PR #1144, opened 2016-11-15
Refactor PlotSchema

etpinard:
This PR significantly refactor `PlotSchema.get()` in an effort to reduce the level of (hard-coded) abstractions required to built the schema and to make it play better for custom modules.

In brief,

- the circa 2014 `_nestedModules` and `_composedModules` attribute [abstractions](https://github.com/plotly/plotly.js/blob/0620416577baf26dcd7501ee19494a7e5af66cdf/src/plots/layout_attributes.js#L178) are gone
- `Plotly.registry` now builds [`Plots.layoutArrayContainer`](https://github.com/plotly/plotly.js/blob/0620416577baf26dcd7501ee19494a7e5af66cdf/src/plots/plots.js#L1617) automatically cc @rreusser 
- add `frames` and animation attributes to `Plotly.PlotSchema.get()` output
- makes array containers schema tree creation more robust - see https://github.com/plotly/plotly.js/commit/b56bf0c0378c623cd80abdab5848675a46e976d7
- fix bug where  `rangeslider` and `rangeselector` where listed as available under `layout.yaxis` 
- removes some useless exports on `Annotations`
- removes a bunch of circular dependencies cc https://github.com/plotly/plotly.js/issues/236

Issue #420, opened 2016-04-13
wishlist for potential breaking changes since v1


etpinard:
_Leave this text box for planned improvements_ 
- [ ] upgrade `d3` to [v4.0.0](https://github.com/d3/d3/blob/master/CHANGES.md#changes-in-d3-40) (now v5) - should be mostly a matter of updating the general update [pattern](https://bl.ocks.org/mbostock/a8a5baa4c4a470cda598).
- [x] remove [`es-promise`](https://github.com/stefanpenner/es6-promise) polyfill from bundle - and thus make IE users add their own Promise polyfill.
- [ ] drop jQuery event support
- [ ] Make components (e.g. `Annotations`, `Shapes`, `RangeSlider`, ...) register-able and remove them from the core bundle.
- [ ] revamp set of default colorscales
- [ ] make config arguments consistent https://github.com/plotly/plotly.js/issues/839
- [ ] remove `config.plot3dPixelRatio` [map](https://github.com/plotly/plotly.js/blob/6b197a0f8f7ec808097e19d8531718d3f2a10dd0/src/plot_api/plot_api.js#L386-L389) for backward compatibility
- [x] Incorporate /remove all unofficial exposed methods (i.e Plotly.Plots, Plotly.Fx, Plotly.Snapshot, Plotly.PlotSchema, Plotly.Queue)
- [ ] drop event-based `Plotly.Snapshot.toImage` and merge `src/snapshot` in `plot_api/to_image.js`
- [ ] remove (some) components for `lib/core.js` - see https://github.com/plotly/plotly.js/pull/845
- [ ] drop `Plotly.relayout` handlers for the `'remove'` and `'add'` special values - see https://github.com/plotly/plotly.js/pull/1086 
- [ ] for log axes, have `layout` express everything in data units instead of linearized units. This applies to `range`, `tick0`, and annotation and image positions.
- [ ] remove support for dates provided as epoch milliseconds
- [ ] remove the distinction between axis name and axis id (ie `xaxis2` vs `x2`) probably by only using the name.
- [ ] rename current gl marker symbol `'cross'` -> `'cross-thin'` (i.e. revert https://github.com/plotly/plotly.js/pull/1482/commits/c0eb06599db481f2137ed97e79bab5cb2bc31351)

etpinard:
On my list:

[...]

PR #1188, opened 2016-11-22
Config options coerce / schema

etpinard:
This PR builds on top of https://github.com/plotly/plotly.js/pull/1144 to bring `config` options into the plot schema.

image

This should be pretty useless in our API libraries - as config options will be listed as first classed attributes on-par with `"data"` / `"layout"`.

This PR also make `gd._context` be filled using `Lib.coerce` making config option input more predictable internally.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -28,6 +28,7 @@ var ErrorBars = require('../components/errorbars');
 var xmlnsNamespaces = require('../constants/xmlns_namespaces');
 var svgTextUtils = require('../lib/svg_text_utils');
 
+var defaultConfig = require('./plot_config');
 var helpers = require('./helpers');
 var subroutines = require('./subroutines');
 
@@ -80,6 +81,7 @@ Plotly.plot = function(gd, data, layout, config) {
 
     // transfer configuration options to gd until we move over to
     // a more OO like model
+
     setPlotContext(gd, config);
 
     if(!layout) layout = {};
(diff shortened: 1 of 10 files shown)


etpinard:
cc @cldougl @Kully it would be nice for `plotly.py` to use plot schema to fill in this [list](https://github.com/plotly/plotly.py/blob/5ba9fa703ff16fab117ad192955e146ffef4bf86/plotly/offline/offline.py#L189-L209) instead of the current hard-coded not-in-sync nightmare.

etpinard:
Moreover, (cc @cldougl @chriddyp @ellecj ) with this PR, the config options declarations now have the same format as the data and layout attributes. It would be nice to eventually display them on https://plot.ly/javascript/reference/ instead of the lonely, outdated https://plot.ly/javascript/configuration-options/

src/plot_api/plot_api.js
```diff
@@ -363,32 +365,49 @@ Plotly.plot = function(gd, data, layout, config) {
...
+    }
+
+    // 'opaque' is special value for setBackground
+    if(config.setBackground === 'opaque') {
+        context.setBackground = opaqueSetBackground;
+    }
+
+    // map plot3dPixelRatio to plotGlPixelRatio for backward compatibility
+    if(config.plot3dPixelRatio && !context.plotGlPixelRatio) {
+        context.plotGlPixelRatio = context.plot3dPixelRatio;
```
alexcjohnson: shouldn't this happen to the input object before coercing, so it can be validated? Anyway if it stays here it would need to be `context.plotGlPixelRatio = config.plot3dPixelRatio;`

etpinard: Thanks for the comment. I discovered some breaking behavior.

As this PR isn't important to anyone right now, I'll drop it from the `1.21.1` release.

etpinard:
Now that https://github.com/plotly/plotly.js/pull/1895 is merged, this PR is now very outdated, so I'll close it.

Go to https://github.com/plotly/plotly.js/issues/420#issuecomment-316410692 for a discussion on  potential v2 config improvements.

PR #1191, opened 2016-11-22
tickvals / ticktext edge cases

### LINKED ISSUE #738: Inconsistency when setting `tickvals` and omitting `ticktext` in date axes
Please see https://jsfiddle.net/0cjprqgy/10/
All ticks of date axis display as "hh:mm:ss aN".

alexcjohnson:
Fixes #738 as well as a bunch of other odd edge cases around tickvals with or without ticktext on all axis types.

@etpinard

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -679,11 +679,11 @@ axes.calcTicks = function calcTicks(ax) {
     // show the exponent only on the last one
     ax._tmax = vals[vals.length - 1];
 
-    // for showing date suffixes: ax._prevSuffix holds what we showed most
-    // recently. Start with it cleared and mark that we're in calcTicks (ie
-    // calculating a whole string of these so we should care what the previous
-    // suffix was!)
-    ax._prevSuffix = '';
+    // for showing the rest of a date when the main tick label is only the
+    // latter part: ax._prevDateHead holds what we showed most recently.
+    // Start with it cleared and mark that we're in calcTicks (ie calculating a
+    // whole string of these so we should care what the previous date head was!)
+    ax._prevDateHead = '';
     ax._inCalcTicks = true;
 
     var ticksOut = new Array(vals.length);
@@ -1107,10 +1107,10 @@ function formatDate(ax, out, hover, extraPrecision) {
     var x = out.x,
         tr = ax._tickround,
         d = new Date(x),
-        // suffix completes the full date info, to be included
+        // headPart completes the full date info, to be included
         // with only the first tick or if any info before what's
         // shown has changed
-        suffix,
+        headPart,
         tt;
     if(hover && ax.hoverformat) {
         tt = modDateFormat(ax.hoverformat, x);
@@ -1129,12 +1129,12 @@ function formatDate(ax, out, hover, extraPrecision) {
         else if(tr === 'm') tt = monthFormat(d);
         else {
             if(tr === 'd') {
-                suffix = yearFormat(d);
+                headPart = yearFormat(d);
 
                 tt = dayFormat(d);
             }
             else {
-                suffix = yearMonthDayFormat(d);
+                headPart = yearMonthDayFormat(d);
 
                 tt = minuteFormat(d);
                 if(tr !== 'M') {
@@ -1151,8 +1151,8 @@ function formatDate(ax, out, hover, extraPrecision) {
         // we get extra precision in array mode or hover,
         // but it may be useless, strip it off
         if(tt === '00:00:00' || tt === '00:00') {
-            tt = suffix;
-            suffix = '';
+            tt = headPart;
+            headPart = '';
         }
         else if(tt.length === 8) {
             // strip off seconds if they're zero (zero fractional seconds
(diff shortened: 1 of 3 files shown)


src/plots/cartesian/axes.js
```diff
@@ -1136,9 +1147,26 @@ function formatDate(ax, out, hover, extraPrecision) {
...
+        }
+        else {
+            tt = tt.replace(/:00$/, '');
+        }
+    }
+
+    if(suffix) {
+        if(hover) {
+            // hover puts it all on one line, so suffix works best up front
+            tt = suffix + (tt ? ' ' + tt : '');
```
alexcjohnson: Previously, hover text on date axes would *not* get the suffix at all. This changes it for all date axes, not just those with array ticks, to include the suffix (as a prefix... hmm, maybe we should change this name...) which is sometimes more information than you need to interpret it, but it's useful often enough that I think this is the right way to go about it.

src/plots/cartesian/set_convert.js
```diff
@@ -313,6 +313,14 @@ module.exports = function setConvert(ax) {
             return c === -1 ? BADNUM : c;
         };
 
+        ax.d2l_noadd = function(v) {
+            // d2c variant that that won't add categories but will also
+            // allow numbers to be mapped to the linearized axis positions
+            var index = ax._categories.indexOf(v);
+            if(index !== -1) return index;
+            if(typeof v === 'number') return v;
+        };
```
alexcjohnson: before this change, sometimes tickvals could add categories to the list, but it happened after autorange calculations so they didn't show up (until you zoom out on the axis). It was weird. Now that can't happen. At some point we should probably look through all other uses of `ax.d2l` (or `ax.d2c` and `ax.d2r`, which are all the same function for category axes) and make sure they shouldn't also be using `d2l_noadd`... but so far I haven't seen any other ill effects.

src/plots/cartesian/axes.js
etpinard: Hmm. 

Looks we're getting a little too much information on regular date axis (e.g. the `date_axis` mock)

gifrecord_2016-11-22_170028

test/jasmine/tests/axes_test.js
```diff
@@ -1535,19 +1535,19 @@ describe('Test axes', function() {
...
                 });
 
                 var expectedText = [
                     'One',
                     '...and a half', // the first two get explicit labels
                     '2.7', // 2.6999999 gets rounded to 2.7
-                    '3',
-                    '3.999' // 3.999 does not get rounded
+                    '30',
+                    '39.999' // 39.999 does not get rounded
```
etpinard: Bonus points for reusing an existing test case. :tada:

src/plots/cartesian/axes.js
```diff
@@ -856,7 +872,7 @@ function autoTickRound(ax) {
         // not necessarily *all* the information in tick0 though, if it's really odd
         // minimal string length for tick0: 'd' is 10, 'M' is 16, 'S' is 19
         // take off a leading minus (year < 0 so length is consistent)
-        var tick0ms = Lib.dateTime2ms(ax.tick0),
+        var tick0ms = Lib.dateTime2ms(ax.tick0) || 0,
```
etpinard: Is that `|| 0` fallback testing somewhere? It looks fragile.

test/jasmine/tests/axes_test.js
```diff
@@ -1499,5 +1499,87 @@ describe('Test axes', function() {
...
+                tickvals: [
+                    '2012-01-01',
+                    new Date(2012, 2, 1).getTime(),
+                    '2012-08-01 00:00:00',
+                    '2012-10-01 12:00:00',
+                    new Date(2013, 0, 1, 0, 0, 1).getTime(),
+                    '2010-01-01', '2014-01-01' // off the axis
+                ],
+                // only the first two have text
+                ticktext: ['New year', 'February'],
```
etpinard: This looks pretty useful. Nice feature to have in the :bank:

src/plots/cartesian/axes.js
```diff
@@ -704,8 +711,11 @@ function arrayTicks(ax) {
     // except with more precision to the numbers
     if(!Array.isArray(text)) text = [];
 
+    // make sure showing ticks doesn't accidentally add new categories
+    var tickVal2l = ax.type === 'category' ? ax.d2l_noadd : ax.d2l;
```
alexcjohnson: right, like https://github.com/plotly/plotly.js/pull/1191#discussion_r89206759 - there are a bunch of uses outside `makeCalcdata`, particularly in gl code that we need to sort out first.

src/plots/cartesian/axes.js
alexcjohnson: Ah, that's an artifact of an older iteration... not needed anymore.

src/plots/cartesian/axes.js
alexcjohnson: Ah... yep, I see where that's coming from. Looks like I need to add some more explicit tests of `Axes.tickText` with hover; the regular (non-hover) cases should all be covered by the `calcTicks` test cases.

alexcjohnson:
Man, this code has a ridiculous number of edge cases, but all the ones I can think of are tested now (including the `date_axis` hovertext you pointed out). Any more?

src/plots/cartesian/axes.js
```diff
@@ -679,11 +679,11 @@ axes.calcTicks = function calcTicks(ax) {
...
-    // for showing date suffixes: ax._prevSuffix holds what we showed most
-    // recently. Start with it cleared and mark that we're in calcTicks (ie
-    // calculating a whole string of these so we should care what the previous
-    // suffix was!)
-    ax._prevSuffix = '';
+    // for showing the rest of a date when the main tick label is only the
+    // latter part: ax._prevDateHead holds what we showed most recently.
+    // Start with it cleared and mark that we're in calcTicks (ie calculating a
+    // whole string of these so we should care what the previous date head was!)
+    ax._prevDateHead = '';
```
etpinard: thanks!

test/jasmine/tests/axes_test.js
```diff
@@ -1578,8 +1616,98 @@ describe('Test axes', function() {
...
+                ['2000-01-01 11:00:15', 'Jan 1, 2000, 11:00:15'],
+                ['2000-01-01 11:00:00.1', 'Jan 1, 2000, 11:00'],
+                ['2000-01-01 11:00:00.0001', 'Jan 1, 2000, 11:00']
+            ]);
+
+            ax.dtick = 1000; // one second
+            ax.range = ['1999-12-31 23:59', '2000-01-01 00:01'];
+            mockCalc(ax);
+
+            checkHovers(ax, [
```
etpinard: looks good!

PR #1172, opened 2016-11-19
use UTC milliseconds for internal representation of dates

alexcjohnson:
@etpinard this builds on the ISO-8601 PR #1158 so note the base branch but we can merge directly to master if #1158 goes in first.

Changes our internal date linearization to use UTC rather than local milliseconds. Every day on a Plotly graph will now be 24 hours long, with no daylight shifts.

Note that for backward compatibility I still have us treat millisecond values (in ranges etc) as they were calculated previously, in local time, and shift them to UTC. I do the same thing with js date objects, on the theory that users generally construct these by using `new Date(y, m, d, H, M, S, ms)` with the date/time components they expect to see on the final graph.

If and when we add support for explicit timezone settings, we can do smarter things for users who specify a timezone, but I think this should remain the default behavior.

Fixes #171 and #811

Issue #811, opened 2016-08-03
Default date ticks are drawn for each day at time: 01:00:00 rather than 00:00:00 


cldougl:
Example: http://codepen.io/plotly/pen/grdEaN?editors=1010

It would be more correct to have the default tick for a day be drawn at time: 00:00:00 rather than the first hour.

etpinard:
pinging @alexcjohnson (again)

alexcjohnson:
I'm pretty sure this is the daylight saving time bug, will go away when we switch to UTC internally, which is part 2 of my cleanup plan, once nothing user-facing involves epoch milliseconds. Thanks for pointing me to this @etpinard

alexcjohnson:
closed by #1194

PR #1194, opened 2016-11-23
Dates: Accept ISO-8601 format, and use UTC milliseconds as the backend

### LINKED ISSUE #171: Plotting issues with daylight savings time
In my timezone (America/Denver), when plotting times during the day that daylight savings time went into effect in the United States, plotted points are off by 1 hour.

The following example plots March 8th @ 11pm, and March 9th @ 12am.

Expected: these points should plot as 1 hour apart.
Actual: these points are plotting as the same point in time, as 12am on March 9th.

Code:

``` javascript
var data = [
  {
    x: ['2015-03-08 23:00:00', '2015-03-09 00:00:00'],
    y: [1, 3],
    type: 'scatter'
  }
];

```

Details: in my cursory tests, all points between March 8th @ 3am (inclusive) and March 9th @ 12 am (exclusive) are plotted 1 hour off, due to confusion with Daylight Savings time.

Tested against plotly.js v1.3.1 (minified)

I suggest that any fix for this should also be tested against points during the day when daylight savings time ended in the U.S., which is November 1, 2015

### LINKED ISSUE #1003: Support ISO 8601 timestamps
It seems that plotly.js does not recognize an ISO 8601 timestamp with a character other than a space separating the date and time, e.g.:
`2016-10-14 18:04:52.142018000` works, but
`2016-10-14T18:04:52.142018000` is not parsed as a timestamp.

Obviously the "T" can be replaced with a space, but it would be nice to support the output of many standard JSON serializes.

alexcjohnson:
### First part: ISO-8601:
fixes #1003

For simplicity, I ended up separately allowing the date/time separator to be space, t, or T, and a timezone spec at the end - ie timezones are now accepted for our normal space-separated format too.

Also @etpinard a perf note - I switched from the manual string chopping we had before to a regexp-based conversion. When I originally wrote this converter a few years ago it was clearly faster to do it manually, but now either because our patterns have gotten more complicated, or browsers have improved their regexp engines, it's faster this way. It's also a lot shorter code-wise (though I wouldn't exactly call it nicer to look at, with that mile-long regexp!)

On Chrome, the old way (extended to ISO-8601) was taking ~2microsec per conversion, the new way is ~1.3microsec. FF is quite a bit slower at both versions, but the regexp is faster there too: ~6.8microsec vs ~10.7microsec.

### The rest: use UTC milliseconds as the back end:
Changes our internal date linearization to use UTC rather than local milliseconds. Every day on a Plotly graph will now be 24 hours long, with no daylight shifts.

Note that for backward compatibility I still have us treat millisecond values (in ranges etc) as they were calculated previously, in local time, and shift them to UTC. I do the same thing with js date objects, on the theory that users generally construct these by using new Date(y, m, d, H, M, S, ms) with the date/time components they expect to see on the final graph.

If and when we add support for explicit timezone settings, we can do smarter things for users who specify a timezone, but I think this should remain the default behavior.

Fixes #171 and #811 

This PR supersedes #1158 and #1172


src/lib/dates.js
```diff
@@ -21,6 +20,11 @@ var ONEHOUR = constants.ONEHOUR;
 var ONEMIN = constants.ONEMIN;
 var ONESEC = constants.ONESEC;
 
+var DATETIME_REGEXP = /^\s*(-?\d\d\d\d|\d\d)(-(0?[1-9]|1[012])(-([0-3]?\d)([ Tt]([01]?\d|2[0-3])(:([0-5]\d)(:([0-5]\d(\.\d+)?))?(Z|z|[+\-]\d\d:?\d\d)?)?)?)?)?\s*$/m;
```
etpinard: Did you cook this up yourself or find from a google search? If the latter, can you paste the reference above?

test/jasmine/tests/lib_date_test.js
```diff
@@ -105,7 +117,9 @@ describe('dates', function() {
                 '2015-01-00', '2015-01-32', '2015-02-29', '2015-04-31', '2015-01-001', // bad day (incl non-leap year)
                 '2015-01-01 24:00', '2015-01-01 -1:00', '2015-01-01 001:00', // bad hour
                 '2015-01-01 12:60', '2015-01-01 12:-1', '2015-01-01 12:001', '2015-01-01 12:1', // bad minute
-                '2015-01-01 12:00:60', '2015-01-01 12:00:-1', '2015-01-01 12:00:001', '2015-01-01 12:00:1' // bad second
+                '2015-01-01 12:00:60', '2015-01-01 12:00:-1', '2015-01-01 12:00:001', '2015-01-01 12:00:1', // bad second
+                '2015-01-01T', '2015-01-01TT12:34', // bad ISO separators
+                '2015-01-01Z', '2015-01-01T12Z', '2015-01-01T12:34Z05:00', '2015-01-01 12:34+500', '2015-01-01 12:34-5:00' // bad TZ info
```
etpinard: looks good.

test/jasmine/tests/lib_date_test.js
```diff
@@ -191,19 +222,20 @@ describe('dates', function() {
...
-        it('should convert any number or js Date within range to a date string', function() {
+        it('should convert numbers or js Dates to strings based on local TZ', function() {
             [
                 new Date(0),
                 new Date(2000),
                 new Date(2000, 0, 1),
                 new Date(),
-                new Date(-9999, 0, 1),
-                new Date(9999, 11, 31, 23, 59, 59, 999)
+                new Date(-9999, 0, 3), // we lose one day of range +/- tzoffset this way
```
etpinard: thanks for that comment

etpinard:
... and don't forget to close https://github.com/plotly/plotly.js/issues/171 and https://github.com/plotly/plotly.js/issues/811 after merging.

test/image/mocks/date_axes.json
```diff
@@ -2,16 +2,89 @@
...
+    "yaxis4": {"domain": [0.48, 0.52]},
+    "yaxis5": {"domain": [0.64, 0.68]},
+    "yaxis6": {"domain": [0.80, 0.84]},
+    "yaxis7": {"domain": [0.96, 1]},
+    "xaxis2": {"anchor": "y2"},
+    "xaxis3": {"anchor": "y3"},
+    "xaxis4": {"anchor": "y4"},
+    "xaxis5": {"anchor": "y5"},
+    "xaxis6": {"anchor": "y6"},
+    "xaxis7": {"anchor": "y7"}
```
alexcjohnson: @etpinard good point about including the UTC behavior in the mocks. I replaced the simple `date_axes.json` with this one that includes tick scales all the way from full years to partial seconds.

Issue #360, opened 2016-03-25
Variable bin sizes for histograms


pollackscience:
It would be useful to allow for variable-sized bins for histogram plotting.  User should be able to give a list of bin edges to xbins or ybins (or some similar attribute), similar to functionality available in matplotlib.

naodell:
I would also like to see this feature implemented.

dschouten:
I second this request!

ndawe:
Would love to have ability to specify the bin edges manually now after wanting to show bins with visually the same width on a histogram with a logarithmic x-axis.

ndawe:
The race is on with Bokeh: https://github.com/bokeh/bokeh/issues/3398 :wink:

rajsahae:
+1
Could we get an estimate of when this could go in? Otherwise I would need to switch to use a different charting library.

sweatyrichard:
Agreed, this is a standard configuration attribute for histograms.

elv1s42:
Any updates on this one?
Would be very useful.

chriddyp:
Any companies out there interested in sponsoring this item?
https://plot.ly/products/consulting-and-oem/

ice-ctrl:
2022. Still hoping for this feature.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #1201, opened 2016-11-26
Date bin shift

### LINKED ISSUE #1151: Extend histogram bin auto-shifting algorithm to dates
We have an algorithm that shifts numeric histogram bins if a lot of the data is near the bin edges - for example, if all data are integers, we'll shift the bin edges to half integers. We should extend this to dates so that if for example all data is on the first of months, the bin edges get put at the middle of months.

This goes along with the desire to show more meaningful bin position hover labels - instead of a single x value as we do with bars, we want to show either both ends (perhaps clipped to what's *actually* in the bin?), or the single value represented if there is only one. Not sure if we have an issue for that already...

alexcjohnson:
Fixes #1151 

Implements an auto-shifting algorithm for nonuniform (n-month) bins in date histograms. Covers 3 cases:
- Most data is on Jan 1 with no time -> shift bins to start July 2 12:00, which is exactly halfway through a non leap year.
- Most data is on the first of months with no time -> shift bins to start on the 16th of every month at noon. This is exactly halfway through 31-day months, which ends up meaning that any month *after* one with 31 days, the hovertext will say the first of that month, months after shorter months the month label will always be correct, but the date/time will be 12-36 hours into that month.
- Most data has no time, only date -> shift bins 12 hours.

A couple of comments
- "Most" was semi-arbitrarily taken to mean 80% of non-blank data. That's a much stronger criterion than we used for numeric bin shifting (there it's >30% near the edges and <10% near the middles, and "near" wasn't exact like it is here) because I figured with date data you basically either have a level of precision (months, days, hours) as part of your data or you don't, whereas numeric data has a lot more variety in the levels of precision that exist.
- In the future, we might want to change the hover format for the date axis if we've found a histogram like this. This might actually be a more generic situation, where we can look at all the data on that axis and see what level of precision it has. Beyond the scope of this PR.


src/plots/cartesian/axes.js
```diff
@@ -507,7 +507,6 @@ axes.autoBin = function(data, ax, nbins, is2d) {
         var distinctData = Lib.distinctVals(data),
             msexp = Math.pow(10, Math.floor(
                 Math.log(distinctData.minDiff) / Math.LN10)),
-            // TODO: there are some date cases where this will fail...
```
alexcjohnson: removed because I have no idea anymore what this was about. We'll wait for a bug report...

test/jasmine/tests/histogram2d_test.js
```diff
@@ -91,8 +91,7 @@ describe('Test histogram2d', function() {
 
             // TODO: even though the binning is done on non-uniform bins,
             // the display makes them linear (using only y0 and dy)
-            // when we sort out https://github.com/plotly/plotly.js/issues/1151
-            // lets also make it display the bins with nonuniform size
+            // Can we also make it display the bins with nonuniform size?
```
alexcjohnson: Removed from the scope of this PR because it turned out to be more of a pain than I thought, and it's a really minor effect visually.

test/jasmine/tests/histogram_test.js
```diff
@@ -156,7 +156,7 @@ describe('Test histogram', function() {
                 nbinsx: 4
             });
 
-            // Note: this gives half-day gaps between all but the first two
+            // TODO: this gives half-day gaps between all but the first two
             // bars. Now that we have explicit per-bar positioning, perhaps
             // we should fill the space, rather than insisting on equal-width
             // bars?
```
alexcjohnson: This effect is a little more important than getting the exact right bin sizes in the 2D case, because here it leads to visible gaps between many of the bars. Unfortunately it's also quite a bit more effort to solve when you consider stacked or grouped traces. Worth making an issue for it?

src/plots/cartesian/axes.js
```diff
@@ -506,102 +507,186 @@ axes.autoBin = function(data, ax, nbins, is2d) {
...
+    if(typeof dummyAx.dtick === 'number') {
+        binStart = autoShiftNumericBins(binStart, data, dummyAx, dataMin, dataMax);
 
-        var bincount = 1 + Math.floor((datamax - binstart) / dummyax.dtick);
-        binend = binstart + bincount * dummyax.dtick;
+        var bincount = 1 + Math.floor((dataMax - binStart) / dummyAx.dtick);
+        binEnd = binStart + bincount * dummyAx.dtick;
     }
     else {
+        // month ticks - should be the only nonlinear kind we have at this point.
```
etpinard: ... after `traces/histogram/clean_bins.js` that is, correct?

src/plots/cartesian/axes.js
```diff
@@ -506,102 +507,186 @@ axes.autoBin = function(data, ax, nbins, is2d) {
...
+
+    dataCount = data.length - blankCount;
+
+    // include bigger exact dates in the smaller ones
+    exactMonths += exactYears;
+    exactDays += exactMonths;
+
+    // unmber of data points that needs to be an exact value
+    // to shift that increment to (near) the bin center
+    var threshold = 0.8 * dataCount;
```
etpinard: non- :no_entry: It would be nice to put these auto-algo threshold fractions in a `constants.js` file somewhere. It's not obvious where that file should be though.

src/plots/cartesian/axes.js
```diff
@@ -930,10 +1014,23 @@ axes.tickIncrement = function(x, dtick, axrev) {
...
+         * So shift 3 days forward, THEN set the new month, then unshift:
+         *   1/31 -> 2/28 (or 29) -> 3/31 -> 4/30 -> ...
+         *
+         * Note that odd behavior still exists if you start from the 26th-28th:
+         *   1/28 -> 2/28 -> 3/31
+         * but at least you can't shift any dates into the wrong month,
+         * and ticks on these days incrementing by month would be very unusual
+         */
+        var y = new Date(x + THREEDAYS);
+        return y.setUTCMonth(y.getUTCMonth() + dtSigned) - THREEDAYS;
```
etpinard: > Anyway I think the advantages far outweigh this risk.

:+1:

test/jasmine/tests/histogram2d_test.js
```diff
@@ -91,15 +91,11 @@ describe('Test histogram2d', function() {
 
             // TODO: even though the binning is done on non-uniform bins,
             // the display makes them linear (using only y0 and dy)
-            // when we sort out https://github.com/plotly/plotly.js/issues/1151
-            // lets also make it display the bins with nonuniform size,
-            // and ensure we don't generate an extra bin on the end (see
-            // first row of z below)
-            expect(out.y0).toBe('1969-07-02 14:24');
-            expect(out.dy).toBe(365.2 * oneDay);
+            // Can we also make it display the bins with nonuniform size?
```
etpinard: ref https://github.com/plotly/plotly.js/issues/360

src/plots/cartesian/axes.js
alexcjohnson: Alright, I'll leave this as is for now but I'm happy to change it if you tell me where to put them. I don't have a good feel for what would make this easy to grok.

etpinard: No action required in this PR.

Issue #558, opened 2016-05-22
Bug: Rogue Line in scatterplot in some environments


alexander-alvarez:
it seems there's a regression:

v1.9.0:

![image](https://cloud.githubusercontent.com/assets/11779993/15451439/6f589862-1f90-11e6-94a9-2602dfeca329.png)

v.1.10.0-v1.11.0
![image](https://cloud.githubusercontent.com/assets/11779993/15451427/004afbea-1f90-11e6-9e82-80c55eea9d20.png)

using `git bisect` I narrowed it down to: https://github.com/plotly/plotly.js/commit/b6d993006264a6f5e66b9f8d519f7eb44a2e46ea

not sure what information I can provide to better help out, but just let me know so I can help out.

mdtusz:
That commit was part of a branch that removed the remaining nested `<svg>` tags for compatibility with Illustrator, so the structure of plots needed to be changed a bit, so it's very possible I missed something in the logic that creates the `clipPath` for the plot. Could you provide the sample data/layout you used to produce this and I'll look into it ASAP.

mdtusz:
That commit was part of a branch that was removing the remaining nested `<svg>` tags for compatibility with Adobe Illustrator. It required changing from a `<svg>` tag that was constrained by it's `viewBox` to a `<g>` tag that has an associated clip - it's most likely that the clip is being miscalculated in the scenario provided.

Could you share the data and layout you have used for this example so we could reproduce? I can't seem to make a (not) working example myself.

alexander-alvarez:
so trying to reproduce it in a fresh environment and no luck...
http://codepen.io/anon/pen/reXdVV

After the `Plotly.relayout()` it seems things go AWOL

would styling/position not captured here affect this?

mdtusz:
It's _possible_, but unlikely. Are you on the most recent release for your local environment? It looks as if the clip is not being properly set on the plot `<g>` element. Which browser are you using? I didn't see any significant inconsistencies when implementing the clip, but it's very possible I may have overlooked something - any extra info will be helpful to debug.

alexander-alvarez:
Running: Chrome 49.0.2623.110 under Linux and Mac

seems like it might be a clip issue:
when [this](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/line_points.js#L53) gets called during relayout, I get negative x values for the points that render to the left of the axes: 

what attributes on objects / functions should i be looking at pertaining to the 'clip'
[...]

PR #1203, opened 2016-11-28
add id to location href w/o hashes when setting clip paths urls

etpinard:
fixes https://github.com/plotly/plotly.js/issues/558 (hopefully :smile:)

See https://github.com/plotly/plotly.js/issues/558#issuecomment-262853476 for rationale.


mdtusz:
That's a pretty funny bug - surprising it took this long to catch! Looks like this should be fairly robust and solve it though.

💃

Issue #1174, opened 2016-11-21
Opacity on line series not reflected properly in legend


adamborowski:
When I set serie opacity to 0.2 for example, I expect to either legend to be unaffected or legend item to have opacity applied to both dot and line
![image](https://cloud.githubusercontent.com/assets/3614676/20487643/7adb008a-b004-11e6-830a-738a199dc8a3.png)

PR #1204, opened 2016-11-28
Apply trace opacity to 'mode: 'lines' legend items

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1174


test/image/mocks/line_scatter.json
```diff
@@ -30,7 +30,8 @@
         9
       ],
       "mode": "lines",
-      "type": "scatter"
+      "type": "scatter",
+      "opacity": 0.5
```
alexcjohnson: Would be a better test to set the opacity on a lines+markers trace - which I don't think is handled correctly in this PR, trace opacity should get set on `traceGroup` rather than independently to lines and markers (and fills!).

etpinard: First, see

image

> trace opacity should get set on traceGroup

that would imply that the trace name (rendered in the `<text class="legendtext">` node) would be semi-transparent whenever `trace.opacity < 1`. 

So, in https://github.com/plotly/plotly.js/pull/1204/commits/d0097980ae8786d4d77348de585a8946771bcfc9 a new `<g>` is added grouping the fill, lines and symbols `<g>` and applying `trace.opacity` once.

alexcjohnson:
> So, in d009798 a new <g> is added grouping the fill, lines and symbols <g> and applying trace.opacity once.

👍 

and then a test with lines+markers? Ideally with different line and marker colors?

Do you know why [28.png changed](https://github.com/plotly/plotly.js/pull/1204/files#diff-c6c980e600bb49bb47c36649957e4925) the way it did? I don't think the new one looks wrong per se, but in addition to the slight transparency to the legend lines, it seems like the lines got a little bit shorter. And in some of the other cases the semitransparent lines got a little thinner. Any idea why?

etpinard:
> and then a test with lines+markers?

Hopefully somethint like https://github.com/plotly/plotly.js/pull/1204/commits/849782d6b25e61cf0e815dd43d7a89ca1dcea323 is what you had in mind.

> I don't think the new one looks wrong per se, but in addition to the slight transparency to the legend lines,

as it should, all traces in `28.json` have `opacity: 0.84` which wasn't honored previously.

> it seems like the lines got a little bit shorter.

I can't detect a difference to be honest, maybe I'm going blind :eyeglasses:

alexcjohnson:
> > it seems like the lines got a little bit shorter.
>
> I can't detect a difference to be honest, maybe I'm going blind :eyeglasses: 

Ok, don't worry about it. Probably just some different antialiasing algorithm when opacity is included.

> Hopefully something like 849782d is what you had in mind.

Technically it does show what I wanted to see - it will fail if the marker and line get `trace.opacity` applied separately rather than to the containing group - though that failure would be rather subtle, especially since you included marker opacity. I was imagining something that will make it blazingly obvious how the opacity was applied, like:
```
Plotly.newPlot(gd,[{
    x: [1, 2, 3],
    y: [1, 2, 3],
    opacity: 0.2,
    line: {width: 10, color: 'red'},
    marker: {size: 20, color: 'blue'}
}], {
    showlegend: true
})
```
And the legend matches the plot:
screen shot 2017-01-11 at 12 27 30 pm
but if it was done wrong, the legend would look like:
screen shot 2017-01-11 at 12 22 28 pm

test/image/mocks/line_scatter.json
```diff
@@ -55,6 +55,13 @@
       "marker": {
         "opacity": 0.707
       }
+    },
+    {
+      "x": [1, 2, 3],
+      "y": [1, 2, 3],
+      "opacity": 0.2,
+      "line": { "width": 10, "color": "red"},
+      "marker": { "size": 20, "color": "blue"}
```
etpinard: image

looks like it is done _right_.

alexcjohnson: Beautiful, thanks :) 💃

etpinard:
Dropping from the `1.22.0` milestone and merging as the `1.21.x` series will get another patch release early next week.

flying-sheep:
Hi! I have a plot where I set per-marker opacity using a array.

The legend takes the opacity of the more transparent points. How can I force the legend’s opacity to be 1?

Issue #810, opened 2016-08-03
Update menus open items


etpinard:
Layout menu feature PR: https://github.com/plotly/plotly.js/pull/770

Open items:
- Should we do some validation for 'args' depending on `method` or just let Plotly[method] error out?
- Should add more global settings:
  - width
  - height (instead of being inferred by the label dimensions)
  - title (above header)
  - header borderwidth
  - active / hover bgcolor
  - gap between buttons
  - gap between header & buttons
- We should probably make dropdown items scrollable
- Should the dropdown arrow direction depend on `direction` see https://github.com/plotly/plotly.js/pull/974#discussion_r80271589

jackparmer:
> We should probably make dropdown items scrollable

Yep. Here's someone struggling with this out-in-the-wild. Also came up from a customer @charleyferrari chatted up today. 

https://plot.ly/~Scochran/159/price-vs-date/

rreusser:
Closing #890 in favor of this. I'll only add that if menus scrolled, that would presumably also solve the hard limit of six items (a very hard limit—the menu disappears if you exceed six).

rreusser:
@etpinard — What do you think about changing the expanded menu to HTML? It will be a bit awkward to position relative to the top-level SVG container, but it would probably be significantly more accessible and require 80% less code. Plus, it's not clear to me why you'd want the *expanded* menu to be rendered to SVG. Just a thought as I look through what it means to implement complex controls in SVG.

Kully:
Another user who is having trouble accessing menu items past the 10th one or so: https://plot.ly/~wholleran/829/completion-percentage-vs-period/?share_key=J0gquGi2vrEhTurmsYsP3s

jackparmer:
Hey @n-riesco - Can you take this before you go onto animations for bars? (At least making it scroll-able). Thanks!

etpinard:
Closing as the main item on the list :arrow_double_up: was completed in https://github.com/plotly/plotly.js/pull/1214 and https://github.com/plotly/plotly.js/issues/1345 repeats the title options. The other items on the list are of much lesser priority.

PR #1214, opened 2016-12-01
scrollable dropdown menus

n-riesco:
This PR addresses one of the points in https://github.com/plotly/plotly.js/issues/810 :

* provides a helper class ScrollBox (that could be reused to provide scroll boxes elsewhere)

* implements scrollable dropdown menus:
  * scrollbars are added only if needed
  * user can scroll by dragging a scrollbar
  * user can scroll by dragging the scrollbox (menu buttons skip the click triggered by `dragend`)

TODO

* [x] Add jasmine tests

EXAMPLE

Here's an example to test the functionality:

```js
// Adapted from https://github.com/plotly/plotly.js/pull/770#issuecomment-234669383

var direction = 'down',
    data = [],
    layout = {
        updatemenus: [{
            direction: direction,
            buttons: [{
                method: 'restyle',
                args: ['line.color', 'red'],
                label: 'red'
            }, {
                method: 'restyle',
                args: ['line.color', 'blue'],
                label: 'blue'
            }, {
                method: 'restyle',
                args: ['line.color', 'green'],
                label: 'green'
            }]
        }, {
            y: 0.7,
            direction: direction,
            buttons: []
        }]
    };

for (var i = 0, n = 20; i < n; i++) {
    data.push({
        y: Array.apply(null, Array(10)).map(() => Math.random()),
        line: {
            shape: 'spline',
            color: 'red'
        },
        visible: i === 0,
        name: 'Data set ' + i,
    });

    var visible = Array.apply(null, Array(n)).map((_, j) => i === j);

    layout.updatemenus[1].buttons.push({
        method: 'restyle',
        args: ['visible', visible],
        label: 'Data set ' + i
    });
}

Plotly.plot('graph', data, layout);
```


etpinard:
Some suggestions:

- It would be nice to add a wheelevent scroll handler, similar to how long vertical legends can be scrolled
- When clicking on the header, I think the set of dropped buttons should be centred on the active button

src/components/updatemenus/scrollbox.js
```diff
@@ -0,0 +1,381 @@
...
+    };
+
+    // compute position of horizontal scroll bar
+    var needsHorizontalScrollBar = (w > boxW),
+        hbarW = ScrollBox.barLength + 2 * ScrollBox.barPad,
+        hbarH = ScrollBox.barWidth + 2 * ScrollBox.barPad,
+        hbarL = boxL,
+        hbarT = (boxB + hbarH < fullHeight) ? boxB : fullHeight - hbarH;
+
+    var hbar = this.container.selectAll('rect.scrollbar-horizontal').data(
```
n-riesco: If we cut horizontal scrollbars, dropdown menus with `direction: 'left' | 'right'` will be limited to fit within the plot area.

For legends, we could make Scrollbox provide the option to disable the horizontal scrollbar.

etpinard:
@n-riesco are you available to finish off https://github.com/plotly/plotly.js/pull/1214#issuecomment-264277841 in the short term?

etpinard:
@n-riesco Looking good. This will make a bunch of users happy :tada: 

I'm wondering if you noticed that menus that have a scrollbox are a little _jumpy_ on click:

gifrecord_2017-01-30_170812

I can help you debug this tomorrow if you like.

n-riesco:
@etpinard I hadn't noticed. I guess that it's caused by removing the clip path before the buttons.

I haven't tested it, but the fix may be as simple as removing [this line](https://github.com/n-riesco/plotly.js/blob/ef5210e97f583beaf23ef99e3fbaa62e9717ed00/src/components/updatemenus/draw.js#L319).

src/components/updatemenus/draw.js
```diff
@@ -315,8 +337,32 @@ function drawButtons(gd, gHeader, gButton, menuOpts) {
...
-    Lib.setTranslate(gButton, menuOpts.lx, menuOpts.ly);
+    scrollBox.enable();
+
+    var active = menuOpts.active,
+        i;
+    if(isVertical) {
+        if(scrollBox._vbar) {
+            var translateY = 0;
+            for(i = 0; i < active; i++) {
+                translateY += menuOpts.heights[i] + constants.gapButton;
```
etpinard: This looks off for `direction: 'up'` when the updatemenu `y` isn't the default value:

```js
var direction = 'up',
    data = [],
    layout = {
        updatemenus: [{
            y: 0,
            direction: direction,
            buttons: []
        }]
    };

for (var i = 0, n = 20; i < n; i++) {
    data.push({
        y: Array.apply(null, Array(10)).map(() => Math.random()),
        line: {
            shape: 'spline',
            color: 'red'
        },
        visible: i === 0,
        name: 'Data set ' + i,
    });

    var visible = Array.apply(null, Array(n)).map((_, j) => i === j);

    layout.updatemenus[0].buttons.push({
        method: 'restyle',
        args: ['visible', visible],
        label: 'Data set ' + i
    });
}

Plotly.plot('graph', data, layout);
```

gives (after clicking on the header)

image

src/components/updatemenus/scrollbox.js
```diff
@@ -0,0 +1,406 @@
...
+    var vbar = this.container.selectAll('rect.scrollbar-vertical').data(
+            (needsVerticalScrollBar) ? [0] : []);
+
+    vbar.exit()
+        .on('.drag', null)
+        .remove();
+
+    vbar.enter().append('rect')
+        .classed('scrollbar-vertical', true)
+        .call(Color.fill, ScrollBox.barColor);
```
etpinard: Would it be ok to chain

```js
.attr('opacity', '0')
.transition()
.attr('opacity', '1')
```

here and for `hbar` to make the scroll bar appear at the same time as the buttons?

src/components/updatemenus/draw.js
n-riesco: Oh! The [position of the scrollbox](https://github.com/n-riesco/plotly.js/blob/13508daf68fcb77499d2b3975448675386753af3/src/components/updatemenus/draw.js#L302) is wrong! This happens in the `up` direction, when the scrollbox doesn't fit within the chart height. Something similar may happen in the `left` direction.

src/components/updatemenus/scrollbox.js
```diff
@@ -0,0 +1,406 @@
...
+    // set initial position
+    this.setTranslate();
+};
+
+/**
+ * If present, remove clip-path and scrollbars
+ *
+ * @method
+ */
+ScrollBox.prototype.disable = function disable() {
```
etpinard: Looks like https://github.com/plotly/plotly.js/pull/1214/commits/9e3ac1b73bad1ae13a4ad41b1ee23d9d8aab528c made this method obsolete :hocho:

src/components/updatemenus/scrollbox.js
```diff
@@ -246,6 +246,8 @@ ScrollBox.prototype.enable = function enable() {
                 .on('.drag', null)
                 .call(onBarDrag);
         }
+
+        this.container.on('wheel', this._onBoxWheel.bind(this));
```
etpinard: Currently, the wheel handler isn't called when one tries to scroll with the cursor over the gap between the buttons (you can try increasing the `gapButton` constant) to make this problem more obvious).

I think the easiest solution would be to make `Scrollbox` add a background `<rect>` as big as the clipped `<rect>` in the container and attach the `wheel` handler on it.

src/components/updatemenus/draw.js
```diff
@@ -253,11 +254,19 @@ function drawButtons(gd, gHeader, gButton, menuOpts) {
 
         exit.transition()
             .attr('opacity', '0')
-            .remove();
+            .remove()
+            .each('end', function() {
```
etpinard: gifrecord_2017-01-31_143701

This makes the scrollbox disappear after the button exit transition is completed (you can chain `.delay(1000)` to the transition to make the problem easier to detect). 

I would prefer making the scrollbox disappear at the same time as the buttons.

src/components/updatemenus/draw.js
```diff
@@ -281,13 +290,23 @@ function drawButtons(gd, gHeader, gButton, menuOpts) {
...
+        y: menuOpts.ly + y0 + menuOpts.pad.t,
         yPad: constants.gapButton,
         xPad: constants.gapButton,
         index: 0,
     };
 
+    var fullLayout = gd._fullLayout,
+        scrollBoxId = 'updatemenus' + fullLayout._uid + menuOpts._index,
+        scrollBoxPosition = {
+            l: menuOpts.lx + menuOpts.borderwidth + x0 + menuOpts.pad.l,
```
etpinard: Nice job taking into account `borderwidth` :tada:

src/components/updatemenus/scrollbox.js
etpinard: Working like a charm now. Thanks!

src/components/updatemenus/scrollbox.js
etpinard: Working great now :+1:

src/components/updatemenus/draw.js
etpinard: Looking good now :+1:

src/components/updatemenus/scrollbox.js
etpinard: The comment :arrow_up: is now obsolete

src/components/updatemenus/draw.js
etpinard: This one hasn't been addressed yet, correct?

etpinard: Looking great now!

etpinard:
Great @n-riesco looks like all the functionality is :ok_hand: 

Ping me once you fixed the merge conflict and are happy with your test cases.

Awesome work :+1:

n-riesco:
@etpinard Ready for review!

This PR also fixes `Drawing.getTranslate` (the previous regexp dropped `-` signs).

test/jasmine/tests/updatemenus_test.js
```diff
@@ -800,3 +801,406 @@ describe('update menus interaction with other components:', function() {
         .then(done);
     });
 });
+
+
+describe('update menus interaction with scrollbox:', function() {
```
etpinard: :100: / :100: tests. Thanks!

src/lib/index.js
```diff
@@ -453,7 +453,7 @@ lib.addStyleRule = function(selector, styleString) {
 
 lib.getTranslate = function(element) {
 
-    var re = /.*\btranslate\((\d*\.?\d*)[^\d]*(\d*\.?\d*)[^\d].*/,
+    var re = /.*\btranslate\((-?\d*\.?\d*)[^-\d]*(-?\d*\.?\d*)[^\d].*/,
```
etpinard: Nice catch :microscope: 

Would you mind adding a test case in [`drawing_test.js`](https://github.com/plotly/plotly.js/blob/dc7621760c2f264ab075ff285ac8e72901939d03/test/jasmine/tests/drawing_test.js#L72) to :lock: this down?

(No rush on this by the way. If you don't have the time to do this by Monday night, I'll merge your PR as is and open up an issue about this).

etpinard:
https://github.com/plotly/plotly.js/pull/1214#discussion_r99401471 is non-blocking.

This PR will be merged and released as part of `1.23.0` on Feb 6.

etpinard:
Merging. Thanks @n-riesco :tada:

ayh9kim:
Will this work for Ropensci/plotly automatically?

PR #1220, opened 2016-12-04
World calendars

alexcjohnson:
@etpinard @n-riesco plots all around the world!

A few TODOs left:

- [x] test against gl2d and gl3d plots
- [x] test against candlestick and OHLC
- [x] get Chinese into `world-calendars` ( @n-riesco is working on this)
- [x] test global `layout.calendar`

The bulk of this PR is in the two "add world calendar support" commits, but it doesn't make sense to look at the two of them separately, the first is incomplete and did some things wrong that the second completed and fixed.


src/plots/cartesian/set_convert.js
```diff
@@ -67,24 +82,170 @@ module.exports = function setConvert(ax) {
         else return BADNUM;
     }
 
-    function fromLog(v) {
-        return Math.pow(10, v);
+    /*
+     * wrapped dateTime2ms that:
+     * - accepts ms numbers for backward compatibility
+     * - inserts a dummy arg so calendar is the 3rd arg (see notes below).
```
alexcjohnson: There's a lot of rearrangement in this commit, but there are two key points. This is one: calendar is always the 3rd argument to these functions - an extremely unfortunate necessity, since we need to allow different calendar systems on the same axis.

src/plots/cartesian/set_convert.js
```diff
@@ -67,24 +82,170 @@ module.exports = function setConvert(ax) {
...
+    ax.p2l = p2l;
+
+    ax.c2p = (ax.type === 'log') ? function(v, clip) { return l2p(toLog(v, clip)); } : l2p;
+    ax.p2c = (ax.type === 'log') ? function(px) { return fromLog(p2l(px)); } : p2l;
 
     /*
-     * fn to make sure range is a couplet of valid & distinct values
+     * now type-specific conversions for **ALL** other combinations
+     * they're all written out, instead of being combinations of each other, for
+     * both clarity and speed.
```
alexcjohnson: ... and this is the other key change here: now *all* possible conversions between d, r, c, l, and p are generated (so you don't need to worry about whether the one you want exists), and they're simplified as much as possible, separately for each axis type.

src/lib/index.js
```diff
@@ -135,6 +135,22 @@ lib.identity = function(d) { return d; };
...
+ * array: the array to map it to
+ * func: the function to apply
+ * x1, x2: optional extra args
+ */
+lib.simpleMap = function(array, func, x1, x2) {
+    var len = array.length,
+        out = new Array(len);
+    for(var i = 0; i < len; i++) out[i] = func(array[i], x1, x2);
+    return out;
+};
```
alexcjohnson: because of the extra args to the axis conversion functions, statements like `ax.range.map(ax.r2l)` would behave... strangely to say the least. This was actually already true for log axes, but it didn't seem to have bitten us yet. This helper fixes that and lets you pass the extra (constant) args straight through. Not super excited about its name, but I wanted to make sure people didn't use it in place of built-in map without understanding it first, that's why I didn't just call it `lib.map`.

test/image/mocks/world-cals.json
```diff
@@ -0,0 +1,443 @@
...
+                "5200-01-03",
+                "5200-01-03",
+                "5200-01-03",
+                "5201-04-01",
+                "5199-01-01"
+            ],
+            "xaxis": "x2",
+            "yaxis": "y2",
+            "ycalendar": "hebrew",
+            "name": "hebrew histogram<br>jalali filtered",
```
alexcjohnson: and displayed against nepali dates - 3 calendars applied to one array!

src/plots/cartesian/set_convert.js
alexcjohnson: you mean the `setConvert` closure? Because each trace can apply its own calendar to this axis and `setConvert` only happens per axis. But at least the axis internals only need to pass the first arg because it defaults to `ax.calendar`.

alexcjohnson:
> Down the road (but hopefully soon), we should require world-calendars in a component module.

I think this would be easy, actually... [here](https://github.com/plotly/plotly.js/blob/86b31eaff0ce6243a7f445e27d1bbb91d7f88ace/src/lib/dates.js#L40) is the only place we directly call out to `world-calendars` (though we probably need another in [`isWorldCalendar`](https://github.com/plotly/plotly.js/blob/86b31eaff0ce6243a7f445e27d1bbb91d7f88ace/src/lib/dates.js#L44)) I just don't know the right pattern - @etpinard if you've got a few free cycles would you want to turn this into a pluggable component?

src/traces/mesh3d/attributes.js
```diff
@@ -37,6 +37,9 @@ module.exports = {
             'jointly represent the X, Y and Z coordinates of the nth vertex.'
         ].join(' ')
     },
+    xcalendar: surfaceAtts.xcalendar,
+    ycalendar: surfaceAtts.ycalendar,
+    zcalendar: surfaceAtts.zcalendar,
```
alexcjohnson: > Hmm. Do dates actually work in mesh3d traces?

[yes](https://github.com/plotly/plotly.js/pull/1220/commits/86b31eaff0ce6243a7f445e27d1bbb91d7f88ace)

alexcjohnson: actually it's surface that has a problem with dates, because `colorscale` and `colorbar` don't do dates... you'll notice it's there but I had to explicitly include `surfacecolor`. We could support this too, but another time.

test/jasmine/tests/lib_date_test.js
```diff
@@ -222,6 +222,68 @@ describe('dates', function() {
                 expect(Lib.ms2DateTime(Lib.dateTime2ms(v[0]), v[1])).toBe(v[2], v);
             });
         });
+
+        it('should work right with inputs beyond our precision', function() {
```
alexcjohnson: worked fine *before* all this crap with regular dates... during my work it broke with regular and world dates, now it works for all.

src/transforms/filter.js
```diff
@@ -188,9 +200,11 @@ function getDataToCoordFunc(gd, trace, target) {
 
         setConvert(ax);
 
-        // build up ax._categories (usually done during ax.makeCalcdata()
-        for(var i = 0; i < target.length; i++) {
-            ax.d2c(target[i]);
+        if(ax.type === 'category') {
```
alexcjohnson: optimization... though it could have led to spurious confusing "invalid date" console messages if your target values are valid in their own calendar but not in the axis calendar.

src/plots/cartesian/set_convert.js
```diff
@@ -112,26 +113,23 @@ module.exports = function setConvert(ax) {
...
         for(i = 0; i < 2; i++) {
             if(ax.type === 'date') {
-                if(!Lib.isDateTime(range[i])) {
+                if(!Lib.isDateTime(range[i], ax.calendar)) {
                     ax[rangeAttr] = dflt;
                     break;
                 }
 
-                if(range[i] < Lib.MIN_MS) range[i] = Lib.MIN_MS;
-                if(range[i] > Lib.MAX_MS) range[i] = Lib.MAX_MS;
```
alexcjohnson: This was a relic (albeit not a very old one, as `MIN_MS` and `MAX_MS` only came in with my work a month or so back...) as `cleanDate` / `dateTime2ms` won't accept strings that give invalid ms output.

src/lib/dates.js
```diff
@@ -204,17 +275,194 @@ function includeTime(dateStr, h, m, s, msec10) {
...
+ *  Date formatting for ticks and hovertext
+ */
+
+/*
+ * convert d3 templates to world-calendars templates, so our users only need
+ * to know d3's specifiers. Map space padding to no padding, and unknown fields
+ * to an ugly placeholder
+ */
+var UNKNOWN = '##';
+var d3ToWorldCalendars = {
```
alexcjohnson: https://github.com/d3/d3-3.x-api-reference/blob/master/Time-Formatting.md

Good call - we'll have to check that when we upgrade d3.

src/lib/dates.js
```diff
@@ -49,6 +49,85 @@ function isWorldCalendar(calendar) {
 // of the unix epoch. From calendars.instance().newDate(1970, 1, 1).toJD()
 var EPOCHJD = 2440587.5;
 
+// each calendar needs its own default canonical tick. I would love to use
+// 2000-01-01 (or even 0000-01-01) for them all but they don't necessarily
+// all support either of those dates. Instead I'll use the most significant
+// number they *do* support, biased toward the present day.
+var CANONICAL_TICK = {
```
alexcjohnson: maybe the world-calendars *plotly.js component*... the package is a pretty thin wrapper on kbwood/calendars so in principle I'd like it to stay in sync with the original.

src/traces/mesh3d/attributes.js
etpinard: Great. Thanks for checking :+1:

etpinard:
> I just don't know the right pattern - @etpinard if you've got a few free cycles would you want to turn this into a pluggable component?

Sure. I'll give this a go this afternoon.

src/plots/cartesian/set_convert.js
etpinard: I knew there was a good reason. Thanks for the info!

I still think we could come up with a better pattern here. Having to remember to pass `ax.calendar` in the trace module (or transform?) calc steps could potentially produce bugs in future PRs. 

Let me test out a few things. Anyhow, I'm _ok_ with the current implementation.

src/lib/dates.js
etpinard: Good call. I've give this a shot!

src/lib/coerce.js
```diff
@@ -267,6 +268,20 @@ exports.valObjects = {
...
+        description: [
+            'A string, one of the calendar systems available',
+            'in the `world-calendars` package, to be used in evaluating',
+            'or displaying date data. Defaults to built-in (Gregorian) dates.',
+            'available calendars:', '*' + calendarList.join('*, *') + '*'
+        ].join(' '),
+        requiredOpts: [],
+        otherOpts: ['dflt'],
+        coerceFunction: function(v, propOut, dflt) {
+            if(v && calendarList.indexOf(v) !== -1) propOut.set(v);
```
etpinard: I'll see what I can do here.

src/traces/scatter/attributes.js
```diff
@@ -41,8 +41,14 @@ module.exports = {
...
         ].join(' ')
     },
+    xcalendar: {
+        valType: 'calendar',
+        role: 'info',
+        description: 'Sets the calendar system to use with `x` date data'
+    },
     y: {
         valType: 'data_array',
+        role: 'info',
```
alexcjohnson: oh good catch - mispaste...

package.json
```diff
@@ -89,7 +89,8 @@
     "superscript-text": "^1.0.0",
     "tinycolor2": "^1.3.0",
     "topojson-client": "^2.1.0",
-    "webgl-context": "^2.2.0"
+    "webgl-context": "^2.2.0",
+    "world-calendars": "0.1.0"
```
alexcjohnson: 👍  when we get @n-riesco 's Chinese in `world-calendars` I'll change the pattern here - and probably bump it to 1.0.0 at that point since it's doing its job.

src/plots/cartesian/set_convert.js
alexcjohnson: yeah, it's a painful pattern for sure. A lot of trace types never see it though, as it's centralized in `ax.makeCalcdata` which gets the trace and data attribute as args already so it can look up the calendar. But if anyone can think of a better pattern I would be happy to refactor (after a bit of a break 😅  )

But - and this is perhaps an important point - the way I have it implemented traces do NOT inherit the axis calendar, so actually you never need to pay attention to `ax.calendar`, just the calendar that's specified in the trace itself. The only calendar inheritance is `layout.calendar` which sets the default for everything else. That's intended for what I expect will really be the most common use case of this, people really living by one of these calendars who want all their date data and all their date axes to use that calendar.

src/traces/scatter/attributes.js
alexcjohnson: https://github.com/plotly/plotly.js/pull/1220/commits/4e9a632aff0eac49e63852b91abd5b68ca1e4a0d

src/components/rangeslider/draw.js
```diff
@@ -352,7 +353,8 @@ function drawRangePlot(rangeSlider, gd, axisOpts, opts) {
 
         mockFigure.layout[oppAxisName] = {
             domain: [0, 1],
-            range: oppAxisOpts.range.slice()
+            range: oppAxisOpts.range.slice(),
+            calendar: oppAxisOpts.calendar
```
alexcjohnson: @etpinard apologies for adding a nearly duplicate mock to test this... but since the fix is in `draw` I didn't really see another way. I tried to at least put both rangesliders together on one plot but it didn't seem to like that at all.

src/components/rangeselector/get_update_object.js
```diff
@@ -42,13 +39,13 @@ function getXRange(axisLayout, buttonLayout) {
...
         case 'backward':
-            range0 = Lib.ms2DateTime(+d3.time[step].utc.offset(base, -count));
+            range0 = axisLayout.l2r(+d3.time[step].utc.offset(base, -count));
             break;
 
         case 'todate':
             var base2 = d3.time[step].utc.offset(base, -count);
 
-            range0 = Lib.ms2DateTime(+d3.time[step].utc.ceil(base2));
+            range0 = axisLayout.l2r(+d3.time[step].utc.ceil(base2));
```
alexcjohnson: This is a weird one. At first when I played with range selectors it seemed like they *just worked* :tm: - if only we were that lucky. They worked (to within errors I would not have been able to notice) as long as the range end is valid in the gregorian calendar and the new start is valid in the axis calendar, but there's no guarantee that that will be true in general. The fix here at least means we properly convert between range values and milliseconds.

What it *doesn't* get right is stepping backward by some months or years - these steps are going to be somewhat different in each calendar.

What it *really* doesn't get right is month or year to date. It's going to go to the beginning of the Gregorian month or year, which is generally totally different from in the other calendar. I propose to leave this as an open item for now, because it's going to have all sorts of edge cases to sort out (like in the Hebrew calendar, when the year changes at month 7, not month 1) for a niche-upon-niche feature.

src/lib/dates.js
```diff
@@ -204,17 +379,321 @@ function includeTime(dateStr, h, m, s, msec10) {
...
+    return dateStr + (headStr ? '\n' + headStr : '');
+};
+
+/*
+ * incrementMonth: make a new milliseconds value from the given one,
+ * having changed the month
+ *
+ * special case for world calendars: multiples of 12 are treated as years,
+ * even for calendar systems that don't have (always or ever) 12 months/year
+ * TODO: perhaps we need a different code for year increments to support this?
```
alexcjohnson: @etpinard I meant to flag this one for you - there would be something nice even with regular dates about allowing `Y1` and `Y5` etc instead of requiring `M12` and `M60` for 1 and 5 year ticks, even though for Gregorian they're equivalent. And if we allow `M12` to mean a year in every calendar *now* it will be tough to undo later if we decide to add `Y1`. On the other hand, someone actually wanting 12-month ticks in a calendar that doesn't have 12 months seems ridiculously unlikely.

src/components/rangeslider/draw.js
etpinard: That's fine :+1: 

By the way, I've been wanting to add support for multiple range sliders per plot for a while now. Would be a nice feature.

src/components/rangeselector/get_update_object.js
etpinard: >  I propose to leave this as an open item for now, because it's going to have all sorts of edge cases to sort out 

I agree :+1: . But let's make the range selector supplyDefaults doesn't allow `axis.calendar !== 'gregorian'` with `axis.rangeselector.buttons[i].stepmode = 'todate'`

test/image/mocks/candlestick_double-y-axis.json
```diff
@@ -1947,8 +1947,8 @@
   "layout": {
     "xaxis": {
       "range": [
-        1452453861917,
-        1471240462978
+        "2016-01-10 14:24:21.917",
+        "2016-08-15 01:54:22.978"
       ]
```
alexcjohnson: https://github.com/plotly/plotly.js/pull/1220/commits/6653da7dbbe79591d6eeb6fdb5203eebc9fa09a8

src/components/calendars/index.js
```diff
@@ -58,11 +59,13 @@ var CANONICAL_TICK = {
...
 // Start on a Sunday - for week ticks
-// Discworld and Mayan calendars don't have 7-day weeks anyway so don't change them.
+// Discworld and Mayan calendars don't have 7-day weeks but we're going to give them
+// 7-day week ticks so start on our Sundays.
 // If anyone really cares we can customize the auto tick spacings for these calendars.
 var CANONICAL_SUNDAY = {
+    chinese: '2000-01-02',
     coptic: '2000-01-03',
-    discworld: '2000-01-01',
+    discworld: '2000-01-03',
```
etpinard: bug fix or typo?

alexcjohnson: see the comment change above - I decided it would be better to keep our Sundays unless and until we support the native weeks in these calendars.

src/lib/dates.js
```diff
@@ -167,11 +171,19 @@ exports.dateTime2ms = function(s, calendar) {
...
+        y = Number(y);
 
         var cDate;
         try {
-            cDate = Registry.getComponentMethod('calendars', 'getCal')(calendar)
-                .newDate(Number(y), m, d);
+            var calInstance = Registry.getComponentMethod('calendars', 'getCal')(calendar);
+            if(isChinese) {
+                var isIntercalary = m.charAt(m.length - 1) === 'i';
+                m = Number(isIntercalary ? m.substr(0, m.length - 1) : m);
```
alexcjohnson: ah good call - I always forget about parseInt.

test/jasmine/tests/lib_date_test.js
```diff
@@ -341,6 +396,51 @@ describe('dates', function() {
         });
     });
 
+    describe('incrementMonth', function() {
+        it('should include Chinese intercalary months', function() {
```
n-riesco: I should add a similar tests to world-calendars

package.json
```diff
@@ -90,7 +90,7 @@
     "tinycolor2": "^1.3.0",
     "topojson-client": "^2.1.0",
     "webgl-context": "^2.2.0",
-    "world-calendars": "0.1.0"
+    "world-calendars": "^1.0.0"
```
etpinard: :beers:

etpinard:
Looks good to me :dancer: after that `parseInt` patch.

test/jasmine/tests/lib_date_test.js
```diff
@@ -341,6 +396,51 @@ describe('dates', function() {
...
+                '2003-06-01',
+                '2004-06-01',
+                '2005-06-01',
+                '2006-06-01',
+                '2007-06-01',
+                '2008-06-01'
+            ];
+            var tick = Lib.dateTime2ms(start, 'chinese');
+            expected.forEach(function(v) {
+                tick = Lib.incrementMonth(tick, 12, 'chinese');
```
n-riesco: 12? I guess `incrementMonth` does some magic over leap years

alexcjohnson: yes exactly: https://github.com/plotly/plotly.js/pull/1220#discussion_r91010748

n-riesco:
I had a quick look through the Chinese commits (I didn't go through the functions in Lib). LGTM

src/components/rangeselector/get_update_object.js
alexcjohnson: no year/month todate (but smaller increments are still allowed) range selectors outside Gregorian calendars: https://github.com/plotly/plotly.js/pull/1220/commits/7bd501fbae2bdf6112df98685292584515e1ba78

PR #1223, opened 2016-12-05
Heatmap contour refactor

rreusser:
Unless something has gone wrong, this PR makes absolutely no changes to the functioning of the codebase. Instead, it separates some of the contour and heatmap logic into separate files.

In one place I've made marginally more substantial change change:

```javascript
z = cleanData(trace);
```

becomes:

```javascript
z = cleanData(trace.z, trace.transpose);
```

since `cleanData` otherwise has no particular dependence on the variable being the `z` property. There is no change to the behavior though. It simply moves in the direction of being able to apply some of the same functions to, say, `trace.y` for carpet plots. There will be more of those changes, but the goal here is to keep things simple since this first step involves moving large chunks of code. Otherwise this PR should be (to the very best of my knowledge) harmless.


rreusser:
ping @alexcjohnson @etpinard 

No big rush, but nothing here is particularly offensive to me so I'm currently building off of this in my carpet/scattercarpet/contourcarpet trace types

src/traces/heatmap/calc.js
```diff
@@ -20,6 +18,10 @@ var colorscaleCalc = require('../../components/colorscale/calc');
 var hasColumns = require('./has_columns');
 var convertColumnXYZ = require('./convert_column_xyz');
 var maxRowLength = require('./max_row_length');
+var cleanData = require('./clean_data');
```
etpinard: I would prefer not calling this module `clean_data`.

Trace modules with `#cleanData` methods are called during the `Plotly.plot` pipeline [here](https://github.com/plotly/plotly.js/blob/8265df5acf85ecfe828e25cde97a2a7a5e997694/src/plots/plots.js#L430-L435). At the moment, only `Scatter` has a [`cleanData`](https://github.com/plotly/plotly.js/blob/8265df5acf85ecfe828e25cde97a2a7a5e997694/src/plots/plots.js#L430-L435) method but still adding a `heatmap/clean_data.js` module that has nothing to do with `cleanData` step could potentially be confusing.

I'd vote for `clean_z.js`, but @rreusser I'll let decide on an alternative file name.

etpinard: Oh I see, you generated `cleanZ` to no `'z'` data arrays. 

So maybe `clean_2d_array.js` instead?

src/traces/heatmap/find_empties.js
```diff
@@ -0,0 +1,103 @@
...
+
+var maxRowLength = require('./max_row_length');
+
+module.exports = function findEmpties(z) {
+    // return a list of empty points in 2D array z
+    // each empty point z[i][j] gives an array [i, j, neighborCount]
+    // neighborCount is the count of 4 nearest neighbors that DO exist
+    // this is to give us an order of points to evaluate for interpolation.
+    // if no neighbors exist, we iteratively look for neighbors that HAVE
+    // neighbors, and add a fractional neighborCount
```
etpinard: let's bump that comment block above the `function` statement.

src/traces/heatmap/make_bound_array.js
```diff
@@ -0,0 +1,78 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Registry = require('../../registry');
+
+module.exports = function makeBoundArray(trace, arrayIn, v0In, dvIn, numbricks, ax) {
+    var arrayOut = [],
+        isContour = Registry.traceIs(trace, 'contour'),
```
etpinard: non- :no_entry: , but it would be nice to remove those `Registry.traceIs` calls in `makeBoundArray` and have two (many more) boolean options instead: `withCenteredBricks` (true for contour and gl2d traces) and `startAtZero` (for histogram2d traces and categorical axes)

That way `makeBoundArray` would no longer need `trace` as an argument

etpinard:
:dancer: thanks!

src/traces/heatmap/make_bound_array.js
rreusser: Agreed. If it's alright, I wanted to keep this PR (almost) entirely just trivial file reorganization so that small but meaningful changes to get too lost in gigantic line-count PRs. In other words, ideally I'd like to push that change off until subsequent PRs that continue to aim for small changes.

PR #243, opened 2016-02-08
WIP: Legend Scrolling

mdtusz:
There's still a bug or two to work out, but it's in a working state.

When you toggle a trace for the first time, the width changes and hides the scrollbar until you scroll (for some reason the text-width is being changed when you toggle for the first time).

Otherwise, it will automatically enable the scroll if the height of the list of traces is greater than the height of the plot. I need to add a check whether the scrollbar will be inserted and rework the details on what sets the bounds, but the main logic is there.

PR #1230, opened 2016-12-06
"calendars" component

etpinard:
to be merged in https://github.com/plotly/plotly.js/pull/1220

This PR makes @alexcjohnson's world-calendars a requirable component in `plotly.js/lib/`. This new `"calendars"` component will be included by default only in the main plotly.js bundle. The changes proposed here effectively :hocho: my concerns in https://github.com/plotly/plotly.js/pull/1220#pullrequestreview-11400570 

Off this branch:

image

That the ~200K `world-calendars` module is only bundled in `dist/plotly.js`, the other plotly.js bundle show little :hamburger: increase.

### DIFF
--- a/src/components/calendars/index.js
+++ b/src/components/calendars/index.js
@@ -35,6 +35,7 @@ var handleTraceDefaults = function(traceIn, traceOut, coords, layout) {
         handleDefaults(traceIn, traceOut, coords[i] + 'calendar', layout.calendar);
     }
 };
+
 // each calendar needs its own default canonical tick. I would love to use
 // 2000-01-01 (or even 0000-01-01) for them all but they don't necessarily
 // all support either of those dates. Instead I'll use the most significant
@@ -164,10 +165,77 @@ function getCal(calendar) {
     return calendarObj;
 }
 
+function makeAttrs(description) {
+    return Lib.extendFlat({}, attributes, { description: description });
+}
+
+function makeTraceAttrsDescription(coord) {
+    return 'Sets the calendar system to use with `' + coord + '` date data.';
+}
+
+var xAttrs = {
+    xcalendar: makeAttrs(makeTraceAttrsDescription('x'))
+};
+
+var xyAttrs = Lib.extendFlat({}, xAttrs, {
+    ycalendar: makeAttrs(makeTraceAttrsDescription('y'))
+});
+
+var xyzAttrs = Lib.extendFlat({}, xyAttrs, {
+    zcalendar: makeAttrs(makeTraceAttrsDescription('z'))
+});
+
+var axisAttrs = makeAttrs([
+    'Sets the calendar system to use for `range` and `tick0`',
+    'if this is a date axis. This does not set the calendar for',
+    'interpreting data on this axis, that\'s specified in the trace',
+    'or via the global `layout.calendar`'
+].join(' '));
+
 module.exports = {
     moduleType: 'component',
     name: 'calendars',
 
+    schema: {
+        traces: {
+            scatter: xyAttrs,
+            bar: xyAttrs,
+            heatmap: xyAttrs,
+            contour: xyAttrs,
+            histogram: xyAttrs,
+            histogram2d: xyAttrs,
+            histogram2dcontour: xyAttrs,
+            scatter3d: xyzAttrs,
+            surface: xyzAttrs,
+            mesh3d: xyzAttrs,
+            scattergl: xyAttrs,
+            ohlc: xAttrs,
+            candlestick: xAttrs
+        },
+        layout: {
+            calendar: makeAttrs([
+                'Sets the default calendar system to use for interpreting and',
+                'displaying dates throughout the plot.'
+            ].join(' ')),
+            'xaxis.calendar': axisAttrs,
+            'yaxis.calendar': axisAttrs,
+            'scene.xaxis.calendar': axisAttrs,
+            'scene.yaxis.calendar': axisAttrs,
+            'scene.zaxis.calendar': axisAttrs
+        },
+        transforms: {
+            filter: {
+                calendar: makeAttrs([
+                    'Sets the calendar system to use for `value`, if it is a date.',
+                    'Note that this is not necessarily the same calendar as is used',
+                    'for the target data; that is set by its own calendar attribute,',
+                    'ie `trace.x` uses `trace.xcalendar` etc.'
+                ].join(' '))
+            }
+        }
+    },
+
+    layoutAttributes: attributes,
 
     handleDefaults: handleDefaults,
     handleTraceDefaults: handleTraceDefaults,
--- a/src/components/calendars/index.js
+++ b/src/components/calendars/index.js
@@ -16,6 +16,25 @@ var constants = require('../../constants/numerical');
 var EPOCHJD = constants.EPOCHJD;
 var ONEDAY = constants.ONEDAY;
 
+var attributes = {
+    valType: 'enumerated',
+    values: Object.keys(calendars.calendars),
+    role: 'info',
+    dflt: 'gregorian'
+};
+
+var handleDefaults = function(contIn, contOut, attr, dflt) {
+    var attrs = {};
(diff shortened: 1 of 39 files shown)


src/constants/numerical.js
```diff
@@ -36,5 +36,11 @@ module.exports = {
...
     ONEHOUR: 3600000,
     ONEMIN: 60000,
-    ONESEC: 1000
+    ONESEC: 1000,
+
+    /*
+     * For fast conversion btwn world calendars and epoch ms, the Julian Day Number
+     * of the unix epoch. From calendars.instance().newDate(1970, 1, 1).toJD()
+     */
+    EPOCHJD: 2440587.5
```
etpinard: I'm not super happy about this, but I had to find a way to make `EPOCHJD` requirable in both `lib/dates.js` and `components/calendars/` in a non-circular way

src/lib/dates.js
```diff
@@ -30,102 +32,40 @@ var DATETIME_REGEXP = /^\s*(-?\d\d\d\d|\d\d)(-(\d?\d)(-(\d?\d)([ Tt]([01]?\d|2[0
...
-    calendar = (isWorldCalendar(calendar) && calendar) || 'gregorian';
-    if(sunday) return CANONICAL_SUNDAY[calendar];
-    return CANONICAL_TICK[calendar];
+    if(isWorldCalendar(calendar)) {
+        return sunday ?
+            Registry.getComponentMethod('calendars', 'CANONICAL_SUNDAY')[calendar] :
+            Registry.getComponentMethod('calendars', 'CANONICAL_TICK')[calendar];
+    }
+    else {
+        return sunday ? '2000-01-02' : '2000-01-01';
```
etpinard: maybe these gregorian defaults should be placed elsewhere ...

src/components/calendars/index.js
```diff
@@ -16,6 +16,25 @@ var constants = require('../../constants/numerical');
 var EPOCHJD = constants.EPOCHJD;
 var ONEDAY = constants.ONEDAY;
 
+var attributes = {
+    valType: 'enumerated',
```
etpinard: no need for a `calendar` valType anymore!

src/traces/box/defaults.js
```diff
@@ -33,9 +34,8 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         return;
     }
 
-    var dfltCalendar = layout.calendar;
-    coerce('xcalendar', dfltCalendar);
-    coerce('ycalendar', dfltCalendar);
+    var handleCalendarDefaults = Registry.getComponentMethod('calendars', 'handleTraceDefaults');
+    handleCalendarDefaults(traceIn, traceOut, ['x', 'y'], layout);
```
etpinard: a little ugly, but `Registry` gets it done

src/components/calendars/index.js
```diff
@@ -164,10 +165,77 @@ function getCal(calendar) {
...
+    'if this is a date axis. This does not set the calendar for',
+    'interpreting data on this axis, that\'s specified in the trace',
+    'or via the global `layout.calendar`'
+].join(' '));
+
 module.exports = {
     moduleType: 'component',
     name: 'calendars',
 
+    schema: {
```
etpinard: some fancy (slightly hacky) declaration to make sure that calendar attributes will show up on https://community.plot.ly/

src/constants/numerical.js
alexcjohnson: If the whole contents of the [`if(calendar)`](https://github.com/plotly/plotly.js/pull/1230/files?diff=split#diff-b896a50c4f63f50409418cd75ff7d0aeR430) [blocks](https://github.com/plotly/plotly.js/pull/1230/files?diff=split#diff-b896a50c4f63f50409418cd75ff7d0aeR503) went into the component, you wouldn't need `EPOCHJD` in dates at all, would you?

etpinard: Sure. This will lead to some return object destructuring though. I'll see how bad it gets.

src/plot_api/plot_api.js
```diff
@@ -312,7 +312,7 @@ Plotly.plot = function(gd, data, layout, config) {
 
         // show annotations and shapes
         Registry.getComponentMethod('shapes', 'draw')(gd);
-        Registry.getComponentMethod('annoations', 'draw')(gd);
```
alexcjohnson: how did this not get caught before?

etpinard: Not sure. `Annotations.draw` gets called twice in `Plotly.plot`. Maybe one too many?

etpinard: The second `Annotations.draw` call was added in https://github.com/plotly/plotly.js/pull/243 [here](https://github.com/plotly/plotly.js/pull/243/files?diff=unified#diff-2941ab69a12080c0633ff4ac8ea3aa83R306).

src/traces/histogram/defaults.js
```diff
@@ -26,6 +27,9 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     var x = coerce('x'),
         y = coerce('y');
 
+    var handleCalendarDefaults = Registry.getComponentMethod('calendars', 'handleTraceDefaults');
+    handleCalendarDefaults(traceIn, traceOut, ['x', 'y'], layout);
+
```
alexcjohnson: previously we coerced calendars after the [return on no data](https://github.com/plotly/plotly.js/pull/1230/files?diff=split#diff-995f21c0a277c39bc47f02f65846d42aR40) - was it intentional to move this up?

etpinard: Oops. My bad.

src/plot_api/plot_api.js
alexcjohnson: I wonder if we can run (most of) our tests in a mode where *all* components are registered, and `Registry.getComponentMethod` throws an error if it doesn't find what it's looking for, rather than returning `noop`?

src/constants/numerical.js
etpinard: I'm not sure I want to go through with this.

Moving all the logic from the `if(calendar)` blocks to the calendars component would involved passing oddly named variables to new routines and return sets of values and even mutating them in [`findExactDates`](https://github.com/plotly/plotly.js/blob/68af28745fba7551e57ae166a95e64bad76bf6d0/src/lib/dates.js#L555-L567).

I understand that if would be nice to put all the world-calendar logic _in one place_, but it would make the code less readable, unless we start duplicating some `dates.js` logic in the calendars component.

The current implementation is both DRY and readable - which gets my vote.

src/traces/histogram/defaults.js
etpinard: done in cb2c54ba96c7997b022346258d8dc4617794aea2

src/plot_api/plot_api.js
etpinard: > I wonder if we can run (most of) our tests in a mode where all components are registered

That's currently the case as most tests have `var Plotly = require('@lib/index.js')` which registers all traces and components.

> Registry.getComponentMethod throws an error if it doesn't find what it's looking for, rather than returning noop?

Yeah, we could mutate the `Registry` export in the test suite. Let me try that.

alexcjohnson: cool - not in this PR though, unless it's super easy :)

src/constants/numerical.js
alexcjohnson: Sure, lets leave it as is.

src/components/calendars/index.js
alexcjohnson: how will we know to keep this up to date?

alexcjohnson:
@etpinard this looks ready to me! Remaining comments are nonblocking from my perspective 💃

src/components/calendars/index.js
etpinard: > how will we know to keep this up to date? 

good question. There's got to be a way to test this.

PR #1233, opened 2016-12-07
Add fallbacks for manual manipulation of slider/frames

rreusser:
cc @bpostlethwaite @etpinard 

This PR fixes two cases:

1. When slider options are added, it needs to *refetch* the slider steps from the d3 data. The mouse events are in a closure, so it was previously using outdated data in callbacks.
1. When slider options are removed, it checks whether the active index is invalid. If it is, it just moves the slider to the first position.

This PR does not fix:

- It does not search for a matching option after the update has occurred. Like if you remove the thid option, it will not fix the active index from 5 -> 4. That can be added, but I did not yet address that case in the interest of stopping blocking @bpostlethwaite ASAP.

This PR also sneaks in `Plot.recomputeFrameHash(gd)` which does what it says. I should add tests for that; it's just a small bonus to prevent @bpostlethwaite from needing to do it manually.

### DIFF
--- a/src/components/sliders/draw.js
+++ b/src/components/sliders/draw.js
@@ -72,10 +72,18 @@ module.exports = function draw(gd) {
         computeLabelSteps(sliderOpts);
 
         Plots.manageCommandObserver(gd, sliderOpts, sliderOpts.steps, function(data) {
-            if(sliderOpts.active === data.index) return;
-            if(sliderOpts._dragging) return;
+            // NB: Same as below. This is *not* always the same as sliderOpts since
+            // if a new set of steps comes in, the reference in this callback would
+            // be invalid. We need to refetch it from the slider group, which is
+            // the join data that creates this slider. So if this slider still exists,
+            // the group should be valid, *to the best of my knowledge.* If not,
+            // we'd have to look it up by d3 data join index/key.
+            var opts = gSlider.data()[0];
 
-            setActive(gd, gSlider, sliderOpts, data.index, false, true);
+            if(opts.active === data.index) return;
+            if(opts._dragging) return;
+
+            setActive(gd, gSlider, opts, data.index, false, true);
         });
 
         drawSlider(gd, d3.select(this), sliderOpts);
@@ -225,6 +233,15 @@ function findDimensions(gd, sliderOpts) {
 }
 
 function drawSlider(gd, sliderGroup, sliderOpts) {
+    // This is related to the other long notes in this file regarding what happens
+    // when slider steps disappear. This particular fix handles what happens when
+    // the *current* slider step is removed. The drawing functions will error out
+    // when they fail to find it, so the fix for now is that it will just draw the
+    // slider in the first position but will not execute the command.
+    if(sliderOpts.active >= sliderOpts.steps.length) {
+        sliderOpts.active = 0;
+    }
+
     // These are carefully ordered for proper z-ordering:
     sliderGroup
         .call(drawCurrentValue, sliderOpts)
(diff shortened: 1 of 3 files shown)


src/components/sliders/draw.js
```diff
@@ -72,10 +72,18 @@ module.exports = function draw(gd) {
...
-            if(sliderOpts._dragging) return;
-
-            setActive(gd, gSlider, sliderOpts, data.index, false, true);
+            // NB: Same as below. This is *not* always the same as sliderOpts since
+            // if a new set of steps comes in, the reference in this callback would
+            // be invalid. We need to refetch it from the slider group, which is
+            // the join data that creates this slider. So if this slider still exists,
+            // the group should be valid, *to the best of my knowledge.* If not,
+            // we'd have to look it up by d3 data join index/key.
+            var opts = gSlider.data()[0];
```
etpinard: Does the same behavior happen for `updatemenus` ?

src/components/sliders/draw.js
```diff
@@ -402,11 +419,21 @@ function setActive(gd, sliderGroup, sliderOpts, index, doCallback, doTransition)
...
+function attachGripEvents(item, gd, sliderGroup) {
     var node = sliderGroup.node();
     var $gd = d3.select(gd);
 
+    // NB: This is *not* the same as sliderOpts itself! These callbacks
+    // are in a closure so this array won't actually be correct if the
+    // steps have changed since this was initialized. The sliderGroup,
+    // however, has not changed since that *is* the slider, so it must
+    // be present to receive mouse events.
+    function getSliderOpts() {
```
etpinard: thanks for the comment :books:

src/plots/plots.js
```diff
@@ -1546,6 +1546,23 @@ plots.computeFrame = function(gd, frameName) {
     return result;
 };
 
+/*
+ * Recompute the lookup table that maps frame name -> frame object. addFrames/
+ * deleteFrames already manages this data one at a time, so the only time this
+ * is necessary is if you poke around manually in `gd._transitionData_frames`
```
etpinard: typo :alarm_clock: `gd._transitionData_frames` -> `gd._transitionData._frames`

test/jasmine/tests/sliders_test.js
```diff
@@ -209,6 +210,59 @@ describe('sliders initialization', function() {
...
+            // Remove the option:
+            gd.layout.sliders[0].steps.pop();
+
+            // And redraw the plot:
+            return Plotly.redraw(gd);
+        }).then(function() {
+            // The selected option no longer exists, so confirm it's
+            // been fixed during the process of updating/drawing it:
+            expect(gd._fullLayout.sliders[0].active).toEqual(0);
+        }).catch(fail).then(done);
```
etpinard: very nice test case.

src/components/sliders/draw.js
rreusser: I think so. Need to investigate make a parallel PR for that. This is one is just of particular concern because of @bpostlethwaite's current project with the slider transform thing.

etpinard: Ok. No need to address `updatemenus` in this PR.

Can you open an issue for that and for your _This PR does not fix:_ point above?

etpinard:
@bpostlethwaite wanna try this branch out in the workspace?

src/components/sliders/draw.js
rreusser: See: #1234 #1235

bpostlethwaite:
Yep will do

bpostlethwaite:
Successfully used this in WS2 :+1: from me

etpinard:
@rreusser :dancer: 

@bpostlethwaite I'll release `1.21.0` on Monday.

PR #1236, opened 2016-12-07
Fix number named frames

rreusser:
This PR adds some casting and fixes some checks to ensure that frames with number names are handled correctly. Technically it could break compatibility with current plots, but only in the sense that the current behavior is broken. This seems like an extremely common problem based on what I've observed.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2327,7 +2327,11 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
             var newFrame = trans._currentFrame = trans._frameQueue.shift();
 
             if(newFrame) {
-                gd._fullLayout._currentFrame = newFrame.name;
+                // Since it's sometimes necessary to do deep digging into frame data,
+                // we'll consider it not 100% impossible for nulls or numbers to sneak through,
+                // so check when casting the name, just to be absolutely certain:
+                var stringName = newFrame.name ? newFrame.name.toString() : null;
+                gd._fullLayout._currentFrame = stringName;
 
                 trans._lastFrameAt = Date.now();
                 trans._timeToNext = newFrame.frameOpts.duration;
@@ -2349,7 +2353,7 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
                 });
 
                 gd.emit('plotly_animatingframe', {
-                    name: newFrame.name,
+                    name: stringName,
                     frame: newFrame.frame,
                     animation: {
                         frame: newFrame.frameOpts,
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2528,7 +2528,7 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
 Plotly.addFrames = function(gd, frameList, indices) {
     gd = helpers.getGraphDiv(gd);
 
-    var hasBeenWarnedAboutNumericNames = false;
+    var numericNameWarningCount = 0;
 
     if(frameList === null || frameList === undefined) {
         return Promise.resolve();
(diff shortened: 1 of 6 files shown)


rreusser:
cc: @Kully re: problems with number-named frames

This PR should hopefully fix the issues. It's still slightly sketchy since it looks up frames by name and exact numeric equality is only a good idea with integers, but it is what it is. It will issue a warning if `addFrames` would overwrite with a number-named frame, but other than that it should just work as expected.

See: http://rickyreusser.com/demos/plotly-unsupported/numbered-frames-bugfix.html
and source: https://github.com/rreusser/demos/blob/master/plotly-unsupported/src/numbered-frames-bugfix.js

for an example that mixes strings and numbers with components.

src/plot_api/plot_api.js
```diff
@@ -2558,6 +2564,25 @@ Plotly.addFrames = function(gd, frameList, indices) {
...
+        if(name && newName && typeof newName === 'number' && _hash[name]) {
+            numericNameWarningCount++;
+
+            Lib.warn('addFrames: overwriting frame "' + _hash[name].name +
+                '" with a frame whose name of type "number" also equates to "' +
+                name + '". This is valid but may potentially lead to unexpected ' +
+                'behavior since all plotly.js frame names are stored internally ' +
+                'as strings.');
+
+            if(numericNameWarningCount > 5) {
```
etpinard: nice touch.

test/jasmine/tests/command_test.js
```diff
@@ -444,6 +444,12 @@ describe('Plots.computeAPICommandBindings', function() {
             expect(result).toEqual([{type: 'layout', prop: '_currentFrame', value: 'framename'}]);
         });
 
+        it('treats numeric frame names as strings', function() {
+            var result = Plots.computeAPICommandBindings(gd, 'animate', [[8]]);
+
+            expect(result).toEqual([{type: 'layout', prop: '_currentFrame', value: '8'}]);
```
etpinard: nicely done

test/jasmine/tests/frame_api_test.js
```diff
@@ -73,13 +74,41 @@ describe('Test frame api', function() {
...
             }).catch(fail).then(done);
         });
 
+        it('casts number names to strings on insertion', function(done) {
+            Plotly.addFrames(gd, [{name: 2}]).then(function() {
+                expect(f).toEqual([{name: '2'}]);
+            }).catch(fail).then(done);
+        });
+
+        it('updates frames referenced by number', function(done) {
```
etpinard: sweet :tada:

src/plot_api/plot_api.js
```diff
@@ -2327,7 +2327,11 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
             var newFrame = trans._currentFrame = trans._frameQueue.shift();
 
             if(newFrame) {
-                gd._fullLayout._currentFrame = newFrame.name;
+                // Since it's sometimes necessary to do deep digging into frame data,
+                // we'll consider it not 100% impossible for nulls or numbers to sneak through,
+                // so check when casting the name, just to be absolutely certain:
+                var stringName = newFrame.name ? newFrame.name.toString() : null;
```
etpinard: we _could_ maybe have

```js
var nameIn = (newFrame.name === 'number') ? newFrame.name.toPrecision(1) : newFrame.name;
var stringName = nameIn ? nameIn.toString() : null;
```

to make sure that we only cast integer to string.

-----

This is probably an overkill though. @rreusser I'll let decide what's best.

etpinard:
:dancer: looks good to me.

src/plot_api/plot_api.js
rreusser: Yeah, I published a module once where I said "meh, probably won't be too many warnings." Haha, [WRONG](https://github.com/scijs/integrate-adaptive-simpson/issues/1).

PR #1237, opened 2016-12-07
use only the calendars we need from 'world-calendars'

etpinard:
.. so that the resulting main plotly.js bundle weighs in at:

image

compared to `567 kB` in https://github.com/plotly/plotly.js/pull/1230


etpinard:
cc @alexcjohnson

alexcjohnson:
easy peasy :) If we add localization support later we can bring the rest in as another component.
💃

Issue #760, opened 2016-07-21
Non scaling stroke on zoom.


mdtusz:
This only applies when using `zoomScroll`, but we should add `vector-effect="non-scaling-stroke"` to scatter lines so they don't embiggen and enshrinken too too much.

E.g. when scrolling on @rreusser 's example [here](http://rickyreusser.com/plotly-expression-transform/)

rreusser:
Correct me if I'm wrong, is it's not (reasonably) possible to apply that to markers, is it? Just paths? Or you can use markers but then they all have to be the same etc. so that this can really only be used on paths?

mdtusz:
As far as I know, it only works on paths. We _could_ counteract marker size changes easily enough as well though - just need to set `r = radius/scale` if I'm not mistaken.

rreusser:
SVG elements get redrawn all the time, so not sure there's a meaningful performance overhead if the transform is just computed to a string once and applied a bunch—thought it's a bit ugly.

Added `vector-effect: non-scaling-stroke;` in the browser and appears to work fine, though obviously disappears as soon as you pause and it's redrawn. It's not immediately apparent to me that this would cause other problems since it's only present until it's redrawn—which only really happens on scroll zoom and soon-to-be transitions, unless I'm mistaken.

rreusser:
Is this the spot for it? https://github.com/plotly/plotly.js/blob/master/src/components/drawing/index.js#L84

mdtusz:
I think that would do it - I was testing in `scatter/plot.js`, but `drawing` may have a nicer deeper effect.

rreusser:
Pardon the hacky informality, but deployed a quick PoC with it added on that line here: http://rickyreusser.com/plotly-expression-transform/

that is,

```javascript
drawing.lineGroupStyle = function(s, lw, lc, ld) {
    s.style('fill', 'none')
    .style('vector-effect', 'non-scaling-stroke')
    .each(function(d) {
```

A grep shows it's used in the following places:

```
src/components/colorbar/draw.js
390:                    .call(Drawing.lineGroupStyle,

src/components/drawing/index.js
83:drawing.lineGroupStyle = function(s, lw, lc, ld) {

src/components/legend/style.js
75:    line.call(Drawing.lineGroupStyle);

src/traces/contour/style.js
39:                .call(Drawing.lineGroupStyle,

src/traces/scatter/style.js
34:        .call(Drawing.lineGroupStyle);

[...]

PR #1239, opened 2016-12-08
[WIP] Carpet plots

rreusser:
Trace types to be implemented:

- [ ] trace type: `carpet`
  - [x] x-cheater
  - [ ] y-cheater
  - [x] first cut
  - [x] setconvert
    - [x] linear interpolation
    - [x] bicubic interpolation
    - [x] derivatives
    - [x] fix clipping and detection of out-of-range points
  - [x] attrs to style the axes (implemented but needs feedback)
  - [x] array tickmode label formatting
  - [x] linear tickmode label formatting
  - [x] graceful degradation when bad data is provided
  - [x] play with with cartesian axes/subplots
  - [x] hide cheater axis by default (if no other no-cheater traces visible on that axis)
  - [x] construct clip shape around the axis edge
  - [x] fix (disable?) hover
  - [ ] fix missing minor gridline on `tickmode: 'array'` axes
  - [x] image tests
  - [ ] good jasmine coverage
- [ ] trace type: `scattercarpet`
  - [x] attributes
  - [x] supply defaults
  - [x] basic functionality
  - [x] clip to axis bounds
  - [x] fix hover
  - [ ] jasmine tests
  - [x] image tests
  - [x] break any connections with and filter invalid points
- [ ] trace type: `contourcarpet`
  - [x] clip to axis bounds
  - [x] attributes
  - [x] supply defaults
  - [x] clean data
  - [x] contour lines
  - [x] filled contour inequalities
  - [x] filled contours
  - [ ] jasmine tests
  - [x] image tests
  - [x] fix fully-filled area bug (one side is closed but doesn't follow the edge
  - [x] legend symbols
  - [x] maybe constraints follow non-colorscale syntax?
  - [x] label inequalities on the plot
  - [x] assign to correct subplot based on carpetid (must currently/redundantly be specified separately)

Current state:

screen shot 2016-12-08 at 08 08 19

Live version: http://rickyreusser.com/demos/plotly-unsupported/carpet.html


rreusser:
Interesting question, @etpinard @alexcjohnson:

If we allow scatter plots and inequality contours, I'm not sure, generally speaking that we can guarantee point will lie on the correct side of the contour. Splines handle affine transforms nicely, but if the grid lines themselves are cubic splines, then the *true* contour lines that cut across those grid lines subject to bicubic interpolation would be more complicated than cubic splines themselves and could probably only be drawn correctly by oversampling.

I'm glad to clarify that if needed, but just something I'm realizing as I go through this. That was part of the initial motivation for a high resolution grid. We could subdivide to reduce the problem, but I'm not yet sure what the best option will be.

alexcjohnson:
When you say "the true contour" are you referring to the exact path the contour takes within one grid cell being different if you draw it in a,b space and map back to x,y vs if you drew it in x,y space in the first place? I'm actually not sure if there *is* an objectively true contour in that sense... in fact if you've chosen a really good carpet transformation, the contours you draw in x,y space might even be more meaningful at sub-cell resolution than in a,b space because the variation of physically relevant properties might be more homogeneous than in a,b space (where they could be stretched along a diagonal for example).

Also what do you mean by "high-resolution grid" - isn't the grid just whatever data we get from the user?

rreusser:
Conclusion after talking with @alexcjohnson: The actual real-world implications of this and the cost of doing better (i.e. numerical spline-fitting) make it negligible for now as long as it's documented as such.

rreusser:
Small status update:

I tried to optimize the attributes and styling for a fairly straightforward case. Using 1-d data, this definition:

```javascript
[{
  carpetid: 'c',
  a: [4, 4, 4, 4.5, 4.5, 4.5, 5, 5, 5, 6, 6, 6].map(a => a * 1e-6),
  b: [1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3].map(b => b * 1e6),
  y: [2, 3.5, 4, 3, 4.5, 5, 5.5, 6.5, 7.5, 8, 8.5, 10],
  type: 'carpet',
  aaxis: { 
    tickprefix: 'a = ',
    ticksuffix: 'm',
    smoothing: 1,
    minorgridcount: 9,
  },
  baxis: { 
    tickprefix: 'b = ',
    ticksuffix: 'Pa',
    smoothing: 1,
    minorgridcount: 9,
  }
}, { 
  carpetid: 'c',
  type: 'scattercarpet',
  a: [4, 4.5, 5, 6].map(a => a * 1e-6),
  b: [1, 1, 1, 1].map(b => b * 1e6),
  line: {shape: 'spline', smoothing: 1}
}]
```

with a couple more scattercarpet traces for the sake of illustration gives this plot:

screen shot 2017-01-17 at 14 07 49

rreusser:
One thing I'd forgotten about. Plotly splines are computed in screen space and so are dependent upon the *aspect ratio* of the plot (not both the height and width separately).

I didn't particularly like the idea of the layout of the carpet axis changing depending upon the shape of the plot, so I did not follow that convention here. It's not a strong effect for cheater plots since the x values are computed based on the y values combined with `cheaterslope` which is usually ~1 or so so that there aren't huge scale differences going on here.

The downside though is that it doesn't follow plotly splines.

Do you think I should change this and do all of the computation in screen coordinates?

src/traces/scattercarpet/plot.js
```diff
@@ -0,0 +1,42 @@
...
+    var carpet = data[0][0].carpet;
+
+    // mimic cartesian plotinfo
+    var plotinfo = {
+        xaxis: Axes.getFromId(gd, carpet.xaxis || 'x'),
+        yaxis: Axes.getFromId(gd, carpet.yaxis || 'y'),
+        plot: plotinfoproxy.plot
+    };
+
+    /* var calcdata = new Array(data.length),
```
etpinard: Ha look took that block from `scatternatery`. Looks like your `data` here is in fact calcdata, so you can :hocho: the commented block no :fearful:

rreusser: Haha indeed. There's going to be a lot of review necessary. There's a lot of code here!

alexcjohnson:
> Do you think I should change this and do all of the computation in screen coordinates?

The way you have it is better, and in fact we could consider altering the other splines accordingly, as generally x and y do not share units so they *should* be treated independently.

rreusser:
Thanks for the feedback @alexcjohnson.

One additional question: what do you think is the right behavior for points that fall outside the defined carpet range? The obvious answer is that they should not be plotted since that doesn't even make sense, but the challenges:

1. floating point errors: e.g. `a` only goes up to 10. Your scatter trace has a point with value 10.000000000000007. Is there precedent for selecting a proper tolerance?
2. line traces. You connect point from `(5, 5)` (inside the valid range) to point `(70, 5)` (outside the valid range). Presumably it should detect this, break the curve into segments, and just draw the single point that *is* valid?

alexcjohnson:
> floating point errors: e.g. a only goes up to 10. Your scatter trace has a point with value 10.000000000000007. Is there precedent for selecting a proper tolerance?

Yes. We do this in different (but always simple) ways around the library, I just tightened up the tolerance for contour range to [1ppm here](https://github.com/plotly/plotly.js/pull/1309/files#diff-adc91b6a2df65ee51ff82a3b36078542R17), it used have a tolerance of 10% of contour spacing which had some effects I decided were strange and unexpected. I'm not sure if it makes sense to try and make them all the same - [which axis ticks to draw](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js#L722), for example, gets a little bigger tolerance, my rationale being that it's related to the number of pixels on a screen, and there could be cases where for example an autorange gets very close to zero but off by more than just a floating point rounding error - my inclination is this case is more like contours actually, since you're explicitly specifying the a/b ranges somewhere, but I don't have a super strong opinion.

> line traces. You connect point from (5, 5) (inside the valid range) to point (70, 5) (outside the valid range). Presumably it should detect this, break the curve into segments, and break any connections with points that are not valid.

Yes, we should try to figure out how to display the line in the correct direction as it exits the defined range. For cartesian lines we do this by always drawing the segment that exits, no matter how far it goes, but it gets clipped by the viewport so that won't work for you. FWIW I don't think we handle it correctly necessarily if *both* ends of a segment are outside but the line connecting them passes through the viewport, so I wouldn't worry too much about that case.

rreusser:
@alexcjohnson I did just implement a clipping path that bounds the carpet axis, so it shouldn't be too difficult to just draw something wherever and it'll just get clipped. The bigger concern is extrapolating outside the valid range which is probably fine for small extrapolations. For bicubic though, it probably needs to switch over to linear extrapolation which just adds a bit of complexity. I have all the derivatives though, so it shouldn't be too bad to just extrapolate linearly. It won't be perfect, but if you're plotting points outside your axis, some of that is on the user anyway so it seems reasonable to just do the best we can.

rreusser:
A couple implementation/design issues:

1. Scatter traces are clipped against the carpet axis. This is definitely the most efficient way to do this (with room for additional optimization), though it clips part of the lines/markers:

That is, this:
unclipped

Becomes this:
clipped

Additionally, scatter traces are drawn in *screen space*, which means the lines don't necessarily follow the axes. Here's a simple example:

screen shot 2017-01-19 at 16 43 37

Note that the lines don't reproduce the sharp corners of the carpet axis. Solving this would require a careful geometry-processing pass that is a bit of work for points and likely prohibitively difficult for splines. Submitting for feedback in case these points present any problems.

alexcjohnson:
> markers and lines at the edges get clipped

This is the optimal handling for lines, but it would be awesome if we could filter, rather than clip, the markers - ie show the *entire* marker if it's in or on the boundary, and *no marker* if it's outside the boundary (even close enough that it would partly show, though as discussed before perhaps with a 1ppm or similar buffer for FP rounding errors). Don't worry about it if you think it's more than a couple of hours work though.

> lines don't necessarily follow the axes. ... I think this is best neglected at the moment since scatter traces are not the major focus of carpet plots.

Agreed. It's not ideal, but it's already a bit of an edge case, lets leave it until someone is bothered by it.

rreusser:
Thanks for the feedback @alexcjohnson. Required a bit of SVG reorganization and a small scatter trace patch, but wasn't too bad:

screen shot 2017-01-19 at 18 11 09

FWIW all of the math is done so that you can compute the derivatives relative to a specific position *in* a specific grid cell (so the discontinuities are fine and has chain rule methods etc to get to screen coordinates), so conforming straight lines to the grid is well-defined and doesn't require too much additional work; it's splines that are real tough and probably require a numerical solver. Gonna avoid that for now; just noting here so there's something to go back to when estimating the cost of this addition pops up down the road. 😄

rreusser:
Another update. This screenshot isn't particularly informative, but the point is: fills and splines are working pretty well.

screen shot 2017-01-31 at 17 52 47

Decisions to be made:

- Should the carpet axis grid fall on top of the fill? this seems highly desirable (or at the very least should perhaps be configurable since otherwise there's no way to interpret the plot.
- What's the best syntax for inequalities instead of the contours pictured above. Currently (and to the extent that I haven't made errors), carpet contours reuse a lot of the regular contour code so that the specification is nearly identical. I think it would be nice to add a `inequality: true | false` property to the `contours` section of the trace definition. The effect would be to tweak the specification of levels so that it plots reasonable-looking inequalities by default (vs a sequence of contours). This is nothing that can't be achieved by specifying, say:

```javascript
contours: {
  start: 1.5,
  end: 1000000,
  step: 10000000
}
```

in order to ensure it's one-sided, but the downside is both in specifying flipped inequalities and in the clumsiness of having to make up arbitrary large numbers to avoid multiple levels.

alexcjohnson:
> Should the carpet axis grid fall on top of the fill? this seems highly desirable (or at the very least should perhaps be configurable since otherwise there's no way to interpret the plot.

Good call, on top only. I don't see any examples of carpet plots that *aren't* built this way, and if you don't want the grid, presumably you can just turn it off (set its width to zero or something)? No need to make a setting for it until someone explicitly requires it - anyway the only case I can see that this would be different from turning the grid off is if your contour mesh doesn't cover the full a/b domain.

> What's the best syntax for inequalities instead of the contours pictured above.

Lets see if we can share syntax with [filters](https://github.com/plotly/plotly.js/blob/master/src/transforms/filter.js) - I can imagine, for example, wanting to accept a band, like `operation: '[]'`. Maybe even reject a band, it's harder to think of uses for that but why not. So how about we use the same, `operation` and `value`?

In a bunch of other cases we let a setting default to `false` rather than inventing a value to keep the type consistent, so I'd propose doing that with `operation` as well. Makes the internal logic a tiny bit simpler too - `if(contours.operation)`

For this purpose there's no difference between `<` and `<=` etc, so we wouldn't need all the `operation` values present in filters, though it might actually be nice to support them all: what if a user adds scatter data to the carpet, using one of these filters, and would like to match it to a constraint using the same filter? And even `'{}'` could be neat - you'd use that to provide an explicit list of contour values to show, rather than being limited to linear spacing.

Last question: how will the colors be set in these cases? The way we describe colorscales right now is an array of length-2 arrays, each `[fraction, color]` where fraction must rise monotonically from 0 to 1. I suppose we can use `fraction=0` to denote "false" (excluded) and `fraction=1` to denote "true" (allowed), is that what you had in mind?

rreusser:
I've been thinking about how to accomplish this. I'm thinking maybe just pass an optional mapping function to the contour function. The really really difficult way to accomplish bands seems to be to actually implement them in the contouring logic.

The much simpler way would seem to be to pass a mapping function. A band from 0 to 1 would be `z * (1 - z) > 0`, which is perfectly straightforward without any other modifications. I think each operation would then just be a carefully constructed mapping and everything else proceeds as normal.

Thinking about the last question now.

alexcjohnson:
Arbitrary functions are nice for flexibility - but I worry about them for security but also because we already have an enumerated syntax for filters.

rreusser:
Sorry, yes. Not arbitrary functions. I just mean internally mapping filter operations to a particular function.

rreusser:
Okay, I've partially implemented constraints. The interesting ones are inequalities since it requires merging and flipping the extracted contour paths.

This code:

```javascript
data: [{
  type: 'carpet'
  a: [0, 1, 2, 3],
  b: [4, 5, 6],
  y: [
    [1, 1, 1, 1],
    [2, 2, 2, 2],
    [3, 3, 3, 3]
  ]
}, {
  type: 'contourcarpet',
  z: [
    [1, 1, 1, 1],
    [2, 2, 2, 2],
    [3, 3, 3, 3]
  ],
  autocontour: false,
  contours: {
    constraint: {
      operation: '][',
      value: [1.5, 2.5]
    }
  },
}]
```

Shades the region *outside* the specified range (following filter transform syntax). The result looks like this:

screen shot 2017-02-02 at 17 09 59

Flipping that backwards, the operation `[]` (open intervals = identical) looks like this:

screen shot 2017-02-02 at 17 10 14

Feedback on syntax welcome!

rreusser:
Specific points about these contours:

1. autocontour will either ignore/not ignore the whole `contours` section, as it does now.
2. if a constraint operation is present, it will overwrite the `size` `start` and `end` properties with whatever it needs to get the contours to work correctly.
3. Labeling the inequalities may be required. Wondering if there's ever been talk of making + labeling contours along the contour itself or whether should push labels to the boundary.

alexcjohnson:
> The interesting ones are inequality ranges since it requires merging and flipping the extracted contour paths.

Would it work to just make it as two regular contours and fiddle with the colors instead (to give out/in/out or in/out/in) rather than stitching together the contours into one? Just a thought, if you're already down that road it may be too late to matter.

Re:
```
  contours: {
    constraint: {
      operation: '[]',
      value: [1.5, 2.5]
    }
  }
```

Hmm, I see the rationale for this, as just `contours: {operation: '[]', value: [1.5, 2.5]}` is a little strange... but it feels too nested to me. @etpinard thoughts? `contours: {constraint: '[]', value: [1.5, 2.5]}` perhaps? It breaks the correspondence with filters, but only a little bit.

> `autocontour` will either ignore/not ignore the whole contours section, as it does now.

Sounds fine, as long as specifying a constraint and value is enough to make `autocontour` default to `false`.

> if a constraint operation is present, it will overwrite the `size` `start` and `end` properties with whatever it needs to get the contours to work correctly.

👍 

> Labeling the inequalities may be required. Wondering if there's ever been talk of making + labeling contours along the contour itself or whether should push labels to the boundary.

There's been "that would be nice" discussion but no action. Lets omit it for the initial PR and put it in the hopper to be implemented for both standard and carpet contours, along with better gradient following for the contours.

rreusser:
> Would it work to just make it as two regular contours and fiddle with the colors instead

(Please correct me if I'm wrong about any of this, but to the best of my understanding:)

I don't think regular contours will work. If you want to fill the area from `0 < z < 1`, regular contours will fill `> 0` and `>1` and one must be drawn with paperbg so that it _looks_ like it's not there. Since these fills definitely need to be semi-transparent and overlap, that seems like it will cause layering problems.

> Hmm, I see the rationale for this, as just contours: {operation: '[]', value: [1.5, 2.5]} is a little strange... but it feels too nested to me. @etpinard thoughts? contours: {constraint: '[]', value: [1.5, 2.5]} perhaps? It breaks the correspondence with filters, but only a little bit.

Agreed on the too nested. I can just hoist it up a level. `value` could be an enumerated list of values; I just haven't done that case yet.

> Sounds fine, as long as specifying a constraint and value is enough to make autocontour default to false.

That's the way it currently works, which didn't really match what I intended or how I thought I'd implemented it, but… 🎉 🎉 🎉 hooray for dumb luck!

alexcjohnson:
> I don't think regular contours will work.

Ah, you're right. I thought perhaps we had created the contour fills non-overlapping but that actually wouldn't even work, it would make unpleasant effects at the boundaries due to antialiasing. Next I started thinking you could avoid stitching by transforming z, ie finding zero-crossings of (z-z1) * (z-z2) but even that, I think there would be problems if there's any single cell that goes all the way from <z1 to >z2. So OK, I guess you pretty much have to flip and stitch.

> I think the MVP for release is labeling the inequalities somewhere, otherwise it's basically impossible to tell which constraint is which.

I see lots of examples of unlabeled constraints online (and we have the advantage of hover labels while the plot is in the browser), but point taken. Wherever we put these labels it's going to be tough to keep them from colliding - if we label them on the carpet, they'll likely collide with each other and the other stuff being shown. If we label them off the edge, they still might collide with each other, and they'll very likely collide with the a/b labels. Maybe we can try to put them off a side that has no a/b labels, and barring that try to scoot them out until they don't hit anything? Dunno, I'm not sure there will be a solution that works all the time, eventually people will just have to add their own annotations (in x/y space of course!) to get it exactly the way they want it.

etpinard:
Perhaps we could keep the same `contours` container as in regular contour traces and have a `constraints` container in parallel. For example:

```js
{
  type: 'contourcarpet',
  z: [ [ /* */ ] ],
  contours: {
    start: 0,
    end: 10,
    size: 1,
    coloring: 'fill',
    showlines: true
  },
  constraints: {
    operation: '[]',
    values: [1.5, 2.5],
    coloring: 'fill',
    showlines: true
  }
}
```

that way we don't have add any _if-start-end-size-are-present-or-if-operation-values-are-present_ logic (which can be confusing to users especially on `restyle`) in the default step. In addition, this would imply that one `contourcarpet` trace could have both visible contours and constraints - maybe that's something that people want or maybe not :smile:

rreusser:
@etpinard yeah, I considered that. The downside is the extra complexity of managing two sets of contours and contour properties etc, but it's certainly not impossible.

rreusser:
@alexcjohnson:

> Last question: how will the colors be set in these cases? The way we describe colorscales right now is an array of length-2 arrays, each [fraction, color] where fraction must rise monotonically from 0 to 1. I suppose we can use fraction=0 to denote "false" (excluded) and fraction=1 to denote "true" (allowed), is that what you had in mind?

I like that. Since it's binary, I'd vote for allowing nulls and then use that to decide whether or not to display the fill/background.

@etpinard:

> In addition, this would imply that one contourcarpet trace could have both visible contours and constraints - maybe that's something that people want or maybe not 😄

As I reorganize, I'm starting to think that's the right option. It's not strictly necessary, but it's not too difficult to add layers (side note: I'm adding junk to the bar layer at the moment. By the time you add an axis and stuff above/below the axis, we might need another layer here.

rreusser:
@etpinard @alexcjohnson The API is feeling better and better, but I need to commit to a decision regarding constraints vs. contours. Part of me likes the concept of drawing both contours and constraints for a single trace, but part of me thinks that goes against the concept of 'traces', which favors simplicity by simply requiring separate traces if you want to represent the same data in different ways. Plus, contours + 1 inequality would be one or two traces (no difference, just a choice), and contours + 2 inequalities would be two or three traces (again, no difference). This suggests something is a little goofy with the logical breakdown.

A list of inequalities could solve it. Multiple constraints feels like a corner case, but to be honest, so does `contours` + `constraint` (singular) except with an API that doesn't feel quite right. The main advantage to multiple constraints would be DRYing the json and that the legend would toggle all inequalities on the trace together.

```javascript
{
  type: 'contourcarpet',
  z: [ [ /* */ ] ],
  contours: {
    start: 0,
    end: 10,
    size: 1,
    coloring: 'fill',
    showlines: true
  },
  constraints: [{
    operation: '[]',
    value: [1.5, 2.5],
    coloring: 'fill',
    showlines: true
  }, {
    operation: '[]',
    value: [3.5, 4.5],
    coloring: 'fill',
    showlines: true
  }
}
```

My favorite approaches:

- Just iron out the contour vs. constraint settings so that it's one or the other
- A list of constraints that are drawn in addition to the optional fill

etpinard:
> Plus, contours + 1 inequality would be one or two traces (no difference, just a choice), and contours + 2 inequalities would be two or three traces (again, no difference). This suggests something is a little goofy with the logical breakdown.

@rreusser you're making a good point here.

How about this:

```js
{
  type: 'contourcarpet',
  z: [ [ /* */ ] ],
  contours: {
    type: 'levels' || 'constraint',

    // coerced if type === 'levels'
    start: 0,
    end: 10,
    size: 1,

    // coerce if type === 'constraint'
    operation: '[]',
    value: [3.5, 4.5],

    // coerced regardless of type
    coloring: 'fill',
    showlines: true
  }
}
```

alexcjohnson:
At the risk of being too opinionated with other peoples' data, I don't think I like the idea of contours + constraints, or multiple constraints, at all, so I think we should try for a mutually exclusive API and if you want both you make multiple traces. In either of those cases, showing all of that together will be too hard to interpret visually; you should just show contours, and perhaps make a custom colorscale that highlights certain boundaries - not sure if we have this documented anywhere, but if you do something like `colorscale: [[0, 'black'], [0.5, 'red'], [0.5, 'blue'], [1, 'white']]` you get a sharp transition from red to blue at the middle of the scale

screen shot 2017-02-06 at 3 02 39 pm

rreusser:
@etpinard @alexcjohnson thanks for the feedback! Agreed on both points. I think we're all on the same page then, and seems the right answer is `type: 'levels' || 'constraint'`, which I think is straightforward.

alexcjohnson:
> seems the right answer is `type: 'levels' || 'constraint'`, which I think is straightforward.

Nice, I like it (sorry @etpinard - I didn't see that before posting my comment)

rreusser:
Next major item to be cemented is the layering. I don't think the existing layers are adequate. I was putting some things in `barlayer` because I needed a couple layers across axis/contour/scatter carpet traces in order to get things to look right. Decision I need to make:

1. Using the existing layers and do whatever is necessary to get subgroups to order/layer correctly
2. It's fine to add maybe `maplayerabove` or something similar in order to make space without having to focus on order

They'll both work. It's not a hard problem to solve. Any guidance on which is preferred would be greatly appreciated!

etpinard:
I'd vote for 

> It's fine to add maybe maplayerabove or something similar in order to make space without having to focus on order

for now ... and maybe try to merge layer content at a later time (maybe once [this](https://github.com/plotly/plotly.js/blob/87b12cfdec747bee88fcaf76e868a37a4c0917c8/src/plots/cartesian/index.js#L97-L105) gets :hocho: )

rreusser:
One more small step forward with a/b axis labels (sparing some sanity on crazy long prefixes/suffixes on the ticks themselves :tada:)

screen shot 2017-02-08 at 17 59 24

rreusser:
Okay, major issue to be solved: Looks like constraints that fill the whole area (e.g. `z: 10, constraint: < 1`) have an issue:



screen shot 2017-02-09 at 12 42 27

rreusser:
Right now you have to add:

```javascript
 colorscale: [[0, 'blue'], [1, 'blue']]
```

in order to set the fill color. Do you think it deserves a shortcut? For convenience, I think that would be good, but on principle, I think it's unpleasant to have shortcut aliases for other properties that already exist. At best, I'd think maybe `countours.constraintcolor` or some other constraint-specific property could maybe override the colorscale if provided.

src/traces/heatmap/convert_column_xyz.js
```diff
@@ -11,47 +11,66 @@
 
 var Lib = require('../../lib');
 
+module.exports = function convertColumnData(trace, ax1, ax2, var1Name, var2Name, arrayVarNames) {
```
etpinard: @rreusser would you mind refactoring this routine here so that it doesn't mutate `trace` (which is the `fullData` item)?

See https://github.com/plotly/plotly.js/pull/1417#discussion_r103071813

rreusser: I think that should be straightforward. I've already implemented calcPriority to reorder the calc step. Calling this is one of the last steps in carpet defaults, so it should be possible to simply move this to the beginning of `calc` instead. It might cascade into `scattercarpet` and `contourcarpet`, but I think the only downside is the inability to bail out early due to invalid data.

src/traces/carpet/calc_gridlines.js
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    var xcp = trace._xctrl;
+    var ycp = trace._yctrl;
+    var nea = xcp[0].length;
+    var neb = xcp.length;
+    var na = trace.a.length;
+    var nb = trace.b.length;
+
+    Axes.calcTicks(axis);
```
etpinard: Headsup @rreusser ! This statement leads to an exception post https://github.com/plotly/plotly.js/pull/1431

I fixed (in a hacky way) with:

image

alexcjohnson: somehow we'd better get `layout.separators` in there instead of hardcoding the defaults...

rreusser:
Cementing API details. `minorgridcount` currently specifies the number of minor grid lines *between* major grid lines. The plot below has `minorgridcount: 1`: 

screen shot 2017-03-16 at 16 23 10

So if you have major grid lines at 1, 2, 3, ..., then to get minor grid lines at each 0.1, you'd use `minorgridcount: 9`. That seems weird. Do you think I should subtract one instead and the count would *include* the major grid lines?

rreusser:
Good thought on standardizing sooner than later. That's a lot to process, for reference, the full current carpet axis logic is:

- if `tickmode: 'array'`:
  - place gridlines at every array index `arraytick0 + n * arraydtick` for `n = ..., -2, -1, 0, 1, 2, ...`
  - place `minorgridcount` minor gridlines between the arrays
- else:
  - place gridlines at every a/b value `tick0 + n * dtick` for `n = ..., -2, -1, 0, 1, 2, ...`
  - place `minorgridcount` minor gridlines between the arrays

rreusser:
Okay, so the changes/behavior for carpet would then be:

#### Major gridlines

- Add `tickmode: 'auto'`, which will place gridlines at nicely chosen values. If you specify `nticks`, it will choose the largest 1/2/5 increment that creates `<= nticks` ticks. If you specify `dtick` (and optionally `tick0`, then it will ignore `nticks` and use the provided spacing.
- Keep `tickmode: 'array'` and by default it will the values in `a` or `b`. (this is peculiar to carpet axes since the data is provided, and it's particularly easy because the curve points are embedded in the spline stuff so don't require any computation)
- If `tickmode: 'array'` *and* `tickvals` are provided though, then it switches over to the interpolation strategy and places ticks explicitly at the `tickvals` specified
- Carpet also has `arraydtick` and `arraytick0` which are peculiar to carpet axes and are useful in particular for the case where `tickvals` default to `a` and `b`.

#### Minor gridlines

Specify `nticks`. You may or may not get the number you require, but it will try to place them at nice 1/2/5 values. If `tickmode: 'array'`, then presumably it's all weird numbers anyway so you'll just get the number you request even if they're at weird values.

etpinard:
:+1: for adding `nticksminor`  to specify the number of minor tick intervals and a more precise `dtickminor`.

We should probably add a `tickminormode` with values `auto` (i.e. use `nticks`) and `linear` to let users easily toggle from the `nticksminor` and `dtickminor` settings. `tickminormode: 'array'` can wait.

I'm not sure how exactly it would work, but we should allow special strings inputs for `dtickminor` for log and date axes (similar to `dtick` of course).

We should probably at some point add a `ticksminor` attribute analogous to `ticks: '' | 'outside' | 'inside'` to allow users to have inside (major) ticks and outside minor ticks and vise-versa.

rreusser:
How does `dtickminor` work? Say you have gridlines at 0.152 and 0.328. `dtickminor` is `0.1`. Where do the minor gridlines end up?

alexcjohnson:
All of these are strange "why the heck would you do this" cases, but here goes:

I would think if you have a weird `dtickminor` (ie that's not an even fraction of the major tick spacing) you'd still just increment by that value from the first tick until you get to or past the last tick. So between 0.152 and 0.328 with `dtickminor: 0.1` you'd just get a single tick at 0.252. Strange request from the user, but that seems like the most faithful interpretation of it.

And if you have `ntickminor: 9` but weird major tick spacing, I would just make exactly 9 evenly-spaced ticks. Even spacing seems far "nicer" than trying to make "nice" spacing and ending up with a weird remainder.

> I like the 10 - 1 = 9 gridlines between the major gridlines style better

Works for me. Igor has made an awful lot of good decisions over the years, you can't go too wrong by following their lead.

rreusser:
Sounds good. Yeah, I agree. They're useless corner cases, but I was thinking of making impossible states impossible, and if they're not impossible, then we need to be set up to handle them reasonably. 😄

rreusser:
A small update. This is what I currently have for carpet legend + labeling the constraints. I'll see if I can get some sensible defaults. For contour plots (not constraints, but that's just a configuration difference), it will show the colorscale by default instead:

screen shot 2017-03-21 at 17 35 07

rreusser:
Oh wow. Didn't explicitly fix this case, but circled back to the old mock and it actually looks pretty good. @bpostlethwaite do you think this works for identifying the constraints in lieu of on-plot labels?

screen shot 2017-03-22 at 13 48 56

bpostlethwaite:
Looks great - yep I think this will close that particular issue/feature

rreusser:
@etpinard It looks like my tests are failing a couple gl3d hover tests:

```
  Test gl3d plots
    - @noCI should display correct hover labels and emit correct event data
    - @noCI should emit correct event data on click
..................................
.............................................................................
Chrome 54.0.2840 (Linux 0.0.0) ERROR
  Uncaught TypeError: Cannot read property 'textContent' of null
  at /node_modules/d3/d3.js:788:0 <- /tmp/b583f447d72339f4472133e4317f6f6d.browserify:13654
Chrome 54.0.2840 (Linux 0.0.0) ERROR
  Uncaught TypeError: Cannot read property 'textContent' of null
  at /node_modules/d3/d3.js:788:0 <- /tmp/b583f447d72339f4472133e4317f6f6d.browserify:13654
Chrome 54.0.2840 (Linux 0.0.0): Executed 797 of 1804 (skipped 2) ERROR (2 mins 34.697 secs / 2 mins 32.006 secs)

Chrome 54.0.2840 (Linux 0.0.0): Executed 797 of 1804 (skipped 2) ERROR (2 mins 34.697 secs / 2 mins 32.006 secs)
```

Before I start digging, do you have a gut reaction whether this is known/expected or whether I might have broken it? I think gl3d tests have been a thing getting some attention lately.

rreusser:
Small update: Zoom and pan currently fail with scatter/contourcarpet. Trying to make the references right and finding myself pretty much at an impasse. When it redraws *without* recalc, the axis references are not correct. Something [here](https://github.com/plotly/plotly.js/pull/1239/files#diff-55e73e1ce7c046db1f44dbeb16ce40caR36) is wrong with the way I'm handling the fake a/b axes objects such that when I try to use them later on, the setConvert functions return NaN. For some reason I still can't quite wrap my head around what happens to all of the object identities on redraw.

src/traces/carpet/defaults.js
```diff
@@ -45,6 +43,8 @@ module.exports = function supplyDefaults(traceIn, traceOut, dfltColor, fullLayou
 
     var len = handleXYDefaults(traceIn, traceOut, coerce);
 
+    setConvert(traceOut);
```
etpinard: :tada:

rreusser:
Outstanding issues on this ticket:

- [ ] tick attributes need a couple name/behavior changes
- [ ] need to give the schema a close look and make sure descriptions are valid
- [ ] need a winding number fix for constraints that fill the whole region
- [ ] need to ensure unused attrs are unset for the toolpanel
- [ ] doesn’t currently obey trace.opacity (thanks, ben) (edited)
- [ ] should finally just make a proper layer for carpet plots that’s *not* barlayer

rreusser:
Re: winding number issues, I made some sorta extensive mocks and nailed down the problems. Hooray! 🎉 Adding the corresponding mocks for ranges. Lots of cases here.

cheater_constraint_less_than_with_hill

rreusser:
So when you combine ranges and winding numbers, there actually turn out to be quite a number of cases that need to be checked and verified. 🎉 

montage

rreusser:
Quick q @etpinard @alexcjohnson: What's more plotly-like for carpet axes? Given `y` alone but not `a` or `b`, what's the desired behavior? Should it fill in `a` and `b` with `[0, 1, 2, ...]` or should it set `visible: false`?

etpinard:
> What's more plotly-like for carpet axes? Given y alone but not a or b, what's the desired behavior? Should it fill in a and b with [0, 1, 2, ...] or should it set visible: false?

It should use `a0` and `da` to fill in a linear space. Or am I confused these `a` / `b` fields with something else?

rreusser:
Nope. You're right. Just made that connection. It just needs to piggyback on heatmap xyz defaults. Confirming/checking now. 👍

rreusser:
@alexcjohnson @etpinard trivial decision that'll be hard to change: should the contourcarpet fill regions be above or below the carpet axis, z-order-wise. I vote below because the carpet axis lines are semi-transparent by default and since the axis lines aren't regular so that the alignment isn't implicit/obvious as it is for contour/heatmap plots.

alexcjohnson:
> I vote below because I've made the carpet axis lines semi-transparent by default and since the axis lines aren't regular so that the alignment with tick labels isn't implicit/obvious as it is for contour/heatmap plots.

I agree. That's how it's handled in all the filled contour carpet examples I see online too.

rreusser:
Caught a lot of bugs writing this one…

screen shot 2017-04-05 at 18 24 12

src/plots/gl3d/layout/axis_attributes.js
```diff
@@ -14,6 +14,15 @@ var extendFlat = require('../../../lib/extend').extendFlat;
 
 
 module.exports = {
+    visible: {
+        valType: 'boolean',
+        role: 'info',
+        description: [
```
etpinard: Could reuse the declaration from `cartesian/` with `visible: axesAttrs.visible,`

src/plots/plots.js
```diff
@@ -1926,7 +1962,34 @@ plots.doCalcdata = function(gd, traces) {
...
+        _module = trace._module;
+
+        // Get the priority from the module definition, otherwise default to zero:
+        priority = (_module && _module.calcPriority) ? _module.calcPriority : 0;
+
+        // Note if a non-zero priority was found so that we only need to sort if
+        // there's anything to sort:
+        hasPriorities = hasPriorities || priority;
+
+        prioritizedFullDataLookup.push({index: i, priority: priority});
```
etpinard: Oh, we need this after all? I thought we concluded it wasn't necessary at some point in #plotly_js. Honesty, it doesn't add much code, so I'm :+1: either way.

src/traces/carpet/attributes.js
```diff
@@ -0,0 +1,121 @@
...
+
+'use strict';
+
+var extendFlat = require('../../lib/extend').extendFlat;
+var fontAttrs = require('../../plots/font_attributes');
+var axisAttrs = require('./axis_attributes');
+var colorAttrs = require('../../components/color/attributes');
+
+module.exports = {
+    carpetid: {
```
etpinard: `id` isn't the most plotly-esque of identifiers. It's probably the best for this job though. But perhaps, `carpetname` or simply `carpet` would sound more familiar to plotly users?

src/traces/carpet/axis_attributes.js
```diff
@@ -0,0 +1,445 @@
...
+            'all at once: line, font, tick, and grid colors.',
+            'Grid color is lightened by blending this with the plot background',
+            'Individual pieces can override this.'
+        ].join(' ')
+    },
+    smoothing: {
+        valType: 'number',
+        dflt: 1,
+        min: 0,
+        max: 1,
```
etpinard: it would be nice to make this consistent with the `scatter` definition of [`smoothing`](https://github.com/plotly/plotly.js/blob/6e547ef794093c206998103f3065d5f2ab2a20b4/src/traces/scatter/attributes.js#L154) where the default value is `1` and the max `1.3`.

src/traces/carpet/axis_attributes.js
```diff
@@ -0,0 +1,445 @@
...
+        dflt: 'value',
+        role: 'info'
+    },
+    tickmode: {
+        valType: 'enumerated',
+        values: ['linear', 'array'],
+        dflt: 'array',
+        role: 'info',
+    },
+    nticks: {
```
etpinard: it would be nice to reuse the `cartesian/` declarations here, when applicable.

src/traces/carpet/axis_defaults.js
```diff
@@ -0,0 +1,238 @@
...
+        family: font.family,
+        size: Math.round(font.size * 1.2),
+        color: dfltFontColor
+    });
+
+    coerce('titleoffset');
+
+    coerce('tickangle');
+
+    var validRange = (
```
etpinard: you can now use [`ax.isValidRange`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/set_convert.js#L406). :recycle:

src/traces/carpet/axis_defaults.js
```diff
@@ -0,0 +1,238 @@
...
+        font = options.font || {},
+        attributes = carpetAttrs[letter + 'axis'];
+
+    options.noHover = true;
+
+    function coerce(attr, dflt) {
+        return Lib.coerce(containerIn, containerOut, attributes, attr, dflt);
+    }
+
+    // I don't know what this does:
```
etpinard: :hocho: 

See https://github.com/plotly/plotly.js/blob/master/src/lib/coerce.js#L316

src/traces/carpet/axis_defaults.js
```diff
@@ -0,0 +1,238 @@
...
+
+    if(!containerOut.title || (containerOut.title && containerOut.title.length === 0)) {
+        delete containerOut.titlefont;
+        delete containerOut.titleoffset;
+    }
+
+    return containerOut;
+};
+
+function setAutoType(ax, data) {
```
etpinard: Hmm. @alexcjohnson moved the auto-type stuff around in https://github.com/plotly/plotly.js/pull/1522/ 

Does this conflict?

lib/index.js
```diff
@@ -36,6 +36,10 @@ Plotly.register([
 
     require('./scattermapbox'),
 
+    require('./carpet'),
+    require('./scattercarpet'),
+    require('./contourcarpet'),
```
etpinard: Maybe carpet traces should be part of the [`cartesian`](https://github.com/plotly/plotly.js/blob/master/lib/index-cartesian.js) bundle?

If `scatterternary` wasn't already in there I'd vote for _no_. But now that I realise `scatteternary` is part of the cartesian bundle, I'm not so sure.

src/traces/carpet/ensure_array.js
```diff
@@ -0,0 +1,27 @@
...
+
+/*
+ * Ensures an array has the right amount of storage space. If it doesn't
+ * exist, it creates an array. If it does exist, it returns it if too
+ * short or truncates it in-place.
+ *
+ * The goal is to just reuse memory to avoid a bit of excessive garbage
+ * collection.
+ */
+module.exports = function ensureArray(out, n) {
```
etpinard: maybe general / reusable enough for `src/lib/`?

src/traces/carpet/index.js
```diff
@@ -0,0 +1,31 @@
...
+Carpet.animatable = true;
+Carpet.calcPriority = 1;
+Carpet.defaultPriority = 1;
+
+Carpet.moduleType = 'trace';
+Carpet.name = 'carpet';
+Carpet.basePlotModule = require('../../plots/cartesian');
+Carpet.categories = ['cartesian', 'carpet', 'carpetAxis', 'notLegendIsolatable'];
+Carpet.meta = {
+    description: [
```
etpinard: TODO :books:

src/traces/carpet/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+var Carpet = {};
+
+Carpet.attributes = require('./attributes');
+Carpet.supplyDefaults = require('./defaults');
+Carpet.plot = require('./plot');
+Carpet.calc = require('./calc');
+Carpet.animatable = true;
+Carpet.calcPriority = 1;
+Carpet.defaultPriority = 1;
```
etpinard: Very clear, very nice :+1:

src/traces/carpet/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+Carpet.moduleType = 'trace';
+Carpet.name = 'carpet';
+Carpet.basePlotModule = require('../../plots/cartesian');
+Carpet.categories = ['cartesian', 'carpet', 'carpetAxis', 'notLegendIsolatable'];
```
etpinard: Nicely done here.

src/traces/carpet/measure_text.js
```diff
@@ -0,0 +1,22 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+var Drawing = require('../../components/drawing');
+
+module.exports = function measureText(tester, text, font) {
```
etpinard: Probably best for `component/drawing/`

src/traces/carpet/plot.js
```diff
@@ -0,0 +1,229 @@
...
+
+        var el = d3.select(this);
+
+        el.text(axis.title || '')
+            .attr('transform',
+                'translate(' + orientation.p[0] + ',' + orientation.p[1] + ') ' +
+                'rotate(' + orientation.angle + ') ' +
+                'translate(0,' + offset + ')'
+            )
+            .classed('user-select-none', true)
```
etpinard: nice touch.

src/traces/contour/colorbar.js
```diff
@@ -30,11 +35,6 @@ module.exports = function colorbar(gd, cd) {
     var cb = drawColorbar(gd, cbId);
     cd[0].t.cb = cb;
 
-    var contours = trace.contours,
```
etpinard: why do we need to bring this block up?

src/traces/contour/make_color_map.js
```diff
@@ -21,6 +21,11 @@ module.exports = function makeColorMap(trace) {
         nc = Math.floor((end - start) / cs) + 1,
         extra = contours.coloring === 'lines' ? 0 : 1;
 
+    if(!isFinite(cs)) {
```
etpinard: why? 

P.S. I apologize for being lazy and not reviewing this PR commit by commit.

src/traces/contour/plot.js
```diff
@@ -277,7 +277,8 @@ function makeLines(plotgroup, pathinfo, contours) {
         .attr('d', function(d) {
             return Drawing.smoothopen(d, smoothing);
         })
-        .style('stroke-miterlimit', 1);
+        .style('stroke-miterlimit', 1)
+        .style('vector-effect', 'non-scaling-stroke');
```
etpinard: learn something new everyday.

https://www.w3.org/TR/SVGTiny12/painting.html#NonScalingStroke

src/traces/contourcarpet/plot.js
```diff
@@ -0,0 +1,305 @@
...
+            d3.select(this).remove();
+        } else {
+            d3.select(this)
+                .attr('d', fullpath)
+                .style('stroke', 'none');
+        }
+    });
+}
+
+/* function clipGaps(plotGroup, plotinfo, cd0, perimeter) {
```
etpinard: TODO or :hocho:

src/traces/contourcarpet/style.js
```diff
@@ -0,0 +1,64 @@
...
+        if(!isFinite(cs)) {
+            cs = 0;
+        }
+
+        c.selectAll('g.contourlevel').each(function() {
+            d3.select(this).selectAll('path')
+                .call(Drawing.lineGroupStyle,
+                    line.width,
+                    line.color,
+                    // contours.coloring === 'lines' ? colorMap(start + i * cs) : line.color,
```
etpinard: todo or :hocho:

src/traces/scattercarpet/index.js
```diff
@@ -0,0 +1,31 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var ScatterCarpet = {};
+
+ScatterCarpet.attributes = require('./attributes');
+ScatterCarpet.supplyDefaults = require('./defaults');
+// ScatterCarpet.colorbar = require('../scatter/colorbar');
```
etpinard: I guess should _just work_ as `scattercarpet` is using the same colorscale attribute as the cartesian version.

src/traces/scattercarpet/select.js
```diff
@@ -0,0 +1,33 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+var scatterSelect = require('../scatter/select');
+
+
+module.exports = function selectPoints(searchInfo, polygon) {
```
etpinard: Hey, why not? Does this work?

src/traces/scattercarpet/style.js
```diff
@@ -0,0 +1,27 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+var scatterStyle = require('../scatter/style');
+
+
+module.exports = function style(gd) {
```
etpinard: This module isn't registered in `scattercarpet/index.js`. TODO or :hocho: ?

lib/index.js
rreusser: Ah, yes. I'm not sure exactly what's best for a default. Easiest for working was just to bundle it. But it's quite a bit of code, generally speaking, so that I feel like it will require some more nuanced per-bundle decisions.

src/lib/carpet.js
```diff
@@ -0,0 +1,11 @@
...
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = require('../src/traces/carpet');
```
rreusser: Maybe it shouldn't be included by default, but see: https://github.com/plotly/plotly.js/blob/3f18627be8cb1be48a9c3450329e0cbed713ffd6/lib/index.js#L41

src/plots/plots.js
rreusser: That's a good question… I thought the same, but there it is. I'll check on the removability.

src/traces/carpet/attributes.js
rreusser: `carpet` works for me. Should an easy change.

src/plots/gl3d/layout/axis_attributes.js
rreusser: Good point. I blindly copied this definition in a rush since it broke gl3d plots entirely until it was present.

src/traces/carpet/attributes.js
rreusser: ✅

src/traces/carpet/axis_defaults.js
rreusser: ✅

src/traces/carpet/ensure_array.js
rreusser: Yes, definitely. Part of the goal here was to reuse existing arrays to the extent possible so that we don't thrash garbage collection *quite* so much.

src/traces/carpet/index.js
rreusser: Where does this end up and what is it intended to communicate?

src/traces/carpet/measure_text.js
rreusser: ✅

src/traces/carpet/plot.js
rreusser: Ah, more things should perhaps have that. The double-click behavior (e.g. legend isolate) is kind of annoying when it selects a bunch of axis labels in the process.

src/traces/contour/colorbar.js
rreusser: Perhaps no good reason. I think this was path-dependent where I first had some extra conditions that required it, then removed the extra conditions. I'll work through this and attempt to keep the original flow intact.

src/traces/contour/make_color_map.js
rreusser: Hoo boy. Okay, so. I've tried to piggyback to the full extent possible on existing contour behavior. Part of that was to use, for example, `start = 5`, `end = Infinity`, `size = Infinity` for the inequality `z > 5`. That causes it to automatically extract the correct contour levels without additional logic. So in that sense, the information is 'correct'. The only caveat is that it required a couple additional checks in places like this.

src/lib/carpet.js
etpinard: Right `lib/carpet.js` is required, but `src/lib/carpet.js` isn't.

src/traces/contourcarpet/plot.js
rreusser: Yup. TODO. I'll remove it. It's in the git history or just copied from contour anyway.

src/lib/carpet.js
rreusser: ✅

src/plots/gl3d/layout/axis_attributes.js
rreusser: ✅

src/plots/plots.js
rreusser: Reverting this. I reordered one of the mocks to ensure that calcdata is in an order that would cause it to fail if this were required but not implemented.

rreusser: ✅

src/traces/carpet/ensure_array.js
rreusser: ✅

src/traces/carpet/index.js
rreusser: Unfortunately also both removed 😄

src/traces/carpet/index.js
rreusser: The names of the categories might not be the _best_, but they're internal at least so can perhaps be changed.

src/traces/carpet/plot.js
rreusser: ⏫ I have no idea what I was referring to in the above comment.

src/traces/contour/colorbar.js
rreusser: See #760

src/traces/contour/plot.js
rreusser: See #760

src/traces/carpet/index.js
etpinard: Oh, I thought that ended up on https://plot.ly/javascript/reference, but I guess not currently. 

It intends to communicate some general info about the trace object especially on how different attribute can interact with one another, the [`heatmap`](https://github.com/plotly/plotly.js/blob/00355fb142ea41a36235676b5180ac9e23c5c9a8/src/traces/heatmap/index.js#L26) meta description, is probably the best we got so far.

src/traces/scattercarpet/index.js
rreusser: Correct. Uncommented. ✅

src/traces/contourcarpet/plot.js
rreusser: ✅

lib/index.js
etpinard: OK. Let's leave carpet traces out of the cartesian bundle to start. We can always add then down the road in a future minor release if choose to.

src/traces/carpet/index.js
etpinard: no need to worry about those category names :+1:

src/traces/contour/make_color_map.js
etpinard: Nice touch. Thanks for :recycle: as much as possible.

src/traces/contour/colorbar.js
rreusser: ✅

src/traces/scattercarpet/select.js
rreusser: Appears so! 🎉

src/traces/scattercarpet/style.js
rreusser: I've included it and restored the same check that scatterternary uses to either run it or take for granted that the scatter trace has already run it.

src/traces/carpet/axis_defaults.js
rreusser: I think *maybe* there was a conflict, but I merge with master and sorted out a couple trivial merge conflicts and verified that axis constraints work.

lib/index.js
rreusser: ✅

rreusser: Is there a bundle which should include it by default?

rreusser:
Ah, I see. My confusion. 👍

src/traces/scattercarpet/style.js
rreusser: ✅

src/traces/scattercarpet/select.js
rreusser: ✅

src/traces/carpet/index.js
rreusser: Added meta. ✅

rreusser:
Superseded by #1595.

PR #1246, opened 2016-12-11
don't auto-run `git push && git push --tags` on version

etpinard:
as discussed privately with @alexcjohnson , the auto `git push && git push --tags` postversion hook is probably too magical. Adding one step to our publishing tasks will allow us to inspect `dist/` bundles and `dist/npm-ls.json` before pushing to the origin.

For example,  the failed `v1.19.0` could have been avoided if this hook wasn't present.

TODO:

- [ ]  update dev-docs once merged


alexcjohnson:
Nice note! 💃

PR #1252, opened 2016-12-14
Calendar fixes

alexcjohnson:
@etpinard can we get this out as a patch release and in streambed ASAP? Fixes the two world calendar bugs we found:
- chinese calendar `ms2DateTime` misinterpreted all month 1's as intercalary: @n-riesco fixed this in https://github.com/alexcjohnson/world-calendars/pull/9 and here I bumped `world-calendars` to 1.0.2
- `tickformat` was broken with world calendars - this fixes it, and tests `Lib.formatDate`.

Also cleans up a few weird edge cases that popped up while writing the tests.


etpinard:
Looking good. Great test :dancer:

PR #1253, opened 2016-12-14
filters use valuecalendar and targetcalendar instead of calendar

alexcjohnson:
Changes filter attribute `calendar` to `valuecalendar` and adds a new one `targetcalendar` to use if an array is provided as `target`. If `target` is a string (ie referring to an attribute in the same trace) it will ignore `targetcalendar` and instead use the calendar for that attribute (so `xcalendar` if target is `'x'`) which is what it was doing previously.

- [x] hmm, maybe we should still allow `targetcalendar` to be specified, like if you're filtering on `text` or something that doesn't have its own calendar attribute? yeah, I'll do that.

<strike>Note that this is a BREAKING change (`calendar` if supplied is not coerced into `valuecalendar`) but only breaking what's been in master for the last 6 days since merging #1220 </strike>

@VeraZab @cldougl @chriddyp @etpinard

### DIFF
--- a/src/components/calendars/index.js
+++ b/src/components/calendars/index.js
@@ -230,11 +230,13 @@ module.exports = {
         },
         transforms: {
             filter: {
-                calendar: makeAttrs([
-                    'Sets the calendar system to use for `value`, if it is a date.',
-                    'Note that this is not necessarily the same calendar as is used',
-                    'for the target data; that is set by its own calendar attribute,',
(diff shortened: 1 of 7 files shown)


alexcjohnson:
added cleanData step to turn `calendar` into `valuecalendar` for backward compatibility

etpinard:
:dancer:

PR #1254, opened 2016-12-14
Add frames accept null

rreusser:
cc: @bpostlethwaite 

I've added the capability to ignore null frames in `addFrames`. I *think* it's good, but honestly I find the index-based addFrames a bit fragile and undesirable in the first place, so I hate to push 'fixes' through too quickly. All it does is continue to the next frame if the input is null.

### DIFF
--- a/test/jasmine/tests/frame_api_test.js
+++ b/test/jasmine/tests/frame_api_test.js
@@ -43,8 +43,8 @@ describe('Test frame api', function() {
             }).catch(fail).then(done);
         });
 
-        it('compresses nulls when adding frames', function (done) {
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/frame_api_test.js
```diff
@@ -43,6 +43,13 @@ describe('Test frame api', function() {
             }).catch(fail).then(done);
         });
 
+        it('compresses nulls when adding frames', function(done) {
```
etpinard: not _just_ nulls to be fair. but yeah this seems like the correct behavoir

etpinard:
:dancer:

rreusser:
Expecting passed tests; waiting for confirmation just to play it safe.

PR #1257, opened 2016-12-16
Convert arrayOk attribute to calcdata in calc step

etpinard:
This PR makes calc-transforms behave as desired on relayout (e.g. zoom).

Moreover, it should also be a perf boost as trace array to calcdata is now called only in the calc as opposed to the plot step (which is called more often e.g. on zoom).

Now, this PR moved around some pretty ancient code. I'm a little afraid about possible regressions. I'm not 100% confident in our test suite for the various `restyle` call that depend on `arraysToCalcdata`, I'll dig into our test suites to verify.

### DIFF
--- a/src/plot_api/subroutines.js
+++ b/src/plot_api/subroutines.js
@@ -236,7 +236,7 @@ exports.doTraceStyle = function(gd) {
             _module = ((cdi[0] || {}).trace || {})._module || {},
             arraysToCalcdata = _module.arraysToCalcdata;
 
-        if(arraysToCalcdata) arraysToCalcdata(cdi);
+        if(arraysToCalcdata) arraysToCalcdata(cdi, cdi[0].trace);
     }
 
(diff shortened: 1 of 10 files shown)


rreusser:
I like the concept, but I'd really need to dig in and use it to fully appreciate and understand the true implications of the changes.

One small note on carpet: the smoothing should debatably be view-dependent, which would mean it makes a big difference whether some parts are called in the plot or calc step. This sounds fine in that context; the only point being that I'm very interested in this despite not fully understanding the implications.

alexcjohnson:
Looks reasonable to me. Sometime we need to have a discussion about what an explicitly *designed* pipeline should look like in the context of updates; but until then seems like this kind of change (accumulating as many tests along the way as possible) is about the only way.

etpinard:
will be part of `1.21.3` set to be released in the week of Jan 2.

Issue #1181, opened 2016-11-21
Should categorical bars / boxes always have a width of 1?


bpostlethwaite:
http://codepen.io/bpostlethwaite/pen/eBvMWM

All data points except two have been filtered but the plot shows the full range of axis points and they are off center. I don't know if this is related to filtering or is more general but this is a way to trigger it.

cc @etpinard @alexcjohnson

alexcjohnson:
that's a funny one, I think it's two bugs:
- the categories are being read from the original trace, pre-filtering, which doesn't seem like what we want
- then when we make bars only at `a` and `f` they each get a width of 5 categories - also probably not what we want, category axis bars should always have width 1 (one category), right? You can also trigger this second one with two overlaid traces:
```
Plotly.newPlot(gd,[
        {x:['a','b','c','d','e'], y:[1,2,3,4,5], type: 'bar'},
        {x:['a','e'], y:[4,2], type:'bar'}
    ],
    {barmode:'overlay'})
```
![screen shot 2016-11-21 at 4 11 41 pm](https://cloud.githubusercontent.com/assets/2678795/20500855/4334e85c-b005-11e6-9aaf-1e1e1854b20f.png)

n-riesco:
@alexcjohnson Forcing a width of 1 for category axis can be done by changing [this line](https://github.com/plotly/plotly.js/blob/48b035873b66682de512d9473c07cb7c1e331552/src/traces/bar/set_positions.js#L189) and [this line](https://github.com/plotly/plotly.js/blob/48b035873b66682de512d9473c07cb7c1e331552/src/traces/bar/set_positions.js#L233).

etpinard:
> the categories are being read from the original trace, pre-filtering, which doesn't seem like what we want

Fixed in https://github.com/plotly/plotly.js/pull/1260

Changing issue title to reflect what's left of @bpostlethwaite 's bug report.

alexcjohnson:
> Should categorical bars always have a width of 1?

I think so - it's misleading otherwise, as otherwise a bar would bleed into neighboring categories, which in principle have no relation to each other.

etpinard:
A similar problem happens for box traces on categorical position axes - see https://github.com/plotly/plotly.js/issues/4510 for more details.

PR #1260, opened 2016-12-19
Ensure axis autorange is computed after transform operations

etpinard:
To fix the bug, we need to _clear_ the `ax._min` and `_ax._max` arrays (generated in `Axes.expand` done in the `_module.calc` call before the `calcTransform` loop) before the _final_ `module.calc` call.

To do so in the least _intrusive_ way possible, this PR splits `Plots.doCalcdata` operations into separate loops - which hopefully will also help readability.

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -1984,10 +1984,11 @@ plots.doCalcdata = function(gd, traces) {
         }
     }
 
-    // clear stuff ...
+    // clear stuff that should recomputed in 'regular' loop
     for(i = 0; i < axList.length; i++) {
         axList[i]._min = [];
         axList[i]._max = [];
+        axList[i]._categories = [];
     }
 
     // 'regular' loop
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -1961,6 +1961,8 @@ plots.doCalcdata = function(gd, traces) {
         }
     }
 
+    var hasCalcTransform = false;
+
     // transform loop
     for(i = 0; i < fullData.length; i++) {
         trace = fullData[i];
@@ -1978,6 +1980,7 @@ plots.doCalcdata = function(gd, traces) {
 
                 _module = transformsRegistry[transform.type];
                 if(_module && _module.calcTransform) {
+                    hasCalcTransform = true;
                     _module.calcTransform(gd, trace, transform);
                 }
             }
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -1951,46 +1952,51 @@ plots.doCalcdata = function(gd, traces) {
...
-            continue;
-        }
-
-        var trace = fullData[i],
-            cd = [];
-
-        // If traces were specified and this trace was not included, then transfer it over from
-        // the old calcdata:
         if(Array.isArray(traces) && traces.indexOf(i) === -1) {
             calcdata[i] = oldCalcdata[i];
```
alexcjohnson: I know this part isn't new here, but... it isn't used anywhere, right? Looks to me like it was already broken by transforms (calcdata index and trace index aren't the same) and will be broken MORE by this PR (any trace not included will be dropped from autorange) so I'd remove it until we decide to do it right AND use it.

alexcjohnson:
LGTM. Great tests! nonblocking [comment](https://github.com/plotly/plotly.js/pull/1260#discussion_r93086039), 💃

src/plots/plots.js
etpinard: cc @rreusser

alexcjohnson: right, makes a lot of sense to want to do that. I guess for autorange we just need to store the various `_min` and `_max` entries in a different way so we can selectively replace only the ones for the traces we're recalculating. And trace index, depending on how this gets used (it's not yet, right?) you just have to be careful about what you mean by index, before or after transforms potentially insert new traces.

VeraZab:
🎉 
range
this fixes: https://github.com/plotly/streambed/issues/8802
💃 💃 💃

etpinard:
will be part of `1.21.3` set to be released in the week of Jan 2.

src/plots/plots.js
```diff
@@ -1951,46 +1952,51 @@ plots.doCalcdata = function(gd, traces) {
...
+                    _module.calcTransform(gd, trace, transform);
                 }
             }
+        }
+    }
+
+    // clear stuff ...
+    for(i = 0; i < axList.length; i++) {
+        axList[i]._min = [];
+        axList[i]._max = [];
```
etpinard: note to self: maybe we should clear `_categories` too to fix https://github.com/plotly/plotly.js/issues/1181 ?

alexcjohnson: Cool - note though that this only fixes the first part of 1181, not the variant of it mentioned in https://github.com/plotly/plotly.js/issues/1181#issuecomment-262069568

alexcjohnson:
nice. 💃

PR #1017, opened 2016-10-06
Range slider second iteration (part 2: on-par range plots !!!)

etpinard:
fixes https://github.com/plotly/plotly.js/issues/472

This PR builds on top of https://github.com/plotly/plotly.js/pull/946 and https://github.com/plotly/plotly.js/pull/962 to bring range plots (the little little subplot within the range slider) on-par with good old cartesian subplots.

That is, **ALL** cartesian trace types are now supported in the range slider - they now appear as they would in a miniature subplot.

PR #1261, opened 2016-12-19
Honor `fixedrange: false` in y-axes anchored to xaxis with range slider

etpinard:
This should make a bunch of folks on https://community.plot.ly/ happy.

Ever since https://github.com/plotly/plotly.js/pull/1017, there's no reason to hard set `fixedrange: true` for all y-axes anchored to a x-axis that has a range slider: the range plot draw [step](https://github.com/plotly/plotly.js/blob/b823573efb1cd910b6f9406502690f4de0c94f6c/src/components/rangeslider/draw.js#L315-L382) can handle y-axis interactions.

To preserve backward compatibility, we must however make sure that y axes anchored to a x-axis with a range slider default to `fixedrange: true`. With this PR, users can now _set back_ `layout.yaxis.fixedrange: false` if they prefer having interactive y-axes along with their range slider plots.

### DIFF
--- a/src/plots/cartesian/layout_defaults.js
+++ b/src/plots/cartesian/layout_defaults.js
@@ -179,5 +179,22 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
                 axLayoutOut.calendar
             );
         }
+
+        coerce('fixedrange');
+    });
+
+    yaList.forEach(function(axName) {
+        axLayoutIn = layoutIn[axName];
+        axLayoutOut = layoutOut[axName];
+
+        var anchoredAxis = layoutOut[axisIds.id2name(axLayoutOut.anchor)];
+
+        var fixedRangeDflt = (
+            anchoredAxis &&
+            anchoredAxis.rangeslider &&
+            anchoredAxis.rangeslider.visible
+        );
+
+        coerce('fixedrange', fixedRangeDflt);
     });
 };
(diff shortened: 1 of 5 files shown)


etpinard:
ping range slider author :hash::one: @mdtusz

src/plots/cartesian/layout_defaults.js
```diff
@@ -115,33 +115,39 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
+        };
 
         handleAxisDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions, layoutOut);
+
+        var positioningOptions = {
+            letter: axLetter,
+            counterAxes: {x: yaList, y: xaList}[axLetter].map(axisIds.name2id),
+            overlayableAxes: {x: xaList, y: yaList}[axLetter].filter(function(axName2) {
+                return axName2 !== axName && !(layoutIn[axName2] || {}).overlaying;
+            }).map(axisIds.name2id)
```
alexcjohnson: I don't suppose you want to use this opportunity to 🔪  these `map` and `filter` calls? And the `(x|y)alist.forEach` calls below?

alexcjohnson:
nonblocking comment only. 💃

src/plots/cartesian/layout_defaults.js
etpinard: Yeah. That would be nice!

I want to clean up a few artefacts from the old `axes.js` module (which would help out @rreusser for carpets). Namely those annoying `._gd` refs and get rid of the circular dependency between `axes.js` and `plots.js`. I'll make sure to remove those `map` / `filter` in that PR!

Issue #1208, opened 2016-11-30
Support annotation lines connecting all annotation container faces


bpostlethwaite:
Support annotation connector arrow at any of the 4 annotation walls. For example:

![untitled presentation](https://cloud.githubusercontent.com/assets/1176674/20760386/e92e6aaa-b6ed-11e6-9450-c2c007557041.png)

alexcjohnson:
Relatedly, controlling the shape of this line - along the lines of scatter `line.shape`, I'd like to have the options `'hv'` (horizontal line coming out of the note, making a right angle and continuing vertically to the arrowhead), `'vh'`, `'hvh'`, `'vhv'`, and possibly some variants of spline. spline itself doesn't mean anything with just two points, but we should invent modes you might call `'hspline'` and `'vspline'`, ie the start and end of the line are horizontal (or vertical) and they're connected with a cubic spline. Those could be nice additions to `line.shape` for scatter as well, then we could remake the sources tab from shareplot as a regular plotly plot!

alexcjohnson:
Trying to work out how this plays with rotated text: I don't like it, but I think we need very different behaviors with and without arrows.

Without an arrow, when we specify an anchor point as one of the edges or corners (which we already support so we can't really change this), we do the rotation and THEN set the anchor point based on the new bounding box. So for example if you say the anchor is left middle and the position is (0,0), no matter what rotation you specify the text will always be to the right of x=0 and centered on y=0.

But this doesn't seem to make sense when there's an arrow, and its tail is attached to the text box. Then it seems to me like we should define the anchor point BEFORE rotating the text. So if you specify left middle, the tail will always be attached next to the beginning of the words, even if you rotate it 180 degrees so this is actually the right side.

If we only ever rotated by multiples of 90 degrees we could keep the no-arrow behavior with arrows... but if you rotate by some other angle that way, the arrow tail will not in general be attached to the text at all, it'll be floating out in space.

@etpinard @cldougl does that make sense? Any thoughts?

etpinard:
> Then it seems to me like we should define the anchor point BEFORE rotating the text. 

That makes the most sense to me.

Adding an option to either define the anchor before or after rotating the text might end up make things too confusing (and hard to implement).

So, I would be ok with having (slightly) different behavior for `showarrow` true and false annotations.

alexcjohnson:
> So, I would be ok with having (slightly) different behavior for showarrow true and false annotations.

Great, I'll go with it, thanks!

[...]

Issue #751, opened 2016-07-19
Data-referenced annotations on 3D graphs


rudresh-ajgaonkar:
Hey, Is it possible to add labels displaying names(Only Name) of the traces without hover over the data point ? 
Also when i hover on the chart, it should show the regular hover text like for example, in case of a 3d scatter plot, on hover it can show x, y , z data and without hover it should show just the trace name. Is this feature implemented ?  type:feature

Rudresh-Ajgaonkar:
I have read about Text Annotations but i was trying to see them if they are applicable to 3D scatter plots too.

etpinard:
Annotations aren't available in 3D graphs yet. I believe you are the first person to ask for that feature, so I'll change the issue title accordingly.

In the meantime, here's how to make a 3D text chart using the scatter3d `'text'` mode: http://codepen.io/etpinard/pen/rLJRdN

Rudresh-Ajgaonkar:
For Now I have used mode : 'market + text' and the following is the output.
![newplot 5](https://cloud.githubusercontent.com/assets/17093368/16964744/10c80594-4dcb-11e6-9b00-e65b9a6bfdae.png)

Only query here is that why does that text appear about legends ? Aa, currently i have kept it hidden by using .pointtext{
  display: none;
}

etpinard:
Unfortunately there's no way to control the apperence of the legend items at the moment.

I'd suggest making two set of `scatter3d` traces with with `mode: 'markers'` (that will appear in the legend) and one with `mode: 'text'` with `showlegend: false` (that won't appear in the legend).

etpinard:
We should clear up a few things about 3D annotations before implementing them. We'd love any feedback for users watching this issue and of course from @monfera @alexcjohnson @rreusser  @cldougl @chriddyp 

## Appearance 

- Should 3D annotations look like permanent hover labels? That is, their `text` (or arrow tail) coordinates would be an `(x,y)` pair and their arrow tip coordinates would be a `(x,y,z)` on a given 3D subplot (aka scene).

- Or, should 3D annotations look like `{ type: 'scatter3d', mode: 'text' }` traces? That is, both their arrow tip and tails position would be determined by a `(x,y,z)` triplet on a given 3D subplot. 

- Or maybe both? In this case, which one should we implement first?

## API

[...]

PR #1265, opened 2016-12-22
Annotation additions: standoff, anchor with arrow, clicktoshow

alexcjohnson:
A few more options for annotations:

- `xanchor` and `yanchor` work along with `showarrow=true` so you can 1) attach the arrow to any edge or corner of the text box, and 2) be sure that regardless of the note size, it will stay away from the point being marked by exactly as many pixels as you specify for the arrow length. One thing to note: the positioning *if you rotate the text* is different from how it works with `showarrow=false`. Described in https://github.com/plotly/plotly.js/issues/1208#issuecomment-268291536:
  - without an arrow, we first rotate the text box, then calculate the new bounding box and find the appropriate anchor on that. So if you say "left bottom" for the anchor, your note is always above and to the right of the specified coordinates.
  - with an arrow, we attach the specified anchor to the tail of the arrow, and THEN rotate the text box around that anchor point. So now if you say "left bottom" for the anchor, the point that was originally the bottom left corner of the text box stays attached to the arrow tail.

- a new attribute was added, `standoff`, which is a number of pixels to move the arrowhead away from the point it's marking. That's particularly useful if you're marking a scatter point with finite size specified in pixels; then you can ensure the arrowhead is always at the edge of the marker regardless of how the plot is zoomed, and the coordinate it is positioned on is exactly the data point you are marking.

Also I redid all of the position calculations in a way that I hope will be much clearer for future expansions.

cc @etpinard


alexcjohnson:
closes #1208

etpinard:
Hmm. Something is off when panning the axes

gifrecord_2016-12-22_110035

alexcjohnson:
> Hmm. Something is off when panning the axes

Ah good catch - part of the incorrect behavior was already there (while panning, if you start from autoranged, annotations don't get deleted when they go off plot) but the console errors are new. I'll fix both pieces.

package.json
```diff
@@ -77,7 +77,7 @@
     "gl-spikes2d": "^1.0.1",
     "gl-surface3d": "^1.3.0",
     "mapbox-gl": "^0.22.0",
-    "mouse-change": "^1.1.1",
+    "mouse-change": "=1.3.0",
```
alexcjohnson: @etpinard @bpostlethwaite temporary change per https://github.com/plotly/streambed/pull/8893

src/plot_api/plot_api.js
```diff
@@ -2093,7 +2093,7 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, traces) {
     // fill in redraw sequence
     var seq = [];
 
-    if(restyleFlags.fullReplot && relayoutFlags.layoutReplot) {
+    if(restyleFlags.fullReplot || relayoutFlags.layoutReplot) {
```
alexcjohnson: @etpinard am I missing something here? Seems like we would have noticed this before if it's a bug, but clicktoshow wouldn't work right without this change.

src/plots/cartesian/graph_interact.js
```diff
@@ -596,6 +598,13 @@ function hover(gd, evt, subplot) {
 
     gd._hoverdata = newhoverdata;
 
+    // TODO: tagName hack is needed to appease geo.js's hack of using evt.target=true
+    // we should improve the "fx" API so other plots can use it without these hack.
```
alexcjohnson: FYI @etpinard

src/components/annotations/calc_autorange.js
```diff
@@ -45,7 +45,7 @@ function annAutorange(gd) {
     // relative to their anchor points
     // use the arrow and the text bg rectangle,
     // as the whole anno may include hidden text in its bbox
-    fullLayout.annotations.forEach(function(ann) {
+    Lib.filterVisible(fullLayout.annotations).forEach(function(ann) {
```
alexcjohnson: `Lib.filterVisible` was unnecessary before because we didn't coerce any other attributes when visible was false, so these annotations wouldn't look like they're on axes. But now I need the position and reference attributes even when annotations aren't visible in order to determine what to show on clicks, hence this filter is needed too.

alexcjohnson:
Added `clicktoshow` to this PR as it's all somewhat coupled.

@etpinard @bpostlethwaite ready for review.

src/plot_api/plot_api.js
etpinard: No. `&&` looks correct here.

That if-else-block looks like:

```js
if(restyleFlags.fullReplot && relayoutFlags.layoutReplot) {
  // full restyle and relayout update
}
else if(restyleFlags.fullReplot) {
  // full restyle update (but not relayout update)
}
else if(relayoutFlags.layoutReplot) {
    // full relayout update (but not restyle update)
}
else {
  // other more specific updates.
}
```

Maybe the problem is in [`layoutReplot`](https://github.com/plotly/plotly.js/blob/3b85f7ff966293d0c4a2b56c6f8e8c744ee96e24/src/plot_api/subroutines.js#L275-L280)?

src/components/annotations/attributes.js
```diff
@@ -286,17 +297,53 @@ module.exports = {
...
+            'you can show multiple annotations. In *onout* mode, a click',
+            'anywhere else in the plot (on another data point or not) will',
+            'hide this annotation.',
+            'If you need to show/hide this annotation in response to different',
+            '`x` or `y` values, you can set `xclick` and/or `yclick`. This is',
+            'useful for example to label the side of a bar. To label markers',
+            'though, `standoff` is preferred over `xclick` and `yclick`.'
+        ].join(' ')
+    },
+    xclick: {
```
etpinard: `xclick` and `yclick` makes sense in the current state of annotations. 

That said, I'm a little afraid about how these attribute will _scale_ when we add data-referenced to other plot types. I'll elaborate on those concerns in more details in https://github.com/plotly/plotly.js/issues/751

etpinard: gifrecord_2017-01-16_172046

fun :tada:

src/plots/cartesian/graph_interact.js
```diff
@@ -596,6 +598,13 @@ function hover(gd, evt, subplot) {
...
+    if(evt.target && evt.target.tagName) {
+        var hasClickToShow = Registry.getComponentMethod('annotations', 'hasClickToShow')(gd, newhoverdata);
+        overrideCursor(d3.select(evt.target), hasClickToShow ? 'pointer' : '');
```
etpinard: :+1: for `pointer` (`'help'` looks weird to me).

src/components/annotations/attributes.js
```diff
@@ -286,17 +297,53 @@ module.exports = {
...
             'If *auto*, the anchor is equivalent to *middle* for',
-            'data-referenced annotations',
-            'whereas for paper-referenced, the anchor picked corresponds',
-            'to the closest side.'
+            'data-referenced annotations or if there is an arrow,',
+            'whereas for paper-referenced with no arrow, the anchor picked',
+            'corresponds to the closest side.'
+        ].join(' ')
+    },
+    clicktoshow: {
```
etpinard: :+1: for `clicktoshow`

Thanks for that very clear description.

src/plot_api/plot_api.js
alexcjohnson: ah OK - thanks, I missed that. I'll look into `layoutReplot`.

alexcjohnson: alright, not sure what I was seeing earlier with `layoutReplot` / `fullReplot`, but putting back the `&&` (https://github.com/plotly/plotly.js/pull/1265/commits/b683903b9c44a83826e52506135bde077ddac0d1) doesn't seem to cause any issues now. Perhaps I fixed it in some other way in the meantime...

etpinard: Great!

src/lib/override_cursor.js
```diff
@@ -0,0 +1,43 @@
...
+
+var STASHATTR = 'data-savedcursor';
+
+/*
+ * works with our CSS cursor classes (see css/_cursor.scss)
+ * to override a previous cursor set on d3 single-element selections,
+ * by moving the name of the original cursor to the data-savedcursor attr.
+ * omit cursor to revert to the previously set value.
+ */
+module.exports = function overrideCursor(el3, csr) {
```
etpinard: It would be nice to a few test cases for `overrideCursor` similar to the ones for `setCursor` [here](https://github.com/plotly/plotly.js/blob/e2105d2d587168baf24fe8bd4cb113f29ec97c50/test/jasmine/tests/lib_test.js#L1145).

test/jasmine/tests/annotations_test.js
```diff
@@ -296,3 +297,131 @@ describe('annotations autosize', function() {
...
+        };
+    }
+
+    var allIndices = layout().annotations.map(function(v, i) { return i; });
+
+    it('should select only clicktoshow annotations matching x, y, and axes of any point', function(done) {
+        gd = createGraphDiv();
+
+        // first try to select without adding clicktoshow, both visible and invisible
+        Plotly.plot(gd, data, layout())
```
etpinard: Very nicely done :tada:

etpinard:
:dancer: after https://github.com/plotly/plotly.js/pull/1265#discussion_r96646019 is addressed.

src/lib/override_cursor.js
alexcjohnson: haha good call - `overrideCursor` was actually broken, though somehow it looked like it was working. Fixed & tested in https://github.com/plotly/plotly.js/pull/1265/commits/ccbffd0fd85efabd17ede2984141c7d47504b449

etpinard: Thanks :tada:

PR #1280, opened 2017-01-04
Contour bg fix

### LINKED ISSUE #591: This contour trace blows up on zoom
The x/y limits are fine, but the coloring breaks on any zoom event, perhaps a rounding issue?

http://codepen.io/cpsievert/pen/qNEXKV

alexcjohnson:
Fixes #591 and a bunch of other edge cases for contour plots related to `autocontour` and / or partial or self-contradictory specification of contour levels.

### DIFF
--- a/src/lib/coerce.js
+++ b/src/lib/coerce.js
@@ -315,9 +315,10 @@ exports.coerce = function(containerIn, containerOut, attributes, attribute, dflt
  */
 exports.coerce2 = function(containerIn, containerOut, attributes, attribute, dflt) {
     var propIn = nestedProperty(containerIn, attribute),
-        propOut = exports.coerce(containerIn, containerOut, attributes, attribute, dflt);
+        propOut = exports.coerce(containerIn, containerOut, attributes, attribute, dflt),
+        valIn = propIn.get();
 
-    return propIn.get() ? propOut : false;
+    return (valIn !== undefined && valIn !== null) ? propOut : false;
 };
 
 /*
--- a/src/traces/contour/attributes.js
+++ b/src/traces/contour/attributes.js
@@ -44,13 +44,15 @@ module.exports = extendFlat({}, {
     },
     ncontours: {
         valType: 'integer',
-        dflt: 0,
+        dflt: 15,
+        min: 1,
         role: 'style',
         description: [
             'Sets the maximum number of contour levels. The actual number',
             'of contours will be chosen automatically to be less than or',
             'equal to the value of `ncontours`.',
-            'Has an effect only if `autocontour` is *true*.'
+            'Has an effect only if `autocontour` is *true* or if',
+            '`contours.size` is missing.'
         ].join(' ')
     },
 
@@ -59,19 +61,29 @@ module.exports = extendFlat({}, {
             valType: 'number',
             dflt: null,
             role: 'style',
-            description: 'Sets the starting contour level value.'
+            description: [
+                'Sets the starting contour level value.',
+                'Must be less than `contours.end`'
+            ].join(' ')
         },
         end: {
             valType: 'number',
             dflt: null,
             role: 'style',
-            description: 'Sets the end contour level value.'
+            description: [
+                'Sets the end contour level value.',
+                'Must be more than `contours.start`'
+            ].join(' ')
         },
         size: {
             valType: 'number',
             dflt: null,
+            min: 0,
             role: 'style',
-            description: 'Sets the step between each contour level.'
+            description: [
+                'Sets the step between each contour level.',
+                'Must be positive.'
+            ].join(' ')
         },
         coloring: {
             valType: 'enumerated',
--- a/src/traces/contour/calc.js
+++ b/src/traces/contour/calc.js
@@ -10,6 +10,7 @@
 'use strict';
 
 var Axes = require('../../plots/cartesian/axes');
+var extendFlat = require('../../lib').extendFlat;
 var heatmapCalc = require('../heatmap/calc');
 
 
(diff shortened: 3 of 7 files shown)


src/lib/coerce.js
```diff
@@ -315,9 +315,10 @@ exports.coerce = function(containerIn, containerOut, attributes, attribute, dflt
  */
 exports.coerce2 = function(containerIn, containerOut, attributes, attribute, dflt) {
     var propIn = nestedProperty(containerIn, attribute),
-        propOut = exports.coerce(containerIn, containerOut, attributes, attribute, dflt);
+        propOut = exports.coerce(containerIn, containerOut, attributes, attribute, dflt),
+        valIn = propIn.get();
 
-    return propIn.get() ? propOut : false;
+    return (valIn !== undefined && valIn !== null) ? propOut : false;
```
alexcjohnson: @etpinard I included this change bcs contours uses coerce2 on some numeric values that can be 0. This still seems a little weird, ie you have to know to check for exactly `false`, and if `false` is an allowed value it still won't be telling you what you think it is. But this fixes it for the exact case I was worried about here.

src/traces/contour/attributes.js
```diff
@@ -44,13 +44,15 @@ module.exports = extendFlat({}, {
...
+        dflt: 15,
+        min: 1,
         role: 'style',
         description: [
             'Sets the maximum number of contour levels. The actual number',
             'of contours will be chosen automatically to be less than or',
             'equal to the value of `ncontours`.',
-            'Has an effect only if `autocontour` is *true*.'
+            'Has an effect only if `autocontour` is *true* or if',
+            '`contours.size` is missing.'
```
alexcjohnson: So if you specify `ncontours` and `contours: {start, end}` we will automatically choose a "nice" size, even though `autocontour` is `false`. Any reason *not* to allow this usage?

Note that the converse is not allowed: you can't specify `autocontour=true` and use your own `contours.size` value instead of `ncontours` - though I suppose it *would* be clear what this means, as it uses the same machinery as axis ticks, ie it would use a canonical tick of 0 so the contours would be all the multiples of `contours.size` between the data min and max values. That would actually be kind of nice, maybe I'll add it as well.

src/traces/contour/defaults.js
```diff
@@ -33,10 +33,19 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
-        autocontour = coerce('autocontour', !(contourStart && contourEnd));
+        missingEnd = (contourStart === false) || (contourEnd === false),
 
-    if(autocontour) coerce('ncontours');
-    else coerce('contours.size');
+        // normally we only need size if autocontour is off. But contour.calc
+        // pushes its calculated contour size back to the input trace, so for
+        // things like restyle that can call supplyDefaults without calc
+        // after the initial draw, we can just reuse the previous calculation
+        contourSize = coerce('contours.size'),
```
alexcjohnson: This line - always coercing `contours.size` rather than only if `autocontour` is `false` - is actually all you need to fix #591

src/traces/contour/calc.js
```diff
@@ -22,30 +23,72 @@ module.exports = function calc(gd, trace) {
...
         contours.start = Axes.tickFirst(dummyAx);
-        contours.size = dummyAx.dtick;
         dummyAx.range.reverse();
         contours.end = Axes.tickFirst(dummyAx);
 
         if(contours.start === trace.zmin) contours.start += contours.size;
         if(contours.end === trace.zmax) contours.end -= contours.size;
 
-        // so rounding errors don't cause us to miss the last contour
-        contours.end += contours.size / 100;
```
alexcjohnson: this is actually [already dealt with in `contour.plot`](https://github.com/plotly/plotly.js/pull/1280/commits/cef9e66d78bae1db4398bdbf469b1625131837ad#diff-249836043c1e84b5269568da1b828e7cR86), which is easier because we only have to do it once, rather than separately for every edge case we're now going to be checking for here.

src/traces/contour/plot.js
```diff
@@ -103,6 +104,11 @@ function emptyPathinfo(contours, plotinfo, cd0) {
             z: cd0.z,
             smoothing: cd0.trace.line.smoothing
         });
+
+        if(pathinfo.length > 1000) {
+            Lib.warn('Too many contours, clipping at 1000', contours);
```
alexcjohnson: somewhat arbitrary... but consistent with what we do for [cartesian ticks](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js#L728) - here I included the warning as the consequences (in terms of super slow perf) are likely much more severe and you will want to see the `contours` object to know why it happened.

src/lib/coerce.js
etpinard: Nice. 

I've been wanting to merge  `coerce2` with `validate` for a while now.

Good enough for now :+1:

src/traces/contour/calc.js
etpinard: nice catch :+1:

src/traces/contour/attributes.js
etpinard: > Any reason not to allow this usage?

I can't think of one. Nice addition :+1: 

> Note that the converse is not allowed:

So,

```js
autocontour: true,
contours: {
  size: 10
}
```

ignores `contours.size`, but then how does

```js
autocontour: false,
contours: {
  size: 10
}
```

_pick_ `contours.start` and `contours.end` using `Axes.tickFirst`?

src/traces/contour/calc.js
```diff
@@ -22,30 +23,72 @@ module.exports = function calc(gd, trace) {
...
+        if(contours.start > contours.end) {
+            contours.start = contours.end = (contours.start + contours.end) / 2;
+        }
 
         // copy auto-contour info back to the source data.
-        trace._input.contours = contours;
+        trace._input.contours = extendFlat({}, contours);
+    }
+    else {
+        // sanity checks on manually-supplied start/end/size
```
etpinard: looks good :+1:

src/traces/contour/defaults.js
etpinard: :+1:

etpinard:
:dancer: looks good to me.

src/traces/contour/attributes.js
alexcjohnson: Hmm, I was starting to think `autocontour` had become completely obsolete, we’d just fill in whatever missing pieces there were. But actually I think I want to make some changes to this:

- with `autocontour=true` we clear `start`, `end`, and `size` and generate them all automatically from the data range and `ncontours`. Perhaps the only reason you'd need to do this is if you previously set some non-auto values and you want to toss them out.
- with `autocontour=false` we still fill in anything that’s missing but won’t ignore anything the user provides. So for example if you only want to specify `contours.end` to ignore some outliers, you can do that, you don’t have to also specify `contours.start` to get it accepted.

I don't think there's anything backward-incompatible in this proposal, except in partial information cases, where we had a bunch of bugs anyway.

alexcjohnson: deferred to #1282

Issue #1268, opened 2016-12-22
3D hover labels don't show up with `mouse-change@1.4.0`


etpinard:
<strike>**that is 3D hover labels are broken in bundles built from a fresh `npm i`**</strike> --- fixed in `v1.21.3`

Commit https://github.com/mikolalysenko/mouse-change/commit/2aba95014765317eed59b5b58c61aaa6342ef78c and more precisely this [line]( https://github.com/mikolalysenko/mouse-change/commit/2aba95014765317eed59b5b58c61aaa6342ef78c#diff-ba527e2055a8d4c2ab96bee04318fc26R181) ensured that mouse change listeners could be detached. Unfortunately, our gl3d scene code relied on the old (`mouse-change@1.3.0`) buggy behavior [here](https://github.com/plotly/plotly.js/blob/b823573efb1cd910b6f9406502690f4de0c94f6c/src/plots/gl3d/scene.js#L204).

Removing that above line 

```js
 scene.glplot.mouseListener.enabled = false;
```

in `plots/gl3d/scene.js` appears to solve the issue. But, I afraid that it might lead to side-effects such as memory leaks. We should look into this in depth before merging a patch.

etpinard:
cc @dfcreative @alexcjohnson

rreusser:
Am I correct in understanding that plotly simply doesn't use `mouseListener.enabled = true` anywhere? Or if that command was broken, perhaps it *does* use it even though it requests not to. At least I'm not immediately able to find a reference to where it *is* used. It sounds like that might indeed leak a bit of memory then if listeners are attached but never detached.

This is a bit confusing.

PR #1281, opened 2017-01-04
Lock mouse-change@1.3.0

etpinard:
... so that users on `v1.21.x` can see 3D hover labels after a fresh `npm i`.

This PR **does not fix** https://github.com/plotly/plotly.js/issues/1268. Making 3D hover labels work with `mouse-change@^1.4.0` will require a patch that I don't want to make on a patch release.


alexcjohnson:
💃

Issue #1278, opened 2017-01-02
Mapbox hover text concatenates values if there is an empty string in `text`


chriddyp:
http://codepen.io/chriddyp/pen/GrKLEV?editors=0010

The third marker has the hover text `a,b,,d` instead of nothing

![hover-text](https://cloud.githubusercontent.com/assets/1280389/21592624/f9e4809c-d0dc-11e6-8c0c-00737467baa3.gif)

first reported here: https://community.plot.ly/t/multiple-fields-in-hovertext-on-satellite-map/2999/2

PR #1283, opened 2017-01-04
Fix scattermapbox and scattergeo handling of '' text items

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1278

I opted for a strategy similar to what's currently done in [`Drawing.textPointStyle`](https://github.com/plotly/plotly.js/blob/1ef50b91e97e3bfc53a856329649c858e7f0d9ec/src/components/drawing/index.js#L319-L321) to handle all valid `text` input combinations.

cc @alexcjohnson @chriddyp


alexcjohnson:
Nice - super clear tests! 💃

Issue #1231, opened 2016-12-06
`newPlot` does not clear histogram axis type!!!


andrewkfiedler:
Here are the relevant codepens to reproduce:
http://codepen.io/andrewkfiedler/pen/XNqMeW?editors=0010
http://codepen.io/andrewkfiedler/pen/ObZpzv?editors=0010

In the first codepen, you can see that plotting a histogram of type 'Number' causes future histograms to be assumed to be of type 'Number' as well.  As such it throws an error (the second plot is attempted 5 seconds after the first, so wait 5 seconds).

In the second codepen, you can see that plotting a histogram of type 'String' causes future histograms to be assumed to be of type 'String' as well.  As such, the numbers are binned as though they are strings.  

Both are examples are using Plotly.newPlot.

etpinard:
What is the desired behavior here in your mind?

andrewkfiedler:
In my mind, Plotly should not crash when switching the type of data used in a histogram.  'Strings' are useful in histograms to bin things into categories, while 'Numbers' are useful to group things into ranges.  

Using one type should not cause Plotly to assume that's the type for all future histograms.  

I'd actually prefer being allowed to specify the type instead of allowing Plotly to make the decision for me, as right now I have to append a https://en.wikipedia.org/wiki/Zero-width_space in order to force String types that sometimes involve Numbers to be seen as such.

andrewkfiedler:
Anything else I can provide to help out on this?

etpinard:
Oh. You're using `newPlot`. Ok that's bad. Sorry I thought you were complaining about multi-trace graphs being cast to the same axis type. `newPlot` (as the name suggests) should clean up everything about the previous graphs.

Thanks for the report.

etpinard:
Changing the issue title to attract more attention.

etpinard:
The issue can be replicated all the back be to `v1.0.0`: **not a regression**.

etpinard:
We haven't fixed the bug yet, but here's a workaround: http://codepen.io/etpinard/pen/dOwJoZ

Our poor user input object management is responsive for this bug.

alexcjohnson:
Just to be clear what @etpinard means by "poor user input object management" - the issue is that we mutate the `layout` object you pass in, so the workaround is to create a new `layout` object for each `newPlot` call.

We can only fix this in v2, at which point we can clone `data` and `layout` and mutate those rather than the ones you pass in.

etpinard:
[...]

PR #1284, opened 2017-01-06
add fallback in autobin routine

etpinard:
so that trying to plot a histogram with categorical sample pts on a linear axis result in blank plot instead of `ax.dtick` exception.

For example,

```js
Plotly.plot('graph', [{
  type: 'histogram',
  x: ['orange', 'apple']
}], {
  xaxis: {
     type: 'linear'
  }
}
```

currently lead to:

image

with this patch, the above snippet generates a blank graph.

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -514,6 +514,11 @@ axes.autoBin = function(data, ax, nbins, is2d, calendar) {
                 distinctData.minDiff / msexp, [0.9, 1.9, 4.9, 9.9], true);
         size0 = Math.max(minSize, 2 * Lib.stdev(data) /
             Math.pow(data.length, is2d ? 0.25 : 0.4));
+
+        // fallback if ax.d2c output BADNUMs
+        // e.g. when user try to plot categorical bins
+        // on a layout.xaxis.type: 'linear'
+        if(!isNumeric(size0)) size0 = 0;
     }
 
     // piggyback off autotick code to make "nice" bin sizes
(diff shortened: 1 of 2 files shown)


etpinard:
This bug was discovered while debugging for https://github.com/plotly/plotly.js/issues/1231 but does not resolve #1231 

@alexcjohnson looks ok?

src/plots/cartesian/axes.js
```diff
@@ -514,6 +514,11 @@ axes.autoBin = function(data, ax, nbins, is2d, calendar) {
                 distinctData.minDiff / msexp, [0.9, 1.9, 4.9, 9.9], true);
         size0 = Math.max(minSize, 2 * Lib.stdev(data) /
             Math.pow(data.length, is2d ? 0.25 : 0.4));
+
+        // fallback if ax.d2c output BADNUMs
+        // e.g. when user try to plot categorical bins
+        // on a layout.xaxis.type: 'linear'
+        if(!isNumeric(size0)) size0 = 0;
```
alexcjohnson: If a bin size of `0` made it through, this could lead to infinite loops or divide-by-zero errors somewhere else. It looks like `Axes.autoTicks` has its own fallback for this so `dummyAx.dtick` won't be zero, which is why this works for now, but perhaps it would be better to use `size0 = 1` instead so we don't have to depend on this behavior from `autoTicks`?

etpinard: done in https://github.com/plotly/plotly.js/pull/1284/commits/755c1439e16374a2f72280f5d96d0d64d620abb9

test/jasmine/tests/axes_test.js
```diff
@@ -1738,4 +1738,95 @@ describe('Test axes', function() {
...
+            );
+
+            expect(out).toEqual({
+                start: -0.5,
+                end: 2.5,
+                size: 1
+            });
+        });
+
+        it('should not error out for categories on linear axis', function() {
```
etpinard: This case and the two below resulted in errors previously.

alexcjohnson:
💃

Issue #1282, opened 2017-01-04
Consistent "auto*" logic


alexcjohnson:
We have a bunch of different `auto*` attributes scattered about.
- Some control start/end/size type combinations (`autocontour`, `autobin(x|y)` -> `{start, end, size}`, `ncontours`/`nbins(x|y)` is also wrapped up here)
- Some control just a range pair (`autorange` -> `range:[min, max]`, `heatmap.zauto` -> `zmin`, `zmax`)
- At least one other case too (`autosize` -> `width`, `height`).

The only unifying aspect seems to be that the behavior of 2 or 3 output attributes gets set by one `auto` attribute setting.

Right now they all work fine if you either set `auto=true`, or if you provide *all* of the resulting attributes. In https://github.com/plotly/plotly.js/pull/1280#discussion_r94659695 we started discussing how to handle when you supply only *some* of the resulting attributes, and the proposal @etpinard and I seem to agree on is:
- with `auto*=true`, you ignore whatever was provided for the other attributes, and generate them all fresh. Perhaps the only reason you'd need to do this is if you previously set some non-auto values and you want to toss them out.
- with `auto*=false`, you accept whatever was provided for the other attributes, and fill in the rest. So if you want to provide one end of an axis or contour range, or you want to provide the contour size but not the range, you can do that and we'll figure out the rest.

And one extra piece that I think is true now but just for completeness:
- In all of these situations we should copy the calculated values back to the input traces/layout. This is debatable, I guess, but it's important in practice right now because there are code paths (like `restyle`/`relayout`) where you can redo `supplyDefaults` without calling `calc`, which is nice because some of these `auto` calculations are expensive.

alexcjohnson:
@rreusser I think the key is to make sure all this logic as far as possible is in `supplyDefaults`, with the occasional things like `auto*` results calculated in a `calc` step and pointed out in comments and/or PR notes. Then it's just on me and @etpinard to review those parts extra carefully to ensure consistency. I'm not going to pick through every line of your drawing code as long as there are tests that it does what it's supposed to, but the default inheritance I am, for exactly this reason.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #1285, opened 2017-01-06
Clear axis types during `Plotly.newPlot`

### LINKED ISSUE #1231: `newPlot` does not clear histogram axis type!!!

etpinard:
fixes #1231

So that calling `plot` then `newPlot` using the same layout object does not lead to any side-effects.

For example,

```js
var layout = {};

var numberTrace = {
  x: [22000, 22100, 2300],
  type: "histogram"
};

var stringTrace = {
  x: ['apple', 'orange'],
  type: "histogram"
};

Plotly.plot(gd, [numberTrace], layout);

console.log(layout.xaxis.type)  // => 'linear'

// so subsequent `newPlot` will try to plot a categorical histogram
// on  a linear axis.

Plotly.newPlot(gd, [stringTrace], layout);
```

-------------

There are probably other _auto_ attributes (referencing https://github.com/plotly/plotly.js/issues/1282) that could/should get the same treatment in `newPlot`. Obviously, the _real_ way to fix this issue would to **not** mutate the user layout object, but that's a much harder project.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -524,6 +524,14 @@ Plotly.redraw = function(gd) {
 Plotly.newPlot = function(gd, data, layout, config) {
     gd = helpers.getGraphDiv(gd);
 
+    // clear 'auto' axis types that could have leaked in user layout
+    // in previous Plotly.plot calls
+    var axList = Plotly.Axes.list({ _fullLayout: layout || {} });
+    for(var i = 0; i < axList.length; i++) {
+        var ax = axList[i];
+        if(ax && ax.type) ax.type = '-';
+    }
+
     // remove gl contexts
     Plots.cleanPlot([], {}, gd._fullData || {}, gd._fullLayout || {});
 


etpinard:
@alexcjohnson thoughts on this?

alexcjohnson:
If we do that, then you can't use `newPlot` in cases where the axis autotype is wrong.

One way I can imagine doing this that wouldn't be *too* invasive:
- when we set an autotype, save a private var (like `_autotype: true`) in the `fullLayout` axis.
- in your new loop in `newPlot`, instead do something like:

```js
for(var i = 0; i < axList.length; i++) {
    var ax = axList[i];
    var axIn = layout[ax._name];
    if(axIn && axIn.type && ax._autotype) axIn.type = '-';
}
```

etpinard:
Ooops. I forgot to push one commit. Thanks for checking!

But I guess most importantly, would you be _ok_ with this kind of (hacky) solution in `newPlot`?

alexcjohnson:
> would you be ok with this kind of (hacky) solution

yes - hack away :)

etpinard:
Closing. We'll come up with a better solution as part of #1850

PR #1288, opened 2017-01-09
allow style field with sub and sup in text

alexcjohnson:
@etpinard a quick fix for a bug I discovered while working on an application with @charleyferrari

He has a workaround (put the styling into a `<span>` nested inside the `<sup>` so it's not urgent, but the fix was easy so here it is.


etpinard:
looks good :dancer:

PR #1291, opened 2017-01-10
Combine geo and ternary genaral update pattern

etpinard:
Cartesian, geo and ternary subplots (i.e. all SVG/d3 subplots) draw traces using the trace modules' `plot` method which expects as input the `calcdata` traces corresponding to all traces of a given trace type on a given subplot.

Previously, geo and ternary subplot used similar but not-exactly-equivalent logic to handle add/remove/modify updates. This PR - mainly by making ternary subplot use `gd.calcdata` instead of `gd._fullData` traces - unifies the geo and ternary logic into `Plots.generalUpdatePerTraceModule` (I hope you like the name). Cartesian subplots on the other hand are (still) in a world of their own - unifying all SVG/d3 subplots will have to wait.

By doing so, this PR also fixes a bug where:

```js
Plotly.plot('graph', [{
  type: 'scattergeo',   // or 'scatterternary'
  visible: false,
  lon: [/* ... */],
  lat: [/* ... */]
}, {
  type: 'scattergeo',   // or 'scatterternary'
  lon: [/* ... */],
  lat: [/* ... */]
}])
```

would lead to exception and fail to render. See commit  in particular this [line](https://github.com/plotly/plotly.js/commit/fc0af0c9a759decb6f1102dad83bfd5873b99457#diff-ad4f76ccd6044ed16514297078e13b84R2081) for more details.

### DIFF
--- a/src/plots/geo/geo.js
+++ b/src/plots/geo/geo.js
@@ -15,6 +15,7 @@ var d3 = require('d3');
 
 var Color = require('../../components/color');
 var Drawing = require('../../components/drawing');
+var Plots = require('../plots');
 var Axes = require('../cartesian/axes');
 var Fx = require('../cartesian/graph_interact');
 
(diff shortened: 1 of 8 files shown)


src/plots/geo/geo.js
```diff
@@ -170,66 +171,13 @@ proto.plot = function(geoCalcData, fullLayout, promises) {
...
-
-    for(i = 0; i < moduleNames.length; i++) {
-        var moduleCalcData = traceHash[moduleNames[i]],
-            _module = moduleCalcData[0][0].trace._module;
-
-        _module.plot(this, filterVisible(moduleCalcData), geoLayout);
-    }
-
-    this.traceHash = traceHash;
+    Plots.generalUpdatePerTraceModule(this, geoCalcData, geoLayout);
```
rreusser: Lots of trace-specific code moved elsewhere. I like it already. 👏

src/plots/plots.js
```diff
@@ -157,6 +157,31 @@ plots.getSubplotData = function getSubplotData(data, type, subplotId) {
...
+ * Get calcdata traces(s) associated with a given subplot
+ *
+ * @param {array} calcData (as in gd.calcdata)
+ * @param {string} type subplot type
+ * @param {string} subplotId subplot id to look for
+ *
+ * @return {array} array of calcdata traces
+ */
+plots.getSubplotCalcData = function(calcData, type, subplotId) {
+    if(plots.subplotsRegistry[type] === undefined) return [];
```
rreusser: `if(!plots.subplotsRegistry[type])` but that's really nitpicky.

src/plots/plots.js
```diff
@@ -2027,3 +2052,67 @@ plots.doCalcdata = function(gd, traces) {
...
+            traceHash[trace.type].push(calcTraces);
+        }
+    }
+
+    var moduleNamesOld = Object.keys(traceHashOld);
+    var moduleNames = Object.keys(traceHash);
+
+    // when a trace gets deleted, make sure that its module's
+    // plot method is called so that it is properly
+    // removed from the DOM.
```
rreusser: 👏 Yes.

src/plots/plots.js
```diff
@@ -2027,3 +2052,67 @@ plots.doCalcdata = function(gd, traces) {
         calcdata[i] = cd;
     }
 };
+
+plots.generalUpdatePerTraceModule = function(subplot, subplotCalcData, subplotLayout) {
+    var traceHashOld = subplot.traceHash,
```
rreusser: I always feel like the mutate-existing-data approach contributes a bit to fragility, but I understand why these things are as they are.

test/jasmine/tests/plots_test.js
```diff
@@ -549,4 +551,165 @@ describe('Test Plots', function() {
...
+                [ { trace: { type: 'C', visible: false } } ]
+            ], out2.traceHash);
+
+            expect(Object.keys(out3.traceHash)).toEqual(['B', 'A', 'C']);
+            expect(out3.traceHash.B.length).toEqual(1);
+            expect(out3.traceHash.B[0][0].trace.visible).toBe(false);
+            expect(out3.traceHash.A.length).toEqual(1);
+            expect(out3.traceHash.A[0][0].trace.visible).toBe(false);
+            expect(out3.traceHash.C.length).toEqual(1);
+            expect(out3.traceHash.C[0][0].trace.visible).toBe(false);
```
rreusser: The tests are great. 👍

src/plots/plots.js
etpinard: Good call. Done in 83c2722b496a00f91619c0eddba866524206760e

PR #1295, opened 2017-01-11
Clear axis types on some restyle calls

etpinard:
which allows for example to restyle histogram `x` from one axis type to another:

gifrecord_2017-01-11_175650

In detail, `Axes.clearTypes` is called during `restyle` in this [block](https://github.com/plotly/plotly.js/blob/6ebf221f16b7dde558fa8e17d78321e6e3cf6be3/src/plot_api/plot_api.js#L1584-L1587). At the moment, the [attributes](https://github.com/plotly/plotly.js/blob/6ebf221f16b7dde558fa8e17d78321e6e3cf6be3/src/plot_api/plot_api.js#L1334-L1338) that trigger an `Axes.clearTypes` call are the following:

```js
[ 'type', 'x', 'y', 'x0', 'y0', 'orientation', 'xaxis', 'yaxis' ]
```

anything missing?

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -133,14 +133,15 @@ axes.coercePosition = function(containerOut, gd, coerce, axRef, attr, dflt) {
 // empty out types for all axes containing these traces
 // so we auto-set them again
 axes.clearTypes = function(gd, traces) {
-    if(!Array.isArray(traces) || !traces.length) {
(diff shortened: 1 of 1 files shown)


etpinard:
See current behavior here: http://codepen.io/etpinard/pen/oBxxpE

etpinard:
Closing. We'll come up with a better solution as part of https://github.com/plotly/plotly.js/issues/1850

PR #1299, opened 2017-01-13
Test IE9 by mocking window globals

etpinard:
which should help with @alexcjohnson 's https://github.com/plotly/plotly.js/pull/1297 and should be less annoying to setup / maintain than the [`appveyor`](https://github.com/plotly/plotly.js/compare/appveyor) CI.


alexcjohnson:
Looks like a good test of the exact change I made here - thanks! 💃 

That said, I'm not sure how realistic this approach really is, or could be made to be.

For one, a complete test would need to mock globals before importing Plotly, as some of the errors occur when the code is first loaded rather than when it runs (which, if we had done it in this case, would have necessitated loading a subset of the full bundle).

Also, simply setting a property of `window` to `undefined` is not enough, it would need to actually throw an error if you try to access it. Can you `delete` it instead? We could define a getter for it that throws an error, but then I'm not sure if `typeof` would could be made to yield `undefined` in case anyone looks for the property that way...

etpinard:
> For one, a complete test would need to mock globals before importing Plotly

Very good point here. Thanks very much. Here's what I came up with https://github.com/plotly/plotly.js/pull/1299/commits/4ce9df075a02f692ecc54d75bc7127c4b798c42a 

Note that each suite under `test/jasmine/bundle_tests/` is run in a different browser window via the [`test_bundle.js`](https://github.com/plotly/plotly.js/blob/master/tasks/test_bundle.js) task.

test/jasmine/bundle_tests/ie9_test.js
```diff
@@ -0,0 +1,40 @@
...
+var createGraphDiv = require('../assets/create_graph_div');
+var destroyGraphDiv = require('../assets/destroy_graph_div');
+
+describe('Bundle with IE9 supported trace types:', function() {
+
+    afterEach(destroyGraphDiv);
+
+    it('[wip] check that ie9_mock.js did its job', function() {
+        expect(window.ArrayBuffer).toBeUndefined();
+        expect(window.Uint8Array).toBeUndefined();
```
alexcjohnson: cool - can we also do something like
```js
expect(function() { return ArrayBuffer; }).toThrow('ReferenceError: ArrayBuffer is not defined');
```
to make sure it's really inaccessible in the way it *actually* gets invoked in the code?

etpinard: done in 7cb691eeda071100573e1a3f3397968a6f9ad300

alexcjohnson:
Nice, I thought it would be harder to get that to work! 💃

Issue #1298, opened 2017-01-12
Initializing `scattergl` with empty traces throws TypeError


rldotai:
# Summary 

1. Initializing plots with empty traces (traces are defined and styled but have no data yet) throws an error when using `scattergl` for the traces' type.
2. This behavior is not consistent with how Plotly acts when using `scatter`, which allows the user to do this.
3. An error also results when using `Plotly.addTraces` with `scattergl` trace that has no data present, which is also inconsistent with how Plotly behaves when the trace is of type `scatter`. 
4. It seems to be due to an assumption that the traces always have _some_ data in plotly.js/src/components/errorbars/index.js`, but I do not understand enough of the code base to create a pull request, hence this issue.

# Details

In some situations I find it helpful to initialize traces with empty arrays for their data prior to adding values through `Plotly.extendTraces()`.
For example, before data starts arriving, we might wish to style the traces, or ensure that the traces are in some particular order.

This behavior appears to be supported in the SVG case (at least for `type: scatter`), but results in an error when using GL version.

The following works, allowing one to define the traces without adding data, and later extend them [see CodePen demo](http://codepen.io/anon/pen/vgGogG?editors=1011)

```js
var data = [
	{y: [], type: 'scatter'}
];

var layout = {
	title: 'Graph Title'
};

Plotly.newPlot(graphDiv, data, layout);

Plotly.extendTraces(graphDiv, {y: [[1, 2, 3]]}, [0])
```

[If we use `type: 'scattergl'` instead, it throws an error](http://codepen.io/anon/pen/dNMxVY?editors=1011)

```js
var data = [
	{y: [], type: 'scattergl'}
];

var layout = {
	title: 'Graph Title'
};

// This throws an error (Plotly 1.21.3)
Plotly.newPlot(graphDiv, data, layout);
```

From the console, we get:

```
Uncaught TypeError: Cannot set property 'trace' of undefined
    at Object.errorBars.calcFromTrace (eval at <anonymous> (bundle.js:3613), <anonymous>:96224:27)
    at LineWithMarkers.proto.updateFancy (eval at <anonymous> (bundle.js:3613), <anonymous>:147929:31)
    at LineWithMarkers.proto.update (eval at <anonymous> (bundle.js:3613), <anonymous>:147793:14)
    at Object.createLineWithMarkers [as plot] (eval at <anonymous> (bundle.js:3613), <anonymous>:148148:10)
[...]

PR #1300, opened 2017-01-13
gl2d non-visible traces improvements

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1298

Previously scattergl traces with `visible: false` or `visible: 'legendonly'` weren't filtered out of the [`convert`](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js) step where several methods assumed non-empty `trace.x` and `trace.y`.

This PR add a third scattergl update path (along side the current fancy and fast paths) for `visible !== true` traces where gl-vis objects are updated with `Float62Array(0)` as data.

### DIFF
--- a/test/image/mocks/gl2d_scatter-marker-line-colorscales.json
+++ b/test/image/mocks/gl2d_scatter-marker-line-colorscales.json
@@ -176,6 +176,22 @@
                 }
             },
             "uid": "0a4533"
+        },
+        {
+            "type": "scattergl",
+            "x": [],
+            "y": []
+        },
+        {
+            "type": "scattergl",
+            "y": []
+        },
+        {
+            "type": "scattergl",
+            "x": []
+        },
+        {
+            "type": "scattergl"
         }
     ],
     "layout": {
--- a/test/image/mocks/gl2d_scatter-marker-line-colorscales.json
+++ b/test/image/mocks/gl2d_scatter-marker-line-colorscales.json
@@ -192,6 +192,18 @@
         },
         {
             "type": "scattergl"
+        },
+        {
+            "type": "scattergl",
+            "x": [1,2,3],
+            "y": [2,1,2],
+            "visible": false
+        },
+        {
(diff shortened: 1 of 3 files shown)


test/image/mocks/gl2d_scatter-marker-line-colorscales.json
```diff
@@ -176,6 +176,22 @@
...
+        {
+            "type": "scattergl",
+            "y": []
+        },
+        {
+            "type": "scattergl",
+            "x": []
+        },
+        {
+            "type": "scattergl"
```
alexcjohnson: and one each that *has* data but `visible: false || 'legendonly'`?

alexcjohnson:
How hard would it be to test toggling one of these traces on/off, maybe by clicking the legend? Might be overkill given the tests you *do* have...

test/image/mocks/gl2d_scatter-marker-line-colorscales.json
etpinard: done in 2563f17f071c9c3eaa0c7127b2878b8c1cb1cac9

etpinard:
> How hard would it be to test toggling one of these traces on/off,

Not hard. At least it we only test the corresponding `Plotly.restyle` call (as clicking on a legend button would require a `setTimeout` as there's no way to know when the Promise is resolved. Oh wait, I guess we could listen to `'plotly_restyle'`. Oh well, I hope you find the restyle tests satisfactory).

We already have one for gl3d [here](https://github.com/plotly/plotly.js/blob/cebdb1578c3122dd3062cc1c993f09ddf71e1166/test/jasmine/tests/gl_plot_interact_test.js#L196-L210).

I added one for gl2d in https://github.com/plotly/plotly.js/pull/1300/commits/a587f8d31246c0a01e8451e889728968a25f87ba

alexcjohnson:
Those tests are fine. No need to do an actual click.
💃

Issue #130, opened 2015-12-17
gl2d known limitations


etpinard:
A list of known limitations in gl2d graphs in comparison with svg2d feature set:
- [x] **image export isn't reliable** (meaning gl2d graphs have no image test support)
- [x] scatter `mode: 'text'` (for text charts) is not available
- [x] _'free'_ axes and _'overlaying'_ axes are not available. But subplots, insets, shared axes work. **Hurdles:** free axes with custom 'position' would need an option to make a gap between the data-box and the axis in gl-plot2d, overlaying axes would need more logic around when to show grid lines, scene backgroud color and have hover events propagate down. :white_check_mark: Fixed with regl.
- [x] data-reference annotations/shapes are not available (paper-reference annotations/shapes work)
- [x] fill options `tonexty`, `tonexty`, and `toself` are not avaliable (`tozerox` and `tozeroy` work) :white_check_mark: Fixed with regl.
- [x] the `marker.symbol` is pretty good since `v1.28.0` but not on-par with svg :white_check_mark: Fixed with regl.
- [ ] px-specified `line.dash` are not available
- [x] no `line.shape: 'spline'`  **update:** `'hv', 'vh', 'hvh', 'vhv'` added in https://github.com/plotly/plotly.js/pull/3087
- [x] axis titles are not centered (positioning algo does not consider text width)
- [x] tick marks don't support `\n` (most common in date axes)
- [x] axes don't support decreasing (custom or with `autorange: 'reversed'`) ranges. :white_check_mark: Fixed with regl.
- [x] hovermode `x` and `y` are not available (toggling hover on/off in modebar works). :white_check_mark: Fixed with regl.
- [x] no support for axis `fixedrange` (would need to tweak gl-select-box). :white_check_mark: Fixed with regl.
- [x] double click and shift click interactions are not supported. :white_check_mark: Fixed with regl.
- [x] cartesian + gl2d subplot on the same graph are not supported. :white_check_mark: Fixed with regl.
- [x] `plotly_hover`, `plotly_unhover` and `plotly_click` are not emitted  [done in #994]
- [x] setting `hoverinfo` to `'none'` should disable the hover spikes. :white_check_mark: Fixed with regl.
- [x] a hover spikes sometimes appears prematurely after initial `Plotly.plot`  call
- [x] select box and lasso selections are not supported
- [x] single-axis zoom/pan and corner zoom interactions are not supported
[...]

PR #1301, opened 2017-01-13
gl2d data-referenced annotations

etpinard:
set to :heavy_check_mark: one item on https://github.com/plotly/plotly.js/issues/130

gifrecord_2017-01-13_171012

TODO:

- [x] add some jasmine test (and maybe clean up our gl2d test suites along the way)


src/plots/gl2d/scene2d.js
```diff
@@ -286,10 +292,11 @@ proto.updateFx = function(options) {
 
     fullLayout.dragmode = options.dragmode;
     fullLayout.hovermode = options.hovermode;
-};
 
-var relayoutCallback = function(scene) {
+    this.graphDiv._fullLayout = fullLayout;
```
alexcjohnson: this seems a bit scary - why is this necessary? I'm just worried that if `gd._fullLayout` and `scene.fullLayout` have diverged, there could somehow be changes in `gd._fullLayout` that this will throw away.

etpinard: I believe the problem is in the [`doModebar`](https://github.com/plotly/plotly.js/blob/5e4bb1a49f93e59e682a142099923dad3a21f861/src/plot_api/subroutines.js#L293-L319) relayout shortcut. 

I'll double-check why the `_fullLayout` gets lost. Thanks for the headsup!

alexcjohnson: would it be an option to not use `this.fullLayout` at all, but only to use `this.graphDiv._fullLayout`?

etpinard: That's an idea.

So, the `_fullLayout`  gets lost in `Plots.supplyDefaults`. We could add a block handling refs on scene objects similar to this [one](https://github.com/plotly/plotly.js/blob/9b7c18f5f576f103b65889574d25ad9924d4d7c1/src/plots/plots.js#L494-L499) for cartesian axes. But I think we could do better.

The reason the lost of reference only affects `dragmode` relayouts is because all other relayout calls pass through `scene2d.plot` which update the `fullLayout` ref [here](https://github.com/plotly/plotly.js/blob/9b7c18f5f576f103b65889574d25ad9924d4d7c1/src/plots/gl2d/scene2d.js#L353) on every pass.

alexcjohnson: alright, well I'm not entirely sure what would happen, but the test would be something like:
- do a relayout that causes a new `gd._fullLayout` to be constructed by `supplyDefaults` but DOESN'T result in a new `scene2d.plot` (like what? change legend background color? change something cosmetic on an axis of a different subplot?)
- trigger this code by changing dragmode
- do something else that causes a full plot redraw (maybe just `Plotly.redraw(gd)`)
- does the initial relayout get reverted?

etpinard: Here's the _cleanest_ way (I think) of updating `gd._fullLayout._plots.xy._scene2d.fullLayout`

https://github.com/plotly/plotly.js/commit/b50b7e99bb8e720240fae7c4ae0e8e3a1bb8a0ff

alexcjohnson: yeah, that seems better, thanks.

etpinard: full solution in https://github.com/plotly/plotly.js/pull/1319

PR #1302, opened 2017-01-13
Draw updatemenus over sliders

etpinard:
as pointed out by @jackparmer in https://plot.ly/~jackp/17246.embed

When a graph has both updatemenus and sliders, updatemenu buttons drop below sliders.
gifrecord_2017-01-13_174700

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -196,8 +196,8 @@ Plotly.plot = function(gd, data, layout, config) {
 
         Registry.getComponentMethod('legend', 'draw')(gd);
         Registry.getComponentMethod('rangeselector', 'draw')(gd);
-        Registry.getComponentMethod('updatemenus', 'draw')(gd);
         Registry.getComponentMethod('sliders', 'draw')(gd);
+        Registry.getComponentMethod('updatemenus', 'draw')(gd);
 
         for(i = 0; i < calcdata.length; i++) {
             cd = calcdata[i];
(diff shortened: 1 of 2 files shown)


etpinard:
I'll add a jasmine test about it next week. I just wanted to make a quick PR so that I didn't forget about it.

rreusser:
Ah, I thought it would be a complicated SVG layering thing. Changing the order seems much simpler.

src/plot_api/plot_api.js
```diff
@@ -196,8 +196,8 @@ Plotly.plot = function(gd, data, layout, config) {
 
         Registry.getComponentMethod('legend', 'draw')(gd);
         Registry.getComponentMethod('rangeselector', 'draw')(gd);
-        Registry.getComponentMethod('updatemenus', 'draw')(gd);
         Registry.getComponentMethod('sliders', 'draw')(gd);
+        Registry.getComponentMethod('updatemenus', 'draw')(gd);
```
rreusser: Real simple. The fact that it's quite this simple is a little concerning though.

- [Updatemenu is drawn on `_infolayer`](https://github.com/plotly/plotly.js/blob/master/src/components/updatemenus/draw.js#L55)
- [Slider is drawn on `_infolayer`](https://github.com/plotly/plotly.js/blob/master/src/components/sliders/draw.js#L29)

I guess that means it works. For some reason I thought the dropdown was drawn on a separate layer (should it be?). The only corner case I can think of:

1. Draw plot with updatemenus and no sliders
2. Add slider
3. Somehow the slider is appended *after* the updatemenu and ends up on top of it

But I guess that's what containerClass is for. So perhaps this is adequate.

rreusser: Ignore this. Seems all good. Had to mentally work through the hierarchy once more, but yeah, this seems good. I'll only add a small note about long-term robustness of relying upon component ordering. I can imagine needing `_upperinfolayer` or something to which expanded components are added. Not necessary now but might be a better solution if more cases arise.

etpinard: > Not necessary now but might be a better solution if more cases arise.

Yeah. Maybe the **best** solution would have the dropdown buttons in a separate layer above `<g class="infolayer">`. 

My solution aims to fix the 99% cases for users ASAP.

rreusser: Yup. That fix is nicely future-aware, but unless I'm wrong there's not actually any difference in the present.

Issue #1274, opened 2016-12-28
Hoverdata undefined when animating.


rreusser:
Investigating the cause of failure for `plotly_click` to work *sometimes* when animations are taking place. Tracing through things:

1. [`Fx.click` fires](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/graph_interact.js#L1325)
2. [`gd._hoverdata` is `undefined`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/graph_interact.js#L1326) *sometimes*

Investigating the further causes…

cc @bpostlethwaite

rreusser:
I'm having a lot of trouble coming up with a sequence of events that reproduces this 100% of the time, but the problem definitely seems to be the way `hoverdata` is set. It seems like *maybe* it's set on `mousemove` and the *used* on click. Need to investigate further. I suspect the cause is that I'm simply failing to set it 100% of the time when animation changes the hoverdata.

etpinard:
> cc @etpinard I might need some help to get to the bottom of this (tomorrow)

I hope 8 days later is ok too :smile:

rreusser:
Haha, yup!

rreusser:
Hmm… I'm having just a bit of trouble reproducing this. I was mistaken and it appears though some interactions are explicitly disabled in dragbox, click animations do seem to work. It might just take a more complicated example to zero in on buggy behavior. Here's an example that animates continuously and seems to work acceptably with click events: http://rickyreusser.com/demos/plotly-unsupported/interaction-failure.html

See also: #1293

rreusser:
Update. Have isolated a better example. Working on debugging.

etpinard:
@rreusser does https://github.com/plotly/plotly.js/pull/1304 indeed close this issue?

rreusser:
I'll try to do a better job of catching related issues!

PR #1304, opened 2017-01-16
Cache the last mousemove event

rreusser:
This PR attempts to solve #1274 by caching the mousemove event globally and then using it to re-set the hoverdata when the plot is clicked.

I don't think this is really satisfactory, but I didn't manage to come up with a more pleasant solution. The problem is that animations may change the data, but the hoverdata is only set when the mouse *moves*, so that if the data changes and you click before moving the mouse, hoverdata won't be correct. The next best solution seems to be to use the click event data to compute the hover data, except `clientX` and `clientY` aren't available on the click event; they're only available on the mousemove event. So the solution here is to simply use the most recent mouse move event, if available. This could work, but I'm concerned about what it means to only consider a single mousemove event. Does clientX mean the same thing against all subplots that might call this?

I don't think it's a good idea to merge this, but submitting it for feedback so we can come up with a better solution.

cc @etpinard @alexcjohnson @bpostlethwaite


rreusser:
The other option is to be more careful about scoping, caching and reusing the last hover data. The more I look at it and think about it, the more I think this might actually be a decent option. As long as it's robust, it seems much simpler than refactoring things to accomplish the same thing.

etpinard:
This patch looks solid. I'm a little concerned about graphs with multiple cartesian subplots. @rreusser have you tried manually testing multi-subplots mocks?

Moreover, but the looks of it https://github.com/plotly/plotly.js/issues/1274 should only affect cartesian subplots. Is this that case?

rreusser:
Yeah, multiple subplots was my main concern here. Wanted to get feedback on if this is even reasonable or if there's a magic shortcut before going too far with it. This will probably require some refinement to ensure it's fully robust and only affects what it should affect.

bpostlethwaite:
@alexcjohnson can you continue to review this one for us?

alexcjohnson:
thanks @bpostlethwaite 

@rreusser I like the concept of caching the event - but I think rather than re-evaluating it on click, I think it needs to happen after any change to the plot. That way if animation actually changes which point you're over, this will be captured by the displayed hover info (and therefore by the click too).

Note that this issue has existed *forever* (though never bubbled up into an issue...) in the case that something else that wasn't directly initiated by the user changes the plot while the mouse is over it - such as streaming, or restyle/relayout/whatever calls coming from some larger javascript app that the plot is embedded in. Would be great if whatever we do here can solve those cases too.

- Streaming: eg https://plot.ly/~streaming-demos/4/double-pendulum-simulation/ - mouse over the history, the hovertext will remain even when the history has disappeared from that point.
- JS app: here's one using postMessage: http://codepen.io/plotly/pen/ogpGbL?editors=101 - same deal, and same happens if a plot is inserted directly in the page rather than through postMessage to an iframe.

rreusser:
Thanks for the context, @alexcjohnson. I think the path forward then is:

1. understand whether hover is attached to the whole plot or to specific subplots so that I can be sure to get subplot behavior correct, and
2. figure out the right hook for reevaluating. In order to get called with minimal additional effort and at the right times, I feel like maybe it should be called at the end of the calc step for baseplot=cartesian type traces.

rreusser:
I've very slightly modified this to cache the last hover handler call instead. Whenever it runs calcdata, it now executes the last hover callback. I still need to figure out whether hoverdata is being updated at the correct time or not.

The gif below shows that updates the hover to reflect new data even when the mouse hasn't moved:

ezgif com-052d59733f

rreusser:
In order to finish this, I the main thing I need is hints regarding the cases in which it's still likely to fail. Here's what I can come up with:

1. is the end of calcdata the correct place to trigger this? It seemed to make sense since that's identically the same as the data changing, which is when this needs a recheck.
2. does it need to be unset? like if the visibility of a trace is toggled? Or perhaps it should be globally unset at the beginning of each replot?
3. other corner cases related to plot types or otherwise?

cc @alexcjohnson @etpinard

alexcjohnson:
It seems a little funny to me to have something as presentational as hover take place during the calc stage. And there may be some changes that affect the hover label but don't recalc - like changing trace color? I *think* `Plots.style` gets called near the end of all of these paths, though that also seems like a little bit of a nonintuitive place to put it. Maybe checking for rehover actually warrants being an explicit step that `Plotly.plot`, `Plotly.restyle`, `Plotly.relayout`, and `Plotly.update` all tack onto their queues?

Another thing I worry about is whether `maindrag` always still exists and represents the actual dragger at the end of the update. We're probably not handling its creation and update using regular d3 idioms... until we do, might rehover fail if you do something like change a subplot size in the update - like `Plotly.relayout(gd, {xaxis.domain: [0, 0.5]})`? In a case like that I'd be fine if we just drop the hover label entirely - something like `try { hover(...) } except { unhover(...) }`.

Anyway the general behavior your demo shows (hover label jumps while animation is smooth) seems totally reasonable.

rreusser:
Agreed. Pardon the burden, but that's exactly the feedback I was hoping for while it was still only a five-line patch. It would be possible to unpack values from the event so everything is definitely current when used later, but I think some of the references were only accessible in the closure, which is where the out-of-date reference issue becomes tricky to resolve.

I'll iterate on this and see if I can make it a bit more satisfying. Thanks!

rreusser:
@alexcjohnson I've taken another shot at refining this by explicitly appending it to the async queue where needed. I've also added matching of the subplot and unsetting the name of the hovered subplot when the mouse leaves maindrag. This prevents some corner cases like when it would reapply the hover inappropriately after the mouse has fully left the plot.

I think it's doing a bit of unnecessary work by applying the hover more often than necessary. It's possible this could be optimized by more carefully relying on flags like `docalc` etc., but I'm hesitant to get too elaborate. Cost = dev time, code complexity, extra processing, Benefit = calling it more often than necessary ensures it's more likely to be correct and not missed.

src/plots/plots.js
```diff
@@ -1934,7 +1934,11 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
         }
     }
 
-    var seq = [plots.previousPromises, interruptPreviousTransitions, prepareTransitions, executeTransitions];
+    function rehover() {
+        plots.rehover(gd);
+    }
```
alexcjohnson: This wrapper isn't necessary, as all the items get `gd` passed as an arg.

rreusser: Ahh, thanks. Wondered about that but assumed the opposite.

src/plots/cartesian/graph_interact.js
```diff
@@ -129,6 +144,11 @@ fx.init = function(gd) {
             maindrag.onmouseout = function(evt) {
                 if(gd._dragging) return;
 
+                // When the mouse leaves this maindrag, unset the hovered subplot.
+                // This may cause problems if it leaves the subplot directly *onto*
+                // another subplot, but that's a tiny corner case at the moment.
+                gd._fullLayout._hoversubplot = null;
```
alexcjohnson: Inset plots seem like a not-so-tiny corner case. But does this actually cause problems? Presumably `mouseout` happens before `mouseover` in that case?

src/plots/cartesian/graph_interact.js
```diff
@@ -114,6 +114,21 @@ fx.init = function(gd) {
...
+                // Track the hovered subplot. This prevents rehover from accidetally
+                // reapplying a hover label after the mouse has left the plot or if
+                // the mouse has entered another subplot.
+                gd._fullLayout._hoversubplot = plotinfo.id;
+
+                gd._fullLayout._rehover();
+
                 fx.hover(gd, evt, subplot);
                 fullLayout._lasthover = maindrag;
                 fullLayout._hoversubplot = subplot;
```
alexcjohnson: Are this `fx.hover` and `fullLayout._hoversubplot = subplot` still necessary then?

src/plots/cartesian/graph_interact.js
rreusser: Yes, which happens first was the question. If mouseover and then mouseout, it would end up unset, but the *only* time that could happen is if you move *exactly* one pixel outside of the inset region, which is where it started to seem like a small corner case. The second pixel moved will immediately fix things.

src/plots/cartesian/graph_interact.js
rreusser: Oops! Good catch. I change this back and forth and accidentally left both. Fixed in the commit below.

rreusser:
@alexcjohnson Thanks for the quick review. I made the changes you pointed out. I can add a test that catches the straightforward cases if you think the proposed fix is starting to look reasonable.

src/plots/cartesian/graph_interact.js
rreusser: Oops. Yeah, you're right. There's still an extra item or two there. Fixing (and looking at the *full* diff when doing so this time)

alexcjohnson:
Yes, this looks great! 🔒  it down with some tests and we should be ready to roll.

> It's possible this could be optimized by more carefully relying on flags like docalc etc., but I'm hesitant to get too elaborate

Agreed 100%

rreusser:
Added test to verify it's updated when data changes (and nothing else, i.e. no replot, no mouse event). Used animation to accomplish this. That doesn't exhaustively test every case, but it at least verifies the fix is occurring.

alexcjohnson:
Could you add a test where an update causes the hover label to disappear, then another update causes one to appear again? Maybe a `relayout` changing an axis range, so we're also testing a route that app developers are likely to use when altering the data in their plots, and similar to the route used by streaming?

rreusser:
I've expanded the test so that it now:

1. shows label
2. animates data (no replot) and confirms point in other trace now hovered
3. relayout to move data offscreen & verify hover label removed
4. relayout to restore original axis limits & verify hover label restored

alexcjohnson:
Awesome, thanks! 💃

PR #1305, opened 2017-01-16
bump mouse-change dep to ^1.4.0

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1268

... and makes 3D hover labels compatible with it.


alexcjohnson:
💃

PR #1306, opened 2017-01-17
Fix range slider svg / eps / pdf exports

etpinard:
Range slider svg / eps / pdf exports are currently failing on plot.ly. 

This PR ensures that

- all range slider `<rect>` elements have set x/y/width/height attributes (no setting those makes the SVG invalid)
- RGBA color are set via `fill` + `fill-opacity` attributes (Batik doesn't like `fill: ${rgba}`)
- all `fill: 'transparent` elements are not drawn on static plots (Batik doesn't like `fill: 'transparent'`)


alexcjohnson:
LGTM 💃

PR #1307, opened 2017-01-17
Fix XSS in trace name on hover

etpinard:
Report:

- make a scatter plot with at least two traces
- edit the name of one of the traces (in the legend) to: 
- hover over any point in the trace with the bad name, the console.log will execute.
- for bonus points, hide that trace name from the legend, and make that trace look like it's part of (or all of!) another trace - the user will see nothing.


etpinard:
cc @alexcjohnson

alexcjohnson:
💃

PR #1309, opened 2017-01-18
Contour bug

alexcjohnson:
@etpinard @chriddyp fixes https://github.com/plotly/streambed/issues/4236 and another edge case I found while investigating it.


src/traces/contour/colorbar.js
```diff
@@ -52,7 +53,7 @@ module.exports = function colorbar(gd, cd) {
         })
         .levels({
             start: contours.start,
-            end: contours.end,
+            end: endPlus(contours),
```
alexcjohnson: `colorbar` actually adds some more to `end` for the same purpose - but it turns out not to matter in this case, additional expansion doesn't have any ill effects that I could find, as long as it always ends up at or beyond the level we used to make the `colorscale`. Sometime though we may want to fold ALL the places we extend a range a little for rounding purposes into the same `endPlus` mold.

alexcjohnson:
Gah, good thing our test cases include yet *another* edge case... more work to do here.

alexcjohnson:
@rreusser the [test image](https://github.com/plotly/plotly.js/pull/1309/files#diff-34eeda8aa1783dd2381a3c2ea97ad0ca) I added here shows some contour crossings - well beyond the scope here to clean those up, this is just to fix the overall topology, but when we get around to making a better smoothing algorithm, these will be good ones to watch and fix. There are also cases where a contour crosses itself, which are not included here but I can pull some up when we get to it.

src/traces/contour/find_all_paths.js
```diff
@@ -84,13 +84,13 @@ function makePath(pi, loc, edgeflag) {
...
+            closedLoop = (locStr === startLocStr) && (marchStep.join(',') === startStepStr);
+
         // have we completed a loop, or reached an edge?
-        if((locStr === startLocStr && marchStep.join(',') === startStepStr) ||
-            (edgeflag && (
-                (marchStep[0] && (loc[0] < 0 || loc[0] > n - 2)) ||
-                (marchStep[1] && (loc[1] < 0 || loc[1] > m - 2))))) {
-            break;
-        }
+        if((closedLoop) || (edgeflag && atEdge)) break;
```
alexcjohnson: I could revert this change actually, as I don't need `atEdge` separately anymore. It's way more readable now, but a bit faster before as it could short-circuit some of the logic...

etpinard: Your call here. I'm fine with it either way.

etpinard:
:dancer:

src/traces/contour/find_all_paths.js
alexcjohnson: Actually, if we're worried about perf, I should get rid of all those `str = pair.join(',')` statements - they make the comparison code shorter, but it's way faster to just compare integers twice. Since `marchStep.join(',')` is in the first clause it would save almost nothing to revert.
https://github.com/plotly/plotly.js/issues/1311

PR #1314, opened 2017-01-18
Prevent uncaught exception in plot_api _relayout()

JonDum:
Happens when `fixedrange` is set to true of a line chart's `yaxis`


etpinard:
Can you share a reproducible example of the bug you're trying to fix? Thank you.

etpinard:

@JonDum if you can reproducible that bug you're seeing with the latest master in a codePen / jsFiddle / jsBin example please open a new issue [here](https://github.com/plotly/plotly.js/issues/new). Thank you.

PR #1315, opened 2017-01-18
Still animate layout objects if there's no transitioned axis

rreusser:
This PR solves an oversight on my part. At the moment if there's no transitioned axis, then layout objects aren't updated. This PR solves that by calling the layout update if the axes are not otherwise affected.

cc @jackparmer

### DIFF
--- a/src/plots/cartesian/transition_axes.js
+++ b/src/plots/cartesian/transition_axes.js
@@ -102,8 +102,7 @@ module.exports = function transitionAxes(gd, newLayout, transitionOpts, makeOnCo
 
     function updateLayoutObjs() {
         function redrawObjs(objArray, method) {
-            var i;
-            for(i = 0; i < objArray.length; i++) {
+            for(var i = 0; i < objArray.length; i++) {
                 method(gd, i);
             }
         }
--- a/src/plots/cartesian/transition_axes.js
+++ b/src/plots/cartesian/transition_axes.js
@@ -100,7 +100,21 @@ module.exports = function transitionAxes(gd, newLayout, transitionOpts, makeOnCo
     var updatedAxisIds = Object.keys(updates);
     var affectedSubplots = computeAffectedSubplots(fullLayout, updatedAxisIds, updates);
 
+    function updateLayoutObjs() {
+        function redrawObjs(objArray, method) {
+            var i;
+            for(i = 0; i < objArray.length; i++) {
+                method(gd, i);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


rreusser:
@etpinard @alexcjohnson Layout animation doesn't currently update container objects unless there's also an axis update. This branch checks for and triggers container object updates in case there is no axis update. I've rebased and added a test for this fix.

rreusser:
Nothing urgent, but @etpinard @alexcjohnson does this seem alright? Trying to close out old things.

src/plots/cartesian/transition_axes.js
```diff
@@ -100,7 +100,20 @@ module.exports = function transitionAxes(gd, newLayout, transitionOpts, makeOnCo
...
+    function updateLayoutObjs() {
+        function redrawObjs(objArray, method) {
+            for(var i = 0; i < objArray.length; i++) {
+                method(gd, i);
+            }
+        }
+
+        redrawObjs(fullLayout.annotations || [], Registry.getComponentMethod('annotations', 'drawOne'));
+        redrawObjs(fullLayout.shapes || [], Registry.getComponentMethod('shapes', 'drawOne'));
+        redrawObjs(fullLayout.images || [], Registry.getComponentMethod('images', 'draw'));
```
etpinard: Looking at [`dragbox.js`](https://github.com/plotly/plotly.js/blob/9948af4f345fac61280677f367d67a4599115c4a/src/plots/cartesian/dragbox.js#L580), images can short circuit that loop over `objArray` above.

test/jasmine/tests/annotations_test.js
```diff
@@ -1344,3 +1344,41 @@ describe('annotation effects', function() {
...
+            expect(expected.length).toEqual(texts.size());
+
+            texts.each(function(i) {
+                expect(Plotly.d3.select(this).text()).toEqual(expected[i]);
+            });
+        }
+
+        Plotly.plot(gd,
+            [{y: [1, 2, 3]}],
+            {annotations: [{text: 'hello'}]}
```
etpinard: It would be nice to :lock: down the behavior for `shapes` and `images` too.

src/plots/cartesian/transition_axes.js
rreusser: Copied the short-circuit code.

test/jasmine/tests/annotations_test.js
rreusser: ✅

etpinard:
Beautifully done. :dancer:

PR #1327, opened 2017-01-23
fixes 3d thumbnail clone when no xaxis in scene

bpostlethwaite:
The test case that now (should) be passing was failing before this fix. This impacted Plotly image server thumbnail generation.

### DIFF
--- a/src/snapshot/cloneplot.js
+++ b/src/snapshot/cloneplot.js
@@ -113,14 +113,26 @@ module.exports = function clonePlot(graphObj, options) {
             };
         }
         for(i = 0; i < sceneIds.length; i++) {
-            var sceneId = sceneIds[i];
+            var scene = newLayout[sceneIds[i]];
 
-            extendFlat(newLayout[sceneId].xaxis, axesImageOverride);
-            extendFlat(newLayout[sceneId].yaxis, axesImageOverride);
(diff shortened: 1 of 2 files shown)


src/snapshot/cloneplot.js
```diff
@@ -115,9 +115,21 @@ module.exports = function clonePlot(graphObj, options) {
...
+                newLayout[sceneId].yaxis = {};
+            }
+
+            if(!newLayout[sceneId].zaxis) {
+                newLayout[sceneId].zaxis = {};
+            }
+
+            extendFlat(newLayout[sceneId].xaxis || {}, axesImageOverride);
+            extendFlat(newLayout[sceneId].yaxis || {}, axesImageOverride);
+            extendFlat(newLayout[sceneId].zaxis || {}, axesImageOverride);
```
alexcjohnson: `|| {}` is unnecessary here (and would have been confusingly useless by itself)

I'd probably also do `var scene = newLayout[sceneIds[i]]` instead of `var sceneId = sceneIds[i]` to be more concise, but that's more of a style issue.

bpostlethwaite:
@alexcjohnson changes made ready for review

alexcjohnson:
💃  Thanks!

Issue #166, opened 2016-01-07
IE9 JS Plotly.redraw and Plotly.addTraces both throws Invalid Argument


micwc:
I used the recommended way to load plotly for IE9 with the typedarray
```
<script>if(typeof window.Int16Array !== 'function')document.write("<scri"+"pt src='extras/typedarray.min.js'></scr"+"ipt>");</script>

<script type="text/javascript" src="plotly.min.js"></script>
```

I am trying to do a simple "bar" chart with three traces. However, both addTraces (adding one trace at a time) and redraw (array of traces) would throw "Invalid Argument" error in IE9 and no bars on graph

IE10+ works
Chrome works

Is there a workaround for this in IE9? It is a little urgent

thanks for any help

etpinard:
Can you screenshot where the error occurs?

(I don't have access to a machine with IE9, and I'd rather not debug using browserstack).

Thank you!

micwc:
![image](https://cloud.githubusercontent.com/assets/16586417/12162035/f964748a-b4ca-11e5-8574-c62f1a7739f1.png)

version 1.3.1
fo.setProperty=function(t,e,r){po.call(this,t,e+"",r)}

let me know if more info is needed, thanks

micwc:
![image](https://cloud.githubusercontent.com/assets/16586417/12162211/ae4d790e-b4cc-11e5-9ce3-5b826114370d.png)

let me know if more info is needed, thanks

etpinard:
@micwc Could paste a screenshot of the callstack leading to the error?

Thank you very much!

micwc:
![image](https://cloud.githubusercontent.com/assets/16586417/12162395/647f6aaa-b4cf-11e5-99cb-cb37fa5130da.png)

![image](https://cloud.githubusercontent.com/assets/16586417/12162402/86d1ae42-b4cf-11e5-9d24-98242eda2d56.png)

![image](https://cloud.githubusercontent.com/assets/16586417/12162427/f50cbae6-b4cf-11e5-930f-2ee4d0eb04c1.png)

etpinard:
@micwc thanks!

This line https://github.com/plotly/plotly.js/blob/master/src/components/legend/index.js#L465 see a little strange.

Perhaps changing it to:

![image](https://cloud.githubusercontent.com/assets/6675409/12162534/8d73d8f4-b4d1-11e5-9f85-e227afa815c5.png)

could fix it.

micwc:
[...]

Issue #1247, opened 2016-12-12
[IE] Y-axis values overlap the title


maxcopters:
Hello,
I have been using plot.ly to draw classical charts, but I observed an issue when my values are very little or very large. This issue only appears on Internet Explorer and Edge (on win10), as I have tested it on Firefox and Chrome.
Here is an example: https://codepen.io/anon/pen/XNPpKN

Could you check this issue?

Thanks!

etpinard:
Thanks for the report.

asolagmbh:
I don't know if it's the same problem or not.
In IE 11 and Edge if you use the pan function for dragging the graph the line of the graph overlap with the y axis. It seems that the line are not cropped in the correct area.
There are plan for resolve this issue? 
Thanks a lot!

maxcopters:
Hello,
I would be very interested in the information whether this ticket could be integrated soon or not.

Regards!

Issue #746, opened 2016-07-14
Malformed hovertext in IE11


cpsievert:
I don't have IE11, but this was reported in https://github.com/ropensci/plotly/issues/636

http://codepen.io/cpsievert/pen/wWpzzr

In IE11:

![image](https://camo.githubusercontent.com/c038af84e05619da3f01ed5c3600a76ef200ac81/687474703a2f2f692e737461636b2e696d6775722e636f6d2f57493669462e706e67)

rreusser:
Everyone is probably good on the IE-debugging front, but just FYI, useful resources:

- [free IE virtual machines from MS](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)
- http://browserstack.com

PR #1332, opened 2017-01-25
IE9 fixes - to make SVG maps compatible

alexcjohnson:
A bunch of random fixes for IE9 behavior, only one was actually *in* maps, oddly enough (so some edge cases may have been broken in other plot types too), but all of which are *accessed by* maps.

- ~~`classList`~~
- ~~`instanceof d3.selection`~~ -> seems to work for built-in types and some user-defined types, but for some reason `d3.selection` broke it, replaced with duck typing
- ~~`console[fn].apply`~~ - I guess these aren't regular functions in IE9, so if there's no `apply` just call the function once for each argument.
- When an invalid attribute value is provided, IE9 throws an error rather than just using a default value. Choropleth with no explicit line width was trying to set `undefinedpx` via `drawing.dashLine`, which in all other browsers silently defaulted to `1px`. I made this explicit at the `supplyDefaults` level, while at the same time defaulting to `0px` in `dashLine` so this can't cause further problems.

cc @chriddyp @etpinard 

Given the scattered nature of these fixes, it's probably worthwhile to have someone spend some time manually trying out more plots & behaviors, ideally on a real local copy of IE9 rather than Browserstack. I saw some sporadic glitches with drag interactions, including one where it seemed like the hover effects were pointing to the old point locations... but couldn't reproduce so I'm concerned it was just an artifact of tunneling through Browserstack.


tasks/test_syntax.js
```diff
@@ -56,7 +60,15 @@ function assertHeaders() {
 
             // parse through code string while keeping track of comments
             var comments = [];
-            falafel(code, {onComment: comments, locations: true}, function() {});
+            falafel(code, {onComment: comments, locations: true}, function(node) {
+                // look for .classList
+                if(node.type === 'MemberExpression') {
+                    var parts = node.source().split('.');
+                    if(parts[parts.length - 1] === 'classList') {
+                        logs.push(file + ' : contains .classList (IE failure)');
```
alexcjohnson: I was hoping eslint would have an option for IE compatibility (actually not even IE11 supports `classList` on SVG elements, apparently)... but I don't see one, so I added it in here. Confirmed that before the change I made to `modebar.js` it failed this test.

test/jasmine/tests/lib_test.js
```diff
@@ -1599,6 +1600,172 @@ describe('Test lib.js:', function() {
...
+    });
+
+    describe('loggers', function() {
+        var stashConsole,
+            stashLogLevel;
+
+        function consoleFn(name, hasApply, messages) {
+            var out = function() {
+                var args = [];
+                for(var i = 0; i < arguments.length; i++) args.push(arguments[i]);
```
alexcjohnson: because `toEqual` doesn't think `arguments` is a regular array

tasks/test_syntax.js
chriddyp: Great idea!

src/components/drawing/index.js
```diff
@@ -112,6 +112,7 @@ drawing.lineGroupStyle = function(s, lw, lc, ld) {
 };
 
 drawing.dashLine = function(s, dash, lineWidth) {
+    lineWidth = +lineWidth || 0;
```
etpinard: Interesting. Maybe this also fixes https://github.com/plotly/plotly.js/issues/166 ?

alexcjohnson: Maybe - I'll try that tonight and see if it implies any further fixes. Any other IE issues I should include in my testing?

etpinard: Maybe

- https://github.com/plotly/plotly.js/issues/1247
- https://github.com/plotly/plotly.js/issues/746

alexcjohnson: I think this does fix #166 - and https://github.com/plotly/plotly.js/pull/1332/commits/4aa1a4535a63633f38256159fc751ca59dd7b2bd should catch any of these `nanpx` or `undefinedpx` errors, at least as long as they go through d3.

Will investigate the other two tomorrow, they may be related.

test/image/index.html
```diff
@@ -5,6 +5,7 @@
 <script type="text/javascript" src="../plotly.js/dist/extras/mathjax/MathJax.js?config=TeX-AMS-MML_SVG"></script>
 <script type="text/javascript" src="../plotly.js/build/plotly.js" charset="utf-8"></script>
 <script type="text/javascript" src="../plotly.js/dist/plotly-geo-assets.js" charset="utf-8"></script>
+<script type="text/javascript" src="../plotly.js/test/image/strict-d3.js" charset="utf-8"></script>
```
etpinard: Great idea here!

First, I was thinking adding this to the image test index might lead to some very annoying debugging sessions when things go wrong. But, by adding it to the test-dashboard as well essentially make my concerns disappear  (except for devs that don't use the test-dashboard cc @rreusser @monfera).

Alternatively, we could mock the `d3.selection.prototype` only in the [`ie9_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/bundle_tests/ie9_test.js) bundle test and test each ie9 falback one by one.

It comes down to choosing between a solution with great coverage (yours) and a less intrusive solution that might lead to less :face_with_head_bandage: down the road. I'm ok with both.

alexcjohnson: We've talked before about wanting a way to report errors out of the image tester - printing the stack trace to a blank image or something. My vote would be to leave this in and work toward that goal. Being strict here will help us even outside IE, as we really never want to implicitly rely on browser defaults.

src/lib/index.js
```diff
@@ -451,8 +451,12 @@ lib.addStyleRule = function(selector, styleString) {
     else lib.warn('addStyleRule failed');
 };
 
-lib.getTranslate = function(element) {
+// TODO (alexcjohnson): the following translate/scale functions should
+// get moved to components/drawing rather than the generic lib.
```
alexcjohnson: I didn't want to move all of these (along with all their references and tests) without a discussion, but I did write a new one before finding this in here and discovering that it already tests the behavior I needed it to have for IE.

src/lib/svg_text_utils.js
```diff
@@ -278,6 +280,13 @@ function encodeForHTML(_str) {
 function convertToSVG(_str) {
     _str = convertEntities(_str);
 
+    // normalize behavior between IE and others wrt newlines and whitespace:pre
+    // this combination makes IE barf https://github.com/plotly/plotly.js/issues/746
+    // Chrome and FF display \n, \r, or \r\n as a space in this mode.
+    // I feel like at some point we turned these into <br> but currently we don't so
+    // I'm just going to cement what we do now in Chrome and FF
+    _str = _str.replace(NEWLINES, ' ');
```
alexcjohnson: As far as I can tell this is purely a bug in IE - though one that persists from 9 through 11... the affected text is invisible, even though `getBoundingClientRect()` on its node shows the correct position and shape, and it reappears when you take out the newline character.

src/lib/index.js
etpinard: :+1: for moving `getTranslate` and `setTranslate` out of `Lib`. Maybe putting them into `Drawing` would make the most sense?

src/lib/svg_text_utils.js
etpinard: :1st_place_medal: debugging.

test/jasmine/tests/hover_label_test.js
```diff
@@ -123,7 +125,8 @@ describe('hover info', function() {
 
             expect(d3.selectAll('g.axistext').size()).toEqual(0);
             expect(d3.selectAll('g.hovertext').size()).toEqual(1);
-            expect(d3.selectAll('g.hovertext').select('text').html()).toEqual('hover text');
+            expect(d3.selectAll('g.hovertext').select('text').html())
+                .toEqual('hover text  with spaces  not newlines');
```
alexcjohnson: None of the image tests in https://github.com/plotly/plotly.js/pull/1332/commits/21059303ee919812e4b2e281df5f71f7c77cf7be actually changed results with the fix there, but they would all have resulted in invisible text in IE before the fix. But this test confirms that we do sanitize newlines for IE compatibility.

etpinard: Nicely done.

src/lib/index.js
alexcjohnson: Yeah, that's where I put mine before noticing these. Great, I'll move them in there.

alexcjohnson: done in https://github.com/plotly/plotly.js/pull/1332/commits/66baf516524d114ba2044b2f432a3c3441c5ea43

Any final comments @etpinard ?

etpinard:
:dancer:

PR #1336, opened 2017-01-26
Fix #1328 skip translatePoint for empty nodes

### LINKED ISSUE #1328: Error when pan scatter with empty text
Change https://github.com/plotly/plotly.js/blob/master/src/components/drawing/index.js#L54
from `if(isNumeric(x) && isNumeric(y)) {` to `if(isNumeric(x) && isNumeric(y) && sel.node()) {`
solved this problem.

Demo: https://jsfiddle.net/indream/6L8nLgr1/

inDream:
Fix #1328

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -51,7 +51,7 @@ drawing.translatePoint = function(d, sel, xa, ya) {
     var x = d.xp || xa.c2p(d.x),
         y = d.yp || ya.c2p(d.y);
 
-    if(isNumeric(x) && isNumeric(y)) {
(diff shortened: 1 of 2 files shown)


etpinard:
@inDream looks good. Big thanks!

Before merging this in, we'll need to add a test case. Would you interested in doing so?

I was able to reproduce this bug using `Plotly.relayout` (see https://jsfiddle.net/6L8nLgr1/2/) which will make adding the test case a lot easier.

inDream:
@etpinard I've added test but don't know whether is it the correct location.

etpinard:
Looking great @inDream 

Thanks very much for doing this. I can't think of a better place to put this test case. Great job :beers:

PR #1337, opened 2017-01-26
Fixes #1324 which occurs due to a collision in the datafn.

### LINKED ISSUE #1324: Missing certain xaxis labels
When using ticktext it appears that with certain data sets there are missing xaxis labels. In this example the xaxis label for 'Jan 1' is missing even though the marker for that data point is drawn and the label is shown on hover:
https://jsfiddle.net/hgqhrnrj/5/

The bug only seems to occur with very specific data sets, for example if you remove or reorder one of the items from the data array then 'Jan 1' gets drawn properly.

rpaskowitz:
In the example Jan 11 is at x = 0 while Jan 1 is at x = 10. Concatenating the two results in both being "Jan 110".

Join the fields with \u0000 instead of simple concatenation to prevent the collision.


src/plots/cartesian/axes.js
```diff
@@ -1604,7 +1604,7 @@ axes.doTicks = function(gd, axid, skipTitle) {
     var axletter = axid.charAt(0),
         counterLetter = axes.counterLetter(axid),
         vals = axes.calcTicks(ax),
-        datafn = function(d) { return d.text + d.x + ax.mirror; },
+        datafn = function(d) { return [d.text, d.x, ax.mirror].join('\u0000'); },
```
etpinard: Nice fix :tada: 

I'm curious to hear why you chose `\u0000` instead of say `-` here?

rpaskowitz: We used \u0000 as a workaround but putting it on the end of our ticktext values, but not causing any visible change. No solid reason we couldn't use something else here.

etpinard: Ok. In this case, I would prefer changing the `join` argument to `-` if you don't mind.

etpinard:
@rpaskowitz It would be nice to add an image test for :lock:ing down this behavior. To do so, add an image test mock data/layout JSON in the [`test/image/mocks/`](https://github.com/plotly/plotly.js/tree/master/test/image/mocks) as generated in https://jsfiddle.net/hgqhrnrj/5/ If you're feeling adventurous, you can try to generate the baseline yourself in our image test docker container (docs [here](https://github.com/plotly/plotly.js/tree/master/test/image)).

src/plots/cartesian/axes.js
rpaskowitz: Updated

rpaskowitz:
Added a mock and tested it pre/post change and confirmed it's working as intended.

I don't think I'll be able to get to generating the baseline however. Any recommendation on adding that in a separate PR, sending me a baseline and my adding it to this PR, other...?

src/plots/cartesian/axes.js
alexcjohnson: Actually, perhaps a character that's not possible in a stringified number, like `'_'`? It would be rather pathological, but Imagine you had
```js
xaxis: {
    type: 'linear',
    title: 'Grade',
    tickvals: [-1, 1],
    ticktext: ['A', 'A-']
}
```

(Nice debugging BTW @rpaskowitz !)

etpinard: Good call @alexcjohnson 

image

:open_mouth:

etpinard:
> I don't think I'll be able to get to generating the baseline however

No worries will take it from here once https://github.com/plotly/plotly.js/pull/1337#discussion_r98594383 is addressed. Thanks again for this awesome PR!

src/plots/cartesian/axes.js
rpaskowitz: image

Will change.

etpinard:
@rpaskowitz I added the baseline image in https://github.com/plotly/plotly.js/pull/1347 which includes your commit. I will close this PR and merge https://github.com/plotly/plotly.js/pull/1347

Thanks again :beers:

Issue #144, opened 2015-12-24
Styling a single marker


aderidder:
Hi,

In a community post I recently enquired whether, in order to change the colour of a single marker in a 10000 marker trace, I'd have to provide the `plotly.restyle` with a complete array with colours, or whether it's possible to update just a single marker. 
Etienne's answer is that at the moment `plotly.restyle` can only be used to restyle a complete trace.

The feature I'm requesting is a more fine-grained control, e.g. the possibility to change the colour of a single marker without having to style the complete trace. 

Thanks,
Sander

etpinard:
@aderidder Thanks for the suggestion. 

I like this idea, but I'm not 100% sure if it belongs in `Plotly.restyle`.

My main concern is coming up with an API on top of the current `Plotly.restyle` API that is intuitive, general and scalable.

The current API supports:

```js
// restyle trace 1 in red
// restyle trace 2 in blue
Plotly.restyle('graph', 'marker.color', ['red', 'blue'], [1, 2]);

// restyle trace 1 pts in red and blue in alternate manner
// restyle trace 2 pts in green and yellow " " "
Plotly.restyle('graph', 'marker.color', [['red', 'blue'], ['green', 'yellow']], [1, 2]);
```

Per-pt restyle could then look like:

```js
// restyle pt 20 of trace 1 in red & pt 21 of trace 1 in blue (and nothing else)
// restyle pt 20 of trace 2 in green & pt 21 of trace 2 in yellow (" " ")
Plotly.restyle('graph', 'marker.color', [['red', 'blue'], ['green', 'yellow']], [1, 2], [20, 21]);
```
The last argument `[20, 21]` will be used to restyle both trace 1 and trace 2. 

**Problem** there would no way to e.g. restyle pts 20 and 21 in trace 1 and 10 and 12 in trace 2 in one `Plotly.restyle` call.

Alternatively, last argument could become a 2d array:

```js
// restyle pt 20 of trace 1 in red & pt 21 of trace 1 in blue (and nothing else)
// restyle pt 10 of trace 2 in green & pt 12 of trace 2 in yellow (" " ")
Plotly.restyle('graph', 'marker.color', [['red', 'blue'], ['green', 'yellow']], [1, 2], [[20, 21], [10, 12]]);
```

but, that might be too confusing / non-intuitive. 

@cldougl @mdtusz thoughts?

alexcjohnson:
[...]

PR #1338, opened 2017-01-27
Contour auto persist

alexcjohnson:
Fixes a bug I found when playing around with contours: If you make a contour plot without specifying anything about contours (so you get the full `autocontour` behavior even though it's not explicitly stated) then you change the `z` range, it would not update the contour range. You can do this in straight plotly.js with `Plotly.restyle`, or in the workspace by making a contour plot and then modifying the data grid. Also ensures the correct level of replotting when you modify any part of an array, ie if `z` triggers a recalc, so does `z[1]`, `z[2][3]`, etc.

Related to #1282 but for now I don't want to go all the way into that, I just want to fix the bug.

cc @etpinard

### DIFF
--- a/test/jasmine/tests/plot_api_test.js
+++ b/test/jasmine/tests/plot_api_test.js
@@ -283,6 +283,28 @@ describe('Test plot api', function() {
             expect(gd._fullData[1].marker.color).toBe(colorDflt[1]);
         });
 
+        it('should redo auto z/contour when editing z array', function() {
+            var gd = {
+                data: [{type: 'contour', z: [[1, 2], [3, 4]]}]
+            };
+
+            mockDefaultsAndCalc(gd);
+
+            expect(gd.data[0].zauto).toBe(true);
+            expect(gd.data[0].zmin).toBe(1);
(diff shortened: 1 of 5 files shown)


test/jasmine/tests/plot_api_test.js
```diff
@@ -283,6 +283,28 @@ describe('Test plot api', function() {
...
+            mockDefaultsAndCalc(gd);
+
+            expect(gd.data[0].zauto).toBe(true);
+            expect(gd.data[0].zmin).toBe(1);
+            expect(gd.data[0].zmax).toBe(4);
+
+            expect(gd.data[0].autocontour).toBe(true);
+            expect(gd.data[0].contours).toEqual({start: 0.5, end: 3.5, size: 0.5});
+
+            Plotly.restyle(gd, {'z[0][0]': 10});
```
etpinard: Really. This works? That's awesome!

etpinard: This makes https://github.com/plotly/plotly.js/issues/144 obsolete.

etpinard:
:dancer: lgtm

Issue #1341, opened 2017-01-28
afterplot event not fired on load


tjmcewan:
The `afterplot` event isn't fired until the chart is interacted with in some way.  It seems like this isn't the intention?  I'd like to be able to hook to an event that is fired when the chart is finished loading for the first time.

Reproduction:

1. Go here: http://codepen.io/plotly/pen/QbZmZY
2. Open your console
3. Reload the page and observe that no event is fired.

If this is not the intention, do you have a recommended workaround?

Thanks. :)

etpinard:
Thanks for the report.

Issue #76, opened 2015-12-04
A few more things to clean up in the promise land


alexcjohnson:
- [ ] Check that we're doing things as synchronously as possible (see [comment on 70](https://github.com/plotly/plotly.js/pull/70/files#r46615093) )
- [x] Make sure _all_ plot_api calls return promises. Looks like at present `plot`, `newPlot`, `redraw`, `restyle`, and `relayout` do, but not `extendTraces`, `moveTraces` etc.

alexcjohnson:
From a separate convo

@etpinard 
> Plotly.plot's promise story is still a bit of a mystery to me
> One thing I do know: we can't test `Plotly.plot` in jasmine without handling the `then()` regardless of the plot type

@alexcjohnson
> huh, OK, it seems weird to me. We do always return a promise, but everything should be synchronous up to that point in the normal case, so by the time the function exits it should be complete.
> might be interesting to watch [`donePlotting`](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L358)
and see if we are really making everything async for some dumb reason...

It's not the end of the world if plotting is less synchronous than it could be, but the experience is better for the viewer (at the very least fewer flashes of half-built plots, and likely faster display overall) if it's sync. For the developer using plotly.js it's probably no different, they should get used to expecting `Plotly.plot` and `Plotly.newPlot` (and everything else in `plot_api` but there's some more for us to clean up there) to return a promise.

mdtusz:
@etpinard @alexcjohnson I see that the promise resolve isn't passed an argument - should we pass in the gd? That's what I would expect to be handed when using the library, and is trivial to implement because the gd is always in scope when Promise.resolve is called.

etpinard:
@mdtusz I vote for passing the `gd`

alexcjohnson:
@mdtusz sure, if you think it would be useful. Generally I don't see why it would, as all these functions take `gd` as an arg so the caller already has it. But I guess there could be cases, like if you bundle a bunch of plot promises together...

mdtusz:
@alexcjohnson my main use-case is if someone does something like this:

``` javascript
// In some place, dynamically adding plot(s)
var myPlot = Plotly.plot('someDiv', data, layout);
var plots = [myPlot];


// Somewhere else in the code, maybe another file
plots.forEach(function(plot){
    plot.then(function(gd){
         gd.className = "visible";
    });
});
```

[...]

PR #1342, opened 2017-01-30
Fire plotly_afterplot after Promise.all callback

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1341

### DIFF
--- a/test/jasmine/tests/plot_api_test.js
+++ b/test/jasmine/tests/plot_api_test.js
@@ -84,6 +84,21 @@ describe('Test plot api', function() {
                 expect(gd._transitionData._frames[2].name).toEqual('frame3');
             }).catch(fail).then(done);
         });
+
+        it('should emit afterplot event after plotting is done', function(done) {
+            var afterPlot = false;
+
+            var promise = Plotly.plot(gd, [{ y: [2, 1, 2]}]);
+
+            gd.on('plotly_afterplot', function() {
+                afterPlot = true;
+            });
+
+            promise.then(function() {
+                expect(afterPlot).toBe(true);
+            })
+            .then(done);
+        });
     });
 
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/plot_api_test.js
```diff
@@ -84,6 +84,21 @@ describe('Test plot api', function() {
                 expect(gd._transitionData._frames[2].name).toEqual('frame3');
             }).catch(fail).then(done);
         });
+
+        it('should emit afterplot event after plotting is done', function(done) {
```
etpinard: This test here fails on `master`.

I'm not exactly sure why though which might me related to https://github.com/plotly/plotly.js/issues/76

Consider:

```js

Plotly.plot(gd, [{}])

gd.on('plotly_afterplot', () => {})
```

I suspect that on `master` the `Plotly.plot` promise is resolved before `gd.on('plotly_relayout, /* */)` is set. By tucking `gd.emit('plotly_afterplot')` is the `Promise.all` callback the `Plotly.plot` promise is resolved after `gd.on`.

alexcjohnson: > I suspect that on master the Plotly.plot promise is resolved before gd.on('plotly_afterplot, /* */) is set.

exactly - in most cases the plotting itself is sync, so your fix here is spot on.

alexcjohnson:
💃

alexander-daniel:
Just wondering, is this only fired on `Plotly.plot` calls?

Curious, as I'd love to use this event with the React component, but that seems to use either:
`Plotly.newPlot`  &
`Plotly.react`

Thanks!

etpinard:
Right, `plotly_afterplot` should be fired when `Plotly.newPlot` completes. Not sure why that doesn't happen.

Issue #1343, opened 2017-01-30
cannot call newPlot with missing data for financial charts


VeraZab:
Should be able to call newPlot on financial traces, even if user hasn't supplied all data yet.
Instead getting 'Uncaught TypeError: Cannot read property 'val' of undefined'
http://codepen.io/veraz/pen/GrygMq?editors=1010

I believe it is because financial traces have a rangeslider, which needs a range in order to be plotted, and the range requires x data to be present...

Also, gd.layout.xaxis.type gets set to 'category', which should not be the case as its not a valid type for financial charts. It should not get set at that point, because the user hasn't entered any data yet, and data type can be linear or date, so we don't know yet.

PR #1348, opened 2017-01-31
Don't try adding rangeslider to visible: false finance traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1343

@VeraZab easy.


alexcjohnson:
💃

Issue #1329, opened 2017-01-24
financial charts: can't edit trace names from legend while {editable: true}


VeraZab:
http://codepen.io/veraz/pen/wgqwKb

etpinard:
Weird. There's a [block](https://github.com/plotly/plotly.js/blob/c9de6c80aca353429a8f153901b36244ca8a955a/src/components/legend/draw.js#L379-L389) in the legend component that should do just that.

PR #1020, opened 2016-10-07
OHLC and Candlestick trace types

etpinard:
This PR adds two new trace types the plotly.js: `ohlc` (short for [open-high-low-close](https://en.wikipedia.org/wiki/Open-high-low-close_chart)) and [`candlestick`](https://en.wikipedia.org/wiki/Candlestick_chart) to the main plotly.js bundle - making [plotlyjs-finance](https://github.com/etpinard/plotlyjs-finance) and the plotly.py figure factories obsolete.

This PR also adds a new partial bundle called `plotly-finance` which includes `ohlc` and `candlestick` along with a few other trace type. We should come to an agreement about which trace type to include before this PR is done.

On the technical side of things, this PR introduces a new way of adding trace types: using transforms. In essence, a `ohlc` trace is transformed internally into one increasing and  one decreasing line trace. Similarly, a `candlestick` trace is transformed transformed internally into one increasing and one decreasing box trace.

PR #1349, opened 2017-02-01
Finance trace editable legend fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1329

In https://github.com/plotly/plotly.js/pull/1020 (more precisely in https://github.com/plotly/plotly.js/pull/1020/commits/fbb299b145d0ed1c16e4802a4a25120938ff4e16) a custom finance trace `editable: true` legend item handler was added to ensure that editing legend text would propagate to the correct finance trace. Unfortunately, the relevant attribute name changed since then, breaking the behavior. Testing this particular `editable: true` isn't trivial hence why I left it out of #1020. 

In this PR, I make sure to :lock: this behavior down.

### DIFF
--- a/test/jasmine/tests/finance_test.js
+++ b/test/jasmine/tests/finance_test.js
@@ -963,3 +963,79 @@ describe('finance charts updates:', function() {
         });
     });
 });
+
+describe('finance charts *special* handlers:', function() {
+
+    afterEach(destroyGraphDiv);
+
+    it('`editable: true` handles should work', function(done) {
+
+        function editText(itemNumber, newText) {
+            var textNode = d3.selectAll('text.legendtext')
+                .filter(function(_, i) { return i === itemNumber; }).node();
+            textNode.dispatchEvent(new window.MouseEvent('click'));
+
+            var editNode = d3.select('.plugin-editable.editable').node();
+            editNode.dispatchEvent(new window.FocusEvent('focus'));
+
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/finance_test.js
```diff
@@ -963,3 +963,79 @@ describe('finance charts updates:', function() {
...
+        function editText(itemNumber, newText) {
+            var textNode = d3.selectAll('text.legendtext')
+                .filter(function(_, i) { return i === itemNumber; }).node();
+            textNode.dispatchEvent(new window.MouseEvent('click'));
+
+            var editNode = d3.select('.plugin-editable.editable').node();
+            editNode.dispatchEvent(new window.FocusEvent('focus'));
+
+            editNode.textContent = newText;
+            editNode.dispatchEvent(new window.FocusEvent('focus'));
```
etpinard: I learned something new -> https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent

etpinard: which is used to invoke the [`makeEditable`](https://github.com/plotly/plotly.js/blob/e7f69c986ba11e6672b0ff55bc76389c934d64cb/src/lib/svg_text_utils.js#L436-L535) handlers.

alexcjohnson:
Gotta love the 1-line fix with 76 lines of tests! 💃

Issue #1250, opened 2016-12-13
Rangeslider only shows for the first x-axis in multiple subplots


hy9be:
For example in case here: http://codepen.io/plotly/pen/648434b7b06688fc6b00067592d616b0.

If I add ``rangeslider: {}`` to ``xaxis2``, no slider will be rendered.
If I add ``rangeslider: {}`` to both ``xaxis`` and ``xaxis2``, only the first x-axis will have the slider.

It seems the range slider will not work for some of the axes in subplots.

hy9be:
@etpinard Is this a bug or an expected behavior. I read the spec for rangeslider here: https://github.com/plotly/plotly.js/issues/279, and did not find a clue.

Thanks!

etpinard:
It is the expected behavior (for now). Would be a nice addition though.

hy9be:
Thanks!. I would like to try to make a PR for this.

I see that there is a flag named ``isMainPlot`` in the drawing function for rangeslider: https://github.com/plotly/plotly.js/blob/master/src/components/rangeslider/draw.js#L334

Not sure if this is key flag for this.

PR #1350, opened 2017-02-01
Range slider data update fix

etpinard:
fixes https://github.com/plotly/streambed/issues/9088

A very subtle change that ensures that new range slider ranges are computed under data array restyle/addTraces/expandTrace/deleteTraces updates.

### DIFF
--- a/src/components/rangeslider/draw.js
+++ b/src/components/rangeslider/draw.js
@@ -82,7 +82,7 @@ module.exports = function(gd) {
         // compute new slider range using axis autorange if necessary
         // copy back range to input range slider container to skip
         // this step in subsequent draw calls
-        if(!opts.range) {
+        if(axisOpts._needsExpand && axisOpts._min.length && axisOpts._max.length) {
             opts._input.range = opts.range = Axes.getAutoRange(axisOpts);
         }
 
--- a/test/jasmine/tests/range_slider_test.js
+++ b/test/jasmine/tests/range_slider_test.js
@@ -536,6 +536,9 @@ describe('the range slider', function() {
 
     describe('in general', function() {
 
+        // lower toBeCloseToArray precision for FF38 on CI
+        var precision = 1e-2;
+
         beforeAll(function() {
             jasmine.addMatchers(customMatchers);
         });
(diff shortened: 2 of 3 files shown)


src/components/rangeslider/draw.js
```diff
@@ -82,7 +82,7 @@ module.exports = function(gd) {
         // compute new slider range using axis autorange if necessary
         // copy back range to input range slider container to skip
         // this step in subsequent draw calls
-        if(!opts.range) {
+        if(axisOpts._needsExpand && axisOpts._min.length && axisOpts._max.length) {
```
etpinard: As the computed range slider range (i.e. `opts.range` here) is copied to the user layer container, `Axes.getAutoRange` would not get called on data array updates (unless one would `delete gd.layout.xaxis.rangeslider.range`).

Note that `xaxis._min` and `xaxis._max` are only defined when an update call passed through the `calc` step as  `setConvert` clears them [here](https://github.com/plotly/plotly.js/blob/e7f69c986ba11e6672b0ff55bc76389c934d64cb/src/plots/cartesian/set_convert.js#L405-L409) in supplyDefaults.

test/jasmine/tests/range_slider_test.js
```diff
@@ -561,6 +568,83 @@ describe('the range slider', function() {
...
+                expect(gd.layout.xaxis.rangeslider.range).toEqual(rangeSliderRange, precision);
+            }
+
+            Plotly.plot(gd, [{
+                y: [2, 1, 2]
+            }], {
+                xaxis: { rangeslider: { range: [-1, 11]} }
+            })
+            .then(function() {
+                assertRange([-0.13, 2.13]);
```
alexcjohnson: A bit confusing, as it looks like this test is identical to the previous one until you squint at the details of `assertRange`. Nonblocking, but I might have made it explicit by saying `assertRanges([-0.13, 2.13], [-1, 11])` or something.

alexcjohnson:
💃

etpinard:
This fix doesn't work for soon-to-be-added multiple range sliders https://github.com/plotly/plotly.js/issues/1250. I'm working on a better solution possible by adding `rangeslider.autorange` boolean attribute.

etpinard:
Dropping from the `1.23.0` milestone.

This bug fix will be part of the multiple range slider #1250 feature PR.

PR #1351, opened 2017-02-02
Vanishing titles

alexcjohnson:
fixes https://github.com/plotly/streambed/issues/8961
The issue turned out to be entering a title again after deleting the default, the hover events for the placeholder text weren't cleared.
@chriddyp you had a comment that it involved annotations as well, but I don't see it, so I'm assuming maybe *that* part actually was https://github.com/plotly/streambed/issues/8423

cc @etpinard

### DIFF
--- a/src/components/titles/index.js
+++ b/src/components/titles/index.js
@@ -18,6 +18,7 @@ var Lib = require('../../lib');
 var Drawing = require('../drawing');
 var Color = require('../color');
 var svgTextUtils = require('../../lib/svg_text_utils');
+var interactConstants = require('../../constants/interactions');
 
 
 var Titles = module.exports = {};
@@ -196,11 +197,11 @@ Titles.draw = function(gd, titleClass, options) {
             .text(txt)
             .on('mouseover.opacity', function() {
                 d3.select(this).transition()
-                    .duration(100).style('opacity', 1);
+                    .duration(interactConstants.SHOW_PLACEHOLDER).style('opacity', 1);
             })
             .on('mouseout.opacity', function() {
                 d3.select(this).transition()
-                    .duration(1000).style('opacity', 0);
+                    .duration(interactConstants.HIDE_PLACEHOLDER).style('opacity', 0);
             });
     }
 
--- a/src/components/titles/index.js
+++ b/src/components/titles/index.js
@@ -192,8 +192,7 @@ Titles.draw = function(gd, titleClass, options) {
         opacity = 0;
         isplaceholder = true;
         txt = placeholderText;
-        fullLayout._infolayer.select('.' + titleClass)
-            .attr({'data-unformatted': txt})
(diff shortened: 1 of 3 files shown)


src/components/titles/index.js
```diff
@@ -192,21 +193,21 @@ Titles.draw = function(gd, titleClass, options) {
...
             .on('mouseout.opacity', function() {
                 d3.select(this).transition()
-                    .duration(1000).style('opacity', 0);
+                    .duration(interactConstants.HIDE_PLACEHOLDER).style('opacity', 0);
             });
     }
 
     if(gd._context.editable) {
         if(!txt) setPlaceholder();
+        else el.on('.opacity', null);
```
alexcjohnson: The fix is just this line. All the rest is cleanup and testing :)

test/jasmine/tests/titles_test.js
```diff
@@ -0,0 +1,103 @@
...
+            expect(+titleEl.style('opacity')).toBe(opacityIn);
+
+            mouseEvent('mouseout', xCenter, yCenter);
+            setTimeout(function() {
+                expect(+titleEl.style('opacity')).toBe(opacityOut);
+                done();
+            }, interactConstants.HIDE_PLACEHOLDER + 50);
+        }, interactConstants.SHOW_PLACEHOLDER + 50);
+
+        return promise;
```
etpinard: Nicely done. :tada: 

If you're up for a challenge, you could try adding a test for title edits like I did [here](https://github.com/plotly/plotly.js/pull/1349/files#diff-7deb9e75965ca15196b29a1f7a4a9210R971) for legend/trace name edits.

etpinard:
:dancer:

test/jasmine/tests/titles_test.js
```diff
@@ -45,6 +46,24 @@ describe('editable titles', function() {
         return promise;
     }
 
+    function editTitle(letter, attr, text) {
+        return new Promise(function(resolve) {
+            gd.once('plotly_relayout', function(eventData) {
+                expect(eventData[attr]).toEqual(text, [letter, attr, eventData]);
+                setTimeout(resolve, 10);
```
alexcjohnson: @etpinard nice editing tests in [finance](https://github.com/plotly/plotly.js/pull/1349/files#diff-7deb9e75965ca15196b29a1f7a4a9210R971)! And good call to include a test like that here too. The only thing I did differently is this: put resolve in a short `setTimeout` instead of manually removing the node. Not sure why `makeEditable` uses a [duration-0 transition](https://github.com/plotly/plotly.js/blob/master/src/lib/svg_text_utils.js#L500) rather than just removing the element synchronously... but this seems to play well with it.

etpinard: Good eyes on 0-second transition :microscope:

Issue #887, opened 2016-08-26
Feature request: Axis boundaries


g-i-o-:
Hi, I just recently started using your plotting library (which is awesome by the way) to do some heatmap plots. One of the main complaints I've received have been the ability to pan outside the plot's boundaries (just click and drag the plot outside the viewport and suddenly the axis range point to some non-data containing area). While this can be mitigated with the `fixedrange` attribute, you end up losing the pan and zoom features for that axis.

So, I was thinking that maybe adding axis boundaries could be a thing? Maybe two more properties could be added to the axes: `bounds` and `boundsmode`. `bounds` gets checked in the panning and zooming logic and, if set, it constrains the range to the specified bounds. `boundsmode` allows the bounds to be computed automatically from the plot (like `autorange`). That way users can zoom and pan al around the plot, but cannot get outside the plot bounds.

As proof of concept I have created this pull request g-i-o-/plotly.js#1 on my fork and was thinking maybe you guys could check it out comment on how to improve and, if you like it, add it or something like that.

---
1. Here we have a nice little heatmap:
   ![newplot](https://cloud.githubusercontent.com/assets/5231862/18020290/ae3232d6-6bae-11e6-867f-a1a49f94a0e1.png)
2. But then some user starts using with the panning tool a bit too much:
   ![newplot 1](https://cloud.githubusercontent.com/assets/5231862/18020304/b8aa7fca-6bae-11e6-9b41-ef38ea6b6f2a.png)

etpinard:
Yes, that would be a great addition to the library.

We already partly have this concept for ternary plots using the axis `min` [attribute](https://github.com/plotly/plotly.js/pull/390/files#diff-c8ed16ffae7f100f28b053863b402b08R50).

So, maybe for consistency, we should add `axis.min` and `axis.max` to cartesian axes with default values being `null` corresponding to no limits. Thoughts?

g-i-o-:
`axis.min` and `axis.max` sounds ok.  Any thoughts on how should `axis.boundsmode` should be called? the idea is that `axis.boundsmode` can be set to `'auto'` and then the `max` and `min` are set using the autorange funcionality.

etpinard:
Why would we need `axis.boundsmode` ?

g-i-o-:
Currently, with `axis.boundsmode` = `'auto'` the bounds are set automatically using the plot's autorange feature.  That way you can just show the data and the plot doesn't allow the user to pan/zoom outside the plot bounds

[...]

Issue #400, opened 2016-04-06
Setting only min or max values for axis and not a range.


shimwell:
Dear all

Many thanks for all your wonderful efforts with Plotly

I just wanted to ask if it might be possible to tweak the range function a little

currently an axis range can be set in the layout like this ...
```
layout = go.Layout(
xaxis=dict(
range=[2, 5]
),
```
in some cases it would be desirable to only specify a min value.
Plotly can then work out the max value automatically.

Or perhaps a user would want to set a max value and have a plotly work out the min value

Either-way it would be nice (for me at least) to only specify one of the range limits and not both. However I can see why this might not be a highly required function and I'm sure there are lots of other things to do first

matplotlib offers this function ...
```
plt.xlim(2,5) #sets a range
plt.gca().set_xlim(xmin=2) # sets only the lower x value and allows computer to work out upper value
plt.gca().set_ylim(ymax=50) # sets only the lower y value and allows computer to work out upper value
```

etpinard:
Thanks for submitting the issue. You're making some valid points.

Kaitensatsuma:
I'd like to add comments to this: it is especially desirable when displaying bar charts that are meant to represent performance against a maximum, otherwise the automatic sizing creates visually misleading results.

since tickmode has an attached option to explicitly name tick points, perhaps adding another option to make them fixed would be a viable feature?

cinkie:
This feature would also be very useful when using subplots. The upper limit can be the same in axis so we could compare more easily.

beeblebrox:
+1.

I am using extendPlots on a larger data set in which I only want to show up to the last two days (xaxis) and using extendTraces. I do not want to limit the points as extendTraces can (it doesn't know dates and it drops points completely).  If range took null for one side I can envision extendTraces auto scrolling the xaxis as points are added always leaving the new points showing.

rneuhausler:
Not sure if people are already doing this, but what's worked for me is using "[]" in place on the value I don't want to specify. Say I only want to set the lower limit to zero, I use **range = [0,[]]**. I'm using plotly in jupyter notebook, written in python 3.

SaturnTeam:
I need this feature too! Sometimes I have only zero values and auto axis sets up to -1 to 1, but negative values are impossible for my data.

etpinard:
> have only zero values and auto axis sets up to -1 to 1

[...]

PR #1353, opened 2017-02-04
Implement Axis boundaries

inDream:
Added two options in layout: `bound` and `autobound` for control the axis boundaries.

@etpinard I would like to know is there a list of functions I need to modify in order to add this feature?
As there're different actions to trigger relayout: click buttons at displayModeBar, scrollZoom at axis, etc.
My draft commit only can handle axis change by drag but not scroll.

Also is there any convenient ways to undo drag / zoom?

Related issues: #887 #400

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1771,7 +1771,8 @@ function _relayout(gd, aobj) {
         doplot: false,
         docalc: false,
         domodebar: false,
-        layoutReplot: false
+        layoutReplot: false,
+        overBounds: false
     };
 
     // copies of the change (and previous values of anything affected)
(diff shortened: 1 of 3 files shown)


etpinard:
Ah I see. Looks like you're trying to add hacks inside `relayout` to make this work. This is the wrong approach I'm afraid. 

You'll need to dive into the [`axes.js`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js) draw functions and limit the visible range there.

Thanks for looking into this :beers:

inDream:
Beside axes auto range, I think prevent drag / zoom also necessary for limiting user only can view area with data. How can I achieve that?

etpinard:
Most of that should be in https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/dragbox.js

etpinard:

I recommend subscribing to https://github.com/plotly/plotly.js/pull/1871 for the latest development of this topic.

Issue #666, opened 2016-06-20
npm run citest-jasmine fails locally


etpinard:
Looks like `npm run citest-jasmine` is calibrated for CircleCI only.

Off 3c85d9e

![image](https://cloud.githubusercontent.com/assets/6675409/16200600/00c468c8-36dc-11e6-85ff-b017380b29e0.png)

I suspect that these issues are related to the browser window size. The failing tests should go away once we add a `window-size` flag in the `karma.ciconf.js` file like we currently do in the (local) [`karma.conf.js`](https://github.com/plotly/plotly.js/blob/3c85d9eb47ce08177d951065fd66a7d9f9671e57/test/jasmine/karma.conf.js#L79).

Moreover, perhaps we should merge the two karma conf files into one and use the CirlcleCI environment [variable](https://circleci.com/docs/environment-variables/#basics) to set up CI specific flags.

alexcjohnson:
Locally I see a default body margin of 8px. I'd rather just set this to 0, then the pixel positions we have to use for mouse events will make more sense anyway. I'll try to address this in my working branch.

etpinard:
@rreusser off `1.14.0`:

```
Firefox 42.0.0 (Mac OS X 10.11.0) mapbox plots should be able to relayout FAILED
    Expected 80,100,904,270 to be close to 80,100,908,270.
```

and

```
Firefox 42.0.0 (Mac OS X 10.11.0) hover on fill should always show one label in the right place FAILED
    Expected 254.27499999999998 to be close to 249.175, -1, 'trace 2:x'.
    assertLabelsCorrect/</<@/var/folders/5s/mm8_jrls0zd34zkb1z71xb940000gp/T/3e5f6eaff748734ec0be007b7921fd04.browserify:152118:17 <- tests/hover_label_test.js:673:0
```

rreusser:
First test: `tests/hover_label_test.js:673:0`
Commit: `b6fc59d`

Issue #444, opened 2016-04-19
Some jasmine tests fail depending on the screen resolution of the computer running the test


monfera:
When running tests in the Default resolution of a 13" MacBook, all of `click_test.js` passed; however, four failures were reported after increasing the resolution to the maximum in System Preferences. After much fiddling on my part, @etpinard suggested it may be to do with the browser window size. Indeed, the browser window was of a different size, depending on the resolution:

- Default resolution: 1035 x 617
- Highest resolution: 792 x 867

Manually resizing the already existing Chrome window and rerunning the test made it pass, confirming the suspicion that failure is due to less than expected viewport width.

mdtusz:
Interesting. Is this because of the viewport width (and plots are being rendered at a different pixel-width?) or because of how OSX/Chrome handles pixel scaling when using the increased resolution mode?

In any case, it would probably be a good idea if we make sure all our tests have explicitly set widths and heights.

monfera:
@mdtusz For this specific issue, it's the former. The devicePixelRatio is the same in both resolutions, the width was different (and Chrome even gave a different aspect ratio for the higher resolution mode). But in general it's true that there are multiple variables to consider, e.g.

- The devicePixelRatio is important for Canvas and WebGL, because, without scaling up with it, the output will be pixelated (aliased). On a multimonitor setup, it can be different per monitor.
- The browser zooming level is another variable. Maybe it's possible for it to get stuck at a non-100% level, failing a bunch of test cases.
- Minimum font size might also impact pixel based comparisons, and potentially even jasmine tests.

I haven't tried to address these other variations in this little PR. 

My $.02 is it's useful to set the screen size in test cases, e.g. to test window size responsiveness. It's also useful to have a basic window size for those type of tests that are not width/height sensitive or haven't (yet?) been sized explicitly. My immediate motivation was to be able to run the test suite without errors so it's but a small step.

PR #1355, opened 2017-02-06
Multiple range sliders

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1250

gifrecord_2017-02-06_153150

This PR also:

- fixes various range-slider+data-update issues as https://github.com/plotly/plotly.js/pull/1350 but in a more robust way, see https://github.com/plotly/plotly.js/commit/cebc31c72f20cd0e1426e8d9258f9e09e269817a for the details.
- add attribute `xaxis.rangeslider.autorange` allowing users to turn on/off range slider autorange computations.


src/components/rangeslider/draw.js
```diff
@@ -92,8 +95,13 @@ module.exports = function(gd) {
...
 
-        var x = margin.l + (graphSize.w * domain[0]),
-            y = fullLayout.height - opts._height - margin.b;
+        var x = margin.l + (graphSize.w * domain[0]);
+
+        var y = (
+            margin.t + graphSize.h * (1 - oppDomain[0]) +
+            tickHeight +
+            opts._offsetShift + constants.extraPad
+        );
```
etpinard: I think I got this right. I would appreciate having something take a second :eye: at it.

alexcjohnson: 👍

alexcjohnson: I guess something in here is responsible for the [small position differences in the mocks](https://github.com/plotly/plotly.js/pull/1355/commits/ffccc75aa40d195d49138fade77a2c00ef89ec6b)? I'm not bothered by the differences, *except* that some of the new ones look a little bit fuzzier than the old ones. Can you try and round it so that the outer edges of the range slider border are exactly on a pixel boundary, even if you have non-integers in `layout.height`, the margins, `rangeslider.borderwidth` and anything else that comes into the calculation?

etpinard: Sure.

etpinard: @alexcjohnson some improvements on:

 https://github.com/plotly/plotly.js/compare/multiple-rangesliders...rangeslider-crisp-round

I think the mocks look better to my :eyes: , would you mind taking a look at them?

alexcjohnson: I like it, thanks! Both the cleaned up edges and the wider grabbers. The one thing that still strikes me as odd is how the corners of the grabbers sometimes get rounded off - one pixel that you'd have expected to be black is missing sometimes. Seems like this violates the spirit of `crispEdges` but oh well. Fiddling in the inspector it seems like if I offset the grabbers by half a pixel in both x and y the corners get drawn correctly.

alexcjohnson: Also, this isn't new but I'm just noticing it now while we're here: should we switch `handleStroke` to be `Color.defaultLineColor = '#444'` instead of `#666` as it is now? And to avoid duplication, `handleFill = '#fff'` is already available as `Color.background` - to explicitly match how we define [zoombox corners](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/dragbox.js#L202-L203)

etpinard: Good call 👍

etpinard: https://github.com/plotly/plotly.js/commit/5a649f03339c155893eeec882b0bd15ffe0e5968 gives

image

is that what you had in mind?

alexcjohnson: The image you've pasted here is blown up (with browser zoom?), so the rounding isn't quite right and consequently there are still some odd 1-pixel effects at the corners. When I look at the rangeslider-crisp-round branch locally it looks *almost* perfect.
screen shot 2017-02-22 at 10 22 28 pm

I realize this is a borderline obsessive ask (which side of the border?) so I did it myself in https://github.com/plotly/plotly.js/commit/29a31d5a0fc7bea0f1b6a7831358ca7a224202e1 and https://github.com/plotly/plotly.js/commit/b49ea77012de7fbb6f80f1af38cd37d389e53433. I just tweaked it a tiny bit to get the symmetry right: both handles the slightest bit more on the dark side. I suppose we could bump the width either up or down 1px so it would be possible to have them each exactly symmetric across the dark/light line, but to my eye anyway symmetry between the two handles is more important. Also included rounding of everything that goes into the grabber position before offsetting, because sometimes we were still getting a non-integer starting point, making the half-pixel offset moot. Kind of annoying but I think now it's really truly guaranteed to be an honest rectangle.

screen shot 2017-02-22 at 10 32 16 pm

Seem reasonable?

alexcjohnson: Aaand in case you weren't 100% convinced I'm being obsessive - I've got a retina display on my laptop, and a non-retina external monitor. The retina display still rounds off all the corners, since it doubles the real number of pixels displayed. I'd say this looks fine, perhaps even better than the perfect sharp corners on a non-retina display - because it's a consistent look and the missing pixels are smaller than the line width so the line really looks rounded, not broken like it did on regular monitors before. Screen shot blown up so you can see the individual retina pixels:
screen shot 2017-02-22 at 11 13 49 pm

alexcjohnson:
💃

Issue #1310, opened 2017-01-18
Two 'bar' traces are not showing with empty data


taranovalexander:
Hi! I try to create two time series with type 'bar'.
One trace didn't contain any data, but this one should be shown in legend. To make this trace visible in legend, I have set x: [null] and y: [null];
But in this case both series are not displaying.
[https://codepen.io/sandertar/pen/VPmOdK](https://codepen.io/sandertar/pen/VPmOdK)
Is there any other way to make label visible in legend for trace without data?

etpinard:
Thanks for writing. Your codepen points a few separate issues. 

First, https://codepen.io/etpinard/pen/xggwYq is probably the best you can do with the current version. You might not like that the `trace 1` legend item is semi-transparent, our apologies :smiley: 

Looks like there's a bug with how the `null` trace interacts with other the trace to compute the bar widths. This bug should be easy to fix.

Moreover, I believe that something like

```js
trace = {
  visible: true,
  x: [],
  y: []
}
```

where the user explicitly set `visible: true` should appear in the legend. At the moment, traces with `x.length === 0` and `y.length === 0` are taken to be `visible: false` no matter what the users inputs. Modifying this logic would be a much longer and harder project; I can't guarantee it's going to happen.

taranovalexander:
Ok, thanks

PR #1004, opened 2016-10-04
Multiple scattergeo improvement

etpinard:
fixes #963

and supersedes https://github.com/plotly/plotly.js/pull/964 

In brief, this PR brings several ideas from `scattermapbox` over to `scattergeo` - while reusing some parts of logic (namely regarding GeoJSON) keeping things :palm_tree: 

In details, this PR:

- fixes #963 (thanks again @cpsievert for letting me know)
- scattergeo traces with non-circle markers now get the correct legend item
- make `scattergeo` use a proper calc step
- make `scattergeo` use `Fx.hover` to pick (using calcdata) and draw its hover labels
  + this means that user no longer have to hover **over** scattergeo data points to trigger hover event / draw labels
  + this means that scattergeo `mode: 'lines'` now (I mean finally) has a hover handlers!
- adds support for `connecgaps` to scattergeo
- adds support for `fill: 'toself'` to scattergeo (similar to how `scattermapbox` implement `fill`)

-------------

:tada: Yay geo line hover text:

gifrecord_2016-10-03_180302

PR #1369, opened 2017-02-09
Skip over placeholder calc traces in bar set positions

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1310

In bar calc, bars with non-numeric positions are not included in the resulting `gd.calcdata` ([ref](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/traces/bar/calc.js#L61-L63)). Later in `Plots.doCalcdata`, those skipped bars are given a placeholder calcdata item ([ref](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/plots/plots.js#L2046-L2048)) to ensure that all calc traces have as many items their corresponding fullData traces (see https://github.com/plotly/plotly.js/pull/1004/commits/dd05c6ee857ff5e5e537ccf946d715e60f9924e2 for more on `placeholder` calcdata items). 

These placeholder calc items contain `x: false, y: false`, which throws off the bar min-difference [here](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/traces/bar/sieve.js#L42); hence the :bug: . In details, `false` is coerced to `0` in [`Lib.distinctVals`](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/lib/search.js#L79) so essentially

```js
data = [{
  x: [1,2,3],
  y: [2,1,2]
}, {
  x: [null],
  y: [/* ... */]
}]
```

was treated essentially the same way as

```js
data = [{
  x: [1,2,3],
  y: [2,1,2]
}, {
  x: [0],
  y: [/* ... */]
}]
```

which leads to large errors especially when `data[0]` has date positions (which are stored in datetimes at the set position step) as reported in #1310.

In brief, the fix is easy, simply skip over `placeholder` calc traces during bar set positions.

### DIFF
--- a/src/traces/bar/set_positions.js
+++ b/src/traces/bar/set_positions.js
@@ -37,7 +37,8 @@ module.exports = function setPositions(gd, plotinfo) {
             fullTrace.visible === true &&
             Registry.traceIs(fullTrace, 'bar') &&
             fullTrace.xaxis === xa._id &&
-            fullTrace.yaxis === ya._id
+            fullTrace.yaxis === ya._id &&
+            !calcTraces[i][0].placeholder
         ) {
             if(fullTrace.orientation === 'h') {
                 calcTracesHorizontal.push(calcTraces[i]);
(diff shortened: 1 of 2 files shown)


alexcjohnson:
💃

Issue #1251, opened 2016-12-14
Bar / histogram traces break on single-ended (zoom) axis drag


alexcjohnson:
- Make a bar or histogram plot. To see it even more clearly, use a plot with bars *and* something else like scatter - for example http://localhost:3000/devtools/test_dashboard/#bar_line
- Position your mouse over an axis end or corner dragger - anywhere the mouse cursor becomes a single-ended arrow
- Drag. The scatter or other traces stay scaled correctly as the axis/axes zoom. The bar trace(s) don't, they just pan.

rpaskowitz:
The issue seems to be slightly more subtle (or has changed since originally reported)
- If you are dragging from the low end of the X axis or the high end of the Y axis, then the bars appear to pan, but moving far further than they aught to, **but**
- If you are dragging from the high end of the X axis, or the low end of the Y axis, then the bars do not move until a click-up event, at which point they redraw.

Fairly easy to see on these mocks as well, with the rendering being corrected once the users stopps the drag.
http://localhost:3000/devtools/test_dashboard/#basic_bar
http://localhost:3000/devtools/test_dashboard/#stacked_bar
http://localhost:3000/devtools/test_dashboard/#grouped_bar

etpinard:
This is a side effect of @rreusser https://github.com/plotly/plotly.js/pull/762 where

```js
subplot.plot
  .selectAll('.points').selectAll('.point')
```

(unfortunately) also selects [`<g points>`](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/traces/bar/plot.js#L45) nodes in bar traces. Easy fix coming up.

PR #1370, opened 2017-02-10
Don't unscale bar layer points on drag

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1251

as mentioned in https://github.com/plotly/plotly.js/issues/1251#issuecomment-278976729 the fix was pretty easy, writing a test for it was a little harder though.

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -712,7 +712,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 // This is specifically directed at scatter traces, applying an inverse
                 // scale to individual points to counteract the scale of the trace
                 // as a whole:
-                .selectAll('.points').selectAll('.point')
+                .select('.scatterlayer').selectAll('.points').selectAll('.point')
                     .call(Drawing.setPointGroupScale, 1 / xScaleFactor, 1 / yScaleFactor);
         }
     }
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/click_test.js
```diff
@@ -809,3 +810,48 @@ describe('Test click interactions:', function() {
...
+                assertScale(d3.select('.plot').node(), 1.14, 1);
+
+                d3.select('.scatterlayer').selectAll('.point').each(function() {
+                    assertScale(this, 0.87, 1);
+                });
+                d3.select('.barlayer').selectAll('.point').each(function() {
+                    assertScale(this, 1, 1);
+                });
+
+                mouseEvent('mouseup', pos[0] + 50, pos[1]);
```
etpinard: Fun fact. This test case doesn't need a `mouseup` event to test what is desired. But, we need to trigger one to not affect other tests downstream (namely in `config_test.js`).

src/plots/cartesian/dragbox.js
```diff
@@ -712,7 +712,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 // This is specifically directed at scatter traces, applying an inverse
                 // scale to individual points to counteract the scale of the trace
                 // as a whole:
-                .selectAll('.points').selectAll('.point')
+                .select('.scatterlayer').selectAll('.points').selectAll('.point')
```
etpinard: @rreusser would you mind reviewing the line?

rreusser: That makes sense to me. It's certainly better, though I can't say from direct knowledge that nothing (questionably) puts stray items in the scatterlayer.

PR #1373, opened 2017-02-13
bower best practice: Updating bower.json to reference non-minified file

willbastian:
The minified version of plotly.min.js results in grunt-contrib-uglify
hanging when it hits the minified plotly.min.js
(gruntjs/grunt-contrib-uglify#233)

Best practices captured at bower/bower#393
specify that the main bower file should point to unminified files.


etpinard:
Thanks very much :beers:

PR #1371, opened 2017-02-10
[EXPERIMENTAL] Run standard --fix

rreusser:
This is a test PR, mainly to get the ball rolling and see what *would* happen if we simply applied [standard](https://github.com/feross/standard) formatting. There are lots of rules that will require manual rewriting.

In particular, the vast majority look like

```
convert.js:219:72: ':' should be placed at the beginning of the line.
index.js:34:3: Split initialized 'var' declarations into multiple statements.
```

It's probably a couple hour pass to go through and fix those, then a bit more work to figure out maybe a few more subtle issues.

See: #950 ping @etpinard

PR #1374, opened 2017-02-13
[EXPERIMENTAL] Run prettier --write

rreusser:
This PR is an alternative to #1371 using `prettier`, which seems to be an alternative to eslint that people are pretty excited about. As opposed to eslint, prettier does not 'fix' any code or throw errors. Instead, it just 100% rewrites it from the AST so that you have no choice over formatting. Which is kind of the point of enforcing formatting in the first place.

Just submitting for the sake of discussion. Presumably this would be a precommit/prepublish hook instead of a yes/no pass/fail step.


rreusser:
`prettier`[lists itself] as beta and notes that it works fine but may undergo a few tweaks before it's fully stable. I saw someone from facebook saying they'd be fine using it in production if just a few more PRs were reviewed/merged.

rreusser:
Closing this. [prettier-standard-formatter](https://github.com/dtinth/prettier-standard-formatter) looks interesting, but seems maybe best to wait just a moment for prettier to stabilize.

Issue #1242, opened 2016-12-11
Official version of plotly.js via Packagist (Composer - PHP)


marclaporte:
Packagist is the main Composer repository. It aggregates public PHP packages installable with Composer: 
https://getcomposer.org/
https://packagist.org/

It makes it easier to install and to keep up to date. If a developer is picking a library (vs another), package management is a criteria. More inclusions in apps means the library is used more, and can lead to more contributions (bug reports of course, but also hopefully patches and new features)

I know plotly.js is not in PHP, but it's common to have packages that have JavaScript but no PHP. This represents millions of downloads. Some notable ones:
https://packagist.org/packages/twbs/bootstrap
https://packagist.org/packages/components/jquery
https://packagist.org/packages/components/jqueryui
https://packagist.org/packages/blueimp/jquery-file-upload
https://packagist.org/packages/ckeditor/ckeditor
https://packagist.org/packages/mbostock/d3
https://packagist.org/packages/tinymce/tinymce
https://packagist.org/packages/moment/moment

Tiki gets dozens of them:
http://sourceforge.net/p/tikiwiki/code/HEAD/tree/trunk/composer.json

Thanks!

etpinard:
Thanks for the headsup.

If you can get 50 :+1: s, we'll spend the time to do this.

fabiomontefuscolo:
Hello @etpinard 

I want to do a pull request with `composer.json`, but after read *PULL_REQUEST_TEMPLATE.MD*, I think it is better talk to you first.

I forked your project and added the file needed by packagist. The URL is https://github.com/fabiomontefuscolo/plotly.js/blob/master/composer.json.

If you accept, there are some other steps to follow:

1. Create an account in packagist.org
2. Submit your package (just ctrl+c/ctrl+v repository url)
3. Connect github webhook to packagist

Unfortunately, I can't do these steps for you, but I think it takes no more than 10 minutes to setup.

The package name in `composer.json` is *plotly/plotly.js*. This means your account name in packagist should be plotly. If you want other account name, you have to change the before slash, like *othername/plotly.js".

Hope this help,

etpinard:
@fabiomontefuscolo Thanks for looking into this!

@jackparmer, @chriddyp or @bpostlethwaite would you mind creating a plotly packagist account?

chriddyp:
Sure! It looks like `composer.json` needs to be part of the repository before we can submit it though:

[...]

PR #1375, opened 2017-02-13
create composer.json file to add plotly in packagist

fabiomontefuscolo:
This file is required to add plotly.js is packagist and solves issue #1242.


etpinard:
Thanks!

PR #1376, opened 2017-02-13
Extend input relayout & restyle objects before looping over key-vals

etpinard:
fixes https://github.com/plotly/streambed/issues/9172

In a few cases, given:

```js
Plotly.relayout(gd, updateObj)

// or
Plotly.update(gd, {}, updateObj)
```

the input `updateObj` is mutated by the relayout logic while converting aliases for example [here](https://github.com/plotly/plotly.js/blob/8269ea00555fb90e7a95657e183ca4c10477e9d7/src/plot_api/plot_api.js#L1958) when relayout'ing annotations.

This bug became very problematic for update menu buttons relayout annotations ([example](http://codepen.io/etpinard/pen/EZrjqy?)) where it caused the buttons to stop working after repeated clicks.

Note that, the restyle object - i.e. as in `Plotly.restyle(gd, restyleObject, traces)` - are never mutated. In this PR, I chose to extend them anyway in case we do start mutating them down the road.


alexcjohnson:
Good call 💃

PR #1379, opened 2017-02-14
Pass customdata through to scatter calcdata

rreusser:
This PR adds a `customdata` attribute to scatter traces that is passed through and assigned to the `data` property of the corresponding calcdata entry.

To do: verify the manner in which this is (or is not?) actually passed all the way through to the dom element.

ping @alexcjohnson @etpinard

### DIFF
--- a/src/traces/scatter/style.js
+++ b/src/traces/scatter/style.js
@@ -25,13 +25,13 @@ module.exports = function style(gd) {
     s.selectAll('g.points')
         .each(function(d) {
             var el = d3.select(this);
-            var pt = el.selectAll('path.point');
+            var pts = el.selectAll('path.point');
             var trace = d.trace || d[0].trace;
 
-            pt.call(Drawing.pointStyle, trace);
+            pts.call(Drawing.pointStyle, trace);
 
             if(trace.customdata) {
-                pt.each(function(cd) {
+                pts.each(function(cd) {
                     d3.select(this).classed('plotly-customdata', cd.data !== null && cd.data !== undefined);
                 });
             }
--- a/src/traces/scatter/style.js
+++ b/src/traces/scatter/style.js
@@ -30,12 +30,6 @@ module.exports = function style(gd) {
 
             pts.call(Drawing.pointStyle, trace);
 
-            if(trace.customdata) {
-                pts.each(function(cd) {
-                    d3.select(this).classed('plotly-customdata', cd.data !== null && cd.data !== undefined);
-                });
-            }
-
             el.selectAll('text')
                 .call(Drawing.textPointStyle, trace);
         });
--- a/src/traces/scatter/style.js
+++ b/src/traces/scatter/style.js
@@ -26,15 +26,18 @@ module.exports = function style(gd) {
         .each(function(d) {
             var el = d3.select(this);
             var pt = el.selectAll('path.point');
+            var trace = d.trace || d[0].trace;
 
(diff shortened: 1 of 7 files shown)


alexcjohnson:
Looks reasonable to me. Do you also want to:
- [ ] add a class like `plotly-customdata` to the elements that actually have `d.data` (I think [scatter/style.js](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/style.js#L28) is the right place (do we need to call out `text` or just `path.point`?) so the API doesn't need to know anything about our selectors to find the right elements to look for `d.data` on.
- [ ] explicitly test passing through an object, not just a string, as that's probably what we'll need in practice.

rreusser:
I've added a `plotly-customdata` class that's set if `data` is anything except `undefined` or `null`. I could probably avoid the loop if `customdata` wasn't even provided. It didn't feel like a nice fit in the scatter tests, but I didn't really want to unit-test an intermediate API for this since it certainly doesn't need to be implemented in the `style` pass.

alexcjohnson:
Nice. Yes, I think it'd be worthwhile to avoid the loop in the (overwhelmingly common) case of no `customdata`.

test/jasmine/tests/scatter_test.js
```diff
@@ -326,3 +332,32 @@ describe('Test scatter', function() {
...
+        }]).then(function() {
+            var points = d3.selectAll('g.scatterlayer').selectAll('.point');
+
+            // Rather than just duplicating the logic, let's be explicit about
+            // what's expected. Specifially, only null and undefined (the default)
+            // do *not* add the class.
+            var expected = [false, false, true, true, true, true, false];
+
+            points.each(function(cd, i) {
+                expect(d3.select(this).classed('plotly-customdata')).toBe(expected[i]);
```
alexcjohnson: Beautiful test. I agree, end-to-end is the right way to test this particular beast.

src/traces/scatter/style.js
```diff
@@ -24,9 +24,16 @@ module.exports = function style(gd) {
...
-            d3.select(this).selectAll('path.point')
-                .call(Drawing.pointStyle, d.trace || d[0].trace);
-            d3.select(this).selectAll('text')
+            var el = d3.select(this);
+            var pt = el.selectAll('path.point');
+
+            pt.call(Drawing.pointStyle, d.trace || d[0].trace);
+
+            pt.each(function(cd) {
+                d3.select(this).classed('plotly-customdata', cd.data !== null && cd.data !== undefined);
```
alexcjohnson: Just a little style point: `pt` makes it sounds like it refers to a single point, but it's all points so I'd change it to `pts`. And elsewhere `cd` refers to a whole `calcdata` array so this tripped me up for a sec - maybe change it to `pointData` or something.

rreusser:
Caveat I just can't ignore: animation. Going to add that to the test to ensure these are properly updated. Since transitions don't work through the style pass, an animation update won't update these. Fixing.

rreusser:
Okay, I've updated the PR (in a test-driven fail-first manner that actually tests what it claims to 👍). It unsets customdata via an animation that doesn't replot in order to verify that it's actually set correctly. This required moving it to `plot.js` which is pretty straightforward and doesn't really affect anything, to the best of my knowledge. I'm currently content with this.

alexcjohnson:
Very nice. 💃

PR #1389, opened 2017-02-16
Allow controlling pan/zoom range entry.

rpaskowitz:
For either UI appearance reasons, or because the range entry doesn't work as one might expect (for example, on http://localhost:3000/devtools/test_dashboard/#axes_category_categoryarray , which should maybe be addressed in addition), allow people to disable the entry. Default to on so that introduction of the option is backwards compatible.

### DIFF
--- a/src/plot_api/plot_config.js
+++ b/src/plot_api/plot_config.js
@@ -48,8 +48,11 @@ module.exports = {
     // new users see some hints about interactivity
     showTips: true,
 
-    // enable/disable direct range entry at the pan/zoom drag points
-    enableRangeEntry: true,
+    // enable axis pan/zoom drag handles
+    showAxisDragHandles: true,
+
+    // enable direct range entry at the pan/zoom drag points (drag handles must be enabled above)
+    showAxisRangeEntryBoxes: true,
 
     // link to open this plot in plotly
     showLink: false,
--- a/src/plot_api/plot_config.js
+++ b/src/plot_api/plot_config.js
@@ -48,6 +48,9 @@ module.exports = {
     // new users see some hints about interactivity
     showTips: true,
 
+    // enable/disable direct range entry at the pan/zoom drag points
+    enableRangeEntry: true,
+
     // link to open this plot in plotly
     showLink: false,
 
--- a/src/plots/cartesian/graph_interact.js
+++ b/src/plots/cartesian/graph_interact.js
@@ -153,36 +153,39 @@ fx.init = function(gd) {
             };
 
             // corner draggers
-            dragBox(gd, plotinfo, -DRAGGERSIZE, -DRAGGERSIZE,
-                DRAGGERSIZE, DRAGGERSIZE, 'n', 'w');
-            dragBox(gd, plotinfo, xa._length, -DRAGGERSIZE,
-                DRAGGERSIZE, DRAGGERSIZE, 'n', 'e');
-            dragBox(gd, plotinfo, -DRAGGERSIZE, ya._length,
-                DRAGGERSIZE, DRAGGERSIZE, 's', 'w');
-            dragBox(gd, plotinfo, xa._length, ya._length,
-                DRAGGERSIZE, DRAGGERSIZE, 's', 'e');
+            if(gd._context.showAxisDragHandles) {
+                dragBox(gd, plotinfo, -DRAGGERSIZE, -DRAGGERSIZE,
+                    DRAGGERSIZE, DRAGGERSIZE, 'n', 'w');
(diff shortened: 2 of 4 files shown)


etpinard:
Ha, yeah that's probably plotly.js' most hidden feature. Good :eye: 

First, can you elaborate on the bug you're seeing? At first glance, those _range entry_ input boxes handlers behave as expected (i.e. as what a plain `relayout` call would), though the way to set axis `range` for categorical axes isn't fabulous.

I'm open to adding a config argument, but I'm not convinced _range entry_ is best at describing this feature. Maybe `showAxisRangeEntryBoxes`?

Moreover, we should make sure that the drag corners 

gifrecord_2017-02-20_114205

can also be optionally taken out.

rpaskowitz:
Here's the type of thing I see happing with a category axis:
rangeentry

Providing a way to disable pan/zoom on the chart area may make sense. I suppose the options would need to be named explicitly with reference to the drag handles since there would still be pan/zoom in the toolbar.  Hopefully it wouldn't be necessary to distinguish between the corners (which do zoom), and the mid-axis (which do pan)?

rpaskowitz:
Oh, and you may be right, what I see for a category axis isn't necessarily a bug, but still isn't what people would expect. (It's almost an argument for it being disabled all the time on category axes since while it works the same way it would via an API, I don't think there's any way for an end-user to understand the behavior). A larger discussion I'm sure, and doesn't necessarily need to be tied up in providing an ad-hoc method to disabling the range entry.

rpaskowitz:
Range entry option renamed and controlling whether the drag handles are shown to be begin with is added.

There is a dependency between the option now (noted in comment). If you disable the drag handles, even if range entry is enabled you won't be able to use it.

src/plot_api/plot_config.js
```diff
@@ -48,8 +48,11 @@ module.exports = {
     // new users see some hints about interactivity
     showTips: true,
 
-    // enable/disable direct range entry at the pan/zoom drag points
-    enableRangeEntry: true,
+    // enable axis pan/zoom drag handles
+    showAxisDragHandles: true,
```
etpinard: A straight-up boolean is fine for now.

I'm thinking down the road, we could allow users to specify which drag handles to show for example:

```
showAxisDragHandles: 'n+s'
```

would only show the north (or top) and south (or bottom) dragger.

But that's for another day. :+1:

src/plots/cartesian/graph_interact.js
```diff
@@ -153,36 +153,39 @@ fx.init = function(gd) {
...
-        // y axis draggers
-        if(isNumeric(x0)) {
-            if(ya.anchor === 'free') x0 -= fullLayout._size.w * xa.domain[0];
-            dragBox(gd, plotinfo, x0, ya._length * 0.1,
-                DRAGGERSIZE, ya._length * 0.8, 'ns', '');
-            dragBox(gd, plotinfo, x0, ya._length * 0.9,
-                DRAGGERSIZE, ya._length * 0.1, 's', '');
-            dragBox(gd, plotinfo, x0, 0,
-                DRAGGERSIZE, ya._length * 0.1, 'n', '');
+        if(gd._context.showAxisDragHandles) {
```
etpinard: Great! Before merging, we'll need to add a few test cases to :lock: this down.

In `test/jasmine/config_test.js`, call `Plotly.plot` with the a config object that set both new options to `false` and make sure that the corresponding handle and range entry boxes don't show up in the DOM. The [`showLink`](https://github.com/plotly/plotly.js/blob/143749ecfecf676adfaa42ff03469e273c3f8cea/test/jasmine/tests/config_test.js#L145-L179) should serve as a good example

rpaskowitz: Got some tests added.

test/jasmine/tests/config_test.js
```diff
@@ -269,4 +271,119 @@ describe('config argument', function() {
...
+            var corner = document.getElementsByClassName('edrag')[0];
+
+            var cornerBox = corner.getBoundingClientRect(),
+                cornerX = cornerBox.left + cornerBox.width / 2,
+                cornerY = cornerBox.top + cornerBox.height / 2;
+
+            click(cornerX, cornerY);
+
+            var editBox = document.getElementsByClassName('plugin-editable editable')[0];
+            expect(editBox).toBeUndefined();
```
etpinard: Wow. Very nicely done!

etpinard:
@rpaskowitz amazing work! Thanks for your efforts :beers: 

This patch will part of the new release `1.24.0`, set to be published early next week.

Issue #1381, opened 2017-02-15
`Virgin Islands (U.S.)` overwrites `United States` in choropleth


chriddyp:
![image](https://cloud.githubusercontent.com/assets/1280389/22960215/0a59e800-f30a-11e6-9685-a3a83da6ff94.png)

USA is getting colored in as 10 instead of 1.

Here is the example: https://plot.ly/~chris/18329.embed

etpinard:
`Virgin Islands (U.S.)` passes the USA regex in the [countrycode](https://github.com/vincentarelbundock/countrycode) package whereas `Virgin Islands` passes the [VIR](https://countrycode.org/virginislands) regex.

@chriddyp would you mind opening an issue upstream in https://github.com/vincentarelbundock/countrycode about this?

cjyetman:
pretty sure this has been fixed already upstream... make sure you're using the [most recent version (0.19) of countrycode](http://cran.r-project.org/web/packages/countrycode/index.html).

etpinard:
Thanks for your help @cjyetman !

PR #1392, opened 2017-02-17
Country regex update

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1381

cc @chriddyp

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "alpha-shape": "^1.0.0",
     "arraytools": "^1.0.0",
     "convex-hull": "^1.0.3",
-    "country-regex": "^1.0.0",
+    "country-regex": "^1.1.0",
     "d3": "^3.5.12",
     "delaunay-triangulate": "^1.1.6",
     "es6-promise": "^3.0.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "alpha-shape": "^1.0.0",
     "arraytools": "^1.0.0",
     "convex-hull": "^1.0.3",
-    "country-regex": "^1.0.0",
+    "country-regex": "^1.1.0",
```
etpinard: see https://github.com/etpinard/country-regex/commit/38259312897aeaf075cd550ced304eb399e1ed10

src/lib/geo_location_utils.js
```diff
@@ -51,7 +51,7 @@ function countryNameToISO3(countryName) {
         var iso3 = countryIds[i],
             regex = new RegExp(countryRegex[iso3]);
 
-        if(regex.test(countryName.toLowerCase())) return iso3;
+        if(regex.test(countryName.trim().toLowerCase())) return iso3;
```
etpinard: In case someone is wondering, `.trim` works in IE9

image

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/trim

chriddyp:
woohoo! :dancer:

Issue #236, opened 2016-02-03
De-circularize plotly.js require statements


etpinard:
@mdtusz 

**Latest list of circular dependencies:**
- Plots < --- > Annotations
- Plots < --- > Legend
- Plots < --- > Axes (the only `Plots` calls in `axes.js` are `traceIs`)
- Plots < --- > plotApi
- plotApi < --- > Annotations
- plotApi < --- > Shapes
- plotApi < --- > Titles
- plotApi < --- > Fx
- plotApit < --- > Legend

**Possible partial solutions**:
- Make a component registry and register the components (e.g. Annotations, Shapes, Titles, Fx, Legend), so `plot_api.js` doesn't have to require the full component modules (similar to trace and subplot modules). **Blocking:** we'll need a centralized api for the component module, e.g. all component modules should have a `draw` method and possible and `init` method.
- Use a more OO pattern (in the long term). Attach `relayout` and `restyle` to a plot object so that files requiring `plot_api.js` (e.g Annotations, Shapes and Fx) can use `gd.relayout()` instead of `plotApi.relayout(gd, ...)`.

etpinard:
The World Calendar push of https://github.com/plotly/plotly.js/pull/1220 added a new circular dependency pattern to `src/`:

![image](https://cloud.githubusercontent.com/assets/6675409/23086177/991fa1fc-f539-11e6-823b-26178d9b68d5.png)

as `src/registry.js` requires `Lib` for and `src/lib/dates.js` requires `Registry` to use the calendars component methods.

To solve this, we can either:

- make `lib/dates.js` a component module (as one could argue that `src/lib/` files should have to use the registry)
- require lib functions directly in `registry.js` instead of requiring the whole `Lib` module

alexcjohnson:
In the course of #1403 I made some circular deps involving `Lib`. In the short term I'm fixing these by breaking up `Lib` and just requiring the pieces I need, but longer term perhaps the right solution is to only allow things in `Lib` that do not `require` code from elsewhere in plotly.js. That would make it safe to import `Lib` anywhere, without forcing us to break it into a million little files. Would be a bit of an effort though...

etpinard:
Refer to PR https://github.com/plotly/plotly.js/pull/2032 for the most up-to-date findings.

PR #1394, opened 2017-02-17
Bump dev deps

etpinard:
... and fix the circular dependency check in `test-syntax`.


tasks/test_syntax.js
```diff
@@ -117,7 +119,7 @@ function assertCircularDeps() {
         // as of v1.17.0 - 2016/09/08
         // see https://github.com/plotly/plotly.js/milestone/9
         // for more details
-        var MAX_ALLOWED_CIRCULAR_DEPS = 17;
+        var MAX_ALLOWED_CIRCULAR_DEPS = 18;
```
etpinard: see https://github.com/plotly/plotly.js/issues/236#issuecomment-280789028

Issue #1333, opened 2017-01-25
choropleth hover labels can get lost


alexcjohnson:
If you zoom into a choropleth (say for example http://localhost:3000/devtools/test_dashboard/#geo_country-names) and hover on a region that goes off the edge, the hover label can get lost. Here's one where you just see the bottom of the Russia label (and if I zoom in a little more it disappears entirely):
![screen shot 2017-01-25 at 9 14 01 am](https://cloud.githubusercontent.com/assets/2678795/22293856/37e0fc9a-e2df-11e6-8069-20c7480b5e46.png)

The label should get pushed down onto the plot, like we do with regular cartesian types:
![screen shot 2017-01-25 at 9 22 14 am](https://cloud.githubusercontent.com/assets/2678795/22294055/d8223b60-e2df-11e6-815f-232bcef9dac5.png)

One thing that may help a bit: the hover labels don't need to be clipped to the same viewport as the plot: the tip of the arrow should be constrained to that viewport, but the rest of the label can go outside (as long as the whole thing stays within`gd`)

etpinard:
> the hover labels don't need to be clipped to the same viewport as the plot

Yeah. That would be a nice improvement to start. Choropleth hover labels should be drawn in the `<g infolayer>` node. While at it, we should :hocho: that useless `<div geo-container>`:

![image](https://cloud.githubusercontent.com/assets/6675409/22432524/223f30d4-e6e3-11e6-9242-67085beb105a.png)

and put the geo nodes in the main `<svg>`.

------------

Down the road, to better handle edge cases, we should consider extending the `hoveron: 'fill'` hover placement algo to geo polygons.

PR #1401, opened 2017-02-21
Better hover for choropleth traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1333

and cleans up a bunch of _geo-only_ code paths and logic.


src/plot_api/plot_api.js
```diff
@@ -2762,11 +2762,6 @@ function makePlotFramework(gd) {
     fullLayout._glcontainer.enter().append('div')
         .classed('gl-container', true);
 
-    fullLayout._geocontainer = fullLayout._paperdiv.selectAll('.geo-container')
-        .data([0]);
-    fullLayout._geocontainer.enter().append('div')
```
etpinard: An artefact from the first iteration back in July 2015 - where I found that [`Fx.lonehover`](https://github.com/plotly/plotly.js/blob/143749ecfecf676adfaa42ff03469e273c3f8cea/src/plots/cartesian/graph_interact.js#L752-L808) worked best off a plain `<div>` (fast-forward today, this not longer the case).

src/plot_api/plot_api.js
```diff
@@ -2810,6 +2805,9 @@ function makePlotFramework(gd) {
     // single ternary layer for the whole plot
     fullLayout._ternarylayer = fullLayout._paper.append('g').classed('ternarylayer', true);
 
+    // single geo layer for the whole plot
+    fullLayout._geolayer = fullLayout._paper.append('g').classed('geolayer', true);
```
etpinard: Now geo lies side-by-side with other SVG subplot types :tada:

src/traces/choropleth/plot.js
```diff
@@ -79,54 +43,20 @@ plotChoropleth.plot = function(geo, calcData, geoLayout) {
...
-            .on('mouseout', function() {
-                Fx.loneUnhover(geo.hoverContainer);
-
-                geo.graphDiv.emit('plotly_unhover', eventData);
-            })
-            .on('mousedown', function() {
-                // to simulate the 'zoomon' event
-                Fx.loneUnhover(geo.hoverContainer);
+            .on('mouseover', function(pt) {
+                geo.choroplethHoverPt = pt;
```
etpinard: `pt` here is essentially the calcdata item corresponding to the feature under the mouse cursor. `geo.choroplethHoverPt` is then used in the `hoverPoints` method. 

Alternatively, we could have looped over all the features and use something like [`turf-inside`](https://github.com/Turfjs/turf-inside) which is more similar to the `hoveron: 'fills'` algorithm, but I thought this here sufficed.

src/traces/choropleth/hover.js
```diff
@@ -0,0 +1,68 @@
...
+        hasText = (parts.indexOf('text') !== -1),
+        hasIdAsNameLabel = !hasName && hasLocation;
+
+    var text = [];
+
+    function formatter(val) {
+        return Axes.tickText(axis, axis.c2l(val), 'hover').text;
+    }
+
+    if(hasIdAsNameLabel) pointData.nameOverride = pt.id;
```
etpinard: see https://github.com/plotly/plotly.js/pull/1401/commits/ec6c1b1651a17738c0dcb289886d4f6664e74780

etpinard:
I can't notice any unwanted differences in the new baselines in https://github.com/plotly/plotly.js/pull/1401/commits/c32d445899038b3f278445221d3017d4a89786bb

If anything, the new baselines appear to better _clip_ the geo subplots:

gifrecord_2017-02-21_132722

I wouldn't mind having another set of :eyes: taking a look at this.

alexcjohnson:
Beautiful. Hover effects work better AND the code and svg structure is way simpler. 🌟 💯 🌟 

I'm not worried about the baseline image differences - seems like just subpixel rounding differences based on the different nesting structure. There does seem to be a little bit of overflow in the clipping, may be worthwhile taking a 🔍  to it at some point but doesn't need to be now - I don't think the effect changed with this PR, it just got a subpixel shift.

💃

Issue #1111, opened 2016-11-04
Data-referenced layout images should contribute to autorange calculations


etpinard:
as discovered by @alexcjohnson in https://github.com/plotly/plotly.js/pull/1110#discussion_r86576948

etpinard:
cc https://github.com/plotly/plotly.js/issues/2704

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #1403, opened 2017-02-21
Array edits

alexcjohnson:
Standardizes our editing behavior for the growing list of container arrays - ie anything in the figure that is an array of objects, like annotations, shapes, sliders, updatemenus - and gets that behavior out of the individual components and collects it in the update machinery. Also ensures that multiple updates within one `relayout` call will behave in well-defined ways.

The behavior (which is mostly how annotations and shapes worked before this but it's extended to be more complete and robust) is:
- If you set an entire element (eg `Plotly.relayout(gd, {'annotations[2]': {...}})`) it always inserts a new element into the array at that position.
- You can delete an element by setting it to null `Plotly.relayout(gd, {'annotations[2]': null})`
- You can edit attributes of existing elements `Plotly.relayout(gd, {'annotations[2].text':'boo'})` but you cannot create new elements by setting attributes of nonexistent elements.
- Order of operations is designed so any arbitrary combination of edits is possible, and will be reversible.
  - First we apply the insertions and edits, in order of ascending indices
  - Then we apply the deletions, in order of decreasing indices
  - it's not allowed to combine inserting/edit/deletion of the same index. The edit will be dropped and a warning logged.

In addition:
- I removed any automatic conversion of positions when changing the reference axes of annotations. Previously we had plotly.js try to keep the pixel position of the annotation fixed when you changed the reference axis. This more or less worked (although a bug with it https://github.com/plotly/streambed/issues/8901 was the original impetus for me to look at this) but it became clear when considering multiple simultaneous edits that there were going to be an incredible number of edge cases that would make this unmanageable, and nearly impossible to divine what the user really wants. Anyway this is really only going to be an issue within the plot.ly workspace, and in there the set of possible individual actions is much smaller (you change axis reference by itself, with no simultaneous changes), so there will be a companion PR to this one over there that we'll need to merge along with the next plotly.js version update.
- But we do still convert positions for log/linear axis changes. This only applies to annotations and images (but we didn't do it at all for images previously) since shapes use data coordinates so need no conversion to log. I moved this out of the `Annotations.draw` method and instead drive it from within `_relayout`.
- Added a new layout attributes key, `_arrayAttrRegexps`, to support variably named container arrays. Right now the examples of this are `mapbox2.layers` and `updatemenus[2].buttons`.
- `getComponentMethod` now returns `Lib.noop` even when the component itself isn't found. I did this so I could look for `draw` and `drawOne` methods based on the array name, but they don't always match, then we fall back on replotting, which is what we want to do anyway.

Three things to keep in mind for later:
- I did not extend this to `restyle` (eg the `transforms` array, but I guess more are coming eg with parcoords) but we should do this too.
- We should convert shapes and annotations to a more d3-idiomatic construction and get rid of `drawOne` entirely
- The log/linear conversion stuff can all disappear too in V2.0 when we change annotations and images to data coordinates rather than linearized coordinates.


rreusser:
Unless you've figured out further details, debugging the command issue…

src/lib/nested_property.js
```diff
@@ -118,19 +120,38 @@ function npGet(cont, parts) {
...
+// }
 
-    return isArray(val) && isNotAContainer;
+/*
+ * Can this value be deleted? We can delete any empty object (null, undefined, [], {})
+ * EXCEPT empty data arrays. If it's not a data array, it's a container array,
+ * ie containing objects like annotations, buttons, etc
+ */
+var DOMAIN_RANGE = /(^|.)(domain|range)$/;
+function isDeletable(val, propStr) {
```
etpinard: Does this work for `domain.x` and `domain.y` ?

etpinard: And more generally, shouldn't all `info_array` attributes (e.g. updemenues/sliders `args`, and `geo.projection.parallels`) be deletable?

src/plot_api/plot_api.js
```diff
@@ -1717,12 +1723,16 @@ Plotly.relayout = function relayout(gd, astr, val) {
     if(flags.docalc) gd.calcdata = undefined;
 
     // fill in redraw sequence
-    var seq = [];
+
+    // even if we don't have anything left in aobj,
+    // something may have happened within relayout that we
+    // need to wait for
+    var seq = [Plots.previousPromises];
```
etpinard: Good eyes :microscope:

src/components/updatemenus/attributes.js
```diff
@@ -49,6 +49,7 @@ var buttonsAttrs = {
 
 module.exports = {
     _isLinkedToArray: 'updatemenu',
+    _arrayAttrRegexps: [/^updatemenus\[(0|[1-9][0-9]+)\]\.buttons/],
```
etpinard: I'm ok with adding `_arrayAttrRegexps` in this PR (there's enough new logic it in already :tired_face: ), but we could probably generate it at runtime down the road in the registry [here](https://github.com/plotly/plotly.js/blob/143749ecfecf676adfaa42ff03469e273c3f8cea/src/registry.js#L92-L95)?

Isn't `_arrayAttrRegexps` simply,

```js
allIsLinkedToArrayAttributes.map((attr) => {
  return module.name + '\[[(0|[1-9][0-9]+)\]\.' + attr;
});
```

rreusser:
Verdict on the sliders testing problem: private (underscored) properties on the container objects are not currently getting transferred over, so it wasn't cleaning up after itself.

src/components/updatemenus/attributes.js
alexcjohnson: that wouldn't work for the other instance of this, [`/^mapbox([2-9]|[1-9][0-9]+)?\.layers/`](https://github.com/plotly/plotly.js/pull/1403/files#diff-26aaae812b2edd7de7c8d4ab79df84f0R19)

src/plot_api/plot_api.js
alexcjohnson: Good tests 🎉

src/plots/array_container_defaults.js
```diff
@@ -64,4 +67,13 @@ module.exports = function handleArrayContainerDefaults(parentObjIn, parentObjOut
...
     }
+
+    // in case this array gets its defaults rebuilt independent of the whole layout,
+    // relink the private keys just for this array.
+    if(Lib.isArray(previousContOut)) {
+        var len = Math.min(previousContOut.length, contOut.length);
+        for(i = 0; i < len; i++) {
+            Lib.relinkPrivateKeys(contOut[i], previousContOut[i]);
+        }
+    }
```
alexcjohnson: @rreusser here's the fix for the command test. However if you add or delete an object, the private key will get reattached to the wrong item, or lost. I'm not sure what to do about this... I could splice the `fullLayout` array in sync with the `layout` array, which would keep the private key with the right index, but I don't know what to do when I delete an element that has private keys. Most of the time it doesn't matter, but in the case of `._remove` I guess it would.

src/plots/plots.js
```diff
@@ -2012,6 +2015,18 @@ plots.doCalcdata = function(gd, traces) {
...
+    // need to re-supply defaults for these objects now, after calc has
+    // finished populating the category mappings
+    // Any component that uses `Axes.coercePosition` falls into this category
+    if(hasCategoryAxis) {
+        var dataReferencedComponents = ['annotations', 'shapes', 'images'];
+        for(i = 0; i < dataReferencedComponents.length; i++) {
+            Registry.getComponentMethod(dataReferencedComponents[i], 'supplyLayoutDefaults')(
+                gd.layout, fullLayout, fullData);
+        }
+    }
```
alexcjohnson: I suppose I could list these components somehow in the registry... worth it or overkill?

BTW you'll see I added category-name coordinates to image and shape test images, annotations didn't need this because 16.json already has it, and that's the failing test that prompted this fix.

src/lib/noop.js
```diff
@@ -0,0 +1,14 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+// Simple helper functions
+// none of these need any external deps
+
+module.exports = function noop() {};
```
alexcjohnson: I thought this was about the shortest possible module one could make... but I suppose you could do:
```js
exports.a = 0;
```
Kind of silly... but needed to de-circularize, short of refactoring `Lib` (see https://github.com/plotly/plotly.js/issues/236 )

tasks/test_syntax.js
```diff
@@ -119,13 +119,14 @@ function assertCircularDeps() {
         // as of v1.17.0 - 2016/09/08
         // see https://github.com/plotly/plotly.js/milestone/9
         // for more details
-        var MAX_ALLOWED_CIRCULAR_DEPS = 18;
+        var MAX_ALLOWED_CIRCULAR_DEPS = 17;
 
         if(circularDeps.length > MAX_ALLOWED_CIRCULAR_DEPS) {
+            console.log(circularDeps.join('\n'));
```
alexcjohnson: Just to see where I had made circular refs when this test was failing. So now if you want to see them all just set `MAX_ALLOWED_CIRCULAR_DEPS = 0` temporarily and run this.

src/lib/nested_property.js
alexcjohnson: Yes, good point. I was trying to match the previous logic and extend it gently but honestly I can't recall anymore why we had that logic - why does an empty data array need to stick around while other empty arrays can be removed entirely? Perhaps just to keep things simple for streaming? Does that point to a more robust condition on what should be deletable?

tasks/test_syntax.js
etpinard: Nice touch 🎆

src/plot_api/plot_api.js
```diff
@@ -2092,9 +2092,6 @@ function _relayout(gd, aobj) {
     var oldWidth = gd._fullLayout.width,
         oldHeight = gd._fullLayout.height;
 
-    // coerce the updated layout
-    Plots.supplyDefaults(gd);
```
etpinard: Really? This doesn't break the autosize teats? I thought I tried  removing it at some point a few weeks back. 

Oh well. One less supply defaults loop ia great news 🐎

src/plot_api/plot_api.js
```diff
@@ -251,7 +251,7 @@ Plotly.plot = function(gd, data, layout, config) {
         // calc and autorange for errorbars
         ErrorBars.calc(gd);
 
-        // TODO: autosize extra for text markers
+        // TODO: autosize extra for text markers and images
```
etpinard: There's actually an issue for that one https://github.com/plotly/plotly.js/issues/1111

src/plots/plots.js
etpinard: Let's leave it like that for now. We could always add that list to the registry down the road if it comes up somewhere else. 👍

src/plots/array_container_defaults.js
rreusser: Thanks! Hmm… scrambling at the moment for a call this afternoon, but let me think about this and process a bit.

src/plot_api/plot_api.js
alexcjohnson: Yeah, this is ridiculously complicated... actually tracing through it I found a very specific new bug, though not sure yet if it's exactly related. But I think we can sort it out without reinstating that redundant `supplyDefaults`:

All of the pathways after `_relayout` in the `relayout`, and `update` wrappers include full `supplyDefaults` - with the exception of `relayout` when the edits object `aobj` has had everything stripped out of it and executed directly within `_relayout` - and THAT can happen if all keys are handled by `editContainerArray`, which is OK because inside `editContainerArray` we call `supplyDefaults` for that particular component array. This won't work for the regex arrays, because the array name doesn't match a component, but then we don't find a `draw` method either so it reverts to a full replot.

So I think the only potential pitfall of this is if a change inside the array affects something in the layout outside the array. One case I thought *might* fit this bill is if you provide a new axis id in a component object (new or edited) - but actually out now as previously we don't generate new axes just from component references, so:

`Plotly.relayout(gd,{'annotations[1]':{text:'boo',x:-1,y:4, xref:'x2', yref:'y2'}})`

on a plot with only `xaxis` and `yaxis` will get drawn on the existing axes... but the bug is that in this case the axes don't get correctly autoranged until you redraw the plot. It's ONLY this case, autorange works if your axis refs are valid.

So, are there any cases I'm not thinking of where changes inside one of these arrays affect `_fullLayout` outside of the array?

src/lib/nested_property.js
alexcjohnson: Alright, letting every array be deletable and running the full test suite was fairly enlightening - There are a lot of things that fail or error out, some of which could be addressed but fundamentally for data arrays there really is a difference between `[]` (which means there is data of that kind, it's just presently empty) and `undefined` (which means the data/attribute should get some default value filled in).

As far as I can tell, for `info_array` attributes there is no difference between `[]` and `undefined` though. So unless we want to try and build in here some way to hunt through the schema (which sounds super painful to me, and potentially not even well-defined given that `nestedProperty` can, on occasion, be called on a root object that's neither a complete trace or layout object but some portion of it) I would argue for some simple tests that might frequently describe an info array as falsely *not* deletable but will never describe a real data array as deletable.

src/plot_api/plot_api.js
etpinard: > > So, are there any cases I'm not thinking of where changes inside one of these arrays affect _fullLayout outside of the array?

I can't think of any other cases.

src/lib/nested_property.js
etpinard: > As far as I can tell, for info_array attributes there is no difference between [] and undefined though

That's correct.

src/lib/push_unique.js
```diff
@@ -0,0 +1,36 @@
...
+ * @return {array}
+ *  ref to array (now possibly containing one more item)
+ *
+ */
+module.exports = function pushUnique(array, item) {
+    if(item instanceof RegExp) {
+        var itemStr = item.toString(),
+            i;
+        for(i = 0; i < array.length; i++) {
+            if(array[i] instanceof RegExp && array[i].toString() === itemStr) {
```
etpinard: Can you add a few test cases for regex items ->

https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/lib_test.js#L1246

src/plot_api/manage_arrays.js
```diff
@@ -0,0 +1,201 @@
...
+ * in ascending order, then removed in descending order.
+ * For example, starting with [a, b, c], if you want to:
+ * - replace b with d:
+ *   {'annotations[1]': d, 'annotations[2]': null} (b is item 2 after adding d)
+ * - add a new item d between a and b, and edit b:
+ *    {'annotations[1]': d, 'annotations[2].x': newX} (b is item 2 after adding d)
+ * - delete b and edit c:
+ *    {'annotations[1]': null, 'annotations[2].x': newX} (c is edited before b is removed)
+ *
+ * You CANNOT combine adding/deleting an item at index `i` with edits to the same index `i`
```
etpinard: Can we add tests for these CANNOT cases?

etpinard: Oops. Just saw https://github.com/plotly/plotly.js/pull/1403/files#diff-70e5a2fafe3d8a7e1ed2e632e9781df9R300

Ignore.

src/plot_api/plot_api.js
```diff
@@ -1876,142 +1898,170 @@ function _relayout(gd, aobj) {
...
+        if(containerArrayMatch) {
+            arrayStr = containerArrayMatch.array;
+            i = containerArrayMatch.index;
+            var propStr = containerArrayMatch.property,
+                componentArray = Lib.nestedProperty(layout, arrayStr),
+                obji = (componentArray || [])[i] || {};
+
+            if(i === '') {
+                // replacing the entire array: too much going on, force recalc
+                if(ai.indexOf('updatemenus') === -1) flags.docalc = true;
```
etpinard: why does `'updatemenus'` need special treatment here?

src/lib/push_unique.js
etpinard: https://github.com/plotly/plotly.js/pull/1403/files#diff-9edb82cf7ed7d529d7e77cce5714360aR1287

I need better :eyeglasses:

src/plot_api/plot_api.js
alexcjohnson: proximate reason: because [this test](https://github.com/plotly/plotly.js/blob/7ed9eb9ffe052c4b2cba183c753e130a42074cb3/test/jasmine/tests/plot_api_test.js#L1296) fails after recalc because `gd.calcdata` is a new object. But what that shows is that we don't actually need a recalc in this case. We probably could exclude from recalc everything that doesn't contribute to autorange (so, I suspect only annotations, shapes, and eventually images need recalc) but without taking the time to investigate all the other cases in detail I erred on the side of caution.

src/plot_api/plot_api.js
alexcjohnson: ah thanks - comment-linked in https://github.com/plotly/plotly.js/pull/1403/commits/46962af8bdb37ee77bef8880c41baf7186f095ac

src/plot_api/plot_api.js
alexcjohnson: fixed that bad-reference bug in https://github.com/plotly/plotly.js/pull/1403/commits/9f89dd690a638a4979d77d050667a1557baf91e5 - it was unrelated to removing this `supplyDefaults`, which I still believe is OK to do.

src/lib/nested_property.js
alexcjohnson: expanded the safe-to-delete list to include all `info_array` attributes in https://github.com/plotly/plotly.js/pull/1403/commits/1af058ab343062365eadb5ec0c60251c4c78eabf

@rreusser this also required making `args` optional for sliders and updatemenus (fallback to `[]` in command.js) since an empty array will now get pruned. Not sure there will be a real use case for this (what method could one call with no args?) but this case pops up in the tests anyhow 🏆  and it does seem more plotly-esque to not require empty parameters.

rreusser: The only use case I'm aware of is animating list of frames `[]` which (serendipitously) functions as a pause button. I've had a bit of trouble with this behavior, so we documented pretty clearly in the web docs that you must use `[null]` instead of `[]` in order to get it passed through.

src/plot_api/plot_api.js
etpinard: Ok. That's fine. Maybe we should open (another) `Registry` list for components that have a `calcAutorange` method (like annotations and shapes) for this purpose here. But that's not necessary in this PR.

src/lib/nested_property.js
```diff
@@ -113,24 +115,36 @@ function npGet(cont, parts) {
...
+ * tests here to find known non-data arrays but don't worry too much about not deleting
+ * some arrays that would actually be safe to delete.
  */
-function isDataArray(val, key) {
-
-    var containers = ['annotations', 'shapes', 'range', 'domain', 'buttons'],
-        isNotAContainer = containers.indexOf(key) === -1;
-
-    return isArray(val) && isNotAContainer;
+var INFO_PATTERNS = /(^|.)((domain|range)(\.[xy])?|args|parallels)$/;
```
etpinard: Ok, well this is the only part I dislike about this PR. I'm ok with leaving as is for now, but should think about ways to :hocho: this down the road.

Solution 1: simply don't delete `info_array` attributes, like `data_array`

Solution 2: use the plot schema to look up all `info_array` attributes. This doesn't have to be done inside nested property where objects are set w/o a reference to their parent object (if any). Maybe this could be done only during `restyle`, `relayout` and `update`.

Solution 3: something else

src/plots/array_container_defaults.js
etpinard: @rreusser can you confirm that is working as it should?

src/components/updatemenus/attributes.js
etpinard: Just wondering why you chose to attach `_arrayAttrRegexps` to the attribute object? 

Alternatively, we could set in the module object [here](https://github.com/plotly/plotly.js/blob/027d97fe3f9c2fb2d5e816d41719df49eb49c763/src/components/annotations/index.js#L15) - and not having to skip it in the plot schema [logic](https://github.com/plotly/plotly.js/pull/1403/files#diff-9d405afa212832aed790fcec3d0e16f9R31).

alexcjohnson: Yes, we could... I just put it there because that's where `_isLinkedToArray` lives and I thought it would be good to keep them together. I don't have a strong opinion either way.

etpinard: > I just put it there because that's where _isLinkedToArray lives

Very good point. Let's keep it that way :+1:

src/lib/nested_property.js
alexcjohnson: added even more complexity https://github.com/plotly/plotly.js/pull/1403/commits/c6378b12e0d2cd99e7fc2541a6dfcf3b4597fdc2 after discussion with @rreusser but opened #1410 to discuss simplification later on.

src/plots/array_container_defaults.js
alexcjohnson: https://github.com/plotly/plotly.js/pull/1403/commits/70c882bceaa5236b7c1e7cdf51903c5d1823449b keeps the private keys in sync to the degree that it can - which is unless you delete elements from the array. Then they get lost. @rreusser as this isn't new behavior are you OK with it as it stands now?

test/jasmine/tests/lib_test.js
```diff
@@ -375,6 +375,30 @@ describe('Test lib.js:', function() {
             expect(obj).toEqual({a: [undefined, {}]});
         });
 
+        it('does not prune inside `args` arrays', function() {
```
etpinard: Nicely done. Thanks!

etpinard:
:dancer: for me! Nice way to make array container a little more solid!

@alexcjohnson can you also open up a `type: maintenance` issue about making trace data arrays use the same logic?

src/lib/noop.js
rreusser: (resisting temptation suggest a [noop factory](https://github.com/sindresorhus/noop3/blob/master/factory.js) approach)

src/plot_api/helpers.js
```diff
@@ -484,3 +484,19 @@ exports.manageArrayContainers = function(np, newVal, undoit) {
         np.set(newVal);
     }
 };
+
+var ATTR_TAIL_RE = /(\.[^\[\]\.]+|\[[^\[\]\.]+\])$/;
```
rreusser: Could you add maybe a single line or two about the sort of pattern this matches? I'm guessing… dot… more than one of… something… that's not brackets… just something that's not brackets or a dot? So

```
.something
something
```

match while

```
.something[]
something[]
```

do not?

rreusser: (not a big deal; just difficult to confirm we're on the same page about what the end of an attr string means by mentally parsing the regexp)

src/lib/noop.js
alexcjohnson: Actually it's important that `noop` is a singleton so you can use [`=== noop`](https://github.com/plotly/plotly.js/pull/1403/files#diff-621069fcc5090ecff8632dbba197ce0eR77) particularly on the results of `Registry.getComponentMethod` to see if it was found.

src/plot_api/manage_arrays.js
```diff
@@ -0,0 +1,211 @@
...
+
+/*
+ * editContainerArray: for managing arrays of layout components in relayout
+ * handles them all with a consistent interface.
+ *
+ * Here are the supported actions -> relayout calls -> edits we get here
+ * (as prepared in _relayout):
+ *
+ * add an empty obj -> {'annotations[2]': 'add'} -> {2: {'': 'add'}}
+ * add a specific obj -> {'annotations[2]': {attrs}} -> {2: {'': {attrs}}}
```
rreusser: I would have expected this to say `{'annotations[2]': {attrs}} -> {2: {attrs}}`, but I may have content confused with the action to insert content.

src/plot_api/manage_arrays.js
```diff
@@ -0,0 +1,211 @@
...
+ * delete an obj -> {'annotations[2]': 'remove'} -> {2: {'': 'remove'}}
+ *               -> {'annotations[2]': null} -> {2: {'': null}}
+ * delete the whole array -> {'annotations': 'remove'} -> {'': {'': 'remove'}}
+ *                        -> {'annotations': null} -> {'': {'': null}}
+ * edit an object -> {'annotations[2].text': 'boo'} -> {2: {'text': 'boo'}}
+ *
+ * You can combine many edits to different objects. Objects are added and edited
+ * in ascending order, then removed in descending order.
```
rreusser: Thanks for enforcing that! As discussed elsewhere, the spec does not require any particular order when iterating over object keys, so I feel better knowing we're not just figuring people deserve to suffer the consequences if they combine operations ambiguously.

src/plot_api/manage_arrays.js
```diff
@@ -0,0 +1,211 @@
...
+        layout = gd.layout,
+        fullLayout = gd._fullLayout;
+
+    if(edits['']) {
+        if(Object.keys(edits).length > 1) {
+            Loggers.warn('Full array edits are incompatible with other edits',
+                componentType);
+        }
+
+        var fullVal = edits[''][''];
```
rreusser: Is it conceivable that `edits['']` could ever not be array-indexable?

src/plot_api/manage_arrays.js
```diff
@@ -0,0 +1,211 @@
...
+ *  in special values of the empty attr)
+ * @param {Object} flags
+ *  the flags for which actions we're going to perform to display these (and
+ *  any other) changes. If we're already `recalc`ing, we don't need to redraw
+ *  individual items
+ *
+ * @returns {bool} `true` if it managed to complete drawing of the changes
+ *  `false` would mean the parent should replot.
+ */
+exports.editContainerArray = function editContainerArray(gd, np, edits, flags) {
```
rreusser: Is `editContainerArray` the best name for this function? Grammar-wise, `applyContainerArrayChanges` *might* be more descriptive, and functionality-wise, I didn't assume by the name that it would also supply defaults and perform drawing updates.

src/plot_api/manage_arrays.js
rreusser: Thanks for enforcing that! As discussed elsewhere, the spec does not require any particular order when iterating over object keys, so I feel better knowing we're not just figuring people deserve to suffer the consequences if they combine operations ambiguously.

src/plot_api/manage_arrays.js
rreusser: I would have expected this to say `{'annotations[2]': {attrs}} -> {2: {attrs}}`, but I may have content confused with the action to insert content.

src/plot_api/manage_arrays.js
rreusser: Is it conceivable that `edits['']` could ever not be array-indexable?

src/plot_api/plot_api.js
```diff
@@ -1753,13 +1765,16 @@ function _relayout(gd, aobj) {
         fullLayout = gd._fullLayout,
         keys = Object.keys(aobj),
         axes = Plotly.Axes.list(gd),
-        i;
+        arrayEdits = {},
+        arrayStr,
+        i,
+        j;
```
rreusser: FYI: I think by far the biggest manual change required to move from eslint to standard is that variable declarations like this must manually be split. The general rule is that initialized declarations must each have their own `var` and uninitialized declarations are fine in the same statement. So no particular feedback here, just that every time I see these, I worry about the big cleanup pass ahead 😬

src/plot_api/plot_api.js
```diff
@@ -1876,142 +1911,170 @@ function _relayout(gd, aobj) {
...
+                    doextra(ptrunk + '.autorange', true);
+                }
+                else if(!parentFull.autorange) {
+                    // toggling log without autorange: need to also recalculate ranges
+                    // because log axes use linearized values for range endpoints
+                    var r0 = ax.range[0],
+                        r1 = ax.range[1];
+                    if(toLog) {
+                        // if both limits are negative, autorange
+                        if(r0 <= 0 && r1 <= 0) {
```
rreusser: This is nothing new and probably not worth addressing, but lots of logic and log calculations deep inside a 350+ line function seem like they'd be difficult to test and maintain. Would be nice if there were a better place for the logic.

src/plot_api/plot_api.js
```diff
@@ -2032,12 +2095,16 @@ function _relayout(gd, aobj) {
         }
     }
 
+    // now we've collected component edits - execute them all together
+    for(arrayStr in arrayEdits) {
+        var finished = manageArrays.editContainerArray(gd,
+            Lib.nestedProperty(layout, arrayStr), arrayEdits[arrayStr], flags);
```
rreusser: Same feedback. If this is the only place it's used, perhaps the supplydefaults and draw parts should be outside the function that manages arrays? You have a better sense than I do though, so feel free to disregard :)

src/plots/array_container_defaults.js
rreusser: My benchmark for success is just whether it passes the component test that was failing due to the `._remove` property getting lost. It was originally just called `.remove` which would still get lost, so I'll check and submit that fix as a separate PR as needed.

src/registry.js
```diff
@@ -9,7 +9,9 @@
 
 'use strict';
 
-var Lib = require('./lib');
+var Loggers = require('./lib/loggers');
+var noop = require('./lib/noop');
```
rreusser:
```javascript
echo "var noop = require('./lib/noop');" |wc -c
34
$ echo "function noop() {}" |wc -c               
19
```

No. Stop. Don't bikeshed, Ricky. Don't bikeshed. LGTM. 👍

test/jasmine/tests/command_test.js
```diff
@@ -594,7 +594,8 @@ describe('attaching component bindings', function() {
             expect(gd.layout.sliders[0].active).toBe(0);
 
             // Check that it still has one attached listener:
-            expect(typeof gd._internalEv._events.plotly_animatingframe).toBe('function');
+            expect(typeof gd._internalEv._events.plotly_animatingframe).toBe('function',
+                gd._internalEv._events.plotly_animatingframe);
```
rreusser: The second argument to `toBe` adds a message? No worries, just checking to make sure there's no unintended difference in function.

src/plots/array_container_defaults.js
rreusser: Wait, no, I must be mistaken. It passes the test in its current form so the changes must have fixed it. Again, that was my only metric for this, so I'm satisfied although I wouldn't have expected fixing private keys to fix the problem. Hmmm.

src/plot_api/helpers.js
alexcjohnson: good call - also added some tests for `hasParent` -> https://github.com/plotly/plotly.js/pull/1403/commits/5fb462fa6b8c256e5fc179c028d59c10ce5c7b6d

src/plot_api/manage_arrays.js
alexcjohnson: right - it's a bit of a weird structure I ended up with as input to `editContainerArray`. It's probably clear, but this is prepared in [`_relayout`](https://github.com/plotly/plotly.js/pull/1403/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R2024) and users never see it. It's structured as `edits[index][attribute]` - so if you're setting an array element all at once, `attribute` will be `''`, and if you're setting the whole array at once, both `attribute` and `index` will be `''`

src/plot_api/manage_arrays.js
alexcjohnson: per https://github.com/plotly/plotly.js/pull/1403#discussion_r102981565 - full-array changes will always have exactly this form. That was just the easiest construction for `_relayout` to do without edge cases, though I agree it looks a little odd in here...

rreusser: All good! Assumed so. Just a PR-review reflex. 👍

src/plot_api/manage_arrays.js
rreusser: Perfect. 👍

test/jasmine/tests/command_test.js
alexcjohnson: yes - that's true for most of the jasmine matchers, and can be really useful in debugging failures.

src/plot_api/manage_arrays.js
alexcjohnson: The other ambiguity-removing thing I did was to make both [_restyle](https://github.com/plotly/plotly.js/pull/1403/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R1403) and [_relayout](https://github.com/plotly/plotly.js/pull/1403/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R1847) error out if you try to edit both an object and one of its parents. That said, perhaps based on some other conversations of ours @etpinard in this case we should just log a warning and drop the child attribute, rather than throwing an error?

src/plot_api/manage_arrays.js
alexcjohnson: The reason it's not `{'annotations[2]': {attrs}} -> {2: {attrs}}` is that `{2: {attrs}}` and `{2: {'': {attrs}}` mean different things: the former would change just `attrs`, leaving intact anything else that's already present in `annotations[2]`. The latter inserts a new object containing just `{attrs}`.

src/plot_api/plot_api.js
alexcjohnson: Interesting... what's "standard"?

src/plot_api/manage_arrays.js
rreusser: So `{'marker.color': 'red', marker: {color: 'red}}` errors out? I don't have strong feelings, though plotly's general solution seems to be to log an error and do the best it can.

src/plot_api/plot_api.js
alexcjohnson: Man, the single biggest change I'm looking forward to in v2.0 is not using linearized coordinates for log axes. All of this crap will disappear then.

src/plot_api/plot_api.js
rreusser: See: https://github.com/plotly/plotly.js/pull/1371

rreusser: Was missing link to project page. It's: https://github.com/feross/standard

rreusser: Long story short, *unconfigurable* eslint.

src/plot_api/plot_api.js
alexcjohnson: For now I'll change the name per https://github.com/plotly/plotly.js/pull/1403#discussion_r102964323 -> https://github.com/plotly/plotly.js/pull/1403/commits/d29cbae52654f1e74d29e77dbc353d021353eb74 but I suspect in implementing #1412 we will restructure it along the lines you're suggesting.

src/plot_api/manage_arrays.js
etpinard: To me, they are two types of plotly.js errors:

- Invalid plot attribute combinations,
- Invalid input to `Plotly.___` methods.

I believe the former should never lead to errors as @rreusser said plotly.js should try to _do the best it can_. This is especially important for non-JS users that send `"data"` and `"layout"` and expect to get in return the _best possible_ :tm: graph.

This case :arrow_up: falls into the second category. `Plotly.__` method are called by JS users only* and throwing errors should help them debug. For example in `Plotly.plot(gd, fig)` if `gd` isn't right we [throw](https://github.com/plotly/plotly.js/blob/2533f04c12f0de69cc1b227d1eccd58b98c5d53f/src/plot_api/helpers.js#L25-L42) an error. I think the same should be expected for invalid relayout update object.

*: Note that updatemenus and slider now expose `Plotly.__` methods to non-JS users now. :expressionless:

rreusser: Good philosophical points, @etpinard!

test/jasmine/tests/plot_api_test.js
```diff
@@ -348,7 +384,45 @@ describe('Test plot api', function() {
                 expect(gd.data[0].zmax).toBe(10);
 
                 expect(gd.data[0].contours).toEqual({start: 3, end: 9, size: 1});
-            });
+            })
+            .catch(fail)
+            .then(done);
```
alexcjohnson: Hah, this test wasn't actually running. More precisely, it was running but not being used. Who wrote that sloppy thing? 🙈

etpinard: image

Not me :smile:

alexcjohnson: The biggest irony is my commit message there:

`fix the tests I must have only thought I had run successfully`

Issue #1407, opened 2017-02-22
Hoverinfo with date axis in 3D broken when using Date objects


jdugge:
If one of the axes in a 3D plot is a date axis and the data for this axis consists of Date objects, the hoverinfo is displayed as `NaN, 0NaN, aNaN:aN`.

![image](https://cloud.githubusercontent.com/assets/2399255/23223690/02e93e36-f92c-11e6-9bab-3519fd839fc3.png)

[Codepen](http://codepen.io/anon/pen/RpbQWQ)

When the date data is provided as date strings, these strings are displayed in the hoverinfo instead, instead of being parsed and formatted as dates.

etpinard:
Thanks for the report.

PR #1414, opened 2017-02-24
3D hover text fixes for date axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1407

### DIFF
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -58,10 +58,10 @@ function render(scene) {
     }
 
     function formatter(axisName, val) {
-        if(typeof val === 'string') return val;
-
         var axis = scene.fullSceneLayout[axisName];
-        return Axes.tickText(axis, axis.c2l(val), 'hover').text;
+        var calendar = trace[axis._id.charAt(0) + 'calendar'];
+
+        return Axes.tickText(axis, axis.d2l(val, 0, calendar), 'hover').text;
     }
 
     var oldEventData;
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -59,9 +59,8 @@ function render(scene) {
 
     function formatter(axisName, val) {
         var axis = scene.fullSceneLayout[axisName];
-        var calendar = trace[axis._id.charAt(0) + 'calendar'];
 
-        return Axes.tickText(axis, axis.d2l(val, 0, calendar), 'hover').text;
+        return Axes.tickText(axis, axis.d2l(val), 'hover').text;
     }
 
     var oldEventData;
--- a/test/jasmine/tests/gl_plot_interact_test.js
+++ b/test/jasmine/tests/gl_plot_interact_test.js
@@ -88,7 +88,7 @@ describe('Test gl plot interactions', function() {
 
         describe('scatter3d hover', function() {
 
-            var node, ptData;
+            var ptData;
 
             beforeEach(function(done) {
                 gd.on('plotly_hover', function(eventData) {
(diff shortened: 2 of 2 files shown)


test/jasmine/tests/gl_plot_interact_test.js
```diff
@@ -100,14 +100,18 @@ describe('Test gl plot interactions', function() {
...
-                expect(node[0].innerHTML).toEqual('x: 140.72', 'x val on hover');
-                expect(node[1].innerHTML).toEqual('y: −96.97', 'y val on hover');
-                expect(node[2].innerHTML).toEqual('z: −96.97', 'z val on hover');
+                var tspan = d3.selectAll('g.hovertext').selectAll('tspan')[0];
+                expect(tspan[0].innerHTML).toEqual(xLabel, 'x val on hover');
+                expect(tspan[1].innerHTML).toEqual(yLabel, 'y val on hover');
+                expect(tspan[2].innerHTML).toEqual(zLabel, 'z val on hover');
+            }
+
+            it('should have', function(done) {
```
alexcjohnson: This isn't new with this PR, but this labeling is a little weird - it looks like you're including the message ('x val on hover' etc) as part of the test description, but doing that doesn't make it read out the right way either in the code or when the test fails.

The way I understand it, these tests should be described as:
```js
describe('Noun', function() {
    it('makes a sentence when you include the "it"', function() {
        expect(true).toBe(false, 'more info if useful pinpoint the problem');
    });
});
```

Which, when it fails, will make a message that starts with a sentence:

```
Chrome 56.0.2924 (Mac OS X 10.12.3) Noun makes a sentence when you include the "it" FAILED
        Expected true to be false, more info if useful pinpoint the problem.
            <traceback>
```

So in this case I'd have expected it to be something like:

```js
describe('gl plot interactions', function() {
    it('makes the right hover text and point data', function() {
        // I'm not sure if the messages add much, I typically only include them
        // when they include some variables either to determine which loop entry
        // we failed on or some input to the calculation that led to the failure
        // but having them doesn't detract from the test usability so you're
        // welcome to keep them if you find them useful
        expect(ptData.x).toBe('140.72');
    });
});
```

src/plots/gl3d/scene.js
```diff
@@ -58,10 +58,10 @@ function render(scene) {
     }
 
     function formatter(axisName, val) {
-        if(typeof val === 'string') return val;
-
         var axis = scene.fullSceneLayout[axisName];
-        return Axes.tickText(axis, axis.c2l(val), 'hover').text;
+        var calendar = trace[axis._id.charAt(0) + 'calendar'];
```
alexcjohnson: don't we actually want to use the axis calendar for this? (which is what you get if you omit calendar as an arg to `d2l`)

src/plots/gl3d/scene.js
```diff
@@ -58,10 +58,10 @@ function render(scene) {
...
+
+        return Axes.tickText(axis, axis.d2l(val, 0, calendar), 'hover').text;
```
alexcjohnson: I don't see a mock with category data in 3D, though it does work:
```js
Plotly.newPlot(gd,[{x:[1,2,3],y:['a','b','c'],z:['2014-01-01','2016-01-01','2015-01-01'], type:'scatter3d'}])
```
I *think* this code is compatible with category data, but can you include a test for it?

etpinard: This one here 

https://github.com/plotly/plotly.js/pull/1414/files#diff-c406dc5b6d20f7f95206516178b5be36R148

should do it.

alexcjohnson: Oh perfect. My turn for new 👓

src/plots/gl3d/scene.js
etpinard: So, if I understand correctly: hover labels should show dates with the calendar of the axes (as opposed to the calendar of the trace)?

alexcjohnson: Yes. That's what we do in 2D, and I think it makes sense, particularly when you're plotting different traces each with its own calendar, you want to be able to compare them within the system they were plotted on, irrespective of what the source data looks like.

test/jasmine/tests/gl_plot_interact_test.js
etpinard: https://github.com/plotly/plotly.js/pull/1414/commits/e9e0c833818d43299a761472ed80ebbc0ebe55c7 - better?

src/plots/gl3d/scene.js
etpinard: https://github.com/plotly/plotly.js/pull/1414/commits/cbf1f8627451308042b1ac845c2a1b2a004f023a - looks ok?

alexcjohnson: Perfect - thanks for the test.

alexcjohnson:
💃 💃

Issue #1206, opened 2016-11-29
Opacity for 3D plots doesn't behave as expected


cldougl:
Originally posted in: https://github.com/plotly/plotly.py/issues/610

Hello,
When plotting a 3D surface using iplot, the opacity parameter does not behave continuously. Setting it to 1.0 will cause full opacity, and setting it to 0.999999 will cause the surface to be VERY transparent, somewhere around 0.5 opacity. It will behave relatively linearly below this point.

Expected behavior: The opacity parameter should be continuous. A tiny change in the parameter value should be imperceptible.

http://codepen.io/plotly/pen/QGqavK

etpinard:
cc @dfcreative

PR #1415, opened 2017-02-24
Fix gl3d surface opacity scaling

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1206

Strangely, trace opacity was scaled by `0.25` before being passed to `gl-surface3d` :

[image](https://github.com/plotly/plotly.js/blame/master/src/traces/surface/convert.js)

Removing this scaling gives:

gifrecord_2017-02-24_120430

which I believe looks like the desired behavior @cldougl can you confirm?


cldougl:
@etpinard yep exactly! Now the decrease appears to be linear/increments as expected.

Issue #1391, opened 2017-02-17
Hover label disappear after dragging axes on this graph  


etpinard:
See report https://community.plot.ly/t/tooltip-is-not-showing-when-we-move-y-axis/3387

etpinard:
To reproduce:

- go to https://jsfiddle.net/1yLavejg/30/
- drag any axes
- hover over data, hover labels don't show up.

etpinard:
Looks like the problem is only showing when one axis has `fixedrange: true`.

PR #1417, opened 2017-02-24
Fix for xyz column heatmap trace 'text' on hover

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1391

### DIFF
--- a/src/traces/heatmap/hover.js
+++ b/src/traces/heatmap/hover.js
@@ -96,8 +96,8 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode, contour)
     if(zmask && !zmask[ny][nx]) zVal = undefined;
 
     var text;
-    if(Array.isArray(trace.text) && Array.isArray(trace.text[ny])) {
-        text = trace.text[ny][nx];
+    if(Array.isArray(cd0.text) && Array.isArray(cd0.text[ny])) {
+        text = cd0.text[ny][nx];
     }
 
     return [Lib.extendFlat(pointData, {
--- a/src/traces/heatmap/calc.js
+++ b/src/traces/heatmap/calc.js
@@ -121,7 +121,7 @@ module.exports = function calc(gd, trace) {
         Axes.expand(ya, yArray);
     }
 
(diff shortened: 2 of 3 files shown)


src/traces/heatmap/hover.js
```diff
@@ -96,8 +96,8 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode, contour)
     if(zmask && !zmask[ny][nx]) zVal = undefined;
 
     var text;
-    if(Array.isArray(trace.text) && Array.isArray(trace.text[ny])) {
-        text = trace.text[ny][nx];
+    if(Array.isArray(cd0.text) && Array.isArray(cd0.text[ny])) {
+        text = cd0.text[ny][nx];
```
etpinard: as for xyz heatmaps, `fullData[i].trace.text` is 1D, whereas `gd.calcdata[i][0].text` is 2D.

Previously, `fullData[i].trace.text` was converted to be 2D during the calc step, but since the calc step isn't called on `relayout(gd, 'xaxis.range', [/* */])`, `text` on this line here was `undefined` after zoom interactions as discovered in https://github.com/plotly/plotly.js/issues/1391.

alexcjohnson: wow, nice detective work!

alexcjohnson: This looks fine as a bugfix, though I would think the more robust solution would not alter `fullData` at all, and *only* keep the re-dimensioned text array in `calcdata`.

etpinard: > and only keep the re-dimensioned text array in calcdata.

I agree 100%. But, I wanted to keep this patch to a minimum as @rreusser's carpet PR already alters [`heatmap/convert_column_xyz.js`](https://github.com/plotly/plotly.js/pull/1239/files#diff-9f910383868c78bde87fcb8d3905fe91R14).

I'll make a comment there.

alexcjohnson: Ah good thinking. :dancer:

Issue #1418, opened 2017-02-24
scattermapbox lines break with trailing null


rreusser:
The following trace breaks scattermapbox:

```javascript
{
  type: 'scattermapbox',
  mode: 'lines+markers',
  lat: [0, 10, null, 20, 30, null],
  lon: [10, 20, null, 30, 40, null],
  connectgaps: false
}
```

The culprit is the trailing null. The result is a generic mapbox error and no lines for that particular trace. It fails identically given more than one trailing null.

Example: http://codepen.io/rsreusser/pen/VpYjON?editors=1010

<img width="536" alt="nulls" src="https://cloud.githubusercontent.com/assets/572717/23320385/88d452f6-faa8-11e6-823c-0f86456d7a21.png">

etpinard:
Thanks for the report!

PR #1421, opened 2017-02-24
Scattermapbox line trailing gaps

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1418

easy. cc @rreusser

### DIFF
--- a/test/jasmine/tests/scattermapbox_test.js
+++ b/test/jasmine/tests/scattermapbox_test.js
@@ -368,7 +368,7 @@ describe('scattermapbox convert', function() {
     it('for lines traces with trailing gaps', function() {
         var opts = _convert(Lib.extendFlat({}, base, {
             mode: 'lines',
-            lon: [10, '20', 30, 20, null, 20, 10, null],
+            lon: [10, '20', 30, 20, null, 20, 10, null, null],
             lat: [20, 20, '10', null, 10, 10, 20, null]
         }));
 
--- a/test/jasmine/tests/scattermapbox_test.js
+++ b/test/jasmine/tests/scattermapbox_test.js
@@ -259,7 +259,7 @@ describe('scattermapbox convert', function() {
         lat: [20, 20, '10', null, 10, 10, 20]
     };
 
-    it('for markers + circle bubbles traces, should', function() {
+    it('should generate correct output for markers + circle bubbles traces', function() {
         var opts = _convert(Lib.extendFlat({}, base, {
             mode: 'markers',
             marker: {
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/scattermapbox_test.js
```diff
@@ -365,6 +365,23 @@ describe('scattermapbox convert', function() {
...
+            lon: [10, '20', 30, 20, null, 20, 10, null],
+            lat: [20, 20, '10', null, 10, 10, 20, null]
+        }));
+
+        assertVisibility(opts, ['none', 'visible', 'none', 'none']);
+
+        var lineCoords = [
+            [[10, 20], [20, 20], [30, 10]],
+            [[20, 10], [10, 20]]
+        ];
```
etpinard: Previously, this was:

```js
var lineCoords = [
  [[10, 20], [20, 20], [30, 10]],
  [[20, 10], [10, 20]].
  []
];
```

which an invalid GeoJSON and made `mapbox-gl` fail.

test/jasmine/tests/scattermapbox_test.js
```diff
@@ -365,6 +365,23 @@ describe('scattermapbox convert', function() {
         expect(actualText).toEqual(['A', 'B', 'C', 'F', '']);
     });
 
+    it('for lines traces with trailing gaps', function() {
```
alexcjohnson: same comment as https://github.com/plotly/plotly.js/pull/1414#discussion_r103071216 - is this an intentional pattern now?

test/jasmine/tests/scattermapbox_test.js
rreusser: Great. I thought this would be a pretty easy fix. Just to confirm, I assume it will also handle `[10, '20', 30, 20, null, 20, 10, null, null]` without problems?

etpinard: > I assume it will also handle [10, '20', 30, 20, null, 20, 10, null, null] without problems?

Yes it does: https://github.com/plotly/plotly.js/pull/1421/commits/1a22bebd1776b2d70658ef3c4101655c3839912e

test/jasmine/tests/scattermapbox_test.js
etpinard: https://github.com/plotly/plotly.js/pull/1421/commits/24d7c47f2ed07b8ca7409fb9fb57c65175c72e47 - better?

alexcjohnson: Yes, much better to my eye - thanks!

Issue #1368, opened 2017-02-09
mesh3d: Different lighting results with different hardware


jackparmer:
Consider this codepen:

http://codepen.io/monfera/full/ggByRW

On my 12 in Macbook, I see this:

![image](https://cloud.githubusercontent.com/assets/1865834/22805728/f5e07f68-eeec-11e6-8cfc-55b196a9915b.png)

@etpinard 's Thinkpad and @monfera 's Macbook show this:

![image](https://cloud.githubusercontent.com/assets/1865834/22805737/0535d7ce-eeed-11e6-946e-c0020f0b13be.png)

Another example - consider this codepen:

http://codepen.io/monfera/pen/XjxZjZ

Result on my 12 in Macbook:

![image](https://cloud.githubusercontent.com/assets/1865834/22805761/220d896e-eeed-11e6-8754-62c483b32a11.png)

Result on larger Macbooks and Thinkpad:

![image](https://cloud.githubusercontent.com/assets/1865834/22805787/3d66c130-eeed-11e6-9f1a-fad66913da2e.png)

***

I tried this in Chrome, Safari, and Firefox with the same result.

Computer specs:

![image](https://cloud.githubusercontent.com/assets/1865834/22805815/53162ab6-eeed-11e6-9ed2-ec1b417fdbb1.png)

Chrome specs:

![image](https://cloud.githubusercontent.com/assets/1865834/22805831/6378ab72-eeed-11e6-967b-bde5a64b9c47.png)

rreusser:
Hmm… looks like flipped normals to me. I'd suggest maybe `gl_FrontFacing` isn't properly supported, but 2016 seems pretty recent for that.

http://stackoverflow.com/questions/24375171/is-there-a-reliable-alternative-to-gl-frontfacing-in-a-fragment-shader

etpinard:
cc @dfcreative which have insight about this.

monfera:
I don't have access to this machine but guessing is always fun:

- perhaps the `gl` floats are `lowp` - I think the standard is quite permissive about precision
- or we're using an extension that silently causes an issue

One thing to try: visiting a bunch of WebGL content on the web to see if *those* work properly.

jackparmer:
Quick update: It doesn’t seem to matter what value I set for `diffuse` or `specular`. I can set them to 0.1 or 1 or 1000 and nothing changes.

monfera:
Do these work fine? http://webglsamples.org/

[...]

PR #1423, opened 2017-02-27
Better face normals for mesh3d

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1368

via @dfcreative 's https://github.com/gl-vis/gl-mesh3d/pull/8 which now appears to behave well on iOS, Windows, Ubuntu (including on CircleCI).

Note the baselines diffs though. The new algorithm clearly changes the look of the 3D meshes. Is that for better or for worse? cc @monfera @rreusser @alexcjohnson @jackparmer


alexcjohnson:
I wouldn't say it's better or worse, but sufficiently different to raise an alarm - it looks like the light source is now coming from exactly the opposite direction? Have we reversed a sign somewhere?

rreusser:
It looks like the [terminators](https://en.wikipedia.org/wiki/Terminator_(solar)) are *exactly* the opposite in the snowden example so that yeah, maybe a flipped sign?

dfcreative:
What is the chance that test images had wrong sign all the time due to rendering in Ubuntu?
I inverted sign in https://github.com/gl-vis/gl-mesh3d/pull/8/commits/c0fb6bcdc41ef9c55787556b925bcfc761327a97, but that breaks own `gl-mesh3d` tests

monfera:
I have a vague recollection of counterintuitive light direction, although I was using OS X, and it was one specific thing (Snowden). I'm wondering if the polygon (triangle) winding order might be anything to do with it, i.e. maybe with the Snowden the triangles are clockwise when seen from the 'outside', and counterclockwise for the others (or the other way around). Just a random thought in a semi-slumber...

alexcjohnson:
If we can clearly demonstrate that the previous behavior is a bug then I suppose there's no problem cementing the new behavior. Presumably the default light location was chosen, in the context of the buggy interpretation, to be some standard-ish above-and-to-the-side illumination for the default camera angle? Should we then alter the default lighting at the same time as we fix the bug? Then the practical results would be limited to those folks who've explicitly set a light source - admittedly exactly the folks who are most likely to care, but a bug needs to be fixed...

If there's any chance the winding order affects any of this, we need to sort that out and fix it first.

rreusser:
For debuggability and my own sake, here's a cube with outward facing normals (at least to the extent that it follows the right hand rule per-face): http://codepen.io/rsreusser/pen/PpPJWW?editors=0010

I'm actually kind of confused what lightposition does. It's clearly in screen space or something bt such that it's not totally clear to me what space the light is oriented in.

etpinard:
The new baselines are here https://github.com/plotly/plotly.js/pull/1423/commits/db6532c505614322fc0406f649f49758e7bf3205

Looking a lot better:

image

etpinard:
@dfcreative one possible red flag: the cube baseline now looks a bit darker

gifrecord_2017-03-01_135302

Is that the expected bahavior? It is just an Ubuntu / nw.js artefact?

etpinard:
@dfcreative any news on my previous :arrow_up: https://github.com/plotly/plotly.js/pull/1423#issuecomment-283434751 ?

I would :heart: to merge this PR in the next few days.

dfcreative:
@etpinard I'll try to look at it tonight, but that seems to be not the only issue. It seems to be snowden' model issue.

dfcreative:
Turns out that default cube setup has broken lighting:
image
That is also true for the [API demo](https://plot.ly/javascript/3d-mesh/#3d-mesh-cube):
image
The @monfera's guess is right, the triangles are in wrong order of vertices. Is there any theoretical background for the order or vertices? Maybe we should sort vertices order before use? Out of ideas for now.

rreusser:
The rule for triangle normals is just the right hand rule. I guess to fix it you'd have to build adjacency data, decide which direction is outward, and then flip normals to match the correctly oriented neighbors. Just a guess. There are plenty of edge cases (that's a pun) where trying to fix it automatically would probably do more harm than good though. I think it's totally fair and reasonable to state the required orientation and require correct input.

alexcjohnson:
> decide which direction is outward

so we can't do a klein bottle? why can't both directions be "outward"?

rreusser:
@alexcjohnson Would the solution be to [`abs(*)` instead of `max(0, *)` the diffuse lighting](https://github.com/stackgl/glsl-diffuse-lambert/blob/master/index.glsl#L4)? The downside is that lights would illuminate from the source direction and also the anti-source direction.

alexcjohnson:
> lights would illuminate from the source direction and also the anti-source direction.

That doesn't sound good... the light needs to act like a real light source and come from one side, so it does its job of helping you to see the shape of the data.

Can't we just make the active surface normal be the one you see? Which I guess means the one that has a positive dot product with the vector from somewhere on the face to the camera?

monfera:
I'm elbow deep in some other task now, but wanted to agree that as @rreusser said, working out face direction from global mesh or adjacency data would be prone to all kinds of issues, and as @alexcjohnson said, it's possible to do the lighting such that it's invariant of the (e.g.) right winded interpretation of the face normal, indeed as needed for rendering a _semi-opaque_ Klein bottle. 

I'm not certain how conducive the current lighting algo is to this (pls. correct me if I'm wrong but it's the diffuse lighting only i.e. not the specular), and I also found e.g. during the brain mesh work that semitransparent rendering didn't quite work out. So due to this we can't render a convincing semi-opaque Klein bottle.

Most likely unrelated, but there's also the possibility of [face culling](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/cullFace) - I suppose we don't cull back faces as otherwise the triangle would be dropped. To render semi-opaque Klein bottles there cannot be face culling.

etpinard:
Hmm. Sounds this issue is from being resolved.

@jackparmer ok to drop this from the `1.25.0` release?

dfcreative:
@etpinard the point is that current [gl-mesh3d](https://github.com/gl-vis/gl-mesh3d/) is not less broken than [gl-mesh3d](https://github.com/gl-vis/gl-mesh3d/pull/8), it is consistent across platforms though, so

etpinard:
Good point @dfcreative 

Any opposition to merging this?

monfera:
Yes I think it's a big improvement!

etpinard:
New baselines with the latest gl-mesh3d commit https://github.com/gl-vis/gl-mesh3d/pull/8/commits/63dba3585a1f73be9c0e18ce41a3704a77f10778 are in https://github.com/plotly/plotly.js/pull/1423/commits/26751ec9b0b100e21d77fa0a42b4c2545fdf7f91 which are essentially the same as from the time of this https://github.com/gl-vis/gl-mesh3d/pull/8#issuecomment-282085143

I'm looking for some much needed :+1:s before going ahead and merging this thing.

alexcjohnson:
😿  the intermittent CI failure I saw [yesterday](https://circleci.com/gh/plotly/plotly.js/3794) is [here](https://circleci.com/gh/plotly/plotly.js/3840) again today:
```
TypeError: unsupported file type: undefined (file:
/var/www/streambed/image_server/plotly.js/build/test_images/gl3d_z-range.svg)
```

alexcjohnson:
Snowden is still lit from the opposite side as before this change. The others too, probably, but the rest are hard to say anything about except "they changed."

So I'll ask again: which is correct and which is a bug? I understand there may be other advantages to this change, but we cannot just flip the lighting on all of our 3d surfaces without clearly determining that the previous behavior was a bug.

rreusser:
FYI: I just learned about this function: http://docs.gl/el3/faceforward

monfera:
@rreusser neat! looks like a WebGL 2 thing, or maybe already prevalent?

rreusser:
@monfera it says OES 1.0, which *should* be webgl 1 unless I'm mistaken. Which isn't unlikely.

monfera:
@rreusser cool, I just saw the GLSL ES 3.1 label on top, but indeed, 1.0 is listed below

rreusser:
@alexcjohnson I'll see if I can phrase it as a challenge. Can someone demonstrate clear behavior/intent with the lighting position in this codepen? http://codepen.io/rsreusser/pen/PpPJWW?editors=0010

It's a cube with normals just averaged to point outward so that it gets shaded like a sphere. Reverse engineering, here's what I know:

- `(x, y, z)` seems to be in screen space.
  - `(x,y,z) = (-100000, 0, 0)` places the light to the left, which makes sense
  - `(x,y,z) = (100000, 0, 0)` places the light to the right, which makes sense
  - `(x,y,z) = (0, 100000, 0)` lights from above
  - `(x,y,z) = (0, -100000, 0)` lights from below
  - I don't know quite why the numbers have to be so big. Are the positions in screen pixels?
  - I can't figure out where `z` puts the lighting

Of course the real approach should be to debug the code/intent itself, but this is my understanding from treating it as a black box, which is what users would do.

rreusser:
After digging into the behavior quite a bit with @alexcjohnson, here's what we conclude:

The current behavior mostly make sense. The really weird stuff is all in the light positioning.

- the magnitudes have to be really giant. Are they pixels or something?
- z exhibits really weird behavior and thresholds

Apart from how it does work, here's how we think it should work:

The light should be positioned effectively in clip coordinates with the addition of an aspect ratio. Conceptual test cases:

- a light position of `[1, 0, 0]` should be just on the right side of the screen (*at the depth of the center*, that is. the projection matrix makes this statement otherwise meaningless). It involves an (inverse?) view matrix somewhere in there, I think.
- a light position of `[10, 0, 0]` would be roughly speaking 10 half-screens worth off to the right (again, at the depth of the center point relative to the eye)
- a light position of `[0, 1, 0]` should be at the top of the screen. If the plot is not square, then one or the other of these will need to take into account an aspect ratio.
- a light position of `[1, 1, 0]` should be at 45 degrees NE regardless of the plot's aspect ratio. This constraint probably determines where the aspect ratio goes
- a light position of `[0, 0, 1]` should be more or less *at* the eye position, unless I'm wrong about scales. At the very least in the direction of the eye. Out of the screen instead of into because of the righthand rule.
- a light position of `[0, 0, 10]` should be way behind the camera
- If we can, it'd be great of `length(light) > 1` implied the light tended not to fall *inside* the screen (I'm not 100% convinced this can be satisfied subject to the other constraints)

Open to feedback, but I think these constraints are enough to minimally fix the parts of the current behavior that are surprising without changing anything substantial. In talking with @alexcjohnson one conclusion is that existing behavior like the weird 1.00002 -> 1.00003 z threshold is best just fixed so that it's probably best *not* to preserve strict backward compatibility.

etpinard:
Dropping from the `1.25.0`. 

To minimize the impact on our users, we should address @rreusser 's `lightpositon` https://github.com/plotly/plotly.js/pull/1423#issuecomment-287411731 proposal in the same release as @dfcreative 's hardware-dependent fixes done in this PR.

etpinard:
Closing this thing for now.

I'll copy https://github.com/plotly/plotly.js/pull/1423#issuecomment-287411731 in https://github.com/plotly/plotly.js/issues/1368 . Future development should refer to https://github.com/plotly/plotly.js/issues/1368

PR #1427, opened 2017-02-28
Show correct curve number in gl2d hover data

etpinard:
Currently, the `curveNumber` field in the gl2d hover data corresponds to the index in `fullData` after `visible: false` are filtered out. That is, `curveNumber` does not match the index in the user data array - as it should and does for all other subplot types. This PR fixes this.


alexcjohnson:
Easy enough 💃

PR #1431, opened 2017-03-01
Axes ref cleanup

etpinard:
No more `ax_gd` !!!


src/plots/cartesian/layout_defaults.js
```diff
@@ -115,18 +115,34 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
     var bgColor = Color.combine(plot_bgcolor, layoutOut.paper_bgcolor);
 
-    var axLayoutIn, axLayoutOut;
+    var axName, axLayoutIn, axLayoutOut;
 
     function coerce(attr, dflt) {
         return Lib.coerce(axLayoutIn, axLayoutOut, layoutAttributes, attr, dflt);
     }
 
-    axesList.forEach(function(axName) {
```
etpinard: as promised in https://github.com/plotly/plotly.js/pull/1261#discussion_r96523260

src/plots/plots.js
```diff
@@ -491,12 +491,10 @@ plots.supplyDefaults = function(gd) {
     // TODO may return a promise
     plots.doAutoMargin(gd);
 
-    // can't quite figure out how to get rid of this... each axis needs
-    // a reference back to the DOM object for just a few purposes
+    // set scale after auto margin routine
     var axList = Plotly.Axes.list(gd);
     for(i = 0; i < axList.length; i++) {
         var ax = axList[i];
-        ax._gd = gd;
```
etpinard: @alexcjohnson this commit is very important.

First, no need to keep `ax_gd` in-sync anymore!

src/plots/cartesian/set_convert.js
```diff
@@ -61,7 +61,8 @@ function num(v) {
  * also clears the autorange bounds ._min and ._max
  * and the autotick constraints ._minDtick, ._forceTick0
  */
-module.exports = function setConvert(ax) {
+module.exports = function setConvert(ax, fullLayout) {
+    fullLayout = fullLayout || {};
```
etpinard: Second, make `setConvert` take in `fullLayout`

src/plots/cartesian/set_convert.js
```diff
@@ -319,7 +320,7 @@ module.exports = function setConvert(ax) {
 
     // set scaling to pixels
     ax.setScale = function(usePrivateRange) {
-        var gs = ax._gd._fullLayout._size,
+        var gs = fullLayout._size,
```
etpinard: ... so that `setScale` can look into `_size`.

src/plots/cartesian/set_convert.js
```diff
@@ -360,7 +361,6 @@ module.exports = function setConvert(ax) {
             Lib.notifier(
                 'Something went wrong with axis scaling',
                 'long');
-            ax._gd._replotting = false;
             throw new Error('axis scaling');
```
etpinard: :hocho: `ax._gd._replotting = false;` on axis scaling errors. Isn't that error throw enough?

src/plots/cartesian/set_convert.js
```diff
@@ -417,6 +417,10 @@ module.exports = function setConvert(ax) {
     ax._min = [];
     ax._max = [];
 
+    // copy ref to fullLayout.separators so that
+    // methods in Axes can use it w/o having to pass fullLayout
+    ax.separators = fullLayout.separators;
```
etpinard: **Important:** link `fullLayout.separators` to full axis object, so that `Axes` method can use it without having to know anything about `fullLayout`.

src/plots/cartesian/layout_defaults.js
```diff
@@ -164,13 +169,7 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
         handlePositionDefaults(axLayoutIn, axLayoutOut, coerce, positioningOptions);
 
-        layoutOut[axName] = axLayoutOut;
-
-        // so we don't have to repeat autotype unnecessarily,
-        // copy an autotype back to layoutIn
-        if(!layoutIn[axName] && axLayoutIn.type !== '-') {
-            layoutIn[axName] = {type: axLayoutIn.type};
-        }
+        axLayoutOut._input = axLayoutIn;
```
etpinard: This is essentially the new `ax_gd`. This pattern is similar to other components that require a ref to the user container to mutate it on interaction (e.g. updatemenus and sliders). Moreover, IMHO this pattern is a lot better than having to keep the full `gd` in-sync.

src/plots/cartesian/axes.js
```diff
@@ -328,14 +328,10 @@ axes.doAutoRange = function(ax) {
...
 
-        if(!axIn) ax._gd.layout[ax._name] = axIn = {};
-
-        if(axIn !== ax) {
-            axIn.range = ax.range.slice();
-            axIn.autorange = ax.autorange;
-        }
+        var axIn = ax._input;
+        axIn.range = ax.range.slice();
+        axIn.autorange = ax.autorange;
```
etpinard: Now, `doAutorange` write in `ax._input` instead of having dig in `_gd.layout ...` :tada:

src/plots/cartesian/axes.js
```diff
@@ -1137,7 +1133,7 @@ axes.tickText = function(ax, x, hover) {
 };
 
 function tickTextObj(ax, x, text) {
-    var tf = ax.tickfont || ax._gd._fullLayout.font;
+    var tf = ax.tickfont || {};
```
etpinard: Oh. I need to double-check this. 

Removing the `|| {}` should do the trick - as axis `tickfont` already default to layout `font`, but removing `|| {}` makes the [`fonts.json`](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/fonts.json) mock fail. Not sure why.

src/plots/cartesian/layout_defaults.js
alexcjohnson: There are still [two](https://github.com/plotly/plotly.js/pull/1431/files#diff-1602f9e66d27cf60c7469f31f03b8053R127) [maps](https://github.com/plotly/plotly.js/pull/1431/files#diff-1602f9e66d27cf60c7469f31f03b8053R136) and a [filter](https://github.com/plotly/plotly.js/pull/1431/files#diff-1602f9e66d27cf60c7469f31f03b8053R133) 🐎

src/plots/cartesian/set_convert.js
alexcjohnson: If we think the error is fatal to the plot then yes, it's enough. But if there would be a use case for catching the error and continuing, then the plot will be in an odd state without resetting `gd._replotting`. I suppose at the very least we can move it to `fullLayout` but we should really figure out a better solution than the fragile `_replotting` flag to prevent infinite recursion.

src/plots/cartesian/set_convert.js
alexcjohnson: shall we call it `_separators` to distinguish it from regular settings managed by `supplyDefaults`? That actually adds overhead (as `relinkPrivateKeys` will look at it) but it does seem to be the pattern we've been using.

src/plots/gl3d/scene.js
```diff
@@ -708,4 +705,11 @@ proto.toImage = function(format) {
     return dataURL;
 };
 
+proto.setConvert = function() {
+    for(var i = 0; i < 3; ++i) {
+        var ax = this.fullSceneLayout[axisProperties[i]];
+        Axes.setConvert(ax, this.fullLayout);
+    }
+};
```
alexcjohnson: Nice simplification! Why don't we need the [`containerOut.setScale = Lib.noop;`](https://github.com/plotly/plotly.js/pull/1431/files#diff-dd23546f2ee4a55150cd584d026e1c93L18) line anymore?

test/jasmine/karma.conf.js
```diff
@@ -35,6 +35,14 @@ function func(config) {
     //     See CONTRIBUTING.md for additional notes on reporting.
     func.defaultConfig.logLevel = config.LOG_INFO;
 
+    // without this, console logs in the plotly.js code don't print to
+    // the terminal since karma v1.5.0
+    //
+    // See https://github.com/karma-runner/karma/commit/89a7a1c#commitcomment-21009216
+    func.defaultConfig.browserConsoleLogOptions = {
+        level: 'log'
+    };
```
alexcjohnson: Oh awesome - that'll be super helpful.

src/plots/cartesian/set_convert.js
etpinard: Good call here. I was thinking maybe down the road we would add per-axis separators, making this line obsolete (by supply defaults logic). But no, you're right,  non-underscore keys in containers should match input attributes.

src/plots/cartesian/set_convert.js
etpinard: Sounds good. I'll try to cook up a test case for this edge case. Any help would be appreciated ;)

src/plots/gl3d/scene.js
etpinard: Not anymore. :hocho:

etpinard: To be more precise, 3d doesn't use any of the _m and friends fields that setScale computes at the moment. So setting setScale to a noop might be a minor :racehorse: boost. But, for 3d annotations, I'm planning on using / mocking those scale fields. To be continued.

src/plots/cartesian/layout_defaults.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1431/commits/5eb62027c58fdcde38d732ba01368849d885a4d3

src/plots/cartesian/set_convert.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1431/commits/6e25640adeac74d6cc7ec9847a5a74a8daff40a9

src/plots/cartesian/set_convert.js
etpinard: done in 96125490b513498b4eec7a7efd9231e54eb09535

alexcjohnson:
Thanks for the changes. I guess you're still sorting something out re: [fonts](https://github.com/plotly/plotly.js/pull/1431#discussion_r103795795) but I'm ready to 💃

src/plots/cartesian/axes.js
etpinard: After investigation, I'll keep the `|| {}`, if ok.

Whenever `ax.showticklabels: false`, [`ax.tickfont`](https://github.com/plotly/plotly.js/blob/954da8a806f814798cf6cbffa30087f9ca1d43d2/src/plots/cartesian/tick_label_defaults.js#L28-L38) isn't coerced and isn't defined here.

We could make sure that `tickTextObj` isn't called when `ax.showticklabels: false`, but this requires adding a few early returns in axes.js, 3d and gl2d convert routines which I believe isn't worth it.

alexcjohnson: 👍

PR #1440, opened 2017-03-03
Clear parcoords canvas specially if no panels exist

monfera:
Reason: if there's no panel, then the panel redaw loop will not enter the loop body that also clears the canvas, causing uncleared lines when only 0 or 1 dimension exists (as visible dimensions) after rendering, then making other dimensions go invisible with `Plotly.restyle`.


etpinard:
Great! Can you add a test case to :lock: this down?

etpinard:
This fix would be nice addition to the next patch release `1.24.1`. @monfera have you had the time to think about a way to add a test case?

monfera:
@etienne my conclusion from our chat is that we'll need to persist the regl object on the <canvas> element; it'll make it possible to have access to it for testing. I can go ahead with it today.

monfera:
@etpinard specifically, we already bind (with D3) the lineLayer object that has a readPixel method for picking - I'll just make a similar method for reading all pixels in a rectangle, so no need for the entire regl object.

etpinard:
> specifically, we already bind (with D3) the lineLayer object that has a readPixel method for picking - I'll just make a similar method for reading all pixels in a rectangle, so no need for the entire regl object.

Sounds good. Thanks!

Issue #1267, opened 2016-12-22
scatter3d traces don't respect trace order when toggled 


etpinard:
See http://codepen.io/etpinard/pen/oYOMZO

![gifrecord_2016-12-22_143738](https://cloud.githubusercontent.com/assets/6675409/21437873/48c19b00-c854-11e6-9963-bd30981bfdd7.gif)

etpinard:
cc @dfcreative

alexcjohnson:
#1466 at least makes the behavior not be path-dependent, so that's good. But there's a bigger issue here: points aren't ordered by depth across traces. The trace that was drawn last always has all its points in front of the other traces, even if they are really farther away. It turns out this is only an issue for traces with opacity. If I change the traces in http://codepen.io/etpinard/pen/oYOMZO to `opacity: 1` I see a mixture of blue and grey throughout, not one color dominant throughout the dense middle part:

![screen shot 2017-03-13 at 4 58 35 pm](https://cloud.githubusercontent.com/assets/2678795/23875549/3daa3ef6-0810-11e7-8a2c-90daecd53c65.png)

This seems to me frankly a major bug for partly-transparent traces. It means the essential 3d-ness of the plot is silently lost, making you think the last trace drawn has a greater extent than the others or something...

rreusser:
[Painter's algorithm](https://en.wikipedia.org/wiki/Painter's_algorithm) is one solution, but that means view-dependent sorting. Unfortunately [order-independent transparency](https://en.wikipedia.org/wiki/Order-independent_transparency) tends to be pretty nontrivial to the best of my knowledge. 😞

dfcreative:
Seems that we have to do single gl-scatter3d render pass for accumulated data, logically. Trying to figure out how are we doing that now.

etpinard:
cc @archmoj

archmoj:
Issue example displayed in a candidate patch: [codepen](https://codepen.io/MojtabaSamimi/pen/XvEXaB?editors=1000).

brianpenghe:
I also have this issue when having opacity = 1. Any solutions?

drscotthawley:
Wow, 4 years old and still no solution.   How can we help? 

Currently seeing this: If I start with the first figure and rotate it a bit toward the right such that the blue points should be in front of the brown points,.... the blue points disappear completely ?? 

[...]

PR #1444, opened 2017-03-03
scattergl lazy init

etpinard:
Instead of initialising 1 `gl-line2d`, 2 `gl-error2d`, 1 `gl-scatter2d` and 1 `gl-scatter2d-(fancy|sdf)` object per `scattergl` why not intialize only the part(s) we need?

On a 5-trace graph of 10,000 pts each, that amounts to a ~300ms-faster `plot` call :racehorse: 

image


src/traces/scattergl/convert.js
```diff
@@ -264,6 +290,15 @@ proto.update = function(options) {
         this.updateFast(options);
     }
 
+    // sort objects so that order is preserve on updates:
+    // - lines
+    // - errorX
+    // - errorY
+    // - markers
+    this.scene.glplot.objects.sort(function(a, b) {
```
etpinard: This part isn't quite right. We'll have to consider trace order too, fix coming on Monday.

The same technique might be extended to 3D to fix https://github.com/plotly/plotly.js/issues/1267

etpinard: done in 81d3bf68d483c2d702a7108026893745a09baaf3

src/traces/scattergl/convert.js
```diff
@@ -100,16 +97,47 @@ function LineWithMarkers(scene, uid) {
...
+    var obj = null;
+
+    function update() {
+        if(!obj) {
+            obj = createFn(glplot, options);
+            obj._trace = _this;
+            obj._index = objIndex;
+        }
+        obj.update(options);
+        return obj;
```
alexcjohnson: what are the `return obj` statements for? Seems like it's meant for method chaining, but then wouldn't it be `return _this`? This way seems dangerous to me since `obj` can be (and very often is) `null`.

alexcjohnson: or no, I guess not `_this`, but the object you create and return [below](https://github.com/plotly/plotly.js/pull/1444/files#diff-202416f5e6d665625197b3b4abf2e142R133)

etpinard: Ha good eyes. I should've :hocho: before pushing. We don't need to support method chaining here (why we anyone clear and/or update and/or dispose in the same statement?) 

:hocho: :hocho: :hocho:

etpinard: done in 34ea31e9d58790928020f4fa7a6b982d6985ae16

alexcjohnson:
💃  nice 🐎 !

Issue #1446, opened 2017-03-06
Latest release leads to infinite loops when calling `restyle` on `plotly_hover`


jayfresh:
Since the latest release, using the hover and unhover events seems to throw a lot of JavaScript errors, ultimately locking up the browser. The easiest way to see this is on the hover event example on the Plotly docs: https://plot.ly/javascript/plotlyjs-events/#hover-event with accompanying codepen here: http://codepen.io/plotly/pen/xgxbrj.

I've tested that this doesn't happen locally for v1.23.1 (and there aren't any "hover" changes in v1.23.2 so I expect that would work, but https://cdn.plot.ly/plotly-1.23.2.min.js is not available).

etpinard:
Thanks for the report.

I suspect https://github.com/plotly/plotly.js/issues/1274 and/or https://github.com/plotly/plotly.js/pull/1376 could have broken your example.

etpinard:
Looks like indeed #1304 is the culprit. cc @rreusser @alexcjohnson 

@jayfresh We're currently working on a fix that will be part of `v1.24.1`

In the meantime adding

```js
  var gd = document.getElementById('myDiv')
  gd._hoverdata = undefined
```

to the `'plotly_unhover'` event handler as in http://codepen.io/etpinard/pen/dvpKLw makes things work again.

jayfresh:
:+1:

jayfresh:
Ah, there's still a problem on my app using `v1.24.1`, but I'm not sure where that's coming from exactly. Trying to replicate on a codepen.

dictaartisan:
I see what is probably the same problem (stack overflow, chrome hanging) when drawing/undrawing an annotation on hover and unhover events

Here's a codepen (shamelessly copying etpinard's code)

http://codepen.io/dictaartisan/pen/XRRYwe

Hover in and out of the points often enough and browser will hang - seems to recover after a while in this case but on another more complex plot I have to kill the browser

Linerd:
Same here, on the latest version. Trying to restyle a scatter 3D plot upon hovering and infinite amount of hover events are triggered until maximum stack call is reached.

PR #1448, opened 2017-03-06
Clear `gd._hoverdata` before emitting `plotly_unhover`

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1446

- so that in the case where a `plotly_unhover` handler calls a Plotly method (restyle, relayout and update which in turn call [`Plots.rehover`](https://github.com/plotly/plotly.js/blob/05c576c87a561da1879b734cf90449c190ccd552/src/plots/plots.js#L2042-L2046) added in https://github.com/plotly/plotly.js/pull/1304) an infinite loop :curly_loop:  isn't triggered (see https://github.com/plotly/plotly.js/issues/1446 for example)
- add test case showing the current infinite loop code path

cc @rreusser @alexcjohnson

### DIFF
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -886,4 +886,62 @@ describe('hover updates', function() {
             return assertLabelsCorrect(null, [103, 100], 'trace 10.5');
         }).catch(fail).then(done);
     });
+
+    it('should not trigger infinite loop of plotly_unhover events', function(done) {
+        var gd = createGraphDiv();
+        var colors0 = ['#00000', '#00000', '#00000', '#00000', '#00000', '#00000', '#00000'];
+
+        function unhover() {
+            return new Promise(function(resolve) {
+                mouseEvent('mousemove', 394, 285);
+                setTimeout(function() {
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/hover_label_test.js
```diff
@@ -886,4 +886,62 @@ describe('hover updates', function() {
...
+            expect(unHoverCnt).toEqual(1);
+
+            return assertLabelsCorrect([400, 200], [435, 198], '3');
+        })
+        .then(unhover)
+        .then(function() {
+            expect(hoverCnt).toEqual(2);
+            expect(unHoverCnt).toEqual(2);
+        })
+        .then(done);
```
etpinard: This test case led to 

image

before the above `dragelement` patch.

alexcjohnson:
Crazy. I certainly would not have caught that, but it makes sense. 💃 

Sometime it might be worth making test cases corresponding to some of our other event-driven examples, both to test the events themselves and to guard against other interactions like this. The Mandelbrot explorer I made is one that comes to mind - where a zoom (relayout) event triggers a restyle (with new heatmap data)

PR #1450, opened 2017-03-07
bump dev deps

etpinard:
featuring most notably some [`uglify-js`](https://github.com/mishoo/UglifyJS2/compare/v2.7.5...v2.8.8) improvements that will trim our minified bundle size by a few bytes in future releases.


etpinard:
Very interestingly, bumping `uglify-js` to `2.8.8` makes some of our gl3d image tests fail

image

 gl3d requires `uglify-js` via the [`cwise`](https://github.com/scijs/cwise/blob/35f89e209b9cf024857149cf55943de55c12d840/lib/cwise-transform.js#L7) browserify transform.

-----------

Pinging scijs pros @rreusser and @dfcreative if you can think of a reason why. See git-compare [here](https://github.com/mishoo/UglifyJS2/compare/v2.7.5...v2.8.8).

etpinard:
Putting this PR on hold until we find the root of the `cwise` + `uglify-js@2.8.x` problem.

etpinard:
PR sent upstream :arrow_right_hook:  https://github.com/scijs/cwise/pull/13

PR #1451, opened 2017-03-07
fix removal of last shape or annotation

alexcjohnson:
Removing the last shape or annotation led to an error in `drawOne`. Images don't have `drawOne` so are immune to this problem. Ironically, images are also the only one that previously had a test for it. Someday we go d3-idiomatic and get rid of `drawOne` entirely...

cc @etpinard - there's one more bug I've uncovered while testing in the workspace, will make another PR for it then may need yet another patch release.


src/components/annotations/draw.js
```diff
@@ -69,7 +69,7 @@ function drawOne(gd, index) {
     fullLayout._infolayer.selectAll('.annotation[data-index="' + index + '"]').remove();
 
     // remember a few things about what was already there,
-    var optionsIn = layout.annotations[index],
+    var optionsIn = (layout.annotations || [])[index],
```
etpinard: Ha. I'm looking forward to making annotations and shapes update _d3-idiomatic_. These user / full container double standard headaches will be gone!

test/jasmine/tests/shapes_test.js
```diff
@@ -320,6 +320,23 @@ describe('Test shapes:', function() {
                 expect(countShapePathsInLowerLayer()).toEqual(0);
                 expect(countShapePathsInSubplots()).toEqual(0);
             })
+            .then(function() {
+                return Plotly.relayout(gd, {'shapes[0]': getRandomShape()});
+            })
+            .then(function() {
```
etpinard: Nice tests! Annotations and shapes are getting nicely :lock: ed down.

etpinard:
:dancer:

Issue #1390, opened 2017-02-17
Overdrawing of images (and probably shapes)


alexcjohnson:
When you combine a data-referenced layer-below image with multiple subplots, you get extra copies of the image (which in certain instances could end up on top even though you asked for below!).
- Open the [layout_image mock](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/layout_image.png) in test_dashboard
- `Plotly.relayout(gd,{plot_bgcolor:'rgba(0,0,0,0)'})
- You'll notice the R logo gets brighter. It's actually being drawn 3 times in the same place, so the effective opacity increases.
  - once in `g.x2y2 g.imagelayer` - where it's supposed to be
  - once in `g.xy g.imagelayer` - even though it gets correctly clipped to x2y2
  - once in `g.xy2 g.imagelayer` - which shouldn't even exist, but the mock forgets to anchor `yaxis2` to `x2`. Perhaps this should be considered a separate bug? ie if `xaxis2.anchor = 'y2'` shouldn't `yaxis2.anchor` automatically default to `x2`? If you drag one of the subplots after setting the background transparent you'll see we're drawing gridlines for `xaxis` onto the upper subplot.

So the easiest test of fixing this bug would be to give that mock the transparent `plot_bgcolor` without changing the baseline image.

Shapes use a similar structure for data-referenced below-data display, so it likely has the same bug, but I haven't tested it there.

alexcjohnson:
Huh, I thought I could wait on this one, but looks like I may need to address it now as part of my work on the array relayout API...

alexcjohnson:
Turns out shapes had a different issue with my branch and do *not* suffer from this bug. I'll keep myself assigned here since I pretty much know what to do, but may be able to keep it separate after all.

alexcjohnson:
closed by #1453

Issue #1452, opened 2017-03-08
Shapes layer: 'below' doesn't play well with subplots


alexcjohnson:
I'm not sure this really *has* a solution even theoretically, if you consider the layering that needs to happen with inset plots (where the background of the inset plot needs to be drawn on top of the data of the main plot) but posting as a bug for discussion purposes since it's at the very least unexpected.

If you make a `layer: 'below'` shape that's data-referenced on one axis and paper-referenced on another, and it crosses two subplots (which could actually be a very nice way to highlight correlations between these subplots) the shape will NOT be below the data of both subplots.

```js
Plotly.newPlot(gd,[
    {y:[1,2,3]},
    {y:[1,2,3],yaxis:'y2'}
],{
    yaxis:{domain:[0,0.4]},
    yaxis2:{domain:[0.6,1]},
    shapes:[{
        xref:'x',yref:'paper',type:'rect',
        x0:0.8,x1:1.2,y0:0,y1:1,
        fillcolor:'#eee',layer:'below'
    }]
})
```

![screen shot 2017-03-08 at 12 38 35 am](https://cloud.githubusercontent.com/assets/2678795/23691624/9e86b9f2-0397-11e7-846f-dfe8d0b87d3f.png)

etpinard:
Nice find. I hope the [`shape_below_traces`](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/shapes_below_traces.json) isn't [wrong](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/shapes_below_traces.png).

I'd vote for making `layer: 'below'` correspongd to below **all** traces linked to the axis set in the shape `xref` or `yref`.

alexcjohnson:
closed by #1453 - now `layer: 'below'` shapes and images are still in front of plot backgrounds in all cases EXCEPT on inset plots (any subplot that overlaps an earlier one without being explicitly overlaying) where the component is NOT referenced to both axes of that subplot. So for example, if `x2/y2` is on top of `x/y`, a shape referenced to `x2/y2` can still be seen with `layer: below` but one referenced to `x2/paper`, `paper/y2`, `x/y`, or anything else cannot be seen where it intersects the `x2/y2` subplot.

This is, I believe, the minimal set of objects it's theoretically impossible to show (short of some sort of horribly complicated stitching together of different objects in different z layers, which wouldn't work that well anyway due to antialiasing and would have strange effects of data objects from the lower plot slicing through them.)

PR #1453, opened 2017-03-08
Image and shape clip paths

alexcjohnson:
In testing axis reference changes for integrating v1.24 into the workspace, I noticed some strange situations with clip paths for shapes and images. Turns out they had different issues but related symptoms. There's one particular edge case that I didn't sort out here because it's a much bigger structural question, so I described it in #1452 instead.

@etpinard please review.


alexcjohnson:
Note: annotations do not have these issues, as they don't use clip paths. We will happily allow the arrow and text box of an annotation to go outside the plot area as long as its reference point is visible.

src/components/images/draw.js
```diff
@@ -143,9 +143,10 @@ module.exports = function draw(gd) {
             yId = ya ? ya._id : '',
             clipAxes = xId + yId;
 
-        if(clipAxes) {
-            thisImage.call(Drawing.setClipUrl, 'clip' + fullLayout._uid + clipAxes);
-        }
+        thisImage.call(Drawing.setClipUrl, clipAxes ?
+            ('clip' + fullLayout._uid + clipAxes) :
+            null
+        );
```
alexcjohnson: Needed to *unset* the clip path if you change from data-referenced to fully paper-referenced.

test/jasmine/tests/shapes_test.js
```diff
@@ -382,6 +382,71 @@ describe('Test shapes:', function() {
...
+                fillcolor: '#eee', layer: 'below'
+            }]
+        }).then(done);
+    });
+
+    afterEach(destroyGraphDiv);
+
+    function getShape(index) {
+        var s = d3.selectAll('path[data-index="' + index + '"]');
+        expect(s.size()).toBe(1);
```
alexcjohnson: before the fix above, the shape was drawn twice in its initial configuration `xref: 'x', yref: 'paper'`: once on subplot `xy` and again on `xy2`

src/components/shapes/draw.js
```diff
@@ -110,10 +110,15 @@ function drawOne(gd, index) {
...
-                'clip' + gd._fullLayout._uid + clipAxes);
-        }
+        // note that for layer="below" the clipAxes can be different from the
+        // subplot we're drawing this in. This could cause problems if the shape
+        // spans two subplots. See https://github.com/plotly/plotly.js/issues/1452
+        var clipAxes = (options.xref + options.yref).replace(/paper/g, '');
+
+        path.call(Drawing.setClipUrl, clipAxes ?
+            ('clip' + gd._fullLayout._uid + clipAxes) :
+            null
```
etpinard: Nice way to unset a clip path. :+1:

test/jasmine/tests/layout_images_test.js
```diff
@@ -425,23 +425,38 @@ describe('images log/linear axis changes', function() {
         // we don't try to figure out the position on a new axis / canvas
         // automatically when you change xref / yref, we leave it to the caller.
 
+        // initial clip path should end in 'xy' to match xref/yref
+        expect(d3.select('image').attr('clip-path') || '').toMatch(/xy\)$/);
```
etpinard: Ha. I didn't know about `.toMatch`. Pretty handy :hammer:

test/jasmine/tests/layout_images_test.js
```diff
@@ -425,23 +425,38 @@ describe('images log/linear axis changes', function() {
...
             // log to paper
             return Plotly.relayout(gd, {'images[0].yref': 'paper'});
         })
         .then(function() {
             expect(gd.layout.images[0].y).toBe(1);
 
+            expect(d3.select('image').attr('clip-path') || '').toMatch(/x\)$/);
+
+            // change to full paper-referenced, to make sure the clip path disappears
+            return Plotly.relayout(gd, {'images[0].xref': 'paper'});
```
etpinard: I'm surprised we weren't testing this before. Oh well, now it's :lock: ed

test/jasmine/tests/shapes_test.js
etpinard: Good eyes. Thanks very much!

etpinard:
Looks like `shapes_below_traces` produced a diff

image

is that on purpose?

alexcjohnson:
> Looks like shapes_below_traces produced a diff, is that on purpose?

Hmm, maybe #1452 needs to be addressed together with this PR after all, that's what's going on here. You'll see that the data in the lower left subplot gets covered up in the new result for that image. But the original is clearly wrong in how it breaks the two shapes that span that subplot and either the one above or to the right into two disjoint pieces.

From a slack chat:

> the reason I said (in #1452 ) "I'm not sure this really has a solution even theoretically” is that there really isn’t a layering that puts the shape in front of the plot area backgrounds but behind the traces, at least not when you consider inset subplots that want to have their own backgrounds in front of the main plot traces.
>
> One easy thing we could do is put the below shapes completely below the plot layers… then you have to make `plot_bgcolor` transparent to see the shapes.
>
> A trickier thing we could do is pull the background rects out of the subplot layers, putting them behind everything else, UNLESS we detect overlapping (but not overlaying) subplots.
>
> That would cover all the use cases I can think of except below shapes that are half referenced to an axis of an inset subplot… which would be an exceedingly weird situation.

@etpinard suggested:
> Maybe we should disallow shapes that span multiple subplots

That would certainly make this easier (the restriction would only apply to `layer: 'below'` shapes and images) - then we'd just have to figure out which subplot contains the shape. But there are some cool uses of subplot-spanning shapes - for example on the standard finance chart that has candlesticks/OHLC in the main plot and volume in a little lower subplot, if you want to highlight a time range. It would look weird to force these to be two separate shapes. Would also be annoying to try and enforce this restriction during editing (particularly via drag)

So I'm leaning toward the "trickier thing": pulling the plot backgrounds back to their own layer except for inset plots.

Thoughts?

alexcjohnson:
#1390 is wrapped up in this too... better fix that at the same time.

etpinard:
> So I'm leaning toward the "trickier thing": pulling the plot backgrounds back to their own layer except for inset plots.

This is definitely the _best_ solution from a user's perspective. If you feel like taking a shot at this, go for it :+1:

alexcjohnson:
https://github.com/plotly/plotly.js/pull/1453/commits/2102b0f39c9fa09a824417699598baeb9e045a83 fixes #1390 and #1452 - I still want to add a couple of tests of the inset vs flat background behavior though

src/components/images/draw.js
```diff
@@ -151,29 +163,44 @@ module.exports = function draw(gd) {
...
     });
+
+    var allSubplots = Object.keys(fullLayout._plots);
+    for(i = 0; i < allSubplots.length; i++) {
+        subplot = allSubplots[i];
+        var subplotObj = fullLayout._plots[subplot];
+
+        // filter out overlaid plots (which havd their images on the main plot)
+        // and gl2d plots (which don't support below images, at least not yet)
+        if(!subplotObj.imagelayer) continue;
```
etpinard: Looks good :+1:

src/plot_api/plot_api.js
```diff
@@ -308,8 +308,7 @@ Plotly.plot = function(gd, data, layout, config) {
 
         // keep reference to shape layers in subplots
         var layerSubplot = fullLayout._paper.selectAll('.layer-subplot');
-        fullLayout._imageSubplotLayer = layerSubplot.selectAll('.imagelayer');
-        fullLayout._shapeSubplotLayer = layerSubplot.selectAll('.shapelayer');
+        fullLayout._shapeSubplotLayers = layerSubplot.selectAll('.shapelayer');
```
etpinard: :pencil2: nicely done

test/jasmine/tests/cartesian_test.js
```diff
@@ -340,6 +352,70 @@ describe('subplot creation / deletion:', function() {
...
+
+            // verify that these are *all* the subplots and backgrounds we have
+            expect(gd.querySelectorAll('.subplot').length).toBe(3);
+            ['xy', 'x2y2', 'xy3'].forEach(function(subplot) {
+                expect(gd.querySelectorAll('.subplot.' + subplot).length).toBe(1);
+            });
+            expect(gd.querySelectorAll('.bg').length).toBe(behindCount + x2y2Count);
+        }
+
+        Plotly.plot(gd, [
```
etpinard: Fantastic test :tada:

etpinard:
:dancer: nicely done!

Don't forget to close off  #1390 and #1452 after merging :beers:

Issue #1335, opened 2017-01-26
Custom mode bar buttons are displayed upside down 


bruno-rino:
The mode bar icons that ship with plotly are for some reason flipped vertically. To compensate, plotly applies transformation matrix that includes a vertical flip (the -1 in the following line):

[`path.setAttribute('transform', 'matrix(1 0 0 -1 0 ' + thisIcon.ascent + ')');`](https://github.com/plotly/plotly.js/blob/master/src/components/modebar/modebar.js#L176)

To illustrate: the following example of how to add a custom icon uses a SVG path that looks good, because it is actually an upside down image (I render the SVG inline for comparison): http://codepen.io/bruno-rino/pen/bgowQx

The `-1` in the transformation matrix should be configurable. Or else any normal SVG path has to be flipped manually (and I don't know even how to go about doing that) before it can be used as an icon in the mode bar.

etpinard:
Ha. Good catch. 

Yes, our stock mode bar icons use bottom-to-top y axis (from [fontastic.me](http://fontastic.me/) I believe) which need to be flipped in the browser. 

We have a few options here:

### 1- reverse our icons

in [`src/fonts/ploticons/`](https://github.com/plotly/plotly.js/tree/master/src/fonts/ploticon) and remove the useless `-1` matrix transform. Problem: some users might be currently using our built icons paths by accessing them [here](https://github.com/plotly/plotly.js/blob/master/build/ploticon.js); this change might be considered _backward incompatible_.

### 2- add a general transform field

for example

```js
modeBarButtonsToAdd: [{
    name: 'my icon',
    icon: /* icon string path */,
    ascent: 100,
    transform: 'matrix(1 0 0 -1 0 100)',
    click: function(gd) { }
}]
```

This solution might be too confusing for some.

### 3- add a `flip` boolean field

or `direction: 'up' | 'down'`, ...

```js
modeBarButtonsToAdd: [{
    name: 'my icon',
    icon: /* icon string path */,
    ascent: 100,
    filp: true,
    click: function(gd) { }
}]
```
------

Personally, I prefer a type **3** solution (once we agree on the field name and values) for its simplicity. unless you think we need to cover other cases through the API.

rreusser:
Perhaps `flipy` is more descriptive both since it's more fully descriptive and since `flipx` could conceivably be useful too (left right arrows without needing two separate icons?)

bruno-rino:
[...]

PR #1461, opened 2017-03-10
Cartesian dropline support

rpaskowitz:
This PR adds support for droplines on cartesian graphs. It works with sub-plots and scatter/line/bar.

Some examples:
drop1
drop2
drop3
drop4

A new option is added to enable this behaviour (showDroplines), and it is disabled by default for backwards compatibility. Even if enabled, the chart must be on 'closest' hovermode and not 'compare'. Without this restriction there can be multiple points hovered simultaneously and that tended to make things confusing if there were multiple droplines.

A couple small things right now:
- ~~The droplines are drawn before the labels, but because the hover event fires on small moves, it can cause a redraw of the dropline causing it to be on top of the label. This should be pretty easily solved.~~
- ~~It would probably be better to draw the dropline to the secondary axis on the right, instead of the left, for series plotted on that axis. Some dropline implementations just draw the line full across which avoids that issue, but I prefer this styling. Getting the dropline to swap sides has the added benefit of making it much more clear to the user which axis they should be reading, which on charts with 2 is always a point of confusion at first glance.~~ The bigger issue is what to do if there are even more axes off to the side, like on `mocks/20.json`, where the line may go to the plot edge and not the true axis, but where drawing the line on top of other axes to reach would look weird.

drop5

As usual, tests are a TODO. Please let me know if you'd prefer them sooner in the future. Also let me know if you'd prefer something other than a PR for a feature that may/may not be merged (which is why I skip tests off the bat)


rpaskowitz:
Related to this PR, but not exclusively, I think it may make sense to swap the hovertext anchor point based on if a data series is plotted on a right or left axis. With the droplines, it becomes more apparent, since often for a series plotted on a right axis the hovertext fully covers the horizontal dropline. Even without droplines though, having the hovertext to the right of the datapoint of a series which is plotted on a right axis makes it more difficult to visualize where it lands on the axis.

Specifically in the following block for the non-rotated case, add a check if the series is on the right axis, and if so, default to `end` and switch to `start` only in the case where the label would be off the left side of the chart.

```
        if(rotateLabels) {
            d.pos = htx;
            anchorStartOK = hty + dy / 2 + txTotalWidth <= outerHeight;
            anchorEndOK = hty - dy / 2 - txTotalWidth >= 0;
            if((d.idealAlign === 'top' || !anchorStartOK) && anchorEndOK) {
                hty -= dy / 2;
                d.anchor = 'end';
            } else if(anchorStartOK) {
                hty += dy / 2;
                d.anchor = 'start';
            } else d.anchor = 'middle';
        }
        else {
            d.pos = hty;
            anchorStartOK = htx + dx / 2 + txTotalWidth <= outerWidth;
            anchorEndOK = htx - dx / 2 - txTotalWidth >= 0;
            if((d.idealAlign === 'left' || !anchorStartOK) && anchorEndOK) {
                htx -= dx / 2;
                d.anchor = 'end';
            } else if(anchorStartOK) {
                htx += dx / 2;
                d.anchor = 'start';
            } else d.anchor = 'middle';
        }
```

I mention it here because I thought I had a bug where I didn't think the horizontal dropline was drawn, when really it was just fully obscured.

My only concern with this is that if you have 2 data points near one another, but plotted on opposite series, it may seem odd to the user that the hovertext swaps positions so drastically.

etpinard:
Thanks for the PR @rpaskowitz !!!

I like this new functionality a lot. Though I'm thinking it should be via `layout.hovermode` instead of some config option. @alexcjohnson @cldougl thoughts?

alexcjohnson:
Yeah, I like it! We should harmonize this behavior with gl2d, which currently does exactly this - always, cannot be turned off, can it? But it uses solid black lines:

screen shot 2017-03-10 at 10 50 56 am

I can see this being a `hovermode` - that works nicely with not allowing it in compare mode, and also allows the end viewer to turn it on and off, not relying on the developer, since it'll show up in the modebar. Then maybe we can turn it off by default for gl2d as well.

Thoughts on colors? For robustness I kind of want to match the box/lasso select outline, which is alternating black/white dashes (drawn as solid white with dashed black on top of it) which probably isn't the prettiest but it always shows up, no matter what data or background it's drawn against.

re: axes off to the side ("free" axes) - perhaps we can just add a tick or arrow on the axis itself but not draw the whole line over to it (wherever it may be), just draw the line to the edge of the plot area? Same goes for subplots that share an axis, we probably don't want the line to go all the way across the other subplot, just add a mark of some sort on the relevant axis.

re: flipping the default label orientation for series drawn against the right axis: I think that's a great idea! The hover label then functions as an arrow pointing toward that axis. There's a chance people will find this *less* intuitive, as the label is *positioned* toward the wrong axis, even if it's *pointing* toward the correct one... but the benefits seem greater so lets do it. A thought on implementation: if we did this based on the pixel position of the axis relative to the hovered-on point, rather than on the side, then if you put a free axis over the middle of the plot, the labels would flip as you cross the axis, which seems like it would be a cool & useful effect - and you'd automatically get it right for free axes off the edge like `mocks/20.json`

rpaskowitz:
> Though I'm thinking it should be via layout.hovermode instead of some config option.

On one hand, I like using `hovermode` since it untangles the fact that this only works on specific hovermodes. On the other hand though, today `hovermode` really impacts what triggers a hover, and what data is included in the hover events, and not what is shown when a hover is triggered. I think that definition highlights why I struggle to name a new hovermode for this, since it would basically be `closest_dropline`, where apart from the UI aspect, it behaves identically to `closest`. Just as today there's no way to use the `hovermode` to control the hovertext...

...I'm picturing something like a `hoverinfo` (except not `hoverinfo`, since that controls what is in the hovertext) that could be set to `text`, `dropline` or `text+dropline` to control whether the current hovertext is shown at all, and this new dropline functionality. Currently you can't subscribe to hover events without the hovertext being shown, so this would provide a nice way to disable that (as opposed to nuking it in CSS or the DOM) as well as add this new functionality. What this doesn't solve is that droplines would still only be shown on `closest` hovermode.

> Thoughts on colors? For robustness I kind of want to match the box/lasso select outline, which is alternating black/white dashes (drawn as solid white with dashed black on top of it) which probably isn't the prettiest but it always shows up, no matter what data or background it's drawn against.

I prefer the look with colors, but understand that it can cause some comprehension issues in certain scenarios. In others though I think it lends clarity and provides easier comparison across series. So far in my testing the colors look pretty good, even with some dense data series. I'd might suggest the white background behind the color, which should increase visibility while still keeping some indicator of the series?

> perhaps we can just add a tick or arrow on the axis itself but not draw the whole line over to it

I could picture a dot/circle on the axis the goes slides along the axis to help make this clear. I'd still likely want to keep the line within the plot/subplot because a second convenience with droplines is not just finding the line on the axis, but also comparing data points of the same or other series with a selected on ("where we higher or lower 7 days ago?")

> if we did this based on the pixel position of the axis relative to the hovered-on point

I'll need to unpack this a bit and learn a bit more about "free" axes, but I see the value of the generic implementation that also handles cases like an axis in the middle.

> We should harmonize this behavior with gl2d

I stumbled across the droplines on gl2d after implementing this and had a moment of "is it somehow using my implementation from cartesian but failing to apply stroke?". Once we settled on behavior to enable and styling, I can take a look at the gl2d implementation, but it could take longer since I've not done any WebGL before.

alexcjohnson:
> today `hovermode` really impacts what triggers a hover, and what data is included in the hover events, and not what is shown when a hover is triggered.

Good point - @etpinard is right that this shouldn't go in config, but we should make it another attribute, independent of `hovermode`, and give it a modebar icon that's only available in `closest` mode so viewers can toggle this.

This also has an analog in gl3d, where each axis independently has an attribute `showspikes` - along with `spikesides` for secondary lines, and `spikethickness` and `spikecolor` to style them. I can't say I'm that excited about the name "spikes" but we do already have it so using a different name could be confusing... That said, unlike 3D I don't think we need to allow this on a per-axis basis, a global setting seems fine. And maybe sometime we will want to enable styling like color and thickness but lets omit that for now.

>  I'd might suggest the white background behind the color, which should increase visibility while still keeping some indicator of the series?

Ah, that's nice - we could use the same coloring as the hover label: light colors get a black contrast, dark colors get a white contrast.

> I could picture a dot/circle on the axis the goes slides along the axis to help make this clear. I'd still likely want to keep the line within the plot/subplot because a second convenience with droplines is not just finding the line on the axis, but also comparing data points of the same or other series with a selected on ("where we higher or lower 7 days ago?")

Oh yes - I wasn't suggesting getting rid of the line within the subplot... just not extending it outside the subplot if the axis isn't at the edge of the subplot. But the point about using this line for comparison is interesting - does this mean we should actually extend these lines the full width of the subplot, not just toward the axis?

> I can take a look at the gl2d implementation, but it could take longer since I've not done any WebGL before.

Doesn't have to be in this PR, nor does it need to be your job unless you're itching to play with webGL. I just want to make sure that whatever we do here will also work for webGL when we get to it.

etpinard:
> That said, unlike 3D I don't think we need to allow this on a per-axis basis, a global setting seems fine

I disagree here. I kinda like the idea of adding `showspikes` to every axis object. I can image someone only wanting spikes on the x (or y) axis of a given subplot. We should make sure that the `xaxis.showspikes` description mentions that it only has an effect when `layout.hovermode` is set to `'closest'`.

> And maybe sometime we will want to enable styling like color and thickness but lets omit that for now.

Agreed. No need to do this in this PR.

> But the point about using this line for comparison is interesting - does this mean we should actually extend these lines the full width of the subplot, not just toward the axis?

For example, on a stacked/coupled subplot layout:

[image](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/stacked_coupled_subplots.png)

I think, yes, when `xaxis.showspikes` is turned on it would be nice to make the spikes reach the x-axis at the bottom. Now, maybe this isn't as obvious for other subplot layouts like:

[image](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/multiple_axes_multiple.png)

here with `yaxis.showspikes: true` should the spikes reach the y axes with non-default `position` settings? I'm not sure. In this case, the spikes would overlap one or several y-axis title(s).

rpaskowitz:
> I think, yes, when xaxis.showspikes is turned on it would be nice to make the spikes reach the x-axis at the bottom. Now, maybe this isn't as obvious for other subplot layouts like:
> Oh yes - I wasn't suggesting getting rid of the line within the subplot... just not extending it outside the subplot if the axis isn't at the edge of the subplot.

Here's what it looks like with dots on the axes. The dropline will draw the the edge of the plot area, but the dot will appear on whichever axis the series is associated to.

> Ah, that's nice - we could use the same coloring as the hover label: light colors get a black contrast, dark colors get a white contrast.

This example also has a white line behind the colored line, which is also a pixel wider on each side. I'm not sure the contrast is needed for alternating colors because the backgrounds purpose is to make the line visible on top of other series, not necessarily visible on top of a series of the same color. Since there could be any number of other colors to differentiate from, alternating backgrounds may not help. I think if the background were black, it would stick out really poorly on a chart with a white background. (that being said, if the chart had a black or dark background, this one would stick out, so score another point for configurability - unless we wanted to use the chart base color to programatically determine the dropline background).

drop6

rpaskowitz:
>  where each axis independently has an attribute `showspikes`
>  I kinda like the idea of adding showspikes to every axis object.
> But the point about using this line for comparison is interesting - does this mean we should actually extend these lines the full width of the subplot, not just toward the axis?

Perhaps we could resolve this stuff by making showspikes configurable with options like `false` (off), `axis` on to plot to the data axis only, `true` to draw both ways. If it's configurable per axis someone could do `true` for the y axis to get the line all the way across, but keep the a axis at `axis`. Alternatively this could be done like `hoverinfo` where we could pack in whether lines are drawn to the data axis, entirely, and if the dot is drawn on the axis line.

rpaskowitz:
Took a re-read, but 
> I think, yes, when xaxis.showspikes is turned on it would be nice to make the spikes reach the x-axis at the bottom

makes sense to me now. It's certainly possible to draw all the way to the axis (I just did it accidentally), so maybe it would be another configurable option in the `showspikes` configuration to have the line drawn all the way to the axis for the cases where one knows it's not going to do something ugly like draw over other axes.

rpaskowitz:
>Took a re-read, but
>>I think, yes, when xaxis.showspikes is turned on it would be nice to make the spikes reach the x-axis at the bottom

>makes sense to me now. It's certainly possible to draw all the way to the axis (I just did it accidentally), so maybe it would be another configurable option in the showspikes configuration to have the line drawn all the way to the axis for the cases where one knows it's not going to do something ugly like draw over other axes.

I think this is actually a non-issue. Between the two screenshots are two different scenarios. In the first example, `mocks/stacked_coupled_subplots`, the 3 plots are all sharing a single x-axis, they don't have different `free` axes, while in `mocks/multiple_axes_multiple` and `mocks/20` there are some series plotted on `free` axes. In the case of shared axes, we can simply draw the line directly to the axis. It is in the `free` case where we don't want to do that, and draw instead to the plot area.

I'm holding off for some more changes before pushing code, but here are some examples of my updated implementation:
`stacked_coupled_subplots` -- shared x-axis
drop_shared_x

`stacked_subplots_shared_yaxis` -- shared y-axis
drop_shared_y

`20` -- free y-axes
drop_free_y

`shared_axes_subplots` -- shared x and y axes. The droplines really helped me understand this chart, unrealistic though it may be, except if someone were trying to make some sort of clever infographic venn like diagram, since at a glance I wasn't sure which sub-plots were using which axes, but it was very clear with the droplines.
drop_shared_xy


My implementation still needs to be updated to handle free x axes, since I'm only handling y right now. And I need to get rid of triple nested ternaries.

An interesting discovery is that while in 'compare' mode, the axis value callouts in shared axis scenarios are arguable in the wrong spot since they seem to be being positioned at the bottom of the y axis instead of on the x axis (which is basically what the earlier implementation of droplines was doing, so I imagine a similar fix is straightforward):
compare_marker

rpaskowitz:
Most points addressed in latest commit.

TODO:
- Decide if we want either dropline to not just go to the axis the value relates to, but full across the chart. Could be configurable as part of showspikes.
- Decide if we we need the background line behind the colored line to change based on chart and/or line color.
- Decide if a modebar is needed to enable/disable this behavior (I'm not 100% on it being a modebar interaction since it wouldn't be a simple on/off, as it's per-axis, including individual axes when there are subplots, and it may not be desirable at all for certain charts - I think it may be best left to the chart designer - one example, x-axis spikes are pretty unnecessary for bar charts).
- Decide if swapping the hovertext position based on axis position should be done as a part of this PR (my feeling now is no, since the axis marker mostly resolves the visibility issue that the problem previously exacerbated, and it's independent)
- Tests, naturally

alexcjohnson:
@rpaskowitz thanks for all the great examples and exploration of so many scenarios!

> > That said, unlike 3D I don't think we need to allow this on a per-axis basis, a global setting seems fine
>
> I disagree here. I kinda like the idea of adding `showspikes` to every axis object. I can image someone only wanting spikes on the x (or y) axis of a given subplot. We should make sure that the `xaxis.showspikes` description mentions that it only has an effect when `layout.hovermode` is set to `'closest'`.

> Decide if a modebar is needed to enable/disable this behavior (I'm not 100% on it being a modebar interaction since it wouldn't be a simple on/off, as it's per-axis, including individual axes when there are subplots, and it may not be desirable at all for certain charts - I think it may be best left to the chart designer - one example, x-axis spikes are pretty unnecessary for bar charts).

True, bar charts don't benefit from a guide line down the bar, and in fact it could obscure things... But we could handle that based on trace type instead. I worry that few people will go to the trouble of understanding its implications fully enough to configure it per-axis, and at least for me personally I would really appreciate the ability to turn this effect on or off (ie via the mode bar). I suppose we could define a global `layout.showspikes` that's controlled by the mode bar and sets the default that per-axis values override... seems potentially confusing though.

> Here's what it looks like with dots on the axes. The dropline will draw the the edge of the plot area, but the dot will appear on whichever axis the series is associated to.

> Decide if we want either dropline to not just go to the axis the value relates to, but full across the chart. Could be configurable as part of showspikes.

I like the look when the line is drawn from the point to the axis, no matter where that axis is displayed (even if it's a free axis?), and your point that this helps explain multi-axis plots is a great observation. Does that mean we don't need the dot on the axis? I like that look too when the axis has a line, but it seems redundant if the line gets there, and I *don't* like it when the axis doesn't have a line or ticks - then it overlaps the tick labels. Might be better just to leave it off?

> Decide if we we need the background line behind the colored line to change based on chart and/or line color.

> This example also has a white line behind the colored line, which is also a pixel wider on each side. I'm not sure the contrast is needed for alternating colors because the backgrounds purpose is to make the line visible on top of other series, not necessarily visible on top of a series of the same color. Since there could be any number of other colors to differentiate from, alternating backgrounds may not help.

I still think it would be worth trying exactly the colors that the hover label gets. Maybe it will look a bit heavy on white-background charts when the contrast color goes to black, but it will guarantee contrast in any situation: what if you have light-colored scatter points either in a thick cloud over a dark background, or with a dark `marker.line` on a light background. In both cases the spike lines would disappear. You're right that in a plot with a jumble of small feature of many different colors this won't really differentiate itself locally, but I still think except in some truly pathological cases it will make a clear feature to guide the eye vertically or horizontally when you consider the whole chart. That said I feel like the extra width of the white (or contrasting) line - which would help for some of these really chaotic cases - is probably too wide and obscures too much, particularly if you're using the line to compare values across the plot.

> Decide if swapping the hovertext position based on axis position should be done as a part of this PR (my feeling now is no, since the axis marker mostly resolves the visibility issue that the problem previously exacerbated, and it's independent)

Agreed, but might be nice to make an issue for it so we don't forget to revisit it later.

rpaskowitz:
Leaving aside the issue of per-legend, or chart-type, etc...fot now as I think that may need some internal plotly discussion. It mostly doesn't matter to me since I won't wind up using the modebar, and needing to enable globally, or per type, or per axis won't make too much of a difference.

> I like the look when the line is drawn from the point to the axis, no matter where that axis is displayed (even if it's a free axis?), and your point that this helps explain multi-axis plots is a great observation. Does that mean we don't need the dot on the axis?
I'm not sure I follow all of this comment. I also like the line drawn from point to axis, but only to the graph edge in the case of a free axis (for the problem of having the line cross the other axes). I like the dot as an indicator on the free axes to help make the association, and while redundant when the axis on the chart-edge is the same as the series axis, I think the dot serves a purpose otherwise it will need to be a learned behaviour that "when the dot isn't shown, it means it's the axis on chart edge".

> I like that look too when the axis has a line, but it seems redundant if the line gets there, and I don't like it when the axis doesn't have a line or ticks - then it overlaps the tick labels. Might be better just to leave it off?

I'll grant it does sometimes look a bit odd when there is no line on the axis, which happens not only when the line is disabled, but if the chart edge isn't at x = 0, so the line is offset from the chart edge. That's still the place at which the dropline itself will end, so I don't think it's too weird, and I think the placement could be adjusted a bit to make sure it doesn't cover the label text.

I think this may speak further to the need for something like the `showspikes` option to be like `hovertext` in accepting various options - this dot could be one of those configurable options - just as disabling tick markers is one, if this looks weird when there's no tick markers, then people should perhaps turn the dot off. My list of available spike options at this point would be `toaxis`, `across, and `marker`. `across` would be 100% width, not just from the point, and would supersede `toaxis`.

> I still think it would be worth trying exactly the colors that the hover label gets.
I can put together some samples with the different background and widths.

etpinard:
> I worry that few people will go to the trouble of understanding its implications fully enough to configure it per-axis, and at least for me personally I would really appreciate the ability to turn this effect on or off (ie via the mode bar). I suppose we could define a global layout.showspikes that's controlled by the mode bar and sets the default that per-axis values override... seems potentially confusing though.

As discussed privately with @alexcjohnson, making this new mode bar button turn on or off the spikes would be pretty easy and shouldn't be a blocker for `(x|y)axis.showspikes`. 

For example, we could do the same as in 3D where the mode bar "hover" handler turns off `showspikes` for all axes on 1st click and on 2nd click turns the original `showspikes` back on - see logic [here](https://github.com/plotly/plotly.js/blob/4b7aa2be1ea3a7ddf3c52cecf364b84388f8c6bc/src/components/modebar/buttons.js#L339-L387).

Alternatively, we could add a _true_ toggle button where the 1st click turns off `showspikes` for every axes, the 2nd click turns on `showspikes` for every axes and the current _reset axes_ mode bar button would not only reset the axis range but also the spike settings back to their original values.

At the moment, I think the second option is the best (i.e. less confusing). Moreover, similar to @alexcjohnson 's idea, I think we should only show this new _toggle spikes_ mode bar button when scatter traces are present on the graph.

--------- 

> - Decide if we we need the background line behind the colored line to change based on chart and/or line color.

I'll let you and @alexcjohnson decide on what the best default spike color is. But I think we should at the very least expose `spikecolor` and `spikethinkness` attributes as part of this PR - on-par with the 3D spike [attributes](https://github.com/plotly/plotly.js/blob/4b7aa2be1ea3a7ddf3c52cecf364b84388f8c6bc/src/plots/gl3d/layout/axis_attributes.js#L17-L48). Maybe we could also expose `spikedash` with the same possible values as the scatter [`line.dash`](https://github.com/plotly/plotly.js/blob/4b7aa2be1ea3a7ddf3c52cecf364b84388f8c6bc/src/traces/scatter/attributes.js#L150-L174).

-----

> I think this may speak further to the need for something like the showspikes option to be like hovertext in accepting various options - this dot could be one of those configurable options - just as disabling tick markers is one, if this looks weird when there's no tick markers, then people should perhaps turn the dot off. My list of available spike options at this point would be toaxis, across, and marker. across would be 100% width, not just from the point, and would supersede toaxis.

I think a [flaglist](https://github.com/plotly/plotly.js/blob/4b7aa2be1ea3a7ddf3c52cecf364b84388f8c6bc/src/lib/coerce.js#L190-:L220) attribute like `spikemode` with possible flags `'toaxis'`, `'marker'` etc might be more _plotly-esque_. But I'm not 100% sure this setting needs to be exposed in this PR.

rpaskowitz:
> But I think we should at the very least expose spikecolor and spikethinkness

I don't think this adds much work (spikedash too), so I'll just include it. Same goes for the flaglist, I don't think it is that big a deal, so may as well get it done (in my own projects, I find this stuff doesn't often get revisited after shipping, so sometimes better to just get it over with).

> At the moment, I think the second option is the best 

If I follow the current recommendations it would entail:
- A per-axis `showspike` (plus associated color, thickness, dash, mode)
- A per-cartesian-chart-type configuration (scatter X and Y true, line X and Y true, bar Y true, row X true, ...) of the above per-axis values
- Allow manual override of the per-cartesian-chart-type configuration through axis layout options
- A modebar that goes `default` -> (`all off` <-> `all on`)
- The 'reset axis' modebar going back to `default`
 - `default` in both these cases meaning whatever the chart configuration happened to be when the spike toggle modebar button was hit (based on per-axis layout options)

If the 'per-axis-cartesian' defaults are desired, my main question would be what to do when there are mixed-series on a chart. Since the properties are currently being discussed as per-axis and not per-series, if there were a scatter+bar chart, then a line would be drawn to the x-axis for the bar as it would have been enabled by the scatter series' presence. I think resolving this would mean moving the spike settings to the series and having a flaglist for `showspikes` like `x+y` (I think it'd probably fine to keep the spike color, thickness, etc... the same for both axes for a given series, I can't picture wanting to draw those in different colors, and in fact, specifying those per-series may actually make more sense than specifying per-axis)

etpinard:
@rpaskowitz @alexcjohnson status update on this thing? It would be nice to include it in next week's `1.26.0`.

rpaskowitz:
I can likely get the spikecolor/thickness stuff done. Before tackling the rest in my last comment, if someone could confirm that I've properly described the functionality, I can get working on that too.

Completing tests would add the most uncertainty to wrapping this up, but if the functionality described is correct, implementing it shouldn't take too long, and I can start looking at the rest.

alexcjohnson:
> A per-axis `showspike` (plus associated color, thickness, dash, mode)

👍 

> A per-cartesian-chart-type configuration (scatter X and Y true, line X and Y true, bar Y true, row X true, ...) of the above per-axis values
> Allow manual override of the per-cartesian-chart-type configuration through axis layout options

I'd vote to omit this for now - just make the default off for all axes, and users can opt into it. We can revisit this later and decide if these warrant turning on by default, once we have the functionality available.

> A modebar that goes `default` -> (`all off` <-> `all on`)
> The 'reset axis' modebar going back to `default`

👍 

 > `default` in both these cases meaning whatever the chart configuration happened to be when the spike toggle modebar button was hit (based on per-axis layout options)

I'd put this in the initial draw - ie alongside [saveRangeInitial](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L155)

rpaskowitz:
>>A per-cartesian-chart-type configuration (scatter X and Y true, line X and Y true, bar Y true, row X true, ...) of the above per-axis values
Allow manual override of the per-cartesian-chart-type configuration through axis layout options

>I'd vote to omit this for now - just make the default off for all axes, and users can opt into it. We can revisit this later and decide if these warrant turning on by default, once we have the functionality available.

I think the issue in omitting this is the different expectations on different plot types, namely not wanting/needing a dropline on both axes for row/column charts. If we don't want to define them on the standard chart types, then I think the "allow manual override" remains from the earlier "per-axis `showspike`". From that perspective, this works for my needs, since we'll be able to configure it per-chart, and the only cases where it will be weird will be when someone takes a standard chart and uses the modebar to turn 'all on'.

The mode bar behavior for `default` -> (`all off` <-> `all on`) will now have `default` only be based on the initial chart rendering, which unless someone manually set up `showspike` on the individual axis, and with no defaults per chart, will be the same as `all off`.

I'll go ahead with that plan.

rpaskowitz:
Spending some time on this today, and re-reading old comments. Main question I have is if people would go for 'showspike' being on the series instead of the axes. It addresses the mixed-mode composite chart, and has a certain other amount of elegance. It may not be required given the dropped requirement for the per-cartesian-chart-type defaults, but when one enables a spike on the x-axis for example, if there are both line and bars, it could still be undesirable.

alexcjohnson:
> Main question I have is if people would go for 'showspike' being on the series instead of the axes

Interesting idea... my gut reaction though is people would find that confusing, and harder to work with given that you often have many traces of the same type on a chart. Also with 3D we already have precedent for `showspike` on the axes.

rpaskowitz:
The various `spike*` configuration options have now been added. Last functional piece should be the modebar.

rpaskowitz:
Modebar support has now been added. I tried fighting a bit with fontello to get a new icon in, but it was a losing battle. Attached is an icon I've made (but feel free to change). Assistance in getting it into the font would be appreciated.


[spikeline.svg.zip](https://github.com/plotly/plotly.js/files/904478/spikeline.svg.zip)

Some notes about the modebar behavior:
- I have made it so that if you are not in 'closest' mode, then when you enable the spikelines you are automatically switched to 'closest'.
- The button properly displays the state for whether spikelines are on. It will only be on when all axes have it enabled as clicking it will enable all spikelines.
- The defaults for showspike have been set to false, but various styles are still set. It is these default styles that are used when enabling through the modebar (but custom styles will not be overridden)
- If you change from 'closest' to another hovermode, the spikelines are all disabled and the toggle is shown as disabled. This is done even if the user had not used the toggle to go to `all on` from `default`, which is probably not desirable (and as I type these notes, I now know why I did this, and why I need't have done it, and will fix it - probably)
- Reset Axes has also been updated to get things back to `default` (which out of the box is everything disabled, but could be something else depending on how the chart was initializes)

src/lib/index.js
```diff
@@ -636,3 +636,31 @@ lib.numSeparate = function(value, separators, separatethousands) {
...
+    else if(dash === 'dashdot') {
+        dash = (3 * dlw) + 'px,' + dlw + 'px,' + dlw + 'px,' + dlw + 'px';
+    }
+    else if(dash === 'longdashdot') {
+        dash = (5 * dlw) + 'px,' + (2 * dlw) + 'px,' + dlw + 'px,' + (2 * dlw) + 'px';
+    }
+    // otherwise user wrote the dasharray themselves - leave it be
+
+    return dash;
+};
```
alexcjohnson: I like the idea of breaking this out - but lets keep it in `drawing`, it fits there thematically and `lib` is a bit overloaded.

src/plots/cartesian/graph_interact.js
```diff
@@ -857,73 +857,95 @@ function createDroplines(hoverData, opts) {
...
+        if(ySpikeLine) {
+            // Background horizontal Line (to y-axis)
+            container.append('line')
+                .attr('x1', xBase)
+                .attr('x2', xEndSpike)
+                .attr('y1', yPoint)
+                .attr('y2', yPoint)
+                .attr('stroke-width', yThickness + 2)
+                .attr('stroke', yContrastColor)
+                .attr('class', 'dropline');
```
alexcjohnson: `.attr` can take an object ie `.attr({x1: xBase, x2: xEndSpike ... })`

also it's a bit nicer to use `.classed('dropline', true)` instead of `.attr('class', 'dropline')`, just because it makes things easier to extend later, if it turns out we need multiple classes or something.

src/plots/cartesian/layout_attributes.js
```diff
@@ -244,9 +244,9 @@ module.exports = {
         role: 'style',
         description: 'Determines whether or not the tick labels are drawn.'
     },
-    showspikes: {
+    showspike: {
```
alexcjohnson: in `gl2d` it's already `showspikes` - which I think makes sense, as even if at any given time you only show one, there are many spikes you *could* show, ie one for each data point.

src/plot_api/plot_api.js
```diff
@@ -151,9 +151,12 @@ Plotly.plot = function(gd, data, layout, config) {
...
 
-    // save initial axis range once per graph
-    if(graphWasEmpty) Plotly.Axes.saveRangeInitial(gd);
-
+    if(graphWasEmpty) {
+        // save initial axis range once per graph
+        Plotly.Axes.saveRangeInitial(gd);
+        // save initial show spikes once per graph
+        Plotly.Axes.saveShowSpikeInitial(gd);
+    }
```
alexcjohnson: Just a heads up, this is going to cause merge conflicts, as in the new axis constraints feature this block got moved - https://github.com/plotly/plotly.js/pull/1522/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R274

src/components/modebar/buttons.js
```diff
@@ -183,7 +183,8 @@ function handleCartesian(gd, ev) {
         var mag = (val === 'in') ? 0.5 : 2,
             r0 = (1 + mag) / 2,
             r1 = (1 - mag) / 2,
-            axList = Axes.list(gd, null, true);
+            axList = Axes.list(gd, null, true),
+            allEnabled = 'on';
```
alexcjohnson: this (and `fullLayout._cartesianSpikesEnabled`) reads to me like a boolean, and it seems like that would simplify some of the logic too.

src/plot_api/plot_api.js
```diff
@@ -153,12 +153,8 @@ Plotly.plot = function(gd, data, layout, config) {
...
     }
 
-    if(graphWasEmpty) {
-        // save initial axis range once per graph
-        Plotly.Axes.saveRangeInitial(gd);
-        // save initial show spikes once per graph
-        Plotly.Axes.saveShowSpikeInitial(gd);
-    }
+    // save initial show spikes once per graph
+    if(graphWasEmpty) Plotly.Axes.saveShowSpikeInitial(gd);
```
alexcjohnson: Ah, you already got this merged, great! didn't notice it before making my previous comment obviously :)

alexcjohnson:
I like it! I played with it a bit and there are a few little bits of style I'd tweak. Definitely up for discussion but here are my diff and a couple of before/after screenshots:

```js
--- a/src/plots/cartesian/graph_interact.js
+++ b/src/plots/cartesian/graph_interact.js
@@ -890,9 +890,10 @@ function createDroplines(hoverData, opts) {
                 .attr('x2', xEndSpike)
                 .attr('y1', yPoint)
                 .attr('y2', yPoint)
-                .attr('stroke-width', yThickness + 2)
+                .attr('stroke-width', yThickness)
// the extra width, particularly with black contrast, seemed too heavy to me
                 .attr('stroke', yContrastColor)
-                .attr('class', 'dropline');
+                .attr('class', 'dropline')
+                .classed('crisp', true);
// this turns on shape-rendering: crisp-edges, which I think looks nicer in general for
// pure vertical or horizontal lines, but particularly if we drop the extra width on the contrast,
// this prevents that line from bleeding around the dashed line by a fractional pixel
 
             // Foreground horizontal line (to y-axis)
             container.append('line')
@@ -903,7 +904,8 @@ function createDroplines(hoverData, opts) {
                 .attr('stroke-width', yThickness)
                 .attr('stroke', yColor)
                 .attr('stroke-dasharray', yDash)
-                .attr('class', 'dropline');
+                .attr('class', 'dropline')
+                .classed('crisp', true);
         }
         // Y axis marker
         if(yMarker) {
@@ -922,22 +924,24 @@ function createDroplines(hoverData, opts) {
             container.append('line')
                 .attr('x1', xPoint)
                 .attr('x2', xPoint)
-                .attr('y1', yEndSpike)
-                .attr('y2', yBase)
-                .attr('stroke-width', xThickness + 2)
+                .attr('y2', yEndSpike)
+                .attr('y1', yBase)
+                .attr('stroke-width', xThickness)
                 .attr('stroke', xContrastColor)
-                .attr('class', 'dropline');
+                .attr('class', 'dropline')
+                .classed('crisp', true);
 
             // Foreground vertical line (to x-axis)
             container.append('line')
                 .attr('x1', xPoint)
                 .attr('x2', xPoint)
-                .attr('y1', yEndSpike)
-                .attr('y2', yBase)
+                .attr('y2', yEndSpike)
+                .attr('y1', yBase)
// start (specifically the dashed line) at the base, so there's always the main color
// immediately at the axis
                 .attr('stroke-width', xThickness)
                 .attr('stroke', xColor)
                 .attr('stroke-dasharray', xDash)
-                .attr('class', 'dropline');
+                .attr('class', 'dropline')
+                .classed('crisp', true);
         }
 
         // X axis marker
```
I thought I would prefer a narrower line, but particularly since it's dashed I ended up concluding I agree with the 3px default.

Before these tweaks:
screen shot 2017-04-07 at 2 03 50 am
screen shot 2017-04-07 at 2 04 05 am

And after:
screen shot 2017-04-07 at 2 32 16 am
screen shot 2017-04-07 at 2 32 26 am

alexcjohnson:
@delekru can you take a look at a modebar icon for toggling these droplines? @rpaskowitz has included one in https://github.com/plotly/plotly.js/pull/1461#issuecomment-292425977 that may be a starting point but I'll defer to your judgment.

alexcjohnson:
And to your notes about behavior @rpaskowitz - it all seems to work as I'd expect, really well done! Not sure I follow the "probably not desirable" part...

I'd be curious to see how this feels if we *were* to enable it for "compare" mode, but definitely not necessary for this PR and could well be a terrible experience.

src/components/modebar/buttons.js
rpaskowitz: I agree, this would be better as a boolean, and is how I had it originally. I wound up changing it to take advantage of the modebar logic which sets the `active` class. Because the `val` on the button is passed through a `data-val` attr on the dom, and the is a strict equality check at https://github.com/plotly/plotly.js/blob/master/src/components/modebar/modebar.js#L211 I found I couldn't use a boolean. Open to suggestions that let me keep this as a boolean, but get to use the state display on the modebar.

rpaskowitz:
I'm mostly ok with the styling tweaks, but I think having the contrast line be an extra pixel wider than the spikeline can still help, otherwise it's possible for the line get a bit lost. For example, in your 'after tweak' screenshot:
image

I think the same could happen for the y-axis spikeline on a bar chart.

> Not sure I follow the "probably not desirable" part...
If someone had done y-axis showspike: true, and x-axis showspike: false, and the chart were originally rendered with hovermode: x -- if the user then changes hovermode to closest, the spikes will be shown. In my current implementation, if they change hovermode back to x, the showspike attributes will all be set to false, so moving back to closest will have them off, even though the user hadn't done any toggle action. To get back to the original they would need to reset the axis. This feels weird to me, but I think I can fix it.

delekru:
@alexcjohnson Looks good! It just needed a tiny bit of fixing for it to be compatible with our fonts.

alexcjohnson:
> I'm mostly ok with the styling tweaks, but I think having the contrast line be an extra pixel wider than the spikeline can still help, otherwise it's possible for the line get a bit lost.

Fair point, it's a nice effect in the light case, but it does bother me in the dark case. Or more precisely, when the contrast line is very different from the plot background, which could even happen to both variants if the plot background was neither black nor white.

So what about making the rear line actually match the plot background (which, to account for potential transparency while making this line itself opaque, can be done by `Color.combine(fullLayout.plot_bgcolor, fullLayout.paper_bgcolor)`)? *That* in turn would fall down when the point color is too close to this color - for example, on a white background if you had a nearly white point - which could still be visible on its own if it had a dark border. In that case we could replace the dashed color with the contrast color? Then the question would be how to tell "too close." `tinycolor` has a utility [`readability`](https://github.com/bgrins/TinyColor#readability) which only uses luminance... that may be sufficient, especially since backgrounds are nearly always somewhere on or near the greyscale. Looks to me like `readability < 1.5` or something like that would be a reasonable cutoff. Only when we're taking this color from the trace of course... if the user specified an explicit `spikecolor` we should honor that even if we think it'll be invisible!

Thoughts? Apologies for the extra complication, just want to make sure we're covering all our bases.

alexcjohnson:
>  In my current implementation, if they change hovermode back to x, the showspike attributes will all be set to false, so moving back to closest will have them off, even though the user hadn't done any toggle action. To get back to the original they would need to reset the axis. This feels weird to me, but I think I can fix it.

Ah I see what you mean now. Yes, that's kind of an obscure edge case but if it's not too cumbersome it would be nice to fix.

etpinard:
peek 2017-04-07 10-26

the icon so far

alexcjohnson:
@etpinard that looks a little small, can we expand it a bit?

etpinard:
peek 2017-04-07 10-36

scaled by 50%

src/lib/index.js
rpaskowitz: Coming in next commit.

src/plots/cartesian/graph_interact.js
rpaskowitz: Both coming in next commit.

rpaskowitz:
Most comments addressed in latest PR. The edge case I described should be fixed, and it has the added bonus of preserving the toggle state, so if you have spikelines enabled then switch out of `closest`, when you return to `closest` the spikelines will still be enabled.

Need to read the comment on background/contrast color in a bit more depth and think about getting that done. I agree that the contrast issue is primarily a problem when it's the chart background that is light, and not the series color itself. For the hover marker, it makes sense to use the contrast color for the font, since the text is drawn on top of the color, but that seems less of an issue here. It's why I preferred just using the white background, but a solution that ensure contrast with the chart background, which may not always be light makes more sense.

src/plots/cartesian/layout_attributes.js
rpaskowitz: Alright, I can change this back. I was thinking of the plural in terms of the axes, not the data points.

src/plots/cartesian/graph_interact.js
```diff
@@ -886,68 +886,86 @@ function createDroplines(hoverData, opts) {
...
-                .attr('y2', yBase)
-                .attr('stroke-width', xThickness)
-                .attr('stroke', xColor)
-                .attr('stroke-dasharray', xDash)
-                .attr('class', 'dropline');
+                .attr({
+                    'x1': xPoint,
+                    'x2': xPoint,
+                    'y1': yEndSpike,
+                    'y2': yBase,
```
alexcjohnson: yBase <-> yEndSpike (for a consistent starting point of the dashes)

rpaskowitz:
Here are some examples where the series color is close to the background color, and using the readability function to swap to using the contrast color instead of the series color:
image
image

I think it works pretty well, and now the spikeline background never stands out, but cases like this make sure it's not lost:
image

With background thickness == spike thickness:
image

src/plots/cartesian/graph_interact.js
rpaskowitz: Done in most recent commit.

alexcjohnson:
Nice, I think it looks fantastic now. @etpinard how do you want to handle the icon? I think that's the only thing left on my list.

etpinard:
> how do you want to handle the icon? I think that's the only thing left on my list.

I'll try pushing to this branch directly. 

@rpaskowitz can you make sure you have the _Allow edits from maintainers_ enabled? See https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ for more info.

rpaskowitz:
Sure. Do you want me to squash anything before you start adding commits on top?

alexcjohnson:
Oh and tests, which you already mentioned at some point @rpaskowitz - I don't think we need too much, just a couple that the attributes get set right on the modebar clicks, and that the right elements show up (or don't show up) on hover.

rpaskowitz:
Github collapsed the comment+reply, but unaddressed what the string you wanted to become a bool. Want to make sure my reply wasn't missed on that.

Allow edits from maintainers looks like it was on, so feel free to add.

src/components/modebar/buttons.js
alexcjohnson: Ah, makes sense. Sure, lets leave it as a string then. Thanks for pointing me back to this.

rpaskowitz:
I've got all but two tests fixed up before I write my own, but the couple failing are a bit of a mystery before I dig in. They do relate to hover, so it could be a new failure, but I figured I'd check if anyone may have a clue before going at it. They are noCI tests, so it could be that I have issues similar to why they were disabled there?

In `gl_plot_interact_test.js` these are failing with what looks like the completely wrong data points are being hovered:
`@noCI should display correct hover labels and emit correct event data`
`@noCI should emit correct event data on click`

Any hints?

etpinard:
@rpaskowitz don't worry too much about those `gl` hover tests. They don't behave well on all hardware at the moment (that's why we skip them on CI hence that `@noCI` tag).

Sorry for the confusion.

rpaskowitz:
Got an image comparison failure on `gl2d_14`, but may need to have project access to download artifacts to see what's up with it?

etpinard:
> Got an image comparison failure on gl2d_14

Hmm. Strange. Let look into this.

src/components/modebar/modebar.js
```diff
@@ -172,8 +172,12 @@ proto.createIcon = function(thisIcon) {
     icon.setAttribute('width', (thisIcon.width / iconHeight) + 'em');
     icon.setAttribute('viewBox', [0, 0, thisIcon.width, iconHeight].join(' '));
 
+    var transform = name === 'toggleSpikelines' ?
```
etpinard: A little hacky here. Ideally, the icon should have been exported 50% bigger. I opted for this hacky fix because implementing some icon `transform` setting as proposed https://github.com/plotly/plotly.js/issues/1335 would be really nice.

src/components/modebar/manage.js
```diff
@@ -138,7 +138,7 @@ function getButtonGroups(gd, buttonsToRemove, buttonsToAdd) {
         addGroup(['hoverClosestGl2d']);
     }
     else if(hasCartesian) {
-        addGroup(['hoverClosestCartesian', 'hoverCompareCartesian']);
+        addGroup(['toggleSpikelines', 'hoverClosestCartesian', 'hoverCompareCartesian']);
```
etpinard: peek 2017-04-07 17-09

Personally, I think I'd prefer placing this new icon first from the right. Thoughts?

rpaskowitz: My reasoning was to put it beside `hovermode: closest` because of how spikeline functionality is tied to that mode. (Needed for any spikelines, and using the toggle will force you into that mode)

etpinard: Good point. I'm ok leaving it there.

src/components/modebar/modebar.js
etpinard: Hmm. I did break something, the new button doesn't seem to work:

peek 2017-04-07 17-28

rpaskowitz: I think that was me. Grab my latest commit.

etpinard:
@rpaskowitz 

> I think that was me. Grab my latest commit.

Confirm. Things are working fine with the latest commit. :+1: 

> Got an image comparison failure on gl2d_14, but may need to have project access to download artifacts to see what's up with it?

Indeed. This has nothing to do with you. The problem is documented in 

https://github.com/gl-vis/gl-scatter2d-sdf/commit/263078a7206d40dbadfb350613bb4701171fbb23#commitcomment-21683726

we'll either wait for a patch in `gl-scatter2d-sdf` or lock down it version in our package.json. At some point on Monday, you'll be able to merge `master` into your branch and get that :white_check_mark: again. Sorry (again) for the inconvenience.

etpinard:
@rpaskowitz I merged master into your branch. Looks like `npm run test-jasmine -- modebar` is failing though. Some tests probably need to be updated for the new modebar button added in this PR.

rpaskowitz:
Yup, I've got fixes for the broken ones, plus a few new modebar tests. Just working on some tests for the spikeline itself now.

rpaskowitz:
Modebar and spikeline tests now added. Fingers crossed for a nice clean build.

Pretty minimal tests on the spikeline drawing itself (plus a small refactor to `hoverChanged` which only fired when evt.target was set, but now I'm re-using that for more than just event firing).

I should have just RTFM on Fx.hover(), as I would've saved a lot of messing around, but in my digging, I may have uncovered a test that is passing only through luck? 

In [hover_label_test.js#L20](https://github.com/plotly/plotly.js/blob/6e547ef794093c206998103f3065d5f2ab2a20b4/test/jasmine/tests/hover_label_test.js#L20) the tests there have `evt.clientX` and `evt.clientY`, but [graph_interact.js#L417](https://github.com/plotly/plotly.js/blob/6e547ef794093c206998103f3065d5f2ab2a20b4/src/plots/cartesian/graph_interact.js#L417) means that if `evt.target` is not set, then `xpx` and `ypx` will be used, and since they are unset, half the length of the axis is used. Since `clientX` and `clientY` were meant to be set to half of the graph anyways, it winds up being the same thing, but for different reasons.

rpaskowitz:
Please hold off on any merge. Some of my arithmetic seems to have broken and lines aren't drawn quite right at the moment.

alexcjohnson:
> I may have uncovered a test that is passing only through luck?

Hah, nice catch @rpaskowitz 🎣 
I wouldn't worry about it here, but if you wouldn't mind can you make an issue for it so we remember to address it?

rpaskowitz:
Erm, not sure about my line issue, I refresh the test page and it went away, and nothing was throwing console errors. I may have loaded last while refactoring dropline->spikeline, and been in a weird middle state - or, my second best guess is it relates to browser resize. But I can't repro, so...things still look good.

rpaskowitz:
Ok, so it is related to browser size/scrolling. Need to dig in a bit. Probably just grabbing the wrong coords as a base.

scroll

rpaskowitz:
Ok, should be better now. May need a bit more testing outside the test framework where the graph can appear anywhere on the page.

src/plots/cartesian/axis_defaults.js
```diff
@@ -75,6 +75,12 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
     coerce('range');
     containerOut.cleanRange();
 
+    coerce2('showspikes');
```
etpinard: Why use `coerce2` here?

rpaskowitz: The short answer is "because it made things work". The slightly longer answer is that with just coerce(), it seems like some of the gl charts wind up including the cartesian axis values, but because the various spike options aren't defined in the gl3d layout options, things exploded pretty badly.

Some things seem to alternately get defined as axes options or layout options, and I saw that coerce2 was defined above to always pass in layout options (which is where these options are defined for the non-gl charts), so it meant that even when this got included by the gl3d charts, it didn't cause issues.

Here's the stack you get on a gl3d chart when this uses `coerce`:
```
Uncaught TypeError: Cannot read property 'dflt' of undefined
    at Object.exports.coerce (coerce.js:242)
    at coerce (axis_defaults.js:30)
    at handleAxisDefaults (axis_defaults.js:81)
    at supplyLayoutDefaults (axis_defaults.js:44)
    at handleGl3dDefaults (defaults.js:92)
    at handleSubplotDefaults (subplot_defaults.js:71)
    at Object.supplyLayoutDefaults (defaults.js:32)
    at Object.plots.supplyLayoutModuleDefaults (plots.js:1050)
    at Object.plots.supplyDefaults (plots.js:475)
    at Object.Plotly.plot (plot_api.js:136)
```

etpinard: I see, thanks for writing down the details @rpaskowitz. Your solution indeed _just works_, but it creates a (small) problem: it throws off `Plotly.validate`, which relies on `Plots.supplyDefaults` to determine if user input is invalid or uselessly defined (see test [suite](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/validate_test.js) for more details). For example, something like:

```js
Plotly.validate(/* */, {
  scene: {
    xaxis: {
        splkedash: 'dash'
    }
  }
}))
```

will be considered valid even though `spikedash` isn't  supported in 3D axes (yet).

-----


So, I'd recommend moving this block to just above this [line](https://github.com/plotly/plotly.js/blob/6e547ef794093c206998103f3065d5f2ab2a20b4/src/plots/cartesian/layout_defaults.js#L184) where it won't be called on 3D axes.

rpaskowitz: Change made and tested locally. Will hold off on commit to see if there's any more feedback.

etpinard: Thanks!

rpaskowitz:
New case I'm fixing for 'across' mode with multiple y-axes.

rpaskowitz:
Is `layout.supplyLayoutDefaults` supposed to be called on drag(pan/zoom) of an axis?

etpinard:
> Is layout.supplyLayoutDefaults supposed to be called on drag(pan/zoom) of an axis?

Yes, on zoom and at the end of (i.e. on mouseup) of pan and drag.

rpaskowitz:
> Yes, on zoom and at the end of (i.e. on mouseup) of pan and drag.

Ok. I think the way I was testing was probably just invalid (I was changing spikemode on the axis directly and not a relayout, so the value was being lost when I did a pan)

alexcjohnson:
I'm making a few edits to this after merging in the latest master (so @rpaskowitz let me know if you want to contribute any more to this branch, otherwise we'll make a merge mess...), also there was still a weird issue when you turn on droplines and then scroll an element the plot is in...

But before I push this in, I wanted to revisit our handling of free axes. The way it works right now the dropline goes to the edge of the plot:

screen shot 2017-04-13 at 2 13 03 pm

There was some discussion a while back eg
> so maybe it would be another configurable option in the showspikes configuration to have the line drawn all the way to the axis for the cases where one knows it's not going to do something ugly like draw over other axes.

But personally I think drawing over other axes is a small price to pay for pointing out which axis we're really referring to, so I'd prefer it this way (at least by default, we could consider adding a config option for it if anyone is really bothered):

screen shot 2017-04-13 at 2 02 30 pm

@rpaskowitz @etpinard thoughts?

There's a further issue where the free axis `side` doesn't match its position relative to the subplot we're hovering on (or in fact if it's in the middle of the subplot), I can fix this easily once we agree which way we go with ^^

etpinard:
>  I think drawing over other axes is a small price to pay for pointing out which axis we're really referring to

I agree with @alexcjohnson here. It's nice to have the spike lines reach to axis where users can easily read off the value.

rpaskowitz:
I think I'm done, so feel free to take the branch over.

> also there was still a weird issue when you turn on droplines and then scroll an element the plot is in

Ugh. I'd take another look, but sounds like you've already got it fixed?

> There's a further issue where the free axis side doesn't match its position relative to the subplot we're hovering on

I think I saw this one one of the mock plots, so will be good to fix up

> But personally I think drawing over other axes is a small price to pay for pointing out which axis we're really referring to, so I'd prefer it this way (at least by default, we could consider adding a config option for it if anyone is really bothered)

I liked the marker as an added way to point out which axis the point is referring to (although it sometimes wound up a bit confusing on those sample charts, where the series colors don't always match the axis colors - which I think is likely, but mainly confusing because some of them _did_ match).

I don't especially dislike the line going all the way to the axis, but think we'd might like a spikemode to control the behavior. Notably, in the `across` mode, I don't think this behavior would make sense because it would no longer be helping identify which axis the number is actually plotted on, and the spikeline is more useful for the within-plot point comparison.

alexcjohnson:
> I don't especially dislike the line going all the way to the axis, but think we'd might like a spikemode to control the behavior. Notably, in the across mode, I don't think this behavior would make sense because it would no longer be helping identify which axis the number is actually plotted on, and the spikeline is more useful for the within-plot point comparison.

True. When we think about free axes, or even anchored axes that might be positioned between two subplots, I suspect there are a bunch of edge cases we're not considering yet. That said, this covers 99% of the cases and won't do *terrible* things in the last few, so I'll try to get this merged today and we can revisit these cases later.

alexcjohnson:
@etpinard @rpaskowitz I'm happy with this now - any comments on these commits before I merge?

rpaskowitz:
Gave the code a brief review, and played with a bunch of charts. Looks to be working as expected.

src/plot_api/plot_api.js
```diff
@@ -2124,14 +2124,16 @@ function _relayout(gd, aobj) {
...
+             * affect reaction to user input. Everything else, assume full replot.
              * height, width, autosize get dealt with below. Except for the case of
              * of subplots - scenes - which require scene.updateFx to be called.
              */
-            else if(['hovermode', 'dragmode'].indexOf(ai) !== -1) flags.domodebar = true;
-            else if(['hovermode', 'dragmode', 'height',
-                'width', 'autosize'].indexOf(ai) === -1) {
+            else if(['hovermode', 'dragmode'].indexOf(ai) !== -1 ||
+                    ai.indexOf('spike') !== -1) {
+                flags.domodebar = true;
```
etpinard: :racehorse:

src/plots/cartesian/layout_attributes.js
```diff
@@ -279,6 +279,59 @@ module.exports = {
...
+        role: 'style',
+        description: 'Sets the width (in px) of the zero line.'
+    },
+    spikedash: {
+        valType: 'string',
+        // string type usually doesn't take values... this one should really be
+        // a special type or at least a special coercion function, from the GUI
+        // you only get these values but elsewhere the user can supply a list of
+        // dash lengths in px, and it will be honored
+        values: ['solid', 'dot', 'dash', 'longdash', 'dashdot', 'longdashdot'],
```
etpinard: Non- :no_entry: , but It would be nice to reuse the list in `traces/scatter/attributes.js`

src/plots/cartesian/layout_attributes.js
```diff
@@ -279,6 +279,59 @@ module.exports = {
...
+        role: 'style',
+        description: [
+            'Sets the style of the lines. Set to a dash string type',
+            'or a dash length in px.'
+        ].join(' ')
+    },
+    spikemode: {
+        valType: 'flaglist',
+        flags: ['toaxis', 'across', 'marker'],
+        extras: ['none'],
```
etpinard: Isn't `spikemode: 'none'` the same as `showspikes: false`?

src/plots/cartesian/layout_defaults.js
```diff
@@ -173,6 +173,12 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
 
         handleAxisDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions, layoutOut);
 
+        coerce('showspikes');
```
etpinard: When `showspikes: false`, the attributes below shouldn't get coerced (as they don't do anything).

etpinard:
@alexcjohnson @rpaskowitz I'm a little concerned about https://github.com/plotly/plotly.js/pull/1461#discussion_r111743208 as commit https://github.com/plotly/plotly.js/pull/1461/commits/69dc78147de4a3255225a1878e653800c6092107 _fixed_ manual hover event emitting, but no tests appeared to fail previous to that commit.

src/plots/cartesian/graph_interact.js
```diff
@@ -629,7 +629,7 @@ function hover(gd, evt, subplot) {
     }
 
     // don't emit events if called manually
-    if(evt.target && !hoverChanged(gd, evt, oldhoverdata)) return;
+    if(!evt.target || !hoverChanged(gd, evt, oldhoverdata)) return;
```
alexcjohnson: now it did ;) https://github.com/plotly/plotly.js/pull/1461/commits/052417bb27cdcaffe449b47bac8f9b7d9a81161f

I'll note in case this confuses someone later: a bunch of the other tests in the `hover_label_test` suite look like they're providing pixel positions but are actually hitting [this block](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/graph_interact.js#L419) that takes the middle of the x axis (and equivalently for y just below)

src/traces/ohlc/attributes.js
```diff
@@ -87,9 +88,9 @@ module.exports = {
                 '`decreasing.line.width`.'
             ].join(' ')
         }),
-        dash: Lib.extendFlat({}, lineAttrs.dash, {
+        dash: Lib.extendFlat({}, dash, {
             description: [
-                lineAttrs.dash,
```
alexcjohnson: previously:
screen shot 2017-04-17 at 5 38 03 pm

etpinard: Thanks very much!

test/jasmine/tests/hover_label_test.js
```diff
@@ -532,6 +533,44 @@ describe('hover info', function() {
...
+        var data = [{x: [1, 2, 3], y: [1, 3, 2], type: 'bar'}];
+        var layout = {width: 600, height: 400};
+        var gd;
+
+        beforeEach(function(done) {
+            gd = createGraphDiv();
+            Plotly.plot(gd, data, layout).then(done);
+        });
+
+        fit('should emit events only if the event looks user-driven', function(done) {
```
etpinard: :sos: `fit` :sos:

alexcjohnson: yeah, linter caught it...

etpinard: Nice test :tada:

src/plots/cartesian/layout_attributes.js
alexcjohnson: I cleaned up a bunch of things in https://github.com/plotly/plotly.js/pull/1461/commits/bd1156704c022f7ba3ea425e5cda82df4c0b06e5
@etpinard can you take a look at this? I tried to turn it into a regular `enumerated` attribute in cases where the manual dasharray isn't allowed (gl) and removed it where it's not supported at all (mapbox). I don't think this syntax is tested anywhere though... I suppose in principle we should provide mocks that use it for all the places it's used in SVG.

etpinard:
:dancer: when tests :white_check_mark:

src/plots/cartesian/layout_attributes.js
etpinard: Looking good :+1:

src/traces/scattermapbox/defaults.js
```diff
@@ -26,14 +26,14 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
     }
 
-    function coerceMarker(attr, dflt) {
-        var attrs = (attr.indexOf('.line') === -1) ? attributes : scatterAttrs;
+    // function coerceMarker(attr, dflt) {
```
etpinard: :hocho: :hocho:

PR #1465, opened 2017-03-10
Rework groupby style syntax into array

rreusser:
As discussed, groupby transform styling permits arbitrary keys which, in addition to not fitting into the plot schema, will fail if a group ends with `*src`. This PR starts the switch to array-of-objects style by first simply breaking all of the tests.

### DIFF
--- a/src/transforms/groupby.js
+++ b/src/transforms/groupby.js
@@ -35,14 +35,24 @@ exports.attributes = {
         ].join(' ')
     },
     style: {
-        valType: 'any',
-        dflt: {},
-        description: [
-            'Sets each group style.',
-            'For example, with `groups` set to *[\'a\', \'b\', \'a\', \'b\']*',
-            'and `style` set to *{ a: { marker: { color: \'red\' } }}',
-            'marker points in group *\'a\'* will be drawn in red.'
-        ].join(' ')
+        target: {
+            valType: 'any',
+            role: 'info',
+            description: [
+                'The group value which receives these styles.'
+            ].join(' ')
+        },
+        value: {
+            valType: 'any',
+            role: 'info',
+            dflt: {},
+            description: [
+                'Sets each group style.',
+                'For example, with `groups` set to *[\'a\', \'b\', \'a\', \'b\']*',
+                'and `style` set to *{ a: { marker: { color: \'red\' } }}',
+                'marker points in group *\'a\'* will be drawn in red.'
+            ].join(' ')
+        },
     }
 };
 
@@ -71,10 +81,26 @@ exports.supplyDefaults = function(transformIn) {
     if(!enabled) return transformOut;
 
     coerce('groups');
-    coerce('style');
+
+    var styleIn = transformIn.style;
+    var styleOut = transformOut.style = [];
+
+    if(Array.isArray(styleIn)) {
+        // If styles are an array, then sanitize them against the schema:
+        for(var i = 0; i < styleIn.length; i++) {
+            styleOut[i] = {};
+            Lib.coerce(styleIn[i], styleOut[i], exports.attributes.style, 'target');
+            Lib.coerce(styleIn[i], styleOut[i], exports.attributes.style, 'value');
+        }
+    } else {
+        // Otherwise, this is deprecated keyed-object-style styles and we
+        // just pass the object through as the index itself:
+        transformOut.style = transformIn.style;
+    }
 
     return transformOut;
 };
+
 
 /**
  * Apply transform !!!
@@ -115,6 +141,7 @@ function pasteArray(newTrace, trace, j, a) {
 }
 
 function transformOne(trace, state) {
+    var i;
     var opts = state.transform;
     var groups = trace.transforms[state.transformIndex].groups;
 
@@ -128,9 +155,20 @@ function transformOne(trace, state) {
 
     var arrayAttrs = PlotSchema.findArrayAttributes(trace);
 
-    var style = opts.style || {};
+    var style = opts.style || [];
+    var styleIndex = style;
 
-    for(var i = 0; i < groupNames.length; i++) {
+    // If the styles are an array, then translate it into an index keyed on the
+    // target. Otherwise just use the object provided for the sake of backward
+    // compatibility.
+    if(Array.isArray(style)) {
+        styleIndex = {};
+        for(i = 0; i < style.length; i++) {
+            styleIndex[style[i].target] = style[i].value;
+        }
+    }
+
+    for(i = 0; i < groupNames.length; i++) {
         var groupName = groupNames[i];
 
         var newTrace = newData[i] = Lib.extendDeepNoArrays({}, trace);
@@ -145,9 +183,9 @@ function transformOne(trace, state) {
 
         newTrace.name = groupName;
 
-        // there's no need to coerce style[groupName] here
+        // there's no need to coerce styleIndex[groupName] here
         // as another round of supplyDefaults is done on the transformed traces
-        newTrace = Lib.extendDeepNoArrays(newTrace, style[groupName] || {});
+        newTrace = Lib.extendDeepNoArrays(newTrace, styleIndex[groupName] || {});
     }
 
     return newData;
--- a/src/transforms/groupby.js
+++ b/src/transforms/groupby.js
@@ -49,7 +49,7 @@ exports.attributes = {
             description: [
                 'Sets each group style.',
                 'For example, with `groups` set to *[\'a\', \'b\', \'a\', \'b\']*',
-                'and `style` set to *{ a: { marker: { color: \'red\' } }}',
+                'and `style` set to *[{target: \'a\': { marker: { color: \'red\' } }}]',
                 'marker points in group *\'a\'* will be drawn in red.'
             ].join(' ')
         },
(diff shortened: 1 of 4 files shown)


src/transforms/groupby.js
```diff
@@ -35,14 +35,24 @@ exports.attributes = {
         ].join(' ')
     },
     style: {
```
rreusser: @etpinard @alexcjohnson Comparing this to similar cases, I'm not 100% sure what `_isLinkedToArray` does or whether it's needed here. I've omitted it.

src/transforms/groupby.js
```diff
@@ -35,14 +35,24 @@ exports.attributes = {
...
-        valType: 'any',
-        dflt: {},
-        description: [
-            'Sets each group style.',
-            'For example, with `groups` set to *[\'a\', \'b\', \'a\', \'b\']*',
-            'and `style` set to *{ a: { marker: { color: \'red\' } }}',
-            'marker points in group *\'a\'* will be drawn in red.'
-        ].join(' ')
+        target: {
+            valType: 'any',
```
rreusser: Does `valType: 'any'` make sense or should it be strings? As-is styling by numbers is implicitly converted to a string-keyed object anyway, so I'm not 100% sure the right answer here.

rreusser: Maybe weird, maybe not:

screen shot 2017-03-10 at 12 18 01

src/transforms/groupby.js
```diff
@@ -128,9 +155,20 @@ function transformOne(trace, state) {
...
     var arrayAttrs = PlotSchema.findArrayAttributes(trace);
 
-    var style = opts.style || {};
+    var style = opts.style || [];
+    var styleIndex = style;
+
+    // If the styles are an array, then translate it into an index keyed on the
+    // target. Otherwise just use the object provided for the sake of backward
+    // compatibility.
+    if(Array.isArray(style)) {
```
rreusser: Note: I could have translated the object in to array style format to get a nice consistent result on `fullData`, except the useful format is actually as originally defined since it needs a lookup so that it would just have to be translated *back* to the key-value style down below. That way would be a bit cleaner and may need to happen; this way just minimizes the back and forth. No strong feelings here.

alexcjohnson: backward-compatibility fixes should go in [cleanData](https://github.com/plotly/plotly.js/blob/master/src/plot_api/helpers.js#L245) so they only happen when new data comes in, and they change the data to current format.

src/transforms/groupby.js
alexcjohnson: not so weird:
```js
0.20000000000000001 === 0.2
true
```

We definitely want numbers treated as strings for these purposes, so `target: 1` should match `'1'` in the data, and vice versa.

`valType: 'string'` might make sense to help accomplish this - with `strict` omitted so numbers get stringified. That would cause us to ignore array, object, null, and undefined targets, which seems right. All else equal I might have permitted booleans to pass through (stringified) but if it simplifies things - which it looks like it does - then lets leave those out.

src/transforms/groupby.js
alexcjohnson: I should probably let @etpinard handle this but I think you want both `_isLinkedToArray` and `_arrayAttrRegexps` - the latter of which won't get used yet but when we extend the array editing stuff I did in `relayout` to `restyle` it will be important.

see https://github.com/plotly/plotly.js/blob/master/src/components/updatemenus/attributes.js#L17-L52

src/transforms/groupby.js
rreusser: Failed to initially appreciate the full meaning of "only when new data comes in." Moved to cleanData. 👍

src/transforms/groupby.js
rreusser: That was my template, but I had a hard time tracking down exactly what it *does*. See: [registry.js#L100-L105](https://github.com/plotly/plotly.js/blob/master/src/registry.js#L100-L105). It seems like since `groupby` has no `layoutAttributes`, this key will have no effect.

src/transforms/groupby.js
rreusser: Sounds good. I guess my discomfort is just in relying upon implicit conversion of keys to strings to sort out numerical identity. Like if your group were `0.2` and your style key were computed in JS as `0.3 - 0.1`, I don't think it would find the right style.

rreusser: Not really a problem in JSON, but a little weird in plain JS API usage:

screen shot 2017-03-10 at 13 54 48

src/transforms/groupby.js
alexcjohnson: right, at the moment `_arrayAttrRegexps` doesn't do anything inside traces, it's just a question of if we want to put it in so we don't forget about this when we *do* get to implementing it. Maybe better not to include it, in case we implement it a different way. I think `_isLinkedToArray` matters to [plot_schema](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_schema.js) already though, doesn't it?

src/transforms/groupby.js
alexcjohnson: I think that would still be a problem with numeric keys, unless you implement some sort of fuzzy equality, which sounds even more dangerous.

src/transforms/groupby.js
rreusser: Added `_isLinkedToArray` but did not add `_arrayAttrRegexps`, which I think only applies to *nested* array attributes like `updatemenus[...].buttons`. Technically this is `transforms[...].styles`, so I'm not totally sure where this would need to be added. At the transform level vs. *inside* styles, perhaps? Glad to add it blindly, but without really knowing how to test it, I think maybe better omit.

src/transforms/groupby.js
rreusser: Agreed. FWIW [`almost-equal`](https://github.com/scijs/almost-equal) combines absolute and relative tolerance in a way that seems generally desirable, but I think definitely best to ignore it and just keep in mind that it's a little picky.

rreusser:
Noticed a small typo in the style docs. No rush on the review, but I'm content with this for now.

rreusser:
Failure appears to be a fluke due to cholorpleth, perhaps? I'll push a dummy commit in a bit unless someone wants to hit the rebuild button.

screen shot 2017-03-10 at 15 23 34

test/jasmine/tests/transform_groupby_test.js
```diff
@@ -401,17 +467,20 @@ describe('groupby', function() {
...
                                 color: 'red'
                             }
                         }
-                    },
-                    b: {
+                    }
+                }, {
+                    target: 'b',
+                    value: {
                         mode: 'markers+lines', // heterogeonos attributes are OK: group 'a' doesn't need to define this
```
alexcjohnson: not that this is new, but heterogeneous

But that makes me wonder: you don't need to supply `style` for every group, do you? If you omit it entirely for a group, you just get the styling contained in the parent trace, yes? Want to test that?

This has an interesting use case, I think, that wouldn't have worked (at least not nearly as well) the old way: you could have a "highlight" style to do something special to one group you're focusing on, and to focus on a different group you just change `target`. And if you want to focus on none, change `target` to some value that's not a group at all (does this work or does it complain?)

Also, there's a case we couldn't have previously: what if two targets are the same? I could imagine wanting to do this in exactly the "focus style" example... you have styles for each group (say, different marker colors) but then there's a different style you want to move between groups... in that case the actual style to apply to a given group would be the composition of all styles targeting the group. Should we support this? Or do something else if you get the same target more than once?

src/transforms/groupby.js
etpinard: Maybe we should name this attributes `styles` instead? All (I think) `isLinkedToArray` attribute have plural names.

test/jasmine/tests/transform_groupby_test.js
etpinard: > If you omit it entirely for a group, you just get the styling contained in the parent trace, yes? Want to test that?

That should be already part of the suite. But yes, we should test that.

> in that case the actual style to apply to a given group would be the composition of all styles targeting the group. Should we support this? Or do something else if you get the same target more than once?

I don't see a problem with supporting multiple styles item with the same `target`. Style items with the same `target` would simply be applied in sequence to the transformed traces.

src/transforms/groupby.js
rreusser: Is that an acceptable change for cleanData?

test/jasmine/tests/transform_groupby_test.js
rreusser: @alexcjohnson No, you don't need to specify for all. Or to be more precise, you don't need to now *if* you didn't before. 😄 

At the moment, I'm internally transforming the style array into a key-value index that matches the old specification format. So if two are the same, the last in the array will overwrite other instances. I think merging would be an appropriate use case mergeDeepNoArrays. I don't personally find it a *particularly* compelling use-case, but it's a pretty straightforward change to add support for pretty weird user input.

src/transforms/groupby.js
etpinard: Yes :+1:

rreusser:
Trying to close out PRs. Reading through this, I'm not sure if I have anything to add to it.

etpinard:
> Trying to close out PRs. Reading through this, I'm not sure if I have anything to add to it.

Looks like https://github.com/plotly/plotly.js/pull/1465/files#r105486264 was never addressed.

Issue #1470, opened 2017-03-13
Range sliders don't work when anchored to log y-axes


etpinard:
See http://codepen.io/etpinard/pen/dvRmGv

First reported on https://community.plot.ly/t/range-slider-doesnt-display-trace-when-y-axis-set-to-log-scale/3627

PR #1472, opened 2017-03-14
Range slider fixes (user set y axis types + trace clearance)

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1470

Easy.


src/components/rangeslider/draw.js
```diff
@@ -376,6 +376,7 @@ function drawRangePlot(rangeSlider, gd, axisOpts, opts) {
         };
 
         mockFigure.layout[oppAxisName] = {
+            type: oppAxisOpts.type,
```
etpinard: Before, we were strictly relying on the axis-auto-type routines in `Plots.supplyDefaults` to determine the range plot y-axis type - which obviously fails when a overrides the y axis type (e.g. in the case of log y axes as presented in https://github.com/plotly/plotly.js/issues/1470)

test/jasmine/tests/range_slider_test.js
```diff
@@ -345,6 +345,90 @@ describe('the range slider', function() {
             })
             .then(done);
         });
+
+        it('should clear traces in range plot when needed', function(done) {
```
etpinard: @alexcjohnson This commit fixes part 1

> the data is gone from the graph, it should disappear from the range slider too. If you delete any but the last trace, the range slider does update correctly, but not when you delete the last one.

of https://github.com/plotly/plotly.js/issues/1473 in a pretty ugly way, unfortunately.

Most trace clearance (i.e. deletion or `visible: false` or `type`-restyle) is handled via this [block](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/index.js#L104-L106):

```js
plotinfo.plot.selectAll('g:not(.scatterlayer)').selectAll('g.trace').remove();
```
except for scatter, heatmap and contour traces. That is, part 1 of #1473 actually _only_ affected scatter, heatmap and contour traces.

Scatter traces are exempted from that `selectAll('').remove()` block since @rreusser's animation PR to smooth out their data updates. They are cleared [here](https://github.com/plotly/plotly.js/blob/88ccce462bea3b38eeb4995937ddfed210b06acc/src/plots/cartesian/index.js#L150). Contour and heatmap groups don't always match one-to-one with trace items (e.g. a contour trace can have a heatmap fill group) so we use their `uid` to tell which trace to clear [here](https://github.com/plotly/plotly.js/blob/88ccce462bea3b38eeb4995937ddfed210b06acc/src/plot_api/plot_api.js#L291), [here](https://github.com/plotly/plotly.js/blob/88ccce462bea3b38eeb4995937ddfed210b06acc/src/plots/plots.js#L581) [here](https://github.com/plotly/plotly.js/blob/88ccce462bea3b38eeb4995937ddfed210b06acc/src/traces/heatmap/plot.js#L38) and [here](https://github.com/plotly/plotly.js/blob/88ccce462bea3b38eeb4995937ddfed210b06acc/src/traces/contour/plot.js#L59).

In brief, this commit ensure that every special `.remove()` call mentioned in the previous paragraph clear traces in the range slider containers as well.

src/plots/plots.js
```diff
@@ -571,23 +571,25 @@ plots.cleanPlot = function(newFullData, newFullLayout, oldFullData, oldFullLayou
...
-            ).remove();
+            oldFullLayout._paper.selectAll(query).remove();
         }
 
-        // clean old colorbars
+        // clean old colorbars and range slider plot
         if(hasInfoLayer) {
             oldFullLayout._infolayer.selectAll('.cb' + oldUid).remove();
+
+            oldFullLayout._infolayer.selectAll('g.rangeslider-container')
```
etpinard: I chose this pattern

```
fullLayout._infolayer.selectAll('g.rangeslider-container')
```

over keeping a ref to some underscore key e.g. `fullLayout._rangeSliders` to make these additions work even when there are no range sliders on the graph without having to resort to `Registry`.

test/jasmine/tests/range_slider_test.js
alexcjohnson: ugly but... ya gotta do what ya gotta do. Nice tests. 👍

alexcjohnson:
looks good to me - 💃  as soon as you update the test image based on #1475 (finishing the loose end there 🎉 )

test/image/mocks/range_slider_multiple.json
```diff
@@ -2,7 +2,7 @@
   "data": [
     {
       "x": [ 1, 2, 3 ],
-      "y": [ 4, 5, 6 ],
+      "y": [ 4, 5e5, 6e8 ],
```
etpinard: @alexcjohnson here's `bar` + log image test :white_check_mark:

PR #1455, opened 2017-03-09
CircleCI with Ubuntu 14.04, 2x parallelism and WebGL jasmine tests 

etpinard:
resolves https://github.com/plotly/plotly.js/issues/241

### Big news in CircleCI land !!!

Our tests are now running on an [Ubuntu 14.04](https://circleci.com/docs/1.0/build-image-trusty/) build image on CircleCI. Previously we were using a (now very outdated) Ubuntu 12.04 build image.

This build image comes with [Chrome 54](https://circleci.com/docs/1.0/build-image-trusty/#google-chrome) by installed by default. This is big news because this Chrome version can create a WebGL context using the `--ignore-gpu-blacklist` (which we were setting previously, but wasn't functional). That is, **our WebGL jasmine tests can now run on CI** :tada: - which for all intents and purposes close #241.

With the WebGL suites `npm run test-jasmine` now takes more than 5 minutes to run on CI, bringing our total CI run time to above 15 minutes which is kind of slow :turtle: . Needless to say, this was a good time to add [2x parallelism](https://circleci.com/gh/plotly/plotly.js/edit#parallel-builds) to our CI builds  :racing_car: . With 2x parallelism the total build times are back to about [10 minutes](https://circleci.com/gh/plotly/plotly.js/3751). See commit https://github.com/plotly/plotly.js/commit/14621e122b5766e9a6360f0192cdd435cac40afd for more info.

Unfortunately, some jasmine tests still need to be skipped namely the `mapbox` suites (which are very resource intensive and end up blowing up the process) and similar for some `parcoords` tests (cc @monfera). Luckily the [`karma-spec-tags`](https://github.com/mnasyrov/karma-jasmine-spec-tags) plugin made skipping those few tests on CI very easy :cocktail:

PR #1478, opened 2017-03-15
Skip gl3d hover and click test on CI

etpinard:
... for now. It's been a rough day on CI for those two tests. 

I could not find a way to make these tests more robust. Even translating the technique of [`gl2d_click_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/gl2d_click_test.js#L61) didn't work. I think we're not mocking the mouse events that trigger hover in the 3d [code](https://github.com/plotly/plotly.js/blob/b1b525ee042fd423a11801c2cca8a83c10e04da4/src/plots/gl3d/scene.js#L67-L124) correctly. At the moment we're simply triggering a [mouseover](https://github.com/plotly/plotly.js/blob/b1b525ee042fd423a11801c2cca8a83c10e04da4/test/jasmine/tests/gl_plot_interact_test.js#L45) event. So, let's skip them for now.

### DIFF
--- a/test/jasmine/tests/gl_plot_interact_test.js
+++ b/test/jasmine/tests/gl_plot_interact_test.js
@@ -78,7 +78,7 @@ describe('Test gl3d plots', function() {
         destroyGraphDiv();
     });
 
-    it('should display correct hover labels and emit correct event data', function(done) {
+    it('@noCI should display correct hover labels and emit correct event data', function(done) {
         var _mock = Lib.extendDeep({}, mock2);
 
         function _hover() {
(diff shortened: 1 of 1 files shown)


etpinard:
Oh boy, looks like another test is failing :rage: 

image

... maybe I rushed into merging that WebGL jasmine test PR.

etpinard:
another one:

image

this is getting ridiculous. Starting to feel like reverting https://github.com/plotly/plotly.js/pull/1455 :cry:

alexcjohnson:
Yep, lets do it. Gotta get a handle on CI and worry about getting these to work repeatably later. 💃

PR #1479, opened 2017-03-15
Let the regl object tree be garbage collected on `Plotly.restyle`

monfera:
Having tried more verbose approaches such as keeping around both the `<canvas>` element and the `regl` object, or just the `<canvas>` element, the simple and effective route from a memory and object lifecycle point of view was to let the previously created `regl` object be garbage collected. This is achieved by ensuring a fresh `<canvas>` is created and the previous one gets garbage collected (again, after trying some alternatives), including its attached `gl` context and `regl` object.

Also, the snapshot making code should no longer remove the canvas as it's not kept around on rerendering.

### DIFF
--- a/src/traces/parcoords/base_plot.js
+++ b/src/traces/parcoords/base_plot.js
@@ -61,7 +61,6 @@ exports.toSVG = function(gd) {
         });
     }
 
-    imageRoot.selectAll('*').remove();
     canvases.each(canvasToImage);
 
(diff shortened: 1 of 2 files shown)


monfera:
Getting a seemingly unrelated CI failure on this 3 times in a row, mostly [the `gl3d_z-range.svg` thingie](https://github.com/plotly/plotly.js/pull/1423#issuecomment-286630436)

PR #1481, opened 2017-03-15
Fix shapes and images referencing a missing subplot

alexcjohnson:
Soooo many edge cases... This one comes up in the workspace trying to move shapes or images between unlinked subplots, but it can also happen on initial draw - for example you put data on `xy` and `x2y2` but put a shape on `xy2` - not sure why you'd do this but it should be allowed, and certainly shouldn't error as it does now.

Meanwhile on the workspace side I'll just move these components to `layer: 'above'` - which may be less confusing there anyway since much of the time the objects would get hidden behind a plot otherwise.

@etpinard please review.

### DIFF
--- a/src/components/shapes/draw.js
+++ b/src/components/shapes/draw.js
@@ -75,10 +75,17 @@ function drawOne(gd, index) {
         drawShape(gd._fullLayout._shapeLowerLayer);
     }
     else {
-        var plotinfo = gd._fullLayout._plots[options.xref + options.yref],
-            mainPlot = plotinfo.mainplot || plotinfo;
-
-        drawShape(mainPlot.shapelayer);
+        var plotinfo = gd._fullLayout._plots[options.xref + options.yref];
+        if(plotinfo) {
+            var mainPlot = plotinfo.mainplot || plotinfo;
+            drawShape(mainPlot.shapelayer);
+        }
+        else {
+            // Fall back to _shapeLowerLayer in case the requested subplot doesn't exist.
+            // This can happen if you reference the shape to an x / y axis combination
+            // that doesn't have any data on it (and layer is below)
+            drawShape(gd._fullLayout._shapeLowerLayer);
+        }
     }
 
     function drawShape(shapeLayer) {
--- a/test/jasmine/tests/layout_images_test.js
+++ b/test/jasmine/tests/layout_images_test.js
@@ -134,6 +134,31 @@ describe('Layout images', function() {
             checkLayers(0, 0, 1);
         });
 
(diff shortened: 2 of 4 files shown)


test/jasmine/tests/layout_images_test.js
```diff
@@ -107,7 +107,7 @@ describe('Layout images', function() {
         it('should draw images on the right layers', function() {
 
             Plotly.plot(gd, data, { images: [{
-                source: 'imageabove',
```
alexcjohnson: These were generating `404` warnings when you run the tests, which seem distracting and necessary, so I just put in `jsLogo` everywhere.

test/jasmine/tests/shapes_test.js
```diff
@@ -103,75 +103,75 @@ describe('Test shapes defaults:', function() {
     });
 });
 
-describe('Test shapes:', function() {
-    'use strict';
+function countShapesInLowerLayer(gd) {
+    return gd._fullLayout.shapes.filter(isShapeInLowerLayer).length;
+}
```
alexcjohnson: Moved these helpers to the root level so I could use them in a different `describe` block (which wants to set up a different plot in `gd`)

test/jasmine/tests/layout_images_test.js
etpinard: Nice touch :ok_hand:

src/components/shapes/draw.js
```diff
@@ -75,10 +75,17 @@ function drawOne(gd, index) {
...
+        var plotinfo = gd._fullLayout._plots[options.xref + options.yref];
+        if(plotinfo) {
+            var mainPlot = plotinfo.mainplot || plotinfo;
+            drawShape(mainPlot.shapelayer);
+        }
+        else {
+            // Fall back to _shapeLowerLayer in case the requested subplot doesn't exist.
+            // This can happen if you reference the shape to an x / y axis combination
+            // that doesn't have any data on it (and layer is below)
+            drawShape(gd._fullLayout._shapeLowerLayer);
```
etpinard: Looks good.

I see that annotations don't need a similar fix. @alexcjohnson Can I ask why?

alexcjohnson: Annotations don't have the concept of layer. They're always on top.

etpinard:
Thanks again for making shapes/images :bullettrain_front: proof!

:dancer: my comment is non-blocking.

src/components/shapes/draw.js
etpinard: Right. That makes sense :+1:

PR #1482, opened 2017-03-15
More scattergl symbols

destradafilm:
Added 49 more symbols to get as close to the 'scatter' plot symbol list to include : 

    circle
    circle-open
    square
    square-open
    diamond
    diamond-open
    cross
    x
    triangle-up
    triangle-up-open
    triangle-down
    triangle-down-open
    triangle-left
    triangle-left-open
    triangle-right
    triangle-right-open
    triangle-ne
    triangle-ne-open
    triangle-nw
    triangle-nw-open
    triangle-se
    triangle-se-open
    triangle-sw
    triangle-sw-open
    pentagon
    pentagon-open
    hexagon
    hexagon-open
    hexagon2
    star
    star-open
    hexagram
    diamond-tall
    diamond-tall-open
    hourglass
    hourglass-open
    bowtie
    bowtie-open
    circle-cross
    circle-x
    square-cross
    square-x
    cross-thin
    asterisk

previous list:
screen shot 2017-03-16 at 10 07 05 am

updated list:
screen shot 2017-03-16 at 10 05 12 am


etpinard:
Thanks a lot!

We should make these on-par with the SVG symbol [list](https://github.com/plotly/plotly.js/blob/master/src/components/drawing/symbol_defs.js) or at the very least we shouldn't add `scattergl`-only symbols.

etpinard:
... and by the way please we don't need those `devtools/` modifications as part of this PR.

destradafilm:
@etpinard Sounds good, I reverted the devtools folder back to master.

I attempted to find all the unicode symbols that I could to match the SVG symbol list, but couldn't find all of them, so this is the best I could do. 

I'm wondering, would it make sense to render the symbols on the `scattergl` from the same SVGs? Or is there a reason it's using Unicode symbols?

destradafilm:
And, I actually counted wrong, there's 35 more symbols so a total of 43

src/constants/gl_markers.js
```diff
@@ -16,6 +16,42 @@ module.exports = {
...
+    'hourglass-open': '⧖',
+    'bowtie': '⧓',
+    'bowtie-open': '⋈',
+    'circle-cross': '⊕',
+    'circle-x': '⊗',
+    'square-cross': '⊞',
+    'square-x': '⊠',
+    'cross-thin': '+',
+    'asterisk': '✱'
+};
```
etpinard: @destradafilm try running `npm run lint-fix`

etpinard:
> I attempted to find all the unicode symbols that I could to match the SVG symbol list, but couldn't find all of them, so this is the best I could do.

That's ok. We don't have to have perfect parity between WebGL and SVG symbol. Like I said in https://github.com/plotly/plotly.js/pull/1482#issuecomment-286928882, we should just make sure that we don't introduce WebGL-only markers - which I think you did, but we should double check possible by adding a [jasmine](https://github.com/plotly/plotly.js/tree/master/test/jasmine/tests) test.

src/constants/gl_markers.js
```diff
@@ -16,6 +16,42 @@ module.exports = {
     'square-open': '□',
     diamond: '◆',
     'diamond-open': '◇',
-    cross: '+',
-    x: '❌'
-};
+    cross: '➕',
```
etpinard: Looks like you changed the `cross` symbol which is breaking the [`gl3d_marker_arrays`](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/gl3d_marker-arrays.json) and the [`gl3d_projection_traces`](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/gl3d_projection-traces.json) mocks.

src/constants/gl_markers.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     'square-open': '□',
     diamond: '◆',
     'diamond-open': '◇',
-    cross: '➕',
+    cross: '+',
```
destradafilm: I changed cross back to original symbol, so that means I removed 'cross-thin'

etpinard: Thanks. Reverting back to the current symbol is the way to go here. 

That said, we could (and should I'd say) make `cross` thicker and add `cross-thin` in [`v2`](https://github.com/plotly/plotly.js/issues/420).

etpinard:
Ok @destradafilm before merging we should make the new WebGL symbols have image test support.

Can you try adding an `scatter3d` and `scattergl` data/layout mock in [`test/image/mock/`](https://github.com/plotly/plotly.js/tree/master/test/image/mocks) showing off all the new symbols?

If you're up for a challenge, you can also try generating the baseline images using our image test docker [container](https://github.com/plotly/plotly.js/blob/master/test/image/README.md).

etpinard:
@destradafilm Are you ok with taking care of https://github.com/plotly/plotly.js/pull/1482#issuecomment-288230265 or do you me to do it? This would make a nice addition to next week's `1.26.0`.

destradafilm:
@etpinard I've been too busy in the past few days to work on it. If you're willing to do it, that would be great! If not, it will probably have to be the end of next week before I can get to it.

etpinard:
@destradafilm no problem. I'll take care of it (should only be a 5-minute job)! 

Thanks again for the PR :beers:

etpinard:
Superseded by https://github.com/plotly/plotly.js/pull/1550

src/constants/gl_markers.js
etpinard: Added to https://github.com/plotly/plotly.js/issues/420

Issue #1480, opened 2017-03-15
Marker styles are not updated with Plotly.restyle in zoomed state until double-click to zoom out


kimsk:
When calling `Plotly.restyle` in zoomed state, the markers style are not updated in UI until I double-click to zoom out.

Codepen: https://codepen.io/karlkim/pen/OpxMxG?editors=1010#0

> Click one of the buttons to trigger `Plotly.restyle` which update style of markers

### Steps to reproduce:

0. Initially, the red marker has `opacity` = 1 and others have `opacity` = 0.3
1. Click `1`, and the blue marker should have `opacity` = 1, and the red marker should have `opacity` = 0.3
2. Zoom into the area around red marker and click `0`.

### Expected Result:

The red marker should have `opacity` = 1 and the blue marker should have `opacity` = 0.3

### Actual Result:

The blue marker still has `opacity` = 1. The marker styles are not updated until I double-click to zoom out.

etpinard:
Thanks for writing in. As a lot of people do, you not providing the right input to `restyle`. Restyle set key-value updates - not a object meant to extend a trace. To update nested keys, use e.g. `marker.opacity: 0.4`.

See https://codepen.io/etpinard/pen/WpZJgb for a working example. And please refer to our https://community.plot.ly/ for more restyle examples.

kimsk:
Thanks for the comment about the input. 

However, the working example does not have marker style updated when zooming in either?

etpinard:
> the working example does not have marker style updated when zooming in either?

I don't understand. Why should the marker style update on zoom. Restyle is only called on button click if I'm not mistaken.

kimsk:
Sorry for the confusion. I mean if I click the buttons to trigger restyle without zooming in, the marker styles are updated immediately. However, if I zoom in and click the buttons, the marker style are not updated until I double-click the graph to zoom out.

etpinard:
No worries @kimsk !

Thanks very for the report. Something is up during restyle in versions `1.21.3` and up. I think https://github.com/plotly/plotly.js/pull/1257 is the culprit. I'll investigate.

etpinard:
Thanks again for the report @kimsk 

This will be fixed in the next release `1.25.0`.

Issue #648, opened 2016-06-15
Move property literals out of Plotly.restyle


monfera:
See items 1, 2, 3 in https://github.com/plotly/plotly.js/pull/617#discussion_r67130816. 
Preceding comment from @etpinard: https://github.com/plotly/plotly.js/pull/617#discussion_r67041656

etpinard:
I think (at least I like to think) that there's a way to achieve this using our attribute declaration. 

Some questions that we need to answer:

- Do we need all restyle `docalc`, `doplot`, ... [distinctions](https://github.com/plotly/plotly.js/blob/9ee4525189b5377cebc35f691e692f7865115ff2/src/plot_api/plot_api.js#L1638-L1643) ? Can you merge some of them?

- Most `docalc` updates relate to our `data_array` and our `arrayOk` attributes. Meaning that we could simply use the attribute declaration to figure our what update path to take. Are there any exceptions to that rule?

- _anything else ?_

etpinard:
From @rreusser 's https://github.com/plotly/plotly.js/issues/1578 

> Carpet is going to add a huge number of attrs to recalcAttrs. (up to 287 total or so!?) The array of attributes is recalculated on every plot and it looks them up with indexOf. It should place them in a hash (e.g. {mode: true, visible: true, ...}) and should be defined in a plot-api-global context since it never changes.

> @alexcjohnson notes that a larger refactoring is desired, but this might be a start since, at least once carpet is merged, it shouldn't take more than five minutes and may very well be at the point of having a non-negligible impact on performance.

etpinard:
A pretty valuable first step was made in https://github.com/plotly/plotly.js/pull/1653/files/413e5459960c4b4c2c4c6212919e3a5c3183f981#diff-2941ab69a12080c0633ff4ac8ea3aa83

etpinard:
https://github.com/plotly/plotly.js/pull/1999 will probably close this.

etpinard:
Superseded by https://github.com/plotly/plotly.js/issues/2025

PR #1488, opened 2017-03-16
Make restyling arrayOk attribute redo calcdata

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1480

I made a blunder in https://github.com/plotly/plotly.js/pull/1257 where conversion of `arrayOk` attributes from `fullData` -> `calcdata` was moved from the `plot` step to the `calc`. I tested `restyle`, it worked (in some situations) but for the wrong reason. Luckily the fixed wasn't too hard.

I wouldn't mind a @alexcjohnson review.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1576,9 +1576,17 @@ function _restyle(gd, aobj, _traces) {
                 helpers.manageArrayContainers(param, newVal, undoit);
                 flags.docalc = true;
             }
-            // all the other ones, just modify that one attribute
-            else param.set(newVal);
+            else {
+                var moduleAttrs = (contFull._module || {}).attributes || {};
+                var valObject = Lib.nestedProperty(moduleAttrs, ai).get() || {};
 
+                if(valObject.arrayOk && (Array.isArray(newVal) || Array.isArray(oldVal))) {
+                    flags.docalc = true;
+                }
+
+                // all the other ones, just modify that one attribute
+                param.set(newVal);
+            }
         }
 
         // swap the data attributes of the relevant x and y axes?
--- a/test/jasmine/tests/plot_api_test.js
+++ b/test/jasmine/tests/plot_api_test.js
@@ -285,6 +285,36 @@ describe('Test plot api', function() {
             expect(gd.calcdata).toBeDefined();
         });
 
+        it('should do full replot when arrayOk attributes are updated', function() {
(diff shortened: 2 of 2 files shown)


src/plot_api/plot_api.js
```diff
@@ -1576,9 +1576,17 @@ function _restyle(gd, aobj, _traces) {
...
                 flags.docalc = true;
             }
-            // all the other ones, just modify that one attribute
-            else param.set(newVal);
+            else {
+                var moduleAttrs = (contFull._module || {}).attributes || {};
+                var valObject = Lib.nestedProperty(moduleAttrs, ai).get() || {};
 
+                if(valObject.arrayOk && (Array.isArray(newVal) || Array.isArray(oldVal))) {
+                    flags.docalc = true;
```
etpinard: Previously, calls like `restyle(gd, 'marker.size', [[10, 30, 20]])` worked only when one or several axis `autorange` were turned on as this [too](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L1673) clear `gd.calcdata`.

This here now ensures that restyling any `arrayOk` key clears `gd.calcdata`.

test/jasmine/tests/plot_api_test.js
```diff
@@ -285,6 +285,36 @@ describe('Test plot api', function() {
...
 
+        it('should do full replot when arrayOk attributes are updated', function() {
+            var gd = {
+                data: [{x: [1, 2, 3], y: [1, 2, 3]}],
+                layout: {}
+            };
+
+            mockDefaultsAndCalc(gd);
+            Plotly.restyle(gd, 'marker.color', [['red', 'green', 'blue']]);
+            expect(gd.calcdata).toBeUndefined();
```
etpinard: this means restyle clear `gd.calcdata`.

etpinard: Note that `Plotly.plot` is mocked in this suite.

alexcjohnson:
Nice job tracking this down - fix & tests look spot on! 💃

src/plot_api/plot_api.js
etpinard: @alexcjohnson using `_module.attributes` here to find the `arrayOk` keys worked surprisingly well. We could declare a few more keys in the attributes like to resolve https://github.com/plotly/plotly.js/issues/648 at some point.

PR #1493, opened 2017-03-17
Update readme tagline

chriddyp:
cc @etpinard 
- Highlight `scientific charts`
- "_over_ 20 " chart types now
- Financial charts too


etpinard:
:dancer:

PR #1494, opened 2017-03-17
Relayout edge case fixes

etpinard:
The two commits here each fix one relayout edge case broken since https://github.com/plotly/plotly.js/pull/1431 and https://github.com/plotly/plotly.js/pull/1403 respectively.

### DIFF
--- a/src/components/annotations/convert_coords.js
+++ b/src/components/annotations/convert_coords.js
@@ -26,6 +26,8 @@ var toLogRange = require('../../lib/to_log_range');
  *     same relayout call should override this conversion.
  */
 module.exports = function convertCoords(gd, ax, newType, doExtra) {
+    ax = ax || {};
+
     var toLog = (newType === 'log') && (ax.type === 'linear'),
         fromLog = (newType === 'linear') && (ax.type === 'log');
 
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -714,6 +714,7 @@ proto.setConvert = function() {
     for(var i = 0; i < 3; ++i) {
         var ax = this.fullSceneLayout[axisProperties[i]];
         Axes.setConvert(ax, this.fullLayout);
+        ax.setScale = Lib.noop;
     }
 };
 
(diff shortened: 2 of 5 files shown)


src/components/annotations/convert_coords.js
```diff
@@ -26,6 +26,8 @@ var toLogRange = require('../../lib/to_log_range');
  *     same relayout call should override this conversion.
  */
 module.exports = function convertCoords(gd, ax, newType, doExtra) {
+    ax = ax || {};
```
etpinard: .. so that e.g. in [`relayout`](https://github.com/plotly/plotly.js/blob/9f4823be4f50239503bb3c5da126db3980619384/src/plot_api/plot_api.js#L1991-L1992) blank axis objects that pass the forgiving [`AX_NAME_PATTERN`](https://github.com/plotly/plotly.js/blob/9f4823be4f50239503bb3c5da126db3980619384/src/plots/cartesian/constants.js#L30) (which also matches `zaxis?` by the way) don't leave to exceptions.

etpinard: cc @alexcjohnson :arrow_heading_up:

src/plots/gl3d/scene.js
```diff
@@ -714,6 +714,7 @@ proto.setConvert = function() {
     for(var i = 0; i < 3; ++i) {
         var ax = this.fullSceneLayout[axisProperties[i]];
         Axes.setConvert(ax, this.fullLayout);
+        ax.setScale = Lib.noop;
```
etpinard: This statement also appears in the 3d [`axis_defaults`](https://github.com/plotly/plotly.js/blob/9f4823be4f50239503bb3c5da126db3980619384/src/plots/gl3d/layout/axis_defaults.js#L56).

etpinard: We need it here to make margin `relayout` calls work. 

I got carried away in #1431 where I :hocho: ed [it](https://github.com/plotly/plotly.js/pull/1431/files#diff-dd23546f2ee4a55150cd584d026e1c93L18). My mistake.

src/components/annotations/convert_coords.js
alexcjohnson: Man, if only I could turn back time and not use linearized coordinates on log axes...

alexcjohnson:
Looks good. I can't really tell if these problems generalize beyond these fixes or not, so I guess if anything else comes up we'll just have to add more test cases...

💃

PR #1496, opened 2017-03-17
Update README.md

cldougl:
add chart images to readme

@etpinard is that ok there or would you prefer to add the image to the doc section?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,10 @@ plotly.js is a high-level, declarative charting library. plotly.js ships with ov
 chart types, including scientific charts, 3D graphs, statistical charts, SVG maps, financial charts,
 and more.
 
-![](https://raw.githubusercontent.com/cldougl/plot_images/add_r_img/plotly_2017.png)
+<p align="center">
+    <a href="https://www.plot.ly/javascript">
+    <img src="https://raw.githubusercontent.com/cldougl/plot_images/add_r_img/plotly_2017.png">
+</a></p>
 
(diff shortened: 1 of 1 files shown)


etpinard:
Looks good! @cldougl can you wrap the image in an anchor so that click on it lead you to https://plot.ly/javascript/ ?

etpinard:
... or to some other page that might be better.

cldougl:
yep - I think https://plot.ly/javascript/ will be best for now

rreusser:
Also, I think it's pretty safe to swap the markdown image for `` (I'm pretty sure that's the attr) in order to center it.

README.md
```diff
@@ -8,7 +8,10 @@ plotly.js is a high-level, declarative charting library. plotly.js ships with ov
 chart types, including scientific charts, 3D graphs, statistical charts, SVG maps, financial charts,
 and more.
 
-![](https://raw.githubusercontent.com/cldougl/plot_images/add_r_img/plotly_2017.png)
+<p align="center">
+    <a href="https://www.plot.ly/javascript">
```
etpinard: maybe a `target="_blank"` here?

cldougl: oh yes, good call

rreusser:
(FYI: I use [vmd](https://github.com/yoshuawuyts/vmd) all the time for working on + previewing github markdown files locally. Auto-updates on save.)

etpinard:
:dancer:

PR #1497, opened 2017-03-17
Support all axis types for clicktoshow annotations

alexcjohnson:
a quick one @etpinard @charleyferrari


etpinard:
Easy :dancer:

Issue #1495, opened 2017-03-17
Histogram colorscales


ni-hil:
Hello,
I'm trying to make an histogram with a colorscale, I found the following example online, but it seems broken.

https://plot.ly/~etpinard/7146/histogram-with-colorscale/#plot

Is it still the right way to do this?

alexcjohnson:
Oh weird. @ni-hil can you do the binning manually and make this with a bar trace? That behavior is [known to work](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/bar-colorscale-colorbar.png).

@etpinard I wasn't aware that we ever supported this usage but I guess you did if you made that plot! I'm kind of in favor of not allowing it though. Since you need to know the bins already in order to specify colors (and therefore you can't change the bins afterward without also recalculating the color array) seems to me the main advantage of using a histogram instead of a bar trace is lost and outweighed by the pitfalls of potential bin/color mismatches.

etpinard:
This is broken since [`1.21.3`](https://github.com/plotly/plotly.js/releases/tag/v1.21.3). I suspect https://github.com/plotly/plotly.js/pull/1257 and/or https://github.com/plotly/plotly.js/pull/1260.

Here's a working version with `1.21.2`: http://codepen.io/etpinard/pen/MpQpmx

I agree with @alexcjohnson colorscales on histograms are ill-defined. But a regression is a regression and we should fix this.

PR #1500, opened 2017-03-20
Bring back histogram colorscale

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1495

which was broken since https://github.com/plotly/plotly.js/pull/1257 where `arrayToCalcdata` was moved from the `plot` to the `calc` step. 

@alexcjohnson looks ok?


alexcjohnson:
Ok, I might still vote to :hocho: this ability in v2.0 but for now this is good :dancer:

PR #1424, opened 2017-02-27
Implement touch interactions in 3D subplots

etpinard:
Brings in @dfcreative 's https://github.com/mikolalysenko/3d-view-controls/pull/1 (via https://github.com/gl-vis/gl-plot3d/pull/7).

This PR partly resolves https://github.com/plotly/plotly.js/issues/480.

Test it on your mobile devices with http://codepen.io/etpinard/full/aJOMzZ/

PR #1507, opened 2017-03-21
Update gl3d camera

dfcreative:
Ok, so here is updated `gl3d/camera`, related to https://github.com/plotly/plotly.js/pull/1424.
Turns out [gl3d/camera](https://github.com/plotly/plotly.js/blob/master/src/plots/gl3d/camera.js) code is 98% the same as [3d-view-controls](https://npmjs.org/package/3d-view-controls), and in some way there is monkey-patching happening rewriting the original 3d-view-controls with the plotly one.
I guess there is optimization possible, but so far that is workable solution.
@etpinard please look at it

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "gl-surface3d": "^1.3.0",
     "mapbox-gl": "^0.22.0",
     "mouse-change": "^1.4.0",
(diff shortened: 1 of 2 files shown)


etpinard:
Will be merged in https://github.com/plotly/plotly.js/pull/1657

PR #1508, opened 2017-03-21
No unwarranted colorscale or related data on parcoords trace/fullTrace

monfera:
Avoid color scale persistence on trace / fullTrace if the user hasn't specified one (as interpreted by [has_colorscale.js](https://github.com/plotly/plotly.js/blob/master/src/components/colorscale/has_colorscale.js#L36-L42))

Before this PR, `colorscale` related details populated `data` and `fullData`, as seen by these test case changes.

### DIFF
--- a/src/traces/parcoords/calc.js
+++ b/src/traces/parcoords/calc.js
@@ -10,16 +10,9 @@
 
 var hasColorscale = require('../../components/colorscale/has_colorscale');
 var calcColorscale = require('../../components/colorscale/calc');
-var Lib = require('../../lib');
 
 
 module.exports = function calc(gd, trace) {
-    var cs = !!trace.line.colorscale && Lib.isArray(trace.line.color);
-    var color = cs ? trace.line.color : Array.apply(0, Array(trace.dimensions.reduce(function(p, n) {return Math.max(p, n.values.length);}, 0))).map(function() {return 0.5;});
-    var cscale = cs ? trace.line.colorscale : [[0, trace.line.color], [1, trace.line.color]];
-
-    trace.line.color = color;
-    trace.line.colorscale = cscale;
 
     if(hasColorscale(trace, 'line')) {
         calcColorscale(trace, trace.line.color, 'line', 'c');
--- a/src/traces/parcoords/defaults.js
+++ b/src/traces/parcoords/defaults.js
@@ -17,9 +17,9 @@ var maxDimensionCount = require('./constants').maxDimensionCount;
 function handleLineDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
 
     coerce('line.color', defaultColor);
-    coerce('line.colorscale');
 
-    if(hasColorscale(traceIn, 'line')) {
+    if(hasColorscale(traceIn, 'line') && Lib.isArray(traceIn.line.color)) {
+        coerce('line.colorscale');
         colorscaleDefaults(traceIn, traceOut, layout, coerce, {prefix: 'line.', cLetter: 'c'});
     }
     else {
--- a/src/traces/parcoords/parcoords.js
+++ b/src/traces/parcoords/parcoords.js
@@ -121,8 +121,14 @@ function model(layout, d, i) {
         dimensions = trace.dimensions,
         width = layout.width;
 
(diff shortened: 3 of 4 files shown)


src/traces/parcoords/calc.js
```diff
@@ -10,16 +10,9 @@
...
 var hasColorscale = require('../../components/colorscale/has_colorscale');
 var calcColorscale = require('../../components/colorscale/calc');
-var Lib = require('../../lib');
 
 
 module.exports = function calc(gd, trace) {
-    var cs = !!trace.line.colorscale && Lib.isArray(trace.line.color);
-    var color = cs ? trace.line.color : Array.apply(0, Array(trace.dimensions.reduce(function(p, n) {return Math.max(p, n.values.length);}, 0))).map(function() {return 0.5;});
-    var cscale = cs ? trace.line.colorscale : [[0, trace.line.color], [1, trace.line.color]];
-
```
monfera: I just moved this block unchanged from here, where it'd be persisted, to parcoords.js below, as this calculation is now considered implementation detail, unexposed on data/fullData

src/traces/parcoords/calc.js
```diff
@@ -10,16 +10,9 @@
...
-    trace.line.color = color;
-    trace.line.colorscale = cscale;
```
monfera: ... so, no more persisting

src/traces/parcoords/defaults.js
```diff
@@ -17,9 +17,9 @@ var maxDimensionCount = require('./constants').maxDimensionCount;
 function handleLineDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
 
     coerce('line.color', defaultColor);
-    coerce('line.colorscale');
 
-    if(hasColorscale(traceIn, 'line')) {
+    if(hasColorscale(traceIn, 'line') && Lib.isArray(traceIn.line.color)) {
+        coerce('line.colorscale');
         colorscaleDefaults(traceIn, traceOut, layout, coerce, {prefix: 'line.', cLetter: 'c'});
```
monfera: only coerce `colorscale` if it's warranted

src/traces/parcoords/parcoords.js
```diff
@@ -121,8 +121,14 @@ function model(layout, d, i) {
         dimensions = trace.dimensions,
         width = layout.width;
 
+    var cs = !!trace.line.colorscale && Lib.isArray(trace.line.color);
+    var lineColor = cs ?
+        trace.line.color :
+        Array.apply(0, Array(trace.dimensions.reduce(function(p, n) {return Math.max(p, n.values.length);}, 0))).map(function() {return 0.5;});
+    var lineColorScale = cs ? trace.line.colorscale : [[0, trace.line.color], [1, trace.line.color]];
+
```
monfera: okay this is the moved logic block

src/traces/parcoords/defaults.js
monfera: There's no point in attempting to use a color scale if `line.color` is not an array but e.g. a plain color, because the `line.color` numerical array values are mapped to the colors on the color scale.
Also, now `line.colorscale` is only performed if it's warranted.

src/traces/parcoords/parcoords.js
etpinard: Hmm. I would prefer not. Per-datum non-pixel related computations should be in the `calc` step.

src/traces/parcoords/defaults.js
etpinard: I believe patching `defaults.js` is all you need to fix this bug.

src/traces/parcoords/parcoords.js
monfera: In the original code, `calc.js` actually persisted the result of these calculations, see above, these lines:
```
-    trace.line.color = color;
-    trace.line.colorscale = cscale;
```
... so instead of this, I'd just _return_ these two derived things (whereas now, `calc.js` returns `{}`)
Is it how you'd prefer it? I'm swamped with preparation so I'd avoid dead-ends and you may already have a preferred way in mind (or see issues with it).

PR #1519, opened 2017-03-27
Bar with non-numeric (x,y) items bug fix

etpinard:
Currently

```js
var data = [{
  x: [5, NaN, 15, 20, NaN, 21],
  y: [20, NaN, 23, 25, NaN, 26],
  type: 'bar'
}];

```

generates

[image](http://codepen.io/trisphx/pen/gmXQJG?editors=0010)

which is wrong.

With this PR, we get

image

### DIFF
--- a/src/traces/bar/calc.js
+++ b/src/traces/bar/calc.js
@@ -51,15 +51,21 @@ module.exports = function calc(gd, trace) {
     var serieslen = Math.min(pos.length, size.length),
         cd = [];
 
-    // set position
+    // set position and size
     for(i = 0; i < serieslen; i++) {
+        var p = pos[i];
+        var s = size[i];
 
         // add bars with non-numeric sizes to calcdata
         // so that ensure that traces with gaps are
         // plotted in the correct order
 
-        if(isNumeric(pos[i])) {
-            cd.push({p: pos[i]});
+        if(isNumeric(p)) {
+            if(isNumeric(s)) {
+                cd.push({ p: p, s: s });
+            } else {
+                cd.push({ p: p });
+            }
         }
     }
 
@@ -81,13 +87,6 @@ module.exports = function calc(gd, trace) {
         b = (isNumeric(b)) ? b : 0;
         for(i = 0; i < cd.length; i++) {
             cd[i].b = b;
-        }
-    }
-
-    // set size
-    for(i = 0; i < cd.length; i++) {
-        if(isNumeric(size[i])) {
-            cd[i].s = size[i];
         }
     }
 
--- a/src/traces/bar/calc.js
+++ b/src/traces/bar/calc.js
@@ -49,24 +49,11 @@ module.exports = function calc(gd, trace) {
 
     // create the "calculated data" to plot
     var serieslen = Math.min(pos.length, size.length),
-        cd = [];
+        cd = new Array(serieslen);
 
     // set position and size
     for(i = 0; i < serieslen; i++) {
-        var p = pos[i];
-        var s = size[i];
-
-        // add bars with non-numeric sizes to calcdata
-        // so that ensure that traces with gaps are
-        // plotted in the correct order
-
-        if(isNumeric(p)) {
-            if(isNumeric(s)) {
-                cd.push({ p: p, s: s });
-            } else {
-                cd.push({ p: p });
-            }
-        }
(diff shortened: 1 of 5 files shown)


src/traces/bar/calc.js
```diff
@@ -51,15 +51,21 @@ module.exports = function calc(gd, trace) {
...
         // plotted in the correct order
 
-        if(isNumeric(pos[i])) {
-            cd.push({p: pos[i]});
+        if(isNumeric(p)) {
+            if(isNumeric(s)) {
+                cd.push({ p: p, s: s });
+            } else {
+                cd.push({ p: p });
+            }
```
alexcjohnson: Nice. Do we need to merge in the [base loop below](https://github.com/plotly/plotly.js/pull/1519/files#diff-54f85e79ac98b8a983427bbd8a31a3b9R76) too?

etpinard: Maybe? The behavior now is similar to how scatter fills in its `arrayOk` attributes. I can double check though.

alexcjohnson: Looks like scatter keeps a 1:1 mapping between calcdata points and the input arrays, filling in non-numerics with [`{x: false, y: false}`](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/calc.js#L117) - that's a good point about other `arrayOK` attributes, maybe it would be easier to keep that 1:1 mapping for bars too? Should be faster too, as we're not doing repeated `.push`, can set the size of `cd` at the start.

etpinard: I prefer removing non-numeric items in the calc step, so that we don't have to do that during the plot step (e.g. on zoom).

alexcjohnson: OK, but then how do you want to solve this for bar color arrays etc? save the original index in each `cd` element? `cd.push({p: p, s: s, i: i})` or something? That would need to be propagated all the way through to `Lib.mergeArray` which seems like a bit of a disaster.

Non-numerics should be regarded as an edge case, so the overhead of managing their objects shouldn't be seen as a concern, but you're right to worry about the overhead of checking each *valid* point. That said, this works for scatter which can have way more points than bars, so I think as long as we make this test super simple it shouldn't be a problem.

etpinard: @alexcjohnson ok, you convinced me. 

With commit https://github.com/plotly/plotly.js/pull/1519/commits/0144f06bbcb46462ca7bd1187481d38d78583e83 bar calcdata has now a 1-1 correspondence with the bar `(x,y)` coordinates. Note that non-numeric bar items are cleaned up in [`bar/plot.js`](https://github.com/plotly/plotly.js/blob/a41b67670db6c35edf8ec901e0692e0eaf3c3cd6/src/traces/bar/plot.js#L82-L87) and `makeCalcdata` (called above the `serieslen` loop in `bar/calc.js`) sets non-numeric values to `BADNUM`.

---

This a pretty big change that might lead to regressions, so this PR will make be part of the next minor release `v1.26.0` only.

test/jasmine/tests/bar_test.js
```diff
@@ -283,6 +283,28 @@ describe('Bar.calc', function() {
...
+    it('should not exclude items with non-numeric y from calcdata (to plots gaps correctly)', function() {
+        var gd = mockBarPlot([{
+            x: ['a', 'b', 'c', 'd'],
+            y: [1, null, 'nonsense', 15]
+        }]);
+
+        var cd = gd.calcdata;
+        assertPointField(cd, 'x', [[0, 1, 2, 3]]);
+        assertPointField(cd, 'y', [[1, NaN, NaN, 15]]);
+    });
```
alexcjohnson: do we have a test with a bad `x` but good `y` for the same point?

etpinard: Good call. Added in https://github.com/plotly/plotly.js/pull/1519/commits/6f0a95d3ecea7f7ae9ed906b88cee0a96284402d

alexcjohnson:
Nice - big improvement. At some point we may want to make sure we have tests that throw some garbage or partially garbage points into ALL our trace types - perhaps just add some junk to various image test mocks that shouldn't impact the images - but it's great that we've locked this down for bars now. 💃

Issue #1200, opened 2016-11-25
Smarter defaults for axis anchors


alexcjohnson:
We could be smarter about axis anchors for subplots, basing their default values off the traces we have.

In a new mock I made for [date axes](https://github.com/plotly/plotly.js/pull/1194/commits/87786480858e8f54532ff2a9330697ff8a3bf076), there are 7 stacked subplots, each with one trace and its own x and y axes. All the traces have:
`xaxis: 'x<n>', yaxis: 'y<n>'`
and all the x axes have layout obects:
`xaxis<n>: {anchor: 'y<n>'}`

With smarter defaults, I would not have needed to include these x axis objects at all since `'x<n>'` is only ever used with `'y<n>'` in the data. In fact, in principle my y axes should also be anchored to the corresponding x axes... I didn't do that, and it didn't matter here because all the x axes have domain `[0, 1]`, but if I **were** to change some of these domains, then I would be in for some confusion as the y axes wouldn't move as I would expect.

This would not be backward compatible, so I guess should be included in v2.0 if we do it, but I can't really think of a situation where people would be depending on the existing defaults.

alexcjohnson:
Just encountered this again as a really peculiar failure mode for gl2d plots: gl2d_stacked_subplots does not have explicit anchors for `yaxis2` and `yaxis3`, which is fine for positioning them (all the x axes have the same left edges) but the plot behaves badly:

- Open http://localhost:3000/devtools/test_dashboard/#gl2d_stacked_subplots - looks correct initially.
![screen shot 2017-03-31 at 11 52 51 pm](https://cloud.githubusercontent.com/assets/2678795/24575207/31317534-166d-11e7-8117-7938f6e6051c.png)
- Click "autoscale" (or zoom in/out) in the modebar, and all but one subplot gets emptied out:
![screen shot 2017-03-31 at 11 43 31 pm](https://cloud.githubusercontent.com/assets/2678795/24575151/fb844746-166b-11e7-84e4-198fc85ae333.png)
- Similar things happen to this plot on `Plotly.redraw`, `Plotly.restyle`, and `Plotly.relayout` calls.
- Behavior after this is weird too: trying to zoom in the top subplot (with a zoombox) causes the top y axis and the BOTTOM x axis to zoom.
- The problem is we have `xy2` and `xy3` subplots as well as `x2y2` and `x3y3` - for some reason the first time around, these are below the real ones, but on redraw they pop to the top.
[...]

PR #1522, opened 2017-03-28
Axis constraints

### LINKED ISSUE #272: Aspect ratio key for 2d plots
Similar to:
https://plot.ly/python/reference/#layout-scene-aspectratio
For 3d plots

alexcjohnson:
Closes #272 - defines and enforces constraints between axis scales, so you can have a fixed aspect ratio plot. Allows arbitrary chaining of many axes together, each with its own aspect ratio, along with unconstrained subplots.

TODO:
- [x] decide what to do if one of the axes with a constraint is also labeled `fixedrange` - should we forbid this combination? Or make the whole constraint group fixed? Or something else? I can't think of a use case for this so I'm inclined to forbid it, at least for now.
- [x] tests for the dynamic behavior
- [x] gl2d compatibility? This might be as simple as constraining the zoom box, given the limited zoom functionality available in gl2d right now, but there is some weird behavior right now that may also need to be addressed (open http://localhost:3000/devtools/test_dashboard/#gl2d_stacked_subplots and click "autoscale", for example)
- [x] enforce constraints in `relayout` calls that don't come from the GUI. This may actually allow me to simplify some things in dragbox.js too...
- [x] allow x-x and y-y constraints, not just x-y

@etpinard @cpsievert @rreusser I'd be grateful if one or two of you could play with the interactions - zoom box, dragging an axis end, dragging a corner of a plot - and tell me if they feel right. The mock in the PR - http://localhost:3000/devtools/test_dashboard/#axes_scalewith if you run the test dashboard - should be a good starting point, as it has two linkages across two coupled subplots.


src/plots/cartesian/dragbox.js
```diff
@@ -578,7 +580,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
         redrawObjs(fullLayout.annotations || [], Registry.getComponentMethod('annotations', 'drawOne'));
         redrawObjs(fullLayout.shapes || [], Registry.getComponentMethod('shapes', 'drawOne'));
-        redrawObjs(fullLayout.images || [], Registry.getComponentMethod('images', 'draw'));
+        redrawObjs(fullLayout.images || [], Registry.getComponentMethod('images', 'draw'), true);
```
etpinard: Nice touch :racehorse:

src/plots/cartesian/dragbox.js
```diff
@@ -578,31 +705,44 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
                 .call(Drawing.setTranslate, plotDx, plotDy)
-                .call(Drawing.setScale, xScaleFactor, yScaleFactor)
+                .call(Drawing.setScale, 1 / xScaleFactor, 1 / yScaleFactor)
 
                 // This is specifically directed at scatter traces, applying an inverse
                 // scale to individual points to counteract the scale of the trace
                 // as a whole:
                 .select('.scatterlayer').selectAll('.points').selectAll('.point')
-                    .call(Drawing.setPointGroupScale, 1 / xScaleFactor, 1 / yScaleFactor);
+                    .call(Drawing.setPointGroupScale, xScaleFactor, yScaleFactor);
```
etpinard: Is this a bug currently?

alexcjohnson: no, I just flipped the definitions of x/yScaleFactor since I noticed they were being inverted in most of their uses.

src/plots/cartesian/dragbox.js
```diff
@@ -473,43 +567,67 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
+
         var ax, i, rangeInitial;
 
-        if(doubleClickConfig === 'autosize') {
+        // For reset+autosize mode:
+        // If *any* of the main axes is not at its initial range
+        // (or autoranged, if we have no initial range, to match the logic in
+        // doubleClickConfig === 'reset' below), we reset.
+        // If they are *all* at their initial ranges, then we autosize.
+        if(doubleClickConfig === 'reset+autosize') {
```
alexcjohnson: This reorg was to fix a subtle bug:
- set explicit initial ranges for both x and y,
- autoscale one of them (doubleclick the middle of the x axis, for example)
- doubleclick the plot repeatedly: it will toggle between autoscaled x, initial y, and autoscaled y, initial x (with this change it toggles between fully autoscaled and fully initial-scaled).

src/plots/cartesian/dragbox.js
```diff
@@ -135,6 +159,8 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 clearSelect(zoomlayer);
             }
             else if(isSelectOrLasso(dragModeNow)) {
+                dragOptions.xaxes = xa;
+                dragOptions.yaxes = ya;
```
alexcjohnson: Fixing what I broke in https://github.com/plotly/plotly.js/pull/1522/commits/98900a27b411e34b2447217abb9ba6ea87baaa48#diff-1bd3dd52cddb8638876bbe17cef28a0dL105 when I consolidated to use `recomputeAxisLists` for the initial settings too (when `dragOptions` isn't yet available, and I didn't see anyone using the `xaxes/yaxes` properties. But it turns out it was subtly broken before anyway, as the version in `recomputeAxisLists` used the wrong names (`dragoptions.xa` instead of `dragoptions.xaxes`).

`dragbox.js` is still a bit of a mess, but... I think it's a little better now.

cpsievert:
@alexcjohnson the zooming on your example looks great, and even does a sensible thing in extreme cases 🍻 

zoom

I am, however, thrown out a bit by the choice of expanding the relevant axis range to match `scaleratio`. I'm more familiar with approaches that shrink/restrict `axisID.domain`'s domain accordingly.

BTW, here's a similar example where it seems the x-axis doesn't expand (i.e., it's wrong)...

screen shot 2017-03-28 at 11 26 23 am

alexcjohnson:
@cpsievert 
> I am, however, thrown out a bit by the choice of expanding the relevant axis range to match `scaleratio`. I'm more familiar with approaches that shrink/restrict `axisID.domain`'s domain accordingly.

Hmm, yeah, I see that in your ggplot example https://github.com/plotly/plotly.js/issues/272#issuecomment-281405318 - it's obviously not the way I think about it, but I'm not sure if I've seen precedent for this way or not - other than I guess the way we handle maps. The only way I can really see shrinking the actual axis size making sense within our structure is some sort of algorithm to automatically reduce the total plot width to minimize white space, but that seems a bit tricky to get right. I'd also say this approach strikes me as rooted in a static presentation of the plot - those long narrow subplots are not so useful once you start zooming and panning. But of course we want to be able to replicate all the plots folks feed in!

Do you think that for ggplotly's purposes you could pull out the overall plot dimensions from ggplot's results and just pass them along to plotly.js? In which case I guess it would basically be ggplot enforcing the initial constraints (with plotly.js just verifying and possibly correcting for small differences in layout) and plotly.js's role in this would primarily be in dynamic behavior.

And re: the part that disappeared from your comment above - do you think the with/width confusion merits changing the name `scalewith` to something else?

cpsievert:
> Do you think that for ggplotly's purposes you could pull out the overall plot dimensions from ggplot's results and just pass them along to plotly.js?

That's essentially what I'm doing in https://github.com/ropensci/plotly/pull/509. As you mention, it unfortunately requires [estimating the axis dimensions](https://github.com/ropensci/plotly/pull/509/files#diff-3afd3a8e6a2cbc84a7afc6d2d06ec5e3R428), which is approximate, at best. It also significantly complicates the logic on my end, but I suppose it's possible. 

> And re: the part that disappeared from your comment above - do you think the with/width confusion merits changing the name scalewith to something else?

I'm not highly opposed to the name, but I see your point. How about `scaleanchor`?

FWIW, at least for the R package, I'm going to add [more robust name checking](https://github.com/ropensci/plotly/issues/911) that should catch instances like this.

src/plots/cartesian/dragbox.js
etpinard: Great. Let's make sure to :lock: this down before merging.

etpinard:
> [from @cpsievert] How about scaleanchor?

:+1: much better than `scalewith` in my mind.

etpinard:
> decide what to do if one of the axes with a constraint is also labeled fixedrange - should we forbid this combination? Or make the whole constraint group fixed? Or something else? I can't think of a use case for this so I'm inclined to forbid it, at least for now.

I'd vote for skipping `coerce('scale*')` when `fixedrange` is turned on. The `scale*` and `fixedrange` sound to me like two conflicting things.

> gl2d compatibility? This might be as simple as constraining the zoom box, given the limited zoom functionality available in gl2d right now, but there is some weird behavior right now that may also need to be addressed (open http://localhost:3000/devtools/test_dashboard/#gl2d_stacked_subplots and click "autoscale", for example)

That would require a few patches in [`gl-select-box`](https://github.com/gl-vis/gl-select-box) which could use some :heart: at the moment. Maybe it's worth investigating using `cartesian/dragbox.js` for gl2d zoom-box interactions instead? I don't see a performance benefit from using a gl solution for zoom-box. Though, hovering, panning and scrolling interactions would have to remain in the gl universe for speed :racehorse:

alexcjohnson:
> I'd vote for skipping `coerce('scale*')` when `fixedrange` is turned on. The `scale*` and `fixedrange` sound to me like two conflicting things.

👍 

> That would require a few patches in [`gl-select-box`](https://github.com/gl-vis/gl-select-box) which could use some :heart: at the moment. Maybe it's worth investigating using `cartesian/dragbox.js` for gl2d zoom-box interactions instead?

I'll take a look at it, that might not be too hard. I ran into a few more issues gl2d zoom on multi-subplot plots - the one I mentioned above seems to be dropped contexts, the relayout makes new ones for some reason; but there's another one that the grey overlay covers the whole plot (margins included) EXCEPT whatever other gl subplots happen to be in front of the plot you're zooming on. That probably wouldn't be hard to fix but if we just swap in dragbox.js it would be moot.

cpsievert:
> Do you think that for ggplotly's purposes you could pull out the overall plot dimensions from ggplot's results and just pass them along to plotly.js?

After thinking about this again, I wouldn't be able to support aspect ratios in `ggplotly()` *on resize* (in such a way that mimics ggplot2's spacing). That's because axis domains (computed in R) depend on the container height/width (which may change after we lose connection to an R session). So, if it isn't a lot of trouble, it would be a win for `ggplotly()` if we could implement the other way (shrinking the domain of the appropriate axis). That being said, I see your point that ggplot2's approach is "rooted in a static presentation of the plot" in that current implementation is more convenient for zooming.

I guess what I'm trying to say is I would still slightly prefer axis shrinking, but I can make do with the current implementation (and understand why you prefer it). We'll just have to educate `ggplotly()` users why the spacing is different. I've tried to stick to the philosophy of "do our damnest to replicate the static result by default", but I suppose we can deviate from that somewhat ;)

alexcjohnson:
After a discussion with @cpsievert I think we'll leave the general structure as it is and let ggplotly handle explicitly setting the subplot shapes to match what ggplot generates.

But in the discussion it came up that there are cases you want to constrain several x (or y) axis scales without a corresponding x<->y constraint. See eg https://github.com/ropensci/plotly/issues/908 - so I'll see if I can add that in here too, letting `scaleanchor` reference ANY other axis, regardless of its orientation. I think there's a good chance that it'll "just work" ™️ but we'll see.

src/plots/cartesian/dragbox.js
alexcjohnson: 🔒 : https://github.com/plotly/plotly.js/pull/1522/commits/86e0d5a400766f7b2dedec98275cf619422ad869#diff-200dbdf53958cf8973413a709518f4bbR313

test/jasmine/tests/click_test.js
```diff
@@ -771,7 +771,7 @@ describe('Test click interactions:', function() {
             var translate = Drawing.getTranslate(mockEl),
                 scale = Drawing.getScale(mockEl);
 
-            expect([translate.x, translate.y]).toBeCloseToArray([61.070, 97.712]);
+            expect([translate.x, translate.y]).toBeCloseToArray([-25.941, 43.911]);
```
alexcjohnson: This test was actually incorrect, as wheel events weren't passing along the mouse coordinates before https://github.com/plotly/plotly.js/pull/1522/commits/86e0d5a400766f7b2dedec98275cf619422ad869#diff-d48932d4d7f81cc4201054d397d53f1aR17 and these matter to wheel zoom.
(the change from -1000 to -20 above is irrelevant, this event handler clips at +/-20 per event fired)

src/plot_api/plot_api.js
```diff
@@ -2079,6 +2079,18 @@ function _relayout(gd, aobj) {
...
+                    pp1 === 'rangemode' ||
+                    pp1 === 'type' ||
+                    pp1 === 'domain' ||
+                    pp1 === 'fixedrange' ||
+                    pp1 === 'scaleanchor' ||
+                    pp1 === 'scaleratio' ||
+                    ai.indexOf('calendar') !== -1 ||
+                    ai.match(/^(bar|box|font)/)) {
+                flags.docalc = true;
+            }
```
alexcjohnson: had to move this block above the following `_has('gl2d')` block, so `scaleanchor/scaleratio` get the necessary recalc. Seems like as a general rule, we might want to reorder these from biggest change to smallest, ie all the recalcs, then all the replots, then the ticks & styles... if we don't find an altogether better way to manage these flags (via the schema?).

src/plots/gl2d/camera.js
```diff
@@ -76,32 +99,114 @@ function createCamera(scene) {
...
+                        dy = result.boxEnd[1] - result.boxStart[1];
+                        var dydx = (dataBox[3] - dataBox[1]) / (dataBox[2] - dataBox[0]);
+
+                        if(Math.abs(dx * dydx) > Math.abs(dy)) {
+                            result.boxEnd[1] = result.boxStart[1] +
+                                Math.abs(dx) * dydx * (Math.sign(dy) || 1);
+
+                            // gl-select-box clips to the plot area bounds,
+                            // which breaks the axis constraint, so don't allow
+                            // this box to go out of bounds
```
alexcjohnson: in contrast, svg lets you make these constrained boxes extend outside the plot area. Didn't seem worth diving into `gl-select-box` to make this marginal feature match up precisely.

src/plots/gl2d/camera.js
```diff
@@ -76,32 +99,114 @@ function createCamera(scene) {
...
+                                result.boxEnd[0] = dataBox[2];
+                                result.boxEnd[1] = result.boxStart[1] +
+                                    (dataBox[2] - result.boxStart[0]) * Math.abs(dydx);
+                            }
+                        }
+                    }
+                    // otherwise clamp small changes to the origin so we get 1D zoom
+                    else {
+                        if(smallDx) result.boxEnd[0] = result.boxStart[0];
+                        if(smallDy) result.boxEnd[1] = result.boxStart[1];
```
alexcjohnson: @etpinard while I was at it I added in a few of the bits of behavior used in svg, such as 1D zoom, minimum zoom size, and clamping panning to pure x (y) if dy (dx) is small enough. Do you want to take it for a spin and see what you think?

src/plots/gl2d/scene2d.js
```diff
@@ -426,6 +427,13 @@ proto.plot = function(fullData, calcData, fullLayout) {
         ax.setScale();
     }
 
+    var mockLayout = {
+        _axisConstraintGroups: this.graphDiv._fullLayout._axisConstraintGroups,
+        xaxis: this.xaxis,
+        yaxis: this.yaxis
+    };
+    enforceAxisConstraints({_fullLayout: mockLayout});
```
alexcjohnson: kind of hacky, but `enforceAxisConstraints ` has to happen after the `Axes.doAutoRange` just above, and these aren't the `gd._fullLayout` axes...

A bunch of things here would break if/when we support overlaid axes...

src/plots/gl2d/scene2d.js
```diff
@@ -629,12 +647,15 @@ proto.draw = function() {
...
-        else if(!result && this.lastPickResult) {
-            this.spikes.update({});
-            this.lastPickResult = null;
-            this.graphDiv.emit('plotly_unhover');
-            Fx.loneUnhover(this.svgContainer);
-        }
+    }
+
+    // Remove hover effects if we're not over a point OR
+    // if we're zooming or panning (in which case result is not set)
```
alexcjohnson: also to match svg

test/jasmine/tests/gl2d_click_test.js
```diff
@@ -246,12 +246,20 @@ describe('Test hover and click interactions', function() {
             pointNumber: 0
         });
 
+        // after the restyle, autorange changes the y range
```
alexcjohnson: For some reason I don't understand, when this test ran before my changes, the y-axis didn't get autoranged after the `restyle`, even though when I make the same `restyle` call on master in the test dashboard it does autorange. Anyway, now it *does* get autoranged during the test, as it should, which is why I needed to alter the `run` the second time around.

test/jasmine/tests/gl2d_click_test.js
```diff
@@ -269,12 +277,23 @@ describe('Test hover and click interactions', function() {
             pointNumber: 0
         });
 
+        // after the restyle, autorange changes the x AND y ranges
+        // I don't get why the x range changes, nor why the y changes in
+        // a different way than in the previous test, but they do look
+        // correct on the screen during the test.
```
alexcjohnson: I think it makes sense that the axis ranges are different here than in the test above, as this is scattergl-fancy... but then why are they the same as the test above during the first `run`? This test suite baffles me...

alexcjohnson:
@etpinard ready for another review. Single gl2d plots work well with scale constraints now (please try it out and let me know what you think). I did not investigate multiple gl2d subplots (coupled or otherwise) as I was thrown off for a while by https://github.com/plotly/plotly.js/issues/1200#issuecomment-290893266 but now that I understand that I could look into this, if we think it's important.

test/jasmine/assets/mouse_event.js
```diff
@@ -14,7 +14,7 @@ module.exports = function(type, x, y, opts) {
         ev;
 
     if(type === 'scroll') {
-        ev = new window.WheelEvent('wheel', opts);
+        ev = new window.WheelEvent('wheel', Object.assign({}, fullOpts, opts));
```
etpinard: can you use `Lib.extendFlat` here instead - in case we run these tests on old browsers some day?

test/jasmine/tests/click_test.js
etpinard: Thanks :+1:

test/jasmine/assets/mouse_event.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1522/commits/745b9535a06c1c850013a99504c53912f5b30700

---

I should really be reviewing all commits before making comments.

test/jasmine/tests/axes_test.js
```diff
@@ -570,6 +572,49 @@ describe('Test axes', function() {
...
+        var gd;
+
+        beforeEach(function() {
+            gd = createGraphDiv();
+            jasmine.addMatchers(customMatchers);
+        });
+
+        afterEach(destroyGraphDiv);
+
+        it('updates ranges when adding, removing, or changing a constraint', function(done) {
```
etpinard: Nice test :tada:

src/plot_api/plot_api.js
etpinard: I really like the idea of adding a `recalc` true flag to the attributes that require a recalc on restyle / relayout. But that can wait. We should do this in one PR crossing off https://github.com/plotly/plotly.js/issues/648.

src/plots/gl2d/camera.js
etpinard: Loving it. Thanks!

src/plots/gl2d/camera.js
etpinard: Agreed. No need to do this for now :+1:

src/plots/gl2d/scene2d.js
etpinard: Very nice. Thanks :+1:

etpinard:
:dancer: if you don't feel like including a patch for https://github.com/plotly/plotly.js/issues/1200 in this PR. 

This gets my vote for _PR of the year_ :trophy:

alexcjohnson:
Thanks @etpinard 😊  I'll put a note about this into #1200 but I think I should move on for now.

martintburgess:
I am trying to build a plot that is: 
- responsive to changes in window size (using `config={'responsive': true}`)
- NOT zoomable (both axes have `fixedrange: true`
- has a fixed aspect ratio/uniform axis units. 
Am I correct in thinking this is not possible because `scaleanchor` does not work with `fixedrange` with this PR?

PR #1505, opened 2017-03-21
Add mouse event to `plotly_click`, `plotly_hover` and `plotly_unhover`

n-riesco:
With this PR, listeners to `plotly_click`, `plotly_hover` and `plotly_unhover` receive the original mouse event (that triggered the event) in `data.event`.

Implements #988 for scatter (@etpinard **edit:** cartesian), pie, ternary, geo and mapbox plots.

Fixes #1456 and #902

PR #1523, opened 2017-03-28
Add 'hovertext' attribute in scatter* traces

etpinard:
Scatter traces (i.e. `scatter`, `scatter3d`, `scattergeo`, `scatterternary` and `scattermapbox`) have a `text` attribute which is displayed on the graph when `mode` contains the `'text'` flag. The `text` content is also displayed in the hover label when `hoverinfo` contains the `'text'` flag.

But, what if a user wants to display some text elements on the chart and some other (distinct) text elements in the hover labels? Currently we would have to use two traces to achieve this.

So, I proposed adding a `hoverinfo` attribute. Now it's a little difficult to preserve backward compatible; we should make sure to agree on the details.


src/traces/scatter/hover.js
```diff
@@ -72,7 +72,9 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
                 yLabelVal: di.y
             });
 
-            if(di.tx) pointData.text = di.tx;
+            if(di.htx) pointData.text = di.htx;
+            else if(trace.hovertext) pointData.text = trace.hovertext;
+            else if(di.tx) pointData.text = di.tx;
```
etpinard: This is what I came up with:

- If trace `hoverinfo` contains a `'text'` flag and `hovertext` is not set the `text` elements will be seen in the hover labels
- If trace `hoverinfo` contains a `'text'` flag and `hovertext` is set, `hovertext` takes precedence over `text` i.e. the `hoverinfo` element will be seen in the hover labels

etpinard:
Would love some :+1: s from @alexcjohnson @bpostlethwaite @n-riesco

alexcjohnson:
👍  seems clear and I think it covers all use cases. I suppose you could argue for the ability to display BOTH text and hovertext in the hover labels, but that opens a big can of worms just to avoid a bit of duplication in hovertext so doesn't seem like a good idea.

Bar and pie could also benefit from this feature.

etpinard:
> Bar and pie could also benefit from this feature.

Good call. Thanks!

etpinard:
Thanks @n-riesco @alexcjohnson , dropping the `discussion needed` label.

etpinard:
TODO:

- add `'hovertext'` for pie, after https://github.com/plotly/plotly.js/pull/1505 is merged
- add `'hovertext'` for scattergeo, after https://github.com/plotly/plotly.js/issues/1529 is fixed

src/traces/pie/attributes.js
```diff
@@ -79,7 +79,27 @@ module.exports = {
 
     text: {
         valType: 'data_array',
```
etpinard: @alexcjohnson I just noticed that `text` for pie is declared as a `data_array` whereas other trace types have it as a `string` with `arrayOk: true`. Was that on purpose or should we allow single-value `text` input for pie traces?

alexcjohnson: That was on purpose - though probably never explicitly discussed. My rationale is that `text` is essentially the `x` data for pies, and they would be meaningless without this being an array of data. Can you think of a counterexample?

test/jasmine/tests/hover_pie_test.js
```diff
@@ -186,44 +185,60 @@ describe('pie hovering', function() {
...
+        function _hover() {
+            mouseEvent('mouseover', 223, 143);
+        }
 
-                mouseEvent('mouseover', 223, 143);
+        function assertLabel(expected) {
+            var labels = d3.selectAll('.hovertext .nums .line');
 
-                var labels = Plotly.d3.selectAll('.hovertext .nums .line');
+            expect(labels[0].length).toBe(expected.length);
```
alexcjohnson: 🐄  `labels.size()`

etpinard: Good :eye: thanks!

src/traces/pie/attributes.js
etpinard: I agree for the on-graph `textinfo: 'text'` part. But if someone wants to show the same `text` piece in all hover labels? But I guess after this PR, `hovertext` will cover that case.

alexcjohnson: Oh sorry, that's `label`... I still don't see a good usage for a single-value `text` but perhaps it would be better to define it consistently with other trace types anyway.

test/jasmine/tests/hover_pie_test.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1523/commits/1bf54e29a966dd8a203c9bd0f9dd022f3928183e

alexcjohnson:
nice. 💃

Issue #1317, opened 2017-01-18
[Bug] Manual width breaks hover boxes


lewisjb:
Example: http://codepen.io/etpinard/pen/VPPvpM

Two cases:
1 (Top example). Manual width > automatic: Hover boxes don't show up
2 (Bottom example). Manual width < automatic: Hover boxes not aligned properly

Related: #1308

alexcjohnson:
Nice find. In the top example the hover box *does* show up... if you hover over exactly the middle of the bar. The issue is that we're not taking the manual width into account in determining the region you can hover over to see the label. We'll have to be careful with how this plays with grouped bars (which can mean that the hoverable region is quite different from the bar itself) and note that the problem is only in "compare data" mode - in "closest data" mode it works as expected.

hy9be:
Run into the same problem. Is anyone working on this issue already? Thanks!

etpinard:
> Is anyone working on this issue already? Thanks

@hy9be no one that I'm aware of. Want to give a shot?

hy9be:
@etpinard Why no. I will see what I could figure out. Any suggestions for directions or where should I start? Looks tricky by a quick glance.

etpinard:
The problem is in https://github.com/plotly/plotly.js/blob/master/src/traces/bar/hover.js where some fields computed in https://github.com/plotly/plotly.js/blob/master/src/traces/bar/set_positions.js are not used properly in the case of custom `width`.

lewisjb:
I had a look into it, it's due to hover using `bargroupwidth`, which seems to always be `0.8`.
But, due to it being a fairly new addition, I wasn't entirely sure about what it's purpose is. I created a test case under the assumption that `bargroupwidth` should be the same as `barwidth` when using a custom `width`; but when I made the changes in set_positions.js, a few other tests failed as they expect `bargroupwidth` to be `0.8`. I haven't looked further into it yet

hy9be:
I took a look at the "closest" hovermode only. The problem is t.barwidth at https://github.com/plotly/plotly.js/blob/master/src/traces/bar/hover.js#L24, is an array of the widths of all bars when manual width is set. Thus barDelta results in a NaN. My solution is that, find the width of the closest bar to the xval and yval hovering point.

But my code
`function closestIndex(list, x) {
        var min,
            chosen = 0;
        for (var i in list) {
            min = Math.abs(list[chosen] - x);
            if (Math.abs(list[i] - x) < min) {
[...]

PR #1527, opened 2017-03-29
Fix hover label positioning for bar trace with set `width`

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1317 - at least the `hovermode: 'closest'` part, see https://github.com/plotly/plotly.js/issues/1317#issuecomment-287477107 for more details.

TODO:

- [x] add more tests!


src/traces/bar/hover.js
```diff
@@ -19,35 +19,38 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
...
+
     var dx, dy;
     if(trace.orientation === 'h') {
         dx = function(di) {
             // add a gradient so hovering near the end of a
             // bar makes it a little closer match
             return Fx.inbox(di.b - xval, di.x - xval) + (di.x - xval) / (di.x - di.b);
         };
-        dy = function(di) {
+        dy = function(di, i) {
```
alexcjohnson: I suppose this (providing `i` as an extra arg) is the most flexible way to handle the issue for future extensibility, but the way that matches better with what we do in scatter - and should be slightly better perf - would be to include the width of each point in `di`

etpinard: Yeah. Good call! Post https://github.com/plotly/plotly.js/pull/1519 this will be easy!

src/traces/bar/set_positions.js
```diff
@@ -392,9 +392,11 @@ function setBarCenter(gd, pa, sieve) {
         for(var j = 0; j < calcTrace.length; j++) {
             var calcBar = calcTrace[j];
 
+            // store the actual bar width and position, for use by hover
+            var width = calcBar.w = (barwidthIsArray) ? barwidth[j] : barwidth;
```
alexcjohnson: I suspect we could clean up the hover routine quite a bit more if we stopped using `.x` and `.y` in `calcdata`... then we'd only have to switch at the beginning of the routine "are `xval` and `yval` position and size or vice versa" but I wasn't sure what the consequences of that would be beyond hover, just adding one more value was safe and easy.

src/traces/bar/hover.js
```diff
@@ -15,50 +15,66 @@ var Color = require('../../components/color');
...
 
+    minPos = (hovermode === 'closest') ?
+        thisBarMinPos :
+        function(di) {
+            /*
+             * In compare mode, accept a bar if you're on it *or* its group.
+             * Nearly always it's the group that matters, but in case the bar
+             * was explicitly set wider than its group we'd better accept the
+             * whole bar.
+             */
```
alexcjohnson: fixes the "compare" part of #1317 - at least if the bar is wider than our default. I guess if it's *narrower* and there are other things on the plot, we could be accepting this point when we shouldn't be... maybe I'll look at that case tomorrow and see if I can trigger it.

src/traces/bar/plot.js
```diff
@@ -61,7 +59,7 @@ module.exports = function plot(gd, plotinfo, cdbar) {
                     // log values go off-screen by plotwidth
                     // so you see them continue if you drag the plot
                     var p0 = di.p + ((poffsetIsArray) ? poffset[i] : poffset),
-                        p1 = p0 + ((barwidthIsArray) ? barwidth[i] : barwidth),
+                        p1 = p0 + di.w,
```
alexcjohnson: a little follow-on benefit of adding `di.w`

src/traces/bar/hover.js
alexcjohnson: Alright, the only case I can see where this pops up is an explicitly narrow (`width<1`) single bar with other things on the plot as well, like:
```js
var data = [
	{type: 'bar', x: [1], y: [2], width: 0.1}, {x:[1.1], y:[1]}
];
var layout = {
	xaxis: {
		range: [-2, 2]
	}
};

Plotly.newPlot(gd, data, layout);
```
then the hoverable region for the bar is so wide it actually extends past the scatter point. But the scatter point still wins when you'd want it to. So if that's all it is, I don't feel like it's worth addressing this very obscure edge case, it'd be just as likely to cause other problems.

alexcjohnson:
@etpinard back to you for review. Do the tests I added satisfy the TODO up top?

src/traces/bar/hover.js
etpinard: I agree with :arrow_up: . Thanks for documenting it!

etpinard:
@alexcjohnson 

> Do the tests I added satisfy the TODO up top?

Absolutely.

:dancer: thanks for wrapping this one up :tada:

PR #1531, opened 2017-03-29
make deleteFrames(gd) delete all frames

rreusser:
This PR changes the behavior of `deleteFrames(gd)` to delete *all* frames. Previously it threw a slice-null error. Contrast with `deleteTraces(gd)` which throws an error with a meaningful message or `animate(gd)` which animates all frames. In this case I chose to match the behavior of animate.


etpinard:
Looks good :dancer:

Issue #1526, opened 2017-03-29
BUG: After a 'pan' restyle function doesn't color points


destradafilm:
Using the plotly codepen for 'click and double-click event'
http://codepen.io/plotly/pen/WoZOdq

If you click a point, it restyles the point colors. But if you 'pan' the plot and then click the point, it doesn't restyle. Only after reseting the plot does it show the point with the new color
![pan then restyle bug](https://cloud.githubusercontent.com/assets/6043499/24473026/1fba6372-1485-11e7-87e6-d1ee3c71c27c.gif)

etpinard:
Ok interesting. Thanks for reporting. This is probably related to https://github.com/plotly/plotly.js/pull/1488 in which your restyle call doesn't trigger a per-pt `recalc`.

By restyling only `marker.color` instead of the whole `marker` container, things are appear to be working correctly: http://codepen.io/etpinard/pen/ZempLG

etpinard:
will be fixed in https://github.com/plotly/plotly.js/pull/1536

PR #1536, opened 2017-03-30
Assume recalc is needed when restyling attribute containers

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1526

So that:

```js
Plotly.restyle(gd, {
  marker: {
    color: ['red', 'blue', 'green']
  }
})
```

works in all situations.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1585,6 +1585,7 @@ function _restyle(gd, aobj, _traces) {
                     flags.docalc = true;
                 }
 
+                // must redo calcdata when restyling array values of arrayOk attributes
                 if(valObject.arrayOk && (Array.isArray(newVal) || Array.isArray(oldVal))) {
                     flags.docalc = true;
                 }
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1580,6 +1580,11 @@ function _restyle(gd, aobj, _traces) {
                 var moduleAttrs = (contFull._module || {}).attributes || {};
                 var valObject = Lib.nestedProperty(moduleAttrs, ai).get() || {};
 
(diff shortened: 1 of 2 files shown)


src/plot_api/plot_api.js
```diff
@@ -1580,6 +1580,12 @@ function _restyle(gd, aobj, _traces) {
                 var moduleAttrs = (contFull._module || {}).attributes || {};
                 var valObject = Lib.nestedProperty(moduleAttrs, ai).get() || {};
 
+                // if restyling entire attribute container, assume worse case
+                if(!valObject.valType) {
```
etpinard: A little bit hacky as this relies on the fact that attribute containers (e.g. `marker`) don't have a `valType` field set in the attribute declarations. 

Morevoer, we could optimize this further by digging into the attribute container for `arrayOk` keys, but I thought this wasn't necessary. For best-performing `restyle` calls, users should target specific attributes e.g. `restyle(gd, 'marker.color', [['red', 'blue', 'green']])`.

alexcjohnson: if an attribute container *did* have a `valType` field, would (or could) the schema validation tests fail?

I agree 💯  it's not worth optimizing this.

etpinard: It would because if it had a `valType` field it would only need a valid `role` (i.e. set to `'info' | 'style' | 'data'`) field to pass this [test](https://github.com/plotly/plotly.js/blob/caade64acd65f7a0dfa52932115a6aeee2aeb37a/test/jasmine/tests/plotschema_test.js#L38-L46).

alexcjohnson: Great, then I'm happy with this hack.

alexcjohnson:
💃

Issue #1529, opened 2017-03-29
scattergeo does not respect datum position in arrayOk attributes


etpinard:
For example

```js
Plotly.plot('graph', [{
  type: 'scattergeo',
  lon: [10, null, 40],
  lat: [0, null, 0],
  marker: {
    color: ['red', 'green', 'blue'],
    size: [20, 50, 10]
  }
}])
```

yields

![image](https://cloud.githubusercontent.com/assets/6675409/24476938/3210ce9e-14a3-11e7-98c0-48d5b8e69815.png)

see http://codepen.io/etpinard/pen/QpJGeq

---

This issues comes from https://github.com/plotly/plotly.js/pull/1004 where we started [skipping](https://github.com/plotly/plotly.js/pull/1004/commits/52c1b89d2f90453f259f766d7c79012e4bff1827#diff-c9c536870c5f780da73fb82c7e5f15bfR42) over items with non-numeric `lon` and/or `lat` coordinates.

Post https://github.com/plotly/plotly.js/pull/1519#discussion_r108450645, I'm having second thoughts about skipping non-numeric items in the calc step though [`mapbox/calc.js`](https://github.com/plotly/plotly.js/blob/master/src/traces/scattermapbox/calc.js) appears to get this right. Ignoring non-numeric items early makes sense for :racehorse: as we save a few `isNumeric` call in hot code paths (e.g. in `<>/plot.js`), but it does lead to some headaches elsewhere in the code (e.g. those [`placeholder`](https://github.com/plotly/plotly.js/pull/1004/commits/dd05c6ee857ff5e5e537ccf946d715e60f9924e2) flags) :confused:

alexcjohnson:
if `isNumeric` is a concern (I'm not sure it is compared to everything else that needs to happen in `<>/plot.js`, but it wouldn't hurt to optimize this) we should be able to ensure `<>/calc.js` replaces all non-numeric points with `BADNUM` and just use `!== BADNUM` which is about as fast as can be imagined.

Right now scatter uses `false` for non-numerics but that was an ancient decision... if we standardize on something it should be `BADNUM`.

PR #1538, opened 2017-03-30
Scattergeo BADNUM

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1529

This PR makes the `scattergeo` calcdata have as many items as there are `lon/lat` coordinates or `locations` items in the full data. In other words, it :hocho: the overly-complicated logic that _skipped_ non-numeric items in `calc.js`. Now, non-numeric coordinates are set to `BADNUM` in `calc.js` and filtered out in `plot.js`. Turns out this cleans up a good amount of code and makes it possible for `scattergeo` to reuse [`scatter/arrays_to_calcdata.js`](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/arrays_to_calcdata.js) straight up :tada: 

Now, commit https://github.com/plotly/plotly.js/commit/a705ed103c86bdbc41630c04b786f41d19e5ed84 breaks `scattermapbox` as it too depends `calcTraceToLineCoords`. I'll fix this on another branch based off this one.

In the meantime (i.e. before adding mucho tests) I would appreciate some :+1: s from @alexcjohnson to get this right once and for all.


src/lib/geojson_utils.js
```diff
@@ -21,18 +23,19 @@
...
+    var lineString = [];
 
     for(var i = 0; i < calcTrace.length; i++) {
         var calcPt = calcTrace[i];
+        var lonlat = calcPt.lonlat;
 
-        lineString.push(calcPt.lonlat);
-
-        if(!connectgaps && calcPt.gapAfter && lineString.length > 0) {
+        if(lonlat[0] !== BADNUM && lonlat[1] !== BADNUM) {
```
alexcjohnson: This is great for robustness (so this comment is nonblocking), but for 🐎  (with sufficient test coverage) we could ensure in calc that if the point is invalid, BOTH components get set to `BADNUM`, so you only need one test.

etpinard: good call

src/traces/scattergeo/calc.js
```diff
@@ -10,45 +10,34 @@
...
-            var lon = trace.lon[i],
-                lat = trace.lat[i];
+            var lon = trace.lon[i];
+            var lat = trace.lat[i];
+            calcPt.lonlat = new Array(2);
 
-            calcPt.lonlat = [+lon, +lat];
-            skip = (!isNumeric(lon) || !isNumeric(lat));
+            calcPt.lonlat[0] = isNumeric(lon) ? +lon : BADNUM;
+            calcPt.lonlat[1] = isNumeric(lat) ? +lat : BADNUM;
```
alexcjohnson: probably all we'd need to do for https://github.com/plotly/plotly.js/pull/1538/files#r109043439 to work is change this to something like
```js
if(isNumeric(lon) && isNumeric(lat)) calcPt.lonlat = [+lon, +lat];
else calcPt.lonlat = [BADNUM, BADNUM]; // or even [BADNUM] or []
```

alexcjohnson:
I guess if the goal here is "get this right once and for all" then we probably should reduce the plot-level filters to a single `!== BADNUM` test, per https://github.com/plotly/plotly.js/pull/1538#discussion_r109043439

After we have that, this looks a great pattern to standardize on. 🎉

src/traces/scattergeo/calc.js
etpinard: Sounds good! Thanks for taking a :eye:

src/traces/scattergeo/plot.js
```diff
@@ -72,66 +66,39 @@ module.exports = function plot(geo, calcData) {
...
 
         if(subTypes.hasMarkers(trace)) {
-            s.selectAll('path.point').data(_calcTrace)
-              .enter().append('path')
-                .classed('point', true);
+            s.selectAll('path.point')
+                .data(Lib.identity)
+             .enter().append('path')
+                .classed('point', true)
+                .each(function(calcPt) { removeBADNUM(calcPt, this); });
```
alexcjohnson: For the record, I just had a discussion with @etpinard about this where my gut reaction was that we should do
```js
s.selectAll('path.point')
    .data(filterOutBadPoints)
  .enter().append('path')
    ...
```
instead of creating all the nodes and then deleting the bad ones.

The conclusion we came to though is that the way it is here is actually better, despite being a bit less "d3-idiomatic." The rationale is that bad points should be considered an edge case, so we don't worry about the overhead of creating and deleting their nodes (which will happen on every replot), just about how the `.each` call to test (good) points for removal compares to whatever `filterOutBadPoints` does. And since the latter would *always* need to duplicate the `calcdata` array, it's almost certainly slower than this `.each`.

alexcjohnson:
Looks great to me (once tests pass, of course) 💃

etpinard:
> (once tests pass, of course)

Will passed once https://github.com/plotly/plotly.js/pull/1543 is merged.

Issue #1539, opened 2017-03-30
Enforce ES5 only


rreusser:
It would be nice to include this as a test step. A quick script:

```javascript
var CLIEngine = require("eslint").CLIEngine;

var cli = new CLIEngine({
    envs: ["browser", "mocha"], // ??? just copied it. not sure what this does.
    useEslintrc: false,
    ignore: false,
    parserOptions: {
        ecmaVersion: 5
    }   
});

var report = cli.executeOnFiles(["dist/plotly.js"]);
var formatter = cli.getFormatter();

console.log(formatter(report.results));

if (report.errorCount > 0) {
    process.exit(1);
}
```

Running this on 1.25.1 gives:

```
$ node es5only.js

/Users/rreusser/plotly/plotly.js/dist/plotly.js
  7:1  error  Parsing error: The keyword 'const' is reserved

✖ 1 problem (1 error, 0 warnings)
```

Remaining tasks:
1. find the right place for it (tasks?)
2. fix the ignore details. Right now it uses eslintignore which excludes everything we want to include. We either need to grab the directory listing and filter out `/\.min\.js$` manually or figure out the right incantation. Should be simple.

etpinard:
> find the right place for it (tasks?)

https://github.com/plotly/plotly.js/blob/master/tasks/test_syntax.js

rreusser:
Can just use ``

var report = cli.executeOnFiles(fs.readdirSync(__dirname + '/dist').filter(x => /[^(min)].js$/.test(x)));

rreusser:
Closing since it's all on the PR now.

PR #1540, opened 2017-03-30
Add ES5 check to syntax test

rreusser:
See #1539. Uses another eslint pass that applies no rules except for an `ecmaVersion: 5` parser setting. Before the const fix, this change causes plotly-latest to return:

```sh
$ npm run test-syntax

/Users/rreusser/plotly/plotly.js/dist/plotly-gl2d.js
  435:1  error  Parsing error: The keyword 'const' is reserved

/Users/rreusser/plotly/plotly.js/dist/plotly-gl3d.js
  1105:1  error  Parsing error: The keyword 'const' is reserved

/Users/rreusser/plotly/plotly.js/dist/plotly-with-meta.js
  1916:1  error  Parsing error: The keyword 'const' is reserved

/Users/rreusser/plotly/plotly.js/dist/plotly.js
  1916:1  error  Parsing error: The keyword 'const' is reserved

✖ 4 problems (4 errors, 0 warnings)

test-syntax error [es5-only syntax]

ok no jasmine suites focus/exclude blocks
ok circular dependencies: 17
ok correct headers and contents in lib/ and src/
ok lower case only file names
ok trailing new line character
/Users/rreusser/plotly/plotly.js/tasks/test_syntax.js:250
        throw new Error('test syntax failed.');
        ^

Error: test syntax failed.
    at process.<anonymous> (/Users/rreusser/plotly/plotly.js/tasks/test_syntax.js:250:15)
    at emitOne (events.js:96:13)
    at process.emit (events.js:188:7)
```

After fixing:

```sh
$ npm run test-syntax

ok es5-only syntax
ok no jasmine suites focus/exclude blocks
ok circular dependencies: 17
ok correct headers and contents in lib/ and src/
ok lower case only file names
ok trailing new line character
```


rreusser:
Note: CI failure can be considered a success. 🎉

tasks/test_syntax.js
```diff
@@ -188,6 +190,47 @@ function assertCircularDeps() {
...
+        }
+    });
+
+    // Filter out min and plotly-geo-assets.js since one is unnecessary
+    // and the other is super slow:
+    var files = fs.readdirSync(path.join(__dirname, '../dist'));
+    var validFiles = [];
+    for(var i = 0; i < files.length; i++) {
+        var f = files[i];
+        var isMin = !/[^(min)]\.js$/.test(f);
```
rreusser: I've inverted negated a negation… that's silly.

etpinard: Why not just use this [list](https://github.com/plotly/plotly.js/blob/master/tasks/util/constants.js#L19), something like:

```js
var validFiles = constants.partialBundlePaths(function(p) { return p.dist; });
validFiles  = validFiles.concat(constants.pathToPlotlyDist)l

var report = cli.executeOnFiles(validFiles);
// ...
```

looks easier to read to me. Thoughts?

rreusser: Sounds great. Didn't know it was a thing. Will update once confirmed it works.

tasks/test_syntax.js
```diff
@@ -188,6 +190,47 @@ function assertCircularDeps() {
...
+        var isGeo = /geo-assets/.test(f);
+        if(!isMin && !isGeo) {
+            validFiles.push(path.join(__dirname, '../dist', f));
+        }
+    }
+
+    var report = cli.executeOnFiles(validFiles);
```
etpinard: Too bad `eslint` can't do this in an async way. This does slow down `test-syntax` a lot. Oh well, it's worth it :+1:

tasks/test_syntax.js
rreusser: Fixed. Included `plotly.js` but left out with `with-meta` to save a few seconds.

tasks/test_syntax.js
rreusser: There's always the possibility of just testing the regular bundle only since there's tons of duplication between the bundles, but I'd need to investigate to figure out which parts are a subset of which parts.

rreusser:
Fixed file lookup. FYI you can just switch ecmaVersion between 5 an 6 to test success/failure.

etpinard:
:dancer: after merging `master`

PR #1543, opened 2017-03-31
Scattermapbox BADNUM + mapbox-gl@0.33.1

etpinard:
to be merged in https://github.com/plotly/plotly.js/pull/1538 - fixing the test failures there.

In brief, this PR:

- makes `scattermapbox/calc.js` reuse `scattergeo/calc.js` :recycle: 
- :hocho: the last remaining signs of calcdata `placeholder` 
- bumps `mapbox-gl` to [`^0.33.1`](https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#0331-march-10-2017). Among a few interesting performance improvements, this new version allows use the [`identity`](https://www.mapbox.com/mapbox-gl-js/example/data-driven-lines/) property function which :hocho: the hacky `stops` logic in `scattermapbox/convert.js` :tada: 

### Important

- `mapbox-gl@0.33.1` does not get along with the `nw.js` version we use in our image test container (same as on the plot.ly image server) - see [here](https://gist.github.com/etpinard/5d58ef47d65d65cd1e93d825659936af#news) for more details. This is probably a good time to bump the `nw.js` version. The latest `nw.js` version is `0.21.4` while we're still using `0.12` cc @bpostlethwaite 

- Note that I couldn't update `mapbox-gl` to their latest version (`v0.34.0`) because that version is currently [broken](https://github.com/mapbox/mapbox-gl-js/pull/4486/) in browserified bundles and that patch hasn't been published yet.


src/plots/plots.js
```diff
@@ -2012,11 +2012,8 @@ plots.doCalcdata = function(gd, traces) {
         //
         // This ensures there is a calcdata item for every trace,
         // even if cartesian logic doesn't handle it (for things like legends).
-        //
-        // Tag this artificial calc point with 'placeholder: true',
-        // to make it easier to skip over them in during the plot and hover step.
         if(!Array.isArray(cd) || !cd[0]) {
-            cd = [{x: false, y: false, placeholder: true}];
+            cd = [{x: false, y: false}];
```
etpinard: Good call. Let me try that!

alexcjohnson:
This looks great! Really nice simplification. Lets get the `nw` version bumped so we can get *all* images tested before merging this, but other than that 💃 !!

src/plots/plots.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1564/commits/b5cd57308c60b584ba47ba93b35bb02efdd15418 

Didn't break any tests. The only place I can think of that this would impact in the legend code, but looks like it isn't making any assumptions on `x` / `y` [here](https://github.com/plotly/plotly.js/blob/master/src/components/legend/style.js).

etpinard:
Replaced by https://github.com/plotly/plotly.js/pull/1564

PR #1544, opened 2017-03-31
Optimize performance of setCategoryIndex

hy9be:
I added an auxiliary map for ``ax._categories`` array to avoid using ``Array.prototype.indexOf`` function searching for the existence of a category.

I tested [here](https://jsfiddle.net/smileyhaowen/cvLzxz7L/2/) and observed ~20% performance improvements on big amount of traces.

The unit test failed at some range_slider cases but I cannot figure out how my changes are related to those test cases.

### DIFF
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -148,8 +148,13 @@ module.exports = function setConvert(ax, fullLayout) {
         // d2l/d2c variant that that won't add categories but will also
         // allow numbers to be mapped to the linearized axis positions
         var index;
-        if(ax._categoriesMap) { index = ax._categoriesMap[v] ? ax._categoriesMap : undefined; }
-        if(index !== undefined) { return index; }
+        if(ax._categoriesMap) {
+            index = ax._categoriesMap[v] ? ax._categoriesMap : undefined;
+            if(index !== undefined) return index;
+        } else {
+            index = ax._categories.indexOf(v);
+            if(index !== -1) return index;
+        }
         if(typeof v === 'number') { return v; }
     }
 
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -126,12 +126,16 @@ module.exports = function setConvert(ax, fullLayout) {
      */
     function setCategoryIndex(v) {
         if(v !== null && v !== undefined) {
-            var c = ax._categories.indexOf(v);
-            if(c === -1) {
+
+            if(ax._categoriesMap[v] === undefined) {
                 ax._categories.push(v);
-                return ax._categories.length - 1;
+
+                var curLength = ax._categories.length - 1;
+                ax._categoriesMap[v] = curLength;
+
+                return curLength;
             }
-            return c;
+            return ax._categoriesMap[v];
         }
         return BADNUM;
     }
@@ -325,6 +329,8 @@ module.exports = function setConvert(ax, fullLayout) {
 
         // TODO cleaner way to handle this case
         if(!ax._categories) ax._categories = [];
+        // Add a map to optimize the performance of category collection
+        if(!ax._categoriesMap) ax._categoriesMap = {};
 
         // make sure we have a domain (pull it in from the axis
         // this one is overlaying if necessary)
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -147,10 +147,8 @@ module.exports = function setConvert(ax, fullLayout) {
     function getCategoryIndex(v) {
         // d2l/d2c variant that that won't add categories but will also
         // allow numbers to be mapped to the linearized axis positions
-        var index;
-        // Get the index from the aux map if it was built, otherwise search the _categories array
         if(ax._categoriesMap) {
-            index = ax._categoriesMap[v];
+            var index = ax._categoriesMap[v];
             if(index !== undefined) return index;
         }
 
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -126,8 +126,12 @@ module.exports = function setConvert(ax, fullLayout) {
      */
     function setCategoryIndex(v) {
         if(v !== null && v !== undefined) {
+            if (ax._categoriesMap === undefined) 
+                ax._categoriesMap = {};
 
-            if(ax._categoriesMap[v] === undefined) {
+            if (ax._categoriesMap[v] !== undefined) {
+                return ax._categoriesMap[v];
+            } else {
                 ax._categories.push(v);
 
                 var curLength = ax._categories.length - 1;
@@ -135,7 +139,6 @@ module.exports = function setConvert(ax, fullLayout) {
 
                 return curLength;
             }
-            return ax._categoriesMap[v];
         }
         return BADNUM;
     }
@@ -143,8 +146,10 @@ module.exports = function setConvert(ax, fullLayout) {
     function getCategoryIndex(v) {
         // d2l/d2c variant that that won't add categories but will also
         // allow numbers to be mapped to the linearized axis positions
-        var index = ax._categories.indexOf(v);
-        if(index !== -1) return index;
+        if(ax._categoriesMap)
+            var index = ax._categoriesMap[v]?ax._categoriesMap:undefined;
+        
+        if(index !== undefined) return index;
         if(typeof v === 'number') return v;
     }
 
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -148,6 +148,7 @@ module.exports = function setConvert(ax, fullLayout) {
         // d2l/d2c variant that that won't add categories but will also
         // allow numbers to be mapped to the linearized axis positions
         var index;
+        // Get the index from the aux map if it was built, otherwise search the _categories array
         if(ax._categoriesMap) {
             index = ax._categoriesMap[v];
             if(index !== undefined) return index;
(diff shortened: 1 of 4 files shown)


etpinard:
Thanks very much for the PR @hy9be :beers: 

Looks like the image mocks testing axis `categoryarray` / `categoryorder` are failing

image

----

To run those tests locally, have a look at image test [README](https://github.com/plotly/plotly.js/blob/master/test/image/README.md).

hy9be:
@etpinard Ok I will take a look.

etpinard:
and also `npm run test-jasmine -- calcdata` is failing (which should be easier to debug than the image tests).

hy9be:
@etpinard It relates to ``cateoryorder`` cases. I will need to take a further look.

hy9be:
@etpinard Is it okay to use the map object in ES6 in Plotly.js? The ES5 map does not allow non-string keys, which failed some cases.

hy9be:
Got it. I will see if I could resolve the breaks in some other ways. Thanks!

hy9be:
@etpinard Had 1 last jasmine test case failed: https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/axes_test.js#L1853-L1877. What is this feature about?

My error is:
```
Expected [ 'B?', 'B->C', 'd', 'd', '' ] to equal [ 'A!', 'B?', 'B->C', 'c', 'd', 'd', '' ].
	    at Object.<anonymous> (/var/folders/4z/81cn93jd36q64qf2lbd48bx0lx05kn/T/tests/axes_test.js:1707:0 <- /var/folders/4z/81cn93jd36q64qf2lbd48bx0lx05kn/T/68c4de161f2f1ec90f156c4f21faa6b8.browserify:195317:29)
	Expected 'a' to be 'A!'.
	    at Object.<anonymous> (/var/folders/4z/81cn93jd36q64qf2lbd48bx0lx05kn/T/tests/axes_test.js:1708:0 <- /var/folders/4z/81cn93jd36q64qf2lbd48bx0lx05kn/T/68c4de161f2f1ec90f156c4f21faa6b8.browserify:195318:42)

```

alexcjohnson:
> https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/axes_test.js#L1853-L1877. What is this feature about?

Haha that test is my doing, it's testing a collection of edge cases with the ticktext/tickvals feature, which lets you simultaneously override both the positions and labels of tick marks. The normal usage is stuff like https://plot.ly/javascript/axes/#enumerated-ticks-with-tickvals-and-ticktext

In the category case, `tickvals` entries can either be category strings, or numbers corresponding to the serial numbers assigned to the categories. The reason we need to allow numbers is they can also be fractional values if you want to put a tick in between categories or something. For example, maybe you want to label two neighboring categories with one tick halfway between them. There would be no way to do this using the category string.

And then after you provide a tick value, you *may* choose to override its text with your own string, but if you don't override it the automatic label is used (in this case the category string).

So in this case the problem is either actually a real regression the test is picking up, or the `mockCalc` function in this test is incompatible with the changes you've made and needs updating. The way to tell the difference would be to create the plot that this test is trying to mock, and see if it behaves correctly. In this case I'd do:
```js
Plotly.newPlot(gd, [{x: ['a', 'b', 'c', 'd'], y: [1, 2, 3, 4]}],
    {xaxis: {
        range: [-0.5, 4.5],
        tickvals: ['a', 1, 1.5, 'c', 2.7, 3, 'e', 4, 5, -2],
        ticktext: ['A!', 'B?', 'B->C']
    }});
```
(the rest of `ax` is probably included just to get the right pieces in place since the test doesn't call `supplyDefaults`)
If I do that on the master branch I see:
screen shot 2017-04-05 at 11 54 09 am
which you can see gives the expected tick labels `[ 'A!', 'B?', 'B->C', 'c', 'd', 'd', '' ]`. What do you see on your branch?

hy9be:
Ah got it. I did not notice the rendering result. Will doublecheck. Thanks @alexcjohnson !

hy9be:
CI reported error:
image

But in my local environment the test suites always pass:
image

hy9be:
Btw is that a way to re-run the CI test? Every time I had to make some dummy commits now...

alexcjohnson:
Ah yes, I've seen that test fail intermittently. You should be able to retry the test by clicking "rebuild" on the test results page, but I'm glad you got this to work anyway. I'm away from my computer but will review this tomorrow night.

rreusser:
(I think you might need to have permissions and be logged in in order to retry.)

hy9be:
@rreusser you are right, the button is disabled:

image

alexcjohnson:
> you are right, the button is disabled:

Ah sorry, I thought I was looking at it signed-out on my phone... well, one more reason for us to clamp down on intermittent test failures!

src/plots/cartesian/set_convert.js
```diff
@@ -148,8 +148,13 @@ module.exports = function setConvert(ax, fullLayout) {
...
         var index;
-        if(ax._categoriesMap) { index = ax._categoriesMap[v] ? ax._categoriesMap : undefined; }
-        if(index !== undefined) { return index; }
+        if(ax._categoriesMap) {
+            index = ax._categoriesMap[v] ? ax._categoriesMap : undefined;
+            if(index !== undefined) return index;
+        } else {
+            index = ax._categories.indexOf(v);
+            if(index !== -1) return index;
+        }
```
alexcjohnson: If this block is *only* here to fix that test, and never gets hit in real code (which looks to be the case to me, since `ax._categories` and `ax._categoriesMap` always get populated simultaneously) then we should leave out the `else` block here, and just make the test more realistic by creating the corresponding `_categoriesMap` in [the mock `ax` object](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/axes_test.js#L1854-L1862).

For the record, this made me curious "do we still need `ax._categories` at all?" But you were right to not remove it, as `getCategoryName` (`ax.c2d` etc, ie hover info) still needs the reverse mapping to be fast.

hy9be: Totally agree. I will make the change on the mock object instead.

Yeah I tried to completely replacing the ``_categories`` array with the map, but there will be some cases we need a reverse mapping. 

Actually, comparing to maintaining a pair of synced-up array and map, I feel there must be some better data structure for a performant categories collection. But I was sort of lazy and did not give it another think anymore.

src/plots/cartesian/set_convert.js
```diff
@@ -152,10 +152,8 @@ module.exports = function setConvert(ax, fullLayout) {
         if(ax._categoriesMap) {
             index = ax._categoriesMap[v];
             if(index !== undefined) return index;
-        } else {
-            index = ax._categories.indexOf(v);
-            if(index !== -1) return index;
```
alexcjohnson: Awesome, I'm glad it was as easy as that :)

Apologies for the iterations, but one more related request since this is a perf PR of a couple of very hot code paths: I don't think that either `getCategoryIndex` or `setCategoryIndex` needs the fallbacks for missing `_categoriesMap` - the previous version, using only `_categories`, didn't have these fallbacks, and every path that creates `_categories` also creates `_categoriesMap`, right?

(also 📚  your [comment above this](https://github.com/plotly/plotly.js/pull/1544/commits/d2be6c94f3291c42666bf5c3f086968f42010ba2#diff-bddbbac00596941548be19d6840c720aL151) is out of date now)

src/plots/cartesian/set_convert.js
alexcjohnson: > I feel there must be some better data structure for a performant categories collection.

Good point, I'm sure there is, though it won't be built in so is unlikely to improve performance, it would just clean up the API... lets not worry about it for now, this two-part solution is light and easy enough to use here but we can look into it further if this comes up in enough other places.

src/plots/cartesian/set_convert.js
hy9be: Right now no functions will fallback to use ``indexOf`` on ``_categories`` anymore.

Currently I made sure ``_categoriesMap`` is always populated accordingly when ``_categories`` is created or initialized with some items for the current code. But this is not very robust way. Some other contributors could change ``_categories``  without updating ``_categoriesMap`` in the future. And it will not be easy for them to debug the unit test and find out the reason behind breaks.

alexcjohnson: Oh sorry, I just meant we can remove the fallbacks for if `_categoriesMap` doesn't even exist, ie in `getCategoryIndex` changing:
```js
if(ax._categoriesMap) {
    var index = ax._categoriesMap[v];
    if(index !== undefined) return index;
}
```
to just:
```js
var index = ax._categoriesMap[v];
if(index !== undefined) return index;
```
and removing from `setCategoryIndex`:
```js
if(ax._categoriesMap === undefined) {
    ax._categoriesMap = {};
}
```

As far as I can tell, *that* much should be robust. Not a huge deal but should be a small performance boost.

But you raise a good point about future changes, I guess that's an argument for converting these two to a single bidirectional map object even if it doesn't yield any additional perf benefits.

hy9be: Oh I get it. Sorry I misunderstood your point.

Well, I feel a liiiittle bit uncomfortable of doing that, because if there is any case that the map is not created, it will result in an uncaught exception, which to me, seems more severe than a messed up chart.

And I tried on [jsperf](https://jsperf.com/check-null-overhead/1), the overhead of checking null is less than 10% on Chrome:
image
And considering the performance of reading a map itself, 10% differences here will not be a big cost for the whole workflow.

For Safari the differences are even less.
image

alexcjohnson: That's fine, I'm pretty confident that it's OK since that's how it worked previously, but I can take responsibility for removing it (in another PR) and ensuring we have sufficient test coverage. You're right that the penalty is much smaller than the improvement you've made here (though I'd be wary of that jsperf result, I suspect in Safari the compiler has optimized away almost everything you see there).

So I think this is ready to go! Nice job @hy9be !

hy9be: Good to know! I used to trust the results I got from jsperf a lot...
Thanks a lot for reviewing my code! @alexcjohnson

PR #1550, opened 2017-04-03
More symbols for gl2d and gl3d plots

etpinard:
This PR supersedes @destradafilm's https://github.com/plotly/plotly.js/pull/1482


test/image/mocks/marker_symbols.json
```diff
@@ -0,0 +1,99 @@
+{
+  "data": [
+    {
+      "x": [1, 2, 3, 4, 5, 6, 7, 8, 9, 1, 2, 3, 4, 5, 6, 7, 8, 9, 1, 2, 3, 4, 5, 6, 7, 8, 9, 1, 2, 3, 4, 5, 6, 7, 8, 9, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+      "y": [1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 5, 5, 5, 5, 5],
+      "mode": "markers",
+      "text": [
+        "marker number: 0<br>marker string: 'circle'",
```
etpinard: @alexcjohnson Turns out we didn't have a mock testing _all_ our marker symbols.

I dug out https://github.com/plotly/streambed/pull/358 and https://github.com/plotly/documentation/pull/16 and made this one here.

alexcjohnson: Good catch! Can you include just a few examples of 1xx (`-open`), 2xx (`-dot`) and 3xx (`-open-dot`) in this mock? It should be able to support just *one* more row 😉 

And then we'll be able to see how this compares with the gl implementation, where I'm pretty sure colors are handled differently in the `-open` case than they currently are for svg, ie we would never have a double line like gl2d currently has.

test/image/mocks/gl3d_marker_symbols.json
```diff
@@ -0,0 +1,163 @@
...
+        "range": [0, 2]
+      }
+    },
+    "annotations": [
+      {
+        "x": 0.5,
+        "y": 1,
+        "xref": "paper",
+        "yref": "paper",
+        "text": "Add 100 (or '-open') to marker number (or string) to remove the fill color,<br>Add 200 (or '-dot') to marker number (or string) to add a dot in the middle,<br>Add 300 (or '-open-dot', 'dot-open') to marker number (or string) for both.",
```
alexcjohnson: I presume `(-open)?-dot` is not supported in gl? That's fine, but we should be clear about this.

src/constants/gl_markers.js
```diff
@@ -17,5 +17,40 @@ module.exports = {
...
+    'diamond-tall-open': '♢',
+    'hourglass': '⧗',
+    'hourglass-open': '⧖',
+    'bowtie': '⧓',
+    'bowtie-open': '⋈',
+    'circle-cross': '⊕',
+    'circle-x': '⊗',
+    'square-cross': '⊞',
+    'square-x': '⊠',
+    'asterisk': '✱'
```
alexcjohnson: There are two other tweaks this is going to need before it can really match how the svg symbols behave:
- all the symbols should have roughly the same scale. You can see from the svg test image that this is a little bit subjective, for example the side of the square is the same as the diameter of the circle, but the outer width of X and + are a little bigger - basically I tweaked each one until put on a plot next to the other I thought they all looked "about the same size"
- the symbol should be placed with its centroid at the data location, not the center of the bounding rect. For most of the symbols these are the same, but for the triangles they are not (and I guess also for the pentagons and 5-sided stars, to a lesser extent).

alexcjohnson: Basically this is going to require passing symbol-dependent offsets and scales to the relevant `gl-vis` modules. @destradafilm  not sure if you want to take a crack at this, I'm afraid it'll be a little bit tedious and @etpinard and I may not be able to get to it for a while.

etpinard:
fyi @dfcreative

src/constants/gl_markers.js
destradafilm: @alexcjohnson I'd like to take a crack at it, but it will take a while for me since I've got a full plate at the moment, as well.

etpinard:
@dfcreative you're ok taking over this one?

dfcreative:
@etpinard yes. Just to clarify: it is expected to just extend various symbols, or to consider @alexcjohnson's suggestions on optical alignment? Also some OS don't have symbols declared, eg. for me it looks like:
image

As a workaround, it is possible to replace bad symbols with ones from http://www.copypastecharacter.com/. But manual alignment is unsustainable in this case because symbol size/glyph vary across different OS.

Good solution − come up with optically balanced marker font (similar to [wavefont](https://npmjs.org/package/wavefont)) and load is via [add-font](https://npmjs.org/package/add-font) or alike.

Which solution sounds more preferable?

etpinard:
> To clarify: it is expected to just extend various symbols, or to consider @alexcjohnson's suggestions on optical alignment?

Yes, make sure @alexcjohnson 's https://github.com/plotly/plotly.js/pull/1550#discussion_r109512434 is addressed. 

>  Also some OS don't have symbols declared,

Do you have an idea of which symbols are available on each of the major OS? (Plotly has a https://www.browserstack.com if you need one). We don't have to include all of them. If we can add the triangles and a few star-like symbols that work in most OS in this PR, I'd be very happy.

> Good solution − come up with optically balanced marker font (with technique similar to wavefont) and load is via add-font or alike.

That's sounds like a better long-term solution. I would prefer waiting for `scattergl` v2 though.

dfcreative:
Ok just finished [a tool](https://dfcreative.github.io/optical-properties/) for aligning markers size/alignment, following @alexcjohnson suggestions. We are to use that in order to look markers consistent across devices/OS, because right now glyphs are very different.

Gotta introduce it into _font-atlas-sdf_ along with handpicked subset of crossplatform symbols.

etpinard:
> Introduced that into font-atlas-sdf.
> Preparing a patch along with handpicked subset of crossplatform symbols to close this issue.

Excellent.

dfcreative:
Branch https://github.com/dfcreative/plotly.js/tree/pr-1482-gl-symbols includes merged master and symbols extended to be cross-platform.

https://github.com/gl-vis/gl-scatter2d-sdf/pull/4 enables markers alignment/size normalization.

test/image/mocks/marker_symbols.json
etpinard: Done in https://github.com/plotly/plotly.js/pull/1781/commits/083b00a9832211dfa1f02e511fdaf8698fedce1d

test/image/mocks/gl3d_marker_symbols.json
etpinard: See https://github.com/plotly/plotly.js/pull/1781/commits/b4db0efc205bb8377872a6747185a332d6a30f4a

etpinard:
Superseded by https://github.com/plotly/plotly.js/pull/1781

PR #1551, opened 2017-04-04
Annotation size & text alignment

alexcjohnson:
Explicit width and height for annotations, along with vertical alignment to complement the existing (horizontal) alignment.

@etpinard a couple of decisions could use 👍 /👎 :
- The attribute name `valign`, and its default value of `'middle'` - we already have default `align: 'center'`, so now if you make a giant annotation the text will end up in the center (as in the bottom left annotation I added to https://github.com/plotly/plotly.js/compare/note-sizing?expand=1#diff-038299cfa4b6b29a6fad48dc78a8c1d0)
- Text is clipped such that the `borderpad` region is always blank - see the bottom center annotation added in that same baseline image. You could argue for allowing it all the way to the border, but to me it looks better this way, and I think this overrides letting the user see another half a letter or whatever. It has a quirk though, that even if I'm clipping the text in principle exactly to its own bounding box, sometimes you lose a tiny bit. I *mostly* fixed this problem by not clipping at all if you don't set an explicit width or height (when clipping shouldn't do anything anyway) but it might still pop up on the in-principle-not-clipped sides of clipped text. I suppose there could be a solution to this involving stretching the clip rectangle on the non-clipped sides, based on whether `width` and/or `height` are set and the values of `align` and `valign` but to me that sounds more complicated than it's worth. You can see what I'm talking about if you look at my intermediate commit https://github.com/plotly/plotly.js/commit/efe83458cc0a576cbc737ebecdb602df63d32474 - the "j"s in justified and the "y" in y-int lost a tiny bit of the left and bottom respectively of their tails.


src/components/annotations/attributes.js
```diff
@@ -49,6 +49,27 @@ module.exports = {
     font: extendFlat({}, fontAttrs, {
         description: 'Sets the annotation text font.'
     }),
+    width: {
+        valType: 'number',
+        min: 0,
+        dflt: 0,
```
etpinard: Obviously the default value isn't `0`, it's computed from the input text. I believe in plotly.js lingo that means the `dflt` value should be set to `null` (similar to e.g. `contours.start` and `xbins.start`)

src/components/annotations/attributes.js
```diff
@@ -63,10 +84,21 @@ module.exports = {
...
-            '(i.e. `text` contains one or more <br> HTML tags).'
+            'Sets the horizontal alignment of the `text` within the box.',
+            'Has an effect only if `text` spans more two or more lines',
+            '(i.e. `text` contains one or more <br> HTML tags) or if an',
+            'explicit width is set to override the text width.'
+        ].join(' ')
+    },
+    valign: {
+        valType: 'enumerated',
+        values: ['top', 'middle', 'bottom'],
```
etpinard: Love it :+1:

etpinard:
> Text is clipped such that the borderpad region is always blank - see the bottom center annotation added in that same baseline image. You could argue for allowing it all the way to the border, but to me it looks better this way, and I think this overrides letting the user see another half a letter or whatever. It has a quirk though, that even if I'm clipping the text in principle exactly to its own bounding box, sometimes you lose a tiny bit. 

Hmm. I guess what we really neat is some auto-textwrapping routine. I can expect some users not liking having to add `<br>` to their annotations `text` fields to see them in full.

src/components/annotations/annotation_defaults.js
```diff
@@ -45,6 +44,11 @@ module.exports = function handleAnnotationDefaults(annIn, annOut, fullLayout, op
     coerce('textangle');
     Lib.coerceFont(coerce, 'font', fullLayout.font);
 
+    coerce('width');
+    coerce('height');
+    coerce('align');
+    coerce('valign');
```
etpinard: I guess we don't need to coerce `valign` is `height` isn't set?

alexcjohnson:
> Hmm. I guess what we really need is some auto-textwrapping routine.

At some point that'd be a nice feature, but not now, it's going to be annoying. For now I just put a note about this in the [description of width](https://github.com/plotly/plotly.js/pull/1551/files#diff-f6bb7a62be1a37018f1c5b143be592aeR60):
`There is no automatic wrapping; use <br> to start a new line.`

src/components/annotations/draw.js
```diff
@@ -118,6 +118,19 @@ function drawOne(gd, index) {
...
+    var annClipID = 'clip' + fullLayout._uid + '_ann' + index;
+    var annTextClip = fullLayout._defs.select('.clips')
+        .selectAll('#' + annClipID)
+        .data(isSizeConstrained ? [0] : []);
+
+    annTextClip.enter().append('clipPath')
+        .classed('annclip', true)
+        .attr('id', annClipID)
+      .append('rect');
+    annTextClip.exit().remove();
```
alexcjohnson: In fact they're not removed - they don't proliferate, ie if you repeatedly make 100 annotations and delete them, you'll never get more than 100 clip paths, but I didn't bother deleting clip paths corresponding to annotations past the end of the new array after removing some. I can revisit that though...

etpinard:
> There is no automatic wrapping; use <br> to start a new line.

Nice. Good enough for now!

src/components/annotations/draw.js
etpinard: Ha I see. We can revisit this when we shift to a more d3-idiomatic pattern in annotation `draw`. Then, we could easily remove `<clipPath>` inside the `exit()` selection similar to how updatemenu buttons are [removed](https://github.com/plotly/plotly.js/blob/ce06ea9cea377c5ed87ed7baa088a14a1997dbb7/src/components/updatemenus/draw.js#L103).

src/components/annotations/attributes.js
alexcjohnson: good call -> https://github.com/plotly/plotly.js/pull/1551/commits/e01a765953a290967bc97b7c72d5a3d9b91f0c03

src/components/annotations/draw.js
alexcjohnson: actually was easy to remove the unneeded clip paths after all -> https://github.com/plotly/plotly.js/pull/1551/commits/b21154bf138e4a1e49019a5ac72a1e7aca9e54bf

src/components/annotations/annotation_defaults.js
alexcjohnson: 👍  -> https://github.com/plotly/plotly.js/pull/1551/commits/2e132bd353c9502b8ef233380528e89b4adb7c1c

src/components/annotations/draw.js
```diff
@@ -72,9 +72,14 @@ function drawOne(gd, index) {
...
         options = fullLayout.annotations[index];
 
+    var annClipID = 'clip' + fullLayout._uid + '_ann' + index;
+
     // this annotation is gone - quit now after deleting it
     // TODO: use d3 idioms instead of deleting and redrawing every time
-    if(!optionsIn || options.visible === false) return;
+    if(!optionsIn || options.visible === false) {
+        d3.selectAll('#' + annClipID).remove();
+        return;
```
etpinard: easy.

etpinard:
Nicely done :dancer: 

This will be part of `1.26.0` (thanks to my `nw.js` upgrade nightmare :rage: )

PR #1552, opened 2017-04-04
Add plotly-mock-viewer as dev dependency

rreusser:
Just a two-line PR that adds [plotly-mock-viewer](http://rickyreusser.com/plotly-mock-viewer/) as a dev dependency you can start with `npm run mock-viewer`.

Pros:
- trivial to add/remove as dev dependency so no interference with plotly.js otherwise
- live-reload
- reloads mock listing on refresh
- passes json mock errors to client
- deployable to web (stays current by using cdn plotly and pulling mocks from plotly/plotly.js repo.)

Cons:
- annoying to have two devtools with complementary features
- more dev deps (budo & es2040 are the the big ones. not huge)
- autocomplete menu could be better

No hard feelings either way. I found it useful. Thought maybe others would too. Glad to transfer to plotly org if useful.


etpinard:
Thanks @rreusser :tada:  !!

We should :hocho: `devtools/test_dashboard/` and the update the CONTRIBUTING.md accordingly while at it in this PR.

But before, I'd like your `plotly-mock-viewer` to have a few more features:

- it should use our [`compress_attributes.js`](https://github.com/plotly/plotly.js/blob/master/tasks/util/compress_attributes.js) transform to trim attribute descriptions from the bundle
- it should inject `MathJax` (maybe optionally)
- make sure it supports all the same fonts as the current test dashboard (see [here](https://github.com/plotly/plotly.js/blob/b19a87f38f92b0659de0c0fb098c8e06fad7c208/devtools/test_dashboard/index.html#L5))
- it should inject @alexcjohnson 's [`strict-d3`](https://github.com/plotly/plotly.js/blob/b19a87f38f92b0659de0c0fb098c8e06fad7c208/test/image/strict-d3.js) to catch errors in old browsers
- it should allow building `plotly.js` in `build/plotly.js` so that `npm run test-image` can use the plotly.js bundle bundled by `plotly-mock-viewer` i.e. no need to `npm run watch` in a separate tab when running pixel-comparison tests

Moreover, I'm thinking it would be nice to have the following down the road:

- some abstraction to make it easier to plot multiple graphs side-by-side for comparison
- some compare with baseline option that way we could :hocho: [`devtools/image_viewer`](https://github.com/plotly/plotly.js/tree/master/devtools/image_viewer) too.

Thoughts?

--- 

By the way, I removed the `feature` label to not confused this with a _real_ plotly.js feature from a user's perspective.

rreusser:
Agreed! Also I think there was something about geojson data that I failed to carry over, though it does handle mapbox. It's perfectly usable as a standalone tool right now, so I think I'll put this on the back burner for just a bit and treat the above as a checklist to be chipped away at before merging.

rreusser:
Current status, @etpinard:

[x] add option to break everything by attempting to load mathjax
[x] add `-k` flag to keep schema metadata (otherwise applies strip transform)
[x] load fonts
[x] use strict-d3
[ ] output bundle (wontfix. see: rreusser/plotly-mock-viewer#9)
[ ] get mathjax working (not sure how to fix. see: rreusser/plotly-mock-viewer#6)
[ ] add to contributing docs (on hold unless moving forward with this)

Except for mathjax and starting to add image comparison features, I don't currently have anything further to add to this PR.

etpinard:
Ok thanks @rreusser! Let me try to get MathJax working before merging this.

rreusser:
Cool. No rush *at all* on mathjax. I was just having a pretty tricky time telling what it was trying to load or what it was supposed to be loading. It looks like I fell back to attempting to just load the live version from embedplot, but the result was still just errors: https://github.com/rreusser/plotly-mock-viewer/blob/master/lib/inject-mathjax.js#L10

etpinard:
Closing this thing. 

I love [`plotly-mock-viewer`](https://github.com/rreusser/plotly-mock-viewer) (in fact I use it everyday), but I don't see the need to make it a dev-dependency for this repo even if/when the unchecked items in https://github.com/plotly/plotly.js/pull/1552#issuecomment-294533888 get resolved. I think `plotly.js` and `plotly-mock-viewer` can happily live side-by-side in separate repos :heart:

PR #1555, opened 2017-04-05
Optimize the performance of scatter trace sorting comparator with map

hy9be:
Construct a map rather than an array to improve the performance of index search. Jasmine tests passed.

### DIFF
--- a/src/traces/scatter/plot.js
+++ b/src/traces/scatter/plot.js
@@ -51,13 +51,13 @@ module.exports = function plot(gd, plotinfo, cdscatter, transitionOpts, makeOnCo
     // Sort the traces, once created, so that the ordering is preserved even when traces
     // are shown and hidden. This is needed since we're not just wiping everything out
     // and recreating on every update.
-    for(i = 0, uids = []; i < cdscatter.length; i++) {
-        uids[i] = cdscatter[i][0].trace.uid;
+    for(i = 0, uids = {}; i < cdscatter.length; i++) {
+        uids[cdscatter[i][0].trace.uid] = i;
     }
 
     scatterlayer.selectAll('g.trace').sort(function(a, b) {
-        var idx1 = uids.indexOf(a[0].trace.uid);
-        var idx2 = uids.indexOf(b[0].trace.uid);
+        var idx1 = uids[a[0].trace.uid];
+        var idx2 = uids[b[0].trace.uid];
         return idx1 > idx2 ? 1 : -1;
     });
 


etpinard:
Looks good to me.

@rreusser can you think of any side effects (e.g. for animations)?

src/traces/scatter/plot.js
```diff
@@ -51,13 +51,13 @@ module.exports = function plot(gd, plotinfo, cdscatter, transitionOpts, makeOnCo
     // Sort the traces, once created, so that the ordering is preserved even when traces
     // are shown and hidden. This is needed since we're not just wiping everything out
     // and recreating on every update.
-    for(i = 0, uids = []; i < cdscatter.length; i++) {
-        uids[i] = cdscatter[i][0].trace.uid;
+    for(i = 0, uids = {}; i < cdscatter.length; i++) {
```
rreusser: Seems entirely equivalent to me 👍 

I didn't anticipate this would be more than like 10-15 or so, which would make the overhead of indexOf negligible. I'll keep this in mind when contemplating orders of magnitude!

etpinard:
@hy9be before merging this, would you sharing a few before / after benchmarks?

(Can't wait to wrap up https://github.com/plotly/plotly.js/pull/1511)

rreusser:
Suggested benchmark strategy for this case:

```javascript
var t0 = performance.now() // if your browser supports it
... sort traces
var t1 = performance.now()
console.log('Time taken to sort:', t0 - t1)
```

The fix seems appropriate either way, it's just very useful to have a sense for the cost and gains of optimizations like this.

hy9be:
Here is my [script](https://jsfiddle.net/smileyhaowen/qwuzevwg/2/).

And some data points collected on my computer:
image

etpinard:
@hy9be thanks for that!

src/traces/scatter/plot.js
hy9be: Yeah, actually there is almost no difference on map and array here if the amount of trace is small. My cases might be a bit crazy (with 5000+ traces). But it seems worthwhile to use map instead here since there is almost no side effect.

etpinard:
By the way, @hy9be thanks for all your contributions this week!

You get my vote for contributor of the week :trophy:

hy9be:
Wow...feel flattered. Thanks! 

I had some exposure to the idea of grammar of graphics. Plotly is not quite an implementation of that but the design is much more orthogonal and abstract than most other charting libraries. I like the project.

Will try to help more.

rreusser:
Thanks for the benchmark! Honestly I didn't think that single call would make a big enough difference to have that obvious of an effect in an end-to-end test. 🐢

Issue #1554, opened 2017-04-05
plots.addLinks is still called even set { showLink: false }


hy9be:
Not sure if this is expected. The main concern here is performance. As some measurements and reflows occurs, this function call always add about 100ms overhead on my 2015 Macbook pro. I have not tested on other devices, but I expect more delays on some older hardwares. 

![image](https://cloud.githubusercontent.com/assets/3392907/24715195/9ae301bc-19f8-11e7-8dee-c27428f0f5dd.png)

etpinard:
Thanks for the report! 

Your patch in #1557 will improve this for most users as `showLink: false` by default. But we should also try to find an alternative for that :turtle:  `getComputedTextLength()` call [here](https://github.com/plotly/plotly.js/blob/ce06ea9cea377c5ed87ed7baa088a14a1997dbb7/src/plots/plots.js#L291).

etpinard:
done in https://github.com/plotly/plotly.js/pull/1557

PR #1557, opened 2017-04-05
Remove the performance overhead of showLink property

hy9be:
A fix for: https://github.com/plotly/plotly.js/issues/1554. Do nothing if showLink is set to be false, to avoid performance overhead caused by measuring and reflows.

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -257,7 +257,7 @@ plots.previousPromises = function(gd) {
  * Add source links to your graph inside the 'showSources' config argument.
  */
 plots.addLinks = function(gd) {
-    if(gd._context.showLink === false) return;
+    if(!gd._context.showLink || !gd._context.showSources) return;
 
     var fullLayout = gd._fullLayout;
 
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -257,6 +257,7 @@ plots.previousPromises = function(gd) {
  * Add source links to your graph inside the 'showSources' config argument.
  */
 plots.addLinks = function(gd) {
+    // Do not do anything if showLink and showSources are not set to true in config
     if(!gd._context.showLink && !gd._context.showSources) return;
 
     var fullLayout = gd._fullLayout;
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -257,6 +257,8 @@ plots.previousPromises = function(gd) {
  * Add source links to your graph inside the 'showSources' config argument.
  */
 plots.addLinks = function(gd) {
+    if(gd._context.showLink === false) return;
```
etpinard: This could break graphs that have the [`showSources`](https://github.com/plotly/plotly.js/blob/ce06ea9cea377c5ed87ed7baa088a14a1997dbb7/src/plot_api/plot_config.js#L67) config option set as it is called [below](https://github.com/plotly/plotly.js/blob/ce06ea9cea377c5ed87ed7baa088a14a1997dbb7/src/plots/plots.js#L308). 

So, replacing this line with something like:

```js
if(!gd._context.showLink || !gd._context.showSources) return;
```

would be safer.

hy9be:
Code updated according to the review feedback.

src/plots/plots.js
hy9be: Got it. Thanks!

etpinard:
Unfortunately, your latest commit made `npm run test-jasmine -- config` fail:

image

hy9be:
Em, Interesting...I will take a look. Did not run the test locally as I thought it is just a trivial change.

hy9be:
I changed the OR condition into AND:
``` javascript
if(!gd._context.showLink && !gd._context.showSources) return;
```
It seems to make sense to me logically, and all tests pass. But please review. Thanks!

hy9be:
I cannot reproduce the errors reported by CI. And it does not have information about lines. Any suggestion on debugging?

etpinard:
CI uses an older version of Chrome (54, the latest is 57 I think), so maybe that's the issue.

Otherwise, have you tried running all the test suites at once with `npm run jasmine`? Maybe on test case isn't cleaning up correctly?

hy9be:
The error doesn't look like related to browser version.

Yes I run the full jasmine test. And only two noCI cases failed:

image
image

hy9be:
Run all the test cases locally again. All 1850 cases passed this time. So I add some comments to re-trigger the CI check.

etpinard:
@hy9be tests are passing. Looks like the CI failure was intermittent. Our apologies.

Great PR. Thanks againg :tada: 

Merging.

Issue #1139, opened 2016-11-14
restyling scatter trace to scattergl can lead to exceptions


etpinard:
See http://codepen.io/theengineear/pen/WowNdV

cc @dfcreative

dfcreative:
Looks like we should do #1567 here.

etpinard:
done in https://github.com/plotly/plotly.js/pull/1567

PR #1567, opened 2017-04-07
Cast to empty array

dfcreative:
Tiny fix for https://github.com/plotly/plotly.js/issues/1139
No idea why it was done for `x` but was not for `y`.

### DIFF
--- a/src/components/errorbars/index.js
+++ b/src/components/errorbars/index.js
@@ -19,7 +19,7 @@ errorBars.calc = require('./calc');
 
 errorBars.calcFromTrace = function(trace, layout) {
     var x = trace.x || [],
-        y = trace.y,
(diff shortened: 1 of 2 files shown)


etpinard:
nice find :microscope:

etpinard:
@dfcreative wanna try adding a jasmine test for this in [`gl_plot_interact_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/gl_plot_interact_test.js)? This particular [test](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/gl_plot_interact_test.js#L165) should is a good example of I'm expecting.

dfcreative:
@etpinard appended test targeting specifically this case

PR #1568, opened 2017-04-07
supplyTransformDefaults to fill in missing transformModules

rreusser:
fullLayout is sometimes an empty object. This fixes an assumption that layout._transformModules is definitely defined.


etpinard:
Looks good. Would you mind adding a test in `plots_test.js` to :lock: this down?

rreusser:
Added. Testing this in isolation meant exporting `supplyTransformDefaults`, which is probably preferable to testing a bigger chunk of code to get at this one behavior. 🔒

etpinard:
:dancer: when tests :white_check_mark:

Issue #1525, opened 2017-03-29
Scattered3D plot loads cpu and memory up to total system freeze


kelegorm:
Example: https://codepen.io/anon/pen/VpVLMN
OS X, Win 10, Chrome.

How to reproduce: run example by link above and wait some time(10-30min). While wait open a Chrome's task manager to look how many memory and cpu browser tab with example takes.

It's not usual memory leaks, it is a memory rage. Default, minimal memory is ~150Mb. Then it grows up to 170 by few seconds, then drops to default, after garbage collector have done it's  job. Max memory level grows too, I have seen tab takes more than 1Gb (comment out marker and colors in example to get 1Gb fast).

When job for GB is becoming bigger, it takes more CPU, up to 100% on my mac.

It's not a problem when you need to show charts just for few minutes, nobody notice. But for dashboards it's crutical bug.

If something isn't clear, ask me.

rreusser:
Have you edited and re-saved the codepen? It looks like most of it is commented out. I just want to be sure we're looking at the same thing.

kelegorm:
Yes, I have tried comment different things, may be they affects problem somehow. So, they just makes leaks faster.

At last time I have left it with x,y,z data only.

rreusser:
I'm not currently able to reproduce it, and I'm not sure I fully understand exactly the scenario that leads to the memory leak. I certainly would be believe there could be a memory leak, but when I run the above codepen it just sits idle. In the four minutes after the initial plot, it allocates about 700kb of memory, mostly after clicking the animate button (which to be clear just runs an action once and then does nothing further) and a bit more when moving the mouse a bit. I'm running Chrome 56.0 on Mac OS X.

The memory profile I see for that plot:

<img width="606" alt="screen shot 2017-03-29 at 16 36 12" src="https://cloud.githubusercontent.com/assets/572717/24475453/de2d8e52-149d-11e7-9808-d280e11c6c4e.png">

If possible, could you profile memory usage and confirm that it's what happens when you run the codepen with exactly the parts commented as are currently saved in that pen?

I believe there's been an effort to stamp out memory leaks in the gl code, so to anticipate what might be the problem here, one small note on animations is that if you keep adding frames, they will persist in memory. 

[...]

PR #1570, opened 2017-04-07
Lock down gl3d memomry management improvements

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1525

This is guarantees, that `npm i` will grab

- https://github.com/gl-vis/gl-plot3d/pull/8/files and
- https://github.com/mikolalysenko/matrix-camera-controller/pull/1/files


package.json
```diff
@@ -82,6 +82,7 @@
     "gl-spikes2d": "^1.0.1",
     "gl-surface3d": "^1.3.0",
     "mapbox-gl": "^0.22.0",
+    "matrix-camera-controller": "^2.1.3",
```
etpinard: `matrix-camera-controller` isn't required directly by plotly.js, but by including it as so in out package.json, we guarantee (in `npm@3` at least) that `npm i` will grab version `2.1.3` and up.

etpinard:
@dfcreative looks ok?

dfcreative:
@etpinard yes, makes total sense.

PR #1573, opened 2017-04-10
annotation hover text

alexcjohnson:
Provides hover text for annotations, using the existing hover label framework so you get stuff like automatic left/right flipping built in. Because this is a bit of a custom scenario anyway, and it's not always clear what style these hover labels should get, I included explicit styling. @etpinard and I iterated on this separately to come up with a structure that should work for styling trace hover items as well:

```js
{
  annotations: [
    {
      ...,
      hovertext: "hello!",
      hoverlabel: {
        font: {
          family: "courier",
          size: 20,
          color: "#f00"
        },
        bgcolor: "#0ff",
        bordercolor: "#080"
      }
    }
  ]
}
```
screen shot 2017-04-10 at 10 28 21 am

### DIFF
--- a/src/components/annotations/annotation_defaults.js
+++ b/src/components/annotations/annotation_defaults.js
@@ -120,6 +120,7 @@ module.exports = function handleAnnotationDefaults(annIn, annOut, fullLayout, op
             color: hoverBorder
         });
     }
+    coerce('captureevents', !!hoverText);
 
     return annOut;
 };
--- a/src/components/annotations/annotation_defaults.js
+++ b/src/components/annotations/annotation_defaults.js
@@ -12,6 +12,7 @@
 var Lib = require('../../lib');
 var Color = require('../color');
 var Axes = require('../../plots/cartesian/axes');
+var constants = require('../../plots/cartesian/constants');
 
 var attributes = require('./attributes');
 
@@ -30,7 +31,7 @@ module.exports = function handleAnnotationDefaults(annIn, annOut, fullLayout, op
     if(!(visible || clickToShow)) return annOut;
 
     coerce('opacity');
-    coerce('bgcolor');
+    var bgColor = coerce('bgcolor');
 
     var borderColor = coerce('bordercolor'),
         borderOpacity = Color.opacity(borderColor);
@@ -106,6 +107,18 @@ module.exports = function handleAnnotationDefaults(annIn, annOut, fullLayout, op
         // so we don't have to do this little bit of logic on every hover event
         annOut._xclick = (xClick === undefined) ? annOut.x : xClick;
         annOut._yclick = (yClick === undefined) ? annOut.y : yClick;
+    }
+
+    var hoverText = coerce('hovertext');
+    if(hoverText) {
+        var hoverBG = coerce('hoverlabel.bgcolor',
+            Color.opacity(bgColor) ? Color.rgb(bgColor) : Color.defaultLine);
+        var hoverBorder = coerce('hoverlabel.bordercolor', Color.contrast(hoverBG));
+        Lib.coerceFont(coerce, 'hoverlabel.font', {
+            family: constants.HOVERFONT,
+            size: constants.HOVERFONTSIZE,
+            color: hoverBorder
+        });
     }
 
     return annOut;
--- a/src/components/annotations/draw.js
+++ b/src/components/annotations/draw.js
@@ -15,6 +15,7 @@ var Plotly = require('../../plotly');
 var Plots = require('../../plots/plots');
 var Lib = require('../../lib');
 var Axes = require('../../plots/cartesian/axes');
+var Fx = require('../../plots/cartesian/graph_interact');
 var Color = require('../color');
 var Drawing = require('../drawing');
 var svgTextUtils = require('../../lib/svg_text_utils');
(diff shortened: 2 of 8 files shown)


src/components/annotations/draw.js
```diff
@@ -96,7 +97,16 @@ function drawOne(gd, index) {
...
+
+    // another group for text+background so that they can rotate together
+    var annTextGroup = annGroup.append('g')
+        .classed('annotation-text-g', true)
+        .attr('data-index', String(index));
+
+    var annTextGroupInner = annTextGroup.append('g')
+        .style('pointer-events', 'all')
+        .call(setCursor, 'default')
         .on('click', function() {
```
alexcjohnson: `plotly_clickannotation` was present, but didn't actually work before this, as far as I can tell, because it didn't have `pointer-events` turned back on. Anyway I had to do this to make hover work. A potential downside to this is that annotations grab mouse interactions even if you don't use them, so if you have a data point that's entirely under an annotation you won't see it in hover. You could imagine only grabbing mouse events if you *are* using them for hover, but then you won't get click events, and at draw time it's generally not known yet whether click events are going to be captured... so I don't see any way around this.

Anyway, I only attached click and hover to the text box, not to the arrows, and both are included in the new test.

src/components/color/index.js
```diff
@@ -57,12 +61,23 @@ color.combine = function(front, back) {
...
     var tc = tinycolor(cstr);
 
-    var newColor = tc.isLight() ?
-        tc.darken(darkAmount) :
-        tc.lighten(lightAmount);
+    if(tc.getAlpha() !== 1) tc = tinycolor(color.combine(cstr, '#fff'));
+
+    var newColor = tc.isDark() ?
+        (lightAmount ? tc.lighten(lightAmount) : '#fff') :
+        (darkAmount ? tc.darken(darkAmount) : '#000');
```
alexcjohnson: I switched from `tc.isLight` to `tc.isDark` just because [`isLight` calls `isDark`](https://github.com/bgrins/TinyColor/blob/master/tinycolor.js#L56) anyway 🐎 .

Thoughts about my TODO?
> TODO: black is what we've always done for hover, but should it be #444 instead?

(ie `Color.defaultLine`)

etpinard: > Thoughts about my TODO?

Sounds good!

src/components/annotations/draw.js
```diff
@@ -106,12 +116,33 @@ function drawOne(gd, index) {
...
-    var annTextGroupInner = annTextGroup.append('g');
+    if(options.hovertext) {
+        annTextGroupInner
+        .on('mouseover', function() {
+            var hoverOptions = options.hoverlabel;
+            var hoverFont = hoverOptions.font;
+            var bBox = this.getBoundingClientRect();
+            var bBoxRef = gd.getBoundingClientRect();
+
+            Fx.loneHover({
```
etpinard: peek 2017-04-10 12-07

so this means the annotation hover takes precedence over trace hover. I don't think this a big deal. But perhaps we make annotation hover pass by `Fx.hover` so that they don't _conflict_ sort of speak.

src/components/annotations/draw.js
etpinard: > I only attached click and hover to the text box, not to the arrows

That's sounds like the desired behavior to me.

src/components/color/index.js
alexcjohnson: done in https://github.com/plotly/plotly.js/pull/1573/commits/919580e6d7f93f21718ee52ca033e2c72c9c5e33

src/components/annotations/draw.js
alexcjohnson: added a new attribute `captureevents` in https://github.com/plotly/plotly.js/pull/1573/commits/da1e9d70be9f51656e7ce6c67703fe8ae371e6fc

This one defaults to false (so data point hover comes through) unless `hovertext` is provided. But as noted in the description, if you use `plotly_clickannotation` events you must explicitly turn this on.

@cldougl once this is merged we should update the (currently broken) example in https://plot.ly/javascript/text-and-annotations/#styling-and-formatting-annotations, and add a section about `plotly_clickannotation` in https://plot.ly/javascript/plotlyjs-events/ that specifically mentions `captureevents`.

cldougl: got it-- just made an [issue](https://github.com/plotly/documentation/issues/742) to keep track. 
@alexcjohnson was there something specifically broken in https://plot.ly/javascript/text-and-annotations/#styling-and-formatting-annotations (just took a quick look and it seemed to work as expected) or should it just be updated to use `plotly_clickannotation`?

src/components/annotations/annotation_defaults.js
```diff
@@ -120,6 +120,7 @@ module.exports = function handleAnnotationDefaults(annIn, annOut, fullLayout, op
             color: hoverBorder
         });
     }
+    coerce('captureevents', !!hoverText);
```
etpinard: maybe `hoverText.length > 0`, as `!![] \\ => true`?

alexcjohnson: For annotations, `hovertext` is a string, I think this is OK. But you're absolutely right when this moves into traces and can be an array.

src/components/annotations/draw.js
alexcjohnson: Thanks @cldougl - I moved further discussion of this into the docs issue you made.

src/components/annotations/annotation_defaults.js
etpinard: Oh right. My bad.

etpinard:
Nicely done :dancer: `1.26.0` will be a good one.

rpaskowitz:
CI tests of master seem to be failing since this merge (local tests are ok for me):

```
Chrome 54.0.2840 (Linux 0.0.0) annotation effects should register clicks and show hover effects on the text box only FAILED
	Expected 0 to equal 1, '0 only (0)'.
	    at /tmp/tests/annotations_test.js:1084:0 <- /tmp/ba4b1756258f0bb2a60f96899bbad74d.browserify:196097:46
```

Merge build: https://circleci.com/gh/plotly/plotly.js/4087#tests/containers/1

Currently affecting #1588 and #1589

alexcjohnson:
> CI tests of master seem to be failing since this merge

Huh, thanks for the heads up @rpaskowitz. Not sure if this is an intermittent issue or a merge issue but I'll sort it out tonight.

etpinard:
> Not sure if this is an intermittent issue

It's intermittent. I've encountered this since Monday too.

Issue #1580, opened 2017-04-11
nestedProperty results in modal Plotly.restyle


rreusser:
See: http://codepen.io/rsreusser/pen/NjKXqJ?editors=0010

It seems to me that `Plotly.restyle` is modal in an undesirable way. That is, the same command succeeds or fails depending on the *value* of current data. Right now `value` of the filter transform is the only property which I'm aware exhibits this since either a number or an array is valid, but there may be others.

Consider:
1. `value` is a number
2. restyle `'value[0]'` ⟶ fails because you can't set property `[0]` of a number
3. restyle `'value': [0, 0]` ⟶ succeeds
4. restyle `'value[0]'` ⟶ succeeds

So the same command both fails and succeeds. You're welcome to close this, but it seems undesirable and uncommon, so I'd love to answer it definitively before it becomes a common usage that can't be undone. (Which is to say, carpet adds another usage of this and I'm currently struggling with the fallout.)

etpinard:
Interesting. Thanks for bringing this up @rreusser :beers: 

What does `relayout` do with e.g axis `range` and `domain`?

rreusser:
That case seems to work as expected. It seems like a bare number `n` gets interpreted as the array `[n, n + 1]`. In that case though, there's no confusion since the range itself can't be a bare number. The confusion here is due to the fact that `{operation: '<', value: 5}` expects a scalar value while `{operation: '[]', value: [4, 5]}` expects an array range.

rreusser:
One proposed fix: Sanitization should be done at the supplydefaults level. If nestedProperty complied and expanded values into an array when requested to set a particular index, _perhaps_ that would resolve this, but then maybe it would cause problems elsewhere. cc @alexcjohnson for ever-insightful API input.

That is, say `value` is 5. You try to set `'value[1]': 6`. First it could wrap 5, then assign your value. You'd get the result `[5, 6]`.

alexcjohnson:
My concern about this is that in general (including most `arrayOk` cases), `5` and `[5]` are not equivalent. When used for alongside a data array, `5` applies to all points whereas `[5]` applies to the first point and the rest receive some sort of default value. Even in this case, you wouldn't *just* change from `5` to `[5, 6]`, would you? You'd change `operation` as well, to `[]` or something, and you also can't use `5` with `[]` so they necessarily need to happen together.

[...]

PR #1581, opened 2017-04-11
Expand values into arrays as needed when setting with nestedProperty

rreusser:
See #1580 

This is just speculative. I don't necessarily think it's a good idea, but I'm submitting it to confirm whether it's actually a _bad_ idea. If the current value is not an array or an object and you try to set an array value, it wraps the current value in an array before setting. Unless I'm mistaken, this should only affect cases that currently throw an error.

### DIFF
--- a/src/lib/nested_property.js
+++ b/src/lib/nested_property.js
@@ -243,8 +243,13 @@ function checkNewContainer(container, part, nextPart, toDelete) {
     if(container[part] === undefined) {
         if(toDelete) return false;
 
-        if(typeof nextPart === 'number') container[part] = [];
-        else container[part] = {};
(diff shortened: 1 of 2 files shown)


rreusser:
The best I can come up with is one of:

1. this PR
2. restyle/relayout/nestedproperty need more knowledge about the schema so they can enable this _only_ for arrayOk attributes
3. we should disallow non-data attributes that fall into this pattern (setting data attributes with an accessor like this isn't a good use case anyway, but #1580 highlights why the difficulty is problematic for non-data attrs that need to get set in the ui)

rreusser:
Closing this as a discussion point but probably not the right solution.

Issue #102, opened 2015-12-10
Add attributes to customize hover text font


mheilman:
The font attribute doesn't affect the text in the "hoverlayer" div.  At the moment, it doesn't look like there's a way to specify that hover text font family.

Here's a minimal example.  The font for the axis labels gets changed, but not for the hover text.

```
<html>
<body>
  <div id="plotly-div"></div>
<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
<script>

var trace1 = { x: [0, 1], y: [4, 5] };
var trace2 = { x: [0, 1], y: [4, 5] };
var data = [trace1, trace2];
var layout = { "font": { "family": "Comic Sans MS" } };
Plotly.plot('plotly-div', data, layout);
</script>
</body>
</html>
```

Note:  I used Comic Sans MS because it's obvious, not because I actually want it in my chart :-)

Semi-related: The "font" property is mislabeled as "textfont" in the [docs](https://plot.ly/javascript/reference/#scatter-textfont-family)).  "textfont" doesn't appear to do anything.

etpinard:
@mheilman `textfont` isn't currently intended to affect hover text. Though, it would be a nice feature.

`textfont` works on text charts i.e. `scatter` trace with `mode: 'text'`. For example,

![image](https://raw.githubusercontent.com/plotly/plotly.js/master/test/image/baselines/text_chart_arrays.png)

mheilman:
Ah, OK, thanks for the clarification on `textfont`.

It'd be nice to be able to set the hover text because I don't think it's possible now to have a consistent font in a scatter plot other than with the default font.  I think CSS would just get overridden because the font is set explicitly in the tag using the `style` attribute, and one can't edit that `style` attribute with, e.g., `querySelectorAll(".hovertext text").forEach(...)` because those text elements are only added when the user hovers over them.

I could be missing something, though.  Maybe there's some way to change the mouse over event, but that seems complicated.

Thanks.

etpinard:
@mheilman you could hook on a hover event handler ([example](https://plot.ly/javascript/hover-events/)).

Something like

```js
graphDiv.on('plotly_hover', function(data) {
  querySelectorAll(".hovertext text").forEach( /* ... */);
// ...
});
```

But, yeah, there's no way customize the font of the hover labels in the plot API at the moment.

I'm renaming this issue as a feature request.

mheilman:
Ah, OK.  Thanks again.

[...]

PR #1582, opened 2017-04-11
Custom trace hover labels

etpinard:
resolves https://github.com/plotly/plotly.js/issues/102

This PR builds on top of https://github.com/plotly/plotly.js/pull/1573 where annotation hover labels were added with support for custom bg color, border color and font settings. Trace hover labels will be customisable using the same attributes as for annotations, but unlike for annotations, these attributes will be `arrayOk: true`. For example,

peek 2017-04-11 14-14

----

Questions: 

- [x] should we add `layout.?axis.hoverlabel` attribute containers to make the the hover label that appear on the x(y) axes on `hovermode: 'x'('y')` customisable? **UPDATE:** commits https://github.com/plotly/plotly.js/pull/1582/commits/f986fd2e34afa4fe625a995831888899d9b2442b and https://github.com/plotly/plotly.js/pull/1582/commits/a8cc8b03efd2934265c1ed1c76092776c1596953 implement `layout.hoverlabel`
- [x] some attribute declarations and logic are duplicated across annotations and trace modules, maybe we should add a new component in e.g. in `src/component/hover/`? This new component would include `graph_interact.js` which is becoming less and less a _cartesian_ module. **Done** in https://github.com/plotly/plotly.js/pull/1613

TODO:

- [x] make sure this works for all trace types
  + e.g. fill calcdata accordingly `heatmap/calc.js` to support 2D data arrays
  + add handlers for trace type that don't use `Fx.hover` (e.g. gl3d. gl2d) on hover


etpinard:
cc @alexcjohnson @n-riesco

alexcjohnson:
> should we add `layout.?axis.hoverlabel` attribute containers to make the the hover label that appear on the x(y) axes on `hovermode: 'x'('y')` customisable?

My gut reaction is to make `layout.hoverlabel`, ie a default label style for the whole plot, which would set defaults for *all* the hover labels, including the new annotation labels, but the colors would only be used (as a default) by the common (axis) labels. It's true that this would preclude different styling for the different axes on which these common labels might appear, but offhand that seems like a fairly obscure use case to me, whereas I can definitely see people wanting a concise way to tailor the style of all the labels across the whole chart.

> maybe we should add a new component in e.g. in `src/component/hover/`? This new component would include `graph_interact.js` which is becoming less and less a _cartesian_ module.

Absolutely. Dunno if now is the time to do it, may be a bit of a project, but if you're excited about it now then by all means go for it.

n-riesco:
> should we add `layout.?axis.hoverlabel` attribute containers to make the the hover label that appear on the x(y) axes on hovermode: 'x'('y') customisable?

I could understand some users wanting the data and the axis labels have the same style.
Neither `layout.hoverlabel` nor `layout.?axis.hoverlabel` would help with that.

I can see a case for `layout.?axis.hoverlabel` in bar charts (so +1 from me).

Also +1 from for `layout.hoverlabel`, because, as a user, if I can style data labels using `data[n].hoverlabel`, I'd also expect to be able to style axis labels.

> `add a new component in e.g. in src/component/hover/`

What Alex said (+1 from me too)

etpinard:
OK. I'll add `layout.hoverlabel` in this PR - where it will set the default hover label settings for trace, common and annotations hover labels (similar to how `layout.font` impacts other font attributes). For example,

```js
layout: {
  hoverlabel: {
    bgcolor: 'white',
    font: {
      size: 20,
      color: 'black'
    }
  }
}

data = [{
  y: [/* ... */],
  hoverlabel: {
     font: { 
       color: 'biue'
     }
  }
}]
```

would make all hover labels have a white bg, common labels will have black text and trace labels blue text. Sounds ok?

Then, if someone really wants per-axis common label customizable down the road, we can always add `?axis.hoverlabel` later.

n-riesco:
> would make all hover labels have a white bg, common labels will have black text and trace labels blue text. Sounds ok?

+1 to white bg
I'd suggest black text as a safer default for all types of labels.

alexcjohnson:
> would make all hover labels have a white bg, common labels will have black text and trace labels blue text. Sounds ok?

I was thinking the colors in `layout.hoverlabel` would only apply to common labels and annotation labels, since the colors in trace labels pull their defaults from a different route and this serves an important purpose disambiguating traces. But other attributes (font family & size, border width) would propagate to traces.

But @n-riesco raises a good point:
> I could understand some users wanting the data and the axis labels have the same style.

That would be annoying the way I'm suggesting, as it would require explicitly coloring all the trace labels. I suppose we could make some boolean attribute like `layout.hoverlabel.applycolorstotraces` for this case, and have it default to `false`? Dunno, maybe that's too much, but it would have the advantage that you can say "oh, see, the trace labels get their own colors because applycolorstotraces is false."

> Then, if someone really wants per-axis common label customizable down the road, we can always add `?axis.hoverlabel` later.

👍

n-riesco:
@alexcjohnson @etpinard I really misunderstood `layout.?axis.hoverlabel`. Please, ignore my comments about it. 

Please, correct me if I'm mistaken. This is how understand `hoverlabel` now:
- `data[n].hoverlabel` is `arrayOk: true` and sets the style of trace hover labels (I'll call these labels *data hover labels*)
- `layout.?axis.hoverlabel` is `arrayOk: false` and would set the style of the hover labels displayed over the `?axis` (I'll call these labels *axis hover labels*).
- `layout.hoverlabel` is `arrayOk: false` and would set the style of any hover labels.

> I was thinking the colors in layout.hoverlabel would only apply to common labels and annotation labels, since the colors in trace labels pull their defaults from a different route and this serves an important purpose disambiguating traces. But other attributes (font family & size, border width) would propagate to traces.

Oh, I missed that at the moment *data hover labels* are coloured after the marker!

> I could understand some users wanting the data and the axis labels have the same style.

When I wrote that, what I had in mind was a bar chart (each bar with different colours). Currently bar charts have *axis hover labels* with a black background, whereas *data hover labels* are coloured after the bar. The use case I had in mind was that both *axis* and *data hover labels* were coloured after the bar.

How about having `layout.hoverlabel.autocolor` to support this use case?
When `layout.hoverlabel.autocolor` is `true` any hover labels (axis, annotations) will use the same colour as the *data hover label* (i.e by default the marker colour, unless `data[n].hoverlabel` is defined).

alexcjohnson:
> How about having `layout.hoverlabel.autocolor` to support this use case?
> When `layout.hoverlabel.autocolor` is `true` any hover labels (axis, annotations) will use the same colour as the *data hover label* (i.e by default the marker colour, unless `data[n].hoverlabel` is defined).

When there's a common label it's because you're in a hover mode (ie compare) that supports multiple labels... so this wouldn't be meaningful in general.

tasks/test_syntax.js
```diff
@@ -177,7 +177,7 @@ function assertCircularDeps() {
         var logs = [];
 
         // see https://github.com/plotly/plotly.js/milestone/9
-        var MAX_ALLOWED_CIRCULAR_DEPS = 12;
+        var MAX_ALLOWED_CIRCULAR_DEPS = 17;
```
etpinard: I got a little too excited in https://github.com/plotly/plotly.js/pull/1613, that PR didn't remove any circular deps. It only made `madge` think that there were less ciruclar deps (their algo must order-dependent).

If you don't believe me, try adding `var Fx = require('../fx');` in `annotations_defaults` like [here](https://github.com/plotly/plotly.js/pull/1582/commits/f986fd2e34afa4fe625a995831888899d9b2442b#diff-81cae611e22b976932e957a99623ac1eL14) and you get back the 12 circular deps of PR #1613 .

etpinard:
@alexcjohnson this PR is ready for review. It's probably best to review it commit by commit after the #1613 merge commit.

To note, `cartesian`, `geo`, `ternary` and `mapbox` subplot use `Fx.hover` while `pie`, `gl3d` and `gl2d` use `Fx.loneHover`. A lot of redundant logic could be :hocho: by making a more general hover routine that all subplot types could call on even grounds. Moreover, making `gl3d` and `gl2d` use `gd.calcdata` would also help DRY things up :palm_tree: Oh well, at least our hover test suites are becoming pretty solid :lock: .

mateuscavaliere:
@etpinard is there a especific way to declare this hoverlabel options in plotly for R? I've tried as a list inside layout option but nothing happens

cpsievert:
@mateuscavaliere the R package only supports [official releases](https://github.com/plotly/plotly.js/releases) of plotly.js. You will have to wait until this feature (which is under development) is included in a future release.

mateuscavaliere:
@cpsievert I thought it was released in v1.26.0

etpinard:
@mateuscavaliere I'm gonna go ahead and delete your comments. This PR is still under review comments of the likes make things a little annoying for reviewers. Please use our https://community.plot.ly/ forums if you have other questions.

src/components/fx/helpers.js
```diff
@@ -0,0 +1,81 @@
...
+var constants = require('./constants');
+
+// look for either subplot or xaxis and yaxis attributes
+exports.getSubplot = function getSubplot(trace) {
+    return trace.subplot || (trace.xaxis + trace.yaxis) || trace.geo;
+};
+
+// convenience functions for mapping all relevant axes
+exports.flat = function flat(subplots, v) {
+    var out = [];
```
alexcjohnson: faster (here and in `p2c` below) to do `var out = new Array(subplots.length)` and avoid `push`?

src/components/fx/hover.js
```diff
@@ -577,9 +579,17 @@ function createHoverText(hoverData, opts) {
             ltext = label.selectAll('text').data([0]);
 
         lpath.enter().append('path')
-            .style({fill: Color.defaultLine, 'stroke-width': '1px', stroke: Color.background});
+            .style({
+                fill: commonLabelOpts.bgcolor || Color.defaultLine,
+                stroke: commonLabelOpts.bordercolor || Color.background,
```
alexcjohnson: might want to be `Color.fill` and `Color.stroke` since these I guess support alpha now

src/plots/gl2d/scene2d.js
```diff
@@ -667,3 +675,18 @@ proto.hoverFormatter = function(axisName, val) {
     var axis = this[axisName];
     return Axes.tickText(axis, axis.c2l(val), 'hover').text;
 };
+
+proto.castHoverOption = function(trace, ptNumber, attr) {
```
alexcjohnson: possible to share this with the one in [gl3d](https://github.com/plotly/plotly.js/pull/1582/commits/6bbfa652b4da40f6c060eb4ef72c97dbea98e253)? Looks nearly identical?

src/components/fx/hover.js
alexcjohnson: I started looking around at other places this might be needed... and there are a bunch potentially (seems like all in this file), including some in spikelines that I didn't catch when reviewing that. I'm not sure if it's really crucial, the distinction is mainly important for export right? And you're not going to export hover effects. Still, best to be consistent.

src/plots/gl2d/scene2d.js
etpinard: It is possible. I didn't do it because as brought up in https://github.com/plotly/plotly.js/pull/1582#issuecomment-297115980, the _real_ solution would be to replace `Fx.loneHover` by something more similar to `Fx.hover` in gl2d and gl3d. But, for the time being, might as well put this logic something in the `Fx` module object.

etpinard: done in 60bd4fc

src/components/fx/hover.js
etpinard: After some thought, I think it would be best to **not** render the alpha channel in all hover labels, as the marker hover labels already [blend](https://github.com/plotly/plotly.js/pull/1582/files#diff-53eaf7f3ac3274b40206f89f8f87a5d6R691) marker opacities with `plot_bgcolor` filtering out the alpha channel. For example,

```js
var trace = {
  x: [1],
  y: [1],
  marker: {
    color: ['rgba(255, 0, 0, 0.3)']
  }
}

var layout = {
  plot_bgcolor: 'grey'
}
```

renders the marker hover label as `rgb(166, 90, 90)` i.e. with no alpha channel.

We could add some special logic so that e.g.

```js
var trace = {
  x: [1],
  y: [1],
  marker: {
    color: ['rgba(255, 0, 0, 0.3)']
  },
  hovelabel: {
    bgcolor: 'rgba(0,0,255,0.3)'
  }
}

var layout = {
  plot_bgcolor: 'grey'
}
```

renders the alpha channel overriding the marker color to hover label color behavior, but quite frankly I don't see the need.

Adding a line in the schema description saying that alpha channel aren't honored sounds sufficient for now. Thoughts?

src/components/fx/helpers.js
etpinard: done in 	e45bff9

src/components/fx/hover.js
alexcjohnson: I've occasionally thought it would be nice to be able to see behind the hover labels - not just behind the name field like you can do now, but for example in a [fractal](http://codepen.io/alexcjohnson/full/KwQWYo/) when I want to draw a small zoom box it would be cool if the whole label were partially transparent. But in cases like that, seems like the *best* solution would be to make another attribute - `hoverlabel.opacity`, akin to `marker.opacity` so you don't need to set the background, the font color, and the border color all to rgba values - in fact, doing it that way would be a substandard experience for exactly the same reason as we have a `marker.opacity` in the first place, that you really want opacity applied to the whole group.

So yeah, I guess I'm agreeing with you that these attributes needn't support opacity on their own. If you feel like putting in `hoverlabel.opacity` now go for it, but it can also be left out of this PR and done later.

etpinard: >  but it can also be left out of this PR and done later.

I'm going to go with that if you don't mind. This PR is big enough.

src/traces/sankey/plot.js
```diff
@@ -124,7 +124,8 @@ module.exports = function plot(gd, calcData) {
     };
 
     var linkHoverFollow = function(element, d) {
-
+        var trace = gd._fullData[d.traceId];
+        var ptNumber = d.originalIndex;
```
etpinard: @monfera does this look ok to you? I'm looking for the data array index corresponding to this link / node (below).

Oh wait. `node` and `link` can have different lengths, correct? So maybe `hoverlabel` should be set per-node and per-link e.g.:

```js
var trace = {
  node: {
     hoverlabel: {
        bgcolor: [/* same length as node data arrays */]
     },
     link: {
       hoverlabel: {
         bgcolor: [/* same length as link data arrays */]
       }
     }
   },
   hoverlabel: {
     // for trace-wide hover label settings (that would be applied to all nodes and links)
   }
}
```

Well the above sounds uselessly complicated for v1 of sankey. So, I'd vote for not allowing array values in `hoverlabel` attribute of sankey traces. Does that sound ok?

monfera: Yes your conclusion sounds good to me. Indeed it'll be an unlikely occurrence that the number of nodes and links match.

etpinard: :white_check_mark: in 	e50886

:lock: in f8258cb

alexcjohnson: We can leave it off for now to get 1.27 out, but we should definitely include custom hovertext (and at that point we might as well include custom styling) for both nodes and links in the near future - I can definitely see people wanting to explain what specifically is included in each of these elements in more detail than is displayed permanently on the diagram. Chatting with @monfera seems like `link.label` fills this role for links but there's no analog for nodes. I'll make an issue for this so we don't forget about it.

etpinard: > I'll make an issue for this so we don't forget about it.

Thanks!

alexcjohnson: https://github.com/plotly/plotly.js/issues/1663

src/components/fx/hover.js
alexcjohnson: https://github.com/plotly/plotly.js/issues/1665

test/jasmine/tests/sankey_test.js
```diff
@@ -260,6 +260,12 @@ describe('sankey tests', function() {
 
         afterEach(destroyGraphDiv);
 
+        function wait() {
+            return new Promise(function(resolve) {
+                setTimeout(resolve, 60);
+            });
+        }
```
alexcjohnson: https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/delay.js ?

test/jasmine/tests/sankey_test.js
```diff
@@ -313,30 +319,23 @@ describe('sankey tests', function() {
...
-                    window.setTimeout(function() {
-                        expect(d3.select('.hoverlayer>.hovertext>text').node().innerHTML)
-                            .toEqual('447TWh', 'tooltip present');
-
-                        mouseEvent('mousemove', 450, 300);
-                        mouseEvent('mouseover', 450, 300);
-
-                        window.setTimeout(function() {
-                            expect(d3.select('.hoverlayer>.hovertext>text').node().innerHTML)
-                                .toEqual('46TWh', 'tooltip jumped to link');
```
alexcjohnson: this `expect` got lost?

etpinard: it's there: https://github.com/plotly/plotly.js/pull/1582/files#diff-6c72cf011d0c1bbd8d91d759790188aeR330

test/jasmine/tests/sankey_test.js
etpinard: That thing got :hocho: 

I've been using this [pattern](https://github.com/plotly/plotly.js/pull/1582/files#diff-846cf5b534aa0d22bdd1da2b43ac3cbaR453) lately. No need for timeouts :tada:

alexcjohnson: Ah nice 🚀

test/jasmine/tests/sankey_test.js
alexcjohnson: 👍  sorry, missed that!

alexcjohnson:
💃  once tests pass. Epic PR @etpinard 🏆

PR #1584, opened 2017-04-11
Enable compress attributes browserify transform by default

etpinard:
as discussed in https://github.com/rreusser/plotly-mock-viewer/issues/5#issuecomment-292581169, we should really be compressing our attribute declarations in **all** browserify builds by default - to make sure that customs bundles don't include (useless) attribute description strings. 

Note that, using the `ignoreTransform` browserify [options](https://github.com/substack/node-browserify#browserifyfiles--opts) with can still produce `dist/plotly-with-meta.js` as before.

TODO:

- [x] In this PR, I'm referring to the `compress_attributes.js` transform using a relative (UNIX) path. We should make sure that this works also on Windows machines before merging. If that doesn't work we can always publish `compress_attributes.js` on npm and refer to it as a dependency in `node_modules/` like we currently do for `glslify`.
- [x] move the `plotschema_test.js` suite to `bundle_tests/` and make it use `plotly-with-meta.js`

cc @rreusser


rreusser:
Apologies I missed in the first place, but I wholeheartedly support this. The only exception would be if it makes it difficult to compile *with* metadata. So my question is: how do you compile *with* metadata when this is present?

rreusser:
Oh, just saw linked comment:

> `--ignore-transform=MODULE`

tasks/util/browserify_wrapper.js
```diff
@@ -38,7 +37,10 @@ module.exports = function _bundle(pathToIndex, pathToBundle, opts) {
     var browserifyOpts = {};
     browserifyOpts.standalone = opts.standalone;
     browserifyOpts.debug = opts.debug;
-    browserifyOpts.transform = outputMinified ? [compressAttributes] : [];
+
+    if(opts.noCompress) {
+        browserifyOpts.ignoreTransform = './tasks/compress_attributes.js';
```
rreusser: 👏 This seems better by default, and `tasks/` seems like a better location than `tasks/util`

tasks/util/watchified_bundle.js
```diff
@@ -6,7 +6,6 @@ var prettySize = require('prettysize');
 
 var constants = require('./constants');
 var common = require('./common');
-var compressAttributes = require('./compress_attributes');
```
rreusser: Nice that this just disappears. Opt-opt also scopes the transform better so it doesn't need to be applied globally 👍

tasks/bundle.js
```diff
@@ -41,7 +41,8 @@ _bundle(constants.pathToPlotlyGeoAssetsSrc, constants.pathToPlotlyGeoAssetsDist,
 // Browserify the plotly.js with meta
 _bundle(constants.pathToPlotlyIndex, constants.pathToPlotlyDistWithMeta, {
     standalone: 'Plotly',
-    debug: DEV
+    debug: DEV,
+    noCompress: true
```
rreusser: 👏

rreusser:
Apologies I didn't leave more meaningful feedback the first time around. This definitely seems like a good idea. It's really the exception rather than the norm that you'd even want the extra metadata. It looks like it needs a quick refresh to resolve a conflict or two, but assuming that's just trivial merging, it gets the 💃 from me.

etpinard:
Can anyone with a window laptop try running `npm run build` off this branch? I'm a little worried about this [line](https://github.com/plotly/plotly.js/pull/1584/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R53) which uses a UNIX-style path.

Maybe @dfcreative could help.

etpinard:
Ping @antoinerg (who has access to a Windows "dev" :computer: )

Could you pull down this branch and run `npm run build` to check that the added UNIX-style path to a browserify transform in our package.json works ok on Windows?

To check that things work:

- `npm run build`, then
- `grep "description:" dist/plotly.js` (or some windows equivalent) should output:

```
    return Lib.extendFlat({}, attributes, { description: description });
 *     extra.description: extra description. N.B we use
 *     opts.description: where & how this font is used
```

i.e. only three occurrences of `description:`

- and `grep "description:" dist/plotly-with-meta.js | wc -l` should be 973 (i.e. 973 lines with `description:` in it)

Thanks!

etpinard:
Fantastic!

Attribute meta will be stripped out of bundles (made with bundlers that support browserify transforms) by default starting in `v1.41.0`!

PR #1585, opened 2017-04-11
Misc performance optimization

hy9be:
Following my attempts in https://github.com/plotly/plotly.js/pull/1544, https://github.com/plotly/plotly.js/pull/1555, and https://github.com/plotly/plotly.js/pull/1557, I did some work to further improve the performance of Plotly. I scan through timeline results I captured on slow charts, which has 200+ traces, and did some caching. I also tried to skip processing on some components (e.g. colorbar) when it is enabled for a graph. 

By doing all these, I observed about 15% improvements on bar chart and marginal improvements on scatter plot.

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -504,6 +504,7 @@ drawing.makeTester = function(gd) {
 // always returns a copy of the bbox, so the caller can modify it safely
 var savedBBoxes = [],
     maxSavedBBoxes = 10000;
+
 drawing.bBox = function(node) {
     // cache elements we've already measured so we don't have to
     // remeasure the same thing many times
@@ -512,12 +513,14 @@ drawing.bBox = function(node) {
         return Lib.extendFlat({}, savedBBoxes[saveNum.value]);
     }
 
-    var test3 = d3.select('#js-plotly-tester'),
-        tester = test3.node();
+    if(!drawing.test3) {
+        drawing.test3 = d3.select('#js-plotly-tester');
+        drawing.tester = drawing.test3.node();
+    }
 
     // copy the node to test into the tester
     var testNode = node.cloneNode(true);
-    tester.appendChild(testNode);
+    drawing.tester.appendChild(testNode);
     // standardize its position... do we really want to do this?
     d3.select(testNode).attr({
         x: 0,
@@ -525,19 +528,21 @@ drawing.bBox = function(node) {
         transform: ''
     });
 
-    var testRect = testNode.getBoundingClientRect(),
-        refRect = test3.select('.js-reference-point')
+    var testRect = testNode.getBoundingClientRect();
+    if(!drawing.refRect) {
+        drawing.refRect = drawing.test3.select('.js-reference-point')
             .node().getBoundingClientRect();
+    }
 
-    tester.removeChild(testNode);
+    drawing.tester.removeChild(testNode);
 
     var bb = {
         height: testRect.height,
         width: testRect.width,
-        left: testRect.left - refRect.left,
-        top: testRect.top - refRect.top,
-        right: testRect.right - refRect.left,
-        bottom: testRect.bottom - refRect.top
+        left: testRect.left - drawing.refRect.left,
+        top: testRect.top - drawing.refRect.top,
+        right: testRect.right - drawing.refRect.left,
+        bottom: testRect.bottom - drawing.refRect.top
     };
 
     // make sure we don't have too many saved boxes,
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -282,7 +282,8 @@ Plotly.plot = function(gd, data, layout, config) {
     // Now plot the data
     function drawData() {
         var calcdata = gd.calcdata,
-            i;
+            i,
+            rangesliderContainers = fullLayout._infolayer.selectAll('g.rangeslider-container');
 
         // in case of traces that were heatmaps or contour maps
         // previously, remove them and their colorbars explicitly
@@ -302,7 +303,7 @@ Plotly.plot = function(gd, data, layout, config) {
                     .selectAll(query)
                     .remove();
 
-                fullLayout._infolayer.selectAll('g.rangeslider-container')
+                rangesliderContainers
                     .selectAll(query)
                     .remove();
             }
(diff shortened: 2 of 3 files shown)


hy9be:
The way I cache those objects is not ideal. Please suggest the desired way to cache objects, during the review. Thanks!

hy9be:
I skipped the colorbar layout step when colorbar is not enabled for a trace: https://github.com/plotly/plotly.js/pull/1585/commits/f4196f9420366961b532d57e1188711f611d6faa. But the change breaks 4 unit test cases so I decide to roll it back for now. Maybe I misunderstood the necessity of having colorbar layouted during calling ``marginPushers``, even colorbar is not enabled for the plot.

src/plot_api/plot_api.js
```diff
@@ -302,7 +303,7 @@ Plotly.plot = function(gd, data, layout, config) {
                     .selectAll(query)
                     .remove();
 
-                fullLayout._infolayer.selectAll('g.rangeslider-container')
+                rangesliderContainers
```
etpinard: Interesting. How does this result in a perf improvement?

src/components/drawing/index.js
```diff
@@ -512,32 +513,36 @@ drawing.bBox = function(node) {
         return Lib.extendFlat({}, savedBBoxes[saveNum.value]);
     }
 
-    var test3 = d3.select('#js-plotly-tester'),
```
etpinard: Nicely done. There's only one tester div per page, so might as well cache in the `Drawing` module object. :+1:

src/plot_api/plot_api.js
```diff
@@ -205,7 +205,7 @@ Plotly.plot = function(gd, data, layout, config) {
             if(trace.visible !== true || !trace._module.colorbar) {
                 Plots.autoMargin(gd, 'cb' + trace.uid);
             }
-            else trace._module.colorbar(gd, cd);
+            else if(trace.marker.colorbar) trace._module.colorbar(gd, cd);
```
etpinard: This new `if()` makes traces that have root-level `colorbar` settings (e.g. `heatmap`, `contour`, `surface`) bypass the colorbar draw step. 

Something like

```js
else if(Plots.traceIs(trace, 'markerColorscale') && Colorbar.hasColorbar(trace)) {
  trace._module.colorbar(gd, cd);
}
```

might make things work.

etpinard:
Thanks (again) @hy9be for these perf improvements :+1:

src/components/drawing/index.js
hy9be: Not sure if I understand correctly. I did cache the tester div in line 517 and 518 already. Is that what you are referring to?

src/plot_api/plot_api.js
hy9be: I made the changes and run test locally. It still fails in the contour/heatmap traces assertion on colorbar, https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/plot_interact_test.js#L223 and https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/plot_interact_test.js#L283. Both ``countColorBars`` function returns 0. 

I will take a look. And it would be great if you could shed some lights here.

src/plot_api/plot_api.js
hy9be: Did not tested separately. Just checked and my tests shows 200ms to 400ms improvement on 200 traces.

When I profile the 200-trace case, the ``select`` and ``selectAll`` are among top 5 of hot functions. So my thought was caching all the constant selected elements in data loops.

etpinard: Right. I understand the value of caching selections, but here you're just assigning the selection to a variable: `.selectAll` is still called inside a loop.

src/components/drawing/index.js
etpinard: I'm saying this is a great commit.

hy9be: Got used to expect modification suggestions in a code review lol. Thanks!

src/plot_api/plot_api.js
hy9be: ``.selectAll('g.rangeslider-container')`` is moved out of the loop, maybe that's where the save comes from.

src/plot_api/plot_api.js
hy9be: I did not find any problems and tried again today, the local test passed this time. Not sure why it failed last time. So I pushed the commit you suggested.

hy9be:
I fetch and merged the latest master and all tests passed locally, but CI still has couple of cases failed. Not sure why.

hy9be:
@etpinard I fixed all jasmine test cases but the CI failed at some other commands which I do not know how to debug. Could you please give me some hints here? Basically I do not really understand the error reported by CI...Thanks!

etpinard:
`npm run test-image-gl2d` is failing. 

I think adding `markerColorscale` to this [list](https://github.com/plotly/plotly.js/blob/6255f6ddfd5f31596e8ff5165b3325c103741acf/src/traces/parcoords/index.js#L22) will fix it. 

Or perhaps, replacing

```js
else if(Plots.traceIs(trace, 'markerColorscale') || Colorbar.hasColorbar(trace)) {}
```

with 

```js
else if(Plots.traceIs(trace, 'markerColorscale') || Plots.traceIs(trace, 'parcoords') || Colorbar.hasColorbar(trace)) {}
```

hy9be:
@etpinard I tried both ways you suggested and did some further investigation, still cannot figure out a way to pass all the image passed.
Considering the amount of improvements we can get from skipping the colorbar layout, and the effort spent and will be spend on this, I am fine with rollbacking this colorbar change and just commit the caching of selections in this PR. Does it sound ok for you guys?

hy9be:
@etpinard I removed the optimization for colorbar. Now all tests passed. Just let you know.

src/plot_api/plot_api.js
etpinard: Ha. Now I see it (it isn't part of the GH diff view). Thanks :+1:

etpinard:
@hy9be looks like this PR is ready to be merged. But before, since you chose to not include that colorbar optimisation, can you rebase and remove all the colorbar-related commits locally and then `git push -f` here?

hy9be:
@etpinard I removed all colorbar related commits. Please see if that looks good. Thanks!

PR #1588, opened 2017-04-12
Passthrough the wheel event if the scrollbar is at the very top or bottom

rpaskowitz:
On a page with several charts, if you happen to have the cursor such that it winds up over a legend with scrollbar as you scroll the page, the scrollbar scrolls, but then captures all further `wheel` events. This change makes it so that the event is passed through to the page if the scrollbar is at the very top or bottom of the box so that the rest of the page can scroll.


etpinard:
Makes sense. Thanks very much!

PR #1590, opened 2017-04-12
add attributes xshift, yshift to annotations

alexcjohnson:
Adds two new attributes to annotations, `xshift` and `yshift`.

Originally I thought I'd accomplish this by just allowing `standoff` to be a 2-vector, but I decided against this for a few reasons:
- matches the rest of our attributes in annotations better, like `ax` and `ay`
- it should have an effect even when there's no arrow
- related to ^^, I felt like this one, particularly for the use cases I expect like pointing a new annotation at an existing one or at a colorbar or something, would be more natural if it moved both ends of the arrow (and the text box) so it works fairly differently from `standoff`
- didn't want to invite https://github.com/plotly/plotly.js/issues/1580 again


src/components/annotations/draw.js
```diff
@@ -524,21 +532,17 @@ function drawOne(gd, index) {
 
                         update[annbase + '.x'] = xa ?
                             xa.p2r(xa.r2p(options.x) + dx) :
-                            ((headX + dx - gs.l) / gs.w);
+                            (options.x + (dx / gs.w));
```
etpinard: Nice simplification :+1:

test/jasmine/tests/annotations_test.js
```diff
@@ -1127,7 +1132,7 @@ describe('annotation effects', function() {
         }
 
         makePlot([
-            {x: 50, y: 50, text: 'hi', width: 50, ax: 0, ay: -40},
+            {x: 50, y: 50, text: 'hi', width: 50, ax: 0, ay: -40, xshift: -50, yshift: 50},
```
etpinard: Is `editable: true` tested with xshift / yshift? It's a little hard to tell from the diff.

test/jasmine/tests/annotations_test.js
```diff
@@ -554,9 +554,9 @@ describe('annotations autorange', function() {
     it('should adapt to relayout calls', function(done) {
         Plotly.plot(gd, mock).then(function() {
             assertRanges(
-                [0.97, 2.03], [0.97, 2.03],
-                ['2000-10-01 08:23:18.0583', '2001-06-05 19:20:23.301'], [-0.245, 4.245],
-                [0.9, 2.1], [0.86, 2.14]
+                [0.91, 2.09], [0.91, 2.09],
```
etpinard: so these values changed because tha `annotations-autorange` mock changed?

src/components/annotations/draw.js
alexcjohnson: motivated by the fact that the old way would have needed to explicitly account for shift - notice though in the [`showarrow: false` case](https://github.com/plotly/plotly.js/pull/1590/files#diff-05274991eb85aefd482f72514dbda1a5R601) I did need to just include shift, due to the auto-anchor business...

test/jasmine/tests/annotations_test.js
alexcjohnson: yes, all the other tests above this in `describe('annotation effects'`, to which I added `xshift: 5, yshift: 5`, are editable and failed before I fixed dragging with shifts.

test/jasmine/tests/annotations_test.js
alexcjohnson: > so these values changed because the `annotations-autorange` mock changed?

correct. I guess that's a hazard of reusing mocks for both images and jasmine tests...

etpinard:
All right. Nicely done :dancer:

PR #1593, opened 2017-04-12
Note test fix

alexcjohnson:
Pretty sure this fixes the intermittent test failure https://github.com/plotly/plotly.js/pull/1573#issuecomment-293684424

There may be other ways to do it (this adds ~1.5 sec to the test), but the main thing I think is avoiding an accidental doubleclick (and the associated autorange, which moves the annotations) by putting a delay after each click.

Tested 20x locally, which was enough to generally cause failure before...

cc @etpinard @rpaskowitz


etpinard:
:dancer: thanks!

PR #1595, opened 2017-04-13
Carpet plot rebase

rreusser:
This PR breaks up a couple not-really-related or loosely-related commits from carpet plots into separate commits and rebases all of carpet plots into one big commit. Supersedes #1239.


src/plots/gl3d/layout/axis_attributes.js
```diff
@@ -14,6 +14,7 @@ var extendFlat = require('../../../lib/extend').extendFlat;
 
 
 module.exports = {
+    visible: axesAttrs.visible,
```
etpinard: `visible` for 3D axes doesn't actually works, correct?

I think you _had_ to put it in there so that `supplyDefaults` doesn't error out.

etpinard:
PR of the year candidate :trophy:

rreusser:
giphy_360

src/plots/gl3d/layout/axis_attributes.js
rreusser: Yeah, correct. Was contemplating the best option as was working back through that. Correct though, it was just to make them *not fail*.

src/lib/ensure_array.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * The goal is to just reuse memory to avoid a bit of excessive garbage
+ * collection.
+ */
+module.exports = function ensureArray(out, n) {
+    if(!Array.isArray(out)) out = [];
+
+    // If too long, truncate. (If too short, it will grow
+    // automatically so we don't care about that case)
+    out.length = n;
```
alexcjohnson: huh, never knew you could resize this way!

src/plots/gl3d/layout/axis_attributes.js
rreusser: Perhaps the right answer is not to remove the attribute and avoid supplying that defaults in that case.

etpinard: `scene.?axis.visible: false` would be a nice option too. So I don't think it's worth changing the logic. Opening a new issue about implementing  `?axis.visible` in 3D will suffice.

src/lib/ensure_array.js
rreusser: Best I can tell, it's actually robust and supported across platforms. It's possible it does weird things internally (like not actually de-allocating that memory. I saw a while ago some discussion of how strings work in js. Hint: it's not straightforward). The main goal was to just overwrite the carpet spline stuff rather than allocating everything from scratch every time since that adds up to lots of garbage collection.

etpinard: From https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/length

> The array length property sets or returns the number of elements in an array.

That's right, [..] **sets** or returns [..].

src/components/drawing/index.js
```diff
@@ -678,3 +678,13 @@ drawing.setPointGroupScale = function(selection, x, y) {
...
+
+drawing.measureText = function(tester, text, font) {
+    var dummyText = tester.append('text')
+        .text(text)
+        .call(drawing.font, font);
+
+    var bbox = drawing.bBox(dummyText.node());
+    dummyText.remove();
+    return bbox;
+};
```
alexcjohnson: I see why you did it this way, but it seems a little redundant, we're making a new node only to clone it in the global tester, then we delete them both. If this ends up getting called a lot, for performance we may want to break up `drawing.bBox` in such a way that this element can get created in the global tester in the first place.

rreusser: That makes sense. I didn't appreciate that the tester was also cloning it. And to add to that, it looks like I only ended up using it in one place. :/

rreusser: (but on the plus side, to answer your question, at least it doesn't end up getting called a lot 😄 )

test/jasmine/tests/carpet_test.js
```diff
@@ -0,0 +1,432 @@
...
+
+        expect(traceOut.a).toEqual([0, 1]);
+        expect(traceOut.b).toEqual([0, 1, 2]);
+        expect(traceOut.x).toEqual([[1, 2, 3], [4, 5, 6]]);
+        expect(traceOut.y).toEqual([[2, 3, 4], [5, 6, 7]]);
+        expect(traceOut.da).toBeUndefined();
+        expect(traceOut.db).toBeUndefined();
+        expect(traceOut.a0).toBeUndefined();
+        expect(traceOut.b0).toBeUndefined();
+        expect(traceOut.visible).toBe(true);
```
etpinard: @rreusser looks like only this assertion fails when removing the `if(!test.supplyDefaults) return;` statement above.

etpinard: by the way @rreusser I'm adding a few `restyle` test cases on a branch, so I can take care of :arrow_heading_up: if you like.

src/lib/ensure_array.js
rreusser: (best-case scenario though is that it actually truncates in-place without actually copying. I mean _maybe_ copies, but `slice()` _definitely_ copies.)

test/jasmine/tests/carpet_test.js
rreusser: If you're already editing it, then 👍 . I think most of the carpet tests should actually be moved out of this file since some of the functions are better suited for lib.

etpinard: done in https://github.com/plotly/plotly.js/pull/1596/commits/c98553a9f14b32ede6afd79ca5a86a3793d72d9c

src/plots/gl3d/layout/axis_attributes.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1599

rreusser: oh wow, nice!

etpinard: https://github.com/plotly/plotly.js/pull/1599 is now merged in this PR.

etpinard:
@rreusser a well-earned :dancer: 

Not sure if @alexcjohnson 's https://github.com/plotly/plotly.js/pull/1595#discussion_r111461982 is blocking though. 

-----------

:tada: :rocket: :rocket:  :rocket:  :rocket:  :rocket:  :rocket: :rocket:  :rocket:  :tada:

alexcjohnson:
> Not sure if @alexcjohnson 's #1595 (comment) is blocking though.

Nah, if it only gets sparing use it's not blocking. Awesome work @rreusser & @etpinard !

rreusser:
Removed nonetheless. You're correct that it really wasn't needed there and wasn't use elsewhere. 👍

src/traces/carpet/attributes.js
```diff
@@ -0,0 +1,121 @@
...
+        description: [
+            'The shift applied to each successive row of data in creating a cheater plot.',
+            'Only used if `x` is been ommitted.'
+        ].join(' ')
+    },
+    aaxis: extendFlat({}, axisAttrs),
+    baxis: extendFlat({}, axisAttrs),
+    font: {
+        family: extendFlat({}, fontAttrs.family, {
+            dflt: '"Open Sans", verdana, arial, sans-serif'
```
alexcjohnson: @rreusser a little late for me to notice this, but is there a reason this font gets its own defaults? Is it too late to make this inherit from the global `layout.font`? I see that the various axis fonts inherit from this, which is great.
cc @etpinard

src/traces/carpet/axis_attributes.js
```diff
@@ -0,0 +1,445 @@
...
+            'If the axis `type` is *date*, it should be date strings,',
+            'like date data, though Date objects and unix milliseconds',
+            'will be accepted and converted to strings.',
+            'If the axis `type` is *category*, it should be numbers,',
+            'using the scale where each category is assigned a serial',
+            'number from zero in the order it appears.'
+        ].join(' ')
+    },
+
+    fixedrange: {
```
alexcjohnson: @rreusser does `fixedrange` do anything on a carpet axis? Can I remove it?

rreusser: 🔪

src/traces/carpet/attributes.js
rreusser: Ah, good point. Yes, I valid to make the switch now. Though it's technically a change, I think the number of actual breakages would be outweighed by API consistency.

alexcjohnson: Great - I just noticed some of these things as I'm pushing more of the restyle/relayout logic into the schema... I'll take care of these after I'm done with that.

etpinard: > Is it too late to make this inherit from the global layout.font?

Nop, not too late. I'd call this a bug fix.

PR #1598, opened 2017-04-13
Mapbox to image context

etpinard:
fixes https://github.com/plotly/plotly.js/pull/1598

`toImage` and `downloadImage` are broken currently for graphs with mapbox subplot and access token set via the `config` argument in `Plotly.plot` (e.g. like on plot.ly embed and shareplot).


etpinard:
@alexcjohnson might as well get this one in `1.26.0` if you don't mind reviewing this.

alexcjohnson:
Yeah, looks legit to me! 💃

PR #1602, opened 2017-04-17
Fix tspan selector

### LINKED ISSUE #1556: <span>'s in text labels move during animations
I have created a scatter plot that changes the opacity of a trace when clicked.
I have also set text labels for each point.
The text labels contain `<span>`'s and I've noticed they move during animations.

Here's a [codepen](http://codepen.io/n-riesco/pen/yMrwgO) that illustrates the issue better than I can describe.

rreusser:
fixes #1556 

I thought this was going to be very problematic since I struggled with it a lot the first time, but it turned out to be pretty simple. Just required selecting `tspan.line` instead of `tspan` when tweaking properties for animation. It was positioning regular inline spans the same as lines, which I think was the extent of the problem.

cc @n-riesco 

No tests added since it's transient behavior that I don't quite know a good way to test.

An animation of this working simultaneously with tspans representing both `<span>` and `<br>` tags:

ezgif-3-da7582232b

### DIFF
--- a/src/traces/scatter/plot.js
+++ b/src/traces/scatter/plot.js
@@ -470,7 +470,7 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
                 var x = d.xp || xa.c2p(d.x),
                     y = d.yp || ya.c2p(d.y);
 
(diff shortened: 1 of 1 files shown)


etpinard:
Nicely done :dancer:

PR #1604, opened 2017-04-19
Don't try to draw spikeline if the axis isn't on the chart.

rpaskowitz:
Noticed the line wasn't being drawn and errors were being logged to console on the last chart at https://plot.ly/javascript/line-charts/. This should clean that up.

### DIFF
--- a/src/plots/cartesian/graph_interact.js
+++ b/src/plots/cartesian/graph_interact.js
@@ -864,8 +864,8 @@ function createSpikelines(hoverData, opts) {
     var c0 = hoverData[0];
     var xa = c0.xa;
     var ya = c0.ya;
-    var showX = xa.showspikes;
-    var showY = ya.showspikes;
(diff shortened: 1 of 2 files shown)


etpinard:
Hmm. I'm not convinced this is the desired behavior (even though this patch clears some console errors). I can see spikes on invisible axes being somewhat useful (?)

etpinard:
cc @alexcjohnson

alexcjohnson:
Nice catch @rpaskowitz! But I think @etpinard is right, we should find a way to still show the spikes. Particularly in `across` mode the spike can be useful without pointing to a labeled value on an axis.

I propose that if the axis is invisible, we act as though it would have been at the farthest left/bottom of all subplots it's used in. Sound reasonable?

rpaskowitz:
Yeah, given the use of comparing data points, still drawing the line makes sense.

etpinard:
Will be fixed in https://github.com/plotly/plotly.js/pull/1980

PR #1609, opened 2017-04-19
Sort transform

etpinard:
Introducing a `sort` transform which can be very useful for example to sort marker points in descending order:

peek 2017-04-19 10-51


etpinard:
cc @alexcjohnson @n-riesco

src/transforms/sort.js
```diff
@@ -0,0 +1,189 @@
...
+            'For example, set `target` to *marker.size* to sort',
+            'about the marker size array.',
+
+            'If an array, *target* is then the data array by which',
+            'the sort transform is applied.'
+        ].join(' ')
+    },
+    order: {
+        valType: 'enumerated',
+        values: ['ascending', 'descending'],
```
etpinard: I chose _ascending_ and `descending` to be familiar with axis [`categoryorder`](https://github.com/plotly/plotly.js/blob/ef0946cfef97c446363edc11edf2e149d9297438/src/plots/cartesian/layout_attributes.js#L578)

src/transforms/sort.js
```diff
@@ -0,0 +1,189 @@
...
+    // special case for 'ids'
+    // -> cast to String
+    if(target === 'ids') return function(v) { return String(v); };
+
+    // otherwise (e.g. numeric-array of 'marker.color' or 'marker.size')
+    // -> cast to Number
+    return function(v) { return +v; };
+}
+
+function getIndices(opts, targetArray, d2c) {
```
etpinard: Challenge to all plotly.js contributors: find a faster sort algorithm that this O(n^2) attempt. Note that binary search is nice, but doesn't work so great for array with duplicate items.

src/transforms/sort.js
```diff
@@ -0,0 +1,189 @@
...
+        for(var j = 0; j < len; j++) {
+            arrayNew[j] = arrayCopy[indices[j]];
+        }
+
+        np.set(arrayNew);
+    }
+};
+
+// TODO reuse for filter.js
+function getTargetArray(trace, target) {
```
etpinard: Thinking about putting this in `src/lib/`. Thoughts?

src/transforms/sort.js
```diff
@@ -0,0 +1,189 @@
...
+
+        return Array.isArray(array) ? array : [];
+    }
+    else if(Array.isArray(target)) return target.slice();
+
+    return false;
+}
+
+// TODO reuse for filter.js
+function getDataToCoordFunc(gd, trace, target, targetArray) {
```
etpinard: Thinking about putting this in `src/plots/cartesian/`. Thoughts?

src/transforms/sort.js
etpinard: cc @rreusser @monfera

src/transforms/sort.js
monfera: 👍

src/transforms/sort.js
monfera: also 👍

src/transforms/sort.js
monfera: Stable sorting may be useful, e.g. the user presorted based on some criteria, we'd keep that, even if this new facility will support sorting based on multiple fields.  Stable sorting is simple with the built-in `Array.prototype.sort` as in this [recently merged thing](https://github.com/plotly/d3-sankey/pull/1/files). The built-in sort is rather fast.

monfera: The function `d2c` seems to always be `getDataToCoordFunc` which looks a bit expensive. Since it's a lot of calls to this function - 2 calls on each comparison - there's probably benefit to first map (e.g. in a loop) to get the `d2c` values for all items, and then sort based on that. One way is sorting value/index pairs as above, or more simply, just the indices in a numeric array. Another is to add the `d2c` value as a property to the objects to be sorted, and then use `return a.d2cValue - b.d2cValue || a.originalIndex - b.originalIndex` similar to the above link in which case the objects themselves are easily sorted (using the accessor is very fast).

rreusser: I think I fixated on stable due to the initial note that "[binary search] doesn't work so great for arrays with duplicate items." To the best of my knowledge `Array.prototype.sort` is [not stable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

monfera: Correct, but with the above trick you can make it so.

monfera: @etpinard would something like this work? I'm probably blind to the importance of the double loop...
```
var indices = d3.range(targetArray.length);
var values = targetArray.map(getSortFunc(opts, d2c));
return indices.sort(function(i, j) {return (values[i] - values[j]) || (i - j);};
```

src/transforms/sort.js
alexcjohnson: sure, there's nothing explicitly cartesian about this but that's where we have all our axis logic, makes more sense there than hiding in a transform. I guess this says at some point we might want to make a `src/components/axes/` or something... but that's a long discussion and even longer project.

src/transforms/sort.js
etpinard: Thanks @monfera , that `.map(d2c)` pass pre-`.sort` sounds like a good idea :racehorse:

monfera: ^^ put in a bit more code for copy/paste into console in isolation and fixed a typo

monfera: ^^ another fix and using actual objects to verify stability of the sorting

etpinard: @monfera how does ^^ handle duplicate items?

etpinard: Some interesting in https://gist.github.com/etpinard/86949a01309d5505fab8ccca2b0cd875

Looks like my dumb attempt isn't that bad after all (barring any typos ...)

src/transforms/sort.js
etpinard: > we might want to make a `src/components/axes/`

I like to think of components as optional, i.e. things that could be left out of `plotly.js/lib/core.js` for users that want to make custom bundle w/o them to shave off some :hamburger:.  As we can't (right?) make the axes logic optional, so I'd vote :-1: for making `axes` a component.  But agreed, most the axes logic isn't strictly cartesian. Maybe it's time for making a `src` directory e.g. called `src/base/` or `src/common/` or `src/plots/common/` for non-`src/lib/` things that are used across multiple plot types?

src/lib/index.js
```diff
@@ -344,6 +344,17 @@ lib.mergeArray = function(traceAttr, cd, cdAttr) {
     }
 };
 
+lib.getTargetArray = function(trace, target) {
+    if(typeof target === 'string' && target) {
+        var array = lib.nestedProperty(trace, target).get();
+
+        return Array.isArray(array) ? array : [];
```
alexcjohnson: Either this should be `: false` or [the one below](https://github.com/plotly/plotly.js/pull/1609/files/0be666f1c94d6762561c09aca3c9965ae09f53fa..823375ccbdf6260c5b7278cbdd4a62908da6ff1c#diff-e64ef427412e9bab4929b8b93ca12d96R355) should be `[]`

I'd probably vote for `false` and then instead of the `if(!len) return;` check, do `if(!targetArray) return;`

That said, this raises the question, in these applications, of what we should do when `targetArray` isn't the same length as the other associated arrays. Looks like for both, it will truncate (or extend) all of them to the length of `targetArray`. Is that what we want? Seems right for sort, but if we agree then lets 🔒  it with a test. For filter there's an argument to be made that eg `[]` and `)(`, and `{}` and `}{`, should be complementary - which they are *within* `targetArray` but not with respect to the items *beyond* it. Maybe that's still what we want, or maybe we don't want the extra complexity in so far as we think this would be incorrect usage with no legitimate use case. I don't have a strong opinion other than that it should be clear how we handle this case.

BTW it took me a minute to recall the motivation for the different cases in this function - can you give the function a docstring?

src/transforms/sort.js
```diff
@@ -0,0 +1,134 @@
...
+
+    if(!len) return;
+
+    var arrayAttrs = PlotSchema.findArrayAttributes(trace);
+    var d2c = Axes.getDataToCoordFunc(gd, trace, target, targetArray);
+    var indices = getIndices(opts, targetArray, d2c);
+
+    for(var i = 0; i < arrayAttrs.length; i++) {
+        var np = Lib.nestedProperty(trace, arrayAttrs[i]);
+        var arrayCopy = np.get().slice();
```
alexcjohnson: Is it actually necessary that *this* be a copy? `arrayNew` is already a different object...

src/lib/index.js
etpinard: > in these applications, of what we should do when targetArray isn't the same length as the other associated arrays. Looks like for both, it will truncate (or extend) all of them to the length of targetArray. Is that what we want? Seems right for sort, but if we agree

Yep, that's how I designed back when `filter` was first added. 

> but if we agree then lets :lock: it with a test. 

Sure thing.

>  can you give the function a docstring?

good call :+1:

src/transforms/sort.js
etpinard: Yeah, good call. only reading the values from those arrays here.

src/lib/index.js
etpinard: done in 2dd1333a0706c91628dc596d147e6f88d18b5183 , 9c70ee963d5af87c81e84af01cda3da81d45dbab and  4f9195785367c253fd27c4b5471b33a45e3e19b6

test/jasmine/tests/transform_sort_test.js
```diff
@@ -146,6 +146,58 @@ describe('Test sort transform calc:', function() {
...
+        expect(out[0].ids).toEqual(['p3', undefined, 'n2', 'p2', undefined, 'n1', 'p1', undefined, 'n0', 'z', undefined]);
+        expect(out[0].marker.color).toEqual([0.4, undefined, 0.3, 0.3, undefined, 0.2, 0.2, undefined, 0.1, 0.1, undefined]);
+        expect(out[0].marker.size).toEqual([10, undefined, 5, 0, undefined, 20, 6, undefined, 10, 1, undefined]);
+
+        expect(out[1].x).toEqual([-2, -1, -2, 0, 1, 3, 1, undefined, undefined]);
+        expect(out[1].y).toEqual([1, 2, 3, 1, 2, 3, 1, undefined, undefined]);
+        expect(out[1].ids).toEqual(['n0', 'n1', 'n2', 'z', 'p1', 'p2', 'p3', undefined, undefined]);
+        expect(out[1].marker.color).toEqual([0.1, 0.2, 0.3, 0.1, 0.2, 0.3, 0.4, undefined, undefined]);
+        expect(out[1].marker.size).toEqual([10, 20, 5, 1, 6, 0, 10, undefined, undefined]);
+    });
```
alexcjohnson: Nice tests! 🎉

alexcjohnson:
Looks great! 💃

test/jasmine/tests/transform_sort_test.js
```diff
@@ -241,4 +242,102 @@ describe('Test sort transform interactions:', function() {
         .catch(fail)
         .then(done);
     });
+
+    it('does not preserve hover/click `pointNumber` value', function(done) {
```
etpinard: @n-riesco @alexcjohnson IMPORTANT: event data `pointNumber` does not correspond to array index in `gd.data` after a sort transform. The same is true for filter transform with `preservegaps: false`. This is an unfortunate consequence of the current transform pipeline. Using `ids` become crucial here. That is, as presented in this test case:

```js
var pt = eventData.points[0]

pt.fullData.ids[pt.pointNumber] 
// => is preserved after a sort transform
```

We could eventually add a `originalPointNumber` field to the event data to solve this problem. Moreover, we should really add an `id` field to the hover & click event data and make the `ids` data array more ubiquitous (it is only available for scatter traces at the moment).

PR #1613, opened 2017-04-20
Fx component

etpinard:
to be merged in https://github.com/plotly/plotly.js/pull/1582

Introducing `src/components/fx/` home to all things hover. As hover logic is becoming more and more centralized and less and less _cartesian-only_, moving this out of `src/plots/cartesian/` is a no-brainer.

Down the road (maybe in this PR if easy) we should make `Fx` a optionally registerable component so that we can remove from our `lib/core.js` module, making for smaller custom bundles for users not caring about hover labels. Note that `src/components/fx/hover.js` is over 1000 LOC.

As a bonus, this reorganisation removed <strike>4</strike> 5 circular dependencies from `src/` (ref https://github.com/plotly/plotly.js/issues/236) :tada:


src/components/fx/constants.js
```diff
@@ -0,0 +1,26 @@
...
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
```
etpinard: @alexcjohnson anything else that might be better off in here?

alexcjohnson: nope, I think you got it all!

alexcjohnson:
Nicely done! 💃

etpinard:
> (maybe in this PR if easy) we should make Fx a optionally registerable component

Yep, this turned out to be easy in https://github.com/plotly/plotly.js/pull/1613/commits/a96af57c50a09fdcd9db3682d5e3b1aceedbd4c2 and https://github.com/plotly/plotly.js/pull/1613/commits/abcf1b034c637aa2a56203bdd7724fa870ec9cf3 :white_check_mark:

alexcjohnson:
> Yep, this turned out to be easy in a96af57 and abcf1b0 ✅

And 🔪 yet another circular dep! 🍻

Issue #1614, opened 2017-04-20
Python API plotly.offline: Save Plot as SVG, "rgba" not translated correctly if using Pie Chart.


dsholes:
With Python API, I have been setting _data.marker.color_ using an "rgba" string. When saving the plot as an SVG image using plotly.offline.plot(figure,image='svg',), the "rgba"  is translated into "rgb" and "fill-opacity" as long as the figure is a bar chart or stacked bar chart, which is the preferred behavior. 

However, when using a Pie chart, the SVG style "fill" for each Slice attempts to use "rgba" (i.e. it is not translated to "rgb" and "fill-opacity"). This is not an issue when viewing the SVG in the browser. However, Inkscape, GIMP (and I assume Illustrator) cannot interpret "rgba" values, and so opening the Pie chart SVG gives only black slices (i.e. the fill for each slice is "Unset").

This is not a huge issue, as users can manually set alpha using _data.opacity_ and pass an "rgb" string to _data.marker.color_, producing an SVG that behaves correctly in Inkscape. 

I just want to highlight the fact that the behavior is different for Bar Charts and Pie Charts.

This may be related to user's issue in plotly.js https://github.com/plotly/plotly.js/issues/707.

alexcjohnson:
Thanks for the report @dsholes - #1615 should fix this.

PR #1615, opened 2017-04-21
Split up fill and fill-opacity for pies

alexcjohnson:
quick fix for #1614 - just forgot to use `Color.fill` there.
cc @etpinard


etpinard:
:dancer: Very nice! This will be part of `1.26.1` shipping today.

Issue #1608, opened 2017-04-19
Marker text size distorted during animations


n-riesco:
Taken from #1607.

Here's a [codepen](http://codepen.io/n-riesco/pen/GmoZEx) that shows the issue.

![peek 2017-04-19 10-49](https://cloud.githubusercontent.com/assets/6199391/25174028/0331d608-24ee-11e7-825a-69820ce5bf7c.gif)

alexcjohnson:
And during zoom (by dragging an axis end or corner). @rreusser it was you who put in the inverse-distortion code for points and lines, right? Should be easy to extend to text?

etpinard:
Closed via https://github.com/plotly/plotly.js/pull/1616 by @rreusser

PR #1616, opened 2017-04-21
Inverse-transform text on layout transition

rreusser:
Addresses #1608. cc @alexcjohnson @n-riesco 

See: http://rickyreusser.com/demos/plotly-unsupported/text-transition.html

(The hover/click behavior is frustratingly finicky, but that's a separate issue. Also the date axes just return NaN which is why things are weird with dates. That's not addressed here.)

As for this PR, I'm a little worried `.points g` is too general a selector, so may need to add a class on text points to ensure *only* text is selected/modified by this tweak. Finally the transform logic is more than a little weird/ugly, but there's probably no avoiding that.

### DIFF
--- a/src/plots/cartesian/transition_axes.js
+++ b/src/plots/cartesian/transition_axes.js
@@ -151,10 +151,10 @@ module.exports = function transitionAxes(gd, newLayout, transitionOpts, makeOnCo
             // scale to individual points to counteract the scale of the trace
             // as a whole:
             .selectAll('.points').selectAll('.point')
-                .call(Drawing.setPointGroupScale, 1, 1)
+                .call(Drawing.setPointGroupScale, 1, 1);
 
         subplot.plot.selectAll('.points').selectAll('g')
-            .each(function () {
+            .each(function() {
                 var el = d3.select(this);
                 var existingTransform = el.attr('transform').match(LAST_TRANSLATION_RE);
                 el.attr('transform', existingTransform || '');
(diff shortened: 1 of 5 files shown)


src/plots/cartesian/transition_axes.js
```diff
@@ -17,6 +17,8 @@ var Drawing = require('../../components/drawing');
 var Axes = require('./axes');
 var axisRegex = /((x|y)([2-9]|[1-9][0-9]+)?)axis$/;
 
+var LAST_TRANSLATION_RE = /translate\([^)]*\)\s*$/;
```
rreusser: Note: this will break or at least do something very unexpected if the last transform is not a vertical offset for the text. Not sure how to handle that more generally.

alexcjohnson:
Adding a class to text groups seems like a great idea - there aren't too many places this would be needed, and then we could ensure this is targeted exactly where we want it. For example, bars with text (eg test_dashboard/#bar_attrs_relative) often the text gets autosized, with a maximum font size that shrinks to fit if necessary, and even auto rotation. This seems like a disaster to replicate on the fly, so I'd vote to just omit bar text from this effect at least for now.

Make sure it works with axis drag zooms and scroll zooms too!

src/plots/cartesian/transition_axes.js
etpinard: Can you just use [`Drawing.getTranslate`](https://github.com/plotly/plotly.js/blob/master/src/components/drawing/index.js#L588) instead?

rreusser: No, for two reasons, I think. First, it doesn't get the last translation explicitly, and second, it teases apart the numbers involved, which is unnecessary here. We only need to preserve the last transform without modification, not parse and reconstruct it.

rreusser:
@alexcjohnson I've pushed a fix that fixes an error I'd made in applying the class and have refactored things into a method that sets the scale for a group of text points. The result works with transitions and zooming:

ezgif-1-0a5d3c4b7c

Last step is to 🔒 the drawing method logic with a quick test. Need to head back to conference, but will do that asap.

rreusser:
Okay, added test. I'm 😄

etpinard:
:dancer:

PR #1619, opened 2017-04-22
Fixing ohlc showing wrong color when opening equals closing

noway:
Haven't found how to contact maintainers (no irc/slack, apart from community.plot.ly which doesn't seem to be the appropriate place?) so posting the PR here.

Fixed the case when opening is equal to closing and it always showing "increasing" color. It should instead determine the color based on the relation of previous close price to close price of the new one, or, if that's equal too, on the color of the previous candle.

So that it renders correctly in the following case:
2017-04-22-200734_266x595_scrot

### DIFF
--- a/src/traces/ohlc/helpers.js
+++ b/src/traces/ohlc/helpers.js
@@ -97,13 +97,46 @@ exports.makeTransform = function(traceIn, state, direction) {
 exports.getFilterFn = function(direction) {
     switch(direction) {
         case 'increasing':
-            return function(o, c) { return o <= c; };
+            return function(o, c, isPrevThisDirection, oprev, cprev) {
+                if (o == c) { 
+                    if (c > cprev) {
+                        return true // increasing
+                    } else if (c < cprev) {
+                        return false // decreasing
+                    } else {
+                        if (isPrevThisDirection === true) {
+                            return true // determine by last candle
+                        } else if (isPrevThisDirection == false) {
+                            return false // determine by last candle
+                        } else {
+                            return true // If we don't have previous data, assume it was increasing
+                        }
+                    }
+                }
+                return o < c; 
+            };
 
         case 'decreasing':
-            return function(o, c) { return o > c; };
+            return function(o, c, isPrevThisDirection, oprev, cprev) { 
+                if (o == c) { 
+                    if (c > cprev) {
+                        return false // increasing
+                    } else if (c < cprev) {
+                        return true // decreasing
+                    } else {
+                        if (isPrevThisDirection === true) {
+                            return true // determine by last candle
+                        } else if (isPrevThisDirection == false) {
+                            return false // determine by last candle
+                        } else {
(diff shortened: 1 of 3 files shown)


etpinard:
> Haven't found how to contact maintainers

Usually, contributors first open up an issue about the problem and discuss it there. Then contributors are encouraged to make a PR to their forked repo ask for a review there. Community PRs that don't reference an issue ticket can sometimes feel a little rude to FOSS maintainers.

No need to close your PR and redo the process though. Take it as an FYI.

---------

About

> Fixed the case when opening is equal to closing and it always showing "increasing" color. It should instead determine the color based on the relation of previous close price to close price of the new one, or, if that's equal too, on the color of the previous candle.

That sounds fair. But, can you confirm that this is the official way of handling this? How does this compare to what other graphing libraries are doing?

noway:
bitcoinwisdom which is a really reputable example deals with it this way.
screen shot 2017-04-27 at 2 41 29 am
(in current version all of those dashes would be green)

src/traces/ohlc/helpers.js
```diff
@@ -97,13 +97,46 @@ exports.makeTransform = function(traceIn, state, direction) {
 exports.getFilterFn = function(direction) {
     switch(direction) {
         case 'increasing':
-            return function(o, c) { return o <= c; };
+            return function(o, c, isPrevThisDirection, oprev, cprev) {
```
etpinard: The algo looks good :+1: 

It might worth the time to make it a little DRYer by wrapping `isPrevThisDirestion`, `oprev` and `cprev` in a closure. That is, something like:

```js
exports.getFilterFn = function(direction) {
  return new _getFilterFn(direction);
};

function _getFilterFn(directon) {
  var isPrevThisDirection = null;
  var oprev = null;
  var cprev = null;
  
  var fn;
  switch(direction) {
    case 'increasing':
      fn = function(o, c) { /* */ };
      break;
    case 'decreasing':
      fn = function(o, c) { /* */ };
      break;
  }

  return function(o, c) {
    var out = fn(o, c);
    isPrevThisDirection = !!out;
    oprev = o;
    cprev = c;
  };
};
```

that way we wouldn't have to patch `candlestick/transform.js` and `ohlc/transform.js`.

noway:
That's a good idea. Fixed, although haven't had a chance to test it.

etpinard:
@noway thanks. If you're up for a challenge you can try adding a few test cases in this [suite](https://github.com/plotly/plotly.js/blob/93d5d0b1af202f36e8af1594ebc3bea3959888ad/test/jasmine/tests/finance_test.js#L379).

noway:
Unfortunately I'd have to pass, can't allocate enough of a chunk of time. Otherwise I'm happy to help :-)

etpinard:
Superseded by https://github.com/plotly/plotly.js/pull/1655

Issue #581, opened 2016-05-30
Line colorscale on scatter graph


MonsieurV:
Hello,

I can't make work the [marker.line.colorscale](https://plot.ly/javascript/reference/#scatter-marker-line-color) property. I'm not sure how it should work, but what I can say is:
- using the `colorscale` property on lines as it is used on [markers](http://codepen.io/etpinard/pen/zqqeBJ) has [no effect](http://codepen.io/anon/pen/jrOKvY)
- if we use the property on a scatter graph with both lines and markers we have [a bug](http://codepen.io/anon/pen/aZbKPx) (or I'm not using it properly?)

So is it possible to color a line following a colorscale on a scatter graph? How we do it?

Moreover the documentation is confusing, as it states the property line is under the mark one:

> Sets the colorscale and only has an effect if `marker.line.color` is set to a numerical array.

Actually it seems not to be the case: http://codepen.io/anon/pen/OXJErp. Should the documentation states `line.color` instead?

I can do a PR to fix the doc, but I want to be sure I'm not misleading first.

MonsieurV:
Seems like the feature was already suggested in #147 [see](https://github.com/plotly/plotly.js/issues/147#issuecomment-194046036)

etpinard:
Thanks for the very detailed issue.

In plotly.js, attributes under `marker.line` describe the marker points' bordering line styling (example [here](http://codepen.io/etpinard/pen/WxNPrG?)) and have **no effect** on a `scatter` trace's line segments.

At the moment, scatter `line` attributes do not support color scales nor per-data-point style setting (as mentioned in https://github.com/plotly/plotly.js/issues/147). In other words, there is no way to apply a colour gradient to a plotly.js line chart.

Adding color scale support to line charts (possibly using svg [gradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Gradients)) would be a great addition.

etpinard:
We should also add this feature for `scatter3d` traces e.g. 

![image](https://cloud.githubusercontent.com/assets/6675409/15657008/0685b544-267b-11e6-9a3b-7918653e1fbc.png)

from [here](http://docs.enthought.com/mayavi/mayavi/building_applications.html).

etpinard:
Assigning the `scatter3d` part to @monfera

monfera:
A comment about SVG (note: no plotly support at the moment) which underlies the regular plots that don't use WebGL:

[...]

PR #1620, opened 2017-04-22
scatter marker color gradients

alexcjohnson:
- radial, horizontal, and vertical gradients for marker fills
- for all svg scatter types

cc @etpinard @charleyferrari - I won't be able to make changes until I'm back online mid-late week but wanted to post this for review.


src/components/drawing/index.js
```diff
@@ -271,24 +272,83 @@ function singlePointStyle(d, sel, trace, markerScale, lineScale, marker, markerL
...
-    var marker = trace.marker,
-        markerLine = marker.line;
+var HORZGRADIENT = {x1: 1, x2: 0, y1: 0, y2: 0};
+var VERTGRADIENT = {x1: 0, x2: 0, y1: 1, y2: 0};
 
-    // allow array marker and marker line colors to be
-    // scaled by given max and min to colorscales
-    var markerScale = drawing.tryColorscale(marker, ''),
-        lineScale = drawing.tryColorscale(marker, 'line');
+drawing.gradient = function(sel, gd, type, color1, color2) {
```
alexcjohnson: I was debating whether this goes in `color` or `drawing` - ended up deciding on `drawing` primarily because it's tied to `gd` (so it can create `defs` in the right place, without having to walk the tree with `Lib.getPlotDiv` for every marker) and nothing else in `color` is... but could be persuaded otherwise.

src/traces/scattercarpet/attributes.js
```diff
@@ -107,7 +107,8 @@ module.exports = {
         line: extendFlat({},
             {width: scatterMarkerLineAttrs.width},
             colorAttributes('marker'.line)
-        )
+        ),
+        gradient: scatterMarkerAttrs.gradient
```
alexcjohnson: TODO: test image with carpet plots. Somehow `imagetest` isn't working for me yet with carpets, must be something I still need to update?

rreusser: `rm -rf node_modules`?

src/components/drawing/index.js
etpinard: This belongs in `Drawing` no doubt. `Color` is really a lib (i.e. utility) module that shouldn't know anything about plotly attributes and especially `gd`.

src/traces/scattercarpet/attributes.js
etpinard: another TODO: looks like this PR broke `gl2d`.

test/image/mocks/ternary_multiple.json
```diff
@@ -31,7 +38,14 @@
             ],
             "subplot": "ternary2",
             "name": "b missing",
-            "uid": "6c0753"
+            "uid": "6c0753",
```
etpinard: might as well :hocho: those `uid` fields.

src/traces/scattercarpet/attributes.js
rreusser: Can you clarify which PR? Do you mean scattercarpet or *this* PR?

src/traces/scatter/marker_defaults.js
```diff
@@ -15,12 +15,18 @@ var colorscaleDefaults = require('../../components/colorscale/defaults');
 
 var subTypes = require('./subtypes');
 
-
+/*
+ * opts: object of flags to control features not all marker users support
```
etpinard: Thanks for :lock: ing this pattern. :tada:

src/traces/scattercarpet/attributes.js
etpinard: This as in https://github.com/plotly/plotly.js/pull/1620/ You can calm down @rreusser :smile:

src/components/drawing/index.js
```diff
@@ -271,24 +272,83 @@ function singlePointStyle(d, sel, trace, markerScale, lineScale, marker, markerL
...
+};
+
+/*
+ * Make the gradients container and clear out any previous gradients.
+ * We never collect all the gradients we need in one place,
+ * so we can't ever remove gradients that have stopped being useful,
+ * except all at once before a full redraw.
+ * The upside of this is arbitrary points can share gradient defs
+ */
+drawing.initGradients = function(gd) {
```
etpinard: Nice pattern. Controlling gradient `defs` in one place should help us in https://github.com/plotly/plotly.js/issues/581 too.

src/plot_api/plot_api.js
```diff
@@ -153,6 +153,9 @@ Plotly.plot = function(gd, data, layout, config) {
         makePlotFramework(gd);
     }
 
+    // clear gradient defs on each .plot call, because we know we'll loop through all traces
+    Drawing.initGradients(gd);
```
etpinard: It would be nice to add jasmine test checking that `restyle` can update gradient `defs` properly.

src/components/drawing/index.js
```diff
@@ -271,24 +272,83 @@ function singlePointStyle(d, sel, trace, markerScale, lineScale, marker, markerL
...
-    var markerScale = drawing.tryColorscale(marker, ''),
-        lineScale = drawing.tryColorscale(marker, 'line');
+drawing.gradient = function(sel, gd, type, color1, color2) {
+    var fullLayout = gd._fullLayout;
+    var gradientID = (
+        type + fullLayout._uid + '-' + color1 + '-' + color2
+    ).replace(/[^\w\-]+/g, '_');
+    var gradient = fullLayout._defs.select('.gradients').selectAll('#' + gradientID).data([0]);
+
+    gradient.enter()
```
etpinard: I suspect this algo would work for adding gradients to bar traces? Would be a cool feature :smirk:

src/traces/scattercarpet/attributes.js
alexcjohnson: scattercarpet image test in https://github.com/plotly/plotly.js/pull/1620/commits/105ce9b4ffdd77e62bc5bbc06f5e8a5f3f9fdb0a
gl2d fix still to come.

test/image/mocks/ternary_multiple.json
alexcjohnson: 🔪 🔪 🔪  https://github.com/plotly/plotly.js/pull/1620/commits/a0742598366d9f407857978ae931db8ea6d20ff7

src/components/drawing/index.js
alexcjohnson: yep, would be cool to add bar gradients... for another PR

test/jasmine/tests/drawing_test.js
```diff
@@ -309,3 +315,83 @@ describe('Drawing', function() {
...
+            // simple scalar restyle doesn't trigger a full replot, so
+            // doesn't clear the old gradients
+            checkGradientIds(gd._fullLayout._uid, [
+                ['radial', '#123', '#fff'],
+                ['radial', '#123', '#eee'],
+                ['radial', '#123', '#ddd'],
+                ['radial', '#456', '#fff'],
+                ['radial', '#456', '#eee'],
+                ['radial', '#456', '#ddd']
+            ]);
```
alexcjohnson: Not happy with this... there are a bunch of ways for it to blow up in actual use, like if someone makes 1000 different gradients via different marker colors, and animates them through 1000 intermediate states, suddenly you have a million gradients lying around.

I think I need to change it to associating the gradient with the point rather than with the color combo - unless someone has another idea how to keep this in check.

src/components/drawing/index.js
```diff
@@ -275,10 +281,20 @@ function singlePointStyle(d, sel, trace, markerScale, lineScale, marker, markerL
...
+        else gradientType = markerGradient && markerGradient.type;
+
         if(gradientType && gradientType !== 'none') {
-            sel.call(drawing.gradient, gd, gradientType, fillColor, gradientColor);
+            var gradientColor = d.mgc;
+            if(gradientColor) perPointGradient = true;
+            else gradientColor = markerGradient.color;
+
+            var gradientID = 'g' + gd._fullLayout._uid + '-' + trace.uid;
+            if(perPointGradient) gradientID += '-' + i;
```
alexcjohnson: Now we're making either one gradient for the whole trace (if we can) or one per point (if any of the attributes varies per point) and it will update if the attributes change. This means that several traces (or several points in one trace with array attributes) will no longer be able to share a gradient def, but I don't think there are any cases where this can blow up.

src/traces/scatter/arrays_to_calcdata.js
```diff
@@ -15,6 +15,9 @@ var Lib = require('../../lib');
 // arrayOk attributes, merge them into calcdata array
 module.exports = function arraysToCalcdata(cd, trace) {
 
+    // so each point knows which index it originally came from
+    for(var i = 0; i < cd.length; i++) cd[i].i = i;
```
alexcjohnson: @etpinard to my comment earlier today about not using `calcdata` as much: I needed to put this in because what I first tried to do, getting the index from the selection `.each(d, i)`, is not reliable, as evidenced by the scattercarpet test image: the 0th point on the plot was actually index 1 from the trace (point 0 is cut off), but then the legend thought IT was point 0 so overwrote the other point 0 gradient.

BUT once we have that, we could get rid of the rest of these `Lib.mergeArray` statements - which, to address your concern earlier, don't do any sanitization anyway, we only do that for the axis data - and just directly access the arrays (if they are arrays) from within the style routines.

I'm not going to do that now, but could be a nice 🚀 at some point.

test/image/mocks/ternary_multiple.json
etpinard: :arrow_heading_up: commit of the year

src/components/drawing/index.js
etpinard: That sounds right :+1:

src/traces/scatter/arrays_to_calcdata.js
etpinard: Good :eye: here and nice fix. But, I think the _solution_ would be to make `Lib.mergeArray` sanatize those arrays instead of :hocho: it. We can debate this in a future issue / PR :smile:

src/components/drawing/index.js
```diff
@@ -298,11 +374,12 @@ drawing.pointStyle = function(s, trace) {
     // allow array marker and marker line colors to be
     // scaled by given max and min to colorscales
     var marker = trace.marker;
-    var markerScale = drawing.tryColorscale(marker, ''),
-        lineScale = drawing.tryColorscale(marker, 'line');
+    var markerScale = drawing.tryColorscale(marker, '');
+    var lineScale = drawing.tryColorscale(marker, 'line');
+    var gd = Lib.getPlotDiv(s.node());
```
etpinard: I wonder if it might be better to add a ref to those gradient defs in `fullData[i]` then to look for `gd`? Or even better refactor those `Drawing` method to expect `fullLayout` as argument.

etpinard:
I'm not a big fan of that `Lib.getPlotDiv` call as highlighted [here](https://github.com/plotly/plotly.js/pull/1620/files#r114479128), but, then again, it's not that big of a deal and I'll call that non- :no_entry: for `1.27.0`

:dancer: nicely done.

src/components/drawing/index.js
alexcjohnson: I'll leave it for now - I agree this is suboptimal but neither of these alternatives seems particularly simpler to me. If I had to pick one I'd probably go for the first actually... stashing the `<defs>` in each trace could come in handy elsewhere too.

Issue #950, opened 2016-09-19
Lint with standardjs


etpinard:
The [standard](https://github.com/feross/standard) set of linting rules are becoming the standard :laughing: for open-source JS libraries. 

We should follow suit.

rreusser:
Strong choice! See also: [standard-format](https://github.com/maxogden/standard-format)

etpinard:
I believe plotly.js should remain a ES5-only repository for the foreseeable future. I also believe that `npm run lint` should catch all non-ES5 slip-ups.

Eslint is configured with ES5-only parser by default ([ref](http://eslint.org/docs/user-guide/configuring#specifying-parser-options)), but standard now uses an ES8 parser :astonished:  ([ref](https://github.com/feross/eslint-config-standard/blob/85872703713090cd84bc80f3b6d2076cc5769ed0/eslintrc.json#L3)) meaning that running the stock `standard` in plotly.js will not catch non-ES5 tokens.

So, I propose using [`standard-own`](https://github.com/o2team/standard-own) which uses that same [engine](https://github.com/flet/standard-engine) as standard but allows uses to specify which ECMA script to parse. I think adding:

```json
{
  "standard-own": {
    "eslintConfig":  {
       "parserOptions": {
          "ecmaVersion": 5
        }
     }
  }
}
```

to the package.json should do the trick.

Alternatively, we could publish an es5 version of [`eslint-config-standard`](https://github.com/feross/eslint-config-standard) and use that?

rreusser:
Interesting. Agreed on keeping it ES5 for the sake of consistency. (though [es2020](https://github.com/yoshuawuyts/es2020) hits the nail on the head, IMO.)

Related discussion: https://github.com/feross/standard/issues/159

rreusser:
Though I'm mostly trying to understand the consequences of the parser vs. the set of rules and what that means for simply enforcing style.

rreusser:
Testing out `./node_modules/.bin/standard --fix` now, just to watch things burn…

etpinard:
To enforce ES5, we could use [`use-es5-only`](https://github.com/azu/use-es5-only) which simply errors out when trying to parse non-es5 tokens (w/o listing detail about the error) or try to write something better using [`esutils`](https://www.npmjs.com/package/esutils#keyword) to provide better info about the non-ES5 token found.

rreusser:
[...]

PR #1624, opened 2017-04-26
Parcoords fonts

monfera:
Support for specifying fonts for `parcoords` dimensions. Example, see JS code 8..12:http://codepen.io/anon/pen/QvdwpX

image


src/traces/parcoords/defaults.js
```diff
@@ -81,6 +81,10 @@ function dimensionsDefaults(traceIn, traceOut) {
     return dimensionsOut;
 }
 
+function coerceFont(fontAttr, coerce, layoutFont, defaultFont) {
+    var fontSpec = Lib.coerceFont(coerce, fontAttr);
+    Lib.coerceFont(coerce, fontAttr, Lib.extendFlat({}, layoutFont, defaultFont, fontSpec));
```
etpinard: Why do you need to call `Lib.coerceFont` twice?

src/traces/parcoords/attributes.js
```diff
@@ -12,10 +12,12 @@ var colorAttributes = require('../../components/colorscale/color_attributes');
 var colorbarAttrs = require('../../components/colorbar/attributes');
 var colorscales = require('../../components/colorscale/scales');
 var axesAttrs = require('../../plots/cartesian/layout_attributes');
+var fontAttrs = require('../../plots/font_attributes');
 
 var extendDeep = require('../../lib/extend').extendDeep;
 var extendFlat = require('../../lib/extend').extendFlat;
 
+
```
etpinard: :hocho:

src/traces/parcoords/attributes.js
```diff
@@ -47,6 +49,10 @@ module.exports = {
         }
     },
 
+    labelfont: extendFlat({}, fontAttrs, {dflt: {size: 10}, description: 'Sets the font for the `dimension` labels.'}),
```
etpinard: This is not our usual pattern.

The default font size is inherited from [`fullLayout.font.size`](https://github.com/plotly/plotly.js/blob/460ac39ae99f0e3de46cb9307dc77f581ecbafcb/src/plots/plots.js#L934) directly as [here](https://github.com/plotly/plotly.js/blob/460ac39ae99f0e3de46cb9307dc77f581ecbafcb/src/components/annotations/annotation_defaults.js#L46) or times a scalar like [here](https://github.com/plotly/plotly.js/blob/460ac39ae99f0e3de46cb9307dc77f581ecbafcb/src/plots/cartesian/axis_defaults.js#L83).

src/traces/parcoords/defaults.js
```diff
@@ -97,4 +100,8 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     if(!Array.isArray(dimensions) || !dimensions.length) {
         traceOut.visible = false;
     }
+
+    coerceFont('labelfont', coerce, layout.font, attributes.labelfont.dflt || {});
```
etpinard:
```js
Lib.coerceFont('labelfont', coerce, layout.font);
```

should be enough, no?

src/traces/parcoords/parcoords.js
```diff
@@ -227,8 +234,6 @@ function styleExtentTexts(selection) {
     selection
         .classed('axisExtentText', true)
         .attr('text-anchor', 'middle')
-        .style('font-weight', 100)
-        .style('font-size', '10px')
```
etpinard: Ha. I see `10px` was used before. I should've caught that. Elsewhere in plotly.js this would be [`12px`](https://github.com/plotly/plotly.js/blob/460ac39ae99f0e3de46cb9307dc77f581ecbafcb/src/plots/layout_attributes.js#L23).

I guess we'll have to keep it this way until v2 :grimacing:

src/traces/parcoords/attributes.js
etpinard: ... as we're (slowly) trying to make our code look more like the `standard` style - which disallows multiple blank lines.

monfera: Ah got it, thanks!

monfera: Btw HUGE fan of standard style

rreusser: Now that there aren't giant outstanding PRs, might be time to revisit: https://github.com/plotly/plotly.js/pull/1371

src/traces/parcoords/attributes.js
monfera: The purpose was to have _smaller_ fonts than what would come from the `layout` defaults, which would result in overly big numbers, looked really bad. The `dimension` labels and tick numbers need to be quite small by default, e.g. around 10px. It's achieved by `parcoords/attributes` specifying these sizes. Is there another way of saying, 'take layout font values but don't adopt large fonts unless the user wants it, in which case they specify it on eg. `labelfont`?

src/traces/parcoords/attributes.js
etpinard: ... we'll have to convince @alexcjohnson first.

rreusser: --> The selling point of prettier is that you don't have to adhere to any style at all. 😄

src/traces/parcoords/defaults.js
monfera: I could do that. The consequence is that the tick and dimension label fonts would be too large:
image

It's even worse if the user specified a larger font centrally, e.g. size 16:
image

Since I can go either way, pls confirm if I should preserve the current fontsize-moderation logic or use the simpler logic. Consistency matters a lot on one hand, and having ergonomically good defaults is also valuable.

src/traces/parcoords/attributes.js
etpinard: True. But you have to do some non-trivial src vs build file trickery.

etpinard: ... adding pre-commit hooks is hardly simplifying anything IMHO.

rreusser: See #1629 for quick experiment. It replaces lint and lint-fix and adds a precommit hook to format staged files without ever having to manually add a precommit hook.

alexcjohnson: I think we're bikeshedding https://github.com/plotly/plotly.js/issues/950#issuecomment-298144207

src/traces/parcoords/parcoords.js
```diff
@@ -227,8 +234,6 @@ function styleExtentTexts(selection) {
     selection
         .classed('axisExtentText', true)
         .attr('text-anchor', 'middle')
-        .style('font-weight', 100)
```
etpinard: @monfera is removing that font-weight setting on purpose? Unsurprisingly, it makes the image test fail.

monfera: As we discussed about font standardization... I assumed, perhaps incorrectly, that we don't want to customize it. I do have a preference for the lower font weight as in the case of `parcoords` the shape of line distribution is more important than the best readability so it's fine to keep the weight=100 but with configuration, the user can probably control the font anyway.

etpinard: We don't set `font-weight` anyway in our code - though inputting text as `<b>text</b>` effectively does [that](https://github.com/plotly/plotly.js/blob/ce970f112a8fff0d09909529be8a713f5d4f36cc/src/lib/svg_text_utils.js#L229). 

This is another thing I wished I would've noticed while reviewing your first parcoords PR. Oh well. I think it's best to drop `font-weight` and make parcoords look a little more plotly-esque.

monfera: Cool, in this case I could guess, if a bit late, how you'd go about it :-)

src/traces/parcoords/defaults.js
```diff
@@ -101,7 +97,15 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
     }
 
-    coerceFont('labelfont', coerce, layout.font, attributes.labelfont.dflt || {});
-    coerceFont('tickfont', coerce, layout.font, attributes.tickfont.dflt || {});
-    coerceFont('rangefont', coerce, layout.font, attributes.rangefont.dflt || {});
+    // make default font size 10px,
+    // scale linearly with global font size
+    var fontDflt = {
+        family: layout.font.family,
+        size: Math.round(layout.font.size * (10 / 12)),
```
etpinard: @monfera this is what I came off with. 

As you pointed out, changing the font size default to 12 makes most parcoord graphs look less-than-optimal. We should fix that text-overlapping problem at some point though - similar to how we do it currently in [cartesian](https://github.com/plotly/plotly.js/blob/2bac6ef08f46dae0117ba34dea1599866f060379/src/plots/cartesian/axes.js#L1844-L1897) axes. But that's for another PR. So, the 10px default value remain. 

But, to make this a little more plotly-esque, we make the parcoord font size scale linearly with `layout.font` similar to what's done for axes [`titlefont`](https://github.com/plotly/plotly.js/blob/2bac6ef08f46dae0117ba34dea1599866f060379/src/plots/cartesian/axis_defaults.js#L81-L85).

monfera: Thanks @etpinard, also for the pointer!

etpinard:
@monfera are you ok with my changes?

monfera:
@etpinard yes, thank you, I should have indicated this more unambiguously in my reply.

monfera:
Also, overall, looks good!

PR #1625, opened 2017-04-26
Remove clip path from contourcarpet

rreusser:
This PR removes the clip path from contourcarpet traces. If everything else is _perfect_, then it's already clipped and this serves no purpose. Expecting pixel-level differences (_half_ of lines along the edge are currently clipped by the clip path), but hopefully no substantial changes so that it can simply be removed.

Otherwise, I think maybe _reversing_ the clip path and changing the `clip-rule` might fix some of the self-intersecting carpet axis issues seen. But this is probably preferable.

### DIFF
--- a/src/traces/contourcarpet/plot.js
+++ b/src/traces/contourcarpet/plot.js
@@ -122,13 +122,6 @@ function plotOne(gd, plotinfo, cd) {
 
     // Draw contour lines:
     makeLines(plotGroup, pathinfo, contours);
-
-    // Clip the boundary of the plot:
-    clipBoundary(plotGroup, carpet);
-}
-
-function clipBoundary(plotGroup, carpet) {
-    plotGroup.attr('clip-path', 'url(#' + carpet.clipPathId + ')');
 }
 
 function makeContourGroup(plotinfo, cd, id) {


rreusser:
For reference, the only thing holding this PR back is a single test failure that I'm having a hard time simplifying to the point that it's debuggable. Specifically, there's no inherent guarantee that the contour splines must fall within the carpet axis (no guarantee of spline monotonicity etc.), so I'm trying to figure out whether content spilling out of the carpet axis is a result of that or whether it's a more problematic issue. If so, my first guess is that it's a result of [these lines](https://github.com/plotly/plotly.js/blob/master/src/traces/contourcarpet/join_all_paths.js#L78-L86) that I adapted from contour plots and struggled to convince myself were strictly correct.

rreusser:
I came up with a potential fix. It's possible to shade regions of the plot that are inverted. Unfortunately it's not at all simple to shade the overlapping region because it's not encircled by the plot boundary. Making this approach robust would mean:

1. taking the cross product of the tangents at the corner to figure out the dominant plot orientation
2. use that to invert the paths accordingly to shade the orientation that's in the minority. That means if it's all a consistent orientation, nothing will be shaded.

Note that this would *not* catch the case where the boundary is nice and the interior extends outside the carpet boundary.

screen shot 2017-04-27 at 14 48 32

rreusser:
Closing this as interesting but probably not a desirable thing. I think we can leave carpet axes that fold back on themselves as an plotter-beware feature for now. It's not reasonably possible to detect and reasonably handle all the pathological cases, so I think it's best to keep it simple.

PR #1626, opened 2017-04-26
Add textpoint class to scatter text points

rreusser:
This PR adds a `'textpoint'` class to text points of scatter traces so that they can more easily and consistently be selected. I have no preference on the class name, but `'textpoint'` seemed reasonably descriptive (better than `'text'`, which is a little vague/overgeneral IMO)


etpinard:
`textpoint` sounds good. :dancer:

PR #1632, opened 2017-04-27
Fix aggressive zooming

dfcreative:
Should fix https://github.com/plotly/phoenix-integration/issues/148#issuecomment-297547082


etpinard:
Amazing. Thanks!

PR #1638, opened 2017-04-28
Introducing 3D annotations

etpinard:
resolves https://github.com/plotly/plotly.js/issues/751 and more precisely implements https://github.com/plotly/plotly.js/issues/751#issuecomment-277280113.

peek 2017-04-28 17-29

This PR is still a very much WIP. Here are some TODOs:

- [x] get the autorange logic hooked in
- [x] get this to work with category axes types
- [x] register the new logic under `components/annotations/` (or a new `components/annotations3d` maybe?)
- [x] DRY up defaults
- [x] tests


src/components/annotations/draw.js
```diff
@@ -168,7 +168,6 @@ function drawOne(gd, index) {
     var font = options.font;
 
     var annText = annTextGroupInner.append('text')
-        .classed('annotation', true)
```
etpinard: This commit is reverted in ababc49 to make the tests added in https://github.com/plotly/plotly.js/pull/1683 :white_check_mark:

etpinard:
Ok. All the functionality is in (I think), so I'll tagged this as `reviewable`.

I'm currently thinking it would be to put this new 3d annotation code under a new component `components/annotations3d/` which would reuse much of `components/annotations/` but would remain separate allowing users that don't not need 3d annotations or 3d graphs for that matter to trim down their bundle size even more in v2.

Any objections before I shuffle the new code around?

src/plots/gl3d/scene.js
```diff
@@ -716,11 +738,121 @@ proto.toImage = function(format) {
...
+
+        // set infinite range so that annotation draw routine
+        // does not try to remove 'outside-range' annotations,
+        // this case is handled in the render loop
+        range: [-Infinity, Infinity]
+    };
+
+    ann._xa = {};
+    Lib.extendFlat(ann._xa, base);
+    Axes.setConvert(ann._xa);
```
alexcjohnson: what do you need from `setConvert`, if you're overriding `_offset` and `l2p`?

alexcjohnson:
> Any objections before I shuffle the new code around?

sounds reasonable, go for it!

src/components/annotations/draw.js
alexcjohnson: I think you've hit on a flaw in how github displays comments... I'm reading:

> This commit is reverted in ababc49 to make the tests added in https://github.com/plotly/plotly.js/pull/1683 :white_check_mark: 

but I'm looking at https://github.com/plotly/plotly.js/pull/1638/commits/a31549d3b7532c48200289607ab435e4e80d891f which came after ababc49 and rereverts the change originally made in https://github.com/plotly/plotly.js/commit/1a8f1ed1e855fb9157f728c191b00784fe3f1258 💫 

Anyway, totally approve of this change. My best guess is these got the same `.annotation` class because at some point in the distant past they weren't all encapsulated in one group like they are now, and I wanted to be able to more easily select the whole annotation. Has always been confusing to deal with...

etpinard: As of a31549d, the annotations `text` nodes have class name `annotation-text`.

src/components/annotations/draw.js
```diff
@@ -62,47 +63,65 @@ function draw(gd) {
...
+ * options (object): this annotation's options
+ * index (int): the annotation to draw
+ * xa (object | undefined): full x-axis object to compute subplot pos-to-px
+ * ya (object | undefined): ... y-axis
+ */
+function drawRaw(gd, options, index, xa, ya) {
+    var fullLayout = gd._fullLayout;
+    var gs = gd._fullLayout._size;
+
+    var className = options._scene ?
```
alexcjohnson: elsewhere `_scene` is an object, but this one is an id - can we call it `_sceneId` instead?

src/plots/gl3d/scene.js
alexcjohnson: > what do you need from setConvert, if you're overriding _offset and l2p?

Oh I see - `ax.l2p` is now controlling all the other `2p` conversions (instead of the local var `l2p`) so you get those for free. Seems fine - tiny tiny 🐎  hit from having to look up attributes more often... but should be negligible.

src/plots/gl3d/layout/layout_attributes.js
```diff
@@ -139,6 +142,83 @@ module.exports = {
...
+        },
+        z: {
+            valType: 'any',
+            role: 'info',
+            description: [
+                'Sets the annotation\'s z position.'
+            ].join(' ')
+        },
+        ax: {
+            valType: 'any',
```
alexcjohnson: `'number'` now since it's always going to be in px?

src/components/annotations/draw.js
```diff
@@ -62,47 +63,65 @@ function draw(gd) {
...
     var annGroup = fullLayout._infolayer.append('g')
-        .classed('annotation', true)
+        .classed(className, true)
         .attr('data-index', String(index))
         .style('opacity', options.opacity);
 
     // another group for text+background so that they can rotate together
     var annTextGroup = annGroup.append('g')
-        .classed('annotation-text-g', true)
-        .attr('data-index', String(index));
```
alexcjohnson: 👍  another relic 🔪

src/components/annotations/draw.js
etpinard: Good call. Done in c084da0e2282c6ea2e680e4e01dfcbbe910a949f

src/plots/gl3d/layout/layout_attributes.js
etpinard: Thanks! Done in adfc305a0918c4afcb8b17b737e281517ec09613

src/plots/gl3d/scene.js
etpinard: > Oh I see - ax.l2p is now controlling all the other 2p conversions (instead of the local var l2p) so you get those for free

Exactly.

src/components/annotations3d/attributes.js
```diff
@@ -75,13 +75,13 @@ module.exports = {
     arrowwidth: annAtts.arrowwidth,
     standoff: annAtts.standoff,
     hovertext: annAtts.hovertext,
-    hoverlabel: annAtts.hoverlabel
+    hoverlabel: annAtts.hoverlabel,
+    captureevents: annAtts.captureevents
```
etpinard: With this, hover labels on 3d annotations are working:
peek 2017-05-25 16-37

alexcjohnson: what about the `plotly_clickannotation` event? Looks like that might need a `sceneId` if there is one? Which brings up the point that for extensibility to other non-cartesian subplots perhaps it should really be called `subplotId`...

etpinard: - commit 8b8a088 generalises `drawRaw`
- commit 1d5606a add `subplotId` to the `ploltly_clickannotation` event data + :lock: it down.

etpinard: After 1d5606a, [this](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/test/jasmine/tests/annotations_test.js#L1202) and [that](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/test/jasmine/tests/annotations_test.js#L1226) assertions are failing, because the event data now contains `subplotId: false` for cartesian/paper-ref annotations.

@alexcjohnson would you prefer leaving `subplotId` out of the cartesian/paper-ref annotation event data or keep it as `subplotId: false`?

alexcjohnson: > would you prefer leaving subplotId out of the cartesian/paper-ref annotation event data or keep it as subplotId: false

haha didn't see this before [my comment](https://github.com/plotly/plotly.js/pull/1638#commitcomment-22295724)

yeah, lets omit it.

etpinard: done in 46ab63d

alexcjohnson:
That's all from my side, looks great! 💃

PR #1639, opened 2017-05-01
numeric sort for component edits

alexcjohnson:
Not the first time I've been bitten by the default sort being alphabetical even if the items are numbers... This made a bug where if you tried to add annotations 9 and 10 in a single `relayout`, it would try to add 10 first but it couldn't do that because 9 didn't exist yet (or if it did, 10 would get inserted in the array in the wrong places) and similar problems on deletion.

@etpinard OK?

### DIFF
--- a/src/plot_api/manage_arrays.js
+++ b/src/plot_api/manage_arrays.js
@@ -101,7 +101,7 @@ exports.applyContainerArrayChanges = function applyContainerArrayChanges(gd, np,
         return true;
     }
 
-    var componentNums = Object.keys(edits).map(Number).sort(),
+    var componentNums = Object.keys(edits).map(Number).sort(function(a, b) { return a - b; }),
         componentArrayIn = np.get(),
         componentArray = componentArrayIn || [],
         // componentArrayFull is used just to keep splices in line between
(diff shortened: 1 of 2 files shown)


src/plot_api/manage_arrays.js
```diff
@@ -101,7 +101,7 @@ exports.applyContainerArrayChanges = function applyContainerArrayChanges(gd, np,
         return true;
     }
 
-    var componentNums = Object.keys(edits).map(Number).sort(),
+    var componentNums = Object.keys(edits).map(Number).sort(function(a, b) { return a - b; }),
```
etpinard: You might want to use [`Lib.sorterAsc`](https://github.com/plotly/plotly.js/blob/ce970f112a8fff0d09909529be8a713f5d4f36cc/src/lib/search.js#L59) to DRY it up :palm_tree:

test/jasmine/tests/annotations_test.js
```diff
@@ -223,6 +223,33 @@ describe('annotations relayout', function() {
         .then(done);
     });
 
+    it('should sort correctly when index>10', function(done) {
```
etpinard: nice test!

etpinard:
:dancer: after that little bit of :palm_tree:

src/plot_api/manage_arrays.js
alexcjohnson: ah thanks - should've guessed we'd have that already exposed somewhere!

etpinard:
:dancer:

PR #1641, opened 2017-05-01
Fix scattercarpet calc

rreusser:
This addresses @alexcjohnson's comment that scattercarpet was doing something weird and not handling array properties correctly. I mostly copied scattercarpet from scatterternary, but I wasn't quite able to reconstruct the removed line in this PR. I replaced it with `arraysToCalcdata` and it seems like things work fine though.

### DIFF
--- a/src/traces/scattercarpet/calc.js
+++ b/src/traces/scattercarpet/calc.js
@@ -12,10 +12,10 @@
 var isNumeric = require('fast-isnumeric');
 
 var Axes = require('../../plots/cartesian/axes');
-var Lib = require('../../lib');
 
 var subTypes = require('../scatter/subtypes');
 var calcColorscale = require('../scatter/colorscale_calc');
+var arraysToCalcdata = require('../scatter/arrays_to_calcdata');
 var lookupCarpet = require('../carpet/lookup_carpetid');
 
 module.exports = function calc(gd, trace) {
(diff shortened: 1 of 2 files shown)


etpinard:
That looks right. I should've caught that while reviewing that carpet PR. My mistake.

@rreusser would you mind a test case to :lock: this down? Something similar to this [suite](https://github.com/plotly/plotly.js/blob/ce970f112a8fff0d09909529be8a713f5d4f36cc/test/jasmine/tests/scattergeo_test.js#L82) and more precisely this [case](https://github.com/plotly/plotly.js/blob/ce970f112a8fff0d09909529be8a713f5d4f36cc/test/jasmine/tests/scattergeo_test.js#L174) would more than suffice.

rreusser:
I'm currently on my way to plotcon and accidentally checked my iPhone cable so I have no wifi today. Anyone is free to seize control of this PR if the fix is needed urgently, otherwise I'll lock this down asap.

alexcjohnson:
Added a test (I want this for my gradient test cases). I did it through a full plot since I wanted to test `restyle` too. I actually thought it might not work on restyle, due to [this line](https://github.com/plotly/plotly.js/blob/master/src/plot_api/subroutines.js#L320) and the fact we didn't make `module.arraysToCalcdata` like we did in [some modules](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/index.js#L26)... but, test passes (as does a sanity check in the dashboard) so I think we're OK, just may point to an optimization we can do later for style changes like this, perhaps whenever we get around to overhauling restyle/relayout.

etpinard:
Oh that arraysToCalcdata check is a thing of the past :hocho:

Restyle now does a `recalc` whenever an arrayOk attribute is or is set to an array.

etpinard:
But yeah, it might be better to :hocho: that in a different PR. 

:dancer:

PR #1652, opened 2017-05-08
Annotationclick event

alexcjohnson:
@etpinard quick one to add the native mouse event to `plotly_clickannotation`


etpinard:
:dancer: nicely done

Issue #1649, opened 2017-05-04
restyle zmin & zmax doesn't cause contour maps to redraw


alexcjohnson:
1. open http://localhost:3000/devtools/test_dashboard/#electric_heatmap - changing colorscale works correctly:
```
Plotly.restyle(gd, {zmin: -50, zmax: 150})
```
2. Instead, first change to a contour map and change zmin/zmax - we don't see the new range:
```
Plotly.restyle(gd, {type: 'contour', zauto: true})
Plotly.restyle(gd, {zmin: -50, zmax: 150})
```
3. but then if you cause a redraw (doubleclick or something) then you do see the new range.

PR #1653, opened 2017-05-08
Contour restyle zmin zmax

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1649

So, turns out that restyling `zmin` and `zmax` for contour traces has been broken since `v1.0.0` and nobody noticed :expressionless:  

This PR fixes this by making `zmin` and `zmax` redo calcdata as the contour [calc](https://github.com/plotly/plotly.js/blob/2105fd48663b616503337518bab28621c1804497/src/traces/contour/calc.js#L26) does use it. Note that this isn't the case for `heatmap` traces where `zmin` and `zmax` are only used in the [plot](https://github.com/plotly/plotly.js/blob/2105fd48663b616503337518bab28621c1804497/src/traces/heatmap/plot.js#L174) step.

cc @alexcjohnson

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1624,8 +1624,8 @@ function _restyle(gd, aobj, _traces) {
                     flags.docalc = true;
                 }
 
-                // some attributes declare a 'recalc' flag
-                if(valObject.recalc) {
+                // some attributes declare an 'editType' flaglist
+                if(valObject.editType === 'docalc') {
                     flags.docalc = true;
                 }
 
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1624,6 +1624,11 @@ function _restyle(gd, aobj, _traces) {
                     flags.docalc = true;
                 }
 
+                // some attributes declare a 'recalc' flag
+                if(valObject.recalc) {
+                    flags.docalc = true;
+                }
+
                 // all the other ones, just modify that one attribute
                 param.set(newVal);
             }
--- a/src/traces/contour/attributes.js
+++ b/src/traces/contour/attributes.js
@@ -128,7 +128,10 @@ module.exports = extendFlat({}, {
         })
     }
 },
-    colorscaleAttrs,
-    { autocolorscale: extendFlat({}, colorscaleAttrs.autocolorscale, {dflt: false}) },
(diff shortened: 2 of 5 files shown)


src/plot_api/plot_api.js
```diff
@@ -1624,6 +1624,11 @@ function _restyle(gd, aobj, _traces) {
                     flags.docalc = true;
                 }
 
+                // some attributes declare a 'recalc' flag
+                if(valObject.recalc) {
+                    flags.docalc = true;
```
etpinard: ... slowly migrating to something better than those ugly `recalcAttrs` list in restyle and relayout.

alexcjohnson: Down the line, this is going to require `valObject.dostyle`, `.doplot`, `.doCamera` etc etc... would we be better off with something like `valObject.editType = 'docalc'` etc? Which presumably we could use like a `flaglist` (and validate as such in `plotschema_test`)?

💯 🍻  for getting this effort going!

etpinard: > valObject.editType = 'docalc'

Right. I guess that scales better. Thanks :+1:

etpinard: done in 413e545

src/plot_api/plot_api.js
```diff
@@ -1624,6 +1624,11 @@ function _restyle(gd, aobj, _traces) {
                     flags.docalc = true;
                 }
 
+                // some attributes declare an 'editType' flaglist
+                if(valObject.editType === 'docalc') {
+                    flags.docalc = true;
+                }
```
alexcjohnson: 👍 
I guess later this can change to something like 
```js
if(valObject.editType) {
    valObject.editType.split('+').forEach(function(flag) {
        flags[flag] = true;
    });
}
```
But don't need to do that now - lets wait until we have validation of `editType`, perhaps in plotschema_test?

alexcjohnson:
Looks good! 💃

src/plot_api/plot_api.js
etpinard: > But don't need to do that now - lets wait until we have validation of editType, perhaps in plotschema_test?

Yep, that's what I was thinking. Referencing https://github.com/plotly/plotly.js/issues/648

PR #1655, opened 2017-05-08
OHLC equal open close cases fix

etpinard:
supersedes @noway's https://github.com/plotly/plotly.js/pull/1619 with tests and lint fixes.

@noway does the test case look ok to you?


src/traces/ohlc/helpers.js
```diff
@@ -95,14 +95,65 @@ exports.makeTransform = function(traceIn, state, direction) {
...
+function _getFilterFn(direction) {
+    var isPrevThisDirection = null;
+    var cPrev = null;
+    var fn;
+
     switch(direction) {
         case 'increasing':
-            return function(o, c) { return o <= c; };
+            fn = function(o, c) {
+                if(o === c) {
```
alexcjohnson: We haven't cast to numbers, have we? Looks like this will break if we mix strings and numbers (or use different string representations of the same number, like `'2.0'` and `'2.00'`

I guess a point should get dropped if *any* of o, h, l, c is not numeric?

src/traces/ohlc/helpers.js
```diff
@@ -95,14 +95,65 @@ exports.makeTransform = function(traceIn, state, direction) {
...
+                            return true; // determine by last candle
+                        } else if(isPrevThisDirection === false) {
+                            return false; // determine by last candle
+                        } else {
+                            return false; // If we don't have previous data, assume it was increasing
+                        }
+                    }
+                }
+                return o > c;
+            };
```
alexcjohnson: This is a lot of duplication with just a few `true` <-> `false` flips... looks like we could keep the perf up without the duplication if we made one function `isIncreasing(o, c)`, which looks at a state var `isPrevIncreasing` (which would also simplify the logic there as we wouldn't need to deal with `null`, just initialize it to `true`), then move the switch logic into creating the wrapper function below:
```
function increasingFilter(o, c) {
    var isThisIncreasing = isIncreasing(o, c);
    isPrevIncreasing = isThisIncreasing; // why add !! to this?
    cPrev = c;
    return isThisIncreasing;
}

function decreasingFilter(o, c) {
    var isThisIncreasing = isIncreasing(o, c);
    isPrevIncreasing = isThisIncreasing; // why add !! to this?
    cPrev = c;
    return !isThisIncreasing; // the only difference...
}
```

etpinard: @alexcjohnson I vaguely remember you mentioning on slack that you had a commit implementing your solution on a branch. If that's not case, I'll try to push something up to get this PR in this week's `1.28.0`.

alexcjohnson: simplified these filters (even more than ^^) in https://github.com/plotly/plotly.js/pull/1655/commits/9575ab65178ef66bb0265391af3d43283f34bf9a

src/traces/ohlc/helpers.js
alexcjohnson: checked all to ensure they're numeric in https://github.com/plotly/plotly.js/pull/1655/commits/597adf96181b40a4eabc7b3622e919d19e1864a0

alexcjohnson:
hah, the [test failure](https://circleci.com/gh/plotly/plotly.js/4485?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) is a real bug that's fixed by https://github.com/plotly/plotly.js/pull/1655/commits/597adf96181b40a4eabc7b3622e919d19e1864a0
old version has a light green bar:
screen shot 2017-06-08 at 4 50 44 pm
new version it's pink:
screen shot 2017-06-08 at 4 50 20 pm
because `'99' > '100'` while `99 < 100`

etpinard:
> hah, the test failure is a real bug that's fixed by 597adf9

Nice :tada:

src/traces/ohlc/helpers.js
```diff
@@ -95,14 +97,38 @@ exports.makeTransform = function(traceIn, state, direction) {
...
+        return isPrevIncreasing;
     }
-};
+
+    function isIncreasing(o, c) {
+        return isNumeric(o) && isNumeric(c) && _isIncreasing(+o, +c);
+    }
+
+    function isDecreasing(o, c) {
+        return isNumeric(o) && isNumeric(c) && !_isIncreasing(+o, +c);
```
etpinard: that `_!isIncreasing` statement is so brilliant :sparkles:

etpinard:
:dancer:

Thanks very much for taking this one to the finish line :racing_car: :checkered_flag:

noway:
Haven't had a chance to look into it all, but thank you for seeing it all through 👍

Issue #1648, opened 2017-05-04
plotly_click and plotly_select buggy on scatter/markers plot


r1z1a:
I made a codepen here:
[Codepen](https://codepen.io/anon/pen/rmzgda?editors=0011)

plotly_click:
So when you click on a given point, it only returns one object even when there is 10 points at each of the 3 locations. Does it not register the other points, that are also with same x and y coordinates?

plotly_selected:
As you can see when you select(box or lasso) any given point or set of points, the total number of objects always appends the NA items into the array. Also the selection does not give more than just x,y, pointNumber and curveNumber the actual data for each point is not passed into plotly_select.

etpinard:
> So when you click on a given point, it only returns one object even when there is 10 points at each of the 3 locations. Does it not register the other points, that are also with same x and y coordinates?

That's been discussed in https://github.com/plotly/plotly.js/issues/720#issuecomment-230900670 - it is the intended behavior I'm afraid.

> As you can see when you select(box or lasso) any given point or set of points, the total number of objects always appends the NA items into the array

Oh. That's a bug. Thanks for discovering it!

> Also the selection does not give more than just x,y, pointNumber and curveNumber the actual data for each point is not passed into plotly_select.

Ok, but what else do you want it there?

etpinard:
> As you can see when you select(box or lasso) any given point or set of points, the total number of objects always appends the NA items into the array

will be via https://github.com/plotly/plotly.js/pull/1656

r1z1a:
> That's been discussed in #720 (comment) - it is the intended behavior I'm afraid.
Yeah I suspected that this is set in stone and would make sense since not everyone would want to return multiple click events from hitting one point. 
I saw that someone has made a topic about adding "jitter" functionality which would solve the issue since you would be able to zoom in very close and points would not be superimposed #299

> Oh. That's a bug. Thanks for discovering it!
No problem, its better that I rewrite my code once the fix is out, than other people struggling with this problem.

> Ok, but what else do you want it there?
Stemming back from first issue where _click will not allow you to hit multiple points and, and since there is no jittering I'm using the _select to pick up the points that are below the top one. It would be nice if it also returned data, like plotly_click. Although I understand you can get the data for whole graph from the div and then go though each point and get the info that way. 
Just thought it would be nice to have consistence between the 2 events since they seem similar in that way. Maybe just for me and way I use them.

Thanks for quick response

etpinard:
[...]

PR #1564, opened 2017-04-07
Scattermapbox BADNUM (w/o upgrading mapbox-gl)

etpinard:
replacement for https://github.com/plotly/plotly.js/pull/1543 to be merged in #1538.

After four days of trying to get newer versions of `mapbox-gl` to work in `nw.js`, I decided it wasn't worth the trouble. So, in brief, this PR:

- makes scattermapbox/calc.js reuse scattergeo/calc.js
- :hocho: the last remaining signs of calcdata placeholder
- will make the test :white_check_mark: in #1538 
- (had to) keep the current hacky convert logic in `scattermapbox/convert.js` :disappointed:

PR #1656, opened 2017-05-08
Skip BADNUM items in scatter select routine

etpinard:
closes https://github.com/plotly/plotly.js/issues/1648

This PR first replaces `{ x: false, y: false }` scatter calcdata items with `{ x: BADNUM, y: BADNUM }` as put forward in https://github.com/plotly/plotly.js/pull/1564/commits/b5cd57308c60b584ba47ba93b35bb02efdd15418 then makes sure to skip those BADNUM items (as opposed to casting them to numbers - see codepen in https://github.com/plotly/plotly.js/issues/1648) in the scatter select routine.

cc @alexcjohnson it would be nice to include this in `1.27.0` if you get the time to review it :smile:

### DIFF
--- a/src/traces/scatter/select.js
+++ b/src/traces/scatter/select.js
@@ -10,6 +10,7 @@
 'use strict';
 
 var subtypes = require('./subtypes');
+var BADNUM = require('../../constants/numerical').BADNUM;
 
 var DESELECTDIM = 0.2;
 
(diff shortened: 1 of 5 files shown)


src/traces/scatter/select.js
```diff
@@ -40,6 +41,11 @@ module.exports = function selectPoints(searchInfo, polygon) {
             di = cd[i];
             x = xa.c2p(di.x);
             y = ya.c2p(di.y);
+
+            if(x === BADNUM || y === BADNUM) {
+                continue;
+            }
```
alexcjohnson: I take it from this fix that `polygon.contains` fails for undefined inputs? I would think for future extensibility it would be better to fix it in there, so if either `x` or `y` is `BADNUM` it's not contained in any polygon.

etpinard: Good call. Done in https://github.com/plotly/plotly.js/pull/1656/commits/80cee013a0a6b70092fc1ecb3ac7ba6ed812d81d

alexcjohnson:
💃

Issue #1316, opened 2017-01-18
[WIP] Crossfilter discussion


monfera:
# Reactive, crossfiltered data visualization

Plotly has originally focused on generating visualizations, and interactivity increased over time. Plotly, by now, has acquired rich layout, style and data update facilities, even animations. Data transformations such as declarative input grouping and filtering have also been added.

As there's growing expectations for fluid, efficient, yet still declarative interactions such as [crossfiltering](http://crossfilter.github.io/crossfilter/), we are starting a discussion with the purpose of shaping an API in line with Plotly conventions, current practices and future expectations. 

Crossfilters are behaviors that let the user _subset a multivariate dataset via direct manipulation across multiple views on that dataset_. It is also known as linked brushing or linked filtering. The set of views included in one crossfilter is called _coordinated views_ by the crossfilter.js doc, or sometimes _linked views_. There is no clear-cut boundary for the functional scope and features of crossfilters.

The archetypal crossfilter example by Mike Bostock, author of crossfilter.js, showing multidimensional filtering and aggregation on a quarter million records, also updating a sample table:
![crossfilterjs](https://cloud.githubusercontent.com/assets/1548516/22087839/98730a70-dde0-11e6-897d-0f48afa19708.gif)

This text is just to start the ball going. There is prior art, surrounding the Plotly toolchain and its dependencies such as D3. Since these tools are in active use and well-documented, this description doesn't detail them, except enlisting them and highlighting some of their properties.

Also, there is a [fantastic discussion](https://cpsievert.github.io/plotly_book/linking-views-without-shiny.html) on the topic by Carson Sievert, including many of the crossfilter concepts. Due to the richness of that material, this writeup can be a bit sparser on the crossfilter behaviors and more detailed on implementation concerns.

It's still useful to start with one way of thinking about interactivity, as crossfiltering is a particular instance of it. Also, crossfilter cores such as `crossfilter.js` usually peer-depend on change propagation or [reactivity](https://en.wikipedia.org/wiki/Reactive_programming). Section 1 may be skipped for directly jumping into the crossfilter-specific part. 

# reactivity

## What runtime changes may occur to a visualization?

Not all types of visualizations require sophisticated updates. For example, a command tool such as the typical use of `ggplot2` is technically a single-step execution even if the dataviz maker may repeatedly invoke it with various projections, aesthetics and data. These are common things that need data flow:
[...]

PR #1657, opened 2017-05-09
Scattergl lasso

dfcreative:
Enables svg interactions for scattergl plot.
Some minor issues TODO:
* [x] zooming loses selection
* [x] autoscale loses `cdata.plot`
* [x] ~~opacity of huge number of points is indistinguishable from non-selected points~~
* [x] blinking of some points on selection
* [x] picking is lost
* [x] fix tests
* [x] low fps on selection (ok at 1e5)
* [x] ~~panning does not clear selection~~
* [x] add tests

Related https://github.com/dfcreative/plotly-contrib/issues/3


jackparmer:
This will be useful for the `crossfilter` project too, where large D3 line and scatter plots may not be zippy enough: https://github.com/plotly/plotly.js/issues/1316

jackparmer:
Hey @dfcreative will `scattergl` have `plotly_selected` and `plotly_deselect` events identical to its SVG counterpart? 
http://codepen.io/plotly/pen/BpayyX

etpinard:
> Hey @dfcreative will scattergl have plotly_selected and plotly_deselect

yes,

etpinard:
> Hey @dfcreative will scattergl have plotly_selected and plotly_deselect

yes,

src/plots/cartesian/dragbox.js
```diff
@@ -79,7 +79,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         var yIDs = [ya0._id];
 
         // if we're dragging two axes at once, also drag overlays
-        subplots = [plotinfo].concat((ns && ew) ? plotinfo.overlays : []);
+        subplots = [plotinfo].concat((ns && ew && plotinfo.overlays) ? plotinfo.overlays : []);
```
etpinard: Why do we need the extra `&& plotinfo.overlays` now?

src/plots/gl2d/index.js
```diff
@@ -82,6 +81,47 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
             delete oldFullLayout._plots[id];
         }
     }
+
+    //since we use cartesian interactions, do cartesian clean
+    cartesian.clean.apply(this, arguments)
+};
+
+exports.drawFramework = function(gd) {
```
etpinard: Nice. We probably don't all of this or at least we should be able to reuse the `Cartesian.drawFramework` directly as:

```js
exports.drawFramework = Cartesian.drawFramework;
```

src/plots/gl2d/index.js
```diff
@@ -108,3 +148,159 @@ exports.toSVG = function(gd) {
...
+        id = plotinfo.id;
+
+    // Layers to keep plot types in the right order.
+    // from back to front:
+    // 1. heatmaps, 2D histos and contour maps
+    // 2. bars / 1D histos
+    // 3. errorbars for bars and scatter
+    // 4. scatter
+    // 5. box plots
+    function joinPlotLayers(parent) {
```
etpinard: ... e.g all of these are useless in gl2d

src/plots/gl2d/scene2d.js
```diff
@@ -359,7 +358,6 @@ proto.destroy = function() {
 
     if(!this.staticPlot) this.container.removeChild(this.canvas);
     this.container.removeChild(this.svgContainer);
-    this.container.removeChild(this.mouseContainer);
```
etpinard: Hmm. That probably breaks [`gl2d/camera.js`](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/src/plots/gl2d/camera.js#L34)

etpinard: So wait, are we :hocho: the gl2d picking routine?

It would be nice to merge the cartesian and gl2d hover logic at some point, but swapping in cartesian hover in gl2d probably means poor performance on graphs with 1e5 and up data points at this stage.

src/traces/scattergl/attributes.js
```diff
@@ -84,5 +84,17 @@ module.exports = {
     fillcolor: scatterAttrs.fillcolor,
 
     error_y: scatterAttrs.error_y,
-    error_x: scatterAttrs.error_x
+    error_x: scatterAttrs.error_x,
+
+    hoveron: {
```
etpinard: This has no effect in gl2d. You must have added this in to make other things not break, correct?

src/traces/scattergl/convert.js
```diff
@@ -511,7 +593,11 @@ proto.updateFancy = function(options) {
             this.scatter.options.borderWidths[i] = 0.5 * borderWidths[index];
 
             for(j = 0; j < 4; ++j) {
-                this.scatter.options.colors[4 * i + j] = colors[4 * index + j];
+                var color = colors[4 * index + j];
+                if(sel && !sel[index] && j === 3) {
+                    color *= DESELECTDIM;
+                }
+                this.scatter.options.colors[4 * i + j] = color;
```
etpinard: Very nice. Thanks for covering both `updateFast` and `updateFancy` !!

src/traces/scattergl/index.js
```diff
@@ -13,10 +13,12 @@ var ScatterGl = {};
 ScatterGl.attributes = require('./attributes');
 ScatterGl.supplyDefaults = require('./defaults');
 ScatterGl.colorbar = require('../scatter/colorbar');
+ScatterGl.hoverPoints = require('../scatter/hover');
 
 // reuse the Scatter3D 'dummy' calc step so that legends know what to do
-ScatterGl.calc = require('../scatter3d/calc');
+ScatterGl.calc = require('../scatter/calc');
 ScatterGl.plot = require('./convert');
+ScatterGl.selectPoints = require('./select');
```
etpinard: :tada:

src/traces/scattergl/index.js
```diff
@@ -13,10 +13,12 @@ var ScatterGl = {};
 ScatterGl.attributes = require('./attributes');
 ScatterGl.supplyDefaults = require('./defaults');
 ScatterGl.colorbar = require('../scatter/colorbar');
+ScatterGl.hoverPoints = require('../scatter/hover');
 
 // reuse the Scatter3D 'dummy' calc step so that legends know what to do
-ScatterGl.calc = require('../scatter3d/calc');
+ScatterGl.calc = require('../scatter/calc');
```
etpinard: ... but that might slow down `scattergl` for 1e5+ point graphs.

src/plots/cartesian/axes.js
```diff
@@ -365,7 +365,11 @@ axes.doAutoRange = function(ax) {
     var hasDeps = (ax._min && ax._max && ax._min.length && ax._max.length);
 
     if(ax.autorange && hasDeps) {
-        ax.range = axes.getAutoRange(ax);
+        // FIXME: scattergl calls it in async fashion, hence there is .range and _._rl mess
+        ax.range = ax._rl = axes.getAutoRange(ax);
+
+        ax._r = ax.range.slice();
+        ax._rl = Lib.simpleMap(ax._r, ax.r2l);
```
etpinard: This is an interesting patch. I wonder if this might be fixing an undiscovered bugs(s).

dfcreative: That is taken from https://github.com/dfcreative/plotly.js/blob/6de41e6d555d038299aa0aef7e075804f011db49/src/plots/cartesian/axes.js#L1657, just optimized it a drop

src/traces/scattergl/attributes.js
dfcreative: Picking was broken without `hoveron` property

src/traces/scattergl/index.js
etpinard: For the record, we must make sure plotting 1e6 points with `scattergl` (passing through `updateFast`) clocks in at around 1 second.

src/plots/gl2d/index.js
```diff
@@ -12,23 +12,22 @@
 var Scene2D = require('./scene2d');
 var Plots = require('../plots');
 var xmlnsNamespaces = require('../../constants/xmlns_namespaces');
-
+var constants = require('../cartesian/constants');
+var cartesian = require('../cartesian')
```
dfcreative: Done

src/plots/gl2d/index.js
dfcreative: Done. Gotta take a second look on optimization.

src/plots/gl2d/index.js
dfcreative: Done.

src/plots/gl3d/camera.js
```diff
@@ -179,8 +180,24 @@ function createCamera(element, options) {
         return false;
     });
 
-    var lastX = 0, lastY = 0;
-    mouseChange(element, function(buttons, x, y, mods) {
+    var lastX = 0, lastY = 0, lastMods = {shift: false, control: false, alt: false, meta: false};
+    mouseChange(element, handleInteraction);
+
+    // enable simple touch interactions
```
dfcreative: That is unmerged https://github.com/plotly/plotly.js/pull/1507 here

src/traces/scattergl/index.js
dfcreative: That is required to obtain selection in [scattergl/select](https://github.com/dfcreative/plotly.js/blob/6de41e6d555d038299aa0aef7e075804f011db49/src/traces/scattergl/select.js#L15). Mb it is possible to optimize that, just haven't got that a chance.

dfcreative: Actually I see no speed difference compared to old implementation for 1e6.

etpinard: Great. I'm surprised, but that's great news :tada:

src/plots/gl2d/scene2d.js
dfcreative: The problem is that `mouseContainer` intercepts mousemove, needed for cartesian selection to work. There seems to be no good way to hook up [gl2d/camera](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/src/plots/gl2d/camera.js#L34) so to enable [cartesian/select](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/src/plots/cartesian/select.js#L24) − these are too separate concerns.
It is possible to switch to cartesian hover only for lasso/select mode, but the pick style differs visually, which is confusing.

dfcreative:
Ok, the progress so far:

* stable interactions, selection works the same way as cartesian svg ones
* 1e6 points are rendered fast, but svg-picking is visually a bit slower than gl2d one
* svg hover does not have spikes, therefore gl2d spikes are lost
* blinking points seems to be caused by [snap-points-2d](https://www.npmjs.com/package/snap-points-2d) within gl-scatter2d

Opacity, fps and panning seems to be features/bugs of cartesian scatter, probably should be addressed separately.

etpinard:
Ok great. Thanks for the details @dfcreative! 

My gut feeling is we shouldn't :hocho: gl2d hover at this stage. It's very much a breaking change in my mind that would have to wait for v2. 

Spikes are now avaliable for svg cartesian, so maybe we hook those in. Oh well, I'm sure we can make gl2d hover and cartesian selection coexist (probably by adding a few if-gl2d early returns in fx/hover and cartesian/graph_interact). I can help you out next week if you want :+1:

dfcreative:
@etpinard that would be great, I am really having a hard time figuring out how to do that.

dfcreative:
Hm just a thought - what if we get rid of `scattergl` at all by reusing [kd-tree](https://github.com/mourner/kdbush) structure for `scatter`? That would allow for rendering 1e6+ custom points with d3.

The principle is illustrated in [supercluster](https://github.com/mapbox/supercluster):
supercluster zoom

An article on that: https://medium.com/@agafonkin/a-dive-into-spatial-search-algorithms-ebd0c5e39d2a.

We already use similar thing, [snap-points-2d](https://www.npmjs.com/package/snap-points-2d) for that in gl-scatter2d, which is unfortunately not that efficient.

@etpinard is there any other purpose of scattergl besides rendering big amount of points?

src/traces/scattergl/index.js
etpinard: @dfcreative can you share benchmark script with me (e.g. via a GitHub gist or codePen)? I'm seeing some a significant slow down.

dfcreative: I have not benchmarked it, just impression from https://gist.github.com/dfcreative/bbf862a685c3f031c44d652a2e906298

etpinard: Cool. Thanks for the gist!

dfcreative:
@etpinard demo for 1e6 points https://dfcreative.github.io/point-cluster/ with canvas2d
Does not look too convincing :confused:

dfcreative:
Ok, figured out that [snap-points-2d here](https://github.com/gl-vis/gl-scatter2d/blob/master/scatter2d.js#L69) is the main perf bottleneck - it recalculates scale index every mousemove.

src/traces/scattergl/index.js
etpinard: Ok. After some more thorough speed tests, making `scattergl` pass through `scatter/calc.js` adds roughly 1 full second to `Plotly.plot` at 1e6 points (clocking in the 2500-3000ms range as opposed to ~1500ms on `master`) on my machine. Moreover, reusing the calcdata-based cartesian hover routines make picking noticeably slower at 1e6 points. 

In brief, we can't go through with this approach. If `scattergl` is slow with 1e6 points, it loses its _raison d'être_.

It worth mentioning that at 1e5, the modifications in this branch perform very (and to my surprise) well. Still, millions of points is what `scattergl` is all about.

So, what should we do? I'm thinking we should start by reusing [`cartesian/select.js`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/select.js) and make its output listen to events on the `gl2d` mouse container (that this PR removed :smirk: ) and draw the SVG lasso/rectangle path above the gl canvas. From there, it shouldn't be too hard to modify `scattergl/select.js` to use the selection indices to dig in `fullData.x` and `fullData.y` (as the calcdata won't match the coordinates w/o calling the slow `scatter/calc.js`). 

@dfcreative let me know if this makes sense or if you have any questions.

dfcreative: @etpinard we can effortlessly get the `mouseContainer` back, it is just a matter of 2 commented LOC.
I will try to optimize _snap-points-2d_ in _gl-scatter2d_ update, I feel optimistic about that, it should add decent selection performance, because now it regenerates hierarchical cache every draw frame for a millions of points.
As for modifying [cartesian/select](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/select.js) that implies implementing part of [gl2d/camera](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/src/plots/gl2d/camera.js) in it, but that seems to be the only choice if we want to keep gl picker.

etpinard: > we can effortlessly get the mouseContainer back, it is just a matter of 2 commented LOC.

Ok, but the mouseContainer sits on top of the cartesian drag layers, so how will the selection handler gets its data from?

> I will try to optimize snap-points-2d in gl-scatter2d update,

That's cool, but the real bottleneck is in `scatter/calc.js` when comparing this branch to `master`.

src/traces/scattergl/select.js
```diff
@@ -0,0 +1,70 @@
...
+
+    if(polygon === false) { // clear selection
+        for(i = 0; i < cd.length; i++) cd[i].dim = 0;
+    }
+    else {
+        for(i = 0; i < cd.length; i++) {
+            di = cd[i];
+            //FIXME: this affects performance for 1e6 points
+            x = xa.c2p(di.x);
+            y = ya.c2p(di.y);
```
alexcjohnson: So... the only things you need from `scatter.calc` are `di.x` and `di.y` right? There's a whole lot that could be cut out from `scatter.calc` then - like, probably 90% of the code, though I'm not sure what fraction of time this corresponds to if `x` and `y` are the only arrays in the trace - but I bet at least removing `Axes.expand` is a decent savings.

Perhaps we could also stash these pixel positions somewhere, so `c2p` only has to be called on the first `selectPoints` after any zoom/pan. Probably `polygon.contains` is the more expensive operation much of the time, and you'd have to be careful about what conditions would invalidate these pixel values, but could be a worthwhile optimization.

It may be a bigger project than you want to take on right now, but I suspect the longer-term solution is to not create a `cd` at all (or just a dummy if you need it to pass a reference to `trace` along here) and do something more like @monfera was describing parcoords doing, and have this picking happen on the GPU. Take this part as speculation since I haven't worked with GPU code, but something like:
- stash pixel positions in x and y as arrays on the GPU during rendering, so the picker loop doesn't have to do any `c2p` calculations
- reimplement `polygon.contains` on the GPU
- pass a selection array back to the CPU for emitting events

src/traces/scattergl/index.js
dfcreative: @etpinard I reduced calc time ~10times, so for `1e6` points it takes ~200ms on my laptop to load.

etpinard: Ok. What about:

```js
var x = [];
var y = [];
var size = []

for( var i = 0; i < 1e6; i++)
{
  x.push(i);
  y.push(i*Math.random());
  size.push(15 * Math.random())
}

var scatterglMarkers = {
  x: x, y: y, type: "scattergl", mode: "markers",
  // marker: {size: size}
};

console.time('scattergl')
Plotly.newPlot('graph',  [scatterglMarkers], {})
.then(() => console.timeEnd('scattergl'))
```

dfcreative: @etpinard ~3300ms vs ~3500ms for 1e6 points

src/traces/scattergl/calc.js
```diff
@@ -0,0 +1,38 @@
...
+
+    var serieslen = Math.min(x.length, y.length), i;
+
+    // create the "calculated data" to plot
+    var cd = new Array(serieslen);
+    for(i = 0; i < serieslen; i++) {
+        cd[i] = {x: x[i], y: y[i]};
+    }
+
+    arraysToCalcdata([{x: false, y: false, trace: trace, t: {}}], trace);
```
alexcjohnson: you've shortened it to one point... but is `arraysToCalcdata` even necessary?

src/traces/scattergl/index.js
etpinard: That's ~3300ms on `master`?

src/traces/scattergl/calc.js
```diff
@@ -0,0 +1,38 @@
...
+    var x = xa.makeCalcdata(trace, 'x'),
+        y = ya.makeCalcdata(trace, 'y');
+
+    var serieslen = Math.min(x.length, y.length), i;
+
+    // create the "calculated data" to plot
+    var cd = new Array(serieslen);
+    for(i = 0; i < serieslen; i++) {
+        cd[i] = {x: x[i], y: y[i]};
+    }
```
alexcjohnson: you might even be able to get rid of this loop... it would take a bit of reshaping in `scattergl/select` but can you just say `cd = [{x: x, y: y}]`? Again, if `select` is the only consumer of `cd` you can put whatever you want in there. Regular scatter is much more constrained because it's constructing `cd` as the d3 data for the trace.

dfcreative: @alexcjohnson there seems to be some intermediary or consumer relying on `cd` of that form, selection just does not work properly that way
```js
var x = cd[0].x, y = cd[0].y
var serieslen = Math.min(x.length, y.length), i;
for (i = 0; i < serieslen; i++) {
    var px = xa.c2p(x[i]), py = xa.c2p(y[i])
    if(polygon.contains([px, py])) {
        selection.push({
            x: px,
            y: py,
        });
    }
}
```
image

src/traces/scattergl/index.js
dfcreative: @etpinard just updated master: ~3000ms vs ~3200ms, I hardly notice the difference due to fluctuations

src/traces/scattergl/calc.js
alexcjohnson: OK, seems like it's not the bottleneck now but at some point we may want to track this down.

src/traces/scattergl/index.js
etpinard: Wow. That's brutal. Does anyone else have to wait `3000ms` when plotting 1e6 points?

src/traces/scattergl/select.js
monfera: It later occurred to me that, unlike with the point picking, the lasso selection needs to be inclusive of all points, not just the visible points, so the raster based approach wouldn't work out. The rectangular selection is still fairly easy on the GPU, simply by checking whether `value === clamp(value, min, max)` as in `parcoords` but the freeform selector would need something like the even-odd raycasting, quadtree or some such...

jackparmer: Should we limit `scattergl` to rectangular selection? That sounds reasonable to me.

src/traces/scattergl/convert.js
```diff
@@ -9,7 +9,7 @@
 
 'use strict';
 
-var createScatter = require('gl-scatter2d');
+var createScatter = require('../../../../gl-scatter2d');
```
etpinard: Can you push your `gl-scatter2d` work to a branch, so that we can work off the same source files?

src/traces/scattergl/select.js
etpinard: > Should we limit scattergl to rectangular selection?

I don't think we can.

cc @bpostlethwaite rob

monfera: yes it'd make the problem more tractable unless someone has a suggestion for cutting the Gordian knot of freestyle lasso, maybe @dfcreative @etpinard @rreusser has add'l ideas

monfera: [if we can access](https://www.khronos.org/registry/webgl/sdk/tests/conformance/rendering/vertex-texture-fetch.html) textures in the vertex shader then it'd be relatively simple to check if a particular point (vertex) is within or outside the rasterized lasso area represented as eg. black area in the texture, any experience @dfcreative @rreusser ?

dfcreative: @monfera could you pls elaborate why selecting only visible points won't work? If there're no plans on deleting/moving selected points, dealing only with visible ones seems to be reasonable. In case of zoom it is possible to reinit selection.

src/traces/scattergl/convert.js
dfcreative: Done

src/traces/scattergl/select.js
dfcreative: @monfera good idea, but that seems to belong to gl-scatter2d* component API. Implementing that within plotly infrastructure seems not elegant. That is mostly why I started [regl-scatter2d](https://github.com/dfcreative/regl-scatter2d)

alexcjohnson: Even-odd raycasting, with prefiltering ala `clamp`, is what we do in SVG land. It's [pretty painless](https://github.com/plotly/plotly.js/blob/master/src/lib/polygon.js#L86-L150), how hard would that really be on the GPU?

dfcreative: @alexcjohnson I imagine `scatter.select(bitmap|polygon|ids|fn)` function. Testing a point over texture in shader is a matter of masking, simply multiplication.

monfera: @dfcreative 

> could you pls elaborate why selecting only visible points won't work

For example, two points have the same exact x/y coordinate, one of them will be rendered on the top, and it'll be duly returned by the point picker (via a bijective rgb mapping) on mouseover, and the occluded point will not be reported - but for the area selector, it's expected that the bottom point is also recovered. This generalizes to arbitrary overlaps esp. that the current hover picker needs to work with the rendered markers in raster space while the area selector needs to check for inclusion of mathematical points (rather than marker pixels) in the lassoed area.

etpinard:
> 1e6 points are rendered fast, but svg-picking is visually a bit slower than gl2d one
> svg hover does not have spikes, therefore gl2d spikes are lost

> My gut feeling is we shouldn't  gl2d hover at this stage. It's very much a breaking change in my mind that would have to wait for v2.

I found a quick and dirty to allow us to use the current gl2d (fast) hover routine in parallel with cartesian lasso/select: draw the scene2d `mouseContainer` as before and set its `style['pointer-events'] = 'none'` when the lasso or select dragmodes are selected.

image
peek 2017-06-06 14-56

src/traces/scattergl/select.js
monfera: @alexcjohnson yes it's fairly simple as a CPU algo for a specific point, but the GPU needs to do it in parallel and it can't even do proper loops or index handling. The best analogy would be [ray marching](https://en.wikipedia.org/wiki/Volume_ray_casting) that underpins almost all [shadertoy](https://www.shadertoy.com/view/Xds3zN) creations, and it samples a 3D space populated with 3D shapes that are individually easy to describe mathematically (eg. modeling a head as a sphere and a neck as a cylinder, combining them via [CSG](https://en.wikipedia.org/wiki/Constructive_solid_geometry). Arbitrary 3D shapes can be built. The raymarching samples the 3D space by shooting rays as per the 2D pixel grid. 

It's possible to do a 2D variant of this; it'd require that the lassoed area is decomposed of mathematically easy to model primitives eg. triangles (so, just triangulation of the lassoed area). It uses [signed distance fields](http://iquilezles.org/www/articles/distfunctions/distfunctions.htm) with which there's ample experience on the team.

As it's not for artistic output but data analysis, I think significant time would go into a robust approach even though it's "only" 2D.

monfera:
@dfcreative 

> Ok, figured out that snap-points-2d here is the main perf bottleneck

yes indeed, btw. it's pretty zippy once the LoD layers are populated, but that's a pretty expensive operation that was done on the CPU, there was even a long, custom sorting algo that I rewrote but still too much latency - that's when we introduced `pointcloud`, it's not doing splatting but has a [quite low, 30ms latency on data change](https://codepen.io/monfera/pen/BLjJVZ)

dfcreative:
@monfera do you cluster points in pointcloud or hide invisible ones in any way? I played around with [kd-tree](https://github.com/dfcreative/kdgrass) and [point-cluster](https://github.com/dfcreative/point-cluster), the trees are ~30% faster on indexing 1e7 points, compared to snap-points-2d, also they are speedy on fetching bounded data, but I could not have figured out how to avoid "patterns" of fetched data:
moire1

monfera:
@dfcreative to ensure very low initial and data update latency (requirement was: below 100ms), `pointcloud` doesn't do anything like that, all 1m points are just sent to the GPU in attrib array buffers. Since it would be expensive (in the fragment shader) to paint over lots of overlapping points, the point size is set dynamically based on visible point count (which depends on zoom level and how crowded the view box is), and it's also helpful as points are always discernible. So ink/overpaint is limited this way, and 1m points render just fine too if the points are small.

etpinard:
@dfcreative here's my attempt at making the test pass:

https://github.com/dfcreative/plotly.js/compare/scattergl-lasso...plotly:scattergl-lasso-etienne

All jasmine tests now pass (locally at least). A few image tests fail though. I suspect your modifications in `cartesian/axes.js` are to blame. I'll figure out what went wrong.

Ok well, this PR is starting to look mergeable. Are you still working on improving performance? Anything I can do to help?

src/plots/gl2d/scene2d.js
etpinard: Probably resolved by 	1451802

src/traces/scattergl/attributes.js
etpinard: Resolved in 648547f from 

https://github.com/dfcreative/plotly.js/compare/scattergl-lasso...plotly:scattergl-lasso-etienne

@dfcreative you might want to merge that commit in too.

dfcreative:
@etpinard awesome! No, I think I've done everything possible for the current "iteration". Making selection faster would require solutions suggested by Robert and Alex, which is another story. Thank you!

etpinard:
> No, I think I've done everything possible for the current "iteration"

Ok. I'm not 100% ready to call this PR _satisfactory_ in terms of performance. I'll spend the day tomorrow trying to speed things up. I'll add a few tests while at it.

etpinard:
>  I'm not 100% ready to call this PR satisfactory in terms of performance.

So, with the latest commits, on my t450s Thinkpad, lasso/selectbox offer decent perf at 1e5 points:

peek 2017-06-06 17-19

at 1e6 points, things start to be a bit more sluggish especially for lasso (overall selectbox is faster than lasso). 

That's probably good enough for v1, right @jackparmer @bpostlethwaite @alexcjohnson ?

src/traces/scattergl/convert.js
etpinard: @dfcreative Something broke selection for _fancy_ traces e.g. with array `marker.size`:

peek 2017-06-06 17-26

```js
var x = [];
var y = [];
var size = []

for( var i = 0; i < 1e4; i++)
{
  x.push(i);
  y.push(i*Math.random());
  size.push(15 * Math.random())
}

var scatterglMarkers = {
  x: x, y: y, type: "scattergl", mode: "markers",
  marker: {size: size, opacity: 1}
};

console.time('scattergl')
Plotly.newPlot('graph',  [scatterglMarkers], {dragmode: 'lasso'})
.then(() => console.timeEnd('scattergl'))
```

etpinard: Did you get the time to investigate :arrow_heading_up: ? If not, I can look at it this afternoon.

etpinard: Well, looks like this was never implemented `this.selectedIds` in the `scene2d.updateFancy` scope doesn't reference anything.

dfcreative: @etpinard done

etpinard: Beautifully done. Thanks very much!

jackparmer:
> That's probably good enough for v1, right

As long as it's faster than picking with 10e5 points in SVG, I'm happy for the customer projects I'm working on (mainly crossfilter)

src/traces/scattergl/index.js
etpinard: With 6ac722d4d440bbff10568cecc46174076071f253 we're back in the ~1500ms for 1e6 first render as on `master` :tada: 

Most of `scattergl/calc.js` is now skipped when dragmode is set to pan/zoom (the default).

 Calling `Plotly.relayout` from pan/zoom to lasso/select now trigger a recalc where `gd.calcdata` is filled so that the selection logic knows what to do.

etpinard:
> As long as it's faster than picking with 10e5 points in SVG

You can't really plot 1e5 pts in SVG, so yeah, it's better.

monfera:
Started leaning on this branch as this feature is essential for crossfiltering if the number of points is over around 1k. No specific need for it now, but at some point, maybe we'd want to add the properties `curveNumber` and `pointNumber` to elements of the `eventData.points` array for feature parity w/ `scattergl`. These numbers can work for indexing `calcdata` eg. gd.calcdata[curveNumber][pointNumber] (speaking of this object, at some point it could have the `data` property that gets fed from `customdata`). The eventData points already have the `id` which is all I need right now.

monfera:
Quick perf check, it works well at 10k points as it's pushed rn:
scattergl-10k

It currently seems overly slow with 125k points where `color` is an array (haven't tested otherwise as crossfilter needs salient selected points):
scattergl-100k

Also, as seen on this 2nd animgif (also present with 10k points) the hover tooltip, for some reason, appears on initial render with no interaction (top left) and is stuck there even when mouse is moved (with or without `hovermode: 'closest'`).

monfera:
⬆️ cursory test shows that `readPixel` in 
```
function SelectBuffer(gl, fbo, buffer) {
  this.gl     = gl
  this.fbo    = fbo
  this.buffer = buffer
  this._readTimeout = null
  var self = this

  this._readCallback = function() {
    if(!self.gl) {
      return
    }
    fbo.bind()
    gl.readPixels(0,0,fbo.shape[0],fbo.shape[1],gl.RGBA,gl.UNSIGNED_BYTE,self.buffer)
    self._readTimeout = null
  }
}
```
is the culprit (slow irrespective of how many of the 125k points is currently in the visible, zoomed domain):
image

etpinard:
TODO:

- [ ] bump `gl-scatter2d-sdf` dep to include https://github.com/gl-vis/gl-scatter2d-sdf/pull/5

dfcreative:
@monfera may I ask you to help reproducing this bug with stuck hover tooltip? I faced it once, but it works fine rest of the time. Is some special case to that?

monfera:
@dfcreative not showing up on my current branch right now, I'll make a minimal example if it occurs again

monfera:
@dfcreative I don't have an isolated case, just a hint at this time - the tooltip seems to get stuck if I delete the `marker.opacity` line here:
```
Plotly.newPlot(
      plotlyChartHolder01,
      [{
        type: 'scattergl',
        mode: 'markers',
        marker: {
          size: 3,
          color: 'rgba(255,255,255, 0.25)',
          opacity: atomic.map(d => 1)
        },
        x: atomic.map(d => d.logValueMmUsd),
        y: atomic.map(d => d.logKilotons),
        ids: atomic.map(d => d.key),
      }],
      {
        hovermode: 'closest',
        dragmode: 'select',
        width: 500
      }
    )
```
In addition, when the `opacity` array is not present, and then do a rectangular lassoing (which is the initial state due to ` dragmode: 'select'`) there's an error complaining that `positions === null`. There are 2300 points. I can't work on a minimal case rn but thought to share it in case the above is a sufficient condition.

src/traces/scattergl/select.js
```diff
@@ -0,0 +1,60 @@
...
+        for(i = 0; i < cd.length; i++) cd[i].dim = 0;
+    }
+    else {
+        for(i = 0; i < cd.length; i++) {
+            di = cd[i];
+            x = xa.c2p(di.x);
+            y = ya.c2p(di.y);
+            if(polygon.contains([x, y])) {
+                selection.push({
+                    id: i,
```
etpinard: TODO: `i` here isn't the `id`, it's the `pointNumber`. 

Would be best to merge https://github.com/plotly/plotly.js/pull/1770 first, as this here will break after commit 3399a58 is applied.

etpinard: done in debdd4645eb77fa901f3305567e1f8170b89b700

test/jasmine/tests/gl2d_click_test.js
```diff
@@ -444,3 +444,149 @@ describe('Test hover and click interactions', function() {
         .then(done);
     });
 });
+
+describe('Test gl2d lasso/select:', function() {
```
etpinard: Painful! For some reason, these tests doesn't work on CI. Looks like the cases below can't even draw the scene:

image

https://circleci.com/gh/plotly/plotly.js/4535#tests/containers/1

Can anyone try to pull down this branch and try this out locally?

dfcreative: Works fine for me:
image
Although couple of times `'should output correct event data for scattergl'` failed due to delay or something.

etpinard:
Would anyone be opposed to merging this without CI test support? 

I can't figure out why these new tests keep failing on CircleCI. With the `@noCI` tag, they will run locally on `npm run test-jasmine` and be part of the pre-release checklist in [`./tasks/noci_test.sh`](https://github.com/plotly/plotly.js/blob/master/tasks/noci_test.sh).

monfera:
@etpinard sounds good to me!

Issue #1662, opened 2017-05-10
Sankey hover labels misaligned when the plot isn't in the corner of the page?


alexcjohnson:
If I load a sankey diagram in the test dashboard - where plots are drawn on the left edge of the page but not at the top - I see hover labels misaligned with the links:

![screen shot 2017-05-10 at 1 09 53 pm](https://cloud.githubusercontent.com/assets/2678795/25911704/d9f8a000-3582-11e7-865c-948c8496c3dd.png)

To reproduce: `npm start` from the plotlyjs repo root, then go to eg http://localhost:3000/devtools/test_dashboard/#sankey_energy

cc @monfera

etpinard:
Blocking or not for `1.27.0`?

alexcjohnson:
yes, blocking `1.27.0` IMO

etpinard:
@monfera can I help with anything?

monfera:
@etpinard thanks a lot, done: https://github.com/plotly/plotly.js/issues/1664

etpinard:
closed by https://github.com/plotly/plotly.js/pull/1664

PR #1664, opened 2017-05-10
Fixes hover tooltip position with non-origin div position

monfera:
Issue reported by @alexcjohnson is https://github.com/plotly/plotly.js/issues/1662

### DIFF
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -125,13 +125,14 @@ module.exports = function plot(gd, calcData) {
 
     var linkHoverFollow = function(element, d) {
 
+        var rootBBox = gd.getBoundingClientRect();
         var boundingBox = element.getBoundingClientRect();
         var hoverCenterX = boundingBox.left + boundingBox.width / 2;
         var hoverCenterY = boundingBox.top + boundingBox.height / 2;
 
         var tooltip = Fx.loneHover({
-            x: hoverCenterX + window.scrollX,
-            y: hoverCenterY + window.scrollY,
+            x: hoverCenterX - rootBBox.left + window.scrollX,
+            y: hoverCenterY - rootBBox.top + window.scrollY,
             name: d3.format(d.valueFormat)(d.link.value) + d.valueSuffix,
             text: [
                 d.link.label,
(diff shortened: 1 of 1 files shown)


src/traces/sankey/plot.js
```diff
@@ -125,13 +125,14 @@ module.exports = function plot(gd, calcData) {
 
     var linkHoverFollow = function(element, d) {
 
+        var rootBBox = gd.getBoundingClientRect();
```
etpinard: Can you use `fullLayout.width` and `fullLayout.height` or even `fullLayout._size` instead? Those `getBoundingClientRect` calls are expensive.

monfera: thanks, I'll check

monfera: ... maybe I could use help for a low-latency fix :-) the code needs to know the positions, not just the width/height - having said this, the speed impact is likely negligible as it's only triggered on one element at a time. Yet if there's a need to quickly switch, I could use suggestions for getting the right position without a bbox call.

alexcjohnson: this matches the pattern we've had to use elsewhere. It's possible that *both* `getBoundingClientRect` calls could be avoided by figuring out the position directly from the link/node element (or by caching it in the element's d3 data or something) but if you use `getBoundingClientRect` for one you'd better use it for both. That could be an optimization later, this is good for now.

src/traces/sankey/plot.js
```diff
@@ -125,13 +125,14 @@ module.exports = function plot(gd, calcData) {
...
         var boundingBox = element.getBoundingClientRect();
         var hoverCenterX = boundingBox.left + boundingBox.width / 2;
         var hoverCenterY = boundingBox.top + boundingBox.height / 2;
 
         var tooltip = Fx.loneHover({
-            x: hoverCenterX + window.scrollX,
-            y: hoverCenterY + window.scrollY,
+            x: hoverCenterX - rootBBox.left + window.scrollX,
+            y: hoverCenterY - rootBBox.top + window.scrollY,
```
alexcjohnson: are the `window.scrollX/Y` shifts still necessary now that you're using a `getBoundingClientRect` differential?

Issue #1666, opened 2017-05-10
Filter transform doesn't restore data text


n-riesco:
I'm using filter transforms to hide and show points in a trace. These points have a `text` and I've noticed that in some cases when I unhide a point, its `text` isn't displayed.

Here's a [codepen](http://codepen.io/n-riesco/pen/aWqXWa) reproducing the issue.

![peek 2017-05-10 18-58](https://cloud.githubusercontent.com/assets/6199391/25913526/f4f539c4-35b2-11e7-9c58-4bf65e6d102b.gif)

etpinard:
Interesting. Thanks for the report @n-riesco !

I can replicate the problem on `master`. By inspecting `fullData` and `calcdata`, the filter transform appear to do the _right thing_. Something must be up in the `plot` routine.

etpinard:
OK. I found the problem. When a filter transform filters out all items in data arrays the resulting `fullData` items is inconsistent with what the plot step except. More precisely, filter transforms allows for `fullData` items containing:

```js
x: [],
y: [],
visible: true
```
whereas (all) other plotly code paths force `visible: false` whenever blank coordinates are found.

In details, logging `x`, `y` and `visible` in `fullData` after `restyle(gd, 'transforms', /* */`)` call gives:

![peek 2017-05-10 17-50](https://cloud.githubusercontent.com/assets/6675409/25922930/23142ea0-35a9-11e7-9873-d7e04639458a.gif)

Luckily the marker points appear to update correctly when `x` and `y` are blank. No luck for the text point nodes though. But, we should **not** rely on the plot step knowing what to do with blank `x` / `y` arrays. Instead we should make sure the filter transform forces in `visible: false`. Perhaps we should call `Plots.supplyDefaults` after the calc transform is applied (probably an overkill)?

--- 

In brief @n-riesco , one workaround would have your toggle buttons never reach that `transform.target = [0, 0, 0]` state. In the case, the restyle command should be `restyle(gd, 'visible', false)`.

n-riesco:
@etpinard How about `filter` transforms setting `visible` to `false` whenever they result in an empty trace? Would that fix the issue?

PR #1672, opened 2017-05-11
Scatter <g textpoint> join fixup

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1666

cc @rreusser and @n-riesco 

Continuing on https://github.com/plotly/plotly.js/issues/1666#issuecomment-300627893, I found a case (see 		310e67a ) where `restyle` was failing on blank `x`, `y` arrays without using filter transforms, so the patch had to be made in `scatter/plot.js` (see 49e5da5)

After this patch:

peek 2017-05-11 11-00

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -47,6 +47,17 @@ drawing.setRect = function(s, x, y, w, h) {
     s.call(drawing.setPosition, x, y).call(drawing.setSize, w, h);
 };
 
+/** Translate / remove node
+ *
+ * @param {object} d : calcdata point item
+ * @param {sel} sel : d3 selction of node to translate
+ * @param {object} xa : corresponding full xaxis object
+ * @param {object} ya : corresponding full yaxis object
+ *
+ * @return {boolean} :
+ *  true if selection got translated
+ *  false if selection got removed
+ */
 drawing.translatePoint = function(d, sel, xa, ya) {
     // put xp and yp into d if pixel scaling is already done
     var x = d.xp || xa.c2p(d.x),
@@ -59,8 +70,12 @@ drawing.translatePoint = function(d, sel, xa, ya) {
         } else {
             sel.attr('transform', 'translate(' + x + ',' + y + ')');
         }
+    } else {
+        sel.remove();
+        return false;
     }
-    else sel.remove();
+
+    return true;
 };
 
 drawing.translatePoints = function(s, xa, ya, trace) {
--- a/src/traces/scatter/plot.js
+++ b/src/traces/scatter/plot.js
@@ -458,8 +458,10 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         join.enter().append('g').classed('textpoint', true).append('text');
 
         join.each(function(d) {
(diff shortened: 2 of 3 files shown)


src/traces/scatter/plot.js
```diff
@@ -458,8 +458,10 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         join.enter().append('g').classed('textpoint', true).append('text');
 
         join.each(function(d) {
-            var sel = transition(d3.select(this).select('text'));
-            Drawing.translatePoint(d, sel, xa, ya);
+            var g = d3.select(this);
+            var sel = transition(g.select('text'));
+            var hasTextPt = Drawing.translatePoint(d, sel, xa, ya);
+            if(!hasTextPt) g.remove();
```
etpinard: Each `<text>` node is wrapped in a `<g class="textpoint">` group. It is on this group that the data-join happens so we must make sure that group is removed whenever `<text>` node are removed to obtain the correct enter and merge selections on updates.

test/jasmine/tests/scatter_test.js
```diff
@@ -383,6 +383,60 @@ describe('end-to-end scatter tests', function() {
...
+            }
+        }])
+        .then(function() {
+            assertNodeCnt(3, 3);
+            assertText(['A', 'B', 'C']);
+
+            return Plotly.restyle(gd, {
+                x: [[null, undefined, NaN]],
+                y: [[NaN, null, undefined]]
+            });
```
etpinard: On `master`, this `restyle` call results in 

```js
d3.selectAll('.point').size() // => 0 (as desired)
d3.selectAll('.textpoint').size() // => 3 (which messes up subsequent updates)
```

src/traces/scatter/plot.js
n-riesco: Good catch!

src/components/drawing/index.js
```diff
@@ -47,6 +47,17 @@ drawing.setRect = function(s, x, y, w, h) {
...
+ *
+ * @param {object} d : calcdata point item
+ * @param {sel} sel : d3 selction of node to translate
+ * @param {object} xa : corresponding full xaxis object
+ * @param {object} ya : corresponding full yaxis object
+ *
+ * @return {boolean} :
+ *  true if selection got translated
+ *  false if selection got removed
+ */
```
alexcjohnson: Thanks 🎎  !

src/traces/scatter/plot.js
alexcjohnson: `Drawing.translatePoint` gets used in [line 436 above](https://github.com/plotly/plotly.js/pull/1672/files#diff-b15c528765724e5b795253c8314bb750R436) too, and could short-circuit the lines after it if it removed the node.

etpinard: Good eyes :eye: thanks!

etpinard: done in 3f7601c66bec6e8a2ef963cdb433938a315392c4

alexcjohnson:
Great! 💃

Issue #1678, opened 2017-05-12
Annotation text not getting placed correctly on relayout when page is scrolled


etpinard:
First reported in https://community.plot.ly/t/text-move-to-wrong-position-when-scrolling-update-or-relayout-graph/4209

:sos: This is a regression `1.27.0` :sos: 

From http://codepen.io/plotly/pen/Kzjamd?editors=0011 

- make the page scrollable (e.g. by opening the codePen console)
- click on a point, get an annotation (this works)
- scroll down
- click on a point, now the annotation text nodes are misplaced
- scroll up
- click on a point, now all the annotation text nodes are placed properly

![peek 2017-05-12 10-26](https://cloud.githubusercontent.com/assets/6675409/26002419/9073f728-36fd-11e7-97b9-d8c87db62d7b.gif)

----

I can't find anything suspicious in the `1.27.0` changelog. Maybe the `Drawing` optimisations of https://github.com/plotly/plotly.js/pull/1585 are to blame?

rreusser:
Hmm.  About the textpoint class? That could possibly have an effect on the transform/x/y properties.

etpinard:
This regression is indeed caused by https://github.com/plotly/plotly.js/pull/1585/commits/29abb5d8ec3dea92cc71f438fffa6785463d1fb1 

I'm not sure why though. I'll try to come up with a test case.

Issue #1675, opened 2017-05-11
No longer possible to run tests for components which import plotly without a browser


wayofthepie:
Recently a change was made to _svg_text_utils.js_ which instantiates a DOMParser at the top level of that module: https://github.com/plotly/plotly.js/blob/ce970f112a8fff0d09909529be8a713f5d4f36cc/src/lib/svg_text_utils.js#L22.

This has broken a number of tests I have which do not use a browser (or jsdom etc..) because the DOMParser will be instantiated on import, giving the following:
```
var parser = new DOMParser();
                 ^

ReferenceError: DOMParser is not defined
    at Object.<anonymous> (/home/user/repo/node_modules/plotly.js/src/lib/svg_text_utils.js:22:18)
...
```
For now I've applied a very quick fix:
```
var parser;
function cachingParser() {
  parser = parser === undefined ? new DOMParser(): parser;
  return parser
};
...
var dom = cachingParser().parseFromString(skeleton, 'application/xml'),
        childNode = dom.documentElement.firstChild;
...
```
Can this be updated to something like the above? If not, then does the DOMParser really need to be at the top level?

etpinard:
Thanks for tracking down this bug. 

I'm curious, what kind of tests are you running in node? I'm surprised that anything plotly.js works in node w/o at least using `jsdom`. That said, we should at the very least repalce that `new DOMParser()` statement  with `new window.DOMParser()` so that it can be more easily mocked in non-browser runtimes. 

> If not, then does the DOMParser really need to be at the top level?

It doesn't need to. But, apparently instantiating `DOMParser` is a non-negligible operation so @hy9be moved it to the module scope in https://github.com/plotly/plotly.js/pull/1585 instead of instantiating it  every time a text node is drawn.

wayofthepie:
I actually am using jsdom :wink:  was half asleep yesterday when I came across this!

I can set up a DOMParser as is with jsDom, so all is good really for me, but ya maybe what you suggested would be better, to use `new window.DOMParser()` as window is generally simulated anyway.

bnbarak:
Temporary fix to run tests that import `Plotly` in a non-browser environment.

```  Javascript
const importPlotly = () => {
  if(window.isTesting)
    return {Plot: null, Plotly: null} // or Plot: <div />

  const Plotly = require('plotly.js/dist/plotly.min.js');
  const Plot = require('react-plotly.js');
  return { Plot, Plotly };
};

[...]

PR #1683, opened 2017-05-12
Drawing bbox fix for window scroll

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1678

I decide to play it safe an revert https://github.com/plotly/plotly.js/pull/1585/commits/29abb5d8ec3dea92cc71f438fffa6785463d1fb1 until we :lock: the `Drawing.bBox` behavior better. Reverting https://github.com/plotly/plotly.js/pull/1585/commits/29abb5d8ec3dea92cc71f438fffa6785463d1fb1 also fixes https://github.com/plotly/plotly.js/issues/1675, but maybe we could do better there.

cc @alexcjohnson

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -589,7 +589,6 @@ drawing.makeTester = function(gd) {
 // always returns a copy of the bbox, so the caller can modify it safely
 var savedBBoxes = [],
     maxSavedBBoxes = 10000;
-
 drawing.bBox = function(node) {
     // cache elements we've already measured so we don't have to
     // remeasure the same thing many times
@@ -598,14 +597,12 @@ drawing.bBox = function(node) {
         return Lib.extendFlat({}, savedBBoxes[saveNum.value]);
     }
 
-    if(!drawing.test3) {
-        drawing.test3 = d3.select('#js-plotly-tester');
-        drawing.tester = drawing.test3.node();
-    }
+    var test3 = d3.select('#js-plotly-tester'),
+        tester = test3.node();
 
     // copy the node to test into the tester
     var testNode = node.cloneNode(true);
-    drawing.tester.appendChild(testNode);
+    tester.appendChild(testNode);
     // standardize its position... do we really want to do this?
     d3.select(testNode).attr({
         x: 0,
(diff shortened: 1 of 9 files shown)


test/jasmine/tests/drawing_test.js
```diff
@@ -348,6 +345,54 @@ describe('Drawing', function() {
...
+            })
+            .then(function() {
+                var node = d3.select('text.annotation').node();
+                expect(Drawing.bBox(node)).toEqual({
+                    height: 14,
+                    width: 41.015625,
+                    left: -20.671875,
+                    top: -11,
+                    right: 20.34375,
+                    bottom: 3
```
etpinard: Commit https://github.com/plotly/plotly.js/pull/1585/commits/29abb5d8ec3dea92cc71f438fffa6785463d1fb1 made the `bottom` and `top` keys here incorrectly change - as the bbox computation were based on a outdated cached tester element.

etpinard: The `width`, `left` and `right` keys are change correctly here in response to the annotation `text` relayout update.

etpinard: image

Ha, those values appear to be machine-dependent (I suspect a font issue) ...

etpinard: fixed in 3fdb3081ad2c8231145a57443cf381b6eac14688

src/components/drawing/index.js
```diff
@@ -598,36 +597,32 @@ drawing.bBox = function(node) {
...
         y: 0,
         transform: ''
     });
 
-    var testRect = testNode.getBoundingClientRect();
-    if(!drawing.refRect) {
-        drawing.refRect = drawing.test3.select('.js-reference-point')
+    var testRect = testNode.getBoundingClientRect(),
+        refRect = test3.select('.js-reference-point')
             .node().getBoundingClientRect();
```
alexcjohnson: If you want to keep the perf gains from #1585 in this fix, I'm pretty sure the only change we'd need is to call this `.getBoundingClientRect()` with each `drawing.bBox` call, rather than caching it. But we can still cache `tester`, and we can cache `test3.select('.js-reference-point').node()`.

Just a question of timing really - now that you have a scrolling test in place it seems safe to me, but you can leave it as is for now if you need to move on to other things.

etpinard: that's in https://github.com/plotly/plotly.js/pull/1690

alexcjohnson:
💃  once #1690 is merged and tests pass

PR #1689, opened 2017-05-15
Sort + set 'categoryarray'

etpinard:
```js
Plotly.newPlot('graph', [{
  mode: 'markers',
  x: ['A', 'B', 'C'],
  y: ['a', 'b', 'c'],
  marker: { size: [20, 30, 50] },
  transforms: [{
    type: 'sort',
    target: [2, 1, 0]
  }]
}])
```

currently gives:

image

where the categories are re-arranged after the `sort` transform - which is consistent with how we handle categories in `filter` transforms. See PR https://github.com/plotly/plotly.js/pull/1260 for more details.

But setting `layout.xaxis.categoryarray: ['A', 'B', 'C]` -- which implies that the category positions are fixed -- curently gives the same results as above.

With commit 7a2a2e1

```js
Plotly.newPlot('graph', [{
  mode: 'markers',
  x: ['A', 'B', 'C'],
  y: ['a', 'b', 'c'],
  marker: { size: [20, 30, 50] },
  transforms: [{
    type: 'sort',
    target: [2, 1, 0]
  }]
}], {
  xaxis: {
    categoryarray: ['A', 'B', 'C']
  }
})
```

now gives:

image

where `categoryarray` is honored.

cc @n-riesco @alexcjohnson

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2003,7 +2003,7 @@ plots.doCalcdata = function(gd, traces) {
         }
     }
 
-    var hasCategoryAxis = plots.initCategories(axList);
+    var hasCategoryAxis = initCategories(axList);
 
     var hasCalcTransform = false;
 
@@ -2039,7 +2039,7 @@ plots.doCalcdata = function(gd, traces) {
             axList[i]._categories = [];
             axList[i]._categoriesMap = {};
         }
-        plots.initCategories(axList);
+        initCategories(axList);
     }
 
     // 'regular' loop
@@ -2085,7 +2085,7 @@ plots.doCalcdata = function(gd, traces) {
     }
 };
 
-plots.initCategories = function(axList) {
+function initCategories(axList) {
     var hasCategoryAxis = false;
 
     // initialize the category list, if there is one, so we start over
@@ -2103,7 +2103,7 @@ plots.initCategories = function(axList) {
     }
 
     return hasCategoryAxis;
-};
+}
 
 plots.rehover = function(gd) {
     if(gd._fullLayout._rehover) {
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -2099,6 +2085,26 @@ plots.doCalcdata = function(gd, traces) {
     }
 };
 
+plots.initCategories = function(axList) {
```
alexcjohnson: is there a reason to export this fn?

etpinard: No.

src/plots/plots.js
```diff
@@ -2099,6 +2085,26 @@ plots.doCalcdata = function(gd, traces) {
...
+    var hasCategoryAxis = false;
+
+    // initialize the category list, if there is one, so we start over
+    // to be filled in later by ax.d2c
+    for(var i = 0; i < axList.length; i++) {
+        axList[i]._categories = axList[i]._initialCategories.slice();
```
rreusser: Certainly would have been caught by now if this could ever be undefined, right? (I'm too cautious about checking for undefined unnecessarily)

etpinard: Yeah. That `_initialCategories` thing is set in the defaults. Calling `Plots.doCalcdata` before `supplyDefaults` is like :gun: yourself in the foot.

rreusser: Haha the squirt gun. And then your socks are wet for the rest of the day.

src/plots/plots.js
alexcjohnson: Not a big deal, but I'd prefer not to since it goes into `Plotly.Plots`

etpinard: Yeah, good call :+1:

alexcjohnson:
💃  very clean! But consider https://github.com/plotly/plotly.js/pull/1689#discussion_r116596355 before merging.

src/plots/plots.js
etpinard: done in b8e5f52

PR #764, opened 2016-07-21
Plotting in child window

nielsenb-jf:
This pull request fixes the issues noted in [issue 702](https://github.com/plotly/plotly.js/issues/702). Essentially, if Plotly was provided an element in a different window or tab, the plot would not be styled correctly and interactivity (hovers, drag zoom, etc.) would not work.

It is squished to one commit due to issues caused by multiple merges during development [see this previous PR](https://github.com/plotly/plotly.js/pull/758).

The following are the big changes to fix this issue:

- Graph divs now keep a reference to their 'parent' document in the `_document' property.
- Styles are now injected to the parent document on the first call to `Plotly.plot`.
- Anything that draws over a document for interactivity purposes (eg. dragcovers), now uses the `_document` reference to ensure it is drawn in the correct document.
- Notifiers now require a graph div as the first argument so that notifications are rendered in the correct place

Issue #702, opened 2016-06-28
Interactive portions of plot do not render correctly in a child window


nielsenb-jf:
When using Plotly to render in a child window, the hover functionality for plots seems to break completely. Any enabled buttons will render in the incorrect location, but seem to function correctly.

[This CodePen](http://codepen.io/anon/pen/RRpMzL) shows the issue.

The attached screenshot demonstrates the issue as well.

I have reproduced in the latest stable Firefox and Chrome.

![screenshot](https://cloud.githubusercontent.com/assets/20195166/16433155/5df01308-3d4e-11e6-8e68-c99c528c3636.png)

etpinard:
Thanks for report.

Interesting. I'm not sure what could cause that.

nielsenb-jf:
Everything renders correctly if you copy the injected styles from the parent window to the child. See [this CodePen](http://codepen.io/anon/pen/ZOaGoY) for a demonstration.

Things break if you try to interact with the plot by clicking on a data point or by zooming by clicking and selecting a zoom area.

etpinard:
@nielsenb-jf Interesting. Thanks for letting us know.

nielsenb-jf:
I have a [branch in my fork](https://github.com/nielsenb-jf/plotly.js/tree/coverslip_fix) that demonstrates fixing interactivity when rendering to a child window. Basically, it renders the drag cover to the document of the child window, instead of always to the parent window where Plotly was instantiated.

Looking at the code more, Plotly makes lots of assumptions about which document / window to render to. I'm not sure if it would be worth a pull request for only fixing this one issue, I'm sure there are others I haven't found yet.

Would there be interest in trying to make Plotly more window / document agnostic? It does potentially enable some interesting multi-window webapps with Plotly. This is a big enough undertaking that I'd rather not start it if there's no upstream interest in incorporating it.

If yes, what should be done about the issue of injecting stylesheets? Should they be injected to the document containing the original element given to Plotly? Is that even feasible?

What about getting the correct parent window? [document.defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView) would be the 'modern' way to do it, but would limit support of IE versions before 9. Is that an issue? I've also found a [polyfill of sorts on StackOverflow](http://stackoverflow.com/a/10260692).

nielsenb-jf:
[...]

PR #1690, opened 2017-05-15
Drawing perf take2

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1675 

Take 2 of https://github.com/plotly/plotly.js/commit/29abb5d8ec3dea92cc71f438fffa6785463d1fb1 from  @hy9be's https://github.com/plotly/plotly.js/pull/1585 to be merged in https://github.com/plotly/plotly.js/pull/1683 

@hy9be would you mind checking that this solution here is as fast as yours?

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -542,11 +542,11 @@ drawing.steps = function(shape) {
 };
 
 // off-screen svg render testing element, shared by the whole page
-// uses the id 'js-plotly-tester' and stores it in gd._tester
+// uses the id 'js-plotly-tester' and stores it in drawing.tester
 // makes a hash of cached text items in tester.node()._cache
 // so we can add references to rendered text (including all info
 // needed to fully determine its bounding rect)
-drawing.makeTester = function(gd) {
+drawing.makeTester = function() {
     var tester = d3.select('body')
         .selectAll('#js-plotly-tester')
         .data([0]);
@@ -579,7 +579,7 @@ drawing.makeTester = function(gd) {
         tester.node()._cache = {};
     }
 
-    drawing.tester = gd._tester = tester;
+    drawing.tester = tester;
     drawing.testref = testref;
 };
 
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -580,7 +580,7 @@ drawing.makeTester = function(gd) {
     }
 
     drawing.tester = gd._tester = tester;
-    drawing.testref = gd._testref = testref;
+    drawing.testref = testref;
 };
 
 // use our offscreen tester to get a clientRect for an element,
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -579,16 +579,17 @@ drawing.makeTester = function(gd) {
         tester.node()._cache = {};
     }
 
-    gd._tester = tester;
-    gd._testref = testref;
+    drawing.tester = gd._tester = tester;
+    drawing.testref = gd._testref = testref;
 };
 
 // use our offscreen tester to get a clientRect for an element,
 // in a reference frame where it isn't translated and its anchor
 // point is at (0,0)
 // always returns a copy of the bbox, so the caller can modify it safely
-var savedBBoxes = [],
-    maxSavedBBoxes = 10000;
+var savedBBoxes = [];
+var maxSavedBBoxes = 10000;
+
 drawing.bBox = function(node) {
     // cache elements we've already measured so we don't have to
     // remeasure the same thing many times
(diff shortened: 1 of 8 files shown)


src/components/drawing/index.js
```diff
@@ -579,16 +579,17 @@ drawing.makeTester = function(gd) {
         tester.node()._cache = {};
     }
 
-    gd._tester = tester;
-    gd._testref = testref;
+    drawing.tester = gd._tester = tester;
+    drawing.testref = gd._testref = testref;
```
alexcjohnson: `gd._testref` currently isn't used anywhere so we might as well get rid of it before someone *does* use it.

`gd._tester` *is* used in a few places, but wouldn't be hard to 🔪  and replace with `drawing.tester`.

src/lib/svg_text_utils.js
```diff
@@ -17,6 +17,17 @@ var Lib = require('../lib');
 var xmlnsNamespaces = require('../constants/xmlns_namespaces');
 var stringMappings = require('../constants/string_mappings');
 
+exports.getDOMParser = function() {
+    if(exports.domParser) {
+        return exports.domParser;
```
alexcjohnson: Lets make `domParser` a module-level global instead of an export - it's unsafe to use in general as it might not exist before calling `exports.getDOMParser()`.

src/components/drawing/index.js
etpinard: :stop_sign:  Hold the presses :stop_sign: 

Caching the tester div and testref path on the `Drawing` module object will make plotly.js even more broken in child windows than it is at the moment. From https://github.com/plotly/plotly.js/issues/702 and the attempted PR https://github.com/plotly/plotly.js/pull/764, plotly.js CSS is  inserted once per `<html>` page - which doesn't propagate to child windows. We should instead insert plotly.js CSS once per `document`.

Similarly, we should be caching `tester` and `testref` once per `document` too. That is, child windows need their own tester nodes. As we don't have access to cache once-per-document object (adding refs to `document` sounds dirty), caching `tester` and `testref` on the `gd` might actually be the best we can do. 

This would imply passing `gd` to `Drawing.bBox` below - which isn't a high price to pay.

src/lib/svg_text_utils.js
etpinard: Good call. Done in e0050f73c57c54755a6131584b152151bf0c7c06

src/components/drawing/index.js
alexcjohnson: > child windows need their own tester nodes

Do they? It looks to me (fiddling in the [codepen](http://codepen.io/anon/pen/RRpMzL) from @nielsenb-jf ) as though if you have a reference to a node, you can interact with it fully even if it's in a different window. It won't get the same CSS applied to it, sure, but we don't depend on that here, do we?

etpinard: Right, you can interact with the base `document` in a child window. But does `getBoundingClientRect()` give you the correct answer in the case?

etpinard: See what happens when setting `layout.title`:

http://codepen.io/etpinard/pen/LyBQLj?editors=0010

gives 

image

which I believe is due to `Drawing.bBox` not computing the correct values.

etpinard: @alexcjohnson can we agree on fixing the child window behavior in an another PR?

I'm ok leaving this PR as is or pushing two more commits replacing `gd._tester` -> `Drawing.tester` and `gd._testref` -> `Drawing.testref`.

alexcjohnson: Sure. child window behavior is strange... like it looks like in IE and Edge we have some even bigger problems with it.

FWIW I think the `layout.title` issue you flagged above is not a bounding box issue, but the fact that the two `<svg>` layers are somehow not on top of each other, the front layer appears down below. the main one.

alexcjohnson: re: `Drawing.` vs `gd.` - we had better pick one and go use only that. As long as we look into this issue in the near future I don't care too much which we use. As is, if there *is* any difference between them, we'll be making a mess of it as soon as we have plots in two different windows... but in an order-dependent way that will be very hard to debug.

etpinard: Ok. Let's go with `Drawing` :+1:

etpinard: Done in 	c3c3b3a and a8fba03

alexcjohnson:
LGTM! 💃

hy9be:
@etpinard I did some quick test and did not find any difference other than some minor variance. So it should be okay in terms of performance. Thanks for checking with me!

etpinard:
@hy9be thanks **you** for checking :ok_hand:

Issue #933, opened 2016-09-13
Additional animation issues


rreusser:
I've found some additional issues related to animation that should be fixed ASAP:
- [x] transforms are an array, which extendDeepNoArrays accidentally avoids extending
- [ ] animate doesn't obey the correct semantics where changes are applied to all traces when trace indices not specified (copy from generalized function as a result of `Plotly.update`?)
- [ ] animate should be able to apply more than one nextFrame per frame in order to keep up with changes
- [x] animation should emit an event on each frame _so that you can figure out which frame it's on_! This is crucial for displaying information about an ongoing animation.
- [ ] convert async sequences to `Lib.syncOrAsync` (see first try in https://github.com/plotly/plotly.js/pull/1693)

rreusser:
@etpinard @monfera Curious to hear your thoughts on the merging issue.

Animations currently expand objects then `extendDeepNoArrays`. That's mostly okay except for a couple issues:

- Transforms don't get properly merged because of the NoArrays part
- `{'marker.color': ...}` and `marker: {color: ...}` both have the same effect, which makes frames nice and easy to write, but is slightly different than the rest of plotly which overwrites all marker properties for the second case.
- animations don't do array indexing like `{'transforms.0.type': ...}`

Long story short, I need to fix a couple bugs, so I either need to commit to the current approach and make it 100% correct, or I need to reevaluate and make the right choice about the most straightforward and sensible way for this to work.

rreusser:
After a brief discussion, the conclusion is that the current behavior is best: attribute strings are good for single properties but awkward for whole frames (which would then need *every* property attribute-stringed, which wouldn't really be acceptable).

Then working through transform array logic…

etpinard:
After https://github.com/plotly/plotly.js/pull/956 it would be easy

- [] add animation for `scatter` trace with `'text'` mode.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #1693, opened 2017-05-17
Move plotly.animate to Lib.syncOrAsync

rreusser:
This PR moves the content of `Plotly.animate` into `Lib.syncOrAsync` so that it better fits into the plotly queue. The particular use-case is `Plotly.restyle` and `Plotly.animate` called in synchronous series (i.e. not promise-chained), though I'm currently unable to produce a situation that actually triggers any problems. I stumbled upon the possibility of a problem while working on the notifier, and solved things just by properly `then`-ing all the plotly API calls. That's good practice anyway, but this is also probably the right way for things to be.

I'm not quite sure how to produce a test case for this (short of inspecting the promise queue, except those are just anonymous callbacks, right?), but I'm glad to try if someone can come up with an idea that's not inherently fragile due to a race condition.

Additional note: there are only ~240 lines due to an indentation change. I've only changed this:

```javascript
return new Promise(...)
```

into this

```javascript
return Lib.syncOrAsync([
  function () {
    return new Promise(...)
  }
], gd);
```

/cc @alexcjohnson @etpinard

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2700,7 +2700,7 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
                 }
             });
         }
-    ]);
+    ], gd);
 };
 
 /**
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2423,280 +2423,284 @@ Plotly.animate = function(gd, frameOrGroupNameOrFrameList, animationOpts) {
         };
     }
 
-    return new Promise(function(resolve, reject) {
-        function discardExistingFrames() {
-            if(trans._frameQueue.length === 0) {
-                return;
-            }
-
-            while(trans._frameQueue.length) {
-                var next = trans._frameQueue.pop();
-                if(next.onInterrupt) {
-                    next.onInterrupt();
-                }
-            }
-
-            gd.emit('plotly_animationinterrupted', []);
-        }
-
-        function queueFrames(frameList) {
-            if(frameList.length === 0) return;
-
-            for(var i = 0; i < frameList.length; i++) {
-                var computedFrame;
-
-                if(frameList[i].type === 'byname') {
-                    // If it's a named frame, compute it:
-                    computedFrame = Plots.computeFrame(gd, frameList[i].name);
-                } else {
-                    // Otherwise we must have been given a simple object, so treat
-                    // the input itself as the computed frame.
-                    computedFrame = frameList[i].data;
-                }
-
-                var frameOpts = getFrameOpts(i);
-                var transitionOpts = getTransitionOpts(i);
-
-                // It doesn't make much sense for the transition duration to be greater than
-                // the frame duration, so limit it:
-                transitionOpts.duration = Math.min(transitionOpts.duration, frameOpts.duration);
-
-                var nextFrame = {
-                    frame: computedFrame,
-                    name: frameList[i].name,
-                    frameOpts: frameOpts,
-                    transitionOpts: transitionOpts,
-                };
-                if(i === frameList.length - 1) {
-                    // The last frame in this .animate call stores the promise resolve
-                    // and reject callbacks. This is how we ensure that the animation
-                    // loop (which may exist as a result of a *different* .animate call)
-                    // still resolves or rejecdts this .animate call's promise. once it's
-                    // complete.
-                    nextFrame.onComplete = callbackOnNthTime(resolve, 2);
-                    nextFrame.onInterrupt = reject;
-                }
-
-                trans._frameQueue.push(nextFrame);
-            }
-
-            // Set it as never having transitioned to a frame. This will cause the animation
-            // loop to immediately transition to the next frame (which, for immediate mode,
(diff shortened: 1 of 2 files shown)


rreusser:
Okay, it passes the tests. That's the first big step. That means it's *worth* inventing a test. I'll just be creative.

rreusser:
I've added a test, but the conclusion is that most of the work was ending up on the main plotly queue anyway via `Plots.transition` so that I'm not 100% convinced this *should* be merged. I need to think about the extreme subtleties just a bit more.

rreusser:
Okay here's my conclusion on this PR:

I can't find a problem it solves. `Plotly.animate` currently does some setup work *synchronously* when called. This PR defers that work until `syncOrAsync` starts executing its callbacks, but since the frame modification functions do their work and resolve synchronously anyway, there's not really a way to get things out of sequence unless you misuse the API in ways that aren't really 'correct' anyway, i.e.:

```javascript
Plotly.plot(...)
Plotly.animate(...)
Plotly.addFrames(...)
```

instead of:

```javascript
Plotly.plot(...).then(() => {
  return Plotly.addFrames(...)
}).then(() => {
  return Plotly.animate(...)
})
```

So in the end I don't really have a strong preference on whether to merge this or not. I think it comes down to a matter of conventions instead of cases that succeed/fail. My question then is:

Should plotly.js API functions do *any* setup when invoked, or should *everything* be wrapped in `Lib.syncOrAsync`?

alexcjohnson:
@rreusser there's not all that much that plotly.js does async, so it's not surprising that you can't pin this down. Mathjax rendering, `geo` and `mapbox` plots fetching data, fetching images... that might be it.

Note that `Lib.syncOrAsync` doesn't, on its own, wait for the existing promise queue to complete; it just ensures proper sequencing of items in its own arguments. It also doesn't necessarily do anything async or promise related, in fact it keeps things sync whenever possible.

If your goal is to wait for the plot to finish anything else it's been doing - which I think is a good goal, and probably important in some cases (a map or a plot with mathjax that animates on load?) then I think you just need to add `Plots.previousPromises`:
```js
return Lib.syncOrAsync([
    Plots.previousPromises, // only async if there *are* promises in the queue
    function() {
        return new Promise(...)
    }
], gd); // gd is just passed along to previousPromises, it's not needed for your own fn
```
then if everything is sync, `done` will just be `undefined`, but since you're making a new promise, it will just be that promise (having chained it onto previous ones if necessary). In other places we've had to [generate](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L387) a [promise](https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/plot_api/plot_api.js#L1873) to keep the API consistent, but you already know you have one.

rreusser:
Thanks for the clarification, @alexcjohnson! Is there a rule of thumb on which API functions should wait for previousPromises to finish? Like it's possible to write a synchronous-*looking* API that queues everything internally and is actually asynchronous, but I'm not sure if that's a good or bad goal. My specific concern is that API commands would get interleaved inside the async parts of `Plotly.plot` (though as you pointed out, even though it's a sequence of promises, do they all resolve synchronously?) leading to nasty race conditions.

alexcjohnson:
> Is there a rule of thumb on which API functions should wait for previousPromises to finish?

Offhand I'd say anything that can modify the plot should wait for `previousPromises` *and* return a promise itself. which is... basically the whole top-level API? Below that level we're basically not exposing anything new anyway, and we'll probably de-expose all of those in v2, so nothing to worry about.

> even though it's a sequence of promises, do they all resolve synchronously?

Are you talking about `gd._promises` or `Lib.syncOrAsync`? `gd._promises` is an array of whatever async stuff happened during plotting, so you could always call `Promise.all` on it if you *need* a promise, but `Plots.previousPromises` will return `undefined` if it's empty so that a `syncOrAsync` call will stay sync. `syncOrAsync` is called with a sequence, but it never creates promises, only waits for promises that its items return before continuing the sequence. If no items return a promise, the whole thing is sync, ie it's done before it returns and it does not return a promise.

etpinard:
> Mathjax rendering, geo and mapbox plots fetching data, fetching images... that might be it.

That is it, as far as I know. Shout-out to https://github.com/plotly/plotly.js/issues/76 :speaker: 


> Offhand I'd say anything that can modify the plot should wait for previousPromises and return a promise itself. which is... basically the whole top-level API?

`Plotly.purge` is an exception.

> Below that level we're basically not exposing anything new anyway, and we'll probably de-expose all of those in v2, so nothing to worry about.

:+1:

alexcjohnson:
> `Plotly.purge` is an exception.

Makes sense, though... what *would* happen if you purge with promises still pending? I guess the hope would be that it just throws an error inside the callback, which doesn't affect anything in the wider app... or perhaps it can still act on the now-removed plot elements and doesn't even throw an error? I guess mostly it will be fine but perhaps there are cases this would pollute `gd` by reattaching some variables to it or something.

etpinard:
Closing. It's probably best to find a scenario where not using `Lib.syncOrAsync` causes a bug before going through with this.

Now part https://github.com/plotly/plotly.js/issues/933

PR #604, opened 2016-06-06
Toimage bug

mdtusz:
Fixes #600.

This PR fixes a nasty little bug with image downloads. The culprit was font-family strings that were quoted and breaking the svg string used for image conversions.

The fix is not exactly _pretty_, but should be more robust now (previously, the quotation marks were escaped _before the svg was converted to a string_ - Chrome seems to not care for that and automatically replaces the double quotation marks.

PR #1694, opened 2017-05-17
Fix toImage for marker gradient

etpinard:
fixes https://github.com/plotly/support-assets/issues/87

The `Plotly.toImage` as well as the to-image modebar button are currently broken for graphs with set `marker.gradient` (added in https://github.com/plotly/plotly.js/pull/1620/f). Note that the image tests work fine inside the image test container - I'm thinking this has to do with nw.js 0.12 which uses a much older Chromium version (we gotta upgrade that thing soon).

The fix applies the same logic as in https://github.com/plotly/plotly.js/pull/604. In #604, the `"` surrounding `font-family` property value (e.g. `style="font-family: "Times New Roman""`) were stripped before serializing with `XMLSerializer` and then put back after. In this PR, the `"` surrounding the `fill` property value for gradients (e.g. `style="fill: url("#dgfdakngkda")"`) are given the same treatment. 

Note that, we don't have to the same with `clip-path` attributes, `XMLSerializer` handles those fine.

----- 

In future PRs, we should make that all new attributes that generate DOM style properties wrapped in `"` go through this process to avoid bugs like this one.

@alexcjohnson

### DIFF
--- a/src/snapshot/tosvg.js
+++ b/src/snapshot/tosvg.js
@@ -16,6 +16,9 @@ var Drawing = require('../components/drawing');
 var Color = require('../components/color');
 
 var xmlnsNamespaces = require('../constants/xmlns_namespaces');
+var DOUBLEQUOTE_REGEX = /"/g;
+var DUMMY_SUB = 'TOBESTRIPPED';
+var DUMMY_REGEX = new RegExp('("' + DUMMY_SUB + ')|(' + DUMMY_SUB + '")', 'g');
 
 
 module.exports = function toSVG(gd, format) {
@@ -90,9 +93,20 @@ module.exports = function toSVG(gd, format) {
             // to a string (browsers convert singles back)
             var ff = txt.style('font-family');
             if(ff && ff.indexOf('"') !== -1) {
-                txt.style('font-family', ff.replace(/"/g, 'TOBESTRIPPED'));
+                txt.style('font-family', ff.replace(DOUBLEQUOTE_REGEX, DUMMY_SUB));
             }
         });
+
+    svg.selectAll('.point').each(function() {
+        var pt = d3.select(this);
+        var fill = pt.style('fill');
+
+        // similar to font family styles above,
+        // we must remove " after the SVG DOM has been serialized
+        if(fill && fill.indexOf('url(') !== -1) {
+            pt.style('fill', fill.replace(DOUBLEQUOTE_REGEX, DUMMY_SUB));
+        }
+    });
 
     if(format === 'pdf' || format === 'eps') {
         // these formats make the extra line MathJax adds around symbols look super thick in some cases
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/snapshot_test.js
```diff
@@ -245,5 +245,53 @@ describe('Plotly.Snapshot', function() {
...
+
+                var textElements = svgDOM.getElementsByTagName('text');
+                expect(textElements.length).toEqual(11);
+
+                for(i = 0; i < textElements.length; i++) {
+                    expect(textElements[i].style['font-family']).toEqual('\"Times New Roman\"');
+                }
+
+                var pointElements = svgDOM.getElementsByClassName('point');
+                expect(pointElements.length).toEqual(3);
```
etpinard: Before this patch, `pointElement.length` was 0 i.e. `Snapshot.toSVG` failed to process them.

src/snapshot/tosvg.js
```diff
@@ -16,6 +16,9 @@ var Drawing = require('../components/drawing');
 var Color = require('../components/color');
 
 var xmlnsNamespaces = require('../constants/xmlns_namespaces');
+var DOUBLEQUOTE_REGEX = /"/g;
+var DUMMY_SUB = 'TOBESTRIPPED';
```
alexcjohnson: is there any way this can occur in other contexts in the SVG string? Or a way to add characters that make it more obviously impossible?

etpinard: Not that I can think of. The gradient `fill` property is set by us and even if `font.family: 'TOBESTRIPPED` then in the DOM this will look like: `font-family: TOBESTRIPPED` (no double quotes) which won't  trigger that `.replace(DOUBLEQUOTE_REGEX, DUMMY_SUB)` call.

alexcjohnson: hmm, looks like this is already a problem in master:
```
Plotly.relayout(gd,{'yaxis.title':'"TOBESTRIPPED"'})
```
on screen:
screen shot 2017-05-17 at 5 42 03 pm

download:
newplot 2

So you're not making a new bug, but we do need something better for this.

etpinard: OK. What's the solution then? 

- We could make the regex more precise [here](https://github.com/plotly/plotly.js/blob/master/src/snapshot/tosvg.js#L114) so that it grabs only `TOBESTRIPPED` inside `style=` props.
- Use another (more esoteric) dummy string
- Use another seriliazer than what's the browsers give us on the `window` object?

alexcjohnson: I don't know, I can't see an easy bulletproof solution. Lets just make an issue for the moment since the problem isn't new. Otherwise this looks great. I was surprised that this showed up in gradients since it didn't show up in clip paths (which also use these urls) but I guess only style attributes get these extra quotes.

etpinard: > I guess only style attributes get these extra quotes.

Exactly!

etpinard: Issue :arrow_right: https://github.com/plotly/plotly.js/issues/1697

alexcjohnson:
💃

src/snapshot/tosvg.js
alexcjohnson: > Issue :arrow_right: https://github.com/plotly/plotly.js/issues/1697

thanks!

Issue #1695, opened 2017-05-17
Add `disablemethod` flag to button and slider steps


rreusser:
An emerging pattern for examples of sufficient complexity is that it's becoming necessary to hook into components via events instead of strictly the command api interface. This might be the case if you have to, say, perform a couple precisely sequenced async actions before animating.

However, a problem with sliders, for example, is that bailing on specifying `method` and `args` as you normally would breaks the coupling that puts the slider in the correct position when you change the frame via some other pathway, like a play button.

I propose two things:

- `disablemethod` for `updatemenus` and `sliders` buttons/steps that leaves everything else unaffected and just skips actually executing the method so that you can hook into an event and call it yourself in a more sophisticated way. Otherwise the behavior will be unaffected.
- allow `null` for a method. Currently the [supplydefaults enum](https://github.com/plotly/plotly.js/blob/43d3710bdb183f9c1681f40efe8649a7238efce5/src/components/updatemenus/attributes.js#L21) blocks what seems like a legitimate usage when you just want a slider with no associated api call. (i.e. why have events if you can't use them?)

/cc @etpinard @alexcjohnson @bpostlethwaite for what might need to be expedited feedback and fixes

rreusser:
Other considerations: Maybe this should be the affirmative `enabled` with `dflt: true` instead. There could also be a separate flag for enabling/disabling the automagic two-way coupling, but that seems orthogonal and not currently necessary to address.

etpinard:
> allow null for a method. Currently the supplydefaults enum blocks what seems like a legitimate usage when you just want a slider with no associated api call. (i.e. why have events if you can't use them?)

Maybe not `null` (as `null` means `restyle(gd, astr, null)` means _set to attribute back to its default), but I like the idea of adding another value for `method`. Maybe `method: 'skip'`?

rreusser:
Sorry, just `null` for the `method`, not the `args`. (as in `Plotly[null]`, which is just a no-op as `Plotly['skip']` would also be). But I do like the idea of `skip`.

etpinard:
> as in Plotly[null], which is just a no-op 

[...]

PR #1699, opened 2017-05-18
Allow the skip api method for sliders and udpatemenus

rreusser:
This PR implements half of #1695: the `skip` method by allowing it as a method and otherwise living with it but mostly ignoring it.

/cc @etpinard @alexcjohnson @n-riesco

### DIFF
--- a/src/components/sliders/attributes.js
+++ b/src/components/sliders/attributes.js
@@ -20,7 +20,7 @@ var stepsAttrs = {
 
     method: {
         valType: 'enumerated',
-        values: ['restyle', 'relayout', 'animate', 'update'],
+        values: ['restyle', 'relayout', 'animate', 'update', 'skip'],
         dflt: 'restyle',
         role: 'info',
         description: [
--- a/src/components/sliders/defaults.js
+++ b/src/components/sliders/defaults.js
@@ -95,11 +95,12 @@ function stepsDefaults(sliderIn, sliderOut) {
         valueIn = valuesIn[i];
         valueOut = {};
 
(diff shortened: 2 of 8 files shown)


alexcjohnson:
Can you add a test that the `plotly_sliderchanged` and `plotly_buttonclicked` events still fire with `skip`?

rreusser:
I've added a test to `updatemenus` but not to `sliders`. I can add to `sliders`, but it's a bit tricky (requires calculating slider click positions and working through DOM nodes) and I don't think it's a particularly high-value test since I've tested the defaults to make sure nothing is skipped accidentally and the line that actually implements this is [in the command code](https://github.com/plotly/plotly.js/pull/1699/files#diff-ed618bfd5577de9b060caf82b52423cfR265), not the component code.

rreusser:
To say that more clearly, as long as the string is coerced correctly, the component code is completely agnostic about what the value is, except to the extent that it feeds back into the state via the command observer, which doesn't apply to this case anyway since the command observer is tested to return that it doesn't know anything about the `'skip'` command.

alexcjohnson:
> I've added a test to updatemenus but not to sliders

Works for me. All looks great! 💃

src/components/sliders/attributes.js
```diff
@@ -20,7 +20,7 @@ var stepsAttrs = {
 
     method: {
         valType: 'enumerated',
-        values: ['restyle', 'relayout', 'animate', 'update'],
+        values: ['restyle', 'relayout', 'animate', 'update', 'skip'],
```
etpinard: A little note on `'skip'` in the description would've been nice. We can get to that later though.

PR #1700, opened 2017-05-18
Add `execute` flag to components

rreusser:
This PR implements the second part of #1695 by adding an `execute` flag to sliders and updatemenus API calls. I had a bit of trouble getting a spy to work, so I omitted the end to end test at the moment. The full extent of the change is to simply sidestep the call to `Plots.executeAPICommand`, so it should have few effects except to skip the API call.

I have *not* yet tested whether everything works identically if the API call is moved into the event handler. Although that's the reason for this PR to exist, it's also a separate usage concern.

/cc @etpinard @alexcjohnson 
#1695

### DIFF
--- a/src/components/sliders/attributes.js
+++ b/src/components/sliders/attributes.js
@@ -56,6 +56,7 @@ var stepsAttrs = {
     },
     execute: {
         valType: 'boolean',
+        role: 'info',
         dflt: true,
         description: [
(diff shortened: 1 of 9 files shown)


alexcjohnson:
If not a spy, do you want to just check that `gd.data` doesn't change with `execute: false` but then does change if you set it `true`?

alexcjohnson:
Looks great, thanks for the test! 💃

PR #1702, opened 2017-05-18
Add `skip` documentation

rreusser:
/cc @etpinard 

Adds missing docs for #1699.


etpinard:
Thanks @rreusser :dancer:

PR #1708, opened 2017-05-19
Replace @src shortcut paths with `../../../src` in test assets

etpinard:
So that they can be more easily reused in other projects e.g. in https://github.com/plotly/support-assets/pull/150

@rreusser @alexcjohnson @n-riesco


alexcjohnson:
makes sense 💃

Issue #1706, opened 2017-05-18
Data ids interfere with transforms


n-riesco:
[This code](http://codepen.io/n-riesco/pen/MmPJYw) shows a bad interaction between data ids and transforms. The plot is made up of three bubbles sorted by marker size.

Clicking once on the button removes the circle in the middle using a filter transform.
Clicking a second time brings the circle back.
The issue is that when circle is back is drawn on top of all the circles (despite the sort transform).

![peek 2017-05-18 22-43](https://cloud.githubusercontent.com/assets/6199391/26224760/c6bfcd48-3c1b-11e7-944e-4f7383376335.gif)

etpinard:
This [`.order()`](https://github.com/d3/d3-3.x-api-reference/blob/master/Selections.md#order) is what we want to use I think, but looks like but doing so it does break some of our tests. :confused:

alexcjohnson:
pinging @rreusser - just to make sure this issue (and @etpinard's fix in #1709) is on your radar re: bar animations

PR #1709, opened 2017-05-19
Selection re-ordering for scatter traces with `ids`

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1706 (an attempt at least)

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -47,7 +47,7 @@ drawing.setRect = function(s, x, y, w, h) {
     s.call(drawing.setPosition, x, y).call(drawing.setSize, w, h);
 };
 
-/** Translate / remove node
+/** Translate node
  *
  * @param {object} d : calcdata point item
  * @param {sel} sel : d3 selction of node to translate
@@ -56,7 +56,7 @@ drawing.setRect = function(s, x, y, w, h) {
  *
  * @return {boolean} :
  *  true if selection got translated
- *  false if selection got removed
+ *  false if selection could not get translated
  */
 drawing.translatePoint = function(d, sel, xa, ya) {
     // put xp and yp into d if pixel scaling is already done
@@ -71,7 +71,6 @@ drawing.translatePoint = function(d, sel, xa, ya) {
             sel.attr('transform', 'translate(' + x + ',' + y + ')');
         }
     } else {
-        sel.remove();
         return false;
     }
 
--- a/src/traces/scatter/plot.js
+++ b/src/traces/scatter/plot.js
@@ -419,9 +419,6 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         var enter = join.enter().append('path')
             .classed('point', true);
 
-        enter.call(Drawing.pointStyle, trace)
-            .call(Drawing.translatePoints, xa, ya, trace);
-
         if(hasTransition) {
             enter.style('opacity', 0).transition()
                 .style('opacity', 1);
@@ -429,6 +426,8 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
 
         var markerScale = showMarkers && Drawing.tryColorscale(trace.marker, '');
         var lineScale = showMarkers && Drawing.tryColorscale(trace.marker, 'line');
+
+        join.order();
 
         join.each(function(d) {
             var el = d3.select(this);
@@ -443,8 +442,6 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
                 }
             }
         });
-
-        join.order();
 
         if(hasTransition) {
             join.exit().transition()
@@ -461,6 +458,8 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         // each text needs to go in its own 'g' in case
         // it gets converted to mathjax
         join.enter().append('g').classed('textpoint', true).append('text');
+
+        join.order();
 
         join.each(function(d) {
             var g = d3.select(this);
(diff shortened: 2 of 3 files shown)


etpinard:
FYI: more info on `selection.order()` [here](https://github.com/d3/d3-3.x-api-reference/blob/master/Selections.md#order)

src/traces/scatter/plot.js
```diff
@@ -483,6 +485,8 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
                 });
             });
 
+        join.order();
```
etpinard: :sos: IMPORTANT :sos: 

This commit breaks `scatter/plot.js` for traces with non-numeric (x,y) positions. 

That is due to [`Drawing.translatePoint`](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/components/drawing/index.js#L74) which is called on the [enter](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/traces/scatter/plot.js#L423) and [merge](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/traces/scatter/plot.js#L433) selections. When points with non-numeric (x,y) are detected, `Drawing.translatePoint` removes them. Unfortunately, this makes the `join` selections out-of-sync with the DOM and leads to _`'TypeError: Cannot read property 'textContent' of null`_ exception thrown by d3.

----

So, I have two questions for the d3-heads looking at this PR:

- Is there a way to re-select the join and call `order()` on that selection? 
- Or should we make our updates more d3-esque by filtering data items with non-numeric position **before** binding them to the selections?

n-riesco: How about `join.filter(onlyNumeric).order()`?

src/traces/scatter/plot.js
```diff
@@ -419,9 +419,6 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         var enter = join.enter().append('path')
             .classed('point', true);
 
-        enter.call(Drawing.pointStyle, trace)
```
etpinard: @n-riesco thanks very much :tada: 

Your solution appears to work:
peek 2017-05-23 11-43



My only concern now is how this interacts with animations. I think the reason why `translatePoints` and `pointStyle` are called on enter selection **and** merge selection has to do with this [block](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/traces/scatter/plot.js#L425-L428) below where entering nodes have their opacity transitioned from 0 to 1. I suspect this commit will break this behavior (although the tests are passing) has the entering aren't positioned and styled properly before opacity transition cc @rreusser . It would be nice to write a test case here :smirk: 

I'm thinking instead, we could keep those `enter.call()`, make `Drawing.translatePoints` not (and maybe never) remove nodes and make the merge selection `.each` block below (which comes after the `.order()`) handle the cases where nodes need to be removed.

Thoughts?

n-riesco: I think the point you make about animations is important.
And the solution you propose (i.e. remove nodes only after we're done with the enter selection) is more robust.
:+1: from me.

n-riesco: @etpinard after thinking a little bit about it. I'm still think that translatePoints and poinStyle should be applied to the merge selection to keep in sync DOM and trace.

etpinard: > I'm still think that translatePoints and pointStyle should be applied to the merge

That's always been the case [here](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/traces/scatter/plot.js#L436) and [here](https://github.com/plotly/plotly.js/blob/cea9455ba353fe62ef374454cc954c4b7945bb0a/src/traces/scatter/plot.js#L439).

test/jasmine/tests/scatter_test.js
```diff
@@ -539,6 +539,42 @@ describe('end-to-end scatter tests', function() {
         .catch(fail)
         .then(done);
     });
+
+    it('should smoothly add/remove nodes tags with *ids* during animations', function(done) {
```
etpinard: This test is failing after c8c3197

src/components/drawing/index.js
```diff
@@ -56,7 +56,7 @@ drawing.setRect = function(s, x, y, w, h) {
  *
  * @return {boolean} :
  *  true if selection got translated
- *  false if selection got removed
+ *  false if selection could not get translated
  */
 drawing.translatePoint = function(d, sel, xa, ya) {
```
etpinard: To my surprise, `Drawing.translatePoint` is called only in `scatter/plot.js` and `box/plot.js`:

image

Note that this patch does **not** affect the `box/plot.js` where points with non-numeric positions are filtered out of the data bind.

test/jasmine/tests/scatter_test.js
etpinard: and fixed in commit 9c35675

Before 9c35675
peek 2017-05-23 14-09


After:

peek 2017-05-23 14-08

etpinard:
Tagging this as `reviewable`, pinging @alexcjohnson (as this is a pretty important PR) and putting this under the `1.28.0` milestone (as the changes here are too heavy for a patch release IMHO).

Thanks again @n-riesco for your help!

src/traces/scatter/plot.js
```diff
@@ -430,6 +427,8 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
         var markerScale = showMarkers && Drawing.tryColorscale(trace.marker, '');
         var lineScale = showMarkers && Drawing.tryColorscale(trace.marker, 'line');
 
+        join.order();
```
alexcjohnson: why do we need `join.order()` twice?

etpinard: one for marker nodes, one for text nodes

alexcjohnson: ah right - didn't notice `join` got reassigned between 👍

alexcjohnson:
Looks great, nice work! 💃

Issue #431, opened 2016-04-15
Use AppVeyor for karma/jasmine CI testing on IE11


etpinard:
_... and maybe on MS Edge too, [down the road](https://github.com/nicolasmccurdy/karma-edge-launcher/issues/12)._

I've made some prelim. work on the [`appveyor`](https://github.com/plotly/plotly.js/compare/appveyor) branch where [`karma-ie-launcher`](https://github.com/karma-runner/karma-ie-launcher) is used to boot up IE.
### Issues that I've ran into
- all our tests that use our [mouseEvent wrapper](https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/mouse_event.js) are failing (because `window.mouseEvent` does not work in IE11, see prelim attempt for writing a wrapper [here](http://codepen.io/etpinard/pen/reYWaJ))
- I had to use `npm install --force` so that `node-sass` installs properly run after run. This has the side effect of making our AppVeyor test runs > 10 minutes long. We have to find a better way of caching packages.

Full CI outputs: https://ci.appveyor.com/project/etpinard/plotly-js/build/7

Any help from anyone with access to a Windows machine and/or with AppVeyor expertise would be greatly appreciated.

mdtusz:
Interesting. Why was/is the node-sass installation failing?

We should probably add

``` yml
cache:
  - $PATH_TO_PLOTLYJS/node_modules     # or maybe it would be a backslash...
```

to the `appveyor.yml` though.

etpinard:
> Interesting. Why was/is the node-sass installation failing?

I got some `Error: EPERM: operation not permitted` off this [build](https://ci.appveyor.com/project/etpinard/plotly-js/build/3). This error apparently has happened to other folks (e.g. [here](http://help.appveyor.com/discussions/problems/3147-error-eperm-operation-not-permitted-rename-on-npm-install)).

etpinard:
Note, we don't have to run all our jasmine tests there. Most of our pixel-based-assertions will probably fail, getting those up-to-speed for IE will be too much of a pain for little gain. We could either tag a few existing tests with a new `@IE` tag or add a seperate suite.

PR #1715, opened 2017-05-23
add empty protocol to link whitelist for IE relative links

alexcjohnson:
Currently relative links (`<a href="/path/to/file">Link!</a>`) that users insert in their plots fail in IE, because it treats the protocol of such a link as `''` - other browsers have these inherit from the page you're in. This fixes that.

@etpinard I can't think of a way to test this (since I can't find any way on Chrome or FF to generate an empty protocol) can you?

@scjody I don't see any way to turn this into an XSS attack, do you?

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -254,11 +254,11 @@ var TAG_CLOSE = {
 
 /*
  * Whitelist of protocols in user-supplied urls. Mostly we want to avoid javascript
- * and related attack vectors. The empty string is there for IE, that treats
- * relative paths as having no protocol, while other browsers have these explicitly
- * inherit the protocol of the page they're in.
+ * and related attack vectors. The empty items are there for IE, that in various
+ * versions treats relative paths as having different flavors of no protocol, while
+ * other browsers have these explicitly inherit the protocol of the page they're in.
  */
-var PROTOCOLS = ['http:', 'https:', 'mailto:', ''];
+var PROTOCOLS = ['http:', 'https:', 'mailto:', '', undefined, ':'];
 
 var STRIP_TAGS = new RegExp('</?(' + Object.keys(TAG_STYLES).join('|') + ')( [^>]*)?/?>', 'g');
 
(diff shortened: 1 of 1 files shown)


etpinard:
> can't think of a way to test this (since I can't find any way on Chrome or FF to generate an empty protocol) can you?

Maybe it's time to resurrect https://github.com/plotly/plotly.js/issues/431? Otherwise, no I can't think of an other way.

scjody:
This sounds OK from a 🐍 point of view to me.

I checked on our Windows machine. In Edge, `/cats` has a protocol of `"http:"` like in Chrome and FF. In IE, `/cats` has a protocol of `undefined`.

More critically, for `data:` URLs, in Edge `data:text/html,<script>alert();</script>` has a protocol of `data:`, and in IE11 it has a protocol of `data`. So we should be good here.

The only issue I can see is that I'm not sure the new test will actually work. In IE11:
```
''.indexOf(undefined)
-1
```

@alexcjohnson Have you been able to test this in IE11? I know Browserling offers access for one thing.

alexcjohnson:
> In Edge, `/cats` has a protocol of `"http:"` like in Chrome and FF.

Ah great - I didn't test Edge, assumed it would behave like IE but even better if it behaves like Chrome and FF!

> In IE, `/cats` has a protocol of `undefined`.

Huh, on Browserstack, IE11 in Win10, I saw `''` and this fix worked. But I can change it to accept anything falsey, still seem reasonable?

alexcjohnson:
Oh, even falsey isn't enough - IE9 gives `':'` for the protocol here 🙄

scjody:
Wow, weird.

Yeah, anything falsey seems reasonable to me.

alexcjohnson:
Alright, @etpinard especially in light of all the divergent IE behaviors we've seen I don't see much point trying to test this... but at least this fix is now linked forever to #431 in case we ever do push that one through.

Anything else?

etpinard:
:dancer: for me.

Issue #1713, opened 2017-05-22
Wrong fullData passed `plotly_click` listeners after zoom


n-riesco:
The object `plotlyEvent` passed to a listener to `plotly_click` events contains amongst other the properties:
- `plotlyEvent.points[0].data` = trace before transforms
- `plotlyEvent.points[0].fullData` = trace after transforms

This [codepen](http://codepen.io/n-riesco/pen/JNwJgr) shows a chart with a `sort` transform that reorders the circles from `A, B, C` to `B, C, A`. It also defines a `plotly_click` listener that outputs the order in `data` and `fullData`.

The screencast below shows how before zooming the order in `fullData` is `B, C, A` but after zooming the order changes to `A, B, C`.

![peek 2017-05-22 21-36](https://cloud.githubusercontent.com/assets/6199391/26327307/c89e2bdc-3f36-11e7-9dba-b00268438038.gif)

etpinard:
> plotlyEvent.points[0].data = trace before transforms
> plotlyEvent.points[0].fullData = trace after transforms

That's the intended behavior - at least to my eyes. Can you elaborate on why this is a big deal for the spotlight plugin?

n-riesco:
Before the zoom, when I click I get:
```js
plotlyEvent.points[0].data.text = ['B', 'C', 'A']
plotlyEvent.points[0].fullData.text = ['A', 'B', 'C']
```
This is expected because of the `sort` transform.

The next time I click (after zooming) I get:
```js
plotlyEvent.points[0].data.text = ['A', 'B', 'C']
plotlyEvent.points[0].fullData.text = ['A', 'B', 'C']
```
The issue is that `plotlyEvent.points[0].data.text` isn't sorted any longer.

etpinard:
Ah I see. Thanks for pointing this out @n-riesco !

It won't be pretty, but I'll figure something out.

PR #1717, opened 2017-05-23
Transformed array on zoom fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1713

As promised, it's not pretty. cc @n-riesco @alexcjohnson

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -523,18 +523,18 @@ function remapTransformedArrays(cd0, newTrace) {
     var oldTrace = cd0.trace;
     var arrayAttrs = PlotSchema.findArrayAttributes(oldTrace);
     var transformedArrayHash = {};
-    var i, ast;
+    var i, astr;
 
     for(i = 0; i < arrayAttrs.length; i++) {
-        ast = arrayAttrs[i];
-        transformedArrayHash[ast] = Lib.nestedProperty(oldTrace, ast).get().slice();
+        astr = arrayAttrs[i];
+        transformedArrayHash[astr] = Lib.nestedProperty(oldTrace, astr).get().slice();
     }
 
     cd0.trace = newTrace;
 
     for(i = 0; i < arrayAttrs.length; i++) {
-        ast = arrayAttrs[i];
-        Lib.nestedProperty(cd0.trace, ast).set(transformedArrayHash[ast]);
+        astr = arrayAttrs[i];
+        Lib.nestedProperty(cd0.trace, astr).set(transformedArrayHash[astr]);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -505,12 +506,38 @@ plots.supplyDefaults = function(gd) {
     // update object references in calcdata
     if((gd.calcdata || []).length === newFullData.length) {
```
etpinard: This block update the `gd.calcdata[i][0].trace` reference so that the plot step can use the correct `fullData` attribute values even in cases where `Plots.doCalcdata` isn't called (e.g. on zoom).

src/plots/plots.js
```diff
@@ -505,12 +506,38 @@ plots.supplyDefaults = function(gd) {
...
+    for(i = 0; i < arrayAttrs.length; i++) {
+        ast = arrayAttrs[i];
+        transformedArrayHash[ast] = Lib.nestedProperty(oldTrace, ast).get().slice();
+    }
+
+    cd0.trace = newTrace;
+
+    for(i = 0; i < arrayAttrs.length; i++) {
+        ast = arrayAttrs[i];
+        Lib.nestedProperty(cd0.trace, ast).set(transformedArrayHash[ast]);
```
etpinard: Here we make sure that the array attributes in `gd.calcdata[i][0].trace` match the transformed array post calc transform. Other non-data-array attributes are left untouched.

src/plots/plots.js
```diff
@@ -505,12 +506,38 @@ plots.supplyDefaults = function(gd) {
...
+            }
         }
     }
 };
 
+function remapTransformedArrays(cd0, newTrace) {
+    var oldTrace = cd0.trace;
+    var arrayAttrs = PlotSchema.findArrayAttributes(oldTrace);
+    var transformedArrayHash = {};
+    var i, ast;
```
alexcjohnson: 🐄  just so I don't get confused, `ast` makes me think "abstract syntax tree", elsewhere we call this `astr`

etpinard: Oh right, `astr` is what I had in mind. Something wrong happened between brain and keyboard.

alexcjohnson:
Nice use of `PlotSchema` to keep this clean!

Yeah, annoying that we have to do this (for the record from a separate chat: the underlying issue is that a `calcTransform` can modify `fullData`, which is really upstream of the `calc` step that it's run in. We haven't come up with a cleaner way to separate this but will keep ruminating on it)

:dancer: after https://github.com/plotly/plotly.js/pull/1717#discussion_r118102595

src/plots/plots.js
etpinard: done in 0a4ce1f

rreusser:
Is it clear what needs to happen with this? I'm still having just a bit of trouble understanding exactly why this is needed and then the downstream problem is with animation.

etpinard:
> Is it clear what needs to happen with this?

First look at the added assertions. If that's not clear, slack me.

Issue #1562, opened 2017-04-06
scatter3d error bars spit out READER WARNING


etpinard:
See http://codepen.io/etpinard/pen/dvEQQj?editors=0010

@dfcreative

dfcreative:
[:hammer:](https://github.com/gl-vis/gl-error3d/pull/2)

PR #1718, opened 2017-05-23
bump gl-error3d to ^1.0.5

etpinard:
to grab https://github.com/gl-vis/gl-error3d/pull/2 and :lock: down the https://github.com/plotly/plotly.js/issues/1562 fix.


etpinard:
Shout out to @dfcreative for the fix!

Issue #1710, opened 2017-05-20
zmin and zmax properties not implemented for mesh3d types colorbar


precisesimulation:
Perhaps just a copy paste bug, but according to:

https://plot.ly/javascript/reference/#mesh3d-colorscale

the mesh3d colorbar should be able to be controlled with the zmin and zmax properties, however no such properties are available.

Also it seems the colorbar is attached to the z value which doesn't seem to be possible to change or general. Seems to me that having the colorbar correlate to the intensity field would make more sense where one could then attach any value one wants (x, y, z, or some other value).

etpinard:
Thanks very much for pointing this out.

`gl-mesh3d` will need to patched in a similar manner to `gl-surface3d` in PR https://github.com/gl-vis/gl-surface3d/pull/7

PR #1719, opened 2017-05-23
Mesh3d colorscale settings

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1710

requires:

- https://github.com/gl-vis/gl-mesh3d/pull/10 set for `gl-mesh3d@1.3.0`
- https://github.com/gl-vis/gl-error3d/pull/3 set for `gl-error3d@1.0.6`


test/image/mocks/gl3d_tetrahedra.json
```diff
@@ -16,6 +16,8 @@
         [1, "rgb(0, 0, 255)"]
       ],
       "intensity": [0, 0.33, 0.66, 1],
+      "cmin": -0.5,
+      "cmax": 0.5,
```
etpinard: Weird. That made the axis labels disappear. I might have to revert back those `gl-error3d` dependencies bump of commit https://github.com/gl-vis/gl-error3d/commit/d756284bbfdaf3dd1f4ce7e69ae8085b9b6d2a64

etpinard: fixed in f55c6f5dad3c302e265a246d47c2b15d563daa52 via https://github.com/gl-vis/gl-error3d/pull/3

etpinard:
All green on CI, ready for review @alexcjohnson

test/jasmine/tests/mesh3d_test.js
```diff
@@ -0,0 +1,57 @@
...
+        })
+        .then(function() {
+            _assert([false, 0, 10], [false, 0, 10]);
+
+            return Plotly.restyle(gd, 'cauto', true);
+        })
+        .then(function() {
+            _assert([true, 0, 3], [true, 0, 3]);
+
+            return Plotly.purge(gd);
```
alexcjohnson: why purge?

src/traces/mesh3d/defaults.js
```diff
@@ -77,23 +76,12 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
 
     if('intensity' in traceIn) {
         coerce('intensity');
-        coerce('showscale', true);
-    }
-    else {
+        colorscaleDefaults(traceIn, traceOut, layout, coerce, {prefix: '', cLetter: 'c'});
```
alexcjohnson: nice when you can remove code to add features and fix bugs 🎉

alexcjohnson:
So I guess previously these colorscale ranges were erroneously taken from the `z` span of the data? Looks great! 💃

test/jasmine/tests/mesh3d_test.js
etpinard: To destroy the webgl context. 

We should perhaps add that Plotly.purge statement in destroyGraphDiv?

etpinard:
> So I guess previously these colorscale ranges were erroneously taken from the z span of the data?

Exactly. `mesh3d` was using the heatmap [colorbar](https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/colorbar.js) logic previously.

test/jasmine/tests/mesh3d_test.js
alexcjohnson: ah interesting... yeah, would be good to put that in `destroyGraphDiv` - but it's frustrating that deleting `gd` from the document isn't sufficient to 🔪 the context - I wonder if we can do something to [watch for removing](https://stackoverflow.com/questions/20156453/how-to-detect-element-being-added-removed-from-dom-element) `gd` and purge automatically? If you think that's not a good idea, we should at least make sure this is well documented somewhere prominent.

Issue #1721, opened 2017-05-24
Fillcolor animation not happening


rreusser:
From @cpsievert, seems fillcolor is not animating. It just transitions instantaneously. This should be easy to implement.

Example: https://codepen.io/rsreusser/pen/VbRwpj?editors=0010

etpinard:
done in #1722

PR #1722, opened 2017-05-24
Transition fillcolor

rreusser:
This PR solves #1721 and transitions fillcolor as expected, which currently only updates on an actual redraw.

An example in which it works: http://rickyreusser.com/demos/plotly-unsupported/fillcolor-animation.html

/cc @cpsievert

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -156,6 +156,16 @@ drawing.dashStyle = function(dash, lineWidth) {
     return dash;
 };
 
+// Same as fillGroupStyle, except in this case the selection may be a transition
+drawing.singleFillStyle = function(sel) {
+    var node = d3.select(sel.node());
+    var data = node.data();
+    var fillcolor = (((data[0] || [])[0] || {}).trace || {}).fillcolor;
+    if(fillcolor) {
+        sel.call(Color.fill, fillcolor);
+    }
+};
+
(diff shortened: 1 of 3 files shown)


rreusser:
Test failure is:

```
Chrome 54.0.2840 (Linux 0.0.0) Test sort transform interactions: does not preserve hover/click `pointNumber` value FAILED
```

@etpinard is that to be expected? Is that an intentionally failing and not yet resolved test?

etpinard:
> @etpinard is that to be expected?

Nop. Your patch shouldn't brake that test.

etpinard:
@rreusser I just pressed rebuild on CI, it might _just_ be an intermittent failure.

rreusser:
Sounds good! Oh… sort suggested deterministic, but hover suggests maybe a little flaky. Thanks!

etpinard:
@rreusser ok if I change the label to `type: feature`?

rreusser:
Yup! I debated. Do the labels have an effect/use, or is it just for record-keeping?

etpinard:
> Do the labels have an effect/use

To make sure `type: feature` don't sneak into patch releases.

rreusser:
@etpinard I've rebased and looks like now passing 👍

etpinard:
Great news :dancer:

PR #1723, opened 2017-05-24
Switching to a SVG presentation attribute to assist Microsoft browsers

monfera:
There is still insufficient support for CSS styles on SVG elements in IE: https://github.com/ropensci/plotly/issues/1023#issuecomment-303529546

### DIFF
--- a/src/traces/sankey/render.js
+++ b/src/traces/sankey/render.js
@@ -264,7 +264,7 @@ function salientEnough(d) {
 }
 
 function linksTransform(d) {
-    return d.horizontal ? 'matrix(1,0,0,1,0,0)' : 'matrix(0,1,1,0,0,0)';
+    return d.horizontal ? 'matrix(1 0 0 1 0 0)' : 'matrix(0 1 1 0 0 0)';
 }
 
 function textGuidePath(d) {
@@ -444,11 +444,11 @@ module.exports = function(svg, styledData, layout, callbacks) {
         .append('g')
         .classed('sankeyLinks', true)
         .style('fill', 'none')
-        .style('transform', linksTransform);
+        .attr('transform', linksTransform);
 
     sankeyLinks.transition()
         .ease(c.ease).duration(c.duration)
-        .style('transform', linksTransform);
+        .attr('transform', linksTransform);
 
     var sankeyLink = sankeyLinks.selectAll('.sankeyLink')
         .data(function(d) {


src/traces/sankey/render.js
```diff
@@ -444,11 +444,11 @@ module.exports = function(svg, styledData, layout, callbacks) {
         .append('g')
         .classed('sankeyLinks', true)
         .style('fill', 'none')
-        .style('transform', linksTransform);
```
etpinard: Good catch. 

Would anyone be interested in making @alexcjohnson 's [`strict-d3`](https://github.com/plotly/plotly.js/blob/master/test/image/strict-d3.js) wrapper throw an error on `selection.style('transform', <>)`?

monfera: Good idea! I'm working on it now in a separate PR.

monfera: ^ cc @alexcjohnson

monfera: The tweening effect when switching between horizontal and vertical layout was a bit like a circus attraction, and interestingly, the switch to `attr` made it noticeably worse, crossing into creepy territory. I'm reworking a chunk of code so that links and nodes don't "fly" on separate arcs but they remain joint instead. It also stops the node rectangle shape morphing too. As I think this PR stands on its legs though (fixing the error) it'll be a separate PR based on this one.

monfera:
The tweening effect when switching between horizontal and vertical layout was a bit like a circus attraction, and interestingly, the switch to attr made it noticeably worse, crossing into creepy territory. I'm reworking a chunk of code so that links and nodes don't "fly" on separate arcs but they remain joint instead. It also stops the node rectangle shape morphing too. As I think this PR stands on its legs though (fixing the error) it'll be a separate PR based on this one.

src/traces/sankey/render.js
monfera: We need to check if the `style`d element is an SVG element or not. (I'm planning to just check the first element in the selection with `sel.node()` as selecting both HTML and SVG elements in the same selection is not commonly done.) A linting problem arises:
```
/Users/robert/repos/plotly/plotly.js/test/image/strict-d3.js
  42:38  error  'SVGElement' is not defined  no-undef
```
Interestingly the linter doesn't balk on a [similar usage](https://github.com/plotly/plotly.js/blob/master/src/lib/index.js#L426).

Have we got an established workaround, or should I just lint-define the `SVGElement` locally?

alexcjohnson: I don't think we have an established workaround, either lint defining it or `var SVGElement = window.SVGElement` would be fine by me.

etpinard: > var SVGElement = window.SVGElement

Yep, that's the way to go here :+1:

alexcjohnson:
> The tweening effect when switching between horizontal and vertical layout was a bit like a circus attraction, and interestingly, the switch to attr made it noticeably worse

Interesting indeed... I certainly wouldn't have expected that to make a difference.

Personally I don't see animating an orientation change as being particularly useful to viewers (did we have an explicit requirement to do this?), so I'd be at least as happy just snapping to the new layout - particularly if it's going to take a noticeable effort to fix this.

alexcjohnson:
This looks to be all set yes? Particularly since #1729 is merged so it's tested...
💃

PR #1726, opened 2017-05-25
support target and popup attributes in pseudo-html text links

alexcjohnson:
Adds two new attributes for more flexibility to determine where links in our pseudo-html text will open:
- `target`, which is preferred over `xlink:show` in modern browsers and has more functionality - but I still include `xlink:show` as a fallback: `'replace'` if target is `'_self'`, `'_parent'`, or `'_top'` and `'new'` for `'_blank'` or a named frame. Some `xlink:show` implementations seem to accept more values to partially bridge the gap with `target` but some don't, and I believe most that do will look at `target` anyway.
- `popup`, which doesn't exist in html but we need a way to insert the "window features" string without accepting arbitrary javascript, so inventing a new attribute seemed like the easiest way. Note that `target` is sent as the second arg (`name`) to `window.open`, which lets you send subsequent requests to the same popup (or, uselessly, to not make a popup at all if you use `'_self'` etc)

So usage would look like:
`'<a href="https://plot.ly" target="fred" popup="width=500,height=400,menubar=no">link</a>'`

@etpinard look reasonable?
@scjody quick 🐍  review? I poked at it for a while and couldn't find any problems but would love a second set of eyes. Note that both attribute values get `encodeForHTML` immediately on extraction.

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -282,11 +282,36 @@ var SPLIT_TAGS = /(<[^<>]*>)/;
 
 var ONE_TAG = /<(\/?)([^ >]*)(\s+(.*))?>/i;
 
-// Style and href: pull them out of either single or double quotes.
-// Because we hack in other attributes with style (sub & sup), drop any trailing
-// semicolon in user-supplied styles so we can consistently append the tag-dependent style
+/*
+ * style and href: pull them out of either single or double quotes. Also
+ * - target: (_blank|_self|_parent|_top|framename)
+ *     note that you can't use target to get a popup but if you use popup,
+ *     a `framename` will be passed along as the name of the popup window.
+ *     per the spec, cannot contain whitespace.
+ *     for backward compatibility we default to '_blank'
+ * - popup: a custom one for us to enable popup (new window) links. String
+ *     for window.open -> strWindowFeatures, like 'menubar=yes,width=500,height=550'
+ *     note that at least in Chrome, you need to give at least one property
+ *     in this string or the page will open in a new tab anyway. We follow this
+ *     convention and will not make a popup if this string is empty.
+ *     per the spec, cannot contain whitespace.
+ *
+ * Because we hack in other attributes with style (sub & sup), drop any trailing
+ * semicolon in user-supplied styles so we can consistently append the tag-dependent style
+ */
 var STYLEMATCH = /(^|[\s"'])style\s*=\s*("([^"]*);?"|'([^']*);?')/i;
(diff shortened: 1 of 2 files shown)


src/lib/svg_text_utils.js
```diff
@@ -282,11 +282,36 @@ var SPLIT_TAGS = /(<[^<>]*>)/;
...
+ *     note that you can't use target to get a popup but if you use popup,
+ *     a `framename` will be passed along as the name of the popup window.
+ *     per the spec, cannot contain whitespace.
+ *     for backward compatibility we default to '_blank'
+ * - popup: a custom one for us to enable popup (new window) links. String
+ *     for window.open -> strWindowFeatures, like 'menubar=yes,width=500,height=550'
+ *     note that at least in Chrome, you need to give at least one property
+ *     in this string or the page will open in a new tab anyway. We follow this
+ *     convention and will not make a popup if this string is empty.
+ *     per the spec, cannot contain whitespace.
```
alexcjohnson: seems to be a point of browser inconsistency whether it still works with whitespace... alternatively I guess we could accept whitespace in the input but strip it from the output, but this was easier.

src/lib/svg_text_utils.js
```diff
@@ -297,14 +322,14 @@ exports.plainText = function(_str) {
 };
 
 function replaceFromMapObject(_str, list) {
-    var out = _str || '';
+    if(!_str) return '';
```
alexcjohnson: 🐎  no need to loop over all these replacements on an empty string

etpinard:
> @etpinard look reasonable?

Very reasonable. I'll leave it out to @scjody to review to :dancer: over the :snake: 

---

Now, we should document these new pseudo-html features. We already have some info under the annotations [`text`](https://github.com/plotly/plotly.js/blob/6cc5ffd7efa73dab266e2e6e1840797b0b2bb06c/src/components/annotations/attributes.js#L34) declaration, but really this extends to all `text` fields printed in SVG cc @cldougl

alexcjohnson:
> Now, we should document these new pseudo-html features

Excellent point. Our pseudo-html applies to really anywhere we display text: annotations, plot and axis titles, trace names. Also tick labels (categories, ticktext, tickprefix, ticksuffix...) and trace text and hovertext though those are generally unclickable so this *particular* change doesn't matter there.

Seems like this format is too complex to include outright in all of these attribute, so we need some dedicated text formatting page that all of these attributes can somehow reference?

etpinard:
> , so we need some dedicated text formatting page that all of these attributes can somehow reference?

:+1:

etpinard:
Maybe open up an issue on https://github.com/plotly/documentation/issues/new so we don't forget about it.

scjody:
🐍 looks good to me.

PR #1727, opened 2017-05-25
Fixup sort transform event data text using

etpinard:
This test doesn't behave well on CI as noticed by @alexcjohnson and @rreusser 

So to remedy this problem, we'll test sort transform event data using hover events only:

- hover and click when invoked programmatically back-to-back don't behave well.
- this test case makes assertions on the event data (i.e. `gd._hoverdata`) which is the same for hover and click
- use `delete gd._lastHoverTime` to :hocho: that annoying wait promise

### DIFF
--- a/test/jasmine/tests/transform_sort_test.js
+++ b/test/jasmine/tests/transform_sort_test.js
@@ -273,7 +273,7 @@ describe('Test sort transform interactions:', function() {
         }
 
         function hover(gd, id) {
-            return new Promise(function(resolve) {
+            return new Promise(function(resolve, reject) {
                 gd.once('plotly_hover', function(eventData) {
                     delete gd._lastHoverTime;
                     resolve(eventData);
@@ -281,6 +281,10 @@ describe('Test sort transform interactions:', function() {
 
                 var pos = getPxPos(gd, id);
                 mouseEvent('mousemove', pos[0], pos[1]);
+
+                setTimeout(function() {
+                    reject('plotly_hover did not get called!');
+                }, 100);
             });
         }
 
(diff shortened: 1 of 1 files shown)


test/jasmine/tests/transform_sort_test.js
```diff
@@ -275,6 +275,7 @@ describe('Test sort transform interactions:', function() {
         function hover(gd, id) {
             return new Promise(function(resolve) {
                 gd.once('plotly_hover', function(eventData) {
```
alexcjohnson: Ah, this pattern is probably the reason we're timing out blind - when we wait for an event to resolve our promise, we need some timeout that causes an explicit error if that event never fires (and gets aborted when the event does fire), so we can see what's happening.

`delete gd._lastHoverTime` may well solve the issue here, but in case it doesn't (or something else crops up when we alter/extend this later) can you add that?

test/jasmine/tests/transform_sort_test.js
```diff
@@ -273,14 +273,18 @@ describe('Test sort transform interactions:', function() {
...
                     delete gd._lastHoverTime;
                     resolve(eventData);
                 });
 
                 var pos = getPxPos(gd, id);
                 mouseEvent('mousemove', pos[0], pos[1]);
+
+                setTimeout(function() {
+                    reject('plotly_hover did not get called!');
```
etpinard: @alexcjohnson something like this?

e.g. if we comment out the above `mouseEvent()` we get:

image

from `fail_test.js`

alexcjohnson: yeah, that's great! Right, and you don't even need to cancel the timeout because after resolve is called, reject doesn't do anything... nice!

In principle to tell *precisely* what's going on we probably need a message - but that can be added during debugging if we get to that point.

alexcjohnson:
💃

etpinard:
cc @rreusser you might want to merge `master` into https://github.com/plotly/plotly.js/pull/1722

PR #1728, opened 2017-05-25
[TEST] Sankey rotation and IE SVG CSS transform CI test

monfera:


.eslintrc
```diff
@@ -16,7 +16,8 @@
     "Uint8Array": true,
     "Int16Array": true,
     "Int32Array": true,
-    "ArrayBuffer": true
+    "ArrayBuffer": true,
+    "SVGElement": false
```
etpinard: maybe `"SVGElement": true` will make eslint stop complaining?

PR #1729, opened 2017-05-25
Prohibit CSS transform usage on SVG elements due to IE

monfera:
As [suggested](https://github.com/plotly/plotly.js/pull/1723#discussion_r118360733) by @etpinard


etpinard:
Nicely done. Thank very much @monfera 

:dancer: when tests pass

PR #1730, opened 2017-05-25
Fix scene-bounds logic for non-linear axis set ranges

etpinard:
Setting axis ranges of non-`linear` 3d axes is broken at the moment and has been probably for long time  if not _forever_. I can't believe nobody noticed this before. Moreover, I can't believe we didn't have any mocks testing that.

Oh well, the new `gl3d_set-ranges` mock should :lock: this one down for good :tada: 

cc @alexcjohnson


alexcjohnson:
Oops 🙃  💃

PR #1731, opened 2017-05-25
Sankey rotation between horizontal and vertical

monfera:
Due to an [IE shortcoming](https://github.com/plotly/plotly.js/pull/1729) a [fix](https://github.com/plotly/plotly.js/pull/1723) was necessary that further deteriorated an already dubious transition between horizontal and vertical Sankey.

This PR has a profound impact on what happens when running the `sankey_energy.json` mock and switching it from horizontal to vertical and back. As a side effect, while the source code is a tad longer, a lot of lines became shorter and simpler as the new logic has fewer conditionalization on the orientation.


alexcjohnson:
Personally the part of this PR I'm most excited about is that the node/link alignment is cleaner now in vertical orientation (as shown in the baseline image change) 🎉 

The transition is good too 😉 

💃

PR #1732, opened 2017-05-25
Fixup camera interaction relayout event data

etpinard:
Another bug I found when working on gl3d annotations.

On scene drag/pan, we fire a `plotly_relayout` event mocking the equivalent `Plotly.relayout` call that would generate the current view.

Previously, the mocked event data looked:

```js
{ scene: { eye: {/**/}, up: {/**/}, center: {/**/} }
```
(or `scene2: // ...` for multiple-scene graphs) which isn't compatible with our plot schema

Now with this PR, the mocked event data is:

```js
{ 'scene.camera': { eye: {/**/}, up: {/**/}, center: {/**/} }
```

cc @monfera


alexcjohnson:
nice catch! 💃

PR #1736, opened 2017-05-26
Add warning if an invalid scale is passed

mdamien:
When an invalid scale is passed, it's silently ignored if it's invalid. This makes it painful to debug (did I use the right parameter ? does this chart support colorscales ?). So I propose to add a warning for invalid scales.

Of course, people still have to do `Plotly.setPlotConfig({logging: true});`

### DIFF
--- a/src/components/colorscale/is_valid_scale.js
+++ b/src/components/colorscale/is_valid_scale.js
@@ -9,11 +9,15 @@
 
 'use strict';
 
+var Lib = require('../../lib');
 var scales = require('./scales');
 var isValidScaleArray = require('./is_valid_scale_array');
 
 
 module.exports = function isValidScale(scl) {
-    if(scales[scl] !== undefined) return true;
-    else return isValidScaleArray(scl);
+    if(scales[scl] !== undefined || isValidScaleArray(scl)) {
+        return true;
+    }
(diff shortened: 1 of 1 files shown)


etpinard:
Thanks very much for the PR!

Personally, I would prefer making sure the _very-undocumented_ [`Plotly.validate`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/validate_test.js) catches all invalid colorscales which [should](https://github.com/plotly/plotly.js/blob/78b82b85206bde93c7117706278ac698b480139d/test/jasmine/tests/validate_test.js#L151) already be the case.

alexcjohnson:
@etpinard I seem to recall we talked at some point about having a config setting that runs `Plotly.validate` during `supplyDefaults` or something, kind of a debug mode - is there an issue for this? @mdamien is right of course that this kind of error is tough to catch, but that's true of every attribute really, they'll all silently ignore invalid input.

etpinard:
> I seem to recall we talked at some point about having a config setting that runs Plotly.validate during supplyDefaults or something, kind of a debug mode - is there an issue for this?

Here' it is :arrow_right: https://github.com/plotly/plotly.js/issues/1741

PR #1737, opened 2017-05-26
clear calcdata before redoing it in animations

alexcjohnson:
Fixes the issue we're seeing with calcTransforms + animations, which was introduced in #1717 
Then the question is: what do we test this with and how?

cc @etpinard @rreusser


src/plots/plots.js
```diff
@@ -1813,6 +1813,10 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
         // of essentially the whole supplyDefaults step, so that it seems sensible to just use
         // supplyDefaults even though it's heavier than would otherwise be desired for
         // transitions:
+
+        // first delete calcdata so supplyDefaults knows a calc step is coming
+        delete gd.calcdata;
```
etpinard: Looks like a solid fix. It will slow down `Plotly.animate` though.

etpinard: There's some [logic](https://github.com/plotly/plotly.js/blob/6b7c9d7aeba575c3ff5a35a03e111c051e001ed7/src/plots/plots.js#L2026) in `Plots.doCalcdata` involving old versus new calcdata that might be obsolete now.

alexcjohnson: will it? We're calling `doCalcdata` no matter what so I'd think this will actually speed it up a tiny bit!

etpinard: Good point :+1:

rreusser:
Here's at least a codepen that isolates this effect (plotly-latest works though so it's not visible on the codepen)

https://codepen.io/rsreusser/pen/rmbrXg?editors=0010

etpinard:
Great thanks!

Issue #699, opened 2016-06-28
Download plot as a png not working in IE from 1.13.0


fixxxr:
From release 1.13.0 the snapshot in IE (svg) seems broken. 
tested with `type: 'surface'` and `type: 'contour'` and for the first it produces a blank svg while for the second, the snapshot seems to stop before the saving of the svg file. 

From 1.12.0 and below the snapshot for `type: 'contour'` has the same behavior as described above.

etpinard:
Thanks for the report.

We did make some changes to our to-image code in `1.13.0` (See PR https://github.com/plotly/plotly.js/pull/604). 

Can you confirm or infirm the following:

Donwload plot with

- `type: 'surface'` does not work in IE11 with plotly.js `1.13.0` ?
- `type: 'surface'` works in IE11 with plotly.js `1.12.0` ?
- `type: 'surface'` works in IE11 with plotly.js `1.11.0` ?
- `type: 'contour'` does not work in IE11 with plotly.js `1.13.0` ?
- `type: 'contour'` does not work in IE11 with plotly.js `1.12.0` ?
- `type: 'contour'` does not work in IE11 with plotly.js `1.11.0` ?

To-image support for IE11 was introduced in 1.11.0 (see PR https://github.com/plotly/plotly.js/pull/446), so there no chance of it working in versions < `1.11.0`.

cc @timelyportfolio

fixxxr:
in order:

1. Yes
2. Yes
3. Yes
4. Yes
5. Yes
6. Yes

Below 1.12.0 is for down to 1.11.0

Thanks

timelyportfolio:
@fixxxr, I'll check.  Could you provide the code you are using to make your `png`?  Sorry for the troubles.

fixxxr:
I am using the snapshot button.
Also I tested the aforementioned releases with the example posted here: [plotly.surface](https://plot.ly/javascript/3d-surface-plots/#topographical-3d-surface-plot), just to be sure that it wasn't related only to my case. I confirm that the issue is reproducible with the code posted on Plotly's page.

timelyportfolio:
Looking into this, and I think I see the issue. I'll keep working on it and hopefully by tomorrow I will have it fixed.

fixxxr:
@timelyportfolio, thank you very much

fixxxr:
hi there, 
any news on this issue?

etpinard:
From @vsabeti :

[...]

PR #1740, opened 2017-05-28
Fix RE used to correct quotes used in URLs for IE.

bobisch:
The `svgToImg()` function tries to normalize the SVG for compatibility
with IE.  However, currently it use a regular expression that is too
greedy when matching the attribute values that need quotes changed from
single to double.  The result is that invalid XML is generated and will
cause IE11 to fail with a `XML 5607` error and `svgToImg()` will never
return.  Edge will not complain but will receive invalid XML, which
causes problems when used.

This change restricts matching to only data surrounded by single quotes.

Seems like a jasmine test should be added for `svgToImg()` along the
lines of

  `it('Returns valid XML',...)`

Also addresses: https://community.plot.ly/t/plotly-toimage-xml5607-error-in-ie11/4223

### DIFF
--- a/src/snapshot/svgtoimg.js
+++ b/src/snapshot/svgtoimg.js
@@ -35,7 +35,7 @@ function svgToImg(opts) {
             // url in svg are single quoted
             //   since we changed double to single
             //   we'll need to change these to double-quoted
-            svg = svg.replace(/(\('#)(.*)('\))/gi, '(\"$2\")');
+            svg = svg.replace(/(\('#)([^']*)('\))/gi, '(\"$2\")');
             // font names with spaces will be escaped single-quoted
             //   we'll need to change these to double-quoted
             svg = svg.replace(/(\\')/gi, '\"');


etpinard:
Thanks very much for the PR!

That might be enough to fix https://github.com/plotly/plotly.js/issues/699

etpinard:
@bobisch I'm having problem replicating the problem.

- Can you confirm that https://codepen.io/etpinard/pen/WjqNLE works as desired in IE11?
- Can you share the `"data"` / "layout"` JSON that led to the `XML 5607` error?

bobisch:
@etpinard - sorry, I thought we had some relevant examples, but I guess none highlight this particular issue.  Unfortunately, I don't think this will do too much to really fix #699 and related issues since those probably have more to do with the IE limitations caused by tainting svg canvas data.  But, this will make it possible to use some work-around(s) such as sending the data to the back-end for rendering or using canvg.

Anyway, to get the error you need to have something that produces at least two `clip-path="url(...` attributes.  In your example I think it works to just add another curve to the plot - see [https://codepen.io/bobisch/pen/JNgXrE](https://codepen.io/bobisch/pen/JNgXrE) and that fails under Edge (or IE) without the patch.

etpinard:
@bobisch thanks for the info!

Bundling [up](https://github.com/etpinard/plotly.js/blob/pr1740-snapshot-ie-fix/dist/plotly.js) your patch and using it in https://codepen.io/etpinard/pen/WjqNLE gives the correct in IE11:

image


:tada::tada::tada::tada::tada::tada::tada:

---

About adding a new test case, I can't find a way to test this logic outside of IE. Mocking the test environment such that it executes this normally IE-only [block](https://github.com/plotly/plotly.js/blob/e37eeaeb73d4d91682bdf84adfa8c871da61a18e/src/snapshot/svgtoimg.js#L34-L41) in `svgtoimg.js` yields a string that Chrome's `DOMParser` can't parse as SVG. We could test the raw string output, but that's too brittle for my tastes. 

What we should really do is implement https://github.com/plotly/plotly.js/issues/431 - so I'd vote for not adding any tests here, merge this in and include it in `1.28.0`. 

Sounds good?

bobisch:
That seems reasonable to me.  Thanks!

Wouldn't it still be reasonable to just add a test that checks to see if `svgToImg()` at least returns a string and that should be valid (parse-able) XML in any browser?  That would catch issues with this code paths mangling of the SVG at least.

etpinard:
> be valid (parse-able) XML in any browser?

That's the problem. XML parsers are different from browser to browser. A string that might be considered valid in one browser won't necessary be valid in another.

Issue #1720, opened 2017-05-24
shapes on category axis are redrawn incorrectly after zoom


cldougl:
example: https://codepen.io/plotly/pen/79661cd85e9e51054039b89d096b0c9a

after zoom/relayout the vertical line is drawn as a rectangle or if `type='line'` is specified the vertical line becomes diagonal.

etpinard:
This is a regression from `1.24.0`. 

Here's working example with `1.23.1`:
https://codepen.io/etpinard/pen/vmqEJg?editors=0010

etpinard:
I suspect PR https://github.com/plotly/plotly.js/pull/1403

etpinard:
Ok, well this is a pretty big problem after all. Annotations, shapes and layout image on `type: 'category'` axes are all broken after zoom, pan or any axis-range relayout calls at the moment.

Diagnostic: in [`Axes.coercePosition`](https://github.com/plotly/plotly.js/blob/a0f833387c0c1a06b234a0b4f026e40a0f867dbc/src/plots/cartesian/axes.js#L117) we attempt to use `ax._categories` to convert string category positions to category indices. Note that, `ax._categories` isn't filled (nor relinked) at that point in [`Plots.supplyDefaults`](https://github.com/plotly/plotly.js/blob/a0f833387c0c1a06b234a0b4f026e40a0f867dbc/src/plots/plots.js#L476). To go around that, annotations, shapes and images defaults are called a second time during [`Plotly.doCalcdata`](https://github.com/plotly/plotly.js/blob/a0f833387c0c1a06b234a0b4f026e40a0f867dbc/src/plots/plots.js#L2112) if `type: 'category'` axes are detected. At this point `ax_.categories` is set, but `Plots.doCalcdata` isn't called on all updates (e.g. not on pan/zoom) whereas `Plotly.supplyDefaults` leading to bad coordinates values in `fullLayout` propagating to the `draw` methods after pan/zoom interactions.

Before PR #1403, annotations and shape [used](https://github.com/plotly/plotly.js/pull/1403/files#diff-05274991eb85aefd482f72514dbda1a5L246) to redo their defaults in their `drawOne` respective methods. We could revert to doing something like that, but we can do better I think. 

[...]

PR #1748, opened 2017-05-30
Improved category axes conversions

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1720

This PR is a rebased version of the [`component-category-axes-relayout-attempt`](https://github.com/plotly/plotly.js/compare/component-category-axes-relayout-attempt) branch with better commit messages.

In brief, category (i.e. string) values of annotation/shape/image position attributes (e.g. `x`, `y`, `x0` ...) referencing `type: 'category'` axes are no longer converted to indices (i.e. numbers) during the defaults step. This implies:

- the annotation/shape/image defaults no longer depend on `ax._categories` (which isn't defined during `Plots.supplyDefaults` which led to :bug: #1720 ) and this ugly category-axes-only [block](https://github.com/plotly/plotly.js/blob/7a281ccbb2046888d2d2587461a99fd51ee63d8e/src/plots/plots.js#L2109-L2115).
- we must redefine what the `r` `Axes.setConvert` data type means for category axes - see commit 	b67657f for more details. This part isn't perfect, but better than what we have currently [accordingly](https://github.com/plotly/plotly.js/commit/b2fa5d551bc5e0cc83f4c3cd7c9009779d558c5f#commitcomment-22332537) to @alexcjohnson

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -10,6 +10,11 @@
 'use strict';
 
 var d3 = require('d3');
+var isNumeric = require('fast-isnumeric');
+
+var numConstants = require('../constants/numerical');
+var FP_SAFE = numConstants.FP_SAFE;
+var BADNUM = numConstants.BADNUM;
 
 var lib = module.exports = {};
 
@@ -86,6 +91,13 @@ lib.filterVisible = require('./filter_visible');
 lib.pushUnique = require('./push_unique');
 
 lib.cleanNumber = require('./clean_number');
+
+lib.num = function num(v) {
+    if(!isNumeric(v)) return BADNUM;
+    v = Number(v);
+    if(v < -FP_SAFE || v > FP_SAFE) return BADNUM;
+    return isNumeric(v) ? Number(v) : BADNUM;
+};
 
 lib.noop = require('./noop');
 lib.identity = require('./identity');
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -92,7 +92,7 @@ lib.pushUnique = require('./push_unique');
 
 lib.cleanNumber = require('./clean_number');
 
-lib.num = function num(v) {
+lib.ensureNumber = function num(v) {
     if(!isNumeric(v)) return BADNUM;
     v = Number(v);
     if(v < -FP_SAFE || v > FP_SAFE) return BADNUM;
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -113,6 +113,12 @@ axes.coercePosition = function(containerOut, gd, coerce, axRef, attr, dflt) {
     containerOut[attr] = ax.cleanPos(pos);
 };
 
+axes.cleanPosition = function(pos, gd, axRef) {
+    var ax = (axRef === 'paper' || axRef === 'pixel') ?
+        { cleanPos: Lib.num } :
+        axes.getFromId(gd, axRef);
+
+    return ax.cleanPos(pos);
 };
 
 axes.getDataToCoordFunc = function(gd, trace, target, targetArray) {
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -99,26 +99,27 @@ axes.coerceRef = function(containerIn, containerOut, gd, attr, dflt, extraOption
  * - for other types: coerce them to numbers
  */
 axes.coercePosition = function(containerOut, gd, coerce, axRef, attr, dflt) {
-    var ax, pos;
+    var cleanPos, pos;
 
     if(axRef === 'paper' || axRef === 'pixel') {
-        ax = { cleanPos: Lib.num };
+        cleanPos = Lib.ensureNumber;
         pos = coerce(attr, dflt);
     } else {
-        ax = axes.getFromId(gd, axRef);
+        var ax = axes.getFromId(gd, axRef);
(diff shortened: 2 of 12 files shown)


src/lib/index.js
```diff
@@ -87,6 +92,13 @@ lib.pushUnique = require('./push_unique');
 
 lib.cleanNumber = require('./clean_number');
 
+lib.num = function num(v) {
```
alexcjohnson: Do we want a more descriptive name for this now that it's not just internal to one module? `ensureNumber`? `safeNumber`?

src/plots/cartesian/axes.js
```diff
@@ -100,33 +99,20 @@ axes.coerceRef = function(containerIn, containerOut, gd, attr, dflt, extraOption
  * - for other types: coerce them to numbers
  */
 axes.coercePosition = function(containerOut, gd, coerce, axRef, attr, dflt) {
-    var pos,
-        newPos;
+    var ax, pos;
 
     if(axRef === 'paper' || axRef === 'pixel') {
+        ax = { cleanPos: Lib.num };
```
alexcjohnson: don't need to pretend we're making an axis anymore... just `cleanPos = Lib.num` or `cleanPos = ax.cleanPos`

src/traces/heatmap/calc.js
```diff
@@ -57,10 +57,15 @@ module.exports = function calc(gd, trace) {
         z = binned.z;
     }
     else {
-        if(hasColumns(trace)) convertColumnData(trace, xa, ya, 'x', 'y', ['z']);
+        if(hasColumns(trace)) {
+            convertColumnData(trace, xa, ya, 'x', 'y', ['z']);
+            x = trace.x;
+            y = trace.y;
```
alexcjohnson: Oh interesting... were we double-converting triplet data on category and date axes before? Do we have a test of these cases?

test/jasmine/tests/annotations_test.js
```diff
@@ -98,16 +98,48 @@ describe('Test annotations', function() {
...
+            });
+
             _supply(layoutIn, layoutOut);
 
+            expect(layoutOut.annotations[0]._xclick).toBe(10, 'paper x');
+            expect(layoutOut.annotations[0]._yclick).toBe(30, 'paper y');
+            expect(layoutOut.annotations[1]._xclick).toBe(1, 'linear');
+            expect(layoutOut.annotations[1]._yclick).toBe(undefined, 'invalid date');
+            expect(layoutOut.annotations[2]._xclick).toBe(2, 'log');
+            expect(layoutOut.annotations[2]._yclick).toBe('A', 'category');
```
alexcjohnson: 🎉

src/traces/heatmap/calc.js
etpinard: > were we double-converting triplet

Right. Previously `ax.d2c` was called twice in `heatmap/calc.js`. For example, on `master` the [`heatmap_xyz-dates-and-categories`](http://rickyreusser.com/plotly-mock-viewer/#heatmap_xyz-dates-and-categories) mock has:

```js
gd._fullLayout.yaxis._categories
// ["3-month", "6-month", "1-year", "2-year", "3-year", "5-year", 0, 1, 2, 3, 4, 5]
```

which still gives the correct result though as the string categories are ignored during `r2p`. 

Now that, starting in this PR, `r2p` convert strings to pixels, this patch is now necessary to make the [`heatmap_xyz-dates-and-categories`](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/heatmap_xyz-dates-and-categories.png) pass.

> Do we have a test of these cases?

Good call. I added `heatmap/calc` test case to :lock: this down more explicitly in 4434fadf634c3e0a877706fea38dc9358bce9d09

src/plots/cartesian/axes.js
etpinard: done in 071249f917ca4bb5995d973bf362fd656a1e663c

src/lib/index.js
etpinard: I went with `ensureNumber` in 071249f917ca4bb5995d973bf362fd656a1e663c

test/jasmine/tests/heatmap_test.js
```diff
@@ -404,6 +409,25 @@ describe('heatmap calc', function() {
         expect(out.y).toBeCloseToArray([-0.5, 0.5]);
         expect(out.z).toBeCloseTo2DArray([[17, 18, 19]]);
     });
+
+    it('should handle the category x/y/z/ column case', function() {
+        var out = _calc({
+            x: ['a', 'a', 'a', 'b', 'b', 'b', 'c', 'c', 'c'],
+            y: ['A', 'B', 'C', 'A', 'B', 'C', 'A', 'B', 'C'],
```
alexcjohnson: can we test dates too? That's the other case where `d2c` shouldn't be called multiple times - actually, that results in a timezone shift because of our legacy support for local milliseconds!
```
ax.d2c('1970-01-02')
86400000
ax.d2c(ax.d2c('1970-01-02'))
68400000
ax.d2c(ax.d2c(ax.d2c('1970-01-02')))
50400000
```

etpinard: :arrow_right: 0fafe14

alexcjohnson:
Looks great! Thanks for the tweaks.
💃

PR #1753, opened 2017-06-02
fix bug with unhover in gl2d

alexcjohnson:
While working on axis constraints by restricting domain, I found that a bunch of tests were failing locally for me, consistently, on master as well as my branch. Most just needed increased tolerances so I'll just include them in the constraints PR. But this one, "scattergl after visibility restyle", pointed to a real bug in gl2d: if you mouseout of the gl2d plot while hovering on a point, it will not unhover.

@etpinard I guess in principle this should be an independent PR into master (in which case it would also need another commit from my branch that cleaned up the test suite so I could see where the failure was) but I need it in my `constrain-domain` branch to get tests to pass and somehow noone else encountered this error?

It would also be nice eventually if the `mouse-change` package would handle this itself somehow, rather than just calling its [`clearState`](https://github.com/mikolalysenko/mouse-change/blob/master/mouse-listen.js#L117) that's indistinguishable from a real hover event. But for now it seemed like we should handle this within the gl2d scene.

### DIFF
--- a/src/plots/gl2d/scene2d.js
+++ b/src/plots/gl2d/scene2d.js
@@ -67,6 +67,11 @@ function Scene2D(options, fullLayout) {
     // last pick result
     this.pickResult = null;
 
+    // is the mouse over the plot?
+    // it's OK if this says true when it's not, so long as
+    // when we get a mouseout we set it to false before handling
+    this.isMouseOver = true;
+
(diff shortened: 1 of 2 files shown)


etpinard:
Looks good! Thanks for the fix! :dancer: when you feel like it.

Issue #955, opened 2016-09-20
incompatible with Angular 2


rolandjitsu:
There seems to be some kind of conflict when using Angular 2 and Plotly. The same issue is somewhat described in [#7218](https://github.com/angular/angular/issues/7218).

For now, the only fix is including plotly before angular's zone.js dependency.

Could there be something done from within this lib?

etpinard:
We'll remove the `es6-promise` polyfill entirely from plotly.js in `v2.0.0` (see complete list of planned modifications [here](https://github.com/plotly/plotly.js/issues/420)).

That said,pPerhaps we could already (in the `v1.x` series) distribute a plotly.js bundle w/o the `es6-promise`? I'd vote :+1: on that if this issue gets more backers.

In the meantime, I'd recommend:

- cloning the repo
- cd into it then `npm i`, 
- removing this [line](https://github.com/plotly/plotly.js/blob/de6a183ca5d40e597bf224a4a0f7a069adf3f44a/src/core.js#L21)
- running `npm run build` and
- then use the resulting `dist/plotly.js` bundle.

I hope this helps.

rolandjitsu:
Thanks for the quick response, looks like a lot of work on the next version. It may make sense to have a patch with the promise polyfill removed already.

[Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) are widely supported (IE not) and it always makes sense to let the user decide on which polyfill to use.

I prefer using npm to mange my dependencies, so rolling a build of my own is not something I would do. For now, I just include before angular deps, which still allows me to be up to date with the newer versions.

I think that a simple doc on including the promise polyfill for those that use browsers which don't support it yet, should be enough. I'm not sure though how much work would be required in order to remove the polyfill, but if it's just a [line](https://github.com/plotly/plotly.js/blob/de6a183ca5d40e597bf224a4a0f7a069adf3f44a/src/core.js#L21), I think it's worth doing it.

Thanks again 👍

brian428:
Was this ever addressed? Trying to add plotly to an Angular 2 project and it is hanging the browser, probably because of the Promise conflict.

amellnik:
Including it from the CDN in `index.html` and using `declare var Plotly: any;` in the controller is pretty hacky, but works fine for me in a few angular-cli apps.

sn0wcat:
it seems to still be broken in angular/cli apps. The hack from amllnik works though.

[...]

Issue #768, opened 2016-07-22
Should we expose a high-level bundler module / CLI ?


etpinard:
.. so that users (not familiar with `browserify` or `webpack`) could make easily make their own custom plotly.js bundle. See discussion starting from https://github.com/plotly/plotly.js/pull/740#issuecomment-234292678

etpinard:
> Does plotly.js-bundler require plotly.js as a dependency? (which wouldn't actually need a circular require, but it's still a little weird.)

I don't think so. We could make the default path to the plotly.js index `require.resolve('plotly.js')` while letting users set their own path. 

If we go along that path, we should move the [`compress_attributes`](https://github.com/plotly/plotly.js/blob/master/tasks/util/compress_attributes.js) transform to this new repo and add a `watchify`-powered `--watch` option.

etpinard:
Closing. 

Between our partial "dist" npm packages (added in https://github.com/plotly/plotly.js/pull/2670) and our [building instructions](https://github.com/plotly/plotly.js/blob/master/BUILDING.md), I don't see this happening.

etpinard:
Also referencing https://github.com/plotly/plotly.js/pull/1584 which applies our compress attribute transform by _default_.

PR #1758, opened 2017-06-04
Remove Promise polyfill

rolandjitsu:
This addresses https://github.com/plotly/plotly.js/issues/955. It basically just removes `es6-promise` polyfill as there is really no need for it unless you use a legacy browser.

I've also added some doc on this matter for users that do still use legacy browsers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ and more.
 * [Contributing](#contributing)
 * [Community](#community)
 * [Clients for R, Python, Node, and MATLAB](#clients-for-r-python-node-and-matlab)
+* [Browser Compatibility](#browser-compatibility)
 * [Creators](#creators)
 * [Copyright and license](#copyright-and-license)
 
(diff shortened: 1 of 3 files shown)


rolandjitsu:
Note that tests were failing before I made any changes.

etpinard:
Thanks for the PR. But, we can't _just_ do that. 

Dropping support for Promise-less browsers will have to wait until [`v2`](https://github.com/plotly/plotly.js/issues/420). You'd be surprised to know how many plotly.js users still need support for IE11. So, I'll have to close this PR until we get serious about making a push for `v2`.

Like I said in https://github.com/plotly/plotly.js/issues/955#issuecomment-248437652, we could publish a  a `dist/` bundle without the Promise-polyfill (if you're up for a challenge, that's a PR we'd accept). Alternatively, we could add an option to include or not include a Promise polyfill in https://github.com/plotly/plotly.js/issues/768.

rolandjitsu:
@etpinard I understand, but it's been a while since https://github.com/plotly/plotly.js/issues/955 has been opened and I thought I should give it a try and remove the polyfill.

As I explained in the issue, users should not be forced to use a polyfill if they don't want to and it should be up to them to provide the necessary APIs that plotly would require.

Anyhow, I also understand that breaking compatibility is not an option. So making an extra bundle does not sound like a bad idea. How should I approach that, because I'm not familiar with the build system for this lib?

spontaliku-softaria:
@etpinard why not just avoid polluting of global scope with your polyfill? I understand you want promises and you need polyfill, I've just wondering why you made graphing library export anything but graphing.

etpinard:
> why not just avoid polluting of global scope with your polyfill?

That's a good idea. That might still be _considered_ a breaking change, but it's worth investigating. 

>  I've just wondering why you made graphing library export anything but graphing.

People make mistakes. Sorry.

rolandjitsu:
@etpinard I'd give it a shot at making another bundle without polyfills, but I need to understand how the bundles are created.

etpinard:
It starts here: https://github.com/plotly/plotly.js/blob/master/tasks/bundle.js

Learning a few things about the browserify [api](https://github.com/substack/node-browserify#api-example) first might help.

spontaliku-softaria:
> People make mistakes. Sorry.

Understood. I've rebuilt it as stated in #955. Sorry I can't help with PR - I'm not familiar with Plotly code.

Issue #1749, opened 2017-05-31
Implement arrayOk behavior for trace 'hoverinfo'


alexcjohnson:
The `plotly_beforehover` event can be used to prevent hover effects dynamically, by attaching a handler that returns `false` if you don't want the hover label. But it has two shortcomings:
- It doesn't include info about which point(s) were going to get a label, only the mouse event. If we move this event to after the hover data has been calculated but before the labels have been displayed, we should be able to fix this.
  - The ideal fix would be to make the event data match most of our other events, and include the plotly data and the mouse event inside one object. But if we did this now, it would be a breaking change. Do we think this is acceptable? I don't see what useful thing people would do with the mouse event by itself, so I wouldn't be *too* worried about this, but it *is* a breaking change.
  - @n-riesco suggests an alternative: we could fire this event after `gd._hoverdata` has been set, so the handler can inspect it, and then if the handler returns `false` we reset `gd._hoverdata` to what it was before (or just clear it?) Perhaps this would be a safer way to do it, then switch to ^^ in v2?
- If you attach multiple handlers, only the return value of the last one will be used. This is both limiting and potentially dangerous (if your app can possibly attach handlers in unknown order) Can we change to pass on the `&&` of *any* of these handlers returning `false`? (ie `undefined` and other falsey values should not count as `false` if we do this, only `false` itself).

@etpinard thoughts?

etpinard:
After some internal discussions, we'll implement the desired by making hoverinfo an `arrayOk` attribute.

 e.g.

```js
trace = {
  x: [1, 2, 3],
  y: [1, 2, 1],
  hoverinfo: ['x+y','skip', 'skip']
}
```

would only show x+y info on the (1,1) point, the other points would get no labels.

---

This should be part of the next minor release `1.28.0`.

PR #1761, opened 2017-06-05
Array support for trace `hoverinfo`

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1749

peek 2017-06-05 13-47

TODO:

- [x] implement for 3d, gl2d, pie - done in e8a9ddd

### DIFF
--- a/src/components/fx/calc.js
+++ b/src/components/fx/calc.js
@@ -10,31 +10,14 @@
 
 var Lib = require('../../lib');
 var Registry = require('../../registry');
-var baseAttrs = require('../../plots/attributes');
 
 module.exports = function calc(gd) {
     var calcdata = gd.calcdata;
     var fullLayout = gd._fullLayout;
 
-    function makeCoerceHoverInfo(fullTrace) {
-        var moduleAttrs = fullTrace._module.attributes;
-        var attrs = moduleAttrs.hoverinfo ?
-            {hoverinfo: moduleAttrs.hoverinfo} :
-            baseAttrs;
-        var valObj = attrs.hoverinfo;
-        var dflt;
-
-        if(fullLayout._dataLength === 1) {
-            var flags = valObj.dflt === 'all' ?
-                valObj.flags.slice() :
-                valObj.dflt.split('+');
-
-            flags.splice(flags.indexOf('name'), 1);
-            dflt = flags.join('+');
-        }
-
+    function makeCoerceHoverInfo(trace) {
         return function(val) {
-            return Lib.coerce({hoverinfo: val}, {}, attrs, 'hoverinfo', dflt);
+            return Lib.coerceHoverinfo({hoverinfo: val}, {_module: trace._module}, fullLayout);
         };
     }
 
--- a/src/components/fx/calc.js
+++ b/src/components/fx/calc.js
@@ -30,17 +30,17 @@ module.exports = function calc(gd) {
         // won't match the data array order.
         if(Registry.traceIs(trace, 'pie')) continue;
 
-        var mergeFn = Registry.traceIs(trace, '2dMap') ? paste : Lib.mergeArray;
+        var fillFn = Registry.traceIs(trace, '2dMap') ? paste : Lib.fillArray;
 
-        mergeFn(trace.hoverinfo, cd, 'hi', makeCoerceHoverInfo(trace));
+        fillFn(trace.hoverinfo, cd, 'hi', makeCoerceHoverInfo(trace));
 
         if(!trace.hoverlabel) continue;
 
-        mergeFn(trace.hoverlabel.bgcolor, cd, 'hbg');
-        mergeFn(trace.hoverlabel.bordercolor, cd, 'hbc');
-        mergeFn(trace.hoverlabel.font.size, cd, 'hts');
-        mergeFn(trace.hoverlabel.font.color, cd, 'htc');
-        mergeFn(trace.hoverlabel.font.family, cd, 'htf');
+        fillFn(trace.hoverlabel.bgcolor, cd, 'hbg');
+        fillFn(trace.hoverlabel.bordercolor, cd, 'hbc');
+        fillFn(trace.hoverlabel.font.size, cd, 'hts');
+        fillFn(trace.hoverlabel.font.color, cd, 'htc');
+        fillFn(trace.hoverlabel.font.family, cd, 'htf');
     }
 };
 
--- a/src/components/fx/calc.js
+++ b/src/components/fx/calc.js
@@ -25,11 +25,17 @@ module.exports = function calc(gd) {
         var cd = calcdata[i];
         var trace = cd[0].trace;
 
-        if(!trace.hoverlabel) continue;
+        // don't include hover calc fields for pie traces
+        // as calcdata items might be sorted by value and
+        // won't match the data array order.
+        if(Registry.traceIs(trace, 'pie')) continue;
 
         var mergeFn = Registry.traceIs(trace, '2dMap') ? paste : Lib.mergeArray;
 
         mergeFn(trace.hoverinfo, cd, 'hi', makeCoerceHoverInfo(trace));
+
+        if(!trace.hoverlabel) continue;
+
         mergeFn(trace.hoverlabel.bgcolor, cd, 'hbg');
         mergeFn(trace.hoverlabel.bordercolor, cd, 'hbc');
         mergeFn(trace.hoverlabel.font.size, cd, 'hts');
--- a/src/components/fx/calc.js
+++ b/src/components/fx/calc.js
@@ -10,9 +10,33 @@
 
 var Lib = require('../../lib');
 var Registry = require('../../registry');
+var baseAttrs = require('../../plots/attributes');
 
 module.exports = function calc(gd) {
     var calcdata = gd.calcdata;
+    var fullLayout = gd._fullLayout;
+
+    function makeCoerceHoverInfo(fullTrace) {
+        var moduleAttrs = fullTrace._module.attributes;
+        var attrs = moduleAttrs.hoverinfo ?
+            {hoverinfo: moduleAttrs.hoverinfo} :
+            baseAttrs;
+        var valObj = attrs.hoverinfo;
+        var dflt;
+
+        if(fullLayout._dataLength === 1) {
+            var flags = valObj.dflt === 'all' ?
+                valObj.flags.slice() :
+                valObj.dflt.split('+');
+
+            flags.splice(flags.indexOf('name'), 1);
+            dflt = flags.join('+');
+        }
+
+        return function(val) {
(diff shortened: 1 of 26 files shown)


src/components/fx/calc.js
```diff
@@ -10,9 +10,33 @@
...
+            var flags = valObj.dflt === 'all' ?
+                valObj.flags.slice() :
+                valObj.dflt.split('+');
+
+            flags.splice(flags.indexOf('name'), 1);
+            dflt = flags.join('+');
+        }
+
+        return function(val) {
+            return Lib.coerce({hoverinfo: val}, {}, attrs, 'hoverinfo', dflt);
```
etpinard: This here coerces `hoverinfo` items with no valid flags to `'all'` if `gd.data.length > 1` or all flags but `'name'` on single-trace graphs. For example,

```js
hoveinfo: ['sda', null, 'hello+sup']

// assuming more than one trace,
// => gets coerced to
hoverinfo: ['all', 'all', 'all']
```

---

This is debatable I feel. Perhaps, `hoverinfo` items with no valid flags should get coerced to `'none'` instead? This would be more in line with e.g. `marker.color` where

```js
marker.color: ['red', null, 'blue']

// gets coerced to
marker.color: ['red', '#444' /* i.e. Color.defaultLine */, 'blue']
```

I'm currently leaning toward the second option, but I have no strong opinion.

alexcjohnson: I think as you have it is probably correct - an invalid entry gets the default behavior, which is `'all'` or all but name. This could come in handy, in as far as the default is really what you want most of the time, for tweaking `hoverinfo` of just a few points by providing those in the array and leaving the rest of the array `undefined` or `null`.

src/traces/pie/plot.js
```diff
@@ -124,20 +124,18 @@ module.exports = function plot(gd, cdpie) {
...
                         y: hoverCenterY,
                         text: thisText.join('<br>'),
                         name: hoverinfo.indexOf('name') !== -1 ? trace2.name : undefined,
                         idealAlign: pt.pxmid[0] < 0 ? 'left' : 'right',
-                        color: pt.hbg || hoverLabelOpts.bgcolor || pt.color,
-                        borderColor: pt.hbc || hoverLabelOpts.bordercolor,
-                        fontFamily: pt.htf || hoverLabelOpts.font.family,
-                        fontSize: pt.hts || hoverLabelOpts.font.size,
-                        fontColor: pt.htc || hoverLabelOpts.font.color
+                        color: Fx.castHoverOption(trace, pt.i, 'bgcolor') || pt.color,
```
etpinard: I never realized that pie/calc.js [sorts](https://github.com/plotly/plotly.js/blob/bc0e06d280be748200870da28bf48180ef5f100b/src/traces/pie/calc.js#L83) calcdata items meaning the fullData array order isn't preserved. That is, we can't fill in pie calcdata items with hoverlabel field in order. Luckily here, we can simply reuse the gl2d/gl3d machinery.

This commit fixes pie traces with array `hoverlabel` values.

src/lib/coerce.js
```diff
@@ -338,6 +339,35 @@ exports.coerceFont = function(coerce, attr, dfltObj) {
...
 
+/** Coerce shortcut for 'hoverinfo'
+ * handling 1-vs-multi-trace dflt logic
+ *
+ * @param {object} traceIn : user trace object
+ * @param {object} traceOut : full trace object (requires _module ref)
+ * @param {object} layoutOut : full layout object (require _dataLength ref)
+ * @return {any} : the coerced value
+ */
+exports.coerceHoverinfo = function(traceIn, traceOut, layoutOut) {
```
etpinard: I hope nobody minds this abstraction. The hoverinfo `dflt` is special, so I think it deserves its own `coerce` function.

src/traces/scattercarpet/attributes.js
```diff
@@ -117,7 +117,7 @@ module.exports = {
     textfont: scatterAttrs.textfont,
     textposition: scatterAttrs.textposition,
     hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
-        flags: ['a', 'b', 'c', 'text', 'name']
+        flags: ['a', 'b', 'text', 'name']
```
etpinard: fyi @rreusser

src/traces/pie/plot.js
alexcjohnson: > This commit fixes pie traces with array hoverlabel values.

Good catch! Is this covered in the tests? If it is, it's not obvious to me which one...

etpinard: > Is this covered in the tests?

yes

image

etpinard: ... which was previously broken.

alexcjohnson: Great - I saw that change, just couldn't tell why it tested this behavior - but a failure is a failure, works for me!

test/jasmine/tests/transform_filter_test.js
```diff
@@ -262,6 +262,25 @@ describe('filter transforms calc:', function() {
...
+            transforms: [{
+                type: 'filter',
+                operation: '>',
+                value: 0
+            }]
+        })]);
+
+        expect(out[0].x).toEqual([1, 2, 3]);
+        expect(out[0].y).toEqual([2, 3, 1]);
+        expect(out[0].hoverinfo).toEqual(['none', 'skip', 'all']);
```
etpinard: Before this patch, we had

```js
out[0].hoverinfo 
// => ['x', 'y', 'text', 'name', 'none', 'skip', 'all']
```

src/plot_api/plot_schema.js
```diff
@@ -156,6 +156,7 @@ exports.findArrayAttributes = function(trace) {
         return stack.join('.');
     }
 
+    exports.crawl(baseAttributes, callback);
```
alexcjohnson: ha nice, that's a lot simpler than I was afraid it might be!

test/jasmine/tests/hover_label_test.js
```diff
@@ -1187,6 +1187,44 @@ describe('Test hover label custom styling:', function() {
                 path: ['rgb(68, 68, 68)', 'rgb(255, 255, 255)'],
                 text: [13, 'Arial', 'rgb(255, 255, 255)']
             });
+
+            // test insufficient arrayOk case
+            return Plotly.restyle(gd, 'hoverinfo', [['none']]);
+        })
+        .then(function() {
+            expect(gd.calcdata[0].map(function(o) { return o.hi; })).toEqual(
+                ['none', 'x+y+z+text', 'x+y+z+text'],
```
etpinard: Before this commit, we had

```js
gd.calcdata[0].map(o => o.hi)
// => // ['none']
```

and no fallbacks for `hoverinfo` in the hover routine. Since, the `hoverinfo` default value isn't trivial, I chose add a (another :expressionless: ) `Lib` function `fillArray`.

`Lib.fillArray` is very similar to `Lib.mergeArray`, but instead of looping over the min of `cd[i].length` and `hoverinfo.length`, it loops over `cd[i].length` allowing calcdata items to be filled in with defaults.

src/lib/index.js
```diff
@@ -350,19 +350,25 @@ lib.noneOrAll = function(containerIn, containerOut, attrList) {
     }
 };
 
-/** merge data array into calcdata items
+lib.mergeArray = function(traceAttr, cd, cdAttr) {
```
alexcjohnson: 🎎 with a description of the difference between `mergeArray` and `fillArray`? And perhaps some indication of *why* one would use merge vs fill? Makes me wonder whether any of our other existing merge usages should be fill...

`mergeArray` doesn't need `fn`?

src/lib/index.js
```diff
@@ -350,19 +350,25 @@ lib.noneOrAll = function(containerIn, containerOut, attrList) {
...
+/** fill data array into calcdata items
  *
  * @param {array} traceAttr : trace attribute
  * @param {object} cd : calcdata trace
  * @param {string} cdAttr : calcdata key
  * @param {function} [fn] : optional function to apply to each array item
  */
-lib.mergeArray = function(traceAttr, cd, cdAttr, fn) {
+lib.fillArray = function(traceAttr, cd, cdAttr, fn) {
     fn = fn || lib.identity;
```
alexcjohnson: 🐎 I wonder whether it would be worth avoiding `fn` entirely if it's not provided? Either
`cd[i][cdAttr] = fn ? fn(traceAttr[i]) : traceAttr[i];`
or even
```
if(fn) {
    for(i = 0; i < cd.length; i++) cd[i][cdAttr] = fn(traceAttr[i]);
}
else {
    for(i = 0; i < cd.length; i++) cd[i][cdAttr] = traceAttr[i];
}
```

src/lib/index.js
etpinard: > :dolls: with a description of the difference between mergeArray and fillArray?

Sure. :+1: 

> Makes me wonder whether any of our other existing merge usages should be fill...

Good point. Many routines in `Drawing` have fallbacks in case a required calcdata field is missing. The fallback pattern works well when the fallback value is constant for all cases (e.g. missing items in a `marker.color` array get drawn as `#444`), but not so well when it is depends on other things (e.g. for `hoverinfo` here which depends on `gd.data.length` and `trace.type`). 

That said, replacing all `Lib.mergeArray` calls with `Lib.fillArray` should not break anything. I could do that if you prefer. It might slow things down a little bit when e.g. 

```js
trace.x.length // => 1e5
trace.marker.color.length // => 2
```
but I'd call that an edge case.

> mergeArray doesn't need fn?

Not at the moment.

alexcjohnson: > That said, replacing all Lib.mergeArray calls with Lib.fillArray should not break anything. I could do that if you prefer. It might slow things down a little bit when e.g.

I'm not worried about perf - yes, shorter attribute array is an edge case I'm comfortable neglecting. But it does change behavior, even if it's behavior we expect users not to be depending on. I'd leave it as is for the moment, for existing uses, and make an issue about it. We'll probably have to look at each attribute individually to see if we think the change is sufficiently unambiguous as an improvement that it's worth the minor incompatibility.

src/lib/index.js
etpinard: That won't make that much of a difference (at least in Chrome 58)

image


https://gist.github.com/etpinard/4a5c7983a619efb4c98cb927529b9af6

src/lib/index.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1761/commits/3c3fda4da845df34f3b730f4f0a2b2b347fd8e6c

etpinard:
tagging this as `reviewable`

src/lib/index.js
alexcjohnson: awesome, thanks for checking.

alexcjohnson:
No more comments from me - looks great! 💃

PR #1766, opened 2017-06-08
gl2d better ticks

etpinard:
:sos: this is debatably a breaking change :sos: 

In brief, this PR removes a block (see commit https://github.com/plotly/plotly.js/commit/3d671773249ea88dfe2986861bcfa75f8dc08936) which dates back to the first gl2d implementation and recently highlighted in https://github.com/plotly/plotly.js/commit/bbac99bd58604642d85a5064643f4f7bd0873f6d (part of the gl2d annotation PR https://github.com/plotly/plotly.js/pull/1301). Back in https://github.com/plotly/plotly.js/pull/1301, I thought that this block was _necessary_ to correctly draw the gl2d axis ticks, but I was wrong. That block was plain wrong from the start. In more technical terms, `ax.setScale()` sets the correct `ax._length`, no need to override it using `glplot.viewBox`.

Removing this block makes the gl2d baselines _more on-par_ with gl2d. To demonstrate, commits https://github.com/plotly/plotly.js/commit/d63578b8f4498f5dc4953f1453c71e2a7a6f1ccc and https://github.com/plotly/plotly.js/commit/fbeb5e901811d761f99df5e1ddf2fdc2e061eba3 regenerate the `scattergl` baselines as SVG `scatter` traces, so comparing (in swipe mode especially) the baselines in commit https://github.com/plotly/plotly.js/commit/32efccd12a42613d9fcf6bac1d7d961cfebe5b82 shows little difference between the SVG and gl2d versions. But comparing the new baselines with the old in the combined diff shows some pretty significant differences.

So, what do you think? Can this be considered a bug fix? Or, should we wait for v2?

cc @alexcjohnson @dfcreative @monfera @rreusser 

TODO
- [x] fixup jasmine tests which rely on wrong axis range values


etpinard:
Full disclosure, I found out that the block removed in 3d67177 was wrong while working on scattergl lasso (PR https://github.com/plotly/plotly.js/pull/1657) where as of the latest commit 1D select box have the wrong width (because they have the wrong `ax._length`):

peek 2017-06-08 10-51

monfera:
Better unity looks great and on first sight I don't see a negative with it, though my work on `gl2d` didn't touch on the axis ticks in this regard.

alexcjohnson:
Looks great, I'm happy to call this unification a bug fix.

Issue #1577, opened 2017-04-10
Specify annotations[0].align/annotations[0].valign on [0, 1] scale


cpsievert:
Currently these attributes may be specified on a [discretized scale with 3 categories](https://github.com/plotly/plotly.js/blob/e01a765953a290967bc97b7c72d5a3d9b91f0c03/src/components/annotations/attributes.js#L81-L102).

Mainly requesting so we can achieve parity with ggplot2 when converting labels/annotations:

```r
qplot(1:10) + 
ggtitle("Hey there") + 
  theme(plot.title = element_text(size = 40, vjust = 1, hjust = 0.75, debug = T))
```

![screen shot 2017-04-10 at 2 12 15 pm](https://cloud.githubusercontent.com/assets/1365941/24878169/b95823ae-1df7-11e7-93d6-59475288f24e.png)

alexcjohnson:
If we did this, ideally we would standardize the implementation so it applies to everything that uses these discrete categories:
- `xanchor`/`yanchor` for annotations, legends, images, colorbars, rangeselectors, sliders, updatemenus
- `scatter.textposition`

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #621, opened 2016-06-09
x, y and z error bars should have different attributes


etpinard:
At the moment, all error bar containers (i.e. `error_x`, `error_y` and `error_z`) are linked to the same attribute declarations found in [`src/components/errorbars/attributes.js`](https://github.com/plotly/plotly.js/blob/9ee53c9af2c516d4c5fd4e80d1cba6e7bab95e48/src/components/errorbars/attributes.js).

But, the error bar attributes contains container-specific attributes, namely, `copy_ystyle` and `copy_zstyle` (see [here](https://github.com/plotly/plotly.js/blob/9ee53c9af2c516d4c5fd4e80d1cba6e7bab95e48/src/components/errorbars/attributes.js#L100-L107)). These attribute should only be shown in the relevant containers.

The solution would be make the `src/components/errorbars/attributes.js` export a function that return attributes specific to the container e.g.:

``` js
// in traces/scatter/attributes.js

var errorBarAttributes = require('../../components/errorbars/attributes');

module.exports = {
  // ...

  error_x: errorBarAttributes('x'),
  error_y: errorBarAttributes('y')
};
```

similar to what PR https://github.com/plotly/plotly.js/pull/609 put forward for colorscale attributes.

PR #1767, opened 2017-06-08
Constrain axes by domain

alexcjohnson:
An alternative mechanism to enforce axis scale constraints: instead of just altering `range` as in #1522 this PR allows you to keep exactly the specified ranges, and restricts `domain` instead. That way no extraneous regions of the data axes are visible.

We add two new attributes to axes to handle this:
- `constrain`: default is `constrain: 'range'` (gives the behavior in #1522), the new behavior of this PR is invoked by `constrain: 'domain'`
- `constraintoward`: which direction to allow the data (either range or domain depending on `constrain`) to float. default is `'center'` (x) or `'middle'` (y) but you can also use `'left'` and `'right'` for x axes, and `'top'` and `'bottom'` for y. For example, `'left'` means that the left end of the range or domain is unchanged when enforcing the constraint, so only the right end will change. This is most useful with `constrain: 'domain'` to control where the whitespace shows up, for example if you want the subplot to push next to a neighboring subplot or next to a legend.
- both of these only have an effect if the axis *is* constrained, ie it and another axis are linked by `scaleanchor` in one of them.
- these settings apply to each axis separately - if you want to constrain by domain in either direction, whichever happens to need constraining, you need to set `constrain: 'domain'` in both axes.

Also included here are a bunch of tweaks to unrelated test suites, so that the whole `npm run test-jasmine` passes on my machine. One of these fixes uncovered a bug - fixed in #1753 which was merged into this branch so this PR will get that fix into master.

cc @etpinard


src/plots/cartesian/dragbox.js
```diff
@@ -543,8 +548,12 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             pushActiveAxIds(yaLinked);
         }
 
+        updates = {};
         for(i = 0; i < activeAxIds.length; i++) {
-            doTicks(gd, activeAxIds[i], true);
+            var axId = activeAxIds[i];
+            doTicks(gd, axId, true);
+            var ax = getFromId(gd, axId);
+            updates[ax._name + '.range'] = ax.range.slice();
```
alexcjohnson: @etpinard this is significantly simpler than what I was doing before ([calculating the range edits again in dragTail](https://github.com/plotly/plotly.js/pull/1767/commits/b1c6b24ed457310c20f4651c621460348fb729d5#diff-1bd3dd52cddb8638876bbe17cef28a0dL644)) and more robust since it's explicitly WYSIWYG - whenever you update tick labels you record that and use it as the basis of the update object to be sent to `relayout`.

Zoombox has a counterpart [here](https://github.com/plotly/plotly.js/pull/1767/commits/b1c6b24ed457310c20f4651c621460348fb729d5#diff-1bd3dd52cddb8638876bbe17cef28a0dR827)

Oh, I'm noticing now that previously we sent each end of the range as a separate edit in the update object (ie `'xaxis.range[0]': 1, 'xaxis.range[1]': 5` rather than `'xaxis.range': [1, 5]`) - this way is more compact but I bet it'll break some applications that listen to relayout events to respond to zooms and inspect the event data to find the range changes, so I think I should change it back to the previous form.

src/plots/cartesian/dragbox.js
```diff
@@ -641,29 +650,17 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
+        // ticksAndAnnotations again - it's unnecessary and would overwrite `updates`
         updateSubplots([0, 0, pw, ph]);
-        Plotly.relayout(gd, attrs);
+        Plotly.relayout(gd, updates);
     }
 
     // updateSubplots - find all plot viewboxes that should be
     // affected by this drag, and update them. look for all plots
     // sharing an affected axis (including the one being dragged)
+    // returns all the new axis ranges as an update object
```
alexcjohnson: oops, this was an aborted attempt, I'm not using this`attrs` so will remove it.

src/plots/cartesian/dragbox.js
alexcjohnson: note that one effect of doing it this way is when you drag axis ends or make a zoombox (any of the GUI interactions) while using `constrain: 'domain'`, the shape and constrained domains of the subplot will never change because we always make proportional updates to all linked ranges. This is NOT the case for generic `relayout` calls: if you `relayout` a single range, it can reshape the subplot, potentially changing which axis is constrained, but always keeping the domains as large as possible within the constraint.

alexcjohnson:
One tricky aspect of this implementation: whenever `ax.domain` gets changed to satisfy a constraint, the input `gd.layout` axis domain is also changed to match, but the original full domain is stashed in `ax._inputDomain`, so that if you change ranges later and the subplot shape changes, it can expand to fill the originally specified space so it doesn't have to get smaller and smaller.

I considered doing this by either 1) letting the `_fullLayout` `ax.domain` be different from the `layout` input value, but I thought that would cause confusion & problems, particularly in the workspace; or 2) using another (private) attribute like `ax._finalDomain` for all actual axis positioning, but I felt like that would be too intrusive for a fairly esoteric feature.

One pitfall with this approach though is that when a plot is saved, `ax._inputDomain` will be lost so later `relayout`s will not be able to access the full original domain. This isn't a problem for pure GUI use because that *can't* change the shape (see https://github.com/plotly/plotly.js/pull/1767#discussion_r120920709) but I suppose could cause confusion if the plot is brought back into the workspace, or into some other application that permits more granular `relayout`s.

etpinard:
cc @cpsievert would might be interested in this feature.

test/jasmine/tests/gl2d_click_test.js
```diff
@@ -119,30 +114,30 @@ describe('Test hover and click interactions', function() {
...
-    function assertHoverLabelStyle(sel, expected) {
+    function assertHoverLabelStyle(sel, expected, msg) {
         if(sel.node() === null) {
             expect(expected.noHoverLabel).toBe(true);
             return;
         }
 
         var path = sel.select('path');
-        expect(path.style('fill')).toEqual(expected.bgColor, 'bgcolor');
-        expect(path.style('stroke')).toEqual(expected.borderColor, 'bordercolor');
```
etpinard: Too bad that `toEqual` can't print a failure `msg`. Oh jasmine.

etpinard: Thanks very much for those test tweaks!

test/jasmine/tests/mapbox_test.js
```diff
@@ -983,7 +983,9 @@ describe('@noCI, mapbox plots', function() {
 });
 
 describe('@noCI, mapbox toImage', function() {
-    var MINIMUM_LENGTH = 1e5;
+    // decreased from 1e5 - perhaps chrome got better at encoding these
+    // because I get 99330 and the image still looks correct
+    var MINIMUM_LENGTH = 8e4;
```
etpinard: Great. As long as the image looks ok! Some background info -> https://github.com/plotly/plotly.js/pull/1598

test/jasmine/tests/gl2d_click_test.js
alexcjohnson: yeah, didn't it used to be able to do that? super annoying.

src/plots/cartesian/dragbox.js
etpinard: > so I think I should change it back to the previous form.

I agree. Listening to `plotly_relayout` and inspecting the event data for `xaxis.range[0]` or `xaxis.range[1]` must be pretty common in userland. We should try to clean that up in v2 or maybe we could add a `plotly_rerange`?

src/plot_api/plot_api.js
```diff
@@ -383,6 +383,7 @@ Plotly.plot = function(gd, data, layout, config) {
         drawFramework,
         marginPushers,
         marginPushersAgain,
+        initInteractions,
```
etpinard: for reference (because this diff isn't great), we now have:

image

src/plot_api/plot_api.js
```diff
@@ -190,8 +190,7 @@ Plotly.plot = function(gd, data, layout, config) {
 
         return Lib.syncOrAsync([
             subroutines.layoutStyles,
-            drawAxes,
-            initInteractions
+            drawAxes
```
etpinard: Yeah, I guess initialising interactions after pushing the margins makes sense :+1:

src/plots/cartesian/axes.js
```diff
@@ -455,6 +455,13 @@ axes.expand = function(ax, data, options) {
         i, j, v, di, dmin, dmax,
         ppadiplus, ppadiminus, includeThis, vmin, vmax;
 
+    // domain-constrained axes: base extrappad on the unconstrained
+    // domain so it's consistent as the domain changes
+    if(extrappad && ax._inputDomain) {
```
etpinard: Interesting choice for having a condition based on `ax._inputDomain`. 

Wouldn't `ax.constrain === 'domain'` be more robust?

src/plots/cartesian/constraint_defaults.js
```diff
@@ -15,10 +15,25 @@ var id2name = require('./axis_ids').id2name;
...
+    if(containerOut.fixedrange) return;
 
-    var constraintOpts = getConstraintOpts(constraintGroups, containerOut._id, allAxisIds, layoutOut);
+    // coerce the constraint mechanics even if this axis has no scaleanchor
+    // because it may be the anchor of another axis.
+    coerce('constrain');
+    Lib.coerce(containerIn, containerOut, {
+        constraintoward: {
+            valType: 'enumerated',
+            values: letter === 'x' ? ['left', 'center', 'right'] : ['bottom', 'middle', 'top'],
```
etpinard: it might time to consider having separate attribute declarations for x axes and y axes (similarly for x,y and z error bars https://github.com/plotly/plotly.js/issues/621). The reference page https://plot.ly/javascript/reference/ would benefit.

src/constants/alignment.js
```diff
@@ -0,0 +1,32 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+// fraction of some size to get to a named position
+module.exports = {
+    // from bottom left: this is the origin of our paper-reference
+    // positioning system
+    FROM_BL: {
```
etpinard: First steps towards https://github.com/plotly/plotly.js/issues/1577 

Nice.

src/plots/cartesian/constraints.js
```diff
@@ -65,10 +70,115 @@ module.exports = function enforceAxisConstraints(gd) {
...
+                    if(rangeShrunk < 1) {
+                        // the range has previously been constricted by ^^, but we've
+                        // switched to the domain-constricted regime, so reset range
+                        ax.range = ax._input.range = ax._inputRange.slice();
+                        factor *= rangeShrunk;
+                    }
 
-            if(normScale !== matchScale) {
-                scaleZoom(axes[axisID], normScale / matchScale);
+                    // TODO
```
etpinard: still `TODO` ?

etpinard: ignore. my bad

src/plots/cartesian/constraint_defaults.js
etpinard: I'm curious, does `Plotly.validate` gives the correct answer here e.g. where `constraintoward: 'bottom'` in an xaxis container?

src/plots/cartesian/axes.js
alexcjohnson: ah possibly... this might get called before `enforceAxisConstraints` though, so I guess I should filter on both.

src/plots/cartesian/constraints.js
alexcjohnson: removed in a later commit I think :)

src/constants/alignment.js
alexcjohnson: Ah, wasn't even thinking of that but good point. There are tons of other places we could make use of these (potentially adding `FROM_BR` and `FROM_TR`...) to make things clearer and more efficient. see eg `ag middle:` in the src directory. I guess perhaps all of these could also allow `[0, 1]`.

src/plots/cartesian/constraints.js
alexcjohnson: yeah sorry - TODO removed [here](https://github.com/plotly/plotly.js/pull/1767/commits/59a092a1fa685fffa25a7027a70ee17040d63588#diff-3ffca45adc4809cf1b19648f91eaf1d8L110)

src/plot_api/plot_api.js
alexcjohnson: just want to call out a note I put in a [commit message](https://github.com/plotly/plotly.js/pull/1767/commits/1570274c426dfda4cf3e57c312d7fa14b531a696):
```
I don't understand why we need to initInteractions twice,
but it seems necessary to do both before and after finalDraw
```

I'll have a perf PR coming after this merges and will look into that in more detail.

src/plots/cartesian/constraint_defaults.js
alexcjohnson: heh, no, `Plotly.validate` doesn't catch `constraintoward: 'bottom'` on an x axis, but it also incorrectly complains about `anchor` and `scaleanchor`:
```
LOG: In layout, key xaxis.scaleanchor is set to an invalid value (y)
LOG: In layout, key xaxis.anchor is set to an invalid value (y)
```

src/plots/cartesian/dragbox.js
alexcjohnson: -> back to independent attributes in https://github.com/plotly/plotly.js/pull/1767/commits/51fc2c694b4e89055bfb9a02e392d1ca41e4ed03

src/plots/cartesian/dragbox.js
alexcjohnson: -> removed this `attrs` in https://github.com/plotly/plotly.js/pull/1767/commits/39319734530d192db712ac8e9f3f0e79180a43da

src/plots/cartesian/axes.js
alexcjohnson: -> tighter condition in https://github.com/plotly/plotly.js/pull/1767/commits/1271382e9ed31a238544fdc59cbe3c3f11cd9e6c

alexcjohnson: FWIW I think there's a good argument that basing this padding on `_inputDomain` is the right thing to do (rather than having the padding be 5% of the constrained domain), but a big part of MY motivation here was that trying to do it the other way means you'd have to make a totally separate attribute in the `ax._min/_max` items in addition to `val` and `pad` because you'd no longer be able to express the padding either in data (`val`) or pixel (`pad`) units, then both the auto-range and the auto-domain routines get more complicated... also it makes the conditions more complicated for which points might possibly be the extremes so must be kept in `ax._min/max`. I didn't want to open that can of worms.

But if anyone has a strong opinion that padding based on the constrained domain is truly the correct behavior we can certainly discuss. The biggest argument in that direction that I can see is that if your data make a perfect square but it's padded on constrained axes, the subplot itself will not end up precisely square. But if you've gotten to that point, perhaps you could be bothered to explicitly set up the domains and dimensions so the axes have equal length :)

src/plots/cartesian/constraint_defaults.js
etpinard: I found a way to fix this. Can I push a commit to this branch?

alexcjohnson: Sure, push it!

etpinard: Fixing `Plotly.validate` for `scaleanchor`, `anchor` and `overlaying` was a little harder than I thought actually, so I made a separate PR -> https://github.com/plotly/plotly.js/pull/1769

src/plots/cartesian/dragbox.js
```diff
@@ -733,6 +735,8 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             var plotDx = xa2._offset - clipDx / xScaleFactor2,
                 plotDy = ya2._offset - clipDy / yScaleFactor2;
 
+            // console.log(subplot.id, editX2, editY2, xScaleFactor2, yScaleFactor2, clipDx, clipDy, plotDx, plotDy);
```
etpinard: :hocho: please

test/jasmine/assets/custom_matchers.js
```diff
@@ -55,14 +55,19 @@ module.exports = {
...
-                    return isClose(element, expected[i], precision);
-                });
-
-                var passed = (
-                    expected.length === actual.length &&
-                    tested.indexOf(false) < 0
-                );
+                var passed;
+
+                if(Array.isArray(actual) && Array.isArray(expected)) {
```
etpinard: :tada:

src/plots/cartesian/dragbox.js
alexcjohnson: thanks -> https://github.com/plotly/plotly.js/pull/1767/commits/fa78376fea1cb78c1b1483a142e45541517872f4

test/jasmine/tests/axes_test.js
```diff
@@ -613,6 +613,204 @@ describe('Test axes', function() {
...
+            // input and full range always match
+            expect(ax.range.map(ax.r2l)).toBeCloseToArray(range.map(ax.r2l), 5,
+                'range, ' + msg + ': ' + ax.range);
+            expect(axIn.range.map(ax.r2l)).toBeCloseToArray(ax.range.map(ax.r2l), 5,
+                'input range, ' + msg + ': ' + ax.range);
+        }
+
+        it('can change per-axis constrain:domain/range and constraintoward', function(done) {
+            Plotly.plot(gd,
+                // start with a heatmap as it has no padding so calculations are easy
```
etpinard: Thanks for all these comments in this suite. They make my life a lot easier.

test/jasmine/tests/axes_test.js
```diff
@@ -613,6 +613,204 @@ describe('Test axes', function() {
...
+                assertRangeDomain('xaxis', xAutorange, [0, 0.5], [0, 0.5]);
+                assertRangeDomain('yaxis', yAutorange, [0, 1], [0.225, 0.775]);
+                assertRangeDomain('xaxis2', xAutorange, [0.5, 1], [0.5, 1]);
+                assertRangeDomain('yaxis2', yAutorange, [0, 1], [0.225, 0.775]);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('can constrain date axes', function(done) {
```
etpinard: Would it be too much to ask to add a `'can constrain category axes'` test case?

alexcjohnson: hah, I set myself up for that one didn't I. Sure. I'll include `'log'` too.

alexcjohnson: category and log axis domain and range constraint tests to match the category one -> https://github.com/plotly/plotly.js/pull/1767/commits/67ea3a8344e8907ebc53c035d8e1222954fd1f30

etpinard:
:dancer: :dancer: :dancer:

PR #1769, opened 2017-06-08
Validate dynamic enumerated

etpinard:
to be merged in @alexcjohnson 's https://github.com/plotly/plotly.js/pull/1767

fixes `Plotly.validate` for axis `anchor`, `overlaying` and `contraintoward`.

### DIFF
--- a/src/lib/coerce.js
+++ b/src/lib/coerce.js
@@ -44,6 +44,20 @@ exports.valObjects = {
             if(opts.coerceNumber) v = +v;
             if(opts.values.indexOf(v) === -1) propOut.set(dflt);
             else propOut.set(v);
+        },
+        validateFunction: function(v, opts) {
+            if(opts.coerceNumber) v = +v;
+
+            var values = opts.values;
+            for(var i = 0; i < values.length; i++) {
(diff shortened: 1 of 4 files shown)


src/lib/coerce.js
```diff
@@ -44,6 +44,20 @@ exports.valObjects = {
             if(opts.coerceNumber) v = +v;
             if(opts.values.indexOf(v) === -1) propOut.set(dflt);
             else propOut.set(v);
+        },
+        validateFunction: function(v, opts) {
```
etpinard: This thing in only called during `Lib.validate`. We need this so that make enumerated attributes like `xaxis.anchor` that declare [regex](https://github.com/plotly/plotly.js/blob/9779c9608ff8554a074285eac67892dc100af27b/src/plots/cartesian/layout_attributes.js#L512-L513) work with `Lib.validate`.

etpinard: oh and I should add, I didn't want to add regex support in the (regular) `coerceFunction` defined above for performance reasons.

alexcjohnson:
Looks great! 💃

Issue #679, opened 2016-06-22
Implement setCursor handler for editable annotations


etpinard:
Currently, hovering over `editable: true` annotations does not change the mouse cursor - making it hard for user to realise that editable annotations can be moved around.

See https://github.com/plotly/plotly.js/pull/660/files#diff-a123bebe752778fa4640d499ad73277cR375 for more info.

rmoestl:
Has been implemented in #2594.

PR #1772, opened 2017-06-09
Misc perf improvements

alexcjohnson:
A bunch of perf improvements. The biggest one is changing how we cache `Drawing.bBox` results https://github.com/plotly/plotly.js/commit/ea883e6912c216bc8243d0f18b67b098233df0e7 - instead of having an element keep track of its own caching, I changed it to just using the contents of the node (its `innerHTML` plus its `style` and `text-anchor` - I'm 90% sure this is sufficient to determine the bbox but needs a bit more testing). This means we're not always invalidating the cache, whether accidentally or out of an abundance of caution, and in fact even unrelated elements that have the same contents and styling can use the same bbox (such as matching labels on different axes).

Another fairly significant one is removing an unnecessary double-draw of ticks and grids during `Plotly.plot` https://github.com/plotly/plotly.js/commit/b4a2f4d581c14c92f1b9733407b1f3a9e0b27aa8 - no idea where that came from but it doesn't seem to cause any problems to pare that down.

Two smaller ones:
- don't draw placeholder titles that we're just going to delete https://github.com/plotly/plotly.js/commit/c02204f13f6508b26039ea14b54944d4352ff7c8
- remove `Lib.getPlotDiv` https://github.com/plotly/plotly.js/commit/032cf7c191cd10000ccc7a77ca3d0cb03edbbeef - this was by far the most intrusive change (I had to pass `gd` around to a bunch of places that didn't have it before), for the smallest benefit, but it turns out `instanceof HTMLElement` (which is actually in `Lib.isPlotDiv` which I *didn't* get rid of... but that one calls it once and `getPlotDiv` called it in a loop) could be kind of slow.

Also I added a perf tester `timeit` to the global scope in the test dashboard, just for convenience.

I'd like to get #1767 merged before finishing this so I can bring the test improvements in here.


src/lib/index.js
```diff
@@ -481,13 +481,6 @@ lib.containsAny = function(s, fragments) {
     return false;
 };
 
-// get the parent Plotly plot of any element. Whoo jquery-free tree climbing!
-lib.getPlotDiv = function(el) {
```
etpinard: Beautiful commit. 

Note that `Lib.isPlotDiv` is currently used in W2. No such problem for `Lib.getPlotDiv`.

src/components/titles/index.js
```diff
@@ -83,7 +89,7 @@ Titles.draw = function(gd, titleClass, options) {
     }
 
     var el = group.selectAll('text')
-        .data([0]);
+        .data(elShouldExist ? [0] : []);
```
etpinard: Nice :beer:

devtools/test_dashboard/devtools.js
```diff
@@ -8,6 +8,8 @@ var credentials = require('../../build/credentials.json');
 var Lib = require('@src/lib');
 var d3 = Plotly.d3;
 
+require('./perf');
```
etpinard: cc @rreusser would be nice to have something like that in https://github.com/rreusser/plotly-mock-viewer  

Moreover, @dfcreative's brilliant :sparkles: https://www.npmjs.com/package/fps-indicator :sparkles: would be a nice addition to our dev tools.

src/components/drawing/index.js
```diff
@@ -654,18 +653,41 @@ drawing.bBox = function(node) {
...
+// doing a simple hash ala https://stackoverflow.com/questions/7616461
+// adds ~15 microsec (nearly all of this is spent in charCodeAt)
+// function hash(s) {
+//   var h = 0;
+//   for (var i = 0; i < s.length; i++) {
+//     h = (((h << 5) - h) + s.charCodeAt(i)) | 0; // codePointAt?
+//   }
+//   return h;
+// }
+function nodeHash(node) {
```
etpinard: Interesting patch. Do you have any benchmark results comparing old and new `Drawing.bBox`?

If you're concerned about memory usage, you might want to consider setting `maxSavedBBoxes` to something less than `1e4`. 10000 text nodes on one page sounds a bit much.

cc'ing @monfera @rreusser who might have experience doing stuff like this.

monfera:
Looks great! 

- I've often seen double or worse rendering in imperative code, hard to automatically catch, I guess this double render was not intended
- bounding box calc is in good part expensive because the layoutung must ve synchronously awaited, so I'd guess avoiding bbox calc on identical tick text is even bigger than avoiding repeat calls
- caching is best if global, but risk is mem leak

src/components/drawing/index.js
alexcjohnson: > Do you have any benchmark results comparing old and new Drawing.bBox?

I don't have any explicit benchmarks, but putting in a few `console.log` statements you can see that previously during redraws (and especially interactions like dragging axis ends) we were only actually using the cache in a few cases, mostly we were redoing the `cloneNode` and `getBoundingClientRect`. Now we hardly ever need to do that. The key is that managing `data-bb` so we can safely use the cache is really hard and we weren't doing it well. This way there's nothing to manage, and `nodeHash` is very fast (as long as I don't actually hash the resulting string 🤔 ).

> 10000 text nodes on one page sounds a bit much.

Yes, perhaps I could drop the limit a bit, but two things to note:
- I'm not storing text nodes, just a lookup object of `{nodeHash: bBox}` where `bBox` is a plain object, my main concern is that `nodeHash` itself can be quite a long string, though I guess a `bBox` object of 6 numbers can also take non-negligible memory.
- A large number of text nodes is exactly when this would be most useful, and wiping the cache will kill this whole gain. Perhaps I could take advantage of `Object.keys` being ordered in insertion order (usually? always?) and drop the oldest 50% of cache entries or something if we hit the limit?

test/jasmine/tests/cartesian_interact_test.js
```diff
@@ -274,7 +274,10 @@ describe('axis zoom/pan and main plot zoom', function() {
     }
 
     function doDblClick(subplot, directions) {
-        return function() { return doubleClick(getDragger(subplot, directions)); };
+        return function() {
+            gd._mouseDownTime = 0; // ensure independence from any previous clicks
+            return doubleClick(getDragger(subplot, directions));
+        };
```
alexcjohnson: Man, this one was brutal - the key was that (locally, in chrome, when running multiple tests) it was seeing a previous click (from the same test, despite the fact that it still cared whether another test had been run before this one) too close to the first click of the doubleclick, and ended up responding as if there were *two* doubleclicks - which I still don't quite understand, I didn't think a triple-click would look like two doubles, but anyway this works and we can use this in the future (similar to your trick with `gd._lastHoverTime` @etpinard ) to cut down on unnecessary delays in the test suite.

For future reference @etpinard since we talked about potentially doing this, I went off on a dead end for a while when I noticed that calling `Plotly.purge(gd)` during `destroyGraphDiv` caused all sorts of problems... That turns out to be unrelated, that sometimes there are some things still waiting to happen on the plot after tests have finished (like `rehover`) which will fail (and confusingly attribute the failure to the next test that's run, since it's happening async) if we've purged the plot of attributes like `emit` but they still complete without failure if all we've done is remove the `gd` from the DOM.

alexcjohnson:
Some profiling on small charts: All results from Chrome on my mac with no profiles running. Running profiles adds ~5% to these times with no screenshots, ~10% with screenshots (interestingly the profiling penalty seems bigger even *fractionally* before the change). All results are averaging ~10 runs of `Plotly.newPlot` on the same data repeatedly, throwing away the first one. This gives extra weight to the caching improvements, but as my interest is mainly in interactivity performance that's what I want.

using mock `category_dtick_3` (small line chart with a linear and a category axis):
- before this branch: 144ms
- after this branch: 63ms

using a chart with ~30 variable-size bubbles, gradient fill, long category names:
- before this branch: 444ms
- after this branch: 270ms

alexcjohnson:
(offline weekend 💃  from @etpinard )

test/jasmine/tests/cartesian_interact_test.js
etpinard: Thanks for writing down the details.

src/components/shapes/draw.js
```diff
@@ -141,6 +140,8 @@ function setupDragElement(gd, shapePath, shapeOptions, index) {
 
     dragElement.init(dragOptions);
 
+    shapePath.node().onmousemove = updateDragMode;
```
etpinard: Referencing the ticket for annotation edits -> https://github.com/plotly/plotly.js/issues/679

src/components/dragelement/index.js
```diff
@@ -57,15 +62,11 @@ dragElement.init = function init(options) {
         startY,
         newMouseDownTime,
         dragCover,
-        initialTarget,
-        initialOnMouseMove;
```
etpinard: Amazing. Thanks for :hocho: that!

test/jasmine/tests/cartesian_interact_test.js
etpinard: image

:tada: nice work

PR #1776, opened 2017-06-11
[WIP] some more performance optimizations

alexcjohnson:
Following on #1772 - some more poking around for performance bottlenecks.

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -14,46 +14,7 @@
 var d3 = require('d3');
 
 var Lib = require('../lib');
-var xmlnsNamespaces = require('../constants/xmlns_namespaces');
 var stringMappings = require('../constants/string_mappings');
-
-var DOM_PARSER;
-
-exports.getDOMParser = function() {
-    if(DOM_PARSER) {
-        return DOM_PARSER;
-    } else if(window.DOMParser) {
-        DOM_PARSER = new window.DOMParser();
-        return DOM_PARSER;
-    } else {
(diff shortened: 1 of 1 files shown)


src/lib/svg_text_utils.js
```diff
@@ -106,9 +65,7 @@ exports.convertToTspans = function(_context, gd, _callback) {
                 'white-space': 'pre'
             });
 
-        result = _context.appendSVG(converted);
-
-        if(!result) _context.text(str);
+        _context.node().innerHTML = convertToSVG(str);
```
etpinard: Hmm. Does this work in IE?

etpinard: http://caniuse.com/#search=innerHTML

image

Inconclusive. We'll have to try it on Browserstack.

etpinard:
If this works out, this PR is :trophy: 

This should get us pretty close to plotly.js working in jsdom out of the box, referencing:

- https://gist.github.com/etpinard/bee7d62b43b6bb286950
- https://github.com/plotly/plotly.js/issues/1675

src/lib/svg_text_utils.js
alexcjohnson: oh crap you're right... fails even in IE11

alexcjohnson:
Alright, I suppose I could write some sort of feature detection to enable or disable this behavior... but I think the better solution (which with any luck will be even faster as well as cross-browser!) will be to just build elements like we do for everything else, rather than building a text string and turning that into elements. Going to give that a shot.

Issue #949, opened 2016-09-19
Rewrite WebGL code in regl


etpinard:
The time has come to :hocho: the [gl-vis](https://github.com/gl-vis) project. 

These multiple small modules are becoming harder and harder to maintain. A lot of helper methods are currently duplicated across several modules - which is adding a lot :hamburger: to our bundle size.

@mikolalysenko's new WebGL framework [`regl`](https://github.com/regl-project/regll) solves a lot of long-standing gl-vis issues such as multiple scopes per gl-context and improved memory management. 

I placed this issue on the v2.0.0 milestone because it _might be_ a good opportunity to make the switch in-phase with bumping [`d3` to `v4`](https://github.com/plotly/plotly.js/issues/424) but this is subject to change.

@dfcreative @monfera @bpostlethwaite

dfcreative:
More I get into gl-vis code more I realize that it is reasonable in terms of size/solutions.
Regl is awesome webgl API wrapper, but I would guess code of gl-vis components does not add much overhead compared to regl, they reuse lots of external packages. Also some packages have nothing to do with regl. For example, [text-cache](http://npmjs.org/package/text-cache) by itself adds 150kb of compressed code (~15%) to the final bundle, and regl is not going to help much with it.

monfera:
I agree text rendering with WebGL has diverse compromises depending on the method, tesselation (`vectorize-text`), texture or SDF methods. I did some benchmarks and at least for the 2D plots I believe we can just overlay a `gl` context with a plain old 2d canvas layer. It's not difficult to keep their projection in sync. 

In the 3D case, the applicability of 2D canvas for text is more of a challenge in that some of the text needs to be presented properly as part of the 3D space, i.e. possibly occluded by things in the 3D scene. Perhaps even for such cases, it's possible to do canvas text by analyzing whether a feature of interest (to which the text label belongs) is visible or not. Such analysis is possible because this is how picking works already. So even for 3D, if it's satisfactory to switch text labels on/off depending on the visibility of their anchors, text can be done with 2D canvas.

[...]

PR #1781, opened 2017-06-12
More gl2d symbols!

etpinard:
Supersedes https://github.com/plotly/plotly.js/pull/1550 with better mocks and fine tuning via https://github.com/gl-vis/gl-scatter2d-sdf/pull/4 and https://github.com/plotly/plotly.js/commit/b1bccdc8993ff73175531939e432a690f998f885

TODO:

- [x] bump `gl-scatter2d-sdf` once @dfcreative 's https://github.com/gl-vis/gl-scatter2d-sdf/pull/4 is merged/published


test/image/mocks/marker_symbols.json
```diff
@@ -0,0 +1,505 @@
+{
+  "data": [
```
etpinard: generated with https://codepen.io/etpinard/pen/LLZGZV

etpinard:
SVG:

marker_symbols


gl2d:

gl2d_marker_symbols

test/image/mocks/gl3d_marker_symbols.json
```diff
@@ -0,0 +1,77 @@
...
+      "text": [
+        "marker symbol: 'circle'",
+        "marker symbol: 'circle-open'",
+        "marker symbol: 'square'",
+        "marker symbol: 'square-open'",
+        "marker symbol: 'diamond'",
+        "marker symbol: 'diamond-open'",
+        "marker symbol: 'cross'"
+      ],
+      "marker": {
```
etpinard: We should make sure new marker symbols are positioned and scaled properly in gl3d before adding them. To be completed in a future PR.

dfcreative:
@etpinard if marker set is going to be limited, a better solution would be pre-generated sdf sprite with all markers included. That would enable all svg markers in _scattergl_ and allow to get rid of _font-atlas-sdf_, which would make code of _gl-scatter2d_ way simpler and performant.
Probably it should be addressed in #949.

alexcjohnson:
Could we make some of the markers be rotated characters? So like ➕  could rotate 45 degrees to make ✖️  rather than having to use letter X.

dfcreative:
@alexcjohnson we use [font-atlas-sdf](https://github.com/dfcreative/font-atlas-sdf) to render symbols, which in turn uses [tiny-sdf](https://github.com/mourner/tiny-sdf) internally to draw characters. It is possible, but would unnecessarily bloat API. In fact it is not the X letter used, but valid [unicode box cross symbol](https://unicode-table.com/en/2573/) ╳, which is rendered so with default ubuntu sans-serif font. We can replace it with [any other cross](https://unicode-table.com/en/search/?q=cross).

alexcjohnson:
OK - just trying to get closer to parity. Lots of good additions here, but I don't think we should include symbols that don't match fairly closely what we do in SVG. To my eye the red ones below are not close enough as shown, and the green ones could probably be used unfilled but not filled.
gl2d-markers

etpinard:
@alexcjohnson 

> To my eye the red ones below are not close enough as shown, and the green ones could probably be used unfilled but not filled.

Done in :arrow_right: https://github.com/plotly/plotly.js/pull/1781/commits/bb9a46c808352503bca429abd4f74ef08c9fb455

Thanks for taking a look :microscope:

dfcreative:
@alexcjohnson thanks for the good point! Fixed in 1f610febece7ec8b7bbee2cf50d7e08016248916

etpinard:
> it's a bit of a shame not to have any "X" symbols...

Yeah, it's a shame. Actually, I should put [it](https://github.com/plotly/plotly.js/blob/85e8770998a0e0ca4ee3a3de3ede6f494a3ec91f/src/constants/gl_markers.js#L20) back in even it doesn't render in `imagetest`.

src/traces/scattergl/convert.js
```diff
@@ -491,29 +500,56 @@ proto.updateFancy = function(options) {
...
+            } else {
+                _borderColors = borderColors;
+            }
+
+            if(symbolSpec.bwFactor) {
+                bwFactor = symbolSpec.bwFactor;
+            } else if(symbolSpec.noBorder) {
+                bwFactor = 0.25;
+            } else if(symbolSpec.noFill) {
+                bwFactor = 0.1;
```
alexcjohnson: so... this means that for these symbols with a smaller `bwFactor` the border will grow more slowly than the others? The result being that for small `borderWidth` these ones be thicker than other symbols, and for large `borderWidth` they will be thinner, and somewhere in the middle they will look similar right? I would think we'd do better with something like a minimum border width, as a fraction of `size`, that corresponds to the line width already baked into the unicode symbol, so the symbol gets *no* explicit border until you provide a width greater than this, and at greater widths you set a border `borderWidth - minBorderWidth`.

alexcjohnson:
@etpinard nice work! just the one more question about how we calculate border widths https://github.com/plotly/plotly.js/pull/1781/files#r121820798 and we should be ready to go. The available symbols are looking really good now. Thanks for bearing with all my nitpicks, I just want to get this all as final as possible so we don't need to come back to it again...

src/traces/scattergl/convert.js
etpinard: Thanks for the headsup. For documentation purposes, using https://codepen.io/etpinard/pen/JJRMxo?editors=0010 as of commit fffc702


SVG gives:

image

and gl:

image

etpinard: @alexcjohnson here's my latest attempt:

https://github.com/plotly/plotly.js/compare/gl2d-more-symbols...gl2d-more-symbols-line-width

alexcjohnson:
```
//this.scatter.options.borderWidths[i] = (bw > minBorderWidth) ? bw - minBorderWidth : 0;

// but I found better results with:
this.scatter.options.borderWidths[i] = (bw > minBorderWidth) ? bw - minBorderWidth : bw;
```

`minBorderWidth` should only be nonzero for `noBorder` or `noFill` symbols right? Does that make it work better to use `bw - minBorderWidth : 0`? As it is it looks funny, as border width is non-monotonic.

etpinard: Yeah, I think that's better:

https://github.com/plotly/plotly.js/commit/3f664a1744467f191a230384dbf74549ab1452a2

alexcjohnson: Awesome - I think we've got it 🎉  💃

alexcjohnson:
💃 💃 💃 !

Issue #382, opened 2016-04-01
Word wrap for long labels


eugenesvk:
Is there an option to reflow long text labels?
- For example, a long label without reflow (plotly tries to autocorrect by slanting the labels instead of reflowing the text, but it still doesn't fit)
![plotlylonglabels1](https://cloud.githubusercontent.com/assets/12956286/14210826/74d2cb16-f833-11e5-8def-3d12e4c01010.PNG)

- And a simple manual fix with a few `<br>` in the code that I'd like to have Plotly be able to do with a simple `wrap: true` option
![plotlylonglabels2](https://cloud.githubusercontent.com/assets/12956286/14210809/595b5754-f833-11e5-8f6a-0c80e7619a1e.PNG)

JS code for this chart
```
var data = [{
// first screenshot labels
	//x: ['giraffes', 'orangutans', "Very Long Text Label That Doesn't Fit Properly"],
// second screenshot labels
	x: ['giraffes', 'orangutans', "Very Long<br>Text Label That<br>Doesn't Fit<br>Properly"],
	y: [20, 14, 23],
	type: 'bar'
}]
Plotly.newPlot('PlotlyDivSimple', data, {})
```

mdtusz:
Unfortunately, there isn't this option built into plotly.js at the moment. The text labels are created using SVG elements, or in WebGL where text-reflow is done manually and it's not *quite* as easy just constraining the width. 

This likely won't be of high priority for us in the *immediate* future, but we're always open to pull requests and are happy to help on them!

eugenesvk:
Ok, understood. Would help with a PR if I had a slightest clue how to do that :)

Pardon my ignorance, but is there no way to introduce a JavaScript function that would be applied to all labels before they're fed to SVG/WebGL, where it becomes complicated?
For example, I've googled found this function
```
function stringDivider(str, width, spaceReplacer) {
    if (str.length>width) {
        var p=width
        for (;p>0 && str[p]!=' ';p--) {
        }
        if (p>0) {
            var left = str.substring(0, p);
            var right = str.substring(p+1);
            return left + spaceReplacer + stringDivider(right, width, spaceReplacer);
        }
    }
    return str;
}
```

that I can use internally to shorten the labels
```
var label = "Very Long Text Label That Doesn't Fit Properly"
x: ['giraffes', 'orangutans', stringDivider( label, 20, "<br>") ],
```
[...]

PR #1783, opened 2017-06-13
Build svg text

alexcjohnson:
For our rich text displays: instead of converting pseudo-HTML into svg XML and then parsing this into elements, this PR converts the pseudo-HTML directly into `<tspan>` (and `<a>`) elements. Interestingly the speed gain isn't as pronounced as it was in #1776 using `innerHTML` but it helps a bit, and also addresses a couple of deficiencies in the previous version.

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -23,12 +23,14 @@ function getSize(_selection, _dimension) {
     return _selection.node().getBoundingClientRect()[_dimension];
 }
 
+var FIND_TEX = /([^$]*)([$]+[^$]*[$]+)([^$]*)/;
+
 exports.convertToTspans = function(_context, gd, _callback) {
     var str = _context.text();
 
     // Until we get tex integrated more fully (so it can be used along with non-tex)
     // allow some elements to prohibit it by attaching 'data-notex' to the original
-    var tex = (!_context.attr('data-notex')) && str.match(/([^$]*)([$]+[^$]*[$]+)([^$]*)/);
+    var tex = (!_context.attr('data-notex')) && str.match(FIND_TEX);
     var parent = d3.select(_context.node().parentNode);
     if(parent.empty()) return;
     var svgClass = (_context.attr('class')) ? _context.attr('class').split(' ')[0] : 'text';
@@ -138,9 +140,12 @@ exports.convertToTspans = function(_context, gd, _callback) {
 
 // MathJax
 
+var LT_MATCH = /(<|&lt;|&#60;)/g;
+var GT_MATCH = /(>|&gt;|&#62;)/g;
+
 function cleanEscapesForTex(s) {
-    return s.replace(/(<|&lt;|&#60;)/g, '\\lt ')
-        .replace(/(>|&gt;|&#62;)/g, '\\gt ');
+    return s.replace(LT_MATCH, '\\lt ')
+        .replace(GT_MATCH, '\\gt ');
 }
 
 function texToSVG(_texString, _config, _callback) {
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -391,7 +391,7 @@ function buildSVGText(containerNode, str) {
         currentNode = nodeStack[nodeStack.length - 1].node;
     }
 
-    var hasLines = str.match(BR_TAG);
+    var hasLines = BR_TAG.test(str);
 
     if(hasLines) newLine();
     else {
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -17,62 +17,6 @@ var Lib = require('../lib');
 var xmlnsNamespaces = require('../constants/xmlns_namespaces');
 var stringMappings = require('../constants/string_mappings');
 
-var DOM_PARSER;
-
-exports.getDOMParser = function() {
-    if(DOM_PARSER) {
-        return DOM_PARSER;
-    } else if(window.DOMParser) {
-        DOM_PARSER = new window.DOMParser();
-        return DOM_PARSER;
-    } else {
-        throw new Error('Cannot initialize DOMParser');
-    }
-};
-
-// Append SVG
-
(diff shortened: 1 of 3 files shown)


src/lib/svg_text_utils.js
```diff
@@ -17,62 +17,6 @@ var Lib = require('../lib');
...
-exports.getDOMParser = function() {
-    if(DOM_PARSER) {
-        return DOM_PARSER;
-    } else if(window.DOMParser) {
-        DOM_PARSER = new window.DOMParser();
-        return DOM_PARSER;
-    } else {
-        throw new Error('Cannot initialize DOMParser');
-    }
-};
```
alexcjohnson: 🔪 `DOMParser` - but now replaced with something that's tested to work in IE 🎉

src/lib/svg_text_utils.js
```diff
@@ -17,62 +17,6 @@ var Lib = require('../lib');
...
-        if(d.indexOf('<') !== -1 || d.indexOf('>') !== -1) { return ''; }
-        return hiddenDiv.html(d).text(); // everything else, let the browser decode it to unicode
-    });
-    hiddenDiv.remove();
-    return replaced;
-};
-
-exports.xml_entity_encode = function(str) {
-    return str.replace(/&(?!\w+;|\#[0-9]+;| \#x[0-9A-F]+;)/g, '&amp;');
-};
```
alexcjohnson: moved these encoders to `snaphost/tosvg`, which is now the only place they're used because `svg_text_utils` is creating text nodes directly, so doesn't need to do any conversion/sanitizing.

test/jasmine/tests/svg_text_utils_test.js
```diff
@@ -301,7 +301,8 @@ describe('svg+text utils', function() {
             var node = mockTextSVGElement('SO<sub>4<br>44</sub>');
             expect(node.html()).toBe(
                 '<tspan class="line" dy="0em">SO​' +
-                    '<tspan style="font-size:70%" dy="0.3em">4</tspan></tspan>' +
+                    '<tspan style="font-size:70%" dy="0.3em">4</tspan>' +
+                    '<tspan dy="-0.21em">​</tspan></tspan>' +
```
alexcjohnson: I don't think this has any visual consequences, but `<sup>`/`<sub>` tags previously didn't get their proper reset elements when they spanned line breaks.

test/jasmine/tests/svg_text_utils_test.js
```diff
@@ -74,7 +74,7 @@ describe('svg+text utils', function() {
         }
 
         afterEach(function() {
-            d3.select('#text').remove();
+            d3.selectAll('.text-tester').remove();
```
alexcjohnson: the cases I added (previously and in the next commit) that test several input strings with identical output were not properly cleaned up with just `select`.

test/jasmine/tests/svg_text_utils_test.js
```diff
@@ -300,12 +300,36 @@ describe('svg+text utils', function() {
...
                     '<tspan style="font-size:70%" dy="0.3em">44</tspan>' +
-                    '<tspan dy="-0.21em">​</tspan></tspan>');
+                    '<tspan dy="-0.21em">\u200b</tspan></tspan>');
+        });
+
+        it('allows nested tags to break at <br>, eventually closed or not', function() {
+            var textCases = [
+                '<b><i><sup>many<br>lines<br>modified',
+                '<b><i><sup>many<br>lines<br>modified</sup></i></b>',
+                '<b><i><sup>many</sup><br><sup>lines</sup></i><br><i><sup>modified',
```
alexcjohnson: This behavior *is* supported by regular HTML - and can be really useful to decouple styling from line breaks.

We tried weakly to support this previously but only allowed closing and reopening a single tag. The new version made it easy to handle arbitrary nesting.

test/jasmine/tests/svg_text_utils_test.js
etpinard: Nice! I've seen those forgotten `<text>` node before while `npm run test-jasmine`. Thanks :tada:

src/lib/svg_text_utils.js
```diff
@@ -264,19 +210,14 @@ var ENTITY_TO_UNICODE = Object.keys(stringMappings.entityToUnicode).map(function
     };
 });
 
-var UNICODE_TO_ENTITY = Object.keys(stringMappings.unicodeToEntity).map(function(k) {
```
etpinard: So, this part was useless?

etpinard: Oh I see, right, this is useless now as we're building the text nodes directly instead of using `window.DomParser`. Nice!

src/lib/svg_text_utils.js
```diff
@@ -346,145 +292,166 @@ function convertToSVG(_str) {
...
+    function exitNode(type) {
+        var innerNode = nodeStack.pop();
+        if(type !== innerNode.type) {
+            Lib.log('Start tag <' + innerNode.type + '> doesnt match end tag <' +
+                type + '>. Pretending it did match.', str);
+        }
+        currentNode = nodeStack[nodeStack.length - 1].node;
+    }
+
+    var hasLines = str.match(BR_TAG);
```
etpinard:
```js
var hasLines = BR_TAG.test(str);
```

would be best here, as `.test` always returns a boolean. Moreover, I remember benchmarking `regex.test` vs `str.match`, and `.test` a while ago (RIP jsFiddle) came out on top.

etpinard:
Looks awesome. Building `tspan` nodes on our own should help us with text wrapping (one of plotly's long-standing issue: https://github.com/plotly/plotly.js/issues/382). 

:dancer:

src/lib/svg_text_utils.js
alexcjohnson: nice catch! done in https://github.com/plotly/plotly.js/pull/1783/commits/3113795db0979c74b7391456a15e30852bcfee15

etpinard: Thanks! :dancer:

PR #1786, opened 2017-06-14
hide off-plot annotations

### LINKED ISSUE #1771: Annotations appear in the corner if they are zoomed off-plot
See eg the annotations-autorange mock - it's easiest to see with annotations that don't have arrows, but happens with all of them: zoom in on one annotation, eg "right no arrow" - the other three annotations from that plot generate junk in the upper left corner.
screen shot 2017-06-09 at 1 43 43 am

alexcjohnson:
fixes #1771 

@etpinard the bug was actually introduced in https://github.com/plotly/plotly.js/commit/aa58118e3665dd60d3fdda689ae58cf1814f99c2 - nice easy fix, plus a test 🔒

### DIFF
--- a/src/components/annotations/draw.js
+++ b/src/components/annotations/draw.js
@@ -316,7 +316,7 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
                         annotationIsOffscreen = true;
                     }
 
-                    if(annotationIsOffscreen) return;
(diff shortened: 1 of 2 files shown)


etpinard:
Thanks very much :dancer: 

For :books: commit aa58118 was first merged in https://github.com/plotly/plotly.js/pull/1638 (gl3d annotations)

Issue #1782, opened 2017-06-13
annotations in IE: bounding boxes and arrowheads broken


alexcjohnson:
Check out `annotations_autorange`, which looks like this on Chrome (and matching in Edge and FF):

![screen shot 2017-06-13 at 11 14 16 am](https://user-images.githubusercontent.com/2678795/27089840-8b6f5570-5029-11e7-96de-c834ea660e94.png)

On IE11 it looks like this:

![screen shot 2017-06-13 at 11 15 25 am](https://user-images.githubusercontent.com/2678795/27089904-b401cd1a-5029-11e7-8b0a-1ba26605baaa.png)

Notice that when width or height are specified explicitly, sizing and clipping works correctly - but the normal autosize does not work.

alexcjohnson:
I was concerned the bbox part might have to do with premature caching - from #1772 - but no, it isn't fixed if you `relayout` one of the annotations to totally different text.

alexcjohnson:
Oh crap, the bbox part *is* from #1772 though - because I used `innerHTML` to generate the node hash. OK, I'm on it.

PR #1788, opened 2017-06-14
fix and test annotations with mathjax

### LINKED ISSUE #1787: annotations with mathjax are misaligned
for example, in the annotations mock, do `Plotly.relayout(gd,{'annotations[14].text':'$x+y+z+q$'})`
on 1.27.1 we get the expected:
screen shot 2017-06-14 at 11 41 53 am
But current master is broken:
screen shot 2017-06-14 at 11 47 31 am

alexcjohnson:
fixes #1787 - the bug came in when we simplified the class structure of annotation elements.
@etpinard another quick fix - lets hope the test image works consistently on CI. If it does, I'll probably add some more mathjax image tests (for elements other than annotations) when I tweak `drawing.bBox` for #1782


etpinard:
image

Hmm. Looks like CI doesn't like MathJax intermittently.

alexcjohnson:
> Hmm. Looks like CI doesn't like MathJax intermittently.

Annoying. I guess this is probably the same sporadic delay we saw in https://github.com/plotly/streambed/issues/9508

I guess what I'll do is take MathJax out of the annotations mock, and make one MathJax mock that we skip in CI for now but tries to cover *everywhere* we support MathJax at present.

etpinard:
> I guess what I'll do is take MathJax out of the annotations mock, and make one MathJax mock that we skip in CI for now but tries to cover everywhere we support MathJax at present.

Very good idea, until we have more time to investigate plotly/streambed#9508

To do so, just add one expression to this [filter](https://github.com/plotly/plotly.js/blob/master/test/image/compare_pixels_test.js#L104) function.

alexcjohnson:
Actually I made a mistake in 033bee3 - fixed in b0d1af2 - that means it's likely MathJax will work fine on CI after all. But I found a bunch of other issues when developing a more complete MathJax test mock so I'm going to defer that to a separate PR if you don't mind.

etpinard:
> so I'm going to defer that to a separate PR if you don't mind.

Sounds good.

etpinard:
:dancer: :dancer: :dancer:

Issue #1768, opened 2017-06-08
Annotation autorange on category axes


alexcjohnson:
Annotation autorange is broken if you position the annotation using the category names for positioning, rather than the category serial number.
```
Plotly.newPlot(gd, [{x: ['a', 'b'], y: ['c', 'd']}], {
    annotations:[{x: 1, y: 1, ax: 30, ay: -30, text: 'boo!'}],
    width: 300,
    height: 300,
    paper_bgcolor:'#ffc'
})
```
works:
![screen shot 2017-06-08 at 12 50 03 pm](https://user-images.githubusercontent.com/2678795/26940554-1a6849c4-4c49-11e7-886d-6ecdedfc1d3e.png)

but if I change to category names:
```
Plotly.newPlot(gd, [{x: ['a', 'b'], y: ['c', 'd']}], {
    annotations:[{x: 'b', y: 'd', ax: 30, ay: -30, text: 'boo!'}],
    width: 300,
    height: 300,
    paper_bgcolor:'#ffc'
})
```
it fails to include the annotation in autorange:
![screen shot 2017-06-08 at 12 51 31 pm](https://user-images.githubusercontent.com/2678795/26940600-401d3e40-4c49-11e7-9ac5-60b10fe95bb4.png)

etpinard:
This appear to be working fine in `1.27.1`: https://codepen.io/etpinard/pen/ZypqJz?editors=1010

etpinard:
PR https://github.com/plotly/plotly.js/pull/1748 strikes again. Not sure why `annotations.json` didn't produce a diff.

etpinard:
Blocking for `1.28.0`

alexcjohnson:
The mock to include this in is `annotations-autorange.json` - make one of the axes categorical and use the category names for a couple of the positions on that axis.

PR #1791, opened 2017-06-15
Fix annotation autorange on category axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1768 by making the category axes `setConvert` methods a bit more of a mess. Note that annotation autorange uses [`ax.r2c`](https://github.com/plotly/plotly.js/blob/ffb3791989ff97de528fc66d0f9748fa2aedec43/src/components/annotations/calc_autorange.js#L61).

Oh well, with this patch here, it seems like all the cases are covered.

Together with https://github.com/plotly/plotly.js/pull/1748, we now have working category annotations on first draw and on pan/zoom :tada: :

peek 2017-06-15 15-21

By the way @alexcjohnson I chose to add a new mock in https://github.com/plotly/plotly.js/commit/f5c038295616012143626272ee2bb7afdc520540 instead of appending `annotation-autorange.json` to test shapes on category axes as well. I hope you don't mind.


alexcjohnson:
Nice and simple, and this actually seems like it's the logical way for `ax.r2c` to work. No problem about splitting off the mock - you're right to include both x and y, shapes and annotations (what about images?) - though at some point I feel like a bit of consolidation may be in order.

But did that mock fail before the patch? The coordinates that lead to the shape and annotations bumping out the autorange didn't come from category names. For annotations we can use a category that's already in the data and let the text box or arrow add extra pixel padding, for shapes it seems like we may need to use a category that's not in the data, so `ax.categoryarray`?

etpinard:
@alexcjohnson 

> Nice and simple, and this actually seems like it's the logical way for ax.r2c to work.

Great.

> (what about images?)

Layout-image-induced autorange bumps aren't a thing ye (seet https://github.com/plotly/plotly.js/issues/1111) and the [`layout_image`](http://rickyreusser.com/plotly-mock-viewer/#layout_image) mock already has a category axis. But yeah, when #1111 will be completed, we should add an image or two to `category-autorange.json`.

> But did that mock fail before the patch?

Oops, I messed that up when I added arrows to the mock. Fixed in -> e4aa7a3deab5b26df802506dd1ea25610002a9d7

> for shapes it seems like we may need to use a category that's not in the data, so ax.categoryarray?

Good call here. That actually uncovered another bug. Done in -> aaf12ba9cbbb465a21d2680efafa43cf8822c3fb

Bonus: annotations and shapes with invalid category coordinates were broken too. Fixed in -> 0758b3b48b8591001c002ede5d1d5c4d8dcc65a5

alexcjohnson:
Beautiful. Gotta love it when all these extra edge cases pop out on fixing related issues.
💃

Issue #1773, opened 2017-06-09
Remove Math.sign


alexcjohnson:
Math.sign is not supported in IE, even IE11. Currently we only seem to have this in gl2d/camera
https://github.com/plotly/plotly.js/blob/master/src/plots/gl2d/camera.js#L134
https://github.com/plotly/plotly.js/blob/master/src/plots/gl2d/camera.js#L152

We should remove these and include `Math.sign` (any other IE quirks?) in our syntax tests.

alexcjohnson:
fixed by #1792

PR #990, opened 2016-09-29
Try out visibility:inherit default for text

rreusser:
See #984 

The goal of this PR is to get circle-ci to see what happens if the visibility default on text is `visibility: inherit` instead of `visibility: visible`. `visible` hard-codes it and is difficult to override. I had an ominous feeling about this change leading to a rabbit hole of related fixes and compensations, but let's see what happens…

Issue #984, opened 2016-09-28
css visibility is overridden by plotly


fresheneesz:
If you use css `visibility:hidden` on an element containing a plotly graph, you can still see the axes numbers and titles.

rreusser:
The reason for this is that [visibility of text nodes is explicitly set to `visibility: 'visible'`](https://github.com/plotly/plotly.js/blob/master/src/lib/svg_text_utils.js#L94).

@etpinard, do you know for sure whether that line is strictly necessary or not? Could it be set only when necessary? Could `visibility: 'inherit'` lead to easier overriding? Plotly seems to work fine if I simply remove that line, though I'm sure there are *many* more uses than I could possibly have tested at a quick glance.

As a workaround, `opacity: 0` hides the plot, though I'm not sure offhand how to set `pointer-events: 'none'` without just placing an empty div over the plot that intercepts pointer events.

etpinard:
Controlling the elements inside a plotly.js graph div with CSS is a bad idea to begin with.

plotly.js graph are meant to be fully and strictly determined by the `data` / `layout` JSON serializable API.

We _could_ set `visibility: 'inherit'` instead of `visibility: 'visible'` in this case, but I'm afraid that won't solve all your issues when pointing CSS to plotly.js elements. 

So, I'd vote for `type: wontfix`. Any objections?

rreusser:
I sympathize that it's nice if properties aren't hard-coded unnecessarily, but I also agree that it's not a good idea in general to expect CSS properties to filter into the SVG. I'd vote for `wontfix`, particularly since there's not a more general path forward getting SVG to play well with CSS and since a combination of `opacity` and, if necessary, transparent divs that mask pointer events (which even more so can't be made to play well with CSS) seem to be a reasonably straightforward workaround.

fresheneesz:
@etpinard I don't think this is an acceptable stance for a library built to be used in a browser environment. You simply aren't going to have graphs outside the context of the dom, and therefore there are a minimum of things that plotly should do to play nice with the dom, including sizing correctly within its parent element, and reacting to things like visibility and opacity of ancestor elements.

Please reopen so we can continue to discuss this for more than 23 hours, this topic deserves more than 3 opinions.

rreusser:
[...]

PR #1792, opened 2017-06-16
No innerHTML and pseudo-html cleanup

alexcjohnson:
Started out as a bugfix but kind of turned into a spring cleaning...

In an effort to fix bounding box calculations in IE (one part of #1782) - which were broken by trying to use the `innerHTML` property of an SVG element - I wound up going through every user of `Drawing.bBox` and trying to make sure it was a `<text>` or `.*-math-group` node that had previously had `svgTextUtils.convertToTspans` called on it. Because if so, I could use the `data-unformatted` attribute, along with whether or not the node was converted to MathJax, in place of the `innerHTML` in the cache key, which is also much smaller than the `innerHTML`, reducing my memory concerns about this cache.

But then while doing THAT it became clear that all the monkey business using `tspan.line` elements to count lines and to properly align multi-line text was a disaster, and not done properly in a lot of places, so I centralized all that in two new utilities:
- `svgTextUtils.positionText` - to be used instead of `Drawing.setPosition`, handles multiline alignment, and also gets called within `convertToTspans` so if you use `transform` instead of the `x` and `y` attributes to position the text, you never even need to use this at all.
- `svgTextUtils.lineCount` - nuff said. No more hacky and/or broken selectors for counting lines of text.

So after that, *almost* nobody outside of `svgTextUtils` even needs to know about `<tspan>` or the `.line` class.

Along the way I found a few places we were not supporting or had bugs supporting pseudo-HTML but easily could, and added it: rangeselectors, updatemenus, sliders, carpet axes (titles and tick labels). And in cases that currently break with MathJax I explicitly cancel it (with the `data-notex` attribute). (MathJax test mocks coming in a future PR)

And while I was working in `test_syntax.js` I took care of #1773 as well.

There are still a few `Drawing.bBox` users that call it on elements larger than a `<text>` or MathJax group node, that at this point in this branch we will NOT cache at all. For 🐎  I need to either convert them to their base nodes or find an alternate way to give them a robust cache key:
- [ ] colorbar/draw:473 & 490
- [ ] titles/index:146 & 171


src/components/rangeselector/draw.js
```diff
@@ -180,23 +180,23 @@ function reposition(gd, buttons, opts, axName) {
...
-            text = button.select('.selector-text');
+        var button = d3.select(this);
+        var rect = button.select('.selector-rect');
+        var text = button.select('.selector-text');
 
-        var tWidth = text.node() && Drawing.bBox(text.node()).width,
-            tHeight = opts.font.size * 1.3,
-            tLines = svgTextUtils.lineCount(text);
+        var tWidth = text.node() && Drawing.bBox(text.node()).width;
+        var tHeight = opts.font.size * 1.3;
```
etpinard: or maybe inside a new `svg_text_utils.js` method?

src/snapshot/tosvg.js
```diff
@@ -90,16 +90,18 @@ module.exports = function toSVG(gd, format) {
...
             var txt = d3.select(this);
 
-            // hidden text is pre-formatting mathjax,
-            // the browser ignores it but it can still confuse batik
-            if(txt.style('visibility') === 'hidden') {
+            // hidden text is pre-formatting mathjax, the browser ignores it
+            // but in a static plot it's useless and it can confuse batik
+            // we've tried to standardize on display:none but make sure we still
+            // catch visibility:hidden if it ever arises
+            if(txt.style('visibility') === 'hidden' || txt.style('display') === 'none') {
```
etpinard: Referencing https://github.com/plotly/plotly.js/pull/990

src/traces/carpet/plot.js
```diff
@@ -214,6 +221,7 @@ function drawAxisTitle(layer, trace, t, xy, dxy, axis, xa, ya, offset, labelClas
         var el = d3.select(this);
 
         el.text(axis.title || '')
+            .call(svgTextUtils.convertToTspans, gd)
```
etpinard: Amazing, Pinging @rreusser

tasks/test_syntax.js
```diff
@@ -70,9 +70,16 @@ function assertSrcContents() {
                 // look for .classList
                 if(node.type === 'MemberExpression') {
                     var parts = node.source().split('.');
-                    if(parts[parts.length - 1] === 'classList') {
+                    var lastPart = parts[parts.length - 1];
+                    if(lastPart === 'classList') {
                         logs.push(file + ' : contains .classList (IE failure)');
                     }
+                    else if(lastPart === 'innerHTML') {
+                        logs.push(file + ' : contains .innerHTML (IE failure in SVG)');
```
etpinard: There might be exceptions to this rule in the future, if we end up using `innerHTML` on non-SVG DOM elements. Oh well, we could always filter the cases out then. We might we want to add a comment here signalling this check is too strict.

test/jasmine/tests/snapshot_test.js
```diff
@@ -224,12 +224,14 @@ describe('Plotly.Snapshot', function() {
...
 
         it('should force *visibility: visible* for text elements with *visibility: inherit*', function(done) {
+            // we've gotten rid of visibility almost entirely, using display instead
             d3.select(gd).style('visibility', 'inherit');
 
             Plotly.plot(gd, subplotMock.data, subplotMock.layout).then(function() {
 
                 d3.select(gd).selectAll('text').each(function() {
                     expect(d3.select(this).style('visibility')).toEqual('visible');
+                    expect(d3.select(this).style('display')).toEqual('block');
```
etpinard: Nice touch. Thanks!

src/lib/svg_text_utils.js
```diff
@@ -40,18 +40,23 @@ exports.convertToTspans = function(_context, gd, _callback) {
...
+        });
 
     function showText() {
         if(!parent.empty()) {
             svgClass = _context.attr('class') + '-math';
             parent.select('svg.' + svgClass).remove();
         }
         _context.text('')
-            .style({
-                visibility: 'inherit',
```
alexcjohnson: Ah, I didn't notice the original issue #984 that led to this change... assumed it was a remnant of a much older time. My goal was to *remove* inline styles whenever possible, rather than setting an explicit reset value, but anyway I [moved](https://github.com/plotly/plotly.js/pull/1792/commits/61dd3b181a9611a24651ce6c88f8f075438e899e) from `visibility:hidden` to `display:none` because only the latter takes the item out of the bounding box, and there are still a couple of places (most notably titles/index:146 & 171 that I'm still working on) we get the bounding box of the combined text+MathJax element. I'll see if I can come up with a test of #984 to include with this PR.

etpinard:
:dancer: after that linting mistake is fixed.

> For :racehorse: I need to either convert them to their base nodes or find an alternate way to give them a robust cache key:

I suppose this will wait for a future PR?

alexcjohnson:
> > For :racehorse: I need to either convert them to their base nodes or find an alternate way to give them a robust cache key:
> 
> I suppose this will wait for a future PR?

I'm going to take a stab at making the ones in `titles` cacheable for this PR, as I think that's actually a big part of the speedup in #1772 that we're losing at the moment. The ones in `colorbar` though I'll probably leave for later.

src/components/rangeselector/draw.js
alexcjohnson: Good call. It's used in a few different ways, so I went for `alignment.js` -> https://github.com/plotly/plotly.js/pull/1792/commits/c5fa44b21dac1c3b4039aa4ca2b998e0d2410b30

tasks/test_syntax.js
alexcjohnson: good thought. Comment added -> https://github.com/plotly/plotly.js/pull/1792/commits/a68f861a8fb57184bd887ce18ec9f72fe34646ce

src/components/drawing/index.js
```diff
@@ -633,23 +633,10 @@ drawing.bBox = function(node) {
...
-
-    if(testNode.nodeName === 'text') {
-        test3.selectAll('tspan.line').attr({
-            x: 0,
-            y: 0
-        });
-    }
+    // standardize its position (and newline tspans if any)
+    d3.select(testNode)
+        .attr('transform', '')
```
alexcjohnson: yes, that's better -> https://github.com/plotly/plotly.js/pull/1792/commits/82ba7eebb16f21933dbae90057984fc5b0a82f09

src/components/legend/draw.js
```diff
@@ -576,11 +577,12 @@ function computeTextDimensions(g, gd) {
     }
     else {
         var text = g.selectAll('.legendtext'),
-            textSpans = g.selectAll('.legendtext>tspan'),
-            textLines = textSpans[0].length || 1;
+            textSpans = text.selectAll('tspan.line'),
+            textLines = textSpans.size() || 1,
+            textNode = text.node();
```
alexcjohnson: sure, that's a bit cleaner -> https://github.com/plotly/plotly.js/pull/1792/commits/27b4be3998b4580446efccde3f3e93183e4e7c6b

src/components/updatemenus/draw.js
```diff
@@ -214,7 +214,7 @@ function drawHeader(gd, gHeader, gButton, scrollBox, menuOpts) {
         .classed('user-select-none', true)
         .attr('text-anchor', 'end')
         .call(Drawing.font, menuOpts.font)
-        .text('▼');
+        .text(constants.arrowSymbol[menuOpts.direction]);
```
etpinard: Nice :tada:

tasks/test_syntax.js
```diff
@@ -77,6 +77,9 @@ function assertSrcContents() {
                     else if(lastPart === 'innerHTML') {
                         logs.push(file + ' : contains .innerHTML (IE failure in SVG)');
                     }
+                    else if(lastPart === 'parentElement') {
+                        logs.push(file + ' : contains .parentElement (IE failure)');
```
etpinard: For reference purposes: https://stackoverflow.com/questions/8685739/difference-between-dom-parentnode-and-parentelement

tasks/test_syntax.js
etpinard: Thanks!

src/constants/alignment.js
```diff
@@ -28,5 +28,7 @@ module.exports = {
         bottom: 1,
         middle: 0.5,
         top: 0
-    }
+    },
+    // multiple of fontSize to get the vertical offset between lines
+    LINE_SPACING: 1.3
```
etpinard: :tada: :tada: :tada: :tada: :tada:

src/plots/gl2d/camera.js
```diff
@@ -131,7 +131,7 @@ function createCamera(scene) {
 
                         if(Math.abs(dx * dydx) > Math.abs(dy)) {
                             result.boxEnd[1] = result.boxStart[1] +
-                                Math.abs(dx) * dydx * (Math.sign(dy) || 1);
+                                Math.abs(dx) * dydx * (dy >= 0 ? 1 : -1);
```
etpinard: [Non-blocking comment] It might be worth making a `Lib.sign` method at some point.

test/jasmine/tests/page_test.js
```diff
@@ -0,0 +1,88 @@
...
+            y: 1,
+            sizex: 0.2,
+            sizey: 0.2,
+            xanchor: 'right',
+            yanchor: 'bottom'
+        }];
+
+        // then merge in a few more with other component types
+        mock.layout.updatemenus = require('@mocks/updatemenus.json').layout.updatemenus;
+        mock.layout.sliders = require('@mocks/sliders.json').layout.sliders;
```
etpinard: Great idea here putting all traces and components in one `data/layout`. Maybe we should do the same in the svg/pdf/eps image export [tests](https://github.com/plotly/plotly.js/blob/master/test/image/export_test.js)

etpinard:
Amazing PR. :dancer: 

Text placement was always a tricky part of adding new layout components. This PR will make that so much easier. :tada:

alexcjohnson:
> Text placement was always a tricky part of adding new layout components. This PR will make that so much easier. 🎉

I hope so! One piece that's still pretty tricky is mathjax (because you have to support async rendering, and any bbox calls need to look for the mathjax group), and I'll try to address that later. In the meantime, we should make a practice of including `data-notex` (as I did with existing components in this PR) anytime we call `convertToTspans` if it hasn't been explicitly tested with tex. That way not only will it work consistently, but we'll have an easy tag to look for when someone wants to upgrade a component with tex compatibility.

PR #1794, opened 2017-06-16
Rework transform style into array syntax

rreusser:
This PR supersedes #1465 to rework keyed styles into an array of objects that's more plotly-friendly.


etpinard:
Well done. :dancer:

Issue #1807, opened 2017-06-21
Event data for transformed traces is inconsistent


rreusser:
This plot is throws an error whenever I hover in 1.28.2:

https://codepen.io/rsreusser/pen/bRRBjg

```
Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.exports.appendArrayPointValue (plotly-1.28.2.js:113510)
    at _hover (plotly-1.28.2.js:113987)
    at Object.hover (plotly-1.28.2.js:113608)
    at SVGRectElement.maindrag.onmousemove (plotly-1.28.2.js:138465)
```

Apologies this isn't a simpler plot type, but I've noticed this while working on groupby and the splinter tests now seem to be failing because of it. I tried and failed to reproduce it with a plain codepen, so this is the first time I've been able to reproduce it in a couple lines.

rreusser:
I don't know for sure, but it appears `trace._arrayAttrs` is undefined sometimes. That causes a failure on [this line](https://github.com/plotly/plotly.js/blob/b831cbccfaacc42a944b2a1af0f34a64ac77e12f/src/components/fx/helpers.js#L97). Is the solution to simply return if there are no arrayAttrs?

etpinard:
Thanks for heads-up.

That codepen you linked is blank. We should add a test before merging your PR

rreusser:
Hahaha fail. https://codepen.io/rsreusser/pen/bRRBjg

etpinard:
Oh hacky finance charts!

rreusser:
To be clear, I did see it on other trace types but wasn't able to reproduce it despite copying the JSON verbatim from `gd.data` in the workspace. I was a bit confused, so jumped on this reproduction, despite the OHLC.

etpinard:
> I did see it on other trace types but wasn't able to reproduce it despite copying the JSON verbatim from gd.data in the workspace. I was a bit confused, so I jumped on this reproduction, despite the OHL

Ok. In this case, the adding a fallback for undefined `_arrayAttrs` would be safest and best.

I'll like to get a fix in today, so I'll try to a few test cases to your https://github.com/plotly/plotly.js/pull/1808 PR. Thanks again for spotting this!

alexcjohnson:
> Ok. In this case, the adding a fallback for undefined `_arrayAttrs` would be safest and best.

[...]

PR #1808, opened 2017-06-22
Fix hover error in appendArrayPointValue

rreusser:
This PR address the issue in #1807 by exiting early if `arrayAttrs` is undefined. This may or may not be an appropriate fix. I could test this by passing it a plain object with no _arrayAttrs, but it seems to be of questionable value since `appendArrayPointValue` isn't tested otherwise. Can add, but I'll wait for a second opinion on the source of the problem.

### DIFF
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -1303,12 +1303,11 @@ describe('Test hover label custom styling:', function() {
 describe('ohlc hover interactions', function() {
     var data = [{
         type: 'candlestick',
+        x: ['2011-01-01', '2012-01-01'],
         open: [2, 2],
         high: [3, 3],
         low: [0, 0],
         close: [3, 3],
-        x: [1, 2, 3],
-        y: [0.12345, 0.23456, 0.34567]
     }];
 
     beforeEach(function() {
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -1299,3 +1299,27 @@ describe('Test hover label custom styling:', function() {
         .then(done);
     });
 });
+
+describe('ohlc hover interactions', function() {
+    var data = [{
+        type: 'candlestick',
+        open: [2, 2],
+        high: [3, 3],
+        low: [0, 0],
+        close: [3, 3],
+        x: [1, 2, 3],
+        y: [0.12345, 0.23456, 0.34567]
+    }];
+
+    beforeEach(function() {
+        this.gd = createGraphDiv();
+    });
+
+    afterEach(destroyGraphDiv);
+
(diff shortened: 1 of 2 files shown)


etpinard:
I don't see the need to test appendArrayPointValue im isolation. We should instead test "real" hover by mocking mouse events or Fx.hover. 

Moreover, I think it would be best to make _arrayAttrs be always defined for every trace type. When no array attributes are present, arrayAttrs should be `[] `

etpinard:
@rreusser have you tried adding any tests for this?

rreusser:
@etpinard Yup. Pushing.

rreusser:
I added it to the `hover_label_test` since it seems loosely related to hover data, but let me know if there's a better place for it.

test/jasmine/tests/hover_label_test.js
```diff
@@ -1299,3 +1299,27 @@ describe('Test hover label custom styling:', function() {
...
+
+describe('ohlc hover interactions', function() {
+    var data = [{
+        type: 'candlestick',
+        open: [2, 2],
+        high: [3, 3],
+        low: [0, 0],
+        close: [3, 3],
+        x: [1, 2, 3],
+        y: [0.12345, 0.23456, 0.34567]
```
rreusser: Oops. Does y do anything for ohlc? This looks like somehow leftover from another test, though it does trigger and respond to the fixed bug correctly.

etpinard: `y` isn't a valid attribute. :hocho:

test/jasmine/tests/hover_label_test.js
```diff
@@ -1299,3 +1299,27 @@ describe('Test hover label custom styling:', function() {
...
+    beforeEach(function() {
+        this.gd = createGraphDiv();
+    });
+
+    afterEach(destroyGraphDiv);
+
+    // See: https://github.com/plotly/plotly.js/issues/1807
+    it('should not fail in appendArrayPointValue', function() {
+        Plotly.plot(this.gd, data);
+        mouseEvent('mousemove', 203, 213);
```
etpinard: Maybe assert that the hover label is showing e.g. by

```js
expect(d3.select('.hovertext').size()).toBe(1);
```

test/jasmine/tests/hover_label_test.js
rreusser: 🔪 'd

test/jasmine/tests/hover_label_test.js
rreusser: Good call. Added ✅

etpinard:
:dancer: not a perfect solution (see https://github.com/plotly/plotly.js/issues/1807#issuecomment-310493998) but good enough for now and, more importantly, good enough for `1.28.3`

Issue #1597, opened 2017-04-13
Outline for text layers


alanocallaghan:
It would be very helpful to be able to add outlines to text layers. This allows for text to be legible on any background (eg white text with black outline, or black text with white outline). 

Also, though not strictly related, there does not seem to be any documentation of the options "familysrc", "sizesrc", or "colorsrc" anywhere.
https://plot.ly/javascript/reference/#bar-textfont
https://plot.ly/r/reference/#scatter-textfont

Related: https://github.com/talgalili/heatmaply/issues/42

etpinard:
> Also, though not strictly related, there does not seem to be any documentation of the options "familysrc", "sizesrc", or "colorsrc" anywhere.

Those shouldn't even be displayed in the docs pages. Thanks for pointing this out.

Alanocallaghan:
No worries, fyi there are a few similar things hanging around in the documentation, eg
https://plot.ly/javascript/reference/#bar-marker-colorbar-tickvalssrc

etpinard:
> It would be very helpful to be able to add outlines to text layers.

About this here, can you share a screenshot to help us understand what you're looking for?

Alanocallaghan:
Google maps uses dark text with a white outline, while white text with a dark outline also works. It's the most effective way to make text readable on any background 

http://asset0.cbsistatic.com/cnwk.1d/i/bto/20081009/google_maps_taxi_ad.png
http://2.bp.blogspot.com/-TGEt81O_qaI/Uxns70ZKVEI/AAAAAAAACE4/V5v81OF9sT4/s1600/white-text-black-outline.png

rreusser:
A note on implementation: for SVG, it requires two copies of the text element with different `stroke-width` properties (that is, it's possible but not just a simple property, which means it requires a bit more effort)

alexcjohnson:
Our much-neglected polar charts do this with the text-shadow style:
![screen shot 2017-04-18 at 10 13 56 am](https://cloud.githubusercontent.com/assets/2678795/25135240/c839cf8e-241f-11e7-997f-0cf6910907b9.png)


```
<text dy="0" y="0" x="0" dx="0" transform="rotate(45) translate(0,15)" style="
  text-anchor: start;
  font-size: 15px;
  font-family: Tahoma, sans-serif;
  fill: gray;
  text-shadow: white -1px 0px 0px, white 1px -1px 0px, white -1px 1px 0px, white 1px 1px 0px;
">0</text>
```
[...]

PR #1815, opened 2017-06-23
Contour line labels

### LINKED ISSUE #1395: contour labels
The 'contour' plot type is very useful but would be even more powerful if it included an automated facility for creating labels.    [Mapbox](https://www.mapbox.com/) seems to incorporate a particularly nice implementation of contour labeling as illustrated here:  https://www.mapbox.com/blog/satellite-map-with-contours/  

Given that the two systems already apparently have some connections perhaps something could be worked out to incorporate Mapbox's approach for contour labeling into Plotly for cases where the data are not defined in terms of lat / long but instead by x and y?

alexcjohnson:
Fixes #1395 

screen shot 2017-06-23 at 5 23 12 pm
screen shot 2017-06-23 at 5 24 15 pm

TODO:
- [x] contourcarpet (cc @rreusser): currently the attributes are there (so we don't get errors) but they don't work. I haven't looked to see how easy it will be to add these there, if it's too hard we could just leave it for later (and remove the attributes and make `supplyDefaults` work without them)
- [x] tests (basic, labels without lines, panned, with formatting) - since this work is all in the plot step it'll probably be all in image tests
- [x] there's no configurability to how many labels you get, or where they go (there is style configurability though, for font, color, size, number format). I spent a good while on the location optimizer - which tries to minimize angle and maximize distance from other labels (especially those on the same level) and from the edge. This looks pleasing to my eye across all our test plots (and on zoom/pan), but I'd be curious @geocosmite if you see a use for more control in this regard.

cc @etpinard


src/traces/contour/plot.js
```diff
@@ -259,11 +264,28 @@ function joinAllPaths(pi, perimeter) {
     return fullpath;
 }
 
-function makeLines(plotgroup, pathinfo, contours) {
+var TRAILING_ZEROS = /\.?0+$/;
+
+function makeLinesAndLabels(plotgroup, pathinfo, gd, cd0, contours, perimeter) {
```
rreusser: Except for the innermost join content, it looks like mostly d3 operations so is probably okay, but this function comes out to 240 lines, which seems a bit on the high side.

rreusser:
Very impressed with the greedy approach!

screen shot 2017-06-23 at 15 03 34
screen shot 2017-06-23 at 15 01 05
screen shot 2017-06-23 at 15 01 52
screen shot 2017-06-23 at 15 02 24

geocosmite:
This looks FABULOUS!

src/traces/contour/plot.js
```diff
@@ -291,18 +313,303 @@ function makeLines(plotgroup, pathinfo, contours) {
...
+                var totalPathLen = pathBounds.total;
+                var pathLen = onPlotMax - onPlotMin;
+
+                var isOpen = d3.select(this).classed('openline');
+
+                if(pathLen < textWidth * constants.LABELMIN) return;
+
+                var maxLabels = Math.min(Math.ceil(pathLen / normLength),
+                    constants.LABELMAX);
+                var dp, p0, pMax, minCost, location, pMin;
```
rreusser: Just a small note of caution that my syntax highlighter picks up `location` since `window.location` is special, though it's 100% fine here.

rreusser:
*Maybe* a minimum path length cost component so that there aren't numbers on closed curves too small to display them? Otherwise I'm having trouble finding any problems with it. 🎉 

screen shot 2017-06-23 at 18 56 44

alexcjohnson:
@rreusser what plot did you see those labels on tiny loops? There is a minimum of 3x the label length but perhaps we need to increase that.

src/traces/contour/plot.js
alexcjohnson: Ah good call, should be ok but I'll change it to avoid any confusion.

rreusser:
@alexcjohnson zoomed in section on: http://rickyreusser.com/plotly-mock-viewer/#contour_transposed

Also, I've been thinking about carpet plots. Since plotting is pretty trace-specific, I think it might have to be duplicated, though it's probably possible to factor out the code since there's nothing really contour-specific. I'll try to throw together a POC (though probably not much work required beyond that).

src/traces/contour/plot.js
rreusser: *Only* danger (I think) is is `var` declaration gets removed. Linter may or may not catch it. So really not dangerous at all… almost certainly 🔫

alexcjohnson:
@rreusser 

> Since plotting is pretty trace-specific, I think it might have to be duplicated, though it's probably possible to factor out the code since there's nothing really contour-specific.  I'll try to throw together a POC (though probably not much work required beyond that).

awesome, that'll help with [your previous comment](https://github.com/plotly/plotly.js/pull/1815#discussion_r123851404) too 🍰

src/traces/contour/plot.js
```diff
@@ -423,7 +423,7 @@ function makeLinesAndLabels(plotgroup, pathinfo, gd, cd0, contours, perimeter) {
 
                 var isOpen = d3.select(this).classed('openline');
 
-                if(pathLen < textWidth * constants.LABELMIN) return;
+                if(pathLen < (textWidth + textHeight) * constants.LABELMIN) return;
```
alexcjohnson: I didn't increase `LABELMIN`, instead included both width and height of the label. That way when you have long labels you'll still get them on about the same length paths as before, but short labels will require a bit longer paths than they did previously:
screen shot 2017-06-26 at 2 34 00 pm

Note that this isn't in the cost function, because that controls label location on the path but only weakly affects whether the path gets a label at all - ie if it pushes past `MAXCOST`

src/traces/contourcarpet/attributes.js
```diff
@@ -136,15 +92,7 @@ module.exports = extendFlat({}, {
...
-        showlines: {
-            valType: 'boolean',
-            dflt: true,
-            role: 'style',
-            description: [
-                'Determines whether or not the contour lines are drawn.',
-                'Has only an effect if `contours.coloring` is set to *fill*.'
-            ].join(' ')
-        },
+        showlines: contourContourAttrs.showlines,
```
etpinard: :palm_tree: thanks!

src/traces/contour/attributes.js
```diff
@@ -106,7 +107,33 @@ module.exports = extendFlat({}, {
...
+        },
+        showlabels: {
+            valType: 'boolean',
+            dflt: false,
+            role: 'style',
+            description: [
+                'Determines whether to label the contour lines with their values.'
+            ].join(' ')
+        },
+        font: extendFlat({}, fontAttrs, {
```
etpinard: I'd vote for `labelfont` here.

cc @cldougl

src/traces/contour/attributes.js
```diff
@@ -106,7 +107,33 @@ module.exports = extendFlat({}, {
...
+            description: [
+                'Sets the font used for labeling the contour levels.',
+                'The default color comes from the lines, if shown.',
+                // TODO: same size as layout.font, or smaller? 80%?
```
etpinard: `0.8` of `layout.font` sounds good to me.

src/traces/contour/plot.js
```diff
@@ -291,18 +311,303 @@ function makeLines(plotgroup, pathinfo, contours) {
...
+                Axes.calcTicks(formatAxis);
+                formatAxis._tmin = null;
+                formatAxis._tmax = null;
+            }
+            contourFormat = function(v) {
+                return Axes.tickText(formatAxis, v).text;
+            };
+        }
+
+        var dummyText = defs.append('text')
```
etpinard: Why did you append this dummy text node to `<defs>` instead of `Drawing.tester`?

etpinard:
> there's no configurability to how many labels you get,

I'm just curious, if we do add configurability down the road, how would it fit with this first version?

Maybe adding, `labelmode: 'auto' || 'manual'` where `'manual'` would be coupled with an array of boolean (e.g. `labels: [true, false, true]`) one value per contour level. I guess some users would want to control to number of labels per contour and the total number of labels per trace (e.g. with an `nlables` attribute similar to `nticks`).

src/traces/contour/attributes.js
alexcjohnson: I played with that a little, and ended up feeling like that was a little small:
screen shot 2017-06-26 at 4 57 18 pm

etpinard: OK. Let's keep it as `1 * layout.font` :+1:

src/traces/contour/plot.js
alexcjohnson: good question! Actually, I bet if it's already in `Drawing.tester` we can avoid the [`cloneNode` and `appendChild`] steps 🐎

src/traces/contour/attributes.js
cldougl: +1 for `labelfont`

alexcjohnson:
> I'm just curious, if we do add configurability down the road, how would it fit with this first version?
>
> Maybe adding, `labelmode: 'auto' || 'manual'` where `'manual'` would be coupled with an array of boolean (e.g. `labels: [true, false, true]`) one value per contour level. I guess some users would want to control to number of labels per contour and the total number of labels per trace (e.g. with an `nlables` attribute similar to `nticks`).

Good question. My feeling is to control which contours to label, the most intuitive way would be to specify an array of levels, which could even be making an array form of `showlabels`, like `showlabels=[10, 20, 50, 100]` And then to control how many labels to show per contour, I'd probably just make some parameter like `labeldensity` that adjusts a few of the now constants (perhaps `LABELDISTANCE`, `LABELMAX`, and `MAXCOST`?) to more or less linearly increase or decrease the number we display (with the current constants what you get with the default `labeldensity=1`), without trying to explicitly specify a number. Partly I say that because of the heuristic nature of the placement algorithm, which would make it hard to arrive at a specific number, and partly because when you zoom and pan you probably don't want the number of labels displayed to be constant.

But I'd like to do major/minor contours first, as we might have different ideas after implementing that.

src/traces/contour/plot.js
alexcjohnson: 🐎  https://github.com/plotly/plotly.js/pull/1815/commits/539ca224de5d350373782eec85a726a6818fe716

src/traces/contour/attributes.js
alexcjohnson: -> `labelFont` in https://github.com/plotly/plotly.js/pull/1815/commits/2116a124eb68aa3f8e7ae051d36bf03862f72afc

src/plot_api/plot_api.js
```diff
@@ -1374,7 +1374,8 @@ function _restyle(gd, aobj, _traces) {
         'line.cmin', 'line.cmax',
         'marker.line.cmin', 'marker.line.cmax',
         'contours.start', 'contours.end', 'contours.size',
-        'contours.showlines',
+        'contours.showlines', 'contours.showlabels', 'contours.labelformat',
```
etpinard: Wanna try adding `editType` flags their attribute declarations instead as in https://github.com/plotly/plotly.js/pull/1653 ?

alexcjohnson: @etpinard I had a misstep in https://github.com/plotly/plotly.js/pull/1815/commits/e5a2f91e0c5e3e7a25fd463398fbd0322c6dd908 when I thought we could hunt for `editType` inside containers, but your awesome plot_api tests showed that wasn't going to work - so removed that piece in https://github.com/plotly/plotly.js/pull/1815/commits/cbdf795d2c1b1154fd5968855636d163d986aab4 - look reasonable? Another baby step to cleaning up `restyle` and `relayout`...

src/plot_api/edit_types.js
```diff
@@ -0,0 +1,84 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var isPlainObject = require('../lib/is_plain_object');
+
+module.exports = {
```
etpinard: good idea :tada: 

I don't think this deserved a file of its own though. Putting this in

https://github.com/plotly/plotly.js/blob/master/src/plot_api/helpers.js

should suffice.

alexcjohnson: Ah sure, I can do that. I was worried about circular deps if it lived in either `plot_api.js` or `plot_schema.js` but looks like it should work fine in `helpers.js`

src/plot_api/plot_schema.js
```diff
@@ -62,7 +64,11 @@ exports.get = function() {
     return {
         defs: {
             valObjects: Lib.valObjects,
-            metaKeys: UNDERSCORE_ATTRS.concat(['description', 'role'])
+            metaKeys: UNDERSCORE_ATTRS.concat(['description', 'role']),
+            editTypes: {
```
etpinard: Ha. Sure, might as well list those things.

src/plot_api/edit_types.js
alexcjohnson: Actually, turns out putting these in helpers makes new circular references and breaks plotting entirely.

- plots/plots, plot_api/plot_schema, plot_api/helpers
- plots/cartesian/axes, components/titles/index, plotly, components/modebar/index, components/modebar/manage, components/modebar/buttons, plots/plots, plot_api/plot_schema, plot_api/helpers

The real culprit here seems to be `plots/plots.js`, that ancient beast is coupled to everything under the sun (and indeed is part of 3 other loops too)... but do you mind if I leave `edit_types.js` as a separate file for now?

etpinard: > do you mind if I leave edit_types.js as a separate file for now?

Sure, that's fine :+1:

src/traces/contourcarpet/index.js
```diff
@@ -15,7 +15,7 @@ ContourCarpet.supplyDefaults = require('./defaults');
 ContourCarpet.colorbar = require('../contour/colorbar');
 ContourCarpet.calc = require('./calc');
 ContourCarpet.plot = require('./plot');
-ContourCarpet.style = require('./style');
```
alexcjohnson: continuing to 🌴  contour/contourcarpet... this also fixes some bugs such as:
- contourcarpet failed with `contours.coloring='lines'`
- if the first contour wasn't shown (if you give an explicit contour start value lower than what the data would be) the colors would have been wrong. For contour this was fixed in #1309 (the key is to use `d.level` rather than `i` in coloring the contour lines/fills) and tested in `contour_edge_cases`.

@etpinard do you think it's sufficient to rely on the contour tests for contourcarpet too - since it's using the same code now - or would you like me to add contourcarpet tests of these cases?

src/traces/contour/style.js
```diff
@@ -31,9 +31,13 @@ module.exports = function style(gd) {
         var line = trace.line;
         var cs = contours.size || 1;
         var start = contours.start;
-        var colorLines = contours.coloring === 'lines';
 
-        var colorMap = makeColorMap(trace);
+        // for contourcarpet only - is this a constraint-type contour trace?
```
alexcjohnson: A bit funny to have `contourcarpet only` code in `contour.style` - but perhaps someday we would like to support constraint-type coloring for regular contour plots too?

src/traces/contourcarpet/index.js
etpinard: > do you think it's sufficient to rely

We have plenty of carpet mocks, maybe you could update one of them with `contours.coloring='lines'` ?

src/traces/contourcarpet/index.js
```diff
@@ -15,7 +15,7 @@ ContourCarpet.supplyDefaults = require('./defaults');
...
+ContourCarpet.style = require('../contour/style');
```
etpinard: Amazing :tada:

src/traces/contourcarpet/index.js
alexcjohnson: hmm, as far as I can tell only `cheater_contour` is *not* constraint-type, but perhaps I can extend that one to cover these cases. Sure.

alexcjohnson: https://github.com/plotly/plotly.js/pull/1815/commits/0ff2dc469ece4897eae42eefe88df1aa03601d28
This mock on master:
screen shot 2017-06-29 at 6 39 10 pm
and now:
screen shot 2017-06-29 at 6 40 20 pm

src/traces/carpet/defaults.js
```diff
@@ -21,6 +21,8 @@ module.exports = function supplyDefaults(traceIn, traceOut, dfltColor, fullLayou
         return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
     }
 
+    traceOut._clipPathId = 'clip' + traceOut.uid + 'carpet';
```
alexcjohnson: @rreusser FYI - it turns out that we weren't applying the clip path to mocks that have the data trace before the carpet trace because the ID was cached during the plot step. I moved `_clipPathId` up to `supplyDefaults` to get around that. It's barely visible in the test images, but that's why they all changed (all the ones that put the data before the carpet anyway).

Also two more changes in here:
- `_clipPathId` rather than `clipPathId` - if we're attaching something to the trace (or layout) that's **not** an attribute, it should either be a function or it should start with an underscore.
- use `Drawing.setClipUrl` rather than directly setting the `clip-path` attribute. This is to handle pages with a `<base>` tag, which is obnoxious but important if it's present - which happens most notably in angular 😑

rreusser: Oh wow… Good catch! Yeah, I can imagine that's a subtle half-border-width-effect. 👍

src/components/annotations/draw.js
```diff
@@ -487,15 +487,15 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
             // to get the parity of the number of intersections.
             if(edges.reduce(function(a, x) {
                 return a ^
-                    !!lineIntersect(headX, headY, headX + 1e6, headY + 1e6,
+                    !!Lib.segmentsIntersect(headX, headY, headX + 1e6, headY + 1e6,
```
rreusser: Not part of this PR and I'm sure no meaningful performance hit, but is `reduce` necessary here? Seems like

```js
for (i = 0; i < edges.length; i++) {
  if (!Lib.segmentsIntersect(...)) return;
}
```

would be more concise and readable and would bail out early when it can.

src/components/annotations/draw.js
```diff
@@ -487,15 +487,15 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
...
                             x[0], x[1], x[2], x[3]);
             }, false)) {
                 // no line or arrow - so quit drawArrow now
                 return;
             }
 
             edges.forEach(function(x) {
```
rreusser: Is `forEach` avoidable here?

src/components/drawing/index.js
```diff
@@ -601,25 +598,37 @@ drawing.makeTester = function() {
...
+ * @param {boolean} inTester: is this element already in `drawing.tester`?
+ *   If you are measuring a dummy element, rather than one you really intend
+ *   to use on the plot, making it in `drawing.tester` in the first place
+ *   allows us to test faster because it cuts out cloning and appending it.
+ *
+ * @param {string} hash: for internal use only, if we already know the cache key
+ *   for this element beforehand.
+ *
+ * @return {object}: a plain object containing the width, height, left, right,
+ *   top, and bottom of `node`
```
rreusser: ❤️  📚

src/components/drawing/index.js
```diff
@@ -601,25 +598,37 @@ drawing.makeTester = function() {
...
  */
 drawing.savedBBoxes = {};
 var savedBBoxesCount = 0;
 var maxSavedBBoxes = 10000;
 
-drawing.bBox = function(node, hash) {
+drawing.bBox = function(node, inTester, hash) {
```
rreusser: 👍 technically an internal API used elsewhere, but a quick grep I can't find any cases where the argument signature change is invalid.

src/lib/geometry2d.js
```diff
@@ -0,0 +1,195 @@
...
+        c = x4 - x3,
+        d = y2 - y1,
+        e = y3 - y1,
+        f = y4 - y3,
+        det = a * f - c * d;
+    // parallel lines? intersection is undefined
+    // ignore the case where they are colinear
+    if(det === 0) return null;
+    var t = (b * f - c * e) / det,
+        u = (b * d - a * e) / det;
```
rreusser: If I were really going to nitpick, I'd suggest comparing bounding boxes first and here multiplying through by `det` to eliminate division, but this is probably not a bottleneck…

src/lib/geometry2d.js
```diff
@@ -0,0 +1,195 @@
...
+ *        and are at the same point
+ * }
+ *
+ * Works by starting from either end and repeatedly finding the distance from
+ * that point to the plot area, and if it's outside the plot, moving along the
+ * path by that distance (because the plot must be at least that far away on
+ * the path). Note that if a path enters, exits, and re-enters the plot, we
+ * will not capture this behavior.
+ */
+exports.getVisibleSegment = function getVisibleSegment(path, bounds, buffer) {
```
rreusser: Oh man, I've implemented something similar to this before. It's such a pain!

src/traces/scattercarpet/plot.js
```diff
@@ -37,6 +38,6 @@ module.exports = function plot(gd, plotinfoproxy, data) {
         // separately to all scattercarpet traces, but that would require
         // lots of reorganization of scatter traces that is otherwise not
         // necessary. That makes this a potential optimization.
-        node.attr('clip-path', 'url(#clip' + carpet.uid + 'carpet)');
+        Drawing.setClipUrl(node, carpet._clipPathId);
```
etpinard: Ha. That means carpet traces will now be functional in Angular 1.0 see https://github.com/plotly/plotly.js/pull/509

alexcjohnson: yeah, as mentioned in https://github.com/plotly/plotly.js/pull/1815#discussion_r124943024

Tried to figure out if we have a way to enforce that... only allow `attr('clip-path', ...)` inside `Drawing.setClipUrl` - like something to add to `strict-d3`? Seemed pretty invasive and hacky though.

src/traces/carpet/plot.js
```diff
@@ -96,13 +90,9 @@ function drawClipPath(trace, t, layer, xaxis, yaxis) {
...
-    trace.clipPathData = 'M' + segs.join('L') + 'Z';
-    clip.attr('id', trace.clipPathId);
-    path.attr('d', trace.clipPathData);
-        // .style('stroke-width', 20)
-        // .style('vector-effect', 'non-scaling-stroke')
-        // .style('stroke', 'black')
-        // .style('fill', 'rgba(0, 0, 0, 0.1)');
+    var clipPathData = 'M' + segs.join('L') + 'Z';
+    clip.attr('id', trace._clipPathId);
+    path.attr('d', clipPathData);
```
rreusser: Ah, thanks for removing. I left those because I so frequently wanted to toggle over to *displaying* the clip path instead of *using* the clip path. But no good reason to keep.

src/traces/contour/attributes.js
```diff
@@ -36,6 +37,7 @@ module.exports = extendFlat({}, {
         valType: 'boolean',
         dflt: true,
         role: 'style',
+        editType: 'docalc',
```
rreusser: 😮 

Now I understand. Nice.

src/traces/contour/plot.js
```diff
@@ -13,14 +13,19 @@ var d3 = require('d3');
...
 var heatmapPlot = require('../heatmap/plot');
 var makeCrossings = require('./make_crossings');
 var findAllPaths = require('./find_all_paths');
 var endPlus = require('./end_plus');
+var constants = require('./constants');
+var costConstants = constants.LABELOPTIMIZER;
 
 
-module.exports = function plot(gd, plotinfo, cdcontours) {
+exports.plot = function plot(gd, plotinfo, cdcontours) {
```
rreusser: Named for testing?

src/traces/contour/plot.js
```diff
@@ -259,50 +264,380 @@ function joinAllPaths(pi, perimeter) {
...
+    lineClip.enter().append('clipPath')
+        .classed('contourlineclip', true)
+        .attr('id', clipId);
+
+    Drawing.setClipUrl(lineContainer, clipId);
+
+    return lineClip;
+};
+
+exports.labelFormatter = function(contours, colorbar, fullLayout) {
```
rreusser: Interesting. Does this have any more general applicability? I feel like piggybacking on this is a common need, and I struggled to figure out just which parts and in what sequence it should be done.

etpinard:
Amazing PR! 

This will surely make the daunting contour code more-easily reusable (`contourgeo` anyone?).

:dancer:

src/traces/contour/plot.js
```diff
@@ -259,50 +264,380 @@ function joinAllPaths(pi, perimeter) {
...
+    return {
+        text: text,
+        width: bBox.width,
+        height: bBox.height,
+        level: level,
+        dy: (bBox.top + bBox.bottom) / 2
+    };
+};
+
+exports.findBestTextLocation = function(path, pathBounds, textOpts, labelData, plotBounds) {
```
rreusser: ❤️ ❤️ ❤️ Reusable for carpet?

src/traces/contourcarpet/index.js
rreusser: This is great. I'd have to dig in further to refresh my memory, but I struggled a bit to wrangle the behavior under control for both constraints and contours since it felt like they had slightly different semantics. Like constraints tend to be a single color while contours have multiple colors. But if it passes the existing image tests and fixes other behaviors, then I'm 100% content.

src/traces/contour/plot.js
rreusser: Reuse in carpet. Got it.

rreusser:
Looks great to me! Thrilled for carpet to get a close second look. One nitpick is that `-5.0` and `5.0` are sometimes hard to distinguish since the `-` tends to blend into the surrounding contour. But I don't really see how we can fix that, unless the label padding were configurable so that you could add a bit more spacing.

💃

src/components/annotations/draw.js
rreusser: Unless that equates to an `all` condition where you actually need to check all. Not 100% sure. The `^` is a bit confusing.

alexcjohnson:
> One nitpick is that `-5.0` and `5.0` are sometimes hard to distinguish since the `-` tends to blend into the surrounding contour.

Hmm, interesting - I suppose we could either add some extra space ahead of a leading `-`, or replace it with a different unicode symbol that's raised up a bit, perhaps? I'll leave it as is for the first iteration but will take a look at it later.

etpinard:
> Hmm, interesting - I suppose we could either add some extra space ahead of a leading -, or replace it with a different unicode symbol that's raised up a bit, perhaps? I'll leave it as is for the first iteration but will take a look at it later.

... or implement https://github.com/plotly/plotly.js/issues/1597

rreusser:
This is the only other variation that jumps out. Which I don't *love*. Or `+1`/`-1` instead of `1`/`-1`, which I don't like either. I think the text shadow method seems best.

overlay_9_3_lg

Issue #1816, opened 2017-06-24
Inverted coordinates returned on click on scattergeo?


collioud:
Hi,

When clicking on a point on a scattergeo plot, the lat/lon returned values seem to be inverted.
See [https://codepen.io/anon/pen/VWzOXq?editors=0010](https://codepen.io/anon/pen/VWzOXq?editors=0010)
Try to click on a point to add 2 traces:  the red one corresponds to lat set to lat_clicked_pt and lon to lon_clicked_pt ;   the green one is when lon and lat are inverted.
Obviously, the green one is expected...

etpinard:
Nice catch! Yep, you're right. These [lines](https://github.com/plotly/plotly.js/blob/5e70d3b04cb32e1c8e5a4a7460915764a338cefe/src/traces/scattergeo/hover.js#L69-L70) should be:

```js
pointData.lon = lonlat[0];
pointData.lat = lonlat[1];
```

PR #1819, opened 2017-06-26
Fixup geo event data

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1816

- lon and lat were swapped previously due to typo
- 'locations' items weren't handled properly

Thanks @collioud for the report.


alexcjohnson:
Haha I guess just because we *have* tests, doesn't guarantee they're testing the correct result!

💃

Issue #1814, opened 2017-06-23
HTML entities no longer render in ticktext


cpsievert:
Introduced in 1.28.0 release

https://codepen.io/cpsievert/pen/ZyJoPW

etpinard:
Thanks for the report. Caused by https://github.com/plotly/plotly.js/pull/1792. I can't believe no test failed to catch this.

etpinard:
cc @alexcjohnson

etpinard:
Here's the list of entities we should support:

https://github.com/plotly/plotly.js/blob/master/src/constants/string_mappings.js

PR #835, opened 2016-08-09
Decode some html entities

etpinard:
In https://github.com/plotly/plotly.js/pull/804, we dropped html entity decoding for performance reasons, but apparently some folks have been using `&amp;`, `&nbsp;`, etc in their graphs. We need to bring back support for these ASAP. 

So, I propose supporting a few entities by translating them one-by-one to unicode.

This PR also DRYs up some code by making sure that `svg_text_utils.js` (for svg text) and `html2unicode` (for WebGL text) use the same mappings.

PR #804, opened 2016-08-01
Remove HTML entity decoding from convertToSVG

scjody:
Remove HTML entity decoding from convertToSVG - this is unnecessary and imposes a performance hit on long strings.

PR #1820, opened 2017-06-26
Fixup html entity number

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1814

Previously HTML entities entered as numbers were decoded to unicode _for free_ via now-removed [`appendSVG`](https://github.com/plotly/plotly.js/commit/cbbb2d695f662c0a8c57a535d579a98a465587a5#diff-f198093b7ae79ab6392778ae670747a6L44) method. When PR https://github.com/plotly/plotly.js/pull/1792 :hocho: `appendSVG`, support for HTML entity numbers was dropped. This PR brings it back.

cc @cpsievert


alexcjohnson:
Looks good. Referencing #804 and #835 for why we're only concerned with these few entities. I may hunt around for a performant way to support *all* entities, but in the meantime this gets a 💃 !

Issue #1813, opened 2017-06-23
Major memory leak in 2D WebGL plots


kevineschmidt:
The Plotly.purge does not fully clean up a 2D OpenGL plot.  It seems that the camera associated with the 2D scene leaves some mouse handlers attached, which then keep a reference to the pickbuffer in  GLPlot2D, and in that, some very large Uint32Arrays.  

I ran the following plnkr in Chrome under an allocation timeline.  Repeatedly click the button.
https://plnkr.co/edit/7fve7JWbJrQy6COqJdIv?p=preview

In the function purgePlot, you can see there are two lines commented out that uninstall them and get rid of that large leak.

kevineschmidt:
Intersting, in the allocation timeline, you don't see much recovered in the summary timeline at top, but the Retained Size is quite large.

Also it seems like you can't just change the code & rerun the allocation - Chrome tools don't seem to pick up the change. Stop plnkr  , close dev tools, make the change, start plnkr, open devtools

etpinard:
cc @dfcreative

dfcreative:
Ok seems that [fixed](https://github.com/dfcreative/plotly.js/tree/mem-unleak).

Before:
![image](https://user-images.githubusercontent.com/300067/27553619-a01dcad8-5a79-11e7-8af3-abc063ef93af.png)

After:
![image](https://user-images.githubusercontent.com/300067/27553664-d971c2f8-5a79-11e7-895d-bc71c1ec8204.png)

dfcreative:
[Fix for scatter3d camera](https://github.com/dfcreative/plotly.js/commit/888ee1b5be96533e446e63cdef6341d584952d38).

etpinard:
@dfcreative amazing :tada: 

PR away :rocket:

etpinard:
fixed in https://github.com/plotly/plotly.js/pull/1821

PR #1821, opened 2017-06-26
Fix scatter camera memory leak

dfcreative:
Covers https://github.com/plotly/plotly.js/issues/1813


etpinard:
Nicely done. :tada:

Issue #1777, opened 2017-06-12
Allow to choose whether to trim trace names in hover info


simoneb:
Because of https://github.com/plotly/plotly.js/blob/master/src/components/fx/hover.js#L703 trace names longer than 15 chars get trimmed. It would be useful to provide an option to decide whether to trim or not

etpinard:
Yeah for sure. Thanks for writing in!

simoneb:
@etpinard thanks for the feedback. 

To expand on this, it would also be useful to be able to specify the maximum number of chars to trim to, or not at all. If you can give me some hints about where you think this configuration should live and what shape it may have based on your coding standards I can also create a PR. My first guess would be an additional option next to `hoverinfo` in a trace configuration, but maybe there's a better, upstream location for such a configuration?

alexcjohnson:
@simoneb definitely, making this attribute an integer seems like the way to go. I'd put it in [hoverlabel](https://github.com/plotly/plotly.js/blob/master/src/components/fx/attributes.js#L15), perhaps `hoverlabel.trimname` or `hoverlabel.namelength`? Then a value of `0` would mean "no trimming" (or perhaps `false` would be clearer? It's not an integer but we do mix types for purposes like that elsewhere) and the `dflt` would be `15`.

etpinard:
+1 for `hoverlabel.namelength`

PR #1822, opened 2017-06-27
Hover label namelength

ckiss:
resolves https://github.com/plotly/plotly.js/issues/1777

### DIFF
--- a/src/components/fx/attributes.js
+++ b/src/components/fx/attributes.js
@@ -38,6 +38,7 @@ module.exports = {
             valType: 'number',
             min: 0,
             dflt: 15,
+            arrayOk: true,
             role: 'style',
             description: 'Sets the length (in number of characters) of the hover labels for this trace'
         }
--- a/src/components/fx/attributes.js
+++ b/src/components/fx/attributes.js
@@ -35,12 +35,18 @@ module.exports = {
             color: extendFlat({}, fontAttrs.color, { arrayOk: true })
         },
         namelength: {
-            valType: 'number',
-            min: 0,
-            dflt: 15,
+            valType: 'integer',
+            min: -1,
             arrayOk: true,
             role: 'style',
-            description: 'Sets the length (in number of characters) of the hover labels for this trace'
+            description: [
+                'Sets the length (in number of characters) of the trace name in',
+                'the hover labels for this trace. -1 shows the whole name',
+                'regardless of length. 0-3 shows the first 0-3 characters, and',
+                'an integer >3 will show the whole name if it is less than that',
+                'many characters, but if it is longer, will truncate to',
+                '`namelength - 3` characters and add an ellipsis.'
+            ].join(' ')
         }
     }
 };
--- a/src/components/fx/attributes.js
+++ b/src/components/fx/attributes.js
@@ -33,6 +33,13 @@ module.exports = {
             family: extendFlat({}, fontAttrs.family, { arrayOk: true }),
             size: extendFlat({}, fontAttrs.size, { arrayOk: true }),
             color: extendFlat({}, fontAttrs.color, { arrayOk: true })
+        },
+        namelength: {
+            valType: 'number',
+            min: 0,
+            dflt: 15,
+            role: 'style',
+            description: 'Sets the length (in number of characters) of the hover labels for this trace'
         }
     }
 };
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -689,7 +689,9 @@ function createHoverText(hoverData, opts, gd) {
             // strip out our pseudo-html elements from d.name (if it exists at all)
             name = svgTextUtils.plainText(d.name || '');
 
-            if(name.length > commonLabelOpts.namelength && commonLabelOpts.namelength > 0) name = name.substr(0, commonLabelOpts.namelength - 3) + '...';
+            if(commonLabelOpts.namelength > 0 && name.length > commonLabelOpts.namelength) {
+                name = name.substr(0, commonLabelOpts.namelength - 3) + '...';
+            }
         }
 
         // used by other modules (initially just ternary) that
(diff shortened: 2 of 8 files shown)


ckiss:
@etpinard as I said I'll do, I opened a new PR, with fixed tests. Let me know what you think. I still have some concerns that I might need to do some updates to the documentation or something. Not sure how that goes here :)

src/components/fx/attributes.js
```diff
@@ -33,6 +33,13 @@ module.exports = {
             family: extendFlat({}, fontAttrs.family, { arrayOk: true }),
             size: extendFlat({}, fontAttrs.size, { arrayOk: true }),
             color: extendFlat({}, fontAttrs.color, { arrayOk: true })
+        },
+        namelength: {
+            valType: 'number',
+            min: 0,
+            dflt: 15,
+            role: 'style',
```
etpinard: We should try adding `arrayOk` support here, so that 

```js
trace = {
  x: [1, 2, 3],
  y: [2, 1, 2],
  namelength: [15, 10, 40]
}
```
would work.

src/components/fx/hover.js
```diff
@@ -689,7 +689,7 @@ function createHoverText(hoverData, opts, gd) {
             // strip out our pseudo-html elements from d.name (if it exists at all)
             name = svgTextUtils.plainText(d.name || '');
 
-            if(name.length > 15) name = name.substr(0, 12) + '...';
+            if(name.length > commonLabelOpts.namelength && commonLabelOpts.namelength > 0) name = name.substr(0, commonLabelOpts.namelength - 3) + '...';
```
etpinard: Does this take care of `namelength = 0` case which would hide the name text node completely?

src/components/fx/attributes.js
etpinard: see https://github.com/plotly/plotly.js/pull/1761 for an example on how to proceed.

src/components/fx/hover.js
etpinard: ++ can you rewrite this as

```js
if(/* */) {

}
```

ckiss: @etpinard my understanding from this [comment](https://github.com/plotly/plotly.js/issues/1777#issuecomment-308321146) is that when `namelength = 0` there should be no trimming happening, instead display the whole text... Am I wrong?

etpinard: Too bad the JS token `Infinity` isn't JSON serializable. I'm ok with making `namelength: 0` means no truncation, even though it feels contradictory to me.

Perhaps we should instead add support for an `Infinity`-like value in `valType: 'number'` attributes. For example, we could make `namelength: null` correspond to `namelength === Infinity`. Though, I can't think of any other attributes that could benefit from it, so I'm not sure it's worth the time.

cc @alexcjohnson

rreusser: `namelength: -1`? In the style of indexOf, that says "does not apply" to me. Sort of.

etpinard: @rreusser solid suggestion, I like it :+1:

alexcjohnson: `-1` sounds good for "no truncation." I'm not sure *why* someone would want to use `namelength=0` to completely hide the name, rather than using `hoverinfo` but I bet someone will come up with a clever reason for it!

src/components/fx/attributes.js
alexcjohnson: > We should try adding arrayOk support here

good call - at first I was worried that this would just be confusing to look at ("I thought this was the same trace, why does it have a different name?") but it could actually come in handy for example if you vary the hover label font size, a bigger one may need more truncation of the name 👍

etpinard:
@ckiss are you planning on addressing the comments above?

I might have time to wrap this one up on Monday.

ckiss:
@etpinard I'll address them Monday morning! Have a great weekend and sorry for the delay!

alexcjohnson:
@ckiss I hope you don't mind, I took the liberty of finishing this up so we can get it in the next minor release. I switched to -1 for "no truncation", fleshed out the `arrayOk` support, and added a test of all these behaviors. Also, I realized that lengths 1-3 would have been meaningless as we had them, as it would be only ellipsis, so for those short lengths I omit the ellipsis and just show the leading characters. Sound reasonable?

@etpinard any comments?

etpinard:
> @etpinard any comments?

94f660a looks good. You'll need to add `'arrayOk'` to this [list](https://github.com/plotly/plotly.js/blob/9ab907fc277123fb84f5cb2c2069702812610407/src/lib/coerce.js#L97) to get the plot-schema tests to pass.

alexcjohnson:
> You'll need to add `'arrayOk'` to this [list](https://github.com/plotly/plotly.js/blob/9ab907fc277123fb84f5cb2c2069702812610407/src/lib/coerce.js#L97) to get the plot-schema tests to pass.

hah, first place we have an integer array, really? 😎  https://github.com/plotly/plotly.js/pull/1822/commits/e9375192e3675f5902bbd0012642018ce1bc949c

src/components/fx/layout_attributes.js
```diff
@@ -57,11 +57,18 @@ module.exports = {
             color: extendFlat({}, fontAttrs.color)
         },
         namelength: {
-            valType: 'number',
-            min: 0,
+            valType: 'integer',
+            min: -1,
             dflt: 15,
```
ckiss: don't you need to remove `dflt` here too?

ckiss:
@alexcjohnson @etpinard sorry I didn't had time to complete this :( I was back from holidays monday and had lots on my plate... thanks for your help @alexcjohnson. I added a comment to your commit. Otherwise I'm happy, although [where](https://github.com/plotly/plotly.js/issues/1777#issuecomment-308321146) this was initially discussed a default value was mentioned.

src/components/fx/layout_attributes.js
alexcjohnson: I kept only the global default, because the trace-level default [inherits](https://github.com/plotly/plotly.js/pull/1822/files#diff-2d7db1bff018d6a0f823bed5c4994b69R18) from the global so doesn't use a default of its own.

You're right that we want a default - in fact, we *need* a default of 15 to preserve backward compatibility - this is just how that default gets propagated.

ckiss:
cool, thanks!

Issue #1385, opened 2017-02-15
Option to prevent symbol truncation at ternary axis boundaries 


geocosmite:
I have scoured the documentation but have been unable to find a way to prevent symbols from being truncated by axis boundaries. While you can avoid the problem by extending the boundaries beyond what you would normally use for the 'scatter' plot type (to slightly below zero for instance for cases where zero is the minimum data value), this is not a possibility with the 'scatterternary' type, at least for cases where points lie on an axis or apex.

To see the impact of the symbol truncation consider the plot below left where points at the apices are almost invisible and those in the middle of the axes also are quite small. This display is problematic because the center point has much greater visual heft compared to the remaining points, particularly compared to those on the apices. The preferred plotting approach is depicted on the right where each point has an equal visual weight irrespective of its position on the plot.

![image](https://cloud.githubusercontent.com/assets/25803192/22993673/0a5201ea-f381-11e6-98ce-8ad65c11edd7.png)

etpinard:
Thanks for pointing this out.

geocosmite:
My pleasure!  I appreciate your taking the time to consider the issue.  

What do you think about the idea of having a boolean attribute to provide the desired behavior?  The default value could be _true_ to be consistent with the current implementation but _false_ would produce a plot analogous to the "desired display" example above.

In most cases my guess is that you would either want no truncation of markers at the axes at all on the plot or to have the plot behave as it does now so this attribute could be defined at the top level of **layout** (_markeraxistrunctate_?).  I suppose, however, that having the ability to define this attribute on the **axis** (_markertruncate_?) or the trace **marker** (_axistruncate_?) levels would allow for finer-grained control at the expense of a more verbose configuration.

etpinard:
Quick question: should the marker point be placed above the axis tick labels? 


![image](https://cloud.githubusercontent.com/assets/6675409/26732709/881e1718-4786-11e7-8958-a01a3cb7b44e.png)


Or should they be placed in-between the axis bounding line and tick labels?

![image](https://cloud.githubusercontent.com/assets/6675409/26732692/736c2d6e-4786-11e7-9018-405d2f29374e.png)

geocosmite:
Hi Étienne.  Thanks for the question.  The marker points near the axis boundaries should go on top of the axis line and tick labels.  My rationale is that this seems to be more consistent with the objective of ensuring that they have visual weights that are equivalent to their counterparts near the center of the plot region.

etpinard:
[...]

PR #1824, opened 2017-06-27
Implement scatter `cliponaxis: false`

etpinard:
Resolves https://github.com/plotly/plotly.js/issues/1385

In brief, this PR adds a new layer to each cartesian and ternary (and eventually geo) subplot where the clipping `<rect>` isn't applied. Instead, points (i.e. markers, text nodes and error bars) are removed if their (x,y) or (a,b,c) coordinates are outside the subplot's current axis range box as discussed in https://github.com/plotly/plotly.js/issues/1385#issuecomment-310176186.

Note that, line and fills are never plotted in the `noclip` layer. This fact led to some non-trivial changes in `scatter/plot.js`. I would enjoy a thorough review from @alexcjohnson and @rreusser there. That said, as the scatterternary module reuses `scatter/plot.js`, implement the `cliponaxis: false`  behavior was pretty DRY :palm_tree:.

TODO:
- [ ] agree on attribute name (cc @cldougl )
- [ ] test animations (cc @rreusser might need your help here)
- [ ] test overlaying axes

Maybes:
- implement `cliponaxis: false` for `scattergeo` (might leave this for another PR)

cc @geocosmite


alexcjohnson:
Right now the layering depends on `cliponaxis` - I think we need to preserve the previous behavior, which probably means keeping the layer structure as it was but pushing the clipping down to the level of point groups and line groups. One simple example where this is not desired:

```js
Plotly.newPlot(gd,[
    {y: [1,1,1,1,1], mode:'markers+lines', marker: {size: 20}},
    {x:[0, 1.5, 2.5, 4], y: [0, 2, 0, 2], mode: 'markers+lines', marker: {size: 20}}
],
{width: 400, height: 400, xaxis: {range: [0, 4]}, yaxis: {range: [0, 2]}})
```
screen shot 2017-06-27 at 6 24 09 pm

```js
Plotly.restyle(gd,'cliponaxis',false)
```
screen shot 2017-06-27 at 6 24 35 pm
The blue points are no longer behind the orange lines.

I guess if you have two marker traces there would be an even more dramatic effect if you made the first one `cliponaxis: false` and the second `true`: the markers of the first would jump on top of the second. And if you had a dense field of markers from one trace, with a sparse `'lines+markers'` in front of it, both `cliponaxis: false`, the top trace would lose its lines and have disconnected markers visible...

Also: we should recalculate visible markers during zoom/pan (as well as during animation, which you already mentioned), so we don't get situations like these (dragging the top right corner out and then in):
screen shot 2017-06-27 at 6 36 33 pm
screen shot 2017-06-27 at 6 36 43 pm

alexcjohnson:
re: attribute name: Since this applies only to markers I'd put it in the `marker` container, but I'm not sure what to call it. Just to throw some things out there:
`marker.edgeeffect: 'clip'/'hide'` (`'hide'`: never clip, only hide out of range markers)
`marker.overflow: true/false` (`true`: allow markers to overflow the subplot bounds)

etpinard:
> Right now the layering depends on cliponaxis - I think we need to preserve the previous behavior, which probably means keeping the layer structure as it was but pushing the clipping down to the level of point groups and line groups.

Yeah, I thought about that, but didn't think it would be a big deal. Guess I was wrong. 

> Since this applies only to markers I'd put it in the marker container,

I guess you didn't see the part about text nodes and error bars.

etpinard:
> Also: we should recalculate visible markers during zoom/pan (as well as during animation, which you already mentioned), so we don't get situations like these (dragging the top right corner out and then in):

That too, I didn't think this would be a big deal. I suspect 99% of the use cases for `cliponaxis: false` would be for image export.

geocosmite:
I'm glad that you guys are doing such a thorough job in considering the edge cases.  

By the way, for scientific plots my intuition would be the opposite of yours:  99% of the use cases will not want the symbols to be truncated by the axis boundaries irrespective of the zoom.   A survey of our users (all scientists admittedly) found that not a single one of them could come up with a use case where symbol truncation was a desirable feature.  Based on this response our current plan is hardwire `cliponaxis: false`.

alexcjohnson:
> > Since this applies only to markers I'd put it in the marker container,
>
> I guess you didn't see the part about text nodes and error bars.

Ah right, sorry. Text should work this way for sure, so the attribute doesn't belong in `marker`.

I'm not sure about error bars though (apologies for not noticing the [comments on this](https://github.com/plotly/plotly.js/issues/1385#issuecomment-310166352) last week) - my gut reaction is to *not* include them outside the clip area, since (unlike markers and text) their ends correspond to actual axis values, so if the axis value at the end of an error bar is not displayed, then the bar should not be displayed either. I suppose you could argue that the *caps* should not be clipped, in the direction perpendicular to the errors, but that's adding a whole other layer of complexity for what seems like little benefit.

etpinard:
> 99% of the use cases will not want the symbols to be truncated by the axis boundaries irrespective of the zoom.

That's not what I meant. I was referring to interacting by panning or via the zoom box on a `cliponaxis: false` graph isn't much of a gain compared to the current behavior.

etpinard:
Closing this thing. :rage:

PR #1829, opened 2017-06-28
Accept empty transforms

rreusser:
`transforms: [{}]` causes plotly to fail with:

screen shot 2017-06-28 at 11 19 27

See: https://codepen.io/rsreusser/pen/GEyWav

I could have it compress empty transforms out of the picture (`[{}] --> []`, but because of indexing (it should work but I'm too worried it'll cause unexpected failures elsewhere), I've elected to just leave `[{}]` --> `[{}]`. This PR prevents it from throwing an error.


etpinard:
Looks good. :dancer:

PR #448, opened 2016-04-19
Fix issue 359 (zoom overlay drawn beneath shapes)

n-riesco:
This PR comes from #395 . I couldn't rebase it, and I ended up cherry-picking the commits.

I followed @alexcjohnson 's advice and created a zoom layer between the info and the hover layers.

The PR is ready for review.

PR #390, opened 2016-04-03
Ternary phase diagrams

alexcjohnson:
### In brief

Add ternary subplots, plus a `scatterternary` type to put on them.
Also a bunch of refactoring to make this (and future) integrations easier:

- collapse `axisid`, `sceneid`, `geoid` into a single `subplotid` attribute type. Ternary subplots are managed by a `subplot` attribute and the next step will be to use this for 3d and geo too rather than attributes called `scene` and `geo`. Potentially also for cartesian, ie `subplot='x2y2'` - but I'm not going to change those ones in this PR.

- provide `axis.color` as an attribute that sets the default for all the other colors associated with the axis. The one kind of tricky part of this is `gridcolor`, which has to be lightened from `axis.color` relative to the background color it will appear on.

- break up `graph_interact` some more, creating a new `dragElement` component, a new cartesian `dragBox` module, and a new `setCursor` function in lib

- unwind all the ugly conditionals in `Titles.draw`, instead passing a configuration object from each caller.

I still need to make some test plots, but I think this is ready for review.

### Limitations

- Only the `scatterternary` trace trypes. `contourternary` will be in a future iteration
- Fills (i.e scatter attributes `'fill'` and `'fillcolor'`) aren't supported at the moment

cc @etpinard @mdtusz @jackparmer

PR #1831, opened 2017-06-28
Fix scatterternary lasso/select drag modes

etpinard:
fixes issue-less bug discovered while working on `cliponaxis: false`

In brief, select/lasso drag modes are currently 100% broken on ternary subplots (e.g http://rickyreusser.com/plotly-mock-viewer/#ternary_markers) since https://github.com/plotly/plotly.js/pull/448 (merged more than 1 year ago :upside_down_face: ). Note that `scattertenary` traces supported select/lasso in their first implementation in https://github.com/plotly/plotly.js/pull/390.

cc @alexcjohnson

### DIFF
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -23,7 +23,7 @@ var MINSELECT = constants.MINSELECT;
 function getAxId(ax) { return ax._id; }
 
 module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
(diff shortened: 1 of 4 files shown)


alexcjohnson:
Ha nice find! Can't decide if I should be relieved or sad that nobody reported this in the year it's been broken... but great fix and beautiful tests (with a bonus type `scattercarpet` no less!)
💃

Issue #1825, opened 2017-06-28
heatmapgl cursor popup can't go wider than the plot is tall


ssfrr:
I use heatmaps to plot spectrograms represented as matrices. In trying to improve performance I tried using `heatmapgl`, but it looks like there's an issue with the cursor popover where it stops following the cursor when the X coordinate is greater than 1023.

Example:
https://jsfiddle.net/xfj1yxyu/3/

EDIT: corrected fiddle URL

etpinard:
cc @dfcreative

ssfrr:
just realized that 1023 isn't special, but that's the height of my plot (well, 1025, but that's close enough). When the plot is taller than it is wide things seem to be OK.

A smaller (e.g. 50x100) `heatmapgl` shows the same behavior.

PR #1833, opened 2017-06-28
Fix order of heatmap z-data

dfcreative:
Should fix https://github.com/plotly/plotly.js/issues/1825

If I get it right, we have rows layout `z: [[x,x,x,x], [x,x,x,x], ...]`, meaning that to retrieve a value we call `z[y][x]`, not `z[x][y]`.

### DIFF
--- a/src/components/fx/helpers.js
+++ b/src/components/fx/helpers.js
@@ -105,11 +105,11 @@ exports.appendArrayPointValue = function(pointData, trace, pointNumber) {
         if(astr === 'ids') key = 'id';
         else if(astr === 'locations') key = 'location';
         else key = astr;
-
         if(pointData[key] === undefined) {
             var val = Lib.nestedProperty(trace, astr).get();
+
             pointData[key] = Array.isArray(pointNumber) ?
-                val[pointNumber[0]][pointNumber[1]] :
+                val[pointNumber[1]][pointNumber[0]] :
                 val[pointNumber];
         }
     }


etpinard:
Hmm. `pointNumber` should give the `z[pointNumber[0]][pointNumber[1]]` value.

We'll have to check with if this is consistent with non-gl `heatmap` and `contour`.

dfcreative:
@etpinard yes, we do `z[y][x]` in regular heatmap: https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/hover.js#L94

etpinard:
but on

https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/hover.js#L38

we set

```js
ny = index[1]
nx = index[0]
```

dfcreative:
Not sure where do we use [`pointData.index !== false`](https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/hover.js#L36) branch at all, because in every case I can spot we calculate `nx, ny` [here](https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/hover.js#L70), which is basically snapping proper `x`/`y` value to bins, which corresponds to `z[y][x]` data layout and the way data is plotted.

etpinard:
See http://rickyreusser.com/plotly-mock-viewer/#zsmooth_methods

dfcreative:
@etpinard yes, this cases are for regular `heatmap`, switching them to `webgl` raises the error covered here.

alexcjohnson:
> but on
>
> https://github.com/plotly/plotly.js/blob/master/src/traces/heatmap/hover.js#L38
>
> we set
>
> ```js
> ny = index[1]
> nx = index[0]
> ```

@etpinard looks consistent to me?
```
nx = Math.round(pointData.index[1]);
ny = Math.round(pointData.index[0]);
```

Anyway, the convention is meant to match the way a matrix is printed:
```
[
  [1, 2, 3],
  [4, 5, 6],
  [7, 8, 9]
]
```
so the first index is row (y) and the second index is column (x)

@dfcreative the `pointData.index !== false` case may be for manually triggering hover - ala https://plot.ly/javascript/hover-events/#triggering-hover-events (note you have to scroll the iframe to see the "go" button)

How about covering the relevant cases with tests, then we don't have to argue about this!

etpinard:
It would be nice to fix this thing in `1.29.0`. I'll try wrapping this tomorrow

etpinard:
Supersedes by https://github.com/plotly/plotly.js/pull/1884

Issue #991, opened 2016-09-29
Provide a data table view usable via JavaScript


sheikirfan:

etpinard:
It is possible. A plotly.py table is simply a regular `heatmap` trace with judiciously placed text annotations. But, as you pointed out, we don't expose any _figure factories_ in plotly.js.

That said, with the new plotly.js [transform](https://github.com/plotly/plotly.js/pull/499) plugin functionality, we would pretty easy to convert the plotly.py [logic](https://github.com/plotly/plotly.py/blob/e8dc76e1175cbe5a2d143b9fad8394305d54e1cd/plotly/tools.py#L6123-L6194) in JS as a transform module.

etpinard:
Anyone interested in doing this may want to look at https://github.com/plotly/plotly.js/pull/1020 first for inspiration.

monfera:
# Description

There's a need for tables in the context of crossfiltering. This ticket is repurposed to 

- clarify and confirm needs for tables
- consider pros/cons of approaches (e.g. as a new plot/trace type in `plotly.js`, or an abstraction that uses existing facilities e.g. `heatmap` and `transform`)
- decide on direction and finalize initial requirements ([Minimum Viable Cathedral](https://twitter.com/worrydream/status/878471652362731524))

## Context

We're working on crossfiltering functionality. On an initial glance, it's just a way of letting widgets constrain dimensions and make the other widgets rerender. But there's a lot more to it - to be described elsewhere, except for the fact that showing tables is common on crossfiltered views.

- Mike Bostock's [example](http://crossfilter.github.io/crossfilter/) with his `crossfilter.js` package 
- [some dashboards](http://climateproxiesfinder.ipsl.fr/) made with [dc.js](https://github.com/dc-js/dc.js/wiki#community-examples)
- stated user needs

The need that a table work well with crossfiltering already poses some demands, and there are general requirements toward a table widget as well.

## Table functionality needed for, or implied by crossfiltering

There are some direct, often obvious needs, such as:

- take tabular (multidimensional) data and render them with dimensions in one direction (usually, columns) and samples, or records, in another (usually, rows)
- support styling options, e.g. row pitch
- have a clear approach for overflow (e.g. scrollable table)
- rerender efficiently if the set of records change

[...]

PR #1834, opened 2017-06-29
[WIP] Table view

monfera:
As per discussion at #991 

Most functions, attributes etc. are not working and not final.

Printing doesn't work as, apparently, nested `font` specs in an array (`dimensions`) balk when printing.

Test cases aren't added to this branch yet.

Shows a variety of column formats (font options, numeric options) and column rearrangement.

initial table


monfera:
Quick memo of our chat : the `dimensions` approach has some impedance mismatch with `workspace` so on a proposal by @etpinard we'll follow a `heatmap` `"z"` [type approach](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/basic_heatmap.json).

To allow per cell styling, the attributes for left vs center vs right alignment, font styling, number formatting, background color etc. will admit either a singular value (e.g. the entire table uses the same font) or an array that follows the [same nested structure]([type approach](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/basic_heatmap.json)) eg.

```
align: [['left','right','center'], ['right','left','center'],['center','right','left']]
```

@etpinard specified these options for future consideration:

```
align: 'left' // applies to all cells
align: ['left', 'right', 'center'] // applies per row
align: [['left'],['right'], ['center']] applies to columns
align: [['left','right','center'], ['right','left','center'],['center','right','left']] // per-cell options
```

monfera:
image

monfera:
The mock that produces the above circus-themed cell smorgasbord:
```
{
    "layout": {
        "width": 1184,
        "height": 400,
        "title": "Widget parameters and cost",
        "font": {
            "color": "blue",
            "size": 2,
            "family": "sans-serif"
        },
        "margin": {"t": 106, "r": 60, "b": 54, "l": 60}
    },

    "data": [{

        "type": "table",

        "domain": {
            "x": [0.05, 0.95],
            "y": [0.05, 0.85]
        },

        "width": [110, 60, 90, 90, 80, 70, 80, 60, 100, 70],
        "height": 30,


        "align": ["right", "center", "center", "right", "right", "right", "left", "right"],
        "valign": ["top", "bottom", "center"],

        "prefix": ["", "", "", "", "U<br>S<br> ", ""],
        "suffix": ["m", ""],

        "font": {
            "color": [
                ["rgb(214,235,14)", "rgb(24,251,11)", "rgb(109,103,67)", "rgb(152,192,23)", "rgb(80,207,160)", "rgb(84,134,44)", "rgb(222,105,214)", "rgb(76,181,63)", "rgb(110,173,82)", "rgb(144,56,245)", "rgb(48,196,142)", "rgb(190,102,179)", "rgb(115,83,61)", "rgb(54,231,74)", "rgb(235,105,39)", "rgb(1,128,246)", "rgb(205,58,176)", "rgb(34,128,113)", "rgb(222,81,3)", "rgb(18,252,69)"],
                ["rgb(107,128,223)", "rgb(95,62,207)", "rgb(110,65,159)", "rgb(17,249,0)", "rgb(184,189,240)", "rgb(234,55,166)", "rgb(254,75,119)", "rgb(215,70,114)", "rgb(102,160,80)", "rgb(7,190,174)", "rgb(177,25,224)", "rgb(237,181,164)", "rgb(122,240,243)", "rgb(54,245,219)", "rgb(65,30,43)", "rgb(67,161,153)", "rgb(77,77,169)", "rgb(11,144,192)", "rgb(27,193,182)", "rgb(176,134,145)"],
                ["rgb(214,204,7)", "rgb(228,251,163)", "rgb(22,8,80)", "rgb(129,125,180)", "rgb(229,211,97)", "rgb(14,77,169)", "rgb(31,68,23)", "rgb(219,106,81)", "rgb(113,123,241)", "rgb(113,228,139)", "rgb(115,124,86)", "rgb(185,227,179)", "rgb(126,218,33)", "rgb(83,39,227)", "rgb(43,40,158)", "rgb(244,77,12)", "rgb(19,15,168)", "rgb(212,66,245)", "rgb(175,76,103)", "rgb(250,61,197)"],
                "red",
                ["magenta", "blue"],
                ["rgb(47,249,84)", "rgb(186,219,118)", "rgb(116,107,204)", "rgb(255,6,79)", "rgb(157,34,21)", "rgb(90,61,107)", "rgb(12,195,215)", "rgb(16,22,219)", "rgb(111,207,137)", "rgb(106,15,130)", "rgb(156,131,30)", "rgb(178,28,31)", "rgb(90,192,19)", "rgb(188,146,244)", "rgb(83,1,104)", "rgb(74,206,15)", "rgb(200,23,125)", "rgb(94,68,55)", "rgb(25,43,18)", "rgb(72,79,254)"]
            ],
            "size": 12,
            "family": "sans-serif"
        },

        "line": {
            "color": [
                ["grey"]
            ],
            "width": 0.5
        },

        "fill": {
            "color": [
                ["#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9","#e0f3f8", "#abd9e9"]
            ]
        },

        "labelfont": {
            "family": "PT Sans Narrow",
            "size": 16,
            "weight": "bold",
            "color": "grey"
        },
        "values": [
            [32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666, 32000, 162666],
            [268630, 489543, 379086, 600000, 489543, 268630, 600000, 379086, 268630, 489543, 379086, 600000, 489543, 268630, 600000, 379086, 268630, 489543, 379086, 600000],
            ["A", "A", "B", "AB", "Y", "Z", "Z", "A", "B", "B", "Y", "Y", "Z", "A", "A", "B", "Y", "Y", "Z", "Z"],
            [160, 1324, 252, 1711, 173, 624, 228, 2474, 371, 3103, 312, 2408, 460, 1727, 754, 2457, 227, 1963, 354, 2797],
            [1125, 9274, 2735, 16920, 2193, 20704, 4814, 34689, 5565, 46554, 5225, 38887, 9746, 37715, 17798, 55250, 5697, 49083, 11226, 81939],
            [9, 794, 34, 1409, 17, 536, 39, 7133, 115, 9843, 107, 7017, 273, 5131, 1210, 9032, 82, 7312, 252, 16778],
            [0.2366, 0.3269, 0.3471, 0.4373, 0.4575, 0.106, 0.568, 0.2164, 0.2366, 0.3269, 0.3471, 0.4373, 0.4575, 0.106, 0.568, 0.2164, 0.2366, 0.3269, 0.3471, 0.4373],
            [5137, 113712, -5909, 102666, -16955, 135803, -28000, 124758, 5137, 113712, -5909, 102666, -16955, 135803, -28000, 124758, 5137, 113712, -5909, 102666],
            [98453, 319366, 234498, 455411, 391333, 147711, 479081, 208910, 124041, 344954, 280876, 501789, 368624, 98453, 429823, 234498, 170419, 391333, 258167, 479081],
            [1417, 23377, 3376, 33336, 5635, 10812, 6898, 21273, 2484, 35126, 5626, 51096, 7384, 22020, 18912, 52449, 7498, 87528, 11359, 107154]
        ],
        "labels": [
            "Height",
            "Width",
            "Cylinder",
            "Weight",
            "Cost",
            "Penalty",
            "H/W",
            "Min H/W",
            "Diameter",
            "RF block"
        ],
        "valueformat": [
            ".2s",
            "#0x",
            null,
            ",.2f",
            "$,",
            "_>5",
            ",.1%",
            null,
            null,
            null
        ]
    }]
}

```

monfera:
I'm leaning toward delete the colorbar. The original idea was that, if the user chooses to color cells by some value, then the colorbar would be handy. But the issue is that there's a combinatorial explosion of possible wishes, eg. the user would use coloring for one column, or multiple columns of the same unit / measure / etc. (comparables); but perhaps it would be rows, or a rectangular area in the table. Heterogeneous coloring can't be done because a colorbar supports one scale only. 

Yet it's simple to achieve the desired effect, because as with the other cell styling options, a per-row, per-column or per-cell color can be directly specified.

If the decision is to retain the colorbar, I could use guidance on how exactly it should behave and how it should interact with possible cell coloring specification that might be in conflict.

jackparmer:
> I'm leaning toward deleting the colorbar.

Yes, definitely. The Python table figure factory doesn't have a colorbar either.

jackparmer:
I agree with the ✅ 's in the MBP section. It seems like those will get us to parity with the [Python table figure factory](https://plot.ly/python/table/), which we already know is a success. 

We can let future features be driven by user and customer demand.

> scrolling or pagination (or and?)

Offering one _or_ the other seems more than sufficient. I've seen table pagination with DataTables a fair amount in Shiny apps:
https://datatables.net/examples/basic_init/alt_pagination.html

monfera:
@etpinard have you got a preference for styling the header row in the JSON spec? Some options:

1. There is no concept of a header row; users need to use the 1st row (`values` and styling) as a header row, with whatever formatting they need (formatting can be per row, column and cell for the table)
2. There is a separate spec in the JSON for the header row; its substructure is the same as the substructure of the table, i.e. the values/style spec is duplicated like:
```
{ 
  "width: [50, 30, ...], // width has to be shared so it's at the top level
  "header": {
    "values": ["Horse power", "Volume", ...],
    "font": {"color": ["red", "blue", ...]},
    "fill": {...}
    ...
   },
  "cells": {
    "values": [[110,202,52,...], [23,45,9,...], ...],
    "font": {"color": ["black", "grey", ...]},
    "fill": {...}
   },
  ...
}

**Comment by alexcjohnson** [comment]
Lets go with (2) - separate spec for the header row, and yes, single-row headers. This is a very common use case so shouldn't require turning all the styles into arrays. Also fits in with the analogous heatmap case - in which axis labels are specified independently from the data.

**Comment by etpinard** [comment]
Yeah (2) sounds good. I'm not a fan of the root-level `width` attribute name as it might be confused for `layout.width`, but I can't think of a better name and `annotations` already have a `width` attribute, so :+1:

**Comment by monfera** [comment]
`columnWidth` would work if you prefer that

**Comment by alexcjohnson** [comment]
I like `columnWidth`, good idea! Makes it clear that it's not the overall table width, and can be an array.

**Comment by etpinard** [comment]
> I like columnWidth

`columwidth` please in this case. plotly.js doesn't like :camel: cases in attributes.

**Comment by monfera** [comment]
Non-garish table example: 

image

**Comment by alexcjohnson** [comment]
> heterogeneous row heights (needed if text is wrapped); maybe option for keeping rows of homogeneous height? e.g. only one row needs to increase its height to fit multiple rows; should the other rows increase in height?

I don't think we need an option for this (if I'm understanding you right) - this should work just like regular html tables, ie all rows get the same height, which is enough to fit the tallest contents. Otherwise items in a row lose connection to each other.

> breaking text to multiple rows at spaces; breaking too long words nevertheless if space based breaking is not enough; perhaps truncation (with or without an ellipsis) as a boolean option?

> links (if needed) and perhaps other content types?

Lets use `svgTextUtils.convertToTspans` so we get all the pseudo-html features. And perhaps we can work wrapping into that routine so it can be used elsewhere too? 👼  Truncation as an option is a good idea, but I'd think for tables wrapping should be the default.

File: test/image/mocks/table.json [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+                ["Width"],
+                ["Cylinder"],
+                ["Weight"],
+                ["Cost"],
+                ["Penalty"],
+                ["H/W"],
+                ["Min H/W"],
+                ["Diameter"],
+                ["RF block"]
+            ],
```
alexcjohnson: Does `header.values` need to be 2D, or is 1D sufficient?

monfera: Nope, it handles a 1D array just fine, I'll need to update the mock, thanks for the reminder!

monfera:
Thanks again for your work! We just had a chat and while the ellipsis approach would solve this case and indeed it's a very compact diff, the other plotly.js plots do something different, eg. the Cartesian plots rotate the labels if there's not enough room. We were planning to add it at some point but it's a good bit of work as rotation of the long texts needs to result in a vertically and horizontally smaller plot (given the same overall layout.width and layout.height) so as the long, tilted labels fit. We were thinking of this text rotation for multiple reasons:

- the ellipsis still leaves part of the text unreadable while rotation can show the full text
- the ellipsis effect can be approximated by the developer setting a character limit and truncating text (yes it's approximate only)
- there can be up to 60 dimensions, where the distance between axes is likely very small
- the iterative text shortening is slightly expensive by its nature

The ellipsis would be an improvement on the current situation but maybe a cheaper way to achieve almost the same thing (without the ellipsis) is the use of `<textPath>` where an invisible `path` is in the DOM and referenced by the `<textPath>` element, with the result that the text is rendered only till it reaches the end of the line.

@etpinard what do you think, would a simple text truncation be a suggested improvement? It needs no iteration.

monfera:
@alexcjohnson thanks for your feedback, I'll look into the text wrapping! btw. when I wrote 'heterogeneous row heights' I referred to the possibility that some rows (the entire row ofc.) would be taller than others to wrap long contents into multiple lines - yes an alternative is to make all rows of equal height (the max). Given that we'll have pagination, and the table may have a lot of rows, probably heterogeneous row heights would provide a better user experience as a single outlier long line would make the entire table look weird, pls. let me know if you see otherwise. (I wasn't shooting for broken lines ie. the cells of a specific line would be of equal height, as per the max wrap count in the row).

An aside is that HTML tables are rather sophisticated for text wrapping and I'm not currently shooting for feature parity with it. For example, if left unspecified, HTML tables will adjust the table column widths so as to maximize the screen use. For example, a column with more / longer content would end up being wider. I doubt the HTML table does an actual MILP optimization but at the simplest, it may use some heuristic eg. total text length, which would of course be crude, as non-proportional fonts and non-uniformly breakable text (dependent on word length, word count  etc.) make it more complex than if text were like sand or water :water

alexcjohnson:
> Given that we'll have pagination, and the table may have a lot of rows, probably heterogeneous row heights would provide a better user experience as a single outlier long line would make the entire table look weird, pls. let me know if you see otherwise. (I wasn't shooting for broken lines ie. the cells of a specific line would be of equal height, as per the max wrap count in the row).

Ah great, I thought I must have misunderstood you, thanks for clarifying! Yes, I think you're correct that each row should size to its own contents, not the tallest of all contents. I wouldn't bother making that an option, at least not initially. Raises a question about how pagination should work with these variable-length rows, do you go for constant number of rows or (as close as possible) constant total height? If the latter, there will still be a possibility that a single row is bigger than the allotted size. To me that kind of consideration kind of argues for scrolling instead of pagination - and personally I think I kind of prefer scrolling anyway, but it's probably a good deal harder to implement in a performant way.

> An aside is that HTML tables are rather sophisticated for text wrapping and I'm not currently shooting for feature parity with it.

Way too sophisticated, I've always been confused by the decisions HTML layout engines make there. Simple and explicit is way better for our purposes.

monfera:
@alexcjohnson thinking about the latter, because otherwise the page height will be overly variable, and the pagination buttons must be kept at place, i.e. there would be too much variance in the gap above the pagination buttons, and a large gap sends the misleading message: I'm the last page!

As the user can jump to arbitrary pages, the entire table has to be wrapped upfront... so we're kinda reimplementing the HTML table functionality... scrolling is the same in this regard (all wrapping has to be precomputed). Yet I'm partial for pagination, at least in the 1st version, for some reasons (just a sampler):

- limited, bounded DOM element count; the mere element count is a big factor in interaction performance, even if we don't `translate` stuff after the initial render
- not straying from plotly.js patterns (I don't think we have scrolling widgets)
- column reordering is much simpler (the header would need to stay in place while normal rows would scroll past)

So as you can see, wrapping, while it sounds innocent as a _feature request_, it triggers a whole lot of design and implementation choices and we'll learn of most of them during the work :-)

alexcjohnson:
I agree that pagination will likely be easier to implement - and I'm not particularly strongly opinionated about which we go with - but to your points:

> page height will be overly variable, and the pagination buttons must be kept at place, i.e. there would be too much variance in the gap above the pagination buttons, and a large gap sends the misleading message: I'm the last page!

This will be an issue - at least in edge cases - no matter what for a paginated table. But scrolling tables can just show partial rows and fill whatever space they are allotted (unless the *whole* table is smaller than the given space, which is also fine).

> limited, bounded DOM element count; the mere element count is a big factor in interaction performance, even if we don't translate stuff after the initial render

In principle I would imagine a *container* big enough for all the rows, but then only *rendering* the visible ones... so then the element count would be only marginally larger for scrolling than for pagination, though the frequency of changing which ones are rendered would be much larger.

> not straying from plotly.js patterns (I don't think we have scrolling widgets)

We do have scrolling [legends](https://github.com/plotly/plotly.js/pull/243) and [dropdowns](https://github.com/plotly/plotly.js/pull/1214)...

> column reordering is much simpler (the header would need to stay in place while normal rows would scroll past)

Not sure I follow... I guess with pagination we could make each column (header included) a group, and move them together, whereas with scrolling it would be easier to make the headers inside one container and the data inside another container, so there would be two things to move in sync whenever a column moves? So pagination does seem simpler but I wouldn't characterize it as *much* simpler.

> wrapping, while it sounds innocent as a _feature request_

Yah for sure, all sorts of issues around wrapping. Just to be clear though, it's not just automatic wrapping that triggers this, but also `<br>` in the text if we support pseudo-html, and also font size if that's allowed to be variable.

monfera:
btw. if scrolling were needed, we'd probably need to do an infinite revolver to keep element bounded, e.g. https://bl.ocks.org/monfera/8b87a9aade2214fc15ea76bdf7abe11a

We'd need to do something similar because in my experience a long SVG surface (multiple k pixels tall or wide) will be slow even if only the visible part is rendered, so basically we'd just mimic the scrolling like in the above example.

Thanks for the clarifications on the other points! Brw if we want to initially go with a scroller rather than a paginator pls tell me today so I can pivot early tomorrow my time.

alexcjohnson:
Ah, the infinite revolver is a cool idea, with the two panes trading order!

Dunno if @jackparmer or @etpinard have other opinions, but from my perspective scrolling is worth it if:
- It can work just like we have it for legends, for example the `legend_scroll` mock: wheel scroll with the same inertia as regular page scrolls, and a permanently visible drag handle (though I notice the legend scroll handle doesn't have a size proportional to the visible fraction, we should fix that).
- It's not going to add more than a couple of days work vs. pagination

monfera:
I'll do a morning sprint on a crude scroller just to see how it behaves performance and jankiness wise.

etpinard:
:+1: for scrolling 

:+1: :+1: for wrapping our current updatemenu/legend scrolling logic into a reusable component (I can help with that)

monfera:
thanks for the feedback, my morning tests look good so far, a scroller is in the making, though, due to the unbounded length of the table compared to the low DOM element count performance starts to suffer, it's an _infinite revolving scroller_ (see link/comments above) now inside the `table` renderer - it uses two panels and each panel holds at least as many rows as what may be visible at a time, given the view height of the (sub)plot

alexcjohnson:
> 👍 👍 for wrapping our current updatemenu/legend scrolling logic into a reusable component (I can help with that)

I doubt it will make sense to combine that with table scrolling - since the scale is very different, legends and updatemenus don't need the dynamic rendering that will be crucial to tables. Anyway for the scrollbar handle specifically I made an issue #1859 - if we want to roll that into componentizing scroll, that would be great, I'm sure it will see additional uses in the future!

monfera:
Besides the need for an infinity revolver,
scroller
column rearrangement still needs to work, i.e. it's not just two large panels doing the rotation, it must be split to columns.

iainmwallace:
It would be very useful if the table had the option to filter columns. This is probably the feature I like most in the Shiny DT package.

monfera:
@iainmwallace filtering is indeed useful and I'm sure it'll be present at some point. To expedite things, the initial version will likely skip it. But it'll be possible to do filtering at the data level, ie. the library user just re-issues the data that contains only the kept rows (users may roll their own input fields and custom logic). There are so ways of doing filtering:

- type value in text field and do simple match
- type value and allow regex or at least *? matches
- select a value in the table to filter for only those rows where the value is the same
- from-to range selection
- combined selections

also, so many affordances for interaction:

- control filtering from outside the table view
- separate text field for entry
- radio buttons, check boxes for options
- modal dialog
- separate UI elements on the side

What may be sufficient and obviously needed functionality for some people may be inadequate or overly broad for others, so any input is appreciated, eg. what specific options you have in mind.

iainmwallace:
Hi Robert,

In my experience with Shiny the filtering options provided the non
technical users of my dashboard a great experience. The UX from my
perspective was also great as it was just two lines of code to display an
intuitive table. Displaying the contents of a csv/excel file is a very,
very common pattern.

There filtering modes in the DT package (
https://rstudio.github.io/DT/shiny.html) depend on the type in the column
- dates/numbers have a range selector
- strings filter based on a drop down selection.

Cheers,

Iain

monfera:
Had to debug the revolver panels, why not document visually how it works. There are two panels, a blue and a red one; each is just a few pixels taller than what's visible vertically at any given time (the few pixels extra tallness is due to the need for a panel to cover an integral number of columns). Once a panel is off the viewport, it jumps to the other end.

revolver

This is how it looks with regular styling:

revolver2

monfera:
... the panels are also visible if `constants.clipView` is set to `true`. This is a mode which, if set, doesn't actually clip, instead, just shows the clip bounds. The end result is a plain scroller, but the clipping is happening at two levels, one is per 'column block' (ie. one column within a revolver panel) and one is for the scrolling viewport. The reason is that, on column drag, the column lifts `constants.uplift` pixels to give visual feedback, Apple Numbers style. Also, the columns are dragged individually. The yellow clip box is wider to allow overdrag of columns by `constant.overdrag` pixels.

ciip

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+                arrayOk: true,
+                role: 'style',
+                description: [
+                    'Sets the cell fill color. It accepts either a specific color',
+                    ' or an array of colors.'
+                ].join('')
+            }
+        },
+
+        font: {
```
alexcjohnson: This is just `fontAttrs` with `arrayOk`, right? 🌴  https://github.com/plotly/plotly.js/blob/master/src/traces/bar/attributes.js#L20-L23

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+    labelfont: extendFlat({}, fontAttrs, {
+        description: 'Sets the font for the `dimension` labels.'
+    }),
+
+    labels: {
+        valType: 'data_array',
+        role: 'info',
+        dflt: [],
+        description: 'The shown name of the columns.'
+    },
```
alexcjohnson: I feel like I asked this already but I don't see it... `labels`/`labelfont` are gone, right? replaced by `header`?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+    header: {
+
+        values: {
+            valType: 'data_array',
+            role: 'info',
+            dflt: [],
+            description: [
+                'Dimension values. `values[n]` represents the value of the `n`th point in the dataset,',
+                'therefore the `values` vector for all dimensions must be the same (longer vectors',
+                'will be truncated). Each value must be a finite number.'
```
alexcjohnson: this is the same description as in `cells.values` - can you make one specific to `header.values` that describes the options for dimensionality? And as to the other `arrayOk` attributes in `header`, would they need to match that dimensionality, or can they be 1D if `values` is 2D?

alexcjohnson: Also other `data_array` items we typically have not provided a `dflt` at all - is it needed here?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+        },
+
+        format: {
+            valType: 'data_array',
+            role: 'info',
+            dflt: [],
+            description: [
+                'Sets the cell value formatting rule using d3 formatting mini-language',
+                'which is similar to those of Python. See',
+                'https://github.com/d3/d3-format/blob/master/README.md#locale_format'
```
alexcjohnson: headers will mostly be strings, not numbers... does this just get ignored in that case? Actually same goes for `cells.format`. Which begs the question of dates too - `axis.tickformat` supports [either number formatting or date formatting](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/layout_attributes.js#L441-L448) depending on the axis type - do we need a `type` field for headers and cells as well?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+            valType: 'string',
+            arrayOk: true,
+            dflt: null,
+            role: 'style',
+            description: 'Suffix for cell values.'
+        },
+
+        height: {
+            valType: 'number',
+            arrayOk: false,
```
alexcjohnson: I understand `arrayOk: false` for `cells.height`, but I could imagine it being useful for headers.

BTW @etpinard re: #1894 ALL of these `headers` arrays would need to be excluded if we ever call `findArrayAttributes` on tables.

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+    cells: {
+
+        values: {
+            valType: 'data_array',
+            role: 'info',
+            dflt: [],
+            description: [
+                'Dimension values. `values[n]` represents the value of the `n`th point in the dataset,',
+                'therefore the `values` vector for all dimensions must be the same (longer vectors',
+                'will be truncated). Each value must be a finite number.'
```
alexcjohnson: not just numbers...

test/image/mocks/table.json
```diff
@@ -0,0 +1,118 @@
...
+                "color": "white"
+            }
+        },
+
+        "cells": {
+
+            "values": [
+                [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19],
+
+                                ["A", "A", "B", "AB", "Y", "Z", "Z", "A", "B", "B", "Y", "Y", "Z", "A", "A", "B", "Y", "Y", "Z", "Z"].slice(0, 20),
```
alexcjohnson: `.slice` in JSON?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,283 @@
...
+
+        format: {
+            valType: 'data_array',
+            role: 'info',
+            dflt: [],
+            description: [
+                'Sets the cell value formatting rule using d3 formatting mini-language',
+                'which is similar to those of Python. See',
+                'https://github.com/d3/d3-format/blob/master/README.md#locale_format'
+            ]
```
alexcjohnson: remember `.join(' ')` (and in `cells.format`) - took me a while to figure out why I couldn't get the test dashboard to start, kept getting the error:
```
"SyntaxError: Unexpected token (99:9) while parsing file:
/Users/alex/plotly/plotlyjs/src/traces/table/attributes.js"
```
which I guess is coming from the browserify transform that strips out descriptions.

alexcjohnson: hmm, still not able to get it running, looks like a circular dep or something? I guess it's not ready for playing with, but I really wanted to see how the infinite scroller feels!

src/traces/table/attributes.js
monfera: Yes, I'll switch to `extendDeep`.

src/traces/table/attributes.js
monfera: Yes, thanks for the reminder, I left it in but served no function, removed (w/ the next squashed push)

src/traces/table/attributes.js
monfera: Indeed, I'll need to revise descriptions in general.

src/traces/table/attributes.js
monfera: In the implementation, most of the header cell and regular cell rendering is the same codepath (minor trivia: there are three `columnBlock`s, one for the header row (or header rows if we want to extend in the future), and one for each of the revolving scroller panels. So I thought without much thinking that the formatting options would be identical. Indeed numeric headers are rare, though date headers may be more common. I'll need to implement proper date formatting as you suggest.

src/traces/table/attributes.js
monfera: ... also, the description should explain how it works if a 1D array is specified, or a mixed 1D / 2D array (ie. an array whose elements include arrays as well as scalars), and also the truncation - if the array is shorter than the row / column count, then the last value is used, making it possible to easily format by column, by row and mixed

src/traces/table/attributes.js
monfera: I wasn't planning with heterogeneous cell heights for the header row for the time being, as it's a single-row header at least for the MVP. Cell merging would be useful for multiline headers in the future. Maybe I'm missing a short term use case of heterogeneous cell heights with the single-row header cells, so a link or drawing might help. It'd complicate things not just in the implementation but also in the design, e.g. the top would look jaggy and the uplift that happens when you start dragging a column would become ineffective.

src/traces/table/attributes.js
monfera: Correct, and it's not dimensions, I'll revise the descriptions.

test/image/mocks/table.json
monfera: they're just there temporarily for quick tests, linter balks too

src/traces/table/attributes.js
monfera: I had two missing `.join(' ')`s due to omitting it in the copy/paste selection. It causes a problem with `npm run build` which I use less often than `budo` but had to fix it already for the crossfilter example (now relies on a dist js file), fix will be in the squashed commit.

src/traces/table/attributes.js
alexcjohnson: Oh, I didn't realize headers were limited to 1 row (saw the 2D `values` in [table_large](https://github.com/plotly/plotly.js/pull/1834/files#diff-438295fda1d4e0516bdc35244aaef4efR28) - still only one row but the structure is obviously there to support multiple rows) - that's fine, I was thinking `height`, if it were an array, would index the rows, not columns or individual cells, so at 1 row there's no reason this should be an array. Actually much more useful from my standpoint would be if headers (at least) would support pseudoHTML via `svgTextUtils.convertToTspans`.

etpinard: no need to set `arrayOk: false`, by default all attributes are `arrayOk: falsy`

src/traces/table/base_plot.js
```diff
@@ -0,0 +1,33 @@
...
+    if(calcData.length) tablePlot(gd, calcData);
+};
+
+exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout) {
+    var hadTable = (oldFullLayout._has && oldFullLayout._has('table'));
+    var hasTable = (newFullLayout._has && newFullLayout._has('table'));
+
+    if(hadTable && !hasTable) {
+        oldFullLayout._paperdiv.selectAll('.table').remove();
+        oldFullLayout._paperdiv.selectAll('.table').remove();
```
etpinard: duplicate?

src/traces/table/base_plot.js
```diff
@@ -0,0 +1,33 @@
...
+        oldFullLayout._glimages.selectAll('*').remove();
```
etpinard: Why?

src/traces/table/plot.js
```diff
@@ -0,0 +1,82 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var table = require('./render');
+
+module.exports = function plot(gd, cdTable) {
```
etpinard: @monfera I'm not a fan of your `plot.js` + `render.js` pattern. I know it makes `render.js` somewhat plotly-agnostic, but it also makes very painful to patch. You couldn't believe how much I was :rage:-ing while working on https://github.com/plotly/plotly.js/pull/1896

So, could you merge `render.js` into `plot.js` to make `table/` look a little more like the other traces?

src/traces/table/plot.js
```diff
@@ -0,0 +1,82 @@
...
+    var svg = fullLayout._paper;
+    var root = fullLayout._paperdiv;
+
+    var gdColumns = {};
+    var gdColumnsOriginalOrder = {};
+
+    var size = fullLayout._size;
+
+    cdTable.forEach(function(d, i) {
+        gdColumns[i] = gd.data[i].header.values.map(function(d) {return d[0];});
```
etpinard: Stuff like is perfect for `calc.js`. You could stash those `gdColumns` arrays in `gd.calcdata` and those will only get recomputed when array attributes are updated.

src/traces/table/render.js
```diff
@@ -0,0 +1,561 @@
...
+        width = layout.width,
+        labels = trace.labels,
+        columnWidths = trace.columnwidth;
+
+    var colCount = labels.length;
+
+    var groupWidth = Math.floor(width * (domain.x[1] - domain.x[0]));
+    var groupHeight = Math.floor(layout.height * (domain.y[1] - domain.y[0]));
+
+    columnWidths = trace.header.values.map(function(d, i) {
```
etpinard: more stuff for `calc.js`

src/traces/table/plot.js
```diff
@@ -0,0 +1,82 @@
...
+        gdColumnsOriginalOrder[i] = gdColumns[i].slice();
+    });
+
+    var hover = function(eventData) {
+        gd.emit('plotly_hover', eventData);
```
etpinard: Before merging, we should make sure `eventData` is on par with cartesian traces.

See https://github.com/plotly/plotly.js/pull/1896 and https://github.com/plotly/plotly.js/pull/1770

src/traces/table/render.js
```diff
@@ -0,0 +1,561 @@
...
+        return Array.isArray(columnWidths) ?
+            columnWidths[Math.min(i, columnWidths.length - 1)] :
+            isFinite(columnWidths) && columnWidths !== null ? columnWidths : 1;
+    });
+
+    var totalColumnWidths = columnWidths.reduce(function(p, n) {return p + n;}, 0);
+    columnWidths = columnWidths.map(function(d) {return d / totalColumnWidths * groupWidth;});
+
+    var pad = layout.margin || {l: 80, r: 80, t: 100, b: 80};
```
etpinard: you shouldn't need that fallback. `layout.margin` gets filled in during supply defaults.

monfera:
@chriddyp here are some quick notes on `table` vs alternatives:

Compared to the current `heatmap` based solution, though I'm not too familiar with it beyond the docs:

- scrolling (infinity scrolling; header row stays visible)
- column drag/drop (maybe linked to sorting in the future)
- text wrap, multirow text (not part of the 1st merge candidate)
- as it's dedicated, doesn't have the artifacts that `heatmap` has, by necessity
- arbitrary coloring of fields (not value based)
- more styling options eg. cell borders; also, per column, per row and/or per cell styling
- left/center/right and top/middle/bottom alignment (heatmap examples show left-aligned but numbers are best shown right-aligned)
- cell value formatting
- direct row height specification
- different column widths can be specified

Compared to a HTML table:
- fits in the rest of the plotly plots, ie. renders in SVG, current view (!) exportable as an image
- column drag&drop
- infinity scroller (admittedly, plain HTML scrolling is pretty good up to a significant row number but with SVG, we can't have too many DOM elements even if they're not individually `transform`ed during scrolling; also, SVG layers can't be too big, otherwise, performance suffers badly - so an infinity scroller was a necessity with SVG)
- easier crossfilter dashboard integration, as it's a plotly plot, not something external to plotly.js
- image export 
- formatting and styling options as in the above block

Compared to DataTable:
- integrates with plotly.js, just like any other plot type
- a standalone DataTable build is a 100kb, adding `table` to plotly.js is less than 10kB

src/traces/table/base_plot.js
monfera: thanks, now removed

src/traces/table/base_plot.js
monfera: thank you, leftover from `parcoords`, now removed

src/traces/table/plot.js
monfera: the latest push fixes it - it wasn't a simple file concat, the main dataflow has been plotlified, eg. referencing `calcdata` etc. and phasing out `model` and `viewModel`, hopefully you had something like this in mind

src/traces/table/plot.js
monfera: With the `render.js` / `plot.js` merge, these calculation steps have moved to `calc.js`.

I'd need a bit of help though. I [persisted stuff](https://github.com/plotly/plotly.js/blob/table/src/traces/table/plot.js#L260) on `gd.calcdata` as you suggested but upon a subsequent `Plotly.restyle` (of some unrelated attribute) it'll be overwritten as though I fully re-stated the `table.json`, so the information is lost. Is there a very simple example for showing how to persist things so they don't get lost on unrelated attribute change?

Somewhat related: to expose column ordering to the user, and also to let the users easily read the column order, this new attribute `columnorder` had been added.

Unrelated question: is there a way for `plot.js` or `calc.js` to know _what_ changed? For example, if the user called `Plotly.restyle` with `line.color` or some such, then a lot of the logic and rendering may not need to be redone.

src/traces/table/render.js
monfera: was also part of the refactoring

src/traces/table/render.js
monfera: thanks, leftover removed!

src/traces/table/attributes.js
monfera: it was done just for emphasis; now removed

monfera:
@alexcjohnson having spent some time with `svgTextUtils.convertToTspans`, it looks like it'd be best to defer (upshot on the bottom):

1. Main issue: it's a large function and expensive call that needs to be done by individual value (here: cell) while `table` just does a straight SVG text render, and with the data binding pattern (all cells on a revolving panel are rendered with a single call). The use of `table` poses a new type of performance requirements as it's a dense 2D grid of unbounded size, and rendering must happen fast.
2. Misc complications (note to future self, or the one working on it):
    - for text truncation, `textpath` is being used which on 1st look doesn't seem compatible with `tspan`s
    - when we switch to multiline text then `convertToTspans` won't support it; if it made to do so, then there's got to be some return value of how many rows (or how many pixels tall) the text is wrapped

I was already thinking about eventually bypassing `D3` to increase SVG text rendering speed a bit, but the shallow `D3` overhead is low compared to SVG paint time.

Caching? Part of the expense is the parsing of the cell `value`s for HTML tags to build a 'document' tree. It could be cached for the entire table, ie. eagerly, or on 1st render on a revolver panel, do the parse and retain the output (even in the form of an `outerHTML` or a retained DOM fragment for cheap insertion - but then the cache may get pretty large, or a MRU/LRU/LFU cache is both complicated and slow if encounters cache misses; eager calc for the entire table is expensive thus UI-blocking). Given the request to work on `crossfilter` it sounds like a future task.

What _could_ be done relatively quickly: enhance the `font` spec with `font-style`,`font-weight` and perhaps `alignment-baseline` or `dy` ie. an entire cell could be turned bold or italic (useful and common feature while heterogeneous styling within a cell is much less common). It'd fit nicely as a HTML/SVG/CSS standard, and would be fast, but maybe not desired for API consistency.

Upshot: 
- for the time being, we could add `font-style` and`font-weight` array attributes for `table` if neeed
- we could gather all text rendering requirements (e.g. multiline rendering, use of `convertToTspans` logic, caching for speed and perhaps even the refactoring of `convertToTspans` if it has perf reserves) and do it as one comprehensive task.

monfera:
As part of some code refactoring, now `table` supports headers with an arbitrary row count:

image

Things in the header, as usual, can be styled per row, column, cell or individually, except for row height, which is uniform across `header` rows (but can be different from the height of regular rows, as seen above).

src/traces/table/attributes.js
monfera: @alexcjohnson the current push should build fine, the attribute documentation join type is fixed. The `table_large.json` is obsolete (will be reworked), but the `table.json` should be OK.

jackparmer:
Can't wait to unleash this to Python/Pandas users, check out this thread:

image

https://forum.modeanalytics.com/t/36z7t8/python-notebooks-best-way-to-format-a-table

monfera:
@jackparmer thanks for the details! The first two features are done, the wrapping (on the tasklist) isn't done yet (only truncation), and manual column resizing isn't on our tasklist though it's an easier task than wrapping. Looking fwd to advancing this WIP item.

etpinard:
Adding this to the [`v1.30.0`](https://github.com/plotly/plotly.js/milestone/26) milestone. @monfera anything we can do to help?

monfera:
@etpinard I'm back on `table` in the morning. Thanks for your question - if there's capacity, these can be done in parallel: 

A. Hover tooltip and events are not yet done - these shouldn't be too complicated

B. Scrollbar: it'd be good to reuse most of of `ScrollBox` but I'm new to it ie. not the most efficient. Also, `table` is a revolving scroller, so there may be wrinkles, or `ScrollBox` may need modification (if the code assumes a static panel to scroll, and/or slow if the scroll distance is large). 

C. Multiline text:

1. Actual text wrapping (where can breaks go, how is text length measured inexpensively etc.), also, reusability [elsewhere](https://github.com/plotly/plotly.js/pull/1834#issuecomment-313485106) 
2. Making `table` work with heterogeneous row height, as needed for multiline text (various rows may have different amounts of text, leading to fewer or more text lines)

D. Test cases 

E. Wheel scroll isn't done yet (drag scrolling is done which just uses the standard `d3.behavior.drag` so wheel scroll will somehow need to talk to the underlying logic lest we duplicate some of `d3.behavior.drag` logic)

I think it'd be easiest if I did C/2 and E at least. Also, due to the revolver nature, if someone familiar with `ScrollBox` works on B, I can take over when it comes to the revolving part. Other than that, the other tasks could be done in parallel. Regarding especially D, it'd be unfair/wrong to let me work on the fun(ctional) part and dish out the less glamorous parts to someone else, so I'm more than glad to cover as much as possible.

etpinard:
@monfera ok, looks like there's quite a bit of work left to do. 

As there are 15 unreleased PRs merged master, I'll drop this one from the `1.30.0` milestone and release `1.30.0` today. I hope you don't mind.

monfera:
@etpinard thanks, indeed, I'm getting close to finishing C/2 now, this itself has been quite tedious as the revolving panels were built when varying row heights weren't yet needed (I did the simplest thing back then that used SVG and permitted scalability of row count). Will switch to the other topics after it's done (plan: today).

monfera:
C/2 Heterogeneous row heigths are now supported with the revolving scroller, the precondition for breaking contents to multiple lines:

scroll

monfera:
Plotly markup and asynchronously loading, size-adjusting LaTeX in combination with revolver panels (debug view showing clip paths etc.)

latex latex

monfera:
Working on the last view element for this PR today, the scrollbar.

Adding a simple vertical scrollbar now, just a grey section looking like the OS X scroll bar. 

I looked into about requiring `scrollbox.js` but the `table` scroller is virtualized, ie. there's no actual contiguous thing that does the scrolling, it's just rendering small blocks of rows and `translate`ing them, so it looked like a lot of work to link them together, for not much to gain.

Also, a native scrollbar would be better, but there are some cons:

- it should be linked to some native element, as the virtualized scroll area doesn't give proper height info; a very tall DOM element may be an issue
- HTML (rather than SVG) would be needed; possible to align analogously with `parcoords` (WebGL + SVG) but tedious (time) and bringing in HTML may bring in several unknowns 
- its size wouldn't be controlled by `plotly.js`

So both alternatives look comparatively time consuming.

Since a simple SVG scrollbar is simple, I'd put it in and we can still revisit it with little incurred time even if it gets replaced in another PR. But not having a scrollbar is not good.

etpinard:
now done in https://github.com/plotly/plotly.js/pull/2052

Issue #698, opened 2016-06-28
Support for making multiple selections using Box selector or Lasso selector tools in Plotly.js


juni-green:
Hi,
Is there any way to handle multiple selections simultaneously i.e. without cancelling the previous selection? Currently, when the Box select cursor is selected (from the modebar) and box selection is applied, the previously selected part of the graph gets deselected and the box re-appears for the newly selected part. I was just wondering if we can select new data points on the graph without deselecting the previous ones.

Any help will be appreciated. 
Thanks!

etpinard:
> Is there any way to handle multiple selections simultaneously i.e. without cancelling the previous selection

Not at the moment. 

This is a limitation of the current implementation. See PR https://github.com/plotly/plotly.js/pull/154 for more information.

That said, what you describe would be a great addition to the library.

toothlessdragon:
This would be awesome!

A great thing to do would be to enable pressing shift+selection to add to the current selection (keeping selection without pressing the shift key working as-is).

drcrook1:
I could really use this functionality as well.  We use plotly for annotating various charts and many need multiple annotations.

etpinard:
cc @dfcreative @chriddyp @jackwluo

etpinard:
@dfcreative

etpinard:
@dfcreative

dfcreative:
[Here](https://github.com/dfcreative/plotly.js/tree/multiselect) is work in progress:

![screen](https://user-images.githubusercontent.com/300067/27558844-7a646cda-5a8c-11e7-8c16-b287bb28c081.gif)

Things to do:

* [ ] merge polygons
* [ ] make proper shift interaction
* [ ] cover scattergl case
* [ ] add multibox selection

etpinard:
Awesome @dfcreative :tada: 

We should also discuss what should happen to the `plotly_selected` event data during multiple selections.

I'd vote for emit a `plotly_selected` event after every selection stroke is completed and making the `points` array include points for all selected currently visible.

Thoughts?

dfcreative:
Progress so far

![multisel](https://user-images.githubusercontent.com/300067/27838388-5b61d43a-60b8-11e7-8cea-62b3acc61450.gif)

etpinard:
Done via @dfcreative 's https://github.com/plotly/plotly.js/pull/2140

PR #1853, opened 2017-07-04
Enable multiple selection for box/lasso tool

dfcreative:
Work in progress.

Related to https://github.com/plotly/plotly.js/issues/698.
Holding shift extends previous selection.
impl

Things to figure out:

* [ ] scattergl hover does not show info-tooltip
* [ ] panning breaks/resets the selection


dfcreative:
Covered by #1869

Issue #1434, opened 2017-03-02
Alignment of horizontal grid lines and tick text is off


Kilian:
Visually, the tick text and the horizontal gridlines are not vertically centered, which makes the axis look slightly-off. This is already noticeable in the [line chart demos](https://plot.ly/javascript/line-charts/):
![selection_012](https://cloud.githubusercontent.com/assets/41970/23499707/7b3f63ae-ff2d-11e6-92a5-01a98fc7b2f6.png)

If you zoom in on this image, you can see it more clearly:

![selection_013](https://cloud.githubusercontent.com/assets/41970/23499750/acd3241e-ff2d-11e6-9616-31fcbd93c91c.png)

**What I would expect** is vertical alignment of the tick text (regardless of font-size) with the grid lines. In the above example, the ticks would need to be moved _2px_ up to get the following alignment (left: new position, right: overlayed on current position):

![selection_014](https://cloud.githubusercontent.com/assets/41970/23499846/0ec1aeb6-ff2e-11e6-8964-219a51778701.png) ![selection_015](https://cloud.githubusercontent.com/assets/41970/23499856/19f40a90-ff2e-11e6-8695-6d9812dbe434.png)

etpinard:
I suspect that off-center offset is on purpose. Though it would be nice to add an axis attribute (e.g. `yaxis.ticktextoffset`) to make this configurable.

alexcjohnson:
I don't think it's on purpose; I know that for log axes we fiddle with the positioning of individual labels (particularly to keep the small 9 from interfering with the bigger 10^n) but I suspect for linear it's just us making a poor approximation to avoid having to calculate bounding boxes individually for each tick label. I'd be happy to consider this a bug.

That said there *are* use cases for offsetting the labels - but perhaps this would be best handled as an anchor (top/middle/bottom, or left/center/right for x axes) rather than a pixel offset?

Like this Vox chart:
![screen_shot_2015-01-14_at_11 34 49_am 0](https://cloud.githubusercontent.com/assets/2678795/23520569/925fd07a-ff49-11e6-89b0-a1247992a143.png)

alexcjohnson:
Closed by #1854

FinnFrotscher:
<img width="883" alt="screen shot 2017-09-08 at 19 55 05" src="https://user-images.githubusercontent.com/11148454/30224624-bf69d6f8-94cf-11e7-87b8-8719be105cae.png">
i have the same issue.

[...]

Issue #484, opened 2016-04-28
Axis labels on very tall graph don't reach end of axis


etpinard:
Originally posted on https://github.com/plotly/plotly.py/issues/450

See codepen: http://codepen.io/etpinard/pen/WwajEX

I was able to reproduce the issue with a `scatter` trace as well.

Something must up in the axes code.

alexcjohnson:
Oh that's just a failsafe against [infinite loops](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js#L810-L811) - we could be smarter about that, perhaps just check to see that each label is closer to the end than the previous one was, though I do think we need *some* sort of upper bound to prevent locking the browser - this can happen easily for example if you specify an explicit tick spacing and then zoom out a lot.

etpinard:
Thanks for the clarification @alexcjohnson :books:

PR #1854, opened 2017-07-06
Fix axis line width, length, and positioning for coupled subplots

alexcjohnson:
@etpinard this is for the first bug we came across this afternoon, but I'm not going to fix it until you sort out the second one (re: [`joinLayer`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/index.js#L350)) because I expect the symptoms are likely to change with that fix - although I don't expect it will be fixed. Turns out that there are a bunch of related issues with the wrong line widths getting applied to different lines, and the wrong positioning of ticks & labels regardless of which width is correct. The mock changes below make all the axis lines semi-transparent so we can tell if ticks and counteraxes abut (as intended) or overlap.

Also to note (will be fixed in this PR): axis lines currently get `stroke-width` set both as a style and an attribute - with different values 🙀 . I'll remove this duplication and standardize on style.


alexcjohnson:
@etpinard this is ready for a look. I ended up changing nearly all the baseline images, so let me explain (and please look closely at the new images, let me know if you think these changes are positive AND acceptable in terms of compatibility).

- #1855 did partially fix this, as I hoped it would: at least after that each axis received its own line width, but the positioning still was off.

- The first few commits are a stab at cleaning up the ancient tangle of code that draws axis lines. The functional changes are all in https://github.com/plotly/plotly.js/pull/1854/commits/eb48a9d15c7d52ee2703021c43ec441d692e0ecf

- I started out fixing how axis lines are drawn, which was broken in a number of multiple-axis situations (interestingly there was a [TODO](https://github.com/plotly/plotly.js/blob/294e5f6c9d2e352c208d355b0e78ba7010afa90a/src/plot_api/subroutines.js#L193) about it dating from 4 years ago when we first added multiple axes!) - see test image 20 https://github.com/plotly/plotly.js/pull/1854/commits/eb48a9d15c7d52ee2703021c43ec441d692e0ecf#diff-2355db501d595161704e15acbd3a9960 where I extended this to all combinations of anchored / free / mirror axes, all with different line widths, to make sure the lines all abut correctly at corners.
- Along the way I discovered we were doing weird things with tick label and title positioning as a function of axis line width - which became blazingly obvious in mock 20 with its super fat lines. Rationalizing that behavior moved the ticks on a substantial fraction of our mocks.
- Then since I was staring at label positions and had modified a bunch of the mocks, I thought it would be a good time to address #1434 - not the part where we proposed making it configurable, just the misalignment. I suppose I could have made that a separate commit, but it's really just [this line](https://github.com/plotly/plotly.js/pull/1854/commits/eb48a9d15c7d52ee2703021c43ec441d692e0ecf#diff-f6d5ed25a5d6f57c70063e1e1bbc407dR1799) which is a bit of a magic number but looks good to me across a wide range of fonts and font sizes. I suppose a better way to handle it would be to use bounding boxes but... there are tons of edge cases around that with tick labels, and this is far simpler (and faster) for a result that will only be minutely different. I'll also note that if you look *very* closely, you'll see the tick labels are not quite consistent in how they align with the ticks. In fact their coordinates *are* consistently aligned, I believe, but the ticks are rendered with `crispEdges` so their positions get rounded while the text is antialiased instead. We could round the text positions manually, but we'd have to turn that off for export... big mess. Lets just leave it as is.

- I only put the baseline for mock 20 in the code commit, as that's got the key features I was fixing. All the other baselines are in the next commit.

And an aesthetic question for you: Notice in mock 20 and the [x line](https://github.com/plotly/plotly.js/pull/1854/commits/eb48a9d15c7d52ee2703021c43ec441d692e0ecf#diff-07cd4d425af26971d4a589aaac8ea059R258) and [y line](https://github.com/plotly/plotly.js/pull/1854/commits/eb48a9d15c7d52ee2703021c43ec441d692e0ecf#diff-07cd4d425af26971d4a589aaac8ea059R275) comments about how we handle free axes: if a free axis is aligned with an opposite-letter anchored axis, I opted to extend its line out the same way as if both axes are anchored - but perhaps it would be better to leave free axes always ending at the edge of the subplot? That would mean for example that the green, yellow, purple, and brown axes in 20 would be shortened to the same span as the blue and red axes.

etpinard:
Wow. Thanks for digging up https://github.com/plotly/plotly.js/issues/1434.

Were you aware of https://github.com/plotly/plotly.js/issues/484 when working on this branch?

----

Starting my review now.

src/plot_api/subroutines.js
```diff
@@ -79,10 +79,8 @@ exports.lsInner = function(gd) {
             return;
         }
 
-        var xa = Plotly.Axes.getFromId(gd, subplot, 'x');
-        var ya = Plotly.Axes.getFromId(gd, subplot, 'y');
-        var xDomain = xa.domain;
-        var yDomain = ya.domain;
+        var xDomain = plotinfo.xaxis.domain;
```
etpinard: Amazing. Note that references are updated in `plotinfo` in [`linkSubplots`](https://github.com/plotly/plotly.js/blob/edf52ca901923959023feeb351ab7f3d4e82faab/src/plots/plots.js#L633) called during supplyDefaults.

src/plot_api/subroutines.js
```diff
@@ -43,7 +43,9 @@ function overlappingDomain(xDomain, yDomain, domains) {
 exports.lsInner = function(gd) {
     var fullLayout = gd._fullLayout;
     var gs = fullLayout._size;
+    var pad = gs.p;
     var axList = Plotly.Axes.list(gd);
+    var hasSVGCartesian = fullLayout._has('cartesian');
```
etpinard: To note, this is needed as gl2d now uses cartesian (SVG layers) subplots to handle selections.

alexcjohnson: Ah right - I figured out that that's the reason, which is why I added `SVG` to the var name, but perhaps I should add the full explanation in a comment. Will become important when we start using SVG axes with GL2D plots @dfcreative

alexcjohnson:
> Were you aware of #484 when working on this branch?

When I first saw the title of that issue "Axis labels on very tall graph don't reach end of axis" I thought it might be related to the axis line being (sometimes) longer than the subplot (and therefore you can't have ticks/labels all the way to the end) - see my "aesthetic question" in https://github.com/plotly/plotly.js/pull/1854#issuecomment-314982882. That's one way in which it actually isn't a purely aesthetic consideration, you could argue that axis lines for free axes should map exactly to the extent of the plot area, to help the viewer map far-away data points onto the axis.

I think I buy that argument actually, so I'd be inclined to make this change. Any objection?

src/plot_api/subroutines.js
```diff
@@ -165,129 +164,219 @@ exports.lsInner = function(gd) {
...
+        var yIsFree = ya.anchor === 'free';
+        var showFreeY = yIsFree && !freeFinished[ya._id];
+        var showLeft = shouldShowLine(ya, xa, 'left');
+        var showRight = shouldShowLine(ya, xa, 'right');
+
+        var xlw = Drawing.crispRound(gd, xa.linewidth, 1);
+        var ylw = Drawing.crispRound(gd, ya.linewidth, 1);
+
+        function findMainAxis(ax) {
+            return ax.overlaying ? Plotly.Axes.getFromId(gd, ax.overlaying) : ax;
```
etpinard: Non- :no_entry: : might be nice to stash references to overlaying and anchor axes in `plotinfo` during [`makeSubplotData`](https://github.com/plotly/plotly.js/blob/edf52ca901923959023feeb351ab7f3d4e82faab/src/plots/cartesian/index.js#L234) and :hocho: those `Axes.getFromId`

src/plot_api/subroutines.js
```diff
@@ -165,129 +164,219 @@ exports.lsInner = function(gd) {
...
+        }
+
+        function findCounterAxes(ax) {
```
etpinard: Would it be too much to ask to move those new `find*` functions outside of `subplotSelection.each` scope? I don't think compilers can optimize that still meaning these functions are redefined for every subplot.

src/plots/cartesian/axes.js
```diff
@@ -1792,7 +1795,9 @@ axes.doTicks = function(gd, axid, skipTitle) {
         }
         else {
             flipit = (axside === 'right') ? 1 : -1;
-            labely = function(d) { return d.dy + d.fontSize / 2 - labelShift * flipit; };
+            labely = function(d) {
+                return d.dy + d.fontSize * 0.35 - labelShift * flipit;
```
etpinard: Non- :no_entry: (might be suited for #1434 - the part where we proposed making it configurable) but this `0.35` would look really nice side-by-side [`LINE_SPACING`](https://github.com/plotly/plotly.js/blob/master/src/constants/alignment.js#L33).

src/plot_api/subroutines.js
```diff
@@ -162,129 +164,132 @@ exports.lsInner = function(gd) {
...
                 .attr('d', (
-                    (showleft ? ('M' + leftpos + ypathSuffix) : '') +
-                    (showright ? ('M' + rightpos + ypathSuffix) : '') +
-                    (showfreey ? ('M' + freeposy + ypathSuffix) : '')) ||
+                    yLinePath(yLinesXLeft, showLeft) +
+                    yLinePath(yLinesXRight, showRight) +
+                    yLinePath(yLinesXFree, showFreeY)) ||
                     'M0,0')
-                .attr('stroke-width', ylw + 'px')
+                .style('stroke-width', ylw + 'px')
```
etpinard: Good catch. Is `.attr('stroke-width', /* */)` a thing?

src/plot_api/subroutines.js
```diff
@@ -162,129 +164,132 @@ exports.lsInner = function(gd) {
...
+        plotinfo.yaxislayer.attr('transform', originY);
         plotinfo.gridlayer.attr('transform', origin);
         plotinfo.zerolinelayer.attr('transform', origin);
         plotinfo.draglayer.attr('transform', origin);
 
         // mark free axes as displayed, so we don't draw them again
-        if(showfreex) { freefinished.push(xa._id); }
-        if(showfreey) { freefinished.push(ya._id); }
+        if(showFreeX) freeFinished[xa._id] = 1;
+        if(showFreeY) freeFinished[ya._id] = 1;
```
etpinard: Smart. I usually set keys like this as `true`. But `1` is truthy **and** three characters shorter!

src/plot_api/subroutines.js
```diff
@@ -165,129 +164,219 @@ exports.lsInner = function(gd) {
...
+            }
+            return 0;
+        }
+
+        /*
+         * x lines get longer where they meet y lines, to make a crisp corner
+         * free x lines are not excluded - they don't necessarily *meet* the
+         * y lines, but it still looks good if the x line reaches to the ends
+         * of the y lines, especially in the case of a free axis parallel to
+         * an anchored axis, like this:
```
etpinard: I agree 100%. Thanks for documenting.

src/plot_api/subroutines.js
```diff
@@ -165,129 +164,219 @@ exports.lsInner = function(gd) {
...
+        var xLinesXRight = xa._length + pad + findCounterAxisLineWidth(xa, ylw, showRight, 'right');
+        var xLinesYFree = gs.h * (1 - (xa.position || 0)) + ((xlw / 2) % 1);
+        var xLinesYBottom = ya._length + pad + xlw / 2;
+        var xLinesYTop = -pad - xlw / 2;
+
+        /*
+         * y lines do not get longer when they meet x axes, because the
+         * x axis already filled that space and we don't want to double-fill.
+         * BUT they get longer if they're free axes, for the same reason as
+         * we do not exclude x axes:
```
etpinard: :+1:

etpinard: mock `20` looks amazing.

src/plot_api/subroutines.js
alexcjohnson: Yes, a lot of properties in SVG you can use either `.attr` or `.style`, with slightly different rules but also very different precedence - see https://css-tricks.com/presentation-attributes-vs-inline-styles/ - attributes are about the lowest, while inline styles are nearly the highest.
@monfera and I had a discussion about this a few days ago, I think for our purposes `.style` should be preferred.

etpinard: Ha, interesting. It's something we might to standardize in `strict-d3.js`.

monfera: I think `stroke-width` as an attribute doesn't need the `'px'` but also, the question kept brewing in my mind and I like the (user-experienced) flexibility of using attributes where possible, as it's easy to restyle them. With `D3` the typical usage is inline styling of the element, which is for our current context, the highest priority, ie. the user would have a hard time overriding it (there are two main ways: 1) user reselects with D3 and changes the inline style; 2) user specifies `!important` styling in the style sheet. So, using `attr` would be more convenient for users if they choose to restyle via CSS.

etpinard:
> you could argue that axis lines for free axes should map exactly to the extent of the plot area, to help the viewer map far-away data points onto the axis.

> I think I buy that argument actually, so I'd be inclined to make this change. Any objection?

Hmm. By _extent of the plot area_, you mean spanning exactly `axis.range`?

src/plot_api/subroutines.js
monfera: ... also, there's the thing that Microsoft browsers (IE, Edge) don't have full CSS support, so if the decision is to go with `style` as much as we can, then we still must use `attr` for `transform` and possibly some more presentation attributes, ie. some `style`->`attr` switch may need to happen during MS testing

alexcjohnson:
> Hmm. By _extent of the plot area_, you mean spanning exactly `axis.range`?

domain, but yeah - the idea is to have it exactly match the x or y pixel range on which we display data.

src/plot_api/subroutines.js
alexcjohnson: Right, `transform` should stay as an `attr` for MS support. I think even some things like `x` and `y` that we've always provided as attributes *can* be styles. I think my own proclivity is to treat positioning as attributes and everything else as style, but I'm sure I've broken that rule. Might be worthwhile to create a list of property names we use and whether each should be used as a style or attribute? And then yes, we could enforce that in `strict-d3`.

I'm sympathetic to the idea of giving users more control, but in the end I think it would be a mistake to encourage overriding internals of the plot with CSS - we should push people to use the plot JSON instead - for three reasons:
1. Portability: with outside CSS the figure JSON no longer fully describes the plot. Even if you don't care about sharing a plot to another server, you might care about the image server or - depending on how the user applied their styling - even in-browser downloading could be affected.
2. Accidental overrides: we use a lot of generic class names like `.point`, `.xtitle`, that could get mangled easily if the page does something else with those classes.
3. Forward compatibility: obviously anyone can peek inside our elements and find selectors for whatever elements they'd like to style, but we've never published this as part of the API, and I don't want to give up the right to refactor our structures and rename classes.

monfera: Yes these points are all true. We don't currently go out of our way to dissuade scenegraph class based styling - we don't generate mangled or one-off class names or otherwise actively dissuade external styling. Also on the forum, SO and gh questions sometimes our answer is that a particular wish isn't supported but can be done with a `d3.selectAll('whatever').style(...)` and our scenegraph class is human friendly so I over-inferred 😄  That and the upside of not having to have heuristics for when to switch to `attr`. Btw. as a developer either way is fine as long as there's a guideline like what you mentioned above @alexcjohnson 

Not sure if the emerging linter solution already covers it, I think this is the [list of attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute#Presentation_attributes) (ie. all _presentation attributes_) that can go either way. 

Unfortunately it's quite Venn-diagrammey and common stuff like `x`, `dx`, `r` etc. ie. drivers of the most basic geometry are not on it so even in the future, those things won't be able to use CSS transitions/animations. (Maybe interesting is that the SMIL deprecation note got retracted.)

etpinard:
@alexcjohnson here's what mpl does:

https://gist.github.com/etpinard/55254d82e1fd48d96a920bb306506238#file-mpl-free-axis-line-span-ipynb

Looks like free axes do span the whole axis domain regardless of their sister axis line width.

---

All in all, we'll probably need some new attribute governing this behavior in the future. I'll let you decide the _best_ default value for now.

etpinard:
And for those you (like me), that had issues understanding the implications of 

> you could argue that axis lines for free axes should map exactly to the extent of the plot area, to help the viewer map far-away data points onto the axis.

Here's the effect as of eb48a9d when relayout'ing `xaxis.linewidth`:

peek 2017-07-13 16-29

src/plot_api/subroutines.js
alexcjohnson: comment added https://github.com/plotly/plotly.js/pull/1854/commits/6d41a3701bdbb041013ca9547f75745d42b34784

src/plot_api/subroutines.js
alexcjohnson: good idea. Actually `makeSubplotData` didn't really seem the place for this, it duplicated a lot of effort that should have just been done once per axis. I put this all into `plots.linkSubplots`. https://github.com/plotly/plotly.js/pull/1854/commits/274526afca78f8ac7d41429f2ead483901050c3e

src/plot_api/subroutines.js
alexcjohnson: ah, I actually meant to un-nest these, thanks for reminding me. https://github.com/plotly/plotly.js/pull/1854/commits/b31e60d3380122ec01e01c1cd5ffb51ab89d0207

src/plots/cartesian/axes.js
alexcjohnson: hah, I found the 0.35 value empirically, but polar had the same one, and it turns out there's a rational basis for exactly this value. https://github.com/plotly/plotly.js/pull/1854/commits/16bae9ca0be3328e6b0f6b7d975ffdb0eafd6720

alexcjohnson:
> you could argue that axis lines for free axes should map exactly to the extent of the plot area, to help the viewer map far-away data points onto the axis.

I tried this out, and I think it both looks good and best represents the data. https://github.com/plotly/plotly.js/pull/1854/commits/739c9988cd1f8c44126d627bbb25b592f05ff159

Also I included `margin.pad` in that test image. It's not shown this way, but `cliponaxis: false` #1861 suddenly makes `margin.pad` a much more useful feature, since you can use it to set the axes back from the plot area (to give extra room so these non-clipped markers will not cover lines and labels) without increasing the axis ranges. cc @geocosmite

src/plots/cartesian/axes.js
monfera: indeed! just out of curiosity, would `alignment-baseline: central` be an alternative, or there's something that would be lost with it? (tweenability comes to mind, smooth with pixels but not nice with discrete levels, though probably tweening isn't applicable here)

alexcjohnson: > would `alignment-baseline: central` be an alternative

Quite possibly! I recall some baseline properties that I tried to use in the past not having sufficient cross-browser support, so we would need to test it. Also I'm not sure how this would work with pseudo-html (subscript/superscript, changing font sizes, multiple lines...) so I'm not going to do it now but we could investigate later.

monfera: If you run into such a thing I'm interested as the upcoming `table` uses it. IIRC `dominant-baseline` had some inconsistencies but don't now recall an issue with `alignment-baseline`. We'll see, I just bought a machine for real-life testing which as MS Edge on it.

src/plots/polar/micropolar.js
```diff
@@ -295,7 +296,7 @@ var µ = module.exports = { version: '0.2.2' };
             angularAxisEnter.append('text').classed('axis-text', true).style(fontStyle);
             var ticksText = angularAxis.select('text.axis-text').attr({
                 x: radius + axisConfig.labelOffset,
-                dy: '.35em',
+                dy: MID_SHIFT + 'em',
```
etpinard: :trophy: for patching the polar code.

test/image/mocks/20.json
```diff
@@ -291,6 +291,6 @@
         },
         "width": 750,
         "height": 600,
-        "margin": {"r": 200, "l": 50, "t": 50, "b": 50}
+        "margin": {"r": 200, "l": 50, "t": 50, "b": 50, "pad": 5}
```
etpinard: :trophy: for bringing `pad` back to life.

src/plots/plots.js
```diff
@@ -691,6 +693,39 @@ plots.linkSubplots = function(newFullData, newFullLayout, oldFullData, oldFullLa
             }
         }
     }
+
+    // while we're at it, link overlaying axes to their main axes and
+    // anchored axes to the axes they're anchored to
+    var axList = Plotly.Axes.list(mockGd, null, true);
+    for(i = 0; i < axList.length; i++) {
```
etpinard: Much better. I'd glad `linkSubplots` is getting more attention. Updating all those refs early during `supplyDefaults` sounds like the way to go.

One small non-:no_entry: comment, that we could easily address later: `Axes.list` is now called twice during `Plots.supplyDefaults` here and after `doAutoMargin` [here](https://github.com/plotly/plotly.js/blob/739c9988cd1f8c44126d627bbb25b592f05ff159/src/plots/plots.js#L500). Maybe we could call it once and stash the result in `fullLayout` instead?

etpinard:
:dancer: solid work.

:smile: to see that `cliponaxis: false` made us discover (and fix) a bunch of little axis bugs.
 
----

> I thought it would be a good time to address #1434 - not the part where we proposed making it configurable, just the misalignment.

Make sure to either close #1434 and open up a new issue about making tick label alignment configurable OR update the #1434 issue title accordingly.

PR #1856, opened 2017-07-06
Log contour

### LINKED ISSUE #641: This contour trace has intersecting contour lines
See http://codepen.io/etpinard/pen/wWGOJw

First reported on http://community.plot.ly/t/possible-bug-intersecting-contours/633

alexcjohnson:
fixes #641 

We have a step in contour path creation that coalesces points that are too close together relative to the average distance between points. Without this we can get strange effects at crossings very close to corners of grid cells, but it was implemented in pixel coordinates, which means on nonlinear grids (particularly log axes with zero or negative positions) the threshold was incorrect. Fixed by calculating this threshold in index units.

Some existing mocks changed. Mostly the changes are minor and possibly even improvements. There's a slightly annoying one in `contour_lines_coloring` (and the others that use that same data) where it adds a new crossing of some nearby contours with differing topology, but to me that's just increasing our motivation to find an altogether better way to interpolate these curves.

@rreusser please review (since hopefully ^^ will land on your plate at some point!)

cc @etpinard


src/traces/contour/find_all_paths.js
```diff
@@ -45,9 +45,10 @@ function equalPts(pt1, pt2, xtol, ytol) {
            Math.abs(pt1[1] - pt2[1]) < ytol;
 }
 
+// distance in index units - uses the 3rd and 4th items in points
 function ptDist(pt1, pt2) {
-    var dx = pt1[0] - pt2[0],
-        dy = pt1[1] - pt2[1];
+    var dx = pt1[2] - pt2[2],
+        dy = pt1[3] - pt2[3];
```
rreusser: Is 0/1 vs 2/3 the difference between data units and grid index units?

alexcjohnson: yes - kinda hacky but since we had this available when calculating the pixel positions I tacked it on.

etpinard:
Looks good! Thanks for the quick fix :tada: 

I can't notice any differences in the updated baseline, except for

peek 2017-07-07 13-57

If this is expected, :dancer:

alexcjohnson:
Yep, the changes are pretty minor - and in fact I notice that the change you highlighted actually swaps one instance of crossing contours for another... so it's basically a wash until we make an altogether different interpolation algorithm.

Issue #1703, opened 2017-05-18
Parallel coordinates plot axes labels overlap


kernc:
On parallel coordinate plot, axes labels overlap when the text is too long, font too large, or there are too many dimensions for given canvas width.

CodePen example: https://codepen.io/anon/pen/LygNGG

A short proposed solution, please have a look at **PR** https://github.com/kernc/plotly.js/pull/1

![screenshot - 05182017 - 07 49 37 pm](https://cloud.githubusercontent.com/assets/684364/26216098/2a5f3190-3c03-11e7-8e06-fc35371b9420.png)

Slightly related: https://github.com/plotly/plotly.js/issues/511

etpinard:
See https://github.com/plotly/plotly.js/pull/1256 for other known limitations.

kernc:
I reviewed the thread. What do you mean by other limitations?

etpinard:
The unchecked boxes:

![image](https://cloud.githubusercontent.com/assets/6675409/26222368/b8885602-3be7-11e7-995f-1eae8c5ec23f.png)

kernc:
Well, the [proposed PR](https://github.com/kernc/plotly.js/pull/1) checks the first item by an _"other way of avoiding overlap"_. 😃 Would you prefer the labels slanted?

etpinard:
> Would you prefer the labels slanted?

Whatever `parcoods` prince @monfera prefers.

bhvbhushan:
@kernc were you able to resolve the issue, with the possible solution that @monfera mentioned [here](https://github.com/plotly/plotly.js/pull/1857#issuecomment-313492943).
If yes, could you please share here. It would be really helpful.

Thanks

harishr1308:
Hi @bhvbhushan, has this been merged? Or, is there any other way to resolve the overlap. Perhaps, having the text on multiple lines?

Thanks

javdev:
we had the same problem, slanting sucks a bit, cause it needs a lot of space. Wouldn't it be a simple idea, to make two lines for the labels and every second label is put in the bottom line?

javdev:
I could do this myself, just adding a newline to half of the labels on at the end and a newline to the other half at the beginning of the labeltext, but the newline gets removed, as it seems. If it only would get trimmed, I could add a little dot, but it gets really removed

gvwilson:
[...]

PR #1857, opened 2017-07-06
Parcoords: ellipsize long labels; show full text in tooltip

kernc:
Ellipsizes labels that are too long to prevent overlapping. But always show full label text in mouse hover tooltip.

Fixes https://github.com/plotly/plotly.js/issues/1703.

### Before
screenshot - 05182017 - 07 49 37 pm

### After
screenshot - 05182017 - 07 47 22 pm

### DIFF
--- a/src/traces/parcoords/parcoords.js
+++ b/src/traces/parcoords/parcoords.js
@@ -598,7 +598,20 @@ module.exports = function(root, svg, styledData, layout, callbacks) {
     axisTitle
         .attr('transform', 'translate(0,' + -c.axisTitleOffset + ')')
         .text(function(d) {return d.label;})
-        .each(function(d) {Drawing.font(axisTitle, d.model.labelFont);});
+        .each(function(d) {Drawing.font(axisTitle, d.model.labelFont);})
+        .each(function(d) {
+            // Ellipsize long labels
+            var maxWidth = d.xScale(1) - 5;
+            while(this.textContent.length > 2 &&
+                  this.getComputedTextLength() > maxWidth) {
+                this.textContent = this.textContent.slice(0, -2) + '…';
+            }
+        });
+
+    // Tooltip
+    axisTitle.enter()
+        .append('title')
+        .text(function(d) {return d.label;});
 
     var axisExtent = axisOverlays.selectAll('.axisExtent')
         .data(repeat, keyFun);


monfera:
Thanks again for your work! We just had a chat and while the ellipsis approach would solve this case and indeed it's a very compact diff, the other plotly.js plots do something different, eg. the Cartesian plots rotate the labels if there's not enough room. We were planning to add it at some point but it's a good bit of work as rotation of the long texts needs to result in a vertically and horizontally smaller plot (given the same overall layout.width and layout.height) so as the long, tilted labels fit. We were thinking of this text rotation for multiple reasons:

the ellipsis still leaves part of the text unreadable while rotation can show the full text
the ellipsis effect can be approximated by the developer setting a character limit and truncating text (yes it's approximate only)
there can be up to 60 dimensions, where the distance between axes is likely very small
the iterative text shortening is slightly expensive by its nature
The ellipsis would be an improvement on the current situation but maybe a cheaper way to achieve almost the same thing (without the ellipsis) is the use of <textPath> where an invisible path is in the DOM and referenced by the <textPath> element, with the result that the text is rendered only till it reaches the end of the line.

@etpinard what do you think, would a simple text truncation be a suggested improvement? It needs no iteration.

etpinard:
> what do you think, would a simple text truncation be a suggested improvement?

This PR _would be_ an improvement for sure. But, implementing a better, more plotly-esque improvement (i.e. rotating the labels) down the road would then be a breaking change.

So, I'm afraid I'll have to vote :-1: on this PR.

monfera:
@kernc I'll need to close this specific PR for the above reason and also due to the need for using plotly-style tooltips which are quite a bit more involved than the `<title>` option. For now, your truncation solution seems to be able to run as `Plotly.newPlot(...).then(function() { ... add-on code ...})` where the add-on code invokes your `d3.selectAll('parcoords text.axisTitle').each(function(d) {..your truncation code...})`. Thanks again for your constructive PR, sometimes it's hard to meet all constraints with an otherwise effective solution.

kernc:
Right, no problem. The add-on thing is what I'm doing now, but I just though this approach might be better than the currently overlapping labels. When you get to implementing the labels properly, please consider that the assumption:
> while rotation can show the full text

is incorrect. The labels in various domains can ordinarily be of arbitrary/mixed length, so if precise look and feel is of importance, some truncation or (imho) ellipsization would still likely find its place place.

monfera:
@kernc thanks, valid point - otherwise, even with fully vertical text, we'd run out of vertical space with no room for the actual `parcoords`. I'll need to check what the Cartesian does with very long labels.

PR #1884, opened 2017-07-14
Fix heatmapgl point number event data

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1825

Luckily, the problem described in #1825 only concerns `heatmapgl` - as opposed to the solution proposed in https://github.com/plotly/plotly.js/pull/1833. 

The patch is simple. It brings `heatmapgl`'s `pointNumber` event data field on par with other 2D trace types. This PR also adds an _asymmetric_ `heatmapgl` hover test case. 

cc @alexcjohnson @dfcreative

### DIFF
--- a/src/traces/contourgl/convert.js
+++ b/src/traces/contourgl/convert.js
@@ -77,7 +77,7 @@ proto.handlePick = function(pickResult) {
         ],
         textLabel: this.textLabels[index],
         name: this.name,
-        pointIndex: [xIndex, yIndex],
+        pointIndex: [yIndex, xIndex],
         hoverinfo: this.hoverinfo
     };
 };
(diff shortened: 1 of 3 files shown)


src/traces/contourgl/convert.js
```diff
@@ -77,7 +77,7 @@ proto.handlePick = function(pickResult) {
         ],
         textLabel: this.textLabels[index],
         name: this.name,
-        pointIndex: [xIndex, yIndex],
+        pointIndex: [yIndex, xIndex],
```
etpinard: thanks @dfcreative for that https://github.com/plotly/plotly.js/issues/1825#issuecomment-311787182

alexcjohnson:
Works for me! 💃

Issue #1682, opened 2017-05-12
Bug: histogram silently ignores data, when bin size is too small


cnowzohour:
When plotting histograms with very fine grained bins, some data will be ignored. See:

https://codepen.io/cnowzohour/pen/qmYmoM

vs

https://codepen.io/cnowzohour/pen/vmjmjM

Both histograms should show 10 bars, but the second one only shows 5 - the only difference being the bin size of 0.005 vs 0.001.

I realized this when visualizing timestamps over one day with a bin size of 10 seconds:

https://github.com/ropensci/plotly/issues/1001

etpinard:
Thanks for the report. Looks like this hard-coded [5000](https://github.com/plotly/plotly.js/blob/6653e71d79313627194628e6d2e41308ac7deca8/src/traces/histogram/calc.js#L109) bin limit is the culprit. For example, bumping that number to 50000 yields:

![image](https://cloud.githubusercontent.com/assets/6675409/26017167/430b22fc-3736-11e7-8abb-893866f00403.png)

as desired.

We should come up with a better upper bound. Maybe something like `x.length / bins.size`?

cnowzohour:
Yeah I guess 5000 can be a bit conservative in cases when the data is sparse and only a small subset of the bins are actually drawn. Also, it would be great if it was somehow visible if this limit is exceeded (and data is omitted). Maybe via a big last bin, containing all the remaining data?

etpinard:
Update: changing the while clause from

```js
while(i < binend && pos.length < 5000) {}
```

to 

```js
while(i < binend && pos.length < (pos0.length / bins.size)) {}
```

breaks some of our tests. We'll have to dig deeper to find a solution.

PR #1887, opened 2017-07-17
Compute histogram bin positions beyond length 5000

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/1682

@alexcjohnson does this look acceptable?

### DIFF
--- a/src/traces/histogram/calc.js
+++ b/src/traces/histogram/calc.js
@@ -106,7 +106,7 @@ module.exports = function calc(gd, trace) {
     // decrease end a little in case of rounding errors
     binend = pr2c(binspec.end) + (i - Axes.tickIncrement(i, binspec.size, false, calendar)) / 1e6;
 
-    while(i < binend) {
+    while(i < binend && pos.length < 1e6) {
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Makes sense, the fix for #484 is going to be very similar. But, like my comment there https://github.com/plotly/plotly.js/issues/484#issuecomment-315094312 I wonder if we still want to have some hard-coded upper bound on the total number, even if it's super high like 1e6?

etpinard:
> though I do think we need some sort of upper bound to prevent locking the browser - this can happen easily for example if you specify an explicit tick spacing and then zoom out a lot.

Yeah, a super high upper bound would make sense. Maybe we could put that upper bound somewhere in a constants file and reuse it in all our `while` loops? Though, locking the browser is less concerning here as this `while` loop is called in the calc step as opposed to #484 which is called in the plot step.

etpinard:
> wonder if we still want to have some hard-coded upper bound on the total number, even if it's super high like 1e6?

done in eef9158d82c1d307b479721b391d074415af46c7

alexcjohnson:
💃

PR #1888, opened 2017-07-17
safer construction of popup click handler

alexcjohnson:
Fixes an XSS vulnerability in links defined with the special `popup` attribute. See the added test case for three variants that were previously able to exploit it.

The fix here constructs the `onclick` handler in a safer way: it re-extracts `href` and `target` from the element attributes, and it restricts the characters allowed in the `popup` (`strWindowFeatures`) attribute since we know there's [only so much](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) that can be used there.

cc @etpinard @scjody - this bug came in with #1783 which was included in v1.28.0

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -258,7 +258,7 @@ var BR_TAG = /<br(\s+.*)?>/i;
 var STYLEMATCH = /(^|[\s"'])style\s*=\s*("([^"]*);?"|'([^']*);?')/i;
 var HREFMATCH = /(^|[\s"'])href\s*=\s*("([^"]*)"|'([^']*)')/i;
 var TARGETMATCH = /(^|[\s"'])target\s*=\s*("([^"\s]*)"|'([^'\s]*)')/i;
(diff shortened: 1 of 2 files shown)


src/lib/svg_text_utils.js
```diff
@@ -459,7 +461,7 @@ function buildSVGText(containerNode, str) {
                         var dummyAnchor = document.createElement('a');
                         dummyAnchor.href = href;
                         if(PROTOCOLS.indexOf(dummyAnchor.protocol) !== -1) {
-                            nodeSpec.href = href;
+                            nodeSpec.href = encodeURI(href);
```
etpinard: Nicely done.

alexcjohnson: FWIW I couldn't identify any vulnerability due to *not* calling `encodeURI` here - quotes and other special characters embedded in `href` still can't break you out of the `href` attribute (while constructing the `onclick` string or otherwise) to where you can cause more damage - but it makes strange items look more url-like.

PR #1891, opened 2017-07-18
Hover fix for colorbars with custom tickval / ticktext

etpinard:
fixes https://community.plot.ly/t/tooltips-dont-show-up-in-a-heatmap/4972 first reported by @empet

Another bug introduced by https://github.com/plotly/plotly.js/pull/1770. This time `appendArrayPointValue` tried to dig into a trace's `colorbar.tickvals` and `colorbar.tickvals` with a 2D `pointNumber` leading a slew of exceptions during hover. Note that `colorbar.ticktext` and `colorbar.tickvals` are both declared as `valType: 'data_array'` at the moment.

The proposed fix in 0aa87c9da6914bad762e36c2571886db4c93ad93 isn't great, but feels the safest to my eyes.

Perhaps a better, more evolved fix would make a distinction between `data_array` attributes that are expected to have the same dimension as of the _main_ data coordinates (in 2D traces, that _main_ data coordinate is `z`). and _other_ `data_array` attributes. To note, we might **have to** do this to get `filter` and `sort` working for 2D traces in general.

cc @alexcjohnson

### DIFF
--- a/test/jasmine/tests/cartesian_interact_test.js
+++ b/test/jasmine/tests/cartesian_interact_test.js
@@ -440,3 +440,83 @@ describe('axis zoom/pan and main plot zoom', function() {
         .then(done);
     });
 });
+
+describe('Event data:', function() {
+    var gd;
+
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+
+    afterEach(destroyGraphDiv);
+
+    function _hover(px, py) {
+        return new Promise(function(resolve, reject) {
+            gd.once('plotly_hover', function(d) {
+                delete gd._lastHoverTime;
+                resolve(d);
+            });
+
+            mouseEvent('mousemove', px, py);
+
+            setTimeout(function() {
+                reject('plotly_hover did not get called!');
+            }, 100);
+        });
+    }
+
+    it('should have correct content for *scatter* traces', function(done) {
+        Plotly.plot(gd, [{
+            y: [1, 2, 1],
+            marker: {
+                color: [20, 30, 10],
+                colorbar: {
+                    tickvals: [25],
+                    ticktext: ['one single tick']
+                }
+            }
+        }], {
+            width: 500,
+            height: 500
+        })
+        .then(function() { return _hover(200, 200); })
+        .then(function(d) {
+            var pt = d.points[0];
+
+            expect(pt.y).toBe(2, 'y');
+            expect(pt['marker.color']).toBe(30, 'marker.color');
+            expect('marker.colorbar.tickvals' in pt).toBe(false, 'marker.colorbar.tickvals');
+            expect('marker.colorbar.ticktext' in pt).toBe(false, 'marker.colorbar.ticktext');
+        })
+        .catch(fail)
+        .then(done);
+    });
+
+    it('should have correct content for *heatmap* traces', function(done) {
+        Plotly.plot(gd, [{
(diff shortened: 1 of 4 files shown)


alexcjohnson:
Hmm, this will work to exclude 1D arrays from getting picked up alongside 2D arrays - so if you want to include it as a quick bug fix I'm OK with it. But there are going to be cases where the aux arrays have the same dimensionality as the main array and yet still shouldn't be included. Like `colorbar.ticktext` on a scatter trace using `marker.color`. I think your "more evolved fix" may be the only way to handle this in the long run. So, another optional schema attribute like `uncorrelatedArray`?

TBH I'm not sure what `filter` and `sort` even *mean* for 2D data... but that's a discussion for another time and you're absolutely right that if we *do* support this in some way it will require this "more evolved fix".

etpinard:
> Like colorbar.ticktext on a scatter trace using marker.color.

Good point here. Maybe it would be best to _manually_ skip `colorbar.ticktext` and `colorbar.tickvals` in [`PlotSchema.findArrayAttributes`](https://github.com/plotly/plotly.js/blob/f56e07161cb90f729c4bda925f4ddc4f09dcc9f6/src/plot_api/plot_schema.js#L144)?

alexcjohnson:
> Maybe it would be best to _manually_ skip `colorbar.ticktext` and `colorbar.tickvals` in [`PlotSchema.findArrayAttributes`](https://github.com/plotly/plotly.js/blob/f56e07161cb90f729c4bda925f4ddc4f09dcc9f6/src/plot_api/plot_schema.js#L144)?

Good idea to handle this in `PlotSchema.findArrayAttributes`. But I'd much prefer to do it via the schema, so it will scale to other potential such attributes later. And reflect this in at least the docstring for `findArrayAttributes`, and possibly in the name - `findCorrelatedArrayAttributes`? At the moment I don't see any more arrays that are *not* correlated, except for 1D `x` an `y` arrays with a 2D `z` (which in principle are correlated to one or the other dimension of `z` but not both), though these seem to be already taken care of almost accidentally since they're part of the hover data before we even get to the generic array attributes.

test/jasmine/tests/cartesian_interact_test.js
```diff
@@ -440,3 +440,83 @@ describe('axis zoom/pan and main plot zoom', function() {
...
+            width: 500,
+            height: 500
+        })
+        .then(function() { return _hover(200, 200); })
+        .then(function(d) {
+            var pt = d.points[0];
+
+            expect(pt.z).toBe(3, 'z');
+            expect(pt.text).toBe(undefined, 'undefined text items are included');
+            expect('id' in pt).toBe(false, 'undefined ids items are not included');
```
etpinard: N.B. this :lock:  0d91074

etpinard:
> But I'd much prefer to do it via the schema

We would prefer that too, but I don't want to add a schema field until we investigate these _uncorrelated_ `valType: 'data_array'` attributes some more - especially within the realm of transforms. Maybe `colorbar.tickvals` and `colorbar.ticktext` shouldn't be declared `valType: 'data_array'` to begin with? But I guess that one is more of a workspace issue.

etpinard:
Ok. I'm merging this thing. 

https://github.com/plotly/plotly.js/issues/1894 documents the situation and potential long-term more-robust solutions.

Issue #1843, opened 2017-07-03
Filter + groupby breaks if filter includes target data


rreusser:
The example below works if `target` is omitted but breaks if `target` is filled in, even when it's exactly the same as the `x` data that does work. Each transform works independently but the combination fails only when target data is specified.

I'm assuming this means groupby is not outputting the correctly grouped target data. Perhaps findDataAttributes doesn't know how to recurse into other transforms.

https://codepen.io/rsreusser/pen/vZjxWM?editors=0010

rreusser:
A small note: `findDataAttributes` *does* now how to recurse into other transforms. Instead it was failing to actually no-array-clone the transform and then split data arrays into grouped subsets. That's been implemented in #1892.

PR #1892, opened 2017-07-18
Filter-groupby interaction

rreusser:
This PR adds a failing test for filter + groupby. The goal is to rework this into a PR that solves the issue. That may or may not be possible due to the way one is a calcTransform and the other is a transform. I'll dig in and see if I can figure out what it would take to fix this.

See: https://github.com/plotly/plotly.js/issues/1843


rreusser:
@alexcjohnson @etpinard I added some more challenging tests for transforms. The actual change to groupby was smaller than I expected (and deserves a close look).

I implemented [recursive transform application](https://github.com/plotly/plotly.js/blob/9b44b4bd597443eb99cf98828e44800e509ab675/src/plots/plots.js#L1055-L1093) to lock down the way transforms are recursively expanded and iterated, but I wasn't actually able to come up with a test case that it solves, so I removed it. It might be worth keeping in mind though in case we need to get more elaborate.

src/transforms/groupby.js
```diff
@@ -137,8 +137,22 @@ function pasteArray(newTrace, trace, j, a) {
     );
 }
 
+// In order for groups to apply correctly to other transform data (e.g.
+// a filter transform), we have to break the connection and clone the
+// transforms so that each group writes grouped values into a different
+// destination. This function does not break the array reference
+// connection between the split transforms it creates. That's handled in
+// initialize, which creates a new empty array for each arrayAttr.
+function cloneTransforms(newTrace) {
```
etpinard: Interesting fix :+1:

etpinard: @rreusser is that something we should do for all transforms?

src/transforms/groupby.js
```diff
@@ -163,12 +177,14 @@ function transformOne(trace, state) {
...
+        cloneTransforms(newTrace);
+
         arrayAttrs.forEach(initializeArray.bind(null, newTrace));
 
-        for(var j = 0; j < len; j++) {
+        for(j = 0; j < len; j++) {
             if(groups[j] !== groupName) continue;
 
-            arrayAttrs.forEach(pasteArray.bind(0, newTrace, trace, j));
+            arrayAttrs.forEach(pasteArray.bind(null, newTrace, trace, j));
```
alexcjohnson: This isn't new here - except that the change you made reminded me how `bind` always confuses me - but it occurs to me that there's no reason to use it at all if we unwind the `forEach` to a `for`. Also `pasteArray` is a bit inefficient in the way it calls `Lib.nestedProperty(newTrace, a)` twice.

src/transforms/groupby.js
alexcjohnson: Oh hah - of course, because the original `Lib.extendDeepNoArrays({}, trace)` doesn't dive into the `transforms` array. Do we need a `Lib.extendDeepOnlyTheseArrays({}, trace, ['transforms'])` ? 🙈 

> is that something we should do for all transforms?

This only applies to transforms that generate multiple traces, right? I do kind of like the idea of recursive transforms, but until we do something like that, that might be more inherently robust, I think we should manage this case-by-case.

rreusser: So… `arrayAttrs` *does* pick up on data_array arrays nested inside transforms. The particular challenge is executing a transform after another has created expanded traces. Because if the first transform splits it into multiple traces, then each expanded trace needs its own copy of the other transforms since they, for example, have different split `target` or `groups` data. By expanding deep with no arrays and then cloning the data arrays (which are in `arrayAttrs`, they all have their own copy that can be filtered appropriately.

src/transforms/groupby.js
rreusser: 🎉 🎉 🎉  Glad to change 😄

src/transforms/groupby.js
rreusser: Oh, I missed a bit of subtlety. Let me check on the identity of the split `transforms`. Unless I'm mistaken, something somewhere *is* cloning those… Let me dig that up.

rreusser: @etpinard I wondered the same. The downfall of transforms at the moment is that there are too many corner cases and different things transforms *might* need to do to lock them down to a completely airtight state such that they can *only* do what they need to do. The result is that there are some patterns like this that should not be a concern of individual transforms at all but which might need to be applied to them all separately. Or if a clear pattern arises, then it can be abstracted. At the moment, the sample size is just a bit to small for me to *really* see overarching patterns.

etpinard: >  Or if a clear pattern arises, then it can be abstracted. At the moment, the sample size is just a bit to small for me to see overarching patterns.

Very good answer. Thanks!

rreusser: @alexcjohnson Never mind. I read my code more carefully. I'm just manually transferring the transforms. 👍

src/transforms/groupby.js
rreusser: @alexcjohnson @etpinard I've refactored… actually a fair amount of groupby. It was mostly trivial reordering and inlining in order to cut down on function calls and `slice` and `concat` operations, just to streamline things a bit function-call and memory-wise.

rreusser:
@monfera's groupby tests saved me here. Refactored, optimized, and fixed some test cases. I'm once again content. 👍

src/transforms/groupby.js
```diff
@@ -158,20 +152,52 @@ function transformOne(trace, state) {
...
+        // initialize, which creates a new empty array for each arrayAttr.
+        transforms = newTrace.transforms;
+        newTrace.transforms = [];
+        for(j = 0; j < transforms.length; j++) {
+            newTrace.transforms[j] = Lib.extendDeepNoArrays({}, transforms[j]);
+        }
+
+        // Initialize empty arrays for the arrayAttrs, to be split in the next step
+        for(j = 0; j < arrayAttrs.length; j++) {
+            Lib.nestedProperty(newTrace, arrayAttrs[j]).set([]);
```
alexcjohnson: nice - what you're doing here is pretty similar in fact to what I'm doing with the aggregate transform 🎉 

I suspect you could speed it up even a bit more by hanging onto these arrays as you create them, and avoid the inner-loop [`Lib.nestedProperty` calls below](https://github.com/plotly/plotly.js/pull/1892/files#diff-2c8b7c893b5eec3feb83ac0af6030ccfR194)

rreusser: Yeah, that was bugging me, though I know the effect is small in the grand scheme of things. You know exactly what you'll need, so I just indexed them directly and then looked them up by group index. It's still lots of nestedProperty instantiation, but unless we get fancy (read: fragile), I think this is somewhat minimal for the task at hand.

rreusser: ✅

src/transforms/groupby.js
```diff
@@ -152,13 +153,15 @@ function transformOne(trace, state) {
         styleLookup[styles[i].target] = styles[i].value;
     }
 
-    var newDataByGroup = {};
+    // An index to map group name --> expanded trace index
+    var groupIndex = {};
 
     for(i = 0; i < groupNames.length; i++) {
         groupName = groupNames[i];
+        groupIndex[groupName] = i;
```
rreusser: Naming is tricky. It's actually the array index that I'm indexing. So it should be called `indexIndex` :/

rreusser: `indexLookup`. Better, maybe. A variable by any other name would smell… as… hmm… 🌹

alexcjohnson:
Looks great! 💃

rreusser:
I'll defer to someone else on whether this is a bugfix or a feature. Since the output for interactions was simply incorrect/nonsensical before, I'd tend to consider this a bugfix.

etpinard:
> I'd tend to consider this a bugfix.

Yeah. This is a :beetle: fix for sure.

Issue #1750, opened 2017-05-31
plotly_clickannotation event not firing on `editable: true`


marcusblevin:
Has something changed with the plotly_clickannotation event? I have tried reverting back to a previous version (1.22.0) and the event is firing but on latest (1.27.0) it is not.

See [CodePen](https://codepen.io/marcusblevin/pen/KmJJEQ?editors=1011) for example using latest version

etpinard:
We apologise if your graphs are currently broken, adding `captureevents: true` to your annotation settings should do the trick - as in https://codepen.io/etpinard/pen/PmMxZL?editors=0011

See

- https://github.com/plotly/plotly.js/pull/1573#discussion_r110697483
- https://github.com/plotly/documentation/issues/742

for more detail on the issue.

marcusblevin:
Thank you, that fixes the issue with the event firing but is that event dependent upon the editable configuration option not being turned on? I saw you had it disabled in the Codepen. The use case I am trying to accomplish is allow the annotation to be moved and capture the click event to load the currently set properties into an editor the user can use to update the annotation. Seems like this is a one or the other solution.

etpinard:
> but is that event dependent upon the editable configuration option not being turned on?

Yeah. I noticed that too. I'm not sure if on purpose and/or the desired behavior.

> The use case I am trying to accomplish is allow the annotation to be moved and capture the click event to load the currently set properties into an editor the user can use to update the annotation

You should listen to `plotly_relayout` instead to accomplish this task.

marcusblevin:
To be more specific, I use the plotly_clickannotation event to load the properties into the editor and to immediately change the focus so the text in the annotation never shows as editable and the edits instead happen in the form inputs. Plotly_relayout would fire after a change to the plot occurs but I want the event to be captured before any changes are made. Granted I could use plotly_relayout to capture the info when the annotation is moved but the loading of the inputs should not be dependent upon the annotation being repositioned.

etpinard:
> to immediately change the focus so the text in the annotation never shows as editable and the edits instead happen in the form inputs.

That's an interesting use case, probably not what we envision `plotly_clickannotation` to be for, but interesting nonetheless.

----

After some thinking a little bit longer, I'm going to call this a bug: `plotly_clickannotation` and, more generally, all `plotly_` events, should be triggered in `editable: true` mode.

Changing the issue title accordingly.

slfan2013:
Hi @etpinard, is there any update on this?

thanks.

PR #1895, opened 2017-07-18
Granular options for on-chart editing

### LINKED ISSUE #1872: Add support for the "config.editable" parameter as an object
The documentations states that the **editable** flag allows us to edit titles, move annotations, etc.
It would be nice to have the option to specify how editable my plot will be.

For example: I'd like to be able to edit annotation positions, or even their text, but I'd like my axis labels and plot title to remain fixed.

alexcjohnson:
Adds a new config option `edits` that's an object of all the separate editable interactions you want to enable or disable. The original `config.editable` should still be a boolean (default `false` as always) that sets the defaults for *all* our supported editable interactions, and `config.edits` can override pieces of it. For example if you *only* want to enable editing legend entries, and axis titles and moving the legend, you could do:

```
Plotly.newPlot(gd, data, layout, {
    edits: {legendText: true, axisTitleText: true, legendPosition: true}
})
```

closes #1872

cc @etpinard @geocosmite

I chose to be fairly granular - for example, annotations have three separate flags (`annotationText`, `annotationPosition`, and `annotationTail`, on the idea that for example you might want to allow someone to move an annotation but not change the length/direction of the arrow) but there are some other things we could segment further if desired (`axisTitleText` could be replaced by `xaxisTitleText` and `yaxisTitleText` for example). Did I strike the right balance?

### DIFF
--- a/src/components/annotations/draw.js
+++ b/src/components/annotations/draw.js
@@ -85,6 +85,8 @@ function drawOne(gd, index) {
 function drawRaw(gd, options, index, subplotId, xa, ya) {
     var fullLayout = gd._fullLayout;
     var gs = gd._fullLayout._size;
+    var edits = gd._context.edits;
+
     var className;
     var annbase;
 
@@ -128,8 +130,11 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
     var annTextGroup = annGroup.append('g')
         .classed('annotation-text-g', true);
 
+    var editTextPosition = edits[options.showarrow ? 'annotationTail' : 'annotationPosition'];
+    var textEvents = options.captureevents || edits.annotationText || editTextPosition;
+
     var annTextGroupInner = annTextGroup.append('g')
-        .style('pointer-events', options.captureevents ? 'all' : null)
+        .style('pointer-events', textEvents ? 'all' : null)
         .call(setCursor, 'default')
         .on('click', function() {
             gd._dragging = false;
@@ -519,7 +524,7 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
 
             // the arrow dragger is a small square right at the head, then a line to the tail,
             // all expanded by a stroke width of 6px plus the arrow line width
-            if(gd._context.editable && arrow.node().parentNode && !subplotId) {
+            if(edits.annotationPosition && arrow.node().parentNode && !subplotId) {
                 var arrowDragHeadX = headX;
                 var arrowDragHeadY = headY;
                 if(options.standoff) {
@@ -601,7 +606,7 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
         if(options.showarrow) drawArrow(0, 0);
 
         // user dragging the annotation (text, not arrow)
-        if(gd._context.editable) {
+        if(editTextPosition) {
             var update,
                 baseTextTransform;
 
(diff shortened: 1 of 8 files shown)


src/plot_api/plot_api.js
```diff
@@ -397,7 +397,7 @@ function opaqueSetBackground(gd, bgColor) {
 }
 
 function setPlotContext(gd, config) {
-    if(!gd._context) gd._context = Lib.extendFlat({}, Plotly.defaultConfig);
+    if(!gd._context) gd._context = Lib.extendDeep({}, Plotly.defaultConfig);
```
etpinard: Nice. Maybe `config` isn't as poorly tested than I thought.

src/components/legend/draw.js
```diff
@@ -750,22 +717,23 @@ function computeLegendDimensions(gd, groups, traces) {
...
-                (gd._context.editable ? 0 : opts.width),
-                legendItem.height
-            );
-        });
-    }
+    traces.each(function(d) {
+        var legendItem = d[0],
+            bg = d3.select(this).select('.legendtoggle');
+
+        bg.call(Drawing.setRect,
```
etpinard: lovely commit :tada:

src/components/annotations/draw.js
```diff
@@ -128,8 +130,11 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
     var annTextGroup = annGroup.append('g')
         .classed('annotation-text-g', true);
 
+    var editTextPosition = edits[options.showarrow ? 'annotationTail' : 'annotationPosition'];
+    var textEvents = options.captureevents || edits.annotationText || editTextPosition;
```
etpinard: No need to address this now, but this seems like the right place to reference https://github.com/plotly/plotly.js/issues/1750

test/jasmine/tests/config_test.js
```diff
@@ -193,77 +198,157 @@ describe('config argument', function() {
...
+                    colorbarPosition: true,
+                    colorbarTitleText: true,
+                    legendPosition: true,
+                    legendText: true,
+                    shapePosition: true,
+                    titleText: true
+                });
+            })
+            .then(function() {
+                return Plotly.newPlot(gd, data, layout, {
```
etpinard: For completeness, calling `Plotly.plot` instead of `Plotly.newPlot` would do the same here?

test/jasmine/tests/config_test.js
```diff
@@ -181,10 +181,15 @@ describe('config argument', function() {
 
         var gd;
 
-        beforeEach(function(done) {
+        beforeEach(function() {
             gd = createGraphDiv();
+        });
+
+        function initPlot(editFlag) {
```
etpinard: Nice tests :tada:

etpinard:
> (axisTitleText could be replaced by xaxisTitleText and yaxisTitleText for example). Did I strike the right balance?

I think so :+1:

etpinard:
:dancer:

test/jasmine/tests/config_test.js
alexcjohnson: yes, I guess since we [don't remove `gd._context` on purge](https://github.com/plotly/plotly.js/blob/e3eba8a7989d9dfd4e54fa5c6ca147bfff9e8a5e/src/plots/plots.js#L1280-L1281) I could omit `data` and `layout` here and that would be equivalent, just changing the config. I've just been tending to avoid `Plotly.plot` since its effects on existing plots are a little strange.

PR #1897, opened 2017-07-19
better zooming behavior for errorbars, bars, and box plots

### LINKED ISSUE #1879: Set non-scaling-stroke effect for error bars
... just like @rreusser 's https://github.com/plotly/plotly.js/pull/761 did for scatter lines,

See also https://github.com/plotly/plotly.js/issues/760 for more info on the topic.

alexcjohnson:
fixes #1879 and a few related effects:
- non-scaling stroke on errorbars, bar outlines, and box plots
- include box plot points in the rescaling we do for scatter points

As far as I can tell that's everything we put on SVG cartesian axes.

Not sure if there's a useful way to test this - I was pondering making a complicated plot and testing that every `<path>` on it either has `vector-effect: non-scaling-stroke` set or matches `.scatterlayer .points .point` or `.boxlayer .points .point` but I kind of feel like that will cause more problems than it solves, and now that we know about these strategies we won't forget to use them when we make new trace types.

cc @etpinard

### DIFF
--- a/src/components/errorbars/plot.js
+++ b/src/components/errorbars/plot.js
@@ -89,6 +89,7 @@ module.exports = function plot(traces, plotinfo, transitionOpts) {
 
                 if(isNew) {
                     yerror = errorbar.append('path')
+                        .style('vector-effect', 'non-scaling-stroke')
                         .classed('yerror', true);
                 } else if(hasAnimation) {
                     yerror = yerror
@@ -117,6 +118,7 @@ module.exports = function plot(traces, plotinfo, transitionOpts) {
 
                 if(isNew) {
                     xerror = errorbar.append('path')
+                        .style('vector-effect', 'non-scaling-stroke')
                         .classed('xerror', true);
                 } else if(hasAnimation) {
                     xerror = xerror
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -747,7 +747,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 .call(Drawing.setTranslate, clipDx, clipDy)
                 .call(Drawing.setScale, xScaleFactor2, yScaleFactor2);
 
-            var scatterPoints = subplot.plot.select('.scatterlayer').selectAll('.points');
+            var scatterPoints = subplot.plot.selectAll('.scatterlayer .points, .boxlayer .points');
 
             subplot.plot
                 .call(Drawing.setTranslate, plotDx, plotDy)
--- a/src/traces/bar/plot.js
+++ b/src/traces/bar/plot.js
@@ -124,8 +124,10 @@ module.exports = function plot(gd, plotinfo, cdbar) {
                     // append bar path and text
                     var bar = d3.select(this);
 
-                    bar.append('path').attr('d',
-                        'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z');
+                    bar.append('path')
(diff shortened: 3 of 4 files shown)


src/components/errorbars/plot.js
```diff
@@ -117,6 +118,7 @@ module.exports = function plot(traces, plotinfo, transitionOpts) {
 
                 if(isNew) {
                     xerror = errorbar.append('path')
+                        .style('vector-effect', 'non-scaling-stroke')
```
etpinard: Something still looks off:

peek 2017-07-19 11-15

Looks like the whiskers appear thicker during zooms.

src/plots/cartesian/dragbox.js
```diff
@@ -747,7 +747,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 .call(Drawing.setTranslate, clipDx, clipDy)
                 .call(Drawing.setScale, xScaleFactor2, yScaleFactor2);
 
-            var scatterPoints = subplot.plot.select('.scatterlayer').selectAll('.points');
+            var scatterPoints = subplot.plot.selectAll('.scatterlayer .points, .boxlayer .points');
```
etpinard: Nice touch :ok_hand:

etpinard:
> Not sure if there's a useful way to test this - 

I can't think of one either. I'm ok leaving this thing untested.

src/components/errorbars/plot.js
alexcjohnson: right - because that's about the *length* of the whisker, not its stroke. I couldn't think of a way to handle this at first but now that you mention it, I suppose we could have the `x` scale of the `y` errorbars (and the `y` scale of `x` error bars) get inverse-scaled the same way as we do for both x and y scales of the points... I can look into this but it may take a little more work, do you think it's worth holding up this PR or shall I tackle it separately later?

etpinard: > do you think it's worth holding up this PR or shall I tackle it separately later?

Not worth holding up this PR. :+1:  

Opening a new issue will be just fine.

alexcjohnson:
> should we scale back contour labels that same what we do for scatter text nodes?

We could consider it. The space available for the labels will not scale back though, so when you zoom out the labels will start to overlap the lines, and when you zoom in they'll get disconnected.

src/traces/bar/plot.js
```diff
@@ -124,8 +124,10 @@ module.exports = function plot(gd, plotinfo, cdbar) {
                     // append bar path and text
                     var bar = d3.select(this);
 
-                    bar.append('path').attr('d',
-                        'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z');
+                    bar.append('path')
```
etpinard: Non- :no_entry: comment 

----------

What about bar text labels?

peek 2017-07-19 11-39

they probably would benefit from the same [tricks](https://github.com/plotly/plotly.js/blob/97d08c8a126e9dd84e95acc8c8edc655eb95ed28/src/plots/cartesian/dragbox.js#L763) scatter text nodes get on scroll

alexcjohnson: These are also tricky - as sometimes they will scale up, sometimes they won't, and sometimes they'll move or rotate after the zoom. Also I played around with including them and the element structure isn't quite set up the same way as scatter text points so it's nontrivial to make it work. I'll add this to the list of possible future improvements.

alexcjohnson:
open items collected in https://github.com/plotly/plotly.js/issues/1899

etpinard:
Solid improvements :dancer:

PR #1898, opened 2017-07-19
smarter tick count limit for cartesian axes

### LINKED ISSUE #484: Axis labels on very tall graph don't reach end of axis

alexcjohnson:
fixes #484 

@etpinard what do you think about this solution: Stopping if two consecutive ticks are truly identical (infinite loop) or if we make more ticks than there are pixels on the axis. I included the same 1000 for short axes and as a fallback in case there are callers with dummy axes that don't have a `_length`. But it seems to me that any case where you'd *really* want a large number of ticks, there will still be less than one per pixel.

### DIFF
--- a/test/jasmine/tests/axes_test.js
+++ b/test/jasmine/tests/axes_test.js
@@ -2225,6 +2225,38 @@ describe('Test axes', function() {
                 ['2000-01-01 11:00:00.0001', 'Jan 1, 2000, 11:00:00.0001']
             ]);
         });
+
+        it('avoids infinite loops due to rounding errors', function() {
+            var textOut = mockCalc({
+                type: 'linear',
+                tickmode: 'linear',
+                tick0: 1e200,
+                dtick: 1e-200,
+                range: [1e200, 2e200]
+            });
+
+            // This actually gives text '-Infinity' because it can't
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/axes_test.js
```diff
@@ -2225,6 +2225,38 @@ describe('Test axes', function() {
...
+                tickmode: 'linear',
+                tick0: 1e200,
+                dtick: 1e-200,
+                range: [1e200, 2e200]
+            });
+
+            // This actually gives text '-Infinity' because it can't
+            // calculate the first tick properly, but since it's not going to
+            // be able to do any better with the rest, we don't much care.
+            expect(textOut.length).toBe(1);
```
etpinard: Nice test!

etpinard:
> what do you think about this solution:

Great solution. Nicely done!

Now, https://codepen.io/etpinard/pen/WwajEX?editors=0010 is showing all its tick labels:

image

:tada:

etpinard:
:dancer:

PR #1901, opened 2017-07-20
copy histogram autobinx/y back to the input trace

### LINKED ISSUE #24: Histogram does not scale to the range of new data on restyle
If you set up a histogram...

```
var data1 = [1, 2, 2, 3, 3, 3, 4, 4, 5];
var hist = document.getElementById('hist_div');
Plotly.plot(hist, [{x: data1, type: 'histogram' }]);
```

... and then update this histogram with wider data followed by a restyle...

```
var data2 = [10, 20, 20, 30, 30, 30, 40, 40, 50];
Plotly.restyle(hist, {x: [data2]});
```

... the histogram keeps the old range (in this case 1 to 5) and does not scale to the range of the new data (which is the case when using scatter style plot). Here is a codepen showing this: http://codepen.io/rasmusab/pen/megKOo

I have not found any documentation regarding whether this is the expected behavior. If it is, what would be the best practice for updating a histogram with new data?

alexcjohnson:
on par with colors and axis ranges

fixes #24 

@etpinard are you OK with this?

### DIFF
--- a/src/traces/histogram/calc.js
+++ b/src/traces/histogram/calc.js
@@ -43,9 +43,11 @@ module.exports = function calc(gd, trace) {
     var pos0 = pa.makeCalcdata(trace, maindata);
 
     // calculate the bins
-    var binAttr = maindata + 'bins',
-        binspec;
-    if((trace['autobin' + maindata] !== false) || !(binAttr in trace)) {
+    var binAttr = maindata + 'bins';
+    var autoBinAttr = 'autobin' + maindata;
+    var binspec = trace[binAttr];
+    if((trace[autoBinAttr] !== false) || !binspec ||
+            binspec.start === null || binspec.end === null) {
         binspec = Axes.autoBin(pos0, pa, trace['nbins' + maindata], false, calendar);
 
         // adjust for CDF edge cases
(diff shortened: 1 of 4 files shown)


src/traces/histogram/calc.js
```diff
@@ -43,9 +43,11 @@ module.exports = function calc(gd, trace) {
...
 
     // calculate the bins
-    var binAttr = maindata + 'bins',
-        binspec;
-    if((trace['autobin' + maindata] !== false) || !(binAttr in trace)) {
+    var binAttr = maindata + 'bins';
+    var autoBinAttr = 'autobin' + maindata;
+    var binspec = trace[binAttr];
+    if((trace[autoBinAttr] !== false) || !binspec ||
+            binspec.start === null || binspec.end === null) {
```
alexcjohnson: note that [`cleanBins`](https://github.com/plotly/plotly.js/blob/bf1aa6ba015c304c8cf07f4027637ac2c23addd9/src/traces/histogram/calc.js#L40) provides a `size` even if there are no `start` or `end`. This also doesn't allow you to specify a `start` and no `end` or things like that. So *partial* bin definition might do strange things... but that I would definitely consider a feature we could propose to add, rather than a bug to fix.

Also note that `!(binAttr in trace)` in the old version never failed, since we default to `null` for `start` and `end` (which might have been important to enable `cleanBins`?) - so in principle I could omit `!binspec` here, just thought it prudent in case there's some context I'm not aware of that can call this `calc`.

etpinard:
> on par with colors and axis ranges

Nicely done and thanks for not forgetting about `histogram2d` :dancer:

Issue #179, opened 2016-01-13
This data makes the surface algorithm blow up


etpinard:
... and freeze the browser tab.

``` js
z =  [[9607345622458638.0, 9607345622458652.0, 9607345622458652.0, 9607345622458646.0, 9607345622458652.0, 9607345622458650.0, 9607345622458650.0, 9607345622458646.0], [9607345622458654.0, 9607345622458640.0, 9607345622458650.0, 9607345622458652.0, 9607345622458652.0, 9607345622458654.0, 9607345622458650.0, 9607345622458652.0], [9607345622458654.0, 9607345622458652.0, 9607345622458638.0, 9607345622458652.0, 9607345622458650.0, 9607345622458652.0, 9607345622458654.0, 9607345622458650.0], [9607345622458650.0, 9607345622458652.0, 9607345622458650.0, 9607345622458632.0, 9607345622458644.0, 9607345622458646.0, 9607345622458646.0, 9607345622458650.0], [9607345622458654.0, 9607345622458652.0, 9607345622458650.0, 9607345622458650.0, 9607345622458638.0, 9607345622458654.0, 9607345622458654.0, 9607345622458650.0], [9607345622458650.0, 9607345622458654.0, 9607345622458650.0, 9607345622458646.0, 9607345622458652.0, 9607345622458638.0, 9607345622458646.0, 9607345622458652.0], [9607345622458654.0, 9607345622458652.0, 9607345622458654.0, 9607345622458650.0, 9607345622458654.0, 9607345622458652.0, 9607345622458640.0, 9607345622458654.0], [9607345622458650.0, 9607345622458652.0, 9607345622458650.0, 9607345622458652.0, 9607345622458650.0, 9607345622458654.0, 9607345622458654.0, 9607345622458638.0]]
```

most likely due to unstable floating point errors.

If that's the case, we should at least warn user when this happens.

alexcjohnson:
#1906 fixes this for SVG, but there's something strange still happening with webGL - the browser still locks up with an unending cascade of these errors:
![screen shot 2017-07-21 at 4 56 06 pm](https://user-images.githubusercontent.com/2678795/28484291-833ed8a6-6e3f-11e7-951b-0a27cc5a0cc2.png)

archmoj:
A [demo](https://codepen.io/MojtabaSamimi/pen/vQzKBp) to illustrate the issue.

gvwilson:
[...]

Issue #970, opened 2016-09-22
colorbar draws outside the bounds


hellochar:
Hey Plotly team,

Small bug - sometimes, depending on the layout's height/vertical margins, one of the colorbar's `.cbfill` will get drawn outside the outline (see http://codepen.io/hellochar/pen/JREjyv):
![image](https://cloud.githubusercontent.com/assets/1079495/18761582/81d98fba-80d3-11e6-8e20-93bdbf37e359.png)

Looks like it's only off by 1 pixel:

![image](https://cloud.githubusercontent.com/assets/1079495/18761654/d42c56e4-80d3-11e6-9c06-0eb6433521af.png)

It is a noticeable artifact that sometimes appears in our charts. Any ideas what's going on here? Happy to submit a PR as well if someone could help me understand the code at 
https://github.com/plotly/plotly.js/blob/master/src/components/colorbar/draw.js#L348

Thanks,
Xiaohan

etpinard:
Thanks for the report.

andi1984:
When `yanchor: 'bottom'` is used it is off by far more pixels also overlapping with mode bar. So this mode is pretty much not usable at the moment. It should simply set `y=0` on this mode.

alexcjohnson:
@etpinard pointed out this issue when I mentioned a related one during #1906, so I might as well combine them and tackle them simultaneously:

> we're not doing crisp rendering of colorbars - particularly the outlines, but line levels should probably be crisp too

alexcjohnson:
Looks like the original issue is masked in v1.29.0 - I suspect the tweaks in #1854 result in the outline covering the out-of-place bar - but there's still a `y="-1"` bar sometimes so I suspect it could become visible in certain cases. Frankly the whole concept of constructing our colorbars from 256 overlapping bars (regardless of how many pixels tall the bar is) seems a bit silly, I'm going to try to replace it with a simple gradient fill.

rreusser:
@alexcjohnson This page: http://rickyreusser.com/plotly-mock-viewer/#contour_heatmap_coloring
Shows this:

<img width="348" alt="screen shot 2017-07-24 at 11 56 53" src="https://user-images.githubusercontent.com/572717/28539622-39b4d9f4-7067-11e7-8c48-1212880fdf78.png">

What do you see?

alexcjohnson:
[...]

PR #1906, opened 2017-07-21
Colorbar loops

alexcjohnson:
fixes *part* of #179 - it works now with heatmap (and in fact any SVG type that makes a colorbar, like bubble charts). The chart it makes looks a bit weird (the colorscale *should* be smooth, and the ticks *should* be evenly spaced), but I think that's probably a *good* thing, as it will alert the user that they're at the limit of javascript's precision.

screen shot 2017-07-21 at 4 37 12 pm

Resulted in a few sub-pixel changes in baselines, I think they're fine but I notice we're not doing crisp rendering of colorbars - particularly the outlines, but line levels should probably be crisp too - so I will make an issue for that look at this again afterward.


etpinard:
Nicely done :dancer: 

> I think they're fine but I notice we're not doing crisp rendering of colorbars - particularly the outlines, but line levels should probably be crisp too - so I will make an issue for that look at this again afterward.

... which might be related to https://github.com/plotly/plotly.js/issues/970

Issue #1907, opened 2017-07-21
showline: true while showticklabels: false breaks plot


cldougl:
example: https://codepen.io/plotly/pen/qXWmpY

both features work independently as expected and the above example seems to work in v 1.28.3

alecrosenbaum:
+1 also experiencing this issue, suddenly a bunch of my graphs broke

PR #1910, opened 2017-07-24
Axes line w/o ticklabels fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1907

which was broken since https://github.com/plotly/plotly.js/commit/eb48a9d15c7d52ee2703021c43ec441d692e0ecf where on this [line](https://github.com/plotly/plotly.js/commit/eb48a9d15c7d52ee2703021c43ec441d692e0ecf#diff-f6d5ed25a5d6f57c70063e1e1bbc407dR1702) `ax.tickfont` is assumed to exist - which isn't the case when `showticklabels` is [false](https://github.com/plotly/plotly.js/blob/10c3ae0f5105e6d617ede522d854112ca277f3cb/src/plots/cartesian/tick_label_defaults.js#L34).

Commit 3c26cbf first adds a mock inspired by https://codepen.io/plotly/pen/qXWmpY and generates a baseline using `v1.28.3` (i.e before the regression). 

Commit fe20694 proposes a fix where a fallback for `tickfont.size` is added, similar to another [fallback](https://github.com/plotly/plotly.js/blob/10c3ae0f5105e6d617ede522d854112ca277f3cb/src/plots/cartesian/axes.js#L760) in `axes.js`. Notice that commit fe20694 did generate a diff for the new mock, but looks like it is only of a few sub-pixels:

peek 2017-07-24 11-57

cc @alexcjohnson


alexcjohnson:
Thanks for ferreting this out @etpinard 

In the case where this fallback is needed, `labelStandoff` [doesn't actually get used](https://github.com/plotly/plotly.js/blob/fe206949211942b9c687bebc39b623cea8f869da/src/plots/cartesian/axes.js#L1783), right? So can we just include `&& ax.showticklabels` in the [conditional](https://github.com/plotly/plotly.js/blob/fe206949211942b9c687bebc39b623cea8f869da/src/plots/cartesian/axes.js#L1708)?

etpinard:
> In the case where this fallback is needed, labelStandoff doesn't actually get used, right? So can we just include && ax.showticklabels in the conditional?

Good call. Done in d951f6a

alexcjohnson:
💃  🎉

tasks/util/container_commands.js
```diff
@@ -59,11 +59,13 @@ function getRunCI(commands) {
...
     var commandsJoined = '"' + commands.join(' && ') + '"';
+    var containerId = '$(docker inspect --format \'{{.Id}}\' ' + constants.testContainerName + ')';
 
     return [
         'sudo',
-        'lxc-attach -n',
-        '$(docker inspect --format \'{{.Id}}\' ' + constants.testContainerName + ')',
+        'lxc-attach',
+        '-n', containerId,
+        '-f', '/var/lib/docker/containers/' + containerId + '/config.lxc',
```
etpinard: :speaker: Big shoutout to `hasimo` on CircleCI forums:

https://discuss.circleci.com/t/container-is-not-defined-on-lxc-attach/14684/5

alexcjohnson: 🥇 🎺  lets get back in business!

Issue #1893, opened 2017-07-18
Per-group names in groupby transform


rreusser:
The problem: When you group more than once trace, the legend is confusing. It shows repeated groups without a good indication of which trace they belong to:

<img width="69" alt="screen shot 2017-07-18 at 12 37 49" src="https://user-images.githubusercontent.com/572717/28336228-f39456da-6bb5-11e7-94db-84ce4c7a23a8.png">

More buggily, as @chriddyp [points out](https://github.com/plotly/streambed/issues/9691#issuecomment-316169810), if you change trace group names in the legend while in edit mode, they revert because there's nowhere in the schema for the input to go. One solution would be to add a `names` field for custom group names, e.g.:

```json
"type": "groupby",
"target": ["morning", "morning", "evening", "evening"],
"names": [
  {
    "value": "morning",
    "name": "kale - morning"
  },
  {
    "value": "evening",
    "name": "kale - evening"
  }
]
```

That at least makes room for the input and seems much less corner-case bound than combining the names into a single string automatically.

On a more UX level, it would be possible to show the legend group name (i.e. parent trace name) and group the traces visually, as in:

```
+- radishes ---+
| o morning    |
| x evening    |
+--------------+

+- carrots ----+
| o morning    |
| x evening    |
+--------------+

+- kale -------+
| o morning    |
| x evening    |
+--------------+
```

This solution resolves the UX issue but not the nowhere-for-editmode-input-to-go issue.

/cc @chriddyp @alexcjohnson @etpinard @bpostlethwaite @cldougl for thoughts.

rreusser:
I like @alexcjohnson's suggestion: `nameformat: "kale - %g"` that substitutes the group name into `%g`. The only problem is that we still don't have a place for edit mode input, though that could just be disabled.

alexcjohnson:
My first reaction would be to do something analogous to `tickvals`/`ticktext`:

```json
"type": "groupby",
"target": ["morning", "morning", "evening", "evening"],
"groupvals": ["morning", "evening"],
"groupnames": ["kale - morning", "kale - evening"]
```

A more concise but restricted way would be a custom string substitution:
```json
"type": "groupby",
"target": ["morning", "morning", "evening", "evening"],
"nameformat": "kale - %t"
```

But yeah, that wouldn't allow editing via the legend, that would have to break out of the substitution format and convert it to the `groupvals`/`groupnames` version.

rreusser:
[...]

PR #1919, opened 2017-07-27
Add ability to rename grouped traces

rreusser:
See #1893 

This PR adds the ability to rename grouped traces. The logic is as follows:

- provides `nameformat` as a `groupby` setting. The default is `%g (%t)`, which expands out into, for example, "groupName (traceName)". (This part is really easy to change. Feedback welcome.)
- You can provide overrides in `groupnames`

The result is that you can apply intelligent naming and also have the ability to override expanded trace names by modifying the legend.

To accomplish the `groupnames` part, I've implemented `Lib.keyedContainer`, which manages modifying and constructing updates for objects like `[{name: ..., value: ...}]`, which are becoming numerous and kind of unwieldy to manage. It's very similar to nestedProperty with the addition that it tracks changes and outputs them in the format `{'styles[0].name': ..., 'styles[0].value': ...}`, which are then easy to plugin into `Plotly.restyle`, as is done when changing legend names.

## To do:
- [x] I've left `groupnames` as `type: "any"` because I need to figure out how to make this a properly nested container array.
- [x] change `[{name: ..., value: ...}]` to `[{group: ..., name: ...}]` to make it a bit more natural.
- [x] I'd love it if the `groupName (traceName)` format were only applied when there's more than one trace, but, unless I'm mistaken, we'd need to expose a bit more global state to transform supplyDefaults, which needs just a bit of thought to make sure we really can't avoid it. 😬 
- [x] Tests. (Holding off until confirmed that this is a good approach).

ezgif-2-6729e8cf33

/cc @etpinard @alexcjohnson


rreusser:
On usage of the fancy container array thing. It's basically used like this:

```javascript
var carr = Lib.keyedContainer(fullInput, 'transforms[' + i + '].groupnames');

if(BLANK_STRING_REGEX.test(text)) {
  carr.remove(legendItem.trace._group);
} else {
  carr.set(legendItem.trace._group, [text]);
}

Plotly.restyle(gd, carr.constructUpdate(), [traceNumber]);
```

When you add keys, it constructs the update. If you remove a key, it slides over all the other indices to keep the storage compressed (rather than a thousand undefined's because you kept modifying the plot).

Really, the whole goal is just to keep things schema-conforming without arbitrary keys. That may or may not be the right approach here, but this library function at least facilitates the process.

rreusser:
I've made my own requested changes to this, including internal cleanup and that it now shows the "groupName" if there's only one trace or "groupName (traceName)" if there's more than one trace. I had to pass `totalTraceCount` all the way through to transform supplyDefaults to accomplish this, which is ugly, though it does work. Very open to feedback on any of this PR (including just scrapping/rethinking it).

src/components/legend/draw.js
```diff
@@ -392,24 +393,57 @@ function drawTexts(g, gd) {
...
                 if(['ohlc', 'candlestick'].indexOf(fullInput.type) !== -1) {
-                    var transforms = legendItem.trace.transforms,
-                        direction = transforms[transforms.length - 1].direction;
+                    transforms = legendItem.trace.transforms;
+                    direction = transforms[transforms.length - 1].direction;
+
+                    update[direction + '.name'] = text;
+                } else if(fullInput.transforms) {
+                    for(i = fullInput.transforms.length - 1; i >= 0; i--) {
+                        if(fullInput.transforms[i].type === 'groupby') {
```
etpinard: Might be nice to add a lib function similar to [`Registry.traceIs`](https://github.com/plotly/plotly.js/blob/b2e54861dfe9960a7bb447f0a622d4198a4e22c7/src/registry.js#L150) for transforms e.g. `Registry,hasTransform('groupby')`

src/components/legend/draw.js
```diff
@@ -392,24 +393,57 @@ function drawTexts(g, gd) {
...
+
+                var p = Plotly.restyle(gd, update, traceIndex);
 
-                    astr = direction + '.name';
+                // If a groupby label is deleted, it seems like we need another redraw in order
+                // to restore the label. Otherwise it simply sets this property and the blank
+                // string is retained.
+                if(needsRedraw) {
+                    p = p.then(function() {
+                        return Plotly.redraw(gd);
```
etpinard: Hmm. We shouldn't have to do this. We should fix [`_restyle`](https://github.com/plotly/plotly.js/blob/b2e54861dfe9960a7bb447f0a622d4198a4e22c7/src/plot_api/plot_api.js#L1295) so that it leads to the correct update.

src/lib/keyed_container.js
```diff
@@ -0,0 +1,116 @@
...
+
+var nestedProperty = require('./nested_property');
+
+// bitmask for deciding what's updated:
+var NONE = 0;
+var NAME = 1;
+var VALUE = 2;
+var BOTH = 3;
+
+module.exports = function keyedContainer(baseObj, path, keyName, valueName) {
```
etpinard: Interesting. I like it. 

@rreusser can you of other situations where this could be useful?

src/plots/plots.js
```diff
@@ -798,7 +798,7 @@ plots.supplyDataDefaults = function(dataIn, dataOut, layout, fullLayout) {
 
     for(i = 0; i < dataIn.length; i++) {
         trace = dataIn[i];
-        fullTrace = plots.supplyTraceDefaults(trace, cnt, fullLayout, i);
+        fullTrace = plots.supplyTraceDefaults(trace, cnt, fullLayout, i, dataIn.length);
```
etpinard: I would prefer, stashing `dataIn.length` on `fullLayout` then passing it around like this.

Something like `fullLayout._dataInLength` that could be set [here](https://github.com/plotly/plotly.js/blob/b2e54861dfe9960a7bb447f0a622d4198a4e22c7/src/plots/plots.js#L453) side-by-side to `fullLayout._dataLength`.

src/plots/plots.js
```diff
@@ -809,7 +809,12 @@ plots.supplyDataDefaults = function(dataIn, dataOut, layout, fullLayout) {
 
             for(var j = 0; j < expandedTraces.length; j++) {
                 var expandedTrace = expandedTraces[j];
-                var fullExpandedTrace = plots.supplyTraceDefaults(expandedTrace, cnt, fullLayout, i);
+                var fullExpandedTrace = plots.supplyTraceDefaults(expandedTrace, cnt, fullLayout, i, dataIn.length);
+
+                // The group key gets cleared. If set, pass it forward
+                if(expandedTrace._group) {
+                    fullExpandedTrace._group = expandedTrace._group;
```
etpinard: Hmm. That's too bad. Do we _really_ need this `_group`, i.e. could _just_ recompute it during `Legend.draw`?

etpinard: Or maybe we could call [`relinkPrivateKeys`](https://github.com/plotly/plotly.js/blob/master/src/lib/relink_private.js) here, to make this solution generalizable?

src/transforms/groupby.js
```diff
@@ -35,6 +35,35 @@ exports.attributes = {
             'with `x` [1, 3] and one trace with `x` [2, 4].'
         ].join(' ')
     },
+    nameformat: {
```
etpinard: Bringing in @alexcjohnson @cldougl @chriddyp @cpsievert @monfera what do you think of this syntax?

A lot of existing attributes could benefit form _special formatting_ character like this one. For example,

```js
trace = {
  y: [1, 2, 1]
  marker: {
    size: [20, 30, 10]
  },
  hovertext: '%marker.size'
}
```

would be a nice shortcut to show each point's `marker.size` value on hover.

etpinard: Personally, I like `%`. Though maybe I'd vote for enforcing it with `{}` e.g. `%{name}` similar to ES6 `${name}`.

etpinard: ... and I would make user spell out `%{name}` and `%{group}` as opposed to e.g. `%t` and `%g`

src/components/legend/draw.js
rreusser: Agreed. I'll try to make that happen.

src/lib/keyed_container.js
rreusser: Yeah. It's also used in group styles, for one, though that case might be slightly more complicated. Also frames, though again, slightly more complicated. It's more a matter of realizing that unless we allow custom keys in the schema, abstracting this pattern seems like the only sane way forward.

src/plots/plots.js
rreusser: Sounds good.

src/plots/plots.js
rreusser: Yeah I didn't like it. I'll try to clean it up.

src/transforms/groupby.js
rreusser: I really like the `%{group}` format.

alexcjohnson: Yeah, this is cool! I hadn't thought about extending this to other keys but I can imagine that going off in all sorts of directions in the future (`%{marker.size*5}`???).

`%{...}` seems sufficiently useless on its own that we can do this without calling it a breaking change.

rreusser: Ah, I see. Like a general substitution framework? Yeah, the goal here was just a one-off custom substitution in which it *only* knows about a couple properties. But definitely possibilities. The only thing then would be choosing awkward names like `%{group}` and `%{name}` (instead of `%{trace}`) so that it's backward compatibility in case we substitute trace properties instead of special/custom/ad hoc keys.

src/components/legend/draw.js
rreusser: Added `Regsitry.hasTransform` and `Registry.getTransformIndices` to accomplish this. It's less efficient than the above code, but so marginally that I think it's fine. ✅

src/components/legend/draw.js
rreusser: I'm having trouble reproducing that this was ever necessary. It seems to update just fine. Giving this the 🔪 and I think it's fine.

src/plots/plots.js
rreusser: Oh, actually `layout._dataLength` *is* the length of the input data that I'm looking for. I think it's best not to make the count depend upon visibility (legend shouldn't fundamentally change when you toggle visibility, right?), so I think it's best to just use `layout._dataLength` for this.

src/transforms/groupby.js
rreusser: @alexcjohnson I've added `Lib.templateString` that performs these substitutions. At the moment it just grabs object properties if it's simple or uses `nestedProperty` if the key is more complicated than just a plain word. For this case the only data available is `%{trace}` and `%{group}`, but it at least opens the door to a unified interface for taking this further (or worst-case-scenario it's only a couple lines of code if it's not more generally useful). Feedback welcome.

rreusser: See: https://github.com/plotly/plotly.js/blob/57dfe4ac7ceb2281c71868f829a04e0a2ac994de/src/lib/index.js#L748

src/plots/plots.js
rreusser: Switched to `relinkPrivateKeys(fullExpandedTrace, expandedTrace)`, which seems to me like probably the most general approach here as long as it's not too expensive.

rreusser:
I've made the requested changes and have added a [test for editing the legend](https://github.com/plotly/plotly.js/pull/1919/commits/42d76e08aca4fc0cfda7c49cf33da4e76138d2ba#diff-8c755d7e7f3ece80edc9a49f41cb21a2R888) that includes regular and grouped traces. As soon as the tests pass, I don't think I have anything further to add.

test/jasmine/tests/register_test.js
```diff
@@ -277,4 +277,54 @@ describe('the register function', function() {
 
         expect(Registry.transformsRegistry['mah-transform']).toBeDefined();
     });
+
+    describe('getTransformIndices', function() {
```
etpinard: thanks for tests!

src/components/legend/draw.js
```diff
@@ -429,18 +427,7 @@ function drawTexts(g, gd) {
...
-                // to restore the label. Otherwise it simply sets this property and the blank
-                // string is retained.
-                if(needsRedraw) {
-                    p = p.then(function() {
-                        return Plotly.redraw(gd);
-                    });
-                }
-
-                return p;
+                return Plotly.restyle(gd, update, traceIndex);
```
etpinard: fun :+1:

src/transforms/groupby.js
```diff
@@ -113,7 +113,7 @@ exports.supplyDefaults = function(transformIn, traceOut, layout, traceIn, inputT
     if(!enabled) return transformOut;
 
     coerce('groups');
-    coerce('nameformat', inputTraceCount > 1 ? '%g (%t)' : '%g');
+    coerce('nameformat', layout._dataLength > 1 ? '%g (%t)' : '%g');
```
etpinard: This makes me very happy.

src/transforms/groupby.js
```diff
@@ -39,11 +39,11 @@ exports.attributes = {
         valType: 'string',
         description: [
             'Pattern by which grouped traces are named. If only one trace is present,',
-            'defaults to the group name (`"%g"`), otherwise defaults to the group name',
-            'with trace name (`"%g (%t)"`). Available escape sequences are `%g`, which',
-            'inserts the group name, and `%t`, which inserts the trace name. If grouping',
+            'defaults to the group name (`"%{group}"`), otherwise defaults to the group name',
+            'with trace name (`"%{group} (%{trace})"`). Available escape sequences are `%{group}`, which',
```
etpinard: Shouldn't this be `%{name}` instead of `%{trace}` so that it corresponds to an attribute just like `group`?

etpinard: ... but wait, `group` isn't a attribute either (`groups` is though). So how should we do this?

rreusser: Yeah, you're right. There are two options here:
1. Expose special sequences. That's what I've done here. So only `group` and `trace` are available. They're custom.
2. Make this a more general pattern of evaluating trace properties, in which case it would indeed be `%{name}` and… well, there's not really a property for the group unless we add it custom which is why I went with 1 here.

src/plots/plots.js
```diff
@@ -811,10 +811,7 @@ plots.supplyDataDefaults = function(dataIn, dataOut, layout, fullLayout) {
                 var expandedTrace = expandedTraces[j];
                 var fullExpandedTrace = plots.supplyTraceDefaults(expandedTrace, cnt, fullLayout, i);
 
-                // The group key gets cleared. If set, pass it forward
-                if(expandedTrace._group) {
-                    fullExpandedTrace._group = expandedTrace._group;
-                }
+                relinkPrivateKeys(fullExpandedTrace, expandedTrace);
```
etpinard: Maybe add a comment here e.g.:

> relink private (i.e. underscore) keys expanded trace to full expanded trace so that transform supply-default methods can set `_` keys for future use.

src/transforms/groupby.js
rreusser: Though to be honest, I'm starting to think we should intercept `%{group}` and %{name}` for this special case but also allow more general evaluation.

test/jasmine/tests/legend_test.js
```diff
@@ -882,4 +884,68 @@ describe('legend interaction', function() {
...
+            var item = d3.selectAll('text.legendtext')[0][index || 0];
+            item.dispatchEvent(new MouseEvent('click'));
+            return delay(20)().then(function() {
+                var input = d3.select('.plugin-editable.editable');
+                input.text(str);
+                input.node().dispatchEvent(new KeyboardEvent('blur'));
+            }).then(delay(20));
+        }
+
+        it('sets and unsets trace group names', function(done) {
```
etpinard: Lovely test.

src/transforms/groupby.js
etpinard: > I'm starting to think we should intercept %{group} and %{name} for this special case

I'm starting to think the same. :+1: for the current implementation.

src/plots/plots.js
rreusser: ✅

test/jasmine/tests/legend_test.js
rreusser: It's testing a fair amount of feature actions which isn't the *best*, but it's not bad for end-to-end.

src/transforms/groupby.js
rreusser: One small change though, right? Perhaps templateString should accept *multiple* objects from which to pull values and it takes the first one with a value. Then it `group` and `name` would be the first preference before grabbing `trace` properties. Specifically:

Now:
```javascript
Lib.templateString('%{group} (%{trace})', {group: ..., trace: ...})
```

Proposed:
```javascript
Lib.templateString('%{group} (%{name})', {group: ..., name: ...}, traceObj);
```

So that technically `%{marker.size}` would be equally valid.

etpinard: This plural vs singular problem came up too in https://github.com/plotly/plotly.js/pull/1770 :arrow_right:  [here](https://github.com/plotly/plotly.js/blob/b2e54861dfe9960a7bb447f0a622d4198a4e22c7/src/components/fx/helpers.js#L105-L106) - where values of _plural_ attributes are listed using a singular (grammatically) key.

Maybe we should start declaring corresponding _singular_ key in the plot schema and have `Lib.templateString` and `appendArrayPointValue` use it.

rreusser: Interesting. The only tricky part is that to be really precise, it would have to be `%{transforms[2].group}`, which it then links up with the schema to realize the pluralized key is in the schema and is a data array, and can be indexed and expanded to `%{transforms[2].groups[12]}`. The problem I see with that is that templateString must then receive the schema and an item index and that the key must contain the transform number. Correct me if I'm mistaken though.

etpinard:
@rreusser any updates on this? My https://github.com/plotly/plotly.js/pull/1919#discussion_r130960693 still stands as of today. 

We don't have to implement a the proposed full attribute-value-to-text-replacement scheme in this PR.

rreusser:
Sounds good @etpinard. Can you clarify that the comment you're referring to is that `%{group}` and `%{trace}` seem reasonable even though they're not attributes? Does that conflict with future-proofing this in case we *do* make them attributes? In other words, I'm just worried that if we make them special cases now, we'll never be able to remove them.

etpinard:
> that %{group} and %{trace} seem reasonable even though they're not attributes?

Exactly.

> Does that conflict with future-proofing this in case we do make them attributes?

`%{trace}` should never conflict (having attributes named `trace` in trace object would be very :confused: )

I guess `%{group}` _could_ conflict with e.g. `%{transforms[i].groups[j]` (not sure what the indices mean here), but to me, it's definitively a special that we should expose.

rreusser:
Thanks for clarifying @etpinard. If that's the case, then I think this is complete. You can see [here](https://github.com/plotly/plotly.js/pull/1919/files#diff-2c8b7c893b5eec3feb83ac0af6030ccfR221) that `group` and `trace` are currently the only attributes exposed though `templateString` is theoretically more powerful and its use *could* be expanded (though if not, it's only a couple lines of code anyway so not a big deal).

rreusser:
Confirming 💃ability then, @etpinard / @alexcjohnson ?

src/transforms/groupby.js
```diff
@@ -35,6 +35,35 @@ exports.attributes = {
...
+        description: [
+            'Pattern by which grouped traces are named. If only one trace is present,',
+            'defaults to the group name (`"%{group}"`), otherwise defaults to the group name',
+            'with trace name (`"%{group} (%{trace})"`). Available escape sequences are `%{group}`, which',
+            'inserts the group name, and `%{trace}`, which inserts the trace name. If grouping',
+            'GDP data by country when more than one trace is present, for example, the',
+            'default "%{group} (%{trace})" would return "Monaco (GDP per capita)".'
+        ].join(' ')
+    },
+    groupnames: {
```
etpinard: Hmm. I missed this during my first review. I'm not a fan of `groupnames` being an array container. 

We could instead declare `groupnames` as a `data_array` and expecting its length to be equal to `groups.length`, but that would lead to some (maybe a lot of) duplication. 

Maybe we could `groupname` attribute to the `styles` container below instead which already expects a target group?

Actually, do we even need a `groupnames` attribute? `groups` items can be any strings, so setting `groupnames[i].name` is equivalent to changing the corresponding `groups` item.

I apologize for not catching this sooner.

alexcjohnson: > setting `groupnames[i].name` is equivalent to changing the corresponding `groups` item.

Not quite - the name will be set by `nameformat` - ie a combination of the group value and the trace name by default - but when you want to override that (perhaps by typing something specific in the editable legend) you'll need this.

> We could instead declare `groupnames` as a `data_array` and expecting its length to be equal to `groups.length`, but that would lead to some (maybe a lot of) duplication.

Or ambiguity, if the names were not perfectly duplicated. What would that mean?

> Maybe we could `groupname` attribute to the `styles` container below instead which already expects a target group?

That sounds great! so each entry in `styles` would have `target` and optionally `value` and/or `name`? No need to have two separate containers both keyed on the group value.

rreusser: The main problem is that `groups` have no correspondence with `groupnames`, which I think rules out a plain array for the API. I think there's a good argument to be made though for merging `styles` with `groupnames`. Actually in hindsight, that does seem best.

etpinard: > That sounds great! so each entry in styles would have target and optionally value and/or name?

Exactly. I'd vote for `name` over `value`, but I don't have a strong opinion here.

alexcjohnson: `value` is all the other styles... actually, would it work to just put `name` *inside* `value`, and not have to create *any* new API? We'd still of course have to point the editing routine at `styles[i].value.name` but seems like that's all, right?

rreusser: I think that's the best option. 👍 Then it really is just a style property (which could be set anyway and in fact we'd be doing something surprising *not* to use what's already there).

rreusser: FWIW, IMO, TBH, I'm even happier now that I took the time to implement keyedContainer…

etpinard: > would it work to just put name inside value

good call :+1:

src/lib/keyed_container.js
```diff
@@ -61,7 +64,11 @@ module.exports = function keyedContainer(baseObj, path, keyName, valueName) {
                 nestedProperty(newValue, valueName).set(value);
             }
 
-            arr[idx] = newValue;
+            // If it's not an unset, force that bit to be unset. This is all related to the fact
+            // that undefined and null are a bit specially implemented in nestedProperties.
+            if(value !== null) {
+                changeType = changeType & ~UNSET;
```
etpinard: What does `& ~null` do?

etpinard:
Well-earned :dancer: @rreusser 

Before merging, would you mind shedding some light about those bitwise operations (in a PR comment or in the code directly, whatever you prefer). Also, could you update the PR header (which still mentions `groupnames`) for us potentially looking over this PR again in the future?

src/lib/keyed_container.js
rreusser: I'm tracking what needs to be updated with a bit mask (which, before adding a bit to track whether it's unset in a way that `nestedProperty` needs to handle specially, originally made 10% more sense since everything else was strictly an OR operation). & ~4 unsets precisely the 4 bit.

rreusser:
I'm having second thoughts about storing `name` inside `styles` because removing `styles` is how the UI reverts a group's styles to the defaults. That means reverting to default styles will unset the name, which suggests the nesting is a bit incorrect. I think it might be better to store the name *next to* the styles, which would then allow them to operate separately.

rreusser:
After a brief offline conversation with @alexcjohnson, I have to agree that the current API seems best and that it should be a concern of, for example, the UI to construct the proper update that doesn't unset the name accidentally. With that then, I'm content.

Below is the last case that was rather tricky. I fixed it by just leaving unused container in place rather than compressing the group styles.

ezgif-4-24052e288f

src/components/legend/draw.js
```diff
@@ -392,24 +393,41 @@ function drawTexts(g, gd) {
...
+
+                    update[direction + '.name'] = text;
+                } else if(Registry.hasTransform(fullInput, 'groupby')) {
+                    var groupbyIndices = Registry.getTransformIndices(fullInput, 'groupby');
+                    var index = groupbyIndices[groupbyIndices.length - 1];
+
+                    var carr = Lib.keyedContainer(fullInput, 'transforms[' + index + '].styles', 'target', 'value.name');
 
-                    astr = direction + '.name';
+                    if(BLANK_STRING_REGEX.test(origText)) {
```
alexcjohnson: hmm, there *is* a use case for a blank name - maybe less so for groupby than for regular traces, but sometimes you want the title for one trace to stand in for a few traces below it. You could argue then that this should just be testing for an actual empty string `''` but a space `' '` should be allowed as the name.

rreusser: okay, I could just check `=== ''` instead.

rreusser: ✅

alexcjohnson:
💃  (again 😄)

rreusser:
Okay, merging this at last. This API seems correct this way, though it makes the updates a bit tricky to compute. Improvements like compressing unused groups out of the picture would be a nice, backwards-compatible enhancement.

PR #1924, opened 2017-08-02
Aggregate transforms

alexcjohnson:
Adds an `aggregate` transform type, which is actually a closer analog of sql `groupby` than the `groupby` transform is because it coalesces all values in each group to a single output, and leaves the results all in a single trace. Each array attribute needs an aggregation function; if not provided we fall back on `"first"` ie the first value encountered in the group.

Note that some data types / aggregation function combinations don't make much sense, but we allow them anyway: date sums (we add milliseconds since 1970 and then convert back to a date), category sums (we add category serial numbers), category averages (we average serial numbers then round). The tests include an example to show how this plays out.

I did NOT implement any sort of binning in this PR. That will be a separate PR so it can add binning to `groupby` as well.

cc @rreusser @etpinard

### DIFF
--- a/src/transforms/aggregate.js
+++ b/src/transforms/aggregate.js
@@ -79,8 +79,17 @@ var attrs = exports.attributes = {
                 'for example a sum of dates or average of categories.',
                 '*median* will return the average of the two central values if there is',
                 'an even count. *mode* will return the first value to reach the maximum',
-                'count, in case of a tie. *stddev* uses the population formula',
-                '(denominator N, not N-1)'
+                'count, in case of a tie.'
+            ].join(' ')
+        },
+        funcmode: {
+            valType: 'enumerated',
+            values: ['sample', 'population'],
+            dflt: 'sample',
+            role: 'info',
+            description: [
+                '*stddev* supports two formula variants: *sample* (normalize by N-1)',
+                'and *population* (normalize by N).'
             ].join(' ')
         },
         enabled: {
@@ -148,17 +157,24 @@ exports.supplyDefaults = function(transformIn, traceOut) {
 
     var aggregationsIn = transformIn.aggregations;
     var aggregationsOut = transformOut.aggregations = new Array(aggregationsIn.length);
+    var aggregationOut;
+
+    function coercei(attr, dflt) {
+        return Lib.coerce(aggregationsIn[i], aggregationOut, aggAttrs, attr, dflt);
+    }
 
     if(aggregationsIn) {
         for(i = 0; i < aggregationsIn.length; i++) {
-            var aggregationOut = {};
-            var target = Lib.coerce(aggregationsIn[i], aggregationOut, aggAttrs, 'target');
-            var func = Lib.coerce(aggregationsIn[i], aggregationOut, aggAttrs, 'func');
-            var enabledi = Lib.coerce(aggregationsIn[i], aggregationOut, aggAttrs, 'enabled');
+            aggregationOut = {};
+            var target = coercei('target');
+            var func = coercei('func');
+            var enabledi = coercei('enabled');
 
             // add this aggregation to the output only if it's the first instance
             // of a valid target attribute - or an unused target attribute with "count"
             if(enabledi && target && (arrayAttrs[target] || (func === 'count' && arrayAttrs[target] === undefined))) {
+                if(func === 'stddev') coercei('funcmode');
+
                 arrayAttrs[target] = 0;
                 aggregationsOut[i] = aggregationOut;
             }
@@ -225,7 +241,7 @@ function aggregateOneArray(gd, trace, groupings, aggregation) {
     var targetNP = Lib.nestedProperty(trace, attr);
     var arrayIn = targetNP.get();
     var conversions = Axes.getDataConversions(gd, trace, attr, arrayIn);
-    var func = getAggregateFunction(aggregation.func, conversions);
+    var func = getAggregateFunction(aggregation, conversions);
 
     var arrayOut = new Array(groupings.length);
     for(var i = 0; i < groupings.length; i++) {
@@ -234,7 +250,8 @@ function aggregateOneArray(gd, trace, groupings, aggregation) {
     targetNP.set(arrayOut);
 }
 
-function getAggregateFunction(func, conversions) {
+function getAggregateFunction(opts, conversions) {
+    var func = opts.func;
     var d2c = conversions.d2c;
     var c2d = conversions.c2d;
 
@@ -371,7 +388,11 @@ function getAggregateFunction(func, conversions) {
                 // is a number of milliseconds, and for categories it's a number
                 // of category differences, which is not generically meaningful but
                 // as in other cases we don't forbid it.
-                return Math.sqrt((total2 - (total * total / cnt)) / cnt);
+                var norm = (opts.funcmode === 'sample') ? (cnt - 1) : cnt;
+                // this is debatable: should a count of 1 return sample stddev of
+                // 0 or undefined?
+                if(!norm) return 0;
+                return Math.sqrt((total2 - (total * total / cnt)) / norm);
             };
     }
 }
--- a/src/transforms/aggregate.js
+++ b/src/transforms/aggregate.js
@@ -1,0 +1,283 @@
+/**
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Axes = require('../plots/cartesian/axes');
+var Lib = require('../lib');
+var PlotSchema = require('../plot_api/plot_schema');
+var BADNUM = require('../constants/numerical').BADNUM;
+
+exports.moduleType = 'transform';
+
+exports.name = 'aggregate';
+
+var attrs = exports.attributes = {
+    enabled: {
+        valType: 'boolean',
+        dflt: true,
+        description: [
+            'Determines whether this aggregate transform is enabled or disabled.'
+        ].join(' ')
+    },
+    groups: {
+        // TODO: groupby should support string or array grouping this way too
+        // currently groupby only allows a grouping array
+        valType: 'string',
+        strict: true,
+        noBlank: true,
+        arrayOk: true,
+        dflt: 'x',
+        description: [
+            'Sets the grouping target to which the aggregation is applied.',
+            'Data points with matching group values will be coalesced into',
+            'one point, using the supplied aggregation functions to reduce data',
+            'in other data arrays.',
+            'If a string, *groups* is assumed to be a reference to a data array',
+            'in the parent trace object.',
+            'To aggregate by nested variables, use *.* to access them.',
+            'For example, set `groups` to *marker.color* to aggregate',
+            'about the marker color array.',
+            'If an array, *groups* is itself the data array by which we aggregate.'
+        ].join(' ')
+    },
+    aggregations: {
+        _isLinkedToArray: 'style',
+        array: {
+            valType: 'string',
+            role: 'info',
+            description: [
+                'A reference to the data array in the parent trace to aggregate.',
+                'To aggregate by nested variables, use *.* to access them.',
+                'For example, set `groups` to *marker.color* to aggregate',
+                'about the marker color array.',
+                'The referenced array must already exist, unless `func` is *count*,',
+                'and each array may only be referenced once.'
+            ].join(' ')
+        },
+        func: {
+            valType: 'enumerated',
+            values: ['count', 'sum', 'avg', 'min', 'max', 'first', 'last'],
+            dflt: 'first',
+            role: 'info',
+            description: [
+                'Sets the aggregation function.',
+                'All values from the linked `array`, corresponding to the same value',
+                'in the `groups` array, are collected and reduced by this function.',
+                '*count* is simply the number of values in the `groups` array, so does',
+                'not even require the linked array to exist. *first* (*last*) is just',
+                'the first (last) linked value.'
+            ].join(' ')
+        },
+    }
+};
+
+/**
+ * Supply transform attributes defaults
+ *
+ * @param {object} transformIn
+ *  object linked to trace.transforms[i] with 'func' set to exports.name
+ * @param {object} traceOut
+ *  the _fullData trace this transform applies to
+ * @param {object} layout
+ *  the plot's (not-so-full) layout
+ * @param {object} traceIn
+ *  the input data trace this transform applies to
+ *
+ * @return {object} transformOut
+ *  copy of transformIn that contains attribute defaults
+ */
+exports.supplyDefaults = function(transformIn, traceOut) {
+    var transformOut = {};
+    var i;
+
+    function coerce(attr, dflt) {
+        return Lib.coerce(transformIn, transformOut, attrs, attr, dflt);
+    }
+
+    var enabled = coerce('enabled');
+
+    if(!enabled) return transformOut;
+
+    /*
+     * Normally _arrayAttrs is calculated during doCalc, but that comes later.
+     * Anyway this can change due to *count* aggregations (see below) so it's not
+     * necessarily the same set.
+     *
+     * For performance we turn it into an object of truthy values
+     * we'll use 1 for arrays we haven't aggregated yet, 0 for finished arrays,
+     * as distinct from undefined which means this array isn't present in the input
+     * missing arrays can still be aggregate outputs for *count* aggregations.
+     */
+    var arrayAttrArray = PlotSchema.findArrayAttributes(traceOut);
+    var arrayAttrs = {};
+    for(i = 0; i < arrayAttrArray.length; i++) arrayAttrs[arrayAttrArray[i]] = 1;
+
+    var groups = coerce('groups');
+
+    if(!Array.isArray(groups)) {
+        if(!arrayAttrs[groups]) {
+            transformOut.enabled = false;
+            return;
+        }
+        arrayAttrs[groups] = 0;
+    }
+
+    var aggregationsIn = transformIn.aggregations;
+    var aggregationsOut = transformOut.aggregations = [];
+
+    if(aggregationsIn) {
+        for(i = 0; i < aggregationsIn.length; i++) {
+            var aggregationOut = {};
+            var array = Lib.coerce(aggregationsIn[i], aggregationOut, attrs.aggregations, 'array');
+            var func = Lib.coerce(aggregationsIn[i], aggregationOut, attrs.aggregations, 'func');
+
+            // add this aggregation to the output only if it's the first instance
+            // of a valid array attribute - or an unused array attribute with "count"
+            if(array && (arrayAttrs[array] || (func === 'count' && arrayAttrs[array] === undefined))) {
+                arrayAttrs[array] = 0;
+                aggregationsOut.push(aggregationOut);
+            }
+        }
+    }
+
+    // any array attributes we haven't yet covered, fill them with the default aggregation
+    for(i = 0; i < arrayAttrArray.length; i++) {
+        if(arrayAttrs[arrayAttrArray[i]]) {
+            aggregationsOut.push({
+                array: arrayAttrArray[i],
+                func: attrs.aggregations.func.dflt
+            });
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+        dflt: 'x',
+        description: [
+            'Sets the grouping target to which the aggregation is applied.',
+            'Data points with matching group values will be coalesced into',
+            'one point, using the supplied aggregation functions to reduce data',
+            'in other data arrays.',
+            'If a string, *groups* is assumed to be a reference to a data array',
+            'in the parent trace object.',
+            'To aggregate by nested variables, use *.* to access them.',
+            'For example, set `groups` to *marker.color* to aggregate',
```
rreusser: Is there a difference between the way backticks and asterisks render in the docs?

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+    aggregations: {
+        _isLinkedToArray: 'style',
+        array: {
+            valType: 'string',
+            role: 'info',
+            description: [
+                'A reference to the data array in the parent trace to aggregate.',
+                'To aggregate by nested variables, use *.* to access them.',
+                'For example, set `groups` to *marker.color* to aggregate',
+                'about the marker color array.',
```
rreusser: Subtle, but maybe "to aggregate *over* the marker color array?"

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+                'The referenced array must already exist, unless `func` is *count*,',
+                'and each array may only be referenced once.'
+            ].join(' ')
+        },
+        func: {
+            valType: 'enumerated',
+            values: ['count', 'sum', 'avg', 'min', 'max', 'first', 'last'],
```
rreusser: Can't imagine median would be very popular… mode…? variance…?

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+     * Normally _arrayAttrs is calculated during doCalc, but that comes later.
+     * Anyway this can change due to *count* aggregations (see below) so it's not
+     * necessarily the same set.
+     *
+     * For performance we turn it into an object of truthy values
+     * we'll use 1 for arrays we haven't aggregated yet, 0 for finished arrays,
+     * as distinct from undefined which means this array isn't present in the input
+     * missing arrays can still be aggregate outputs for *count* aggregations.
+     */
+    var arrayAttrArray = PlotSchema.findArrayAttributes(traceOut);
```
rreusser: Of subtle interest is the fact that `arrayAttrs` *does* include transform data arrays themselves. I glossed over this groupby since it wasn't critically important for performance (iterated groupbys are definitely a corner case) and since the groups are already decided by the time this transform modifies its own `arrayAttrs` so that the result is correct. Perhaps it should filter `transform[i]` for `i > transformIn._index`…

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+                'For example, set `groups` to *marker.color* to aggregate',
+                'about the marker color array.',
+                'The referenced array must already exist, unless `func` is *count*,',
+                'and each array may only be referenced once.'
+            ].join(' ')
+        },
+        func: {
+            valType: 'enumerated',
+            values: ['count', 'sum', 'avg', 'min', 'max', 'first', 'last'],
+            dflt: 'first',
```
rreusser: Presumably any data array encountered by default just returns the first entry as a way of dealing with the issue of needing to specify operations for all data arrays?

test/jasmine/tests/transform_multi_test.js
```diff
@@ -446,6 +439,88 @@ describe('multiple transforms:', function() {
...
+        expect(trace1Out.y).toEqual([5]);
+
+        var trace2Out = gd._fullData[1];
+        expect(trace2Out.x).toEqual([4, -5]);
+        expect(trace2Out.y).toEqual([5, 4]);
+    });
+
+    it('always executes groupby before aggregate', function() {
+        // aggregate and groupby wouldn't commute, but groupby always happens first
+        // because it has a `transform`, and aggregate has a `calcTransform`
```
rreusser: Is this desirable? Should it be *possible* to group and then apply an aggregation to each group?

test/jasmine/tests/transform_aggregate_test.js
```diff
@@ -0,0 +1,190 @@
...
+            y: ['1995-01-15', '2005-03-15', '1990-12-23', '2001-01-01', 'not a date'],
+            text: ['2001-01-01 12:34', '2001-01-01 12:35', '2001-01-01 12:36', '2001-01-01 12:37', ''],
+            hovertext: ['a', '2001-01-02', '2001-01-03', '2001-01-04', '2001-01-05'],
+            customdata: ['2001-01', 'b', '2001-03', '2001-04', '2001-05'],
+            transforms: [{
+                type: 'aggregate',
+                // groups can be any type, but until we implement binning they
+                // will always compare as strings = so 1 === '1' === 1.0 !== '1.0'
+                groups: [1, 2, '1', 1.0, 1],
+                aggregations: [
```
rreusser: This would be a case where `Lib.keyedContainer` would be usable, though since we probably don't modify it internally, it probably serves no purpose. But for the workspace it could be possible to import it from plotly lib. It basically just helps you use this arrangement as if it's a direct key-value store.

jackparmer:
Perhaps @domoritz is interested in this as well.

domoritz:
Thanks for pinging me @jackparmer. Will there be a way to provide custom aggregators like user defined aggregates in databases?

src/transforms/aggregate.js
alexcjohnson: Ah good catch - not quite sure how it renders, but we've tried to use backticks for attribute names and asterisks for string attribute values, looks like I mixed them up a bit here.

src/transforms/aggregate.js
alexcjohnson: > By default, presumably any data array just returns the first entry as a way of dealing with the issue of needing to specify operations for all data arrays?

[exactly](https://github.com/plotly/plotly.js/pull/1924/files#diff-f12eb70a20dd2c3229f24292419ffe64R154)

src/transforms/aggregate.js
alexcjohnson: Ah interesting... I will have to look for cases where this matters, I can't quite tell offhand whether transforming the earlier transforms is merely unnecessary or can actually lead to bugs. Combine that with [groupby happening first](https://github.com/plotly/plotly.js/pull/1924#discussion_r130778479) and the condition is a bit more complicated than just `i > transformIn._index` anyway.

test/jasmine/tests/transform_multi_test.js
alexcjohnson: You mean the other way around - should it be possible to aggregate and then group? Yes of course, there are definitely use cases for it - say you want to do a `'count'` aggregation, then group by that count - one group for all entries with one sample, a separate group for entries with two samples, etc etc. But it seems like for the moment we have a deeper structural problem with that, so this test documents it.

alexcjohnson:
@domoritz 

> Will there be a way to provide custom aggregators like user defined aggregates in databases?

Unlikely - for portability & security we do not allow any code in the plot JSON. But I definitely would not be averse to adding more functions. Can you give me examples of the kind of thing people do with this? [All the custom aggregator docs](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-aggregate-transact-sql), seem to talk about "concatenate" - that we seems generally useful enough that we should add it (would require an additional parameter for the join string), @rreusser suggested some others (median, mode, variance), I could throw in RMS... what else? I know we're not going to cover all edge cases this way but we should try to grab all the semi-common ones.

I was also imagining adding some special ones for x/y/z coordinates to generate error bars automatically (min/mean/max or mean +/- std. dev).

src/transforms/aggregate.js
rreusser: All good. I didn't notice which was which. 😄

src/transforms/aggregate.js
rreusser: I think it's just a tiny bit of unnecessary work we can 🔪 at some point.

test/jasmine/tests/transform_multi_test.js
rreusser: Ah, correct. The other way around. 👍

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+function aggregateOneArray(gd, trace, groupings, aggregation) {
+    var attr = aggregation.array;
+    var targetNP = Lib.nestedProperty(trace, attr);
+    var arrayIn = targetNP.get();
+    var conversions = Axes.getDataConversions(gd, trace, attr, arrayIn);
+    var func = getAggregateFunction(aggregation.func, conversions);
+
+    var arrayOut = new Array(groupings.length);
+    for(var i = 0; i < groupings.length; i++) {
+        arrayOut[i] = func(arrayIn, groupings[i]);
```
rreusser: One more comment: is it necessary to split the arrays or would it be possible to use online algorithms to just make one pass without splitting into arrays? Like [mean and variance](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm), for example. Going through the list with online algorithms in mind:

- `count`: easy
- `sum`: easy
- `avg`: see: [Algorithms_for_calculating_variance](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm)
- `min`: easy
- `max`: easy
- `first`: easy
- `last`: easy, maybe?
- `variance`: see: [Algorithms_for_calculating_variance](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm)

alexcjohnson: Good point - I think for the moment I'll leave it as is, all the existing aggregations would be fairly easy to replicate online (and yes, `'last'` is easy - perhaps even easier than `'first'`, you just always replace the output with the new value) but some of the others we've listed in the comments might be trickier (median and mode, in particular). I'll keep this in mind though as a potential performance gain for later.

To note from a private convo with @rreusser - the case where this is most important is when you have a large number of small groups - thousands of groups aggregating just a couple of items each - in which case creating all the little `groupings` arrays entails significant overhead. If we see unreasonable drag in this case, switching to online algorithms would be the fix.

rreusser: Oh. Yeah. Easy. Obv.

src/transforms/aggregate.js
etpinard: For the record, this special attribute / value formatting isn't currently used anywhere. I originally though it could be on https://plot.ly/javascript/reference/, but no-one ever got to implementing it.

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+            'If a string, *groups* is assumed to be a reference to a data array',
+            'in the parent trace object.',
+            'To aggregate by nested variables, use *.* to access them.',
+            'For example, set `groups` to *marker.color* to aggregate',
+            'about the marker color array.',
+            'If an array, *groups* is itself the data array by which we aggregate.'
+        ].join(' ')
+    },
+    aggregations: {
+        _isLinkedToArray: 'style',
```
etpinard: For the record, the `_isLinkedToArray` values are used in the python api to build the _graph objects_ e.g. `go.Annotation`. So, we should change this line to `_isLinkedToArray: 'aggregation'`.

This won't do anything at the moment as transforms and anything inside them don't have corresponding python graph objects (I think) at the moment, but might as well stay consistent.

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+        array: {
```
etpinard: why not `target` as in `groupby` and `filter`?

src/transforms/aggregate.js
```diff
@@ -0,0 +1,283 @@
...
+        for(i = 0; i < aggregationsIn.length; i++) {
+            var aggregationOut = {};
+            var array = Lib.coerce(aggregationsIn[i], aggregationOut, attrs.aggregations, 'array');
+            var func = Lib.coerce(aggregationsIn[i], aggregationOut, attrs.aggregations, 'func');
+
+            // add this aggregation to the output only if it's the first instance
+            // of a valid array attribute - or an unused array attribute with "count"
+            if(array && (arrayAttrs[array] || (func === 'count' && arrayAttrs[array] === undefined))) {
+                arrayAttrs[array] = 0;
+                aggregationsOut.push(aggregationOut);
```
etpinard: So in general, `aggregationsIn.length !== aggregationsOut.length`. Hmm, that might cause issues in the workspace (but I sure you thought about that). Maybe we could instead add an `aggragations[i].enabled` attribute to make more similar to other `isLinkedToArray` items?

src/transforms/aggregate.js
alexcjohnson: oh haha copy/paste error - thanks.

test/jasmine/tests/transform_aggregate_test.js
```diff
@@ -0,0 +1,190 @@
...
+        Plotly.newPlot(gd, [{
+            x: ['2001-01-01', '', '2001-01-03', '2001-01-05', '2001-01-07'],
+            y: ['1995-01-15', '2005-03-15', '1990-12-23', '2001-01-01', 'not a date'],
+            text: ['2001-01-01 12:34', '2001-01-01 12:35', '2001-01-01 12:36', '2001-01-01 12:37', ''],
+            hovertext: ['a', '2001-01-02', '2001-01-03', '2001-01-04', '2001-01-05'],
+            customdata: ['2001-01', 'b', '2001-03', '2001-04', '2001-05'],
+            transforms: [{
+                type: 'aggregate',
+                // groups can be any type, but until we implement binning they
+                // will always compare as strings = so 1 === '1' === 1.0 !== '1.0'
```
etpinard: image

learn something new everyday.

test/jasmine/tests/transform_aggregate_test.js
```diff
@@ -0,0 +1,190 @@
...
+                }
+            },
+            transforms: [{
+                type: 'aggregate',
+                groups: ['a', 'b', 'a', 'a', 'a'],
+                aggregations: [
+                    // missing array - the entry is ignored
+                    {array: '', func: 'avg'},
+                    {array: 'x', func: 'sum'},
+                    // non-numerics will not count toward numerator or denominator for avg
```
etpinard: Maybe we could add a [`nansum`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.nansum.html) func down the road.

etpinard: It would be nice to add a line about this behavior in the attribute `description`.

alexcjohnson: or another attribute (that could apply to other functions too) for how to handle bad values.

test/jasmine/tests/transform_multi_test.js
etpinard: > we have a deeper structural problem

... to say the least. :confused:

src/transforms/aggregate.js
alexcjohnson: good idea. We're never going to be able to ensure equal in/out lengths, as we have to add entries for missing arrays - but we can at least ensure that all inputs contribute and the extras go at the end of the array so entries that *do* appear in the input have the same index in the output. added `enabled` -> https://github.com/plotly/plotly.js/pull/1924/commits/6d2d32cb43d260fa057d8630ae85d9c1f579c06d

test/jasmine/tests/transform_aggregate_test.js
alexcjohnson: expanded description along with adding a few more functions in https://github.com/plotly/plotly.js/pull/1924/commits/b6ccc01878ebf7edf164576932e2153d8c0088f5

src/transforms/aggregate.js
alexcjohnson: `array` -> `target` in https://github.com/plotly/plotly.js/pull/1924/commits/ffc4ee2c59d126420d748f757db8f0bce9c11123

rreusser: To recap:

In filter, `target` = data by which filter is applied
In sort, `target` = data by which data is sorted
In groupby, `groups` = data by which trace is grouped (should this be `target`?)
In aggregate, `target` = data by which trace is aggregated

test/jasmine/tests/transform_aggregate_test.js
```diff
@@ -0,0 +1,225 @@
...
+                line: {width: [1, 1, 2, 2, 1]}
+            },
+            transforms: [{
+                type: 'aggregate',
+                groups: [1, 2, 1, 1, 1],
+                aggregations: [
+                    {target: 'x', func: 'mode'},
+                    {target: 'y', func: 'median'},
+                    {target: 'marker.size', func: 'rms'},
+                    {target: 'marker.line.width', func: 'stddev'}
```
rreusser: Maybe:

`{target: 'marker.line.width', func: 'stddev', population: true | false}`

or


`{target: 'marker.line.width', func: 'stddev', sample: true | false}`

rreusser: Or @alexcjohnson suggests: `{target: 'marker.line.width', func: 'stddev', variant: 'population' | 'sample'}`

Are we going to want these config keys to go in that container and not conflict with anything else? Feels just a little loose to lump in their config with the parent container as opposed to 

```javascript
{
  target: 'marker.line.width',
  func: 'stddev',
  funcconfig: {
    variant: 'population' | 'sample'
  }
}
```

But that's just getting verbose for no reason.

src/transforms/aggregate.js
alexcjohnson: > In groupby, `groups` = data by which trace is grouped (should this be `target`?)

aggregate has *both* `groups` (in the main container) and `target` (in each `aggregation`) - I suppose we could make *all* of these be `target` but that's starting to sound a bit confusing.

rreusser: Ah, I thought maybe I was missing one. That works for me.

test/jasmine/tests/transform_aggregate_test.js
etpinard: Attribute _variants_ are often set under `*mode` in other contexts. I'd vote for:

```js
funcmode: 'population' | 'sample'
```

etpinard:
:dancer: for me

w/ or w/o `'population'` vs `'sample'` of https://github.com/plotly/plotly.js/pull/1924#discussion_r131023581 which we can always add in another PR.

test/jasmine/tests/transform_aggregate_test.js
rreusser: The only downside of `funcmode` might be that the enum values are coupled to `type`. That is if `concat` has `funcmode` too, then do we just set dynamic enum values? Does that affect the docs?

src/transforms/aggregate.js
```diff
@@ -371,7 +388,11 @@ function getAggregateFunction(func, conversions) {
                 // is a number of milliseconds, and for categories it's a number
                 // of category differences, which is not generically meaningful but
                 // as in other cases we don't forbid it.
-                return Math.sqrt((total2 - (total * total / cnt)) / cnt);
+                var norm = (opts.funcmode === 'sample') ? (cnt - 1) : cnt;
+                // this is debatable: should a count of 1 return sample stddev of
+                // 0 or undefined?
+                if(!norm) return 0;
+                return Math.sqrt((total2 - (total * total / cnt)) / norm);
```
rreusser: I felt compelled to verify the formula, though I trust it's correct. I ran it against [a simple online calculator](https://www.mathsisfun.com/data/standard-deviation-calculator.html) and copied the function into a super quick test: https://gist.github.com/rreusser/ffe0a6cea78dd153c6eea1d54ac11ea1

It gets a hearty 👍 from me. 👏

rreusser: Actually, make that 💃

test/jasmine/tests/transform_aggregate_test.js
alexcjohnson: This should be fine. Yes, we'll have to describe it in detail in `description`, and manage it dynamically in `supplyDefaults`, but that seems to me better than making different names for each `func` that gets varying `mode`s.

Issue #1923, opened 2017-08-02
Plot fails to sanitize invalid html strings


rreusser:
See: https://codepen.io/rsreusser/pen/xLOPGv

### Expected behavior:
1. plot is created
2. title is set to invalid string: `"failure<sub>2</sub></sub>"`
3. plot is restyled to valid string: `"success"`
4. plot has title `"success"`

### Observed behavior:
As soon as the title is made invalid in step 2, further updates don't work.

rreusser:
Ohhhh, I was being really silly here. It rejects the promise. Switching that to `catch` makes it succeed. Facepalm.

Nonetheless, I don't necessarily see why the update shouldn't just sanitize the input and do its best.

PR #1926, opened 2017-08-02
Ignore bare closing tags in html-like input

rreusser:
See #1923. This PR modifies svg text utils to ignore dangling closing tags when there's nothing on the stack.

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -402,7 +402,10 @@ function buildSVGText(containerNode, str) {
     function exitNode(type) {
         // A bare closing tag can't close the root node. If we encounter this it
         // means there's an extra closing tag that can just be ignored:
-        if(nodeStack.length === 1) return;
+        if(nodeStack.length === 1) {
+            Lib.log('Ignoring unexpected end tag </' + type + '>.', str);
+            return;
+        }
 
         var innerNode = nodeStack.pop();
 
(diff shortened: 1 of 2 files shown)


src/lib/svg_text_utils.js
```diff
@@ -400,7 +400,12 @@ function buildSVGText(containerNode, str) {
     }
 
     function exitNode(type) {
+        // A bare closing tag can't close the root node. If we encounter this it
+        // means there's an extra closing tag that can just be ignored:
+        if(nodeStack.length === 1) return;
```
alexcjohnson: do you want to log this like we do below for `if(type !== innerNode.type)`? Though, we don't do anything with this so maybe we should get rid of them both...

rreusser: That seems reasonable. Do those log messages show up on-screen? Does that seem desirable?

alexcjohnson: just in the console, but I think they also go to sentry when they happen on plot.ly

rreusser: Sounds good.

alexcjohnson: or maybe not... not really sure, I haven't looked there. But anyway `Lib.log` points to `console.trace` if it exists, otherwise it falls back on `console.log`

rreusser: Added message ✅

alexcjohnson:
Nice catch and nice fix! 💃

PR #1928, opened 2017-08-04
Fix toSVG for marker gradient legend items

etpinard:
fixes an issue-less bug I discovered while working on `image-exporter`.

See https://github.com/plotly/plotly.js/pull/1694 for more on the subject, this PR here applies the same fix to legend items.

cc @alexcjohnson


alexcjohnson:
Nice catch, nice fix! 💃

Issue #755, opened 2016-07-20
Hover label incorrect when number is very small (or large?)


rreusser:
**Live example**: http://codepen.io/rsreusser/pen/JKLGOL?editors=0010
(change min, max to see effect on when labels displayed correctly. I haven't figured out the pattern yet.)

**Expected behavior**: very small values show a hover label that represents the value. Shows scientific notation if beyond meaningful suffixes.

**Observed behavior**: Depending on the overall magnitudes involved, SI suffixes often just go missing. Which wouldn't be a big deal except for the fact that `1` when it should read `1n` is actually incorrect.

See SI prefixes here: http://www.npl.co.uk/reference/measurement-units/si-prefixes/

PR #1930, opened 2017-08-04
Better exponent treatment beyond SI prefixes

### LINKED ISSUE #1920: Inconistent numeric format on logarithmic axis labels
Hi all,

The following example shows inconsistent x-axis labels. 
Result: 10e+21, 100e+21, 1e+24, 10e+24, 100e+24, 1e+27
Expected: 1e+22, 1e+23, 1e+24, 1e+25, 1e+26, 1e+27

Can this be improved? Thanks!

```
<html>
<head> <script src="https://cdn.plot.ly/plotly-latest.min.js"></script> </head>
<body> 
<div id="TEST" class="plotly-graph-div"></div>
<script>
    window.PLOTLYENV=window.PLOTLYENV || {};
    window.PLOTLYENV.BASE_URL="https://plot.ly";
    Plotly.newPlot('TEST', [{"mode":"lines","y":[0.0,1.0],"type":"scatter","x":[1.0e22,1.0e27]}],
               {"xaxis":{"type":"log"},"margin":{"l":50,"b":50,"r":50,"t":60}}, {showLink: false});
</script>
</body>
</html>
```
CrossRef: [https://github.com/sglyon/PlotlyJS.jl/issues/132](https://github.com/sglyon/PlotlyJS.jl/issues/132)

alexcjohnson:
Fixes #1920 

In addition to no longer rounding the exponent to a factor of 3 when you get past the SI prefixes, I also changed it to `exponentformat: 'power'` there - ie "&times;10<sup>16</sup>" instead of "e+16" as we had before. I suspect using `exponentformat: 'e'` as the fallback was a decision made before we supported pseudo-html in tick labels.

screen shot 2017-08-04 at 5 30 13 pm

cc @etpinard @AndiMD

### DIFF
--- a/src/constants/numerical.js
+++ b/src/constants/numerical.js
@@ -47,5 +47,11 @@ module.exports = {
     /*
      * Are two values nearly equal? Compare to 1PPM
      */
-    ALMOST_EQUAL: 1 - 1e-6
+    ALMOST_EQUAL: 1 - 1e-6,
+
+    /*
+     * not a number, but for displaying numbers: the "minus sign" symbol is
+     * wider than the regular ascii dash "-"
+     */
+    MINUS_SIGN: '\u2212'
 };
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -1056,7 +1056,7 @@ function autoTickRound(ax) {
 
         var rangeexp = Math.floor(Math.log(maxend) / Math.LN10 + 0.01);
         if(Math.abs(rangeexp) > 3) {
-            if(ax.exponentformat === 'SI' || ax.exponentformat === 'B') {
+            if(isSIFormat(ax.exponentformat) && !beyondSI(rangeexp)) {
                 ax._tickexponent = 3 * Math.round((rangeexp - 1) / 3);
             }
             else ax._tickexponent = rangeexp;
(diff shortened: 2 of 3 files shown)


src/constants/numerical.js
```diff
@@ -47,5 +47,11 @@ module.exports = {
...
      * Are two values nearly equal? Compare to 1PPM
      */
-    ALMOST_EQUAL: 1 - 1e-6
+    ALMOST_EQUAL: 1 - 1e-6,
+
+    /*
+     * not a number, but for displaying numbers: the "minus sign" symbol is
+     * wider than the regular ascii dash "-"
+     */
+    MINUS_SIGN: '\u2212'
```
alexcjohnson: Currently this is only used in cartesian tick labels (and hover text) and this does get reused in other places... but I suspect there are places in other plot types where we use a regular dash that should be replaced by this one.

rreusser:
It's not quite related, but this reminded me of #755 and led to #1932. 🎉  The test I added for that PR will break when this is merged and will need to be updated to fancy exponent formatting.

Otherwise this PR looks good to me. 💃

src/plots/cartesian/axes.js
```diff
@@ -1359,6 +1361,21 @@ function formatLinear(ax, out, hover, extraPrecision, hideexp) {
...
+}
+
+// are we beyond the range of common SI prefixes?
+// 10^-16 -> 1x10^-16
+// 10^-15 -> 1f
+// ...
+// 10^14 -> 100T
+// 10^15 -> 1x10^15
+// 10^16 -> 1x10^16
+function beyondSI(exponent) {
```
etpinard: Beautiful.

etpinard:
Solid work. Thanks!

Issue #34, opened 2015-11-24
'marker+text' mode for bar charts


jackparmer:
Currently, we have to make annotations for bar charts because 'markers+text' mode (Reference: https://plot.ly/javascript/reference/#scatter-mode) does not extend to bar charts like this one:

https://plot.ly/javascript/bar-charts/#bar-chart-with-direct-labels

This can be tedious for a large number of bars, especially if someone is using the Web App. It would be nice if 'markers+text' mode existed for horizontal and vertical bar charts.

etpinard:
having `mode: 'makers+text'` for bar charts doesn't make sense for as `mode: 'text'` would be a degenerate text chart.

etpinard:
I would use `textposition` instead, similar to what we do for [pie charts](https://github.com/plotly/plotly.js/blob/master/src/traces/pie/attributes.js#L112).

jackparmer:
Ah, great points! That's a sweet idea!

chriddyp:
related: https://github.com/plotly/plotly.py/issues/356

cldougl:
Cool - thanks for the `textposition` tip @etpinard !

ryanlin1986:
This is a productivity boost.

alexcjohnson:
In fact more pieces of the pie chart text functionality should be replicated here - like `textposition='auto'` putting labels inside the bars if they fit (or if there's another bar stacked on top), and outside otherwise; shrinking the label if need be when it's displayed inside; `insidetextfont` and `outsidetextfont` so you can have differences (usually just different color, like white inside and black outside) in addition to `textfont` controlling them both.

etpinard:
Perhaps bar chart with `barmode: 'stacl'` should show totals: https://github.com/plotly/plotly.js/issues/85

n-riesco:
@jackparmer @cldougl The PR for #80 is almost completed. I'm planning to work on this issue now.

n-riesco:
@etpinard @alexcjohnson I have a PR that implements this feature ready [for review on my repo](https://github.com/n-riesco/plotly.js/pull/4).

etpinard:
Thanks @n-riesco I'll take at your PR before days end. Cheers :beers:

etpinard:
Done in https://github.com/plotly/plotly.js/pull/1159

PPPW:
Dear alexcjohnson and etpinard,
[...]

PR #1931, opened 2017-08-04
Bar text constraint configuration

rreusser:
This PR does two things:

1. adds `constraintext: 'both' | 'none' | 'inside' | 'outside'` in order to resolve [this comment](https://github.com/plotly/plotly.js/issues/34#issuecomment-320344424) on #34.
2. slightly modifies the `textpad` for outside text in that it keeps the padding but does not factor it into the width of the bar. That means the text is still offset from the bar just a bit but can grow slightly larger before it's constrained.


test/jasmine/tests/bar_test.js
```diff
@@ -95,6 +95,7 @@ describe('Bar.supplyDefaults', function() {
         expect(traceOut.texfont).toBeUndefined();
         expect(traceOut.insidetexfont).toBeUndefined();
         expect(traceOut.outsidetexfont).toBeUndefined();
+        expect(traceOut.constraintext).toBe();
```
alexcjohnson: This looks odd... `toBeUndefined`?

src/traces/bar/plot.js
```diff
@@ -335,23 +337,26 @@ function getTransformToMoveInsideBar(x0, x1, y0, y1, textBB, orientation) {
...
-function getTransformToMoveOutsideBar(x0, x1, y0, y1, textBB, orientation) {
+function getTransformToMoveOutsideBar(x0, x1, y0, y1, textBB, orientation, constrained) {
     var barWidth = (orientation === 'h') ?
             Math.abs(y1 - y0) :
             Math.abs(x1 - x0),
         textpad;
 
-    // apply text padding if possible
+    // Keep the padding so the text doesn't sit right against
+    // the bars, but don't factor it into barWidth
```
alexcjohnson: Ah, good call.

src/traces/bar/plot.js
```diff
@@ -335,23 +337,26 @@ function getTransformToMoveInsideBar(x0, x1, y0, y1, textBB, orientation) {
...
     if(barWidth > 2 * TEXTPAD) {
         textpad = TEXTPAD;
-        barWidth -= 2 * textpad;
     }
 
     // compute rotation and scale
-    var rotate = false,
+    var rotate = false;
```
alexcjohnson: This may be a silly question but... why do we have an `if(rotate)` below if `rotate = false` and is never changed?

test/jasmine/tests/bar_test.js
rreusser: Oops. Good catch.

src/traces/bar/plot.js
rreusser: Good catch. I'll clean it up.

alexcjohnson:
Looks great. I had also suggested expanding to fill the bar gap for outside text in stacked mode or with only one bar, but that seems a good deal tricker to implement and this is already a big win, so 💃

test/jasmine/tests/bar_test.js
rreusser: ✅

src/traces/bar/plot.js
rreusser: 🔪 'd

rreusser:
Ah, yeah. So just to confirm, does this win on pragmatism? I hate to introduce more attributes that don't have good longevity because they don't address the true problem.

Issue #1711, opened 2017-05-20
Hovervalues can be shown incorrectly with a large range of values


precisesimulation:
Sometimes with very small values the hoverdata shows incorrect value probably due to some truncation, rounding, or formating. Hovering over the right left point shows _10_ instead of _~0_

`Plotly.newPlot('myDiv',[{x: [0, 1],y: [1e-8, 1],type: 'scatter'}]);`

Note that it works fine with for example 

`Plotly.newPlot('myDiv',[{x: [0, 1],y: [1e-8, 2e-8],type: 'scatter'}]);`

when the _y_ values does not span a big range.

etpinard:
Probably a duplicate of https://github.com/plotly/plotly.js/issues/755

Related:

- https://github.com/plotly/plotly.js/issues/912
- https://github.com/plotly/plotly.js/issues/803
- https://github.com/plotly/plotly.js/issues/780

etpinard:
@rreusser can you confirm that https://github.com/plotly/plotly.js/pull/1932 also resolved this issue?

rreusser:
Confirmed!

<img width="150" alt="screen shot 2017-08-07 at 09 59 05" src="https://user-images.githubusercontent.com/572717/29037109-15ddb7e4-7b57-11e7-8fe9-688413343c1c.png">

PR #1932, opened 2017-08-04
Fix hover label exponents

### LINKED ISSUE #755: Hover label incorrect when number is very small (or large?)

rreusser:
This PR fixes #755 and #1711. After a bit of digging:

**The problem**: Hover labels are sometimes wrong by arbitrarily gigantic orders of magnitude, which is bad. Hover labels piggyback on `Axes.tickText`, which sometimes hides exponents and just shows 1/2/5 values. That's *always* invalid for hover labels since their exponent is not shown elsewhere as it is for axes.

**The solution**: whenever `'hover'` is passed to `tickText`, this PR prevents it from making any automated decision about whether to hide the exponent, instead just obeying the axis formatting.

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -1207,8 +1207,8 @@ axes.tickText = function(ax, x, hover) {
         return showAttr !== 'all' && x !== first_or_last;
     }
 
-    if (hover) {
-        hideexp = 'never'
+    if(hover) {
+        hideexp = 'never';
     } else {
         hideexp = ax.exponentformat !== 'none' && isHidden(ax.showexponent) ? 'hide' : '';
     }
@@ -1350,14 +1350,14 @@ function formatCategory(ax, out) {
 }
 
 function formatLinear(ax, out, hover, extraPrecision, hideexp) {
-    if (hideexp === 'never') {
+    if(hideexp === 'never') {
         // If this is a hover label, then we must *never* hide the exponent
         // for the sake of display, which could give the wrong value by
         // potentially many orders of magnitude. If hideexp was 'never', then
         // it's now succeeded by preventing the other condition from automating
         // this choice. Thus we can unset it so that the axis formatting takes
         // precedence.
-        hideexp = ''
+        hideexp = '';
     } else if(ax.showexponent === 'all' && Math.abs(out.x / ax.dtick) < 1e-6) {
         // don't add an exponent to zero if we're showing all exponents
         // so the only reason you'd show an exponent on zero is if it's the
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Looks good to me. I could imagine with `showexponent: 'none'` wanting the hover labels to have no exponents either - but that's a weird case and I could equally imagine wanting the exponents *only* on hover in that case. If you want that kind of control you can make your own hover text :)

This whole system of `formatLinear`/`formatLog`/`numFormat` is a bit of a palimpsest at this point and probably should be completely refactored... but for now this is great.

I took the liberty of merging #1930 first, so you can fix these tests 🎉  and then 💃

rreusser:
Haha good word. Yeah, the logic was about a 6/10 for subtle and intricate, so I was a bit worried I'd miss some cases. As is often the case then, I focused mainly on fixing the bug while adding and modifying as little as possible. I don't *quite* understand the above comment about `showexponent:  'none'` (like `0.0000000000000000112` then?) but seems like the top priority is that they're *not wrong*.

src/plots/cartesian/axes.js
```diff
@@ -1346,10 +1350,18 @@ function formatCategory(ax, out) {
...
-    // ONLY tick to get an exponent (first or last)
-    if(ax.showexponent === 'all' && Math.abs(out.x / ax.dtick) < 1e-6) {
+    if(hideexp === 'never') {
+        // If this is a hover label, then we must *never* hide the exponent
+        // for the sake of display, which could give the wrong value by
+        // potentially many orders of magnitude. If hideexp was 'never', then
+        // it's now succeeded by preventing the other condition from automating
+        // this choice. Thus we can unset it so that the axis formatting takes
+        // precedence.
+        hideexp = '';
```
etpinard: Thanks for the :books: @rreusser

etpinard:
Very nice PR @rreusser thanks for doing this :tada: 

:hocho:'ing two issues in ~10 new LOCs is amazing!

etpinard:
@rreusser this PR breaks [this](https://github.com/plotly/plotly.js/blob/4825f8ee0a6d3372c59c1fa6bcac390fccef7a1d/test/jasmine/tests/gl_plot_interact_test.js#L89) gl3d hover label test. CI :white_check_mark: as this particular test is skipped on CI. 

Can you make sure `npm run test-jasmine -- gl_plot_interact` passes?

rreusser:
Sure. En route to meeting with Alex today so will fix during the flight, along with the final trace naming issue.

Issue #1912, opened 2017-07-25
"Error: <path> attribute d" when traces have one element and fill tozerox


sporto:
When traces have only one element and fill is set to tozerox.

I see this error:

```
Error: <path> attribute d: Unexpected end of attribute. Expected number, "…673599549.5,135L".
```

See fiddle here:

https://codepen.io/sporto/pen/eEYwmW?editors=1010

Thanks

sporto:
Same error show if the trace has nulls

![image](https://user-images.githubusercontent.com/1005498/28550854-7486141a-7127-11e7-9f36-3e28079f953e.png)

rreusser:
@sporto I submitted #1933 to fix the behavior when either the first or last point is isolate. There's a still a bit of interpretation regarding what the fill should look like when it encounters isolated points (should it ignore them? pass through them?) but at least this way it's consistent and doesn't give errors.

rreusser:
@sporto #1933 has been merged and will be released mid-this-week with `1.30.0`, I believe. The error is gone and the fill now passes through single isolated points, which I think is as desired. Please let us know if you still have problems with this once it's released.

sporto:
Thanks, I will

Issue #1205, opened 2016-11-29
Filling two scatter traces without endpoints connection


sandertar:
If one scatter trace doesn't completely enclose the other, then the endpoints of each trace will be connected  with straight lines, and closed shape will be filled.
Are there any ways to fill each trace independently from other traces and close shape with vertical lines which perpendicular  to xaxis?
[http://codepen.io/sandertar/pen/YprYjm](http://codepen.io/sandertar/pen/YprYjm)

![selection_018](https://cloud.githubusercontent.com/assets/13880528/20698003/fd6b119c-b627-11e6-97e0-8240c1a58890.png)
![selection_021](https://cloud.githubusercontent.com/assets/13880528/20698430/2a698518-b62b-11e6-8363-39c094a177ad.png)

etpinard:
Thanks for the report.

Here's a workaround: http://codepen.io/etpinard/pen/RoLmdV

etpinard:
By the way, your codepen uses a _very very_ old version of plotly.js. You should try using

```js
<script type="text/javascript" src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```

instead in the future.

alexcjohnson:
FYI you're definitely right to consider this a bug @taranovalexander but there are several things that make this difficult to fix:

- What happens if the endpoint of `trace 1` does not match a data point in `trace 0` but rather between two data points? For straight line joins we can fairly easily find the intersection, but we will need to calculate it differently for each `line.shape` value (and `'spline'` is particularly annoying)
- What if `trace 0` is not monotonic in x? Which in fact can happen even if the `trace0` `x` *data* is monotonic with `line.shape='spline'`, but there are some truly horrible things that can happen if `x` is not monotonic.
- What if `trace 1` starts *before* `trace 0`? Then we need to fill down to zero, then up the side of `trace 0`... and this gets even more convoluted if `trace 1` starts after `trace 0` as in your example, but then you add a `trace 2` on top that starts before both of them.

One way we could think of solving this would be to draw these fills to zero but then clip them with the previous fills. That seems like it would give decent results - a bit weird looking in some of the non-monotonic cases but since those probably don't have much physical meaning we may not care.

[...]

PR #1933, opened 2017-08-05
Fix scatter fill with isolated endpoints

rreusser:
See: #1912 

This PR fixes scatter fill behavior in which single-point segments cause d3 errors and undesirable visual appearance when they're either the first or last line segments. It fixes it by simply drawing them as empty paths instead of cutting them out entirely.

**The good**: current behavior for isolated points is unaffected *unless* they're the first or last point

**The bad**: single points *do* affect the fill if they're the first or last segment but *do not* if they're interior points, which is a bit inconsistent.

**Question**: Should the fill go through isolated interior points? It's easy to accomplish but may change current instead of simply fixing endpoint cases.

### Input:

```json
"data": [ 
  {"x": [1.5], "y": [1.75], "fill": "tozeroy"},
  {"x": [0, 1, 2, 3], "y": [1.5, null, 1.5, 1.5], "fill": "tozeroy"},
  {"x": [0, 1, 1.5, 2, 2.5, 3], "y": [1.25, 1.25, null, 1.35, null, 1.25, 1.25], "fill": "tozeroy"},
  {"x": [0, 1, 2, 3], "y": [1.0, 1.1, null, 1.0], "fill": "tozeroy"},
  {"x": [0, 1.5, 3], "y": [0.75, null, 0.75], "fill": "tozeroy"}
]
```

### Before:

screen shot 2017-08-04 at 18 06 13

### After:

screen shot 2017-08-04 at 18 09 13


rreusser:
I've been thinking about this a bit more. There are a few options here, I guess. Perhaps the fill for isolated first/last segments should be drawn as actual isolated points that don't create any filled area. My solution here was that the first and last segments *always* box out some area, even if they are isolated, while interior isolated points do not. That's the inconsistency that bothers me just a bit. The bottom line though is that the current behavior just leads to errors, so I'm trying to decide whether it's worthwhile to aim for a slightly more nuanced fix or whether this solution solves it reasonably enough. Feedback appreciated.

alexcjohnson:
Lets just call it a bug in the interior case as well - if we consider this behavior to be as intended:
```
Plotly.newPlot(gd,[{y: [1, 3, null, 2, 4, null, 3, 5], fill: 'tozeroy'}], {width: 400, height: 400})
```
screen shot 2017-08-07 at 10 33 33 am

Then it makes little sense that dropping the middle segment to 1 point would disconnect it from the fill as it does now:
```
Plotly.newPlot(gd,[{y: [1, 3, null, 2, null, 3, 5], fill: 'tozeroy'}], {width: 400, height: 400})
```
screen shot 2017-08-07 at 10 35 14 am
It should really look like:
screen shot 2017-08-07 at 10 37 06 am

etpinard:
@rreusser @alexcjohnson not sure if you are aware of https://github.com/plotly/plotly.js/issues/1205, it might be related.

alexcjohnson:
#1205 is an independent, and trickier, issue - I'll add a comment over there.

rreusser:
I've pushed a modified version that simply removes `lineSegments` entirely rather than just using it differently. That is, empty lines segments are drawn as empty line segments instead of getting culled.

This changes my above baseline so that the single points *do* connect with the fill:

scatter_fill_corner_cases

rreusser:
With the most recent pushed code, it looks like this:

```json
{
  "y": [1, 3, null, 2, null, 3, 5],
  "fill": "tozeroy"
}
```

screen shot 2017-08-07 at 10 10 06

alexcjohnson:
@rreusser looks great! To play devil's advocate, can you include a test with some other `line.shape` values and `fill: 'tonexty'` between traces that both include isolated end and mid points? I don't think it will apply the line shape to the connections across null values - and that seems fine to me - but I do want to make sure that the lines and fills all still follow each other.

rreusser:
Good call, @alexcjohnson. There was a bug. It was considering single points to be a closed spline since the first point is (trivially but not meaningfully) the same as the first. I've modified the scatter trace not to try to draw single points as closed curves. I could make this change to the drawing routine itself, but I don't have strong feelings since I don't know the deep implications here.

Note that it connects spline gaps with straight lines so that for these cases with no more than two points per segment, the appearance is the same:

screen shot 2017-08-07 at 11 08 44

rreusser:
Oops. Also adding tonexty test.

alexcjohnson:
To put my mind at ease, can you also either make some of these segments 3+ points so spline is visibly different from straight, and/or change to one of the stepped shapes (ideally `'hv'` or `'vh'` so it's asymmetric)?

rreusser:
@alexcjohnson it now looks like this. Glad to make more, but should probably split it into new mocks if so.

screen shot 2017-08-07 at 11 29 03

alexcjohnson:
That mock is thing of beauty 😻 

💃

Issue #1886, opened 2017-07-17
[Sankey] Can't link 2 nodes with >3 links


jackparmer:
@alishobeiri has been making some really nice Sankey examples in Python:

https://plot.ly/~alishobeiri/1257/plotly-sankey-diagrams/

I think we found an issue where in plotly.js >=1.28.3, you can't connect 2 nodes with more than 3 links. Ali made a nice reproducible example here in Python:

https://plot.ly/~alishobeiri/1358/problem-first-identified-on-wapo-sankey/

To confirm this is a plotly.js issue, you can see that the WaPo example (many nodes with >3 links) works here using https://rawgit.com/monfera/plotly-webgl-parcoords/sankey-example-01/plotly-sankey.min.js:

https://codepen.io/plotly/pen/yXEJYZ

However if you change the plotly.js version to 1.28.3 it doesn't work:

https://codepen.io/plotly/pen/GEapyx

(Or using plotly-latest.min.js):

https://codepen.io/plotly/pen/zzQvjq

etpinard:
cc @monfera

monfera:
indeed, another reproduction is https://codepen.io/anon/pen/ayBoGG

monfera:
It should be fixed by the above commit.

@jackparmer thanks for the clear details! A minor note, the referred WaPo drone codepen mock precedes the decision to switch to a columnar representation so it won't work with the fix.

Indeed great examples by @alishobeiri, I especially love the Scottish referendum ones! The colors and opacity change on hover works out very well. An unrelated thing: now the tooltip value seems to be out of the hover tooltip box, I'll check what's going on.

![image](https://user-images.githubusercontent.com/1548516/28998639-a954a7bc-7a30-11e7-9710-afc47a091c7f.png)

monfera:
Interestingly the tooltip value is properly inside the box on my machine so I'm not yet sure how to reproduce it locally, will check.

![image](https://user-images.githubusercontent.com/1548516/28998650-05fdb896-7a31-11e7-9093-da3204340e4b.png)

PR #1934, opened 2017-08-05
Sankey fix for 4+ multilinks

monfera:
Purports to fix https://github.com/plotly/plotly.js/issues/1886


src/traces/sankey/render.js
```diff
@@ -155,7 +155,7 @@ function linkModel(uniqueKeys, d, l) {
     var tc = tinycolor(l.color);
     var basicKey = l.source.label + '|' + l.target.label;
     var foundKey = uniqueKeys[basicKey];
-    uniqueKeys[basicKey] = (foundKey === void(0) ? foundKey : 0) + 1;
+    uniqueKeys[basicKey] = (foundKey === void(0) ? 0 : foundKey) + 1;
```
etpinard: @monfera thanks for updating the baseline to :lock: this fix down :tada: 

For documentation purposes, can you explain what's going on this line?

monfera: The code relies on a unique key for the purpose of entering / exiting the proper links. A simple solution would be to lean on the index for this. But it's better to associate semantically meaningful elements, e.g. some specific node or link, with the DOM element - useful for future possible animations or tweening. As we don't require an unique identifier array from users, an approximation is

- for nodes: the node label
- for links: the label of the nodes they link

As there can be more than one link between two arbitrary nodes, the data binding key is made unique by a lookup that generates a new key for additional links. The ternary got flipped during an unrelated refactoring and it only came up with 4+ links.

alexcjohnson: might be clearer if we just use the fact that `foundKey` will always be truthy if it exists?
```
uniqueKeys[basicKey] = (foundKey || 0) + 1;
var key = basicKay + (foundKey ? '' : ('__' + foundKey));
```

Which brings up: it might be very uncommon, but what if the user actually had a `label` like `'name'` and another like `'name__1'`? I could see this popping out of some upstream disambiguation... Wouldn't it be more robust to always add the suffix like:
```
var keyCounter = uniqueKeys[basicKey] || 0;
uniqueKeys[basicKey] = keyCounter + 1;
var key = basicKey + '_' + keyCounter;
```

monfera: @alexcjohnson yes, the 1st suggestion works out if we 1-index things as you do, and indeed there is a small risk of name collision - commit on the way

alexcjohnson:
Nice, looks good to me! 💃

src/traces/sankey/render.js
etpinard: Thanks for :books: @monfera 

:dancer:

PR #1939, opened 2017-08-08
Improve `Plotly.toImage`

etpinard:
The bulk of this PR is in commit https://github.com/plotly/plotly.js/commit/8866525c6f5129526165d618e7b77d44b30e0da3 (where several improvement to `Plotly.toImage` are made) and https://github.com/plotly/plotly.js/commit/dd003ceec40dd89473aea3e0df0f4fa23738f457 (where the `setBackground` config option is now a little more portable)

With https://github.com/plotly/plotly.js/commit/8866525c6f5129526165d618e7b77d44b30e0da3, `Plotly.toImage` can now accept `data/layout/config` (plain) object as in input. Previously, only _existing_ graph divs were supported. This is important for image generation performance.

In details, to generate an image using only _official_ API methods (i.e. none of the `Plotly.Snapshot` methods), currently one must first call `Plotly.newPlot` and then pass the `gd` to `Plotly.toImage`. To ensure that `toImage` does not mutate the existing graph div, `toImage` clones it (i.e. extends the graph's data/layout) and calls `Plotly.plot` on an off-screen `<div>`. That is, `Plotly.plot` is called twice for each generated image. Needless to say, we can do better.

So, now calling `toImage` as:

```js
Plotly.toImage({
  data: [/* */],
  layout: {/* */},
  config: {/* */}
})
.then(imgData => {})
```
assumes that the data/layout/config containers are _new_ (i.e. aren't linked to an existing graph) meaning that the containers aren't extended and `Plotly.plot` is only called once per generated image :racehorse:


src/plot_api/plot_api.js
```diff
@@ -404,6 +404,11 @@ function opaqueSetBackground(gd, bgColor) {
     setBackground(gd, bgColor);
 }
 
+function blendSetBackground(gd, bgColor) {
+    var blend = Color.combine(bgColor, 'white');
```
etpinard: Implements [this](https://github.com/plotly/streambed/blob/7311d4386d80d45999797e87992f43fb6ecf48a1/image_server/server_app/main.js#L214-L218) (old) image server logic in plotly.js

src/plot_api/to_image.js
```diff
@@ -8,101 +8,191 @@
...
-        };
+    function convert() {
+        return new Promise(function(resolve, reject) {
+            var svg = toSVG(clonedGd);
 
-        if(!isSizeGood(opts.width) || !isSizeGood(opts.height)) {
-            reject(new Error('Height and width should be pixel values.'));
-        }
+            if(format === 'svg' && imageDataOnly) {
+                return resolve(svg);
```
etpinard: Fast pass for `svg` with `imageDataOnly` where `Snapshot.toSVG` gives the correct result.

Note that PDF and EPS image should first be generated as `svg` and them converted to pdf and eps using another library. In-house PDF and EPS image generation is outside the scope of plotly.js.

src/plot_api/plot_api.js
alexcjohnson: doesn't this end up visually equivalent to `opaqueSetBackground`? And if so can we remove `opaqueSetBackground`?

src/plot_api/to_image.js
```diff
@@ -8,101 +8,191 @@
...
+        return !(attr in opts) || Lib.validate(opts[attr], attrs[attr]);
+    }
+
+    if(!isBadlySet('width') || !isBadlySet('height')) {
+        throw new Error('Height and width should be pixel values.');
+    }
+
+    if(!isBadlySet('format')) {
+        throw new Error('Image format is not jpeg, png, svg or webp.');
+    }
```
alexcjohnson: Nice job using `coerce` for this! But `isBadlySet` is confusing me... it's `true` if you either haven't set the attribute or if it has a valid input? Wouldn't that be `isValidOrImplied` or something? Anyway I don't see how you can use the same logic for `format` (which has a default) as for `width` and `height` (which don't).

Dimensions you could just look at `fullOpts` after `coerce` to see if you have something that's not `undefined` (the implicit `dflt`)...

src/plot_api/to_image.js
```diff
@@ -8,101 +8,191 @@
...
+
+    // extend layout with image options
+    var layoutImage = Lib.extendFlat({}, layout);
+    if(width) layoutImage.width = width;
+    if(height) layoutImage.height = height;
+
+    // extend config for static plot
+    var configImage = Lib.extendFlat({}, config, {
+        staticPlot: true,
+        plotGlPixelRatio: 2,
```
alexcjohnson: is `plotGlPixelRatio` something people might want to override? Would it work if they could?

src/plot_api/to_image.js
alexcjohnson: is it worth handling `imageDataOnly=false` here too - and just adding the prefix? Or is it more complicated than that?

src/plot_api/to_image.js
alexcjohnson: Oh I see, it *is* OK to omit dimensions, then they default to what's in the layout... might be clearer if this was just used as the default `width = coerce('width', layout.width)`?

test/jasmine/tests/toimage_test.js
```diff
@@ -51,31 +54,51 @@ describe('Plotly.toImage', function() {
...
+        .then(function(gd) {
+            try {
+                Plotly.toImage(gd, {format: 'x'});
+            } catch(e) {
+                errors.push(e.message);
+            }
+        })
+        .then(function() {
+            expect(errors.length).toBe(1);
+            expect(errors[0]).toBe('Image format is not jpeg, png, svg or webp.');
```
alexcjohnson: You're not a fan of `expect(function() { ... }).toThrow(...)`?

src/plot_api/plot_api.js
etpinard: Maybe?

src/plot_api/to_image.js
etpinard: Yeah, it shouldn't be too hard. I'll add that in.

src/plot_api/to_image.js
etpinard: > might be clearer if this was just used as the default width = coerce('width', layout.width)?

right, but `layout.width` might not exist at this stage.

src/plot_api/to_image.js
etpinard: Yeah, potentially. Good :eyes:

src/plot_api/to_image.js
alexcjohnson: OK, but if it's not in `opts` and it's not in `layout` then what width do we get?

etpinard: We get the dflt from [`plots/layout_attributes.js`](https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/plots/layout_attributes.js#L63)

alexcjohnson: Alright, I don't want to beat a dead horse and it's not a big deal, but I believe it would still work just fine if you coerce with a `dflt` which might be `undefined` at this point, then you coerce with the real `dflt` later on, overriding that `undefined`.

Anyway my initial comment stands, about `isBadlySet` being misnamed and confusing logic.

src/plot_api/plot_api.js
etpinard: RE `'opaque'` vs `'blend'`, they don't generate the same PNG (`gm` compare generates a diff) but I can't find a case where the diff is detectable to my :eyes:

So, I guess a can make the new image server use `'opaque'` instead of `'blend'` and :hocho: `blendSetBackground`.

etpinard: done in https://github.com/plotly/plotly.js/pull/1939/commits/22a598bb814498a2057c2ea665195e77c0a2f598

src/plot_api/to_image.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1939/commits/4ec3e0f90ce2689bb617cb1b0dce86a3e593d035

test/jasmine/tests/toimage_test.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1939/commits/43a615022281b582b0b4b283ef778870f535c54f

src/plot_api/to_image.js
etpinard: done in https://github.com/plotly/plotly.js/pull/1939/commits/7565d0b8089c01562b5a953d6e2bfff6cea96048

src/plot_api/to_image.js
etpinard: isBadlySet -> isImpliedOrValidin  in https://github.com/plotly/plotly.js/pull/1939/commits/3c42efbe778adc1afd7113e49dda7395dceaa5d8

src/plot_api/to_image.js
```diff
@@ -8,101 +8,192 @@
...
+            var svg = toSVG(clonedGd);
 
-        if(!isSizeGood(opts.width) || !isSizeGood(opts.height)) {
-            reject(new Error('Height and width should be pixel values.'));
-        }
+            if(format === 'svg') {
+                if(imageDataOnly) {
+                    return resolve(svg);
+                } else {
+                    return resolve('data:image/svg+xml,' + encodeURIComponent(svg));
```
etpinard: as on https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/snapshot/svgtoimg.js#L64

alexcjohnson:
Love it. 💃

src/plot_api/to_image.js
etpinard: This broke a lot of tests. It made the `toImage` promise resolved before removing the clone graph div.

Fixed in https://github.com/plotly/plotly.js/pull/1939/commits/816df165605a401764140396aeec2e6b0a516bc3

Issue #1864, opened 2017-07-07
Stacked Histogram not rendering correctly when sample sizes are different 


cldougl:
It seems that stacked hists aren't stacking correctly when the traces have different sample sizes.

![screen shot 2017-07-07 at 3 09 29 pm](https://user-images.githubusercontent.com/2789078/27973000-514b6f18-6326-11e7-8bed-089548d0a4e3.png)

Example: https://codepen.io/plotly/pen/owPQqg
Working Example: https://codepen.io/plotly/pen/BzqwbE

(noticed this recently in the python doc example and I think it worked at some point despite different sample sizes).

alexcjohnson:
haha @cldougl when I open your [working example](https://codepen.io/plotly/pen/BzqwbE) I sometimes get it "working"
![screen shot 2017-08-07 at 11 33 16 am](https://user-images.githubusercontent.com/2678795/29033837-50237828-7b64-11e7-9a62-034759e68790.png)

And sometimes it automatically offsets one of the traces (presumably via the [edge-avoidance algo](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js#L667) though I'm a bit surprised that this would be so unstable):
![screen shot 2017-08-07 at 11 33 01 am](https://user-images.githubusercontent.com/2678795/29034312-f1788780-7b65-11e7-9ec3-177bcbf2314d.png)

And in fact sometimes your [non-working example](https://codepen.io/plotly/pen/owPQqg) does this and "sort of works":
![screen shot 2017-08-07 at 11 36 12 am](https://user-images.githubusercontent.com/2678795/29034091-21c8bf28-7b65-11e7-8dae-15fa9a2ee939.png)

I say "sort of works" because this is still not the outcome we want. Yes, the orange trace gets stacked on the blue, but the bin width used for the orange trace is not what's displayed as the bar widths.

I think in the end this may be the same as https://github.com/plotly/plotly.js/issues/50 - currently our oldest open bug! 🏆  So I'm pretty sure this is not a regression, just a hazard of using `random` in a bug report!

alexcjohnson:
closed by #1944

Issue #109, opened 2015-12-14
Tick label position incorrect for PDF and PNG with extra resolution


chrisjcameron:
In the sample plot linked below, the Y axis labels have three spaces as the ticksuffix to avoid hugging the axis too closely with the larger font. This looks fine when viewed in the Jupyter notebook or online at plot.ly. If I download the figure in PNG format, it looks fine. If I select PNG with extra resolution or PDF format, the suffix spacing is ignored. This seems like a bug in the compositing. 

Standard Res PNG
![demo](https://cloud.githubusercontent.com/assets/1383080/11785912/e2c3a9de-a251-11e5-8da8-21906902ad8a.png)

High Res PNG
![demo_high_res](https://cloud.githubusercontent.com/assets/1383080/11785910/e224939e-a251-11e5-83d3-6b5ef101ad22.png)

Original file located at https://plot.ly/~chris.j.cameron/95/proportion-of-nodes-adopted-vs-time-steps/

etpinard:
@chrisjcameron Bug confirmed. Thanks

etpinard:
This may be related to our use of [`class="crisp"`](https://github.com/plotly/plotly.js/blob/a27d62ca9b157db473e2a3bd8b51634fff2433b5/src/css/_base.scss#L23) in the axis drawing [code](https://github.com/plotly/plotly.js/blob/a27d62ca9b157db473e2a3bd8b51634fff2433b5/src/plots/cartesian/axes.js#L1427). 

I doubt that batik understands DOM classes.

etpinard:
Closing. This bug isn't related to plotly.js, but to our image-exporting services.

PR #1944, opened 2017-08-10
Match hist bins

### LINKED ISSUE #50: Dissimilar histogram bins when they should default to all the same
Histogram charts display some text on hover even when the bins are empty.
This only seems to happen in "Compare data on hover" mode.

JSBin: https://jsbin.com/pamafun/2/edit?js,output

example:
hover on empty bins 1
hover on empty bins 2

alexcjohnson:
Fixes #50 and #1864 
- All autobinned histograms on a given subplot, when `barmode` is not `'overlay'`, get the same bin size and alignment
- That size is the minimum any of them were auto-assigned
- If there are any manually binned histograms in the group, match the autobinned ones to the first manual one:
  - More than one manual histogram, if they're inconsistent there's no way to fully satisfy the constraints so don't even bother looking at it, just use the first one.
  - In case the manual bins are larger than the autobins but not as large as the next larger autobin step, expand (and shift) the autobins to match exactly
  - Otherwise shrink the autobins to the next largest integer fraction of the manual bin size. Align the bin edges so that there is an autobin center at the center of each manual bin. In this case the manual bins will still be represented incorrectly when they're drawn (with the narrower width of the autobins) but there's nothing we can do about that (stacked/grouped bars must have matching widths) at least the autobins will be drawn with the correct width.


src/traces/histogram/calc.js
```diff
@@ -180,10 +157,181 @@ module.exports = function calc(gd, trace) {
...
+        }
+
+        // Even non-autobinned traces get included here, so we get the greatest extent
+        // and minimum bin size of them all.
+        // But manually binned traces won't be adjusted, even if the auto values
+        // are inconsistent with the manual ones (or the manual ones are inconsistent
+        // with each other).
+        //
+        // TODO: there's probably a weird case here where a larger bin pushes the
+        // start/end out, then it gets shrunk and doesn't make sense with the smaller bin.
```
alexcjohnson: In a small set of cases, by fixing the major problem (bin sizes getting drawn wrong) we introduce a smaller problem (bin edges get shifted so an undesirably large fraction of data is ambiguously right at bin edges). This is possible because we set `minSize` and `minStart` independently so they can come from different traces. The vast majority of the time though the bin edges are still positioned nicely at the end of this, and the fix would need to involve keeping `minStart` based on the same trace that gave `minSize` while expanding it to encompass all of these traces, which wouldn't be *too* hard for numeric bins but gets tricky for `'M<#>'` bins. To me then this isn't worth fixing.

src/traces/histogram/calc.js
```diff
@@ -180,10 +157,181 @@ module.exports = function calc(gd, trace) {
...
+    }
+
+    return trace[binAttr];
+}
+
+/*
+ * return an array of traces that are all stacked or grouped together
+ * TODO: only considers histograms. Should we also harmonize with bars?
+ * in principle people can mix and match these, but bars always
+ * specify their positions explicitly...
```
alexcjohnson: We *could* try to treat bars the same as we treat manually-binned histograms in adjusting the autobins. That has more edge cases to consider though, nonuniform bar spacing being perhaps the hardest to manage, so I'm inclined not to worry about this one either. If someone is grouping/stacking bars with histograms I'd like to think it would be clear to them that they need to set the histogram binning to match the bar positions...

src/traces/histogram/calc.js
```diff
@@ -27,60 +28,35 @@ module.exports = function calc(gd, trace) {
...
-                binspec.start = pa.c2r(pa.r2c(binspec.start) - binspec.size);
-            }
-            else {
-                binspec.end = pa.c2r(pa.r2c(binspec.end) + binspec.size);
-            }
-        }
+    // the raw data was prepared in calcAllAutoBins (during the first trace in
+    // this group) and stashed. Pull it out and drop the stash
+    var pos0 = trace._pos0;
+    delete trace._pos0;
```
etpinard: Maybe `calcAllAutoBins` should return a tuple e.g. `[binspec, pos0]`?

src/traces/histogram/calc.js
etpinard: I don't mind not fixing this either. It would be nice to add a test case to document this behavior though.

src/traces/histogram/calc.js
etpinard: > I'm inclined not to worry about this one either

No worries for me either.

src/traces/histogram/calc.js
```diff
@@ -180,10 +157,181 @@ module.exports = function calc(gd, trace) {
...
+ * bin specs, there's nothing we can do, but if possible we will try to use the
+ * smallest bins of any of the auto values for all histograms grouped/stacked
+ * together.
+ */
+function calcAllAutoBins(gd, trace, pa, maindata) {
+    var binAttr = maindata + 'bins';
+
+    // all but the first trace in this group has already been marked finished
+    // clear this flag, so next time we run calc we will run autobin again
+    if(trace._autoBinFinished) {
```
etpinard: Maybe we should move auto-bins computations to `histograms/set_positions` (which loops over all traces of a given trace type) to avoid having to use `_` flags like this one?

test/jasmine/tests/histogram_test.js
```diff
@@ -162,10 +163,20 @@ describe('Test histogram', function() {
...
     describe('calc', function() {
-        function _calc(opts) {
-            var base = { type: 'histogram' },
-                trace = Lib.extendFlat({}, base, opts),
-                gd = { data: [trace] };
+        beforeAll(function() {
+            jasmine.addMatchers(customMatchers);
+        });
+
+        function _calc(opts, extraTraces) {
```
etpinard: This PR probably deserves an image test too.

src/traces/histogram/calc.js
etpinard: ... maybe `histogram/calc.js` should only sanitize `pos` and `size` values and call `arraysToCalcdata`?

alexcjohnson: > maybe `histogram/calc.js` should only sanitize `pos` and `size` values and call `arraysToCalcdata`, and leave the rest to `set_positions.js`?

Pro: that would avoid coupling between traces in a part of the code (`calc`) that's supposed to be operating on one trace at a time... and `set_positions` is only called after `calc` (as opposed to every replot), which is when it's needed.

Con: we'd need to be saving these sanitized `pos0` during `calc`, rather than making an actual `calcdata[i]` array - or perhaps stashing it in `calcdata[i][0]` as we used to do with trace-wide stuff (and still do here and there) but then `setPositions` would need to fill in the rest of `calcdata[i]` once it determined the bin spec.

So I feel like in the end it would be more confusing that way. Thoughts?

Incidentally, it looks like we're probably calling `setPositions` twice unnecessarily if we have both histogram and bar traces on the same plot. I'll investigate...

src/traces/histogram/calc.js
alexcjohnson: perhaps... we'd still need to stash `_pos0` because it could be used in a totally separate call to `calc`, but at least that could all be dealt with inside `calcAllAutoBins`, so yeah that would be easier to understand. 👍  (unless we end up moving to `setPositions`...)

src/traces/histogram/calc.js
etpinard: > So I feel like in the end it would be more confusing that way. Thoughts?

Good point. I don't have a strong opinion. This whole `calc` vs `set_positions` debate was just something that came to mind, not anything blocking that's for sure.

alexcjohnson: OK cool - I'm going to leave it then, but this discussion will be useful to keep in mind for the future, I suspect at some point we will want to reimagine the whole pipeline to be a bit more flexible - particularly in regards to minimizing redraw work.

test/jasmine/tests/histogram_test.js
alexcjohnson: modified test images in https://github.com/plotly/plotly.js/pull/1944/commits/3f173a447fcb09096026230501e0662916af8971 - previously trace1 would have gotten bin size 2, so bin centers at 1 & 2 for trace0, 0.5, 2.5, and 4.5 for trace1 and displayed bar widths of 0.5. Now they all get bin size = displayed bar width = 1

src/traces/histogram/calc.js
alexcjohnson: tested in https://github.com/plotly/plotly.js/pull/1944/commits/ec12eab5e17fcdf2985d1b888b44dda99614fbf3

src/traces/histogram/calc.js
alexcjohnson: reorg so `_pos0` is only in `calcAllAutoBins` in https://github.com/plotly/plotly.js/pull/1944/commits/89093ad688568d229baa967b440a7a8cd92970be

src/traces/histogram/calc.js
```diff
@@ -166,115 +163,116 @@ module.exports = function calc(gd, trace) {
...
-        // if the bins were not specified.
-        // in that case this will remain in the trace, so that future updates
-        // which would change the autobinning will not do so.
-        tracei._input[autoBinAttr] = tracei[autoBinAttr];
-    }
+    pos0 = trace._pos0;
+    delete trace._pos0;
 
-    return trace[binAttr];
+    return [trace[binAttr], pos0];
```
etpinard: Awesome. Thanks!

etpinard:
:dancer: 

though we should probably add a few examples on https://github.com/plotly/documentation to help  out users at some point.

---

FYI @alexcjohnson, the next oldest bug is https://github.com/plotly/plotly.js/issues/109 but maybe I just take that one.

alexcjohnson:
Haha you noticed my pattern 🥂 Some of them are getting a liiittle too old.

jurasource:
This seems to generate errors when I have multiple traces.  Only happens when I toggle one on/off a few times.
I managed to narrow down the bug to release 1.30.0.
No example I'm afraid, jsfiddle etc are blocked by my firewall.

`plotly-latest.js:168259 Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.calc (plotly-latest.js:168259)
    at Object.829.plots.doCalcdata (plotly-latest.js:152035)
    at 753.Plotly.plot (plotly-latest.js:130164)
    at Object.725.lib.syncOrAsync (plotly-latest.js:126168)
    at Object.restyle (plotly-latest.js:131297)
    at handleClick (plotly-latest.js:116491)
    at plotly-latest.js:116410`

etpinard:
@jurasource that error looks similar to https://github.com/plotly/plotly.js/issues/2020 - can you confirm?

jurasource:
@etpinard yes, it's the same error

Issue #1941, opened 2017-08-08
missing ticktext in `parcoords`


monfera:
This ticket is just a reference to the [original report](https://github.com/ropensci/plotly/issues/1096#issuecomment-321055902) as it needs to be investigated in `plotly.js`.

PR #1945, opened 2017-08-10
Proper `parcoords` axis tick scale when `ticktext` is unordered

monfera:
Fixes https://github.com/plotly/plotly.js/issues/1941 whose original report is https://github.com/ropensci/plotly/issues/1096#issuecomment-321055902.

The problem arose due to not properly supporting unordered tick labels (`ticktext`), fixed with this commit, line 72.

While at it, an additional safeguard was added in line 560, for the eventuality that a falsey `ticktext` eg. `0` is used such that the corresponding `tickvals` value is non-zero.

A test change was made to exercise both the out of order `ticktext` and the falsey value (`0`). It is expected that the `0` tick label does show up; and that it shows up as zero rather than its corresponding `tickvals` value which is 3. Note: the purpose of `tickvals` is the appropriate vertical placement of the tick to give the user flexibility, for example, uneven distance between adjacent ticks.


src/traces/parcoords/parcoords.js
```diff
@@ -557,7 +557,7 @@ module.exports = function(root, svg, styledData, layout, callbacks) {
                     .outerTickSize(2)
                     .ticks(wantedTickCount, d.tickFormat) // works for continuous scales only...
                     .tickValues(d.ordinal ? // and this works for ordinal scales
-                        sdom.map(function(d, i) {return texts && texts[i] || d;}) :
+                        sdom.map(function(d, i) {return texts && texts[i] !== undefined ? texts[i] : d;}) :
```
etpinard: Maybe filter out `null` too?

Actually, looks like cartesian `tickvals` / `ticktext` just stringifies falsy values: https://codepen.io/etpinard/pen/aywJNK?editors=0010 - which might be considered a bug, but I'd vote for making the parcoord implementation consistent(ly buggy).

monfera: What this does right now: if `ticktext[i]` is `undefined`, ie. the array is nonexistent or too short, then it'll use the `tickvals[i]` value (which can be the intention), represented as the `d`  in the green line. If it's however defined, then it'll be used (and stringified) whether falsy or not, in line with what you describe for Cartesian. Is it OK or would you rather that a missing `ticktext` array or missing `ticktext[i]` yields an `'undefined'` as the tick label?

etpinard: > Is it OK or would you rather that a missing ticktext array or missing ticktext[i] yields an 'undefined' as the tick label?

Ideally, it would be nice to make parcoords reuse some of the `cartesian/axes.js` routines.

Now, should we consider the current cartesian behavior a bug? I'd vote yes. I'd think only `typeof strings` ticktext values should be visible cc @alexcjohnson

monfera: What should be the next step? It'd be good to see the missing tick label fixed promptly and continuing the above convo here, or I could copy it in a separate issue `to be discussed`.

etpinard: > What should be the next step?

Enjoy your weekend. We'll revisit this Monday :tada:

alexcjohnson: I think you guys get this but just to be clear, there's a definite use case for `tickvals` without `ticktext` at all - when you want ticks at specific positions but labeled as regular numbers - see eg https://github.com/plotly/plotly.js/blob/master/test/image/mocks/axes_enumerated_ticks.json - this one has both completely missing `ticktext` (for the y axis) and shorter `ticktext` (for the x axis, ticks at 4 and 5 have no corresponding text). Some things to note about [the current cartesian implementation](https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/plots/cartesian/axes.js#L878-L879) though:
- It stringifies any `ticktext[i]` it finds that's not `undefined` - including `null`, so it matches what you have in this PR. Does this count as a bug? Perhaps we should filter out `null`, as we've generally taken that to be indistinguishable from `undefined` within the json (but not within `restyle/relayout` calls so it can be tricky to even insert a `null` value...) but I would vote to leave numbers and even booleans to be stringified - it's kind of weird, but I feel like the user's intent is clear in those cases.
- It doesn't *just* stringify the number if no `ticktext[i]` is found - it formats it similarly to regular numeric ticks, just with potentially [more digits](https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/plots/cartesian/axes.js#L780) than automatic ticks get, since these are not guaranteed to be round numbers. But it follows all the other formatting you might have provided: `exponentformat`, `tickprefix`, etc etc. I haven't looked at parcoords in enough detail to know if there is equivalent subtlety here, but if there is it should be included.

alexcjohnson: oh sorry @etpinard I didn't notice that in your codepen even `undefined` seems to be stringified - I don't understand how that happened, will have to investigate...

alexcjohnson: Ah, it still gets [here](https://github.com/plotly/plotly.js/blob/f42802689a99cba4057434664853e848c75625d6/src/plots/cartesian/axes.js#L1191) - which stringifies anything. This should definitely change... I think unfortunately both of these blocks are required in the cartesian case (~line 878 is for showing all ticks, ~line 1191 gets invoked more directly on hover) but yeah, would be great if they could share logic for excluding `undefined` and `null`.

monfera: With `parcoords` it's the situation you describe when the `tickvals` array is present and `ticktext` is missing: ticks will be placed possibly unevenly, while the ticks will be labeled with the numbers. The third axis is an example of that (0, 0.5, 1, 2, 3).

image

Currently, the dimension in this case is considered categorical and `tickvals` do not get formatted. But as you say it'd probably better be, which is a one-liner change (+ test). The result looks like this, wondering if it should be included in this PR or a separate PR, or await further discussion.

image

monfera: @alexcjohnson to summarise if I understand your latest comments and Étienne's comment on the top, this PR needs these: 

1. in `parcoords`, if `ticktext[i]` is either `undefined` or `null`, then we consider it unsupplied, ie. revert to just using `tickvals[i]` (the current PR currently checks for `undefined` only but I can extend)
2. `tickvals[i]` should be subject to `tickformat` as on my above screenshot with the percentages in column 3 and 4.
3. these things should be locked down in test so that, when we get around to axis unification, we can detect regressions
4. I'd create an issue that refers to comments in this PR that would be the basis for considering a change in cartesian such that it also does what's described in point 1 of this list

alexcjohnson: @monfera that all sounds exactly right 💯

etpinard: > in parcoords, if ticktext[i] is either undefined or null, then we consider it unsupplied, ie. revert to just using tickvals[i] (the current PR currently checks for undefined only but I can extend)

Maybe even: if ticktext[i] is either undefined or null **or false** [...]?

alexcjohnson: I'd let `false` pass through (ie get stringified) - not super useful, but it would be weird to stringify `true` but turn `false` into some number. `null` and `undefined` don't have opposites in the same way.

monfera: Thanks for the clarifications! The commit is done but I want to do some code simplification once `crossfilter` is tightened up. I can push where I am though.

monfera:
The last commit unifies the discussed logic, please take another look at it 👍

src/traces/parcoords/parcoords.js
```diff
@@ -569,18 +572,7 @@ module.exports = function(root, svg, styledData, layout, callbacks) {
...
-                                if(text === null || text === undefined) {
-                                    return dd;
-                                } else {
-                                    return text;
-                                }
-                            } else {
-                                return dd;
-                            }
-                        }) :
+                        sdom.map(toText(Lib.identity), texts) :
```
alexcjohnson: Am I missing someting or are the `()` not in the right place here? 
`sdom.map(toText(Lib.identity, texts))`

monfera: Thanks @alexcjohnson for the catch, thought it was magic that it still worked! Turns out, the `domainScale` change (same PR, higher up in the code) made `scale.domain()` to return the desired values anyway, which got mapped with the identity function (as `texts` was not supplied ie. falsey).

alexcjohnson: :) surprised me too, esp. since this seems to be fully tested. Thanks for figuring out what's really happening.

PR #1871, opened 2017-07-11
Implement `axis.layer` with 'above traces' and 'below traces' values

etpinard:
to be merged in https://github.com/plotly/plotly.js/pull/1861 with (hopefully) all the requirements for https://github.com/plotly/plotly.js/issues/1385

This PR adds an axis attribute `layer` with two possible values: `'above traces'` (the default and current behavior) and `'below traces`' which allows together with `cliponaxis: false` to draw markers and text nodes above axis lines and tick labels. See `cliponaxis_false` mock in e44aa4d for an example.

We chose to make `layer` have a _hard_ `'above traces'` default value corresponding to the behavior on `master`. This might be revisited in [`v2`](https://github.com/plotly/plotly.js/issues/420).

This PR also fixes a bug in the `cliponaxis: false` implementation in 7b62b73 and lints a few things.

-----

In a private convo with @alexcjohnson , we opted to **not** include special auto-range logic for `cliponaxis: false` traces. This issue will be dealt with by having a few more axis options in a later PR. I'll open up a ticket about this tomorrow. **DONE** in https://github.com/plotly/plotly.js/issues/1876

Unless someone :no_entry: me, <strike>I'll won't add a workaround patch for the cartesian line clip path problems on pan (see https://github.com/plotly/plotly.js/pull/1861#issuecomment-313755166) nor will I fix the IE / Edge issues from https://github.com/plotly/plotly.js/pull/1861#issuecomment-313755166. In trying to fix these, I broke some animation tests; I think it would be wise to wait until Ricky is back to working full-time before making another attempt. I'll document this things in new separate issues tomorrow.</strike> **FIXED** in e234827

PR #1952, opened 2017-08-15
Ternary axis layer

etpinard:
Implements https://github.com/plotly/plotly.js/pull/1871 for ternary subplots


etpinard:
I still have to add a few jasmine tests. I'm opening this PR early to tag it under the `v1.30.0` milestone.

src/plots/ternary/ternary.js
```diff
@@ -95,50 +96,75 @@ proto.makeFramework = function() {
...
 
-    var grids = _this.plotContainer.select('.grids').selectAll('g.grid')
-        .data(['agrid', 'bgrid', 'cgrid']);
-    grids.enter().append('g')
-        .attr('class', function(d) { return 'grid ' + d; })
-        .each(function(d) { _this.layers[d] = d3.select(this); });
-
-    _this.plotContainer.selectAll('.backplot,.grids')
-        .call(Drawing.setClipUrl, clipId);
+    toplevel.order();
```
etpinard: see https://bl.ocks.org/mbostock/1440766

src/plots/ternary/ternary.js
```diff
@@ -95,50 +96,75 @@ proto.makeFramework = function() {
...
-    var plotLayers = [
-        'draglayer',
-        'plotbg',
-        'backplot',
-        'grids',
-        'frontplot',
-        'aaxis', 'baxis', 'caxis', 'axlines'
-    ];
+
+    var plotLayers = ['draglayer', 'plotbg', 'backplot', 'grids'];
```
etpinard: The handful of lines below are a little bit verbose. If anyone knows a cleaner / better way to do this, let me know.

etpinard:
Looks like CircleCI is down for the moment. Nonetheless, `type: reviewable`.

src/plots/ternary/ternary.js
alexcjohnson: dunno if it's worthwhile - shorter but not necessarily cleaner - but you could do something like
```js
var plotOverLayers = ['frontplot'];
['a','b','c'].forEach(function(axLetter) {
  var axName = axLetter + 'axis';
  var isBelow = ternaryLayout[axName].layer === 'below';
  (isBelow ? plotLayers : plotOverLayers).push(axName, axLetter + 'line');
});
[].push.apply(plotLayers, plotOverLayers);
```

etpinard:
cc @geocosmite

alexcjohnson:
The `cliponaxis: true` trace with `axis.layer: 'below traces'` in the new test image looks funny, how the markers cover up half the axis line. But I think trying to solve that - moving the edge of the clip path back to the edge of the axis line, for example - would cause more problems than it solves. This combination doesn't have a use case that's obvious to me, but if you moved the clip path back it would make a subpixel gap so the more common use case `cliponaxis: true` and `axis.layer: 'above traces'` would look bad.

So lets keep it as is. No more comments, the one above is arguably not even an improvement 🙈
💃

src/plots/ternary/ternary.js
alexcjohnson: dunno if `selection.order` wasn't in d3 back when I started using it or I just couldn't find it... but we need to use this maaaany more places in our code!

PR #1975, opened 2017-08-30
fix and test shapes on overlaid axes

### LINKED ISSUE #1974: Shape type 'path' SVG causes TypeError: Cannot read property 'append' of undefined
Upgrading Plotly.js from v.1.24 to v.1.30 resulted in
	Failed: TypeError: Cannot read property 'append' of undefined
	at drawShape (plotly.js:120051)
	when SVG type 'path' Shape is used

To reproduce use the plotly-1.24.0.min.js library - the SVG shape shows.
Switch to plotly-1.30.0.min.js - TypeError: Cannot read property 'append' of undefined is thrown

https://codepen.io/dkolev8/pen/VzVywM

I found out that the last Plotly.js version that works with this SVG shape is v.1.24.1 
Looks like the breaking changes are introduced in [v.1.24.2](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#1242----2017-03-10)

alexcjohnson:
fixes #1974 

Interesting that we had a test of shapes with an overlaid axis but... we didn't actually put any shapes on this axis! now we do 😅 

cc @etpinard

### DIFF
--- a/src/components/shapes/draw.js
+++ b/src/components/shapes/draw.js
@@ -77,7 +77,7 @@ function drawOne(gd, index) {
     else {
         var plotinfo = gd._fullLayout._plots[options.xref + options.yref];
         if(plotinfo) {
-            var mainPlot = plotinfo.mainplot || plotinfo;
+            var mainPlot = plotinfo.mainplotinfo || plotinfo;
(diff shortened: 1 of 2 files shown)


etpinard:
:dancer:

PR #1979, opened 2017-09-01
Scale image in `Plotly.toImage`

etpinard:
This PR adds a `scale` option to `Plotly.toImage` (and `Plotly.downloadImage`) which _scales_ the generated image by a given factor. Note that, _scaling_ an image isn't the same as just relayouting it to some factor times the current layout `width`/`height` values. Scaling scales every feature of a graph - this should help users that want higher resolution image exports.

The new `scale` has an effect on all the supported formats. For `png`, `jpeg` and `webp` this is accomplished using  [`context.drawImage`](https://developer.mozilla.org/en/docs/Web/API/CanvasRenderingContext2D/drawImage)'s width/height options. For `svg`, we tweak the [`viewBox`](https://developer.mozilla.org/en/docs/Web/SVG/Attribute/viewBox) attribute similar to during scroll/pan interactions.

### DIFF
--- a/src/plot_api/to_image.js
+++ b/src/plot_api/to_image.js
@@ -40,6 +40,14 @@ var attrs = {
             'Defaults to the value found in `layout.height`'
         ].join(' ')
     },
+    scale: {
+        valType: 'number',
+        min: 0,
+        dflt: 1,
+        description: [
+            '...'
+        ].join(' ')
+    },
     setBackground: {
         valType: 'any',
         dflt: false,
@@ -111,6 +119,7 @@ function toImage(gd, opts) {
     var format = coerce('format');
     var width = coerce('width');
     var height = coerce('height');
+    var scale = coerce('scale');
     var setBackground = coerce('setBackground');
     var imageDataOnly = coerce('imageDataOnly');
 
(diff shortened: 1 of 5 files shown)


src/plot_api/to_image.js
```diff
@@ -128,7 +140,6 @@ function toImage(gd, opts) {
     // extend config for static plot
     var configImage = Lib.extendFlat({}, config, {
         staticPlot: true,
-        plotGlPixelRatio: config.plotGlPixelRatio || 2,
```
etpinard: Ooops. This should have been in a separate commit. In brief, this line was useless, so :hocho:

src/snapshot/svgtoimg.js
```diff
@@ -33,6 +33,12 @@ function svgToImg(opts) {
         }
 
         var canvas = opts.canvas;
+        var scale = opts.scale || 1;
+        var w0 = opts.width || 150;
+        var h0 = opts.height || 300;
```
alexcjohnson: 150 <-> 300 ?

alexcjohnson:
Brilliant. 💃  after https://github.com/plotly/plotly.js/pull/1979#discussion_r136489977

src/snapshot/svgtoimg.js
etpinard: Good eyes! Done in https://github.com/plotly/plotly.js/pull/1979/commits/633882b8596fc620c122ff14389003e80808ae8f

etpinard:
Thanks for the quick review @alexcjohnson !

I'll leave this one unmerged until `1.30.1` is released early next week.

Issue #1977, opened 2017-08-31
`showspikes: true` doesn't work if `showticklabels: false`


shogun70:
See https://codepen.io/anon/pen/mMvybw

My use case for this is when tick-labels for a "category" axis are too long or too numerous I want to hide the labels and show a bootstrap tooltip at the end of the spikeline. I use `hovermode: 'closest'` so I don't get the built-in `commonLabel` on hover.

I'm currently working around this issue using `tickfont.color: 'transparent'`instead of `showticklabels: false`.

etpinard:
Thanks for the report.

Related: https://github.com/plotly/plotly.js/pull/1604

shogun70:
Thanks, that was fast.

etpinard:
Thanks for 10/10 bug report @shogun70 :ok_hand:

PR #1980, opened 2017-09-01
Make hover spikes work when no tick labels are present

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1977 and makes https://github.com/plotly/plotly.js/pull/1604 obsolete.

@alexcjohnson I wouldn't mind a +1 on my approach before I add a test or two.

### DIFF
--- a/test/jasmine/tests/hover_spikeline_test.js
+++ b/test/jasmine/tests/hover_spikeline_test.js
@@ -4,6 +4,7 @@ var Plotly = require('@lib/index');
 var Fx = require('@src/components/fx');
 var Lib = require('@src/lib');
 
+var fail = require('../assets/fail_test');
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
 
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Looks solid to me. Might want to calculate the rest of the bbox attributes just so people can pretend it's a real bbox (I see one caller that looks for `.height` for example).

test/jasmine/tests/hover_spikeline_test.js
```diff
@@ -34,6 +38,20 @@ describe('spikeline', function() {
...
 
+        it('draws lines and markers on enabled axes w/o tick labels', function(done) {
+            Plotly.relayout(gd, {
+                'xaxis.showticklabels': false,
+                'yaxis.showticklabels': false
+            })
+            .then(function() {
+                Fx.hover('graph', {xval: 2, yval: 3}, 'xy');
+                expect(d3.selectAll('line.spikeline').size()).toEqual(4);
+                expect(d3.selectAll('circle.spikeline').size()).toEqual(1);
```
alexcjohnson: is it worth verifying that these things have reasonable dimensions/positions?

etpinard: Good call. done in 92b606b

alexcjohnson:
Looks fantastic - thanks for testing positions. 💃

Issue #1914, opened 2017-07-25
Setting y-axis to invisible does not clear its title


hylkevds:
Changing the visibility of an y-axis to false with Plotly.update does not clear the axis title from the graph. The axis itself (ticks, numbers, etc.) is removed.

The following jsfiddle (https://jsfiddle.net/tf6urjuq/) shows the issue. It's a modification of the "multiple y-axes example", with only 2 traces. The last two lines hide the second y-axis:
```
layout.yaxis2.visible = false;
Plotly.update('myDiv', data, layout);
```
But the text "yaxis2 title" is still visible.

Interestingly, when the axis is re-enabled with a different position, the title is moved to the new position.

rreusser:
Actually that's a good question: should a hidden axis display the title or not?

hylkevds:
Unsetting or changing the title doesn't work. In my jsFiddle, add `layout.yaxis2.title = "";` between the two plot command, and the title still shows.

rreusser:
Ah yes. Apologies. I meant it's unclear to me that visibility should affect the title at all. Right now it does and right now it's buggy. The part I was thinking through out loud is how best to fix it. It's definitely buggy though. 👍🏼

PR #1981, opened 2017-09-01
Clear axis title on full axis redraw

etpinard:
fixes https://github.com/plotly/plotly.js/issues/1914

This should do the trick :tophat:


alexcjohnson:
man, `axid === 'redraw'` makes me cringe but... yeah, all my doing. Looks great, and if we ever clean up that hack the test will make sure we get this right.

💃

etpinard:
> man, axid === 'redraw' makes me cringe

Yeah. `Axes.doTicks` is probably the hackiest function in all of plotly.js at the moment. :computer:

PR #1986, opened 2017-09-05
Introduce better regl management

dfcreative:
WIP.

Following regl-summer meetings.

## Questions

* Do we have to initialize regl-instances in `plot_api` or we can create only canvases?
  * :+1: easier to pass just canvases
  * :-1: no extensions management, we need them merged
* What is the use-case for `c.overdrag` transform in parcoords?


etpinard:
Thanks for making this PR @dfcreative :tada:

PR #1991, opened 2017-09-06
Test mathjax

alexcjohnson:
- Tests the current state of our MathJax rendering, including a "visual TODO list" of things that we would like to support but currently either do not or they are actively broken (`$` is replaced by `H` in the mock where it wouldn't work correctly)
- Fixes a bug with our `Plotly.plot` promise chain that meant multiple MathJax items in different parts of the plot would render but the plot would report being done before it really had finished. Prior to this change, the mock fails.

cc @etpinard - hopefully this will fix our MathJax image export issues.


src/plot_api/plot_api.js
```diff
@@ -107,7 +107,7 @@ Plotly.plot = function(gd, data, layout, config) {
     // any part of the plotting code can push to gd._promises, then
     // before we move to the next step, we check that they're all
     // complete, and empty out the promise list again.
-    gd._promises = [];
+    if(!Array.isArray(gd._promises)) gd._promises = [];
```
alexcjohnson: This I didn't write a test for... but I made this change so that if you make a plot, then change it before its promises have finished, you should still wait for them (via [`previousPromises`](https://github.com/plotly/plotly.js/pull/1991/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R376)) so you don't get a race condition between them and the new plot (with the old ones potentially overwriting elements of the new plot you made).

I haven't actually managed to cause any ill effects without this change in manual testing, but I think it's safer this way.

src/plot_api/plot_api.js
```diff
@@ -384,14 +384,16 @@ Plotly.plot = function(gd, data, layout, config) {
...
-        Plots.rehover
+        Plots.rehover,
+        Plots.previousPromises
     ];
 
-    Lib.syncOrAsync(seq, gd);
-
     // even if everything we did was synchronous, return a promise
     // so that the caller doesn't care which route we took
-    return Promise.all(gd._promises).then(function() {
```
alexcjohnson: This was the crux of the issue: if you have more than one async item in `seq`, only the promises added to `gd._promises` by the first one will exist when this `Promise.all` is called (at the synchronous return of `Lib.syncOrAsync`), so any added after that will not be awaited. But `Plots.previousPromises` (added to the end of `seq` above) looks at `gd._promises` when it's actually invoked so it gets them all.

etpinard: Solid investigation work :microscope: :tada: :1st_place_medal: 

Is this enough to close https://github.com/plotly/plotly.js/issues/76 you think?

etpinard:
I tested this locally using the old (i.e. current aka `imagetest`) image server and I didn't detect any flakiness. The new `image-exporter` is working beautifully too.

To make sure (sure), I'll manually rebuild this branch on CI a few times.

etpinard:
Moreover, I'm thinking that we could start testing `MathJax` using jasmine in a separate [`bundle_tests/`](https://github.com/plotly/plotly.js/tree/master/test/jasmine/bundle_tests) suite. The MathJax `<script>` could be added [here](https://github.com/plotly/plotly.js/blob/253eb247523e1f69041342ba7f0918ce8ac85a59/test/jasmine/karma.conf.js#L227).

src/plot_api/plot_api.js
alexcjohnson: > Is this enough to close #76 you think?

I think so!

alexcjohnson:
> To make sure (sure), I'll manually rebuild this branch on CI a few times.

One unrelated error, but the mathjax part succeeded 4/4 times on circle and all the times we've done it locally - looks pretty solid to me, ready to go?

etpinard:
> 4/4 times

I want 5/5; so one more run ->  https://circleci.com/gh/plotly/plotly.js/4955

etpinard:
It's green :white_check_mark: 

:dancer: :dancer: :dancer:

Issue #1325, opened 2017-01-21
Animating the layout of secondary axis fails


emanuelsetitinger:
Please find a reproducible example on CodePen: http://codepen.io/EmanuelS/pen/vgmKRm

This example shows a plot of one x-axis and two y-axes. While animating the range of the x-axis and the primary y-axis works perfectly fine, the range of the secondary y-axis doesn't change after Plotly.animate.

rreusser:
Thanks for the report and example. I'll take a look at this later today.

alexcjohnson:
Looks like this is https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/transition_axes.js#L18
-> matches `'x2axis'` instead of `'xaxis2'` - I'm abstracting all these regexes right now, should fix this but I'll include a test to confirm.

rreusser:
Ah! That definitely looks like an error on my part.

PR #1869, opened 2017-07-11
regl-scatter2d

dfcreative:
Work in progress, early stages. PR is for discussions.

* [x] make proper creation of glcontainer and glcanvas
* [x] make regl plot object
* [x] setup parcoord, pointcloud and scatter to use regl plot
* [x] make sure scatter updates are passing
* [x] integrate 
* [x] 2 scattergl plots case
* [x] shared context with parcoords case
* [x] hi-precision mode https://github.com/plotly/plotly.js/pull/1114
* [ ] ~~crossfilter case~~ ?
* [x] multiple plots
* [x] shared regl context
* [x] regl-scatter2d component
* [x] clustering points (kdtree) for boosting hover
* [x] scatter/scattergl image tests
* [ ] ~~selection API for regl-scatter2d component~~ handled by trace
* [x] set proper name, remove scattergl
* [x] sprite with all regular scatter symbols instead of font-atlas
* [x] all FIXME/TODOs in code
* [x] speed up autoscale
* [x] ~~background image https://github.com/plotly/plotly.js/issues/1603#issuecomment-315196515~~ a separate PR
* [x] wrong coloring https://github.com/plotly/plotly.js/issues/1909
* [ ] pointcloud/scatterregl two plots
* [x] extra update after updateRange call for half-transparent lines
* [x] point size discrepancy with scatter
* [x] test memory use
* [x] fills: tonexty, tonextx, tozerox, tozeroy
* [x] multiselect #1853
* [ ] speed up hover (3 strategies) − on hold
* [x] optimize creating select2d plot by cloning positions and unchangeble attribs
* [ ] reuse scatter2d tree for picking (can save up to 500ms for 1e6 points)
* [ ] subrange selection by the scatter2d tree (visible range + region)
* [x] ~~multiselect interaction: should persist selection on panning, zooming~~ leave it for a separate PR
* [x] all gl2d image tests
* [x] jasmine tests help required @etpinard 
* [x] benchmark plot relayout/restyle test-case required @etpinard 
* [x] hover/pick modes
* [x] entering selection mode in `gl2d_12` respans axes
* [ ] make better pick image
* [ ] running `Plotly.plot(container, mock.data)` two times breaks selection
* [ ] running `Plotly.plot(container, mock.data)` two times with modified `data.type` breaks selection and panning
* [x] axes range detection discrepancy image vs image in `gl2d_axes_labels`, `gl2d_axes_lines` or `gl2d_axes_range_type`
[...]

PR #1999, opened 2017-09-13
restyle/relayout refactor

alexcjohnson:
Something we've wanted to do for a looong time... get (nearly all) the attribute names and regex tests out of `_restyle` and `_relayout`, and use the schema instead:
- We already had `editType` as an optional field in the attribute schema, this makes it mandatory and uses it as the key way to determine what flavor of redraw to invoke.
- Adds a new optional field `impliedEdits` for correlated changes - for example `range` has `impliedEdits: {autorange: false}` because setting an explicit range is expected to disable the associated autorange. Note that the converse `autorange.impliedEdits: {range: undefined}` is used as it was before just to record the previous values for use in undo/redo, which we've discussed removing in v2.

Also includes a fix for #1325 as that came along essentially for free with generalizing our attribute string regexes.

I want to do a bit more testing of specific restyle/relayout calls to make sure we're doing the right redraws in all cases - especially of base/module interactions like `hoverlabel`, and module/module interactions like `heatmapgl`/`colorbar`. But I'd appreciate a high-level review ASAP @etpinard @rreusser


src/components/annotations/attributes.js
```diff
@@ -21,6 +20,7 @@ module.exports = {
         valType: 'boolean',
         role: 'info',
         dflt: true,
+        editType: 'docalcAutorange',
```
rreusser: Would `docalc+autorange` be a sensible name for this? It sounds like an enum camelcased and concatenated, but I could be mistaken.

rreusser: Ah, I see. Unless I'm mistaken, this adds no new labels. That's just plugging into what already exists. Still seems like maybe an enum in disguise, but nothing to change here then.

src/components/annotations/attributes.js
```diff
@@ -41,19 +42,23 @@ module.exports = {
...
         description: [
             'Sets the angle at which the `text` is drawn',
             'with respect to the horizontal.'
         ].join(' ')
     },
-    font: extendFlat({}, fontAttrs, {
+    font: fontAttrs({
+        editType: 'docalcAutorange',
+        colorEditType: 'doarraydraw',
```
rreusser: Could you point me toward what `doarraydraw` does? It's not entirely clear from the name, and if it appears somewhere else in the source, it's not jumping out at me.

src/components/colorscale/attributes.js
```diff
@@ -13,6 +13,8 @@ module.exports = {
         valType: 'boolean',
         role: 'info',
         dflt: true,
+        editType: 'docalc',
+        impliedEdits: {zmin: undefined, zmax: undefined},
```
rreusser: Same with impliedEdits. I assume these are attributes that are automatically applied within the same attr path, but I'm unable to find where it's implemented and it's not immediately clear what the `^` prefix means.

src/components/annotations/attributes.js
rreusser: One additional note while digging through here is that the `do-` prefix seems to be carried forward from what was originally just internal variable names. Would `editType: 'calcAutorange'` or `editType: 'calc+autorange'` make sense? It seems a bit friendlier if you're trying to make sense of the schema as a standalone concept.

src/components/fx/layout_attributes.js
```diff
@@ -29,38 +35,34 @@ module.exports = {
...
-        font: {
-            family: extendFlat({}, fontAttrs.family, {
-                dflt: constants.HOVERFONT
-            }),
-            size: extendFlat({}, fontAttrs.size, {
-                dflt: constants.HOVERFONTSIZE
-            }),
-            color: extendFlat({}, fontAttrs.color)
-        },
+        font: fontAttrs,
```
rreusser: ❤️

src/components/annotations/attributes.js
alexcjohnson: `docalcAutorange` is a single thing, means "do a recalc only if there's an autoranged axis related to this attribute" - Actually, ideally we'd specify what to do if there *isn't* an autoranged axis... so like `docalcAutorange+dostyle` or something.

> Would `editType: 'calcAutorange'` make sense?

Yeah totally, good idea! And to help us out (to figure out what `editType` a new attribute should get) I should probably describe somewhere what these flags actually mean in terms of the code paths they lead to.

src/lib/relative_attr.js
```diff
@@ -0,0 +1,47 @@
...
+var ASCEND = /^(.*)(\.[^\.\[\]]+|\[\d\])$/;
+var SIMPLEATTR = /^[^\.\[\]]+$/;
+/*
+ * calculate a relative attribute string, similar to a relative path
+ *
+ * @param {string} baseAttr:
+ *   an attribute string, such as 'annotations[3].x'. The "current location"
+ *   is the attribute string minus the last component ('annotations[3]')
+ * @param {string} relativeAttr:
+ *   a route to the desired attribute string, using '^' to ascend
```
rreusser: Ah, I think I understand. So relative to `a.b.c`, `^foo` would be `a.foo`. I don't know a solution, but one challenge with the schema is that it implements a lot of very precise semantics that can only be understood by locating the part of the code that handles them and working backwards (which isn't always possible since you might not even know that python is doing something with `role`, for example). I'm tempted to take a stab at `docs/schema.md`.

src/lib/relative_attr.js
```diff
@@ -0,0 +1,47 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+
+var ASCEND = /^(.*)(\.[^\.\[\]]+|\[\d\])$/;
```
rreusser: Could you add a very short comment about what this matches? In particular figuring out whether `^` is negation or an actual carat takes a good amount of staring and recalling js-specific regex escaping rules since this particular use could very reasonably involve either one.

src/components/colorscale/attributes.js
rreusser: Never mind. Github was suppressing the plot_api diff. Found it.

src/plot_api/plot_api.js
```diff
@@ -1528,57 +1430,14 @@ function _restyle(gd, aobj, _traces) {
...
-                doextra('xtype', 'scaled', i);
-            }
-            else if(['y0', 'dy'].indexOf(ai) !== -1 &&
-                    contFull.y && contFull.ytype !== 'scaled') {
-                doextra('ytype', 'scaled', i);
+            valObject = PlotSchema.getTraceValObject(contFull, param.parts);
+
+            if(valObject && valObject.impliedEdits && newVal !== null) {
+                for(var impliedKey in valObject.impliedEdits) {
+                    doextra(Lib.relativeAttr(ai, impliedKey), valObject.impliedEdits[impliedKey], i);
```
rreusser: One question about implied edits is whether it moves away from declarative plots. That is, if you set `foo` (which, e.g., implicitly also sets `bar: false`), is it valid to then set `bar: true`? Is there pathway that might accidentally circumvent the implied edits and lead to a bit of inconsistency? Would it be better/worse to think of  implied edits as a concern of the plotly API consumer rather than the plotly API itself? Like the workspace would recognize the attribute you're modifying has an implied edit and send *both* of those to the API? It seems nice from an end-user perspective; my only concern is that it might be surprising/confusing to modify an attribute and find that additional changes have been made to your plot state.

src/plot_api/plot_api.js
```diff
@@ -1665,41 +1520,20 @@ function _restyle(gd, aobj, _traces) {
                 flags.docalc = true;
             }
             else {
-                var aiHead = param.parts[0];
-                var moduleAttrs = (contFull._module || {}).attributes;
-                var valObject = moduleAttrs && moduleAttrs[aiHead];
-                if(!valObject) valObject = Plots.attributes[aiHead];
                 if(valObject) {
-                    /*
-                     * In occasional cases we can't the innermost valObject
```
rreusser: "can't ____ the innermost…"

rreusser: Oh never mind. This is in the removed section. 👍

src/plot_api/plot_schema.js
```diff
@@ -208,6 +216,182 @@ exports.findArrayAttributes = function(trace) {
     return arrayAttributes;
 };
 
+/*
+ * Find the valObject for one attribute in an existing trace
```
rreusser: Is a `valObject` just a value? Or is it a nested property object? Or something else?

rreusser: Less vaguely, `valObjects` are defined [here](https://github.com/plotly/plotly.js/blob/7ea0d250ab6731be0b98d40ee0fd9e6bf667986c/src/lib/coerce.js#L21), but I'm trying to process how the return value relates to those definitions.

src/plot_api/plot_schema.js
```diff
@@ -287,26 +461,37 @@ function getLayoutAttributes() {
...
-    Object.keys(Registry.componentsRegistry).forEach(function(k) {
-        var _module = Registry.componentsRegistry[k];
-
-        if(!_module.layoutAttributes) return;
-
-        if(_module.schema && _module.schema.layout) {
-            Object.keys(_module.schema.layout).forEach(function(v) {
-                insertAttrs(layoutAttributes, _module.schema.layout[v], v);
-            });
+    for(key in Registry.componentsRegistry) {
```
rreusser: 🐎

src/plots/cartesian/constants.js
```diff
@@ -7,25 +7,25 @@
...
 'use strict';
+var counterRegex = require('../../lib').counterRegex;
 
 
 module.exports = {
 
     idRegex: {
-        x: /^x([2-9]|[1-9][0-9]+)?$/,
-        y: /^y([2-9]|[1-9][0-9]+)?$/
+        x: counterRegex('x'),
```
rreusser: 👏

src/components/annotations/attributes.js
rreusser: Never mind. Found it in the suppressed diff.

src/components/annotations/attributes.js
rreusser: Maybe `calcIfAutorange`?

> to figure out what editType a new attribute should get

FWIW, I have a history of making bad choices here for lack of really understanding the implications of some schema concepts. If you want I can try making `docs/schema.md` that explains, at the very least, what some of the schema items link up with. Some of the concepts like `_isLinkedToArray` get a little subtle. Even for the fifth time interacting with it, I think a brief description with links to a couple relevant spots in the code would help.

alexcjohnson: > Maybe `calcIfAutorange`?

🏆 

> If you want I can try making `docs/schema.md`

Definitely time for something like that! @etpinard thoughts about how/where to do this? Should it be a standalone thing, or somehow included in `Plotly.PlotSchema.get()` or...?

rreusser: My vote would be for not including in the code so that it's as accessible and readable as possible without having to realize it's an API function and instantiate anything in order to access.

src/lib/relative_attr.js
rreusser: (re-reading, matches `foo.string` or `foo[5]` or `foo.bar[5]`, I think, while capturing the final part as well as everything before it. It's not so bad, I guess.

src/plot_api/plot_schema.js
```diff
@@ -287,26 +277,37 @@ function getLayoutAttributes() {
...
+         * not yaxis (rangeselector, rangeslider) - delete from y schema.
+         * Note that the input attributes for xaxis/yaxis are the same object
+         * so it's not possible to only add them to xaxis from the start.
+         * If we ever have such asymmetry the other way, or anywhere else,
+         * we will need to extend both this code and mergeComponentAttrsToSubplot
+         * (which will not find yaxis only for example)
+         */
+        if(schema && (schema.subplots || schema.layout)) {
+            var subplots = schema.subplots;
+            if(subplots && subplots.xaxis && !subplots.yaxis) {
```
etpinard: What if the `xaxis` and `yaxis` attributes were different objects?

etpinard: ... would this simplify things?

etpinard: I'm thinking we _might_ have to make the `xaxis` and `yaxis` attribute objects different down the road if we (ever) want to have more details per-axis description (similar to https://github.com/plotly/plotly.js/issues/621)

src/plot_api/plot_api.js
alexcjohnson: This PR by itself isn't moving toward or away from anything as far as users are concerned... all of these `impliedEdits` were previously handled by dedicated blocks in `_restyle`/`_relayout`. Without them, a call like `relayout(gd, 'xaxis.range', [5, 10])` wouldn't do anything if `xaxis.autorange` was true.

Looking ahead though to `Plotly.react`, I'm hoping that by having `impliedEdits` in the schema we can help people make these implied changes in a transparent way - so if you have a `range` control that modifies the state and passes it back to the plot in `Plotly.react`, it would be able to look up the implied changes we think it should make, but the control would have final say over whether to apply those changes or not.

src/registry.js
```diff
@@ -116,6 +150,41 @@ function findArrayRegexps(_module) {
...
+    var componentSchema = exports.componentsRegistry[componentName].schema;
+    if(!componentSchema || !componentSchema.transforms) return;
+
+    var transformAttrs = componentSchema.transforms[transformType];
+    if(transformAttrs) {
+        Lib.extendDeep(exports.transformsRegistry[transformType].attributes, transformAttrs);
+    }
+}
+
+function mergeComponentAttrsToSubplot(componentName, subplotName) {
```
etpinard: Very clean. Good call making a distinction between `subplots` and _regular_ layout attributes.

:+1:

src/plot_api/plot_api.js
rreusser: Got it. Thanks for the clarification.

src/plot_api/plot_schema.js
alexcjohnson: hah yeah, `coerce.valObjects` *describes* `valObject`s but it isn't itself at set of `valObject`s. What we mean by it in this context is a schema entry describing one attribute. I'll see what I can do in terms of documenting this better...

rreusser: Not a big deal. Just another part of the code I haven't gone too deep in. To clear up a common misconception though, naming in CS is a very easy problem (randomly generated strings will do). Naming _for maximum interpretability_ is the hard part. 😄

src/plot_api/plot_schema.js
alexcjohnson: > What if the xaxis and yaxis attributes were different objects?

Then we'd need to extend this section 😈  Currently we don't have that so I didn't implement it, just added the comment above:
```
 * If we ever have such asymmetry the other way, or anywhere else,
 * we will need to extend both this code and mergeComponentAttrsToSubplot
 * (which will not find yaxis only for example)
```

alexcjohnson: I suppose it would be useful to add an error check in there though, so it alerts us to fix it if we ever do add a y-axis-only, or different x and y, component attributes. I'll give that a shot.

src/lib/regex.js
```diff
@@ -0,0 +1,23 @@
...
+// Simple helper functions
+// none of these need any external deps
+
+/*
+ * make a regex for matching counter ids/names ie xaxis, xaxis2, xaxis10...
+ *  eg: regexCounter('x')
+ * tail is an optional piece after the id
+ *  eg regexCounter('scene', '.annotations') for scene2.annotations etc.
+ */
+exports.counter = function(head, tail, openEnded) {
```
etpinard: :sparkles:  Brilliant :sparkles:

src/lib/regex.js
```diff
@@ -0,0 +1,23 @@
...
+    return new RegExp('^' + head + '([2-9]|[1-9][0-9]+)?' +
```
rreusser: Ah, so `xaxis0` and `xaxis1` are strictly invalid? 0 makes sense. 1 seems slightly asymmetric, but nothing new.

src/lib/regex.js
etpinard: One of the things I've been wanting to do for like 2 years now. Thanks a lot for doing this @alexcjohnson :tada:

src/plot_api/plot_schema.js
```diff
@@ -208,6 +208,142 @@ exports.findArrayAttributes = function(trace) {
     return arrayAttributes;
 };
 
+/*
```
etpinard: :sos: semi-useless lint comment alert :sos: 

jsDOC is enabled on comments starting with `/**` (see http://usejsdoc.org/about-getting-started.html). As we don't extract those jsDOC comment (yet), adding this extra `*` would _just_ make that comment look prettier in my editor. Non-blocking.

src/plot_api/plot_schema.js
```diff
@@ -208,6 +208,142 @@ exports.findArrayAttributes = function(trace) {
...
+ */
+exports.getLayoutValObject = function(parts) {
+    var valObject = layoutHeadAttr(parts[0]);
+
+    return recurseIntoValObject(valObject, parts, 1);
+};
+
+function layoutHeadAttr(head) {
+    if(head in baseLayoutAttributes) return baseLayoutAttributes[head];
+    if(head in Registry.traceLayoutAttributes) return Registry.traceLayoutAttributes[head];
```
etpinard: I vaguely remember reading that `k in obj` is much slower than `obj[k] !== undefined`, but maybe newer browsers have optimized this. `k in obj` is safer though as some of our objects have keys set to `undefined`.

src/plot_api/plot_schema.js
```diff
@@ -208,6 +208,142 @@ exports.findArrayAttributes = function(trace) {
...
+    var key, _module;
+
+    for(key in Registry.componentsRegistry) {
+        _module = Registry.componentsRegistry[key];
+        if(head === _module.name) return _module.layoutAttributes;
+    }
+
+    for(key in Registry.subplotsRegistry) {
+        _module = Registry.subplotsRegistry[key];
+        if(_module.attrRegex && _module.attrRegex.test(head)) return _module.layoutAttributes;
```
etpinard: Beautiful.

src/plot_api/plot_schema.js
rreusser: Is `baseLayoutAttributes.hasOwnProperty('head')` also (effectively) equivalent?

src/plot_api/plot_schema.js
```diff
@@ -208,6 +208,142 @@ exports.findArrayAttributes = function(trace) {
...
+            if(!isIndex(index) || index >= valObject.items.length) return false;
+            valObject = valObject.items[index];
+        }
+    }
+
+    return valObject;
+}
+
+function isIndex(val) {
+    return val === Math.round(val) && val >= 0;
```
etpinard: [`Number.isInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) is a thing, but IE doesn't support it, so :ok_hand:

rreusser: Oh, javascript.

```
> Number.isInteger(new Number(5))
< false
```

src/plot_api/plot_schema.js
etpinard: I feel like `coerce.valObjects` should be named `coerce.valObjectMeta` instead. To me and to [`PlotSchema.isValObject`](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plot_api/plot_schema.js#L130) a val object is any object with a `valType` key - which is essentially what `getTraceValObject` and `getLayoutValObject` return.

src/components/annotations/attributes.js
```diff
@@ -164,13 +164,16 @@ module.exports = {
         min: 0.3,
         dflt: 1,
         role: 'style',
-        description: 'Sets the size (in px) of annotation arrow head.'
+        description: [
+            'Sets the size of the annotation arrow head, relative to `arrowwidth`.',
+            'A value of 1 (default) gives a head about 3x as wide as the line.'
```
etpinard: Thanks :books:

src/components/fx/hover.js
```diff
@@ -580,24 +580,27 @@ function createHoverText(hoverData, opts, gd) {
...
 
         lpath.enter().append('path')
-            .style({
-                fill: commonLabelOpts.bgcolor || Color.defaultLine,
-                stroke: commonLabelOpts.bordercolor || Color.background,
-                'stroke-width': '1px'
-            });
+            .style({'stroke-width': '1px'});
+
+        lpath.style({
```
etpinard: nicely done.

alexcjohnson: it was so hard to figure out what flags would happen in a particular restyle call before... I *thought* I got the same ones for `layout.hoverlabel` but [this test](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/test/jasmine/tests/hover_label_test.js#L1201) was failing until I made this change so it must have been hitting some other conditional I didn't notice.

src/components/annotations/attributes.js
etpinard: I'd vote for adding a attribute-like edit type declaration object in [`edit_types.js`](https://github.com/plotly/plotly.js/blob/master/src/plot_api/edit_types.js) with a `description` field.  For example,

```js
var traceEditTypes: {
  docalc: {
    description: [
       // ....
    ].join(' ')
  },
  // ....
}
```

That `description` field would get stripped off by the [`compress_attributes`](https://github.com/plotly/plotly.js/blob/master/tasks/util/compress_attributes.js) transform. `Plotly.PlotSchema.get().defs.editTypes` would then return the list of edit types with their description.


I see your point @rreusser about relying an API function to show docs being less-than-ideal. But since we already have a hard time keeping https://github.com/plotly/documentation/ up to speed, adding another layer of static (e.g. a github wiki or a `docs/` directory) documentation sounds dangerous. No docs is better than bad docs, right? That said, it wouldn't hurt adding a few lines in our CONTRIBUTING.md over the edit types.

src/components/colorscale/color_attributes.js
```diff
@@ -9,80 +9,104 @@
...
+ * Make all the attributes for a regular colorscale:
+ *  color, colorscale, cauto, cmin, cmax, autocolorscale, reversescale
+ *
+ * @param {string} context:
+ *   the container this is in (*marker*, *marker.line* etc)
+ * @param {optional string} editTypeOverride:
+ *   most of these attributes already require a recalc, but the ones that do not
+ *   have editType *dostyle* or *doplot* unless you override (presumably with *docalc*)
+ * @param {optional bool} autoColorDflt:
+ *   normally autocolorscale.dflt is `true`, but pass `false` to override
```
etpinard: brilliant :sparkles:

src/components/errorbars/attributes.js
```diff
@@ -89,48 +96,57 @@ module.exports = {
...
     },
     tracerefminus: {
         valType: 'integer',
         min: 0,
         dflt: 0,
-        role: 'info'
+        role: 'info',
+        editType: 'dostyle'
     },
     copy_ystyle: {
```
etpinard: Should we :hocho: these in v2?

etpinard: (I'd vote _yes_)

src/components/fx/layout_attributes.js
```diff
@@ -29,38 +35,34 @@ module.exports = {
...
         values: ['x', 'y', 'closest', false],
+        editType: 'domodebar',
         description: 'Determines the mode of hover interactions.'
     },
 
     hoverlabel: {
         bgcolor: {
             valType: 'color',
             role: 'style',
+            editType: 'none',
```
etpinard: Interesting. This means a round of `supplyDefaults` is enough?

src/plot_api/edit_types.js
```diff
@@ -54,5 +58,53 @@ module.exports = {
...
+ * @param {object} attrs: the attributes to override. Will not be mutated.
+ * @param {string} editTypeOverride: the new editType to use
+ * @param {bool|'nested'} overrideContainers: should we override editType for containers or just
+ *   `valObject`s? 'nested' will override editType for all but the top level. Containers below
+ *   the absolute top level (trace or layout root) DO need an editType, to handle the case
+ *   where you edit the whole container, but in some cases you want to provide it explicitly
+ *   as it may be more expansive than the attribute editTypes, since you don't know if there
+ *   is a data array in the container (which if provided directly triggers an automatic docalc)
+ */
+function overrideAll(attrs, editTypeOverride, overrideContainers) {
```
etpinard: I'd vote for a making the call signature `overrideAll(attrs, editTypeOverride, opts)` where `opts` can have key `overrideContainers`. It would make lines like this [one](https://github.com/plotly/plotly.js/pull/1999/commits/f49ae5ea580d5826666574c041a744e89180d873#diff-8e9a79b541026f6567178f4c9d990c99R288) more readable.

src/plot_api/plot_schema.js
```diff
@@ -259,16 +259,16 @@ exports.getTraceValObject = function(trace, parts) {
...
+        // then look in the subplot attributes
         if(!valObject) {
             var subplotModule = _module.basePlotModule;
             if(subplotModule && subplotModule.attributes) {
                 valObject = subplotModule.attributes[head];
             }
         }
+
+        // finally look in the global attributes
+        if(!valObject) valObject = baseAttributes[head];
```
etpinard: Why do we have look for base attributes last now?

src/plot_api/plot_schema.js
```diff
@@ -504,7 +544,8 @@ function mergeValTypeAndRole(attrs) {
             description: [
                 'Sets the source reference on plot.ly for ',
                 attrName, '.'
-            ].join(' ')
+            ].join(' '),
+            editType: 'none'
```
etpinard: cc https://github.com/plotly/documentation/issues/810

src/plots/cartesian/layout_attributes.js
```diff
@@ -92,9 +100,11 @@ module.exports = {
         valType: 'info_array',
         role: 'info',
         items: [
-            {valType: 'any'},
-            {valType: 'any'}
+            {valType: 'any', editType: 'doplot', impliedEdits: {'^autorange': false}},
```
etpinard: very clean syntax :+1:

src/plots/font_attributes.js
```diff
@@ -8,33 +8,67 @@
...
+ *   @param {optional bool} arrayOk:
+ *     should each part (family, size, color) be arrayOk? default false.
+ *   @param {string} editType:
+ *     the editType for all pieces of this font
+ *   @param {optional string} colorEditType:
+ *     a separate editType just for color
+ *
+ * @return {object} attributes object containing {family, size, color} as specified
+ */
+module.exports = function(opts) {
```
etpinard: :1st_place_medal:

src/plots/geo/layout/layout_attributes.js
```diff
@@ -254,4 +255,4 @@ module.exports = {
     },
     lonaxis: geoAxesAttrs,
     lataxis: geoAxesAttrs
-};
+}, 'doplot', true);
```
etpinard: Nice. I might try to add fine-grained edits in my geo refactor.

src/plot_api/plot_schema.js
alexcjohnson: plotly.js would be [very unhappy](https://github.com/plotly/fast-isnumeric/blob/master/index.js#L3-L6) if you use `new Number` in your plot.

src/components/errorbars/attributes.js
alexcjohnson: `traceref` or `copy_ystyle`? The former absolutely 🔪  but I kinda like the latter.

src/plots/gl2d/index.js
```diff
@@ -27,6 +30,21 @@ exports.attrRegex = constants.attrRegex;
...
+// this could potentially be just `layoutAttributes` but it would
+// still need special handling somewhere to give it precedence over
+// the svg version when both are in use on one plot
+exports.layoutAttrOverrides = overrideAll(Cartesian.layoutAttributes, 'doplot', true);
+
+// similar overrides for base plot attributes (and those added by components)
+exports.baseLayoutAttrOverrides = overrideAll({
+    plot_bgcolor: Plots.layoutAttributes.plot_bgcolor,
+    hoverlabel: fxAttrs.hoverlabel
+    // dragmode needs docalc but only when transitioning TO lasso or select
```
etpinard: FYI: this won't be needed after https://github.com/plotly/plotly.js/pull/1869

src/components/fx/layout_attributes.js
alexcjohnson: `supplyDefaults`, plus `Plots.rehover` that also always gets called on edits.

src/plots/gl3d/index.js
```diff
@@ -30,6 +33,13 @@ exports.attributes = require('./layout/attributes');
 
 exports.layoutAttributes = require('./layout/layout_attributes');
 
+exports.baseLayoutAttrOverrides = overrideAll({
+    hoverlabel: fxAttrs.hoverlabel
+    // dragmode needs docalc but only when transitioning TO lasso or select
```
etpinard: Hmm. There's no lasso or select modes in gl3d.

src/components/errorbars/attributes.js
etpinard: I was referring to `copy_ystyle` and `copy_zstyle`, but I don't mind keeping them in v2 :smile:

src/plot_api/plot_schema.js
alexcjohnson: > Why do we have look for base attributes last now?

In case a specific trace alters the base behavior - which is assumed to require a longer redraw path, like [`box.name`](https://github.com/plotly/plotly.js/pull/1999/files#diff-bbaff551e896584401309e06c8e48a83R54) needing calc because it might be used as the coordinate.

src/plots/gl3d/index.js
alexcjohnson: haha good catch. 🔪

src/plots/gl3d/layout/defaults.js
```diff
@@ -59,7 +59,7 @@ function handleGl3dDefaults(sceneLayoutIn, sceneLayoutOut, coerce, opts) {
     var bgcolor = coerce('bgcolor'),
         bgColorCombined = Color.combine(bgcolor, opts.paper_bgcolor);
 
-    var cameraKeys = Object.keys(layoutAttributes.camera);
+    var cameraKeys = ['up', 'center', 'eye'];
```
etpinard: Is that to improve perf :racehorse: ?

If so, why not, I think those `camera` attributes are pretty much :lock: ed forever.

src/components/errorbars/attributes.js
alexcjohnson: yeah, they're kind of ugly, but it's such a common use case to want the same style for both/all directions (and in fact I'd say such an *uncommon* use case to have *different* styles) that I think they deserve to stay.

src/plots/gl3d/layout/defaults.js
alexcjohnson: not 🐎 , just to avoid having to strip out `editType`

src/plots/pad_attributes.js
```diff
@@ -8,29 +8,35 @@
 
 'use strict';
 
+// This is used exclusively by components inside component arrays,
```
etpinard: Thanks for the note. We might have to make `pad_attributes.js` export a function similar to `font_attributes.js` is ever we want to add `pad` to e.g. `legend` and `colorbar`.

src/traces/scattercarpet/attributes.js
```diff
@@ -49,19 +52,6 @@ module.exports = {
             '`ternary<i>.sum`.'
         ].join(' ')
     },
-    sum: {
```
etpinard: good catch!

src/plot_api/plot_api.js
```diff
@@ -2231,73 +1991,15 @@ function _relayout(gd, aobj) {
...
-                    ai.match(/^(bar|box|font)/)) {
-                flags.docalc = true;
-            }
-            else if(fullLayout._has('gl2d') &&
-                (ai.indexOf('axis') !== -1 || ai === 'plot_bgcolor')
-            ) {
-                flags.doplot = true;
-            }
-            else if(fullLayout._has('gl2d') &&
+            if(fullLayout._has('gl2d') &&
```
etpinard: to :hocho: in https://github.com/plotly/plotly.js/pull/1869

src/lib/regex.js
alexcjohnson: We sanitize a few of these (turning `scene1` into `scene` etc) on the way in using [`cleanLayout`](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plot_api/helpers.js#L56) but yeah, it's always been that way. The idea is that most of the time you don't have more than one, so needing to add the `1` would be annoying and ugly.

src/plot_api/plot_api.js
```diff
@@ -1607,10 +1466,6 @@ function _restyle(gd, aobj, _traces) {
             else if(ai === 'colorbar.tick0' || ai === 'colorbar.dtick') {
                 doextra('colorbar.tickmode', 'linear', i);
```
etpinard: Why does this remain? Should this get done automatically via `impliedEdits`?

test/jasmine/tests/plotschema_test.js
```diff
@@ -147,6 +147,18 @@ describe('plot schema', function() {
         });
     });
 
+    it('includes xaxis-only items on only the x axis, not y or bare', function() {
+        var items = ['rangeselector', 'rangeslider'];
+
+        var layoutAttrs = plotSchema.layout.layoutAttributes;
+
+        items.forEach(function(item) {
```
etpinard: I'm curious. How did you find out about this? Did you compare `Plotly.PlotSchema.get()` to some base schema output?

etpinard: .... I'm thinking maybe we should start committing a `plot-schema.json` in this repo (possible in `dist/` alongside the bundles). i'm sure some users would find it useful. Thoughts?

src/plots/gl2d/index.js
```diff
@@ -27,6 +30,21 @@ exports.attrRegex = constants.attrRegex;
 
 exports.attributes = require('../cartesian/attributes');
 
+// gl2d uses svg axis attributes verbatim, but overrides editType
+// this could potentially be just `layoutAttributes` but it would
+// still need special handling somewhere to give it precedence over
+// the svg version when both are in use on one plot
```
etpinard: > but it would still need special handling somewhere to give it precedence over the svg version when both are in use on one plot

Can you elaborate on this? Depending on how complicated that needed _special handling_, I might vote for trying to :hocho: those *AttrOverride` fields - which feels like a layer of abstraction we could avoid.

src/plots/gl2d/index.js
```diff
@@ -27,6 +30,21 @@ exports.attrRegex = constants.attrRegex;
...
+exports.layoutAttrOverrides = overrideAll(Cartesian.layoutAttributes, 'doplot', true);
+
+// similar overrides for base plot attributes (and those added by components)
+exports.baseLayoutAttrOverrides = overrideAll({
+    plot_bgcolor: Plots.layoutAttributes.plot_bgcolor,
```
etpinard: I think this too could be :hocho: after #1869 where gl2d subplot with reused svg axis layering.

src/plots/gl2d/index.js
```diff
@@ -27,6 +30,21 @@ exports.attrRegex = constants.attrRegex;
...
+    hoverlabel: fxAttrs.hoverlabel
```
etpinard: Why does gl2d layout `hoverlabel` need to go through `doplot`?

src/plot_api/plot_api.js
alexcjohnson: Oh good catch - yes, the impliedEdits are there, I'll :hocho: this. One more down!

Though I notice from looking at `PlotSchema.get()` that the `{attr: undefined}` entries in `impliedEdits` (needed just for undo/redo) get stripped by `extendDeep`... not sure if that's worth worrying about, I wouldn't want to mangle `extendDeep` just for that purpose.

test/jasmine/tests/plotschema_test.js
alexcjohnson: Ha, can't remember what prompted me to dive into it but it was looking at rangeslider attributes. A committed schema could certainly be useful!

src/components/annotations/attributes.js
etpinard: Wait, where's is `doarraydraw` being used? I can't find it (even in the suppressed diffs :upside_down_face: )

src/plots/gl2d/index.js
alexcjohnson: That was basically in response to [this block](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plot_api/plot_api.js#L2223-L2227) that says "any axis attribute when the plot has gl2d needs to be treated as `doplot`." - But yeah, not the nicest structure, I'm open to suggestions.

src/plots/gl2d/index.js
alexcjohnson: good question, seems like this was needed to get a test to pass but I'll look at it again.

src/components/annotations/attributes.js
alexcjohnson: I think at one point I had `none` instead of `doarraydraw` but that looked weird. Then at another point I had a block just for debugging that checked that everything labeled `doarraydraw` was captured by an `arrayContainers` block and threw an error otherwise, but I removed that once I was satisfied that it was never getting hit... happy to change it if you think it would make more sense another way.

src/lib/regex.js
etpinard: We should perhaps try to sanatize all subplots attribute names in `cleanLayout`. Looks like we only sanatize cartesian, gl2d and gl3d axes at the moment.

src/plot_api/plot_schema.js
etpinard: :speaker: plotly.js is 100% with `new Number()` numbers :speaker:

src/plot_api/edit_types.js
etpinard: Actually, `overrideAll(attrs, editTypeOverride, 'nested')` is very readable, so maybe that third argument could accept values `'nested'` and 'from-root'` (instead of `true`).

src/plot_api/plot_schema.js
etpinard: Makes sense, thanks!

This means though that the order by which the attributes are displayed on https://plot.ly/javascript/reference/ and friends will change. I'm not sure if this is a big deal cc @cldougl

cldougl: that's no problem (hoping to have search implemented on this page pretty soon anyway)

etpinard: Oh sorry. False alarm. This patch is in `getTraceValObject` which isn't called during `PlotSchema.get()`.

src/plot_api/plot_api.js
etpinard: We do have another extend flavor [`extendDeepAll`](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/lib/extend.js#L37) which I think does just that: deep extends and keeps `undefined` values.

src/components/annotations/attributes.js
etpinard: If I understand correctly, `doarraydraw` is just a placeholder for edits in array containers that are handled by `applyContainerArrayChanges` in [`manage_array.js`](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plot_api/manage_arrays.js#L73). 

In this case, I prefer setting `editType: 'doarraydraw'` over `editType: 'none'`. Just make sure that this behavior is documented in the edit type `description` field (cc https://github.com/plotly/plotly.js/pull/1999#discussion_r138708703).

etpinard:
I say this [often](https://github.com/plotly/plotly.js/pulls?utf8=%E2%9C%93&q=is%3Apr%20PR%20of%20the%20year), but really, this deserves a PR of the year :trophy:. We've been taking (cc @bpostlethwaite ) about doing this for ~ 3 years now and finally seeing it happen feels like :tada: :champagne: :sparkles: 

List of blocking comments:

- confirmation that the schema changes don't break plotly.py graph object generation
- add a few assertions in the `test/jasmine/bundle_tests/` suites which checks that new `src/registry.js` logic does the right thing.
- adding `description` fields to edit types declarations: https://github.com/plotly/plotly.js/pull/1999#discussion_r138708703
- rename `coerce.valObjects` to avoid confusion: https://github.com/plotly/plotly.js/pull/1999#discussion_r138694738
- and yeah, as in the PR description: _I want to do a bit more testing of specific restyle/relayout calls to make sure we're doing the right redraws in all cases_

I noted a few things that didn't work out perfectly. I don't think any one them is blocking for now. But we might want to (or have to) clean them up in future iterations to make the whole dynamic schema + `editType` machinery more maintainable:

- those `*AttrOverrides` fields in base plot module objects. I feel with a bit more work and https://github.com/plotly/plotly.js/pull/1869), we won't need them at all. 
- `xaxis` / `yaxis` asymmetries (as in mentioned in https://github.com/plotly/plotly.js/pull/1999#discussion_r138679149). Looks like there isn't a clear solution at this stage, but it's something we'll probably have to improve at some point.
- add some sort of _edit convert_ method to component & baseplot modules to factor out colorbar [pixel<->faction](https://github.com/plotly/plotly.js/pull/1999#discussion_r138679149) and axis range [log<->linear](https://github.com/plotly/plotly.js/blob/7ea0d250ab6731be0b98d40ee0fd9e6bf667986c/src/plot_api/plot_api.js#L1870) logic similar to annotations/shapes/images `convert_coords.js`
- add a way to declare correspondence between attributes of different name across trace types (e.g. to generalize [`type` edits](https://github.com/plotly/plotly.js/blob/7ea0d250ab6731be0b98d40ee0fd9e6bf667986c/src/plot_api/plot_api.js#L1470) involving pie traces, `scattergeo` `lon`/`lat` should correspond to `x`/`y`, ...)
- figure which _special_ update keys (e.g. `'reverse'`, [`'swapaxis'`](https://github.com/plotly/plotly.js/blob/7ea0d250ab6731be0b98d40ee0fd9e6bf667986c/src/plot_api/plot_api.js#L1546), ... ) -- if any -- we want to keep in v2

-----

Moreover but unrelated, while reviewing `plot_api.js`, I also found that this [line](https://github.com/plotly/plotly.js/blob/f49ae5ea580d5826666574c041a744e89180d873/src/plot_api/plot_api.js#L1486) can now be removed (the `pie` clean up is now done [here](https://github.com/plotly/plotly.js/blob/f49ae5ea580d5826666574c041a744e89180d873/src/plot_api/plot_api.js#L1486)).

src/plot_api/plot_api.js
alexcjohnson: Nice, I didn't know about `extendDeepAll`, but that's exactly it - if you need to do a deep extend with attributes, it now needs to be changed to `extendDeepAll`. https://github.com/plotly/plotly.js/pull/1999/commits/238e2482f06810fd759d14aebee386724c0e1e45 does this, and also removes some of the deep extends entirely 🐎  - along with a test that we don't have any empty `impliedEdits`.

I'll also note that over the course of this PR I found a lot of places where we were doing
```js
extend(Flat|Deep)({}, {some object we're creating here}, {more stuff})
```
The first `{}` is unnecessary, because we don't mind mutating a new and disposable object. I removed it whenever I noticed this 🐎

src/traces/scattergl/attributes.js
```diff
@@ -58,7 +57,7 @@ module.exports = overrideAll({
             description: 'Sets the style of the lines.'
         }
     },
-    marker: extendDeep({}, colorAttributes('marker'), {
+    marker: extendFlat({}, colorAttributes('marker'), {
```
alexcjohnson: these attributes should be pretty well tested, but just to confirm this wasn't changing anything, I checked that `JSON.stringify(attrsDeep) === JSON.stringify(attrsFlat)` ✅

src/plot_api/edit_types.js
alexcjohnson: Turns out we weren't using the default no-containers behavior at all - I guess if that's the behavior you needed, some of the other attributes would have different editTypes already so you couldn't use `overrideAll` at all - so I changed it to only `'nested'` and `'from-root'` - https://github.com/plotly/plotly.js/pull/1999/commits/b9826c81c0b5ab49048e0e601825aa045402aa5e

src/components/annotations/attributes.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
         valType: 'boolean',
         role: 'info',
         dflt: true,
-        editType: 'docalcAutorange',
+        editType: 'calcIfAutorange',
```
rreusser: <3

src/plot_api/plot_schema.js
alexcjohnson: tested for y-only or x/y differences in https://github.com/plotly/plotly.js/pull/1999/commits/cb94e95d992c914faec1930da45b3a2a0af24b40

src/plot_api/plot_schema.js
alexcjohnson: `valObjectMeta` - and descriptions for `editType` and `impliedEdits` that make their way into the schema - in https://github.com/plotly/plotly.js/pull/1999/commits/96cc57fe5a7ce905901ecf18250ddcc2cc0c66f8

src/plots/gl2d/index.js
alexcjohnson: > Why does gl2d layout `hoverlabel` need to go through `doplot`?

didn't look into *why*, but I can confirm that currently this override is needed.

src/plot_api/plot_schema.js
```diff
@@ -64,7 +64,7 @@ exports.get = function() {
     return {
         defs: {
             valObjects: Lib.valObjectMeta,
-            metaKeys: UNDERSCORE_ATTRS.concat(['description', 'role']),
+            metaKeys: UNDERSCORE_ATTRS.concat(['description', 'role', 'editType', 'impliedEdits']),
```
alexcjohnson: With this addition, the schema still works as expected in `plotly.py` - thanks @cldougl for helping me figure out how to test that.

alexcjohnson: (and with the schema already in `dist` we won't need to build it in streambed anymore)

etpinard: I believe we should _still_ generate the schema in streambed, and remove the `*src` key [handler](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plot_api/plot_schema.js#L355) from `plot_schema.js`.

test/jasmine/bundle_tests/component_last_test.js
```diff
@@ -0,0 +1,15 @@
...
+var Scatter3d = require('@lib/scatter3d');
+var Filter = require('@lib/filter');
+var Calendars = require('@lib/calendars');
+
+var checkComponent = require('../assets/check_component');
+
+describe('Bundle with a component loaded after traces and transforms', function() {
+    'use strict';
+
+    Plotly.register([Bar, Scatter3d, Filter, Calendars]);
```
etpinard: thanks :tada:

src/plot_api/plot_api.js
```diff
@@ -2157,6 +2148,37 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, traces) {
     });
 };
 
+// empty out types for all axes containing these traces
+// so we auto-set them again
+var axLetters = ['x', 'y', 'z'];
+function clearAxisTypes(gd, traces, layoutUpdate) {
+    if(typeof traces === 'number') traces = [traces];
+    else if(!Array.isArray(traces) || !traces.length) {
```
etpinard: Do we really need this?

`traces` should be already sanatize [`helpers.coerceTraceIndices`](https://github.com/plotly/plotly.js/blob/790d470b901a38713728b5c9dc10bc624fed39ad/src/plot_api/helpers.js#L448) called [here](https://github.com/plotly/plotly.js/blob/790d470b901a38713728b5c9dc10bc624fed39ad/src/plot_api/plot_api.js#L1323).

src/plot_api/plot_api.js
```diff
@@ -2157,6 +2148,37 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, traces) {
...
+                var axAttr = ax._name;
+                var sceneName = ax._id.substr(1);
+                if(sceneName.substr(0, 5) === 'scene') {
+                    if(layoutUpdate[sceneName] !== undefined) continue;
+                    axAttr = sceneName + '.' + axAttr;
+                }
+                var typeAttr = axAttr + '.type';
+
+                if(layoutUpdate[axAttr] === undefined && layoutUpdate[typeAttr] === undefined) {
+                    Lib.nestedProperty(gd.layout, typeAttr).set(null);
```
etpinard: Much better than `delete /* */` :ok_hand:

package.json
```diff
@@ -119,6 +119,7 @@
     "gzip-size": "^3.0.0",
     "image-size": "^0.5.1",
     "jasmine-core": "^2.4.1",
+    "jsdom": "^9.5.0",
```
etpinard: Why not use the latest version [`jsdom@11`](https://www.npmjs.com/package/jsdom)?

alexcjohnson: Because `9.5.0` is what we have in streambed, and it looked like it was going to take me more than 5 minutes to figure out the new API 🙈 

But you're right, not exactly a good reason - I'll try to upgrade it.

src/plot_api/plot_api.js
alexcjohnson: I suppose I can move the `helpers.coerceTraceIndices` call into `restyle` and `update` instead of `_restyle`, then we can get rid of this.

tasks/util/make_schema.js
```diff
@@ -1,25 +1,24 @@
...
-                fs.writeFileSync(schemaPath, plotSchemaStr);
+        var w = new JSDOM('', {runScripts: 'dangerously'}).window;
 
-                console.log('ok ' + path.basename(schemaPath));
-            }
-        });
+        // jsdom by itself doesn't support getContext, and adding the npm canvas
+        // package is annoying and platform-dependent.
+        // see https://github.com/tmpvar/jsdom/issues/1782
+        w.HTMLCanvasElement.prototype.getContext = function() { return null; };
```
alexcjohnson: dunno why it didn't happen with the old version, but with the new jsdom without this patch you get an ugly but nonblocking error `Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)`

src/plot_api/plot_api.js
```diff
@@ -2152,10 +2154,6 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, traces) {
 // so we auto-set them again
 var axLetters = ['x', 'y', 'z'];
 function clearAxisTypes(gd, traces, layoutUpdate) {
-    if(typeof traces === 'number') traces = [traces];
```
etpinard: thanks!

tasks/util/make_schema.js
etpinard: Yep, no need to add a dependency for this :ok_hand:

test/jasmine/assets/check_ticks.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+var d3 = require('d3');
+
+module.exports = function checkTicks(axLetter, vals, msg) {
```
etpinard: I've been trying to centralized common custom assertions like this one in:

https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/custom_assertions.js

I'll let decide whether separate files or a common module is better for stuff like this.

test/jasmine/assets/negate_if.js
```diff
@@ -0,0 +1,16 @@
...
+ * Conditionally negate an expectation:
+ *
+ *   negateIf(expect(x), xIsSomethingElse).toBe(0);
+ *
+ * is equivalent to:
+ *
+ *   if(xIsSomethingElse) expect(x).not.toBe(0);
+ *   else expect(x).toBe(0);
+ */
+module.exports = function negateIf(expectation, negate) {
```
etpinard: Non-blocking but I wonder if there's a way to rewrite this a custom jasmine matcher. So that we have something like:

```js
expect(x).negateIf(negate).toBe(0)
```

test/jasmine/assets/check_ticks.js
alexcjohnson: oh that's nice - yeah, I'll put `checkTicks` in there.

test/jasmine/assets/negate_if.js
alexcjohnson: `expect(x).negateIf(negate).toBe(0)` would definitely be more readable - I'll give it a shot.

test/jasmine/tests/histogram_test.js
```diff
@@ -463,6 +463,10 @@ describe('Test histogram', function() {
     describe('plot / restyle', function() {
         var gd;
 
+        beforeAll(function() {
+            jasmine.addMatchers(customMatchers);
```
etpinard: (note-to-self comment) I'm getting tired of these `jasmine.addMatchers`. I'll try to find a way to add our custom matchers _globally_ once and for all.

rreusser: ❤️ 🎉 🥇 👏 💯

alexcjohnson: global addMatchers, and turning `negateIf` into a method (which I could only see how to do globally) ->https://github.com/plotly/plotly.js/pull/1999/commits/0729921645f88839df29fff32097f90a60a42143 hopefully it works right on circle 🙏

test/jasmine/assets/check_ticks.js
alexcjohnson: https://github.com/plotly/plotly.js/pull/1999/commits/e4227aad6073d4119cea1d67c3bac2a162701581

etpinard:
:dancer: :dancer: :dancer:

PR #2011, opened 2017-09-18
add minimum ify-loader version to readme

bpostlethwaite:
To account for recent fix found here: https://github.com/browserify/ify-loader/issues/5#issuecomment-330236079

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![circle ci](https://circleci.com/gh/plotly/plotly.js.png?&style=shield&circle-token=1f42a03b242bd969756fc3e53ede204af9b507c0)](https://circleci.com/gh/plotly/plotly.js)
 
 #### 📢 Announcement!
-#### Seats are still available for a 2 day, Plotly.js master class in NYC, November 16-17. 
+#### Seats are still available for a 2 day, Plotly.js master class in NYC, November 16-17.
 #### [Registration here](https://plotcon.plot.ly/workshops) 🎨 📈 🗽
 
 ***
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,7 +111,7 @@ Important: the plotly.js code base contains some non-ascii characters. Therefore
 
 #### Webpack Usage with Modules
 
-Browserify [transforms](https://github.com/substack/browserify-handbook#transforms) are required to build plotly.js, namely, [glslify](https://github.com/stackgl/glslify) to transform WebGL shaders and [cwise](https://github.com/scijs/cwise) to compile component-wise array operations. To make the trace module system work with Webpack, you will need to install [ify-loader](https://github.com/hughsk/ify-loader) and add it to your `webpack.config.json` for your build to correctly bundle plotly.js files.
+Browserify [transforms](https://github.com/substack/browserify-handbook#transforms) are required to build plotly.js, namely, [glslify](https://github.com/stackgl/glslify) to transform WebGL shaders and [cwise](https://github.com/scijs/cwise) to compile component-wise array operations. To make the trace module system work with Webpack, you will need to install [ify-loader@v1.1.0+](https://github.com/hughsk/ify-loader) and add it to your `webpack.config.json` for your build to correctly bundle plotly.js files.
```
rreusser: Not a big deal, but appears `brfs` [is also used](https://github.com/mapbox/mapbox-gl-shaders/blob/b51b85ffb8c512e228c36c5405293ce51d123519/package.json#L24), unless I'm mistaken. Though that module is deprecated anyway. Also, thought brfs would show up in more places.

bpostlethwaite: Do users wanting to build plotly.js need to explicitly add brfs?

rreusser: To the best of my knowledge, they don't need to add any of the transforms explicitly, do they? It's picked up in the dev dependencies and `package.json` of the corresponding module, right? Like you don't have to specify anything about `glslfify`, for one, to get `ify-loader` to use glslify when building plotly.js, I think…? Correct me if I'm wrong.

bpostlethwaite: that's correct. I think we should simplify this section. We don't need to mention any of this stuff other than "use ify-loader" and provide a simple webpack config example.

rreusser: For reference, here's [what I was using](https://github.com/rreusser/plotly-webpack/blob/master/webpack.config.js) as a minimal example:

```javascrip
// index.js:
var Plotly = require('plotly.js');
```

And the `webpack.config.js`:

```javascript
module.exports = {
    entry: "./index.js",
    output: {
        path: __dirname,
        filename: "plotly-bundle.js"
    },
    module: {
        loaders: [{
            test: /\.js$/,
            loader: 'ify-loader'
        }]
    }
};
```

Maybe just suggest the `loaders` section?

README.md
```diff
@@ -109,9 +109,23 @@ Important: the plotly.js code base contains some non-ascii characters. Therefore
...
 
 
-#### Webpack Usage with Modules
+#### Building plotly.js with Webpack
 
-Browserify [transforms](https://github.com/substack/browserify-handbook#transforms) are required to build plotly.js, namely, [glslify](https://github.com/stackgl/glslify) to transform WebGL shaders and [cwise](https://github.com/scijs/cwise) to compile component-wise array operations. To make the trace module system work with Webpack, you will need to install [ify-loader](https://github.com/hughsk/ify-loader) and add it to your `webpack.config.json` for your build to correctly bundle plotly.js files.
+For plotly.js to build with Webpack you will need to install [ify-loader@v1.1.0+](https://github.com/hughsk/ify-loader) and add it to your `webpack.config.json`. This adds Browserify transform compatibility to Webpack which is necessary for some plotly.js dependencies.
+
+A repo that demonstrates how to build plotly.js with Webpack can be found [here](https://github.com/rreusser/plotly-webpack). In short add `ify-loader` to the `module` section in your `webpack.config.js`:
+```json
```
rreusser: s/json/js/

etpinard:
LGTM :dancer:

rreusser:
I've fixed up my instructions to turn this from an issue-demonstration repo into more of a webpack-howto repo: https://github.com/rreusser/plotly-webpack

PRs for clarity welcome, though it's pretty simple as-is.

Issue #2014, opened 2017-09-19
Sankey hover text placement


jackparmer:
@alexcjohnson noticed a hover placement issue with Sankey - here the 633 appears on top of the white box instead of inside it. Not sure if it's a JSON construction issue or a Plotly.js issue.

![image](https://user-images.githubusercontent.com/1865834/30574500-5d0e73dc-9cc8-11e7-803b-d577e8f65488.png)

The wapo hover text can also go off into the ether and does not always seem to point at the selected line:

![image](https://user-images.githubusercontent.com/1865834/30574529-88e96ec6-9cc8-11e7-81ff-748bd5144dc4.png)

Both of these examples are from https://medium.com/@plotlygraphs/4-interactive-sankey-diagram-made-in-python-3057b9ee8616

/cc @monfera @alishobeiri

monfera:
Looks like the out-of-box number happens on items that show a `Source:` and a `Target:` line, but no title above it, as on the above example, because the few items with a `label` are fine:
![image](https://user-images.githubusercontent.com/1548516/30584156-2e9abaf4-9d29-11e7-8377-45ccaa9984f6.png)

This codepen example also has links both with and without `label`, and both render fine:

![image](https://user-images.githubusercontent.com/1548516/30584366-c6364cd4-9d29-11e7-8c79-18b0f925372d.png)

![image](https://user-images.githubusercontent.com/1548516/30584334-b29b323e-9d29-11e7-9ed2-1974db25b694.png)

I'll look at how the data is different here.

monfera:
Reopening to be able to link the other fix (closing only when that's done too)

monfera:
@alexcjohnson the only remaining problem is the sometimes misplaced tooltip, seen when the browser window [here](https://plot.ly/~alishobeiri/1367.embed) is wide and `div.svg-container` gets a `margin: auto`, as inlined style. Any quick thought on where this `auto` is coming from? I copied `data` and `layout` and I didn't get that. I can look it up if it'd take time.

monfera:
Added an effective [fix](https://github.com/plotly/plotly.js/pull/2027), and, depending on the answer I'd need to add an appropriate test. It looks like our [guideline](https://plot.ly/javascript/responsive-fluid-layout/) on fluid layout tampers with the `gd` `margin` rather than the `paperdiv` (=== `div.svg-container`) margin.

monfera:
[...]

PR #2016, opened 2017-09-19
Sankey default link label

monfera:
Fixes https://github.com/plotly/plotly.js/issues/2014

### DIFF
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -145,7 +145,7 @@ module.exports = function plot(gd, calcData) {
             y: hoverCenterY - rootBBox.top,
             name: d3.format(d.valueFormat)(d.link.value) + d.valueSuffix,
             text: [
-                d.link.label,
+                d.link.label || '',
                 ['Source:', d.link.source.label].join(' '),
                 ['Target:', d.link.target.label].join(' ')
             ].filter(renderableValuePresent).join('<br>'),
--- a/test/jasmine/tests/sankey_test.js
+++ b/test/jasmine/tests/sankey_test.js
@@ -112,6 +112,8 @@ describe('sankey tests', function() {
                 expect(fullTrace.link.target)
                     .toEqual([], 'presence of link target array is guaranteed');
 
+                expect(fullTrace.link.label)
+                    .toEqual([], 'presence of link target array is guaranteed');
             });
 
         it('\'Sankey\' specification should have proper types',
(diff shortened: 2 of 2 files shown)


monfera:
Changes 
image

to

image

src/traces/sankey/defaults.js
```diff
@@ -31,13 +31,13 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
     }));
 
-    coerce('link.label');
     coerce('link.source');
     coerce('link.target');
     coerce('link.value');
     coerce('link.line.color');
     coerce('link.line.width');
 
+    coerce('link.label', traceOut.link.value.map(function() {return '';}));
```
alexcjohnson: 🐎  (also for `link.color`) we're always doing this `map` even if we don't need the default. Would it be worth either:
- running that after the `coerce` call, ie:
```js
var linkLabel = coerce('link.label');
if(!linkLabel) traceOut.link.label = traceOut.link.value.map(function() {return '';});
```
- building into `coerce` that you can pass a function as `dflt` and it will only be called if needed?
```js
coerce('link.label', function() { return traceOut.link.value.map(function() {return '';}); });
```
thoughts?

alexcjohnson: Also typically when we rely on a previously coerced value to modify later behavior, usually we just save it from its coercion: `var linkValue = coerce('link.value')`

Which brings up the question: some of these inputs are strictly necessary in order to have meaningful output. What's the minimal set here, without which we should [automatically hide the trace](https://github.com/plotly/plotly.js/blob/790d470b901a38713728b5c9dc10bc624fed39ad/src/traces/scatter/defaults.js#L34-L37)? If there's no `link.value` this `map` will throw an error, which we definitely don't want.

alexcjohnson: Oh I guess you default to `[]` so it won't error... but it still doesn't seem meaningful right?

monfera: Good point, definitely worth avoiding useless work. Whenever adding some array attributes, I felt it'd be good to be able to specify a default value for the array elements (in `attributes.js`) rather than just the whole array. But your suggestion is more flexible because it allows a function (default element values depending on other values) so I'll

etpinard: @monfera why don't you move this piece of logic to the `calc` step?

monfera: @etpinard I think of array values as conceptually default-able (though there's no `elementDefault` in `attributes.js`), and saw usages elsewhere that were also in `defaults.js`. But I can do it in `calc` or even in the tooltip caller code `linkLabel || ''` - after all, only the tooltip has a problem with `undefined`, which is external to `sankey`. As @alexcjohnson had an alternative suggestion, I wonder which direction would be best in this case.

alexcjohnson: Ah, if you can handle it entirely at the point of hover - so no fallback processing needs to happen at all during plotting - that seems like the best of all, performance-wise, and we do have other examples of arrays that may or may not exist and we leave it until the value is needed to test this rather than trying to fill it in at `supplyDefaults` or even `calc`.

monfera: The fix got pivoted to this consensus, the `assertLabel` test function now also checks whether the value no longer breaches the tooltip bounding box for a new test case (and other tests that use it already). The test properly breaks when I comment out the fix.

monfera: The commit is in, that's all there is to it, plus the tests 😄

alexcjohnson:
Super - thanks for the quick turnaround! 💃

Issue #2015, opened 2017-09-19
Minimal Sankey values don't visibly render


monfera:
See bottom right corner:

![image](https://user-images.githubusercontent.com/1548516/30585887-7e130bcc-9d2e-11e7-8dc6-c3c0973d063f.png)

Both the node and the capture zone are rendered, but the node has a height of 0px, I'll check what's going on as I remember the logic was to make it at least 1px even with negligible `value`.

Saw it when working on https://github.com/plotly/plotly.js/issues/2014

monfera:
Desired view, from the fix in progress:

![image](https://user-images.githubusercontent.com/1548516/30594467-72760f6a-9d4e-11e7-8600-01c0d4315af8.png)

PR #2017, opened 2017-09-19
Sankey: render a thin but visible node even if `value` is very low

monfera:
Fixes https://github.com/plotly/plotly.js/issues/2015 (see before/after example there)
Issue found when looking at https://github.com/plotly/plotly.js/issues/2014

Jasmine test produces this DOM tree:

image

### DIFF
--- a/src/traces/sankey/render.js
+++ b/src/traces/sankey/render.js
@@ -182,7 +182,7 @@ function nodeModel(uniqueKeys, d, n) {
         zoneThicknessPad = c.nodePadAcross,
         zoneLengthPad = d.nodePad / 2,
         visibleThickness = n.dx + 0.5,
-        visibleLength = n.dy - 0.5;
+        visibleLength = Math.max(1, n.dy - 0.5);
 
     var basicKey = n.label;
     var foundKey = uniqueKeys[basicKey];
--- a/test/jasmine/tests/sankey_test.js
+++ b/test/jasmine/tests/sankey_test.js
@@ -311,6 +311,29 @@ describe('sankey tests', function() {
                     done();
                 });
         });
+
+        it('\'node\' remains visible even if \'value\' is very low', function(done) {
+
+            var gd = createGraphDiv();
+            var minimock = [{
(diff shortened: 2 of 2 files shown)


alexcjohnson:
Good catch, and really nice test!
💃

alexcjohnson:
though - just a nitpick, but it doesn't look like the node is aligned quite properly with the link in this case?

monfera:
I can put a bit more into alignment but maybe it could be an independent ticket:

The alignments, in general, aren't perfect. When developing the `sankey`, there were tradeoffs due to these factors:

- crisp, uniform thickness node borders requires that SVG `shape-rendering` is set to `crispEdges` - compare [Mike's original](https://bost.ocks.org/mike/sankey/) with [our take](https://codepen.io/monfera/full/jLVNxM/) which gives more consistent apparent box borders
- even if we have this, constant node border (stroke) thickness and constant color requires that the `rect` borders are snapped - for example, moving a line half a pixel can result in a less contrasty, blurred line (it's done with things like adding 0.5px and ceil/truncate)
- there were some cross-browser inconsistencies (can't remember)
- `d3-sankey` is a black box and returns node geometry and link geometry separately (also, it has no awareness of pixels or snapping)
- unlike `d3-sankey`, we more clearly support [multiedges](https://github.com/d3/d3-sankey/pull/19) - for this reason, we fill shade rather than stroke shade the links - and various options lead to various overlaps between adjacent multiedges, so if we further tweak alignment, we also need to ensure there's no negative banding artifact with multiedges

alexcjohnson:
@monfera I see what you're talking about with regular (non-tiny) nodes, and I can see why you did it the way you did. I think the normal case is fine, strikes a good balance between the competing visual requirements. But the new tiny one really seems to be hanging off the bottom of the link. I think it's worth spending a few minutes fixing it just in this special case.

monfera:
@alexcjohnson ah got it, will do!

alexcjohnson:
#2021 addresses my comment, so this should be good to go! 💃

PR #2022, opened 2017-09-19
Rewrite legend visibility toggling

rreusser:
Legend visibility toggling didn't work at all with `groupby`, which led to mostly a rewrite. By happenstance, it toggled them as a group, which wasn't really the desired behavior. More problematically though, it also threw lots of `Uncaught bad container` errors because it really just wasn't accounted for at all. This PR rewrites legend visibility toggling and tries to preserve existing behavior while fixing `groupby`.

In particular, one tricky case is `legendgroup`, both toggling and isolating:

regular-trace-toggle

And the other is `groupby` transforms, both toggling and isolating:

groupby-visibility-toggling

### DIFF
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -573,11 +573,11 @@ function handleClick(g, gd, numClicks) {
         var hasLegendgroup = legendgroup && legendgroup.length;
         var traceIndicesInGroup = [];
         var tracei;
-        if (hasLegendgroup) {
-            for (i = 0; i < fullData.length; i++) {
+        if(hasLegendgroup) {
+            for(i = 0; i < fullData.length; i++) {
                 tracei = fullData[i];
-                if (!tracei.visible) continue;
-                if (tracei.legendgroup === legendgroup) {
+                if(!tracei.visible) continue;
+                if(tracei.legendgroup === legendgroup) {
                     traceIndicesInGroup.push(i);
                 }
             }
@@ -599,9 +599,9 @@ function handleClick(g, gd, numClicks) {
                     break;
             }
 
-            if (hasLegendgroup) {
-                for (i = 0; i < fullData.length; i++) {
-                    if (fullData[i].visible && fullData[i].legendgroup === legendgroup) {
+            if(hasLegendgroup) {
+                for(i = 0; i < fullData.length; i++) {
+                    if(fullData[i].visible && fullData[i].legendgroup === legendgroup) {
                         setVisibility(fullData[i], nextVisibility);
                     }
                 }
@@ -611,17 +611,17 @@ function handleClick(g, gd, numClicks) {
         } else if(numClicks === 2) {
             // Compute the clicked index. expandedIndex does what we want for expanded traces
             // but also culls hidden traces. That means we have some work to do.
-            var clickedIndex;
+            var clickedIndex, isIsolated, isClicked, isInGroup, otherState;
             for(clickedIndex = 0; clickedIndex < fullData.length; clickedIndex++) {
                 if(fullData[clickedIndex] === fullTrace) break;
             }
 
-            var isIsolated = true;
+            isIsolated = true;
             for(i = 0; i < fullData.length; i++) {
-                var isClicked = fullData[i] === fullTrace;
-                if (isClicked) continue;
+                isClicked = fullData[i] === fullTrace;
+                if(isClicked) continue;
 
-                var isInGroup = (hasLegendgroup && fullData[i].legendgroup === legendgroup);
+                isInGroup = (hasLegendgroup && fullData[i].legendgroup === legendgroup);
 
                 if(!isInGroup && fullData[i].visible === true && !Registry.traceIs(fullData[i], 'notLegendIsolatable')) {
                     isIsolated = false;
@@ -642,9 +642,9 @@ function handleClick(g, gd, numClicks) {
                         setVisibility(fullData[i], true);
                         break;
                     case true:
-                        var otherState = isIsolated ? true : 'legendonly';
-                        var isClicked = fullData[i] === fullTrace;
-                        var isInGroup = isClicked || (hasLegendgroup && fullData[i].legendgroup === legendgroup);
+                        otherState = isIsolated ? true : 'legendonly';
+                        isClicked = fullData[i] === fullTrace;
+                        isInGroup = isClicked || (hasLegendgroup && fullData[i].legendgroup === legendgroup);
                         setVisibility(fullData[i], isInGroup ? true : otherState);
                         break;
                 }
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -618,12 +618,8 @@ function handleClick(g, gd, numClicks) {
         } else if(numClicks === 2) {
             // Compute the clicked index. expandedIndex does what we want for expanded traces
             // but also culls hidden traces. That means we have some work to do.
-            var clickedIndex, isIsolated, isClicked, isInGroup, otherState;
-            for(clickedIndex = 0; clickedIndex < fullData.length; clickedIndex++) {
-                if(fullData[clickedIndex] === fullTrace) break;
-            }
-
-            isIsolated = true;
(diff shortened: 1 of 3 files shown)


rreusser:
Oh, and to preempt the question, the answer is, "Yes." 😄 

combo

src/components/legend/draw.js
```diff
@@ -520,52 +570,116 @@ function handleClick(g, gd, numClicks) {
...
+            var nextVisibility;
+
+            switch(fullTrace.visible) {
+                case true:
+                    nextVisibility = 'legendonly';
+                    break;
+                case false:
+                    nextVisibility = false;
+                    break;
+                default:
```
etpinard: :hocho: that `default` ?

src/components/legend/draw.js
```diff
@@ -480,26 +480,76 @@ function setupTraceToggle(g, gd) {
...
+            var carr = carrs[fullInput.index];
+            if(!carr) {
+                var groupbyIndices = Registry.getTransformIndices(fullInput, 'groupby');
+                var lastGroupbyIndex = groupbyIndices[groupbyIndices.length - 1];
+                carr = Lib.keyedContainer(fullInput, 'transforms[' + lastGroupbyIndex + '].styles', 'target', 'value.visible');
+                carrs[fullInput.index] = carr;
+            }
+
+            // If not specified, assume visible:
+            var curState = carr.get(fullTrace._group) || true;
```
etpinard: Why does this happen?

etpinard: this `===` that `|| true` fallback.

src/components/legend/draw.js
```diff
@@ -480,26 +480,76 @@ function setupTraceToggle(g, gd) {
...
+    }
+
+    function setVisibility(fullTrace, visibility) {
+        var fullInput = fullTrace._fullInput;
+        if(Registry.hasTransform(fullInput, 'groupby')) {
+            var carr = carrs[fullInput.index];
+            if(!carr) {
+                var groupbyIndices = Registry.getTransformIndices(fullInput, 'groupby');
+                var lastGroupbyIndex = groupbyIndices[groupbyIndices.length - 1];
+                carr = Lib.keyedContainer(fullInput, 'transforms[' + lastGroupbyIndex + '].styles', 'target', 'value.visible');
```
etpinard: `keyedContainer` for the win :trophy: 

By the way, is `carr` short for _container array_?

etpinard: ... no need to rename `carr` to something more explicit, I'm just curious.

src/components/legend/draw.js
```diff
@@ -520,52 +570,116 @@ function handleClick(g, gd, numClicks) {
...
+                    break;
+                default:
+                case 'legendonly':
+                    nextVisibility = true;
+                    break;
+            }
+
+            if(hasLegendgroup) {
+                for(i = 0; i < fullData.length; i++) {
+                    if(fullData[i].visible && fullData[i].legendgroup === legendgroup) {
```
etpinard: maybe `fullData[i].visible !== false` instead of `fullData[i].visible` would be best, to make things more explicit. It is easy to forget that `visible` supports two truthy values.

src/components/legend/draw.js
```diff
@@ -520,52 +570,116 @@ function handleClick(g, gd, numClicks) {
...
+                }
+            } else {
+                setVisibility(fullTrace, nextVisibility);
+            }
         } else if(numClicks === 2) {
-            var sameAsLast = true;
+            // Compute the clicked index. expandedIndex does what we want for expanded traces
+            // but also culls hidden traces. That means we have some work to do.
+            var clickedIndex, isIsolated, isClicked, isInGroup, otherState;
+            for(clickedIndex = 0; clickedIndex < fullData.length; clickedIndex++) {
```
etpinard: Isn't `clickedIndex` simply `fullTrace.index`?

src/components/legend/draw.js
```diff
@@ -520,52 +570,116 @@ function handleClick(g, gd, numClicks) {
 
         Plotly.relayout(gd, 'hiddenlabels', hiddenSlices);
```
etpinard: I think we can return early after `Plotly.relayout` for pie traces.

etpinard: Ignore :arrow_heading_up: I missed to if block levels.

src/components/legend/draw.js
```diff
@@ -520,52 +570,116 @@ function handleClick(g, gd, numClicks) {
...
+                key = updateKeys[j];
+                val = attrUpdate[key] = attrUpdate[key] || [];
+                val[carrIdx[i]] = update[key];
+            }
+        }
+
+        // The length of the value arrays should be equal and any unspecified
+        // values should be explicitly undefined for them to get properly culled
+        // as updates and not accidentally reset to the default value. This fills
+        // out sparse arrays with the required number of undefined values:
```
etpinard: Very nice touch :ok_hand:

etpinard:
Very lovely PR :heart: with strong tests :hammer_and_pick: 

I made a few comments; I'll make a second review after reading your answers.

As an aside, `legend/draw.js` now probably has more click & edit logic than actual _drawing_ logic. Do you think factoring out click & edit logic into a separate module (i.e. a few methods that take in clicked data and return restyle/relayout update objects) would make the legend code easier to maintain?

src/components/legend/draw.js
rreusser: Yes, it's somewhat of a holdover. It's not great. Maybe `kcont`?

src/components/legend/draw.js
rreusser: Good catch. This isn't a great conditional. If it's really `fullData`, it should *always* be present and what it means to be. I think I'll just remove this.

src/components/legend/draw.js
rreusser: Let's see… it might be. I need to think about this just a bit. `fullTrace.index` vs. `fullTrace._fullInput.index` vs. `fullTrace._expandedIndex` obey slightly different rules. I think more often that not the answer was to avoid at all costs needing the index at all. I'll give the above a try.

src/components/legend/draw.js
etpinard: `carr` or `kcont`, you're call.

src/components/legend/draw.js
rreusser: ✅

src/components/legend/draw.js
rreusser: Haha couldn't remove it. Modified it to be a bit more targeted/explicit. The issue is that the container doesn't always have `visible` defined. "Oh, so just use `fullTrace.visible`." That didn't work either. Broke some tests. Aimed for some middle ground that passes all the tests.

src/components/legend/draw.js
rreusser: `kcont`. ✅

src/components/legend/draw.js
rreusser: ✅

src/components/legend/draw.js
rreusser: Ah, good catch. `clickedIndex` isn't actually used at all, which was the end result of trying to get this right (i.e. don't use it at all). I've removed this code.

src/components/legend/draw.js
```diff
@@ -516,20 +516,28 @@ function handleClick(g, gd, numClicks) {
...
-            // If not specified, assume visible:
-            var curState = carr.get(fullTrace._group) || true;
+            var curState = kcont.get(fullTrace._group);
+
+            // If not specified, assume visible. This happens if there are other style
+            // properties set for a group but not the visibility. There are many similar
+            // ways to do this (e.g. why not just `curState = fullTrace.visible`??? The
+            // answer is: because it breaks other things like groupby trace names in
+            // subtle ways.)
+            if(curState === undefined) {
```
etpinard: Thanks for the comment :books:

etpinard:
:dancer: thanks very much @rreusser :tada:

Issue #1996, opened 2017-09-10
geo.scope='usa' with points outside USA gives console errors on hover


jackparmer:
https://plot.ly/~jackp/17904/

Change `geo.scope` to `north america` and hover works again

![image](https://user-images.githubusercontent.com/1865834/30245796-f21d0e10-95b2-11e7-8f6f-b5df0da3aa77.png)

jackparmer:
Is `usa` for `geo.scope` only meant to be used for `choropleth`?

etpinard:
> Is usa for geo.scope only meant to be used for choropleth?

Nop. https://plot.ly/~jackp/17904/ should work.

Thanks

etpinard:
@jackparmer found a fix for this. Is this urgent? Or can it wait 1-2 weeks? If so, I'll include it only in my upcoming geo refactor PR.

Issue #1995, opened 2017-09-08
add a miter limit on scattergeo


alexcjohnson:
Cartesian scatter lines have a miter limit of 2 https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/plot.js#L42

This is important for sharp turns in the data for two related reasons:
- with only lines, it looks like the data point is more extreme than it really is
- with markers+lines, the line seems to poke through the marker and then leave from the same side it came from:
![screen shot 2017-09-08 at 3 52 39 pm](https://user-images.githubusercontent.com/2678795/30228943-e5fc55c4-94ad-11e7-8546-4d56ff40c2f7.png)
(in case it's not clear: this is a black point on a red line on an existing scattergeo trace)

We should add this to scattergeo too, and investigate whether any other trace types need it. I suspect not though: scatterternary and scattercarpet inherit from regular scatter so should already have it, and the others are all webgl (cc @dfcreative I think we discussed this already at some point, it's at the very least a completely different kind of implementation)

cc @etpinard 

I'll also note (in case this question comes up regarding any other trace types) that for contour lines we chose an even shorter miter limit of 1, because even if you specify no smoothing these are always modeling the data as if it were continuous and smooth - so we want to round off even sort-of-sharp corners.

Issue #1698, opened 2017-05-18
Can't set longitude range between [160, -160]


chriddyp:
Trying to center a scatter map around Fiji. I want the longitude to be between 140 and -140, like this image (I manually zoomed and panned to focus here)

![image](https://cloud.githubusercontent.com/assets/1280389/26181706/399e3f52-3b40-11e7-8ee0-ada17e7e1022.png)

Setting the longitude range to `[140, -140]` does not seem to work. I've also tried rolling over with `[140, 220]` to no avail. Here is a Codepen: https://codepen.io/chriddyp/pen/OmovzL?editors=1010. Here is a graph in the chart editor: https://plot.ly/create/?fid=chris:18383

🎩  tip to @jackparmer who originally discovered this.

etpinard:
Looks like all custom `lonaxis` ranges that cross the antimeridian (where lon = -180 = 360) are broken.

This might be a good reason to start using d3 v4's [`d3-geo`](https://github.com/d3/d3-geo) module which ships with a [`fitExtent`](https://github.com/d3/d3-geo#projection_fitExtent) method that probably does a better [job](https://github.com/d3/d3-geo/blob/2c51b71d55eb68c4f571b06c92ea502342975849/src/projection/fit.js) than my ad hoc [attempt](https://github.com/plotly/plotly.js/blob/master/src/plots/geo/set_scale.js).

etpinard:
@chriddyp I found a way to make this work without waiting for a patch:

- set `projection.rotation.lon` to `-180` (or `180`) - the algo which sets its default is off currently
- set `lonaxis.range` to `[140, 220]` - the `[140, -140]` does not work.
- see https://codepen.io/etpinard/pen/pWgWLv?editors=0010

chriddyp:
cc @jackparmer in case you still need to make this example ^^

Issue #1486, opened 2017-03-16
Add `geo.center` attribute to specify map view without rescaling


brian428:
When using an autosized plot, focusing on a map area by using ranges on scattergeo lat long causes the plot width and height to change. I don't think this should happen. The plot should always use all available space. 

I created a codepen to demonstrate the issue: http://codepen.io/brian428/pen/EWbKgQ

Just change the `setRange` flag to `true` to see the effect it has on the plot size.

etpinard:
That's the intended behavior I'm afraid. Can I _why_ this shouldn't happen?

brian428:
I assumed that the geo range would set the extent of what you SEE in the map, but I don't understand why it would change the width and height. If you don't set a range and you zoom into the plot, it doesn't alter the size. So if this is the desired behavior when setting the range, is there some way to specify the visible extent being shown on the map without changing the size of the map itself?

etpinard:
@brian428 that's more like it! Pro tip: opening an issue saying something _should_ work a certain way without explanation is one way to make a OSS maintainer pissed off.  On that topic, you might want to read [this](https://nolanlawson.com/2017/03/05/what-it-feels-like-to-be-an-open-source-maintainer/).

------

As you point out, geo subplots are missing a few attributes to fully described the state. At first, I was convinced that `lonaxis.range` and `lataxis.range` were and should be enough, but no. 

To better explain the problem here's a little background information: in general one can't satisfy (or _fill_) both given a width and height without distorting the map. That's because a map's width and height aren't independent: they are constrained by the map projection. So, plotly fills one dimension to its maximum (in your case your map has full height) and the width is then determined by the map projection.  When setting the lon/lat ranges you essentially set the bounds for which width/height constrain is applied. If you're interested, the logic is [here](https://github.com/plotly/plotly.js/blob/master/src/plots/geo/set_scale.js).

But now, how does one show a subset of a map without rescaling it? At the moment, you can't. So I'm thinking about adding a `geo.center` attribute similar to how `mapbox` subplot are specified. Together with `geo.scale` one could then get your desired result.

It's about time I give this a shot as this will also resolved https://github.com/plotly/plotly.js/issues/292 fairly easily.

brian428:
[...]

PR #2030, opened 2017-09-22
Geo refactor and lasso/select-box selections

etpinard:
In brief, this PR brings geo subplots up-to-speed with other subplots. :rocket: 

Geo subplots have long suffered from a few design mistakes (e.g. as discussed https://github.com/plotly/plotly.js/issues/292 and https://github.com/plotly/plotly.js/issues/1486). This PR attempts to solve these issues - which are required to make lasso/select-box selections work.

Since the geo subplot constructor code was mostly rewritten (see https://github.com/plotly/plotly.js/commit/4c1ddb80dd39ae6d1ba5d51d21b95b36919a8bc5), I decided to include the new lasso/select-box logic in this PR as well. I hope the reviewers won't mind. You may consider geo as a _new_ subplot type and this its introduction PR. 

----

#### Stuff implemented:
- :globe_with_meridians:  `geo.center` a new attribute which should've been there from the start https://github.com/plotly/plotly.js/issues/1486
- :microscope:  `scattergeo` and `choropleth`(selections about centroid) lasso and select box

peek 2017-09-21 23-10

#### Stuff fixed:
- :tada: no view reset on updates https://github.com/plotly/plotly.js/issues/292 (thanks to `geo.center`) :lock: in https://github.com/plotly/plotly.js/commit/3bd89d32ce50cbf771a539cf6dce013d26d552da
- :white_check_mark: `scattergeo` markers outside 'usa' scope now render correctly https://github.com/plotly/plotly.js/issues/1996 :lock:  in https://github.com/plotly/plotly.js/commit/eaf44a8fb898e55a3f8679469425ab1ade7d6f7b
- :world_map:  cross anti-meridian `lonaxis` ranges now behave correctly https://github.com/plotly/plotly.js/issues/1698 :lock: in https://github.com/plotly/plotly.js/commit/e59c90b0609693b7f62a5cee247b7acb40fa61f7
- :mag: add miter limit to scattergeo lines https://github.com/plotly/plotly.js/issues/1995 :lock: in https://github.com/plotly/plotly.js/commit/825e24d2065114be8f5b045d63cd5235686f8ff1
- :mag_right:  `marker.opacity`wasn't applied properly on `scattergeo` bubbles :lock: in https://github.com/plotly/plotly.js/commit/7e5bb41fa699e49f0c0eff84e4ea58cb044ccb20

#### Baseline updates:

Unfortunately, **all** geo baselines had to be updated. 

- Most geo subplots now render a few pixels off their position in the old baselines (see https://github.com/plotly/plotly.js/commit/a30beeaf306b3774d318bbc34cdb7829901919da). 
- The center of some scoped maps are shifted (see https://github.com/plotly/plotly.js/commit/be59f693383b0b746bc8947ad8e8c2c5cccb239d) due to the new default logic (of https://github.com/plotly/plotly.js/commit/a0be08795acfc0a976f7ef141a30a4a1d4e65704). 
- Maps with conic projections needed to be updated for similar reasons (see https://github.com/plotly/plotly.js/commit/32a29c7191efc9a96faca97ccad8a30733f09d9c). 
- Finally, `geo_fill` was adapted to showcase the new `geo.center` attribute (see https://github.com/plotly/plotly.js/commit/36833d4446c3b86f447407773ab8446db24e883a)

#### Stuff I could do in this PR, but could wait for v2:
- Rename (+ `cleanLayout` map) `geo.projection.scale` :arrow_right:  `geo.zoom` (as in mapbox subplots)
- Rename (+ `cleanData` map) choropleth `z` :arrow_right:  `values` (similar to pie traces)

#### Open items:
- Base layer ordering changes whenever a choropleth trace is present on the subplot (see this [block](https://github.com/plotly/plotly.js/commit/4c1ddb80dd39ae6d1ba5d51d21b95b36919a8bc5#diff-b4732bdd35eeb4830531de6212e422e2R239)). This logic is meant to preserve backward compatibility (i.e. to mimic what that [block](https://github.com/plotly/plotly.js/blob/60b445d2916547f1d1a548d64209989688d7a617/src/traces/choropleth/plot.js#L65-L72) does in the current `master`). But really, we should make the base layer ordering consistent for all trace type combinations in v2. But in the meantime, we should make the base layer ordering customizable (similar to `axis.layer`).
- Rewrite drag interactions using `dragElement` (i.e. get rid of `d3.behavior.drag`) and implement dragmode: 'zoom'` zoombox
- Make `scattergeo/plot.js` update inner nodes instead of purging them. This is required for smooth animations and should be a performance boost.

#### Stuff I believe should wait for v2:
- Make geo subplot respect the `scrollZoom` config option https://github.com/plotly/plotly.js/issues/143. Unfortunately. geo assumes `scrollZoom: true`, but the `scrollZoom` default is `faise`. We could add some logic to make `scrollZoom` default to `true` if a geo subplot is present to preserve backward compatibility, but that hardly sounds worth it (to me at least).
- Replace `PlotlyGeoAssets` global with some field on `Plotly`. No need to have another global variables, but this isn't a huge priority.

-----

cc @alexcjohnson @rreusser  @chriddyp @jackparmer @cldougl 
and @cpsievert (this here might make him start using `geo` in the R API :stuck_out_tongue_closed_eyes:  )


src/plots/geo/constants.js
```diff
@@ -135,12 +135,23 @@ params.fillLayers = ['ocean', 'land', 'lakes'];
...
+];
+
+params.layersForChoropleth = [
+    'bg',
+    'ocean', 'land',
+    'subunits', 'countries', 'coastlines',
+    'lataxis', 'lonaxis', 'frame',
+    'backplot',
+    'rivers', 'lakes',
+    'frontplot'
```
alexcjohnson: I'm a bit confused by the choropleth distinction here... it seems to me that either:
- adding a choropleth (even if it only shows up in some other part of the plot) will reorder other layers such that there are visual changes, which would be bad, right?
- if there are no visual changes, we should be able to use the same layer structure whether or not there is a choropleth present - with at most an extra layer just for choropleth, separate from backplot (though I doubt it, seems like choropleth is the only thing that uses backplot currently, right?)

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+    var t = projection.translate();
+
+    if(
+        !isFinite(b[0][0]) || !isFinite(b[0][1]) ||
+        !isFinite(b[1][0]) || !isFinite(b[1][1]) ||
+        isNaN(t[0]) || isNaN(t[0])
+    ) {
+        Lib.warn('Invalid geo settings');
+
+        // TODO fallback to default ???
```
alexcjohnson: Yes, I would think so - this is the equivalent of setting invalid range on a cartesian plot, right? In that case we autorange and push the autorange back to the input `layout`.

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    // px coordinates of view mid-point,
+    // useful to update `geo.center` after interactions
+    var midPt = this.midPt = [
+        b[0][0] + (b[1][0] - b[0][0]) / 2,
```
alexcjohnson: `(b[0][0] + b[1][0]) / 2`?

src/plots/geo/constants.js
etpinard: Yeah, pasting what I wrote above under _open items_:

> Base layer ordering changes whenever a choropleth trace is present on the subplot (see this [block](https://github.com/plotly/plotly.js/commit/4c1ddb80dd39ae6d1ba5d51d21b95b36919a8bc5#diff-b4732bdd35eeb4830531de6212e422e2R239)). This logic is meant to preserve backward compatibility (i.e. to mimic what that [block](https://github.com/plotly/plotly.js/blob/60b445d2916547f1d1a548d64209989688d7a617/src/traces/choropleth/plot.js#L65-L72) does in the current `master`). But really, we should make the base layer ordering consistent for all trace type combinations in v2. But in the meantime, we should make the base layer ordering customizable (similar to `axis.layer`).

src/plots/geo/geo2.js
etpinard: Good catch :microscope: . Thanks!

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+
+    // adjust projection to user setting
+    projection
+        .scale(projLayout.scale * s)
+        .translate([t[0] + (midPt[0] - t[0]), t[1] + (midPt[1] - t[1])])
+        .clipExtent(b);
+
+    // the 'albers usa' projection does not expose a 'center' method
+    // so here's this hack to make it respond to 'geoLayout.center'
+    if(geoLayout._isAlbersUsa) {
```
alexcjohnson: would there be a way to fold this into `getProjection`, perhaps where you fill in [`dummyFn`](https://github.com/plotly/plotly.js/pull/2030/commits/4c1ddb80dd39ae6d1ba5d51d21b95b36919a8bc5#diff-b4732bdd35eeb4830531de6212e422e2R501)? I guess that would have to also override `.translate` so perhaps it's more trouble than it's worth... still `_isAlbersUsa` is a bit annoying.

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+    var basePaths = _this.basePaths;
+
+    function isAxisLayer(d) {
+        return (d === 'lonaxis' || d === 'lataxis');
+    }
+
+    function isTopoLayer(d) {
+        return (
+            constants.fillLayers.indexOf(d) !== -1 ||
+            constants.lineLayers.indexOf(d) !== -1
```
alexcjohnson: 🐎  is it worth turning this (and `isAxisLayer`) into a hash lookup `{ocean: 1, land: 1, lakes: 1}` instead of `indexOf`? in fact it looks like `indexOf` is the only way `fillLayers` and `lineLayers` are ever used

src/plots/geo/constants.js
alexcjohnson: Ah OK, missed that note, thanks. 👍

src/plots/geo/geo2.js
etpinard: Very good idea :racehorse:

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+        var path = basePaths[d];
+        var adj = constants.layerNameToAdjective[d];
+
+        if(d === 'frame') {
+            path.datum(constants.sphereSVG);
+        } else if(isTopoLayer(d)) {
+            path.datum(topojsonFeature(topojson, topojson.objects[d]));
+        } else if(isAxisLayer(d)) {
+            path.datum(makeGraticule(d, geoLayout))
+                .attr('fill', 'none')
```
alexcjohnson: 🐎  these constant `.attr` calls can go in `enter()` - but also for consistency switch to `.style` for fill and stroke?

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+    if(dragMode === 'pan') {
+        _this.bgRect.node().onmousedown = null;
+        framework.call(createGeoZoom(_this, geoLayout));
+        framework.on('dblclick.zoom', zoomReset);
+    }
+
+    framework.on('mousemove', function() {
+        var lonlat = _this.projection.invert(d3.mouse(this));
+
+        if(!lonlat || isNaN(lonlat[0]) || isNaN(lonlat[1])) return;
```
alexcjohnson: In `geo_orthographic` if I mouse off the globe while hovered on a point, this block gets hit, and I don't get an unhover event. I thought I'd be able to fix it by adding `Fx.loneUnhover` before the `return` here but that didn't seem to work. Can you take a look @etpinard ?
screen shot 2017-09-22 at 12 06 03 pm

alexcjohnson: And very often the hover tip still shows after I mouse out this way. Every once in a while I actually get a hover tip for a totally different element that I haven't ever hovered on...

etpinard: Can you check if that's also true in `latest`?

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+
+proto.makeFramework = function() {
+    var _this = this;
+    var fullLayout = _this.graphDiv._fullLayout;
+    var clipId = 'clip' + fullLayout._uid + _this.id;
+
+    var defGroup = fullLayout._defs.selectAll('g.clips')
+        .data([0]);
+    defGroup.enter().append('g')
+        .classed('clips', true);
```
alexcjohnson: We've got a few of these scattered about... should we just make a `g.clips` from the beginning when we make `fullLayout._defs`?

src/plots/geo/geo2.js
alexcjohnson: And while we're on the topic, I notice that hover tips are still present while dragging the projection - ideally they should disappear.

alexcjohnson: Behavior on master is similar, yes, though the second issue (hover tip for a totally different element) seems to be new:
kapture 2017-09-22 at 12 29 42

src/traces/choropleth/plot.js
```diff
@@ -122,9 +106,11 @@ function style(geo) {
             )
         );
 
-        s.selectAll('path.choroplethlocation').each(function(pt) {
+        s.selectAll('path.choroplethlocation').each(function(_, i) {
+            var pt = calcTrace[i];
```
alexcjohnson: why is`pt` not just the first arg?

src/traces/choropleth/hover.js
```diff
@@ -12,23 +12,35 @@
...
-    if(!pt) return;
+    for(i = 0; i < cd.length; i++) {
+        pt = cd[i];
+        isInside = false;
 
-    var centroid = geo.projection(pt.properties.ct);
+        if(pt._polygons) {
+            for(j = 0; j < pt._polygons.length; j++) {
+                if(pt._polygons[j].contains([xval, yval])) {
+                    isInside = !isInside;
```
alexcjohnson: Just to have a record of it - @etpinard and I discussed this in chat: in principle it would be faster to keep the nesting structure of the geojson, that tells us the difference between exterior and internal polygons. Then we could:
- Look for the first exterior loop we are inside
- Don’t check any more exterior loops after that
- Only check the interior loops for that particular exterior loop

But for now there can only be at most 200 locations, for ~ 200 countries, and polygon.contains should short-circuit most of these based on their bounding boxes, so this is not a big deal. If we ever add US counties or something we can revisit. But then @etpinard points out that this becomes complicated in a few obnoxious cases of [higher-order nesting](https://en.wikipedia.org/wiki/India%E2%80%93Bangladesh_enclaves) which this simple but slower algorithm (are you inside an odd number of polygons?) gets right.

src/plot_api/subroutines.js
```diff
@@ -513,6 +513,13 @@ exports.doModeBar = function(gd) {
         subplotObj.updateFx(fullLayout);
     }
 
+    subplotIds = Plots.getSubplotIds(fullLayout, 'geo');
+    for(i = 0; i < subplotIds.length; i++) {
+        subplotLayout = fullLayout[subplotIds[i]];
+        subplotObj = subplotLayout._subplot;
+        subplotObj.updateFx(fullLayout, subplotLayout);
+    }
```
alexcjohnson: This section is getting a bit overgrown... can we pull these out into the plot modules like:
```js
for(i=0; i < fullLayout._basePlotModules.length; i++) {
    var updateFx = fullLayout._basePlotModules[i].updateFx;
    if(updateFx) updateFx(fullLayout);
}
```

etpinard: Good call. I'll need to make the `updateFx` API consistent across subplots, but yeah, it'll be very much worthwhile.

test/jasmine/tests/select_test.js
```diff
@@ -703,4 +703,45 @@ describe('Test select box and lasso per trace:', function() {
         .then(done);
     }, LONG_TIMEOUT_INTERVAL);
 
+    it('should work on choropleth traces', function(done) {
+        gd = createGraphDiv();
```
alexcjohnson: `beforeEach`?

src/traces/scattergeo/plot.js
```diff
@@ -62,6 +62,7 @@ module.exports = function plot(geo, calcData) {
                 .data([{geojson: lineData, trace: trace}])
               .enter().append('path')
                 .classed('js-line', true)
+                .style('stroke-miterlimit', 2);
```
alexcjohnson: ❤️ 
I also looked around at the paths generated by geojson/d3, which do not have any explicit `stroke-miterlimit` set (so they get the default of 4) and they generally look fine, but in certain cases might also look nicer at 2.
Here's the northeast US (global scale) at default 4:
screen shot 2017-09-22 at 4 30 22 pm

and at 2:
screen shot 2017-09-22 at 4 30 40 pm

alexcjohnson:
If I do `Plotly.relayout(gd,{coastlinewidth:10})` (still using `geo_orthographic`), occasionally (I really can't figure out when ☹️ ) the viewport shrinks:
before:
screen shot 2017-09-22 at 4 27 30 pm

after:
screen shot 2017-09-22 at 4 28 13 pm

And then when I do a restyle the viewport expands again.
Doesn't seem to happen on master... but it's a different system there because any relayout there resets the zoom.

test/image/mocks/geo_across-antimeridian.json
```diff
@@ -0,0 +1,89 @@
+{
+    "data": [
+        {
+            "type": "scattergeo",
+            "lon": [170, 180, -180, -170],
+            "lat": [-50, -45, -10, -20.5],
+            "marker": { "size": 20 }
```
alexcjohnson: for completeness, want to include lines here? Hard to imagine that they would do something funny like try to wrap the other way around the globe, but ya never know if it's not tested!

etpinard:
> occasionally (I really can't figure out when  ) the viewport shrinks:

Hmm. That shouldn't happen. I thought I fixed it. Oh well. Thanks for pointing that out.

alexcjohnson:
re: https://github.com/plotly/plotly.js/pull/2030/commits/eaf44a8fb898e55a3f8679469425ab1ade7d6f7b - [geo_scattergeo-out-of-usa.png](https://github.com/plotly/plotly.js/blob/eaf44a8fb898e55a3f8679469425ab1ade7d6f7b/test/image/baselines/geo_scattergeo-out-of-usa.png) looks the same as https://plot.ly/~jackp/17904/ and the errors @jackparmer mentioned were appearing on hover, so just running the image test won't catch them, will it?

etpinard:
> re: eaf44a8 - geo_scattergeo-out-of-usa.png looks the same as https://plot.ly/~jackp/17904/ and the errors @jackparmer mentioned were appearing on hover, so just running the image test won't catch them, will it?

Oops. My mistake. I'll :lock: this down using a jasmine test instead.

alexcjohnson:
> `marker.opacity` wasn't applied properly on `scattergeo` bubbles

Was it just the default being applied improperly, or was opacity not supported before?

I'm a bit torn about whether this is the right default for scattergeo or not. On one hand, transparency lets you see borders behind the bubble, so it's clearer where it is. On the other hand, all those extra lines (and colors, if you color the water or something) make it harder to quickly gauge the bubble size and/or color. It's also a behavior change people might not want in their existing plots. Thoughts @jackparmer @chriddyp @cldougl ?

etpinard:
> Was it just the default being applied improperly, or was opacity not supported before?

Good question. I'll investigate.

test/image/mocks/geo_across-antimeridian.json
etpinard: Works fine with `mode: 'lines'`

image

You scared me there, because I know `mapbox-gl` does the wrong thing in this case.

src/plots/geo/geo2.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/ab059900a1b877c4528a029553e22979d0e16e25

src/plots/geo/geo2.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/eeee9118294af10660a7bcedfbbcea12ec275579

src/plots/geo/geo2.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/eeee9118294af10660a7bcedfbbcea12ec275579

src/traces/choropleth/plot.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/71c535541bb4816310f782ce9ee836247cf54c78

test/jasmine/tests/select_test.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/a3df09ebea3552210125b1de41b8afb103db125c

test/image/mocks/geo_across-antimeridian.json
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/05e819429e76d6a49e770441d54086b05aebb8ed

src/plot_api/subroutines.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/2488f4b71ff79ef00a476f845f7e29fac8b7cc0b

test/jasmine/tests/geo_test.js
```diff
@@ -1150,9 +1150,22 @@ describe('Test geo interactions', function() {
                     done();
                 });
             });
-
         });
     });
+
+    it('should not throw during hover when out-of-range pts are present in *albers usa* map', function(done) {
```
etpinard: which :white_check_mark: https://github.com/plotly/plotly.js/pull/2030#issuecomment-331560323

Note that I kept the `scattergeo-out-of-usa` because the image test does :lock: down the bug fix.

See the rogue point(s) in the upper-left corner:

image


this thing :arrow_double_down: 

image

But yeah, adding a jasmine test documents the behavior better, so here it is.

src/plots/geo/geo2.js
etpinard: Done in https://github.com/plotly/plotly.js/commit/46d29d5af1daedd639742228a746fce1a804b3ac using a `relayout` call for maximum robustness.

src/plots/geo/geo2.js
etpinard: > And very often the hover tip still shows after I mouse out this way.

This one was easy, using `dragElement.unhover` instead of `Fx.loneUnhover` did the trick in 9bb7d91 Unfortunately, I can get the test I wrote working on CI.

>  (hover tip for a totally different element) seems to be new

This turned out to be caused by two separate bugs:

-https://github.com/plotly/plotly.js/pull/2030/commits/1c26ec4d12f542f289634ecf6e8bae049d4e73f5 made sure that `lon`/`lat` coordinate are converted correctly to px in scattergeo/hover.js
- https://github.com/plotly/plotly.js/pull/2030/commits/651399e430dfdff6f91b3c06cebeba07898d82b7 is a little more subtle - we must make sure that location polygons that cross the antimeridian have their negative longitude offset by 360 degrees so that `polygon.contains` doesn't get confuse. I found that only Russia and Fiji had such polygons.

So now, hovering over the North Pole on `geo_orthographic` gives:

peek 2017-09-27 10-34

src/traces/scattergeo/plot.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/303de22a6d8b8e56f82c88783469a155f4629c7b

src/plots/geo/geo2.js
etpinard: There's no easy way unfortunately, if we add a special mocked `.center` method for `albersUsa`, we would also need to have special logic in `.fitExtent` which would otherwise clear the `translate` settings applied by `.center`.

src/plots/geo/geo2.js
alexcjohnson: That was quite a deep dive - nicely done, and thanks for tacking on the first bug even though it wasn't new.

src/traces/scattergeo/plot.js
alexcjohnson: Sweet, thanks for the special mock to call this out - rather than relying on subpixel tweaks elsewhere to test it.

src/plots/geo/geo2.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2030/commits/8f9b75aefcb8ac965405cdf1b8a92fd5b40aa62a

src/components/annotations/draw.js
```diff
@@ -194,7 +194,7 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
 
     var isSizeConstrained = options.width || options.height;
 
-    var annTextClip = fullLayout._defs.select('.clips')
+    var annTextClip = fullLayout._topclips
```
etpinard: @alexcjohnson IMPORTANT: `Annotation` used to put its clips in `fulLayout._defs`, even though `Annotation` draws in `toppaper`. Clip ids are global to the page, so things still work fine (at in Chrome and FF). But, who knows maybe some other browsers don't apply clip ids outside the `<svg>` where the `<defs>` is located.

alexcjohnson: Nice find! Definitely not the intent, thanks for cleaning that up!

alexcjohnson:
@etpinard and I discussed bubble opacity a bit more - I was on the fence but he won me over with keeping symmetry with regular `scatter`. So the behavior introduced here is great.

src/plots/geo/geo2.js
etpinard: With https://github.com/plotly/plotly.js/pull/2030/commits/190c7338b3f2a6271d54bf90129129374e5596f8 choropleth hover now works over Antarctica. Is wasn't easy, but I think I found a solid way to do this

peek 2017-09-27 18-32

src/plots/geo/geo2.js
```diff
@@ -0,0 +1,612 @@
...
+        this.basePaths[k].attr('d', pathFn);
+    }
+
+    for(k in this.dataPaths) {
+        this.dataPaths[k].attr('d', function(d) { return pathFn(d.geojson); });
+    }
+
+    for(k in this.dataPoints) {
+        this.dataPoints[k]
+            .attr('display', hideShowPoints)
```
etpinard: > marker.opacity wasn't applied properly on scattergeo bubbles
> Was it just the default being applied improperly, or was opacity not supported before?

This here is why suddenly `marker.opacity` is applied correctly in `scattergeo` traces. [Previously](https://github.com/plotly/plotly.js/blob/549ee3866d86b64bd34a574573bd5596a57c1ed0/src/plots/geo/geo.js#L447) we used `opacity` 0 :left_right_arrow:  1 to hide points that over the map's edges which erroneously overrode whatever `Drawing.pointStyle` set.  See https://github.com/plotly/plotly.js/pull/1861#discussion_r126199803 for why we now use `display` hide/show points.

> @etpinard and I discussed bubble opacity a bit more - I was on the fence but he won me over with keeping symmetry with regular scatter. So the behavior introduced here is great.

:ok_hand:

src/plots/geo/geo2.js
alexcjohnson: Wow OK - Antarctica is especially annoying because, at least the way you did this (and I don't see a better way, this looks great) we *can't* envision fixing this at the topojson level, or we'd draw a spurious ray to the pole.

etpinard: > we can't envision fixing this at the topojson level, or we'd draw a spurious ray to the pole.

The special patches added in `choropleth/plot.js` can't _just_ be moved to the topojson creations step. Look what happens when longitudes [-180, 0] are mapped to [180, 360]:

peek 2017-09-28 09-40

which I think is due to d3 [optimising](https://bl.ocks.org/mbostock/5735770) around longitude -180/180 (not 360). So, to avoid spurious drawings and coordinate duplication, we'll _always_ have to patch the geojson after the topojson is fetched in plotly.js.

---

Thinking about this more deeply, perhaps we should instead 1) map longitudes of antimeridian crossing polygons using `lon => Math.sin(lon)` (or `cos` :slightly_smiling_face: ) in `choropleth/plot.js` 2) add a flag in their corresponding `calcdata` items 3) apply `Math.sin` to the lon/lat mouse position in `choropleth/hover.js` when computing the distance from these flagged calcdata items. This might be more robust than the ad-hoc offset or stitch techniques of 190c733. Thoughts?

etpinard:
> If I do Plotly.relayout(gd,{coastlinewidth:10}) (still using geo_orthographic), occasionally (I really can't figure out when  ) the viewport shrinks:

(Referencing commits in: https://github.com/plotly/plotly.js/compare/geo-refactor+lasso...try-bigger-clippad)

FYI I was able to reproduce the bug using a simple `relayout` call - see https://github.com/plotly/plotly.js/commit/3ad3d1e412342c13071ab36a283700fbb8d7613f

The problem can be fixed by increasing the `clipPad` geo constant from `1e-3` to `1e-2`. Now, why? Without a `clipPad` some clipped projections (e.g. `orthographic`), compute different bounds for different `geo.projection.rotation` values. But this shouldn't happen. Bounds should be a function of the lon/lat axis `range` only. This is possibly due to rounding errors in d3, but my investigation stopped there.

@alexcjohnson you can comment this [line](https://github.com/plotly/plotly.js/compare/geo-refactor+lasso...try-bigger-clippad#diff-1f829b9e62be68acb1eabdb81bf5d1cbR275) to draw the lon/lat range-box polygon on the map if you feel like investigating further. 

Bumping `clipPath` to 1e-2 does affect the baselines - see https://github.com/plotly/plotly.js/commit/3cd0ca51c9cf085cb919191535a0ad881793387f, but not noticeably. The jasmine tests are passing fine. 

There might be other situation where `clipPath: 1e-2` will fail too. Maybe setting it to `0.1` would be safer. Oh well, I couldn't get geo to break with `clipPath: 1e-2`.

alexcjohnson:
> > If I do Plotly.relayout(gd,{coastlinewidth:10}) (still using geo_orthographic), occasionally (I really can't figure out when  ) the viewport shrinks:

> There might be other situations where `clipPath: 1e-2` will fail too. Maybe setting it to `0.1` would be safer. Oh well, I couldn't get geo to break with `clipPath: 1e-2`. 

I'm uncomfortable with this solution. For example, what if you zoom in sufficiently that 0.1 (or even 0.01) is a noticeable fraction of the range? I think we need to step back and look into where this is coming from a little closer.

The bug exists on master (in fact it's worse, you just get the top or bottom half of the viewport rather than a centered half, see below) so lets ignore it for this PR and make a new bug report for it.
screen shot 2017-09-28 at 12 32 09 pm

src/plots/geo/geo2.js
alexcjohnson: For the record (a conversation @etpinard and I had elsewhere) I worry that `sin` or `cos` mapping would introduce other artifacts, as it maps two `lon` values onto the same output.

I think we're always going to need special handling of some sort or other, because the globe simply doesn't map onto flat space as assumed by `polygon.contains`. That could either be as you've introduced here, or making a special `polygon.contains` just for lat/lon, with special handling for crossing the antimeridian (with additional edge cases for encircling either pole).

Given that at present we only have three entities that can trigger this, I think the solution you have here is perfect. If we extend beyond this in the future (one tricky one that occurs to me is if we allow filled areas in `scattergeo` with `hoveron: 'fills'`) we can generalize at that point.

alexcjohnson:
💃 🕺 💃 🕺 

Fantastic work - so much more solid and well-reasoned!

Issue #2024, opened 2017-09-20
Aggregation transform throws when aggregations not defined


bpostlethwaite:
@alexcjohnson 

Aggregations are scooped out of incoming transform here:
https://github.com/plotly/plotly.js/blob/master/src/transforms/aggregate.js#L158

and we check the length here:
https://github.com/plotly/plotly.js/blob/master/src/transforms/aggregate.js#L159

But later we check if it is defined at all:
https://github.com/plotly/plotly.js/blob/master/src/transforms/aggregate.js#L166

Partial aggregation transforms are easily composed in a UI and sending a partially complete transform without aggregations defined results in:
```
Cannot read property 'length' of undefined TypeError: Cannot read property 'length' of undefined
```
on line https://github.com/plotly/plotly.js/blob/master/src/transforms/aggregate.js#L159

PR #2031, opened 2017-09-22
Add Aggregation user input _index

bpostlethwaite:
fixes https://github.com/plotly/plotly.js/issues/2024
and adds _index to fullData `aggregation.aggregations` to make back referencing into user data possible.

cc @alexcjohnson

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -1039,9 +1039,21 @@ plots.supplyTransformDefaults = function(traceIn, traceOut, layout) {
             _module = transformsRegistry[type],
             transformOut;
 
+        /*
+         * Supply defaults may run twice. First pass runs all supply defaults steps
+         * and adds the _module to any output transforms.
+         * If transforms exist another pass is run so that any generated traces also
+         * go through supply defaults. This has the effect of rerunning
+         * supplyTransforms. If the transform does not have a `transform` function
+         * it could not have generated any new traces and the second stage is
+         * unnecessary. We detect this case with the following variables.
+         */
+        var isSecondStage = transformIn._module && transformIn._module === _module,
+            doSecondStage = _module && typeof _module.transform === 'function';
+
         if(!_module) Lib.warn('Unrecognized transform type ' + type + '.');
 
-        if(_module && _module.supplyDefaults) {
+        if(_module && _module.supplyDefaults && (!isSecondStage || doSecondStage)) {
             transformOut = _module.supplyDefaults(transformIn, traceOut, layout, traceIn);
             transformOut.type = type;
             transformOut._module = _module;
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -1044,16 +1044,16 @@ plots.supplyTransformDefaults = function(traceIn, traceOut, layout) {
          * and adds the _module to any output transforms.
          * If transforms exist another pass is run so that any generated traces also
          * go through supply defaults. This has the effect of rerunning
-         * supplyTransforms. If the transform does not have a `transform` function
-         * it could not have generated any new traces and the second stage is
-         * unnecessary. We detect this case with the following variables.
+         * supplyTransformDefaults. If the transform does not have a `transform`
+         * function it could not have generated any new traces and the second stage
+         * is unnecessary. We detect this case with the following variables.
          */
-        var isSecondStage = transformIn._module && transformIn._module === _module,
-            doSecondStage = _module && typeof _module.transform === 'function';
+        var isFirstStage = !(transformIn._module && transformIn._module === _module),
+            doLaterStages = _module && typeof _module.transform === 'function';
 
         if(!_module) Lib.warn('Unrecognized transform type ' + type + '.');
 
-        if(_module && _module.supplyDefaults && (!isSecondStage || doSecondStage)) {
+        if(_module && _module.supplyDefaults && (isFirstStage || doLaterStages)) {
             transformOut = _module.supplyDefaults(transformIn, traceOut, layout, traceIn);
             transformOut.type = type;
             transformOut._module = _module;
--- a/test/jasmine/tests/transform_aggregate_test.js
+++ b/test/jasmine/tests/transform_aggregate_test.js
@@ -245,4 +245,40 @@ describe('aggregate', function() {
         expect(traceOut.marker.line.width).toBeCloseToArray([0.5, 0], 5);
         expect(traceOut.marker.color).toBeCloseToArray([Math.sqrt(1 / 3), 0], 5);
     });
+
+    it('links fullData aggregations to userData via _index', function() {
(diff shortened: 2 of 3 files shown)


etpinard:
Very interesting fix. Transforms that do not expand traces shouldn't have to go through `supplyTransformDefaults` twice. At the moment, all transforms but `groupby` don't expand traces. So from a pref standpoint, this PR is :ok_hand: 

cc @rreusser 's who in https://github.com/plotly/plotly.js/pull/1830 had to go-around a few transform design problems.

src/plots/plots.js
```diff
@@ -1039,9 +1039,21 @@ plots.supplyTransformDefaults = function(traceIn, traceOut, layout) {
             _module = transformsRegistry[type],
             transformOut;
 
+        /*
+         * Supply defaults may run twice. First pass runs all supply defaults steps
```
rreusser: Questions:

- I assume this refers to the way `supplyTraceDefaults` runs [on traces and then again on expanded traces](https://github.com/plotly/plotly.js/blob/master/src/plots/plots.js#L799-L812)?
- Which `_module` is added? I think that sometimes refers to the trace module and sometimes to the transform module (and every now and then to the base plot module, though usually it's called `baseModule` or something, I think)

src/plots/plots.js
```diff
@@ -1039,9 +1039,21 @@ plots.supplyTransformDefaults = function(traceIn, traceOut, layout) {
...
+         * and adds the _module to any output transforms.
+         * If transforms exist another pass is run so that any generated traces also
+         * go through supply defaults. This has the effect of rerunning
+         * supplyTransforms. If the transform does not have a `transform` function
```
rreusser: *`supplyTransformDefaults`?

src/plots/plots.js
rreusser: Answer to second: the *transform* module.

src/plots/plots.js
```diff
@@ -1039,9 +1039,21 @@ plots.supplyTransformDefaults = function(traceIn, traceOut, layout) {
...
+         * it could not have generated any new traces and the second stage is
+         * unnecessary. We detect this case with the following variables.
+         */
+        var isSecondStage = transformIn._module && transformIn._module === _module,
```
rreusser: Subtle, but will this detect second stage incorrectly if the first and second stage both have the same `_module`? Or maybe it's the second and third stage that would be indistinguishable?

rreusser: Is a third stage possible if multiple transforms expand things multiple times? (e.g. groupby + groupby) I think this makes sense to me, but just making sure transform module identity equates to `isSecondStage`. (Or is it `isNotFirstStage`?) My understanding is just a little loose here.

src/plots/plots.js
bpostlethwaite: Yep that is what it is referring to

src/plots/plots.js
bpostlethwaite: right, more than 2 stages... need to think about that. Suggestions?

bpostlethwaite: Right I think we are ok. I should call this variable `isNotFirstStage` though. Thanks

rreusser: Things get fundamentally a bit shaky after one level. Like `groupby` + `groupby`. It does mostly what you'd expect, but it turns out *styling* is the big challenge. Think SQL. You process the data and process the data until you have a nice query that spits out data. Then you insert it into a plot and you're happy. Plotly's transforms incrementally apply styles *as* as sequence of transforms is applied so that you'd have to track styles as well as the pathway a data point took through the query to really sort things out.

rreusser: Sorry, specifics: `groupby` adds a `_group` property to traces so that you can figure out which group split generated a trace. But it should really be tracking that as a transform index and a group value. This comes into play with transform vs. calcTransform since you start having to worry about *when* a trace was split/grouped/styled/etc.

rreusser: Apologies if that's still a bit loose on the specifics. /cc @alexcjohnson since we've talked this through before in a fair bit of detail. I think our conclusion at the time was that the basic use cases are handled just fine and account for 97+% of realistic use-cases.

bpostlethwaite:
@rreusser some changes here: https://github.com/plotly/plotly.js/pull/2031/commits/ec0a65f0f6dee798e912f61ea4180e5d0969958c

Does that seem more clear? It seems to be more honest about the possibility of more than 2 stages.

test/jasmine/tests/transform_aggregate_test.js
```diff
@@ -225,4 +245,40 @@ describe('aggregate', function() {
...
+
+        var traceOut = gd._fullData[0];
+        var fullAggregation = traceOut.transforms[0];
+        var fullAggregations = fullAggregation.aggregations;
+        var enabledAggregations = fullAggregations.filter(function(agg) {
+            return agg.enabled;
+        });
+
+        expect(enabledAggregations[0].target).toEqual('x');
+        expect(enabledAggregations[0]._index).toEqual(0);
```
rreusser: Is there any way to test this that doesn't rely upon an underscored property? Like could `supplyTransformDefaults` delete the first aggregation if it's overridden by another? Would that be adequate to sort things out in the workspace or does the workspace need to know indices?

rreusser:
Above corner-case comments notwithstanding, this all looks reasonable to me.

test/jasmine/tests/transform_aggregate_test.js
bpostlethwaite: The only reason for indicies is for Workspace. supplyTransformDefaults doesn't care about them.

bpostlethwaite: I need a way in Workspace to take a fullData transform Aggregration and write back to the correct userData index. I need some way to correlate userData indicies with the fullData equivalents.

bpostlethwaite: maybe it shouldn't be underscored? Something like {inputIndex}. the only problem with that is there are precedents for things being added to fullData for consumption bu users that have an underscore. `_input` for example.

rreusser: So the workspace accesses indices directly? (That's okay, just making sure. plotly.js's 'private' properties are fairly publicly accessed, so it's nothing new. Just means they need to be modified with great care.)

bpostlethwaite: yep, and the workspace is the _only_ thing that needs them. Plotly.js doesn't care as it rebuilds new arrays each time.

etpinard:
Ok. Looks like @rreusser approves this. 

:dancer: for me @bpostlethwaite 

This will be part of `1.31.0` set to be released no-latter than October 5.

PR #2032, opened 2017-09-23
Eliminate circular dependencies

### LINKED ISSUE #236: De-circularize plotly.js require statements

rreusser:
resolves #236.

This PR eliminates all circular dependencies (🎉), mainly by adding `gd._plotAPI.*` (as suggested in #236) which binds (curries?) the `gd` to the arguments automatically. That is, you can call `gd._plotAPI.redraw()` or `gd._plotAPI.restyle([....])`. This eliminates the need for `var Plotly = require('../../plotly')`, which accounted for the vast majority of circular dependencies.

I didn't really find a need for nontrivial changes (has the component registry been implemented since #236?) since all other circular dependencies could be resolved by moving one or two functions to their own files. Though it's a bit ugly to require `get_subplot_id.js` separately. I'm open to different thoughts on this.

I expect a couple failing tests to remain since `spyOn` got somewhat more tricky to enforce and required binding the API manually in order to make sure it was the correct, spied-upon API.

Edit: *almost*. One circular dependency remains. Remaining issue:

- [ ] `toImage` requires plot api --> circular

### DIFF
--- a/src/plots/font_weight.js
+++ b/src/plots/font_weight.js
@@ -1,0 +1,13 @@
+/**
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+// TODO make this a plot attribute?
+module.exports.fontWeight = 'normal';
--- a/src/plots/get_subplot_ids.js
+++ b/src/plots/get_subplot_ids.js
@@ -1,0 +1,54 @@
+/**
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+
+var Registry = require('../registry');
+
+/**
+ * Get the ids of the current subplots.
+ *
+ * @param {object} layout plotly full layout object.
+ * @param {string} type subplot type to look for.
+ *
+ * @return {array} list of ordered subplot ids (strings).
+ *
+ */
+module.exports = function getSubplotIds(layout, type) {
+    var _module = Registry.subplotsRegistry[type];
+
+    if(!_module) return [];
+
+    // layout must be 'fullLayout' here
+    if(type === 'cartesian' && (!layout._has || !layout._has('cartesian'))) return [];
+    if(type === 'gl2d' && (!layout._has || !layout._has('gl2d'))) return [];
+    if(type === 'cartesian' || type === 'gl2d') {
+        return Object.keys(layout._plots || {});
+    }
+
+    var attrRegex = _module.attrRegex,
+        layoutKeys = Object.keys(layout),
+        subplotIds = [];
+
+    for(var i = 0; i < layoutKeys.length; i++) {
+        var layoutKey = layoutKeys[i];
+
+        if(attrRegex.test(layoutKey)) subplotIds.push(layoutKey);
+    }
+
+    // order the ids
+    var idLen = _module.idRoot.length;
+    subplotIds.sort(function(a, b) {
+        var aNum = +(a.substr(idLen) || 1),
+            bNum = +(b.substr(idLen) || 1);
+        return aNum - bNum;
+    });
+
+    return subplotIds;
+};
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -12,12 +12,12 @@
 var d3 = require('d3');
 var isNumeric = require('fast-isnumeric');
 
-var Plotly = require('../plotly');
 var PlotSchema = require('../plot_api/plot_schema');
 var Registry = require('../registry');
 var Lib = require('../lib');
 var Color = require('../components/color');
 var BADNUM = require('../constants/numerical').BADNUM;
+var Axes = require('./cartesian/axes');
 
 var plots = module.exports = {};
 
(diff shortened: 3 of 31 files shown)


src/plots/get_subplot_ids.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Get the ids of the current subplots.
+ *
+ * @param {object} layout plotly full layout object.
+ * @param {string} type subplot type to look for.
+ *
+ * @return {array} list of ordered subplot ids (strings).
+ *
+ */
+module.exports = function getSubplotIds(layout, type) {
```
rreusser: I don't like this. Moving this one file broke a long circular chain in which `Plots` was required by `plots/cartesian/axis_ids.js`, resulting in a circular dependency.

src/plots/font_weight.js
```diff
@@ -0,0 +1,13 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+// TODO make this a plot attribute?
+module.exports.fontWeight = 'normal';
```
rreusser: This seems silly to have in its own file but was another one of those couple things that required a giant circular require of top-level `Plots` by some deeply nested component.

src/plots/plots.js
```diff
@@ -1865,7 +1811,7 @@ plots.extendTrace = function(destTrace, srcTrace) {
  * The result is the original object reference with the new contents merged in.
  */
 plots.extendLayout = function(destLayout, srcLayout) {
-    return plots.extendObjectWithContainers(destLayout, srcLayout, plots.layoutArrayContainers);
+    return plots.extendObjectWithContainers(destLayout, srcLayout, Registry.layoutArrayContainers);
```
rreusser: `plots` vs. `Registry` was aliased but not updated in quite a few places. It should be otherwise equivalent.

test/jasmine/tests/lib_test.js
```diff
@@ -2008,59 +2008,59 @@ describe('Queue', function() {
         })
         .then(function() {
             expect(gd.undoQueue.index).toEqual(1);
-            expect(gd.undoQueue.queue[0].undo.args[0][1]['marker.color']).toEqual([null]);
-            expect(gd.undoQueue.queue[0].redo.args[0][1]['marker.color']).toEqual('red');
+            expect(gd.undoQueue.queue[0].undo.args[0][0]['marker.color']).toEqual([null]);
```
rreusser: With `gd` gone from the first argument, these all shifted over one.

tasks/test_syntax.js
```diff
@@ -195,7 +195,7 @@ function assertCircularDeps() {
         var logs = [];
 
         // see https://github.com/plotly/plotly.js/milestone/9
-        var MAX_ALLOWED_CIRCULAR_DEPS = 17;
+        var MAX_ALLOWED_CIRCULAR_DEPS = 0;
```
etpinard: :tada:

src/plots/font_weight.js
alexcjohnson: This is only used in one place, `titles` - everywhere else we just implicitly inherit `font-weight: normal`.

One *can* set `font-weight: bold` or something on `gd` or a parent, and have that inherited by text inside the plot. We could argue about whether that's a reasonable thing to allow (in particular, exports would NOT inherit this bolding) but it seems silly to ONLY explicitly remove this from the title. I'd 🔪  this file AND the place it's used in `titles/index.js`. Later if we decide this is important we can apply normal weight to our whole `<svg>` elements or something.

etpinard: By the way, that font weight key has to be exposed on `Plots` for [this](https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/styles/open_office_2015.js#L12).

alexcjohnson: > By the way, that font weight key has to be exposed on `Plots` for [this](https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/styles/open_office_2015.js#L12).

Ah, crazy - as it stands now this PR will break that, but now I get why this was in there in the first place. So it's really just to allow us to make titles bold by default. Then lets have this an attribute of the `Titles` object, ie `Titles.fontWeight`, so it can be overridden by saying `var Titles = require('./components/titles'); Titles.fontWeight = 900;`

rreusser:
I'm going to put this on ice for just a moment since I got a bit tripped up trying to get rid of the last circular dependency. It suggests moving toward a register approach is probably the right answer, but it gets a little treacherous to keep things testable (spies, in particular) and to avoid introducing any extra circularity. It's proved very doable, but will require just a bit of finesse to really nail it.

rreusser:
I wonder if this is related to circular dependencies: https://github.com/ropensci/plotly/issues/483

etpinard:
Closing this thing - as this was _put on ice_ :ice_skate: 

Refer to https://github.com/plotly/plotly.js/milestone/9 for more on the topic

PR #2040, opened 2017-09-28
Throttle selectPoints

### LINKED ISSUE #2005: Throttle box/lasso selection
The UI gets laggy during box/lasso selection on large plots. Throttling it like we do for [hover](https://github.com/plotly/plotly.js/blob/master/src/components/fx/hover.js#L69) should help.

alexcjohnson:
closes #2005 - adds throttling to box and lasso selection. I opted to throttle just the `selectPoints` portion, because It's where nearly all the expense of that drag handler lives, so this way keeps the selection *region* feeling snappy even if it takes some time for the selected *points* to catch up.

Here's what it looked like before:
unthrottled-select
and now with throttling:
throttled-select
Using this plot:
```js
var x = new Array(3e5), y = new Array(x.length);
for(var i = 0; i < x.length; i++) { x[i] = Math.random(); y[i] = Math.random(); }
Plotly.newPlot(gd,
    [{x: x, y: y, type: 'scattergl', mode: 'markers', marker: {opacity: 0.1}}],
    {width: 600, height: 600})
```


src/components/fx/hover.js
```diff
@@ -68,26 +68,12 @@ var HOVERTEXTPAD = constants.HOVERTEXTPAD;
...
-    // Queue up the next hover for 100ms from now (if no further events)
-    gd._hoverTimer = setTimeout(function() {
-        _hover(gd, evt, subplot, noHoverEvent);
-        gd._lastHoverTime = Date.now();
-        gd._hoverTimer = undefined;
-    }, constants.HOVERMINTIME);
+    Lib.throttle(
+        function() { _hover(gd, evt, subplot, noHoverEvent); },
+        constants.HOVERMINTIME,
+        gd._fullLayout._uid + constants.HOVERID
```
alexcjohnson: using `fullLayout._uid` to construct the key to what we're throttling means we can 🔪  `gd._lastHoverTime` and `gd._hoverTimer`

src/lib/get_graph_div.js
```diff
@@ -0,0 +1,39 @@
...
+ * or its id string
+ *
+ * @param {HTMLDivElement|string} gd: a graph element or its id
+ *
+ * @returns {HTMLDivElement} the DOM element of the graph
+ */
+// Get the container div: we store all variables for this plot as
+// properties of this div
+// some callers send this in by DOM element, others by id (string)
+module.exports = function(gd) {
```
alexcjohnson: oops, this file was supposed to go in the next commit https://github.com/plotly/plotly.js/pull/2040/commits/aa434dd3b9b785dac6d37b24be2d89d1e52b8203

docs cleaned up in 0861736

src/plots/cartesian/select.js
```diff
@@ -186,9 +186,15 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
         selection = [];
         for(i = 0; i < searchTraces.length; i++) {
             searchInfo = searchTraces[i];
-            [].push.apply(selection, fillSelectionItem(
```
alexcjohnson: not a reported bug AFAIK, but this use of `push.apply` overloads the stack for large plots. Doesn't look to me as though any of the other places we use `push.apply` are susceptible to this issue.

src/plots/cartesian/select.js
```diff
@@ -208,21 +208,23 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
...
-        throttle.clear(throttleID);
-
-        if(!dragged && numclicks === 2) {
-            // clear selection on doubleclick
-            outlines.remove();
-            for(i = 0; i < searchTraces.length; i++) {
-                searchInfo = searchTraces[i];
-                searchInfo.selectPoints(searchInfo, false);
-            }
+        throttle.done(throttleID).then(function() {
```
alexcjohnson: we need to wait for any throttling-delayed move event before we process the mouseup event, so we have the correct data.

test/jasmine/tests/select_test.js
```diff
@@ -545,108 +584,70 @@ describe('select box and lasso', function() {
...
-                expect(selectedCnt).toBe(1, 'plotly_selected call count');
-                expect(deselectCnt).toBe(1, 'plotly_deselect call count');
-            });
+            return doubleClick(250, 200);
+        })
+        .then(deselectPromise)
+        .then(function() {
+            // deselect gives an extra plotly_selected event on the first click
+            // event data is undefined
+            checkEventCounts(1, 2, 1, 'deselect');
```
alexcjohnson: The `gd.once` we were using in this test before hid this behavior - which exists on master as well as in this branch. I imagine we would prefer to not emit these events with undefined data - which happen I guess when we get [here](https://github.com/plotly/plotly.js/blob/80508e8a60698459f01375c1aa90a8a7e52e0096/src/plots/cartesian/select.js#L225) but `dragged === false`. @etpinard shall I remove that extra event and put this test back as it appeared to be previously?

test/jasmine/tests/select_test.js
```diff
@@ -545,108 +584,70 @@ describe('select box and lasso', function() {
...
+            resetEvents(gd);
             drag([[300, 200], [300, 300], [400, 300], [400, 200], [300, 200]]);
+            return selectedPromise;
+        })
+        .then(function() {
             assertPoints([{lon: 20, lat: 20}]);
             assertLassoPoints([
                 [13.28, 25.97], [13.28, 14.33], [25.71, 14.33], [25.71, 25.97], [13.28, 25.97]
             ]);
+            checkEventCounts(4, 1, 0, 'lasso');
```
alexcjohnson: 4 `plotly_selecting` events is real - because there are 5 points in the lasso path.

rreusser:
Interestingly, I noticed today that selecting a box + heatmapgl can be *very* slow. I'm curious whether that's the same pathway.

src/components/dragelement/unhover.js
```diff
@@ -12,14 +12,15 @@
...
+var getGraphDiv = require('../../lib/get_graph_div');
 
 var hoverConstants = require('../fx/constants');
 
 var unhover = module.exports = {};
 
 
 unhover.wrapped = function(gd, evt, subplot) {
-    if(typeof gd === 'string') gd = document.getElementById(gd);
+    gd = getGraphDiv(gd);
```
etpinard: NIce :palm_tree:

src/plots/cartesian/select.js
etpinard: Anyone interested in writing a syntax [test](https://github.com/plotly/plotly.js/blob/549ee3866d86b64bd34a574573bd5596a57c1ed0/tasks/test_syntax.js#L80-L93) to :lock: this down?

src/plots/cartesian/constants.js
```diff
@@ -47,6 +47,12 @@ module.exports = {
     // delay before a redraw (relayout) after smooth panning and zooming
     REDRAWDELAY: 50,
 
+    // throttling limit (ms) for selectPoints calls
+    SELECTDELAY: 100,
+
+    // cache ID suffix for throttle
+    SELECTID: '-select',
```
etpinard: Non-blocking, but we should start thinking about moving selection things out of `plots/cartesian/` and into e.g. `component/fx/` or a new `component/select` module.

src/plots/cartesian/select.js
```diff
@@ -183,27 +181,35 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
...
-        selection = [];
-        for(i = 0; i < searchTraces.length; i++) {
-            searchInfo = searchTraces[i];
-            var thisSelection = fillSelectionItem(
-                searchInfo.selectPoints(searchInfo, poly), searchInfo
-            );
-            if(selection.length) {
-                for(var j = 0; j < thisSelection.length; j++) {
-                    selection.push(thisSelection[j]);
+        throttle.throttle(
```
etpinard: :ok_hand:

src/plots/cartesian/select.js
```diff
@@ -208,21 +208,23 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
...
-
-        if(!dragged && numclicks === 2) {
-            // clear selection on doubleclick
-            outlines.remove();
-            for(i = 0; i < searchTraces.length; i++) {
-                searchInfo = searchTraces[i];
-                searchInfo.selectPoints(searchInfo, false);
-            }
+        throttle.done(throttleID).then(function() {
+            throttle.clear(throttleID);
```
etpinard: Maybe `throttle.done()` should call `throttle.clear()` before returning a promise?

test/jasmine/tests/select_test.js
```diff
@@ -159,6 +165,42 @@ describe('select box and lasso', function() {
         });
     });
 
+    var selectingCnt, selectingData, selectedCnt, selectedData, deselectCnt, doubleClickData;
+    var selectedPromise, deselectPromise;
+
+    function resetEvents(gd) {
```
etpinard: Ha. This will conflict massively with https://github.com/plotly/plotly.js/pull/2030/commits/e53a908d725489c09857c64be00693fa6388d623 

Not sure who should get priority here.

etpinard:
> that selecting a box + heatmapgl can be very slow

I didn't know you can select heatmapgl traces :confused:

src/lib/throttle.js
```diff
@@ -0,0 +1,101 @@
...
+ * last callback received while waiting.
+ * So the first and last events in a train are always executed (eventually)
+ * but some of the events in the middle can be dropped.
+ *
+ * @param {function} callback: the function to throttle
+ * @param {string} id: an identifier to mark events to throttle together
+ * @param {number} minInterval: minimum time, in milliseconds, between
+ *   invocations of `callback`
+ */
+exports.throttle = function throttle(callback, minInterval, id) {
```
etpinard: I think `throttle(id, minInterval, callback)` would a call signature more familiar to most (node.)js devs.

src/lib/throttle.js
```diff
@@ -0,0 +1,101 @@
...
+                delete timerCache[idi];
+            }
+        }
+        cache = timerCache[id] = {ts: 0, timer: null};
+    }
+
+    _clearTimeout(cache);
+
+    function exec() {
+        callback();
```
etpinard: So `callback` can't itself by async - which is fine but deserves a mention in the docstring.

test/jasmine/tests/select_test.js
etpinard: >  I imagine we would prefer to not emit these events with undefined data

Hmm. I'm thinking some users might find it useful to have `plotly_selected` fired even when the selection polygon is smaller than the minimum allowed. In this case, `plotly_selected` essentially mimics `plotly_deselect` - which some users might not know about. 

So, I'd vote for keeping the current behavior and updating the test (good catch by the way :medal_sports: )

But, maybe this is something we might want to change in v2, especially now that click selections are on the table for future development.

src/plots/cartesian/select.js
alexcjohnson: > Maybe `throttle.done()` should call `throttle.clear()` before returning a promise?

I don't think so. It doesn't come into play here, but you could imagine wanting to finish whatever event had been throttled when `done` was called, without cutting off throttling of future events in the train. I guess that's saying I didn't mean `done` to mean you were necessarily done forever, just done with whatever is in the queue right now.

etpinard: Got it. Thanks!

test/jasmine/tests/select_test.js
alexcjohnson: #2030 is epic enough already, you don't need to add a big merge resolution to it. You can merge first.

test/jasmine/tests/select_test.js
```diff
@@ -130,88 +135,6 @@ describe('Test select box and lasso in general:', function() {
...
-                .then(done);
-        });
-
-        it('should be appended to the zoom layer', function(done) {
-            var x0 = 100,
-                y0 = 200,
-                x1 = 150,
-                y1 = 250,
-                x2 = 50,
-                y2 = 50;
```
alexcjohnson: `(x2, y2)` is in the top margin of this plot, so the doubleclick didn't do anything. But since these tests didn't actually check that the handlers ever got called, it didn't cause a failure.

src/plots/cartesian/select.js
alexcjohnson: we *do* use `push.apply` in a few other places - but they get max a few items in the array so it's not a problem - so if we wanted to 🔒  it, we'd have to also remove those other uses.

alexcjohnson:
@etpinard I had to do (as expected) a fair amount of work pruning merge conflicts in [`select_test`](https://github.com/plotly/plotly.js/blob/6da3cfa39e88697f1db57f20bd76debc201706c9/test/jasmine/tests/select_test.js) - would you mind giving it a look to make sure I preserved your intentions?

src/lib/throttle.js
alexcjohnson: -> https://github.com/plotly/plotly.js/pull/2040/commits/b6d64be17950cf6b31189e512be4c481252f61b4

src/lib/throttle.js
alexcjohnson: -> https://github.com/plotly/plotly.js/pull/2040/commits/b6d64be17950cf6b31189e512be4c481252f61b4

test/jasmine/tests/select_test.js
etpinard: Nice investigation work. Thanks!

etpinard:
> - would you mind giving it a look to make sure I preserved your intentions?

`select_test.js` is looking very nice. That `resetEvents` abstraction is :100: 

I'm looking forward to adding tests for bar/histogram traces tomorrow. :beers:

etpinard:
Great. Tests are :white_check_mark: :arrow_right_hook: :dancer:

src/plots/cartesian/select.js
dfcreative: @alexcjohnson funny enough I stumbled upon this stack overflow in selection for regl. I replaced it with `.concat` here https://github.com/plotly/plotly.js/pull/1869/files#diff-8c5d8f688f2bb41a18d35397e5f58864R211, since it is being called only as many times as there is traces per subplot, which is small. (related [tweet](https://twitter.com/DimaYv/status/923640516104933376))

PR #2041, opened 2017-09-28
Normalize zoom speed and wheel behavior across trace types

rreusser:
This PR attempts to normalize mouse wheel behavior across cartesian plots as well as gl2d and gl3d trace types. In particular, it:

1. multiplies the default zoom speed of gl plots by ten in order to get them more inline with non-gl plots
2. <strike>makes zoom speed configurable through `layout.zoomspeed` (should this be a config parameter?)</strike> *Edit: no longer configurable; just faster*
3. removes logic that prevents mousewheel zooming while gl plots are in pan mode (corresponding to analogous svg behavior, which does not block mousewheel zoom in pan mode)

### DIFF
--- a/src/plots/gl2d/camera.js
+++ b/src/plots/gl2d/camera.js
@@ -256,13 +256,15 @@ function createCamera(scene) {
     }
 
     result.wheelListener = mouseWheel(element, function(dx, dy) {
+        if (!scene.scrollZoom) return false;
+
         var dataBox = scene.calcDataBox(),
             viewBox = plot.viewBox;
 
         var lastX = result.lastPos[0],
             lastY = result.lastPos[1];
 
-        var scale = Math.exp(10.0 * dy / (viewBox[3] - viewBox[1]));
+        var scale = Math.exp(5.0 * dy / (viewBox[3] - viewBox[1]));
 
         var cx = lastX /
                 (viewBox[2] - viewBox[0]) * (dataBox[2] - dataBox[0]) +
--- a/src/plots/gl2d/camera.js
+++ b/src/plots/gl2d/camera.js
@@ -256,7 +256,7 @@ function createCamera(scene) {
     }
 
     result.wheelListener = mouseWheel(element, function(dx, dy) {
-        if (!scene.scrollZoom) return false;
+        if(!scene.scrollZoom) return false;
 
         var dataBox = scene.calcDataBox(),
             viewBox = plot.viewBox;
--- a/src/plots/gl2d/camera.js
+++ b/src/plots/gl2d/camera.js
@@ -262,7 +262,7 @@ function createCamera(scene) {
         var lastX = result.lastPos[0],
             lastY = result.lastPos[1];
 
-        var scale = Math.exp(scene.fullLayout.zoomspeed * 10.0 * dy / (viewBox[3] - viewBox[1]));
+        var scale = Math.exp(10.0 * dy / (viewBox[3] - viewBox[1]));
 
         var cx = lastX /
                 (viewBox[2] - viewBox[0]) * (dataBox[2] - dataBox[0]) +
--- a/src/plots/gl2d/camera.js
+++ b/src/plots/gl2d/camera.js
@@ -262,34 +262,27 @@ function createCamera(scene) {
         var lastX = result.lastPos[0],
             lastY = result.lastPos[1];
 
-        switch(scene.fullLayout.dragmode) {
-            case 'zoom':
-                break;
+        var scale = Math.exp(scene.fullLayout.zoomspeed * 10.0 * dy / (viewBox[3] - viewBox[1]));
 
-            case 'pan':
-                var scale = Math.exp(0.1 * dy / (viewBox[3] - viewBox[1]));
+        var cx = lastX /
+                (viewBox[2] - viewBox[0]) * (dataBox[2] - dataBox[0]) +
+            dataBox[0];
+        var cy = lastY /
+                (viewBox[3] - viewBox[1]) * (dataBox[3] - dataBox[1]) +
+            dataBox[1];
 
-                var cx = lastX /
-                        (viewBox[2] - viewBox[0]) * (dataBox[2] - dataBox[0]) +
-                    dataBox[0];
-                var cy = lastY /
-                        (viewBox[3] - viewBox[1]) * (dataBox[3] - dataBox[1]) +
-                    dataBox[1];
+        dataBox[0] = (dataBox[0] - cx) * scale + cx;
+        dataBox[2] = (dataBox[2] - cx) * scale + cx;
+        dataBox[1] = (dataBox[1] - cy) * scale + cy;
+        dataBox[3] = (dataBox[3] - cy) * scale + cy;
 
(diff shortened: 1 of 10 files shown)


src/plots/plots.js
```diff
@@ -1332,6 +1332,7 @@ plots.purge = function(gd) {
     delete gd._transitionData;
     delete gd._transitioning;
     delete gd._initialAutoSize;
+    delete gd._transitioningWithDuration;
```
rreusser: This was just one minor cleanup tweak I noticed along the way.

src/plots/gl2d/camera.js
```diff
@@ -262,34 +262,27 @@ function createCamera(scene) {
         var lastX = result.lastPos[0],
             lastY = result.lastPos[1];
 
-        switch(scene.fullLayout.dragmode) {
```
etpinard: Why is this gone? `fullLayout.dragmode: false` should still be a thing.

src/plots/gl2d/camera.js
```diff
@@ -262,34 +262,27 @@ function createCamera(scene) {
...
-                scene.setRanges(dataBox);
-
-                result.lastInputTime = Date.now();
-                unSetAutoRange();
-                scene.cameraChanged();
-                scene.handleAnnotations();
-                scene.relayoutCallback();
-                break;
-        }
+        var scale = Math.exp(scene.fullLayout.zoomspeed * 10.0 * dy / (viewBox[3] - viewBox[1]));
```
etpinard: Nice. Can we put that `10.0` in https://github.com/plotly/plotly.js/blob/master/src/constants/interactions.js ?

src/plots/plots.js
etpinard: Good catch :ok_hand:

alexcjohnson:
Nice, definitely better parity this way. Playing with it now though, I wonder if something a little slower (maybe half this speed?) might be best for both versions. I still think you're right not to have it configurable though.

Also, we decided long ago that 3D should have scroll zoom on always, but for SVG you need to have `config.scrollZoom: true` to enable it. Seems like gl2D should respect this as well, right?

src/plots/gl2d/camera.js
rreusser: From what I can tell, `dragmode: false` isn't an option, is it? If I set that, it defaults to `zoom`. This is removed because the dragmode should not be affecting mousewheel behavior at all.

rreusser:
Mousewheel speed is a contentious issue. I'd support half speed for all. They at least felt same-ish
 to me with these numbers. 🐭 

Can you clarify the last point? For me when scrollZoom: false,
- scatter: no mousewheel zoom
- scattergl: mousewheel zoom 
- scatter3d: mousewheel zoom

alexcjohnson:
> Can you clarify the last point? For me when scrollZoom: false,
> - scatter: no mousewheel zoom
> - scattergl: mousewheel zoom 
> - scatter3d: mousewheel zoom

I think scatter and scattergl should match, but scatter3d should have wheel zoom enabled always.

src/plots/gl2d/camera.js
rreusser: Ahhh possible bug…… 🕷 

When setting `dragmode: false` with gl3d, dragging is disabled but `gd._fullLayout.dragmode = 'zoom'`. That suggests it's sanitizing it but using `gd.layout.dragmode` instead of `gd._fullLayout.dragmode`.

alexcjohnson: Looks like we only support `layout.scene.dragmode: false` (ie 3D), not `layout.dragmode: false`
@rreusser good catch that this should not be disabled in zoom mode anyway.

rreusser: The problem then is that `dragmode: false` is not a valid coercible default. `gl3d` seems to sidestep the coerced default and look at the user data instead of sanitized data. Which works, though it's perhaps a bit sketchy.

rreusser: Oh, I see. It *inherits* false from `layout.dragmode` despite `layout.dragmode: false` not actually being a valid default. I'm okay with that.

etpinard: Oopppppppps, that line is in `gl2d/camera.js (i.e. **NOT** gl3d). Funny that thing has been there since [v1.0.0](https://github.com/plotly/plotly.js/blame/master/src/plots/gl2d/camera.js#L110). 

:hocho: :hocho: :hocho:  :hocho:  :hocho:  :hocho:

rreusser:
Okay, I've 1/2'd all zoom speeds (including svg, which previously wasn't affected by this PR) and `scattergl` now obeys `config.scrollZoom`.

etpinard:
> I think scatter and scattergl should match, but scatter3d should have wheel zoom enabled always.

:+1: for me here.

But, I think in `v2` all subplot types should obey `scrollZoom`. Moreover, I think should add a corresponding `layout` attribute as people might want to turn on/off scroll interactions via a `relayout` call. Thoughts?

rreusser:
Agreed. I actually rather dislike scrollZoom being a config parameter. I find embedplot, for one, suddenly has dramatic scroll interactions I never saw when developing the plot.

src/plots/gl2d/camera.js
rreusser: Yeah. Definitely undesirable. 😄 🔪

rreusser: (next most glaring thing, as briefly mentioned to @alexcjohnson is double-click to revert zoom to autoscale, which is the top top top thing that always gets me about scattergl plots.)

etpinard: > which is the top top top thing that always gets me about scattergl plots.)

No need to waste time on this. Double-click will _just work_ after @dfcreative 's https://github.com/plotly/plotly.js/pull/1869

rreusser: Works for me. Should have bothered to do it long ago, but I'm content to wait for #1869. 🎉

rreusser:
Okay, I'm now content with this. Had to fix test values. The big change is cartesian zoom speed *= 1/2. Which doesn't seem all that noticeable because of inertia. A much bigger change is gl zoom speed *= 5.

alexcjohnson:
💃  from my side. I have more thoughts about `scrollZoom` but we can discuss that more in the v2 issue #420

etpinard:
@rreusser what about those scroll factors. Looks like cartesian, gl2d and gl3d all use different one. Is that why you didn't bother putting them in a `constants.js` file?

rreusser:
@etpinard Yes. I debated whether to copy the window size logic. `gl3d`, for one, takes for granted that the *window* height is actually relevant: 

```javascript
-camera.zoomSpeed * flipY * dy / window.innerHeight * (t - view.lastT()) / 100.0;
```

Perhaps that should be modified to use the *plot* height and conform to the same math as SVG.

etpinard:
> Perhaps that should be modified to use the plot height and conform to the same math as SVG.

No need. I was just curious why my comment didn't get addressed.

I'm happy with the current state of this PR :dancer:

Issue #170, opened 2016-01-07
Lasso and select-box support for traces other than scatter marker and scatter text


etpinard:
Lasso and select-box were added in https://github.com/plotly/plotly.js/pull/154

Adding support for other trace types should be as easy as adding a `selectPoints` to the other trace modules.

etpinard:
cc @dfcreative @chriddyp @jackwluo

jackparmer:
Top ones for crossfilter project:
- [ ] Bar charts
- [ ] Histograms
- [ ] Choropleths
- [ ] geoscatter

jackparmer:
For bar chart and histogram selection, I think just touching a bar should include it in a selection. Thoughts @cpsievert based on plotly.js selections in Shiny apps?

etpinard:
For the record, I'd vote for having to select 1/2 of the bar's area. Not a strong opinion though.

morganh35:
I am anxious to get this feature working and would be willing to help implement it. 
Is there an estimate of the timeline for this feature?

cpsievert:
> For bar chart and histogram selection, I think just touching a bar should include it in a selection. 

Yea, we can already link views in this way via the R package (via click or hover):

![plotlylinkedbars](https://user-images.githubusercontent.com/1365941/28929768-d0d081a6-783e-11e7-88d8-757dfd62bb8f.gif)



> Thoughts @cpsievert based on plotly.js selections in Shiny apps?

R users can leverage most common plotly.js events (e.g., ,`plotly_click`, `plotly_hover`, `plotly_selected`, `plotly_relayout`, etc) both inside and outside of shiny. Regardless, being able to brush multiple bars would be of some use. 

> I'd vote for having to select 1/2 of the bar's area. Not a strong opinion though.

No strong opinion here either, but I think I'd prefer that we force selection of the entire bar. One reason being that you could imagine a system where you could partially select a bar that represented multiple observations

jackparmer:
@morganh35 We'll be rolling out lasso or box select for the rest of Plotly.js trace types over the next few months. If you need this feature for a project at your workplace, we'd love to [hear from you](https://plotly.typeform.com/to/mH1Cpb)

monfera:
For the record, there are some traces that work with `crossfilter` in a view mode, ie. crossfiltering yields a feedback, even though they weren't yet made suitable for `crossfilter`:

[...]

Issue #1862, opened 2017-07-07
Add `marker.opacity` to bar traces


saimaparveen:
Hi,

Opacity of bar chart not changing.
https://codepen.io/saimaparveen/pen/WOgKYJ

If i remove type:'bar' for scatter plot opacity is working fine.

Thanks,
Saima Parveen

etpinard:
CodePen w/o commented-out blocks: https://codepen.io/etpinard/pen/xrayzz?editors=1010

@saimaparveen Thanks for report!

saimaparveen:
Is there any workaround for this issue?

etpinard:
> Is there any workaround for this issue?

Try plotting separate traces with different scalar opacity values or wait for a patch.

etpinard:
After looking at this more closely, it appears that `marker.opacity` has never been part of `bar` trace schema in `v1.x`. So this ticket isn't a bug. Passing the `data` array from the codepen above to `Plotly.validate` correctly outputs _key marker.opacity is not part of the schema_ .

 I wouldn't be opposed to adding `marker.opacity` to `bar` traces, but this will be a low-priority ticket in the short term. Community PRs are welcomed though.

Note that trace-wide `opacity` scalar values due work for bars and setting `marker.color` as `rgba` values also works. See https://codepen.io/etpinard/pen/JJqjgy?editors=1010

Issue #1848, opened 2017-07-04
Persistent, stateful hover, click, and selection properties


chriddyp:
In Dash, the `Graph` component stores the data from hover, click, and selection events as part of the component's state and passes that to the user as `hoverData`, `clickData`, `selectionData`.

In some cases, the user may want to initialize their app with a point already "clicked", a region "preselected", or a point "hovered".

This is related to "[Customized Click, Hover, and Selection Styles](https://github.com/plotly/plotly.js/issues/1847)". The viewer will know that the graph has been pre-selected, clicked, or hovered through whatever custom styles the developer added.

Ideally, this API would match the same event data. For example, when hovering over a point, the event data might be:

```
{
    "points": [{"x": 1, "y"; 4}]
}
```

Ideally, the user would be able to pass that in as part of the figure to trigger those styles themselves. That's how the Dash API currently works although styles are not applied:
```
Graph(
    figure={...},
    hoverData={"points": [{"x": 1, "y": 4}]}
)
```

***

cc @alexcjohnson @etpinard @monfera @jackparmer @charleyferrari

etpinard:
So,

```js
Plotly.newPlot('graph', {
  data: [{
    mode: 'markers',
    x: [1, 2, 3],
    y: [2, 1, 2]
  }],
  hoverdata: {
    points: [{x: 1, y: 2}]
  })
```

would render the graphs and one hover label at `(x=1, y=2)`?

etpinard:
See also https://github.com/plotly/plotly.js/issues/257

jackparmer:
An issue from an R / Shiny user on this subject:
https://community.plot.ly/t/how-do-i-retain-the-information-about-both-the-lasso-selection-and-the-clicked-point/5849

rreusser:
Here's somewhat of a breakdown of how I'm thinking this can exist:

- `hoverpoints` and `hoverids` (`selectedpoints` + selectedids`?) seem like properties that go _somewhere_. My preference would be for putting them in `data` itself along with the traces they affect. I don't necessarily see points selected across traces as mutually exclusive or needing to be specified outside the traces themselves. A second choice would be for them to go in layout along with a trace number, but it seems like this could cause the usual trace indexing headaches.
- these seem like general trace properties that apply to at least *many* types of traces, if not all. This suggests they should be defined outside particular traces and merged into specific trace attributes where they apply.
- @etpinard what did we decide was best for the attribute structure for modified properties?
    ```javascript
    {
[...]

PR #2045, opened 2017-09-29
Bar | histogram lasso / select box dragmode  

etpinard:
Knocking out a few more items in https://github.com/plotly/plotly.js/issues/170

peek 2017-09-29 12-04


src/traces/bar/plot.js
```diff
@@ -69,12 +70,18 @@ module.exports = function plot(gd, plotinfo, cdbar) {
...
+                        di.ct = [x1, (y0 + y1) / 2];
                     }
                     else {
                         x0 = xa.c2p(p0, true);
                         x1 = xa.c2p(p1, true);
                         y0 = ya.c2p(s0, true);
                         y1 = ya.c2p(s1, true);
+
+                        // for selections
+                        di.ct = [(x0 + x1) / 2, y1];
```
etpinard: :sos: debated logic alert :sos: 

I chose to make a bar _selected_ when the lasso/select polygon contains this point here which corresponds to the middle of the position coords at full span on the size axis.

Looking for :+1: s from @alexcjohnson @rreusser  @monfera @chriddyp @cpsievert

src/traces/bar/select.js
```diff
@@ -0,0 +1,52 @@
...
+                    y: di.y
+                });
+                di.dim = 0;
+            } else {
+                di.dim = 1;
+            }
+        }
+    }
+
+    node3.selectAll('.point').style('opacity', function(d) {
```
etpinard: This operation here is similar to what we currently do for `scatter` traces. 

But note that, unlike `scatter` traces, `bar` traces do not have a `marker.opacity` attribute (see https://github.com/plotly/plotly.js/issues/1862) so currently, this here is equivalent to adding an alpha channel to `marker.color`, `marker.line.color` and `textcolor`.

cc https://github.com/plotly/plotly.js/issues/1848

src/traces/bar/plot.js
cpsievert: That seems reasonable (especially with a 1-to-1 mapping from the (input) data value to bar), but another important question is what data do you attach to the select event when a single bar represents an aggregation of multiple values? It seems important to provide options as to how you'd like to respond to the event, so I if it were me, I would emit the raw data, the aggregated data, and the selection region (on the data scale).

etpinard: > so I if it were me, I would emit the raw data, the aggregated data, and the selection region (on the data scale).

Right now, we emit raw data and selection region. Adding the aggregated data is an interesting idea.

cpsievert: And, just to be clear, assuming the model doesn't allow for partially selecting a bar that represents multiple values, I would actually vote for the selection region to cover the *entire* bar in order for a selection to be made...

The problem with that though is that it will be *very* confusing for most users if we implement two different selection criteria based on whether an aggregation is taking place

etpinard: > assuming the model doesn't allow for partially selecting a bar that represents multiple values,

Here, are you referring to histograms or bar traces with `barmode: 'stack'`?

cpsievert: Furthermore, partial selection is an advanced concept that advanced users could still hack together by stacking together individual observations so that they "look aggregated"

cpsievert: > Here, are you referring to histograms or bar traces with barmode: 'stack'?

Histograms

etpinard: Ok, so maybe `histogram` and `bar` should have different select logic? 

- Selecting a bar would mean encircling the most data-like pt on the bar
- Selecting an histogram item would mean encircling the entire rectangle

cpsievert: > Ok, so maybe histogram and bar should have different select logic?

I'm leaning towards yes the more I think about it. I was apprehensive at first to suggest different criteria since it's potentially confusing, but the simple fact that selecting the whole bar will *always* work makes up for it

alexcjohnson: I would offer a few arguments in favor of just keeping the "select the end" behavior:
- Some users do their own aggregation, making a `bar` trace that conceptually represents a histogram. What would we do in that case, introduce an option for whether to select the end or the whole thing?
- It would be meaningless (or at least ambiguous) to disaggregate by selecting a portion of the bar. Which sample is on top vs on bottom?
- All our other selections work by collapsing the entity to a single point to be selected
- It makes it that much harder to select just the bars you want: you could still select all the *small* bars, but you wouldn't have any way to do the likely more common act of selecting all the *largest* bars.

cldougl: I would lean towards a consistent logic for bars & hists (and maybe the possibility of adding modes in the future). I think having a different behaviour across those two charts would cause more confusion than benefits across users, especially given how often I see users creating their own "histograms" with the `bar` trace as @alexcjohnson mentioned above.

test/jasmine/tests/select_test.js
```diff
@@ -708,4 +722,100 @@ describe('Test select box and lasso per trace:', function() {
...
+        Plotly.plot(gd, fig)
+        .then(function() {
+            return _run(
+                [[350, 200], [400, 200], [400, 250], [350, 250], [350, 200]],
+                function() {
+                    assertPoints([
+                        [0, 4.9, 0.371], [0, 5, 0.368], [0, 5.1, 0.356], [0, 5.2, 0.336],
+                        [0, 5.3, 0.309], [0, 5.4, 0.275], [0, 5.5, 0.235], [0, 5.6, 0.192],
+                        [0, 5.7, 0.145],
+                        [1, 5.1, 0.485], [1, 5.2, 0.40900000000000003], [1, 5.3, 0.327],
```
alexcjohnson: can we use `toBeCloseTo` above in `makeAssertPoints` and avoid `0.40900000000000003` etc?

src/traces/bar/plot.js
rreusser: More attributes are probably not a good idea, but perhaps the bar default would be data endpoints and the histogram default would be centroid. And if you dig deep enough to notice this subtlety, it'll lead to an attribute that can be toggled. If additional attributes had no cost, that'd be my preference, but additional attributes *do* have some small but nonzero cost (bundle size, documentation size, discoverability, maintenance, sprawl, configurability over good defaults, etc.)

chriddyp: > I think having a different behaviour across those two charts would cause more confusion than benefits across users

I agree. I think there should be one simple and predictable way for this to work.

If the app developer really wants to allow users to select "partial bars" because it represents important aggregated data, then I'd argue that they should be more explicit about this in their UI and draw a rug plot below their histogram and allow users to select the exact partial sets of points that way. Or even a box plot with points or eventually a violin plot with points.

test/jasmine/tests/select_test.js
etpinard: done in 3c73301

src/traces/bar/plot.js
etpinard: Ok. Thanks everyone for your input. I believe the _keep-as-is_ side wins.

alexcjohnson:
No more comments from my side - nicely done! 💃

PR #2046, opened 2017-10-02
Annotations offline

### LINKED ISSUE #992: The annotation arrow isn't rendered properly on load
Annotations seem to be missing the arrow head when the graph is first rendered. Strangely, once I either change the zoom, or click on one of the axes, the arrowhead changes to the correct display.

Here's what it looks like onload:
plotly annotation bug

This is what it looks like after doing one of those actions I mentioned:
plotly annotation bug correct

alexcjohnson:
Supposed to be a quick Friday night bugfix but it turned a bit more involved:
- Fixes #992 (annotations not styling correctly if the plot is off-DOM when it's first rendered).
- Updates the style of `drawArrowHead` a bit.
- Tests (via `strict-d3`) that we don't use `selection.style` as a getter there or anywhere else.
- Gets rid of all the other uses of `selection.style` as a getter - some others actually could cause bugs of their own (like in `colorbar.draw`), some were benign (if they dealt with eg hover labels so we know the element is displayed) but it's easier - and in some cases simpler code even - to just be consistent in not using it.

cc @etpinard

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -212,7 +212,12 @@ plots.redrawText = function(gd) {
 plots.resize = function(gd) {
     return new Promise(function(resolve, reject) {
 
-        if(!gd || d3.select(gd).style('display') === 'none') {
+        function isHidden(gd) {
+            var display = getComputedStyle(gd).display;
+            return !display || display === 'none';
+        }
+
+        if(!gd || isHidden(gd)) {
             reject(new Error('Resize must be passed a plot div element.'));
         }
 
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -213,12 +213,12 @@ plots.resize = function(gd) {
     return new Promise(function(resolve, reject) {
 
         function isHidden(gd) {
-            var display = getComputedStyle(gd).display;
+            var display = window.getComputedStyle(gd).display;
             return !display || display === 'none';
         }
 
         if(!gd || isHidden(gd)) {
-            reject(new Error('Resize must be passed a plot div element.'));
+            reject(new Error('Resize must be passed a displayed plot div element.'));
         }
 
         if(gd._redrawTimer) clearTimeout(gd._redrawTimer);
--- a/test/jasmine/assets/custom_assertions.js
+++ b/test/jasmine/assets/custom_assertions.js
@@ -49,14 +49,14 @@ exports.assertStyle = function(dims, color, opacity) {
 exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
     if(!msg) msg = '';
 
-    var path = g.select('path').node();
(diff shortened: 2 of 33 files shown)


src/plots/plots.js
```diff
@@ -212,7 +212,12 @@ plots.redrawText = function(gd) {
 plots.resize = function(gd) {
     return new Promise(function(resolve, reject) {
 
-        if(!gd || d3.select(gd).style('display') === 'none') {
+        function isHidden(gd) {
+            var display = getComputedStyle(gd).display;
```
etpinard: I'd prefer `window.getComputedStyle` here.

test/jasmine/assets/custom_assertions.js
```diff
@@ -37,16 +37,28 @@ exports.assertStyle = function(dims, color, opacity) {
...
                 .toEqual(opacity[i], 'to have correct pt opacity');
         });
     });
 };
 
+exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
+    if(!msg) msg = '';
+
+    var path = g.select('path').node();
+    expect(getComputedStyle(path).fill).toBe(expectation.bgcolor, msg + ': bgcolor');
```
etpinard: `window.getComputedStyle`

src/plots/plots.js
etpinard: ... in an effort to make plotly.js work in jsdom

etpinard: and readability (I think).

test/jasmine/assets/custom_assertions.js
etpinard: ... but this is in a test file. I don't care about that as much. Your call.

test/jasmine/assets/custom_assertions.js
```diff
@@ -37,16 +37,28 @@ exports.assertStyle = function(dims, color, opacity) {
...
                 .toEqual(color[i], 'to have correct pt color');
-            expect(+point.style('opacity'))
+            var op = this.style.opacity;
+            expect(op === undefined ? 1 : +op)
                 .toEqual(opacity[i], 'to have correct pt opacity');
         });
     });
 };
 
+exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
```
etpinard: :tada:

test/jasmine/assets/custom_assertions.js
alexcjohnson: updated them all -> https://github.com/plotly/plotly.js/pull/2046/commits/78f5222c9c93e875eb758738d13ccd1054c0e549

tested that we only call `getComputedStyle` on `window` and only as many times as expected -> https://github.com/plotly/plotly.js/pull/2046/commits/97cea609b9fb75f5cec9e40893497b4b831cc0d2

etpinard:
Nicely done :dancer:

PR #2050, opened 2017-10-02
Bar axis autorange fix

### LINKED ISSUE #2049: Something is up with bar autorange with y minute-precision data on date axes
See https://codepen.io/etpinard/pen/mBMgLR

First reported in https://community.plot.ly/t/bar-chart-axis-not-showing-duration-properly-line-chart-does/5990/

alexcjohnson:
fixes #2049 - cleaner handling of base and zero for bar axis autorange.

We don't need to include 0 explicitly in the autorange for bars (unless we're normalizing), nor should we be including the bar base unless that was explicitly specified.

Most of the time these extras were just redundant, but for bars on date axes (which is not really a kosher use of bars but we don't forbid it) the zero (1970) is arbitrary and it's confusing and useless if this gets included in the autorange.

cc @etpinard

### DIFF
--- a/src/traces/bar/calc.js
+++ b/src/traces/bar/calc.js
@@ -63,7 +63,11 @@ module.exports = function calc(gd, trace) {
     if(Array.isArray(base)) {
         for(i = 0; i < Math.min(base.length, cd.length); i++) {
             b = sa.d2c(base[i], 0, scalendar);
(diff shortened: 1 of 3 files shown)


etpinard:
Looks good :dancer: 

Thanks :tada:

PR #2052, opened 2017-10-03
Table view - squashed

monfera:
Squashed version of #1834 

Before starting review, I'd push proper mocks, also as codepen examples. I'll also need to recheck attributes and add jasmine tests.


src/traces/table/attributes.js
```diff
@@ -0,0 +1,278 @@
...
+            dflt: [],
+            description: [
+                'Dimension values. `values[n]` represents the value of the `n`th point in the dataset,',
+                'therefore the `values` vector for all dimensions must be the same (longer vectors',
+                'will be truncated). Each value must be a finite number.'
+            ].join(' ')
+        },
+
+        format: {
+            valType: 'data_array',
```
etpinard: Maybe this should be an `arrayOk` `valType: 'string'`, to format all values to same way more easily?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,278 @@
...
+            valType: 'string',
+            arrayOk: true,
+            dflt: null,
+            role: 'style',
+            description: 'Suffix for cell values.'
+        },
+
+        height: {
+            valType: 'number',
+            dflt: 28,
```
etpinard: Interesting number. How did you come up with that?

src/traces/table/attributes.js
```diff
@@ -0,0 +1,278 @@
...
+                arrayOk: true,
+                role: 'style',
+                description: [
+                    'Sets the cell fill color. It accepts either a specific color',
+                    ' or an array of colors.'
+                ].join('')
+            }
+        },
+
+        font: {
```
etpinard: please use https://github.com/plotly/plotly.js/blob/master/src/plots/font_attributes.js

src/traces/table/attributes.js
```diff
@@ -0,0 +1,278 @@
...
+    },
+
+    cells: {
+
+        values: {
+            valType: 'data_array',
+            role: 'info',
+            dflt: [],
+            description: [
+                'Dimension values. `values[n]` represents the value of the `n`th point in the dataset,',
```
etpinard: Shouldn't `values` be 2D? That is, `values[n][m]` represents the value of ....

src/traces/table/attributes.js
```diff
@@ -0,0 +1,278 @@
...
+                role: 'style'
+            },
+            color: {
+                valType: 'color',
+                arrayOk: true,
+                role: 'style'
+            }
+        }
+    }
+};
```
etpinard: You'll need to add `editType` declarations to these attributes.

See https://github.com/plotly/plotly.js/pull/1999

monfera:
@etpinard Bird mock is now added, I'll need to replace the other tech test mocks too (they have JS etc.)

image

src/traces/table/calc.js
```diff
@@ -0,0 +1,131 @@
...
+    columnWidths = columnWidths.map(function(d) {return d / totalColumnWidths * groupWidth;});
+
+    var calcdata = {
+        key: trace.index,
+        translateX: domain.x[0] * gd._fullLayout._size.w,
+        translateY: gd._fullLayout._size.h - domain.y[1] * gd._fullLayout._size.h,
+        size: gd._fullLayout._size,
+        width: groupWidth,
+        height: groupHeight,
+        columnOrder: columnOrder, // will be mutated on column move, todo use in callback
```
etpinard: Still TODO?

src/traces/table/calc.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+    var totalColumnWidths = columnWidths.reduce(function(p, n) {return p + n;}, 0);
+
+    // fit columns in the available vertical space as there's no vertical scrolling now
+    columnWidths = columnWidths.map(function(d) {return d / totalColumnWidths * groupWidth;});
+
+    var calcdata = {
+        key: trace.index,
+        translateX: domain.x[0] * gd._fullLayout._size.w,
+        translateY: gd._fullLayout._size.h - domain.y[1] * gd._fullLayout._size.h,
```
etpinard: `(1 - domain.y[1]) * gd._fullLayout._size.h`

src/traces/table/calc.js
```diff
@@ -0,0 +1,131 @@
...
+    return wrap(calcdata);
+};
+
+function xScale(d) {
+    return d.calcdata.columns.reduce(function(prev, next) {
+        return next.xIndex < d.xIndex ? prev + next.columnWidth : prev;
+    }, 0);
+}
+
+function makeRowBlock(anchorToRowBlock, auxiliary) {
```
etpinard: Might be nice to reuse some of the `heatmap/` calc routines. For example `clean2dArray` for cells.values.

@monfera have you looked into `Heatmap.calc`?

src/traces/table/constants.js
```diff
@@ -0,0 +1,33 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
+    maxDimensionCount: 60,
+    overdrag: 45,
+    cellPad: 8,
+    latexCheck: /^\$.*\$$/,
```
etpinard: That's probably something we have already in https://github.com/plotly/plotly.js/blob/master/src/lib/svg_text_utils.js

src/traces/table/defaults.js
```diff
@@ -0,0 +1,66 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Lib = require('../../lib');
+var attributes = require('./attributes');
+
+function defaultColumnOrder(traceIn, coerce) {
```
etpinard: Can this be done in `calc`?

src/traces/table/index.js
```diff
@@ -0,0 +1,30 @@
...
+Table.plot = require('./plot');
+
+Table.moduleType = 'trace';
+Table.name = 'table';
+Table.basePlotModule = require('./base_plot');
+Table.categories = [];
+Table.meta = {
+    description: [
+        'Table view for multidimensional exploratory data analysis.',
+        'The samples are specified in `dimensions`.',
```
etpinard: Hmm. That is not right.

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+var Drawing = require('../../components/drawing');
+var extendFlat = require('../../lib/extend').extendFlat;
+var svgUtil = require('../../lib/svg_text_utils');
+var raiseToTop = require('../../lib').raiseToTop;
+var cancelEeaseColumn = require('../../lib').cancelTransition;
+
+module.exports = function plot(gd, calcdata) {
+
+    if(c.clipView) {
+        gd._fullLayout._paper.attr('height', 2000);
```
etpinard: `clipView: false` in constants.js. So is this block only useful for debugging? It so, :hocho: please.

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+    table.enter()
+        .append('g')
+        .classed('table', true)
+        .attr('overflow', 'visible')
+        .style('box-sizing', 'content-box')
+        .style('position', 'absolute')
+        .style('left', 0)
+        .style('overflow', 'visible')
+        .style('shape-rendering', 'crispEdges')
+        .style('pointer-events', 'all'); // todo restore 'none'
```
etpinard: Why would we need to restore `'none'` here?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        .attr('transform', function(d) {
+            return 'translate(' + d.translateX + ',' + d.translateY + ')';
+        });
+
+    var tableControlView = table.selectAll('.tableControlView')
+        .data(gup.repeat, gup.keyFun);
+
+    tableControlView.enter()
+        .append('g')
+        .classed('tableControlView', true)
```
etpinard: `table-control-view` - use dash-case for class names please.

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        .style('box-sizing', 'content-box')
+        .on('mousemove', function() {tableControlView.call(renderScrollbarKit, gd);})
```
etpinard: Have you thought about `touch` handlers?

Have you looked into [`dragElement`](https://github.com/plotly/plotly.js/blob/master/src/components/dragelement/index.js) to DRY this up?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+            makeDragRow(gd, tableControlView, null, d.scrollY + d3.event.deltaY)(d);
+            d.scrollbarState.wheeling = false;
+        })
+        .call(renderScrollbarKit, gd);
+
+    tableControlView
+        .attr('transform', function(d) {return 'translate(' + d.size.l + ' ' + d.size.t + ')';});
+
+    if(!c.clipView) {
+        tableControlView.attr('clip-path', function(d) {return 'url(#scrollAreaBottomClip_' + d.key + ')';});
```
etpinard: Please use [`Drawing.setClipUrl`](https://github.com/plotly/plotly.js/blob/8b35a8e0a0484466d1f49cde769321aa6a120a35/src/components/drawing/index.js#L766).

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+
+    cellsColumnBlock
+        .call(d3.behavior.drag()
+            .origin(function(d) {
+                d3.event.stopPropagation();
+                return d;
+            })
+            .on('drag', makeDragRow(gd, tableControlView, -1))
+            .on('dragend', function() {
+                // fixme emit plotly notification
```
etpinard: Yes, this should emit `plotly_restyle`.

src/lib/gup.js
```diff
@@ -0,0 +1,34 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var identity = require('./identity');
+
+function wrap(d) {return [d];}
+
+module.exports = {
```
etpinard: I like this. Thanks @monfera :tada:

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        .data(gup.repeat, gup.keyFun);
+
+    columnBoundary.enter()
+        .append('g')
+        .classed('columnBoundary', true);
+
+    var columnBoundaryClippath = yColumn.selectAll('.columnBoundaryClippath')
+        .data(gup.repeat, gup.keyFun);
+
+    // SVG spec doesn't mandate wrapping into a <defs> and doesn't seem to cause a speed difference
```
etpinard: Please add your `<clipPath>` nodes in `fullLayout._clips`, for consistency with other trace types.

src/traces/table/plot.js
etpinard: ... and those class names would be best defined in `table/constants.js`

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        .style('pointer-events', 'stroke');
+
+    columnBoundaryRect
+        .attr('width', function(d) {return d.columnWidth;})
+        .attr('height', function(d) {return d.calcdata.height + c.uplift;});
+
+    updateBlockYPosition(null, cellsColumnBlock, tableControlView);
+};
+
+function renderScrollbarKit(tableControlView, gd) {
```
etpinard: Another scroller. That's now three separate implementations (legend, update menus and this here). :rage:

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+            d.valign = gridPick(d.calcdata.cells.valign, col, i);
+            d.cellBorderWidth = gridPick(d.calcdata.cells.line.width, col, i);
+            d.font = font;
+        });
+}
+
+function sizeAndStyleRect(cellRect) {
+    cellRect
+        .attr('width', function(d) {return d.column.columnWidth;})
+        .attr('stroke-width', function(d) {return d.cellBorderWidth;})
```
etpinard: Please use [`Color.stroke`](https://github.com/plotly/plotly.js/blob/8b35a8e0a0484466d1f49cde769321aa6a120a35/src/components/color/index.js#L84) :recycle: whenever you can. That split between `stroke` and `stroke-opacity` is for Adobe Illustrator compat I believe.

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+            var userBrokenText = (typeof userSuppliedContent !== 'string') || userSuppliedContent.match(/<br>/i);
+            var userBrokenText2 = (typeof userSuppliedContent === 'string') && userSuppliedContent.match(/<br>/i);
+            var prefix = latex ? '' : gridPick(d.calcdata.cells.prefix, col, row) || '';
+            var suffix = latex ? '' : gridPick(d.calcdata.cells.suffix, col, row) || '';
+            var format = latex ? null : gridPick(d.calcdata.cells.format, col, row) || null;
+            var prefixSuffixedText = prefix + (format ? d3.format(format)(d.value) : d.value) + suffix;
+            d.latex = latex;
+            d.mayHaveMarkup = (typeof userSuppliedContent === 'string') && userSuppliedContent.match(/[<&>]/);
+            var hasWrapSplitCharacter;
+            var hwsc = function(prefixSuffixedText) {return prefixSuffixedText.indexOf(c.wrapSplitCharacter) !== -1;};
```
etpinard: Oh boy. I really wished the text-wrapping logic would've been added to `svg_text_utils.js` so that other parts of plotly.js can use it.

@monfera is there any reason to keep this here?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+
+function columnMoved(gd, calcdata, i, indices) {
+    var o = calcdata[i][0].gdColumnsOriginalOrder;
+    calcdata[i][0].gdColumns.sort(function(a, b) {
+        return indices[o.indexOf(a)] - indices[o.indexOf(b)];
+    });
+
+    calcdata[i][0].columnorder = indices;
+
+    gd.emit('plotly_restyle');
```
etpinard: ... with corresponding event data.

etpinard: ... as reordering columns (I hope) can be done via `Plotly.restyle` too

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+}
+
+function splitToCells(d) {
+    var fromTo = rowFromTo(d);
+    return d.values.slice(fromTo[0], fromTo[1]).map(function(v, i) {
+        // By keeping identical key, a DOM node removal, creation and addition is spared, important when visible
+        // grid has a lot of elements (quadratic with xcol/ycol count).
+        // But it has to be busted when `svgUtil.convertToTspans` is used as it reshapes cell subtrees asynchronously,
+        // and by that time the user may have scrolled away, resulting in stale overwrites. The real solution will be
+        // to turn `svgUtil.convertToTspans` into a cancelable request, in which case no key busting is needed.
```
etpinard: > to turn `svgUtil.convertToTspans` into a cancelable request

Why not in this PR :smile: ?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        // block.allRowsHeight relies on being invalidated, so enabling this opt may not be safe
+        // if(pages.length > 1) break;
+    }
+
+    return pages;
+}
+
+function updateBlockYPosition(gd, cellsColumnBlock, tableControlView) {
+    // fixme this function is THE performance hotspot
+    var d = cellsColumnBlock[0][0].__data__;
```
etpinard: `cellsColumnBlock.data()`

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        });
+
+    cellsColumnBlock
+        .attr('transform', function(d) {
+            var yTranslate = firstRowAnchor(d.rowBlocks, d.page) - d.scrollY;
+            return 'translate(0 ' + yTranslate + ')';
+        });
+
+    // conditionally rerendering panel 0 and 1
+    if(gd) {
```
etpinard: How can `gd` be falsy here?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+        d.scrollY = optionalPosition === void(0) ? d.scrollY + multiplier * d3.event.dy : optionalPosition;
+        var cellsColumnBlock = tableControlView.selectAll('.yColumn').selectAll('.columnBlock').filter(cellsBlock);
+        updateBlockYPosition(gd, cellsColumnBlock, tableControlView);
+    };
+}
+
+function conditionalPanelRerender(gd, tableControlView, cellsColumnBlock, pages, prevPages, d, revolverIndex) {
+    var shouldComponentUpdate = pages[revolverIndex] !== prevPages[revolverIndex];
+    if(shouldComponentUpdate) {
+        window.clearTimeout(d.currentRepaint[revolverIndex]);
```
etpinard: no need to spell out `window` when using `setTimeout` and `clearTimeout`

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+            //                               scrollable window, as increases to row heights may need scrollbar updates
+            renderScrollbarKit(tableControlView, gd);
+        }
+
+        cellTextHolder
+            .attr('transform', function() {
+                // this code block is only invoked for items where d.cellHeightMayIncrease is truthy
+                var element = this;
+                var columnCellElement = element.parentNode;
+                var box = columnCellElement.getBoundingClientRect();
```
etpinard: Can we avoid these `getBoundingClientRect` calls?

src/traces/table/plot.js
```diff
@@ -0,0 +1,892 @@
...
+                var rectBox = d3.select(element.parentNode).select('.cellRect').node().getBoundingClientRect();
+                var currentTransform = element.transform.baseVal.consolidate();
+                var yPosition = rectBox.top - box.top + (currentTransform ? currentTransform.matrix.f : c.cellPad);
+                return 'translate(' + xPosition(d, d3.select(element.parentNode).select('.cellTextHolder').node().getBoundingClientRect().width) + ' ' + yPosition + ')';
+            });
+
+        // MathJax styling has to be killed for HTML DOM color specifications to seep through
+        cellTextHolder.selectAll('svg').selectAll('*')
```
etpinard: Looks hacky. Can we do better?

test/jasmine/tests/table_test.js
```diff
@@ -0,0 +1,972 @@
...
+        });
+
+        it('\'domain\' specification should have a default', function() {
+            var fullTrace = _supply({});
+            expect(fullTrace.domain).toEqual({x: [0, 1], y: [0, 1]});
+        });
+
+        it('\'dimension\' specification should have a default of an empty array', function() {
+            var fullTrace = _supply({});
+            expect(fullTrace.dimensions).toEqual([]);
```
etpinard: What is this? A copy from `parcoords`?

test/jasmine/tests/table_test.js
```diff
@@ -0,0 +1,972 @@
...
+            }));
+
+            expect(fullTrace.line).toEqual({
+                color: '#444'
+            });
+        });
+    });
+});
+
+describe('@noCI table', function() {
```
etpinard: Why `@noCI`?

etpinard:
A few (opinionated) comment from https://codepen.io/monfera/pen/ZXXMLo?editors=1010#0

- I find column swapping a little too _bubbly_:

peek 2017-10-03 11-08

Is there an easy way to reduce the recoil action on drag-end?

- It would be nice to add `plotly_hover` and `plotly_click` events on the cells:
- It would be nice to make the header values (and maybe even the cells?) `editable: true`

test/image/mocks/table_plain_birds.json
```diff
@@ -0,0 +1,90 @@
...
+            "font": {
+                "family": "Arial",
+                "size": [[14, 12]],
+                "color": ["white", "white", "white", "white", "white", ["pink", "white"], "white"]
+            }
+        },
+
+        "cells": {
+
+            "values": [
```
etpinard: Hmm. So here `values[i]` represent columns - which I think is fair for `table`. But this means that `cells.values` is the transpose of the `z` attribute in `heatmap` traces.  Which at the very least write this down in the `table/index.js` module description.

src/lib/gup.js
etpinard: Would you mind reusing this in `parcoords`.

src/traces/table/plot.js
monfera: Yes, invaluable for development even to the last touches, planning to 🔪 it just before the merge

src/traces/table/constants.js
monfera: Will check!

src/traces/table/attributes.js
monfera: Just took a look at what'd look balanced with a font height of 12px although it could as well be computed
image

src/traces/table/attributes.js
monfera: I'll need to revise all attribute descriptions, indeed it's a nested array:
image

src/traces/table/calc.js
monfera: I thought that a callback on column rearrangement (and scroll) would be useful but as I didn't get this as a piece of requirement, more pressing things took my time. I left in comments to revisit, whether in this PR or in a follow-up PR. Due to the target date, a follow-up PR would be safer.

etpinard: Thanks for the info. No action required. :+1:

src/traces/table/calc.js
monfera: Thanks, I'll look into reusability there!

src/traces/table/plot.js
monfera: I'm on the fence about `pointer-events` - right now, one can't copy-paste cell text, as its results would be interesting in cases of LaTeX or other formatted cell content. But maybe a better choice would be to enable copying text in which case `'none'` won't be good.

`'none'` is good if table cell contents are not meant to be copied, ie. the user only scrolls up/down and rearranges columns. During development, I rarely activate `'none'` because it makes Dev Tools element selection nigh impossible; `todo` is an easy to find tagging of what to revisit.

alexcjohnson:
thinking ahead to `Plotly.react`, we'll probably need an attribute for scroll position, so the trace attributes fully specify the view. Also for export, people may want to be able to snapshot at the current view.

src/traces/table/plot.js
monfera: I'll do the classname conversion to dashes, serves code unity! Seen it both w/ dashes and even all-lowercase w/o separators in `plotly.js`. Btw. if there's technical merit pro/cons I'd be interested to know too.

Glad to refactor classnames into `constants` and also, the selection `'.classname'`s can be glued together from that. I'd probably leave the this 2nd thing to the end so higher prio changes can be done 1st (worst case, minor, isolated follow-up PR w/ extraction, also including the same dashing+extraction for `parcoords`).

rreusser:
@alexcjohnson That's true. The only caveat is that high-frequency updates like scroll position shouldn't go through controlled update cycles or even `setState` (which is delayed by one frame and so makes things unnecessarily jumpy). In other words, I don't think the scroll position should be controlled. Which is equivalent to saying it would just be `initialScrollY`/`initialScrollX` or similar.

src/traces/table/plot.js
monfera: Yes, both column drag&drop and scrolling work on mobile, though on the codepen exampe it also drags the document somewhat so it's not immediately obvious (iOS needs page-level specs so that the document not be dragged around, I can add a demo later). But _kinetic_ scrolling doesn't now work on mobile (not sure of its urgency).

I'll look into `dragElement` to see if it'd yield code saving or unification benefit. On an initial look, immersing myself would take significant time as `dragElement` is hundreds of code lines and modifies the DOM so may or may not be a good fit (as I don't know it it's hard to tell).

So maybe, as there are other, functional gaps / regressions to cover, and a number of important PR feedback items, it could also be a separate PR that would also involve `parcoords` which also has dragging.

src/traces/table/plot.js
monfera: Neat, will check!

src/traces/table/plot.js
monfera: Agree with your sentiment... I looked into these scrollers and didn't find an easy way for either using them or reworking them to cover the `table` use case, it's an infinite revolver scroller, our discussion: https://github.com/plotly/plotly.js/pull/1834#issuecomment-313725969

As the scrollbars and event captures didn't take much time and there were more difficult issues, I spent < 1 day putting in the scrollbars remembering that maybe you or us would check how unification could be done. I'd have spent a lot more time trying to unify while the core of `table` also wasn't stable so it was a risk reduction tradeoff (maybe a bad tradeoff on my part).

src/traces/table/plot.js
etpinard: No need to switch to `dragElement` in this PR. But we should really start centrailizing all drag handlers soon.

src/traces/table/plot.js
etpinard: That :rage: wasn't meant for you. Adding a separate scroller for `table` at this stage was definitively the correct decision.

It's just one big tech-debt that we'll have to pay back soon (I hope) :moneybag:

src/traces/table/plot.js
monfera: Agreed, it should be extracted out. I scrambled during the implementation, as the significant code size of `svg_text_utils`, the previously unknown (to me) MathJax, the need for doing asynchronous cell content population and bumping row heights and revolver panel offsets on taller content yielded quite a few challenges. 

Maybe I'd need to consider alternative work processes - and glad to receive input on this - but it felt more manageable to work out the kinks while not risking other plot types, and then unify them once it proved itself. You were kind to offer developing some parts of `table` and I think we even talked about specific items (scrolling etc.), maybe my avoidance of bothering you w/ these wasn't ideal.

alexcjohnson: > I really wished the text-wrapping logic would've been added to svg_text_utils.js so that other parts of plotly.js can use it.

That would certainly be the great, and it will have to happen at some point. I feel like a lot of the complexity here has to do with:
- figuring out whether we're allowed to apply formatting or not - which, you're right, doesn't belong in `svg_text_utils`, but I feel like it would be simpler and more robust to *try* to apply formatting if the user has specified formatting, then back off (and just cast as string?) if it didn't work, rather than trying to determine ahead of time whether formatting is allowed. Then perhaps we wouldn't need all the `typeof` checks and the confusing `userBrokenText` and `userBrokenText2`?
- trying to figure out if we can avoid `convertToTspans` altogether in favor of [this block](https://github.com/plotly/plotly.js/pull/2052/files?diff=unified#diff-1e0dea3a9361823fe80cbbdc50befd7eR509), I guess for performance purposes? But if shortcuts like that exist to improve over the performance of `convertToTspans`, shouldn't we use those shortcuts *in* `convertToTspans` so everyone can benefit?

If this section *does* need to have this kind of logic, I feel like it could be made a lot clearer with `if` branches instead of ternary operators.

src/traces/table/plot.js
etpinard: Ok, thanks for the info.

src/traces/table/plot.js
alexcjohnson: The key test here (which I assume will fail given the `getBoundingClientRect` call) is to call `Plotly.newPlot` into an element that's not in the DOM, and insert it in the DOM afterward - ala https://codepen.io/anon/pen/XjakQy
If you really *do* need a `getBoundingClientRect` result, use `Drawing.bBox`.

src/traces/table/plot.js
monfera: Good plan, totally warranted! I feel we're pushing the perf boundaries with new plot types, e.g. `parcoords` and `table` which must support an essentially unlimited row count (plain content tested with 1m rows and 20 columns just fine; Bird mock is just 7k for saving space) as well as both doing asynchronous rendering. So I think it's fitting that unification happens once we have diverse experience from all use cases, eg. small legend panels and crazy high performance alike (our preexisting things don't use revolvers etc.)

src/traces/table/plot.js
alexcjohnson: in regular `convertToTspans` MathJax gets colored [here](https://github.com/plotly/plotly.js/blob/master/src/lib/svg_text_utils.js#L118) based on the color that was already applied to the text it was converted from - does that not apply here?

src/traces/table/plot.js
monfera: @alexcjohnson yes the naming of certain vars in the determination of which code paths to hit would benefit from improvement; I still wanted to spend a bit of time cleaning up that section, also, logic-wise (and we could save some CPU cycles, some `regex` is called twice).

`convertToTspans` is long code that does much heavy lifting and my perf measurements showed it to be a main bottleneck, which is why it's avoided when possible (this code split was a reaction to benchmarking). I think that most often, users don't have a simultaneous need for cell content formatting _and_ actual infinity scrolling with no jank. So, as you can see in the Bird example, scrolling is pretty smooth with plain content (we have more headroom if need be) even though it applies `D3` numerical formatting, pre/postfix, alignment.

I agree with the suggestion that such performance-minded shortcuts may benefit all callers of `convertToTspans` and it boils down to two things:

1. Development time, which is dominated by having to ensure, also with manual testing, that nothing breaks; also, getting more intimate in `convertToTspan` as it's also imperative code that needs time (I found that doing two hard things at the same time is non-linear, risk-wise and I was already at my comfort level). 
2. Is there a current or foreseen need? Saving CPU cycles is always a good idea, though in a lot of cases, plots have but a handful or maybe a couple of dozen text elements, and even if more, often, there's no need to support their redrawing with a high frame rate as in a scroller. `table` is quite special in that it's both absolutely text heavy and benefits from very fast renders, otherwise scroll is janky.

One more note about `convertToTspan`, which generalizes to our other functions such as `Drawing.font` (already in use in `table`) and other things that @etpinard suggested in this review such as `dragElement`, other scrollers etc. Basically all these things couple computational tasks with rendering tasks. Moreover, rendering is in-place and often singleton, ie. requires a `gd` object, builds on a number of assumptions which were naturally true when making them or extracting them out. It sounds abstract so here's an example:

LaTeX rendering has high latency. I started coding toward saving previously shown cells so that, if the user has already scrolled there, there would be no need for a fresh call to `convertToTspan`. But the way it works is, it _computes_ what to put in the DOM and then _renders_ it too. So I can't just cache the results and reapply it anytime. Faced with the choice of querying and saving DOM fragments, and subsequently reapplying them, I thought let's find some other way. Maybe the best place to do caching is in `convertToTspan` itself but this component may not have enough info on when to release (memory leak). What would yield more flexibility is,

1. the component would return _something_
2. that _something_ could be manipulated by the caller who has info on memory use, eg. it can bind memory to DOM elements which go away after not needed (GC) or the trace can use some MRU/LRU cache component as it befits

The _something_ could be 

- a HTML `innerText`/`outerText` (obv. limited as it can't do just styling like mixing color)
- a function that takes a DOM node and alters or recreates desired structures
- a virtual DOM fragment

I haven't thought much about it but some small vDOM utility like Preact (3kB) would do the job (adding React would be too much payload for a task like this and Preact API is almost identical).

In general, I wouldn't mind eventually breaking up these utilities and uncoupling calculations (-> pure functions) and rendering (which, if we want, can also be represented declaratively, eg. via some `diff` or `ensure` patch not unlike a vDOM object, or if not, it can be a returned function that takes a DOM node as input).

src/traces/table/plot.js
monfera: Yes, although I haven't yet tested it and maybe it'll need an `.exit().remove()` to `yColumn` if not done yet, so let me check.

src/traces/table/plot.js
monfera: I'm meek 😺 Also, 'convertToTspans' creates temporary DOM nodes at the `document.body` level that get absorbed upon completion and there may be other lifecycle issues to wrap up, lest I add memory leaks

src/traces/table/plot.js
monfera: `gd` is only needed (and passed on by the caller) if there's a possible need for revolver panel rerender. If the user scrolls down, then maybe one panel jumps ahead of the other, implying that its contents must be rendered. But if this function merely gets called due to a row becoming taller (eg. to fit LaTeX or wrapped text content) then it just bumps the block Y positions but doesn't rerender the block contents.

Should I split this function to two parts?

src/traces/table/plot.js
monfera: I'll check your link, maybe it solves the need for the current hack, I wasn't happy it drew in black when `fill` was white but thought maybe there's a better way.

test/jasmine/tests/table_test.js
monfera: Yes, vestigial. I'll read through the code before the next round.

test/jasmine/tests/table_test.js
monfera: Basically I didn't intend to check in `table_test.js` 😃

monfera:
@etpinard I can taper down the bouncy column drag to less or non-bouncy easing.

monfera:
@alexcjohnson @rreusser is it decided that plot state will be fully specified? That'd include things like zoom level, pan position, the use of a selection box or even the contours of a lasso. Not suggesting it's not a desired thing at all though!

@rreusser I'm not sure if it adds to or questions anything you said, but wanted to mention that controlling scroll position _might be_ desirable. For example, some journal incorporates  a `table` and links scroll position to an advance-based story arc, or part of some product tour. I think it'd apply to other things as well. I prefer observables over fully restating the state because it allows lightweight, minimal updates upon very specific signals, ie. you save on both spec diffing and incremental rerendering. Such signals may be a lot of things besides scroll position; could be for responsive features such as `devicePixelRatio`, `window`/`document` or container element size, or even things like mouse position (for hover).

rreusser:
@monfera That's valid. I was wondering about that. I guess the key part is that we somehow feeding UI scrolling from feeding back into controlled UI updates.

src/traces/table/plot.js
alexcjohnson: Folks will definitely want to copy text. I wouldn't worry about LaTeX, as long as it's MathJax it's just paths anyway.

src/traces/table/plot.js
alexcjohnson: probably just an order-of-operations issue, we assume the text is fully styled before being converted to MathJax. It would probably be good for us to break this out like we did with `positionText` so it can be applied either before or after conversion.

monfera: Ah, thanks for the lead!

src/traces/table/plot.js
alexcjohnson: > Development time, which is dominated by having to ensure, also with manual testing, that nothing breaks; also, getting more intimate in `convertToTspan` as it's also imperative code that needs time (I found that doing two hard things at the same time is non-linear, risk-wise and I was already at my comfort level).

That's fair, I know `convertToTspans` is a cumbersome chunk of code even after several attempts to clean it up over the years. We may also want to order our pushes wisely - for example perhaps we explore KaTeX first, then merge wrapping in. Please make an issue to tackle this though. There shouldn't be too much manual testing needed, between images and jasmine we have fairly good coverage of `convertToTspans`

> Is there a current or foreseen need?

The biggest need is for functionality (wrapping) rather than perf, though the latter certainly wouldn't hurt. You're right that table is the heaviest user of text, but there are a couple of cases that can get pretty heavy during dynamic zoom/pan:
- tick labels, particularly categories
- scatter text labels

> In general, I wouldn't mind eventually breaking up these utilities and uncoupling calculations (-> pure functions) and rendering

That sounds like a good general strategy, my concern is mainly about a proliferation of different patterns scattered about the codebase, with no clear plan to reunify them later. That's the reason for the fairly strong preference to use what's present and either fix or extend it to fit new needs, rather than augmenting it from the outside.

alexcjohnson:
> is it decided that plot state will be fully specified? That'd include things like zoom level, pan position, the use of a selection box or even the contours of a lasso.

Yes. Specify all the things.

> I guess the key part is that we somehow avoid feeding UI scrolling from feeding back into controlled UI updates.

We can certainly throttle these events (wooo we have a [util](https://github.com/plotly/plotly.js/blob/master/src/lib/throttle.js) for that 🎉 ) but we definitely want users to be able to watch the scroll state, and for example highlight points on a separate plot corresponding to the visible rows, and vice versa - click a point and scroll to that row in the table (hmm, those are a little more involved, as they require not only the pixel scroll, which is required to fully specify the state, but also some way to translate that bidirectionally to row numbers. Anyway my more immediate concern is other edits not resetting the table scroll - like the issues @etpinard fixed for geo plots in #2030.

alexcjohnson:
Looks to me like `shape-rendering: crispEdges` is getting inherited in a few places it shouldn't: the scroll handle and MathJax are where I notice it.

src/traces/table/plot.js
monfera: Added the requested separate issue: https://github.com/plotly/plotly.js/issues/2053

src/traces/table/plot.js
monfera: Added this item too to https://github.com/plotly/plotly.js/issues/2053

src/traces/table/plot.js
monfera: Separate item: https://github.com/plotly/plotly.js/issues/2054

src/traces/table/plot.js
monfera: I think that https://github.com/plotly/plotly.js/issues/2053 should address it. I added a 'core' line now to express this. Here, the difference between parent and child bounding box top matters. What do people, or we, do with detached nodes? I'm trying to minimize open points for this specific PR, unless of course users would run into serious issues with it otherwise.

src/lib/gup.js
monfera: Added note to https://github.com/plotly/plotly.js/issues/2058

src/traces/table/plot.js
monfera: Included it in https://github.com/plotly/plotly.js/issues/2056

src/traces/table/plot.js
monfera: Issue https://github.com/plotly/plotly.js/issues/2057 created for `dragElement` related unification

monfera:
@etpinard regarding editability: I've coded with eventual editing in mind, for example, async cell loading is effectively, post-render editing :-) But there are bound to be questions. We'd need some SVG text cursor handling; maybe undo/redo. If there's cell editing, it'd be logical to support other things such as line addition/deletion. Adding an item into https://github.com/plotly/plotly.js/issues/2056

src/traces/table/attributes.js
```diff
@@ -67,9 +67,9 @@ module.exports = {
             role: 'info',
             dflt: [],
             description: [
-                'Dimension values. `values[n]` represents the value of the `n`th point in the dataset,',
-                'therefore the `values` vector for all dimensions must be the same (longer vectors',
-                'will be truncated). Each value must be a finite number.'
+                'Header cell values. `values[m][n]` represents the value of the `n`th point in column `m`,',
```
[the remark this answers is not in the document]
monfera: Currently, even header rows must be 2D. It'd make sense to allow the common 1D case, though adding it later doesn't break existing JSONs. It's 2D in small part due to code uniformity - header and content are handled by the same code path except stuff like no scrolling for header rows and acting as column drags.

But the more important part was a previous PR comment suggesting it, e.g. to have a 2nd row for units of measure.

A `<br>` would often but not always suffice. For example, top cell may be left-aligned text and bottom cell may be right-aligned LaTeX with another font size. Also, multirow header is implemented (took a bit of time) and causes no issues. 

One more reason for this common-path implementation style was to eventually support other row groups, e.g. aggregate rows but that's just a concept now.

src/traces/table/plot.js
monfera: @alexcjohnson @etpinard Fully agree w/ the order of operations (LaTeX coloring), a quick initial look didn't lead to a 5-minute solution, plan to do it tomorrow. Other tasks are, 

- ensuring `table` works fine with `Plotly.restyle`, and 
- more `default`ing shouls happens (eg. now the user always needs to specify column order) 

which seem more pressing than internal changes.

src/traces/table/attributes.js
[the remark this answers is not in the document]
etpinard: Ok. But please add a _allow 1D header values_ items to https://github.com/plotly/plotly.js/issues/2056

src/traces/table/plot.js
etpinard: Ok. Not a great pattern, but no action required for now.

src/traces/table/plot.js
etpinard: Ok. Please add item to https://github.com/plotly/plotly.js/issues/2056

etpinard:
@monfera here's a list of open items:

### :no_entry: Blocking items

- Make all mocks pure JSON !!!
- Add baselines !!!
- Make tests :white_check_mark: on CI !!!
- Remove all `clipView` dev-only blocks
- Remove all tests case that were just copied over from `parcoords` 
- Make cell text copy-able (i.e. remove `pointer-events: 'none'`) https://github.com/plotly/plotly.js/pull/2052/files#r142486070
- Use `Drawing.setClipUrl` https://github.com/plotly/plotly.js/pull/2052#discussion_r142415932
- Use `Color.stroke` https://github.com/plotly/plotly.js/pull/2052/files#r142419913

### :tropical_drink: Items that shouldn't take too long, but can wait for https://github.com/plotly/plotly.js/issues/2056

- Support `header.format` scalar value https://github.com/plotly/plotly.js/pull/2052/files#r142410512
- Reuse `plots/font_attributes.js` https://github.com/plotly/plotly.js/pull/2052/files#r142410981
- Move `<clipPath>` to `fullLayout._clips` https://github.com/plotly/plotly.js/pull/2052/files#r142418318

### :leaves: Items that you should add to #2056 

- Support for 1D header values https://github.com/plotly/plotly.js/pull/2052#discussion_r142827253
- Better handling of MathJax styling https://github.com/plotly/plotly.js/pull/2052#discussion_r142469257

etpinard:
Oh god, `toImage` doesn't work. @monfera is this a known issue? Or some regression that creep in https://codepen.io/monfera/pen/aLVEOK ?

monfera:
@etpinard `toImage` should work fine except the codepen version has rerender temporarily disabled, to be able to show subplots with scatter with no rerendering on scatter interaction. Here's the image from the feature branch (code state as per this PR):

image

etpinard:
image

This sums up this PR. A very nice looking and performant trace type. :clap: to @monfera for his very user-first approach. But, unfortunately, not much work was put in for reusability for our other components.

I hope @monfera will have time to make a followup PR knocking out a few items in https://github.com/plotly/plotly.js/issues/2056 in the next few days.  Factoring out the text wrap logic ( https://github.com/plotly/plotly.js/issues/2053 ) we'll make a bunch of folks happy, I hope this happens soon. Items https://github.com/plotly/plotly.js/issues/2057 and https://github.com/plotly/plotly.js/issues/2054 are less important to me in the short-term.

Let's get this thing in `master` :dancer: :dancer:

monfera:
Thanks @etpinard and @alexcjohnson for the numerous review points!

Issue #2044, opened 2017-09-29
scattergeo hoverinfo breaks when it's an array


cpsievert:
https://codepen.io/cpsievert/pen/NagamG

Discovered via https://github.com/ropensci/plotly/issues/1127

etpinard:
Should be an easy fix. I'll try to include in `1.31.0`.

etpinard:
Oh wait, you have `hoverinfo` as an array (not `hovertext` like a I was expected - I must be going blind :eyeglasses: ). That's not _supposed_ to work. 

`hoverinfo` is flaglist attribute. Setting it to `"lon+lat"` gives the desired behavior:

https://codepen.io/etpinard/pen/veZRLw?editors=0010

cpsievert:
Why, then, would `arrayOk` be `true`?

![screen shot 2017-09-29 at 1 57 43 pm](https://user-images.githubusercontent.com/1365941/31031465-3cba5778-a51e-11e7-894e-60b42f59f9cf.png)

etpinard:
Oh sorry, I'm not going blind, but I might have Alzheimer's disease.

I forgot about https://github.com/plotly/plotly.js/pull/1761 - which didn't bother testing any geo traces. On it.

PR #2055, opened 2017-10-03
ArrayOk hoverinfo fixups

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2044 in 0a77239, but really, these few commits here should have been part of https://github.com/plotly/plotly.js/pull/1761. Adding anything hover-related is still a bit of a pain right now. As a lot of logic is scattered around `components/fx`, the base plot and trace modules, things don't _just work_ unfortunately. Manually checking every trace type is in order until we better centralize our hover label logic.

This PR also add a fairly general custom assertion in 4cf489e that checks every part of the hover label DOM nodes. I also caught a few others bug fixed and :lock: in b24759d and 85d7061.

cc @alexcjohnson

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -443,12 +443,10 @@ drawing.tryColorscale = function(marker, prefix) {
 var TEXTOFFSETSIGN = {start: 1, end: -1, middle: 0, bottom: 1, top: -1};
 drawing.textPointStyle = function(s, trace, gd) {
     s.each(function(d) {
-        var p = d3.select(this),
-            text = d.tx || trace.text;
+        var p = d3.select(this);
+        var text = Lib.extractOption(d, trace, 'tx', 'text');
 
-        if(!text || Array.isArray(text)) {
-            // isArray test handles the case of (intentionally) missing
-            // or empty text within a text array
+        if(!text) {
             p.remove();
             return;
         }
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -887,6 +887,9 @@ drawing.setTextPointsScale = function(selection, xScale, yScale) {
         var transforms;
         var el = d3.select(this);
         var text = el.select('text');
+
+        if(!text.node()) return;
+
         var x = parseFloat(text.attr('x') || 0);
         var y = parseFloat(text.attr('y') || 0);
 
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -447,8 +447,7 @@ lib.castOption = function(trace, ptNumber, astr, fn) {
  *  @return {any}
  */
 lib.extractOption = function(calcPt, trace, calcKey, traceKey) {
-    var calcVal = calcPt[calcKey];
-    if(calcVal || calcVal === 0) return calcVal;
+    if(calcKey in calcPt) return calcPt[calcKey];
 
     // fallback to trace value,
     //   must check if value isn't itself an array
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -437,6 +437,27 @@ lib.castOption = function(trace, ptNumber, astr, fn) {
     }
 };
 
+/** Extract option from calcdata item, correctly falling back to
+ *  trace value if not found.
+ *
+ *  @param {object} calcPt : calcdata[i][j] item
+ *  @param {object} trace : (full) trace object
+ *  @param {string} calcKey : calcdata key
+ *  @param {string} traceKey : aka trace attribute string
+ *  @return {any}
+ */
+lib.extractOption = function(calcPt, trace, calcKey, traceKey) {
+    var calcVal = calcPt[calcKey];
+    if(calcVal || calcVal === 0) return calcVal;
+
+    // fallback to trace value,
+    //   must check if value isn't itself an array
+    //   which means the trace attribute has a corresponding
+    //   calcdata key, but its value is falsy
+    var traceVal = lib.nestedProperty(trace, traceKey).get();
+    if(!Array.isArray(traceVal)) return traceVal;
+};
+
 /** Returns target as set by 'target' transform attribute
  *
  * @param {object} trace : full trace object
--- a/src/traces/choropleth/hover.js
+++ b/src/traces/choropleth/hover.js
@@ -9,6 +9,7 @@
 
 'use strict';
 
+var Lib = require('../../lib');
 var Axes = require('../../plots/cartesian/axes');
 var attributes = require('./attributes');
 
@@ -80,12 +81,8 @@ function makeHoverInfo(pointData, trace, pt, axis) {
 
     if(hasZ) text.push(formatter(pt.z));
     if(hasText) {
-        var tx;
-
-        if(pt.tx) tx = pt.tx;
-        else if(trace.text) tx = trace.text;
-
-        if(!Array.isArray(tx)) text.push(tx);
+        var tx = Lib.extractOption(pt, trace, 'tx', 'text');
+        if(tx && tx !== '') text.push(tx);
     }
 
     pointData.extraText = text.join('<br>');
--- a/src/traces/choropleth/hover.js
+++ b/src/traces/choropleth/hover.js
@@ -9,9 +9,9 @@
 
 'use strict';
 
-var Lib = require('../../lib');
 var Axes = require('../../plots/cartesian/axes');
 var attributes = require('./attributes');
+var fillHoverText = require('../scatter/fill_hover_text');
 
 module.exports = function hoverPoints(pointData, xval, yval) {
     var cd = pointData.cd;
@@ -81,8 +81,7 @@ function makeHoverInfo(pointData, trace, pt, axis) {
 
     if(hasZ) text.push(formatter(pt.z));
     if(hasText) {
-        var tx = Lib.extractOption(pt, trace, 'tx', 'text');
-        if(tx && tx !== '') text.push(tx);
+        fillHoverText(pt, trace, text);
     }
 
     pointData.extraText = text.join('<br>');
--- a/src/traces/choropleth/hover.js
+++ b/src/traces/choropleth/hover.js
@@ -53,17 +53,17 @@ module.exports = function hoverPoints(pointData, xval, yval) {
 };
 
 function makeHoverInfo(pointData, trace, pt, axis) {
-    var hoverinfo = trace.hoverinfo;
+    var hoverinfo = pt.hi || trace.hoverinfo;
 
     var parts = (hoverinfo === 'all') ?
         attributes.hoverinfo.flags :
         hoverinfo.split('+');
 
-    var hasName = (parts.indexOf('name') !== -1),
-        hasLocation = (parts.indexOf('location') !== -1),
-        hasZ = (parts.indexOf('z') !== -1),
-        hasText = (parts.indexOf('text') !== -1),
-        hasIdAsNameLabel = !hasName && hasLocation;
+    var hasName = (parts.indexOf('name') !== -1);
+    var hasLocation = (parts.indexOf('location') !== -1);
+    var hasZ = (parts.indexOf('z') !== -1);
+    var hasText = (parts.indexOf('text') !== -1);
+    var hasIdAsNameLabel = !hasName && hasLocation;
 
     var text = [];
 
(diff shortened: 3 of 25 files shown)


src/traces/scattercarpet/hover.js
```diff
@@ -67,7 +67,7 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     var tj = ij[1] - j0;
 
     var xy = carpet.evalxy([], i0, j0, ti, tj);
-    text.push('y: ' + xy[1].toFixed(3));
+    text.push('y = ' + xy[1].toFixed(3));
```
etpinard: cc @rreusser unless that `y:` was on purpose? This here turns it into `y =` just like the `a =` and `b =` rows

peek 2017-10-03 17-01

alexcjohnson: Most other places (heatmap/contour, 3d, ternary) we use `:` - perhaps we should standardize on that here too?

rreusser: Good catch. Thank you. That seems correct.

rreusser: Oh. Haha, my mistake then. Perhaps it's the `a/b = ` that are the odd ones out. Need to check the consistent choice here.

rreusser: https://github.com/plotly/plotly.js/blob/master/src/components/fx/hover.js#L697-L698

src/traces/choropleth/hover.js
```diff
@@ -79,7 +79,14 @@ function makeHoverInfo(pointData, trace, pt, axis) {
...
 
     if(hasZ) text.push(formatter(pt.z));
-    if(hasText) text.push(pt.tx);
+    if(hasText) {
+        var tx;
+
+        if(pt.tx) tx = pt.tx;
+        else if(trace.text) tx = trace.text;
+
+        if(!Array.isArray(tx)) text.push(tx);
```
alexcjohnson: nice careful logic - `tx` could be an array if `trace.text` is too short to populate `pt.tx` or if it has falsey elements (hmm, what about `0`?)

This seems like a bit of logic that's repeated a bunch - and is probably done wrong in some places where it's *not* repeated... worth abstracting?

etpinard: Worth it for sure. Shouldn't be too bad.

test/jasmine/assets/custom_assertions.js
```diff
@@ -59,6 +59,92 @@ exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
     expect(textStyle.fill).toBe(expectation.fontColor, msg + ': font.color');
 };
 
+function assertLabelContent(label, expectation, msg) {
+    var lines = label.selectAll('tspan');
```
alexcjohnson: in case we want to test styled labels at some point, `selectAll('tspan.line')`

etpinard: Good eyes. Thanks!

src/traces/choropleth/hover.js
etpinard: (the hardest part will be to figure out how to call this new function)

test/jasmine/assets/custom_assertions.js
```diff
@@ -59,6 +59,92 @@ exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
...
+            fail('Expected a multi-line label, found single');
+        }
+    }
+}
+
+function count(selector) {
+    return d3.selectAll(selector).size();
+}
+
+exports.assertHoverLabelContent = function(expectation, msg) {
```
alexcjohnson: I like this fn - really nice idea! I find the signature a little unintuitive though... `expectation` has to be something like `[['a = 0.200', 'b = 3.500', 'y = 2.900'], 'a = 0.2']`... perhaps it would be clearer like `{nums: 'a = 0.200\nb = 3.500\ny = 2.900', name: 'a = 0.2'}`? Array vs multiline string I don't feel that strongly about, but I think named items instead of the outer array would be quite a bit easier to read, and possibly easier to write too.

src/traces/scattercarpet/hover.js
rreusser: Sorry, to be less vague, `:` seems correct. `=` seems incorrect.

test/jasmine/assets/custom_assertions.js
etpinard: I get annoyed of typing object keys in assertion functions arguments when writing tests.

But I guess there's value in being a little more verbose for project-wide custom assertions.

etpinard: ... and good call about the somewhat useless single-vs-multi line logic. Our `convertToTspans` tests are there for that.

src/traces/scattercarpet/hover.js
etpinard: done in 2fe641d

src/traces/choropleth/hover.js
etpinard: See bc8294d and b637179 for what I came up with.

test/jasmine/assets/custom_assertions.js
etpinard: done in 294714b

test/jasmine/assets/custom_assertions.js
etpinard: done in 294714b

Thoughts?

src/lib/index.js
```diff
@@ -437,6 +437,27 @@ lib.castOption = function(trace, ptNumber, astr, fn) {
...
+lib.extractOption = function(calcPt, trace, calcKey, traceKey) {
+    var calcVal = calcPt[calcKey];
+    if(calcVal || calcVal === 0) return calcVal;
+
+    // fallback to trace value,
+    //   must check if value isn't itself an array
+    //   which means the trace attribute has a corresponding
+    //   calcdata key, but its value is falsy
+    var traceVal = lib.nestedProperty(trace, traceKey).get();
+    if(!Array.isArray(traceVal)) return traceVal;
```
alexcjohnson: Nice, I like it. Thanks for accepting `0`.

> not super happy with the name

meh, it's a very specific concept that we probably don't want to go into any more detail with in the name. Seems fine to me.

> not super happy about it being on Lib

It definitely belongs next to `castOption` but yeah, if we make a `common/` that might be more appropriate for these.

Also a style question: I just noticed all over this file we start the docs immediately after `/**`, I thought the convention was to leave that first line blank and start on the next line?

src/traces/scatter/fill_hover_text.js
```diff
@@ -0,0 +1,36 @@
...
+    var fill = Array.isArray(contOut) ?
+        function(v) { contOut.push(v); } :
+        function(v) { contOut.text = v; };
+
+    var htx = Lib.extractOption(calcPt, trace, 'htx', 'hovertext');
+    if(htx && htx !== '') return fill(htx);
+
+    var tx = Lib.extractOption(calcPt, trace, 'tx', 'text');
+    if(tx && tx !== '') return fill(tx);
+};
```
alexcjohnson: Nice! 🌴  Does this really belong in `scatter` though? Perhaps in `fx`?

src/lib/index.js
etpinard: image

from http://usejsdoc.org/howto-commonjs-modules.html

So, I think both are acceptable.

src/traces/scatter/fill_hover_text.js
etpinard: I have no strong opinion here. This routine felt similar to 

https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/arrays_to_calcdata.js

that's why I put it in `traces/scatter/`

test/jasmine/assets/custom_assertions.js
```diff
@@ -60,88 +60,99 @@ exports.assertHoverLabelStyle = function(g, expectation, msg, textSelector) {
...
 function count(selector) {
     return d3.selectAll(selector).size();
 }
 
+/**
+ * @param {object} expectation
+ *  - nums {string || array of strings}
+ *  - name {string || array of strings}
+ *  - axis {string}
+ * @param {string} msg
```
alexcjohnson: Thanks - I find that a lot easier, especially as you point out since this lives in a different file from where it gets called.

src/traces/choropleth/hover.js
etpinard: Ooops I forget to update `choropleth/hover.js`, now done in -> 7f2604a6987925e0f211fe1f35b0821f0832dafb

src/lib/index.js
alexcjohnson: perhaps overly pedantic, but I read that as `/** Single-line description */` or
```
/**
 * Multiline still looks like it always
 * leaves the first line blank.
 */
```

src/traces/scatter/fill_hover_text.js
alexcjohnson: That's fine, we certainly have a long history of traces pulling from each other, particularly from scatter, and there *is* something nice about it being in the `traces` dir. Perhaps at some point we introduce a `traces/common` dir for stuff like this but leave it as is for now.

src/traces/choropleth/hover.js
```diff
@@ -80,12 +81,8 @@ function makeHoverInfo(pointData, trace, pt, axis) {
...
     if(hasZ) text.push(formatter(pt.z));
     if(hasText) {
-        var tx;
-
-        if(pt.tx) tx = pt.tx;
-        else if(trace.text) tx = trace.text;
-
-        if(!Array.isArray(tx)) text.push(tx);
+        var tx = Lib.extractOption(pt, trace, 'tx', 'text');
+        if(tx && tx !== '') text.push(tx);
```
alexcjohnson: `''` is already falsey, right?

alexcjohnson:
Take a look at https://github.com/plotly/plotly.js/pull/2055/commits/7f2604a6987925e0f211fe1f35b0821f0832dafb#r142793932 but other than that I'm happy - patching a functionality hole while DRYing up a bunch of things -> 💃 !

src/traces/choropleth/hover.js
etpinard: Good catch. Strictly speaking, I think this should be:

```js
if(tx !== undefined && tx !== '') {
  text.push(tx);
}
```

as `Lib.extractOption` returns `undefined` for all falsy values except `0` in calcdata. But, `trace.text` is allowed to be `''` (which is the default by the way).

---

To make this easier to maintain (and remember :slightly_smiling_face: ), I'm thinking either 

- (1) we could remove the `dflt: ''` line in the `text` (and `hovertext`) attribute declarations. This leads to a slight change in behavior (but might more consistent) as now:

```js
{
   text: 'on-chart text',
   hovertext: ''
}
```

would use `''` for the hover label text field (i.e. not show anything). On `master`, `hovertext: ''` is ignored and `'on-chart text'` shows in the hover labels.

- (2) we could make `Lib.extractOption` return `undefined` when `trace.text: ''`, making it a little more opinionated (I think this only makes sense for string attributes).

src/components/drawing/index.js
```diff
@@ -887,6 +887,9 @@ drawing.setTextPointsScale = function(selection, xScale, yScale) {
         var transforms;
         var el = d3.select(this);
         var text = el.select('text');
+
+        if(!text.node()) return;
```
etpinard: Caught another bug. `dragmode: 'pan'` is broken on http://rickyreusser.com/plotly-mock-viewer/#text_chart_arrays 

This fixes it:

peek 2017-10-04 18-24

alexcjohnson: ha, good catch! 🎉

src/traces/choropleth/hover.js
etpinard: Here's what I came up with https://github.com/plotly/plotly.js/pull/2055/commits/3931273b331971085d39ce5bbb1fa7e9d4720c38

I made `Lib.extractOption` less opinionated, and made `scatter/fill_hover_text` handle falsy but valid text values.

alexcjohnson: Yep, I like it. There are some edge cases this makes a little weird - eg trying to make one point have text but no hovertext, while all the others have both - but there are ways to get what you want (in that example using `hoverinfo` as an array) and I don't see any way to make *everything* more intuitive. So lets keep this, it's nice clean logic, will make sense to people, and will do what people expect 99.9% of the time.

Issue #1660, opened 2017-05-10
y-range and extreme outliers: traces vanish progressively


czellweg:
When enabling `yaxis.range` and the data contains extreme outliers, the traces vanish, i.e. they will progressively vanish. See the following codepens that demonstrates this behaviour:

* http://codepen.io/czellweg/pen/rmJvzW (OK)
* http://codepen.io/czellweg/pen/pPaVWj (trace with outlier shown only as markers, no more lines)
* http://codepen.io/czellweg/pen/MmQGEy (all traces gone)

It seems that it depends on the magnitude the outlier is outside the `yaxis.range` values, e.g. if `yaxis.range=[0,20]` and the outlier is 20'000'000, only markers are shown. If `yaxis.range=[0,20]` and the outlier is 2'000'000'000, all traces are gone.

Expected behaviour would be:

* traces are shown, including markers and lines if `markers+lines` has been configured

etpinard:
Thanks very much for this report.

Looks like your pens are working ok using `v1.7.0` -> http://codepen.io/etpinard/pen/gWvjBv

I suspect the same issue as in https://github.com/plotly/plotly.js/issues/403 caused by https://github.com/plotly/plotly.js/pull/353 which unfortunately is a little hard to solve as mentioned in https://github.com/plotly/plotly.js/issues/403#issuecomment-247362096

czellweg:
Ok, good to know. `v1.7.0` is a bit far back (March 22, 2016) but I will follow the developments on the mentioned issues. Thanks.

alexcjohnson:
closed by #2060

PR #353, opened 2016-03-23
Correctly compute ultra-zoomed line paths [fixes #217]

etpinard:
@alexcjohnson @mdtusz 

Issue https://github.com/plotly/plotly.js/issues/217 is two months old now and after numerous attempts at fixing it the *right way* by adding more complexity to the already-complex [line decimation](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/line_points.js) algorithm, I gave up.

This PR removes the clip-multiple constrain in the linear to pixel convert routine. This has three effects:

1. Ultra-zoomed line paths (where 4 segment pts are beyond the clip-multiple constrain) are now computed correctly, fixing #217.
1. Far-off-edge points now have coordinates of very large magnitude that some browser's svg renderers **may not** understand. Note: I couldn't find any documentation on the svg limits.
1. The line decimation algorithm is slowed down as now far-off-edge points are not de facto clusters anymore.

I decided to make this PR because I thought that fixing #217 was more important than side-effects 2. and 3. above. I hope you agree.

Issue #217, opened 2016-01-26
Line paths are not correctly computed when their points are well-beyond the axis ranges


ifll:
When building a scatter plot with both x and y logarithmic axis there is an error when placing a line shape on the plot.

The line is sometimes well displayed but when trying to zoom it disappears or appears not well placed.

The code used for displaying the line is the following:

- inside the layout variable for the plot making place the following:

var layout2 = {
			  .
                          .
                          .
			  shapes: [
			           
					//line 45º
					
					{
					  type: 'line',
					  xref: 'x',
					  yref: 'y',
					  x0: 0.0001,
					  y0: 0.0001,
					  x1: 10000,
					  y1: 10000,
					  line: {
					    color: 'rgb(0, 0, 0)',
					    width: 1
					  }
					},
			           
			           
			           
			           ],
     
				  .
				  .
				  .
			};

there are around 1000 points to be displayed on the scatter plot

Do any of you know if this is a bug or if is is not the proper way to place a line on the plot and there is another method which could work?

Thank you.

etpinard:
Thanks for writing in.

Can you elaborate on what you mean by *not well-placed* ?

Looks like log shapes are behavior as expected when no data is present: http://codepen.io/etpinard/pen/NxYyzL

Could you share the data you're using?

ifll:
Unfortunately data I am using is not sharable...

Even though, within the example you have posted in your answer there appears the same problem I have been encountering on my plot, as when you zoom on the plotly plot around the line shape created it disappears (it might be that 2 or 3 zooms are required to check this problem).

Besides this, by non-well displayed I mean what I have explained in the previous paragraph of this answer. Also another effect of "non-well displayed" is that the line shape is displayed not complete but for a small range of its completion, or even that it is displayed not where it should be placed.

Hope this helps to understand the problem I have been encountering and thank you very much for your help.

etpinard:
I can't replicate it.

Could you at least share a screenshot of the *not-well-displayed* line?

etpinard:
Moreover, what browser are you using?

ifll:
Here you can see some images of the problem I have reported and obtained from the link you have posted on your first answer on this chat.

This images are a followup of zooming on the plot created on the link you have posted.

[...]

PR #2060, opened 2017-10-04
Support extreme off-plot data points in scatter lines

### LINKED ISSUE #403: Lines disappear when displaying small numbers and large numbers are off screen
Can see this clearly here: https://plot.ly/~KenCoder/11/percent-vs-date/

Move your mouse in the middle of the chart  - the hover displays the data, but the line isn't visible. The cause seems to be a large value on 10/1/1924. Even though that value is off screen, it's messing up the rest of the points. Pressing auto-scale shows the large point (and the rest of the line) but reset axes reverts the error.

I believe this was introduced in the last few months, as these charts worked for us in a prior version of plotly.js

alexcjohnson:
Fixes #403 and #1660 (which is a duplicate as expected) - when scatter line points go extremely far off the plot, the SVG would break. Turns out both Chrome and FF have limits at +/- 2^29 px (about 500 million), after which the path (and sometimes other paths too) disappears. I didn't check other browsers (except nw, I'll make a note in the code about that).

Basically a brute-force solution: once you get 20 screen widths away (so, max of ~100k px for a very large screen), clip the path to where it intersects that rectangle, then follow it around on that rectangle as it moves outside the rectangle (so that fills will be handled correctly) then revert to a regular path when it passes back inside that rectangle. That way paths inside the +/- 20-screen rectangle are precisely correct, preserving fix #353 of bug #217. One can ponder the effects of this on other line shapes - they all have edge cases in which this alters them, but to my mind it's not worth the extra complexity of dealing with that.

For in-bounds points the overhead is pretty minimal - 6 or 8 extra simple logic operations, quite a bit less than we pick up from clustering. For out-of-bounds points (which hopefully are rare enough that we don't often need to be too worried about their impact) there are a few line intersections to calculate, quite a bunch of annoying logic, and perhaps more importantly one out-of-bounds point can turn into 2 or 3 clipped points in the path, though on the flip side, many out-of-bounds points together generally collapse into just a few perimeter points.

cc @etpinard


test/image/mocks/axes_range_manual.json
```diff
@@ -42,13 +46,18 @@
...
+        -1e11,
         4,
         5,
         6,
         7,
         8
       ],
-      "type": "scatter"
+      "type": "scatter",
+      "mode": "lines"
```
alexcjohnson: nw has problems with displaying the markers for this trace if you include them: you get the axes and gridlines but no traces drawn at all. If you drop the exponent 100 to 12 or less, it works again. Clipping markers should be easier than lines, but I opted to leave that for another time. If it's only nw that cares, perhaps image-exporter will moot the issue - but it bears a little more investigation -> #2061

test/jasmine/assets/custom_matchers.js
```diff
@@ -127,7 +127,7 @@ var matchers = {
                     'to be close to',
                     arrayToStr(expected.map(arrayToStr)),
                     msgExtra
-                ].join(' ');
+                ].join('\n');
```
alexcjohnson: This way the two arrays always start at the same column, so it's easy to match points and see what failed.

test/image/mocks/axes_range_manual.json
alexcjohnson: BTW I modified this mock because, while `ultra_zoom` does test the functionality, I wanted to cover a few more cases with the image tests. But [scatter_test](https://github.com/plotly/plotly.js/pull/2060/files#diff-88806d1982a87e90bee0165cdabd5b9a) covers many many more cases.

etpinard: using the new image exporter and :hocho: that `"mode": "lines",` line, we get:

image

src/traces/scatter/line_points.js
```diff
@@ -10,76 +10,241 @@
...
+        // we may need to loop around a corner point
+        if(pts[pti - 1][0] !== pt[0] && pts[pti - 1][1] !== pt[1]) {
+            updateEdge([lastXEdge, lastYEdge]);
+        }
+        updateEdge(pt);
+        lastFarPt = null;
+        lastXEdge = lastYEdge = 0;
+    }
+
+    function addPt(pt) {
```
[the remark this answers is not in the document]
alexcjohnson: That's correct. And good call on adding some test cases for other `line.shape` values. I actually might want to think about this a little - `spline` is probably going to have to stay untouched, but `[hv]+` may have relatively simple ways to make them behave perfectly correctly, which they don't now because we linearly interpolate between the on-screen and extreme points.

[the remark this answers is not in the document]
alexcjohnson: Spline actually works remarkably well using the linear algorithm - because Catmull-Rom smoothing reaches a steady limit to the curve shape on-screen as the next point moves farther and farther off-screen.

https://github.com/plotly/plotly.js/pull/2060/commits/37b5cdc559788192ea9cfcdabcf5ee51605f7f18 adds special handling for the right-angle line shapes - not too bad, certainly less involved than linear! Test-wise, in the interest of time I only altered ultra-zoom, did not add extra jasmine tests for these, I hope that's OK.

test/jasmine/tests/plots_test.js
```diff
@@ -408,6 +408,19 @@ describe('Test Plots', function() {
         beforeEach(function(done) {
             gd = createGraphDiv();
             Plotly.plot(gd, [{ x: [1, 2, 3], y: [2, 3, 4] }], {}).then(done);
+
+            // hacky: simulate getting stuck with these flags due to an error
+            // see #2055 and commit 6a44a9a - before fixing that error, we would
+            // end up in an inconsistent state that prevented future Plotly.newPlot
+            // because _dragging and _dragged were not cleared by purge.
```
alexcjohnson: That's the first time I'd seen an error that blocked all future `Plotly.newPlot` calls! We really need to encapsulate all these `gd.***` vars better...

src/traces/scatter/line_points.js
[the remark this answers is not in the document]
etpinard: >  I hope that's OK.

Yeah, image test support should suffice here. :+1:

src/plots/plots.js
```diff
@@ -1341,14 +1341,25 @@ plots.purge = function(gd) {
     delete gd._promises;
     delete gd._redrawTimer;
     delete gd.firstscatter;
-    delete gd.hmlumcount;
-    delete gd.hmpixcount;
+    delete gd._hmlumcount;
+    delete gd._hmpixcount;
```
etpinard: good catch!

test/jasmine/tests/plots_test.js
etpinard: Yep, very nice. Thanks!

src/traces/scatter/line_points.js
```diff
@@ -10,76 +10,311 @@
...
+
+        if(ptInt1) out.push(ptInt1);
+        if(ptInt2) out.push(ptInt2);
+        return out;
+    }
+
+    // hvh and vhv we sometimes have to move one of the intersection points
+    // out BEYOND the clipping rect, by a maximum of a factor of 2, so that
+    // the midpoint line is drawn in the right place
+    function getABAEdgeIntersections(dim, limit0, limit1) {
```
etpinard: Nice job making this file very readable, even though the algo is quite heavy :muscle: 

I'm thinking we might want to release this module here as a separate package at some point.

test/image/mocks/ultra_zoom.json
```diff
@@ -1,33 +1,15 @@
...
-                0.0001,
-                10000
-            ]
-        }
+        {"x": [10000, 0.0001], "y": [0.0001, 10000], "fill": "tozeroy", "fillcolor": "blue"},
+        {"x": [10000, 0.0001], "y": [0.0001, 10000], "fill": "tozeroy", "line": {"shape": "hvh"}},
+        {"x": [10000, 0.0001], "y": [0.0001, 10000], "fill": "tozeroy", "line": {"shape": "vhv"}},
+        {"x": [0, 5001], "y": [5002, 0], "line": {"shape": "hv"}},
+        {"x": [5002, 0], "y": [0, 5003], "line": {"shape": "vh"}},
+        {"x": [0, 5000, 5001, 5002, 5003], "y": [-5000, 5000, 5000, 5000, 15000], "line": {"shape": "spline"}}
```
etpinard: Solid mock :lock:

etpinard:
A very impressive PR. `line_points.js` is definitively in the running for plotly's most elegant algorithm :trophy: 

:dancer: :dancer: :dancer: 

--- 

Don't forget to close https://github.com/plotly/plotly.js/issues/1660 too (I don't think Github is smart enough to auto-close multiple issues)

PR #2064, opened 2017-10-06
Fix for table scroll glitch

monfera:
The glitch sometimes occurs when manually stopping a fast mouse wheel scroll in a table with text wrapping that increases cell height - thanks @rreusser for the report!


etpinard:
:dancer:

PR #2065, opened 2017-10-06
Roll out General Update Pattern utils to `parcoords` and `sankey`

monfera:
More traces to use these simple utils that yield a simple and robust pattern for constructing the scene graph (and wrapping/unwrapping data). I didn't prefix the functions as character count loss wouldn't have improved, so it was a drop-in replacement of preexisting, separate function defs.


etpinard:
:dancer:

Issue #2056, opened 2017-10-03
Open items for future PRs related primarily to `table`


monfera:
- [ ] add event callbacks to `table` https://github.com/plotly/plotly.js/pull/2052#discussion_r142450559 including plotly_hover and plotly_click events on the cells https://github.com/plotly/plotly.js/pull/2052#issuecomment-333874540
- [x] roll back `gup` utils to `parcoords`: https://github.com/plotly/plotly.js/pull/2052#discussion_r142444229
- [x] get rid of camelcased classnames (use dashed) and put all classnames into `constants.js`: https://github.com/plotly/plotly.js/pull/2052#discussion_r142415494
- [ ] discuss and plan cell editing and possibly other related functionality ([request](https://github.com/plotly/plotly.js/pull/2052#issuecomment-333874540), [reply](https://github.com/plotly/plotly.js/pull/2052#issuecomment-333992189))
- [ ] represent full `table` state via attributes (in) and callbacks (out) such that users can specify and/or receive, in logical units (row item) where we are in scrolling; similar full state representation for other interactions such as column drag&drop https://github.com/plotly/plotly.js/pull/2052#issuecomment-333949519
- [ ] put back `valign` (used to be there but needs a bit of work to handle cases of LaTeX etc.)
- [ ] do more to avoid wrapping: initial render should try to render straight away (no wrapping) and iff it breaches constraints, wrapping is invoked - this makes the current 2-step process into a 3-step process but CPU upside is big if no or few cells need wrapping (worth optimizing for the plainer case and negligible cycle loss even in the worst case)
- [ ] support individual border lines in addition to boxes; maybe border collapse?
- [x] Support for 1D header values https://github.com/plotly/plotly.js/pull/2052#discussion_r142827253
- [x] Better handling of MathJax fill color styling https://github.com/plotly/plotly.js/pull/2052#discussion_r142469257
- [ ] Support `header.format` scalar value https://github.com/plotly/plotly.js/pull/2052/files#r142410512
- [x] Reuse `plots/font_attributes.js` https://github.com/plotly/plotly.js/pull/2052/files#r142410981
- [x] Move `<clipPath>` to `fullLayout._clips` https://github.com/plotly/plotly.js/pull/2052/files#r142418318
- [ ] Jasmine tests
- [x] Checking for render glitch (high speed scrolling and then stopping) and fixing it (issue found by @rreusser )
- More `defaults.js` logic to let users make do with fewer `attributes`, for example, 
[...]

PR #2067, opened 2017-10-06
Lisp-casing class, id names and factoring them out in `table`, `parcoords`, `sankey`

monfera:
An item in https://github.com/plotly/plotly.js/issues/2056

@etpinard is it how you meant it, or should the test cases also pull the class names from the constants?


etpinard:
Nicely done :dancer:

PR #2068, opened 2017-10-06
IE/Edge SVG export fixes

### LINKED ISSUE #2063: SVG downloads in IE/Edge contain literal data URL
When using `downloadImage` to save a SVG snapshot via MS IE or Edge, the saved file contains the literal data URL value rather than the decoded SVG source code.

I believe this is a consequence of https://github.com/plotly/plotly.js/commit/22d902ba00822496e7a9edbf1ce73689b78a37c1 before which `url` contained the raw SVG source. That commit is probably correct to only put URL values in `url`, but that means the URL should be decoded before constructing a Blob and saving as a file.

See my PR here for a fix: https://github.com/arbscht/plotly.js/pull/1

alexcjohnson:
Based off https://github.com/arbscht/plotly.js/pull/1 by @arbscht

Fixes #2063 and possibly #699

@arbscht can you confirm that this produces good output for you - particularly can you verify that a zoomed-in plot is still clipped correctly in the exported SVG?

cc @etpinard @bpostlethwaite

### DIFF
--- a/src/snapshot/filesaver.js
+++ b/src/snapshot/filesaver.js
@@ -53,7 +53,11 @@ var fileSaver = function(url, name) {
 
         // IE 10+ (native saveAs)
         if(typeof navigator !== 'undefined' && navigator.msSaveBlob) {
-            navigator.msSaveBlob(new Blob([url]), name);
+            // At this point we are only dealing with a SVG encoded as
+            // a data URL (since IE only supports SVG)
+            var encoded = url.split(/^data:image\/svg\+xml,/)[1];
+            var svg = decodeURIComponent(encoded);
+            navigator.msSaveBlob(new Blob([svg]), name);
             resolve(name);
         }
 
--- a/test/jasmine/tests/download_test.js
+++ b/test/jasmine/tests/download_test.js
@@ -32,7 +32,6 @@ describe('Plotly.downloadImage', function() {
 
     afterAll(function() {
         document.createElement = createElement;
-        navigator.msSaveBlob = msSaveBlob;
     });
 
     beforeEach(function() {
@@ -43,6 +42,7 @@ describe('Plotly.downloadImage', function() {
         destroyGraphDiv();
         Lib.isIE = isIE;
         slzProto.serializeToString = serializeToString;
+        navigator.msSaveBlob = msSaveBlob;
     });
 
     it('should be attached to Plotly', function() {
--- a/test/jasmine/tests/download_test.js
+++ b/test/jasmine/tests/download_test.js
@@ -12,37 +12,27 @@ describe('Plotly.downloadImage', function() {
     'use strict';
     var gd;
 
-    // override click handler on createElement
-    //  so these tests will not actually
-    //  download an image each time they are run
-    //  full credit goes to @etpinard; thanks
     var createElement = document.createElement;
-    var msSaveBlob = navigator.msSaveBlob;
-    var isIE = Lib.isIE;
     var slzProto = (new window.XMLSerializer()).__proto__;
     var serializeToString = slzProto.serializeToString;
 
-    beforeAll(function() {
-        document.createElement = function(args) {
+    beforeEach(function() {
+        gd = createGraphDiv();
+
+        // override click handler on createElement
+        //  so these tests will not actually
+        //  download an image each time they are run
+        //  full credit goes to @etpinard; thanks
+        spyOn(document, 'createElement').and.callFake(function(args) {
             var el = createElement.call(document, args);
             el.click = function() {};
             return el;
-        };
-    });
-
-    afterAll(function() {
-        document.createElement = createElement;
-    });
-
-    beforeEach(function() {
-        gd = createGraphDiv();
+        });
     });
 
     afterEach(function() {
         destroyGraphDiv();
-        Lib.isIE = isIE;
-        slzProto.serializeToString = serializeToString;
-        navigator.msSaveBlob = msSaveBlob;
+        delete navigator.msSaveBlob;
     });
 
     it('should be attached to Plotly', function() {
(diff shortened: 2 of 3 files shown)


arbscht:
@alexcjohnson Thanks, I can confirm this works for me. SVG files are downloaded, valid, and clipped correctly when zoomed.

alexcjohnson:
Thanks @arbscht - and thanks again for the fix that got us most of the way there! I'm going to wait for a review from @etpinard or @bpostlethwaite before merging - which with any luck will NOT come today 🇨🇦 🦃

arbscht:
As a sidenote, I don't think this really affects #699. This change only fixes a regression to do with SVG files generated in IE/Edge. Downloading PNG files with MS browsers is still not supported in any code path as far as I can see. In fact, `filesaver.js` assumes that the desired format is SVG when creating blobs for MS browsers, without consulting `format`.

That assumption would need to be changed if one were to implement the solution outlined by @etpinard in https://github.com/plotly/plotly.js/issues/699#issuecomment-255117576 but I'd look at revisiting that separately.

alexcjohnson:
> As a sidenote, I don't think this really affects #699. This change only fixes a regression to do with SVG files generated in IE/Edge.

Ah good read - seems that issue morphed after it was originally posted, as the initial report did deal with a regression introduced in v1.13.0 but later on a discussion of possible improvements for Edge specifically was added. Good, I will leave #699 open after this gets merged.

etpinard:
> Downloading PNG files with MS browsers is still not supported in any code path

I doubt that something we'll ever support.

etpinard:
> Downloading PNG files with MS browsers is still not supported in any code path
> I doubt that something we'll ever support.

Oops. Let me try that again:

I don't think we'll ever support downloading PNGs in IE. But, making this work in Edge should be doable.

etpinard:
So I would either rename https://github.com/plotly/plotly.js/issues/699 or close and open a new issue named: `toImage` in Edge.

src/snapshot/filesaver.js
```diff
@@ -53,7 +53,11 @@ var fileSaver = function(url, name) {
 
         // IE 10+ (native saveAs)
         if(typeof navigator !== 'undefined' && navigator.msSaveBlob) {
-            navigator.msSaveBlob(new Blob([url]), name);
+            // At this point we are only dealing with a SVG encoded as
+            // a data URL (since IE only supports SVG)
+            var encoded = url.split(/^data:image\/svg\+xml,/)[1];
```
etpinard: I wonder what's better (i.e. faster) 

```js
url.split(/^data:image\/svg\+xml,/)[1];

// or as in plot_api/to_image.js
url.replace(/^data:image\/svg\+xml,/, '');
```

test/jasmine/tests/download_test.js
```diff
@@ -59,8 +70,55 @@ describe('Plotly.downloadImage', function() {
...
+            return Plotly.downloadImage(gd, {
+                format: 'svg',
+                height: 300,
+                width: 300,
+                filename: 'plotly_download'
+            });
+        })
+        .then(function() {
+            expect(savedBlob).toBeDefined();
+            if(savedBlob === undefined) return;
```
etpinard: Perhaps

```js
if(saveBlob === undefined) {
  fail('undefined save blob');
}
```

would be more readable than an `expect` + early `return` sequence.

test/jasmine/tests/download_test.js
```diff
@@ -59,8 +70,55 @@ describe('Plotly.downloadImage', function() {
...
 
+    it('should produce the right SVG output in IE', function(done) {
+        // mock up IE behavior
+        Lib.isIE = function() { return true; };
+        slzProto.serializeToString = function() {
+            return serializeToString.apply(this, arguments)
+                .replace(/(\(#)([^")]*)(\))/gi, '(\"#$2\")');
+        };
+        var savedBlob;
+        navigator.msSaveBlob = function(blob) { savedBlob = blob; };
```
etpinard: It would be best to use jasmine [`spyOn` and `callFake`](https://jasmine.github.io/2.0/introduction.html#section-Spies:_<code>and.callFake</code>) that way we wouldn't have to worry about restoring the unmocked behavior.

etpinard:
Great PR. 

Thanks very much @arbscht for bringing this to our attention. You win the _community PR of the week_ :trophy:
 
I gave up on https://github.com/plotly/plotly.js/issues/699 a long time ago. This will make a bunch of people happy :tada: 

@alexcjohnson I made a few comments. All of them are non-blocking, if you're pressed for time.

src/snapshot/filesaver.js
alexcjohnson: I don't see perf as too important in this block. An advantage (at least it looks like an advantage to me...) of the current one is that if `url` does NOT start with the expected `'data:image/svg+xml,'` it will throw an error, rather than saving garbage, which seems easier to investigate.

etpinard: > it will throw an error, rather than saving garbage,

Good point. :+1:

test/jasmine/tests/download_test.js
alexcjohnson: ha, you can only `spyOn` something that already exists - so `msSaveBlob` I have to handle manually. But `Lib.isIE` and `serializeToString` (and perhaps `document.createElement`??) I can switch to `spyOn().and.callFake`

alexcjohnson:
> So I would either rename #699 or close and open a new issue named: toImage in Edge.

A new issue seemed clearest -> #2077

test/jasmine/tests/download_test.js
etpinard: Looks good!

etpinard:
:dancer: nicely done.

Braintelligence:
@alexcjohnson I'm not sure if this is the right place to mention this, but I think I encountered a bug for IE SVG export. If you download an SVG with Internet Explorer Version 11.0.9600.19137 that renders via trident engine (Edge) you get this:
`<g class='plot' clip-path='url('https://www.domain.tld#clip00a1b1xyplot')' transform='translate(110 30)'>`
instead of this:
`<g class="plot" transform="translate(110, 30)" clip-path="url(https://www.domain.tld#clip189888xyplot)">`
which results in a wrongly formatted XML (see the `'` that should be `"`).

I'm not sure how to fix this. Is this a very specific problem with this browser running on trident?

Issue #1630, opened 2017-04-27
formatTime error when starting test dashboard


meteoric199:
运行 `npm start` 时,`tasks\util\common.js` `48`行报错.
```
date.toString().match(/\(([A-Za-z\s].*)\)/)[1]
```
但是有时 `date.toString()` 的值是 `Thu Apr 27 2017 18:31:16 GMT+0800 (中国标准时间)`

系统语言:中文简体
时区: `(UTC+08:00)beij北京,重庆...`

English:

When I  run `npm start` , there has a error in` line 48` in the file of `tasks\util\common.js`.
```
date.toString().match(/\(([A-Za-z\s].*)\)/)[1]
```
sometimes the value of `date.toString()` is  `Thu Apr 27 2017 18:31:16 GMT+0800 (中国标准时间)`

System Language:Chinese-Simplified 
Time zone:  `(UTC+08:00)BeiJing,ChongQing ...`.

I'm not good at English.If there have some errors ,please point out.

error:
```
TypeError: Cannot read property '1' of null
    at Object.exports.formatTime (D:\workspace\tool\plotly.js\tasks\util\common.js:48:52)
    at Browserify.<anonymous> (D:\workspace\tool\plotly.js\tasks\util\watchified_bundle.js:77:36)
    at emitOne (events.js:96:13)
    at Browserify.emit (events.js:188:7)
    at DestroyableTransform.end [as _flush] (D:\workspace\tool\plotly.js\node_modules\watchify\index.js:81:15)
    at DestroyableTransform.<anonymous> (D:\workspace\tool\plotly.js\node_modules\through2\node_modules\readable-stream\lib\_stream_transform.js:115:49)
    at DestroyableTransform.g (events.js:291:16)
    at emitNone (events.js:86:13)
    at DestroyableTransform.emit (events.js:185:7)
    at prefinish (D:\workspace\tool\plotly.js\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:494:12)
    at finishMaybe (D:\workspace\tool\plotly.js\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:502:7)
    at endWritable (D:\workspace\tool\plotly.js\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:514:3)
    at DestroyableTransform.Writable.end (D:\workspace\tool\plotly.js\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:484:41)
```

[...]

PR #2069, opened 2017-10-06
fix Chinese formatDate error on npm start

Handsome2734:
Fix formatDate error on `npm start` when system language is Chinese. #1630


alexcjohnson:
Thanks! I bet for compatibility with any arbitrary language we could do something even more permissive like `/\(([^\(\)]*\)\)/` - or even drop the timezone, it's not the most crucial bit of info there. But I'm glad this fixes it for Chinese anyway. 💃

PR #2070, opened 2017-10-06
Reuse font_attributes in `table`

monfera:
As in https://github.com/plotly/plotly.js/issues/2056


etpinard:
There's another one here:

https://github.com/plotly/plotly.js/blob/6075d867c6130c4bb18a517cb93fe7ec807cc5a5/src/traces/table/attributes.js#L217

PR #2072, opened 2017-10-06
Allow 1D header in `table`

monfera:
An item in https://github.com/plotly/plotly.js/issues/2056

One mock updated to show its use.

### DIFF
--- a/src/traces/table/defaults.js
+++ b/src/traces/table/defaults.js
@@ -52,6 +52,12 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     Lib.coerceFont(coerce, 'cells.font', fontDflt);
 
     coerce('header.values');
+    for(var i = 0; i < traceOut.header.values.length; i++) {
+        if(!Array.isArray(traceOut.header.values[i])) {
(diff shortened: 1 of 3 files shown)


src/traces/table/defaults.js
```diff
@@ -52,6 +52,12 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     Lib.coerceFont(coerce, 'cells.font', fontDflt);
 
     coerce('header.values');
+    for(var i = 0; i < traceOut.header.values.length; i++) {
+        if(!Array.isArray(traceOut.header.values[i])) {
+            traceOut.header.values[i] = [traceOut.header.values[i]];
+        }
+    }
+
```
alexcjohnson: This makes me vaguely uneasy - having a different structure of the attribute in `data` vs `_fullData`. For example, the workspace looks at `_fullData` for what the current values are in order to construct controls, but then feeds control output back into `data`. It might be fine, but if there's a way to push this farther down, into `data_preparation_helper` perhaps, that would make me more comfortable.

monfera: Thanks, will do!

PR #2074, opened 2017-10-09
No default needed for table line

monfera:
Currently, the JSON spec must have a `line` spec, which is not ideal as it can be supplied with good defaults.


alexcjohnson:
Good catch, thanks! 💃

PR #2075, opened 2017-10-09
Drag outside window

alexcjohnson:
`document` keeps getting `mousemove` events even outside the window, until `mouseup`. Who knew?!? Not me anyway... but now that I do, this quick change allows dragging to keep going outside the window. Tested manually in all my local browsers, IE9 - 11 and iOS on Browserstack, and of course the test suites still work (after removing `mouseout` which is no longer part of the interaction).

Two notes:
- I kept `coverSlip` around, because that's an easy way to prevent selecting text as you drag around.
- Some [tips](https://stackoverflow.com/questions/1685326/responding-to-the-onmousemove-event-outside-of-the-browser-window-in-ie) say you need to use `setCapture` / `releaseCapture` in IE, but a) that seems to be older versions than we target, and b) might have been mitigated by `coverSlip` anyway. I saw no difference if I included it in our code.
- I included the IE polyfills in test_dashboard, commented out, to make it easier to test down to IE9 in the future.

cc @etpinard

### DIFF
--- a/devtools/test_dashboard/index.html
+++ b/devtools/test_dashboard/index.html
@@ -19,6 +19,10 @@
   </div>
   <div id="snapshot"></div>
 
+  <!-- uncomment below for IE9/10 support -->
+  <!-- <script>if(typeof window.Int16Array !== 'function')document.write("<scri"+"pt src='../../dist/extras/typedarray.min.js'></scr"+"ipt>");</script>
+  <script>document.write("<scri"+"pt src='../../dist/extras/request_animation_frame.js'></scr"+"ipt>");</script> -->
+
   <script type="text/javascript" src="../../dist/extras/mathjax/MathJax.js?config=TeX-AMS-MML_SVG"></script>
   <script id="source" type="text/javascript" src="../../build/plotly.js"></script>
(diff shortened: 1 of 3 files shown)


etpinard:
Looks good to me. 

I'm curious though, what made you investigate `document` mouse events on drag in the first place?

alexcjohnson:
> Looks good to me. 

is that a 💃 ?

> I'm curious though, what made you investigate `document` mouse events on drag in the first place?

Just noticed that google maps allows dragging like that and wondered how 🤔

etpinard:
:dancer: 

I asked the question in case we could have locked some behavior down in a meaningful test.

alexcjohnson:
> I asked the question in case we could have locked down some behavior in a meaningful test.

I thought about that a little and didn't see anything. Our own [mouse_event](https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/mouse_event.js#L27) helper will fail outside the window (`document.elementFromPoint` returns `null`) so I don't really see a way to trigger the browser's internal behavior.

Fortunately the new behavior is actually simpler on our side than the old, as it doesn't need to worry about `mouseOut`.

PR #2076, opened 2017-10-09
Phase out *-baseline in `table` as IE doesn't follow this part of the SVG standard

monfera:
Local, recent Firefox doesn't adhere to `alignment-baseline` here, this tests effects on CI.


monfera:
It fixed the local FF misalignment of unwrapped lines:

image

... and it didn't break the same thing which evidently worked fine on CI (test cases intact).

Also, adding a similar fix to `parcoords`, where the effect of the problem (locally) was much less noticeable, and similarly, it worked fine on CI. 

So I'm proposing this for merging.

monfera:
Thanks @alexcjohnson for bringing up an issue seen in CI (separate one, not dealt with by this PR) that led to to these FF problems.

alexcjohnson:
I do like those mocks a bit better, but now I'm rather confused... parcoords plots look fine in IE11 but according to a quick check on browserstack neither `alignment-baseline` nor `dominant-baseline` works there? How is that possible? Suffice it to say it doesn't seem to me like these attributes (or styles?) are really ready for prime time. How hard would it be to 🔪  them entirely?

dominant baseline chrome - http://bl.ocks.org/eweitnauer/7325338 (assumed to be the authority)
dominant baseline chrome

dominant baseline ff - seems to work but somewhat differently (baseline-shift fails, dominant-baseline has small differences)
dominant baseline ff

dominant baseline ie11 - fail
dominant baseline ie11

alignment baseline chrome - https://bl.ocks.org/emmasaunders/0016ee0a2cab25a643ee9bd4855d3464
alignment baseline chrome

alignment baseline ff - fail
alignment baseline ff

alignment baseline ie11 - fail
alignment baseline ie11

monfera:
Indeed! This says it's not supported by IE: https://msdn.microsoft.com/en-us/library/gg558060(v=vs.85).aspx

Looks like we need to do the `dy: 0.35em` trick for `central`/`middle` and something analogous for `hanging`.

monfera:
There are still some vertical misalignments specifically in CI, adding a separate item. The purpose of this ticket is, after @alexcjohnson 's IE testing, merely the replacement of `*-baseline` attributes with something even IE adheres to. This is a big improvement because, without this PR, Firefox and especially IE rendering have serious alignment problems so its separate merge is suggested.

PR #2081, opened 2017-10-12
Return empty set for invisible bar trace

monfera:
Downstream processing expects an array here, even when a trace is not `visible`:

```
                    var thisSelection = fillSelectionItem(
                        searchInfo.selectPoints(searchInfo, poly), searchInfo
                    );
                    if(selection.length) {
                        for(var j = 0; j < thisSelection.length; j++) { // <- won't have a length as selectPoints is undefined
                            selection.push(thisSelection[j]);
                        }
                    }
```

(side note: `fillSelectionItem` handles the case of a non-array `selection`, in which case it just passes it on ie. doesn't try to set `curveNumber` etc. properties on `undefined`; the result of which is the same `undefined` as the value of the first arg)

There's a similar `return;` in `scatter`, `scattergeo`, `scattermapbox`.

Is it a good direction @etpinard ?

### DIFF
--- a/test/jasmine/tests/select_test.js
+++ b/test/jasmine/tests/select_test.js
@@ -142,6 +142,7 @@ describe('Test select box and lasso in general:', function() {
             .map(function(v) { return 'id-' + v; });
         mockCopy.data[0].customdata = mockCopy.data[0].y
             .map(function(v) { return 'customdata-' + v; });
+        addInvisible(mockCopy);
 
         var gd;
         beforeEach(function(done) {
@@ -213,6 +214,7 @@ describe('Test select box and lasso in general:', function() {
     describe('lasso events', function() {
         var mockCopy = Lib.extendDeep({}, mock);
         mockCopy.layout.dragmode = 'lasso';
+        addInvisible(mockCopy);
 
         var gd;
         beforeEach(function(done) {
@@ -293,6 +295,9 @@ describe('Test select box and lasso in general:', function() {
     });
 
     it('should skip over non-visible traces', function(done) {
+        // note: this tests a mock with one or several invisible traces
+        // the invisible traces in the other tests test for multiple
+        // traces, with some visible and some not.
         var mockCopy = Lib.extendDeep({}, mock);
         mockCopy.layout.dragmode = 'select';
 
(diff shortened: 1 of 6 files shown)


etpinard:
Good catch.

I'd vote for making:

```js
if(selection.length) { /*

// turn into
if(selection && selection.length) { /*
```

that way we won't have to remember to always return an array in all the trace module `selectPoints` methods.

monfera:
@etpinard it's not `selection` but `thisSelection` that is `undefined`.

Also, we can't just do `if(thisSelection && selection.length` because there's an `else` branch too:

```
            function() {
                selection = [];
                    searchInfo = searchTraces[i];
                    var thisSelection = fillSelectionItem(
                        searchInfo.selectPoints(searchInfo, poly), searchInfo
                    );
                    if(selection.length) {
                        for(var j = 0; j < thisSelection.length; j++) {
                            selection.push(thisSelection[j]);
                        }
                    }
                    else selection = thisSelection;
                }

                eventData = {points: selection};
                fillRangeItems(eventData, poly, pts);
                dragOptions.gd.emit('plotly_selecting', eventData);
            }
```
We probably don't want `eventData.points` to end up being `undefined`.

Looking at this block a bit more, it feels like it could be simplified: `selection` is initialized (to `[]`) but its initial value is discarded in the `else` branch; also, @alexcjohnson is the purpose of the `else` branch to ensure `selection` identical with `thisSelection`, and is it important? It probably has no impact on performance and the `if/else` could be simplified to just one straight code path if a shallow copy is fine too.

As resolution of this `plotly.js` issue but ancillary to `crossfilter` work I tried to avoid delving into it too deep, and always returning the same `[...]` type - set, or empty set, doesn't matter -
 seemed elegant and of the smallest code impact.

etpinard:
> always returning the same [...] type - set, or empty set, doesn't matter -
seemed elegant and of the smallest code impact.

Ok, if you think that's more elegant, I won't argue with you. Just make sure lock that in tests for all `selectPoints` methods.

alexcjohnson:
Good catch @monfera - I don't care exactly what route you take, but as long as the event always has a `points` array (even if it's empty) and it's robust against missing inputs, I'm happy.

test/jasmine/tests/select_test.js
```diff
@@ -824,3 +855,16 @@ describe('Test select box and lasso per trace:', function() {
         .then(done);
     });
 });
+
+// to make sure none of the above tests fail with extraneous invisible traces,
+// add a bunch of them here
+function addInvisible(fig, canHaveLegend) {
```
etpinard: Brilliant :sparkles:

etpinard:
:dancer: nice tag-team effort on this one 

:muscle:  @monfera @alexcjohnson :muscle:

monfera:
@alexcjohnson incredible commit, thanks!

Issue #248, opened 2016-02-10
Tools (e.g. select) should not execute on contextmenu


joshua-gould:
All tools (e.g. select) have their click handler invoked when right-clicking.

etpinard:
Can I ask you why?

joshua-gould:
I'd like to create a custom popup menu on right-click (contextmenu event)

etpinard:
@joshua-gould I labelled this issue with the `community` tag meaning that the no plotly.js team member will dedicate time to it, but PR are welcomed to address this issue.

AlexVvx:
Hi, guys. I pulled pr about context menu. When {rightClick: false} is in options, then 'contextmenu' event will be fired on element and default browser menu prevented. Please take a look at https://github.com/AlexVvx/plotly.js/pull/1/files

alexcjohnson:
@AlexVvx interesting, thanks for the PR! I wonder though if there isn't a simpler solution - we probably don't really want to trigger `dragelement` if you right click no matter what. Can we just look at `e.button` and/or `e.buttons` and bail out of `onStart` if it's not a left-click? If we do that, your own `contextmenu` event handler should pick up the event, right?

I haven't looked at browser support of the button attributes - note that we target IE9+ in addition to current versions of Chrome, FF, Edge, and Safari.

Touch interactions are an interesting case, you generally get `contextmenu` on a long press there, right? But might this already work correctly, since in touch environments we don't create a `coverSlip` at all (which I'm assuming is what interferes with the expected `contextMenu` on desktop)?

AlexVvx:
Sounds good, applied changes in [new pr](https://github.com/AlexVvx/plotly.js/pull/2)

alexcjohnson:
closed by @AlexVvx in https://github.com/plotly/plotly.js/pull/2087

PR #2087, opened 2017-10-13
Check if not a left-click, then bail out of onStart

AlexVvx:
Related to [#248](https://github.com/plotly/plotly.js/issues/248)

This PR adds check in dragElement. On start, if not a left-click, then bail out of onStart. The purpose of this check is to let developers to use their own context menu instead of browser context menu. Without that check there were no possibility to see 'contextmenu' event when click inside plot container.


test/jasmine/tests/dragelement_test.js
```diff
@@ -122,6 +122,19 @@ describe('dragElement', function() {
...
 
+    it('should not add a cover slip div to the DOM when right click', function() {
+        function countCoverSlip() {
+            return d3.selectAll('.dragcover').size();
+        }
+
+        var options = { element: this.element, gd: this.gd };
+        dragElement.init(options);
+
+        options.element.onmousedown({button: 2})
```
AlexVvx: Not sure, that it's a best way to emulate right click, however I can't see any other options.

alexcjohnson: looks like in [mouse_event](https://github.com/plotly/plotly.js/blob/f1faa2871dbf182847579166af47d1cc2a983707/test/jasmine/assets/mouse_event.js#L11) we have `buttons` but if we add `button` the same way it should allow right-click simulation?

AlexVvx:
I wasn't able to find 'Buttons' in ie spec, but it works well in ie9, so I changed everything to 'buttons'.

test/jasmine/tests/dragelement_test.js
```diff
@@ -122,6 +122,18 @@ describe('dragElement', function() {
...
+    it('should not add a cover slip div to the DOM when right click', function() {
+        function countCoverSlip() {
+            return d3.selectAll('.dragcover').size();
+        }
+
+        var options = { element: this.element, gd: this.gd };
+        dragElement.init(options);
+
+        mouseEvent('mousedown', this.x, this.y, { buttons: 2 });
+        expect(countCoverSlip()).toEqual(0);
```
alexcjohnson: Great! Very clear correspondence to the test above. Two last changes I would suggest, then I think this is ready to go:
- 🌴  Can you move `countCoverSlip` out one scope so we don't have to repeat it?
- Can you continue through `mouseup` as in the test below, and verify that `handleClick` has not been called? (that one ironically looks annoying to DRY 🌴  so you can just repeat the pieces you need.)

AlexVvx: Sure

alexcjohnson:
Fantastic @AlexVvx - thanks for the PR and updates! 💃  - we're calling this a bug fix, so it can go in the release we're planning later today.

PR #2091, opened 2017-10-16
yet another contour algorithm bug fix

### LINKED ISSUE #2088: Contour generation fails for this z data
See https://codepen.io/etpinard/pen/NaEWmN?editors=1010

First reported in https://community.plot.ly/t/plotly-contour-problem/6299

alexcjohnson:
I'm always amazed how many edge cases (literal edge cases in fact 🖼 ) there are with contour plots, and that our tests still hadn't (is that too optimistic a word? should I say haven't?) covered them all.

Fixes #2088


alexcjohnson:
Ah crap - it's not that simple 😢 
screen shot 2017-10-16 at 1 54 32 pm

src/traces/contour/find_all_paths.js
```diff
@@ -198,41 +193,45 @@ function makePath(pi, loc, edgeflag, xtol, ytol) {
...
+                        if(j === i) {
                             // the path is now closed
-                            pi.paths.push(pts.concat(edgepath2));
+                            pi.paths.push(pts.concat(edgepathj));
                         }
                         else {
-                            pi.edgepaths[edgei2] =
-                                pi.edgepaths[edgei2].concat(pts, edgepath2);
+                            if(j > i) j--;
+                            pi.edgepaths[j] = edgepathj.concat(pts, edgepathi);
```
alexcjohnson: in the end it was all about these two lines - getting the right index (`if(j > i) j--;`) and using the right paths (previously we were doubling up `edgepath2 <-> edgepathj` but because of the poor usage it was hard to tell that. Surprising that we hadn't hit this before in our tests, but now we do!

rreusser: Nice! Then this also applies to carpet contour plots which use the same function. 🎉

etpinard:
Nicely done :dancer:

PR #2093, opened 2017-10-16
Remove smith attribute

alexcjohnson:
We had a plan a long time ago to support smith charts, but never did.
If/when we decide to do this, it probably won't be implemented with a simple global attribute anyhow.

cc @cldougl


etpinard:
:dancer:

cldougl:
💃

PR #2099, opened 2017-10-18
Selection on `choropleth` to work even if an invisible trace is present

monfera:
Box select and lasso to work in the presence of an invisible `choropleth` trace. Updated test case signals if the small fix is commented out.

### DIFF
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -23,6 +23,11 @@ var MINSELECT = constants.MINSELECT;
 
 function getAxId(ax) { return ax._id; }
 
+function visible(searchInfo) {
+    var cd0 = searchInfo.cd[0];
+    return cd0 && cd0.trace && cd0.trace.visible === true;
+}
+
 module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
     var zoomLayer = dragOptions.gd._fullLayout._zoomlayer,
         dragBBox = dragOptions.element.getBoundingClientRect(),
@@ -190,6 +195,7 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
                 selection = [];
                 for(i = 0; i < searchTraces.length; i++) {
                     searchInfo = searchTraces[i];
+                    if(!visible(searchInfo)) continue;
                     var thisSelection = fillSelectionItem(
                         searchInfo.selectPoints(searchInfo, poly), searchInfo
                     );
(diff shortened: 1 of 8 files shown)


monfera:
Thanks @alexcjohnson, I understand the general direction for preventing the situation from emerging, not sure how I'd go about it as `node3` etc. are handled outside the scope of selection. I'd need to urgently work on `table.restyle` but circle back to this ASAP.

alexcjohnson:
@monfera the idea is that the only reason `node3` would not be set is if the trace is not visible. So if we filter invisible traces out before even calling `selectPoints`, we don't have to worry about this kind of error in any of the individual modules.

etpinard:
> but I wonder if we should preempt problems like this by just filtering out invisible traces (visible !== true, to account for legendonly)

This is a much better solution in my opinion.

src/plots/cartesian/select.js
```diff
@@ -190,6 +195,7 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
                 selection = [];
                 for(i = 0; i < searchTraces.length; i++) {
                     searchInfo = searchTraces[i];
+                    if(!visible(searchInfo)) continue;
```
etpinard: Nice. Does that mean we can :hocho: those `visible !== false` clauses [here](https://github.com/plotly/plotly.js/blob/master/src/traces/bar/select.js#L20), [here](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/select.js#L29), [here](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergeo/select.js#L25), [here](https://github.com/plotly/plotly.js/blob/86e7135f54fedfa1e89b3e299833491369baa96e/src/traces/scattermapbox/select.js#L27) and [here](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/select.js#L29)?

monfera: Great find, indeed these are the only callers, and `select` check is almost always on the top, except `scattermapbox/select/js` where it's okay to set the `_hasDimmedPts` to `false` anyway:

```
    trace._hasDimmedPts = false;

    if(!subtypes.hasMarkers(trace)) return [];
```
Pushing a new commit now

monfera:
moved check more upstream on Alex's suggestion

etpinard:
:dancer: for me. Great PR @monfera !

monfera:
Thanks @etpinard and @alexcjohnson for the added clarity and code size reducing suggestions!

Issue #2092, opened 2017-10-16
Restyling table traces


VeraZab:
is there something that I'm doing wrong here:
https://codepen.io/veraz/pen/XeydOX?editors=1010

can't seem to be able to use Plotly.restyle on table traces

cc @monfera

monfera:
Table restyle needs a bit of work, on the list of the todo gh items, will look into it.

VeraZab:
we need restyle to put this into the workspace..

monfera:
I hastily assumed `restyle` doesn't work en bloc as `table` code was not tested (manually or otherwise) for `restyle`, in which case it's almost certain it'd have issues. Writing jasmine test cases, the couple of `restyle` things covered so far do work so I gave a shot to this one. It seems to be OK with the current build, and some slight changes to the codepen:

https://codepen.io/anon/pen/boZbRq?editors=1010

I'm carrying on with the test cases and manual tests as there may well be other `restyle` tests that'd still fail now.

VeraZab:
oups, looked at your new pen, and can see from my old one that I was not using the plotly-latest.js bundle I was using:
<script src="https://rawgit.com/monfera/plotly.js/table-squashed/dist/plotly.min.js"></script>

Because I forked the pen from one of your old ones :)

But even when I changed to plotly-latest, and try to do a restyle call to columnwidth, same as you did in your codepen, my table doesn't update. The console shows: Plotly.version
"1.31.1" And when I go back to your codepen and play with changing columnwidth a few times, sometimes it works, sometimes it doesn't

my edited codepen: https://codepen.io/veraz/pen/XeydOX?editors=1010

maybe it's a codepen thing, but in our tests, can we try using restyle on the same property a few times, to see if it behaves as expected? thanks!

monfera:
@VeraZab just a guess, the thing with `columnWidth` is that `layout.width` combined with `data[n].domain` actually determine the full table width; therefore `columnWidth` gets scaled. If you try it with either a smaller `layout.width`, or a smaller `domain`, you'll get narrower, equal-size columns. My example shows various column widths because `columnWidth` is supplied in an array of different numbers.

Hmm just looked the `attributes`, the doc for `columnWidth` doesn't say the above, so some kind of update is needed to bring user expectations and code in sync.

PR #2107, opened 2017-10-19
Improvements to table: `restyle` fix, 0 row/column and misc. fixes; jasmine tests

monfera:
- `restyle` fix was urgent due to https://github.com/plotly/plotly.js/issues/2092
- why not take the opportunity to add jasmine coverage for most of what exists
- tests uncovered misc. special case issues which were fixed

### DIFF
--- a/test/jasmine/tests/table_test.js
+++ b/test/jasmine/tests/table_test.js
@@ -1,0 +1,289 @@
+var Plotly = require('@lib/index');
+var Lib = require('@src/lib');
+var Plots = require('@src/plots/plots');
+var Table = require('@src/traces/table');
+var cn = require('@src/traces/table/constants').cn;
+
+var createGraphDiv = require('../assets/create_graph_div');
+var destroyGraphDiv = require('../assets/destroy_graph_div');
+
+var mockMulti = require('@mocks/table_latex_multitrace.json');
+
+// mock with two columns; lowest column count of general case
+var mock2 = Lib.extendDeep({}, mockMulti);
+mock2.data = [mock2.data[2]]; // keep the subplot with two columns
+
+// mock with one column as special case
+var mock1 = Lib.extendDeep({}, mock2);
+mock1.data[0].header.values = mock1.data[0].header.values.slice(0, 1);
+mock1.data[0].cells.values = mock1.data[0].cells.values.slice(0, 1);
+
+// mock with zero columns; special case, as no column can be rendered
+var mock0 = Lib.extendDeep({}, mock1);
+mock0.data[0].header.values = [];
+mock0.data[0].cells.values = [];
+
+var mock = require('@mocks/table_plain_birds.json');
+
+describe('table initialization tests', function() {
+
+    'use strict';
+
+    describe('table global defaults', function() {
+
+        it('should not coerce trace opacity', function() {
+            var gd = Lib.extendDeep({}, mock1);
+
+            Plots.supplyDefaults(gd);
+
+            expect(gd._fullData[0].opacity).toBeUndefined();
+        });
+
+        it('should use global font as label, tick and range font defaults', function() {
+            var gd = Lib.extendDeep({}, mock1);
+            delete gd.data[0].header.font;
+            delete gd.data[0].cells.font;
+            gd.layout.font = {
+                family: 'Gravitas',
+                size: 20,
+                color: 'blue'
+            };
+
+            Plots.supplyDefaults(gd);
+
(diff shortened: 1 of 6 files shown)


VeraZab:
I'll test out this branch on streambed, see if the restyle calls I need work now, thanks!

VeraZab:
ok, I see the improvements, 

here's a few observations: 
1) there's  inconsistency where I I need to wrap my data into an extra set of arrays to make some calls work:
for example: https://codepen.io/veraz/pen/EwMLBQ?editors=1010
I can make Plotly.restyle('graphDiv', {'header.values': [[['A'], ['B'], ['C']]]}) work, with one extra set of arrays. It's the same for a few other calls like: Plotly.restyle(gd, {'cells.font.size': [[16, 5]]})

2) A restyle to `cells.fill.color` behaves unexpectedly, it should color by row right? not by column..

I'd really like to see that extra [] wrapping fixed in this pr, would really help for the workspace integration.

Thanks!

test/jasmine/tests/table_test.js
```diff
@@ -0,0 +1,290 @@
...
+            }
+
+            restyleValues('cells', 0, [['new cell content 1', 'new cell content 2']])()
+                .then(restyleValues('cells', 2, [[0, 0.1]]))
+                .then(restyleValues('header', 1, [['Species top', 'Species bottom']]))
+                .then(done);
+        });
+
+        it('Calling `Plotly.relayout` with string should amend the preexisting parcoords', function(done) {
+            expect(gd.layout.width).toEqual(1000);
```
VeraZab: preexisting 'tables' instead of 'parcoords'

test/jasmine/tests/table_test.js
```diff
@@ -0,0 +1,290 @@
...
+            Plotly.relayout(gd, 'width', 500).then(function() {
+                expect(gd.data.length).toEqual(1);
+                expect(gd.layout.width).toEqual(500);
+                done();
+            });
+        });
+
+        it('Calling `Plotly.relayout` with object should amend the preexisting parcoords', function(done) {
+            expect(gd.layout.width).toEqual(1000);
```
VeraZab: can we add a few tests testing some of those array attributes?

monfera:
Hi @VeraZab yes #1 is the way `plotly.js` works when it's an array attribute of traces (of which there can be multiple). Re #2, it can do per column, per row, per cell or combination. The idea is that if the columns in a three-column table are `[A, B, C]`, then eg. `B` can be a scalar value or color; but it can also be an array. So you can have 3 columns, 2 rows styled as `[[A1, A2], [B1, B2], [C1, C2]]` on a per cell basis, or you can do just `[[A1, A2]]` ie. leave other columns undefined, in which case they inherit colors, so it's basically per-row styling. I'll make an example.

monfera:
Per column: as you did above
Per row: https://codepen.io/anon/pen/vePreY?editors=1010
Combined: https://codepen.io/anon/pen/XeGYzB?editors=1010

monfera:
@VeraZab this version replaces cell contents in one call, ie. not of just one column at a time: https://codepen.io/anon/pen/dVrKqb?editors=1010 - I'm adding jasmine test cases for such things, and what you're pointing out above as well.

VeraZab:
oh for #1, yeah sorry there's a bunch of things that I'm realizing are not working as expected in the workspace, so I was making the calls from the workspace console to test your pr, rather from the workspace ui, which already has those conventions pre-built, yeah I think that's fine.

> @VeraZab this version replaces cell contents in one call, ie. not of just one column at a time: https://codepen.io/anon/pen/dVrKqb?editors=1010 - I'm adding jasmine test cases for such things, and what you're pointing out above as well.

yeah, both are supported right, from what I've seen in my tests, and that's great, should be that way, and yes, let's add some tests

Ok, well, aside from a few extra tests, I'm good with this pr 💃

etpinard:
Oh shit, what Vera points out will be very tricky to fix. We can't rely on what we currently do for other 2d array trace types as table is column major.

VeraZab:
@etpinard I think I've made a mistake in that comment, about array wrapping.

I've just checked for a scatter trace, and it's true that when I want to send an update to `marker.color`, I don't just do Plotly.restyle(gd, {'marker.color': [1, 2, 3]})

I wrap the [1, 2, 3] array into another array and the correct call is
Plotly.restyle(gd, {'marker.color': [[1, 2, 3]]})

so my bad, I think this pr's good to go, after tests

Issue #2086, opened 2017-10-12
display data range for histogram hover text


chriddyp:
![screen shot 2017-10-12 at 6 41 02 pm](https://user-images.githubusercontent.com/1280389/31522766-98017968-af7d-11e7-8a19-62238e8abb99.png)

In this histogram, my data is binned by month. However, when I hover over a bar, I see the midpoint of the data. I wasn't sure whether to interpret this as the midpoint of the histogram bar or the start of the histogram bar. I would prefer if this was more explicit and said: `October 1-September 1`

alexcjohnson:
related: #69 - this would be nice for all histograms, not just on date axes.

etpinard:
I'd vote for merging this with #69

etpinard:
or maybe closing #69 and keeping this thing.

alexcjohnson:
Closing #69 and bringing the important comments here.

I said:
> What I'd prefer to do is make smarter hover text for histograms. So the axis labels (which may or may not correspond to bins) can stay as they are, unless you use `ticktext`, but the hover text will be smarter - perhaps even showing the actual range of data aggregated in the bin, which would provide more useful information than just what the precise edges of the bin are.

@mewalig replied:
> in case it is helpful to know, in my experience documenting ranges for financial services transactions, being able to show the ranges in the form of "x.00 - y.99" (to however many zeros and nines the creator wants/needs) is the most useful, because it shows not only the exact range, but also conveys the rounding approach that was used (e.g. "first round to two decimal places, then put in the corresponding bin"). I'm not saying that is what this library should default to, but it would be great if it was able to support a way to achieve that.

alexcjohnson:
Actually might be natural for me to tackle this alongside #2071

alexcjohnson:
It turns out to be super easy to have this report the *actual* data range in each bin, and there's something nice about this from a theoretical standpoint as it gives you more real information about your data, rather than just clarifying the meaning of what you can already see on the plot.

If the data in the bin span a range, you see that range:
![screen shot 2017-10-13 at 2 47 25 pm](https://user-images.githubusercontent.com/2678795/31561326-832b60fa-b025-11e7-877b-71e38de2b19d.png)

If the data in the bin are all at the same value, you just see that value:
![screen shot 2017-10-13 at 2 47 40 pm](https://user-images.githubusercontent.com/2678795/31561378-a504a86c-b025-11e7-8be6-e77c0218dcd8.png)

[...]

Issue #2071, opened 2017-10-06
Histogram events


alexcjohnson:
Currently histogram trace events (`click`, `hover`, `selecting`, `selected`) yield the *bin number* in the `pointNumber` field, and don't do anything to connect this back to the input arrays. That seems awkward for applications like crossfilter. How should we handle this?

- (1) Leave it as is, let the developer work it out.
  - pro: Small event data. Direct correspondence between the event on screen and the event data. Also inertia!
  - con: It can be cumbersome (and slow) to work out the bin contents. Also `pointNumber` isn't actually a point number (ie you can't look up items in `gd.data` from this)
- (2) Report one `point` for each item in the selected bin, and don't report the bin itself at all
  - pro: Most natural for crossfilter, as we connect directly back to the input arrays. In as far as this is what you care about, there's no special case to deal with histogram traces, they work just like other traces.
  - con: Event data could become huge and slow to create. What about empty bins, we wouldn't be reporting *anything* even though (for click and hover events) there is hover data there. Devs still may want to know what bin is being selected.
- (3) Keep the current structure, but add an array of actual point numbers to each bin data object. Perhaps bin number could move from `pointNumber` to `binNumber` and the associated input data could be an array, `pointNumbers`?
  - pro: report all the things! Doesn't have as much overhead as reporting a whole point for each item, either in object size (an array of numbers vs an array of objects with lots of keys) or in CPU (we could probably construct all these arrays of indices during `histogram.calc` with minimal overhead, and just spit them out on demand)
  - con: histogram becomes a special case for any event-handling code - that said, you won't need to switch on trace type *per se* but just on the structure of the event data you get. And other traces may follow, for example looking at https://simonbjohnson.github.io/Ebola-3W-Dashboard/ I notice the pie charts are really aggregated like histograms - separate issue on the topic forthcoming, I think this is in fact a very common use case.

I'm leaning (3) but would love to hear opinions particularly from @cpsievert @chriddyp @monfera

cpsievert:
(3) seems reasonable to me

chriddyp:
(3) works for me

monfera:
seconding (3)

etpinard:
(3) ftw :1st_place_medal:

etpinard:
ping @alexcjohnson 

looks _implements ####_ doesn't close issues automatically.

alexcjohnson:
closed by #2113

PR #2113, opened 2017-10-23
Histogram events & bin hover label improvements

alexcjohnson:
Implements #2071 (histogram events get the input points selected) and #2086 (better display of data ranges for histogram bins) for both `histogram` and the 2d histogram types (`histogram2d` and `histogram2dcontour`)


src/lib/search.js
```diff
@@ -12,6 +12,11 @@
 var isNumeric = require('fast-isnumeric');
 var loggers = require('./loggers');
 
+// don't trust floating point equality - fraction of bin size to call
+// "on the line" and ensure that they go the right way specified by
+// linelow
+var roundingError = 1e-9;
```
alexcjohnson: I'm a little surprised that this hadn't come up before (other than the test [below](https://github.com/plotly/plotly.js/pull/2113/commits/6dc99047d171ac7acea56d5471cb0cbdff2ecdb2#diff-8c89958f1286cac36ab2325921671724R426) that was actually wrong before), and obviously it's a bit questionable exactly what small fraction to put here... but if we just use straight `<` vs `<=` etc it's pretty easy to trick `findBin` on the edges.

src/components/fx/hover.js
```diff
@@ -1063,14 +1063,13 @@ function cleanPoint(d, hovermode) {
     d.y0 = Lib.constrain(d.y0, 0, d.ya._length);
     d.y1 = Lib.constrain(d.y1, 0, d.ya._length);
 
-    // and convert the x and y label values into objects
-    // formatted as text, with font info
+    // and convert the x and y label values into formatted text
     if(d.xLabelVal !== undefined) {
-        d.xLabel = ('xLabel' in d) ? d.xLabel : Axes.hoverLabelText(d.xa, d.xLabelVal);
+        d.xLabel = ('xLabel' in d) ? d.xLabel : getDimText(d, 'x');
```
alexcjohnson: @etpinard I made `getDimText` in parallel with you adding `d.(x|y)Label` and `Axes.hoverLabelText` - I suppose in principle we could use `d.(x|y)Label` for this too instead of adding `d.(x|y)LabelVal(0|1)`, maybe that would actually be simpler... let me take a look.

src/traces/histogram/calc.js
```diff
@@ -171,15 +171,19 @@ module.exports = function calc(gd, trace) {
...
             };
-            if(uniqueValsPerBin) {
-                cdi.p0 = cdi.p1 = (inputPoints[i].length) ? pos0[inputPoints[i][0]] : pos[i];
-            }
-            else {
-                cdi.p0 = roundFn(binEdges[i]);
-                cdi.p1 = roundFn(binEdges[i + 1], true);
+
+            // pts and p0/p1 don't seem to make much sense for cumulative distributions
+            if(!cumulativeSpec.enabled) {
```
alexcjohnson: A questionable decision... this gets at some of the same issues that led to `cumulative.currentbin` but in principle a CDF shows the sum of all the data prior to a specific point, not the data within that bin. perhaps I should shift `p0` and `p1` depending on `currentbin` so that the hover label at least gets the description "all the data prior to X" precisely correct, even though it'll be a different value than the center of the bar...

And then what about `pts`? You could say it should be included and again should be "all the data prior to X" but then it would be meaningless to select a single bar and not all the bars before it. Alternatively `pts` could contain "all the data that was added in this bar" which might be more useful in terms of selecting a single bar, but it's not really what that bar means.

test/jasmine/tests/histogram_test.js
```diff
@@ -664,3 +666,207 @@ describe('Test histogram', function() {
...
+        _test(0.05, 0.3, [0, 2, 4, 6], false, [0, 1.9, 2, 3.9, 4, 5.9]);
+        _test(0, 0.1, [0, 1, 2], false, [0, 0.9, 1, 1.9]);
+        _test(0, 0.001, [0, 1, 2], false, [0, 0.999, 1, 1.999]);
+        _test(0.1, 0.1, [115, 125, 135], false, [115, 124.9, 125, 134.9]);
+        _test(0.1, 0.01, [115, 125, 135], false, [115, 124.99, 125, 134.99]);
+        _test(10, 100, [5000, 6000, 7000], false, [5000, 5900, 6000, 6900]);
+
+        // too small of a right gap: stop disambiguating data at the edge
+        _test(0, 0.0009, [0, 1, 2], false, [0, 1, 1, 2]);
+        _test(0.1, 0.009, [115, 125, 135], false, [115, 125, 125, 135]);
```
alexcjohnson: @etpinard @chriddyp as discussed offline this morning - compare to lines 699 and 701 above that are just on the other side of the cutoff I implemented in https://github.com/plotly/plotly.js/pull/2113/commits/ad0e08f14bc18e3b8f5a570d5372db15696d86d0 - seem reasonable?

alexcjohnson: to be clear, what these tests are indicating is that the hover labels for these bins will be `115 - 124.99` and `125 - 134.99` on the disambiguated side of the cutoff, and `115 - 125` and `125 - 135` on the too-many-digits-to-disambiguate side.

test/jasmine/tests/histogram_test.js
```diff
@@ -664,3 +666,207 @@ describe('Test histogram', function() {
...
+        _test(0, 28 * day, [cn1_00, cn1_01, cn1_02], cn,
+            [cn1_00, Lib.dateTime2ms('2000-12-01', cn), cn1_01, Lib.dateTime2ms('2001-12-01', cn)]);
+        _test(14 * day, 14 * day, [cn1_00 - 14 * day, cn1_01 - 14 * day, cn1_02 - 14 * day], cn,
+            [cn1_00, Lib.dateTime2ms('2000-12-01', cn), cn1_01, Lib.dateTime2ms('2001-12-01', cn)]);
+
+        _test(0, 353 * day, [cn1_00, cn1_10, cn1_20], cn,
+            [cn1_00, Lib.dateTime2ms('2009-01-01', cn), cn1_10, Lib.dateTime2ms('2019-01-01', cn)]);
+        // occasionally we give extra precision for world dates (month when it should be year
+        // or day when it should be month). That's better than doing the opposite... not going
+        // to fix now, too many edge cases, better not to complicate the logic for them all.
```
alexcjohnson: The algorithm I came up with for figuring out how many digits to show is basically:
- find the minimum distance from data points to bin edges from either side - this is just two numbers for the whole distribution
- for two consecutive bins, find the largest digit that changes in both the left and right gaps (with a little shifting to account for rounding errors)

This algorithm is a little complicated to state, but it's quick to calculate and it's robust as far as I can see, on numeric, category, and gregorian date axes (though even in that case there is a weird [edge case](https://github.com/plotly/plotly.js/pull/2113/commits/6cb4e7130e4a5a1de7e5819fc6eb7c487c7997a4#diff-8c89958f1286cac36ab2325921671724R819) that confuses it). However, the larger variability of month and year lengths in world calendars causes it to sometimes think a digit doesn't change when in the actual data it does, and increasing the tolerance to catch this breaks some much more common cases.

I suspect we could fix this by looking at *all* bin edges independent of each other, but that's far more computationally intensive, and has some other potential weird pitfalls. I don't think providing a bit of extra precision in the label for a few world calendar cases justifies all that extra complexity.

src/components/fx/hover.js
alexcjohnson: yeah, better to use `d.(x|x)Label` -> https://github.com/plotly/plotly.js/pull/2113/commits/51ad8c239653ba0bb69387cefd791409fa6208db

src/plots/cartesian/axes.js
```diff
@@ -1221,7 +1221,7 @@ axes.hoverLabelText = function(ax, val) {
     var tx = axes.tickText(ax, ax.c2l(logOffScale ? -val : val), 'hover').text;
 
     if(logOffScale) {
-        return val === 0 ? '0' : '-' + tx;
+        return val === 0 ? '0' : MINUS_SIGN + tx;
```
alexcjohnson: been around for a while, using regular dash instead of unicode minus. tested in https://github.com/plotly/plotly.js/pull/2113/commits/3aa03f7911fe6988fafcc71c1451b9c9f17bb76f#diff-846cf5b534aa0d22bdd1da2b43ac3cbaR517

test/jasmine/tests/hover_label_test.js
```diff
@@ -501,6 +501,128 @@ describe('hover info', function() {
...
+            }], {
+                width: 500,
+                height: 400,
+                margin: {l: 0, t: 0, r: 0, b: 0}
+            })
+            .then(function() {
+                _hover(gd, 250, 200);
+                assertHoverLabelContent({
+                    nums: '3',
+                    axis: '3.3'
```
alexcjohnson: ⚠️  another potentially controversial decision: whenever *every* bin contains a single unique value, label exactly that value, not a range, even if those single values are not at the same places within the bins. But as soon as you have any bin with two distinct values in it, we use the range logic.

Note: for histogram2d I applied this independently to each axis, not to each brick. So if one x brick has only the point (2, 3) and one above it has only the point (2.1, 4), this will make the x bins show ranges, because *that one x bin* has multiple values, even though *each brick* has unique values in it.

etpinard: Nice touch here. I was worried that category histograms would get e.g, [`apples - lemons`](http://rickyreusser.com/plotly-mock-viewer/#hist2d_summed) bin hover labels. :ok_hand:

test/jasmine/tests/hover_label_test.js
```diff
@@ -501,6 +501,128 @@ describe('hover info', function() {
...
+                xbins: {start: -0.5, end: 8.5, size: 3},
+                type: 'histogram'
+            }], {
+                width: 500,
+                height: 400,
+                margin: {l: 0, t: 0, r: 0, b: 0}
+            })
+            .then(function() {
+                _hover(gd, 250, 200);
+                assertHoverLabelContent({
```
etpinard: Would mind checking that `plotly_hover` does emit the correct event data?

src/traces/histogram/calc.js
etpinard: > so that the hover label at least gets the description "all the data prior to X"

Yeah. I think something along those lines would be useful down the road. 

> You could say it should be included and again should be "all the data prior to X" but then it would be meaningless to select a single bar and not all the bars before it. Alternatively pts could contain "all the data that was added in this bar" 

Sounds to me like both _all-data-prior-to-X_ and _data-added-in-this-bar_ point lists could be useful, so we'll might have to emit both in the future. Leaving this discussion for another time is :ok_hand: with me. I doubt that most plotly.js users even know about cumulative histograms.

test/jasmine/tests/histogram_test.js
alexcjohnson: The issue with dates is that the actual time increments - the "digits" we're looking to see if they have changed - have different sizes in terms of milliseconds. The *bin sizes* being different should not be a problem, though people could do things to thwart the algorithm, like having the first two bin edges on fairly round values but some subsequent edges at not-so-round values. The idea that two edges are enough to determine the biggest digit that always changes comes from the observation that if *one* bin is extra-round, its neighbor will not be. So to do this robustly with nonuniform bins, we could either:
- Do this same algorithm on *all* bin edges. That's different from looking at all bin edges independently as I suggested above for world calendars, though that would solve this problem too - what I'm suggesting here is to calculate just one left-edge-gap and one right-edge-gap as we do now, but then test those gaps against all bin edges instead of the first two.
- As a last resort, make a new attribute for folks to set their own rounding.

One common case mentioned in #360 is equally-sized bins on a log axis. There, the whole assumption of a single digit to round to is wrong, and we'll have to do something completely different or just give up on the idea of disambiguating the edges, and let the regular axis-formatting-for-hover system handle it unchanged.

src/traces/histogram/calc.js
alexcjohnson: Great, I'll make a new issue  to discuss how this should work for CDFs.

test/jasmine/tests/hover_label_test.js
alexcjohnson: That *can* happen - but only if the bins *actually* contain apples and lemons.
```js
Plotly.newPlot(gd, [{
  x: ['apples','apples','apples','lemons','lemons','peaches','peaches','pears'],
  type: 'histogram',
  xbins: {start: -0.5, end: 3.5, size: 2}
}],{
  width: 400,
  height: 400
})
```

screen shot 2017-10-24 at 11 43 19 am

It's possible to put as many categories as you want into one bin, and as with numbers we'll only show the first and last ones in the label. Never happens automatically since it's in general confusing, but if you want to do it we won't stop you!

etpinard: > but if you want to do it we won't stop you!

Cool. Would you mind adding a test case for this situation?

test/jasmine/tests/hover_label_test.js
alexcjohnson: Good catch -> tested in https://github.com/plotly/plotly.js/pull/2113/commits/2696c1c9ded8fb57c61ef103f08c374dda2a4f53

test/jasmine/tests/hover_label_test.js
alexcjohnson: > Cool. Would you mind adding a test case for this situation?

https://github.com/plotly/plotly.js/pull/2113/commits/0331a16845be937de4cedac301979a0139249fa1

etpinard:
Nicely done :dancer:

Issue #1876, opened 2017-07-12
Upcoming axis auto-range improvements


etpinard:
This issue attempts to combine ideas from several individual feature requests into coherent picture (i.e. the least number of new attribute possible).

In a private conversation with @alexcjohnson , we proposed adding two new (cartesian) axis attributes: `bounds` and `boundmode`. `bounds` would expect 2-item array values similar to the current `range`. `bounds` can be thought of as representing the maximum possible extend of the axis ranges. In turn, `boundmode` would determine if `bounds` applies during the auto-range routine (i.e. on _first_ view - discuss in https://github.com/plotly/plotly.js/pull/1861#issuecomment-314518657) or during interactions as well (as proposed in https://github.com/plotly/plotly.js/issues/887). For example,

```js
xaxis: {
  autorange: true,
  boundmode: 'soft',
  bounds: [0, 100]
}
```
would restrict the auto-range logic to compute an x-axis `range` between 0 and 100. This might be useful when plotting percentages where users don't want the axis ranges to go beyond 100%.

```js
xaxis: {
  boundmode: 'hard',
  bounds: [0, 100]
}
```
would restrict the axis range to `[0,100]` in the auto-range logic **and** during interactions.

In doing so, 1-way `bounds` values such as `[null, 100]` or `[0, null]` should be allowed. `null` items here mean _don't restrict the min (or max) bound_ acting. This behavior is similar to `cmin` and `cmax` for color scales. This call signature should also be extended to the axis `range` attribute - taking care of https://github.com/plotly/plotly.js/issues/400

-----

A few more things to think about:

- Adding a `'tight'` value for axis `autorange` would take care of https://github.com/plotly/plotly.js/issues/928
- Maybe adding `'once'` value for axis `autorange` for https://github.com/plotly/plotly.js/issues/1849
- Maybe allow _span_ axis `range` (and `bounds`) values for https://github.com/plotly/plotly.js/issues/387

alexcjohnson:
One other idea we've discussed is autoranging one axis based on only the data that's *visible* on the opposite axis or axes - for example zoom in on a particular x range and autorange y based only on the x data in that range. Perhaps `autorange: 'visible'`?

petergaultney:
[...]

Issue #1774, opened 2017-06-09
Setting different color for each part of Box Plot 


alenan2016:
I would like to color the top of the box and the bottom of the box in different colors, and also top of the whisker and the bottom of the whisker in different colors. Is it possible somehow? Please see my example here: (my apologize, the example is in the Highcharts, in Highcharts I only can color top and bottom of the whisker  in different colors): https://jsfiddle.net/hew8nq5u/4/ Please show me an example if it is possible. Thank you in advance,
Best regards,
alenan2016

etpinard:
Here's what you can do now: https://codepen.io/etpinard/pen/mwVxBE?editors=0010

We're missing a few attributes:

- `whiskerthinkess`
- `whiskercolor`
- `boxmeancolor`

alenan2016:
Thank you for your interest and help.

slfan2013:
eagerly wanting this as well. thanks.

Braintelligence:
@etpinard Would love to have `whiskercolor`.

slfan2013:
@Braintelligence Maybe set `boxpoints` to `outliers` and then change the `markers`? https://plot.ly/javascript/reference/#box-boxpoints

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2116, opened 2017-10-24
Violin plots

etpinard:
[Violin](https://en.wikipedia.org/wiki/Violin_plot) plots are coming to plotly.js :tada: :violin: :tada: 

Python users can already create violins using @cldougl's [`create_violin`](https://github.com/plotly/plotly.py/blob/master/plotly/figure_factory/_violin.py) figure factory ([example](https://plot.ly/python/violin-plot/)); this PR will bring violin plots to all plotly.js consumer with an API very similar to the `box` trace type.

----

**IMPORTANT:** After the first push of 2017/10/24, this PR is very much a WIP. Several API decisions remain to be made. See the first few comments on commit https://github.com/plotly/plotly.js/pull/2116/commits/3438eae9199ef5df36a907ccb3b8b54ad16fcd6e for more info.


src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+
+    bandwidth: {
+        valType: 'number',
+        min: 0,
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Sets the bandwidth used to compute the kernel density estimate.',
+            'By default, the bandwidth is determined by Silverman\'s rule of thumb.'
+        ].join(' ')
```
etpinard: No _hard_ bandwidth. This [rule-of-thumb](https://en.wikipedia.org/wiki/Kernel_density_estimation#A_rule-of-thumb_bandwidth_estimator) that depends on the sample length and standard deviation seems pretty popular in other libraries.

src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+        items: [
+            {valType: 'any', editType: 'plot'},
+            {valType: 'any', editType: 'plot'}
+        ],
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Sets the span in data space for which the density function will be computed.',
+            'By default, the span goes from the minimum value to maximum value in the sample.'
+        ].join(' ')
```
etpinard: To address, @alexcjohnson 's concern (from a private convo):

> mmm, looking through example images I don’t actually see the behavior I was imagining… might be something to play with at some point but not for now, unless some standard existing package does it.

> One thing I see a lot in examples that seems to me a *bad* idea is truncating the violin at value where it has a finite width - particularly if that value is actually a data value. Unfortunately one of the examples I see that does exactly that is on our own site… https://plot.ly/python/basic-statistics/ out[5]

> unless there’s some physical limit to the variable… that seems like it would justify doing this, but aren’t you still throwing away area (and therefore visual weight) from the points at the end?

> at any rate, seems like if one is going to truncate like that it should be done explicitly, and should not be the default.

> Ideally I’d like you to only be able to truncate by value (per the argument about physical limits) rather than “truncate at the ends of the data range” but I imagine people will complain if they can’t do both…

etpinard: To note, `span` was inspired [seaborn](http://seaborn.pydata.org/generated/seaborn.violinplot.html)'s `cut` argument.

src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+        valType: 'number',
+        min: 0,
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Sets the bandwidth used to compute the kernel density estimate.',
+            'By default, the bandwidth is determined by Silverman\'s rule of thumb.'
+        ].join(' ')
+    },
+    scaleby: {
```
etpinard: inspired by [seaborn](http://seaborn.pydata.org/generated/seaborn.violinplot.html)'s `scale` attribute.

src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+            {valType: 'any', editType: 'plot'}
+        ],
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Sets the span in data space for which the density function will be computed.',
+            'By default, the span goes from the minimum value to maximum value in the sample.'
+        ].join(' ')
+    },
+    side: {
```
etpinard: to remake, seaborn's 

image

src/traces/box/plot.js
```diff
@@ -161,7 +161,8 @@ function plotPoints(sel, plotinfo, trace, t) {
     var bdPos = t.bdPos;
     var bPos = t.bPos;
 
-    var mode = trace.boxpoints;
+    // TODO ... unfortunately
+    var mode = trace.boxpoints || trace.points;
```
etpinard: We might want to rename `boxpoints` -> `points` as well as `boxmean` (maybe to  `showmean` + `showsd` ?)

src/traces/violin/index.js
```diff
@@ -0,0 +1,40 @@
...
+    style: require('./style'),
+    hoverPoints: require('../box/hover'),
+    selectPoints: require('../box/select'),
+
+    moduleType: 'trace',
+    name: 'violin',
+    basePlotModule: require('../../plots/cartesian'),
+    // TODO
+    // - should maybe rename 'box' category to something more general
+    categories: ['cartesian', 'symbols', 'oriented', 'box', 'showLegend'],
```
etpinard: this `box` category is currently used:

image

etpinard: ... which for the most part are common places for `violin` traces too.

src/traces/violin/layout_attributes.js
```diff
@@ -0,0 +1,19 @@
...
+'use strict';
+
+var boxLayoutAttrs = require('../box/layout_attributes');
+
+// TODO update descriptions
+// ... or maybe just reuse box* attributes?
+module.exports = {
+    violinmode: boxLayoutAttrs.boxmode,
+    violingap: boxLayoutAttrs.boxgap,
+    violingroupgap: boxLayoutAttrs.boxgroupgap
```
etpinard: Important: Should `box` and `violin` the same _gap_, _groupgap_ and even _mode_ attributes?

In other words, should `violin` be thought as a different _mode_ for `box`?

src/traces/violin/calc.js
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+'use strict';
+
+var Lib = require('../../lib');
+var boxCalc = require('../box/calc');
+
+var kernels = {
+    gaussian: function(v) {
+        return (1 / Math.sqrt(2 * Math.PI)) * Math.exp(-0.5 * v * v);
```
etpinard: We could support more:

https://en.wikipedia.org/wiki/Kernel_(statistics)#Kernel_functions_in_common_use

src/plots/cartesian/constants.js
```diff
@@ -57,18 +57,13 @@ module.exports = {
...
-    // 3. errorbars for bars and scatter
-    // 4. scatter
-    // 5. box plots
+    // Layers to keep trace types in the right order
     traceLayerClasses: [
         'imagelayer',
         'maplayer',
         'barlayer',
         'carpetlayer',
+        'violinlayer',
```
etpinard: Making sure violins are under boxes always.

src/traces/violin/index.js
```diff
@@ -0,0 +1,40 @@
...
+module.exports = {
+    attributes: require('./attributes'),
+    layoutAttributes: require('./layout_attributes'),
+    supplyDefaults: require('./defaults'),
+    supplyLayoutDefaults: require('./layout_defaults'),
+    calc: require('./calc'),
+    setPositions: require('../box/set_positions'),
+    plot: require('./plot'),
+    style: require('./style'),
+    hoverPoints: require('../box/hover'),
```
etpinard: Should violin show hover labels about the kernel density curve?

src/traces/box/defaults.js
```diff
@@ -77,4 +88,10 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     }
 
     coerce('hoveron');
+}
+
+module.exports = {
+    supplyDefaults: supplyDefaults,
+    handleSampleDefaults: handleSampleDefaults,
+    handlePointsDefaults: handlePointsDefaults
```
etpinard: I made a separate reorganisation commit, to show a new way to factor out common trace module blocks. I think this method is a little more consistent with ES6 modules. For example here, `supplyDefaults` would be the default export.

src/traces/violin/attributes.js
alexcjohnson: I love that this is two numbers (or date strings, I guess), one for each end. - that's way more flexible than seaborn's one number or ggplot2's boolean. I also love that it's data values, rather than a delta from the ends of the distribution. Between the two of those you could do what seems to be really the *right* thing with seaborn's [cut example](http://seaborn.pydata.org/_images/seaborn-violinplot-12.png) and lop off unphysical values < 0 but not restrict the upper end.

Ideally I would prefer if the default were no clipping at all (which, as discussed, probably means data bounds extended by 2 or 3 times the bandwidth), rather than clipped at the data bounds. The challenge then is to make it easy for people who do want to clip tightly to the data bounds - particularly if we think about the distribution changing over time, having to manually update the bounds for either of these cases seems awkward.

I suppose we could define two special values of `span[i]` - one for "clip tightly to this end of the data" and another for "do not clip this end"? `'tight'` and `'loose'` perhaps? Alternatively there could be another attribute for this, which would be nice as we wouldn't need to mix numbers and strings (or whatever the special value is) but it seems tricky to cover all cases this way, like if you want the low end tight and the upper end loose.

src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+        valType: 'enumerated',
+        values: ['both', 'left', 'right'],
+        dflt: 'both',
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Determines which side of the position line the density function making up',
+            'one half of a is plotting.',
+            'Useful when comparing two violin traces under *overlay* mode, where one trace.'
```
alexcjohnson: where one trace what?

src/traces/violin/attributes.js
alexcjohnson: nice! also `'top'` and `'bottom'` for horizontal violins.

src/traces/violin/calc.js
alexcjohnson: Good thinking to make this extensible down the line. There would be something nice about using one of the polynomial kernels that goes smoothly to zero at a finite position but to start just gaussian seems fine. That's the ggplot2 default anyway, and I can't find anything to say which one seaborn uses.

src/traces/violin/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+        },
+        width: {
+            valType: 'number',
+            role: 'style',
+            min: 0,
+            dflt: 2,
+            editType: 'style',
+            description: 'Sets the width (in px) of line bounding the violin(s).'
+        },
+        smoothing: scatterAttrs.line.smoothing,
```
alexcjohnson: Is `smoothing` needed? Seems weird, especially since we're just kind of making up points for the path.

src/traces/violin/index.js
alexcjohnson: It's tricky to figure out what someone would want to see labeled on the curve. The only thing I can think of that might be cool is a label that moves continuously (along the distribution axis) with the mouse, so you can look at a peak or a valley or something and see exactly what data value it's at... would help you read quantitative differences off several violins. Would a label like that get any value reported for the density? It wouldn't mean much on its own so could be omitted, though it would have meaning relative to other such values on the same or different violins.

src/traces/violin/layout_attributes.js
cldougl: I could see users wanting to use the `gap` attributes on violins the same way they would with boxes. imo it seems pretty intuitive for these charts to behave the same way and have the same grouping attributes available.

src/traces/violin/attributes.js
etpinard: > I suppose we could define two special values of span[i]

That would be dangerous, as we want the `span` items to be set in `d` coordinates. So a violin trace with `'tight'` and `'loose'` categories would totally break this.

@alexcjohnson 's `'tight'` and `'loose'` suggestions made me think of https://github.com/plotly/plotly.js/issues/1876 where the specs `axis.bounds` and `axis.boundsmode` are written down.

For consistency, I'm thinking about adding a `spanmode` attribute alongside `span` with possible values `'tight'`, `'soft'`, `'manual'`

```js
{
   spanmode: 'soft',
   span: [null, 10]
   // where the null means pick the 'soft' default value
   // i.e. data min minus 2 bandwidths
}

{
   spanmode: 'hard',
   span: [0, null]
   // where the null means pick the 'hard' default value
   // i.e. the data max.
}
```

We could also add `'tight-soft'` and `'soft-tight'` spanmode values to allow users to pick hard and soft defaults for each ends without having to set `span` at all.

src/traces/box/plot.js
etpinard: I thought about combining display attributes for box, mean line and standard dev line inside the violins in one flaglist attributes e.g. `stats: 'box+mean+sd'` or `innermode: /* */`, but when considering https://github.com/plotly/plotly.js/issues/1774 it might be best to have separate booleans e.g. `showbox`, `showmean`, `showsd` to keep things consisting with features like axis line and grids

So, unless someone opposes:

```js
showmeanline: true,
meanlinecolor: 'blue',
meanlinewidth: 1,
meanlinedash: 'dot'

showbox: true,
boxlinecolor: 'black'
boxlinewidth: 2,
boxfillcolor: 'red'
```

src/traces/violin/index.js
etpinard: Here's what I came up with:

peek 2017-10-27 14-36

alexcjohnson: In my excitement at how beautiful this effect is, I missed an important piece: you should see both the kde and the y value in that label, so you can use it to read out the exact peak/valley locations for example. Should show just enough digits that each pixel is a different y value.

src/traces/violin/attributes.js
etpinard: Done in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/attributes.js
etpinard: Replaced by scalegroup (taken from pie traces) and scalemode in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/attributes.js
etpinard: As of https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8, `side` is now an enumerated with values `'both'`, `'negative'` and `'positive'` so that the same set of values work with horizontal and vertical violins.

src/traces/box/plot.js
etpinard: Done in 6ffc379 with the help of `Lib.findPointOnPath` added in https://github.com/plotly/plotly.js/pull/2116/commits/4a40fc726cc42c161b2866a5f732b8c6b309c224

src/traces/violin/index.js
etpinard: Replaced by `box-violin` in https://github.com/plotly/plotly.js/pull/2116/commits/ea43b251428047d63afc2a332d308796700a11ac

src/traces/violin/layout_attributes.js
etpinard: Violin-specific violinmode, violingroup and violingroupgap are implemented in https://github.com/plotly/plotly.js/pull/2116/commits/7769f203a5b7697f682b6c5df19bc65677b72827

src/traces/violin/calc.js
etpinard: I added the _epanechnikov_ kernel in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/index.js
etpinard: Implemented in https://github.com/plotly/plotly.js/pull/2116/commits/bc6bc025ce3717c3e640f43eb71dd3ab2f719e01

etpinard: ... now looks like:

peek 2017-10-30 23-38

src/traces/violin/attributes.js
etpinard: Improved description in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/attributes.js
etpinard: `line.smoothing` was :hocho: in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/plot.js
```diff
@@ -0,0 +1,92 @@
...
+                        x: d.pos + bPos + (density[i].v / scale),
+                        y: density[i].t
+                    });
+                }
+                for(i--; i >= 0; i--) {
+                    lineData.push({
+                        x: d.pos + bPos - (density[i].v / scale),
+                        y: density[i].t
+                    });
+                }
```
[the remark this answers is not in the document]
etpinard: Better drawing algo in https://github.com/plotly/plotly.js/pull/2116/commits/ad51966b2c72b59633415f711349b6836a9cf3e8

src/traces/violin/calc.js
etpinard: ... and of course, I could add a few more if desired. Note that other kernels make the violins look a little less smooth than the gaussian. Perhaps this is why other libraries (e.g. seaborn and ggplot) only use the gaussian kernel for violin plots :thinking:

etpinard:
@alexcjohnson tagging this thing as `reviewable`. :checkered_flag: 

I still need to fill in a few descriptions :pencil2: , add a few tests (especially for one-sided violins) :lock:  and I think something is off in the way I'm computing `scalegroup` fields :computer: , but oh well, this thing is in a good enough state to started reviewing :mag_right:

alexcjohnson:
Beautiful! 🎵  Now we really have some music for our 💃 !

Issue #2073, opened 2017-10-06
Aggregated pie charts


alexcjohnson:
A common use case for pie charts is like histograms - aggregating all entries for a given label into a single slice, either with an explicit `values` field or without (then the value is just the count of items). I'm thinking about cases like https://simonbjohnson.github.io/Ebola-3W-Dashboard/ (the two pies in the middle are aggregating row counts by either activity or country) and the canonical "sales by region" pies where each slice is the sum of revenue within some category label.

In principle you can already do this with an `aggregate` transform, but this seems like a common enough use case (and transforms have enough drawbacks) that both we and our users would benefit from it being built in `pie` functionality.

Things to consider:
- I think it's straightforward to do this within the `pie` type, rather than making a new type (like the distinction between `bar` and `histogram`). Does this cause any problems? At least at first that would restrict it to discrete labels, I suppose down the line we could add binning but that'd be a weird thing to do with a pie.
- `arrayOk` attributes - do we need to provide eg `color` redundantly for every item? Do we just take the first value we find? Do we make some way to provide this once per slice, like a `distinctlabels` attribute and all the `arrayOk` items just map to that? I think we do NOT want to match the way we (accidentally) did it with `histogram` where these attributes map to bins (in this case that would mean the distinct labels, in whatever order they might show up)
- Event data - see #2071 

cc @etpinard @monfera

monfera:
### Lego-like construction

On one hand, bundling logic with view has benefits. A user who needs a histogram has the effortless task of just saying, `type: 'histogram'`. If she follows the docs and supplies proper attributes, she's in business.

On the other hand, we know that `histogram` is, conceptually, `binning` (a _model_) plus `bar` (a _view_). We also know that there is an infinite variety of _models_ that nicely combine with bars; we could have `waterfall`, `quantilebars` (showing count or median of quantiles) etc. so I'm a bit uneasy with the coupling of _models_ and _views_ as `histogram` does.

Yet, `histogram` is by far the most common specialized, aggregate barchart - in addition to, or second only to a plain _aggregate_ barchart. What you ask about pies is equally true for bars, ie. by analogy of `histogram`, we could have an `aggregatebar` which takes multivariate data like `parcoords` or `table` does and aggregates by some dimension.

Now it turns out that `histogram` is also an `aggregatebar` - where the aggregation is based on a tacked-on dimension we can just call `bin`.

[...]

Issue #2048, opened 2017-10-02
Custom data item aren't included in pie trace event data


etpinard:
See https://codepen.io/anon/pen/OxjXWP?editors=0011

First reported in https://community.plot.ly/t/customdata-for-pie-chart-is-undefined/6021

Bug from https://github.com/plotly/plotly.js/pull/1770

etpinard:
Oh pie hover is a big mess (cc https://github.com/plotly/plotly.js/issues/1456). 

Unfortunately, I'll have to punt this out of my TODO list for `1.31.0` :football:

alexcjohnson:
Fixed in #2117

Issue #1984, opened 2017-09-04
histogram colored bars


chriddyp:
as reported in https://community.plot.ly/t/coloring-selected-range-in-histogram/5619/1

it looks like `marker.color` is documented in the schema
![image](https://user-images.githubusercontent.com/1280389/30036523-3144cd10-9182-11e7-8e7f-cf5ce6a477d6.png)

but i couldn't get it to work
![image](https://user-images.githubusercontent.com/1280389/30036533-4a717a54-9182-11e7-90ac-361efb2fda34.png)

etpinard:
Here's one that works: http://rickyreusser.com/plotly-mock-viewer/#histogram_colorscale

But really, how is `marker.color` supposed to work for histogram where `x.length` in general doesn't match `marker.color.length`?

alexcjohnson:
right, we had already discussed removing histogram colorscale functionality, at least as implemented now - see https://github.com/plotly/plotly.js/issues/420#issuecomment-287809360 and #1500 - because it only works accidentally and depends on you explicitly making the correspondence between bins and color values. But I could certainly see supporting a histogram-specific colorscale functionality, where probably you would specify it similarly to `#histogram_colorscale` but instead of a `color` array you'd connect the color to either the `x` or the `y` values.

@chriddyp the issue with doing it the way you tried is, as @etpinard alluded, the values you want to color don't match up with the values you're binning... so it's redundant (and ambiguous - what happens when two values that get binned together have different colors?) and also generally you'd want to color by either the bin value or the bin height rather than the pre-binned values.

chriddyp:
>  but instead of a color array you'd connect the color to either the x or the y values.

Yeah, that sounds right. For more context, they would likely be using this to make these types of interactions (currently made with a bar chart):
![histogram-colored-bar](https://user-images.githubusercontent.com/1280389/30075597-3b2e6c10-9244-11e7-9bf8-c64455f7aa80.gif)

jmmease:
If `marker.color` is specified as an array of numbers then perhaps an aggregate function from the transform framework (`count`, `sum`, `mean`, etc.) could be specified (e.g. `marker.coloragg`) to be applied prior to mapping the result to a final color using `marker.colorscale`.

[...]

Issue #1963, opened 2017-08-18
Fx.hover does not work for scattergl


kyle-strait:
Hello,

I recently switched my app from using scatter to scattergl and realized that the Fx.hover method does not work correctly for that plot type. This [codepen](https://codepen.io/kyle-strait/pen/gxvzLX) shows that the change to scattergl breaks the hover call. I did not see this listed anywhere as a gl2d limitation so I wanted to get it out there. I also realize that the plan is for Fx to be removed in the future however I was hoping there may be some plan to amend the hover method to work with scattergl.

A quick look through made it seem like one of the sticking points is that the hover method is looking for a "hoveron" field in the trace which is not present for scattergl. It is then set to an empty string and the hover code moves on without looking for any points to execute the hover on.

Thanks!

etpinard:
That's correct, `Fx.hover` only works for cartesian subplots (and maybe `pie` - I don't remember).

You can try using [`Fx.loneHover`](https://github.com/plotly/plotly.js/blob/ff8ecb1b6a63559a897c2ba744cecdef0c6660a8/src/components/fx/hover.js#L122) the same [way](https://github.com/plotly/plotly.js/blob/ff8ecb1b6a63559a897c2ba744cecdef0c6660a8/src/plots/gl2d/scene2d.js#L681) our gl2d code invokes it to get the desired result.

`Fx.hover` is one our _unofficial_ methods that we'll remove in [`v2`](https://github.com/plotly/plotly.js/issues/420), maybe we should we should consider adding an official `Plotly.hover` method?

kyle-strait:
Hi @etpinard thank you for your reply.

I will give Fx.loneHover a try and see if I can get it to do what we need.

As for an official Plotly.hover method I personally think it could be very helpful. In our case we have points that are being used in multiple locations (i.e. a map and the plots) and they are linked. I am currently using the hover capabilities to perform "highlighting" when there is a hover on a point in a map that links to a point on the plot. I don't know how common of a use case that is but it is pretty key to the application we are working on.

lnunno:
It'd be nice if this was mentioned in the Plotly docs since [the official docs use Fx.hover directly](https://plot.ly/javascript/hover-events/#triggering-hover-events) with no mention of scattergl.

arlowhite:
@etpinard where is the **scattergl** code that creates hovers in compare mode?
(such that the labels don't overlap each other)
Looking at `loneHover`, I don't see how it deals with multiple points.

[...]

PR #2117, opened 2017-10-25
Pie aggregation and event cleanup

### LINKED ISSUE #1456: onClick for pies slices doesn't return any trace info
Code sample:
http://codepen.io/anon/pen/ZeLdXK?editors=0011

If i click any pie other that the biggest one ( first ) no there is no any trace info in event data. neither in points[0], nor in event.trace. Actually, there is "event.points.trace = undefined".
console.log(e) gives only:
image

alexcjohnson:
Implements #2073 and fixes #1456 

Previously if we encountered duplicate labels we would just drop all but the first one. This PR just does the natural thing and sums all the values we find for each label. Additionally, you can now make a pie chart with no `values` array at all, in which case we just count occurrences of each label.

To support this, event data has been augmented with a `pointNumbers` field, which is an array of indices of points in the input data represented by the slice. The `pointNumber` field remains, but *only* if the slice came from a single input point. Other fields have been renamed to be more consistent with events on other trace types, but for backward compatibility I kept `i` (replaced by `pointNumber` and like it, only present for non-aggregated slices) and `v` (replaced by `value`) but they are destined for removal at some point.

Since the `trace` field to link to the trace for this pie was both nonstandard and mostly broken (#1456) I replaced it with `data` and `fullData` as in other event types, while ensuring it's always available.


src/traces/pie/helpers.js
```diff
@@ -25,3 +25,16 @@ exports.formatPieValue = function formatPieValue(v, separators) {
     }
     return Lib.numSeparate(vRounded, separators);
 };
+
+exports.getFirstFilled = function getFirstFilled(array, indices) {
```
alexcjohnson: For `arrayOk` attributes, as well as slice color, I chose to take the first non-empty item corresponding to each label. I suppose in principle we could refine that further to the first *valid* item for each label?

src/traces/pie/plot.js
```diff
@@ -95,7 +95,19 @@ module.exports = function plot(gd, cdpie) {
...
+                    if(Array.isArray(hoverinfo)) {
+                        // super hacky: we need to pull out the *first* hoverinfo from
+                        // pt.pts, then put it back into an array in a dummy trace
+                        // and call castHoverinfo on that.
+                        // TODO: do we want to have Fx.castHoverinfo somehow handle this?
+                        // it already takes an array for index, for 2D, so this seems tricky.
+                        hoverinfo = Fx.castHoverinfo({
+                            hoverinfo: [helpers.castOption(hoverinfo, pt.pts)],
+                            _module: trace._module
+                        }, fullLayout2, 0);
```
alexcjohnson: @etpinard thoughts? I didn't see a reasonable way to build this into `Fx.castHoverInfo` since an array index already has a meaning there...

src/traces/pie/plot.js
```diff
@@ -37,28 +37,10 @@ module.exports = function plot(gd, cdpie) {
...
-        var rSmall = cd0.r * Math.cos(tiltRads);
-
-        var pieParts = pieGroup.selectAll('g.part')
-            .data(trace.tilt ? ['top', 'sides'] : ['top']);
-
-        pieParts.enter().append('g').attr('class', function(d) {
-            return d + ' part';
-        });
-        pieParts.exit().remove();
-        pieParts.order();
```
alexcjohnson: Man, I totally forgot that I had written all this crap for 3D pies, I thought #1152 was all of it - but now between this commit and https://github.com/plotly/plotly.js/pull/2117/commits/679ed9cf051eda44d53fa3aa3b6a031cd34ddab5 I think it's finally all gone.

src/components/fx/hover.js
```diff
@@ -155,7 +155,7 @@ exports.loneHover = function loneHover(hoverItem, opts) {
 
 // The actual implementation is here:
 function _hover(gd, evt, subplot, noHoverEvent) {
-    if((subplot === 'pie' || subplot === 'sankey') && !noHoverEvent) {
+    if(subplot === 'sankey' && !noHoverEvent) {
```
alexcjohnson: I couldn't see any purpose to having pie call `Fx.hover` other than throttling - which really shouldn't be an issue for `pie`, so I removed it from this special handling. Am I missing anything? Should we do the same for `sankey` and 🔪 this block entirely?

test/jasmine/tests/pie_test.js
```diff
@@ -522,33 +478,44 @@ describe('Test event property of interactions on a pie plot:', function() {
...
-            var point0 = futureData.points[0],
-                evt = futureData.event;
-            expect(point0.originalEvent).toEqual(evt, 'points');
-            expect(evt.clientX).toEqual(pointPos[0], 'event.clientX');
-            expect(evt.clientY).toEqual(pointPos[1], 'event.clientY');
+            checkEventData(futureData);
+        });
+
+        it('should not emit a hover if you\'re dragging', function() {
+            gd._dragging = true;
```
alexcjohnson: It occurred to me when fixing this bit of logic, it would perhaps be better to check `event.buttons` instead? Seems both cleaner and more powerful, for example it would prevent hover if you drag over a plot from somewhere else on the page. So in principle perhaps *all* our hover handlers should contain this check?

test/jasmine/tests/pie_test.js
```diff
@@ -522,33 +478,44 @@ describe('Test event property of interactions on a pie plot:', function() {
...
+            mouseEvent('mouseover', pointPos[0], pointPos[1]);
+            expect(futureData).toBeUndefined();
+        });
+
+        it('should not emit a hover if hover is disabled', function() {
+            Plotly.relayout(gd, 'hovermode', false);
+            mouseEvent('mouseover', pointPos[0], pointPos[1]);
+            expect(futureData).toBeUndefined();
```
alexcjohnson: This is also different from before, but consistent with other subplot types, I believe.

test/jasmine/tests/pie_test.js
```diff
@@ -522,33 +478,44 @@ describe('Test event property of interactions on a pie plot:', function() {
...
-                evt = futureData.event;
-            expect(point0.originalEvent).toEqual(evt, 'points');
-            expect(evt.clientX).toEqual(pointPos[0], 'event.clientX');
-            expect(evt.clientY).toEqual(pointPos[1], 'event.clientY');
+            checkEventData(futureData);
+        });
+
+        it('should not emit an unhover if you didn\'t first hover', function() {
+            mouseEvent('mouseout', pointPos[0], pointPos[1]);
+            expect(futureData).toBeUndefined();
```
alexcjohnson: kind of a weird one... but lets say you dragged over the pie from another subplot, so there was no hover event - you shouldn't get an unhover event even if you mouse up and then mouse out of that slice.

src/traces/pie/plot.js
```diff
@@ -95,68 +104,78 @@ module.exports = function plot(gd, cdpie) {
...
-                        hasHoverData = false;
+                    if(hasHoverLabel) {
+                        Fx.loneUnhover(fullLayout2._hoverlayer.node());
+                        hasHoverLabel = false;
                     }
                 }
 
                 function handleClick() {
-                    gd._hoverdata = [pt];
-                    gd._hoverdata.trace = cd0.trace;
```
alexcjohnson: We probably didn't mean to attach `trace` as a property of the *whole array* 🙈

test/jasmine/tests/pie_test.js
```diff
@@ -400,43 +407,24 @@ describe('Test event property of interactions on a pie plot:', function() {
             click(pointPos[0], pointPos[1]);
             expect(futureData.points.length).toEqual(1);
 
-            var trace = futureData.points.trace;
-            expect(typeof trace).toEqual(typeof {}, 'points.trace');
```
alexcjohnson: > We probably didn't mean to attach `trace` as a property of the *whole array* 🙈 

And yet we had a test to verify that that's exactly what we did 🤔

src/components/fx/helpers.js
```diff
@@ -100,22 +100,58 @@ exports.appendArrayPointValue = function(pointData, trace, pointNumber) {
...
+        if(pointData[key] === undefined) {
+            var val = Lib.nestedProperty(trace, astr).get();
+            var pointVal = getPointData(val, pointNumber);
+
+            if(pointVal !== undefined) pointData[key] = pointVal;
+        }
+    }
+};
+
+exports.appendArrayPointValues = function(pointData, trace, pointNumbers) {
```
etpinard: Maybe `appendArrayMultiPointValue` or `appendArrayMultiPointValues` would more typo-proof.

etpinard: ... and please add jsDOC :arrow_heading_up:

src/components/fx/helpers.js
```diff
@@ -100,22 +100,58 @@ exports.appendArrayPointValue = function(pointData, trace, pointNumber) {
...
-                pointData[key] = val[pointNumber];
+            for(var j = 0; j < pointNumbers.length; j++) {
+                keyVal[j] = getPointData(val, pointNumbers[j]);
             }
+            pointData[key] = keyVal;
         }
     }
 };
+
+var pointKeyMap = {
```
etpinard: :sparkles: at some point we might want to move this to the attribute declarations.

src/components/fx/helpers.js
```diff
@@ -100,22 +100,58 @@ exports.appendArrayPointValue = function(pointData, trace, pointNumber) {
...
+    labels: 'label',
+    values: 'value',
+    'marker.colors': 'color'
+};
+
+function getPointKey(astr) {
+    return pointKeyMap[astr] || astr;
+}
+
+function getPointData(val, pointNumber) {
```
etpinard: This thing is very similar to [`Lib.castOption`](https://github.com/plotly/plotly.js/blob/9bfbabfba9a5975edb0b6f49c625f6f7db7acd4b/src/lib/index.js#L431-L445). I wonder if we could combine them? Note that `getPointData` should not fallback to trace-wide properties.

src/components/fx/hover.js
etpinard: Looks like `sankey` does `Fx.hover` [here](https://github.com/plotly/plotly.js/blob/9bfbabfba9a5975edb0b6f49c625f6f7db7acd4b/src/traces/sankey/plot.js#L134) and [here](https://github.com/plotly/plotly.js/blob/9bfbabfba9a5975edb0b6f49c625f6f7db7acd4b/src/traces/sankey/plot.js#L191), but if you feel adventurous, sure, go ahead and patch `sankey/plot.js` to :hocho: this block.

src/traces/pie/plot.js
etpinard: Wow. Brutal. Good catch :baseball:

test/jasmine/tests/pie_test.js
```diff
@@ -376,10 +356,37 @@ describe('Test event property of interactions on a pie plot:', function() {
     beforeEach(function() {
         gd = createGraphDiv();
         mockCopy = Lib.extendDeep({}, mock);
+        Lib.extendFlat(mockCopy.data[0], {
+            ids: ['marge', 'homer', 'bart', 'lisa', 'maggie'],
+            customdata: [{1: 2}, {3: 4}, {5: 6}, {7: 8}, {9: 10}]
```
etpinard: Nice. I was wondering if `customdata` would get passed properly :tada:

test/jasmine/tests/pie_test.js
```diff
@@ -376,10 +356,37 @@ describe('Test event property of interactions on a pie plot:', function() {
...
+        expect(point.pointNumbers).toEqual([4]);
+        expect(point.data).toBe(gd.data[0]);
+        expect(point.fullData).toBe(gd._fullData[0]);
+        expect(point.label).toBe('4');
+        expect(point.value).toBe(5);
+        expect(point.color).toBe('#1f77b4');
+        expect(point.id).toEqual(['maggie']);
+        expect(point.customdata).toEqual([{9: 10}]);
+
+        // for backward compat - i/v to be removed at some point?
```
etpinard: Blocking: yeah, we should remove that in `v2` - would you mind adding it to #420 ?

test/jasmine/tests/pie_test.js
etpinard: Yes this is how it should be.

For the record, to get hover event data w/o visible hover labels, one should set `hoverinfo: 'none'` in all the traces of interest.

test/jasmine/tests/pie_test.js
```diff
@@ -376,10 +356,37 @@ describe('Test event property of interactions on a pie plot:', function() {
...
         mockCopy = Lib.extendDeep({}, mock);
+        Lib.extendFlat(mockCopy.data[0], {
+            ids: ['marge', 'homer', 'bart', 'lisa', 'maggie'],
+            customdata: [{1: 2}, {3: 4}, {5: 6}, {7: 8}, {9: 10}]
+        });
     });
 
     afterEach(destroyGraphDiv);
 
+    function checkEventData(data) {
```
etpinard: A long-overdue test abstraction. Thanks!

src/traces/pie/plot.js
etpinard: Maybe something like:

```js
Fx.castHoverinfo = function(trace, fullLayout, pt) {
   if(pt.pointNumber) {
     // as currently
   } else if(pt.pointNumbers) {
     // new pie logic
   }
}

// instead of
Fx.castHoverinfo = function(trace, fullLayout, ptNumber) {
  // ...
}
```

but your hack (with that comment) is fine :+1:

src/traces/pie/helpers.js
etpinard: Picking the first non-empty item is fine :+1:

test/jasmine/tests/pie_test.js
etpinard: Which by the way :white_check_mark: https://github.com/plotly/plotly.js/issues/2048

etpinard:
I should reference https://github.com/plotly/plotly.js/issues/1963 in this PR as now `pie` is locked into a hover pattern outside the realm of `Fx.hover`. Maybe a better solution to #1963 is persistent `hover` point list as discussed in https://github.com/plotly/plotly.js/issues/1848

src/traces/pie/plot.js
alexcjohnson: I think I should leave it as is for now, but if we unify this behavior with histograms (see #1984) then it may be worth making this change.

test/jasmine/tests/pie_test.js
alexcjohnson: that will need a bunch more testing and ideally rolling out to all subplots at once, so I'll defer it -> #2120

src/components/fx/helpers.js
alexcjohnson: For the multipoint case it's nice to keep the `nestedProperty` call out of this function, as that would slow it down dramatically. Between that and the fact that, as you mention, this one does not return trace-wide properties (it would return `undefined` but in fact it never gets there as it's iterating over `arrayAttrs`) these functions seem fairly distinct - though perhaps `Lib.castOption` could call this one?

src/components/fx/hover.js
alexcjohnson: simplified sankey -> https://github.com/plotly/plotly.js/pull/2117/commits/e08e37ae58296d0abcedb6f02f21f9153bb6676b
Also: I notice that the way we had it was only throttling *event emission*, not the actual item selection which is normally the slow part. Again, throttling shouldn't be an issue for either of these trace types, this is just more confirmation of that.

src/components/fx/helpers.js
alexcjohnson: good call on `appendArrayMultiPointValues ` -> with docs in 
https://github.com/plotly/plotly.js/pull/2117/commits/4ba02497d920b1ff77cb0c878aff8209120104e1

src/traces/pie/defaults.js
```diff
@@ -19,13 +19,14 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
 
     var vals = coerce('values');
-    if(!Array.isArray(vals) || !vals.length) {
-        traceOut.visible = false;
-        return;
-    }
-
     var labels = coerce('labels');
     if(!Array.isArray(labels)) {
+        if(!Array.isArray(vals) || !vals.length) {
```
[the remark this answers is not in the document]
alexcjohnson: sure -> https://github.com/plotly/plotly.js/pull/2117/commits/ee1f8c4f776fbc8dff26c4f630a6669d5e0b3df9
It was done in the image test I added, but nice to have it in jasmine too.

test/jasmine/tests/pie_test.js
alexcjohnson: commented https://github.com/plotly/plotly.js/issues/420#issuecomment-339475968

src/components/fx/hover.js
etpinard: https://github.com/plotly/plotly.js/commit/e08e37ae58296d0abcedb6f02f21f9153bb6676b is very nice. :ok_hand:

src/components/fx/helpers.js
etpinard: Right, it might be nice to move `getPointData` to `Lib` at some point, but let's keep this as is for now.

etpinard:
Lovely feature and solid clean up: looking great :dancer: 

Make sure to close #2073, #1456 and https://github.com/plotly/plotly.js/issues/2048

Issue #2119, opened 2017-10-25
updatemenus are drawn below colorscale


gdkrmr:
Similar issue to here #1302

![screenshot_2017-10-25_17-33-45](https://user-images.githubusercontent.com/12512930/32007867-11ff0bf4-b9ab-11e7-8650-34cb98d88e8c.png)

alexcjohnson:
Good catch, these should definitely go on top.

gdkrmr:
btw: the same happens with legends.

alexcjohnson:
> btw: the same happens with legends.

@gdkrmr Do you mean updatemenus go below legends (which, if so, will be fixed by #2121) or that legends go below colorscales (for which neither option will satisfy everyone, normally you just want to keep them from overlapping each other)?

gdkrmr:
the drop down menu shows below the legend

![screenshot_2017-10-26_17-31-22](https://user-images.githubusercontent.com/12512930/32062055-a5324d0a-ba73-11e7-8e98-b90a0d6d349b.png)

alexcjohnson:
> the drop down menu shows below the legend

Great, #2121 fixes that.

alexcjohnson:
closed in #2121

PR #2121, opened 2017-10-25
Layer fix

### LINKED ISSUE #2110: Plotly shape always drawn under the trace when the plot type is pie
Code reference:
https://codepen.io/dipan66/pen/dVEXOy

In this, we are trying to draw a gauge chart by drawing a needle shape over a pie. The needle is always being drawn below the trace, though the layer attribute is set to "above". Need help / thoughts on how we can go about and get the needle shape to be drawn above the pie. 

Se are using Plotly.js v1.25.0

alexcjohnson:
fixes #2110 and #2119 
- put `layer-above` at the bottom of `toppaper` so shapes (and images) in this layer end up above *all* trace types
- put `updatemenus` into their own layer - `menulayer`, naturally, because the real reason they need to be on top is that they contain hidden elements that, whenever and wherever they are shown, need to take precedence over everything else.


src/components/shapes/draw.js
```diff
@@ -57,7 +57,7 @@ function draw(gd) {
 function drawOne(gd, index) {
     // remove the existing shape if there is one.
     // because indices can change, we need to look in all shape layers
-    gd._fullLayout._paper
+    gd._fullLayout._paperdiv
```
alexcjohnson: Really we should just move shapes (and annotations for that matter) to the standard d3 update pattern like images already use (and which did not require this kind of change to adapt to the changed layering)... but that's a bigger project than I had an appetite for right now!

src/components/updatemenus/draw.js
```diff
@@ -54,7 +54,7 @@ module.exports = function draw(gd) {
      */
 
     // draw update menu container
-    var menus = fullLayout._infolayer
+    var menus = fullLayout._menulayer
```
etpinard: Ha I guess that does it. I first thought about splitting the updatemenu buttons from their header `<g>` i.e. the header would have remained in `<g infolayer>` and the buttons moved to `<g menulayer>`.

Can anyone think of a situation where one would want to place an updatemenu below other `<g infolayer>` features?

src/components/shapes/draw.js
etpinard: > but that's a bigger project than I had an appetite for right now!

Fixing two :beetle: in one PR, but hungry for more :hamburger:

src/components/updatemenus/draw.js
alexcjohnson: I can't see when you'd want the updatemenu below anything else, but I guess it could be useful to separate `<g menulayer>` itself into two pieces, so that regardless of the order you draw the menus, when they open up the transient parts will be on top. The only case I can think of that would really *need* this is if you have two menus that each overlaps the other when opened - one above that opens downward and one below that opens upward for example. If you have a lot of menus that might become necessary at times.

alexcjohnson: Updatemenus were already smart enough to put the dropdown buttons (but not always-visible buttons) on top - exactly what I was suggesting 🎉 🏆 But there was a bug with this that if a new menu was added when other menus already existed, this would get added on top of the dropdown buttons -> fixed in https://github.com/plotly/plotly.js/pull/2121/commits/e0a10af575aa864365248a1e210961fb02bf8092

src/components/updatemenus/draw.js
```diff
@@ -97,6 +97,9 @@ module.exports = function draw(gd) {
 
     // remove exiting header, remove dropped buttons and reset margins
     if(headerGroups.enter().size()) {
+        // make sure gButton is on top of all headers
+        gButton.node().parentNode.appendChild(gButton.node());
```
etpinard: Nicely done.

PR #2122, opened 2017-10-26
harden locationToFeature against non-string country names

alexcjohnson:
Before, with `locationmode: 'country names'`, non-string locations would throw an error:

```
Uncaught TypeError: Cannot read property 'trim' of null
    at countryNameToISO3 (geo_location_utils.js:54)
```

Now they're just ignored silently - as opposed to strings that are unrecognized that generate a console warning. If it's just empty it seems clear you don't even want a warning. I suppose if it's not empty but also not a string it could be nice to still generate a warning, though that's not how most of plotly.js handles invalid data...

cc @etpinard

### DIFF
--- a/src/lib/geo_location_utils.js
+++ b/src/lib/geo_location_utils.js
@@ -23,6 +23,8 @@ var locationmodeToIdFinder = {
 };
 
 exports.locationToFeature = function(locationmode, location, features) {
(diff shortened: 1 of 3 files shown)


etpinard:
Beautifully done. Thanks :dancer:

PR #2126, opened 2017-10-26
Transform inverse mapping

monfera:
- generating original points for `transform` results so that, upon interaction such as crossfiltering, the original (constituent) points can be recovered
- cascading these points through a (purportedly) arbitrary sequence of `filter` and `aggregation` steps (`groupby` not yet supported)
- adding test cases for these (also, we didn't have heterogeneous `transforms` testing with both `filter` and `aggregate` in the past)

There's still so much to do (PS: now documented the follow-up issue https://github.com/plotly/plotly.js/issues/2128); eg. more test cases; covering `groupBy` and `sort`. There's plenty of things to iterate on here.

On the pro side, the code change is rather small in terms of lines; also, the logic didn't need to change. So it's an add-on patch rather than anything heavy.

### DIFF
--- a/src/transforms/aggregate.js
+++ b/src/transforms/aggregate.js
@@ -221,9 +221,9 @@ exports.calcTransform = function(gd, trace, opts) {
     var indexToPoints = {};
     var groupings = [];
 
-    var prevTransforms = trace.transforms.filter(function(tr) {return tr.indexToPoints;});
+    var prevTransforms = trace.transforms.filter(function(tr) {return tr._indexToPoints;});
     var originalPointsAccessor = prevTransforms.length ?
-        function(i) {return prevTransforms[prevTransforms.length - 1].indexToPoints[i];} :
+        function(i) {return prevTransforms[prevTransforms.length - 1]._indexToPoints[i];} :
         function(i) {return [i];};
 
     for(i = 0; i < groupArray.length; i++) {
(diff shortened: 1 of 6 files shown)


src/transforms/aggregate.js
```diff
@@ -215,20 +215,34 @@ exports.calcTransform = function(gd, trace, opts) {
...
+            indexToPoints[groupIndices[vi]] = originalPointsAccessor(i);
+        }
+        else {
+            groupings[groupIndex].push(i);
+            indexToPoints[groupIndices[vi]] = indexToPoints[groupIndices[vi]].concat(originalPointsAccessor(i));
         }
-        else groupings[groupIndex].push(i);
     }
 
+    opts.indexToPoints = indexToPoints;
```
alexcjohnson: Since this isn't a real attribute but we're putting it back into `fullData` lets give it an `_` - ie `opts._indexToPoints`.

test/jasmine/tests/transform_multi_test.js
```diff
@@ -277,8 +277,60 @@ describe('multiple transforms:', function() {
...
+            expect(gd._fullData[0].y).toEqual([9]);
+            expect(gd._fullData[0].marker.size).toEqual([10]);
+
+            expect(gd._fullData[0].transforms[0].indexToPoints).toEqual({0: [1], 1: [3], 2: [4]});
+            expect(gd._fullData[0].transforms[1].indexToPoints).toEqual({0: [1, 3], 1: [4]});
+            expect(gd._fullData[0].transforms[2].indexToPoints).toEqual({0: [4]});
+
+            done();
+        });
+    });
```
alexcjohnson: Thanks for testing this! Nice to see `indexToPoints` cascade through the transforms.

PR #2135, opened 2017-10-31
Persistent point selection

rreusser:
This PR addresses #1848. There is useful code, but it's not yet fully functional. In particular the things that need doing:

1. invoke `plot` to make the draw updates happen. I think the best thing, at least for scatter, is to piggyback on the transition capability that bypasses a full replot. Many properties can be handled via that pathway for scatter. For non-scatter traces, I'm not 100% certain what the right solution is yet. We might need `module.drawSelect` to mutate the DOM colors/opacities/etc directly instead of calling a full replot and in order to keep, say, bar drawing logic inside the bar trace (as opposed to worrying about drawing bars within the cartesian plot select code)
2. Make `selectedids` work the same as `selectepoints`. Right now that's partially omitted.

/cc @etpinard


alexcjohnson:
scatter and some other types have a `style` method, and a corresponding `editType: 'style'` that gets picked up by `restyle` to short-circuit a full redraw. I'd think it would be pretty good performance just to use that pathway - and that would make it easy to put deselected points back to the normal style.

rreusser:
@alexcjohnson — @etpinard and I talked a bit about this and decided the best option for a start is to style them directly. I was concerned we'd lose out on a lot of properties, but if we start with just opacity and color, then direct styling (as opposed to a full `plot` pass) seems best.

etpinard:
Here's what I come up with:

https://github.com/plotly/plotly.js/compare/persistent-point-selection-etienne

- `selectPoints` long longer update DOM nodes. It now only calls `polygon.contains` and updates the `selected` calcdata accordingly
- DOM node updates are now in the `style` method
- Selection are made persistent via @rreusser 's `updatedSelectedState` in `cartesian/select.js` and `scatter/apply_selection_to_calcdata.js` where the `selected` calcdata fields are filled in. 
- Note that I made `Scatter,style` accept single-trace selection (as opposed to graph-wide only) for :racehorse:

rreusser:
@etpinard Feel free to file a separate PR or just seize control of this PR.

etpinard:
Ok, tagging this as `status: reviewable`, even though test coverage is limited and some tests are failing on CI. I'd recommend reviewing this thing commit-by-commit. I've put down some addition info in each commit message.

peek 2017-11-02 15-36

In brief, much of the work here was put into replacing the DOM node styling calls in the trace module `selectPoints` methods by `selected`-and -`unselected`-attribute-dependent node updates in the trace module `style` methods - which are now called in every `plotly_selecting` interactions as well as on `plotly_deselect`.

For most trace types, `marker.opacity`, `marker.color` and `textfont.color` have now `selected` and `unselected` flavours. `marker.opacity` and `textfont.color` mimic the currently dimming `plotly_selecting` updates, while `marker.color` adds the ability to highlight points using an arbitrary color (which should make @monfera :smiley: )

I've use `selectedpoints` as the attribute which list the indices of the coordinates in `gd.data` that are selected. Perhaps, `selectedindices` would be better. Thoughts? Note that `selectedpoints: []` means an empty selection (i.e. all pts are styled using the `unselected` attributes) and `selectedpoints: undefined` means no selection at all (i.e. all pts are styled using the normal trace-level attributes). Oh and I should add: `Plotly.restyle(gd, 'selectedpoints', [[/* new selected indiced */]])` just works<sup>TM</sup>.

In https://github.com/plotly/plotly.js/issues/1848#issuecomment-337244988, there was also talks about adding a `selectedids` attributes which would look into the traces `ids` array to determine the selected indices. I was thinking about skipping this part in this PR, unless someone really wants it included. 

cc @monfera @dfcreative @alexcjohnson @bpostlethwaite @cpsievert

src/components/drawing/index.js
```diff
@@ -426,6 +425,35 @@ drawing.pointStyle = function(s, trace, gd) {
...
+
+    s.style('opacity', function(d) {
+        return d.selected ?
+            (selectedAttrs.marker || {}).opacity :
+            (unselectedAttrs.marker || {}).opacity;
+    });
+
+    // which is slightly different than:
+    // ((d.mo + 1 || opacity + 1) - 1) * (d.dim ? DESELECTDIM : 1);
+    // in https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/select.js
```
alexcjohnson: This means that if marker opacity is an array, you won't be able to keep that during selection. I feel like we can do better than that:
- Only give `[un]selected.marker.opacity` a default value if you don't set any other `[un]selected` attributes - so if you want to keep opacity intact but just change marker color during selection, you can do that.
- If only one of the `[un]selected` opacities is set, the other state should revert to the `arrayOk` logic `(d.mo + 1 || marker.opacity + 1) - 1`
- The same is true for colors: if only one is used and the base attribute is an array, the other state should revert to the `arrayOk` logic - perhaps we want to stash this color in `d` though, as this logic is a bit more involved? Anyway, that would allow selections like in parcoords (deselected is gray, selected keeps its color) or like our S&P notifiers (deselected keeps its color, selected is bright red)
- And for both opacity and color, if neither is set you don't even need to do that loop.
- Another minor 🐎  - this will probably naturally happen with the above, but you can move `(selectedAttrs.marker || {}).opacity` calculation out of the `.each` loop - also `smc` and `usmc` (Semper Fi!) below. That said I'd love it if some day these attributes all become `arrayOk` in which case they'd have to move at least partially back into the loop... But that's not for this PR, single-value is great for now.
- My suggestions may make a mess with gradients... perhaps for now if you have a gradient and only one selection color we just revert to uniform `marker.color` for the other state?

monfera:
The bar looks good on initial render except where I have a small question:
image
The `selectedpoints` is `[0, 1, 4]` but the last highlighted orange bar is of index 5. The reason is that the `]3.5, 4.5[` filter transform eliminates one point. So I just wanted to make sure it's OK that post-filter point indices are used rather than pre-filter.

In the second bar trace, ie. last trace, both `selected` and `unselected` are specified as orange. When clearing the selection, it'll be all green but iirc there's some todo item for bars, perhaps this.
tmp
That the legend also shows both traces in green is probably the same thing.

etpinard:
> The selectedpoints is [0, 1, 4] but the last highlighted orange bar is of index 5. The reason is that the ]3.5, 4.5[ filter transform eliminates one point. So I just wanted to make sure it's OK that post-filter point indices are used rather than pre-filter.

Good catch. Thanks! I think `selectedpoints` should correspond to the pre-transform indices. What do you think?

> In the second bar trace, ie. last trace, both selected and unselected are specified as orange. When clearing the selection, it'll be all green but iirc there's some todo item for bars, perhaps this.

This is the desired. When clearing the selection, bar should show up in `marker.color` i.e. ignoring `selected.marker.color` and `unselected.marker.color`.

etpinard:
> The selectedpoints is [0, 1, 4] but the last highlighted orange bar is of index 5. The reason is that the ]3.5, 4.5[ filter transform eliminates one point. So I just wanted to make sure it's OK that post-filter point indices are used rather than pre-filter.
Good catch. Thanks! I think selectedpoints should correspond to the pre-transform indices. What do you think?

But perhaps making `selectedpoints` match post-transform indices would be best from a performance :racehorse: standpoint. I have a feeling that mapping selection (which consider post-transform traces only) to pre-transform indices will slow down `plotly_selecting` interactions. Users interacting with transforms and `selectedpoints` will probably be inclined to use `ids` (and `selectedids`) anyhow.

alexcjohnson:
> > I think selectedpoints should correspond to the pre-transform indices. What do you think?

> But perhaps making `selectedpoints` match post-transform indices would be best from a performance :racehorse: standpoint.

I think we should try to get back to the input indices. Determining the input indices shouldn't be too hard - `calcdata` should have that information now. Efficiently getting back from input indices to the post-transform indices may require caching something about the mapping in that direction though.

This raises a question though of what to do when only one or a few of the indices that led to an aggregated point are selected - if you set `selectedpoints` via crossfilter or something. I guess at this point if any part is selected, the result is selected?

monfera:
Really neat, it's incredibly useful that even in the absence of `selected` / `unselected` attributes, it defaults to the standard selection styling, which is the use of 0.2 for unselected marker opacity. It means that all preexisting plots benefit, library using and workspace / dashboard alike. It also means that `crossfilter` users don't need to manually spec these if they're happy with the default styling.

When removing these attributes, the bars show up in green, it's probably the same thing as above but just in case:
image

monfera:
> This is the desired. When clearing the selection, bar should show up in marker.color 

Haha thanks, I missed that attrib on 1st sight, explains some of my other screenshots too.

> I think we should try to get back to the input indices. Determining the input indices shouldn't be too hard

Re the availability of info, there can be an arbitrary sequence of `transform`s and via the recent https://github.com/plotly/plotly.js/pull/2126 we have the original points in `_.fullData`, at least for `aggregate` and `filter` (`sort` and `groupby` will have to be covered).

Re the question: `crossfilter` must be aware of the original points, which was the reason for the above PR. The reason is that selecting in one plot (eg. a `scatter`) will have to be reflected in another plot (e.g. a `bar`) that has aggregations, so the bars can be shorter or taller.

So I could eliminate some code from `crossfilter` if permanent selections used the original points, because then we could just `Plotly.restyle(gd, 'selectedpoints', [...])` the attribute in those plots that need to reflect a filtering in some other (originating) plot. We can't do so otherwise because the it'd only work in the special case that dashboard plots in a crossfiltered ensemble have no `transform`s.

As I think handling aggregates like this is not prepared for in this PR, ie. when `Plotly.restyle(gd, 'selectedpoints', [...])` a plot with `transform`s, a bar will either be selected or not, but it won't be shorter or longer as there are fewer points in an aggregate group.

So if this reuse is not possible right now, maybe it's more expedient to stick to post-transform points but I'm not too certain about it from the PoV of other usages, maybe @chriddyp as an originator of the linked issue can comment?

monfera:
> This raises a question though of what to do when only one or a few of the indices that led to an aggregated point are selected - if you set selectedpoints via crossfilter or something. I guess at this point if any part is selected, the result is selected?

In `crossfilter`, it's neither, but the usage is different: all the original points still show up on a `bar`, and a new trace is temporarily added that shows eg. the new aggregate results.

It looks like persistent selections can't be relied on by `crossfilter` for restyling (two of the above items already make it so and there may be more) BUT it's still incredibly useful as `crossfilter` can eventually pluck values from the `selected` / `unselected` attributes so styling is consistent. For this reason I'd like if `_fullData` had these attributes with defaults (e.g. 0.2 opacity for `unselected`) which is probably the case already. Now `crossfilter` just hardcodes such styling.

monfera:
The `selected` / `unselected` attributes in the mock are `marker: {...}` which makes sense because markers are selected or not. What happens if we're styling something else, e.g. `trace` level stuff like `[trace.]opacity` or even something like `x`?

chriddyp:
> So if this reuse is not possible right now, maybe it's more expedient to stick to post-transform points but I'm not too certain about it from the PoV of other usages, maybe @chriddyp as an originator of the linked issue can comment?

I'm a little lost in the discussion as I think it's relating to transforms? If so, I don't have an opinion, Dash doesn't use transforms or aggreations atm.

alexcjohnson:
I think we should start out (this PR) with my original suggestion: if *any* input point in a given aggregated output point is selected, the whole output point is selected. But then we should have a discussion about how best to display partial selections going forward.

@chriddyp (and @jackparmer you may have opinions here too): with transforms (and a few non-transformed types like histogram and now pie #2117) a single displayed element may correspond to multiple input points. So if you ask to select some input points, but not all the points in a given displayed element, how is that displayed element drawn?

Seems like this may need to be trace-type-dependent - and maybe even configurable. For example, scatter points might just have three states they can be drawn in: deselected, selected, and partially-selected; seems like continuously variable selectedness for scatter is unlikely to make sense. But in some cases (like if the point shows the average of a bunch of data, with size being the count or sum of sizes) you might want to show the selected and deselected components as two separate points.

For histograms, or aggregated bars or pies, seems like generally you'd want to show the aggregation results for just the selected points, overlaid on the full results - which most of the time would mean the bar/slice gets divided into a selected and a deselected part, but for some aggregations (min, max) the result could be the whole bar even if it's partially selected. And for others still (avg) the result could be bigger than the original bar.

Partially-selected choropleth regions? Dunno... try to fill it up from the bottom to the required area fraction?

etpinard:
> This raises a question though of what to do when only one or a few of the indices that led to an aggregated point are selected

and the same logic works for `groupby` transforms? I _think_ so, but I wouldn't mind having someone else confirm.

alexcjohnson:
> and the same logic works for groupby transforms?

groupby maps *points* 1:1 - but the output points can be in different traces even though the input points are all from the same trace. So I don't think there's any issue there, it's only aggregate transforms and self-aggregating traces that cause problems.

monfera:
@alexcjohnson on the 1:1 mapping w/ trace split for `groupby`, there's also the composition of various types of transforms, some of them repeated and/or in multiple places in the sequence, might complicate things a bit. They compose in nontrivial ways.

src/plots/cartesian/select.js
```diff
@@ -221,6 +225,7 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
                     searchInfo.selectPoints(searchInfo, false);
                 }
 
+                updateSelectedState(gd, searchTraces);
                 gd.emit('plotly_deselect', null);
```
etpinard: @alexcjohnson The failing test on CI

image

can be fixed by moving `updateSelectedState` after `gd.emit`. I don't think this is desired behavior though. We should emit `plotly_selected` and `plotly_deselect` after `module.style()` is completed.

So, there must a race condition somewhere in `select_test.js`. Maybe we're missing a `Lib.clearThrottle` call?

src/components/drawing/index.js
etpinard: > Only give [un]selected.marker.opacity a default value if you don't set any other [un]selected attributes - so if you want to keep opacity intact but just change marker color during selection, you can do that.

But then `[un]selected.marker.opacity` won't appear in `_fullData`, and @monfera :smirk: won't be able to populate his style panels. Personally, I think keeping the same default selection behavior is more important, so thanks @alexcjohnson for pointing this out!

> My suggestions may make a mess with gradients... perhaps for now if you have a gradient and only one selection color we just revert to uniform marker.color for the other state?

Good point here. I haven't tested any of the mocks with `marker.gradient` turned on.

etpinard: Revised logic in 1859256 and d3cdd6f with tests.

src/plots/cartesian/select.js
etpinard: Shamefully fixed via https://github.com/plotly/plotly.js/pull/2135/commits/6437081d71ccaa5d144d9ee63b327625f643509e

etpinard:
Ok, I think I'm now happy with this PR functionality-wise :rocket: as well as test-coverage-wise :lock: 

There's a lot of things going on here, so let me recap what happened since https://github.com/plotly/plotly.js/pull/2135#issuecomment-341556479:

- `[un]selected.marker.size` support has been added for `scatter`, `scattergeo`, `scatteternary` and `scattercarpet`
- PR https://github.com/plotly/plotly.js/pull/2163 got merged into this branch, making `selectedpoints` work correctly for histograms and transformed traces.
- An attempt at resolving @alexcjohnson 's https://github.com/plotly/plotly.js/pull/2135#discussion_r148815836 was made  in 1859256 and d3cdd6f 
- `selectedids` got :hocho: for now, I'll make a note of it in https://github.com/plotly/plotly.js/issues/1848 after this is PR is merged
- got the tests to :white_check_mark: 
- added mucho tests :snake:

monfera:
I looked into this grand PR two weeks ago and a lot of improvements were put in place since, and it's the single biggest step toward fully representing plot state, I'm voting for 💃 

Caution: we shouldn't immediately use it in `streambed` as the `crossfilter` restyling needs to be updated. Currently, resetting the `crosssfilter` filter on a plot will still leave it with deemphasized points, I guess because of the permanence of the selection:

permanent

So, putting it in `streambed` immediately wouldn't be good for `crossfilter`. cc @jackparmer

jackparmer:
@etpinard Let's keep this off streambed for a few weeks if possible

src/traces/box/calc.js
```diff
@@ -200,7 +200,13 @@ function calcSelection(cd, trace) {
...
         for(var i = 0; i < cd.length; i++) {
             var pts = cd[i].pts || [];
-            Lib.tagSelected(pts, trace);
+            var ptNumber2cdIndex = {};
+
+            for(var j = 0; j < pts.length; j++) {
+                ptNumber2cdIndex[pts[j].i] = j;
+            }
+
+            Lib.tagSelected(pts, trace, ptNumber2cdIndex);
```
etpinard: Quick fixup. I noticed that `selectedpoints` -> calcdata was off for box and violins while making

peek 2017-11-17 14-26

```js
var gen = v => Math.round(Math.random(v) * 10)
var n = 100
var x = new Array(n).fill(0).map(gen)
var y = new Array(n).fill(0).map(gen)
var c = '#984ea3'
var usmo = 0

Plotly.newPlot(gd, [{
  mode: 'markers',
  x: x,
  y: y,
  marker: {color: c},
  unselected: { marker: {opacity: usmo} }
}, {
  type: 'violin',
  name: 'x',
  x: x,
  points: 'all',
  pointpos: 1.5,
  box: {visible: true},
  yaxis: 'y2',
  marker: {color: c},
  unselected: { marker: {opacity: usmo} }
}, {
  type: 'violin',
  name: 'y',
  y: y,
  points: 'all',
  pointpos: 1.5,
  box: {visible: true},
  xaxis: 'x2',
  marker: {color: c},
  unselected: { marker: {opacity: usmo} }
}], {
  xaxis: {
    domain: [0, 0.48]
  },
  yaxis: {
    domain: [0, 0.48]
  },
  xaxis2: {
    domain: [0.52, 1]
  },
  yaxis2: {
    domain: [0.52, 1]
  },
  showlegend: false,
  dragmode: 'lasso',
  hovermode: 'closest',
  width: 500,
  height: 500
})
gd.on('plotly_selected', d => {
  if(d && d.points && d.points.length) {
    var sel = d.points.map(p => p.pointIndex)
    Plotly.restyle(gd, 'selectedpoints', [sel.slice(), sel.slice(), sel.slice()])
  } else {
    Plotly.restyle(gd, 'selectedpoints', null)
  }
})
```

jackparmer:
One question that came up in the workshop is whether this will work with animations. E.g., will points selected with the persistent selections API stay selected through frame transitions?

etpinard:
> will points selected with the persistent selections API stay selected through frame transitions?

yes

alexcjohnson:
Tons of great stuff here, nice job! 💃

PR #2139, opened 2017-11-02
Add class to `colorbar`; extract out class names

monfera:
See the separate commits.

### DIFF
--- a/src/components/colorbar/constants.js
+++ b/src/components/colorbar/constants.js
@@ -1,0 +1,26 @@
+/**
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
+    cn: {
+        colorbar: 'colorbar',
+        cbbg: 'cbbg',
+        cbfill: 'cbfill',
+        cbfills: 'cbfills',
+        cbline: 'cbline',
+        cblines: 'cblines',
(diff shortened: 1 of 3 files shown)


etpinard:
Can I ask why? Looks good though.

monfera:
@etpinard `crossfilter` mostly operates via the regular API, eg. `Plotly.addTraces`for adding a salient trace that shows the selected points (will likely switch to the persistent selection for this), but not always. For example, adding a salient trace will duplicate the `colorbar`, but the original one shouldn't be shown. It is possible to `Plotly.restyle` the original trace to make the `colorbar` go away, and another `Plotly.restyle` to make it come back. These calls are expensive and the dashboard is quite slow as it is. To meet functional needs at decent speed, there are some shortcuts, mostly, setting the `opacity` of something to zero and back. These calls are infinitely fast, just one `selection.style`.

While this relies on implementation detail - the scenegraph - , it's unexposed and internal to the `crossfilter` code ie. we can trivially remove these once there is a fast, single-call API (such as `Plotly.react` as you suggest) that can take care of all changes with no duplicated overhead.

In the process, I saw a missing classname and thought it's good practice to add a class name anyway (and extract out the class names). As a matter of convention I automatically add class names to scenegraph elements because not doing so might introduce shallow, avoidable gotchas during dev, and makes the code more fragile, eg. when changing the code. Relying on SVG tags eg. `circle` is not as good because, in the future, we may always replace a terminal node with a `g` and there goes some selection that might have relied on tags.

etpinard:
:dancer:

PR #2140, opened 2017-11-02
Add multiselect

dfcreative:
Reproduces #1853

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -942,11 +942,13 @@ function showSelect(zoomlayer, dragOptions) {
         paths.push(ppts.join('L') + 'L' + ppts[0]);
     }
 
-    outlines.enter()
-        .append('path')
-        .attr('class', function(d) { return 'select-outline select-outline-' + d; })
-        .attr('transform', 'translate(' + xs + ', ' + ys + ')')
-        .attr('d', 'M' + paths.join('M') + 'Z');
+    if(paths.length) {
+        outlines.enter()
+            .append('path')
+            .attr('class', function(d) { return 'select-outline select-outline-' + d; })
+            .attr('transform', 'translate(' + xs + ', ' + ys + ')')
+            .attr('d', 'M' + paths.join('M') + 'Z');
+    }
 }
 
 function updateZoombox(zb, corners, box, path0, dimmed, lum) {
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -140,7 +140,10 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
                 // to pan (or to zoom if it already is pan) on shift
                 if(e.shiftKey) {
                     if(dragModeNow === 'pan') dragModeNow = 'zoom';
-                    else dragModeNow = 'pan';
+                    else if(!isSelectOrLasso(dragModeNow)) dragModeNow = 'pan';
+                }
+                else if(e.ctrlKey) {
+                    dragModeNow = 'pan';
                 }
             }
             // all other draggers just pan
@@ -168,12 +171,30 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             else if(isSelectOrLasso(dragModeNow)) {
                 dragOptions.xaxes = xa;
                 dragOptions.yaxes = ya;
+
+                // take over selection polygons from prev mode, if any
+                if(e.shiftKey && plotinfo.selection.polygons && !dragOptions.polygons) {
+                    dragOptions.polygons = plotinfo.selection.polygons;
+                    dragOptions.mergedPolygons = plotinfo.selection.mergedPolygons;
+                }
+                // create new polygons, if shift mode
+                else if(!e.shiftKey || (e.shiftKey && !plotinfo.selection.polygons)) {
+                    plotinfo.selection = {};
+                    plotinfo.selection.polygons = dragOptions.polygons = [];
+                    dragOptions.mergedPolygons = plotinfo.selection.mergedPolygons = [];
+                }
+
                 prepSelect(e, startX, startY, dragOptions, dragModeNow);
             }
         }
     };
 
     dragElement.init(dragOptions);
+
+    // FIXME: this hack highlights selection once we enter select/lasso mode
+    if(isSelectOrLasso(gd._fullLayout.dragmode) && plotinfo.selection) {
+        showSelect(zoomlayer, dragOptions);
+    }
 
     var x0,
         y0,
@@ -526,6 +547,9 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         }
 
         updateSubplots([x0, y0, pw - dx, ph - dy]);
+
+        if(plotinfo.ondrag) plotinfo.ondrag.call([x0, y0, pw - dx, ph - dy]);
+
         ticksAndAnnotations(yActive, xActive);
     }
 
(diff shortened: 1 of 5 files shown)


dfcreative:
@etpinard can you please rerun the test? It is fine on my machine.

etpinard:
Yep, [`npm run test-jasmine -- --bundleTest=ie9_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/bundle_tests/ie9_test.js) correctly fails as `poly-bool` has a few `Unit8Array` calls:

image

Now, I don't think multi-selections in IE9 is a must. Maybe there's a way to make `poly-bool` not invoke `Unit8Array` on load.

dfcreative:
@etpinard turns out `poly-bool` is a huge dependency with redundancy. Replaced it with `polybooljs` - faster and smaller alternative

src/plots/cartesian/select.js
```diff
@@ -226,10 +259,30 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
             else {
                 dragOptions.gd.emit('plotly_selected', eventData);
             }
+
+            if(currentPolygon && dragOptions.polygons) {
+                // save last polygons
+                dragOptions.polygons.push(currentPolygon);
+
+                // we have to keep reference to arrays, therefore just replace items
+                dragOptions.mergedPolygons.splice.apply(dragOptions.mergedPolygons, [0, dragOptions.mergedPolygons.length].concat(mergedPolygons));
```
etpinard: What does this do? Could we just use a plain-old for-loop?

src/plots/cartesian/dragbox.js
```diff
@@ -924,9 +973,7 @@ function removeZoombox(gd) {
 }
 
 function isSelectOrLasso(dragmode) {
-    var modes = ['lasso', 'select'];
-
-    return modes.indexOf(dragmode) !== -1;
+    return dragmode === 'lasso' || dragmode === 'select';
```
etpinard: Way better :racehorse: . Thanks!

src/plots/cartesian/select.js
```diff
@@ -226,10 +259,31 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
...
+
+                // we have to keep reference to arrays container
+                dragOptions.mergedPolygons.length = 0;
+                [].push.apply(dragOptions.mergedPolygons, mergedPolygons);
+            }
         });
     };
 };
 
+function joinPolygons(list, poly) {
```
etpinard: Joining the polygons sounds very reasonable to me. Some users might expect that selections within selections would remove the inner polygons. Perhaps we'll need a layout options (e.g. `layout.multiselectionmode: 'join' | 'cut'`).

Any thoughts about this @alexcjohnson @chriddyp @cldougl @monfera ?

cldougl: so `'join'` mode would always add points the the selection (and do nothing if a selected point is selected a second time) and `'cut'` mode would always remove points from the selection (and do nothing if the point was not previously selected)?

cldougl: +1 for a `cut` mode (after seeing the join in action)

alexcjohnson: My gut reaction: if you start a new selection *inside* an existing one, it's purely subtractive, if you start *outside* the existing ones it's purely additive. Would have to play with it implemented to see if this really feels intuitive though...

dfcreative: Ok, added subtract selection by holding <kbd>Alt</kbd> in e8c18

etpinard: > My gut reaction: if you start a new selection inside an existing one, it's purely subtractive, if you start outside the existing ones it's purely additive.

+1

alexcjohnson: > Ok, added subtract selection by holding <kbd>Alt</kbd> in e8c18f0

That *is* more powerful (as well as easier to code 😅 ), my only concern is whether people will be able to find it. Though I guess if they can find add they can find subtract...

(and I never saw `<kbd>` before, nice! 🎉 )

chriddyp:
This is so awesome.
multiple-selections

src/plots/cartesian/select.js
etpinard: Nice solution @dfcreative . The only (potential) problem I can think of is mobile and tablets users complaining no having access to Shift and Alt keys.

dfcreative: @etpinard good point, we can use @alexcjohnson strategy for that. But I'd wait for persistent selections PR merged first

etpinard:
> But I'd wait for persistent selections PR merged first 

Sounds good @dfcreative Looks like everyone is happy with e8c18f0 :tada: 

Now, would you mind adding a few test cases in https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/select_test.js to :lock: down the behavior?

src/plots/cartesian/dragbox.js
```diff
@@ -168,13 +171,31 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             else if(isSelectOrLasso(dragModeNow)) {
                 dragOptions.xaxes = xa;
                 dragOptions.yaxes = ya;
+
+                // take over selection polygons from prev mode, if any
+                if((e.shiftKey || e.altKey) && (plotinfo.selection && plotinfo.selection.polygons) && !dragOptions.polygons) {
```
etpinard: Would it be possible to move this block into `prepSelect` so that all subplot types that support selections can allow multi-select? Right now, only cartesian and gl2d plot types allow it.

src/plots/cartesian/dragbox.js
```diff
@@ -168,13 +171,31 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
+
                 prepSelect(e, startX, startY, dragOptions, dragModeNow);
             }
         }
     };
 
     dragElement.init(dragOptions);
 
+    // FIXME: this hack highlights selection once we enter select/lasso mode
+    if(isSelectOrLasso(gd._fullLayout.dragmode) && plotinfo.selection) {
```
etpinard: I'm not a big fan of this behavior. @dfcreative is there a particular reason why you added this?

src/plots/cartesian/dragbox.js
```diff
@@ -526,6 +547,9 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         }
 
         updateSubplots([x0, y0, pw - dx, ph - dy]);
+
+        if(plotinfo.ondrag) plotinfo.ondrag.call([x0, y0, pw - dx, ph - dy]);
```
etpinard: why do we need this?

test/jasmine/tests/select_test.js
```diff
@@ -193,15 +236,37 @@ describe('Test select box and lasso in general:', function() {
...
+                    pointNumber: 4,
+                    x: 0.013,
+                    y: 6.875,
+                    id: 'id-0.013',
+                    customdata: 'customdata-6.875'
+                }], 'with the correct selecting points (1)');
+            })
+            .then(function() {
+                // sub selection
+                drag([[219, 143], [219, 183]], {altKey: true});
```
etpinard: Brilliant tests :sparkles: 

Thanks

src/plots/cartesian/dragbox.js
dfcreative: Not relevant since #2135

src/plots/cartesian/dragbox.js
dfcreative: Removed

test/jasmine/tests/select_test.js
dfcreative: :) you're welcome!

src/plots/cartesian/dragbox.js
dfcreative: Yep

etpinard:
Beautifully done @dfcreative, merge away :dancer:

PavelGolodoniuc:
Is there a way to disable multiselect mode and return to the old behaviour when Shift was invoking pan mode?

dfcreative:
@PavelGolodoniuc pan mode can be enabled by <kbd>Ctrl</kbd> key, is that a case?

PavelGolodoniuc:
@dfcreative it doesn't seem to be the case. When "Box select" is mode is selected Ctrl key doesn't seem to enable pan. Also our application traditionally used Ctrl key over the plot for a user-defined functionality.

PavelGolodoniuc:
@dfcreative, @etpinard  Can the multiselect mode including Shift/Alt/Ctrl handler be disabled completely in the Box select mode? In our application we do not allow irregular selections in general and thus would like to disable this functionality completely.

ammu1395:
what are the changes that we need to make to enable this in the plotly

PR #2141, opened 2017-11-03
Mock additions for persistent point selection

monfera:
New mocks as part of reviewing https://github.com/plotly/plotly.js/pull/2135 to see it with various traces, and also also because of Étienne's travel.

### DIFF
--- a/test/image/mocks/point-selection-bar.json
+++ b/test/image/mocks/point-selection-bar.json
@@ -1,0 +1,100 @@
+{
+  "data": [{
+    "mode": "lines+markers+text",
+    "x": [1, 2, 3, 4, 5, 6],
+    "y": [1, 3, 2, 4, 5, 7],
+    "ids": ["a", "b", "c", "d", "e", "f"],
+    "text": ["a", "b", "c", "d", "e", "f"],
+    "marker": {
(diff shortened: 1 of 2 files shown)


etpinard:
Can you please make a PR to https://github.com/plotly/plotly.js/pull/2135 instead of `master`?

monfera:
Ah indeed, thanks, changed!

monfera:
The bar looks good on initial render except where I have a small question:
image
The `selectedpoints` is `[0, 1, 4]` but the last highlighted orange bar is of index 5. The reason is that the `]3.5, 4.5[` filter transform eliminates one point. So I just wanted to make sure it's OK that post-filter point indices are used rather than pre-filter.

In the second bar trace, ie. last trace, both `selected` and `unselected` are specified as orange. When clearing the selection, it'll be all green but iirc there's some todo item for bars, perhaps this.
tmp
That the legend also shows both traces in green is probably the same thing.

etpinard:
Superseded by https://github.com/plotly/plotly.js/pull/2135/commits/3d2049c0c6ada1c741da27184704e1c477227a6b in https://github.com/plotly/plotly.js/pull/2135

Issue #2136, opened 2017-10-31
Setting default colors


cixzhang:
Could we have a way to override the default colors for plotlyjs?

I'm willing to work on this feature.

etpinard:
Can you be more specific? 

You would like to make the default trace-to-trace color [sequence](https://github.com/plotly/plotly.js/blob/4b8b10df27c53390751f01e949ab3a75e912367c/src/components/color/attributes.js#L12-L24) customizable?

cixzhang:
@etpinard Yup! Exactly! I'd like to be able to set a specific color scheme for all chart traces to use.

etpinard:
Ok. I'm thinking this should be a config [option](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_config.js) e.g. `traceColorDefaults`.

cixzhang:
That sounds good to me! I'll take a stab at it tonight.

cixzhang:
Following the PR guidelines, I made a PR here in my own fork: https://github.com/cixzhang/plotly.js/pull/1.

Please have a look when you have time!

alexcjohnson:
@cixzhang Great PR! @etpinard and I had a discussion about this offline though, and we concluded it actually should be a [top-level layout attribute](https://github.com/plotly/plotly.js/blob/master/src/plots/layout_attributes.js), not a config option. The advantage of this is that the result is portable (on plot.ly for example `data` and `layout` get saved with the plot but `config` is context-dependent), and easier to change (via `relayout` / `update`) - which can be useful for example if you want to try out a bunch of different themes / templates for your plot (cc @bpostlethwaite )

We're thinking `layout.colorway` - sound reasonable?

cixzhang:
Thanks for the feedback! That makes sense to me. I'll take a look at implementing it as a layout attribute this evening.

cixzhang:
I've created a new PR with `layout.colorway` on my fork here: https://github.com/cixzhang/plotly.js/pull/2

cixzhang:
I have updated my PR here: https://github.com/cixzhang/plotly.js/pull/2. Let me know if there's anything else I should do. :)

bpostlethwaite:
Thanks @cixzhang we appreciate the PR and will get to it ASAP. Unfortunately that may be a few days as our core plotly.js team is either on vacation or dealing with unforeseen events. Everything should be back to normal next week. Thanks again.

etpinard:
Done in #2156

PR #2156, opened 2017-11-13
adding layout.colorway

cixzhang:
For https://github.com/plotly/plotly.js/issues/2136

This PR adds the `layout.colorway` attribute which takes a list of colors to use as the default set of colors for all traces. By default, it should use the existing `Color.defaults`.


etpinard:
To add the baseline corresponding to your new mock, see 

https://github.com/plotly/plotly.js/blob/master/test/image/README.md#c-generate-or-update-existing-baseline-image

If you're having issues with our docker image-test container, let us know.

src/plots/layout_attributes.js
```diff
@@ -175,5 +175,12 @@ module.exports = {
         role: 'info',
         editType: 'legend',
         description: 'Determines whether or not a legend is drawn.'
-    }
+    },
+    colorway: {
+        valType: 'colorlist',
+        dflt: colorAttrs.defaults,
+        role: 'style',
+        editType: 'calc',
```
etpinard: Comment for future development: note here that only `pie` traces require this to be `editType: 'calc'`. For most trace types, `editType: 'plot'` would have sufficed and even `editType: 'style'` might have been good enough for cartesian traces. Oh well, we might have to add-trace-type-specific `editType` flags for layout attributes down the road to improve performance.

test/image/mocks/layout-colorway.json
```diff
@@ -0,0 +1,32 @@
...
+    {"y": [8, 7, 7, 6, 5, 4, 4, 3, 2, 2, 3]},
+    {"y": [7, 7, 6, 5, 4, 4, 3, 2, 2, 3, 8]},
+    {"y": [7, 6, 5, 4, 4, 3, 2, 2, 3, 8, 7]},
+    {"y": [6, 5, 4, 4, 3, 2, 2, 3, 8, 7, 7]},
+    {"y": [5, 4, 4, 3, 2, 2, 3, 8, 7, 7, 6]},
+    {
+      "labels": ["a","b","c","c","c","a","d","e","f","f","g","h"],
+      "type": "pie",
+      "domain": {"x": [0, 0.4]},
+      "xaxis": "x2",
```
etpinard: `xaxis` and `yaxis` don't do anything for pie traces. Would you mind :hocho: them?

test/image/mocks/layout-colorway.json
```diff
@@ -0,0 +1,32 @@
...
+      "#E83898",
+      "#A83DD1",
+      "#5A38E8",
+      "#3C71DE"
+    ],
+    "xaxis": {
+      "domain": [0.4, 1]
+    },
+    "yaxis": {
+      "anchor": "y2"
```
etpinard: `anchor` for y-axes expects x-axis ids, so this thing here gets ignored. :hocho:

src/lib/coerce.js
```diff
@@ -141,6 +141,22 @@ exports.valObjectMeta = {
             else propOut.set(dflt);
         }
     },
+    colorlist: {
```
etpinard: Would you mind adding a [`Lib.validate`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/lib_test.js#L928) test case for the new colorlist `valType`?

etpinard:
A very lovely addition :heart: 

Thanks very much @cixzhang :tada:

PR #1993, opened 2017-09-06
Shared canvases

dfcreative:
Comes after https://github.com/plotly/plotly.js/pull/1989.

Makes use of pre-created 3-canvas strategy, so that no new canvases created.

PR #2029, opened 2017-09-21
Update mapbox-gl requirement

pfhayes:
mapbox-gl 0.22.1 is very old and transitively
pulls in a vulnerable version of concat-stream.
See https://snyk.io/vuln/npm:concat-stream:20160901

PR #2159, opened 2017-11-14
Introduce global contexts

dfcreative:
Excerpt from #1869.
Covers #1993 and #1989.


dfcreative:
Apparently there are some IE9-related issues with regl.

src/plot_api/plot_api.js
```diff
@@ -42,6 +42,8 @@ var enforceAxisConstraints = axisConstraints.enforce;
 var cleanAxisConstraints = axisConstraints.clean;
 var axisIds = require('../plots/cartesian/axis_ids');
 
+var createRegl = require('regl');
```
etpinard: Ha. `regl` must be invoking one of many globals that IE9 doesn't support at runtime.

As this file here `plot_api.js` is required in all partial bundles (even that gl-less ones), this makes the IE9 test fail.

So, I think that best to solve this would be use a `Registry.getComponentMethod`-style pattern so that `regl` is only required in bundles that expose `regl` traces.

dfcreative: @etpinard do I get it right that we don't have this safe-require pattern yet? What if I just require regl directly when the `regl` trace type is detected?

etpinard: >  What if I just require regl directly when the regl trace type is detected?

You mean something like:

```js
if(fullLayout._has('regl')) {
  var createREGL = require('regl')
}
```

in `plot_api.js`? That could work, `regl` would still got bundled up in all partial bundles.

src/plot_api/plot_api.js
```diff
@@ -2754,7 +2788,7 @@ function makePlotFramework(gd) {
         .classed('plotly', true);
 
     // Make the svg container
-    fullLayout._paperdiv = fullLayout._container.selectAll('.svg-container').data([0]);
+    fullLayout._paperdiv = fullLayout._container.selectAll('.svg-container').data([{}]);
```
etpinard: Why change `[0]` to `[{}]`?

src/plots/plots.js
```diff
@@ -569,15 +569,29 @@ plots.createTransitionData = function(gd) {
...
 
-    for(var i = 0; i < basePlotModules.length; i++) {
+    for(i = 0; i < basePlotModules.length; i++) {
         var _module = basePlotModules[i];
 
         if(_module.name === category) return true;
     }
 
+    // check trace
+    var modules = this._modules || [];
```
etpinard: brilliant :sparkles:

src/traces/parcoords/base_plot.js
```diff
@@ -41,22 +37,20 @@ exports.toSVG = function(gd) {
...
             xmlns: xmlnsNamespaces.svg,
             'xlink:href': imageData,
-            x: size.l + size.w * domain.x[0] - c.overdrag,
-            y: size.t + size.h * (1 - domain.y[1]),
-            width: (domain.x[1] - domain.x[0]) * size.w + 2 * c.overdrag,
-            height: (domain.y[1] - domain.y[0]) * size.h,
+            x: 0,
+            y: 0,
+            width: canvas.width,
+            height: canvas.height,
```
etpinard: Lovely cleanup :heart:

src/plot_api/plot_api.js
```diff
@@ -2765,10 +2799,13 @@ function makePlotFramework(gd) {
     // TODO: sort out all the ordering so we don't have to
     // explicitly delete anything
     fullLayout._glcontainer = fullLayout._paperdiv.selectAll('.gl-container')
-        .data([0]);
+        .data([{}]);
```
etpinard: what about this one here?

src/plot_api/plot_api.js
```diff
@@ -195,6 +195,40 @@ Plotly.plot = function(gd, data, layout, config) {
...
+                pick: false
+            }, {
+                key: 'focusLayer',
+                context: false,
+                pick: false
+            }, {
+                key: 'pickLayer',
+                context: false,
+                pick: true
+            }]);
```
etpinard: I think you need a [key-function](https://bost.ocks.org/mike/constancy/) here to make sure to not create these canvas nodes on every _hard_ redraw that go through `drawFramework`.

Sometime like:

```js
fullLayout._glcanvas = fullLayout._glcontainer.selectAll('.gl-canvas').data([{
  /* .. as currently ... */
}], function(d) { return d.key; });
```

etpinard: This makes me wonder: we probably don't have sufficient test coverage for `Plotly.restyle` calls that morph a gl-based trace types into a SVG-based. For example off the regl2d branch we should have:

```js
Plotly.newPlot(gd, [{
  type: 'scattergl',
  // ...
}])
.then(function() {
   d3.selectAll('canvas').size() // => 3

  return Plotly.restyle(gd, 'type', 'scatter')
})
.then(function() {
  d3.selectAll('canvas').size() // => 0
})
```

As restyling a `parcoords` traces into any other trace types doesn't make much sense at the moment. You could instead try something like:

```js
Plotly.newPlot(gd, [{
  type: 'parcoords',
  // ...
}])
.then(function() {
   d3.selectAll('canvas').size() // => 3

  return Plotly.deleteTraces(gd, [0])
})
.then(function() {
  d3.selectAll('canvas').size() // => 0
})
```

src/plot_api/plot_api.js
dfcreative: that is the way `parcoords` is doing it's thing: https://github.com/plotly/plotly.js/pull/2159/files#diff-2fad23de82bf8c7625ea8e11436233b7R290
I was not really digging in why is it done this and not another way. Before it used own container for storing this data.

etpinard: Does it work ok with `data([0])`?

dfcreative: @etpinard that is the point that parcoords puts `vm` property there via `extendFlat`, probably it is possible to find a better storage for it, I just did not figure out what.

src/plots/plots.js
```diff
@@ -609,6 +609,16 @@ plots.cleanPlot = function(newFullData, newFullLayout, oldFullData, oldFullLayou
 
     var hasPaper = !!oldFullLayout._paper;
     var hasInfoLayer = !!oldFullLayout._infolayer;
+    var hadGl = oldFullLayout._has && oldFullLayout._has('gl');
+    var hasGl = newFullLayout._has && newFullLayout._has('gl');
+
+    if(hadGl && !hasGl) {
+        if(oldFullLayout._glcontainer !== undefined) {
+            oldFullLayout._glcontainer.selectAll('.gl-canvas').data([]).exit().remove();
```
etpinard:
```js
oldFullLayout._glcontainer.selectAll('.gl-canvas').remove()
```

should suffice.

etpinard:
@dfcreative Good news the parcoords tests are passing again

Bad news other `@noCI` tests are failing.

image


Can you try making `npm run test-jasmine -- --tags=noCI` pass?

dfcreative:
@etpinard fixed these four noCI tests, seems that others (mapbox ones) are not related to this PR but are specific to my machine.

etpinard:
> seems that others (mapbox ones) are not related to this PR but are specific to my machine.

Yeah, those mapbox tests are pretty resource-intensive; that doesn't surprise me. Hopefully https://github.com/plotly/plotly.js/pull/2029 will fix those issues.

-----

Great PR. Merge away :dancer:

PR #2163, opened 2017-11-15
Persistent point selection with transforms compatibility

etpinard:
These few commits essentially addresses https://github.com/plotly/plotly.js/pull/2135#issuecomment-341811820 while cleaning up a few things in event data land - mainly making hover and selection event data reuse the same logic.

----

In brief, this PR adds `pointIndex` event data field and uses its value to fill in `selectedpoints` with input indices. Note that `pointIndex` is the same as `pointNumber` in all non-transformed traces except histograms. With the help of @monfera 's `indexToPoints`, this part was relatively easy :tada: 

I chose to add a new field instead of _fixing_ the existing `pointNumber` field to retain backward compatibility (I suspect a certain support-assets project we did back in May would be break if we tweaked `pointNumber`  :smirk: ), but yeah `pointIndex` is what users should use from now on. We could even :hocho: `pointNumber` in v2.

`pointIndex` also has a multi-value version called `pointIndices` for histogram traces and aggregation transforms. 

I'll wait for @alexcjohnson to :+1: the commits below before pushing new test cases and incorporating them in https://github.com/plotly/plotly.js/pull/2135 

cc @monfera


etpinard:
Referencing https://github.com/plotly/plotly.js/pull/2126#pullrequestreview-72395582 which first put forward something similar to this PR's `pointIndex`.

src/traces/scatterternary/event_data.js
```diff
@@ -0,0 +1,37 @@
...
+        var cdi = cd[pointNumber];
+
+        // N.B. These are the scale coordinates !!!
+        //
+        // On master, hover events get the non-scaled coordinates
+        // whereas selection events get the scaled version.
+        // Note also that the hover labels show the scaled version.
+        //
+        // What about the 'raw' input coordinates?
+        // Should we include them in parallel here or replace a/b/c with them?
```
alexcjohnson: I think the scaled (normalized) coords are the most useful here, and folks can get the raw data back out easily enough using the indices, lets not add them here.

src/lib/index.js
```diff
@@ -465,6 +465,57 @@ lib.extractOption = function(calcPt, trace, calcKey, traceKey) {
...
+        for(var k in indexToPoints) {
+            var pts = indexToPoints[k];
+            for(var j = 0; j < pts.length; j++) {
+                ptIndex2ptNumber[pts[j]] = k;
+            }
+        }
+    }
+
+    function isPtIndexValid(v) {
+        return lib.validate(v, {valType: 'integer', min: 0});
```
alexcjohnson: ha, that's super 🌴 but might be a bit slow for this 🌶 path - non-negative integers are just
`typeof v === 'number' && v >= 0 && v % 1 === 0` right?

etpinard: Thanks!

:racehorse: > :palm_tree: in :hot_pepper: code paths

alexcjohnson: Oh hmm I guess this is looking at user input so it should be `isNumeric(v)` instead of `typeof v === 'number'` ie allow index `'1'` as well as `1`

src/traces/histogram/calc.js
```diff
@@ -512,3 +510,19 @@ function cdf(size, direction, currentBin) {
         }
     }
 }
+
+function calcSelection(cd, trace) {
+    if(Array.isArray(trace.selectedpoints)) {
+        var ptNumber2cdIndex = {};
+
+        // make histogram-specific pt-number-to-cd-index map object
```
alexcjohnson: don't we want this available for use later, ie without a recalc? I would have imagined creating this in the [main binning loop](https://github.com/plotly/plotly.js/blob/315e6327bbd512b20cec61e651f6bb945c1521d0/src/traces/histogram/calc.js#L128) and stashing it in the full trace, also possibly creating it as an array rather than a map, seems like that would be more efficient and would work just as well?

etpinard: > don't we want this available for use later, ie without a recalc?

I don't think so. Calling `restyle` with `selectedpoints` has to trigger a recalc. On user selections, the calcdata items are mutated [directly](https://github.com/plotly/plotly.js/blob/261387dcb04502b82634e2ea220d3d828d3d3f8e/src/traces/bar/select.js#L33) w/o needing to refer back to input data-array indices. Am I missing something?

etpinard: > I would have imagined creating this in the main binning loop 

Good call, done in -> 7543dc6

src/traces/scatterternary/event_data.js
etpinard: done in c8d715b

src/lib/index.js
etpinard: done in c8d715b

src/traces/histogram/calc.js
alexcjohnson: > Calling `restyle` with `selectedpoints` has to trigger a recalc.

hmm ok, I was hoping we could avoid that but we can discuss optimizing that later.

alexcjohnson:
💃

PR #2169, opened 2017-11-17
Fill in 'id' in bar calcdata so that ErrorBars.plot knows what do it with it

### LINKED ISSUE #2168: Regression observed on errors-bars in bar plot
Hi,

I recently observed  a regression over `error-bars`, I can't say which version introduced the regression.

Here's a working version in **plotly-1.2.0** : [sample OK](https://codepen.io/anon/pen/vWWpoR)

Here's a sample where I observe the regression (**plotly-latest**) (only the first error-bar appear) : [sample KO](https://codepen.io/anon/pen/mqqXJj) 

Maybe I've missed something from [the doc](https://plot.ly/javascript/error-bars/).

Thanks for your time and consideration.

etpinard:
fixes #2168 

`ErrorBars.plot` looks for `id` in calcdata [here](https://github.com/plotly/plotly.js/blob/82326bad3490be56d9c068122e49dd7975ef4f64/src/components/errorbars/plot.js#L37-L39) to construct a d3 key-function. Previously, `bar` traces did not fill `id`, so commit https://github.com/plotly/plotly.js/commit/129ad6c35f2b772b1f010351cbe9729d7600960a broke bar error bar with ids as `trace.ids` was now defined but [`d.id`](https://github.com/plotly/plotly.js/blob/82326bad3490be56d9c068122e49dd7975ef4f64/src/components/errorbars/plot.js#L38) `=== undefined` for all bar calcdata items.

A pretty solid Friday-afternoon-three-line bug fix. cc @alexcjohnson


alexcjohnson:
Nice and easy 🍹 💃

Issue #2002, opened 2017-09-13
cliponaxis: false fails with date & log axes


alexcjohnson:
markers and text disappear when you use `cliponaxis: false` with a date axis:
```js
var x = [];
var y = [];
for(var i = 1; i < 8; i++)
{
  x.push('2017-11-' + i);
  y.push(Math.floor(Math.random() * 10000));
}

Plotly.newPlot(gd,
  [{
      x: x,
      y: y,
      type: 'scatter',
      text: y,
      fill: 'tonexty',
      mode: 'lines+markers+text',
      textposition: 'top',
      cliponaxis: false
  }],
  {width: 400, height: 400}
);
```
![screen shot 2017-09-13 at 10 18 18 am](https://user-images.githubusercontent.com/2678795/30382363-ea3d3f5a-986c-11e7-838c-fb513cd6d995.png)
also fails with log axes, but works fine if the x axis is either linear or category:
```js
var x = [];
var y = [];
for(var i = 1; i < 8; i++)
{
  x.push(String.fromCharCode(i+64));
  y.push(Math.floor(Math.random() * 10000));
}

Plotly.newPlot(gd,
  [{
      x: x,
      y: y,
      type: 'scatter',
      text: y,
      fill: 'tonexty',
      mode: 'lines+markers+text',
      textposition: 'top',
      cliponaxis: false
  }],
  {width: 400, height: 400}
);
```
![screen shot 2017-09-13 at 10 22 26 am](https://user-images.githubusercontent.com/2678795/30382581-83dd3ade-986d-11e7-9cc6-63e722d016ab.png)

discovered while investigating #2000

etpinard:
Oh well, this [is](https://github.com/plotly/plotly.js/blob/11bb997e78128b9f47db2ed392982f82a026e589/src/plots/cartesian/set_convert.js#L440) dumb. My mistake. Adding `d2r()` should do the trick.

alexcjohnson:
> Adding `d2r()` should do the trick.

Will probably have to be `d2c` for `x/y` and `r2c` for `range` (can we trust `ax._rl` to exist?) but yeah. And calendars 🙄

PR #2177, opened 2017-11-20
Cliponaxis dates

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2002 - where trace input calendars made this a little trickier than expected.


alexcjohnson:
> calendars made this a little trickier than expected

As always...

Looks great, nice simplifications, despite the extra trickiness you actually ended up -3 LOC in `src` 🎉 

💃

PR #2182, opened 2017-11-27
`table` `columnwidth` to accept strings

monfera:
As per https://github.com/plotly/streambed/issues/10338#issuecomment-346610238 and  https://github.com/plotly/streambed/issues/10338#issuecomment-346699036


etpinard:
Looks good. Thanks :dancer:

PR #2183, opened 2017-11-27
Update "Save and edit plot in cloud" text

chriddyp:


alexcjohnson:
Or maybe `'Edit in Chart Studio'`? That gives you some indication of *why* you'd want to send it there.

jackparmer:
We might want to eventually  think about adding another `/external` endpoint and modebar button for `Add to Plot.ly dashboard` (not in this PR obviously, but while on the subject :) ).

chriddyp:
Alright, I updated to `Edit in Chart Studio`. Good to 💃 @alexcjohnson ?

alexcjohnson:
💃

PR #2187, opened 2017-11-28
Upgrade bundling

dfcreative:
* introduces `minify-stream` instead of `uglifyjs2` - less errors, better options
* removes `patch_minified` (seems that `uglify-es` fixes that since tests pass)
* introduces `bubleify` to handle ES6 dependencies


tasks/util/constants.js
```diff
@@ -84,16 +84,29 @@ module.exports = {
...
         output: {
             beautify: false,
             ascii_only: true
-        }
+        },
+        sourceMap: false
+    },
+
+    bubleifyOptions: {
+        target: {
```
etpinard: Hmm. We have to target browsers waaaaaaaay older than that. Can bubleify guarantee to output code that IE11 understands?

etpinard:
@dfcreative can you run `npm run bundle && git diff -- dist/README.md` and paste a screenshot of the diff to get an idea of how the new minifiers perform?

tasks/util/constants.js
dfcreative: @etpinard that is the best bubleify can do, it is targeted only for ES6 → ES5, not ES5 → ES3 etc.

etpinard: Hmm. Well that won't work then. plotly.js needs to be pure ES5.

dfcreative:
image

:/

tasks/util/constants.js
dfcreative: @etpinard that is pure ES5, same as what we have now. `target` flag only indicates what new ES6 features are supported by target environment, eg. some browsers have `Object.assign` etc, so since we indicate oldest target possible, we make sure we get pure ES5.

etpinard: Could we put these config options in the package.json instead, so that all consumers bundling with browserify or webpack+ify-loader get the same results as us in our `dist/`?

etpinard: Can the output bundle be minified by plain-old uglify-js? e.g. does

```
browserify -t bubleify src/plotly.js | uglify-js > bundle.js
```

work?

etpinard: I'm not a big fan of this `target` option. I find it pretty confusing. Could we instead manually list all the [transforms](https://buble.surge.sh/guide/#command-line-options) we want turned on?

tasks/util/patch_minified.js
```diff
@@ -1,22 +0,0 @@
...
- * - https://github.com/plotly/plotly.js/pull/1094
- *
- * For example, this routine replaces
- *  'require("+o(s)+")' -> 'require("+ o(s) +")'
- *
- * But works for any 1-letter variable that uglify-js may output.
- *
- */
-module.exports = function patchMinified(minifiedCode) {
-    return minifiedCode.replace(PATTERN, NEW_SUBSTR);
```
etpinard: One of the hackiest thing in plotly.js at the moment. Good riddance :tada:

tasks/stats.js
```diff
@@ -1,6 +1,6 @@
 var path = require('path');
 var fs = require('fs');
-var spawn = require('child_process').spawn;
+var spawn = require('cross-spawn');
```
etpinard: Ha nice catch. That's for that `spawn('npm', ['ls', '--json', '--only', 'prod']);` statement below - which will be obsolete when we switch to `npm@5` as the new `package-lock.json` file will do that same as the current `dist/npm-ls.json`.

dfcreative:
Actually that diff is not that scary, the actual master has the same size: 
image
It is just the `dist/README.md` we have is old.

etpinard:
> It is just the dist/README.md we have is old.

Ha. I see. Yeah we have a bunch of new things in master that didn't get released yet. :ok_hand:

etpinard:
For reference:

on the latest master:

image

off this branch:

image

Lovely :heart:

etpinard:
Great PR. Thanks for doing this @dfcreative 

Getting rid of `patch_minified.js` well worth swapping `uglify-js` for `uglify-es`.

Merge away :dancer:

ndabAP:
My Webpack asset code splitted bundle increased by approximately 200 kilobytes after updating to 1.32.0.

etpinard:
That's because we added a lot of features (sorry :wink: ). 

But still 200 kB seems larger than the worse case increase (see bundle size changes from `1.31.2` to `1.32.0` [here](https://github.com/plotly/plotly.js/commit/90218e5e9b73fd503f366b8ccf531464df2f9f2b#diff-90524ffa151b37c8b277925c017ad1ba)). Perhaps something is off. Would you care to share a reproducible example of your bundle setup and config?

ndabAP:
> That's because we added a lot of features 

Got it. Would still be nice to be more selective about the features. It would be cool to have the possibility to omit special features like chart zooming, saving as picture or maybe animations to get a smaller package.

> Would you care to share a reproducible example of your bundle setup and config?

Unfortunately, my asset has a lot of Vue.js related stuff in it. It would take weeks to sort it out ;) It's not exactly 200 kilobytes, though.

etpinard:
> Would still be nice to be more selective about the features.

Yep, we know. That'll have to wait for [v2](https://github.com/plotly/plotly.js/issues/420) though.

Issue #856, opened 2016-08-16
Provide proper i18n support


jnehlmeier:
Currently Plotly has poor i18n support.
- You can not change d3 locale (issue #380)
- Adjustment of rendered precision on date axis does not work when applying custom `axis.tickformat` if dates need to be rendered differently because of locale. If you want to keep that feature you have to patch hardcoded format strings: https://github.com/plotly/plotly.js/blob/ad88139911d09b52cfafecc2bc7e72fc267b3a7f/src/plots/cartesian/axes.js#L879
- Error messages, notifier messages, hover texts for default mode bar buttons are hardcoded in english https://github.com/plotly/plotly.js/issues/288

Currently I have to manually patch plotly for each release, which is pretty annoying.

etpinard:
Thanks for writing in. 

It shouldn't be too hard to pass these locale options in the plot [config](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_config.js) and have the axes code look up `gd._context` for their values.

I added a link to a past issue discussing mode bar internationalisation and I marked #380, #288 as duplicate.

fhurta:
I'm voting for this feature!

sclausen:
:+1: The only thing that keeps us from moving from Highcharts to plotly.js

etpinard:
@sclausen 

> us

Who would that be?

florianleger:
Hi there,
I've made a fork of Plotly.js to handle french translations and d3 locales for my company:
https://github.com/florianleger/plotly.js
([diff](https://github.com/plotly/plotly.js/compare/master...florianleger:master))
I explain in the README how it works and how to add new languages, I thought that it might be helpful to others until a patch came out.

alexcjohnson:
I believe that between #2195 and #2207 this issue on i18n can be closed - but before I do so, I'd love some 👍 /👎  on the framework from folks who've weighed in here.

[...]

Issue #265, opened 2016-02-19
Three-variable hover labels should adapt to supplied axis titles


etpinard:
As described in http://stackoverflow.com/questions/35467316/rename-axes-in-plotly-3d-hover-text

fixxxr:
Hi there, 

is this enhancement available for plotly.js? If yes where I can find a usage example?

thanks in advance

ischurov:
@fixxxr, no it is not yet implemented (otherwise this bug would have status 'closed', but it's 'open').

fixxxr:
I saw that but since the label was added 17 days ago I thought It was already implemented

thanks for the quick answer

etpinard:
@fixxxr labels are described in the [contributing guidelines](https://github.com/plotly/plotly.js/blob/master/CONTRIBUTING.md#issue-labels).

etpinard:
Merging into https://github.com/plotly/plotly.js/issues/2618

PR #183, opened 2016-01-14
Replace json assets by js files

etpinard:
@mdtusz @bpostlethwaite 

Webpack does not bundle up json files by default. It is possible using user-defined [loaders](https://webpack.github.io/docs/using-loaders.html) e.g. this [one](https://github.com/webpack/json-loader) for json files. But this is an extra hurdle for webpack users that we can *fix* on our end.

This PR takes one step towards supporting Webpack bundling without any special configs.

Note that I used this [script](https://gist.github.com/etpinard/bac3656466458607ddd0) to check for non-js files that are bundled into plotly.js.

PR #2195, opened 2017-12-07
Localization

alexcjohnson:
- Creates a framework for localization:
  - Dictionaries can either be registered (in the bundle or afterward) or provided to a plot as config
  - Locale to use must be provided as plot config
- Wraps all displayed string literals in `gd`-specific localization function (did I miss any?)
- Collects all translation key strings (on build or test) and barfs if any non-literals are translated

- *Does not localize date tick labels.* This will be a separate PR, using the same `config.locale` but pulling `world-calendars` locale data in, so no additional translations will be needed there.

cc @bpostlethwaite @etpinard


src/components/modebar/buttons.js
```diff
@@ -47,31 +49,31 @@ var modeBarButtons = module.exports = {};
 
 modeBarButtons.toImage = {
     name: 'toImage',
-    title: 'Download plot as a png',
+    title: function(gd) { return _(gd, 'Download plot as a png'); },
```
alexcjohnson: It's kind of annoying to have to turn all these button titles into functions... just to follow the rule that you can only translate a literal string. In principle we could do this translation in `ModeBar.createButton` by adding some complexity to the `find_locale_strings` task to allow one non-literal translation and look for keys inside `modebarButtons`.

src/plots/cartesian/dragbox.js
```diff
@@ -296,7 +296,7 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         dragTail(zoomMode);
 
         if(SHOWZOOMOUTTIP && gd.data && gd._context.showTips) {
-            Lib.notifier('Double-click to<br>zoom back out', 'long');
+            Lib.notifier(Lib._(gd, 'Double-click to<br>zoom back out'), 'long');
```
alexcjohnson: Seems funny to include `<br>` in the translation - should we 🔪  that and maybe just set notifier max width in css? (it's a `<div>` so we can do that 😄 )

src/traces/box/calc.js
```diff
@@ -122,7 +123,17 @@ module.exports = function calc(gd, trace) {
...
             posLetter: posLetter,
-            valLetter: valLetter
+            valLetter: valLetter,
+            labels: {
+                med: _(gd, 'median'),
+                min: _(gd, 'min'),
+                q1: _(gd, 'q1'),
+                q3: _(gd, 'q3'),
+                max: _(gd, 'max'),
+                mean: trace.boxmean === 'sd' ? _(gd, 'mean ± σ') : _(gd, 'mean'),
```
alexcjohnson: also not sure about making `mean` and `mean ± σ` separate translations... most languages will probably leave ` ± σ` intact, but some might want to change it (Greek?)

src/traces/heatmap/calc.js
```diff
@@ -82,21 +83,21 @@ module.exports = function calc(gd, trace) {
 
     function noZsmooth(msg) {
         zsmooth = trace._input.zsmooth = trace.zsmooth = false;
-        Lib.notifier('cannot fast-zsmooth: ' + msg);
+        Lib.notifier(_(gd, 'cannot use zsmooth: "fast"') + '<br>' + msg);
```
alexcjohnson: Do we really want this (and the 3 msgs below) to be notified? Can I just turn it into a console warning and leave it in English?

src/traces/ohlc/transform.js
```diff
@@ -132,6 +133,11 @@ exports.calcTransform = function calcTransform(gd, trace, opts) {
         close = trace.close,
         textIn = trace.text;
 
+    var openName = _(gd, 'Open') + ': ';
+    var highName = _(gd, 'High') + ': ';
+    var lowName = _(gd, 'Low') + ': ';
+    var closeName = _(gd, 'Close') + ': ';
```
alexcjohnson: Should `:` be part of the translation? (Applies a bunch of other places as well)

alexcjohnson: I suppose in principle one could make that argument about other punctuation as well (`,`, `(`, `)`...) but that seems like a can of worms we don't really want to open...

etpinard:
cc https://github.com/plotly/plotly.js/issues/856

src/traces/ohlc/transform.js
etpinard: > Should : be part of the translation?

I think so. For example, in french it is customary to add a nbsp before `:`, `;`, `!` and `?` e.g. `Open: 10` would be `Ouverture : 10`.

src/plots/cartesian/set_convert.js
```diff
@@ -381,9 +381,6 @@ module.exports = function setConvert(ax, fullLayout) {
         }
 
         if(!isFinite(ax._m) || !isFinite(ax._b)) {
-            Lib.notifier(
```
etpinard: Let me guess: you removed this so that we don't have to pass `gd` to `setConvert` ?

Regardless, I'm ok with making this just an (English-only) error :wink:

src/plots/plots.js
```diff
@@ -586,8 +606,8 @@ plots._hasPlotType = function(category) {
     var modules = this._modules || [];
 
     for(i = 0; i < modules.length; i++) {
-        var _ = modules[i];
-        if(_.categories && _.categories.indexOf(category) >= 0) {
```
etpinard: `modulei` is way better here. Thanks.

src/plots/plots.js
```diff
@@ -418,6 +419,25 @@ plots.supplyDefaults = function(gd) {
...
+    // These ones get used as default values so need to be known at supplyDefaults
+    // others keep their blank defaults but render the placeholder as desired later
+    // TODO: make these work the same way, only inserting the placeholder text at draw time?
+    // The challenge is that this has slightly different behavior right now in editable mode:
+    // using the placeholder as default makes this text permanently (but lightly) visible,
+    // but explicit '' for these titles gives you a placeholder that's hidden until you mouse
+    // over it - so you're not distracted by it if you really don't want a title, but if you do
+    // and you're new to plotly you may not be able to find it.
+    // When editable=false the two behave the same, no title is drawn.
+    newFullLayout._dfltTitle = {
```
etpinard: Thanks for the note :books:

src/traces/box/hover.js
```diff
@@ -162,7 +159,7 @@ function hoverOnBoxes(pointData, xval, yval, hovermode) {
 
         pointData2[vLetter + '0'] = pointData2[vLetter + '1'] = valPx;
         pointData2[vLetter + 'LabelVal'] = val;
-        pointData2[vLetter + 'Label'] = prefixes[i] + ': ' + Axes.hoverLabelText(vAxis, val);
+        pointData2[vLetter + 'Label'] = t.labels[attr] + ': ' + Axes.hoverLabelText(vAxis, val);
```
etpinard: Way better. thanks!

src/traces/heatmap/calc.js
etpinard: > Can I just turn it into a console warning and leave it in English?

I vote :+1:

etpinard: I think `Lib.notifier` should only be used for things that might affect the end user. 

`Lib.{log,warn,error}` should suffice for messages that only help developers debug .

src/traces/scattergeo/calc.js
```diff
@@ -40,5 +42,14 @@ module.exports = function calc(gd, trace) {
     calcMarkerColorscale(trace);
     calcSelection(calcTrace, trace);
 
+    if(len) {
+        calcTrace[0].t = {
+            labels: {
+                lat: _(gd, 'lat') + ': ',
+                lon: _(gd, 'lon') + ': '
```
etpinard: This is an interesting case here. One could argue that labels `lat` and `lon` could be configured using a layout attribute e.g.

```
  geo: {
    lonaxis: {
       hoverprefix: 'longitude' 
    },
    lataxis: {
       hoverprefix: 'latitude'
     }
  }
}
```



So I ask: where do we draw the line between configurable hover axis labels and dictionaries?



cc https://github.com/plotly/plotly.js/issues/265

tasks/find_locale_strings.js
```diff
@@ -0,0 +1,85 @@
...
+var srcGlob = path.join(constants.pathToSrc, '**/*.js');
+
+var common = require('./util/common');
+
+var EXIT_CODE = 0;
+
+var localizeRE = /(^|[\.])(_|localize)$/;
+
+// main
+findLocaleStrings();
```
etpinard: Lovely script :ok_hand:

test/jasmine/assets/supply_defaults.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var Plots = require('@src/plots/plots');
+
+/**
+ * supplyDefaults that fills in necessary _context
+ */
+module.exports = function supplyDefaults(gd) {
```
etpinard: :+1:

test/jasmine/tests/localize_test.js
```diff
@@ -0,0 +1,109 @@
...
+            expect(_(gd, 'c')).toBe('c-ctx');
+            expect(_(gd, 'd')).toBe('d-reg');
+            expect(_(gd, 'e')).toBe('e');
+        })
+        .catch(failTest)
+        .then(done);
+    });
+
+    it('does not generate an automatic base language dictionary in context', function(done) {
+        plot('fr', {'fr-QC': {fries: 'poutine'}})
```
etpinard: Test case of the year :trophy:

dist/translation-keys.txt
```diff
@@ -0,0 +1,60 @@
+Autoscale
```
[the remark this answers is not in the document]
alexcjohnson: > are we planning on having a set on official dictionaries

They're not going to be very big, and seems really useful so I'd vote yes - with the caveat that they would probably mostly have to be contributed by users. We might have to make use of google translate during review, as a sanity check...

Also it occurs to me now, some of these keys might not be obvious out of context - perhaps I should comment each with the file & line it came from? Something like:
```
kde        src/traces/violin/calc.js:73
```
and then we could comment there with more information `// kernel density estimate` (maybe I could even put that comment into `translation-keys.txt`?)

src/plots/cartesian/dragbox.js
alexcjohnson: > should we 🔪 that and maybe just set notifier max width in css?

Actually we already have it, and as currently set we just break at the next word:
screen shot 2017-12-07 at 2 51 29 pm 2

I'm inclined to take the `<br>` out here and just leave it to break where it will. We can of course still use `<br>` in cases where it conveys some meaning, but this is not such a case.

src/plot_api/plot_config.js
```diff
@@ -126,5 +126,19 @@ module.exports = {
 
     // Set global transform to be applied to all traces with no
     // specification needed
-    globalTransforms: []
+    globalTransforms: [],
+
+    // Which localization should we use?
+    // Should be a string like 'en' or 'en-US'.
+    locale: 'en',
```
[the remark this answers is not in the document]
alexcjohnson: > So should our default locale be en-US?

Ha sure -  might be important for dates too. Then I can make and pre-register `en` and `en-US` dictionaries to override these appropriately.

src/traces/scattergeo/calc.js
alexcjohnson: Seems to me if we were to allow configuring labels like this, it would work like the `dfltTitle` items - the translation result would be used as the default in `supplyDefaults`, but any other value provided with the figure gets used verbatim. We can't expect to translate user-provided strings, but for the most part it doesn't matter, usually translation is not about viewing other people's plots in your own language, but about making your own plot work and function the way you want it to.

src/plots/cartesian/dragbox.js
alexcjohnson: `<br>` removed in https://github.com/plotly/plotly.js/pull/2195/commits/5d6f26772191935a6026f0acad7efaa2b90d2dea

src/traces/heatmap/calc.js
alexcjohnson: moved these to `Lib.warn` and turned log level 1 on by default in https://github.com/plotly/plotly.js/pull/2195/commits/0f8520e846f4251b477e265fb12cfd4fa1d41509

see also https://github.com/plotly/plotly.js/issues/420#issuecomment-350083770

src/traces/ohlc/transform.js
alexcjohnson: moved colons into the translation keys in https://github.com/plotly/plotly.js/pull/2195/commits/bf2c6dbf8619718acd41ff0519f9c00abd0f62fb

src/plot_api/plot_config.js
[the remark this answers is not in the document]
alexcjohnson: default locale to `'en-US'`, with pre-registered dictionaries for this and `'en'`, in https://github.com/plotly/plotly.js/pull/2195/commits/04ed6f058443ddaa6782ebb3e6e2155e1e3a4bb2.
This also conveniently documents the locale module format.

dist/translation-keys.txt
```diff
@@ -1,60 +1,56 @@
...
+lower fence:                                          // traces/box/calc.js:134
+max:                                                  // traces/box/calc.js:132
+mean ± σ:                                             // traces/box/calc.js:133
+mean:                                                 // traces/box/calc.js:133
+median:                                               // traces/box/calc.js:128
+min:                                                  // traces/box/calc.js:129
+q1:                                                   // traces/box/calc.js:130
+q3:                                                   // traces/box/calc.js:131
+trace                                                 // plots/plots.js:439
+upper fence:                                          // traces/box/calc.js:135
```
alexcjohnson: @etpinard @bpostlethwaite thoughts on this format, with the (first) source line for each key included for a translator to use for more context?

Also having all these strings together raises some questions about whether they're all actually what we want, or if any could be made more consistent:
- Is the verb "Double-click" or "Double click"?
- Most of the hover label names (the items ending in `':'`) are lowercase, but some (OHLC, sankey) are capitalized - is this what we want?
- In making this PR I standardized the `"Click to enter <item> title"` ones to have `<item>` capitalized (previously one or two of them were lowercase) - is that correct?
- Most of the `Lib.notifier` items have no punctuation but some of them do. Should we standardize?

test/image/index.html
```diff
@@ -4,6 +4,9 @@
 <!-- this index file gets copied in to the image server docker -->
 <script type="text/javascript" src="../plotly.js/dist/extras/mathjax/MathJax.js?config=TeX-AMS-MML_SVG"></script>
 <script type="text/javascript" src="../plotly.js/build/plotly.js" charset="utf-8"></script>
+<script type="text/javascript">
+    Plotly.setPlotConfig({logging: 0});
+</script>
```
alexcjohnson: @etpinard I added this so `geo_choropleth-text` would render with no console logs (which trip the image server to throw an error). That way the image server keeps the same definition of "error" as it had before. I'm kind of inclined though to either:
1. Downgrade the `Location with id *** does not have a matching topojson feature at this resolution.` message from a `Lib.warn` to a `Lib.log`, or
2. Remove the offending ids from that mock. It's a lot though: ARB, BHR, CEB, CSS, EAP, EAS, ECA, ECS, EMU, EUU, FCS, HIC, HKG, HPC, LAC, LCN, LDC, LIC, LMC, LMY, MEA, MIC, MLT, MNA, NAC, NOC, OEC, OED, OSS, SAS, SGP, SSA, SSF, SST, UMC, WLD

Thoughts?

etpinard: How about both?

> Downgrade the Location with id *** does not have a matching topojson feature at this resolution. message from a Lib.warn to a Lib.log,

Yes, that should certainly be `Lib.log` :ok_hand: 

> Remove the offending ids from that mock. It's a lot though: ARB, BHR, CEB, CSS, EAP, EAS, ECA, ECS, EMU, EUU, FCS, HIC, HKG, HPC, LAC, LCN, LDC, LIC, LMC, LMY, MEA, MIC, MLT, MNA, NAC, NOC, OEC, OED, OSS, SAS, SGP, SSA, SSF, SST, UMC, WLD

I'd vote for removing the offending ids from the mock, but adding a jasmine test spying on `Lib.log` (if not done already). This part here could be done in a separate PR.

lib/locales.js
```diff
@@ -0,0 +1,14 @@
...
+* Copyright 2012-2017, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = [
```
etpinard: Interesting. I think registering only one local (en-US) in the dist bundles should suffice, but then again as only the spelling of colo(u)rscale differs between the two, including both en and en-US is fine by me :ok_hand:

dist/translation-keys.txt
etpinard: > thoughts on this format,

I like this format a lot. Nicely done :ok_hand: 

> Is the verb "Double-click" or "Double click"?

I have no preference.

> Most of the hover label names (the items ending in ':') are lowercase, but some (OHLC, sankey) are capitalized - is this what we want?

I'd vote for standardizing to lower case.

>  I standardized the "Click to enter <item> title" [...] is that correct?

fine by me.

> Most of the Lib.notifier items have no punctuation but some of them do. Should we standardize?

No preference.

lib/index-basic.js
```diff
@@ -15,4 +15,7 @@ Plotly.register([
     require('./pie')
 ]);
 
+// locales
+Plotly.register(require('./locales.js'));
```
etpinard: Maybe it would better to put these `Plotly.register` calls in [`src/core.js`](https://github.com/plotly/plotly.js/blob/master/src/core.js) so that we don't have to duplicate them?

That would also mean moving locale modules to a new `src/locales/` directory, as we wouldn't want to require `lib/` files from `src/` right?

src/traces/scattergeo/calc.js
etpinard: > Seems to me if we were to allow configuring labels like this, it would work like the dfltTitle items 

Good point. There's no downside to allowing `lat:`/`lon:` to be translated via dictionaries even if we add `{lat,lon}axis.hoverprefix` at some point.

tasks/bundle.js
```diff
@@ -54,3 +58,13 @@ constants.partialBundlePaths.forEach(function(pathObj) {
...
+
+// "Browserify" the locales
+var localeGlob = path.join(constants.pathToLib, 'locale-*.json');
+glob(localeGlob, function(err, files) {
+    files.forEach(function(file) {
+        var outName = 'plotly-' + path.basename(file).replace(/.json$/, '.js');
+        var outPath = path.join(constants.pathToDist, outName);
+        wrapLocale(file, outPath);
+    });
+});
```
alexcjohnson: @etpinard what do you think about this solution?
- Turned the locales into json files
- Explicitly loaded just `'en'` and `'en-US'` into the built bundles (ie I got rid of `locales.js` so that if/when we add more locales they will not automatically go into all the bundles) but we'll still put new locales into `lib/` for use if you make your own bundle.
- Built separate minified js files for each locale, so if you load via a script tag (from CDN or script) you can first load plotly.js and then in the next script tag load the locale you want.

lib/index-basic.js
```diff
@@ -16,6 +16,9 @@ Plotly.register([
 ]);
 
 // locales
-Plotly.register(require('./locales.js'));
+Plotly.register([
+    require('./locale-en.json'),
+    require('./locale-en-us.json')
```
etpinard: Ha. Webpack users won't like that. Webpack (v1 at least) needs a special loader to bundle JSON files unlike browserify. See https://github.com/plotly/plotly.js/pull/183

tasks/bundle.js
etpinard: > Turned the locales into json files

Requiring JSON files isn't ideal as described in https://github.com/plotly/plotly.js/pull/2195#discussion_r155848922 

I'd vote for converting those `lib/locale-???` files into plain js files. 

> so that if/when we add more locales they will not automatically go into all the bundles) but we'll still put new locales into lib/ for use if you make your own bundle.

great :ok_hand: 

> Built separate minified js files for each locale, so if you load via a script tag (from CDN or script) you can first load plotly.js and then in the next script tag load the locale you want.

Lovely :heart: and by the looks of it out [push-to-cdn](https://github.com/plotly/streambed/blob/master/bin/update_plotlyjs_cdn.sh) script should _just work_<sup>tm</sup>.

tasks/util/wrap_locale.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+module.exports = function wrap_locale(pathToInput, pathToOutput) {
+    intoStream(prefix)
+        .pipe(addStream(fs.createReadStream(pathToInput)))
+        .pipe(addStream(intoStream(suffix)))
+        .pipe(minify(constants.uglifyOptions))
+        .pipe(fs.createWriteStream(pathToOutput))
+        .on('finish', function() {
+            logger(pathToOutput);
+        });
```
etpinard: Smooth work with `add-stream` and `info-stream`. It would be nice to rewrite the [`addHeadersInDistFiles`](https://github.com/plotly/plotly.js/blob/6b4b89280c49eb70ffa4b43d37a57de0552d746c/tasks/header.js#L16-L42) step in `tasks/header.js` using streams for :racehorse:

lib/index-basic.js
alexcjohnson: ok, back to js https://github.com/plotly/plotly.js/pull/2195/commits/1275661c6c4d84cbbdf795f764623515b7c0d566

src/plots/command.js
```diff
@@ -114,7 +114,7 @@ exports.manageCommandObserver = function(gd, container, commandList, onchange) {
     } else {
         // TODO: It'd be really neat to actually give a *reason* for this, but at least a warning
         // is a start
-        Lib.warn('Unable to automatically bind plot updates to API command');
+        Lib.log('Unable to automatically bind plot updates to API command');
```
alexcjohnson: `updatemenus.json` emits this twice, and yet everything in its menus seems to work as expected... not sure what this is really about but since it doesn't (necessarily) mean anything is broken it seems more appropriate as a `log` than a `warn`.

etpinard: >  it seems more appropriate as a log than a warn.

I agree :ok_hand:

test/image/index.html
alexcjohnson: alright, did both.
- missing location downgraded to `Lib.log` https://github.com/plotly/plotly.js/pull/2195/commits/4b8f9802c3da064a3838c8a7757e170dac952bf3 (we have tests of these messages, modified in that commit)
- removed offending ids anyway https://github.com/plotly/plotly.js/pull/2195/commits/8b9e7d88def8fc3f887a33078cfbd119979c068d

dist/translation-keys.txt
alexcjohnson: I made all the hover labels lower case, standardized on "Double-click" (when used as a verb, which is the only way we use it in plotly.js text), and tweaked a few other messages -> https://github.com/plotly/plotly.js/pull/2195/commits/97821e912fe3c1363ddf1426529be7bd6eb5523e

jackparmer:
woot! A mere 95 files changed! 😄  📚

etpinard:
:dancer: nicely done

Issue #1546, opened 2017-04-03
plotly_doubleclick also triggers plotly_click


druvisc:
Double clicking on something drawn always triggers a 'plotly_click' event as well.

http://codepen.io/anon/pen/EWrrwW

In my actual scenario, the 'plotly_click' event almost always gets triggered (also when clicking on the plot's 'background' where nothing is drawn).
Is it possible to make it into an actual 'double click' event (adding a delay after the 'click' event)?

monfera:
@druvisc thanks for the report! It's indeed the case that currently, a `plotly_click` event is also raised. We _could_ make it so that it doesn't. But, since a double-click physically starts with a single mouse click, and within a short interval, followed by another click, we'd need to wait at least as much, before plotly.js could tell apart whether it's a single click, or just the first click in a double-click. `plotly_click` is now low latency and for some uses it can be beneficial.

Therefore we're planning to introduce a new plotly.js event called `plotly_singleclick` that would be of a higher latency than `plotly_click` because it first needs to know that it won't be a double-click, but helpful for general use cases when a couple hundred milliseconds of latency doesn't matter. We'd still keep around the current `plotly_click` as it works now, as well as of course `plotly_doubleclick`.

For the moment, this logic can be done on your end, either by capturing a `plotly_click` and not invoking your click callback if a `plotly_doubleclick` is subsequently registered, or follow a workaround by @etpinard where you detect a double-click event yourself: http://codepen.io/etpinard/pen/JWaEXG?editors=0010

etpinard:
Referencing other event discussions in https://github.com/plotly/plotly.js/issues/145

sardbaba:
@druvisc please take a look at the following code, which uses (and require) d3js:

```JavaScript
plotlyClickEvents() {
    /**
     * Distinguish between single and double click
     * @see http://bl.ocks.org/ropeladder/83915942ac42f17c087a82001418f2ee
     */
    function clickcancel() {
        var dispatcher = d3.dispatch('d3_plotly_singleclick', 'd3_plotly_doubleclick');
        function cc(selection) {
            var down, tolerance = 5, last, wait = null, args;

            // euclidean distance
            function dist(a, b) {
                return Math.sqrt(Math.pow(a[0] - b[0], 2) + Math.pow(a[1] - b[1], 2));
            }

            selection.on('mousedown', function() {
                down = d3.mouse(document.body);
                last = +new Date();
                args = arguments;
            });

[...]

PR #2200, opened 2017-12-12
Polar 2.0

etpinard:
should close everything listed under the on-par polar [milestone](https://github.com/plotly/plotly.js/milestone/2) :tada: 

--------------

I'm making a PR at this stage to:

- hopefully get a few :+1: s about the attributes came up with. I would appreciate if @alexcjohnson @cldougl @chriddyp could take a :eyes: at commits https://github.com/plotly/plotly.js/commit/3c8ae5beb88b32ab2158255a15eca5157fbb5b4e and https://github.com/plotly/plotly.js/commit/853581e3ede84aa412d17fdb3b5d43e2177f63ae,
- pin down the last open items, and
- set a firm base branch for @dfcreative 's upcoming `scatterpolargl` work.

### Working features so for:

- `scatterpolar` is on-par with `scattertenary` e.g. lines, markers, text, some fills options and line shapes as well as point selections are supported.
- `scatterpolar` accepts input angles in both degrees and radians using a polar-only `thetaunit` attribute
- `polar` subplots have fully configurable radial and angular axes styling.
- users can set the direction and starting angle of each `polar` subplots. (see [`polar_direction`](https://github.com/plotly/plotly.js/blob/543b61531073bb3f7f81157b8d28e56c2b495f46/test/image/baselines/polar_direction.png))
- `polar` subplots can be carve into sectors (see [`polar_sector`](https://github.com/plotly/plotly.js/blob/543b61531073bb3f7f81157b8d28e56c2b495f46/test/image/baselines/polar_sector.png))
- `polar` subplots can display angular tick labels in degrees or in fractions of pi.
- `polar` subplots accepts category as angles (see [`polar_categories`](https://github.com/plotly/plotly.js/blob/543b61531073bb3f7f81157b8d28e56c2b495f46/test/image/baselines/polar_categories.png))

Quick demo:

peek 2017-12-12 09-40

### Biggest open item: zoombox interactions

What should zoom interactions look like? At first, me and @alexcjohnson thought that a viewbox-based zoom system would suffice. In details, we thought about exposing a set of scalar attributes e.g. `polar.x`, `polar.y` and `polar.zoom` that determine the location and zoom-level of the polar subplots with respect to their `polar.domain` bounding-box. But there's a problem with such system: for some sets of `x/y/zoom` values, viewers won't see the subplot radial and/or angular axes - making data point only discernible via hover. 

While adding functionality for non-full `polar.sector` values, I thought that perhaps we could make zoombox interactions correspond to a combination of polar `sector` and radial axis range relayout updates. That is, zooming in on a point would correspond to carving a sector around it and narrowing the `radialaxis.range` (note that radial axis ranges that don't start at 0 are common in other libraries e.g. [matlab](https://www.mathworks.com/help/matlab/ref/rlim.html)). 

@nicolaskruchten pointed out that generalising `polar.sector` to some 4-pt polar bounding box attribute could allow for even better zooming interactions (we should probably draw that on a whiteboard and take a screenshot to make it easier to grasp).

So I ask: is `x/y/zoom` zooming more intuitive than some fancy sector/radial range algo?

----

cc'ing @jackparmer @cpsievert @monfera @bpostlethwaite who might also be interested.


src/plots/cartesian/axes.js
```diff
@@ -819,6 +823,13 @@ axes.calcTicks = function calcTicks(ax) {
         vals.push(x);
     }
 
+    // If same angle over a full circle, the last tick vals is a duplicate.
+    //
+    // TODO must do something similar for angular date axes.
+    if(ax._id === 'angular' && Math.abs(rng[1] - rng[0]) === 360) {
```
etpinard: I had to hack a few things deep into `axes.js` to get the angular axes looking the way I wanted. I hope @alexcjohnson won't mind.

src/plots/polar/layout_attributes.js
```diff
@@ -0,0 +1,327 @@
...
+        // similar to dtick, one way to achieve e.g.:
+        // - period that equals the timeseries length
+        //  http://flowingdata.com/2017/01/24/one-dataset-visualized-25-ways/18-polar-coordinates/
+        // - and 1-year periods (focusing on seasonal change0
+        //  http://otexts.org/fpp2/seasonal-plots.html
+        //  https://blogs.scientificamerican.com/sa-visual/why-are-so-many-babies-born-around-8-00-a-m/
+        //  http://www.seasonaladjustment.com/2012/09/05/clock-plot-visualising-seasonality-using-r-and-ggplot2-part-3/
+        //  https://i.pinimg.com/736x/49/b9/72/49b972ccb3206a1a6d6f870dac543280.jpg
+        //  https://www.climate-lab-book.ac.uk/spirals/
+    },
```
etpinard: Some screenshot of date angular axes:

image

image

image


image

src/plots/cartesian/axes.js
```diff
@@ -1905,7 +1905,7 @@ axes.doTicks = function(gd, axid, skipTitle) {
     var axLetter = axid.charAt(0),
         counterLetter = axes.counterLetter(axid),
         vals = axes.calcTicks(ax),
-        datafn = function(d) { return [d.text, d.x, ax.mirror].join('_'); },
+        datafn = function(d) { return [d.text, d.x, ax.mirror, d.font, d.fontSize, d.fontColor].join('_'); },
```
etpinard: Cartesian axes use a special [`ticks`](https://github.com/plotly/plotly.js/blob/599dff93f9923e7ef4d6446de734cfb4eb39d663/src/plot_api/subroutines.js#L484) update pathway that.

This here gets `Axes.doTicks` to update properly. I'm pretty sure some axis style updates are currently broken for ternary subplots. I'll lock those down in tests soon.

chriddyp:
🎉 🏆

alexcjohnson:
Re: zooming interaction: I think we may need two different modes, eventually if not right away, because there are two very different ways polar plots get used. One is where you're drawing a real 2D space (where zero is a unique point and shapes have meaning) just in polar coordinates. That's where you would want zoombox-type interactions that I was thinking of. I notice though that none of your examples are of this type (except maybe the cardioid). Your examples all have independent R and Theta variables, just mapped onto a polar space mostly to show periodicity. Incidentally, this points to another way the behavior may want to depend on this distinction: what do you do with negative R? If you're mapping a real space, then it should show up as a positive R in the opposite direction, but if you're not it should disappear at the origin.

So how should zooming work in the independent-variable case? I still need to look at some more example plots with the question "how would I want to zoom this to explore THIS data set", but as a first cut, I feel like the most useful for people would be to independently move each of the axis ends - so min and max radius (dragging them like you drag an axis end in cartesian? @etpinard points out that this is ambiguous with interacting with data near the origin though... not sure how to handle that.), and for angle I guess initial angle (rotating the whole plot) and period (so points at the initial angle would stay put, points at initial angle + period would move with the cursor, points at initial angle + twice the period would move twice as fast as the cursor... per private discussion with @etpinard this interaction should NOT happen when the period is physically significant, like 360 degrees or 2Pi - perhaps we can have an attribute to enable/disable it, with a smart default set depending on whether an explicit period is provided or not)?

I also liked the way our old polar plots had the radial axis on top of the data, with an outline around the text in the background color. Perhaps it could be configurable whether this axis is above or below the data?

etpinard:
> I also liked the way our old polar plots had the radial axis on top of the data, with an outline around the text in the background color. Perhaps it could be configurable whether this axis is above or below the data?

Absolutely. I'll add a `polar.radialaxis.layer` enumerated attribute. I could even make that the default if you prefer. Moreover, it's true that polar charts would benefit a lot from https://github.com/plotly/plotly.js/issues/1597

chriddyp:
> What should zoom interactions look like?

In antenna design, where `r` represents a meaningful amplitude in DBs, it seems like its helpful to be able to zoom just `r` in and out, i.e. change the "`rmin`" value: 
Here is the example from matlab

src/traces/scatterpolar/plot.js
```diff
@@ -10,33 +10,59 @@
...
+            var cdi = moduleCalcData[i][j];
+            var r = cdi.r;
+
+            if(r !== BADNUM) {
+                var rr = r - radialRange[0];
+                if(rr >= 0) {
+                    var rad = cdi.rad;
+                    cdi.x = rr * Math.cos(rad);
+                    cdi.y = rr * Math.sin(rad);
+                    continue;
```
etpinard: cc @dfcreative

src/plots/cartesian/axes.js
etpinard: :lock: in https://github.com/plotly/plotly.js/pull/2200/commits/7b0f718d8961794fc395ef249cdb6462c25914b9

src/plots/polar/constants.js
```diff
@@ -15,15 +15,17 @@ module.exports = {
...
-        'axes',
-        'lines',
-        'frontplot'
+        'angular-grid',
+        'radial-grid',
+        'frontplot',
+        'angular-axis',
+        'radial-axis',
+        'angular-line',
+        'radial-line'
```
etpinard: cc @alexcjohnson 's 

> I also liked the way our old polar plots had the radial axis on top of the data, with an outline around the text in the background color. Perhaps it could be configurable whether this axis is above or below the data?

from https://github.com/plotly/plotly.js/pull/2200#issuecomment-351127938

etpinard:
@alexcjohnson I just pushed about 20 commits where I:

- updated the attributes as requested in https://github.com/plotly/plotly.js/commit/3c8ae5beb88b32ab2158255a15eca5157fbb5b4e   and https://github.com/plotly/plotly.js/pull/2200/commits/853581e3ede84aa412d17fdb3b5d43e2177f63ae (where I just made a few comments), 
- added support for `radialaxis.title`, 
- merged master :sweat: and 
- fixed a few things while :palm_tree: a few other things

----

I'll spend a few hours tomorrow trying to get angular dates to work better with `angularaxis.period`. The main open item I believe is what to do with negative `r` values (referencing https://github.com/plotly/plotly.js/commit/3c8ae5beb88b32ab2158255a15eca5157fbb5b4e#r26416872). Other than that, this thing is pretty close.

alexcjohnson:
Currently if you zoom so some points are below the minimum radius, points disappear (as they should) but so do lines and fills (this is the polar_categories mock):
screen shot 2018-01-04 at 8 17 32 pm
Seems like the lines and fills should continue into the center, right? That would be as analogous as we can get to how cartesian handles off-plot points.

alexcjohnson:
Radial axis dragging seems to be circling the wrong point:
kapture 2018-01-04 at 21 46 18
Also we need a cursor or something (`move`? `grab` with fallbacks for IE?) to show its grabber.

alexcjohnson:
I don't see where you're generating fractions of Pi, but at some precision (denominator >= 100?) it should switch to decimal - decimal fractions of Pi? Also should hovertext have these Pi fractions/decimals too instead of raw decimal numbers?
screen shot 2018-01-04 at 10 29 16 pm

alexcjohnson:
I find the radial zoom effect a lot of fun but a bit confusing to operate. I always know what's going to happen when I mouse up, that part is great - but I don't always know where to put my cursor to get what I want. It sometimes swaps between keeping the donut, keeping the hole, and keeping the whole thing, and it's not obvious to me when and why it's going to change. Might help to have some kind of handle for the starting point, where you moused down? And possibly some sort of guides delineating regions with different results, or just a clearer rule about which is which? 

Perhaps it could be as simple as: take the ray from the center toward where you moused down; the 180 degree sector clockwise from that ray gives you the donut between mouse down and mouse up, the sector ccw from that ray gives you a new outer radius only (set by where you mouse up, ignoring where you moused down).

It's also not obvious to me how to cancel the zoom in case I discover I've started it badly or something - which in cartesian I can do by making the box very small so it disappears. Sometimes I can't find any place to put the cursor to do this. Perhaps if there was a starting handle, I would be tempted to do the same thing as in cartesian, bring my cursor close to that point and the box would disappear?

src/plots/cartesian/axes.js
```diff
@@ -974,10 +993,15 @@ axes.autoTicks = function(ax, roughDTick) {
         ax.tick0 = 0;
         ax.dtick = Math.ceil(Math.max(roughDTick, 1));
     }
+    else if(ax._id === 'angular') {
+        ax.tick0 = 0;
+        base = getBase(1);
```
alexcjohnson: what am I missing? Isn't `getBase(1)` just 1^something === 1?

src/plots/cartesian/axes.js
alexcjohnson: What if you use something else as the period of a linear angular axis? Also do we need to do something more flexible to handle *nearly* duplicate values so we don't get overlaps?

src/plots/cartesian/axes.js
```diff
@@ -1399,6 +1424,66 @@ function formatLinear(ax, out, hover, extraPrecision, hideexp) {
     out.text = numFormat(out.x, ax, hideexp, extraPrecision);
 }
 
+function formatAngle(ax, out, hover, extraPrecision, hideexp) {
```
alexcjohnson: Ah nice, here it is 🎉  I guess even though it's currently only used for polar, there *is* something nice about having this in `axes.js` (granted that perhaps all of this should go somewhere else outside cartesian at some point) because I could imagine us wanting to support fractions and/or multiples of π on cartesian axes at some point.

But yeah, previous comments hold: we need a cutoff to switch to decimal, and then perhaps hover can use π too!

alexcjohnson:
It wasn't obvious to me at first that `pan` mode meant "rotate the plot" - I just stumbled on rotating and wondered why I couldn't zoom anymore. Maybe that's OK, but alternatively, perhaps you could grab in the ring of the angular axis labels to rotate, without going into a separate pan mode? Also it would be neat if it would snap to straight orientations, ie with zero angle at N, S, E, or W.

etpinard:
@alexcjohnson I think all your comments from last week (starting after https://github.com/plotly/plotly.js/pull/2200#issuecomment-355406922) have been addressed. I didn't push the improvements I've made to drag interactions to this branch yet (I'm looking for some feedback first :wink: ), they are [here](https://github.com/plotly/plotly.js/compare/polar...polar-dev).

Here's the new angular drag layer in action:

peek 2018-01-11 18-24

where I kept the `crosshair` cursor on the radial drag layer, and made the angular layer have `move`, to not confused them as they partially overlap.

------------

Concerning `dragmode: 'zoom'`, I added a few ways to _cancel_ the zoom event:

for example, if we did reach min zoom for two-sided re-range level and then moved back in the opposite direction reaching the mouse down radial distance, the zoom event is cancelled:

peek 2018-01-11 18-27

if we did cross the origin, we get:

peek 2018-01-11 18-30



Other than that, `dragmode: 'zoom'` behaves the same as previously. I hoping to add way to (easily) cancel zoom events without over-complicating things. I hope you'll like the results.

alexcjohnson:
Nice! I'll check out the other changes later tonight, but re: drag interactions:
- Angular - love it. Only addition I'd like at some point is snap to NSEW (for both the angular orientation and radial line orientation) but that can go in a TODO issue.
- Radial - definite improvement but I still find it confusing. I don't love the statefulness, ie if you start dragging inward it behaves one way, outward a different way, crossing the origin changes what you can do later... I feel like we can find a solution that only depends on the start and end points of the drag. Also wonder if we can get away from all angular dependence and only care about initial and final radii? It just always seems like angular dependence will give weird invisible boundaries with unexpected effects when you cross them. Sorry to be a pain about this one, it's a hard problem I don't have real clear answers on, just kind of thinking out loud and I'm probably going to contradict some suggestions I made earlier:

There are four radial range change you might want: zoom to a range, give just a new outer bound, give just a new inner bound, or abort.
- Abort: can we just say `Math.abs(r1 - r0) < MINDRAG` means abort?
  - The limitation this imposes is that you can't zoom straight to a tiny slice of the original radial range. But you can always zoom multiple times if that's what you want, and that's the same limitation we have with cartesian zoom.
- New outer bound: how about starting OR ending the drag within a certain region around the center (perhaps again `MINDRAG` px? Or that might be too small, we might want `2 * MINDRAG`)? So then you could either drag from the center outward to your new outer bound, or start at your new outer bound and drag to the center.
  - The limitation this imposes is that you can't make a very small tweak to the minimum bound, because it'll snap to the existing minimum. That doesn't strike me as a big deal though, it's a fairly weird case. And once we have `hole` as an available option (which is mostly when you'd want a variable inner bound anyhow!), we can remove snapping for the drag endpoint because any endpoint inside the hole counts as "no change to the minimum". So you can drag from the outside inward to an arbitrary number of pixels away from the edge of the hole without snapping to the edge. Seems though like the start of the drag will still need to snap to the hole edge, otherwise you can't ever hit the edge exactly.
- New inner bound: similar to new outer bound, either drag from the new edge you want all the way to (beyond) the outer edge, or drag from near the outer edge (snap to the edge) to the new edge you want.
- Zoom to a range is anything not covered by the above: your starting and ending radii are the new range.

src/plots/polar/layout_attributes.js
```diff
@@ -211,16 +210,15 @@ var angularAxisAttrs = {
...
-        dflt: 0,
         editType: 'calc',
         role: 'info',
         description: [
             'Sets that start position (in degrees) of the angular axis',
-            'Note that by default, polar subplots are orientation such that the theta=0',
-            'corresponds to a line pointing right (like what mathematicians prefer).',
-            'For example to make the angular axis start from the North (like on a compass),',
-            'set `angularaxis.rotation` to *90*.'
+            'By default, polar subplots will `direction` set to *counterclockwise*',
```
alexcjohnson: ~~will~~ with

src/plots/polar/layout_defaults.js
```diff
@@ -113,6 +107,26 @@ function handleDefaults(contIn, contOut, coerce, opts) {
...
 
             case 'angularaxis':
+                // We do not support 'true' date angular axes yet,
+                // users can still plot dates on angular axes by setting
+                // `angularaxis.type: 'category'`.
+                //
+                // Here, if a date angular axes is detected, we make
+                // all its corresponding traces invisible, so that
+                // when we do add support for data angular axes, the new
+                // behavior won't conflict with existing behavior
```
alexcjohnson: Beautifully done. Where should we document the use of categories for displaying polar dates? Just in an example somewhere? I'm thinking that doing it really right would involve categories like "Jan 2010", "Feb 2010", ... "Jan 2011" etc., `angularaxis.period` and then `tickvals`/`ticktext` to get labels to just be "Jan", "Feb", "Mar"...

test/jasmine/tests/polar_test.js
```diff
@@ -232,8 +232,21 @@ describe('Test relayout on polar subplots:', function() {
...
+        .then(function() {
+            expect(gd._fullLayout.polar._subplot.viewInitial['radialaxis.range'])
+                .toBeCloseToArray([-0.53, 1.158]);
+
+            return Plotly.relayout(gd, 'polar.radialaxis.type', 'linear');
+        })
+        .then(function() {
+            expect(gd._fullLayout.polar._subplot.viewInitial['radialaxis.range'])
+                .toBeCloseToArray([0, 11.225]);
+        })
```
alexcjohnson: Good catch! Just for my peace of mind so I don't have to trust that `viewInitial` does the right thing, can you also `expect` that `fullLayout.polar.radialaxis.range` matches `viewInitial[...range]`?

src/traces/scatter/line_points.js
```diff
@@ -60,9 +60,10 @@ module.exports = function linePoints(d, opts) {
 
     // turn one calcdata point into pixel coordinates
     function getPt(index) {
-        var x = xa.c2p(d[index].x);
-        var y = ya.c2p(d[index].y);
-        if(x === BADNUM || y === BADNUM) return false;
+        var di = d[index];
+        var x = xa.c2p(di.x);
+        var y = ya.c2p(di.y);
+        if(x === BADNUM || y === BADNUM) return di.intoCenter || false;
```
alexcjohnson: Very clean solution, love it!

test/image/mocks/polar_fills.json
```diff
@@ -13,13 +13,37 @@
     "theta": [ 0, 90, 180, 270, 350, 360 ],
     "fill": "toself",
     "hoveron": "fills"
+  }, {
+    "type": "scatterpolar",
+    "mode": "lines",
+    "r": [5, 1, 2, 4, 5],
```
alexcjohnson: This test doesn't show that lines go to the center while markers are not displayed there (though I've seen in playing with it that you indeed got it right 🎉 ) - can you just show markers here, and bump the r=2 value down so no marker *should* show at the center?

src/plot_api/plot_api.js
```diff
@@ -1946,7 +1946,7 @@ function _relayout(gd, aobj) {
 
                 // clear polar view initial stash for radial range so that
                 // value get recomputed in correct units
-                if(fullLayout._subplots.polar && fullLayout[p.parts[0]]) {
+                if(Array.isArray(fullLayout._subplots.polar) && fullLayout._subplots.polar.length && fullLayout[p.parts[0]]) {
```
alexcjohnson: Oh, I didn't notice this logic before - should it be further restricted to when the axis (I guess this is `p.parts[1]`) is `radialaxis`?

src/plots/polar/polar.js
```diff
@@ -807,6 +788,23 @@ proto.updateMainDrag = function(fullLayout, polarLayout) {
...
+            var updateObj = {};
+            for(var k in _this.viewInitial) {
+                updateObj[_this.id + '.' + k] = _this.viewInitial[k];
+            }
+
+            gd.emit('plotly_doubleclick', null);
+            Plotly.relayout(gd, updateObj);
+        }
+
+        Fx.click(gd, evt, _this.id);
```
alexcjohnson: Hmm, this is not specific to polar, you're just replicating what we have elsewhere... but it occurs to me, having the `doubleClick` handled before `click` (in `Fx.click`) is both different from native events (which always seem to put the `doubleClick` last, though some browsers - [*cough* IE *cough*](https://stackoverflow.com/questions/5497073/how-to-differentiate-single-click-event-and-double-click-event) - don't emit the second `click` at all) AND potentially incorrect: if `Plotly.relayout` is synchronous, then `Fx.click` will make it look like you clicked on whatever point happens to be under your cursor after autoranging.

test/jasmine/tests/polar_test.js
```diff
@@ -496,3 +499,153 @@ describe('Test relayout on polar subplots:', function() {
...
+            assertEventCnt({
+                plotly_hover: 2,
+                plotly_unhover: 1,
+                plotly_click: 3,
+                plotly_doubleclick: 1,
+                plotly_relayout: 1
+            }, 'after doubleclick');
+        })
+        .catch(fail)
+        .then(done);
```
alexcjohnson: Great test, very readable!

Presumably polar also handles right-clicks the same way as cartesian? Want to 🔒 it ala https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/click_test.js#L173-L213?

src/plots/polar/polar.js
etpinard: There's https://github.com/plotly/plotly.js/issues/1546 that discuss the subject where @monfera argued for a new event called `plotly_singleclick` to go around this problem. Not sure what's best.

src/plots/polar/layout_defaults.js
etpinard: In https://github.com/plotly/plotly.js/pull/2200/commits/8b4c2b5bd9e4303599f3a407a7261bb4c3a31708, I restricted the period for category axes to be `>=` than the number of categories, as I couldn't think of a use case for winding category labels. I don't think they make sense **unless** the users set angular `tickvals`/`ticktext`. That might not be too much to ask though :thinking: 

So, we could allow `period < ax._categories.length` to make date-like categories work with a period less than the data range or simply wait for a true official handling of angular dates. I'm not sure it's worth the trouble, but I'll leave it up to you to decide.

test/jasmine/tests/polar_test.js
etpinard: Good call. Fixed in https://github.com/plotly/plotly.js/pull/2200/commits/473643cf8dfb12f1543bd8dcb01ca00c568f6398

test/image/mocks/polar_fills.json
etpinard: See https://github.com/plotly/plotly.js/pull/2200/commits/473643cf8dfb12f1543bd8dcb01ca00c568f6398
for what I came up with. I'm not 100% sure this is what you meant though.

src/plot_api/plot_api.js
etpinard: Good call -> https://github.com/plotly/plotly.js/pull/2200/commits/473643cf8dfb12f1543bd8dcb01ca00c568f6398

test/jasmine/tests/polar_test.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2200/commits/473643cf8dfb12f1543bd8dcb01ca00c568f6398

etpinard:
Another day, another polar zoombox attempt, see https://github.com/plotly/plotly.js/compare/polar...polar-dev

@alexcjohnson First off, thanks very much for the long writeup in https://github.com/plotly/plotly.js/pull/2200#issuecomment-357117251, I think this is by far the best version so far. This gif should cover all the cases:

peek 2018-01-12 14-18

alexcjohnson:
> Another day, another polar zoombox attempt, I think this is by far the best version so far.

Love it, I think we have a winner! 🏆 Particularly nice how if you start at one end, there's no snap to the other end - because, really, why would you do that, it would be a noop. That gets around my concern about not being able to make a small change to the inner radius. Very very nice.

I could imagine decreasing the snap length a little bit, at least for small plots. But we can play with that later, the hard part is the mechanics and I think you've nailed it.

src/plots/polar/layout_defaults.js
alexcjohnson: I wouldn't *prohibit* `period < ax._categories.length` - someone will find a use for it for sure. As long as ticks only display for the first revolution, ideally with the ability to simplify them with `tickvals`/`ticktext`, and hovertext displays the actual angular data value so you know which revolution you've hovered on. But it's also a weird case so I'm fine with leaving this feature in a polar TODO issue.

test/image/mocks/polar_fills.json
alexcjohnson: Yep, you got it, thanks. I was just looking to see a corner in the line at the origin, with no corresponding marker. Lo and behold, there it is!
screen shot 2018-01-12 at 5 53 38 pm

src/plots/polar/polar.js
alexcjohnson: Ah thanks for bringing up that issue - I'll just add this comment there.

src/plots/cartesian/axes.js
etpinard: good catch -> 7f7193d

src/plots/polar/constants.js
```diff
@@ -33,6 +33,9 @@ module.exports = {
     cornerLen: 25,
     cornerHalfWidth: 2,
 
-    MINDRAG: 8,
-    MINZOOM: 20
+    // smallest radial distance [px] allowed for a zoombox
+    MINZOOM: 20,
+    // distance [px] off (r=0) or (r=radius) where we transition
+    // from single-sided to two-sided radial zoom
+    OFFEDGE: 20
```
etpinard: per https://github.com/plotly/plotly.js/pull/2200#issuecomment-357393518

> I could imagine decreasing the snap length a little bit

I made this smaller than last Friday's version.

test/jasmine/tests/scatterpolar_test.js
```diff
@@ -31,14 +39,80 @@ describe('Test scatterpolar trace defaults:', function() {
...
+    }, {
+        desc: 'on a `thetaunit: radians` polar subplot',
+        patch: function(fig) {
+            fig.layout.polar.angularaxis.thetaunit = 'radians';
+            return fig;
+        },
+        nums: 'r: 4.022892\nθ: 2.239991',
+        name: 'Trial 3'
+    }, {
+        desc: 'on work on log radial axis',
```
alexcjohnson: ~on work~ on log radial axis

src/plots/polar/polar.js
```diff
@@ -1071,6 +1093,31 @@ function pathSectorClosed(r, sector) {
...
 
+// inspired by https://gist.github.com/buschtoens/4190516
+function pathAnnulus(r0, r1, _sector) {
+    var sector = _sector.slice();
+    var s0 = deg2rad(sector[0]);
+    var s1 = deg2rad(sector[1]);
+
+    // make a small cut on full sectors so that the
+    // inner region isn't filled
+    if(isFullCircle(sector)) s1 -= 1e-3;
```
alexcjohnson: FWIW pies solved this by making [two arcs at each radius](https://github.com/plotly/plotly.js/blob/ef181daa3f18483af769cb776059b9709b17311b/src/traces/pie/plot.js#L216-L219) so the annulus really is complete.

*Probably* not going to cause any issues, it's just missing a fraction of a pixel, but I worry a little about numerical stability - it has to determine the center of the circle based on the very small vector from `p00` to `p01` and I'm not sure how much precision SVG uses.

alexcjohnson:
Two problems I just encountered with interaction effects, both while looking at the polar_direction mock:
- Rotating the plot by dragging the angular axis: seems like this is centered on the wrong point (though dragging the radial axis around has the correct center now)
- Dragging the radial axis around: *sometimes* after I've done this once, it doesn't seem to allow it anymore, I can only change the outer limit of the radius no matter which direction I start the drag. But sometimes it still works indefinitely... I can't really figure out the pattern.

src/plots/polar/constants.js
alexcjohnson: > I made this smaller than last Friday's version.

Looks good. I wish there was a way to indicate to the user which behavior they're going to get, like we do with single- or double-ended arrows for the cartesian axis draggers... but I don't see any obvious choices among built-in cursors. I suppose we could do some super-complicated thing where we choose among all eight single-ended arrows depending on which direction you are from the center... but that might even be more confusing since it would make it look like those eight sectors had differences in behavior. Lets leave it as is for now, and see how it feels as it starts getting used.

src/plots/polar/polar.js
```diff
@@ -911,9 +916,19 @@ proto.updateAngularDrag = function(fullLayout, polarLayout) {
             strTranslate(_this.xOffset2, _this.yOffset2) + strRotate([-da, cxx, cyy])
         );
 
-        _this.clipPaths.circle.select('path').attr('transform',
-            strTranslate(cxx, cyy) + strRotate(da)
-        );
+        // 'un-rotate' marker and text points
```
alexcjohnson: Oh man, that is such a satisfying effect 🍹

alexcjohnson:
@etpinard Fantastic work! Aside from the three quick comments above (at least I hope the interaction bugs are quick 😅 ) the only thing I'd like to see before merging is an issue collecting open items, so we can point people to that when questions arise, and use it to prioritize future additions.

src/components/dragelement/index.js
```diff
@@ -76,6 +76,10 @@ dragElement.unhoverRaw = unhover.raw;
  *          numClicks is how many clicks we've registered within
  *          a doubleclick time
  *          e is the original mousedown event
+ *      clampFn (optional, function(dx, dy) return [dx2, dy2])
+ *          Provide custom clamping function for small displacements.
+ *          By default, clamping is done using `minDrag` to x and y displacements
+ *          independently.
```
alexcjohnson: Nicely done. And it works great to solve the radial/angular drag bug! 🎯

alexcjohnson:
A hard-fought 💃  - excellent work.

etpinard:
Merging. 

For future development ideas and updates, please subscribe to https://github.com/plotly/plotly.js/issues/2255

etpinard:
Oh and please note that (r,t) _legacy_ polar charts are now deprecated! Please switch over to `scatterpolar`.

src/plots/polar/layout_attributes.js
srpn: Great feature !! Can you please include the code (layout config attributes) for the above example screenshots?

Issue #1842, opened 2017-07-03
Per-axis separators attribute


ndabAP:
```javascript
var trace = {
  x: ['2000-01-30', '2000-01-31'],
  y: [1.5, 1.7],
  type: 'scatter' 
};

var data = [trace];

var layout = {
  xaxis: {
    type: 'date',
  },
  yaxis: {
    tickformat: ',.'
  },
};

Plotly.plot('myDiv', data, layout);
```
Last two axis labels are wrong and the comma format isn't applied. The problem doesn't occur when setting `layout.separators = ',.'`.

https://codepen.io/anon/pen/LLmWpm

etpinard:
What _should_ it do? It's not obvious to me from reading the [docs](https://github.com/d3/d3-format/blob/master/README.md#locale_format).

You might be looking for [`layout.separator`](https://plot.ly/javascript/reference/#layout-separators) attribute.

ndabAP:
@etpinard, last two axis labels should be 1,65 and 1,7 not 000000 ...

What if my x axis values should be comma decimal and my y axis should be dot decimal notation? E. G.:

x: 1.5, 1.7
y: 2.1, 2.5

should be displayed as

x: 1,5, 1,1
y: 2.1, 2.5

Is this possible?

etpinard:
> Is this possible?

Yes, using `layout.separtors` as in https://codepen.io/etpinard/pen/awKoBP?editors=0010

ndabAP:
@etpinard, thanks for your answer. But it seems that it is not possible to set the separators independently per axis, e. g.  `yaxis2` has a different separator than `yaxis`. Is this correct?

etpinard:
> But it seems that it is not possible to set the separators independently per axis, e. g. yaxis2 has a different separator than yaxis. Is this correct?

That's correct. `layout.separators` is a global setting and there's no corresponding per-axis attribute.

Can I ask? Why do you different separator styles on different axes?

ndabAP:
> Why do you different separator styles on different axes?

Sure. The user within my application can choose their column format which can be a comma or a dot decimal notation. They can upload data and visualize it. Maybe one source of the data has dot, another comma decimal notation. Either way, the user can choose their format.

etpinard:
> Maybe one source of the data has dot, another comma decimal notation.

Hmm. But `layout.seperators` is meant to control the output format of those numbers on the graph. Wouldn't you want that to be consistent?

[...]

PR #2207, opened 2017-12-15
Date and number localization

### LINKED ISSUE #1264: hoverformat/tickformat overwrites separators
It's not possible to set the separators of thounds and comma when also setting the tickformat or hoverformat.

Settings:
```js
data = [{x:[0],y[2323.45]]
layout={
  separators: ',.',
  yaxis: {
    hoverformat: ',.2r'
  }
}
```

Expected output:
When hover, following number should be printed: "2.323,45".

Example: [Codepen](http://codepen.io/anon/pen/zobmyW)

I want to reformat ticks and hover to not show "k" as thounds, but also keep the localized thousand separator and comma.

alexcjohnson:
- Adds date and number localization to the string localization framework of #2195 
- Uses `d3.locale` format
- Pulls in and cleans up all the information we already have in `world-calendars`
  - None of these localizations are built into the bundle, but they are provided as requireable modules in `lib` and as standalone scripts in `dist`, as started by #2195.
  - Does NOT provide localizations for non-gregorian calendars. `world-calendars` *does* have a few of these, but mostly these calendars are only in use in one region and are already localized to this region. If there is a need we can address this later.
  - Some required information for `d3.locale` is not available in `world-calendars`: AM/PM translations (the `period` field), `dateTime` and `time` format fields (only relevant for custom formats anyway so this doesn't seem like a big deal) and, most importantly, number formatting (`decimal` and `thousands`, we only have partial support for `groupings` and `currency` right now but they should be included too if they're different from English). Now that we have these localization skeletons we can encourage community members to fill in the rest, and even add more.
- Fixes #1264 - and incidentally (as noted in a TODO) in V2 we could consider removing `layout.separators` altogether and *only* using locales.

cc @etpinard @VeraZab


src/core.js
```diff
@@ -65,6 +65,12 @@ exports.register([
     require('./components/rangeselector')
 ]);
 
+// locales en and en-US are required for default behavior
+exports.register([
+    require('./locale-en'),
+    require('./locale-en-US')
+]);
```
alexcjohnson: 🎩  `bundle_tests/core_test` 🌟 

Since we're constructing our own `d3.locale` now no matter what, `locale-en` needs to be part of the core so it can be used as the fallback for missing pieces of other locales. Then for simplicity I moved the default `locale-en-US` to the core as well, which has the nice side-effect of keeping these redundant locales out of `dist/`.

@etpinard seem reasonable? Are you OK with putting these two in the root `src/` directory? At least as long as we're not going to put any other locales into the core it doesn't seem like they warrant their own directory.

Minor downside is the `lib/index*` files don't demonstrate how to register locales, but we should be able to to better than documenting it there anyhow - somewhere we will want examples of how to use locales both in the build process and via script tags.

src/lib/dates.js
```diff
@@ -433,10 +433,12 @@ function formatTime(x, tr) {
...
-var yearFormat = utcFormat('%Y'),
-    monthFormat = utcFormat('%b %Y'),
-    dayFormat = utcFormat('%b %-d'),
-    yearMonthDayFormat = utcFormat('%b %-d, %Y');
+// TODO: do these strings need to be localized?
+// ie this gives "Dec 13, 2017" but some languages may want eg "13-Dec 2017"
+var yearFormatD3 = '%Y';
+var monthFormatD3 = '%b %Y';
+var dayFormatD3 = '%b %-d';
+var yearMonthDayFormatD3 = '%b %-d, %Y';
```
alexcjohnson: This is for our default, adaptive date labels. I'm not too worried about it, as these formats are at least unambiguous. But it shouldn't be too hard to allow these to be localized too, essentially as an extension to the `d3.locale` format objects, should our international users feel strongly about it (and contribute their own alternative format strings).

src/plot_api/plot_config.js
```diff
@@ -137,11 +137,20 @@ module.exports = {
...
+    dictionaries: {},
+
+    // Localization specs for dates and numbers
+    // Each localization should be an object with keys matching most of d3.locale,
+    // see https://github.com/d3/d3-3.x-api-reference/blob/master/Localization.md
+    // {'da': {months: [...], shortMonths: [...], ...}, ...}
+    // Unlike d3.locale, every key is optional, we will fall back on English ('en').
+    // Currently `grouping` and `currency` are ignored for our automatic number
+    // formatting, but can be used in custom formats.
+    formats: {}
```
alexcjohnson: I suppose we could combine these into one, so context looks more like the register-able locale modules. I didn't do that because I wanted to preserve compatibility with #2195 but that hasn't been released to npm yet so we *can* change it if we want. Thoughts?

src/plots/cartesian/set_convert.js
```diff
@@ -447,9 +447,19 @@ module.exports = function setConvert(ax, fullLayout) {
     ax._min = [];
     ax._max = [];
 
-    // copy ref to fullLayout.separators so that
+    // Fropagate localization into the axis so that
```
alexcjohnson: I love to Fropagate whenever I can ;)

src/plots/plots.js
```diff
@@ -563,6 +567,83 @@ function remapTransformedArrays(cd0, newTrace) {
...
+            }
+            formats = Registry.formatRegistry;
+        }
+
+        var baseLocale = locale.split('-')[0];
+        if(formatDone || baseLocale === locale) break;
+        locale = baseLocale;
+    }
+
+    // lastly pick out defaults from english (non-US, as DMY is so much more common)
```
etpinard: I strongly agree with picking `en` (not `en-us`) as our default :ok_hand:

src/plots/plots.js
```diff
@@ -563,6 +567,83 @@ function remapTransformedArrays(cd0, newTrace) {
...
+    if(!formatDone) includeFormat(Registry.formatRegistry.en);
+
+    return formatObj;
+}
+
+/**
+ * getFormatter: combine the final separators with the locale formatting object
+ * we pulled earlier to generate number and time formatters
+ * TODO: remove separators in v2, only use locale, so we don't need this step?
```
etpinard: I'm in strong agreement here too :ok_hand:

src/core.js
```diff
@@ -68,7 +68,7 @@ exports.register([
 // locales en and en-US are required for default behavior
 exports.register([
     require('./locale-en'),
-    require('./locale-en-US')
+    require('./locale-en-us')
```
alexcjohnson: how did this not fail for me locally but only on CI?

src/registry.js
```diff
@@ -328,23 +329,38 @@ function getTraceType(traceType) {
  *  the dictionary mapping input strings to localized strings
  *  generally the keys should be the literal input strings, but
  *  if default translations are provided you can use any string as a key.
+ * @param {object} module.format
+ *  a `d3.locale` format specifier for this locale
+ *  any omitted keys we'll fall back on en-US
```
etpinard: Hmm. Isn't the fallback `en` ?

tasks/pull_date_format.js
```diff
@@ -0,0 +1,169 @@
+var path = require('path');
+var fs = require('fs');
+var constants = require('./util/constants');
+var common = require('./util/common');
+
+var args = process.argv.slice(2);
+var argLocale = args[0];
+
+var pathToEn = path.join(constants.pathToLib, 'locale-en.js');
+var pathToWCRegions = path.join(__dirname, '../node_modules/world-calendars/dist/regional');
```
etpinard: `path.join(require.resolve('world-calendars'), 'dist', 'regional')`

would be a bit more robust

tasks/pull_date_format.js
```diff
@@ -0,0 +1,169 @@
...
+            format.shortMonths = wcDataObj.monthNamesShort;
+            format.date = translateDateFormat(wcDataObj.dateFormat);
+            // leave out `dateTime` and `time` fields. We could include them as cmd line args later
+            // for now these will get inherited from english.
+
+            var outStr = JSON.stringify(outObj, null, 4)
+                .replace(/'/g, '\\\'') // escape single quotes
+                .replace(/"([A-Za-z]\w+)":/g, '$1:') // unquote simple identifier keys
+                .replace(/([^\\])"/g, '$1\'') // replace unescaped doublequotes with singlequotes
+                .replace(/\\"/g, '"'); // unescape escaped doublequotes
```
etpinard: Nasty. That most have been painful to do. :clap:

tasks/pull_date_format.js
```diff
@@ -0,0 +1,169 @@
...
+
+    return out;
+}
+
+/**
+ * Either put the whole array on one line, split it into 2 lines,
+ * or leave it as is (one line per item), whatever we can fit into 90 characters.
+ * A trailing comma may still be after the 90-character limit
+ */
+function shortenArrays(s) {
```
etpinard: That's brilliant :sparkles: 

Maybe down the road, we could factor this out and use it on our `test/image/mocks/` JSONs to make them a bit more readable.

tasks/pull_date_format.js
```diff
@@ -0,0 +1,169 @@
...
+        });
+    });
+}
+else {
+    pullOneLocale(argLocale);
+}
+
+function pullOneLocale(locale) {
+    var pathToInput = path.join(pathToWCRegions, locale + '.js');
+    var pathToOutput = path.join(constants.pathToLib, 'locale-' + locale.toLowerCase() + '.js');
```
etpinard: I wonder if putting all the locales in a new subdirectory `lib/locales/` (and `dist/locales/`) would make things more user-friendly?

src/core.js
etpinard: > @etpinard seem reasonable?

This is very reasonable :ok_hand: 

> but we should be able to to better than documenting it there anyhow 

Yep, we should at least open up an issue on https://github.com/plotly/documentation/issues and / or add a _Locale_ section to `dist/README.md` via [`tasks/stats.js`](https://github.com/plotly/plotly.js/blob/master/tasks/stats.js).

src/plots/cartesian/set_convert.js
```diff
@@ -447,7 +447,7 @@ module.exports = function setConvert(ax, fullLayout) {
     ax._min = [];
     ax._max = [];
 
-    // Fropagate localization into the axis so that
+    // Propagate localization into the axis so that
```
etpinard: :+1:

src/locale-en.js
```diff
@@ -0,0 +1,37 @@
...
+        shortDays: ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'],
+        months: [
+            'January', 'February', 'March', 'April', 'May', 'June',
+            'July', 'August', 'September', 'October', 'November', 'December'
+        ],
+        shortMonths: [
+            'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
+            'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'
+        ],
+        periods: ['AM', 'PM'],
```
etpinard: I'm curious here. Do locales w/o `periods` set fallback to using a 24-hour clock?

src/plots/plots.js
etpinard: Don't forget to add it #420 :smoking:

src/traces/heatmap/hover.js
```diff
@@ -29,7 +29,6 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode, hoverLay
         zmask = cd0.zmask,
         range = [trace.zmin, trace.zmax],
         zhoverformat = trace.zhoverformat,
-        _separators = trace._separators,
```
etpinard: So we probably can :hocho: these now:

image

src/plot_api/plot_config.js
etpinard: So, `dictionaries` + `formats` would become `locale` as config argument?

If so, I'd vote :+1: for consistency with what `Plotly.register` expects.

etpinard: ... and don't worry about breaking things that haven't been released :wink:

src/plots/plots.js
etpinard: That said, I just remembered that someone (in https://github.com/plotly/plotly.js/issues/1842) has asked for per-axis `separators` settings. I didn't really understand that person's use case though. Personally, I think `separators` are better served as locale/context options, but maybe I'm wrong.

src/locale-en.js
alexcjohnson: `periods` is only relevant if you use `%p` (and `%I` for 12-hour hours) in a custom format, our standard formatting always uses a 24-hour clock. Locales that don't define `periods` will inherit `['AM', 'PM']` from `en`. Seems like most places don't care about this, and some that do are still OK with AM/PM (since it's latin anyway).

src/plots/plots.js
alexcjohnson: > picking `en` (not `en-us`) as our default

OK, so what we're doing right now is a bit schizophrenic, but it has a certain logic to it: `en-US` is the default locale, but if you specify a different locale and it has an incomplete definition (After potentially combining the region and language definitions), it will fill in with non-translated strings (for which we've use US english, `en-US`) but `en` for the missing formatting fields. But the ONLY difference between `en` and `en-US` formatting is if you make a custom `tickformat` with [`%x`](https://github.com/d3/d3-time-format) (and possibly `%c`) in it - you get DD/MM/YYYY format from `en` vs MM/DD/YYYY from `en-US`. Seemed to me that since the US is [about the only country in the world](https://en.wikipedia.org/wiki/Date_format_by_country) that uses the utterly nonsensical MM/DD/YYYY format, if you've already said you're not in the US we shouldn't fall back on that.

For strings, the only place non-translated is different is different is color/colour. I guess I don't care much about that, it currently only shows up in "Click to enter Colo(u)rscale title," and I see no legitimate reason to give an incomplete set of string translations unless you're in another english region (in which case `en` is part of the inheritance chain anyway so you'll get your silly extra "u")

Any quibbles with that logic?

src/plots/plots.js
alexcjohnson: added https://github.com/plotly/plotly.js/issues/420#issuecomment-352099689

> That said, I just remembered that someone (in #1842) has asked for per-axis separators settings.

commented over there https://github.com/plotly/plotly.js/issues/1842#issuecomment-352105362

src/registry.js
alexcjohnson: to be precise, we'll fall back on no translation (which means `en-US`) - more detail https://github.com/plotly/plotly.js/pull/2207#discussion_r157288668

src/plots/plots.js
etpinard: > Any quibbles with that logic?

None from me :ok_hand:

tasks/pull_date_format.js
alexcjohnson: happy to make `locales/` directories if you think it's better. Definitely seems nice for `lib/`, but how would it play with the CDN? [Sync](https://github.com/plotly/streambed/blob/master/bin/update_plotlyjs_cdn.sh) probably has to change, is it OK with serving from subdirectories?

src/core.js
alexcjohnson: Added a check for case correctness of all the requires https://github.com/plotly/plotly.js/pull/2207/commits/43b73e0a6a54bdc0586973696d1f0c9cc2bf0f79

src/plot_api/plot_config.js
alexcjohnson: combined `dictionaries` + `formats` -> `locales` in https://github.com/plotly/plotly.js/pull/2207/commits/27037af204754658a68af4d9d75c1dd530939e79

alexcjohnson:
> > Fixes #1264 - and incidentally
> I didn't see a new test case for this specifically

Good call -> https://github.com/plotly/plotly.js/pull/2207/commits/2a8d38f2f05e6e9f8b472aa17fc215547106d0fc

tasks/pull_date_format.js
alexcjohnson: Neat, I didn't know about `require.resolve` - took a little fiddling but -> https://github.com/plotly/plotly.js/pull/2207/commits/d0e01584136a68d4373b7896ccbe758c9adbb527

src/traces/heatmap/hover.js
alexcjohnson: :hocho: `trace._separators` in https://github.com/plotly/plotly.js/pull/2207/commits/592fc9eacd31f7d36a172edc6694ae570bca38bb

tasks/pull_date_format.js
etpinard: update from private convo:

image

alexcjohnson: lib/locale -> https://github.com/plotly/plotly.js/pull/2207/commits/614d5befb50a6b95774fbc04b5fd58596b5cb66d

src/core.js
alexcjohnson: updated `dist/README.md` -> https://github.com/plotly/plotly.js/pull/2207/commits/331b2daca8bb6398d01d6b0190e9e1bfba9e980c

will open an issue at https://github.com/plotly/documentation once this is merged.

alexcjohnson:
> I'm thinking we should make a pinned post on https://community.plot.ly/ like @chriddyp did to advertise the Dash workshops to attract potential contributors.

That's a great idea - wait until this is published (`1.32`), then start including localizations in `1.33`? I'll try to get @stelajohnson to help me fill in the dictionary for `sq` (Albanian) so potential contributors have a complete example to work off, @etpinard perhaps you want to do (or find someone to do) `fr`? Is `fr-QC` different enough to be noticed in plotly.js?

etpinard:
Amazing PR :dancer:

etpinard:
Don't forget to either close (or update with remaining open items) https://github.com/plotly/plotly.js/issues/856 after merging.

PR #2210, opened 2017-12-16
word breaking in notifier

### LINKED ISSUE #2201: Long words in notifier messages can go off the edge
Thinking about German, now that we have localization abilities (#2195) - we should allow word breaking in `Lib.notifier`, which already has automatic line breaking.

long word:
screen shot 2017-12-12 at 2 16 48 pm 2
vs long message of short words:
screen shot 2017-12-12 at 2 20 14 pm 2
seems like an easy fix https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

alexcjohnson:
Fixes #2201
Break and hyphenate (in some browsers) long words. Not needed for English, but potentially important for eg German (long compound words?) and maybe asian languages with no spaces between word characters.

Chrome - regular words:
chrome words
long words:
chrome long

FF - regular words:
ff words
long words:
ff long

taken from https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/
but modified: using `word-break`, no matter what I did there would be cases that short words got broken unnecessarily, so I removed it. Also `-moz-hyphens` didn't do anything - which was expected as `hyphens` is supposed to be supported in FF, but I don't ever seem to see hyphens there regardless.

### DIFF
--- a/build/plotcss.js
+++ b/build/plotcss.js
@@ -50,7 +50,7 @@ var rules = {
     "X .select-outline-2": "stroke:black;stroke-dasharray:2px 2px;",
     Y: "font-family:'Open Sans';position:fixed;top:50px;right:20px;z-index:10000;font-size:10pt;max-width:180px;",
     "Y p": "margin:0;",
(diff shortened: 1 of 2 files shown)


etpinard:
well played :dancer:

PR #2212, opened 2017-12-18
allow use of element ID in Plots.resize

alexcjohnson:
Also `Plots.redrawText` while we're at it - this is the other one in `Plots` that might occasionally need to be called from outside.

cc @etpinard @VeraZab

### DIFF
--- a/test/jasmine/tests/plots_test.js
+++ b/test/jasmine/tests/plots_test.js
@@ -6,6 +6,7 @@ var d3 = require('d3');
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
 var supplyAllDefaults = require('../assets/supply_defaults');
+var failTest = require('../assets/fail_test');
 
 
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/plots_test.js
```diff
@@ -369,7 +370,7 @@ describe('Test Plots', function() {
                 .then(done);
         });
 
-        afterEach(destroyGraphDiv);
+        afterAll(destroyGraphDiv);
```
alexcjohnson: Oops, if we're doing `beforeAll` we'd better match it with an `afterAll`! The second test (`should resize...`) didn't consider that it might be running against a missing plot (`mainSvgs.length` was `0`) so was in fact a noop. The third test (`should update...`) I guess was OK, `gd` was still an object with properties even if it wasn't attached to the DOM...

etpinard: Nice find :microscope:

etpinard:
:dancer:

Issue #163, opened 2016-01-06
Local jasmine test should use local topojson to plot geo maps


etpinard:
More info: https://github.com/plotly/plotly.js/pull/154#issuecomment-168828320

PR #2213, opened 2017-12-18
Heatmap equal gaps

### LINKED ISSUE #2142: Nonuniform heatmap brick gaps
The first and last gap are different from other gaps (which our test image doesn't catch because it only has 2 gaps in each direction...)
screen shot 2017-11-03 at 2 39 41 pm

alexcjohnson:
Fixes #2142 - heatmaps with more than 3 bricks in one direction had nonuniform gaps. Looking into it in more detail, the bricks at the edges were also not centered on their desired data values, as the gap was applied only to the inside edges of these gaps.

This PR makes all the gaps symmetric (aside from rounding to keep the brick edges as crisp as possible).

Note the two commits here:
- https://github.com/plotly/plotly.js/commit/2802c7989f0b50fa9b847040e55eedade1f3f3c1 just changes the mock & baseline image to make the issue obvious
- https://github.com/plotly/plotly.js/commit/0d1709b2e85856912964bbb730a054d87805e41c changes the algorithm and updates the baseline image and jasmine tests - please see the image diff in that commit to get a feel for what the new algorithm does differently. In particular, the bricks are all now a bit smaller than they were before, since some of the gaps had to expand to be as big as requested, as well as adding gaps around the edges to keep the bricks centered.

cc @CalvinFernandez @etpinard


test/jasmine/bundle_tests/choropleth_test.js
```diff
@@ -5,24 +5,34 @@ var PlotlyChoropleth = require('@lib/choropleth');
...
+        .then(function() {
+            var nodes = d3.selectAll('g.trace.choropleth');
 
-        expect(nodes.size()).toEqual(4);
-    });
+            expect(nodes.size()).toEqual(4);
+        })
+        .catch(failTest)
+        .then(done);
+    }, LONG_TIMEOUT_INTERVAL);
```
alexcjohnson: We've seen this one fail a lot - and this morning seems like circle is super stressed or something, every intermittent test failure we've ever had seems to be biting us today. Choropleth needs to make a request for map data - it's local but still perhaps slow? Just give it some more time before we fail.

etpinard: > it's local but still perhaps slow?

Actually, no. Geo subplots in jasmine tests get their map data from https://cdn.plot.ly/world_110m.json and friends.

You brought this up previously (see https://github.com/plotly/plotly.js/issues/163) but I never got around to fixing that.

----

:ok_hand: for bumping the timeout interval.

etpinard:
Excellent fix :dancer:

PR #2215, opened 2017-12-19
allow hover labels to extend to the edges of the plot

### LINKED ISSUE #527: Small domain causing hovers to not appear
Ok, maybe "sensible" and "nonsensical" are not really true, I'm just thinking domain should be `[0,1]`, but maybe that is not necessary.

This bug was discovered via this SO post. http://stackoverflow.com/questions/37178664/plotly-hover-text-not-displaying/37180688#37180688

[This fiddle](https://jsfiddle.net/abalter/0cjprqgy/) has the following code, and hover text is not working. If I make the single change of `domain` from `[0.85, 0.9]` to `[0.85, 1.9]`, the hovers show. But the whole point is to have the line of markers above other plots as in http://plnkr.co/edit/vvStyutowlwubUdddY37

``` javascript
var data =
{
    "x":["2014-02-10 00:00:00.0","2014-02-18 00:00:00.0","2014-02-24 00:00:00.0"],
  "y":[0,0,0],
  "text":["gemcitabine","gemcitabine + Abraxane","Xeloda"],
    "hoverinfo": "all",
  "name":"Treatment",
  "type":"scatter",
  "mode":"markers",
  "marker":
  {
    "size":9,
    "color":"#800000"
    },
  "uid":"c2e171"
};

var layout = 
{
    "title":"Treatments",
  "height":600,
  "width":655,
  "autosize":true,
  "yaxis":
  {
    "titlefont":
    {
        "size":12,
      "color":"#800000"
    },
    "domain":[0.85,0.9],
    "showgrid":false,
    "showline":false,
    "showticklabels":false,
    "zeroline":true,
    "type":"linear",
    "range":[0,0],
    "autorange":true
  },
  "hovermode":"closest",
  "xaxis":
  {
    "type":"date",
    "range":[1389215256994.8186,1434909143005.1814],
    "autorange":true
  }
};

Plotly.plot('graph', [data], layout);
```

alexcjohnson:
fixes #527 

@etpinard another quick bugfix for you: Allow hover labels to push out not just to the edges of the subplot but to the edges of the graph div.

There's one downside to this fix: The axis label can get partially covered by point labels in certain extreme cases:
screen shot 2017-12-18 at 11 40 47 pm
The common label is supposed to say "day 1" and the "1" is mostly covered. If this label were much longer, more of it would be hidden. To my mind this is a good tradeoff, as a) the y labels are usually more important, so it's better to show as many of these as you can, and b) you can always switch to "closest" mode to narrow in on what info is displayed.

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -452,7 +452,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
     var hoverLabels = createHoverText(hoverData, labelOpts, gd);
 
-    hoverAvoidOverlaps(hoverData, rotateLabels ? 'xa' : 'ya');
+    hoverAvoidOverlaps(hoverData, rotateLabels ? 'xa' : 'ya', fullLayout);
 
     alignHoverText(hoverLabels, rotateLabels);
 
@@ -791,7 +791,7 @@ function createHoverText(hoverData, opts, gd) {
 // know what happens if the group spans all the way from one edge to
 // the other, though it hardly matters - there's just too much
 // information then.
-function hoverAvoidOverlaps(hoverData, ax) {
+function hoverAvoidOverlaps(hoverData, ax, fullLayout) {
     var nummoves = 0,
 
         // make groups of touching points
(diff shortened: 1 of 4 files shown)


etpinard:
> To my mind this is a good tradeoff, as a) the y labels are usually more important, so it's better to show as many of these as you can, and b) you can always switch to "closest" mode to narrow in on what info is displayed.

I agree. This is a good tradeoff.

Nice fix and test :dancer:

PR #2217, opened 2017-12-20
German translation  

JonasGutermuth:
Not finished translation for #856:
```
      //Missing context
      "open:":"Offen:",                                                  // traces/ohlc/transform.js:136
      "high:":"",                                                  // traces/ohlc/transform.js:137
      "low:":"",                                                   // traces/ohlc/transform.js:138
      "kde:":"",                                                   // traces/violin/calc.js:73
      "outgoing flow count:":"",                                   // traces/sankey/plot.js:143
      "incoming flow count:":"",                                   // traces/sankey/plot.js:142

      //Missing translation
      "Toggle Spike Lines":"",                                      // components/modebar/buttons.js:548
      "Toggle show closest data on hover":"",                      // components/modebar/buttons.js:353
      "Turntable rotation":""                    
```


lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'max: ': 'Max.: ',                                                   // traces/box/calc.js:132
+        'mean ± σ: ': 'Mittelwert ± σ: ',                                            // traces/box/calc.js:133
+        'mean: ': 'Mittelwert: ',                                                  // traces/box/calc.js:133
+        'median: ': 'Median: ',                                                // traces/box/calc.js:128
+        'min: ': 'Min.: ',                                                 // traces/box/calc.js:129
+
+        // Missing context
+        'open: ': 'Offen: ',                                                  // traces/ohlc/transform.js:136
+        'high: ': '',                                                  // traces/ohlc/transform.js:137
+        'low: ': '',                                                   // traces/ohlc/transform.js:138
```
alexcjohnson: open, high, low, and close (translated above as schließen) all pertain to the financial chart which in English has the same name (OHLC). These refer to stock prices: the first and last trade (usually within one trading day) and the highest and lowest prices of any trade that day.

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'kde: ': '',                                                   // traces/violin/calc.js:73
```
alexcjohnson: kde is Kernel Density Estimate - the value that's displayed as the thickness of a violin plot.

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'outgoing flow count: ': '',                                   // traces/sankey/plot.js:143
+        'incoming flow count: ': '',                                   // traces/sankey/plot.js:142
```
alexcjohnson: incoming/outgoing flow count are for [sankey (flow) diagrams](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/sankey_energy.png): how many links are there entering this node, and how many leave this node.

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+
+        // Missing translation
+        'Toggle Spike Lines': '',                                      // components/modebar/buttons.js:548
+        'Toggle show closest data on hover': '',                      // components/modebar/buttons.js:353
+        'Turntable rotation': ''                                     // components/modebar/buttons.js:288
```
alexcjohnson: `Toggle Spike Lines`: spike lines are what we call the lines to the axis or across the plot that can show up when you hover on a point - this button turns them on or off. See original PR https://github.com/plotly/plotly.js/pull/1461

`Toggle show closest data on hover`: sometimes there are two buttons available: "show closest data on hover" and "compare data on hover". Other times there is only one button, because only "closest" is meaningful, and this button turns it on or off (no hover effects at all).

`Turntable rotation`: This is a funny one: in 3D plots, you have two different rotation modes: "Orbital rotation" (translated above as "Orbitalrotation") allows you to grab the scene and rotate it in a path-dependent way along any rotation axis (ie at any instant the rotation axis is perpendicular to both the 3-vector you started at, and the 3-vector implied by your mouse motion. "Turntable rotation" maps x motion of the mouse to rotation around the z axis, and y motion of the mouse to tilt of the z axis toward or away from the viewer. This does not allow you to tip the scene sideways but can be more intuitive to use. It is named after the old record player turntables.

alexcjohnson:
@JonasGutermuth thanks for kicking this off! Any other German speakers care to weigh in or help with the remaining items? Anyone used an OHLC or violin plot in German?

Please pay close attention to whitespace and punctuation: No translation keys or values should start or end with spaces, and if a key ends in a colon `:` that means we will display a data value after this, so the translation value should also end with a colon or other appropriate label/data separator.

lib/locales/de.js
JonasGutermuth: Just the number of links or their weight, too

Translation suggestion:
'outgoing flow count:': 'Anzahl ausgehender Verbindungen:',  
'incoming flow count:': 'Anzahl ausgehender Verbindungen:',

lib/locales/de.js
JonasGutermuth: I know what it does, but I'm not aware of the German word. 

The closest translation I've got in mind is "Fadenkreuz" (engl. crosshairs). But I think it is a bad translation. I'm sure there is a better translation.

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
 module.exports = {
     moduleType: 'locale',
     name: 'de',
-    dictionary: {},
+    dictionary: {
+        'Autoscale': 'Automatische Skalierung',                                              // components/modebar/buttons.js:139
+        'Box Select': 'Box Auswahl',                                            // components/modebar/buttons.js:103
```
jdugge: "Rechteckauswahl"

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'Click to enter X axis title': 'Klicken, um den Titel der X-Achse einzugeben',                           // plots/plots.js:435
+        'Click to enter Y axis title': 'Klicken, um den Titel der Y-Achse einzugeben',                           // plots/plots.js:436
+        'Compare data on hover': 'Über die Daten fahren, um sie zu vergleichen',                                 // components/modebar/buttons.js:167
+        'Double-click on legend to isolate one trace': 'Daten isolieren durch Doppelklick in der Legende',           // components/legend/handle_click.js:90
+        'Double-click to zoom back out': 'Herauszoomen durch Doppelklick',                         // plots/cartesian/dragbox.js:299
+        'Download plot as a png': 'Download als png',                                // components/modebar/buttons.js:52
+        'Edit in Chart Studio': 'Im Chart Studio bearbeiten',                                  // components/modebar/buttons.js:76
+        'IE only supports svg.  Changing format to svg.': 'IE unterstützt nur SVG-Dateien.  Format wird zu SVG gewechselt.',        // components/modebar/buttons.js:60
```
jdugge: 'IE unterstützt nur svg-Dateien.  Format wird zu svg gewechselt.'

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'Pan': 'Verschieben',                                                   // components/modebar/buttons.js:94
+        'Produced with Plotly': 'Erstellt mit Plotly',                                  // components/modebar/modebar.js:256
+        'Reset': 'Zurücksetzen',                                                 // components/modebar/buttons.js:432
+        'Reset axes': 'Achsen zurücksetzen',                                            // components/modebar/buttons.js:148
+        'Reset camera to default': 'Kamera auf Standard zurücksetzen',                               // components/modebar/buttons.js:314
+        'Reset camera to last save': 'Kamera auf letzte Speicherung zurücksetzen',                             // components/modebar/buttons.js:322
+        'Reset view': 'Ansicht zurücksetzen',                                           // components/modebar/buttons.js:583
+        'Reset views': 'Ansichten zurücksetzen',                                           // components/modebar/buttons.js:529
+        'Show closest data on hover': 'Zeige nähste Daten beim Überfahren',                            // components/modebar/buttons.js:157
```
jdugge: 'Zeige näheste Daten beim Überfahren'

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'Snapshot succeeded': 'Snapshot erfolgreich',                                    // components/modebar/buttons.js:66
+        'Sorry, there was a problem downloading your snapshot!': 'Es gab ein Problem beim Herunterladen des Snapshots', // components/modebar/buttons.js:69
+        'Taking snapshot - this may take a few seconds': 'Erstelle einen Snapshot - dies kann einige Sekunden dauern',         // components/modebar/buttons.js:57
+        'Zoom': 'Zoom',                                                  // components/modebar/buttons.js:85
+        'Zoom in': 'Hineinzoomen',                                               // components/modebar/buttons.js:121
+        'Zoom out': 'Herauszoomen',                                              // components/modebar/buttons.js:130
+        'close: ': 'schließen',                                                // traces/ohlc/transform.js:139
```
jdugge: @alexcjohnson The English key (`'close: '`) has a space after the colon. Is this correct? In https://github.com/plotly/plotly.js/pull/2217#issuecomment-353177120, you said that "No translation keys or values should start or end with spaces"

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'trace': 'Datenspur',                                                 // plots/plots.js:439
+        'lat: ': 'Lat.: ',                                                   // traces/scattergeo/calc.js:48
+        'lon: ': 'Lon.: ',                                                  // traces/scattergeo/calc.js:49
+        'q1: ': 'q1',                                                    // traces/box/calc.js:130
+        'q3: ': 'q3',                                                    // traces/box/calc.js:131
+        'source: ': 'Quelle: ',                                                // traces/sankey/plot.js:140
+        'target: ': 'Ziel: ',                                          // traces/sankey/plot.js:141
+        'lower fence: ': 'Untere Schranke',                                           // traces/box/calc.js:134
```
jdugge: 'Untere Schranke: '

lib/locales/de.js
jdugge: 'Schluss: ' (https://books.google.com.au/books?id=UjgPCgAAQBAJ&lpg=PA93&dq=balkenchart&pg=PA93#v=onepage&q&f=false)

lib/locales/de.js
jdugge: 'Eröffnung: '
'Höchstkurs: '
'Tiefstkurs: '

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        // Missing context
+        'open: ': 'Offen: ',                                                  // traces/ohlc/transform.js:136
+        'high: ': '',                                                  // traces/ohlc/transform.js:137
+        'low: ': '',                                                   // traces/ohlc/transform.js:138
+        'kde: ': '',                                                   // traces/violin/calc.js:73
+        'outgoing flow count: ': '',                                   // traces/sankey/plot.js:143
+        'incoming flow count: ': '',                                   // traces/sankey/plot.js:142
+
+        // Missing translation
+        'Toggle Spike Lines': '',                                      // components/modebar/buttons.js:548
```
jdugge: 'Bezugslinien an-/abschalten' (this is what Excel calls spike lines)

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
...
+        'Toggle show closest data on hover': '',                      // components/modebar/buttons.js:353
```
jdugge: 'Anzeige der nähesten Daten an-/abschalten'

lib/locales/de.js
jdugge: Blender calls turntable rotation "Drehscheibenorbit"... not sure if this is a commonly understood term, though

lib/locales/de.js
jdugge: 'Dichte: ' (or more accurately 'Dichteschätzung: ', or even more accurately 'Kerndichteschätzung: '). Unfortunately, there doesn't seem to be a common abbreviation for this term in German.

lib/locales/de.js
alexcjohnson: Right, we should remove all of these trailing spaces. Since this is the first full translation @JonasGutermuth was in uncharted territory, and `translation-keys.txt` doesn't make this clear due to the path/line comments. My apologies for not making this more clear earlier!

lib/locales/de.js
alexcjohnson: I don't think orbital / turntable rotation is particularly commonly understood in English either ;) As long as it's clear to users that these are two different rotation modes, I'm happy.

lib/locales/de.js
alexcjohnson: Just the number of links. Normally incoming and outgoing have the same total weight, which is displayed as the node size and in its label.

lib/locales/de.js
```diff
@@ -11,7 +11,68 @@
 module.exports = {
     moduleType: 'locale',
     name: 'de',
-    dictionary: {},
+    dictionary: {
+        'Autoscale': 'Automatische Skalierung',                                              // components/modebar/buttons.js:139
+        'Box Select': 'Box Auswahl',                                            // components/modebar/buttons.js:103
+        'Click to enter Colorscale title': 'Klicken, um den Farbskalatitel einzugeben',                       // plots/plots.js:437
```
jnehlmeier: I wonder if 'click to' and thus 'Klicken, um' is really needed. 

In HTML you should never use a hyperlink named "here" like "[click here](www.google.com) to open Google". It is always preferable to use "[Open Google](www.google.com)". I feel like help texts in tooltips, etc should act the same.

Same for related translations. What do you think?

alexcjohnson: Our rationale was that it’s not obvious that this is a clickable link at all, unlike a specially-styled item within text - and we purposely make this placeholder text lighter than other text, not more prominent like links usually are, so it doesn’t distract you too much if you don’t want a title at all.

lib/locales/de.js
JonasGutermuth: this is of course better

lib/locales/de.js
JonasGutermuth: Wikipedia uses SVG instead of svg. The double space is also in english.

lib/locales/de.js
```diff
@@ -12,62 +12,62 @@ module.exports = {
...
+        'Toggle Spike Lines': 'Bezugslinien an-/abschalten',                                                // components/modebar/buttons.js:548
+        'open:': 'Eröffnung:',                                                                              // traces/ohlc/transform.js:136
+        'high:': 'Höchstkurs:',                                                                             // traces/ohlc/transform.js:137
+        'low:': 'Tiefstkurs:',                                                                              // traces/ohlc/transform.js:138
+        'Toggle show closest data on hover': 'Anzeige der nähesten Daten an-/abschalten',                   // components/modebar/buttons.js:353
+        'incoming flow count:': 'Anzahl eingehender Verbindungen',                                          // traces/sankey/plot.js:142
+        'outgoing flow count:': 'Anzahl ausgehender Verbindungen',                                          // traces/sankey/plot.js:143
+        'kde:': 'Kerndichteschätzung'                                                                       // traces/violin/calc.js:73
```
JonasGutermuth: @jdugge:
'Dichte: ' (or more accurately 'Dichteschätzung: ', or even more accurately 'Kerndichteschätzung: '). Unfortunately, there doesn't seem to be a common abbreviation for this term in German.

lib/locales/de.js
```diff
@@ -12,62 +12,62 @@ module.exports = {
...
+        'Click to enter X axis title': 'Klicken, um den Titel der X-Achse einzugeben',                      // plots/plots.js:435
+        'Click to enter Y axis title': 'Klicken, um den Titel der Y-Achse einzugeben',                      // plots/plots.js:436
+        'Compare data on hover': 'Über die Daten fahren, um sie zu vergleichen',                            // components/modebar/buttons.js:167
+        'Double-click on legend to isolate one trace': 'Daten isolieren durch Doppelklick in der Legende',  // components/legend/handle_click.js:90
+        'Double-click to zoom back out': 'Herauszoomen durch Doppelklick',                                  // plots/cartesian/dragbox.js:299
+        'Download plot as a png': 'Download als png',                                                       // components/modebar/buttons.js:52
+        'Edit in Chart Studio': 'Im Chart Studio bearbeiten',                                               // components/modebar/buttons.js:76
+        'IE only supports svg.  Changing format to svg.': 'IE unterstützt nur SVG-Dateien.  Format wird zu SVG gewechselt.', // components/modebar/buttons.js:60
```
JonasGutermuth: @jdugge:
'IE unterstützt nur svg-Dateien. Format wird zu svg gewechselt.'
@JonasGutermuth:
Wikipedia uses SVG instead of svg. The double space is also in english.

alexcjohnson: Thanks for pointing out the double space - Since this gets displayed with normal whitespace rules, only one space is actually shown, but it needs the spaces to match for translation. I'll leave it for now, but may strip out the second space (from the source and any translations) at some point.

re: capitalization of SVG - I don't have a strong opinion for English (anyone else feel strongly that we should capitalize it?), but we should probably keep SVG and PNG consistent within any given locale, so should we also use `'Download als PNG'` above?

lib/locales/de.js
alexcjohnson: As this is used as a hover label prefix, I'd tend to place a premium on keeping it short. If you think a violin plot user (someone who already has a basic understanding of what information a violin plot conveys) who sees a point on the curve labeled with `'Dichte:'` would understand it to mean the kernel density estimate, then I'd vote for that instead.

alexcjohnson:
@JonasGutermuth looking great! Just two final comments but I'll leave them to your judgment. Let me know if you'd like to make further changes or if you're happy as it is, and I'll merge.

alexcjohnson:
Also happy new year! We'll need to merge master to get tests to pass, with the new year in file headers.

lib/locales/de.js
JonasGutermuth: You are right. Wikipedia uses also PNG instead of png.

lib/locales/de.js
alexcjohnson: > I wonder if 'click to' and thus 'Klicken, um' is really needed.

FYI, I thought it best to leave this in here, as @JonasGutermuth did an admirable job keeping the translation short. But @jnehlmeier it's true that this may not be completely necessary. We've been discussing this in other contexts, and if the translation is much longer than the English and users can be expected to understand that this is a clickable link without the "Click to", you're right it may be better to omit it so it doesn't take up too much space - see https://github.com/plotly/plotly.js/pull/2252#discussion_r161635675

jnehlmeier: Good design is to never tell a user that something is clickable, it must visually be clear that it is clickable. Desaturated, light text might look good, but gives poor visual clues. 

So maybe you should think about restyling the elements in question. E.g. make everything clickable the same color, or add a consistent edit icon, change the mouse cursor on hover, or even remove clickable placeholder text and instead use input boxes in some edit panel.

alexcjohnson: (Not to beat a dead horse, and I don't mean to say we would never change this design, @jnehlmeier raises good points, just hoping this conversation may be useful to future translation contributors)

> Good design is to never tell a user that something is clickable

Definitely a good rule of thumb, though a similar thing happens in Powerpoint and Keynote:
screen shot 2018-01-16 at 6 55 54 am

In our [Chart editor](plot.ly/create) (which is the original context for which `editable: true` was created, though of course we hope it finds other uses) we *do* have input boxes in an edit panel as an option, and these placeholders *do* get an edit cursor on hover (which titles do not get if they are not editable). Our design goal was to make it clear that you *can* just click and type, but in a soft enough way that users won't feel the need to delete placeholders for titles they don't want, and they will understand that, just like in PPT/Keynote, when you're done editing and view your plot in a non-editable context these placeholders disappear if not replaced by other text.

PR #2218, opened 2017-12-20
scatter & bar hover fixes for #780

### LINKED ISSUE #780: Bar trace hover text not shown when using bar and scatter with large data set
[Example here](https://jsfiddle.net/0sgj3dzb/1/). With the 'compare data' option selected, only the hover text for the scatter trace is shown. Either reducing the number of data points or removing one of the traces appears to eliminate the issue.

alexcjohnson:
Fixes #780 - @etpinard another quick bugfix for you to look at.

Scatter hover was too permissive and too flat in its distance function in compare mode - so it would call the wrong point "closest" sometimes. That was the main culprit in #780, we were matching a scatter point with one x value to a bar with a different x and concluding that they really shouldn't "compare" with each other, so tossing out the bar.

Bar hover was also too restrictive, it would not allow you to be hovered on a bar if the mouse was over the gap between bars. Now in compare mode it allows that, which makes for a more pleasant (and useful) effect as you mouse over, particularly when the bars are very narrow: the hover labels don't flash on/off anymore, there's always one showing.

For both trace types I left "closest" hover behavior as it was. In particular this means the "flashing" effect for bars is still there in this mode. Perhaps there would be a way to fix it at least for very thin bars, by rounding outward or something in our acceptance criteria, but at least this doesn't seem to lead to any explicit bugs.

### DIFF
--- a/src/traces/scatter/hover.js
+++ b/src/traces/scatter/hover.js
@@ -18,35 +18,44 @@ var fillHoverText = require('./fill_hover_text');
 var MAXDIST = Fx.constants.MAXDIST;
 
 module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
-    var cd = pointData.cd,
-        trace = cd[0].trace,
-        xa = pointData.xa,
-        ya = pointData.ya,
-        xpx = xa.c2p(xval),
-        ypx = ya.c2p(yval),
-        pt = [xpx, ypx],
-        hoveron = trace.hoveron || '';
+    var cd = pointData.cd;
+    var trace = cd[0].trace;
+    var xa = pointData.xa;
+    var ya = pointData.ya;
+    var xpx = xa.c2p(xval);
+    var ypx = ya.c2p(yval);
+    var pt = [xpx, ypx];
+    var hoveron = trace.hoveron || '';
+    var minRad = (trace.mode.indexOf('markers') !== -1) ? 3 : 0.5;
 
     // look for points to hover on first, then take fills only if we
     // didn't find a point
     if(hoveron.indexOf('points') !== -1) {
         var dx = function(di) {
+                // dx and dy are used in compare modes - here we want to always
+                // prioritize the closest data point, at least as long as markers are
+                // the same size or nonexistent, but still try to prioritize small markers too.
+                var rad = Math.max(3, di.mrc || 0);
+                var kink = 1 - 1 / rad;
+                var dxRaw = Math.abs(xa.c2p(di.x) - xpx);
+                var d = (dxRaw < rad) ? (kink * dxRaw / rad) : (dxRaw - rad + kink);
+                return d;
+            },
+            dy = function(di) {
+                var rad = Math.max(3, di.mrc || 0);
+                var kink = 1 - 1 / rad;
+                var dyRaw = Math.abs(ya.c2p(di.y) - ypx);
+                return (dyRaw < rad) ? (kink * dyRaw / rad) : (dyRaw - rad + kink);
+            },
+            dxy = function(di) {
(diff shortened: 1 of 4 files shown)


src/traces/scatter/hover.js
```diff
@@ -18,35 +18,44 @@ var fillHoverText = require('./fill_hover_text');
...
         var dx = function(di) {
-                // scatter points: d.mrc is the calculated marker radius
-                // adjust the distance so if you're inside the marker it
-                // always will show up regardless of point size, but
-                // prioritize smaller points
+                // dx and dy are used in compare modes - here we want to always
+                // prioritize the closest data point, at least as long as markers are
+                // the same size or nonexistent, but still try to prioritize small markers too.
                 var rad = Math.max(3, di.mrc || 0);
-                return Math.max(Math.abs(xa.c2p(di.x) - xpx) - rad, 1 - 3 / rad);
```
etpinard: I think I copied this piece of logic elsewhere e.g. in [`scattergeo/hover.js`](https://github.com/plotly/plotly.js/blob/82f4c161cfa2035a4de18c0e2071ab7d6ab6f7f1/src/traces/scattergeo/hover.js#L45). Should we update it there too?

test/jasmine/tests/hover_label_test.js
```diff
@@ -1108,6 +1108,43 @@ describe('hover info on stacked subplots', function() {
 });
 
 
+describe('hover on many lines+bars', function() {
+    'use strict';
+
+    afterEach(destroyGraphDiv);
+
+    it('shows hover info for both traces', function(done) {
```
etpinard: Nice test :lock:

test/jasmine/tests/hover_label_test.js
```diff
@@ -1274,7 +1311,7 @@ describe('hover updates', function() {
 
     afterEach(destroyGraphDiv);
 
-    function assertLabelsCorrect(mousePos, labelPos, labelText) {
+    function assertLabelsCorrect(mousePos, labelPos, labelText, msg) {
```
etpinard: Great. So none of the current tests needed to be updated, is that correct?

test/jasmine/tests/hover_label_test.js
```diff
@@ -1383,14 +1420,14 @@ describe('hover updates', function() {
...
-            return assertLabelsCorrect([351, 251], [358, 272], '2');
+            return assertLabelsCorrect([351, 251], [358, 272], '2', 'events 0');
         })
         .then(unhover)
         .then(function() {
             expect(hoverCnt).toEqual(1);
             expect(unHoverCnt).toEqual(1);
 
-            return assertLabelsCorrect([400, 200], [435, 198], '3');
+            return assertLabelsCorrect([420, 100], [435, 198], '3', 'events 1');
```
etpinard: Ha. Found one. Why did this change?

src/traces/scatter/hover.js
alexcjohnson: scattergeo should be fine, as it only uses the `closest` form (ie real quadrature distance, not just x or just y). I suppose the `minRad` change is still relevant there, to better handle line-only traces, but that's it.

test/jasmine/tests/hover_label_test.js
alexcjohnson: `x=400` is a little too far away using the new `compare` logic. I don't know exactly where the center of the marker is (that's why I've moved to more explicitly-sized-and-ranged plots for tests like this, so we can easily reason about the pixel positions of different points), probably the test would have worked just increasing to `405` or something but I just moved the mouse to a point I *knew* would be in range.

etpinard:
Looks great. Thanks for the clarifications. :dancer:

Issue #1396, opened 2017-02-20
Colorbar tick labels overlap with colorbar title


jdugge:
When using a `colorbar` with `titleside: 'right'` and a small value for `len`, the colorbar tick labels can overlap with the colorbar title:

![image](https://cloud.githubusercontent.com/assets/2399255/23126883/18ab6be0-f779-11e6-8f8e-c5b3060c131a.png)

(Codepen: http://codepen.io/anon/pen/YNoOML)

```
var data = [ {
  z: [[0, 100],
       [75, 0]],
  type: 'heatmap',
  colorbar:{
    len: 0.35,
    title: 'My long label',
    titleside:'right',
    tickvals:[0,50,100],
    tickfont: {color: 'red'}
  }
}]; 

var layout = {};

Plotly.newPlot('myDiv', data, layout);
```

The workaround for avoiding this overlap given in https://github.com/plotly/plotly.js/issues/938#issuecomment-247093233 (using an invisible `tickprefix`) only works for titles that are to the left of the tick labels, unfortunately.

Is there a way to avoid this overlap with titles to the right of the tick labels?

hiteshsahu:
**You need to add <br\> with the title text** 

Add `<br\>` as prefix if you want to add space on top of the title. Add `<br\>` with a `&nbsp;` if you want to add space after the title.
  ```
 colorbar:{
           title: 'Speed(RPM)<br\> &nbsp;',
           titleside:'top',
           titlefont: {color: 'blue'}
         },
```

You can use <br\> trick on other elements of Plotly as well. Hope it help somebody.

alexcjohnson:
closed by #2227

PR #2227, opened 2017-12-29
Clean subplots

### LINKED ISSUE #2125: Automatically remove unused subplots
This is largely an issue with the plotly workspace, but I imagine other applications can run into it as well:

If you change trace types between different subplots, for example from heatmap to surface, bar to pie, it's possible to end up with an empty subplot (cartesian in those cases, but could be others) that nevertheless exists in `layout`, with automatically-entered attributes like `range` and `autorange` but possibly with explicitly-provided attributes as well. At present we still display these subplots, even though they're useless.

The proposal (discussed offline with @etpinard and @bpostlethwaite) is:
- Only show subplots that are referenced by some object - be it a trace, shape, annotation, or image.
- Even hidden objects (`visibility: (false|'legendonly')`) count as a subplot reference. So a subplot that's blank just because you turned off all the traces on it from the legend, that will still render.
- If there are *no* subplots left after we axe all of these, make a blank cartesian subplot so it's not just blank.
- Do not delete the unused axis and subplot objects from `layout`, in case the user changes back to the original type, so they don't lose the configuration they did there.

In principle this could be regarded as a breaking change, but in practice it seems like the only thing it breaks is useless plots. If anyone has examples otherwise though I'd love to hear about it!

alexcjohnson:
Closes #2125

Updates subplot creation logic. The meat of this is in https://github.com/plotly/plotly.js/commit/3625be4e516376ff4ce65f0c0223d1d9b99c2d24, which makes it so new subplots are created only by traces (visible OR invisible) and components (annotations, shapes, images), but NOT by axis, scene, or subplot objects.

The primary motivation for this is if you make a trace of one type and change it to another, you won't get subplots corresponding to the old type even if we automatically pushed information for the old subplot back into the layout (for example axis ranges). We can leave those objects in the layout without them being rendered, and yet if you come back to the old trace types later any configuration you had used previously will still be available - a nice side effect in editor contexts.

I added a few more `_private` attributes in the course of this work - which makes pieces a bit more coupled, not so great for testing... but my reasoning is that the logic is fairly complex (we don't want to be redoing it all the time) and nonlocal (the axes/subplots need to know about what's in data and component arrays) so it seemed cleaner to just do it right once during `supplyDefaults` and reference that forever after.

In order to get this to work right, I altered the svg structure a bit, and simplified some of the rendering logic such as `Axes.doTicks` a little. This had some happy side effects like removing excess subplot containers and redundant gridlines when you have multiple cartesian subplots (leads to smaller svg files and presumably a bit faster tick/grid rendering - this is why a few mocks in 2de217e got thinner gridlines, they're no longer being overdrawn), and along the way I fixed a couple of other random subtle bugs:
- colorbar title-avoids-tick-labels algorithm was incorrect when the top of the colorbar wasn't at the top of the plot area (where it is by default). This is why the `airfoil` mock changed in 2de217e, that title was avoiding the wrong label. The fix for this one is wrapped up with other logic changes in the big commit https://github.com/plotly/plotly.js/commit/3625be4e516376ff4ce65f0c0223d1d9b99c2d24 - the code in `colorbars/draw` changed, and now it's fixed, but it's only right in conjunction with other changes in `cartesian/axes` so I couldn't split it out into a separate commit.
- `scene.cameraposition` -> `scene.camera` conversion had a typo, fixed in https://github.com/plotly/plotly.js/commit/26aaffc6153e50f4b68a7f12b4831b2b8f7a5546. It turns out a couple of mocks do use `cameraposition` and changed as a result of this -> https://github.com/plotly/plotly.js/commit/276d1d3c2d652bf7d5a42b8635a5533d5b3052fa

@etpinard please review. Apologies for the giant monolithic commit, fraid I couldn't see a reasonable way to split it up. That probably speaks to the need to refactor this code more than I had the stamina for in this PR...

cc @bpostlethwaite @nicolaskruchten


dist/translation-keys.txt
```diff
@@ -1,12 +1,12 @@
 Autoscale                                              // components/modebar/buttons.js:139
 Box Select                                             // components/modebar/buttons.js:103
-Click to enter Colorscale title                        // plots/plots.js:437
+Click to enter Colorscale title                        // plots/plots.js:338
```
alexcjohnson: hmm, I should have just ignored the changes to this file, since they're only line number changes... I should revisit this so it doesn't get regenerated just because you run `test-syntax` locally.

src/plot_api/subroutines.js
```diff
@@ -349,60 +344,52 @@ exports.lsInner = function(gd) {
...
+
+    // is this a free axis? free axes can only have a subplot side-line with all(ticks)? mirroring
+    if(!anchorAx) return false;
+
+    // in order to handle cases where the user forgot to anchor this axis correctly
+    // (because its default anchor has the same domain on the relevant end)
+    // check whether the relevant position is the same.
+    var sideIndex = alignmentConstants.FROM_BL[side];
+    if(counterAx.side === side) {
+        return anchorAx.domain[sideIndex] === ax.domain[sideIndex];
```
alexcjohnson: This happens fairly often in our mocks, and is super easy for users to do: for example make stacked subplots `xy` and `x2y2`, then leave `y.anchor` out because everything looks right even though it defaults to `x` and not `x2`. Perhaps we should make better default logic (pick the first non-overlaying axis that will actually make a subplot with this axis, or perhaps the first axis that specifies this axis as its own anchor?) but that logic is kinda tricky and nearly circular. Anyway what this does is as long as the subplot you're drawing *looks* like the two axes are anchored together, it will join the lines neatly at the corner.

This wasn't needed before because we generated so many extra overlaying subplots, *one* of them was sure to make a crisp corner...

src/plots/cartesian/axes.js
```diff
@@ -1847,7 +1774,7 @@ axes.doTicks = function(gd, axid, skipTitle) {
     if(axLetter === 'x') {
         sides = ['bottom', 'top'];
         transfn = function(d) {
-            return 'translate(' + ax.l2p(d.x) + ',0)';
+            return 'translate(' + (ax._offset + ax.l2p(d.x)) + ',0)';
```
alexcjohnson: At some point I think we'd be better off just merging `ax._offset` into `ax.*2p` and getting rid of all the `<g>` translations we have scattered about. The fact that it's not in there is a remnant of the days when plots were nested `<svg>` elements so within a subplot pixel coordinates always started at (0,0). Here at least I'm getting rid of the `<g>` translations associated with ticks and grids, but it means I have to put them back into the tick and grid *paths* until `*2p` catches up.

src/plots/cartesian/axes.js
```diff
@@ -2326,54 +2263,59 @@ axes.doTicks = function(gd, axid, skipTitle) {
...
-                        ticksides[i] = true;
-                    }
-                }
+        if(ax.ticks) {
+            var mainSign = ticksign[2];
+            var tickpath = tickpathfn(ax._mainLinePosition + pad * mainSign, mainSign * ax.ticklen);
+            if(ax._anchorAxis && ax.mirror && ax.mirror !== true) {
+                tickpath += tickpathfn(ax._mainMirrorPosition - pad * mainSign, -mainSign * ax.ticklen);
             }
+            drawTicks(mainPlotinfo[axLetter + 'axislayer'], tickpath);
```
alexcjohnson: instead of trying to draw all the ticks in a single loop, now I'm splitting it up into the main subplot (the main subplot is the one you're anchored to and may get ticks on one or both sides) and all other subplots (which, if they get any ticks, always get both sides).

This also simplifies figuring out where to put labels, since they always go only on at `ax._mainLinePosition` and we then no longer need to pay attention *here* to if that's an anchored or free position or which side of the subplot to pick out.

src/plots/cartesian/axes.js
```diff
@@ -2326,54 +2263,59 @@ axes.doTicks = function(gd, axid, skipTitle) {
...
         return drawLabels(ax._axislayer, ax._pos);
     }
-    else {
+    else if(fullLayout._has('cartesian')) {
         subplots = axes.getSubplots(gd, ax);
-        var alldone = subplots.map(function(subplot) {
+
+        // keep track of which subplots (by main conteraxis) we've already
+        // drawn grids for, so we don't overdraw overlaying subplots
+        var finishedGrids = {};
```
alexcjohnson: note that you can't do something simple like pick the main axis for each overlaying subplot, as that subplot may not exist (say for example you have `x2y2` overlaying `xy` - for `y2` the main counteraxis is `x` but there is no `xy2` subplot. So instead I just make sure we only draw the grid once for any given main axis.

src/plots/cartesian/axis_ids.js
```diff
@@ -65,7 +64,7 @@ function listNames(gd, axLetter, only2d) {
     var names = filterAxis(fullLayout, '');
     if(only2d) return names;
 
-    var sceneIds3D = Plots.getSubplotIds(fullLayout, 'gl3d') || [];
```
alexcjohnson: removed one circular dep here (only 16 to go 🙄 ) by having `_subplots` already calculated. I removed `Plots.getSubplotIds` entirely since it had become a 1-liner just looking up entries in `fullLayout._subplots`.

src/plots/plots.js
```diff
@@ -894,18 +893,27 @@ plots.supplyDataDefaults = function(dataIn, dataOut, layout, fullLayout) {
...
         cnt++;
+
+        // TODO: do we really want color not to increment for explicitly invisible traces?
+        // This logic is weird, but matches previous behavior: traces that you explicitly
+        // set to visible:false do not increment the color, but traces WE determine to be
+        // empty or invalid (and thus set to visible:false) DO increment color.
+        // I kind of think we should just let all traces increment color, visible or not.
+        // see mock: axes-autotype-empty vs. a test of restyling visible: false that
+        // I can't find right now...
+        if(fullTrace._input.visible !== false) colorCnt++;
```
alexcjohnson: @etpinard thoughts? Should `visible: false` traces still be allowed to increment color? I guess one reason to permit disqualified traces to increment color is in case they become valid later on (like by streaming the first data point into one) but I feel like that argument applies to explicitly `visible: false` traces too, ie showing one would not be expected to change all the other colors. But I guess technically that would be a breaking change so maybe should wait for v2? As I have it here we preserve backward compatibility.

test/jasmine/tests/toimage_test.js
```diff
@@ -149,7 +149,7 @@ describe('Plotly.toImage', function() {
         .then(function() { return Plotly.toImage(gd, {format: 'png', imageDataOnly: true}); })
         .then(function(d) {
             expect(d.indexOf('data:image/')).toBe(-1);
-            expect(d.length).toBeWithin(53660, 1e3, 'png image length');
+            expect(d.length).toBeWithin(54660, 3e3, 'png image length');
```
alexcjohnson: This had been failing on my computer for ages, though it always works on CI... now there's just one test, in gl3d code, that consistently fails for me, though I need to be careful to have the chrome window on my secondary monitor; if I let it live on my laptop screen (retina) a whole bunch of other tests fail too. Must be a `devicePixelRatio` dependence or something. Would love to sort this out at some point, makes it annoying to run the full test suite if I'm not at home, and god forbid I ever upgrade this monitor and only have high-res devices to work with!

src/plots/plots.js
etpinard: You make a great point about streaming data. So yeah, I agree: `visible: false` traces should be allowed to increment color in v2. Let's add that to #420 :smoking:

src/plots/plots.js
```diff
@@ -1051,49 +1059,63 @@ plots.supplyFrameDefaults = function(frameIn) {
...
+            }
+        }
     }
 
     if(visible) {
         coerce('customdata');
         coerce('ids');
 
-        var _module = plots.getModule(traceOut);
-        traceOut._module = _module;
```
etpinard: This will conflict with https://github.com/plotly/plotly.js/pull/1869/commits/a02ed20026a80c530f3f40d66dd079ec024bba96

src/plots/plots.js
```diff
@@ -1352,21 +1368,37 @@ function calculateReservedMargins(margins) {
...
-    Plotly.Axes.supplyLayoutDefaults(layoutIn, layoutOut, fullData);
+    var Cartesian = Registry.subplotsRegistry.cartesian;
+
+    // check if any components need to add more base plot modules
+    // that weren't captured by traces
+    for(i = 0; i < components.length; i++) {
+        _module = Registry.componentsRegistry[components[i]];
+
+        if(_module.includeBasePlot) {
+            _module.includeBasePlot(layoutIn, layoutOut);
```
etpinard: Nicely done here.

src/plots/plots.js
```diff
@@ -1352,21 +1368,37 @@ function calculateReservedMargins(margins) {
...
+        }
+    }
+
+    // make sure we *at least* have some cartesian axes
+    if(!basePlotModules.length) {
+        basePlotModules.push(Cartesian);
```
etpinard: In v2, I would like to remove `scatter` (and thus the cartesian base module) from the core bundle so that e.g. gl3d-only custom bundles don't have to require `Cartesian` for nothing. 

In this case, should empty graphs (i.e. with no `basePlotModules`) remain blank (i.e. with no cartesian axes)?

src/plots/ternary/ternary.js
```diff
@@ -156,6 +156,9 @@ proto.updateLayers = function(ternaryLayout) {
             } else if(d === 'grids') {
                 grids.forEach(function(d) {
                     layers[d] = s.append('g').classed('grid ' + d, true);
+
+                    var fictID = (d === 'bgrid') ? 'x' : 'y';
+                    layers[d].append('g').classed(fictID, true);
```
etpinard: why do we need this now?

src/snapshot/cloneplot.js
```diff
@@ -101,8 +100,11 @@ module.exports = function clonePlot(graphObj, options) {
         }
     }
 
-    var sceneIds = Plots.getSubplotIds(newLayout, 'gl3d');
-
+    // TODO: does this scene modification really belong here?
+    // If we still need it, can it move into the gl3d module?
+    var sceneIds = Object.keys(newLayout).filter(function(key) {
+        return key.match(/^scene\d*$/);
+    });
```
etpinard: `snapshot/cloneplot` will soon (I hope) be obsolete. `Plotly.toImage` does not use it anymore, only the old image server still does. All thumbnail-specific layout patches are now in `image-exporter`.

test/jasmine/tests/annotations_test.js
```diff
@@ -721,8 +721,8 @@ describe('annotations autorange', function() {
                 text: 'LT',
                 x: -1,
                 y: 3,
-                xref: 'x5', // will be converted to 'x' and xaxis should autorange
-                yref: 'y5', // same 'y' -> yaxis
+                xref: 'xq', // will be converted to 'x' and xaxis should autorange
+                yref: 'yz', // same 'y' -> yaxis
```
etpinard: Why did you change this?

test/jasmine/tests/gl_plot_interact_test.js
```diff
@@ -309,8 +309,8 @@ describe('Test gl3d plots', function() {
         .then(function() {
             expect(countCanvases()).toEqual(1);
             expect(gd.layout.scene).toEqual(sceneLayout);
-            expect(gd.layout.xaxis).toBeUndefined();
-            expect(gd.layout.yaxis).toBeUndefined();
+            expect(gd.layout.xaxis === undefined).toBe(true);
+            expect(gd.layout.yaxis === undefined).toBe(true);
```
etpinard: Something is off with `toBeUndefined`?

test/jasmine/tests/plots_test.js
```diff
@@ -758,4 +671,134 @@ describe('Test Plots', function() {
             });
         });
     });
+
+    describe('subplot cleaning logic', function() {
```
etpinard: Lovely tests

test/jasmine/tests/transform_groupby_test.js
```diff
@@ -9,6 +9,15 @@ var customAssertions = require('../assets/custom_assertions');
 var assertDims = customAssertions.assertDims;
 var assertStyle = customAssertions.assertStyle;
 
+
+function supplyDataDefaults(dataIn, dataOut) {
+    return Plots.supplyDataDefaults(dataIn, dataOut, {}, {
```
etpinard: It's getting more difficult to call anything but `Plots.supplyDefaults` in tests due to the large amount of things to mocks. We should maybe start standardising around [`assets/supply_defaults.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/supply_defaults.js) instead?

src/components/colorbar/draw.js
```diff
@@ -433,7 +436,7 @@ module.exports = function draw(gd, id) {
                                 selection: d3.select(gd).selectAll('g.' + cbAxisOut._id + 'tick'),
                                 side: opts.titleside,
                                 offsetLeft: gs.l,
-                                offsetTop: gs.t,
+                                offsetTop: 0,
```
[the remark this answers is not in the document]
etpinard: or maybe https://github.com/plotly/plotly.js/issues/1396

src/plots/plots.js
```diff
@@ -1352,21 +1368,37 @@ function calculateReservedMargins(margins) {
...
+    }
+
+    // ensure all cartesian axes have at least one subplot
+    if(layoutOut._has('cartesian') || layoutOut._has('gl2d')) {
```
etpinard: Why do we need the `|| layoutOut._has('gl2d')` part here?

src/components/colorbar/draw.js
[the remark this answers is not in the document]
alexcjohnson: I don't think it fixes #970 but yes, it does fix #1396 🎉

src/plots/plots.js
```diff
@@ -162,31 +162,6 @@ plots.getSubplotData = function getSubplotData(data, type, subplotId) {
...
-/**
- * Get calcdata traces(s) associated with a given subplot
- *
- * @param {array} calcData (as in gd.calcdata)
- * @param {string} type subplot type
- * @param {string} subplotId subplot id to look for
- *
- * @return {array} array of calcdata traces
- */
-plots.getSubplotCalcData = function(calcData, type, subplotId) {
```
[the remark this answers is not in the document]
alexcjohnson: > Why did you move this to a separate file?

I don't recall if it made more circular deps or not, but I don't like `plots/plots` - it has its fingers in everything. Plus I was making 2 (3 as of the next commit) similar simple functions for filtering down `data` or `calcdata` that only depend on `Registry` so it seemed natural to group them in their own file.

src/plots/plots.js
alexcjohnson: Perhaps we just take the first `basePlotModule` and make a subplot from it? Or maybe that's too complicated and we only care about empty plots for contexts like the editor, so if there's no `cartesian` we just leave it blank. Either way shouldn't be too hard, there's not that much code involved in the empty fallback.

src/plots/ternary/ternary.js
alexcjohnson: Yeah, that was a bit of a hack, we could probably find a way to clean it up for derived axes like this. When I got rid of `overgrid` and put everything in `gridlayer`, it meant nesting inside `gridlayer` and [selecting](https://github.com/plotly/plotly.js/pull/2227/files/2c9d90fd26952b91290540c35ef284f9aa6d90ed..3625be4e516376ff4ce65f0c0223d1d9b99c2d24#diff-f6d5ed25a5d6f57c70063e1e1bbc407dR2199) that nesting later.

test/jasmine/tests/annotations_test.js
alexcjohnson: > Why did you change this?

Now an annotation is enough to create a new subplot, so I had to switch these to invalid references.

test/jasmine/tests/gl_plot_interact_test.js
alexcjohnson: when you `expect(somePotentiallyBigObject).toBeUndefined()`, a failure swamps you with the whole stringified object. The killer was `expect(scene)` whose string representation was so big it took me forever to find the tracebacks. If the object you might get on failure is small, `toBeUndefined` is still a good idea as it shows you what you got.

test/jasmine/tests/transform_groupby_test.js
alexcjohnson: right, that's the disadvantage of this kind of precalculation, which I've added a lot of lately 🙈

👍  for moving toward `assets/supply_defaults.js`.

src/plots/cartesian/include_components.js
```diff
@@ -0,0 +1,73 @@
...
+module.exports = function makeIncludeComponents(containerArrayName) {
+    return function includeComponents(layoutIn, layoutOut) {
+        var array = layoutIn[containerArrayName];
+        if(!Array.isArray(array)) return;
+
+        var Cartesian = Registry.subplotsRegistry.cartesian;
+        var idRegex = Cartesian.idRegex;
+        var subplots = layoutOut._subplots;
+        var xaList = subplots.xaxis;
+        var yaList = subplots.yaxis;
```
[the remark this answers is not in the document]
alexcjohnson: Ah right - but actually `listIds` is the direct match to `fullLayout._subplots.(x|y)axis` (with some `_subplots.gl3d` shenanigans thrown in for good measure), as `list` returns the axis objects. I'll take a stab at removing `listIds` and simplifying the others.

src/plots/cartesian/graph_interact.js
```diff
@@ -100,36 +92,42 @@ module.exports = function initInteractions(gd) {
...
             // x axis draggers - if you have overlaid plots,
             // these drag each axis separately
-            if(isNumeric(y0)) {
-                if(xa.anchor === 'free') y0 -= fullLayout._size.h * (1 - ya.domain[1]);
-                dragBox(gd, plotinfo, xa._length * 0.1, y0,
+            if(subplot === xa._mainSubplot) {
+                // the y position of the main x axis line
+                var y0 = xa._mainLinePosition;
+                if(xa.side === 'top') y0 -= DRAGGERSIZE;
+                // if(xa.anchor === 'free') y0 -= fullLayout._size.h * (1 - ya.domain[1]);
```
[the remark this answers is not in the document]
alexcjohnson: thx -> 698561c

src/plots/plots.js
```diff
@@ -526,6 +419,50 @@ plots.supplyDefaults = function(gd) {
...
+    var out = {};
+    var i, j;
+
+    if(!collectableSubplotTypes) {
+        collectableSubplotTypes = [];
+
+        var subplotsRegistry = Registry.subplotsRegistry;
+        var subplotTypes = Object.keys(subplotsRegistry);
+
+        for(i = 0; i < subplotTypes.length; i++) {
```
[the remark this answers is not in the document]
alexcjohnson: Thanks for pointing that out! Swapped out `Object.keys` for `for-in` throughout plots/plots -> 8804311

src/plots/plots.js
```diff
@@ -831,8 +776,60 @@ plots.linkSubplots = function(newFullData, newFullLayout, oldFullData, oldFullLa
...
+
+    // and lastly, delete any titles we don't need anymore
+    // check if subplots have changed, and if so clear old titles
+    if(oldSubplotList.xaxis && oldSubplotList.yaxis) {
+        var oldAxIDs = oldSubplotList.xaxis.concat(oldSubplotList.yaxis);
+        var newAxIDs = newSubplotList.xaxis.concat(newSubplotList.yaxis);
+
+        for(i = 0; i < oldAxIDs.length; i++) {
+            if(newAxIDs.indexOf(oldAxIDs[i]) === -1) {
+                oldFullLayout._infolayer.selectAll('.g-' + oldAxIDs[i] + 'title').remove();
```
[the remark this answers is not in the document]
alexcjohnson: Ha good catch! I had basically duplicated it there -> moved the new one into `Cartesian.clean` and deleted the (less effective) old one -> 4786b6e

src/plots/plots.js
alexcjohnson: In fact we don't! Good 👀 
We *depend on* gl2d inside `finalizeSubplots` but it's irrelevant if we have *only* gl2d. 6e62b35

src/plots/cartesian/include_components.js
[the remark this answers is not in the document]
alexcjohnson: Cleaned up `list` and `listIds` in 0506f88 and 2aa1545

src/plots/plots.js
alexcjohnson: I'd be surprised if that's the only conflict this PR makes with regl-2d 😈

src/plots/plots.js
```diff
@@ -831,8 +776,60 @@ plots.linkSubplots = function(newFullData, newFullLayout, oldFullData, oldFullLa
 
         ax._anchorAxis = ax.anchor === 'free' ?
             null :
-            Plotly.Axes.getFromId(mockGd, ax.anchor);
+            axisIDs.getFromId(mockGd, ax.anchor);
+    }
+
+    for(i = 0; i < axList.length; i++) {
```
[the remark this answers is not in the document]
alexcjohnson: > Any reason for putting this here instead of e.g. `Cartesian.plot`?
>
> I'm thinking this has to do with being compatible with `editType: 'ticks'`, but I want to confirm.

Actually I'm not entirely sure. `_mainSubplot` felt structural to me, in a way that I wanted it available sooner rather than later, but it's possible it could be deferred to `Cartesian.plot`.

src/plots/plots.js
etpinard: done in https://github.com/plotly/plotly.js/issues/420#issuecomment-354509483

src/plots/ternary/ternary.js
etpinard: Ok thanks for the info. I'll need to hack in something similar for polar. Maybe I'll try to clean it up :ok_hand:

etpinard:
A well-earned :dancer:

I'll take care of https://github.com/plotly/plotly.js/pull/2227#discussion_r159070279 while trying to get the tests to pass.

PR #2232, opened 2018-01-03
don't write out strings from find_locale_strings during tests

alexcjohnson:
but we still want to run the rest of the script as it ensures we wrapped all strings

@etpinard as discussed in https://github.com/plotly/plotly.js/pull/2227#discussion_r159028225 so we're not modifying `dist/` when we don't intend to. `build` will still produce a new `translation_keys.txt` file.


etpinard:
Nicely done :dancer:

PR #2235, opened 2018-01-07
You are using a HTML5 DOCTYPE. For HTML5 script tags you can omit the type attribute when its value …

jbampton:
…is JavaScript. From MDN -> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script - Omitted or a JavaScript MIME type: For HTML5-compliant browsers this indicates the script is JavaScript. HTML5 specification urges authors to omit the attribute rather than provide a redundant MIME type.  https://www.w3.org/TR/2011/WD-html5-20110405/scripting-1.html#script


alexcjohnson:
Thanks @jbampton - It does seem like the wider web world is getting rid of these tags. For reference, see https://stackoverflow.com/a/23254634/9188800 - since that was posted in 2014, google and amazon have both joined facebook getting rid of it, linkedin and instagram still have it.

PR #2105, opened 2017-10-19
Restore dispatching 'click' on right-click

AlexVvx:
Issue [#6353](https://community.plot.ly/t/plotly-click-event-stopped-firing-after-update-on-right-click/6353) and [#6352](https://community.plot.ly/t/annotation-right-click/6352)

Restore dispatching 'click' on right-click

Issue #2057, opened 2017-10-03
Centralize drag handlers


monfera:
Followup of https://github.com/plotly/plotly.js/pull/2052#discussion_r142463300

alexcjohnson:
#2241 adjusted `dragElement` to take care of `click` events as well. This has a number of benefits: it simplifies all the `click` vs (drag)`done` logic, and ensures that you do exactly one of those - a mousedown/up must be either a click or a drag. It also allows us to capture right clicks. Some TODOs came up in that PR:
- Any trace type that uses `selection.on('click', clickHandler)` would benefit from being converted to `dragElement`, for the above mentioned reasons. This includes at least `geo`, `mapbox`, `parcoords`, `pie`, `sankey`, `table`. `geo` has a couple of additional issue: right-click starts a pan/select, which then  you have to click again to get out of; it also needs to have hover effects removed when you start selecting or panning.
- Annotations have a `plotly_annotationclick` event that should also be extended to support right-click.
- Possibly create a config parameter to manage right-click behavior. Right now the hacky behavior that some users depend on is that if you attach a `contextmenu` handler to `gd`, and call `preventDefault()` in it, this prevents the native context menu from appearing, and then a right click (for those traces using `dragElement` anyway) get a `plotly_click` event and the dev can look at `button(s)`, `ctrlKey`, etc to determine what kind of a click it was. But we don't want this to always, as mostly devs won't be thinking about what kind of click they're getting, and folks may want to keep the native context menu accessible. A config option seems like a clean way to expose this.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2241, opened 2018-01-10
Right click fix

### LINKED ISSUE #2101: Some folks are using right-click on data points in their apps
... and PR https://github.com/plotly/plotly.js/pull/2087 broke that :bomb: 

See https://community.plot.ly/t/plotly-click-event-stopped-firing-after-update-on-right-click/6353 and https://community.plot.ly/t/annotation-right-click/6352.

What should we do?

- Revert #2087 ?
- add a config option that allows `plotly_click` on right-clicks?
- anything else?

alexcjohnson:
fixes #2101 
supersedes #2105 - @AlexVvx I started with your PR, thinking I could just add the right tests to it and be done... unfortunately I found that we had some deeper problems in both the source code and in the tests so it became a much bigger project.

The test problem is that `assets/click` was unrealistic - it generated a sequence of events that's fine for regular clicks, but right clicks and ctrl-clicks (which we were using all over the place to test "modified click behavior", making a bunch of trace types *look* like they supported right-click when they really didn't) take a different pathway, and importantly for some users, it also matters whether the `contextmenu` event is canceled.

The source problem is that we're inconsistent with what we call a click - we have parallel systems handling drag and click events. This isn't only a problem for right clicks (which don't trigger `click` events at all, so it's a huge problem for them.), it's also inconsistent in subtle ways for our other interactions. For example, independent systems might see both a drag (to make a pan, zoom, or select) AND a click; or they might see neither one! Neither situation is desirable: users should be able to expect that one mouse action will create one response. So I changed the way `dragElement` works: it now takes an extra option, `clickFn`, ie what to do in the event of a click. Now you know that the sequence of events in a `dragElement` action will be either:

`prepFn` -> `moveFn` 1 or more times -> `doneFn` (a drag, even if it ends exactly where it started)

or:

`prepFn` -> `clickFn` (a click)

This has a nice side effect that `doneFn` can be much simpler, it only has one argument (the original event) and doesn't need all the `if(dragged)` branches it had before. `clickFn` is also pretty simple, and to my eye keeping the two separate makes them much more readable.

There are a whole bunch of `TODO`s scattered about in here. Before we merge this we should probably collect them into a new issue or two and 🔪 them from the code - most are about right-click scenarios we could support but currently do not, there's at least one about an unrelated interaction issue I encountered along the way. I don't think any are necessary for 1.32 though.

@etpinard please take a look!

### DIFF
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -334,25 +334,28 @@ module.exports = function draw(gd) {
                 xf = dragElement.align(newX, 0, gs.l, gs.l + gs.w, opts.xanchor);
                 yf = dragElement.align(newY, 0, gs.t + gs.h, gs.t, opts.yanchor);
             },
-            doneFn: function(dragged, numClicks, e) {
-                if(dragged && xf !== undefined && yf !== undefined) {
+            doneFn: function() {
+                if(xf !== undefined && yf !== undefined) {
                     Plotly.relayout(gd, {'legend.x': xf, 'legend.y': yf});
-                } else {
-                    var clickedTrace =
-                        fullLayout._infolayer.selectAll('g.traces').filter(function() {
-                            var bbox = this.getBoundingClientRect();
-                            return (e.clientX >= bbox.left && e.clientX <= bbox.right &&
-                                e.clientY >= bbox.top && e.clientY <= bbox.bottom);
-                        });
-                    if(clickedTrace.size() > 0) {
-                        if(numClicks === 1) {
-                            legend._clickTimeout = setTimeout(function() { handleClick(clickedTrace, gd, numClicks); }, DBLCLICKDELAY);
-                        } else if(numClicks === 2) {
-                            if(legend._clickTimeout) {
-                                clearTimeout(legend._clickTimeout);
-                            }
+                }
+            },
+            clickFn: function(numClicks, e) {
+                var clickedTrace =
+                    fullLayout._infolayer.selectAll('g.traces').filter(function() {
+                        var bbox = this.getBoundingClientRect();
+                        return (e.clientX >= bbox.left && e.clientX <= bbox.right &&
+                            e.clientY >= bbox.top && e.clientY <= bbox.bottom);
+                    });
+                if(clickedTrace.size() > 0) {
+                    if(numClicks === 1) {
+                        legend._clickTimeout = setTimeout(function() {
                             handleClick(clickedTrace, gd, numClicks);
+                        }, DBLCLICKDELAY);
+                    } else if(numClicks === 2) {
+                        if(legend._clickTimeout) {
+                            clearTimeout(legend._clickTimeout);
                         }
+                        handleClick(clickedTrace, gd, numClicks);
                     }
                 }
             }
--- a/src/plots/cartesian/graph_interact.js
+++ b/src/plots/cartesian/graph_interact.js
@@ -84,10 +84,6 @@ module.exports = function initInteractions(gd) {
                 dragElement.unhover(gd, evt);
             };
 
-            maindrag.onclick = function(evt) {
-                Fx.click(gd, evt, subplot);
-            };
-
             // corner draggers
             if(gd._context.showAxisDragHandles) {
                 dragBox(gd, plotinfo, xa._offset - DRAGGERSIZE, ya._offset - DRAGGERSIZE,
--- a/src/traces/pie/plot.js
+++ b/src/traces/pie/plot.js
@@ -171,7 +171,7 @@ module.exports = function plot(gd, cdpie) {
 
                 function handleClick() {
                     // TODO: this does not support right-click. If we want to support it, we
-                    // would likely need to change mapbox to use dragElement instead of straight
+                    // would likely need to change pie to use dragElement instead of straight
                     // mapbox event binding. Or perhaps better, make a simple wrapper with the
                     // right mousedown, mousemove, and mouseup handlers just for a left/right click
                     // mapbox would use this too.
--- a/src/traces/pie/plot.js
+++ b/src/traces/pie/plot.js
@@ -170,6 +170,11 @@ module.exports = function plot(gd, cdpie) {
                 }
 
                 function handleClick() {
+                    // TODO: this does not support right-click. If we want to support it, we
+                    // would likely need to change mapbox to use dragElement instead of straight
+                    // mapbox event binding. Or perhaps better, make a simple wrapper with the
+                    // right mousedown, mousemove, and mouseup handlers just for a left/right click
+                    // mapbox would use this too.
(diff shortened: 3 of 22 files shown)


src/components/legend/draw.js
```diff
@@ -334,25 +334,28 @@ module.exports = function draw(gd) {
...
-                    if(clickedTrace.size() > 0) {
-                        if(numClicks === 1) {
-                            legend._clickTimeout = setTimeout(function() { handleClick(clickedTrace, gd, numClicks); }, DBLCLICKDELAY);
-                        } else if(numClicks === 2) {
-                            if(legend._clickTimeout) {
-                                clearTimeout(legend._clickTimeout);
-                            }
+                }
+            },
+            clickFn: function(numClicks, e) {
```
etpinard: Wow. This looks so much cleaner by splitting `done` and `click` logic. :100:

src/plots/cartesian/graph_interact.js
```diff
@@ -84,10 +84,6 @@ module.exports = function initInteractions(gd) {
                 dragElement.unhover(gd, evt);
             };
 
-            maindrag.onclick = function(evt) {
-                Fx.click(gd, evt, subplot);
```
etpinard: this is now defined in `cartesian/dragbox.js` correct?

src/traces/pie/plot.js
```diff
@@ -170,6 +170,11 @@ module.exports = function plot(gd, cdpie) {
                 }
 
                 function handleClick() {
+                    // TODO: this does not support right-click. If we want to support it, we
+                    // would likely need to change mapbox to use dragElement instead of straight
```
etpinard: we would need to change _pie_ to use dragElement,

correct?

test/jasmine/tests/click_test.js
```diff
@@ -65,24 +66,62 @@ describe('Test click interactions:', function() {
...
+            click(marginPos[0], marginPos[1]);
+            expect(futureData).toBe(undefined);
+            expect(clickPassthroughs).toBe(1);
+            expect(contextPassthroughs).toBe(0);
         });
 
         it('should contain the correct fields', function() {
             click(pointPos[0], pointPos[1]);
             expect(futureData.points.length).toEqual(1);
+            expect(clickPassthroughs).toBe(2);
```
etpinard: so, in a real world scenario, this here would be `.toBe(0)`?

etpinard: i.e. the `plotly_click` eats up `gd.on('click')`?

test/jasmine/tests/click_test.js
```diff
@@ -98,49 +137,79 @@ describe('Test click interactions:', function() {
...
+            // Testing the specific behavior that some users depend on
+            // See https://github.com/plotly/plotly.js/issues/2101
+            // If and only if you cancel contextmenu, by doing something like:
+            //
+            //   gd.addEventListener('contextmenu', function(e) { e.preventDefault(); })
+            //
+            // then we pass right-click on data points through to plotly_click events.
+            // Devs using this need to be aware then to check eventData.event
+            // and figure out if it had a button or ctrlKey etc.
+            it('should contain the correct fields', function() {
```
etpinard: Amazing. Thanks for :lock: this down!

src/traces/pie/plot.js
etpinard: Perhaps we could discuss future `dragElement` generalizations in https://github.com/plotly/plotly.js/issues/2057? 

`table`, `sankey` and `parcoords` would also benefit from using `dragElement`.

test/jasmine/tests/dragelement_test.js
```diff
@@ -76,34 +76,66 @@ describe('dragElement', function() {
...
+            element: this.element,
+            gd: this.gd,
+            clickFn: function() {
+                args = arguments;
+            },
+            moveFn: function() {
+                expect('should not call moveFn').toBe(true);
+            },
+            doneFn: function() {
+                expect('should not call doneFn').toBe(true);
```
etpinard: Oh nice. I forget we had unit tests for `dragElement` :ok_hand:

test/jasmine/tests/select_test.js
```diff
@@ -98,11 +98,11 @@ function assertEventCounts(selecting, selected, deselect, msg) {
 
 // events for box or lasso select mouse moves then a doubleclick
 var NOEVENTS = [0, 0, 0];
-// deselect gives an extra plotly_selected event on the first click
-// event data is undefined
-var BOXEVENTS = [1, 2, 1];
+// deselect used to give an extra plotly_selected event on the first click
+// with undefined event data - but now that's gone, since `clickFn` handles this.
```
etpinard: Oh. Some users might currently rely on `plotly_selected` sending `undefined` as event data.

etpinard:
Well done. I'm a big fan of the `doneFn` / `clickFn` split in `dragElement` :tada: 

I'm a little concern with how that affects `plotly_selected` event listeners that expect `undefined` (see https://github.com/plotly/plotly.js/pull/2241#discussion_r160723541). I hope there's a way to patch `cartesian/select.js` to leave the `select_test.js` suite unchanged, I don't think we can consider this a bug fix (in v1 at least).

src/plots/cartesian/graph_interact.js
alexcjohnson: > this is now defined in `cartesian/dragbox.js`, correct?

yep, now it's here: https://github.com/plotly/plotly.js/pull/2241/files#diff-1bd3dd52cddb8638876bbe17cef28a0dR175

src/traces/pie/plot.js
alexcjohnson: comment added to #2057 - I'll leave the TODOs present in the code then unless you object.

test/jasmine/tests/click_test.js
alexcjohnson: real world it would be `1`. `dragElement` eats up one click event but generates a new one. That was important in certain cases when we were using that synthetic event to trigger the `selection.on('click')` to trigger the `plotly_click`. We could consider removing it now, per [this TODO](https://github.com/plotly/plotly.js/pull/2241/files#diff-cb022fcf45f61114c1c73baba3164c1aR103) so it doesn't bubble up the tree after we've already generated our own events from it, but leaving it in is the most backward-compatible way to handle this.

test/jasmine/tests/select_test.js
alexcjohnson: put back the old behavior in https://github.com/plotly/plotly.js/pull/2241/commits/974cebf4e6f596d4fde8c84053aef43100b2ea10 and commented in #420 about removing it later.

etpinard:
Looks amazing :dancer:

Issue #1504, opened 2017-03-21
Axis titles overlapping axis tick labels


cbailiss:
Hello

I am having problems with the axis titles being far too close or overlapping the text of the labels (tick text) on each axis.

Example:
https://plot.ly/~cbailiss/5/incidence-by-age-range-and-gender/

The two axis titles should be further away from the numbers/categories on each axis.  How can this be accomplished?

If the font size of the axis tick labels is increased, both the positions of the axis titles and the positions of the plot (i.e. where the x and y axes are drawn) do not change to adapt to the increased space occupied by the axis tick labels.  Ever more of the labels overlaps with the title.  Eventually, the labels themselves do not fit into the plot.
https://plot.ly/~cbailiss/7/incidence-by-age-range-and-gender/

It seems the rendering/layout algorithm doesn't pay attention to the width of the labels, which seems like a significant bug.  Note, in the above examples, I am specifying a tickformat.  Is this the cause of the bug?

For reference, the two plots above were produced from R, with the following code:
```
plot_ly(AgeGenderIncidence3, x = ~AgeRange2, y = ~F, type = 'bar', name = 'F', marker = list(color = 'rgb(255,91,188)')) %>%
  add_trace(y = ~M, name = 'M', marker = list(color = 'rgb(91,217,255)')) %>%
  layout(title="<b>Incidence By Age Range and Gender</b>", 
         titlefont=list(family="Arial", size=18, color="#2E008B"),
         xaxis = list(title = '<b>Age Range</b>', 
                      titlefont=list(family="Arial", size=16, color="#2E008B"),
                      tickfont=list(family="Arial", size=11, color="#343434")), 
         yaxis = list(title = '<b>Incidence</b>', tickformat=',.0', hoverformat=',.0',
                      titlefont=list(family="Arial", size=16, color="#2E008B"),
                      tickfont=list(family="Arial", size=14, color="#343434")), 
         barmode = 'group')
```
In the second example above, the size 14 was changed to size 20.

An example (produced with a different library) that is more what I am looking for is:
https://1drv.ms/i/s!ApqRkiJX6gQSg5odXumwaUNXuSl23g

Thanks

Chris

etpinard:
duplicate of https://github.com/plotly/plotly.js/issues/296 or maybe I'm not understanding this issue correctly.

cbailiss:
Possibly, but that issue doesn't refer to the (a) axis titles, or (b) the effects that tickformat has.  
[...]

Issue #1199, opened 2016-11-25
Add legend 'auto' x|y and/or 'container' (x|y)ref


salim-b:
The problem is that the horizontal legend overlaps the x-axis title. It occurs if the legend is positioned below (rather than to the right of) the plot area using `layout = {legend: {"orientation": "h"}};`. It aggravates in combination with multiline x-axis titles (using `<br>` tags).

Currently there's no way to control the spacing between the legend and the plotting area. Everything is auto-scaled, it is useless to set the bottom margin (`layout = {margin: {"b": ...}};`) in this case because it refers to the space _below_ the legend. The only way to make the whole x-axis title visible is to increase the height of the whole plot until the autoscaling sufficiently spreads the legend and the plotting area apart.

I try to illustrate the problem with the following 3 images which differ only in frame height:

_height 700px:_
![example_height700](https://cloud.githubusercontent.com/assets/20040931/20580578/6831cab0-b1d3-11e6-8162-c67bd0c11cf3.png)

_height 1000px:_
![example_height1000](https://cloud.githubusercontent.com/assets/20040931/20580582/6cd3f91c-b1d3-11e6-904f-edb748be01f7.png)

_height 1300px:_
![example_height1300](https://cloud.githubusercontent.com/assets/20040931/20580585/71203f4e-b1d3-11e6-81ad-6f0b3e8b0b3c.png)

[Here](https://plot.ly/~slim-b/225) is the above plot hosted on plot.ly.

n-riesco:
This is happening because `layout.margin.b` is set to 0 and the default position for horizontal legends is:
```js
x: 0,
xanchor: 'left',
y: 1.1,
yanchor: 'bottom'
```

@etpinard Would it make sense to make the defaults smarter and check the value of `layout.margin.b`?

salim-b:
>... and check the value of `layout.margin.b`?

I don't think `layout.margin.b` is central to this. If you play around a little with different combinations of `layout.margin.b` and `layout.legend.yanchor`, you quickly see that in the plot above you'll get nowhere desirable... actually the default of `layout.legend.yanchor`, which seems to be `top` not `bottom` (at least when I'm plotting in `R`), leads to the best results.

The problem in my eyes is that – when `layout.legend.orientation` is horizontal and the rest of the legend positioning parameters left to its default values – the legend overlaps the plotting area in the first place. In the default case the legend should always be placed below the x-axis, then it would be fine.

etpinard:
[...]

Issue #296, opened 2016-02-29
Make auto-margin step dependent on axis label span 


mnoble01:
Hi there, I have some graphs with particularly long labels/axis string values and I'm wondering if there's a way to get them to fit without changing the `layout.margin` option because I'm using dynamic data.

Here's CodePen showing a longer label being cut off on both a vertically and horizontally-oriented bar chart - http://codepen.io/mnoble01/pen/zqxZQx.

etpinard:
There's no automated way to do that at the moment. Making the [`auto-margin`](https://github.com/plotly/plotly.js/blob/e39244de7791c58656e6bfdd4c5eb0aa435c7990/src/plots/plots.js#L876-L954) step dive into the axis labels would be a great addition though.

Renaming this issue accordingly.

aburato:
![image](https://cloud.githubusercontent.com/assets/17029400/14012557/c85d3c50-f1a6-11e5-8b31-a797ac7fa681.png)

Hi guys, adding my +1 to this... Please remember to take Pies into account as well :)

kbeck527:
+1 from me too

gracecarey:
+1

tilleryj:
+1

druvisc:
+1

merlux:
+1

hy9be:
Observed the same issue.

edwardbetan:
+1

donnyv:
+1

brian428:
+1

hy9be:
+1

bpaszcza:
+1

mariehooper:
+1

aaneitchik:
+1

merlux:
Little workarround:

You can calculate the needed margins for your layout out of the SVG:

```
 function calculateLegendMargins(containingElement){
      var refBB = $(containingElement).find(".gridlayer")[0].getBBox();
      var legendContainingBB = $(containingElement).find(".plot")[0].getBBox();
      var legendBB = $(containingElement).find(".xaxislayer")[0].getBBox();

      var margin = {};
      margin.b = legendBB.height;
      margin.r = (legendBB.width+legendBB.x+legendContainingBB.x) - (refBB.width+refBB.x);
      return margin;
} 

```
The Plot will "jump" on pageload, but the legends should be in place.
Only working with Plotly v1.24.2, only tested with single bar plot.

Use the function with:
```
[...]

PR #2243, opened 2018-01-10
automatic margins for long labels

nicolaskruchten:
Proof of concept for addressing #296  and #1504.

The `Plots.autoMargin` and `calcBoundingBox` machinery is all already in place so I've wired them together such that margins can only grow, otherwise it's easier to trigger interactions between various axes such that the layout machinery loops without converging. Unfortunately this means that it's possible to get into situations where e.g. the left-hand margin is big enough to accommodate some long Y-axis labels which are then hidden because some long X-axis labels caused the plot to become shorter.

Known issues with this PR: 
* the axis autoMargin calls don't mesh very well with the legend autoMargin calls, so it's still possible for the legend to sit on top of tick labels. Resolving that seems like it would involve modifying `Plots.doAutoMargin` to sum the requested `_pushmargins` somehow. This is a known issue: #1199
* this behaviour should have some controls associated with it to turn it on/off


nicolaskruchten:
Also note that I didn't test this with subplots, I wanted to kick off some discussion first :)

nicolaskruchten:
And here's what the example looks like, with a legend, two Y axes and some big axis titles:

image

etpinard:
This looks like a very solid first step. :muscle: 

> this behaviour should have some controls associated with it to turn it on/off

I think so. I'd vote for something like `layout.automargins` which would be `false` by default in v1.x and `true` starting in [`v2`](https://github.com/plotly/plotly.js/issues/420).

This new attribute will require some fancy`editType` and `relayout`  logic similar to [`autorange`](https://github.com/plotly/plotly.js/blob/90218e5e9b73fd503f366b8ccf531464df2f9f2b/src/plot_api/plot_api.js#L1885-L1893) to work correctly on updates. Moreover, we should make sure that labels automargins are only computed when needed (e.g. when axis labels or the plot `domain` change). `Plots.doAutoMargins` can slow things down.

We should also test various combinations of things that can push margins e.g. colorbars, legend, range sliders, update menus and layout sliders (anything else?).

Working this patch in `Axes.doTicks` should make this new logic work automatically with ternary and polar (coming soon), but we'll need to patch [`pie/plot.js`](https://github.com/plotly/plotly.js/blob/master/src/traces/pie/plot.js) to make sure pie labels get the same treatment - that could be in a different PR though :wink: 

One final comment, should `anchor: 'free'` axes also be allowed to bump the margins? I think so, but that might be debatable and/or pretty tricky to get right in situations like mock [20](https://rreusser.github.io/plotly-mock-viewer/#20).

nicolaskruchten:
Re the testing with other things that push margins, it behaves just as detailed in #1199 ... no improvement but no worse either. The margins move such that there is enough room for the labels, and then the legend still sits on top of the labels. So I would consider that a separate issue that we can tackle as part of solving #1199 etc?

src/plots/cartesian/tick_label_defaults.js
```diff
@@ -50,6 +50,8 @@ module.exports = function handleTickLabelDefaults(containerIn, containerOut, coe
     }
 
     if(axType !== 'category' && !options.noHover) coerce('hoverformat');
+
+    coerce('ticklabelsautomargin');
```
etpinard: @nicolaskruchten you'll have to add an option to opt-out of coercing `ticklabelautomargin` in `handleTickLabelDefaults` so that other callees 

image


don't cough on it.

src/plots/cartesian/tick_label_defaults.js
```diff
@@ -50,6 +50,8 @@ module.exports = function handleTickLabelDefaults(containerIn, containerOut, coe
     }
 
     if(axType !== 'category' && !options.noHover) coerce('hoverformat');
+
+    if(axType === 'cartesian') coerce('ticklabelsautomargin');
```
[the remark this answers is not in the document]
etpinard: Oh does `coerce('ticklabelsautomargin')` have an effect when `showticklabels` is false? More precisely, do the axis title font settings and/or `ticklen` affect to auto-margin computations?

src/plots/cartesian/layout_attributes.js
```diff
@@ -344,6 +344,16 @@ module.exports = {
         editType: 'ticks',
         description: 'Determines whether or not the tick labels are drawn.'
     },
+    ticklabelsautomargin: {
```
etpinard: This might be too verbose. Would `(x|y)axis.automargin` suffice? Or maybe `tickautomargin`? Or perhaps we're looking at this the wrong way, something like:

```js

var layout = {
  automargin: true || false,
  automarginmode: 'all' || 'ticks' || 'title' || 'xaxis.title' || 'legend' || ...
  // or any combination of these using '+',  
  // a 'flaglist' attribute like scatter 'mode'
}
```

might scale better.

@alexcjohnson @cldougl @chriddyp any thoughts on this?

src/plots/cartesian/tick_label_defaults.js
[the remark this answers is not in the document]
nicolaskruchten: I'm not sure what `coerce` actually does, to be honest, so I'm not sure if it will have an impact. What exactly does it do? Is it necessary here?

To answer your question more generally, the auto-margin calculation explicitly takes the axis title font size into account, and for the rest relies on the pre-existing bounding-box calculation.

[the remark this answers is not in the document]
etpinard: > I'm not sure what coerce actually does,

In brief, `coerce` fills in `fullLayout` with a sanitise version of its user layout value. Almost everywhere downstream of the `Plots.supplyDefaults` uses `fullLayout` and `fullData` value instead of `gd.layout` and `gd.data` including your new logic in `axes.js`. 

So, yes it is necessary :smile: 

>  the auto-margin calculation explicitly takes the axis title font size into account,

Ok great. So, auto-margin doesn't _just_ depend on tick settings like the attribute name `ticklabelsautomargin` is suggesting. We'll need to think about how to name this new attribute some more cc https://github.com/plotly/plotly.js/pull/2243#discussion_r165100920

> for the rest relies on the pre-existing bounding-box calculation.

Ok, so this should depend on `ticklen` (we should double-check though). Knowing which attributes  impact the auto-margin computation is important to make sure relayout calls like `Plotly.relayout(gd, 'xaxis.ticklen', /**/)` update the (auto) margin when `ticklabelsautomargin` is turned on.

src/plots/cartesian/layout_attributes.js
cldougl: +1 for `automargin` / `automarginmode` or at least `(x|y)axis.automargin` - this is more clear if attributes other than ticks/tick labels contributes to this (now or in the future).

nicolaskruchten:
@etpinard which files should I look into for managing the dependencies among attributes for restyle/relayouts? I.e. the fact that my new option depends on `showticklabels` etc: how do I encode that?

fabiangehring:
Great work, thx. For me, missing automatic label length is one of the big issues when using plotly. Are you still working on this? Would appreciate!

nicolaskruchten:
Disclaimer: this is what it looks like in an extreme case with subplots... Subplots don't push against each other, just the edges of the canvas, and pushing is just in one axis, so angled x-ticks don't push against the sides.

image

nicolaskruchten:
Also disclaimer: this PR will not fix #2428 :)

nicolaskruchten:
@etpinard @alexcjohnson ready for review! :D

etpinard:
Looking good! Thanks for :microscope:  and :pencil2: https://github.com/plotly/plotly.js/issues/2428

> One final comment, should anchor: 'free' axes also be allowed to bump the margins? I think so, but that might be debatable and/or pretty tricky to get right in situations like mock 20.

Don't hate me for this, but I think we should try to get support for `anchor: 'free'` in this PR (I can help out tomorrow). Main reason: after this PR, graphs like:

```js
Plotly.newPlot(gd, [{
  y: ['one', 'two', 'three']
}, {
  x: [1, 2, 3],
  y: ['one', 'twooooooooooooooooooooooooooooooooooooooooooo', 'three'],
  yaxis: 'y2'
}], {
  showlegend: false,
  xaxis: {
    domain: [0.2, 1]
  },
  yaxis2: {
    automargin: true,
    overlaying: 'y',
    position: 0
  }
})
```

will look like:

image


but should really look like:

image

Adding support for `automargin` for `anchor: 'free'` axes in a future PR isn't really an option, as it would cause a _breaking_ change (unless we turn `automargin` into an enumerated with values `true`, `false` or `free-also` which would be pretty rough on the :eyes: ). 

Luckily, this should be pretty easy, a simple patch like this:

image

got the trivial case working. I suspect this might break `doTicks` for non-cartesian axes, but we can easily fix that with:

image


and similarly in `src/plots/ternary/ternary.js`.

nicolaskruchten:
Sure, that seems pretty easy! I'll try to get that integrated tomorrow. Is
it ever possible to have neither position nor anchoredAxis?

etpinard:
> it ever possible to have neither position nor anchoredAxis?

That shouldn't happen, I don't think.

nicolaskruchten:
Ready for re-review.

src/plot_api/edit_types.js
```diff
@@ -47,6 +47,7 @@ var layoutOpts = {
         '*plot* calls `Plotly.plot` but without first clearing `gd.calcdata`.',
         '*legend* only redraws the legend.',
         '*ticks* only redraws axis ticks, labels, and gridlines.',
+        '*margins* recomputes ticklabel automargins.',
```
alexcjohnson: Ah yes, that's a much better solution 🎉

src/plots/polar/polar.js
```diff
@@ -286,7 +286,10 @@ proto.updateRadialAxis = function(fullLayout, polarLayout) {
         position: 0,
 
         // dummy truthy value to make Axes.doTicks draw the grid
-        _counteraxis: true
+        _counteraxis: true,
+
+        // don't use automargins routine for labels
+        automargin: false
```
alexcjohnson: might be nice to let polar and ternary contribute to automargin in the future - but since this is an axis-level attribute I guess there's no problem ignoring that part for now.

alexcjohnson:
Nice job @nicolaskruchten - nothing more from me, looks great! 💃

etpinard:
All right! Time to merge this thing :dancer:

yniknafs:
@etpinard Hey Étienne. Can't thank you enough for all your work on plotly. Whats the timeframe for merging these fresh PR merges into the distributed plotly package. Curious if I should pull the repo and make my own local copy or wait for it to be in the distribution.

etpinard:
This PR has been released in [`1.35.0` back on March 7 ](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#1350----2018-03-07).

yniknafs:
@etpinard 

Gotcha, awesome. I just saw there was a new release yesterday, thanks for that. 

I'm puling my hair out trying to figure out why my automargin isn't working. Not sure if you can help me at all. both the legend + axis label collide with the axis labels.

Here's my code for the layout: 
```
const layout = (scale = 'linear', layoutData, name) => {
  let orientation = 'h';
  let subtitle = '';
  if (layoutData !== undefined) {
    orientation = (layoutData.legendOrientation === 'h') ? 'h' : 'v';
    subtitle = layoutData.subtitle;
  }

  const layoutObject = {
    title: (name) ? `<b>${name} (Expression)</b> <br> ${subtitle}` : '<b>No Gene Selected</b>',
    height: (layoutData) ? layoutData.height : 600,
    width: (layoutData) ? layoutData.width : '',
    autosize: true,
    showlegend: (layoutData) ? layoutData.showLegend : true,
    legend: {
      orientation,
    },
    font: {
      family: 'Helvetica',
    },
    yaxis: {
      title: (!name) ? '<b>TPM</b>' : `<b> ${name} (TPM) </b>`,
      automargin: true,
      showline: false,
      ticks: 'outside',
      showticklabels: true,
      rangemode: 'tozero',
      type: scale,
    },
    xaxis: {
      title: (layoutData) ? layoutData.xaxis : '',
      automargin: true,
      // showgrid: false,
      // zeroline: false,
      // showline: scale === 'log',
      // ticks: 'outside',
      // showticklabels: true,
      // type: 'category',
      // categoryorder: 'array',
      // categoryarray: [],
    },
  };
```

The data is a little complex, but I'll try to make a codepen to emulate the exact plot I'm doing if you don't have any ideas initially. 

screen shot 2018-04-19 at 11 35 05 am

screen shot 2018-04-19 at 11 36 15 am

screen shot 2018-04-19 at 11 38 48 am

etpinard:
That's part of the know issues in https://github.com/plotly/plotly.js/pull/2243#issue-162224099

> the axis autoMargin calls don't mesh very well with the legend autoMargin calls, so it's still possible for the legend to sit on top of tick labels.

yniknafs:
@etpinard Got it. In my second screenshot, though. I didn't use a legend at all, and my xaxis label is still colliding with the tick labels. 

I commented out all legend options and it still happens. 

image

nicolaskruchten:
@yniknafs can you please create a new issue for this and I'll investigate?

nicolaskruchten:
One thought on this is that the `automargins` code bails out if the new margins are bigger than half the plot dimension, which is likely what's happening in your screenshot above.

alexcjohnson:
Aside from the legend overlap, this question about where the axis title goes has come up before - though I'm not aware of an issue discussing it. @yniknafs as you've deduced, we currently have the title avoid the bounding box of each tick label independently, so if there's a long label near the edge somewhere, that won't cause the title to shift outward (until you pan the axis to put that long label in conflict with the title, then it'll move). But you're right, in certain cases it would be nice to have a mode where the title stays out beyond the longest label even if they wouldn't actually cover each other.

alexcjohnson:
> actually automargins is that mode

Ah I see, didn't realize those two were coupled but you're right, that does generally work and seems like a good standard behavior. I still think it would be nice to be able to decouple them, ie *by default* it works like you have it but you could use another attribute to force either separate-label avoidance with `automargins` or all-label avoidance without.

yniknafs:
@nicolaskruchten @alexcjohnson 
Will great a new issue with this. FYI, I tried to make the plot bigger to test the 1/2size bailout, and it seems that no matter how large I make the plot, I still see the behavior. 

image

johnbenac:
Please add this to the 3D scene. It wont work in scatter3D. I'm getting the following error:

`ValueError: Invalid property specified for object of type plotly.graph_objs.layout.scene.XAxis: 'automargin'

    Valid properties:
        autorange
            Determines whether or not the range of this axis is
            computed in relation to the input data. See `rangemode`
            for more info. If `range` is provided, then `autorange`
            is set to False.
        backgroundcolor
            Sets the background color of this axis' wall.
        calendar
            Sets the calendar system to use for `range` and `tick0`
            if this is a date axis. This does not set the calendar
            for interpreting data on this axis, that's specified in
            the trace or via the global `layout.calendar`
        categoryarray
            Sets the order in which categories on this axis appear.
            Only has an effect if `categoryorder` is set to
            "array". Used with `categoryorder`.
        categoryarraysrc
            Sets the source reference on plot.ly for  categoryarray
            .
`

because the automargins property belongs to "plotly.graph_objs.layout.XAxis:" and not "plotly.graph_objs.layout._**scene**_.XAxis:"

For my particular scatter3d plots, I need to use the scene.

PR #2251, opened 2018-01-15
Mute event listener logging

### LINKED ISSUE #1795: Chrome is logging [Violation] errors in verbose console mode
First reported https://community.plot.ly/t/violation-at-console/4518/7 

To see it, open up https://codepen.io/etpinard/pen/ZyLQYy and turn on _Verbose_ mode

image

The error message points to this [line](https://github.com/plotly/plotly.js/blob/2940d950d7e39c1be44462681ddd1c2139a0a649/src/plots/cartesian/dragbox.js#L440) in `dragbox.js`.

```
[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. 
Consider marking event handler as 'passive' to make the page more responsive.
```

Observed in Chrome 59 and 58.

---

Anyone know what this is about?

dfcreative:
Fixes #1795


alexcjohnson:
Ah, I see - so basically this is just attaching the event in a different way that allows us to tell the browser "yes, we know this isn't passive, stop bugging us"?

I like the `_onwheel` etc solution. LGTM! 💃

src/plots/cartesian/dragbox.js
```diff
@@ -434,9 +434,14 @@ module.exports = function dragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
     // everything but the corners gets wheel zoom
     if(ns.length * ew.length !== 1) {
-        // still seems to be some confusion about onwheel vs onmousewheel...
-        if(dragger.onwheel !== undefined) dragger.onwheel = zoomWheel;
-        else if(dragger.onmousewheel !== undefined) dragger.onmousewheel = zoomWheel;
```
etpinard: Can we keep that `onmousewheel` handler in case we need it in older browsers?

alexcjohnson: I think the only difference between what @dfcreative has here and what we had before is that now we *always* attach to either `wheel` or `mousewheel`, whereas before we could potentially attach to neither if the browser doesn't support either one (I had to play with this - you may already know this but what's going on is if the browser supports `wheel`, then `element.onwheel = null` before any handler is attached, not `undefined`). And even if there is a case where the browser really doesn't support either, there shouldn't be any problem attaching to a nonexistent event.

etpinard: Oh I see. My bad. :ok_hand:

dfcreative:
@etpinard resolved

src/plots/cartesian/dragbox.js
```diff
@@ -1041,8 +1041,14 @@ function calcLinks(constraintGroups, xIDs, yIDs) {
...
-    if(element.onwheel !== undefined) element.onwheel = handler;
-    else if(element.onmousewheel !== undefined) element.onmousewheel = handler;
+    var wheelEventName = element.onwheel !== undefined ? 'wheel' : 'mousewheel';
+
+    if(element._onwheel) {
+        element.removeEventListener(wheelEventName, element._onwheel);
+    }
+    element._onwheel = handler;
+
+    element.addEventListener(wheelEventName, handler, {passive: false});
```
etpinard: Unfortunately, this might break older browsers, see:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support

I couldn't find a list of _which_ older browsers does not support `{passive: true|false}`, but we should probably play it safe here.

alexcjohnson: Good catch @etpinard - do we need a `has-passive` ala https://github.com/dfcreative/has-hover? Or even better, a module that would figure this out once and then export the correct 3rd arg to mark the listener active or passive:
```js
var eventOpts = require('event-opts');
element.addEventListener(wheelEventName, handler, eventOpts.active); // false or {passive: false}
element.addEventListener(wheelEventName, handler, eventOpts.passive); // false or {passive: true}
```

dfcreative: There is a polyfill [default-passive-events](https://www.npmjs.com/package/default-passive-events) for that. We could use detection logic from there, I agree with @alexcjohnson.

dfcreative: @etpinard @alexcjohnson https://github.com/plotly/plotly.js/pull/2251/commits/a32a21aac96e6efdeed5dffbcf193ecbf838a533 introduces `eventListenerOptionsSupported` method to detect proper way to attach events.

etpinard:
Nicely done. Merge away (when you get the chance) :dancer:

etpinard:
Dima must be travelling today. Merging this thing so that it's part of `1.33.0`.

PR #2257, opened 2018-01-17
localize "new text"

alexcjohnson:
@etpinard a quick one, a missing translation I noticed while working on localizing the editor https://github.com/plotly/react-plotly.js-editor/pull/246


etpinard:
:dancer: good catch!

PR #2263, opened 2018-01-18
fixes #2262 - Selection displayed on incorrect subplot

### LINKED ISSUE #2262: Selection displayed on incorrect subplot
When using subplots and performing selections if the selection is not removed by double clicking prior to performing a select on a different subplot the selection is drawn on the previous subplot rather than the current subplot.  However, when selecting only horizontal or vertical the handles are shown on the correct subplot.  The data returned from the event however is correct.

paulsputer:
Transform now set on updates as per path.

### DIFF
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -44,8 +44,22 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
         yAxisIds = dragOptions.yaxes.map(getAxId),
         allAxes = dragOptions.xaxes.concat(dragOptions.yaxes),
         filterPoly, testPoly, mergedPolygons, currentPolygon,
-        subtract = e.altKey;
+        subtract = e.altKey,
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Nice find! For perf though, I suspect this just needs to be moved outside the `enter()` where we had it [originally](https://github.com/plotly/plotly.js/pull/2263/files#diff-8c5d8f688f2bb41a18d35397e5f58864L71), to just `outlines.attr('transform', ...)` like you added, but doesn't need to be called in `moveFn` (which is called repeatedly on mousemove).

Also we may want to leave #2262 open until we figure out what we want to do with https://github.com/plotly/plotly.js/issues/2262#issuecomment-358737387

paulsputer:
I see yea, also looks like we need to reset the path to `path0` otherwise we get a draw error until a new path is created.

paulsputer:
Closing PR as it appears to have been fixed in ebaf705c61d6840734d2f63f8b7b7a454a316395

Issue #2253, opened 2018-01-16
candlestick graph hoverinfo shows strange labels for open/close/low/high


AfroMetal:
With update of `dash-core-components` to 0.17.0 my candlestick (Python: `plotly.graph_objs.Candlestick`) chart behaves strangely on hover. `hoverinfo` is set to `x+y` for date on x axis and open/close/high/low values on candle. It looks like an error in the underlying `plotly.js`

Before (dcc 0.15.5, based on old plotly.js) it showed only the values.

But now (dcc 0.17.0, based on new plotly.js 1.32.0) each candle value has a unwanted label `min:`, `max:`, `q1:`, `q3:` and `median:`.

First mentioned in (with example screenshots): https://community.plot.ly/t/dcc-graph-candlestick-hoverinfo-shows-strange-labels/7764

alexcjohnson:
Thanks @AfroMetal - candlestick is showing a bit more of its box plot roots than it should. Possibly came in with localization #2195 ?

PR #2264, opened 2018-01-18
Candlestick hover label fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2253

cc @alexcjohnson

### DIFF
--- a/src/traces/box/calc.js
+++ b/src/traces/box/calc.js
@@ -136,6 +136,11 @@ module.exports = function calc(gd, trace) {
             }
         };
 
+        // don't show labels in candlestick hover labels
+        if(trace._fullInput && trace._fullInput.type === 'candlestick') {
+            delete cd[0].t.labels;
+        }
+
         fullLayout[numKey]++;
         return cd;
     } else {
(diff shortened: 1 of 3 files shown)


src/traces/box/calc.js
```diff
@@ -136,6 +136,11 @@ module.exports = function calc(gd, trace) {
             }
         };
 
+        // don't show labels in candlestick hover labels
+        if(trace._fullInput && trace._fullInput.type === 'candlestick') {
+            delete cd[0].t.labels;
+        }
```
alexcjohnson: not a big deal, but 🐎  I might invert this logic and only add labels if we *do* want them.

alexcjohnson:
nonblocking https://github.com/plotly/plotly.js/pull/2264#discussion_r162461911, looks great 💃

src/traces/box/calc.js
etpinard: I thought about that. But, I prefer writing candlestick-only blocks like this one as _exceptions_ to a rule (the rule here being fill up `cd[0].t.labels`). Here's another [example](https://github.com/plotly/plotly.js/blob/55c4b3521c0cd426de6a52582f69852233c0284e/src/plots/cartesian/type_defaults.js#L121). 

I can't wait to rewrite `candlestick` and `ohlc` as good old (non-transform) trace types :upside_down_face:

alexcjohnson: >  I prefer writing candlestick-only blocks like this one as exceptions to a rule

Good point.

PR #2268, opened 2018-01-19
Update es.js

n-riesco:
* Added translation for `Click to enter radial axis title`
* Added translation for `new text`

### DIFF
--- a/lib/locales/es.js
+++ b/lib/locales/es.js
@@ -21,6 +21,7 @@ module.exports = {
         'Click to enter Plot title': 'Introducir el título de la Gráfica', // plot_api/plot_api.js:579
         'Click to enter X axis title': 'Introducir el título del eje X', // plots/plots.js:301
         'Click to enter Y axis title': 'Introducir el título del eje Y', // plots/plots.js:302
+        'Click to enter radial axis title': 'Introducir el título del eje radial',
         'Compare data on hover': 'Comparar datos al pasar por encima', // components/modebar/buttons.js:167
         'Double-click on legend to isolate one trace': 'Haga doble-clic en la leyenda para aislar una traza', // components/legend/handle_click.js:90
         'Double-click to zoom back out': 'Haga doble-clic para restaurar la escala', // plots/cartesian/dragbox.js:335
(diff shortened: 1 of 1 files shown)


lib/locales/es.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
         'Click to enter Plot title': 'Introducir el título de la Gráfica', // plot_api/plot_api.js:579
         'Click to enter X axis title': 'Introducir el título del eje X', // plots/plots.js:301
         'Click to enter Y axis title': 'Introducir el título del eje Y', // plots/plots.js:302
+        'Click to enter radial axis title': 'Introducir el título del eje radial',
```
n-riesco: Just nit-picking: here I've translated the original `radial axis`, but I think `polar axis` is more common.

lib/locales/es.js
```diff
@@ -60,6 +61,7 @@ module.exports = {
         'mean:': 'media:',                                            // traces/box/calc.js:133
         'median:': 'mediana:',                                        // traces/box/calc.js:128
         'min:': 'mín:',                                               // traces/box/calc.js:129
+        'new text': 'nuevo texto',
```
n-riesco: Grammatically, the translation should be `texto nuevo` (as is `insertar un texto nuevo`), but in software `nuevo texto` is more common (I guess this is the result of the influence of the original English).

lib/locales/es.js
alexcjohnson: Isn't `polar axis` ambiguous between angular and radial? Anyway that's a separate discussion, within the existing context this looks great, thanks!

lib/locales/es.js
alexcjohnson: Insidious, isn't it 😈 

Thanks @n-riesco , I'll merge once I get tests to pass; the new `regl` code is awesome but it's wreaking havoc on CI right now...

lib/locales/es.js
n-riesco: Oh, I've just realised that `radial axis` here is not the actual axis (i.e. the polar axis used as a reference to measure the angular coordinate).

alexcjohnson:
cc @JonasGutermuth @jdugge - at some point we'll need a more formal process for communicating new translation keys, but for now since we only have a few languages this'll do. Care to translate "Click to enter radial axis title" (new with polar 2.0 #2200) and "new text" (for blank annotations, I missed this in my initial string wrapping push) to German?

Issue #720, opened 2016-07-06
Hovering over super-imposed pts (only) considers hoverinfo of top trace


bxlxux:
Hello - I came across this bug https://github.com/plotly/plotly.js/issues/493 though I'm not sure it's the same issue.

I have this test codepen running with "hovermode":"closest" set. The text displays when hovering over the lines. However, when I change the cdn to 1.10.1 or later the hover no longer works. I can only get the text to display when removing "hovermode":"closest", but it's displaying all text based upon x and not the way I'd like it to which is "closest".

[http://codepen.io/bxlxux/pen/jALWAK](url)

etpinard:
Thanks for the report.

Here's a more compact reproducible example: http://codepen.io/etpinard/pen/ZOJQyk?

To note, the problem is also present on `master` and seems to be only affecting scattergeo traces with mode set to `lines`. Changing the issue title accordingly.

etpinard:
Oh wait. My mistake your example isn't a `scattergeo` graph.

The problem is in the cartesian hover code.

**Preliminary diagnostic:** Hovering of superimposed picks the last trace and honours the `hoverinfo` setting present in that last trace. For example, if the last trace has `hoverinfo: 'none'` then no hover labels are shown.

See https://github.com/plotly/plotly.js/pull/438, https://github.com/plotly/plotly.js/pull/495 and https://github.com/plotly/plotly.js/issues/678 for more info on the topic.

Thanks again for this report. In the meantime here's an easy workaround: http://codepen.io/etpinard/pen/qNXNwz 

where simply reordering the traces gets you back to the desired behaviour.

etpinard:
More generally,

 _Should hovering over super-imposed traces on `hovermode: 'closest'` show multiple hover labels_?

etpinard:
Ok. I found a way to fix this (see [`closest-hoverinfo-none`](https://github.com/plotly/plotly.js/compare/closest-hoverinfo-none)) but I'm not convinced it is the way to go.

At the moment, hovering with `hovermode: 'closest'` over super-imposed points picks out the most above point - regardless of its `hoverinfo` value. The same point is used to dispatch `plotly_click` and `plotly_hover` info - which _a priori_ should not care about the `hoverinfo` value.

[...]

PR #2270, opened 2018-01-19
Constraint-type contours for regular contour (not on a carpet)

alexcjohnson:
Extends `contours.type: 'constraint'` to regular `contour` traces, not just `contourcarpet`.

cc @bpostlethwaite - lets not merge until we've waited a bit to see if we need a patch release, but you can use this branch for integration.


src/traces/contour/attributes.js
```diff
@@ -155,6 +183,53 @@ module.exports = extendFlat({
...
+        value: {
+            valType: 'any',
+            dflt: 0,
+            role: 'info',
+            editType: 'calc',
+            description: [
+                'Sets the value or values by which to filter by.',
+
+                'Values are expected to be in the same type as the data linked',
+                'to *target*.',
```
etpinard: Hmm. I think this was copied from `transforms/filter.js`, `target` makes no sense here.

src/traces/contour/close_boundaries.js
```diff
@@ -32,36 +34,31 @@ module.exports = function(pathinfo, operation, perimeter, trace) {
...
+            if(v2 < boundaryMin || v1 > boundaryMax) {
+                pi0.prefixBoundary = true;
             }
             break;
         case '][':
-        case ')(':
-            v1 = Math.min.apply(null, trace.contours.value);
-            v2 = Math.max.apply(null, trace.contours.value);
+            v1 = Math.min.apply(null, contours.value);
+            v2 = Math.max.apply(null, contours.value);
```
etpinard: Shouldn't we use `Lib.aggNums` here in case `contours.value` has some non-numeric values?

src/traces/contour/calc.js
```diff
@@ -9,94 +9,14 @@
...
-            var sizeOut;
-            if(start === end) sizeOut = 1;
-            else sizeOut = autoContours(start, end, trace.ncontours).dtick;
-
-            inputContours.size = contours.size = sizeOut;
-        }
-    }
-
+    var cd = heatmapCalc(gd, trace);
+    setContours(trace);
```
etpinard: Lovely :recycle:

src/traces/contour/constraint_value_defaults.js
```diff
@@ -9,14 +9,13 @@
 
 'use strict';
 
-var constraintMapping = require('./constraint_mapping');
 var isNumeric = require('fast-isnumeric');
+var COMPARISON_OPS2 = require('../../constants/filter_ops').COMPARISON_OPS2;
```
etpinard: Looks like `constraint_value_defaults.js` is only required in `constraint_defaults.js`. Would you mind merging them in a single file?

src/traces/contour/hover.js
```diff
@@ -9,8 +9,26 @@
...
 
+var Color = require('../../components/color');
+
 var heatmapHoverPoints = require('../heatmap/hover');
 
 module.exports = function hoverPoints(pointData, xval, yval, hovermode, hoverLayer) {
-    return heatmapHoverPoints(pointData, xval, yval, hovermode, hoverLayer, true);
+    var hoverData = heatmapHoverPoints(pointData, xval, yval, hovermode, hoverLayer, true);
+
+    if(hoverData) {
```
etpinard: Looks like `contourcarpet` does not have a `hoverPoints` handler. Could it now reuse this one here?

src/constants/filter_ops.js
```diff
@@ -0,0 +1,36 @@
...
+* Copyright 2012-2018, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
```
etpinard: :heart:

src/traces/contourcarpet/defaults.js
```diff
@@ -13,17 +13,19 @@ var Lib = require('../../lib');
...
 var attributes = require('./attributes');
+var handleConstraintDefaults = require('../contour/constraint_defaults');
+var handleContoursDefaults = require('../contour/contours_defaults');
 var handleStyleDefaults = require('../contour/style_defaults');
-var handleFillColorDefaults = require('../scatter/fillcolor_defaults');
-var plotAttributes = require('../../plots/attributes');
-var supplyConstraintDefaults = require('./constraint_value_defaults');
-var addOpacity = require('../../components/color').addOpacity;
 
 module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
```
etpinard: Much cleaner. Thanks!

src/traces/contour/hover.js
```diff
@@ -9,8 +9,26 @@
...
+                    hoverPt.color = Color.addOpacity(trace.fillcolor, 1);
+                }
+                else if(trace.contours.showlines && Color.opacity(trace.line.color)) {
+                    hoverPt.color = Color.addOpacity(trace.line.color, 1);
+                }
+            }
+        });
+    }
+
+    return hoverData;
```
etpinard: Moverover, should we show hover labels for x/y/z points outside the constraints? I'm not sure what's best here. Maybe that's why Ricky didn't bother implementing hover labels for `contourcarpet` in the first place.

etpinard: One more comment: looks like contours would benefit from a version of `hovermode: 'compare'` that allows showing multiple labels of super-imposed traces as proposed in https://github.com/plotly/plotly.js/issues/720#issuecomment-230900670

Currently, `hovermode: 'closest'` isn't that useful on the `contour_constraints` mock:

peek 2018-01-29 18-35

src/traces/contour/hover.js
alexcjohnson: I'm guessing it will take some work to sort out hover for `contourcarpet`. Unlike `scattercarpet` where you just have to look at the final position of each point, `contourcarpet` we would actually have to invert (x,y) back to (a,b), which I don't think has been implemented (it's in principle multi-valued anyway).

src/traces/contour/close_boundaries.js
alexcjohnson: `handleConstraintValueDefaults` already cleans this up. Not sure if I actually *agree* with all the choices made there, but it doesn't look to me as though non-numeric values can get through.

src/traces/contour/hover.js
alexcjohnson: > Moverover, should we show hover labels for x/y/z points outside the constraints?

I think so. In fact it's possible that there actually *aren't* any data points (only interpolations) within the constraints, for `[]` type. Anyway I could see people using hover to see "how bad would it be to operate out here"

> One more comment: looks like contours would benefit from a version of `hovermode: 'compare'` that allows showing multiple labels of super-imposed traces as proposed in https://github.com/plotly/plotly.js/issues/720#issuecomment-230900670

Yes definitely - but perhaps we tackle that all at once as part of #720? There are going to be some weird cases with contours, like if two constraints overlay but with misaligned grids.

src/traces/contour/attributes.js
alexcjohnson: good catch - https://github.com/plotly/plotly.js/pull/2270/commits/7c7a01135b0319fa536d7520b6fc33f8275d27c5

src/traces/contour/constraint_value_defaults.js
alexcjohnson: ah yes, no need for separate files - https://github.com/plotly/plotly.js/pull/2270/commits/7e837efd67cdd56c5934fbf92d4f8fbbd545c305

src/traces/contour/hover.js
etpinard: > Yes definitely - but perhaps we tackle that all at once as part of #720?

Absolutely :ok_hand:

alexcjohnson:
> One additional comment: why did the airfoil baseline change exactly?

Because the streamlines and pressure contours weren't supposed to have legend entries at all, they're regular contours, not constraints. I suppose we could make it possible to give legend entries for contour plots with no colorscale, but that's not in scope here.

etpinard:
Looks great! Thanks for clarifying the `airfoil` diff. There's definitely room for making the hover behavior more configurable, but that can wait for (a) future PR(s).

:dancer:

PR #2277, opened 2018-01-22
Make sankey respect layout.colorway

### LINKED ISSUE #2276: Sankey plot doesn't use layout.colorway
The sankey plot doesn't use the colors from the new colorway layout option, example:
https://codepen.io/anon/pen/aEMvBX

As you can see, the bar chart at the bottom uses the red and green colours perfectly, but sankey doesn't.

alexcjohnson:
fixes #2276 

@monfera another quick one for your review. This slipped through during #2156

### DIFF
--- a/test/jasmine/tests/sankey_test.js
+++ b/test/jasmine/tests/sankey_test.js
@@ -12,6 +12,7 @@ var fail = require('../assets/fail_test');
 var mouseEvent = require('../assets/mouse_event');
 var assertHoverLabelStyle = require('../assets/custom_assertions').assertHoverLabelStyle;
 var supplyAllDefaults = require('../assets/supply_defaults');
+var defaultColors = require('@src/components/color/attributes').defaults;
 
 describe('sankey tests', function() {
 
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/sankey_test.js
```diff
@@ -31,7 +32,7 @@ describe('sankey tests', function() {
         var traceOut = { visible: true },
             defaultColor = '#444';
 
-        Sankey.supplyDefaults(traceIn, traceOut, defaultColor, layout);
+        Sankey.supplyDefaults(traceIn, traceOut, defaultColor, Lib.extendFlat({colorway: defaultColors}, layout));
```
monfera: Might be an overly cautious idea, how about using non-default colors for `colorway`here, or just altering a `sankey` mock to lock in a visible difference.

monfera:
💃 thanks Alex, just tried such a thing locally as well, looks great! 
image

test/jasmine/tests/sankey_test.js
alexcjohnson: Good call. Actually, given that `('color/attributes').defaults` no longer appears explicitly in the sankey code, seems to me that an end-to-end test of the full default-color case may be the most useful of all, so I just stripped explicit colors out of the dark mock -> https://github.com/plotly/plotly.js/pull/2277/commits/f2ebc7908d2abf26abdfd02e1eeb882310dcd707

monfera: Yes, makes sense, and works great with a custom `colorway` too!

PR #2278, opened 2018-01-22
Intermittent errors fix

dfcreative:
@alexcjohnson this may (partially) fix intermittent test failures.

The reason seems to be too many webgl contexts and/or losing `gd` or alike.
Splitting `gl_plot_interact` by `gl2d_plot_interact` and `gl3d_plot_interact` seems to helped with some tests.
I would suggest squashing merge this, since there are debugging commits.

PR #2279, opened 2018-01-23
Fixedrange click

### LINKED ISSUE #2265: plotly_click doesn't fire when `fixedrange` is true
Hello

When `fixedrange=true` for xaxis and yaxis
`plotly_click` events do not trigger.

Please see
https://codepen.io/sporto/pen/WdLvro?editors=0010

Thanks

alexcjohnson:
fixes #2265 

In #2241 we consolidated click behavior into the drag handler, for cartesian subplots. But this inadvertently disabled click when the drag handler was not attached - ie when there's nothing to drag because you're in zoom or pan mode and all the axes are `fixedrange`!

@dfcreative please take a look!

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -113,12 +113,14 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
     var dragger = makeRectDragger(plotinfo, ns + ew + 'drag', cursor, x, y, w, h);
 
+    var allFixedRanges = !yActive && !xActive;
+
     // still need to make the element if the axes are disabled
     // but nuke its events (except for maindrag which needs them for hover)
     // and stop there
-    if(!yActive && !xActive && !isSelectOrLasso(fullLayout.dragmode)) {
+    if(allFixedRanges && !isMainDrag) {
         dragger.onmousedown = null;
-        dragger.style.pointerEvents = isMainDrag ? 'all' : 'none';
+        dragger.style.pointerEvents = 'none';
         return dragger;
     }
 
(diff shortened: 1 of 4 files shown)


alexcjohnson:
@dfcreative 5th time's a charm 🙃  Lets get #2278 or some squashed version merged ASAP and 🙏  our tests get more reliable!

src/plots/cartesian/dragbox.js
```diff
@@ -113,12 +113,14 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
+    var allFixedRanges = !yActive && !xActive;
+
     // still need to make the element if the axes are disabled
     // but nuke its events (except for maindrag which needs them for hover)
     // and stop there
-    if(!yActive && !xActive && !isSelectOrLasso(fullLayout.dragmode)) {
+    if(allFixedRanges && !isMainDrag) {
         dragger.onmousedown = null;
-        dragger.style.pointerEvents = isMainDrag ? 'all' : 'none';
+        dragger.style.pointerEvents = 'none';
```
dfcreative: That is called just once per `dragger` element, right? Ie. it does not need to be reset to pointerEvents `'all'`.

dfcreative:
@alexcjohnson looks good, although not sure I can review the whole logic.
Tests are merged, should pass fine now.

src/plots/cartesian/dragbox.js
alexcjohnson: Good thinking - I had to check, but yes, if you change it later so this needs `pointerEvents`, it will be reinstated by [`dragElement.init`](https://github.com/plotly/plotly.js/blob/master/src/components/dragelement/index.js#L103)

PR #2280, opened 2018-01-23
fix ghost fill-tonext when either trace is emptied out

### LINKED ISSUE #2271: Restyling x or y values with null values creates a ghost when filling is enabled
If you restyle the x or y values of a trace that forms part of a filled area with an array of `null` values, you get a ghost in the shape of the original filled area. It shows up at any zoom level and regardless of how you pan.

[CodePen](https://codepen.io/ChristofferHS/pen/PEXgLv)

alexcjohnson:
Fixes #2271 - we were short-circuiting a little too much behavior when no line is generated by a trace.

@dfcreative another quick one to look at, if you would!

### DIFF
--- a/src/traces/scatter/link_traces.js
+++ b/src/traces/scatter/link_traces.js
@@ -9,12 +9,11 @@
 'use strict';
 
 module.exports = function linkTraces(gd, plotinfo, cdscatter) {
-    var cd, trace;
+    var trace, i;
     var prevtrace = null;
 
-    for(var i = 0; i < cdscatter.length; ++i) {
-        cd = cdscatter[i];
-        trace = cd[0].trace;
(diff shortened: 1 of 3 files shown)


dfcreative:
Turns out this test case does not work with `scattergl`, although it should, I should have a look.
Also is that correct that the legend shows an empty trace?
image
In all looks good!

alexcjohnson:
>  is that correct that the legend shows an empty trace?

Yes, trace 1 has `mode: 'lines', line: {width: 0}`, rendering it invisible (could have just done `mode: 'none'` like trace 2 but without the `fill`, but this works too)

dfcreative:
:dancer: :)

Issue #2282, opened 2018-01-23
relayout of height/width with scattergl does not update positioning correctly


cpsievert:
https://codepen.io/cpsievert/pen/rpbKEz

Originally identified in https://github.com/ropensci/plotly/pull/1178#issuecomment-359141174

bpostlethwaite:
@dfcreative fyi

dfcreative:
Should be fixed at #2283

PR #2283, opened 2018-01-23
Layout resize fix

dfcreative:
Addresses #2282 and https://github.com/ropensci/plotly/pull/1178#issuecomment-359141174.

@alexcjohnson please take a glance.


alexcjohnson:
good catch 💃

PR #2285, opened 2018-01-23
fix the title side for odd cheaterslope situations

### LINKED ISSUE #2254: carpet plot a/b axis title positioning bug
In certain configurations - looks like left-side labels with the title aligned either exactly vertical or clockwise of vertical, either the a axis or the b axis - the title gets positioned over top of the carpet rather than outside it.
screen shot 2018-01-16 at 10 31 59 am
This is the `cheater` mock, restyled to `cheaterslope = -2`

alexcjohnson:
Fixes #2254 

The mock I added here really highlights the fact that we don't explicitly include labels/titles in autoranging the containing axes for carpets... but that's outside the scope of this fix. Actually, this mock isn't even autoranged, or you wouldn't be able to see some of the titles at all! Mostly though this is only a problem for tiny carpets as used in these mocks, most real-world examples are just fine.

Anyway, before the fix this mock would have looked kind of like:

cheaterslope before fix

Ie some of the titles are just completely in the wrong locations.

@bpostlethwaite please review - I would like to get this fixed before we make a slider for `cheaterslope` that makes it super easy to access these conditions 😅


alexcjohnson:
ah shoot, the failing `airfoil` mock pointed out that this doesn't work with `showticklabels: 'none'` - hold on, a little more work to do...

alexcjohnson:
@bpostlethwaite ready for review

bpostlethwaite:
Good catch - :dancer: 
Is there an issue for implementing the full autoranging behaviour for carpet axes titles/labels?

alexcjohnson:
> Is there an issue for implementing the full autoranging behaviour for carpet axes titles/labels?

Good point, no, but now there is -> https://github.com/plotly/plotly.js/issues/2289

Issue #975, opened 2016-09-23
histogram2d / heatmap with z: separate color for empty bins


ajrouvoet:
2d histograms with `histfunc = avg|sum` are useful for plotting measurements on irregular 2D grids. 
The downside of this approach is that the representation of zero measurements collapses with e.g. n times a measurement of zero.

For these applications it would be useful if there was some way of either treating empty bins separately, or having the ability of setting a custom histfunc.

etpinard:
That would be really nice to allow for a special _no-data_ color - even heatmap traces could benefit from it.

In the meantime, here's a workaround: http://codepen.io/etpinard/pen/xVoJoj?

ElessarWebb:
If I were to implement this feature and submit a PR (assuming it wouldn't be too hard), what would be the timeline for it to end up in a release?

etpinard:
> If I were to implement this feature and submit a PR [...] timeline for it to end up in a release

The next minor release ([`v1.18.0`](https://github.com/plotly/plotly.js/milestone/11)) should be out in about 7 to 10 days. We usually make minor release (i.e. release that add features) every two weeks or so.

>  (assuming it wouldn't be too hard)

Adding this feature will be similar to and have to touch the same files as PR https://github.com/plotly/plotly.js/pull/868

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2288, opened 2018-01-24
Fix coloring and hover data for nonuniform heatmap & contour bricks

### LINKED ISSUE #2233: Invalid heatmap values
This heatmap shows that my values are (15, 22.75), (22.75, 28), (26, 11), the first two are wrong.

https://jsfiddle.net/zc5r6v6t/3/

plotly

alexcjohnson:
Fixes #2233 - we were finding brick edges (as midpoints between data values) and then forever after using the midpoints of bricks as the hover data and (heatmap with `zsmooth: 'best'` or contour with `coloring: 'heatmap'`) the point of exact color matching, rather than using the original data values for these purposes.

@dfcreative one more bugfix PR to review before I move onto building new stuff?


src/traces/heatmap/plot.js
```diff
@@ -415,3 +383,64 @@ function plotOne(gd, plotinfo, cd) {
...
+        bin0: bin0,
+        frac: frac,
+        bin1: Math.round(bin0 + frac / (interp - bin0))
+    };
+}
+
+function findInterpFromCenters(pixel, centerPixArray) {
+    // if(pixel <= centerPixArray[0]) return {bin0: 0, bin1: 0, frac: 0};
+    var maxBin = centerPixArray.length - 1;
+    // if(pixel >= centerPixArray[lastCenter]) return {bin0: lastCenter, bin1: lastCenter, frac: 0};
```
dfcreative: Are these being temporarily commented out?

dfcreative:
@alexcjohnson is that ok that we see `NaN` in empty blocks?

image

src/traces/heatmap/plot.js
alexcjohnson: Ah thanks - those are obsolete -> https://github.com/plotly/plotly.js/pull/2288/commits/3581c87fbe9b151162dc854992a963aa3599702d

alexcjohnson:
> is that ok that we see `NaN` in empty blocks?

I think so - it's certainly better than saying `0`. Might be nice to be able to customize what NaN displays as, but this is a reasonable default I think. See also #975

alexcjohnson:
> Not relevant to this PR, but is that fine that hover labels appear at different positions than with `contour`

I believe this behaves as expected. For a heatmap, the hover label should not point to the data value but to the edge of the brick, whereas contour map labels should point exactly to the data value.

Heatmaps you're defining, either implicitly or explicitly, bricks of a certain value, and you're saying "this is the value that applies to this brick". That metaphor gets a little bit muddied by smoothing the heatmap, but that's the idea anyway. Whereas with contour maps, you're implying that there's some continuous function and you've only sampled it at certain locations, the data points, but it has a distinct value at every (x,y) pair. But we don't extrapolate beyond the range of the data, even as far as the edge of the corresponding heatmap brick.

dfcreative:
Looks good then! :dancer:

PR #2293, opened 2018-01-25
Unlock gl-shader

dfcreative:
Resolves https://github.com/gl-vis/gl-mesh3d/pull/12, https://github.com/gl-vis/gl-error3d/pull/4, https://github.com/gl-vis/gl-axes3d/pull/11 and yarn bundling mentioned by @nicolaskruchten


dfcreative:
@alexcjohnson could you please have a look :)

dfcreative:
@nicolaskruchten can you please try bundling w/yarn the `git@github.com:plotly/plotly.js.git#7b1bb6a5aa9fae0d4d05420f7ae81798f4fc5ef8` dist? That should fix the issue.

package.json
```diff
@@ -69,18 +69,19 @@
     "es6-promise": "^3.0.2",
     "fast-isnumeric": "^1.1.1",
     "font-atlas-sdf": "^1.3.3",
+    "gl-axes3d": "^1.2.6",
```
alexcjohnson: why are we adding a package but making no direct reference to it?

dfcreative: @alexcjohnson good point. I wonder if we should remove `gl-shader` dependency as well, since we're not using it by plotly directly.

alexcjohnson: Sounds reasonable to me, but after the mess we had yesterday with `npm ls` I'd like to figure this out with @etpinard 's input.

alexcjohnson:
💃  from my side, but I'd like to hear from @nicolaskruchten whether this accomplished the goal.

nicolaskruchten:
💃 from me, I tested all permutations of installer=(yarn, npm) and bundler=(browserify, webpack) and the WebGL bug doesn't appear :)

etpinard:
This PR seem to have made the tests fail on `master` 

https://circleci.com/gh/plotly/plotly.js/6120

etpinard:
Re-running the tests on `master` w/o cache (i.e. with a brand new `node_modules/` folder) helps a little bit:

https://circleci.com/gh/plotly/plotly.js/6129


Only 3 `surface` mocks are failing:

image

etpinard:
After re-running the test w/o cache off this PR's last commit https://github.com/plotly/plotly.js/pull/2293/commits/0d1d19be1cece1a2ef3342b62997cd9f87ca28cb, the same `surface` mocks are failing. 

https://circleci.com/gh/plotly/plotly.js/6130?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This is good news. At least we have some sort of consistency.

dfcreative:
@etpinard added baselines https://github.com/plotly/plotly.js/pull/2306/commits/8456f9584b8d19a66252841cb587d4286013fb3a, seems there is very subtle shading or dithering difference, as if color conversion is a bit wrong. I'll try to reproduce that.

dfcreative:
@etpinard tbh that is pretty mystical difference. That is not even clear if the behavior before was correct. That seems like internal WebGL thing − as if order of binding affects attribute/uniform precision [here](https://github.com/gl-vis/gl-surface3d/blob/master/surface.js#L349) or in a place like that.

PR #2305, opened 2018-01-29
 Implement notched box plots, closes #2286 

### LINKED ISSUE #2286: Feature request: notched box plots
I searched thoroughly for such option, but found only this unanswered post on plotly community website: [Add option notch to box plots](https://community.plot.ly/t/add-option-notch-to-box-plots/7396).

Here is some short introduction: https://sites.google.com/site/davidsstatistics/home/notched-box-plots
Here is a reference for notches from MATLAB: https://www.mathworks.com/help/symbolic/mupad_ref/notched.html,
And here is a question from stackexchange to highlight how notches should be sometimes rendered "inside-out": https://stats.stackexchange.com/questions/151580/how-to-interpret-notched-box-plots

Is such an option in-scope of plotly?

krassowski:
I added a boolean option "notched" (default false) and a numerical "notchwidth" (default 0.25).
The options define if the box trace should be notched or not and if yes, determine the width of the notch (proportional to the width of the box).

I modified the svg path-generation code, utilizing conditional (ternary) operators so the common parts of the box (top & bottom edges for vertical, left & right edges for horizontal) are shared.

I added two test baselines: `box_notched` and `box_horz_notched`.

Preliminary review was given by @alexcjohnson in my fork: https://github.com/krassowski/plotly.js/pull/1. 

Update: notch width behavior was changed to follow Matlab convention.
The previous version of this PR used the ggplot2 way.

Please let me know what you think.

box notched

box horiz notched


src/traces/box/defaults.js
```diff
@@ -29,6 +29,9 @@ function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
     coerce('whiskerwidth');
     coerce('boxmean');
 
+    var notched = coerce('notched');
+    if(notched) coerce('notchwidth');
```
etpinard: I wonder if we can get away with just a `notchwidth` similar to how whiskers are set (with `whiskerwidth`)?

krassowski: Are you suggesting to remove `if(notched)`?
Btw. this line was modified after this comment: https://github.com/krassowski/plotly.js/pull/1#discussion_r164287086

etpinard: > Are you suggesting to remove if(notched)?

Yes, that's what I'm suggesting. Waiting on a concensus.

alexcjohnson: I think @etpinard was suggesting removing the `notched` attribute altogether and just defaulting to `notchwidth=0` as no notch. I thought about this but would prefer the existing way so that we can pick a nice default `notchwidth` rather than everyone making up random values. One thing we *could* do though is make the default for notched be `true` if `traceIn.notchwidth !== undefined`. That way if you do want a custom width you don’t also need to provide `notched`.

etpinard: >  was suggesting removing the notched attribute altogether and just defaulting to notchwidth=0 as no notch

yes, this is what I meant. Sorry for the confusion.

etpinard: > I thought about this but would prefer the existing way so that we can pick a nice default notchwidth rather than everyone making up random values

Good point here. I wonder if we should do the same for whiskers in v2? That is, adding a `showwhiskers` boolean and making the `whiskerswidth` default _smarter_?

alexcjohnson: > I wonder if we should do the same for whiskers in v2?

My argument doesn't really apply to whiskers - though I'm not sure I get what you mean by *smarter* default. If we're happy with the default `whiskerwidth` staying at `0.5` then people can use `0` (a very non-random number 😄 ) to disable whiskers and any other number for a custom width.

alexcjohnson: @krassowski in talking with @etpinard offline, we agreed that despite the correspondence with `whiskerwidth` the Matlab convention makes more sense, to us anyway - so that's 0=no notch, 0.5=meet in the middle, with a default of 0.25.

And re: my comment:
> One thing we *could* do though is make the default for notched be `true` if `traceIn.notchwidth !== undefined`.

That's just removing `attributes.notched.dflt` and changing its coerce line to:
`var notched = coerce('notched', traceIn.notchwidth !== undefined);`

You'll also want to merge in the latest master so tests pass, then I think we're ready to go!

src/traces/box/plot.js
```diff
@@ -101,6 +101,8 @@ function plotBoxAndWhiskers(sel, axes, trace, t) {
     var wdPos = t.wdPos || 0;
     var bPosPxOffset = t.bPosPxOffset || 0;
     var whiskerWidth = trace.whiskerwidth || 0;
+    var notched = trace.notched || false;
+    var nw = notched ? 1 - trace.notchwidth : 1;
```
alexcjohnson: I think we're off by a factor of 2 here - `1 - 2 * trace.notchwidth`?

src/traces/box/attributes.js
```diff
@@ -85,6 +85,27 @@ module.exports = {
...
+        role: 'style',
+        editType: 'calcIfAutorange',
+        description: [
+            'Determines whether or not notches should be drawn.'
+        ].join(' ')
+    },
+    notchwidth: {
+        valType: 'number',
+        min: 0,
+        max: 1,
```
alexcjohnson: also in the matlab convention the max is `0.5`

krassowski:
It seems that test "select box and lasso per trace" fails. I have no idea why.

alexcjohnson:
> It seems that test "select box and lasso per trace" fails. I have no idea why.

Seems to be a bad day for that test. You didn't break it, it's just glitchy.

💃 - once tests pass I'll merge this. Thanks for your work @krassowski!

test/image/mocks/box_horz_notched.json
```diff
@@ -0,0 +1,290 @@
...
+            68
+         ],
+         "line":{
+            "color":"#1c9099"
+         },
+         "type":"box",
+         "name":"Notch width = 0.1",
+         "notched":true,
+         "orientation": "h",
+         "notchwidth":0.1
```
alexcjohnson: Actually looks like the mocks and baseline images got out of sync. This one uses 0.2 in both images.
https://6171-45646037-gh.circle-artifacts.com/0/home/ubuntu/plotly.js/build/test_images_diff/diff-box_notched.png
https://6171-45646037-gh.circle-artifacts.com/0/home/ubuntu/plotly.js/build/test_images_diff/diff-box_horz_notched.png

krassowski: You are right. I updated the baselines, hope that all will work this time.

PR #2306, opened 2018-01-29
Unbind VAO in gl-surface3d

dfcreative:
This should fix gl3d-surface tests https://github.com/plotly/plotly.js/pull/2293#issuecomment-361330182


dfcreative:
@etpinard not sure what was the tiny shading difference caused by, seems like some internal webgl thing. Not even sure what image is more correct - before or after, because code after seems "more right". As if shading quality is different, or dithering enabled, or some floats got extra-precision or something alike.

etpinard:
This looks right!

Thanks for the quick fix :dancer: 

cc @alexcjohnson and anyone else that opend a PR since #2293

etpinard:
Oh and re https://github.com/plotly/plotly.js/pull/2293#discussion_r163965601, yes we can remove `gl-shader` from the plotly.js package.json as now every gl-vis module we use is compatible with the latest gl-shader version.

Issue #1798, opened 2017-06-19
Make CircleCI use node@8 and npm @5


etpinard:
Both LTS, now available on CircleCI - see https://circleci.com/blog/npm5-and-nodejs8/

etpinard:
Actually, node@8 (with npm@5) isn't available on CircleCI's preinstalled images yet. See

https://circleci.com/docs/1.0/build-image-trusty/#nodejs

for most up-to-date list.

etpinard:
While doing this, it might be worth migrating to CircleCI 2.0: 

https://circleci.com/docs/2.0/migrating-from-1-2/

etpinard:
cc https://github.com/plotly/streambed/pull/9973

bpostlethwaite:
fyi https://github.com/tessel/node-usb/issues/193#issuecomment-323506658
and I'm seeing these errors building plotly.js 
```
 npm install
npm ERR! code 128
npm ERR! Command failed: /usr/bin/git clone -q git://github.com/mapbox/mapbox-gl-shaders.git /root/.npm/_cacache/tmp/git-clone-d8de79a2
npm ERR! fatal: could not create leading directories of '/root/.npm/_cacache/tmp/git-clone-d8de79a2': Permission denied
npm ERR! 

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2017-08-29T20_24_01_459Z-debug.log
root@stage-streambed-fl6p:/var/www/streambed# rm -r /root/.npm
root@stage-streambed-fl6p:/var/www/streambed# npm install
npm ERR! code 128
npm ERR! Command failed: /usr/bin/git clone -q git://github.com/mapbox/mapbox-gl-shaders.git /root/.npm/_cacache/tmp/git-clone-146d42f8
npm ERR! fatal: could not create leading directories of '/root/.npm/_cacache/tmp/git-clone-146d42f8': Permission denied
npm ERR! 

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2017-08-29T20_24_15_797Z-debug.log
```

bpostlethwaite:
sorry comment link here: https://github.com/npm/npm/issues/16898

etpinard:
https://github.com/plotly/plotly.js/issues/1745#issuecomment-307132219

Bumping to `npm@5.0.3` should solve this.

bpostlethwaite:
```
[...]

PR #2258, opened 2018-01-17
Regl-based scattergl and scatterpolargl traces

etpinard:
This PR is a big one :muscle: :rocket: :rotating_light: 

About 99.99% of it is the fruit of @dfcreative's 6-month worth of work. @dfcreative rewrote our `scattergl` trace type using `regl`. In brief, `scattergl` traces are now drawn in a single canvas -- no matter how many subplots there are on the graphs -- that sits above cartesian SVG axes. Reusing the SVG axes logic fixed about 10 bugs by itself. Plotting in a single canvas improves performance for multi-subplot graphs immensely as well as removing those annoying _too many active WebGL contexts_ warnings. For more info on our new regl-based `scattergl` trace type, see https://github.com/plotly/plotly.js/pull/1869. For the sake of brevity, all that work was squashed into a single commit b64928d

The remaining 0.01% of this PR adds a new trace type: `scatterpolargl` which, as the name suggests, is a scatter trace type that'a plotted on polar subplots using WebGL. This new trace type doesn't bring much new code into the library, it simply combines logic from `scattergl` and `scatterpolar` on top of polar (SVG) subplots from #2200.

PR #2310, opened 2018-01-30
Test with CircleCI 2.0

etpinard:
CircleCI has mostly stopped adding new images compatible with its 1.0 framework, so it's about time we pay back some tech-debt :pick: :moneybag:  In addition, upgrading to 2.0, makes our tests run in about 10 minutes - roughly 5 minutes less than for the current master. :tada:

Note that our jasmine tests now run on Chrome 59 (no more Chrome 54 as on the current master). We're using node 6.x + npm 3.x images (same as the current master). An upcoming PR (cc https://github.com/plotly/plotly.js/issues/1798) will bump those to node 8.x + npm 5.x - which should speed up our tests even more :racing_car: 

The only _tricky_ part here was to make the image tests work. CircleCI 2.0 is setup to fetch a single docker container and run tests directly in that container. In our CircleCI 1.0 config, we fetched our `imagetest` docker container and made requests from a "main" container to a port mapped to the nw.js server inside `imagetest`. Circle 2.0 doesn't allow communication between containers, so I had to update some of our `tasks/` script, mostly by removing a few hacky lines  :stuck_out_tongue:  

About our increasingly-annoying intermittent test failures, I still noticed some unfortunately. But hopefully this is a step in the right direction.

cc @alexcjohnson @dfcreative @scjody (who's been asking me to do this for a while)


tasks/noci_test.sh
```diff
@@ -10,7 +10,4 @@ npm run test-jasmine -- --tags=noCI --nowatch || EXIT_STATE=$?
 # mapbox image tests take too much resources on CI
 npm run test-image -- mapbox_* --queue || EXIT_STATE=$?
 
-# run gl2d image test again (some mocks are skipped on CI)
-npm run test-image-gl2d || EXIT_STATE=$?
```
etpinard: fixup for https://github.com/plotly/plotly.js/pull/2258/commits/9f304e658bfd8cecd6a5a011faa46e6161deaa58 

No need for this line anymore as now all gl2d mocks are tested on CI.

package.json
```diff
@@ -44,7 +44,7 @@
     "start-image_viewer": "node devtools/image_viewer/server.js",
     "start": "npm run start-test_dashboard",
     "baseline": "node tasks/baseline.js",
-    "preversion": "npm-link-check && npm dedupe",
+    "preversion": "npm-link-check && npm dedupe && npm ls --prod",
```
etpinard: cc @alexcjohnson @dfcreative 

This commit partly reverts https://github.com/plotly/plotly.js/commit/1d3a3f0ead64168f1764ad7fe215936964bd116d

`npm ls --prod` will fail only if a bundle dependency (i.e. ignoring dev-dependencies) is (1) missing, (2) needs to be updated or (3) is extraneous. This step ensures that we're not bundling `dist/` bundles with incorrect dependencies.

.circleci/test.sh
```diff
@@ -1,21 +1,29 @@
 #!/bin/bash
 
+# override CircleCi's default run settings
+set +e
+set +o pipefail
```
etpinard: I don't like CircleCI's new defaults where a given job stop executing after the first test failure. I prefer running all the tests no matter what, to see if a given commit makes multiple tests fail. I hope you agree with me.

alexcjohnson:
LGTM! Only thing that occurs to me is we could balance the containers a bit more, perhaps taking the bundle tests out of jasmine and lumping with syntax, that would save another minute or two. But not a big deal, this is already a big win! 💃

etpinard:
Great. Merging. Next stop updating to node 8.x. I'll should have time to do that by week's end.

Issue #2299, opened 2018-01-25
restyle of marker opacity not working for scattergl 


cpsievert:
https://codepen.io/cpsievert/pen/Rxzzzz

cc @dfcreative

alexcjohnson:
Interestingly `Plotly.restyle(gd, {"marker.opacity": 0.1});` seems to work fine, as does the `trace.opacity` attribute on the initial draw - it's just `Plotly.restyle(gd, {"opacity": 0.1});` that fails.

@dfcreative you probably just have to explicitly define `opacity` in `scattergl/attributes`, like we do in eg [`surface/attributes`](https://github.com/plotly/plotly.js/blob/f48d03ec5da0219419faac55366646479931d65d/src/traces/surface/attributes.js#L226) so you can get `editType: 'calc'` instead of `editType: 'style'` that you get from the base [opacity attribute](https://github.com/plotly/plotly.js/blob/f48d03ec5da0219419faac55366646479931d65d/src/plots/attributes.js#L55). And assuming that fixes it, take a quick spin through the rest of those base attributes and see if anything else is broken the same way.

etpinard:
Better yet, maybe we could make [`ScatterGl.style`](https://github.com/plotly/plotly.js/blob/fdf20cb46bd9224ee70be4a3a78b39d6046a6142/src/traces/scattergl/index.js#L1214-L1221) handle trace opacity updates.

dfcreative:
@alexcjohnson thank, fixed in https://github.com/plotly/plotly.js/pull/2311/commits/a9d387aa14d3e2ade6768bcd2bbc20dcc22782b3. Making `ScatterGl.style`to do options recalculation (that is required to update opacity, since resulting opacity depends on trace.opacity, marker.opacity and color alpha) would be unnecessary complexity and mostly duplication.

etpinard:
fixed in https://github.com/plotly/plotly.js/pull/2311

Issue #2298, opened 2018-01-25
addTraces upon plotly_selected event with scattergl not rendering correctly


cpsievert:
https://codepen.io/cpsievert/pen/MrMMvj

Brush the blue marker. A red marker should appear at (1, 2), but it seems to be transparent.

cc @dfcreative

dfcreative:
@etpinard `scatter` mode adds a new trace without `selectedpoints` set on it. Ie. new trace looks like as if it's points are selected, but in fact it has no selection. Is that desired behavior? Or we better off showing new traces with all poins in `unselected` mode?

etpinard:
What does the svg `scatter` implementation do in this case?

dfcreative:
Should be fixed with #2311

cpsievert:
I modified the example slightly so the red marker should be placed directly on top of the blue marker, but again, the red marker isn't rendered.

May seem like a silly bug, but I need this fixed before I can update the R package...

dfcreative:
@cpsievert https://github.com/plotly/plotly.js/pull/2367 fixes that.
It swaps canvas order.
The initial guess was that selected points are rendered as the top layer over the not selected ones. That may break cases where selected points overlap not selected ones.
Ultimately we have to choose one way or another. Selected points are a separate layer, we cannot mix them into main points layer to make per-point layering, since that will significantly reduce performance.

etpinard:
I prefer the behavior on `master` with the selected points on top.

@cpsievert here's one way to ensure that the red marker (from the new trace added by `addTraces`) appears above the blue marker: https://codepen.io/etpinard/pen/VQMOOd?editors=0010

cpsievert:
@etpinard populating `selectedpoints` in that way will require some non-trivial changes on my end -- I imagine it should only be necessary for `scattergl`, though?

etpinard:
> I imagine it should only be necessary for scattergl, though?

yes.

chriddyp:
Looks like #2433 and https://community.plot.ly/t/webgl-context-lost-after-updating-10-times/8581/8?u=chriddyp is actually this issue. In Dash, we were recommending that users replot their data after selection to customize selection styles. 

This no longer works since the second trace (with _identical_ points as the first trace) now appears below the first trace that was originally plotted and selected. Here is a simple example: https://codepen.io/chriddyp/pen/YebmJL?editors=1010. Selecting the point used to display the second trace with the orange point on it, now it persists the first trace on top of it.

[...]

PR #2311, opened 2018-01-30
Refactor scattergl selection

dfcreative:
This enhances scattergl selection handling, fixing #2298.

* [x] test
* [x] #2299 
* [x] #2315


src/plots/cartesian/select.js
```diff
@@ -278,7 +278,7 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
             else {
                 // TODO: remove in v2 - this was probably never intended to work as it does,
                 // but in case anyone depends on it we don't want to break it now.
-                gd.emit('plotly_selected', undefined);
+                gd.emit('plotly_selected', {points: [], range: null});
```
dfcreative: That gets triggered whenever we click on a point in select/lasso mode. The event receives `undefined` data object, which makes user check it manually. Simple compatibility is better.

src/traces/scattergl/index.js
```diff
@@ -886,17 +883,77 @@ function plot(container, subplot, cdata) {
...
         }
-    }
 
-    // make sure selection layer is initialized if we require selection
-    var dragmode = layout.dragmode;
 
-    if(dragmode === 'lasso' || dragmode === 'select') {
-        if(scene.select2d && scene.selectBatch) {
-            scene.scatter2d.update(scene.unselectedOptions);
+        // update selection
```
dfcreative: Separated code chunks are in single location now

src/plots/cartesian/select.js
etpinard: Sure, simple compatibility is better. That's why we'll do this in v2. Until then, we must preserve backward compatibility. Please undo this change.

src/traces/scattergl/attributes.js
```diff
@@ -82,6 +83,10 @@ var attrs = module.exports = overrideAll({
         marker: scatterAttrs.unselected.marker
     },
 
+    opacity: extendFlat({}, plotAttrs.opacity, {
+        editType: 'calc'
+    }),
```
etpinard: Looks good. Thanks!

I think the best way to test this would be to spy on `ScatterGl.calc` checking that `Plotly.relayout(gd, 'opacity', /**/)` does invoke it - similar to this [suite](https://github.com/plotly/plotly.js/blob/18dac1f4210378945baacbffdacc06f883fc55a7/test/jasmine/tests/plot_api_test.js#L569).

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -645,4 +646,28 @@ describe('Test gl2d plots', function() {
...
+        }];
+
+        Plotly.plot(gd, dat, {width: 500, height: 500})
+        .then(function() {
+            expect(ScatterGl.calc).toHaveBeenCalledTimes(1);
+
+            return Plotly.restyle(gd, {'opacity': 0.1});
+        })
+        .then(function() {
+            expect(ScatterGl.calc).toHaveBeenCalledTimes(2);
```
etpinard: Brilliant test :sparkles:

src/traces/scattergl/index.js
etpinard: On the `gl2d_12` mock, calling `Plotly.restyle(gd, 'selected.marker.color', 'blue')` leads to:

image

etpinard:
Adding a test for https://github.com/plotly/plotly.js/issues/2298 will be tricky. @dfcreative is their a way to assert that some internal field (I guessing in `gd._fullLayout._plots.xy._scene`) update properly after `appendTraces`?

dfcreative:
@etpinard it is possible to test renderer options, since they directly affect `regl-scatter2d` state, thanks for the hint! Otherwise I'd just `readPixels`, which is bulletproof solution :)

src/traces/scattergl/index.js
dfcreative: Should be fixed with 98cddf3

etpinard:
Wow. Nice job @dfcreative this is looking very solid.

:dancer: (cc @alexcjohnson )

etpinard:
... @dfcreative make sure to close all the related issues after you merge this.

PR #2313, opened 2018-01-31
Minor changes to animation code ahead of the array animation fix

monfera:
These should involve no actual logic change, see the commit texts.


src/plot_api/plot_api.js
```diff
@@ -2634,20 +2635,20 @@ Plotly.addFrames = function(gd, frameList, indices) {
...
-            Lib.warn('addFrames: overwriting frame "' + _hash[name].name +
+            Lib.warn('addFrames: overwriting frame "' + _frameHash[name].name +
                 '" with a frame whose name of type "number" also equates to "' +
                 name + '". This is valid but may potentially lead to unexpected ' +
                 'behavior since all plotly.js frame names are stored internally ' +
                 'as strings.');
 
-            if(numericNameWarningCount > 5) {
-                Lib.warn('addFrames: This API call has yielded too many warnings. ' +
+            if(numericNameWarningCount === numericNameWarningCountLimit) {
```
etpinard: Why `===` instead of `>` ?

src/plots/plots.js
```diff
@@ -2252,14 +2252,6 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
         var module = contFull._module;
 
         if(!module) continue;
-
-        if(!module.animatable) {
-            var thisUpdate = {};
-
-            for(var ai in data[i]) {
-                thisUpdate[ai] = [data[i][ai]];
```
etpinard: Strange. I wonder what Ricky had in mind here. Good catch :ok_hand:

src/plot_api/plot_api.js
monfera: Explanation of 1st commit: it looked to me as though `Lib.warn` continued to be invoked past the 5 warnings, because I didn't see an actual condition that'd bypass this code path, so I added an extra condition in the `if` line. With the commit, if `numericNameWarningCount >= numericNameWarningCountLimit` then it doesn't descend in the `if`, so it won't warn there anymore. The condition for raising the "Too many warnings" is an equality check, because `numericNameWarningCount` is initially 0 but gets incremented past the condition check ie. before this check. If `numericNameWarningCount` is 5 here, then we've been through this conditional branch 5 times (with `numericNameWarningCount` values 1, 2, 3, 4, 5); we emit the "too many warning" things and won't come into this conditional branch again. Hope I got it right but maybe you're seeing an off-by-one error?

etpinard: Ok. That make sense. Can we add a test for this? Using jasmine `spyOn(Lib, 'warn')` should make this task relatively easy.

src/plots/plots.js
monfera: Could've been just some initial logic that remained there.

test/jasmine/tests/frame_api_test.js
```diff
@@ -37,6 +37,44 @@ describe('Test frame api', function() {
...
+                    return Plotly.addFrames(gd, [{name: 8, data: [{x: [3, 2, 1]}]}]);
+                })
+                .then(function() {return Plotly.addFrames(gd, [{name: 8, data: [{x: [3, 2, 1]}]}]);})
+                .then(function() {return Plotly.addFrames(gd, [{name: 8, data: [{x: [3, 2, 1]}]}]);})
+                .then(function() {return Plotly.addFrames(gd, [{name: 8, data: [{x: [3, 2, 1]}]}]);})
+                .then(function() {return Plotly.addFrames(gd, [{name: 8, data: [{x: [3, 2, 1]}]}]);})
+                .then(function() {
+                    // Five (`var numericNameWarningCountLimit = 5`) warnings and one warning saying that there won't be more warnings
+                    expect(Lib.warn.calls.count()).toEqual(5 + 1);
+                }).catch(fail).then(done);
```
etpinard: Excellent test. Thanks!

etpinard:
Good stuff. Thanks for taking on the animation code. 

:dancer:

Issue #221, opened 2016-01-27
[Feature request] 'symlog' scale


liuyigh:
Dear fellow developers and plot.ly users,

I would like to request a new feature: "symmetrical log" scale. Matplotlib has it:

http://matplotlib.org/examples/pylab_examples/symlog_demo.html

Also see: What is the difference between 'log' and 'symlog'? [link](http://stackoverflow.com/questions/3305865/what-is-the-difference-between-log-and-symlog)

It would be great if plot.ly can adopt this scale. Thanks!

-- Y

mdtusz:
This isn't a priority right now, but shouldn't be too difficult to implement down the road. We're always open to pull requests as well!

veged:
maybe any workarounds here without changes in the library itself? for example as in Highcharts http://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/samples/highcharts/yaxis/type-log-negative/

veged:
@etpinard is there any reason to keep current `log` scale in state of unsupported negative values? maybe it will be enough to change https://github.com/plotly/plotly.js/blob/master/src/lib/to_log_range.js to properly support negative values?

I can make PR if you show me all places to change exept those I found by myself:

```
src/lib/to_log_range.js
17:module.exports = function toLogRange(val, range) {

src/lib/index.js
25:lib.toLogRange = require('./to_log_range');

src/components/annotations/convert_coords.js
13:var toLogRange = require('../../lib/to_log_range');
45:        if(toLog) newVal = toLogRange(currentVal, ax.range);

src/components/images/convert_coords.js
13:var toLogRange = require('../../lib/to_log_range');
57:                newPos = toLogRange(currentPos, ax.range);
```

alexcjohnson:
symlog is a little funny, as it's not a smooth mapping. A variant I've seen before though that *is* smooth is arcsinh (inverse hyperbolic sine) - this one is particularly nice as you can define both the zero and the linear-ish range, using `y' = arcsinh( (y - y0) / yL )`. This is good for situations like a noisy signal with a nonzero base level, where you want to see both small and large peaks popping out with either sign.

PavelGolodoniuc:
@alexcjohnson, this is exactly the approach we are taking in our signal processing app where log scale would be ideal but where non-zero values are commonly present.

![image](https://user-images.githubusercontent.com/9332295/36184616-cecada62-116f-11e8-8fa8-60451176c114.png)

[...]

PR #2318, opened 2018-02-01
A few more ways to configure `marker.size` scales

etpinard:
This PR adds a few attributes to the `marker` container for scatter-like traces. 

These new attributes are:

-  `marker.sizemax`: Sets the maximum size (in px) of the rendered marker points. Analogous to the existing `marker.sizemax`.
- `marker.sizedatamin`: Sets the minimum `marker.size` value to be scaled. Values smaller than `sizedatamin` are displayed with `sizemin`.
- `marker.sizedatamax`: Sets the maximum `marker.size` value to be scaled. Values greater than `sizedatamin` are displayed with `sizemax`.

and `marker.sizemode` has two new possible values:

- `'log-diameter'`: the base10 log of the `marker.size` data is proportional to marker points' diameter .
- `'log-area'`, the base10 log of the `marker.size` data is proportional to marker points' area.

Alternatively, we could add a new attribute (named e.g. `marker.sizescaletype: 'linear' || 'log'`), but while writing the description I thought that a looping log-scale settings into `sizemode` would be clearer for users.

cc @alexcjohnson


greg9504:
Hello, 
We hadn't heard from you so we had gone ahead and implemented this too.  I've attempted to merge the two here:
https://github.com/greg9504/plotly.js
Your version does not handle negative values, and the log handling only converts the size data.  The sizedatamin and sizedatamax should be converted too.  See this commit:
https://github.com/greg9504/plotly.js/commit/7ea8b88d7160dcd8135c89482107d66c42ef6af4

The biggest difference is that sizeref is NOT used.  Also I'm not too sure how the area mode is meant to be used.

etpinard:
Hi @greg9504 

First off, sorry for the wait. It's been a pretty crazy couple months in plotly.js land feature-wise. 

Thanks for sharing your commit. It definitely makes it easier to understand what you're going after.

---------------

Now regarding the discrepancies between your commit and this PR:

> Your version does not handle negative values

A while back, we (i.e me and. @alexcjohnson ) chose to ignore negative values in `marker.size` (i.e. making size=0) so that `marker.size` keeps its _physical_ meaning if I remember correctly. Referencing the pre-open-source (private) PR here: https://github.com/plotly/streambed/pull/2947/ 

I'm not oppose to adding support for negative values in `marker.size` scales, but it would have to come under a new attribute to not break the current default behavior.

> and the log handling only converts the size data. The sizedatamin and sizedatamax should be converted too

Personally, I thought `sizedatamin` and `sizedatamax` are more meaning if we keep them in true (i.e. not-log) data-space having the same units as the `marker.size` values. Would you mind sharing one of your use cases to enlighten us on the topic?

One more thing I noticed:

We can do better than `sizedataislog` which doesn't scale well. Let's say we want to add _symlog_ scale option down the road (referencing https://github.com/plotly/plotly.js/issues/221)?. I thought all scale flavours could easily be encapsulated in `marker.sizemode`. What do you think?

-----

Thanks for your help and patience. We're looking forward to having this feature part of plotly.js!

greg9504:
Hello,
I can't read plotly/streambed#2947 but if I understand what you are saying is this a correct way of thinking of the current size array?:

- array of integer values that are literal marker size values (physical meaning), and this is why negative values do not make sense.  

What we had thought was that the size array could point to any type of data.  In our case this can be measured data (things like temperature, pressure etc.) or calculated data.  This data can have negative values, log scale etc.  In the original proposal sent to Rob Williams it's clear that is what we intended, I believe we even provided some sample Matlab code to calculate the symbol sizes from the data values.

I can see from plotlys point of view that interpreting the marker.size array as literal marker sizes makes sense, and it keeps the interface simple.  Perhaps there should have been another field called:
marker.sizedata - array field used to calculate marker.size values.  If set Plotly handles filling in the marker.size array using the scaling method as implemented in my code? 

As part of the proposal Plotly also needs to provide a marker size legend.  Example from the proposal:
image

In order to create that legend and label it correctly the marker.sizedata, marker.sizedatamin, marker.sizedatamax, marker.sizemin, marker.sizemax, and the code to calculate the marker size are all needed.  

I will post up a plot json config with some test data shortly.

greg9504:
Hello,
Any updates to this?

etpinard:
Closing (due to numerous merge conflicts). Discussion moved to -> https://github.com/plotly/plotly.js/issues/3129

PR #2323, opened 2018-02-01
Switch to node8 & npm5

etpinard:
Tech-debt week continues :pick: :moneybag: :tada: 

This PR makes CI use node v8.9.4 and npm v5 (which works much better v3 :smirk: ). This CI image use Chrome 64 (i.e. the latest version) which should make it easier for us to make our tests run consistency locally and on CI.

Using npm5 gives us access to `package-lock.json` file (more info [here](https://docs.npmjs.com/files/package-lock.json)) which contains a complete description of the `node_modules/` tree. Now, `npm i` (should) give the same result everywhere everytime. This makes a few our tooling helpers obsolete: `dist/npm-ls.json` is gone and no need to `dedupe` on `preversion`.

A few more things:
- dev deps have been updated. Most notably `eslint@4` (which runs a little faster now) and `karma@2`,
- [`check-node-version`](https://www.npmjs.com/package/check-node-version) now runs on `preversion` ensuring that whoever bundles the `dist/` use node8 and npm5.

cc @alexcjohnson @dfcreative @nicolaskruchten @monfera


.eslintrc
```diff
@@ -65,6 +66,7 @@
     "no-use-before-define": [2, "nofunc"],
     "no-loop-func": [2],
     "no-console": [0],
-    "no-unused-labels": [2]
+    "no-unused-labels": [2],
+    "no-useless-escape": [0]
```
alexcjohnson: how badly do we break this rule?

alexcjohnson:
Looks great, lets do it! 💃

.eslintrc
etpinard: this bad:

image

etpinard: ... and they are not auto-fixable (using `eslint . --fix`) unfortunately.

etpinard: ... and I don't want to waste time with linting stuff. I really think we should be switching to `standardjs` and be done with it (cc https://github.com/plotly/plotly.js/issues/950) :stuck_out_tongue:

etpinard:
Ok. Merging this in. 

------

Devs, please make sure to update to node8/npm5 and `rm -rf node_modules && npm i` after doing so. :rocket:

Issue #2228, opened 2017-12-29
BUG: Figure breaks after animating a property of an array element


jmmease:
It seems that something goes wrong when using `Plotly.animate` to change a property of an array element. I hit this while working on a custom trace type, but it is reproducible on the Parallel Coordinate trace.

**Note:** I understand animated transitions are not supported for the Parallel Coordinates trace, but my expectation is that an animate command shouldn't cause an error and should behave like a relayout/restyle command.

Parallel Coordinate Example Pen: https://codepen.io/anon/pen/ZvKGxW

Summary: 
 - Animating a change to `line.color` works fine
 - Restyling a change to `dimensions[0].constraintrange` works fine
 - Animating a change to `dimensions[0].constraintrange` breaks the plot (Nothing shows up afterward) 

Example code for reference:
```
var trace = {
  type: 'parcoords',
  line: {
    color: 'blue'
  },
	
  dimensions: [{
    range: [1, 5],
    constraintrange: [1, 2],
    label: 'A',
    values: [1,4]
  }, {    
    range: [1,5],
    label: 'B',
    values: [3,1.5],
    tickvals: [1.5,3,4.5]
  }]
};

var data = [trace]

Plotly.plot('graphDiv', data).then(function() {
	// ## This animate command works as expected (line turns red) ##
	// Plotly.animate('graphDiv', 
	// 			{data: [{'line.color': 'red'}],
	// 			 traces: [0],
	// 			 layout: {}})
	
	// ## This restyle command works as expected (constraintrange extends to [1, 4]) ##
	// Plotly.restyle('graphDiv', {'dimensions[0].constraintrange': [[1, 4]]}, 0)
	
	// ## This animate command breaks things and the plot doesn't show up ##
	Plotly.animate('graphDiv', 
                                {data: [{'dimensions[0].constraintrange': [1, 4]}],
                                  traces: [0],
                                  layout: {}})
});
```

etpinard:
The fix must be done inside the [`Plots.transitions`](https://github.com/plotly/plotly.js/blob/48512eb5db6227be88790b8a5202ec06ea21b918/src/plots/plots.js#L2046) helpers most likely in [`Plots.computeFrame`](https://github.com/plotly/plotly.js/blob/48512eb5db6227be88790b8a5202ec06ea21b918/src/plots/plots.js#L1836). All trace types should _work_ with `Plotly.animate` the fallback being an equivalent `restyle`, `relayout` or `update` call.

[...]

PR #2324, opened 2018-02-02
Add parcoords 'dimension' to list of array containers found in traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2228

We already had logic built-in that merge data array container in frame. Previously only one special key was listed ('transforms`), now parcoords `dimensions` is there too.

@monfera would you mind reviewing this thing?


monfera:
Just tested, fixes the issue and the test covers it, looks great! 💃

Issue #2242, opened 2018-01-10
Mesh3d hover text not shown


Helsioufy:
I was trying to draw a mesh3d with customized hover text in the below example:

https://codepen.io/Elsioufy/pen/JMLEmP

Yet the hover text did not appear to be shown in the graph. it is working with scatter3d and surface though.

``` 

var gd = document.getElementById('graph');

var x = [1, 1, 2, 3, 4, 2];
var y = [2, 1, 3, 4, 5, 3];
var z = [3, 7, 4, 5, 3.5, 2];

// fill in 'text' array for hover
var text = x.map((xi, i) => `
ts: ${xi}<br>
hz: ${y[i]}<br>
ftt: ${z[i]}`)

Plotly.plot(gd, [{
  type: 'mesh3d',
  text: text,
  hoverinfo: 'text',
  x: x,
  y: y,
  z: z,
  text: text,
}], {
  scene: {
    xaxis: { title: 'x'},
    yaxis: { title: 'y'},
    zaxis: { title: 'z'},
  }
});
```
![no_text](https://user-images.githubusercontent.com/16387211/34777195-3af30aca-f622-11e7-883f-4da0c7ad1f43.png)

alexcjohnson:
This trace also seems to suffer from misplaced hover ala #2208

PR #2327, opened 2018-02-02
Add support for (hover) 'text' in mesh3d traces

etpinard:
Call this a bug fix or a new feature, this PR closes https://github.com/plotly/plotly.js/issues/2242

### DIFF
--- a/src/traces/surface/attributes.js
+++ b/src/traces/surface/attributes.js
@@ -11,6 +11,7 @@
 var Color = require('../../components/color');
 var colorscaleAttrs = require('../../components/colorscale/attributes');
 var colorbarAttrs = require('../../components/colorbar/attributes');
+var baseAttrs = require('../../plots/attributes');
 
 var extendFlat = require('../../lib/extend').extendFlat;
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
@@ -242,7 +243,9 @@ var attrs = module.exports = overrideAll({
         zmax: extendFlat({}, colorscaleAttrs.zmax, {
             description: 'Obsolete. Use `cmax` instead.'
         })
-    }
+    },
+
+    hoverinfo: extendFlat({}, baseAttrs.hoverinfo, {editType: 'calc'})
 }, 'calc', 'nested');
 
 attrs.x.editType = attrs.y.editType = attrs.z.editType = 'calc+clearAxisTypes';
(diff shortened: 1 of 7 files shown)


alexcjohnson:
Not sure how much you want to let this PR expand, but here's a definitely-a-bug that popped up as I was trying out your PR: `hoverinfo` has the wrong `editType`:
`Plotly.restyle(gd,{hoverinfo:'x+y'})`
does not update the displayed hover info for mesh3d. If you follow it with
`Plotly.redraw(gd)`
or include something else in the restyle though, it does, like
`Plotly.restyle(gd,{hoverinfo:'x+text',text:'AAAAAAAH'})`

Also it looks like `surface` doesn't support `text` as a single string (only as an array) and both `surface` and `scatter3d` have the same `hoverinfo` restyle problem as `mesh3d`.

etpinard:
@alexcjohnson good catch. Thanks.

I think I fixed them all in https://github.com/plotly/plotly.js/pull/2327/commits/8ae2d0e0acc669f8e097cc4e9353e85e3e34ed1c by overriding the `hoverinfo` attribute declaration with `editType: 'calc'`.

alexcjohnson:
https://github.com/plotly/plotly.js/commit/8ae2d0e0acc669f8e097cc4e9353e85e3e34ed1c looks great, thanks for the tests 🔒 
Do you want to extend `surface` to support `text` as a single string now too, so it matches the new `mesh3d.text`, or leave that in an issue?

etpinard:
> Do you want to extend surface to support text as a single string now too

Oops. I pushed too soon. That's in https://github.com/plotly/plotly.js/pull/2327/commits/d20bdc1aae4e7ec42e27678c78951c429687b9ca

src/traces/surface/attributes.js
```diff
@@ -113,9 +113,18 @@ var attrs = module.exports = overrideAll({
     },
 
     text: {
-        valType: 'data_array',
-        description: 'Sets the text elements associated with each z value.'
+        valType: 'string',
+        role: 'info',
+        dflt: '',
+        arrayOk: true,
+        editType: 'calc',
```
alexcjohnson: Oh this is in an `overrideAll` so you don't need `editType` - which I think applies to a few more of the additions in this PR too.

alexcjohnson: (I see you 🔪 some `editType` lines, just not the one I made the comment directly on 😅 )

etpinard: Oops. Fixed in -> 13a3ecf

alexcjohnson:
Nice - slowly getting all these trace types up to par with each other. 💃

Issue #2317, opened 2018-02-01
Bug - Hoverformat for second axis fails if axis is not visible


emanuelnastasa:
**Observed behavior:**
Having a chart with 2 Y axes, if you mark the second Y axis as "visible:false" the hoverformat property for that axis is ignored and the values are shown in exponential form.

**Expected behavior:**
The hoverformat is respected regardless of the visible parameter.

**Steps to reproduce:**
1.Access the JSFiddle:https://jsfiddle.net/a75dhu5z/
2. Change layout.yaxis2.visible between false and true and observe the value displayed in the hover tool-tip of the top point in the orange trace.

**Build found**
It's reproducing in all builds from 1.31 to the latest release 1.33.1

etpinard:
Thanks for the report.

PR #2329, opened 2018-02-02
Coerce hoverformat on visible-false axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2317

`visible: false` axes should still support `hoverformat` as demonstrated in the https://github.com/plotly/plotly.js/issues/2317 fiddle. 

To do so, this PR move the `coerce('hoverformat')` out of `handleTickLabelDefaults` -- which does not get called for `visible: false` axes -- to the base axis default handler.

### DIFF
--- a/src/components/colorbar/defaults.js
+++ b/src/components/colorbar/defaults.js
@@ -53,11 +53,9 @@ module.exports = function colorbarDefaults(containerIn, containerOut, layout) {
 
     handleTickValueDefaults(colorbarIn, colorbarOut, coerce, 'linear');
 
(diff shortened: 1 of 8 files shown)


etpinard:
Note that commit https://github.com/plotly/plotly.js/commit/86067a6eae0d6fadc427477b6a4dc86fb1e42e74#diff-9169f5ae07a510e5c981221d9970ee67 was culprit for #2317 from PR #2200

alexcjohnson:
Very nice - 💃 . Some good cleanup in here as well 🎁

Issue #2338, opened 2018-02-06
Multi-type subplot problem


alexcjohnson:
the mock https://rreusser.github.io/plotly-mock-viewer/#plot_types has two problems:
- two of the ternary axes don't draw, and generate errors in d3
![screen shot 2018-02-06 at 12 31 55 pm](https://user-images.githubusercontent.com/2678795/35874150-c3e7759c-0b39-11e8-862d-262f62635597.png)
- the ternary and geo subplots have broken hover text (no main hover label, little blank common label)
![screen shot 2018-02-06 at 12 33 00 pm](https://user-images.githubusercontent.com/2678795/35874195-e7eca2c8-0b39-11e8-85f6-303207906c78.png)

Interestingly `ternary_multiple` and `geo_multi-geos` are both fine… @etpinard speculates that cartesian+non-cartesian may be the culprit. Also weird that this mock renders just fine in the images tests on CI.

PR #2339, opened 2018-02-06
Negate axis offset and then turn it into a string

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2338

Not 100% sure how to test this. There's probably a way to add a check in `d3-strict`? But maybe just a jasmine test that doesn't fail would do the trick.

cc @alexcjohnson

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -175,9 +175,10 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         subplots = subplots.concat(overlayedSubplots);
     }
 
-    var len = subplots.length,
-        xaArray = new Array(len),
-        yaArray = new Array(len);
+    var len = subplots.length;
+    var xaArray = new Array(len);
+    var yaArray = new Array(len);
+    var supportsCompare = false;
 
     for(var i = 0; i < len; i++) {
         var spId = subplots[i];
@@ -185,6 +186,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         // 'cartesian' case
         var plotObj = plots[spId];
         if(plotObj) {
+            supportsCompare = true;
 
             // TODO make sure that fullLayout_plots axis refs
             // get updated properly so that we don't have
@@ -202,6 +204,8 @@ function _hover(gd, evt, subplot, noHoverEvent) {
     }
 
     var hovermode = evt.hovermode || fullLayout.hovermode;
+
+    if(hovermode && !supportsCompare) hovermode = 'closest';
 
     if(['x', 'y', 'closest'].indexOf(hovermode) === -1 || !gd.calcdata ||
             gd.querySelector('.zoombox') || gd._dragging) {
--- a/src/components/modebar/manage.js
+++ b/src/components/modebar/manage.js
@@ -82,10 +82,13 @@ function getButtonGroups(gd, buttonsToRemove, buttonsToAdd) {
     var hasTernary = fullLayout._has('ternary');
     var hasMapbox = fullLayout._has('mapbox');
     var hasPolar = fullLayout._has('polar');
+    var allAxesFixed = areAllAxesFixed(fullLayout);
 
     var groups = [];
 
     function addGroup(newGroup) {
+        if(!newGroup.length) return;
+
         var out = [];
 
         for(var i = 0; i < newGroup.length; i++) {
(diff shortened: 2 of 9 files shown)


src/plots/ternary/ternary.js
```diff
@@ -328,12 +328,12 @@ proto.adjustLayout = function(ternaryLayout, graphSize) {
     _this.layers.bgrid.attr('transform', bTransform);
 
     var aTransform = 'translate(' + (x0 + w / 2) + ',' + y0 +
-        ')rotate(30)translate(0,-' + aaxis._offset + ')';
+        ')rotate(30)translate(0,' + -aaxis._offset + ')';
```
etpinard: When `_offset` is negative we use to get:

image

that is double `-` in the translate transform string.

Looks like `nw.js@0.12` doesn't care, so this went unnoticed in our image tests.

alexcjohnson:
Great, was hoping it would be that simple - does this fix the hover labels too?

A `d3-strict` test could work - if setting `transform` check that `--` doesn't appear in the string?

etpinard:
> does this fix the hover labels too?

This one is bit tricky, `hovermode` defaults to `'x'` in `plot_types.json`, which isn't compatible with subplot types other than `cartesian`. With `hovermode: 'closest'`, everything works fine. Now, gl3d and pie seem not to care when `hovermode: 'x'` is set (they behave the same as for `hovemode: 'closest'`) whereas geo, ternary and polar `hovermode: 'x'` show an odd-looking arrow at the bottom of the corresponding subplots.

Clearly those odd-looking arrow need to go. But I ask: should `hovermode: 'x'` for subplot types without and x-axis behave like `hovermode: 'closest'` or like `hovermode: false`?

alexcjohnson:
> should `hovermode: 'x'` for subplot types without and x-axis behave like `hovermode: 'closest'` or like `hovermode: false`?

I think anything that's not explicitly `hovermode: false` should behave like `'closest'` on subplots where compare mode doesn't make sense.

etpinard:
> I think anything that's not explicitly hovermode: false should behave like 'closest' on subplots where compare mode doesn't make sense.

Sounds good. Let me fix that in this PR.

etpinard:
@alexcjohnson if you're looking to fix an easy (I think) bug before the `1.34.0` release, you can fixing the `hovermode` part of https://github.com/plotly/plotly.js/issues/2338 as described in https://github.com/plotly/plotly.js/pull/2339#issuecomment-363531321 . Feel free to push to this branch directly.

I have to keep working on the `mapbox-gl` update until Wednesday.

src/components/fx/hover.js
```diff
@@ -203,6 +205,8 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
     var hovermode = evt.hovermode || fullLayout.hovermode;
 
+    if(hovermode && !supportsCompare) hovermode = 'closest';
```
alexcjohnson: The second part of #2338 - fixes hover labels for geo and ternary on the `plot_types` mock

etpinard: Looks good. Not sure what's the best to :lock: this down?

Maybe just checking that `hovermode: 'x'` behaves the same as `hovermode: 'closest'` in a cartesian + geo (or ternary or polar) graph?

src/components/modebar/manage.js
```diff
@@ -100,55 +103,71 @@ function getButtonGroups(gd, buttonsToRemove, buttonsToAdd) {
...
-        addGroup(['hoverClosestPie']);
-    } else if(hasMapbox) {
-        addGroup(['resetViewMapbox', 'toggleHover']);
-    } else if(hasGeo) {
-        addGroup(['zoomInGeo', 'zoomOutGeo', 'resetGeo']);
-        addGroup(['hoverClosestGeo']);
-    }
+    addGroup(dragModeGroup);
+    addGroup(zoomGroup.concat(resetGroup));
+    addGroup(hoverGroup);
```
etpinard: Nice cleanup. Looking good!

src/components/fx/hover.js
alexcjohnson: good call -> https://github.com/plotly/plotly.js/pull/2339/commits/16b01f002b4b2874ff191bde6ce9cc9f9f38f1db

etpinard:
Nicely done :dancer:

alexcjohnson:
Likewise, 💃 on the `--` fix and `d3-strict` test 🎉

Issue #1850, opened 2017-07-04
a version of .newPlot that does the .update logic behind the scenes


chriddyp:
In Dash, I'd like the Graph component to do smart `restyle` and `relayout` updates when I call `newPlot` or `plot` (or some new method). Dash users deal with `figure`s and not diffs.

Programming this logic myself is a lot of work that other groups could benefit from if it was done centrally in plotly.js. I wrote a lot of the logic in the workspace and it took a while to work through all the kinks. Some of the edge cases included:
- calling `newPlot` instead of `update` at the start
- calling `addTraces` manually instead of `update`
- calling `newPlot` instead of `update` when switching between 2D, 3D, or ternary
- calling `newPlot` when switching to a financial chart type
- deleting cartesian axes when swithing to ternary or mapbox

[Here is that code](https://github.com/plotly/streambed/blob/master/shelly/filewell/static/filewell/src/components/workspace/panels/PlotPanel.react.js#L367-L613)

This would form the basis of a plotly.js React component.

Dash users will benefit from this immediately with:
- Faster updates enabling their apps to refresh every 50ms and mimic live-streaming
- Mapbox updates with full map redraws

Ideally, this logic would work its way into a robust `.animate` method as well ([.animate method that can handle any type of figure diff #1849](https://github.com/plotly/plotly.js/issues/1849))

***

cc @rreusser @alexcjohnson @etpinard @cpsievert @monfera @jackparmer

etpinard:
> - calling newPlot instead of update when switching between 2D, 3D, or ternary
> - calling newPlot when switching to a financial chart type
> - deleting cartesian axes when swithing to ternary or mapbox

Those are arguably `newPlot` bugs.

etpinard:
I'd vote for adding a new `Plotly` method (roll the :drum: ) `Plotly.react`.

monfera:
Arriving at the delta spec between two subsequent specs sounds similar to a DOM diffing algo. With Étienne's `Plotly.react` it'd be black box to Dash / Python but internally, at least from some point onward, it'd be good to handle updates incrementally. The current `Plotly.restyle` calls occupy the middle ground, not granular enough for the fastest updates and not aggregate enough to allow full restatement of a plot.

rreusser:
[...]

Issue #2007, opened 2017-09-15
sanitize all subplot attribute names


alexcjohnson:
From @etpinard in https://github.com/plotly/plotly.js/pull/1999#discussion_r138750557 - we've added a lot of stuff and not all of it is handled equivalently in `cleanLayout` / `cleanData`. Do we want to?

> We should perhaps try to sanitize all subplots attribute names in `cleanLayout`. Looks like we only sanitize cartesian, gl2d and gl3d axes at the moment.

etpinard:
Closing. Probably not worth adding more to `cleanData` and `cleanLayout` at this stage if we're planning on :hocho: them in v2.

Issue #1741, opened 2017-05-29
Should we add add *strict* (or *debug*) mode that runs `Plotly.validate` on every `newPlot` and `update` call before attempting to draw?


etpinard:
which would essentially be a more general way to resolve https://github.com/plotly/plotly.js/pull/1736 

I'm thinking this strict/debug mode would be set as a config argument. We could either log the attribute errors (as `Plotly.validate` currently does) or throw an error.

rreusser:
Random question: does a bundle with stripped metadata affect at all what or how it's able to report errors?

etpinard:
>  does a bundle with stripped metadata affect at all what or how it's able to report errors

In its current implementation, `Plotly.validate` doesn't log the attribute descriptions, so no, it wouldn't matter.

etpinard:
cc @phxnsharp if we weren't aware of this.

phxnsharp:
This sounds like exactly what we would prefer. Is there a sense of what type of performance hit this would involve?

etpinard:
> Is there a sense of what type of performance hit this would involve?

At the moment, `Plotly.validate` is fairly fast. It clocks in at no worse than `50ms`. `Plotly.validate`'s execution time, in its current state, _scales_ with the number of `"data"` and `"layout"` attributes set.

That said, if we choose to implement https://github.com/plotly/plotly.js/issues/1998, `Plotly.validate` will likely slow down a little bit. It we only check for data arrays lengths, the performance hit will be small. But if we choose to check for invalid data items (e.g. `null` in scatter marker traces), this will make `Plotly.validate`'s execution time _scale_ with the length of the data array - which will slow it down considerably for graphs with many data points.

etpinard:
cross-referencing https://github.com/plotly/plotly.js/issues/167

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2341, opened 2018-02-06
Plotly.react

alexcjohnson:
Introduces `Plotly.react` - an idempotent, full-state, self-diffing update method with a signature matching `Plotly.newPlot` but using the faster update pathways of `Plotly.restyle/relayout/update`.

Diffing is done on `fullData` and `fullLayout`, not on `data` and `layout`, for a few reasons:
- We don't copy `data` and `layout`, so there's nothing to diff there if the user modifies them in place. But `_full*` get regenerated as new objects with every `supplyDefaults` call (which must happen on every update regardless).
- No need to worry about the data types we're diffing, as `supplyDefaults` cleans that up. Also we won't diff user attributes that don't go into the plot, or irrelevant leftovers of old plot types.

We do NOT dive into data arrays (`valType: 'data_array'` or `arrayOk: true`) in the diffing algorithm. There are two ways to tell `Plotly.react` that the data have changed:
- `layout.datarevision`: can take any value - string, number, whatever - and if it is not `===` its previous value, we treat the data as changed. Data arrays themselves are ignored during diffing. Note that right now there's no difference in the pathway we'd take for one vs all data arrays changing, but in the future if we implement partial recalc we can make a similar flag for each array, or perhaps each trace.
- If you omit `datarevision`, we assume data arrays are being used as immutable and compare them with `===` to determine if anything changed. This has some known issues (resulting in slow recalc when it wouldn't be needed) that we should fix:
  - At least one trace type (surface) fills in index arrays for missing dimensions. These are never `===` each other. We should provide these later in the pipeline, not attached to `_fullData`, or fill them in as `_private` attributes which do not get diffed.
  - At least one trace type (scatter) will slice uneven arrays to the same length, which will also make them compare unequal. We should just record a private `_length` or something instead.

see #1850 
cc @etpinard @chriddyp @nicolaskruchten @VeraZab @jackparmer


src/components/legend/defaults.js
```diff
@@ -58,6 +56,8 @@ module.exports = function legendDefaults(layoutIn, layoutOut, fullData) {
 
     if(showLegend === false) return;
 
+    layoutOut.legend = containerOut;
```
alexcjohnson: legend's visibility is controlled by `showlegend`, which is *outside* the `legend` container itself, unlike most of our newer containers that have `visible` *inside* the container. Unless and until we change that (v2?) the legend container should not exist when it's not needed. This was causing me problems as `relinkPrivateKeys` was humorously removing `newFullLayout.legend` if it was empty and `oldFullLayout.legend` existed... so its existence would toggle on `Plotly.react` calls, making it look like the legend was always in need of updating 💫

src/plot_api/plot_api.js
```diff
@@ -1632,26 +1639,6 @@ function _restyle(gd, aobj, traces) {
         }
     }
 
-    // check axes we've flagged for possible deletion
-    // flagAxForDelete is a hash so we can make sure we only get each axis once
-    var axListForDelete = Object.keys(flagAxForDelete);
```
alexcjohnson: This isn't really related to `Plotly.react` but noticed as I was working here: post #2227 we don't need to delete unused axes anymore - in fact we don't *want* to.

alexcjohnson: Oh but it's not self-contained here - I need to get rid of `flagAxForDelete` above...

src/plot_api/plot_api.js
```diff
@@ -2208,6 +2195,356 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+    function addFrames() { return Plotly.addFrames(gd, frames); }
+
+    gd = Lib.getGraphDiv(gd);
+
+    var oldFullData = gd._fullData;
+    var oldFullLayout = gd._fullLayout;
+
+    // you can use this as the initial draw as well as to update
+    if(!Lib.isPlotDiv(gd) || !oldFullData || !oldFullLayout) {
+        plotDone = Plotly.newPlot(gd, data, layout, config);
```
alexcjohnson: `Plotly.react` is a one-stop-shop, it can replace `Plotly.newPlot` and `Plotly.plot` entirely.

src/plot_api/plot_api.js
```diff
@@ -2208,6 +2195,356 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+        var immutable = newFullLayout.datarevision === undefined;
+
+        var restyleFlags = diffData(gd, oldFullData, newFullData, immutable);
+        var relayoutFlags = diffLayout(gd, oldFullLayout, newFullLayout, immutable);
+
+        // clear calcdata if required
+        if(restyleFlags.calc || relayoutFlags.calc) gd.calcdata = undefined;
+
+        // Note: what restyle/relayout use impliedEdits and clearAxisTypes for
+        // must be handled by the user when using Plotly.react.
```
alexcjohnson: ⚠️ not sure if there's something we can do to make this easier for users (pull out the relevant sections from `_restyle` and `_relayout` perhaps into helper functions?), but I don't think we should be trying to do anything with `impliedEdits` or `clearAxisTypes` inside `Plotly.react`.

src/plot_api/plot_api.js
```diff
@@ -2208,6 +2195,356 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+
+        // fill in redraw sequence
+        var seq = [];
+
+        if(frames) {
+            gd._transitionData = {};
+            Plots.createTransitionData(gd);
+            seq.push(addFrames);
```
alexcjohnson: anyone want to comment on how we should handle `frames` here?

src/plot_api/plot_api.js
```diff
@@ -2208,6 +2195,356 @@ Plotly.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+            var oldConfig = Lib.extendDeep({}, gd._context);
+            gd._context = undefined;
+            setPlotContext(gd, config);
+            configChanged = diffConfig(oldConfig, gd._context);
+        }
+
+        gd.data = data || [];
+        helpers.cleanData(gd.data, []);
+        gd.layout = layout || {};
+        helpers.cleanLayout(gd.layout);
```
alexcjohnson: I'm a little ambivalent about this - there's part of `cleanData`, fillling in uids, that seems like it may not even belong to `cleanData` but anyway we need to do it. But the rest, do we need to be cleaning obsolete usages on every `Plotly.react` call? These routines mutate the inputs so after the first call the user shouldn't have any of them left; that said users can always provide totally new data/layout...

src/plots/cartesian/graph_interact.js
```diff
@@ -18,7 +20,13 @@ var makeDragBox = require('./dragbox').makeDragBox;
 module.exports = function initInteractions(gd) {
     var fullLayout = gd._fullLayout;
 
-    if((!fullLayout._has('cartesian') && !fullLayout._has('gl2d')) || gd._context.staticPlot) return;
+    if(gd._context.staticPlot) {
+        // this sweeps up more than just cartesian drag elements...
+        d3.select(gd).selectAll('.drag').remove();
```
alexcjohnson: Far as I could tell this is the only change needed so `config` can be altered with a regular `Plotly.plot` call - everything else I tried just seemed to work. If we make more of a `supplyDefaults` type framework for config at some point we may be able to minimize the update pathway there too, like modebar-only changes... but it's probably not a very common use case anyhow.

src/plots/plots.js
```diff
@@ -281,13 +281,19 @@ var extraFormatKeys = [
...
-        newFullLayout = gd._fullLayout = {},
-        newLayout = gd.layout || {};
+    var oldFullLayout = gd._fullLayout || {};
 
-    var oldFullData = gd._fullData || [],
-        newFullData = gd._fullData = [],
-        newData = gd.data || [];
+    if(oldFullLayout._skipSD) {
+        delete oldFullLayout._skipSD;
+        return;
```
alexcjohnson: my hack to prevent running `supplyDefaults` twice in full-replot update pathways.

src/plot_api/plot_api.js
alexcjohnson: finished deleting in https://github.com/plotly/plotly.js/pull/2341/commits/aab4f11ff1200b8a58d287a03670034354a6b060

alexcjohnson:
@etpinard I still don't understand why I have this big `package-lock` diff - I can remove it from the PR, but I'd much rather get to a situation where I'm *not* constantly generating a different one from what's in the repo...

test/jasmine/tests/plot_api_test.js
```diff
@@ -2171,6 +2171,239 @@ describe('Test plot api', function() {
...
+
+                data[0].type = 'heatmap';
+                delete layout.scene;
+                return Plotly.react(gd, data, layout);
+            })
+            .then(function() {
+                countCalls({plot: 1});
+
+                // ideally we'd just do this with `surface` but colorbar attrs have editType 'calc' there
+                // TODO: can we drop them to type: 'colorbars' even for the 3D types?
```
etpinard: Yeah, this should work with `editType: 'colorbars'`.

src/components/annotations/attributes.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
         valType: 'boolean',
         role: 'info',
         dflt: true,
-        editType: 'calcIfAutorange',
+        editType: 'calcIfAutorange+arraydraw',
```
etpinard: Can you clarify why we now need `arraydraw` here and in `shapes/attributes.js`?

src/components/legend/defaults.js
etpinard: Very nice catch. 



> until we change that (v2?)

I'd vote +1 here.

src/lib/push_unique.js
```diff
@@ -30,7 +32,7 @@ module.exports = function pushUnique(array, item) {
         }
         array.push(item);
     }
-    else if(item && array.indexOf(item) === -1) array.push(item);
+    else if((item || item === 0) && array.indexOf(item) === -1) array.push(item);
```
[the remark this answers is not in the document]
etpinard: None from me. Let's just make sure it's documented in this its test [suite](https://github.com/plotly/plotly.js/blob/ef2bff1904b50fa6fd01655ac37a588866a0e330/test/jasmine/tests/lib_test.js#L1281).

src/plot_api/plot_api.js
etpinard: _the dream_

src/plot_api/plot_api.js
etpinard: I agree, the `uid` generation routine should be taken out. It has no place in `cleanData`. 

In `v2`, I'm thinking `cleanData` and `cleanLayout` should be called only under a special config flag (e.g. `compatibility: true`, or a more granular `compatibilityLevel: 'v1'`) whereas assigning uid to traces will always have to be done.

cc https://github.com/plotly/plotly.js/issues/2007

src/plot_api/plot_api.js
etpinard: Can you clarify what you mean by: _must be handling by user_?

src/plot_api/plot_api.js
etpinard: Looks good to me :ok_hand:

src/plots/layout_attributes.js
```diff
@@ -182,4 +182,18 @@ module.exports = {
...
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'If provided, a changed value tells `Plotly.react` that',
+            'one or more data arrays has changed. This way you can modify',
+            'arrays in-place rather than making a complete new copy for an',
+            'incremental change.',
+            'If NOT provided, `Plotly.react` assumes that data arrays are',
+            'being treated as immutable, thus any data array with a',
+            'different identity from its predecessor contains new data.'
```
etpinard: This is brilliant :sparkle:

src/plots/cartesian/graph_interact.js
etpinard: What about other svg-based subplot types? For example, I'm thinking we could add a similar `.remove()` statement for polar [here](https://github.com/plotly/plotly.js/blob/ef2bff1904b50fa6fd01655ac37a588866a0e330/src/plots/polar/polar.js#L534)

src/lib/push_unique.js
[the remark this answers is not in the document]
etpinard: Oh oops, that's done already. Thanks!

src/components/legend/defaults.js
alexcjohnson: added to #420

VeraZab: > the legend container should not exist when it's not needed. 

so by default, in _fullLayout, the .legend key wouldn't be there?
we read from _fullLayout to display widgets in the editor, if there's no .legend key, or a ._fullLayout.legend.visible key, then the widget won't display..

We could have a work around, but it's kind of one of the basic premises of widget visibility in the editor..

src/plots/layout_attributes.js
nicolaskruchten: yes, yes, yes!

src/plot_api/plot_api.js
alexcjohnson: `impliedEdits` are things like turning off `axis.autorange` when you set an explicit `range`. If you use `relayout` et al we do that for you, so `Plotly.relayout(gd, 'yaxis.range', [5, 10])` automatically sets `yaxis.autorange: false` - otherwise the range wouldn't actually change, as autorange would put it back to what you started with.

The equivalent with `Plotly.react` would be:
```js
layout.yaxis.range = [5, 10];
// you have to explicitly disable autorange when using `Plotly.react`
// even if you never explicitly enabled it in the first place, since plotly.js
// inserted `yaxis.autorange: true` in your layout object
layout.yaxis.autorange = false;
Plotly.react(gd, data, layout);
```

`clearAxisTypes` is similar: using `Plotly.restyle(gd, 'x', [['a', 'b', 'c']], [0])` will clear the axis types of the axes trace 0 is displayed on, so the autotype algorithm runs again. With `Plotly.react` you'll have to explicitly remove `xaxis.type` if you want to invoke autotype.

I suppose there's a discussion to be had about this: it would certainly be nice to just make the change you want, the same as you do with `relayout`, and have us figure out what we need to do to make it happen. On the other hand, with `Plotly.react` you're saying "this is the new plot state I want rendered", and for us to comb through it and say "Oh, I see you made this change, you must have wanted this other change too" seems heavy-handed and error-prone - what if you really don't want that other change, there's no way to tell us that (as there would be with `relayout` eg `Plotly.relayout(gd, {'yaxis.range': [5, 10], 'yaxis.autorange': true})` (which seems a contrived example, but you might actually do that for example to turn the range from decreasing autorange to increasing autorange).

So rather than bake this into `Plotly.react` as we do with `relayout` et al, I think we should try and find a way to help the user access the recommendations we would make with `impliedEdits` and `clearAxisTypes`, but leave it up to them whether to follow those recommendations.

src/plots/cartesian/graph_interact.js
alexcjohnson: we could... though this one kills the polar draggers too, hence my comment. I originally put the `remove` inside the one `if((!fullLayout._has('cartesian') && !fullLayout._has('gl2d')) || gd._context.staticPlot)` block, and it broke all polar drag and hover effects.

You were saying something about overly broad `selectAll` statements the other day 😏 It wasn't obvious to me how to make this more specific though...

etpinard: > though this one kills the polar draggers too,

Oh nice. Is this tested? If not, would you mind :lock: ing this down?

src/components/annotations/attributes.js
alexcjohnson: `relayout` uses [`containerArrayMatch`](https://github.com/plotly/plotly.js/blob/master/src/plot_api/container_array_match.js) to figure out what edits need to be handled with the special add/remove/edit syntax, and `arraydraw` was just a placeholder for "relayout will handle this, don't worry about it".

Here we're letting the user do their own editing, so there's much less of that complexity. So now I use `editType: 'arraydraw'` as the signal to try and draw components using their `draw` or `drawOne` component methods, hence it has to actually be there!

test/jasmine/tests/plot_api_test.js
```diff
@@ -2171,6 +2171,239 @@ describe('Test plot api', function() {
...
+        function countPlots() {
+            plotCalls = 0;
+
+            gd.on('plotly_afterplot', function() { plotCalls++; });
+            subroutines.layoutStyles.calls.reset();
+        }
+
+        function countCalls(counts) {
+            var callsFinal = Lib.extendFlat({}, counts);
+            // TODO: do we really need to do layoutStyles twice in Plotly.plot?
```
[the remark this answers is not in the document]
alexcjohnson: I suspect this is about automargins - right now the drawing pipeline is like:
- draw everything that can alter margins
- check if anything is too big and needs to change margins
- draw that stuff again if the margins changed
- then draw all the stuff that can't change the margins

we really should refactor it to something like:
- *calculate* the sizes of everything that can alter margins, without actually drawing it
- calculate all the automargins (which determines the positions of everything)
- draw everything

or perhaps:
- draw everything that can alter margins, but don't position it
- calculate all the automargins
- position the predrawn items
- draw everything else

src/components/legend/defaults.js
alexcjohnson: @VeraZab you must not have been dependent on the existence of a `legend` container up to now, as it was unreliable ^^ - `showlegend` is still there, but now it's *reliably* the case that if `_fullLayout.showlegend=false` there will be NO `_fullLayout.legend` container.

src/plot_api/plot_api.js
etpinard: > what if you really don't want that other change, there's no way to tell us that 

Thanks for writing this. This makes :100: % total sense :books: 

>  think we should try and find a way to help the user access the recommendations we would make with impliedEdits and clearAxisTypes, but leave it up to them whether to follow those recommendations.

Perhaps all we need is to (under a certain config option) log whenever we mutate user data or user layout. Alternatively, `Plotly.validate` could be use here e.g.

```js
Plotly.validate([{
  y: [1, 2, 1]
}], {
  xaxis: {
     autorange: true,
     range: [1, 2]
  }
})
```

could log something like: _increasing range detected, range values ignored under autorange_. Then together with https://github.com/plotly/plotly.js/issues/1741, it should make it fairly easy for plotly.js users to debug.

src/components/annotations/attributes.js
etpinard: Very clear. Thanks :books:

src/plots/plots.js
alexcjohnson: good call -> `_skipDefaults` in https://github.com/plotly/plotly.js/pull/2341/commits/1074804c79f23779bb94571283e2954a23d27af6

src/plots/cartesian/graph_interact.js
alexcjohnson: 🔒 https://github.com/plotly/plotly.js/pull/2341/commits/0548db3dd9e5eda6774d7e6f7f58a44c7453aef3

nicolaskruchten:
I don't know enough about the implementation at this point to comment on that, but the approach and API look pretty awesome from the point of view of `react-plotly.js` :)

devtools/test_dashboard/devtools.js
```diff
@@ -222,8 +226,16 @@ function searchMocks(e) {
     });
 }
 
+function getNameFromHash() {
+    return window.location.hash.replace(/^#/, '');
+}
+
+function getResultClass(name) {
+    return 'search-result' + (getNameFromHash() === name ? ' search-result__selected' : '');
+}
```
alexcjohnson: FYI just added this so I could click through the list of mocks and not lose track of which one I'm on... Something I'd wanted to do for a long time but now that I'm trying to load EVERY mock in sequence it became important enough to actually implement.

chriddyp:
This is really great, the design with `datarevision` is really great and Dash users will definitely be able to use.

A couple questions:
1. If the `datarevision` remains the same, can the user omit the data?
i.e.
```
newPlot(gd, [{'x': [1, 2, 3], 'y': [3, 1, 2]}], {'datarevision': 1})
react(gd, [{'marker': {'color': 'blue'}}], {'datarevision': 1})
```
Would that work? If so, then Dash users could potentially omit their data from even being transported if it remained the same.

2. Does `plotly.react` keep the zoom layer or reset it? i.e. if the user zooms in on a map, and then changes a dropdown, will the graph reset or not? I believe that it should (since that is what `newPlot` currently does).

3. Do you imagine that `datarevision` will eventually allow specificity of which trace's data changed?

alexcjohnson:
Re: image diffs - I can't tell why [`gl3d_ribbons`](https://6629-45646037-gh.circle-artifacts.com/0/var/www/streambed/image_server/plotly.js/build/test_images/gl3d_ribbons.png) [changed](https://6629-45646037-gh.circle-artifacts.com/0/var/www/streambed/image_server/plotly.js/build/test_images_diff/diff-gl3d_ribbons.png)... looks like a subpixel change but I don't know what caused it. But I think `gl2d_axes_labels2` was incorrect before and is more correct now ([diff](https://6629-45646037-gh.circle-artifacts.com/0/var/www/streambed/image_server/plotly.js/build/test_images_diff/diff-gl2d_axes_labels2.png) / [new](https://6629-45646037-gh.circle-artifacts.com/0/var/www/streambed/image_server/plotly.js/build/test_images/gl2d_axes_labels2.png)) (it has fill to zero so the autorange should go to zero. The autorange actually goes a little past zero now, rather than ending exactly at zero as it does with regular `scatter`, but the x axis is also a little different in the two cases. Again I'm not sure what I did that changed it though...

src/traces/surface/defaults.js
```diff
@@ -29,12 +29,12 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
 
-    traceOut._xlength = z[0].length;
-    traceOut._ylength = z.length;
-
-    coerce('x');
+    var x = coerce('x');
     coerce('y');
 
+    traceOut._xlength = (Array.isArray(x) && Array.isArray(x[0])) ? z.length : z[0].length;
+    traceOut._ylength = z.length;
```
alexcjohnson: `_xlength` and `_ylength` are used in two places, to try and describe how much of each array to use - one in `surface/convert` (only for the `surface` type) and again in `gl3d/scene` (for all 3d types). The two do things in different ways, and especially since surface can use x/y arrays of either 1 or 2 dimensions, they weren't getting things right. I was trying to do better at handling mismatched arrays but this at least gets it right again for arrays of the correct length (and brings back the original `gl3d_ribbons` mock), and I've added an item for surface to #2353 to explore the mismatched cases more fully.

etpinard:
Awesome stuff :dancer: 

Follow and/or comment on https://github.com/plotly/plotly.js/issues/2353 for more on this topic.

alexcjohnson:
@chriddyp 

> 1. If the `datarevision` remains the same, can the user omit the data?
> i.e.
> ```
> newPlot(gd, [{'x': [1, 2, 3], 'y': [3, 1, 2]}], {'datarevision': 1})
> react(gd, [{'marker': {'color': 'blue'}}], {'datarevision': 1})
> ```
> Would that work? If so, then Dash users could potentially omit their data from even being transported if it remained the same.

Sounds dangerous to me. In some cases this would work, *if* you insert empty arrays for the data (if you don't even do that, we'll see no array vs used-to-be-an-array and do a recalc regardless of `datarevision`). But if anything else in the diff triggers a `recalc` (like, for example, changing the overall plot size, or autoranging an axis), we'll pick up the omitted data anyway and you're hosed.

> 2. Does `plotly.react` keep the zoom layer or reset it? i.e. if the user zooms in on a map, and then changes a dropdown, will the graph reset or not? I believe that it should (since that is what `newPlot` currently does).

`Plotly.react` itself does nothing about the reverse data flow - UI-driven changes to the plot pushing back up into the plot state you're holding onto elsewhere - though since these UI-driven changes mutate `gd.data` and `gd.layout` and *most* of the time these are the same objects you provided with `Plotly.newPlot` (or `Plotly.react`), you may in fact have these changes in your state whether you wanted them or not (probably not with Dash though, as I'm assuming a new plot state coming from Python would generate new `data` and `layout` objects). If you've picked up these changes and included them in the new state you send to `Plotly.react`, then you'll see them persist on the plot. If you haven't, you won't.

> 3. Do you imagine that `datarevision` will eventually allow specificity of which trace's data changed?

Yes, as mentioned in the opening comment:

> Note that right now there's no difference in the pathway we'd take for one vs all data arrays changing, but in the future if we implement partial recalc we can make a similar flag for each array, or perhaps each trace.

etpinard:
cc @VeraZab @nicolaskruchten this will part of `1.34.0` that we'll release on Monday (100% guarantee this time I promise). Sorry for the wait.

src/plot_api/plot_api.js
igorpejic: This has most probably caused:
https://github.com/plotly/plotly.js/issues/2392

PR #2360, opened 2018-02-12
fix errorbars for Plotly.react and for uneven data arrays

alexcjohnson:
Noticed this morning during a discussion with @VeraZab - error bars still filled in `[]` during `supplyDefaults`. When fixing that I also noticed that we weren't allowing uneven `array`/`arrayminus`, not a big deal but if you're doing manual data entry it would be nice to see the two halves of the errorbar show up independently. Ideally this would also remove the cap from the missing end of the bar, as I have it here it just sets the missing end to zero length so that the cap shows up on top of the data point; if anyone can think of a use case for this as a final state of the plot (as opposed to an intermediate state while you're typing the rest of the data) I'd be happy to extend it to omit that cap but if not it seemed like unnecessary extra complexity.

### DIFF
--- a/src/components/errorbars/compute_error.js
+++ b/src/components/errorbars/compute_error.js
@@ -30,18 +30,26 @@ module.exports = function makeComputeError(opts) {
         symmetric = opts.symmetric;
 
     if(type === 'data') {
-        var array = opts.array,
-            arrayminus = opts.arrayminus;
+        var array = opts.array || [];
 
-        if(symmetric || arrayminus === undefined) {
+        if(symmetric) {
             return function computeError(dataPt, index) {
                 var val = +(array[index]);
                 return [val, val];
             };
         }
         else {
+            var arrayminus = opts.arrayminus || [];
             return function computeError(dataPt, index) {
-                return [+arrayminus[index], +array[index]];
+                var val = +array[index];
+                var valMinus = +arrayminus[index];
+                // in case one is present and the other is missing, fill in 0
+                // so we still see the present one. Mostly useful during manual
+                // data entry.
+                if(!isNaN(val) || !isNaN(valMinus)) {
+                    return [valMinus || 0, val || 0];
+                }
+                return [NaN, NaN];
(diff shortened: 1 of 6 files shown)


src/components/errorbars/compute_error.js
```diff
@@ -30,18 +30,26 @@ module.exports = function makeComputeError(opts) {
         symmetric = opts.symmetric;
 
     if(type === 'data') {
-        var array = opts.array,
-            arrayminus = opts.arrayminus;
+        var array = opts.array || [];
 
-        if(symmetric || arrayminus === undefined) {
```
alexcjohnson: Note that this (removing `|| arrayminus === undefined`) isn't actually a change in behavior, since we were filling in `[]` at the `supplyDefaults` step anyway

test/image/mocks/basic_error_bar.json
```diff
@@ -20,6 +20,11 @@
         ],
         "visible": true
       },
+      "error_x": {
+        "type": "data",
+        "symmetric": false,
+        "visible": true
+      },
```
alexcjohnson: added x errorbars with no data so that the image wouldn't change, but it would have failed before this fix in `plot_api_test`

src/components/errorbars/plot.js
```diff
@@ -105,7 +104,9 @@ module.exports = function plot(traces, plotinfo, transitionOpts) {
 
                 yerror.attr('d', path);
             }
+            else yerror.remove();
```
alexcjohnson: another bug fixed here (and tested below) - if you deleted some data from your errorbar data arrays, the corresponding errorbar was not removed.

etpinard: Good catch :fishing_pole_and_fish:

test/image/mocks/basic_error_bar.json
etpinard: thanks for the comment :books:

Issue #2172, opened 2017-11-20
Traces inside rangeslider do not stack vertically for stacked subplots


jackparmer:
Here's an example:

![image](https://user-images.githubusercontent.com/1865834/33000095-ef4a87fa-cd73-11e7-98e4-1f4e5714f233.png)

The rangeslider here should have its own 5 "subplots" so that the traces are ordered vertically in the same way.

This is going to very nice combined with vertical line hover spanning subplots:
https://github.com/plotly/plotly.js/issues/2155

etpinard:
So should it _always_ do that? (not obvious to me)

jackparmer:
> So should it always do that? (not obvious to me)

The only case I can see for not doing it is if the y axis ranges for all of the subplots are the same, so maybe it should be optional.

alexcjohnson:
I feel like the range slider should *always* mirror exactly the y layout of the subplots it's controlling - seems like it would be nonintuitive to see traces overlaid there but not on the main plot.

The only case I can see where you might not want the y *ranges* in the range slider to match what's on the main plot is if we implement autorange to visible data (see https://github.com/plotly/plotly.js/issues/1876#issuecomment-314913229) - then you might want the range slider to keep the full y autorange (because that's the visible data there) even as the main plot zooms in to smaller y ranges.

jumanbar:
I have another example to add, this time with histograms. Consider this arrangement made with the iris dataset (using R & plotly):

![](https://user-images.githubusercontent.com/1039904/51279180-65dbe780-19bb-11e9-9309-26eb3f4a852e.png "Multiple histograms with rangeslider.")

The rangeslider below shows a weird mix between the subplots. In fact, if you examine it closely, it's just the superposition of the three plots, resulting in this kind of things:

![](https://user-images.githubusercontent.com/1039904/51279354-ed295b00-19bb-11e9-9ebd-8ffc86b9bd39.png "Detail of the rangeslider, showing bar-stacking done wrong.")

Compare this to the same data in one stacked histogram:

![](https://user-images.githubusercontent.com/1039904/51279565-66c14900-19bc-11e9-8996-6b737e975100.png "Now the rangeslider is OK, because it just mirrors the main plot.")

IMO, the rangeslider for the multiple stacked-histogram plot should look either like the one in the last example or simply a histogram of the whole dataset with just one color.

Even when there's only 1 color, the histogram shown in the rangeslider should stack the bars of the subplot instead of simply putting one over the others.

[...]

PR #2364, opened 2018-02-14
Rangeslider allow zoom on oppaxis

TomDemulierChevret:
By default we can't currently zoom on the y-axis if we set a rangeslider on the x-axis.
If we write `fixedrange: false` in our y-axis config, it will override the lock put by the rangeslider and allow us to zoom both in x-and y-axis.

The issue is that the rangeslider is kinda broken if we zoom on y-axis since it will not keep the initial range on y-axis (like it does for the x-axis).

This PR :
- maintain the correct y-axis range for data previsualization in the rangeslider even when the user zoom
- add a visual clue to show the user where he has zoomed alongside the y-axis

Here is a set of screenshot to illustrate the issue and the PR :
### Without zoom
no-zoom
### With zoom on y-axis before PR
before-pr
### With zoom on y-axis after PR
after-pr

If you have multiple y-axis and not all of them are zoomable, then you must add `anchor: 'y'` in your x-axis config (with `y` replaced with the name of one of the zoomable y-axis).


etpinard:
Thanks for the PR. Impressive work and clear explanation!

But I have to say I'm not 100% sure the current behavior is incorrect. It might just be a matter of preference. Perhaps we can add an attribute e.g.  `xaxis.rangeslider.anchor: 'fixed' || 'mirror'` (we can probably come up with a better name though :thinking: ) to toggle between the current and your behavior.

TomDemulierChevret:
Just added a new attributes `perpendicularaxesinitialrange` to switch between the new and old behaviour.
I put the default to new behaviour as the old one may not be the most intuitive.

TomDemulierChevret:
Not really sure why the last test is failling, any hint from your side ?

src/components/rangeslider/attributes.js
```diff
@@ -89,5 +89,17 @@ module.exports = {
             'If visible, perpendicular axes will be set to `fixedrange`'
         ].join(' ')
     },
+    perpendicularaxesinitialrange: {
```
etpinard: This is probably a little to verbose for our needs. Attribute name rarely have more than two words smashed together.

The best option that comes to my mind is `xaxis.rangeslider.fixedanchor: false || true`, but `anchor` is used already elsewhere to declare a different thing.

Looking for opinions from @alexcjohnson @cldougl and @chriddyp

etpinard:
> Not really sure why the last test is failling, any hint from your side ?

We've been fighting some intermittent test failures lately. Your latest commit is now :white_check_mark:

src/components/rangeslider/attributes.js
TomDemulierChevret: Yeah, I find it also too verbose but I prefered to use a clear name until we have a better proposal (since anchor is already used).

alexcjohnson:
referencing also https://github.com/plotly/plotly.js/issues/2172 - I love the effect here @TomDemulierChevret, just want to make sure we have a path forward when we add potentially stacked multiple axes. I guess it just means we'd end up with multiple clear and shaded regions stacked up, which should be fine.

src/components/rangeslider/attributes.js
alexcjohnson: given that the range slider is always an x axis, we could use something like `fixedyrange` - and if we ever make a y axis range slider it can get a `fixedxrange` attribute.

TomDemulierChevret: It does makes sense, I was not quite sure rangeslider was for x-axis only.

TomDemulierChevret:
By stacked multiple axes, you're referencing multiple x-axis right ?
Because if you're talking only about multiple y-axis, only one instance of each [clear/shaded] region is added.

src/plots/cartesian/axes.js
```diff
@@ -389,6 +389,10 @@ axes.saveRangeInitial = function(gd, overwrite) {
             ax._rangeInitial = ax.range.slice();
             hasOneAxisChanged = true;
         }
+        // store the initial range for the rangeslider if we zoom on oppaxis
+        if((isNew && ax.fixedrange === false) || (overwrite && hasChanged)) {
+            ax._rangesliderInitialRange = ax.range.slice();
```
alexcjohnson: Hmm, one issue with this: `_rangesliderInitialRange` isn't part of the figure JSON... so if you zoom in on the y axis, then you decide you like the plot and you save it, or you snapshot it even, the rangeslider y axis will not have the same range in the snapshot or when you reopen the plot.

That seems to me to mean that we need a real attribute for the range of each y axis on the rangeslider, and perhaps even an attribute for whether to autorange that y axis on the range slider - it would be cool if the y axis could be told to autorange in the range slider even if you start out zoomed in on the main plot, wouldn't it?

BTW I came to this realization as I was thinking "what tests should we have for this feature", noting that a big part of this is drawing code so it should have an image test, but then realizing as it stands we *can't* make an image test because we have no mechanism to force `_rangesliderInitialRange` to be different from the y axis range in an image test, therefore we can't make the opp axis masks appear there.

TomDemulierChevret: Right, I didn't think about that.
I'll look into it.

TomDemulierChevret:
Added new attributes to the figure JSON as suggested by @alexcjohnson.
It allows to make the rangeslider y-axis work in 3 ways :
- `rangesliderFixedrange: false` => the range is the same than the one displayed on the main plot (old behaviour before the PR)
- `rangesliderFixedrange: true` => the range is fixed and set via `rangesliderRange: [min, max]`
- `rangesliderFixedrange: 'auto'` [default] => the range is fixed and set via the range of the y-axis (or autorange if none provided)

alexcjohnson:
Good call putting the new attributes into the y axis - originally I was thinking that they belonged in the `rangeslider` container, but that wouldn't support multiple y axes correctly. A couple of tweaks I'd suggest:

- We should only coerce these attributes if `letter === y`.
- Instead of `rangesliderFixedrange: (true | false | 'auto')` I might suggest `rangeslidermode: ('fixed' | 'match' | 'auto')` - I think that's a bit more descriptive, and we've been moving away from these mixed-type attributes anyway. Also despite this being javascript, for portability we use only lowercase in attribute names (so `rangesliderRange -> rangesliderrange`). @etpinard opinions about this?
- There may be places (`Axes.expand`?) where we need to check for `rangeslidermode: 'auto'` alongside `autorange: true`, in order to make sure the autorange precursor calculations are done. To test this: make a plot where the y axis is NOT autoranged at the beginning, but it IS supposed to be autoranged in the rangeslider, and see if it gets handled correctly.

I'll try out `plot_api_test` on this branch tomorrow, see if I can figure out what the problem is there.
the image test failed because we had `master` in an improper state that just got fixed late this afternoon - if you merge in master *again*, that one should work again.

src/components/rangeslider/constants.js
```diff
@@ -31,9 +31,13 @@ module.exports = {
     grabAreaMaxClassName: 'rangeslider-grabarea-max',
     handleMaxClassName: 'rangeslider-handle-max',
 
+    maskMinOppAxisClassName: 'rangeslider-mask-min-opp-axis',
+    maskMaxOppAxisClassName: 'rangeslider-mask-max-opp-axis',
+
     // style constants
 
     maskColor: 'rgba(0,0,0,0.4)',
+    maskOppColor: 'rgba(0,0,0,0.2)',
```
etpinard: Just noticed this. Why pick a different opacity value here?

alexcjohnson: I like this effect - you see that the y axis is constrained, but it's not shaded as darkly as the x axis constraint, which aligns with the fact that you can only affect the x axis constraint with the range slider handles.

But a different thought occurred to me on looking at this: this is semitransparent black, so does that mean range sliders don't work on a black background? That's a totally separate issue, doesn't need to be addressed in this PR, but we should look into it.

etpinard:
Hmm. I think people will inevitably confuse `yaxis?.rangesliderrange` with `xaxis?.rangeslider.range`, especially when we implement y-axis (i.e. vertical) range sliders.

Perhaps, it would have been best to make range sliders a root layout component? :rage: 

Oh well, for now I can think of 3 options:

```js
//  option 1
xaxis: {
  rangeslider: {
     // ...
   }
},
yaxis: {
   // finding a good concise name for this thing is very hard
   oppositerangeslider: {
      mode: '  ',
      range: [/* */],
   }
}

// option 2
xaxis: {
  rangeslider: {
     yrangemode: ' ',
     yrange: [/* */],
     y2rangemode: ' ',
     y2range: [/* */],
     // ...
  }
},

// option 3 - which would be more similar to other 
// attributes with axis counters
xaxis: {
   rangeslider: {
     rangemodey: ' ',
     rangey: [/* */],
     rangemodey2: ' ',
     rangey2: [/* */],
     // ....
}
```

After writing the 3 options down, I'd say I prefer option 3 by far. It keeps range sliders attributes contained in their _base_ x-axis object. While axis counters (e.g. 'y', 'y2', 'y3', ...)  aren't great, they are already a pretty big part of the plotly.js api. Thoughts?

src/components/rangeslider/constants.js
TomDemulierChevret: Indeed, that's what I wanted to show with the 2 different values of opacity.

And yeah, the black backgrond will be an issue.

alexcjohnson:
> After writing the 3 options down, I'd say I prefer option 3 by far.

I was hoping to avoid something like that, but I just realized there's a use case where we *need* to do it this way. Kinda convoluted, but imagine you have two side-by-side x axes sharing a y axis, each x axis has its own range slider, and you want different range settings for the y axes in the two rangesliders. Would be weird, as the y axes on the main plots are one axis so they *must* share a range... but I guess I can imagine wanting to restrict the range of the y axes in each range slider to just the data on the respective subplot.

This is not possible with `yaxis.rangesliderrange` - one way or another the attribute path must include both axis ids/names.

Option 3 is pretty good, but let me add one more to the mix:
```js
xaxis: {
  rangeslider: {
    yaxis: {range: [/* */], rangemode: ' '},
    yaxis2: {range: [/* */], rangemode: ' '},
  }
}
```

alexcjohnson:
(@TomDemulierChevret sorry for the back-and-forth here, attribute specs can be really tricky to get right, especially when we start thinking of all the possible edge cases involved)

etpinard:
@alexcjohnson oh right, true. A given y axis can be linked to multiple x range sliders in general. Nice catch!



> Option 3 is pretty good, but let me add one more to the mix:

I like it, but I still prefer my option 3 by a slight margin (mainly due to possible confusion between `layout.yaxis` and `layout.xaxis.rangeslider.yaxis`), but since we might have to add more _opposite_ axis attributes down the road (e.g. grid line style in the range slider panel), your solution should scale better. 

We could also do something like:

```js
xaxis: {
  rangeslider: {
      oppaxis: {
          rangemode: ' ',
          range: [/* */]
       },
       oppaxis2: {

       }
   }
}
```

that way, the same `oppaxis?` containers could be used in future `yaxis.rangeslider` containers?


But really, maybe we should thinking about pulling `rangeslider` out of the axis containers entirely, wouldn't something like this be easier to manage and scale:

```js
layout: {
  xaxis: {},
  yaxis: {},
  yaxis2: {},

  rangesliders: [{
    xref: 'x',
    yref: ['y', 'y2'],

    xaxis: {
      range: [/* */], // equivalent to current `xaxis.rangeslider.range`
      /// style attributes ...
    },
    yaxis: { 
      rangemode: ' ',
      range: [/* */]
    }
  }]
```

alexcjohnson:
re: `xaxis.rangeslider.oppaxis` vs `xaxis.rangeslider.yaxis`: `oppaxis` seems ambiguous to me - what if the y axes are `yaxis2` and `yaxis4` - are those `oppaxis` and `oppaxis2` or are they `oppaxis2` and `oppaxis4`? If we use actually the same names then it's clear that `xaxis.rangeslider.yaxis2` means "the xaxis rangeslider's representation of yaxis2."

re: `xaxis.rangeslider` vs `rangesliders[0].xref`: either way this is a separate issue, we wouldn't change that in this PR. But it also seems to me like that would make it harder to find the rangeslider for the axis you're looking at, would make it possible to define two rangesliders for one x axis, which I don't think we want, and would complicate the simple case (adding a rangeslider to your one x axis - why should it need to go in an array?).

etpinard:
> oppaxis seems ambiguous to me -

Yeah true, that was a long shot. :+1: 



> we wouldn't change that in this PR.

Absolutely. I was just thinking out loud for possible (v2) api improvement.

---------

Ok, so I guess it comes down to:

```js
// option 3
xaxis: {
   rangeslider: {
     rangemodey: ' ',
     rangey: [/* */],
     rangemodey2: ' ',
     rangey2: [/* */],
     // ....
}
```

vs

```js
// option AJ
xaxis: {
  rangeslider: {
    yaxis: {range: [/* */], rangemode: ' '},
    yaxis2: {range: [/* */], rangemode: ' '},
  }
}
```

Thinking for the future when we may allow per-axis range-slider specific style settings e.g. `xaxis.rangeslider.yaxis.gridcolor: 'red'`, I think option AJ has to upper hand. But perhaps @cldougl could help us pick a winner?

TomDemulierChevret:
@alexcjohnson No problem, better get it right now than having to come back to it later.

@etpinard From a user point of view, I find the 'AJ' option much clearer to read and write.

cldougl:
definitely pro: 
```
xaxis: {
  rangeslider: {
    yaxis: {range: [/* */], rangemode: ' '},
    yaxis2: {range: [/* */], rangemode: ' '},
  }
}
```

 (`rangemodey2` is kind of hard for me to read and I think the nested option looks a little cleaner)

alexcjohnson:
Thanks @cldougl - OK @TomDemulierChevret I think we have consensus. Now the question is how to know which y axes to `coerce` inside `rangeslider`... I don't see a pre-made function that does exactly this (kinda unfortunate, @etpinard am I missing something?), but it seems like passing [`subplots.cartesian`](https://github.com/plotly/plotly.js/blob/ce5bc971770daaf3ed60d374f23a8af1bb61fe0b/src/plots/cartesian/layout_defaults.js#L70) to [`Axes.findSubplotsWithAxis`](https://github.com/plotly/plotly.js/blob/ce5bc971770daaf3ed60d374f23a8af1bb61fe0b/src/plots/cartesian/axes.js#L1710), then for each one stripping off the x part, then calling [`Axes.id2name`](https://github.com/plotly/plotly.js/blob/ce5bc971770daaf3ed60d374f23a8af1bb61fe0b/src/plots/cartesian/axes.js#L41) on it - that should give the right set of names.

We also have to be careful how we specify this in the `attributes` file, I guess it'll be a container `yaxis` inside `rangeslider.attributes` but we need to do something to make it clear to users that this means any y axis name. @etpinard do we have a precedent for this?

etpinard:
> I don't see a pre-made function that does exactly this (kinda unfortunate, @etpinard am I missing something

I can't find anything either.



> but we need to do something to make it clear to users that this means any y axis name. @etpinard do we have a precedent for this?

Currently only subplot containers (e.g. `scene`, `geo`, ...) and cartesian axes support counter suffixes. 

There's a meta key in the attribute declaration called `_isSubplotObj`:

image


that makes `Plotly.validate` tests like this [one](https://github.com/plotly/plotly.js/blob/ce5bc971770daaf3ed60d374f23a8af1bb61fe0b/test/jasmine/tests/validate_test.js#L301-L334) pass. Too bad it's called `_isSubplotObj` something like `_supportsCounterSuffixes` would be best. Oh well.

TomDemulierChevret:
This last version should implement what has been said above.
Need to pass the tests now. 😄

TomDemulierChevret:
Hum, not sure why the last tests are failling.
Any idea ?

alexcjohnson:
@TomDemulierChevret looking at the image tests, it seems like we lost the second y axis data in the range sliders. To see what failed and how, you can look in the `artifacts` tab and dig in to find `test_images_diff` which has only the failed images shown as diffs, then find those images in `test_images` and compare with the ones in the repo's `test/image/baseline` folder - eg compare https://7347-45646037-gh.circle-artifacts.com/0/var/www/streambed/image_server/plotly.js/build/test_images/range_slider_axes_double.png to https://github.com/plotly/plotly.js/blob/master/test/image/baselines/range_slider_axes_double.png

In the jasmine tests, there's one explicit failure (do you see this when you run locally?) and a bunch of timeouts. The timeouts are probably because we forgot to catch errors in the promise chain in these tests - notice the async ones end in just `.then(done);`, compare this to the more robust version eg [here](https://github.com/plotly/plotly.js/blob/f8e7ee4086b1b0920d2fec07a717d9fcc619ff1f/test/jasmine/tests/plot_interact_test.js#L120-L121) where we use `assets/fail_test` - that should turn the timeouts into tracebacks you can see and fix.

TomDemulierChevret:
Ok, I didn't know about the artifacts.
And yes I had the same timeout locally.

I'll look into both of these issue.

TomDemulierChevret:
Both issue seem to be fixed now (other yaxis traces are correctly drawn and the jasmie issue wath about a bad html element reference).

src/components/rangeslider/range_attributes.js
```diff
@@ -0,0 +1,40 @@
...
+            {valType: 'any', editType: 'calc'},
+            {valType: 'any', editType: 'calc'}
+        ],
+        editType: 'calc',
+        description: [
+            'Sets the range of this axis for the rangeslider.'
+        ].join(' ')
+    },
+    editType: 'calc'
+};
```
alexcjohnson: @etpinard help me out here - what's the right way to get these attributes into the schema? Just attach them as `rangeslider.attributes.yaxis`? How then do we document that `yaxis` is a stand-in for "all the y axis names relevant to this x axis"?

src/components/rangeslider/defaults.js
```diff
@@ -38,6 +44,31 @@ module.exports = function handleDefaults(layoutIn, layoutOut, axName) {
     coerce('autorange', !axOut.isValidRange(containerIn.range));
     coerce('range');
 
+    var subplots = layoutOut._subplots;
+    if(subplots) {
+        var yIds = subplots.yaxis;
+        var yNames = Lib.simpleMap(yIds, axisIds.id2name);
```
alexcjohnson: This is a bit tricky actually - we don't want *all* y axes, we only want the ones that apply to this x axis. It *might* be sufficient to look at `subplots.cartesian`, filtering on the x axis ID and taking the y axis IDs (note that you can't do this with `subplotId.indexOf(xId) !== -1` because `x` is a valid ID - you have to do something like `subplotId.substr(0, subplotId.indexOf('y')) === xId`, like [here](https://github.com/plotly/plotly.js/blob/113b3618732f28d1a2158771544bf4e60e00c084/src/plots/grid.js#L353))

I say *might* because I'm not sure what happens with rangesliders now if we have overlaying x axes... that would be very weird and confusing, so I think I'd be OK ignoring the y data for these axes, but we should confirm that.

src/components/rangeslider/defaults.js
```diff
@@ -38,6 +44,31 @@ module.exports = function handleDefaults(layoutIn, layoutOut, axName) {
...
+        for(var i = 0; i < yNames.length; i++) {
+            var yName = yNames[i];
+            if(!containerIn[yName]) {
+                containerIn[yName] = {};
+            }
+            if(!containerOut[yName]) {
+                containerOut[yName] = {};
+            }
```
alexcjohnson: Do we *need* to push `{}` back to `containerIn`? I know we have many examples to the contrary 🙈 but we try not to modify the inputs any more than we have to. Also at this point we know `containerOut[yName]` doesn't exist yet. So ideally this section would look something like (would require changing `coerceRange` but should actually make it simpler):
```js
rangeContainerIn = containerIn[yName] || {};
rangeContainerOut = containerOut[yName] = {};
```

src/components/rangeslider/defaults.js
TomDemulierChevret: Couldn't we just use : `layoutOut[yName].anchor === axisIds.name2id(axName)` in the `for` loop ?

src/components/rangeslider/range_attributes.js
alexcjohnson: Oh sorry, you partially answered this [before](https://github.com/plotly/plotly.js/pull/2364#issuecomment-367424841)
> There's a meta key in the attribute declaration called `_isSubplotObj`

So we'll need a `Plotly.validate` test with eg `xaxis: {rangeslider: {yaxis2: {...}}}`, but this still doesn't answer the documentation question, how users are expected to know to use the respective y axis names here.

src/components/rangeslider/defaults.js
alexcjohnson: In principle you can have a y axis shared across several x axes, or positioned as a free axis, and not anchored to the one with the range slider, so unfortunately I don't think that will work.

src/plots/cartesian/axes.js
```diff
@@ -362,10 +363,30 @@ axes.doAutoRange = function(ax) {
...
+            switch(axeRangeOpts.rangemode) {
+                case 'auto':
+                    axeRangeOpts.range = hasDeps ? axes.getAutoRange(ax) : ax.range.slice();
+                    break;
+                case 'match':
+                    axeRangeOpts.range = ax.range.slice();
+                    break;
+                default:
+                    axeRangeOpts.range = ax.range.slice();
+                    break;
```
alexcjohnson: `case 'match'` and `default` are identical, is that expected? If so this can probably be condensed to a simple `if(rangemode === 'auto')`. Regardless, we should have a jasmine test for each of the three `rangemode` cases.

test/jasmine/tests/range_slider_test.js
```diff
@@ -90,12 +91,14 @@ describe('the range slider', function() {
...
                 var maskMin = children[2],
-                    handleMin = children[5];
+                    handleMin = children[7];
 
                 expect(gd.layout.xaxis.range).toBeCloseToArray([4, 49], -0.5);
                 expect(maskMin.getAttribute('width')).toEqual(String(diff));
                 expect(handleMin.getAttribute('transform')).toBe('translate(' + (diff - 2.5) + ',0.5)');
-            }).then(done);
+            })
+            .catch(failTest)
```
alexcjohnson: Thanks for adding these!

src/plots/cartesian/axes.js
TomDemulierChevret: Indeed, I forgot to refactor it down to a simple if (initialy there was more case).

src/components/rangeslider/defaults.js
TomDemulierChevret: Indeed, it is not required anymore (it was in one of the first version).

src/components/rangeslider/range_attributes.js
TomDemulierChevret: I'm not really sure on how to use this metakey, neither how this validate test works.

src/plots/cartesian/axes.js
```diff
@@ -362,10 +363,23 @@ axes.doAutoRange = function(ax) {
...
+        var axeRangeOpts = ax._anchorAxis.rangeslider[ax._name];
+        if(axeRangeOpts) {
+            if(axeRangeOpts.rangemode === 'auto') {
+                axeRangeOpts.range = hasDeps ? axes.getAutoRange(ax) : ax.range.slice();
+            } else {
+                axeRangeOpts.range = ax.range.slice();
+            }
+        }
+        axIn = ax._anchorAxis._input;
+        axIn.rangeslider[ax._name] = Object.assign({}, axeRangeOpts);
```
etpinard: Please use `Lib.extendFlat` instead of `Object.assign` which isn't IE-compatible.

src/components/rangeslider/range_attributes.js
etpinard: No worries @TomDemulierChevret , I'll pull down your branch and see what I can do :ok_hand:

src/components/rangeslider/defaults.js
```diff
@@ -38,6 +44,34 @@ module.exports = function handleDefaults(layoutIn, layoutOut, axName) {
...
+                return subplotId.substr(subplotId.indexOf('y'), subplotId.length);
+            });
+        var yNames = Lib.simpleMap(yIds, axisIds.id2name);
+        for(var i = 0; i < yNames.length; i++) {
+            var yName = yNames[i];
+
+            var rangeContainerIn = containerIn[yName] || {};
+            var rangeContainerOut = containerOut[yName] = {};
+
+            if(rangeContainerIn.range && layoutOut[yName].isValidRange(rangeContainerIn.range)) {
```
etpinard: Hmm. The default `rangemode` value should be `'match'` to match (:laughing: ) the current behavior. We could change that in `v2`, but in the meantime we'll have to make sure this PR doesn't induce a breaking change,

src/plots/cartesian/axes.js
TomDemulierChevret: Indeed, completly forgot this since I'm mainly using Typescript (or ES2017 with Babel/polyfills).

src/components/rangeslider/defaults.js
TomDemulierChevret: Done.

src/plots/cartesian/axes.js
TomDemulierChevret: What kind of jasmine test do you want for the 3 rangemode value ?

src/components/rangeslider/defaults.js
```diff
@@ -38,6 +44,32 @@ module.exports = function handleDefaults(layoutIn, layoutOut, axName) {
...
+                coerceRange(rangeContainerIn, rangeContainerOut, 'rangemode', 'fixed');
+            } else {
+                coerceRange(rangeContainerIn, rangeContainerOut, 'rangemode');
+            }
+
+            coerceRange(rangeContainerIn, rangeContainerOut, 'range', layoutOut[yName].range.slice());
```
etpinard: We should only coerce `xaxis?.rangeslider.yaxis?.range` when `rangemode: 'fixed'`. For `'auto'`, the range value in punched in during `Axes.doAutoRange`. For `'match'`, the range value isn't used.

If it helps you out, I would write this block as:

```js
var rangemodeDflt;
if(rangeContainerIn.range && layoutOut[yName].isValidRange(rangeContainerIn.range)) {
     rangemodeDflt = 'fixed';
}

var rangeMode = coerceRange(rangeContainerIn, rangeContainerOut, 'rangemode', rangemodeDflt);
if(rangeMode === 'fixed') {
     coerceRange(rangeContainerIn, rangeContainerOut, 'range');
}
```

src/components/rangeslider/draw.js
```diff
@@ -136,21 +137,31 @@ module.exports = function(gd) {
 
         opts._rl = [range0, range1];
 
+        if(oppAxisRangeOpts.rangemode === 'auto' || oppAxisRangeOpts.rangemode === 'fixed') {
```
etpinard: I'd vote for using `oppAxisRangeOpts.rangemode !== 'match'` instead here and in all the other places below would make thing more readable. Not a big deal though.

src/components/rangeslider/range_attributes.js
```diff
@@ -0,0 +1,40 @@
...
+        dflt: 'match',
+        role: 'style',
+        editType: 'calc',
+        description: [
+            'Determines whether or not the range of this axis in',
+            'the rangeslider use the same value than in the main plot',
+            'when zooming in/out.',
+            'If *auto*, the autorange will be used.',
+            'If *fixed*, the `range` is used.',
+            'If *match*, the current range is used.'
```
etpinard: the current range _of the corresponding y-axis on the main subplot_ is used - would make things clearer.

src/components/rangeslider/range_attributes.js
```diff
@@ -0,0 +1,40 @@
...
+        ].join(' ')
+    },
+    range: {
+        valType: 'info_array',
+        role: 'style',
+        items: [
+            {valType: 'any', editType: 'calc'},
+            {valType: 'any', editType: 'calc'}
+        ],
+        editType: 'calc',
```
etpinard: Post https://github.com/plotly/plotly.js/issues/2385, I think the`editType`s for these new attributes can be relaxed to `'plot'`.

src/components/rangeslider/defaults.js
TomDemulierChevret: Well, I have an issue with that :
- the rangeslider `drawis s called twice on the first render
- the second call have the wrong `axisOpts` object, the range in `rangeslider.yaxis` is not initialized by the `doAutoRange` method (which is called only once before the first `draw` call)
- it then obviously crashes when trying to access range members on this second render call

Since I've no idea where this second draw call comes from (not from my side, I've check and relayout are done before the second draw call), I've limited coerce to `rangeMode !== 'match'`.

alexcjohnson: @TomDemulierChevret I'll take a look at this, and try to push a few last additions like a new test image. I think the double-draw is the same thing I was fighting with last night https://github.com/plotly/plotly.js/pull/2451#discussion_r172742441

TomDemulierChevret: I would be glad if you could add the test image, I've some issue currently with my docker config which prevent me from doing it on my work station.

alexcjohnson: Yep, both @etpinard and I are working on parts of this, trying to get it into v1.35 which we're hoping to release this afternoon.

etpinard:
@TomDemulierChevret would if be ok if we start pushing a few commits to this branch directly? We're very close to merging your PR. Thanks again for all your efforts!

TomDemulierChevret:
Yes of course, no problem.

test/jasmine/tests/validate_test.js
```diff
@@ -505,4 +505,25 @@ describe('Plotly.validate', function() {
...
+                    yaxis2: { rangemode: 'fixed' },
+                    yaxis3: { range: [0, 1] }
+                }
+            },
+            yaxis: {},
+            yaxis2: {},
+            yaxis3: {}
+        });
+
+        expect(out).toBeUndefined();
```
etpinard: As promised in https://github.com/plotly/plotly.js/pull/2364/files#r171957082

Please note, the `_isSubplotObj` flag in the range slider opposite axis attributes is very important. Without it, this test would spit out:

image

that is `Plotly.validate` would think that `rangeslider.yaxis2` and `rangeslider.yaxis3` aren't part of the schema.

etpinard: Note that this appears to work in `plotly.py` as well.

Copying the new `plot-schema.json` over [there](https://github.com/plotly/plotly.py/blob/master/plotly/package_data/default-schema.json) gives:

image

i.e. does not fail like:

image

cc @cldougl @Kully

src/plots/cartesian/axis_defaults.js
```diff
@@ -49,6 +49,12 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
 
     var autoRange = coerce('autorange', !containerOut.isValidRange(containerIn.range));
 
+    // both x and y axes may need autorange done just for the range slider's purposes
+    // the logic is complicated to figure this out later, particularly for y axes since
+    // the settings can be spread out in the x axes... so instead we'll collect them
+    // during supplyDefaults
+    containerOut._rangesliderAutorange = false;
```
etpinard: Very clear comment. Thanks!

src/components/rangeslider/defaults.js
```diff
@@ -60,15 +60,18 @@ module.exports = function handleDefaults(layoutIn, layoutOut, axName) {
...
             var rangemodeDflt;
-            if(rangeContainerIn.range && layoutOut[yName].isValidRange(rangeContainerIn.range)) {
+            if(rangeContainerIn.range && yAxOut.isValidRange(rangeContainerIn.range)) {
                 rangemodeDflt = 'fixed';
             }
 
             var rangeMode = coerceRange(rangeContainerIn, rangeContainerOut, 'rangemode', rangemodeDflt);
             if(rangeMode !== 'match') {
-                coerceRange(rangeContainerIn, rangeContainerOut, 'range', layoutOut[yName].range.slice());
+                coerceRange(rangeContainerIn, rangeContainerOut, 'range', yAxOut.range.slice());
```
etpinard: Ok. Looks like `rangeslider.yaxis?.range` is still getting coerced when `rangemode: 'auto'` though this appears standard for subplot `xaxis` and `yaxis` `range` under `autorange: true`, so this sounds fine :ok_hand: . 

That said, it would be nice in the future to make:

```js
Plotly.validate([], {
  xaxis: {
    autorange: true,
    range: [0, 1]
  }
})
```

spit out an `unused` log message for `xaxis.range`

test/image/mocks/range_slider_rangemode.json
```diff
@@ -0,0 +1,30 @@
...
+      "anchor": "y2",
+      "domain": [0.58, 1],
+      "range": [1.5, 3.5],
+      "rangeslider": {
+        "range": [-2, 4],
+        "yaxis2": {"rangemode": "fixed", "range": [0, 10]}
+      },
+      "title": "Rangeslider Y2 rangemode fixed"
+    },
+    "yaxis": {"type": "log", "range": [2, 6], "title": "Y explicit range"},
```
etpinard: nice mock!

src/components/rangeslider/oppaxis_attributes.js
```diff
@@ -0,0 +1,45 @@
...
+    // but this is the flag we use to denote attributes that
+    // support yaxis, yaxis2, yaxis3, ... counters
+    _isSubplotObj: true,
+
+    rangemode: {
+        valType: 'enumerated',
+        values: ['auto', 'fixed', 'match'],
+        dflt: 'match',
+        role: 'style',
+        editType: 'calc',
```
etpinard: @alexcjohnson we should be able to change these to `editType: 'plot'`, correct?

alexcjohnson: oh hmm... actually I don't think we can, it's like `ax.autorange` which is still `calc`, because this affects `_rangesliderAutorange` which determines whether we even bother calculating `ax._min/_max` during the `calc` step.

test/jasmine/tests/range_slider_test.js
```diff
@@ -11,716 +11,907 @@ var createGraphDiv = require('../assets/create_graph_div');
...
+            return Plotly.relayout(gd, 'xaxis.rangeslider.yaxis.rangemode', 'match');
+        })
+        .then(function() {
+            expect(gd.layout.xaxis.rangeslider.yaxis).toEqual({ rangemode: 'match' });
+        })
+        .catch(failTest)
+        .then(done);
+    });
+
+    it('should update rangeslider x/y ranges when data changes even if main axes are not autoranged', function(done) {
```
etpinard: Thanks for testing this case!

src/components/rangeslider/oppaxis_attributes.js
etpinard: Ok. I tried switching them to `'plot'`, but that didn't break any test :confused:

xapple:
So how do we make the range slider keep a constant y axis with the latest version of plotly? What's the configuration?

PR #2365, opened 2018-02-14
Better face normals for mesh3d (take 2)

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1368 (at least it try to).

Take on https://github.com/plotly/plotly.js/pull/1423 including the changes from https://github.com/gl-vis/gl-mesh3d/pull/8 in plotly.js.

Note that this PR does not attempt to resolve https://github.com/plotly/plotly.js/issues/1368#issuecomment-345301984.

By opening a PR, we (i.e. mostly @jackparmer ) hope to test out a few more `mesh3d` graphs with the commits of https://github.com/gl-vis/gl-mesh3d/pull/8 - see https://github.com/gl-vis/gl-mesh3d/pull/8#issuecomment-363236128


alexcjohnson:
@etpinard @dfcreative After playing with this for a while, I think @dfcreative's comment last year was actually 🎯 

> What is the chance that test images had wrong sign all the time due to rendering in Ubuntu?

When I load the Snowden mocks on my machine, both master and this branch look like the new baselines, so lets do this! (with a published version of `gl-mesh3d`)

We still have seriously broken `lightposition` - generally x is left (-) to right (+) and y is bottom (-) to top(+), but z has effects that defy easy categorization other than z=1 being "special", and sometimes x and y flip randomly and they don't seem to have the same scales, ie x=y does not put the light at 45 degrees... but that applies to both `mesh3d` and `surface` and isn't new here - I'll make a new issue for it.

jackparmer:
And he said let there be light!

etpinard:
Has anyone tested some of empet's creations?

Issue #1755, opened 2017-06-02
Memory leak in dynamically updated scatter plot.


adamshapiro0:
Hi,

I'm using scatter plots as part of a real-time data display where I'm updating the data set relatively frequently (large volume of data @ ~5-10Hz). I've noticed a significant growth in memory usage as the plot is updated on the order of 100s of MB over a short span of time. I made a small pen to demonstrate, just generating random data and restyling the plot:

https://codepen.io/adamshapiro0/pen/jwOWgZ

I've found a few things after a bunch of digging and googling:
1. This seems to happen in both Chrome (58.0.3029.96) and Firefox (53.0.3)
2. The tab's overall memory usage _and_ javascript memory usage as reported by Chrome's task manager grow continuously while the tab is visible
3. When another tab is selected, the memory usage drops back down to a reasonable level, but appears to maybe grow slowly (hard to tell)
4. In the Chrome inspector, the heap usage recorded by the profiler appears to drop completely to normal on each garbage collect (i.e., the task manager looks like the usage is growing without bound but the inspector profile and heap snapshots don't match that)

I can't tell if this is an issue with Plotly or a deeper HTML5/v8 issue, but since it seems to affect both browsers I'm leaning towards Plotly.

This might be related to https://github.com/plotly/plotly.js/issues/957 and https://github.com/plotly/plotly.py/issues/487, but it's not quite clear.

Any thoughts?

etpinard:
Interesting. This is the first I head about (possible) memory leakage in an SVG-based trace type (#957 and  plotly/plotly.py#487 identify problems with the WebGL-based `surface` trace type).

I remember @alexcjohnson taking noticing some memory issues in codepen itself than couldn't be replicated in a plain HTML page. @adamshapiro0 have you noticed this leaky behavior outside of codepen?

etpinard:
@adamshapiro0 is using `restyle` in the referenced codepen. I wonder if `extendTraces` has the same apparent leaky tendencies.

adamshapiro0:
(Sorry for the delay - traveling.)

@etpinard, I tried copying the codepen code directly into a local HTML file and I _think_ I'm seeing slow memory growth. It's a bit hard to tell because of variation due to garbage collection, but I watched it for a while and it did at least seem to be growing on average. I can try to play some more to confirm when I have a bit more time.

dfcreative:
If there is memory leak, it is tiny/very hard to catch. 30s memory screening:
[...]

PR #2366, opened 2018-02-14
Break circular refs in old fullData and fullLayout

alexcjohnson:
cc @etpinard @dfcreative This doesn't seem to be the whole story, but it does look like it helps the garbage collector in repeated-restyle workloads like #1755


alexcjohnson:
huh, this causes some problems in the tests... will need to investigate.

alexcjohnson:
hmm, after some more testing I no longer think this is a significant benefit, and the test errors make it clear this was an overly blunt approach anyway - if there *is* a benefit to be had here, it'll be better to target exactly the items known to contain possible circular references, such as `trace._expandedInput` and `trace._fullInput`.

PR #2375, opened 2018-02-16
Relink less stuff

alexcjohnson:
@etpinard a quick change to improve update performance with complicated `customdata`

### DIFF
--- a/src/lib/relink_private.js
+++ b/src/lib/relink_private.js
@@ -43,8 +43,9 @@ module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
             if(k === 'customdata' || k === 'ids') continue;
 
             // recurse into arrays containers
-            for(var j = 0; j < fromVal.length; j++) {
-                if(isPlainObject(fromVal[j]) && isPlainObject(toVal[j]) && (toVal[j] !== fromVal[j])) {
+            var minLen = Math.min(fromVal.length, toVal.length);
+            for(var j = 0; j < minLen; j++) {
+                if((toVal[j] !== fromVal[j]) && isPlainObject(fromVal[j]) && isPlainObject(toVal[j])) {
                     relinkPrivateKeys(toVal[j], fromVal[j]);
                 }
             }
--- a/src/lib/relink_private.js
+++ b/src/lib/relink_private.js
@@ -20,12 +20,9 @@ var isPlainObject = require('./is_plain_object');
  * This prevents deepCopying massive structures like a webgl context.
  */
 module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
-    var keys = Object.keys(fromContainer || {});
-
-    for(var i = 0; i < keys.length; i++) {
-        var k = keys[i],
-            fromVal = fromContainer[k],
-            toVal = toContainer[k];
+    for(var k in fromContainer) {
+        var fromVal = fromContainer[k];
+        var toVal = toContainer[k];
 
         if(toVal === fromVal) {
             continue;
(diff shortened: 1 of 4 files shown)


tasks/util/make_schema.js
```diff
@@ -12,6 +12,7 @@ module.exports = function makeSchema(plotlyPath, schemaPath) {
         // package is annoying and platform-dependent.
         // see https://github.com/tmpvar/jsdom/issues/1782
         w.HTMLCanvasElement.prototype.getContext = function() { return null; };
+        w.URL.createObjectURL = function() { return null; };
```
alexcjohnson: @etpinard `npm run build` works again with this change.

src/lib/relink_private.js
```diff
@@ -37,9 +37,14 @@ module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
         }
         else if(isArray(fromVal) && isArray(toVal) && isPlainObject(fromVal[0])) {
 
+            // filter out data_array items that can contain user objects
+            // most of the time the toVal === fromVal check will catch these early
+            // but if the user makes new ones we also don't want to recurse in.
+            if(k === 'customdata' || k === 'ids') continue;
```
etpinard: For reference, you should probably right down somewhere that values in the `ids` arrays should be strings.

But yeah, :+1: for being safe.

tasks/util/make_schema.js
etpinard: Nicely done!

src/lib/relink_private.js
alexcjohnson: > For reference, we should probably right down somewhere that values in the `ids` arrays should be strings.

I can put that in its attribute `description`, but is it important that these are strings and not numbers? Do they get converted to actual strings during use?

alexcjohnson: > Do they get converted to actual strings during use?

yep: `cd[i].id = String(trace.ids[i]);`

test/jasmine/tests/lib_test.js
```diff
@@ -1984,6 +1984,109 @@ describe('Test lib.js:', function() {
...
+            expect(toContainer.customdata[0]._x).toBeUndefined();
+            expect(toContainer.customdata[0]._y).toBeUndefined();
+            expect(toContainer.ids[0]._i).toBeUndefined();
+        });
+
+        it('ignores any values that are ===', function() {
+            var accesses = 0;
+
+            var obj = {
+                get _x() { accesses++; return 1; },
```
etpinard: Really, this is valid ES5?

src/lib/relink_private.js
etpinard: > I can put that in its attribute description

That would be great!

test/jasmine/tests/lib_test.js
alexcjohnson: Yes, I believe so, though don't use it in `src` as it's not supported by IE or Safari. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

src/lib/relink_private.js
alexcjohnson: comment added in https://github.com/plotly/plotly.js/pull/2375/commits/62e54b70c87f8438d2ebd64fb8db90dc8b926785

etpinard:
Looking great :dancer:

test/jasmine/tests/lib_test.js
alexcjohnson: Oh actually I was reading the browser compatibility table wrong... in fact looks like this basic setter/getter usage is fine even in IE and Safari, though I still feel like we should avoid it as it can  make things confusing. I've only ever used it for testing and debugging. If, for example, a property gets set but you can't figure out *where* it's getting set, attach a setter for that property that throws an error.

Issue #2001, opened 2017-09-13
[BUG] Text is cut off on bars where textposition=outside


RedShift1:
When drawing a bar plot with
`{ textposition: outside }`

the text of the highest values is sometimes cut off:

![plotly_bars_text](https://user-images.githubusercontent.com/3098010/30381763-a0b1197a-989d-11e7-83f4-d81cc58a8b2d.png)

JSFiddle: https://jsfiddle.net/vuqugx8v/

alexcjohnson:
Yes, this is the same issue as #2000 - closing this one, we will make sure the fix handles both scatter and bar traces.

babykick:
Has the bug been Fixed or not yet? I recently use Dash to plot bar chart, the cut off problem still exists.

dfaivre:
In case some one stumbles across this, use the `cliponaxis: false` option (from the above [referenced PR](https://github.com/plotly/plotly.js/pull/2378))

pietersv:
I just ran into a case where setting `cliponaxis: false` just wouldn't work.  One workaround is to explicitly set the yaxis range to `[0,7000]`  but I think more direct is to set the top margin to 20px or something large enough to show the label

**Set top margin to 20px**
![image](https://user-images.githubusercontent.com/10464727/141403212-34aa1de9-a8d0-4425-9c8e-05383ada7233.png)

**Set yaxis range a little wider:**
![image](https://user-images.githubusercontent.com/10464727/141403108-bad7c3d6-1b7a-476b-ac71-6e93ab732fd1.png)

brandonerose:
for a bar chart the only thing that works for me is calculating the max value and multiplying the max by 1.1 to extend the range. Hope this gets fixed.

brandonerose:
I found a useful workaround! If you add the label as an **annotation** instead of texposition outside it will not get cut off :)

emilykl:
special case of https://github.com/plotly/plotly.js/issues/2000

emilykl:
Reopening this one as it best captures the specific issue addressed by #7675

Issue #2000, opened 2017-09-13
Text should contribute to autorange calculations


RedShift1:
When drawing a scatter plot with
`{ mode: 'lines+markers+text', fill: 'tonexty' }`
or
`{ mode: 'lines+markers+text', fill: 'tozeroy' }`

the text of the first point and the last point is cut off:

![plotly_filled](https://user-images.githubusercontent.com/3098010/30363974-8ba93368-9862-11e7-82d2-58fcebd29552.png)

JSFiddle: https://jsfiddle.net/sd7drgph/

alexcjohnson:
[Known issue](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L268) but doesn't look like we have an issue for it - thanks for bringing this up @RedShift1 

I was going to suggest `cliponaxis: false` #1861 (your leftmost text item would overlap the tick labels but at least would be visible) but I notice that has problems with date axes... another bug report coming :)

alexcjohnson:
Merging #2001 -> ensure our solution works for both scatter and bar traces.

etpinard:
This would be a nice addition post https://github.com/plotly/plotly.js/issues/2704

emilykl:
still an issue

PR #2378, opened 2018-02-16
Implement `cliponaxis: false` for bar text

etpinard:
Friday Afternoon Feature. This one was talked about in https://github.com/plotly/plotly.js/issues/2000 and https://github.com/plotly/plotly.js/issues/2001 as one way to solve problems with undesired clipped text nodes in bar traces. Note that this PR does not attempt to make on-graph text contribue to the autorange computations.

cc @alexcjohnson @cldougl (and I think this should make @chriddyp happy too :wink: )


src/components/drawing/index.js
```diff
@@ -100,9 +100,11 @@ drawing.hideOutsideRangePoint = function(d, sel, xa, ya, xcalendar, ycalendar) {
     );
 };
 
-drawing.hideOutsideRangePoints = function(traceGroups, subplot) {
+drawing.hideOutsideRangePoints = function(traceGroups, subplot, selector) {
     if(!subplot._hasClipOnAxisFalse) return;
 
+    selector = selector || '.point,textpoint';
```
alexcjohnson: `'.point,.textpoint'` (you lost a `.`)

alexcjohnson: do we not have an image test that catches this?

etpinard: No. That thing only gets called during pan and zoom. Nice catch. I'll add a pan test to :lock: this down.

etpinard: Fixed and :lock: in -> https://github.com/plotly/plotly.js/pull/2378/commits/b9513a22396454123649581c38903e802bceefa3

alexcjohnson:
Very nice, relieves one of the big pain points until we really do text in autorange. 💃

chriddyp:
YESSSSSSSSSS! Love Friday Afternoon Features :1st_place_medal:

frankhumnoi:
Oo

Issue #2370, opened 2018-02-15
Issue with hover labels


Maio93:
Hello,
I customized the hover event to trigger it on each subplots at the same time. When Plotly creates labels, it plots them in a single group. Is it possible force Plotly to plot each labels separately?

like in this example that uses an older version of Poltly: https://jsfiddle.net/2v2qp0hy/2/
this is my example on codpen: https://codepen.io/maio93/pen/OQxoZx

Thank you

etpinard:
Thanks for the report. 

Looks like things were working fine up until `1.32.0`. Here's a working version with `1.31.2`: https://codepen.io/etpinard/pen/RQLzwq

Maio93:
Perfect, thank you very much

PR #2377, opened 2018-02-16
Fix color clustering

dfcreative:
That fixes #2376, #2374  and #2354.
By accident pushed https://github.com/plotly/plotly.js/commit/41c067a2550cb576924f129002092ffc976bed61 to master, this should fix that.

cc @alexcjohnson @etpinard 

* [x] add baseline images

PR #2379, opened 2018-02-17
Generalize spikelines

### LINKED ISSUE #2336: Regression: spikelines no longer supported on bar charts
I've been trying out the latest release (v1.33.1) and found that spikelines no longer appear on bar-charts. 

This seems to be a regression from the current version I'm using (v1.30.1)

I've made what is hopefully an illustrative demo:
    https://codepen.io/anon/pen/gvMBYR

alexcjohnson:
Fixes #2336 and cleans up our spikelines framework so that each module's `hoverPoints` method can define its own spikelines behavior. Ensures that spikelines work correctly for all cartesian trace types, including `scattergl`, see https://github.com/plotly/plotly.js/pull/2247#issuecomment-364740538


test/image/mocks/gl2d_line_select.json
```diff
@@ -1,14 +0,0 @@
...
-        "x": [1, 2, 3],
-        "y": [1, 2, 1]
-    }],
-    "layout": {
-        "dragmode": "select",
-        "showlegend": false,
-        "width": 400,
-        "height": 400
-    }
-}
```
alexcjohnson: cc @dfcreative - needed so tests will pass, if this PR merges before #2377 you can merge master and put this back in there.

src/components/fx/constants.js
```diff
@@ -9,9 +9,6 @@
 'use strict';
 
 module.exports = {
-    // max pixels away from mouse to allow a point to highlight
-    MAXDIST: 20,
```
alexcjohnson: not using this constant anymore, we just have the defaults in `layout.hoverdistance` and `layout.spikedistance`, with which the users of this constant were in conflict.

src/components/fx/helpers.js
```diff
@@ -62,19 +62,16 @@ exports.getClosest = function getClosest(cd, distfn, pointData) {
...
-// it to see its hover info, so distance is infinite outside.
-// But make distance inside be at least 1/4 MAXDIST, and a little bigger
-// for bigger bars, to prioritize scatter and smaller bars over big bars
-//
-// note that for closest mode, two inbox's will get added in quadrature
-// args are (signed) difference from the two opposite edges
-// count one edge as in, so that over continuous ranges you never get a gap
-exports.inbox = function inbox(v0, v1) {
-    if(v0 * v1 < 0 || v0 === 0) {
-        return constants.MAXDIST * (0.6 - 0.3 / Math.max(3, Math.abs(v0 - v1)));
```
alexcjohnson: This built-in constant return value for `Fx.inbox` was limiting our flexibility. `inbox` is now a much simpler function (so in fact we could just 🔪 but it still makes the code a little more concise I think), just returning `Infinity` outside the box or the `passVal` you give it inside the box.

src/components/fx/hover.js
```diff
@@ -464,47 +492,28 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
         if(!point) return null;
         return {
             xa: point.xa,
             ya: point.ya,
-            x0: point.x0,
-            x1: point.x1,
-            y0: point.y0,
-            y1: point.y1,
+            x: point.xSpike !== undefined ? point.xSpike : (point.x0 + point.x1) / 2,
+            y: point.ySpike !== undefined ? point.ySpike : (point.y0 + point.y1) / 2,
```
alexcjohnson: If a module wants spikes NOT to just be drawn at the midpoint of the box hover labels are attached to, it just needs to specify `xSpike` and/or `ySpike`.

src/components/fx/hover.js
```diff
@@ -653,20 +662,31 @@ function createHoverText(hoverData, opts, gd) {
...
+            var parts = Array.isArray(traceHoverinfo) ? traceHoverinfo : traceHoverinfo.split('+');
+            if(parts.indexOf('all') === -1 &&
+                parts.indexOf(hovermode) === -1) {
+                showCommonLabel = false;
+                break;
+            }
         }
+
+        // xyz labels put all info in their main label, so have no need of a common label
+        if(allHaveZ) showCommonLabel = false;
```
alexcjohnson: This logic about whether to show the common label has gotten a bit more complex... previously it was being implicitly handled by `c0.distance <= opts.hoverdistance`, since most of the cases that shouldn't show a common label (like heatmaps) were also setting a pseudo-distance greater than the distance limit, which was always too convoluted to really understand, and we can't do anymore anyway due to the way we're calling `hoverPoints` separately for spikelines if no points were found in the first round for hover labels. Now it's at least more explicit what's going on, but there might be a way to clean it up even more by perhaps having the `hoverPoints` methods flag points that should or should not get a common label.

src/components/fx/hover.js
```diff
@@ -464,47 +492,28 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
-            var xa = point.xa,
-                ya = point.ya,
-                xpx = xa.c2p(xval),
-                ypx = ya.c2p(yval),
-                dxy = function(point) {
-                    var rad = point.kink,
-                        dx = (point.x1 + point.x0) / 2 - xpx,
-                        dy = (point.y1 + point.y0) / 2 - ypx;
-                    return Math.max(Math.sqrt(dx * dx + dy * dy) - rad, 1 - 3 / rad);
-                };
```
alexcjohnson: We're not replicating the distance function here, because that only works for `scatter` anyhow. Now we're letting each `hoverPoints` module tell us both a `distance` (based on `hovermode`) and a `spikeDistance` (always as if in `closest` mode, but may be calculated differently or omitted if that hover label should not make a spike line)

test/jasmine/tests/box_test.js
```diff
@@ -266,6 +266,7 @@ describe('Test box hover:', function() {
             fig.layout.hovermode = 'x';
             return fig;
         },
+        pos: [215, 200],
```
alexcjohnson: I'm not exactly sure how this happened, but it seems like box points are a little bit greedier now about taking hover events. The behavior feels right to me when playing with it, so I just moved this point over a little so that it's again grabbed by the box as intended.

test/jasmine/tests/scatterpolargl_test.js
```diff
@@ -39,8 +39,8 @@ describe('Test scatterpolargl hover:', function() {
 
     [{
         desc: 'base',
-        nums: 'r: 2.920135\nθ: 138.2489°',
-        name: 'Trial 4'
+        nums: 'r: 3.886013\nθ: 125.2822°',
+        name: 'Trial 3'
```
alexcjohnson: This (and below) is a weird one - as far as I can tell the new point is in fact the closest to the cursor, and the old one is a couple of pixels farther away, but I don't know what was making us pick the other one previously. I will note however that `scattergl` (which this inherits from) and `scatter` use different distance functions (mostly having to do with special handling when you're inside two markers, to prioritize the smaller marker), I didn't try to harmonize that here but we probably should at some point.

alexcjohnson:
@etpinard I had a spurious test failure in a gl test - do we want to allow tests to be marked BOTH `@gl` and `@flaky`?

etpinard:
> do we want to allow tests to be marked BOTH @gl and @flaky?

I don't see a problem here.

src/components/fx/hover.js
```diff
@@ -653,20 +662,31 @@ function createHoverText(hoverData, opts, gd) {
     // show the common label, if any, on the axis
     // never show a common label in array mode,
     // even if sometimes there could be one
-    var showCommonLabel = c0.distance <= opts.hoverdistance &&
-                          (hovermode === 'x' || hovermode === 'y');
+    var showCommonLabel = (
+        (t0 !== undefined) &&
```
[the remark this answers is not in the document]
alexcjohnson: Related to https://github.com/plotly/plotly.js/pull/2379#discussion_r168913340
Previously `c0.distance <= opts.hoverdistance` was strangely being used as a proxy for which trace types should show the common label. I suspect that clause could be removed entirely at this point, in fact... But now that we're not using that, it's important to ensure that there *is* a common label to display, otherwise you can get that little empty arrow in some cases.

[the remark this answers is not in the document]
etpinard: Good to know. Thanks!

src/traces/bar/hover.js
```diff
@@ -49,15 +51,24 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
             return Math.max(thisBarMaxPos(di), di.p + t.bardelta / 2);
         };
 
+    function _positionFn(_minPos, _maxPos) {
+        return Fx.inbox(_minPos - posVal, _maxPos - posVal,
+            maxHoverDistance - Math.min(1, Math.abs(_maxPos - _minPos) / pRangeCalc));
```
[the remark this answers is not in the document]
alexcjohnson: Ah, good call - it was actually wrong -> https://github.com/plotly/plotly.js/pull/2379/commits/ad38f8d4eee8d8cac6dbe6a9e1bf7656f5429c50

alexcjohnson:
> Now, this PR almost fixes #2370

And now it *really* does -> https://github.com/plotly/plotly.js/pull/2379/commits/eb98c9159c639457d2ace818443feb68e9bf330f

alexcjohnson:
> I'd like to see a few words in the spikedistance and hoverdistance attribute descriptions about how they only apply to point-like objects.

Good point. How's https://github.com/plotly/plotly.js/pull/2379/commits/31e4e344fcd55394cc4668a84be2a0fd616ab19f?

etpinard:
Nicely done. Thanks very much :dancer:

Issue #860, opened 2016-08-17
Data array attributes should support Typed arrays (e.g. Float32Array)


joshua-gould:
http://codepen.io/anon/pen/wWOmNQ

monfera:
@joshua-gould Thanks for your report, iirc @etpinard answered but I don't see it here, just wanted to mention that typed arrays are not yet supported except for `pointcloud` (typed array demo [here](http://codepen.io/monfera/pen/BLjJVZ)). 

However there are plans to introduce this option. The approach isn't to just not fail when a typed array is supplied, but also, to handle things efficiently with typed arrays in mind, at least for plots where performance matters. For this reason there's no imminent plan to map typed arrays into untyped arrays just to make it work, because the API might change somewhat, similar to how `pointcloud` has an `xy` typed array attribute instead of a separate `x` and `y` attribute.

etpinard:
Initial support for typed arrays got added for https://github.com/plotly/plotly.js/pull/2388/

Discussion on JSON-serializable typed array data should continue in https://github.com/plotly/plotly.js/issues/1784

Issue #1784, opened 2017-06-14
Allow x,y,z in WebGL plots to accept Float32Array's


jackparmer:
It looks like currently only `pointcloud` accepts Float32Array's:

alexcjohnson:
Interesting, definitely seems doable, but is the numpy buffer format accessible from other languages? It looks like a fairly straightforward encoding, so seems like even if it's not natively available elsewhere we could likely generate it - hopefully just translating headers from whatever *is* natively available.

jackparmer:
Here's an [R package](https://cran.r-project.org/web/packages/RcppCNPy/vignettes/RcppCNPy-intro.pdf) for writing numpy buffers.

etpinard:
Referencing https://github.com/plotly/plotly.js/issues/860

rreusser:
Very loosely related, only because Mikola wrote so much useful stuff that's not fully utilized: https://github.com/scijs/ndarray/issues/18

There's been talk of letting the scijs/ndarray constructor accept a plain object with format {data: [...], shape: [...], stride: [...], offset: ...}. It's a bit annoying/inefficient to pack and unpack those via [ndarray-unpack](https://github.com/scijs/ndarray-unpack) and [ndarray-pack](https://github.com/scijs/ndarray-pack), but it's pretty trivial. Like if a numpy buffer could get unpacked into an scijs ndarray and then sent to plotly as an array of arrays via `ndarray-unpack`. The annoying part is mashalling to/from typed arrays, but maybe the above code handles that.

maartenbreddels:
Let me add my 2c on this.
In the ipywidgets ecosystem we have two levels, the higher level at the widget level is where you define how an object gets serialized into json, which is what happens [here](https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/serialize.py#L95) for instance.

Let me repeat the stripped version here:
```python
def array_to_json(ar, obj=None):
    return {'buffer':memoryview(ar), 'dtype':str(ar.dtype), 'shape':ar.shape}
```

[...]

PR #2388, opened 2018-02-20
Typed arrays support

etpinard:
Resolves https://github.com/plotly/plotly.js/issues/860 (hopefully)

Supporting typed arrays inputs should allow users to build more memory-conscious apps around plotly.js (cc https://github.com/plotly/plotly.js/issues/1784). Moreover, typed arrays will allow us to bypass numeric-object-to-number coercion speeding up the `calc` step for various trace types (see proof-of-concept in https://github.com/plotly/plotly.js/commit/45c2f35ac2332a78147b7238b9a2847115192254).

-----------

Now, a few things to take in considerations that the commits below do not cover:

- Typed arrays are nice drop-in replacements for 1D arrays, but what should we do with 2D data arrays? It would easy to extend support for arrays of typed arrays (e.g `[new Float32Array([1,2,3]), new Float32Array([2,1,3])]`), but that doesn't sound very user friendly. Maybe adding first-class [`ndarray`](https://github.com/scijs/ndarray) support would be worthwhile?
- Should we add support for a JSON-serializable version of typed arrays? Something like:

```js
trace = {
  y: {
     type: 'float32',
     vals: [/* */]
  }
}

// or
trace = {
  yarraytype: 'float32',
  y: [/* ... */]
}
```

wouldn't conflict with the existing api, and should allow us to bring some performance benefit to r/python/dash users that declare their data types.


src/plots/cartesian/set_convert.js
```diff
@@ -405,10 +405,15 @@ module.exports = function setConvert(ax, fullLayout) {
         if(axLetter in trace) {
             arrayIn = trace[axLetter];
             len = trace._length || arrayIn.length;
-            arrayOut = new Array(len);
 
-            for(i = 0; i < len; i++) {
-                arrayOut[i] = ax.d2c(arrayIn[i], 0, cal);
+            if(ax.type === 'linear' && Lib.isTypedArray(arrayIn) && arrayIn.subarray) {
+                arrayOut = arrayIn.subarray(0, len);
```
etpinard: More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray

Bypassing `isNumeric(v) ? Number(v) : BADNAM` coercion saves about ~50ms at 1e4 points for SVG scatter. :racehorse:

jmmease:
@jackparmer Yeah, having this alongside plotly/plotly.py#942 will let us transfer numpy arrays from Python into Plotly.js traces without any data marshaling!  Using the binary ipywidgets protocol for the Python -> JS transfer already reduces the plot time for a 1-million point scattergl plot from ~24 seconds (using `iplot`) to ~4 seconds (using `FigureWidget`). I'm excited to see how far down we'll be able to push that with these changes.

etpinard:
@jmmease Thanks for input! I'm curious though: how does one _transfer_ numpy arrays to JS `Float32Array`s ?

jmmease:
@etpinard When working in the Jupyter Notebook, the [ipywidgets](http://ipywidgets.readthedocs.io/en/stable/index.html) library handles syncronizing back-end Python objects with a front-end JavaScript model (This is done over ZMQ and Websockets, see [here](http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Low%20Level.html)).

The Python -> JavaScript serialization logic has special handling for binary Python buffers (specifically `memoryview`, `bytearray`, and `bytes` objects). As I understand it, these buffers are transferred to the front end in binary form, without any ASCII encoding.

So, during serialization on the Python side, I wrap numpy arrays in a `memoryview` and combine them in a `dict` with the numpy datatype and shape metadata. (see [here](https://github.com/plotly/plotly.py/blob/27d41776bef33f5be1ab7ebf96a3a8dd583fc894/plotly/serializers.py#L15))

Then, during deserialization on the JavaScript side, the binary buffers are passed to the constructor of a Typed array (see [here](https://github.com/plotly/plotly.py/blob/27d41776bef33f5be1ab7ebf96a3a8dd583fc894/js/src/Figure.js#L481)), where the numpy datatype metadata is used to lookup the appropriate TypedArray constructor (See [here](https://github.com/plotly/plotly.py/blob/ipyplotly_integration/js/src/Figure.js#L438)).

Currently, I immediately convert the constructed typed array into a standard array (see [here](https://github.com/plotly/plotly.py/blob/27d41776bef33f5be1ab7ebf96a3a8dd583fc894/js/src/Figure.js#L482)), but with your changes, it looks like we won't need to do this conversion, and will be able to pass the typed arrays directly into Plotly.js. This should make for a really efficient interactive visualization experience for Jupyter Notebook users, especially when used in conjunction with WebGL traces!

jmmease:
@etpinard I really like the idea of a JSON-serializable encoding of typed arrays.  I'd like to find an efficient way to save figures involving large arrays to disk.

In your first example (copied below) what did you picture the value of the `vals` property being?

```
trace = {
  y: {
     type: 'float32',
     vals: [/* */]
  }
}
```

I assume that a JSON list of numbers should be supported, but this wouldn't really offer efficiency gains in terms of storage size and (de)serialization time.  Would it make sense to also support a HEX-string encoding of the typed array buffer?

If there is a `shape` property alongside `type` and `vals` then this same HEX-string approach could also be used to encode multi-dimensional arrays (e.g. by assuming row-major ordering).

etpinard:
Ok, tagging this thing as `reviewable`. 

------

@jmmease I won't attempt to add support for JSON-seriliazable typed arrays in this PR. The tedious 
parts are done (I think), it shouldn't be too hard to decoder (e.g. for HEX-strings like you proposed in https://github.com/plotly/plotly.js/pull/2388#issuecomment-368310394) once this thing is merged. We can continue discussing ways to encode large typed arrays inside JSONs in https://github.com/plotly/plotly.js/issues/1784

src/lib/is_array.js
```diff
@@ -8,15 +8,11 @@
...
- * Return true for arrays, whether they're untyped or not.
- */
-
 // IE9 fallback
 var ab = (typeof ArrayBuffer === 'undefined' || !ArrayBuffer.isView) ?
     {isView: function() { return false; }} :
     ArrayBuffer;
 
-module.exports = function isArray(a) {
+exports.isArrayOrTypedArray = function(a) {
```
alexcjohnson: perusing https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView - there's also `DataView` that passes but looks like it's a little too generic to be meaningful to us as an input. Should we be concerned about this?

src/plots/cartesian/set_convert.js
alexcjohnson: Typed arrays can't hold `undefined`, only `NaN`, right? So do we need to switch to `NaN` as `BADNUM`? Though this has the follow-on problem that `NaN === NaN` doesn't work, so `v === BADNUM` would have to be replaced by `IsNaN(v)` when we already know it's a number and I guess `IsNaN(v) && typeof v === 'number'` if we don't... That's pretty annoying. But if we DON'T, won't we have errors dealing with missing data in typed arrays?

src/plots/cartesian/set_convert.js
```diff
@@ -405,10 +405,15 @@ module.exports = function setConvert(ax, fullLayout) {
...
+            } else {
+                arrayOut = new Array(len);
```
alexcjohnson: In cases like this where we know the length and we know we're dealing with numbers, would we benefit by using typed arrays internally? When they're supported of course - we'd need a helper to revert to `Array` in IE9. And subject to my [question about `NaN`](https://github.com/plotly/plotly.js/pull/2388/commits/45c2f35ac2332a78147b7238b9a2847115192254#r171131109).

For a later time regardless...

src/traces/bar/calc.js
```diff
@@ -11,6 +11,7 @@
 
 var isNumeric = require('fast-isnumeric');
 
+var Lib = require('../../lib');
```
alexcjohnson: you go back and forth between this and `isArrayOrTypedArray = require('../../lib').isArrayOrTypedArray;` - not a big deal but the latter seems marginally preferable to me when you don't need anything else from `Lib`.

src/traces/carpet/map_2d_array.js
```diff
@@ -8,6 +8,8 @@
 
 'use strict';
 
+var isArrayOrTypedArray = require('../../lib').isArrayOrTypedArray;
```
alexcjohnson: this file is `map_2d_array` - and we're not supporting typed 2D arrays, at least not yet, right?

test/jasmine/tests/parcoords_test.js
```diff
@@ -615,7 +615,7 @@ describe('@gl parcoords', function() {
             function restyleDimension(key, setterValue) {
 
                 // array values need to be wrapped in an array; unwrapping here for value comparison
-                var value = Lib.isArray(setterValue) ? setterValue[0] : setterValue;
+                var value = Array.isArray(setterValue) ? setterValue[0] : setterValue;
```
alexcjohnson: Why did you need to do this?

src/plots/cartesian/set_convert.js
```diff
@@ -406,29 +406,33 @@ module.exports = function setConvert(ax, fullLayout) {
...
-            } else {
-                arrayOut = new Array(len);
-
-                for(i = 0; i < len; i++) {
-                    arrayOut[i] = ax.d2c(arrayIn[i], 0, cal);
+            if(Lib.isTypedArray(arrayIn)) {
+                if(len === arrayIn.length) {
+                    return arrayIn;
+                } else if(arrayIn.subarray) {
+                    return arrayIn.subarray(0, len);
```
alexcjohnson: This should work for linear and log axes, but what happens if you feed numeric data to a non-numeric (date or category) axis? Both of those could be real use cases, but both *also* alter the input numbers, so I think they need to bail out to the `d2c` block below.

src/components/colorscale/has_colorscale.js
```diff
@@ -6,24 +6,22 @@
...
     var container = containerStr ?
             Lib.nestedProperty(trace, containerStr).get() || {} :
             trace,
         color = container.color,
         isArrayWithOneNumber = false;
 
-    if(Array.isArray(color)) {
+    if(Lib.isTypedArray(color)) {
+        isArrayWithOneNumber = true;
+    } else if(Array.isArray(color)) {
```
alexcjohnson: again to my question about missing data in typed arrays... here can't we just let these typed arrays drop into the for loop as well?

src/traces/heatmap/has_columns.js
```diff
@@ -6,9 +6,10 @@
...
 */
 
-
 'use strict';
 
+var Lib = require('../../lib');
+
 module.exports = function(trace) {
-    return !Array.isArray(trace.z[0]);
+    return !Lib.isArrayOrTypedArray(trace.z[0]);
```
alexcjohnson: Would it work to provide a 2D array as an array of typed arrays? Not that I want to encourage this, far better for us to provide a solution based on a single typed array...

alexcjohnson: > Would it work to provide a 2D array as an array of typed arrays?

Should have looked at the [next commit](https://github.com/plotly/plotly.js/pull/2388/commits/09d37b670b013e71b607bd99ee010b1c4ddff02e#diff-ccbe72dad26409c70d2a8001ee8134ffR453) before commenting 🏆

src/traces/carpet/map_1d_array.js
```diff
@@ -16,7 +18,7 @@
 module.exports = function mapArray(out, data, func) {
     var i;
 
-    if(!Array.isArray(out)) {
+    if(!isArrayOrTypedArray(out)) {
         // If not an array, make it an array:
         out = [];
     } else if(out.length > data.length) {
```
alexcjohnson: ⚠️ this block has a `.slice` ⚠️

test/jasmine/tests/parcoords_test.js
etpinard: Because `Lib.isArray` is no more. It was replaced by `Lib.isArrayOrTypedArray` and `Lib.isTypedArray`. Here, these parcoord test don't use typed arrays, so good old `Array.isArray` suffice.

src/traces/heatmap/has_columns.js
etpinard: Yep, arrays of typed arrays for should work after this PR.

src/traces/carpet/map_1d_array.js
etpinard: That's ok though:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice

src/traces/carpet/map_2d_array.js
etpinard: Good eyes. Only the inner arrays should be allowed to be typed arrays.

src/traces/carpet/map_1d_array.js
alexcjohnson: Ah great. So this is fine, but I guess 🐎 at some point we could make a helper that uses `slice` on regular arrays and `subarray` on typed arrays, for cases like this where we don't need a copy.

src/plots/cartesian/set_convert.js
etpinard: As far as I know, typed arrays "call" `Number(v)` on each item before "inserting" them into the typed array:

image

so yeah (I think) that means `NaN`s are the only non-number values accepted in typed arrays.

etpinard: ... so yeah `connectgaps: false` is probably broken for `scatter` with typed arrays. Let me check.

src/plots/cartesian/set_convert.js
etpinard: That's an idea. Saving a few `isNumeric` (assuming `isNaN` is faster) calls downstream of the `makeCalcdata` should help a little bit.

src/plots/cartesian/set_convert.js
alexcjohnson: Just for completeness while you're looking into edge cases, `(+/-)Infinity` are also allowed in float typed arrays. Aside from users directly providing `Infinity`, I presume this comes up in gl code when we drop doubles to singles, if the input is too large. I believe `(+/-)Infinity` and `NaN` are the only non-numerics that have representations in standard floating point encodings.

etpinard: To me surprise, `connectgaps: false` is working fine with typed arrays:

image

----------------

Here are all the places we check equality with `BADNUM`:


image

by the looks of it, these are all places that check `gd.calcdata` values (i.e. not in a typed array). Luckily `isNumeric` is used elsewhere (which works for fine detecting `NaN`s in typed arrays`).

I'll test out a few more case this afternoon.

src/lib/is_array.js
etpinard: improved in f0395b5965b037af7d3ac24e04057ba532a35d1b

src/traces/bar/calc.js
etpinard: cleaned up in 9b8382602b0bf0c8e8f436ae24f6701b6684a7a4

src/plots/cartesian/set_convert.js
etpinard: Done and :lock: in 6dd2f69e1489982646a36fc488e9b23b5f5f315d

src/components/colorscale/has_colorscale.js
etpinard: done and :lock: in 306986d83c76d65ddc7a8edd0d463d86ba7d0db5

alexcjohnson:
Excellent work - should be a big step toward higher performance! 💃 🍾

src/plots/cartesian/set_convert.js
etpinard: All right, I couldn't find an example of `BADNUM` breaking things for typed array. 

Replacing `BADNUM`s with `NaN`s (and `isNaN`) might help us be more  _typed-array_ friendly down the road, but as long as we're using arrays of object of our `calcdata` for most things, `BADNUM` should work just fine. 

Note that these two [lines](https://github.com/plotly/plotly.js/blob/d4cb0c43570c7619d68af75da93475822a17871a/src/traces/scattergl/index.js#L65-L66) in scattergl calc aren't necessary (as `Number(BADNUM) // -> NaN`), but it makes things clear that `BADNUM` and typed array aren't the best of friends, so I'll keep it there.

jmmease:
@etpinard I just started playing with these changes and I was wondering if the following behavior is expected.

When I create a scatter (or scattergl) trace using a typed array as the x/y values the trace displays properly.  When I look at the gd.data[0].x property I see a typed array, but when I look at the gd._fullData[0].x attribute I see a standard (non-typed) array.  

For memory efficiency, I was thinking it would be nice for _fullData to store the typed version as well.  I can open a separate issue, just wanted to get your initial take on it.

etpinard:
> When I create a scatter (or scattergl) trace using a typed array as the x/y values the trace displays properly. When I look at the gd.data[0].x property I see a typed array, but when I look at the gd._fullData[0].x attribute I see a standard (non-typed) array.

That shouldn't [happen](https://github.com/plotly/plotly.js/blob/cea4fd44ecdee62574b7a9dee1e95dffcf05b0ed/test/jasmine/tests/axes_test.js#L2548-L2578). Would you mind sharing a reproducible example?

jmmease:
Never mind, I'm not able to reproduce it. Must be a bug somewhere else in my code. Thanks!

Issue #2372, opened 2018-02-15
Scatter Plot Matrix (aka SPLOM) discussion


etpinard:
SPLOMs are coming to plotly.js. 

For the uninitiated, docs on the python api `scatterplotmatrix` figure factory are [here](https://plot.ly/python/scatterplot-matrix/). Seaborn calls it a [pairplot](https://seaborn.pydata.org/generated/seaborn.pairplot.html#seaborn.pairplot). Matlab has [plotmatrix](https://www.mathworks.com/help/matlab/ref/plotmatrix.html) draw function.

Some might say that SPLOMs are already part of plotly.js: all we have to do is generate traces for each combination of variables and plot them on an appropriate axis layout ([example](https://codepen.io/etpinard/pen/Vbzxqa?editors=1010)). 

But, this technique has a few limitations and inconveniences:

- data arrays are duplicated, which impacts performance when the number of variables and/or the data array lengths are large 
- creating the axes layout and correctly linking the splom traces is tedious. Note that the python api exposes a few [tools](https://plot.ly/python/subplots/) to make this smoother, but these aren't available to plotly.js users.
- ....
- _feel free to edit and append this list_

Numerous solutions are available. This issue will attempt to spec out the best one.

cc @dfcreative @alexcjohnson @cldougl @chriddyp

etpinard:
### Solution 1 (aka splom overlord)

Add a new do-it-all `splom` (and possible a `splomgl` too) trace type that generates its own _internal_ scatter traces and its own axes - with an api similar to `parcoords`:

```js
trace = {
  dimensions: [{
     values: [/* */],
     // some scatter style props ...
     // some axis props reused from cartesian axes
  }],
  // some splom-wide options e.g.:
  showdiagonal: true || false,
  showupperhalf: true || false,
  showlowerhalf: true || false,
  direction: 'top-left-to-bottom-right' || 'bottom-left-to-top-right',
  // ...
}

```

etpinard:
### Solution 2 (tooling)

Port `make_subplots` and `append_traces` from the python api in plotly.js ([docs](https://plot.ly/python/subplots/)). For example:

```js
var Plotly = require('plotly.js')

var fields = [
   [/* */],
   [/* */],
   // ...
]

var layout = Plotly.makeSubplots({rows: fields.length, cols: fields.length})
var data = []

for (var i = 0; i < fields.length; i++) {
  for (var j = 0; j < fields.length; j++) {
    var trace = {
[...]

Issue #1894, opened 2017-07-18
Better schema declarations for array attributes that don't scale with coordinate arrays


etpinard:
See https://github.com/plotly/plotly.js/pull/1891 for context.

At the moment, `axis.tickvals`, `axis.ticktext`, `axis.categoryarray`, `colorbar.tickvals` and `colorbar.ticktext` are listed as `valType: 'data_array'`. This is causing issues for `colorbar.tickvals` and `colorbar.ticktext` as routines that use the output of [`PlotSchema.findArrayAttributes`](https://github.com/plotly/plotly.js/blob/f56e07161cb90f729c4bda925f4ddc4f09dcc9f6/src/plot_api/plot_schema.js#L144) are expecting those _data_ arrays to scale with the coordinate arrays.

So, as @alexcjohnson points out, maybe we should tag these _uncorrelated_ data arrays in the schema (e.g. `colorbar.tickvals` would have `uncorrelatedArray: true`) - which would be a lot more robust than the solution in https://github.com/plotly/plotly.js/pull/1891/commits/aa9312313393764f63c5e7076986c45d73fa24a6

But perhaps, we should make these attributes something else then `valType: 'data_array'` instead. For plotly.js, setting those attributes to `valType: 'info_array'` with `freeLength: true` would work, but that might break a few things downstream (e.g. the plotly workspace).

Somewhat related: we should also think about _correlated_ arrays mean for trace with 2D coordinate arrays, so that we can eventually extend `filter` and `sort` transform to 2D data.

rreusser:
A quick grep shows *three* things in plotly.js that depend on `'data_array'`:

- https://github.com/plotly/plotly.js/blob/master/src/lib/coerce.js#L23-L35
- https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_schema.js?utf8=%E2%9C%93#L151-L155
- https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_schema.js?utf8=%E2%9C%93#L368

So seems not tightly coupled within plotly.js. Does changing the field type affect the backend processing? On principle thought, I think it'd be nice if `data_array` meant identically that it has 1-1 correspondence with the trace data.

etpinard:
PR https://github.com/plotly/plotly.js/pull/2399 added _free-length_ info array declarations (to support `layout.grid` and `layout.grid.subplots`).

[...]

PR #2399, opened 2018-02-23
Layout grids

alexcjohnson:
An easy way to make a grid of subplots: `layout.grid` can arrange cartesian subplots either as coupled layouts (x axes span entire columns and y axes span rows) or independent (each grid cell gets a specified x/y subplot; axes are coupled only as specified by the subplots). Other types of subplot (and non-subplot traces that support `domain`) can be placed in the grid too, by specifying `domain: {row, column}`

Created to support SPLOM https://github.com/plotly/plotly.js/issues/2372#issuecomment-366049231, but hopefully its usefulness will be more general than that.


test/image/mocks/grouped_bar.json
```diff
@@ -33,11 +33,6 @@
     "xaxis": {
       "type": "category"
     },
-    "barmode": "group",
-    "categories": [
```
alexcjohnson: maybe this was supposed to be `xaxis.categories`? Anyway it was ignored so 🔪

test/jasmine/tests/gl3d_plot_interact_test.js
```diff
@@ -1020,8 +1020,8 @@ describe('@gl Test gl3d annotations', function() {
         camera.eye = {x: x, y: y, z: z};
         scene.setCamera(camera);
         // need a fairly long delay to let the camera update here
-        // 200 was not robust for me (AJ), 300 seems to be.
-        return delay(300)();
+        // 300 was not robust for me (AJ), 500 seems to be.
+        return delay(500)();
```
alexcjohnson: Should I be worried that I've had to increase this delay twice now? I just ignored this test for a while, but it had been failing for me locally for at least the last 6 months...

src/lib/coerce.js
```diff
@@ -257,19 +257,56 @@ exports.valObjectMeta = {
             'An {array} of plot information.'
         ].join(' '),
         requiredOpts: ['items'],
-        otherOpts: ['dflt', 'freeLength'],
+        // set dimensions=2 for a 2D array
+        // `items` may be a single object instead of an array, in which case
+        // `freeLength` must be true.
```
alexcjohnson: I needed 1D and 2D arbitrary-length `info_array` for specifying axes and/or subplots for the grid. I didn't actually end up using `coerce` with these, since there's validation involved where allowed values depend on both the available axes/subplots and on earlier elements in the same array; but it's still important for documentation, and for `Plotly.react` it's important that these are marked as `info_array` and not `data_array`.

Anyway, they work, in case we find a need in the future... tested below.

src/lib/coerce.js
```diff
@@ -257,19 +257,56 @@ exports.valObjectMeta = {
...
+            // simplified coerce function just for array items
+            function coercePart(v, opts, dflt) {
+                var out;
+                var propPart = {set: function(v) { out = v; }};
+
+                if(dflt === undefined) dflt = opts.dflt;
+
+                exports.valObjectMeta[opts.valType].coerceFunction(v, propPart, dflt, opts);
+
+                return out;
```
alexcjohnson: I had change this section to support non-array `items` - but also for existing `info_array` attributes this should be substantially faster than using regular `coerce` for each of the elements, since we're not constructing fake attribute strings and using them in the multiple `nestedProperty` constructs that `coerce` does.

src/plots/grid.js
```diff
@@ -0,0 +1,409 @@
...
+    },
+    roworder: {
+        valType: 'enumerated',
+        values: ['top to bottom', 'bottom to top'],
+        dflt: 'top to bottom',
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'Is the first row the top or the bottom? Note that columns',
+            'are always enumerated from left to right.'
```
alexcjohnson: until someone from a RTL country asks I guess? 😅

src/plots/grid.js
```diff
@@ -0,0 +1,409 @@
...
+        items: {valType: 'enumerated', values: [counterRegex('xy').toString(), '']},
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'Used for freeform grids, where some axes may be shared across subplots',
+            'but others are not. Each entry should be a cartesian subplot id, like',
+            '*xy* or *x3y2*, or ** to leave that cell empty. You may reuse x axes',
+            'within the same column, and y axes within the same row.',
+            'Non-cartesian subplots and traces that support `domain` can place themselves',
+            'in this grid separately using the `gridcell` attribute.'
```
alexcjohnson: There would be something nice about specifying non-cartesian subplots in this array as well... but there are two reasons I didn't allow that:
- It's easy to specify `'scene2'` and `'geo3'` etc but how do you specify a trace (like pie)? You'd have to use `uid` or something. I feel like unless we can get everything, we should only allow cartesian subplots this way.
- What if you use the `xaxes` and `yaxes` form (or one of the defaults with `pattern`) but you still want to insert something else? Like, say you have a splom but you want to put pies along the diagonal, or a splom showing only above the diagonal and you want to put some unrelated subplots below the diagonal?

Also I should mention it might even be useful to allow cartesian axes to be inserted in the grid separately, like with a `row` or `column` attribute that plucks the appropriate `domain` from the grid. Like if you make a splom (using `xaxes` and `yaxes`) and want something special along the diagonal (box plots for example) - these may want to share a y axis with the grid but have a separate x axis (categorical in that case). We might be able to do this with `overlaying` but that might not work because the axis it's overlaying doesn't actually have a subplot on the diagonal... anyway that seems a kind of hacky solution, would be cleaner to just specify the column.

src/plots/grid.js
```diff
@@ -0,0 +1,409 @@
...
+        editType: 'calc',
+        description: [
+            'If no `subplots`, `xaxes`, or `yaxes` are given but we do have `rows` and `columns`,',
+            'we can generate defaults using consecutive axis IDs, in two ways:',
+            '*coupled* gives one x axis per column and one y axis per row.',
+            '*independent* uses a new xy pair for each cell, left-to-right across each row',
+            'then iterating rows according to `roworder`.'
+        ].join(' ')
+    },
+    gap: {
```
alexcjohnson: I used `xgap` and `ygap` to match `heatmap`, but I felt like it's also useful to have a symmetric `gap` attribute (that's only used to set the defaults for `xgap` and `ygap`) - thoughts? Keep this (and add it to `heatmap`), or remove it and just use `xgap`/`ygap`?

Also, for completeness, in other contexts I've seen either `grout` or `gutter` used for this effect, do we want to stick with `gap` or use one of these?

src/plots/grid.js
etpinard: ... and also allowing multiple grids could be a nice way to generate grids with [multi-colspan](https://github.com/plotly/plotly.py/blob/master/plotly/tests/test_core/test_tools/test_make_subplots.py#L849-L852), multi-rowspan cells and crazy shared axes [pattern](https://github.com/plotly/plotly.py/blob/cb71b64e22c29e91d11862c32dd533c2442b1d1f/plotly/tests/test_core/test_tools/test_make_subplots.py#L1627-L1629).

src/lib/coerce.js
alexcjohnson: These ones are for function arguments, and we know that the functions being called have a maximum of 3 arguments. The `[{}, {}, {}]` form ensures we never go past three args. I suppose it *probably* wouldn't hurt to change it, but I didn't feel like taking that risk as I'm not sure how well tested the edge cases are for these components.

src/plots/grid.js
alexcjohnson: I think multiple overlaying grids is likely to be more confusing than helpful. colspan/rowspan would be easier to understand and use as just that - a way to specify colspan and rowspan within the existing grid, some of them wouldn't even work as a different grid, unless you monkey with `grid.domain`, which would be super confusing, and getting the gaps to line up correctly would be automatic with colspan/rowspan but nontrivial between grids having a different number of rows and columns.

I suppose for my other use case, inserting extra subplots in an otherwise fully coupled layout, we could support providing `xaxes`, `yaxes`, *and* `subplots`.

> we could probably do `pie` (i.e. first pie trace in `gd.data`), `pie2` (2nd pie trace in `gd.data`), `pie3` and similar for other traces that provide their own subplot. But yeah, it's not ideal.

Yeah, my concern with that is order in `gd.data` can be changed, or you can delete or change the type of an earlier trace, and then you'd have to know to update the reference in the grid. Also it's ambiguous, does `pie2` mean the pie in trace 2 or the second pie?

Anyway, it's pretty low-footprint, if we do think of a nice way to do this I don't think it would be a problem to add it and continue supporting the way I have it now.

src/plots/grid.js
alexcjohnson: Sounds good, I'll 🔪 `gap` and keep `xgap` and `ygap`.

The meaning of horizontal/vertical spacing in the python wrapper is different from what I have for `xgap`/`ygap` - the python version seems to use absolute domain fraction (hence the need to divide the default by the number of cols/rows, whereas in my current implementation it's a fraction of the grid period, ie `xgap = xgapPx / (xgapPx + xplotPx)`. That allows me to use a constant value for the default, and I feel like it's easier to use as you can think of how the gap compares to one cell of the grid, rather than how it compares to the whole plot area... do you agree or should I switch to what we did in python?

etpinard: I prefer your implementation :ok_hand:

test/image/mocks/grid_subplot_types.json
```diff
@@ -0,0 +1,32 @@
...
+    {"y": [4, 6, 5], "xaxis": "x2"},
+    {"x": [1, 2, 3], "y": [8, 9, 7], "yaxis": "y2"},
+    {"type": "surface", "z": [[1, 2], [4, 1]], "showscale": false},
+    {"type": "pie", "labels": ["a", "b"], "values": [1, 6], "domain": {"row": 0, "column": 2}},
+    {"type": "scatterternary", "a": [1, 1, 3], "b": [1, 3, 1], "c": [3, 1, 1], "mode": "markers"},
+    {
+      "type": "table",
+      "header": {"values": ["a", "b"]},
+      "cells": {"values": [["c", "d"], ["e", "f"]]},
+      "domain": {"row": 2, "column": 0}
```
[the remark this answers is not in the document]
alexcjohnson: huh, seems like a table rendering issue. Here's what I see in test_dashboard:
table bug
And if I move it a little bit, I can get the other side to (almost) disappear:
screen shot 2018-02-23 at 5 22 26 pm
Lets treat this as a separate issue?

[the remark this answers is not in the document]
etpinard: > Lets treat this as a separate issue?

yep that's fine -> https://github.com/plotly/plotly.js/issues/new

src/plots/domain.js
```diff
@@ -69,4 +71,60 @@ module.exports = function(opts, extra) {
         }),
         editType: opts.editType
     };
+
+    if(!opts.noGridCell) {
```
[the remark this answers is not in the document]
alexcjohnson: `noGridCell` shows up in the domain for the [grid itself](https://github.com/plotly/plotly.js/pull/2399/files#diff-c7482eca8726e0b7849cfc77ffc99d9cR148) - you can't put the grid in itself! 😄

test/image/mocks/grid_subplot_types.json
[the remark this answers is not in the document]
alexcjohnson: -> #2401

test/jasmine/tests/lib_test.js
```diff
@@ -797,6 +797,77 @@ describe('Test lib.js:', function() {
...
+                };
+                expect(coerce({}, {}, attrs, 'x')).toEqual([[1, 4], [7, 10]]);
+                expect(coerce({x: []}, {}, attrs, 'x')).toEqual([[1, 4], [7, 10]]);
+                expect(coerce({x: [[0, 3], [8, 11]]}, {}, attrs, 'x'))
+                    .toEqual([[0, 3], [8, 11]]);
+                expect(coerce({x: [[10, 5, 10], [6], [1, 2, 3]]}, {}, attrs, 'x'))
+                    .toEqual([[1, 5], [6, 10]]);
+            });
+
+            it('supports unbounded 2D freeLength arrays', function() {
```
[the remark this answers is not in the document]
alexcjohnson: Good call, turns out `Plotly.validate` didn't dig into `info_array` - was there any reason for that? Anyway, now it does -> https://github.com/plotly/plotly.js/pull/2399/commits/40dd784546bbc41f8a1e874bd737135ff235cdb9
I had to tweak the behavior there, `''` vs `undefined` - seems reasonable, shouldn't affect behavior but cleans up `validate`.

src/plots/grid.js
alexcjohnson: 🔪 gap and made the defaults (for independent axes, not for coupled) match the python version 0.2 and 0.3 instead of the 0.25 for both I had previously. -> https://github.com/plotly/plotly.js/pull/2399/commits/670bdd59bb696fb5ef6bf0d785a183632451ba8c

src/plots/grid.js
```diff
@@ -0,0 +1,409 @@
...
+var counterRegex = require('../lib/regex').counter;
+var cartesianIdRegex = require('./cartesian/constants').idRegex;
+
+
+var gridAttrs = exports.attributes = {
+    rows: {
+        valType: 'integer',
+        min: 1,
+        role: 'info',
+        editType: 'calc',
```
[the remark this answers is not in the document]
alexcjohnson: Good call - I used `calc` because that's what we had for cartesian `axis.domain` but neither it nor the grid attributes actually needed it -> https://github.com/plotly/plotly.js/pull/2399/commits/6b8d461f05eb5daee02caf9f13d2a7619303b6f5

src/plots/domain.js
```diff
@@ -69,4 +71,60 @@ module.exports = function(opts, extra) {
...
+        out.row = {
+            valType: 'integer',
+            min: 0,
```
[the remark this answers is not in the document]
alexcjohnson: > It might be worth adding a `dflt: 0`, so that `domain: {column: 2}` defaults to subplot `xy2`?

Yeah, makes sense - if you have a grid, by default all subplots should go in it. You can always set explicit `domain.x` and `domain.y` anyhow, this is only setting the defaults for those attributes -> https://github.com/plotly/plotly.js/pull/2399/commits/4b43e35587ea990f24f4598113189069878476a1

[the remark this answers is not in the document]
alexcjohnson: > It might be nice to allow `domain: {row: ''}` to match the 'x'/'x2'/'x3' counter

As discussed offline, I omitted this one (though if you say `''` you'll get `0` now anyway with the change ^^ so the result is the same)

alexcjohnson:
I never did figure out what was causing the failing test, so I just [marked it `@flaky`](https://github.com/plotly/plotly.js/pull/2399/commits/446f9f95d7539c5dc603b362e900688634eb8b0a) and it runs there. Not exactly the intent of `@flaky`, as this test simply fails in the `@gl` suite now on CI (but not locally, for me anyway). I found an issue with the test suite that I *hoped* would show what was going on - [dropped promise errors](https://github.com/plotly/plotly.js/pull/2399/commits/cd6e8232121b819b1618da9de9975faf9f0b3e86) - but no luck there (did lots of testing on the [grid-ci-test branch](https://github.com/plotly/plotly.js/tree/grid-ci-test)), and I noticed that I forgot to update parcoords for [`Plotly.react`](https://github.com/plotly/plotly.js/pull/2399/commits/1aec7f7aee678161b214d095db2eb857fbcdb2a4)... but never could figure out why that test stalls on CircleCI. I hope you don't mind @etpinard 😓

src/traces/parcoords/calc.js
```diff
@@ -15,15 +15,23 @@ var wrap = require('../../lib/gup').wrap;
...
+        calcColorscale(trace, color, 'line', 'c');
     }
 
     return wrap({
         lineColor: color,
         cscale: cscale
     });
 };
+
+function constHalf(len) {
```
etpinard: Thanks for making this more readable albeit less _functional_. :ok_hand:

src/traces/parcoords/defaults.js
```diff
@@ -17,14 +17,20 @@ var handleDomainDefaults = require('../../plots/domain').defaults;
...
-        coerce('line.color', defaultColor);
+    var lineColor = coerce('line.color', defaultColor);
+
+    if(hasColorscale(traceIn, 'line') && Lib.isArray(lineColor)) {
+        if(lineColor.length) {
+            coerce('line.colorscale');
+            colorscaleDefaults(traceIn, traceOut, layout, coerce, {prefix: 'line.', cLetter: 'c'});
+            // TODO: I think it would be better to keep showing lines beyond the last line color
+            // but I'm not sure what color to give these lines - probably black or white
+            // depending on the background color?
```
etpinard: I agree here.

Currently, scatter traces with `marker.color` arrays shorter than the coordinates get black markers - regardless of `plot_bgcolor`. Adding some logic around the background color would be even better.

This probably deserves a new issue.

test/jasmine/tests/validate_test.js
```diff
@@ -440,4 +440,69 @@ describe('Plotly.validate', function() {
...
+        );
+        assertErrorContent(
+            out[3], 'dynamic', 'layout', null,
+            ['grid', 'yaxes', 3], 'grid.yaxes[3]',
+            'In layout, key grid.yaxes[3] (set to \'y4\') got reset to \'\' during defaults.'
+        );
+        assertErrorContent(
+            out[4], 'dynamic', 'layout', null,
+            ['xaxis', 'range', 1], 'xaxis.range[1]',
+            'In layout, key xaxis.range[1] (set to \'lots\') got reset to \'50\' during defaults.'
```
etpinard: Hmm. Why does `xaxis.range[1]` get reset to `'50'`? here? I thought the `xaxis.range[1]` default was 6?

src/plots/domain.js
```diff
@@ -90,6 +91,7 @@ exports.attributes = function(opts, extra) {
         out.column = {
             valType: 'integer',
             min: 0,
+            dflt: 0,
```
etpinard: Great. Can we :lock: this down in a jasmine test?

etpinard:
> so I just marked it @flaky and it runs there. Not exactly the intent of @flaky, as this test simply fails in the @gl suite now on CI

That's how many of `@flaky` tests behave unfortunately. When someone discovers flaky behavior locally, we should really make an effort to fix this. In brief, the `@flaky` tag is reserved for tests that don't behave consistently on CI - which are much harder to fix.

> I hope you don't mind @etpinard sweat

I'm happy with parcoords tests that ensure we don't slice arrays during the defaults.

test/jasmine/tests/validate_test.js
alexcjohnson: It gets `range[0] * 10` https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/set_convert.js#L324

etpinard: ok thanks :ok_hand:

src/lib/coerce.js
etpinard: Good point here about `args` being at-most a three-item array. Let's keep that as it :ok_hand: 

But this has me thinking: should we change attributes like `ticktext` and `tickvals` to free-length `info_array` instead of their current `data_array` val type? That way, we could get rid of this ugly [hack](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/plot_api/plot_schema.js#L180).

src/traces/parcoords/defaults.js
alexcjohnson: Great -> #2405

src/plots/domain.js
alexcjohnson: good call -> https://github.com/plotly/plotly.js/pull/2399/commits/794669bf39accfc5bf3faa394c6c1c0d5d9a6e2e

src/lib/coerce.js
alexcjohnson: > should we change attributes like `ticktext` and `tickvals` to free-length `info_array` instead of their current `data_array` val type?

Hmm, possibly... would certainly be nice to keep `data_array` out of `layout`, but those are potentially long enough (and likely enough to be connected to a data source) to still warrant being called `data_array` so I'm a bit ambivalent about it. We could discuss this in a separate issue.

etpinard: > We could discuss this in a separate issue.

Sounds good :+1:

etpinard: Let's continue that discussion in -> https://github.com/plotly/plotly.js/issues/1894

etpinard:
Great. Looks like all my comments have been addressed. :dancer:

Issue #2354, opened 2018-02-09
scattergl fill - autorange and restyle


alexcjohnson:
`scattergl` doesn't match the `scatter` svg autorange behavior with fill: for example, `fill: 'tozeroy'` like in the `gl2d_axes_labels2` mock should trigger the y axis to autorange to zero, and the x axis to autorange tightly rather than padded, despite the markers.

Also: open the same mock and `Plotly.restyle(gd, {fill: 'none'})` - the fill will not disappear, though if you *add* a fill to a plot that doesn't have one, it *will* appear.

etpinard:
> Also: open the same mock and Plotly.restyle(gd, {fill: 'none'}) - the fill will not disappear, though if you add a fill to a plot that doesn't have one, it will appear.

cc @dfcreative 

-----

`scattergl` probably suffers from other autorange bugs and/or discrepancies.

etpinard:
> Plotly.restyle(gd, {fill: 'none'}) - the fill will not disappear

this one is fixed via https://github.com/plotly/plotly.js/pull/2377. The autorange issues prevails.

etpinard:
@dfcreative I found a pretty bad typo in the scattergl calc step where the marker options container is refered to `options.markers` [here](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/traces/scattergl/index.js#L110), [here](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/traces/scattergl/index.js#L114), [here](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/traces/scattergl/index.js#L115) and [here](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/traces/scattergl/index.js#L121) instead of `options.markers` as set [here](https://github.com/plotly/plotly.js/blob/1256c4f9de37ad68ba82d28a048dbe96bc204c4d/src/traces/scattergl/index.js#L513).

[...]

PR #2404, opened 2018-02-26
On-par autorange for scattergl

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2354 - to be merged into https://github.com/plotly/plotly.js/pull/2388

This gets `scattergl` autorange to parity with `scatter` by reusing its calc-step logic straight-up for data arrays of length less than `1e5`. I made a few more improvements to `Scattergl.calc` that will be highlighted in 0aa0f5e. 

I also made a linting commit 5316c47 mostly renaming `container` -> `gd` making `scattergl/index.js` look a little more like the rest of our codebase. I hope @dfcreative won't mind.


src/traces/scattergl/index.js
```diff
@@ -47,116 +48,73 @@ function calc(gd, trace) {
...
 
-    var x = xaxis.type === 'linear' ? trace.x : xaxis.makeCalcdata(trace, 'x');
-    var y = yaxis.type === 'linear' ? trace.y : yaxis.makeCalcdata(trace, 'y');
-
-    var count = trace._length, i, xx, yy;
+    // we need hi-precision for scatter2d,
+    // regl-scatter2d uses NaNs for bad/missing values
+    //
+    // TODO should this be a Float32Array ??
+    var positions = new Array(count2);
```
etpinard: @dfcreative could we set up `positions` with a `Float32Array` here?

src/traces/scattergl/index.js
```diff
@@ -47,116 +48,73 @@ function calc(gd, trace) {
...
+    var count = trace._length;
+    var count2 = count * 2;
     var stash = {};
+    var i, xx, yy;
 
+    var x = xa.makeCalcdata(trace, 'x');
+    var y = ya.makeCalcdata(trace, 'y');
 
-    var x = xaxis.type === 'linear' ? trace.x : xaxis.makeCalcdata(trace, 'x');
-    var y = yaxis.type === 'linear' ? trace.y : yaxis.makeCalcdata(trace, 'y');
```
etpinard: **IMPORTANT** with typed array support in mind (-> https://github.com/plotly/plotly.js/pull/2388), I made all traces pass through `makeCalcdata` unlike previously where x/y array on linear axes bypassed it. Please note that `makeCalcdata` creates a new array (i.e. `x` isn't the same as `trace.x`), unless `trace.x` is a typed array. So memory-conscious user should switch to using typed arrays.

Note that at `1e6` on my setup, one makeCalcdata` call clocks in at roughly `25ms`. So using typed arrays can save about `50ms`

etpinard: Even with the extra arrays, memory consumption appears fine. I wouldn't having @dfcreative double-check though.

etpinard:
For reference, here's a few benchmarks I took down while working on this PR:

At 1e6 points:

- xy `makeCalcdata`: ~ 50ms (this can be :hocho: by using typed arrays)
- generating the kdtree: ~ 200-300ms (which @dfcreative is hoping to :hocho: soon)
- autorange (the fast/approximative routine): < 5ms
- total first render time: 1200-1500ms (with set ranges or not)

At 1e5-1 points:

- xy `makeCalcdata`: < 5ms
- generating the kdtree: ~20ms
- autorange (the slow but on-par with svg routine): < 100-120ms (can be :hocho: when setting axis ranges)
- total first render time: 300ms

A few more notable benchmarks:

- At 1e4 points, first render time for scattergl is ~150ms (vs ~800ms for svg scatter).
- the slow but on-par with svg autorange routine would clock in at ~800ms at 1e6 points. So yeah, using a fast approximative version for big data array is still the right call.

I used Chrome 64 on a Lenovo t450s with Ubuntu 16.04.

src/traces/scattergl/index.js
alexcjohnson: I guess there are situations we accept now that would have broken previously (but only on linear axes) - the "junk" characters stripped by [`cleanNumber`](https://github.com/plotly/plotly.js/blob/master/src/lib/clean_number.js) - worth a test case?

src/traces/scattergl/index.js
```diff
@@ -47,116 +48,73 @@ function calc(gd, trace) {
...
+        positions[i * 2 + 1] = yy === BADNUM ? NaN : yy;
+    }
 
-    if(!x) {
-        x = Array(count);
-        for(i = 0; i < count; i++) {
-            x[i] = i;
+    if(xa.type === 'log') {
+        for(i = 0; i < count2; i += 2) {
+            positions[i] = xa.d2l(positions[i]);
```
alexcjohnson: 🐎 we've already been through `makeCalcData` so we should be able to do `c2l` here

src/traces/scattergl/index.js
```diff
@@ -47,116 +48,73 @@ function calc(gd, trace) {
...
+    // for graph with very large number of points where it
+    // performs poorly.
+    // In big data case, fake Axes.expand outputs with data bounds,
+    // and an average size for array marker.size inputs.
+    if(count < TOO_MANY_POINTS) {
+        ppad = calcMarkerSize(trace, count);
+        calcAxisExpansion(gd, trace, xa, ya, x, y, ppad);
+    } else {
+        if(markerOptions) {
+            ppad = 2 * (markerOptions.sizeAvg || Math.max(markerOptions.size, 3));
```
alexcjohnson: How did you find `sizeAvg` to work in practice? A little hard to say I guess, since we don't have a lot of real data > 1e5 points to play with... You're right that we don't want to just use `sizeMax`, it's worth accepting a bit of clipping in order to generally have less wasted space, and much of the time the largest point won't be on any edge... just wondering how that balance plays out in practice, whether we would be better off with something like halfway between the average and max.

Anyway perhaps we don't have a god way to answer that question right now. I'll just mention that in case we do want to try and do better later, we could find some heuristics that only add a little bit of computation, like binning data points into top/middle/bottom thirds, and only using the top third for the top padding... maybe even with a smooth weighting of the size based on how far it is from the edge. That would still be far faster than the full calculation but could do a better job reducing wasted space without too much clipping.

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -833,3 +833,103 @@ describe('@gl Test gl2d plots', function() {
...
+                Plotly.newPlot(gd, glFig).then(function() {
+                    glRangeX = gd._fullLayout.xaxis.range;
+                    glRangeY = gd._fullLayout.yaxis.range;
+                })
+                .then(function() {
+                    return Plotly.newPlot(gd, svgFig);
+                })
+                .then(function() {
+                    expect(gd._fullLayout.xaxis.range).toBeCloseToArray(glRangeX, 'x range');
+                    expect(gd._fullLayout.yaxis.range).toBeCloseToArray(glRangeY, 'y range');
```
alexcjohnson: Beautiful test!

src/traces/scattergl/index.js
etpinard: Yes. Good call!

src/traces/scattergl/index.js
dfcreative: Float32 is not enough precision for some timedate and precise linear/etc ranges.

src/traces/scattergl/index.js
```diff
@@ -393,17 +376,21 @@ function sceneOptions(container, subplot, trace, positions) {
...
+        var multiOpacity = Lib.isArrayOrTypedArray(markerOpts.opacity);
+        var multiSize = Lib.isArrayOrTypedArray(markerOpts.size);
+        var multiLineWidth = Lib.isArrayOrTypedArray(markerOpts.line.width);
+
+        var isOpen;
+        if(!multiSymbol) isOpen = OPEN_RE.test(markerOpts.symbol);
 
         // prepare colors
-        if(multiMarker || Array.isArray(markerOpts.color) || Array.isArray(markerOpts.line.color) || Array.isArray(markerOpts.line) || Array.isArray(markerOpts.opacity)) {
+        if(multiSymbol || multiColor || multiLineColor || multiOpacity) {
```
dfcreative: <3

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -834,3 +833,103 @@ describe('@gl Test gl2d plots', function() {
...
+    describe('should return the same value as SVG scatter for ~small~ data', function() {
+        var specs = [
+            {name: 'lines+markers', fig: require('@mocks/gl2d_10.json')},
+            {name: 'bubbles', fig: require('@mocks/gl2d_12.json')},
+            {name: 'line on log axes', fig: require('@mocks/gl2d_14.json')},
+            {name: 'fill to zero', fig: require('@mocks/gl2d_axes_labels2.json')},
+            {name: 'annotations', fig: require('@mocks/gl2d_annotations.json')}
+        ];
+
+        specs.forEach(function(s) {
```
dfcreative: Lovely!

src/traces/scattergl/index.js
etpinard: -> https://github.com/plotly/plotly.js/issues/2417

src/traces/scattergl/index.js
etpinard: Thanks for pointing this out! Done in -> 
98d2407

src/traces/scattergl/index.js
etpinard: Added mock in c15722f.

On master, the new mock would have looked like:

image

etpinard:
@dfcreative @alexcjohnson all comments have been addressed.

dfcreative:
LGTM! :dancer:

alexcjohnson:
💃 from me too! 🎉

Issue #2154, opened 2017-11-13
Hover tooltip positioned wrong in stacked barchart when number of bar is 1


alexbyj:
my macbook screen resolution : 1680x1050
browser: firefox 56.0.2 (64-bit)
reproducible jsfiddle link : https://jsfiddle.net/o9fc44pg/1/
used library : https://cdn.plot.ly/plotly-latest.min.js
Is there any limit of tooltip label character length?
screenshot 
<img width="737" alt="screen shot 2017-11-13 at 9 45 24 am" src="https://user-images.githubusercontent.com/10076271/32734491-cba72d56-c857-11e7-91d8-446aed3fc4e5.png">

alexcjohnson:
Thanks @alexbyj - specifically the problem here is with the centered label: when there isn't enough room to put the label either entirely to the right or entirely to the left of the object it's labeling, we center it; but the text is not positioned properly in this case. This situation doesn't arise very often, which is why it's not well tested I guess.

Probably duplicate of #865

nishantlnmiit:
I am also facing the same issue while using this with shiny ggplot2 library
Please help me out.

Issue #865, opened 2016-08-19
Hover label elements are misplaced on small viewports


eagor:
text itself seems to be aligned properly, but the background rects are overlapping.

![image](https://cloud.githubusercontent.com/assets/2478452/17800390/776f0e18-65eb-11e6-98c4-3117b1b7e6bf.png)

Demo:
http://codepen.io/eagor/pen/RROKKK

eagor:
did some more testing. Looks like it's not long text to be the issue:

http://codepen.io/eagor/pen/dXLNzX

etpinard:
Looks like the issue is with small `width` / `height` viewports in general. 

Thanks for posting!

PR #2407, opened 2018-02-27
Fix text position in centered tooltips #2154

rmoestl:
Suggested fix for #2154. Added test mock seems to cover #865 as well.

Regarding the fix:
- Deleted the lines in `alignHoverText` because it fixes the scenario in question and because I wasn't able to figure out the intent of these four lines of code.
- To figure out the intent, I also tracked down the origin of this piece of code: it was introduced in 2e69310. I studied the change set of this commit and its message, but wasn't able to understand the exact intent either.
- I'm not entirely sure that the deleted lines are relevant for certain kinds of plots. I guess a more experienced contributer is able to make a better judge.

Regarding tests:
- I added a *mock* that reproduces the issue and have noticed that for each mock there's an image as the rendered result. This image is missing in this commit, because I'm not sure how to run the image pixel tests described in [image test README](https://github.com/plotly/plotly.js/blob/master/test/image/README.md) and trigger the required `hover` event.

Regarding related issue #865:
- Added test mock shows the problem when two tooltips overlap.

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -1124,10 +1124,6 @@ function alignHoverText(hoverLabels, rotateLabels) {
             tx2x = txx + alignShift * (d.txwidth + HOVERTEXTPAD),
             offsetX = 0,
             offsetY = d.offset;
-        if(d.anchor === 'middle') {
-            txx -= d.tx2width / 2;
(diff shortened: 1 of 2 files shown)


rmoestl:
Accidentally created pull request in plotly.js main repository.

PR #2411, opened 2018-02-27
fall back for contour labels when there's only one contour

### LINKED ISSUE #2398: Contour plot uncaught ax.dtick error: NaN
When I create a contour plot with my configuration with small data, the contour plot works fine however when I try it with bigger data set, it throw me an "uncaught ax.dtick error: NaN". Here are some example.
[Contour plot with Nan error](https://codepen.io/claudericka/pen/WMKXxO)
[Contour plot without error](https://codepen.io/claudericka/pen/GQBOWZ)

Thanks.

alexcjohnson:
Fixes #2398 - the test case I added is the simplest distillation of the plot in the initial report.

Also: I broke up `Axes.calcTicks` -> new `prepTicks` only finishes setting up the axis object (`tick0`, `dtick`, `_tickround`, and `_tickexponent` - all the things `Axes.tickText` will need), but does not actually calculate all the ticks. This is potentially a noticeable improvement for constraint-type contours, where I set an arbitrarily large `nticks` merely to ensure we get a reasonable number of digits of precision. Carpet can take advantage of this as well in one place.

cc @etpinard

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -742,11 +742,8 @@ function autoShiftMonthBins(binStart, data, dtick, dataMin, calendar) {
 // Ticks and grids
 // ----------------------------------------------------
 
-// calculate the ticks: text, values, positioning
-// if ticks are set to automatic, determine the right values (tick0,dtick)
-// in any case, set tickround to # of digits to round tick labels to,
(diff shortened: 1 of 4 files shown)


etpinard:
Looks good :dancer:

PR #2028, opened 2017-09-21
Histogram edge cases

alexcjohnson:
Fixes #1978 - some edge cases of single-bin histogram handling

cc @etpinard

PR #2413, opened 2018-02-27
Fix yet another histogram edge case

### LINKED ISSUE #2393: Wrong histogram for a certain data point
Instead of two bars, one at 4 and another at 1, there is only one single bar at 4 when I try to plot a histogram for just [4, 1]

[Bug on Codepen](https://codepen.io/Pradhyo/pen/ZrRYqB)

It plots correctly for [5, 1] and some other values I tried.

alexcjohnson:
Fixes #2393 - this bug was actually introduced with #2028 fixing some other histogram edge cases. But in that fix I accidentally conflated single-value (which should use the special logic added there to readjust bins) and single-bin (which can result from `Axes.autoBin` even with multiple distinct values in the data array). This special logic is only triggered with `barmode: 'overlay'` which is why that's all I tested here; it already worked correctly in any other `barmode`.

cc @etpinard

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -568,7 +568,7 @@ axes.autoBin = function(data, ax, nbins, is2d, calendar) {
             start: dataMin - 0.5,
             end: dataMax + 0.5,
             size: 1,
-            _count: dataMax - dataMin + 1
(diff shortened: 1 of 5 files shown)


etpinard:
Fantastic :dancer:

PR #2414, opened 2018-02-27
push hover labels in to the bar group extent in compare mode

### LINKED ISSUE #2350: Better positioning of hover labels for bars in compare hovermode
As reported by @saimaparveen with https://codepen.io/saimaparveen/pen/zRKXge - in `layout.hovermode: 'x'|'y'` the hover labels for bars go to the edge of the "category", which in some cases is confusingly far from any bars. We should move it in to match the actual bars we found.

alexcjohnson:
Fixes #2350 

Adds an extra step to `Bar.setPositions` to record the position extent of each bar group, which includes two loops over the data: one to find the min and max position of the trace group, and a second to calculate the extents. I did it this way because calculating this on the fly in `hoverPoints` is fairly awkward, you either have to loop over traces once for each trace (repetitious) or you could modify the hover data after finishing `hoverPoints`, manipulating all the bar traces in the hover data to match their position axis coordinates (adding type-dependent logic to `Fx.hover` directly)... this way seemed cleaner. There may be ways to move some of these calculations into existing loops in `setPositions` but the way these are set up this seems like it would entail a lot of duplication, and anyway I don't expect this to be a noticeable overhead for bars.

cc @etpinard

### DIFF
--- a/src/traces/bar/hover.js
+++ b/src/traces/bar/hover.js
@@ -128,8 +128,9 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     pointData[sizeLetter + '0'] = pointData[sizeLetter + '1'] = sa.c2p(di[sizeLetter], true);
     pointData[sizeLetter + 'LabelVal'] = size;
 
-    pointData[posLetter + '0'] = pa.c2p(minPos(di), true);
-    pointData[posLetter + '1'] = pa.c2p(maxPos(di), true);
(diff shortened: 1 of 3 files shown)


etpinard:
:+1: for putting the hard parts in `setPositions` as opposed to beefing up `hoverPoints`.

:dancer: :dancer: :dagger: for your third bug fix of the day.

Issue #2331, opened 2018-02-04
Parcoords multiselection


monfera:
Currently, `parcoords` permits the selection of one contiguous range within each dimension (via brushing, or via an API call). We're working on support for multiselections, ie. an OR relation among multiple ranges within a specific dimension. Also, any of the 60+ dimensions could have multiple selections. The goal is to retain as much speed as feasible.

Currently, a promising route for the multi-selection is the use of a pixel-based bitmap (likely via textures) for each dimension. Basically, instead of arbitrary floating-point(*) ranges, filtering would have no higher resolution than what can be selected on the screen, or discerned as lines, ie. the pixel raster. A mechanical interpretation is that brushing would turn pixels `on`/`off`, enabling or disabling lines that go through that pixel. This is sufficient for mouse based interactions, but in theory(**), a floating-point range can be more accurate, allowing subpixel selection via the API.

(*) (**) the 32 bit floating points are effectively 23 bit integers in this regard, which is quite a bit more than the (for example) 12 bit precision (=4096) that characterizes a 4k tall `parcoords`, or the 8-10 bit precision on views of regular height eg. on a dashboard. On the other hand, `parcoords` is an aggregate view where going more accurate than a pixel limit may not be useful even through applying the constraints via the API. So this approach is workable iff the visible raster is deemed sufficient as constraint resolution via the API. (It's not possible to go subpixel level with the mouse, unless with browser zooming or similar.)

The motive is that the vertical height of the parcoords is bounded by reasonable monitor resolutions, and WebGL `gl.MAX_TEXTURE_SIZE` is at least 4096, and `gl.MAX_TEXTURE_IMAGE_UNITS` is at least 8, and each texture pixel can be up to 4 bytes (64 bits) so even in the worst case, there's plenty of room for a full-height 4k monitor display and our almost 64 dimensions. The resulting vertex shader algo is a simple texture lookup, and given that 64 bits make up the texture pixels and we have almost 64 dimensions, it could reduce to simple bit fiddling operations.

There are alternatives, but as this approach is promising and would have fairly bounded runtime (to be tested), and represents no limit on how many ranges can be selected, we should discuss if this is an acceptable solution, before enlisting more involved alternatives.

alexcjohnson:
Would this replace the existing system entirely, or only be used on multirange selection axes while the existing system continues to be used for a single selection range?

[...]

PR #2415, opened 2018-02-27
multiple selections on parcoords axes

monfera:
The branch so far contains the SVG part but not yet the WebGL part. Upcoming commits will have 

- working out the API part (Alex); currently, `constraintrange` is diverted to a new meaning
    -> note: current `attributes.js` specification for `constraintrange` is just a hint for the current structure but not what it'll end up being, e.g. it now assumes two intervals
- if needed, changes to interactions, eg. now a new brushing will make a new section, but perhaps a shift+brush is preferred
- WebGL filtering (Robert)
- added mock for testing multiselect features (Robert)
- updated test cases as needed (mostly dependent on the API choices)
- updated mocks due to pixel level differences and API changes
- breakup of some functions or nested ternary conditions
- solution for the remaining two lint issues

Changes:

- factored out brushing
- replaced d3.svg.brush with our own implementation supporting multiple selections
- eliminated non-brush related logic from axisbrush.js for eventual reusability
- general code improvements in areas that were in attention

Due to the need for supporting very numerous intervals (up to 1000 per axis) and the possibility for up to 60 axes, the SVG solution has to be economical, otherwise thousands of DOM elements would be added. I found that SVG slowness has a nontrivial component that scales with concurrently existing DOM element count, roughly; [other](https://codepen.io/monfera/pen/xXmpbY) 
 [experiments](https://codepen.io/monfera/pen/rLYqWR) show the technique of using a single DOM element for showing a lot of glyphs. The current solution uses two layers of `stroke-dasharray` to represent filter coverage but it's fairly easy to switch to eg. `path` (would need points ie. longer `d` than the current `stroke-dasharray`).

There subselections are not fused during drag interaction but they do get fused on mouse release.

More description to come. Code is liberally commented and glad to add more.


src/traces/parcoords/attributes.js
```diff
@@ -80,8 +80,22 @@ module.exports = {
             valType: 'info_array',
             role: 'info',
             items: [
-                {valType: 'number', editType: 'calc'},
-                {valType: 'number', editType: 'calc'}
+                {
```
monfera: This just shows the current structure, there won't be a limit of two specified intervals.

src/traces/parcoords/axisbrush.js
```diff
@@ -0,0 +1,520 @@
...
+            .on('dragstart', function(d) {
+                var e = d3.event;
+                e.sourceEvent.stopPropagation();
+                var y = d.unitScaleInOrder(d.unitScale.invert(d3.mouse(this)[1]));
+                var unitLocation = d.unitScaleInOrder.invert(y);
+                var b = d.brush;
+                var intData = getInterval(b, d.unitScaleInOrder, y);
+                var unitRange = intData.interval;
+                var pixelRange = unitRange.map(d.unitScaleInOrder);
+                var s = b.svgBrush2;
```
monfera: upcoming commit will remove the `2` post-fix (we used to have both D3 and our own brush for manual regression testing but not anymore)

src/traces/parcoords/parcoords.js
```diff
@@ -218,13 +233,36 @@ function viewModel(model) {
             xScale: xScale,
             x: xScale(i),
             canvasX: xScale(i) * canvasPixelRatio,
-            unitScale: unitScale(height, c.verticalPadding),
+            // fixme remove the old unitScale
```
monfera: `unitScale` is still being used for its inverse mapping due to how `y` values are arriving from the underlying `d3` interaction callback calls, but could in theory be unified, simplifying the current two-step map eg. `var y = d.unitScaleInOrder(d.unitScale.invert(d3.mouse(this)[1]));`

src/traces/parcoords/attributes.js
alexcjohnson: (for my own notes, @monfera don't worry about this unless it's blocking you)

If we merge in master, particularly #2399, we can use 2D and/or unbounded free-length `info_array`.
I'm not sure the best way to handle it though - since the vast majority of cases will still use a single range, I'd like to keep `constraintrange` 1D rather than forcing people to wrap it into 2D. So the two options that occur to me are:
1. make it a length-2N array, ie `[min1, max1, min2, max2, min3, max3...]` - this would be fairly simple for us to work with, and seems easy to *describe* to users, not sure how intuitive it will be for them to use though.
2. make a separate 2D attribute `constraintranges` that we switch to when there are multiple ranges. That would be more plumbing on our side, and we'd have to decide which one wins in a conflict, but the structure may be more intuitive.

@etpinard thoughts?

etpinard: Users should **not** have to make `constraintrange` 2D to set a single constraint. 

I'll like to add a third option, make `constraintrange` 1D or 2D. This would be a first for info arrays, but note that some data arrays already use this pattern (e.g. heatmap `x` and `y`).

monfera: Another aspect is that the functionality is [initially discussed](https://github.com/plotly/plotly.js/issues/2331#issuecomment-364920486) specifically for ordinal dimensions, where it'd be natural to simply list values, eg. `[1, 5, 8, 12]`, and even to mix singular values and ranges, like `[1, [5, 8], 12]` meaning 1, 5 to 8 and 12. It would be possible to support this, but it wouldn't blend well with using an 1D array for a single range, because then an 1D array with two elements could be interpreted as either a single range, or as two discrete values.

test/image/mocks/gl2d_parcoords.json
```diff
@@ -62,6 +62,8 @@
       },
       {
         "label": "Min height width",
+        "range": [200000, -30000],
+        "constraintrange": [0, 100000],
```
alexcjohnson: This is covered by the `gl2d_parcoords_constraints` mock too, but I added a reversed-axis constraint to this during my tests.

test/image/mocks/gl2d_parcoords_constraints.json
```diff
@@ -0,0 +1,178 @@
...
+          0, 0, 0, 0,
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+        ]
+      },
+      {
+        "label": "±[5,10],0",
+        "constraintrange": [[-10, -5], [-1, 1], [5, 10]],
+        "values": [
+          -9, -9, -7, -7,
+          -10.001, 10.001, -4.98, 4.98, -10, 10, -5, 5,
```
alexcjohnson: Nearly every sample on this plot is testing something... the values arrays are broken into lines based on which dimension is being tested. This dimension is testing that precision on the outer bounds of the constraint range is limited by floating point (OK, I could have added another zero or two, `10.00001` or so...) and precision in the gaps between selection ranges is not as good, but this test (`4.98`) shows at least one thousandth of the range. The number of vertical pixels is 2048, and as expected if I move to `4.99` one of these values gets incorrectly marked valid. Notice that for this test to pass, all the lines going to `-0.7` in the preceding and following dimensions should be grey, and all the lines going to `0.7` should be colored.

alexcjohnson: I didn't notice any performance issue with making the filter mask a fixed 2048 height, since it only has a width of 8. In principle we should be able to double that if more precision is required.

alexcjohnson:
@etpinard I need to clean up some syntax and other test issues, but calling this reviewable as tests pass locally. Have a look!

test/jasmine/tests/parcoords_test.js
```diff
@@ -8,7 +8,7 @@ var attributes = require('@src/traces/parcoords/attributes');
 var createGraphDiv = require('../assets/create_graph_div');
 var delay = require('../assets/delay');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
-var fail = require('../assets/fail_test');
```
alexcjohnson: FYI `fail` is a jasmine global, we were overloading it here, so among other things if you *don't* bring in `assets/fail_test` but you thought you had, linters won't catch it. We should probably change that in the rest of our test suites too...

alexcjohnson:
I can’t figure out what’s going on with these tests. Seems like since I `@noCi`’d my new tests and it’s still hanging, perhaps the increased resource needs from multiselect are just reducing the threshold we can run before hanging. I guess I can look for the memory leak and see if fixing that allows more tests to run, or we can break the tests up further - shift more tests to `@flaky`, or even make a totally new tag...

alexcjohnson:
I've reproduced the memory issue - it clearly behaves like a real leak, building up to multiple GB after 1000 or so lines are added in sequence, not dropping after you stop adding points and wait (so it's not just slow garbage collection) until you purge the plot with eg `Plotly.newPlot`.
screen shot 2018-03-16 at 12 11 24 am

Generated using this code - note CPU usage never went above ~20% during this test, but I couldn't leave the console open or the console process would eat 100% CPU and grind to a halt, though the plot would keep going just fine:
```js
var trace = {type: 'parcoords', dimensions: []};
for(var i = 0; i < 5; i++) {
  trace.dimensions.push({values: [Math.random()], label: i, range: [0, 1]});
}
Plotly.newPlot(gd,[trace]);
var add = setInterval(function() {
  trace.dimensions.forEach(function(d) { d.values.push(Math.random()); });
  gd.layout.title = String(trace.dimensions[0].values.length);
  Plotly.redraw(gd);
}, 500);
```
 Will look into this more tomorrow - perhaps looking back at older plotly.js versions to see where the leak was introduced (which is why I used `redraw` instead of `react`).

alexcjohnson:
- [X] memory leak fixed -> https://github.com/plotly/plotly.js/pull/2415/commits/2911ae699a82b173be63be6f241d2f03fb9326a3 and cleaner create/update pattern in https://github.com/plotly/plotly.js/pull/2415/commits/fa1a436daa5621e6dd4addca6752724531136a57
- [ ] click on/near an ordinal value to toggle it (perhaps `cursor: pointer` to indicate this?)

alexcjohnson:
@etpinard check it out, I added click-to-select for ordinal ranges. Should be ready to go.

etpinard:
All right. Well, this is great! We should probably rename this PR _Parcoords 2.0_.

That click-to-select functionality is pretty sweet:

peek 2018-03-19 11-02

I won't pretend to understand more than 10% of the parcoods code (I would appreciate a walkthrough at some point :wink: ), but after some thorough testing, this thing deserves a :dancer:

PR #2416, opened 2018-02-27
Clean up removed axes and axis lines

### LINKED ISSUE #2392: Deleted axes not cleaned up properly
See eg:

```js
Plotly.newPlot(gd,
    [{y: [1,2,3]}, {y: [10, 30, 20], yaxis: 'y2'}],
    {yaxis2: {side: 'right', overlaying: 'y'}}
);
Plotly.deleteTraces(gd, [1]);
```
screen shot 2018-02-21 at 3 00 43 pm

reported in https://github.com/plotly/react-plotly.js/issues/45

### LINKED ISSUE #2410: Showline not cleaned up properly
Similar to #2392 ...

```javascript
Plotly.newPlot(gd,
    [{y: [1,2,3]}, {y: [10, 30, 20]}],
    {yaxis: {showline: true}}
);
Plotly.relayout(gd, {'yaxis.showline': false});
```

Results in the y-axis line still being shown.

Ref https://github.com/plotly/react-plotly.js/issues/53

alexcjohnson:
Fixes #2410 - hide axis lines that have been removed - this one was straightforward, just needed to move the path attr setter out to where it's always called
Fixes #2392 - this one is a bit more complicated, it's hard to figure out a reasonable d3-idiomatic way to do this because it's not obvious where the axis objects that we need to remove had been created in the first place, ie independent or overlaying some other axis. In principle we should refactor the d3 data for drawing axes to reflect this structure, but for now I just added some more explicit cleanup code.

cc @etpinard @nicolaskruchten

### DIFF
--- a/src/plot_api/subroutines.js
+++ b/src/plot_api/subroutines.js
@@ -272,6 +272,7 @@ exports.lsInner = function(gd) {
          *    -----
          *     x2
          */
+        var xPath = 'M0,0';
         if(shouldShowLinesOrTicks(xa, subplot)) {
             leftYLineWidth = findCounterAxisLineWidth(xa, 'left', ya, axList);
             xLinesXLeft = xa._offset - (leftYLineWidth ? (pad + leftYLineWidth) : 0);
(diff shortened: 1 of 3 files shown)


src/plot_api/subroutines.js
```diff
@@ -272,6 +272,7 @@ exports.lsInner = function(gd) {
          *    -----
          *     x2
          */
+        var xPath = 'M0,0';
```
etpinard: maybe a good candidate for a new addition to `src/constants/` (e.g. `blankPath`)?

alexcjohnson: We do use it a fair amount... though we use it for purposes other than just blank (as the start of a real path for example), and between the open `'M0,0'` and closed `M0,0Z` variants (though maybe we don't need the closed one and could switch to open?), I feel like hiding it in a constant may make its purpose less clear rather than more.

src/plots/cartesian/index.js
```diff
@@ -515,6 +524,11 @@ function purgeSubplotLayers(layers, fullLayout) {
     }
 }
 
+function removeSubplotExtras(subplotId, fullLayout) {
+    fullLayout._draggers.selectAll('g.' + subplotId).remove();
+    fullLayout._defs.select('#clip' + fullLayout._uid + subplotId + 'plot').remove();
```
etpinard: What about axis titles? They are removed [above](https://github.com/plotly/plotly.js/blob/315272be116c29c108487537dcf6c0b2f7e7a2c2/src/plots/cartesian/index.js#L296) in an old vs new axis loop, could we combine that in `removeSubplotExtras`?

alexcjohnson: Titles are per axis, because there's only ever one copy per axis, so they're inserted with no reference to the subplot and it would be incorrect to tie their removal to the disappearance of any given subplot.

etpinard:
Thanks for making `1.35.0` a pretty solid release :dancer:

PR #2106, opened 2017-10-19
Hoverlabel.zformat

jdugge:
Add a `zhoverformat` option to `heatmap` and `contour` traces which specifies the number format to use for displaying `z` values.

PR #2422, opened 2018-02-28
Z exponent fix

### LINKED ISSUE #2420: Contour plot "undefined" label when z >= 10 million
When contour z-points are greater than 10 million then the tooltip value is suffixed with "undefined". 

The suffix should be B or a multiple of 10M or it should render with an exponent. It should also be implemented for even larger values (1e8+).
```
var data = [ {
    z: [
        [1e6, 5e6, 1e7, 1.5e7],
        [1.2e6, 5.6e6, 1.1e7, 1.6e7]
    ], type: 'contour'
}];
```
undefined

alexcjohnson:
Fixes #2420 - #2106 improved formatting here but broke the default behavior for z values that need exponents. Fortunately the fix was a one-liner.

cc @etpinard

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -1528,7 +1528,7 @@ function numFormat(v, ax, fmtoverride, hover) {
     if(hover) {
         // make a dummy axis obj to get the auto rounding and exponent
         var ah = {
-            exponentformat: ax.exponentformat,
+            exponentformat: exponentFormat,
             dtick: ax.showexponent === 'none' ? ax.dtick :
                 (isNumeric(v) ? Math.abs(v) || 1 : 1),
             // if not showing any exponents, don't change the exponent
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -502,8 +502,7 @@ describe('hover info', function() {
                     nums: 'x: 1\ny: 3\nz: 2',
                     name: 'two'
                 });
-            })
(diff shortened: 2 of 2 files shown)


src/plots/cartesian/axes.js
```diff
@@ -1528,7 +1528,7 @@ function numFormat(v, ax, fmtoverride, hover) {
     if(hover) {
         // make a dummy axis obj to get the auto rounding and exponent
         var ah = {
-            exponentformat: ax.exponentformat,
+            exponentformat: exponentFormat,
```
etpinard: Nice catch :fishing_pole_and_fish:

test/jasmine/tests/hover_label_test.js
```diff
@@ -550,6 +549,45 @@ describe('hover info', function() {
...
+            })
+            .then(function() {
+                [
+                    [expFmt(MINUS_SIGN + '1.234568', 23), MINUS_SIGN + '10B', MINUS_SIGN + '10k'],
+                    [MINUS_SIGN + '0.01', MINUS_SIGN + '10n', '0'],
+                    [expFmt('1.234568', -23), '10n', '0.01'],
+                    ['123.4568', '12.34568B', expFmt('1', 23)]
+                ]
+                .forEach(function(row, y) {
+                    row.forEach(function(zVal, x) {
```
etpinard: :sparkles: test! Why create 12 test case when we can loop over 12 bricks!

etpinard:
:dancer: nicely done

Issue #808, opened 2016-08-02
Relayout'ing orientation of scrollable legends is off  


etpinard:
See http://codepen.io/etpinard/pen/oLPpkj

In detail:
- going from `legend.orientation` `'v'` to `'h'` does not remove the scroll bar
- going from `legend.orientation` `'v'` to `'h'` does not remove the scroll handlers
- going from `legend.orientation` `'h'` to `'v'` does not add a scroll bar immediately (the scroll bar appears only after attempting to scroll).

Issue #1859, opened 2017-07-07
Updatemenu scroll handle has the wrong size


alexcjohnson:
(Edit: this has been fixed for legend, still an issue for updatemenus)
Look for example at the `legend_scroll` mock: there are 23 traces, and 13 are visible at a time in the legend. So the scroll handle should take up ~13/23 of the legend height. But it looks like it just has a fixed height of 20px (which might be reasonable as a *minimum* in case the legend is absurdly long relative to its viewport, unless the viewport itself is super small, then the minimum might drop to half its height?)
![screen shot 2017-07-07 at 1 37 52 am](https://user-images.githubusercontent.com/2678795/27944474-1991d3b2-62b5-11e7-88ca-01f77b5b7795.png)

I think updatemenu scroll handles do [the same thing](https://github.com/plotly/plotly.js/pull/1214/files#diff-6df5a05bf905f2429d5e0ef121f31d08R62) but have not observed it.

n-riesco:
@alexcjohnson Both legends and updatemenus use a constant-sized scrollbars.

alexcjohnson:
Thanks @n-riesco - It's not super high priority (I called it a bug not because it's broken, but because it's different from basically every other scrollbar out there), but do you want to take this one when you need a break from other work? Should be easy...

n-riesco:
@alexcjohnson sure, I can look into it (as long as this isn't urgent).

alexcjohnson:
#2426 fixed the legends part of this.
Updatemenus *do* have this problem too, so leaving this issue open.

PR #2426, opened 2018-02-28
Legend scroll fix

### LINKED ISSUE #2387: Clicking legend item causes legend to scroll
Taking from https://community.plot.ly/t/clicking-legend-entry-causes-legend-to-scroll/3361

This came up in the Dash workshop with one of the attendee's apps. When they click on a legend item, the legend scrolls before the click is fired, causing the legend click to fire on a different item. Very annoying!

One of the attendees and I recreated an example here: https://codepen.io/chubukov/pen/rJdZdz?editors=0010. I can't get it ti misbehave on my computer, but the attendee can. The attendee claims that it is part of the grouped legend functionality, as this example (without groups) https://codepen.io/chriddyp/pen/RQMBmJ?editors=1111 works fine for them.

Attendee says:
> I wonder if it has something to do with the extra entries per legendgroup. I played with markers vs lines also, don't think that matters.
> It's only clicking in specific spots that triggers the scroll. It seems to be just under the letters that tends to do it. Maybe just when the pointer is about to shift from the hand to the arrow? Not sure exactly, maybe there's some automated way to click in every pixel and see?

alexcjohnson:
Fixes #2387 - don't allow dragging from the scroll box of a legend, only from the scroll bar itself, and even then not on right clicks, only left click. I also made the scroll bar handle a little bit wider so it's a bit easier to grab. And another part of the issue is that the scrollbar would jump to be centered on the mouse cursor, rather than simply moving with it from wherever you grabbed it. I fixed that too.

Fixes the legend part of #1859 - makes the scrollbar scale to show the fraction of the legend that's visible.

cc @etpinard

### DIFF
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -207,6 +207,10 @@ module.exports = function draw(gd) {
     // legend, background and border, scroll box and scroll bar
     Drawing.setTranslate(legend, lx, ly);
 
+    // to be safe, remove previous listeners
+    scrollBar.on('.drag', null);
+    legend.on('wheel', null);
+
     if(opts._height <= legendHeight || gd._context.staticPlot) {
         // if scrollbar should not be shown.
         bg.attr({
@@ -226,6 +230,9 @@ module.exports = function draw(gd) {
         });
 
         scrollBox.call(Drawing.setClipUrl, clipId);
+
+        Drawing.setRect(scrollBar, 0, 0, 0, 0);
+        delete opts._scrollY;
     }
     else {
         var scrollBarHeight = Math.max(constants.scrollBarMinHeight,
@@ -234,9 +241,10 @@ module.exports = function draw(gd) {
             scrollBarHeight -
             2 * constants.scrollBarMargin;
         var scrollBoxYMax = opts._height - legendHeight;
+        var scrollRatio = scrollBarYMax / scrollBoxYMax;
 
-        var scrollBarY = constants.scrollBarMargin;
-        var scrollBoxY = scrollBox.attr('data-scroll') || 0;
+        // scrollBoxY is 0 or a negative number
+        var scrollBoxY = Math.max(opts._scrollY || 0, -scrollBoxYMax);
 
         // increase the background and clip-path width
         // by the scrollbar width and margin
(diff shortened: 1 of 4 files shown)


alexcjohnson:
Since this doesn't show up in test images, here's what the new scrollbars look like - a little wider and a lot taller:
screen shot 2018-02-28 at 4 21 05 pm

test/jasmine/tests/legend_scroll_test.js
```diff
@@ -96,6 +97,69 @@ describe('The legend', function() {
...
+            mouseEvent('mousedown', x, y0, opts);
+            mouseEvent('mousemove', x, y1, opts);
+            mouseEvent('mouseup', x, y1, opts);
+
+            expect(finalDataScroll).not.toBe(initialDataScroll);
+
+            return finalDataScroll;
+        }
+
+        it('should scroll on dragging the scrollbar', function() {
```
etpinard: Are these general legend scrollbox tests and/or tests that :lock: the fix for #2387? (I'm asking as I would be surprised if #2387 was reproducible in a jasmine test).

etpinard:
Probably related to https://github.com/plotly/plotly.js/issues/808. The new variable-height scroll bar doesn't update. For example, from [`legend_scroll`](https://rreusser.github.io/plotly-mock-viewer/#legend_scroll) repeatedly calling `Plotly.deleteTraces(gd, [0])` gives:

peek 2018-02-28 16-47

No need to fix this in this PR though, just confirming that scrollable legends don't update well as first reported in https://github.com/plotly/plotly.js/issues/808

test/jasmine/tests/legend_scroll_test.js
alexcjohnson: Yes, these do test #2387, if I'm interpreting it correctly. What I think was happening in #2387 is that *somehow* the [`scrollBox` drag](https://github.com/plotly/plotly.js/pull/2426/commits/f75f5ba325e7c80d93272b71585d0e31ceac4f48#diff-f48c57dae9095e916f13f364ae740f69L296) was being triggered - and that's what I decided (with no dissent) had no reason ever to happen. So this test shows that regular drag works, then the [next one](https://github.com/plotly/plotly.js/pull/2426/commits/f75f5ba325e7c80d93272b71585d0e31ceac4f48#diff-ed9e861dab1a2e0cbdeacb9c7bd74db1R145) shows that the `scrollBox` won't drag even if it does somehow get the event, and then the [third new test](https://github.com/plotly/plotly.js/pull/2426/commits/f75f5ba325e7c80d93272b71585d0e31ceac4f48#diff-ed9e861dab1a2e0cbdeacb9c7bd74db1R154) shows that right-click (the only form of #2387 I could reproduce on my own computer) even on the `scrollBar` will not trigger drag.

test/jasmine/tests/legend_scroll_test.js
```diff
@@ -100,19 +101,19 @@ describe('The legend', function() {
...
                 2 * constants.scrollBarMargin;
             var initialDataScroll = scrollBox.getAttribute('data-scroll');
             var dy = 50;
             var finalDataScroll = '' + Lib.constrain(initialDataScroll -
                 dy / scrollBarYMax * scrollBoxYMax,
                 -scrollBoxYMax, 0);
 
-            var scrollBarBB = scrollBar.getBoundingClientRect();
-            var y0 = scrollBarBB.top + scrollBarBB.height / 2;
+            var y0 = scrollBarBB.top + scrollBarBB.height / 5;
```
alexcjohnson: notice that I moved the drag point away from the center of the `scrollBar`, thereby testing that it drags from where you grab it rather than snapping to the center.

alexcjohnson:
Fixed #808 (and https://github.com/plotly/plotly.js/pull/2426#issuecomment-369395487) in https://github.com/plotly/plotly.js/pull/2426/commits/569b37893c7082e9ed43740dd87a2d2fc36ce41b

src/components/legend/draw.js
```diff
@@ -234,9 +241,10 @@ module.exports = function draw(gd) {
             scrollBarHeight -
             2 * constants.scrollBarMargin;
         var scrollBoxYMax = opts._height - legendHeight;
+        var scrollRatio = scrollBarYMax / scrollBoxYMax;
 
-        var scrollBarY = constants.scrollBarMargin;
-        var scrollBoxY = scrollBox.attr('data-scroll') || 0;
```
etpinard: Yes. Way better to use `_` then a DOM attr :tada:

src/components/legend/draw.js
```diff
@@ -243,8 +243,7 @@ module.exports = function draw(gd) {
         var scrollBoxYMax = opts._height - legendHeight;
         var scrollRatio = scrollBarYMax / scrollBoxYMax;
 
-        // scrollBoxY is 0 or a negative number
-        var scrollBoxY = Math.max(opts._scrollY || 0, -scrollBoxYMax);
+        var scrollBoxY = Math.min(opts._scrollY || 0, scrollBoxYMax);
```
etpinard: Thank you!

etpinard:
Amazing PR! :dancer:

PR #2429, opened 2018-03-01
ZERO circular dependency 

etpinard:
resolves https://github.com/plotly/plotly.js/issues/236 and closes [milestone/9](https://github.com/plotly/plotly.js/milestone/9).

I started working on this PR when I noticed that the new layout `grid` module from https://github.com/plotly/plotly.js/pull/2399 has a circular dependency:

image

This went unnoticed because some other PR(s) removed one or two circular dependencies. On master we now have:

image

whereas our `test-syntax` script was happy with [16 or less](https://github.com/plotly/plotly.js/blob/113b3618732f28d1a2158771544bf4e60e00c084/tasks/test_syntax.js#L264).

----

So after fixing the `grid` circular dep (that's in commit https://github.com/plotly/plotly.js/commit/bee37dd1ca6f9f96d9c466baa8c1d8a22f7e33d2), I tried to see what it would take to :hocho: all circular dependencies. In brief, we had:

- to register `errorbars` as a component and use `Registry.getComponentMethod` when needed (commit https://github.com/plotly/plotly.js/commit/87275afe4b9db337c3201cbbee629d7f3b80f8bb),
- register the top-level api methods (i.e. the `Plotly[method]` functions) as first proposed in the last serious attempt https://github.com/plotly/plotly.js/pull/2032#issuecomment-332272263

and to my surprise this worked as a charm :four_leaf_clover: :champagne: :tada: 

---------

Looking forward, removing these circular deps isn't just _nice_, it should allow us to trim our bundle size using [`browser-pack-flat`](https://github.com/goto-bus-stop/browser-pack-flat) or even [`tinyify`](https://github.com/browserify/tinyify). Preliminary attempts were positive: I think we'll be able to trim tens of kBs out of our bundles. I'll open up a new PR for this soon. Moreover, there were reports of some dev tools (the new FF ones I think) balking on our source code, this might help.

@alexcjohnson @nicolaskruchten hopefully you'll enjoy this. :stuck_out_tongue:


nicolaskruchten:
Anything that makes bundling easier, faster, smaller :)

src/plots/plots.js
```diff
@@ -19,8 +19,8 @@ var axisIDs = require('../plots/cartesian/axis_ids');
 var Lib = require('../lib');
 var _ = Lib._;
 var Color = require('../components/color');
+var Grid = require('../components/grid');
```
alexcjohnson: Do you want to go the last inch here and turn `Grid.method` into `Registry.getComponentMethod`? Then `grid` becomes an optional component, right?

test/jasmine/tests/register_test.js
```diff
@@ -277,53 +151,119 @@ describe('the register function', function() {
...
+    it('returns an empty array if none present', function() {
+        expect(Registry.getTransformIndices({
+            transforms: [
+                {type: 'filter'},
+                {type: 'groupby'}
+            ]
+        }, 'degauss')).toEqual([]);
+    });
+
+    it('returns an empty array if none present', function() {
```
alexcjohnson: test case description copied from above

But also, if you leave these un-nested, the descriptions should reference the method being tested.

src/registry.js
```diff
@@ -209,8 +209,16 @@ exports.getComponentMethod = function(name, method) {
...
 /**
+ * Call registered api method.
  *
+ * @param {string} name : api method name
+ * @param {array} args : array of argument passed to api method
+ * @return {any} : returns api method output
  */
+exports.call = function(name, args) {
+    return exports.apiMethodRegistry[name].apply(null, args);
+};
```
alexcjohnson: Can we get rid of the array wrapping the args, like other `call` methods (`d3.select().call` and native `function.call`)? Borrowing from d3 and condensing, seems like it would just be:
```
return exports.apiMethodRegistry[name].apply(null, [].slice.call(arguments, 1));
```
or, perhaps marginally faster if we don't care about `this` being the method name:
```
return exports.apiMethodRegistry[name].apply([].slice.call(arguments));
```

tasks/test_syntax.js
```diff
@@ -260,12 +260,9 @@ function assertCircularDeps() {
...
         var logs = [];
 
-        // see https://github.com/plotly/plotly.js/milestone/9
-        var MAX_ALLOWED_CIRCULAR_DEPS = 16;
-
-        if(circularDeps.length > MAX_ALLOWED_CIRCULAR_DEPS) {
+        if(circularDeps.length) {
             console.log(circularDeps.join('\n'));
-            logs.push('some new circular dependencies were added to src/');
+            logs.push('some circular dependencies were found to src/');
```
alexcjohnson: found *in* src/

alexcjohnson:
@etpinard this is brilliant. Beautiful. Heroic even. Amazing that even with the extra machinery it's a net decrease in LOC.

Just a few comments, the only substantive one being the [`Registry.call` signature](https://github.com/plotly/plotly.js/pull/2429#discussion_r171728952).

Love it. End of an era.

src/registry.js
etpinard: That's what I was going to do until I saw this [block](https://github.com/plotly/plotly.js/blob/113b3618732f28d1a2158771544bf4e60e00c084/src/plots/command.js#L264-L281) (called from updatemenus and silders) where the argument list is built as an array.

I think I would prefer keeping this pattern. I could rename `Registry.call` -> `Registry.apply` to avoid confusion on the call signature if you prefer.

alexcjohnson: Ah right, that's the one place you use `Registry.call` without brackets, and it would be awkward to try and fit that into the `call` pattern. It's a shame that this one call adds cruft to the hundred other ones though. What about creating both `Registry.call` *and* `Registry.apply` methods? Or, since at least for now that's the only place we'd use `Registry.apply`, just inlining it there:
```
return Registry.apiMethodRegistry[method].apply(null, allArgs).catch(...
```

src/plots/plots.js
etpinard: done in -> https://github.com/plotly/plotly.js/pull/2429/commits/9246fdad6a85f812d23900e772ae210bac52d283

test/jasmine/tests/register_test.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2429/commits/e05f4f0db66c6797a88d62eaa67a07a7e1f5fe32

test/jasmine/tests/register_test.js
```diff
@@ -277,53 +151,119 @@ describe('the register function', function() {
...
+        name: 'newtype',
+        categories: ['red', 'green'],
+        calc: function() { return 42; },
+        plot: function() { return 1000000; },
+        basePlotModule: { name: 'newbaseplot' }
+    };
+
+    beforeAll(function() {
+        Registry.register(fakeModule);
+    });
```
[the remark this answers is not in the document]
etpinard: Good catch. Fixed in https://github.com/plotly/plotly.js/pull/2429/commits/e05f4f0db66c6797a88d62eaa67a07a7e1f5fe32

tasks/test_syntax.js
etpinard: Good :eye: done in https://github.com/plotly/plotly.js/pull/2429/commits/d27b3baa880b8b41fd61d9caf3050d26ce7a7f19

src/registry.js
etpinard: Done in https://github.com/plotly/plotly.js/pull/2429/commits/705aed622d42648d55b4497c93486ea96921e1a0 :rocket:

etpinard:
A couple more things I want to add:

- first, I'll reference https://github.com/plotly/plotly-webpack/issues/2#issuecomment-333256054 where the first observations on `browser-pack-flat` +/- circular dependencies were made,
- second, errorbars and grid components are now totally optional and could be taken out of the core bundle in v2.
- lastly, now that we're registering plot api methods, we could make some of them _optional_ in v2 (i.e. move them out of `src/core.js`). If we choose to do, we'll just have to make sure the components that call api methods register them beforehand (similar to how trace modules register subplot modules).

src/plot_api/to_image.js
```diff
@@ -196,7 +196,7 @@ function toImage(gd, opts) {
     }
 
     return new Promise(function(resolve, reject) {
-        Registry.call('plot', [clonedGd, data, layoutImage, configImage])
+        plotApi.plot(clonedGd, data, layoutImage, configImage)
```
alexcjohnson: just curious, why did you choose not to use `Registry` here?

alexcjohnson:
💃 💃 💃

src/plot_api/to_image.js
etpinard: `Plotly.toImage` shouldn't be in a circular dep with the other `Plotly` methods. `Plotly.toImage` should call `Plotly.plot` but never the other way around.

Ideally, I think the api method registry should only be used in `components/` and in the subplot modules (for interaction updates). `Plotly` methods need to call component methods (e.g `.draw`) and some components need to call `Plotly` methods (e.g. updatemenus / mode bar button click) - this is circular, hence the need for an api method registry.

Issue #2417, opened 2018-02-28
Look for ways to improve big-data axis-expand algo


etpinard:
... most notably for `scattergl`, but also perhaps for all traces types that support data arrays with length greater than 1e5 in the future.

From @alexcjohnson in https://github.com/plotly/plotly.js/pull/2404#discussion_r170957224 :

> How did you find sizeAvg to work in practice? A little hard to say I guess, since we don't have a lot of real data > 1e5 points to play with... You're right that we don't want to just use sizeMax, it's worth accepting a bit of clipping in order to generally have less wasted space, and much of the time the largest point won't be on any edge... just wondering how that balance plays out in practice, whether we would be better off with something like halfway between the average and max.

> Anyway perhaps we don't have a god way to answer that question right now. I'll just mention that in case we do want to try and do better later, we could find some heuristics that only add a little bit of computation, like binning data points into top/middle/bottom thirds, and only using the top third for the top padding... maybe even with a smooth weighting of the size based on how far it is from the edge. That would still be far faster than the full calculation but could do a better job reducing wasted space without too much clipping.

alexcjohnson:
From https://github.com/plotly/plotly.js/pull/2437#discussion_r172549249:

> In comparison, at `1e5 - 1` pts (i.e. below the big-data threshold), we get:
> 
> ![image](https://user-images.githubusercontent.com/6675409/37040298-7f4f737c-2127-11e8-8819-d1c2ba9a1214.png)
> 
> 
> @alexcjohnson I'll reference https://github.com/plotly/plotly.js/issues/2417 - but honestly feel free to close it if you think `Axes.expand` is satisfactory now. 

Any improvement we make at this point will come with both a performance cost and a loss of generality - so lets close for now, and reopen if we hit a specific use case where the current behavior isn't good enough.

PR #635, opened 2016-06-13
Refactor plot autosize

etpinard:
fixes https://github.com/plotly/plotly.js/issues/537

This PR sets up a base branch for the in-progress autosize refactor starting out by @n-riesco in #577 patched in #629 and then reverted in #633.

Important comments:

- https://github.com/plotly/plotly.js/pull/629#issuecomment-225591388
- https://github.com/plotly/plotly.js/pull/577#discussion_r66688752

PR #2437, opened 2018-03-05
Autosize / autorange fixes

### LINKED ISSUE #2385: Plotly.react should autosize
A difference in result between `Plotly.newPlot` and `Plotly.react` - the latter doesn't pick up changes in the DOM that affect autosizing the plot. It should, as these two methods should always yield the same result regardless of their histories.

Original report: https://github.com/plotly/react-plotly.js/issues/41

Probably just need to pull in the [autosize handling](https://github.com/plotly/plotly.js/blob/ce5bc971770daaf3ed60d374f23a8af1bb61fe0b/src/plot_api/plot_api.js#L2053-L2064) from `relayout` into `react` as well.

### LINKED ISSUE #2428: autorange and pushmargin interaction
Double-clicking on this plot shouldn't cause a jump: https://codepen.io/nicolaskruchten/pen/yvwOJK

Per @alexcjohnson:

> I’m looking at  https://github.com/plotly/plotly.js/blob/master/src/plots/plots.js#L1675
> it might be enough to rearrange that and say something like `if(marginsChangedStuff) { if(replotting) { autoRangeAllAxes } else { Plotly.plot } }`

alexcjohnson:
Fixes #2428 - despite its recent posting, I think this is one of our oldest bugs. You see it in the way a lot of autoranged plots will change just a little bit if you double-click them to cause the autorange to rerun... including a whole lot of our test images. They don't do that anymore. This happened when the margins get pushed automatically (mostly by the legend) and the axes are padded, because `Axes.expand` (called during `calc`) depended on `ax._length`, which could change *after* `calc` due to margin pushers. As a nice side benefit, now that `Axes.expand` is independent of axis length all the layout attributes that affect axis lengths can have their `editType` downgraded from `calc` to `plot`, speeding up all these edits.

Fixes #2385 as well - `Plotly.react` now handles autosize and size changes correctly. This was *mostly* just a matter of adding the special autosize handling that we already had in `relayout` into `react` as well, but then when I dropped axis size changes to `editType: 'plot'` it also required some more careful handling of `ax._min/_max`.

cc @etpinard


src/plot_api/plot_api.js
```diff
@@ -2313,6 +2322,17 @@ exports.react = function(gd, data, layout, config) {
...
         var relayoutFlags = diffLayout(gd, oldFullLayout, newFullLayout, immutable);
 
+        // TODO: how to translate this part of relayout to Plotly.react?
+        // // Setting width or height to null must reset the graph's width / height
+        // // back to its initial value as computed during the first pass in Plots.plotAutoSize.
+        // //
+        // // To do so, we must manually set them back here using the _initialAutoSize cache.
+        // if(['width', 'height'].indexOf(ai) !== -1 && vi === null) {
+        //     fullLayout[ai] = gd._initialAutoSize[ai];
+        // }
```
alexcjohnson: On closer inspection I *think* I understand what this is about, but I don't think it's actually the correct behavior, for example if `gd` was resized external to plotly.js it wouldn't make sense to revert to when the plot was first drawn. I'm not quite sure how to handle this though.

src/plots/cartesian/autorange.js
```diff
@@ -0,0 +1,324 @@
...
+var Lib = require('../../lib');
+var FP_SAFE = require('../../constants/numerical').FP_SAFE;
+
+module.exports = {
+    getAutoRange: getAutoRange,
+    makePadFn: makePadFn,
+    doAutoRange: doAutoRange,
+    needsAutorange: needsAutorange,
+    expand: expand
+};
```
alexcjohnson: I initially pulled this module out of `axes.js` to avoid circular references; then I ended up moving `clearCalc` into `setConvert` instead of here, which mooted the circular reference problem, but I still like having these separate from the behemoth `axes.js`.

src/plots/gl2d/scene2d.js
```diff
@@ -261,7 +260,7 @@ proto.updateSize = function(canvas) {
     }
 
     // make sure plots render right thing
-    if(this.redraw) this.redraw();
+    // if(this.redraw) this.redraw();
```
alexcjohnson: Oh I forgot to remove this line... but @dfcreative @etpinard I wanted to ask about it anyway: if I leave it in, then occasionally - most noticeably when I autorange an old gl2d plot (ie pointcloud) - I get some extra tick labels: the new tick values get placed on the old axis range (if they're visible) then the correct tick labels get drawn over them without removing the obsolete ones.

Anyway I couldn't see any ill effects from removing it. OK?

src/plots/gl2d/scene2d.js
```diff
@@ -432,32 +433,13 @@ proto.plot = function(fullData, calcData, fullLayout) {
...
-            bounds[k] = Math.min(bounds[k], traceObj.bounds[k]);
-            bounds[k + 2] = Math.max(bounds[k + 2], traceObj.bounds[k + 2]);
-        }
-    }
-
     for(i = 0; i < 2; ++i) {
-        if(bounds[i] > bounds[i + 2]) {
-            bounds[i] = -1;
-            bounds[i + 2] = 1;
-        }
```
alexcjohnson: Also seems like `this.bounds` has stopped being used, so I 🔪 it. Can't wait until we can 🔪 this whole subplot type...

src/plot_api/plot_api.js
etpinard: :grimacing: auto-sizing always has a way to make my brain hurt.

Referencing https://github.com/plotly/plotly.js/pull/635 which did a lot of damage the streambed :face_with_head_bandage:

src/plots/cartesian/autorange.js
```diff
@@ -0,0 +1,324 @@
...
+function expand(ax, data, options) {
+    if(!needsAutorange(ax) || !data) return;
+
+    if(!ax._min) ax._min = [];
+    if(!ax._max) ax._max = [];
+    if(!options) options = {};
+    if(!ax._m) ax.setScale();
+
+    var len = data.length;
+    var extrapad = options.padded || false;
```
etpinard: Correct me if I'm wrong, but this used to be the only place in `expand()` that used `ax._length` as:

https://github.com/plotly/plotly.js/blob/8731cb5d87f305fff8386b5d8eb2f0280752393f/src/plots/cartesian/axes.js#L455

alexcjohnson: > this used to be the only place in expand() that used `ax._length`

Exactly - taking that out, and instead holding on to `extrapad` inside the `ax._min/_max` entries, to be evaluated during `getAutorange`, is what fixed #2428 and enabled `calc`->`plot`

etpinard: Great! That `makePadFn` helper above is brilliant :sparkles: and `expand()` is much clearer and well commented :books: 

Now, would mind mentioning in the jsDoc above that `expand()` **should not** depend on pixel values (e.g ax._length), but only on `calcdata` things?

src/plots/cartesian/autorange.js
```diff
@@ -0,0 +1,324 @@
...
+    var tozero = options.tozero && (ax.type === 'linear' || ax.type === '-');
+
+    var i, j, k, v, di, dmin, dmax, ppadiplus, ppadiminus, includeThis, vmin, vmax;
+
+    function getPad(item) {
```
etpinard: We could maybe rename this function. `getPad` might be confused with the output of `makePadFn`

alexcjohnson: haha I noticed that too... I'll change it.

src/plots/cartesian/autorange.js
```diff
@@ -0,0 +1,324 @@
...
+        // doAutoRange will get called on fullLayout,
+        // but we want to report its results back to layout
+
+        var axIn = ax._input;
+        axIn.range = ax.range.slice();
+        axIn.autorange = ax.autorange;
+    }
+}
+
+function needsAutorange(ax) {
```
etpinard: It feels awkward to export this now that auto-range routines are out of `axes.js`.

Maybe we should move [`fastAxisExpand`](https://github.com/plotly/plotly.js/blob/8731cb5d87f305fff8386b5d8eb2f0280752393f/src/traces/scattergl/index.js#L144-L161) out of scattergl/index.js and into this file?

alexcjohnson: Or just bake the `fastAxisExpand` functionality into `expand` to be used automatically when the paddings are constant - then everyone wins! I think that should be possible... I'll try it.

src/plots/cartesian/layout_attributes.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
         values: ['normal', 'tozero', 'nonnegative'],
         dflt: 'normal',
         role: 'style',
-        editType: 'calc',
+        editType: 'plot',
```
etpinard: It would be nice to add a few test like this [one](https://github.com/plotly/plotly.js/blob/8731cb5d87f305fff8386b5d8eb2f0280752393f/src/traces/scattergl/index.js#L144-L161) that :lock: the new `editType: 'plot'`.

src/plots/cartesian/autorange.js
etpinard: > then everyone wins!

That's an idea! But perhaps the [`TOO_MANY_POINTS`](https://github.com/plotly/plotly.js/blob/8731cb5d87f305fff8386b5d8eb2f0280752393f/src/traces/scattergl/index.js#L42) constant should remain per trace type?

src/plots/cartesian/layout_attributes.js
etpinard: ... as removing `editType: 'calc'` is a pretty be deal :racehorse: we wouldn't want to regress from that in the future!

src/plots/gl2d/scene2d.js
etpinard: Yeah, looks like in `scene2d.plot` this

https://github.com/plotly/plotly.js/blob/8731cb5d87f305fff8386b5d8eb2f0280752393f/src/plots/gl2d/scene2d.js#L474

does the same thing. :hocho: away!

src/traces/scattergl/index.js
```diff
@@ -104,14 +103,10 @@ function calc(gd, trace) {
...
-    } else {
-        if(markerOptions) {
-            ppad = 2 * (markerOptions.sizeAvg || Math.max(markerOptions.size, 3));
-        }
-        fastAxisExpand(xa, x, ppad);
-        fastAxisExpand(ya, y, ppad);
+    } else if(markerOptions) {
+        ppad = 2 * (markerOptions.sizeAvg || Math.max(markerOptions.size, 3));
     }
+    calcAxisExpansion(gd, trace, xa, ya, x, y, ppad);
```
etpinard: Very clean. This is great! 

To note though, this nice combined pathway is little slower at 1e6 pts. [Previously](https://github.com/plotly/plotly.js/pull/2404#issuecomment-368595753), `fastAxisExpand` would clock at ~5ms for one  coordinate, now we're more in a 20-30ms range. Even though this is significant, I think it's worthwhile to make scattergl autorange even more _on-par_ with the rest of the library. :+1: 

The failing tests are in this new [suite](https://github.com/plotly/plotly.js/blob/3bff26f7ef21c8ca43122e86214d268f57b77418/test/jasmine/tests/gl2d_plot_interact_test.js#L880) that got put in during https://github.com/plotly/plotly.js/pull/2404. Should be an easy fix.

test/jasmine/tests/plot_api_test.js
```diff
@@ -535,35 +549,69 @@ describe('Test plot api', function() {
...
+                gd = mock({
+                    data: [{y: [1, 2]}, {y: [4, 3], xaxis: 'x2', yaxis: 'y2'}],
+                    layout: {
+                        xaxis2: {domain: [0.6, 0.9]},
+                        yaxis2: {domain: [0.6, 0.9]}
+                    }
+                });
+
+                Plotly.relayout(gd, attr, axLayoutEdits[attr]);
+                expectReplot(attr);
```
etpinard: :100: / :100: test!

src/traces/scattergl/index.js
alexcjohnson: Huh, let's unpack that performance issue. Previously the inner loop was:
```js
for(var i = 0; i < vals.length; i += 2) {	
  var v = vals[i];	
  if(v < b0) b0 = v;	
  if(v > b1) b1 = v;	
}
```

this afternoon it was:
```js
for(i = 0; i < data.length; i++) {
  v = data[i];
  if(Math.abs(v) < FP_SAFE) {
    // data is not linearized yet so we still have to filter out negative logs
    if(v < vmin && (!isLog || v > 0)) vmin = v;
    if(v > vmax) vmax = v;
  }
}
```
The old code was accidentally omitting half the points (`i+= 2`), so we should be comparing 10ms to 20-30 ms (unless the 20-30ms is for both coordinates?). Then I'm doing two extra things: 1) comparing to `FP_SAFE` so we don't get a broken graph from numbers very close to the limit of floating point numbers, and 2) ensuring positive numbers when on a log scale. The new code also has a bunch more logic before and after this loop, but nothing else scales with the data size, all that is presumably <1ms. In my own tests don't see the extra logic *in* the loop coming anywhere near doubling the time. What am I missing?

Anyway I noticed I could optimize it a bit by making two versions of the loop, one for log and one for other types. Seems worth a couple of extra lines since this gets so much traffic -> https://github.com/plotly/plotly.js/pull/2437/commits/247c626a61a1fe813577f1e6fdf5fee9824680e6

Updated the failing test in https://github.com/plotly/plotly.js/pull/2437/commits/cf4842609e240a708cac093e07470bbaae3ca1a9

src/plot_api/plot_api.js
alexcjohnson: Oy, I think we've all fought with that one at various points... If you don't mind, I'm going to leave this as it is now - with the TODO here to remind us about it if a related bug pops up, but without trying to implement anything around it. We have some great tests in [config_test](https://github.com/plotly/plotly.js/pull/2437/commits/9f774c0b650b8f913b00337afc1bdcf3735e7c04#diff-42c68411da55979155205ae2842e992fR87) that I extended to make sure that all of *those* test cases work the same in `relayout` and `react`, I'd like to just freeze it until we have more test cases to look at.

src/plots/gl2d/scene2d.js
alexcjohnson: :hocho -> https://github.com/plotly/plotly.js/pull/2437/commits/0ecaaaf8dd3fbb5d3aeb5e6aab4f4b0a0b86cfa6

src/plots/cartesian/autorange.js
alexcjohnson: fuller docs -> https://github.com/plotly/plotly.js/pull/2437/commits/9913afe4549580bb4b0271a5cf645633a8dfdab1

src/plots/cartesian/autorange.js
alexcjohnson: renamed -> https://github.com/plotly/plotly.js/pull/2437/commits/9913afe4549580bb4b0271a5cf645633a8dfdab1

src/traces/scattergl/index.js
etpinard: Great! You're latest patch in 247c626 got `Axes.expand` within 5ms of the `fastAxisExpand` with `i++` swapped in for `i += 2` at 1e6 pts. Considering it's more robust and works with traces other than `scattergl`, I'll call this a win. :1st_place_medal:

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -925,8 +925,8 @@ describe('Test scattergl autorange:', function() {
                 marker: {size: ms}
             }])
             .then(function() {
-                expect(gd._fullLayout.xaxis.range).toBeCloseToArray([-0.05, 1.05], 'x range');
-                expect(gd._fullLayout.yaxis.range).toBeCloseToArray([-0.11, 1.11], 'y range');
+                expect(gd._fullLayout.xaxis.range).toBeCloseToArray([-0.119, 1.119], 2, 'x range');
+                expect(gd._fullLayout.yaxis.range).toBeCloseToArray([-0.199, 1.199], 2, 'y range');
```
etpinard: In comparison, at `1e5 - 1` pts (i.e. below the big-data threshold), we get:

image


@alexcjohnson I'll reference https://github.com/plotly/plotly.js/issues/2417 - but honestly feel free to close it if you think `Axes.expand` is satisfactory now.

alexcjohnson: Right, so I guess that result is saying that in this case the [constant pixel padding we choose](https://github.com/plotly/plotly.js/pull/2437/commits/3875fcc1d767af30fb5e8e08a9548191d754c70e#diff-e03c5a2a46f9392088513c32645ea9d5R107) is a bit bigger than it would need to be... but this is a random case, a lot of real cases will have correlations, like all the biggest markers are near one edge, so I think this is a pretty good compromise.

etpinard: > so I think this is a pretty good compromise.

Yep. I think so too :ok_hand:

PR #2440, opened 2018-03-05
Fix layout of centered hover info #2154

### LINKED ISSUE #2154: Hover tooltip positioned wrong in stacked barchart when number of bar is 1

rmoestl:
This fixes #2154: when hover info (aka tooltip) is rendered in the center due to a lack of space left or right, the text overflowed the box and the secondary hove info (trace name) was hidden underneath the primary one.

Final solution:
2154_layout_centered-hover-info-final

Hint: part of the discussion took place in https://github.com/rmoestl/plotly.js/pull/1.

### DIFF
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -1179,24 +1179,20 @@ describe('hover info', function() {
             expect(elem1BB.top - elem2BB.top).toBeWithin(0, tolerance, msg);
         }
 
-        it('renders labels inside boxes', function(done) {
+        it('renders labels inside boxes', function() {
             _hover(gd, 300, 150);
 
             var nodes = ensureCentered(centeredHoverInfoNodes());
             assertElemInside(nodes.primaryText, nodes.primaryBox, 'Primary text inside box');
             assertElemInside(nodes.secondaryText, nodes.secondaryBox, 'Secondary text inside box');
-
-            done();
         });
 
-        it('renders secondary info box right to primary info box', function(done) {
+        it('renders secondary info box right to primary info box', function() {
             _hover(gd, 300, 150);
 
             var nodes = ensureCentered(centeredHoverInfoNodes());
             assertElemRightTo(nodes.secondaryBox, nodes.primaryBox, 'Secondary box right to primary box');
             assertTopsAligned(nodes.secondaryBox, nodes.primaryBox, 'Top edges of primary and secondary boxes aligned');
-
-            done();
         });
     });
 });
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/hover_label_test.js
```diff
@@ -1109,6 +1109,96 @@ describe('hover info', function() {
...
+
+        function assertTopsAligned(elem1, elem2, msg) {
+            var elem1BB = elem1.getBoundingClientRect();
+            var elem2BB = elem2.getBoundingClientRect();
+
+            // Hint: toBeWithin tolerance is exclusive, hence a
+            // diff of exactly 1 would fail the test
+            var tolerance = 1.1;
+            expect(elem1BB.top - elem2BB.top).toBeWithin(0, tolerance, msg);
+        }
```
alexcjohnson: Love these assert functions! They'd be good candidates for `assets/custom_assertions` but we can leave them here until someone else finds a use for them elsewhere.

test/jasmine/tests/hover_label_test.js
```diff
@@ -1109,6 +1109,96 @@ describe('hover info', function() {
...
+
+        it('renders labels inside boxes', function(done) {
+            _hover(gd, 300, 150);
+
+            var nodes = ensureCentered(centeredHoverInfoNodes());
+            assertElemInside(nodes.primaryText, nodes.primaryBox, 'Primary text inside box');
+            assertElemInside(nodes.secondaryText, nodes.secondaryBox, 'Secondary text inside box');
+
+            done();
```
alexcjohnson: do these test cases need to be async? Can't we just take out `done` from the function args and here?

alexcjohnson:
@rmoestl looks great, beautiful tests. Just the one quick question https://github.com/plotly/plotly.js/pull/2440#discussion_r172212582 and then I think we're ready to 💃

test/jasmine/tests/hover_label_test.js
rmoestl: Good point. They don't need to be async. I'll go and change that.

etpinard:
Great PR @rmoestl !

Thanks very much :tada:

alexcjohnson:
Can we close #865 now too?

alexcjohnson:
Interesting - looks like the text gets pushed up to avoid overlap with the orange bar, but the rect doesn't. @rmoestl since you've got your head wrapped around this part of the code, can you hunt this last bit down as well?

PR #2447, opened 2018-03-06
Bundle with browser-pack-flat plugin

etpinard:
A nice benefit from removing all circular deps in https://github.com/plotly/plotly.js/pull/2429, [`browser-pack-flat`](https://github.com/goto-bus-stop/browser-pack-flat) now allows us to trim some bundle bits (especially in our minified bundle). 

Here's what `browser-pack-flat` can trim:

image

that's about 10 gzip+minified `kB` per bundle. Not bad I think.

**That said**, using `browser-pack-flat` comes at a price :moneybag: , `npm run build` is not a bit slower and consumes a lot more memory. So I had to use [`run-series`](https://github.com/feross/run-series) to reduce memory consumption and make `npm run build` succeed consistently. Note also, I thinking of only using `browser-pack-flat` during `npm run build` to not slow `npm run watch` and `npm start`.

------------------

Referencing https://github.com/plotly/plotly-webpack/issues/2 that first discussed bundling with `browser-pack-flat`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1079,6 +1079,30 @@
         "umd": "3.0.1"
       }
     },
+    "browser-pack-flat": {
+      "version": "3.0.8",
+      "resolved": "https://registry.npmjs.org/browser-pack-flat/-/browser-pack-flat-3.0.8.tgz",
+      "integrity": "sha512-1pKozTPswxbugz552PJoSzgn3LB54maoyRz2aHVxHF/gtCJmhQbxE3SsGcfs+VMl1Aqst1kk8vJxvNa0pJ/2cg==",
+      "dev": true,
+      "requires": {
+        "JSONStream": "1.3.2",
+        "combine-source-map": "0.8.0",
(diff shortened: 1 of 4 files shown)


alexcjohnson:
Seems like a clear win, and all the choices you made seem right. One question though: why did the raw sizes increase?

etpinard:
> One question though: why did the raw sizes increase?

Good question. I think it has to do with the long variable names `browser-pack-flat` uses to denote exported fields in the global scope (example [here](https://github.com/goto-bus-stop/browser-pack-flat#what-exactly)) `brower-pack-flat` seems to be more liberal on `\n` too. Uglifyjs can easily :hocho: those things hence to smaller minified bundles.

alexcjohnson:
Also:

> browser-pack-flat does some more things like rewriting top-level variables in modules in case there is another variable with the same name in another module

Those names probably start getting pretty long in our bundle.

Anyway anyone who cares about size will use the minified versions, and this is a clear win. 💃

Issue #2421, opened 2018-02-28
Scatterpolar bug: not all theta labels showing


yanndebelgique:
https://codepen.io/yannbel/pen/rJoEKy
Current behavior:
Theta labels 'H','I','J' not showing. only labels 'A','B','C', 'D', 'E','F','G' showing
Expected behavior:
Example: All the theta labels should be visible on the graph i.e. ['A','B','C', 'D', 'E','F','G','H','I','J']

etpinard:
Thanks for the report.

yanndebelgique:
Thanks for the great work with Plotly.js! awesome library

Issue #2255, opened 2018-01-16
Polar 2.0 open items


etpinard:
Followup issue of https://github.com/plotly/plotly.js/pull/2200 and the on-par polar [milestone](https://github.com/plotly/plotly.js/milestone/2).

------

### Features we could easily add
- [ ] radial and angular axis `spikes`
- [ ] support for `radialaxis.side: 'center'` to display radial axis labels along invisible radial axis lines
- [x] radial axis _hole_ (similar to pie `hole`, mpl [example](https://github.com/matplotlib/matplotlib/issues/4217)) - done in https://github.com/plotly/plotly.js/pull/2977
- [x] support square grid lines for category angular axes [example](https://matplotlib.org/_images/radar_chart.png) - done in https://github.com/plotly/plotly.js/pull/2739
- [ ] some way to _snap_ to NSEW corners during angular drag
- [ ] way to activate radial (and possibly angular down the road) autorange on double click (i.e. implement `doubleClick: 'autosize'` config option
- [ ] radial (and angular?) axis zeroline
- [ ] polar (r,theta) data-referenced annotations
- [ ] trace-specified and radial-axis calendars
- [ ] make polar respect `gd._context.showAxisDragHandles`? 

### Potential features that require some thought

- [x] How to replace legacy polar's `area` trace type? I'd vote for replacing it with `polarbar` or `wedge`. Most of the `bar` logic could be reused here. **Moved to ->** https://github.com/plotly/plotly.js/issues/2810 **Done in ->** https://github.com/plotly/plotly.js/pull/2954
- How to replace legacy polar's (r,t) `bar` traces? I'd vote simply removing them, unless someone out there finds them useful.
- What to do with `r < 0` values? In #2200, they are simply dismissed, but perhaps we should add an option to plot them on the opposite side of the origin. See https://github.com/plotly/plotly.js/pull/2200#commitcomment-26416872
- Date angular axes. See examples in https://github.com/plotly/plotly.js/pull/2200#discussion_r156397924) and some considerations in the last part of https://github.com/plotly/plotly.js/commit/1e81039384ebf89de376c3ddd004d0f103278fa6#r26653662 **Moved to -->** https://github.com/plotly/plotly.js/issues/4379
- Should we allow `angularaxis.period` to be less then the number of categories on categorical angular axes? See https://github.com/plotly/plotly.js/pull/2200#discussion_r161143299
[...]

PR #2449, opened 2018-03-06
Fix category angular axis mocked range logic

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2421

- :hocho: useless axex.expand call in scatterpolar/calc, this is an artefact from an earlier angular dates attempt
- set mocked angular axis range using max of set 'period' and total categories length.
- more on period vs category limitations and possible improvements in #2255 

cc @alexcjohnson


alexcjohnson:
Nice job. 💃

PR #2451, opened 2018-03-07
Fix rangeslider titles with stacked y axes

### LINKED ISSUE #2443: rangeslider with multiple yaxis (after v1.32.0)
We have a plot with multiple y axes and a rangeslider. This produces something like:

image

This worked perfectly up until v1.31.1. The next v1.32.0 breaks it and renders rangeslider over one of the plots:

image

Reproducible example is here: https://codepen.io/anon/pen/qxzVpW

In the HTML view replace https://cdn.plot.ly/plotly-1.31.0.min.js script with the 1.32.0 or the latest from the CDN to reproduce the issue.

May I suggest that xaxis.rangeslider layout config option has something like "domain: [ number, number ]" rather than "thickness" option to specify where exactly I want to place the slider?

alexcjohnson:
Fixes #2443 - this was actually two issues in one:
- rangeslider positioning didn't take into account *all* linked y axes, just the main one.
- the default anchors for cartesian axes were chosen prior to sorting the axis lists - so if `y2` showed up in the data before `y`, then `y2` became the default.

I also noticed while working on this that titles with `xaxis.side = 'top'` was mangled. I simplified the whole business by having the rangeslider itself draw the (bottom) title and bailing out of title drawing in `Axes.doTicks`.

cc @etpinard


src/components/rangeslider/draw.js
```diff
@@ -151,18 +157,29 @@ module.exports = function(gd) {
...
+        Titles.draw(gd, axisOpts._id + 'title', {
+            propContainer: axisOpts,
+            propName: axisOpts._name + '.title',
+            placeholder: fullLayout._dfltTitle.x,
+            attributes: {
+                x: axisOpts._offset + axisOpts._length / 2,
+                y: y + opts._height + opts._offsetShift + 10 + 1.5 * axisOpts.titlefont.size,
+                'text-anchor': 'middle'
+            }
+        });
```
alexcjohnson: Where to put this `Titles.draw` call totally confused me. First I put it after `Plots.automargin` below... that failed, calling `Titles.draw` twice: first with the adjusted margins and THEN with the un-adjusted margins - the reason being that `Plots.autoMargin` invokes a whole new `Plotly.plot` call, and in that one we get the adjusted margin, but then *after* that we come back here and draw the title with the original margin. Instead, I had to put it here, so that the draw with the original margin happens before we adjust the margins, then the draw with the adjusted margin happens inside `Plots.autoMargin`. Or I suppose I could have added some MathJax so `Plots.autoMargin` would be async, that would have fixed it too 🙄 

@etpinard this is what got me thinking about altering our plotting pipeline with some intermediate step that's just concerned with arranging all the components but doesn't draw anything other than prerendering and stashing all the text we need sizes for.

Also, I started trying to pull out the `10 + 1.5 * axisOpts.titlefont.size` that's repeated from `Axes.doTicks` into some sort of `Titles.offset` function (that's why I started changing the structure in `titles/index`), but gave up for now as there's too much logic concerning side and other axis attributes in `doTicks`.

src/components/rangeslider/draw.js
```diff
@@ -151,18 +157,31 @@ module.exports = function(gd) {
         // update current range
         setPixelRange(rangeSlider, gd, axisOpts, opts);
 
-        // update margins
+        // title goes next to range slider instead of tick labels, so
+        // just take it over and draw it from here
+        if(axisOpts.side === 'bottom') {
+            Titles.draw(gd, axisOpts._id + 'title', {
```
etpinard: Nice solution!

src/components/rangeslider/draw.js
```diff
@@ -151,18 +157,31 @@ module.exports = function(gd) {
...
+                propContainer: axisOpts,
+                propName: axisOpts._name + '.title',
+                placeholder: fullLayout._dfltTitle.x,
+                attributes: {
+                    x: axisOpts._offset + axisOpts._length / 2,
+                    y: y + opts._height + opts._offsetShift + 10 + 1.5 * axisOpts.titlefont.size,
```
etpinard: Here, that `10` px offset comes from eyeballing it?

src/plots/plots.js
```diff
@@ -1390,6 +1385,11 @@ plots.supplyLayoutModuleDefaults = function(layoutIn, layoutOut, fullData, trans
         Cartesian.finalizeSubplots(layoutIn, layoutOut);
     }
 
+    // sort subplot lists
+    for(var subplotType in layoutOut._subplots) {
+        layoutOut._subplots[subplotType].sort(Lib.subplotSort);
```
etpinard: So, if I get this right, PR https://github.com/plotly/plotly.js/pull/2227 (which got merged in `1.32.0`) is the culprit for https://github.com/plotly/plotly.js/issues/2443 ?

src/components/rangeslider/draw.js
etpinard: ... or is there a similar offset for range-slider-less axes?

test/image/mocks/range_slider_multiple.json
```diff
@@ -20,14 +20,17 @@
     "xaxis": {
       "domain": [ 0, 0.45 ],
       "range": [ 1, 2 ],
-      "rangeslider": {}
+      "rangeslider": {},
+      "title": "X",
+      "side": "top"
```
etpinard: Nice touch :1st_place_medal:

test/image/mocks/range_slider_axes_stacked.json
```diff
@@ -0,0 +1,11 @@
+{
+  "data": [
+    {"y": [1, 2, 3], "yaxis": "y2"},
+    {"fill": "tozeroy", "y": [30, 20, 10]}
+  ],
+  "layout": {
+    "xaxis": {"rangeslider": {}, "title": "x"},
+    "yaxis": {"title": "y", "domain": [0, 0.25]},
+    "yaxis2": {"title": "y2", "domain": [ 0.3, 1]}
```
etpinard: Referencing https://github.com/plotly/plotly.js/issues/2172 and more specifically https://github.com/plotly/plotly.js/issues/2172#issuecomment-347279990 - so if I understand correctly, the baseline for this mock is _wrong_ as the desired behavior would be replicate the stacked y-axes in the range slider?

src/components/rangeslider/draw.js
```diff
@@ -96,11 +96,17 @@ module.exports = function(gd) {
...
-            domain = axisOpts.domain,
-            oppDomain = oppAxisOpts.domain,
-            tickHeight = (axisOpts._boundingBox || {}).height || 0;
+        var margin = fullLayout.margin;
+        var graphSize = fullLayout._size;
+        var domain = axisOpts.domain;
+        var tickHeight = (axisOpts._boundingBox || {}).height || 0;
+
+        var oppBottom = Infinity;
+        var subplotData = Axes.getSubplots(gd, axisOpts);
```
etpinard: ... I hope these patched lines won't conflict with https://github.com/plotly/plotly.js/pull/2364

src/components/rangeslider/draw.js
alexcjohnson: copied from `doTicks` https://github.com/plotly/plotly.js/blob/dfe13b5dcfb9f4e632c00fe8b5ad2959db60ed01/src/plots/cartesian/axes.js#L2031
That's what made me start to try and abstract this out, before deciding the logic was a little too convoluted to want to tackle it right now.

src/plots/plots.js
alexcjohnson: Yes, I think so, at least that was the main part of it.

src/components/rangeslider/draw.js
etpinard: Great. Thanks!

test/image/mocks/range_slider_axes_stacked.json
alexcjohnson: yes, most likely this mock will change with #2172. Though reconciling #2172 with #2364 may be a bit of a challenge...

etpinard:
Great. Looks like all my questions have been answered :dancer: (or maybe wait until https://github.com/plotly/plotly.js/pull/2364 is merged - your call)

PR #2454, opened 2018-03-07
fix #1645 - ensure we don't draw ticks if there are none to draw

### LINKED ISSUE #1645: Axis: Dragging categories out of chart area
When a chart has categories in the x-axis, it is possible to drag the last entry out of the chart area.

I've made a screencast displaying this behavior:

axis slider

The chart used is found on the bar chart example site https://plot.ly/javascript/bar-charts/

Steps for reproduction:
* Open the referenced example page
* Scroll down to any chart with a categories x-axis
* Start dragging from the middle of the axis to the left

alexcjohnson:
@etpinard a quick one I noticed while perusing old issues.

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -522,22 +522,27 @@ axes.calcTicks = function calcTicks(ax) {
     // find the first tick
     ax._tmin = axes.tickFirst(ax);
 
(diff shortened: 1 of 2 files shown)


etpinard:
Ah nice :dancer: 

I actually tried to fix that bug once (a late Friday pm effort). Glad it's finally fixed!

PR #2458, opened 2018-03-08
Polar + flat-pack

### LINKED ISSUE #2456: Polar plots broken in v1.35.0 (visible on docs site)
If you go to the [Polar Charts in plotly.js](https://plot.ly/javascript/polar-chart/) section of the documentation (tested in Chrome 64) you will notice that the grid lines are plotted and the radial axis ticks & labels are present, but the angular axis labels and the traces are not. For example, [this first codepen](https://codepen.io/plotly/pen/LQepNO) does not work.

Instead, an error message like the one below appears in the console. This is either a problem with `plotly.js` or a problem with the docs, but I can't imagine this not being a problem.

```
Uncaught TypeError: node.getAttribute is not a function
    at Array.d3_selectionPrototype.classed (VM456 plotly-latest.js:754)
    at ax._transfn (VM456 plotly-latest.js:173293)
    at SVGTextElement.<anonymous> (VM456 plotly-latest.js:27849)
    at VM456 plotly-latest.js:1028
    at d3_selection_each (VM456 plotly-latest.js:1034)
    at Array.d3_selectionPrototype.each (VM456 plotly-latest.js:1027)
    at positionLabels (VM456 plotly-latest.js:27845)
    at SVGTextElement.<anonymous> (VM456 plotly-latest.js:27808)
    at VM456 plotly-latest.js:1028
    at d3_selection_each (VM456 plotly-latest.js:1034)
```

Changing the script source to [v1.34.0](https://cdnjs.cloudflare.com/ajax/libs/plotly.js/1.34.0/plotly.js) seems to fix the problem, so this could be a regression introduced in v1.35.0.

alexcjohnson:
@etpinard I *think* this will fix #2456 but not quite sure how to test it since everything looks fine in the dashboard... I guess using flat-pack in our tests - jasmine? image?

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -1787,7 +1787,7 @@ axes.doTicks = function(gd, axid, skipTitle) {
                 var anchor = labelanchor(angle, d);
                 var thisLabel = d3.select(this),
                     mathjaxGroup = thisLabel.select('.text-math-group'),
-                    transform = transfn(d) +
+                    transform = transfn.call(thisLabel.node(), d) +
                         ((isNumeric(angle) && +angle !== 0) ?
                         (' rotate(' + angle + ',' + labelx(d) + ',' +
                             (labely(d) - d.fontSize / 2) + ')') :


etpinard:
:dancer: thanks!

I'll see if I can find a decent way to test browser-pack-flat bundle without blowing it up our test times.

PR #2461, opened 2018-03-08
throw an error if the test window is hidden

alexcjohnson:
...when we try to emit a mouse event, which the browser may ignore if it's hidden

@etpinard this would have alerted me a bit quicker yesterday why my tests were failing sporadically 😅 

```
Error: document.visibilityState = "hidden" - Please make the window visible.
```


etpinard:
Smart. :dancer:

Issue #749, opened 2016-07-18
Improve 'relink private keys' steps


etpinard:
### Synopsis

At the moment, during `Plots.supplyDefaults` we loop over all keys found in the _old_ trace and layout containers in order to _relink_ all _private keys_ **not set** during the defaults step in the new trace and layout containers (remember, `fullData` and `fullLayout` start as empty `[]` and `{}` respectively in every `Plots.supplyDefaults` call).

These _private keys_ are linked to subplot instances (e.g `fullLayout.scene._scene`), computed values (e.g. `fullLayout.xaxis._m`), graph svg selections (e.g. `fullLayout._infolayer`) etc ...
### In brief

Looping over all keys is dumb, slow and hard to maintain ; we can do better!
### Observations

Not relinking private keys in trace objects results in:
- geo trace visibility toggle not functioning (`_module` isn't defined anymore for `visible: false` traces)
- contour restyle not working as `trace._emptypoints` isn't relinked

Not relinking array container results in:
- annotations not behaving correctly on zoom -> double-click interactions
### Possible solution

Make the trace and base plot module (and possibly the component module) `clean` method relink the _sticky_ keys that need to be relinked during `Plots.supplyDefaults`.  This would make the `clean` methods handle all old container --> new container logic in one go.

For example, most base plot modules have a  `clean` method that looks like:

``` js
oldSubplotIds.forEach((id) => {
  if(!newFullLayout[id]._subplot && !!oldFullLayout[id]._subplot) {
     oldFullLayout[id]._subplot.destroy();
     return;
  }

  // it would be then be easy to relink the active subplot objects here
  newFullLayout[id]._subplot = oldFullLayout[id]._subplot;
});
```

It might be convenient to put private keys generated in [`makePlotFramework`](https://github.com/plotly/plotly.js/blob/7ae437bbbb81ec1f49fee7145fc796d64e688bc6/src/plot_api/plot_api.js#L2592) under one object in fullLayout (e.g  `fullLayout_base._toppaper`) to keep track of private keys more easily.

etpinard:
Closing this thing for the time being. `relinkPrivateKeys` is getting better and more thoroughly tested, thanks to notably https://github.com/plotly/plotly.js/pull/2375.

alexcjohnson:
[...]

PR #2465, opened 2018-03-09
Clean up constraint groups

### LINKED ISSUE #2462: For swapping between scatter and scatter3d, scaleanchor stays in layout and is no longer ignored by scatter3d plot, causing error
I have a use case where users need to look at a plot in 2D sometimes, but change to a 3D view at others.  This is done via a simple toggle.  Under version 1.31.2, this worked fine, any parameters the chart did not care about were ignored, but we also only conditionally added the `xaxis: { scaleanchor: 'y' }` when the user chose to look at a 2D plot of the data.  Now, trying in both 1.34.0 and 1.35.0, this causes the 3D plot to error and not render.  

We are using react-plotlyjs to construct our plots, but that is at version 1.1.1, which still creates a new chart everytime the inputs change, so it doesn't seem to be a matter of that hanging onto something, but if it is, I'll move this issue over there instead.  Just updating plotly.js to the latest versions seems to make this behavior show up.  

This is the error that I get when I try to swap:
image

To show the behavior I was seeing before, and expected I've put together [this codepen](https://codepen.io/enddev/pen/yKLgbg).

Now, by only changing the version of plotly.js, [this codepen](https://codepen.io/enddev/pen/wmvJbm) shows the error behavior.  Additionally, by changing the starting value of dim3d on line 16 to true, the 3D plot renders fine the first time, and switching to 2D gives a different error (see below), but switching back, the 3D plot gives the same error as above again, and the toggle gets off somehow, rendering the 2D when it's supposed to render 3D.
image

Just in case it matters, and to provide more information, our project is built with webpack 3.10.0, with typescript 2.0.3, but since the codepen has the same behavior, it probably isn't related to current environment.  I've also attempted to update to the latest react-plotly (1.7.0) hoping it might change something, but I can't see any difference in behavior.

alexcjohnson:
Fixes #2462

The problem was that `_axisConstraintGroups` gets reset in cartesian-only code, so with no cartesian module, the old one just gets added back in by `relinkPrivateKeys`. So I cleaned that up, but also realized that without cartesian we don't even need to do some of the steps in `Plotly.plot` (including the one that errored) 🐎 In fact either one of those would have been enough to fix this bug, but I figured both can't hurt.

@etpinard we should really rethink `relinkPrivateKeys` - can we perhaps figure out which keys we really *need* to relink, and give them some stronger marker than just the generic underscore, which also means private, not an attribute? There are probably other keys that can cause similar problems...

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -153,6 +153,8 @@ exports.plot = function(gd, data, layout, config) {
 
     var fullLayout = gd._fullLayout;
 
+    var hasCartesian = fullLayout._has('cartesian');
+
     // Legacy polar plots
     if(!fullLayout._has('polar') && data && data[0] && data[0].r) {
         Lib.log('Legacy polar charts are deprecated!');
@@ -426,16 +428,18 @@ exports.plot = function(gd, data, layout, config) {
         addFrames,
         drawFramework,
         marginPushers,
-        marginPushersAgain,
-        positionAndAutorange,
-        subroutines.layoutStyles,
-        drawAxes,
+        marginPushersAgain
+    ];
+    if(hasCartesian) seq.push(positionAndAutorange);
+    seq.push(subroutines.layoutStyles);
+    if(hasCartesian) seq.push(drawAxes);
+    seq.push(
         drawData,
         finalDraw,
         initInteractions,
         Plots.rehover,
         Plots.previousPromises
-    ];
+    );
 
     // even if everything we did was synchronous, return a promise
     // so that the caller doesn't care which route we took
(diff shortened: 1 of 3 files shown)


etpinard:
Looks good! :dancer: 



> we should really rethink relinkPrivateKeys - can we perhaps figure out which keys we really need to relink,

In https://github.com/plotly/plotly.js/issues/749, I thought about making each module (trace, subplot and component) list which keys need to be relink on redraws. At that time, it sounds like a bigger hassle that it was worth. But now, with almost 2-years worth of additional tests, we could maybe try to swap the generic _loop-over-all-keys_ relink step with something more fine-grained?

PR #2474, opened 2018-03-13
Perf drawFramwork

etpinard:
Here are two perf commits in `Cartesian.drawFramework` that will help splom subplot generation. :racehorse: 

At 400 subplots (or 20 splom dimensions :smirk: ), these commits are able to speed up `drawFramework` from ~1000ms down under 200ms.

Note that it might be nice to make the new `joinLayer` from 7d881f6 a standard `Lib` function and replace all our `selectAll().data([0])` with it. Waiting on @alexcjohnson before going through with it.

### DIFF
--- a/src/components/colorbar/draw.js
+++ b/src/components/colorbar/draw.js
@@ -246,18 +246,18 @@ module.exports = function draw(gd, id) {
 
         // now draw the elements
         var container = Lib.ensureSingle(fullLayout._infolayer, 'g', id, function(s) {
-            s.classed(cn.colorbar, true);
-            s.each(function() {
-                var s = d3.select(this);
-                s.append('rect').classed(cn.cbbg, true);
-                s.append('g').classed(cn.cbfills, true);
-                s.append('g').classed(cn.cblines, true);
-                s.append('g').classed(cn.cbaxis, true).classed(cn.crisp, true);
-                s.append('g').classed(cn.cbtitleunshift, true)
-                    .append('g').classed(cn.cbtitle, true);
-                s.append('rect').classed(cn.cboutline, true);
-                s.select('.cbtitle').datum(0);
-            });
+            s.classed(cn.colorbar, true)
+                .each(function() {
+                    var s = d3.select(this);
+                    s.append('rect').classed(cn.cbbg, true);
+                    s.append('g').classed(cn.cbfills, true);
+                    s.append('g').classed(cn.cblines, true);
+                    s.append('g').classed(cn.cbaxis, true).classed(cn.crisp, true);
+                    s.append('g').classed(cn.cbtitleunshift, true)
+                        .append('g').classed(cn.cbtitle, true);
+                    s.append('rect').classed(cn.cboutline, true);
+                    s.select('.cbtitle').datum(0);
+                });
         });
 
         container.attr('transform', 'translate(' + Math.round(gs.l) +
--- a/src/components/colorbar/draw.js
+++ b/src/components/colorbar/draw.js
@@ -245,10 +245,9 @@ module.exports = function draw(gd, id) {
         cbAxisOut.setScale();
 
         // now draw the elements
-        var container = fullLayout._infolayer.selectAll('g.' + id).data([0]);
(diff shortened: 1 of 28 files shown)


alexcjohnson:
Wow, I would not have guessed that those two were such a big drain, impressive find! Yes please, 🔪 all the `selectAll(...).data([0])`! The name `joinLayer` has never meant much to me though... I suppose I'd get used to it, but what about something like `Lib.ensureSingle`?

etpinard:
> I suppose I'd get used to it, but what about something like Lib.ensureSingle

I like the sound of `ensureSingle`. Thanks!

src/components/colorbar/draw.js
```diff
@@ -245,10 +245,9 @@ module.exports = function draw(gd, id) {
         cbAxisOut.setScale();
 
         // now draw the elements
-        var container = fullLayout._infolayer.selectAll('g.' + id).data([0]);
-        container.enter().append('g').classed(id, true)
-            .classed(cn.colorbar, true)
-            .each(function() {
+        var container = Lib.ensureSingle(fullLayout._infolayer, 'g', id, function(s) {
+            s.classed(cn.colorbar, true);
+            s.each(function() {
```
alexcjohnson: you don't like chaining here? ie
```js
s.classed(...)
    .each(...)
```

etpinard: Good catch. Thanks!

test/jasmine/tests/click_test.js
```diff
@@ -418,8 +418,8 @@ describe('Test click interactions:', function() {
             var node = document.querySelector('rect.nwdrag');
             var pos = getRectCenter(node);
 
-            expect(node.classList[0]).toBe('drag');
-            expect(node.classList[1]).toBe('nwdrag');
+            expect(node.classList[1]).toBe('drag');
+            expect(node.classList[0]).toBe('nwdrag');
```
alexcjohnson: OK as is, but this seems ripe for a custom matcher like:
```
expect(node).toBeClassed(['drag', 'nwdrag', 'cursor-nw-resize']);
```
That would be order-independent but also verify these are the *only* classes applied.

src/plot_api/plot_api.js
```diff
@@ -245,9 +245,7 @@ exports.plot = function(gd, data, layout, config) {
                 .attr('height', fullLayout.height);
         }
 
-        return Lib.syncOrAsync([
-            subroutines.layoutStyles
-        ], gd);
```
alexcjohnson: Too funny, so it just works! We may never know what obviated this...

src/plot_api/plot_api.js
```diff
@@ -245,9 +245,7 @@ exports.plot = function(gd, data, layout, config) {
...
+        return Plots.previousPromises(gd);
```
etpinard: @alexcjohnson tests are :white_check_mark: after removing the extra `layoutStyles` call. Good riddance :boat:

src/plot_api/plot_api.js
etpinard: _just works_<sup>TM</sup>

test/jasmine/tests/click_test.js
etpinard: Good call.

`cartesian_test.js` uses assertions like:

```js
expect(fills[0].classList.contains('js-tozero')).toBe(true)
```

which could work here too, but yeah checking adding a custom assertion that test multiple class names at once should be worthwhile.

src/components/colorbar/draw.js
etpinard: done in -> 8e6032d

test/jasmine/tests/click_test.js
etpinard: done in b974eab

alexcjohnson:
Beautiful! 💃 💃 💃

etpinard:
Ok. This probably deserves to be in the next minor release `1.36.0`, so I'll wait a few more days in case we need to make another patch release in `1.35.x` series.

PR #2483, opened 2018-03-20
Czech localization

fhurta:
Contains Czech localization.
(Unfotunatelly, I don't know other czech users of Plotly.js, so hopefully some find this and approve/make corrections.)

### DIFF
--- a/lib/locales/cs.js
+++ b/lib/locales/cs.js
@@ -11,7 +11,64 @@
 module.exports = {
     moduleType: 'locale',
     name: 'cs',
-    dictionary: {},
+    dictionary: {
+        'Autoscale': 'Auto rozsah',                                                             // components/modebar/buttons.js:139
+        'Box Select': 'Obdélníkový výběr',                                                                    // components/modebar/buttons.js:103
+        'Click to enter Colorscale title': 'Klikněte pro zadání názvu barevné škály',                     // plots/plots.js:437
+        'Click to enter Component A title': 'Klikněte pro zadání názvu komponenty A',            // plots/ternary/ternary.js:386
+        'Click to enter Component B title': 'Klikněte pro zadání názvu komponenty B',            // plots/ternary/ternary.js:400
+        'Click to enter Component C title': 'Klikněte pro zadání názvu komponenty C',            // plots/ternary/ternary.js:411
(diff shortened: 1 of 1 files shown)


etpinard:
Thanks very much!

alexcjohnson:
Thanks @fhurta ! I hope you don't mind that I asked a Slovak friend to review 🇨🇿 🇸🇰 - that's as close as I could get but he assures me he is qualified 😅 His comments:
> I think it's good. I would only change "střední" to "průměr" (literally meaning Average), because "průměr" is a noun while "střední" is an adjective and if you needed a noun you'd need to say "střední hodnota" (=mean value).

What do you think, is that reasonable?

Also a couple of things I noticed:
- There are two new strings that haven't made it into whichever file you based this off: ["Click to enter radial axis title"](https://github.com/plotly/plotly.js/blob/master/dist/translation-keys.txt#L10) and ["new text"](https://github.com/plotly/plotly.js/blob/master/dist/translation-keys.txt#L50) - can you add these?
- "kde:" (kernel density estimate) - is that understandable as is, to Czech speakers who know what a violin plot is, or is there a reasonable (and compact) translation?

fhurta:
The comments look reasonable (as I only use scatter graph, I don't know the context how the titles are used exactly). I added the two missing items. Regarding the changes:

 - `mean` - is the context here average value or mean value? if noun is expected here then I would translate it to either `průměr` or `střed` or `střední hodnota`

- `kde` - I googled it and haven't found any czech equivalent. I found some diploma thesis using this term, so I expect it is known term to violin plots users (or someone more familiar with it may suggest something).

alexcjohnson:
OK great - thanks for adding the new ones. Lets leave `kde` as is. `mean` and `mean ± σ` are used in hover labels for box plots, depending on the information shown:
screen shot 2018-03-27 at 8 02 40 am
screen shot 2018-03-27 at 8 02 18 am
Your call whether an adjective or a noun makes most sense here, but since it's in a hover label it's preferable to keep it short.

fhurta:
OK then. Seems 'průměr' fits the best here. Updated.

alexcjohnson:
Thanks for the quick updates, I think it's ready now!

Issue #1988, opened 2017-09-05
Replace getBoundingClientRect calls in axes.js


etpinard:
From a private convo with @alexcjohnson about https://github.com/plotly/plotly.js/pull/1980 (possibly a duplicate of https://github.com/plotly/plotly.js/issues/298):

> alexj 
also not for a patch release, but do you have any sense after doing this fix what it would take to get rid of those `getBoundingClientRect` calls? They’re some of the worst ones we have left, and if we could figure out how to turn them into `Drawing.bBox` calls plus some calculations it would help a lot with #1902, #888, etc…

> etienne 
yeah for sure, `Drawing.bbox` + `ax._offset` and `ax._length` should be enough to come up with a formula for `ax._boundingBox`

> alexj
doesn’t it include tick labels? 
oh I see tht’s the `Drawing.bbox`

> etienne
exactly
e.g. `yaxis._boundingBox.left = ax._offset - /* widest bbox of all y tick labels */` (edited)
`- ax.tickwidth`  ..

> alexj
right - probably with some tweaks depending on `showline`, `ticks`, etc… tons of edge cases there but it can certainly be done.
can probably be simplified by caching some of those results when the pieces are created.

> etienne
right

> alexj 
anyway would be awesome if you can :hocho: those while this stuff is fresh in your mind.

PR #2487, opened 2018-03-20
More optimization for cartesian subplots

etpinard:
to be more merged in https://github.com/plotly/plotly.js/pull/2474 

In progress for now, as I'll try to speed up a few more things in the coming days.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -347,7 +347,7 @@ exports.plot = function(gd, data, layout, config) {
 
     // draw ticks, titles, and calculate axis scaling (._b, ._m)
     function drawAxes() {
-        return Axes.doTicks(gd, 'redraw');
+        return Axes.doTicks(gd, graphWasEmpty ? '' : 'redraw');
     }
 
     // Now plot the data
--- a/src/plot_api/subroutines.js
+++ b/src/plot_api/subroutines.js
@@ -124,31 +124,20 @@ exports.lsInner = function(gd) {
 
         var xDomain = plotinfo.xaxis.domain;
         var yDomain = plotinfo.yaxis.domain;
-        var plotgroupBgData = [];
+        var plotgroup = plotinfo.plotgroup;
+        var plotgroupBg;
 
         if(overlappingDomain(xDomain, yDomain, lowerDomains)) {
-            plotgroupBgData = [0];
-        }
-        else {
+            var pgNode = plotgroup.node();
+            plotgroupBg = plotinfo.bg = Lib.ensureSingle(plotgroup, 'rect', 'bg');
+            pgNode.insertBefore(plotgroupBg.node(), pgNode.childNodes[0]);
+        } else {
+            plotgroupBg = plotgroup.select('rect,bg');
+            if(plotgroupBg.size()) plotgroupBg.remove();
+
             lowerBackgroundIDs.push(subplot);
             lowerDomains.push([xDomain, yDomain]);
         }
-
-        // create the plot group backgrounds now, since
-        // they're all independent selections
-        var plotgroupBg = plotinfo.plotgroup.selectAll('.bg')
-            .data(plotgroupBgData);
-
-        plotgroupBg.enter().append('rect')
-            .classed('bg', true);
-
-        plotgroupBg.exit().remove();
-
-        plotgroupBg.each(function() {
-            plotinfo.bg = plotgroupBg;
-            var pgNode = plotinfo.plotgroup.node();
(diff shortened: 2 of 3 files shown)


etpinard:
I was hoping to improve a few more things in this PR, namely:

- optimize the number of [`positionLabels`](https://github.com/plotly/plotly.js/blob/f9a15be72e54562eb48a805e886f50947a55402a/src/plots/cartesian/axes.js#L1785-L1816) calls in `Axes.doTicks` which is called multiples time per axis
- speed up [`fixLabelOverlaps`](https://github.com/plotly/plotly.js/blob/f9a15be72e54562eb48a805e886f50947a55402a/src/plots/cartesian/axes.js#L1829-L1882)
- and maybe take a stab at https://github.com/plotly/plotly.js/issues/1988

but I'll have to leave that for later.

----

@alexcjohnson would mind taking a :eyes: at this?

src/plot_api/plot_api.js
```diff
@@ -347,7 +347,7 @@ exports.plot = function(gd, data, layout, config) {
 
     // draw ticks, titles, and calculate axis scaling (._b, ._m)
     function drawAxes() {
-        return Axes.doTicks(gd, 'redraw');
+        return Axes.doTicks(gd, graphWasEmpty ? '' : 'redraw');
```
alexcjohnson: OK... this is the only caller using `doTicks(gd, '')` right? From the logic below it looks like there may have been one previously (or someone just calling `doTicks(gd)`) but I don't see it now. I'm a little reticent to add new branches inside `doTicks` when what we really need to do is unpack it into more digestible pieces - but that's probably a project for another time refactoring all of `axes.js`.

src/plot_api/subroutines.js
```diff
@@ -124,31 +124,20 @@ exports.lsInner = function(gd) {
...
 
         if(overlappingDomain(xDomain, yDomain, lowerDomains)) {
-            plotgroupBgData = [0];
-        }
-        else {
+            var pgNode = plotgroup.node();
+            plotgroupBg = plotinfo.bg = Lib.ensureSingle(plotgroup, 'rect', 'bg');
+            pgNode.insertBefore(plotgroupBg.node(), pgNode.childNodes[0]);
+        } else {
+            plotgroupBg = plotgroup.select('rect,bg');
```
alexcjohnson: `rect.bg` perhaps?

src/plot_api/subroutines.js
```diff
@@ -124,31 +124,20 @@ exports.lsInner = function(gd) {
...
+            if(plotgroupBg.size()) plotgroupBg.remove();
```
alexcjohnson: does this `if` make a difference? Does it take a non-negligible time to `remove` an empty selection?

src/plots/cartesian/axes.js
```diff
@@ -2012,14 +2014,12 @@ axes.doTicks = function(gd, axid, skipTitle) {
...
         // configure their own titles.
-        var ax = axisIds.getFromId(gd, axid);
 
         // rangeslider takes over a bottom title so drop it here
         if(ax.rangeslider && ax.rangeslider.visible && ax._boundingBox && ax.side === 'bottom') return;
 
-        var avoidSelection = d3.select(gd).selectAll('g.' + axid + 'tick');
         var avoid = {
-            selection: avoidSelection,
+            selection: tickLabels,
```
alexcjohnson: nice find!

src/plot_api/plot_api.js
etpinard: > OK... this is the only caller using doTicks(gd, '') right? From the logic below it looks like there may have been one previously (or someone just calling doTicks(gd)) but I don't see it now.

That's correct.

src/plot_api/subroutines.js
etpinard: Nice catch. I guess that means we have some untested code branches in there :rage:

alexcjohnson: > we have some untested code branches in there

maybe, maybe not. `rect,bg` would delete all `rect` elements (are there any others? if not, then what you have there will work, it just won't be as efficient) and all `bg` elements (which don't exist)

etpinard: fixed in 5307125ef9655bec0c2e24ba444b3cfc4c1c4f48

src/plot_api/subroutines.js
etpinard: that `if` did not improve things -> :hocho: in 5307125ef9655bec0c2e24ba444b3cfc4c1c4f48

alexcjohnson:
Fantastic- really whittling down the overhead! :dancer: :racehorse:

PR #2490, opened 2018-03-22
Added missing 'Saturday' string to locale-en

vdh:
For some reason `days` is missing everyone's favourite one 😉

### DIFF
--- a/src/locale-en.js
+++ b/src/locale-en.js
@@ -15,7 +15,7 @@ module.exports = {
         'Click to enter Colorscale title': 'Click to enter Colourscale title'
     },
     format: {
-        days: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday'],
+        days: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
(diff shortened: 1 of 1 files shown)


etpinard:
Thanks! Did that lead to any noticeable bugs?

vdh:
@etpinard Well from what I've seen trying to use i18n, it doesn't look like the locale settings have been used anywhere yet?

alexcjohnson:
Ha, good catch, wonder how we missed that? This will only show up with `%A` in a custom format, and we don't test every possible day and month in each format. We have some related tests in [lib_date_test](https://github.com/plotly/plotly.js/blob/9a79056bb223e13995be8debd22f60c9d56cbef0/test/jasmine/tests/lib_date_test.js#L523) - would you mind adding one that covers all the weekdays?

vdh:
@alexcjohnson I tried writing a test, but I wasn't able to figure out how to connect the `locale-en` file to the test.

alexcjohnson:
Ah you're right, sorry about that - we've just used `d3.time.format.utc` in `lib_date_test`, you don't get locales until you've done a `supplyDefaults` step. We've got some related tests in `localize_test`, we can probably model it off [this one](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/localize_test.js#L52) - but it's a little more involved than I had realized, I can take it if you want.

etpinard:
Great :dancer:

PR #2492, opened 2018-03-22
Fix scattergl selectedpoints clearance under select/lasso drag modes

etpinard:
fixes a bug brought up by @cpsievert in https://github.com/plotly/plotly.js/issues/2298#issuecomment-375111803 - though this PR does not fix the entire issue.

I decided also to push two DRY-ing commits :palm_tree: , one of which we'll be used during the splom (see https://github.com/plotly/plotly.js/issues/2372) push.

cc @dfcreative


etpinard:
@dfcreative can you review this PR please?

etpinard:
All right - merging this.

src/traces/scattergl/convert.js
```diff
@@ -0,0 +1,397 @@
...
+function getSymbolSdf(symbol) {
+    if(symbol === 'circle') return null;
+
+    var symbolPath, symbolSdf;
+    var symbolNumber = Drawing.symbolNumber(symbol);
+    var symbolFunc = Drawing.symbolFuncs[symbolNumber % 100];
+    var symbolNoDot = !!Drawing.symbolNoDot[symbolNumber % 100];
+    var symbolNoFill = !!Drawing.symbolNoFill[symbolNumber % 100];
+
+    var isDot = constants.DOT_RE.test(symbol);
```
alexcjohnson: I know this isn't new here, just noticing it: this won't work if `symbol` is provided by number in the first place. Better to just test the number once we have it - ie `isDot = symbolNumber >= 200` (and for [isOpen](https://github.com/plotly/plotly.js/pull/2492/commits/1c6144e291d3d1e191211f4f208672c224bfce4a#diff-202416f5e6d665625197b3b4abf2e142R94) `isOpen = (symbolNumber % 200) >= 100` - perhaps we should make these into helper functions in the `Drawing` module?). Probably not a very well-known feature, but it *is* supported by SVG, and could be useful particularly for symbol arrays, as you could use a typed array.

Dunno if anyone would use both names and numbers in the same plot, but if they did, it would also help to key the `SYMBOL_SDF` cache off the number.

etpinard: Good eye. This probably deserve an issue of its own.

src/traces/scattergl/convert.js
```diff
@@ -0,0 +1,397 @@
...
+        errorY: null,
+        selected: null,
+        unselected: null
+    };
+
+    if(trace.visible !== true) return opts;
+
+    if(subTypes.hasMarkers(trace)) {
+        opts.marker = convertMarkerStyle(trace);
+        opts.selected = convertMarkerSelection(trace, trace.selected);
```
dfcreative: In some reason `trace.selected` is empty here for multiple colors/multiple opacities #2500

etpinard: Ok - I'll have a look at this tomorrow

Issue #2495, opened 2018-03-23
Feature request: allow limiting dragmode 'select' to horizontal or vertical only


RedShift1:
As a follow-up from [this forum thread](https://community.plot.ly/t/only-allow-vertical-selection-with-dragmode-select/9157), in some datasets it doesn't make sense to be able to select both x axis values and y axis values using "dragmode: select". Perhaps dragmode can be extended to 'select', 'selectx' and 'selecty', to respectively allow the user to select only horizontally or vertically.

etpinard:
That would be nice.

GlennMatthys:
Discussion in PR #2498, implementation in PR #2506.

etpinard:
Done in https://github.com/plotly/plotly.js/pull/2506

PR #2498, opened 2018-03-26
Allow clamping select mode to horizontal or vertical

RedShift1:
Here's some shitty code for feature request #2495. Use layout.dragclamp 'h' or 'v' to clamp the selection horizontally or vertically. There's some duplicated code now, perhaps those polygon manipulations should be moved into their own functions?

### DIFF
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -167,8 +167,7 @@ module.exports = function prepSelect(e, startX, startY, dragOptions, mode) {
             dy = Math.abs(y1 - y0);
 
         if(mode === 'select') {
-            if(dy < Math.min(dx * 0.6, MINSELECT)) {
-                // horizontal motion: make a vertical box
+            if(gd.layout.dragclamp === 'h') {
                 currentPolygon = [[x0, 0], [x0, ph], [x1, ph], [x1, 0]];
                 currentPolygon.xmin = Math.min(x0, x1);
                 currentPolygon.xmax = Math.max(x0, x1);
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Thanks for the PR! You're right though, we can do a bit better - first off, lets think about the attribute name: it doesn't apply to all drags, just selection drags; clamp may be OK, though I kind of like the sound of `selectdirection` or `selectaxis` better than `selectclamp`... dunno, @etpinard any other thoughts?

Next, we need `layout.selectclamp` (or whatever we call it) to be a recognized attribute [here](https://github.com/plotly/plotly.js/blob/master/src/components/fx/layout_attributes.js) as `valType: 'enumerated'` with values `['h', 'v', '']` (or perhaps `['h', 'v', 'any']`?), coerced from `layout` into `fullLayout` [here](https://github.com/plotly/plotly.js/blob/master/src/components/fx/layout_defaults.js#L19).

Then, as you anticipated, we should avoid duplication. Something like:
```js
var selectDirection = fullLayout.selectclamp;
if(!selectDirection) { // or !== 'any' if we go with that
    if(dy < Math.min(dx * 0.6, MINSELECT)) selectDirection = 'h';
    else if(dx < Math.min(dy * 0.6, MINSELECT)) selectDirection = 'v';
}

if(selectDirection === 'h') {
   // existing code
}
else if(selectDirection === 'v') {
    // ...
}
else {
    // ...
}
```

Lastly we'll need a test that covers the new functionality, probably [here](https://github.com/plotly/plotly.js/blob/9a79056bb223e13995be8debd22f60c9d56cbef0/test/jasmine/tests/select_test.js#L141) by dragging a diagonal (that would normally give a box constraining both dimensions) after having `relayout` into each mode and verifying (`assertRange(selectedData.range, ...)`) that you only constrain the expected dimension.

RedShift1:
Once you decide on a name I'll create seperate PRs for
* Adding the attribute to the layout object
* Moving the polygon calculations into separate functions
* Re-implement the logic switching between clamped/non-clamped
* Tests

etpinard:
+1 for `selectdirection: 'h' | 'v' | 'any' (the default)`

alexcjohnson:
superseded by #2506

PR #2493, opened 2018-03-23
point-cluster for scattergl

dfcreative:
This PR includes improvement on regl-scatter2d & introduces point-clustering.
Btw selection, panning & updating are significantly faster.
Clustering is somewhat faster.
Introducing async clustering - later.

cc @etpinard

PR #2499, opened 2018-03-26
Point cluster dev

dfcreative:
supersedes  #2493


etpinard:
image

looks like you brought in some non-ES5 tokens making our image tests 100% not functional.

dfcreative:
@etpinard is the _test-image_ build fail caused by scattergl update? Looks like something other causes that.

etpinard:
you mean this:

image

see -> https://github.com/plotly/plotly.js/pull/2493#issuecomment-375717494

etpinard:
Looks like 

image

is also having issues in nw.js

---- 

@dfcreative Can you try setting up https://gist.github.com/etpinard/d27a44bd5dbee5490f20 ?

You don't anything docker related. Just make sure to use `nw.js` version 0.12  -> https://dl.nwjs.io/v0.12.0/

src/traces/scatterpolargl/index.js
```diff
@@ -106,6 +106,12 @@ function plot(container, subplot, cdata) {
         if(options.line && !scene.line2d) scene.line2d = true;
         if((options.errorX || options.errorY) && !scene.error2d) scene.error2d = true;
 
+        stash.tree = cluster(positions);
+
+        if(options.marker) {
```
etpinard: this should probably be `options.marker && count >= TOO_MANY_POINTS` like in `scattergl/convert.js`

etpinard:
@dfcreative nice! You got the tests to :white_check_mark: 

----

I noticed a few things by looking at the diff:

- `scatterpolargl` tree logic isn't the same as `scattergl` https://github.com/plotly/plotly.js/pull/2499/files#r177424508

- colors in gl2d_scatter-color-clustering are a little different. Is this on purpose?

peek 2018-03-27 09-33

- there's a lot more `@flaky` scattergl tests. Can you think of any reason what that's the case? This added flakiness might be hidding real problems - as @alexcjohnson noticed before in https://github.com/plotly/plotly.js/pull/2415

- can you please check that this new `regl-scatter2d` version is at least as good as the previous in terms of memory consumption?

etpinard:
@dfcreative I'm noticing that drag performance slows down after toggling `layout.dragmode` from 'drag' to 'select' back to 'drag' . Is this expected? 

Moreover, with 

```js
var x = []
var y = []
var ms = []
var N = 1e6
var i = 0

for(; i < N; i++) {
  x.push(Math.random())
  y.push(Math.random())
  ms.push(i % 2 ? 20 : 10)
}

console.time('plot')
Plotly.newPlot(gd, [{
  type: 'scattergl',
  mode: 'markers',
  x: x,
  y: y,
  marker: {size: ms}
}], {
  dragmode: 'closest',
  //xaxis: {range: [-0.1, 2.2]},
  //yaxis: {range: [-0.1, 1.1]}
}, {scrollZoom: true})
```

selection is slower than on `master` (see https://codepen.io/etpinard/pen/QQBzyV). Is this expected for 1e6 random points?

etpinard:
... Oh and we should make sure the new `regl-scatter2d` works in IE11.

dfcreative:
> scatterpolargl tree logic isn't the same as scattergl
Fixed. Although that affects only data layering order, for polar plots that is a bit less actual.

> colors in gl2d_scatter-color-clustering are a little different. Is this on purpose?
That is not clustering, that is adjusted shader behavior in case of no-border https://github.com/dfcreative/regl-scatter2d/commit/023c421c42ee52757f56e42d42dafc739151dc3d. It used to add thin black pixel outline instead of keeping the point color, now that "antialias" has the point color.

> flakiness
`test-jasmine2` passed or failed sporadically, regardless of changes. According to errors, `fullLayout` is `undefined`.

> IE11
Thanks, fixed. In fact it supports multiple colors now in IE and works even better.
image

> Memory
> Selection performance

Working on that.

dfcreative:
Ok, with the latest commit selection should work almost as fast as possible, init time is ok, no tree overcalculation and observable memory leaks.

package.json
```diff
@@ -86,8 +86,8 @@
     "glslify": "^6.1.1",
     "has-hover": "^1.0.1",
     "has-passive-events": "^1.0.0",
-    "kdgrass": "^1.0.1",
     "mapbox-gl": "0.44.1",
+    "math-log2": "^1.0.1",
```
alexcjohnson: did this get used?

dfcreative: Whoops.

etpinard:
Well this is amazing! Performance off the latest commit is nothing short of spectacular. Great job @dfcreative :tada: All my concerns have been addressed.

Now, would you mind publishing a major version of `regl-scatter2d` (that's `v3.0.0` I think) and `point-cluster` and update the package.json and package-lock.json accordingly? Once that's done, we'll merge this thing right up and move on the sploms :rocket:

etpinard:
Ok. Well-earned :dancer: Let's merge this thing.

Issue #2500, opened 2018-03-26
Multiple color/opacity scattergl selection is broken


dfcreative:
Selection in `gl2d_scatter-color-clustering.json` mock does not work.

`trace.selected` does not get populated [here](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js#L42).

etpinard:
fixed in #2503

PR #2503, opened 2018-03-27
Fix scattergl unselected marker opacity for array marker opacity traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2500 by handling the case where `marker.opacity` is an array and no `[un]selected` attributes are set (more info [here](https://github.com/plotly/plotly.js/blob/c3852183a1dd8b080e3c0b6ff872ab15bc656518/src/lib/coerce.js#L476-L480)). To be merged in https://github.com/plotly/plotly.js/pull/2499 to not have to regenerate the baselines once in master.

cc @dfcreative @alexcjohnson


test/image/mocks/gl2d_selectedpoints.json
```diff
@@ -0,0 +1,14 @@
...
+    "name": "array marker opacity edge case",
+    "type": "scattergl",
+    "mode": "markers",
+    "x": [1, 2, 3, 4, 5, 6],
+    "y": [1, 3, 2, 4, 5, 7],
+    "marker": {
+      "size": 20,
+      "opacity": [0.9, 0.8, 0.7, 1, 0.6, 0.8]
+    },
+    "selectedpoints": [1, 4, 2]
```
etpinard: On master, this gives

image

where `selectedpoints` is essentially ignored.

src/traces/scattergl/convert.js
```diff
@@ -41,6 +42,14 @@ function convertStyle(gd, trace) {
         opts.marker = convertMarkerStyle(trace);
         opts.selected = convertMarkerSelection(trace, trace.selected);
         opts.unselected = convertMarkerSelection(trace, trace.unselected);
+
+        if(Array.isArray(trace.marker.opacity)) {
+            var mo = trace.marker.opacity;
+            opts.unselected.opacity = new Array(mo.length);
```
alexcjohnson: what if `trace.unselected.marker.opacity` is already defined?

etpinard: I guess that would override it. Good eye. Thanks!

test/image/mocks/gl2d_selectedpoints.json
```diff
@@ -10,5 +10,24 @@
...
+    "mode": "markers",
+    "x": [1, 2, 3, 4, 5, 6],
+    "y": [3, 5, 4, 6, 7, 9],
+    "marker": {
+      "size": 20,
+      "opacity": [0.9, 0.8, 0.7, 1, 0.6, 0.8]
+    },
+    "unselected": {
+      "marker": {
+        "opacity": 0
```
etpinard: Before this commit, this gave:

image

that is `unselected.marker.opacity` was overrode.

alexcjohnson:
Looks great 💃

Issue #778, opened 2016-07-26
Smarter defaults for layout.hovermode


joshua-gould:
Please see http://codepen.io/anon/pen/akKOvE for an example.

etpinard:
Confirmed. Thanks.

etpinard:
Looking at this issue again, I'm thinking this is the desired behavior under `layout.hovermode: 'x'` which is the default in your case because not all traces are _horizontal_ (see logic [here](https://github.com/plotly/plotly.js/blob/32ccae5a5e86bcd8c974fea923436223d6fb723d/src/plots/cartesian/graph_interact.js#L46-L54)).

Setting `layout.hovermode` to either `'y'` or `'closest'` gives the desired behavior.

alexcjohnson:
setting `layout.hovermode` to `'y'` in this case mucks up hover for the rug trace, so that's not a good option. But I think what this points to is that "compare" mode is only a good default hover mode in certain (albeit very common) cases - where there are multiple traces of the same type with monotonic data mostly in common for the "compare" axis. What we should really do is default to `'closest'` if any of those conditions is not met. That would help in even relatively simple cases like one or more scatter traces with nonmonotonic points (and perhaps we could even figure out if it's `y` that's monotonic instead of `x` and switch to `hovermode: 'y'`). Renaming this issue accordingly.

etpinard:
Adding this to the v2 milestone

archmoj:
We could possibly close this now that `hovermode` is going to default to `closest` in v2. Right?

nicolaskruchten:
Yes.

PR #2505, opened 2018-03-28
Introducing splom traces

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2372

This is a feature branch built off PRs https://github.com/plotly/plotly.js/pull/2474, https://github.com/plotly/plotly.js/pull/2487 and https://github.com/plotly/plotly.js/pull/2499 where all splom development should start off from. Other splom-related branches on `origin` like `splom` and `splom-dev` are now obsolete.

cc @dfcreative @alexcjohnson


etpinard:
As of the commit fda1b34, 

### things that work

- splom markers thanks to https://github.com/dfcreative/regl-scattermatrix and their styling thanks to 1c6144e291d3d1e191211f4f208672c224bfce4a (I haven't tested `arrayOk` attributes though)
- zoom and drag
- graph with multiple splom traces
- splom traces that generate `layout.grid`
- splom traces that default their corresponding axis title
- is pretty fast (more on that soon)

### things that do not work

- image generation in nw.js (that's why the new splom mock don't have corresponding baselines). Note that image generation via the modebar button appear to work fine.
- something is off with splom + other cartesian trace types
- hover
- selection
- legend items styling
- `mode: 'lines'`
- splom attributes `xdirection`, `ydirection`, `showdiagonal`, `showupperhalf` and `showlowerhalf` (or shouldn't that be a flaglist by the way :thinking: )
- probably more things I can't think about right now

---

Here's what the `splom_iris` mock looks like:

peek 2018-03-27 22-58

src/plot_api/plot_schema.js
```diff
@@ -486,7 +486,7 @@ function getLayoutAttributes() {
 
         if(!_module.layoutAttributes) continue;
 
-        if(_module.name === 'cartesian') {
+        if(_module.name === 'cartesian' || _module.name === 'splom') {
```
alexcjohnson: Can we do this just based on `_module.attr`?
`if(Array.isArray(_module.attr)) { _module.attr.forEach(... handleBasePlotModule) }`

src/traces/splom/attributes.js
```diff
@@ -0,0 +1,121 @@
...
+        description: ''
+    },
+    ydirection: {
+        valType: 'enumerated',
+        values: ['top', 'bottom'],
+        dflt: 'bottom',
+        role: 'info',
+        editType: 'plot',
+        description: ''
+    },
```
alexcjohnson: `layout.grid` doesn't have an analog of `xdirection`, do we need that? And in place of `ydirection` we have `roworder: ('top to bottom'|'bottom to top')` which feels clearer to me than `('top'|'bottom')`. Actually I wonder if these belong in `splom` attributes at all, or if we should just let `layout.grid` handle it. It's nice for the simple single-trace case, but once you get two `splom` traces, seems like only one (the first?) matters, so having one in each trace would be confusing.

src/traces/splom/attributes.js
```diff
@@ -0,0 +1,121 @@
...
+    xaxes: makeAxesValObject('x'),
+    yaxes: makeAxesValObject('y'),
+
+    showdiagonal: {
+        valType: 'boolean',
+        role: 'info',
+        dflt: true,
+        editType: 'plot',
+        description: ''
+    },
```
alexcjohnson: Eventually we will want the diagonals to support various types: scattergl, histogram, box, violin. So I'm wondering if instead of `showdiagonal` we should have `diagonal: {visible: (true|false)} to support future expansion:
```js
diagonal: {
    visible: (true|false),
    type: (scattergl|histogram|box|violin),
    // then attributes specific to each type, as well as some overrides per dimension
}
```

src/traces/splom/attributes.js
```diff
@@ -0,0 +1,121 @@
...
+    showlowerhalf: {
+        valType: 'boolean',
+        role: 'info',
+        dflt: true,
+        editType: 'plot',
+        description: ''
+    },
```
alexcjohnson: If we get rid of `showdiagonal` I'm not sure it makes sense to turn these two into a little flaglist - perhaps just leave them as two booleans.

src/traces/splom/defaults.js
```diff
@@ -0,0 +1,137 @@
...
+'use strict';
+
+var Lib = require('../../lib');
+
+var attributes = require('./attributes');
+var subTypes = require('../scatter/subtypes');
+var handleMarkerDefaults = require('../scatter/marker_defaults');
+var handleLineDefaults = require('../scatter/line_defaults');
+var PTS_LINESONLY = require('../scatter/constants').PTS_LINESONLY;
+var OPEN_RE = /-open/;
```
alexcjohnson: Again, going to [break for symbol numbers](https://github.com/plotly/plotly.js/pull/2492#discussion_r177247877)

src/traces/splom/defaults.js
```diff
@@ -0,0 +1,137 @@
...
+
+    for(i = 0; i < dimLength; i++) {
+        xaxesDflt[i] = 'x' + (i ? i + 1 : '');
+        yaxesDflt[i] = 'y' + (i ? i + 1 : '');
+    }
+
+    var xaxes = coerce('xaxes', xaxesDflt);
+    var yaxes = coerce('yaxes', yaxesDflt);
+
+    // TODO what to do when xaxes.length or yaxes.length !== dimLength ???
```
alexcjohnson: I guess if `dimLength` is longer it's easy to just drop the later dims. What if `xaxes.length !== yaxes.length` or `dimLength` is shorter than `xaxes` or `yaxes`? The *easy* thing to do is just truncate them all to the same length, and perhaps until we have a clear use case otherwise we should just do that.

Related though: if you have several `splom` traces but they don't share all the same dimensions, you could of course handle this by specifying `xaxes` and `yaxes` for each, but could you also just pass in an empty dimension for ones you want to skip? If that takes more than 5 minutes to implement though, skip it for now and leave it as an item for future expansion.

test/image/mocks/splom_iris.json
```diff
@@ -691,7 +691,6 @@
   "layout": {
 	"title": "Iris dataset splom",
   	"width": 600,
-	"height": 500,
-	"showlegend": false
+	"height": 500
```
etpinard: first cut baselines :tada: 

cc @dfcreative

src/traces/splom/defaults.js
alexcjohnson: Oh you already have `dimensions[i].visible: false`, nice! That should take care of my "related" point, it'll still reserve x/y axes for these dimensions, right?

test/image/mocks/splom_iris.json
etpinard: ... notice how the legend items aren't showing the correct symbol. Fixing that.

etpinard: Ignore last comment. Something else fixed that.

test/image/mocks/splom_with-cartesian.json
```diff
@@ -0,0 +1,38 @@
+{
+  "data": [{
+    "type": "splom",
+    "name": "splom",
```
etpinard: one for @alexcjohnson :tada:

src/traces/splom/attributes.js
etpinard: > Actually I wonder if these belong in splom attributes at all, or if we should just let layout.grid handle it. It's nice for the simple single-trace case, but once you get two splom traces, seems like only one (the first?) matters, so having one in each trace would be confusing.

I agree 100% here. I'm thinking of :hocho: ing `xdirection` and `ydirection`. Users won't have a way to plot their splom dimensions from right to left until `layout.grid.columnorder` is implemented, but I don't think that's a big deal. Any objections?

src/traces/splom/attributes.js
etpinard: > Eventually we will want the diagonals to support various types: scattergl, histogram, box, violin.

Interesting. So you're thinking a single splom trace could generate scatter panels AND histogram or box or violin traces on the diagonal? That would put us on-par with some of the seaborn. Yeah nice. I can't really think of any drawbacks from that approach except this might cause headaches to the workspace team.

src/traces/splom/defaults.js
etpinard: > The easy thing to do is just truncate them all to the same length, and perhaps until we have a clear use case otherwise we should just do that.

That sounds right to me. I'll go with that unless anyone disagrees.

> That should take care of my "related" point, it'll still reserve x/y axes for these dimensions, right?

Yes that _should_ take care of it, but I haven't tried that yet. Putting that on my TODO list :memo:

etpinard: .... thanks for all these comments by the way. I hope you don't the `_hasOnlyLargeSplom` flag too hacky.

src/traces/splom/attributes.js
alexcjohnson: > a single splom trace could generate scatter panels AND histogram or box or violin traces on the diagonal?

Yes exactly. An extra complication is that this would create even more x or y axes (for the histogram count or box/violin category), depending on the type and orientation of these diagonal plots; but once we do that, it seems like the implementation should be fairly straightforward. But anyway none of that has to happen now, as long as we set up the attributes we DO need now so they'll support this.

src/components/fx/hover.js
```diff
@@ -339,8 +338,15 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         // the rest of this function from running and failing
         if(['carpet', 'contourcarpet'].indexOf(trace._module.name) !== -1) continue;
 
-        subplotId = helpers.getSubplot(trace);
-        subploti = subplots.indexOf(subplotId);
+        if(trace.type === 'splom') {
+            // splom traces do not generate overlay subplots,
+            // it is safe to assume here splom traces correspond to the 0th subplot
+            subploti = 0;
+            subplotId = subplots[subploti];
```
alexcjohnson: OK for now, but we should remove this restriction at some point - it would be weird to overlay a splom on something else, but you never know what weird use cases users will come up with.

src/traces/splom/index.js
```diff
@@ -183,24 +184,81 @@ function plotOne(gd, cd0) {
...
+        if(trace.yaxes[i] === ya._id) yi = i;
+    }
+
+    var x = dimensions[xi].values;
+    var y = dimensions[yi].values;
+
+    var id, ptx, pty, dx, dy, dist, dxy;
+    var minDist = maxDistance;
+
+    if(hovermode === 'x') {
```
alexcjohnson: It's a little weird to use compare mode for splom - I guess maybe there could be uses for it, but especially since the same data will appear as both x and y, compare mode will confuse people. Eventually we should default to `closest` when there's a splom, see #778. For now lets just encourage splom users to specify `closest` explicitly.

dy:
Spotted a bug. With `regl.clear` or `dirty` flag refactoring switching to select mode now does now clear properly. Needs investigation.
capture

src/components/fx/helpers.js
```diff
@@ -11,10 +11,30 @@
...
 
 // look for either subplot or xaxis and yaxis attributes
+// does not handle splom case
 exports.getSubplot = function getSubplot(trace) {
     return trace.subplot || (trace.xaxis + trace.yaxis) || trace.geo;
 };
 
+// is trace in given list of subplots?
+// does handle splom case
+exports.isTraceInSubplots = function isTraceInSubplot(trace, subplots) {
```
dy: :pray:

src/traces/splom/index.js
```diff
@@ -220,30 +220,16 @@ function hoverPoints(pointData, xval, yval, hovermode) {
     var id, ptx, pty, dx, dy, dist, dxy;
     var minDist = maxDistance;
 
-    if(hovermode === 'x') {
```
alexcjohnson: OK sure, that'll make people happier until we do a better job with defaults - just make a comment about why we're omitting `'x'`, so we don't forget about this issue.

src/components/fx/helpers.js
alexcjohnson: Not sure how this will work with `show(upper|lower)half` - actually, it's probably better to let it pass here - in fact in case there's ever another trace type that spans multiple subplots, we could do `if(trace.xaxes && trace.yaxes)` instead of `if(trace.type === 'splom')` - and then bail out during `hoverPoints` for the trace if that subplot is blank.

src/traces/splom/index.js
dy: Removed for now.

src/traces/splom/defaults.js
etpinard: -> move to https://github.com/plotly/plotly.js/issues/2507 thanks!

src/traces/splom/defaults.js
etpinard: done in 07d59d553bc4bd5c762a72a6e8e34b0f489d0e80

dy:
Selection tests to cover:
* [ ] `trace.select` but no `trace.unselect`
* [ ] selection should be reflected in every subplot
* [ ] proper select event data

etpinard:
@dy @alexcjohnson commits https://github.com/plotly/plotly.js/pull/2505/commits/74abb7fe9cb915505d4d2df9c3b710e8d9ac6c1b and https://github.com/plotly/plotly.js/pull/2505/commits/062d4a3195920e2769d3f45da65dadddde97ed7a are a little tricky. splom and scattergl can interact in weird ways on select as scattergl clears only one subplot and splom clears the entire canvases.

src/plots/sort_modules.js
```diff
@@ -0,0 +1,33 @@
...
+'use strict';
+
+function sortBasePlotModules(a, b) {
+    var nameA = a.name;
+    var nameB = b.name;
+
+    // always plot splom before cartesian (i.e. scattergl traces)
+    if(nameB === 'splom' && nameA === 'cartesian') {
+        return 1;
+    }
```
alexcjohnson: I don't think this is sufficient... if there are more than those two modules you're not even guaranteed that `sort` will ever be called with that exact pair; and if it is called with that pair, you're not guaranteed that they'll be in that order, nor that [`0` will preserve order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

Seems to me you need to either put `splom` before everything, or `cartesian` after everything, eg:
```js
if(nameB === 'splom') return 1;
if(nameA === 'splom') return -1;
return 0;
```

etpinard: thanks for the tip!

etpinard:
Here's what happens with when allowing cross-subplot selections (like we have currently on `master`):

peek 2018-04-12 11-46

Three things to notice:

- the event data `pointNumber` values correspond to the pt indices in every `dimensions[i].values`. I think that's correct. Attribute `selectedpoints` follows the same rule.
- the selection outlines do not match the selected points when selecting across subplots.
- multi-select works per subplot, trying to multi-select across subplot doesn't work at the moment.

With commits https://github.com/plotly/plotly.js/pull/2505/commits/71a0395044e42b3c69cdedb36bc14f5f962601ce and https://github.com/plotly/plotly.js/pull/2505/commits/9d0bcfd1fb6f5699feb73f2e1ccac880f2fa7b15 we have a slightly more consistent (albeit incomplete) behavior:

peek 2018-04-12 11-50


In brief, with these commits:

- we clear selection outlines when selecting in a different subplot than the previous selection
- we disallow multi-select across subplots

src/components/fx/helpers.js
etpinard: > Not sure how this will work with show(upper|lower)half

Turns out this works fine for splom traces with `show(upper|lower)half` and `diagonal.visible: false` as hovering over their omitted subplots never triggers `Fx.hover` as we don't draw those subplots and more precisely we don't draw their corresponding `<g .draglayer>`.

If we choose at some point to draw only one drag layer for all interaction types for all subplots, we'll of course have to change this logic here. But if ok, I'll leave this out of this PR and list it in an upcoming _splom open items_ issue.

src/components/fx/hover.js
etpinard: If ok, I'd leave out splom trace on overlaid axes from this PR.

src/plots/sort_modules.js
etpinard: fixed in https://github.com/plotly/plotly.js/pull/2505/commits/31e34fac459e22ececca32bdc4583378741b2780

test/jasmine/tests/plots_test.js
```diff
@@ -154,6 +153,26 @@ describe('Test Plots', function() {
...
+            });
+
+            expect(names).toEqual([
+                'splom',
+                'cartesian',
+                'gl3d',
+                'geo',
+                'pie',
+                'ternary'
+            ]);
```
alexcjohnson: great test!

etpinard:
with commit https://github.com/plotly/plotly.js/pull/2505/commits/a79f5dc1fc29c2c08f901e888b00fba908121fc4 splom interactions are now :lock:ed , I'll tag this PR as reviewable. 

So, here's an updated pre-merge checklist:

- [ ] maybe tweak some test values for CI
- [ ] approval from @alexcjohnson 
- [ ] test @dy `regl-line2d` update
- [ ] adapt `_splomGrid` code for new `regl-line2d` (if necessary)
- [ ] release / publish new versions of `regl-scattermatrix`, `regl-scatter2d` and `regl-line2d`
- [ ] update package.json and package-lock.json

Note that I just made two comments:

- https://github.com/plotly/plotly.js/pull/2505#discussion_r181237581
- https://github.com/plotly/plotly.js/pull/2505#discussion_r181237788

arguing for leaving some previous unaddressed in this PR.

alexcjohnson:
As long as all the TODOs added here are linked to the upcoming splom open items issue(s) I agree, they're all fine to omit from this PR. Looking amazing guys! 💃

etpinard:
Here are some first-render benchmarks with `showupperhalf: false` and `diagonal.visible: false`:

| # of dimensions | # of values per dimension | regl-line2d grid lines? | time [ms] |
| --------------------- | ---------------------------------- | -------------------- | ------------ |
| 10 | 20k | no | 400-500 |
| 20 | 5k | yes | 600-700 |
| 30 | 2k | yes | 900-1000 |
| 40 | 1k | yes | 1300-1400 |
| 50 | 800 | yes | 2000-2300 |

computed using:

```js
var Nvars = /* fill this in */
var Nrows = /* fill this in */
var dims = []

for(var i = 0; i < Nvars; i++) {
  dims.push({values: []})
  
  for(var j = 0; j < Nrows; j++) {
     dims[i].values.push(Math.random())
  }
}

Plotly.purge(gd);

console.time('splom')
Plotly.newPlot(gd, [{
  type: 'splom',
  dimensions: dims,
  showupperhalf: false,
  diagonal: {visible: false},
}], {width: 2000, height: 2000})
console.timeEnd('splom')
```

src/traces/splom/attributes.js
amalinafz: hi @alexcjohnson, i'm really interested to use the diagonal type. Is it available to use?

alexcjohnson: Currently the diagonal can only have scatter type. Follow #2555 for more info, but we do not have any immediate plans for when other types will be added.

amalinafz: oh okay. Thanks

PR #2506, opened 2018-03-29
Configurable select direction

GlennMatthys:
Proper implementation of #2495. This PR includes:
* Adding the "selectdirection" attribute to layout
* Implementation in Cartesian select
* Image mocks for the various "selectdirection" options

Also adds a "d" selectiondirection to only allow diagonal selects.

Working on writing a proper test for this functionality, I'm still figuring out how to do the necessary calculations.


src/components/fx/layout_attributes.js
```diff
@@ -101,5 +101,17 @@ module.exports = {
...
+    },
+    selectdirection: {
+        valType: 'enumerated',
+        role: 'info',
+        values: ['h', 'v', 'd', 'any'],
+        dflt: 'any',
+        description: [
+            'When "dragmode" is set to "select", this limits the selection of the drag to',
+            'horizontal, vertical or diagonal. "h" only allows horizontal selection,',
+            '"v" only vertical, "d" only diagonal and "any" sets no limit.'
```
alexcjohnson: Nice, I like `'d'`, I guess what it allows that `'any'` doesn't is selecting regions that are very narrow in height or width, which would otherwise devolve to pure horizontal or vertical selections. Cool.

src/components/fx/layout_defaults.js
```diff
@@ -17,6 +17,7 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
     }
 
     coerce('dragmode');
+    coerce('selectdirection', 'any');
```
alexcjohnson: since the attribute has `dflt: 'any'` already, you do not need to include it here.

test/image/mocks/limit_select_any.json
```diff
@@ -0,0 +1,15 @@
...
+    { "x": [1,2,3], "y":[-1,-2,-3], "type": "bar" },
+    { "x": [1,2,3], "y":[-1,-2,-3], "type": "bar", "yaxis":"y2" },
+    { "x": [1,1,1,1,1,2,2,2,3,3], "type": "histogram", "yaxis":"y2" }
+  ],
+  "layout": {
+    "yaxis": { "domain": [0,0.49] },
+    "yaxis2": { "domain": [0.51, 1] },
+    "showlegend": false,
+    "dragmode": "select",
+    "selectdirection": "any"
```
alexcjohnson: I don't think we need to make image test mocks for `selectdirection`, since it has no effect on the static images. For the jasmine tests, if you don't like the mock ([14](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/14.json)) that we use in the neighboring tests, you can use any other predefined mock, or just make a new plot explicitly in the test, something like:
```js
Plotly.newPlot(gd, [
    {x: [1,1,1,1,1,2,2,2,3,3], type: 'histogram'},
    {x: [1,2,3], y:[-1,-2,-3], type: 'bar'}
], {
    showlegend: false,
    dragmode: 'select'
})
.then(function() {
    expect(gd.fullLayout.selectdirection).toBe('any');

    // do a drag and check that the selection is correct for the default `selectdirection`

    return Plotly.relayout(gd, {selectdirection: 'h'});
})
.then(function() {
    // do probably the SAME drag and check that the selection is correct for the `selectdirection='h'`

    return Plotly.relayout(gd, {selectdirection: 'v'});
})
.then(function() {
    // ...
})
.catch(fail) // hmm, we should change the name of this import to failTest, since fail is already a jasmine global, for sync failure, we're overriding it here with our async version (we've already done this in other places)
.then(done);
```

alexcjohnson:
> I'm still figuring out how to do the necessary calculations.

While it's nice to be able to write a test where you've pre-calculated what the outcome should be, it's not always necessary: as long as you can demonstrate a difference between the different modes, you can just write tests, watch them fail when you run them, and change them so they pass. But if you do want to do the calculation yourself, I've had good luck setting explicit `width`, `height`, `margin`, and axis `range` so you can make an easy calculation of data value -> pixel position.

GlennMatthys:
@alexcjohnson can you have a look at [this commit](https://github.com/plotly/plotly.js/pull/2506/commits/c49d829dd4b9fa5ead634039dbe9217b1647b4c2) to see if I'm on the right track?

test/jasmine/tests/select_test.js
```diff
@@ -552,6 +552,54 @@ describe('@flaky Test select box and lasso in general:', function() {
...
+                    expect(selectedData.points.length).toBe(1);
+                    expect(selectedData.points[0].x).toBe(1.8);
+                    expect(selectedData.points[0].y).toBe(3);
+                });
+
+                // do a drag and check that the selection is correct for the default `selectdirection`
+
+                return Plotly.relayout(gd, {selectdirection: 'h'});
+            })
+            .then(function() {
```
alexcjohnson: Yep, this looks good! Make sure you keep chaining the async events - your `selectedPromise.then...` is not being waited for by the following code so it's not entirely clear what sequence things will happen in. Try something like:
```js
Plotly.newPlot(...)
.then(function() {
    resetEvents(gd);
    drag(selectPath);
    return selectedPromise;
})
.then(function() {
    expect(selectedData.points.length).toBe(1);
    // etc

    return Plotly.relayout(...);
})
```

GlennMatthys:
Here's a tool I used to help create the tests: https://jsfiddle.net/c0ojh6vq/8/ (just shows you the mouse hover position on the drag layer).

GlennMatthys:
The CI test appears to be failing on an unrelated part before the selectdirection tests ("Test polar interactions: should response to drag interactions on plot area FAILED"), are these tests known to fail from time to time?

alexcjohnson:
Oh sorry about letting this one slip through last week - thanks for pinging me again. Yes, the test failure is unrelated, we should probably mark that test as `@flaky` as I have seen that fail spuriously before.

Anyway I could rerun it for you but it still looks like the promise chain in your test is branched - see https://github.com/plotly/plotly.js/pull/2506#discussion_r179503484 - can you rearrange so every `.then` is chained off the original `Plotly.newPlot`? The ['should skip over BADNUM items'](https://github.com/ECraneWorldwide/plotly.js/blob/69c3157dce84e5a4e45c98b603b096d5a6fa2157/test/jasmine/tests/select_test.js#L495) test is a good reference for this structure.

alexcjohnson:
@etpinard care to take a quick look?

etpinard:
image

one more for the flaky gang?

GlennMatthys:
I wanted to finish this up but I'm struggling to get the tests to succeed. How do you calculate the select path? To which element are these coordinates relative to? Is there some offset that needs to be taken into account?

src/components/fx/layout_defaults.js
```diff
@@ -17,6 +17,7 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
     }
 
     coerce('dragmode');
+    coerce('selectdirection');
```
etpinard: This only has an effect for `dragmode: 'select'`, correct? So this be:

```js
var dragMode = coerce('dragmode');
if(dragMode === 'select') {
  coerce('selectdirection')
}
```

alexcjohnson:
@GlennMatthys I think it's all set now, check out https://github.com/plotly/plotly.js/pull/2506/commits/6d4a4f82fad6e3ca427dafa5781bd21e3cd015b9

But for reference, if you're contributing more in the future, it's relative to the corner of the plot, including the margins. That's why in that commit you can see I set the margins, size, and axis ranges explicitly so I know x is 100px->0, 300px->4, and y is the opposite 100px->4, 300px-> 0

src/plots/cartesian/select.js
```diff
@@ -192,6 +192,8 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
         }
     }
 
+    var direction = fullLayout.selectdirection;
```
etpinard: `direction` doesn't get used in the outer scope. Could be move that `var direction` into the `moveFn` scope below?

alexcjohnson: good call -> pushed to inner scope https://github.com/plotly/plotly.js/pull/2506/commits/9dd166c69bd654bd0a70dac6d2f0cb4f3d6c7235

src/components/fx/layout_defaults.js
alexcjohnson: yep -> conditional coerce https://github.com/plotly/plotly.js/pull/2506/commits/9dd166c69bd654bd0a70dac6d2f0cb4f3d6c7235

alexcjohnson:
Not sure if CI is extra-flaky today or if splom and other new features are causing it to bog down... but anyway I moved those new shape tests to `@flaky` as well... 🤞

etpinard:
Nicely done :dancer:

GlennMatthys:
Can the coordinate system for selections be documented somewhere, perhaps in the contributing guide? This should help newcomers orient themselves, plus the tips like fixing the margins and axis ranges to produce a much more reliable canvas to do select tests on. Some visual representation would be nice too:

plotly_test_select

alexcjohnson:
> Can the coordinate system for selections be documented somewhere, perhaps in the contributing guide?

That's a good idea - @etpinard where would be the right place for this?

I like your diagram. Note that [175, 175] and [225, 225] go one (data) unit down and right of where you have them, ie 175 is actually where you've drawn 225.

GlennMatthys:
Made a new PR so this can be discussed more appropriately, also fixed the diagram.

PR #2511, opened 2018-03-30
Support ragged tables and show full table borders

### LINKED ISSUE #2133: autofill header title
It would be very useful to be able to make table headers of unequal lengths, and have plotly.js fill the gaps

ex: https://codepen.io/veraz/pen/RjNEwE

### LINKED ISSUE #2401: Table borders sometimes get cut off
Sometimes a table can get one of its borders cut off - not sure if it's being clipped or crisp-rendered or what:
table bug
And if I move it a little bit, I can get the other side to (almost) disappear:
screen shot 2018-02-23 at 5 22 26 pm

Discovered as part of https://github.com/plotly/plotly.js/pull/2399 - specifically in the [grid_subplot_types](https://github.com/plotly/plotly.js/blob/82fbae247ae38a087a012673face02970f76603c/test/image/mocks/grid_subplot_types.json) mock

alexcjohnson:
Fixes #2133 - allows ragged columns for both headers and cells, and even a mix of 1D and 2D - see the [table_ragged mock](https://github.com/plotly/plotly.js/compare/ragged-table?expand=1#diff-a3328e7369b3d7608647573bbb1909fa)

Fixes #2401 - how some sides of the table were often missing their border - I extended the `clippath` around each column to include space for the cell borders. This has some strange effects right now if you use different border widths per cell, see eg cell `5` in `table_ragged`:
screen shot 2018-03-30 at 9 36 36 am
On master this still isn't what I'd call "right", but its quirks are at least a little subtler:
screen shot 2018-03-30 at 10 07 58 am
Semi-opaque borders (on this branch) do even weirder things:
screen shot 2018-03-30 at 9 40 36 am
That said semi-opaque borders already behave oddly on master:
screen shot 2018-03-30 at 9 42 28 am

I'm not sure it's worth doing any better right now, but some thoughts for future reference:
- In order to fully display the borders where they overlap, we'd need to display *all* borders above *all* cell backgrounds - ie a separate border layer, with its own infinite scrollers etc... seems like more trouble (and overhead) than this edge case really warrants.
- It's still ambiguous what order to draw the borders in - perhaps we want wider borders to override thinner ones, so we order them by thickness? That takes care of one common use case, using a thicker border to highlight a cell. But it doesn't cover using a different *color* to highlight a cell - actually I didn't explore array line colors at all when making this PR.
- Another possibility I considered - both to fix #2401 and to clear up ambiguities - is to move the border to fully inside each cell. Then none of the borders would overlap, so the semi-transparent issue goes away, there's no ordering, and no clippath extension needed. But it messes up the *most* common case, of a consistent border everywhere, by doubling the effective width of interior lines vs peripheral lines.
- Anyway to really do borders right we need a different API that respects the fact that there's one line per junction between cells, not one line per cell. I'm not sure what that would look like, but it would for example allow horizontal-only or vertical-only lines, or outlining a group of cells, neither of which we can do currently. Though THIS gets ambiguous with column reordering...
- I almost want to remove per-cell lines entirely, and only support per-cell fill colors, which don't have ANY of these issues. We could then still add support for horizontal-only or vertical-only lines, and could support semiopaque borders without any of this weird overdrawing by only expanding the clippath along the outer edges.

cc @etpinard - if you think it's worthwhile I can move the discussion ^^ into a separate issue and gather input from table users like Dash and editor folks...


etpinard:
... but yes input from users would be nice.

etpinard:
Nice fixes! :dancer:

PR #2513, opened 2018-03-30
fixes #2512 - skip over BADNUM lonlat in scattermapbox select

### LINKED ISSUE #2512: Issue selecting points in scattermapbox
Brushing points in [this codepen](https://codepen.io/cpsievert/pen/Brrqqv) yields the following console errors (and points are not highlighted):

screen shot 2018-03-30 at 2 28 49 pm

etpinard:
cc @alexcjohnson @cpsievert

### DIFF
--- a/src/traces/scattermapbox/select.js
+++ b/src/traces/scattermapbox/select.js
@@ -10,6 +10,7 @@
 
 var Lib = require('../../lib');
 var subtypes = require('../scatter/subtypes');
+var BADNUM = require('../../constants/numerical').BADNUM;
 
 module.exports = function selectPoints(searchInfo, polygon) {
     var cd = searchInfo.cd;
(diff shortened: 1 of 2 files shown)


alexcjohnson:
What if you have `lon` but not `lat` for a point, or vice versa? Is that case handled elsewhere?

etpinard:
> Is that case handled elsewhere?

Yes, early on in `calc` -> 

https://github.com/plotly/plotly.js/blob/master/src/traces/scattergeo/calc.js#L36-L37

and note that `scattermapbox` reuses the `scattergeo` calc method:

https://github.com/plotly/plotly.js/blob/master/src/traces/scattermapbox/index.js#L17

alexcjohnson:
Excellent. 💃

PR #2516, opened 2018-04-02
fix plotinfo when mainplot is no longer linked

### LINKED ISSUE #2509: Plotly.react sequence and subplot bgcolor
The following Pen shows a simple pair of calls to `Plotly.react` which causes the `plot_bgcolor` to be applied to the gridlines of one of the subplots instead of its background: https://codepen.io/nicolaskruchten/pen/QmQXMK

alexcjohnson:
Fixes #2509 - and probably various other bugs resulting from the fact that `mainplot` wasn't cleared when a previously overlaid subplot became independent.

cc @etpinard

### DIFF
--- a/src/plots/cartesian/index.js
+++ b/src/plots/cartesian/index.js
@@ -372,6 +372,7 @@ function makeSubplotData(gd) {
             overlays.push(k);
         } else {
             subplotData.push(k);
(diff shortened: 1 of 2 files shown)


etpinard:
looks good :dancer:

Issue #2510, opened 2018-03-29
Turn finance charts into first-class trace types


nicolaskruchten:
The following Pen shows an issue with a sequence of `Plotly.react` calls not meshing well with candlestick charts: https://codepen.io/nicolaskruchten/pen/qooWqr

alexcjohnson:
Update: This is an issue with `Plotly.react` but NOT with `Plotly.restyle`. I spent a couple of hours fighting with it before concluding (after talking with @etpinard) that my time would be better spent turning `candlestick` and `OHLC` into first-class chart types instead of transform hacks. I'm not even going to *look* at #2508 until this is done, because the OHLC transform hacks are so different from our other transforms that the framework will be far easier to reason about and fix once they're gone. Anyway this is a change we've been discussing for ages and will allow us to fix several more otherwise-intractable bugs.

While investigating this I found a different class of errors, that may or may not impact plotly.js directly but will definitely impact the editor and perhaps other apps: traces mutating the "public" parts of `_full(Data|Layout)` outside of the `supplyDefaults` framework: either:
- adding new properties that look like attributes (because they aren't prefaced with `_`) but are not
- reshaping an existing attribute, as when we reshape 1D Z arrays into 2D
- filling in auto values only during `calc` and NOT writing them back to `data` and `layout` - this means if you make a small change to the plot that doesn't trigger `calc`, these quantities will disappear from `_full*` and will not be available to the editor.

I've fixed many of these, and explicitly ignored the others in the relevant test ([this one](https://github.com/plotly/plotly.js/blob/3d65de3adb20ca0c75e9ecb67739c5d801cc3b24/test/jasmine/tests/plot_api_test.js#L2837), where we draw and then redraw a plot with `Plotly.react` and verify that nothing was done; now I'm also testing that `fullData` and `fullLayout` are unchanged in their supposedly public portions), so that the problem is at least bounded.

alexcjohnson:
Incidentally, I came across this page that has a whole bunch of variants on candlestick charts, that we will suddenly be free to add as new features once these are their own trace type http://stockcharts.com/school/doku.php?id=chart_school:chart_analysis:arms_candlevolume
In particular, check out this one that uses color (black/red) to denote whether the closing price is higher or lower than the previous day, and filling (closed/open) to denote whether the price increased or decreased during the day
[...]

Issue #2508, opened 2018-03-29
Plotly.react sequence and transforms


nicolaskruchten:
The sequence of calls to `Plotly.react` laid out in this pen https://codepen.io/nicolaskruchten/pen/pLamVv?editors=1010 causes display problems in the last step. The buttons allow you to step through the sequence using a mix of `newPlot` and `react` to taste.

nicolaskruchten:
Related to #2470?

alexcjohnson:
A new rule I'm considering implementing & enforcing, to help ensure transforms work correctly with all trace types:
- *Every* trace must define `_length`, and MUST do it during `supplyDefaults` so it's available before any transforms operate.
- This is (at least for now) only appropriate for traces that have 1D data - so it applies for example to heatmaps if you provide triplet data (`x`/`y`/`z` are all 1D) but not if `z` is 2D.
- When it does not apply, set `_length = null`, and all transforms that operate on the assumption of 1D data (which is currently all of them) will be automatically disabled.

@etpinard thoughts?

etpinard:
I like it. It sounds like a nice way to :lock: the current behavior. If ever we start supporting transforms on 2D arrays, we can reassess.

alexcjohnson:
> If ever we start supporting transforms on 2D arrays, we can reassess.

Yes - just looking at `table` traces... it would be sweet to be able to filter/sort/aggregate a `table`, but this has (at least) two problems right now:
- `cells.values` is a 2D array... our transforms wouldn't know what to do with this, and generally you'd want the transform target to be one column of that, but we don't have a mechanism to pick out a column.
- `header.values` is also a 2D array, so it'll show up in `_arrayAttrs` and if we try to continue using automatic array collection we'll have to find a way to exclude headers.

Anyway trying to generalize this to other 2D trace types seems tough, given all the different combinations of 1D & 2D data available within a single trace type, with some of the 1D arrays matching columns of the 2D, some matching rows, some allowing 1D or 2D, etc etc... So either each transform would have to be specifically extended for each 2D trace type it wants to support or we'd need a way to describe (in the schema?) which dimensions of which attributes relate to each other.

So with `_length = null` we can disable this explicitly, then re-enable it type-by-type as we figure it out.

etpinard:
> So with _length = null we can disable this explicitly, then re-enable it type-by-type as we figure it out.

Great :ok_hand:

PR #2525, opened 2018-04-04
React finance precursor

alexcjohnson:
Various fixes to our `supplyDefaults` framework to make `Plotly.react` happier. Since this is no longer part of fixing finance charts with `react` per se (#2510 is going a different route now), making a separate PR for these here.

cc @etpinard

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2429,8 +2429,14 @@ function diffData(gd, oldFullData, newFullData, immutable) {
         gd: gd
     };
 
+
+    var seenUIDs = {};
+
     for(i = 0; i < oldFullData.length; i++) {
-        trace = newFullData[i];
+        trace = newFullData[i]._fullInput;
+        if(seenUIDs[trace.uid]) continue;
+        seenUIDs[trace.uid] = 1;
+
         diffOpts.autoranged = trace.xaxis ? (
             Axes.getFromId(gd, trace.xaxis).autorange ||
             Axes.getFromId(gd, trace.yaxis).autorange
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -380,7 +380,7 @@ plots.supplyDefaults = function(gd) {
         if(_module.cleanData) _module.cleanData(newFullData);
     }
 
-    if(oldFullData.length === newData.length) {
+    if(oldFullData.length === newFullData.length) {
         for(i = 0; i < newFullData.length; i++) {
             relinkPrivateKeys(newFullData[i], oldFullData[i]);
         }
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2364,14 +2364,15 @@ plots.doCalcdata = function(gd, traces) {
     // clear stuff that should recomputed in 'regular' loop
     if(hasCalcTransform) clearAxesCalc(axList);
 
-    // 'regular' loop
-    for(i = 0; i < fullData.length; i++) {
+    function calci(i, isContainer) {
+        trace = fullData[i];
+        _module = trace._module;
+
+        if(!!_module.isContainer !== isContainer) return;
+
         var cd = [];
 
-        trace = fullData[i];
-
         if(trace.visible === true) {
-            _module = trace._module;
 
             // keep ref of index-to-points map object of the *last* enabled transform,
             // this index-to-points map object is required to determine the calcdata indices
@@ -2405,6 +2406,11 @@ plots.doCalcdata = function(gd, traces) {
 
         calcdata[i] = cd;
     }
+
+    // 'regular' loop - make sure container traces (eg carpet) calc before
+    // contained traces (eg contourcarpet)
+    for(i = 0; i < fullData.length; i++) calci(i, true);
+    for(i = 0; i < fullData.length; i++) calci(i, false);
 
     Registry.getComponentMethod('fx', 'calc')(gd);
 };
--- a/src/traces/carpet/calc.js
+++ b/src/traces/carpet/calc.js
@@ -52,8 +52,8 @@ module.exports = function calc(gd, trace) {
     trace.setScale();
 
     // Convert cartesian-space x/y coordinates to screen space pixel coordinates:
-    t.xp = trace.xp = map2dArray(trace.xp, x, xa.c2p);
-    t.yp = trace.yp = map2dArray(trace.yp, y, ya.c2p);
+    t.xp = trace._xp = map2dArray(trace._xp, x, xa.c2p);
+    t.yp = trace._yp = map2dArray(trace._yp, y, ya.c2p);
 
     // This is a rather expensive scan. Nothing guarantees monotonicity,
     // so we need to scan through all data to get proper ranges:
@@ -88,7 +88,7 @@ module.exports = function calc(gd, trace) {
 
     // Tabulate points for the four segments that bound the axes so that we can
     // map to pixel coordinates in the plot function and create a clip rect:
-    t.clipsegments = calcClipPath(trace.xctrl, trace.yctrl, aax, bax);
+    t.clipsegments = calcClipPath(trace._xctrl, trace._yctrl, aax, bax);
 
     t.x = x;
     t.y = y;
--- a/src/traces/carpet/calc.js
+++ b/src/traces/carpet/calc.js
@@ -17,25 +17,37 @@ var calcLabels = require('./calc_labels');
 var calcClipPath = require('./calc_clippath');
 var clean2dArray = require('../heatmap/clean_2d_array');
 var smoothFill2dArray = require('./smooth_fill_2d_array');
+var hasColumns = require('./has_columns');
+var convertColumnData = require('../heatmap/convert_column_xyz');
+var setConvert = require('./set_convert');
 
 module.exports = function calc(gd, trace) {
-    var xa = Axes.getFromId(gd, trace.xaxis || 'x');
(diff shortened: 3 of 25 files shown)


test/jasmine/tests/plot_api_test.js
```diff
@@ -2855,11 +2855,64 @@ describe('Test plot api', function() {
...
+                                delete ax.range;
+                            }
+                        }
+                    });
+                    out.data.forEach(function(trace) {
+                        if(trace.type === 'contourcarpet') {
+                            delete trace.zmin;
+                            delete trace.zmax;
+                        }
+                    });
```
alexcjohnson: Having collected all the patches we currently need to get `fullJson` to match up from run to run, we can then go about cleaning them up in subsequent PRs. In the meantime @nicolaskruchten @VeraZab the editor may need to be aware that these attributes are inconsistently available in the meantime 🙈 

Note that I used `Plotly.Plots.graphJson` to generate `fullJson`, which already removes underscored attributes and functions.

src/plots/plots.js
```diff
@@ -2406,6 +2407,11 @@ plots.doCalcdata = function(gd, traces) {
         calcdata[i] = cd;
     }
 
+    // 'regular' loop - make sure container traces (eg carpet) calc before
+    // contained traces (eg contourcarpet)
+    for(i = 0; i < fullData.length; i++) calci(i, true);
+    for(i = 0; i < fullData.length; i++) calci(i, false);
```
alexcjohnson: Because I moved `convertColumnData` and `setConvert` for `carpet` traces from the `supplyDefaults` step to the `calc` step, we need to ensure `carpet.calc` happens before `contourcarpet.calc` and `scattercarpet.calc`. I'm not sure if we're ever going to have another situation like this where one trace essentially lives inside another, but in case we do this seemed like a reasonably generic way to handle it.

src/plots/plots.js
```diff
@@ -2364,14 +2364,15 @@ plots.doCalcdata = function(gd, traces) {
...
 
-    // 'regular' loop
-    for(i = 0; i < fullData.length; i++) {
-        var cd = [];
-
+    function calci(i, isContainer) {
         trace = fullData[i];
+        _module = trace._module;
+
+        if(!!_module.isContainer !== isContainer) return;
```
etpinard: Ha. Nice fix. I wonder if other carpet-dependent blocks (e.g. [here](https://github.com/plotly/plotly.js/blob/3d65de3adb20ca0c75e9ecb67739c5d801cc3b24/src/plots/plots.js#L948-L963)) could be simplified (or at least generatlized) using `_module.isContainer`?

test/jasmine/tests/plot_api_test.js
etpinard: Brilliant test :sparkles:

src/traces/ohlc/ohlc_defaults.js
```diff
@@ -24,17 +22,13 @@ module.exports = function handleOHLC(traceIn, traceOut, coerce, layout) {
...
-        len = Math.min(len, x.length);
-        if(len < x.length) traceOut.x = x.slice(0, len);
-    }
+    if(x) len = Math.min(len, x.length);
 
-    if(len < open.length) traceOut.open = open.slice(0, len);
-    if(len < high.length) traceOut.high = high.slice(0, len);
-    if(len < low.length) traceOut.low = low.slice(0, len);
-    if(len < close.length) traceOut.close = close.slice(0, len);
+    traceOut._inputLength = len;
```
etpinard: Any reason for naming this `_inputLength`? Parcoords and splom use `_commonLength` to stash their common `dimensions[i].values` length - which I guess is a little different than the finance trace case here and the regular `traceOut._length` used in (all?) other traces.

src/traces/carpet/calc.js
```diff
@@ -52,8 +52,8 @@ module.exports = function calc(gd, trace) {
     trace.setScale();
 
     // Convert cartesian-space x/y coordinates to screen space pixel coordinates:
-    t.xp = trace.xp = map2dArray(trace.xp, x, xa.c2p);
-    t.yp = trace.yp = map2dArray(trace.yp, y, ya.c2p);
+    t.xp = trace._xp = map2dArray(trace._xp, x, xa.c2p);
+    t.yp = trace._yp = map2dArray(trace._yp, y, ya.c2p);
```
etpinard: Are these `xp` and `yp` values used anywhere? [Here](https://github.com/plotly/plotly.js/blob/3d65de3adb20ca0c75e9ecb67739c5d801cc3b24/src/traces/carpet/plot.js#L81-L82) they appear to be reset during the plot step.

etpinard: Are these xp and yp values used anywhere? Here they appear to be reset during the plot step.

src/plots/plots.js
```diff
@@ -380,7 +380,7 @@ plots.supplyDefaults = function(gd) {
         if(_module.cleanData) _module.cleanData(newFullData);
     }
 
-    if(oldFullData.length === newData.length) {
+    if(oldFullData.length === newFullData.length) {
```
etpinard: How hard would it be to :lock: this fix?

src/plot_api/plot_api.js
```diff
@@ -2429,8 +2429,14 @@ function diffData(gd, oldFullData, newFullData, immutable) {
         gd: gd
     };
 
+
+    var seenUIDs = {};
+
     for(i = 0; i < oldFullData.length; i++) {
-        trace = newFullData[i];
+        trace = newFullData[i]._fullInput;
+        if(seenUIDs[trace.uid]) continue;
```
etpinard: similarly here, can we :lock: this thing?

src/traces/ohlc/ohlc_defaults.js
alexcjohnson: This will be converted to [`_length`](https://github.com/plotly/plotly.js/blob/3d65de3adb20ca0c75e9ecb67739c5d801cc3b24/src/plots/cartesian/set_convert.js#L408) when I de-transform these traces, but I didn't want to use that (or `_commonLength`) here since the post-transform traces have different lengths.

src/plots/plots.js
alexcjohnson: Probably some such blocks could... though I don't see much to do with that `carpetDependents` loop. It might be nice to move it out of plots/plots and into the `carpet` module, but probably not in this PR.

src/traces/carpet/calc.js
alexcjohnson: ha good call. Nothing fails if you 🔪 these lines, which is good because the `calc` step had *better* not generate pixel positions, the axis scaling isn't known yet at that point!
Removed in https://github.com/plotly/plotly.js/pull/2525/commits/0221510d0e1443aec95f7830659cfdffcd6e6772

src/traces/carpet/calc.js
```diff
@@ -65,10 +65,6 @@ module.exports = function calc(gd, trace) {
     // create conversion functions that depend on the data
     trace.setScale();
 
-    // Convert cartesian-space x/y coordinates to screen space pixel coordinates:
-    t.xp = trace._xp = map2dArray(trace._xp, x, xa.c2p);
-    t.yp = trace._yp = map2dArray(trace._yp, y, ya.c2p);
```
etpinard: :tada:

src/traces/carpet/calc.js
alexcjohnson: wow, then `map_2d_array` became totally unused -> https://github.com/plotly/plotly.js/pull/2525/commits/b0af416a3561da25af1fa726da99be30de1b6c68

src/plots/plots.js
alexcjohnson: I was going to add a `Plotly.react` test in `transform_groupby_test` for this and [diffing on `_fullInput`](https://github.com/plotly/plotly.js/pull/2525#discussion_r179326587), hence https://github.com/plotly/plotly.js/pull/2525/commits/c2b11dc5a8665c8c1e86b7f7fc969b15184e86f2, since that's the obvious place where fullData will have a different length from data after finance is converted. But on second thought perhaps that can of worms should wait for #2508? I'll make a note in that issue if you 👍. The fact that all the existing `Plotly.react` and other tests pass means these didn't cause *problems* at least.

test/jasmine/tests/plot_api_test.js
VeraZab: @alexcjohnson only zmin and zmax?

alexcjohnson: @VeraZab `zmin` and `zmax` are actually not a problem for you, I don't think, as they only apply to `contourcarpet` where they're not even attributes (though they really should be... and if we make them into bon-a-fide attributes this error will go away).

It's the ones farther up that might cause headaches: `tick0`, `dtick`, and (for 3D and maybe occasionally for 2D) `range`.  I *think* this is only when some other attribute indicates these attributes are auto-determined, so it still may not be an issue as you probably don't even want to display them in these cases, but I'm not sure about that.

The way to test would be to arrange a situation where you DO see those attributes, then make some totally unrelated minor change (tweak a color or something in a different panel) and go back and see if they're still where they're supposed to be.

Or just ignore it for now 😎 but if you happen to come across a bug that looks like ^^ yell at me to fix this.

src/plots/plots.js
etpinard: > But on second thought perhaps that can of worms should wait for #2508?

sounds good :ok_hand:

etpinard:
Great PR :dancer:

Issue #2174, opened 2017-11-20
Needs discussion: Registering bespoke trace types


jackparmer:
@dfcreative had this interesting idea from listening to some plotly.js workshop participants today and I thought I'd record it here. Dima, let me know if I got it totally wrong, am missing key parts, etc.

It would be nice if there were a way amateur developers could create and register their own plotly.js traces.

I'm imagining something like the Plotly.py's FigureFactory. Customized traces could be composed using D3 and registered as a trace type at runtime (see pseudocode below).

Right now, to create a new trace type on a Plotly.js coordinates system, the options are  labor intensive or at least daunting:
- Submit a PR complete with tests, mocks, etc
- Fork plotly.js and hack it in
- Make a D3 or WebGL chart _entirely from scratch_, writing your own axes, annotations, zoom/pan behavior, event emitters, and hover labels.

If developers could more easily add their own D3 creations to the Plotly.js coordinate system, such that they were treated like a trace, we'd have a lot more flexibility in helping users prototype ideas and patches without the herculean upfront effort of cementing them monolithically into plotly.js.

For example, say I wanted to add something like this phylogeny tree to my Plotly.js dashboard:
https://bl.ocks.org/mbostock/c034d66572fd6bd6815a

I'm imaging something like a `Plotly.register` routine that adds `phylogeny` as a trace type at runtime:

```
Plotly.register({
    factory: phyFactory, // function that returns some D3 object
    coordinates: 'polar',
    type: 'phylogeny',
    xaxis: 'xaxis1',
    yaxis: 'yaxis1,'
    // A bunnnnch of other stuff...
})
```

Then elsewhere in your JS code, you could add this pseudo-trace to a Plotly DIV like normal, but with a custom `data` object that gets passed onto the `factory` function provided in `Plotly.register`. Something like:

```
Ploty.plot( [{
    type: 'phylogeny',
    data: newickJsonString,
    innerRadius: ... // etc
}] )
```

Ideally, as long as the D3 developer followed some guidelines, they would get a ton of boilerplate for free like a superb axes system, hover labels, zoom/pan configurability, and event emitters.

***

[...]

PR #2527, opened 2018-04-05
Splom zoom perf

etpinard:
to be merged into https://github.com/plotly/plotly.js/pull/2505

------------

This PR features performance improvements on zoom interactions geared towards splom traces, but will benefit all cartesian subplots especially on multi-subplot graphs.

I opened a separate PR because I'm hoping to improve a few more things before the splom PR is ready to be merged (cc @dy )

### Things improved so far

- `relayout(gd, 'xaxis.range', [])` is now using a special (much) faster relayout pathway
- `relayout(gd, 'dragmode', '')` now only has to update one DOM node per graph!
- subplot updates during panning are now more granular (and hence faster for splom-only, scattergl-only cases and graphs w/o points that need to be scaled on pan)

### Things I'll like to improve in the coming days

- double-click (i.e auto/reset range) performance. This thing currently goes through `editType: 'calc'` which is slow as a :turtle:. I'll try to cache the autorange value somewhere (and clearing it when new data comes in) and use the `axrange` pathway to speed thing up. This will benefit big data gl and svg graphs too!
- speed up defaults (`Plots.supplyDefaults` takes > 100ms at 50x50 subplots), Getting rid of `indexOf` into axis lists would be a good start.
- cache a few (sometimes) large selections during the plot step to :hocho: `selectAll` on pan
- speed up `axrange` pathway even more by targeting minimal set of axes to update (via `Axes.doTicks`) instead of redraw ticks and labels for all axes.

cc @alexcjohnson


src/plots/plots.js
```diff
@@ -627,24 +627,22 @@ plots.createTransitionData = function(gd) {
...
+    // check trace modules
     var modules = this._modules || [];
-
     for(i = 0; i < modules.length; i++) {
-        var modulei = modules[i];
-        if(modulei.categories && modulei.categories.indexOf(category) >= 0) {
-            return true;
-        }
+        var name = modules[i].name;
+        if(name === category) return true;
```
etpinard: so calls like `fullLayout._has('scattergl')` just works w/o having to set a `scattergl` category in the trace module.

src/plots/cartesian/graph_interact.js
```diff
@@ -139,17 +137,29 @@ module.exports = function initInteractions(gd) {
...
+};
+
+// Minimal set of update needed on 'modebar' edits.
+// We only need to update the <g .draglayer> cursor style.
+//
+// Note that changing the axis configuration and/or the fixedrange attribute
+// should trigger a full initInteractions.
+exports.updateFx = function(fullLayout) {
+    var cursor = fullLayout.dragmode === 'pan' ? 'move' : 'crosshair';
+    setCursor(fullLayout._draggers, cursor);
```
alexcjohnson: Ah, great solution - much simpler than what I was thinking, and at least as fast 🐎

src/traces/scatter/index.js
```diff
@@ -34,7 +34,7 @@ Scatter.animatable = true;
 Scatter.moduleType = 'trace';
 Scatter.name = 'scatter';
 Scatter.basePlotModule = require('../../plots/cartesian');
-Scatter.categories = ['cartesian', 'symbols', 'markerColorscale', 'errorBarsOK', 'showLegend', 'scatter-like'];
+Scatter.categories = ['cartesian', 'svg', 'symbols', 'markerColorscale', 'errorBarsOK', 'showLegend', 'scatter-like', 'draggedPts'];
```
etpinard: **IMPORTANT** (and possibly worth refactoring)

scattergl and splom traces are currently declared with `'cartesian'` category so that cartesian axis logic that they share just works. But this means that not all 'cartesian' traces are _svg_ or more precisely not all 'cartesian' traces are plotted in `<g .cartersianlayer>`. So to avoid looping over all subplots on drag to shift `plotinfo.plot` and its clip `<rect>` for scattergl and splom traces, I made an `'svg'` category. This is a slight misnomer though, as heatmaps and histogram2d are here considered _svg_ even though they are built using a `<canvas>`. Oh well. The 'svg' category mirrors the 'gl' and 'regl' category listed in scattergl and splom trace modules. Any objections?

Moreover, I added a 'draggedPts' category to avoid those costly `selectAll('.point')` on pan when we don't need them. For example, graphs with contour lines feel faster on pan as we no longer traverse the `<g .cartesianlayer>` looking for points to scaled.

src/plots/cartesian/axes.js
```diff
@@ -2188,7 +2188,7 @@ axes.doTicks = function(gd, axid, skipTitle) {
             }
             drawTicks(mainPlotinfo[axLetter + 'axislayer'], tickpath);
 
-            tickSubplots = Object.keys(ax._linepositions);
+            tickSubplots = Object.keys(ax._linepositions || {});
```
etpinard: Unfortunate, but important as `relinkPrivateKeys` does relink empty objects.

etpinard: ... does **not** relink

src/plots/plots.js
```diff
@@ -627,24 +627,22 @@ plots.createTransitionData = function(gd) {
...
+        // N.B. this is modules[i] along with 'categories' as a hash object
+        var _module = Registry.modules[name];
```
alexcjohnson: TBH I've forgotten why we wrapped modules like this in the registry... not a big deal, but would it be better to attach the hash directly to the module, or even make `categories` as a hash from the beginning? Is it ever important that it's a list?

src/traces/scatter/index.js
alexcjohnson: I think this is fine - `'cartesian'` means "data relates to cartesian axes" and `'svg'` means "data is drawn into an SVG container" (heatmaps use a canvas to generate the image but then put it into an svg `<image>` element)

`'draggedPts'` is still potentially overkill, as the trace type doesn't necessarily tell you if the traces contain these elements - all of these trace types can be drawn with no points or text, only lines and fills...

src/plot_api/plot_api.js
```diff
@@ -1772,6 +1683,26 @@ exports.relayout = function relayout(gd, astr, val) {
...
+
+        if(flags.axrange) {
+            // N.B. leave as sequence of subroutines (for now) instead of
+            // subroutine of its own so that finalDraw always gets
+            // executed after drawData
+            seq.push(
+                // TODO
+                // no test fail when commenting out doAutoRangeAndConstraints,
+                // but I think we do need this (maybe just the enforce part?)
+                // Am I right?
```
alexcjohnson: Huh, I'd have thought so... perhaps worth spending a little time looking for something that does fail if this is removed, so it can be 🔒 down.

alexcjohnson:
> double-click (i.e auto/reset range) performance. This thing currently goes through `editType: 'calc'` which is slow as a :turtle:

`Axes.expand` (which happens during `calc` does nothing if the axis isn't autoranged - which can speed up initial drawing but means if you turn on `autorange` later you don't know whether `ax._min/_max` are populated. We could either calculate min/max no matter what during `calc`, or at least stash the arrays & options used to generate them so that when `autorange` is enabled we can pull it back out and proceed with the calculation without needing to go all the way through `calc` again.

There would also be some advantages to keeping separate track of each trace / object that called `Axes.expand` - ie keep a separate `_min/_max` for each trace/shape/annotation etc. and only combine them when finally calculating the range. That would avoid `calc` for visibility changes, for example (you'd just have to know to include or not include that object's `_min/_max`) and for moving annotations/shapes, and down the road would allow `calc` per trace.

src/plots/cartesian/axes.js
etpinard: does **not** relink empty objects.

src/plot_api/plot_api.js
etpinard: @alexcjohnson Tests are currently passing even when `subroutines.doAutoRangeAndConstraints` is commented out, because of this block:

https://github.com/plotly/plotly.js/blob/2a2582022105abd640ef3e5a38d98fb343a06779/src/plot_api/plot_api.js#L2088-L2107

which makes axis range relayout calls altering constrained axes go though the `calc: true` edit pathway which (of course) calls `doAutoRangeAndConstraints` itself.

-------------

Interestingly, commenting out that [block](https://github.com/plotly/plotly.js/blob/2a2582022105abd640ef3e5a38d98fb343a06779/src/plot_api/plot_api.js#L2088-L2107) which sets `flags.calc = true` does not make any test fails when `subroutines.doAutoRangeAndConstraints` is called during `axrange` edits. Perhaps we don't need?

src/plots/plots.js
etpinard: Referencing https://github.com/plotly/plotly.js/issues/2174 where we like to make building custom trace modules more user friendly:

- we should try to **not** mutate the input module object, so that means not replacing the `categories` list with a hash object nor augmenting the input module object with e.g. a `categoryHash` key.
- listing categories as an array feels more natural than `{cat1: 1, cat2: 1}` hash object

So, I'm voting for the status quo.

src/plot_api/plot_api.js
alexcjohnson: It doesn't look to me as though I wrote all that great test coverage for this section... the two key points of it are:
- You can `relayout` a constrained axis, particularly to a smaller range, and have the other axes in the group update (expanding *or* shrinking) to continue to meet the constraints (note that this won't work with `react`, there you'd have to manually update *all* axes in the constraint group)
- If you `relayout` (or `react`) two or more axes in a group in such a way as to explicitly violate the constraints, one or more of *those* axes will expand (never shrink) its range to fit the constraints.

So if you want to clean this up we should first add a few tests explicitly for ^^. I see some [gui tests](https://github.com/plotly/plotly.js/pull/1522/files#diff-200dbdf53958cf8973413a709518f4bbR389) and one [restyle test](https://github.com/plotly/plotly.js/pull/2465/files#diff-5a140ce1cc87a420496039bbf8d699f1R1265) that uses constraints, but unfortunately no `relayout` or `react` tests of these edge cases.

src/components/shapes/draw.js
```diff
@@ -42,7 +42,11 @@ function draw(gd) {
     // Remove previous shapes before drawing new in shapes in fullLayout.shapes
     fullLayout._shapeUpperLayer.selectAll('path').remove();
     fullLayout._shapeLowerLayer.selectAll('path').remove();
-    fullLayout._shapeSubplotLayers.selectAll('path').remove();
+
+    for(var k in fullLayout._plots) {
+        var shapelayer = fullLayout._plots[k].shapelayer;
+        if(shapelayer) shapelayer.selectAll('path').remove();
```
alexcjohnson: Interesting - why is this better? Wasn't `fullLayout._shapeSubplotLayers` essentially just an array of all the `shapelayer` nodes? Or is the win actually below in [subroutines.js](https://github.com/plotly/plotly.js/pull/2527/commits/d3fe40dab5708bf94f83b472986726e8f8ff65a9#diff-07cd4d425af26971d4a589aaac8ea059L534) that you don't have to do a `selectAll` to find these?

src/plots/cartesian/dragbox.js
```diff
@@ -1105,7 +1105,7 @@ function calcLinks(gd, xaHash, yaHash) {
     var yaHashLinked = {};
     var yaxesLinked = [];
     for(yLinkID in yLinks) {
-        var ya = getFromId(gd, xLinkID);
+        var ya = getFromId(gd, yLinkID);
```
alexcjohnson: yikes, good catch!

src/components/shapes/draw.js
etpinard: >  Or is the win actually below in subroutines.js that you don't have to do a selectAll to find these?

That. `selectAll` are bad.

src/plot_api/plot_api.js
etpinard: Ok if I open up an issue about this and defer this to a later PR?

alexcjohnson: > Ok if I open up an issue about this and defer this to a later PR?

Yep for sure - that was the "if you want to clean this up" clause

etpinard: Moved to -> https://github.com/plotly/plotly.js/issues/2540

alexcjohnson:
@etpinard really impressive - and counterintuitive - work here!
💃 💃 🐎 🚀

Issue #2193, opened 2017-12-05
Fixed size circle shape


chriddyp:
I'm helping out a Dash community user who could use interactive shape circles (https://community.plot.ly/t/moving-the-location-of-a-graph-point-interactively/7161/1)

I'd like a shape circle that has a fixed size in points but is positioned relative to data - something like a marker point that I can drag around but not necessarily resize.

Right now, circle shapes change size when you drag them around:
![circle shape](https://user-images.githubusercontent.com/1280389/33621771-ebaad17a-d9b9-11e7-9cfb-8dd59a1b9e70.gif)

This is somewhat related to https://github.com/plotly/plotly.js/issues/2038

alexcjohnson:
Cool idea, and there are a lot of variants of this that could be useful. Even for (fully or partially) paper-referenced shapes, it might be nice to be able to size them in px, like if you want to draw your own arrow pointing to a particular axis value, or you have a logo or something that you want a fixed size even if the plot resizes.

Seems to me the most flexible way to do this would be adding attributes like:
```js
xmode: 'scaled' | 'pixel' // scaled (dflt) -> x0/x1 are data vals or plot units, anchorx is ignored
                          // pixel -> x0/x1 are pixels relative to anchorx
anchorx: (any) // data val or plot fraction, depending on `xref`
ymode, anchory: same for y
```

Not sure about the naming, but that would leave the flexibility to do things like make a shape off to the side of a data point, like `(x|y)shift` for annotations, not just centered.

rmoestl:
I think it is worth noting that the API suggested by @alexcjohnson will support **two different uses cases**:
1. **Scenario A:** to render a fixed size shape positioned relative to data (blue rectangle below) as suggested by @chriddyp. E.g. useful to annotate a specific data point.
2. **Scenario B:** to render fixed size shapes relative to the plotting area (red rectangle). E.g. useful to place an icon of some sort at the top right corner of a plot.

Despite it increases the API surface, I can clearly see the benefit of having an **anchor**. When annotating a specific data point you certainly would want the control to render it 3px left the data point, for example. 

Regarding the **naming**, I find `anchorx` quite good. Might `xsizing` be more clear than `xmode`? But I also saw "mode" as parts of an attribute name already. So `xmode` might be more consistent within plotly.js...

One **question**, though: Should we take the dimensions of a fixed size shape into account when calculating the **range of an axis**? 
[...]

PR #2532, opened 2018-04-06
Fixed size shapes

rmoestl:
This implements *fixed size shapes* specified in #2193. Please refer to the commit message for a rough overview of the feature.


src/components/shapes/attributes.js
```diff
@@ -71,13 +75,41 @@ module.exports = {
             'the date to unix time in milliseconds.'
         ].join(' ')
     }),
+    xsizemode: {
+        valType: 'enumerated',
+        values: ['data', 'pixel'],
+        dflt: 'data',
```
alexcjohnson: Ah, I hadn't thought of this before, but `'data'` mode applies not just to axis scaling but also plot fraction scaling, right? ie when `xref='paper'` this option is still relevant, right? So how about we change `'data'` to `'scaled'`? That seems like it encompasses both cases and differentiates them from `'pixel'` sizing...

rmoestl: Good point. I'll change that.

rmoestl:
The new test image is here: https://github.com/plotly/plotly.js/blob/fixed-size-shapes/test/image/baselines/fixed_size_shapes.png

test/image/mocks/fixed_size_shapes.json
```diff
@@ -0,0 +1,255 @@
...
+        "line": {
+          "color": "rgba(96, 171, 50, 1)"
+        }
+      }
+    ]
+  },
+  "config": {
+    "editable": "true"
+  }
+}
```
alexcjohnson: Very nice test image, and I think the interactions feel very natural as you have them. Other than adding a trailing newline to make the [syntax test happy](https://circleci.com/gh/plotly/plotly.js/8691?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) the only change I'd like to see is to use the other three axis types (log, date, category) on one axis each - no need to add more subplots, just convert a couple of the existing axes.

alexcjohnson:
One thing I just noticed is the cursors start out wrong (at least in some cases) - the "top" cursor shows for most (or all!) of the shape, then after you cause a redraw (in the gif below I zoom in then double click to autorange again) the cursors look correct.

I haven't looked to see if this is exclusive to pixel-sized shapes, nor if it's new with this PR... can you check it out @rmoestl ?

weird shapes cursor

rmoestl:
I'll check out the cursor position thing.

rmoestl:
Re: drag cursors, unfortunately I was not yet able to reproduce this strange behavior. 

Looking at the code, the bounding box of the shape is obtained in `setupDragElement` and later used in the mouse move event handler to calculate the cursor. Can resizing cause this behavior? I don't think so because a resize triggers drawing the shape which in turn triggers a new call of `setupDragElement`. In addition your recording shows that you've not yet resized the respective shape.

Do you have some hints on how to reproduce the bug? I was testing in Chrome and FF by the way. Though I assume `getBoundingClientRect` is working correct in all major browsers.

alexcjohnson:
Thanks for the test image tweaks - love it!

I figured out what the bug is - it's when you scroll the page after the plot draws (which I was doing naturally with your mock because it's fairly tall, whereas the original `shapes` mock is much shorter - if you shrink your browser window so you need to scroll that one you can see it there too).

It's not new in this PR, but I think it has a simple solution so you might as well address it here. The issue is that `getBoundingClientRect` gets called in the outer scope of `setupDragElement`, and gets used later in `updateDragMode`, at which point the actual bbox is different if you've scrolled. The simple solution is to move `getBoundingClientRect` to the scope where it's used.

rmoestl:
Wow, good catch!!

alexcjohnson:
Great, glad that's all it took! That bug fix may actually help performance too - ironically, since now we're calling `getBoundingClientRect` many more times, but we're *not* calling it during draw, only during mouse moves.

I'm happy with this now, lets go! 💃

Juanlu001:
Thanks for this new feature! As a Python user, I understand that the `offline` mode always uses the latest release from the CDN so this should be ready to use. Have the online docs been update accordingly?

rmoestl:
@Juanlu001 ICYMI, online docs have been updated: https://plot.ly/javascript/reference/#layout-shapes

PR #2533, opened 2018-04-06
Fix containment check for reversed ranges

### LINKED ISSUE #2531: Text missing if axis range is reversed and cliponaxis=false
I have a pull request (https://github.com/darabos/plotly.js/pull/1) for it, but I made it against my own repo now to follow the contribution guidelines.

I committed the "before" baseline to demonstrate the issue. You can see the fix in the second commit's diff:
cliponaxis-before-after

This example is a bit unusual. Who uses upside-down bar charts?! In real life I hit this issue with a horizontal bar chart where I wanted the order of the groups in the chart to match the order in the legend:
horizontal-bar-chart

I set `autorange: 'reverse'` and the texts disappeared:
bars-reversed

If I remove `cliponaxis: false`, the texts are clipped:
bars-clipped

I'm open to suggestions, but my PR is aimed at fixing the version with the missing texts. Thanks!

darabos:
Fixes #2531. Thanks!


etpinard:
Beautiful fix. Thanks very much :tada:

PR #2536, opened 2018-04-06
fix hover bug

dimin:


etpinard:
Thanks for the PR. Would you mind sharing a reproducible example of bug you're trying to fix?

etpinard:
ping @dimin can you help us figuring what's causing this bug?

etpinard:
Closing until someone someone can reproduce this _bug_.

dimin:
Sorry Etienne for such a delay, I finally found the reason for the problem and it was in the third-party library that was messing with D3. Sorry for bugging you so stupidly!

PR #2561, opened 2018-04-16
Finance refactor

### LINKED ISSUE #2510: Turn finance charts into first-class trace types

alexcjohnson:
Closes #2510 - Makes finance traces (OHLC, candlestick) into 🥇 first-class trace types.

Along with this I added selection to both of these trace types. I did *not* however add custom selection styles. I'd like to leave this for a later PR.

cc @etpinard


src/components/rangeslider/draw.js
```diff
@@ -455,7 +455,8 @@ function drawRangePlot(rangeSlider, gd, axisOpts, opts) {
             id: id,
             plotgroup: plotgroup,
             xaxis: xa,
-            yaxis: ya
+            yaxis: ya,
+            isRangePlot: true
```
etpinard: I don't think this range-plot range slider fix got :lock: anywhere. We could add a case in `select_test.js` selects a point in the main plot and asserts that it is really that main-plot plot that gets the selected style.

src/plot_api/plot_schema.js
```diff
@@ -285,8 +287,22 @@ exports.getTraceValObject = function(trace, parts) {
             }
         }
 
-        // finally look in the global attributes
+        // next look in the global attributes
         if(!valObject) valObject = baseAttributes[head];
+
+        // finally check if we have a transform matching the original trace type
+        if(!valObject && transforms) {
```
etpinard: Why do we need this?

src/plots/cartesian/constants.js
```diff
@@ -68,6 +68,7 @@ module.exports = {
         'carpetlayer',
         'violinlayer',
         'boxlayer',
```
etpinard: Can `box` traces and `candlestick` coexist on the same subplot?

src/plots/cartesian/index.js
```diff
@@ -203,7 +203,7 @@ function plotOne(gd, plotinfo, cdSubplot, transitionOpts, makeOnCompleteCallback
     // remaining plot traces should also be able to do this. Once implemented,
     // we won't need this - which should sometimes be a big speedup.
     if(plotinfo.plot) {
-        plotinfo.plot.selectAll('g:not(.scatterlayer)').selectAll('g.trace').remove();
+        plotinfo.plot.selectAll('g:not(.scatterlayer):not(.ohlclayer)').selectAll('g.trace').remove();
```
etpinard: This query is getting ridiculous. 

Can we instead loop over [`constants.traceLayerClasses`](https://github.com/plotly/plotly.js/blob/18e19a86a3e683c1d11f6f4a04eb7f89d1b6108f/src/plots/cartesian/constants.js#L64)?

src/plots/cartesian/index.js
```diff
@@ -224,43 +224,50 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
     var oldModules = oldFullLayout._modules || [],
         newModules = newFullLayout._modules || [];
 
-    var hadScatter, hasScatter, hadGl, hasGl, i, oldPlots, ids, subplotInfo, moduleName;
+    var hadScatter, hasScatter, hadOHLC, hasOHLC, hadGl, hasGl, i, oldPlots, ids, subplotInfo, moduleName;
```
etpinard: This had/has logic is also getting ridiculous. 

What we need to do is find all the _gone_ modules (i.e. modules with `had??=true` and `has??=false`), loop over them, and then do what we need to do the cleanup the DOM. 

Better yet -- as more and more trace type will follow scatter and ohlc in not expecting a `selectAll('g.trace').remove()` before each `_module.plot()` -- we could add a trace module method (maybe called `cleanDOM`) to each trace modules that need to remove things for the DOM (or destroy WebGL objects). 

To note, geo subplots (and maybe others?) solve this problem in a different way by using `Plots.generalUpdatePerTraceModule` which calls `_module.plot` of visible **and**  _gone_ modules and expects the trace module plot methods to clean up the DOM accordingly (e.g. like [here](https://github.com/plotly/plotly.js/blob/18e19a86a3e683c1d11f6f4a04eb7f89d1b6108f/src/traces/choropleth/plot.js#L34) for choropleth traces). At present though, I think adding `cleanDOM` to each trace module would be best, as it would allow us to handle the `scattergl` and range-slider cases more easily.

N.B. the splom PR (ref: https://github.com/plotly/plotly.js/pull/2505) also patches this code here, so maybe addressing this comment could be done in a different PR

src/plots/plots.js
```diff
@@ -366,6 +366,10 @@ plots.supplyDefaults = function(gd) {
     newFullLayout._basePlotModules = [];
     newFullLayout._subplots = emptySubplotLists();
 
+    // for traces to request a default rangeslider on their x axes
+    // eg set `_requestRangeslider.x2 = true` for xaxis2
+    newFullLayout._requestRangeslider = {};
```
etpinard: It would be nice to add a mock with a ohlc or candlestick on a x-axis other than `layout.xaxis` to :lock: this down.

src/traces/candlestick/index.js
```diff
@@ -8,13 +8,11 @@
...
 'use strict';
 
-var Registry = require('../../registry');
-
 module.exports = {
     moduleType: 'trace',
     name: 'candlestick',
     basePlotModule: require('../../plots/cartesian'),
-    categories: ['cartesian', 'showLegend', 'candlestick'],
+    categories: ['cartesian', 'showLegend', 'candlestick', 'boxLayout'],
```
etpinard: So users making graphs with candlestick traces are allowed to set `layout.boxgap` and friends?

src/traces/ohlc/calc.js
```diff
@@ -0,0 +1,163 @@
...
+
+function calc(gd, trace) {
+    var xa = Axes.getFromId(gd, trace.xaxis);
+    var ya = Axes.getFromId(gd, trace.yaxis);
+
+    var tickLen = convertTickWidth(gd, xa, trace);
+    var minDiff = trace._minDiff;
+    delete trace._minDiff;
+    var x = trace._xcalc;
+    delete trace._xcalc;
```
etpinard: Why do we need to delete `_xcalc` and `_minDiff`?

src/traces/ohlc/calc.js
```diff
@@ -0,0 +1,163 @@
...
+
+    var cd = [];
+    for(var i = 0; i < x.length; i++) {
+        var xi = x[i];
+        var oi = o[i];
+        var hi = h[i];
+        var li = l[i];
+        var ci = c[i];
+
+        if(xi !== undefined && oi !== undefined && hi !== undefined && li !== undefined && ci !== undefined) {
```
etpinard: Hmm. you mean `BADNUM` here?

src/traces/ohlc/calc.js
```diff
@@ -69,8 +70,14 @@ function calcCommon(gd, trace, x, ya, ptFunc) {
...
 
         if(xi !== undefined && oi !== undefined && hi !== undefined && li !== undefined && ci !== undefined) {
-            // increasing carries over from previous if ci===oi
-            increasing = (ci === oi) ? increasing : ci > oi;
+            if(ci === oi) {
+                // if open == close, look for a change from the previous close
+                if(cPrev !== null && ci !== cPrev) increasing = ci > cPrev;
+                // else (c === cPrev or cPrev is null) no change
+            }
+            else increasing = ci > oi;
```
etpinard: What does this fix exactly?

src/traces/ohlc/helpers.js
```diff
@@ -1,147 +0,0 @@
...
-// from a clear transforms container. The mutations inflicted are
-// cleared in exports.clearEphemeralTransformOpts.
-exports.pushDummyTransformOpts = function(traceIn, traceOut) {
-    var transformOpts = {
-
-        // give dummy transform the same type as trace
-        type: traceOut.type,
-
-        // track ephemeral transforms in user data
-        _ephemeral: true
```
etpinard: Good to see one of the hackiest things I wrote for plotly :hocho:

src/traces/ohlc/hover.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // don't make .yLabelVal or .text, since we're managing hoverinfo
+    // put it all in .extraText
+    pointData.extraText = textParts.join('<br>');
+
+    // this puts the label at the midpoint of the box, ie
+    // halfway between open and close, not between high and low.
+    // TODO: the spike also links to this point, whereas previously
+    // it linked to close. Is one better?
+    pointData.y0 = pointData.y1 = ya.c2p(di.yc, true);
```
etpinard: I think linking to _this point_ is better :ok_hand:

src/traces/ohlc/hover.js
```diff
@@ -0,0 +1,111 @@
...
+    if(Color.opacity(lc) && container.line.width) pointData.color = lc;
+    else pointData.color = container.fillcolor;
+
+    pointData.x0 = xa.c2p(di.pos + centerShift - displayHalfWidth, true);
+    pointData.x1 = xa.c2p(di.pos + centerShift + displayHalfWidth, true);
+
+    pointData.xLabelVal = di.pos;
+
+    pointData.spikeDistance = dxy(di) * spikePseudoDistance / hoverPseudoDistance;
+    pointData.xSpike = xa.c2p(di.pos, true);
```
etpinard: From `candlestick_double-y-axis`, are we expected spike lines over this point to be black:

peek 2018-04-16 15-31


?

src/traces/ohlc/style.js
```diff
@@ -0,0 +1,35 @@
...
+
+    s.each(function(d) {
+        var trace = d[0].trace;
+
+        d3.select(this).selectAll('path').each(function(di) {
+            var dirLine = trace[di.dir].line;
+            d3.select(this)
+                .style('fill', 'none')
+                .call(Color.stroke, dirLine.color)
+                .call(Drawing.dashLine, dirLine.dash, dirLine.width)
```
etpinard: In the `finance_style` mock, the dash line pattern makes the new split increasing/decreasing legend item look a little strange:

image

compare to e.g. for the `ohlc_first` mock with `layout.showlegend: true`:

image

which looks amazing.

Should we just not call `Drawing.dashLine` here?

test/jasmine/tests/plot_api_test.js
```diff
@@ -2276,6 +2276,90 @@ describe('Test plot api', function() {
...
+                    name: 'Gruyere cooling off'
+                },
+                name: 'Emmenthaler'
+            }];
+
+            Plotly.plot(gd, data);
+
+            // Even if both showlegends are false, leave trace.showlegend out
+            // My rationale for this is that legends are sufficiently different
+            // now that it's worthwhile resetting their existence to default
```
etpinard: :ok_hand:

test/jasmine/tests/plot_api_test.js
```diff
@@ -2781,6 +2865,53 @@ describe('Test plot api', function() {
             .then(done);
         });
 
+        it('can change data in candlesticks multiple times', function(done) {
```
etpinard: Should we added add a finance trace case for the [_can redraw with no changes as a noop_](https://github.com/plotly/plotly.js/blob/7d91d29a48bc882afccda635054dfa1a7ece85fd/test/jasmine/tests/plot_api_test.js#L2856) set of tests?

test/jasmine/tests/select_test.js
```diff
@@ -1299,10 +1299,77 @@ describe('@flaky Test select box and lasso per trace:', function() {
                 null, BOXEVENTS, 'violin select'
             );
         })
-        .catch(fail)
+        .catch(failTest)
         .then(done);
     });
 
+    ['ohlc', 'candlestick'].forEach(function(type) {
```
etpinard: So, apart from selection and the new legend items, can you list all the new features this PR adds to finance traces?

etpinard: .... mostly for changelog purposes :books:

src/components/fx/hover.js
```diff
@@ -844,6 +837,13 @@ function createHoverText(hoverData, opts, gd) {
             text += (text ? '<br>' : '') + d.text;
         }
 
+        // used by other modules (initially just ternary) that
+        // manage their own hoverinfo independent of cleanPoint
+        // the rest of this will still apply, so such modules
+        // can still put things in (x|y|z)Label, text, and name
+        // and hoverinfo will still determine their visibility
+        if(d.extraText !== undefined) text += (text ? '<br>' : '') + d.extraText;
```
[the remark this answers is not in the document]
alexcjohnson: Our previous users of `extraText` use *only* `extraText`, no x/y/z text, so they're essentially independent of the intervening block. But finance traces use `extraText` in place of `yLabel`, and still use `xLabel`. Contrary to the comment (that I just moved down here with `extraText`) previously if you had both parts, the x/y/z labels would obliterate `extraText`.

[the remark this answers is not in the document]
etpinard: Thanks for the headsup :ok_hand:

src/plots/cartesian/index.js
alexcjohnson: > Can we instead loop over `constants.traceLayerClasses`?

Somehow we need to distinguish between trace types that know how to update and those that don't - I feel like rather than adding another category for this, we should just teach the remaining types how to update. Probably actually not that hard to do, right? Just adding the right `exit()` statements and moving things appropriately between `enter()` and `each`?

I wouldn't want to do that in this PR of course...

src/plots/cartesian/index.js
alexcjohnson: Definitely a separate PR - there will be a bit to discuss, between this one and https://github.com/plotly/plotly.js/pull/2561#discussion_r181852562 - created #2563, OK to defer both of these for now?

src/traces/ohlc/calc.js
alexcjohnson: This is code that used to be in [`_isIncreasing`](https://github.com/plotly/plotly.js/pull/2561/commits/84c36a91091aa5524e6b2a9f0788766b622146de#diff-e872430731145abebb1f8f55bf310949L108) that I didn't copy over right the first time, but it affects fallbacks when `open===close`, as tested [here](https://github.com/plotly/plotly.js/pull/2561/files#diff-7deb9e75965ca15196b29a1f7a4a9210R586)

test/jasmine/tests/plot_api_test.js
alexcjohnson: > Should we added add a finance trace case for the can redraw with no changes as a noop set of tests?

It's there -> https://github.com/plotly/plotly.js/blob/7d91d29a48bc882afccda635054dfa1a7ece85fd/test/jasmine/tests/plot_api_test.js#L2826
and the previous `fullJson` exemption was cleared in this PR -> https://github.com/plotly/plotly.js/pull/2561/files#diff-5a140ce1cc87a420496039bbf8d699f1L2913

src/traces/ohlc/style.js
alexcjohnson: True, it looks weird... but it also looks weird in the plot. I think that just says anyone who would really use `line.dash` here would want it to be a very short dash pattern, like `'dot'`. Hardly anyone will use it, but I don't think we should prohibit it, nor should we draw different dashes in the legend vs the plot.

I've changed that mock to `dash: 'dot'` https://github.com/plotly/plotly.js/pull/2561/commits/f84dfae9d46b48fb84fd1be21c710a7dc2d346dd and I have another mock in https://github.com/plotly/plotly.js/pull/2561/commits/0b7541e34d63ce74cddb5d243a652b2df0430b70 that shows it without dashes, so users can see the normal look.

src/plots/cartesian/constants.js
alexcjohnson: good 👀 - they probably could have before this PR, but this morning they couldn't. Now they can https://github.com/plotly/plotly.js/pull/2561/commits/0b7541e34d63ce74cddb5d243a652b2df0430b70

src/plots/plots.js
alexcjohnson: the mock in https://github.com/plotly/plotly.js/pull/2561/commits/0b7541e34d63ce74cddb5d243a652b2df0430b70 🔒 this.

src/traces/candlestick/index.js
alexcjohnson: yes... I included `boxgroupgap` in the mock in https://github.com/plotly/plotly.js/pull/2561/commits/0b7541e34d63ce74cddb5d243a652b2df0430b70 to 🔒 it

src/traces/ohlc/calc.js
alexcjohnson: per https://github.com/plotly/plotly.js/pull/2561/files/82677ac8745ca56da740302977533dffcd5521b6#diff-6877a6db77ade9f43ac4ccd6acbb2713R114:

> when a trace uses this in its calc step it deletes _minDiff, so that next calc this is done again in case the data changed.

I guess in principle we wouldn't need to delete `_xcalc` but it makes me feel better, as we're guaranteed it doesn't get used outside of `calc`. It's only there for 🐎 in the first place, so we don't have to calculate it in both `convertTickWidth` (for all traces) and then again in `calc` (for each trace separately)

src/traces/ohlc/hover.js
alexcjohnson: > are we expecting spike lines over this point to be black

Yeah, that color fails the [`tinycolor.readability` test](https://github.com/plotly/plotly.js/blob/b938842854538d3b6b0bc3ff6eb2cb77964dc899/src/components/fx/hover.js#L1296) so we just pick a color that contrasts with the background.

src/traces/ohlc/calc.js
alexcjohnson: good call, yes `BADNUM` -> https://github.com/plotly/plotly.js/pull/2561/commits/dc016858219587070e633f32b248ae5497e384b4

src/traces/splom/base_plot.js
```diff
@@ -22,7 +22,7 @@ var SPLOM = 'splom';
 function plot(gd) {
     var fullLayout = gd._fullLayout;
     var _module = Registry.getModule(SPLOM);
-    var splomCalcData = getModuleCalcData(gd.calcdata, _module);
+    var splomCalcData = getModuleCalcData(gd.calcdata, _module)[0];
```
alexcjohnson: @etpinard I changed the API for `getModuleCalcData` in https://github.com/plotly/plotly.js/pull/2561/commits/0b7541e34d63ce74cddb5d243a652b2df0430b70 - so that cartesian could avoid double-drawing the same trace when two trace modules share a plotting step (`box` and `candlestick`). In principle we could extend this pattern (a `plot` step removes the `calcdata` items it plotted and returns the items that are left) which could have a small 🐎 benefit when you have many traces of several types, you don't have to search the whole array each time, only the items that haven't already been handled.

src/plots/cartesian/index.js
etpinard: Moved to https://github.com/plotly/plotly.js/issues/2563 

Thanks!

src/plots/cartesian/index.js
etpinard: > OK to defer both of these for now?

Absolutely :ok_hand:

src/traces/ohlc/calc.js
etpinard: Thanks for the :books: 

So in theory we could set `trace._minDiff = null` instead of using `delete` here with no side-effects? Some benchmarks: https://jsperf.com/delete-vs-undefined-vs-null/16

src/plots/get_data.js
```diff
@@ -36,20 +36,40 @@ exports.getSubplotCalcData = function(calcData, type, subplotId) {
...
         var trace = cd[0].trace;
+        if(trace.visible !== true) continue;
 
-        if((trace._module === _module) && (trace.visible === true)) moduleCalcData.push(cd);
+        // we use this to find data to plot - so if there's a .plot
+        if(trace._module.plot === _module.plot) {
+            moduleCalcData.push(cd);
+        }
+        else {
+            remainingCalcData.push(cd);
```
etpinard: Brilliantly efficient. :heart: this.

src/traces/ohlc/style.js
etpinard: > True, it looks weird... but it also looks weird in the plot.

Very good point. https://github.com/plotly/plotly.js/commit/f84dfae9d46b48fb84fd1be21c710a7dc2d346dd is :ok_hand:

src/traces/splom/base_plot.js
etpinard: :+1: for https://github.com/plotly/plotly.js/commit/0b7541e34d63ce74cddb5d243a652b2df0430b70 

> In principle we could extend this pattern (a plot step removes the calcdata items it plotted and returns the items that are left) which could have a small racehorse benefit when you have many traces of several types, you don't have to search the whole array each time, only the items that haven't already been handled.

Good idea here. We should keep that in mind when we'll :hocho: `calcdata` completely.

src/components/legend/style.js
```diff
@@ -52,7 +52,9 @@ module.exports = function style(s, gd) {
     .each(styleBoxes)
     .each(stylePies)
     .each(styleLines)
-    .each(stylePoints);
+    .each(stylePoints)
+    .each(styleCandles)
+    .each(styleOHLC);
```
etpinard: :rotating_light:  Non-blocking thinking-out-loud comment alert :rotating_light: 

We should perhaps make these trace module method (e.g. named `_module.legendStyle`) instead of having the legend component handle all trace types by itself.

alexcjohnson: Good idea, that would be way more efficient too - instead of calling every style routine on every trace and quitting if it's not the right one, just something like:

```js
.each(function(d) { d.trace._module.legendStyle(...) })
```

src/traces/ohlc/calc.js
alexcjohnson: That's true - sure, I'll change it to `= null`.

alexcjohnson: -> null https://github.com/plotly/plotly.js/pull/2561/commits/f498bd04e0541a4b6de7f7785d752fd6f1bb6481

src/components/rangeslider/draw.js
alexcjohnson: Ah, thanks for reminding me about the rangeslider + selection fix! 🔒-> https://github.com/plotly/plotly.js/pull/2561/commits/c77a8a319ccfc6aeff044e48cee561e281dec21a

etpinard:
Amazing work :dancer:

PR #2574, opened 2018-04-19
Better cartesian trace updates and removal

### LINKED ISSUE #2563: Better update pattern for cartesian SVG
As @etpinard pointed out during the finance refactor PR https://github.com/plotly/plotly.js/pull/2561#discussion_r181852562 and https://github.com/plotly/plotly.js/pull/2561#discussion_r181856275 we have a mess when redrawing cartesian SVG traces, both re: deleting and redrawing existing traces and re: removing completely unused modules.

The first part we can handle by making all the trace modules support more d3-idiomatic `enter/exit/each` logic as suggested in [this TODO](https://github.com/plotly/plotly.js/blob/82677ac8745ca56da740302977533dffcd5521b6/src/plots/cartesian/index.js#L201). The second has a few ways to go about it. Quoting @etpinard:

> What we need to do is find all the _gone_ modules (i.e. modules with `had??=true` and `has??=false`), loop over them, and then do what we need to do the cleanup the DOM. 
> 
> Better yet -- as more and more trace type will follow scatter and ohlc in not expecting a `selectAll('g.trace').remove()` before each `_module.plot()` -- we could add a trace module method (maybe called `cleanDOM`) to each trace modules that need to remove things for the DOM (or destroy WebGL objects). 
> 
> To note, geo subplots (and maybe others?) solve this problem in a different way by using `Plots.generalUpdatePerTraceModule` which calls `_module.plot` of visible **and**  _gone_ modules and expects the trace module plot methods to clean up the DOM accordingly (e.g. like [here](https://github.com/plotly/plotly.js/blob/18e19a86a3e683c1d11f6f4a04eb7f89d1b6108f/src/traces/choropleth/plot.js#L34) for choropleth traces). At present though, I think adding `cleanDOM` to each trace module would be best, as it would allow us to handle the `scattergl` and range-slider cases more easily.

etpinard:
Resolves #2563

I hope @alexcjohnson will like this solution.


src/plots/get_data.js
```diff
@@ -45,26 +45,42 @@ exports.getSubplotCalcData = function(calcData, type, subplotId) {
...
-exports.getModuleCalcData = function(calcdata, typeOrModule) {
+exports.getModuleCalcData = function(calcdata, arg1) {
     var moduleCalcData = [];
     var remainingCalcData = [];
-    var _module = typeof typeOrModule === 'string' ? Registry.getModule(typeOrModule) : typeOrModule;
-    if(!_module) return [moduleCalcData, calcdata];
+
+    var plotMethod;
+    if(arg1) {
+        var typeOfArg1 = typeof arg1;
```
alexcjohnson: Does a falsy `arg1` happen?

Also, I recall seeing somewhere that js can optimize `typeof arg1 === 'string'` internally so it never has to generate the string values, whereas with `var typeOfArg1 = typeof arg1; typeOfArg1 === 'string'` it can't do that so it's actually better to write out `typeof arg1 === '...'` in each clause. I can't find it now though...

alexcjohnson:
Beautiful. Only the one very nonblocking comment, lets do it! 💃

src/plots/get_data.js
etpinard: Thanks for tips! 

Simplified in -> https://github.com/plotly/plotly.js/pull/2574/commits/e5c860a10488e0ddfdb8eba445909a90b643f3c2

etpinard:
Thanks for the review @alexcjohnson, I think I'll wait a few days before merging this thing in case we need to make another patch release in the `1.36.x` series.

alexcjohnson:
Good call - like my upcoming react/transforms fix...

Issue #2470, opened 2018-03-12
transforms + Plotly.react bug


alexcjohnson:
A user encountered a problem with `Plotly.react` and transforms, changing pie aggregation & labels simultaneously. The following examples start with A/B/C and switch to X/Y/Z on clicking the button.

Works fine using `Plotly.restyle`:
https://codepen.io/alexcjohnson/pen/geaRej?editors=0010
![screen shot 2018-03-12 at 5 23 07 pm](https://user-images.githubusercontent.com/2678795/37310417-3cdb6b84-261a-11e8-8d89-ec28b6a78585.png)

But not with `Plotly.react`:
https://codepen.io/alexcjohnson/pen/xWwrXR?editors=0010
![screen shot 2018-03-12 at 5 24 20 pm](https://user-images.githubusercontent.com/2678795/37310438-45a6fecc-261a-11e8-8ab1-42d88105d021.png)

However `Plotly.react` works fine using the same aggregation that's built into pies since #2117:
https://codepen.io/alexcjohnson/pen/BroZNx?editors=0010

reported by @snazzypants in https://github.com/plotly/react-plotly.js/issues/62

etpinard:
Possibly related: https://community.plot.ly/t/plotly-js-cannot-read-property-length-of-undefined-when-clearing-filter-transforms/8890/4

Issue #1410, opened 2017-02-23
Stop pruning in nestedProperty?


alexcjohnson:
`Lib.nestedProperty` started as a nice compact, decoupled way to manage edits to traces and layout objects. We have a system that prunes empty containers, which was originally meant so that unnecessary pieces would not be left around after multiple operations. For example, you have a line trace with default properties:

`{x: [1, 2, 3], y: [1, 2, 3]}`

then you edit the line color `restyle(gd, {'line.color': 'red'})`

`{x: [1, 2, 3], y: [1, 2, 3], line: {color: 'red'}}`

And you don't like that, so you undo it `restyle(gd, {'line.color': null})`. Without pruning you get:

`{x: [1, 2, 3], y: [1, 2, 3], line: {}}`

Which is functionally the same, but it's not a perfect undo.

However, some empty containers cannot be removed (see https://github.com/plotly/plotly.js/pull/1403#discussion_r102331575 and https://github.com/plotly/plotly.js/pull/1403#discussion_r102759815 ) so this has gotten too coupled to our particular JSON structure. Can we live without doing any pruning since it would simplify the logic so much? @etpinard and I are leaning toward doing that but wanted to invite discussion.

rreusser:
I can imagine it would work, but is it too ugly to use a flag in the plot schema to prevent pruning but allow it by default?

rreusser:
Ohhh, never mind. Yeah, that's the coupling obviously.

alexcjohnson:
> Ohhh, never mind. Yeah, that's the coupling obviously.

and worse, `nestedProperty` isn't always called with a trace or complete `layout` as the root object.

rreusser:
Yeah, I think it's excellent to keep it fully decoupled and unaware of the schema. It seems to me like pruning is both a perfectly nice/valid thing and also something that shouldn't not be related to nested properties.

etpinard:
cc @bpostlethwaite and @VeraZab can you think of any potential problems that stopping pruning empty containers could cause in W2?

VeraZab:
I'm not sure I understand what the situation would look like without pruning..
So instead of calling `restyle(gd, {'line.color': null})` I would call `restyle(gd, line: {})` ?

Of the top of my head, I can think of one situation where we use `null`ing:
when I want to remove a shape or an image, I call `relayout(gd, {'images[2]': null})`
How would I do this now?

alexcjohnson:
[...]

PR #2577, opened 2018-04-22
Transform react

alexcjohnson:
A collection of fixes for transforms, particularly in conjunction with `Plotly.react` - closing #2470 and #2508, and along the way completing #1410.

The primary issue with transforms and `react` was the remapping of transformed arrays that we do at the end of `supplyDefaults` - the signal *not* to do this before a `recalc`, as initiated by `Plotly.restyle/relayout/update` was that we had already deleted `gd.calcdata`; but in `Plotly.react` we call `supplyDefaults` before we know what kind of a change we have, so we haven't deleted `calcdata` yet. The solution was pretty simple, to conditionally break out the remapping step and apply it later as required. This is https://github.com/plotly/plotly.js/commit/fcc459d985f0a3a80826da8df51b2daf150a4634

The other major thing I did here was to standardize the meaning of `trace._length`. It can either be a positive integer or `null`, and every trace type MUST define this during `supplyDefaults`:
- If it's an integer, that means the trace data is 1D and all the `arrayAttrs` describe the same collection of objects, point-by-point.
  - Transforms are allowed, and they will operate on the first `_length` points in each array.
- If it's `null`, that means either it has 2D data, or it's 1D but the arrays have different meanings - like nodes and links in `sankey`, or vertices and vertex indices in `mesh3d`.
  - Transforms are disabled. If we ever develop a transform that makes sense for any of these cases, we'll have to revise this condition, but currently all our transforms assume 1D data.

cc @etpinard @nicolaskruchten 
cc @bpostlethwaite @VeraZab this may require some minor updates to the transforms in chart studio for robustness, though I suspect they'll mostly work already... one thing to note is I had the built-in transforms here update `_length` when they run, but I don't think that's strictly necessary, since the regular `supplyDefaults` runs after the transform anyway.


nicolaskruchten:
Exciting! So how can a UI like react-chart-editor determine if a given trace in a figure is elligible for transforms? Is this available in the schema or in `_fullData` or somewhere or should we inline a list of trace types that support transforms?

alexcjohnson:
> So how can a UI like react-chart-editor determine if a given trace in a figure is eligible for transforms?

If you have `_fullData` look at `trace._length`: if it's a positive integer you can transform it, if it's `null` you can't. One caveat though, if the trace isn't `visible` it does not need to define `_length` so you wouldn't know. I suppose that applies everything though so it must be standard for the editor to not be able to really edit invisible traces.

We could add this to the schema I guess, but there are a few cases where the `type` isn't enough to know whether you can transform it or not. Chiefly `heatmap` and `contour`, where if `z` is 1D you *can* transform it, but if it's 2D you can't.

src/traces/heatmap/convert_column_xyz.js
```diff
@@ -13,24 +13,16 @@ var Lib = require('../../lib');
...
-    var col1 = trace[var1Name].slice(),
-        col2 = trace[var2Name].slice(),
-        textCol = trace.text,
-        colLen = Math.min(col1.length, col2.length),
-        hasColumnText = (textCol !== undefined && !Array.isArray(textCol[0])),
-        col1Calendar = trace[var1Name + 'calendar'],
-        col2Calendar = trace[var2Name + 'calendar'];
+    var colLen = trace._length;
+    var col1 = trace[var1Name].slice(0, colLen);
+    var col2 = trace[var2Name].slice(0, colLen);
```
etpinard: Oh so nice to `.slice()` only once!

src/traces/histogram/defaults.js
```diff
@@ -22,8 +22,10 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
     }
 
-    var x = coerce('x'),
-        y = coerce('y');
+    var x = coerce('x');
+    var y = coerce('y');
+    var hasX = x && x.length;
```
etpinard: I suspect `x && x.length` is faster than `Array.isArray(x) && x.length`. Might be worth standardizing `lib/is_array.js` (as e.g. `Lib.isNonEmptyArray`) ?

src/traces/mesh3d/defaults.js
```diff
@@ -86,4 +86,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     }
 
     coerce('text');
+
+    // disable 1D transforms
+    traceOut._length = null;
```
etpinard: Actually, `mesh3d` only expects 1D data, so `_length` could have a meaning and transforms could work.

etpinard: Ah nevermind, i/j/k and x/y/z lengths don't match for `mesh3d` traces. Can you add a comment about this above this line like you did for `sankey` traces?

src/transforms/sort.js
```diff
@@ -116,27 +116,18 @@ exports.calcTransform = function(gd, trace, opts) {
...
+    sortedArray.sort(getSortFunc(opts, d2c));
 
-                // clear sortedArray item to get correct
-                // index of duplicate items (if any)
-                sortedArray[j] = null;
-                break;
-            }
-        }
+    for(i = 0; i < len; i++) {
+        indices[i] = sortedArray[i].i;
```
etpinard: Brilliant :sparkles:

src/lib/nested_property.js
```diff
@@ -115,44 +113,21 @@ function npGet(cont, parts) {
...
- * to delete that.
- *
- * So we do some simple tests here to find known non-data arrays but don't worry too
- * much about not deleting some arrays that would actually be safe to delete.
+ * *args* arrays get passed directly to API methods and we should respect null if
+ * the user put it there, but otherwise null is deleted as we use it as code
+ * in restyle/relayout/update for "delete this value" whereas undefined means
+ * "ignore this edit"
  */
-var INFO_PATTERNS = /(^|\.)((domain|range)(\.[xy])?|args|parallels)$/;
```
etpinard: TODO:

- [ ] check that the old toolpanel does not break

src/lib/is_array.js
```diff
@@ -18,10 +18,20 @@ var dv = (typeof DataView === 'undefined') ?
...
     return ab.isView(a) && !(a instanceof dv);
-};
+}
+
+function isArrayOrTypedArray(a) {
+    return Array.isArray(a) || isTypedArray(a);
+}
+
+function is1D(a) {
+    return !isArrayOrTypedArray(a[0]);
```
etpinard: Maybe `isArray1D` would be clearer as this assumes that you're passing an array (or at least something you can index in). 

We might want to add a comment above about that assumption too, in case someone in the future wants to turn this into:

```js
return Array.isArray(a) && !isArrayOrTypedArray(a[0]);
```

src/plot_api/plot_api.js
```diff
@@ -2267,7 +2267,10 @@ exports.react = function(gd, data, layout, config) {
         gd.layout = layout || {};
         helpers.cleanLayout(gd.layout);
 
-        Plots.supplyDefaults(gd);
+        // "true" skips updating calcdata and remapping arrays from calcTransforms,
+        // which supplyDefaults usually does at the end, but we may need to NOT do
+        // if the diff (which we haven't determined yet) says we'll recalc
+        Plots.supplyDefaults(gd, true);
```
etpinard: I'm not a big fan of pattern. 

At the very least we should turn this into `Plots.supplyDefaults(gd, {skipDefaultsUpdateCalc: 1})` to make it more readable and extendable. 

Personally, I'd vote for removing the newly added `Plots.supplyDefaultsUpdateCalc()` from `Plots.supplyDefaults` and call `Plotly.supplyDefaultsUpdateCalc` after `Plots.supplyDefaults` in the places that new it.

On `master`, we have:

image

where the calls in `validate.js` and `rangeslider/draw.js` (and maybe more) don't need to update calc.

test/jasmine/tests/plot_api_test.js
```diff
@@ -2955,6 +2955,190 @@ describe('Test plot api', function() {
...
+                expect(gd._fullData.length).toBe(1);
+                var dims = gd._fullData[0].dimensions;
+                expect(dims.length).toBe(expectedVals.length);
+                expectedVals.forEach(function(expected, i) {
+                    expect(dims[i].values).toEqual(expected);
+                });
+            };
+        }
+
+        function reactWith(fig) {
```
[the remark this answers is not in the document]
etpinard: ... and I'm impressed that bbe3533 and standardizing `_length` was all we needed to get transforms to work with `splom` and `parcoords` :tada:

src/lib/nested_property.js
etpinard: Moreover @alexcjohnson, what if in the very unlikely scenario of someone writing in that their app relies on the old pruning behavior, what should we do?

I suspect reverting this commit would break some test(s) added in this PR (which one by the way?), so maybe we could formulate a solution for that hypothetical user?

etpinard:
Oh, one more thing. I don't think the checklist in https://github.com/plotly/plotly.js/issues/2508#issuecomment-378952799 got completed.

src/traces/heatmap/xyz_defaults.js
```diff
@@ -37,12 +39,14 @@ module.exports = function handleXYZDefaults(traceIn, traceOut, coerce, layout, x
         if(!isValidZ(z)) return 0;
 
         coerce('transpose');
+
+        traceOut._length = null;
```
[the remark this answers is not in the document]
alexcjohnson: > What if instead of null, 2D arrays would link a length-2 array to _length of row and column lengths?

I could imagine that being helpful but it sounds like a bit of a project in itself - not quite sure how it would mesh with the logic you linked for example, which may be at least partially after reshaping from columns to a grid. Anyway we can add that later, so I'd like to defer it for now.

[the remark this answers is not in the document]
etpinard: > so I'd like to defer it for now. 

Sounds good.

src/lib/nested_property.js
alexcjohnson: The reason I made this commit now was the `trace._length = null`, which in some cases was getting pruned later on in `supplyDefaults` by attributes with `dflt: null` and breaking my [redraw all the things test addition](https://github.com/plotly/plotly.js/pull/2577/files#diff-5a140ce1cc87a420496039bbf8d699f1R3276). I could have handled this in a few different ways:
- use something other than `null` for `_length` - I considered `false` but to me that had the implication of "has no length" rather than what I wanted, "(1D) length does not apply." Anyway the whole idea that we're pruning in the middle of `supplyDefaults` which is really only supposed to be *building* the `_full*` objects I wanted to avoid.
- make a special code path that does no pruning or unsetting - like `nestedProperty.onlySet(val)` - but that's adding complexity so not ideal.
- get rid of all `dflt: null` (and perhaps even enforce that) throughout our attributes. I still think we should do that, but I opted not to do it here.

Anyway, to the question:
> what if in the very unlikely scenario of someone writing in that their app relies on the old pruning behavior, what should we do?

If it leads to a visible change in the plot, we should fix it so the API does not contain such a dependence on the existence of empty containers ;)
If it just alters the data structures (`data` and `layout` I guess, after a `restyle`/`relayout`) I can't really predict why a user would be bothered by this, but we can help them wrap their `restyle`/`relayout` with something like `if(val === null) { pruneContainersPulledFromThisCommit(...) }`

src/traces/histogram/defaults.js
alexcjohnson: > e.g. `Lib.isNonEmptyArray`

Ah hmm, before I do that, I want to check this against our previous behavior for trace types that don't necessarily need both coordinates specified - like scatter, which can use `x0`/`dx` along with a `y` array. For example if `x=[], y=[1, 2, 3]` we should treat that as `_length=0`, not `3`, right? I'll need to look over our test coverage for those cases...

src/lib/nested_property.js
etpinard: Great answer. Thanks!

src/lib/is_array.js
alexcjohnson: https://github.com/plotly/plotly.js/pull/2577/commits/2a41f9ed9184014a543004ae6a5042334ccf8fec -> `isArray1D`

src/plot_api/plot_api.js
alexcjohnson: cleaner `supplyDefaults` API -> https://github.com/plotly/plotly.js/pull/2577/commits/2a41f9ed9184014a543004ae6a5042334ccf8fec

alexcjohnson:
> Oh, one more thing. I don't think the checklist in https://github.com/plotly/plotly.js/issues/2508#issuecomment-378952799 got completed.

completed in https://github.com/plotly/plotly.js/pull/2577/commits/965bcfb0d20717ba2aa2e2b1f4cda2d049ff4f08

test/jasmine/tests/plot_api_test.js
```diff
@@ -3211,86 +3217,127 @@ describe('Test plot api', function() {
...
+                mapboxAccessToken: require('@build/credentials.json').MAPBOX_ACCESS_TOKEN
+            });
+
+            _runReactMock(['scattermapbox', require('@mocks/mapbox_bubbles-text.json')], done);
+        });
+
+        it('tested every trace & transform type at least once', function() {
+            for(var itemType in typesTested) {
+                expect(typesTested[itemType]).toBeGreaterThan(0, itemType + ' was not tested');
+            }
```
alexcjohnson: So when a new trace or transform type is added we'll be forced to add it to this suite as well. (note above I excluded `contourgl` and, on CI, `scattermapbox`)

I couldn't think of a good way to ensure that every *component* gets tested as well... in fact looking over the list now I don't see shapes or 3D annotations, I'll have to add those. If anyone has ideas how to enforce this please chime in!

test/jasmine/tests/plot_api_test.js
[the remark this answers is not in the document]
nicolaskruchten: I would actually name this `reactTo` personally, in both senses of the word 'to' (with and towards) :)

test/jasmine/tests/plot_api_test.js
```diff
@@ -3027,70 +3219,127 @@ describe('Test plot api', function() {
...
+            })
+            .then(function() {
+                expect(fullJson()).toEqual(initialJson);
+                countCalls({});
+            })
+            .catch(failTest)
+            .then(done);
+        }
+
+        mockList.forEach(function(mockSpec) {
```
nicolaskruchten: Is there anything we can think of that could be asserted about the cartesian product here? I.e. given `M1` and `M2` is there anything in the JSON representation that we want to try to assert about `M1.reactWith(M2).reactWith(M1)` ?

nicolaskruchten: There are likely some things that are desirable that are already true in all cases and might be worth locking down, and some things that are desirable that are true in all but a small subset of cases and could be fixed easily (low-hanging fruits). There are also probably a large number of desirable things that aren't true and that are hard to fix, which I'm less interested in in the short run.

alexcjohnson: I tested a few of those transitions in https://github.com/plotly/plotly.js/pull/2577/commits/fcc459d985f0a3a80826da8df51b2daf150a4634#diff-5a140ce1cc87a420496039bbf8d699f1R3114 - and in fact after this PR (knock wood) I kind of don't think there will be a lot of nasty issues uncovered by testing this cartesian product. It's also going to be a very long test if you want to cover every combination, so it sounds to me more like we'd want to test them all once, offline, find the subset that fail, and incorporate just those into a new test.

src/traces/contourcarpet/calc.js
```diff
@@ -99,7 +99,7 @@ function heatmappishCalc(gd, trace) {
         z: z,
     };
 
-    if(trace.contours.type === 'levels') {
+    if(trace.contours.type === 'levels' && trace.contours.coloring !== 'none') {
```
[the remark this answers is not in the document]
alexcjohnson: possibly for the editor - `zmin`, `zmax`, and `zauto` were set in `_fullData`, implying that these were useful attributes for this case when in fact they are ignored.

test/jasmine/tests/plot_api_test.js
alexcjohnson: OK - still sounds a bit finicky... perhaps a later iteration we could add a test that ensures we've covered every attribute anywhere in the schema - that would be a significantly more stringent coverage test for traces and transforms as well! But if you don't mind I'll defer this for now.

etpinard: > But if you don't mind I'll defer this for now.

No problem :+1:

etpinard:
@alexcjohnson you might need to add a longer test timeout in one of your new tests in `plot_api_test.js` to make them pass consistently

image


Apart from that, this PR is :medal_sports: 

:dancer: for me.

test/jasmine/tests/plot_api_test.js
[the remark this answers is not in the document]
alexcjohnson: good call @nicolaskruchten -> `reactTo` in https://github.com/plotly/plotly.js/pull/2577/commits/cd08479f925745225692dd3440a24cd14ed52fbf

alexcjohnson:
Toolpanel works fine, the only thing I had to do was remove this line:
https://github.com/plotly/streambed/blob/master/shelly/plotlyjs/static/plotlyjs/src/plotly.js#L29

As far as I can tell `Plotly.Heatmap.hasColumns` isn't used anywhere anyway, but if I'm missing something and it still is needed we can point it to `Lib/isArray1D`.

PR #1972, opened 2017-08-28
Image testing with `image-exporter` electron runner

etpinard:
... or in other words, bye bye old image server and (hopefully) bye bye `imagetest` docker container.

To try this out:

- pull down https://github.com/plotly/image-exporter in a directory adjacent to plotly.js (or change the path to `image-exporter` in the plotly.js package.json)
- `npm i`
- then, try `npm run baseline` and  `npm run test-image`
- you also try  `npm run test-export` but this one needs https://github.com/plotly/image-exporter/pull/6 to work properly.
- :tada: 

What's new:

- `electron` based testing using [`image-exporter`](https://github.com/plotly/image-exporter) runner app
- no http requests are made, the runner app reads the JSON mocks directly from your file system
- a _pure_ node-js image comparison routine using Mapbox's [`pixelmatch`](https://github.com/mapbox/pixelmatch) and [`pngjs`](https://github.com/lukeapage/pngjs). That means bye-bye [`gm`](http://aheckmann.github.io/gm/) and `imageMagick`

Results:

- much faster test runs. On my machine, `npm run test-image` runs in under 1:30 minute (**N.B.** that's over 400 mocks folks)
- Should work consistently across all machines.

TODO:

- if this works out well for everyone, delete:
  + all `imagetest` related things in `tasks`
  + `test/image/assets/get_image_request_options.js`

PR #2579, opened 2018-04-24
Even better cartesian trace updates and removal

etpinard:
to be merged in #2574.

This PR implements @alexcjohnson 's idea (:pencil2: down in https://github.com/plotly/plotly.js/issues/2563#issuecomment-381980172) of making cartesian trace module layers (e.g. `<g.scatterlayer>`, `<g.barlayer>`, etc) data driven. Getting this to work was easy. Getting this to work well -- that is making sure layers don't get removed/added when they don't need to -- was a little trickier. Why? In brief d3's `select()` can mutate node data in some scenario (see [example](https://codepen.io/etpinard/pen/YLwvBE)). This is not the first time we noticed this behavior. Ricky once wrote:

https://github.com/plotly/plotly.js/blob/4ed586a6402073cc5c50a40cad5f652d7472fcce/src/traces/scattercarpet/plot.js#L33-L34

Oh well, see https://github.com/plotly/plotly.js/commit/284c206d89b8dfcdea6af2cb525c42872051bca5 for more details.

This PR also makes heatmap/contour/carpet plot methods know how to remove old traces in https://github.com/plotly/plotly.js/commit/da63ed4fa9030caba7d7ecdbebc24e8728753d39. These modules don't use `selectAll().data()` to build multiple traces. They instead use `trace.uid` to build and keep track of nodes. Previously, these old nodes were removed by looping over all traces and using some pretty nasty `selectAll()`. This significantly improves redraw perf on large (multi-subplot, multi-trace) graphs. For example, using this [script](https://codepen.io/etpinard/pen/NMPYow) with 30x30 subplots/traces and calling `relayout(gd, 'xaxis.range', [/*any*/])` used to take ~8000ms, and now is under ~3000ms (which is still slow, but hey it's a start :stuck_out_tongue_winking_eye: ).

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -34,7 +34,7 @@ case $1 in
 
     jasmine2)
         retry npm run test-jasmine -- --tags=gl --skip-tags=noCI,flaky
-        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI
+        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI,gl
         npm run test-bundle || EXIT_STATE=$?
         exit $EXIT_STATE
         ;;
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -34,7 +34,7 @@ case $1 in
 
     jasmine2)
         retry npm run test-jasmine -- --tags=gl --skip-tags=noCI,flaky
-        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI,gl
+        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI
         npm run test-bundle || EXIT_STATE=$?
         exit $EXIT_STATE
         ;;
--- a/src/plots/cartesian/constants.js
+++ b/src/plots/cartesian/constants.js
@@ -61,15 +61,16 @@ module.exports = {
     DFLTRANGEY: [-1, 4],
 
     // Layers to keep trace types in the right order
+    // N.B. each  'unique' plot method must have its own layer
     traceLayerClasses: [
-        'imagelayer',
-        'maplayer',
+        'heatmaplayer',
+        'contourcarpetlayer', 'contourlayer',
         'barlayer',
         'carpetlayer',
         'violinlayer',
         'boxlayer',
         'ohlclayer',
-        'scatterlayer'
+        'scattercarpetlayer', 'scatterlayer'
     ],
 
     layerValue2layerClass: {
--- a/src/plots/cartesian/index.js
+++ b/src/plots/cartesian/index.js
@@ -189,35 +189,33 @@ function plotOne(gd, plotinfo, cdSubplot, transitionOpts, makeOnCompleteCallback
 
     var layerData = [];
 
-    for(var i = 0; i < traceLayerClasses.length; i++) {
-        var className = traceLayerClasses[i];
+    for(var i = 0; i < modules.length; i++) {
+        _module = modules[i];
+        var name = _module.name;
 
-        for(var j = 0; j < modules.length; j++) {
-            _module = modules[j];
+        if(Registry.modules[name].categories.svg) {
+            var className = (_module.layerName || name + 'layer');
+            var plotMethod = _module.plot;
 
-            if(_module.basePlotModule.name === 'cartesian' &&
-                (_module.layerName || _module.name + 'layer') === className
-            ) {
-                var plotMethod = _module.plot;
+            // plot all traces of this type on this subplot at once
+            cdModuleAndOthers = getModuleCalcData(cdSubplot, plotMethod);
+            cdModule = cdModuleAndOthers[0];
+            // don't need to search the found traces again - in fact we need to NOT
+            // so that if two modules share the same plotter we don't double-plot
+            cdSubplot = cdModuleAndOthers[1];
 
-                // plot all traces of this type on this subplot at once
-                cdModuleAndOthers = getModuleCalcData(cdSubplot, plotMethod);
-                cdModule = cdModuleAndOthers[0];
-                // don't need to search the found traces again - in fact we need to NOT
-                // so that if two modules share the same plotter we don't double-plot
-                cdSubplot = cdModuleAndOthers[1];
-
-                if(cdModule.length) {
-                    layerData.push({
-                        className: className,
-                        plotMethod: plotMethod,
-                        cdModule: cdModule
-                    });
-                }
-                break;
+            if(cdModule.length) {
+                layerData.push({
+                    i: traceLayerClasses.indexOf(className),
+                    className: className,
+                    plotMethod: plotMethod,
+                    cdModule: cdModule
+                });
             }
         }
     }
+
+    layerData.sort(function(a, b) { return a.i - b.i; });
 
     var layers = plotinfo.plot.selectAll('g.mlayer')
         .data(layerData, function(d) { return d.className; });
--- a/src/plots/cartesian/index.js
+++ b/src/plots/cartesian/index.js
@@ -10,10 +10,13 @@
 'use strict';
 
 var d3 = require('d3');
+
+var Registry = require('../../registry');
 var Lib = require('../../lib');
 var Plots = require('../plots');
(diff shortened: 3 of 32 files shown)


src/traces/histogram2d/index.js
```diff
@@ -15,6 +15,7 @@ Histogram2D.attributes = require('./attributes');
 Histogram2D.supplyDefaults = require('./defaults');
 Histogram2D.calc = require('../heatmap/calc');
 Histogram2D.plot = require('../heatmap/plot');
+Histogram2D.layerName = 'heatmaplayer';
```
etpinard: **N.B.** Trace modules that reuse a plot method verbatim from another module need to declare `layerName` so that their traces are combined into the same trace module layer.

etpinard: .... there's probably a what to :hocho: `layerName` entirely by building some trace-to-layerName map object within the `Registry`, but oh well I didn't think adding a few of :arrow_heading_up: would be too bad. Thoughts?

src/traces/contour/plot.js
```diff
@@ -33,41 +33,30 @@ exports.plot = function plot(gd, plotinfo, cdcontours) {
...
+    var y = cd[0].y;
+    var contours = trace.contours;
+    var id = 'contour' + trace.uid;
+    var xa = plotinfo.xaxis;
+    var ya = plotinfo.yaxis;
+    var fullLayout = gd._fullLayout;
+    var pathinfo = emptyPathinfo(contours, plotinfo, cd[0]);
 
     // use a heatmap to fill - draw it behind the lines
+    var heatmapColoringLayer = Lib.ensureSingle(contourLayer, 'g', 'heatmapcoloring');
```
etpinard: See https://github.com/plotly/plotly.js/pull/2579/commits/48ee0d87cbcc872cb816ace72336c35711b6a7a0 for explanation.

src/plots/cartesian/index.js
```diff
@@ -179,40 +182,75 @@ exports.plot = function(gd, traces, transitionOpts, makeOnCompleteCallback) {
...
+                cdSubplot = cdModuleAndOthers[1];
+
+                if(cdModule.length) {
+                    layerData.push({
+                        className: className,
+                        plotMethod: plotMethod,
+                        cdModule: cdModule
+                    });
+                }
+                break;
```
alexcjohnson: Does this need to be a 2D loop? Could we just loop over modules, include the layer index in `layerData`, and sort by that at the end? I guess we'd also need to track which layers we've already included, to ensure uniqueness. But still seems faster than looping over all combinations of `className` and `_module`.

alexcjohnson:
re: tricky `.select()` behavior - see https://github.com/d3/d3/issues/1443#issuecomment-345548077

>  I've added an extension to d3-selection, selection.selectWithoutDataPropagation which does `d3.select(this.node().querySelector(name))`

In case using `selectAll` in these instances becomes a 🐎 issue or something...

src/traces/heatmap/plot.js
```diff
@@ -18,32 +19,28 @@ var xmlnsNamespaces = require('../../constants/xmlns_namespaces');
...
 var maxRowLength = require('./max_row_length');
 
+module.exports = function(gd, plotinfo, cdheatmaps, heatmapLayer) {
+    var i;
 
-module.exports = function(gd, plotinfo, cdheatmaps) {
-    for(var i = 0; i < cdheatmaps.length; i++) {
-        plotOne(gd, plotinfo, cdheatmaps[i]);
+    heatmapLayer.selectAll('.hm > image').each(function(d) {
+        var oldTrace = d.trace || {};
```
alexcjohnson: when is there no `d.trace`?

src/traces/carpet/plot.js
```diff
@@ -18,15 +18,20 @@ var svgTextUtils = require('../../lib/svg_text_utils');
...
-    }
+module.exports = function plot(gd, plotinfo, cdcarpet, carpetLayer) {
+    var i;
+
+    carpetLayer.selectAll('g.trace').each(function() {
+        var classString = d3.select(this).attr('class');
+        var oldUid = classString.split('carpet')[1].split(/\s/)[0];
+        for(i = 0; i < cdcarpet.length; i++) {
+            if(oldUid === cdcarpet[i][0].trace.uid) return;
+        }
```
alexcjohnson: 🌴 🐎 perhaps make a common helper `getUidsFromCalcdata` or something, that can be called once up front and returns a hash `{uid0: 1, uid1: 1, ...}`, then `if(!uidHash[oldUid]) d3.select(this).remove();`?
Looks like the same helper would work for all modules in this commit.

src/traces/histogram2d/index.js
alexcjohnson: I like it this way - explicit and clear.

src/plots/cartesian/constants.js
```diff
@@ -61,15 +61,16 @@ module.exports = {
...
-        'maplayer',
+        'heatmaplayer',
+        'contourcarpetlayer', 'contourlayer',
         'barlayer',
         'carpetlayer',
         'violinlayer',
         'boxlayer',
         'ohlclayer',
-        'scatterlayer'
+        'scattercarpetlayer', 'scatterlayer'
```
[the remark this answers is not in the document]
alexcjohnson: is there a reason some of the new ones share lines?

alexcjohnson:
@etpinard this is amazing - you make it look so easy, I thought it would be way more involved 😅
Just a few 🐎 suggestions/questions, then this'll be ready to go!

src/traces/heatmap/plot.js
etpinard: when a heatmap trace is completely off screen:

https://github.com/plotly/plotly.js/blob/bab2a5b692a21200e295c3909056818591e839ff/src/traces/heatmap/plot.js#L138-L142

tasks/util/container_commands.js
```diff
@@ -34,12 +35,17 @@ containerCommands.dockerRun = [
...
     var _commands = Array.isArray(commands) ? commands.slice() : [commands];
+    var cmd;
 
-    if(isCI) return getRunCI(_commands);
+    if(isCI) {
+        _commands = [containerCommands.ping].concat(_commands);
+        cmd = getRunCI(_commands);
+    } else {
+        _commands = [containerCommands.setup].concat(_commands);
+        cmd = getRunLocal(_commands);
```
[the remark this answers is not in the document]
etpinard: Yeah it's a little heavy, but at least now `npm run test-image` always works<sup>TM</sup>.

I should probably give https://github.com/plotly/plotly.js/pull/1972 another go at some point in the next few weeks. We really don't need to be running pixel comparison tests with a server, a CLI runner should do just fine.

src/traces/heatmap/plot.js
alexcjohnson: This is not a big deal, but wouldn't an off-screen heatmap just not contribute to the `'.hm > image'` selection at all?

etpinard: removing that `|| {}` makes

https://github.com/plotly/plotly.js/blob/c0b38dab5044310468355bfa533ccad516399f87/test/jasmine/tests/heatmap_test.js#L482-L506

fail with

image

etpinard: ... I didn't investigate further.

alexcjohnson: Alright, not what my quick glance at the code suggested but you're right, no need to dig in any further.

.circleci/test.sh
```diff
@@ -34,7 +34,7 @@ case $1 in
 
     jasmine2)
         retry npm run test-jasmine -- --tags=gl --skip-tags=noCI,flaky
-        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI
+        retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI,gl
```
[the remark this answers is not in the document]
etpinard: reverted in 8c9a7bf

alexcjohnson:
Beautiful - thanks for those couple of changes. This is a huge improvement not just in 🐎 but in structure and maintainabilty! 💃

Issue #1462, opened 2017-03-10
[Enhancement] Legend and traces three-way toggling


chrisvwn:
May be related to #674, but on advice from @rpaskowitz setting up as a separate request. I would like to suggest a 3-way toggling so that an entry can be on, off or highlighted. Sometimes, when trying to visually analyze a subset of plots in a crowded graph it might help to either dim the rest or just make a selected few stand out? So you still want the context of the other plots for comparison just not at the same visibility level as the selected subset.

Thanks to @rpaskowitz for suggesting the following: 

> I'd recommend doing something like that with a hover interaction, for example: http://c3js.org/samples/timeseries.html

My rejoinder: 

> Almost exactly what I was thinking only with a little more persistence so that one can have multiple highlighted.

jensb:
I would second that request. That would be quite useful.

etpinard:
Similar to https://github.com/plotly/plotly.js/issues/674. This should be doable using `plotly_legendclick` and `plotly_legenddoubleclick` added in https://github.com/plotly/plotly.js/pull/2581

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #665, opened 2016-06-20
Add ability to disable legend trace toggling


etpinard:
Some [folks](http://community.plot.ly/t/disable-legend-click-functionality-hiding-traces/1345) don't want it apparently; we should make this optional.

Maybe a boolean `layout.legend.tracetoggle` attribute?

hectorip:
I have implemented it could you check it, please?
 
https://github.com/hectorip/plotly.js/pull/1

I have created a pull-request in my own fork as stated in the guidelines. 
 Thanks in advance.

etpinard:
@hectorip you're off to a great start. I think that's all you need.

A couple recommendations before making your PR to the main repo:

- no need for that `true` in `coerce('tracetoggle', true);` In this case, the default value is static (i.e. it doesn't depend on other attributes).
- make sure `npm run lint` is clear of errors. I think you'll only need to remove that `console.log` 
- Instead of adding an image baseline mock json, you should add a jasmine test that make sure that clicking on the legend items with `tracetoggle: false` doesn't do anything. See suite [`legend_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/legend_test.js) and [`legend_scroll_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/legend_scroll_test.js) for inspiration.

Thanks for your efforts :beers:

john-soklaski:
@etpinard @hectorip What do you think about returning false from a callback to disable the default behavior?

I have a branch of plotly where I've added a [plotly_legend_toggle](https://github.com/plotly/plotly.js/compare/master...john-soklaski:toggle-legend-event) event so I can  know when to apply some custom scaling to the chart when a user toggles a trace.

It seems like there could be other uses for being able to influence the behavior of a toggle or react to it happening.

etpinard:
>  What do you think about returning false from a callback to disable the default behavior?

We _could_, but the plotly.js `data` / `layout` API needs to be JSON serializable to ensure that all our API [libraries](https://github.com/plotly/plotly.js#clients-for-r-python-node-and-matlab) have access to this feature.

@john-soklaski To treat the custom handler case, we would need a config option as proposed in https://github.com/plotly/plotly.js/issues/65 - which isn't at all incompatible with the `legend.tracetoggle` boolean attribute.

john-soklaski:
@etpinard That makes sense, thanks. Glad to see it would still be a possibility :)

[...]

Issue #65, opened 2015-12-03
click/hover events for labels/legends


tlrdstd:
I would like to be able to know when the user hovers or clicks on my label or legend items. Specifically, I would like to be able to highlight a given pie slice or bar trace when a particular label or legend item is hovered over. d3pie does this, and an example can be seen here: http://d3pie.org/#generator-start

As a starting point, it would be useful to have this `click` handler fire an event after it calls `Plotly.restyle` or `Plotly.relayout`: https://github.com/plotly/plotly.js/blob/47b13b0c2ecffae3ecb0b37d957978f3004a7ac7/src/components/legend/index.js#L484

A `hover` handler for the same traces would help a lot as well.

etpinard:
I really like this idea. 

Custom legend item click handlers could be set in the [plot config](https://github.com/plotly/plotly.js/blob/47b13b0c2ecffae3ecb0b37d957978f3004a7ac7/src/plot_api/plot_config.js) e.g.:

```js
Plotly.plot('graph', [], {}, {
   onLegendItemClick: function(data) {
      /* do something wilh data */
   }
}
```

The [current handler](https://github.com/plotly/plotly.js/blob/47b13b0c2ecffae3ecb0b37d957978f3004a7ac7/src/components/legend/index.js#L484) would simply be the default.

alexcjohnson:
> The current handler would simply be the default.

or, as in some other cases like [plotly_beforeplot](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L48-L49), it will still execute the default handler unless you return `false` from the custom handler

Abildtoft:
Thanks for an awesome library. Can I ask what the status is with regards to this?

abalter:
Here is a hacky way to do it. I did "inspect element" on the legend and found that it has a class `legendtoggle`. Using this, 

```
$('.legendtoggle').on('click', function(data)
{
    alert("legend clicked");
})
```

It doesn't say which trace item was clicked. However, if you are truly dedicated, you can climb back up to the legend group tag, then back down to the `text` element and grab the `data-unformatted` value.

walknwind:
We would like to give a user the ability to select series colors after viewing a plot. Ideally they would interact with the legend to do so. Another vote to resolve this issue :)

mauritzvdworm:
Another vote here.

[...]

PR #2581, opened 2018-04-25
Legend events

etpinard:
This PR adds legend two events `plotly_legendclick` and `plotly_legenddoubleclick`. Note that, as per @alexcjohnson 's recommendation in https://github.com/plotly/plotly.js/issues/65#issuecomment-162097388, these new events are emitted using `Events.triggerHandlers` which allows users to cancel the default click/doubleclick behavior via the (last) handler's return value. This PR chose this strategy despite this note:

https://github.com/plotly/plotly.js/blob/4ed586a6402073cc5c50a40cad5f652d7472fcce/src/lib/events.js#L93-L96

Is it fair to say `Events.triggerHandlers` is a good idea afterall or is there a better solution?

-----

Tagging this PR `discussion needed` as there are as few debatable things here:

- should we really use `Events.triggerHandlers`? One drawback of using it: the custom handlers must be called **before** the default handlers, meaning the users won't have access to the to-be-updated `visible` trace attribute values - which could be confusing for users that want to augment the default behavior (but maybe listening to `plotly_restyle` is good enough for that use case?)
- should `plotly_legendclick` and `plotly_legenddoubleclick` be the same event? For example, maybe both click and doubleclick could trigger `plotly_legendclick` where the click/doubleclick distinction would be made using addition event data field (e.g `numClick: 1 || 2`). This would allows users to define their own click-vs-doubleclick timeouts.
- can anyone think of any other useful field to include in the event data?

----

Referencing numerous related legend feature requests:

- click/hover events for labels/legends https://github.com/plotly/plotly.js/issues/65 (this PR would close the _click_ part)
- Add ability to disable legend trace toggling https://github.com/plotly/plotly.js/issues/665 (this PR adds a _way_, but it might be nice to add attributes for this so that the behavior can be exported)
- https://github.com/plotly/plotly.js/issues/674, https://github.com/plotly/plotly.js/issues/1462 are more specific requests. They would all be possible in user-land through the events added in this PR.

cc @jackparmer @willdurand

### DIFF
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -13,6 +13,7 @@ var d3 = require('d3');
 var Lib = require('../../lib');
 var Plots = require('../../plots/plots');
 var Registry = require('../../registry');
+var Events = require('../../lib/events');
 var dragElement = require('../dragelement');
 var Drawing = require('../drawing');
 var Color = require('../color');
@@ -336,21 +337,45 @@ module.exports = function draw(gd) {
                             e.clientY >= bbox.top && e.clientY <= bbox.bottom);
                     });
                 if(clickedTrace.size() > 0) {
-                    if(numClicks === 1) {
-                        legend._clickTimeout = setTimeout(function() {
-                            handleClick(clickedTrace, gd, numClicks);
-                        }, DBLCLICKDELAY);
-                    } else if(numClicks === 2) {
-                        if(legend._clickTimeout) {
-                            clearTimeout(legend._clickTimeout);
-                        }
-                        handleClick(clickedTrace, gd, numClicks);
-                    }
+                    clickOrDoubleClick(gd, legend, clickedTrace, numClicks, e);
                 }
             }
         });
     }
 };
+
+function clickOrDoubleClick(gd, legend, clickedTrace, numClicks, evt) {
+    var datum = clickedTrace.data()[0][0];
+
+    var evtData = {
+        event: evt,
+        node: clickedTrace.node(),
+        itemNumber: datum.i,
+        curveNumber: datum.trace.index,
+        data: gd.data,
+        layout: gd.layout,
+        frames: gd._transitionData._frames,
+        config: gd._context,
+        fullData: gd._fullData,
+        fullLayout: gd._fullLayout
+    };
+
+    var returnVal;
+
+    if(numClicks === 1) {
+        legend._clickTimeout = setTimeout(function() {
+            returnVal = Events.triggerHandler(gd, 'plotly_legendclick', evtData);
+            if(returnVal !== false) handleClick(clickedTrace, gd, numClicks);
+        }, DBLCLICKDELAY);
+    }
+    else if(numClicks === 2) {
+        if(legend._clickTimeout) clearTimeout(legend._clickTimeout);
+        gd._legendMouseDownTime = 0;
+
+        returnVal = Events.triggerHandler(gd, 'plotly_legenddoubleclick', evtData);
+        if(returnVal !== false) handleClick(clickedTrace, gd, numClicks);
+    }
+}
 
 function drawTexts(g, gd) {
     var legendItem = g.data()[0][0],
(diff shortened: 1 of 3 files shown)


alexcjohnson:
> should we really use `Events.triggerHandlers`? One drawback of using it: the custom handlers must be called **before** the default handlers, meaning the users won't have access to the to-be-updated `visible` trace attribute values - which could be confusing for users that want to augment the default behavior (but maybe listening to `plotly_restyle` is good enough for that use case?)

Not sure what that deprecation comment is about (@bpostlethwaite ?) but this seems like a clean solution to me. I don't see ^^ as a downside, in fact it's really the only option if you want to be able to avoid the default behavior in the first place right? I suppose we could make even more events (`plotly_legendclicked`?) but unless someone is really bothered by it I'd just leave it at `plotly_restyle`.

> should `plotly_legendclick` and `plotly_legenddoubleclick` be the same event? For example, maybe both click and doubleclick could trigger `plotly_legendclick` where the click/doubleclick distinction would be made using addition event data field (e.g `numClick: 1 || 2`). This would allows users to define their own click-vs-doubleclick timeouts.

referencing https://github.com/plotly/plotly.js/issues/1546 - if we're ever going to add `plotly_singleclick` then perhaps for consistency we should have three events here - or at least ensure that the two we're introducing match names <-> behaviors with two of the (eventually) three data click events (so if this one behaves like `plotly_singleclick` perhaps name it `plotly_legendsingleclick`?)

> can anyone think of any other useful field to include in the event data?

For use with our favorite transform, `groupby`, it would be nice to include `_expandedIndex` and maybe `_group` if it exists.

bpostlethwaite:
I think JQuery deprecated certain behaviours with triggerHandler for versions >2 but I can't find any reference to this with a quick search. IMO if this function has useful functionality or more importantly an intuitive and useful API then we can delete that comment.

etpinard:
Currently (i.e. excluding the new events in this PR),  `Events.triggerHandler` is used for `plotly_beforeplot` and `plotly_beforehover`. These two events are likely to get :hocho: in v2: I can't think of a good use case for `plotly_beforeplot` outside the old workspace and there are now better ways to disable hover than `plotly_beforehover`. 

So, I suspect when that comment about :hocho: `Events.triggerHandler` got :pen:  we were already planning on :hocho: `plotly_beforeplot` and `plotly_beforehover` meaning `Events.triggerHandler` would be w/o a callee in our codebase. Now, looks like `Events.triggerHandler` is here to stay.

etpinard:
> For use with our favorite transform, groupby, it would be nice to include _expandedIndex and maybe _group if it exists.

done in https://github.com/plotly/plotly.js/pull/2581/commits/02a85bf609bddfad3d6e9d80ee44b3bed40478e4

etpinard:
Oh maybe we wanted to deprecate `Events.triggerHandlers` because it doesn't work properly with [`gd.once`](https://nodejs.org/api/events.html#events_emitter_once_eventname_listener) spitting out

image

etpinard:
@alexcjohnson 

- so https://github.com/plotly/plotly.js/pull/2581#issuecomment-385549532 turned out to be fixable (see https://github.com/plotly/plotly.js/pull/2581/commits/9a7cc143ffffd8a7c1c117fd2174a1532fd2f521) albeit in a bit of hacky way
- from our private convo, I made `plotly_legendclick` get triggers on mousedown in https://github.com/plotly/plotly.js/pull/2581/commits/fde209f007ec2dc90aa22992c4a09fddedab417a
- and I added `label` to the event data for pie traces in https://github.com/plotly/plotly.js/pull/2581/commits/05e1b6b8c21258dc619b3732f9b59bdb827d1f1c

alexcjohnson:
Very nice. 💃

jacobq:
Are these new events in the docs somewhere? (I don't see them [where I expected them to be](https://plot.ly/javascript/plotlyjs-events/#additional-events).)

Issue #2459, opened 2018-03-08
Regression bug: LASSOing has become significantly slower between 1.31 to 1.32


daattali:
Codepen with sample code: https://codepen.io/anon/pen/NYWNde

Having 10,000 points in a scatterplot, if I wanted to LASSO select some points with plotly1.31 it was fairly smooth. If I try the same thing with plotly1.32+ it becomes agonizingly slow.

![yearlyadepthornshark-max-14mb](https://user-images.githubusercontent.com/952340/37163463-d8158df8-22c5-11e8-9d84-6c98b0206f05.gif)

On Windows 7 Pro, using Chrome v64

Originally reported to @cpsievert at https://github.com/ropensci/plotly/issues/1216

etpinard:
Thanks for the report! To be more precise, the issue is with SVG `scatter` (i.e. `scattergl` is fine).

etpinard:
Arrg. I suspect https://github.com/plotly/plotly.js/pull/2135 is the culprit here. We'll have to speed up `_module.style`. Thanks again for the report!

etpinard:
One more reason to start stashing points selections (cc https://github.com/plotly/plotly.js/issues/2548)

daattali:
As the issue opener, I just want to confirm this is indeed fixed :)

Issue #2548, opened 2018-04-11
Stash SVG points selections for faster pan


etpinard:
PR https://github.com/plotly/plotly.js/pull/2527 introduced a 'draggedPts' trace category (see https://github.com/plotly/plotly.js/pull/2527#discussion_r179595111) which was later deemed a potential overkill by @alexcjohnson in https://github.com/plotly/plotly.js/pull/2527#discussion_r179606748 to make the pan hot code path avoid large `.selectAll()` (which can be very slow). 

What we should try to do instead is: stash those points selections (probably in `fullLayout._plots[subplotId]`) early (probably during the trace plot step) and use them directly in pan (and scroll?) hot code path.

etpinard:
While doing this, we should investigate ways to remove more things from that hot pan code path like various `fullLayout._has` return values (mentioned [here](https://github.com/plotly/plotly.js/pull/2527/files#diff-1bd3dd52cddb8638876bbe17cef28a0dR699)). 

Moreover, we could perhaps improve performance even more by combining `Drawing.setScale` and `Drawing.setTranslate` into one method ([ref](https://github.com/plotly/plotly.js/pull/2527/files#diff-1bd3dd52cddb8638876bbe17cef28a0dR778)).

PR #2583, opened 2018-04-27
SVG trace on selection perf

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2459

There's still plenty of room for improvements (for example via https://github.com/plotly/plotly.js/issues/2548), but this PR gets scatter selection perf back to a level comparable to before the persistent selection PR (in https://github.com/plotly/plotly.js/pull/2135). Unfortunately, this means adding yet another method on the trace module objects. I hope @alexcjohnson won't mind. In brief, the new `styleOnSelect` methods only update DOM styles corresponding the `selected` and `unselected` attributes and thus by-passes much of the regular `style` DOM updates.

In numbers from the start to end of the selecting [throttle](https://github.com/plotly/plotly.js/blob/05b0ec98268928e91c0032bb2a5053ebf839435f/src/plots/cartesian/select.js#L279-L301) using this [codepen]( https://codepen.io/anon/pen/NYWNde) from #2459, before this PR we had ~300-400ms: and after: ~40-60ms

In gifs:

before

peek 2018-04-27 10-19

after

peek 2018-04-27 10-20


src/plots/cartesian/select.js
```diff
@@ -404,18 +404,20 @@ function updateSelectedState(gd, searchTraces, eventData) {
         var len = items.length;
         var item0 = items[0];
         var trace0 = item0.cd[0].trace;
+        var _module = item0._module;
+        var fn = _module.styleOnSelect || _module.style;
```
alexcjohnson: I like `styleOnSelect`, very nice solution!

But please call it something more descriptive than `fn` here - `styleSelection`?

etpinard: sure thing :wink:

src/traces/bar/style.js
```diff
@@ -35,34 +35,55 @@ module.exports = function style(gd, cd) {
...
+
+function styleOnSelect(gd, cd) {
+    var s = cd[0].node3;
+    var trace = cd[0].trace;
+
+    if(trace.selectedpoints) {
+        Drawing.selectedPointStyle(s.selectAll('path'), trace, gd);
+        Drawing.selectedPointStyle(s.selectAll('text'), trace, gd);
+    } else {
+        stylePoints(s, trace, gd);
```
alexcjohnson: Why do we need this `else`? I don't see a corresponding case in the previous code...

BTW, I'm hesitant to suggest this after the `.select().select()` data-mangling mess, but does it work to do `Drawing.selectedPointStyle(s.selectAll('path,text'), trace, gd)`? Is that any faster?

etpinard: Ah oops, this should be `Drawing.selectedTextStyle(s.selectAll('text'), trace, gd)`. I'll add a test.

etpinard: > Why do we need this else? I

To get back to the original state after double-click. `Drawing.selectedPointStyle` and `Drawing.selectedTextStyle` only handle `selected` / `unselected` styling off a "base" state given by `stylePoints`.

src/plots/cartesian/select.js
etpinard: done in 0e91c6e

src/traces/bar/style.js
etpinard: > I'll add a test.

added in https://github.com/plotly/plotly.js/pull/2583/commits/01df2d2179d0a117fc17deee288047a637cffab0

etpinard:
Commit https://github.com/plotly/plotly.js/pull/2583/commits/3776e8f20299cd9b3ecb5db59950a3355481a075 is a little messy, so I'll explain it here:

Previously (i.e. since the persistent selection PR https://github.com/plotly/plotly.js/pull/2135), [un]selected styles were applied to the DOM nodes assuming that the base style had been already applied to DOM nodes. Meaning that to get [un]selected style right, we needed to call `Drawing.pointStyles` first and then `Drawing.selectedPointStyle()` + `Drawing.selectedTextStyle()`. This turned out to be slow especially during selection (i.e. on drag). In other words, before commit 3776e8f20299cd9b3ecb5db59950a3355481a075, calling `Drawing.selected(Point|Text)Style` without `Drawing.pointStyle` lead to bugs. 

Now post commit 3776e8f20299cd9b3ecb5db59950a3355481a075, `Drawing.pointStyles` knows how to apply [un]selected styles. This means DOM nodes `style` attributes are updated only once per `Plots.style` call. This should improve first-render perf for traces with [un]selected attributes. 

Moreover, `Drawing.selected(Point|Text)Style` don't assume a base state anymore. They know how to handle cases where certain [un]selected keys are unset. This means that `styleOnSelect` (which bypasses `Drawing.pointStyle` for speed) now does the right thing<sup>TM</sup>.

src/components/drawing/index.js
```diff
@@ -355,88 +425,36 @@ function singlePointStyle(d, sel, trace, markerScale, lineScale, marker, markerL
...
- * except all at once before a full redraw.
- * The upside of this is arbitrary points can share gradient defs
- */
-drawing.initGradients = function(gd) {
-    var gradientsGroup = Lib.ensureSingle(gd._fullLayout._defs, 'g', 'gradients');
-    gradientsGroup.selectAll('linearGradient,radialGradient').remove();
 };
 
-drawing.singlePointStyle = function(d, sel, trace, markerScale, lineScale, gd) {
+drawing.makePointStyleFns = function(trace) {
```
etpinard: Define trace-wide per-pt functions w/o knowledge of d3 or DOM method so that non-svg module (e.g. `scattermapbox`) can reuse them :recycle:

src/components/drawing/index.js
```diff
@@ -455,84 +473,119 @@ drawing.makeSelectedPointStyleFns = function(trace) {
...
-            // save for selectedTextStyle
+            // save for Drawing.selectedTextStyle
             d.mrc2 = mrc2;
         });
     }
+
+    s.each(function(d) {
+        for(var i = 0; i < seq.length; i++) {
+            seq[i](d3.select(this), d);
+        }
```
alexcjohnson: interesting approach with `seq` - it's not obvious to me that this is meaningfully different from 3 `s.each` calls, or 3 `if(functionExists)` blocks inside one `s.each`, did it make a big difference to do it this way?

etpinard: d3's each are a little slower than regular forEach and for-loops. Nothing worth losing sleep over, but we do gains a few ms using this sequence thing.

etpinard: ... by gain, I mean lose a few ms and gain performance. :smile:

alexcjohnson:
Looks great! Thanks for the help understanding https://github.com/plotly/plotly.js/pull/2583/commits/3776e8f20299cd9b3ecb5db59950a3355481a075

💃

PR #2587, opened 2018-04-27
Blank editable legend item fixes

### LINKED ISSUE #2576: Deleting groupby names via legend doesn't update correctly
- Create a legend with groupby items, in editable mode
- Delete one of the default labels -> after blur, it stays empty (**the default label should come back**).
- Do something else that redraws the graph (including typing in another label) -> now the default label comes back

see https://codepen.io/alexcjohnson/pen/erNPjB?editors=0010 - based on the mock at
https://github.com/plotly/plotly.js/blob/4ed586a6402073cc5c50a40cad5f652d7472fcce/test/jasmine/tests/legend_test.js#L918-L931

alexcjohnson:
Fixes #2576 
- In editable mode, pads all blank or short legend items out to the length of the longest one (or 4 characters at least) - note this is the character length, not the svg text length which will usually be shorter.
- `groupby` and regular traces work the same now - deleting the value completely leaves it blank, rather than reverting to the default like `groupby` had been doing.

cc @etpinard

### DIFF
--- a/src/components/legend/draw.js
+++ b/src/components/legend/draw.js
@@ -41,9 +41,9 @@ module.exports = function draw(gd) {
 
     if(!gd._legendMouseDownTime) gd._legendMouseDownTime = 0;
 
-    var opts = fullLayout.legend,
-        legendData = fullLayout.showlegend && getLegendData(gd.calcdata, opts),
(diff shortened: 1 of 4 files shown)


etpinard:
Looks good :dancer:

PR #2590, opened 2018-04-27
fix spikesnap: cursor in FF

### LINKED ISSUE #2588: Spike line shifted in Firefox
The spike lines don't work correctly in Firefox when setting to spikesnap: 'cursor'.
The example below works well in Chrome or IE, but in Firefox the spike lines are shifted.

https://codepen.io/sunqingsong/pen/RyoBjp

alexcjohnson:
Fixes #2588 
`evt.offset(X|Y)` isn't implemented consistently in Firefox vs other browsers. Stopped using it and blacklisted it as a property.

The little offset I saw in Chrome https://github.com/plotly/plotly.js/issues/2588#issuecomment-385072182 seems to be a weird interaction of Chrome and my second monitor (non-retina, vs. the laptop's retina monitor which does not show this offset) that shows up even in bare `mouseover` events, so doesn't seem like we can fix that.

cc @etpinard

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -279,11 +279,9 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         if(hasUserCalledHover) {
             if('xpx' in evt) xpx = evt.xpx;
             else xpx = xaArray[0]._length / 2;
-            evt.pointerX = xpx + xaArray[0]._offset;
 
             if('ypx' in evt) ypx = evt.ypx;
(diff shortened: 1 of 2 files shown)


etpinard:
:1st_place_medal: bug fix day for @alexcjohnson 

:dancer:

etpinard:
oh by the way, I wasn't able to reproduce https://github.com/plotly/plotly.js/issues/2588#issuecomment-385072182 on my laptop.

sunqingsong:
Thank you for the quick fix!

PR #2592, opened 2018-04-29
Update modebar buttons when locale changes

### LINKED ISSUE #2484: Changing locale does not update mode bar buttons labels
I created pen initially in German - the buttons titles and dates on X-axis appears in German. Then changed the locale to English. The date format on X-axis updated but the buttons still contain the German texts.

https://codepen.io/exavier/pen/gegOmV?editors=1010

alexcjohnson:
Fixes #2484 

@etpinard another quick one I didn't quite finish on bug day :) 
FWIW it fixes not only the encouraged usage (`Plotly.react`) but also the old `Plotly.plot` call with new `config`.

### DIFF
--- a/src/components/modebar/modebar.js
+++ b/src/components/modebar/modebar.js
@@ -52,10 +52,13 @@ proto.update = function(graphInfo, buttons) {
     else this.element.className = 'modebar';
 
     // if buttons or logo have changed, redraw modebar interior
-    var needsNewButtons = !this.hasButtons(buttons),
-        needsNewLogo = (this.hasLogo !== context.displaylogo);
+    var needsNewButtons = !this.hasButtons(buttons);
+    var needsNewLogo = (this.hasLogo !== context.displaylogo);
+    var needsNewLocale = (this.locale !== context.locale);
 
-    if(needsNewButtons || needsNewLogo) {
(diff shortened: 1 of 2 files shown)


etpinard:
Beautiful fix :dancer:

vdgodse:
@alexcjohnson Shouldn't it also check if event handlers were updated? Callbacks may need to be updated while using plotly with React hooks, so that it doesn't use stale values from the old closure. I ran into an issue where it seems like plotly is not updating the onClick handler for modebar buttons.

alexcjohnson:
@vdgodse please file a new issue showing the misbehavior you've discovered, and we'll take a look. If at all possible, try to reduce it to plain javascript, ie `Plotly.newPlot(...).then(...).then(() => Plotly.react(...))` as that's far easier to debug than issues inside a React build (and easier to share in Codepen or the like).

vdgodse:
Sure. I'll open an issue once I reduce it to plain javaScript.

vdgodse:
> Sure. I'll open an issue once I reduce it to plain javaScript.

Update: Created an issue here: [https://github.com/plotly/plotly.js/issues/5796#issue-935706684](url) along with codepen demo

Issue #2038, opened 2017-09-26
Ability to interactively change length and rotate line shapes


chriddyp:
Some dash users are interested in creating image analysis apps using background images and interactive line segments to measure distances on the image.

In order to do this, they need to be able to easily rotate and resize line segments.

Currently, it is possible to translate line segments but not reshape them.

![move-line](https://user-images.githubusercontent.com/1280389/30875280-ff9b1dd2-a2c0-11e7-8719-9e8030bf047c.gif)

cc @alexcjohnson

rreusser:
Would be neat to have a dedicated 'ruler' feature with which you could click and measure, say slope or distance in data coords, though definitely a moderately tricky feature by itself.

alexcjohnson:
More detail: horizontal and vertical lines can't be reshaped, only moved. Diagonal lines *can* be reshaped, but only within the quadrant they start in. Sounds like they're using the same algorithm as shapes with area, which isn't really appropriate there. They should use something closer to what annotations use - which also includes invisible handles which are wider than the line itself, particularly at the end, to make it easier to grab.

A question: should we snap to horizontal and vertical, and/or the original angle the line had? Most of the time this is useful, but on occasion you really do want slightly off-angle lines. Annotations currently have no snapping. I suppose one option would be no snapping at first, but some time later add snapping at the same time as we add the ability to *select* just one endpoint/corner and move it pixel-by-pixel with arrow keys.

rmoestl:
I'm currently working on that issue (without snapping, see above) and a major question arose:

Should the *reshape line* mode (as opposed to *move line* mode) be active "near" both ends only or would it be okay that the reshape mode is active for the first 1/3 and last 1/3 of the line (similar to https://github.com/plotly/plotly.js/blob/5bc25b490702e5ed61265207833dbd58e8ab27f1/src/components/dragelement/cursor.js#L24)?

IMHO it'd be nicer to have the *reshape line* mode be active only "near" both ends of the line. However, in my attempt to implement this, I came across worrisome differences between what `getClientBoundingRect()` is returning for SVG elements in different browsers. For example Chrome does not include the stroke while FF does. And in addition FF is returning way too big bounding boxes for my test cases, which means that the sensitive area for *reshape line* would be off of the helper path's visible area. For more info about those browser inconsistencies see [SVGElement.getBoundingClientRect(): which bounding box? · Issue #339 · w3c/svgwg](https://github.com/w3c/svgwg/issues/339).

[...]

PR #2594, opened 2018-05-01
Move and resize line shapes

rmoestl:
This implements a *more intuitive movement and resizing of line shapes* specified in #2038. Please refer to the commit message for a rough overview of the feature.

In addition renamed image test 'fixed_size_shapes' to 'shapes_fixed_size' to better fit the de-factor naming convention for image tests.


alexcjohnson:
Sweet! I haven't looked at the code yet, just playing with your awesome `shapes_move-and-reshape-lines` mock 🎉 The mechanics of it are great! But I noticed a few things:
- [ ] I like the `move` cursor over the middle section, but how about `grab` for the ends? Won't work in IE but I think that's OK, it'll fall back to `default` there I guess, which is what you use now.
- [ ] Not part of this PR but I also notice we don't do anything special with the cursor when an annotation is draggable. It's a little tricky what to do with the text box, as click edits the text and drag moves it... but the arrow itself could safely get a `move` cursor. Your call whether to add that to this PR or make a new issue for it.
- [ ] When you drag a shape off the edge of an autoranged plot, dragging still works but part of the shape is hidden by the `clip-path` - could be helpful to remove the clipping during the move. I guess to be precise about it, if one of the axes is autoranged and the other is fixed, we should only stop clipping on the autoranged axis. Note that there are single-axis `clipPath`s already defined and available for this purpose.
- [ ] Also maybe not part of this PR, but came up while I was playing: seems like there's something off with y-autorange + fixed-size shapes. When one end of the green line is highest, the autorange is about right, but when the other end is highest it reserves too much space. Does not seem to happen in the x direction:
screen shot 2018-05-01 at 9 33 26 am

alexcjohnson:
Good call renaming 'fixed_size_shapes' to 'shapes_fixed_size' - I hadn't noticed that but yes, it's nice to keep all the shapes mocks together in our various listings.

src/components/shapes/draw.js
```diff
@@ -146,39 +154,90 @@ function setupDragElement(gd, shapePath, shapeOptions, index) {
...
+    function obtainSensoryElement() {
+        return isLine ? createLineDragHandles() : shapePath;
     }
 
-    function startDrag(evt) {
-        // setup conversion functions
-        xa = Axes.getFromId(gd, shapeOptions.xref);
-        ya = Axes.getFromId(gd, shapeOptions.yref);
+    function createLineDragHandles() {
+        var minSensoryWidth = 10,
```
alexcjohnson: Might want to move this into https://github.com/plotly/plotly.js/blob/master/src/constants/interactions.js?

This behavior for annotations is even more ugly and hard-coded
https://github.com/plotly/plotly.js/blob/2b1713d583f9871b2cbf98a566e61a473632afc2/src/components/annotations/draw.js#L536-L540
It wouldn't need to be changed now, but if we were to move this to a constants file and reuse it in annotations, I like the `Math.max` logic you have here better than the `width + 6` currently in annotations (though the little-bit-extra around the head should be preserved).

alexcjohnson:
Oh hmm, re: the y-autorange issue: I think I just got the anchor point outside the actual range of the line, so it was autoranging to the anchor point, which is totally reasonable! I'm not really sure how to handle this, but as it is it's too easy to get into strange unintuitive situations, since there's nothing to show you visually where the anchor point is.

The easy way to handle this would be just not to allow GUI resizing of pixel-sized shapes, only moving them. Or perhaps allow resizing but not past the anchor point, so the shape always contains the anchor? That would cause problems for lines though, as you'd have the same problem this PR is intended to fix, of not being able to change the direction the way you want to. A more complete solution might be to show the anchor point during the resize - that would let people see what they're really doing. Perhaps a crosshair, white with a black outline like we make for zoom box corners?

etpinard:
> but I also notice we don't do anything special with the cursor when an annotation is draggable

That one has its own issue already -> https://github.com/plotly/plotly.js/issues/679

rmoestl:
Re: cursors for annotations
> but the arrow itself could safely get a move cursor
Done.
>  It's a little tricky what to do with the text box, as click edits the text and drag moves it...
What about the `text` or `pointer` cursor for the text and the `move` cursor for the rect around the text? E.g. Trello is using `text` and `pointer` cursors to indicate direct editing of card and list titles.

rmoestl:
Re: auto-range, anchor points and fixed size shapes, in my opinion showing the anchor point (and perhaps a thin line connecting the shape with the anchor point) would the way to go.

alexcjohnson:
> What about the `text` or `pointer` cursor for the text and the `move` cursor for the rect around the text? E.g. Trello is using `text` and `pointer` cursors to indicate direct editing of card and list titles.

Currently the two behave the same when you click or drag them, I think - so I'd keep the cursor the same for both elements. Maybe `pointer` is best to capture the multiple possibilities? I suppose we could also use `text` but change it to `move` once you start dragging? Dunno, I don't have a strong opinion.

rmoestl:
Short comment on showing the **anchor point**: dropped the idea to additionally display a thin line. IMO showing the cross-hair is enough of a visual cue. Felt quite natural to me, please use https://github.com/plotly/plotly.js/blob/move-and-resize-line-shapes/test/image/mocks/shapes_fixed_size.json to see how it behaves in different edge cases.

alexcjohnson:
> dropped the idea to additionally display a thin line. IMO showing the cross-hair is enough of a visual cue. Felt quite natural to me

Agreed, this feels very clear to me as you have it now, makes it really obvious what's going on and how to get the alignment you want. Nice! Thanks for adding the annotation cursors as well, make sure you close #679 in addition to #2038 when this is merged 🎉 

I haven't looked at the code yet, but two comments on the mechanics:
- The cross gets blurry much of the time. `.crisp` can help, though even with that I've seen cases where corner pixels get dropped unexpectedly. The best solution is to round its positioning to  exact half-integer pixels - hence all the `.5`'s [here](https://github.com/plotly/plotly.js/blob/57eb8ef3314017b262ed43fbb50edc3033020d14/src/plots/cartesian/dragbox.js#L1014-L1041) - but that one doesn't need to explicitly round because it's positioned based on a mouse event... yours is placed based on data.

only a little blurry:
less blurry

very blurry:
blurry

- I'm not quite sure what to do with the (very unusual, but supported) case of one axis fixed-pixels and the other data- or plot-referenced... Having the anchor cross appear at one edge of the non-pixel dimension feels weird though. Perhaps it would be enough to put it at the center of that dimension? Even better might be to put it at the center AND make it just a bar - a horizontal bar if only the y direction is fixed pixels, a vertical bar if only x is fixed. Like we do for 1D zooms:
screen shot 2018-05-08 at 11 38 34 am 2

> Also, in case there's a place in the code to gather functions together that generate some kind of "symbol" like the cross-hair shape, please let me know.

No, not at present. We do have nearly a identical path defined in [ternary](https://github.com/plotly/plotly.js/blob/57eb8ef3314017b262ed43fbb50edc3033020d14/src/plots/ternary/ternary.js#L439). That one is smaller than the one you made, which probably makes sense, so I'm not sure DRYing this up is a good goal, though there would be something to be said for collecting these paths at least to ensure the differences between them are intentional. Anyway I wouldn't worry about that in this PR but it's a good point to keep in mind for later.
screen shot 2018-05-08 at 11 28 35 am 2

src/components/shapes/draw.js
```diff
@@ -108,20 +108,24 @@ function drawOne(gd, index) {
...
+        setClipPath(path, gd, options);
 
         if(gd._context.edits.shapePosition) setupDragElement(gd, path, options, index, shapeLayer);
     }
 }
 
+function setClipPath(shapePath, gd, shapeOptions) {
+    // note that for layer="below" the clipAxes can be different from the
+    // subplot we're drawing this in. This could cause problems if the shape
+    // spans two subplots. See https://github.com/plotly/plotly.js/issues/1452
```
alexcjohnson: That issue is closed - is there still a case where this currently breaks? I see https://github.com/plotly/plotly.js/issues/1452#issuecomment-285522747 about layering and inset plots, is that what you're referring to?

alexcjohnson: Oh sorry, you just copied this comment on refactor. NVM!

rmoestl:
> I'm not quite sure what to do with the (very unusual, but supported) case of one axis fixed-pixels and the other data- or plot-referenced

Me neither :smile:. I was thinking about that case. My second option would have been a bar with the full length of the not-pixel-sized dimension of the shape. But then came to the conclusion that this would probably confuse people.

My rationale for the current solution is that (i) positioning the anchor cross on the x0 / y0 coordinate of the not-pixel-sized dimension gives the user a useful hint where the shape "starts" and (ii) drawing a cross even in this case is a more consistent user experience.

Guess, it's best to test it out.

alexcjohnson:
Nicely done. I like it with the small centered bars for the one-axis-fixed-pixels case. Now that they're centered, rather than sitting on one (seemingly arbitrary) edge, I could see going back to a cross, but to my eye anyway it's a little bit clearer and more intuitive as a bar.

The drawing is very crisp now too, thanks for cleaning that up. I think we're ready! 💃

PR #2603, opened 2018-05-02
Fix Plotly.react add/remove traces bug

### LINKED ISSUE #2602: Error when adding a new trace to a plot with Plotly.react
After updating to plotly.js 1.37.0, I'm getting the following error when adding a trace to a plot that already has at least one trace. This then seems to sometimes cause the trace to not render.

```
Error while plotting: TypeError: Cannot read property '0' of undefined
    at Object.plots.supplyDefaultsUpdateCalc (plotly.js:143231)
    at Object.exports.react (plotly.js:123205)
    at factory.js:100

```
Recreated with codepen: https://codepen.io/steph_curried/pen/gzWBQE
I had to use chrome dev tools console, codepens console does not show the error message.

alexcjohnson:
Fixes #2602 - not sure how this worked before #2577 but it must have. And I'm kind of amazed that we didn't have an add/remove traces test for `Plotly.react` before... oh well, now we do!

cc @etpinard

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2372,7 +2372,7 @@ function diffData(gd, oldFullData, newFullData, immutable) {
     if(oldFullData.length !== newFullData.length) {
         return {
             fullReplot: true,
-            clearCalc: true
+            calc: true
         };
     }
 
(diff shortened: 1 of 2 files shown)


etpinard:
Nicely done :dancer:

Issue #2082, opened 2017-10-12
Add an event for panning and zooming that show live details as the range changes


Holyphoenix:
It would be useful to have an event that is fired any time the range of a chart changes (not just after the pan or zoom is complete).  This event could be used to make a second chart scroll or zoom with a first chart.

I want to use it to make a chart with the x-axis scrollable on my page over my other, quite large chart, that has many subplots.  I could also see it being useful for things like a dashboard where charts are separated to a point where they cannot just be subplots, but you want live updates to the charts as other charts are changing.

alexcjohnson:
Good idea, similar to the `selecting` vs `selected` events. I imagine it would be very easy to slow the system down unbearably this way (panning & live zooming take shortcuts to adjust the plots much faster than a normal `relayout` with a changed range) but there are cases where this would work fine, or other things you might want to do with those ranges besides updating other plots.

etpinard:
cc https://github.com/plotly/plotly.js/issues/145

zhihua-chen:
Would like the same for rotating 3d plots.

zhihua-chen:
How about this PR for gl3d plots?

https://github.com/zhihua-chen/plotly.js/pull/1

suriyaprashath:
Any update on this feature request?

zhihua-chen:
Sorry haven’t gotten around to writing the tests yet. Busy with work...

etpinard:
Adding to `v1.48.0`

PR #2606, opened 2018-05-03
Emit `plotly_relayouting` event during drag interactions

zhihua-chen:
Currently a gl3d plot does not emit 'plotly_relayout' while dragging is happening. The event is not emitted until 'mouseup'. This PR will let a gl3d emit the 'plotly_relayout' event while the mouse is moving.


zhihua-chen:
@etpinard Thanks for the feedback! When you said 3D drag is "already inconsistent", were you referring to 3D zooming which does emit 'plotly_relayout' events as the zooming is happening?

Haven't really studied the performance issue. FYI, in my situation, I'm trying to create a synchronized set of 3d scatter plots. Trying to make all of them rotate in real time as you drag on one of them... So far things are pretty smooth with a set of 3 synchronized plots with ~4500 points in each.

alexcjohnson:
As you point out @zhihua-chen `wheel` is already emitting a `relayout` event for every DOM event... which is different from 2D where we [wait](https://github.com/plotly/plotly.js/blob/398eeb3e5f3ef697609235d47d7ad78cada79cac/src/plots/cartesian/dragbox.js#L451-L459) a somewhat arbitrary [delay time](https://github.com/plotly/plotly.js/blob/398eeb3e5f3ef697609235d47d7ad78cada79cac/src/plots/cartesian/constants.js#L51) after the last `wheel` event we see before firing the `relayout`. We really should make 3D work like 2D here, at least once we have these intermediate events implemented.

I can't imagine that simply emitting the event would impact performance noticeably, right? It's how the developer responds to that event that matters, but I think we should let them decide how much performance hit is acceptable in their own application. Likewise I don't think we should throttle these events, let developers do that themselves if it suits their application - though at some point perhaps it would be neat to build that into our events API, like `gd.onThrottled(event, callback, minDelay)`?

Can we do this in a way that explicitly mirrors `relayout` so that a) it's clear what the event data structure will be, and b) it will apply not just to drags but wheel events as well? That could be `plotly_relayouting`, or `plotly_relayoutpending`, something like that?

etpinard:
> That could be plotly_relayouting

I'm a fan of this. This mirrors `plotly_selecting` for selection.

@zhihua-chen would you be interested in adding `plotly_relayouting` to our other subplot types?

zhihua-chen:
@etpinard Sure. Do you have a list of the plot types so I don’t leave anything out?

etpinard:
@zhihua-chen sure!

- 3D (just make sure `plotly_relayouting` is emitted not `plotly_relayout` on mousemove)
- Cartesian (`plotly_relayouting` should be emitted from [`zoomMove`](https://github.com/plotly/plotly.js/blob/dfb56d3a7b99ecf979365ed9b9c2c40747cdd915/src/plots/cartesian/dragbox.js#L281-L353) and [`plotDrag`](https://github.com/plotly/plotly.js/blob/dfb56d3a7b99ecf979365ed9b9c2c40747cdd915/src/plots/cartesian/dragbox.js#L481-L564))
- Polar (in [`zoomMove`](https://github.com/plotly/plotly.js/blob/dfb56d3a7b99ecf979365ed9b9c2c40747cdd915/src/plots/polar/polar.js#L641-L684))
- Ternary (in [`zoomMove`](https://github.com/plotly/plotly.js/blob/master/src/plots/geo/zoom.js) and [`plotDrag`](https://github.com/plotly/plotly.js/blob/master/src/plots/geo/zoom.js))
- Geo (in the various `handleZoom` handlers of [`geo/zoom.js`](https://github.com/plotly/plotly.js/blob/master/src/plots/geo/zoom.js))
- Mapbox (in the mousemove [handler](https://github.com/plotly/plotly.js/blob/dfb56d3a7b99ecf979365ed9b9c2c40747cdd915/src/plots/mapbox/mapbox.js#L159-L172))

zhihua-chen:
Writing down some notes and questions as I go through the code.

What payload should be emitted with these events?

With gl3d, this was easy, just emit the updated camera location. 

What about other chart types? Should the event payloads for “relayouting” be standardized somehow?

People must have thought about this when they decided to emit the “relayout” event. Are payloads for this event already consistent across chart types?

Is there a place to document these events?

etpinard:
> What about other chart types? Should the event payloads for “relayouting” be standardized somehow?

Easy answer. The payload should be the same as the associated `plotly_relayout` event.

etpinard:
> Is there a place to document these events?

We have https://plot.ly/javascript/plotlyjs-events/, but you don't need to worry about it.

zhihua-chen:
When should the event be `plotly_relayouting` and when should it simply be `plotly_relayout`?

For zooming in cartesian plots, it makes sense to have a relayouting event in `zoomMove` for intermediate drag action and a relayout for when the drag action ends and plot gets resized.

For zoomwheel and panning in cartesian plots, there are no intermediate steps. Once the you move the mouse or scroll the wheel, the plot is redrawn immediately (a small delay is set in the case of zoomwheel.)

Currently, zoomwheel emits a `plotly_relayout` in `dragTail` for each wheel scroll. This is great. Should we make panning behave the same way: emitting a 'plotly_relayout` every time the mouse moves?

alexcjohnson:
> When should the event be `plotly_relayouting` and when should it simply be `plotly_relayout`?

Every time we actually alter `gd.layout` (via a `Plotly.relayout` call, or directly if we've short-circuited that) there should be an associated `plotly_relayout` event. If we've only altered the appearance on-screen but haven't propagated the change to `gd.layout` yet, it should be a `plotly_relayouting` event instead. So with that in mind:

> Currently, zoomwheel emits a `plotly_relayout` in `dragTail` for each wheel scroll.

`dragTail` is what actually updates `gd.layout`, so that's the correct place to emit `plotly_relayout`. But in fact we update the viewbox (on-screen appearance) in principle many times before we get to `dragTail` (notice that `zoomWheel` calls [`clearTimeout(redrawTimer);`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/dragbox.js#L416)) so it would be reasonable to add a `plotly_relayouting` event inside `zoomWheel`.

> Should we make panning behave the same way: emitting a `plotly_relayout` every time the mouse moves?

[panning](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/dragbox.js#L485) works the same as zoom - on `mousemove` we just change the viewbox but do not alter `gd.layout` and redraw properly until `mouseup`. So like zoom, mousemove should have a `plotly_relayouting` event, not `plotly_relayout`.

zhihua-chen:
Thanks for the detailed response @alexcjohnson, that is really helpful.

Please forgive this newbie for some more questions.


Question 1: why the multiple layers of drawing action? What determines what gets redrawn during the interaction and what gets redrawn at the relayout?

Related question 2: zoomWheel employs a debounce, why doesn’t panning? gl3d interactions do not use a debounce right now, do they need one?

zhihua-chen:
All right. I have added layouting events to all chart types @etpinard listed.

etpinard:
> All right. I have added layouting events to all chart types @etpinard listed.

Nice job @zhihua-chen . Would you be interesting in adding test cases for each of these new event triggers?

etpinard:
@zhihua-chen are you still interested in trying to bring this PR to the finish line?

If not, I might have time to add a few commits in time for 1.47.0.

zhihua-chen:
Sorry. Please go ahead and do whatever you would like. I’m too distracted
at this moment...

Zhihua

archmoj:
Solving merge conflicts for this PR is rather risky considering the changes since last year. I suggest we may close this PR and open a feature request to possibly address `relayouting` for gl3d and other traces. @etpinard what do you think?

etpinard:
:+1: for opening a new PR. I would wait until this new PR exists before closing this one here, to keep an active placeholder.

etpinard:
Closing. I just realised there's an issue opened for this already -> https://github.com/plotly/plotly.js/issues/2082

PR #2607, opened 2018-05-03
Override toImage modebar button opts via config

nicolaskruchten:
This PR allows for selective overriding of default behaviour of the `toImage` modebar button. Specifying nothing falls through to the current defaults. Note sure about the implications of renaming the button, wrt to locale files.

### DIFF
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -47,19 +47,29 @@ var modeBarButtons = module.exports = {};
 
 modeBarButtons.toImage = {
     name: 'toImage',
-    title: function(gd) { return _(gd, 'Download plot as a png'); },
+    title: function(gd) { return _(gd, 'Download plot'); },
     icon: Icons.camera,
     click: function(gd) {
-        var format = 'png';
+        var toImageButtonDefaults = gd._context.toImageButtonDefaults;
+        var opts = {format: toImageButtonDefaults.format || 'png'};
 
         Lib.notifier(_(gd, 'Taking snapshot - this may take a few seconds'), 'long');
 
         if(Lib.isIE()) {
             Lib.notifier(_(gd, 'IE only supports svg.  Changing format to svg.'), 'long');
-            format = 'svg';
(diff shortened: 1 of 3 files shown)


etpinard:
Nicely done. :dancer:

src/components/modebar/buttons.js
```diff
@@ -47,19 +47,26 @@ var modeBarButtons = module.exports = {};
...
-        var format = 'png';
+        var toImageButtonDefaults = gd._context.toImageButtonDefaults;
+        var opts = {format: toImageButtonDefaults.format || 'png'};
 
         Lib.notifier(_(gd, 'Taking snapshot - this may take a few seconds'), 'long');
 
         if(Lib.isIE()) {
             Lib.notifier(_(gd, 'IE only supports svg.  Changing format to svg.'), 'long');
-            format = 'svg';
+            opts.format = 'svg';
```
alexcjohnson: also we should change the conditional to `if(Lib.isIE() && opts.format !== 'svg')`

src/components/modebar/buttons.js
```diff
@@ -47,19 +47,26 @@ var modeBarButtons = module.exports = {};
 
 modeBarButtons.toImage = {
     name: 'toImage',
-    title: function(gd) { return _(gd, 'Download plot as a png'); },
+    title: function(gd) { return _(gd, 'Download plot'); },
```
alexcjohnson: We can't make dynamic strings here, so `_(gd, 'Download plot as a ' + format)` won't work, but perhaps we can keep `'Download plot as a png'` as the default string and only degrade to `'Download plot'` if it's not a png? Two reasons:
- I don't want to unnecessarily break all our existing translations, for folks who are *not* using this new option.
- I do find "as a png" helpful, to clarify that it's not the data or json for example, though I guess the icon mitigates that substantially.

nicolaskruchten:
Thanks @alexcjohnson ! I've implemented your suggestions

alexcjohnson:
Thanks! 💃

Issue #2600, opened 2018-05-02
scattergl calculates wrong error bar value if y value is a string


slioxo:
I stumbled upon this problem today. As you can see [here](https://codepen.io/anon/pen/deWjJE), when the y-value is given as string and the error value as float, the error value is presumably parsed as string and then added using the "+" operator, which results in string concatenation and therefore a wrong value.

For example, if your y-value is 2 and the error value is 2 then the error bar should be drawn up to 4. When y is a string though, the error bar is drawn up to 22.

When using 'scatter' instead of 'scattergl', this bug does not occur.

alexcjohnson:
Also it gets autorange even *more* wrong, and the errorbars disappear on redraw (the latter also reported in #2450)

Thanks for the report!

Issue #2450, opened 2018-03-07
scattergl layering and errorbar issues


alexcjohnson:
A couple of remaining scatter/scattergl differences, see https://codepen.io/alexcjohnson/pen/YemVOd
![screen shot 2018-03-06 at 6 53 21 pm](https://user-images.githubusercontent.com/2678795/37065664-b3fad89e-216f-11e8-861a-3028adc470a6.png)
- [ ] fill tozero breaks in some weird ways - here it's filling to the wrong final y value, but if you slide `trace1.y[0]` up and down a little bit you see all sorts of odd behavior.
- [x] fills are behind everything else, they should be on top of some of the lines/points/errorbars (note, I didn't include fill tonext in this example, but it makes fill layering even more complicated, see https://github.com/plotly/plotly.js/pull/2442#discussion_r172405040) - **Fixed in** -> https://github.com/plotly/plotly.js/pull/3087
- [x] x errorbars aren't supported? **done ->** #2620

Initially I thought errorbar colors were also a scattergl issue, but scatter also doesn't get the colors I would have expected. That may have to be a v2 change at this point...

cc @dfcreative @etpinard

tgoldina:
One more issue with error bars in scattergl, "markers" mode: error bars disappear on any relayout action: try switching to box select, zooming or panning in https://codepen.io/anon/pen/WJxKpP

etpinard:
thanks for the report @tgoldina

etpinard:
now in https://github.com/plotly/plotly.js/issues/4017

PR #2620, opened 2018-05-08
scattergl errorbars fixes

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2600 and the last item off https://github.com/plotly/plotly.js/issues/2450#issue-302915305 (_x errorbars aren't supported?_). 

Does not attempt to fix the on-relayout issue of https://github.com/plotly/plotly.js/issues/2450#issuecomment-384363970 nor the incorrect errorbar layering order described in https://github.com/plotly/plotly.js/issues/2450, @dy will take care of them once he finishes his `regl-component` module.

cc @alexcjohnson


src/traces/scattergl/convert.js
```diff
@@ -360,10 +360,11 @@ function convertLinePositions(gd, trace, positions) {
     };
 }
 
-function convertErrorBarPositions(gd, trace, positions) {
+function convertErrorBarPositions(gd, trace, pos, x, y) {
     var calcFromTrace = Registry.getComponentMethod('errorbars', 'calcFromTrace');
-    var vals = calcFromTrace(trace, gd._fullLayout);
-    var count = positions.length / 2;
+    var _trace = Lib.extendFlat({}, trace, {x: x, y: y});
```
alexcjohnson: This seems weird to me - putting arrays of `calcdata` values into a fake trace. I don't know if category or date data would make any sense in this context but if we *were* to support them this would break.

This is the only caller of `calcFromTrace`, right? Can we just pass in the x and y calc'ed arrays separately from the trace?

src/traces/scattergl/convert.js
```diff
@@ -360,42 +362,60 @@ function convertLinePositions(gd, trace, positions) {
...
-        }
-
-        return errors;
-    }
+    function convertOneAxis(coords, ax) {
+        var axLetter = ax._id.charAt(0);
+        var opts = trace['error_' + axLetter];
+
+        if(opts && opts.visible && (ax.type === 'linear' || ax.type === 'log')) {
+            var computeError = makeComputeError(trace['error_' + axLetter]);
```
alexcjohnson: `makeComputeError(opts)`?

src/traces/scattergl/convert.js
alexcjohnson: Ah nice - that's more of a refactor than I was expecting but you're right, much better. 🌴 before, but as well as the 🐎 overhead, the scaffolding meant it didn't even really result in shorter code.

alexcjohnson:
Very nice! 💃 after the minor comment above.

Issue #2501, opened 2018-03-27
mode:text for regl-scatter2d


jackparmer:
I have a few bio customers who need to display hundreds to thousands of single letter markers like in the below graph:

![image](https://user-images.githubusercontent.com/1865834/37943372-50436efc-3145-11e8-945a-6a7290cc3365.png)

@dfcreative already has a great example of single-letter rendering for regl-scatter2d here:
https://dfcreative.github.io/regl-scatter2d/copyrights.html

I'd like to propose that `mode: text` be enabled for `scattergl`, but only work for single letters (for now). 

Note that `mode: text` currently does not do anything for `scattergl`:  https://codepen.io/plotly/pen/dmZNjM

We've tried creating graphs like the above with SVG, but for a larger number of letters, they are too slow. Here is an SVG example with a `mode: text` `scatter` trace overlaid on a `heatmap` trace: https://plot.ly/~jackp/18395/ (Note how long it takes to update when dragging the slider).

etpinard:
I have my doubt about the single-letter approach. If we do so, then we would stuck with that until we make breaking API changes. @dfcreative How hard would it be to support text of all lengths?

And by the way @jackparmer about https://plot.ly/~jackp/18395/ are you unhappy about the first-render performance or just the drag hangups (or both :smirk: )? Simply improvement drag performance for SVG text nodes shouldn't be too hard.

jackparmer:
> are you unhappy about the first-render performance or just the drag hangups (or both 😏 )

Mostly the drag performance.

> How hard would it be to support text of all lengths?

Dima told me up to 2 weeks.

dfcreative:
@etpinard depends on the requirements. There are many ways to do that, the simplest is canvas2d texture, that may take a day or two. [Font-texture](https://webglfundamentals.org/webgl/lessons/webgl-text-glyphs.html) or something with sdf would take a week or two. Same for [vector-based](http://wdobbie.com/post/gpu-text-rendering-with-vector-textures/) approach.

jackparmer:
I think we should do the `sdf` method after `splom` if we want multi-character marker text in `scattergl`. No point in adding `mode: text` to `scattergl` if it's not going to be significantly faster than SVG `scatter`.

etpinard:
Related: https://github.com/plotly/plotly.js/issues/1116

etpinard:
done in https://github.com/plotly/plotly.js/pull/2737, followup in https://github.com/plotly/plotly.js/issues/2780

PR #2623, opened 2018-05-09
Faster svg pan

### LINKED ISSUE #2548: Stash SVG points selections for faster pan

etpinard:
resolves #2548 (there's still room for improvement, but it's enough to close that ticket for now I think).

In brief this PR:
- makes pan/scroll go through the "inverse scale" per-point loop only when needed. This loop updates the `transform` attr of every point on every mousemove to counteract the scale of trace group during pan/scroll. It turns out that can be extremely costly on graphs with many pts and even more costly for text nodes.
- stashes "hot" svg selections that are used during pan and scrolled to update the view w/o going through a relayout. Note that `plots/cartersian/dragbox.js` is now free of (potentially) slow `selectAll` calls!
- updates optimization flags `hasSVG`, `hasScattergl` and friends only on mousedown instead of on every mousemove.

Taking https://plot.ly/~jackp/18395.embed to benchmark things:

before each mousedown interval took around 30-40ms:

peek 2018-05-09 16-52

off this branch we're down in around 1ms :

peek 2018-05-09 16-54

Note that this PR improves scroll performance a little as stashing shaves about 5ms per scroll interval, but is still pretty slow (we might have to wait for scattergl text there https://github.com/plotly/plotly.js/issues/2501).

cc @alexcjohnson @jackparmer

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -1057,33 +1057,24 @@ drawing.setScale = function(element, x, y) {
     return transform;
 };
 
-drawing.setPointGroupScale = function(selection, x, y) {
-    var t, scale, re;
+var SCALE_RE = /\s*sc.*/;
 
-    x = x || 1;
-    y = y || 1;
+drawing.setPointGroupScale = function(selection, xScale, yScale) {
+    xScale = xScale || 1;
+    yScale = yScale || 1;
 
-    if(x === 1 && y === 1) {
-        scale = '';
-    } else {
-        // The same scale transform for every point:
-        scale = ' scale(' + x + ',' + y + ')';
-    }
 
-    // A regex to strip any existing scale:
-    re = /\s*sc.*/;
+    // The same scale transform for every point:
+    var scale = (xScale === 1 && yScale === 1) ?
+        '' :
+        ' scale(' + xScale + ',' + yScale + ')';
 
     selection.each(function() {
-        // Get the transform:
-        t = (this.getAttribute('transform') || '').replace(re, '');
+        var t = (this.getAttribute('transform') || '').replace(SCALE_RE, '');
         t += scale;
         t = t.trim();
-
-        // Append the scale transform
         this.setAttribute('transform', t);
     });
-
-    return scale;
 };
 
 var TEXT_POINT_LAST_TRANSLATION_RE = /translate\([^)]*\)\s*$/;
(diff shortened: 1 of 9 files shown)


src/components/drawing/index.js
```diff
@@ -1057,38 +1056,32 @@ drawing.setScale = function(element, x, y) {
...
+    var scale = (xScale === 1 && yScale === 1) ?
+        '' :
+        ' scale(' + xScale + ',' + yScale + ')';
 
     selection.each(function() {
-        // Get the transform:
-        t = (this.getAttribute('transform') || '').replace(re, '');
+        var t = (this.getAttribute('transform') || '').replace(SCALE_RE, '');
         t += scale;
         t = t.trim();
```
alexcjohnson: I don't know whether the string manipulations here are a meaningful contribution to the performance... but this could all be simplified a bit (`indexOf` & `.substr` instead of a regex & `.replace`, no `.trim`) if we could omit the space before the word `scale`. It seems to work fine in my browsers, though in principle it's not supposed to be allowed: ["The individual transform definitions are separated by whitespace and/or a comma."](https://www.w3.org/TR/SVG/coords.html#TransformAttribute)

Alternatively, we could imagine stashing the pre-scaled transform as perhaps a `data-initialtransform` attribute on each element and avoid all this processing...

Not necessary, just a possibility if you think there would be a benefit.

alexcjohnson:
@etpinard looks great! Just the one nonblocking comment, lets do it. 💃

src/components/drawing/index.js
etpinard: I haven't benchmarked this particular block yet, but I know that `Drawing.setTextPointsScale` is 4-5 times slower than this routine here `Drawing.setPointGroupScale`. 

Spending time on making these routines faster would help improve scroll performance. So yeah, thanks for writing this down. I just opened a new ticket about this -> https://github.com/plotly/plotly.js/issues/2625

Issue #2595, opened 2018-05-01
Unnecessary clipPaths


alexcjohnson:
Potential 🐎 issue with many-subplot graphs: seems like we’re making `clipPath.axesclip` for every possible combination of axes… even though `clipPath.plotclip` only gets the subplots we use. Dunno if maybe shapes can require other combinations? In any event we should be able to be more discriminating in which of these we make.
![screen shot 2018-05-01 at 9 23 45 am](https://user-images.githubusercontent.com/2678795/39475821-dbdf6576-4d27-11e8-97a2-1606fc094b5e.png)

etpinard:
These "unnecessary" clip paths are added to the DOM in `Axes.makeClipPaths` 

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/plots/cartesian/axes.js#L1457-L1495

Some benchmarks using https://codepen.io/etpinard/pen/wjmqmO:, at 50 dimensions, `Axes.makeClipPaths` takes around ~50ms, at 20 dimensions we're down to ~8ms - which is non-negligeable but not a major perf hit. Note that `makeClipPaths` is only called during the `lsInner` subroutines which is skipped notably during "axrange" and "modebar" editTypes.

That said, as far as I can tell, these per-axes clip paths are only used for shapes and layout images. So, we could easily skip `Axes.makeClipPaths` when the `_hasOnlyLargeSplom` flag

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/plots/plots.js#L431-L440

is turned on. Alternatively, we could call `Axes.makeClipPaths` only when there are visible shapes or layout images on the graph. This could make some relayout logic trickier though.

alexcjohnson:
Good to know, thanks for benchmarking it. Sounds like adding a `_hasOnlyLargeSplom` switch is a no-brainer.

Down the line though we should be able to do even better by only making the ones we need period, by building up the required list during `supplyDefaults` and using that during `makeClipPaths` instead of the maximal list we generate there now. Changing axis references for shapes is already `editType: 'calc'`, we could do that with images as well (I don't see much reason to make those operations fast, if we ever autorange images #1111 we'd probably need to do that anyway), that seems like it would solve the `relayout` concerns.

etpinard:
> Sounds like adding a _hasOnlyLargeSplom switch is a no-brainer.

[...]

Issue #2549, opened 2018-04-11
Speed up cartesian axes supplyDefaults


etpinard:
From https://github.com/plotly/plotly.js/pull/2527, Plots.supplyDefaults takes > 100ms at 50x50 subplots which feels like an eternity when zooming.

- Getting rid of `indexOf` into axis lists would be a good start (perhaps experimenting with ES6 `Map` and `Set` could be worthwhile)
- Perhaps we could try opening a partial `Plots.supplyDefaults` pathway, that would only update a subset of the graph's cartesian axes and their dependencies.

etpinard:
> So I'm thinking of moving this block out of Plots.linkSubplots and into the lsInner subroutine

Done in https://github.com/plotly/plotly.js/pull/2628. 

PR https://github.com/plotly/plotly.js/pull/2628 also added set axis anchors to all sploms except the ones with `showlowerhalf: false` so that the "find main subplot" loop is skipped when we can.

etpinard:
From https://github.com/plotly/plotly.js/pull/3070#issuecomment-426792309, 50 dims splom traces spend about 40ms in [`axis_autotype.js`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axis_autotype.js). Improving perf there could be very beneficial.

etpinard:
Referencing https://github.com/plotly/plotly.js/issues/749 - moving away from the "relink everything approach" should speed up `supplyDefaults` in some scenarios.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #2547, opened 2018-04-11
Per axis `axrange` edit type


etpinard:
PR https://github.com/plotly/plotly.js/pull/2527 added `editType: 'axrange'` which improves cartesian axis range relayout performance. But we could do even better if made 'axrange' call `Axes.doTicks` only on the required axes instead of clearing them and starting from scratch. Improvements would be most noticeable in multi-subplot layouts. The relayout logic already [tracks altered axes](https://github.com/plotly/plotly.js/blob/2a2582022105abd640ef3e5a38d98fb343a06779/src/plot_api/plot_api.js#L1853-L1857), so in theory, we could pass that info to 'axrange' update sequence.

Related: https://github.com/plotly/plotly.js/issues/298

etpinard:
Pasting the note from https://github.com/plotly/plotly.js/pull/2628#discussion_r187696239:

N.B. @alexcjohnson I didn't implement per-axis "axrange" edit for `Plotly.react`. My main concern was improvement perf on dragend axis range relayout calls. I didn't think adding an exception to the react diffing algo was worth it for this case.

in case we want to improve `Plotly.react` speed down the road.

Issue #2540, opened 2018-04-11
Investigate relayout & react + axis constraints behavior


etpinard:
In PR https://github.com/plotly/plotly.js/pull/2527, a few oddities with relayout on graphs with axis constraints (i.e. `(x|y)axis.constrain: 'range' || 'domain'`) were detected. 

From https://github.com/plotly/plotly.js/pull/2527#discussion_r180543504 and below, tests coverage was deemed inadequate to assume anything about the current behavior. The first step would be to beef up the coverage there. Next, determine whether or not graphs with axis constraints do need to go though the editType `calc` on axis range relayouts. Finally, we should lock down the limitations (if any) of editType `axrange` for graphs with axis constraints.

cc @alexcjohnson

etpinard:
Found a bug in https://codepen.io/etpinard/pen/qYopzp 

![peek 2018-05-10 12-26](https://user-images.githubusercontent.com/6675409/39880977-5aa565be-544d-11e8-8386-f15d94eee9ad.gif)

introduced in `1.36.0` (probably from PR https://github.com/plotly/plotly.js/pull/2527) where the y-axis is getting re-position properly. This should be a good starting point for this ticket here.

etpinard:
PR https://github.com/plotly/plotly.js/pull/2628 fixed the :beetle: reported above in https://github.com/plotly/plotly.js/issues/2540#issuecomment-388106902, but many issues remain.

Pasting https://github.com/plotly/plotly.js/pull/2628#issuecomment-388446268 here:

A few more comments on constrained axes, that should be followed up in https://github.com/plotly/plotly.js/issues/2540:

- Could making range relayout go through editType "plot" (instead of "calc") be sufficient? When replacing this line

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/plot_api/plot_api.js#L2039

with `flags.plot = true`, no tests are failing. 

- I think `Plotly.react` is broken with constrained axes. I'm not sure if it's always been broken or if the recently-added "axrange" edit did it :confused: 

and @alexcjohnson 's https://github.com/plotly/plotly.js/pull/2628#issuecomment-388458992

> Could making range relayout go through editType "plot" (instead of "calc") be sufficient?

[...]

PR #2628, opened 2018-05-11
Fix relayout constrained axes + various multi-subplot perf improvements

etpinard:
This PR started with a fix for relayout on constrained axes https://github.com/plotly/plotly.js/issues/2540#issuecomment-388106902. This fix made the `rangesAltered` stash in relayout

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/plot_api/plot_api.js#L2028-L2047

work properly and lead me to use the same stash to implement per-axis "axrange" edits (as planned in https://github.com/plotly/plotly.js/issues/2547). While at it, I added a few more splom-related perf improvements for multi-subplot graphs, namely

-  skip `Axes.makeClipPaths` under `hasOnlyLargeSploms` planned in https://github.com/plotly/plotly.js/issues/2595#issuecomment-388123728
- move costly "find main subplot" for all anchor-less axes loop out of `supplyDefaults` -> `lsInner` as planned in https://github.com/plotly/plotly.js/issues/2549#issuecomment-382754397
- made splom generated axes set an anchor axis so that that costly "find main subplot" loop is skipped except when the lower half is omitted.

--------------

Some benchmarks:

- Implementing per-axis "axrange" relayout edits trims ~200ms on e.g `Plotly.relayout(gd, 'xaxis.range[0], 0)` on a 50-dimensions splom. At 15 dimensions (i.e. with svg grid lines), it trims ~100ms.
- Making all sploms except those with `showlowerhalf: false` have a set anchor trims 40-50ms.

So, starting with https://codepen.io/etpinard/pen/wjmqmO and benchmarking  `Plotly.relayout(gd, 'xaxis.range', [0, 1])` we get:

| # dims | # values per dim | results w/ 1.37.1 [ms] | results w/ this branch [ms] |
| ---------| --------------------- | -------------------------- | ------------------------------- |
| 15 | 100 | 160 | 40 | 
| 30 | 100 | 220 | 130 |
| 50 | 100 | 500  | 250 |


src/plot_api/plot_api.js
```diff
@@ -2171,7 +2169,9 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
         if(relayoutFlags.layoutstyle) seq.push(subroutines.layoutStyles);
         if(relayoutFlags.axrange) {
             seq.push(
-                subroutines.doTicksRelayout,
+                function(gd) {
+                    return subroutines.doTicksRelayout(gd, relayoutSpecs.rangesAltered);
```
etpinard: N.B. @alexcjohnson I didn't implement per-axis "axrange" edit for `Plotly.react`. My main concern was improvement perf on dragend axis range relayout calls. I didn't adding an exception to the react diffing algo was worth it for this case.

src/plot_api/subroutines.js
```diff
@@ -324,6 +329,48 @@ exports.lsInner = function(gd) {
     return gd._promises.length && Promise.all(gd._promises);
 };
 
+function findMainSubplot(ax, fullLayout) {
```
etpinard: @alexcjohnson I didn't bother the "find main subplot" algo itself as you proposed in https://github.com/plotly/plotly.js/issues/2549#issuecomment-382775857, I was happy enough with skipping this whole thing except when sploms with `showlowerhalf: false` with https://github.com/plotly/plotly.js/pull/2628/commits/c947b2c0070833199ceff56fc4bafc59d539f3d1

alexcjohnson:
> Could making range relayout go through editType "plot" (instead of "calc") be sufficient?

Do we have a test of setting mutually-exclusive range values, as per the [comment on that line](https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/plot_api/plot_api.js#L2034-L2038)? If we do, and the result correctly fixes them for the constraints, then go for it!

src/plots/cartesian/axes.js
```diff
@@ -1498,28 +1498,50 @@ axes.makeClipPaths = function(gd) {
...
+ * Signature 3: Axes.doTicks(gd, 'redraw')
+ *   use this to clear and redraw all axes on graph
+ *
+ * Signature 4: Axes.doTicks(gd, '')
+ *   use this to draw all axes on graph w/o the selectAll().remove()
+ *   of the 'redraw' signature
+ *
+ * Signature 5: Axes.doTicks(gd, [axId, axId2, ...])
+ *   where the items are axis id string,
+ *   use this to update multiple axes in one call
```
alexcjohnson: This was already a bit out of control, do we really want to add another signature?

Can we perhaps break it into 2 separate functions, one for signatures 1 & 2 and a second for 3, 4, 5? That's how it functions anyway - 3,4,5 just map into a bunch of calls to signature 2 (ignoring `skipTitle`) and then return.

etpinard: Sure that would be nice. But since `doTicks` is one of our most referenced function in existing issue, I'm a little reluctant to do so. 

But if you insist, than I'm thinking of making `Axes.doTicks` the multi-cartesian-axis version (that accepts `''`, `'redraw'` and arrays of axis ids), and adding `Axes.doTicksSingle` (that accepts one axis id string or an axis object). 

On `master`, we have:

image

alexcjohnson:
💪 Well done, this is a really fantastic collection of boosts and fixes! Thanks for the `Axes.doTicks` separation, a nice first step toward more comprehensible `axes.js`.

https://github.com/plotly/plotly.js/pull/2628#issuecomment-388458992 is non-blocking.
💃

etpinard:
> #2628 (comment) is non-blocking.

Moved to https://github.com/plotly/plotly.js/issues/2540#issuecomment-388830438

Issue #2480, opened 2018-03-19
Sankey nodes overlap by default


rahulsindc:
Hi there,

As you can see [here](https://rahulsindc.github.io/BF-visualisations/funding.sankey.html), a relatively simply sankey will often end up with nodes overlapping with other nodes, and the outgoing flows stacked as well. The code generating this diagram can be found [here](https://github.com/rahulsindc/BF-visualisations/blob/master/funding.sankey.html)

Is there a reason this is happening?

Thanks!
R

etpinard:
Thanks for the report.

rahulsindc:
Update: It appears to be browser-specific; Chrome 65 has this issue, but the diagram renders property on Safari 11 (both on Mac OS 10.13 High Sierra)

alexcjohnson:
And yet a different result in FF (image is Chrome / FF / Safari):
![screen shot 2018-04-29 at 1 18 54 pm](https://user-images.githubusercontent.com/2678795/39409159-11f51828-4bb0-11e8-810b-d5c6a6445929.png)

Here it is as a pared-down codepen https://codepen.io/alexcjohnson/pen/NMdMbE?editors=1010
Even in Safari it's not quite right - there is no overlap at least, but there's a net upward shift from left to right that should not be there.

alexcjohnson:
Update: the issue here is with a link with `value: 0` (element 5 in this graph's `link.value` array, that leads to node 15 having zero size). If I replace that with anything positive the layout is fixed. The zero results in a `NaN` in the node list, and I suspect the breakage and the browser-dependence of the problem have to do with strange handling of `NaN` in `Array.sort` or the like.

Obviously we need to prevent a zero (or otherwise invalid) `value` from breaking things, but the question then is, what *should* this do? Should we delete it completely? Draw it as a very thin node & link (labeled as size zero)?

Negative or non-numeric values I feel confident we should ignore, but I feel like there may be cases you still want a zero-size node to show up. Thoughts?

PR #2629, opened 2018-05-11
More robust Sankey

alexcjohnson:
Make Sankey more accepting of messy data (fixing #2480)
- Numeric strings are OK
- Zero (and negative) link values get ignored completely. Note: this is not what I proposed in https://github.com/plotly/plotly.js/issues/2480#issuecomment-387201484 - thinking a bit more about it, links with no value should really be not links at all, and nodes with no size shouldn't be nodes at all. If you really want a node & link to show up with zero value, create one with an explicitly tiny value.
- Unused nodes and malformed links are ignored as well. Since there are now various places this can happen and in other contexts we simply ignore partially-invalid data points, I also removed the logged error on missing nodes - which happened during `supplyDefaults`, so was called frequently, and had a fairly slow implementation so could be a drag on large Sankey traces.

cc @etpinard


src/plot_api/plot_schema.js
```diff
@@ -437,6 +437,8 @@ function recurseIntoValObject(valObject, parts, i) {
     return valObject;
 }
 
+// note: this is different from Lib.isIndex, this one doesn't accept numeric
+// strings, only actual numbers.
```
alexcjohnson: I initially made `isIndex` into a `Lib` function thinking I could reuse it here... but in this context it's in principle important that an index is actually a number, to distinguish an array index from an object property that happens to be a number. That said I don't think it's possible to use a number as an object property in `nestedProperty`...

src/traces/sankey/render.js
```diff
@@ -63,44 +66,79 @@ function switchToSankeyFormat(nodes) {
...
+        else removedNodes = true;
+    }
+
+    // need to re-index links now, since we didn't put all the nodes in
+    if(removedNodes) {
+        for(i = 0; i < links.length; i++) {
+            links[i].source = nodeIndices[links[i].source];
+            links[i].target = nodeIndices[links[i].target];
+        }
+    }
```
alexcjohnson: For the most part it didn't cause problems to have extra unused nodes in the list, but sometimes the layout algorithm would reserve extra space for these nodes (leaving extra margins or making the whole diagram S-shaped, for example) so it's better to remove them. The `sankey_messy` mock includes a whole bunch of extra nodes both before (to test this link reassignment) and after the real ones to test this.

src/plot_api/plot_schema.js
etpinard: Oh we could use this for validating `selectedpoints` items as done here:

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/lib/index.js#L497-L499

etpinard: ... by _this_ I mean `Lib.isIndex`

src/traces/sankey/calc.js
```diff
@@ -14,13 +14,16 @@ var wrap = require('../../lib/gup').wrap;
 
 function circularityPresent(nodeList, sources, targets) {
 
-    var nodes = nodeList.map(function() {return [];});
+    var nodes = nodeList.map(function() { return []; });
```
etpinard: Would be nice to reuse:

https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/lib/matrix.js#L13-L17

test/jasmine/tests/sankey_test.js
```diff
@@ -58,30 +58,31 @@ describe('sankey tests', function() {
...
-                    link: {
-                        value: [1],
-                        source: [0],
-                        target: [1]
-                    }
-                });
+    describe('No warnings for missing nodes', function() {
+        // we used to warn when some nodes were not used in the links
+        // not doing that anymore, it's not really consistent with
+        // the rest of our data processing.
```
etpinard: nicely done :ok_hand:

etpinard:
Great PR :dancer:

etpinard:
... my comments are non-blocking.

src/plot_api/plot_schema.js
alexcjohnson: Good call -> https://github.com/plotly/plotly.js/pull/2629/commits/cdcd6200907d5caad164933ebbf388bc6532bf19

We probably either have or *should* have this in other places too... `mesh3d` is the other obvious place, but it has bigger issues -> #2630

PR #2641, opened 2018-05-16
3D cone traces

etpinard:
Cones are coming to plotly.js

peek 2018-05-16 13-58

### Things we need to decide on

- [x] How to name this trace? I chose `cone`, like a _cone_ trace, similar to `surface`. But since a cone trace has often many cones, maybe `cones` would be better.  `cone3d` similar to `mesh3d` could work, but I think the _3d_ is not necessary, _quiver_ is more popular in 2d. **Decision:** we're going with `cone`.
- [x] How to name the meshgrid attribute? I chose `vx`, `vy` and `vz` (_v_ for vector), but maybe `meshgrid.x`, `meshgrid.y` and `meshgrid.z` would be best. Note that these attributes are optional. **Decision:** <strike>going with `cones.(x|y|z)` in cb7ef43933987885b170a8e497dee103c8cdea88 which will allows us to configure non-cartesian-product meshgrids if we want down the road.</strike> **Removed from this 1st implementation** -> 5a42de085bf5829a5108a7c8551d92a95cb48d6f
- [x] Should the colorbar show normalize vector norm (i.e. 0 to 1) values or absolute vector norms? Maybe we need an attribute to enable both behaviors? **Decision:** going with absolut vector norms in https://github.com/plotly/plotly.js/pull/2641/commits/a2db5671b9ed6401c14a2f3db7e0395bec00f10d
- [x] The color values (i.e. intensity) for the cones represent the vector norm which is positive definite. Maybe we should restrict `cmin` and `cmax` to be positive also? Moreover, maybe we should make `Viridis` the default colorscale. **Decision:** Let's **not** restrict cmin, cmax in case we add an `intensity` field to color the cones by something other than their vector norm down the road. No go for Viridis default until v2.

### Things we need to fix/add

- [x] Computed trace bounds are off. They only consider the cones' tip, ignoring their lengths and directions. This can make autoranged axes look weird:

image

**Update:** Ok, but not great solution done in https://github.com/plotly/plotly.js/pull/2641/commits/db8222b715bb9103b58576c96387970645da16c5 and improved multiple times in subsequent commits.

- [x] Need to plug in `size` settings (attributes `sizeref` and `sizemode`) - done in https://github.com/plotly/plotly.js/pull/2641/commits/863b0daabccfbc4d2ccafa0c08a35d2f2ad8dfc5
- [x] Hover doesn't work - done in 71e46af
- [x] Image generation fails in `imagetest` (works fine in orca) - **Update:** non-perfect solution in https://github.com/plotly/plotly.js/pull/2641/commits/aaf7249a775b01ac151a760f6267c1806d82f85c where `gl3d_cone-*` mocks are tests using in orca during `./tasks/noci_test.sh`.

### Things that work ok, but could improve

- [x] We currently call `cone2mesh` twice which can be slow (~300ms for 20k cones) per `.plot` call. Once in `calc` to get the colors (i.e. intensity) values and another in `convert` to compute the correctly scaled coordinate (note the the scene's scale isn't available during `calc`). I would be nice to split gl-cone3d's `createConePlot` into two subroutines. **Update:** not any more after https://github.com/plotly/plotly.js/pull/2641/commits/a2db5671b9ed6401c14a2f3db7e0395bec00f10d

### Things that could wait future iterations

- [ ] make `x`, `y` and `z` optional, generate linspaces when omitted
- [x] plug in gl-mesh3d's lighting settings - done in 3dd6cf5d98dbdb7486a93fed0a05e475c30241bd
- <strike>[ ] providing an underdetermined meshgrid can lead to exceptions in gl-cone3d's `createConePlot`</strike> not an issue until we figure how to implement our version of _meshgrid_.


alexcjohnson:
Cool!

> How to name this trace? I chose `cone`, like a _cone_ trace, similar to `surface`. But since a cone trace has often many cones, maybe `cones` would be better.  `cone3d` similar to `mesh3d` could work, but I think the _3d_ is not necessary, _quiver_ is more popular in 2d.

Agreed - a cone is a 3D shape, it would have no meaning in 2D, so just `cone` - `3d` is unnecessary. We don't pluralize any of our other types (`parcoords` looks plural, but it's "coordinates" that's plural, not the lines connecting these coordinates, which are really the objects on the plot)

> How to name the meshgrid attribute? I chose `vx`, `vy` and `vz` (_v_ for vector), but maybe `meshgrid.x`, `meshgrid.y` and `meshgrid.z` would be best. Note that these attributes are optional.

I think we should have the required data be the vector field `(x, y, z) -> (u, v, w)`, then probably we need a few ways to specify whether and how to interpolate cones onto that field (and I think this is what you're already doing, but in case it's not clear already all of these arrays are assumed to be 1D - 3D leads to too many ambiguities as well as extra headaches for our back end):
- If you add nothing else, draw cones at exactly the specified positions `(x, y, z)`
- You can provide 3 more data arrays, then the cones are interpolated onto these positions. Perhaps like MATLAB we call these `(cx, cy, cz)`? or `cones: {x, y, z}`?
- It would also be nice to be able to specify just a grid for each dimension; what should this look like? I don't think it can just be `x0, dx` etc because we need both bounds; `x0, dx, x1` would do this but we have no precedent for that. For histograms and contours we have `{start, end, size}`, so how about `cones: {xgrid: {start, end, size}, ygrid: {start, end, size}, zgrid: {start, end, size}}`? `start` and `end` could be optional, then we would just bound the grid to the data.
- In principle it seems like you could mix and match these per dimension... like you provide `cones: {x, y, zgrid}` and the result is cones explicitly positioned in `x` and `y`, then repeated for every value implied by `zgrid`. But that could get confusing, I could imagine people wanting to think of `cones.x`, `cones.y`, and `cones.z` each as an independent array, ie cone positions are the cartesian product of the three rather than element-by-element combinations. Too many possibilities, if you want to start with just the first option that's fine, we can add the others later if there's demand.

> Should the colorbar show normalize vector norm i.e. 0 to 1) values or absolute vector norms? Maybe we need an attribute to enable both behaviors?

Absolute vector norm seems far more useful. Leave this attribute out and let people normalize on their own, until someone complains. The hover data is always going to show the absolute vector anyway, right? Would be weird if this disagreed with the colorscale.

> The color values (i.e. intensity) for the cones represent the vector norm which is positive definite. Maybe we should restrict `cmin` and `cmax` to be positive also? Moreover, maybe we should make `Viridis` the default colorscale.

Much as I like `Viridis`, lets not diverge from the default colorscales we have for all other traces until we can change them as well.

Is there a use case to supply color as an independent data array? Perhaps someone wants to color just by `w` (vertical flux) or **∇∙v** or **∇×v** or who knows what... If we allow that, `cmin` and `cmax` could be anything. Anyway I don't see a need to explicitly force them to be positive even when color is the norm, they'd be positive by default and only go negative if the user explicitly chooses that.

alexcjohnson:
> Computed trace bounds are off. They only consider the cones' tip, ignoring their lengths and directions.

Are the cones positioned with their tips at the data positions? That seems odd, I'd have thought the data position would go at its center of mass, ie 3/4 of the way from the tip to the base.

etpinard:
> Are the cones positioned with their tips at the data positions?

Yes. That's how it works currently.

alexcjohnson:
Huh, MATLAB places the cones with their tails at the data positions. That also seems wrong to me...

src/traces/cone/attributes.js
```diff
@@ -80,9 +80,28 @@ var attrs = {
...
+        values: ['scaled', 'absolute'],
+        role: 'info',
+        editType: 'calc',
+        dflt: 'scaled',
+        description: [
+            'Sets the mode by which the cones are sized.',
+            'If *scaled*, `sizeref` scales such that the reference cone size',
+            'for the maximum vector magnitude is 1.',
+            'If *absolute*, `sizeref` scales such that the reference cone size',
+            'for vector magnitude 1 is one grid unit.'
```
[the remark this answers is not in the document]
etpinard: To be honest, I'm not sure. I copied this from https://github.com/gl-vis/gl-cone3d#constructor I'll have to inspect  

https://github.com/gl-vis/gl-cone3d/blob/c675b25e5991e63d0ad23ed24d712c496c28cd72/cone.js#L138-L246

a little bit more.

alexcjohnson:
> We currently call `cone2mesh` twice which can be slow (~300ms for 20k cones) per `.plot` call. Once in `calc` to get the colors (i.e. intensity) values and another in `convert` to compute the correctly scaled coordinate

If it's just for color (which, per https://github.com/plotly/plotly.js/pull/2641#issuecomment-389644575, might not not even come from the cone lengths), can't we just find the bounds of the data provided, as a simple loop over the cone data? That should be a heck of a lot faster (quick test: ~5ms for 20k data points) and, I might argue, more faithful to the data than basing the color bounds on the interpolated data used in the actual cones.

src/traces/cone/convert.js
```diff
@@ -48,25 +49,44 @@ function convert(scene, trace) {
...
-    var meshData = convert(scene, data);
-    var mesh = createMesh(gl, meshData);
-    var result = new Mesh3DTrace(scene, mesh, data.uid);
 
-    result.data = data;
-    mesh._trace = result;
+    var meshData = convert(scene, data);
+    var mesh = createConeMesh(gl, meshData);
+
+    var pts = createScatterPlot({
```
etpinard: a little bit hacky and not very efficient, but this works:

peek 2018-05-16 17-24

alexcjohnson: Nice, that was fast! Is it possible to get `(u, v, w)` in the hovertext as well?

src/plots/gl3d/scene.js
```diff
@@ -480,9 +480,12 @@ proto.plot = function(sceneData, fullLayout, layout) {
             var axLetter = axis._name.charAt(0);
 
             for(j = 0; j < objects.length; j++) {
-                var objBounds = objects[j].bounds;
-                sceneBounds[0][i] = Math.min(sceneBounds[0][i], objBounds[0][i] / dataScale[i]);
-                sceneBounds[1][i] = Math.max(sceneBounds[1][i], objBounds[1][i] / dataScale[i]);
+                var obj = objects[j];
+                var objBounds = obj.bounds;
+                // add worse-case pad for cone traces
+                var pad = (obj._trace.data._normMax || 0) * dataScale[i];
```
etpinard: This is what I came up with to solve the "cones out of scene" problem.

```js
Plotly.newPlot(gd, [{
	type: 'cone',
      "x": [1, 2, 3],
      "y": [1, 2, 3],
      "z": [1, 2, 3],
      "u": [1, 0, 0],
      "v": [0, 3, 0],
      "w": [0, 0, 2],
}])
```

now gives:

image

src/traces/cone/convert.js
etpinard: It's possible!

peek 2018-05-16 23-31

but maybe someone would argue that 6 fields in one hover label in too much by default. Maybe we could just show the vector norm value? Or maybe have `hoverinfo: 'x+y+z+norm'` by default and optionally up to: `x+y+z+u+v+w+norm+text`

alexcjohnson: > `'x+y+z+norm'` by default and optionally up to: `x+y+z+u+v+w+norm+text`

I like it!

src/plots/gl3d/scene.js
alexcjohnson: Looks good! I suppose based on https://github.com/gl-vis/gl-cone3d/issues/6 we could drop it to 3/4 of that value, though then if we make that configurable it would depend on the offset... more trouble than it's worth, let's keep what you have here.

src/traces/cone/attributes.js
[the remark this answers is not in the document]
kig: I'll have a new look at the cone sizing parameters. Pre-meshgrid they were based on the position space units and scaled so that two cones with maximum norm, pointing to the same direction, would have their tip and base touch. Like <|<|. That would be with coneSize 2. With coneSize 1, you'd get no overlapping cones even when they're |><|.

The absolute sizing ditched the automatic cone scaling and used the cone vector norm to figure out the size of the cone model, so that a norm of 1 with absolute size factor 1 results in a cone that's one unit tall in the position unit space. Absolute size factor 2 => 2 unit tall cone, etc.

With the meshgrid sampling (and the realization that the cone positions may not be on a nice regular grid), I changed the automatic cone sizing to scale based on the smallest distance between two adjacent cone positions (adjacent in the sense that they've got successive indices in the positions array). This is not the right thing to do though, doing a smallest distance between any two cone positions would be, but that's O(n^2) with a naive algo. 

Another idea floated on the sizing was to have pixel-based sizing as well, which'd probably work like "invert the projection, view and model transform matrices for cone model, scale cone model according to viewport size" but would result in orthogonal camera cones which might be confusing. (And it sounds like a bit of a pain to implement.)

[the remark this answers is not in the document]
alexcjohnson: @kig that all sounds great. I guess we only need to worry about the smallest-distance calculation when using the `(x, y, z)` data to position the cones; With the cartesian product of x/y/z interpolation grids each dimension should be pretty compact, even if the values are out of order. But I guarantee at some point we'll see a data set where the closest points are not successive. I don't know if there's a JS implementation already available but it's possible to do it in O(n log(n)^2) http://www.cs.ucsb.edu/~suri/cs235/ClosestPair.pdf

Lets not worry about pixel-based sizing, at least not for now. I guess I can see using this if we also do automatic grid spacing, ie when you zoom in and the cone spacing gets larger, at some point we decrease the grid size so we draw more cones between the existing ones. That's definitely not a v1 feature 😅

src/traces/cone/convert.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2641/commits/e4a203563ff02dca646aa904eda423da6df601d4

src/traces/cone/attributes.js
```diff
@@ -64,20 +64,37 @@ var attrs = {
...
-        description: 'Sets the x coordinates of the vector field mesh.'
-    },
-    vy: {
-        valType: 'data_array',
-        editType: 'calc',
-        description: 'Sets the y coordinates of the vector field mesh.'
-    },
-    vz: {
-        valType: 'data_array',
+    cones: {
```
etpinard: @alexcjohnson I went with `cones.(x|y|z)` which looks a lot like attribute containers `contours` and `bins`. Hope you'll like it.

src/traces/cone/index.js
```diff
@@ -22,7 +22,16 @@ module.exports = {
...
+            'Use cone traces to visualize vector fields.',
+            '',
+            'Specify a vector field using 6 1D arrays,',
+            '3 position arrays `x`, `y` and `z`',
+            'and 3 vector component arrays `u`, `v`, `w`.',
+            '',
+            'If nothing else is provided, the cones are drawn exactly at the positions given',
+            'by `x`, `y` and `z`.',
+            'To specify the cone positions manually, use arrays attributes',
+            '`cones.x`, `cones.y` and `cones.z`.'
```
alexcjohnson: "To display cones at different positions from the field data, use array attributes `cones.x`, `cones.y` and `cones.z`. Where these positions do not match data points, the field is interpolated ..." (How, what algorithm? And does the algorithm extend outside the convex hull of the data? Perhaps also mention interpolation in `attributes.cones.description`?)

etpinard: > How, what algorithm?

Can @kig help us out here?

etpinard:
I still need to fixup this case brought up by @alexcjohnson 

```js
// seems like something’s off with the axis padding - cones at 1 & 3 give axis ranges `[-2.46, 6.46]`? Shouldn’t that be `[0, 4]` if we’re padding by one full cell?

Plotly.newPlot(gd,[{type: 'cone', x: [1, 1, 1, 1, 3, 3, 3, 3], y: [1, 1, 3, 3, 1, 1, 3, 3], z: [1, 3, 1, 3, 1, 3, 1, 3],
u: [1, 2, 3, 4, 1, 2, 3, 4], v: [1, 1, 2, 2, 3, 3, 4, 4], w: [4, 4, 1, 1, 2, 2, 3, 3]}])
```

but other than, this PR seems ready. Dropping the _discussion needed_, tagging as _reviewable_.

src/plots/gl3d/scene.js
```diff
@@ -508,7 +508,7 @@ proto.plot = function(sceneData, fullLayout, layout) {
                 var obj = objects[j];
                 var objBounds = obj.bounds;
                 // add worse-case pad for cone traces
-                var pad = (obj._trace.data._normMax || 0) * dataScale[i];
+                var pad = 0.75 * (obj._trace.data._compMax || 0) * dataScale[i];
```
etpinard: This gives a better estimate. The data off https://github.com/plotly/plotly.js/pull/2641#issuecomment-390040970 now gives:

image

alexcjohnson: Great! Yeah, I was wrong with `[0, 4]` - this looks awesome.

test/jasmine/tests/cone_test.js
```diff
@@ -0,0 +1,93 @@
...
+
+    it('should set `visible: false` for traces missing x,y,z,u,v,w arrays', function() {
+        var keysToDelete = ['x', 'y', 'z', 'u', 'v', 'w', ''];
+
+        keysToDelete.forEach(function(k) {
+            gd = makeGD();
+            delete gd.data[0][k];
+
+            supplyAllDefaults(gd);
+            expect(gd._fullData[0].visible).toBe(!k, 'missing array ' + k);
```
alexcjohnson: Clever, but maybe a little too compact to include `''` in this set 😈

etpinard: ok, fixed in https://github.com/plotly/plotly.js/pull/2641/commits/8dca9ae9960e3811ddd9d7d913710624ed2cb3d3

src/plots/gl3d/scene.js
etpinard: latest attempt: https://github.com/plotly/plotly.js/pull/2641/commits/3d26d475209d744b889ad45920dce4cf359b1529

test/jasmine/tests/cone_test.js
```diff
@@ -117,6 +117,55 @@ describe('@gl Test cone autorange', function() {
...
+            w.push(0);
+
+            return Plotly.restyle(gd, {
+                x: [x], y: [y], z: [z],
+                u: [u], v: [v], w: [w]
+            });
+        })
+        .then(function() {
+            _assertAxisRanges('after adding one cone outside range but with norm-0',
+                [-0.45, 6.45], [-0.45, 6.45], [-0.45, 6.45]
```
etpinard: @alexcjohnson unfortunately I can't get this to "stick" to the previous `range[0]` values without breaking other gl3d mocks that depend on

https://github.com/plotly/plotly.js/blob/fb751582506a4d71d23a89786ba8138473e33cd9/src/plots/gl3d/scene.js#L479-L483

which scales the autorange bounds with the `dataScale`:

https://github.com/plotly/plotly.js/blob/fb751582506a4d71d23a89786ba8138473e33cd9/src/plots/gl3d/scene.js#L394-L407

Hopefully this iteration is _close enough_.

--------------

As a side note, we should spend some time replacing this `dataScale` hack. This thing was added to make date axes work properly in 3D. Timestamps don't fit in a Float32Array, so we can't just convert the datestrings to timestamps, but we do this better in `scattergl`, where we use Float64Array or split the data into two Float32Arrays when not supported.

src/traces/cone/convert.js
```diff
@@ -107,6 +108,9 @@ function convert(scene, trace) {
     meshData.fresnel = trace.lighting.fresnel;
     meshData.opacity = trace.opacity;
 
+    // stash autorange pad value
+    trace._pad = anchor2coneSpan[trace.anchor] * meshData.vectorScale * trace._normMax * trace.sizeref;
```
alexcjohnson: Awesome, I think we have a winner now! I don't see any cases (As I change the cone spacing, cone count in each direction, and vector length) where this fails to encompass all the cones, nor any where the padding looks unpleasantly large. 🏆

test/jasmine/tests/cone_test.js
```diff
@@ -102,7 +102,7 @@ describe('@gl Test cone autorange:', function() {
         })
         .then(function() {
             _assertAxisRanges('scaled up',
-                [-0.39, 4.39], [-0.39, 4.39], [-0.39, 4.39]
+                [-0.66, 4.66], [-0.66, 4.66], [-0.66, 4.66]
```
alexcjohnson: Can you add one more test, to 🔒 this in a little more since it went through so many iterations: scale up the x/y/z arrays. At least one of the previous iterations had a problem with this but otherwise looked pretty good.

etpinard: there's this one:

https://github.com/plotly/plotly.js/blob/abb11e003e8c454e703668c12990e3c122500f29/test/jasmine/tests/cone_test.js#L121-L144

am I missing anything?

alexcjohnson: It's close, but there's a difference between adding another point that *doesn't* change the minimum spacing between cones, and expanding the whole x/y/z space and therefore expanding the minimum spacing as well.

etpinard: got it. Thanks!

etpinard: done in -> https://github.com/plotly/plotly.js/pull/2641/commits/2c08357fc40005ac2505d63b374e784a0599cc9e

alexcjohnson:
Alright! I'm happy if you are! 💃 💃 💃

test/jasmine/tests/cone_test.js
etpinard: New issue about the `dataScale` mess -> https://github.com/plotly/plotly.js/issues/2646

etpinard:
Ok. Merging this thing!

Shoutout to @kig for this feature :tada: 

Thanks very much @alexcjohnson for the help making this plotly-proof! :muscle: 

This will be part of v1.38.0 set to be released on Tuesday. :rocket: 

-----

by the way @kig I published `gl-cone3d@1.0.0` to npm.

PR #2647, opened 2018-05-21
Fix single cone

etpinard:
A follow-up on https://github.com/plotly/plotly.js/pull/2641, bringing in @kig's https://github.com/gl-vis/gl-cone3d/pull/9 and a few other cone tweaks.

cc @alexcjohnson


test/image/mocks/gl3d_cone-single.json
```diff
@@ -0,0 +1,27 @@
...
+      "showscale": false
+    },
+    {
+      "type": "cone",
+      "name": "green cone",
+      "x": [3], "y": [3], "z": [3],
+      "u": [0], "v": [0], "w": [2],
+      "colorscale": "Greens",
+      "showscale": false
+    }
```
alexcjohnson: Just to play devil's advocate re: autoscale - perhaps the single-cone mock should really be just one single cone, so the data has zero range without considering the cone itself? And multi-trace could be incorporated into another mock?

etpinard: yeah good call -> https://github.com/plotly/plotly.js/pull/2647/commits/7da71ae44f0297535ebd26049f76a21292e37fed

alexcjohnson:
Beautiful! Thanks for cleaning up multitrace and `compMax` along the way! 💃

PR #2649, opened 2018-05-22
fix #2598 - sloppy click on cartesian zoom

### LINKED ISSUE #2598: sloppy plotly_click (slight cursor move during click)
Sometimes, a user moves the mouse slightly during a click.
In this case, `plotly_click` does not fire, nor does zoom happen due to some minimum drag length for zoom. I've seen users frustrated by this behavior.
I think `plotly_click` should fire when there's a slight drag of 1 or 2 pixels. Or when less than the minimum drag length.
Maybe this behavior would need a new config property for developers to opt-in to?
Alternatively, create a new event? `plotly_tinydrag`, `plotly_sloppy_click`

alexcjohnson:
As mentioned in a comment, zoom takes over `minDrag`, so it also has to take over `gd._dragged`
Also removed a little obsolete behavior while I was at it.

cc @etpinard

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -252,6 +252,9 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         zb,
         corners;
 
+    // zoom takes over minDrag, so it also has to take over gd._dragged
+    var zoomDragged;
+
     // collected changes to be made to the plot by relayout at the end
     var updates = {};
 
@@ -266,6 +269,7 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         path0 = 'M0,0H' + pw + 'V' + ph + 'H0V0';
         dimmed = false;
         zoomMode = 'xy';
+        zoomDragged = false;
 
         zb = makeZoombox(zoomlayer, lum, xs, ys, path0);
 
@@ -340,6 +344,9 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         }
         box.w = box.r - box.l;
         box.h = box.b - box.t;
+
+        if(zoomMode) zoomDragged = true;
+        gd._dragged = zoomDragged;
 
         updateZoombox(zb, corners, box, path0, dimmed, lum);
         dimmed = true;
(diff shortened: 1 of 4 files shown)


src/plots/cartesian/dragbox.js
```diff
@@ -266,6 +269,7 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         path0 = 'M0,0H' + pw + 'V' + ph + 'H0V0';
         dimmed = false;
         zoomMode = 'xy';
+        zoomDragged = false;
```
etpinard: this one is in `zoomPrep`

src/plots/cartesian/dragbox.js
```diff
@@ -341,6 +345,9 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
         box.w = box.r - box.l;
         box.h = box.b - box.t;
 
+        if(zoomMode) zoomDragged = true;
+        gd._dragged = zoomDragged;
```
etpinard: and this one move in `zoomMove`

test/jasmine/assets/click.js
```diff
@@ -43,10 +45,16 @@ module.exports = function click(x, y, optsIn) {
...
+    var slop = opts.slop || [];
+    var slopX = slop[0] || 0;
+    var slopY = slop[1] || 0;
+    var slopOpts = Lib.extendFlat({}, downOpts);
+    delete slopOpts.button;
 
     mouseEvent('mousemove', x, y, moveOpts);
     mouseEvent('mousedown', x, y, downOpts);
     if(callContext) mouseEvent('contextmenu', x, y, downOpts);
+    if(slopX || slopY) mouseEvent('mousemove', x + slopX, y + slopY, slopOpts);
```
etpinard: So, `slop` is only used for testing? This is fine, but perhaps we could add a comment about this in the jsDoc above so that we don't feel like :hocho: down the road.

alexcjohnson: sure -> https://github.com/plotly/plotly.js/pull/2649/commits/c56f223fd0aff569ac9c2da3f2ae79957dab8908

etpinard:
Thanks! Looks good :dancer:

Issue #2631, opened 2018-05-13
Violin Span Calculation Issue


kmui2:
Regarding this line of code in plotly.js: https://github.com/plotly/plotly.js/blob/9d6144304308fc3007f0facf2535d38ea3e9b26c/src/traces/violin/calc.js#L51

Why is taking the minimum (`Math.min(bandwidthDflt, bandwidth)`) needed here? Should it just not take the min and use only `bandwidth`? I can see a problem where I defined `trace.bandwidth` and don't want to use `bandwidthDflt`.

etpinard:
See -> https://github.com/plotly/plotly.js/pull/2552#discussion_r181206232

PR #2650, opened 2018-05-22
Violin span fix

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2631 and supersedes https://github.com/plotly/plotly.js/pull/2552

This PR uses @joshua-gould's https://github.com/plotly/plotly.js/pull/2552/commits/9329a4dd5e8065626710d8e6d311ef8997c1d9a7 where the set bandwidth value is used to determine the number of pts in the kde estimate (as opposed to the min of the set bandwidth and the Silverman's rule of thumb).

Moreover, this PR uses Scott's rule of thumb when the Silverman value yields too many kde points. This happens when the IQR is not a very good estimate of the distribution's spread like in the data of https://github.com/plotly/plotly.js/pull/2552#issue-181309494.

cc @alexcjohnson


alexcjohnson:
I still have two concerns about this:
- The user can still lock things up with a custom bandwidth that's sufficiently small (and potentially even automatically, say you provide a million zeros and one one? That might still be OK though...) - we need *some* absolute limit.
- A sharp transition from Silverman to Scott means that some distributions that are *actually* narrower will end up looking wider. In your example say you tweaked it to be just on the Silverman side of the limit: you'd get 1e5 bandwidths in the span, then narrow it a little and it would pop out to essentially what's in your test image, looks like ~50 bandwidths in the span. I guess I don't mind if we reach a point where distribution width stops affecting bandwidth, but not if the dependence goes the wrong direction.

What if the automatic bandwidth just used Silverman but limited it to `> (max - min)/100` or something? Maybe `/100` is too wide still? `/1000`? I think a narrower result would do a better job showing the shape of this particular distribution anyhow, don't worry about matching Seaborn. And then the `1e5` limit would constrain all cases, manual or automatic. That's far narrower than one pixel, until you've zoomed waaaay in.

etpinard:
> What if the automatic bandwidth just used Silverman but limited it to > (max - min)/100 or something?

Going with `/100` in -> https://github.com/plotly/plotly.js/pull/2650/commits/36cc37d11bd046c4e7d6bf20d8302f564a51d094

src/traces/violin/calc.js
```diff
@@ -80,21 +80,27 @@ function silvermanRule(len, ssd, iqr) {
...
+    // Limit how small the bandwidth can be.
+    //
+    // Silverman's rule of thumb can be "very" small
+    // when IQR does a poor job at describing the spread
+    // of the distribution.
+    // We also want to limit custom bandwidths
+    // to not blow up kde computations.
+    return ((cdi.max - cdi.min) / bw) < 1e5 ?
+        bw :
+        (cdi.max - cdi.min) / 100;
```
alexcjohnson: wait, that's still going to jump from 1/1e5 up to 1/100. I had in mind something like:
```js
if(trace.bandwidth) {
    return Math.max(trace.bandwidth, span / 1e5);
}
else {
    return Math.max(silvermanValue, span / 100);
}
```

etpinard: Oops. My bad -> https://github.com/plotly/plotly.js/pull/2650/commits/c55407079d9006f753156c4446b2bcd4104a1e9a

src/traces/violin/calc.js
```diff
@@ -98,9 +90,17 @@ function calcBandwidth(trace, cdi, vals) {
     // of the distribution.
     // We also want to limit custom bandwidths
     // to not blow up kde computations.
-    return ((cdi.max - cdi.min) / bw) < 1e5 ?
-        bw :
-        (cdi.max - cdi.min) / 100;
+
+    if(trace.bandwidth) {
+        return Math.max(trace.bandwidth, span / 1e4);
```
alexcjohnson: From slack, for the record:

@etpinard: limiting custom bandwidth at (span / 1e5) is still pretty slow. ok if I increase it to span / 1e4 ?

**Me**: sure, that’s still more than 1 bw/px. I’m not quite so worried about it though, at that point the user has asked for something explicitly ridiculous… but I guess we don’t want our mock to take forever to render!

alexcjohnson:
Looks great! 💃

PR #2656, opened 2018-05-23
Fixup noci_test.sh

etpinard:
... and a few dependency updates.

`npm run test-jasmine -- --tags=noCI` which is part of the [pre-release checklist](https://github.com/plotly/dev-docs/blob/master/advanced/plotly.js.md#checklist) has been broken since https://github.com/plotly/plotly.js/pull/2577 as running just the `@noCI` tests doesn't test all the avaiable trace and transforms. **Solution:** tag the react tests with `@noCIdep` and run that tag in `./tasks/noci_test.sh`.


etpinard:
oh oops, forgot to tag @alexcjohnson

PR #2657, opened 2018-05-23
remove cached `trace._interpz` from heatmaps/contour maps with gaps

### LINKED ISSUE #2652: React contour problems
Split from https://github.com/plotly/plotly.js/issues/2624, with reproducible example and explanations in https://codepen.io/etpinard/pen/aGxMVX

cc @alexcjohnson

alexcjohnson:
Fixes #2652 - turns out this isn't a react-specific issue, it's just about changing z data for contours with gaps in the data (and heatmaps with `connectgaps: true`). We were caching the interpolation results, even across recalc calls, because this can be a pretty slow (iterative) process. I removed that cache, after first trying to figure out whether we could salvage it, based on comparing the incoming known (non-interpolated) z values to the old ones. It turned out that was difficult (and expensive) to get right, especially for edge cases like removing a previously filled-in value without changing anything else.

We're doing a better job than we used to do (3+ years ago when this cache was added) at avoiding unnecessary recalcs, so I don't think this should be too big a cost, but this is more motivation to push in that direction, especially things like recalc'ing only the traces that need it.

cc @etpinard

### DIFF
--- a/src/traces/heatmap/interp2d.js
+++ b/src/traces/heatmap/interp2d.js
@@ -10,8 +10,8 @@
 
 var Lib = require('../../lib');
 
-var INTERPTHRESHOLD = 1e-2,
-    NEIGHBORSHIFTS = [[-1, 0], [1, 0], [0, -1], [0, 1]];
+var INTERPTHRESHOLD = 1e-2;
+var NEIGHBORSHIFTS = [[-1, 0], [1, 0], [0, -1], [0, 1]];
 
 function correctionOvershoot(maxFractionalChange) {
     // start with less overshoot, until we know it's converging,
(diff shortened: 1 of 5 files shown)


src/traces/heatmap/interp2d.js
```diff
@@ -10,34 +10,37 @@
...
-        }
-    }
-    else {
-        // one pass to fill in a starting value for all the empties
-        iterateInterp2d(z, emptyPoints);
-    }
+/*
+ * interp2d: Fill in missing data from a 2D array using an iterative
+ *   poisson equation solver with zero-derivative BC at edges.
+ *   Amazingly, this just amounts to repeatedly averaging all the existing
```
etpinard: _Amazingly_ :sparkles:

test/jasmine/tests/contour_test.js
```diff
@@ -458,4 +458,46 @@ describe('contour plotting and editing', function() {
...
+            }]);
+        })
+        .then(function() {
+            expect(gd.calcdata[0][0].z).toEqual([[1, 2], [2, 4], [1, 2]]);
+            expect(gd.calcdata[0][0].zmask).toEqual([[1, 1], [0, 1], [1, 1]]);
+
+            return Plotly.react(gd, [{
+                type: 'contour',
+                // notice that this one is the same as the previous, except that
+                // a previously present value was removed...
```
etpinard: thanks for this comment :books:

alexcjohnson:
Benchmarks - compare a full heatmap, which plots in around 70ms for me at n=100 (100x100 heatmap):
```js
function hmFull(n) {
  var z = new Array(n); for(var i = 0; i < n; i++) { z[i] = new Array(n); for(var j = 0; j < n; j++) z[i][j] = Math.random(); }
  var t0 = performance.now();
  Plotly.newPlot(gd, [{z: z, type: 'heatmap', showscale: false}], {width: 500, height: 500})
  return performance.now() - t0;
}
```

to a gapped one:
```js
function hmPartial(n) {
  var z = new Array(n); for(var i = 0; i < n; i++) {
    z[i] = new Array(n);
    for(var j = 0; j < n; j++) z[i][j] = null; z[i][Math.floor(Math.random() * n)] = Math.random();
  }
  var t0 = performance.now();
  Plotly.newPlot(gd, [{z: z, type: 'heatmap', showscale: false, connectgaps: true}], {width: 500, height: 500})
  return performance.now() - t0;
}
```

A fairly full heatmap (n=100, m=50, so about 50% gaps) spends about 40ms on `interp2d`.
A fairly empty one (n=100, m=5, so about 95% gaps) spends about 450ms on `interp2d`. Only twice as many gap points to fill in, but it iterates much more slowly.

PR #701, opened 2016-06-28
Streamtubes [WIP]

monfera:
image
# Goals

Streamtubes is a new trace type that gives plotly.js the ability to generate streamtubes. Inspirations:
- science, e.g. [connectomics](http://web.cse.msstate.edu/~szhang/papers/Zhang-2004-VIM.pdf)
- industry, e.g. [energy](http://d381qxy5z1upqu.cloudfront.net/wp-content/uploads/2015/09/slide_dernier.jpg)
- general tensor-based visualizations, for [flow or other things in 3D](http://teaching.csse.uwa.edu.au/units/CITS4241/Handouts/Lecture21.html)
- space filling or stick [molecular visualization](http://3.bp.blogspot.com/_DZH2cmCoois/Rp_NJ5IavrI/AAAAAAAAClQ/JBTxF7vYHwM/s1600/figure+19-15.jpg), [molecular structure](https://en.wikipedia.org/wiki/Biomolecular_structure#/media/File:Protein_alphabet.tif) or [molecular marking](http://blogs.plos.org/wp-content/uploads/2011/06/Fractal-globule-480.jpg)
- progress toward more general use in [engineering or related presentation](http://www.rishabheng.com/projects-portfolio/)
- [art](https://twitter.com/choldgraf/status/747880611327578112)

Direct motivations included:
- ability to do 3D scatterplots with curved lines, gradient colors and varying line width
- avoid the Z-fighting issue that's present in `scatter3d` when point markers are of a different color than the lines
- provide nice, realistic connections among markers and connecting lines
- benefit from our revamped lighting and reflection support for aesthetic or realistic rendering

Immediate goals: a `streamtubes` trace type, which
- has an API as close to `scatter3d` as possible
- generates realistic looking 3D markers (in this round, icospheres) and connecting lines (in this round, of spherical cross-section)
- each marker should be able to have a different color and size
- connecting lines should support varying radius, varying color (gradient or discrete)
- use splines for smoothness of a fine trace, or for curviness with low sample count, and for smoothly varying radius and color

Nice to have goals:
- provide text labeling of markers
- support the plane projection as in `scatter3d`
- support a color legent as in `scatter3d`
# Previous steps

Development of [#617](https://github.com/plotly/plotly.js/pull/617/files) (now merged; [example1](http://codepen.io/monfera/pen/OXWLZp) [example2](http://codepen.io/monfera/pen/oLYOjq) [example3](http://codepen.io/monfera/pen/gMLEEB) and see also the test cases in the changeset)
[...]

PR #2658, opened 2018-05-23
3D streamtube traces

etpinard:
_Note that this branch is @kig 's https://github.com/plotly/plotly.js/tree/gl-vis-streamtube rebased with the 2018/05/28 master._

----------

Streamtubes are coming to plotly.js, this time for real (https://github.com/plotly/plotly.js/pull/701 :smirk: )

peek 2018-05-23 17-15

But first, we'll need to bring this branch up-to-speed with the work done for cones in https://github.com/plotly/plotly.js/pull/2641 :hammer_and_pick: 

-------------

Here's a list of TODOs:

- [x] get https://github.com/gl-vis/gl-streamtube3d/pull/1 merged
- [x] figure out how to split stream _tubes_ from stream _lines_:
  + see https://github.com/plotly/plotly.js/tree/gl-vis-streamline
  + and https://github.com/plotly/plotly.js/commit/812be20750e21e0a1831975001c248d365850f73#r29119106
- [x] figure out how hover should work? Just like cones with gl-scatter3d points at the vector fields' (x,y,z) coordinates or something that would suit streamtubes better
- [x] figure out if we expose or not "meshgrid" attributes?
  + meshgrid should be more useful for tubes compared to cones
  + [MATLAB](https://www.mathworks.com/help/matlab/ref/streamtube.html) and the current [`gl-streamtube3d`](https://github.com/gl-vis/gl-streamtube3d/blob/076b0a6e86ccd1f00be6a9635eccfd4744acccbb/example/streamtube.js#L61) iteration allows users to define _starting positions_, would that be enough for the v1?
https://github.com/plotly/plotly.js/commit/812be20750e21e0a1831975001c248d365850f73#r29119106
  + we should think about an API allow user to "fill the scene with tubes" even in scenarios where several streamlines converge and end.
- [x] properly scaffold `streamtubes` as in https://github.com/plotly/plotly.js/pull/2641/commits/fa32a74bfcb00107116304c57720f89f7b161f2e
- [x] hook autorange logic
   + need to consider startx/starty/startz as well as vector field x/y/z,
   + need to consider `tubeScale`
- [x] Implement auto "widthScale" algo (see https://github.com/plotly/plotly.js/commit/812be20750e21e0a1831975001c248d365850f73#r29119106)
- [x] plotly.js-ize attributes
  + [x] no camelCase
  + [x] use colorscale, cmin and cmax as for `cone` traces
  + [x] `bounds` isn't a per-trace setting in plotly.js
  + [x] find out what `maxLength` does
  + [x] find out what `widthScale` does (would using `sizeref` in plotly.js like for cones be ok?)
  + [x] fill in attribute descriptions
- [x] properly hook in colorbar (we might be able to simply reuse `cone/calc.js` and `cone/colorbar.js` here)
- [x] add a few image mocks
  + [x] add one with the data of https://github.com/gl-vis/gl-streamtube3d/blob/master/example/dataset-wind.js
- [x] add jasmine tests similar to the cone [suite](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/cone_test.js)
- [x] add react test as in https://github.com/plotly/plotly.js/pull/2656/commits/07fc2a02720bd1a9baf1d5744ab5870c592ece19
- [x] get syntax and plotschema tests to pass
- [x] publish `gl-streamtube3d` to npm and update package.json dep
- [x] open "open item" issue

cc @alexcjohnson @jackparmer @kig


etpinard:
**Update:**

Here's a few problems I noticed yesterday:

`gl-streamtube3d` exposes a `meshgrid` attribute that "samples" (i.e. downscales) the vector field about the cartesian product of arrays of x,y,z coordinates. But now, @kig how can users specify the vector field's own x,y,z coordinates (like what the `positions` setting does for `gl-cone3d`)?

From https://github.com/plotly/plotly.js/commit/812be20750e21e0a1831975001c248d365850f73#r29129877, `gl-streambed3d`'s `widthScale` setting has a hard-coded default that does a very poor job in some cases. :sos: We must fix this before the v1 release. :sos:  I think @kig will be taking this one.

On-par with `cone` traces, the colorscale's cmin and cmax auto value are computed using the u/v/w max vector norm. For cones, this turns out to be exactly the span of color values displayed, but for streamtubes cmin-cmax overestimate the span of color values displayed, as the tubes aren't drawn at the vector field coordinates. So, should we compute the max vector norm at the tubes' location to determine cmin and cmax? I can see an argument for both methods. Computing cmin/cmax just from u/v/w is probably more "true to the data" whereas computing cmin/cmax from the vector field interpolated to the tubes' locations will give better aesthetically-pleasing results.

As for hover, it's an interesting problem. I'm thinking of three families of solution. 1) Just show the vector field x/y/z and u/v/w pt values just like for cones. This solution is true to the input data, but doesn't offer info about the tubes. 2) Just show the coordinates of the tubes' vertices and the u/v/w components at their location. Here, we're entirely in "tube-space", the input x/y/z and u/v/w don't show up on hover. 3) Show the tubes' starting position (i.e. items from `startx`, `starty` and `startz`) and the tubes' center coordinates plus (optionally) the scalar velocity at the tubes' center. When hovering over a section of a tube, it seems useful to me to know where that tracer started from.

Thoughts?

alexcjohnson:
> Computing cmin/cmax just from u/v/w is probably more "true to the data" whereas computing cmin/cmax from the vector field interpolated to the tubes' locations will give better aesthetically-pleasing results.

"True to the data" is my preference here - it shows you a bit of what you've missed about the data by converting to tubes. It shouldn't often be very far off the interpolated values, right?

> As for hover, it's an interesting problem.

Interesting indeed - I actually think "true to the data" would be really confusing here, as there's nothing to "hover on" at the data point. Particularly since this is 3D, you could end up hovering on something at a totally different depth from the tube you're inspecting. The tube vertices (2) don't mean anything. I assume when you say this you're talking about the surface of the tube? I'm thinking the ideal solution (tell me how painful this would be to calculate...) would be to find the exact tube surface point the cursor is over (not the closest vertex) and project from there back to the centerline of the tube; show x/y/z and u/v/w/norm for that point (or perhaps the closest actual data point? that could be confusing though); and perhaps somehow highlight the whole tube, so you can see not just where it came from but the whole streamline? That last part seems more like a "nice to have" that we could add later though.

etpinard:
> "True to the data" is my preference here 

Fantastic :ok_hand: 

> I assume when you say this you're talking about the surface of the tube?

Yes, exactly.

> I'm thinking the ideal solution (tell me how painful this would be to calculate...) would be to find the exact tube surface point the cursor is over (not the closest vertex) and project from there back to the centerline of the tube; show x/y/z and u/v/w/norm for that point

I agree, this sounds like ideal. I'll have to dig into how the tubes coordinates are computed to answer the "how painful this would be" question. Maybe @kig would be interested in helping us out?

kig:
> gl-streamtube3d exposes a meshgrid attribute that "samples" (i.e. downscales) the vector field about the cartesian product of arrays of x,y,z coordinates. But now, @kig how can users specify the vector field's own x,y,z coordinates (like what the positions setting does for gl-cone3d)?

The streamtube visualization needs a continuous vector field. It works by sampling the vector field at each of the starting points, then moving a small amount along the sampled vector, sampling again, moving again, repeat until far enough from the starting point to create a stream segment. Push stream segment to the stream points array, continue until out of bounds or reach maxLength segments in stream points array.

The meshgrid parameter defines the vector field does for the streamtubes. It maps each of the vectors to a position in space. And since it's a nicely defined grid, it's easy to sample values inside each grid cell by interpolating between the 8 corners of the cell according to the sample position. 

Doing arbitrary vector positions for the vector field would be doable if you've got some nice algorithm to sample them as a continuous vector field.

Re: widthScale, auto scale, I wrote an auto scaler this morning. https://github.com/gl-vis/gl-streamtube3d/commit/97c89d7d0e4e4ed1074d47fbfc715bb2d149841c -- it also removes the widthScale parameter and replaces it with tubeSize and absoluteTubeSize params.

The auto scaler keeps track of the maximum divergence value encountered and uses that to calculate the radius of the tubes. The tubeSize param controls the auto scaled width so that tubeSize of 1 would scale two max divergence tubes at adjacent starting positions so that they don't overlap.

The absoluteTubeSize param overrides the auto scaler, and is used to multiply the divergence value to get the tube radius in coordinate space units.

The divergence used for the streamtubes is || divF || instead of the divF 3-vector. I don't know if that's the right thing to do or if we should asymmetrically scale the tubes according to the 3D divergence. 

> I'm thinking the ideal solution (tell me how painful this would be to calculate...) would be to find the exact tube surface point the cursor is over (not the closest vertex) and project from there back to the centerline of the tube; show x/y/z and u/v/w/norm for that point

This sounds doable, the tubes definition contains all the sampled interpolated data points, and picking should get you the triangle & data point matching that. What would you need for this?

etpinard:
> Re: widthScale, auto scale, I wrote an auto scaler this morning. gl-vis/gl-streamtube3d@97c89d7 -- it also removes the widthScale parameter and replaces it with tubeSize and absoluteTubeSize params.

Thanks @kig ! I'll try to incorporate your work in this PR in the next few hours.

etpinard:
> The streamtube visualization needs a continuous vector field.
> The meshgrid parameter defines the vector field for the streamtubes. It maps each of the vectors to a position in space. And since it's a nicely defined grid, it's easy to sample values inside each grid cell by interpolating between the 8 corners of the cell according to the sample position.

Thanks for writing this down @kig . You're absolutely right, the "mesh" requirements for streamtubes are much more stringent than for cones. Too bad x/y/z in will have a slightly different meaning in `streamtube` vs in `cone` traces, but I think we can live with that.

Having x/y/z generate of "mesh" of `x.length * y.length * z.length` items using a cartesian product and requiring u/v/w to be of length `x.length * y.length * z.length` isn't the most plotly.js-esque API unfortunately. I guess we _could_ make u/v/w 3D arrays. This would be a first though, and I doubt that's better. Perhaps requiring x/y/z to list all coordinates (so that x/y/z and u/v/w would have matching lengths) would be best albeit (way) less efficient. For example,

```js
var x = [], y = [], z = [];
var u = [], v = [], w = [];

for(var i = 0; i < nx; i++) {
  for(var j = 0; j < ny; j++) {
    for(var k = 0; k < nz; k++) {
      x.push(i); y.push(j); z.push(k);

      u.push(1+Math.sin(i));
      v.push(Math.cos(j));
      w.push(Math.sin(k*0.3)*0.3);
    }
  }
}
```

etpinard:
> And it'll ensure users can switch between streamtube and cone relatively robustly. Then we could fill in all the corresponding gaps in cone data handling as well.

This is winning argument :trophy: . @kig would you be interesting in converting the `gl-streambed3d` meshgrid argument to x/y/z positions?

kig:
By converting the meshgrid argument to x/y/z positions, do you mean that it'd be like an "expanded" meshgrid? You'd have the same x coordinates for every y and the same y coords for every z, and they'd be sorted small to big?

For example,

    var x = [0,1,2,0,1,2,0,1,2,0,1,2],
        y = [4,4,4,5,5,5,4,4,4,5,5,5],
        z = [6,6,6,6,6,6,7,7,7,7,7,7];

    var uniq = function(arr, e) {
      if (arr.length === 0 || e > arr[arr.length-1]) {
        arr.push(e);
      }
      return arr; 
    };

    var meshgrid = [x.reduce(uniq, []), y.reduce(uniq, []), z.reduce(uniq, [])];
    => [ [0,1,2], [4,5], [6,7] ]

kig:
Btw the cone meshgrid argument works in a similar fashion. It interprets the vector array as a vector field, and then uses the cone positions as points to sample from the vector field.

etpinard:
> it'd be like an "expanded" meshgrid? You'd have the same x coordinates for every y and the same y coords for every z, and they'd be sorted small to big?

Yes, exactly. Commit https://github.com/plotly/plotly.js/pull/2658/commits/e99ffc9612118f0e3ae1f999798c23bbf24079ed add a x/y/z columns -> meshgrid vector convertion in plotly.js before calling `tube2mesh`.

----

> It interprets the vector array as a vector field, and then uses the cone positions as points to sample from the vector field.

We chose to not expose gl-cone3d's "meshgrid" setting to plotly.js users in https://github.com/plotly/plotly.js/pull/2641/commits/5a42de085bf5829a5108a7c8551d92a95cb48d6f. In this context, (if I understand correctly), gl-cone3d [assumes](https://github.com/gl-vis/gl-cone3d/blob/8d0c829e49b8680d22de11377c02e3f7dcc7f39c/cone.js#L187) that the x/y/z `positions` and u/v/w `vectors` input match.  So now, plotly users using the same x/y/z and u/v/w columns can plot `streamtube` and `cone` traces with no array additional array manipulations.

src/traces/streamtube/convert.js
```diff
@@ -38,19 +38,19 @@ proto.handlePick = function(selection) {
...
         selection.traceCoordinate = [
             fromDataScale(pos[0], 'xaxis'),
             fromDataScale(pos[1], 'yaxis'),
             fromDataScale(pos[2], 'zaxis'),
 
-            uu, vv, ww,
-            Math.sqrt(uu * uu + vv * vv + ww * ww)
+            null, null, null,
+
+            selection.data.intensity * this.data._normMax
```
etpinard: First cut streamtube hover:

peek 2018-06-01 17-28

kig:
> plotly users using the same x/y/z and u/v/w columns can plot streamtube and cone traces with no array additional array manipulations.

That makes sense, yeah. You may want to have some utils to create a sparser version of a dataset for use with cones. Streams can work with arbitrarily dense datasets, but cone plots might get a bit unreadable / low-performance with dense fields.

Perf-wise, as long as you have fewer than 100k cones, should be fine rendering-wise. (My guess is that they're geometry processing limited - 48 vertices per cone and a bit complex vertex shader.) In comparison, stream plot performance is driven by the number of starting points and the maxLength parameter - amount of data points shouldn't affect stream plot perf.

etpinard:
> Perf-wise, as long as you have fewer than 100k cones, should be fine rendering-wise. (My guess is that they're geometry processing limited - 48 vertices per cone and a bit complex vertex shader.) In comparison, stream plot performance is driven by the number of starting points and the maxLength parameter - amount of data points shouldn't affect stream plot perf.

Thanks for the info @kig !

etpinard:
Tagging as `reviewable`.

alexcjohnson:
Looks like there's a broken tube in `gl3d_streamtube-first`:
screen shot 2018-06-05 at 3 33 18 pm
near here:
screen shot 2018-06-05 at 3 34 00 pm

I also notice that when multiple tubes come together and overlap we don't do anything to stop this, for example this set of 3 near the one I flag ^^ that spend most of their length conjoined. Is that a normal way of handling things? Not what I would have expected comparing these to 2D streamline plots, for example, where a line usually disappears when it gets too close to another.

Related: what if a line gets too close to *itself*, or goes on forever? Like, what would happen if we express the Lorentz attractor as a vector field and try to draw tubes on it? Or a magnetic field, whose lines *always* close?

src/lib/zip3.js
```diff
@@ -0,0 +1,17 @@
...
+
+'use strict';
+
+module.exports = function zip3(x, y, z) {
+    var result = new Array(x.length);
+    for(var i = 0; i < x.length; i++) {
+        result[i] = [x[i], y[i], z[i]];
+    }
+    return result;
+};
```
alexcjohnson: Nice to pull this into `lib` but are we moving away from the model of including all of these in `lib/index` and using them from there?

src/traces/streamtube/attributes.js
```diff
@@ -0,0 +1,151 @@
...
+        ].join(' ')
+    },
+
+    // TODO
+    //
+    // Should add 'absolute' (like cone traces have), but currently gl-streamtube3d's
+    // `absoluteTubeSize` doesn't behave well enough for our needs.
+    //
+    // 'fixed' would be a nice addition to plot stream 'lines', see
+    // https://github.com/plotly/plotly.js/commit/812be20750e21e0a1831975001c248d365850f73#r29129877
```
alexcjohnson: Thanks for the 📚 - you're collecting these for a "streamtubes open items" issue?

src/plots/gl3d/scene.js
```diff
@@ -88,24 +88,27 @@ function render(scene) {
...
             if(isHoverinfoAll || hoverinfoParts.indexOf('w') !== -1) {
-                coneTx.push('w: ' + formatter('zaxis', selection.traceCoordinate[5]));
+                vectorTx.push('w: ' + formatter('zaxis', selection.traceCoordinate[5]));
             }
             if(isHoverinfoAll || hoverinfoParts.indexOf('norm') !== -1) {
-                coneTx.push('norm: ' + selection.traceCoordinate[6].toPrecision(3));
+                vectorTx.push('norm: ' + selection.traceCoordinate[6].toPrecision(3));
+            }
+            if(isHoverinfoAll || hoverinfoParts.indexOf('divergence') !== -1) {
+                vectorTx.push('divergence: ' + selection.traceCoordinate[7].toPrecision(3));
```
alexcjohnson: `cone` with `hoverinfo:'all'`?

src/traces/streamtube/attributes.js
```diff
@@ -0,0 +1,151 @@
...
+            'Sets the y components of the starting position of the streamtubes',
+        ].join(' ')
+    },
+    startz: {
+        valType: 'data_array',
+        editType: 'calc',
+        description: [
+            'Sets the z components of the starting position of the streamtubes',
+        ].join(' ')
+    },
```
alexcjohnson: would this be better as `starts: {x, y, z}`? I can imagine later on when we figure out automatic starts wanting to tweak them with some other attributes that might be nice to collect all in a sub-container.

src/lib/zip3.js
etpinard: I like to keep subplot-specific lib function out of Lib (zip3 only makes sense for gl3d traces), but maybe it would less confusing to make `src/lib/index.js` collect everything in `src/lib/`. I don't have a strong opinion here.

etpinard:
> I also notice that when multiple tubes come together and overlap we don't do anything to stop this, for example this set of 3 near the one I flag ^^ that spend most of their length conjoined. Is that a normal way of handling things? Not what I would have expected comparing these to 2D streamline plots, for example, where a line usually disappears when it gets too close to another.

cc @kig 

> or goes on forever?

`gl-streamtube3d` has a maxLength (mapped to `maxdisplayed` in plotly.js) which sets the maximum number of segments per tube. The default is 1000 segments. So don't worry about tubes that go on forever.

src/lib/zip3.js
alexcjohnson: It's true, it's nice not to be carrying code in every build if only one subplot type uses it... but then, if it's really only `gl3d`, does it belong in `plots/gl3d/zip3.js`? You nailed my concern anyway, that I don't want to have to wonder every time I use a lib function whether I need to pull it in directly or if I can just require `lib/index` and pull all the pieces I need out from there.

src/traces/streamtube/convert.js
```diff
@@ -0,0 +1,217 @@
...
+        // to generate "well-defined" tubes
+        var sy0 = sceneLayout.yaxis.d2l(trace._ybnds[0]) * dataScale[1];
+        var sx, sz;
+        if(valsx.length > 2 && valsz.length > 2) {
+            sx = valsx.slice(1, valsx.length - 1);
+            sz = valsz.slice(1, valsy.length - 1);
+        } else {
+            // use all pts on 2x2 and 1x1 planes
+            sx = valsx.slice();
+            sz = valsz.slice();
```
alexcjohnson: Presumably this `length > 2` condition should be done separately for x and z? Also, I don't know what model is used for extending the field outside the x/y/z bounds, but if it's uncertain enough that at 3+ points it's best not to start at the last coordinate, then I'd think at 2 points it'd be best to use a single point halfway between the two.

src/traces/streamtube/index.js
```diff
@@ -0,0 +1,56 @@
...
+        out.tubev = pt.traceCoordinate[4];
+        out.tubew = pt.traceCoordinate[5];
+
+        out.norm = pt.traceCoordinate[6];
+        out.divergence = pt.traceCoordinate[7];
+
+        // Does not correspond to input x/y/z, so delete them
+        delete out.x;
+        delete out.y;
+        delete out.z;
```
alexcjohnson: 🍰

kig:
> I also notice that when multiple tubes come together and overlap we don't do anything to stop this, for example this set of 3 near the one I flag ^^ that spend most of their length conjoined. Is that a normal way of handling things? Not what I would have expected comparing these to 2D streamline plots, for example, where a line usually disappears when it gets too close to another.

This could be done in a post processing pass, checking if any two points have a distance below an epsilon -> throw away the shorter stream suffix. As it is, the stream generator just generates a stream from each starting point and draws them all out.

src/traces/streamtube/convert.js
etpinard: > Also, I don't know what model is used for extending the field outside the x/y/z bounds

Currently, positions outside the mesh get (u=0, v=0, w=0) if I understand correctly:

https://github.com/gl-vis/gl-streamtube3d/blob/0359ed9696de5af570fa6a5e36bdfbca2013220b/streamtube.js#L228-L231

cc @kig

src/lib/zip3.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2658/commits/5c2ddfc916d5895f416ebd0a98816fc87e244476

src/plots/gl3d/scene.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2658/commits/9a532508dcf4bdff899e647e1c1995f924271500

src/traces/streamtube/attributes.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2658/commits/22f0c7e5396518f8ed0d4fbf45257deb4fa3a7d2

src/traces/streamtube/convert.js
etpinard: New attempt in https://github.com/plotly/plotly.js/pull/2658/commits/83f3bdc71c3a3b20715f536f8f368e050656fa40

etpinard:
@alexcjohnson are any of your remarks in https://github.com/plotly/plotly.js/pull/2658#issuecomment-394836109 blocking for a v1 release?

@kig would you be interested in looking at the broken tube @alexcjohnson noticed in `gl3d_streamtube-first` ?

src/traces/cone/attributes.js
```diff
@@ -161,11 +160,13 @@ var attrs = {
     }
 };
 
-extendFlat(attrs, colorAttrs('', 'calc', true), {
-    showscale: colorscaleAttrs.showscale,
+extendFlat(attrs, colorscaleAttrs('', {
+    colorAttr: 'u/v/w norm',
```
etpinard: This give better attribute descriptions for cone `colorscale` and friends. For example, for `cmin` we have

> "Sets the lower bound of the color domain. Value should have the same units as the u/v/w norm and if set, `cmax` must be set as well."

alexcjohnson:
> are any of your remarks in https://github.com/plotly/plotly.js/pull/2658#issuecomment-394836109 blocking for a v1 release?

Just the broken tube - if there's a benign explanation for it then fine, I'd just like us to make sure we're not setting ourselves up for breaks in some broad class of conditions we haven't fully explored yet.

kig:
Right, two ideas of what might be causing this. 

The coordinate system generated based on the velocity vector is not continuous, so when a stream passes through a discontinuous angle, the coordinate axes suddenly change, which would cause a >< kink in the tube.

Alternative explanation: the divergence function returns a zero for some reason.

alexcjohnson:
Looking at that point again - I can confirm that it's not *broken*, it does all seem to connect ala pinched sausage links. But we're off the edge of the data here, the largest x in the data is 4 and we're almost at x=5 here. So I'm not sure what extrapolation is going on but I can't imagine it would take what looks like a fairly large and stable divergence and take it through zero before coming back out to essentially the same magnitude at the next point. Is there any chance it's taking two neighboring rings and rather than connecting them directly, it's connecting them inverted? Meaning it's making a mesh that connects points at the top of the tube on the first ring to points at the bottom of the next ring, and so on? I feel like I see some tight folding that makes it look like a wrapped present, which seems like it could happen with an inversion scenario like that:
screen shot 2018-07-03 at 10 01 06 pm

etpinard:
Ok merging this thing.

Go to https://github.com/plotly/plotly.js/issues/2781 for a list of open items.

kig:
> Is there any chance it's taking two neighboring rings and rather than connecting them directly, it's connecting them inverted?

Yeah, that's probably what happens. At some angles, the coordinate system generated from only the velocity vector suddenly swaps its axes. I can fix it by also sending the previous velocity vector, then the shader could compare the coordinate systems to avoid sudden swaps. Or do it in JS and pass up vectors to the shader.

Issue #2614, opened 2018-05-04
GL plots seem to be using up multiple GL contexts with Plotly 1.37.1


brian428:
Previously, we were able to show 7-8 or more WebGL plots at once. With 1.37.1, we're hitting the browser GL context limit (typically 16) much earlier. When showing 7 plots, the first two immediately lose their GL contexts and appear as blank.

Here is a repro case: http://jsfiddle.net/brian428/fqhb7kk9/

It seems like Plotly is using up multiple GL contexts for each scattergl plot. With 7 plots, there are 5 console warnings about losing contexts, so it appears that each scattergl plot is using **3** GL contexts for some reason. 

We need to show multiple plots together (and allow each one to be individually manipulated by the user), so this is a blocking issue. Can someone take a look?

etpinard:
Duplicate of https://github.com/plotly/plotly.js/issues/2333

brian428:
Thanks, I missed that one. Would it help you if add my repro case to that ticket?

brian428:
Also just to be clear, we had been on 1.35 I believe. So it seems like maybe something changed between 1.35 and 1.37 that made this worse. Maybe the addition of Splom?

etpinard:
@brian428 sorry for the wait. We just merged https://github.com/plotly/plotly.js/pull/2662 which makes `scattergl` graphs use 2 (not 3) WebGL contexts as was the case before `1.36.0`.

A little more info on the topic: we use two WebGL contexts per scattergl (and splom) graph to improve selection (via select-box and lasso) performance. Since `1.33.0` , WebGL contexts are shared across subplots on the same graph `<div>`. So, one way to decrease the number of created contexts would be to merge your graphs into subplots on the same graph `<div>` when possible. 

Now, ticket https://github.com/plotly/plotly.js/issues/2333 has some thoughts on sharing WebGL contexts across graph `<div>`s. Hopefully using an external library like [`multi-regl`](https://github.com/regl-project/multi-regl) will be enough. If not, we might have to add another API method e.g. `Plotly.dashboard` to solve this problem.

I'll close this issue. But please feel free to comment on https://github.com/plotly/plotly.js/issues/2333.

PR #2662, opened 2018-05-24
Create two (not three) WebGL contexts for scattergl/splom graphs

etpinard:
As described in https://github.com/plotly/plotly.js/issues/2614, starting in version 1.36.0 all regl based graphs created 3 WebGL contexts (one for each [canvas layer](https://github.com/plotly/plotly.js/blob/dfb56d3a7b99ecf979365ed9b9c2c40747cdd915/src/plot_api/plot_api.js#L212-L224)). But one of these layer (the _pick_ layer) is only used in parcoords traces, so we can create two contexts instead of three on scattergl and splom graphs. 

Why did this happen? I got sloppy in commit https://github.com/plotly/plotly.js/pull/2505/commits/469e4cdb2a045d57ca5e062cd01444a3756f5fd0 where I failed to bring [this `d.pick` check](https://github.com/plotly/plotly.js/commit/469e4cdb2a045d57ca5e062cd01444a3756f5fd0#diff-e03c5a2a46f9392088513c32645ea9d5L341) into `prepareRegl`.

Reducing the number of WebGL context created does two things:

- allows more scattergl/splom graphs to be displayed on a given page before hitting the browsers' WebGL context limit (= 16 on recent Chrome versions)
- speed up first render benchmarks by ~40ms, which is significant (> 15%) on graphs with 1e5 points or less.

cc @alexcjohnson @dy

### DIFF
--- a/test/jasmine/tests/gl2d_plot_interact_test.js
+++ b/test/jasmine/tests/gl2d_plot_interact_test.js
@@ -8,7 +8,7 @@ var ScatterGl = require('@src/traces/scattergl');
 
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
-var fail = require('../assets/fail_test');
+var failTest = require('../assets/fail_test');
 var mouseEvent = require('../assets/mouse_event');
 var touchEvent = require('../assets/touch_event');
 var drag = require('../assets/drag');
@@ -202,7 +202,7 @@ describe('@gl Test gl plot side effects', function() {
         .then(function() {
             expect(gd.querySelector('.gl-canvas-context').width).toBe(300);
         })
-        .catch(fail)
+        .catch(failTest)
         .then(done);
     });
 });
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -945,7 +945,21 @@ describe('@gl Test gl2d plots', function() {
             expect(opts.positions)
                 .toBeCloseToArray([1, 1, 2, 2, 3, 1]);
         })
-        .catch(fail)
+        .catch(failTest)
+        .then(done);
+    });
+
+    it('should create two WebGL context per graph', function(done) {
```
alexcjohnson: contexts

alexcjohnson:
Nice, glad it was that easy! 💃

test/jasmine/tests/gl2d_plot_interact_test.js
etpinard: done -> a3dce12

Issue #2667, opened 2018-05-25
Selection vs Pan double-click bug


nicolaskruchten:
The following sequence exposes what looks like a bug:

1. Render a plot with https://codepen.io/nicolaskruchten/pen/QrXYbJ for the color settings
2. Click on the box-select tool
3. Select a point
4. Click on the pan tool
5. Pan
6. Double-click on the plot
7. For me this resets the selection and not the pan.

alexcjohnson:
you can even omit 5. Pan

etpinard:
Ok. On it.

etpinard:
Bug introduced in 1.36.0. Working fine in 1.35.2: https://codepen.io/etpinard/pen/aGgXjJ?editors=1010

PR #2668, opened 2018-05-25
Fix select <-> pan double-click behavior

etpinard:
Commit https://github.com/plotly/plotly.js/commit/20820cb70c4879c91e8ade621c829f1b5ac4c069 fixes https://github.com/plotly/plotly.js/issues/2667 which has been broken since https://github.com/plotly/plotly.js/pull/2527/commits/bb022818e7fe6c2c808ad440a633558398853325 (from one of the splom perf PRs) where drag handlers are no-longer reintialized when relayout'ing `dragmode`.

cc @alexcjohnson


alexcjohnson:
Good catch! 💃

Issue #2400, opened 2018-02-23
NPM modules for dists


nicolaskruchten:
Idea: we could have an NPM module called e.g. `plotly.js-dist` which contained only precompiled bundles. Or we could have an NPM module per bundle. These modules would have zero dependencies so they would never throw warnings on installation and they would download really fast and require no special Webpack loaders or Browserify transforms so they would work smoothly with create-react-app or Next.js or whatever comes next :)

etpinard:
Great idea. 

I'll even suggest making one `npm` registry per partial bundle that is `plotly.js-basic`, `plotly.js-cartesian`, `plotly.js-geo`, `plotly.js-gl3d`, `plotly.js-gl2d`, `plotly.js-mapbox`, `plotly.js-finance` that would only include one `.js` and one `.min.js` file.

etpinard:
Related: https://github.com/plotly/plotly.js/issues/768

PR #2670, opened 2018-05-25
NPM pkg for dist bundles

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2400

A new `sync_pacakges.js` script assembles "sub plotly.js packages" in the git-ignored `build/` folder that include a minimal package.json and README.md as well as one single "bundle" file. Along with the main dist/plotly.js bundle, each partial bundle has it own "sub plotly.js package".

After running, `node tasks/sync_pacakges.js`, we get:

image

A sample package.json looks like:

image

and a sample README.md looks like:

image

Some questions left to answer:

- Should we add the minified bundles to each sub package? **Answer:** no.
- Should we add `dist/topojson/` to all sub packages?
- What to do with `dist/extras/mathjax/`? Should we include it too?

@nicolaskruchten what do you think?


etpinard:
TODO: 

- [ ] we'll need to patch `tasks/stats.js` to make `dist/README.md` know about these npm "sub" packages
- [ ] patch up the root README too.

nicolaskruchten:
IMO no need to add minified versions. The main advantage I see for these packages is for people to import plotly.js in a more friendly way into their build systems like webpack and friends. In my experience, it's unhelpful to load minified files into such systems as it messes with their downstream minification.

etpinard:
> Feels to me like we may want to include topojson anyway since otherwise they'd have to include the full plotly.js anyway just to get that, right?

We could alternatively add `plotly.js-topojson` npm package.

cpsievert:
I would vote in favor of minified bundles if it's not too much overhead. It'd be much easier for me to leverage partial bundles in the R package that way.

nicolaskruchten:
(no objections to minified bundles if it's helpful for the R world!)

etpinard:
Just curious @cpsievert, why would an npm package be easier for you than downloading the `min.js` file from https://cdn.plot.ly/plotly-latest.min.js ?

cpsievert:
@etpinard do we publish minified partials to the cdn (e.g., https://cdn.plot.ly/plotly-basic.min.js doesn't exist)? 

I don't currently leverage partial builds in R, and I could do it via github releases, but it'd be nice to have a lighter-weight solution as well.

etpinard:
> do we publish minified partials to the cdn (e.g., https://cdn.plot.ly/plotly-basic.min.js doesn't exist)?

yes -> https://cdn.plot.ly/plotly-basic-latest.min.js or https://cdn.plot.ly/plotly-basic-1.38.1.min.js

etpinard:
All you want to know about partial bundle is here -> https://github.com/plotly/plotly.js/tree/master/dist#plotlyjs-basic

cpsievert:
Ahhh, ok, I use the CDN then, sorry for the noise!

BUILDING.md
```diff
@@ -1,5 +1,7 @@
 # Building plotly.js
 
+The easiest way to bundle plotly.js into your application is to use the one of the distributed plotly.js packages on npm. These distributed packages should just work with **any** build framework. That said, if you're looking to save a few bytes, read the section below corresponding to your building framework.
```
alexcjohnson: use ~the~ one of the distributed...

Do you want to include a link that points to a list of these modules on npm? I guess we can add that later once it actually exists ;)

README.md
```diff
@@ -32,48 +32,45 @@ and more.
...
+Starting in `v1.15.0`, plotly.js ships with several _partial_ bundles (more info [here](https://github.com/plotly/plotly.js/blob/master/dist/README.md#partial-bundles)).
+
+Starting in `v1.39.0`, plotly.js publishes _distributed_ npm packages with no dependencies. For example, run `npm install plotly.js-geo-dist` and add `import Plotly from 'plotly.js-geo-dist';` to your code to start using the plotly.js geo package.
 
-If you would like to manually pick which plotly.js modules to include, you can create a *custom* bundle by using `plotly.js/lib/core`, and loading only the trace types that you need (e.g. `pie` or `choropleth`). The recommended way to do this is by creating a *bundling file*:
+If you would like to manually pick which plotly.js modules to include, you'll first need to run `npm install plotly.js` and then create a *custom* bundle by using `plotly.js/lib/core`, and loading only the trace types that you need (e.g. `pie` or `choropleth`). The recommended way to do this is by creating a *bundling file*. For example, in CommonJS:
```
nicolaskruchten: I might prefix this with something like "If none of the distributed NPM  packages meet your needs, and you would like to..." to encourage people to go with the flow :)

nicolaskruchten:
What will the NPM module page look like for these new modules? they'll just inherit the same readme, github URL and homepage URL as the main module right? so basically the npmjs.org pages will look the same as the current one except for the callouts here?

image

etpinard:
> What will the NPM module page look like for these new modules?

Like their READMEs e.g.:

https://github.com/etpinard/plotly.js/blob/npm-pkgs-for-dist-bundles-build/build/plotly.js-cartesian-dist/README.md

README.md
```diff
@@ -32,48 +32,45 @@ and more.
...
 ```
 
-and use the plotly.js `dist` file(s).
+and use the `Plotly` object in the window scope.
 
 #### Install with `npm`
 
 ```bash
-npm install plotly.js
+npm install plotly.js-dist
```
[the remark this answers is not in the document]
etpinard: Personally, using the CDN is still the "quickest" way to start using plotly.js (hence its first spot in the "Quick start options" section), but ok.

README.md
etpinard: I think we should encourage people that want to save bytes, to save bytes, but ok.

nicolaskruchten:
Re the README how are you telling NPM that it should use that readme instead of the main one? AFAIK it just grabs the readme from the `repository` field in `package.json` .. ?

etpinard:
> AFAIK it just grabs the readme from the repository field in package.json .. ?

Oh really? I thought it grabbed the package.json from where we run `npm publish`. Let me double-check.

etpinard:
Thanks for bringing this up @nicolaskruchten !

We have to list the files manually in this case, like `react`'s mono-repo does e.g. for [`react-dom`](https://github.com/facebook/react/blob/15767a8f8f798f0d80df4ddbc7cfaebdc29996a8/packages/react-dom/package.json#L24-L34).

josemarluedke:
Are there any plans to get this merged in and released?

alexcjohnson:
> Are there any plans to get this merged in and released?

Yes, it's ready to go but because it will result in substantial changes to the documentation we want to merge it immediately before releasing v1.39.0. That should happen within the next week.

Issue #2669, opened 2018-05-25
Selection box stays after panning


destradafilm:
I see an issue when I draw a selection box on the points on a scattergl plot, and then I pan.
 
After I select a range of points and then I use the modifier key 'alt-select' (to subract) or 'shift-select' (to add) I can see the previously drawn selection box and it has not moved even though the plot has shifted range. 

EXAMPLE:
http://recordit.co/Z6r2ZPUMoo

Also, I can't find documentation about the modifier key 'add-select' or 'subtract-selection' tools. 
Is there a dragmode: 'select-add' or a way to programmatically use these dragmodes?

etpinard:
Ha. Nice catch @destradafilm thanks for the report.

This bug is similar to https://github.com/plotly/plotly.js/issues/2667 (the optimizations of https://github.com/plotly/plotly.js/commit/bb022818e7fe6c2c808ad440a633558398853325 are likely to blame) but will require another patch.

I'll try to squeeze that one in before 1.38.1 .

PR #2676, opened 2018-05-28
Fix shift selection after pan

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2669

before:

peek 2018-05-28 17-51

now:

peek 2018-05-28 17-52

cc @alexcjohnson

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -178,23 +178,23 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             prepSelect(e, startX, startY, dragOptions, dragModeNow);
         } else {
             dragOptions.clickFn = clickFn;
+            clearAndResetSelect();
 
-            if(allFixedRanges) {
-                clearAndResetSelect();
-            } else if(dragModeNow === 'zoom') {
-                dragOptions.moveFn = zoomMove;
-                dragOptions.doneFn = zoomDone;
+            if(!allFixedRanges) {
+                if(dragModeNow === 'zoom') {
+                    dragOptions.moveFn = zoomMove;
+                    dragOptions.doneFn = zoomDone;
 
-                // zoomMove takes care of the threshold, but we need to
-                // minimize this so that constrained zoom boxes will flip
-                // orientation at the right place
-                dragOptions.minDrag = 1;
+                    // zoomMove takes care of the threshold, but we need to
+                    // minimize this so that constrained zoom boxes will flip
+                    // orientation at the right place
+                    dragOptions.minDrag = 1;
 
-                zoomPrep(e, startX, startY);
-            } else if(dragModeNow === 'pan') {
-                dragOptions.moveFn = plotDrag;
-                dragOptions.doneFn = dragTail;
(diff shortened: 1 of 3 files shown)


src/plots/cartesian/dragbox.js
```diff
@@ -178,6 +178,8 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             prepSelect(e, startX, startY, dragOptions, dragModeNow);
         } else {
             dragOptions.clickFn = clickFn;
+            // clear selection polygon cache (if any)
+            plotinfo.selection = false;
```
alexcjohnson: Should this be part of `clearSelect`? That gets called in all 3 blocks below (directly in the first and third, and in the second it's inside `zoomPrep` - looks like these can be simplified anyway) but also in [`zoomWheel`](https://github.com/plotly/plotly.js/blob/7784a10880545ef79b5dbc269d5450f337c6807d/src/plots/cartesian/dragbox.js#L397) - does that case have the same bug?

etpinard: Ah. Good catch. Zoom has the problem. 

Working on a fix :pick:

src/plots/cartesian/dragbox.js
```diff
@@ -393,9 +393,7 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             return;
         }
 
-        if(redrawTimer === null) {
-            clearSelect(zoomlayer);
-        }
+        clearSelect(zoomlayer);
```
etpinard: this fixes what I botched in https://github.com/plotly/plotly.js/commit/51c3128718ff61ccaad259a7ff65b76c06b2d2a3

etpinard: Oh wait, this worked fine in https://github.com/plotly/plotly.js/commit/51c3128718ff61ccaad259a7ff65b76c06b2d2a3

It is the optimization of https://github.com/plotly/plotly.js/commit/bb022818e7fe6c2c808ad440a633558398853325 that broke this (except during the very first scroll event),

src/plots/cartesian/dragbox.js
etpinard: fixed and :lock: in https://github.com/plotly/plotly.js/pull/2676/commits/f682fb8979da56083c4a513fe7cbe642f323a2e3

src/plots/cartesian/dragbox.js
```diff
@@ -196,11 +194,18 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             } else if(dragModeNow === 'pan') {
                 dragOptions.moveFn = plotDrag;
                 dragOptions.doneFn = dragTail;
-                clearSelect(zoomlayer);
+                clearAndResetSelect();
```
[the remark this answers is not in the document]
etpinard: Ooops, I missed the one in `zoomPrep`. Cleaned up in -> https://github.com/plotly/plotly.js/pull/2676/commits/34a0cf7faefa75caffd3c572b80209f9ddc6303d

PR #2681, opened 2018-05-30
Automargin fixes

### LINKED ISSUE #2651: React margin problems
Split from https://github.com/plotly/plotly.js/issues/2624, with reproducible example and explanations in: https://codepen.io/etpinard/pen/vjaGLW

cc @alexcjohnson

alexcjohnson:
Fixes #2651 - and a bunch of other things I encountered along the way:
- Stops writing `trace.uid` back to the user input data, instead reusing the values from `oldFullData` if the user didn't specify `uid` in the trace. One more unexpected mutation gone 🎉 but more importantly, `Plotly.react` users were probably trashing this anyway when making a new figure state, so they were triggering a more aggressive redraw than was really necessary.
- Significant 🌴 for colorbars, as well as making them work in some cases where they should (splom, scatterpolar[gl]) or worked but broke on editing (parcoords, histogram)
- Automargin logic is now a bit more encapsulated: for example `clearPushMargins` we used to need in rangesliders and updatemenus (and margin clearing by missing colorbars) are no longer necessary: we start a replot by clearing the list of allowed margin pusher ids, then delete disallowed ids when actually calculating the margins. Edits that don't go through the full replot can still adjust the margins by changing, adding, or deleting pushers, but they don't need to worry about deleting their old pushers on a full replot.

The automargin system is still a mess in some fundamental ways that I didn't address here - really we need to break it out from the drawing steps of all of these components, so we can calculate the margins once and for all, then draw everything; as opposed to drawing, noticing we need to change the margins, then drawing all over again... but I think this is a step in the right direction.

cc @etpinard


src/components/colorbar/connect.js
```diff
@@ -0,0 +1,85 @@
...
+    var vals;
+    if(Array.isArray(valAttr)) {
+        for(var i = 0; i < valAttr.length; i++) {
+            vals = container[valAttr[i]];
+            if(vals) break;
+        }
+    }
+    else vals = container[valAttr];
+
+    if(vals) {
```
etpinard: (Non-blocking) I'm curious to see what would happen (i.e. which tests would fail if any) if we drop the logic around vals here. cmin/cmax or zmin/zmax should always be computed during the calc step which should happen before this block.

test/jasmine/tests/legend_test.js
```diff
@@ -524,18 +524,36 @@ describe('legend relayout update', function() {
     afterEach(destroyGraphDiv);
 
     it('should hide and show the legend', function(done) {
-        var mockCopy = Lib.extendDeep({}, mock);
```
etpinard: Is this commit locking down fixes from this PR, or just improving our test case coverage?

src/components/colorbar/connect.js
```diff
@@ -0,0 +1,85 @@
...
+ *   name of the attribute holding the value of the minimum color
+ * @param {string} moduleOpts.max
+ *   name of the attribute holding the value of the maximum color
+ * @param {Optional(string)} moduleOpts.vals
+ *   name of the attribute holding the (numeric) color data
+ *   used only if min/max fail. May be omitted if these are always
+ *   pre-calculated.
+ */
+module.exports = function connectColorbar(gd, cd, moduleOpts) {
+    if(typeof moduleOpts === 'function') return moduleOpts(gd, cd);
```
etpinard: So, contour still gets its own version on `module._colorbar` in 

https://github.com/plotly/plotly.js/blob/fa8614750586a2901ccf65a520521bb570eb7026/src/traces/contour/colorbar.js

etpinard: ... and that's the only exception?

src/plot_api/plot_api.js
```diff
@@ -252,18 +253,24 @@ exports.plot = function(gd, data, layout, config) {
         var calcdata = gd.calcdata;
         var i, cd, trace;
 
-        Registry.getComponentMethod('legend', 'draw')(gd);
-        Registry.getComponentMethod('rangeselector', 'draw')(gd);
-        Registry.getComponentMethod('sliders', 'draw')(gd);
-        Registry.getComponentMethod('updatemenus', 'draw')(gd);
+        // First reset the list of things that are allowed to change the margins
```
etpinard: (for documentation purposes only) this patch is in `function marginPushers() {}`

src/plot_api/plot_api.js
```diff
@@ -354,6 +361,10 @@ exports.plot = function(gd, data, layout, config) {
         initInteractions,
         Plots.addLinks,
         Plots.rehover,
+        // TODO: shouldn't really need doAutoMargin here, but without it some
+        // edits don't manage to tweak the margins the way they should
+        // in the tests, the only one that presently breaks is axis.position
+        Plots.doAutoMargin,
```
etpinard: This is a little unfortunate. This could be a big performance hit. Would it be too much to ask to try to remove this in this PR?

etpinard: If only axis.position is affected, we could add a special (hopefully temporary) relayout flag to call `Plots.doAutoMargin` after `finalDraw`

src/plots/plots.js
```diff
@@ -1076,10 +1079,10 @@ plots.supplyFrameDefaults = function(frameIn) {
     return frameOut;
 };
 
-plots.supplyTraceDefaults = function(traceIn, colorIndex, layout, traceInIndex) {
+plots.supplyTraceDefaults = function(traceIn, colorIndex, layout, traceInIndex, uid) {
     var colorway = layout.colorway || Color.defaults;
-    var traceOut = {},
-        defaultColor = colorway[colorIndex % colorway.length];
+    var traceOut = {uid: uid};
```
etpinard: Or alternatively, I think we could move this line:

image

to just before this line:

image

etpinard: the above comment is non-blocking.

etpinard:
@alexcjohnson Thanks for taking this on. Auto-margin and colorbars were very much in need of some :heart: .

I'm a little concerned with:

- the additional `doAutoMargin` call in `Plotly.plot` - https://github.com/plotly/plotly.js/pull/2681#discussion_r191894083
- the additional `drawMarginPushers` call during axrange relayouts - https://github.com/plotly/plotly.js/pull/2681#discussion_r191877868
- updatemenus / sliders asymmetries - https://github.com/plotly/plotly.js/pull/2681#discussion_r191879632

My other comments aren't as important.

src/components/updatemenus/draw.js
```diff
@@ -63,10 +67,15 @@ module.exports = function draw(gd) {
         .classed(constants.containerClassName, true)
         .style('cursor', 'pointer');
 
-    menus.exit().remove();
-
-    // remove push margin object(s)
-    if(menus.exit().size()) clearPushMargins(gd);
+    menus.exit().each(function() {
+        // Most components don't need to explicitly remove autoMargin, because
+        // marginPushers does this - but updatemenu updates don't go through
```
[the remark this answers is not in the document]
alexcjohnson: Oh man, I missed sliders in this PR but sure enough, they still have the older pattern - thanks, I'll take a look at those.

src/plot_api/plot_api.js
alexcjohnson: I'll take a look again, didn't really understand why it was needed so I wasn't confident it was *only* `ax.position` - that's just the only one that we test directly that triggered it. Anyway it's a very minor performance hit if the margins *don't* change at that point.

etpinard: >  Anyway it's a very minor performance hit if the margins don't change at that point.

Oh right, `doAutoMargin` is smart<sup>TM</sup> like that. This additional `doAutoMargin` isn't a big then.

src/plot_api/plot_api.js
```diff
@@ -1718,6 +1717,22 @@ exports.relayout = function relayout(gd, astr, val) {
...
+    // executed after drawData
+    var doTicks = rangesAltered ?
+        function(gd) { return subroutines.doTicksRelayout(gd, rangesAltered); } :
+        subroutines.doTicksRelayout;
+
+    seq.push(
+        doTicks,
+        subroutines.drawData,
+        subroutines.finalDraw,
+        subroutines.drawMarginPushers
```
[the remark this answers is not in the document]
alexcjohnson: Good Q, can't remember now why I needed that, I'll take another look at it...

src/traces/cone/colorbar.js
```diff
@@ -20,10 +19,7 @@ module.exports = function colorbar(gd, cd) {
 
     gd._fullLayout._infolayer.selectAll('.' + cbId).remove();
 
-    if(!trace.showscale) {
-        Plots.autoMargin(gd, cbId);
-        return;
-    }
+    if(!trace.showscale) return;
```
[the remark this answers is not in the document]
alexcjohnson: No, it's in this commit - the `autoMargin` I removed here is just the one to clear the entry for this colorbar, but that's not needed anymore due to `plots.clearAutoMarginIds`

src/components/colorbar/connect.js
alexcjohnson: Yes, contour and those that inherit from it (`contourcarpet`, `contourgl`, `histogram2dcontour`).

I should note that there's one other place `module.colorbar` gets used, that's in [`colorbar.draw`](https://github.com/plotly/plotly.js/pull/2681/commits/fa8614750586a2901ccf65a520521bb570eb7026#diff-ad110e994c21bb045d288b9c96e4e6ccR452) for getting the `nestedProperty` string for title editing. If contour colorbars weren't at the module root we'd need to set a `container` property of the *function* which would be a bit hacky... alternatively we could either a) make this function be `module.colorbar.func`, or b) incorporate contour into this routine directly, triggered by some other property of `module.colorbar`.

test/jasmine/tests/legend_test.js
alexcjohnson: Pretty sure at least one of these was broken before this PR, but I don't recall which anymore.

src/plots/plots.js
alexcjohnson: yeah probably - but I thought it would be better in `supplyDefaults` so we don't add another order-dependence to the `plot` step - carpet dependents need this too. I think they're already required to `plot` after the carpet for other reasons but still...

etpinard: > but I thought it would be better in supplyDefaults so we don't add another order-dependence to the plot step - carpet dependents need this too.

Good point. Your way is :ok_hand:

src/components/colorbar/connect.js
alexcjohnson: yeah, looks like calculating the range while drawing the colorbar is obsolete, good call! Though I can't seem to run the image tests all the way through locally anymore, so we'll see in a few minutes -> https://github.com/plotly/plotly.js/pull/2681/commits/f9fbbc647cad99497514c8a575c111b7222f8d73

src/traces/cone/colorbar.js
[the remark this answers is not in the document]
etpinard: Thanks for headsup :ok_hand:

src/components/colorbar/connect.js
etpinard: > alternatively we could either a) make this function be module.colorbar.func

This sounds cleaner to me. But no need to do this in this PR.

src/components/colorbar/connect.js
etpinard: image

:white_check_mark: nice!

test/jasmine/assets/custom_assertions.js
```diff
@@ -216,3 +216,29 @@ exports.assertElemInside = function(elem, container, msg) {
...
+ *   heightLessThan (height must be less than this)
+ */
+exports.assertPlotSize = function(opts, msg) {
+    var width = opts.width;
+    var height = opts.height;
+    var widthLessThan = opts.widthLessThan;
+    var heightLessThan = opts.heightLessThan;
+
+    var actualWidth = d3.select('.ygrid').node().getBoundingClientRect().width;
+    var actualHeight = d3.select('.xgrid').node().getBoundingClientRect().height;
```
etpinard: (non-blocking) but wouldn't `d3.select('.subplot.xy > .plot').node().getBoundingClientRect` be more general (e.g. would work even on graphs with `showgrid: false`)?

alexcjohnson: Hmm, good point, but `.plot` shrinks to the data, which may not reach the edges of the plot area (or could even go beyond with `cliponaxis: false`). Possibly `.bglayer .bg` though...

etpinard: > or could even go beyond with cliponaxis: false

Yeah, good point here. Though problem it turns out.

src/plots/plots.js
```diff
@@ -1603,11 +1603,21 @@ function setupAutoMargin(fullLayout) {
...
+ * @param {string} id - an identifier unique (within this plot) to this object,
+ *     so we can remove a previous margin expansion from the same object.
+ * @param {object} o - the margin requirements of this object, or omit to delete
+ *     this entry (like if it's hidden). Keys are:
+ *     x, y: plot fraction of the anchor point.
+ *     xl, xr, yt, yb: if the object has an extent defined in plot fraction,
+ *         you can specify both edges as plot fractions in each dimension
+ *     l, r, t, b: the pixels to pad past the plot fraction x[l|r] and y[t|b]
+ *     pad: extra pixels to add in all directions, default 12 (why?)
+ */
```
alexcjohnson: I mentioned this in the commit message:

> Also includes a generalization of `Plots.autoMargin` so an object can
have extent in plot fraction, as opposed to just position in plot fraction
and extent in pixels. We could use this to allow for example colorbars
to be sized based on the actual margins, not the pre-margin-pushing plot size
as they are now, see note at colorbar/draw:119-124

but for more detail: colorbars (and possibly a few others?) when sized as plot fraction they use the initial margins to convert plot fraction to pixels - meaning that if the margins expand beyond that, a colorbar with length "1" would actually be taller than the plot area. This is probably fine if it's the colorbar itself that's expanding the margin - that means it's off-center and thus not aligned with the plot, so who really cares if it's a little bigger than the plot? But if something *else* (eg a rangeslider, or axis labels) expanded the margin, all of a sudden the colorbar is misaligned. Using `yt/yb` we can size it correctly without getting into an infinite loop of automargin calls. Actually now that I think of the case ^^ which would be pretty ugly, I may try to do that in this PR as well.

src/plots/plots.js
```diff
@@ -1632,7 +1632,13 @@ plots.autoMargin = function(gd, id, o) {
...
         }
         else {
-            var pad = o.pad === undefined ? 12 : o.pad;
+            var pad = o.pad;
+            if(pad === undefined) {
+                var margin = fullLayout.margin;
+                // if no explicit pad is given, use 12px unless there's a
+                // specified margin that's smaller than that
+                pad = Math.min(12, margin.l, margin.r, margin.t, margin.b);
+            }
```
alexcjohnson: @etpinard seem reasonable? I started looking into this due to test failures at https://github.com/plotly/plotly.js/pull/2681/commits/d02c76b3d1082996fa1c2543d7c8f08e08a256c0 - where some colorbars that were previously not able to expand the margin because they were internally pixel-sized suddenly became able to expand it when I switched it to plot-fraction-sized as described in https://github.com/plotly/plotly.js/pull/2681#discussion_r192409090... but then this 12px padding overrode user-supplied margins that were smaller - and that didn't seem right to me. Some of the baseline images with small explicit margins changed as a result.

The 12px here is an odd hard-coded feature - I suppose an alternative would be to make this an attribute in its own right (`layout.margin.autopad`?) and put the `min(12, margin.l...)` logic in `supplyDefaults`.

test/jasmine/assets/custom_assertions.js
alexcjohnson: switched to `.bglayer .bg` in https://github.com/plotly/plotly.js/pull/2681/commits/bb6c7201820468747976de41baa90f65a08eb4ee though this has a caveat about `margin.pad` https://github.com/plotly/plotly.js/pull/2681/commits/5d3132423a1a4a347528d80feb03ba6ad7bc356e. I still think it's an improvement: `margin.pad` isn't used very much and can be corrected for if you really want, and the bg rect is always present.

src/components/updatemenus/draw.js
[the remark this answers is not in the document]
alexcjohnson: Updated sliders in https://github.com/plotly/plotly.js/pull/2681/commits/3deab4c4aefeac557058800a46707d5669b0a334

src/plot_api/plot_api.js
alexcjohnson: On further investigation, this `doAutoMargin` call is definitely necessary for now. The reason is that axis automargin is not part of `marginPushers` - they reserve the *right* to push the margins there now with `Axes.allowAutoMargin` but the actual adjustment comes later during `Axes.doTicks`.

As discussed above, this `doAutoMargin` call won't actually redraw anything if the margins didn't change, so there's usually a minimal cost to this, but when the margins *do* change this triggers a full replot.

Long-term it would be far better to completely separate out margin calculations from drawing, so we *never* need to draw multiple times for a single change. But this addition doesn't represent a performance hit relative to the status quo.

src/components/sliders/draw.js
```diff
@@ -36,10 +36,20 @@ module.exports = function draw(gd) {
...
         .style('cursor', 'ew-resize');
 
-    sliders.exit().remove();
+    function clearSlider(sliderOpts) {
+        if(sliderOpts._commandObserver) {
+            sliderOpts._commandObserver.remove();
+            delete sliderOpts._commandObserver;
+        }
+
+        Plots.autoMargin(gd, autoMarginId(sliderOpts));
```
etpinard: Could we add a comment similar to what's above https://github.com/plotly/plotly.js/pull/2681/files#r191879632 here?

src/plot_api/plot_api.js
etpinard: > The reason is that axis automargin is not part of marginPushers - they reserve the right to push the margins there now with Axes.allowAutoMargin but the actual adjustment comes later during Axes.doTicks.

Thanks for writing this down! Could update that comment in the code above this line to reflect these new findings?

> Long-term it would be far better to completely separate out margin calculations from drawing, so we never need to draw multiple times for a single change. But this addition doesn't represent a performance hit relative to the status quo.

The _dream_.

etpinard: Oh and about

> Long-term it would be far better to completely separate out margin calculations from drawing, so we never need to draw multiple times for a single change. But this addition doesn't represent a performance hit relative to the status quo.


Is there an issue about this already? If not, we should open one.

src/components/colorbar/draw.js
```diff
@@ -539,14 +540,35 @@ module.exports = function draw(gd, id) {
...
+            var marginOpts = {};
+            var tFrac = FROM_TL[opts.yanchor];
+            var bFrac = FROM_BR[opts.yanchor];
+            if(opts.lenmode === 'pixels') {
+                marginOpts.y = opts.y;
+                marginOpts.t = outerheight * tFrac;
+                marginOpts.b = outerheight * bFrac;
+            }
+            else {
+                marginOpts.t = marginOpts.b = 0; // TODO - not zero?
```
etpinard: Could you expand on that `TODO` comment?

src/plots/plots.js
etpinard: > seem reasonable?

Very reasonable. :1st_place_medal: 

> I suppose an alternative would be to make this an attribute in its own right (layout.margin.autopad?) and put the min(12, margin.l...) logic in supplyDefaults.

Maybe down the road, but no need in this PR.

src/components/colorbar/draw.js
alexcjohnson: Ah thanks, that was just for me to check during development, it's correct as zeros -> removed in https://github.com/plotly/plotly.js/pull/2681/commits/230a782766ae9864eb5ac803c5fb069bedb9ab25

src/components/sliders/draw.js
alexcjohnson: sure -> https://github.com/plotly/plotly.js/pull/2681/commits/230a782766ae9864eb5ac803c5fb069bedb9ab25

src/plot_api/plot_api.js
alexcjohnson: comments improved in https://github.com/plotly/plotly.js/pull/2681/commits/230a782766ae9864eb5ac803c5fb069bedb9ab25, and new issue #2704 to track it.

etpinard:
Awesome :dancer:

PR #2682, opened 2018-05-30
Fix for number 0 treated as invalid in hover (#1)

### LINKED ISSUE #2660: Tooltip text not working when text is 0 (number)
[Example that displays the error](https://codesandbox.io/s/oj90qzq026)

I want to be able to get a value from the text property in the data object. I noticed that if the value is the number 0 it won't be displayed in the plot. 

```javascript
          text: [0, 0, 0, 0, 0],
          hoverinfo: "x+text+name",
```

The code above will make the tooltip only show x and name. The text property won't be visible.
image

All other values except 0 will render properly, as can be seen in the code example linked above. Also if all the numbers are strings they will be shown correctly.

I guess that there is a null-check somewhere that treats 0 as false and therefore won't add it to the tooltip.

dalle:
* treat number 0 as valid in hover fixes #2660

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -650,7 +650,7 @@ drawing.textPointStyle = function(s, trace, gd) {
         var p = d3.select(this);
         var text = Lib.extractOption(d, trace, 'tx', 'text');
 
-        if(!text) {
+        if(!text && text !== 0) {
             p.remove();
             return;
         }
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Thanks @dalle - looks perfect! 💃

etpinard:
sweet PR @dalle thanks very much :tada:

dalle:
@alexcjohnson @etpinard Thank you for all the praise.

PR #2683, opened 2018-05-30
improve docs on loading and applying locales

cpsievert:


dist/README.md
```diff
@@ -36,25 +36,32 @@ You can grab the relevant MathJax files in `./dist/extras/mathjax/`.
 ### To include localization
 
 Plotly.js defaults to US English (en-US) and includes British English (en) in the standard bundle.
-Many other localizations are available - here is an example using Swiss-German (de-CH),
-see the contents of this directory for the full list.
-They are also available on our CDN as https://cdn.plot.ly/plotly-locale-de-ch-latest.js OR https://cdn.plot.ly/plotly-locale-de-ch-1.38.1.js
-Note that the file names are all lowercase, even though the region is uppercase when you apply a locale.
+This example uses Swiss-German (de-CH), but many other localizations are available (see the contents of this directory for the full list). All available locales have working date localizations, but only a subset will localize on-graph text (e.g., modebar controls). See [here](https://github.com/plotly/plotly.js/pulls?q=is%3Apr+label%3A%22type%3A+translation%22+is%3Aclosed) for a list of "fully-supported" locales.
```
etpinard: Perhaps even better, we could [ask translation authors](https://github.com/plotly/plotly.js/blob/master/.github/PULL_REQUEST_TEMPLATE.md#translations) to add an entry in a table of "fully" supported localizaion instead of linking to a list of PRs/

dist/README.md
```diff
@@ -36,25 +36,32 @@ You can grab the relevant MathJax files in `./dist/extras/mathjax/`.
...
 
-*After* the plotly.js script tag, add:
+#### Loading and registering locales
+
+Some locales build on other locales (e.g., "de-CH" builds on "de"). In this case, first load the base locale (e.g., "de"), then load the actual locale (e.g., "de-CH"):
 
 ```html
+<script src="plotly-locale-de.js"></script>
```
etpinard: I think this section is misleading. a priori `de-CH` doesn't _depend_ on `de`. At the moment, `de-CH` fallbacks to `de` just because `de-CH` isn't "fully" supported.

etpinard: So, I propose we use a "fully" supported language for this example and invite folks we want to use a non-fully supported localization (e.g. `de-CH`) to contribute by making a PR.

etpinard:
I'd vote for making all locale dist files (base and regional) standalone, that way users only have to import one file to localize their app. 

The only drawback I can think is for app that want to display multiple locales. In this case, importing a base locale along with its regional variants will lead to a few duplicated kBs.

alexcjohnson:
> I'd vote for making all locale dist files (base and regional) standalone, that way users only have to import one file to localize their app.
>
> The only drawback I can think is for app that want to display multiple locales. In this case, importing a base locale along with its regional variants will lead to a few duplicated kBs.

What if we do this in code? ie have `lib/locales/de-ch.js` be:
```js
module.exports = Lib.extendDeep({}, require('de'), {
    name: 'de-CH',
    format: {
        decimal: '.',
        thousands: ','
    }
});
```

Then `dist/plotly-locale-de-ch.js` could contain the output of that code, so it would automatically get updates from `de`. Users providing both as script tags would still have to carry the extra bytes, but the most byte-conscious users make custom bundles, and a bundle containing both would *not* have any extra bytes.

etpinard:
> What if we do this in code? ie have lib/locales/de-ch.js be:
> so it would automatically get updates from de. Users providing both as script tags would still have to carry the extra bytes, but the most byte-conscious users make custom bundles, and a bundle containing both would not have any extra bytes

Amazing. :1st_place_medal: 

This doesn't solve the `pr-br` vs `pt` case though. We should try to find a `pt` speaker.

alexcjohnson:
> This doesn't solve the `pr-br` vs `pt` case though. We should try to find a `pt` speaker. 

That's a (hopefully) uncommon case, where [3/4 of the language's speakers are in a single non-base region](https://en.wikipedia.org/wiki/List_of_territorial_entities_where_Portuguese_is_an_official_language) (and 20x the number in the base region) - I have no idea how different the two are, but I think it's OK to leave as is for now - hopefully a user from Portugal will come along and make us a `pt` and then we can collapse the two if they *do* have enough overlap that it's justified.

etpinard:
Moved to https://github.com/plotly/plotly.js/issues/2991

Issue #2688, opened 2018-06-01
restyle textposition produces javascript error


kochelmonster:
Try out the following example and click on "No Text" link.
```html
<head>
  <!-- Plotly.js -->
  <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
</head>

<body>

  <div id="myDiv"><!-- Plotly chart will be drawn inside this DIV --></div>
  <script>
var xValue = ['Product A', 'Product B', 'Product C'];

var yValue = [20, 14, 23];

var trace1 = {
x: xValue,
y: yValue,
type: 'bar',
text: yValue,
textposition: 'auto',
hoverinfo: 'none',
marker: {
  color: 'rgb(158,202,225)',
  opacity: 0.6,
  line: {
    color: 'rbg(8,48,107)',
    width: 1.5
  }
}
};

var data = [trace1];

var layout = {
title: 'January 2013 Sales Report'
};

Plotly.newPlot('myDiv', data, layout);

function no_text() {
  Plotly.restyle('myDiv', {textposition: ['none']}, [0]);
}

  </script>

  <a href="javascript:no_text()">No Text</a>

</body>
```

alexcjohnson:
Thanks @kochelmonster - there's another bug if you delete the text array:
```js
Plotly.restyle(gd, {text: null})
```

The text doesn't disappear, and gets stuck in place if you pan/zoom on the plot:
![screen shot 2018-06-01 at 5 02 05 pm](https://user-images.githubusercontent.com/2678795/40863384-9a254e12-65bd-11e8-9c51-dae85581e973.png)

alexcjohnson:
@etpinard seems like these came in with some of the 🐎 optimizations between 1.36 and 1.37 https://codepen.io/alexcjohnson/pen/jKbmmV

etpinard:
> optimizations between 1.36 and 1.37

Ok. On it!

etpinard:
It's from https://github.com/plotly/plotly.js/pull/2574, where we :hocho: that ugly `selectAll('.trace')` on each `Cartesian.plot` call.

PR #2689, opened 2018-06-01
Fixup bartext removal in update selections

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2688 which was a regression since commit https://github.com/plotly/plotly.js/pull/2574/commits/6d88255bc76ce8b0a7b6c1a90ec00fb5886d6227 where all `Bar` started using a d3-iomatic update pattern.

cc @alexcjohnson


alexcjohnson:
That was quick! Looks great :dancer:

Issue #2665, opened 2018-05-24
Plotly.react() in non-WebGL contexts


nicolaskruchten:
If you run this pen https://codepen.io/nicolaskruchten/pen/pVXgmB with WebGL disabled (e.g. in Firefox with webgl.disabled=true in about:config) you'll see that `react`ing between the two states doesn't clear the error message. 

This currently affects the `react-chart-editor` on platforms like iPads.

Issue #2666, opened 2018-05-24
Missing error messages for some WebGL trace types


nicolaskruchten:
Trying to use `scattergl`, `scatterpolargl` (and possible others) in non-WebGL contexts (e.g. Firefox with webgl.disabled=true in about:config or on an iPad) does not result in an error message but instead partially renders: just the SVG bits like axes.

PR #2697, opened 2018-06-05
Make all gl traces show "no webgl message" and fail gracefully

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2665 (the fail _gracefully_ part) and https://github.com/plotly/plotly.js/issues/2666 (the missing message part).

Most importantly, I found a pretty solid way to test no-webgl environments. Adding a `--disable-webgl`  flag to the Chrome CLI in our Karma config did the trick.

@alexcjohnson @nicolaskruchten

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -199,6 +199,8 @@ function sceneUpdate(gd, subplot) {
             Lib.extendFlat(scene, initOpts, resetOpts);
         };
 
+        scene.init();
+
         // apply new option to all regl components (used on drag)
         scene.update = function update(opt) {
             var opts = new Array(scene.count);
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -166,7 +166,7 @@ function sceneUpdate(gd, subplot) {
     var scene = subplot._scene;
     var fullLayout = gd._fullLayout;
 
-    var reset = {
+    var resetOpts = {
         // number of traces in subplot, since scene:subplot → 1:1
         count: 0,
         // whether scene requires init hook in plot call (dirty plot call)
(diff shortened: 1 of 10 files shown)


src/traces/scattergl/index.js
```diff
@@ -326,7 +326,15 @@ function plot(gd, subplot, cdata) {
     var width = fullLayout.width;
     var height = fullLayout.height;
 
-    prepareRegl(gd, ['ANGLE_instanced_arrays', 'OES_element_index_uint']);
+    var success = prepareRegl(gd, ['ANGLE_instanced_arrays', 'OES_element_index_uint']);
+    if(!success) {
+        scene.error2d = false;
+        scene.line2d = false;
+        scene.scatter2d = false;
+        scene.fill2d = false;
```
alexcjohnson: what happens if we add some other component to support a new feature, would we also have to add it here? Will the tests below fail and force that, or would we need to add a new test for the new feature? Just wondering if there's some way we can preempt these concerns like
```js
if(!success) {
  for(var key in scene) {
    if(something about key or scene.key) scene[key] = false;
  }
}
```

nicolaskruchten:
Nice! with reference to the gif above, I would say that the error message should have an opaque background (white?) and a border (grey?) so it stands out a bit on top of the axes. Other than that 👍

test/jasmine/bundle_tests/no_webgl_test.js
```diff
@@ -0,0 +1,105 @@
...
+        Plotly.react(gd, require('@mocks/gl2d_12.json'))
+        .then(function() {
+            checkNoWebGLMsg(true);
+            return Plotly.react(gd, require('@mocks/10.json'));
+        })
+        .then(function() {
+            checkNoWebGLMsg(false);
+        })
+        .catch(failTest)
+        .then(done);
```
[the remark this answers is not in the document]
etpinard: new stylez

peek 2018-06-06 15-10

src/traces/scattergl/index.js
etpinard: all DRY'ed up in eb594d1bf96d0640b5fe7d5b75aac67f3e2418a0

alexcjohnson:
Nicely done! 💃

PR #2698, opened 2018-06-05
Fix scatter3d scalar hovertext handling

etpinard:
Easy fix cc @alexcjohnson 

But: wow, I wish hover from subplot type to another other would be done in a DRYer way :tired_face:


alexcjohnson:
yep, at some point we should make an attempt to 🌴 it up - will probably uncover and fix a bunch more bugs that way like I did (am doing...) with colorbars in #2681 

But nice quick fix 💃

etpinard:
> we should make an attempt to  it up

To me, the first step to :palm_tree: ing this up would be to make all hover-related routines use just `_fullData` (i.e. no `calcdata`) so that e.g. scattergl and scatter can share the same `hoverPoints` method.

PR #2705, opened 2018-06-07
fix line decimation for segments crossing the viewport

### LINKED ISSUE #2664: Line that should render on x axis rendered above it
When plotting a line chart with several points at (x,0) the interpolated line between them renders above the X axis line for certain data sets.

You can view an example [here](https://jsfiddle.net/bh0etd08/). Zoom in such that your x window is roughly 2 to 1 and your y window is -0.04 to .06. As you continue to zoom you will see the incorrect line, even though the only points in that portion of the chart are: (-1,0) and (3.214112,0).

newplot

Thanks,
Jon

alexcjohnson:
Fixes #2664 - the problem was the increasing decimation tolerance as points get farther from the viewport - if a *segment* crosses the viewport we need to use the on-screen (minimum) tolerance even though points on both sides are far enough away to get a high tolerance. Adds a little extra calculation into the decimation algorithm unfortunately, but I don't think it's too big a hit (nor do I see any higher-performance alternative).

cc @etpinard

### DIFF
--- a/test/jasmine/tests/scatter_test.js
+++ b/test/jasmine/tests/scatter_test.js
@@ -489,6 +489,9 @@ describe('Test scatter', function() {
                 [-822, 20], [-802, 2], [-801.5, 1.1], [-800, 0],
                 [900, 0], [901.5, 1.1], [902, 2], [922, 20]
             ], [
+                // all that's really important here (and the next check) is that
+                // the points [-800, 0] and [900, 0] are connected. What we do
+                // with other points beyond those doesn't matter too much.
                 [-822, 20], [-800, 0],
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/scatter_test.js
```diff
@@ -472,6 +474,47 @@ describe('Test scatter', function() {
                 expect(ptsOut2[0]).toBeCloseTo2DArray(ptsExpected[i].map(reverseXY), 1, i);
             });
         });
+
+        it('works when far off-screen points cross the viewport', function() {
```
etpinard: Nicely done. :dancer:

PR #2710, opened 2018-06-08
Plots.resize - additional check for gd.layout

rafalbromirski:
While working with `plotly` & `react-plotly`, from time to time I could notice minor error in my console:

```
Uncaught TypeError: Cannot read property 'width' of undefined
```

First, I thought the error is somewhere on my side but later it was clear it was inside `plotly` library. I backtracked the error and ended up here:

```js
if(gd.layout.width && gd.layout.height)
```

=> https://github.com/plotly/plotly.js/blob/master/src/plots/plots.js#L75-L113

Reproducing it was possible by changing my routes faster than usual – not waiting for plotly to finish rendering charts – chaos monkey style. Apparently when you delete DOM node from your HTML, `gd` variables still holds the reference to the element which doesn't exist anymore. In that case if you try to call `gd.layout.width/height` it will throw an error because `layout` is `undefined`.

### How to reproduce it:
- `npm start`
- open JS console and type:
```js
gd.layout
=> {updatemenus: Array(18), xaxis: {…}, yaxis: {…}, height: 450, width: 1100, …}
```
- remove `gd` element:
```js
gd.parentElement.removeChild(gd)
```
- try to call it again:
```js
gd.layout
=> undefined
```

If you call `gd`, it still points to `#graph`: 
```js
gd
=> <div class="dashboard-plot" id="graph"></div>
```

### DIFF
--- a/test/jasmine/tests/plot_promise_test.js
+++ b/test/jasmine/tests/plot_promise_test.js
@@ -2,7 +2,7 @@ var Plotly = require('@lib/index');
 var Events = require('@src/lib/events');
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
-
+var failTest = require('../assets/fail_test');
 
 describe('Plotly.___ methods', function() {
     'use strict';
@@ -468,7 +468,9 @@ describe('Plotly.___ methods', function() {
                 expect(gd).toBeDefined();
                 expect(typeof gd).toBe('object');
                 expect(gd.layout).toBeDefined();
-            }).then(done);
+            })
+            .catch(failTest)
+            .then(done);
         });
 
         it('should return a rejected promise if gd is hidden', function(done) {
@@ -478,7 +480,9 @@ describe('Plotly.___ methods', function() {
             }, function(err) {
                 expect(err).toBeDefined();
                 expect(err.message).toBe('Resize must be passed a displayed plot div element.');
-            }).then(done);
+            })
+            .catch(failTest)
+            .then(done);
         });
 
         it('should return a rejected promise if gd is detached from the DOM', function(done) {
@@ -488,7 +492,30 @@ describe('Plotly.___ methods', function() {
             }, function(err) {
                 expect(err).toBeDefined();
                 expect(err.message).toBe('Resize must be passed a displayed plot div element.');
-            }).then(done);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('should return a resolved promise if plot has been purged and there is nothing to resize', function(done) {
+            var resizePromise = Plotly.Plots.resize(initialDiv);
+
+            Plotly.purge(initialDiv);
+            destroyGraphDiv();
+
+            resizePromise
+                .catch(failTest)
+                .then(done);
+        });
+
+        it('should return a resolved promise if plot has been hidden and gd is hidden', function(done) {
+            var resizePromise = Plotly.Plots.resize(initialDiv);
+
+            initialDiv.style.display = 'none';
+
+            resizePromise
+                .catch(failTest)
+                .then(done);
         });
 
         it('errors before even generating a promise if gd is not defined', function() {
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -90,7 +90,7 @@ plots.resize = function(gd) {
 
         gd._redrawTimer = setTimeout(function() {
             // return if there is nothing to resize
-            if(gd.layout.width && gd.layout.height) {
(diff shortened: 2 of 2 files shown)


alexcjohnson:
Interesting, thanks for bringing this up @rafalbromirski! Seems like a reasonable check but the source must be a little different from what you've identified. The reason `gd` has no `layout` when you detach it in the test dashboard is that we [recreate it if needed every second](https://github.com/plotly/plotly.js/blob/49819237db1673dc98b087e6e1e7a9a9b607784b/devtools/test_dashboard/devtools.js#L157) so you're not looking at the same `gd`. If before calling `gd.parentElement.removeChild(gd)` you say `gd2 = gd`, you can see that at the end `gd2` is different from `gd`, and `gd2` still has a layout.

However if you call `Plotly.purge(gd)` (possibly an [unmount](https://github.com/plotly/react-plotly.js/blob/6fe45a76d1542419fdbb9379941a631706bcf0a8/src/factory.js#L142) before this timer runs out?) we DO remove `layout`, and everything else we had attached to the `<div>`. So in that case the check you added makes a lot of sense, `gd` was valid at the time `resize` was called so we should not generate an error.

Can you add a test to the [resize promise test suite](https://github.com/plotly/plotly.js/blob/49819237db1673dc98b087e6e1e7a9a9b607784b/test/jasmine/tests/plot_promise_test.js#L453-L498) mimicking this situation? I'm thinking:
```js
var resizePromise = Plotly.Plots.resize(initialDiv);

// now purge and detach the element before the resize timer runs out
Plotly.purge(initialDiv);
initialDiv.parentElement.removeChild(initialDiv);

// then wait for the promise as usual, if it doesn't error the test passes
resizePromise.catch(failTest).then(done);
```
hmm, looks like that test file doesn't have our robust promise error handling (`failTest`) yet - needs it added like in [this commit](https://github.com/plotly/plotly.js/commit/26278de2dcc7771d8c76c45df8fa4e1623a06ea2#diff-739250b254e51d88177d4372123fb2d5), otherwise a failure is just manifest as a timeout because `done` never gets called.

rafalbromirski:
@alexcjohnson thanks for explaining! Will look into tests in my spare time.

rafalbromirski:
@alexcjohnson I tried to follow your lead but I got an error:

```
Uncaught TypeError: Cannot convert undefined or null to object thrown
```
and it seems like this error comes from `_redrawTimer` method (https://github.com/plotly/plotly.js/blob/master/src/plots/plots.js#L91-L111)

Maybe because I'm calling `purge` before and almost all plotly attributes are gone?

I tried different solution and it seems like it's working fine. Could you take a look? Not sure if it's the right one though...

test/jasmine/tests/plot_promise_test.js
```diff
@@ -488,7 +492,34 @@ describe('Plotly.___ methods', function() {
             }, function(err) {
                 expect(err).toBeDefined();
                 expect(err.message).toBe('Resize must be passed a displayed plot div element.');
-            }).then(done);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        xit('should return a resolved promise and purge plotly attributes - tmp', function(done) {
```
rafalbromirski: @alexcjohnson you would have to enable it to see the error

alexcjohnson: Playing with this a little... I think this version of the test *is* correct, and what's needed is just to interpret the comment in the code `// return if there is nothing to resize` a little more broadly - if the plot has been purged (in which case it has no `layout`) that's another "nothing to resize." So perhaps:
```js
if(!gd.layout || (gd.layout.width && gd.layout.height)) {
```
Which, at least if I run it in the console (`Plotly.Plots.resize(gd);Plotly.purge(gd);`) successfully avoids throwing an error.

Hmm, it also occurs to me, looking at the code, we should bail out if `isHidden(gd)` as well, ie when `resize` was called the plot was visible, but before `_redrawTimer` fired it has been hidden/removed. That's yet another "nothing to resize." So I guess:
```js
if(!gd.layout || (gd.layout.width && gd.layout.height) || isHidden(gd)) {
```
The only other test we do outside `_redrawTimer` is `!gd`, but I don't see any way `gd` could be truthy before and falsy 100ms later, it's still going to be a DOM element, the worst that can happen is it gets detached; so I think that should do it!

rafalbromirski:
@alexcjohnson done. I added additional code which handles early return (it still returns `resolved` promise like it did before) but if somebody will change it in the future you will at least have tests for it.

alexcjohnson:
Beautiful, thanks @rafalbromirski! 💃

etpinard:
Lovely tests @rafalbromirski 

Thanks very much!

PR #2712, opened 2018-06-08
A few surface contours fixes

etpinard:
Fixes two surface contour line bugs uncovered in @alexcjohnson 's https://codepen.io/alexcjohnson/pen/gKLBgR?editors=0010:

- bug 1: contour directions (x,y or z) that have a 0-length first level (these are computed in gl-surface3d) did not pass the correct contour color to the shaders. See https://github.com/gl-vis/gl-surface3d/pull/11
- bug 2: `contour.(x|y|z).highlight` wasn't working properly for the same mock (possibly due to that same 0-length first level condition)

Note that both these bugs aren't noticeably on our https://rreusser.github.io/plotly-mock-viewer/#gl3d_contour-lines


alexcjohnson:
Awesome, thanks for taking a look at this!

There's a 3rd bug in that codepen (and your new mock here), that contour lines I expected to be straight are not rendered straight. It looks to me as though it's actually the surface that's not getting drawn in quite the right place - perhaps it's trying to smooth the sharp edges a bit or something - and then the contours are at the correct coordinates but faithfully following the incorrect surface. See what happens to hover when I zoom in on the corner of the bottom:
screen shot 2018-06-08 at 2 47 27 pm

Anyway, this is still 2 bugs down so lets get it in! 💃
Do you want me to make a new issue for the 3rd one?

etpinard:
> Do you want me to make a new issue for the 3rd one?

That would be great!

PR #2714, opened 2018-06-08
Support transforms that make their own data without trace input data

alexcjohnson:
Some transforms in Plotly cloud pull their data from other traces, the trace being "transformed" has no data on its own. This allows those transforms to continue to operate despite the `_length` requirement from #2577. Also allows transforms that make their own data from scratch, such as the `linemaker` transform in the test I added.

Calling this a bugfix since it fixes our broken application code :)

cc @etpinard @VeraZab


etpinard:
Now part of `v.1.38.3`. We can merge this master :dancer:

alexcjohnson:
Already included in #2725

Issue #2700, opened 2018-06-06
This cone trace with `sizemode: 'absolute'` punches through scene wall


etpinard:
See https://codepen.io/etpinard/pen/mKEoro?editors=0010

this should not happen. Something must be wrong with how we add autorange pads here for `sizemode: 'absolute'`:

https://github.com/plotly/plotly.js/blob/ece927fd075fd55f1b7adae38a37f2b24551988a/src/traces/cone/convert.js#L111-L113

Issue #2701, opened 2018-06-06
Cone traces with irregular grids do not scale cone correctly


etpinard:
See https://codepen.io/etpinard/pen/gKMJyZ that has the default `sizemode` and `sizeref` and compare with https://codepen.io/etpinard/pen/eKzawB?editors=0010 that has `sizeref: 10`. Cone should always be visible using the default `sizeref` so something is off. 

I believe this is (another) side-effect of our `dataScale` hack (more on that in: https://github.com/plotly/plotly.js/issues/2646) where `sizeref: 1` should correspond to two cones of max norm at the minimal adjacent position  _just_ touching each other. But by the time `gl-cone3d` computes the max length the x/y/z positions and u/v/w vectors are scaled meaning that we should also scale `sizeref` before passing it to `gl-cone3d`. Doing this might be tricky as x/y/z and u/v/w are vectors and sizeref is a scalar, but hopefully there's a way to do so in plotly.js.

If so, fixing this should also fix https://github.com/plotly/plotly.js/issues/2700

etpinard:
by _irregular_ I mean grids that generate unequal x-axis, y-axis, z-axis ranges.

etpinard:
Closed in #2715 (for now at least)

Issue #2646, opened 2018-05-18
Replace "dataScale" hack in 3D subplots


etpinard:
From https://github.com/plotly/plotly.js/pull/2641#discussion_r189329872

We should spend some time replacing this `dataScale` hack. This thing was added to make date axes work properly in 3D. Timestamps don't fit in a Float32Array, so we can't just convert the datestrings to timestamps, but we do this better in `scattergl`, where we use Float64Array or split the data into two Float32Arrays when not supported.

Fixing this should improve autorange results and make incorporating upcoming gl-vis components much easier!

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2715, opened 2018-06-08
Cone sizeref fixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2700 and https://github.com/plotly/plotly.js/issues/2701 (with the help https://github.com/gl-vis/gl-cone3d/pull/12).

This PR also improves hover on cones by picking the entire cone as opposed to overlaying `gl-scatter3d` points and using those for picking. Note that, the hover label remain at the same positions as before.

I still need to fixup a few tests, but I would appreciate a review of commit https://github.com/plotly/plotly.js/commit/4836861130a1a2944d42ca2fc2025efa2027d68d (where I tried to make sense of `sizemode: 'absolute'`) and commit https://github.com/gl-vis/gl-cone3d/pull/12/commits/3b3ef9d465a8e30d14a9a4a1cc88d4deebec64a9 (where I impleted a "local" find-vectorScale algo as proposed by @alexcjohnson  in https://github.com/gl-vis/gl-cone3d/pull/12/commits/3b3ef9d465a8e30d14a9a4a1cc88d4deebec64a9)


src/traces/cone/convert.js
```diff
@@ -88,15 +87,21 @@ function convert(scene, trace) {
...
-    coneOpts[sizeMode2sizeKey[trace.sizemode]] = trace.sizeref;
     coneOpts.coneOffset = anchor2coneOffset[trace.anchor];
 
-    var meshData = conePlot(coneOpts);
+    if(trace.sizemode === 'scaled') {
+        // unitless sizeref
+        coneOpts.coneSize = trace.sizeref || 0.5;
+    } else {
+        // sizeref here has unit of velocity
+        coneOpts.coneSize = (trace.sizeref / trace._normMax) || 0.5;
```
etpinard: Previously, `coneSize` under `sizemode: 'absolute'` was scaled using the max norm of the "scaled" u/v/w vectors (that same `dataScale` than in https://github.com/plotly/plotly.js/issues/2646). This made `sizeref` under `sizemode: 'absolute'` had little meaning. 

Now, `sizeref` under `sizemode: 'absolute' is scaled by the "input" u/v/w max norm should make this thing a little less confusing.

src/traces/cone/attributes.js
```diff
@@ -112,20 +112,26 @@ var attrs = {
...
         dflt: 'scaled',
         description: [
-            'Sets the mode by which the cones are sized.',
-            'If *scaled*, `sizeref` scales such that the reference cone size',
-            'for the maximum vector magnitude is 1.',
-            'If *absolute*, `sizeref` scales such that the reference cone size',
-            'for vector magnitude 1 is one grid unit.'
+            'Determines whether `sizeref` is set as a *scaled* (i.e unitless) scalar',
+            '(normalized by the max u/v/w norm in the vector field) or as',
+            '*absolute* value (in unit of velocity).'
```
alexcjohnson: I wouldn't use the word "velocity" here - you can show many other quantities as vector fields (electric fields, as just one common example). Just say "in the same units as the vector field" or some such.

src/traces/cone/attributes.js
```diff
@@ -112,20 +112,26 @@ var attrs = {
...
         role: 'info',
         editType: 'calc',
         min: 0,
-        dflt: 1,
-        description: 'Sets the cone size reference value.'
+        description: [
+            'Adjusts the cone size scaling.',
+            'The size of the cones is determined by their u/v/w norm multiplied a factor and `sizeref`.',
+            'This factor (computed internally) corresponds to the minimum "time" to travel across two',
+            'two successive x/y/z positions at the average velocity of those two successive positions',
```
alexcjohnson: two twos is one two too many

src/traces/cone/convert.js
etpinard: > What if the field is uniformly zero? 

Oh right, `Infinity` is truthy. Good call!

etpinard: > When is the || 0.5 (in either case) necessary?

Currently, the `sizeref` attribute declaration does not set a `dflt` field. In practice I guess it does have a hard default of `0.5`. But technically under the `sizemode: 'absolute'`, the sizemode default is `0.5 * trace._normMax`.

alexcjohnson:
Great! Smoothing off a few more rough edges with cones. 💃

PR #2724, opened 2018-06-11
fix #2722 - get the right traces in hist calc for cross-trace coupling

### LINKED ISSUE #2722: Error hiding and redisplaying multiple histograms
```
Uncaught (in promise) TypeError: Cannot read property 'length' of undefined
    at Object.calc (plotly-latest.js:168006)
    at calci (plotly-latest.js:148106)
    at Object.plots.doCalcdata (plotly-latest.js:148130)
    at exports.plot (plotly-latest.js:123908)
    at Object.lib.syncOrAsync (plotly-latest.js:119125)
    at Object.restyle (plotly-latest.js:125048)
    at EventEmitter.<anonymous> (pen.js:69)
    at EventEmitter.emit (plotly-latest.js:29592)
    at HTMLDivElement.plotObj.emit (plotly-latest.js:117770)
    at plotly-latest.js:125052
```

https://codepen.io/anon/pen/oyZMOG

To trigger this error, on the line chart, first, disable trace A by clicking on it in the legend. Then, on the line chart, disable trace B by clicking on it in the legend. Finally, on the line chart, re-enable trace B by clicking on it in the legend. The shown stack trace is produced in the JavaScript console and trace B is not updated on the histogram. I have found more than one way to trigger this error but this is the most easily reproducible.

Also, after triggering this error, moving the mouse over the histogram throws continuous errors.

Is there a supported way for syncing traces on different charts? This was the best that I could come up with.

@rahul0705

alexcjohnson:
@etpinard a quick one to fix #2722

### DIFF
--- a/src/traces/histogram/calc.js
+++ b/src/traces/histogram/calc.js
@@ -23,7 +23,7 @@ var oneMonth = require('../../constants/numerical').ONEAVGMONTH;
 var getBinSpanLabelRound = require('./bin_label_vals');
 
 module.exports = function calc(gd, trace) {
-    // ignore as much processing as possible (and including in autorange) if bar is not visible
+    // ignore as much processing as possible (and including in autorange) if not visible
     if(trace.visible !== true) return;
 
     // depending on orientation, set position and size axes and data ranges
@@ -435,6 +435,7 @@ function getConnectedHistograms(gd, trace) {
     for(var i = 0; i < fullData.length; i++) {
         var tracei = fullData[i];
         if(tracei.type === 'histogram' &&
+            tracei.visible === true &&
             tracei.orientation === orientation &&
             tracei.xaxis === xid && tracei.yaxis === yid
         ) {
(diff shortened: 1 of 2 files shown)


src/traces/histogram/calc.js
```diff
@@ -435,6 +435,7 @@ function getConnectedHistograms(gd, trace) {
     for(var i = 0; i < fullData.length; i++) {
         var tracei = fullData[i];
         if(tracei.type === 'histogram' &&
+            tracei.visible === true &&
```
alexcjohnson: To match the condition [above](https://github.com/plotly/plotly.js/pull/2724/files#diff-e02f19bac6aa5dbda7a48a07b3ad45fdR27) filtering out invisible traces. Interestingly the initial draw (if you just set `visible: 'legendonly'` on the first trace as of `newPlot`) works OK, though I suspect that in certain cases this will lead us to choose the wrong autobins... this fixes that potential problem as well.

etpinard:
Nicely done :dancer:

Issue #2345, opened 2018-02-07
Uncaught in promise: Mapbox error


jackparmer:
This map throws a Mapbox+Plotly.js console error on load and does not fire any selection events afterwards:
`Uncaught (in promise) Error: Mapbox error.`

https://plot.ly/~jackp/18343/

This map is the exact same Plotly graph, but has different geojson overlayers. It does not fire any console errors and responds to selection events.

https://plot.ly/~jackp/18345/

I'm guessing mapbox-gl is having an issue with the geojson in Graph1, but not in Graph2. The geojson for problematic Graph1 is here for convenience:

https://github.com/jackparmer/mapbox-counties/tree/master/1999

It would be nice if the mapbox error was caught and Plotly.js events still fired for Graph1.

etpinard:
I think this happens when the geojson source data is (somewhat) invalid.

We should try running http://geojsonlint.com/ on those geojsons

jackparmer:
> We should try running http://geojsonlint.com/ on those geojsons

I did some linting with this project and didn't find any issues: https://github.com/mapbox/geojsonhint

http://geojsonlint.com/ has some right-hand rule guidance but it's the same for the working and non-working geojson files.

etpinard:
Looks like plain `mapbox-gl` (even their latest version `v.0.44.0`) outputs the same error. From https://codepen.io/etpinard/pen/aqBLgv?editors=0010, I get

![image](https://user-images.githubusercontent.com/6675409/35937723-20b7adce-0c15-11e8-80b1-12d1c13adba4.png)

Maybe we should open a bug report there?

--------

> It would be nice if the mapbox error was caught and Plotly.js events still fired for Graph1.

I can catch that error no problem. But I don't get why events wouldn't get triggered regardless. How are you attaching your event handlers?

leejt489:
Hi, I'm also getting a similar error. I have a map that loads fine, but then when I update it with a slider this error fires in browser with the callback to update. Any update? I think my GeoJSON is fine.

etpinard:
@leejt489 can you please share the problematic geojson in question? Better yet, can you a reproducible plotly.js or [mapbox-gl](https://codepen.io/etpinard/pen/NzxeJJ) example? That way we can determine if it is a problem with plotly.js or mapbox-gl. Thanks.

leejt489:
[...]

PR #2734, opened 2018-06-13
Fix mapbox layout layer updates

etpinard:
- when a new source is given, remove layer, update source and add new layer in that order to avoid mapbox-gl errors.

Some more info about how mapbox layout layers are updated:

These are the available update pathways:

https://github.com/plotly/plotly.js/blob/27c77c11f2eab9cf773543926938743d2eb2b0fc/src/plots/mapbox/layers.js#L35-L51

where `needsNewSource` is fairly dumb (but fast):

https://github.com/plotly/plotly.js/blob/27c77c11f2eab9cf773543926938743d2eb2b0fc/src/plots/mapbox/layers.js#L53-L62

which explains why 

```js
// new source object on each react
function makeFigure() {
  return {
    // ...
   layout: {
      mapbox: {
        layers: [{source: {/**/}}]
      }
   }
  }
}

Plotly.react(gd, makeFigure())
Plotly.react(gd, makeFigure())
```

can behave differently than

```js
// same source object
var source = {/* */}

function makeFigure() {
  return {
    // ...
   layout: {
      mapbox: {
        layers: [{source: source}]
      }
   }
  }
}

Plotly.react(gd, makeFigure())
Plotly.react(gd, makeFigure())
```

So perhaps we could add `uid` attribute in each mapbox layer so that users that create a new source object on each react call can see their map update faster. Thoughts?

cc @alexcjohnson

### DIFF
--- a/src/plots/mapbox/layers.js
+++ b/src/plots/mapbox/layers.js
@@ -39,8 +39,9 @@ proto.update = function update(opts) {
         this.updateLayer(opts);
     } else if(this.needsNewSource(opts)) {
         // IMPORTANT: must delete layer before source to not cause errors
-        this.updateLayer(opts);
+        this.removeLayer();
         this.updateSource(opts);
+        this.updateLayer(opts);
     } else if(this.needsNewLayer(opts)) {
         this.updateLayer(opts);
     } else {
@@ -87,8 +88,7 @@ proto.updateLayer = function(opts) {
     var map = this.map;
     var convertedOpts = convertOpts(opts);
 
-    if(map.getLayer(this.idLayer)) map.removeLayer(this.idLayer);
-
+    this.removeLayer();
     this.layerType = opts.type;
 
     if(isVisible(opts)) {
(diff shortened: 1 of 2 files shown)


etpinard:
This PR may help with https://github.com/plotly/plotly.js/issues/2345#issuecomment-363661069 but not the original #2345 problem.

alexcjohnson:
> So perhaps we could add `uid` attribute in each mapbox layer so that users that create a new source object on each react call can see their map update faster. Thoughts?

Perhaps we should compare `source` as (JSON) strings if they fail strict equality (and use `_compareAsJSON` so `Plotly.react` does this too)?

https://github.com/plotly/plotly.js/blob/27c77c11f2eab9cf773543926938743d2eb2b0fc/src/plot_api/plot_api.js#L2495

Could be a bit slow if `source` is large, but I feel like `uid` is asking users to get a bit too involved in our low-level optimizations.

etpinard:
> Could be a bit slow if source is large,

That was my main concern.

alexcjohnson:
Anyway my comment ^^ is just a 🐎 question, could be addressed now or not, up to you. The bugfix here looks great, so if you want to leave it at that 💃 !

etpinard:
> Anyway my comment ^^ is just a  question, could be addressed now or not, up to you.

Great. I'll defer it if you don't mind. I'd like to investigate `_compareAsJSON` on some real-life geojsons before implementing it.

PR #2739, opened 2018-06-15
Polar polygon grids

etpinard:
This PR adds an attribute to polar 2.0 charts that allows users to draw polar radial grid lines and the angular axis line as polygons when the angular axis has `type: 'category'`. in brief, this allows users to create _poorman's_ radar charts (i.e. radar charts with the same radial scale for all variables).

This new attribute is tentatively named `polar.usepolygons`. I'm not particularly happy about the name. I think we should make this a `polar.` attribute as it affects the look of both radial axis grid lines and the angular axis line. I'm open to suggestions.

Adding this feature turned out to a little more difficult than I thought. Quite a bit of "polygon-only" needed to be added in https://github.com/plotly/plotly.js/commit/c3f5b1baaa2f8ad8a909b44ac9603ab3f09bd7f8. Getting this to work for sectors for tricky (though there's probably a simpler algo out there). We need to _snap_ `radialaxis.angle` to one of the vertex angles to make sure radial axis scale remain the same as the data scale. Angular drag needed some special treatment. Here's what it looks like:

peek 2018-06-15 18-18

TODO:
- [x] agree on attribute name
- [x] maybe make zoombox handles follow the polygon, this would make zoombox interaction lose their rotational symmetry though - not sure what's best
- [x] maybe restrict `theta` values to string or "integer" to not show "out-of-scale" values
- [x] maybe make the angular drag layer a "polygon" annulus
- [x] add a few jasmine tests

cc @alexcjohnson @chriddyp


chriddyp:
> `polar.usepolygons`

Would something like `layout.radialaxis.gridlinesshape = 'linear' | 'circular'` work? Matching the the `scatter.line.shape` properties

etpinard:
> Would something like layout.radialaxis.gridlinesshape = 'linear' | 'circular' work?

It would _work_, but

> it affects the look of both radial axis grid lines and the angular axis line

chriddyp:
throwing some ideas out there:
- `radar: true`
- `spider: true`
- `shape: 'linear' | 'circular'`

alexcjohnson:
`radar` is ambiguous - actual radar screens have circles on them, right? And `spider: false` doesn’t really scream “circular” to me. So I’d prefer both shapes to be explicitly stated, like `gridshape: (linear|circular)`. 

> maybe make zoombox handles follow the polygon, this would make zoombox interaction lose their rotational symmetry though

Yes. Rotational symmetry is already gone. 

> maybe restrict theta values to string or "integer" to not show "out-of-scale" values

This isn’t already the case for category axes?

> maybe make the angular drag layer a "polygon" annulus

Would be weird if there are little gaps between the angular and radial drag areas, so probably yes.

etpinard:
> Yes. Rotational symmetry is already gone.

Ok. So should the cursor follow the handle, or should it just snap to the closest polygon edge?

alexcjohnson:
> So should the cursor follow the handle, or should it just snap to the closest polygon edge?

Mmm, it could actually be a bit weird if the handles move around everywhere the cursor goes, especially as they slip over the corners and would have to bend in arbitrary places. And there would be something nice about leaving the handles in the center of the polygon edge, where we know there's no data, only connecting lines. So yeah, I guess snap to the nearest edge center, sound reasonable?

That said, currently (circular mode) the starting handle is fixed and only the ending handle moves around with the cursor, I wonder if it wouldn't look cleaner to have the ending handle stay at the same angle as the starting handle no matter what? That's how we do it in cartesian with one axis `fixedrange` and I like it, makes clear that motion in the other direction is irrelevant. So I might argue for snapping the starting handle to the nearest edge center, then keeping the ending handle on the same edge no matter where the cursor goes.

Regardless (and this may already be implied but just to be clear) I think it's important that the radial value implied by any given cursor location is given by the *polygon* it's on, NOT its distance from the center point.

etpinard:
@alexcjohnson @chriddyp 

Going with `gridshape: 'circular' | 'linear'` in https://github.com/plotly/plotly.js/pull/2739/commits/a26b56f5e999e572acac449c767cc69bca1053ef

etpinard:
> I wonder if it wouldn't look cleaner to have the ending handle stay at the same angle as the starting handle no matter what?

peek 2018-06-19 16-08

I like it. @alexcjohnson What do you think?

etpinard:
> And there would be something nice about leaving the handles in the center of the polygon edge, where we know there's no data, only connecting lines. So yeah, I guess snap to the nearest edge center, sound reasonable?

peek 2018-06-20 09-15

Is this what you had in mind @alexcjohnson ?

etpinard:
> maybe restrict theta values to string or "integer" to not show "out-of-scale" values

About :arrow_heading_up: , this doesn't apply to polar subplots at the moment. Only cartesian data-ref'ed layout items support real numbers between categories. For example,

```js
Plotly.newPlot(gd, [{
  x: ['a', 'b', 0.5],
  y: [1, 2, 3]
}, {
  type: 'scatterpolar',
   r: [1, 2, 3],
   theta: ['a', 'b', 0.5]
}], {
  xaxis: {
    type: 'category',
	domain: [0, 0.5]
  },
  polar: {
    domain: {x: [0.5, 1]},
    angularaxis: {type: 'category'}
  },
  annotations: [{
    xref: 'x', x: 1.5,
    yref: 'y', y: 2,
    showarrow: false,
    text: 'text at (1.5, 2)'
  }]
})
```

gives:

image

But we should keep this in mind when we'll add polar data-ref'ed annotations.

alexcjohnson:
> Is this what you had in mind?

I was imagining we'd put straight-line handles in the center of whatever polygon edge you're over - or, per our discussion yesterday on slack, whichever edge the mouse was over when you started the zoom, like this:
screen shot 2018-06-20 at 9 21 07 am

Also the cursor should be *on* the polygon edge, not at the same radius as the polygon vertices. That's what I meant above:
> the radial value implied by any given cursor location is given by the *polygon* it's on, NOT its distance from the center point.

---
> About ⤴️ , this doesn't apply to polar subplots at the moment. Only cartesian data-ref'ed layout items support real numbers between categories.

Got it, thanks for clarifying.

etpinard:
> I was imagining we'd put straight-line handles in the center of whatever polygon edge you're over

and what if the cursor is on a vertex?

alexcjohnson:
> and what if the cursor is on a vertex?

break the symmetry however you like - but I don't think it's worth making bent handles at vertices, in fact I'd argue it's better to use the edge centers anyway because we know we're not obscuring any data points there.

etpinard:
Wow, I totally misread that comment of yours. @alexcjohnson here's my latest attempt :arrow_down: 

peek 2018-06-20 13-31

etpinard:
@alexcjohnson latest attempt :arrow_heading_down: 

peek 2018-06-20 18-31

alexcjohnson:
Love the look and feel now, nicely done! I still see some discrepancy between mouse position and polygon edge. In this clip, I'd expect the mouse to *always* be exactly on an edge of the polygon, but sometimes it's inside and sometimes outside:
polar poly zoom
Also, I can't angular-drag the upper-right subplot there (the green one) - should I be able to? I don't see anything in `layout.polar3` that I would expect to prevent it...

etpinard:
> Also, I can't angular-drag the upper-right subplot there (the green one) - should I be able to? I don't see anything in layout.polar3 that I would expect to prevent it...

See:

https://github.com/plotly/plotly.js/blob/40040a527935b91aafd83649059e48012f0486c9/src/plots/polar/polar.js#L1145-L1149

etpinard:
I didn't expect adding almost 1000 lines for this thing, but oh well here it is :white_check_mark:

test/jasmine/tests/polar_test.js
```diff
@@ -1034,7 +1073,226 @@ describe('Test polar interactions:', function() {
...
+        function _assert(msg, angle) {
+            expect(gd._fullLayout.polar.radialaxis.angle)
+                .toBeCloseTo(angle, 1, msg + ' - angle');
+        }
+
+        Plotly.plot(gd, [{
+            type: 'scatterpolar',
+            // octogons have nice angles
+            r: [1, 2, 3, 2, 3, 1, 2, 1, 2],
+            theta: ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'a']
```
alexcjohnson: hmm, not for this PR of course (and would be useful in circular mode as well), but this makes me realize we should have a mode in `scatterpolar` to join the endpoints together so you don't need to repeat the initial point when the data are meant to be periodic. `trace.connectends` perhaps, to mirror `trace.connectgaps`?

src/plots/polar/polar.js
```diff
@@ -532,8 +532,24 @@ proto.updateAngularAxis = function(fullLayout, polarLayout) {
 
     Axes.doTicksSingle(gd, ax, true);
 
+    // angle of polygon vertices in radians (null means circles)
+    // TODO what to do when ax.period > ax._categories ??
```
alexcjohnson: the "angular period > _categories" subplot in your mock looks good - does that mean this TODO is done?

src/plots/polar/polar.js
```diff
@@ -1116,26 +1278,229 @@ function computeSectorBBox(sector) {
...
+    var dcos = clampTiny(Math.cos(v1) - Math.cos(v0));
+
+    if(dsin && dcos) {
+        // given
+        //  g(x) := v0 -> v1 line = m * x + b
+        //  h(x) := mr * x
+        // solve g(xstar) = h(xstar)
+        var m = dsin / dcos;
+        var b = yp - m * xp;
+        var mr = Math.sin(a) / Math.cos(a);
```
alexcjohnson: `Math.tan(a)`

test/jasmine/tests/polar_test.js
etpinard: Good call. Added to -> https://github.com/plotly/plotly.js/issues/2255

src/plots/polar/polar.js
etpinard: That mock _looks_ good, but I'm not it's correct. I still don't what the case `ax.period` > `ax._categories` can and should represent.

etpinard: ... and there's an item about this in https://github.com/plotly/plotly.js/issues/2255

src/plots/polar/polar.js
```diff
@@ -1116,26 +1278,229 @@ function computeSectorBBox(sector) {
...
+            ind = i;
+        }
+    }
+    return ind;
+}
+
+// find intersection of 'v0' <-> 'v1' edge with a 'radial' line
+// (i.e. a line that starts from the origin at angle 'a')
+// given an (xp,yp) pair on the 'v0' <-> 'v1' line
+// (N.B. 'v0' and 'v1' are angles in radians)
```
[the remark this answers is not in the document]
etpinard: > whew, I think I understand the goal of this function: so you have two rays at angles v0 and v1, and given points on each ray at equal radius and a segment connecting them, first find the radius (or the segment) for which that segment intersects [xp, yp]. Then, find the point on a ray at angle a that intersects this segment. Did I get that right?

Yeah, you got it right except that a priori we don't know the points on each ray. We only know their angles. 

> 
If a always matches either v0 or v1 this looks fine, but the else clause looks like it will give the wrong answer if a is anything else (since the norm of the result won't be r then).

Good eye. Thanks! 

When computing the the zoombox coordinates, `a` matches `v0` so no bugs there. When computing the (clipped) polygon coordinates (when `polar.sector` isn't [0,360]) `a` is either `polar.sector[0]` or `polar.sector[1]`.

src/plots/polar/polar.js
```diff
@@ -1116,26 +1278,229 @@ function computeSectorBBox(sector) {
...
+        [x0, m * x0 + t + yp],
+        [x1, m * x1 + t + yp]
+    ];
+}
+
+function makeCycleIndexFn(len) {
+    return function(index) {
+        return index < 0 ? len + index :
+            index < len ? index : index - len;
+    };
```
alexcjohnson: [`Lib.mod`](https://github.com/plotly/plotly.js/blob/master/src/lib/mod.js)?

alexcjohnson:
@etpinard beautiful, your trigonometry teacher would be very proud 🥇
Sounds like https://github.com/plotly/plotly.js/pull/2739#discussion_r197305287 may contain a small bug, otherwise just minor 🐄 comments. Lets do it! 💃

src/plots/polar/polar.js
etpinard: Oh, so nice. Thanks for tip. Sub'ed in -> https://github.com/plotly/plotly.js/pull/2739/commits/1ca65080d138a42442197b9389c04ab9b0cdbeca

src/plots/polar/polar.js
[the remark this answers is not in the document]
etpinard: At lot of little things weren't perfect for non-[0,360] polar sectors. Thanks for bringing this up. https://github.com/plotly/plotly.js/pull/2739/commits/7d405a04a7298f74ef2533b2951b2a1dd72fb3ab?short_path=c565e02#diff-c565e0252037c7b20b8005938d85799a fixes this.

src/plots/polar/polar.js
etpinard: Actually, this isn't so :laughing: , `Math.tan` isn't bounded. I didn't noticed anything wrong before because:

```js
Math.tan(Math.PI / 2)
// => 16331239353195370
```

(in Chrome 67 at least).

Oh well, https://github.com/plotly/plotly.js/pull/2739/commits/7d405a04a7298f74ef2533b2951b2a1dd72fb3ab made things more robust.

alexcjohnson:
Very nice. 💃 stands (or continues dancing?), lets do it!

PR #2752, opened 2018-06-25
Isosurface traces

kig:
Isosurface trace

isosurface_plotly

@etpinard @alexcjohnson 

Isosurface trace for plotly.js integration WIP.

- [x] Add support for axis-aligned clipping planes to construct a clipped version of the isosurface
- [x] Should isocaps be a separate trace or part of the isosurface? Now the isosurface has isocaps turned on and merged into the isosurface mesh. The lib supports breaking them into two and having separate colorscales for each.
- [x] Figure out how to add intensity values to the hover info popup
- [x] How to pass in data, now it's following the streamtube plot method of passing in x, y, z and u arrays, then constructing a meshgrid out of the x,y,z -values and using that as the coords for the u array of intensities. Is this ok?
- [x] make plotly.js compute the trace bounds. They shouldn't be a user setting.


alexcjohnson:
Looking good! @etpinard will likely have more to say when he's back tomorrow, but some thoughts of my own to start:

> Add support for axis-aligned clipping planes to construct a clipped version of the isosurface

👍 the simplest way to do this would be 6 attributes `(x|y|z)(min|max)` and that might be the best in fact - we could do `xrange: [min, max]` or even `range: {x: [min, max]}` but those all seem like they make it harder to do the most common thing, which is to add a single clipping plane so you can see the data there.

> Should isocaps be a separate trace or part of the isosurface? Now the isosurface has isocaps turned on and merged into the isosurface mesh. The lib supports breaking them into two and having separate colorscales for each.

Part of the same trace. My gut reaction is the isosurfaces should by default be colored according to the isocap colorscale, but that you should be free to provide separate colors for each isosurface. What do you mean by "separate colorscales"? Each isosurface just gets a single color, right?

> How to pass in data, now it's following the streamtube plot method of passing in x, y, z and u arrays, then constructing a meshgrid out of the x,y,z -values and using that as the coords for the u array of intensities. Is this ok?

I think the input data format is good, though `u` I might rename to `v` or `value` - Or I guess `intensity` for consistency with `mesh3d` but I don't particularly like that attribute, I wish we had used `value` there.

I'm not sure what you mean by "constructing a meshgrid out of the x,y,z -values and using that as the coords for the u array" though - won't the meshgrid you actually display be some sort of interpolation between the provided (x, y, z) values to find the isovalue?

src/traces/isosurface/attributes.js
```diff
@@ -0,0 +1,125 @@
...
+        valType: 'number',
+        editType: 'calc',
+        description: 'Sets the minimum iso bound of the isosurface.'
+    },
+
+    imax: {
+        valType: 'number',
+        editType: 'calc',
+        description: 'Sets the maximum iso bound of the isosurface.'
+    },
```
alexcjohnson: `imax`/`imin` is OK if we rename `u` to `intensity`... but if it turns to `v` or `value` these should be `vmax`/`vmin`. I presume that if you only provide one of these, the other one is treated as +/-Infinity for the purpose of constructing the isocaps?

src/traces/isosurface/attributes.js
```diff
@@ -0,0 +1,125 @@
...
+
+var extendFlat = require('../../lib/extend').extendFlat;
+
+var attrs = {
+    x: {
+        valType: 'data_array',
+        role: 'info',
+        editType: 'calc+clearAxisTypes',
+        description: [
+            'Sets the x coordinates of the isosurface'
```
alexcjohnson: perhaps "Sets the x coordinates of the volume data" (from which the isosurface will be calculated)

src/traces/isosurface/convert.js
```diff
@@ -0,0 +1,205 @@
...
+
+    isosurfaceOpts.values = trace.u;
+
+    isosurfaceOpts.colormap = parseColorScale(trace.colorscale);
+    // isosurfaceOpts.capsColormap = parseColorScale(trace.capscolorscale);
+    isosurfaceOpts.vertexIntensityBounds = [trace.cmin, trace.cmax];
+    isosurfaceOpts.isoBounds = [trace.imin, trace.imax];
+
+    isosurfaceOpts.isoCaps = trace.isocaps;
+    isosurfaceOpts.singleMesh = trace.singlemesh === undefined ? true : trace.singlemesh;
```
alexcjohnson: What does `singlemesh` do? BTW logic like this belongs as `attributes.singlemesh.dlft = true`.

etpinard:
Thanks for the PRs @kig! Looking great :tada: 

Unfortunately, I won't have time to look at them in detail until we release `1.39.0`.

Now, would you be interested in looking at the broken tubes we noticed in https://github.com/plotly/plotly.js/pull/2658#issuecomment-394836109 to help us merge the streamtube PR?

src/traces/isosurface/attributes.js
etpinard: I remember us having the same concerns about _intensity_ in surface traces. We ended up naming that attribute [`surfacecolor`](https://github.com/plotly/plotly.js/blob/32e6984a46f5c295ca0814aa33d7f066c64ce5d5/src/traces/surface/attributes.js#L127-L132).

I think naming this field `u` or `v` would confusing as it is an array of scalars, not a component of a vector field. I like `value`, but maybe `volume` would better (that's how [MATLAB seems to call this field](https://www.mathworks.com/help/matlab/ref/isosurface.html))?

etpinard: About `imin` and `imax`, MATLAB calls it _isovalue_. Wouldn't `isomin` and `isomax` be better?

etpinard:
In a private convo, @alexcjohnson wrote:

> It looks like it [i.e. `gl-isosurface3d`] is picking the closest grid points inside the min/max range, and making a surface from those points, rather than interpolating to the actual positions of the min and max values.

@kig is this in fact the current behavior?

kig:
> > Should isocaps be a separate trace or part of the isosurface? Now the isosurface has isocaps turned on and merged into the isosurface mesh. The lib supports breaking them into two and having separate colorscales for each.
>
> Part of the same trace. My gut reaction is the isosurfaces should by default be colored according to the isocap colorscale, but that you should be free to provide separate colors for each isosurface. What do you mean by "separate colorscales"? Each isosurface just gets a single color, right?

In some traces, it's nice to have a separate colorscale for the caps. E.g. https://gl-vis.github.io/gl-isosurface3d/smooth_skin.html

isosurface_diff_colorscale

kig:
>I think the input data format is good, though u I might rename to v or value - Or I guess intensity for consistency with mesh3d but I don't particularly like that attribute, I wish we had used value there.

Got it, I'll rename u to v. 

> I'm not sure what you mean by "constructing a meshgrid out of the x,y,z -values and using that as the coords for the u array" though - won't the meshgrid you actually display be some sort of interpolation between the provided (x, y, z) values to find the isovalue?

Hmm, the isosurface now works by running marching cubes on the box lattice defined by |x|,|y|,|z| and u. The result of this is a mesh with vertices at integers <code>[0...|x|-1], [0...|y|-1], [0...|z|-1]</code>. Then the mesh builder moves the vertices to the corresponding x/y/z coords, roughly: <code>verts[i][j][k].xyz = vec3(meshgrid.x[i], meshgrid.y[j], meshgrid.z[k])</code>.

In other words, it constructs a mesh with marching cubes with vertices at the meshgrid positions. (If the meshgrid coords grow at a fixed rate, this amounts to scaling and translating the mesh.)

src/traces/isosurface/convert.js
```diff
@@ -0,0 +1,205 @@
...
+    isosurfaceOpts.dimensions = [xs.length, ys.length, zs.length];
+    isosurfaceOpts.meshgrid = [
+        toDataCoords(xs, 'xaxis'),
+        toDataCoords(ys, 'yaxis'),
+        toDataCoords(zs, 'zaxis')
+    ];
+
+    // var bounds = [
+    //    isosurfaceOpts.boundmin || [xs[0], ys[0], zs[0]],
+    //    isosurfaceOpts.boundmax || [xs[xs.length - 1], ys[ys.length - 1], zs[zs.length - 1]]
```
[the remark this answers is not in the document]
kig: Ok, will do, <code>(x|y|z)(min|max)</code>

alexcjohnson:
> This is the issue with marching cubes.

You should be able to get the desired result using marching cubes, following the example of our 2D SVG `contour` plot, which uses marching SQUARES but extends it:

- First, we do one level at a time, not two... so the binary values are below/above rather than between/outside. That way you can't lose the surface if it transitions too quickly between the two bounds.
- Then we recognize that each of the "crossings" identified just refers to "somewhere on the line segment connecting two adjacent grid points" but we refine that crossing point by linear interpolation between those same two grid points, and use THAT as the mesh vertex, but connect these into a mesh the same way you do currently.

> In some traces, it's nice to have a separate colorscale for the caps. E.g. https://gl-vis.github.io/gl-isosurface3d/smooth_skin.html

Yes, the *caps* get a colorscale. The *isosurfaces*, especially with the refinement ^^, will just get a single color each (one for the low, one for the high) so I don't think it's a good idea to make people pull those colors out of a colorscale, just specify one or two colors.

jackparmer:
@kig - Can you please give us an update on this one?

etpinard:
now in https://github.com/plotly/plotly.js/pull/3311

PR #2753, opened 2018-06-25
Volume traces

kig:
Volume trace

volume_plotly2

@etpinard @alexcjohnson 

Volume rendering trace for plotly.js. Renders 3D volumes as stacks of translucent images. Integration work in progress.

- [x] Volume data is passed in in the streamtube / isosurface fashion: x,y,z, and u arrays of coordinates and intensity, each array has the same length. The x, y, z arrays are used to construct a meshgrid that determines the shape of the u array.
- [x] scene.js transparent pass results in nothing drawn, the trace is now using a hack where it's rendered in the opaque pass.
- [x] There's an opacityscale parameter to control the mapping of intensity values to opacity multipliers, this is passed in as a 256-element array of numbers 0..1. Which is arbitrary and a bit of a pain to construct.
- [x ] When rotating the volume, the rendering jumps when changing the image stack render order. The stack layers may not line up perfectly.
    - This is because the stacks should really be rendered as voxels to get proper z-sorting. Voxels would add up to around 6 triangles per voxel, or 12 million tris for a 128x128x128 volume. Rendering as a point cloud is another option.
- [x] The volume lib represents volumes with textures that have the layer images stacked vertically. This runs into trouble with higher-res volumes, exceeding max texture size.
    - Fixable by using 3D textures, which is easiest in WebGL2. Possible to do with a hack for WebGL1, but that'll require multiple textures for storing the volume.
- [x] <strike>Asymmetric volume dimensions seem buggy, try with 64x32x64 for example</strike>
- [x] <strike>Hover?</strike>
    - Hover doesn't really have a nice mapping here. We could ray march into the volume and look for an intensity value above a hover cutoff (so it'd work a bit like a voxel isosurface wrt picking.)


src/traces/volume/attributes.js
```diff
@@ -0,0 +1,127 @@
...
+        editType: 'calc',
+        description: 'Sets the intensity values of the volume.'
+    },
+
+    opacityscale: {
+        valType: 'data_array',
+        role: 'info',
+        editType: 'calc',
+        description: 'Sets the opacity scale of the volume, which opacity to use for which intensity. Array of 256 values in 0..1 range.'
+    },
```
alexcjohnson: Is this mapping over the same range as `colorscale`? If so, can we specify it similarly, as a piecewise-linear list of pairs `[[0, opacity0], [v1, opacity1], ..., [1, opacityN]]`?

Is this used along with `trace.opacity`? Like the two are multiplied together perhaps? However that works it should be in the description.

src/traces/volume/attributes.js
```diff
@@ -0,0 +1,127 @@
...
+        editType: 'calc',
+        description: 'Sets the maximum intensity bound of the volume.'
+    },
+
+    opacity: {
+        valType: 'number',
+        role: 'info',
+        editType: 'calc',
+        description: 'Sets the opacity of the volume.'
+    },
```
alexcjohnson: `opacity` is a standard attribute for all traces (unless you disable it with `'noOpacity'` in `module.categories`) so you shouldn't need this attribute, nor the `coerce('opacity')`, nor even `trace.opacity === undefined ? 1 : trace.opacity` - you should just be able to do `volumeOpts.opacity = trace.opacity;` since opacity defaults to 1.

src/traces/volume/defaults.js
```diff
@@ -0,0 +1,59 @@
...
+        return;
+    }
+
+    coerce('imin');
+    coerce('imax');
+    coerce('opacity');
+    coerce('opacityscale');
+
+    coerce('boundmin');
+    coerce('boundmax');
```
alexcjohnson: Similar to isosurface, could we change these to `(x|y)(min|max)`? Doesn't look like these are implemented yet, and it doesn't feel to me as though they're as important here as in isosurface, so I'd be OK with removing them for the first version of this trace type.

src/traces/volume/convert.js
```diff
@@ -161,8 +182,6 @@ function convert(gl, scene, trace) {
         trace.imax === undefined ? trace.cmax : trace.imax
     ];
 
-    volumeOpts.opacity = trace.opacity === undefined ? 1 : trace.opacity;
```
alexcjohnson: You can still set this, if it'll get used downstream - which it looks like it will based on your amended description for `opacityscale`, thanks for fleshing that out!

My point was just that you don't need to handle the `undefined` case here, since the global `trace.opacity` attribute has a default of 1 you know at this point that you can only have a number between 0 and 1.

kig: Ohh, right, thanks for catching this.

src/traces/volume/attributes.js
```diff
@@ -0,0 +1,111 @@
...
+var fromMesh3d = ['opacity', 'lightposition', 'lighting'];
+
+fromMesh3d.forEach(function(k) {
+    attrs[k] = mesh3dAttrs[k];
+});
+
+attrs.hoverinfo = extendFlat({}, baseAttrs.hoverinfo, {
+    editType: 'calc',
+    flags: ['x', 'y', 'z', 'intensity', 'text', 'name'],
+    dflt: 'x+y+z+intensity+text+name'
```
etpinard: We should remove `hoverinfo` completely, as we're not implementing a hover handler in this first iteration.

src/traces/volume/convert.js
```diff
@@ -0,0 +1,241 @@
...
+    var dataScale = scene.dataScale;
+    var volumeOpts = {};
+
+    function toDataCoords(arr, axisName) {
+        var ax = sceneLayout[axisName];
+        var scale = dataScale[axisName2scaleIndex[axisName]];
+        return simpleMap(arr, function(v) { return ax.d2l(v) * scale; });
+    }
+
+    var xs = getSequence(trace.x);
```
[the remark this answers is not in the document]
kig: Ooh, good to see. I'll change it to use Lib.distinctVals to have a single reference for this kind of function.

FWIW, getSequence operates on an array of sequences sorted in ascending order and returns the unique values of the first sequence. E.g. [1,1,2,2,3,3,1,1,2,2,3,3] -> [1,2,3]. So it is a bit more performant for this kind of data (no need to sort, can bail out after the first sequence), but I don't think it's worth the maintenance burden.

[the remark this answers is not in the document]
etpinard: Thanks for the info @kig !

> I don't think it's worth the maintenance burden.

I think so too. :+1:

kig:
@etpinard The gl-texture2d failure is likely because a 128x128x128 volume tries to create a 16kpx high texture. 64x64x64 _should_ work. I'm working on a proper fix to this issue by laying out the slices on a grid (and eventually using that as a raymarchable volume).

etpinard:
> The gl-texture2d failure is likely because a 128x128x128 volume tries to create a 16kpx high texture. 64x64x64 _should_ work. I'm working on a proper fix to this issue by laying out the slices on a grid (and eventually using that as a raymarchable volume).

Awesome :ok_hand:

etpinard:
@archmoj could you try pulling down this branch, `npm i`, `npm start` and opening up mock `gl3d_volume-simple`. I'm getting

image

and I'd like to confirm I'm not the only one.

archmoj:
@etpinard Yes I am getting the same errors on Ubuntu & Windows too.

Issue #2687, opened 2018-05-31
Update( ) fails if the user is interacting with the plot


Watashiku:
Hi !

I use an heatmap to represent large datasets. After plotly_relayout events I download and process data (asynchronous, can take some time), then display them. However Plotly will return an undefined error from a promise if the update is done while dragging. I used a setTimeout in the codepen to copy the behavior of my program.

CodePen : https://codepen.io/watashiku/pen/pKJJNW?editors=1010
Error message : 
```
plotly-latest.js:128472 Uncaught TypeError: Cannot read property 'length' of undefined
    at exports.drawData (VM750 plotly-latest.js:128472)
    at Object.lib.syncOrAsync (VM750 plotly-latest.js:119125)
    at relayout (VM750 plotly-latest.js:125433)
    at Object.exports.call (VM750 plotly-latest.js:153164)
    at VM750 plotly-latest.js:133800
    at Object.lib.syncOrAsync (VM750 plotly-latest.js:119125)
    at Object.dragTail [as doneFn] (VM750 plotly-latest.js:133798)
    at HTMLDocument.onDone (VM750 plotly-latest.js:102144)
```
Steps to reproduce : drag the plot once, then quickly start dragging again and hold it one second - the bug appears if you are dragging during the update triggered by the previous drag. 

It creates two major problems :
 - The plot will be "locked" on the screen : panning or zooming will not affect the plot (only the axes in the background),
 - There is no plotly_afterplot event triggered by Plotly which means there is no way for me to tell the user that the plot is ready nor somehow handling the situation by myself (and the error cannot be catched outside of Plotly due to the way it is raise I think?).

Notes : 
 - When the bug happens clicking on the graph fixes it
 - The behavior is similar for react, restyle and addtrace. It does not appear with plot (but this creates traces for every relayout)

alexcjohnson:
Looks like the same root issue as #2643, or perhaps #2644. Regardless we clearly need some better solutions for updating a plot that the user may already be interacting with...

Watashiku:
Yes those issues are very similar !

Is there a way for us to detect when the issue happens until it is fixed ? I did not find a catch case on the promise that fails. The only thing I found is catching all the document.windows error and look for an undefined one :confused:

jacobq:
Is this the same problem I am running into in this example?
https://codepen.io/jacobq_/pen/KexEEB

Simply clicking somewhere on the plot area while it's running generates errors like these:
[...]

Issue #2644, opened 2018-05-17
errors during react() while drag-selecting


nicolaskruchten:
I'm trying to upgrade my crossfilter example to be more declarative by using `react()` but seeing errors. This is my attempt: https://github.com/plotly/plotly.js-crossfilter.js/blob/23b1ff9bd8dc0a3923caf29e4d3feb960ec31234/script.js

While drag-selecting on top of the Infant Mortality chart, the dashed-line selection box does not appear, and errors appear in the console:

```
plotly-latest.min.js:7 Uncaught (in promise) TypeError: Cannot read property '0' of undefined
    at Object.m.supplyDefaultsUpdateCalc (plotly-latest.min.js:7)
    at Object.r.react (plotly-latest.min.js:7)
    at redraw (script.js:64)
    at n.hist_im_select (script.js:121)
    at n.emit (plotly-latest.min.js:7)
    at HTMLDivElement.t.emit (plotly-latest.min.js:7)
    at plotly-latest.min.js:7
```

etpinard:
that one sounds similar to https://github.com/plotly/plotly.js/issues/2602, might have been fixed https://github.com/plotly/plotly.js/pull/2603

nicolaskruchten:
I'm using the latest CDN build so I don't think that's it. It would be really nice for this to at least have the same behaviour as in #2643 ... /cc @alexcjohnson  :)

nicolaskruchten:
Ah, for clarity, here is the diff between the two cases: https://github.com/plotly/plotly.js-crossfilter.js/commit/23b1ff9bd8dc0a3923caf29e4d3feb960ec31234

alexcjohnson:
Alright, managed to reproduce and understand what's going on here:
- The user starts a selection on the infant mortality plot
- The event callback calls `Plotly.react` on the same plot, giving it new data arrays.
- This change triggers a "recalc" ie the most complete redraw we do - and the signal we've used for `Plotly.plot` to know to recalc is to delete `gd.calcdata`.
- But `Plotly.plot` bails out because you're in the middle of a drag:

https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L139-L145

- Then the *next* time we call `Plotly.react` the data arrays may have *not* changed, so we don't delete `gd.calcdata`, we expect it to be there to be relinked to the modified traces.
[...]

Issue #2643, opened 2018-05-17
undefined being thrown while redrawing with plotly_selecting


nicolaskruchten:
Crossfiltering works great with `plotly_selecting` except that I had to add some `catch(function(){})` to suppress errors I was seeing. https://github.com/plotly/plotly.js-crossfilter.js/commit/c48b5beaf1eafbb4c5f68a1923084607c61591c5

Would be great to not have to catch-and-drop `undefined`.

here's the error I was seeing, sorry for the minification:

```
Uncaught (in promise) undefined
Promise.then (async)
r.redraw @ plotly-latest.min.js:7
t @ plotly-latest.min.js:7
redraw @ script.js:68
hist_im_select @ script.js:120
n.emit @ plotly-latest.min.js:7
t.emit @ plotly-latest.min.js:7
(anonymous) @ plotly-latest.min.js:7
l @ plotly-latest.min.js:7
r.throttle @ plotly-latest.min.js:7
i.moveFn @ plotly-latest.min.js:7
M @ plotly-latest.min.js:7
```

alexcjohnson:
I suspect it's this: 

https://github.com/plotly/plotly.js/blob/fb751582506a4d71d23a89786ba8138473e33cd9/src/plot_api/plot_api.js#L139-L145

Are you adding/deleting traces to the same plot that the selection is happening on? I guess we're rejecting here because the plot hasn't actually redrawn yet, so if you expected to be able to do something like look at `gd.data` or the DOM, it would not be in the correct state. We should at least add a message to the rejection stating the reason. Alternatively, since we're not rejecting due to an error but just due to the plot not being done yet, we could imagine returning a promise that only resolves when the plot *is* done... wouldn't affect your usage but I worry that this could potentially deadlock existing apps. Dunno quite how it would work but some combination of event bindings that mean two plots are simultaneously waiting for each other to finish?

nicolaskruchten:
Yes, I'm adding/removing to the same chart that's being dragged on, so as to colour them. Maybe I should restyle?

I should add that the error message is kinda gross, throwing `undefined`... Is this normal on `Promise.reject()` ?

alexcjohnson:
> Yes, I'm adding/removing to the same chart that's being dragged on, so as to colour them. Maybe I should restyle?

[...]

PR #2757, opened 2018-06-27
Add guards for unsafe dereferencing (ref #2687)

jacobq:
* In drawData
* In supplyDefaultsUpdateCalc

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -484,7 +484,7 @@ plots.supplyDefaults = function(gd, opts) {
 plots.supplyDefaultsUpdateCalc = function(oldCalcdata, newFullData) {
     for(var i = 0; i < newFullData.length; i++) {
         var newTrace = newFullData[i];
-        var cd0 = oldCalcdata && oldCalcdata[i] && !oldCalcdata[i][0];
+        var cd0 = oldCalcdata && oldCalcdata[i] && oldCalcdata[i][0];
         if(cd0 && cd0.trace) {
             var oldTrace = cd0.trace;
             if(oldTrace._hasCalcTransform) {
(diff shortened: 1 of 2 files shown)


jacobq:
OK, I think this is ready for review (I'm done doing `git push -f` here ;))

etpinard:
Thanks very much for the PR!

I tested out https://codepen.io/watashiku/pen/pKJJNW?editors=1010 (from https://github.com/plotly/plotly.js/issues/2687#issue-328179735) off your branch and your patch didn't appear to fix the issues reported, it only hides the console errors.

Moreover, your patch made a few of tests fail.

I hope I'm wrong, but I think fixing the issues of https://github.com/plotly/plotly.js/issues/2687, #2643, and #2644 will have to wait a multi-week effort to implement https://github.com/plotly/plotly.js/issues/2644#issuecomment-391152782

src/plots/plots.js
```diff
@@ -484,7 +484,7 @@ plots.supplyDefaults = function(gd, opts) {
 plots.supplyDefaultsUpdateCalc = function(oldCalcdata, newFullData) {
     for(var i = 0; i < newFullData.length; i++) {
         var newTrace = newFullData[i];
-        var cd0 = oldCalcdata[i][0];
+        var cd0 = oldCalcdata && oldCalcdata[i] && !oldCalcdata[i][0];
```
jacobq: Eek, I just noticed a really bad typo here: there shouldn't be a `!`

jacobq:
@etpinard Please check/re-run that latest CI build. It looks to me like the failures aren't being caused by the changes in this PR (got stuck on "install dependencies" step). I'm curious to see if the test failures you saw earlier were because of my typo (had accidentally put a `!` in part of the sanity checking code, rendering it completely wrong).

You are right that this PR doesn't resolve the larger issue. It's just intended to prevent uncaught exceptions.

etpinard:
> It's just intended to prevent uncaught exceptions.

... and that's why I'd vote :-1: on merging this patch.

jacobq:
If you don't want to merge it that's fine, but I don't understand how broken code is better than gracefully broken code.

etpinard:
> but I don't understand how broken code is better than gracefully failing code.

Because I don't think adding fallbacks to these broken things doesn't make plotly.js more graceful in my mind. I could be wrong though.

Issue #2663, opened 2018-05-24
Calling Plots.resize on a (very) small <div> leads to SVG errors


etpinard:
See https://codepen.io/etpinard/pen/KRjKpg?editors=0010

PR #2758, opened 2018-06-27
Sanitize margin after 'autosize' relayouts

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2663 (aka the react-plotly.js issue that resulted in negative svg mesures)

There's probably a _better_ fix out there, but our autosize logic is such a mess that I was happy with this here. I hope @alexcjohnson will agree.

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -345,6 +345,7 @@ plots.supplyDefaults = function(gd, opts) {
 
         if(!newLayout.width) newFullLayout.width = oldWidth;
         if(!newLayout.height) newFullLayout.height = oldHeight;
+        plots.sanitizeMargins(newFullLayout);
     }
     else {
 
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -357,7 +358,7 @@ plots.supplyDefaults = function(gd, opts) {
             initialAutoSize = missingWidthOrHeight && (autosize || autosizable);
 
         if(initialAutoSize) plots.plotAutoSize(gd, newLayout, newFullLayout);
-        else if(missingWidthOrHeight) plots.sanitizeMargins(gd);
```
alexcjohnson: 🤦‍♂️

alexcjohnson:
a mess indeed. This looks good to me! 💃

Issue #2469, opened 2018-03-12
plotly.js theme / style extraction and application


bpostlethwaite:
initial requirements
============

1. Generic style only. Items that apply to any data setup such as trace color, size, background, and fonts are applied. 
2. The template data shall be stored without any design point data. For example, the x and y source data shall not be included. If an attribute such as color, size, or opacity is bound to a variable (has individual values), then the individual values will not be stored.
3. The template shall be applied to an existing graph (a new graph is not generated).
4. A mechanism shall be enabled through the API to report items from the template that do not apply. Example 1) The template has a different number of traces than the plot to which it is being applied. Example 2) The template is from a line plot and being applied to a heat map, therefore the point glyph cannot be applied.

cc @alexcjohnson @etpinard for implementation discussions
cc @chriddyp @sglyon for possible applications to and requirements of `plotly.[py|jl]` and `Dash`
cc @nicolaskruchten @VeraZab for general input.

To get the conversation rolling how about I take a stab at an API:
```javascript
const figureStyle = plotly.extractStyle(gd || {data, layout}, config)
const info = plotly.applyStyle(gd || {data, layout}, {figureStyle [, config]} [, config])
```
where `info` is an object containing a subset of the structure in `figureTheme` documenting each item that was not applied and possibly an optional `reason` why it was not applied.

note: It would be great if this operation was immutable however point 3. requires mutation. Perhaps someone has suggestions for an API that allows both immutable application _and_ mutation.

bpostlethwaite:
Okay how about:

```javascript
const figureStyle = plotly.extractStyle(gd || {data, layout}, config)

// applyStyle does not mutate the gd or data/layout you pass in or update the plot. Simply returns a cloned and updated {data, layout} containing styling information and an info object. Useful for using Plotly.js to build themed templates rather than operate directly on a plot. The config can control whether the data arrays are cloned or referenced.
const {info, figureWithStyle} = plotly.applyStyle({data, layout}, {figureTheme [, config]} [, config])

// mutates the {data, layout} you pass in and then calls plotlyjs.react to update the plot.
const {info, figureWithStyle} = plotly.updateStyle(gd, {figureTheme [, config]} [, config])
```

alexcjohnson:
[...]

PR #2761, opened 2018-06-27
Templates

alexcjohnson:
For #2469 

The first 6 commits - up through https://github.com/plotly/plotly.js/commit/3dee25c4e1e1f35bfaab06554bec434a23f81348 - are essentially prep work, mostly standardizing and simplifying our handling of container arrays.

Then there's one large commit https://github.com/plotly/plotly.js/commit/5cca8d399c10cc3a4ad79136471c4202653a69e1 that implements the coerce-level templating that I settled on, after first trying to merge templates into `data` and `layout` directly at the beginning of `supplyDefaults`, which failed for a variety of reasons.

After that are a bunch of cleanup commits, like https://github.com/plotly/plotly.js/commit/8e2a3210adb53679f095585c06ef705ab42c30b6 making array objects (which may not exist in `layout` if they came from named template items) work with GUI editing.

TODO:
- [x] Jasmine tests - the committed ones right now are still just the skeleton from @bpostlethwaite's initial work and don't actually apply.
- [x] Checking if the template is compatible with the plot it was applied to. This is a bit ill-defined, but I guess means mostly do all the trace types and layout containers fit what's in the template.

@etpinard anything here look problematic to you?


etpinard:
cc @nicolaskruchten this is happening :tada:

src/plots/cartesian/layout_attributes.js
```diff
@@ -513,6 +513,16 @@ module.exports = {
     tickformatstops: {
         _isLinkedToArray: 'tickformatstop',
 
+        visible: {
```
etpinard: Maybe `enabled` (similar to transforms) here would make more sense.

test/jasmine/tests/mapbox_test.js
```diff
@@ -605,6 +612,20 @@ describe('@noCI, mapbox plots', function() {
...
+        .then(function() {
+            expect(getLayerLength(gd)).toEqual(1);
+            expect(countVisibleLayers(gd)).toEqual(1);
+
+            // re-show it
+            return Plotly.relayout(gd, 'mapbox.layers[0].visible', true);
+        })
+        .then(function() {
+            expect(getLayerLength(gd)).toEqual(2);
+            expect(countVisibleLayers(gd)).toEqual(2);
```
etpinard: Nice tests. Thanks!

src/components/rangeselector/button_attributes.js
```diff
@@ -10,6 +10,12 @@
 
 
 module.exports = {
+    visible: {
```
etpinard: Brilliant commit :sparkles:

src/traces/parcoords/merge_length.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * mergeLength: set trace length as the minimum of all dimension data lengths
+ *     and propagates this length into each dimension
+ *
+ * @param {object} traceOut: the fullData trace
+ * @param {Array(object)} dimensions: array of dimension objects
+ * @param {string} dataAttr: the attribute of each dimension containing the data
+ * @param {integer} len: an already-existing length from other attributes
+ */
+module.exports = function(traceOut, dimensions, dataAttr, len) {
```
etpinard: Nicely done :1st_place_medal:

src/components/annotations/defaults.js
```diff
@@ -9,15 +9,93 @@
 
 'use strict';
 
+var Lib = require('../../lib');
+var Axes = require('../../plots/cartesian/axes');
 var handleArrayContainerDefaults = require('../../plots/array_container_defaults');
-var handleAnnotationDefaults = require('./annotation_defaults');
```
etpinard: YES. Thanks for doing this :100:

src/plots/polar/layout_defaults.js
```diff
@@ -49,6 +49,10 @@ function handleDefaults(contIn, contOut, coerce, opts) {
         }
 
         var axIn = contIn[axName];
+        // Note: does not need template propagation, since coerceAxis is still
+        // based on the subplot-wide coerce function. Though it may be more
+        // efficient to make a new coerce function, then we *would* need to
+        // propagate the template.
```
etpinard: thanks for the note :books:

src/plots/array_container_defaults.js
```diff
@@ -34,7 +34,6 @@ var Template = require('../plot_api/plot_template');
  *          - parentObj {object} (as in closure)
  *          - opts {object} (as in closure)
  *          - itemOpts {object}
- *              - itemIsNotPlainObject {boolean}
```
etpinard: Nice cleanup!

src/plot_api/plot_template.js
```diff
@@ -12,7 +12,56 @@
 var Lib = require('../lib');
 var plotAttributes = require('../plots/attributes');
 
-var TEMPLATEITEMNAME = exports.TEMPLATEITEMNAME = 'templateitemname';
+var TEMPLATEITEMNAME = 'templateitemname';
```
etpinard: I'm usually not a fan of three-word attribute names, but I can't really think of a better alternative, so :+1:

src/plots/cartesian/layout_attributes.js
etpinard: .... but maybe it isn't worth breaking symmetric with all other templated arrays :thinking:

alexcjohnson: Right, `enabled` certainly makes more sense here. I thought about that but deferred to symmetry at the time. But the way this ended up working out, it would be easy to make `visibilityAttr: 'enabled'` or something an option to `handleArrayContainerDefaults`. I'll give that a try.

alexcjohnson: switched to `enabled` in https://github.com/plotly/plotly.js/pull/2761/commits/fb489aa0fcd2c526335f09d2068f3659683ad10d

etpinard: I'm glad this wasn't too hard to implement :tada:

src/plots/cartesian/layout_attributes.js
```diff
@@ -61,6 +61,11 @@ module.exports = {
         dflt: '-',
         role: 'info',
         editType: 'calc',
+        // we forget when an axis has been autotyped, just writing the auto
+        // value back to the input - so it doesn't make sense to template this.
+        // TODO: should we prohibit this in `coerce` as well, or honor it if
+        // someone enters it explicitly?
+        _noTemplating: true,
```
alexcjohnson: I prevented `makeTemplate` from pulling out `axis.type` with this flag, as it seems more likely to cause problems than fix them, but I did NOT prevent `coerce` from using it if someone makes a template of manually and includes it (to say "always make this axis categorical" for example). Seem reasonable?

src/plot_api/validate.js
```diff
@@ -397,6 +397,8 @@ function isInSchema(schema, key) {
 }
 
 function getNestedSchema(schema, key) {
+    if(key in schema) return schema[key];
+
```
alexcjohnson: Not part of templating, just noticed it as there's similar logic in templates and I was looking to see how you handled it here 🙈

test/image/mocks/shapes.json
```diff
@@ -18,7 +18,7 @@
         "yaxis2":{"title":"category","range":[0,1],"domain":[0.6,1],"anchor":"x2","type":"category","showgrid":false,"zeroline":false,"showticklabels":false},
         "height":400,
         "width":800,
-        "margin": {"l":20,"r":20,"top":10,"bottom":10,"pad":0},
```
alexcjohnson: I didn't want to change the image so just deleted these bad attrs uncovered when testing this fix.

Perhaps we should make a test that validates all our mocks, to ensure we're not accidentally encouraging mistaken usage like this? Not in this PR of course...

src/plot_api/plot_template.js
```diff
@@ -203,6 +203,8 @@ exports.arrayTemplater = function(container, name, inclusionAttr) {
         // it's explicitly marked visible - in which case it gets NO template,
         // not even the default.
         out[inclusionAttr] = itemIn[inclusionAttr] || false;
+        // special falsy value we can look for in validateTemplate
+        out._template = false;
```
alexcjohnson: This case is just special because we were told to look for a template by name and it wasn't found.

Otherwise `_template` can be `null` (no template was found but that's fine) or `undefined` (we don't need to propagate the template into this container).

src/plot_api/validate.js
```diff
@@ -38,7 +38,7 @@ var isArrayOrTypedArray = Lib.isArrayOrTypedArray;
  *  - {string} msg
  *      error message (shown in console in logger config argument is enable)
  */
-module.exports = function valiate(data, layout) {
+module.exports = function validate(data, layout) {
```
etpinard: good :eye:

test/image/mocks/shapes.json
etpinard: > Perhaps we should make a test that validates all our mocks, to ensure we're not accidentally encouraging mistaken usage like this?

That's a great idea :+1:

src/plot_api/template_api.js
```diff
@@ -269,3 +269,187 @@ function getNextPath(parent, key, path) {
...
+    var fullLayout = figure._fullLayout;
+    var fullData = figure._fullData;
+
+    if(!isPlainObject(layoutTemplate)) {
+        errorList.push({code: 'layout'});
+    }
+    else {
+        // TODO: any need to look deeper than the first level of layout?
+        // I don't think so, that gets all the subplot types which should be
+        // sufficient.
```
etpinard: What about `xaxis.rangeslider.yaxis2` from https://github.com/plotly/plotly.js/pull/2364?

src/plot_api/template_api.js
```diff
@@ -269,3 +269,187 @@ function getNextPath(parent, key, path) {
...
+ * Note that not every issue we identify here is necessarily a problem,
+ * it depends on what you're using the template for.
+ *
+ * @param {object|DOM element} figure: the plot, with {data, layout} members,
+ *     to test the template against
+ * @param {Optional(object)} template: the template, with its own {data, layout},
+ *     to test. If omitted, we will look for a template already attached as the
+ *     plot's `layout.template` attribute.
+ *
+ * @returns {array} array of error objects each containing:
```
etpinard: THanks for making this have a similar API to `Plotly.validate` !

alexcjohnson: One difference is I didn't make the object fields consistent - they all have `code` but each error, in fact even more specific than each `code`, includes whatever extra data it deems relevant. Given the nature of the issues (some apply to a specific trace, some to a path, some to a trace type but not a specific trace, etc... as opposed to `Plotly.validate` where you can basically always trace the issue back to a specific path) it seemed a bit to constraining to give them all the same fields.

src/plot_api/template_api.js
alexcjohnson: Hmm, good point. You're right, though it's a lot of complication (for example for `xaxis.rangeslider.yaxis` we'd have to look for `xaxis<N>.rangeslider.yaxis<M>` for any combination of `<N>` and `<M>` and only generate an error if we don't find *any* of them. And the only time this would alert you that the template isn't appropriate is if the template has a `xaxis<N>.rangeslider.yaxis<M>` but does NOT have `yaxis<M>` - ie your template had opinions about how that y axis should work inside the rangeslider, but left the y axis itself untouched. This is of course possible, just seems very unlikely.

src/plots/cartesian/layout_attributes.js
```diff
@@ -60,6 +61,11 @@ module.exports = {
         dflt: '-',
         role: 'info',
         editType: 'calc',
+        // we forget when an axis has been autotyped, just writing the auto
+        // value back to the input - so it doesn't make sense to template this.
+        // Note: we do NOT prohibit this in `coerce`, so if someone enters a
+        // type in the template explicitly it will be honored as the default.
+        _noTemplating: true,
```
etpinard: [`polar.angularaxis.type`](https://github.com/plotly/plotly.js/blob/c2457ff7d2a3189d35a190b9e4bb315f49dc080d/src/plots/polar/layout_attributes.js#L139-L156) does not inherit from `type` in  `cartesian/layout_attributes.js`. Should it also have `_noTemplating: true`?

src/plot_api/template_api.js
etpinard: > This is of course possible, just seems very unlikely.

Right. I don't this needs special handling. A comment about this would suffice.

test/jasmine/tests/template_test.js
```diff
@@ -0,0 +1,338 @@
...
+        checkValidate(noDataMock, [{
+            code: 'data',
+            msg: 'The template has no key data.'
+        }],
+        // check only the first error - we don't care about the specifics
+        // uncovered after we already know there's no template.data
+        1);
+    });
+
+    it('catches missing template.data', function() {
```
etpinard: :cow2: catches missing _template.layout_.

src/plot_api/template_api.js
alexcjohnson: actually wasn't all that hard - recursed into layout in https://github.com/plotly/plotly.js/pull/2761/commits/15931cf17ebb77baa108b85d80237bfd40cec700

src/plots/cartesian/layout_attributes.js
alexcjohnson: good call - `angularaxis.type` gets `_noTemplating` in https://github.com/plotly/plotly.js/pull/2761/commits/b1c6f0ae0c0d47dab78df8255cd373316a3c3a97

test/jasmine/tests/template_test.js
alexcjohnson: good catch! https://github.com/plotly/plotly.js/pull/2761/commits/818cac94c8c96dc9fe39eef143f605ea12ee5966

etpinard:
Fantastic. :dancer:

PR #2762, opened 2018-06-28
Revamp of mode bar's createIcon

vdh:
As a fix for #1335, a backwards-compatible change to `createIcon` in [modebar.js](https://github.com/plotly/plotly.js/blob/master/src/components/modebar/modebar.js).

1. If an existing `icon` config is given, it behaves the same as before.
2. If an `icon` config defines `height`, it will be used instead of `ascent` and `descent`.
3. If an `icon` config defines `transform`, it will be used instead of the default transform.
4. If an `icon` config doesn't define either a `transform` or an `ascent`, it will not be transformed at all.
5. As an extra bonus, if an `icon` config is a function it will be called instead of `createIcon`, in the case where someone wants to completely customise the icon element generation (i.e. inline their own SVGs, or another element).

The [pull_font_svg.js](https://github.com/plotly/plotly.js/blob/master/tasks/util/pull_font_svg.js) util was also updated to regenerate a new [ploticon.js](https://github.com/plotly/plotly.js/blob/master/build/ploticon.js), containing `height` and `transform` props instead of the previous `ascent` and `descent` props. Assuming that this icon data is only used for its intended purpose in the mode bar (i.e. not used/manipulated manually for anything else by users), this shouldn't cause any breakages.


src/components/modebar/modebar.js
```diff
@@ -155,7 +155,13 @@ proto.createButton = function(config) {
     button.setAttribute('data-toggle', config.toggle || false);
     if(config.toggle) d3.select(button).classed('active', true);
 
-    button.appendChild(this.createIcon(config.icon || Icons.question, config.name));
+    var icon = config.icon;
+    if(typeof icon === 'function') {
+        button.appendChild(icon());
```
etpinard: Why would `config.icon` ever be a function?

src/components/modebar/modebar.js
```diff
@@ -169,7 +175,9 @@ proto.createButton = function(config) {
  * @Return {HTMLelement}
  */
 proto.createIcon = function(thisIcon, name) {
-    var iconHeight = thisIcon.ascent - thisIcon.descent,
+    var iconHeight = thisIcon.height !== undefined ?
```
etpinard: Let's be more strict here:

```js
isNumeric(this.height) ? Number(thisIcon.height) : thisIcon.ascent - this.decent
```

vdh:
@etpinard I've added some tests, how do they look?

src/components/modebar/modebar.js
etpinard: We like to keep our API JSON-serializable, unless we have to (like modebar button click handlers), so I'd vote :-1: 

The way you implemented the use case you described seems a little hacky. Maybe instead we could add a field named e.g `activeIcon` that would be used when `active` class is turned on?

etpinard:
> I've added some tests, how do they look?

Your tests look amazing. Thanks very much!

src/components/modebar/modebar.js
vdh: @etpinard
This issue arose because the current custom icon support is *extremely* tailored and twisted to *only* support the structure that the Plotly icons have been stored in. I mean, you've got:

- Forced transformation inversion of the entire image… just because…?
- Limited to only one path element
- The `viewBox` is pieced together from disparate config variables
- `ascent` and `descent` is never documented anywhere, leaving poor users to guess at (or poke at source code) to even understand how this will change the resulting SVG.
- The fact that the Javascript is manually building the SVGs piece by piece in the first place, instead of loading the SVGs in directly, or applying them via CSS…
- A custom transformation being applied to one special icon, but not actually via the **icon's** own config… no… it's actually applied via **the name** of the button config that just so happens to use that icon…???

I've tried very hard with this PR to keep everything supported, smooth over some of those wrinkles, while also opening it up enough for sane custom icons… so why is this small `function`-based escape hatch so bad? Amongst a config *already* full of existing support for functions?

What if someone wants to use SVG sprite sheets? What if they just want to load in their own SVG source files without jumping through hoops? What if they want to use CSS-based icon sets, to load the images in via their existing asset systems? Or heaven forbid, use *more* than just one path element?! Arbitrarily forcing your users to be restricted to one little use case for their SVGs just to force a JSON structure is maddening! It's not even like I'm forcing this option either, the JSON structure option *still* works just fine!

I'm sorry to get so heated about this, but calling it "a little hacky" is pot calling the kettle black 😢

alexcjohnson: Haha @vdh you raise some good points. FWIW most of the quirks you mention above are due to the structure of SVG fonts - for example https://stackoverflow.com/questions/14203365/why-svg-font-is-mirrored - because that's where we do our own design in the first place, and at least at one point we were also using these icons as a font, though I'm not sure we're still doing that anywhere... not that this justifies an obtuse API, just to explain how we got here.

You're right that `config` items don't have a *hard* requirement for JSON (unlike `data` and `layout` where we cannot allow functions, for portability). @etpinard has good reason to encourage JSON, as it makes cross-language integrations (Python, R, etc) much cleaner. And, pots and kettles notwithstanding, new API gets held to a higher standard than old locked-in API, #420 is bloated enough already 😉 

But I have to say I can see the appeal of a more flexible arrangement, and I don't really see a good alternative here. Multiple paths is a good example, what if you want regions with different colors or something? The use cases you've mentioned could be covered by a string of raw HTML, but I don't think that's better, as it could open an XSS vector. So yeah, I think in this case a function may be our best option.

src/components/modebar/modebar.js
```diff
@@ -178,12 +187,19 @@ proto.createIcon = function(thisIcon, name) {
...
-    path.setAttribute('transform', transform);
+
+    if(thisIcon.transform) {
+        path.setAttribute('transform', thisIcon.transform);
+    }
+    else if(thisIcon.ascent !== undefined) {
+        // Legacy icon transform calculation
+        var transform = name === 'toggleSpikelines' ?
+            'matrix(1.5 0 0 -1.5 0 ' + thisIcon.ascent + ')' :
+            'matrix(1 0 0 -1 0 ' + thisIcon.ascent + ')';
```
alexcjohnson: Thanks for moving `pull_font_svg` to the new `transform` property - it's definitely cleaner to calculate that upfront - and for maintaining the legacy version to support any users who *did* go to the trouble of figuring out `ascent` and `descent`. But I don't think we need to carry the `toggleSpikelines` hack over here - lets take it out, then we don't even need `name` in `createIcon` 🎉

src/components/modebar/modebar.js
etpinard: @vdh sorry, I didn't mean to criticize your implementation by calling it _hacky_ Finding the right hack can be the smartest thing one can do in some situations. But we can't expect every plotly.js user to be come up with that hack. I was simply wondering if there was another (I won't call it less hacky, say more user-intermediate-user-friendly) solution to your use case.

etpinard: > Multiple paths is a good example, what if you want regions with different colors or something? The use cases you've mentioned could be covered by a string of raw HTML, but I don't think that's better, as it could open an XSS vector. So yeah, I think in this case a function may be our best option.

I can get behind this (at least as a solution for v1).

:ok_hand: for `typeof icon === 'function'`.

alexcjohnson:
@etpinard @vdh I removed the legacy `toggleSpikeLines` hack as I mentioned https://github.com/plotly/plotly.js/pull/2762#discussion_r199466432 -> https://github.com/plotly/plotly.js/pull/2762/commits/586ff25c8966e880b8cffb2f28ad6d857d8380ab
If you're OK with this, I think we're ready to go! 💃

etpinard:
Thanks @alexcjohnson ! Merging this thing :dancer:

palimondo:
Heh, this broke my code temporarily, but the fix was just to replace my own `iconHeight` computation from `ascent` and `descent` with `height` and using the precomputed `transform` instead of building one myself from `ascent`… so I just wanted to say kudos! This is much saner serialization in the `json` alone (I haven't looked at the rest of changes...).

Issue #2740, opened 2018-06-16
plotly_afterplot Event isn't triggered on pan/zoom


JOE95443:
In previous versions of Plotly.js (1.35.2 works) I would use the `plotly_afterplot` event to detect when a user had panned the graph, but this event is no longer being called when the user pans or zooms.

The issue can be reproduced using your plotly_afterplot [example](https://codepen.io/plotly/pen/YeGGdy) on Codepen. The event only seems to be called when the plot is first drawn and when the axes are reset. I've tested this on the latest version of both Chrome and Safari on MacOS; neither of which have events called from simply panning or zooming the plot.

alexcjohnson:
This is likely due to our performance improvements, so axis range changes don’t require a full replot anymore. The `plotly_relayout` event is really the one you want anyway, as it also tells you what the change was.

JOE95443:
Thank you, I can confirm `plotly_relayout` works as expected. I think the documentation for `plotly_afterplot` needs updating and then this issue is resolved.

alexcjohnson:
Good call @JOE95443 - the example at https://plot.ly/javascript/plotlyjs-events/#afterplot-event no longer acts as described.

@etpinard I seem to recall us discussing this before and concluding we no longer even need afterplot, you can just use eg `Plotly.newPlot(...).then(...)` to accomplish its original purpose (and as this issue shows, the other times it currently fires it really shouldn't be used). So should we update the docs to describe it as deprecated and add this to #420?

alexcjohnson:
It's true, if there were one event that always fired after any change to the plot, that could be useful. But what if in v2 this event was `plotly_react`?

etpinard:
> But what if in v2 this event was plotly_react?

Aren't we planning on renaming `Plotly.react` -> `Plotly.plot`  in v2?

alexcjohnson:
> Aren't we planning on renaming `Plotly.react` -> `Plotly.plot`  in v2?

sure :) Alright, fiddling a little it seems like `plotly_afterplot` was actually closer to firing after every GUI-initiated change than I thought, so perhaps it *is* worth extending it to be the one "any change of state" event and then renaming it (amid coalescing with other events) in v2.

etpinard:
> it to be the one "any change of state" event

Sounds great :ok_hand:

Issue #2704, opened 2018-06-06
Automargin pipeline


alexcjohnson:
Our automatic margin expansion system is a mess: components generally determine their sizes - and thus the amount they need to increase the margin - only while being drawn. So in order for the new margin to take effect, the component gets drawn, which updates the margin calculation, then all of these components need to be drawn again if the new margin is different from the old one. The situation is even worse for axis automargins, since axes are drawn later in the process, changes they make to the margins result in an entire redraw of the plot, ie nearly doubling the initial plot time.

Much better would be to separate out margin calculations from component drawing; that way we can determine all sizes before we draw anything, then each component need only be drawn once. The challenge with this is that these sizes depend on text bounding boxes, so in order for this not to add its own overhead we may need to create the text elements as part of the margin calculation step, stash them somewhere hidden, and then pull them back out during the drawing step and attach them to the component. Or perhaps the two steps are 1) component creation and sizing, 2) component positioning and updating?

Some related discussion happened during #2681. See also #1988 and #2434.

nicolaskruchten:
> we can determine all sizes before we draw anything, then each component need only be drawn once

This wasn't possible the first time round because as the margins change, some components change how they draw themselves... for example as the left margin increases, the x-axis tick labels compress and then draw themselves on an angle.

etpinard:
After an investigation, I'm realising that our "pipeline problems" are a little deeper than first expected, specifically for cartesian axes.

In brief, cartesian auto-margin push values depend on the axis range (as the axis range determines which tick labels appear on the graph and thus their size) AND cartesian axis auto-range depends on the margins (via `ax._length` which is used as a scale factor for padded auto-range computations). Unfortunately this can lead to infinite loops (cc https://github.com/plotly/plotly.js/issues/4028). Margins and axis ranges get even more intertwined when [axis constraints are set](https://github.com/plotly/plotly.js/blob/19c0683f24ac26977d7adc51a7ad94c32852f137/src/plots/cartesian/constraints.js#L191-L345).

To illustrate, consider:

```js
Plotly.newPlot('graph', [{
  mode: 'markers',
  marker: {size: 100},
  x: [1, 200, 6000]
}], {
  width: 400, height: 400,
  yaxis: {
    visible: false,
    range: [-1, 3]
  },
  xaxis: {
    tickfont: {size: 32},
    tickangle: 'auto',
    automargin: true,
    autorange: true
  },
  margin: {l: 0, t: 0, r: 0, b: 0}
[...]

PR #2773, opened 2018-07-04
Emit 'plotly_afterplot' on all edits

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2740 and more precisely implements https://github.com/plotly/plotly.js/issues/2740#issuecomment-398198401 where we make `plotly_afterplot` fire after every edit type.

I'm not super proud of commit https://github.com/plotly/plotly.js/commit/cb92d7c31cb6bd27a57e2c88a2404c0d7faa5c31, but please consider this a 'temporary' fix until we improve the automargin pipeline as discussed in #2704

cc @alexcjohnson

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -152,7 +152,6 @@ exports.plot = function(gd, data, layout, config) {
     Plots.supplyDefaults(gd);
 
     var fullLayout = gd._fullLayout;
-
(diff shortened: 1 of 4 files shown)


alexcjohnson:
Yep, the real ugliness lies in our loopy automargin pipeline, this is fine. Nicely done! 💃

PR #2775, opened 2018-07-04
Fix bandwidth for single-value violins

etpinard:
bug brought up by @dmt0 

cc @alexcjohnson


etpinard:
> but consider making an issue for ^^

done in -> https://github.com/plotly/plotly.js/issues/2779

Issue #2748, opened 2018-06-19
Can get an x-only zoombox even when xaxis is fixedrange


alexcjohnson:
Only the y-only zoombox should show up in this case, but if the y range is too small it'll show the x-only zoombox (then do nothing on mouseup)
![kapture 2018-06-19 at 16 20 42](https://user-images.githubusercontent.com/2678795/41622274-15d3dcea-73dd-11e8-986b-40be76a0df37.gif)
(this is mock #12 with `Plotly.relayout(gd,'xaxis.fixedrange',true)`)

PR #2776, opened 2018-07-04
Fix x-only zoom moves when `xaxis.fixedrange: true`

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2748

Now zooming on `12.json` with `Plotly.relayout(gd,'xaxis.fixedrange',true)` gives

peek 2018-07-04 18-00

cc @alexcjohnson


alexcjohnson:
🎉 💃

Issue #2777, opened 2018-07-04
violin: hiding a box within violin produces a black box


dmt0:
To reproduce:
https://codepen.io/dmt0/pen/ZRNpEQ
1. newPlot @ step 0
2. react @ step 1

etpinard:
I'll try to squeeze that one in tomorrow am before 1.39.0

PR #2785, opened 2018-07-05
Fix box & violin inner parts removal

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2777 and a bunch of other restyle/react bugs introduced in https://github.com/plotly/plotly.js/pull/2579, where the box and violin meanlines and points d3-data-binds assume a clean parent `<g>` and didn't care about clearing parts that should no longer be visible.

cc @alexcjohnson @dmt0

### DIFF
--- a/src/traces/box/plot.js
+++ b/src/traces/box/plot.js
@@ -102,6 +102,7 @@ function plotBoxAndWhiskers(sel, axes, trace, t) {
 
     var fn;
     if(trace.type === 'box' ||
+        trace.type === 'candlestick' ||
         (trace.type === 'violin' && (trace.box || {}).visible)
     ) {
         fn = Lib.identity;
--- a/src/traces/box/plot.js
+++ b/src/traces/box/plot.js
@@ -292,6 +292,9 @@ function plotBoxMean(sel, axes, trace, t) {
     var bPos = t.bPos;
     var bPosPxOffset = t.bPosPxOffset || 0;
 
+    // to support violin mean lines
+    var mode = trace.boxmean || (trace.meanline || {}).visible;
+
     // to support for one-sided box
     var bdPos0;
     var bdPos1;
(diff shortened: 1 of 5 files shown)


src/traces/box/plot.js
```diff
@@ -109,7 +100,14 @@ function plotBoxAndWhiskers(sel, axes, trace, t) {
...
 
-    var paths = sel.selectAll('path.box').data(Lib.identity);
+    var fn;
+    if(trace.type === 'box' ||
+        (trace.type === 'violin' && (trace.box || {}).visible)
+    ) {
+        fn = Lib.identity;
+    }
+
+    var paths = sel.selectAll('path.box').data(fn || []);
```
alexcjohnson: I like this pattern! But two comments about it:
- `trace.box` and `trace.meanline` are [deleted](https://github.com/plotly/plotly.js/blob/c4453da2aeca42e925e67086c464dbb0b285051e/src/traces/violin/defaults.js#L49) rather than left with `visible: false` - so can't we just say `(trace.type === 'violin' && trace.box)` etc?
- I'm a bit uneasy relying on uninitialized vars being `undefined` - seems like it makes things more fragile, for example if we use `fn` again, or wrap this block in a loop or something... I'd rather a pattern like
```js
var fn;
if(...) fn = Lib.identity;
else fn = []; // or just var fn = []; and drop the else, dunno which is better

var paths = sel.selectAll('path.box').data(fn);
```

src/traces/box/plot.js
```diff
@@ -102,6 +102,7 @@ function plotBoxAndWhiskers(sel, axes, trace, t) {
 
     var fn;
     if(trace.type === 'box' ||
+        trace.type === 'candlestick' ||
```
alexcjohnson: Perhaps `trace.type !== 'violin' || trace.box`? NBD either way, we'd of course figure it out pretty quick if we add any other trace type that reuses box drawing.

alexcjohnson:
Very nicely done - getting us that much closer to regular d3 idioms - and great tests! Both of my comments are nonblocking - 💃

PR #2788, opened 2018-07-06
Better axis range animations

etpinard:
Axis range animations like https://codepen.io/plotly/pen/GjkZNk (which is featured on https://plot.ly/javascript/animations/) are broken on master ever since https://github.com/plotly/plotly.js/pull/2579 (the cartesian d3-iomatic push, again):

peek 2018-07-06 17-20

This PR makes that example run smoothly again:

peek 2018-07-06 17-22

**but** what's more exciting, this fix -- along with commit https://github.com/plotly/plotly.js/commit/c70819066392b95da67a4bbdd943694beffb6860 -- allows us to smoothly animate axis ranges on cartesian subplots with traces other than `scatter`. For example a bar chart:

peek 2018-07-06 17-23

### DIFF
--- a/test/jasmine/tests/animate_test.js
+++ b/test/jasmine/tests/animate_test.js
@@ -3,6 +3,7 @@ var Lib = require('@src/lib');
 var Registry = require('@src/registry');
 var Plots = Plotly.Plots;
 
+var d3 = require('d3');
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
 var failTest = require('../assets/fail_test');
@@ -862,6 +863,11 @@ describe('animating scatter traces', function() {
             });
         })
         .then(function() {
+            // sanity-check that scatter points and bars are still there
+            var gd3 = d3.select(gd);
+            expect(gd3.select('.scatterlayer').selectAll('.point').size()).toBe(3, '# of pts on graph');
+            expect(gd3.select('.barlayer').selectAll('.point').size()).toBe(3, '# of bars on graph');
+
             // the only redraw should occur during Cartesian.transitionAxes,
             // where Registry.call('relayout') is called leading to a _module.plot call
             expect(gd._fullData[0]._module.basePlotModule.transitionAxes).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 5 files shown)


etpinard:
cc @alexcjohnson

test/jasmine/tests/animate_test.js
```diff
@@ -836,4 +837,43 @@ describe('animating scatter traces', function() {
...
+            // the only redraw should occur during Cartesian.transitionAxes,
+            // where Registry.call('relayout') is called leading to a _module.plot call
+            expect(gd._fullData[0]._module.basePlotModule.transitionAxes).toHaveBeenCalledTimes(1);
+            expect(gd._fullData[0]._module.plot).toHaveBeenCalledTimes(1);
+            expect(gd._fullData[1]._module.plot).toHaveBeenCalledTimes(1);
+            expect(Registry.call).toHaveBeenCalledTimes(1);
+
+            var calls = Registry.call.calls.all();
+            expect(calls.length).toBe(1, 'just one Registry.call call');
+            expect(calls[0].args[0]).toBe('relayout', 'called Registry.call with');
```
alexcjohnson: This is a pretty low-level assertion, not clear to me that it would catch breaks other than the exact change you made here. Would it be possible to do something like:

```js
    var animatePromise = Plotly.animate(...)
    expect(stuff).toBe(nearWhereItStarted)
    return animatePromise;
})
.then(function() {
    expect(stuff).toBe(nearWhereItFinishes);
    expect(theFullDataAndRegistryCallStuff);
```
ie something about the scatter marker and bar positions being updated async that would have failed before this change?

alexcjohnson:
Awesome find, that we get a whole new class of animations with this little change!!! 🎉 🏆

test/jasmine/tests/animate_test.js
etpinard: > ie something about the scatter marker and bar positions being updated async that would have failed before this change? 

The marker and bar position does change, it's the subplot and clip rect translate transform that change during axis range transitions via `Cartesian.transitionAxes` (which mimics the cartesian dragbox logic).

Better test added in -> 19300f9

alexcjohnson:
💃

Issue #2787, opened 2018-07-06
Mapbox subplots are broken in minified bundles


etpinard:
Starting in 1.39.0, 

- https://codepen.io/etpinard/pen/jKgNPW using https://cdn.plot.ly/plotly-1.39.0.js works fine
- https://codepen.io/etpinard/pen/LrwPVm using https://cdn.plot.ly/plotly-1.39.0.min.js fails to render the tiles. 

This one will be fun to debug :expressionless:

etpinard:
Minimal example: https://gist.github.com/etpinard/9dc32b5c8d934c6c3a482a24c2d089cc

etpinard:
Looks like bumping `minify-stream` from 1.1.0 to 1.2.0 in https://github.com/plotly/plotly.js/pull/2656/ did this.

etpinard:
:arrow_heading_up: , bumping minify-stream isn't the whole story. minify-stream@1.1.0 breaks for mapbox-gl@0.45.0+ (but worked fine with 0.44.x which we were using before 1.39.0) :rage:

etpinard:
https://github.com/plotly/plotly.js/pull/2789 fixes this by disabling the `compress` step during minification.

This solution is far from perfect, so I'll leave this issue open until/if we figure a better solution.

etpinard:
https://codepen.io/etpinard/pen/wxwPXJ is working fine with https://cdn.plot.ly/plotly-1.39.1.min.js

PR #2789, opened 2018-07-06
Set minify option 'compress' to false to fix mapbox-gl minification

etpinard:
a very unfortunate fix to https://github.com/plotly/plotly.js/issues/2787, we have a bunch of additional bytes, but at-least mapbox-gl is working.

image

cc @alexcjohnson 

Does anyone know a way to skip uglify-js (-es, terser or other uglify-js folks) the `compress` step on only selected files? cc @dy


alexcjohnson:
yes, I suppose we'd better merge and release this, then find a better solution later. Annoying but 💃

PR #2792, opened 2018-07-09
Better mapbox minification fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2787, though I wouldn't call it a great fix, but it's probably the best we can do in this repo.

By playing around with [terser's compression options](https://github.com/fabiosantoscode/terser#compress-options) (which is the default minifier in our dev-dep [minify-stream](https://github.com/goto-bus-stop/minify-stream)), I found that setting `typeofs: false` on bundles that include mapbox-gl fixes https://github.com/plotly/plotly.js/pull/2789. Moreover, I added a test to make sure this doesn't happen again.

With this PR our bundle sizes are back to where they were before https://github.com/plotly/plotly.js/pull/2789.

### DIFF
--- a/tasks/util/browserify_wrapper.js
+++ b/tasks/util/browserify_wrapper.js
@@ -4,7 +4,6 @@ var path = require('path');
 var browserify = require('browserify');
 var minify = require('minify-stream');
 
-var constants = require('./constants');
 var compressAttributes = require('./compress_attributes');
 var strictD3 = require('./strict_d3');
 
@@ -45,7 +44,7 @@ module.exports = function _bundle(pathToIndex, pathToBundle, opts, cb) {
     }
 
     var b = browserify(pathToIndex, browserifyOpts);
-    var pending = opts.pathToMinBundle ? 2 : 1;
+    var pending = pathToMinBundle ? 2 : 1;
 
     function done() {
         if(cb && --pending === 0) cb(null);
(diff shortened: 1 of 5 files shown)


etpinard:
cc @alexcjohnson

alexcjohnson:
Good find! While we’re at it, would it be worth including all trace types in that test? The `plot_types` mock is a good start, but also may be nice to include parcoords, sankey, table, scattergl...

Also, is it possible to base this flag on whether mapbox is in the bundle definition, rather than naming specific bundles? I wouldn’t want to have to remember this if we add another that includes mapbox. Or should we just leave this option in all bundles? Does it add a noticeable number of bytes? Anyway it says IE10 and below might not like this compression so it may be best to just make it consistent.

etpinard:
> The plot_types mock is a good start, but also may be nice to include parcoords, sankey, table, scattergl...

I hope you like https://github.com/plotly/plotly.js/pull/2792/commits/dd8bf6caa9d0b49a029fb6fa191db8b150712fd3 and https://github.com/plotly/plotly.js/pull/2792/commits/89d7aa835f40508a4fbfd9da2383c0436ec0f761 where I reused the `Plotly.react` noop test lists.

 

> Or should we just leave this option in all bundles? Does it add a noticeable number of bytes?

That's what I went for in the end in https://github.com/plotly/plotly.js/pull/2792/commits/2a3afe7ebd607ad01e050890ef7d9e2e0a907aa7. I find the maintainability gains very nice here. It does adds a few bytes in some but not all bundles. For reference, in red is with `typeofs` compression sizes, in green is w/o:

image

alexcjohnson:
😍 love it! Funny that in at least one case (Cartesian) minified in bigger bit gzipped is *smaller* this way!?
:dancer:

.circleci/test.sh
```diff
@@ -29,13 +29,13 @@ case $1 in
 
     jasmine)
         npm run test-jasmine -- --skip-tags=gl,noCI,flaky || EXIT_STATE=$?
+        npm run test-bundle || EXIT_STATE=$?
```
etpinard: which gives:

image

alexcjohnson: ⚖️ 👍 
💃 still applies :)

Issue #2791, opened 2018-07-09
Redraw errors after Plotly.update in a scattergl with subplots 


vincex:
https://codepen.io/anon/pen/wxwpgm

In this pen after an amount of time a plotly.update is executed and the visibility for trace 2 is set to false.
The update is executed without clear the gl context.

This issue seems realted to a check inside plot_api.js at line 212:

```js
function drawFramework() {
...
...
    if(!fullLayout._glcanvas && fullLayout._has('gl')) {
            fullLayout._glcanvas = fullLayout._glcontainer.selectAll('.gl-canvas').data([{
```
the data for each gl-canvas is recreated in any case even if there is already a data with a regl reference.
This leads to the clearGlCanvases that does nothing if regl is not defined.

```js
module.exports = function clearGlCanvases(gd) {
    var fullLayout = gd._fullLayout;

    if(fullLayout._glcanvas && fullLayout._glcanvas.size()) {
        fullLayout._glcanvas.each(function(d) {
            if(d.regl) d.regl.clear({color: true, depth: true});
        });
    }
};
```

I hope this helps.

etpinard:
A reference must get lost when calling `Plotly.update` with the same layout object as the first `newPlot`.

Calling `Plotly.update('myDiv', {visible: [false]}, {}, [0])` works just fine: https://codepen.io/etpinard/pen/QBLVBz?editors=1010 or its `react` variant 

```js
  data[0].visible = false;
  Plotly.react('myDiv', data, layout); 
```

as in https://codepen.io/etpinard/pen/GBKXYQ?editors=1010 works fine too.

Thanks for the report.

PR #2793, opened 2018-07-09
Do not clear _glcanvas ref during layout replot edits

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2791

Some edits may call `makePlotFramework` in plot_api.js, we can't clear the ref to glcanvas there as we need to properly clear the canvases.

cc @alexcjohnson @dy


dy:
LGTM! :dancer:

PR #2801, opened 2018-07-10
Use updated fullLayout._size in scene.clear

### LINKED ISSUE #2731: selection on scattergl after relayout
I'm testing @jmmease's plotly.py 3.0.0 https://github.com/plotly/plotly.py/pull/942 and seeing an issue with selection on `scattergl` with 1M points. This doesn't happen on the default graph, it only appears after I have modified the graph's width, height, and margin using `Plotly.relayout`.

scatergl-select-box-2

scatergl-select-box

This is a pretty uncommon workflow (calling relayout after plot draw) so it's not a high priority issue at all for plotly.py.

etpinard:
fixes #2731 

Such that after relayout edits that do not trigger a recalc, selection can properly clear the canvas.

cc @alexcjohnson @dy

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -188,7 +188,6 @@ function sceneOptions(gd, subplot, trace, positions, x, y) {
 // make sure scene exists on subplot, return it
 function sceneUpdate(gd, subplot) {
     var scene = subplot._scene;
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Nice find, beautiful test! :dancer:

Issue #2786, opened 2018-07-06
Deleting a transform in lines plot creates erros


dmt0:
https://codepen.io/dmt0/pen/xzoBVN
newPlot @ step0; react @ step1; hover a mouse over plot, see console

```
Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.r.makeEventData (plotly-1.39.0.min.js:7)
    at plotly-1.39.0.min.js:7
    at plotly-1.39.0.min.js:7
    at l (plotly-1.39.0.min.js:7)
    at Object.r.throttle (plotly-1.39.0.min.js:7)
    at Object.r.hover (plotly-1.39.0.min.js:7)
    at SVGRectElement.u.onmousemove (plotly-1.39.0.min.js:7)```

**Comment by dmt0** [comment]
OK cool, so I have a workaround for react-chart-editor

**Comment by bhogan-mitre** [comment]
@dmt0 for what it's worth, I'm still seeing this from react-chart-editor v0.18.2 after deleting a transform.

```
core.js:1449 ERROR TypeError: Cannot read property 'length' of undefined
    at Object.makeEventData (plotly.js:83549)
    at _hover (plotly.js:84245)
    at plotly.js:83734
    at exec (plotly.js:101160)
    at Object.throttle (plotly.js:101169)
    at Object.hover (plotly.js:83731)
    at SVGRectElement.maindrag.onmousemove [as __zone_symbol__ON_PROPERTYmousemove] (plotly.js:113860)
    at SVGRectElement.wrapFn (zone.js:1188)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:421)
    at Object.onInvokeTask (core.js:4751)
```

The traces appear fine from GraphCreatePanel, but the error is still thrown on hover.

PR #2805, opened 2018-07-11
Transform removal from react fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2786

cc @alexcjohnson @dmt0


alexcjohnson:
🙄 cc #749 
💃

Issue #2755, opened 2018-06-26
Contour plot labels issue - Max Contour/Z data value limited to 9772


Arkangus:
Hello,

I have been trying to do some contour plot with some datasets. However, in some case the labels included 'undefined'.

![plot 3](https://user-images.githubusercontent.com/40596006/41903898-5bded604-7937-11e8-980b-f546a35e8a4a.png)

Using the Chart Studio environment with a MWE ([https://plot.ly/~Arkange/3/](https://plot.ly/~Arkange/3/)),  I isolated the issue to the Max Contour value.
With one unique dataset, when Max Contour is set to 9772 or below, the right labels appear. When it is set to 9773 or above, 'undefined' appears and the values are not correct.

**### With Max Contour = 9773:**

![issue1](https://user-images.githubusercontent.com/40596006/41904499-db82ed4a-7938-11e8-9309-9ffa8a1b459b.png)

**### With Max Contour = 9772:**

![issue2](https://user-images.githubusercontent.com/40596006/41904504-ddfe6fc2-7938-11e8-84c1-23866c249f39.png)

From the small tests I have made:
- I tried with values ranging from 3 to 6, or from 3000 to 7000, or from 3000 to 11000. As long as some z values are above 9773 **and** not restricted by Max Contour, the issue appears
- it does not depend on the Min Contour value or Step Size.

I have not systematically explored negative values or changed x- and y-axis datasets.

Arkangus:
This is actually true in absolute value.
Using another data set where the contours start value was -9.8E3, the issue appeared. It dissapeared when I set the Min Contour value to -9772.

etpinard:
Thanks for the report!

PR #2806, opened 2018-07-11
Fix contour labels that required "thousands" suffixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2755, by setting `exponentformat` to 'B' in contour label mock axis object- to prevent faulty fallback in `Axes.prepTicks` and `Axis.tickText`.

cc @alexcjohnson


alexcjohnson:
Nice, I was hoping that one would be easy. Curious though, how did [contour_edge_cases](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/contour_edge_cases.png) not run into this problem?

At some point we should make this `exponentformat` configurable, but that's a `feature` rather than a `bug`.

💃

etpinard:
Those `undefined` text pieces reported in #2755 came from this block:

https://github.com/plotly/plotly.js/blob/194ba1dbb4990c35fbc8979367f01680eb2ad7df/src/plots/cartesian/axes.js#L1338-L1340

where `exponent := 4` (4 for thousands) which isn't covered in `contour_edge_cases` mock.

Issue #2809, opened 2018-07-12
Error when zooming out in scattergeo plot


collioud:
Hello,

I have a scattergeo plot using the Aitoff projection. When positionning the mouse outside the projection area (but within the plot), then zooming out with the mousewheel, an error occur (see Firefox console stack below):

```
TypeError: point is undefined   plotly-latest.js:22222:7
    projection  https://cdn.plot.ly/plotly-latest.js:22222:7
    outside  https://cdn.plot.ly/plotly-latest.js:119841:18
    handleZoom  https://cdn.plot.ly/plotly-latest.js:119859:12
    event  https://cdn.plot.ly/plotly-latest.js:18407:40
    d3_eventDispatch/dispatch.of/<   https://cdn.plot.ly/plotly-latest.js:18443:11
    zoomed  https://cdn.plot.ly/plotly-latest.js:19365:7
    mousewheeled  https://cdn.plot.ly/plotly-latest.js:19472:7
    d3_selection_onListener/<  https://cdn.plot.ly/plotly-latest.js:19023:9
```

Is this the expected behaviour (tell that you are zooming outside the projection scope)?

See this codepen for example: 
[Codepen](https://codepen.io/anon/pen/xJGLMo)

BTW, is there a way to limit the zoom range in scattergeo plots? Maybe a zoom level available somewhere?

Thank you!
Cheers

etpinard:
Thanks for the report!

PR #2811, opened 2018-07-12
Guard against undefined projection.invert results 

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2809

@alexcjohnson some d3.geo projections return `undefined` when trying to map outside-subplot x/y px values to lon/lat.


alexcjohnson:
:dancer: very nice 👍

Issue #2794, opened 2018-07-09
Line animations fail in some frames for some lines.


iandanforth:
OS: OSX 10.13.2
Browsers: Chrome/Firefox/Safari latest

![plotly-bug](https://user-images.githubusercontent.com/446062/42472851-a3d0dd96-8377-11e8-83a1-a75a576b7357.gif)

As you can see in this gif, most of the frames animate properly. Some however have the lines "jumping" to their final values rather than being smoothly transitioned.

This animated scatter is being created from the Python API, but I suspect this is more appropriate for the js side to investigate. (Correct me if I'm wrong).

Here is the script to recreate the above chart. 

**Notes**
 - The issue seems to be largely independent of how many lines are being drawn. (You can change line_count below to see that).
 - The issue goes away if the values remain constant within each line. Comment out line 43 to see that.

```python
import numpy as np
import colorlover as cl
from plotly.offline import plot

sim_duration = 20.0
time_inc = 0.1
line_count = 120
times = np.arange(0.0, sim_duration, time_inc)

# Setting colors for plot.
potvin_scheme = [
    'rgb(115, 0, 0)',
    'rgb(252, 33, 23)',
    'rgb(230, 185, 43)',
    'rgb(107, 211, 100)',
    'rgb(52, 211, 240)',
    'rgb(36, 81, 252)',
    'rgb(0, 6, 130)'
]
# It's hacky but also sorta cool.
c = cl.to_rgb(cl.interp(potvin_scheme, line_count))
c = [val.replace('rgb', 'rgba') for val in c]
c = [val.replace(')', ',{})') for val in c]


def get_color(trace_index: int) -> str:
    # The first and every 20th trace should be full opacity
    alpha = 0.2
    if trace_index == 0 or ((trace_index + 1) % 20 == 0):
        alpha = 1.0
    color = c[trace_index].format(alpha)
    return color

# Per Motor Unit Force

start = np.ones((line_count, len(times)))
inds = np.reshape(np.arange(1.0, 41.0, 40 / len(times)), (1, len(times)))
vals = (-np.log(inds) + 4) / 0.18
vals = np.repeat(vals, line_count, axis=0)


all_array = np.ones((line_count, len(times))) * np.reshape(np.arange(0.0, 35.0, 35 / line_count), (line_count, 1))
all_array += vals  # <---- COMMENT OUT TO REMOVE CURVE
data = []
annotations = []
anno_offsets = {
    0: 20,
    19: 30,
    39: 40,
[...]

PR #2814, opened 2018-07-13
Do not collapse duplicate end points when `line.simplify: false`

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2794 by allowing smooth transitions of lines with the same number of coordinates:

peek 2018-07-13 10-06

cc @alexcjohnson


etpinard:
Minimal example for reference: https://codepen.io/etpinard/pen/djoQXJ?editors=1010

alexcjohnson:
It's a good sign when you can fix a bug in negative LOC :) 💃

Issue #2767, opened 2018-06-29
scattergl select + pan bug


etpinard:
Go to https://codepen.io/etpinard/pen/eKbLgj?editors=1010 for reproducible example and the steps to reproduce it.

![peek 2018-06-29 14-50](https://user-images.githubusercontent.com/6675409/42109512-dee2e718-7bab-11e8-8d24-1516148a51a8.gif)

cc @dy

PR #2815, opened 2018-07-13
Fix scattergl select -> dblclick -> pan bug

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2767 

Can't stop won't stop fixing bugs this week.

peek 2018-07-13 16-09

I'm not 100% this is the best to fix this bug (@dy would you mind reviewing this?), but the new test case should be a good starting point if ever there's a better solution out there.

cc @alexcjohnson

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -616,6 +616,12 @@ function plot(gd, subplot, cdata) {
                 }
             });
         }
+    } else {
+        if(scene.scatter2d) {
+            // reset scatter2d opts to base opts,
+            // thus unsetting markerUnselectedOptions from selection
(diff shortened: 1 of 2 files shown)


alexcjohnson:
So the bug isn't manifest in the figure JSON at all, just on screen? The test looks great, but I agree, @dy should review the actual code change.

etpinard:
> So the bug isn't manifest in the figure JSON at all, just on screen?

correct.

etpinard:
Ping @dy !

Would be nice to include this in 1.39.3

alexcjohnson:
💃

dy:
Lgtm :dancer:!

Issue #2546, opened 2018-04-11
Faster axis autorange relayout


etpinard:
A follow-up on https://github.com/plotly/plotly.js/pull/2527 where I first noticed:

> double-click (i.e auto/reset range) performance. This thing currently goes through editType: 'calc' which is slow as a turtle. I'll try to cache the autorange value somewhere (and clearing it when new data comes in) and use the axrange pathway to speed thing up. This will benefit big data gl and svg graphs too!

and @alexcjohnson replied:

> Axes.expand (which happens during calc does nothing if the axis isn't autoranged - which can speed up initial drawing but means if you turn on autorange later you don't know whether ax._min/_max are populated. We could either calculate min/max no matter what during calc, or at least stash the arrays & options used to generate them so that when autorange is enabled we can pull it back out and proceed with the calculation without needing to go all the way through calc again.

> There would also be some advantages to keeping separate track of each trace / object that called Axes.expand - ie keep a separate _min/_max for each trace/shape/annotation etc. and only combine them when finally calculating the range. That would avoid calc for visibility changes, for example (you'd just have to know to include or not include that object's _min/_max) and for moving annotations/shapes, and down the road would allow calc per trace.

----

Getting this right will improve double-click axis autorange performance and play an important role in https://github.com/plotly/plotly.js/issues/2358 (faster `extendTraces`) and other autorange-related improvements (e.g. https://github.com/plotly/plotly.js/issues/387)

etpinard:
Swapping `editType: 'calc'` for `editType: 'plot'` in the axis `autorange` declaration went almost too smoothly. Just one test was falling (apart from tests checking for undefined `ax._min` and `ax._max`) :

https://github.com/plotly/plotly.js/blob/0de19be151be79fa55283fc6f478a1b90f45609e/test/jasmine/tests/click_test.js#L754-L789

By this old little test was enough to highlight a pretty important problem with this approach and 'calcIfAutorange' edits - which has the name suggests leads to a `calc` edit pathway if one or more axis has `autorange: true`.

For example, scatter `marker.size` is a 'calcIfAutorange' attribute as it can expand the axis ranges. So consider,

```js
Plotly.newPlot(gd, [{
  y: [1, 2, 1] 
}], {
  xaxis: {range: [0, 2]}
})

// then
Plotly.restyle(gd, 'marker.size', 30)

// then (equivalent to double-click)
Plotly.relayout(gd, 'xaxis.autorange', true)

// then 
[...]

PR #2823, opened 2018-07-18
Faster axis autorange + remove calcIfAutorange edit type

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2546 and brings cartesian axis autorange relayout calls (e.g. on-plot-area double-click or the "autoscale" mode bar button) to perform to a level comparable to axis range relayout calls (e.g zoom in/out).

In brief, cartesian `axis.autorange` is now declared as `editType: 'axrange'` and edit type 'calcIfAutorange' is gone.

As mentioned in https://github.com/plotly/plotly.js/issues/2546#issuecomment-405388288, adapting the update pathway ended up being a little trickier than I'd imagine (especially for annotations, see https://github.com/plotly/plotly.js/commit/a8fb653ddca6dd00407b4362131ba2ceb3e6d511). But I'm pretty please with the results: a lot of old delicate "find-autoranged-axes" logic in restyle, relayout and react is gone (see https://github.com/plotly/plotly.js/commit/138a84f613c7c6185c6c1d1d3978fda4754f76e5)

However, there's one drawback that is important to mention. While this PR speeds up axis autorange relayouts, it does _slow down_:
- first render for graphs with set axis ranges (as `Axes.expand` is no longer skipped), and 
- updates of old 'calcIfAutorange' attributes (that are now declared as `editType: 'calc'`) on graphs with set ranges. The most significant (I think) is scatter `marker.size` where updating its values will now **always** lead to a calc edit pathway. 

It would be nice to open a new edit pathway with a targeted  `calc` step where only the relevant `_module.calc` / `_module.setPositions` or component `calcAutorange` would be called. But for the time being, I much prefer having fast double-click and mode bar interactions than fast `marker.size` restyle calls. I hope you agree. 

This PR should be part of the next minor release 1.40.0, not a patch release in the 1.39.x series.

cc @alexcjohnson

### DIFF
--- a/src/components/annotations/attributes.js
+++ b/src/components/annotations/attributes.js
@@ -19,7 +19,7 @@ module.exports = templatedArray('annotation', {
         valType: 'boolean',
         role: 'info',
         dflt: true,
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Determines whether or not this annotation is visible.'
         ].join(' ')
@@ -28,7 +28,7 @@ module.exports = templatedArray('annotation', {
     text: {
         valType: 'string',
         role: 'info',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets the text associated with this annotation.',
             'Plotly uses a subset of HTML tags to do things like',
@@ -41,14 +41,14 @@ module.exports = templatedArray('annotation', {
         valType: 'angle',
         dflt: 0,
         role: 'style',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets the angle at which the `text` is drawn',
             'with respect to the horizontal.'
         ].join(' ')
     },
     font: fontAttrs({
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         colorEditType: 'arraydraw',
         description: 'Sets the annotation text font.'
     }),
@@ -57,7 +57,7 @@ module.exports = templatedArray('annotation', {
         min: 1,
         dflt: null,
         role: 'style',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets an explicit width for the text box. null (default) lets the',
             'text set the box width. Wider text will be clipped.',
@@ -69,7 +69,7 @@ module.exports = templatedArray('annotation', {
         min: 1,
         dflt: null,
         role: 'style',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets an explicit height for the text box. null (default) lets the',
             'text set the box height. Taller text will be clipped.'
@@ -130,7 +130,7 @@ module.exports = templatedArray('annotation', {
         min: 0,
         dflt: 1,
         role: 'style',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets the padding (in px) between the `text`',
             'and the enclosing border.'
@@ -141,7 +141,7 @@ module.exports = templatedArray('annotation', {
         min: 0,
         dflt: 1,
         role: 'style',
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
         description: [
             'Sets the width (in px) of the border enclosing',
             'the annotation `text`.'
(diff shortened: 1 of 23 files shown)


test/jasmine/tests/finance_test.js
```diff
@@ -931,6 +931,58 @@ describe('finance charts updates:', function() {
...
+                tickLen: 0.3,
+                pathd: 'M-20.25,134.55H0M0,81V193.5M20.25,85.5H0'
+            });
+            return Plotly.restyle(gd, 'tickwidth', 0);
+        })
+        .then(function() {
+            _assert('set rng / no tickwidth', {
+                tickLen: 0,
+                pathd: 'M0,134.55H0M0,81V193.5M0,85.5H0'
+            });
```
etpinard: This asserting here fails on `master`, so count this as a bug fix.

src/plot_api/plot_api.js
```diff
@@ -1708,6 +1708,7 @@ function addAxRangeSequence(seq, rangesAltered) {
         subroutines.doTicksRelayout;
 
     seq.push(
+        subroutines.doAutoRangeAndConstraints,
```
etpinard: Adding `doAutoRangeAndConstraints` to the 'axrange' update sequence does work against some of the work done in https://github.com/plotly/plotly.js/pull/2628, but only by 1ms in the most extremes case (a 50x50 splom) when there are no axis constraints.

test/jasmine/tests/axes_test.js
```diff
@@ -1858,24 +1858,21 @@ describe('Test axes', function() {
             expect(ax._max).toEqual([{val: 6, pad: 10, extrapad: true}]);
         });
 
-        it('should return early if no data is given', function() {
+        it('should fail if no data is given', function() {
```
alexcjohnson: you're confident we always send an array to `expand`? I guess in principle if we don't have a required array the trace will be set `visible: false`...

src/traces/scatter/calc.js
```diff
@@ -96,8 +96,9 @@ function calcAxisExpansion(gd, trace, xa, ya, x, y, ppad) {
         yOptions.padded = false;
     }
 
-    Axes.expand(xa, x, xOptions);
-    Axes.expand(ya, y, yOptions);
+    // N.B. asymmetric splom traces call this with undefined xa or ya
```
alexcjohnson: not `undefined`, that would error here as well as above... it's [`{}`](https://github.com/plotly/plotly.js/blob/24b291285a15650a76ae6c78e2762a8e88f2027d/src/traces/splom/index.js#L66-L67)

src/components/annotations/attributes.js
```diff
@@ -19,7 +19,7 @@ module.exports = templatedArray('annotation', {
         valType: 'boolean',
         role: 'info',
         dflt: true,
-        editType: 'calcIfAutorange+arraydraw',
+        editType: 'calc+arraydraw',
```
alexcjohnson: Not for now of course, but annotations (showing, hiding, dragging in GUI) are a great argument in favor of each trace and item tracking its own autorange contributions, so it doesn't take a full recalc to move one annotation.

test/jasmine/tests/annotations_test.js
```diff
@@ -737,6 +737,43 @@ describe('annotations autorange', function() {
...
+        })
+        .then(function() {
+            _assert('auto rng / big tx', [-0.22, 3.57], [0.84, 3.365]);
+            return Plotly.relayout(gd, 'annotations[0].text', 'a');
+        })
+        .then(function() {
+            _assert('auto rng / small tx', [-0.18, 3.035], [0.84, 3.365]);
+        })
+        .catch(failTest)
+        .then(done);
```
alexcjohnson: Beautiful test!

src/components/annotations/draw.js
```diff
@@ -250,11 +250,6 @@ function drawRaw(gd, options, index, subplotId, xa, ya) {
         var outerWidth = Math.round(annWidth + 2 * borderfull);
         var outerHeight = Math.round(annHeight + 2 * borderfull);
 
-
-        // save size in the annotation object for use by autoscale
-        options._w = annWidth;
-        options._h = annHeight;
```
alexcjohnson: Good catch, great to see all the pruning you manage to do in this PR!

alexcjohnson:
@etpinard this is great, I have no qualms about this tradeoff and anyway we have a route to improve it even more later on with per-trace updates. A couple of minor comments but nothing blocking. 💃

test/jasmine/tests/axes_test.js
etpinard: Yes, `expand` is only called from trace module `calc` or `setPosition` or component `calcAutorange` which are skipped for `visible: false` traces / component items.

I thought removing that `if(!data) return;` was a nice way to `expand` is only called when needed.

src/components/annotations/attributes.js
etpinard: > so it doesn't take a full recalc to move one annotation.

Absolutely! I'll open an issue about this once this PR is merged.

src/traces/scatter/calc.js
etpinard: good catch -> 026cd53

etpinard:
Ok let's start merging things for 1.40.x

Issue #1852, opened 2017-07-04
Persistent and Multiple Clicking and Unclicking Behaviour


chriddyp:
Very closely related to [Persistent, stateful hover, click, and selection properties](https://github.com/plotly/plotly.js/issues/1848) and [Customized Click, Hover, and Selection Styles or Traces](https://github.com/plotly/plotly.js/issues/1847) is the ability for users to select one or more points through clicking.

Under the hood, this might just be a new "select" mode (like "Box Select" or "Lasso Select") that allows users to select and de-select points through clicking. Once clicked, the style of the clicked points would remain persistent.

From a UI perspective, the user would be informed that points are clickable and unclickable through:
- `cursor: pointer`
- customized hover on clickable points like bigger points or an outline or a tooltip
- A persistent or hover-activated "x" for unclicking points.

See this tableau dashboard for an example: https://public.tableau.com/profile/joe3537#!/vizhome/WorldBankIndicator/WorldData

![image](https://user-images.githubusercontent.com/1280389/27839063-9e15303e-60bc-11e7-9c65-e24a04f60834.png)

Related to [Persistent, stateful hover, click, and selection properties](https://github.com/plotly/plotly.js/issues/1848), users should be able to programatically specify which points are selected in order to initialize their graphs with pre-selected data.

***

cc @alexcjohnson @etpinard @monfera @jackparmer @charleyferrari

chriddyp:
A few requests for this from the Dash community (https://community.plot.ly/t/highlight-clicked-marker-in-scattermapbox-graph/8315, https://community.plot.ly/t/is-there-a-way-to-clear-clickdata/8708).

If you are using Dash in your company or organization and can fund the development of this feature, [please reach out](https://plotly.typeform.com/to/rkO85m)

alexcjohnson:
A few notes after discussing this offline with @chriddyp 
- Using click to create or alter selections need not be mutually exclusive with lasso or box select; those are set by `layout.dragmode` because the property you're controlling is "what happens when the user drags on the plot." Here the question is "what happens when the user clicks on the plot" and it could be really nice to, for example, draw a lasso to start your selection but then click individual points to add or remove them from the selection.
[...]

PR #2824, opened 2018-07-18
1852 persistent click

rmoestl:
Though not complete, this pull request will introduce the select-on-click / persistent selection feature discussed in issue #1852.

The feature requires a change of the trace selection interface and how selection state is determined and controlled in general.

Tests (image and Jasmine alike) are primarily caused by trace types not yet migrated to the new selection interface.

### DIFF
--- a/src/components/modebar/manage.js
+++ b/src/components/modebar/manage.js
@@ -185,7 +185,6 @@ function areAllAxesFixed(fullLayout) {
 }
 
 // look for traces that support selection
-// to be updated as we add more selectPoints handlers
 function isSelectable(fullData) {
     var selectable = false;
 
@@ -194,7 +193,7 @@ function isSelectable(fullData) {
 
         var trace = fullData[i];
 
-        if(!trace._module || !trace._module.selectPoints) continue;
+        if(!trace._module || !trace._module.getPointsIn || !trace._module.toggleSelected) continue;
 
         if(Registry.traceIs(trace, 'scatter-like')) {
             if(scatterSubTypes.hasMarkers(trace) || scatterSubTypes.hasText(trace)) {
@@ -205,7 +204,7 @@ function isSelectable(fullData) {
                 selectable = true;
             }
         }
-        // assume that in general if the trace module has selectPoints,
+        // assume that in general if the trace module has getPointsIn and toggleSelected,
         // then it's selectable. Scatter is an exception to this because it must
         // have markers or text, not just be a scatter type.
         else selectable = true;
--- a/src/lib/set_operations.js
+++ b/src/lib/set_operations.js
@@ -12,9 +12,19 @@
 /*
  * Computes the set difference of two arrays.
  *
- * Returns all elements of a that are not in b.
+ * @param {array} a
+ * @param {array} b
+ * @returns all elements of a that are not in b.
+ *      If a is not an array, an empty array is returned.
+ *      If b is not an array, a is returned.
  */
 function difference(a, b) {
+    if(!Array.isArray(a)) {
+        return [];
+    }
+    if(!Array.isArray(b)) {
+        return a;
+    }
     return a.filter(function(e) {
         return b.indexOf(e) < 0;
     });
--- a/src/lib/set_operations.js
+++ b/src/lib/set_operations.js
@@ -1,0 +1,25 @@
+/**
+* Copyright 2012-2018, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+
+/*
+ * Computes the set difference of two arrays.
+ *
+ * Returns all elements of a that are not in b.
+ */
+function difference(a, b) {
+    return a.filter(function(e) {
+        return b.indexOf(e) < 0;
+    });
+}
+
+module.exports = {
+    difference: difference
+};
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -356,10 +356,12 @@ function selectOnClick(gd, numClicks, evt, xAxes, yAxes, outlines) {
             searchInfo = searchTraces[i];
             trace = searchInfo.cd[0].trace;
 
-            // Start new selection if needed
-            if(!retainSelection) {
+            // Clear old selection if needed
+            if(!retainSelection || clearEntireSelection) {
                 searchInfo._module.toggleSelected(searchInfo, false);
                 if(outlines) outlines.remove();
+
+                if(clearEntireSelection) continue;
             }
 
             // Determine clicked points,
@@ -384,23 +386,9 @@ function selectOnClick(gd, numClicks, evt, xAxes, yAxes, outlines) {
             allSelectionItems = allSelectionItems.concat(fillSelectionItem(traceSelection, searchInfo));
         }
 
-        // TODO Use the clearEntireSelection flag now
-        // Hack to achieve regl traces to set selectBatch to null in case no point is selected anymore
-        // TODO check in advance if a click clear the entire selection, because in this
-        // case just call toggleSelected(searchInfo, false) on all traces and be done. The `shouldSelect` above might
-        // become obsolete.
-        if(allSelectionItems.length === 0) {
-            for(i = 0; i < searchTraces.length; i++) {
-                searchTraces[i]._module.toggleSelected(searchTraces[i], false);
-            }
-        }
-
         // Grand selection state update needs to be done once for the entire plot
         eventData = {points: allSelectionItems};
         updateSelectedState(gd, searchTraces, eventData);
-
-        // Remove outlines if no point is selected anymore
-        if(allSelectionItems.length === 0 && outlines) outlines.remove();
     }
 
     function clickedPtsFor(searchInfo, hoverData) {
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -397,10 +397,15 @@ function selectOnClick(gd, numClicks, evt, xAxes, yAxes, outlines) {
 
     function clickedPtsFor(searchInfo, hoverData) {
         var clickedPts = [];
+        var hoverDatum;
 
-        for(var i = 0; i < hoverData.length; i++) {
-            var hoverDatum = hoverData[i];
+        if(hoverData.length > 0) {
+            hoverDatum = hoverData[0];
             if(hoverDatum.fullData._expandedIndex === searchInfo.cd[0].trace._expandedIndex) {
+                // TODO hoverDatum not having a pointNumber but a binNumber seems to be an oddity of histogram only
+                // Not deleting .pointNumber in histogram/event_data.js would simplify code here and in addition
+                // would not break the hover event structure officially
+                // documented at https://plot.ly/javascript/hover-events/
                 if(hoverDatum.pointNumber !== undefined) {
                     clickedPts.push({
                         pointNumber: hoverDatum.pointNumber
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -116,7 +116,8 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
         cd = gd.calcdata[i];
         trace = cd[0].trace;
 
-        if(trace.visible !== true || !trace._module || !trace._module.selectPoints) continue;
+        if(trace.visible !== true || !trace._module ||
+          !trace._module.toggleSelected || !trace._module.getPointsIn) continue;
 
         if(dragOptions.subplot) {
             if(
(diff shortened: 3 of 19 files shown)


rmoestl:
Since this is a PR in progress, here's a quick overview of what is missing:

- Transition not yet migrated trace types to new selection interface
- Introducing the `clickmode` attribute and the corresponding logic in `dragbox.js`
- Emitting the correct `plotly_selected`, `plotly_deselected` etc. events after a click-on-select action
- Writing Jasmine tests ensuring click-to-select works for each supported trace type
- Writing Jasmine tests ensuring that a combination of subsequent polygon and click selects work as expected
- Clean up things (added TODOs in particular) and here and there adjust coding style
- Maybe adapt other bases plot types besides `cartesian` to use the new trace selection interface?!
- Also adapt API documentation of `hovermode` because click-to-select works best with `closest`
- Should a selection be cleared when user clicks on an empty area?
- Read through issue #1852 to ensure we haven't forgotten anything

src/components/modebar/manage.js
```diff
@@ -194,7 +193,7 @@ function isSelectable(fullData) {
 
         var trace = fullData[i];
 
-        if(!trace._module || !trace._module.selectPoints) continue;
+        if(!trace._module || !trace._module.getPointsIn || !trace._module.toggleSelected) continue;
```
etpinard: It might be better to make `Plotly.register` set a boolean `_module.selected` somewhere [here](https://github.com/plotly/plotly.js/blob/24b291285a15650a76ae6c78e2762a8e88f2027d/src/registry.js#L224-L267) and use that boolean downstream (e.g. in `Plots.supplyDefaults`) instead of looking for `getPointsIn` and `toggleSelected`.

src/lib/set_operations.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+function difference(a, b) {
+    if(!Array.isArray(a)) {
+        return [];
+    }
+    if(!Array.isArray(b)) {
+        return a;
+    }
+    return a.filter(function(e) {
+        return b.indexOf(e) < 0;
```
etpinard: Looks like `Lib.difference` is only called on array of number. We could take advantage of that by using a hash table instead of `indexOf` which can be slow for arrays with length > 1e5.

Something like:

```js
var small;
var large;
var hash = {};
var out = [];
var i;

if(a.length < b.length) {
  small = a;
  large = b;
} else {
  small = b;
  large = a;
}

for(i = 0; i < small.length; i++) {
  hash[small[i]] = 1;
}

for(i = 0; i < large.length; i++) {
  var l = large[i];
  if(hash[l]) {
    out.push(l)
  }
}

return out;
```

src/plots/cartesian/select.js
```diff
@@ -349,11 +323,246 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
+
+    for(i = 0; i < gd.calcdata.length; i++) {
+        cd = gd.calcdata[i];
+        trace = cd[0].trace;
+
+        if(trace.visible !== true || !trace._module ||
+          !trace._module.toggleSelected || !trace._module.getPointsIn) continue;
+
+        // TODO is dragOptions.subplot is ever set? If not, delete.
+        // if(dragOptions.subplot) {
```
etpinard: Looks like [ternary subplots still set that _subplot_ field](https://github.com/plotly/plotly.js/blob/24b291285a15650a76ae6c78e2762a8e88f2027d/src/plots/ternary/ternary.js#L458) as well as geo and mapbox subplots.

src/traces/scattergl/index.js
```diff
@@ -590,32 +592,30 @@ function plot(gd, subplot, cdata) {
             null;
     });
 
-    if(selectMode) {
```
etpinard: Ah no. We need to keep that `selectMode` condition.

As creating one regl context takes about ~300ms, we don't want to create the selection regl context when in non-select interaction modes.

But perhaps, you only remove it before implementing `clickmode`. Please ignore if that's the case.

src/traces/scattergl/index.js
```diff
@@ -823,49 +823,91 @@ function calcHover(pointData, x, y, trace) {
...
+    if(hasText) {
+        styleTextSelection(calcData);
     }
 
-    // make sure selectBatch is created
+    // TODO remove eventually
+    // console.log('els  : ' + JSON.stringify(scene.selectBatch[stash.index]));
+    // console.log('unels: ' + JSON.stringify(scene.unselectBatch[stash.index]));
+    // console.log('selection: ' + JSON.stringify(selection));
+    return selection;
```
etpinard: I have some bad news. The new selection algo perform very poorly above 1e5 points (which in practice only affects `scattergl`). From

```js
var N = 1e5
var x = []
var y = []
var ms = []

for(var i = 0; i < N; i++) {
  x.push(Math.random())
  y.push(Math.random())
  ms.push(i % 2 ? 20 : 10)
}

console.time(N)
Plotly.newPlot(gd, [{
  type: 'scattergl',
  mode: 'markers',
  x: x,
  y: y,
  marker: {size: ms}
}], {
  dragmode: 'select',
})
console.timeEnd(N)
```

selecting is painfully slow.

----

I'll try to pin down what's causing this slow down later today.

src/plots/cartesian/select.js
```diff
@@ -308,18 +276,24 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
 
                 updateSelectedState(gd, searchTraces);
+
+                // clear visual boundaries of selection area if displayed
+                outlines.remove();
+
                 gd.emit('plotly_deselect', null);
             }
             else {
+                // TODO What to do with the code below because we now have behavior for a single click
```
etpinard: Thanks for writing down that `TODO`.

I'm thinking when `clickmode: 'select'` we'll emit `plotly_selected` with the selected points info. Otherwise, we'll have to keep that dummy event below.

src/plots/cartesian/select.js
```diff
@@ -308,18 +276,24 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
                 for(i = 0; i < searchTraces.length; i++) {
                     searchInfo = searchTraces[i];
-                    searchInfo._module.selectPoints(searchInfo, false);
+                    searchInfo._module.toggleSelected(searchInfo, false);
                 }
 
                 updateSelectedState(gd, searchTraces);
+
+                // clear visual boundaries of selection area if displayed
+                outlines.remove();
```
etpinard: Why did you move this after `updateSelectedState`?

src/lib/set_operations.js
etpinard: ... it might be nice to look at the lodash's implementation.

src/plots/cartesian/select.js
```diff
@@ -349,11 +323,246 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
+                searchInfo._module.toggleSelected(searchInfo, false);
+                if(outlines) outlines.remove();
+
+                if(clearEntireSelection) continue;
+            }
+
+            // Determine clicked points,
+            // call selection modification functions of the trace's module
+            // and collect the resulting set of selected points
+            clickedPts = clickedPtsFor(searchInfo, hoverData);
```
etpinard: Does the `clickedPts`  array ever have more than one item?

src/lib/set_operations.js
```diff
@@ -9,25 +9,33 @@
...
-    if(!Array.isArray(a)) {
-        return [];
+    if(!Array.isArray(a)) return [];
+    if(!Array.isArray(b)) return a;
+
+    var hash = {};
+    var out = [];
+    var i;
+
+    for(i = 0; i < b.length; i++) {
```
etpinard: This commit makes selection of 1e5 scattergl markers work ok, but still about an order of magnitude too slow. 

I'll have to dig deeper tomorrow.

src/components/modebar/manage.js
rmoestl: Neat! Would `_module.selectable` even be a better name?

src/plots/cartesian/select.js
rmoestl: Thx! Really great you're also going over my TODOs because some of them would have required your or AJ's input anyways!

src/traces/scattergl/index.js
rmoestl: Oh, I'll have to dig into that again. With `if(selectMode)` scattergl traces weren't updated on select-on-click when in zoom/pan dragMode. I guess I'll have to adapt the logic when `selectMode` flag is set to `true`.

src/traces/scattergl/index.js
rmoestl: Okay, thanks! Let me know if I can be of any assistance, e.g. if some of my code isn't clear...

src/plots/cartesian/select.js
rmoestl: Thx.

src/plots/cartesian/select.js
rmoestl: TBH, I'm not sure. Fun fact: I even had it beneath emitting `plotly_selected` event which then broke some unit tests that assume outline elements are no longer in DOM when `plotly_selected` event is received.

But why did I really moved it below `updateStelectedState`? As far as I can remember, in the early commits, I thought I could move the line before the `if(numClicks === 2)` condition. For some reason that wasn't correct and I moved it back into the if block. So I guess no rationale behind it other than me not putting it back where it always should have been.

src/plots/cartesian/select.js
rmoestl: At the moment not. I changed it back when I was about to integrate the `box` trace type. See https://github.com/plotly/plotly.js/pull/2824/commits/231fa6d1933597bebaa271b5434f9f0cf56df8ca.

I figured it'd be a good idea to be ready for `hoverdata` one day containing more than one valid data point. E.g. when multiple data points are at the same location that was clicked.

Is allocating an array costly in terms of performance? Happy to change it back to non-array style if that is the case! Or, if we want to keep it, I could add a comment!

rmoestl:
Thanks @etpinard for taking the time to do such a thorough review! :bowing_man:
I'll gradually work through the feedback.

src/plots/cartesian/select.js
etpinard: > Is allocating an array costly in terms of performance?

It _can_ be costly, but I'll have to test on some real-world data to know for sure.

src/traces/scattergl/index.js
rmoestl: As far as I understand, the selection regl context is created by a call to `createScatter()`. When user switches to lasso or box mode, the plot is re-rendered for some reason and regl contexts are creating again.

I think you are absolutely right to take `clickMode` into account to decide if `scene.select2d` context needs to be created.

For now, I'll reintroduce the `selectMode` condition and add a TODO for later when I introduce `clickmode`.

src/plots/cartesian/select.js
rmoestl: For now I postpone moving back to a non-array return type until there's proof that it hurts performance considerably.

src/lib/set_operations.js
rmoestl: Have you dug deeper since then?

etpinard: I haven't had the time unfortunately.

etpinard: ... I find some time to review it before the meeting on Thursday though.

rmoestl: No problem. There's no hurry at the moment.

etpinard: Ok very briefly just to get a conversion started (and not break my promise):

Doing some order of operation analysis in the hot selection code path comparing:

https://github.com/plotly/plotly.js/blob/def6aa5a24527c68e90f50485255170609640e43/src/plots/cartesian/select.js#L283-L286

on master to this branch's

https://github.com/plotly/plotly.js/blob/bd04fd8cfcf3a12a485b4d7d48dc69b8f9b5a3d9/src/plots/cartesian/select.js#L242-L254

we can see why this branch is roughly an order of magnitude slower:

On master that block scales as O(n). To handle addition/subtraction/merge selections, we use an algo acting on the selections polygon with I believe is O(number-of-vertices^2) but results nonetheless in not that many computations in most use cases (e.g. rectangles with 4 vertices)

On this branch, we have

```js
if(!retain) toggleSelected('clear')
// which is currently O(2*n)

getPolyonTester()
// not sure why this one is called in the hot code path

getPointsIn()
// this one is O(n) assuming 'fast' is-in-polygon computations

toggleSelected('add or substract')
// this one currently is O(n + p)

difference()
// O(n + p)

toggleSelected('clear')
// this one is O(n + (n-p))
```

which when added all up gives a total of O(6n + p) where n is the number of points and p is number of selected points ~ roughly one order of magnitude.

So, we should try to combine all the `toggleSelected` loops into one. Don't get me wrong, I don't think splitting up `selectedPoints` into `toggleSelected` and `getPointsIn` was a bad idea, but perhaps we could bring a version of `selectPoints` that reuses `getPointsIn`? We can talk more about this tomorrow during the meeting. Thanks for your hard work!

rmoestl: In hindsight it seams the new approach of splitting up `selectedPoints` wasn't such a good idea :-( I'm not sure if it's possible to speed things up that much with that approach. To be honest, it feels like to be at square one again.

Thinking back, the reason for splitting up `selectPoints` was to cover a couple of edge cases when users would use click-to-select and polygon-select interchangeably. Back then we discussed possible approaches [here](https://github.com/plotly/plotly.js/issues/1852#issuecomment-399213759).

Let me jot down **some thoughts for the upcoming meeting**:

## Current approach on master
The approach to selections before tackling persistent select was as follows: each trace type module supporting selection exposes a `selectPoints` function that accepts a `searchInfo` parameter describing the actual trace and a `polygon` object that offers a `contains` function. `selectPoints` iterates through the data points of the passed trace (identified by `searchInfo`) and calls `polygon.contains` for each data point to see if the data point is within the selection area. Adding/subtracting from an existing selection is supported by wrapping multiple `polygons` into one multiPolygon and passing that to `selectPoints`.

## Possible new approach to support click-to-select
The whole polygon testing approach is covered in `lib/polygon.js` which exposes `tester` (for testing a single polygon) and `multiTester` (wrapper for multiple `tester` objects). What about having a third `tester` type that tests by data point index? What would that mean for the client code?
- Go back to `selectPoints` in trace type modules
- Add data point index as a parameter to the signature of `contains`
- In `selectPoints` implementations do not only pass the `x` and `y` (or `lan`, `lat` etc.) of a data point when calling `contains` but also the data point index
- Adapt `multiTester` to be able to wrap the new index-based tester

etpinard: Your new approach sound very promising! I'd say give it a shot with `scatter` and `scattergl` and ping so that I can quickly review and perf-test it. :ok_hand:

rmoestl: @etpinard proof-of-concept is ready to review. It was easier to start from scratch. So the new approach can be found at https://github.com/plotly/plotly.js/tree/1852-persistent-click-via-selectPoints. (Although it would make reviewing easier, I figured code is not quite ready for a new PR).

etpinard: I made a few comments on https://github.com/plotly/plotly.js/commit/2fbaaaa5c48e9215a264234ae683142353e17968 . Your solution is looking great! :gem: 

I'd say go ahead and try to implement it for all selectable trace modules. :rocket:

PR #2826, opened 2018-07-19
Changes to make redrawObjs work by convention

wayne-o:
We're looking to extend plotly so that we can embed some HTML components in graphs.

What we have so far works quite well but we are having to use a custom build of plotly to achieve what we want. 

I think that with few small modifications we could actually have plotly much more extensible which would be awesome! 

I may be wrong here and there may well already be a way to achieve what I'm aiming for but this is how I think it could work:

 - Allow images module to expose a `drawOne` method (even if it just proxies to `draw`) - this would make it uniform with the other modules that require `draw`ing
 - possibly allow modules to advertise that they are `draw`-able and will require re`draw`ing on pan / animation / whatever
 - extend `registry` to export a method allowing us to retrieve all registered modules (perhaps using a filter like `isDrawable:true` or whatever - we could check for the existence of drawOne before calling but I'm not sure how wastefull it would be to be passing about all modules and doing the checks when it's not needed
 - In places like `dragbox` instead of calling stuff explicitly like this: `redrawObjs(gd._fullLayout.annotations || [], Registry.getComponentMethod('annotations', 'drawOne'));` - we can now retrieve all modules using the filter and itterate over them calling `drawOne`

### DIFF
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -18,8 +18,8 @@ module.exports = {
     supplyLayoutDefaults: require('./defaults'),
     includeBasePlot: require('../../plots/cartesian/include_components')('images'),
 
-    draw: drawModule.draw,
-    updateOnPan: drawModule.draw,
+    draw: drawModule,
+    updateOnPan: drawModule,
 
     convertCoords: require('./convert_coords')
 };
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -20,7 +20,6 @@ module.exports = {
 
     draw: drawModule.draw,
     updateOnPan: drawModule.draw,
-    updateOnPanShortCircuit: true,
 
     convertCoords: require('./convert_coords')
 };
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -11,12 +11,14 @@
 module.exports = {
     moduleType: 'component',
     name: 'images',
-
+    isDrawable: true,
+    
     layoutAttributes: require('./attributes'),
     supplyLayoutDefaults: require('./defaults'),
     includeBasePlot: require('../../plots/cartesian/include_components')('images'),
 
     draw: require('./draw'),
+    drawOne: require('./draw'),
 
     convertCoords: require('./convert_coords')
 };
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -8,17 +8,19 @@
 
 'use strict';
 
+var drawModule = require('./draw');
+
 module.exports = {
     moduleType: 'component',
     name: 'images',
-    isDrawable: true,
 
     layoutAttributes: require('./attributes'),
     supplyLayoutDefaults: require('./defaults'),
     includeBasePlot: require('../../plots/cartesian/include_components')('images'),
 
-    draw: require('./draw'),
-    drawOne: require('./draw'),
+    draw: drawModule.draw,
+    updateOnPan: drawModule.draw,
+    updateOnPanShortCircuit: true,
 
     convertCoords: require('./convert_coords')
 };
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -12,7 +12,7 @@ module.exports = {
     moduleType: 'component',
     name: 'images',
     isDrawable: true,
-    
+
     layoutAttributes: require('./attributes'),
     supplyLayoutDefaults: require('./defaults'),
     includeBasePlot: require('../../plots/cartesian/include_components')('images'),
--- a/src/components/images/index.js
+++ b/src/components/images/index.js
@@ -20,6 +20,7 @@ module.exports = {
 
     draw: drawModule,
     updateOnPan: drawModule,
+    updateOnPanShortCircuit: true,
 
     convertCoords: require('./convert_coords')
 };
--- a/src/registry.js
+++ b/src/registry.js
@@ -72,20 +72,20 @@ exports.apiMethodRegistry = {};
  *
  */
 exports.register = function register(_modules) {
-    if(!_modules) {
+    if (!_modules) {
         throw new Error('No argument passed to Plotly.register.');
-    } else if(_modules && !Array.isArray(_modules)) {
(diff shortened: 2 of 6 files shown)


wayne-o:
My only issue that needs resolving here is that `gd._fullLayout` doesn't seem to take new nodes for new component types - how can I make it so that this is extensible?

wayne-o:
I have managed to get `gd._fullLayout` to pass my custom config with a _lot_ of jiggery pokery.

So now the only things that are different are the need to re-export these using a custom build:

```
Plotly.registry =  require('./plotly.js/src/registry')
Plotly.axes = require('./plotly.js/src/plots/cartesian/axes');
Plotly.plotTemplate = require('./plotly.js/src/plot_api/plot_template');
Plotly.color = require('./plotly.js/src/components/color');
Plotly.lib = require('./plotly.js/src/lib');
Plotly.lib.setCursor = require('./plotly.js/src/lib/setCursor');
Plotly.lib.svgTextUtils = require('./plotly.js/src/lib/svg_text_utils');
Plotly.drawing = require('./plotly.js/src/components/drawing');
Plotly.d3 = require('d3');
Plotly.plots = require('./plotly.js/src/plots/plots');
Plotly.templatedArray = require('./plotly.js/src/plot_api/plot_template').templatedArray;
Plotly.cartesianConstants =  require('./plotly.js/src/plots/cartesian/constants');
Plotly.handleArrayContainerDefaults = require('./plotly.js/src/plots/array_container_defaults');
```
`registry` is probably not needed using this PR but the rest are needed so that I can work out positioning and allowing some defaults etc

wayne-o:
Annoyingly, I've just realised I still need to do this: `registry.getComponentMethod('customItems', 'draw')(myPlot);`

which means `registry` still needs to be exported - i think the naming of `isDrawable` is wrong but I am unure what groups these modules that need painting / repainting so not sure what the best name would be - IMO if they can be grouped together then they should be as::

```
        var componentsNeedingDraw = Registry.getModules('component', true);
        componentsNeedingDraw.forEach(function(item) {
             Registry.getComponentMethod(item.name, 'draw');
        });
```

As with `reDraw`

etpinard:
Thanks very much for the PR!

I think I understand what you're trying to do in your app, but perhaps sharing an example of your custom component could help us find the best solution going forward.

To my understanding, you're trying to:

1) group all component that need to redraw during panning and axis transitions (we can probably do better than `isDrawable`, but that's a detail)
2) loop over that group during pan and axis transitions (instead of those _hard-coded_ `getComponentMethod` blocks)
3) register a custom module with the appropriate tag and :crossed_fingers: that it _just works_<sup>TM</sup>.

From what I'm seeing there's nothing wrong in trying to generalize component draw calls during panning, but I'm afraid that it will :lock: down the what is currently an internal API. What if we want to optimize those on-pan / on-axis-transition update methods? Note that even `Annotations.drawOne` can be pretty slow when >10 annotations are present on a subplot.

So here's what I proposing: 

- instead of tagging the component methods with `isDrawable`, let's _tag_ them by adding a new method (that I would call `updateOnPan`, but maybe we could do better) that wrap `draw` or `drawOne` accordingly. For example, we'd have:

```js
// in components/annotations/index.js
// and components/shapes/index.js
exports.updateOnPan = exports.drawOne;
```

and we'd have to think of something clever for images, to make sure we can short-circuit the loop as we do currently. By calling `updateOnPan` instead of `drawOne` we would leave ourselves room to optimize that method down the road w/o changing the API.

- as this is a hot code path, we should stash a list of component modules with an `updateOnPan` on `Plotly.registry`
- loop over that list of modules on pan and during axis transitions
- now we feel confident about advertising a way to add custom pannable component in user-land.

--------

I don't get the part about `_fullLayout` 

> My only issue that needs resolving here is that gd._fullLayout doesn't seem to take new nodes for new component types - how can I make it so that this is extensible?

could you expand on that?

wayne-o:
Hi - thanks for responding

 - could you expand on that?

I actually resolved that by including the `defaults` and `attributes` bits :)

wayne-o:
I'll start on the suggestions on  Monday btw

wayne-o:
@etpinard  - is there anything else you need from me here?

src/registry.js
```diff
@@ -108,6 +108,18 @@ exports.register = function register(_modules) {
     }
 };
 
+exports.getUpdateOnPanComponents = function() {
```
etpinard: I was imagining a stashed list (similar to [these](https://github.com/wayne-o/plotly.js/blob/1479e9fd9880717659f8ee571202e5aba01d9626/src/registry.js#L23-L33)), not a getter method. This way we won't have to loop over all the registered components at the fps drag rate.

src/components/images/index.js
```diff
@@ -16,7 +18,9 @@ module.exports = {
     supplyLayoutDefaults: require('./defaults'),
     includeBasePlot: require('../../plots/cartesian/include_components')('images'),
 
-    draw: require('./draw'),
+    draw: drawModule,
+    updateOnPan: drawModule,
+    updateOnPanShortCircuit: true,
```
etpinard: Not a bad solution. Nice! 

Could you just add a comment to tell future devs what this flag is about?

etpinard:
Sorry for the wait @wayne-o, I just made a few comments.

To note, this PR will only be part of the [v1.40.x](https://github.com/plotly/plotly.js/milestone/36) series, which is a little backed up at the moment. Don't expect your work to be released until at least two weeks. Sorry for the inconvenience.

etpinard:
@wayne-o will you have the time to address my comments before next Tuesday? If so, this PR will be part of the 1.40.0 release.

etpinard:

@wayne-o feel free to open a new PR with comments:

- https://github.com/plotly/plotly.js/pull/2826#discussion_r206232414
- https://github.com/plotly/plotly.js/pull/2826#discussion_r206232747

addressed when (or if) you'll have the time to look at this again.

PR #2827, opened 2018-07-19
Scattergeo selection fixups

etpinard:
fixes a few issues with scattergeo selection:

- [un]selected.marker.opacity never got coerced correctly, and no one has reported this before @rmoestl did a few hours ago :expressionless: 
- some geo projections return `[NaN, NaN]` when trying to convert bad lon/lat coordinates into pixels, but some others (e.g. robinson -> https://rreusser.github.io/plotly-mock-viewer/#geo_country-names) throw up on bad lon/lat. From now on, we won't take any chances.

@alexcjohnson


alexcjohnson:
Very nice, :dancer:  Thanks for the report @rmoestl!

Issue #2819, opened 2018-07-17
Only first trace toggled when horizontal legend and legendPosition: true


AlanJMac:
When the legend layout is set to the horizontal orientation and the legendPosition config variable is set to true then clicking the trace names in the legend will only toggle the visibility of the first trace. 
The behaviour is as expected when the legend is vertically orientated or if its position is not editable.

The layout and config to reproduce this with 2 or more traces is as follows: 
````
var layout = {
  legend: {
    orientation: 'h'
  }
};

var config = {
  edits: {
    legendPosition: true
  }
};
````
This can be seen in the following codepen: https://codepen.io/alanjmac/pen/NBNmaQ

PR #2829, opened 2018-07-19
Position-editable horizontal legend toggle fixups

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2819, an edge case that made 

https://github.com/plotly/plotly.js/blob/0426b61b9a6034b73adca10970923d8ced6cc71b/src/components/legend/draw.js#L343-L354

not grab the correct legend item bounding boxes when `edits.legendPosition` was turned on. 

cc @alexcjohnson


alexcjohnson:
Beautiful. Thanks for sorting out that flakiness. 💃

Issue #2828, opened 2018-07-19
Plotly.react secondary axis draw error


nicolaskruchten:
This Pen https://codepen.io/nicolaskruchten/pen/VBmWLN?editors=0010 shows a `Plotly.react` call which should move the secondary x-axis but results in the old x-axis to remain in the plot.

etpinard:
Ok. Thanks for the report

```js
Plotly.newPlot(gd, steps[0]())
Plotly.relayout(gd, {
  'xaxis2.overlaying': 'free',
  'xaxis2.domain': [0.37, 1]
})
```

is enough to reproduce the bug also.

PR #2831, opened 2018-07-20
Fix overlaying subplot configuration relayouts

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2828 and moves a few "clean up old layer" blocks out of `Plots.supplyDefaults` and into `Cartesian.drawFramework` (cc https://github.com/plotly/plotly.js/issues/2549).

cc @alexcjohnson @nicolaskruchten


alexcjohnson:
Lovely. Nice bug fix and a structural step in the right direction to boot! :dancer:

PR #2837, opened 2018-07-23
Faster trace visibility toggling

etpinard:
... by changing `visible` from a 'calc' to a 'plot' edit type and using a "visible batch" in the scattergl code, similar to how selections are currently handled.

Graphs will slow calc steps will benefit the most. For example, on a 1e6 scattergl graph,

```js
setTimeout(() => {
  console.time('1')
  console.timeEnd('1')
}, 500)
```

clocks in at about 30ms which is about 50x faster than on master :racehorse: 

Eventually, we could augment this PR by adding a new edit type that would bypass `drawFramework` and `initInteractions` (which are very :turtle:  for sploms), but I'll leave that for later if ok.

Note that commits https://github.com/plotly/plotly.js/commit/3790734abdf0c6c37f1864cba726ed2493820376 and https://github.com/plotly/plotly.js/commit/b816a8e68b8ddb86faef8b10f14f03ff584123ef might be a bit controversial and may have some side-effects during subplot/trace removal the spli through our tests, so I wouldn't mind a second opinion there.

I would appreciate if @dy could take a look at https://github.com/plotly/plotly.js/commit/e85725aec106b8ce629ec03691f4a9508f8e81ff and https://github.com/plotly/plotly.js/commit/cf65abd64816506309fb4d66689cbf6d1fb2d3aa

--------

To be merged in https://github.com/plotly/plotly.js/pull/2823 

cc @alexcjohnson and @dy

### DIFF
--- a/test/jasmine/tests/splom_test.js
+++ b/test/jasmine/tests/splom_test.js
@@ -579,6 +579,41 @@ describe('@gl Test splom interactions:', function() {
         .catch(failTest)
         .then(done);
     });
+
+    it('should toggle trace correctly', function(done) {
+        var fig = Lib.extendDeep({}, require('@mocks/splom_iris.json'));
+
+        function _assert(msg, exp) {
+            for(var i = 0; i < 3; i++) {
+                var draw = gd.calcdata[i][0].t._scene.draw;
+                expect(draw).toHaveBeenCalledTimes(exp[i], msg + ' - trace ' + i);
+                draw.calls.reset();
+            }
+        }
+
+        Plotly.plot(gd, fig).then(function() {
+            spyOn(gd.calcdata[0][0].t._scene, 'draw');
+            spyOn(gd.calcdata[1][0].t._scene, 'draw');
+            spyOn(gd.calcdata[2][0].t._scene, 'draw');
+
+            return Plotly.restyle(gd, 'visible', 'legendonly', [0, 2]);
+        })
+        .then(function() {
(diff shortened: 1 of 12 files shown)


test/jasmine/tests/splom_test.js
```diff
@@ -579,6 +579,41 @@ describe('@gl Test splom interactions:', function() {
...
+
+        function _assert(msg, exp) {
+            for(var i = 0; i < 3; i++) {
+                var draw = gd.calcdata[i][0].t._scene.draw;
+                expect(draw).toHaveBeenCalledTimes(exp[i], msg + ' - trace ' + i);
+                draw.calls.reset();
+            }
+        }
+
+        Plotly.plot(gd, fig).then(function() {
```
etpinard: This is broken on master, consider this a bug fix.

alexcjohnson:
Love the concept! But it looks like autoranged axes aren't updated with the faster logic. I'll hold off reviewing until that's sorted out as I suspect it's going to be a bit involved - likely by maintaining separate `_min` and `_max` for each trace? We've been talking about doing that for a while anyhow.

autorange on visibility

etpinard:
> But it looks like autoranged axes aren't updated with the faster logic

... and looks like we don't have any tests covering these cases :expressionless:

etpinard:
Abandoned. See https://github.com/plotly/plotly.js/pull/2860

PR #2847, opened 2018-07-26
fix #1363 - honor tenths of milliseconds in old numeric date data

### LINKED ISSUE #1363: Fractional milliseconds in time series are not handled correctly.
Hi, 

When I'm using scatter chart to display time series with fractional milliseconds, the fractional part is not displayed correctly. Multiple samples become one, this looks strange, for sure it's not the user expects to see on high resolution. The problem persists in all versions of plotly.js since 1.21.0. Version 1.20.5 doesn't have this problem.

Example on jsfiddle: 
[1.20.5 (expected behaviour)](https://jsfiddle.net/c1kt3r82/143/)
[1.23.0 (issue is reproduced)](https://jsfiddle.net/c1kt3r82/144/)

alexcjohnson:
@etpinard a quick one while looking over old bug reports

### DIFF
--- a/src/plots/cartesian/set_convert.js
+++ b/src/plots/cartesian/set_convert.js
@@ -92,7 +92,14 @@ module.exports = function setConvert(ax, fullLayout) {
         // as (local) ms if that fails.
         var ms = dateTime2ms(v, calendar || ax.calendar);
         if(ms === BADNUM) {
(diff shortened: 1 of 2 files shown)


etpinard:
Looks good, nice test :dancer:

PR #2849, opened 2018-07-26
Per-trace axis extremes

etpinard:
to be merged in https://github.com/plotly/plotly.js/pull/2837

By trying to address @alexcjohnson 's https://github.com/plotly/plotly.js/pull/2837#issuecomment-407550652, I opened up a pretty big can of worms. There are twice as many commits here than in #2837 :scream: 

Oh well, I think this is for the best. In brief, 

- https://github.com/plotly/plotly.js/commit/67db3331360cd855a5462d1474e8cdfe5f912dee adds a test that fails in #2837
- `Axes.expand` is replaced by `Axes.findExtremes`. To note, we no longer mutate `_min` and `_max` array into the axis to expand, but instead we fill in `_extremes: {x: {min: [], max: [], y: {min: [], max: []}}` objects in the traces and layout components that **can** expand the axis range. That `extremes` object is slightly cumbersome, I'm open to suggestions to make it smoother to work with. 
- `Axes.getAutoRange` now must look for the `_extremes` objects -- skipping the `visible:false` items -- corresponding to a given axis and then computes the auto range
- https://github.com/plotly/plotly.js/commit/4c250c3c90f061ecbe1836f8ef07ed9b561d9f6b is **very important**, `setPositions` was skipped previously during 'plot' edits, we must now call it to make sure bar and box (and OHLC?) can restyle `visible` correctly. I'll need to double-check a for more cases.
- one polar mock is currently failing, but I think it is more correct now. **Update:** the baseline is correct, the old `concatExtremes` didn't pick the right values, fixed in 66df51e8cf9d99e69f3968eb5e73929c55c233a2
- <strike>We should now be able to change annotations and shapes `visible` from an edit type 'calc' to 'plot' :racehorse: **done in** -> https://github.com/plotly/plotly.js/pull/2849/commits/9a7801339f0572659bd702eac3428daa2421d8db</strike> **reverted** in https://github.com/plotly/plotly.js/pull/2849/commits/5cfee13a5bba6f6aa9c505aa3ccb4ff5d68a35ca will be for later.


src/plots/cartesian/autorange.js
```diff
@@ -174,15 +178,40 @@ function makePadFn(ax) {
     return function getPad(pt) { return pt.pad + (pt.extrapad ? extrappad : 0); };
 }
 
-function doAutoRange(ax) {
+function concatExtremes(gd, ax, k) {
```
etpinard: Hmm. Come to think of it, this thing will be a drag. Looping over all traces, annotations and shapes for every axis, doesn't scale very well.

src/plots/plots.js
```diff
@@ -2455,10 +2455,14 @@ plots.doCalcdata = function(gd, traces) {
         }
     }
 
-    // find array attributes in trace
     for(i = 0; i < fullData.length; i++) {
         trace = fullData[i];
+
+        // find array attributes in trace
```
alexcjohnson: 🤔 isn't `PlotSchema.findArrayAttributes(trace)` self-documenting enough?

src/components/annotations/calc_autorange.js
```diff
@@ -34,63 +34,70 @@ function annAutorange(gd) {
...
+    var headSize = 3 * ann.arrowsize * ann.arrowwidth || 0;
+    var startHeadSize = 3 * ann.startarrowsize * ann.arrowwidth || 0;
+
+    var axId = ax._id;
+    var letter = axId.charAt(0);
+    var pos = ann[letter];
+    var apos = ann['a' + letter];
+    var ref = ann[letter + 'ref'];
+    var aref = ann['a' + letter + 'ref'];
+    var shift = ann[letter + 'shift'];
```
alexcjohnson: if you did `* (letter === 'x' ? 1 : -1)` the `switch` below would collapse.

src/plots/cartesian/axes.js
```diff
@@ -50,6 +50,7 @@ axes.getFromTrace = axisIds.getFromTrace;
 var autorange = require('./autorange');
 axes.expand = autorange.expand;
```
alexcjohnson: 🔪

src/plots/cartesian/constraints.js
```diff
@@ -143,15 +144,17 @@ exports.enforce = function enforceAxisConstraints(gd) {
                         var newVal;
                         var k;
 
-                        for(k = 0; k < ax._min.length; k++) {
-                            newVal = ax._min[k].val - getPad(ax._min[k]) / m;
+                        var minArray = concatExtremes(gd, ax, 'min');
```
alexcjohnson: Hmm, so looks like we can call `concatExtremes` twice with the same results? Could we cache? Like, clear it maybe during `supplyDefaults`, then in `concatExtremes` first look for the cached value before calculating it and setting the cache?

test/jasmine/tests/bar_test.js
```diff
@@ -1328,7 +1328,51 @@ describe('A bar plot', function() {
...
+
+            var setPositions = gd._fullData[0]._module.setPositions;
+            expect(setPositions).toHaveBeenCalledTimes(calls);
+            setPositions.calls.reset();
+        }
+
+        Plotly.plot(gd, [
+            {type: 'bar', x: [1, 2, 3], y: [1, 2, 1]},
+            {type: 'bar', x: [1, 2, 3], y: [-1, -2, -1]}
+        ])
```
alexcjohnson: Is there a test with stacked bars, where the extremes of one trace depend on the traces before it (so may need to be recalculated if an earlier trace changes visibility)?

src/plots/cartesian/autorange.js
```diff
@@ -173,15 +189,66 @@ function makePadFn(ax) {
...
+
+    for(i = 0; i < fullData.length; i++) {
+        var trace = fullData[i];
+        var extremes = trace._extremes;
+
+        if(trace.visible === true) {
+            if(Registry.traceIs(trace, 'cartesian') || Registry.traceIs(trace, 'gl2d')) {
+                var axId = ax._id;
+                if(extremes[axId]) {
+                    out = out.concat(extremes[axId][ext]);
```
alexcjohnson: In principle we can collapse `out` more than just `concat` - new items can be mooted or can moot existing items based on the [three criteria used in `expand`/`findExtremes`](https://github.com/plotly/plotly.js/pull/2849/files#diff-a701daa267f7b5c5bce68efe85034854R418)

I suppose it's worth testing with some real examples though - `doAutoRange` scales as `O(min.length * max.length)` which says it's a good idea to minimize the length of the min and max arrays, but excluding items from the `min` and `max` arrays scales as something like `O(finalArray.length * uncompressedTotalLength)` which is also sort of quadratic... not exactly a clear win either way.

alexcjohnson:
> one polar mock is currently failing, but I think it is more correct now.

I think I agree - the new one (radial autorange still happens even though angular data is invalid) matches how we handle cartesian with invalid data on the opposite axis, right?

src/plots/cartesian/autorange.js
alexcjohnson: Actually there’s a probably much more common case where I’m pretty sure it’s a big win to collapse more here: lots of traces each with only one or two min/max entries. Like 100 line or bar traces, or 100 marker traces with constant marker size. Collapsing here takes ~100 * 2 comparisons, then `doAutorange` only has one calculation to do, whereas simple concat here leaves 100 * 100 calculations for `doAutorange`

src/plots/cartesian/autorange.js
etpinard: Improvements in -> 66df51e8cf9d99e69f3968eb5e73929c55c233a2

src/plots/cartesian/constraints.js
etpinard: Some improvement in 66df51e8cf9d99e69f3968eb5e73929c55c233a2. `concatExtremes` is still called twice on graphs with constrained axes, so would do better if we want.

test/jasmine/tests/bar_test.js
etpinard: > Is there a test with stacked bars,

Thanks for writing this down. Visibility restyles are broken on stacked bars, but for a different reason: `Bar.setPositions` mutates `calcdata[i][j].b` which is first set in `Bar.calc` :upside_down_face: 

> where the extremes of one trace depend on the traces before it 

Yep, that's exactly why commit https://github.com/plotly/plotly.js/commit/4c250c3c90f061ecbe1836f8ef07ed9b561d9f6b is very important, 'plot' edit types now go call `setPositions` for this reason.

etpinard:
I'm starting to think we should abandon this PR and #2837 :cry: 

Consider

```js
Plotly.newPlot(gd, [{
  visible: false,
  y: [1, 2, 1]
}])

```

on this branch, this snippet shows a blank graph. That's because data arrays aren't coerced for `visible:false` traces and `_module.calc` is skipped for `visible!==true` traces.

So, one way to possibly save this PR is to make `_module.calc` run on 'legendonly' traces (note that data arrays are coerced for 'legendonly' traces) and make `visible` restyle calls on `visible:false` traces go through the 'calc' edit pipeline. Performance improvements would still be detected on legend toggling (where we toggle between visible `'legendonly'` and `true`) - which is really the main purpose of this PR.

etpinard:
Abandoned. Recycled commits in Abandoned. See https://github.com/plotly/plotly.js/pull/2860

PR #2860, opened 2018-07-31
Recycled commits from abandoned fast trace toggle PRs

etpinard:
based of https://github.com/plotly/plotly.js/pull/2823, supersedes https://github.com/plotly/plotly.js/pull/2837 and https://github.com/plotly/plotly.js/pull/2849.

I decided to abandoned the fast trace toggle PRs after https://github.com/plotly/plotly.js/pull/2849#issuecomment-408958619 and realizing that even a special `'legendonly'` pathway wouldn't be enough (it would break categorical axes).

So, this PR attempts to :recycle: some of the work done in the past week, some of which have small but noticeable performance benefits.

Namely,

- `fullLayout._modules` now includes trace modules of `visible!==true` traces. In particular, this leads to `fullLayout._has('gl')` returning `true` and some performance benefits for scattergl traces. For example, 

```js
setTimeout(() => {
  console.time('1')
  console.timeEnd('1')
}, 500)
```

is 100ms faster now for scattergl graphs (no matter the number of points) as we no longer clear and reinit a WebGL context.

- `Axes.expand` and `ax._min`/`ax._max` are replaced by `findExtremes` and `_extremes` containers in full trace and annotations/shapes items. Arguably, this improves maintenance as `findExtremes` does not mutate the full axis objects; it should make it easier to reuse `cartesian/autorange.js` in other subplot types. Moreover, this appears to have some performance benefits, perhaps as the pseudo O(n^2) `expand`/`findExtremes` algo is now done over a per-trace-not-axis-wide sample. For example, with

```js
// this is the largest N before we use the average marker.size
// one Axes.expand / findExtremes call here clocks in at 20-50ms on average
var N = 1e5-1
var N2 = 1e5-1

var x = []
var y = []
var ms = []
var x2 = []
var y2 = []
var ms2 = []
var i

for(i = 0; i < N; i++) {
  x.push(Math.random())
  y.push(Math.random())
  ms.push(i % 2 ? 20 : 10)
}
for(i = 0; i < N2; i++) {  
  x2.push(Math.random() + 2)
  y2.push(Math.random())
  ms2.push(i % 2 ? 10 : 20)
}

Plotly.newPlot('graph', [{
  type: 'scattergl',
  mode: 'markers',
  x: x, y: y,
  marker: {size: ms}
}, {
  type: 'scattergl',
  mode: 'markers',
  x: x2, y: y2,
  marker: {size: ms2}
}])
```

The sum of the `findExtremes` computation times (2 per x/y axes) appear 10ms faster than  the corresponding `Axes.expand` calls on master. Note that `Axes.getAutorange` is 10x slower on this branch (most likely from concatenating trace `_extremes`), but still clocks in well below the 1ms mark.

- finally, this PR adds several tests the assert axis range after `visible` restyles.

**But**, please note this PR does **not** attempt to make trace and annotations/shapes `visible` a `'plot'` edit type. They will remain a :turtle: 'calc' edit type for the next few months :disappointed: 

cc @alexcjohnson


etpinard:
Oh and I kept the work of https://github.com/plotly/plotly.js/commit/f1cda60e8681f840b60a2d1e59d8c8e108d79526 squashed in https://github.com/plotly/plotly.js/pull/2860/commits/dfada6ac7e144696d4edc42c33f1754d331cc93d that move the bar calc 'bar' mesures (i.e. `calcdata[i][j].b`) from `Bar.calc` to `Bar.setPositions` so that `Bar.setPositions` doesn't mutate what `Bar.calc` computed.

src/traces/scattergl/index.js
```diff
@@ -360,22 +359,26 @@ function clearViewport(comp, vp) {
     gl.clear(gl.COLOR_BUFFER_BIT);
 }
 
+function repeat(opt, cnt) {
+    var opts = new Array(cnt);
+    for(var i = 0; i < cnt; i++) {
+        opts[i] = opt;
+    }
+    return opts;
+}
```
alexcjohnson: non blocking, would probably make a good `Lib` function. Pretty sure we use this in other places as well, though I'm not quite sure how to search for it...

src/plots/get_data.js
```diff
@@ -69,6 +69,7 @@ exports.getModuleCalcData = function(calcdata, arg1) {
     for(var i = 0; i < calcdata.length; i++) {
         var cd = calcdata[i];
         var trace = cd[0].trace;
+        // N.B. 'legendonly' traces do not make it pass here
```
alexcjohnson: ~pass~ past

src/traces/bar/set_positions.js
```diff
@@ -633,14 +638,21 @@ function normalizeBars(gd, sa, sieve) {
...
 
 function getAxisLetter(ax) {
     return ax._id.charAt(0);
 }
 
+function putExtremes(cd, ax, extremes) {
+    for(var i = 0; i < cd.length; i++) {
+        cd[i][0].trace._extremes[ax._id] = extremes;
+    }
+}
```
alexcjohnson: Now that we're not trying to make `_extremes` survive visibility changes, does it actually help to have the same thing repeated in all traces here? Seems like that will just waste time later on. Can we just put it in the first one and clear all the others?

src/traces/box/set_positions.js
```diff
@@ -86,23 +86,26 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
         vpadminus: dPos + pad[0] * padfactor,
         vpadplus: dPos + pad[1] * padfactor
     });
+
+    for(i = 0; i < boxList.length; i++) {
+        calcTrace = calcdata[boxList[i]];
+        // set the width of all boxes
+        calcTrace[0].t.dPos = dPos;
+        // link extremes to all boxes
+        calcTrace[0].trace._extremes[posAxis._id] = extremes;
```
alexcjohnson: same question as for bars... can we just put these extremes in the first one?

src/plots/cartesian/autorange.js
```diff
@@ -174,15 +189,68 @@ function makePadFn(ax) {
...
+
+    for(i = 0; i < fullData.length; i++) {
+        var trace = fullData[i];
+        var extremes = trace._extremes;
+
+        if(trace.visible === true) {
+            if(Registry.traceIs(trace, 'cartesian')) {
+                var axId = ax._id;
+                if(extremes[axId]) {
+                    out = out.concat(extremes[axId][ext]);
```
alexcjohnson: I think my https://github.com/plotly/plotly.js/pull/2849#discussion_r205597761 and https://github.com/plotly/plotly.js/pull/2849#discussion_r205625136 still hold, and we should prune these the same way as in `findExtremes` to minimize `O(min.length * max.length)` in `doAutoRange`.

src/plots/polar/layout_defaults.js
```diff
@@ -55,6 +55,7 @@ function handleDefaults(contIn, contOut, coerce, opts) {
         // propagate the template.
         var axOut = contOut[axName] = {};
         axOut._id = axOut._name = axName;
+        axOut._traceIndices = subplotData.map(function(t) { return t.index; });
```
alexcjohnson: In cartesian you use `t._expandedIndex` - should it be the same here?

src/plots/cartesian/autorange.js
alexcjohnson: oh sorry, you did that in https://github.com/plotly/plotly.js/pull/2860/commits/a0bfaf36fe7406b3868ed882c09c73e71074999f 🎉  and also 🌴  🏆 💯

alexcjohnson:
Great job plucking out the good bits, this still ended up as a pretty big PR, that cleans up a lot of things! Should we give it a `type: bug` as well? It still includes the splom fix https://github.com/plotly/plotly.js/pull/2837#discussion_r204580575

src/plots/cartesian/autorange.js
etpinard: Yeah I tried to squash 72f06a6 into  a0bfaf3, but that was too much of a :face_with_head_bandage:

src/plots/polar/layout_defaults.js
etpinard: Good eye. Thanks!

etpinard:
> It still includes the splom fix #2837 (comment)

Yes, it does. Thanks!

src/traces/bar/set_positions.js
etpinard: Thanks for pointing this out, but turns out we still need to attach `_extremes` to all bar traces at the moment for error bars. Error bars concat their min/max extremes into the `trace._extremes` in which they are linked and assume that `trace._extremes[ax._id]` exists.

We could alternatively linked error bar `_extremes` in their `trace.error_(x|y)` containers and make `getAutorange` look for `error_(x|y)` in each trace. Or perhaps, making error bar create `trace._extremes[ax._id]` containers when needed would suffice.

But, I wouldn't worry too much about performance. I can't imagine `getAutorange` taking more than 1ms even in the worse of scenario.

src/traces/box/set_positions.js
etpinard: ... for _consistency_ I'm leaning on voting for making all traces have an `_extremes` container. Like in my last comment, I can't imagine `getAutorange` ever being a performance drag.

src/traces/bar/set_positions.js
alexcjohnson: Ah OK. I suppose the first one could have the actual values, then all the others could just get `[]` to give something for errorbars to append to... but yeah, no big deal, no block. These extras will get pruned quickly during your improved `_concat` that I hadn't noticed when I made that comment.

src/traces/scattergl/index.js
etpinard: done in https://github.com/plotly/plotly.js/pull/2860/commits/33b4085010209af0f37b8a2ca538115be76bfc84

etpinard: I didn't find any other usage of a repeat helper in `src/`, but making a lib function now should help us remember that this helper exists.

src/plots/get_data.js
etpinard: fixed in https://github.com/plotly/plotly.js/pull/2860/commits/89aebd1737bb3fe9b82ff61b8793050901eeeba7

src/plots/polar/layout_defaults.js
etpinard: fixed and :lock: in https://github.com/plotly/plotly.js/pull/2860/commits/d233f3bbadc2aacd6539fd397c701cf7dd9e5796

alexcjohnson:
Beautiful. 🍋 -> 🍹 
💃

Issue #2858, opened 2018-07-30
pie legends with conflicting colors


alexcjohnson:
Pointed out by @nicolaskruchten 

If you make several pie charts that share one or more labels but give them different colors, you get one legend entry per label, using the first color only:
https://codepen.io/alexcjohnson/pen/Zjvqee?editors=0010
![screen shot 2018-07-30 at 4 09 30 pm](https://user-images.githubusercontent.com/2678795/43420830-fade941a-9412-11e8-9031-e19bf20a2cd8.png)
and clicking the legend entry for a label makes all copies - even the ones with a different color - disappear. This is confusing, but it's not 100% clear what the right fix is. The case I'm imagining is pies representing distinct data but expressed with the same labels. Like two survey questions, "How many scoops of ice cream do you want" and "How many cherries should there be on top", each with answers "0", "1", "2", "3 or more"

The two options that come to mind for how to resolve this:
- Make one legend entry per label/color combination. This seems the most flexible, as you could for example show/hide each slice independently. But it could be confusing, especially since we don't have a way to split up the legend or to interleave subheadings, to have multiple entries with matching labels.
- Show all of the colors next to each other in a single legend entry, kind of like we do for candlestick (though that's just 2 colors, this could be arbitrarily many). This would be more compact, could avoid the confusion ^^ (though could add more confusion if there are too many colors), and if the pies were related enough in meaning you might want to analyze them together - for example, yes/no/maybe questions, you want to hide all the maybes and just compare yes and no for all of them.

etpinard:
What if we added a `piegroup` attribute? 

- Traces with different or unset `piegroup` would (like the 1st bullet pt) have one legend entry per label+color combination. 
- Traces with the same `piegroup` value would behave the second bullet pt by showing all colors (maybe cap that at 4) associated with traces of the same `piegroup`

alexcjohnson:
> What if we added a piegroup attribute?

We already have a [`scalegroup` attribute](https://github.com/plotly/plotly.js/blob/master/src/traces/pie/attributes.js#L135) so would this need to be `colorgroup`? Or perhaps we can reuse `legendgroup` for this purpose? Anyway seems like the two issues are distinct, so it's not simply a matter of renaming `scalegroup`, we need both to exist (though I suppose we could discuss whether their defaults should be linked)

etpinard:
[...]

Issue #2834, opened 2018-07-23
Fields missing in fullData


dmt0:
1. Missing marker.colorbar.tickfont.color
https://codepen.io/dmt0/pen/djNNpa?editors=1010 -  **done in** https://github.com/plotly/plotly.js/pull/2850

2. For Violin plots missing:
- bandwidth
- span
- box.*  **done in** https://github.com/plotly/plotly.js/pull/2850
- meanline.* **done in** https://github.com/plotly/plotly.js/pull/2850

https://codepen.io/dmt0/pen/MBpxLv?editors=1010

3. For choropleth:
- marker.line.color

etpinard:
> Missing marker.colorbar.tickfont.color

Yeah, that one sounds like a bug



> For Violin plots missing:
> 
> bandwidth
> span

Those are computed fields. They don't have a fixed default. Their values are punching in `gd.calcdata[i][j].t` not `gd,_fullData`. We could copy them back in fullData to make the editor happy, but I'm afraid similar issues are going to come up in other trace types. How does the editor handle other computed values like axis `dtick`, `tick0`, histograms `xbins` etc?


> meanline.*

That's the expected behavior. We remove the `meanline.` container when the violins don't have mean lines:

https://github.com/plotly/plotly.js/blob/dcaebc289b00a5da3f180d0eb42d193c96e063b1/src/traces/violin/defaults.js#L51-L54

etpinard:
> meanline.*
> 
> That's the expected behavior

Oh wait, maybe having `gd._fullData[i].meanline.visible = false` would have more sense for the editor instead of removing the container completely.

etpinard:
... would be nice to talk about these during the weekly meeting on Thursday.

etpinard:
generated pie `marker.colors` are also missing from `fullData`: https://codepen.io/veraz/pen/YjxgvE?editors=1010

cc @VeraZab

etpinard:
Pie `marker.colors`, violin `span`, `bandwidth` all fall under the same category: their defaults are _computed_ during the calc step (which occurs after `Plots.supplyDefaults`) and reside in `gd.caldata`, **not** `gd._fullData`.

These attributes differ from e.g. contour's `contours.(start|end|size)` or histogram's `(x|y).bins` which have their default values computed in calc but punched back into `gd.data` to preserve the correct `autocontours` / `autobins` behavior.

So, I see two possible solutions:

[...]

PR #2867, opened 2018-08-03
pie color fix & enhancements

alexcjohnson:
- Fix inheritance of explicit colors by later traces (see https://github.com/plotly/plotly.js/issues/2834#issuecomment-408965483)
- Allow inheritance (of explicit colors) by earlier traces too - @etpinard I did this by adding a `calcInteractions` step immediately after `calc`. In principle I feel like we could move `setPositions` over to `calcInteractions`, but not vice versa because `setPositions` is tied to cartesian subplot. But I didn't touch `setPositions` in this PR. I also tried putting this code into `Pie.plot` but this failed because the colors are needed for the legend, which gets drawn before the data.
- Add `layout.piecolorway` and `layout.extendpiecolors` for more control over pie colors (cc @nicolaskruchten @VeraZab)
- Does not address #2858

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2435,8 +2435,6 @@ plots.doCalcdata = function(gd, traces) {
 
     // for sharing colors across pies (and for legend)
     fullLayout._piecolormap = {};
-    fullLayout._piecolorway = null;
-    fullLayout._piedefaultcolorcount = 0;
 
     // If traces were specified and this trace was not included,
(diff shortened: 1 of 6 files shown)


nicolaskruchten:
Wow, this seems amazing, thanks Alex! It'll really help make the UX of the editor wayyy better :)

alexcjohnson:
Closing this PR, will recreate it once #2868 is merged.

Issue #2810, opened 2018-07-12
Polar 2.0 replacement for legacy "area" type


alexcjohnson:
A request has come in for an updated version of the old `area` type. As mentioned by @etpinard in #2255:

> How to replace legacy polar's `area` trace type? I'd vote for replacing it with `polarbar` or `wedge`. Most of the `bar` logic could be reused here.

Yes, it seems like this trace type is precisely `bar` transformed onto polar axes. For consistency with other trace type names perhaps it should be `barpolar` though?

Looks to me as though essentially all the existing `bar` trace attributes will translate to the new type straightforwardly. In fact, there's even an analog of `orientation`, though instead of `(x|y)` we'd have `(radial|angular)`. For the request at hand only `radial` (which matches the legacy `area` type) is necessary, but some day someone will want `angular` as well, like https://peltiertech.com/time-is-on-my-side/
![ring chart](https://peltiertech.com/wp-content/img200808/donut-clock-in-excel.png)

The `layout` attributes (`barmode`, `barnorm`, `bargap`, `bargroupgap`) to the extent that we want them at all, we can make into attributes of the `polar<N>` subplot. I would suggest omitting (at least initially) `barmode: 'group'` and `'relative'`, and therefore also `bargroupgap`. Probably also omit `barnorm`, that only enhances the confusion over whether it's the radius or area that encodes value.

spicynaresh:
add Aster Plot

PR #2868, opened 2018-08-03
setPositions -> crossTraceCalc

etpinard:
as requested by @alexcjohnson 

`setPositions` is currently only used for bar, histogram, box and violin traces, but will soon (cc https://github.com/plotly/plotly.js/pull/2867 and https://github.com/plotly/plotly.js/issues/2810) be used in non-cartesian subplots and not just to _set_ positions but a few other "cross-trace" things.


src/traces/histogram/index.js
```diff
@@ -12,7 +12,7 @@
 /**
  * Histogram has its own attribute, defaults and calc steps,
  * but uses bar's plot to display
- * and bar's setPositions for stacking and grouping
+ * and bar's crossTraceCalc (formally known as setPositions) for stacking and grouping
```
alexcjohnson: ~formally~ formerly

alexcjohnson:
Looks great, thanks for doing that! 💃 after fixing the little typo 😅

src/traces/histogram/index.js
etpinard: done in -> e2360382c4e13dc22b8cce928ad4e1bfa7c4ad87

PR #2870, opened 2018-08-04
Pie colors fix & enhancements

alexcjohnson:
Reboot of #2867

Original commit https://github.com/plotly/plotly.js/commit/d50334e8a16b033f4613cc310a17559e03865afd:
- Fix inheritance of explicit colors by later traces (see https://github.com/plotly/plotly.js/issues/2834#issuecomment-408965483)
- Allow inheritance (of explicit colors) by earlier traces too by adding a `calcInteractions` step immediately after `calc`.
- Add `layout.piecolorway` and `layout.extendpiecolors` for more control over pie colors (cc @nicolaskruchten @VeraZab)
- Does not address #2858

Update after #2868, commit https://github.com/plotly/plotly.js/commit/ded6e945250b910261101ccd92acd20c662abe30:
- Get rid of `calcInteractions` and uses `crossTraceCalc` for pies
- Update `crossTraceCalc` to work with pie (and other subplot-less traces)
- Simplify some things by moving this into `doCalcdata`. In particular, it doesn't need to happen in any other places besides `doCalcdata`, and `Errorbars.calc` can move in here too since it just needed to happen after `crossTraceCalc`.

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2236,8 +2236,6 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
 
         plots.supplyDefaults(gd);
         plots.doCalcdata(gd);
-        plots.doCrossTraceCalc(gd);
-        Registry.getComponentMethod('errorbars', 'calc')(gd);
 
         return Promise.resolve();
     }
@@ -2503,8 +2501,6 @@ plots.doCalcdata = function(gd, traces) {
     // clear stuff that should recomputed in 'regular' loop
     if(hasCalcTransform) clearAxesCalc(axList);
 
-    var calcInteractionsFuncs = [];
-
     function calci(i, isContainer) {
         trace = fullData[i];
         _module = trace._module;
@@ -2530,12 +2526,6 @@ plots.doCalcdata = function(gd, traces) {
 
             if(_module && _module.calc) {
                 cd = _module.calc(gd, trace);
-
-                // Some modules need to update traces' calcdata after
-                // *all* traces have been through calc - so later traces can
-                // impact earlier traces.
-                var calcInteractions = _module.calcInteractions;
-                if(calcInteractions) Lib.pushUnique(calcInteractionsFuncs, calcInteractions);
             }
         }
 
@@ -2561,9 +2551,10 @@ plots.doCalcdata = function(gd, traces) {
     for(i = 0; i < fullData.length; i++) calci(i, true);
     for(i = 0; i < fullData.length; i++) calci(i, false);
 
-    for(i = 0; i < calcInteractionsFuncs.length; i++) calcInteractionsFuncs[i](gd, calcdata);
+    plots.doCrossTraceCalc(gd);
 
     Registry.getComponentMethod('fx', 'calc')(gd);
+    Registry.getComponentMethod('errorbars', 'calc')(gd);
 };
 
 function clearAxesCalc(axList) {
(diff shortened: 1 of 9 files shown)


src/plots/plots.js
```diff
@@ -2561,9 +2551,10 @@ plots.doCalcdata = function(gd, traces) {
     for(i = 0; i < fullData.length; i++) calci(i, true);
     for(i = 0; i < fullData.length; i++) calci(i, false);
 
-    for(i = 0; i < calcInteractionsFuncs.length; i++) calcInteractionsFuncs[i](gd, calcdata);
+    plots.doCrossTraceCalc(gd);
```
alexcjohnson: Hmm, I guess we don't need to export `doCrossTraceCalc` anymore since it's only used here...

test/jasmine/tests/annotations_test.js
```diff
@@ -775,7 +775,7 @@ describe('annotations autorange', function() {
             });
         })
         .then(function() {
-            _assert('auto rng / big tx', [-0.22, 3.57], [0.84, 3.365]);
+            _assert('auto rng / big tx', [-0.22, 3.59], [0.84, 3.365]);
```
alexcjohnson: Doesn't matter for CI, and has nothing to do with this PR, but this change makes this test pass on my Mac.

test/jasmine/tests/bar_test.js
```diff
@@ -1801,14 +1801,6 @@ function mockBarPlot(dataWithoutTraceType, layout) {
     supplyAllDefaults(gd);
     Plots.doCalcdata(gd);
 
-    var plotinfo = {
-        xaxis: gd._fullLayout.xaxis,
-        yaxis: gd._fullLayout.yaxis
-    };
-
-    // call Bar.crossTraceCalc
-    Bar.crossTraceCalc(gd, plotinfo);
```
alexcjohnson: ⚠️ ⚠️ ⚠️  Removed because this is included in `doCalcdata` now, but in fact the test FAILS if we leave this in. That means `Bar.crossTraceCalc` is not idempotent, though I didn't really look into what it's doing. I don't think this is particularly a problem, but it might indicate it's more fragile than it should be.

src/plots/plots.js
alexcjohnson: stopped exporting -> https://github.com/plotly/plotly.js/pull/2870/commits/0bf82c862047304d4c1d1c6f205f773b6ceec044

test/jasmine/tests/bar_test.js
etpinard: > That means Bar.crossTraceCalc is not idempotent, though I didn't really look into what it's doing.

Yep, I noticed that last week. This commit https://github.com/plotly/plotly.js/pull/2860/commits/dfada6ac7e144696d4edc42c33f1754d331cc93d helps, but there are more cases where `Bar.setPositions` oops I mean `Bar.crossTraceCalc` mutates `gd.calcdata`.

test/jasmine/tests/pie_test.js
```diff
@@ -138,6 +138,50 @@ describe('Pie traces:', function() {
...
+        .then(_checkSliceColors(['255,0,0', '0,0,0', '0,0,0', '255,0,0']))
+        .then(function() {
+            return Plotly.newPlot(gd, data2);
+        })
+        .then(_checkSliceColors(['0,0,0', '255,0,0', '255,0,0', '0,0,0']))
+        .catch(failTest)
+        .then(done);
+    });
+
+    it('can use a separate pie colorway and disable extended colors', function(done) {
```
etpinard: Great tests!

etpinard:
Very nice.

Merging the old `doSetPositions` with `doCalcdata` is a brilliant idea. I can't imagine a case where we'd want to call `doSetPositions` w/o first calling `doCalcdata` and vice-versa. As, `doCrossTraceCalc` is clever enough to only call the required `_module.crossTraceCalc` methods, this won't lead any perf loses.

As an aside, I think it might be best at some point to move the `ErrorBars.calc` call down to the trace module's own `calc` methods (similar to `Colorscale.calc`) to spare us a another loop over all the traces, but this can wait.

:dancer:

alexcjohnson:
> I think it might be best at some point to move the `ErrorBars.calc` call down to the trace module's own `calc` methods (similar to `Colorscale.calc`) to spare us a another loop over all the traces

Excellent idea. In general it needs to be in `crossTraceCalc` though, as it can be impacted by stacking (leaving aside the question of whether it's actually a good idea to put errorbars on stacked items... sometimes perhaps, other times 🙈 )

Issue #2859, opened 2018-07-31
Sankey separator


simonerubiu:
Hi there,
I need to set the decimal and thousands separators in sankey chart, I found out the layout.separators option that works for other kind of charts, but when I use this options in sankey nothing happen.
I think this is a bug, can you explain me how to fix this issue?? or if there is a branch where this issue is just resolved.

Thanks in advance ;)

alexcjohnson:
Thanks @simonerubiu - You're correct that our Sankey diagrams don't currently support `layout.separators`, and you're correct that this is a bug :)

Unfortunately our number formatting code is a bit chaotic at the moment, and ripe for refactoring... There's [`Lib.numSeparate`](https://github.com/plotly/plotly.js/blob/0c1d6d5ebbd2334f712aede2a32236fb451838b7/src/lib/index.js#L902) that handles separators for simple floating-point representations, [`axes->numFormat`](https://github.com/plotly/plotly.js/blob/0c1d6d5ebbd2334f712aede2a32236fb451838b7/src/plots/cartesian/axes.js#L1242) that handles rounding and exponents along with separators but is tightly coupled to cartesian axes, and [`fullLayout._d3locale.numberFormat`](https://github.com/plotly/plotly.js/blob/0c1d6d5ebbd2334f712aede2a32236fb451838b7/src/plots/plots.js#L371) for using d3 format strings (including localization and override by `layout.separators`)

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2894, opened 2018-08-10
sankey separators

simonerubiu:
Referring to #2859 I find out a solution to solve this issue with sankey separators.
While fixing this issue I find out that the separatethousands options there isn't in sankeycharts, and  I try to implement it in the correct way. Another problem I found is that the numSeparate function not works properly with 4length numbers,


etpinard:
Thanks very much for your help @simonerubiu !

So, if I understand correctly this PR does 4 things:

1. fixes the italian translation
2. fixes `layout.separators` for sankey traces (as tracked by https://github.com/plotly/plotly.js/issues/2859)
3. adds `separatethousands` to sankey traces
4. fixes a bug with `Lib.numSeparate`

which is a little much for _one_ PR. Moreover, we wouldn't want to pollute our commit history with those ".gitingore" -> "revert .gitignore" commits off this branch.

So, I suggest:

- making 1 PR just for item 1,
- create a reproducible example for the `Lib.numSeparate` bug you discovered so that we can confirm it's in fact a bug
- make a separate PR for items 2 and 3 and make sure all the tests pass

Note also, you changed the file mode in the files you modified from `100644 → 100755`. Please try to avoid that.

Thanks again for contributing to plotly.js. Let me know if you have any other questions.

etpinard:
I'll close due to lack of activity. 

@simonerubiu feel free to open another PR (or PRs) following my recommandations in https://github.com/plotly/plotly.js/pull/2894#issuecomment-412087609 

Thanks again for contributing to plotly.js!

Issue #2888, opened 2018-08-09
scatterpolargl in 'lines' mode smears on axis rotation


jonmmease:
For a `scatterpolargl` trace in `lines` `mode`, when the axes are rotated by clicking and dragging on an axis label, the lines smear across the plot. The smearing is removed on release:

CodePen: https://codepen.io/anon/pen/MBzLKp

Rotation example:
![scatterpolargl_smearing](https://user-images.githubusercontent.com/15064365/43892568-4b3381d0-9b9a-11e8-9a4c-0eeecea013b0.gif)

FWIW, this doesn't occur in `markers` `mode`.

alexcjohnson:
But it's so pretty :)

Confirmed - in fact, doesn't seem to happen in `lines+markers` mode, only in `lines` with or without `fill`

etpinard:
Thanks for the report @jonmmease !

Issue #2887, opened 2018-08-09
Scatterpolar hoverinfo inconsistent in 'toself' fill state


jonmmease:
During a [discussion in the plotly.py forums](https://community.plot.ly/t/windrose-chart-customization/12294/6) in came up that the hoverlabel for the `scatterpolar` trace has a different behavior when the `fill` state is set to `toself` than when it's set to `none`.

CodePen: https://codepen.io/anon/pen/EpOGdX

`fill: 'none'` for `scatterpolar`
![screen shot 2018-08-09 at 5 52 32 am](https://user-images.githubusercontent.com/15064365/43891927-91b48ca0-9b98-11e8-8f5c-eae8ef370d64.png)

`fill: 'toself'` for `scatterpolar`
![screen shot 2018-08-09 at 5 52 59 am](https://user-images.githubusercontent.com/15064365/43891938-9d875670-9b98-11e8-9388-26264a00c9fa.png)

It also seems that the `hoverinfo` propery has no effect when `fill` is `toself`.

I did the same comparison with the `scatterpolargl` trace and it has the behavior that I expected:

`fill: 'toself;` for `scatterpolargl`
![screen shot 2018-08-09 at 5 56 18 am](https://user-images.githubusercontent.com/15064365/43892066-f6761244-9b98-11e8-9817-8a2a199025dc.png)

I don't know if this is a bug in `scatterpolar` or `scatterpolargl`, but I find the current `scatterpolargl` behavior (The behavior consistent with `fill` of `none`) to be more helpful.

alexcjohnson:
This is the [`hoveron`](https://plot.ly/javascript/reference/#scatterpolar-hoveron) attribute, which defaults to `fills` in this case. This is consistent with `scatter` behavior, `scatterpolargl` seems to be the odd one out.

etpinard:
@alexcjohnson is right `scatterpolargl` is the odd one out. I didn't know fills in `scatterpolargl` were a thing. That's what you get from reusing polar + scattergl logic I guess. But I doubt they function properly yet along have the correct hoverinfo.

etpinard:
Looking at this more closely, the _real_ problem is with `scattergl`.

[...]

PR #2737, opened 2018-06-14
Enable text for scattergl trace

dy:
Implements https://github.com/plotly/plotly.js/issues/2501

* [x] Text rendering via [gl-text](https://ghub.io/a-vis/text)
  * [x] antialiasing
  * [x] kerning
  * [x] 1e6 characters (test)
  * [x] horiz/vert alignment
* [x] `gl_text_chart_*` mocks

Issue #2797, opened 2018-07-10
Scattergl box/lasso selection on broken overlaying-y-axis subplots


vincex:
https://codepen.io/anon/pen/BPaQOe?editors=1010

If you try to select some points with lasso or box selection only the last added trace is visible.
The "autoscale" buttons fix the visualization problem until a new selection is made.

I dont have a deep knowledge of ploty and I am not able to investigate further. But from what i was able to see seems that with the autoscale the redraw of each subplot is in sync, instead with the selection the redraw of each subplot is async and only the last one is redrawed.

Thanks for your time guys.

etpinard:
Thanks for the report!

etpinard:
Related to https://github.com/plotly/plotly.js/issues/2562, which might be the best way to fix this bug.

@dy what do you think?

etpinard:
More minimal reproducible example: https://codepen.io/etpinard/pen/ajbGOx

PR #2895, opened 2018-08-10
Better polar setConvert + a few misc polar touch ups

etpinard:
This is a preliminary PR for `barpolar`.

I dove back hard into the polar code in the past 10 days and I noticed a few hacks that were worth cleaning up before adding `barpolar`.

Commit https://github.com/plotly/plotly.js/commit/3d8eef2b7b7e258e094395e65186ddea725c6f0b is the big one. In brief, it locks down what 'c'alc means for angular coordinates and adds a new coordinate system (I called it 'g'eometric) that depends on things that interactions can update (i.e. `radialaxis.range` and `angularaxis.rotation`). I invite you to read the `setConvertPolar` docstring as well as commit https://github.com/plotly/plotly.js/commit/91064cac5c3f7ed14741c96641f9a7436a19c33b for more info. It's still not perfect: ideally we should try to factor things in `Axes.doTicks` to no longer rely on 2 radial and 2 angular axis objects per polar subplots, but this feels like a step in the right direction.

After that, this PR also:

- adds support `r0/dr` and `theta0/dtheta`. I decided to compute the `dtheta` default using the angular period and the number of `r` coordinates in the trace - which seems to give pretty user-friendly results and will be useful for `barpolar`.
- fixes https://github.com/plotly/plotly.js/issues/2888
- adds on-graph text to `scatterpolargl` following-up on #2737
- :hocho: un-implemented `hoveron` from the `scattergl` and `scatterpolargl` attributes as discussed in https://github.com/plotly/plotly.js/issues/2887#issuecomment-411911140

cc @alexcjohnson


src/traces/bar/calc.js
```diff
@@ -15,17 +15,11 @@ var arraysToCalcdata = require('./arrays_to_calcdata');
...
-    // and data ranges
-    // note: this logic for choosing orientation is
-    // duplicated in graph_obj->setstyles
+    var xa = Axes.getFromId(gd, trace.xaxis || 'x');
+    var ya = Axes.getFromId(gd, trace.yaxis || 'y');
+    var size, pos;
 
-    var xa = Axes.getFromId(gd, trace.xaxis || 'x'),
-        ya = Axes.getFromId(gd, trace.yaxis || 'y'),
-        orientation = trace.orientation || ((trace.x && !trace.y) ? 'h' : 'v'),
```
alexcjohnson: Nice catch!

src/plots/polar/set_convert.js
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Angular axis coordinate systems flow:
+ * + for linear axes:
+ *   - d2c (in calc) angles -> 'data' radians
+ *   - c2g (in plot) 'data' -> 'geometric' radians (fn of ax rotation & direction)
+ *   - t2g (in updateAngularAxis) 'tick' value (in degrees) -> 'geometric' radians
+ * + for category axes:
+ *   - d2c (in calc) just like for cartesian axes
+ *   - c2g (in plot) category indices -> 'geometric' radians
+ *   - t2g (in updateAngularAxis) 'tick' value (as category indices) -> 'geometric' radians
```
alexcjohnson: This would be easier for me to understand if instead of describing `t2g` etc, just describe `t` and `g` (and `c` and `d` if they're different from cartesian) for the linear and category cases)

alexcjohnson: (like https://github.com/plotly/plotly.js/blob/def6aa5a24527c68e90f50485255170609640e43/src/plots/cartesian/set_convert.js#L35-L50)

src/plots/polar/set_convert.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    // N.B. we mock the axis 'range' here
+    ax.setGeometry = function() {
+        var sector = polarLayout.sector;
+        var dir = {clockwise: -1, counterclockwise: 1}[ax.direction];
+        var rot = deg2rad(ax.rotation);
+
+        ax.rad2g = function(v) { return dir * v + rot; };
+        ax.g2rad = function(v) { return (v - rot) / dir; };
```
alexcjohnson: As discussed in slack, seems like we can 🔪 `ax.rad2*` and `ax.*2rad` entirely now, so we don't need to think about the difference between `rad` and `g` to understand all this.

src/plots/polar/polar.js
```diff
@@ -156,6 +156,32 @@ proto.updateLayers = function(fullLayout, polarLayout) {
...
+ *   used for calcdata -> geometric conversions (aka c2g) during the plot step
+ *   + setGeometry setups ax.c2g given ax.rotation, ax.direction & ax._categories,
+ *                 and mocks ax.range
+ *   + setScale setups ax._m,ax._b with that mocked ax.range
+ *
+ * - this.xaxis
+ * - this.yaxis
+ *   setup so that polar traces can reuse plot methods of Cartesian traces
+ *   which mostly rely on 2pixel methods (e.g ax.c2p)
+ */
```
alexcjohnson: Thanks! 📚

src/plots/polar/set_convert.js
etpinard: done inhttps://github.com/plotly/plotly.js/pull/2895/commits/f5bceda77844160879033d283f2a4cf914924a7a

src/plots/cartesian/axes.js
```diff
@@ -1682,7 +1682,7 @@ axes.doTicksSingle = function(gd, arg, skipTitle) {
     var valsClipped = vals.filter(clipEnds);
 
     // don't clip angular values
-    if(ax._id === 'angular') {
+    if(ax._id === 'angularaxis') {
```
alexcjohnson: not new in this PR, but I think you have an `axid` here.

src/traces/scatterpolar/attributes.js
```diff
@@ -28,6 +28,49 @@ module.exports = {
...
+    dtheta: {
+        valType: 'number',
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'Sets the theta coordinate step.',
+            'By default, the `dtheta` step equals the subplot\'s period divided',
+            'by the length of the `r` coordinates.'
+        ].join(' ')
+    },
```
alexcjohnson: One of the most underrated features, IMHO. Thanks for adding it!

src/traces/scattergl/index.js
```diff
@@ -302,6 +302,8 @@ function sceneUpdate(gd, subplot) {
             }
             if(scene.scatter2d) {
                 clearViewport(scene.scatter2d, vp);
+            } else if(scene.line2d) {
+                clearViewport(scene.line2d, vp);
```
alexcjohnson: The first arg to `clearViewport` is just used to find the `regl._gl` context... and so I guess the reason `select2d` is off by itself is that select gets its own context separate from the main data context?

Might be clearer then to refactor this as something like
```js
var anyComponent = scene.scatter2d || scene.line2d || (scene.glText || [])[0];
if(anyComponent) clearViewport(anyComponent, vp);
```

Which would also make it easier to add `error2d` (can that ever be the only thing drawn?) and `fill2d` to the lookup list, looks plausible that these also have bugs similar to #2888.

BTW, looking at what `clearViewport` does - ie clears a rectangle from the context - what happens if two polar subplots have overlapping bounding boxes? Like if you made a hexagonal packing of polar subplots or you do something funny with sectors, like you make one large pac-man subplot eating a smaller one?

etpinard: > what happens if two polar subplots have overlapping bounding boxes?

This happens -> https://github.com/plotly/plotly.js/issues/2797

src/plots/polar/polar.js
```diff
@@ -993,11 +993,13 @@ proto.updateRadialDrag = function(fullLayout, polarLayout) {
             var moduleCalcDataVisible = Lib.filterVisible(moduleCalcData);
             var _module = moduleCalcData[0][0].trace._module;
             var polarLayoutNow = gd._fullLayout[_this.id];
+            var isGL = Registry.traceIs(k, 'gl');
+
+            if(isGL && _this._scene) _this._scene.clear();
```
alexcjohnson: is it possible to not have `_this._scene` at this point if there's a `gl` trace?

BTW this would be clearer with `traceType` instead of `k`.

I don't see a problem at the moment because there's only one `gl` type that can be here, but for future reference seems like if we ever add another we'll have to make sure only the first one clears the scene.

test/jasmine/tests/polar_test.js
```diff
@@ -1136,7 +1143,9 @@ describe('Test polar interactions:', function() {
                 fig.layout.margin = {l: 50, t: 50, b: 50, r: 50};
 
                 if(s.patch) s.patch(fig);
-                nTraces = fig.data.length;
+                nTraces = fig.data.reduce(function(acc, trace) {
+                    return (trace.type === 'scatterpolargl') ? ++acc : acc;
```
alexcjohnson: hmm, OK. I always find prefix `++` confusing, and perhaps because it encourages stuff like that `reduce` also often comes out confusing to me. I realize it's better perf this way, but that's irrelevant to a test.
`fig.data.filter(function(trace) { return trace.type === 'scatterpolargl'; }).length`?

src/plots/cartesian/axes.js
```diff
@@ -554,7 +554,7 @@ axes.calcTicks = function calcTicks(ax) {
 
     // If same angle over a full circle, the last tick vals is a duplicate.
     // TODO must do something similar for angular date axes.
-    if(ax._id === 'angular' && Math.abs(rng[1] - rng[0]) === 360) {
+    if(ax._id === 'angularaxis' && Math.abs(rng[1] - rng[0]) === 360) {
```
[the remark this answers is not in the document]
etpinard: good call -> 0c517d197dce032bcd5835adf80952672c2341fe

src/plots/cartesian/axes.js
etpinard: went one step further in 0c517d197dce032bcd5835adf80952672c2341fe

src/plots/polar/set_convert.js
etpinard: latest attempt in -> 4c11c84251cfe8aa532f42a26022803e867f4ed4

test/jasmine/tests/polar_test.js
etpinard: you're right, filter is better: b4700cabda7752b3d21f1277a37dd912664a2805

src/traces/scattergl/index.js
etpinard: > Might be clearer then to refactor this as something like

Absolutely -> c3044e094eacfc25fbd2f85648c66c4bf8693994

> add error2d (can that ever be the only thing drawn?)

Traces with error bar will always have either an active `scene.scatter2d` or `scene.line2d`.


> BTW, looking at what clearViewport does - ie clears a rectangle from the context -

We should take a look at how much of a perf gain is to clear only a fraction of the canvas on  pan/select. If that's negligible, then I'm thinking we should be keeping track of a graph-wide regl-draw-queue instead, always clear the whole canvas and draw all regl parts on pan/select. This should make this a lot more manageable. cc @dy

src/plots/polar/polar.js
etpinard: Thanks for tip. Improvements in -> b81885d7315b1d16e609e7d1ca7af3311ea3dbaa

src/plots/polar/set_convert.js
alexcjohnson: Much clearer - thanks!

src/traces/scattergl/index.js
alexcjohnson: > Traces with error bar will always have either an active `scene.scatter2d` or `scene.line2d`.

Even if `mode: 'none'`? That works fine for `scatter`, and sometimes tries to work with `scattergl` #2900

`fill` works with `mode: 'none'` too.

etpinard: Bulletproofed in https://github.com/plotly/plotly.js/pull/2895/commits/75786e40279e1cfbbc04333be7fb0d15d2f3be39

alexcjohnson:
Nice work! 💪 💃

PR #2899, opened 2018-08-13
Multi-axis-type sploms

etpinard:
This PR fixes axis tick rendering for categorical sploms For example, on master

```js
Plotly.newPlot(gd, [{
  type: 'splom',
  dimensions: [{
    values: ['a','b','d']
  }, {
    values: ['a', 'b', 'c']
  }]
}])
```

renders as:

image

and add supports for sploms that generate axis of different types, see 9da5a1bcf270250011a1515fada1e5002cc75f14 for an example.

cc @alexcjohnson


src/plots/cartesian/type_defaults.js
```diff
@@ -89,9 +89,10 @@ function setAutoType(ax, data) {
     }
     else if(d0.type === 'splom') {
         var dimensions = d0.dimensions;
+        var diag = d0._diag;
```
etpinard: Recall that `_diag` is:

https://github.com/plotly/plotly.js/blob/def6aa5a24527c68e90f50485255170609640e43/src/traces/splom/defaults.js#L104-L124

test/image/mocks/splom_multi-axis-type.json
```diff
@@ -0,0 +1,45 @@
...
+  ],
+  "layout": {
+    "hovermode": "closest",
+    "margin": {
+      "b": 80,
+      "l": 80,
+      "r": 30,
+      "t": 30
+    },
+    "yaxis2": {"type": "category"},
```
etpinard: This here might be tricky to find for users as e.g. nothing indicates that `yaxis2` corresponds to the boolean data array.

Perhaps we could include this in the trace `dimensions[i]` containers?

src/plots/cartesian/type_defaults.js
alexcjohnson: ah ok - a little confusing as usually `xa` is a full axis object, an id would be `xId` or something... but anyway nice fix!

src/traces/splom/index.js
```diff
@@ -36,19 +36,39 @@ function calc(gd, trace) {
...
+            if(xa) {
+                makeCalcdata(xa, dim);
+                if(ya && ya.type === 'category') {
+                    ya._categories = xa._categories.slice();
+                }
+            } else if(ya) {
+                makeCalcdata(ya, dim);
+                if(xa && xa.type === 'category') {
+                    xa._categories = ya._categories.slice();
+                }
```
alexcjohnson: How could you get here? if `xa` exists you'll skip the `else if(ya)` block entirely.

But this raises an interesting (if possibly nonsensical) point - what if `xa.type !== ya.type`? I guess  this could happen if one is numeric or date and the other is categorical, or if one is linear and the other is log. If the answer is "all hell breaks loose", which wouldn't surprise me, it may be appropriate to just log a warning and hide that dimension entirely. That seems like an abuse of the concept of a splom.

test/image/mocks/splom_multi-axis-type.json
alexcjohnson: > Perhaps we could include this in the trace `dimensions[i]` containers?

That's a great idea - `dimension.type = 'category'` to override the autotype for both axes. That would certainly reduce (though not eliminate) the chance of mismatches like https://github.com/plotly/plotly.js/pull/2899#discussion_r209777451

src/traces/splom/index.js
alexcjohnson: > a users sets ax.type in the layout
> ... [doesn't] happen by accident.

Hmm could happen in the editor though. @nicolaskruchten @VeraZab not sure how best to handle this. If we add [`dimension.type`](https://github.com/plotly/plotly.js/pull/2899#discussion_r209731332) we should steer people there somehow when there's a splom on that axis. But if they still set an axis type directly would we want to find the corresponding opposite axis in the splom and set it to match, to avoid this error?

src/plots/cartesian/type_defaults.js
etpinard: cleaned up in -> https://github.com/plotly/plotly.js/pull/2899/commits/f3c73dbd774f2280b7e0cdfee0d4170e158b889d

src/traces/splom/index.js
nicolaskruchten: hoo boy. Well right now we don't support SPLOMs yet, and I'm thinking that when we do we should lock down axes/subplots quite tightly so that people can't break them too much :)

test/image/mocks/splom_multi-axis-type.json
```diff
@@ -16,11 +16,13 @@
         },
         {
           "label": "bool",
-          "values": [false, true, true, true, false, true, false, false, false, true]
+          "values": [false, true, true, true, false, true, false, false, false, true],
+          "axis": {"type": "category"}
```
alexcjohnson: Yeah, that's way cleaner! 🎉

Good call making an `axis` container, I can see us wanting to allow other axis settings (gridlines, tick spacing, tick text...) to be set at the dimension level in the future.

I started thinking, since it's (generally) going to apply this to 2 axes, perhaps it should be called `axes`... but then that (elsewhere in the schema) implies a list of axes rather than axis properties. So I think it's good the way you have it.

test/jasmine/tests/splom_test.js
```diff
@@ -357,6 +357,42 @@ describe('Test splom trace defaults:', function() {
...
+        });
+
+        var fullLayout = gd._fullLayout;
+        expect(fullLayout.xaxis.type).toBe('linear');
+        expect(fullLayout.yaxis.type).toBe('linear');
+        expect(fullLayout.xaxis2.type).toBe('category');
+        expect(fullLayout.yaxis2.type).toBe('category');
+    });
+
+    it('axis type setting should be skipped when dimension is not visible', function() {
```
alexcjohnson: agreed, good call.

test/jasmine/tests/splom_test.js
```diff
@@ -359,6 +359,36 @@ describe('Test splom trace defaults:', function() {
...
+                values: [2, 1, 2]
+            }]
+        }, {
+            xaxis: {type: 'category'},
+            yaxis: {type: 'linear'}
+        });
+
+        var cd = gd.calcdata[0][0];
+
+        expect(cd.t._scene.matrixOptions.data).toBeCloseTo2DArray([[2, 1, 2]]);
```
alexcjohnson: This is good, but I have various questions about what happens downstream from `calc` with the missing dimension(s)... like, `cdata` and `ldata` arrays don't match up with the dimension array anymore, does this get handled correctly in plotting? The next commit (`visibleDims`) seems like it assuages my concerns a bit, but would you mind including a mismatched-type dimension (somewhere in the middle of `dimensions`) in a mock so we can see it all the way to the end?

src/traces/splom/index.js
```diff
@@ -59,16 +59,17 @@ function calc(gd, trace) {
             xa = AxisIDs.getFromId(gd, trace._diag[i][0]);
             ya = AxisIDs.getFromId(gd, trace._diag[i][1]);
 
+            // if corresponding x & y axes don't have matching types, skip dim
+            if(xa && ya && xa.type !== ya.type) continue;
```
alexcjohnson: Seems likely to confuse people if they encounter this issue... do you want to include a `Log` message, or perhaps find a way to surface this in `Plotly.validate`?

test/image/mocks/splom_multi-axis-type.json
alexcjohnson: haha exactly my rationale in https://github.com/plotly/plotly.js/pull/2899#discussion_r210114042 - I should read *all* your comments before making my own. Anyway, love it.

test/jasmine/tests/splom_test.js
```diff
@@ -1039,4 +1040,105 @@ describe('@gl Test splom select:', function() {
...
+                drawCnt: 1,
+                matrixTraces: 2,
+                selectBatch: null,
+                unselectBatch: []
+            });
+        })
+        .then(function() { return Plotly.relayout(gd, 'dragmode', 'pan'); })
+        .then(function() {
+            _assert('under dragmode:pan with NO active selection', {
+                updateCnt: 1,       // to clear off 1 matrixTrace
```
etpinard: Before this commit, this 

peek 2018-08-15 12-46

was happening.

test/jasmine/tests/splom_test.js
```diff
@@ -1039,4 +1040,105 @@ describe('@gl Test splom select:', function() {
...
+                drawCnt: 0,
+                matrixTraces: 1,
+                selectBatch: null,
+                unselectBatch: null
+            });
+        })
+        .then(function() { return Plotly.relayout(gd, 'dragmode', 'select'); })
+        .then(function() {
+            _assert('under dragmode:select', {
+                updateCnt: 3,     // updates positions, viewport and style in 3 calls
```
etpinard: Recall that `relayout(gd, 'dragmode', 'select')` results in a plot edit here because of:

https://github.com/plotly/plotly.js/blob/9d570fcc7cae551b3b4041010652ea412dd2ddaa/src/plot_api/plot_api.js#L1980-L1986

test/jasmine/tests/splom_test.js
etpinard: Good call, done in -> https://github.com/plotly/plotly.js/pull/2899/commits/390f29202e4230035e094e9731f4f643e5ca9ea9?short_path=10d6861#diff-10d68616ed667ce29da903fcd8380c29

src/traces/splom/index.js
etpinard: As this is part of the calc step (which is outside the scope of `Plotly.validate`) and that we can't easily move this to the defaults step (as axis defaults are coerced after trace defaults), I chose to only log something in https://github.com/plotly/plotly.js/pull/2899/commits/182390442ee0843e022e32a311baf138a18c4570

Note that I chose I was first planning on using `Lib.warn` instead of `Lib.log`, but currently the image server bails whenever a mock results in a `Lib.warn`, so the mock added in https://github.com/plotly/plotly.js/pull/2899/commits/390f29202e4230035e094e9731f4f643e5ca9ea9 would fail to generate an image. I hope you're ok with just a `Lib.log` here.

test/jasmine/tests/splom_test.js
alexcjohnson: Nice find, thanks for including the fix here!

alexcjohnson:
Excellent, this is really going to help robustness and usability for splom traces. 💃

Issue #2902, opened 2018-08-14
Dist versions do not work with Browserify


m90:
This is picking up: https://github.com/plotly/react-plotly.js/issues/93

I'm running into issues using the `dist` versions included in `plotly.js` and `plotly.js-dist` in a Browserify bundle.

I am basically just doing what the README tells me:

```js
var Plotly = require('plotly.js-dist');

console.log(Plotly);
```

When I try to Browserify this script it tells me:

```
Error: Cannot find module '../src/transforms/aggregate' from '/home/frederik/projects/plotly-browserify-issue/node_modules/plotly.js-dist'
```

When I try to load the dist bundle from the `plotly.js` package as seen for example [in `react-plotly.js`]( https://github.com/plotly/react-plotly.js/blob/6fe45a76d1542419fdbb9379941a631706bcf0a8/src/react-plotly.js#L2) I will see the following error:

```
Error: Cannot find module './core' from '/home/frederik/projects/plotly-browserify-issue/node_modules/plotly.js/dist'
```

I can however bundle plotly if I use `require('plotly.js')` as the main entry point.

I created an example repository for reproducing the issue over here: https://github.com/m90/plotly-browserify-issue

m90:
It looks like this is currently expected and would need https://github.com/browserify/detective/pull/79 to land in order to work.

Other related issues:
- https://github.com/browserify/browserify/issues/1773
- https://github.com/browserify/browserify/issues/1847

which probably means that `dist` is not what Browserify users (or libraries that depend on plotly) should be using.

etpinard:
Thanks very much for the detailed report. I was unaware that browserify couldn't bundle bundles it already bundled. 

It looks like https://github.com/browserify/detective/pull/79 is planned for browserify v17, so that's probably the best solution going forward.

At present though, if someone could give [`derequire`](https://www.npmjs.com/package/derequire) a shot, we will gladly merge that PR, but no plotly.js team member will dedicate time on this issue. That's mainly because bundling the "main" plotly.js package with browserify should [_just work_](https://github.com/plotly/plotly.js/blob/master/BUILDING.md#browserify) - that what we use internally afterall. In other words, for browserify users, `plotly.js-dist` only helps them save a few bytes of downloaded npm packages.

m90:
> That's mainly because bundling the "main" plotly.js package with browserify should just work - that what we use internally afterall.

[...]

PR #2905, opened 2018-08-15
Apply derequire to generated bundles

m90:
Rewrite require calls in the generated bundles so that Browserify is able
to consume the bundles when bundling new bundles. Solves issue #2902.

I can successfully use the bundle this builds in Browserify as well as Webpack (CRA). That being said I am not a Webpack Pro, so if someone else could double check this it'd be much appreciated.

### DIFF
--- a/tasks/util/browserify_wrapper.js
+++ b/tasks/util/browserify_wrapper.js
@@ -3,6 +3,8 @@ var path = require('path');
 
 var browserify = require('browserify');
 var minify = require('minify-stream');
+var derequire = require('derequire');
+var through = require('through2');
 
 var constants = require('./constants');
 var compressAttributes = require('./compress_attributes');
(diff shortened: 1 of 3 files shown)


tasks/util/browserify_wrapper.js
```diff
@@ -80,3 +84,14 @@ function logger(pathToOutput) {
     var log = 'ok ' + path.basename(pathToOutput);
     console.log(log);
 }
+
+function applyDerequire() {
```
etpinard: Ah nice, you turned the plugin into a transforms for speed. Love it!

etpinard:
Thanks very much @m90 :tada: :beers: :dancer: 

I can confirm that bundling using browserify and webpack@3  the dist bundle from `npm run build` works!

Merging.

PR #2906, opened 2018-08-15
Bump dev deps that require sshpk

etpinard:
which should solve

image

by making 

image

turn into 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -186,15 +186,15 @@
       }
     },
     "abab": {
-      "version": "1.0.4",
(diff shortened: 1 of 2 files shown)


alexcjohnson:
LGTM. There's a `karma@3.0.0` now, which doesn't look at all risky to update to, as the only breaking change is [dropping node4](https://github.com/karma-runner/karma/releases) but also doesn't seem to matter to us...
💃

etpinard:
> There's a karma@3.0.0 now,

Yeah, good call. I'll make another bump dev dep PR soon after 1.40.0 is out.

PR #2910, opened 2018-08-16
Use new multistop gradients for continuous colorbars

### LINKED ISSUE #970: colorbar draws outside the bounds

alexcjohnson:
Avoids using lots of little rectangles for continuous colorscales, which was causing small colorbars to draw outside their boundaries due to some annoying and probably still wrong logic where we expand them a little in order to ensure overlaps so we don't see gaps due to antialiasing, so now I think we really can close #970. And yes github, I mean it this time.

I also fixed the original problem of slightly overflowing colorbars, but now that we're not using this for continuous colorscales it would only arise if you were to create a contour plot with very dense contours, particularly if you don't show the contour *lines* at all, only fills. Anyway there are other benefits to using a gradient:
- better performance (we have one node with a gradient fill instead of 253 slightly-overlapping solid-fill nodes)
- more accurate (the rounding and expansion was shifting the colors a little bit - this is part of why some baseline images changed, though some had the overflow issue)
- in principle supports colorscales with opacity, though I haven't tested this, since we're not overlapping. That said we still wouldn't be able to do contour colorscales with opacity and no line, as that case still relies on overlaps. However I think in principle we could create a gradient for that case as well (using two very close stops to abruptly shift from one color to the next) in which case even that one could support opacity.

cc @etpinard


etpinard:
Nice PR :dancer: 

Good to know also that SVG `<linearGradient>` are supported all the way to IE9

image

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient

... so this shouldn't break anyone's graphs.

PR #2911, opened 2018-08-16
Add support for encoding TypedArrays as primitive objects for serialization

jonmmease:
## Overview
This PR implements a proposed approach encoding TypedArrays as primitive representation objects.  See some related discussion in https://github.com/plotly/plotly.js/issues/1784.

## Background
Plotly.js gained native support for typed arrays in https://github.com/plotly/plotly.js/pull/2388. This provides significantly improved performance when working with large arrays.  plotly.py version 3 takes advantage of typed array support by converting numpy arrays into binary buffers on the Python side, and then converting these buffers directly into TypedArrays on the JavaScript side (See https://github.com/plotly/plotly.js/pull/2388#issuecomment-367377411 for more info).

One downside of working with TypedArrays is that there isn't a standard way (at least that I've been able to find) to serialize them to JSON. This PR aims to provide a few options of remedies to this problem.

## Use cases
There are at least 5 use cases directly relevant to plotly.py where a serialized representation of TypedArrays will be very useful.

 1. ~The reason I'm working on this today is because the lack of serialization support for TypedArrays is the reason that `FigureWidget` instances containing numpy arrays cannot be rendered statically using nbconvert, nbviewer, and by extension Plotly Cloud.  With these changes I could update the JavaScript model for `FigureWidget` to not include TypedArrays, but instead primitive representation objects that can be serialized.~ (This PR is no longer needed for this use case)

 2. With these changes I could update the plotly.py JSON serializer to encode numpy arrays as base64 strings, which can be up to 10 times faster than the current method of first converting them to lists.

 3. This JSON representation can be written to disk and then opened in the JupyterLab Chart Editor more efficiently.

 4. This JSON representation should make the plotly.py orca integration more responsive for figures with large numpy arrays.

 5. This JSON representation should make Dash more responsive when working with figures with large numpy arrays.

## Typed Array Representation
This PR introduces the concept of an encoded `TypedArray`.  An encoded TypedArray is a vanilla JavaScript object that contains `dtype` and `value` properties.
 - The `dtype` property is a string indicating the data type of the `TypedArray` (`'int8'`, `'float32'`, `'uint16'`, etc.).
 - The `value` property is a primitive JavaScript object that stores the typed array data. It can be one of the following:
   i. Standard JavaScript `Array`
   ii. A base64 encoded `string`
   iii. An `ArrayBuffer` object
   iv. A `DataView` object

Encodings (i) and (ii) can be directly serialized to a string representation. Encodings (iii) and (iv) are useful when working with frameworks that already have support for serializing these more primitive binary representations.

## Decoding and Encoding
 - [x] A new top-level `Plotly.decode` function is introduced.  This function inputs a JavaScript value, and returns a copy where all encoded `TypedArray` instances have been decoded into proper `TypedArrays`. 

 - [x] A new top-level `Plotly.encode` function is introduced. This function inputs a JavaScript value  and returns a copy where all `TypedArray` instances are encoded as base64-encoded typed array representations.

## Future
 - To support multi-dimensional arrays, an encoded typed array representation object could optionally include a `shape` parameter, indicating the size of each dimension.  Plotly.js does not currently support a homogenous multi-dimensional array type, so initially these would be decoded into nested primitive arrays.

 - Would it be possible to encode datetime arrays more efficiently with a base64 buffer?


chriddyp:
Really appreciate this thorough write ups @jonmmease ! Technologically, I'll leave it up to the rest, but I do really like the sound of the use cases that you've enumerated 👍 

> With these changes I could update the plotly.py JSON serializer to encode numpy arrays as base64 strings, which can be up to 10 times faster than the current method of first converting them to lists.

Dash directly uses this JSON serializer (`plotly.utils.PlotlyJSONEncoder`), so this would be a nice win.

When you say 10 times faster, in which ways is it faster? I'm assuming there are 4:
1 - Speed to serialization (`json.dumps`)
2 - Size for network transfer (this is orders of magnitude more significant than 1, especially when Dash apps are deployed on remote servers. In Dash, we also gzip all of the requests and responses, so I'm not sure what impact this would have.)
3 - Speed of de-serialization (`JSON.parse`)
4 - Speed of charting

jonmmease:
@chriddyp The 10X was my off-the-cuff test of encoding a 1 million element numpy array of random float64 values into a Python string.

Something like `json.dumps(arr.tolist())` vs `base64.encode(memoryview(arr))).decode()`

Comparisons of (2) (3) and (4) would definitely be interesting as well!

src/lib/is_array.js
```diff
@@ -38,8 +38,15 @@ function isArray1D(a) {
     return !isArrayOrTypedArray(a[0]);
 }
 
+function isPrimitiveTypedArrayRepr(a) {
+    return (a !== undefined && a !== null &&
```
etpinard: `Lib.isPlainObject` should get rid of the first three checks.

package.json
```diff
@@ -59,6 +59,7 @@
     "@plotly/d3-sankey": "^0.5.0",
     "alpha-shape": "^1.0.0",
     "array-range": "^1.0.1",
+    "base64-arraybuffer": "^0.1.5",
```
etpinard: https://github.com/niklasvh/base64-arraybuffer/blob/master/lib/base64-arraybuffer.js seems like a legit and fast implmentation. Nice find!

Now, we'll have to check-in the associated `package-lock.json` change, to get CI to build properly.

test/image/mocks/typed_array_repr_scatter.json
```diff
@@ -0,0 +1,13 @@
+{
+  "data": [{
+    "type": "scatter",
+    "x":  {"dtype": "float64", "data": [3, 2, 1]},
```
etpinard: Not sure about using `"data"`, `"data"` has a pretty important meaning already for plotly.js. I'd vote for `values`.

etpinard: or maybe just `"v"` as calling a base64 string a set of _values_ sounds wrong.

src/lib/coerce.js
```diff
@@ -521,3 +539,48 @@ function validate(value, opts) {
     return out !== failed;
 }
 exports.validate = validate;
+
+var dtypeStringToTypedarrayType = {
```
etpinard: I think that all of them, except for `Uint8ClampedArray`:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays#Typed_array_views

Might as well add it here for completeness.

src/lib/coerce.js
```diff
@@ -521,3 +539,48 @@ function validate(value, opts) {
...
+};
+
+/**
+ * Convert a primitive TypedArray representation object into a TypedArray
+ * @param {object} v: Object with `dtype` and `data` properties that
+ * represens a TypedArray.
+ *
+ * @returns {TypedArray}
+ */
+function primitiveTypedArrayReprToTypedArray(v) {
```
etpinard: I'm curious. Have you benchmarked this routine here for bas64 string corresponding to 1e4, 1e5, and 1e6 pts?

src/lib/coerce.js
```diff
@@ -33,8 +37,18 @@ exports.valObjectMeta = {
         otherOpts: ['dflt'],
         coerceFunction: function(v, propOut, dflt) {
             // TODO maybe `v: {type: 'float32', vals: [/* ... */]}` also
-            if(isArrayOrTypedArray(v)) propOut.set(v);
-            else if(dflt !== undefined) propOut.set(dflt);
+            if(isArrayOrTypedArray(v)) {
+                propOut.set(v);
+            } else if(isPrimitiveTypedArrayRepr(v)) {
```
etpinard: Ideally, the `{dtype: '', data: ''}` -> typed array conversion should happen during the `calc` step. More precisely somewhere here:

https://github.com/plotly/plotly.js/blob/24a0f91a6fcbe6dfb7fbb68feecb6257c0bb32ba/src/plots/polar/set_convert.js#L103-L136

Depending on how slow this conversion can be, moving it to the calc step will help ensure faster interactions (note that the `calc` is skipped on e.g. zoom and pan). 

This might be a fairly big job though, you'll have to replace all the `isArrayOrTypedArray` calls upstream of the calc step with something like `isArrayOrTypedArrayOrIsPrimitiveTypedArrayRepr` (or something less verbose :smile: ).

So I guess, we should first benchmark `primitiveTypedArrayReprToTypedArray` what kind of potential perf gain we could get.

jonmmease: @etpinard  Maybe I'm misunderstanding something. As it is now, I thought these conversions would be happening in the `supplyDefaults` logic for the various traces. Does that logic run on events like zoom/pan?

Either way, I'll get some performance numbers for `primitiveTypedArrayReprToTypedArray`. I was hoping I wouldn't need to retrace all of the steps you went through to add the original `TypedArray` support!

alexcjohnson:
Thanks for bringing this issue to the fore @jonmmease - your format proposal looks great, and I'd be very happy to have this format locked in as an official part of plotly.js. My main concern is where it fits in the pipeline. Leaving the representation in the figure and making conversion part of `coerce` means:
- It will be repeated on every `supplyDefaults` - ie on every change to the plot, which is a lot of overhead
- Potentially a problem for `Plotly.react` as it would create arrays with new identity each time, making it look like the data changed when it didn't.
- We don't have an actual array in the figure object, so calls like this (which you likely don't use, but they are supported) won't work:
```js
Plotly.newPlot(gd,[{y:new Int8Array([1,2,3,4,5])}]);
Plotly.restyle(gd, 'y[2]', 6);
```

I guess the first two could potentially be fixed by @etpinard's suggestion of moving the conversion to `calc`, but it wouldn't affect the third one. I guess it would be OK in principle to just say we don't support this kind of mutation on the new data type, but doesn't seem ideal.

Alternatively, would it be reasonable to have official serialization/deserialization routes, that can be used both on a complete figure and on arguments to `restyle` et al, so we can keep `gd.data` in native format? Does that cause a problem for any of the use cases you enumerated above?

jonmmease:
I'll look it over some more tonight, but one quick thought.  For plotly.py all I really need is a way to check equality between my data model (representation array), and whatever Plotly.js stores in `data`. If it would be better on the Plotly.js side to store the converted TypedArray in `data`, then I could perform the conversion from representation to TypedArray in my equality checks. If the `isPrimitiveTypedArrayRepr` and `primitiveTypedArrayReprToTypedArray` were public so that I could import and use them from `FigureWidget` then I think that's all I would need.

If we go this route, could the `supplyDefaults` logic modify the input container (from representation to TypedArray) as well as the output? Or is that frowned upon 🙂 Is there currently a better place to do that?

alexcjohnson:
> If we go this route, could the `supplyDefaults` logic modify the input container (from representation to TypedArray) as well as the output? Or is that frowned upon 🙂 Is there currently a better place to do that?

Yeah, we do modify the input in a few places, but we're trying to break that habit. This is why I'm angling for an explicit deserialize step when new data comes in (and an explicit serialize step when saving). And as the serial format is really all about interfacing with the world outside of javascript, it seems like it should be kept separate from the regular pipeline, to be invoked by whatever application it is that's doing that out-of-js interfacing.

> If the `isPrimitiveTypedArrayRepr` and `primitiveTypedArrayReprToTypedArray` were public so that I could import and use them from FigureWidget then I think that's all I would need.

OK great, lets see how far we can get that way!

jonmmease:
Quick update. It turns out that I was able to solve the FigureWidget serialization problem by customizing the ipywidgets serialization logic. So my use-case (1) is no longer dependent on any changes here. And there's no need to make the methods discussed above public.

Given that, I do think it makes sense to work towards some form of dedicated serialization pathway. 

Maybe something like...
```javascript
// inVal is something from the outside with typed array representation objects
var inVal = {...} 

// Plotly.import converts these to TypedArrays
Plot.newPlot(gd, Plotly.import(inVal)) 

// Do stuff

// outVal has TypedArrays encoded as base64 representation objects
var outVal = Plotly.export(gd, {typedArrayRepr: 'base64'}) 
```
what do you think?

src/lib/coerce.js
jonmmease: Added in 5030d3a

package.json
jonmmease: Done in 5030d3a

src/lib/is_array.js
jonmmease: Done in 5030d3a

test/image/mocks/typed_array_repr_scatter.json
jonmmease: I renamed `data` to `value` (singular) in 5030d3a. Does that work for you? `v` felt too short 🙂

jonmmease:
Latest push reverts all coerce.js changes and moves the typed array conversion logic to a new `Plotly.import` function.

src/lib/is_array.js
```diff
@@ -5,6 +5,7 @@
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
+var Lib = require('../lib');
```
etpinard: You'll need to require `'./is_plain_object.js` to avoid a circular dependency pattern

lib/index -> lib/is_plain_object -> lib/index -> lib/is_array -> lib/index -> lib/is_plain_object

etpinard:
@jonmmease your

```js
// inVal is something from the outside with typed array representation objects
var inVal = {...} 

// Plotly.import converts these to TypedArrays
Plot.newPlot(gd, Plotly.import(inVal)) 

// Do stuff

// outVal has TypedArrays encoded as base64 representation objects
var outVal = Plotly.export(gd, {typedArrayRepr: 'base64'}) 
```

sounds solid. :1st_place_medal: 

Personally, I find `Plotly.import` a little confusion. At first-order, I think of `import` as _importing as module_, not a dataset. Perhaps `Plotly.decodeTypedArrays` would be better?

src/plot_api/plot_api.js
```diff
@@ -2156,6 +2157,82 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
 };
 
+/**
+ * Convert a primitive TypedArray representation object into a TypedArray
+ * @param {object} v: Object with `dtype` and `data` properties that
+ * represens a TypedArray.
+ *
+ * @returns {TypedArray}
+ */
+var dtypeStringToTypedarrayType = {
```
etpinard: We'll need to append this list:

https://github.com/plotly/plotly.js/blob/a8c621706326dafbfe385418874a54c4723cbc4e/.eslintrc#L12-L21

to make `npm run lint` pass.

jonmmease:
Relevant question from the plotly.py forums: https://community.plot.ly/t/offline-plot-to-div-encode-numpy-data-as-binary-blob/12965

>Is there any way to use plotly.offline to store (numpy) data in a div or HTML as a binary blob then have plotly js render it properly?
>
> The HTML file has bits that look like…
“z”: [[0.2866461675771665, 0.36531080671829425, -0.2904632669675007, -0.36149370732795966, -0.049367492181059396, -0.08192231566757988, 0.6849181745604624, 1.268579555423972, 0.2944241042103686, -0.2851544876150079, -0.9762164627387184, 0.1527361407626408] …
>
> I’m trying to reduce the size of my HTML files that have several image plots each with 10^5 - 10^6 data points.
>
> I’ve taken to stripping some of the precision so the json (string)-encoded data have fewer digits, but this seems silly.

This reminds me that the encoding should be able to handle the 2-dimensional use case as well.

I think I'll add an optional `shape` field to the encoding. If `shape` is absent or a scalar array then the representation will convert into a TypedArray. But if the shape is an array with more than one element, then the representation will be converted into a nested set of arrays.

jonmmease:
Thanks for the feedback @etpinard , I'm going to circle back around to this in a week or two, after the plotly.py 3.2 release.

In terms of naming. I was thinking of keeping "TypedArray" out of the name so that we could eventually add other encodings if useful.  How about `Plotly.import` -> `Plotly.decode` and `Plotly.export` -> `Plotly.encode`?

`Plotly.decode` would be generous about accepting the various encodings of `TypedArrays` (and maybe eventually various encoding of other types like images). `Plotly.encode` would always produce a JSON-serializable representation of the object.

I was also wondering if it would make sense to allow `Plotly.decode` to accept compressed representations. It would be nice, from the Python side, to be able to encode arrays as base64 and then gzip up the whole figure bundle before transporting to Plotly.js. It think Dash already does this, but it would be nice if were baked in to plotly.py and Plotly.js. 

@chriddyp would compression as the figure-level be useful to Dash or do you already compress at a higher level?  Also, what do you use from decompression on the JavaScript side?

etpinard:
> How about Plotly.import -> Plotly.decode and Plotly.export -> Plotly.encode?

:ok_hand: 



> I was also wondering if it would make sense to allow Plotly.decode to accept compressed representations.
> to be able to encode arrays as base64 and then gzip up the whole figure bundle before transporting to Plotly.js.

Interesting, but perhaps this is out of the scope of plotly.js? How big our common front-end unzip libraries? If we want to have all decode/encode/compress/decompress logic in one place, maybe we should explore placing these `Plotly.decode` / `Plotly.encode` methods in a separate npm package.

jonmmease:
Good point regarding scope.  The typed array stuff is pretty tied to Plotly.js, but the compression can happen where-ever.

I was picturing using the decompression from orca eventually, but it probably makes more sense to just add this as an option to orca down the road (if it proves helpful), rather than Plotly.js.

jonmmease:
Ok, I believe I have finished the implementation and testing of the new `Plotly.decode` and `Plotly.encode` functions for 1-dimensional TypedArray encoding.

Handling multi-dimensional arrays will be a bit more complicated, and a bit less useful since Plotly.js doesn't (yet?) use a homogenous multi-dimensional array type internally, so I'd like to put this off for a future PR.

Might it be feasible to get this into 1.41?  My hope was to use this in plotly.py 3.3, where I'm going to introduce new `plotly.io.write_figure` and `plotly.io.read_figure` functions for reading and writing figures to/from disk. Then, we'll update the JupyterLab chart editor to read and write these same files from the file system (https://github.com/plotly/jupyterlab-chart-editor/issues/20).  With compression on top, I think we'll have a really efficient storage format for figure's involving large arrays!

etpinard:
@jonmmease after a long talk with @alexcjohnson, we came to a few conclusions:

- a JSON-to-typed-array encode/decode util will be very useful not just for plotly.py but also for plotly cloud and the react-chart-editor
- decoding works best when called **before** `Plotly.newPlot`, mainly when considering applications like the react-chart-editor that would need to decode before populating its table columns.
- when called before `Plotly.newPlot` the decoding util doesn't need to know anything about plotly.js, except that it uses JSON-compatible objects as arguments.
- the only short-term consumer of these encode/decode util will be plotly.py. Adding typed-array support on plotly cloud will be a big project. Adding it to the react-chart-editor doesn't appear to be a priority.
- and finally, we doubt that `Plotly.encode` and `Plotly.decode` are valid additions to plotly.js (not plotly in general :smile: ) as we try to keep plotly.js' scope only as wide as it _needs_ to be.

So, I'm proposing two solutions:

- make a npm package called `@plotly/typed-json` and add it as a dependency to plotly.py. I would be happy to dedicate some time to this later on this week after 1.41.0 is released
- or, move the new encode/decode to the plotly.py front-end for now, until we have a better idea of where this encode/decode util fits in the plotly realm.

jonmmease:
Thanks for taking the time to think through this in detail @etpinard and @alexcjohnson.  The approach of eventually creating a separate `@plotly/typed-json` npm package makes fine sense to me.

plotly.py does't really need this new npm package for itself (it already as a version of the `decode` function in the implementation of the `FigureWidget` JavaScript library). The benefit for plotly.py users will be when other Plotly.js-based endpoints are able to accept (and return) encoded typed arrays.  The ones in the front of my mind are orca, the JupyterLab chart editor, the dash figure component, and cloud.

So I'm happy to plan on pursuing the npm package approach in not too distant future, but I don't think it needs to be this week.  It's not blocking anything, just the next step in efficiency for large datasets. Maybe after 1.42? Thanks!

test/jasmine/tests/decode_typed_arrays_test.js
```diff
@@ -0,0 +1,160 @@
...
+var typedArraySpecs = [
+    ['int8', new Int8Array([-128, -34, 1, 127])],
+    ['uint8', new Uint8Array([0, 1, 127, 255])],
+    ['uint8_clamped', new Uint8ClampedArray([0, 1, 127, 255])],
+    ['int16', new Int16Array([-32768, -123, 345, 32767])],
+    ['uint16', new Uint16Array([0, 345, 32767, 65535])],
+    ['int32', new Int32Array([-2147483648, -123, 345, 32767, 2147483647])],
+    ['uint32', new Uint32Array([0, 345, 32767, 4294967295])],
+    ['float32', new Float32Array([1.2E-38, -2345.25, 2.7182818, 3.1415926, 2, 3.4E38])],
+    ['float64', new Float64Array([5.0E-324, 2.718281828459045, 3.141592653589793, 1.8E308])]
```
archmoj, line 26: It might be of interest to add `BigInt`s to the list.

src/plot_api/plot_api.js
```diff
@@ -2192,6 +2193,181 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+        return 'uint16';
+    } else if(typeof Int32Array !== 'undefined' && v instanceof Int32Array) {
+        return 'int32';
+    } else if(typeof Uint32Array !== 'undefined' && v instanceof Uint32Array) {
+        return 'uint32';
+    } else if(typeof Float32Array !== 'undefined' && v instanceof Float32Array) {
+        return 'float32';
+    } else if(typeof Float64Array !== 'undefined' && v instanceof Float64Array) {
+        return 'float64';
+    }
```
archmoj, line 2317: It might be of interest to support `BigInt`s as well.

src/plot_api/index.js
```diff
@@ -16,6 +16,8 @@ exports.restyle = main.restyle;
 exports.relayout = main.relayout;
 exports.redraw = main.redraw;
 exports.update = main.update;
+exports.decode = main.decode;
+exports.encode = main.encode;
```
archmoj, line 20: We may start with underscore (i.e. kind of private) methods here.
For example:
```js
exports._decode = main.decode;
exports._encode = main.encode;
```
If we want to expose these functionality, then using a more specific name could be considered.
For example: 
```js
exports.decodeArray = main.decode;
exports.encodeArray = main.encode;
```

src/plot_api/plot_api.js
```diff
@@ -2192,6 +2193,181 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+ */
+function performEncode(v) {
+    if(Lib.isTypedArray(v)) {
+        return encodeTypedArray(v);
+    } else if(Array.isArray(v)) {
+        return v.map(performEncode);
+    } else if(Lib.isPlainObject(v)) {
+        var result = {};
+        for(var k in v) {
+            if(v.hasOwnProperty(k)) {
```
archmoj, line 2349: Here we may list the keys using [Object.getOwnPropertyNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames) and loop through them.

archmoj:
> One downside of working with TypedArrays is that there isn't a standard way (at least that I've been able to find) to serialize them to JSON. This PR aims to provide a few options of remedies to this problem.
> 
What's the status of this problem in 2020?

archmoj:
> > One downside of working with TypedArrays is that there isn't a standard way (at least that I've been able to find) to serialize them to JSON. This PR aims to provide a few options of remedies to this problem.
> 
> What's the status of this problem in 2020?

Besides there is `BSON` file format: https://en.wikipedia.org/wiki/BSON

src/plot_api/plot_api.js
archmoj: Big integers could be useful for storing time in milliseconds.

archmoj:
> Plotly.js does not currently support a homogenous multi-dimensional array type, so initially these would be decoded into nested primitive arrays.

`ndarray` is applied in the `surface` trace code:
https://github.com/plotly/plotly.js/blob/a54c5028e814760b95e6243b07d7c21151166b82/src/traces/surface/convert.js#L14-L15

src/plot_api/plot_api.js
```diff
@@ -2192,6 +2193,181 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+    } else if(dtype === 'int32' && typeof Int32Array !== 'undefined') {
+        return Int32Array;
+    } else if(dtype === 'uint32' && typeof Uint32Array !== 'undefined') {
+        return Uint32Array;
+    } else if(dtype === 'float32' && typeof Float32Array !== 'undefined') {
+        return Float32Array;
+    } else if(dtype === 'float64' && typeof Float64Array !== 'undefined') {
+        return Float64Array;
+    }
+}
```
archmoj, line 2222: You may consider rewrite this:
```js
function getArrayType(t) {
    return (typeof t !== 'undefined') ? t : undefined;
}
var validInt8Array = getArrayType(Int8Array);
var validUint8Array = getArrayType(Uint8Array);
...

/**
 * Get TypedArray type for a given dtype string
 * @param {String} dtype: Data type string
 * @returns {TypedArray}
 */
function getTypedArrayTypeForDtypeString(dtype) {
    switch(dtyle) {
        case 'int8':
            return validInt8Array;
        case 'uint8':
            return validUint8Array;
        ...
    }
}
```

src/plot_api/plot_api.js
```diff
@@ -2192,6 +2193,181 @@ exports.update = function update(gd, traceUpdate, layoutUpdate, _traces) {
...
+        var result = {};
+        for(var k in v) {
+            if(v.hasOwnProperty(k)) {
+                result[k] = performDecode(v[k]);
+            }
+        }
+        return result;
+    } else {
+        return v;
+    }
```
archmoj, line 2278: You can drop last `else` statement and simply `return v;`

archmoj:
Right now, one could call:
```js
Plotly.newPlot(gd, Plotly.decode({
    'data': [{
        'type': 'scatter',
        'x': {'dtype': 'float64', 'value': 'AAAAAAAACEAAAAAAAAAAQAAAAAAAAPA/'},
        'y': {'dtype': 'float32', 'value': 'AABAQAAAAEAAAIA/'},
        'marker': {
            'color': {
                'dtype': 'uint16',
                'value': 'AwACAAEA',
            },
        }
    }]
}));
```
But wondering if we could/should expand `data_array` valType to support a call like this instead? i.e. without exposing `decode` function.
```js
Plotly.newPlot(gd, 
    'data': [{
        'type': 'scatter',
        'x': {'dtype': 'float64', 'value': 'AAAAAAAACEAAAAAAAAAAQAAAAAAAAPA/'},
        'y': {'dtype': 'float32', 'value': 'AABAQAAAAEAAAIA/'},
        'marker': {
            'color': {
                'dtype': 'uint16',
                'value': 'AwACAAEA',
            },
        }
    }]
});
```

jonmmease:
That was my original design.  I don't remember all of the details, but as I recall it got a little messy to work through what should get stored in `data` and `fullData` (whether to store the input data structure, or the converted Typed Array) to handle restyle/react updates.

I don't have a strong preference if there's a clean way to handle it all internally.

archmoj:
> That was my original design. I don't remember all of the details, but as I recall it got a little messy to work through what should get stored in `data` and `fullData` (whether to store the input data structure, or the converted Typed Array) to handle restyle/react updates.
> 
> I don't have a strong preference if there's a clean way to handle it all internally.

We actually have this comment here:
https://github.com/plotly/plotly.js/blob/a54c5028e814760b95e6243b07d7c21151166b82/src/lib/coerce.js#L34

test/jasmine/tests/decode_typed_arrays_test.js
ejoliet: Seems that is not yet supporting longs. How is ploty dealing with BigInt currently? is there any support planned?

astroboylrx:
How may one use this feature when exporting an interactive plot to HTML from Python? The html file I got still store data in ascii format. I couldn't find proper documentation for using this.

Issue #2893, opened 2018-08-10
Fractional milliseconds are rounded to 0.1ms precision


SemmZemm:
The fractional milliseconds are rounded to 0.1ms precision when being provided as epoch ms. 
https://jsfiddle.net/c1kt3r82/766/

Couldn't you make behave it like v1.20.5? 
https://jsfiddle.net/c1kt3r82/767/

alexcjohnson:
Heh, recently epoch ms had coarser rounding than strings, until #2847 brought the two into harmony.
3 things to say about this:
- That v1.20.5 behavior is quite fragile, as you can see if you decrease the overall range a bit, among other things. I was unaware that it even worked to the degree that you demonstrate in your fiddle though, so thanks for pointing it out!
- At least until v2 where we intend to alter how epoch ms work (given their strange timezone behavior), we do not intend to allow them any better features than strings. Our official stance is that they are tolerated but discouraged in v1.
- We can certainly have a discussion about how to handle *all* dates though - 0.1ms precision was chosen to work consistently across the whole supported date range, but that's thousands of years and numbers closer to the unix epoch can do better than this. Within a reasonable range of modern dates we should be able to increase the resolution, looks like 1µs is not out of the question in fact for the current time, though it will take some investigation to see how far into the future (and past) we can allow this. And extending this adaptive handling throughout the rest of our date handling machinery will take a fair bit of effort and a lot of testing.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #2912, opened 2018-08-16
fix #2892 - don't allow auto date ticks < our 100 microseconds limit

### LINKED ISSUE #2892: Fractional milliseconds issue
If I use x=[1486549278145, 1486549278145.02, 1486549278145.04, 1486549278145.06, 1486549278145.08, 1486549278145.1] the graph doesn't display anything at all

https://jsfiddle.net/c1kt3r82/763/

alexcjohnson:
At least until we do something more substantial in #2893, be more consistent about the 100 microsecond limit.

cc @etpinard


alexcjohnson:
oh hmm there's another way to trigger this same problem...

etpinard:
Great. One more for 1.40.0 :dancer:

Issue #1697, opened 2017-05-17
Find better solution to "TOBESTRIPPED" dummy string in toSVG to handle quoted style attributes


etpinard:
See https://codepen.io/etpinard/pen/vmzpPy?editors=1010 and https://github.com/plotly/plotly.js/pull/1694#pullrequestreview-38774918

etpinard:
Suggestions from @dy, use:

- https://www.npmjs.com/package/split-string-words, or
- maybe https://www.npmjs.com/package/balanced-match

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #2355, opened 2018-02-09
Plotly uses inline CSS


sanmai-NL:
Which necessitates the `unsafe-inline` Content Security Policy directive, or another workaround. Can we get rid of the inline styling somehow?

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution.

l.addStyleRule @ plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7
plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7 Uncaught TypeError: Cannot read property 'insertRule' of null
    at Object.l.addStyleRule (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
    at Object.1.../src/lib (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
    at i (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
    at plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7
    at Object.724.../build/plotcss (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
    at i (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
    at plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7
    at Object.12.../src/core (plotly__SHA256:3b219dbb9a6101432f4c65d4d4f35a0c5bff2ed2b9d43a007d46faa937b2ac81.mjs:7)
[...]

PR #2914, opened 2018-08-16
Fix `toSVG` for graphs that call `Drawing.gradient`

etpinard:
This PR fixes `toSVG` (and hence `toImage` and `downloadImage`) for graphs with filled colorbars (from https://github.com/plotly/plotly.js/pull/2910) and contour legend items (added in https://github.com/plotly/plotly.js/pull/2891) in "newer" browsers.

These bugs are related to double-nested `"` inside svg node `style` - which browsers have a hard time serializing (cc https://github.com/plotly/plotly.js/issues/1697). Why were they unnoticed in their respective PRs: our old nw.js image-server doesn't care about those double-nested `"`. 

What saved us from releasing these bugs in 1.40.0 is `./tasks/noci_test.sh` which test mapbox and cone mocks using orca (which relies on a much newer version of Chromium) where the mocks with colorbars failed to export.

Now, this PR offers two solutions:

- in https://github.com/plotly/plotly.js/commit/b32aac06ce2033224f76931e216f41a609f1b286, which implements a solution very similar to https://github.com/plotly/plotly.js/pull/1928
- another in https://github.com/plotly/plotly.js/commit/1c8017ef88c3a04d1bbc555b72a4a42ca0848847, where we take a different approach: setting gradient URLs as node attribute, not node style as @alexcjohnson suggested in a private convo.

The second solution seems to work well, and it leads to :hocho: a fairly big hacky block in `toSVG`. Thoughts?

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -334,11 +334,7 @@ drawing.gradient = function(sel, gd, gradientID, type, colorscale, prop) {
             });
         });
 
-    // Set url to gradient definition as an 'attribute', so that
-    // we don't have to fiddle with quotes inside of quotes in Snapshot.toSVG.
-    // Clear corresponding 'style' setting to avoid potential conflicts.
-    sel.attr(prop, 'url(#' + fullID + ')')
-        .style(prop, null)
+    sel.style(prop, 'url(#' + fullID + ')')
         .style(prop + '-opacity', null);
 };
 
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -334,7 +334,11 @@ drawing.gradient = function(sel, gd, gradientID, type, colorscale, prop) {
             });
         });
 
-    sel.style(prop, 'url(#' + fullID + ')')
+    // Set url to gradient definition as an 'attribute', so that
+    // we don't have to fiddle with quotes inside of quotes in Snapshot.toSVG.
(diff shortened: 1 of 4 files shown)


alexcjohnson:
The first solution (extending the `toSVG` patch) is certainly the smaller conceptual change. Switching to attributes is certainly appealing though - and we have at least one issue https://github.com/plotly/plotly.js/issues/2355 encouraging us to switch whole hog. My main concern about this is how easy it is to (accidentally) override attributes with css. I worry that we will
- get lots of users complaining that plotly.js doesn't work in their page/app because other css (which they may or may not control but anyway are not expecting will impact plots) is interfering with the plot.
- get users intentionally overriding plot styling with css, thereby breaking portability of these plots

So I guess in the end I'd vote for the `toSVG` patch, at least until we have a serious conversation / investigation about attributes.

etpinard:
> So I guess in the end I'd vote for the toSVG patch, at least until we have a serious conversation / investigation about attributes.

Sounds good. I'll revert https://github.com/plotly/plotly.js/pull/2914/commits/1c8017ef88c3a04d1bbc555b72a4a42ca0848847 so it will remain in git history in case we change our mind later.

alexcjohnson:
> I'll revert 1c8017e so it will remain in git history in case we change our mind later.

Good call. 💃

PR #2917, opened 2018-08-17
Heatmap ordering bug

### LINKED ISSUE #2907: heatmap ordering bug
Heatmaps have a similar issue to one that was fixed for contour maps in #2891 - ordering gets messed up if you hide and reshow traces. See https://codepen.io/alexcjohnson/pen/OwYrKg - initially electric is in front as it should be:
screen shot 2018-08-15 at 3 44 19 pm
But if you click "Hide & Show Trace 0" then Greys is in front:
screen shot 2018-08-15 at 3 45 41 pm
"Hide & Show Trace 1" will bring electric back to the front.

See https://github.com/plotly/plotly.js/pull/2891/commits/597a845428ce57d4020bb8d7a0bd11e9d48bde10#r209345643 - Should be just a matter of replicating https://github.com/plotly/plotly.js/pull/2891/commits/597a845428ce57d4020bb8d7a0bd11e9d48bde10 for `heatmap` - and probably for `carpet` too, though it's probably unintended anyway if carpets overlap each other!

alexcjohnson:
Fixes #2907 - dynamic ordering of heatmap and carpet traces

:hocho: the now-obsolete `getUidsFromCalcData`

But then I realized this trace group management pattern is identical for all cartesian traces, so I made a single helper `makeTraceGroups` to handle this in all types, except `scatter` since animations give it somewhat different needs.

cc @etpinard


alexcjohnson:
Huh, on my machine as well as on CI `connectgaps_2d` looks fine, but somehow GH can't display it? I don't know why there are image diffs here. `connectgaps_2d` I think I can at least see something that changed at the edges of the clip path, `contour_heatmap_coloring` I really can't even see anything.

src/traces/contour/plot.js
```diff
@@ -636,8 +636,6 @@ function clipGaps(plotGroup, plotinfo, clips, cd0, perimeter) {
     else clipId = null;
 
     plotGroup.call(Drawing.setClipUrl, clipId);
-    plotinfo.plot.selectAll('.hm' + cd0.trace.uid)
-        .call(Drawing.setClipUrl, clipId);
```
alexcjohnson: Alright, GH is working again with these images, must've been an internal glitch last night.

The reason for the image diffs is that in master (right) we were applying the `clip-path` to both the heatmap group and the contour group... and now that the heatmap is inside the contour group that was redundant but apparently it was still having an effect on antialiasing. The `Drawing.setClipUrl ` here that was supposed to be clipping the heatmap actually wasn't doing anything, since I removed the uid-based class from the heatmap.
screen shot 2018-08-17 at 9 59 51 am
So in fact the image diff in this PR probably nearly (or exactly) reverts the diff from `contour-legend` https://github.com/plotly/plotly.js/pull/2891/files?short_path=2b47456#diff-2b474566a55f847908961b61d6e0cba4

src/plots/cartesian/make_trace_groups.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports = function makeTraceGroups(gd, plotinfo, cdModule, traceLayer, cls, plotOneFn) {
+    var traces = traceLayer.selectAll('g.' + cls.replace(/\s/g, '.'))
+        .data(cdModule, function(cd) { return cd[0].trace.uid; });
+
+    traces.exit().remove();
+
+    traces.enter().append('g')
+        .attr('class', cls);
+
+    traces.each(function(cd) {
```
etpinard: Ah very nice! That `cls.replace(/s/g/, '.')` is a nice touch. 

I think `makeTraceGroups` should work with all svg-based subplot types, so perhaps we could move this to `src/lib/`? No sure where it would belong. There's [`lib/gup.js`](https://github.com/plotly/plotly.js/blob/master/src/lib/gup.js) which has "general-update-pattern" helpers, maybe that would be a good fit.

Personally, I would not have included `plotOneFn` as an argument and here just called `traces.order()`, return `traces` and make `_module.plot` handle the `trace.each` "plot-one" logic. This way we wouldn't be tied down to one `plotOneFn` call signature and we wouldn't have to pass `gd`, `plotinfo` to `makeTraceGroups`, perhaps making things more reusable. But that's really more of a personal opinion, so consider this comment non-blocking :ok_hand:

src/plots/get_data.js
```diff
@@ -124,20 +124,3 @@ exports.getSubplotData = function getSubplotData(data, type, subplotId) {
...
     return subplotData;
 };
-
-/**
- * Get a lookup object of trace uids corresponding in a given calcdata array.
- *
- * @param {array} calcdata: as in gd.calcdata (or a subset)
- * @return {object} lookup object of uids (`uid: 1`)
- */
-exports.getUidsFromCalcData = function(calcdata) {
```
etpinard: Thanks!

src/traces/contour/plot.js
```diff
@@ -636,8 +636,6 @@ function clipGaps(plotGroup, plotinfo, clips, cd0, perimeter) {
     else clipId = null;
 
     plotGroup.call(Drawing.setClipUrl, clipId);
-    plotinfo.plot.selectAll('.hm' + cd0.trace.uid)
```
etpinard: nice find!

src/plots/cartesian/make_trace_groups.js
alexcjohnson: I was trying to match the existing pattern (from `contourPlot.plotWrapper`) but yeah, taking `plotOne` out would be cleaner and even more generalizable, I'll do that. Originally I wasn't returning the final selection but now that we have that this will be easy.

src/traces/bar/plot.js
```diff
@@ -32,128 +31,128 @@ var attributes = require('./attributes'),
...
-function plotOne(gd, plotinfo, cd, plotGroup) {
     var xa = plotinfo.xaxis;
     var ya = plotinfo.yaxis;
     var fullLayout = gd._fullLayout;
-    var cd0 = cd[0];
-    var t = cd0.t;
-    var trace = cd0.trace;
 
-    if(!plotinfo.isRangePlot) cd0.node3 = plotGroup;
+    var bartraces = Lib.makeTraceGroups(barLayer, cdbar, 'trace bars').each(function(cd) {
```
etpinard: Ha yes. `Lib.makeTraceGroups` looks very similar to `Lib.ensureSingle` which have similar purpose. :ok_hand:

etpinard:
> Alright, more general pattern in 69dee75, and used it to  pie in 9ddad4c and fix choropleth e892923 and scattergeo 521d27f

Wow. That's great! Thanks for adding test to scattergeo and choropleth :lock: this down for good. I'll take care of `scatterpolar` in the `barpolar` PR.

> ( may make a merge conflict with #2918)

I'll take care of it before my fix-intermittent-gl-test-failure commit.

Once again -> :dancer:

PR #2918, opened 2018-08-17
no more fail for async tests, only failTest

alexcjohnson:
cc @etpinard


test/jasmine/tests/download_test.js
```diff
@@ -2,7 +2,7 @@ var Plotly = require('@lib/index');
 var createGraphDiv = require('../assets/create_graph_div');
 var destroyGraphDiv = require('../assets/destroy_graph_div');
 var textchartMock = require('@mocks/text_chart_arrays.json');
-var fail = require('../assets/fail_test');
+var failTest = require('../assets/fail_test');
```
alexcjohnson: FWIW there was only one actual bug caught in this change, `download_test` has one real Jasmine `fail` in it which we were overriding with our async version:
https://github.com/plotly/plotly.js/blob/2efd0c400471e41e0df93067d8053002809039dd/test/jasmine/tests/download_test.js#L90
which I guess would still show up as a test failure... it just wouldn't give the right message.

alexcjohnson:
@etpinard do you want to use this PR, since it's all about tests, to play with the streamtube test failure? Or make a new one to sort that out?

etpinard:
> do you want to use this PR, since it's all about tests, to play with the streamtube test failure?

Good idea! I'll push a few commits to this branch.

etpinard:
:dancer:

PR #2921, opened 2018-08-17
support 'e' and 'E' exponentformat on log axes

### LINKED ISSUE #2919: `exponentformat` of 'E' does not work with log scaling
When using `layout.yaxis.type = 'log'` the `layout.yaxis.exponentformat` parameter works with `SI`, `power`,  and `none`, but setting to `e` or `E` triggers the `power` behavior. 

See [example here](https://jsbin.com/vopevam)

alexcjohnson:
fixes #2919 - not sure why I thought at the time that `e` and `E` should be treated as synonymous with `power` but that doesn't really make much sense.
cc @etpinard


test/image/mocks/gl2d_12.json
```diff
@@ -734,7 +734,7 @@
                 "size": 0,
                 "color": ""
             },
-            "exponentformat": "e",
```
alexcjohnson: @etpinard I suppose since this behavior has existed so long, this could be considered a breaking change... but it also seems to me to be clearly wrong. Are you OK with calling it a bug?

test/image/mocks/14.json
```diff
@@ -2,432 +2,432 @@
...
-            "zerolinecolor": "#000", 
-            "zerolinewidth": 1, 
-            "linecolor": "#000", 
-            "linewidth": 1, 
-            "anchor": "y", 
-            "position": 0, 
-            "mirror": true, 
-            "overlaying": false, 
+            },
+            "exponentformat": "power",
```
etpinard: ah. there is it :mag:

test/image/mocks/gl2d_12.json
etpinard: Yes, I'm ok calling this a bug as `"e"` isn't the `exponentformat` default, so I can't happen "by accident".

etpinard:
Beautiful :dancer:

test/image/mocks/14.json
alexcjohnson: heh yeah, sorry - I should have split out linting those trailing spaces (we don't test that in the mocks?) into its own commit.

PR #2931, opened 2018-08-20
Fix Plotly.download using graph id as 1st argument

### LINKED ISSUE #2930: TypeError: Cannot create property '_snapshotInProgress' on string 'plot'
This is my first post on here, so apologies if I have missed something ...

I'm getting a type error when using Plotly.downloadImage(). According to the docs [here](https://plot.ly/javascript/plotlyjs-function-reference/#common-parameters):

> The first argument to each function on this page is a reference to this element, and it can be either a DOM node, i.e. the output of document.getElementById(), or a string, in which case it will be treated as the id of the div.

But looking at /src/snapshot/download.js, it looks like passing the string (the id of the div) shouldn't work as the method doesn't look like it tries to find the DOM element from the passed string.... I've created a simplified CodePen that replicates the issue: https://codepen.io/DrRochy/pen/LJYbja

So, it seems to work correctly if you pass the object, but not when you just pass the string.

What's really weird is I am sure I have passed the id of the div element and this has worked, so I am very baffled how!

etpinard:
fixes #2930

cc @alexcjohnson

### DIFF
--- a/src/snapshot/download.js
+++ b/src/snapshot/download.js
@@ -45,19 +45,19 @@ function downloadImage(gd, opts) {
             reject(new Error('Sorry IE does not support downloading from canvas. Try {format:\'svg\'} instead.'));
         }
 
-        _gd._snapshotInProgress = true;
+        if(_gd) _gd._snapshotInProgress = true;
         var promise = toImage(gd, opts);
 
         var filename = opts.filename || gd.fn || 'newplot';
         filename += '.' + opts.format;
 
         promise.then(function(result) {
-            _gd._snapshotInProgress = false;
+            if(_gd) _gd._snapshotInProgress = false;
             return fileSaver(result, filename);
         }).then(function(name) {
             resolve(name);
         }).catch(function(err) {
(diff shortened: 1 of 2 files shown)


src/snapshot/download.js
```diff
@@ -43,19 +45,19 @@ function downloadImage(gd, opts) {
             reject(new Error('Sorry IE does not support downloading from canvas. Try {format:\'svg\'} instead.'));
         }
 
-        gd._snapshotInProgress = true;
+        _gd._snapshotInProgress = true;
```
alexcjohnson: Won't this fail for the plain object case (`_gd === undefined`)?

etpinard: Good eye. This made me realize `Plotly.download` with figure objects as first argument was also broken. Fixed and :lock: ed in -> https://github.com/plotly/plotly.js/pull/2931/commits/a5af83c0275a6ae82c4b3cb9377982cc24cf268b

alexcjohnson:
💃

PR #2932, opened 2018-08-20
improve pseudo-html entity conversion

### LINKED ISSUE #2927: Some HTML entities are double-decoded.
Titles etc have a strange double-decoding bug for HTML entities. 

For example, it is impossible to include a literal  `  &lt;b&gt;   ` within a title or tooltip - it is always converted to `<b>`.   This has two consequences:

1. If I actually want a raw  angle-bracket, for example to say `Threshold < 3`, then this is brittle, because sometimes the bracket could be interpreted as beginning a tag.

2. There is a possible security risk - it is impossible to enforce an "htmlspecialchars()" conversion to make user input safe, because entities are double-decoded.

Try for example:

```
layout = {
  title: "This &lt;b&gt;could break&lt;/b&gt; &lt;script&gt;alert(\"?\")&lt;/script&gt; protection",
};
```

1. This wrongly shows the words "could break" in bold. This is a bug.

2. However, there seems to be some special-case handling of "script" because the alert does not trigger, and the script tag is shown literally.  This indicates a belt-and-braces fix in the special-case of "script", I think. So while the behaviour is safe, it's confusing, given (1).

Thanks for your time and your help.

alexcjohnson:
- Fix #2927 - Move entity conversion later in the convertToTspans pipeline, so we don't try to interpret `&lt;` as `<` and make a tag out of it
- Support all numbered entities, decimal and hex - including the correct ampersand code!
- Improve performance - instead of running the string through a separate regex replace for every entity, just run it through one regex replace that finds all entities. Note that this would make it feasible, performance-wise, to include ALL named html entities, but the full list is over 2000 items, and there are other options (unicode literals, or numbered entities if you want your code/JSON to be pure ASCII) so I suggest we keep it at just the existing set.
- Make the GL version work the same as the SVG version. Rather than fix the `fixEntities` function there, which was overly aggressive about finding entities, I just pointed it to the SVG version. It's possible that the simple loop-based function there would be faster, but making it work correctly would have slowed it down and made it more complicated, and anyway the other transformations in `html2unicode` use regex replacements so I guess this wasn't hyper-optimized anyway.

cc @etpinard


test/jasmine/tests/svg_text_utils_test.js
```diff
@@ -300,16 +300,56 @@ describe('svg+text utils', function() {
...
             );
 
-            expect(node.text()).toEqual('100μ & < 10 > 0  100 × 20 ± 0.5 °');
+            expect(node.text()).toBe('100μ & < 10 > 0  100 × 20 ± 0.5 °');
         });
 
         it('decodes some HTML entities in text (number case)', function() {
             var node = mockTextSVGElement(
-                '100&#956; &#28; &#60; 10 &#62; 0 &#160;' +
+                '100&#956; &#38; &#60; 10 &#62; 0 &#160;' +
```
alexcjohnson: No, the ampersand is not `&#28;`...

etpinard: Oops. My bad.

alexcjohnson: oh haha I assumed I did that 🙈

src/lib/svg_text_utils.js
```diff
@@ -276,19 +278,44 @@ exports.plainText = function(_str) {
...
+exports.convertEntities = convertEntities;
+
+function fromCodePoint(code) {
+    // Don't allow overflow. In Chrome this turns into � but I feel like it's
+    // more useful to just not convert it at all.
+    if(code > 0x10FFFF) return;
+    var stringFromCodePoint = String.fromCodePoint;
+    if(stringFromCodePoint) return stringFromCodePoint(code);
+
+    // IE doesn't have String.fromCodePoint
```
etpinard: To be fair, not just IE doesn't support `fromCodePoint`, 2015 Chrome and 2014 FF didn't have it either.

src/lib/html2unicode.js
```diff
@@ -10,7 +10,7 @@
 'use strict';
 
 var toSuperScript = require('superscript-text');
-var stringMappings = require('../constants/string_mappings');
+var fixEntities = require('./svg_text_utils').convertEntities;
```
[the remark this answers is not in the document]
alexcjohnson: good call -> :hocho: `string_mapping.js` in  https://github.com/plotly/plotly.js/pull/2932/commits/a943db2a7470e66baf2ebfeac0a6ca70c252d565

src/lib/svg_text_utils.js
alexcjohnson: Fair enough - though we don't support such old versions of Chrome and FF, do we?

etpinard:
Very nice PR. :1st_place_medal: 

I tried to find a case where the new `convertEntities` logic would contradict the works of https://github.com/plotly/plotly.js/pull/804, but to no avail.

I'll give you a :dancer: , but I wouldn't mind waiting a few days before merging it in case we need to patch `1.40.0`.

------------------

Now about,

> Note that this would make it feasible, performance-wise, to include ALL named html entities, but the full list is over 2000 items, and there are other options (unicode literals, or numbered entities if you want your code/JSON to be pure ASCII) so I suggest we keep it at just the existing set.

If someone really wants us to decode all named html entities, I suggest we make an optional component (similar to calendars) that would require some third-party module (e.g. [`he`](https://www.npmjs.com/package/he)) to handle the conversions.

src/lib/svg_text_utils.js
etpinard: > do we?

We used to have a blob on https://plot.ly/javascript/ (that I can't find right now) saying _plotly.js is supported in all SVG-enabled browsers_. But yeah, the fraction of users still using Chrome41 and FF29 must be tiny.

Issue #2922, opened 2018-08-17
Find robust way to split jasmine tests in different karma runs


etpinard:
We've been plagued by various intermittent test failures on CI in container [`jasmine2`](https://github.com/plotly/plotly.js/blob/18b12f2af3c4dc30b63a2044de44588406d29be4/.circleci/test.sh#L37-L38) lately, for example: PR https://github.com/plotly/plotly.js/pull/2850 and most recently the [August 17 `master`](https://circleci.com/workflow-run/2a55b085-1b20-4819-b5a2-6f3c8d6e133c).

These failures can't be reproduced locally and are almost impossible  to reproduce manually during an ssh session. We suspect that these failures are due to CI ressources blowing up during our `@gl` tests.

Branch [`try-fix-test-fail`](https://github.com/plotly/plotly.js/tree/try-fix-test-fail) attempted a few ways to "dilutate" those expansive `@gl` in multiple test runs. The results were mitigated, as the intermittent failures seemed to persist. But note that, this wasn't a particularly great attempt: running the `npm run test-jasmine` command one half of the files that contain `@gl` and then the second half resulted in [25 tests in the first run and 80+ in the second](https://circleci.com/gh/plotly/plotly.js/12628?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) - as tests under `@gl` aren't evenly distributed from test file to test file.

Alternatively, we could try to incorporate [`karma-parallel`](https://github.com/joeljeske/karma-parallel) module into our tests, but at first glance it looks incompatible with the [`karma-spec-tags`](https://www.npmjs.com/package/karma-jasmine-spec-tags) plugin we're using to group tests by `@` tags.

---------------

So, in a private convo @alexcjohnson propose a slightly more evolved solution:

> I’m thinking (for performance as well) we should look for a way (that doesn’t require us to manage it all the time) to split the flaky and gl jasmine tests into like chunks of 10 or 20 tests, and run each of THOSE in a retry loop. Also because then if there’s a real failure in there, it will only have one small chunk to retry 5x.

But finding all `it` test cases in a `describe('@gl .,.'` block isn't trivial, so first we'll need to move all those `@gl` tags in `describe` statements down to their child `it` blocks. We should then make sure no tags are set in `describe` block in a [test syntax test](https://github.com/plotly/plotly.js/blob/master/tasks/test_syntax.js).

PR #2933, opened 2018-08-21
Shard gl jasmine tests

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2922 (hopefully)

This PR adds a _sharding_ script that splits our jasmine test suites into groups of a given number of `it` blocks. We use it here to shard jasmine tests with tag `@gl`, but the same logic could eventually be used in other jasmine tests.

To make the sharding logic simple and predictable, we enforce the following jasmine test tag _rules_:

- `@gl`, `@noCI`, `@noCIdep` and `@flaky` are the only allowed tags
- `@gl` and `@flaky` can't be set in `describe` blocks, only in `it` blocks
- a given `it` block can't have both a`@gl` and a `@flaky` tag (i.e. all `@gl` are _flaky_ and are in runned in retry loop on ci, setting both tags is redundant)

Along the way, commit 77125072d0f568f4f0726f87680b9de84e5e88ed made `gl2d_plot_interact_test.js` and `gl2d_click_test.js` more robust and bdf468ee77546661d306d91618edad9d6e4931ab  moved the `@flaky` test run from `jasmine)` to `jasmine2)` on CI. Things appear to be working fine: [many :green_book: runs are occuring in succession on CI](https://circleci.com/gh/plotly/workflows/plotly.js/tree/shard-gl-tests).

cc @alexcjohnson hopefully this will be enough to not have to worry about intermittent failures for a while.

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -13,12 +13,14 @@ retry () {
     local n=0
 
     until [ $n -ge $MAX_AUTO_RETRY ]; do
+        if [ $n -ge 1 ]; then
+            echo ''
+            echo run $n of $MAX_AUTO_RETRY failed, trying again ...
+            echo ''
+            sleep 15
+        fi
         "$@" && break
         n=$[$n+1]
-        echo ''
-        echo run $n of $MAX_AUTO_RETRY failed, trying again ...
-        echo ''
-        sleep 15
     done
 
     if [ $n -eq $MAX_AUTO_RETRY ]; then
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -40,7 +40,7 @@ case $1 in
         SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl))
 
         for s in ${SHARDS[@]}; do
-            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI,flaky
+            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI
         done
 
         retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -4,6 +4,7 @@
 set +e
 set +o pipefail
 
+ROOT=$(dirname $0)/..
 EXIT_STATE=0
 MAX_AUTO_RETRY=5
 
@@ -34,7 +35,12 @@ case $1 in
         ;;
 
     jasmine2)
-        retry npm run test-jasmine -- --tags=gl --skip-tags=noCI,flaky
+        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl))
+
+        for s in ${SHARDS[@]}; do
+            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI,flaky
+        done
+
         retry npm run test-jasmine -- --tags=flaky --skip-tags=noCI
         exit $EXIT_STATE
         ;;
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -8,22 +8,27 @@ ROOT=$(dirname $0)/..
 EXIT_STATE=0
 MAX_AUTO_RETRY=5
 
+log () {
+    echo -e "\n$1"
+}
+
 # inspired by https://unix.stackexchange.com/a/82602
 retry () {
-    local n=0
+    local n=1
 
     until [ $n -ge $MAX_AUTO_RETRY ]; do
-        if [ $n -ge 1 ]; then
-            echo ''
-            echo run $n of $MAX_AUTO_RETRY failed, trying again ...
-            echo ''
-            sleep 15
-        fi
-        "$@" && break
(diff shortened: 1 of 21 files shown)


.circleci/test.sh
```diff
@@ -34,7 +35,12 @@ case $1 in
         ;;
 
     jasmine2)
-        retry npm run test-jasmine -- --tags=gl --skip-tags=noCI,flaky
+        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl))
```
etpinard: `node tasks/shard_jasmine_tests.js --tag=gl` outputs:

```
gl2d_plot_interact_test.js
gl3d_plot_interact_test.js
parcoords_test.js
gl2d_click_test.js
splom_test.js,cartesian_test.js,gl2d_scatterplot_contour_test.js,gl2d_pointcloud_test.js,gl_plot_interact_basic_test.js
streamtube_test.js,gl2d_date_axis_render_test.js,cone_test.js
```

which is then converted into a bash array (that's the outer `()`) and then looped below.

tasks/shard_jasmine_tests.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+});
+
+var tag = argv.tag;
+var limit = argv.limit;
+
+glob(path.join(pathToJasmineTests, '*.js'), function(err, files) {
+    if(err) throw err;
+
+    var file2cnt = {};
```
etpinard: to note, with `--limit=20` and `--tag=gl`, this fills up to:

```
{ 'cartesian_test.js': 1,
  'cone_test.js': 5,
  'gl_plot_interact_basic_test.js': 2,
  'gl2d_click_test.js': 20,
  'gl2d_date_axis_render_test.js': 4,
  'gl2d_plot_interact_test.js': 28,
  'gl2d_pointcloud_test.js': 2,
  'gl2d_scatterplot_contour_test.js': 2,
  'gl3d_plot_interact_test.js': 26,
  'parcoords_test.js': 30,
  'splom_test.js': 10,
  'streamtube_test.js': 10 }
```

test/jasmine/tests/gl2d_click_test.js
```diff
@@ -180,9 +180,10 @@ describe('Test hover and click interactions', function() {
         gd = createGraphDiv();
     });
 
-    afterEach(function() {
+    afterEach(function(done) {
         Plotly.purge(gd);
         destroyGraphDiv();
+        setTimeout(done, 1000);
```
etpinard: this _appears_ to reduce the number of intermittent failures (and the number of retry attempts) _a little_.

alexcjohnson:
Nice! 💃 

Non-blocking: This leaves `test-jasmine2` a bit long, would be great if we can shift the burden a bit onto other suites, to get the total time under 10 minutes. But not now, lets merge this and get ourselves up and running again!

etpinard:
> have only the final retry run to completion, let all the others fail fast.

That's a great idea. I'll push another commit this morning.

----------

I was also looking at ways to make the retry not waste time bundling on every attempt. I couldn't find anything in user land, but this could be a nice karma plugin. Bundling our test code takes about ~10s, over multiple retries we can easily spend 1 minute just bundling code.

test/jasmine/karma.conf.js
```diff
@@ -235,6 +241,8 @@ func.defaultConfig = {
         suppressPassed: true,
         suppressSkipped: false,
         showSpecTiming: false,
+        // use 'karma-fail-fast-reporter' to fail fast w/o conflicting
+        // with other karma plugins
         failFast: false
```
alexcjohnson: So this `failFast` shouldn't ever be used?

.circleci/test.sh
```diff
@@ -8,22 +8,27 @@ ROOT=$(dirname $0)/..
...
-        fi
-        "$@" && break
+        "$@" --failFast && break
+        log "run $n of $MAX_AUTO_RETRY failed, trying again ..."
         n=$[$n+1]
     done
 
     if [ $n -eq $MAX_AUTO_RETRY ]; then
+        log "one last time, w/o failing fast"
+        "$@" && n=0
```
alexcjohnson: man, shell scripts are ugly... that said, very nicely done 👍

etpinard: That was the cleanest solution I could think of :expressionless:

test/jasmine/karma.conf.js
etpinard: Setting it to `true` won't break anything, but it won't actually _fail fast_.

etpinard: More on this topic: to make `karma-spec-reporter` _fail fast_, one needs to add it to the _plugins_ list: https://github.com/mlex/karma-spec-reporter#configuration

When doing so, it conflicts with `karma-browserify`. 

There might be a way, to list them (order dependent?) so that they don't conflict, but using [`karma-fail-fast-reporter`](https://github.com/michaelbull/karma-fail-fast-reporter) was easy enough.

alexcjohnson:
💃 is still dancing.

PR #2747, opened 2018-06-19
Fix #2743 - old date timezone precision in Chrome 67+

alexcjohnson:
cc @etpinard - the test is somewhat circular, I don't really see a way around this though.

PR #2934, opened 2018-08-21
Use Node 10 and npm 6 for development

etpinard:
... and Chrome 68 on CI.

Node 10 will become LTS in October, but why wait? 

Switching to Node 10 makes us the latest major npm version [with improved package-lock and faster installs](https://blog.npmjs.org/post/173260195980/announcing-npm6) and the latest v8 engine, speeding up our builds and CI runs (e.g. the `build` job on CI runs 30s faster, `syntax` is 30s faster too).

The switch to Chrome 68 appears to make two tests consistently fail:

image

image

The second one is probably related to https://github.com/plotly/plotly.js/pull/2747 (we might need to put a `@noCI` tag on that one).


etpinard:
cc @alexcjohnson

alexcjohnson:
Whew, fixing the timezone issue was harder than it should have been, thanks for persevering there! This looks great. 💃 
Oddly though, it seems like a wash in terms of total test time - `build` is faster but `test-jasmine` and `test-jasmine2` are both slower... oh well, still better to be up to date.

etpinard:
> but test-jasmine and test-jasmine2 are both slower.

Interesting. I guess Chrome 68 is slower than Chrome 64?

PR #2936, opened 2018-08-22
fix logic for hiding zero lines when they conflict with axis lines

### LINKED ISSUE #2874: zeroline doesn't display when showline=true
Using `rangemode: 'tozero'` and `zeroline: true` to always display zero line doesn't display zero line when `showline=true` is added to the axis. 
Example: https://codepen.io/plotly/pen/WvPgPP?editors=1010

alexcjohnson:
Fixes #2874 

The logic was quite a bit more complicated than just switching `ax` -> `counterAxis`... but seems to work in all cases now.

cc @etpinard

### DIFF
--- a/src/plot_api/subroutines.js
+++ b/src/plot_api/subroutines.js
@@ -89,8 +89,9 @@ function lsInner(gd) {
 
         // figure out the main axis line and main mirror line position.
         // it's easier to follow the logic if we handle these separately from
-        // ax._linepositions, which are really only used by mirror=allticks
-        // for the non-main-subplot ticks.
+        // ax._linepositions, which are only used by mirror=allticks
+        // for non-main-subplot ticks, and mirror=all(ticks)? for zero line
+        // hiding logic
         ax._mainLinePosition = getLinePosition(ax, counterAx, ax.side);
         ax._mainMirrorPosition = (ax.mirror && counterAx) ?
             getLinePosition(ax, counterAx,
@@ -270,7 +271,7 @@ function lsInner(gd) {
             // each subplot that gets ticks from "allticks" gets an entry:
             //    [left or bottom, right or top]
             extraSubplot = (!xa._anchorAxis || subplot !== xa._mainSubplot);
-            if(extraSubplot && xa.ticks && xa.mirror === 'allticks') {
+            if(extraSubplot && (xa.mirror === 'allticks' || xa.mirror === 'all')) {
                 xa._linepositions[subplot] = [xLinesYBottom, xLinesYTop];
             }
 
@@ -306,8 +307,8 @@ function lsInner(gd) {
             yLinesXLeft = getLinePosition(ya, xa, 'left');
             yLinesXRight = getLinePosition(ya, xa, 'right');
 
-            extraSubplot = (!ya._anchorAxis || subplot !== xa._mainSubplot);
-            if(extraSubplot && ya.ticks && ya.mirror === 'allticks') {
+            extraSubplot = (!ya._anchorAxis || subplot !== ya._mainSubplot);
+            if(extraSubplot && (ya.mirror === 'allticks' || ya.mirror === 'all')) {
                 ya._linepositions[subplot] = [yLinesXLeft, yLinesXRight];
             }
 
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -2146,6 +2146,46 @@ axes.doTicksSingle = function(gd, arg, skipTitle) {
         return trace.fill && trace.fill.charAt(trace.fill.length - 1) === axLetter;
     }
 
+    function lineNearZero(ax2, position) {
+        if(!ax2.showline || !ax2.linewidth) return false;
+        var tolerance = Math.max((ax2.linewidth + ax.zerolinewidth) / 2, 1);
+
+        function closeEnough(pos2) {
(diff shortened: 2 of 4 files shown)


test/jasmine/tests/axes_test.js
```diff
@@ -2783,6 +2783,103 @@ describe('Test axes', function() {
...
+                assertZeroLines(['x', 'x', 'y', 'y', 'x2', 'y2']);
+                return Plotly.relayout(gd, {'xaxis.showline': true, 'xaxis.mirror': 'all'});
+            })
+            .then(function() {
+                assertZeroLines(['x', 'x', 'y', 'x2']);
+                return Plotly.relayout(gd, {'yaxis.showline': true, 'yaxis.mirror': 'all'});
+            })
+            .then(function() {
+                // x axis still has a zero line on xy2, and y on x2y
+                // all the others have disappeared now
```
etpinard: thanks for the :books:

src/plot_api/subroutines.js
```diff
@@ -306,8 +307,8 @@ function lsInner(gd) {
             yLinesXLeft = getLinePosition(ya, xa, 'left');
             yLinesXRight = getLinePosition(ya, xa, 'right');
 
-            extraSubplot = (!ya._anchorAxis || subplot !== xa._mainSubplot);
-            if(extraSubplot && ya.ticks && ya.mirror === 'allticks') {
+            extraSubplot = (!ya._anchorAxis || subplot !== ya._mainSubplot);
+            if(extraSubplot && (ya.mirror === 'allticks' || ya.mirror === 'all')) {
```
etpinard: Nice find! Is there a test :lock: ing this down?

src/plots/cartesian/axes.js
```diff
@@ -2146,6 +2146,46 @@ axes.doTicksSingle = function(gd, arg, skipTitle) {
...
+        }
+        var linePositions = ax2._linepositions || {};
+        for(var k in linePositions) {
+            if(closeEnough(linePositions[k][0]) || closeEnough(linePositions[k][1])) {
+                return true;
+            }
+        }
+    }
+
+    function anyCounterAxLineAtZero(counterAxis, rng) {
```
etpinard: Very nice function name.

src/plots/cartesian/axes.js
```diff
@@ -2146,6 +2146,46 @@ axes.doTicksSingle = function(gd, arg, skipTitle) {
...
+        var mainCounterAxis = counterAxis._mainAxis;
+        if(!mainCounterAxis) return;
+
+        var zeroPosition = ax._offset + (
+            ((Math.abs(rng[0]) < Math.abs(rng[1])) === (axLetter === 'x')) ?
+            0 : ax._length
+        );
+
+        var counterLetterAxes = axes.list(gd, counterLetter);
```
etpinard: It would nice to avoid looping over all axes here.

Can we skip this loop when `fullLayout._plots[counterAxis._mainSubplot].overlays` is empty, and only check `lineNearZero(counterAxis, zeroPosition)`?

src/plots/cartesian/axes.js
```diff
@@ -2182,12 +2222,19 @@ axes.doTicksSingle = function(gd, arg, skipTitle) {
                     break;
                 }
             }
-            var rng = Lib.simpleMap(ax.range, ax.r2l),
-                showZl = (rng[0] * rng[1] <= 0) && ax.zeroline &&
+            var rng = Lib.simpleMap(ax.range, ax.r2l);
+            var zlData = {x: 0, id: axid};
```
etpinard: No need this do in this PR (opening a new issue would be fine), but we'll need to propagate these changes to large-sploms' regl-based grid line logic:

https://github.com/plotly/plotly.js/blob/b5f8b2300d041f12df90647fe61dba93f7d10c4b/src/traces/splom/base_plot.js#L171-L183

where I didn't just reuse the `axes.js` logic because I wanted to avoid looping over the traces (to compute that `hasBarsOrFill` boolean). Thinking about this again, looping over traces isn't really a performance hit for sploms (I can't really imagine a user wanting to plot more 10 splom traces on the same graph).

So it would be worth trying to expose and :recycle: this logic here for large-sploms.

src/plot_api/subroutines.js
alexcjohnson: ha, there was a test (the subplot case) locking down the *changes* but not the parts that remained -> added these in https://github.com/plotly/plotly.js/pull/2936/commits/545ad472b8e8727e19208bac22fd568e17c781a2

src/plots/cartesian/axes.js
alexcjohnson: good call - it was almost that simple. one more relevant test in https://github.com/plotly/plotly.js/pull/2936/commits/bcd7c0666c80250dd4c78f230032df4facad2c37 and 🐎 in https://github.com/plotly/plotly.js/pull/2936/commits/0f59ca2b1bc0189960c1d558ba7414c1e3b54e2b

src/plots/cartesian/axes.js
alexcjohnson: Ah, thanks for pointing that out! https://github.com/plotly/plotly.js/issues/2938

src/plots/cartesian/axes.js
etpinard: Good to hear it was that easy. :100:

etpinard:
Great! :dancer: 

Thanks for including that perf fix and open an issue for the splom case. :guitar:

PR #2939, opened 2018-08-22
Replot when webgl buffer dims don't match canvas dims

etpinard:
... as this can unfortunately happen when relayouting to large width/height on some browsers e.g. Chrome 68 and Chrome 64 (that we use on CI).

Note that this doesn't just affect sploms, but all regl-based trace type and presumably all WebGL-trace types.

cc @alexcjohnson

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -242,6 +242,24 @@ exports.plot = function(gd, data, layout, config) {
             fullLayout._glcanvas
                 .attr('width', fullLayout.width)
                 .attr('height', fullLayout.height);
+
+
+            var regl = fullLayout._glcanvas.data()[0].regl;
+            if(regl) {
+                if(
+                    fullLayout.width !== regl._gl.drawingBufferWidth ||
+                    fullLayout.height !== regl._gl.drawingBufferHeight
+                 ) {
+                    // Unfortunately, this can happen when relayouting to large
+                    // width/height on some browsers.
+                    Lib.log([
+                        'WebGL context buffer and canvas dimensions do not match,',
+                        'due to browser/WebGL bug.',
+                        'Clearing graph and plotting again.'
+                    ].join(' '));
+                    exports.newPlot(gd, gd.data, gd.layout);
+                }
+            }
         }
 
         return Plots.previousPromises(gd);
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -255,9 +255,12 @@ exports.plot = function(gd, data, layout, config) {
                         Lib.error(msg);
                     } else {
                         Lib.log(msg + ' Clearing graph and plotting again.');
-                        fullLayout._redrawFromWrongGlDimensions = 1;
                         Plots.cleanPlot([], {}, gd._fullData, fullLayout, gd.calcdata);
-                        exports.plot(gd, gd.data, gd.layout);
+                        Plots.supplyDefaults(gd);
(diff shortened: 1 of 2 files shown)


etpinard:
Some links relevant this to bug:

- https://github.com/mrdoob/three.js/issues/5194
- https://bugs.chromium.org/p/chromium/issues/detail?id=445542
- http://jsfiddle.net/862de4nj/47/

src/plot_api/plot_api.js
```diff
@@ -242,6 +242,24 @@ exports.plot = function(gd, data, layout, config) {
...
+                    fullLayout.height !== regl._gl.drawingBufferHeight
+                 ) {
+                    // Unfortunately, this can happen when relayouting to large
+                    // width/height on some browsers.
+                    Lib.log([
+                        'WebGL context buffer and canvas dimensions do not match,',
+                        'due to browser/WebGL bug.',
+                        'Clearing graph and plotting again.'
+                    ].join(' '));
+                    exports.newPlot(gd, gd.data, gd.layout);
```
etpinard: Hmm. This may not be the best idea actually. If the canvas and gl buffer dimensions still don't match after the replot, we could have an infinite loop.

alexcjohnson: I'm not sure how that would happen, but if you're worried about it we could do something to limit to 1 redraw like
```
if(!gd._reglRedrawing) {
    gd._reglRedrawing = true;
    return exports.newPlot(...).then(function() { gd._reglRedrawing = false; });
}
```

etpinard: That's essentially when I went for in https://github.com/plotly/plotly.js/pull/2939/commits/85a646f374ab26b191a2c0eb9a1af470b0b72386

etpinard:
New attempt in -> https://github.com/plotly/plotly.js/pull/2939/commits/6fcec6eeea723cb5cc40fe01813760d078510fb1

Note that, we need to call `Plots.supplyDefaults` and `Plots.doCalcdata` after `Plots.cleanPlot` the have correct scene reference before `_module.plot`.

src/plot_api/plot_api.js
```diff
@@ -242,6 +242,28 @@ exports.plot = function(gd, data, layout, config) {
...
+                    if(fullLayout._redrawFromWrongGlDimensions) {
+                        Lib.error(msg);
+                    } else {
+                        Lib.log(msg + ' Clearing graph and plotting again.');
+                        Plots.cleanPlot([], {}, gd._fullData, fullLayout, gd.calcdata);
+                        Plots.supplyDefaults(gd);
+                        fullLayout = gd._fullLayout;
+                        Plots.doCalcdata(gd);
+                        fullLayout._redrawFromWrongGlDimensions = 1;
+                        return drawFramework();
```
alexcjohnson: Nice! Still a substantial win as we don't need to redo all the drawing.

So, there's still recursion here - but it's just back into this function. So instead of a `_redrawFromWrongGlDimensions` flag on `_fullLayout` (that we'd probably have to handle more carefully in principle due to `relinkPrivateKeys`) could we use a local var in the outer scope?

etpinard: Ok, done in https://github.com/plotly/plotly.js/pull/2939/commits/3879a41d94423b5fb73a7547213fc6bf4bdc1b0a

etpinard: one less private key to worry about when we'll :hocho: `relinkPrivateKeys`.

PR #2944, opened 2018-08-27
1852 persistent click via select points

rmoestl:
*Superseding PR #2824*, this PR will introduce the select-on-click / persistent selection feature discussed in issue #1852.

A few remarks:
- There's one uncertainty regarding testing interactions that should have no effect but might be processed asynchronously. See open TODOs in `select_test.js`.
- For now I've decided to not clear selection when user clicks a blank spot. I felt, that this cause losing a selection to easily. Instead clear a selection by click is easy with double click in lasso / select mode, and in zoom / pan mode one only needs to select one point and click it again.
- As per the original request in #1852 we've not decided how the user "would be informed that points are clickable".


etpinard:
Amazing PR @rmoestl. Nothing short of an herculean effort :muscle: 

Thanks for meticulously testing click-to-select on :sparkle: ALL :sparkle: selectable trace types!

I'll make a more technical tomorrow morning (UTC-4). 

---------------

In the meantime, could you clarify:

> As per the original request in #1852 we've not decided how the user "would be informed that points are clickable".

Is this referring to Chris' opening bullet point:

> From a UI perspective, the user would be informed that points are clickable and unclickable through:
> - cursor: pointer

? 

Looking at the changed lines quickly, I don't think https://github.com/plotly/plotly.js/issues/1852#issuecomment-391401079 got addressed:

> But I'd also be OK with just adding something to the docs for clickmode - like "clickmode: 'select' with hovermode: 'x' can be confusing, consider explicitly setting hovermode: 'closest' when using this feature"

it should be fairly easy to patch `fx/layout_defaults.js` accordingly.

and about your

> For now I've decided to not clear selection when user clicks a blank spot. I felt, that this cause losing a selection to easily. Instead clear a selection by click is easy with double click in lasso / select mode, and in zoom / pan mode one only needs to select one point and click it again.

I agreed :100: % here

etpinard:
> so I'm happy to change it to AJ's suggestion

Yep, let's go with AJ's suggestion :rocket:

rmoestl:
> Yep, let's go with AJ's suggestion

Incorporated AJ's suggestion and went over the description again.

I'm no longer entirely sure if I understood the event emitting part correctly: Is it correct, that when `clickmode` flag list is missing `event` (i.e. just `select`), neither `plotly_click` nor `plotly_selected` should fire?

src/plots/cartesian/dragbox.js
```diff
@@ -178,7 +183,19 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
+                // Note: clearing / resetting selection state only happens, when user
+                // triggers at least one interaction in pan/zoom mode. Otherwise, the
+                // select/lasso outlines are deleted (in plots.js.cleanPlot) but the selection
+                // cache isn't cleared. So when the user switches back to select/lasso and
+                // 'adds to a selection' with Shift, the "old", seemingly removed outlines
+                // are redrawn again because the selection cache still holds their coordinates.
+                // However, this isn't easily solved, since plots.js would need
+                // to have a reference to the dragOptions object (which holds the
+                // selection cache).
+                clearAndResetSelect();
```
etpinard: Here's a gif that hopefully clearifies things (as we might need input from @alexcjohnson )

peek 2018-08-29 12-18

Note that the go from select -> pan -> select w/o actually panning. 

Personally, I wouldn't consider this a bug as the "cached" selection outlines are still "correct" (i.e. they are drawn at the correct x/y coordinates). Now, is this the "ideal" behavior, I'm not so sure.

test/jasmine/tests/select_test.js
```diff
@@ -109,6 +117,582 @@ var BOXEVENTS = [1, 2, 1];
...
+              return _clickPt(mock14Pts[7], { shiftKey: true });
+          })
+          .then(function() {
+              assertSelectedPoints(35);
+          })
+          .catch(failTest)
+          .then(done);
+    });
+
+    it('@flaky is supported by scattergl in pan/zoom mode', function(done) {
```
etpinard: Post https://github.com/plotly/plotly.js/pull/2933, I think we make an effort to tag all tests involving any gl-based trace `@gl`, and not `@flaky`.

All `@gl` tests are considered _flaky_ (i.e. they're runned in a retry loop on CI). Moreover, `@gl` tests are sharded to try to limit to ressource used in a given karma process.

test/jasmine/tests/select_test.js
```diff
@@ -109,6 +117,582 @@ var BOXEVENTS = [1, 2, 1];
...
+            testCase('scattercarpet', require('@mocks/scattercarpet.json'), 532, 178,
+              [undefined, [], [], [], [], [], [2]], { width: 1100, height: 450 }),
+
+            // scatterpolar and scatterpolargl do not support pan (the default),
+            // so set dragmode to zoom
+            testCase('scatterpolar', require('@mocks/polar_scatter.json'), 130, 290,
+              [[], [], [], [19], [], []], { dragmode: 'zoom' }),
+            testCase('scatterpolargl', require('@mocks/glpolar_scatter.json'), 130, 290,
+              [[], [], [], [19], [], []], { dragmode: 'zoom' }),
+            testCase('splom', require('@mocks/splom_lower.json'), 427, 400, [[], [7], []])
```
etpinard: We should split this list into two, one with svg-based mock (and tag `@flaky`) and another with gl-based mocks (and tag `@gl`). Hopefully, this will be enough to make the splom test pass consistently on CI :crossed_fingers:

src/components/fx/layout_attributes.js
```diff
@@ -18,6 +18,28 @@ fontAttrs.family.dflt = constants.HOVERFONT;
...
+            'The *select* flag enables selecting single',
+            'data points via click. This mode also supports persistent selections,',
+            'meaning that pressing Shift while clicking, adds to / subtracts from an',
+            'existing selection. *select* with `hovermode`: *x* can be confusing, consider',
+            'explicitly setting `hovermode`: *closest* when using this feature.',
+            'Selection events are sent accordingly as long as *event* flag is set as well.',
+            'When the *event* flag is missing, `plotly_click` and `plotly_selected`',
+            'events are not fired.'
+        ].join(' ')
+    },
```
alexcjohnson: Probably want to add [`extras: ['none']`](https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/traces/scatter/attributes.js#L109) here as we [don't allow empty flaglists](https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/lib/coerce.js#L246)
Not sure why someone would want a click to do nothing, but as usual that's no reason to forbid it 😄

src/plots/cartesian/dragbox.js
alexcjohnson: I like the behavior that we don't clear the selection itself until the last possible moment... but it would certainly be nice if the outline were redrawn when you switch back to a select mode. It's a sufficiently uncommon edge case though that I'd rather not hold up the PR for it - just leave it as is for now and make a new issue to follow up on this later.

src/plots/cartesian/select.js
```diff
@@ -349,6 +289,303 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
+    var searchTraces;
+    var searchInfo;
+    var currentSelectionDef;
+    var selectionTester;
+    var traceSelection;
+    var thisTracesSelection;
+    var pointOrBinSelected;
+    var subtract;
+    var eventData;
+    var i;
```
alexcjohnson: 🐄 we usually combine uninitialized vars multiple per line, like https://github.com/plotly/plotly.js/pull/2944/files#diff-8c5d8f688f2bb41a18d35397e5f58864R54

src/plots/cartesian/select.js
```diff
@@ -349,6 +289,303 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
...
+      (evt.shiftKey || evt.altKey) &&
+      (plotinfo.selection && plotinfo.selection.selectionDefs) &&
+      !dragOptions.selectionDefs
+    ) {
+        // take over selection definitions from prev mode, if any
+        dragOptions.selectionDefs = plotinfo.selection.selectionDefs;
+        dragOptions.mergedPolygons = plotinfo.selection.mergedPolygons;
+    } else if(
+      (!evt.shiftKey && !evt.altKey) ||
+      ((evt.shiftKey || evt.altKey) && !plotinfo.selection)
```
alexcjohnson: 🐄 logic here is a bit tough to read (and I think could be simplified). If we start with:
`var hasModifierKey = evt.shiftKey || evt.altKey;`
then this is (?):
`!hasModifierKey || (hasModifierKey && !plotinfo.selection)`
and then it's clear it simplifies to:
`!hasModifierKey || !plotinfo.selection`

alexcjohnson:
> Is it correct, that when clickmode flag list is missing event (i.e. just select), neither plotly_click nor plotly_selected should fire?

Yes, agreed.

test/jasmine/tests/select_test.js
```diff
@@ -1134,6 +1155,38 @@ describe('@flaky Test select box and lasso in general:', function() {
...
+        var fig = Lib.extendDeep({}, require('@mocks/0.json'));
+
+        fig.layout.dragmode = 'select';
+        Plotly.plot(gd, fig)
+          .then(function() {
+              return drag([[350, 100], [400, 400]]);
+          })
+          .then(function() {
+              _assertSelectedPoints([49, 50, 51, 52, 53, 54, 55, 56, 57]);
+              return doubleClick(500, 200, { shiftKey: true });
```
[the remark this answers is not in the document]
rmoestl: To test that under the hood the selection cache (`dragOptions.selectionDefs`) is cleaned up appropriately. When you were holding the Shift key while clearing the selection (through double-click in lasso/select or clicking the last selected point), then starting a new selection holding the Shift key was adding to the previously cleared selection. The problem was that the selection cache wasn't cleared appropriately. See this commit: https://github.com/plotly/plotly.js/commit/bb44b6d16263c694daf5945f1ef83afd8aee1bcc

test/jasmine/tests/select_test.js
```diff
@@ -630,7 +630,8 @@ describe('@flaky Click-to-select', function() {
             testCase('scatterpolar', require('@mocks/polar_scatter.json'), 130, 290,
               [[], [], [], [19], [], []], { dragmode: 'zoom' }),
             testCase('scatterpolargl', require('@mocks/glpolar_scatter.json'), 130, 290,
-              [[], [], [], [19], [], []], { dragmode: 'zoom' })
+              [[], [], [], [19], [], []], { dragmode: 'zoom' }),
+            testCase('splom', require('@mocks/splom_lower.json'), 427, 400, [[], [7], []])
```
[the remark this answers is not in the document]
rmoestl: Yes, it is. Succeeds locally, but maybe the `@gl` flag you mentioned somewhere else is fixing it.

src/plots/cartesian/select.js
```diff
@@ -553,7 +553,7 @@ function isOnlyThisBinSelected(searchTraces, clickedPtInfo) {
         if(isSameTrace) {
             trace = clickedPtInfo.searchInfo.cd[0].trace;
             if(trace.selectedpoints.length === clickedPtInfo.pointNumbers.length) {
-                for(i = 0; i > clickedPtInfo.pointNumbers.length; i++) {
+                for(i = 0; i < clickedPtInfo.pointNumbers.length; i++) {
```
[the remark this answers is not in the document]
rmoestl: Looking at the code again, it would have been caught by a Jasmine test. However, the calling code in the if condition in `selectOnClick` was "covering" the bug due to an additional call to `isPointOrBinSelected`. This is somewhat redundant and I'll simply the if condition appropriately.

src/lib/polygon.js
```diff
@@ -174,34 +174,31 @@ polygon.tester = function tester(ptsIn) {
     };
 };
 
+// TODO Somewhat redundant to multiTester in 'lib/select.js'
```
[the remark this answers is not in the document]
rmoestl: > It's not just used by cartesian though, right?

Exactly. E.g. it is used in `polar` and `choropleth`.

> I don't think polygon.multiTester gets called anywhere now

Nope. It's called in `polygon.tester` in the first line.

src/plots/cartesian/select.js
```diff
@@ -185,14 +187,14 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
         }
 
         // create outline & tester
-        if(dragOptions.polygons && dragOptions.polygons.length) {
+        if(dragOptions.selectionDefs && dragOptions.selectionDefs.length) {
             mergedPolygons = mergePolygons(dragOptions.mergedPolygons, currentPolygon, subtract);
             currentPolygon.subtract = subtract;
-            testPoly = multipolygonTester(dragOptions.polygons.concat([currentPolygon]));
+            selectionTester = multiTester(dragOptions.selectionDefs.concat([currentPolygon]));
```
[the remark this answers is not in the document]
rmoestl: No problem, it was just a necessity from a maintenance perspective.

src/plots/cartesian/dragbox.js
rmoestl: I created issue #2955 referencing this discussion here.

test/jasmine/tests/select_test.js
rmoestl: I'll split them up. :crossed_fingers:

src/lib/select.js
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param {Array} pt - [0] is the x coordinate, [1] is the y coordinate of the point.
+     * @param {*} arg - An optional parameter to pass down to wrapped testers.
+     * @param {number} pointNumber - The point number of the point within the underlying data array.
+     * @param {number} searchInfo - An object identifying the trace the point is contained in.
+     *
+     * @return {boolean} true if point is considered to be selected, false otherwise.
+     */
+    function contains(pt, arg, pointNumber, searchInfo) {
+        var yes = false;
```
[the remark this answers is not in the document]
rmoestl: Sorry, I wasn't looking through that code too much since I just moved `multiTester` from `lib/polygon.js`. But I remember this `yes` identifier catching my attention once in a while. I'm going to change it.

src/components/fx/layout_attributes.js
alexcjohnson: > Does that mean, that someone can define `clickmode: 'none'`?

Exactly, then nothing at all will happen on a click. `extras` are single items you can use instead of the flag list - in some other places we have `'all'` for example, so you don't have to specify each individual piece. For the most part `'none'` should *just work* because it will fail all the `.indexOf(flag)` tests.

src/lib/select.js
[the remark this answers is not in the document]
alexcjohnson: Ah, I hadn't noticed that it was already there before. Well, moving is the best time to clean house :) thanks!

src/components/fx/layout_attributes.js
rmoestl: Okay, so I'll just add `clickmode: 'none'`.

src/plots/cartesian/select.js
rmoestl: Done.

test/jasmine/tests/select_test.js
```diff
@@ -109,6 +117,582 @@ var BOXEVENTS = [1, 2, 1];
...
+          .catch(failTest)
+          .then(done);
+
+        function clickPtImmediately() { return _immediateClickPt({ x: 610, y: 342 }); }
+        function clickPt() { return _clickPt({ x: 610, y: 342 }); }
+        function clickBox() { return _clickPt({ x: 565, y: 329 }); }
+    });
+
+    describe('is disabled when clickmode does not include \'select\'', function() {
+        // TODO How to test for pan and zoom mode as well? Note, that
```
rmoestl: @etpinard or @alexcjohnson Would you mind to take a look at this question? Maybe it's just my lack of understanding how Plotly works, but I wonder how I can be sure when a click has been processed when the click itself should not lead to any change.

src/plots/cartesian/select.js
rmoestl: Btw. I quickly looked through ESLint to see if there's an appropriate rule. First, I thought https://eslint.org/docs/rules/one-var would fit our purpose, but IMO it's not.

src/lib/polygon.js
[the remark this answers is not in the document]
etpinard: Ideally, all of `cartesian/select.js` should be moved out of `cartesian/` - as @alexcjohnson said it is not _just_ used for cartesian subplots anymore.

What I was suggesting is to move the contents of `lib/select.js` into the same file as `cartesian/select.js` as the contents of `lib/select.js` aren't used anywhere else.

Once that's done, we can move `cartesian/select.js` to either `components/fx` (side-by-side with the hover routines) or in a new `components/` directory e.g. `components/select`.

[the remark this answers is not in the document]
rmoestl: Done. Moved contents of `lib/select.js` into cartesian/select.js`.

rmoestl:
While testing the issues raised by Etienne, I think I discovered another bug in `scattergl` that already exists prior click-to-select.

Steps to reproduce:
1. render a scattergl plot with markers
2. switch to lasso mode
3. select a data point
4. clear selection with double click
5. switch to zoom mode
6. switch back to lasso mode
7. **Bug:** although there's no active selection, all markers are rendered as deselected (the green markers of trace 2 in the screenshot below)

image

etpinard:
> I suggest to consider this a separate issue (if not already known).

Thanks for the report -> https://github.com/plotly/plotly.js/issues/2958

rmoestl:
@etpinard, regarding this bug

> Now, I noticed one potential bug (I can't remember if this is something we agreed on earlier):
> 
> - start from mock `12.json` (but I think all selectable mocks behaves this way)
> - call `relayout('clickmode', 'select')`
> - call `relayout('dragmode', 'select')`
> - select something
> - call `relayout('dragmode', 'pan') // or 'zoom'`
> - shift-click-select a pt
> - this clear all selected pts

This is a little unfortunate. It all happens when the selection state is cleared in `dragbox`, see https://github.com/plotly/plotly.js/blob/ac4b909748166cb553bdea888923d77ac28dbe7d/src/plots/cartesian/dragbox.js#L197
I believe we do that here because if we've drawn polygons previously in lasso / select and then we pan or zoom the plotting area, these polygons are no longer valid. Why? Because panning / zooming isn't implemented for those polygons. Haven't looked at the code explicitly, but that's my interpretation when testing it interactively.

If my interpretation is correct, what could we do?
1. Zoom and pan polygons accordingly (superb UX-wise)
2. Defer clearing selection state until pan an zoom really happens (i.e. calling `clearAndResetSelect` not in `dragElement.prepFn` but in first call of `dragElement.moveFn`. Not perfect of course. A user would still expect that selection is retained although he zoomed / panned.

test/jasmine/tests/select_test.js
```diff
@@ -615,16 +617,32 @@ describe('Click-to-select', function() {
...
                   var mockCopy = Lib.extendDeep(
                     {},
                     testCase.mock,
                     defaultLayoutOpts,
-                    customLayoutOptions);
+                    customLayoutOptions,
+                    customConfigOptions);
 
-                  Plotly.plot(gd, mockCopy.data, mockCopy.layout)
+                  Plotly.plot(gd, mockCopy.data, mockCopy.layout, mockCopy.config)
```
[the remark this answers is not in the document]
rmoestl: Pushed another commit to fix this.

Problem was that in pan mode, mapbox module was registering new click listeners on each relayout, but not unregister accordingly. Ended up having multiple click listeners that were both calling `selectOnClick`. One click listener then overruled the other one: first listener selected a point, second listener was deselecting the point again.

etpinard:
@rmoestl about https://github.com/plotly/plotly.js/pull/2944#issuecomment-417450674. First, thanks very much for the detailed response.  

Thinking about this again, I'm starting to believe that the current behavior isn't that bad. Or, in other words, _fixing_ it won't be worth it. The current behavior only looks odd when users shift-select pt in `zoom` or `pan` mode before actually zoom or panning. Clearing selection state during `prepFn` is what we want until find a solution to your (1) pt, but that's for another PR. Consider my comment as non-blocking.

test/jasmine/tests/select_test.js
etpinard: Hmm. I'm not entirely sure I understand your question here. Under dragmode `pan` and `zoom` and clickmode `event`, the "selection code" shouldn't be executed.

>  I could try to handle the plotly_click event

So yeah, asserting that `plotly_click` returns some info about `mock14Pts[1]` would suffice.

Now, if `clickmode: 'select'` both `plotly_click` and `plotly_selected` aren't fired under dragmode `pan` and `zoom`. To check that indeed the selection _happened_, I would simply check

```js
expect(gd.data[0].selectedpoints).toEqual([14]);
```

I hope this helps!

rmoestl: Let me try to ask the question in a different way: I want to test that click-to-select is disabled when clickmode is `event`. 

So the thing is, when a click event happens, plotly.js is dealing with that click. But maybe in an asynchronous fashion, for example because of throttling. The question is, when can I be sure plotly.js has finished the asynchronous processing of the click.

What I want to avoid is the following:
1. emulating the click
2. asserting selection state hasn't changed (too early!)
3. plotly.js asynchronous handling of the click event runs and potentially changes selection state because of a bug

rmoestl: I fixed those tests and in addition added some more that ensure plotly_click is triggered after plotly_selected.

test/jasmine/tests/select_test.js
```diff
@@ -693,24 +673,78 @@ describe('Click-to-select', function() {
...
+    describe('triggers \'plotly_selected\' before \'plotly_click\'', function() {
+        [
+            testCase('cartesian', require('@mocks/14.json'), 270, 160, [7]),
+            testCase('geo', require('@mocks/geo_scattergeo-locations.json'), 285, 240, [1]),
+            testCase('ternary', require('@mocks/ternary_markers.json'), 485, 335, [7]),
+            testCase('mapbox', require('@mocks/mapbox_0.json'), 650, 195, [[2], []], {},
+              { mapboxAccessToken: require('@build/credentials.json').MAPBOX_ACCESS_TOKEN }),
+            testCase('polar', require('@mocks/polar_scatter.json'), 130, 290,
+              [[], [], [], [19], [], []], { dragmode: 'zoom' })
+        ].forEach(function(testCase) {
```
etpinard: Thanks for standardizing this :ok_hand:

test/jasmine/tests/select_test.js
```diff
@@ -109,6 +123,630 @@ var BOXEVENTS = [1, 2, 1];
...
+            // so set dragmode to zoom
+            testCase('scatterpolar', require('@mocks/polar_scatter.json'), 130, 290,
+              [[], [], [], [19], [], []], { dragmode: 'zoom' }),
+            testCase('scatterpolargl', require('@mocks/glpolar_scatter.json'), 130, 290,
+              [[], [], [], [19], [], []], { dragmode: 'zoom' }).enableGl(),
+            testCase('splom', require('@mocks/splom_lower.json'), 427, 400, [[], [7], []]).enableGl()
+        ]
+          .forEach(function(testCase) {
+              var ciAnnotation = testCase.gl ? 'gl' : 'flaky';
+              it('@' + ciAnnotation + ' trace type ' + testCase.label, function(done) {
```
etpinard: Hmm. Looking at this again, I don't this this will work. [`shard_jasmine_tests`](https://github.com/plotly/plotly.js/blob/master/tasks/shard_jasmine_tests.js) looks for "static" tags only. So here, the `@gl` tags won't count towards the shard rankings. 

It's probably best (albeit less DRY) to just split this specs list into two: one with all `@flaky` tags and the with all `@gl` tags.

test/jasmine/tests/select_test.js
```diff
@@ -109,6 +123,630 @@ var BOXEVENTS = [1, 2, 1];
...
+              });
+          });
+    });
+
+    describe('triggers \'plotly_selected\' before \'plotly_click\'', function() {
+        [
+            testCase('cartesian', require('@mocks/14.json'), 270, 160, [7]),
+            testCase('geo', require('@mocks/geo_scattergeo-locations.json'), 285, 240, [1]),
+            testCase('ternary', require('@mocks/ternary_markers.json'), 485, 335, [7]),
+            testCase('mapbox', require('@mocks/mapbox_0.json'), 650, 195, [[2], []], {},
```
etpinard: We usually skip (with tag `@noCI`) mapbox tests on CI as they tend to be too resource-intensive. 

@rmoestl do you recall any of your newly added mapbox test failing on CI?

rmoestl: No, I don't. Anyways, I've separated the (new) mapbox tests to use the `@noCI` annotation.

test/jasmine/tests/select_test.js
rmoestl: No problem! Just pushed a commit that splits up those tests by CI annotation.

test/jasmine/tests/select_test.js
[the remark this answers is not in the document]
etpinard: Problem fixed :white_check_mark:

src/lib/polygon.js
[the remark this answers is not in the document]
etpinard: @rmoestl thanks for moving the contents to `cartesian/select.js`.

Now, would you mind :hocho: the `multiTester` routine and the `TODO` above from this file. `polygon.multiTester` isn't used anywhere now.

test/jasmine/tests/select_test.js
etpinard: Nicely done. Conversation resolved.

test/jasmine/tests/select_test.js
etpinard: Great. Thank you!

test/jasmine/tests/select_test.js
etpinard: Fantastic.

src/components/fx/layout_defaults.js
```diff
@@ -16,6 +16,8 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
         return Lib.coerce(layoutIn, layoutOut, layoutAttributes, attr, dflt);
     }
 
+    coerce('clickmode');
```
etpinard: We're still missing @alexcjohnson smart default suggestion where when a user set `clickmode: 'select'` the default `hovermode` value becomes `'closest'`.

test/jasmine/tests/select_test.js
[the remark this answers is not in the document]
etpinard: Ok. Thanks for the info. As shift+doubleClick isn't a common interaction, would you mind adding a comment above this line about this?

etpinard:
@rmoestl you're down the :three: unresolved items (if I counted correctly):

- https://github.com/plotly/plotly.js/pull/2944/files#r215695342
- https://github.com/plotly/plotly.js/pull/2944/files/23024b2f045f186fbd002cde737ee15dd2092864..cbd9f0769e69fc0a09ef5415d94eaa167bcf46fd#r215692154
- https://github.com/plotly/plotly.js/pull/2944/files/a6e49fde61fffca2be2cf5ba3d7690f90904b48f..bb44b6d16263c694daf5945f1ef83afd8aee1bcc#r215696218

src/components/fx/layout_defaults.js
rmoestl: Oh, sorry, I misunderstood that. I thought mentioning in `clickmode`'s description that hovermode set to closest is best for most applications was "enough". 

So what we want to have is when `clickmode` includes the `select` flag and nothing is set explicitly for `hovermode`, the default of `hovermode` should be `closest`. Is that right?

src/lib/polygon.js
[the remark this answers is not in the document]
rmoestl: Sorry, to repeat that, but isnt't `polygon.multiTester` called here https://github.com/plotly/plotly.js/blob/cbd9f0769e69fc0a09ef5415d94eaa167bcf46fd/src/lib/polygon.js#L34?

test/jasmine/tests/select_test.js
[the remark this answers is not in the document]
rmoestl: I just added a commit adding the comment.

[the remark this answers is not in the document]
etpinard: Nicely done!

src/components/fx/layout_defaults.js
alexcjohnson: > So what we want to have is when `clickmode` includes the `select` flag and nothing is set explicitly for `hovermode`, the default of `hovermode` should be `closest`. Is that right? 

Yep! Which should be as simple as modifying the `hovermodeDflt` logic below.

src/lib/polygon.js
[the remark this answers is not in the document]
etpinard: My bad! I grep'ed for `multiTester` not `multitester`. You're correct, `multitester` is currently used in `tester` in  `lib/polygon.js`.

Now, if I recall correctly, that `polygon.multitester` call in `polygon.tester` was only intended for mulit-selections. Git blame points us to https://github.com/plotly/plotly.js/commit/b800e5537fc67d5b0493b33a57418c0a3a2fff9f which is consistent.

When commenting out this line

https://github.com/plotly/plotly.js/blob/cbd9f0769e69fc0a09ef5415d94eaa167bcf46fd/src/lib/polygon.js#L34

no test fail.

So yeah, I think it's safe to :hocho: everything about `multitester` in `lib/polygon.js`.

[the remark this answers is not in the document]
alexcjohnson: I do think we have enough tests that you could try replacing that line with
`if(Array.isArray(ptsIn[0][0])) throw new Error('multitester called!');`
or something and see if it ever happens.

But if it *does*, could that line just point to the new `multiTester`? Of course doing that would make a circular reference unless these are brought into the same file - which might be better anyway, I feel like these belong together (even if we can't quite agree on where that together should be...)

src/components/fx/layout_defaults.js
rmoestl: Done!

src/lib/polygon.js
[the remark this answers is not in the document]
rmoestl: Running all tests didn't throw the error. I left throwing the error in place and added a nice comment to it, so that future readers understand the "why".

src/lib/polygon.js
```diff
@@ -31,7 +31,11 @@ var polygon = module.exports = {};
  *          returns boolean: is pt inside the polygon (including on its edges)
  */
 polygon.tester = function tester(ptsIn) {
-    if(Array.isArray(ptsIn[0][0])) return polygon.multitester(ptsIn);
+
+    // Throw an error if any code tries to pass this no
+    // longer supported argument shape. The local multitester became
+    // obsolete when introducing click-to-select in PR #2944.
+    if(Array.isArray(ptsIn[0][0])) throw new Error('multitester called!');
```
alexcjohnson: I guess this is the safest way to go if we're not confident enough to remove this entirely just yet - or perhaps even safer would be a `Lib.warn` so the whole plot doesn't freeze, just selection or hover breaks? @etpinard thoughts? Seems to me if we do leave this here, we should make an issue and tag it with `v1.42` or `v1.43` to remove this block.

etpinard: I'm voting for :hocho: this thing entirely. I'm 99.99% percent confident this won't break anything.

rmoestl: :hocho: Removed it entirely.

etpinard:
Amazing work @rmoestl . A hard-fought :dancer: 

This thing will be in the running for _PR of the year_ :trophy:

rmoestl:
:bowing_man: Thanks for your support, @etpinard and @alexcjohnson!

PR #2877, opened 2018-08-06
Welsh UK locale formats type:translation

HywelMJ:
Welsh UK locale formats

References: http://www.comisiynyddygymraeg.cymru/English/Part%203/10%20Locales%20alphabets%20and%20character%20sets/10.1%20Locales/Pages/10-1-3-Differences-between-British-English-and-British-Welsh-locales.aspx
http://www.unicode.org/cldr/charts/28/verify/dates/cy.html

PR #2945, opened 2018-08-27
Welsh translation

alexcjohnson:
supersedes #2877 
cc @HywelMJ - I tweaked formatting to make the syntax test happy, and swapped `tres` to `olin` - look good?
@etpinard I also put in unix-style `\n` line endings, so the optional `\r\n` test isn't necessary now - we could add it later but that can be a separate discussion.


HywelMJ:
Yes, I think everything looks OK. Thanks very much.

PR #2949, opened 2018-08-28
disable hover for sankey traces when hovermode is false

### LINKED ISSUE #2782: Hoverinfo 'none' and 'skip' do not work with sankey traces
`layout.hovermode: false`, `hoverinfo: 'none'` and `hoverinfo: 'skip'` all seem to not work on `sankey` traces.

antoinerg:
Fixes #2782 

Simple fix: hover handler functions now simply return if hovermode is false.

cc @etpinard

### DIFF
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -177,6 +177,7 @@ module.exports = function plot(gd, calcData) {
     };
 
     var linkUnhover = function(element, d, sankey) {
+        if(gd._fullLayout.hovermode === false) return;
         d3.select(element).call(linkNonHoveredStyle.bind(0, d, sankey, true));
         gd.emit('plotly_unhover', {
             event: d3.event,
@@ -240,6 +241,7 @@ module.exports = function plot(gd, calcData) {
     };
 
     var nodeUnhover = function(element, d, sankey) {
+        if(gd._fullLayout.hovermode === false) return;
         d3.select(element).call(nodeNonHoveredStyle, d, sankey);
         gd.emit('plotly_unhover', {
             event: d3.event,
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -130,6 +130,7 @@ module.exports = function plot(gd, calcData) {
     };
 
     var linkHover = function(element, d, sankey) {
+        if(gd._fullLayout.hovermode === false) return;
         d3.select(element).call(linkHoveredStyle.bind(0, d, sankey, true));
         gd.emit('plotly_hover', {
             event: d3.event,
@@ -143,6 +144,7 @@ module.exports = function plot(gd, calcData) {
     var outgoingLabel = _(gd, 'outgoing flow count:') + ' ';
 
     var linkHoverFollow = function(element, d) {
+        if(gd._fullLayout.hovermode === false) return;
         var trace = d.link.trace;
         var rootBBox = gd._fullLayout._paperdiv.node().getBoundingClientRect();
         var boundingBox = element.getBoundingClientRect();
@@ -193,6 +195,7 @@ module.exports = function plot(gd, calcData) {
     };
 
     var nodeHover = function(element, d, sankey) {
+        if(gd._fullLayout.hovermode === false) return;
         d3.select(element).call(nodeHoveredStyle, d, sankey);
         gd.emit('plotly_hover', {
             event: d3.event,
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/sankey_test.js
```diff
@@ -620,3 +642,8 @@ function assertLabel(content, style) {
         fontColor: style[4]
     });
 }
+
+function assertNoLabel() {
+    var g = d3.selectAll('.hovertext');
+    expect(g[0].length).toBe(0);
```
alexcjohnson: `g.size()` is better than `g[0].length`

test/jasmine/tests/sankey_test.js
```diff
@@ -482,6 +482,28 @@ describe('sankey tests', function() {
...
+
+        it('should not show labels if hovermode is false', function(done) {
+            var gd = createGraphDiv();
+            var mockCopy = Lib.extendDeep({}, mock);
+
+            function _hover(px, py) {
+                mouseEvent('mousemove', px, py);
+                mouseEvent('mouseover', px, py);
+                Lib.clearThrottle();
+            }
```
alexcjohnson: Can we push this to an outer scope so we're not just repeating the function from above?

test/jasmine/tests/sankey_test.js
antoinerg: Good point!

src/traces/sankey/plot.js
```diff
@@ -143,6 +144,7 @@ module.exports = function plot(gd, calcData) {
     var outgoingLabel = _(gd, 'outgoing flow count:') + ' ';
 
     var linkHoverFollow = function(element, d) {
+        if(gd._fullLayout.hovermode === false) return;
```
alexcjohnson: Do we need to bail out of the `unhover` handlers as well, to ensure that we don't emit `plotly_unhover` events? Or does this already avoid these events?

antoinerg: @alexcjohnson It was indeed emitting a bunch of `plotly_unhover` events. It is fixed in my next commit.

alexcjohnson: Great! I guess then we should 🔒 the events too, by adding a `hovermode: false` case to the [click/hover events test](https://github.com/plotly/plotly.js/blob/45f06c447af2bc6b5e7196b2b949d6b4b4737ca6/test/jasmine/tests/sankey_test.js#L509)

test/jasmine/tests/sankey_test.js
```diff
@@ -601,6 +609,30 @@ describe('sankey tests', function() {
...
+
+        it('should not output hover/unhover event data when hovermoder is false', function(done) {
+            var fig = Lib.extendDeep({}, mock);
+
+            Plotly.plot(gd, fig)
+            .then(function() { return Plotly.relayout(gd, 'hovermode', false); })
+            .then(function() { return _hover('node'); })
+            .then(failTest).catch(function(err) {
+                expect(err).toBe('plotly_hover did not get called!');
+            })
```
alexcjohnson: Ah, clever 🎉 
Does `failTest` give an intelligible message if we get there? ie if we run this test without your previous commit, would it be clear what went wrong?

antoinerg: It returns: `Chrome 68.0.3440 (Windows 10 0.0.0) sankey tests Test hover/click event data: should not output hover/unhover event data when hovermoder is false FAILED`
followed by:
`Expected Object({ event: [object MouseEvent], points: [ Object({ pointNumber: 4, label: 'Solid', color: 'rgba(148, 103, 189, 0.8)', sourceLinks: [ Object({ pointNumber: 60, label: '', color: 'rgba(0,0,96,0.2)', source: <circular reference: Object>, target: Object({ pointNumber: 26, label: 'Thermal generation', color: 'rgba(227, 119, 194, 0.8)', sourceLinks: [ Object({ pointNumber: 63, label: '', color: 'rgba(0,0,96,0.2)', source: <circular reference: Object>, target: Object, value: 787.129, originalIndex: 63, dy: 84.76807405050613, ty: 0, sy: 0, trace: Object, curveNumber: 0 }), Object({ pointNumber: 62, label: '', color: 'rgba(0,0,96,0.2)', source: <circular reference: Object>, target: Object, value: 525.531, originalIndex: 62, dy: 56.595870211663566, ty: 0, sy: 84.76807405050613, trace: Object, curveNumber: 0 }), Object({ pointNumber: 64, label: '', color: 'rgba(0,0,96,0.2)', source: <circular reference: Object>, target: Object, value: 79.329, originalIndex: 64, dy: 8.543156898491352, ty: 0,  ... to be undefined.`

What would be preferred?

alexcjohnson: That looks great, assuming there's a line number along with it too, so you can tell it's the `_hover` on line 618 that caused the failure.

alexcjohnson: (no it doesn't give a line number, but neither do our normal `failTest` invocations... lets leave it as is at least for now)

alexcjohnson:
Nicely done @antoinerg! 💃

PR #2954, opened 2018-08-29
Add barpolar traces

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2810 (not everything item on there, but enough to replace legacy polar's `area` trace type)

Say hello to Polar 2.0's version of 'area' traces: `barpolar`. As discussed previously, this trace type is "really" like a bar chart but in polar coordinates, hence its name. Here's a much-awaited updated version of our [wind rose example](https://plot.ly/javascript/wind-rose-charts/):

peek 2018-08-29 15-50

Moreover, `barpolar` isn't just "really like" `bar` from a user's standpoint, but also from a dev's standpoint: I was able to reuse almost all of `Bar.crossTraceCalc` (formally known as `Bar.setPositions`), its default logic is a mix of `ScatterPolar` and `Bar` routines.

The _hard_ part about this PR was to keep everything DRY. Many path-drawing routines in `plots/polar.js` had to be generalized in order to be used in `BarPolar.plot`. See commits https://github.com/plotly/plotly.js/commit/4d203883d35995047dad83737da5c645defdda99, https://github.com/plotly/plotly.js/commit/95119c455dd86eb87ccf8560bbf724780d8084cb and https://github.com/plotly/plotly.js/commit/facb2f8e57948ac9615d9113c592a279959b2bc3 for more details.

In brief, this first version of `barpolar`, includes support for:
- barmode stacked and overlay
- bar width, offset and base
- per-subplot bargap
- `hovermode: 'closest'`, selections

But, does not implement:
- orientation 'angular'
- on-graph text

cc @alexcjohnson @antoinerg


src/traces/barpolar/layout_attributes.js
```diff
@@ -0,0 +1,40 @@
...
+            'With *stack*, the bars are stacked on top of one another',
+            'With *overlay*, the bars are plotted over one another,',
+            'you might need to an *opacity* to see multiple bars.'
+        ].join(' ')
+    },
+    bargap: {
+        valType: 'number',
+        // TODO is this correct?
+        // I think 0.2 (like for bars) is a little much, maybe 0.1 would work better?
+        dflt: 0.2,
```
etpinard: Ping. `bar` uses 0.2 as default (unless when [_gapless_](https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/traces/bar/layout_defaults.js#L54). I think 0.1 looks before though for `barpolar`. 

image

image


Thoughts?

src/traces/barpolar/plot.js
```diff
@@ -0,0 +1,127 @@
...
+                // can skip calling _module.style during interactions
+                dPath = 'M0,0Z';
+            } else {
+                // TODO is this where we want for to select for barpolar?
+                var rg1 = radialAxis.c2g(s1);
+                var thetagMid = (thetag0 + thetag1) / 2;
+                di.ct = [
+                    xa.c2p(rg1 * Math.cos(thetagMid)),
+                    ya.c2p(rg1 * Math.sin(thetagMid))
+                ];
```
etpinard: Similar to cartesian bars, we select about the tip of bar at mid position (i.e. angle). But perhaps we could do better in polar coordinates?

src/traces/barpolar/plot.js
```diff
@@ -0,0 +1,127 @@
...
+
+                // TODO round up bar borders?
+                // if so, factor out that logic from Bar.plot
```
etpinard: I'm talking about this logic:

https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/traces/bar/plot.js#L97-L132

@alexcjohnson do you think it's worth implementing in polar coordinates?

src/plots/polar/polar.js
```diff
@@ -129,6 +128,8 @@ proto.updateLayers = function(fullLayout, polarLayout) {
             switch(d) {
                 case 'frontplot':
                     sel.append('g').classed('scatterlayer', true);
+                    // TODO add option to place in 'backplot' layer??
+                    sel.append('g').classed('barlayer', true);
```
etpinard: At the moment, bars are drawn above the grid lines.

But perhaps, we could add an option to draw bars below grid lines?

src/lib/angles.js
```diff
@@ -10,25 +10,241 @@
...
+ */
+function angleDist(a, b) {
+    return Math.abs(angleDelta(a, b));
+}
+
+/**
+ * is angle inside sector?
+ *
+ * @param {number} a : angle to test in *radians*
+ * @param {2-item array} sector : sector angles in *degrees*
```
etpinard: In degrees here as before this PR this routine was only used with `layout.polar.sector` as argument which is always in degrees.

But I'm thinking of standardizing all angles in this file to be in radians.

src/lib/index.js
```diff
@@ -729,7 +729,7 @@ lib.ensureSingle = function(parent, nodeType, className, enterFn) {
     var sel = parent.select(nodeType + (className ? '.' + className : ''));
     if(sel.size()) return sel;
 
-    var layer = parent.append(nodeType).classed(className, true);
+    var layer = parent.append(nodeType).classed(className || '', true);
```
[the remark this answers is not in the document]
etpinard: > did anything fail from this or you just got ugly class="undefined"?

Just that.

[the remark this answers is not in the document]
etpinard: > should we just not call classed in this case?

Yes. This sounds like a better way forward than my lazy attempt :arrow_up:

src/lib/angles.js
```diff
@@ -10,25 +10,119 @@
...
+ *
+ * solution taken from: https://stackoverflow.com/a/2007279
+ *
+ * @param {number} a : first angle in *radians*
+ * @param {number} b : second angle in *radians*
+ * @return {number} angular delta in *radians*
+ */
+function angleDelta(a, b) {
+    var d = b - a;
+    return Math.atan2(Math.sin(d), Math.cos(d));
```
alexcjohnson: hmm OK... I'm sure it's not used enough to really matter but this feels heavy just to wrap to `[-π,π]` - can it just be `Lib.mod(d-π, 2π) - π`? Anyway this is equivalent to `wrap180(b - a)` but in radians, right? It could just be done that way too...

src/plots/polar/polar.js
alexcjohnson: > But perhaps, we could add an option to draw bars below grid lines?

True, for polar bars that could be nice to help with relative sizes. Not necessary in v1 though.

src/traces/barpolar/attributes.js
```diff
@@ -0,0 +1,54 @@
...
+    //     editType: 'calc+clearAxisTypes',
+    //     description: 'Sets the orientation of the bars.'
+    // },
+
+    base: barAttrs.base,
+    offset: barAttrs.offset,
+    width: barAttrs.width,
+
+    text: barAttrs.text,
+    // hovertext: barAttrs.hovertext,
```
alexcjohnson: I guess `hovertext` is omitted just because, without on-graph text it's equivalent to `text`? That's fine, but we probably want to give `text` a separate description from the one it's inheriting from `scatter`. I haven't looked at all the other descriptions to see if they make sense as inherited...

If/when we do implement on-graph text, we can probably reuse logic from `pie` to fit text into annular sectors (probably not the outside text logic though).

src/traces/barpolar/hover.js
```diff
@@ -0,0 +1,78 @@
...
+    var radialAxis = subplot.radialAxis;
+    var angularAxis = subplot.angularAxis;
+    var xa = subplot.xaxis;
+    var ya = subplot.yaxis;
+    var inboxFn = subplot.vangles ? isPtInsidePolygon : Lib.isPtInsideSector;
+
+    if(radialAxis.range[0] > radialAxis.range[1]) {
+        xval = -xval;
+        yval = -yval;
+    }
```
alexcjohnson: This confuses me... I see you have a test for it so I guess I can't complain too much, but wouldn't it leave `rVal` unchanged while flipping `thetaVal` 180º?

src/traces/barpolar/hover.js
```diff
@@ -0,0 +1,78 @@
...
+
+    var rVal = Math.abs(radialAxis.g2p(Math.sqrt(xval * xval + yval * yval)));
+    var thetaVal = Math.atan2(yval, xval);
+
+    // TODO add padding around sector to show labels,
+    // when hovering "close to" them
+    var distFn = function(di) {
+        var rBnds = [di.s0, di.s1].map(radialAxis.c2p);
+        var thetaBnds = [di.p0, di.p1].map(angularAxis.c2g).map(Lib.rad2deg);
```
alexcjohnson: 🐎 is there a reason not to do the inverse of these `map`s on `rVal` and `thetaVal`? Then in here I guess all you'd need to do is restrict `thetaBnds` to the same range?

Relatedly, what if your bars wrap around more than 360º, there must be a cut in `thetaVal` somewhere, what if there's a bar that crosses the cut, will its "in bounds" appear to be the entire rest of the circle?

alexcjohnson: Ah I see - yes, looks like `isAngleInsideSector` accounts for this aliasing, so long as none of the conversions you do to `thetaBnds` do any wrapping of their own 🎉 but this requires periodicity of 360, hence this mapping is required to be on this side. Well OK, I guess we won't have enough data here to worry *that* much about 🐎 (so this is all nonblocking) - but if we do want to improve this it would probably just require modifying `inboxFn` to work with the calcdata periodicity rather than 360.

src/traces/barpolar/layout_attributes.js
alexcjohnson: 90º bars are unusually large - what value looks best when each bar is only ~15-30º?

src/traces/barpolar/plot.js
alexcjohnson: I think this is still correct - the tip at mid-position is still the location most representative of the data embodied by this bar.

src/traces/barpolar/plot.js
```diff
@@ -0,0 +1,127 @@
...
+
+                dPath = pathFn(rp0, rp1, thetag0, thetag1);
+            }
+
+            Lib.ensureSingle(bar, 'path')
+                .style('vector-effect', 'non-scaling-stroke')
```
alexcjohnson: do we need a `stroke-miterlimit` here? `pie` uses [`stroke-linejoin: 'round'`](https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/traces/pie/plot.js#L34-L36) as you can see if you give pies a very wide linewidth... but I think we probably want more precision here, so a regular mitered corner with a large enough `stroke-miterlimit` that the 90º corners are not clipped, but small enough that two neighboring bars that meet at the center with wide line width, no matter what angle the bars subtend, will never show crossing tips.

src/traces/barpolar/hover.js
```diff
@@ -0,0 +1,78 @@
...
+    var cdi = cd[index];
+    var rg = radialAxis.c2g(cdi.s1);
+    // TODO include offset here?
+    var thetag = angularAxis.c2g(cdi.p);
+    var xp = xa.c2p(rg * Math.cos(thetag));
+    var yp = ya.c2p(rg * Math.sin(thetag));
+
+    // TODO use 'extents' like in Bar.hover?
+    pointData.x0 = pointData.x1 = xp;
+    pointData.y0 = pointData.y1 = yp;
```
alexcjohnson: This is good, but I wonder if we can flip the hover label so it's outside the bar pointing toward it? For the most part this means to the left if the bar is on the left half of the plot, unless the bar is inverted... so here ideally the label would be on the left for both purple bars:
barpolar-hover

test/image/mocks/polar_funky-bars.json
```diff
@@ -0,0 +1,63 @@
...
+      "domain": {"row": 1, "column": 0},
+      "radialaxis": {
+        "categoryorder": "category descending"
+      }
+    },
+    "polar4": {
+      "domain": {"row": 1, "column": 1},
+      "angularaxis": {"direction": "clockwise"},
+      "radialaxis": {"angle": 90, "side": "counterclockwise"}
+    },
```
alexcjohnson: Good lord those are atrocious. But yes, I suppose we should support all of this. Nicely done. 🙈 
Of course the next axis type I'm going to ask for (specifically for polar because it maps data -> area, though occasionally someone wants this for cartesian) is a `sqrt` mapping... but that can wait.

test/jasmine/tests/barpolar_test.js
```diff
@@ -0,0 +1,326 @@
...
+            y: 200,
+            extraText: 'r: 2<br>θ: 360°',
+            color: '#1f77b4'
+        }
+    }]
+    .forEach(function(specs) {
+        it('should generate correct hover labels ' + specs.desc, function(done) {
+            run(specs).catch(failTest).then(done);
+        });
+    });
```
alexcjohnson: Great test suite - that's a heck of a lot of cases covered!

src/lib/angles.js
alexcjohnson: > But I'm thinking of standardizing all angles in this file to be in radians.

That sounds like a good idea, would probably simplify quite a few things. BTW `wrap360` is really just `Lib.mod`, right? And I'm thinking perhaps that we want a version of `Lib.mod` that restricts to +/- 1/2 the value, rather than 0-1x as `Lib.mod` does now - which would then cover both `angleDelta` and `wrap180`.

src/traces/barpolar/hover.js
etpinard: Good idea. Is this what you had in mind?

peek 2018-08-31 14-46


-> https://github.com/plotly/plotly.js/pull/2954/commits/e6c4dadecc2ce0a2d79d1f9c86242aa21a7acd75

src/lib/index.js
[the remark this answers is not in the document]
etpinard: Improved in https://github.com/plotly/plotly.js/pull/2954/commits/a32484a60ca6bad98cb4483d6489c6f1e234ca38

src/traces/barpolar/attributes.js
etpinard: > I guess hovertext is omitted just because, without on-graph text it's equivalent to text? 

Exactly. 

> That's fine, but we probably want to give text a separate description from the one it's inheriting from scatter

Good call 

> I haven't looked at all the other descriptions to see if they make sense as inherited...

`barpolar` width, base and offset aren't exactly equivalent to the ones for `bar`

Improvements in https://github.com/plotly/plotly.js/pull/2954/commits/01e15b401f3e22d930848822fbce9b62a9ff499b

src/traces/barpolar/hover.js
etpinard: > but wouldn't it leave rVal unchanged while flipping thetaVal 180º?

Good call. That's much easier to read -> https://github.com/plotly/plotly.js/pull/2954/commits/971c6f6386c2c73a7dbd993d5953be842eb784e5

src/lib/angles.js
etpinard: Standardized to radians in https://github.com/plotly/plotly.js/pull/2954/commits/e93897473bec9cf27f7e4a63cf85ce55ac4de1b6



> BTW wrap360 is really just Lib.mod, right? And I'm thinking perhaps that we want a version of Lib.mod that restricts to +/- 1/2 the value, rather than 0-1x as Lib.mod does now - which would then cover both angleDelta and wrap180.

I'll have to look at this in more detail :thinking:

src/traces/barpolar/plot.js
etpinard: Playing around with `stroke-miterlimit`, I couldn't get the mocks to produce a diff. So it's a little hard to tell if it improves the bar's appearance. Perhaps this is due to us using (old) nw.js. It might be worth trying different  `stroke-miterlimit` value in orca.

src/traces/barpolar/layout_attributes.js
etpinard: A few more cases:

at 0.1

image

at 0.2

image


---- 

0.1

image


0.2

image

src/traces/barpolar/hover.js
alexcjohnson: OK, I still don't get why reversed *radial* range needs a change in *angle* (please don't bother explaining it 🙈 ) but it's clear anyway - the commit that I commented on didn't have the comment above this, that helps a lot.

src/traces/barpolar/hover.js
alexcjohnson: Ha sorry, I missed that you *did* clean this up in https://github.com/plotly/plotly.js/pull/2954/commits/7ae3b7cd11211e52467a4a70218fb3c859b70356, and then you removed the last `map` inside `distFn` in https://github.com/plotly/plotly.js/pull/2954/commits/e93897473bec9cf27f7e4a63cf85ce55ac4de1b6 - beautiful.

src/traces/barpolar/plot.js
alexcjohnson: Here is the effect I was worried about:  two slices meet at a sharp point, but the strokes end up making them look like they cross. For the angles in the codepen below it only seems to be a problem if I insert a large `stroke-miterlimit`, ~7 or above:
https://codepen.io/alexcjohnson/pen/WgOpqP
screen shot 2018-09-04 at 12 19 44 pm
but for other angles (somewhat wider but not too wide) the [default value of 4](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit) might be problematic. For scatter traces I [set it to 2](https://github.com/plotly/plotly.js/blob/854bbc931c218be579850b45a04b666d40fc53eb/src/traces/scatter/plot.js#L39) mostly so that very spiky traces don't look like they peak higher than they really do. I suspect 2 would be a good value here too, it's conceivable the default of 4 is fine or has such uncommon failures that we wouldn't care.

src/traces/barpolar/layout_attributes.js
etpinard: :hammer_and_pick: `0.1`  in -> https://github.com/plotly/plotly.js/pull/2954/commits/6494302c665e1ef18a7c798310f5423fd8f1318e

src/traces/barpolar/plot.js
etpinard: `2` seems like to way to go -> https://github.com/plotly/plotly.js/pull/2954/commits/183b73490bf5eabc1914efefb451d731e48771b8

test/image/mocks/polar_bar-overlay.json
```diff
@@ -6,12 +6,15 @@
...
   }, {
     "type": "barpolar",
-    "r": [1, 2, 3],
-    "theta": [180, 90, 360],
-    "opacity": 0.7
+    "r": [3, 3],
+    "theta": [0, 45],
+    "opacity": 0.7,
+    "marker": {
+      "line": {"width": 5}
```
etpinard: which looks like

image

closeup

image

with the default `stroke-mitelimit`.

src/traces/histogram/calc.js
```diff
@@ -176,15 +176,17 @@ module.exports = function calc(gd, trace) {
...
 
-            // pts and p0/p1 don't seem to make much sense for cumulative distributions
+            // setup hover and event data fields,
+            // N.B. pts and "hover" positions ph0/ph1 don't seem to make much sense
+            // for cumulative distributions
             if(!cumulativeSpec.enabled) {
                 cdi.pts = inputPoints[i];
                 if(uniqueValsPerBin) {
-                    cdi.p0 = cdi.p1 = (inputPoints[i].length) ? pos0[inputPoints[i][0]] : pos[i];
+                    cdi.ph0 = cdi.ph1 = (inputPoints[i].length) ? pos0[inputPoints[i][0]] : pos[i];
```
etpinard: ping @alexcjohnson I hope you don't mind.

The bar corners p0/p1/s0/s1 are now set during `Bar.crossTraceCalc` (moved from `Bar.plot`) so that we don't have to duplicate logic in `BarPolar.plot`. In turn, we needed to rename histogram p0/p1 to not break histogram hover.

src/lib/angles.js
etpinard: In https://github.com/plotly/plotly.js/pull/2954/commits/8575b27ad369921ab7e842493fb68d6fb566efea, I replaced `wrap360` with its `mod` equivalent - which turns out well as we no longer need to convert thing to degrees `mod(v, 2 * Math.PI)` works just fine.

In the same commit, I added a "restricts to +/- 1/2 the value" version of `mod` that I named `Lib.modHalf` (please let me know if you can think of a better name for this thing).

src/lib/angles.js
etpinard: Using `mod` appears to work fine. Updated in -> https://github.com/plotly/plotly.js/pull/2954/commits/939bdad1d24a2ca99bab9844d312ee65c514f7a7

src/lib/mod.js
```diff
@@ -12,7 +12,23 @@
...
  */
-module.exports = function mod(v, d) {
+function mod(v, d) {
     var out = v % d;
     return out < 0 ? out + d : out;
+}
+
+/**
+ * sanitized modulus function that always returns in the range [-1.5*d, 1.5*d]
+ * rather than (-d, 0] if v is negative
```
alexcjohnson: Behavior looks correct, but shouldn't the doc be:
```
sanitized modulus function that always returns in the range[-d, d]
rather than (-2*d, 0] if v is negative and [0, d) if v is positive
```
That said, since "half" is in the name I might have used d2 as the argument - since that's still in the end the value we're shifting by... not a big deal though, if it makes more sense to you this way feel free to keep it.

etpinard: > I might have used d2 as the argument

Good call. Done in https://github.com/plotly/plotly.js/pull/2954/commits/d2484f39265223943f199658500f2218e9eb929b

Bonus that made the doc correct.

src/lib/angles.js
```diff
@@ -37,8 +37,7 @@ function isFullCircle(aBnds) {
  * @return {number} angular delta in *radians*
  */
 function angleDelta(a, b) {
-    var d = b - a;
-    return Math.atan2(Math.sin(d), Math.cos(d));
+    return mod((b - a) - PI, twoPI) - PI;
```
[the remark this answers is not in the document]
etpinard: Oops. I shouldn't have read that comment of yours all the way to end.

Done in -> https://github.com/plotly/plotly.js/pull/2954/commits/d2484f39265223943f199658500f2218e9eb929b

src/lib/mod.js
alexcjohnson: > that made the doc correct

Almost :neckbeard: 
`[-1.5*d, 1.5*d]` -> `[-0.5*d, 0.5*d]` or `[-d/2, d/2]`

alexcjohnson:
Very nice 💃

PR #2956, opened 2018-08-30
Bump regl trace deps

etpinard:
cc @alexcjohnson 

The first commits applies https://github.com/a-vis/regl-splom/pull/9 and the other essentially reverts https://github.com/plotly/plotly.js/pull/2783

Note that `gl-text@1.1.6` includes an additional commit (-> https://github.com/a-vis/gl-text/commit/95036634910f6801b0eb583a58ab286c98b0c500) that wasn't bundled before. @dy do you remember what https://github.com/a-vis/gl-text/commit/95036634910f6801b0eb583a58ab286c98b0c500 was for?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,29 +35,6 @@
         "commander": "^2.15.1"
       }
     },
-    "@etpinard/gl-text": {
-      "version": "1.1.6",
-      "resolved": "https://registry.npmjs.org/@etpinard/gl-text/-/gl-text-1.1.6.tgz",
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Would be nice to know what https://github.com/a-vis/gl-text/commit/95036634910f6801b0eb583a58ab286c98b0c500 was about @dy - and potentially add a test that requires it assuming it was a bugfix - but as it doesn't break any of our existing tests I'm happy to take this as is.
💃

Issue #1217, opened 2016-12-02
add _real_ stacked area charts [feature request]


salim-b:
I'm filing this issue as a gathering point for the feature request of **_real_ stacked area charts**.

The current solution to create stacked area charts is [to plot cumulative variables](https://plot.ly/javascript/filled-area-plots/#stacked-area-chart) which has multiple drawbacks:
- Correct (i.&nbsp;e. non-cumulative) hover values have to be set manually by using `text: '...'`.
- The resulting chart [is incorrect if some traces are shown and others hidden (by clicking on legend items)](https://github.com/plotly/plotly.js/issues/344).

A real solution would be to have an argument like `layout = {linemode: 'stack'}`, [same as there is for bar charts](https://plot.ly/javascript/bar-charts/#stacked-bar-chart).

I'm aware that @etpinard [stated some time ago](https://github.com/plotly/plotly.js/issues/344#issuecomment-198549402):
> We are planning on adding area charts to our list of trace types at some point this year. You are not the first person to ask for this feature.

Therefore I hope opening this issue won't be seen as an annoying harassment! Since it is an often requested feature, I think it important to have a place for users like me to gather all relevant information and potential progress on this.

Some more information:
- relevant bug reports:
  - `plotly.js`: ~~<https://github.com/plotly/plotly.js/issues/344>~~
  - `plotly` for `R`: ~~<https://github.com/ropensci/plotly/issues/686>~~ and ~~<https://github.com/ropensci/plotly/issues/810>~~
- [Generating a stacked area chart in `ggplot2`](http://chrisalbon.com/r-stats/stacked-area-graph.html) and then converting it to plotly using `ggplotly` doesn't work as supposed either (the plot doesn't get rescaled correctly if traces in between are unselected, there will just be space left).
- proposed workaround by [@etpinard](https://github.com/plotly/plotly.js/issues/344#issuecomment-198549402) (in `JS`): <https://codepen.io/etpinard/pen/yOgdOb>

treadmillian:
What are the timescales for this requested feature?  This is one reason why I've turned to HightCharts which does this perfectly.

Reubentrapdoor:
Been monitoring this one for a long time hoping to see some progress, all the workarounds I've seen are way too buggy to use in any production environment

nicolaskruchten:
[...]

Issue #1397, opened 2017-02-20
This bubble charts isn't working


suribabu39:
Hi, 
For certain range of values(x, y, size in trace) bubble chart is not plotting, Maybe it is plotting but clearly not visible. For most of range it is working but for certain range of values it is not working. Is there any constraint on the value ranges or should I maintain a ratio?
One such example where it doesn't plot any chart is here (http://codepen.io/suribabu_95/pen/egwPrG?editors=0010)

Thank You

etpinard:
Strange.

Back in `v1.7.0`, your bubble chart would've looked like http://codepen.io/etpinard/pen/VPJVyv

At first glance, it looks like very large bubbles are throwing off the marker sizing routine.

> Is there any constraint on the value ranges or should I maintain a ratio?

Yes. Try using `marker.sizeref` as in http://codepen.io/etpinard/pen/VPJVyv

etpinard:
This bug is a side-effect of (the very controversial) https://github.com/plotly/plotly.js/pull/353

etpinard:
More closely, the problem lies inside the autorange [routine](https://github.com/plotly/plotly.js/blob/d8441a080aba8269ce20260a69a2bf8f68fb13ba/src/plots/cartesian/axes.js#L235-L247): where very large bubbles make

```js
dp = ax._length - minpt.pad - maxpt.pad;
```

always be negative and hence excluded from the the autorange computation. Perhaps `dp` should be constrained?

etpinard:
Referencing https://github.com/plotly/plotly.js/issues/1352 which also highlights `getAutoRange` issues.

etpinard:
Fixed _as best we can_ in https://github.com/plotly/plotly.js/pull/2960/commits/1f4898c7b638aa6711c6abbb6691cfe3a4024e1a

PR #2960, opened 2018-09-01
Stacked Area Charts

alexcjohnson:
We held out a long time on this one, but stacked area charts are finally coming to plotly.js.

The API is as discussed in #1217:
- Provide matching `stackgroup` attributes to some `scatter` traces and they become a stack.
- There are no plot-wide stacking attributes; stack-wide attributes are in the trace definitions, and we'll take a value for each attribute from the first trace in the stack that contains that attribute, *visible or not* (so the stack doesn't fall apart if you hide the first trace). This is different from, and more powerful than, how we describe bar stacking/grouping - and should be reviewed with an eye toward eventually using a similar framework for bars.
- The data for all traces in the stack are sorted by position, and gaps in each trace are filled in either with zeros or interpolations
- The one item from #1217 I did *not* include here is `stackgaps: 'interrupt'`. That's going to require some finicky drawing code, particularly if we want to support arbitrary `line.shape` so I'll leave it for later. But `'infer zero'` (default) and `'interpolate'` are included here.
- Another open item is to improve hover info. What I did here matches stacked bars, but both of them, particularly if you normalize the results, would benefit from more options - normalized vs raw data, (sub)totals.

In order to make it work well in various edge cases I made a number of preparatory changes:
- `Lib.sort` https://github.com/plotly/plotly.js/commit/c87ccb33f7232727dbf7bb32589c3767fb184238 wraps the built-in `Array.sort` with a check for whether the array is already perfectly sorted (or perfectly reversed), that for arrays of length 1e5+ can be a 10x or better speedup for already-sorted arrays, and should have very little penalty for unsorted arrays. For stacked area I expect the vast majority of the time the data will already be sorted, so that's why I implemented this now and this is the only place I used it, but I bet there are other places it would be useful as well.
- Some edge case improvements in autorange https://github.com/plotly/plotly.js/commit/1f4898c7b638aa6711c6abbb6691cfe3a4024e1a - I changed a few baseline images (and one mock), I hope you'll agree these were actually incorrect before.
- Better ordering of hover labels when traces have matching data (such as in a bar or area stack when one trace is zero) - try to preserve the stacking order https://github.com/plotly/plotly.js/commit/2fde3dc183522879aaacca41f9aa6a453ea31162
- Continue lines off the edge toward invalid log values https://github.com/plotly/plotly.js/commit/68b489d3e3e0e0a791c81dd8bf2fd43c8bb17c2e - I think I hadn't done this before (for scatter) out of caution lest we draw something misleading, I opted to just not draw the line at all. But particularly with fills, and even more so with stacked fills, this gets confusing and misleading as the fills would just connect across the missing point(s). I opted to draw these lines straight toward the edge if one dimension went invalid (since in principle they're going infinitely far away) or at a slope of 1 on a log/log plot if both dimensions go invalid simultaneously. Note that there are cases here where a separate point will move across these lines if you flip between linear and log axes, but that was already possible with finite data; this is just an extreme case of the same. (note the `axes_range_type` baseline change belongs in this commit but I put it in the autorange commit instead)

cc @etpinard @antoinerg @nicolaskruchten


src/plots/cartesian/autorange.js
```diff
@@ -95,76 +102,83 @@ function getAutoRange(gd, ax) {
...
+                        maxbest = maxpt;
+                        mbest = dv / dp;
+                    }
+                }
+                else if(dv / axLen > mbest) {
+                    // in case of padding longer than the axis
+                    // at least include the unpadded data values.
+                    minbest = {val: minpt.val, pad: 0};
+                    maxbest = {val: maxpt.val, pad: 0};
+                    mbest = dv / axLen;
```
alexcjohnson: @etpinard we had an issue recently where someone used ridiculously large markers and autorange broke, didn't we? I can't find it, but this fixes that case, by trying again with no padding.

test/image/mocks/scatter_fill_corner_cases.json
```diff
@@ -75,7 +75,6 @@
     {
       "x": [1.5],
       "y": [1.25],
-      "fill": "tonexty",
```
alexcjohnson: Oh this change I think is actually required due to a change in the stacked area commit https://github.com/plotly/plotly.js/pull/2960/commits/be38e932ec8e0ac350a6f21d2d8e6fabf4f8cfed#diff-33c02cd37e7a4c951059a3c93221ac4eR175 - we were accidentally treating a length-1 trace as filling to itself (since its start and end points are the same!) but we shouldn't do that... therefore this trace, since it's the first on its subplot, should interpret `'tonexty'` as `'tozeroy'`.

src/traces/scatter/calc.js
```diff
@@ -138,8 +249,34 @@ function calcMarkerSize(trace, serieslen) {
...
+ * note that scatter and scattergl each get their own first trace
+ * note also that I'm doing this during calc rather than supplyDefaults
+ * so I don't need to worry about transforms, but if we ever do
+ * per-trace calc this will get confused.
+ */
+function setFirstScatter(fullLayout, trace) {
+    var subplotAndType = trace.xaxis + trace.yaxis + trace.type;
+    var firstScatter = fullLayout._firstScatter;
+    if(!firstScatter[subplotAndType]) firstScatter[subplotAndType] = trace.uid;
+}
```
alexcjohnson: In fact, `scatter_fill_corner_cases` top subplots were *also* prevented from filling to zero with `'tonexty'` because only one subplot could have the "first scatter" trace on it. This commit :hocho: `gd.firstscatter` and replaces it with one trace (uid) per subplot, attached to `fullLayout`. The stacked area mocks 🔒 this.

src/traces/scatter/plot.js
```diff
@@ -416,12 +424,24 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
...
 
-        if(showMarkers) {
-            markerFilter = (trace.marker.maxdisplayed || trace._needsCull) ? visFilter : Lib.identity;
-        }
+        if(showMarkers || showText) {
+            var showFilter = Lib.identity;
+            // if we're stacking, "infer zero" gap mode gets markers in the
+            // gap points - because we've inferred a zero there - but other
+            // modes (currently "interpolate", later "interrupt" hopefully)
+            // we don't draw generated markers
```
alexcjohnson: @etpinard @nicolaskruchten do you agree with this choice? It only applies to points we generate in one trace to match the positions from another trace - those are the "gap points"

src/traces/scatter/stack_defaults.js
```diff
@@ -0,0 +1,101 @@
...
+
+                        // TODO: should we copy the new value to the first trace instead?
+                        // if we did that, the editor could show controls in the
+                        // first trace always; otherwise it would only show controls
+                        // in a later trace, if that trace happens to already have a value...
+                        // I think it's probably better to ignore the editor for this
+                        // purpose though, as it's probably going to have to use some
+                        // special logic anyway, it will likely want to just
+                        // take the value out of layout._scatterStackOpts, set the
+                        // new value into the first trace, and clear all later traces.
```
alexcjohnson: @nicolaskruchten I think we already settled on what I have here, but before I 🔪 this comment just take a last look at it please.

src/traces/scatter/stack_defaults.js
```diff
@@ -0,0 +1,101 @@
...
+                        if(!firstTrace) {
+                            delete groupOpts.traces[0][attr];
+
+                            // orientation can affect default fill of previous traces
+                            if(isOrientation) {
+                                for(var j = 0; j < groupOpts.traces.length - 1; j++) {
+                                    var trace2 = groupOpts.traces[j];
+                                    if(trace2._input.fill !== trace2.fill) {
+                                        trace2.fill = groupOpts.fillDflt;
```
etpinard: I forgot if we discussed this before, but is there a use case for having a stacked area trace with a `fill` setting other than `tonexty` (`tonextx` for horizontal stacks)?

If not, we could skip `coerce('fill')` for stacked area traces and infer the `fill` value later on in scatter `plot.js` and `line_points.js` and :hocho: this hacky block.

src/plots/cartesian/autorange.js
alexcjohnson: Ah, yep - I thought it was more recent but that seems like the one. Are you happy with this solution, enough that we should consider #1397 closed with this PR?

src/plots/cartesian/axes.js
```diff
@@ -2411,11 +2413,12 @@ function swapAxisGroup(gd, xIds, yIds) {
...
 
-    var allAxKeys = Object.keys(xFullAxes[0]),
-        noSwapAttrs = [
-            'anchor', 'domain', 'overlaying', 'position', 'side', 'tickangle'
-        ],
-        numericTypes = ['linear', 'log'];
+    var allAxKeys = Object.keys(axAttrs);
+
+    var noSwapAttrs = [
+        'anchor', 'domain', 'overlaying', 'position', 'side', 'tickangle', 'editType'
```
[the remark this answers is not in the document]
alexcjohnson: > How can axAttrs with editType get here?

Because I'm using the schema instead of `fullLayout` to find the relevant attributes now. The reason for this change is that not all the attributes that should be swapped are always coerced - the one that prompted this change was `rangemode`, since I changed its behavior here, but there are others I'm sure.

It wouldn't be a big deal to swap a non-attribute like `editType`, but in principle we shouldn't, as we won't swap other non-attributes.

Of course it's not clear anyone is even using `Axes.swap` anymore but it's here 😐

src/traces/scatter/calc.js
```diff
@@ -20,36 +20,142 @@ var arraysToCalcdata = require('./arrays_to_calcdata');
...
+            cdi.id = String(ids[i]);
         }
     }
 
     arraysToCalcdata(cd, trace);
     calcColorscale(trace);
     calcSelection(cd, trace);
 
-    gd.firstscatter = false;
+    if(stackGroupOpts) {
```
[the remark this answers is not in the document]
alexcjohnson: That particular block perhaps, if we decide we want a `stackgaps` analog for bars. Which begs the question though... should it be called something that doesn't have "stack" in the name? What if we want interpolation for grouped bars? Seems like an odd choice to use bars with interpolation - and for bars "infer zero" and "interrupt" would be equivalent visually, though I guess hover would be different - but that might not be a strong enough argument to omit it.

Some of the `cross_trace_calc` logic would need to be different, as I don't see anything reasonable to do with multiple matching positions and bars (other than potentially summing the values, but that's a different feature altogether)

src/traces/scatter/calc.js
```diff
@@ -138,8 +249,34 @@ function calcMarkerSize(trace, serieslen) {
...
+
+function getStackOpts(trace, fullLayout, xa, ya) {
+    var stackGroup = trace.stackgroup;
+    if(!stackGroup) return;
+    var stackOpts = fullLayout._scatterStackOpts[xa._id + ya._id][stackGroup];
+    var stackAx = stackOpts.orientation === 'v' ? ya : xa;
+    // Allow stacking only on numeric axes
+    // calc is a little late to be figuring this out, but during supplyDefaults
+    // we don't know the axis type yet
+    if(stackAx.type === 'linear' || stackAx.type === 'log') return stackOpts;
```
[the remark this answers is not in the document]
alexcjohnson: > I forgot how we handle stacking dates and categories for bars.

Not very gracefully 🤦‍♂️ 
screen shot 2018-09-04 at 7 30 59 pm

> Is this for theoretical reasons? Or is this something that could be done in a future iteration?

Theoretical. Zero is meaningless on non-numeric axes, so there's no rational baseline by which to convert a value to a "size" that would make sense to stack on another. I suppose in principle we could accept numeric data for all but the first trace in a stack... which would mean "number of categories" or "number of milliseconds" (and in the date case perhaps we could also accept some units so you could use data in weeks or hours or whatever) but that would be a totally different feature that we should wait on a real use case before implementing.

nicolaskruchten:
So from a high-level API standpoint, why do we want `stackgroup` again? Is it just so as to match a potential future equivalent for `bar`? Because as a standalone API it's kind of ungainly, and I can't imagine a use-case for have some areas stacked and some not in the same plot... ?

alexcjohnson:
> So from a high-level API standpoint, why do we want `stackgroup` again? Is it just so as to match a potential future equivalent for `bar`? Because as a standalone API it's kind of ungainly, and I can't imagine a use-case for have some areas stacked and some not in the same plot... ?

It's unusual for sure, but I wouldn't want to rule it out. What if you have one stack series for data and another for fits? Then one stack would need its fill removed, since they're overlapping. Or prediction/extrapolation - these might not overlap but still you might want different styling for corresponding items in each stack. Or two back-to-back stacks, like those plots that have male on one side and female on the other, with the axis in the middle (we could manage that one with an analog of `barmode: 'relative'`, or perhaps even better two axes with a constraint so you don't need to flip your data... but you see the point)

> What do you think about adding a `'stack'` flag to scatter `mode` to make it easier to toggle stacked areas on and off?

`mode` is otherwise all about *how* to draw the series, not *where* to draw it... and the one bit of *how* that stacking impacts (`fill`) isn't even part of `mode`.

But, perhaps both of these concerns could be assuaged by making a boolean `stack` attribute, then giving `stackgroup` a default value but only coercing it when `stack` is `true`? (and for completeness, if you provide only a `stackgroup` let `stack` default to `true`). That way the usual behavior would be to just use the boolean but the full flexibility would still be available (if perhaps buried in the UI)

src/traces/scatter/stack_defaults.js
alexcjohnson: There's certainly a use case for no fill. Other than that no I don't see a use case, but I'm not sure that helps us remove this block if we still want to support `fill: 'none'`.

test/image/mocks/log_lines_fills.json
```diff
@@ -0,0 +1,33 @@
...
+        "x": [-1, 1.5, 2, 0, 1, 3, 4, 5, 6, 1, -1, -1, 3, 3, 0, 7, 7, -1, -1],
+        "y": [-1, 1.5, 1, -1, -1, 1, 1, 0, 2, 2, 0, 2, 3, 4, 5, 6, -2, -2, -1],
+        "mode": "markers+lines", "fill": "toself", "yaxis": "y2"
+    }, {
+        "x": [-1, 1.5, 2, 0, 1, 3, 4, 5, 6, 1, -1, -1, 3, 3, 0, 7, 7, -1, -1],
+        "y": [-1, 1.5, 1, -1, -1, 1, 1, 0, 2, 2, 0, 2, 3, 4, 5, 6, -2, -2, -1],
+        "mode": "markers+lines", "fill": "toself", "xaxis": "x2", "yaxis": "y2"
+    }, {
+        "x": [0.01, 0.1], "y": [0.01, 0.1],
+        "mode": "markers", "xaxis": "x2", "yaxis": "y2"
```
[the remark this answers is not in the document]
alexcjohnson: good call -> extended the title https://github.com/plotly/plotly.js/pull/2960/commits/eed76c60ef9144957e59f26e1e0212da9cc3bc4f

src/lib/search.js
```diff
@@ -113,3 +113,44 @@ exports.roundUp = function(val, arrayIn, reverse) {
...
+ * Tweak to Array.sort(sortFn) that improves performance for pre-sorted arrays
+ *
+ * Motivation: sometimes we need to sort arrays but the input is likely to
+ * already be sorted. Browsers don't seem to pick up on pre-sorted arrays,
+ * and in fact Chrome is actually *slower* sorting pre-sorted arrays than purely
+ * random arrays. FF is at least faster if the array is pre-sorted, but still
+ * not as fast as it could be.
+ * Here's how this plays out sorting a length-1e6 array:
+ *
+ * Calls to Sort FN  |  Chrome bare  |  FF bare  |  Chrome tweak  |  FF tweak
```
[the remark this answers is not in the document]
alexcjohnson: sure. ➕versions -> https://github.com/plotly/plotly.js/pull/2960/commits/13b54daad8578286d5625cf2b47370078938ae0d

src/plots/plots.js
```diff
@@ -1178,6 +1181,14 @@ plots.supplyTraceDefaults = function(traceIn, traceOut, colorIndex, layout, trac
 
         plots.supplyTransformDefaults(traceIn, traceOut, layout);
     }
+    else if(_module && Registry.traceIs(traceOut, 'alwaysSupplyDefaults')) {
+        // Some types need *something* from supplyDefaults always, even if
+        // visible: false. Looking at you scatter: stack options even from
+        // hidden traces can control other traces in the stack.
+        // These types should bail out ASAP if visible is false.
+        // But we don't need any other cross-module attrs ^^ in this case.
+        _module.supplyDefaults(traceIn, traceOut, defaultColor, layout);
```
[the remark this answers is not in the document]
alexcjohnson: I don't think so - this `else if` is in reference to an `if(visible)` above. That change in #2860 ensures we still call `plot` for these traces and base plot modules but `supplyDefaults` still bails.

src/traces/scatter/attributes.js
```diff
@@ -72,6 +72,69 @@ module.exports = {
             'See `y0` for more info.'
         ].join(' ')
     },
+
+    stackgroup: {
```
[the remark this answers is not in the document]
alexcjohnson: (discussed in https://github.com/plotly/plotly.js/pull/2960#issuecomment-418584839)

nicolaskruchten:
OK, I'll buy the "back to back stacks" argument :)

Could we make sure the documentation clearly explains whether or not stack normalization applies across or within subplots please? I don't know the answer but I'd like to and I think we should canonicalize it in the docs!

src/traces/scatter/plot.js
nicolaskruchten: Yup.

nicolaskruchten:
I think we can live without an extra `stack` attribute personally :)

src/traces/scatter/stack_defaults.js
nicolaskruchten: Yup.

src/plots/cartesian/axes.js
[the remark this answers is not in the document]
etpinard: Makes sense :+1:

src/traces/scatter/stack_defaults.js
alexcjohnson: Great, 🔪 the TODO in https://github.com/plotly/plotly.js/pull/2960/commits/209235452b6dbe5dc57d255b06aa5995a32c328e

src/plots/cartesian/autorange.js
etpinard:
```
var trace1 = {
  x: [10000, 0, 30, 40000],
  y: [6000, 0, 80, 600],
  mode: 'markers',
  marker: {
    size: [70000, 0, 30, 9000],
    //sizeref: 100
  }
};

var data = [trace1];

var layout = {
  title: 'Marker Size',
  showlegend: false,
  height: 600,
  width: 600
};

```

now renders as 

image

which I would call _correct_.

With a "reasonable" `marker.sizeref` of 500, we get

image

where `getAutoRange` did its job.

src/plots/plots.js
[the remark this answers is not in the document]
etpinard: Ok, my mistake I meant to write _gets called even on traces with `visible:'legendonly'`_ which of course predates #2860 (my shameless reference to a PR I made did go so well here :face_with_head_bandage: 

So, w/o this block, stack options even from hidden traces (i.e. legendonly traces) can _still_ control other traces in the stack. 

In other words, do we need to extra piece of logic "just for" `visible:false`?

src/traces/scatter/cross_trace_calc.js
```diff
@@ -0,0 +1,185 @@
...
+        // values at the same position and this trace only has one value there,
+        // we just duplicate that one value rather than insert a zero.
+        // We also make it look like a real point - because it's ambiguous which
+        // one really is the real one!
+        if(index && position === calcTrace[index - 1][posAttr]) {
+            var prevEntry = calcTrace[index - 1];
+            newEntry.s = prevEntry.s;
+            // TODO is it going to cause any problems to have multiple
+            // calcdata points with the same index?
+            newEntry.i = prevEntry.i;
```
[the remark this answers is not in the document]
etpinard: > To be clear, what's happening here is one trace has multiple values at the same position; other traces get their value(s) at that position duplicated out to the same multiplicity.

Thanks for info. I forgot about that "more data issues" section in #1217. This line makes sense now :ok_hand: Conversation resolved :white_check_mark:

src/traces/scatter/stack_defaults.js
etpinard: > There's certainly a use case for no fill. 

Absolutely. Good point!

--------

Referecing

> mode is otherwise all about how to draw the series, not where to draw it... and the one bit of how that stacking impacts (fill) isn't even part of mode.

from https://github.com/plotly/plotly.js/pull/2960#issuecomment-418584839 (your reply to `mode: 'stack'`)

That _how_ vs _where_ argument makes sense to me :ok_hand: 

But it does suggest that the `fill` attribute isn't optimal. Here's a proposal for v2... maybe v3 :smirk: 

- :hocho: `fill: 'none'`
- add `'fill'` flag in scatter `mode`
- add that `'fill'` flag the `mode` default whenever `stackgroup` is set
- coerce scatter `fill` (that maybe we should call `fills` similar to the `markers` and `lines` attribute containers) only if `'fill'` is part of `mode`
- make `tonexty` the only valid `fills` value when `stackgroup` is set (`tonextx` for horizontal stacks). `Lib.log` something if some other value is set.

I think implementing this proposal would make this hacky logic obsolete.

etpinard: oh and by the way, I'm now convinced that adding a `'stack'` flag in mode isn't worth it :ok_hand:

src/traces/scatter/attributes.js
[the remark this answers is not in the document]
etpinard: Resolved.

src/plots/plots.js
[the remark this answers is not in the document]
alexcjohnson: >  do we need to extra piece of logic "just for" `visible:false`?

Hmm, good question. I could see arguments either way. On the one hand you could say `visible: false` should be the equivalent of removing the trace altogether, just keeping it available in case you want it later. On the other hand I could imagine someone making an app with its own legend or equivalent outside of plotly.js, and being confused that when they hide the bottom stack item the stack config is lost...

etpinard:
Down to :two: unresolved comments:

- https://github.com/plotly/plotly.js/pull/2960#discussion_r215731745
- https://github.com/plotly/plotly.js/pull/2960#discussion_r215728099

src/plots/plots.js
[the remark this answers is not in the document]
alexcjohnson: Alright, you've won me over, both have some arguments in their favor but ignoring `visible: false` gives cleaner, nicer code and is probably more consistent with the rest of our logic. https://github.com/plotly/plotly.js/pull/2960/commits/9fdf5226430c756a8844847e1f2471e76e60a72c

etpinard:
Nicely done :dancer:

Issue #2229, opened 2018-01-02
On new or unified plot/trace types relating to `parcoords` and `sankey`


monfera:
Technical notes re https://github.com/plotly/plotly.js/issues/2221#issuecomment-354816207

On multidimensional explorers: made a separate issue as this topic is deep and the topic of "New charts" will likely be quite broad.

We've been planning multidimensional extensions eg. SPLOM. The above directions also make a lot of sense, and as usual there's always the tradeoff among implementation time (ultimately $), payload size and functionality. A nice option would be a kind of unification of plots, eg. `parcoords` and `sankey`, both of which are relatively compact as they do one thing without much configuration. Another option is a new plot type, or a new trace type eg. on the substrate of `parcoords`.

Here's a quick note on the expected challenges of integrating the new thing with either:

- `parcoords` had been written for the express purpose of performance, so it uses WebGL for rendering. It uses `GL.LINES` as that's the most compact geometry ie. fastest to do the interactions with (there were response time criteria). To make lines thicker, we'd need to convert to GL.TRIANGLES or similar, ie. 2-3x work in the vertex shader (which also does GPU crossfiltering). WebGL does have a [`lineWidth`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth) method but the standard permits that implementations cap line width to 1px, which browsers increasingly opted for over time...  If Sankey-like splines are also needed, that's another layer of performance hit and implementational complexity. Yet it'd be possible to add an SVG or polygonal WebGL trace type for those cases where data points are not in the multiple 10k range and more diverse geometry eg. thicker lines are needed. `parcoords` is already multilayer, the axes are in SVG and there could be more SVG layers.
- `sankey` is an alternative target. After all, it internally works as a layered graph, ie. the internal representation is already close to the axis cadence of parallel coordinates and their ilk. But our implementation uses the heuristic as it is in `d3-sankey` (we only added support for multiedges via a [PR](https://github.com/d3/d3-sankey/pull/19)), and it's free to arrange the edges, resulting in the observed line discontinuity that's definitely not in the style of parallel coordinates. This freedom yields a more optimal arrangement in the case of general Sankey work, ie. minimizing the line crossings, not a concern for parcoords-like work. We'd need to somehow add configuration for bypassing the heuristic or the entire `d3-sankey` in favor of a parcoords-like continuous line layout.

[...]

Issue #2221, opened 2017-12-23
New charts wishlist


jackparmer:
The most requested feature in the plotly user [survey](https://plotly.typeform.com/to/v2LvEr) has been "More chart types" 2 years in a row.

This issue inventories remaining chart types (or ["traces"](https://github.com/plotly/plotly.js/tree/master/src/traces) in plotly parlance) that are still requested for plotly.js and [Chart Studio](https://plot.ly/online-chart-maker/). Pipelines to make most of these charts already exists in Python and R by combining lower level plotly.js traces (lines, SVG shapes, bars, area fills, etc).

### "Business charts"

- [x] Funnel charts [Example](https://jakezatecky.github.io/d3-funnel/)
- [x] Stacked area charts [Example](https://bl.ocks.org/mbostock/3885211)
- [x] Waterfall charts [Example](http://bl.ocks.org/chucklam/f3c7b3e3709a0afd5d57)
- [x] Hero numbers [Example](https://bl.ocks.org/mbostock/7004f92cac972edef365)
- [x] Gauges [Example](https://www.highcharts.com/demo/gauge-solid)
- [x] Bullet charts [Example](https://bl.ocks.org/mbostock/4061961)
- [x] Tree maps https://github.com/plotly/plotly.js/issues/1038
- [x] Voronoi tree maps [Example](https://youtu.be/mBX1p1YS-F4) & [resource](https://github.com/Kcnarf/d3-voronoi-treemap)
- [ ] Marimekko charts #4019
- [ ] Timeline https://github.com/plotly/plotly.js/issues/2803 
- [ ] Gantt charts [Discussion](https://community.plot.ly/t/how-to-use-plotly-to-make-gantt-chart-with-discontinuous-tasks/1870/12)
- [ ] Aster plots (https://github.com/plotly/plotly.js/issues/2901)
- [ ] Cost curve charts [Example](https://grist.files.wordpress.com/2007/12/mckinsey_mid_range_abatement_curve.jpg)

### Geo 

- [x] County choropleths (w/o zoom for perf.) [Example](https://bl.ocks.org/mbostock/4060606)
- [ ] Contours on maps ("weather maps") [Example](https://plot.ly/~AdamKulidjian/6254/)
- [x] MapboxGL heatmap support [Example](https://www.mapbox.com/mapbox-gl-js/example/heatmap-layer/)

### Scientific 

- [x] 3d streamtubes [Example](https://gl-vis.github.io/gl-streamtube3d/)
[...]

Issue #2908, opened 2018-08-15
Multiple issues with transforms


dmt0:
I’ve been trying to add transforms to multiple trace types with varying degrees of success. Documentation doesn’t make it clear which trace types support transforms. All I know from discussion with @etpinard, is that anything that takes 2d arrays as input doesn’t support them.
`react-chart-editor` as of now does not prepopulate target/targetsrc:

```
      "transforms": [
        {
          "type": "filter",
          "target": [],
          "targetsrc": null
        }
      ]
```

**Results:**

scatter3d: 
- filter: many warnings: `[.Offscreen-For-WebGL-0x7fef4b0d6800]RENDER WARNING: Render count or primcount is 0.`

-> https://codepen.io/etpinard/pen/dgWJZg

scatter3d + marker:’line’
- filter: `Uncaught (in promise) TypeError: Cannot read property 'slice' of undefined
    at LinePlot.proto.update`

-> https://codepen.io/etpinard/pen/Edmobo

cone, streamtube, scattergeo, table, heatmap (with 1d array input), contour (with 1d array input)
- any transforms: as I add them to `data`, they never get added to `fullData`

choropleth, scatterternary, scatterpolar, pie:
- aggregate: `Error while plotting: TypeError: Cannot set property 'type' of undefined at Object.plots.supplyTransformDefaults`

pie:
- filter: `Error: <path> attribute d: Expected number, "MNaN,NaNa189.5,18…".`

-> https://codepen.io/etpinard/pen/wYdppB

2d contour histogram:
- filter: `Error while plotting: TypeError: Cannot read property 'length' of undefined
    at makeCrossings (plotly-with-meta.js?ab3a:146042)`

**Edit:**
Add this one to the list:
- When defining a `scatter` with just Y defined, and than add an aggregate:
`Uncaught TypeError: Cannot set property 'type' of undefined
    at Object.plots.supplyTransformDefaults`
https://codepen.io/dmt0/pen/xJoWoY

This one got **fixed in:** #3093

etpinard:
Ok. Thanks for the report!

What does `targetsrc: null` which isn't a plotly.js attribute) do?. I suspect it injects `target: []` in the user data, but I would like to confirm

etpinard:
But yeah, we could do better than declaring `transforms` in the schema for all trace types  (including the ones that don't support them yet):

[...]

PR #2963, opened 2018-09-01
Parallel Categories (parcats) trace type for multi dimensional categorical data

jonmmease:
Continuation of https://github.com/jonmmease/plotly.js/pull/1, now against `master` in the plotly.js repo.

I believe I addressed all of the outstanding code comments brought up in the old PR. Below is the original post, and my last few comments on the implementation.

---
@alexcjohnson @etpinard @monfera @chriddyp @jackparmer 
## Introduction
This PR is a proposal and an implementation of a new trace type for the interactive exploration of multi-dimensional categorical data sets.  My working name for the trace is "Parallel Categories" or `parcats` for short.

The concept of this trace has been discussed previously in the following plotly.js issues:
 * https://github.com/plotly/plotly.js/issues/2221#issuecomment-354560667
 * https://github.com/plotly/plotly.js/issues/2229

I also briefly showed a prototype of this diagram to @chriddyp over screenshare several months ago.

## Related work
The closest prior art to the Parallel Categories Diagram is the Parallel Sets Diagram by Robert Kosara and Caroline Ziemkiewicz.

### Parallel Sets implementations / descriptions
Here are a collection of existing implementations / descriptions of the Parallel Sets Diagram

#### Parallel Sets Java Program
https://eagereyes.org/parallel-sets

This is a stand-alone Java program by Kosara that implements a Parallel Sets Diagram

parsets-titanic-dimensions-thumb

#### Parallel Sets from the DataViz catalog
https://datavizcatalogue.com/methods/parallel_sets.html

parallel_sets

#### D3 implementation of Parallel Sets
https://www.jasondavies.com/parallel-sets/

screen shot 2018-07-27 at 1 55 59 pm

## What's different about the Parallel Categories Diagram?
The primary difference between this Parallel Categories Diagram (parcats from here on) and the Parallel Sets Diagram (parsets from each on) is that the parcats diagram supports a more flexible path coloring scheme.

In all of the examples of parsets diagrams that I have found, the colors of the paths correspond to states in the left-most (or top-most) dimension. In contrast, for the parcats diagram, color may correspond to a column in the dataset that may or may not be present as a dimensions in the diagram. 

This, admittedly modest, extension has several advantages. Path colors may be set using a numeric array and a color map just like many other plotly.js trace types (scatter, parcoords, etc.).  This makes it possible to use the parcats diagram combined with other traces in brushing/crossfiltering configurations.

## Dragging and Brushing example
Here is an example of visualizing a 5-dimensional data set with two continuous dimensions and 3 categorical dimensions.  This is accomplished by displaying the two continuous dimensions in a 2D scatter plot and the 3 categorical dimensions using the parcats diagram.

I created this example using a branch of plotly.py version 3 built against this branch of plotly.js. 

First I show the drag interactions supported by the diagram. Categories (the rectangles) and dimensions labels (dimensions are the columns of rectangles) can be dragged to reorder categories and dimensions. Upon release, the diagram animates to a relaxed state with equal spacing between dimensions and categories.

Selection events in the scatter plot are use to update the colors of both the selected points in the scatter plot, and the corresponding paths in the parcats diagram.  Similarly, click events on categories and paths in the parcats diagram are used to update the colors in both diagrams.

parcats_brushing

As far as I'm aware, this is the only visualization of multi-dimensional categorical data that supports this kind of two-way data brushing. And, combined with plotly.py version 3, it is certainly the only visualization of this type that would be easily accessible to Python users.

## Color bundling
There are two modes for how the colors of paths are arranged.

In the example above, color is not considered when sorting the paths. This is desirable in a brushing scenario so that the paths remain stable as the colors change during interactions. This behavior is specified by setting the `bundlecolors` property to `false`. 

Setting the `bundlecolors` property to `true` causes paths with like colors to be bundled together as they pass through each category.  This results in a cleaner looking diagram and is preferable in cases where the positions of paths do not need to remain stable as colors change.

For example:
parcats_bundlecolors

## Mocks
Several simple mocks have been added as a part of the current test suite.

`parcats_basic`
screen shot 2018-07-27 at 2 28 47 pm

`parcats_bundled`
screen shot 2018-07-27 at 2 28 53 pm

`parcats_unbundled`
screen shot 2018-07-27 at 2 29 01 pm

## API notes
I tried to model the API as closely as possible after existing trace conventions. There is a top-level `dimensions` property with `label` and `values` sub-properties just as with the `parcats` trace. Path colors/colorscales are specified under a `dimension.marker` parent property. 

## Alternative approach
In the issues cited at the beginning of this PR there was some discussion on the possibility of adding categorical support to the existing Parallel Coordinates Diagram.  This diagram was already well under development for our internal needs at the time of these discussions, so I did not pursue this approach.

## TODO
Some items that I know still need to be done
 - [x] Font styling support
 - [x] Complete attribute descriptions
 - [x] Complete the test suite. In terms of my personal testing standards I'd estimate that the test suite is about 50% complete.
 - [x] Examples!

## Request for comments
So the top-level question for the plotly.js team is, are you all interested in having this diagram be part of plotly.js? It's not the most common use-case, but I think it would be another differentiating feature for the plotly ecosystem.

If you all are interested, I have internal funding to put a bunch more time into this through September. And if we can get it merged in during that time, I can continue helping out with basic maintenance after that.

Let me know what you think!

---
I just added a mock that demonstrates the `color` hovermode (`parcats_hovermode_color`).

The basic idea here is that when you hover on a category, only the paths of a single color are highlighted. What's really useful about this is that the tooltip can then display the absolute probability of the paths of that color that pass through a given category. You can also display the conditional probabilities (Probability of blue given category A, probability of category A given blue).

hovermode_color

Does `hovermode` of being an enumeration of `none`, `category`, or `color` seem like a reasonable way to specify this?

---
@alexcjohnson 
Regarding dragging, I went ahead and added a `sankey` style `arrangement` property to control the dragging behavior. There are three modes (names taken from `sankey.arrangement`)

 - `perpendicular` (now the default): categories only drag vertically,
    dimension labels drag horizontally.
 - `freeform`: category labels can drag vertically and horizontally
    (in which case they pull the dimension along with them). Here
    dragging a category can reorder the categories and dimensions.
 - `fixed`: dragging of dimensions and categories is disabled.

---
@alexcjohnson 
I took a look back through the multi-label hover logic, and added a new `'dimension'` hover mode to show it off. This hover mode will display a label for each category in the current dimension (See `parcats_hovermode_dimension` mock).

screen shot 2018-08-20 at 2 28 08 pm

If you notice the hover label on `B`, you can see how the label is pushed downward to keep it from overlapping with the label for `C`.  It might be nice in some cases to also push labels upward to avoid collisions, but at the moment this only pushes things down.

Now that there are multiple hovermodes and configurable hoverinfo I do like this mode as an option. 

---
Font support added for dimension labels (`labelfont`) and category labels (`categorylabelfont`). `labelfont` matches the corresponding property name in `parcoords`.


alexcjohnson:
> It might be nice in some cases to also push labels upward to avoid collisions, but at the moment this only pushes things down.

Not sure how easy it will be to reuse, but check out `hoverAvoidOverlaps` which does shift both ways. 

https://github.com/plotly/plotly.js/blob/19f3800bac848e702948fcc32302ded56ea1cdb7/src/components/fx/hover.js#L963

jonmmease:
Parcats update @alexcjohnson @etpinard @nicolaskruchten 

I've generated new baseline images for the `parcats` mocks, fixed up the remaining syntax errors, and now the CI tests are passing.

I've also finished fleshing out the `parcats` jasmine test suite to the point that I'm pretty comfortable with the test coverage (between jasmine tests and image baseline tests).

I have about 3 weeks left before my funding line disappears, so I'm going to move on to writing documentation.

Even if you don't have time to look at the code much this week, I'd appreciate another pass through `src/traces/parcats/attributes.js` since changes here will affect the documentation.

Thanks!

src/traces/parcats/attributes.js
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+module.exports = {
+    domain: domainAttrs({name: 'parcats', trace: true, editType: 'calc'}),
+    hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
+        flags: ['count', 'probability'],
+        editType: 'plot'
+        // plotAttrs.hoverinfo description is appropriate
+    }),
+    hovermode: {
```
etpinard: A little late to the party, but I'd say having `hovermode` as a layout and a trace may sounds very confusing. 

It would be nice to make whether attribute name we pick here also describe the "separate vs combined" cases for finance trace hover labels (cc https://github.com/plotly/plotly.js/pull/2959).

src/traces/parcats/attributes.js
```diff
@@ -0,0 +1,193 @@
...
+            valType: 'data_array',
+            role: 'info',
+            editType: 'calc',
+            description: [
+                'Sets the order in which categories in this dimension appear.',
+                'Only has an effect if `categoryorder` is set to *array*.',
+                'Used with `categoryorder`.'
+            ].join(' ')
+        },
+        categorylabels: {
```
etpinard: Would it be a stretch to call this `ticktext` and `categorylabelfont` -> `tickfont` to preserve more "symmetry" with other trace types?

src/traces/parcats/index.js
```diff
@@ -0,0 +1,33 @@
...
+Parcats.colorbar = {
+    container: 'line',
+    min: 'cmin',
+    max: 'cmax'
+};
+
+Parcats.moduleType = 'trace';
+Parcats.name = 'parcats';
+Parcats.basePlotModule = require('./base_plot');
+Parcats.categories = ['noOpacity', 'markerColorscale'];
```
etpinard: :zap: `markerColorscale` which is now obsolete (from https://github.com/plotly/plotly.js/pull/2681/).

src/traces/parcats/attributes.js
jonmmease: I just ready through #2959 and I agree it sounds like we could use a new property to specify different approaches to displaying tooltips.  Some ideas based on synonyms for "mode":
 - `hoverbehavior`
 - `hoverform`
 - `hoverkind`
 - `hovermethod`
 - `hovertype`
 - `hovervariety`

Any of those sound reasonable? cc @alexcjohnson

src/traces/parcats/attributes.js
jonmmease: Ok, after playing with `categoryorder`, `categoryarray`, `ticktext`, and `tickvals` in `layout.xaxis` I think I do see the parallel.

What I'm calling `categorylabels` would be equivalent to `ticktext`  if we were to introduce a `tickvals` property and always keep it equal to `categoryarray`.  I don't want to actually introduce a `tickvals` property (because I don't see how it would be useful in general), but if we think of it as being there and always fixed to be equal to `categoryarray` then I think the parallel holds.

So I think I'm fine with these renames, any thoughts @alexcjohnson?

src/traces/parcats/attributes.js
jonmmease: Sure thing!

src/traces/parcats/index.js
jonmmease: Done in https://github.com/plotly/plotly.js/pull/2963/commits/b1ef6c94789fb10a3a57dba29a3c0170a9d9553c

src/traces/parcats/attributes.js
jonmmease: Done in https://github.com/plotly/plotly.js/pull/2963/commits/5a19499b5fe56f4f50c44939c4d2784491d356f7!

jonmmease:
Some initial work on a brushing example for the documentation

```javascript
var categoricalDimensionLabels = [
    'aspiration',
    'body-style',
    'engine-location',
    'num-of-cylinders',
];

var rawData;
d3.csv("https://raw.githubusercontent.com/mtrebi/d3_cars/e90ede77443df72e714863f876e921cfb2977b56/data/imports-85.csv",
    function(carsData) {
        rawData = carsData;

        // Preprocess Data
        var mpg = carsData.map(function(row) { return row['highway-mpg'] });
        var horsepower = carsData.map(function(row) { return row['horsepower'] });

        carsData.forEach(function(row) {
            // Reduce number of categories in num-of-cylinders
            var numCyls = row['num-of-cylinders'];
            if (numCyls < 4) {
                row['num-of-cylinders'] = '< 4'
            } else if(numCyls > 4) {
                row['num-of-cylinders'] = '> 4'
            }
        });

        var categoricalDimensions = categoricalDimensionLabels.map(function(dimLabel) {
            // Extract column
            var values = carsData.map(function(row) {
                return row[dimLabel]
            });

            return {
                values: values,
                label: dimLabel
            };
        });

        // Colors
        var color = new Int8Array(carsData.length);
        var colorscale = [[0, 'gray'], [0.5, 'gray'], [0.5, 'red'], [1, 'red']];

        // Layout
        var layout = {
            width: 800,
            height: 800,
            xaxis: {title: 'Horsepower'},
            yaxis: {domain: [0.6, 1], title: 'MPG'},
            dragmode: 'lasso',
            hovermode: 'closest'
        };

        // Build Traces
        var traces = [
            {type: 'scatter',
                x: horsepower,
                y: mpg,
                marker: {colorscale: colorscale, color: color, cmin: 0, cmax: 1},
                mode: 'markers',
            },
            {type: 'parcats',
                domain: {y: [0, 0.4]},
                dimensions:categoricalDimensions,
                line: {
                    colorscale: colorscale,
                    cmin: 0,
                    cmax: 1,
                    color: color,
                    shape: 'hspline'},
                labelfont: {size: 14},
                bundlecolors: false
            }
        ];

        // Make plot
        Plotly.newPlot(gd, traces, layout);

        // Update color on selection and click
        var update_color = function(points_data) {
            if (points_data) {
                var new_color = new Int8Array(carsData.length);
                for(var i = 0; i < points_data.points.length; i++) {
                    new_color[points_data.points[i].pointNumber] = 1;
                }

                Plotly.restyle(gd,
                    {'marker.color': [new_color, undefined],
                        'line.color': [undefined, new_color]}, [0, 1])
            }
        };

        gd.on('plotly_selected', update_color);
        gd.on('plotly_click', update_color);
    });
```
parcatsbrushing

alexcjohnson:
@jonmmease that brushing example is great! I wonder if we can make it look explicitly like a selection though, rather than just a `line.color`... ie like in `scatter` and various other traces, a [`selectedpoints` attribute](https://github.com/plotly/plotly.js/blob/b6d52839a85706758238242a2d8f689419d74a07/src/plots/attributes.js#L104-L115) and [`selected`/`unselected` styles](https://github.com/plotly/plotly.js/blob/b6d52839a85706758238242a2d8f689419d74a07/src/traces/scatter/attributes.js#L447-L518)?

I don't think this (yet?) has the problem I describe in https://github.com/plotly/plotly.js/issues/3012 of overspecifying the selection, since selections in `parcats` are not persistent within a dimension in the same way as they are in `parcoords`, though it's probably worth thinking at this point about whether we envision something like that in the future, ie that ability to restrict the selection to one or more categories within each dimension, where the final selection is the intersection of all the dimension selections...

jonmmease:
Here's an slightly updated brushing example that uses the `selectedpoints` property of the scatter trace (and still the color property of `parcats`). And it sets the `bundlecolors` property to `false` to keep the colors within each category bundled together.

CodePen: https://codepen.io/jonmmease/full/EepbVE/
parcatsbrushing2

Some thoughts on selections to follow...

jonmmease:
In my own analysis work I have tended to use numeric color vectors, rather than the `selectedpoints` index vectors, to visually encode the relationships between trace types.  This is because I often want to be able to 'brush' with multiple colors in order to compare different clusters of behavior across dimensions/visualizations.  I also like having the ability to use the colorbar to label the meaning of the clusters, and it's nice that this approach works with every trace type that support colorscales.

Here's an example of this kind of workflow:

CodePen: https://codepen.io/jonmmease/full/bxmoxO/

parcatsbrushingmulticolor

I think this approach is the most general, and allows people like me to register callbacks and build custom interaction workflows. But I don't think anything in the current design precludes the introduction of selection and constraint modes in the future:

To implement selection we could add `selectedpoints`, `selected`, and `unselected` properties that would behave the same way as those for `scatter`.  So the styles in `selected` and `unselected`  would take priority over those specified in `line` if the selection mode is active (I think there would need to be a new property to indicate that the selection mode is active).  We could use keyboard modifiers to extend the selection (shift) or subtract from the selection (alt) just like with `scatter`.

To implement a constraint mode (analogous to `parcoords`) we could introduce the concept of constrained categories per dimension, analogous the constraint ranges in `parcoords`.  Paths that pass through constrained categories are displayed according to their `line.color`, paths that don't are displayed in gray.

Do you agree that the current set of properties could be extended to support these modes in the future without altering the existing properties?

alexcjohnson:
> Do you agree that the current set of properties could be extended to support these modes in the future without altering the existing properties?

Those are all good points @jonmmease - thanks for writing that up! Yes, we could add this in the future, the only slight concern I see is that `layout.clickmode` would presumably control this behavior in `parcats`, so if someone turns on `select` there before it exists (to control the subplots that support it now, like the cartesian subplot in your example) then it will also turn this on for `parcats` when we add it. I think that's an acceptably small change, but we should be aware of it.

jonmmease:
I just noticed that the generated schema for this trace includes several top-level trace properties that aren't supported.

In particular `selectedpoints`, `hoverlabel`, `opacity`, `ids`, and `customdata`.  I see that these properties are listed in `src/plots/attributes.js`. Is it possible to remove these properties from the generated plot schema for this trace?  I'm hoping to avoid having these properties show up in plotly.py without a bunch of codegen special cases 🙂  @alexcjohnson @etpinard

etpinard:
> I just noticed that the generated schema for this trace includes several top-level trace properties that aren't supported.

Here's one way: set `selectedpoints`, `hoverlabel`, `opacity`, `ids`, and `customdata` to `undefined` in the `parcats/attributes`. The keys will still be in the `PlotSchema.get()` output:

image

but they'll be gone after being `JSON.stringify`'ed, and thus won't show up in the reference pages.

alexcjohnson:
> I just noticed that the generated schema for this trace includes several top-level trace properties that aren't supported.

There's a module `categories: ['noOpacity']` which prevents *coercing* `opacity` (so it's not in `fullData`, which is important for the editor) but does not take it out of the schema (and anyway only handles `opacity`). @etpinard 's solution would fix the schema but not `fullData`. Sounds like what we really want is something like:
```js
module.omitGlobalAttrs: ['selectedpoints', 'hoverlabel', 'opacity', 'ids', 'customdata']
```
that would do both.

There are probably quite a few other traces that omit some of these - in fact, probably everything that sets `traceOut._length = null` has no use for `selectedpoints`, `ids`, and `customdata`...

etpinard:
> which prevents _coercing_ `opacity` (so it's not in `fullData`, which is important for the editor)

Good point here. But I think for this PR my suggestion in https://github.com/plotly/plotly.js/pull/2963#issuecomment-423569219 will be _good enough_. The react-chart-editor (aka RCE) doesn't know how to handle traces with `dimensions` yet, so having a few extra keys in fullData won't affect anymore at this stage. But, once merged we should open a new issue (similar to https://github.com/plotly/plotly.js/issues/2834 and https://github.com/plotly/plotly.js/issues/2908) about this.

jonmmease:
Sounds good, thanks @etpinard and @alexcjohnson. I'll give the `undefined` approach a try for this PR.

jonmmease:
In 273d2d0 I set the unsupported root-level properties to undefined.

In 4bc3d35 I had to add a few guard conditions to `test/jasmine/bundle_tests/plotschema_test.js` to not crash on `undefined` schema attributes.

jonmmease:
I've started playing with this from Python and the default font size (which I believe matches `parcoords`) is feeling pretty small in the context of the notebook.

screen shot 2018-09-25 at 8 48 18 am

Any objection to my increasing this a little bit?

src/components/fx/index.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
     unhover: dragElement.unhover,
 
     loneHover: require('./hover').loneHover,
+    multiHovers: require('./hover').multiHovers,
```
alexcjohnson: 🐄 this isn't new with your PR, but it's exacerbated here: we have multiple matching `require` statements - can you pull that out into one eg `var hoverModule = require('./hover');`?

src/traces/parcats/attributes.js
```diff
@@ -0,0 +1,202 @@
...
+    sortpaths: {
+        valType: 'enumerated',
+        values: ['forward', 'backward'],
+        dflt: 'forward',
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Sets the path sorting algorithm.',
+            'If `forward`, sort paths based on dimension categories from left to right.',
+            'If `backward`, sort paths based on dimensions categories from right to left.'
```
alexcjohnson: I still think we can do better with sorting in general. Take for example `parcats_bundled`:
screen shot 2018-09-28 at 11 22 31 am
In the `Three` dimension all the grey paths come together, yet the B->11 and C->11 paths are interleaved. And the red A->11 goes all the way to the bottom, but it could go to the top of the red bundle, resulting in less self-crossings. We should be able to find an algorithm that untangles these.

I won't hold up the release for this, as long as:
1. @jonmmease you can make an issue to revisit this later, and
2. @etpinard you won't consider it a breaking change if we implement this and make it the default sorting in the future.

Sound reasonable?

src/traces/parcats/calc.js
```diff
@@ -0,0 +1,526 @@
...
+        markerColorscale = Drawing.tryColorscale(line);
+    } else {
+        markerColorscale = Lib.identity;
+    }
+
+    // Build color generation function
+    function getMarkerColorInfo(index) {
+        var value;
+        if(!line) {
+            value = parcatConstants.defaultColor;
```
alexcjohnson: isn't this case ruled out (and the default supplied) by `supplyDefaults`? And it looks like this line is the only use of anything in `parcatConstants`, so perhaps we can 🔪 that?

src/traces/parcats/attributes.js
etpinard: > 2\. you won't consider it a breaking change if we implement this and make it the default sorting in the future.

It won't, I promise. :smirk:

src/traces/parcats/defaults.js
```diff
@@ -0,0 +1,122 @@
...
+var handleArrayContainerDefaults = require('../../plots/array_container_defaults');
+
+var attributes = require('./attributes');
+var mergeLength = require('../parcoords/merge_length');
+
+function handleLineDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
+
+    if(traceIn.line) {
+        coerce('line.shape');
+    }
```
alexcjohnson: why are we looking for `traceIn.line` and not just always `coerce('line.shape')` so we always have a default in `fullData`?

src/traces/parcats/parcats.js
```diff
@@ -0,0 +1,2086 @@
...
+        traceY = layout.height - domain.y[1] * layout.height + margin.t;
+
+    // Handle path shape
+    // -----------------
+    var pathShape;
+    if(trace.line && trace.line.shape) {
+        pathShape = trace.line.shape;
+    } else {
+        pathShape = 'linear';
+    }
```
alexcjohnson: yeah this kind of logic should be already taken care of by `supplyDefaults` so you can just say `pathShape = trace.line.shape`

alexcjohnson:
@jonmmease this is looking great! Aside from the comments above (all minor and straightforward, I think) the only thing I'd like to see is a mock that puts two `parcats` traces side-by-side, to verify that this works. Can you just replace two of the existing mocks with one combined mock? Maybe even using `layout.grid`, looks as though that's plumbed up correctly but would be nice to 🔒 it down!

src/components/fx/index.js
jonmmease: Done in 2100d96

src/traces/parcats/calc.js
jonmmease: Yes, thanks for catching these holdovers from before I understood how `supplyDefaults` worked 🙂 Done in eba1b4a

src/traces/parcats/defaults.js
jonmmease: Done in eba1b4a

src/traces/parcats/parcats.js
jonmmease: Done in eba1b4a

src/traces/parcats/attributes.js
jonmmease: Yeah, I agree that we could certainly find a better default path sorting algorithm.  I'll make an issue to track this shortly.

jonmmease:
Thanks @alexcjohnson!  I really appreciate your time on this.

In 4117612 I merged the font and colorbar mocks into `parcats_grid_subplots`, which displays 4 parcats traces in a 2x2 grid using `layout.grid` subplots.  I was pretty happy to see that this just worked 🙂

I also took the opportunity to throw in a `parcats` trace with pseudo-HTML and MathJax category labels.

parcats_grid_subplots

Let me know if anything else comes to mind!

src/traces/parcats/attributes.js
jonmmease: See https://github.com/plotly/plotly.js/issues/3059

alexcjohnson:
Fantastic grid mock @jonmmease! Good idea to include pseudo-html. And I'm glad that it *just worked* ™️ 🎉  Having multiple traces also shows off that the default coloring behavior, which I hadn't noticed before, is to pull from the trace color sequence - hence the orange and green coloration when you don't specify a color. Is that really what we want, or would it be better to use `'lightgray'` as you had in the old code (that wasn't being used anyway because you had coerced `line.color`)? `parcats` traces don't share subplots, which is normally the reason to pull trace colors from a sequence.

jonmmease:
That's a good point and I don't have a strong preference. The grey felt like a pretty boring default and I thought our default blue was a nicer starting point. I also liked that the default colors could be specified in a template with the `colorway` property, but perhaps that's overloading `colorway` too much.

In any case, if you have a preference for a different default I'm happy to make a change 🙂

nicolaskruchten:
Random last-minute thought: is "parcats" a more easily-grokked/commonly-used name than something like "alluvial diagram" ? I would maybe consider renaming this trace :)

jonmmease:
From my reading "alluvial diagram" is most often used interchangeably with sankey diagram. For example [Wikipedia](https://en.wikipedia.org/wiki/Alluvial_diagram) and [datavizproject](https://datavizproject.com/data-type/alluvial-diagram/) both define it to be essentially equivalent to our sankey diagram.

nicolaskruchten:
OK.

alexcjohnson:
> The grey felt like a pretty boring default and I thought our default blue was a nicer starting point.

That makes sense, why don't we just use `layout.colorway[0]` though, so all `parcats` traces get the same default color. I think that's the last item, then we'll be ready to merge!

Looking at the examples you showed in the head of this PR, they include a feature we don't, which is categorical coloring based on the first dimension. Not needed for this PR though - I'll make another new issue.

jonmmease:
Sounds good. I'll do that tonight. What would be the best way to get at colorway from inside supplyDefaults?

alexcjohnson:
> What would be the best way to get at colorway from inside supplyDefaults?

`layout` is there already as an arg to `handleLineDefaults` - you should be able to just grab `layout.colorway[0]`.

jonmmease:
Alright, I just pushed the `colorway[0]` change.  Almost there!

alexcjohnson:
Beautiful! Yeah, to my eye a consistent blue (when that's the head of the colorway) is definitely better.

Dunno if @etpinard wanted to take a last look at this, but from my side it's ready to go! 💃

etpinard:
I apologise if this might have been discussed before, but on

image

and most baselines, why isn't the parcats trace centered?

jonmmease:
The domain in most of the mocks is set to `"domain": {"x": [0.125, 0.625],"y": [0.25, 0.75]}`.  I made it a bit off center in `x` to test that the resulting SVG geometry lines up where it should.  This geometry is checked explicitly in a couple of tests in `parcats_test.js`.  Then I stuck with the geometry for the rest of the mocks to have a consistent frame for computing mouse interaction locations.

etpinard:
> "domain": {"x": [0.125, 0.625],"y": [0.25, 0.75]}

Wow. I totally missed that we :eye: the mock JSONs. My bad. 

Let's merge this thing.

PR #2965, opened 2018-09-03
- fix issue scattergl and fill to zero and undefined value

ErwanMAS:
Example of the issue =>  https://codepen.io/erwanmas/pen/aaJXLY

When we add the first point and the last point , we must be sure that the first/last point does not have a NaN value .


alexcjohnson:
Thanks for the PR (and demonstration of the issue) @ErwanMAS ! Looks like a good solution to me, just two things I'd like to add to it:
- [ ] Can you include a few more invalid values in the mock - `undefined`, `''`, `'N/A'` perhaps? I'm assuming they've all been turned into the same value by the time we get to your new code, but while we're here lets verify that.
- [ ] Do the same for `tozerox`

src/traces/scattergl/index.js
```diff
@@ -439,10 +439,17 @@ function plot(gd, subplot, cdata) {
...
-                    pos = pos.concat(srcPos);
-                    pos.push(srcPos[srcPos.length - 2]);
-                    pos.push(0);
+                    var firstdef = 1;
+                    while(isNaN(srcPos[firstdef])) {
+                        firstdef += 2;
+                    }
+                    var lastdef = srcPos.length - 1;
+                    while(isNaN(srcPos[lastdef])) {
+                        lastdef += -2;
```
alexcjohnson: `-= 2`?

src/traces/scattergl/index.js
```diff
@@ -439,10 +439,17 @@ function plot(gd, subplot, cdata) {
...
+                    }
+                    pos = [ srcPos[firstdef - 1], 0 ];
+                    pos = pos.concat(srcPos.slice(firstdef - 1, lastdef + 1));
```
alexcjohnson: 🐎 so we don't need to `slice` unnecessarily, can we keep track of whether there actually *were* invalid points on either end?

src/traces/scattergl/index.js
```diff
@@ -439,10 +439,17 @@ function plot(gd, subplot, cdata) {
...
+                    pos = pos.concat([ srcPos[lastdef - 1], 0 ]);
```
alexcjohnson: I notice the existing code didn't take advantage of this, but `push` is variadic:
`pos.push(srcPos[lastdef - 1], 0);`

src/traces/scattergl/index.js
```diff
@@ -439,10 +439,17 @@ function plot(gd, subplot, cdata) {
                 var pos = [], srcPos = (lineOptions && lineOptions.positions) || stash.positions;
 
                 if(trace.fill === 'tozeroy') {
-                    pos = [srcPos[0], 0];
-                    pos = pos.concat(srcPos);
-                    pos.push(srcPos[srcPos.length - 2]);
-                    pos.push(0);
+                    var firstdef = 1;
+                    while(isNaN(srcPos[firstdef])) {
```
alexcjohnson: oh hmm - this checks for invalid `y`, what about invalid `x`?

test/image/mocks/fill_trace.json
```diff
@@ -0,0 +1,52 @@
+{
+  "data": [
```
etpinard: Would you mind renaming this mock `gl2d_fill_trace.json`? We like to prefix gl-based mocks by either `gl2d_` or `gl3d_`.

Would should probably write down in the CONTRIBUTING file at some point.

alexcjohnson: Good point for this PR - but rather than describing this in `CONTRIBUTING.md` how about we make the image test system support subfolders and give them a little more organization? It's getting impossible to find what you want in `mocks` and `baselines`...

ErwanMAS:
i create a new pull request [#2979]

PR #2973, opened 2018-09-05
ensure makesData transforms work with Plotly.react

alexcjohnson:
@nicolaskruchten try this with https://github.com/plotly/react-chart-editor and the fit transform.


etpinard:
:dancer: nicely done.

Issue #2969, opened 2018-09-04
Add support for responsive charts


antoinerg:
In order for charts to look great on any devices/orientation, I suggest that we add the option to resize charts whenever the window's size changes.

For example, it is common for a user to switch his phone from landscape to portrait mode which results in the following unaesthetic/broken result:

![before](https://user-images.githubusercontent.com/301546/45044626-a0616b00-b03e-11e8-9179-786769ca3489.gif)

etpinard:
Good call!

This would probably work best as new [config](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_config.js) option (i.e. I don't think this should be a layout attribute).

bpostlethwaite:
This could have interesting implications (I think positively so!) for Plotly's own embed and react-plotly.js 

cc @nicolaskruchten

etpinard:
> basically baking in https://plot.ly/javascript/responsive-fluid-layout/ via a config option?

We could even shoot for moving all the [embedplot](https://github.com/plotly/streambed/tree/master/shelly/embedplot/static/embedplot) logic in plotly.js.

antoinerg:
> even though the window doesn't change size, the containing element does (via a resize operation or a CSS reflow

`ResizeObserver` is a proposal to react to any events that resize a element. However, only Chrome implements it thus far ([caniuse#ResizeOberserver](https://caniuse.com/#search=ResizeObserver)) so we would need to resorts to a polyfill.

In the meantime, I will submit a PR doing what most (if not all) other plotting libraries claiming to be responsive do: simply react to window resize events.

antoinerg:
Closed by PR https://github.com/plotly/plotly.js/pull/2974

nicolaskruchten:
Let's make sure we get this in the documentation as well! :) /cc @cldougl what's the process for doing this?

cldougl:
Yep- you can open an issue in the documentation repo: https://github.com/plotly/documentation/issues
If you have a link to a test or codepen example that exhibits the new feature that would be super helpful to include in the issue.

PR #2974, opened 2018-09-05
add support for responsive charts

antoinerg:
Initial implementation of responsive charts as proposed in #2969.

By setting `{responsive:true}` in a plot's config, it will be automatically resized when the window size changes. For now it only reacts to window's resize but could be improved by using `ResizeObserver` down the road.

Viewing a responsive chart on a phone going from landscape to portrait now results in the following experience:
after

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -385,6 +385,15 @@ function emitAfterPlot(gd) {
         fullLayout._redrawFromAutoMarginCount--;
     } else {
         gd.emit('plotly_afterplot');
+
+        // make the figure responsive
+        if(gd._context.responsive && !gd._responsiveChartHandler) {
+            // Keep a reference to the resize handler to purge it down the road
+            gd._responsiveChartHandler = function() {Plots.resize(gd);};
+
+            // Listen to window resize
+            window.addEventListener('resize', gd._responsiveChartHandler);
+        }
     }
 }
 
@@ -3154,6 +3163,11 @@ exports.purge = function purge(gd) {
     var fullLayout = gd._fullLayout || {};
     var fullData = gd._fullData || [];
     var calcdata = gd.calcdata || [];
+
+    // remove responsive handler
+    if(gd._responsiveChartHandler) {
+        window.removeEventListener('resize', gd._responsiveChartHandler);
+    }
 
     // remove gl contexts
     Plots.cleanPlot([], {}, fullData, fullLayout, calcdata);
--- a/test/jasmine/tests/config_test.js
+++ b/test/jasmine/tests/config_test.js
@@ -6,6 +6,7 @@ var destroyGraphDiv = require('../assets/destroy_graph_div');
 var click = require('../assets/click');
 var mouseEvent = require('../assets/mouse_event');
 var failTest = require('../assets/fail_test');
+var delay = require('../assets/delay');
 
 describe('config argument', function() {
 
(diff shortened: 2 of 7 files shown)


src/plot_api/plot_api.js
```diff
@@ -385,6 +385,15 @@ function emitAfterPlot(gd) {
...
         gd.emit('plotly_afterplot');
+
+        // make the figure responsive
+        if(gd._context.responsive && !gd._responsiveChartHandler) {
+            // Keep a reference to the resize handler to purge it down the road
+            gd._responsiveChartHandler = function() {Plots.resize(gd);};
+
+            // Listen to window resize
+            window.addEventListener('resize', gd._responsiveChartHandler);
+        }
```
alexcjohnson: Code looks great, but I'm a little surprised to see it so late in the sequence... `resize` has an internal delay, and calls `Plotly.relayout` which will wait for any pending drawing on the plot to finish before making its changes. So to avoid missing events that happen halfway through an async draw (mathjax or maps) I'd have thought we should put this in the synchronous block.

src/plot_api/plot_api.js
```diff
@@ -3155,6 +3164,11 @@ exports.purge = function purge(gd) {
     var fullData = gd._fullData || [];
     var calcdata = gd.calcdata || [];
 
+    // remove responsive handler
+    if(gd._responsiveChartHandler) {
+        window.removeEventListener('resize', gd._responsiveChartHandler);
+    }
```
alexcjohnson: We also need to `delete gd._responsiveChartHandler` so it would reattach if we make a new plot in this div later on. The rest of this happens in `Plots.purge` (called a few lines down), and the `delete` and `window.removeEventListener` need to be in the same place (since there are a couple of places we call `Plots.purge` outside `Plotly.purge`, so seems like it would be safer to put both in `Plots.purge`

alexcjohnson:
@antoinerg thanks for pushing on this one, it's looking good!

I'm not sure how feasible it is to write a test for my [async comment](https://github.com/plotly/plotly.js/pull/2974#discussion_r215378845) - you could try resizing synchronously after `Plotly.plot` returns but that might not do it (or might not do it consistently). I guess perhaps we could artificially inject a delayed promise into `gd._promises` but... might be too hacky to be worth it.

But the comment about [purging and replotting](https://github.com/plotly/plotly.js/pull/2974#discussion_r215381319) we can definitely test, also lets test that we have exactly one resize handler and one redraw after making a plot, then editing something (causing a new `Plotly.plot` call), and THEN resizing the window. That one looks like it already works, lets keep it that way 😈

src/plot_api/plot_api.js
antoinerg: It doesn't have to be this late. To be honest, I wasn't sure where it'd be best to put it in the sequence. Should I put at line 193?https://github.com/plotly/plotly.js/blob/8457b40375b7baae8a2722bfbc27263be6193dfb/src/plot_api/plot_api.js#L190-L196

alexcjohnson: > Should I put it at line 193?

Seems like a good place!

alexcjohnson: Oh one more thing occurs to me: a user could in principle start with `responsive` and then turn it off using `Plotly.react`... not sure why you'd want to do this, but it should work!

test/jasmine/tests/config_test.js
```diff
@@ -546,10 +548,27 @@ describe('config argument', function() {
...
+                viewport.set(newWidth, newHeight);
+            })
+            // Wait for resize to happen (Plotly.resize has a 100ms timeout)
+            .then(delay(200))
+            // Check final figure size
+            .then(function() {
+                checkLayoutSize(newWidth, newHeight);
+            })
+            .catch(failTest);
+        }
```
alexcjohnson: Good idea to pull this function out! Nice and 🌴. You can probably simplify it a bit though:
```js
function testResponsive() {
    checkLayoutSize(startWidth, startHeight);
    viewport.set(newWidth, newHeight);

    return delay(200)()
    .then(function() {
        checkLayoutSize(newWidth, newHeight);
    })
    .catch(failTest);
}
```
And use it like:
```js
Plotly.plot(gd, data, {}, {responsive: true})
.then(testResponsive)
.then(done);
```

test/jasmine/tests/config_test.js
```diff
@@ -560,17 +579,60 @@ describe('config argument', function() {
...
+            var promise = Plotly.plot(gd, data, {}, {responsive: true});
+
+            testResponsive(promise)
+            .then(done);
+        });
+
+        it('should still be responsive if the plot is redrawn', function(done) {
+            var promise = Plotly.plot(gd, data, {}, {responsive: true})
+            .then(function() {
+                Plotly.restyle(gd, 'y[0]', data[0].y[0] * 2);
```
alexcjohnson: You need to `return` these `Plotly.*` calls inside `.then` - as with any function that itself returns a promise - so the new promise gets added to the chain.

test/jasmine/tests/config_test.js
antoinerg: Good call :) Thank you again

src/plot_api/plot_api.js
```diff
@@ -2234,6 +2243,9 @@ exports.react = function(gd, data, layout, config) {
             gd._context = undefined;
             setPlotContext(gd, config);
             configChanged = diffConfig(oldConfig, gd._context);
+
+            // remove responsive handler
+            Lib.clearResponsive(gd);
```
alexcjohnson: hmm actually there's another way to turn off `config.responsive` on a `div`, by calling `newPlot` again. I think the removal belongs in `Plotly.plot` instead, next to where you add the handler:
```js
if(gd._context.responsive) {
    if(!gd. _responsiveChartHandler) { ...attach }
}
else {
    Lib.clearResponsive(gd);
}
```
Then it would handle both cases.

alexcjohnson:
Alright, looks great now - thanks for the quick turnaround on my comments! 💃

etpinard:
@antoinerg's 1st feature! :champagne:

PR #2977, opened 2018-09-06
Add `polar.hole`

etpinard:
checking off the third item on the https://github.com/plotly/plotly.js/issues/2255  open-item list 

peek 2018-09-06 17-51

Commit-by-commit:
- https://github.com/plotly/plotly.js/commit/39b2ee34de7ae1541918aae469139b8edb6b11e6 fixes barpolar + scatterpolar ordering (I'm :smile: to have caught that one before `v1.41.0`)
- https://github.com/plotly/plotly.js/commit/ea7fd2cb1e91cf278da1c11359e599e565411c0b adds `polar.hole` along with a test mock
- f547692eed05512d2ac928f99de5665300f7a6b0 adds a few TODO comments that could be taken care in this PR if they get enough :+1: s from the reviewers

I'll add a few hover tests on subplots with holes and maybe a zoombox test tomorrow morning.

cc @alexcjohnson @antoinerg


src/plots/polar/polar.js
```diff
@@ -333,9 +335,11 @@ proto.mockCartesianAxis = function(fullLayout, polarLayout, opts) {
 
     ax.setRange = function() {
         var sectorBBox = _this.sectorBBox;
-        var rl = _this.radialAxis._rl;
-        var drl = rl[1] - rl[0];
         var ind = bboxIndices[axId];
+        var rl = _this.radialAxis._rl;
+        var radius = _this.radius;
+        var innerRadius = _this.innerRadius;
+        var drl = radius * (rl[1] - rl[0]) / (radius - innerRadius);
```
alexcjohnson: you don't have access to `hole` here to just do `(rl[1] - rl[0]) / (1 - hole)`?

etpinard: Yep, the `polayLayout` is passed to the `proto.mockCartesianAxis`, so yeah this line could be:

```js
var drl = (rl[1] - rl[0]) / (1 - polarLayout.hole)
```

if you prefer.

src/plots/polar/polar.js
```diff
@@ -597,6 +597,9 @@ proto.updateAngularAxis = function(fullLayout, polarLayout) {
     }
     _this.vangles = vangles;
 
+    // TODO maybe two arcs is better here?
+    // maybe split style attributes between inner and outer angular axes?
```
alexcjohnson: If we can do what we need right now without this, we can add it in later. But it opens up a can of worms if you combine hole and sector... do you allow separate styling of the radial segments? and joining these paths cleanly where they meet is a headache - it was painful enough to get that right for cartesian https://rreusser.github.io/plotly-mock-viewer/#20 but here you also need to deal with weird angles and curves... I don't even know in principle how to handle that!

src/plots/polar/polar.js
```diff
@@ -937,6 +940,8 @@ proto.updateRadialDrag = function(fullLayout) {
     var tx = cx + (radius + bl2) * Math.cos(angle0);
     var ty = cy - (radius + bl2) * Math.sin(angle0);
 
+    // TODO add 'inner' drag box when innerRadius > 0 !!
```
alexcjohnson: not quite sure what this means, but if it's something like the clamped inner edge still jumps to be a circle at the center then yeah, we'd better avoid that.

src/plots/polar/polar.js
alexcjohnson: nonblocking but that would be marginally simpler.

src/plots/polar/constants.js
```diff
@@ -24,6 +24,9 @@ module.exports = {
         'frontplot',
         'angular-axis',
         'radial-axis',
+        // TODO hmm, putting angular-line above radial-axis
+        // can sometimes hide radial ticks and label
+        // this happens especially for the 'inner' angular axis
```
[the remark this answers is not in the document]
alexcjohnson: Also I guess if we add shadows around the radial tick labels, like the legacy polar plots have, it would be important to bring these in front of the lines.

test/image/mocks/polar_hole.json
```diff
@@ -0,0 +1,80 @@
...
+      "angularaxis": {"direction": "clockwise"},
+      "domain": {"row": 1, "column": 0}
+    },
+    "polar4": {
+      "hole": 0.4,
+      "domain": {"row": 1, "column": 1},
+      "sector": [0, 180],
+      "angularaxis": {"direction": "clockwise"},
+      "radialaxis": {"angle": 90, "side": "counterclockwise"}
+    },
```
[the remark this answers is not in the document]
alexcjohnson: Oh also - I see it in your gif at the head of this PR, but can you put something in this mock showing `cliponaxis: false` on the inner edge?

alexcjohnson:
@etpinard beautifully done. Nice catch with the bar/scatter ordering, looks like we missed that before because the traces in the related mock are semi-transparent...

I like the way you collected `TODO` items in a commit at the end. Not sure if I'm organized enough to do that myself but I can try 😅

src/plots/polar/polar.js
alexcjohnson: Oh right! Yes, we should add that. Would be confusing if there's room for one but it's not there.

src/plots/polar/constants.js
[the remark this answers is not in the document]
etpinard: > are there any downsides to moving angular-line behind radial-axis?

Here are the results:

https://github.com/plotly/plotly.js/compare/polar-hole...polar-lines-below-ticks

I can't detect anything that looks _off_. Note that cartesian subplots also put axis lines below axis ticks:

https://github.com/plotly/plotly.js/blob/907070555924a034a7e3c2052d6fe114144e3215/src/plots/cartesian/index.js#L492-L498

Ok to merge these `polar-lines-below-tick` commits in this PR?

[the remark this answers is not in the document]
alexcjohnson: I don't detect any changes at all, even if git does. So seems fine, but I'd like to see a mock that's *designed* to change in this case - so, fat lines of weird colors overlapping ticks/labels of unrelated colors - and verify that we like the results

src/plots/polar/polar.js
etpinard: peek 2018-09-07 16-41

I'm really liking this effect.

src/plots/polar/constants.js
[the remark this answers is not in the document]
etpinard: With https://github.com/plotly/plotly.js/pull/2977/commits/9c3dcef207184909f5637fbb0721ce7740f9f9bf `polar_hole` now produces this diff

peek 2018-09-07 17-49

src/plots/polar/polar.js
etpinard: Leaving this for _later_, if ok.

etpinard: -> https://github.com/plotly/plotly.js/issues/2255#issuecomment-419575450

src/plots/polar/polar.js
etpinard: Added in https://github.com/plotly/plotly.js/pull/2977/commits/d46614605b5dd410c255df625460be9d8da15ad7

test/image/mocks/polar_hole.json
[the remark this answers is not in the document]
etpinard: More cases in https://github.com/plotly/plotly.js/pull/2977/commits/8e5833dcc42fbc404f2eba0a3305d4d4c49e8ed9

src/plots/polar/polar.js
etpinard: Simpler is better if it's _marginally_ simpler -> https://github.com/plotly/plotly.js/pull/2977/commits/35199c638fb00e939b7b5e38b99a4207dbd9ee21

etpinard:
Tagging as `reviewable`.

alexcjohnson:
💃

PR #2978, opened 2018-09-06
fix layer ordering of scatter subtypes

### LINKED ISSUE #2972: Line getting drawn over marker after react()
This pen https://codepen.io/nicolaskruchten/pen/XPaKwP shows that if you change from `markers` to `markers+lines` for a `scatter` trace, the line ends up being drawn over the markers.

alexcjohnson:
Fixes #2972 - and some other related ordering bugs on mode changes. Basically just by using more standard d3 idioms 🙄 

cc @etpinard


test/jasmine/tests/scatter_test.js
```diff
@@ -629,6 +630,98 @@ describe('end-to-end scatter tests', function() {
...
+        // visit each case N times, in an order that covers each *transition*
+        // from any case to any other case.
+        var indices = [];
+        var curIndex = 0;
+        for(i = 1; i < cases.length; i++) {
+            for(j = 0; j < cases.length; j++) {
+                indices.push(curIndex);
+                curIndex = (curIndex + i) % cases.length;
+            }
+        }
```
alexcjohnson: Notice here we have 15 cases, so we're covering 15*14 = 210 transitions between cases. Takes about 4 seconds locally but totally worth it, as it uncovered a few other layering problems besides the lines/markers one that prompted this effort.

src/traces/scatter/style.js
```diff
@@ -37,6 +43,9 @@ function style(gd, cd) {
 
 function stylePoints(sel, trace, gd) {
     Drawing.pointStyle(sel.selectAll('path.point'), trace, gd);
+}
+
+function styleText(sel, trace, gd) {
```
alexcjohnson: Previously text was just shoved into the same group as points, which had the pleasant side-effect of letting them be styled together. Now that they're in separate groups they need independent outer selections. AFAICT none of the other users of `stylePoints` actually *wanted* to be styling text anyway (as opposed to the users of `style`), so this may be a slight 🐎 improvement... but @etpinard please take a look.

alexcjohnson: oh hmm, I guess the image failure is telling me `geo` needs this...

alexcjohnson: alright, put back into scatterGeo -> https://github.com/plotly/plotly.js/pull/2978/commits/7c54fc2f0bd38ecac52c314b736e86ef035c3511

Note that `scattergeo` still [removes everything on update](https://github.com/plotly/plotly.js/blob/33e1acfdd5ffb8a87a637e0868827abad60dabbf/src/traces/scattergeo/plot.js#L37) so these ordering bugs should not matter there, but we should be able to use essentially the same structure there as here to switch to a more standard update pattern without introducing similar bugs.

test/jasmine/tests/scatter_test.js
alexcjohnson: Oh hmm, actually this would work perfectly if `cases.length` were prime, but as it is I duplicated some and missed some others. I wonder if there is a generic efficient way to do this with a composite number of cases? Anyway *most* of the transitions are covered so I think this is fine as is but I am curious now...

src/traces/scatter/plot.js
```diff
@@ -157,12 +153,17 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
...
-        line = trace.line,
-        tr = d3.select(element);
+    var trace = cdscatter[0].trace;
+    var line = trace.line;
+    var tr = d3.select(element);
+
+    var errorBarGroup = ensureSingle(tr, 'g', 'errorbars');
+    var lines = ensureSingle(tr, 'g', 'lines');
+    var points = ensureSingle(tr, 'g', 'points');
+    var text = ensureSingle(tr, 'g', 'text');
```
etpinard: Hmm. So now all scatter traces (no-matter their `mode`, `fill` or `error_x`/`error_y` settings) get four sub `<g>`:

image

What if instead we made a "mode" groups data-bind:

```js
var modeData = []
if(trace.error_y.visible || trace.error_x.visible) modeData.push('errorbars');
if(trace.mode.indexOf('lines') !== -1) modeData.push('lines');
if(trace.mode.indexOf('markers') !== -1) modeData.push('points');
if(trace.mode.indexOf('text') !== -1) modeData.push('text');

 var modeGroups = tr.selectAll('g.mode'))
        .data(modeData, identity);
 modeGroups.exit().remove();
 modeGroups.enter().append('g')
        .classed('mode', true)
        .attr('class', identity);
 modeGroups.order();
```

That way those empty `<g>` would be gone :small_airplane:

etpinard:
TODO:

- [ ] make sure all the examples on https://plot.ly/javascript/animations/ work ok

src/traces/scatter/style.js
etpinard: > alright, put back into scatterGeo -> 7c54fc2

Glad to see our tests caught that. 

Now that only scattergeo uses it, maybe we could move `styleText` to 

https://github.com/plotly/plotly.js/blob/master/src/traces/scattergeo/style.js ?

etpinard: Ah never mind, `styleText` is used in `styleOnSelect`.

test/jasmine/tests/scatter_test.js
etpinard: Amazing testing!

src/traces/scatter/plot.js
```diff
@@ -97,51 +99,45 @@ module.exports = function plot(gd, plotinfo, cdscatter, scatterLayer, transition
...
+            .remove();
 
-        tr.selectAll('.js-fill').call(Drawing.setClipUrl, plotinfo.layerClipId);
+        fillJoin.order().each(function(d) {
+            // make a path element inside the fill group, just so
+            // we can give it its own data later on and the group can
+            // keep its simple '_*Fill' data
+            trace[d] = ensureSingle(d3.select(this), 'path', 'js-fill');
+        });
     });
```
etpinard: A little hard to tell from the diff, but wow `createFills` is so much cleaner now. :1st_place_medal:

src/traces/scatter/plot.js
```diff
@@ -517,28 +518,16 @@ function plotOne(gd, idx, plotinfo, cdscatter, cdscatterAll, element, transition
...
-    // there's no timing involved:
-    pointSelection.each(makePoints);
-
-    join.enter().append('g')
-        .classed('points', true)
-        .each(makePoints);
-
-    join.exit().remove();
+    points.datum(cdscatter);
+    text.datum(cdscatter);
```
etpinard: Nice use of `.datum`

src/traces/scatter/style.js
etpinard: > AFAICT none of the other users of stylePoints actually wanted to be styling text anyway (as opposed to the users of style), so this may be a slight  improvement... but @etpinard please take a look.

:+1: for having text and points in different groups!

antoinerg:
I made sure all the animations on https://plot.ly/javascript/animations/ still work! Great work @alexcjohnson 🎉

src/traces/scatter/plot.js
alexcjohnson: (discussed offline but for future reference)

I tried to do this... and it was going well for a while, but I ran into issues with the fact that various disjoint pathways (eg styling, errorbars) expect the trace data to be attached to these groups. And if you use `selection.data` to control creation of these groups, you can't use it to also pass data down to its children. We could nest another single group inside, just for the purpose of passing down the data, but that seems like it's creating exactly the same kind of problem we're trying to avoid. It's almost like we want a parallel data binding, like extending d3 to allow selecting & operating based on a different attribute from `__data__`... or perhaps we make a separate `Lib` function specifically for this purpose: ensure the ordering and conditional existence of a bunch of groups, but bind the same data to each group.

Anyway, for now we agreed to leave it as is.

etpinard: > Anyway, for now we agreed to leave it as is.

:ok_hand:

test/jasmine/tests/scatter_test.js
alexcjohnson: Found an algorithm that works efficiently for any `cases.length` -> https://github.com/plotly/plotly.js/pull/2978/commits/69888187c9de68a3d8a6d75cd1bb84e6c0c66f1a

etpinard:
Nicely done :dancer:

Issue #2765, opened 2018-06-29
scattergl missing lines after non-line trace


etpinard:
See https://codepen.io/etpinard/pen/bKQjMe

plotly.js version of https://github.com/a-vis/regl-line2d/issues/35

cc @dy

brian428:
Just adding a +1 here to get any updates, since this is also affecting us.

ErwanMAS:
https://codepen.io/anon/pen/wEGKew

PR #2979, opened 2018-09-07
Fix fill tozero with undefined value and overlaping of fill area

ErwanMAS:
This is my last version  

 example  of the current issue (  https://codepen.io/erwanmas/pen/JaymVr )

 - when undefined values with option fill tozerox or tozeroy
 - order of fill scattergl when overlap


etpinard:
Thanks again @ErwanMAS for helping us out!

Judging by:

peek 2018-09-07 11-49

does this PR also fix https://github.com/plotly/plotly.js/issues/2765 ?

test/image/mocks/gl2d_fill_trace_filly.json
```diff
@@ -0,0 +1,52 @@
+{
+  "data": [
```
etpinard: Thanks for covering all these cases!

Now, WebGL image tests are expensive (especially on CI), so would you mind collecting all these cases in a single json mock using subplots?

src/traces/scattergl/index.js
```diff
@@ -417,6 +432,24 @@ function plot(gd, subplot, cdata) {
         }
         if(scene.line2d) {
             scene.line2d.update(scene.lineOptions);
+            scene.lineOptions = scene.lineOptions.map(function(lineOptions) {
+                if(lineOptions && lineOptions.positions) {
```
etpinard: Can we combine this new logic here with one of the loops in

https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js#L387-L468

to alleviate potential performance loss.

test/image/mocks/gl2d_fill_trace_filly.json
ErwanMAS: do you a example of subplots ?  @etpinard  

For #2765  i dont really understand why my patch resolve it .

alexcjohnson: The easiest way to make subplots is with the `layout.grid` object - see eg the `stacked_area` [mock](https://github.com/plotly/plotly.js/blob/cfc720bd3c127ea0e206b4e3c2217b356ca92425/test/image/mocks/stacked_area.json#L81) and [image](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/stacked_area.png)
Note in particular `"pattern": "independent"` which makes a separate x/y axis pair for each subplot, ie `xaxis/yaxis` in the first cell, `xaxis2/yaxis2` in the second (which you reference in the traces you want to show up there as `"xaxis": "x2", "yaxis": "y2"` and so on.

ErwanMAS: @etpinard for #2765 ... does not solve more complex thing . example

https://codepen.io/erwanmas/pen/NLwXab

ErwanMAS: only one mock with subplots

src/traces/scattergl/index.js
ErwanMAS: Not in the current state . You have 3 loops ... 2 of this loops currently can not be trigger because line.shape is not managed for scattergl . And the last loop is only for one case , connectgaps .

etpinard: Good eye, we can't _combine_ this loop with the existing loops in convert.js. 

That said, as your new loop over positions doesn't depend on viewport attributes, would you mind moving it over to [convert.js](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js#L387-L468)? We try to keep computations in `ScatterGl.plot` to a minimum. Things that depend on the viewport (e.g. pixel positions) must be in `plot` as the viewport isn't known during `convert`, otherwise the heavy-lifting (which scale as `Math.min(x.length, y.length)` belongs in `convert`.

Sorry for the confusion.

ErwanMAS: a example https://codepen.io/erwanmas/pen/qMoVWW

etpinard: > I can not push the fill to zero logic here because , this part is only used when you have lines

Hmm. I'm not sure I understand. 

You cannot push the fill to logic here [in convert.js, you mean?] because this part [the part you added in `ScatterGl.plot`] is only used when you have lines?

But this block

https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js#L387-L468

is also only used when _you have lines_.

------------



> i added some basics shape for line ( vhv/hvh) , and i activated the options for hv/vh .

Nice. Thanks! :tada:

src/traces/scattergl/attributes.js
```diff
@@ -54,6 +54,8 @@ var attrs = module.exports = overrideAll({
     line: {
         color: scatterLineAttrs.color,
         width: scatterLineAttrs.width,
+        shape: scatterLineAttrs.shape,
+        smoothing: scatterLineAttrs.smoothing,
```
etpinard: Ha. I see what you did there. You had to add `smoothing` so that you could reuse `handleLineShapeDefaults` in the defaults directly.

But `smoothing` and `line.shape: 'spline'` aren't implemented in scattergl, so we can't just reuse the scatter attributes and defaults. It's important to have the correct attributes declaration as these are used to generate doc pages like this one: https://plot.ly/javascript/reference/

src/traces/scattergl/index.js
```diff
@@ -259,15 +259,28 @@ function sceneUpdate(gd, subplot) {
         // draw traces in proper order
         scene.draw = function draw() {
             var i;
+            var allReadyDrawLines = Lib.repeat(false, scene.count);
+            var allReadyDrawMarkers = Lib.repeat(false, scene.count);
```
etpinard: Do we still need this extra logic now that https://github.com/plotly/plotly.js/pull/2990 is merged?

ErwanMAS: IN https://codepen.io/erwanmas/pen/JaymVr
or 
https://github.com/plotly/plotly.js/pull/2979/files#diff-2f69844ce1835e61b2c9811d852a7dd9

look for the lines of the green trace 

It's clearly visible when you don't use the default colors ... because default colors have a alpha channel .

src/traces/scattergl/attributes.js
ErwanMAS: OK ... so i will copy/paste instead .

src/traces/scattergl/index.js
ErwanMAS: The part i added in ScatterGl.plot   is still used when you need to do a fill without lines ., but this block 
https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/convert.js#L387-L468  is only used whn you need to draw a line

src/traces/scattergl/line_shape_defaults.js
```diff
@@ -0,0 +1,15 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+module.exports = function handleLineShapeDefaults(traceIn, traceOut, coerce) {
+    var shape = coerce('line.shape');
```
etpinard: No need for a new file here.

Putting that `coerce('line.shape')` in the `if(subTypes.hasLines(traceOut))` block will suffice.

ErwanMAS: last version of commit resolve this

src/traces/scattergl/index.js
etpinard: OK true. But there's must _still_ be another way to move this block into convert.js.

Of course, you don't have to do this (thanks again for this PR!), but at some point someone will to improve performance.

src/traces/scattergl/index.js
```diff
@@ -259,15 +259,28 @@ function sceneUpdate(gd, subplot) {
...
             for(i = 0; i < scene.count; i++) {
                 if(scene.fill2d && scene.fillOptions[i]) {
                     // must do all fills first
                     scene.fill2d.draw(i);
+                    if(scene.line2d && scene.lineOptions[i] && (scene.fillOptions[i].fillmode === 'tozeroy' || scene.fillOptions[i].fillmode === 'tozerox')) {
+                        scene.line2d.draw(i);
+                        allReadyDrawLines[i] = true;
+                    }
+                    if(scene.scatter2d && scene.markerOptions[i] && (!scene.selectBatch || !scene.selectBatch[i]) && (scene.fillOptions[i].fillmode === 'tozeroy' || scene.fillOptions[i].fillmode === 'tozerox')) {
```
etpinard: Looks like this isn't 100% correct still. 

Referencing https://github.com/plotly/plotly.js/issues/2450 and codepen https://codepen.io/alexcjohnson/pen/YemVOd?editors=1010 - which doesn't render correctl.

I think our (not yours) "must do all fills first" assumption is wrong.

etpinard:
@ErwanMAS thanks again for you hard work.

If I understand correctly, this PR fixes 2 bug (fill tozero with missing coordinates and overlapping fills ordering) and adds a feature (all scatter `line.shape` values except `'spline'`).

Currently, 

- the fill tozero bug fix looks good :white_check_mark: 
- I still have doubts about the overlapping fills fix. It is an improvement for sure, but it doesn't resolve all our fill/ordering issues in scattergl. :no_entry_sign: 
- the new `line.shape` feature looks good :white_check_mark:

In its current form, this PR could only be released in the next minor (`v1.42.0`  which is probably two weeks away) as it adds a new feature. If you're looking to have your two bug fixes released sooner, I'd recommend opening another PR or two (one for each bug fix).

test/image/mocks/gl2d_fill_trace_tozero_order.json
```diff
@@ -0,0 +1,105 @@
...
+    },
+    {
+	"x": [ 1,  2,  2,  3,  4,  4,5
+      ],
+	"y": [ 0,  0,50,50,50,  0,0
+      ],
+	"type": "scattergl",
+	"fill": "tozeroy",
+	"mode": "none",
+	"showlegend": false,    "xaxis": "x3", "yaxis": "y3" ,
```
etpinard: Can you turn on `showlegend` in these mocks, to help us determine the trace ordering? Thanks!

src/traces/scattergl/index.js
ErwanMAS: i updated with my branch , and i add a new test ( https://github.com/plotly/plotly.js/pull/2979/files#diff-81d378492b6b240494eadcde98f23458 )

ErwanMAS:
@etpinard   , i released a new version of my patch .

ErwanMAS:
@etpinard   i updated my patch . and i added a new test case . It will fix the overlapping issues

etpinard:
Thanks @ErwanMAS !

In the future, would you mind not `git push -f` new commits on branches associated with open PRs? This makes somewhat annoying to keep it with your work. For example, from the Github UI I can't tell which of your commits are "new". If you need to bring your branch up-to-date with the current master, use `git merge master` instead. Thank you!

src/traces/scattergl/index.js
```diff
@@ -261,11 +261,8 @@ function sceneUpdate(gd, subplot) {
             var i;
             for(i = 0; i < scene.count; i++) {
                 if(scene.fill2d && scene.fillOptions[i]) {
-                    // must do all fills first
```
etpinard: Hmm. Looks like this broke the `gl2d_scatter_fill_self_next` mock:

peek 2018-10-04 17-07

The first two traces have `fill: 'tonext'` meaning their fills should be behind both the `data[0]` and `data[1]` markers. See corresponding SVG structure:

image


available here -> https://rreusser.github.io/plotly-mock-viewer/#scatter_fill_self_next

ErwanMAS: i pushed a new version .

src/traces/scattergl/index.js
```diff
@@ -259,25 +259,56 @@ function sceneUpdate(gd, subplot) {
         // draw traces in proper order
         scene.draw = function draw() {
             var i;
+            var we_must_draw_previous_line_marker = 0;
             for(i = 0; i < scene.count; i++) {
+                var we_defer_somes_draw = 0;
+                if(scene.fill2d && scene.fillOptions[i] && scene.fillOptions[i].fillmode && scene.fillOptions[i].fillmode === 'tonext') {
```
etpinard: I think you got the fill part right :tada: 

The logic is a little hard to follow; it would be nice to to :recycle: the logic from the SVG version:

https://github.com/plotly/plotly.js/blob/88d48c57ea7e9876fce134a6731ff7542178e3a8/src/traces/scatter/plot.js#L95-L105

by using `_nexttrace` and `_prevtrace` which are defined during [`linkTraces`](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/link_traces.js) called in scattergl [here](https://github.com/plotly/plotly.js/blob/88d48c57ea7e9876fce134a6731ff7542178e3a8/src/traces/scattergl/index.js#L369.

Moreover, looks like the error bars are drawn above the lines, SVG scatter does it the other way around.

----

Oh well, I'll try out a few things and hopefully that one or two commits will be enough to bring this to the finish line.

Thanks again.

etpinard: @ErwanMAS I made two commits on branch [`etienne-scattergl-ordering`](https://github.com/plotly/plotly.js/compare/etienne-scattergl-ordering) which appear to make things work while reusing the SVG scatter logic, See:

- https://github.com/plotly/plotly.js/commit/075bd145e9122a87456cf511d729b2e05e45859a
- https://github.com/plotly/plotly.js/commit/6ae6283657c6293640bedde0d7cfa1748320a35d

Do they look ok to you? If so, I'll release them along with your work in next week's `v1.42.0`.

ErwanMAS: @etpinard for me this seem to be good .  Can you take the last version of https://github.com/plotly/plotly.js/blob/4b2b0d5d34e648e361532fe23f5dbff536792e1c/test/image/mocks/gl2d_scatter_fill_self_next_vs_nogl.json . The new version has now 4 charts .

etpinard:
Ok thanks again @ErwanMAS !

I opened a new PR (https://github.com/plotly/plotly.js/pull/3087) with your commits from this PR along with my small improvements. Closing this PR in favour of #3087

PR #2983, opened 2018-09-07
fix legend when figures has large margins

### LINKED ISSUE #2971: Horizontal legend wrap calculated before automargin
Hi,

The horizontal legend wrapping seems to be calculated before any auto margin calculations are applied.
This means that the legend can appear off the page if longer axis labels are used.

https://jsfiddle.net/vj5a6r4x/

antoinerg:
Fixes #2971 

Simple fix: the width of a legend was compared to user-supplied margins instead of computed margins. 

For testing, I added a baseline image. Should I do this in jasmine instead?

@etpinard @alexcjohnson


src/components/legend/draw.js
```diff
@@ -639,12 +639,12 @@ function computeLegendDimensions(gd, groups, traces) {
         });
 
         // check if legend fits in one row
-        oneRowLegend = (fullLayout.width - (fullLayout.margin.r + fullLayout.margin.l)) > borderwidth + fullTracesWidth - traceGap;
+        oneRowLegend = (fullLayout.width - (fullLayout._size.r + fullLayout._size.l)) > borderwidth + fullTracesWidth - traceGap;
```
alexcjohnson: `fullLayout.width - (fullLayout._size.r + fullLayout._size.l) === fullLayout._size.w`

alexcjohnson:
> For testing, I added a baseline image. Should I do this in jasmine instead?

An image is the way to go here, since it's a presentation element, so it's harder (and more fragile in case we change element structure) in jasmine to be sure we're testing the right thing.

src/components/legend/draw.js
antoinerg: Nothing gets by you :) Thank you again for the quick review!

alexcjohnson:
Looks great! Nice detective work @antoinerg 💃

PR #2986, opened 2018-09-07
Emit 'plotly_webglcontextlost' event on WebGL context lost

etpinard:
cc @alexcjohnson 

I'm not sure if there's a way to actually test this.

This PR makes regl-based graphs emit an event on WebGL context lost from one of our canvases.

Note that gl3 already sort of does this:

https://github.com/plotly/plotly.js/blob/ff3c324bc09029766b9e4b7ab517b83166a7e363/src/plots/gl3d/scene.js#L231-L234

but logs something instead of emitting an event. Personally, I think emitting an event is better, but maybe consistency with gl3d is best (this is, before the v2 push).

### DIFF
--- a/test/jasmine/tests/gl2d_plot_interact_test.js
+++ b/test/jasmine/tests/gl2d_plot_interact_test.js
@@ -218,14 +218,17 @@ describe('Test gl plot side effects', function() {
     it('@gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
         var _mock = Lib.extendDeep({}, require('@mocks/gl2d_12.json'));
 
+        function _trigger(name) {
+            var ev = new window.WebGLContextEvent('webglcontextlost');
+            var canvas = gd.querySelector('.gl-canvas-' + name);
+            canvas.dispatchEvent(ev);
+        }
+
         Plotly.plot(gd, _mock).then(function() {
             return new Promise(function(resolve, reject) {
                 gd.once('plotly_webglcontextlost', resolve);
                 setTimeout(reject, 10);
-
-                var ev = new window.WebGLContextEvent('webglcontextlost');
-                var canvas = gd.querySelector('.gl-canvas-context');
-                canvas.dispatchEvent(ev);
+                _trigger('context');
             });
         })
         .then(function(eventData) {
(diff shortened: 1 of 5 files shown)


alexcjohnson:
> Personally, I think emitting an event is better, but maybe consistency with gl3d is best

👍 for events in all cases.

> I'm not sure if there's a way to actually test this.

Maybe not to cause a real context loss, but can we mock the native event, just so we can be sure the plotly event is triggered on *any* of the canvases losing their context?

etpinard:
> Maybe not to cause a real context loss, but can we mock the native event, just so we can be sure the plotly event is triggered on any of the canvases losing their context?

Good call. Done in -> 7e282c37f

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -214,6 +214,27 @@ describe('Test gl plot side effects', function() {
...
+    it('@gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
+        var _mock = Lib.extendDeep({}, require('@mocks/gl2d_12.json'));
+
+        Plotly.plot(gd, _mock).then(function() {
+            return new Promise(function(resolve, reject) {
+                gd.on('plotly_webglcontextlost', resolve);
+                setTimeout(reject, 10);
+
+                var ev = new window.WebGLContextEvent('webglcontextlost');
+                var canvas = gd.querySelector('.gl-canvas-context');
```
alexcjohnson: What I meant by "triggered on any of the canvases losing their context" is, for this one plot we make 3 canvases, each with its own webgl context... any of them could in principle lose its context, right?

```
> gd.querySelectorAll('canvas')
NodeList(3) [
    canvas.gl-canvas.gl-canvas-context,
    canvas.gl-canvas.gl-canvas-focus,
    canvas.gl-canvas.gl-canvas-pick
]
```

etpinard: Yeah, you caught me being lazy. 

Better tests in https://github.com/plotly/plotly.js/pull/2986/commits/905956a62560d92cfc5b7b92e93b1b9cb80da546

etpinard: even better tests in https://github.com/plotly/plotly.js/pull/2986/commits/3a38d739b7e33415e883f1599f062cd7c3163f8e

alexcjohnson:
💃 
Make sure we add a mention of this to https://plot.ly/javascript/plotlyjs-events/

etpinard:
> Make sure we add a mention of this to https://plot.ly/javascript/plotlyjs-events/

will be in https://github.com/plotly/documentation/issues/1085

Issue #2915, opened 2018-08-16
[Feature] Show name of a Pie/Donut in Multi Pie.


ebellumat:
I use `annotations` to put the name of a Pie inside the hole of a donut. But, with a variable number of Pie's, it's very difficult to calculate coordinates. Eventually they get misaligned. You will could include some option to display the name of the pie chart on top, or in the center, rather than just display it on the hover.

![image](https://user-images.githubusercontent.com/26942680/44231066-483bf500-a173-11e8-9456-50ce6ed03e1a.png)

alexcjohnson:
Absolutely, this has come up before, would be very useful!

etpinard:
Done in https://github.com/plotly/plotly.js/pull/2987

Issue #75, opened 2015-12-04
Pie charts don't always seem to fit labels to a constrained space


tlrdstd:
Maybe it's just me, but I keep ending up with pie chart labels that look like this: http://codepen.io/tlrdstd/pen/QyWybK

![image](https://cloud.githubusercontent.com/assets/194300/11580670/9a700422-9a07-11e5-8927-3f24562b0d32.png)

It happens when I assign a fixed size to the pie chart, either by passing it in the `layout` hash, or in `style` attributes on the div. The pie chart always grows to fit the space, regardless of whether there is room to fit the labels. With a 300x300 square in this example, both the title and the text are cut off. If I increase to a 500x500 square, the title fits, but the text is still cut off because the pie chart grows to match.

Converting the square to a rectangle solves the problem, but that doesn't fit my design space.

I haven't dug into this in the code yet...after trying to solve variations on the same problem in other libraries (flot, d3pie, and uvcharts come to mind), I'm just assuming it ain't easy. d3pie does the best job of plotting labels of anything I've seen, though it too forgets to check the edges of the box.

alexcjohnson:
We certainly could check for labels going off the edge of the plot, and shrink them to fit, that probably wouldn't be very hard.

The problem gets worse if you have several pies, their respective labels do nothing to avoid each other. That seems like a harder one to fix...

mosteele:
This is not just you! Mine are getting cut off when embedded.
![image](https://cloud.githubusercontent.com/assets/2314995/22720022/3d7b332c-ed5d-11e6-8479-2a3cd0ce3c7a.png)

mosteele:
FYI - I as able to fix this by adding margins, but to do it I had to give up the go.Pie() approach. What about adding margins as a parameter of the go.Pie method?

Thanks! LOVE PLOTLY! <3

lucky6qi:
Any workaround solution for pit chart auto margin?

phanigokaraj:
guys, any workaround for this?

nicolaskruchten:
Related #3607

etpinard:
Merging into (updated) https://github.com/plotly/plotly.js/issues/4223

PR #2987, opened 2018-09-08
Pie title

codrut3:
Hi,

I tried to solve #2915 in this PR. I added an attribute `title` to the pie chart, with the option to show it either inside the hole, or outside.

Showing the title inside is easy. On the other hand, I couldn't come up with an algorithm for placing the title outside that works well in all cases. The problem is that the title may intersect labels / other pies / other images or even go out the chart. So what I do, I consider 2 cases:

- there are several pies per column. Then I scale the pie until enough space for the title is left. There is no other solution, because in the following example the pies leave no free space.
newplot

- if there is one pie per column, I place the title high enough to not intersect pulled slices, and leave space for at most one label below it. It may still intersect slice labels that were moved around, or even be just badly placed. Feel free to suggest another approach :) Maybe one could use an attribute `titleheight` and let the user decide the optimal placement?

I made some mocks, so you can see how it looks.


src/traces/pie/attributes.js
```diff
@@ -180,6 +180,33 @@ module.exports = {
...
+        editType: 'calc',
+        description: [
+            'Sets the title of the pie chart.',
+            'If it is empty, no title is displayed.'
+        ].join(' ')
+    },
+    titleposition: {
+        valType: 'enumerated',
+        values: ['inhole', 'outside'],
+        dflt: 'outside',
```
alexcjohnson: Can we make this look a bit more like [`textposition`](https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/attributes.js#L512)? I'm not sure *quite* all of those options make sense, but most of them do. What you have now corresponds to `'top center'` and `'middle center'`, and it's fine to start with just those two options, but I can certainly see wanting `'top left'`, `'top right'`, and `'bottom (left|center|right)'` in the future.

Also I'd remove the `dflt` here and make it dependent on `hole`: if it's zero, default to `'top center'`, otherwise `'middle center'`.

src/traces/pie/defaults.js
```diff
@@ -65,6 +65,15 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         }
     }
 
+    var title = coerce('title');
+    if(title) {
+        var titlePosition = coerce('titleposition');
+
+        if(titlePosition === 'inhole' || titlePosition === 'outside') {
```
alexcjohnson: You don't need this `if` - after `coerce`, we know `titlePosition` has a valid value. (but if you want to prohibit center positioning when there's no hole, this would be the place:
`if(!hole && titlePosition === 'middle center') traceOut.titlePosition = 'top center';`

src/traces/pie/plot.js
```diff
@@ -454,6 +501,72 @@ function transformOutsideText(textBB, pt) {
...
+    return {
+        x: cd0.cx,
+        y: cd0.cy - (1 + maxPull) * cd0.r - (titleSpace / 2),
+        scale: Math.min(scaleX, scaleY)
+    };
+}
+
+function isSinglePie(trace) {
+    // check if there is a single pie per y-column
+    if(trace.domain.y[0] === 0 && trace.domain.y[1] === 1) return true;
```
alexcjohnson: I don't think we want to treat single-pie-per-column differently from multiple, but it's not clear to me offhand whether to shrink all or none. The argument for shrinking all of them is that there's also `layout.title` at the top, and we don't want pie titles to overlap with that. The argument for shrinking none is that we already put other labels in the margins - cartesian axis  titles, ticks, & tick labels. People making multi-subplot plots are used to leaving space for this between their subplots, and in fact if you make a `layout.grid` (see https://codepen.io/alexcjohnson/pen/NBZYpm/) there's a gap by default between neighboring subplots, which is easy to increase (`layout.grid.ygap`).

I guess *most* of our subplots - whether single-trace like pie or multi-trace like cartesian - stay within their domains, which would argue for shrinking all. The only exceptions I think are cartesian and ternary subplots.

src/traces/pie/plot.js
```diff
@@ -454,6 +501,72 @@ function transformOutsideText(textBB, pt) {
...
+    return false;
+}
+
+function getTitleSpace(trace, plotSize) {
+    var chartHeight = plotSize.h * (trace.domain.y[1] - trace.domain.y[0]);
+    // leave 3/2 * titlefont.size free space. We need at least titlefont.size
+    // space, and the 1/2 * titlefont.size is a small buffer to avoid the text
+    // touching the pie.
```
alexcjohnson: This should be based on `titleBB.height` - since you can include `'multi<br>line<br>titles'`
But a buffer of half the font size is still good.

Actually, in principle anything that depends on the size of text needs to happen in the (potentially async) callback 3rd argument to `convertToTspans`, in order to handle MathJax (`title: '$x+y=z$'` etc). That may throw a bit of a wrench into this whole process, since `scalePies` happens so early on in the process... seems like perhaps all the pie titles need to be rendered upfront, stashed somewhere, and *all* the existing plotting happens in their combined completion callback. Super annoying, I know...

test/jasmine/tests/pie_test.js
```diff
@@ -182,6 +182,53 @@ describe('Pie traces:', function() {
...
+        Plotly.newPlot(gd, [{
+            values: [2, 2, 2, 2],
+            title: 'Pie',
+            titleposition: 'inhole',
+            hole: 0,
+            type: 'pie',
+        }], {height: 300, width: 300})
+        .then(function() {
+            var title = d3.selectAll('.titletext text');
+            expect(title.size()).toBe(0);
```
alexcjohnson: Mentioned above, but I feel like in this case it'd be better to just push the title outside.

test/image/mocks/pie_title_pull.json
```diff
@@ -0,0 +1,21 @@
+{
+  "data": [
+    {
+      "values": [
+        50, 49, 48, 47, 46, 45, 44, 43, 42, 41
+      ],
+      "pull": [
+        0, 0.5, 0, 0.5, 0, 0.5, 0, 0.5, 0, 0.5
```
alexcjohnson: Thanks for considering `pull` in all this - nice touch!

test/image/mocks/pie_title_inhole.json
```diff
@@ -0,0 +1,18 @@
+{
+  "data": [
+    {
+      "values": [955, 405, 360, 310, 295],
+      "labels": ["Mandarin", "Spanish", "English", "Hindi", "Arabic"],
+      "textinfo": "label+percent",
+      "hole": 0.3,
+      "title": "Num. speakers",
```
alexcjohnson: Looks like this one is too small already to trigger scaling down - would you mind making this a series including both large and small titles, and perhaps multiline labels in both cases? (Also multiline for outside labels)

alexcjohnson:
@codrut3 thanks for the PR, this is looking really strong!

One other question that occurred to me is how will these titles interact with outside slice labels? These labels already have known open issues (eg https://github.com/plotly/plotly.js/issues/75) so I won't be surprised if the answer is "not well" and I don't think we need to find a solution to that now, but it's worth keeping in mind and asking whether there's anything simple we could do to improve it in the short term.

src/traces/pie/attributes.js
etpinard: > What you have now corresponds to 'top center' and 'middle center', and it's fine to start with just those two options, but I can certainly see wanting 'top left', 'top right', and 'bottom (left|center|right)' in the future.

> Also I'd remove the dflt here and make it dependent on hole: if it's zero, default to 'top center', otherwise 'middle center'.

voting :+1:

src/traces/pie/plot.js
etpinard: > The only exceptions I think are cartesian and ternary subplots.

Polar axes are also drawn out-of-their domains (for most cases that only implies angularaxis tick and tick labels). In other words only the _plotting_ area is constraint by the domain.

If I understand correctly, long outside labels (e.g. https://rreusser.github.io/plotly-mock-viewer/#pie_style_arrays) do not have an effect on the margins, but arguably _could_ (just like cartesian axis tick labels now have an `automargin:true` option). 

All in all, I'm leaning on the side on consistency with outside labels, that is don't touch the margins but shrink the pies.



> I don't think we want to treat single-pie-per-column differently from multiple

I agreed :100: % here. So yeah, let's shrink them all!

codrut3:
Hi @alexcjohnson @etpinard ,

 Thank you for your comments and for taking the time to review my PR! I will probably need a few days to make the changes. I will add all the options that you mentioned to `titleposition`, pre-render the title and shrink the pie to position it. But I need to think about MathJax. Can't I just set `data-notex`? :smile:
 
Also, the title may intersect outside labels. I also noticed that outside labels can go outside the plot or overlap other pies. This is, like, a whole issue on its own. I will think about this too.

alexcjohnson:
> Can't I just set `data-notex`?

Actually that's a good point, since we don't support LaTeX in pie slices either for now (see the [mathjax test image](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/mathjax.png)), I wouldn't be opposed to `data-notex` here. I suspect there's not *too* much intersection between pie users and LaTeX users... :neckbeard:

codrut3:
Hi,
I've made some changes. Can you have a look and tell me if it is fine so far?

I added the following options to `titleposition`: top left, top center, top right, middle center, bottom left, bottom center, bottom right. I did not add middle left and right, because I am not sure how they would look.

Also, I added unit tests that:
- check correct placement of multiline titles in all 7 cases,
- check correct placement when some slices are pulled,
- check correct scaling of title when placed in the hole ('middle center'),
- check correct placing of very large title.

If this is fine, next I will try to do something about titles intersecting outside labels.

Also, note that placing the title to the left (or right) means placing it on the left (or right) margin of the pie (`x = center.x - radius`). So not the left edge of the subplot, which would be further away. Is this good? What do you think?

src/traces/pie/attributes.js
codrut3: Done, added all options mentioned above.

src/traces/pie/plot.js
codrut3: Done. I always shrink the pies to gain space for the title.

src/traces/pie/plot.js
codrut3: I handled multiline titles.
Making it work for MathJax is much harder. Maybe I could just use `data-notex`?

test/jasmine/tests/pie_test.js
codrut3: Done.

test/image/mocks/pie_title_pull.json
codrut3: Thanks!

test/image/mocks/pie_title_inhole.json
codrut3: Done. I added other mocks using multiline titles and scaling.

test/jasmine/tests/pie_test.js
```diff
@@ -183,52 +183,236 @@ describe('Pie traces:', function() {
...
+            title: 'Test<BR>Title',
+            titleposition: 'middle center',
+            titlefont: {
+                size: 12
+            },
+            hole: 0,
+            type: 'pie',
+            textinfo: 'none'
+        }], {height: 300, width: 300})
+        .then(_verifyTitle(false, false, true, false, true))
```
alexcjohnson: `_verifyTitle` is beautifully DRY 🌴 ! But you're running it synchronously here, because the argument to `then` is the *result* of `_verifyTitle`, not a function to call when the `newPlot` promise resolves. Which turns out to be ok, as this plot (with no MathJax or maps) renders synchronously. My preferred way to handle this is to return a function from `_verifyTitle`:

```js
function _verifyTitle(checkLeft, checkRight, checkTop, checkBottom, checkMiddleX) {
    return function() {
        var title = d3.selectAll('.titletext text');
        ...
    };
}
```

src/traces/pie/defaults.js
```diff
@@ -65,19 +65,17 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         }
     }
 
+    handleDomainDefaults(traceOut, layout, coerce);
+
+    var hole = coerce('hole');
     var title = coerce('title');
     if(title) {
         var titlePosition = coerce('titleposition');
```
alexcjohnson: I still think we should make the default be `'middle center'` when there's a `hole` (ie `var titlePosition = coerce('titleposition', hole ? 'middle center' : 'top center');` and remove `attributes.titleposition.dflt`) - or maybe when `hole > 0.5` or something... would be strange to put the title in the center when the hole is very small!

But I'm not wedded to that idea; if you've considered it and still think `'top center'` should be the default in all cases we can keep it that way.

src/traces/pie/plot.js
alexcjohnson: per https://github.com/plotly/plotly.js/pull/2987#issuecomment-421176694 - yes, just use `data-notex`.

alexcjohnson:
@codrut3 thanks for the changes, they look fantastic! Lovely tests and very thorough test images 🏆 

> Also, note that placing the title to the left (or right) means placing it on the left (or right) margin of the pie (`x = center.x - radius`). So not the left edge of the subplot, which would be further away. Is this good? What do you think?

Agreed 💯 - thanks for pointing this out.

> If this is fine, next I will try to do something about titles intersecting outside labels.

You're certainly welcome to do that here, but that can also be a separate PR (ideally addressing #75 too).

So the only things I'd still like to see before merging are `_verifyTitle` https://github.com/plotly/plotly.js/pull/2987#discussion_r219370769, `data-notex` https://github.com/plotly/plotly.js/pull/2987#discussion_r219372847 and finalizing the default `titleposition` https://github.com/plotly/plotly.js/pull/2987#discussion_r219372162

codrut3:
Hi @alexcjohnson ,
Thank you for the review! I've changed `_verifyTitle` and `titleposition` as you say. However, what should I do about `data-notex`? I set it for the title on line 323. Should I write a test?

etpinard:
Amazing PR @codrut3 

Adding this to the `1.42.0` milestone.

alexcjohnson:
Ah sorry, I hadn't noticed that `data-notex` was already in there, I thought it was still a "should I do this" question. Nothing further is needed. Fantastic work @codrut3, thanks for the quick updates! 💃 

@etpinard whenever you're ready to start merging 1.42 features go for it!

etpinard:
> whenever you're ready to start merging 1.42 features go for it!

We'll make one more release under 1.41.x next week, then start prepping 1.42.0.

nicolaskruchten:
Wow, this is really exciting! Hey @ebellumat once this is done I actually would consider adding first-class multi-pie support to PivotTable.js and react-pivottable :)

etpinard:
Starting merging things for 1.42.0 now.

ebellumat:
> Wow, this is really exciting! Hey @ebellumat once this is done I actually would consider adding first-class multi-pie support to PivotTable.js and react-pivottable :)

Wow, I was waiting for this, I did not imagine it would be so fast! Congratulations to those involved. And Thanks to @nicolaskruchten to remember that pull request in PivotTable.js

PR #2992, opened 2018-09-10
scatter3d now performs axis transformation on error bars

### LINKED ISSUE #2962: Error bars not scaled correctly in logarithmic 3D plots
Plotly.js translation of https://github.com/plotly/plotly.py/issues/1139 reported by @jdieg0

> generating a Scatter plot with a logarithmic y axis and given error_y works fine whereas creating a similar plot with Scatter3d with error bars results in extremely large error bars. The proportions of the error bars in the 3D plot are correct if type='log' is omitted.

Codepen: https://codepen.io/anon/pen/YOZyQq

In `scatter` the error bars work as expected.

```javascript
// ## Constants
var x = [0., 1., 2., 3., 4., 5., 6., 7., 8., 9.]
var y = x.map(function() {return 0})
var z = [1.e+00, 1.e+01, 1.e+02, 1.e+03, 1.e+04,
         1.e+05, 1.e+06, 1.e+07, 1.e+08, 1.e+09]

var error_z = [3.e-01, 3.e+00, 3.e+01, 3.e+02, 3.e+03, 
               3.e+04, 3.e+05, 3.e+06, 3.e+07, 3.e+08]

// ## Scatter 2D
var scatter = {
  'type': 'scatter',
  'error_y': {'array': error_z,
              'type': 'data',
              'visible': true
             },
  'x': x,
  'y': z
}

var layout = {'yaxis': {'type': 'log', 'title': 'z'},
              'xaxis': {'title': 'x'},
              'width': 800,
              'height': 800};

var data = [ scatter ];

// ## Plot scatter 2D
Plotly.newPlot('myDiv', data, layout);
```
newplot-3

In `scatter3d` the errors bars grow very large

```javascript
var scatter3d = {
  'type': 'scatter3d',
  'error_z': {'array': error_z,
              'type': 'data',
              'visible': true
             },
  'x': x,
  'y': y,
  'z': z
}

var data3d = [ scatter3d ];
var layout3d = {'scene': {'zaxis': {'type': 'log'}},
                'width': 800,
                'height': 800};

// ## Plot scatter 3D
Plotly.newPlot('myDiv', data3d, layout3d);
```
newplot-4

Perhaps `error_z.type='data'` is not implemented?

antoinerg:
Fixes #2962 

Applies axis transformation on error bars in scatter3d to deal with log scale appropriately.


src/traces/scatter3d/calc_errors.js
```diff
@@ -19,9 +19,10 @@ function calculateAxisErrors(data, params, scaleFactor) {
     for(var i = 0; i < data.length; i++) {
         var errors = computeError(+data[i], i);
 
+        var point = axis.d2l(data[i], 0, calendar) * scaleFactor; // A bit wasteful
         result[i] = [
-            -errors[0] * scaleFactor,
-            errors[1] * scaleFactor
+            (axis.d2l(data[i] - errors[0], 0, calendar) * scaleFactor) - point || -point,
+            (axis.d2l(data[i] + errors[1], 0, calendar) * scaleFactor) - point || -point
```
alexcjohnson: Since you brought `calendar` in here I'm assuming we're now starting to think about all axis types (including date and category), not just numeric (linear & log). At that level of generality, `d` formats cannot be added/subtracted, but `c` formats can. In fact since the errors are relative values, and for non-numeric axis types there isn't even really a concept of "zero" (see https://github.com/plotly/plotly.js/pull/2960#discussion_r215098710 for example) errorbar values should be assumed to be in `c` format (so milliseconds or number of categories)

Of course `+data[i]` from the `computeError` call above *also* won't work for these axis types, but that's actually a good thing, because it will prevent `'fraction'` and `'sqrt'` errorbars on these axes so you can only use `constant` or `data`. (hmm... except in pathological cases like when you have `'2018'` as a date or a bunch of numbers as "categories"... that's going to be a bug, we should really not try to do `+data[i]` on non-numeric axes and just pass `undefined` or something)

As you point out, all those `d2l` calls will be a bit of a drag... I suspect it'll be best to make a special case for log axes and leave the other types as they were.

antoinerg: As you suggested, I now make a special case for log axes and leave the other types as they were.

antoinerg:
I tried to reuse code from `scattergl` but it turns out it bails out of drawing error bars with negative lower bounds (see #3009)

etpinard:
Not the most ideal fix, but good enough considering the current state of 3D.

Reusing more things for cartesian is probably going to be difficult before we tackle https://github.com/plotly/plotly.js/issues/2646, which we may not tackle before the regl rewrite (https://github.com/plotly/plotly.js/issues/949), so let's get this fix in ASAP :dancer:

antoinerg:
Ok sounds good! I'll merge it now :)

Issue #2897, opened 2018-08-12
`showticklabels` in ternary is broken if false


nicolaskruchten:
In this pen https://codepen.io/nicolaskruchten/pen/XBGXXQ you can see some breakage in ternary plots when setting `showticklabels` to `false`.

PR #2993, opened 2018-09-10
Ternary w/o ticks fixes

etpinard:
Fixes https://github.com/plotly/plotly.js/issues/2897 

I initially wanted to fix this issue by implementing `editType: 'ticks'` in ternary axes (and polar axes which use `Axis.doTicks` in a similar way), but that turned out to be a little too messy for a bug-fix PR. I'm even starting to second-guess `editType: 'ticks'` all together. 

So, ternary axis tick and tick labels updates as now _as bad_ as for polar subplots - where they make use of a "state" fields (one for each axis) on the ternary subplot objects. Oh well, at least we now have a few tests locking the behavior down.

cc @alexcjohnson @antoinerg


src/plots/ternary/index.js
```diff
@@ -17,17 +17,29 @@ var TERNARY = 'ternary';
...
+        'If *ternary* (the default value), the data refer to `layout.ternary`.',
+        'If *ternary2*, the data refer to `layout.ternary2`, and so on.'
+    ].join(' ')
+};
 
-exports.layoutAttributes = require('./layout/layout_attributes');
+exports.layoutAttributes = require('./layout_attributes');
 
-exports.supplyLayoutDefaults = require('./layout/defaults');
+exports.supplyLayoutDefaults = require('./layout_defaults');
```
alexcjohnson: Good call flattening this structure 👍

test/jasmine/tests/ternary_test.js
```diff
@@ -382,6 +382,76 @@ describe('ternary plots', function() {
...
+            });
+        }
+
+        Plotly.plot(gd, fig)
+        .then(function() {
+            return toggle(
+                '.aaxis > .ytick > text',
+                'ternary.aaxis.showticklabels',
+                [true, false], [4, 0]
+            );
```
alexcjohnson: Minor style comment: you can put the extra function wrapping inside the definition of `toggle`, so that the individual calls are just `.then(toggle(...))`

Regardless `toggle` is a very nice simplification!

src/plots/ternary/ternary.js
```diff
@@ -430,6 +457,11 @@ proto.drawAxes = function(doTitles) {
     }
 };
 
+function strTickLayout(axLayout) {
+    return axLayout.ticks + String(axLayout.ticklen) + String(axLayout.showticklabels);
```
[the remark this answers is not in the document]
etpinard: `Axes.doTicks` is able to re-style existing nodes (e.g. relayout calls involving `tickfont`) e.g. see this test:

https://github.com/plotly/plotly.js/blob/2a667d03f1045b75c5c568bd30e33f28df36b9d5/test/jasmine/tests/ternary_test.js#L352-L383

but it can't handle hide/show. Cartesian axes rely on this block:

https://github.com/plotly/plotly.js/blob/2a667d03f1045b75c5c568bd30e33f28df36b9d5/src/plots/cartesian/axes.js#L1550-L1555

which isn't compatible with non-cartesian subplots.

test/jasmine/tests/ternary_test.js
etpinard: Good call. Done in -> 68e0628b20e2bf8f05eaac28e094e60584669f2b

alexcjohnson:
Nice fix. 💃

PR #1804, opened 2017-06-20
Enable touch interactions for select/lasso/others

dfcreative:
Resolves https://github.com/plotly/plotly.js/issues/1790

We have to avoid creating `dragCover` for touch-only devices, because `touchmove`, unlike `mousemove`, requires the same element where `touchstart` happened.

Is there any particular reason we don't use `document` for catching mouse events? What makes `dragCover` different from `document`?

@etpinard @alexcjohnson 

TODO:

* [x] `dragCover` does not hide cursor for panning and makes it a text cursor, we have to disable 
 selection in a way similar to [this](https://github.com/dfcreative/mucss/blob/master/selection.js)
* [x] panning/zooming do not work for scatter2d, that is related to [gl2d/camera](https://github.com/plotly/plotly.js/blob/master/src/plots/gl2d/camera.js)
* [x] no mode cursor for scatter2d #1489
* [x] panning is mixed up with page scrolling if body content is long
* [x] double click should reset the selection

PR #2997, opened 2018-09-11
DragElement: initialize clientX and clientY for touch events

### LINKED ISSUE #1967: Regression: touch to simulate hover on scatter plots
On touch devices, scatter plots used to have "touch to simulate hover" in Plotly 1.28.3, but this seems to have been removed in Plotly 1.29 and later. Is this a regression/bug?

Can touch to simulate hover (Plotly 1.28.3):
https://codepen.io/anon/pen/YxvzZR

Cannot touch to simulate hover (Plotly latest - 1.30.0 at time of writing):
https://codepen.io/plotly/pen/WvPgPP

The two examples are the same except for the Plotly version. Behaviour verified on Ipad running iOS 10.3.2 and a mobile phone running Android 6.0

jonfunkhouser:
Fixes #1967 

Fx.hover() uses evt.clientX and evt.clientY to calculate an offset for the hover. In a TouchEvent, these properties are undefined, which results in the hover being rejected.

https://github.com/plotly/plotly.js/blob/master/src/components/fx/hover.js#L285

### DIFF
--- a/src/components/dragelement/index.js
+++ b/src/components/dragelement/index.js
@@ -137,6 +137,12 @@ dragElement.init = function init(options) {
         initialEvent = e;
         rightClick = e.buttons === 2 || e.ctrlKey;
 
+        // fix Fx.hover for touch events
+        if(typeof e.clientX === 'undefined' && typeof e.clientY === 'undefined') {
+            e.clientX = startX;
+            e.clientY = startY;
+        }
+
         newMouseDownTime = (new Date()).getTime();
         if(newMouseDownTime - gd._mouseDownTime < DBLCLICKDELAY) {
             // in a click train
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -13,6 +13,7 @@ var click = require('../assets/click');
 var delay = require('../assets/delay');
 var doubleClick = require('../assets/double_click');
 var failTest = require('../assets/fail_test');
+var touchEvent = require('../assets/touch_event');
 
 var customAssertions = require('../assets/custom_assertions');
 var assertHoverLabelStyle = customAssertions.assertHoverLabelStyle;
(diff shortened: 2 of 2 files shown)


etpinard:
Thanks very much for this PR @jonfunkhouser !

It's probably the least-bad way to fix the regressions introduced in https://github.com/plotly/plotly.js/pull/1804. If I understand correctly, `plotly_hover` is now emitted correctly on touch and there's no way to trigger `plotly_click` on touch?

At some point, we'll probably need another attribute (e.g. `layout.touchmode`) or maybe new mobile-specific modebar buttons that could toggle between "hover on touch" / "click on touch" / "select on touch", but this seems like the correct "default" behavior, so :ok_hand: 

-----

Now, @jonfunkhouser have you checked that all the features added in https://github.com/plotly/plotly.js/pull/1804 still work on your branch?

jonfunkhouser:
It ends up emitting `plotly_click`, not `plotly_hover`. 

`Fx.click()` is called by dragElement's `onDone()` handler, which then calls `hover()` to determine what point was clicked: https://github.com/plotly/plotly.js/blob/master/src/components/fx/click.js#L22

The above `hover()` call was returning before `gd._hoverData` could be set, which then stopped `plotly_clicked` from being emitted.

I'm just now starting to validate the features in #1804, I'll reply again when that's complete.

etpinard:
> I'm just now starting to validate the features in #1804, I'll reply again when that's complete.

Thanks very much !

etpinard:
@jonfunkhouser I was able to test your patch on my Android phone. I must say: this is a wonderful PR.

I confirm that touching on points triggers the `plotly_click` event and does not trigger `plotly_hover` nor `plotly_unhover`. This is in fact how plotly.js behaved in `v1.28.3` the last release before https://github.com/plotly/plotly.js/pull/1804 which slipped in this regression. So this is fantastic :tada: 

Now, there's still no way to "hover" when `layout.dragmode` is either `'select'` and `'lasso'`. This is different from the behavior in `v1.28.3` where "hover" worked in any `dragmode`, **but** back then there was no way to select points!

Applying your patch on-top of the latest master (to include click-to-select from https://github.com/plotly/plotly.js/pull/2944) and setting `layout.clickmode: 'select'` disables "hover" on this branch. That's _ok_ as that never worked. This would be a cool feature though.

In brief, this PR:

- brings back the touch "hover" behavior of pre-#1804 in `dragmode` `'pan'` and `'zoom'`,
- ... while keeping all the touch "drag" behaviors implemented in #1804,
- touch "hover" still does not work in `dragmode` `'select'` and `'lasso'`
- touch "select" under `clickmode` `'select'` remains unimplement 

----

It would be nice to lock the touch hover behavior down so that regression like https://github.com/plotly/plotly.js/issues/1967 don't happen again. We do have few touch tests, for example this one that tests selections:

https://github.com/plotly/plotly.js/blob/5fa493aca6d820580aca6986fb505d4aad729911/test/jasmine/tests/select_test.js#L1052-L1085

@jonfunkhouser Would you be interesting in adding a touch hover test somewhere [`hover_label_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/hover_label_test.js)?

etpinard:
Thanks for adding that test @jonfunkhouser 

I pushed a small commit to fix up `npm run test-jasmine` https://github.com/plotly/plotly.js/pull/2997/commits/fca6f9b115a7b9f2eb35b0f32f0e263865754b3a

I'll now merge this PR and release it as part of 1.41.1

Issue #3003, opened 2018-09-12
Issue with fill in stackgroups


dmt0:
`fill: 'tonexty'` (the default when you stack) fills to the other trace even though they’re in different `stackgroups`. Should fill the first *of each stack* to zero.

https://codepen.io/dmt0/pen/dqemWo?editors=0010 newPlot @ 0

alexcjohnson:
Whew, I can make all sorts of fun stuff happen with multiple stackgroups... this example has them out of order in `gd.data`:
<img width="449" alt="screen shot 2018-09-12 at 4 36 31 pm" src="https://user-images.githubusercontent.com/2678795/45451796-29fde200-b6aa-11e8-9432-46907ae3610d.png">
But even if they're in the right order (ie all 'a' group together, then all 'b' group together, then all unstacked) there are still problems right now. Thanks for the report @dmt0!

Issue #3002, opened 2018-09-12
Error when deleting traces with stackgaps


dmt0:
When deleting traces with stackgaps:
```
factory.js?b9b2:113 Error while plotting: TypeError: Cannot read property ‘stackgaps’ of undefined
   at makePoints (plotly-with-meta.js?ab3a:163163)
   at plotOne (plotly-with-meta.js?ab3a:163270)
   at SVGGElement.eval (plotly-with-meta.js?ab3a:162817)
   at eval (plotly-with-meta.js?ab3a:19075)
   at d3_selection_each (plotly-with-meta.js?ab3a:19081)
   at Array.d3_selectionPrototype.each (plotly-with-meta.js?ab3a:19074)
   at Object.plot [as plotMethod] (plotly-with-meta.js?ab3a:162816)
   at SVGGElement.eval (plotly-with-meta.js?ab3a:116057)
   at eval (plotly-with-meta.js?ab3a:19075)
   at d3_selection_each (plotly-with-meta.js?ab3a:19081)
```

To reproduce
https://codepen.io/dmt0/pen/dqemWo?editors=0010 newPlot @ 0, react @ 1

Issue #932, opened 2016-09-12
Animate trace entrance/exit


rreusser:
[This stackoverflow question](http://stackoverflow.com/questions/39415791/using-plotly-js-animation-feature/39457872) makes it clear that people may expect `Plotly.animate` to animate the _entrance_ of traces. It wasn't the focus, so right now it only animates _existing_ traces. I think it's a reasonable thing people will definitely expect though, so opening this for discussion as a feature request.

See also: possible connection to #920 regarding entrance/exit of traces.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #3005, opened 2018-09-13
Fix stacked area layering and deletion bugs

alexcjohnson:
Lumped fixes for #3002 and #3003 into one PR, since they touch the same code.
- #3003 had two related components. First, which trace is "next" when we fill between two traces (and if there even *is* a next or should we fill to zero), stacked or otherwise, needs to take account of `stackgroup`. The "next" trace must be in the same stack group, or no stack group. Second, once you've done this, it doesn't make much sense to preserve the trace ordering as it is in `gd.data` exactly. Traces in the same stack group, or that fill to each other, need to be in consecutive order - or you can get weird effects like https://github.com/plotly/plotly.js/issues/3003#issuecomment-420801898. As mentioned there, I chose to push traces only backward to meet the first in their stack, but otherwise to preserve the existing order as much as possible. The new mock demonstrates this.
- #3002 pointed out that we were actually redrawing points even if we were about to remove their trace entirely! So not only does this work now, but deleting scatter traces will be faster in general now.


src/traces/scatter/plot.js
```diff
@@ -72,7 +72,7 @@ module.exports = function plot(gd, plotinfo, cdscatter, scatterLayer, transition
             });
         });
     } else {
-        scatterLayer.selectAll('g.trace').each(function(d, i) {
+        join.each(function(d, i) {
```
alexcjohnson: @etpinard it doesn't seem like animations can remove traces, is that correct? That's what it looks like, since I don't see a `join.exit()` associated with them (since `isFullReplot = !transitionOpts`)... yet the comment above implies that you can *add* traces:
```js
// Must run the selection again since otherwise enters/updates get grouped together
// and these get executed out of order. Except we need them in order!
```
So... do I need to worry about animations deleting traces or not?

etpinard: > do I need to worry about animations deleting traces or not?

Right, removing traces with `'frame.redraw': false` (i.e. without a replot at the end of the animation), doesn't work currently:

peek 2018-09-13 16-17

https://codepen.io/etpinard/pen/ZMRReB

so I don't think you broke anything here.

alexcjohnson:
> Would you mind adding a few words about this behavior in the fill attribute description?

Good call, I added it to both `fill` and `stackgroup` -> https://github.com/plotly/plotly.js/pull/3005/commits/fe0c988adfabf500961bb9a881b86c6b7c80d1c5

etpinard:
nicely done :dancer:

src/traces/scatter/plot.js
etpinard: Ah Ricky wrote an issue about this https://github.com/plotly/plotly.js/issues/932

PR #3011, opened 2018-09-14
scatter3d display errorbars when error_z is not specified

### LINKED ISSUE #2881: scatter3d error bars not displayed unless `error_z` is defined
If you look at this Pen https://codepen.io/nicolaskruchten/pen/QBZaYE you'll see two problems:

1. Error bars don't display unless `error_z` is there (i.e. just `x` or `y` or both gives broken output)
2. `thickness` doesn't seem to do anything

This seems to be true for the latest version as well as 1.38.0, which I randomly tested.

antoinerg:
Fixes #2881 by reverting to default values when error_z is not specified by the user.


src/traces/scatter3d/convert.js
```diff
@@ -124,9 +125,9 @@ function calculateErrorParams(errors) {
         if(e && e.copy_zstyle !== false) e = errors[2];
         if(!e) continue;
 
-        capSize[i] = e.width / 2;  // ballpark rescaling
+        capSize[i] = (e.width || 0) / 2;  // ballpark rescaling
         color[i] = str2RgbaArray(e.color);
-        lineWidth = e.thickness;
+        lineWidth = (e.thickness || errorBarsAttributes.thickness.dflt);
```
etpinard: Hmm. Strange, judging from

https://github.com/plotly/plotly.js/blob/e545415463730291201e0adadb876874680c4361/src/components/errorbars/defaults.js#L71

`fullData[i]._thickness` should have a set value by the time it gets here.

etpinard: Note that having values set in `fullData` is (very) important for the react-chart-editor (aka RCE). RCE uses `fullData` to fill in its UI components.

etpinard:
Looks good :dancer:

Issue #2938, opened 2018-08-22
Update zero line logic for sploms


alexcjohnson:
#2936 fixes zero line logic for regular cartesian plots, but large sploms do their own thing for performance. As @etpinard points out, this should be udpated too:

> No need this do in this PR (opening a new issue would be fine), but we'll need to propagate these changes to large-sploms' regl-based grid line logic:
> 
> https://github.com/plotly/plotly.js/blob/b5f8b2300d041f12df90647fe61dba93f7d10c4b/src/traces/splom/base_plot.js#L171-L183
> 
> where I didn't just reuse the `axes.js` logic because I wanted to avoid looping over the traces (to compute that `hasBarsOrFill` boolean). Thinking about this again, looping over traces isn't really a performance hit for sploms (I can't really imagine a user wanting to plot more 10 splom traces on the same graph).
> 
> So it would be worth trying to expose and :recycle: this logic here for large-sploms.

PR #3015, opened 2018-09-14
Update splom zero line logic

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2938

cc @antoinerg @alexcjohnson


test/image/mocks/splom_large.json
```diff
@@ -297,6 +297,14 @@
     "yaxis": {
       "zerolinecolor": "red",
       "zerolinewidth": 2
+    },
+    "xaxis2": {
+      "zeroline": false
+    },
+    "yaxis2": {
```
etpinard: which currently renders as:

image

notice the missing zero line on the second row from the top.

alexcjohnson:
Nice, thanks! 💃

Issue #2898, opened 2018-08-13
`dtick` in geo not react()'ing


nicolaskruchten:
As seen in this pen https://codepen.io/nicolaskruchten/pen/ZjPLOX it looks like `geo.lonaxis.dtick` isn't `react()`ing to changes as I would expect... newPlot@0 then react@1 shows no change (though it should) and then react@2 does (as it does if you change anything other than `dtick`, in this case `gridwidth`).

Issue #2353, opened 2018-02-08
open items uncovered during Plotly.react push


alexcjohnson:
Some rules that were good ideas before but become crucial with `Plotly.react` so that we recognize a non-change correctly:
- Never create new arrays in the `supplyDefaults` step - for example reshaping 1D to 2D arrays. That should be done in a `calc` step and NOT pushed back into `_fullData`.
- Don't modify attributes after finishing `supplyDefaults`, unless you're going to push the results back to the input (`data`/`layout`) as well, like `autorange`/`range`.
- When you're stashing things on `_full*`, always make `_private` attributes.

Most of the instances of this are taken care of in #2341, but a few are open:
- [ ] `tick0`, `dtick` -> default filled in late (After SD) - also for tickmode=array, not needed at all
- [ ] `camera` doesn’t push back into `_fullLlayout` after drag, only `layout`?
- [ ] tests for `scatter3d` with uneven or missing x/y/z
- [ ] tests for `surface` with uneven, missing, or ragged x/y/z
- [x] `autocolorscale`: first time it gets set on, second time off - is that intentional? cauto sometimes shows a similar effect… see `gl3d_scatter-color-mono-and-palette` **resolved in:** https://github.com/plotly/plotly.js/pull/3341
- [ ] `contourcarpet`: `zmin`/`zmax` get filled in (after `supplyDefaults`) but they aren't actually attributes of `contourcarpet`. They should be!
- [ ] `autorange: 'reversed'` should get replaced in `layout` once it has been acted on. There's a polar mock where this doesn't happen until the second draw.

etpinard:
Referencing https://github.com/plotly/plotly.js/pull/2341/files#r166474049 about possibly improving _implied_ edits. 

Pasting content from https://github.com/plotly/plotly.js/pull/2341/files#r166675758:

>  think we should try and find a way to help the user access the recommendations we would make with impliedEdits and clearAxisTypes, but leave it up to them whether to follow those recommendations.

Perhaps all we need is to (under a certain config option) log whenever we mutate user data or user layout. Alternatively, `Plotly.validate` could be use here e.g.

```js
Plotly.validate([{
  y: [1, 2, 1]
}], {
  xaxis: {
     autorange: true,
     range: [1, 2]
  }
})
```

could log something like: _increasing range detected, range values ignored under autorange_. Then together with https://github.com/plotly/plotly.js/issues/1741, it should make it fairly easy for plotly.js users to debug.

PR #3016, opened 2018-09-14
Don't skip over dtick and tick0 in geo subplot in diffLayout

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2898, a better fix would be to make `dtick` and `tick0` for all axes _just work_ with `Plotly.react` with no added logic (that's one of the items in https://github.com/plotly/plotly.js/issues/2353), but that will be for another day.

cc @alexcjohnson @antoinerg


alexcjohnson:
💃

PR #3017, opened 2018-09-15
Fix #3013 - broken stacked area gap insertion case

### LINKED ISSUE #3013: Stacked Area cross-trace calc error
This figure causes an error and doesn't render, as far as I can tell because the second trace doesn't have `x`:

```
{
  "data": [
      {
        "type": "scatter",
        "x": [1,2,3],
        "y": [1,2,3],
        "stackgroup": 1,
      },
      {
        "type": "scatter",
        "y": [1,2,3],
        "stackgroup": 1
      }
    ],
    "layout": {},
    "frames": []
  }
```

This one fails for a different reason (orientation=h) with a similar stack trace:

```
{
  "data": [
    {
      "type": "scatter",
      "orientation": "h",
      "stackgroup": 1,
      "y": [
        16.99,
        10.34,
        21.01,
        23.68,
        24.59,
        25.29,
        8.77,
        26.88,
        15.04,
        14.78,
      ]
    },
    {
      "type": "scatter",
      "mode": "markers",
      "stackgroup": 1,
      "y": [
        1.01,
        1.66,
        3.5,
        3.31,
        3.61,
        4.71,
        2,
        3.12,
        1.96,
        3.23,
      ]
    }
  ],
  "layout": {},
  "frames": []
}
```

alexcjohnson:
@etpinard @antoinerg the perils of nested functions - I had started out reusing the variable name `cd` in the inner scope, then thought better of it and made a new local var `calcTrace` but didn't update all references...


etpinard:
Ah good catch. :dancer: 

Maybe we should consider enabling https://eslint.org/docs/rules/no-shadow? That would require some work though, as `no-shadow` fails in more than 500 times in our codebase.

alexcjohnson:
> Maybe we should consider enabling https://eslint.org/docs/rules/no-shadow?

mmm, interesting, that's a good idea, I'll make a `maintenance` issue for it.

Issue #2303, opened 2018-01-28
Latex does not display properly in legend


LiteracyFanatic:
Using Latex markup in the names of legend entries results in text which overlaps the marker. A codepen showing the problem can be found [here](https://codepen.io/anon/pen/vdBmGL) but the same problem is present in the plots used for the [documentation](https://plot.ly/javascript/LaTeX/) as well.

![image](https://user-images.githubusercontent.com/18249574/35477926-e096244c-0394-11e8-9fb8-b41b9b9a76d0.png)

alexcjohnson:
Thanks @LiteracyFanatic - This seems to have been broken for a while, prior to v1.28 it works. Possibly broke in #1783?

etpinard:
Related https://github.com/plotly/plotly.js/issues/966

etpinard:
Fixed in https://github.com/plotly/plotly.js/pull/3018

Issue #966, opened 2016-09-22
MathJax makes the legend scroll unnecessarily


rreusser:
http://codepen.io/rsreusser/pen/mARpJd?editors=1010

<img width="198" alt="2016-09-22 10 48 51" src="https://cloud.githubusercontent.com/assets/572717/18753106/6269cf12-80b2-11e6-8ba2-31b54ff73461.png">

I get a scrollbar when my legend includes mathjax. In this case it's not catastrophic, but in [another example](http://demos.rickyreusser.com/elastodynamics/complex.html), I just had to drop the mathjax altogether because the legend names ended up blank if they included any mathjax.

etpinard:
Oh MathJax

rreusser:
Yeah, I'm assuming that it's because of asynchronous measuring/rendering passes. I mainly raise this because I feel like it's an argument in favor of KaTeX if the support ever becomes viable.

etpinard:
Yeah. We've been wanting to switch over to KaTeX for about two years now.

I wonder how hard that would be?

etpinard:
Most likely fixed as part of https://github.com/plotly/plotly.js/pull/2426

PR #996, opened 2016-09-29
add streaming maxpoints max and dflt

cldougl:
Add max and default vals for `maxpoint` attribute for streams
so the values shown [here](https://github.com/plotly/streambed/blob/f95f3303b3600451902efa6d86bada58b9f727ee/shelly/streaming/streamingviews.py#L97) make it into the reference pages

PR #3018, opened 2018-09-15
Fix support for Latex in legends

jonmmease:
This PR introduces a few small fixes for MathJax Latex names in the legend.  I know this has worked in the past, but it has not worked consistently in my testing. When it "doesn't work" the entire legend is not shown, and there are SVG console errors complaining that `rect` width/height cannot be set to NaN.

I believe this to be a concurrency issue that is addressed in 74328f3.  The issue is that the call to `drawTexts` (line 123 in `src/components/legend/draw.js`) triggers an async rendering call for MathJax text, and then the call to `computeLegendDimensions` (new line 130) assumes the text has already been rendered.

Commit f23eb74 introduces no new logic, it just applies the indent that was omitted in 74328f3 to make the logical changes clearer.

Commit 1482077 applies the same horizontal offset to latex legend values as are applied to non-latex legend values (this was zero previously). And then I moved this to a more appropriate location and refactored to extract the constant in https://github.com/plotly/plotly.js/pull/3018/commits/222357c57b3267f32e54080e12c3d52506a7347e.

Commits acf5d7b  and 6c27224 re-enable the Latex legend string in the mathjax mock.  The mock now renders consistently as follows:

mathjax


etpinard:
Looking good to my :eyes: 

Thanks for splitting commits [f23eb74](https://github.com/plotly/plotly.js/commit/f23eb74d422395c246600622f95103131dcf9749) and  [74328f3](https://github.com/plotly/plotly.js/commit/74328f359d829bdd3205cc9d0803b8efd79ac1d1) making this PR very easy to review.

etpinard:
Does this fix https://github.com/plotly/plotly.js/issues/2303 or/and https://github.com/plotly/plotly.js/issues/966 ?

jonmmease:
The alignment issue in #2303 is fixed by
> Commit 1482077 applies the same horizontal offset to latex legend values as are applied to non-latex legend values (this was zero previously). And then I moved this to a more appropriate location and refactored to extract the constant in 222357c.

I don't think the scrollbar issue in #966 is reproducible in the current master.  I've never seen a scroll bar (with or without this fix), but I can confirm that the particular example in the codepen of #996 renders fine now and without a scroll bar. So I'd call it fixed 🙂

alexcjohnson:
> I don't think the scrollbar issue in #966 is reproducible in the current master.

#966 seems to have disappeared between 1.34 and 1.35 despite the fact that they both have the horizontal alignment problem. Likely #2426 fixed the scrollbar issue, but since MathJax was unusable in legends until this PR, I'll let you close #966 when this PR is merged.

etpinard:
Awesome work @jonmmease 

This one will be released today as part of `1.41.1`.

etpinard:
I'm all set to release 1.41.1, so I'll go ahead and merge this PR. I hope @jonmmease won't mind.

Thanks again for this fix!

xhlulu:
Amazing! I was waiting for this one to fix my regression app. Any idea when 1.41.1 will be bundled for plotly.py? Thanks!

jonmmease:
@xhlulu it will be in plotly.py 3.3 which should be out by the end of next week. If plotly.js releases a 1.41.2 next week we'll try to line up with that.

xhlulu:
Thank you

PR #3021, opened 2018-09-17
Make polar be more graceful on non sense inputs

etpinard:
This PR fixes two issue-less bugs discovered by @dmt0:

- `polar.hole=1` case https://codepen.io/dmt0/pen/pOVdKx
- `polar.sector` that spans more than 360 degrees https://codepen.io/dmt0/pen/LJmOQV

This PR also:

- adds a small perf boost on main-drag mouse-move,
- fixes a bug-less typo in the splom defaults (discovered by @bpostlethwaite )
- tags the `plotly_webglcontextlost` test in `gl2d_plot_interact_test.js` as `@noCI`.
- tags a newly-added select test (in the click-to-select PR) as `@flaky`

The last two items should help with the (several) intermittent test failures we've been seeing lately.

cc @alexcjohnson @antoinerg


src/plots/polar/polar.js
```diff
@@ -420,24 +423,31 @@ proto.updateRadialAxis = function(fullLayout, polarLayout) {
...
-    .selectAll('path').attr('transform', null);
-
-    updateElement(layers['radial-line'].select('line'), radialLayout.showline, {
-        x1: innerRadius,
-        y1: 0,
-        x2: radius,
-        y2: 0,
-        transform: trans
-    })
+    updateElement(
```
etpinard: Recall that `updateElement` is 

https://github.com/plotly/plotly.js/blob/7f0fc265ff57a85974d347b772e00e17368ab941/src/plots/polar/polar.js#L1347-L1355

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -215,7 +215,7 @@ describe('Test gl plot side effects', function() {
         .then(done);
     });
 
-    it('@gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
+    it('@noCI @gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
```
etpinard: 4 successes in 4 `test-jasmine2` runs with this `@noCI`

test/jasmine/tests/polar_test.js
```diff
@@ -628,6 +631,35 @@ describe('Test relayout on polar subplots:', function() {
...
+            });
+        }
+
+        Plotly.plot(gd, [{
+            type: 'scatterpolar',
+            r: ['a', 'b', 'c']
+        }])
+        .then(function() { _assert('base', true); })
+        .then(function() { return Plotly.relayout(gd, 'polar.hole', 1); })
+        .then(function() { _assert('hole=1', false); })
```
alexcjohnson: So it doesn't fail... does it work? ie draws all data on the rim regardless of radial value (within the radial range)? Do you want to include this in a mock?

etpinard: Here's `hole=1` subplot:

peek 2018-09-17 17-11

where angular drag still drag still works.

> ie draws all data on the rim regardless of radial value (within the radial range)?

it shows no data points.

> Do you want to include this in a mock?

done in https://github.com/plotly/plotly.js/pull/3021/commits/cb5b3368ed47f1db3904543006e07bba37d8f956

alexcjohnson: > it shows no data points.

This comment is nonblocking (perhaps add to #2255 if you don't want to address it here):

With `cliponaxis: false` (which is the default anyway for polar) I would have expected it to show the data - and it currently doesn't, even if you explicitly specify a range. The argument for this behavior is to have `hole: 1` match `hole: 0.9999` as closely as possible. Perhaps not useful, though I guess I could see someone wanting to just show where their data is on an angular scale without involving a radial scale at all. That said they can always just use `hole: 0.9999` as a workaround, so it's not *really* any missing functionality... but it feels hacky.

alexcjohnson:
Very nice bunch of cleanup @etpinard! 💃

test/jasmine/tests/polar_test.js
etpinard: > The argument for this behavior is to have `hole: 1` match `hole: 0.9999` as closely as possible.

Ah I see. Good point.

> Perhaps not useful, though I guess I could see someone wanting to just show where their data is on an angular scale without involving a radial scale at all

I'll wait until someone complains about it (maybe @nicolaskruchten ?) before addressing this. Now in -> https://github.com/plotly/plotly.js/issues/2255#issuecomment-422409171

PR #3023, opened 2018-09-18
Fix #3022 - validate + polar.bargap

### LINKED ISSUE #3022: Polar bargap looks invalid to Plotly.validate
I made this plot:
```js
Plotly.newPlot(gd,[
  {type:'barpolar',r:r,theta:th,marker:{color:c,colorscale:'Jet'},subplot:'polar2'},
  {type:'scatterpolar',mode:'markers',theta:mth,r:mr,marker:{size:ms,color:mc,colorscale:[[0,'#600'],[1,'#f00']],sizeref:0.3}}
], {
  polar2: {domain:{x:[0,1],y:[0,1]},bargap:0.5,hole:0.83,radialaxis:{visible:false},angularaxis:{ticks:'',showline:false,nticks:40}},
  polar: {bgcolor:'#d3d3d3',hole:0.2,domain:{x:[0.1,0.9],y:[0.1,0.9]},
    radialaxis:{gridcolor:'#fff',showline:false,showticklabels:false,ticks:''},
    angularaxis:{gridcolor:'#fff',nticks:40,showline:false,ticks:'',showticklabels:false}}
})
```
But although it works correctly, `Plotly.validate complains`:
```js
Plotly.validate(gd.data, gd.layout)
LOG: In layout, key polar2.bargap is not part of the schema
```

@etpinard comments:
> ah right, that's a first
> a trace layout attribute inside a subplot container

(cc @Kully - you can ignore that one if you run `Plotly.validate`)

etpinard:
@alexcjohnson @antoinerg this solution isn't as strict as it could be (see comment in test), but at least it fixes the false-negative case of #3022.

### DIFF
--- a/src/plots/polar/legacy/area_attributes.js
+++ b/src/plots/polar/legacy/area_attributes.js
@@ -12,19 +12,24 @@ var scatterAttrs = require('../../../traces/scatter/attributes');
 var scatterMarkerAttrs = scatterAttrs.marker;
 var extendFlat = require('../../../lib/extend').extendFlat;
 
-var deprecationWarning = 'Area traces are deprecated!';
+var deprecationWarning = [
+    'Area traces are deprecated!',
+    'Please switch to the *barpolar* trace type.'
+].join(' ');
 
 module.exports = {
     r: extendFlat({}, scatterAttrs.r, {
         description: [
             deprecationWarning,
-            scatterAttrs.r.description
+            'Sets the radial coordinates',
+            'for legacy polar chart only.'
         ].join(' ')
     }),
     t: extendFlat({}, scatterAttrs.t, {
         description: [
             deprecationWarning,
-            scatterAttrs.t.description
+            'Sets the angular coordinates',
+            'for legacy polar chart only.'
         ].join(' ')
     }),
     marker: {
--- a/src/plots/polar/legacy/area_attributes.js
+++ b/src/plots/polar/legacy/area_attributes.js
@@ -10,15 +10,48 @@
 
 var scatterAttrs = require('../../../traces/scatter/attributes');
 var scatterMarkerAttrs = scatterAttrs.marker;
+var extendFlat = require('../../../lib/extend').extendFlat;
+
+var deprecationWarning = 'Area traces are deprecated!';
 
 module.exports = {
-    r: scatterAttrs.r,
-    t: scatterAttrs.t,
+    r: extendFlat({}, scatterAttrs.r, {
+        description: [
+            deprecationWarning,
+            scatterAttrs.r.description
+        ].join(' ')
+    }),
+    t: extendFlat({}, scatterAttrs.t, {
+        description: [
+            deprecationWarning,
+            scatterAttrs.t.description
+        ].join(' ')
+    }),
     marker: {
-        color: scatterMarkerAttrs.color,
-        size: scatterMarkerAttrs.size,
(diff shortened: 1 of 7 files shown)


test/jasmine/tests/validate_test.js
```diff
@@ -539,4 +539,36 @@ describe('Plotly.validate', function() {
...
+            out[0], 'unused', 'layout', null, ['polar3'], 'polar3',
+            'In layout, container polar3 did not get coerced'
+        );
+
+        // Plotly.validate should be more strict here.
+        //
+        // It should log an 'unused' warning for `polar2.bargap`,
+        // but trace layout attribute set in subplot containers are considered
+        // valid as long as one trace on that graph has those trace layout
+        // attributes.
```
alexcjohnson: In fact it's not just `Plotly.validate` that should be more strict - `supplyDefaults` coerces `polar2.bargap` unnecessarily, and if it didn't do that, `Plotly.validate` would at least return `'unused'` - which you might even argue is a better message than `'schema'` in this case, but that's a secondary concern. Might be worthwhile fixing this for `supplyDefaults` for the editor's sake.

etpinard: Thanks for the tip.

Done in -> https://github.com/plotly/plotly.js/pull/3023/commits/c5cf89b71cd091a5b742fb2a80618672cce34cc7

src/plots/polar/legacy/axis_attributes.js
```diff
@@ -13,6 +13,8 @@ var axesAttrs = require('../../cartesian/layout_attributes');
 var extendFlat = require('../../../lib/extend').extendFlat;
 var overrideAll = require('../../../plot_api/edit_types').overrideAll;
 
+var deprecationWarning = 'Legacy polar charts are deprecated!';
```
etpinard: Should _fix_ https://github.com/plotly/documentation/issues/1095

src/plots/polar/legacy/area_attributes.js
```diff
@@ -10,15 +10,48 @@
...
 var scatterAttrs = require('../../../traces/scatter/attributes');
 var scatterMarkerAttrs = scatterAttrs.marker;
+var extendFlat = require('../../../lib/extend').extendFlat;
+
+var deprecationWarning = 'Area traces are deprecated!';
 
 module.exports = {
-    r: scatterAttrs.r,
-    t: scatterAttrs.t,
+    r: extendFlat({}, scatterAttrs.r, {
```
etpinard: Follow-up on https://github.com/plotly/plotly.js/pull/3023/commits/1c0d3ed0512ef826bf2e6dc9d576bccfff2f18f3

alexcjohnson: Ah great, nice pattern for adding these warnings!

`scatterAttrs.r` already [asks users to switch to `scatterpolar`](https://github.com/plotly/plotly.js/blob/2a3dc2656d3c832ceb24b3c22c9155aa02fd7c7a/src/traces/scatter/attributes.js#L548) - so here we should switch that so the deprecation warning points people to `barpolar`. Similarly for the layout attrs, can we point people to `polar` subplots?

etpinard: new attempt: https://github.com/plotly/plotly.js/pull/3023/commits/019bacfc88069f51133f624c49063e5b01c94d58

alexcjohnson:
Love it. :dancer:

Issue #2940, opened 2018-08-22
Prevent dragging legend out of plot div?


brian428:
Currently, if a plot is editable and set to a constrained size, it's possible to drag the legend out of the plot div and permanently "lose" it. It would be nice if the drag behavior was changed to prevent this, or to provide a config option to turn on constraints on dragging the legend (or anything, really) out of the plot div.

alexcjohnson:
Can you demonstrate in a codepen? I can't figure out how to reproduce the effect you're describing. By default the margins expand to fit wherever you drag it, but if I turn that off (`margin.autoexpand: false`) it always seems to jump back onto the plot when I let go of it.
![legend-off-edge](https://user-images.githubusercontent.com/2678795/44485789-c1c65e00-a61f-11e8-96d7-0dade754d75e.gif)

alexcjohnson:
@brian428 we still have not been able to reproduce this. Can I close it or do you have a demonstration of the issue?

alexcjohnson:
Closing due to lack of activity. Feel free to reopen if anyone can reproduce.

PR #3024, opened 2018-09-18
improve legend positioning in edge cases

### LINKED ISSUE #2131: Negative positions get ignored on horizontal wrapped legends
https://codepen.io/etpinard/pen/ZaEEyL?editors=0010

First reported on https://community.plot.ly/t/legend-positioning-outside-plot/6528/3

### LINKED ISSUE #2337: The legend loses his position on a small screen with lot of items and horizontal orientation
If you add a lot of items, then set the legend horizontal (`layout.legend.orientation = 'h'`) and the width of the graphic to a small value (like a smartphone), the legend (the scrollbox) is wrongly positioned.

See the [JSFiddle](https://jsfiddle.net/6oq39dpy/6/).

antoinerg:
This PR improves how we deal with:
- positioning large legends in small figures (closes #2337)
- legends with negative x positions (closes #2131)

To achieve this, I made the following changes:
- in `autoMargin`: if margins get larger than half the figure's size, scale down the margins instead of setting them to zero!
- honors positioning even if legends are large by removing lines `lx = gs.l;` and `ly = gs.t;`


antoinerg:
cc @alexcjohnson @etpinard

src/components/legend/draw.js
```diff
@@ -170,10 +170,9 @@ module.exports = function draw(gd) {
 
             // Make sure the legend left and right sides are visible
             var legendWidth = opts._width,
-                legendWidthMax = gs.w;
+                legendWidthMax = gs.w * Math.max(1 - opts.x, 1);
```
alexcjohnson: This calculation makes sense for the (default) `xanchor: 'left'` - seems like there's something similar for `'right'` like `* Math.max(opts.x, 1)`, not sure what it should look like for `'center'`, and `'auto'` should presumably depend on which alignment it auto-chose.

(and similarly with `legendHeightMax` and `yanchor`)

alexcjohnson:
Can you check what happens with `config: {editable: true}` when you drag the legend around and bump up against these various limits? Does the legend end up where you expected after you drop it? Is it ever possible to drag it totally off the edge and lose it (cc #2940)? If you're in one of these situations (esp. off the edge where the legend has increased the margin) and you drag it just a few pixels, is the result always a small change in the expected direction or does it jump one way or another?

I don't think this needs any new tests added, unless you encounter and fix any weird behavior.

antoinerg:
As you know, I encountered some weird behavior and I will need to better spec out how legend positioning/sizing should be handled to submit a solid PR. To be continued!

etpinard:
> To be continued!

Thanks for the update!

Braintelligence:
@etpinard Is there any ETA on merging this? 😸

etpinard:
Not soon, unfortunately.

etpinard:
@antoinerg I'll like to bring this PR back to life. Don't worry, I handle all the work :wink: 

Back in October, you wrote:

> I encountered some weird behavior and I will need to better spec out how legend positioning/sizing should be handled to submit a solid PR

Could you elaborate on the "weird behavior" you encountered. I suspect merging master might help. PR https://github.com/plotly/plotly.js/pull/3323 fixed a few "odd" auto-margin scenarios.

antoinerg:
I would have to refresh my memory. I do remember however having troubles with drag and dropping the legend in a new position. Once dropped, it would drastically change position.

Let me get back to you on this one

etpinard:
PR https://github.com/plotly/plotly.js/pull/4160 essentially supersedes this PR here. Ok if we close it?

antoinerg:
> PR #4160 essentially supersedes this PR here. Ok if we close it?

Sure! Thanks for taking this on :)

Issue #3026, opened 2018-09-19
Double click zoom out behaviour


pfbuxton:
**Moved from plotly.py**

Hello,

Here is a movie of the problem behaviour (taken from https://plot.ly/python/zoom-events/; but it's the same using any plotly):
![temp1](https://user-images.githubusercontent.com/4923588/45748960-24e8e780-bc02-11e8-9a0d-1b5fb09131c7.gif)

1. Zoom in by selecting a region.
2. Zoom out by double clicking.
3. Zoom in by horizontal selection.

The problem is that stage 3 also changes the vertical zoom back to that in stage 1.

alexcjohnson:
Confirmed - thanks for the report @pfbuxton!

etpinard:
Hmm. That's bad. I guess I should try to fix this thing ASAP.

Assigned.

etpinard:
This is a regression that split in 1.40.0, most likely from https://github.com/plotly/plotly.js/pull/2823

PR #3028, opened 2018-09-19
Reset updates object on every zoomDone

etpinard:
... so that 'old' keys from two-sided zoomboxes don't trickle into single-side zoombox updates.

Fixes https://github.com/plotly/plotly.js/issues/3026, which was made apparent staring from `v1.40.0` (and PR https://github.com/plotly/plotly.js/pull/2823), as we no longer run a `calc` edit when autoranging cartesian axes (e.g on double-click).

This one is fairly bad bug, so I'll make a release as soon as this thing is merged.

cc @alexcjohnson @antoinerg

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -85,6 +85,8 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
     var editX, editY;
     // graph-wide optimization flags
     var hasScatterGl, hasOnlyLargeSploms, hasSplom, hasSVG;
+    // collected changes to be made to the plot by relayout at the end
+    var updates;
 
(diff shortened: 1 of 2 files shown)


src/plots/cartesian/dragbox.js
```diff
@@ -386,6 +385,8 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
     }
 
     function zoomDone() {
+        updates = {};
```
etpinard: Note that `updates` is also "reset" in 

https://github.com/plotly/plotly.js/blob/7c03fb46d002da6fd85fde6b9cefe23f6d72568e/src/plots/cartesian/dragbox.js#L598-L647

which is called on pan and scroll.

PR #3030, opened 2018-09-20
Fix pie's support for individual stroke width values [2989]

### LINKED ISSUE #2989: Giving stroke / border on one specific marker in Pie Chart
I read the doc at [https://plot.ly/javascript/reference/#pie](https://plot.ly/javascript/reference/#pie), it is said that we can provide an array of number (stroke's width) to give on a specific marker in pie chart.
image
But if we provide an array in "**width**" field like this:
```javascript
var data = [{
  values: [20, 26, 55],
  labels: ['Residential', 'Non-Residential', 'Utility'],
  type: 'pie',
  pull: [0.1, 0, 0],
  marker: {
    colors: ['rebeccapurple', 'purple', 'mediumpurple'],
    line: {
      // WORKS
      // width: 3
      // ERROR
      width: [3, 0, 0]
    }
  }
}];

```
It will produce such an error:
image

What i'm trying to achieve is something like this: (harcoded in inspection tools)
image

Codepen Link: https://codepen.io/rmdwirizki/pen/vzdNjp

rmoestl:
Fixes #2989

### DIFF
--- a/src/components/legend/get_legend_data.js
+++ b/src/components/legend/get_legend_data.js
@@ -59,7 +59,8 @@ module.exports = function getLegendData(calcdata, opts) {
                         label: labelj,
                         color: cd[j].color,
                         i: cd[j].i,
-                        trace: trace
+                        trace: trace,
+                        pts: cd[j].pts
                     });
 
                     slicesShown[lgroup][labelj] = true;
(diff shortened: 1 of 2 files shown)


etpinard:
Brilliant fix :dancer:

Issue #2968, opened 2018-09-04
`autosize:true` doesn't respect div's max-height when its height isn't specified


antoinerg:
`autosize: true` resizes a chart according to the parent container's width and height. 

However, if the parent container's size is determined by css' `max-height`, the chart is too tall and overflows.

Example https://codepen.io/antoinerg/pen/BOZWxx/

antoinerg:
Closed by #3033

PR #3033, opened 2018-09-20
fallback to max-width and max-height if necessary

antoinerg:
Simple fix to issue #2968

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -1387,8 +1387,8 @@ plots.plotAutoSize = function plotAutoSize(gd, layout, fullLayout) {
         // but don't enforce any ratio restrictions
         var computedStyle = isPlotDiv ? window.getComputedStyle(gd) : {};
 
-        newWidth = parseFloat(computedStyle.width) || fullLayout.width;
-        newHeight = parseFloat(computedStyle.height) || fullLayout.height;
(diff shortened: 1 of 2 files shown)


src/plots/plots.js
```diff
@@ -1387,8 +1387,8 @@ plots.plotAutoSize = function plotAutoSize(gd, layout, fullLayout) {
         // but don't enforce any ratio restrictions
         var computedStyle = isPlotDiv ? window.getComputedStyle(gd) : {};
 
-        newWidth = parseFloat(computedStyle.width) || fullLayout.width;
-        newHeight = parseFloat(computedStyle.height) || fullLayout.height;
+        newWidth = parseFloat(computedStyle.width) || parseFloat(computedStyle.maxWidth) || fullLayout.width;
+        newHeight = parseFloat(computedStyle.height) || parseFloat(computedStyle.maxHeight) || fullLayout.height;
```
etpinard: I'm curious. Can we have a situation where `computedStyle.height` > `computedStyle.maxHeight`?

If so, here we honour `computedStyle.height`, but shouldn't `computedStyle.maxHeight` win out? 

(This is probably more a of theoretical question than a real-life situation).

etpinard:
> It looks like `computedStyle.height` will never be greater than `computedStyle.maxHeight` because it's the computed style ie. css logic was applied.

Thanks for the info!

Let's :dancer:

antoinerg:
Closes #2968

PR #3040, opened 2018-09-21
Polar 'text' in hover labels fixes

etpinard:
Thanks for the report @alexcjohnson !


alexcjohnson:
That was quick! Looks fantastic. 💃

Issue #2970, opened 2018-09-05
Hovering on boxes/violins shows only median label under "opposite" hovermode


konieczny9lives:
When box plot is the only trace, all statistics (_min,q1,median,q3,max_) are displayed on hover.
If you add another trace as a subplot, only _median_ is displayed on hover.

This only happens with **horizontal** box plot as far as I know.

Reproduction:
https://codepen.io/kotwgarnku/pen/gdxaPN

alexcjohnson:
Thanks @konieczny9lives! There's also a very strange interaction between the two plots in that codepen, that sometimes (for example, click "show closest data on hover" in the top modebar) the hover labels for the top plot show up at the positions they should have for the bottom plot !?!
![screen shot 2018-09-05 at 9 08 31 am](https://user-images.githubusercontent.com/2678795/45095283-b4af7180-b0eb-11e8-8ade-f7135dcced45.png)

konieczny9lives:
Great catch @alexcjohnson!

In regard to the original issue, I just found out that if you add `hovermode: "closest"` to layout options, all statistics (_min,q1,median,q3,max_) are displayed as intended.

alexcjohnson:
Well, *mostly* as intended... they're showing up horizontal (which requires them to be weirdly shifted vertically)
![screen shot 2018-09-05 at 9 32 53 am](https://user-images.githubusercontent.com/2678795/45096530-ba5a8680-b0ee-11e8-946a-2f7c1f52c1aa.png)


when they should be showing up tilted, which seems to me much more readable:
![screen shot 2018-09-05 at 9 31 38 am](https://user-images.githubusercontent.com/2678795/45096481-a44cc600-b0ee-11e8-97f4-88b1d8f9f6db.png)

etpinard:
Related: https://github.com/plotly/plotly.js/issues/2194

stargazer33:
It seems, this is the same problem as here:
https://community.plot.ly/t/boxplot-in-a-subplot-very-limited-information-in-hover/13109

```
It is a “subplot”; a histogram and a boxplot together, they share the X axis.
When I position the mouse cursor over boxplot I see ONLY the median values of both boxplots at the same time!
```

etpinard:
After some investigation, lines:

https://github.com/plotly/plotly.js/blob/2f0d56eaf71c4e979203d542a5e08233156ac68d/src/components/fx/hover.js#L580-L582

,

[...]

Issue #2194, opened 2017-12-06
non-linear box/violin hover label quirks


alexcjohnson:
Horizontal violins have some odd hover label behavior. Labels are horizontal, not tipped as they are for horizontal boxes and bars, and (perhaps related) they're anchored to the middle of the violin, not the edge (or off the edge) as they are for vertical violins
![screen shot 2017-12-06 at 11 59 07 am](https://user-images.githubusercontent.com/2678795/33674572-f96df4ee-da7d-11e7-8b8e-92c4dedf4339.png)

And at least in `violin_non-linear` the KDE label disappears past the max point, even though the KDE line is still there.
![screen shot 2017-12-06 at 11 59 35 am](https://user-images.githubusercontent.com/2678795/33674570-f639af7a-da7d-11e7-9222-237bcefee2b6.png)

cc @etpinard

etpinard:
This seems to happen on _non-linear_ axes for both violin and box traces.

Mock [`box_grouped_horz`](https://rreusser.github.io/plotly-mock-viewer/#box_grouped_horz) + `Plotly.restyle(gd, 'type', 'violin')` appears to work fine. Whereas [`violin_non-linear`](https://rreusser.github.io/plotly-mock-viewer/#violin_non-linear) + `Plotly.restyle(gd, 'type', 'box')` also shows the problem.

PR #3043, opened 2018-09-24
Rotate hover labels in a few more scenarios

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2194 and might offer a solid enough workaround for https://github.com/plotly/plotly.js/issues/2970 to close it.

cc @alexcjohnson @antoinerg

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -237,6 +237,10 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         vLinePoint: null
     };
 
+    // does subplot have one (or more) horizontal traces,
+    // (to determine whether we rotate the labels or not)
+    var hasOneHorizontalTrace = false;
+
     // Figure out what we're hovering on:
     // mouse location or user-supplied data
 
@@ -256,6 +260,9 @@ function _hover(gd, evt, subplot, noHoverEvent) {
             trace = cd[0].trace;
             if(trace.hoverinfo !== 'skip' && helpers.isTraceInSubplots(trace, subplots)) {
                 searchData.push(cd);
+                if(trace.orientation === 'h') {
+                    hasOneHorizontalTrace = true;
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/components/fx/hover.js
```diff
@@ -237,6 +237,10 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         vLinePoint: null
     };
 
+    // does subplot have one (or more) horizontal traces,
+    // (to determine whether we rotate the labels or not)
+    var hasOneHorizontalTrace = false;
```
etpinard: I'm having some seconds thoughts about this here.

Under `layout.hovermode: 'closest'`, subplots that have one (or more) "horizontal" traces get rotated labels. I'm starting to think this might be an overkill for subplots with horizontal bar traces. Perhaps, only horinzontal box and violin traces should lead to rotate the labels?

Any objections?

alexcjohnson: I think you're right, a single trace of horizontal bars (or histogram) still looks better with unrotated labels. But as soon as you start showing multiple labels (which happens for a single trace of box or violin) it's important to rotate the labels.

etpinard: Here's what I came up with in: https://github.com/plotly/plotly.js/pull/3043/commits/d9f3db7b44c3a7ff13ee4cd79c5a99ba87a3e2fe

alexcjohnson:
💃 
Lets leave #2970 open though, I'm not convinced we can't still do better in a backward compatible way. Does this PR fix the multi-plot misalignment issue there though?

etpinard:
> Does this PR fix the multi-plot misalignment issue there though?

It fixes https://github.com/plotly/plotly.js/issues/2970#issuecomment-418732786:

peek 2018-09-25 13-34

etpinard:
> Lets leave #2970 open though,

Sure. I'll rename the issue accordingly.

Issue #2981, opened 2018-09-07
Hoverlabel custom color bugs


etpinard:
From @nicolaskruchten :

While testing this I also found some cases where the hoverlabel text doesn't automatically contrast with the `bgcolor` in a sufficient way ... setting `layout.hoverlabel.bgcolor = "white"` gives this, where issue 1 is that the bottom overlabel doesn't have black text and issue 2 is that the trace-name portion of the label doesn't have black text. Setting `layout.hoverlabel.font.color = "black"` only corrects issue 1.

![image](https://user-images.githubusercontent.com/203523/44963084-d6470800-aef5-11e8-87fe-e4744cf49d6a.png)

nicolaskruchten:
Pen to show this easily: https://codepen.io/nicolaskruchten/pen/pOdRbo

Issue #2342, opened 2018-02-06
hover label customizations


chriddyp:
I'm working on Dash Themes and I'd like to customize the look and feel of the hover labels more. In order of importance:
1 - Customize the background and text color of the trace name. Currently the background is semi-transparent and the text color is dark, rendering the name of the trace illegible. 
![screen shot 2018-02-06 at 5 49 51 pm](https://user-images.githubusercontent.com/1280389/35888968-ad86b51e-0b67-11e8-9d31-bcd222e75dd8.png)
![screen shot 2018-02-06 at 5 49 39 pm](https://user-images.githubusercontent.com/1280389/35888969-ad96402e-0b67-11e8-9204-380bd75dcf25.png)
2 - Customize border color and width
3 - Generic CSS? For example, it'd be nice if I could experiment with box-shadow like https://bl.ocks.org/sarubenfeld/56dc691df199b4055d90e66b9d5fc0d2 in the same style as my theme's buttons:
![image](https://user-images.githubusercontent.com/1280389/35889241-ba9116e0-0b68-11e8-958c-c07afe7ebbfb.png)

brianstamper:
Possibly related, allow control over hover text alignment independent of hover placement position left/right/center, as mentioned here https://github.com/plotly/plotly.js/issues/260 and here https://stackoverflow.com/questions/50003531/r-plotly-hover-label-text-alignment

etpinard:
In brief, we should add attributes:

- `hoverlabel.namecolor`
- `hoverlabel.bgnamecolor`
- `hoverlabel.borderwidth` (`hoverlabel.bordercolor` is already a thing)

etpinard:
and let's merge in https://github.com/plotly/plotly.js/issues/1665

- `hoverlabel.opacity`

kristianmandrup:
Could you please add `hoverlabel.nameborderwidth` as well :)

antoinerg:
@chriddyp suggested supporting generic CSS. That would cover all possible styling option of CSS available now and in the future. Something like `hoverlabel.style` or `hoverlabel.css` ? What do you think?

alexcjohnson:
> supporting generic CSS

Sure, we support this already in our pseudo-HTML - ie `<span style="...">styled text</span>` so there's precedent for it. I suspect there will be easy ways to use this to break our fragile sizing and positioning system, but you'll know pretty quickly if that has happened 😅 And I'm sure people would find used for it!

[...]

PR #3048, opened 2018-09-25
Fix hover label coloring on white bgcolor

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2981, by doing two things:

- contrast common hover label font color w/ bgcolor
- do not use `hoverlabel.bgcolor` to determine for "name" part of hover label

cc @alexcjohnson @antoinerg 

More to come in https://github.com/plotly/plotly.js/issues/2342, but this is probably the best we can do until then.

### DIFF
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -2062,7 +2062,7 @@ describe('Test hover label custom styling:', function() {
             });
             assertCommonLabel({
                 path: ['rgb(255, 255, 255)', 'rgb(68, 68, 68)'],
-                text: [13, 'Arial', 'rgb(255, 255, 255)']
+                text: [13, 'Arial', 'rgb(68, 68, 68)']
             });
         })
         .then(function() {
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/hover_label_test.js
```diff
@@ -2062,7 +2062,7 @@ describe('Test hover label custom styling:', function() {
             });
             assertCommonLabel({
                 path: ['rgb(255, 255, 255)', 'rgb(68, 68, 68)'],
-                text: [13, 'Arial', 'rgb(255, 255, 255)']
```
alexcjohnson: Ha yeah, that would be a problem! FWIW, we'd have had a *chance* anyway of catching this bug when we wrote the test if it said something like
```
path: {bg: 'rgb(255, 255, 255)', border: 'rgb(68, 68, 68)'}, // or fill & line...
text: [13, 'Arial', 'rgb(255, 255, 255)']
```
So it'd be immediately obvious that the path bg matches the font color, and the font cannot be seen!
As it is `path` has two colors so it's unclear what they mean. `text` is unambiguous because all items are different types, so it's OK as an array, though still perhaps awkward as you need to know the order to *write* the test.

Issue #2555, opened 2018-04-13
Splom open items


etpinard:
Follow-up from https://github.com/plotly/plotly.js/pull/2505 and https://github.com/plotly/plotly.js/pull/2527

### Things that were left out

- Splom traces placed on overlaid axes are broken, see https://github.com/plotly/plotly.js/pull/2505#discussion_r181237788
- Multi-selection across subplot is disallowed. To do so, we would need to share `selectedpoints` info across multiple splom-generated subplots, see https://github.com/plotly/plotly.js/pull/2505#issuecomment-380855752

### Things we could do, but would require some thought

- Add _sub-trace_ on the diagonal (e.g. with `diagonal.mode: 'scattergl' || 'histogram' || 'box' || 'violin'`, see https://github.com/plotly/plotly.js/pull/2505#discussion_r177823962
- Add attribute in `dimensions` that could default the splom generated axes' style (e.g. `dimensions[i].showgrid: false` would remove grid lines from all axes corresponding to `dimensions[i]`)
- Make one single cartesian `<g .draglayer>` covering the whole plot area instead of multiple (12 I think) per subplot. This would speed up initial rendering (especially for splom, but only for all graphs with many subplots), make cross-subplot selection a little easier to manage, see https://github.com/plotly/plotly.js/pull/2505#discussion_r181237581
- optimize `regl-line2d` for grid data

### Related issues that would benefit splom performance:

- faster axis autorange relayout: https://github.com/plotly/plotly.js/issues/2546
- per-axis 'axrange' relayout: https://github.com/plotly/plotly.js/issues/2547
- stash pts selections for pan: https://github.com/plotly/plotly.js/issues/2548
- speed up cartesian axis defaults: https://github.com/plotly/plotly.js/issues/2549

etpinard:
Pasting benchmarks from https://github.com/plotly/plotly.js/pull/2505#issuecomment-381665092 out for future reference:

---------

Here are some first-render benchmarks with `showupperhalf: false` and `diagonal.visible: false`:

| # of dimensions | # of values per dimension | regl-line2d grid lines? | time [ms] |
| --------------------- | ---------------------------------- | -------------------- | ------------ |
| 10 | 20k | no | 400-500 |
| 20 | 5k | yes | 600-700 |
| 30 | 2k | yes | 900-1000 |
| 40 | 1k | yes | 1300-1400 |
| 50 | 800 | yes | 2000-2300 |

computed using:

```js
[...]

Issue #2553, opened 2018-04-13
Implement 'style' editType for scattergl and splom traces


etpinard:
Currently, scattergl and splom traces have to go through `editType: 'calc'` pathway to update style attributes. We should open up a new pathway to for things like `marker.color` and `line.width` that doesn't have to go through all the data type conversion done in the `calc` step.

etpinard:
Splom part done in https://github.com/plotly/plotly.js/pull/3057

Shouldn't be too hard to implement this for `scattergl` now.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #2562, opened 2018-04-16
Implement better (more maintainable) way to redraw gl canvases on incremental updates


etpinard:
Now that we have multiple regl-based traces sharing the same gl canvases and relying on the same global `src/lib/clear_gl_canvases.js` routine, we'll need a better way to redraw things on updates (e.g. on selections but also restyle and relayout calls). 

Here's an example from #2505, `style` here is called during selections:

https://github.com/plotly/plotly.js/blob/a7b08b326e208b90779ec714de92b25955937fdd/src/traces/splom/index.js#L416-L441

Here, we clear the canvases and redraw splom things with new selected/unselected batches, but then ooops we also need to add logic for scattergl traces in the splom method so that they don't disappear on splom updates.

--------

So, I propose (and I'm sure @dy will find a way to improve this proposal) to store a graph-wide _queue_ of regl module _draw_ calls (e.g. `fullLayout._reglDrawQueue`) that we could call everything we clear the canvases to update something. The new update sequence will (1) clear the canvases (or only parts of it if we can) (2) update selected buffers (3) call `reglDrawQueue` to redraw all things.

etpinard:
Done in https://github.com/plotly/plotly.js/pull/3067, merged in master during https://github.com/plotly/plotly.js/pull/3057

PR #3057, opened 2018-09-28
Aggressive splom perf

etpinard:
Completes the splom part of https://github.com/plotly/plotly.js/issues/2553 and checks off two O(n^2) items in https://github.com/plotly/plotly.js/issues/2555.

Each commit here is fairly self-contained. The most _aggresive_ optimizations are in https://github.com/plotly/plotly.js/commit/17e30d2f5f5bdb6296f90fdceea514b8c5dd8518 (aka `supplyDefaults`-bypass) and https://github.com/plotly/plotly.js/commit/4792f08771750a421e3e33895f602580ffa7415b (aka no-need-for-`<rect.bg>`-nor-subplot-`<clipPath>`) Note that commit https://github.com/plotly/plotly.js/commit/3574ece2a1eff74e1009055d9bb1ca699bafc1f9 was required to not have to re-init the splom objects on `editType: 'style'` updates.

In terms of performance gains, using:

```js
var Nvars = 50
var Nrows = 800
var dims = []

for(var i = 0; i < Nvars; i++) {
  dims.push({values: []})
  for(var j = 0; j < Nrows; j++) {
     dims[i].values.push(Math.random())
  }
}

Plotly.purge(gd);

console.time('splom')
Plotly.newPlot(gd, [{
  type: 'splom',
  dimensions: dims,
  showupperhalf: false,
  diagonal: {visible: false}
  // totaling 1225 subplots
}], {width: 2000, height: 2000},
   {showAxisDragHandles: false, showAxisRangeEntryBoxes: false}
)
console.timeEnd('splom')
```

and comparing against `v1.41.3` on Chrome 69 (which significantly slower than Chrome 66, by the way), we get gains of:

- ~200 ms on first render, thanks to the no-need-for-`<rect.bg>`-nor-subplot-`<clipPath>`
- ~200 ms on `relayout(gd, 'xaxis.range', [/* */])`, thanks to https://github.com/plotly/plotly.js/commit/cc5f0caafe230c21b8f3f25d3b7b9b244be5ebf8 and the `supplyDefaults`-bypass
- huge improvements for `restyle(gd, 'marker.color', /* scalar */)`: ~150ms vs 1000ms
- huge improvements (but not good enough) for `restyle(gd, 'marker.size', /* scalar */)`: ~400ms vs 1000ms
- huge improvements (but not good enough) for `restyle(gd, 'marker.color', '/* arrayOk */)`: ~800ms vs 2000ms

cc @alexcjohnson @antoinerg @archmoj


etpinard:
Here are some benchmarks in the splom pipeline off this branch. At 50 dims and 800 rows, we clock:

- `supplyDefaults`: 100-150 ms
- `doCalcdata`: ~20 ms
- `drawFramework` (where the `<g.subplot>` are added to the DOM): 200-300 ms
   + room for improvement here!
- `lsInner` (where subplot, bg, lines and clip paths are updated): 40-60 ms
- `drawAxes`: 300-400 ms
- `initInteractions` (where the drag layers are added to the DOM): 
   + ~100-150 ms (with `showAxisDragHandles` and `showAxisRangeEntryBoxes` false)
   + ~1000 ms (default),
- `createREGL`: ~300-400 ms on a 2000x2000px canvas
- splom grid update/draw: ~150-200 ms, 
   + we could add `dimensions[i].axis.showgrid: false` to skip this
- splom init (the `createMatrix` call): ~200 ms (not sure why)
- splom update/draw: ~400-600 ms

which gives a total range of 1800 to 2500 ms.

src/traces/splom/base_plot.js
```diff
@@ -168,46 +167,40 @@ function makeGridData(gd) {
...
-            if(
-                trace.type === 'splom' &&
-                scene && scene.matrix && scene.matrix.destroy
-            ) {
-                scene.matrix.destroy();
-                cd0.t._scene = null;
+            if(oldFullLayout._splomScenes) {
+                var scene = oldFullLayout._splomScenes[oldTrace.uid];
+                if(scene && scene.destroy) scene.destroy();
+                // must first set scene to null in order to get garbage collected
```
alexcjohnson: interesting... how did you figure this out?

src/traces/splom/base_plot.js
```diff
@@ -168,46 +167,40 @@ function makeGridData(gd) {
...
-    for(i = 0; i < newModules.length; i++) {
-        if(newModules[i].name === 'splom') {
-            hasSplom = true;
-            break;
-        }
-    }
+                if(oldTrace.uid === newTrace.uid && newTrace.type === 'splom') {
+                    continue oldLoop;
+                }
+            }
```
alexcjohnson: unlikely to matter much, as having a large number of traces on a splom plot would be weird... but you could reduce this to two separate loops, one to collect the `uid` of each `newFullData` `splom` trace into an object `newUids = {uid1: 1, uid2: 1}` and the second would just do `if(newUids[oldTrace.uid]) continue`

src/plot_api/plot_api.js
```diff
@@ -1738,8 +1738,8 @@ function addAxRangeSequence(seq, rangesAltered) {
     // subroutine of its own so that finalDraw always gets
     // executed after drawData
     var doTicks = rangesAltered ?
-        function(gd) { return subroutines.doTicksRelayout(gd, rangesAltered); } :
-        subroutines.doTicksRelayout;
+        function(gd) { return Axes.doTicks(gd, Object.keys(rangesAltered), true); } :
+        function(gd) { return Axes.doTicks(gd, 'redraw'); };
```
alexcjohnson: > just calling 'Axes.doTicks' instead of full `doTicksRelayout` which includes drawing the title

Is that sufficient though? If you move the ticks and as a result the title gets pushed away a different amount to avoid them, is that handled correctly?

src/plot_api/plot_api.js
```diff
@@ -1701,15 +1701,26 @@ exports.relayout = function relayout(gd, astr, val) {
         seq.push(subroutines.layoutReplot);
     }
     else if(Object.keys(aobj).length) {
-        Plots.supplyDefaults(gd);
+        var flagList = Object.keys(flags).filter(function(k) { return flags[k]; });
+
+        // Optimization mostly for large splom traces where
+        // Plots.supplyDefaults can take > 100ms
+        if(flagList[0] === 'axrange' && flagList.length === 1) {
```
alexcjohnson: Interesting... this seems safe to me, I'm mostly frustrated that we're locking ourselves into the "older" `relayout` for performance here, `react` won't be able to match it. Not to say we shouldn't do this - this is a fantastic optimization! But it might be worthwhile looking carefully at `supplyDefaults`, if there are any perf gains to be had there it would benefit everybody. Not in this PR of course...

src/plot_api/subroutines.js
```diff
@@ -103,25 +116,21 @@ function lsInner(gd) {
...
     var lowerDomains = [];
-    subplotSelection.each(function(d) {
-        var subplot = d[0];
-        var plotinfo = fullLayout._plots[subplot];
+    // no need to draw background when paper and plot color are the same color,
+    // activate mode just for large splom (which benefit the most from this
+    // optimization), but this could apply to all cartesian subplots.
+    var noNeedForBg = (
+        fullLayout._hasOnlyLargeSploms &&
+        fullLayout.paper_bgcolor === fullLayout.plot_bgcolor
```
alexcjohnson: oh interesting - nice find! I guess this is what you're implying, but would `paper_bgcolor === plot_bgcolor` be sufficient? Is there something special about `splom` in this regard?

Also I guess to be precise we need to know that the color is fully opaque too (or fully transparent)... though a semitransparent `paper_bgcolor` is a strange and dangerous thing.

`overlappingDomain` might itself be a heavy lift for a large number of subplots (am I reading this right that for NxN subplots this is O(N^4), ie O(N^2) tests per subplot???), all for the fairly unusual case of inset plots, or otherwise partially occluding subplots. Perhaps it would be worthwhile doing a faster check for whether `overlappingDomain` is even necessary to call, by looping over all x axes to see if any of their domains overlap, then the same over all y axes? That would only be O(N^2) to possibly (usually!) avoid O(N^4).

src/plot_api/subroutines.js
```diff
@@ -466,12 +466,34 @@ exports.drawMainTitle = function(gd) {
...
+
+        var arraysToCalcdata = _module.arraysToCalcdata;
+        if(arraysToCalcdata) arraysToCalcdata(cd, trace);
+
+        var editStyle = _module.editStyle;
+        if(editStyle) editStyleCalls.push({fn: editStyle, cd0: cd0});
+    }
+
+    if(editStyleCalls.length) {
+        clearGlCanvases(gd);
```
alexcjohnson: Is this going to cause problems if there's a `scattergl` or other regl trace along with `splom`? Do we need to implement `editStyle` for all of these at once?

alexcjohnson:
> - `initInteractions` (where the drag layers are added to the DOM): 
>    + ~100-150 ms (with `showAxisDragHandles` and `showAxisRangeEntryBoxes` false)
>    + ~1000 ms (default),

Wow - time to experiment with a single drag element that registers many separate regions?

alexcjohnson:
Fantastic work @etpinard - very impressive results!

Just two questions of safety https://github.com/plotly/plotly.js/pull/3057#discussion_r221407615 https://github.com/plotly/plotly.js/pull/3057#discussion_r221413342 and a couple of potential small improvements https://github.com/plotly/plotly.js/pull/3057#discussion_r221406493 https://github.com/plotly/plotly.js/pull/3057#discussion_r221411382

The single drag element is already mentioned in #2555 and obviously should not be done here, that should be its own PR.

src/traces/splom/base_plot.js
etpinard: Checking GPU memory consumption in the Chrome task manager (`<shift>+<esc>`) before and after calling `Plotly.purge(gd)`.

src/plot_api/plot_api.js
etpinard: > But it might be worthwhile looking carefully at supplyDefaults, if there are any perf gains to be had there it would benefit everybody. Not in this PR of course...

Yeah for sure. https://github.com/plotly/plotly.js/issues/2549 will remain open after this PR.

src/plot_api/subroutines.js
etpinard: > Is there something special about `splom` in this regard?

There's nothing special about splom in this regard. I added a check for  `_hasOnlyLargeSploms` for two (somewhat poor) reasons:

- I was scared of potentially missing a funky inset/subplot scenario
- many jasmine tests were failing as the custom assertion `assertPlotSize` relies on the [`<rect.bg>` bounding box](https://github.com/plotly/plotly.js/blob/5c62c8e937c92f3c267722b29613ebe2a15413c2/test/jasmine/assets/custom_assertions.js#L264) and I couldn't find a good replacement for it.

> Also I guess to be precise we need to know that the color is fully opaque too (or fully transparent)... though a semitransparent paper_bgcolor is a strange and dangerous thing.

Good call. I'll add a check for this.

> overlappingDomain might itself be a heavy lift for a large number of subplots (am I reading this right that for NxN subplots this is O(N^4), ie O(N^2) tests per subplot???), all for the fairly unusual case of inset plots, or otherwise partially occluding subplots. Perhaps it would be worthwhile doing a faster check for whether overlappingDomain is even necessary to call, by looping over all x axes to see if any of their domains overlap, then the same over all y axes? That would only be O(N^2) to possibly (usually!) avoid O(N^4).

That loop clocks in at 20-40ms at 50 dimensions, so I didn't bother optimizing it for now. You're right, it is an expansive loop, but it doesn't query nor append anything on the DOM, so it's not too too slow.

But yeah, it might be worth experimenting restricting the `hasOnlyLargeSploms` regime to graphs with only splom-generated cartesian subplots (i.e. no insets, no overlaps). This may help us speed up `drawFramework` too.

src/plot_api/subroutines.js
etpinard: Good point. I might _have_ implement `editType: 'style'` for scattergl traces in this PR.

A better solution would be to implement https://github.com/plotly/plotly.js/issues/2562

src/traces/splom/base_plot.js
alexcjohnson: Alright, I have to wonder
- How general that observation is - does this apply across browsers/OS's? does it apply when the console is closed?
- Assuming it does apply in some real-world situations, where else should we be doing the same thing? Would it be worthwhile making a `Lib.delete(container, key)` or something to encourage reuse?

src/plot_api/subroutines.js
alexcjohnson: OK, other than the semitransparent check these are nonblocking for this PR but please put removing the `_hasOnlyLargeSploms` check and avoiding the O(N^4) loop into issues somewhere so we don't forget about them.

src/plot_api/plot_api.js
etpinard: I think this is sufficient. The `drawMainTitle` subroutine 

https://github.com/plotly/plotly.js/blob/e646042b40a4fc123cdd84d4e525ebe64df0c088/src/plot_api/subroutines.js#L428-L441

doesn't attempt to scoot to title away from tick labels.

alexcjohnson: Ah you're right - I missed that this was only about the main title, not the axis titles 👍

src/plot_api/plot_api.js
etpinard: Added supply-default bypass to `Plotly.update` in https://github.com/plotly/plotly.js/pull/3057/commits/6de1ae4f1da1f5d84fd5c8d95d27f688132e2b17

src/plot_api/subroutines.js
etpinard: >  other than the semitransparent check these are nonblocking for this PR but please put removing the _hasOnlyLargeSploms check 

Commit https://github.com/plotly/plotly.js/pull/3057/commits/aaf63127bbad48adef96a19a4d456b7e66ce68b5 implements this optimization to all cartesian subplots and add a check for semi-transparent bg colors.

> avoiding the O(N^4) loop into issues somewhere so we don't forget about them.

Now in https://github.com/plotly/plotly.js/issues/2555#issuecomment-426343601

src/traces/splom/base_plot.js
etpinard: Thanks for the tip. Done -> https://github.com/plotly/plotly.js/pull/3057/commits/478c6693ecb639f8b718b0ab5a608cefd6920c55

src/plot_api/subroutines.js
```diff
@@ -125,7 +125,8 @@ function lsInner(gd) {
     // activate mode just for large splom (which benefit the most from this
     // optimization), but this could apply to all cartesian subplots.
     var noNeedForBg = (
-        fullLayout._hasOnlyLargeSploms &&
+        Color.opacity(fullLayout.paper_bgcolor) === 1 &&
+        Color.opacity(fullLayout.plot_bgcolor) === 1 &&
```
alexcjohnson: you don't have to test both, do you? you're just going to check that they match below anyway.

But also if `Color.opacity(fullLayout.plot_bgcolor) === 0` regardless of `paper_bgcolor` - if the user has explicitly said "make the plot transparent" we don't need these rects even if there *is* overlap.

test/jasmine/assets/custom_assertions.js
```diff
@@ -261,7 +258,7 @@ exports.assertPlotSize = function(opts, msg) {
     var widthLessThan = opts.widthLessThan;
     var heightLessThan = opts.heightLessThan;
 
-    var plotBB = d3.select('.bglayer .bg').node().getBoundingClientRect();
+    var plotBB = d3.select('.plotclip > rect').node().getBoundingClientRect();
```
alexcjohnson: ah nice - wouldn't work if we were testing *position*, but it works for *size*

src/plot_api/subroutines.js
etpinard: See https://github.com/plotly/plotly.js/pull/3067

src/plot_api/subroutines.js
etpinard: > you don't have to test both, do you?

Ha I knew that you were going to make that comment! I left it that way because I thought it made things nicely verbose and self-documented w/o sacrificing perf.

> But also if `Color.opacity(fullLayout.plot_bgcolor) === 0` regardless of `paper_bgcolor` - if the user has explicitly said "make the plot transparent" we don't need these rects even if there _is_ overlap.

Good point. An even better improvement would be to make one `<path d="">` to draw all bg rectangles as a single node. Ok if I leave this for latter?

alexcjohnson: OK, to my mind it's actually more confusing to read this way but I'm not going to insist. Sure, feel free to leave out the transparent case and possible node combination as long as they're recorded in an issue somewhere.

alexcjohnson:
Beautiful. 💃

etpinard:
Improvemens (e.g. no axis type clearance) over https://github.com/plotly/plotly.js/pull/3057/commits/078c0864014fb93c8ca8b0525353144a9d0eece0 in -> https://github.com/plotly/plotly.js/compare/splom-perf...splom-axis-dim-match

test/jasmine/tests/splom_test.js
```diff
@@ -785,30 +907,48 @@ describe('Test splom interactions:', function() {
...
+                fullYaxes: ['linear', 'category', 'category', 'log']
+            });
+
+            return Plotly.restyle(gd, {
+                'showlowerhalf': false,
+                'diagonal.visible': false
+            });
+        })
+        .then(function() {
+            // TODO should we delete layout.xaxis and layout.yaxis4 here?
```
alexcjohnson: no, we should leave them there in case the user brings diagonal or lowerhalf back later

src/traces/splom/defaults.js
```diff
@@ -158,9 +158,10 @@ function handleAxisDefaults(traceIn, traceOut, layout, coerce) {
         }
     }
 
-    // when lower half is omitted, override grid default
-    // to make sure axes remain on the left/bottom of the plot area
-    if(!showLower) {
+    // when lower half is omitted, or when just the diagonal is gone,
+    // override grid default to make sure axes remain on
+    // the left/bottom of the plot area
+    if(!showLower || (!showDiag && showUpper && showLower)) {
```
alexcjohnson: I could imagine `splom` *always* setting the defaults to `'bottom'` and `'left'`... but I guess in principle this way at least if you're in one of the situations that doesn't need this, you could make a bigger grid than the splom and put other things in other cells without having to tweak the axis positions. So yeah, nice fix 🎉

alexcjohnson:
Nice perseverance, excellent additions. Lets do it! 💃

src/traces/splom/defaults.js
etpinard: > I could imagine `splom` _always_ setting the defaults to `'bottom'` and `'left'`

Yeah, that's how it was initially, but https://github.com/plotly/plotly.js/commit/c947b2c0070833199ceff56fc4bafc59d539f3d1#diff-f4f66cea15079ea63f9be7a90528570c added this block for perf benefits.

PR #3067, opened 2018-10-02
Add redrawReglTraces subroutine

etpinard:
resolves https://github.com/plotly/plotly.js/issues/2562, to be merged into https://github.com/plotly/plotly.js/pull/3057 to addressed @alexcjohnson 's https://github.com/plotly/plotly.js/pull/3057#discussion_r221413342

In brief, this PR makes regl-based traces coexist better. All their "draw" calls are done in one single subroutine. This preserves ordering and makes it easier to redraw the canvas on drag and selections. For now on, regl-based traces (excluding `parcoords`) only update their buffers during `_module.plot` and do not do any "drawing".

cc @dy @archmoj @antoinerg


test/jasmine/tests/plots_test.js
```diff
@@ -153,26 +153,6 @@ describe('Test Plots', function() {
 
             testSanitizeMarginsHasBeenCalledOnlyOnce(gd);
         });
-
-        it('should sort base plot modules on fullLayout object', function() {
```
alexcjohnson: It's great that we don't need to sort anymore, but what was the issue that forced us to sort before? Can we do something to test that the underlying issue has been solved?

etpinard: It's already covered in https://rreusser.github.io/plotly-mock-viewer/#splom_with-cartesian where the blue scattergl pt in the lower-left quadrant sits on top of the orange splom trace.

There's also:

https://github.com/plotly/plotly.js/blob/9772ef6dd5ec5417ab50dd378aaa536a8c845336/test/jasmine/tests/splom_test.js#L1063-L1099

src/plot_api/subroutines.js
```diff
@@ -613,6 +612,61 @@ exports.drawData = function(gd) {
...
+// but all traces need to be redrawn following clearGlCanvases.
+//
+// Note that _module.plot for regl trace does NOT draw things
+// on the canvas, they only update the buffers.
+// Drawing is perform here.
+//
+// TODO try adding per-subplot option using gl.SCISSOR_TEST for
+// non-overlaying, disjoint subplots.
+//
+// TODO try to include parcoords in here.
```
alexcjohnson: Does it break if a parcoords trace is mixed in with other regl traces? Or is this TODO just an optimization?

etpinard: > Does it break if a parcoords trace is mixed in with other regl traces

It's no worse than it is now. parcoords + other regl trace works fine on first-render, but breaks during cartesian drag and selection.

alexcjohnson:
Beautiful cleanup @etpinard - dunno if you want someone better versed in the regl code to weigh in, but from my standpoint this is ready to 💃 !

src/plot_api/subroutines.js
etpinard: This was discussed during the regl-based scattergl push, but I couldn't find a dedicated issue on this topic, so -> https://github.com/plotly/plotly.js/issues/3069

Issue #2703, opened 2018-06-06
Include valid config attributes with plot schema


cpsievert:
It would help me throw informative errors/warnings in the R package.

etpinard:
Good call. I tried that once (in https://github.com/plotly/plotly.js/pull/1188), but was a little too ambitious with that PR.

etpinard:
could be a great community contribution.

jonmmease:
This would also be helpful for plotly.py.  We've fallen behind the valid list of config options a couple of times now. e.g. https://github.com/plotly/plotly.py/issues/1290

PR #3068, opened 2018-10-02
modeBar styling

### LINKED ISSUE #3063: Modebar aesthetic configurability
Would like the modebar to have a few more programmatic options, especially for extremely styled Dash apps:

1. Programmatic ability to set the modebar in a vertical orientation. Bokeh does a nice job with this:
modebar_vertical

2. Programmatic ability to set the background bar color, background bar opacity, and icon colors:
modebar_plotly

3. While we're at it, update the plotly logo in the modebar

Related: 
https://github.com/plotly/plotly.js/issues/2171 
https://github.com/plotly/streambed/issues/9699
(Perhaps we should create a plotly/plotly.js "modebar" label?)

antoinerg:
Initial attempt at closing #3063 

There are several little changes in different places and some may not be satisfactory. This initial implementation will be a good starting point to further discuss how to resolve #3063.

Note that currently, this PR doesn't have the hover effect on buttons. To do this in CSS, we would need to inject a new `<style>` element onto the page to style the pseudo-class `:hover`. Alternatively, we could change the fill color in javascript onmouseover. Let me know what you would prefer.


src/components/modebar/modebar.js
```diff
@@ -173,31 +180,42 @@ proto.createButton = function(config) {
...
+        if(thisIcon.color) {
+            path.setAttribute('fill', thisIcon.color);
+        }
+
+        icon.appendChild(path);
     }
 
-    icon.appendChild(path);
+    if(thisIcon.svg) {
+        icon.innerHTML = thisIcon.svg;
```
antoinerg: `test-syntax` fails because or `innerHTML` :(

etpinard: This should be fine. `innerHTML` only causes problems on SVG nodes. 

Looks like the concerns written down in 

https://github.com/plotly/plotly.js/blob/9772ef6dd5ec5417ab50dd378aaa536a8c845336/tasks/test_syntax.js#L106-L109

have become reality.

antoinerg:
BTW, here's what it looks like with a vertical orientation and black icons:
chrome_2018-10-02_17-02-54

src/components/modebar/modebar.js
antoinerg: `thisIcon.svg` isn't a SVG document but rather its content. `icon` is the SVG document so IE doesn't like this. Let me fix this :)

antoinerg: Fixed in commit e870e34

antoinerg: I got rid of `innerHTML` altogether and now use `DOMParser` in commit 61d1c7b !

src/fonts/ploticon/ploticon.svg
```diff
@@ -27,5 +27,37 @@
...
+      .cls-2 {
+        fill: #25fefd;
+      }
+
+      .cls-3 {
+        fill: #fff;
+      }
+    </style>
+  </defs>
+  <title>plotly-logomark</title>
```
etpinard: Cool. Can you commit `build/ploticon.js` and `build/plotcss.js`?

alexcjohnson:
If we think people will really change the logo color, then it would be worth using the monochrome version. But I suspect anyone who cares that much what the modebar looks like is going to drop the plotly logo entirely, so we should optimize the look of the existing logo, which is the current multicolor version - so I'd vote to keep the color.

Do we want to reverse the order of the vertical modebar? We should at least move the logo to the top corner, but I think it might work better - at least for people already used to the horizontal one - to have the whole thing flipped as if we had just rotated the existing modebar around the logo.

antoinerg:
@alexcjohnson I agree we should move the logo to the top. 

As for the icons, I'd rather have the + icon be to the left or above the - icon. Which is why I prefer the current order of icons. But this is a matter of preference. BTW, it's quite easy to change the order in the code.

Note that in order for icons to line up well both vertically and horizontally, it's better to have them be squares. With this change, some icons now look smaller than they used to (in particular the icons related to hover).

Horizontal


Vertical

alexcjohnson:
Good call on the ➕ / ➖ order. I suppose we could flip *groups* but... then it gets confusing if people want to set their own icons. OK, lets leave it as you have it now.

Agreed on making them all square in vertical mode. In horizontal mode I do think it looks a bit better with the original sizes, if it's not too difficult to maintain both behaviors.

etpinard:
> Good call on the  /  order. I suppose we could flip _groups_ bu

I'd vote for making the snapshot button appear at the bottom of the vertical list and the hover buttons at the top under `orientation: 'v'`.

alexcjohnson:
> I'd vote for making the snapshot button appear at the bottom of the vertical list and the hover buttons at the top under `orientation: 'v'`.

Right, that's what I meant by flipping groups - ie the order of groups gets reversed, but within each group the order is maintained. But @etpinard don't you think this would make adding/removing icons (using `modeBarButtons(ToAdd|ToRemove)?`) confusing?

etpinard:
> don't you think this would make adding/removing icons (using `modeBarButtons(ToAdd|ToRemove)?`) confusing?

I don't think so, if we document it properly. It might a good excuse to :white_check_mark:  https://github.com/plotly/plotly.js/issues/2703

antoinerg:
Reversing the order of groups looks like this:


As long as the ➕ is above ➖, I don't have a preference.

src/plots/layout_attributes.js
```diff
@@ -224,5 +224,30 @@ module.exports = {
             'or a logo image, for example. To omit one of these items on the plot,',
             'make an item with matching `templateitemname` and `visible: false`.'
         ].join(' ')
+    },
+    modeBarStyle: {
```
etpinard: no :camel: in data/layout attributes!

etpinard: ... but wait do we really want to make these layout attributes?

etpinard: Ha I see, @alexcjohnson , @nicolaskruchten and @jackparmer have agreed on making these layout attributes in https://github.com/plotly/plotly.js/issues/3063. Not much I can say I guess.

antoinerg: Maybe you could chime in on how we should name those attributes. I didn't realize there no 🐫 in data/layout attributes! 😳

antoinerg: I vote for @etpinard's suggestion. What do you guys think? @alexcjohnson @nicolaskruchten @jackparmer

antoinerg:
Commit f9e8777 does all the styling via CSS injected into the page. The hover and active styles are now working again so there is no regression. The `style` element is properly removed when the plot is purged.

I think the PR now covers all that was requested in #3063.

test/jasmine/tests/modebar_test.js
```diff
@@ -288,9 +294,15 @@ describe('ModeBar', function() {
...
 
+            var style = document.querySelector('style[id*="modebar"]');
+            expect(style).toBeTruthy();
+
             modeBar.destroy();
 
             expect(modeBarParent.querySelector('.modebar')).toBeNull();
+
+            style = document.querySelector('style[id*="modebar"]');
+            expect(style).toBeNull();
```
antoinerg: This is failing on the CI even though it's working on my Windows 10 laptop and my Linux workstation at home 🤷‍♂️. Do you have any idea why?

etpinard: `npm run test-jasmine -- modebar` works fine but

`npm run test-jasmine -- --nowatch` does not.

Something from another suite (before `modebar_test.js` alphabetically) must be not get cleared properly.

antoinerg: Ok, hopefully I will find the misbehaving tests with this command:
`find test/jasmine/tests/ -type f -printf "%f\n" | xargs -I {} xvfb-run -a npm run test-jasmine -- {} modebar_test.js --nowatch`

antoinerg: Ok I found a few. My intuition tells me the problem is that they solely use `destroyGraphDiv` to clear the stage which is not good enough sometimes. For example, responsive plots attach event listeners on the window, `style` elements are injected into the page on load, and so on. `Plotly.purge` should probably be used instead. I will confirm tomorrow that this is indeed the case.

alexcjohnson: Might be best to just pre-clean for this suite, rather than try and sanitize all the other tests. So something up at the top of the suite like
```js
beforeAll(function() {
    d3.selectAll('style[id*="modebar"]').remove();
});
```

alexcjohnson: Or you could put `d3.selectAll('style[id*="modebar"]').remove();` into `destroyGraphDiv`. That routine is wholly predicated on the idea that we only have one plot at a time in the tests, so there should be no problem deleting style sheets added by a different plot. Possibly a whole `Plotly.purge(gd)` could go into `destroyGraphDiv` but that might have been omitted for performance reasons...

antoinerg: I decided to make my query selector more explicit to target the current figure's modebar. That way, I didn't have to touch the other tests.

I also feel like a `Plotly.purge(gd)` could/should go into `destroyGraphDiv` but that seemed like a pretty big change just to get this PR through.

src/css/_modebar.scss
```diff
@@ -44,19 +39,21 @@
...
+}
 
-    &.modebar-btn--logo {
-        padding: 3px 1px;
+.modebar.vertical {
+    .modebar-group {
+        display: block;
+        float: none;
+        margin-left: 0px;
+        margin-top: 8px;
```
etpinard: peek 2018-10-04 13-41

Looks like the vertical orientation has more top margin than the what the horizontal orientation has to its right.

Is this on purpose? This is just something I noticed, I don't have a strong opinion on what this _should_ look like.

build/ploticon.js
```diff
@@ -120,5 +120,9 @@ module.exports = {
...
+    },
+    'newplotlylogo': {
+        'name': 'newplotlylogo',
+        'svg': '<svg xmlns=\'http://www.w3.org/2000/svg\' viewBox=\'0 0 132 132\'><defs><style>&#xD;\n      .cls-1 {&#xD;\n        fill: #119dff;&#xD;\n      }&#xD;\n&#xD;\n      .cls-2 {&#xD;\n        fill: #25fefd;&#xD;\n      }&#xD;\n&#xD;\n      .cls-3 {&#xD;\n        fill: #fff;&#xD;\n      }&#xD;\n    </style></defs><title>plotly-logomark</title><g id=\'symbol\'><rect class=\'cls-1\' width=\'132\' height=\'132\' rx=\'6\' ry=\'6\'/><circle class=\'cls-2\' cx=\'78\' cy=\'54\' r=\'6\'/><circle class=\'cls-2\' cx=\'102\' cy=\'30\' r=\'6\'/><circle class=\'cls-2\' cx=\'78\' cy=\'30\' r=\'6\'/><circle class=\'cls-2\' cx=\'54\' cy=\'30\' r=\'6\'/><circle class=\'cls-2\' cx=\'30\' cy=\'30\' r=\'6\'/><circle class=\'cls-2\' cx=\'30\' cy=\'54\' r=\'6\'/><path class=\'cls-3\' d=\'M30,72a6,6,0,0,0-6,6v24a6,6,0,0,0,12,0V78A6,6,0,0,0,30,72Z\'/><path class=\'cls-3\' d=\'M78,72a6,6,0,0,0-6,6v24a6,6,0,0,0,12,0V78A6,6,0,0,0,78,72Z\'/><path class=\'cls-3\' d=\'M54,48a6,6,0,0,0-6,6v48a6,6,0,0,0,12,0V54A6,6,0,0,0,54,48Z\'/><path class=\'cls-3\' d=\'M102,48a6,6,0,0,0-6,6v48a6,6,0,0,0,12,0V54A6,6,0,0,0,102,48Z\'/></g></svg>'
```
etpinard: Could we trim those `\n` and spaces?

src/components/modebar/modebar.js
```diff
@@ -45,13 +45,28 @@ var proto = ModeBar.prototype;
...
         this.element.className = 'modebar modebar--hover';
     }
     else this.element.className = 'modebar';
 
+    if(fullLayout.modebar.orientation === 'v') {
+        this.element.className += ' vertical';
+        buttons = buttons.reverse();
+    }
+
+    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId, 'background-color: ' + fullLayout.modebar.bgcolor);
```
etpinard: Do we _have_ to use CSS here? Could we instead add `mouseover` / `mouseout` handlers to the modebar buttons, similar to how we handle updatemenu buttons:

https://github.com/plotly/plotly.js/blob/30ed4a4a59789a486b70dda360d630d290811816/src/components/updatemenus/draw.js#L219-L221

https://github.com/plotly/plotly.js/blob/30ed4a4a59789a486b70dda360d630d290811816/src/components/updatemenus/draw.js#L464-L467

that way we wouldn't have to hijack the `<style>` dynamically.

antoinerg:
Commit b7c2355 updates the CSS to place tooltips on the left of vertical modebars!

src/components/modebar/modebar.js
```diff
@@ -45,13 +45,28 @@ var proto = ModeBar.prototype;
...
+    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn .icon path', 'fill: ' + fullLayout.modebar.color);
+    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn:hover .icon path', 'fill: ' + fullLayout.modebar.activecolor);
+    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn.active .icon path', 'fill: ' + fullLayout.modebar.activecolor);
```
etpinard: These don't appear to update properly on `relayout` e.g. `Plotly.relayout(gd, 'modebar.bgcolor', 'blue')`

etpinard: peek 2018-10-04 13-54

moreover, `modebar.color` doesn't seem to work on some icons.

src/css/_modebar.scss
antoinerg: Nice catch 👀 ! No, it's not on purpose. I'd rather have the logo at the top. Fixing this right now!

src/components/modebar/modebar.js
antoinerg: Let me investigate `relayout`!

As for the screen capture: the gray icons are the ones that are currently active (gray being the default `activecolor`). It does look like a bug the first time you try it but it's not. Maybe I should implement what @nicolaskruchten suggested over here: https://github.com/plotly/plotly.js/issues/3063#issuecomment-426720053. This shouldn't be a problem in Dash however since they usually explicitly style everything.

etpinard: > the gray icons are the ones that are currently active (gray being the default `activecolor`).

Ha I see. My mistake!

antoinerg: Commit 97f25e2 fixes the issue with `relayout`. 

The style rules were just appended to the same `style` element and would have less priority than previous rules. This is fixed by starting with a fresh `style` element each time.

src/components/modebar/modebar.js
etpinard: > Since we already add/remove class name `active` to the icons

Right, getting "active" to work would require some work. The modebar button click handlers would need to update their style attribute instead of appending/removing the `"active"` class name. 

Consider my comment non-blocking.

build/ploticon.js
antoinerg: Fixed by commit dc7877b

src/components/modebar/modebar.js
etpinard: Cool. Let's :lock: this down in a jasmine test though.

antoinerg: In commit f5cc0e9, we test each new layout attributes and make sure they behave properly when using `relayout`!

etpinard: Great tests. Thanks :tada:

alexcjohnson:
@antoinerg it's looking very close! I think the default color logic needs a bit of tweaking though - the inactive color is often too saturated and sometimes too close to the active color, and they're often too prominent in general.

See if you can use `Color.contrast` without `lightAmount` or `darkAmount`, followed by `Color.addOpacity` (just using different opacity for active and inactive), to make colors that closely match the existing colors when the background is white.

That strategy won't be able to exactly match our current grays (`rgba(0,31,95,0.3)` and `rgba(0,22,72,0.5)`), which are tinted a bit blue, unless perhaps you replace `Color.defaultLine` when returned by `Color.contrast` with a different color based off those two...

antoinerg:
I followed your suggestion @alexcjohnson:
```javascript
coerce('modebar.bgcolor', Color.addOpacity(layoutOut.paper_bgcolor, 0.5));
var modebarDefaultColor = Color.contrast(Color.rgb(layoutOut.modebar.bgcolor));
coerce('modebar.color', Color.addOpacity(modebarDefaultColor, 0.3));
coerce('modebar.activecolor', Color.addOpacity(modebarDefaultColor, 0.7));
```
which gives the much improved result seen below:
chrome_2018-10-05_15-05-13

Do you have any other comments/suggestions?

alexcjohnson:
Indeed, much improved! I'm a little surprised that the red background counts as dark according to `Color.contrast`, the inactive buttons are a little hard to see there. But... really, who makes a plot with a bright red background? To my eye all the others look good, so I'd say we're ready to go!
💃

jackparmer:
🎉 🥇 🔍

PR #3070, opened 2018-10-03
Axis autotype update

### LINKED ISSUE #1413: Cant print 'None' string in x axis in Scatter plot if first value
If the first value in x series is a string 'None' the plot is not rendered.

http://codepen.io/ssjunior/pen/qrBMNG

### LINKED ISSUE #2473: Improved detection of NaN values in axis_autotype
Dear Plotly,

Thank you for your open source plotly.js library. We are finding it very useful for our renewable energy and meteorological data. We have found an issue and would like to report it in hopes of finding a resolution.

We have time series data for example 2 months of Clear Sky Insolation on a Horizontal Surface at a given location. We like to plot this data using a line and histrogram chart. When we have missing data, we use 'nan' to fill the missing data values to exclude from the charts. Sometimes (and we don't exactly know the threshold), if we have too many missing data values, the line chart Y axis is not sorted by value and it adds 'nan' as a value and the entire graph skews, as if it includes 'nan' as a value and sorts the y axis based on randomness. We are not sure how to resolve this other than not show a chart if we have too many missing values for a given time series and parameter, although sometimes it appears to work as expected. Here is our application power.larc.nasa.gov/data-access-viewer

I've attached 2 screen shots showing how the line charts excludes 'nan' values and removes them from the line in the first image, and in the second image how it adds 'nan' to the Y axis as if it is a value.

Thank you for your assistance with this isssue.

Feel free to reach me at tyler.d.bristow@nasa.gov

powerchart1
powerchart2

### LINKED ISSUE #3039: Boolean data is autotyped like a number
```js
> Plotly.newPlot(gd,[{y:[true,false,true]}])
> gd._fullLayout.yaxis.type
< "linear"
```
screen shot 2018-09-21 at 4 31 27 pm

It should be treated as categories instead:
```js
Plotly.newPlot(gd,[{y:[true,false,true]}],{yaxis:{type:'category'}})
```
screen shot 2018-09-21 at 4 30 09 pm

alexcjohnson:
Fixes #3039 - treat `boolean` as category data in autotype

Fixes #2473, fixes #1413, by counting only distinct values while determining date and category autotype, as discussed in https://github.com/plotly/plotly.js/issues/2473#issuecomment-372825967. This way we don't need to include any explicit "missing" values; data with up to 2 non-numeric values for every numeric value will still be interpreted as numbers, but even a single non-numeric string (including the previous special values `'None'` and `''`) will be interpreted as a valid category if there are no numbers present.

Note that as part of ^^ I also converted the date determination to counting distinct values. Nobody had complained about this part, but if I didn't do that there could be some strange cases where you have only date strings and numbers but our result is `'category'` 🤔

There's probably a way to improve performance (and make it clearer that we can't generate such strange results) by combining `moreDates` and `category` into one loop... but I don't think that's a major cost anymore at least since we switched to testing max 1000 points.

cc @etpinard @antoinerg @archmoj

### DIFF
--- a/test/jasmine/tests/axes_test.js
+++ b/test/jasmine/tests/axes_test.js
@@ -189,6 +189,22 @@ describe('Test axes', function() {
             fullData = [];
         });
 
+        describe('autotype', function() {
+            function fullTrace(mods) {
+                return Lib.extendDeep(
+                    {type: 'scatter', xaxis: 'x', yaxis: 'y'},
+                    mods
+                );
+            }
+            it('treats booleans as categories', function() {
+                fullData = [fullTrace({x: [0, 1, 2], y: [true, false, true]})];
+                layoutIn = {xaxis: {}, yaxis: {}};
+                supplyLayoutDefaults(layoutIn, layoutOut, fullData);
+                expect(layoutOut.xaxis.type).toBe('linear');
+                expect(layoutOut.yaxis.type).toBe('category');
+            });
+        });
+
         it('should set undefined linewidth/linecolor if linewidth, linecolor or showline is not supplied', function() {
             layoutIn = {
                 xaxis: {},
--- a/test/jasmine/tests/axes_test.js
+++ b/test/jasmine/tests/axes_test.js
@@ -190,18 +190,76 @@ describe('Test axes', function() {
         });
 
         describe('autotype', function() {
-            function fullTrace(mods) {
-                return Lib.extendDeep(
+            function supplyWithTrace(trace) {
+                var fullTrace = Lib.extendDeep(
                     {type: 'scatter', xaxis: 'x', yaxis: 'y'},
-                    mods
+                    trace
                 );
-            }
-            it('treats booleans as categories', function() {
-                fullData = [fullTrace({x: [0, 1, 2], y: [true, false, true]})];
                 layoutIn = {xaxis: {}, yaxis: {}};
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/axes_test.js
```diff
@@ -190,18 +190,76 @@ describe('Test axes', function() {
...
+                checkTypes('category', 'category');
+            });
+
+            it('lets a single number beat up to two distinct categories', function() {
+                supplyWithTrace({
+                    x: ['2.1', 'N/A', '', 'N/A', '', 'N/A', 'N/A', '', '', ''],
+                    y: [0, 'None', true, 'None', 'None', 'None', 'None', 'None']
+                });
+                checkTypes('linear', 'linear');
+            });
```
etpinard: Beautiful! This test case here is enough to convince this PR is _non-breaking_.

etpinard:
> Note: I'll check this myself off my latest splom commits.

The results are in: these commits here don't make much of a difference in axis-autotype perf.

**But** I never realized how much time large splom traces spend in `axis_autotype`. In https://codepen.io/etpinard/pen/wjmqmO, we spend roughly 40ms in `axis_autotype` ! I'll add a comment about this in https://github.com/plotly/plotly.js/issues/2549. No need to handle this now of course.

:dancer:

PR #3072, opened 2018-10-03
Fix parcats + Plotly.react bug

etpinard:
a follow-up on @jonmmease 's https://github.com/plotly/plotly.js/pull/2963, making `Plotly.react` result in a `noop` when reacting from the same `parcats` figure.

I noticed that one of our `@noCI` was failing, at first I thought that simply appending of test mock list would suffice, but no, we also needed to patch `parcats` to make that `@noCI` test pass.

While at it, I also added some tolerance here and there the make the whole suite pass on my setup.

cc @alexcjohnson


src/traces/parcats/calc.js
```diff
@@ -480,19 +480,19 @@ function validateDimensionDisplayInds(trace) {
 function validateCategoryProperties(dim, uniqueInfoDim) {
 
     // Update categoryarray
-    dim.categoryarray = uniqueInfoDim.uniqueValues;
+    dim._categoryarray = uniqueInfoDim.uniqueValues;
```
alexcjohnson: looks reasonable, seems like we always go through this step so simply adding new `_` attributes should be safe (in fact this may fix a bug if there are any `plot` pathways that will redraw a `parcats` without hitting `calc`.

@jonmmease I notice the docstring for this function is copypasta... care to enlighten us as to what's really going on here?

alexcjohnson:
I also notice we sometimes overwrite `displayindex` during calc - presumably that can cause problems too?

https://github.com/plotly/plotly.js/blob/e43d0f85149570e9d8194bc1383ebd40465fdf9f/src/traces/parcats/calc.js#L463-L470

jonmmease:
I’ll take a look this evening...

src/traces/parcats/calc.js
jonmmease: Yeah, sorry, this function has an unhelpful docstring and its name didn't evolve with the change in how category order is specified.  The purpose of this function is to set `categoryarray` and `ticktext` with values that are consistent with the unique values found for this dimension.

A better docstring would be something like
```
/**
 * Update category properties based on the unique values found for this dimension
 * ...
 */
```

jonmmease:
Yeah, `displayindex` is overwritten if the user specified `displayindex` values are not consistent across dimensions. So perhaps we need to introduce an `_displayindex` property here as well.

I'm not very familiar with the `react` pathway yet, but let me know if there's anything here you'd like me to dig into.  Thanks @etpinard!

etpinard:
> Yeah, `displayindex` is overwritten if the user specified `displayindex` values are not consistent across dimensions. So perhaps we need to introduce an `_displayindex` property here as well.

I'll take :eyes:

src/traces/parcats/calc.js
etpinard: Fixed in https://github.com/plotly/plotly.js/pull/3072/commits/3bb0a774623d32bd48c6db854f4e3d8b7ffcf0cd

etpinard:
> So perhaps we need to introduce an `_displayindex` property here as well.

Done and :lock: ed in https://github.com/plotly/plotly.js/pull/3072/commits/157a6e28aa3d79d17320bddef5029e1804306420

src/traces/parcats/calc.js
```diff
@@ -456,16 +456,23 @@ function getUniqueInfo(values, uniqueValues) {
...
+ * Otherwise, replace the display order with the dimension order
  * @param {Object} trace
  */
 function validateDimensionDisplayInds(trace) {
-    var displayInds = trace.dimensions.filter(visible).map(function(dim) {return dim.displayindex;});
-    if(!isRangePermutation(displayInds)) {
-        trace.dimensions.filter(visible).forEach(function(dim, i) {
-            dim.displayindex = i;
-        });
+    var visibleDims = Lib.filterVisible(trace.dimensions);
```
alexcjohnson: ha nice - I thought about mentioning that one before, do you want to use it to replace the other `filter(visible)` instances here? the `!('visible' in dimension)` clause is likely another relic of @jonmmease's education into `supplyDefaults` 😁

etpinard: Ok - done in https://github.com/plotly/plotly.js/pull/3072/commits/24af02a7ded0d15fe750e8876b049c23adb66b60

jonmmease:
Thanks @etpinard!

Issue #2904, opened 2018-08-15
scattergl reversed lines bug


someusersomeuser:
https://codepen.io/anon/pen/EpJObQ
for scattergl 
if autorange: 'reversed'
with mode: 'lines+markers' works fine, but with mode: 'lines' - one incorrect horisontal line (hints display correctly on free space)

etpinard:
Thanks for the report!

archmoj:
As well as { autorange:'reversed' }, I observed that the issue also exists using 
{ ranges:[a, b] } when a > b.

archmoj:
The bug is actually related to the modification of scale parameter within the vertex-shader of regl-line2d library: node_modules/regl-line2d/miter-vert.glsl
For any reversed axis the scale value was replaced by positive epsilon: 1e-6 by the max function. This resulted in display of an almost horizontal/vertical line (depending on which axis was reversed).

The users may also wait for additional tests to be completed and possibly a new release of the mentioned library.

In the meantime using the following function inside the vertex-shader file could help fix the issues e.g. reversed bounds in 2D space.
<pre>
float repaireScale( float a ) {
	float b = a;
	if (abs(b) < MIN_DIFF) {
		if (b < 0.0) b = -MIN_DIFF;
		else b = MIN_DIFF;
	}
	return b;
}

void main() {
	vec2 aCoord = aCoord, bCoord = bCoord, prevCoord = prevCoord, nextCoord = nextCoord;

	vec2 newScale = scale; 
	newScale.x = repaireScale(newScale.x);
	newScale.y = repaireScale(newScale.y);
  
	vec2 scaleRatio = newScale * viewport.zw;
  
  ...

}
</pre>

PR #3078, opened 2018-10-04
issue #2904

archmoj:
Bug fix for 2D line plots.
Negative scales are handled in the webgl vertex shader so that the line plots would be displayed in layouts with reversed ranges.
@etpinard 
@alexcjohnson

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8861,8 +8861,8 @@
       }
     },
     "regl-line2d": {
-      "version": "git://github.com/archmoj/regl-line2d.git#506970ff38c2e55419aaf95f5f816fba4d3d3073",
-      "from": "git://github.com/archmoj/regl-line2d.git#506970ff38c2e55419aaf95f5f816fba4d3d3073",
+      "version": "git://github.com/archmoj/regl-line2d.git#e613c7f11907fea6bc533d970203d3396b0ef651",
+      "from": "git://github.com/archmoj/regl-line2d.git#e613c7f11907fea6bc533d970203d3396b0ef651",
       "requires": {
         "array-bounds": "^1.0.0",
         "array-normalize": "^1.1.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -8861,8 +8861,8 @@
       }
     },
     "regl-line2d": {
-      "version": "git://github.com/archmoj/regl-line2d.git#e613c7f11907fea6bc533d970203d3396b0ef651",
-      "from": "git://github.com/archmoj/regl-line2d.git#e613c7f11907fea6bc533d970203d3396b0ef651",
+      "version": "git://github.com/archmoj/regl-line2d.git#87dbc502d0a22611ed135b6d525e7bc7bad3932f",
+      "from": "git://github.com/archmoj/regl-line2d.git#87dbc502d0a22611ed135b6d525e7bc7bad3932f",
       "requires": {
         "array-bounds": "^1.0.0",
         "array-normalize": "^1.1.3",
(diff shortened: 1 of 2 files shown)


etpinard:
Thanks @archmoj ! 

Pro tip: make the PR title descriptive, so that we have an idea of what the PR is about just by looking at the title. No one remember issue numbers. Thanks again!

etpinard:
@archmoj I'm curious. What made you decide to patch the `regl-line2d`? 

From https://codepen.io/anon/pen/EpJObQ, swapping `mode: 'lines'` for `mode: 'lines+markers'` makes things work leading me to believe that we "could" fix this in plotly.js' [`scattergl/index.js`](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/index.js).

archmoj:
@etpinard 
When using only lines (without markers, etc.) using reversed ranges used to create a horizontal or vertical line rather than actual graph.
Please try this: 
<pre>
Plotly.newPlot(gd, [{ "type": "scattergl", "mode": "lines", "x": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], "y": [0, 1, 0, 7, 7, 0, 2, 4, 3, 0], "line": { "width": 10, "dash": "solid" } }], { xaxis:{ range: [-1, 10] }, yaxis:{ range: [10, -1] }}) ;
</pre>

etpinard:
Yeah I know, but 

```
Plotly.newPlot(gd, [{ "type": "scattergl", "mode": "markers+lines", "x": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], "y": [0, 1, 0, 7, 7, 0, 2, 4, 3, 0], "line": { "width": 10, "dash": "solid" } }], { xaxis:{ range: [-1, 10] }, yaxis:{ range: [10, -1] }}) ;
```

gives the correct result:

image


with the current master `regl-line2d` shader.

archmoj:
I see that the point of a patch in javascript file.
However with in the shader we defined and used 'scale' both as a local variable and uniform!
1. uniform vec2 scale, translate; // global
2. vec2 scale = max(scale, MIN_DIFF); // local
And I think in the example image provided above the corners may not be properly rendered?

archmoj:
To compare with the above image here is an example of applying only `lines` with round corners (using new regl-line2d shader). 
newplot 22 - copy
<pre>Plotly.newPlot(gd, [{ "type": "scattergl", "mode": "lines", "x": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], "y": [0, 1, 0, 7, 7, 0, 2, 4, 3, 0], "line": { "width": 10, "dash": "solid" } }], { xaxis:{ range: [-1, 10] }, yaxis:{ range: [10, -1] }}) ;</pre>

etpinard:
> However with in the shader we defined and used 'scale' both as a local variable and uniform!

> To compare with the above image here is an example of applying only `lines` with round corners (using new regl-line2d shader).

Thanks for the explanation. This makes it very clear. Don't be shy to write down comment on "important" lines on your github PRs to add a little bit of backstory to your work. This makes the review much easier (and faster).

--------

Now, I think the next step would be the make a PR to https://github.com/a-vis/regl-line2d and ping @dy

archmoj:
Not being listed in the contributors of regl I was not able to push my branch for the fix to `a-vis/regl-line2d` (rather than my [fork](https://github.com/archmoj/regl-line2d/tree/rev-bounds2d)) so that it could be on `a-vis`. A similar permission may be required for `regl` npm packages.
@dy 
@alexcjohnson 
@bpostlethwaite

alexcjohnson:
Looks good to me! 💃

etpinard:
Lovely fix :1st_place_medal:

Issue #3077, opened 2018-10-04
Rotate 3D axis labels


alexcjohnson:
Long axis titles on 3D scenes overlap the tick labels and data, because they're simply centered on a point outside the axis and not rotated:
![screen shot 2018-10-04 at 3 54 01 pm](https://user-images.githubusercontent.com/2678795/46499366-dfc1d980-c7ed-11e8-8b32-70cdf4879a24.png)
Our test images don't include a lot of long 3d titles, but you can see this effect a little bit in [gl3d_world-cals](https://github.com/plotly/plotly.js/blob/master/test/image/baselines/gl3d_world-cals.png)

Feels like the easiest solution - which is also consistent with our 2D axis titles - would be to rotate these labels parallel to the axis. But potentially 1-letter titles ("X" etc) should remain unrotated?

The degree of overlap also depends on the viewpoint, so the logic for where to put this relative to the axis line and tick labels probably needs a bit of adjustment:
![3d title overlap](https://user-images.githubusercontent.com/2678795/46499814-1f3cf580-c7ef-11e8-8c55-6b570435fbe3.gif)

etpinard:
We could probably merge https://github.com/plotly/plotly.js/issues/511 with this new ticket.

alexcjohnson:
good call. I'll close #511. Referencing http://community.plot.ly/t/axes-labels-overlapping-on-3d-surface-plot/859

alexcjohnson:
a `titleangle` attribute would let people (manually) fix the overlap for one particular camera angle, but as soon as the user starts rotating the scene, the overlap returns. The two options I see are (1) the title rotates to align with the axis, or (2) the title stays horizontal but changes to right-aligned on the left side and left-aligned on the right side. (1) is my strong preference as it's more compact (ie less likely to overflow the scene), and also helps guide your eye as to the representation of that variable.

MeganBeckett:
Hi there! Is this now possible? ie. to make long axis titles lie parallel to the axes in 3d plots? It seems something has been developed, but I can't see how to implement in plotly (I'm also using plot_ly in R)?

archmoj:
@MeganBeckett 
Yes. The latest versions of plotly R/js/py do support auto rotation by default. 
Also in case you may be interested in line breaks or using text formatting in 3D scenes, here is an example: https://rreusser.github.io/plotly-mock-viewer/#gl3d_formatted-text-on-multiple-lines

Issue #511, opened 2016-05-05
Overlapping 3D axes labels


etpinard:
reposted from http://community.plot.ly/t/axes-labels-overlapping-on-3d-surface-plot/859

> I was wondering if there is a way to offset axis labels for 3D plots so that they don't overlap the ticks? This seems to be because the text is set to centre justification and the centre point is too close to the axis when the label is not parallel with its axis. Maybe there's a way to get the text to always remain parallel with its axis? [This](Example Project6) is the project where this is an issue.

mraxilus:
I too would like this behaviour. It seems rather cluttered to have them crossing over. For me, if the axis label is any longer than about 9 characters they cross in nearly all orientations.

alexcjohnson:
merging into #3077

PR #3084, opened 2018-10-07
Rotate axis labels that are longer than 4 characters in 3D scenes (issue #3077)

archmoj:
This PR is to enable the interactive alignment of axis tiles (labels) that are long i.e. to avoid the overlap of layout descriptions with data representation: issues #3077 and #511 . 
For various reasons this functionality is not applied for labels that are shorter than 4 characters e.g. 'x', 'y', 'z'.  As a result during the image test there were only 2 images that needed to be replaced which now have vertical titles on their z-axis.
To achieve this, changes are made to textVert.glsl shader as well as axes.js of gl-axes3d module. 
Having this new setup, it could also be possible to orient axis ticks as well (once there is a config option). 
It may also be possible to remove horizontal & vertical alignments locks in the shader and to try orienting every title towards its axis.
@etpinard 
@alexcjohnson


package.json
```diff
@@ -78,7 +78,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.0.0",
     "gl-plot2d": "^1.3.1",
-    "gl-plot3d": "^1.5.5",
+    "gl-plot3d": "git://github.com/archmoj/gl-plot3d.git#316309c4ede55f990a61df5ba91f0f52bad07374",
```
archmoj: https://github.com/gl-vis/gl-axes3d/compare/master...archmoj:rotate3d-axis-lables

archmoj: Today I worked on another version of gl-axes3d for better alignments (here: gl-vis/gl-axes3d@master...archmoj:rotate3d-axis-lables). Therefore the package.json files of gl-plot3d & plotly.js need second updates.

package.json
```diff
@@ -78,7 +78,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.0.0",
     "gl-plot2d": "^1.3.1",
-    "gl-plot3d": "^1.5.5",
+    "gl-plot3d": "git://github.com/archmoj/gl-plot3d.git#c173daf6ed8bb74b5db9feedee9f66b4e39a638d",
```
archmoj: https://github.com/gl-vis/gl-axes3d/compare/master...archmoj:rotate3d-axis-lables

etpinard:
Let's try to get this part of `v1.42.0`

etpinard:
Here's a look at  @archmoj's "dynamic" behavior:

peek 2018-10-09 12-24

alexcjohnson:
FYI @archmoj and I agreed that he will extend this so the titles exactly match the axis angles, rather than being restricted to exactly horizontal or vertical.

archmoj:
Alex, what do you think about these?
newplot 40
newplot 39
newplot 41

archmoj:
@alexcjohnson This could be a good version to test & merge. 
@etpinard Could you please help with updating the baseline images? This time the image test on CircleCi stopped after encountering only a few differences. So I was not able to collect the rest of changes. Thank you!

archmoj:
No need for generating baseline images. They are uploaded. I am looking forward for your review. 
@etpinard @alexcjohnson

alexcjohnson:
Whew, late night @archmoj! 😴 

Tick labels on all 3 axes behave the same now, which is great 🎉 and the titles are still fantastic 🏆 
But the condition for flipping the tick labels between rotated left and rotated right seems weird. Particularly if I point one of the axes close to straight in/out of the screen, I can get labels rotated far past vertical. They're still all determining when to flip independently, and again I wouldn't spend time on that issue right now, but in lieu of a solution like that (which, based on some experimentation like the bottom image below I'm not even sure is a good idea anymore) we should simply restrict their individual angles to [-90, 90]. Some examples of what I see now:

All "moneyness" tick labels are upside-down:
screen shot 2018-10-12 at 9 24 47 am

140 and 160 are past vertical but at least consistent with 40-120, yet 180 and 200 flip to right-side-up:
screen shot 2018-10-12 at 9 31 21 am

archmoj:
Thanks for your observations. I agree for ticks we could also the set the default tolerance to zero. 
On another topic, what do you think about the maximum number of characters in titles so that stay horizontal? I was wondering it may be better to use a number less than 4?

alexcjohnson:
>  I was wondering it may be better to use a number less than 4?

I don't have a strong opinion about this. 1-letter titles should definitely NOT be rotated, as then you don't have enough context to figure out how to read it. Long titles must be rotated (that's how we got into this issue after all!) But I could see an argument for anywhere from 2 to 4 letters being the longest unrotated title.

At some point we will make this configurable, for now lets leave it at 4 unless anyone else has strong feelings otherwise.

alexcjohnson:
Much better. I can still find occasional cases where it doesn't behave quite as I would like, but I think we should merge this and move on. Thanks @archmoj! Can you work with @etpinard to get these changes published in the gl-* packages and updated back to npm references? Then 💃 !!!

etpinard:
Hmm @archmoj 

The `gl3d_surface-lighting` mock should not have changed:

peek 2018-10-12 14-33

as it has

image

archmoj:
I think the image above compares the deleted version with the new version. But if we compare the master version with the new image, there is actually no difference. And the angles are explicitly defined for each axes as expected.

antoinerg:
Congratulations!

Issue #2371, opened 2018-02-15
scrollZoom no longer works when page is scrollable


etpinard:
Originally posted https://community.plot.ly/t/scrollzoom-doesnt-work-in-chrome/7356/5

Example: http://marfoldi.web.elte.hu/plotly.html

Possible culprit: https://github.com/plotly/plotly.js/pull/2296

alexcjohnson:
I can't reproduce this on any of my browsers (mac OS 10.13, latest Chrome, FF, Safari) - The OP seems to be on Windows, but I guess you see it on Ubuntu as well @etpinard ?

etpinard:
Yep. I can reproduce on my Ubuntu machine fine. @dfcreative can you double check on Windows?

dfcreative:
@etpinard only the last plot is working, but all three of them should work, that is what I see. Confirm the bug.

ChristofferHS:
In the example, none of the plots work for me, neither in Chrome nor Firefox.

A couple of observations (Windows):

- in the example link in the opening post, if I 1) open the developer tools or some other window that hides part of the right hand side of the plots and then 2) refresh the page so that the plots are drawn less wide and then close window, all the plots are now working for me (both in Chrome and Firefox, but the window that was opened has to be closed first).

- a `div` with styling appears to give a working plot for me: 

```
<body>
<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
<div id="plot_0" style="width:110vw;height:100%"></div>
<div id="plot_1"></div>
<div id="plot_2" style="width:70vw"></div>
</body>
<script>
	Plotly.newPlot('plot_0', [{x: [1, 2, 3, 4, 5], y: [1, 2, 4, 8, 16]}], {title: 'has style, can scroll'}, {scrollZoom: true});
	Plotly.newPlot('plot_1', [{x: [1, 2, 3, 4, 5], y: [16, 8, 4, 2, 1]}], {title: 'has no style, cannot scroll'}, {scrollZoom: true});
	Plotly.newPlot('plot_2', [{x: [1, 2, 3, 4, 5], y: [10, 1, 10, 1, 10]}], {title: 'has style, can scroll'}, {scrollZoom: true});
</script>
```

(for some reason, all plots are scrollable on CodePen with this code)

marfoldi:
Any news on this one?
Unfortunately for me the additional styling on the div is not resolving the issue.

Thanks,
Peter

dy:
[...]

PR #3085, opened 2018-10-09
#2371: Not disabling scrollZoom when page (plot) has scroll bar

marfoldi:
Used to disable scrollzoom when the plot had scrollbars. Ended up with a plot which is not zoomable with mouse scroll on a page which had additional content (had a scrollbar).

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -434,13 +434,6 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
         recomputeAxisLists();
 
-        // if the plot has scrollbars (more than a tiny excess)
-        // disable scrollzoom too.
-        if(pc.scrollHeight - pc.clientHeight > 10 ||
-                pc.scrollWidth - pc.clientWidth > 10) {
-            return;
-        }
-
         clearTimeout(redrawTimer);
 
         var wheelDelta = -e.deltaY;


etpinard:
Thanks for the PR!

I'm a little confused with some of the reports in #2371. Would you sharing an example of the situation you're trying to solve?

marfoldi:
Sure. The scrollToZoom/PinchToZoom is not working (tested on Windows and on Android) when the page is scrollable (the piece of code which I just removed disables it). You can mimic the behavior by opening e.g. the following page: https://codepen.io/plotly/pen/QjKZbg and after the plot is loaded you resize your window. Also when I've opened the issue I've linked an example (http://marfoldi.web.elte.hu/plotly) which contains multiple plots (if you have a kinda regular resolution there should be a scrollbar on page) and none of the plots are zoomable by scrolling (however all of them has the `scrollZoom: true` property).

etpinard:
Thanks again for the PR and let me apologize for the delayed response (our 1.42.0 release cycle was a little hectic).

I think I agree that from a pure plotly.js standpoint your patch does indeed fix a bug. As `scrollZoom` is false by default, we _should_ honor `scrollZoom:true` no matter what. Now unfortunately, the code you're :hocho: is ancient: there might some apps in the wild that depend on this extra "scroll-or-not-to-scroll" logic. 

So, I'll suggest making the `scollZoom` config option accept another value. I'm thinking `scrollZoom: 'always'` would always activate scroll and `scrollZoom: true` would behave the same as currently.

src/plots/cartesian/dragbox.js
```diff
@@ -434,13 +434,6 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
         recomputeAxisLists();
 
-        // if the plot has scrollbars (more than a tiny excess)
-        // disable scrollzoom too.
-        if(pc.scrollHeight - pc.clientHeight > 10 ||
```
etpinard: Hmm. But the comment says if the **plot* had scrollbars, not if the page has some.

Maybe we aren't using the correct DOM properties.

etpinard: So in https://codepen.io/plotly/pen/QjKZbg, 

on full-screen I got:

image

but after resize to half-screen-width, I get e.g.

image

----

So perhaps we should compare `pc.scrollWidth` with `fullLayout.width` and `pc.scrollHeight` with `fullLayout.height` to determine if scroll should be activated.

etpinard:
Now in https://github.com/plotly/plotly.js/pull/3424 - thanks again @marfoldi !

PR #3087, opened 2018-10-09
Scattergl improvements for: fill tozero with bad values fix, fills layering fix, add some line.shape values

etpinard:
This PR builds up on @ErwanMAS's https://github.com/plotly/plotly.js/pull/2979 where 
I added a few commits to make his fix more in line with some `scatter` logic.

In brief, this PR:

- has a fix for scattergl fill `'tozero'` with bad (e.g. undefined) values
- multiple scattergl fill ordering fixes
- add support for `line.shape` values `'hv', 'vh', 'hvh', 'vhv'` (only missing `'spline'`)

Set to be part of `v1.42.0`, cc @plotly/plotly_js


src/traces/scattergl/convert.js
```diff
@@ -393,17 +393,74 @@ function convertLinePositions(gd, trace, positions) {
...
             for(i = 0; i < count - 1; i++) {
-                if(isNaN(positions[i * 2]) || isNaN(positions[i * 2 + 1])) {
+                if(isNaN(positions[i * 2]) || isNaN(positions[i * 2 + 1]) || isNaN(positions[i * 2 + 2]) || isNaN(positions[i * 2 + 3])) {
+                    if(!isNaN(positions[i * 2]) && !isNaN(positions[i * 2 + 1])) {
+                        linePositions.push(positions[i * 2]);
+                        linePositions.push(positions[i * 2 + 1]);
+                    } else {
+                        linePositions.push(NaN);
+                        linePositions.push(NaN);
+                    }
```
alexcjohnson: I'm not quite sure what impact this change had, but I'm curious, why is it needed here and not in `'vh'` below?

alexcjohnson:
@ErwanMAS great work here. I'm really impressed by the new mocks, especially putting SVG and GL side-by-side, that helps a lot since it's otherwise not always obvious without poring over the data whether the filling & ordering is really supposed to be as it appears in the baseline image or not.

@etpinard beautiful simplifications in the last couple of commits. The mocks look fairly complete, so unless my question https://github.com/plotly/plotly.js/pull/3087#discussion_r224286320 points to some otherwise untested case this should be ready to go!

src/traces/scattergl/convert.js
alexcjohnson: BTW it's not new in this PR, but `linePositions.push(NaN); linePosition.push(NaN);` is equivalent to `linePositions.push(NaN, NaN);` right? Could probably shorten this whole section quite a bit that way...

ErwanMAS: @alexcjohnson you are right something was wrong .

@etpinard can you cherry pick this commit 

https://github.com/plotly/plotly.js/commit/cab8a06a9fd819a159f16cb5f796c36b5c9c644d

etpinard: > can you cherry pick this commit

git-applied in d8297a0

>  Could probably shorten this whole section quite a bit that way...

done in d708132

alexcjohnson:
Love it. Thanks for the quick turnaround @ErwanMAS @etpinard !
💃

etpinard:
:100: thank yous to @ErwanMAS , amazing work :1st_place_medal:

PR #3090, opened 2018-10-10
improve autosize in edge cases

antoinerg:
In this PR, I address some shortcomings of https://github.com/plotly/plotly.js/pull/3056 by doing the following things:

- provide fixed non-zero width/height when autosize:true and container's size is zero
- when setting container's widht/height, ignore autosize:true if width/height is explicitly specified by the user.
- 🔒 down in test that explicit width/height win over autosize:true

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -488,6 +488,10 @@ function setPlotContext(gd, config) {
     if(context.setBackground === 'transparent' || typeof context.setBackground !== 'function') {
         context.setBackground = setBackground;
     }
+
+    // Check if gd has a specified widht/height to begin with
+    context._hasZeroHeight = context._hasZeroHeight || gd.clientHeight === 0;
+    context._hasZeroWidth = context._hasZeroWidth || gd.clientWidth === 0;
 }
 
 function plotPolar(gd, data, layout) {
@@ -3247,9 +3251,6 @@ exports.purge = function purge(gd) {
 function makePlotFramework(gd) {
     var gd3 = d3.select(gd);
     var fullLayout = gd._fullLayout;
-
-    // Check if gd has a specified height
-    fullLayout._hasZeroHeight = gd.clientHeight === 0;
 
     // Plot container
     fullLayout._container = gd3.selectAll('.plot-container').data([0]);
--- a/src/plot_api/subroutines.js
+++ b/src/plot_api/subroutines.js
@@ -54,8 +54,8 @@ function lsInner(gd) {
     var i, subplot, plotinfo, xa, ya;
 
     fullLayout._paperdiv.style({
-        width: (fullLayout.autosize) ? '100%' : fullLayout.width + 'px',
-        height: (fullLayout.autosize && !fullLayout._hasZeroHeight) ? '100%' : fullLayout.height + 'px'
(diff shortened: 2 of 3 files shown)


src/plot_api/plot_api.js
```diff
@@ -488,6 +488,10 @@ function setPlotContext(gd, config) {
     if(context.setBackground === 'transparent' || typeof context.setBackground !== 'function') {
         context.setBackground = setBackground;
     }
+
+    // Check if gd has a specified widht/height to begin with
+    context._hasZeroHeight = context._hasZeroHeight || gd.clientHeight === 0;
```
etpinard: Why are we stashing things on `_context` now?

src/plot_api/plot_api.js
```diff
@@ -3248,9 +3252,6 @@ function makePlotFramework(gd) {
     var gd3 = d3.select(gd);
     var fullLayout = gd._fullLayout;
 
-    // Check if gd has a specified height
-    fullLayout._hasZeroHeight = gd.clientHeight === 0;
```
antoinerg: The reason I removed _hasZeroHeight from fullLayout and put it into context is because it would get cleared by a call to `Plotly.newPlot`. However, because this call does not remove the plot from the container, `gd.clientHeight` would be non-zero. This is not because the container has a specified width/height in CSS but simply because it expanded for the plot. As soon as the plot is removed, it goes back to being zero size. Not good!

src/plot_api/subroutines.js
```diff
@@ -54,8 +54,8 @@ function lsInner(gd) {
     var i, subplot, plotinfo, xa, ya;
 
     fullLayout._paperdiv.style({
-        width: (fullLayout.autosize) ? '100%' : fullLayout.width + 'px',
-        height: (fullLayout.autosize && !fullLayout._hasZeroHeight) ? '100%' : fullLayout.height + 'px'
+        width: (fullLayout.autosize && !gd._context._hasZeroWidth && !gd.layout.width) ? '100%' : fullLayout.width + 'px',
+        height: (fullLayout.autosize && !gd._context._hasZeroHeight && !gd.layout.height) ? '100%' : fullLayout.height + 'px'
```
antoinerg: The `&& !gd.layout.width` and `&& !gd.layout.height` conditions are there to ensure that explicit width/height wins over `autosize:true`.

src/plot_api/plot_api.js
etpinard: Ok. That looks good to me. Stashing those things `_context` is "great", but I can't think of a better place, so :ok_hand: 

The problems brought up in https://github.com/plotly/plotly.js/pull/3056#issuecomment-428305909 seemed to now be resolved.

I'll test this thing on stage after lunch.

src/plot_api/subroutines.js
antoinerg: I just realized that one way of making sure we're not breaking anything out there would be to add a `&& gd._context.responsive`. That way, we would absolutely have no regression for people not using `responsive`! After all, setting the container's size to 100% is only useful for responsiveness in modern CSS layouts.

etpinard: > I just realized that one way of making sure we're not breaking anything out there would be to add a `&& gd._context.responsive`

:+1:

PR #3091, opened 2018-10-10
prevent vertical modebar from overflowing

antoinerg:
This PR makes vertical modebars behave like horizontal ones.

Previously, a vertical modebar could overflow the figure's container:
sharex_2018-10-10_14-06-36

With this PR, it doesn't and wrap the items:

This patch relies on `display:flex` which is used on plotly's official webpage and is now [very well supported](https://caniuse.com/#search=flex).


etpinard:
Cool!

Arguably an even better fix would be to collapse the modebar into a 

image

menu on small `<div>`s

etpinard:
:dancer:

Issue #3055, opened 2018-09-27
Template issues


chriddyp:
- [ ] Parcoords line colors not updating with `template` and `react`. The line should be blue but it’s still hotpink: https://codepen.io/chriddyp/pen/EeQoGE
- [x] I couldn’t get the carpet grid colors to update. I might be doing something wrong here with the API though: https://codepen.io/chriddyp/pen/BOYJGy **Fixed in:** #3095
- [x] Similarly I couldn’t get the radar grid lines to updates: https://codepen.io/chriddyp/pen/qMxpJy **Fixed in:** https://github.com/plotly/plotly.js/pull/3092

etpinard:
Potentially related to: https://github.com/plotly/plotly.js/issues/3032

etpinard:
Some info:

Parcoords line colors: 

The template propagate fine during `react`, the problem happens during the parcoords update. We can replicate this bug using `restyle` and `relayout`: https://codepen.io/etpinard/pen/YJVrKm

-----------

Carpet grid colors:

No bug here (I think). You' need to include `startlinecolor` and `endlinecolor` or `color` in your template: e.g. https://codepen.io/etpinard/pen/ZqKXOo?editors=1010

-----------

Polar grid lines:

This line:

https://github.com/plotly/plotly.js/blob/5c4094fc49e910da953b80b06f14e44332afd223/src/plots/polar/layout_defaults.js#L52-L56

is off. In fact, we _do_ need to propagate the template so that the attributes that use [`coerce2`](https://github.com/plotly/plotly.js/blob/5c4094fc49e910da953b80b06f14e44332afd223/src/plots/cartesian/line_grid_defaults.js#L30-L32) can reach into `_template`. This should be an easy fix, PR incoming.

jonmmease:
So I've been able to get the carpet grid colors working in the carpet trace directly, but they don't seem to be coming through from the `data` portion of the template.

https://codepen.io/jonmmease/pen/NOjyya

Does my template `data` specification here look right?

etpinard:
Good :eye: @jonmmease , fixed in https://github.com/plotly/plotly.js/pull/3095

etpinard:
All the "real" template issues listed here have been resolved. 

The parcoords problems are now in -> https://github.com/plotly/plotly.js/issues/3101

Closing.

PR #3092, opened 2018-10-10
Use Template.newContainer in polar axis defaults

etpinard:
checking off the last item in https://github.com/plotly/plotly.js/issues/3055

cc @plotly/plotly_js


src/plots/polar/layout_defaults.js
```diff
@@ -49,11 +51,7 @@ function handleDefaults(contIn, contOut, coerce, opts) {
         }
 
         var axIn = contIn[axName];
-        // Note: does not need template propagation, since coerceAxis is still
-        // based on the subplot-wide coerce function. Though it may be more
-        // efficient to make a new coerce function, then we *would* need to
-        // propagate the template.
```
alexcjohnson: huh, so was this comment wrong all along, or did something change since I wrote this?

alexcjohnson:
Very nice. Regardless of the answer to my question, 💃

src/plots/polar/layout_defaults.js
etpinard: That comment was only _partially_ correct/wrong.

We didn't have to propagate the templates for attributes getting coerced by the `coerce` function in the `polar/layout_defaults.js` scope, **but** we do need to for attributes getting coerced by `coerce2` like here:

https://github.com/plotly/plotly.js/blob/5c4094fc49e910da953b80b06f14e44332afd223/src/plots/cartesian/line_grid_defaults.js#L30-L32

alexcjohnson: > but we do need to for attributes getting coerced by coerce2 like here

Ah... there it is, thanks.

PR #3095, opened 2018-10-11
Use Template.newContainer in carpet axis defaults

etpinard:
which should check off the second item in #3055

cc @plotly/plotly_js


alexcjohnson:
💃 thanks for cleaning up all my loose ends 😏

PR #3096, opened 2018-10-11
implement hoverinfo 'none' and 'skip' in sankey

### LINKED ISSUE #2782: Hoverinfo 'none' and 'skip' do not work with sankey traces

antoinerg:
Closes #2782

### DIFF
--- a/src/plot_api/edit_types.js
+++ b/src/plot_api/edit_types.js
@@ -10,6 +10,7 @@
 
 var Lib = require('../lib');
 var extendFlat = Lib.extendFlat;
+var extendDeepAll = Lib.extendDeepAll;
 var isPlainObject = Lib.isPlainObject;
 
 var traceOpts = {
@@ -115,7 +116,7 @@ function falseObj(keys) {
  * @return {object} a new attributes object with `editType` modified as directed
  */
 function overrideAll(attrs, editTypeOverride, overrideContainers) {
-    var out = extendFlat({}, attrs);
+    var out = extendDeepAll({}, attrs);
     for(var key in out) {
         var attr = out[key];
         if(isPlainObject(attr)) {
--- a/src/plot_api/edit_types.js
+++ b/src/plot_api/edit_types.js
@@ -10,7 +10,6 @@
 
 var Lib = require('../lib');
 var extendFlat = Lib.extendFlat;
-var extendDeepAll = Lib.extendDeepAll;
 var isPlainObject = Lib.isPlainObject;
 
 var traceOpts = {
@@ -116,7 +115,7 @@ function falseObj(keys) {
  * @return {object} a new attributes object with `editType` modified as directed
  */
 function overrideAll(attrs, editTypeOverride, overrideContainers) {
-    var out = extendDeepAll({}, attrs);
+    var out = extendFlat({}, attrs);
     for(var key in out) {
         var attr = out[key];
         if(isPlainObject(attr)) {
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -122,7 +122,14 @@ var attrs = module.exports = overrideAll({
             role: 'style',
             description: 'Sets the thickness (in px) of the `nodes`.'
         },
-        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {valtype: 'enumerated', values: ['all', 'none', 'skip'], flags: [], arrayOk: false }),
+        hoverinfo: {
+            valtype: 'enumerated',
+            values: ['all', 'none', 'skip'],
+            dflt: 'all',
+            role: 'info',
+            editType: 'none',
+            description: plotAttrs.hoverinfo.description
+        },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The nodes of the Sankey plot.'
     },
@@ -182,7 +189,14 @@ var attrs = module.exports = overrideAll({
             role: 'info',
             description: 'A numeric value representing the flow volume value.'
         },
-        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {valtype: 'enumerated', values: ['all', 'none', 'skip'], flags: [], arrayOk: false }),
+        hoverinfo: {
+            valtype: 'enumerated',
+            values: ['all', 'none', 'skip'],
+            dflt: 'all',
+            role: 'info',
+            editType: 'none',
+            description: plotAttrs.hoverinfo.description
+        },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
     }
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -9,7 +9,6 @@
 'use strict';
 
 var fontAttrs = require('../../plots/font_attributes');
-var plotAttrs = require('../../plots/attributes');
 var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
@@ -127,7 +126,11 @@ var attrs = module.exports = overrideAll({
             dflt: 'all',
             role: 'info',
             editType: 'none',
-            description: plotAttrs.hoverinfo.description
+            description: [
+                'Determines which trace information appear when hovering nodes.',
+                'If `none` or `skip` are set, no information is displayed upon hovering.',
+                'But, if `none` is set, click and hover events are still fired.'
+            ].join(' ')
         },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The nodes of the Sankey plot.'
@@ -194,7 +197,11 @@ var attrs = module.exports = overrideAll({
             dflt: 'all',
             role: 'info',
             editType: 'none',
-            description: plotAttrs.hoverinfo.description
+            description: [
+                'Determines which trace information appear when hovering links.',
+                'If `none` or `skip` are set, no information is displayed upon hovering.',
+                'But, if `none` is set, click and hover events are still fired.'
+            ].join(' ')
         },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -185,5 +185,7 @@ module.exports = overrideAll({
         hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
-    }
+    },
+    hoverinfo: undefined,
+    hoverlabel: undefined
 }, 'calc', 'nested');
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -122,7 +122,7 @@ var attrs = module.exports = overrideAll({
             role: 'style',
             description: 'Sets the thickness (in px) of the `nodes`.'
         },
-        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
+        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {valtype: 'enumerated', values: ['all', 'none', 'skip'], flags: [], arrayOk: false }),
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The nodes of the Sankey plot.'
     },
@@ -182,10 +182,10 @@ var attrs = module.exports = overrideAll({
             role: 'info',
             description: 'A numeric value representing the flow volume value.'
         },
-        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
+        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {valtype: 'enumerated', values: ['all', 'none', 'skip'], flags: [], arrayOk: false }),
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
-    },
+    }
 }, 'calc', 'nested');
 // hide unsupported top-level properties from plot-schema
 attrs.hoverinfo = undefined;
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -14,6 +14,7 @@ var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
 
+var extendFlat = require('../../lib/extend').extendFlat;
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
 
 module.exports = overrideAll({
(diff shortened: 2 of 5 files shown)


antoinerg:
Note that right now with `hoverinfo: 'skip'` and `hoverinfo: 'none'`, hovering on a node still highlights its links:


I figured that if a user doesn't want any highlighting, he/she can use `hovermode: false`. Is that reasonable?

alexcjohnson:
alternate proposal, before we create some working `hoverinfo` behavior in `sankey`: https://github.com/plotly/plotly.js/issues/3097#issuecomment-429077334

antoinerg:
Ok so commit 880973a implements the changes discussed in issue #3097 ie. adding attributes `(node|link).hover(info|label)`. 

I removed all the existing `hoverinfo` flags except for 'all', 'none' and 'skip' since they're the only ones implemented right now.

src/plot_api/edit_types.js
```diff
@@ -115,7 +116,7 @@ function falseObj(keys) {
  * @return {object} a new attributes object with `editType` modified as directed
  */
 function overrideAll(attrs, editTypeOverride, overrideContainers) {
-    var out = extendFlat({}, attrs);
+    var out = extendDeepAll({}, attrs);
```
antoinerg: Using `extendDeepAll` instead of `extendFlat` because the latter throws away keys with undefined values.

src/traces/sankey/defaults.js
```diff
@@ -24,6 +25,11 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     coerce('node.thickness');
     coerce('node.line.color');
     coerce('node.line.width');
+    coerce('node.hoverinfo');
+    Registry.getComponentMethod(
```
antoinerg: I could have used a one-liner and call `var Fx = require('../../components/fx'); Fx.supplyDefaults (...)`. I used the Registry but don't know which one is better 🤷‍♂️.

src/traces/sankey/attributes.js
```diff
@@ -127,6 +122,8 @@ module.exports = overrideAll({
             role: 'style',
             description: 'Sets the thickness (in px) of the `nodes`.'
         },
+        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
```
alexcjohnson: Hmm, if these have no flags they should probably not be `flaglist` attributes. I realize that at some point we will likely implement flags but in the meantime they should probably be `enumerated`. Also `plotAttrs.hoverinfo` is `arrayOk`, does that work here?

src/traces/sankey/attributes.js
```diff
@@ -185,6 +182,11 @@ module.exports = overrideAll({
             role: 'info',
             description: 'A numeric value representing the flow volume value.'
         },
+        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
+        hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
-    }
+    },
```
alexcjohnson: trailing comma

src/traces/sankey/defaults.js
alexcjohnson: `Registry` is useful when a component may or may not be in the bundle, or to break circular references. I can't really see us making `Fx` optional any time soon, so I'd probably just require the defaults only:
`var handleFxDefaults = require('../../components/fx/defaults');`

That said, it would be even better to dive even one level deeper and call `handleHoverLabelDefaults` directly, using new `coerce` functions based on the one used here, like how `polar` makes `coerceAxis`:

https://github.com/plotly/plotly.js/blob/c922c61234977756d04598a29d448dd7efb74c63/src/plots/polar/layout_defaults.js#L42-L44

Two reasons for this: it'll work with templates (without having to propagate the template explicitly into `node` and `link`, which doesn't look like it's been done here, see eg [Etienne's recent carpet fix](https://github.com/plotly/plotly.js/pull/3095/files#diff-55e73e1ce7c046db1f44dbeb16ce40caR38)), and it's actually using the attributes defined in this trace in case we ever need them to diverge from the ones in `fxAttrs`

test/jasmine/tests/sankey_test.js
```diff
@@ -433,10 +436,14 @@ describe('sankey tests', function() {
...
-                    'hoverlabel.font.size': 20,
-                    'hoverlabel.font.color': 'black'
+                    'node.hoverlabel.bgcolor': 'red',
+                    'node.hoverlabel.bordercolor': 'blue',
+                    'node.hoverlabel.font.size': 20,
+                    'node.hoverlabel.font.color': 'black',
+                    'link.hoverlabel.bgcolor': 'red',
+                    'link.hoverlabel.bordercolor': 'blue',
+                    'link.hoverlabel.font.size': 20,
+                    'link.hoverlabel.font.color': 'black'
```
alexcjohnson: Lets use different values for node & link, just for good measure.

test/jasmine/tests/sankey_test.js
```diff
@@ -610,27 +666,36 @@ describe('sankey tests', function() {
...
+            });
+        }
+
+        it('should not output hover/unhover event data when hovermoder is false', function(done) {
+            var fig = Lib.extendDeep({}, mock);
+
+            Plotly.plot(gd, fig)
+            .then(function() { return Plotly.relayout(gd, 'hovermode', false); })
+            .then(function() { return assertNoHoverEvents('node');})
+            .then(function() { return assertNoHoverEvents('link');})
```
alexcjohnson: Nonblocking, but for extra style points you can move the function wrapper into `assertNoHoverEvents` so this becomes just: `.then(assertNoHoverEvents('link'))`

src/traces/sankey/attributes.js
alexcjohnson: to be clear: if `arrayOk` is not supported, you don't need to add it now but we should take it out of the attribute spec.

src/traces/sankey/attributes.js
antoinerg: Fixed in 706e7ae

test/jasmine/tests/sankey_test.js
antoinerg: Good idea! Fixed in 706e7ae

src/traces/sankey/attributes.js
antoinerg: `arrayOk` is not supported right now. I set it to `false` for the time being.

test/jasmine/tests/sankey_test.js
alexcjohnson: Thanks. Lets let the reviewer mark a conversation resolved though, even if it's a simple one like this. Makes it a lot easier to tell what I need to look at to ensure all my comments have been addressed.

src/traces/sankey/attributes.js
```diff
@@ -122,7 +121,14 @@ var attrs = module.exports = overrideAll({
             role: 'style',
             description: 'Sets the thickness (in px) of the `nodes`.'
         },
-        hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {flags: []}),
+        hoverinfo: {
+            valtype: 'enumerated',
+            values: ['all', 'none', 'skip'],
+            dflt: 'all',
+            role: 'info',
+            editType: 'none',
```
alexcjohnson: `overrideAll` is just going to wipe this out... if you really want to set `editType` per attribute, just stop using `overrideAll`, and just set it explicitly in each attribute.

test/jasmine/tests/sankey_test.js
antoinerg: Makes total sense! Sorry about that

src/traces/sankey/attributes.js
antoinerg: Do we even need `overrideAll` anymore?

src/traces/sankey/attributes.js
```diff
@@ -122,7 +121,14 @@ var attrs = module.exports = overrideAll({
...
+            description: plotAttrs.hoverinfo.description
```
alexcjohnson: I suspect this isn't going to work with https://github.com/plotly/plotly.js/blob/master/tasks/compress_attributes.js (which removes `description`, and other fields that aren't needed at run time, from attributes when enabled). Anyway you might want to modify the description to mention that this only applies to nodes (and below for links)

src/traces/sankey/attributes.js
alexcjohnson: probably not, just make sure that if you take it out, `restyle` works correctly with attributes you set to an `editType` less than `calc`.

src/plot_api/edit_types.js
antoinerg: Outdated, marking as resolved.

src/traces/sankey/attributes.js
antoinerg: fyi, the description did appear in `plotly-schema.json` after running `npm run build`.

Anyway, I modified the description as you suggested in 7a32036.

src/traces/sankey/defaults.js
antoinerg: I am not sure I'm properly addressing this comment. Can you check commit fe7ab49 and tell me whether this is what you had in mind? Thanks @alexcjohnson

antoinerg:
Thank you for your help @alexcjohnson :)

I think that commit 644c4e3 makes attribute coercion more efficient as you suggested doing!

src/traces/sankey/defaults.js
```diff
@@ -13,34 +13,50 @@ var attributes = require('./attributes');
...
+    coerceLink('line.color');
+    coerceLink('line.width');
+    coerceLink('hoverinfo');
+    handleHoverLabelDefaults(linkIn, linkOut, coerceLink, layout.hoverlabel);
 
-    coerce('link.color', traceOut.link.value.map(function() {
+    coerceLink('color', linkOut.value.map(function() {
         return tinycolor(layout.paper_bgcolor).getLuminance() < 0.333 ?
             'rgba(255, 255, 255, 0.6)' :
             'rgba(0, 0, 0, 0.2)';
```
alexcjohnson: not new here, but this bothers me 🐎 
`tinycolor` is kind of slow, and this result is a constant. Can you pull it out of the loop?

alexcjohnson:
@antoinerg this looks great now! https://github.com/plotly/plotly.js/pull/3096#discussion_r227459050 would be lovely but nonblocking - otherwise 💃

src/traces/sankey/defaults.js
antoinerg: Yes of course!

alexcjohnson: https://github.com/plotly/plotly.js/pull/3096/commits/dc8ba12f16245f89814ba4e8f733f733337ee12a -> thanks!

antoinerg:
Ok awesome! I'll just wait for the test to complete and merge it :)

src/traces/sankey/defaults.js
```diff
@@ -13,37 +13,55 @@ var attributes = require('./attributes');
...
+    coerceLink('line.color');
+    coerceLink('line.width');
+    coerceLink('hoverinfo');
+    handleHoverLabelDefaults(linkIn, linkOut, coerceLink, layout.hoverlabel);
+
+    var defaultLinkColor = tinycolor(layout.paper_bgcolor).getLuminance() < 0.333 ?
+                'rgba(255, 255, 255, 0.6)' :
+                'rgba(0, 0, 0, 0.2)';
+
+    coerceLink('color', linkOut.value.map(function() {
```
etpinard: For the next time you're working in these files @antoinerg, here's `Lib.repeat`:

https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/lib/index.js#L167-L181

src/traces/sankey/attributes.js
```diff
@@ -185,6 +190,21 @@ module.exports = overrideAll({
...
+                'If `none` or `skip` are set, no information is displayed upon hovering.',
+                'But, if `none` is set, click and hover events are still fired.'
+            ].join(' ')
+        },
+        hoverlabel: fxAttrs.hoverlabel, // needs editType override,
         description: 'The links of the Sankey plot.'
     }
 }, 'calc', 'nested');
+// hide unsupported top-level properties from plot-schema
+attrs.hoverinfo = undefined;
```
etpinard: Oh so, trace-level `hoverinfo` is no longer available for sankey traces? Luckily it never worked so this isn't breaking change. 

Does this PR close https://github.com/plotly/plotly.js/issues/3097 ?

src/traces/sankey/attributes.js
```diff
@@ -185,6 +190,21 @@ module.exports = overrideAll({
...
+attrs.hoverlabel = undefined;
```
etpinard: But wait, trace `hoverlabel` was working before this. Hmm, I'd call this a breaking change.

etpinard: ... I think we should either make node/link `hoverlabel` inherit from trace `hoverlabel` or add some [`cleanData`](https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/plot_api/helpers.js#L206) logic.

src/traces/sankey/attributes.js
antoinerg: It wouldn't close it. 

Issue #3097 is about supporting `hoverinfo` options other than the basic `skip` and `none`. Although the current doc say that we support many:
> Any combination of "label", "text", "value", "percent", "name" joined with a "+" OR "all" or "none" or "skip"

none actually worked except `all`. This PR is about slely dealing witk `skip` and `none` which are really important (ie. whether to hide the hoverlabels and whether to emit hover events at the same time).

src/traces/sankey/defaults.js
antoinerg: Good to know about the existence `Lib.repeat`! I just mindlessly reused the `.map` iterator that was already there! I should know better, `for` loops are much faster than `map` :racehorse: !

src/traces/sankey/attributes.js
antoinerg: @etpinard It's great that you noticed that change and I agree this should be handled better! 

FYI, having a `hover(label|info)` in each `node` and `link` was the fruit of discussions with @alexcjohnson. In the future, we could add a `hovertemplate` on each to provide more control. That's one way of dealing with point 2 of this comment https://github.com/plotly/plotly.js/pull/3126#issue-224077784 which we can discuss about later on :)

alexcjohnson: Ah indeed - good catch @etpinard - I'd go for:

> make node/link `hoverlabel` inherit from trace `hoverlabel`

That would be more friendly for the (probably most common) case that users want node and link hover labels styled the same.

etpinard: ... and maybe we should have a trace `hoverinfo` that acts similarly.

alexcjohnson: `hoverinfo` is trickier - could be nice to inherit `skip`/`none`, but when we get around to supporting a real flaglist, links and nodes will have completely different options.

src/traces/sankey/attributes.js
etpinard: Good to know. Thanks!

src/traces/sankey/attributes.js
etpinard: e.g. `trace.hoverinfo: 'none'` would be equivalent to `node.hoverinfo: 'none'` + `link.hoverinfo: 'none'`.

Not all the flags to be implemented in https://github.com/plotly/plotly.js/issues/3097 should be available from the trace hoverinfo, but I think setting trace-wide `'none'` and `'skip'` could be useful.

PR #3109, opened 2018-10-15
add 'width' to box and violin trace

Kully:
re: https://github.com/plotly/documentation/pull/1090
inspiration for joyplots found in community discussion here: https://community.plot.ly/t/ridgeline-joy-plots-in-dash-automatic/12602

First PR for js, whooho 🎉 

---

This is my first shot at adding a `width` parameter for violin and box (I am calling the param `vwidth` for now). So far it works okay. The only thing I am noticing is that

1. all traces in the `data` object have to be set to the same value in order for there to be any update to the plot

Running this script

```javascript
function violin(y0, fillcolor, vwidth) {
    var v = {
      type: 'violin',
      vwidth: vwidth,
      x: [0, 5, 7, 8],
      points: 'none',
      side: 'positive',
      box: {
        visible: false
      },
      boxpoints: false,
      line: {
        color: 'black'
      },
      fillcolor: fillcolor,
      opacity: 0.6,
      meanline: {
        visible: false
      },
      y0: y0
    };
    return v
};

var data = [
  violin(0.0, '#8dd3c7', 0.4),
  violin(0.1, '#d3c78d', 0.4),
  violin(0.2, '#c78dd3', 0.4),
];

var layout = {
  title: "Horizontal JoyPlot",
  xaxis: {
    zeroline: false
  },
  violingap: 0
};

```

I get:

newplot

and tweaking

```javascript
var data = [
  violin(0.0, '#8dd3c7', 1),
  violin(0.1, '#d3c78d', 1),
  violin(0.2, '#c78dd3', 1),
];
```

to

```javascript
var data = [
  violin(0.0, '#8dd3c7', 1),
  violin(0.1, '#d3c78d', 0.4),
  violin(0.2, '#c78dd3', 1),
];
```

I get the following:

newplot 1

And changing the snippet to

```javascript
var data = [
  violin(0.0, '#8dd3c7', 0.4),
  violin(0.1, '#d3c78d', 0.4),
  violin(0.2, '#c78dd3', 0.4),
];
```

where all the `vwidths` are 0.4, we get what we should be getting.

newplot 2

---

I am contemplating switching `vwidth` to a layout attribute in the same category as `violingap`. I would appreciate some input before continuing.

cc @etpinard


alexcjohnson:
Welcome to plotly.js @Kully 🎉  Thanks for the PR!

Lets call this parameter just `width` like we do in `bar` traces, and keep it in the trace, not in `layout`. And we might as well allow it for both `box` and `violin` traces.

> all traces in the data object have to be set to the same value in order for there to be any update to the plot

I suspect in fact all that matters as you have it now is the *last* trace (of the right type on the same subplot). That's because you're taking `trace.vwidth`, where `trace` is just the last one in the loop above, and applying it to the whole `boxList` (or `violinList`) of traces. What you should be doing instead is setting this per-trace down `here`[https://github.com/plotly/plotly.js/blob/f76ca97902a85e1e27c1ae1d4012676766cfc378/src/traces/box/cross_trace_calc.js#L109) by checking if `calcTrace[0].trace` exists and using it instead of `dPos`. (BTW factor of 2? `dPos` is the *half* width, and in your example there's only one half of the violin, but in general I think we want this to be the full width even if you're only showing half; and definitely full width for `box`)

src/plots/plots.js
```diff
@@ -2612,6 +2612,9 @@ function doCrossTraceCalc(gd) {
                     fullLayout[sp];
 
                 for(j = 0; j < methods.length; j++) {
+                    console.log(methods);
+                    console.log(methods[j])
+                    console.log(methods[j].type)
```
alexcjohnson: Please don't commit `console.log` 🐄

src/traces/box/cross_trace_calc.js
```diff
@@ -77,6 +77,11 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
     var boxdv = Lib.distinctVals(pointList);
     var dPos = boxdv.minDiff / 2;
 
+    // override dPos if violin width given
+    if (vwidth != false) {
```
alexcjohnson: 🐄 I mentioned just using `if(width)` already, and anyway this block will move... but never use `==` or `!=`. Always use `===` or `!==`.

src/traces/box/attributes.js
```diff
@@ -255,5 +255,16 @@ module.exports = {
             'Do the hover effects highlight individual boxes ',
             'or sample points or both?'
         ].join(' ')
-    }
+    },
+    vwidth: {
+        valType: 'number',
+        min: 0,
+        role: 'info',
+        dflt: false,
```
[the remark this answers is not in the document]
alexcjohnson: (also this is in `box/attributes` but the `description` below says violins...)

src/traces/box/cross_trace_calc.js
Kully: Is this just a js convention? Do the alternatives work the same way?

jonmmease: JavaScript has permission to do evil casting with `==` but not with `===` 🙂 

screen shot 2018-10-16 at 6 34 57 pm

alexcjohnson: > The advice given to JavaScript beginners is to completely forget about == and to always use ===. It turns out that that rule is universally true.

http://2ality.com/2011/12/strict-equality-exemptions.html

src/traces/violin/defaults.js
```diff
@@ -29,6 +29,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     coerce('scalegroup', traceOut.name);
     coerce('scalemode');
     coerce('side');
+    coerce('vwidth');
```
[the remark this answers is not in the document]
Kully: `scalegroup` and `scalecount` seem to not do anything when there is `width`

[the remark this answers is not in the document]
alexcjohnson: Exactly, that's why we don't want to `coerce` them - every attribute that makes it to `traceOut` should *do* something.

alexcjohnson:
@Kully re: tests - I think all we need for this feature is one or two image tests - covering both box and violin, with at least two different explicit widths and one automatic width on the same subplot.

src/traces/box/cross_trace_calc.js
```diff
@@ -100,8 +100,15 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
-        // set the width of all boxes
-        calcTrace[0].t.dPos = dPos;
+        // set the width of all boxes and
+        // override this width with
+        // trace.width if it exists
+        if(calcTrace[0].trace && calcTrace[0].trace.vwidth) {
+            calcTrace[0].t.dPos = calcTrace[0].trace.vwidth; 
+        } else {
+            calcTrace[0].t.dPos = dPos;   
+        }
```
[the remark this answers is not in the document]
Kully: really nifty line! 👏

src/traces/box/cross_trace_calc.js
```diff
@@ -90,23 +90,35 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
+    }
+
     // autoscale the x axis - including space for points if they're off the side
     // TODO: this will overdo it if the outermost boxes don't have
     // their points as far out as the other boxes
     var extremes = Axes.findExtremes(posAxis, boxdv.vals, {
-        vpadminus: dPos + pad[0] * padfactor,
-        vpadplus: dPos + pad[1] * padfactor
+        vpadminus: max_half_width + pad[0] * padfactor,
+        vpadplus: max_half_width + pad[1] * padfactor
```
alexcjohnson: Ah good catch - unfortunately the `TODO` concern above this gets more acute with custom widths. If you show two traces, one off to the left with a very narrow width and a wide one to the right, the one on the left will be padded as much as the one on the right, giving an inappropriately large autorange.

It *looks* to me as though we could pretty easily fix this by calculating `extremes` per trace. That would go below in the trace loop and might be as simple as:
```js
var thisDPos = calcTrace[0].t.dPos = (calcTrace[0].trace.width / 2) || dPos;
var positions = calcTrace.map(function(di) { return di.pos });
var extremes = Axes.findExtremes(posAxis, positions, {
    vpadminus: thisDPos + pad[0] * padfactor,
    vpadminus: thisDPos + pad[1] * padfactor
});
calcTrace[0].trace._extremes[posAxis._id] = extremes;
```
That wouldn't address the original `TODO` - which I guess is about a per-trace `pad` / `padfactor` depending on `trace.pointpos`, I'd have to look to see if there's an easy solution to that, there may be... but if you don't feel like digging into that feel free to just move the `TODO` along with this change.

src/traces/violin/defaults.js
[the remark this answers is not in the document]
alexcjohnson: Oh sorry, I marked this resolved when I saw that you added the conditional `coerce` block... but I guess you removed it again in a later commit. We do want the conditional, per my previous comment.

src/traces/box/cross_trace_calc.js
```diff
@@ -90,18 +90,33 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
     var groupgap = fullLayout[traceType + 'groupgap'];
     var padfactor = (1 - gap) * (1 - groupgap) * dPos / fullLayout[numKey];
 
+    // Find maximum trace width
+    // we baseline this at dPos
+    var max_half_width = dPos;
```
etpinard: Camel case please :camel:

src/traces/box/attributes.js
```diff
@@ -255,5 +268,6 @@ module.exports = {
             'Do the hover effects highlight individual boxes ',
             'or sample points or both?'
         ].join(' ')
-    }
+    },
```
etpinard: :zap: that `,` 

Not sure why our linting didn't pick this up.

src/traces/box/cross_trace_calc.js
```diff
@@ -90,18 +90,33 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
+    for(i = 0; i < boxList.length; i++) {
+        calcTrace = calcdata[boxList[i]];
+
+        if(calcTrace[0].trace.width) {
+            if(calcTrace[0].trace.width / 2 > max_half_width) {
```
etpinard: Hmm. What if `trace.width ` is smaller than the minimum-diff-between-distinct-values? Does this still work?

src/traces/box/attributes.js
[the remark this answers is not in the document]
alexcjohnson: I was thinking of [`nticks`](https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/plots/cartesian/layout_attributes.js#L224), [`nbins(x|y)`](https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/traces/histogram/attributes.js#L132), also looks like we do that for [`maxdisplayed`](https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/traces/scatter/attributes.js#L360).

`null` is certainly more meaningful on its own than `0`, and it's necessary if `0` is a valid value distinct from "auto" - like [`bar.offset`](https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/traces/bar/attributes.js#L142) where `0` means "center the bar at the data value" and "auto" means "if the bars are grouped, offset them all so they end up side-by-side".

It's just a little funny though with `valType: 'number'` - it means `null` looks invalid, so if you try and set it we'll determine it to be invalid and fall back on the default, which I guess is fine as long as `null` *is* the default, but it would still fail `Plotly.validate`. Also you can't set it at all via `restyle/relayout`, because `null` there means "unset" - again, not immediately a problem as long as this is the default. BUT... what if you have a template that overrides the default? Seems to me then you would be unable to revert to auto with any setting at all of the user attribute, right?

`valType: 'angle'` adds a special value `'auto'`, which might be the best of both worlds? Its meaning is clear (and distinct from `0`), and it's not intercepted by plotlyjs like `null` is. Perhaps we should allow `valType: 'number'` to specify `allowAuto: 'true'`, or `extras: ['auto']`?

src/traces/box/attributes.js
Kully: I added this as I placed the width property at the end of this list the first time around. I then moved to the middle to a more appropriate spot but forgot to ⚡️  the `,`

src/traces/box/cross_trace_calc.js
Kully: It actually does work. See [this line](https://github.com/plotly/plotly.js/blob/eb5e9da1d476351244faff9b69cc9c53d53b9238/src/traces/box/cross_trace_calc.js#L118) where the width is set.

These lines are affecting the axis autoscale of the plotted points. But if a violin has trace.width = 0.00001 (much less than the min-diff as you mention) then that trace will shrink to the whatever the width is set to but the "camera" will not zoom in anymore. The view of all the traces is always calibrated to the violin/box with the maximum width set, which doesn't loose any desired layout that you could want with violins or boxes, if you decide to use width at all.

alexcjohnson:
> anything else that needs updating/adding

https://github.com/plotly/plotly.js/pull/3109#discussion_r226722187 is still open, right? Only coerce `scalegroup` and `scalemode` if there's no `width`?

alexcjohnson:
Also https://github.com/plotly/plotly.js/pull/3109#discussion_r226720869 was never addressed

src/traces/box/cross_trace_calc.js
Kully: For

```javascript
  violin1 => x0=0.0, width=0.1
  violin2 => x0=0.1, width=1.0
```
with changes

and PRE your change above:

before change

Kully: so not that much improvement. We need a better algo for this.

alexcjohnson: Ah, right - because we don't take `trace.side` into account - if it's `'positive'` we should set `vpadminus: 0`, if `'negative'` we should set `vpadplus: 0`.

Kully: Muuuch better:

newplot

src/traces/box/attributes.js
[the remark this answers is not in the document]
etpinard: > BUT... what if you have a template that overrides the default? Seems to me then you would be unable to revert to auto with any setting at all of the user attribute, right?

Interesting point. This is somewhat related to https://github.com/plotly/plotly.js/issues/2834, I'll comment over there.

etpinard:
@Kully do you think you'll have time to fix the failing tests today? If not, this thing we'll have to wait for `v1.43.0`.

Kully:
> @Kully do you think you'll have time to fix the failing tests today?

I won't be able to finish it today.

etpinard:
> I'm not very happy with the fact that the violin has gotten smaller by default,

Yes, that's a problem. The existing baselines should not changed in this PR.

Kully:
@alexcjohnson @etpinard all done and all passing!

I fixed the autorange issue that was breaking the image tests and tweaked where we expect the hoveron info for kde is supposed to be for a couple of the jasmine violin tests - this is expected as I changed the vpad logic in this PR.

Please feel free to leave comments. It looks finished to me.

src/traces/box/cross_trace_calc.js
```diff
@@ -90,19 +95,46 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
+    var maxHalfWidth = dPos;
+    for(i = 0; i < boxList.length; i++) {
+        calcTrace = calcdata[boxList[i]];
+
+        if(calcTrace[0].trace.width) {
+            if(calcTrace[0].trace.width / 2 > maxHalfWidth) {
+                maxHalfWidth = calcTrace[0].trace.width / 2;
+            }
+        }
+    }
```
alexcjohnson: Is this loop needed anymore?

src/traces/box/cross_trace_calc.js
```diff
@@ -90,19 +95,46 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
+            vpadminus = 0;
+            vpadplus = thisDPos + pad[1] * padfactor;
+        } else if(calcTrace[0].trace.side === 'negative') {
+            vpadminus = thisDPos + pad[0] * padfactor;
+            vpadplus = 0;
+        } else {
+            // if side === 'both'
+            vpadminus = thisDPos + pad[0] * padfactor;
+            vpadplus = thisDPos + pad[1] * padfactor;
+        }
```
alexcjohnson: 🌴 
```js
var side = calcTrace[0].trace.side;
var vpadminus = (side === 'positive') ? 0 : (thisDPos + pad[0] * padfactor);
var vpadplus = (side === 'negative') ? 0 : (thisDPos + pad[1] * padfactor);
```

alexcjohnson: I'm a bit worried actually that this will exacerbate the issue in the TODO above. For example what if you have a single-sided violin but the points get drawn on the other side? It looks to me like this algorithm will cut them off entirely. @etpinard thoughts?

test/jasmine/tests/violin_test.js
```diff
@@ -523,6 +523,7 @@ describe('Test violin hover:', function() {
             Plotly.plot(gd, fig).then(function() {
                 mouseEvent('mousemove', 250, 250);
                 assertViolinHoverLine([299.35, 250, 200.65, 250]);
+                // assertViolinHoverLine([178.67823028564453, 250, 80, 250]);
```
alexcjohnson: 🔪

src/traces/box/cross_trace_calc.js
Kully: > I'm a bit worried actually that this will exacerbate the issue in the TODO above

I noticed another bug - if we choose to show the box, it gets cut in half if we choose `positive` or `negative` for `side`
half violin

DATA:
```javascript
var data = [{
    type: 'violin',
    x: [0, 5, 7, 8],
    points: 'all',
    side: 'positive',
    box: {
      visible: true
    },
    boxpoints: true,
    line: {
      color: 'black'
    },
    fillcolor: '#8dd3c7',
    opacity: 0.6,
    meanline: {
      visible: true
    },
    y0: 0.0
}];
```

I think we are going to have to correlate the currently `0` vpad with the span of the box and visible points.

The fact that the image tests passed tells me an example like ^ is not in mocks.  I think we should add a single sided violin baseline.

alexcjohnson: > What does 🌴 mean? _Don't repeat yourself_?

yeah sorry, DRY - I guess that's just a plotly.js convention.

> I noticed another bug - if we choose to show the box, it gets cut in half if we choose `positive` or `negative` for `side`

That's intentional, eg for two back-to-back violins. https://github.com/plotly/plotly.js/blob/master/test/image/baselines/violin_side-by-side.png

src/traces/violin/defaults.js
```diff
@@ -26,9 +26,15 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
-    coerce('scalemode');
     coerce('side');
+    var width = coerce('width');
+    if(!width) {
+        coerce('scalegroup', traceOut.name);
+        coerce('scalemode');
+    } else {
+        traceOut.scalegroup = '';
+        traceOut.scalemode = 'width';
+    }
```
alexcjohnson: Does anything break if we 🔪 the `else` entirely? Unnecessary attributes should simply not be included in `traceOut` at all.

Kully: yeah the violins are not drawn properly: the outlines are filled with rgb(0,0,0)

etpinard:
Hmm. Looks like `{violin|box}gap` and `{violin|box}groupgap` still have an effect here when `width` is set. Meaning `width` corresponds to the box/violin width **only** when `*gap` and `*groupgap` are set to `0`.

Is this desired behavior? Or should we do like `bar` traces and ignore `*gap` and `*groupgap` attributes whenever a user sets `width`?

etpinard:
> Or should we do like `bar` traces and ignore `*gap` and `*groupgap` attributes whenever a user sets `width`?

I implemented this along with fixing https://github.com/plotly/plotly.js/pull/3109#discussion_r230226013 and improving the attribute descriptions in -> 

https://github.com/plotly/plotly.js/compare/joyplots...joyplots-etpinard

Kully:
> I implemented this along with fixing [#3109 (comment)](https://github.com/plotly/plotly.js/pull/3109#discussion_r230226013) and improving the attribute descriptions in ->

Nice stuff!

Kully:
> I implemented this along with fixing [#3109 (comment)](https://github.com/plotly/plotly.js/pull/3109#discussion_r230226013) and improving the attribute descriptions in ->
> 
> [joyplots...joyplots-etpinard](https://github.com/plotly/plotly.js/compare/joyplots...joyplots-etpinard)

I gave this a quick scan - if you are fixing the coercing issue with these two variables, I'm good with that. Looks good to merge in.

test/jasmine/tests/violin_test.js
```diff
@@ -533,7 +550,7 @@ describe('Test violin hover:', function() {
 
             Plotly.plot(gd, fig).then(function() {
                 mouseEvent('mousemove', 300, 250);
-                assertViolinHoverLine([299.35, 250, 250, 250]);
+                assertViolinHoverLine([178.67823028564453, 250, 80, 250]);
```
etpinard: @Kully do you _why_ this test had to be modified?

Kully: I wasn't too sure to be honest. I ran the test a few times but there were only two values the assert was saying was valid, iirc.

alexcjohnson: Presumably it changed because autorange can shrink depending on `side` now, so the violin moved.

Kully:
You can get a cool effect by stacking violins with big widths on top of each other:
newplot

Kully:
I think I have solved the points getting cut off issue except for when `jitter` is other than 0. Now points that are below a positive trace _are_ in the range, but right at their midpoints. Same goes with negative violin traces:

newplot

Kully:
@etpinard I mistakenly didn't test your code after merging it into my PR, and now my violin points are not showing up.

Do you have any idea why that might be the case?

Kully:
It may have to do with my logic for setting the vpads, but its strange that setting `pointpos` to `0` still doesnt show the points. Here is an example to show you what is happening:

```javascript
function violin(x, y0, fillcolor, width, side, pointpos) {
    var v = {
        type: 'violin',
        width: width,
        // scalemode: scalemode,
        // scalegroup: scalegroup,
        x: x,
        points: 'all',
        jitter: 0,
        pointpos: pointpos,
        side: side,
        box: {
          visible: true,
        },
        boxpoints: true,
        line: {
          color: 'black'
        },
        fillcolor: fillcolor,
        opacity: 0.6,
        meanline: {
          visible: true
        },
        y0: y0
    };
    return v;
}

var array = [];
var length = 20
for(i = 0; i < length; i++) {
  array.push(12 + 0.5 * i);
}
// var array = [5, 8, 9, 10, 11, 4, 3, 3, 3, 2];

var data = [
  violin([0, 5, 7, 8], 0.0, '#8dd3c7', 0, 'negative', 0.0),
  violin([20, 25, 27, 28], 0.0, '#8dd3c7', 0, 'positive', 0.0),
  // violin(array, 0.1, '#c78dd3', 0, 'positive'),
];

var layout = {
  title: "Joyplot - Violin with multiple widths",
  xaxis: {
    zeroline: false
  },
  violingap: 0
};

```

in master this shows up as:

newplot

but in my PR now:

newplot 1

etpinard:
Now in https://github.com/plotly/plotly.js/pull/3234

PR #3111, opened 2018-10-16
support dom element as argument to Plotly.makeTemplate

antoinerg:
As a first-time user of templates, I expected the following to just work:
```
var.gd = document.getElementById('graph');
Plotly.plot(gd,{...});2,3], y:[4,9,3], line: {color: 'red'}}
...
var layout = Plotly.makeTemplate(gd);
```

This PR enables calling `makeTemplate` with a DOM element.

### DIFF
--- a/test/jasmine/tests/template_test.js
+++ b/test/jasmine/tests/template_test.js
@@ -141,6 +141,33 @@ describe('makeTemplate', function() {
 
         expect(template).toEqual(expected);
     });
+
+    it('works on DOM element', function(done) {
+        var mock = Lib.extendDeep({}, scatterFillMock);
+        var gd = createGraphDiv();
+
+        Plotly.newPlot(gd, mock)
+      .then(function() {
+          var template = Plotly.makeTemplate(gd);
+          delete(template.layout.xaxis);
+          delete(template.layout.yaxis);
+          expect(template).toEqual({
+              data: {scatter: [
+                {fill: 'tonext', line: {shape: 'spline'}},
+                {fill: 'tonext'},
+                {fill: 'toself'}
+              ] },
+              layout: {
+                  title: 'Fill toself and tonext',
+                  width: 400,
+                  height: 400
+              }
+          });
+      })
+      .catch(failTest)
+      .then(destroyGraphDiv)
+      .then(done);
+    });
 });
 
 // statics of template application are all covered by the template mock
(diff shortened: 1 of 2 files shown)


test/jasmine/tests/template_test.js
```diff
@@ -141,6 +141,33 @@ describe('makeTemplate', function() {
...
+
+    it('works on DOM element', function(done) {
+        var mock = Lib.extendDeep({}, scatterFillMock);
+        var gd = createGraphDiv();
+
+        Plotly.newPlot(gd, mock)
+      .then(function() {
+          var template = Plotly.makeTemplate(gd);
+          delete(template.layout.xaxis);
+          delete(template.layout.yaxis);
```
antoinerg: For this test to pass, some elements needed to be deleted (namely `layout.xaxis` and `layout.yaxis`. The result is not exactly the same as calling `makeTemplate` with the figure object. Isn't this problematic? 🤔

test/jasmine/tests/template_test.js
```diff
@@ -141,6 +141,33 @@ describe('makeTemplate', function() {
 
         expect(template).toEqual(expected);
     });
+
+    it('works on DOM element', function(done) {
+        var mock = Lib.extendDeep({}, scatterFillMock);
+        var gd = createGraphDiv();
+
+        Plotly.newPlot(gd, mock)
+      .then(function() {
```
alexcjohnson: weird indentation from here down?

alexcjohnson:
Thanks @antoinerg - beautiful test. After the indentation fix, 💃 !

test/jasmine/tests/template_test.js
antoinerg: I am puzzled. Do you know why it passed the syntax test?

test/jasmine/tests/template_test.js
alexcjohnson: Yep, that's to be expected 👍

test/jasmine/tests/template_test.js
alexcjohnson: I don't think the syntax test cares about method call indentation. It's possible we explicitly disabled that check, perhaps since d3 in particular encourages unusual conventions like
```js
d3.select(cls)
    .attr({...})
    .style({...})
  .select('path')
    .attr({...});
```
but also sometimes it looks best to have all the method calls indented from the initial line and sometime you want them aligned with the initial line.

Issue #3101, opened 2018-10-12
Parcoords line color restyle problems


etpinard:
See https://codepen.io/etpinard/pen/YJVrKm

archmoj:
Version 1.35.2 released on March 9, 2018 looks like the last old version of plotly in which one of the restyle or relayout functions works:
One could refer to [this code pen demo](https://codepen.io/MojtabaSamimi/pen/mzXGKr), and also test the functionality with version 1.36.0 or higher.

archmoj:
@etpinard Just noticed that the codepen I'd created above was showing the latest version in which the restyle call use to work properly on parcoords (of course before it was fixed again in v1.42.0 ).
So this other [demo](https://codepen.io/MojtabaSamimi/pen/GYagEb) actually displays the performance at version 1.41.3 with a white context layer at the start.
If required I may add another test to lock this as well?

etpinard:
![peek 2018-10-30 11-01](https://user-images.githubusercontent.com/6675409/47727714-4b536700-dc33-11e8-845f-52c29f41c206.gif)

Now I see. That first rendered view is missing a grey line. Good eye!



> If required I may add another test to lock this as well?

That would be great! No need to do this now, opening a new issue about it would suffice.

Issue #3099, opened 2018-10-11
parcoords + hovermode modebar button bug


alexcjohnson:
Deselected lines disappear after you click "toggle show closest data on hover" a couple of times.
https://codepen.io/mobigliani/pen/bmqVgZ
They come back as soon as you re-filter (drag one of the selection regions, or create a new one).

Also if I select only a small region at the bottom, sometimes the selection region gets reset to the whole range, but this does not seem to recalculate the selected lines. I have only seen this on the first column, and even there it seems to be intermittent.

![parcoords hovermode bug](https://user-images.githubusercontent.com/2678795/46828228-b1517a80-cd68-11e8-9d7d-c8274e63cbe5.gif)

alexcjohnson:
Closed in #3123

PR #3112, opened 2018-10-16
Parcoords issues with extra update call

archmoj:
Regarding issues #3099 and #3101 this patch removed an extra update call which may be the source of this problem. Please notice that lineLayerMaker itself also calls the update method to redraw.

@alexcjohnson

### DIFF
--- a/src/traces/parcoords/parcoords.js
+++ b/src/traces/parcoords/parcoords.js
@@ -457,8 +457,7 @@ module.exports = function(root, svg, parcoordsLineLayers, styledData, layout, ca
     parcoordsLineLayer
         .each(function(d) {
             if(d.viewModel) {
-                if(d.lineLayer) d.lineLayer.update(d);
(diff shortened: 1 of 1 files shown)


alexcjohnson:
This doesn't seem to me to be *removing* an update, so much as converting an update into a new object creation call. Which does seem to fix the issue, but also slows down updates a good deal (and might lead to a memory leak? That I'm not sure about, depends on what's contained in the `linelayer` object)

Anyway, I think we had better look a little deeper and see what it would take to fix `linelayer.update` rather than just bypassing it.

archmoj:
Yes I totally agree with you that this requires more tests.

archmoj:
We could close this PR now that the issues are addressed in these two different PRs: 
[PR for issue 3099](https://github.com/plotly/plotly.js/pull/3123) 
[PR for issue 3101](https://github.com/plotly/plotly.js/pull/3125)

PR #3114, opened 2018-10-16
ignore invalid trace indices in restyle and update

### LINKED ISSUE #3113: `Plotly.restyle` and (`Plotly.update`) throws error if trace indice is invalid or out of bound
`restyle` (and `update`) should not throw an error if you give invalid trace indices, it should just toss them (this behavior was requested by @alexcjohnson)!

Here's a codepen reproducing the issue: https://codepen.io/anon/pen/RexqKO

antoinerg:
Fixes #3113

### DIFF
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -491,7 +491,15 @@ exports.coerceTraceIndices = function(gd, traceIndices) {
         return gd.data.map(function(_, i) { return i; });
     }
     else if(Array.isArray(traceIndices)) {
-        return traceIndices.filter(function(i) {return i < gd.data.length;});
+        var traceIndicesOut = [];
+        for(var i = 0; i < traceIndices.length; i++) {
+            if(Lib.isIndex(traceIndices[i], gd.data.length)) {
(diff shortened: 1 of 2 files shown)


src/plot_api/helpers.js
```diff
@@ -490,6 +490,9 @@ exports.coerceTraceIndices = function(gd, traceIndices) {
     else if(!Array.isArray(traceIndices) || !traceIndices.length) {
         return gd.data.map(function(_, i) { return i; });
     }
+    else if(Array.isArray(traceIndices)) {
+        return traceIndices.filter(function(i) {return i < gd.data.length;});
+    }
```
antoinerg: Should I do a `Lib.warn('indice is not number or is out of bound')?

antoinerg: Hooray for `lib.isIndex`. I really should spend some time to find out all the functions we have in `lib`!

alexcjohnson:
Thanks for the quick fix @antoinerg ! 🚀 Very nice tests.
💃

etpinard:
:sos: delete branch alert :sos:

PR #3119, opened 2018-10-17
add cors support for devtools http server

antoinerg:
This allows talking to the devtools http server from any webpage. This enables my prototype image viewer (hosted on another domain) to retrieve mocks and images from localhost!


antoinerg:
BTW, here's a [preview of that image viewer](https://antoinerg.github.io/plotly-image-viewer/#/). It will fetch the list of mocks (as well as the images) from `localhost:3000` provided the dev server is running from the current PR :)

PR #3120, opened 2018-10-17
implied edits connecting height/width/autosize

alexcjohnson:
So you can turn autosize back on again, using `relayout` or `update`, after it's been turned off and replaced by explicit `height` & `width`.

@antoinerg please review. Shouldn't have any adverse impact on the autosize work you did recently as this only affects edit behavior (and only the older modes, not `react`)

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -1922,8 +1922,24 @@ function _relayout(gd, aobj) {
         // back to its initial value as computed during the first pass in Plots.plotAutoSize.
         //
         // To do so, we must manually set them back here using the _initialAutoSize cache.
-        if(['width', 'height'].indexOf(ai) !== -1 && vi === null) {
-            fullLayout[ai] = gd._initialAutoSize[ai];
+        // can't use impliedEdits for this because behavior depends on vi
+        if(['width', 'height'].indexOf(ai) !== -1) {
+            if(vi) {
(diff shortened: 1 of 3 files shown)


src/plot_api/plot_api.js
```diff
@@ -1922,8 +1922,24 @@ function _relayout(gd, aobj) {
         // back to its initial value as computed during the first pass in Plots.plotAutoSize.
         //
         // To do so, we must manually set them back here using the _initialAutoSize cache.
-        if(['width', 'height'].indexOf(ai) !== -1 && vi === null) {
-            fullLayout[ai] = gd._initialAutoSize[ai];
+        // can't use impliedEdits for this because behavior depends on vi
+        if(['width', 'height'].indexOf(ai) !== -1) {
+            if(vi) {
+                doextra('autosize', null);
```
alexcjohnson: FYI `doextra(attr, null)` works (with some caveats) just like `relayout(gd, attr, null)` in that it deletes that attribute from `layout`

alexcjohnson:
Oh hmm, @antoinerg turns out I *did* impact at least your autosize tests - need to look at whether those failures make sense.

antoinerg:
I don't think it interferes with any of my work on `responsive` and `autosize`. The tests that are failing are from before my time. I will need to take a closer look 👀

PR #3123, opened 2018-10-18
Display parcoords deselected lines

archmoj:
This PR could fix the issue #3099 i.e. before the implementation of hovermode for parcoords.

@alexcjohnson

### DIFF
--- a/src/traces/parcoords/parcoords.js
+++ b/src/traces/parcoords/parcoords.js
@@ -456,12 +456,18 @@ module.exports = function(root, svg, parcoordsLineLayers, styledData, layout, ca
 
     parcoordsLineLayer
         .each(function(d) {
+
             if(d.viewModel) {
                 if(d.lineLayer) d.lineLayer.update(d);
                 else d.lineLayer = lineLayerMaker(this, d);
 
                 d.viewModel[d.key] = d.lineLayer;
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Looks great @archmoj ! Can you add a test to 🔒 this fix? I was going to suggest modeling it after this test:

https://github.com/plotly/plotly.js/blob/3f95f04d53e16b6074b59365d5420a07ef75733e/test/jasmine/tests/parcoords_test.js#L655-L663

(with the inverse expectation of course, `foundPixel` should be `true`) And while I think the inner part of that test may still be good, the test itself is broken - it doesn't bother to test that `d3.selectAll('.parcoords-lines')` actually yields any elements! (add the line `expect(d3.selectAll('.parcoords-lines').size()).not.toBe(0);` and it fails) So the `.each` never runs. Must have been refactored so the `.parcoords-lines` class is never used.

Don't spend any time fixing that test now unless it's trivial, but please do come up with a similar test showing that all the layers have some data in them after an edit. No need to simulate clicking the modebar button, some other edit like the one I mentioned in #3099 `Plotly.relayout(gd,'paper_bgcolor','#eff')` will suffice.

archmoj:
Thank you @alexcjohnson 
:dancer:

PR #3125, opened 2018-10-18
Enable restyle to redraw parcoords #3101

archmoj:
This PR enables callbacks e.g. restyle  to redraw parcoords.
This fixes https://github.com/plotly/plotly.js/issues/3101
@alexcjohnson


src/traces/parcoords/parcoords.js
```diff
@@ -456,18 +456,20 @@ module.exports = function(root, svg, parcoordsLineLayers, styledData, layout, ca
...
             if(d.viewModel) {
-                if(d.lineLayer) d.lineLayer.update(d);
-                else d.lineLayer = lineLayerMaker(this, d);
+                if((!d.lineLayer) ||
+                    (callbacks)) { // recreate in case of having callbacks e.g. restyle. Should we test for callback to be a restyle?
+                    d.lineLayer = lineLayerMaker(this, d);
+                } else d.lineLayer.update(d);
 
-                d.viewModel[d.key] = d.lineLayer;
+                if(d.key) {
```
alexcjohnson: when is `d.key` falsey? Makes sense not to set `d.viewModel[d.key]` if it doesn't have a `key` but what about `d.lineLayer.render`?

src/traces/parcoords/parcoords.js
```diff
@@ -456,18 +456,20 @@ module.exports = function(root, svg, parcoordsLineLayers, styledData, layout, ca
...
     parcoordsLineLayer
         .each(function(d) {
-
             if(d.viewModel) {
-                if(d.lineLayer) d.lineLayer.update(d);
-                else d.lineLayer = lineLayerMaker(this, d);
+                if((!d.lineLayer) ||
+                    (callbacks)) { // recreate in case of having callbacks e.g. restyle. Should we test for callback to be a restyle?
+                    d.lineLayer = lineLayerMaker(this, d);
+                } else d.lineLayer.update(d);
```
alexcjohnson: https://github.com/plotly/plotly.js/issues/3101#issuecomment-430694449 implies that in an earlier version this worked correctly using `update` even if there *are* callbacks, yes? It would be much preferred if we can get it to work that way again.

archmoj: I don't see an update call in the old version 1.35.2. It may used to create new lines by default which is not efficient.

src/traces/parcoords/parcoords.js
archmoj: Not sure if we always get a good d.key, so I thought we could check the index before setting the array.
In respect to your comment I put the render call outside the if statements. 
Now with the increase of delay, we've got success for the rest of the tests.

src/traces/parcoords/parcoords.js
alexcjohnson: ha you're right, there isn't even an update method in `lineLayer` at that version! That was added by... uh, [me](https://github.com/plotly/plotly.js/blame/6de004240acfba04987ded259c3511d2526d5240/src/traces/parcoords/lines.js#L295) in https://github.com/plotly/plotly.js/pull/2415 😳

alexcjohnson:
Alright, at some point we should dive in and figure out what it will take to make `lineLayer.update` really handle arbitrary updates so they'll be faster. But in the meantime this looks like a solid fix. Only thing we need now is a test or two showing that this fix does allow us to change line color or colorscale.

test/jasmine/tests/config_test.js
```diff
@@ -7,6 +7,7 @@ var click = require('../assets/click');
 var mouseEvent = require('../assets/mouse_event');
 var failTest = require('../assets/fail_test');
 var delay = require('../assets/delay');
+var RESIZE_DELAY = 300;
```
alexcjohnson: cc @antoinerg - bad day for circleci it seems, we were getting repeated test failures here so I encouraged @archmoj to try increasing this delay. At some point we may want to change this so we're waiting on some event (listen for `plotly_afterplot` perhaps?) rather than a fixed delay, that should be more robust (and usually faster!)

archmoj:
The restyle function is tested interactively. Do we still need to create a jasmine test?

alexcjohnson:
> Do we still need to create a jasmine test?

Yes - especially since at some point we want to improve performance by fixing `update`, we should write a few tests that we can rely on during that work to ensure nothing has changed.

archmoj:
There was a test available for colorscale; so I  added another one for line color.

archmoj:
Yes totally agree on that.

src/traces/parcoords/parcoords.js
```diff
@@ -456,16 +456,16 @@ module.exports = function(root, svg, parcoordsLineLayers, styledData, layout, ca
 
     parcoordsLineLayer
         .each(function(d) {
-
             if(d.viewModel) {
-                if(d.lineLayer) d.lineLayer.update(d);
-                else d.lineLayer = lineLayerMaker(this, d);
+                if((!d.lineLayer) ||
```
etpinard: That extra `( )` style feels somewhat to our codebase. Maybe we could add https://eslint.org/docs/rules/no-extra-parens to lintinting rules.

etpinard:
If I understand correctly, PR https://github.com/plotly/plotly.js/pull/2415 was also the culprit for bug https://github.com/plotly/plotly.js/issues/3099 (fixed in https://github.com/plotly/plotly.js/pull/3123)?

archmoj:
@etpinard 
For the case of FireFox, the program actually get into plot_api and `Plots.cleanPlot([], {}, gd._fullData, fullLayout);`. That explains why newFullLayout was empty on FireFox.
But when we define `width` & `height` in the layout the issue goes away.
Do we have any default values for `width` and `height`? Or are they declared by the browser/window when not being specified?

etpinard:
Very interesting. FF must not like this block

https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/plot_api/plot_api.js#L261-L279

added in https://github.com/plotly/plotly.js/pull/2939

Could you check if `regl._gl.drawingBufferWidth` and `regl._gl.drawingBufferHeight` are defined in FF?

archmoj:
On FF when width and height are not declared in the layout we get float values (e.g. fullLayout.width) that we shouldn't get. So they could be rounded somewhere upstream? by either floor function or another round to nearest function. What do you think @etpinard ?
`fullLayout.widthXheight= [1082.05, 450]`
`regl._gl.drawingBufferWidthXheight= [1082, 450]`

test/jasmine/tests/config_test.js
archmoj: @antoinerg What value you suggest we use here for `RESIZE_DELAY` now that the CI is back to normal?

etpinard:
@archmoj do you need help adding a test for this thing? This is a pretty bad bug that we should merge in before releasing 1.42.0

archmoj:
Thank you! Yes actually I am working on a jasmine test for this. I would let you know.

test/jasmine/tests/parcoords_test.js
```diff
@@ -868,7 +871,65 @@ describe('parcoords basic use', function() {
...
     });
 
-    it('@gl Calling `Plotly.restyle` with a string path should amend the preexisting parcoords', function(done) {
+    function _getGrayRatio(secondPass, msg)
+        var canvases = d3.selectAll('.gl-canvas');
+        expect(canvases.size()).toBe(3, msg);
+        
+        canvases.each(function(element, index) {
+          
+            if (index == 0) { // FIXME: we assumed here that the context is the first item but may be not.
```
etpinard: Yes, the context _should_ always be the first one, if it's not maybe other tests will fail (not just this one). No need to for an additional check.

test/jasmine/tests/parcoords_test.js
```diff
@@ -868,7 +871,65 @@ describe('parcoords basic use', function() {
 
     });
 
-    it('@gl Calling `Plotly.restyle` with a string path should amend the preexisting parcoords', function(done) {
+    function _getGrayRatio(secondPass, msg)
+        var canvases = d3.selectAll('.gl-canvas');
+        expect(canvases.size()).toBe(3, msg);
+        
+        canvases.each(function(element, index) {
```
etpinard: ... but alternatively, you could select `.gl-canvas-context` instead of all three `<canvas>` elements

test/jasmine/tests/parcoords_test.js
```diff
@@ -868,7 +871,65 @@ describe('parcoords basic use', function() {
...
+                var imageArray = readPixel(this, 0, 0, this.width, this.height);
+                var totalRGB = 0;
+                var n = imageArray.length;
+                for(var i = 0; i < n; i++) {
+                    var r = imageArray[i][0];
+                    var g = imageArray[i][1];
+                    var b = imageArray[i][2];
+                    totalRGB += r + g + b;
+                }
+                if(secondPass > 0) {
```
etpinard: I don't like this pattern. Let's keep `getGrayRatio` as just a _getter_ function and move the `expect` statement to down to `it` block below:

```js
        var totalRGB;
        
        Plotly.plot(gd, mock3)
        .then(function() {
            totalRGB = _getGrayRatio();

            mockCopy.data[0].dimensions[1].constraintrange = [0.4, 0.6];          
            return Plotly.react(gd, mockCopy);
        })
        .then(function() {
            var totalRGB2 = _getGrayRatio();
            expect(totalRGB2).toBe(totalRGB, 'after react');
        })
        .catch(failTest)
        .then(done);
```

package-lock.json
```diff
@@ -17,7 +17,7 @@
     "3d-view-controls": {
       "version": "2.2.2",
       "resolved": "https://registry.npmjs.org/3d-view-controls/-/3d-view-controls-2.2.2.tgz",
-      "integrity": "sha512-WL0u3PN41lEx/4qvKqV6bJlweUYoW18FXMshW/qHb41AVdZxDReLoJNGYsI7x6jf9bYelEF62BJPQmO7yEnG2w==",
+      "integrity": "sha1-cXP8GX6efk28YyE0OEZwR9vIT6I=",
```
etpinard: Can you undo these package-lock changes? They have nothing to do with this PR.

etpinard:
Now in https://github.com/plotly/plotly.js/pull/3178

Issue #3010, opened 2018-09-14
displaying left aligned and right aligned data in hover items


chriddyp:
Somewhat related to #3007. The ability to better display left aligned and right aligned data like this:

![capture d ecran 2018-09-13 a 22 07 08](https://user-images.githubusercontent.com/1280389/45525562-8859bc00-b7a1-11e8-9b84-42cb830274c4.png)

Example taken from https://www.nytimes.com/interactive/2016/04/29/upshot/money-race-and-success-how-your-school-district-compares.html

etpinard:
Related to https://github.com/plotly/plotly.js/issues/2342 also.

etpinard:
Related to https://github.com/plotly/plotly.js/issues/1882 as well.

etpinard:
Attributes for vertical alignment in hover labels would be nice also, from @alexcjohnson's https://github.com/plotly/plotly.js/issues/3442#issuecomment-455603144 

> But an attribute to switch that to middle-aligned or bottom-aligned would be great to add at some point.

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

Issue #2618, opened 2018-05-07
Axis title could also change the labels in the hovering popup


adube:
This is an enhancement suggestion.

When setting the title of an axis, such as:

```
        xaxis: {
            type: 'linear',
            zeroline: false,
            title: "hellox"  // <----
        },
```

it would make sense to have the same title as default value in the popup being shown on mouse hovering.  Currently, the default value stays "x", where it could have been "hellox" instead.

See an example here: https://codepen.io/anon/pen/qoGQvx

See also the riginal question asked on [Stack Overflow](https://stackoverflow.com/questions/49820560/how-to-change-only-the-title-of-hovering-popup-in-plotly-js?noredirect=1#comment87407776_49820560)

Please, let me know if that makes sense.

alexcjohnson:
Thanks @adube - yes, makes a lot of sense. I don't think we can *always* do this, there are times when the title is too long, or displayed on one subplot and not others, etc... so it would probably have to be opt-in unless we can come up with a sufficiently robust criterion for automatically using the title.

adube:
@alexcjohnson Thanks for your comment.

I agree.  As a way to not _always_ enable this behaviour, we could have some sort of "flag" option (boolean) which would be `false` by default to not break the current behaviour.  When setting it to `true`, users could expect limitations such as title being too long and the other things you mentioned.

I'm not familiar enough with the code to be able to determine a more robust solution at this point, but I think the above suggestion could do the trick.

nicolaskruchten:
I think this makes a ton of sense in general and especially in 3d and ternary contexts: 3d because after a few rotations or whatever, it's no longer visually clear what `x`, `y`, and `z` mean, and ternary because there's no reasonable way a user could know what `A`, `B` and `C` map to ;)

etpinard:
Somewhat a duplicate of https://github.com/plotly/plotly.js/issues/265

adube:
Indeed, it is.

vvch:
> I agree. As a way to not _always_ enable this behaviour, we could have some sort of "flag" option (boolean) which would be `false` by default to not break the current behaviour. When setting it to `true`, users could expect limitations such as title being too long

This option (e.g., 'hovertitle') may also contain text (for each axis) as alternative short hover-only title for this axis. That is, default 'false' value means current behaviour, 'true' means use axis label (maybe with some automatic abbreviation), and text value means use this value.

vvch:
[...]

Issue #2504, opened 2018-03-27
Pass customdata to 'fill' or trace itself.


marcodlk:
I am a plotly.py/Dash user and recently ran into a little blocker regarding the customdata feature.

**The situation:**
I am using go.Scatter with options 
`
name='trace_name',  
mode='lines',  
fill='toself',  
hoveron='fills+points',  
customdata=some_preset_data,  
text=some_preset_text  
` 
to make traces that each represent a polygon. 

**Limits of current customdata feature:**
The hoveron='fills+points' option allows the user to see text and interact with customdata when hovering/clicking on points, which is great. However, when hovering over the fill (which gives the feeling of representing the trace itself), the user is limited to seeing the name property ('trace_name' here) on hover and only being able to interact with the curveNumber of the trace - customdata and text don't show up because they are 'unique' to each point.

**Feature request:**
As a user, I would like to be able to attach customdata and/or text to fill elements or the trace itself. Purely from a user point of view, there seems to be a disconnect there, where the 'fills' and 'points' in the hoveron='fills+points' options have highly different levels of hoveron customizability/functionality.

Unfortunately, I am not a javascript programmer and cannot add much (if anything) to how this would be implemented, but I was directed here from the plotly/dash forum as this seems to be an issue pertinent to plotly.js itself.

Relevant plotly/dash forum:
https://community.plot.ly/t/dcc-graph-only-curvenumber-is-showing-up/9225

etpinard:
That would be nice for sure! Thanks for writing in.

The hardest part about this here is to find a suitable API. For point data, coordinates and fields like `customdata` (and `hovertext` for that matter) match 1-to-1. For fills, the relationship isn't bijective. We could of course tell users to make their `customdata` match the number of polygons or more technically the number of `null` in their coordinates arrays plus one. Something like this has been proposed already for other things like https://github.com/plotly/plotly.js/issues/147. But then, what should we do for `hoveron: 'fill+points'`? We won't be able to honor both the `customdata` to polygon and `customdata` to coordinate maps. So perhaps the solution is to add a new field something like `customfilldata`? But that's sounds a little dirty. I hope we can find a cleaner solution.

marcodlk:
Would it be possible to have something like: 
```python
customdata={
    'points': customdata_for_points,
    'fill': customdata_for_fills
},
customdatamode='points+fills'
```
[...]

PR #3126, opened 2018-10-18
support template string on hover

### LINKED ISSUE #3007: Template strings in hover data
This could include both attributes that exist in the `trace`
```
trace.hovertemplate = '{text}<br>{marker.size} Items<br>{y} Dollars'
```

But also data that is computed on-the-fly by plotly.js, like the bin widths:
```
trace.hovertemplate = 'The number of people between the age of {binleft} and {binright} is {bintotal}.'
```

This means you could express the existing box and violin text with some type of `hovertemplate`

This could span across traces, like the hover data on stacked bars
```
trace.template = 'This represents {trace.y} out of {data.total}'
```

Or one day, perhaps even trace level computations, like the sum of the area
```
texttemplate = 'The sum of the money is {trace.sum}'
```

antoinerg:
Closes #3007 

However, there are several caveats:

1 - In the process of making that PR, I realized that there are a lot of computed values on hover that are not emitted in hover events. Because we are passing the hover event data to the template, only a limited subset of the total information is currently available. Enriching the event data for each trace is probably outside the scope of the current PR since it should benefit BOTH `hovertemplate` and `plotly_hover` events. However, it is something that can be done fairly easily and separately (some traces already have an `event_data.js` file in place where such code would go)!

2 - Some traces have many hover labels. For example, in `box` traces, there are six labels when hovering on a box. As of right now, the numerical value for each of those is assigned to variable `y` regardless of what it represents (`q(1|3)`, `min`, `max`, etc.). We should certainly add a `label` variable alongside in order for template `'%{label}: %{y}'` to exactly reproduce the current hover labels while enabling number formatting. All traces with several different hover labels would have to implement this change.

In the future, to support different templates for each hover labels, we could either allow `hovertemplate` to be an object with a key for each label (`{q1: 'templatestring', min: 'templatestring', ...}`)  OR we can move everything in nested attributes :  `{q1: {hovertemplate: '...'}, min: {hovertemplate: '...'}, ...}`. The latter has the advantages of allowing further customization for each label (think color or hiding)  since each of them could have a nested `hoverinfo` and `hoverlabel`. We need to think carefully about this one. The latter approach is what @alexcjohnson and I decided to do in our latest work for `sankey` traces (see PR https://github.com/plotly/plotly.js/pull/3096).

3 - Right now, `hovertemplate` doesn't control the content of the secondary labels. As @alexcjohnson pointed out, this could be supported by a special tag `<extra>Secondary box content here</extra>`.

Please chime in :)

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -144,7 +144,8 @@ exports.loneHover = function loneHover(hoverItem, opts) {
         rotateLabels: false,
         bgColor: opts.bgColor || Color.background,
         container: container3,
-        outerContainer: outerContainer3
+        outerContainer: outerContainer3,
+        hovertemplate: opts.hovertemplate || false
     };
 
     var hoverLabel = createHoverText([pointData], fullOpts, opts.gd);
@@ -200,7 +201,8 @@ exports.multiHovers = function multiHovers(hoverItems, opts) {
         rotateLabels: false,
         bgColor: opts.bgColor || Color.background,
         container: container3,
-        outerContainer: outerContainer3
+        outerContainer: outerContainer3,
+        hovertemplate: opts.hovertemplate || false
     };
 
     var hoverLabel = createHoverText(pointsData, fullOpts, opts.gd);
@@ -948,6 +950,17 @@ function createHoverText(hoverData, opts, gd) {
             // if 'name' is also empty, remove entire label
             if(name === '') g.remove();
             text = name;
+        }
+
+        // hovertemplate
+        var trace = d.trace, hovertemplate = opts.hovertemplate || trace.hovertemplate || false;
+        if(hovertemplate) {
+            var i = d.index;
+            var obj = {
+                x: trace.x[i],
+                y: trace.y[i]
+            };
+            text = Lib.templateString(hovertemplate, {text: text}, d.cd[i], obj, trace, gd._fullLayout);
         }
 
         // main label
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -676,12 +676,6 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         hoverData[itemnum].hovertemplate = ht || pt.trace.hovertemplate || false;
         hoverData[itemnum].eventData = [eventData];
 
-        // Add axis information to eventData if hovertemplate
-        if(hoverData[itemnum].hovertemplate) {
-            eventData.xaxis = gd._fullLayout[pt.trace.xaxis + 'axis'];
-            eventData.yaxis = gd._fullLayout[pt.trace.yaxis + 'axis'];
-        }
-
         newhoverdata.push(eventData);
     }
 
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -145,9 +145,9 @@ exports.loneHover = function loneHover(hoverItem, opts) {
         bgColor: opts.bgColor || Color.background,
         container: container3,
         outerContainer: outerContainer3,
-        hovertemplate: opts.hovertemplate || false
+        hovertemplate: opts.hovertemplate || false,
+        eventData: opts.eventData || {}
     };
-
     var hoverLabel = createHoverText([pointData], fullOpts, opts.gd);
     alignHoverText(hoverLabel, fullOpts.rotateLabels);
 
@@ -202,7 +202,8 @@ exports.multiHovers = function multiHovers(hoverItems, opts) {
         bgColor: opts.bgColor || Color.background,
         container: container3,
         outerContainer: outerContainer3,
-        hovertemplate: opts.hovertemplate || false
+        hovertemplate: opts.hovertemplate || false,
+        eventData: opts.eventData || {}
     };
 
     var hoverLabel = createHoverText(pointsData, fullOpts, opts.gd);
@@ -686,7 +687,8 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         container: fullLayout._hoverlayer,
         outerContainer: fullLayout._paperdiv,
         commonLabelOpts: fullLayout.hoverlabel,
-        hoverdistance: fullLayout.hoverdistance
+        hoverdistance: fullLayout.hoverdistance,
+        eventData: gd._hoverdata
     };
 
     var hoverLabels = createHoverText(hoverData, labelOpts, gd);
@@ -957,8 +959,9 @@ function createHoverText(hoverData, opts, gd) {
         // hovertemplate
         var trace = d.trace;
         var hovertemplate = opts.hovertemplate || hoverData[curveNumber].hovertemplate || false;
+
         if(hovertemplate) {
-            text = Lib.hovertemplateString(hovertemplate, d, gd._hoverdata[curveNumber], trace);
+            text = Lib.hovertemplateString(hovertemplate, d, opts.eventData[curveNumber] || {}, trace);
 
             var EXTRA_STRING_REGEX = /<extra>(.*)<\/extra>/;
             text = text.replace(EXTRA_STRING_REGEX, function(match, extra) {
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -672,7 +672,9 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         var eventData = helpers.makeEventData(pt, pt.trace, pt.cd);
         newhoverdata.push(eventData);
 
-        hoverData[itemnum].hovertemplate = pt.trace.hovertemplate || false;
+        var ht = false;
+        if(pt.cd[pt.index] && pt.cd[pt.index].ht) ht = pt.cd[pt.index].ht;
+        hoverData[itemnum].hovertemplate = ht || pt.trace.hovertemplate || false;
         hoverData[itemnum].eventData = [eventData];
     }
 
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -971,7 +971,7 @@ function createHoverText(hoverData, opts, gd) {
         if(hovertemplate) {
             text = Lib.hovertemplateString(hovertemplate, hovertemplateLabels, eventData, trace);
 
-            var EXTRA_STRING_REGEX = /<extra>(.*)<\/extra>/;
+            var EXTRA_STRING_REGEX = /<extra>([\s\S]*)<\/extra>/;
             text = text.replace(EXTRA_STRING_REGEX, function(match, extra) {
                 name = extra; // Assign name for secondary text label
                 return ''; // Remove from main text label
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -146,6 +146,7 @@ exports.loneHover = function loneHover(hoverItem, opts) {
         container: container3,
         outerContainer: outerContainer3,
         hovertemplate: opts.hovertemplate || false,
+        hovertemplateLabels: opts.hovertemplateLabels || false,
         eventData: opts.eventData || {}
     };
     var hoverLabel = createHoverText([pointData], fullOpts, opts.gd);
@@ -203,6 +204,7 @@ exports.multiHovers = function multiHovers(hoverItems, opts) {
         container: container3,
         outerContainer: outerContainer3,
         hovertemplate: opts.hovertemplate || false,
+        hovertemplateLabels: opts.hovertemplateLabels || false,
         eventData: opts.eventData || {}
     };
 
@@ -959,9 +961,9 @@ function createHoverText(hoverData, opts, gd) {
         // hovertemplate
         var trace = d.trace;
         var hovertemplate = opts.hovertemplate || hoverData[curveNumber].hovertemplate || false;
-
+        var hovertemplateLabels = opts.hovertemplateLabels || d;
         if(hovertemplate) {
-            text = Lib.hovertemplateString(hovertemplate, d, opts.eventData[curveNumber] || {}, trace);
+            text = Lib.hovertemplateString(hovertemplate, hovertemplateLabels, opts.eventData[curveNumber] || {}, trace);
 
             var EXTRA_STRING_REGEX = /<extra>(.*)<\/extra>/;
             text = text.replace(EXTRA_STRING_REGEX, function(match, extra) {
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -670,6 +670,10 @@ function _hover(gd, evt, subplot, noHoverEvent) {
     for(itemnum = 0; itemnum < hoverData.length; itemnum++) {
         var pt = hoverData[itemnum];
         var eventData = helpers.makeEventData(pt, pt.trace, pt.cd);
+
+        // Add axis information to eventData
+        eventData.xaxis = gd._fullLayout[pt.trace.xaxis + 'axis'];
+        eventData.yaxis = gd._fullLayout[pt.trace.yaxis + 'axis'];
         newhoverdata.push(eventData);
 
         var ht = false;
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -974,7 +974,7 @@ function createHoverText(hoverData, opts, gd) {
         var hovertemplateLabels = d.hovertemplateLabels || d;
         var eventData = d.eventData[0] || {};
         if(hovertemplate) {
-            text = Lib.hovertemplateString(hovertemplate, hovertemplateLabels, eventData, trace);
+            text = Lib.hovertemplateString(hovertemplate, hovertemplateLabels, eventData);
 
             text = text.replace(EXTRA_STRING_REGEX, function(match, extra) {
                 name = extra; // Assign name for secondary text label
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -895,7 +895,7 @@ function createHoverText(hoverData, opts, gd) {
 
     // then put the text in, position the pointer to the data,
     // and figure out sizes
-    hoverLabels.each(function(d, curveNumber) {
+    hoverLabels.each(function(d) {
         var g = d3.select(this).attr('transform', '');
         var name = '';
         var text = '';
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -884,7 +884,7 @@ function createHoverText(hoverData, opts, gd) {
 
     // then put the text in, position the pointer to the data,
     // and figure out sizes
-    hoverLabels.each(function(d) {
+    hoverLabels.each(function(d, curveNumber) {
         var g = d3.select(this).attr('transform', '');
         var name = '';
         var text = '';
@@ -956,11 +956,7 @@ function createHoverText(hoverData, opts, gd) {
         var trace = d.trace, hovertemplate = opts.hovertemplate || trace.hovertemplate || false;
         if(hovertemplate) {
             var i = d.index;
-            var obj = {
-                x: trace.x[i],
-                y: trace.y[i]
-            };
-            text = Lib.templateString(hovertemplate, {text: text}, d.cd[i], obj, trace, gd._fullLayout);
+            text = Lib.templateString(hovertemplate, gd._hoverdata[curveNumber], {text: text}, d.cd[i], trace, gd._fullLayout);
         }
 
         // main label
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -969,7 +969,6 @@ function createHoverText(hoverData, opts, gd) {
         }
 
         // hovertemplate
-        var trace = d.trace;
         var hovertemplate = d.hovertemplate || false;
         var hovertemplateLabels = d.hovertemplateLabels || d;
         var eventData = d.eventData[0] || {};
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -955,7 +955,7 @@ function createHoverText(hoverData, opts, gd) {
         // hovertemplate
         var trace = d.trace, hovertemplate = opts.hovertemplate || trace.hovertemplate || false;
         if(hovertemplate) {
-            text = Lib.templateString(hovertemplate, gd._hoverdata[curveNumber], trace);
+            text = Lib.hovertemplateString(hovertemplate, d, gd._hoverdata[curveNumber], trace);
         }
 
         // main label
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -126,7 +126,7 @@ exports.loneHover = function loneHover(hoverItem, opts) {
         fontColor: hoverItem.fontColor,
 
         // filler to make createHoverText happy
-        trace: {
+        trace: hoverItem.trace || {
             index: 0,
             hoverinfo: ''
         },
@@ -183,7 +183,7 @@ exports.multiHovers = function multiHovers(hoverItems, opts) {
             fontColor: hoverItem.fontColor,
 
             // filler to make createHoverText happy
-            trace: {
+            trace: hoverItem.trace || {
                 index: 0,
                 hoverinfo: ''
             },
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -765,11 +765,13 @@ function createHoverText(hoverData, opts, gd) {
             if(allHaveZ && hoverData[i].zLabel === undefined) allHaveZ = false;
 
             traceHoverinfo = hoverData[i].hoverinfo || hoverData[i].trace.hoverinfo;
-            var parts = Array.isArray(traceHoverinfo) ? traceHoverinfo : traceHoverinfo.split('+');
-            if(parts.indexOf('all') === -1 &&
-                parts.indexOf(hovermode) === -1) {
-                showCommonLabel = false;
-                break;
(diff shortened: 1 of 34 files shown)


src/components/fx/hover.js
```diff
@@ -950,6 +952,12 @@ function createHoverText(hoverData, opts, gd) {
             text = name;
         }
 
+        // hovertemplate
+        var trace = d.trace, hovertemplate = opts.hovertemplate || trace.hovertemplate || false;
+        if(hovertemplate) {
+            text = Lib.templateString(hovertemplate, gd._hoverdata[curveNumber], trace);
```
antoinerg: We can pass additional objects as argument to `Lib.templateString` so the sky is the limit in terms of what values we want to make accessible here!

However, as mentioned in point 1, we should probably put this data into `_hoverdata` since it's emitted on hover.

mojoaxel:
@antoinerg You should consider adding a bunch of examples to [plotly/documentation](/plotly/documentation) to demonstrate the new possibilities with this.

etpinard:
> Enriching the event data for each trace is probably outside the scope of the current PR since it should benefit BOTH hovertemplate and plotly_hover events

Matching event data `points` keys with the upcoming `hovertemplate` `%{}` strings is a fantastic idea. The `points` keys are documented [here](https://plot.ly/javascript/plotlyjs-events/#event-data). The list isn't complete, but garanting that keys emitted during hover/click events work in `hovertemplate` will be a big plus in terms of user-discoverability and maintainability.

We'll need to handle traces that support a `hoveron` attribute (e.g. scatter traces that can "hover-on" points and fills) which don't always emit the same event data keys (see https://github.com/plotly/plotly.js/issues/2504 for more info). Perhaps we could have nested `hovertemplate` for each case e.g. `fills.hovertemplate` and `points.hovertemplate`?

----------------

> Some traces have many hover labels. For example, in `box` traces, there are six labels when hovering on a box.
> In the future, to support different templates for each hover labels, we could either allow `hovertemplate` to be an object with a key for each label

Thanks writing down thoughts about traces that can generate multiple hover labels. I agree, we'll need to implement per-label templating and styling as some point. 

In the case of `box` traces, I'm not a fan of the `{q1: {hovertemplate: '...'}, min: {hovertemplate: '...'}, ...}` proposal as it would conflict with https://github.com/plotly/plotly.js/issues/1059. I think it's important to make a distinction between cases where:

-  one part of a trace can generate multiple hover labels (e.g. 'boxes' hover) and
-  traces with multiple "parts" each with their own hover labels (e.g. sankey, but also parcats).

> this could be supported by a special tag <extra>Secondary box content here</extra>

Maybe these special tags could help us out. Consider:

```
boxes.hovertemplate: `
  <mean><span style="color:blue">MEAN: %{y}</span><name>%{data.name}</name></mean>
  <q1>Q1: %{y}</q1>
  <q3>q3: %{y}</q3>
`
```

would show three labels: one for the mean, q1 and q3 when hovering on a box. The mean hover label would have the trace "name" attach to it in a secondary label and its text would drawn in blue.

Now, those `<-->` will conflict with other pseudo-html tags we support (e.g. `<b>` as `b` is an attribute for `scatterternary` traces), so we should use some other escape sequence. But all in all, this could be a nice way to support all the per-label customizations our users can think of.

---

Let me write down a few less-important comments here about the implementation:

- What happens when template keys don't exist e.g. `show this: %{cow}`. How should this render out? e.g. `show this: undefined` or `show this: ` or something else.
- As `hovertemplate` overrides `hoverinfo`, we should not coerce `hoverinfo` if one provides a `hovertemplate`.
- What should `%{y}` in a template w/o `:` format mean. Does it mean: print the `y` event data value using the default plotly formatting or print the raw `y` value? 
- At some point, we should implement `hoverinfo` as a `arrayOk` attribute to allow user to set different templates for each point.
- To close out https://github.com/plotly/plotly.js/issues/3145, maybe we should add `layout.xaxis.hovertemplate` and `layout.yaxis.hovertemplate`

nicolaskruchten:
I'm curious to talk about the interaction between `hoverinfo` and `hovertemplate`. I was thinking that `hovertemplate` could accept `%{hoverinfo}` key and then that could be the default value, so we could always use the two together.

alexcjohnson:
> * one part of a trace can generate multiple hover labels (e.g. 'boxes' hover) and

The box-like case (incl. violin, ohlc, candlestick) seems like the toughest nut to crack. There's also the issue of whether you make one or many labels, ie `hoverlabel.split` https://github.com/plotly/plotly.js/pull/2959. As mentioned this morning I think we should not enable `hovertemplate` for these types in this PR. I can see a variety of ways to go with the API but none of them would be affected by the API for the simple case of a single label for hovering on a single class of objects. Let's make a new issue to discuss this, including enabling/disabling individual labels, independent styling (not just of the font as in `<mean><span style="color:blue">` but also the `bgcolor` etc that wouldn't work with the mega-template idea).

> * traces with multiple "parts" each with their own hover labels (e.g. sankey, but also parcats).

Again as I mentioned earlier, I like the plan of describing each class of hover label within its own container. The difficult case for this currently is scatter `hoveron: 'fills'`, which currently seems to be broken for the simple case of just one trace anyway. In the short term we should just fix this to ignore `hoverinfo` and always display the trace `name` since that's the only thing it can display, longer term we should make a new container (`fills`?) and put its config in there, like we do for sankey.

etpinard:
>  In the short term we should just fix this to ignore hoverinfo and always display the trace name since that's the only thing it can display, 

:+1: 

> longer term we should make a new container (`fills`?)

Good plan. Related: https://github.com/plotly/plotly.js/issues/420#issuecomment-419299905

etpinard:
> which currently seems to be broken for the simple case of just one trace anyway

now in -> https://github.com/plotly/plotly.js/issues/3203

alexcjohnson:
> * What happens when template keys don't exist e.g. `show this: %{cow}`. How should this render out? e.g. `show this: undefined` or `show this: ` or something else.

Is there ever a case where this is not simply a mistake the user will want to know about and fix? If not, I'd say leave it `show this: %{cow}` in the output and maybe `Lib.warn` about it.

> * As `hovertemplate` overrides `hoverinfo`, we should not coerce `hoverinfo` if one provides a `hovertemplate`.

👍 I do think there's a possibility of having `hoverinfo` create the default `hovertemplate`, but it's fairly complicated and connected to a bunch of other things like `hovermode` and the existence of other traces. Probably ignore this for now but perhaps explore it later.

> * What should `%{y}` in a template w/o `:` format mean. Does it mean: print the `y` event data value using the default plotly formatting or print the raw `y` value?

I like the idea of this being the default plotly formatting. That way we don't need to augment the d3 format spec at all. I don't see a benefit to raw.

> * At some point, we should implement `hoverinfo` as a `arrayOk` attribute to allow user to set different templates for each point.

👍 

> * To close out #3145, maybe we should add `layout.xaxis.hovertemplate` and `layout.yaxis.hovertemplate`

interesting... sure, perhaps in conjunction with the effort to have `hoverinfo` create `hovertemplate`?

etpinard:
Great - thanks @alexcjohnson I think all my concerns have been resolved (at least for this first iteration).

@antoinerg should be all set now :tada:

antoinerg:
- [ ] Return template key when it does not exist and fire a `Lib.warn`
- [ ] Return default formatted value when no format is specified
- [ ] Do not coerce hoverinfo if one provides a hovertemplate
- [ ] For scatter fills, always display the trace name
- [ ] Add a special tag to control the secondary box content
- [ ] Suport for pie
- [ ] Support for bar

etpinard:
To add some additional info about `Fx.hover` vs `Fx.loneHover`:

- `Fx.hover` works per subplot, it loops over all the traces that subplot to (1) find the closest(s) pts, (2) draw labels and (3) emit hover events
- `Fx.loneHover` does not try to find the closests pts, it assumes that's already taken care of. For example, for `pie` traces the "closest" pt is always the one you're hovering on, so listening to DOM `mouseover` events suffices. Note that `loneHover` was originally added in to add hover labels to gl3d subplots, where the "picking" logic is done in the shaders.

src/components/fx/hover.js
```diff
@@ -950,6 +963,21 @@ function createHoverText(hoverData, opts, gd) {
...
 
+        // hovertemplate
+        var trace = d.trace;
+        var hovertemplate = d.hovertemplate || false;
+        var hovertemplateLabels = d.hovertemplateLabels || d;
+        var eventData = d.eventData[0] || {};
+        if(hovertemplate) {
+            text = Lib.hovertemplateString(hovertemplate, hovertemplateLabels, eventData, trace);
+
+            var EXTRA_STRING_REGEX = /<extra>([\s\S]*)<\/extra>/;
```
etpinard: :cow2: Most browsers probably have a way to optimise this, but we usually define regexes in the outer module scope so we only have to compile them once on script load.

antoinerg: @etpinard I was also thinking I could move this whole block into a function of its own.

src/lib/index.js
```diff
@@ -1008,6 +1008,60 @@ lib.templateString = function(string, obj) {
...
+ *  Lib.templateString('name: %{trace[0].name}', {trace: [{name: 'asdf'}]}) --> 'name: asdf'
+ *  Lib.templateString('price: %{y:$.2f}', {y: 1}) --> 'price: $1.00'
+ *
+ * @param {string}  input string containing %{...:...} template strings
+ * @param {obj}     data object containing fallback text when no formatting is specified, ex.: {yLabel: 'formattedYValue'}
+ * @param {obj}     data objects containing substitution values
+ *
+ * @return {string} templated string
+ */
+
```
etpinard: another :cow2: (minor coding stylez) valid jsDocs start with `/**` nor `/*` like plain multi-line comments and usually there's no newline between the trainling `*/` and the start of the `function` .

src/lib/index.js
```diff
@@ -1008,6 +1008,60 @@ lib.templateString = function(string, obj) {
...
+                // getterCache[key] = getterCache[key] || lib.nestedProperty(obj, key).get;
+                // value = getterCache[key]();
+                value = getterCache[key] || lib.nestedProperty(obj, key).get();
+                if(value) getterCache[key] = value;
+            }
+            if(value !== undefined) break;
+        }
+
+        if(value === undefined) {
+            lib.warn('Variable \'' + key + '\' in hovertemplate could not be found!');
```
etpinard: We should probably add a limit to how many of these `lib.warn` get printed in the console. As these happen during mousemove, we wouldn't want to flood the console with identical warning messages.

src/lib/index.js
```diff
@@ -1008,6 +1008,60 @@ lib.templateString = function(string, obj) {
...
+            value = match;
+        }
+
+        if(format) {
+            value = d3.format(format.replace(/^:/, ''))(value);
+        } else {
+            if(labels.hasOwnProperty(key + 'Label')) value = labels[key + 'Label'];
```
etpinard: Very interesting. I've never been a big fan of `hasOwnProperty` (probably because it's a long word to type out), but **look at this**:

image

image


using http://jsbench.github.io/#1aebf699d73fb743127903c0b0a8bece

... so maybe we should start switching to `hasOwnProperty` in hot codepath :fire:

src/lib/index.js
```diff
@@ -1008,6 +1008,60 @@ lib.templateString = function(string, obj) {
...
+            if(value !== undefined) break;
+        }
+
+        if(value === undefined) {
+            lib.warn('Variable \'' + key + '\' in hovertemplate could not be found!');
+            value = match;
+        }
+
+        if(format) {
+            value = d3.format(format.replace(/^:/, ''))(value);
```
etpinard: Similarly here, let's bring this regex up in the outer scope.

src/plots/attributes.js
```diff
@@ -129,6 +129,19 @@ module.exports = {
         ].join(' ')
     },
     hoverlabel: fxAttrs.hoverlabel,
+    hovertemplate: {
```
etpinard: TODO: let's move this down to the trace `attributes.js` file.

By declaring `hovertemplate` per-trace, we'll be able to list of the (event data) keys available for each trace.

src/plots/attributes.js
```diff
@@ -129,6 +129,19 @@ module.exports = {
...
+        valType: 'string',
+        role: 'info',
+        dflt: false,
```
etpinard: We should make `dflt: ''` so that the default and the valid values are all `typeof string`. Luckily in JS `Boolean('') //=> false` , so you won't have to update all the `if(hovertemplate)` statements.

src/traces/pie/event_data.js
```diff
@@ -24,6 +24,8 @@ module.exports = function eventData(pt, trace) {
         label: pt.label,
         color: pt.color,
         value: pt.v,
+        percent: pt.percent,
+        text: pt.text,
```
etpinard: Feature! Nice!

src/traces/scatter/defaults.js
```diff
@@ -38,6 +38,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         'lines+markers' : 'lines';
     coerce('text');
     coerce('hovertext');
+    coerce('hovertemplate');
```
etpinard: We shouldn't `coerce` this when `hoveron: 'fills'`

src/plots/attributes.js
```diff
@@ -129,6 +129,19 @@ module.exports = {
...
+        editType: 'none',
+        description: [
```
etpinard: It might be nice to add support for `arrayOk:true` `hovertemplate`, e.g.

```js
trace = {
  x: [1, 2, 3],
  y: [2, 1, 2],
  hovertemplate: ['first pt<br>%{y}', '%second pt<br>%{y}<extra>LOOK!</extra>', '']
}
```

etpinard: ... it be _should_ be fairly to do so:

- handle `hovertemplate` in https://github.com/plotly/plotly.js/blob/master/src/components/fx/calc.js
- add some logic for `calcdata[i][j].ht` (`ht` for `hovertemplate`) in `fx/hover.js`
- do something similar in `pie/plot.js` using `helpers.castOption()`

etpinard:
@antoinerg great work! This PR is looking strong :muscle: 

I made a few comments, but looks like you're very close. 

Thinking about this again, I'd like getting `arrayOk:true` support out in this first iteration. Let me know if you need help.

Test coverage is looking solid. I would like a few more `Lib.hovertemplateString` cases that combine `%{}`, `<extra>` and other pseudo-html things like `<b>` and `<br>` and `<span>` (see https://rreusser.github.io/plotly-mock-viewer/#pseudo_html for more examples).

src/plots/attributes.js
etpinard: Ha I see. You declared `hovertemplate` in `plots/attributes.js` but then overrode it in the trace modules. 

We should instead remove `hovertemplate` form `plots/attributes.js`, declare it in full in `scatter/attributes.js` and `extendFlat` its `description` in `bar`, `pie` and `histogram` attribute files.

src/lib/index.js
antoinerg: Would something like throttling be satisfactory? I am pondering what to do :thinking:

src/components/fx/hover.js
antoinerg: Fixed in 0659d20

src/lib/index.js
antoinerg: Fixed in 8f440b0

src/lib/index.js
```diff
@@ -1008,6 +1008,60 @@ lib.templateString = function(string, obj) {
...
+        for(i = 2; i < args.length; i++) {
+            obj = args[i];
+            if(obj.hasOwnProperty(key)) {
+                value = obj[key];
+                break;
+            }
+
+            if(!SIMPLE_PROPERTY_REGEX.test(key)) {
+                // getterCache[key] = getterCache[key] || lib.nestedProperty(obj, key).get;
+                // value = getterCache[key]();
```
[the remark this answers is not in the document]
antoinerg: Fixed in 41ab464

src/lib/index.js
antoinerg: Done in 37e40f9

src/plots/attributes.js
antoinerg: Fix in 8e8b75b

src/plots/attributes.js
antoinerg: I made a function similar to `plots/domain.js`as we discussed but I still need to extend their description at the trace-level. Let me know if it looks good in 8e8b75b

src/traces/scatter/defaults.js
antoinerg: Done in 2015c57

src/lib/index.js
etpinard: I would just define a variable e.g. `numberOfHoverTemplateWarnings` in the outer scope, increment it here, and skip this `lib.warn` call here if greater than e.g. `10`.

src/plots/attributes.js
etpinard: 8e8b75b is brilliant. I'm a big fan of putting this new little module in `component/fx/` :ok_hand:

src/lib/index.js
antoinerg: Should we reset `numberOfHoverTemplateWarnings` to 0 each time `hovertemplate` changes? We probably don't want this function to go quiet forever :thinking:

alexcjohnson: I wouldn't worry about bringing the messages back, but do let the user know what happened - ie after the last displayed message add another like
`Lib.warn('Too many hovertemplate warnings - additional warnings will be suppressed')`

etpinard: > Should we reset `numberOfHoverTemplateWarnings` to 0 each time `hovertemplate` changes?

That would be nice, but not necessary for now in my opinion. Feel free to try to implement it if you're up for a challenge though.

antoinerg: Done in 3068d7d

antoinerg: BTW, I went with Alex's suggestion for now

src/components/fx/hover.js
```diff
@@ -670,6 +670,10 @@ function _hover(gd, evt, subplot, noHoverEvent) {
     for(itemnum = 0; itemnum < hoverData.length; itemnum++) {
         var pt = hoverData[itemnum];
         var eventData = helpers.makeEventData(pt, pt.trace, pt.cd);
+
+        // Add axis information to eventData
+        eventData.xaxis = gd._fullLayout[pt.trace.xaxis + 'axis'];
+        eventData.yaxis = gd._fullLayout[pt.trace.yaxis + 'axis'];
```
antoinerg: Should I explicitly select the axis' attributes we care about and create a smaller object to include in `eventData`? As far as I know, only `title` was requested.

antoinerg:
For good measure, I added initial support for #2618 in f89baca by making the axes' title available in `hovertemplate`. I think it's a must have.

src/components/fx/hover.js
antoinerg: @nicolaskruchten If we decide to be more explicit and only require the attributes that are important, what else beside `(x|y)axis.title` would you like to have available in `hovertemplate`?

etpinard: Hmm aren't `xaxis` and `yaxis` already part of the event data:

image

?

antoinerg: You are absolutely right: it's already there :tada:! I removed the extra code in 8d95f05 and still passes the test.

test/jasmine/tests/pie_test.js
```diff
@@ -731,6 +731,63 @@ describe('pie hovering', function() {
...
+            })
+            .then(_hover)
+            .then(function() {
+                assertLabel(
+                    ['33.3%'].join('\n'),
+                    null,
+                    'hovertemplate %{percent}'
+                );
+
+                return Plotly.restyle(gd, 'hovertemplate', '%{label}<extra></extra>');
```
etpinard: Could we add one arrayOk `hovertemplate` test for pies?

src/traces/bar/attributes.js
```diff
@@ -59,6 +60,7 @@ module.exports = {
 
     text: scatterAttrs.text,
     hovertext: scatterAttrs.hovertext,
+    hovertemplate: hovertemplateAttrs(),
```
etpinard: We should append the `hovertemplate` descriptions with a list of flags (i.e. event data keys) available.

etpinard:
Getting close, once:

- https://github.com/plotly/plotly.js/pull/3126#discussion_r232826022
- https://github.com/plotly/plotly.js/pull/3126#discussion_r232826560

are :white_check_mark: , I believe we'll be in :dancer: land.

src/traces/bar/attributes.js
antoinerg: On a related note: I like that for some traces, like `pie`, we test that event data has the correct keys:
https://github.com/plotly/plotly.js/blob/ad937d20a83fb3d183916e715c1a3aac6cef5921/test/jasmine/tests/pie_test.js#L824-L856

We could potentially use the same array of field names for the test and for the description making sure we have one source of truth :thinking:. This improvement could be done in another PR.

test/jasmine/tests/pie_test.js
antoinerg: Absolutely! There it is 1e4bd33

src/traces/bar/attributes.js
etpinard: > We could potentially use the same array of field names for the test and for the description making sure we have one source of truth

Yeah, good call. We usually put things like that in "constants" files like this one:

https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/constants.js

> This improvement could be done in another PR.

That's up to you!

src/traces/scatter/attributes.js
```diff
@@ -205,7 +205,7 @@ module.exports = {
         ].join(' ')
     },
     hovertemplate: hovertemplateAttrs({}, {
-        keys: ['marker.size', 'marker.color']
+        keys: ['marker.size']
```
etpinard: Why is `marker.color` gone? In fact, all `arrayOk` attributes should be part of this list. They get added to the event data via 

https://github.com/plotly/plotly.js/blob/2ceb5c3e7981cc850727f7406f6677c4d74be044/src/components/fx/helpers.js#L161-L186

Now, maybe we don't need to hardcode all the `arrayOk` attributes into `keys` here, maybe we could just mention that all "per-point" values are available for `hovertemplate`.

antoinerg: `marker.color` is gone because it is not emitted in the event data. It used to work because `hovertemplateString` would be called with the full `trace` object as one of its argument but I removed this in 43a4cd9 to make sure everything is on par with event data. Therefore, if we want `marker.color`, I need to add it to the event data for `scatter`.

etpinard: > `marker.color` is gone because it is not emitted in the event data

Try

```js
Plotly.newPlot(gd, [{
  y: [1, 2, 3],
  marker: {color: [1, 2, 3]}
}])
gd.on('plotly_hover', console.log)
```

antoinerg: It works if it's an array but not if it isn't? :thinking: That's why the test was failing: the mock I was using doesn't specify `marker.color` as an array.
```
Plotly.newPlot(gd, [{
  y: [1, 2, 3],
  marker: {color: 'blue'}
}])
gd.on('plotly_hover', console.log)
```

etpinard: > t works if it's an array but not if it isn't?

yes, that's the expected behaviour.

antoinerg: Ok so I added it and now run the test with a mock that specifies marker size and color as arrays. And it now passes.

Now for extra points, I just need to add those extra event data keys in `constants.js`.

antoinerg: Done in 583eb97

etpinard: From that same https://github.com/plotly/plotly.js/pull/3126#discussion_r233487700

> Now, maybe we don't need to hardcode all the `arrayOk` attributes into `keys` here,

I think it might be best to **NOT** hardcoded all `arrayOk` attributes into that `keys` list, as it will be hard to remember to append it everytime we add another `arrayOk` attribute. We should keep that `keys` list only for keys that have no corresponding schema attributes e.g `binNumber` in histogram traces, **and** we should mention that all `arrayOk` (aka per-point) attribute are available to `hovertemplate` in the `components/fx/hovertemplate_attributes'` description.

Sorry for not been clearer.

etpinard:
Beautifully done. :dancer:

chriddyp:
Nice example of this new `hovertemplate` string (courtesy @nicolaskruchten )
```
hovertemplate:
       "<b>%{text}</b><br><br>" +
       "%{yaxis.title}: %{y:$,.0f}<br>" +
       "%{xaxis.title}: %{x:.0%}<br>" +
       "Number Employed: %{marker.size:,}"
```
image

Braintelligence:
Maybe stupid question: Is it possible to style the hovertemplate more? Like having a table structure?

etpinard:
> s it possible to style the hovertemplate more? Like having a table structure?

No. That'll be in https://github.com/plotly/plotly.js/issues/3010

CZechBoY:
Does this work with scatter3D?

Issue #2951, opened 2018-08-28
pie insidetext colour by default isn't readable


dsmmcken:
If a user creates a pie chart using all default settings and the text labels default to inside the slice, then some labels become unreadable depending on slice colour (below WCAG guidelines). For inside labels, colour should auto set to contrasting. A plot produced with all default settings should always be usable.

![screen shot 2018-08-28 at 6 10 12 pm](https://user-images.githubusercontent.com/1576283/44754270-2df90400-aaef-11e8-9727-9b190421af9b.png)

[Link to codepen](https://codepen.io/anon/pen/bxwGqz)

Furthermore, if a user sets a custom marker colours array, there is no way for them to set the text colour to contrast or set text colours individually.

alexcjohnson:
Thanks @dsmmcken - this issue also pertains to bars:
![screen shot 2018-08-28 at 6 46 52 pm](https://user-images.githubusercontent.com/2678795/44755142-da88b500-aaf2-11e8-990f-e2232e1e448d.png)
We should:
- make the default flip from black to white per slice the same way hover labels do
- allow an array for font properties - was left as a TODO and never followed up on:
https://github.com/plotly/plotly.js/blob/bff736d1efdbb471e212374c50c12044ccedcad5/src/traces/pie/attributes.js#L172

Bars already allow array font properties:
![screen shot 2018-08-28 at 6 56 27 pm](https://user-images.githubusercontent.com/2678795/44755426-2556fc80-aaf4-11e8-87e6-99082e983c25.png)
But there's a bug (see trace 0 above, the lower set of positive bars) where `insidetextfont` is ignored when you explicitly set `textposition: 'outside'` but the text gets drawn inside anyway because another set of bars got stacked on top.
https://codepen.io/alexcjohnson/pen/aamzMP?editors=1010

rmoestl:
## Tasks
To summarize, we need to implement **two new features** and **fix one bug**:
- Choose a contrasting text color for inside labels by default
- Allow to set font properties for inside labels on a per slice basis
- Fix bar (and potentially pie) to use `insidetextfont` if the bar label is drawn inside the bar (to avoid overlapping) even if `textposition` is set to `outside`

## API changes
From an **API** point of view, I believe we need to address how font properties are inherited from `layout.font` all the way down to `pie.insidetextfont` and `bar.insidetextfont`. With the new default behavior (contrasting), `layout.font.color` can no longer be inherited by `pie.insidetextfont.color`.

[...]

PR #3130, opened 2018-10-19
2951 contrasting pie and bar inside text

rmoestl:
Improves a few things around text font styling for **pie** and **bar**. Specified and discussed in #2951. 

Rough overview:

- If no custom font color is specified in trace, an inside text label will render with a color contrasting the bar / pie fill color. This improves out-of-the-box readability.
- Adds support for array (`arrayOk`) font properties in pie.
- Bar `textposition: 'outside'` labels pushed `inside` by a bar being stacked on top of it are rendered with inside text font properties now (bugfix).
- Font properties are now applied correctly throughout a selection interaction as well (bugfix).


rmoestl:
> Very nice improvements for the legibility of the figures! tada

Thanks Antoine! :bowing_man:

alexcjohnson:
in [bar_attrs_relative](https://github.com/plotly/plotly.js/pull/3130/files?short_path=f2db335#diff-055dc9f6b5ae185a53b5bf0a29766fbc) why does text inside the orange bars end up white?

screen shot 2018-10-22 at 1 30 26 pm

Elsewhere (like hover labels, and pie labels elsewhere in this PR) that orange color keeps black text as a contrast.

screen shot 2018-10-22 at 1 28 30 pm

BTW very nice that the text that got pushed out of the small red bar turned to black even though the text *inside* the red bars is white 🎉

rmoestl:
> in bar_attrs_relative why does text inside the orange bars end up white?

The problem was I had no test to verify that contrasting text colors are correct if no explicit marker colors are set. I mistakenly assumed `calcdata[i].mc` is always set just like it is the case for `pie` where `calcdata[i].color` has always a value. I've added a test to pie tough to cover the "no marker colors set" case.

alexcjohnson:
@rmoestl looks fantastic!

@etpinard I have one question about how this will play with RCE (cc @nicolaskruchten @VeraZab) - which currently doesn't seem to support `(inside|outside)textfont` at all but presumably at some point it will. I'm assuming because `insidetextfont.color` is [`undefined`](https://github.com/plotly/plotly.js/pull/3130/files?short_path=5a6cbbd#diff-53de192dfcd2487ecdc3dec4d837ede6R165) by default, we wouldn't show the corresponding color picker - so perhaps we should accept and fill in something like `'auto'` instead? (see https://github.com/plotly/plotly.js/pull/3109#discussion_r227543929 for discussion of `null` and other options)

nicolaskruchten:
We can certainly add widgets for these attributes to RCE.

In general, this kind of layered automatic computation is a bit tricky for a UI... we've resorted to this kind of thing in the figure-wide hover label spec:

image

alexcjohnson:
OK great - then lets keep this as is -> 💃 

I still have [concerns about how stuff like this will interact with templates](https://github.com/plotly/plotly.js/pull/3109#discussion_r227543929) but those should be addressed in a broader context.

PR #3131, opened 2018-10-20
Automatic alignment of ticks in 3D scenes

archmoj:
This PR improves automatic alignment of ticks on 3D axes. The best single orientation for each axis (depending on the camera position and orientation) is found and used by vertex shader.
In the [earlier work](https://github.com/plotly/plotly.js/pull/3084) it was possible for the ticks on an axis to align themselves in two different directions and stay upwards. But such behavior was not completely desirable in some perspective views. 
@alexcjohnson


package.json
```diff
@@ -78,7 +78,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.0.0",
     "gl-plot2d": "^1.3.1",
-    "gl-plot3d": "^1.5.7",
+    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#240870c6cbf602b956363b86287e516a45969ea5",
```
archmoj: https://github.com/gl-vis/gl-plot3d/compare/master...align-ticks

archmoj: https://github.com/gl-vis/gl-axes3d/compare/master...align-ticks

alexcjohnson:
@archmoj interesting - The code looks great and it does exactly what we discussed, can't argue with any of that! But now that I see it and play with it, I'm actually not convinced that this is better than having each label independently restricted to [-90,90] degrees rotation as implemented in https://github.com/plotly/plotly.js/pull/3084. I kind of think the improved consistency from label to label is not worth the reduced legibility of individual labels. What do you think?

@etpinard @antoinerg thoughts?

antoinerg:
> such behavior was not completely desirable in some perspective views.
Hello @archmoj, can you point me to some examples of those perspective views where the results are undesirable.

> I kind of think the improved consistency from label to label is not worth the reduced legibility of individual labels.
@alexcjohnson I do agree that the labels oriented with an angle outside [-90, 90] look weird to me and are probably a bit harder to read. On the other hand, I am not sure I fully appreciate what we're gaining in consistency. I need to see an example and test it out more.

archmoj:
> > such behavior was not completely desirable in some perspective views.
> > can you point me to some examples of those perspective views where the results are undesirable.
@antoinerg These [old and new](https://github.com/plotly/plotly.js/pull/3131/files?short_path=f23102e#diff-f23102edc6472edaa010688cb311a8ab) files may be easy to compare using `Swipe` option.

archmoj:
> What do you think?

We could setup the shader to be able to serve both functionalities (as well as others e.g. option:2 for horizontal/vertical, option:8 for 8 directions ...). For example I could simply allocate option:0 (now used for debugging) for the ranges between [-90,90]. Then the question would be which option 0/1 should be the default behavior. I personally think that when either the ticks include some text or the user need to print the graph it would be nice to give them options to switch to this new behavior.

archmoj:
@antoinerg
Here is an example of the previous behavior: 
created by @alexcjohnson before.

label flip

alexcjohnson:
> This is maybe not the right place to bring this up but why aren't we making the label facing us and always horizontal?

Absolutely, a reasonable place to bring it up. I can certainly see cases where horizontal is better, and a lot of simple situations fall into this camp. But there are also cases - particularly when there are lots of ticks (so it can be ambiguous which label goes with which tick) or long labels (so they can overlap each other) - where to my eye it's much better to rotate the labels. So if we only have one behavior I think it should be this one (specifically restricting each label independently to [-90, 90] degrees) but I'd be happy for us to enable fixed angle (normally fixed at 0) and consistently rotated angles in a subsequent PR.

alexcjohnson:
OK great - I'm glad this now leaves all the mocks as they were before the PR. Since this doesn't have any immediate effect anymore, let's wait and see if @etpinard's opinion changes the result at all.

alexcjohnson:
Oh actually, I forgot, this PR includes the Chrome 51 title rotation fix, right? In that case, lets go ahead and merge the upstream PRs and this, and we can discuss whether to change the default behavior and/or add another config parameter in a separate PR. 💃

Issue #2999, opened 2018-09-11
Scattergl does not always render markers


rhamner:
Occasionally when trying to render markers with scattergl, I get the following:

Offscreen-For-WebGL-000000C6A5E1D5E0]GL ERROR :GL_INVALID_OPERATION : glDrawElements: attempt to access out of range vertices in attribute 0

When this happens, most of the markers are not rendered. Hover and click still work for them though. An example of hovering over a point that doesn't render is included in the screenshot below:

![Image of fake hover](http://i65.tinypic.com/bfj9k6.png)

That plot looks like the following when all points are rendered:

![image](https://user-images.githubusercontent.com/20448486/45388752-85af6900-b5df-11e8-85ec-a9606079f42e.png)

If I switch to 'scatter' or 'pointcloud' or use lines instead of markers, it works fine (lines example):

![image](https://user-images.githubusercontent.com/20448486/45388678-497c0880-b5df-11e8-97de-81ffbe6c1cd0.png)

Stranger, I often get this error with a subset of the data but not with the full data set. The full data set looks like this using markers and scattergl:

![image](https://user-images.githubusercontent.com/20448486/45388724-6b758b00-b5df-11e8-9b4a-9a09609d97a2.png)

When I switch to 'edit in chart studio' the problem goes away. When I dump the plot data to json then bring into codepen, the problem goes away (https://codepen.io/rhamner/pen/zJWREL). It is probably important to know that this is the same machine...the issue appears with the same data set on the same machine sometimes.

Any idea on what could be causing this? Unfortunately, I can't get the issue to reproduce with the exact same data in codepen so I'm hoping the warning/error, screenshots, and actual data set in that codepen link are sufficient.

I saw a maybe similar issue reported here:

https://github.com/plotly/dash-core-components/issues/204

etpinard:
What version of plotly.js are you using?

rhamner:
1.40.1. I have tried using https://cdn.plot.ly/plotly-latest.min.js also.

etpinard:
Ok, and are you using dash?

rhamner:
I am not. Just straight javascript/plotly.js

etpinard:
> I am not. Just straight javascript/plotly.js

Can you share a snippet of your work then to help us out?

etpinard:
... or is the JS code in your https://codepen.io/rhamner/pen/zJWREL equivalent to the one that yield those 

[...]

PR #3098, opened 2018-10-11
Scatterpolargl fixes

etpinard:
fixes #3094 and brings `scatterpolargl` axis expansion on-par with `scatterpolar`.

cc @plotly/plotly_js

PR #3132, opened 2018-10-21
Fix to enable switching modes using plotly react method on scattergl traces

archmoj:
This PR resolves the issue #2999 by providing additional checks before calling destroy functions of scattergl. Without these checks in scenarios namely switching back and forth between different modes e.g. `markers` to `lines` and  then back to `markers` the program (depending on the device/browser) may complain and/or stop.
@alexcjohnson

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -296,25 +296,24 @@ function sceneUpdate(gd, subplot) {
         // remove scene resources
         scene.destroy = function destroy() {
             if((scene.fill2d) &&
-               (scene.fill2d.destroy))
-                scene.fill2d.destroy();
+               (scene.fill2d.destroy)) {
+                scene.fill2d.destroy();}
             if((scene.scatter2d) &&
-               (scene.scatter2d.destroy))
-                scene.scatter2d.destroy();
+               (scene.scatter2d.destroy)) {
+                scene.scatter2d.destroy();}
             if((scene.error2d) &&
-               (scene.error2d.destroy))
-                scene.error2d.destroy();
+               (scene.error2d.destroy)) {
+                scene.error2d.destroy();}
             if((scene.line2d) &&
-               (scene.line2d.destroy))
-                scene.line2d.destroy();
+               (scene.line2d.destroy)) {
+                scene.line2d.destroy();}
             if((scene.select2d) &&
-               (scene.select2d.destroy))
-                scene.select2d.destroy();
+               (scene.select2d.destroy)) {
+                scene.select2d.destroy();}
             if(scene.glText) {
                 scene.glText.forEach(
                     function(text) {
-                        if(text.destroy)
-                            text.destroy();
+                        if(text.destroy) text.destroy();
                     }
                 );
             }
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -295,27 +295,15 @@ function sceneUpdate(gd, subplot) {
 
         // remove scene resources
         scene.destroy = function destroy() {
-            if((scene.fill2d) &&
-               (scene.fill2d.destroy)) {
-                scene.fill2d.destroy();}
-            if((scene.scatter2d) &&
-               (scene.scatter2d.destroy)) {
-                scene.scatter2d.destroy();}
-            if((scene.error2d) &&
-               (scene.error2d.destroy)) {
-                scene.error2d.destroy();}
-            if((scene.line2d) &&
-               (scene.line2d.destroy)) {
-                scene.line2d.destroy();}
-            if((scene.select2d) &&
-               (scene.select2d.destroy)) {
-                scene.select2d.destroy();}
+            if(scene.fill2d && scene.fill2d.destroy) scene.fill2d.destroy();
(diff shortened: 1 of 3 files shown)


etpinard:
> Without these checks in scenarios namely switching back and forth between different modes e.g. `markers` to `lines` and then back to `markers` the program (depending on the device/browser) may complain and/or stop.

Were you able to replicate the problem. I'm no opposed to this solution, but I'd like to understand why this happens.

archmoj:
@etpinard Could you please try [this demo](https://codepen.io/MojtabaSamimi/pen/vVroZV/?editors=1111)? 
The graph may stop updates after few switch modes.

etpinard:
Hmm. I can't replicate. 

Do you have an idea _why_ this happens? It smells like a race condition, but I'd like to confirm.

archmoj:
We used to have similar checks before calling destroy in traces/splom/index.js
It could be a race condition. Or the the object key is not completely deleted.
After this fix I still get some warnings on FireFox after revision 6:
`WebGL warning: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one.`

etpinard:
> We used to have similar checks before calling destroy in traces/splom/index.js

Hmm I see. Referencing:

https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/traces/splom/index.js#L155-L157

I put those lines in. I think they were intended to fix intermittent failures on CI.

Note that removing that `&& scene.matrix.destroy` does not make any test fail locally on my machine. Maybe someone could try to double-check?

----------

All in all, I'll like to get to the bottom of this issue. It doesn't have to be now though. A lot of users have commented on https://github.com/plotly/plotly.js/issues/2999, so maybe we should try to merge this fix in before 1.42.0 and take a deeper dive later.

src/traces/scattergl/index.js
```diff
@@ -295,13 +295,27 @@ function sceneUpdate(gd, subplot) {
 
         // remove scene resources
         scene.destroy = function destroy() {
-            if(scene.fill2d) scene.fill2d.destroy();
-            if(scene.scatter2d) scene.scatter2d.destroy();
-            if(scene.error2d) scene.error2d.destroy();
-            if(scene.line2d) scene.line2d.destroy();
-            if(scene.select2d) scene.select2d.destroy();
+            if((scene.fill2d) &&
```
etpinard: Non-blocking comment: We (me and @alexcjohnson ) would usually write these lines as

```js
if(scene.fill2d && scene.fill2d.destroy) {
    scene.fill2d.destroy();
}

File: src/traces/scattergl/index.js [hunk 2]
```diff
@@ -295,13 +295,27 @@ function sceneUpdate(gd, subplot) {
...
+                scene.error2d.destroy();}
+            if((scene.line2d) &&
+               (scene.line2d.destroy)) {
+                scene.line2d.destroy();}
+            if((scene.select2d) &&
+               (scene.select2d.destroy)) {
+                scene.select2d.destroy();}
             if(scene.glText) {
-                scene.glText.forEach(function(text) { text.destroy(); });
+                scene.glText.forEach(
```
etpinard: and this thing as

```js
scene.glText.forEach(function(text) {
    if(text.destroy) text.destroy();
});

etpinard:
Since I (and presumably other people) can't replicate the problem, it might be hard to write a jasmine test to lock this down.

archmoj:
Interesting observation: Using the code identical to the one presented in the codepen on Chrome the `scene.destroy` method is not called. But on FireFox `scene.destroy` is called. Could you point me to the where the destroy method is triggered?

etpinard:
> Could you point me to the where the destroy method is triggered?

here for scattergl -> 

https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/plots/cartesian/index.js#L315-L320

here for splom ->

https://github.com/plotly/plotly.js/blob/7ae6fd68533d6489e3161cdbdfe2a44217f530e1/src/traces/splom/base_plot.js#L177-L186


and similar logic will be added for scatterpolargl in https://github.com/plotly/plotly.js/pull/3098

archmoj:
This should fix the issue for switching `modes`. 
But we may need to open a new PR for the issue of switching line `types` where on FF we could get warning messages like this one after few update calls `WebGL warning: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one.`

etpinard:
Thanks for the update @archmoj !

It would be nice to investigate (and hopefully fix :smile: ) that FF too-many-webgl-context bug this afternoon. I can help out.

archmoj:
Thanks @etpinard for the help on this.
I think we could merge this branch.
And then regarding the warning messages we create another branch & PR to resolve this [PHX issue](https://github.com/plotly/phoenix-integration/issues/207).
What do you think?

etpinard:
> I think we could merge this branch.

We'll need to add a test before merging. 

> And then regarding the warning messages we create another branch

Creating another branch is fine :ok_hand:

etpinard:
I can help you add that test if you like.

archmoj:
Great! Any help on tests would be much appreciated.

Issue #704, opened 2016-06-29
scatter3d: suppressing connecting lines causes WebGL warning


scottwittenburg:
I am using plotly.js version 1.12.0.  In my browser console, I do:

``` javascript
container = document.querySelector('.plotly-container')
Plotly.newPlot(container, [{x: [1, 2, 3], y: [3, 2, 1], z: [2, 3, 1], type: 'scatter3d'}])
```

and everything is fine.  If I don't want the lines between the points, I do:

``` javascript
Plotly.newPlot(container, [{x: [1, 2, 3], y: [3, 2, 1], z: [2, 3, 1], type: 'scatter3d', mode:'markers+text'}])
```

and then, while I no longer see the lines between the points, I get many of this warning:

```
[.CommandBufferContext.Offscreen-MainThread-0x3969ca4b4a00]RENDER WARNING: Render count or primcount is 0.
```

I'm using Ubuntu 14.04 and the above warnings only seem to appear in Chrome (for me, Version 51.0.2704.106 (64-bit)).  When in Firefox, I do not see the warnings.

Thank you!

etpinard:
Thanks for report. I can confirm this issue on Chrome 51 and Ubuntu 16.04. 

The plot still renders fine to my eyes. @scottwittenburg can you confirm?

scottwittenburg:
That is correct.  The plot renders fine, and the only issue is the console warnings.

etpinard:
cc @dfcreative

etpinard:
Oh actually, I was able to replicate the same warning message with:

```js
Plotly.newPlot(gd, [{
  type: 'scatter3d',
  mode: 'text',
  x: [1,2,3],
  y: [2,1,2],
  z: [2,1,3]
}])
```

I'm thinking this has to do with setting `mode: 'text'` when `text` is empty.

etpinard:
Fixed in https://github.com/plotly/plotly.js/pull/3171

Issue #1976, opened 2017-08-31
`gl3d_multi_scene` mock is throwing WebGL rendering warnings


etpinard:
See http://rickyreusser.com/plotly-mock-viewer/#gl3d_multi-scene

![image](https://user-images.githubusercontent.com/6675409/29903341-608fb80a-8dd1-11e7-808a-1e0e2c0915f5.png)

in Chrome `60.0.3112.113` on Ubuntu 16.04.

rreusser:
Same on OSX + Chrome.

etpinard:
cc @archmoj

etpinard:
updated link: https://rreusser.github.io/plotly-mock-viewer/#gl3d_multi-scene

archmoj:
Good call @etpinard! 
This one is definitely related to gl-scatter3d and I believe it happens when the line mode is enabled and has only one point which produce 0 lines! So I need to fix gl-scatter3d there.
But my observation here is that one could use `mode:'markers'` to remove the warnings on this mock.
Also I observed that if we use `mode:'text'` instead we get similar warnings on master branch. But using the fix for issue #704 those warnings would be removed.

PR #3133, opened 2018-10-22
Fix to remove webgl warnings resulted from blank texts in scatter3d

archmoj:
This PR resolves issue #704.
@alexcjohnson


package.json
```diff
@@ -80,7 +80,7 @@
     "gl-plot2d": "^1.3.1",
     "gl-plot3d": "^1.5.7",
     "gl-pointcloud2d": "^1.0.1",
-    "gl-scatter3d": "^1.0.11",
+    "gl-scatter3d": "git://github.com/gl-vis/gl-scatter3d.git#2ee9fbb3c29af694ad64258a27dc04d617896e84",
```
archmoj: https://github.com/gl-vis/gl-scatter3d/compare/master...issue-704

alexcjohnson: BTW now that you have PRs in these other repos, even better than the branch link would be to link to that PR directly
https://github.com/gl-vis/gl-scatter3d/pull/9

alexcjohnson:
Are the warnings that this PR suppresses accessible to us in a jasmine test? Or is there some other effect of the error that we *can* test?

package.json
archmoj: Thanks for the note. But the other PR on the gl-scatter3d repo is not opened by me.

etpinard:
> Are the warnings that this PR suppresses accessible to us in a jasmine test?

Unfortunately no.

> Or is there some other effect of the error that we can test?

`gd._fullLayout.scene._scene.glplot.objects[0].glyphBuffer.length` was 0 before and is `2616` after plotting:

```js
Plotly.newPlot(gd, [{
  type: 'scatter3d',
  mode: 'text',
  x: [1,2,3],
  y: [2,1,2],
  z: [2,1,3]
}])
```

@archmoj is this expected? Could we use it to :lock: down this fix?

Also referencing https://github.com/gl-vis/gl-scatter3d/pull/9/files#r227874673

archmoj:
Thanks @etpinard.
Yes this is really helpful.

etpinard:
- append an `it` to this [describe block](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/gl3d_plot_interact_test.js#L24)
- call

```js
Plotly.plot(gd, [{
  type: 'scatter3d',
  mode: 'text',
  x: [1,2,3],
  y: [2,1,2],
  z: [2,1,3]
}])
```

- assert that `gd._fullLayout.scene._scene.glplot.objects[0].glyphBuffer.length` isn't 0

- add a comment above that asserting linking to issue https://github.com/plotly/plotly.js/issues/704

package.json
archmoj: My apologies @alexcjohnson, 
Here is the link to the open PR:
https://github.com/gl-vis/gl-scatter3d/pull/9

archmoj:
The jasmine test is now able to catch the bug. Many thanks @etpinard for the helpful introduction to this test.
Next step: I need to finalize the changes to gl-scatter3d module, and I may try to fix this issue as well.
https://github.com/plotly/plotly.js/issues/1976

archmoj:
Regarding this issue and other issues I noticed in gl-scatter3d module please refer to [this demo](https://codepen.io/MojtabaSamimi/pen/OBrdyE).
To test improvements and new functionalities (e.g. handling JSON data in the text) with the new version please run the JavaScript code (excluding its first line) directly in the console.

etpinard:
Nice demo @archmoj 

Have you check that those non-string items in `text` behave the same way in for `scatter` traces:

https://codepen.io/etpinard/pen/qJgPvX ?

archmoj:
Thank you @etpinard for this demonstration. Good call if needed I could also fix the other trace in order to e.g. display JSON object properly?

etpinard:
> I could also fix the other trace in order to e.g. display JSON object properly?

No need. I don't think we should support that. We could do better than rendering `[object object]` but that'll be a another day.

archmoj:
So now on scatter3d we render `undefined` as empty text. Sounds OK? @etpinard
Using your [demo for scatter](https://codepen.io/etpinard/pen/qJgPvX ) one can see they are rendered as blank which is similar to the new scatter3d behavior. 
Please note that we used to render them as `undefined` which could have been a bit long & annoying.
Any other suggestion? @alexcjohnson

alexcjohnson:
I'd call it a bug if `undefined` (or `null`) *ever* gets drawn as something other than a missing / blank value. No need to hunt around and fix other trace types here, but it's definitely a positive change for `scatter3d`.

etpinard:
> Sounds OK?

Looks good! Could you add JSON version of your https://codepen.io/MojtabaSamimi/pen/OBrdyE to our test mocks?

archmoj:
Great idea @etpinard . 
Yes I'll do it right now.

archmoj:
There is a question. JSON format only has `null` and does not have `undefined`. So how we could/should handle cases where the user used `undefined` and of course not "undefined" (as string) there?
Should we those cases to `null` or `""` (blank) or  `"undefined"` (string)? I prefer not to use `"undefined"` as string in this scenario. Your thoughts? @etpinard @alexcjohnson

etpinard:
> There is a question. JSON format only has `null` and does not have `undefined`. So how we could/should handle cases where the user used `undefined` and of course not "undefined" (as string) there?
> Should we those cases to `null` or `""` (blank) or `"undefined"` (string)? I prefer not to use `"undefined"` as string in this scenario.

`null`, `undefined`, `''` should all result in a blank text entry.

archmoj:
Yes, but it is not clear to me how one could mimic the case of `undefined` using JSON input? 
One idea could be to write nothing between two commas. Right?

etpinard:
Pro tip @archmoj you can go ahead and close "duplicate" PRs. If something goes wrong, we can always reopen closed PRs.

archmoj:
Thanks for the note @etpinard. OK I would follow that next time.

PR #3136, opened 2018-10-22
add build artifact to CI build

antoinerg:
Testing the artifacts features of circleCI. 

The idea is to store the `plotly.js` that gets built for each commit/branch. This way, it would be extremely easy to test a particular branch without having to checkout, build and serve over HTTP locally. Moreover, if the artifacts are served by circleCI over HTTPS, it will be easy to test them out in Codepen and quickly swap between versions to check whether it correctly fixes a given issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,9 @@ jobs:
           command: |
             npm run pretest
             npm run cibuild
+      - run:
+          name: Save build artifact
+          command: cp build/plotly.js $CIRCLE_ARTIFACTS/
       - save_cache:
           paths:
             - node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,8 @@ jobs:
           command: ./.circleci/test.sh syntax
 
   publish:
+    docker:
+      - image: circleci/node:10.9.0    
     working_directory: ~/plotly.js
     steps:
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,9 +35,6 @@ jobs:
             npm run cibuild
       - store_artifacts:
           path: ~/plotly.js/build/plotly.js
-      - run:
-          name: Save build artifact
-          command: cp build/plotly.js $CIRCLE_ARTIFACTS/
       - save_cache:
           paths:
             - node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,8 @@ jobs:
             npm run pretest
             npm run cibuild
       - store_artifacts:
-          path: ~/plotly.js/build/plotly.js
+          path: build/plotly.js
+          destination: /plotly.js
       - save_cache:
           paths:
             - node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,7 @@ jobs:
           command: ./.circleci/test.sh image
       - store_artifacts:
           path: build
+          destination: /
 
   test-image2:
     docker:
@@ -107,6 +108,7 @@ jobs:
           command: ./.circleci/test.sh image2
       - store_artifacts:
           path: build
+          destination: /
 
   test-syntax:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,9 +122,12 @@ jobs:
 
   publish:
     docker:
-      - image: circleci/node:10.9.0    
+      - image: circleci/node:10.9.0
     working_directory: ~/plotly.js
     steps:
+      - checkout
+      - attach_workspace:
+          at: ~/plotly.js
       - store_artifacts:
           path: build/plotly.js
           destination: /plotly.js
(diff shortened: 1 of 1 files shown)


jonmmease:
This would be awesome from the plotly.py side.  I've been wanting to figure out a way to build plotly.py against plotly.js master, and this would help a lot.

Would the plot schema be available as an artifact as well?  (plotly.py uses this during the build process for code generation)

jonmmease:
In terms of artifacts, all we need is `plot-schema.json` and `plotly.min.js`.  For automation purposes, do you have any thoughts on how the plotly.py build process could find out what the latest plotly.js build is?

antoinerg:
It should be possible to automate using CircleCI API!

I think I should store the artifacts as the last build step. This way, we would get the URLs to the artifacts using [one API call which lists the artifacts of the latest build](https://circleci.com/docs/api/?shell#artifacts-of-the-latest-build):
```
$ curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js/latest/artifacts?branch=master\&filter=successful
...
[ {
  "path" : "plotly.js",
  "pretty_path" : "plotly.js",
  "node_index" : 0,
  "url" : "https://16312-45646037-gh.circle-artifacts.com/0/plotly.js"
} ]
```

antoinerg:
Ok so the following API call returns exactly what we want!
``` bash
$ BRANCH=circle-build-artifact
$ curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js/latest/artifacts?branch=$BRANCH\&filter=successful
[ {
  "path" : "plot-schema.json",
  "pretty_path" : "plot-schema.json",
  "node_index" : 0,
  "url" : "https://16381-45646037-gh.circle-artifacts.com/0/plot-schema.json"
}, {
  "path" : "plotly.js",
  "pretty_path" : "plotly.js",
  "node_index" : 0,
  "url" : "https://16381-45646037-gh.circle-artifacts.com/0/plotly.js"
}, {
  "path" : "plotly.min.js",
  "pretty_path" : "plotly.min.js",
  "node_index" : 0,
  "url" : "https://16381-45646037-gh.circle-artifacts.com/0/plotly.min.js"
} ]
```

I was wondering if I should simply add the whole `dist/` folder instead? cc @etpinard @alexcjohnson

alexcjohnson:
@antoinerg cool feature!

> I was wondering if I should simply add the whole `dist/` folder instead?

I don't see what that would gain us... and there's a lot in there, do we even update all of them during the test run? I'd leave it at just these items for now, as long as the naming matches `dist/` (which it does) we can always augment it later.

While you're in this code, does this same feature let us give the image test artifacts a simpler location? I would love it if we could get at these without diving so deep into the directory tree

screen shot 2018-10-22 at 8 57 16 pm

Can you change it so it just looks like:

screen shot 2018-10-22 at 8 57 16 pm

(I would also love it if there were a directory like `test_images_failed/` of just the images that produced diffs - in addition to the existing two directories. But that's a rather different issue)

antoinerg:
@alexcjohnson  With commit 82b640b, it now looks like this for both `test-image` and `test-image2`:
screenshot from 2018-10-22 21-34-53

alexcjohnson:
Looks great to me! 💃

antoinerg:
I love my new Linux laptop. Having a functional bash shell is awesome!

Here are one-liners to the URLs of `plotly.min.js` (whether they exist or not) for all the `publish` jobs:
```
curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js?limit=100\&filter=completed | \
    jq '.[] | select(.workflows.job_name=="publish") | .build_num' | \
    xargs -I {} echo https://{}-45646037-gh.circle-artifacts.com/0/plotly.min.js
```
Same thing but for a given branch:
```
BRANCH=circle-build-artifact
curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js/tree/$BRANCH?limit=100\&filter=completed | \
    jq '.[] | select(.workflows.job_name=="publish") | .build_num' | \
    xargs -I {} echo https://{}-45646037-gh.circle-artifacts.com/0/plotly.min.js
```

I can't wait to merge this in `master`!

antoinerg:
@jonmmease 

In https://github.com/plotly/plotly.js/pull/3136#issuecomment-432040200, I show how to get the list of artifacts of the latest job built (which may or may not be `publish` if the workflow is still running).

In https://github.com/plotly/plotly.js/pull/3136#issuecomment-432057809, I show how to retrieve the build number of the most recent `publish` jobs that are completed which we can use to guess the URL where the artifact will be sitting. To be safe, you should probably use that build number to make another [API call to get the artifacts associated with that build number](https://circleci.com/docs/api/#artifacts-of-a-build). In bash it would look like:
```
BRANCH=circle-build-artifact
curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js/tree/$BRANCH?limit=100\&filter=completed | \
   jq '.[] | select(.workflows.job_name=="publish") | .build_num' | \
   head -n 1 | \
   xargs -I {} curl https://circleci.com/api/v1.1/project/github/plotly/plotly.js/{}/artifacts
```

Good night and thank you all for your help :)

.circleci/config.yml
```diff
@@ -107,6 +108,7 @@ jobs:
           command: ./.circleci/test.sh image2
       - store_artifacts:
           path: build
+          destination: /
```
etpinard: Nice touch here :art:

PR #3137, opened 2018-10-22
prevent our loggers from throwing their own errors in a few more cases

alexcjohnson:
cc @antoinerg - a sort of followup to #3114

### DIFF
--- a/src/lib/loggers.js
+++ b/src/lib/loggers.js
@@ -61,12 +61,24 @@ loggers.error = function() {
  * apply like other functions do
  */
 function apply(f, args) {
-    if(f.apply) {
-        f.apply(f, args);
+    if(f && f.apply) {
+        try {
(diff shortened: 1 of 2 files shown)


src/lib/loggers.js
```diff
@@ -61,12 +61,24 @@ loggers.error = function() {
  * apply like other functions do
  */
 function apply(f, args) {
-    if(f.apply) {
-        f.apply(f, args);
+    if(f && f.apply) {
+        try {
+            // `this` should always be console, since here we're always
+            // applying a method of the console object.
+            f.apply(console, args);
```
alexcjohnson: In current browsers this works anyway, but `console` is really the right `this` here and some old browsers (Chrome 51 and presumably before) need it to be correct. See eg https://github.com/plotly/phoenix-integration/issues/223#issuecomment-431863801

PR #3139, opened 2018-10-22
fix for transforms operating on auto-invisible traces

alexcjohnson:
Transforms mean `supplyDefaults` gets run twice. This is mostly fine, but there are some unwanted effects from running it on a trace object that has already had its defaults filled in. The one I'm concerned with here is that if a trace is explicitly `visible: false` we don't even run the module defaults, but occasionally there are things that automatically trigger `visible: false` in the module defaults, after we've already set some properties in `fullTrace`. Because the `visible: false` only makes sense based on those properties, we still want them to be part of the final trace (for example, to make RCE and toolpanel happy). So in this PR I detect this situation and clear `expandedTrace.visible` so that the original logic will run again the second time through `supplyDefaults`.

@etpinard can you think of any unwanted consequences of this? cc @antoinerg @archmoj - one of the darker, dingier corners of plotly.js...

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -951,6 +951,8 @@ plots.supplyDataDefaults = function(dataIn, dataOut, layout, fullLayout) {
         fullTrace._expandedIndex = cnt;
 
         if(fullTrace.transforms && fullTrace.transforms.length) {
+            var sdInvisible = trace.visible !== false && fullTrace.visible === false;
+
             var expandedTraces = applyTransforms(fullTrace, dataOut, layout, fullLayout);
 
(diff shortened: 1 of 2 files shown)


etpinard:
> can you think of any unwanted consequences of this?

Nop, I can't think of any.

Jugging from a quick grep, this `splom` partial visibility scenario is the only case where we set `traceOut.visible` to false other than when data arrays are missing or empty.


I'd say :dancer:

etpinard:
Referencing https://github.com/plotly/plotly.js/pull/1830, which highlights a similar (very) dark corner of transform-land.

PR #3143, opened 2018-10-23
sankey - fix issue with large padding

### LINKED ISSUE #3140: Sankey issue
Here is the codepen for the reported bug: https://codepen.io/rbakker/pen/gBjXde
The pen shows two plotly plots with Sankey diagram with the same data.
The only difference is that the first plot uses node.pad = 10, the second uses node.pad = 2.
In the first plot the nodes are not drawn at all.

antoinerg:
Closes #3140 !

If `node.pad` is too large and overridden during the rendering step, a `Lib.warn` call is made to let the user know.


test/jasmine/tests/sankey_test.js
```diff
@@ -771,6 +771,27 @@ describe('sankey tests', function() {
             .then(done);
         });
     });
+
+    it('emits a warning if node.pad is too large', function(done) {
```
alexcjohnson: can you also test something about the node & link thicknesses? Either here or in an image...

antoinerg: I went ahead and added a baseline in commit 352cc61!

antoinerg:
@alexcjohnson What are the next steps I should do before merging this PR? Should I publish `@plotly/d3-sankey` to NPM?

alexcjohnson:
💃

PR #3149, opened 2018-10-24
Improvements for CI build artifacts

etpinard:
I'm a <strike>big</strike> huge fan of @antoinerg 's https://github.com/plotly/plotly.js/pull/3136, so I couldn't resist improving it slightly.

This PR:
- makes the build artifacts get "published" even when the tests fail (which can be useful in some situations)
- echoes the artifact URLs in the publish step using Circle's [built-in environment variables](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables), which can serve as an easier click-to alternative to @antoinerg's [fancy shell footwork](https://github.com/plotly/plotly.js/pull/3136#issuecomment-432059992).

cc @plotly/plotly_js

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,8 +162,4 @@ workflows:
             - build
       - publish:
           requires:
-            - test-jasmine
-            - test-jasmine2
-            - test-image
-            - test-image2
-            - test-syntax
+            - build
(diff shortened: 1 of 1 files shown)


antoinerg:
Thank you @etpinard  :) I like your improvements as well :tada:

The only reason I had `publish` be the last build step was to have the following URL point to it:
`https://circleci.com/api/v1.1/project/github/plotly/plotly.js/latest/artifacts?branch=master\&filter=successful`. It turns out this was unreliable because `latest` could point to another job if the tests were still running. Therefore, it makes sense to run `publish` right after `build`!

:dancer:

Issue #3058, opened 2018-09-28
Prune unsupported global-level trace attributes


alexcjohnson:
There are a bunch of [attributes](https://github.com/plotly/plotly.js/blob/master/src/plots/attributes.js) that we [insert in the schema](https://github.com/plotly/plotly.js/blob/67a05c707e02aed77674e7a0c557a56907338421/src/plot_api/plot_schema.js#L464), and largely [include in `_fullData`](https://github.com/plotly/plotly.js/blob/67a05c707e02aed77674e7a0c557a56907338421/src/plots/plots.js#L1094) for all trace types. But many trace types don't support all of these attributes. As a simple workaround in @jonmmease's `parcats` PR, https://github.com/plotly/plotly.js/pull/2963/commits/273d2d04744efdcff37cd84cd30328644b42c997 (following @etpinard's suggestion https://github.com/plotly/plotly.js/pull/2963#issuecomment-423569219) hides a the ones `parcats` doesn't use from the schema, but they will still show up in `_fullData`.

As I said in https://github.com/plotly/plotly.js/pull/2963#issuecomment-423574232:

> There's a module `categories: ['noOpacity']` which prevents *coercing* `opacity` (so it's not in `fullData`, which is important for the editor) but does not take it out of the schema (and anyway only handles `opacity`). @etpinard 's solution would fix the schema but not `fullData`. Sounds like what we really want is something like:
> ```js
> module.omitGlobalAttrs: ['selectedpoints', 'hoverlabel', 'opacity', 'ids', 'customdata']
> ```
> that would do both.
> 
> There are probably quite a few other traces that omit some of these - in fact, probably everything that sets `traceOut._length = null` has no use for `selectedpoints`, `ids`, and `customdata`...

And @etpinard responded https://github.com/plotly/plotly.js/pull/2963#issuecomment-423646309:

> Good point here. But I think for this PR my suggestion in https://github.com/plotly/plotly.js/pull/2963#issuecomment-423569219 will be _good enough_. The react-chart-editor (aka RCE) doesn't know how to handle traces with `dimensions` yet, so having a few extra keys in fullData won't affect anyone at this stage. But yeah, once merged we should open a new issue (similar to https://github.com/plotly/plotly.js/issues/2834 and https://github.com/plotly/plotly.js/issues/2908) about this.

So here's that issue :)

alexcjohnson:
[...]

PR #3150, opened 2018-10-24
sankey hover improvements

antoinerg:
Addressing comments of @etpinard on PR https://github.com/plotly/plotly.js/pull/3096.

- [x] Replace `.map` call by `Lib.repeat`
- [x] Enable trace-level `hover(info|label)`

I also DRYed up some tests!

### DIFF
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -14,10 +14,21 @@ var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
 
+var extendFlat = require('../../lib/extend').extendFlat;
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
 
 module.exports = overrideAll({
-    hoverinfo: plotAttrs.hoverinfo,
+    hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
+        flags: [],
+        arrayOk: false,
+        description: [
+            'Determines which trace information appear on hover.',
+            'If `none` or `skip` are set, no information is displayed upon hovering.',
+            'But, if `none` is set, click and hover events are still fired.',
+            'Note that this attribute is superseded by `node.hoverinfo` and `node.hoverinfo`',
+            'for nodes and links respectively.'
+        ].join(' ')
+    }),
     hoverlabel: fxAttrs.hoverlabel,
     domain: domainAttrs({name: 'sankey', trace: true}),
 
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -9,13 +9,16 @@
 'use strict';
 
 var fontAttrs = require('../../plots/font_attributes');
+var plotAttrs = require('../../plots/attributes');
 var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
 
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
 
-var attrs = module.exports = overrideAll({
+module.exports = overrideAll({
+    hoverinfo: plotAttrs.hoverinfo,
+    hoverlabel: fxAttrs.hoverlabel,
     domain: domainAttrs({name: 'sankey', trace: true}),
 
     orientation: {
@@ -205,6 +208,3 @@ var attrs = module.exports = overrideAll({
         description: 'The links of the Sankey plot.'
     }
 }, 'calc', 'nested');
-// hide unsupported top-level properties from plot-schema
-attrs.hoverinfo = undefined;
-attrs.hoverlabel = undefined;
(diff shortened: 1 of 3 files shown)


src/traces/sankey/defaults.js
```diff
@@ -53,16 +55,14 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
+    handleHoverLabelDefaults(linkIn, linkOut, coerceLink, hoverlabelDefault);
 
     var defaultLinkColor = tinycolor(layout.paper_bgcolor).getLuminance() < 0.333 ?
                 'rgba(255, 255, 255, 0.6)' :
                 'rgba(0, 0, 0, 0.2)';
 
-    coerceLink('color', linkOut.value.map(function() {
-        return defaultLinkColor;
-    }));
+    coerceLink('color', Lib.repeat(defaultLinkColor, linkOut.value.length));
```
etpinard: thanks :beers:

src/traces/sankey/defaults.js
```diff
@@ -21,6 +21,8 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
     }
 
+    var hoverlabelDefault = Lib.extendDeep(layout.hoverlabel, traceIn.hoverlabel);
```
etpinard: Nice touch :paintbrush:

src/traces/sankey/attributes.js
```diff
@@ -9,13 +9,16 @@
...
+var plotAttrs = require('../../plots/attributes');
 var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
 
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
 
-var attrs = module.exports = overrideAll({
+module.exports = overrideAll({
+    hoverinfo: plotAttrs.hoverinfo,
```
etpinard: We'll need to extend the default `hoverinfo` declaration 

https://github.com/plotly/plotly.js/blob/36d9fb4ad341510823e6e25cb9202867ac94a94a/src/plots/attributes.js#L117-L130

with new flags. I'm thinking `'none'`, `'skip'` and `'all'`(the default) which I think should resolve @alexcjohnson 's concern https://github.com/plotly/plotly.js/pull/3096#discussion_r227626370

We should probably have a special description too to explain how trace `hoverinfo` can propagate to the node/link `hoverinfo`.

antoinerg: @etpinard  I'm having trouble setting `hoverinfo.flags` to an empty array `[ ]`. Running `Plotly.PlotSchema.get().traces.sankey.attributes.hoverinfo` always return 5 flags.

The following doesn't work:
```
var attrs = module.exports = overrideAll({
    hoverinfo: plotAttrs.hoverinfo,
    ...
})
attrs.hoverinfo.flags = []
```

The following also doesn't work:
```
var attrs = module.exports = overrideAll({
        flags: ['random']
    }),
    ...
})
```
although it replaces the first array element with `'random'`.

alexcjohnson: see https://github.com/plotly/plotly.js/issues/3058#issuecomment-429069306 - it's just a problem with the schema, the modules themselves are fine.

etpinard: > the modules themselves are fine.

Right, `gd._fullData[0]._module.attributes.hoverinfo.flags` should have the new list.

antoinerg: Thank you guys for the quick reply! But won't this be an issue for the automatically generated documentation?

alexcjohnson: sure will 😅

alexcjohnson: and already is - feel like tackling #3058, or at least that part of it?

src/traces/sankey/attributes.js
```diff
@@ -9,13 +9,27 @@
...
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
 
+var extendFlat = require('../../lib/extend').extendFlat;
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
 
-var attrs = module.exports = overrideAll({
+module.exports = overrideAll({
+    hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
+        flags: [],
```
etpinard: Nicely done, `plots.hoverinfo` already has the correct `extras` field

https://github.com/plotly/plotly.js/blob/d5ebb50a727821f6705d72182d7822d0097e8a41/src/plots/attributes.js#L121

src/traces/sankey/attributes.js
etpinard: That or, remove `hoverinfo` from the global (plots/attributes.js) trace attributes.

etpinard:
Nice job @antoinerg this PR is already worthy of a :dancer:

I'd say tackling https://github.com/plotly/plotly.js/issues/3058 or https://github.com/plotly/plotly.js/pull/3150#discussion_r228172990 will deserve a PR of its own.

src/traces/sankey/attributes.js
antoinerg: @alexcjohnson I will look into #3058! I actually wanted to improve on plot.ly/javascript/reference so I guess that would be a good start for me since they're related :)

Issue #3116, opened 2018-10-17
Outside text relativebarmode confusing when bars disconnected


eivindjahren:
With relative barmode and `{textposition: 'outside'}`, outside text will smartly position the text of the first bar as 'inside'. However, when there are two disconnected bars this results in the text of both bars being 'inside'.

https://codepen.io/haskellelephant/pen/JmpJMZ

alexcjohnson:
Interesting, there's at least one clear bug here, that the orange bar always gets its text pushed inside if it has a `base`, even if there are no other bars in that column at all (ie if you hide the blue one).

But the logic for the blue trace is a bit more complicated. I guess the simplest option would be if the trace stacked on it has a `base` at all, no matter the value, we don't automatically push text inside. Basically, once you've specified `base` (for the next trace) you're on your own with `textposition` and it's up to you to ensure it doesn't overlap another bar. We could *try* to figure out whether there's room outside a bar for text, based on the other bars around, but I suspect no matter what we'll make the wrong choice sometimes in some users eyes at least, so a better bet would be to make a clear statement: 

If the trace above has a base -> no auto-inside text, you have to do it yourself.

Thanks @eivindjahren - another very clear and helpful report!

eivindjahren:
@alexcjohnson I have made a work in progress PR on my fork (first PR here so needed to conform to the contribution guidelines). It should implement your proposed solution. I still want to test it out a bit more, but hopefully this is in the right direction.

alexcjohnson:
Great PR @eivindjahren - I'm always happy to see a one-line change in `src` backed up by 60 lines of tests!

> Another solution is to have 'outside' always mean 'outside', if you want the smart behavior you use 'auto'.

Possible, though the existing `'auto'` logic depends on the text size relative to the bar height, and ends up meaning "usually inside, but put it outside if we can and the bar is too small" - would be nice to clarify this behavior (for both `'auto'` and `'outside'`) in the `textposition` description though!

eivindjahren:
@alexcjohnson I tried to summarize the behavior of *outside* in the `textposition` description, and adressed your comment about splitting the testdata trace. Not sure if the description is helpful though.

PR #3156, opened 2018-10-25
Don't override outside when base is set

eivindjahren:
Adresses #3116 (see also https://codepen.io/haskellelephant/pen/JmpJMZ). When base is set, which bar is outmost is difficult to calculate. Instead, outside is chosen for all bars in that case.


etpinard:
Very nice PR. I can't think of any unwanted side-effects here. 

:dancer: merging :dancer:

PR #3160, opened 2018-10-25
circleci update plot-schema.json during publish job

### LINKED ISSUE #3159: CircleCI publishes outdated `plot-schema.json`
On CircleCI `publish` happens right after the `build` job.

The build job on the CI is `npm run cibuild` which doesn't update the `plot-schema.json`. We should update it either by:
1. adding code to https://github.com/plotly/plotly.js/blob/master/tasks/cibundle.js 
2. creating a new task only for that purpose which would only be executed in the `publish` job.

Option 1 adds 15 seconds to the build step. So I would favor option 2 since it can be runned in parallel.

### master
``` 
[antoine@newton:~/plotly/plotly.js]$ time npm run cibuild

> plotly.js@1.41.3 cibuild /home/antoine/plotly/plotly.js
> npm run preprocess && node tasks/cibundle.js

###
> plotly.js@1.41.3 preprocess /home/antoine/plotly/plotly.js
> node tasks/preprocess.js

ok plotly-geo-assets.js
ok plotly.js
ok plotly.min.js

real    0m39.960s
user    0m59.195s
sys     0m2.339s
```
### Option 1
```
[antoine@newton:~/plotly/plotly.js]$ time npm run cibuild

> plotly.js@1.41.3 cibuild /home/antoine/plotly/plotly.js
> npm run preprocess && node tasks/cibundle.js


> plotly.js@1.41.3 preprocess /home/antoine/plotly/plotly.js
> node tasks/preprocess.js

ok plotly-geo-assets.js
ok plotly-with-meta.js
ok plot-schema.json
ok plotly.js
ok plotly.min.js

real    0m53.048s
user    1m21.699s
sys     0m3.376s
```

antoinerg:
Fixes #3159

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,6 +130,9 @@ jobs:
       - checkout
       - attach_workspace:
           at: ~/plotly.js
+      - run:
+          name: Update plot-schema.json
+          command: node tasks/bundle_plot_schema.js
       - store_artifacts:
           path: build/plotly.js
           destination: /plotly.js
--- a/tasks/bundle_plot_schema.js
+++ b/tasks/bundle_plot_schema.js
@@ -1,0 +1,19 @@
+var constants = require('./util/constants');
+var _bundle = require('./util/browserify_wrapper');
+var makeSchema = require('./util/make_schema');
+
+/*
+ * Trimmed down version of ./bundle.js for CI testing
+ *
+ * Outputs:
+ *
+ *  - plotly-with-meta.js bundle in dist/
+ *  - plot-schema.json in dist/ (for reference)
+ */
+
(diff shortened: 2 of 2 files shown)


antoinerg:
@etpinard Can you look at this PR again?

antoinerg:
Is the current PR OK?

The other alternative is to run `npm run build` during the `publish` job and not have code duplication. This will come at the expense of building everything which 2-3 minutes. Since it runs in parallel and it's faster than `test-(image|jasmine)`, it wouldn't slow down the whole suite.

etpinard:
I like this solution using `./util/browserify_wrapper` is DRY enough 

Thanks! :dancer:

Issue #3115, opened 2018-10-17
Text does not show up on bars with empty area


eivindjahren:
If a bar does not have any width or height, text does not show up, even when `{textposition: 'outside'}` is set.

https://codepen.io/haskellelephant/pen/PyQemG

eivindjahren:
The relevant line is https://github.com/plotly/plotly.js/blob/d5ebb50a727821f6705d72182d7822d0097e8a41/src/traces/bar/plot.js#L81 as it removes a bar group if the associated bar is empty so text is never attached to it.

etpinard:
Perhaps we should do like `barpolar` does and draw invalid bars with a blank path

weilu:
Still doesn't show the text for default textposition, which is "auto". Setting `{textposition: "outside"}` explicitly fixed it for me.

alb-erandelcastillo:
@etpinard Issue still occurs for bar charts within subplots, where label text doesn't appear for traces with 0 height. The issue can be recreated with this:

`chart = make_subplots(rows=2, cols=1)
                
chart.append_trace(
    go.Bar(
        x=['a','b','c','d'],
        y=[2, 4, 0, 4],
        hoverinfo=["y"],
        text=['a text','b text','c text','d text'],
        textposition="outside"
        ),
    row=2, col=1
    )
    
chart.append_trace(
    go.Bar(
        x=['a','b','c','d'],
        y=[0, 0, 0, 1],
        hoverinfo=["y"],
        text=['a text','b text','c text','d text'],
        textposition="outside"
        ),
    row=2, col=1
    )
        
chart.append_trace(
    go.Bar(
        x=['a','b','c','d'],
        y=[0, 5, 3, 1],
        hoverinfo=["y"],
        text=['a 2','b 2', 'c 2', 'd 2'],
        textposition="outside"
        ),
    row=1, col=1
    )
            
chart.update_layout(
        barmode="stack", showlegend=False, xaxis=dict(showticklabels=False))`

mangel2095:
Hi, I am puting labels with the total value of the each bar.

with `{textposition: "auto"}`, I had this result:

![imagen](https://user-images.githubusercontent.com/18307144/70850651-c210eb00-1e5a-11ea-8eb6-34328c0c652d.png)

When I change that value by `"outsite"`, I get this:

![imagen](https://user-images.githubusercontent.com/18307144/70850715-5e3af200-1e5b-11ea-8433-132127173492.png)

[...]

PR #3164, opened 2018-10-26
Draw bar group if there is text

eivindjahren:
Adresses #3115.

The groups `g.point` contains the path for bars, but also the corresponding
text element. Now, the group is made if it should contain text, even
when it will not contain a path.

### DIFF
--- a/src/traces/bar/plot.js
+++ b/src/traces/bar/plot.js
@@ -76,9 +76,10 @@ module.exports = function plot(gd, plotinfo, cdbar, barLayer) {
                 di.ct = [(x0 + x1) / 2, y1];
             }
 
+            var text = getText(trace, i);
             if(!isNumeric(x0) || !isNumeric(x1) ||
                     !isNumeric(y0) || !isNumeric(y1) ||
-                    x0 === x1 || y0 === y1) {
+                    ((x0 === x1 || y0 === y1) && !text)) {
                 bar.remove();
                 return;
             }
@@ -120,11 +121,13 @@ module.exports = function plot(gd, plotinfo, cdbar, barLayer) {
                 y1 = fixpx(y1, y0);
             }
 
-            Lib.ensureSingle(bar, 'path')
-                .style('vector-effect', 'non-scaling-stroke')
-                .attr('d',
-                    'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
-                .call(Drawing.setClipUrl, plotinfo.layerClipId);
+            if(x0 !== x1 && y0 !== y1) {
+                Lib.ensureSingle(bar, 'path')
+                    .style('vector-effect', 'non-scaling-stroke')
+                    .attr('d',
+                        'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
+                    .call(Drawing.setClipUrl, plotinfo.layerClipId);
+            }
 
             appendBarText(gd, bar, cd, i, x0, x1, y0, y1);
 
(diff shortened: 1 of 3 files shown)


src/traces/bar/plot.js
```diff
@@ -120,11 +122,13 @@ module.exports = function plot(gd, plotinfo, cdbar, barLayer) {
...
-                .attr('d',
-                    'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
-                .call(Drawing.setClipUrl, plotinfo.layerClipId);
+            if(!is_empty) {
+                Lib.ensureSingle(bar, 'path')
+                    .style('vector-effect', 'non-scaling-stroke')
+                    .attr('d',
+                        'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
+                    .call(Drawing.setClipUrl, plotinfo.layerClipId);
+            }
```
alexcjohnson: Then we also need to remove the bar if it later disappears. Off your branch I can do:
```js
Plotly.newPlot(gd,
    [{type:'bar',y:[0,1,0,1],text:'hi',textposition:'outside',cliponaxis:false}],
    {width: 400, height: 400})
```
Which looks great:
screen shot 2018-10-29 at 11 14 56 am
But then if I do:
```js
Plotly.restyle(gd,'y',[[1,0,1,0]])
```
The newly-removed bars are still there (with their text moved to the right place):
screen shot 2018-10-29 at 11 16 12 am

etpinard:
Hmm. I think I would prefer using the "blank-path" style solution as described in https://github.com/plotly/plotly.js/issues/3115#issuecomment-433096687

test/image/mocks/bar_attrs_relative.json
```diff
@@ -9,7 +9,7 @@
             "type":"bar"
         }, {
             "width":[0.4,0.6,0.8,1],
-            "text":["Three",2,"inside text",0],
```
alexcjohnson: Hmm, I think we need to figure out how to handle these cases. If I put values in where you put `""` - `"zero"` here and `"empty"` in the last trace - I get the following unfortunate situation:
screen shot 2018-10-29 at 11 23 46 am
I'm not sure why "zero" is showing up below where its bar would have stacked, seems like it should be above and it should push "4" inside the blue bar. "empty" also shows up below where its bar would be if that bar were positive... in an ideal world we might detect that its neighboring bars are negative - in this case only "outside text" to the right, but if it weren't the first bar, perhaps we would require the bar before to also be negative - we treat this as a "negative zero" and move the "empty" label down below the green -1? Anyway that's a pretty weird edge case, I wouldn't worry about it too much, if we could get it to work right with only a "positive zero" I'd be satisfied.

alexcjohnson:
> Hmm. I think I would prefer using the "blank-path" style solution as described in #3115 (comment)

That would certainly make the logic easier, no need for a special case on removing a previously-visible bar.

test/image/mocks/bar_attrs_relative.json
eivindjahren: I guess that the "zero" and " empty" text is showing up near the empty red bars is the new expected behavior. However, `textposition: 'auto'` is placing 'zero' below the empty bar and not pushing '4' inside. I'll look around and see if the blank-path solution solves this any better.

etpinard:

I made a quick attempt at the blank path solution (more in: https://github.com/plotly/plotly.js/issues/3115#issuecomment-433096687), and I faced the same problems described in https://github.com/plotly/plotly.js/pull/3164#discussion_r228981528 where the `bar_attrs_relative.json` mock is acting up.

Moving (back) the discussion to https://github.com/plotly/plotly.js/issues/3115

PR #3168, opened 2018-10-26
Fix scatter fill regression

etpinard:
This PR fixes an unreleased regression originating from https://github.com/plotly/plotly.js/pull/3087. I'd like to get this in before the 1.42.0 release.

Scatter fills on master currently do not restyle correctly and lead to these logs:

image

but **no** failed tests. That's because those logs come from:

https://github.com/plotly/plotly.js/blob/0a310c99455bf7389f4bb891392c806b59d93d47/src/components/drawing/index.js#L201-L207

which do **not** make `failTest` throw, leading to false-positive tests.

----------

The first commit fixes and :lock:s the bug, and the second commit :hocho:s that try-catch (which is at least 4 years old) which is probably a bit more dangerous, but I'm willing to take that chance. I hope @plotly/plotly_js agrees.

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -198,13 +198,7 @@ drawing.fillGroupStyle = function(s) {
     s.style('stroke-width', 0)
     .each(function(d) {
         var shape = d3.select(this);
-        try {
-            shape.call(Color.fill, d[0].trace.fillcolor);
-        }
-        catch(e) {
-            Lib.error(e, s);
-            shape.remove();
-        }
+        shape.call(Color.fill, d[0].trace.fillcolor);
     });
 };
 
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Good catch! I noticed those logs but studiously ignored them 🙈 - glad someone is paying more attention!!

> The first commit fixes and 🔒s the bug

How does this 🔒 it?

> the second commit 🔪s that try-catch ... which is probably a bit more dangerous, but I'm willing to take that chance.

Me too. We have enough tests at this point, don't worry about removing it. As evidenced here, a `try/catch` like this is more likely to paper over a problem than to provide useful behavior!

etpinard:
> How does this  it?
> 
> >

By spying on `Lib.error`.

etpinard:
... by I removed that spy in the second commit https://github.com/plotly/plotly.js/pull/3168/commits/1f3c10c735927d57b78151be8c4d43bdf35e53d7 as now the `fillGroupStyle` errors lead to `failTest`.

alexcjohnson:
Got it, thanks! 💃

Issue #3142, opened 2018-10-23
width.bar doesnt show if width is typed array


Kully:
re the python bug report here: https://github.com/plotly/plotly.py/issues/1231

bar.width cannot be a typed array. In the codepen example, I used a `Float64Array`.

See this codepen: https://codepen.io/plotly/pen/KGGarx

```javascript
var a = [0.1, 0.4, 0.7]

var myArray = a;
var myTypedArray = new Float64Array(a);

var trace1 = {
     type: 'bar',
     width: myArray,
     x: [1, 2, 3],
     xaxis: 'x1',
     y: [3, 2, 1]
}

var trace2 = {
     type: 'bar',
     width: myTypedArray,
     x: [1, 2, 3],
     xaxis: 'x2',
     y: [3, 2, 1]
}

var data = [trace1, trace2];

var layout = {
  xaxis1: {'domain': [0, 0.45]},
  xaxis2: {'domain': [0.55, 1]},
};

Plotly.newPlot('div1', data, layout);
```

etpinard:
Thanks for the report!

PR #3169, opened 2018-10-26
fix support for typed arrays in bar 'width' and 'offset'

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3142 and a similar unlisted bug for bar `offset`.

cc @plotly/plotly_js

### DIFF
--- a/src/traces/bar/cross_trace_calc.js
+++ b/src/traces/bar/cross_trace_calc.js
@@ -356,7 +356,7 @@ function applyAttributes(sieve) {
 
         if(isArrayOrTypedArray(offset)) {
             // if offset is an array, then clone it into t.poffset.
-            newPoffset = offset.slice(0, calcTrace.length);
+            newPoffset = Array.prototype.slice.call(offset, 0, calcTrace.length);
 
             // guard against non-numeric items
             for(j = 0; j < newPoffset.length; j++) {
@@ -377,12 +377,12 @@ function applyAttributes(sieve) {
             t.poffset = offset;
         }
 
-        var width = fullTrace._width || fullTrace.width,
-            initialBarwidth = t.barwidth;
+        var width = fullTrace._width || fullTrace.width;
+        var initialBarwidth = t.barwidth;
 
         if(isArrayOrTypedArray(width)) {
             // if width is an array, then clone it into t.barwidth.
-            var newBarwidth = width.slice(0, calcTrace.length);
+            var newBarwidth = Array.prototype.slice.call(width, 0, calcTrace.length);
 
             // guard against non-numeric items
             for(j = 0; j < newBarwidth.length; j++) {
--- a/test/jasmine/tests/bar_test.js
+++ b/test/jasmine/tests/bar_test.js
@@ -398,6 +398,46 @@ describe('Bar.crossTraceCalc (formerly known as setPositions)', function() {
     });
 
     it('should guard against invalid width items', function() {
+        var w = [0.1, 0.4, 0.7];
+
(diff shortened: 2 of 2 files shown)


src/traces/bar/cross_trace_calc.js
```diff
@@ -356,7 +356,7 @@ function applyAttributes(sieve) {
 
         if(isArrayOrTypedArray(offset)) {
             // if offset is an array, then clone it into t.poffset.
-            newPoffset = offset.slice(0, calcTrace.length);
+            newPoffset = Array.prototype.slice.call(offset, 0, calcTrace.length);
```
etpinard: This here converts a typed array to a regular array. Note that `slice` for typed array works, but below e.g.

https://github.com/plotly/plotly.js/blob/0a310c99455bf7389f4bb891392c806b59d93d47/src/traces/bar/cross_trace_calc.js#L432-L435

we expect computed offsets and widths to be within regular arrays.

alexcjohnson: Is it more complicated than turning those `Array.isArray` calls into `isArrayOrTypedArray`? The way you have it here seems OK, because bars never have *that* many elements, but it seems a shame to convert back to regular arrays early if it's not too hard to support typed arrays.

etpinard: > Is it more complicated than turning those `Array.isArray` calls into `isArrayOrTypedArray`?

Yes because the rest of `crossTraceCalc` assumes you have arrays you can `.push`ed into e.g.:

https://github.com/plotly/plotly.js/blob/aea6b8d796d367c3930da9753a8b9151e10cd5d8/src/traces/bar/cross_trace_calc.js#L405-L407

which can't be done using typed arrays.

test/jasmine/tests/bar_test.js
```diff
@@ -397,6 +397,46 @@ describe('Bar.crossTraceCalc (formerly known as setPositions)', function() {
...
+            offset: new Float32Array(o),
+            y: [1, 2, 3]
+        }]);
+
+        var cd = gd.calcdata;
+        assertArrayField(cd[0][0], 't.poffset', o);
+        assertArrayField(cd[1][0], 't.poffset', o);
+        assertPointField(cd, 'x', [
+            [0.5, 1.8, 3.1],
+            [0.5, 1.8, 3.099]
```
alexcjohnson: Huh, took me a while to figure out where these numbers came from... surprised I had never noticed this before but it seems weird that `offset` sets the left edge of each bar rather than its center. Oh well, a little late to change that. Why does the second one get `3.099` instead of `3.1`?

src/traces/bar/cross_trace_calc.js
alexcjohnson: Ah OK, makes sense. We can revisit if/when we get rid of `calcdata`, before that it doesn't make much difference anyway.

test/jasmine/tests/bar_test.js
etpinard: I'm not sure. 

With a regular array, we get:

```
[0.5, 1.7999999999999998, 3.1]
```

with a typed array, we get:

```
[0.5000000014901161, 1.8000000059604644, 3.099999988079071]
```

alexcjohnson: ah ok, that's just because it's a `Float32Array` ie single precision, not double. But it's still closer to 3.1 than to 3.099 😏

alexcjohnson:
💃 💪

Issue #2844, opened 2018-07-26
Plotly on Mobile - page scrolling


maverick-2:
Hi developers,
I have troubles scrolling a page with bar chart on mobile devices. The chart’s default behaviour is zooming. It makes page scrolling impossible, unless label area is used for that. This is not intuitive, and confusing for users.
Is there a way to switch off zooming behaviour, and enable normal scrolling instead. Other libraries do that, for example:
https://www.highcharts.com/

I understand that pinch/zoom and other mobile stuff is work in progress, but can you enable at least simple page scrolling??
Thank you!

etpinard:
plotly.js disables scroll zoom by [default](https://github.com/plotly/plotly.js/blob/1ecef456e418667e1b0378b7c45c8bb9eff94dcd/src/plot_api/plot_config.js#L69-L70), so I'm a little puzzled by this report.

Are you referring to graphs on plotly cloud's shareplot view (e.g. https://plot.ly/~etpinard/7609)?

Alternatively, this might be a side-effect of https://github.com/plotly/plotly.js/issues/2371

etpinard:
Or perhaps, you're referring to 3D or/and geo subplots that indeed have no way to disable zoom at the moment, ref https://github.com/plotly/plotly.js/issues/143

maverick-2:
Hi Étienne,

Thank you for prompt response.
I'm talking about a bar chart, like one of these: 
https://plot.ly/javascript/bar-charts/

Is it possible to change scroll zoom settings dynamically? On mobile device I want to have scroll zoom enabled by default, and on desktop – disabled.

etpinard:
You could use a package like https://www.npmjs.com/package/has-hover to determine if you're on desktop or mobile then

```js
var hasHover = require('has-hover')

Plotly.newPlot(gd, [/**/], {/**/}, {
  scrollZoom: !hasHover
})
```

etpinard:
@maverick-2 did https://github.com/plotly/plotly.js/issues/2844#issuecomment-408425623 solve your issue? Would it be ok if we closed this ticket? Thank you.

maverick-2:
Étienne,

I set scrollZoom to true, then to false, but did not notice any change in behaviour.
https://codepen.io/Igor1812/pen/ajqrYL

When I swipe on the bar aria, the default behaviour of the chart is ZOOM, but I need the chart just to be scrollable together with the web page. 
Should I use any specific version of Plotly? 
I use plotly-latest.min.js

[...]

PR #3170, opened 2018-10-26
Add dragmode: false

jonfunkhouser:
This PR adds the `dragmode: false` functionality mentioned in https://github.com/plotly/plotly.js/issues/2844#issuecomment-409940288

I added an example chart with this setting to my existing sample, so this can be previewed on mobile devices: https://jonfunkhouser.github.io/plotly-mobile/


etpinard:
Thanks very much for this PR and let me apologize for the delayed response (our 1.42.0 release cycle was a little hectic).

Could you confirm that hover still works on `dragmode: false`?

jonfunkhouser:
Sorry, I was away last week. Yes, hover still works in `dragMode: false`. I updated the sample site to include `plotly_hover` debug output.

etpinard:
Thanks @jonfunkhouser !


peek 2018-11-05 18-13

I wonder if we could do better. By dragging we effectively halt hover and the hover labels (if any) remain in place. Maybe instead we could treat mouse "drag" as a mouse "move" under `dragmode: false`. But that might make things too confusing though.

----

Regardless, we'll need to add a test to :lock: down the hover on `dragmode: false` behavior. @jonfunkhouser would you be interested?

jonfunkhouser:
Yes, I'll add a test for that case. I'll also investigate your suggestion, it makes a lot of sense.

etpinard:
Brilliant PR, this will be part of our next minor release.

Issue #3186, opened 2018-10-30
plots in plotly.js tutorial docs not displaying in Internet Explorer 11


priyatharsan:
This issue is specific only to the plots in `CodePen`.

![codepen](https://user-images.githubusercontent.com/25846124/47735601-367ecf80-dc43-11e8-87d7-932ca49cd38c.gif)

etpinard:
Can you try opening https://codepen.io/ and any other example codepen in IE11 to help us out? Thanks.

Braintelligence:
@etpinard Current release https://github.com/plotly/plotly.js/releases/tag/v1.42.0 doesn't work for me in IE 11 anymore. Something about `The object doesn't support the attribute or method "remove"`
The release before works fine...

priyatharsan:
Yeah I get this message when opening https://codepen.io/ ,
![capture](https://user-images.githubusercontent.com/25846124/47739122-0c311000-dc4b-11e8-9fcf-6988576ef4b5.PNG)

like I doubted it's related to codepen and not plotly but maybe I could be wrong. 
Also, the full page view is working fine in IE11 
![capture1](https://user-images.githubusercontent.com/25846124/47739392-a6915380-dc4b-11e8-8a0a-8dcbf2438fde.PNG)

But not the pen view: 
![capture3](https://user-images.githubusercontent.com/25846124/47739404-adb86180-dc4b-11e8-98e1-6434ee1f0a63.PNG)

Braintelligence:
IE 11 doesn't support childNode.remove() https://caniuse.com/#search=remove

Braintelligence:
Providing this polyfill before triggering Plotly works: https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove#Polyfill

etpinard:
@Braintelligence which example are you looking at?

alexcjohnson:
This seems to be the offender:
https://github.com/plotly/plotly.js/blob/0124828f061cc1483bf09781c81c12050971bc62/src/lib/index.js#L716

Braintelligence:
@etpinard I hope alex hit the spot. Sadly I'm not looking at any example, I'm looking at a website in the work that is still under NDA 🙈.

priyatharsan:
[...]

PR #3187, opened 2018-10-30
Fix modebar <style> update in IE11

etpinard:
This PR is a fixup of https://github.com/plotly/plotly.js/pull/3068 for IE11, which hopefully will be enough to close https://github.com/plotly/plotly.js/issues/3186. It would be nice to add a test before merging this in.

cc @plotly/plotly_js

### DIFF
--- a/test/jasmine/assets/unpolyfill.js
+++ b/test/jasmine/assets/unpolyfill.js
@@ -1,0 +1,22 @@
+/*
+ * Un-polyfills - to be included in karma.conf.js to catch
+ * browser-dependent errors.
+ */
+
+'use strict';
+
+(function(arr) {
+    arr.forEach(function(item) {
+        Object.defineProperty(item, 'remove', {
+            configurable: true,
+            enumerable: true,
+            writable: true,
+            value: function remove() {
+                throw Error([
+                    'test/jasmine/assets/unpolyfill.js error: calling ChildNode.remove()',
+                    'which is not available in IE.'
+                ].join(' '));
+            }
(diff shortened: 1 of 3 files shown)


test/jasmine/assets/unpolyfill.js
```diff
@@ -0,0 +1,22 @@
+/*
+ * Un-polyfills - to be included in karma.conf.js to catch
+ * browser-dependent errors.
+ */
+
+'use strict';
+
+(function(arr) {
+    arr.forEach(function(item) {
```
etpinard: Taken from @alexcjohnson 's https://github.com/plotly/plotly.js/issues/3186#issuecomment-434418436, this gave 

image

before the previous commit.

I chose to throw instead of logging to avoid situations like the one solved in https://github.com/plotly/plotly.js/pull/3168.

antoinerg:
Beautiful! Thanks @etpinard for fixing my mistake so quickly!

:dancer: when the tests pass!

test/jasmine/assets/unpolyfill.js
etpinard: Note: we could also add `unpolyfill.js` to our image tests (like we do for [`strict-d3`](https://github.com/plotly/plotly.js/blob/master/test/image/strict-d3.js)).

I chose not to as throwing errors in our jasmine test should be enough to catch these bugs and debugging the old nw.js image server is a pain.

etpinard:
Merging.

PR #3191, opened 2018-10-31
Use paper_bgcolor for parcats text-shadow instead of hardcoding to white

jonmmease:
Fix to make parcats look good with a dark theme.  Previously the SVG text shadow behind the category labels was hardcoded to white. But this makes the text unbearably blurry when the font is white and the background is dark:

newplot 70

This PR makes the text-shadow take on the color of the `paper_bgcolor`. This doesn't change the appearance with a white background, but significantly improves the situation for dark background:

newplot 72

This image is from a new `parcats_dark` mock to 🔒 it in.

**What about Sankey?**
The text shadow logic was largely borrowed from the `sankey` trace type, so I took a look at this trace on a black background as well.

newplot 73

It looks like the `sankey` trace doesn't pick up it's default font color from `layout.font.color`, and it draws all node labels on top of paths, so it's default behavior still works reasonable well on a dark background.  So I didn't make any changes to it.


alexcjohnson:
Definitely an improvement! I'm just wondering if we want to go one step further and make a new attribute (`tickshadow`?) that defaults to `paper_bgcolor` but people could set it to whatever they want (including transparent).

alexcjohnson:
I guess though if we want this to be categorized as a bugfix, so it can go in a patch, it should stand as is...

etpinard:
Nice fix :dancer:

etpinard:
Looks like @jonmmease isn't online at the moment. I'll merge this in to make it part of 1.42.1

PR #3178, opened 2018-10-29
Parcoords line.color restyle fix

etpinard:
fixes #3101 and supersedes #3125 ... and the last PR before `v1.42.0`

PR #3195, opened 2018-10-31
Second jasmine test to lock issue 3101

archmoj:
Fixes #3178.
The `getAvgPixelByChannel` function is revised and now accepts an id to select which gl layer (e.g. focus/pick/context) to be used by different tests. 
The new jasmine test, checks for the parcoords context layer to remain visible after a restyle call.
@etpinard 
@alexcjohnson

### DIFF
--- a/test/jasmine/tests/parcoords_test.js
+++ b/test/jasmine/tests/parcoords_test.js
@@ -45,6 +45,23 @@ function purgeGraphDiv(done) {
     return delay(50)().then(done);
 }
 
+function getAvgPixelByChannel(id) {
+    var canvas = d3.select(id).node();
+
+    var imgData = readPixel(canvas, 0, 0, canvas.width, canvas.height);
+    var n = imgData.length * 0.25;
+    var r = 0;
+    var g = 0;
+    var b = 0;
+
+    for(var i = 0; i < imgData.length; i++) {
+        r += imgData[i++];
+        g += imgData[i++];
+        b += imgData[i++];
+    }
+    return [r / n, g / n, b / n];
+}
+
 describe('parcoords initialization tests', function() {
 
     'use strict';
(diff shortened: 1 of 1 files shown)


test/jasmine/tests/parcoords_test.js
```diff
@@ -45,6 +45,23 @@ function purgeGraphDiv(done) {
     return delay(50)().then(done);
 }
 
+function getAvgPixelByChannel(id) {
+    var canvas = d3.select(id).node();
+
+    var imgData = readPixel(canvas, 0, 0, canvas.width, canvas.height);
+    var n = imgData.length * 0.25;
```
archmoj: A minor change here:
instead of dividing by 4, we use multiply by 0.25 so that we return rgb fractions and avoid possibility of rounding small values to 0.

alexcjohnson: shouldn't cause problems but... is there actually a difference between these in js? anyway `n` should always be an integer, right?

https://github.com/plotly/plotly.js/blob/458f5938a538517922afb08790a164339e8d1970/test/jasmine/assets/read_pixel.js#L11

etpinard: I'm curious. Can you share an example of this happening in JS?

alexcjohnson: (am I missing something? why didn't that link get turned into a code snippet?)

archmoj: You are totally right and I tested this on IE, FF and Chrome too. 
In JS there is actually no difference in JS between multiplying by 0.25.
Thanks for the note.

etpinard: > (am I missing something? why didn't that link get turned into a code snippet?)

image

it's rendering fine now. I'll blame the hiccup on Microsoft. 

> Thanks for the note.

Thank you for double checking!

etpinard:
Thanks for the test @archmoj 

Let's merge this in :dancer:

Issue #3197, opened 2018-10-31
Plotly 1.42.1 goes unresponsive


rfriedma:
I recently updated from plotly 1.41 to 1.42.1 in order to get some bug fixes. With the same data, plotting is now going unresponsive. I have two links below showing the same data in two versions of plotly where one hangs.
I'm using Chrome 70.0.3538.67 (Official Build) (64-bit)

Here's 1.42.1:
https://codepen.io/rfried/pen/VEJOJL

Here's 1.41.0:
https://codepen.io/rfried/pen/dgBxEg

Can anyone determine something wrong with my data or explain why plotly is hanging?

Thanks

alexcjohnson:
there's a runaway `while` loop here:
![screen shot 2018-10-31 at 2 27 42 pm](https://user-images.githubusercontent.com/2678795/47810261-423cc580-dd19-11e8-804f-d8225ae668ae.png)
https://github.com/plotly/plotly.js/blob/6fda5955d97436534f786f5dbff7f2e121273bb7/src/traces/scattergl/index.js#L401-L407
Came in with #3087 
@archmoj want to take this?

etpinard:
@archmoj I'm on it. We'll have a PR out shortly.

rfriedma:
Thanks for the quick response and fix :)

etpinard:
@rfriedma thank **you** for the report!

PR #3199, opened 2018-10-31
Guard against infinite loops during scattergl line/fill positions cleanup

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3197

cc @plotly/plotly_js

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -395,7 +395,6 @@ function plot(gd, subplot, cdata) {
             scene.line2d.update(scene.lineOptions);
             scene.lineOptions = scene.lineOptions.map(function(lineOptions) {
                 if(lineOptions && lineOptions.positions) {
-                    var pos = [];
                     var srcPos = lineOptions.positions;
 
                     var firstptdef = 0;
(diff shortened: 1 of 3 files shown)


etpinard:
Codepen from #3197 but using a bundle off this branch: https://codepen.io/etpinard/pen/xyvGRJ

src/traces/scattergl/index.js
```diff
@@ -395,15 +395,16 @@ function plot(gd, subplot, cdata) {
...
 
                     var firstptdef = 0;
-                    while(isNaN(srcPos[firstptdef]) || isNaN(srcPos[firstptdef + 1])) {
+                    while(firstptdef < srcPos.length && (isNaN(srcPos[firstptdef]) || isNaN(srcPos[firstptdef + 1]))) {
                         firstptdef += 2;
                     }
                     var lastptdef = srcPos.length - 2;
-                    while(isNaN(srcPos[lastptdef]) || isNaN(srcPos[lastptdef + 1])) {
-                        lastptdef += -2;
+                    while(lastptdef > -1 && (isNaN(srcPos[lastptdef]) || isNaN(srcPos[lastptdef + 1]))) {
```
alexcjohnson: even better: `while(lastptdef > firstptdef && ...` would skip this second loop altogether in the failing case.

BTW do we need this `pos = pos.concat`... thing at all? Can't we just do `lineOptions.positions = srcPos.slice(...)`? It's needed in the fill section below but I don't see why it'd be needed here. In fact, even that `slice` seems like it could be omitted when we didn't drop any points on either end.

etpinard:
Thanks for the quick review @alexcjohnson !

I'll make another patch release tomorrow morning.

PR #3202, opened 2018-11-01
Few more optional packages in lock

archmoj:
In this PR few more packages are highlighted by npm as `optional` which may help improve the build times.
@etpinard


etpinard:
:dancer: nice!

PR #3184, opened 2018-10-30
optional packages

archmoj:
This PR adds optional packages info which may help improve the build times.
@etpinard

PR #3211, opened 2018-11-05
Histogram TypedArray support

### LINKED ISSUE #3210: histogram autobinning regression with TypedArray
It looks like the new autobinning logic has a regression with typed arrays. See https://github.com/plotly/plotly.py/issues/1257.

CodePen: https://codepen.io/jonmmease/pen/mQdXJK?editors=1010

Standard array (1.42.2):
newplot-1

Typed array (1.42.2):
newplot-2

Typed array (1.41.3):
newplot-3

alexcjohnson:
Fixes #3210 

I rearranged `Lib` array utilities into one file (`lib/array`) and added one new one: `Lib.concat`, a clone of `Array.concat` that works with `TypedArrays` and mixed types.

For now I only used it in one place, where it was needed to fix the histogram bug.


alexcjohnson:
@etpinard how do I avoid making this mistake with `package-lock` again? Is there a way to ensure that these `optional` flags don't get wiped out when I `npm i`?

test/jasmine/tests/lib_test.js
```diff
@@ -2459,6 +2459,58 @@ describe('Test lib.js:', function() {
             expect(toContainer).toEqual(expected);
         });
     });
+
+    describe('concat', function() {
+        var concat = Lib.concat;
+        it('works with multiple Arrays', function() {
+            expect(concat([1], [[2], 3], [{a: 4}, 5, 6]))
+                .toEqual([1, [2], 3, {a: 4}, 5, 6]);
```
etpinard: `Lib.concat` is fantastic :tada: 

Could we `spyOn(Array.prototype, 'concat')` in this test-case to make sure it is called in the "all regular arrays" case?

etpinard:
> how do I avoid making this mistake with `package-lock` again? Is there a way to ensure that these `optional` flags don't get wiped out when I `npm i`?

I'm voting for reverting all `optional: true` additions -> https://github.com/plotly/plotly.js/pull/3184#issuecomment-435926533

test/jasmine/tests/lib_test.js
alexcjohnson: Good call - https://github.com/plotly/plotly.js/pull/3211/commits/1ca72c39344c54f086bf7d5f728337322c6d68fe
`toEqual` calls `Array.concat` profusely, so I had to rearrange a little to get this to work right...

etpinard:
Nicely done :dancer:

Issue #3201, opened 2018-10-31
Plot schema suggests candlestick hoverinfo can be an array


cpsievert:
but it doesn't appear to like arrays -- https://codepen.io/cpsievert/pen/aReZpe

PR #3213, opened 2018-11-05
Finance traces hoverinfo arrayOk

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3201 and correct set `parcats` hoverinfo as `arrayOk:false` (cc @jonmmease )

cc @plotly/plotly_js


alexcjohnson:
Beautiful. 💃

Issue #3190, opened 2018-10-30
Pie text colors don't restyle/react


alexcjohnson:
See https://codepen.io/alexcjohnson/pen/jeogoy?editors=0010 - this edits `titlefont.color`, and you can see that it doesn't change from red to yellow as it should via `react`. The same is true via `restyle`, and of `(inside|outside)textfont.color`.

Thanks @nicolaskruchten!

Looks like pie overrides the text color `editType`:

https://github.com/plotly/plotly.js/blob/1819584564dcd378df366a8cfa6811ca670c0d4a/src/traces/pie/attributes.js#L21

But these colors are set during `Pie.plot`, not `Pie.style`.

alexcjohnson:
`bar` edits these colors correctly, but I notice it does so just by leaving the text font properties as [`calc` edits](https://github.com/plotly/plotly.js/blob/1819584564dcd378df366a8cfa6811ca670c0d4a/src/traces/bar/attributes.js#L19). It's important that family and size be `calc` (or perhaps `plot`?) for both of these types, because whether the text goes inside or outside can depend on its size relative to the bars as drawn... but color should be able to be moved to `style`. The challenge is just figuring out which font to use - it might not be obvious whether the text you're styling is inside or outside.

PR #3214, opened 2018-11-05
pie/bar textfont-color restyles

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3190

cc @plotly/plotly_js

### DIFF
--- a/src/traces/pie/attributes.js
+++ b/src/traces/pie/attributes.js
@@ -18,7 +18,7 @@ var extendFlat = require('../../lib/extend').extendFlat;
 var textFontAttrs = fontAttrs({
     editType: 'calc',
     arrayOk: true,
-    colorEditType: 'style',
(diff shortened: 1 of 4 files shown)


src/traces/pie/attributes.js
```diff
@@ -18,7 +18,7 @@ var extendFlat = require('../../lib/extend').extendFlat;
 var textFontAttrs = fontAttrs({
     editType: 'calc',
     arrayOk: true,
-    colorEditType: 'style',
+    colorEditType: 'plot',
```
alexcjohnson: Sure, would be nice someday to get `style` to work but this isn't a real high-traffic pathway (nor a heavy plot type, unless someone's really abusing it!)

alexcjohnson:
Great tests! 💃

Issue #1849, opened 2017-07-04
.animate method that can handle any type of figure diff


chriddyp:
In Dash, I'd like users to be able to toggle on and off an `animate` property which would either smoothly transition points (if available) or just update the chart.

Right now, `Plotly.animate` doesn't work with all chart transitions and data transitions. While I could program this myself in the Graph component, it would be great if `plotly.js` could do this so that everyone could benefit. We could even use this in the workspace to smooth out transitions!

Ideally, I would just replace the `.newPlot(id, data, layout)` call with a `.animate(id, data, layout)` call and call it a day.

In particular, here are some of the edge cases:
- `.animate` to work before `.plot` is called
- `.animate` to work across all chart types (even if it's not a smooth animation)
- `.animate` to autorange (expand axes if the new data was larger than the existing data)
- `.animate` when new traces are added or removed
- `.animate` when chart types have been changed

I'm OK if this is a new method or a new config option or whatever.

***

cc @rreusser @alexcjohnson @etpinard @cpsievert @monfera @jackparmer

etpinard:
> .animate to work across all chart types (even if an

_even if an_ :question:

etpinard:
> .animate to autorange (expand axes if the new data was larger than the existing data)

That's arguably not always the desired behavior. We'll probably need another attribute.

chriddyp:
> > .animate to autorange (expand axes if the new data was larger than the existing data)
> That's arguably not always the desired behavior. We'll probably need another attribute.

If `autosize=true`, I think that I'd argue that this is the expected behaviour. If the user wants to restrict the range to a certain view, they can set the range manually.

monfera:
I 👍 -d @etpinard's comment b/c sometimes expansion (or shrinking) is not the intent - indeed, keeping the range(s) constant is achievable by explicitly setting the range. I was thinking of a boolean `stickyness` attribute (true: retain the initial autodomain vs. false: expand/shrink as needed) but it may be unnecessary complication as the range can be directly set as you suggest. It's likely I didn't have the same thing in mind as @etpinard, just wanted to add my thought.

Btw. it might occur more often that data is _removed_, yet one wants to preserve the original full domain (eg. when eliminating points via crossfiltering), than that data is _added_, yet one would _not_ want to see them.

[...]

PR #3217, opened 2018-11-05
Transitions with Plotly.react

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1849 (hopefully).

Transitions ([don't call them animations](https://github.com/plotly/plotly.js/issues/1849#issuecomment-429122676)) are coming to `Plotly.react` :tada: 

In this iteration, I opted for a single `transition` option container. For example,

```js
var transition = {
   duration: 2000,
   easing: 'cubic-in-out'
}

var data = [{y: [1,2,1], marker: {size: 20}}]

var layout = {
  xaxis: {range: [-1, 3]},
  yaxis: {range: [0, 3]},
  transition: transition
}

Plotly.newPlot(gd, data, layout)

data[0].y = [2, 1, 2]
data[0].marker.size = 40
data[0].marker.color = 'red'
Plotly.react(gd, data, layout)
```

gives:

peek 2018-11-05 17-34

which appears a lot smoother in reality than in that gif.

----

Here are more examples (taken from https://plot.ly/javascript/animations/ but converted to using `Plotly.react`):

- Trace coordinate transitions: https://codepen.io/etpinard/pen/KrpMQa
- Axis range transitions: https://codepen.io/etpinard/pen/wQazrq?editors=1010
- Trace transition with object constancy: https://codepen.io/etpinard/pen/XybjZG?editors=1010

--------

In brief to get transitions working with `Plotly.react`, this PR:

- added `anim: true` to animatable attributes and picked them off during `react` diffing
- added a version of `Plots.transition` that assumes that `gd._fullData` / `gd._fullLayout` is the "new" state (N.B. `Plotly.animate` assumes that `gd._fullData` / `gd_fullLayout` is the "old" state)
- make that new `Plots.transition` figure out if `redraw` after completion is required from restyle / relayout flags
- track axes with `autorange:true` and altered ranges during `react` diffing to efficiently compute "layout" transitions
- diff `fullLayout` before `fullData` to track animatable `datarevision` changes

----

TODO:

- [ ] get general approval from `react-plotly.js` and `dash` folks
      + [ ] @nicolaskruchten
      + [ ] @chriddyp
      + [ ] @alexcjohnson 
- [x] add tests
- [x] DRY up `Plotly.transition` <--> `Plotly.transition2` **done in:** https://github.com/plotly/plotly.js/pull/3217/commits/1ed4e425232a8a88c6891683269c0dee83f7d79f

cc @plotly/plotly_js


chriddyp:
awesome! is there a prerelease build (on a cdn?) that I can try this out on? I recall that we were creating builds on  commits now

antoinerg:
@chriddyp https://17875-45646037-gh.circle-artifacts.com/0/plotly.min.js

chriddyp:
In Dash-land, for larger features like this, we usually create pre-releases and share them with the community for a couple of weeks to get feedback. If OK with plotly.js, I'd like to do the same with this PR. I will manage the community and I'll create a pre-release in `dash-core-components` from this PR.

I'm frequently surprised by what folks end up doing with abstract features like this and it's a nice, low-risk way to test the waters.

etpinard:
> If OK with plotly.js

If dash users are made aware of the limitations, mainly:

https://github.com/plotly/plotly.js/blob/1cb0d5f6712d2b46ff6e4ebe3352392fde3d7cc4/src/plots/plots.js#L2547-L2549

than yeah I'm ok with it.

etpinard:
Update from today's meeting: I'll attempt to add a flag that makes trace transitions "win out" over layout transitions.

etpinard:
> If the major use of this is object-constancy

Part of the examples I made :arrow_up: https://codepen.io/etpinard/pen/XybjZG?editors=1010

nicolaskruchten:
Ah the `ids` key ok. But what about trace-to-trace constancy?

nicolaskruchten:
Here's an example: https://codepen.io/nicolaskruchten/pen/pQddZm?editors=0010

In this case I would expect the red trace to vanish and the blue trace to translate to where the red one used to be, but I don't think I can express that right now.

etpinard:
`uid` should be enough in this case. No need for another attribute.

nicolaskruchten:
OK, so we should evolve this PR a bit to only transition matching-`uid`-traces? @chriddyp you're OK with requiring `uid` for transitioning to work? I think we have to, otherwise we just encourage gratuitous transitions and/or miss some important ones...

alexcjohnson:
Presumably this will work the same way as `uirevision` #3236 - if you don't provide `uid`, we assign one that's random at first but stays with traces of the same index. That machinery is nothing new. So internally we only need to key off `uid`, but to the user it looks like "`uid` if you provide that, index if you don't"

nicolaskruchten:
I could live with that. We still need to make my example work tho right? Or
did I do it wrong?

etpinard:
Transitions from `.animate` and `.react` don't support trace-in / trace-out tweening at the moment.

nicolaskruchten:
(per verbal conversation: I'm not talking about tweening in/out but rather trace-constancy)

etpinard:
yeah, getting https://codepen.io/nicolaskruchten/pen/pQddZm?editors=0010 to work shouldn't be too. I suspect something is up in the react diffing. 

https://codepen.io/etpinard/pen/yQpNLM?editors=0010 work ok if when the before and after `"data"` have the same number of traces and the same trace ordering.

nicolaskruchten:
Just so we don't forget: animations should track `uid` regardless of the order of the traces in subsequent `react()` calls, which isn't the case right now: https://codepen.io/nicolaskruchten/pen/yQpNWE?editors=0010

etpinard:
Dropping from 1.43.0.

etpinard:
**Update:** the trace "object-constancy" bugs are now fixed:

- data-length mismatch case: https://codepen.io/etpinard/pen/QzZdKN?editors=0010
- out-of-order case: https://codepen.io/etpinard/pen/jXeyMr?editors=0010

not much extra logic was needed (see commits https://github.com/plotly/plotly.js/pull/3217/commits/4ef048f36c9fe043ed4d2cc37f5109033b1d9f7b, https://github.com/plotly/plotly.js/pull/3217/commits/61ee4e97610105b08ab640a2197059d31eed3a2b and https://github.com/plotly/plotly.js/pull/3217/commits/33a9530e97cb36296d8be49429a22d5f17b55a3a for the details), d3's and the key function in this block here:

https://github.com/plotly/plotly.js/blob/05d0759b2628d8e0fb28a2317e22c3af547ab98e/src/traces/scatter/plot.js#L36-L37

did the rest.

nicolaskruchten:
@chriddyp can you elaborate on your comment on "transition revision" before we merge this guy in?

etpinard:
> Update from today's meeting: I'll attempt to add a flag that makes trace transitions "win out" over layout transitions.

Done in https://github.com/plotly/plotly.js/pull/3217/commits/8ab38be3d84a23902430e979a9a9123018642da4 via `transition.ordering: 'trace first'`. I'm not 100% sold on that attribute name though. Please let me know if you can think of a better name for it.

Example: https://codepen.io/etpinard/pen/jXdLgG

-------

With that :arrow_up: , I'll tag this PR as `reviewable`.

nicolaskruchten:
So... if I delete the "traces first" line, it seems to move and pan simultaneously, which is the ideal, right? I thought that wasn't possible somehow?

etpinard:
> it seems to move and pan simultaneously, which is the ideal, right? I thought that wasn't possible somehow?

That's way layout transitions (like https://plot.ly/javascript/animations/#animating-the-layout) always do.

etpinard:
... you can't animate marker color/width though during those layout transition at the moment though.

nicolaskruchten:
Ah I see, the blue/black transition was so subtle I missed it. Adding a size transition as well makes it clearer what the tradeoff is.

etpinard:
@archmoj is going to review this one.

src/plots/plots.js
```diff
@@ -2360,8 +2367,214 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
...
+        var fullLayout = gd._fullLayout;
+        var basePlotModules = fullLayout._basePlotModules;
+        var i;
+
+        var axisTransitionOpts;
+        var traceTransitionOpts;
+        var transitionedTraces;
+
+        var allTraceIndices = [];
+        for(i = 0; i < fullData.length; i++) {
```
archmoj: Whereas `i` is used in two functions below, would you mind using another variable name for this loop?

src/plot_api/plot_api.js
```diff
@@ -2913,10 +2966,12 @@ function getDiffFlags(oldContainer, newContainer, outerparts, opts) {
...
         // short-circuit based on previous calls or previous keys that already maximized the pathway
-        if(flags.calc) return;
+        if(flags.calc && !opts.transition) return;
 
         var oldVal = oldContainer[key];
         var newVal = newContainer[key];
+        var parts = outerparts.concat(key);
+        astr = parts.join('.');
 
         if(key.charAt(0) === '_' || typeof oldVal === 'function' || oldVal === newVal) continue;
```
archmoj: Wondering could we detect some of these cases a bit earlier?
For example `key` may be tested at the started of the loop.

src/plot_api/plot_api.js
```diff
@@ -2913,10 +2966,12 @@ function getDiffFlags(oldContainer, newContainer, outerparts, opts) {
 
     for(key in oldContainer) {
         // short-circuit based on previous calls or previous keys that already maximized the pathway
-        if(flags.calc) return;
+        if(flags.calc && !opts.transition) return;
```
archmoj: Can we move these tests up or down?

src/plots/cartesian/transition_axes.js
```diff
@@ -169,79 +65,71 @@ module.exports = function transitionAxes(gd, newLayout, transitionOpts, makeOnCo
...
-            axis.range[1] = r0[1] * (1 - progress) + progress * r1[1];
+        var editX = !!xr1;
+        var editY = !!yr1;
+        var viewBox = [];
 
-            viewBox[2] = subplot.xaxis._length * ((1 - progress) + progress * dx2 / dx1);
+        if(editX) {
+            var dx0 = xr0[1] - xr0[0];
+            var dx1 = xr1[1] - xr1[0];
+            viewBox[0] = (xr0[0] * (1 - progress) + progress * xr1[0] - xr0[0]) / (xr0[1] - xr0[0]) * xlen;
```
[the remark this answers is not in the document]
etpinard: > Could we check to avoid division by zero here

No need. Updates with equal ranges (`xr1` and `xr0`, here) never get here.

src/plot_api/plot_api.js
etpinard: Why?

src/plot_api/plot_api.js
etpinard: I don't see the need.

src/plots/plots.js
etpinard: Cleaned up in https://github.com/plotly/plotly.js/pull/3217/commits/ec4d8b9250c1a0b598bc960cbc777d2d996e2ee2

src/plot_api/plot_api.js
```diff
@@ -2843,31 +2862,41 @@ function diffData(gd, oldFullData, newFullData, immutable) {
...
+            getDiffFlags(oldFullData[i]._fullInput, trace, [], diffOpts);
+        }
     }
 
     if(flags.calc || flags.plot) {
         flags.fullReplot = true;
     }
 
+    if(transition && flags.nChanges && flags.nChangesAnim) {
+        flags.anim = flags.nChanges === flags.nChangesAnim && sameTraceLength ? 'all' : 'some';
```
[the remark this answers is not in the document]
etpinard: Sure -> https://github.com/plotly/plotly.js/pull/3217/commits/ec4d8b9250c1a0b598bc960cbc777d2d996e2ee2

src/plot_api/plot_api.js
archmoj: First I thought it may be independent from the `key` and loop.
Resolved.

src/plots/cartesian/transition_axes.js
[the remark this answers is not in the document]
archmoj: But what if `xr0[0] === xr0[1]`?
Is that included in the test?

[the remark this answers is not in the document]
etpinard: Not this one, but this one done here:

https://github.com/plotly/plotly.js/blob/3b9188df7b2cd6c57e9941cffa10555e36ce6fc3/src/plots/cartesian/set_convert.js#L379-L443

which is called early during the axis defaults.

etpinard:
Thanks for the review. I think the failing test is due to the joyplot PR and the boxpoint hover fix not getting along. I'll fix that later this morning. Nothing to worry!

Issue #3083, opened 2018-10-06
Please publish locales on NPM


brylie:
The Plotly.js locales are not included in the primary NPM package, probably to keep the package size small. This makes it difficult to include the locales in projects that import Plotly.js from NPM. Please publish the Plotly.js locales as one or more packages on NPM, so they can easily be included in packages that use NPM to manage the Plotly.js dependency.

etpinard:
By "primary", I suspect you mean https://www.npmjs.com/package/plotly.js-dist ?

brylie:
Jep. There is a fork on NPM that includes locales, but it hasn't been updated in a year. It might be worth separating locales into their own packages, so they can be selectively included in projects where only a few languages are needed.

etpinard:
Interesting idea @brylie , I thought about adding a `plotly.js-topojson` during https://github.com/plotly/plotly.js/pull/2670, but yeah locale could benefit from having their own npm packages as well.

Locales files aren't that big, so I'm thinking of publishing a single `plotly.js-locales` package and then ask users to require the correct file e.g. `require('plotly.js-locales/fr.js')`.

brylie:
What might it look like using `import`?

PR #3223, opened 2018-11-07
Publish `plotly.js-locales` to npm

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3083, a continuation from https://github.com/plotly/plotly.js/pull/2670 (the partial-bundle-pkgs PR)

As locale files are fairly small, I decided to publish only one `plotly.js-locales` package (as opposed to one per locale) and ask users (in the `plotly.js-locales` README) to grab to relevant locale as:

```js
// ES6 module
import Plotly from 'plotly.js';
import locale from 'plotly.js-locales/fr';

// CommonJS
var Plotly = require('plotly.js');
var locale = require('plotly.js-locales/fr');

// then
Plotly.register(locale);
```

that is, users would npm-download all the locales, but only bundle the locale(s) they need.

Reviewers can look at:

https://github.com/plotly/plotly.js/compare/publish-locales...publish-locales-build

to see to generated git-ignored files in `build/`.

cc @plotly/plotly_js and @nicolaskruchten who may find this useful for RCE and dash users.

### DIFF
--- a/tasks/sync_packages.js
+++ b/tasks/sync_packages.js
@@ -205,6 +205,7 @@ function syncLocalesPkg(d) {
             '',
             '// then',
             'Plotly.register(locale);',
+            'Plotly.setPlotConfig({locale: \'fr\'})',
             '```',
             '',
             copyrightAndLicense
--- a/tasks/sync_packages.js
+++ b/tasks/sync_packages.js
@@ -2,6 +2,7 @@ var path = require('path');
 var fs = require('fs-extra');
 var exec = require('child_process').exec;
 var runSeries = require('run-series');
+var glob = require('glob');
 
 var common = require('./util/common');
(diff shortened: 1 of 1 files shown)


tasks/sync_packages.js
```diff
@@ -205,6 +205,7 @@ function syncLocalesPkg(d) {
             '',
             '// then',
             'Plotly.register(locale);',
+            'Plotly.setPlotConfig({locale: \'fr\'})',
```
alexcjohnson: Good call 👍

alexcjohnson:
@etpinard this looks great! Assuming you've verified that the built files work as described in the README (and they certainly look like they will) 💃

etpinard:
> Assuming you've verified that the built files work as described

Verified :white_check_mark:

PR #3230, opened 2018-11-08
upgrade of modules that use glsl-out-of-range 1.0.3

archmoj:
This PR is the follow up of a readme typo issue (related to npm install description) within `glsl-out-of-range` module as reported and fixed here https://github.com/plotly/glsl-out-of-range/pull/1.
New gl-vis modules are published in order to use the latest `glsl-out-of-range`.
@alexcjohnson 
@etpinard


etpinard:
:dancer: though you probably didn't need to publish a new version for patch in the `README.md`.

PR #3231, opened 2018-11-08
Bump browserify

etpinard:
to resolve https://github.com/plotly/plotly.js/network/alert/package-lock.json/cached-path-relative/open

where now

image


alexcjohnson:
💃

PR #3233, opened 2018-11-08
Issue 3224

### LINKED ISSUE #3224: 3d Line Plot - Uncaught ax.dtick error: NaN
This can be seen on the 3d Line Plot demo by zooming in and swirling the mouse around wildly.

The first example on this page: https://plot.ly/javascript/3d-line-plots/

Plotly will freeze up and print the following to the console: Uncaught ax.dtick error: NaN

archmoj:
Fixes #3224 by checking NaN values not to be used as ax.dtick
@etpinard

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -755,6 +755,9 @@ axes.autoTicks = function(ax, roughDTick) {
     // prevent infinite loops
     if(ax.dtick === 0) ax.dtick = 1;
 
+    // prevent issue https://github.com/plotly/plotly.js/issues/3224
+    if (Number.isNaN(ax.dtick)) ax.dtick = 1;
+
     // TODO: this is from log axis histograms with autorange off
     if(!isNumeric(ax.dtick) && typeof ax.dtick !== 'string') {
         var olddtick = ax.dtick;
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -756,7 +756,7 @@ axes.autoTicks = function(ax, roughDTick) {
     if(ax.dtick === 0) ax.dtick = 1;
 
     // prevent issue https://github.com/plotly/plotly.js/issues/3224
-    if (Number.isNaN(ax.dtick)) ax.dtick = 1;
+    if(Number.isNaN(ax.dtick)) ax.dtick = 1;
 
     // TODO: this is from log axis histograms with autorange off
     if(!isNumeric(ax.dtick) && typeof ax.dtick !== 'string') {


src/plots/cartesian/axes.js
```diff
@@ -755,6 +755,9 @@ axes.autoTicks = function(ax, roughDTick) {
     // prevent infinite loops
     if(ax.dtick === 0) ax.dtick = 1;
 
+    // prevent issue https://github.com/plotly/plotly.js/issues/3224
+    if(Number.isNaN(ax.dtick)) ax.dtick = 1;
```
etpinard: Do we know under which circumstance `ax.dtick` comes through as `NaN` here?

etpinard:
> @etpinard Could you please let me know where to put a jasmine test for this bug?

Somewhere at the end of this `describe` block

https://github.com/plotly/plotly.js/blob/2009f0988fad6e0f281b941178b82e5162281c9b/test/jasmine/tests/gl3d_plot_interact_test.js#L24

src/plots/cartesian/axes.js
etpinard: > It could simply be triggered for example using gl3d_ibm-plot by panning the graph to outside.

Hmm. I can't replicate. Can you take a screenshot of 

```js
gd._fullLayout.scene.camera.eye
```

in the console once after that error appears.

-----

> I am wondering now that we may need to replace the condition checks in few other places there too?

I'm actually thinking the opposite: we shouldn't be calling `Axes.autoTicks` in the first place whenever `ditck===NaN`. But first, I'll like to know how this happens.

alexcjohnson: > I'm actually thinking the opposite: we shouldn't be calling `Axes.autoTicks` in the first place whenever `ditck===NaN`. But first, I'll like to know how this happens.

Agreed. For one thing, `dtick=1` would draw [1000 ticks](https://github.com/plotly/plotly.js/blob/2009f0988fad6e0f281b941178b82e5162281c9b/src/plots/cartesian/axes.js#L561) on date or otherwise large-range axes which would bog things down and look ugly. What's `roughDTick` when we get `NaN` here, and why?

archmoj: @etpinard here is a [demo](https://codepen.io/MojtabaSamimi/pen/bQpvzJ)

archmoj: And this is [another example](https://codepen.io/MojtabaSamimi/pen/KrzRVd) example now without date and logs.

etpinard: Right, this is consistent with what I was thinking: we get those `Uncaught ax.dtick error: NaN` errors in the console in situations where ticks can't be seen.

etpinard:
Great fix :dancer:

src/plots/gl3d/layout/tick_marks.js
```diff
@@ -47,7 +47,9 @@ function computeTickMarks(scene) {
         axes._length = (glRange[i].hi - glRange[i].lo) *
             glRange[i].pixelsPerDataUnit / scene.dataScale[i];
 
-        if(Math.abs(axes._length) === Infinity) {
+        if(Math.abs(axes._length) === Infinity ||
+           Math.abs(axes._length) === -Infinity ||
```
alexcjohnson: wait, how can `Math.abs(...) === -Infinity`?

archmoj: Thanks for the note. Just removed that line.

alexcjohnson: Thanks. Anyway much better to fix the issue here than in `axes.js`!

💃

caseymarquis:
I've updated to the latest, and the issue still occurs. You can reproduce by zooming in the maximum amount and then rotating rapidly:
https://codepen.io/anon/pen/xQORpB

To reproduce this in an automated way, I think you would need to force a rapid camera rotation from code, which I do not know enough about plotly to do. (I also posted this on the original issue, but I'm not sure if anyone is looking at comments on closed issues.)

PR #3234, opened 2018-11-09
Joyplots2

Kully:
Rebasing from [this commit](0dea0e9a12d5248d492289a61138f6a7706ecba0) from original Joyplots PR https://github.com/plotly/plotly.js/pull/3109


Kully:
to do:
- [ ] vpads work for jitter != 0
- [ ] vpads fully reveal all points (not half of the points)

Kully:
```javascript
function violin(x, y0, fillcolor, width, side, pointpos) {
    return {
        type: 'violin',
        width: width,
        x: x,
        points: 'all',
        jitter: 0,
        pointpos: pointpos,
        side: side,
        box: {
          visible: false,
        },
        boxpoints: false,
        line: {
          color: 'black',
        },
        fillcolor: fillcolor,
        opacity: 0.6,
        meanline: {
          visible: true
        },
        y0: y0
    };
}

var array = [];
var length = 20
for(i = 0; i < length; i++) {
  array.push(12 + 0.5 * i);
}

var data = [
  violin([0, 5, 7, 8], 0.0, '#8dd3c7', 0, 'positive', 1),
  violin([20, 25, 27, 28], 0.0, '#8dd3c7', 1, 'positive', 1),
];

var layout = {
  title: "Joyplot - Violin with multiple widths",
  xaxis: {
    zeroline: false
  },
  violingap: 0
};

```

newplot

test/image/mocks/violin_positive_and_negative.json
```diff
@@ -0,0 +1,35 @@
...
+            "pointpos": 0,
+            "jitter": 0,
+            "x": [20, 25, 27, 28],
+            "side": "negative",
+            "line": {
+                "color": "black"
+            },
+            "fillcolor": "#d3c78d",
+            "opacity": 0.6,
+            "y0": 0.0
```
etpinard: This one here appears to have too much padding above and below.

src/traces/box/cross_trace_calc.js
```diff
@@ -90,22 +95,30 @@ function setPositionOffset(traceType, gd, boxList, posAxis, pad) {
...
+
+        var trace = calcTrace[0].trace;
+        var vpadminus_pos_side = (trace.pointpos <= 0) ? (-trace.pointpos) * padfactor * fullLayout[numKey] : 0;
+        var vpadplus_neg_side = (trace.pointpos >= 0) ? trace.pointpos * padfactor * fullLayout[numKey] : 0;
+        var side = calcTrace[0].trace.side;
+        var vpadminus = (side === 'positive') ? vpadminus_pos_side + thisDPos * 0.05 : (thisDPos + pad[0] * padfactor);
+        var vpadplus = (side === 'negative') ? vpadplus_neg_side + thisDPos * 0.05 : (thisDPos + pad[1] * padfactor);
+        var extremes = Axes.findExtremes(posAxis, positions, {
+            vpadminus: vpadminus,
+            vpadplus: vpadplus
```
etpinard: let's use the [`padded: true`](https://github.com/plotly/plotly.js/blob/ad937d20a83fb3d183916e715c1a3aac6cef5921/src/plots/cartesian/autorange.js#L294-L295) option instead of those adding `* 0.05` in `vpadminus` and `vpadplus` above.

etpinard:
Hmm. Something is still off:

```js
function violin(x, y0, fillcolor, width, side, pointpos) {
    return {
        type: 'violin',
        width: width,
        x: x,
        points: 'all',
        jitter: 0,
        pointpos: pointpos,
        side: side,
        line: {
          color: 'black',
        },
        fillcolor: fillcolor,
        opacity: 0.6,
        meanline: {
          visible: true
        },
        y0: y0
    };
}

var array = [];
var length = 20
for(i = 0; i < length; i++) {
  array.push(12 + 0.5 * i);
}

var data = [
  violin([0, 5, 7, 8], 0.0, '#8dd3c7', 0, 'positive', 2),
  violin([20, 25, 27, 28], 0.0, '#8dd3c7', 1, 'positive', 0),
];

var layout = {
  title: "Joyplot - Violin with multiple widths",
  xaxis: {
    zeroline: false
  },
  violingap: 0
};

```

gives

image


... we don't see the pts from the first trace.

Kully:
> we don't see the pts from the first trace.

I think that this was broken before this PR - traces with pointpos 2 are not visible cause they are too high and out of the scope.

etpinard:
Here's my latest attempt:

https://github.com/plotly/plotly.js/compare/joyplots2...joyplots2-etpinard

- I haven't fixed up the jasmine tests,
- we could probably improve the axis-expand algo for `scalemode: 'count'` traces
- single-box and single-violin graphs may look less-than-optimal to some, e.g.

image

@Kully would you mind testing my branch out, before I continue?

etpinard:
OK @Kully I pushed my commits from the now-deleted `joyplot2-etpinard` branch here.

Before fixing up the tests, I think we should spend some time:

- to come up with a special case for single-box/violin subplots to have them better match the baselines on `master`
- investigate what can be done to remove the extra padding on [`violin_side-by-side`](https://rreusser.github.io/plotly-mock-viewer/#violin_side-by-side)

before adapting the tests to the new autorange results.

---------------

I should have time to look at this again later this week (Thursday or Friday), but feel free to try out to :white_check_mark: the two items above in the meantime.

etpinard:
Dropping from  1.43.0 (unless @Kully has some unpushed commits that address https://github.com/plotly/plotly.js/pull/3234#issuecomment-441828927).

Kully:
> unless @Kully has some unpushed commits that address

I do not

etpinard:
@plotly/plotly_js tagging this as `reviewable`. It's about time!

The improved algo is in https://github.com/plotly/plotly.js/pull/3234/commits/bf6cfa2adf15d82b5f33613f65f12fd96286e09f which :hocho: all our concerns from https://github.com/plotly/plotly.js/pull/3234#issuecomment-441310856 (I think :wink: ).

Jasmine tests are now passing thanks to https://github.com/plotly/plotly.js/pull/3234/commits/12ff323156830f777b84974c08c5c43bc6c9e591

etpinard:
@Kully could you take a few minutes this week to double-check my latest attempt?

test/image/mocks/violin_box_multiple_widths.json
```diff
@@ -0,0 +1,44 @@
...
+            "side": "positive",
+            "line": {
+                "color": "black"
+            },
+            "fillcolor": "#c78dd3",
+            "opacity": 0.6,
+            "y0": 0.2
+    }],
+    "layout": {
+        "title": "Joyplot - Violin with multiple widths",
```
nicolaskruchten: Let's get rid of the name "joyplot" here if possible, in favour of "ridgeplot"

etpinard: Ok done in https://github.com/plotly/plotly.js/pull/3234/commits/c1d6e76e064ec17afcfacee95c4634cbd08c7195

etpinard:
@antoinerg this PR is a bit of a mess, but could I ask you to review it before next Tuesday (the 22nd)?

antoinerg:
@etpinard Looks very nice!

However, I noticed that the axis range does not always update properly. For example, for the mock `violin_box_multiple_widths.json`, when I hide the box by clicking on its name on the legend I end up with:
violin_box_multiple_widths_unselect_box
As you can see, one of the violin is clipped. Is this the expected behavior?

etpinard:
> when I hide the box by clicking on its name on the legend I end up with:


Thanks. Fixed in a237252f330ab2439176d16abbba4a206043b0e1

antoinerg:
Nice fix :ok_hand: !

Another question about the current behavior for automatic width: we resize the violins even when the axis range isn't modified. For example, the `auto` trace goes from small to very large upon hiding trace `with: 0.4`:
screenshot_2019-01-21_15-02-57_01
screenshot_2019-01-21_15-03-21_02

I just want to make sure this is what we want. Maybe we should add a Jasmine test to :lock: that down?

@etpinard let me know what you think!

etpinard:
> Maybe we should add a Jasmine test to  that down?

That change in violin width comes from 

https://github.com/plotly/plotly.js/blob/a237252f330ab2439176d16abbba4a206043b0e1/src/traces/box/cross_trace_calc.js#L73-L74

where we have two distinct violin position vals at first, and then just once after that `restyle` call.

Here's the new test: ef52b8c6e

antoinerg:
Great job @etpinard ! :dancer:

etpinard:
Thanks for catching the bug from https://github.com/plotly/plotly.js/pull/3234#issuecomment-456187108 @antoinerg !

etpinard:
In brief, this PR:

- adds `width` to box and violin traces
- improves box/violin autorange calculations

PR #3236, opened 2018-11-09
Preserve ui state across Plotly.react redraws

alexcjohnson:
For https://github.com/plotly/dash-core-components/issues/198

- Track gui-driven changes via internal-use variants of `restyle`/`relayout`/`update` -> `_guiRestyle` etc that wrap the original methods to record the original state.
- Add a class of `uirevision` attributes (that all inherit directly or indirectly from `layout.uirevision` so in the simplest case you can just use that and forget the others) that can be used to maintain the state user interaction has left the plot in. If `uirevision` exists and did not change, any attribute the user has modified via the built-in GUI will keep the user-specified value *unless* it has a new explicit value in the new figure.

cc @chriddyp @nicolaskruchten 

@etpinard best to review commit-by-commit. The first six (and the [eighth](https://github.com/plotly/plotly.js/commit/10ddbb9c67c4e9abf1b8a1f51cd0ada340f0f7db) as well as test reorg in https://github.com/plotly/plotly.js/commit/a4437472023bbe49f22e0b33f7d50561227ba766) are pre-cleaning and various obscure bugfixes. The meat of the changes are in https://github.com/plotly/plotly.js/commit/a11ec44f388741dd21046d08a782e4e76fc8ba22 (tracking changes), https://github.com/plotly/plotly.js/commit/4af2bf8b6e8b190b3f253934cee51ec7a90f7acd (adding the new attributes & inheritance), and https://github.com/plotly/plotly.js/commit/03baca7efe26c130fbcc559e9724930b79bbbc77 (responding to the changes in `Plotly.react`). I missed `selectedpoints` in the first iteration, that's added in https://github.com/plotly/plotly.js/commit/6a9d0d78e9d7da0882ec8d9fa8214f4fc805cf69, and tests are added in the last commit https://github.com/plotly/plotly.js/commit/54304b40aee7c97a81b64964b67123ac05b5eb82.


test/jasmine/assets/custom_assertions.js
```diff
@@ -130,35 +135,63 @@ exports.assertHoverLabelContent = function(expectation, msg) {
...
-            bboxes.sort(function(a, b) {
-                return (b.bbox.left + b.bbox.right - a.bbox.left - a.bbox.right) / 2;
+            o2.forEach(function(item, i) {
+                item.order = i;
+                delete item.bbox;
             });
-            expect(bboxes.map(function(d) { return d.index; })).toEqual(expectation.hOrder);
         }
+        observed.sort(labelSorter);
+        expected.sort(labelSorter);
```
alexcjohnson: A couple of hover label tests - one box and one violin I think - were failing on my machine and had been for months. AFAICT the order I see is the correct one, and yet on CI (and I guess on other folks' linux boxes) we get a different order for some reason. In the end we don't care about the order, except in the tests where we explicitly provide an order in which case it's not based on the selection order but on the bounding box order... so I made the change here to compare observed and expected labels in an order-independent way.

test/jasmine/tests/lib_test.js
```diff
@@ -2631,9 +2631,8 @@ describe('Queue', function() {
             return Plotly.relayout(gd, 'updatemenus[0]', null);
         })
         .then(function() {
-            // buttons have been stripped out because it's an empty container array...
```
alexcjohnson: turns out this was a mistake, applying our logic self-consistently we should *not* delete this array.

src/plot_api/plot_api.js
```diff
@@ -2074,7 +2072,7 @@ function _relayout(gd, aobj) {
                 if(manageArrays.isAddVal(vi)) {
                     undoit[ai] = null;
                 } else if(manageArrays.isRemoveVal(vi)) {
-                    undoit[ai] = obji;
+                    undoit[ai] = (nestedProperty(layout, arrayStr).get() || [])[i];
```
alexcjohnson: We might not have anyone who cares about undo/redo anymore, but we have it so here's a fix. We were missing the `.get()` the way this was written before.

src/components/modebar/buttons.js
```diff
@@ -370,53 +370,47 @@ modeBarButtons.hoverClosest3d = {
...
     var axes = ['xaxis', 'yaxis', 'zaxis'];
-    var spikeAttrs = ['showspikes', 'spikesides', 'spikethickness', 'spikecolor'];
 
     // initialize 'current spike' object to be stored in the DOM
     var currentSpikes = {};
-    var axisSpikes = {};
     var layoutUpdate = {};
 
     if(val) {
-        layoutUpdate = Lib.extendDeep(layout, val);
```
alexcjohnson: I guess this worked, but it wasn't pretty or efficient, calling relayout with an entire copy of the layout object! Anyway it would not work at all with `uirevisions`, as then *everything* would look like it got edited by the GUI.

src/components/modebar/buttons.js
```diff
@@ -526,12 +522,10 @@ modeBarButtons.toggleHover = {
...
     click: function(gd, ev) {
-        toggleHover(gd);
+        var layoutUpdate = getNextHover3d(gd, ev);
+        layoutUpdate.hovermode = getNextHover(gd);
 
-        // the 3d hovermode update must come
-        // last so that layout.hovermode update does not
-        // override scene?.hovermode?.layout.
-        handleHover3d(gd, ev);
+        Registry.call('relayout', gd, layoutUpdate);
```
alexcjohnson: combined into one call so it's a) more efficient and b) not order-dependent

src/core.js
```diff
@@ -29,7 +29,8 @@ var plotApi = require('./plot_api');
 var methodNames = Object.keys(plotApi);
 for(var i = 0; i < methodNames.length; i++) {
     var name = methodNames[i];
-    exports[name] = plotApi[name];
+    // _ -> private API methods, but still registered for internal use
+    if(name.charAt(0) !== '_') exports[name] = plotApi[name];
```
alexcjohnson: For this PR I chose to draw a distinction between GUI-driven changes - which we record so they can be persisted later - and other random `restyle` and `relayout` calls. My thinking was that other calls are explicitly being invoked by the developer, so they can also decide for themselves whether to persist them by taking into account the accompanying setting when the figure is next updated.

This line prevents exposing these variants (as `Plotly._guiRelayout` etc) but I guess I could see an argument for allowing app developers to access them. Say for example they want to extend our modebar (or make other similar controls) with things like a log/linear axis control. If they did this with `_guiRelayout` it could be all self-contained in the plot widget rather than tied into the higher-level logic about generating the figure. That said in this particular case it could lead to subtle problems since the form of `axis.range` is different for log and linear axes, so a figure that doesn't know what kind of axis it's plotting onto could supply a faulty explicit range...

So after all that I chose not to expose this to app developers, but I'm happy to discuss.

src/components/rangeslider/draw.js
```diff
@@ -286,7 +286,10 @@ function setDataRange(rangeSlider, gd, axisOpts, opts) {
         dataMax = clamp(opts.p2d(opts._pixelMax));
 
     window.requestAnimationFrame(function() {
-        Registry.call('relayout', gd, axisOpts._name + '.range', [dataMin, dataMax]);
+        var update = {};
+        update[axisOpts._name + '.range[0]'] = dataMin;
+        update[axisOpts._name + '.range[1]'] = dataMax;
+        Registry.call('_guiRelayout', gd, update);
```
alexcjohnson: In an earlier iteration I was requiring that all update values be primitives - no arrays or objects, because there could be conflicts between this and other updates that do break it down. For example, first you update `xaxis.range` and store its initial value as the `preGUI` value; then later you update `xaxis.range[0]`, we see that this doesn't have a `preGUI` value so we store it - but what we're actually storing is the previous *post*GUI value (and anyway we have two conflicting values stored). However now this is taken care of in [`storeCurrent`](https://github.com/plotly/plotly.js/pull/3236/commits/a11ec44f388741dd21046d08a782e4e76fc8ba22#diff-2941ab69a12080c0633ff4ac8ea3aa83R1419) so we could in principle put this call back to the more compact form it had before.

src/plots/gl2d/scene2d.js
```diff
@@ -313,27 +313,21 @@ proto.updateRefs = function(newFullLayout) {
...
+    // make a meaningful value to be passed on to possible 'plotly_relayout' subscriber(s)
+    var update = {};
+    update[xaxis._name + '.range'] = xaxis.range.slice();
+    update[yaxis._name + '.range'] = yaxis.range.slice();
+    update[xaxis._name + '.autorange'] = xaxis.autorange;
+    update[yaxis._name + '.autorange'] = yaxis.autorange;
 
-    update[xaxis._name] = xaxis.range.slice(0);
-    update[yaxis._name] = yaxis.range.slice(0);
+    Registry.call('_storeDirectGUIEdit', graphDiv.layout, graphDiv._fullLayout._preGUI, update, true);
```
alexcjohnson: This is a bit confusing... the `true` last arg means "apply this change to the input object" (ie `gd.layout`). But in the end this is the only caller that used that argument, everyone else (among the callers that short-circuit `restyle/relayout` so need to use `_storeDirectGUIEdit`) had some other way of applying the change... so rather than making it more explicit I think I'll just :hocho: that arg and apply the change here.

alexcjohnson: > I think I'll just 🔪 that arg and apply the change here.

done in https://github.com/plotly/plotly.js/pull/3236/commits/86b90d4e00ca98d2e0aa7d689d561548c258c7ff
Anyway only used for `pointcloud` at this point, which is mostly obsolete and partly broken but at least it's not *more* broken now...

src/plots/plots.js
```diff
@@ -479,6 +479,24 @@ plots.supplyDefaults = function(gd, opts) {
     // relink functions and _ attributes to promote consistency between plots
     relinkPrivateKeys(newFullLayout, oldFullLayout);
 
+    // For persisting GUI-driven changes in layout
+    // _preGUI and _tracePreGUI were already copied over in relinkPrivateKeys
+    if(!newFullLayout._preGUI) newFullLayout._preGUI = {};
+    // track trace GUI changes by uid rather than by trace index
```
alexcjohnson: 👀 @chriddyp 

This is important, as it has implications for how you use this feature: GUI changes to a trace (mostly just `visible: 'legendonly'` and `selectedpoints` but there's also `dimensions[i].constraintrange` in `parcoords`, and `name` and `colorbar.title` in editable mode) are tracked by the trace's `uid`, NOT by index.

That doesn't mean it will break if you don't provide a `uid` - since #2681 `uid` is provided consistently from call to call, so without `uid` we'll implicitly fall back on index. But if you want to ensure visibility follows traces of the same *meaning*, even if they change index, just provide a `uid` for that trace.

This could happen for example if you have a multiselect box, or bank of checkboxes, for which traces to include in a plot, and the user selects or deselects an earlier item after already having selected a later one.

src/traces/parcoords/plot.js
```diff
@@ -103,7 +119,17 @@ module.exports = function plot(gd, cdparcoords) {
...
-        gd.emit('plotly_restyle');
+        // TODO: we can't really store this part of the interaction state
+        // directly as below, since it incudes data arrays. If we want to
+        // persist column order we may have to do something special for this
+        // case to just store the order itself.
+        // Registry.call('_storeDirectGUIEdit',
+        //     gd.data[inputIndices[i]],
+        //     fullLayout._tracePreGUI[gd._fullData[fullIndices[i]]._fullInput.uid],
+        //     {dimensions: gdDimensions[i]}
+        // );
```
alexcjohnson: Feels like the better way to handle this is to create a `dimensionorder` attribute, so we can just store that, rather than directly mutating the `dimensions` array.

Like `columnorder` in `table` traces, but [that feature isn't fully built out either](https://github.com/plotly/plotly.js/pull/3236/commits/a11ec44f388741dd21046d08a782e4e76fc8ba22#diff-1e0dea3a9361823fe80cbbdc50befd7eR591) so I left both of these as open items for now.

test/jasmine/tests/colorbar_test.js
```diff
@@ -354,5 +355,49 @@ describe('Test colorbar:', function() {
...
+            })
+            .then(function() {
+                expect(gd.data[0].colorbar.x).toBeWithin(0.591, 0.01);
+                expect(gd.data[0].colorbar.y).toBeWithin(0.045, 0.01);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('can drag marker-level colorbars in editable mode', function(done) {
```
alexcjohnson: root-level dragging worked before (the above test), but was not 🔒 by a test. Dragging `marker.colorbar` (this test) was broken.

src/plot_api/plot_api.js
```diff
@@ -2362,6 +2362,193 @@ exports._guiRestyle = guiEdit(restyle);
...
+    {pattern: /^(scene\d*)\.camera/},
+    {pattern: /^(geo\d*)\.(projection|center)/},
+    {pattern: /^(ternary\d*\.[abc]axis)\.(min|title)$/},
+    {pattern: /^(polar\d*\.(radial|angular)axis)\./},
+    {pattern: /^(mapbox\d*)\.(center|zoom|bearing|pitch)/},
+
+    {pattern: /^legend\.(x|y)$/, attr: 'editrevision'},
+    {pattern: /^(shapes|annotations)/, attr: 'editrevision'},
+    {pattern: /^title$/, attr: 'editrevision'}
+];
```
alexcjohnson: This is a bit annoying, having to explicitly match attribute patterns to `uirevision` attributes, but I thought it was important for example to control all the spikes and hover/drag modes from `modebar.uirevision` and not from all the separate `axis` and `scene.uirevision` attributes. But we might be able to simplify it to only explicitly include patterns when the corresponding `uirevision` attribute is not directly adjacent to the edited attribute or one of its parents...

src/plot_api/plot_api.js
```diff
@@ -2362,6 +2362,193 @@ exports._guiRestyle = guiEdit(restyle);
...
+        var match = key.match(spec.pattern);
+        if(match) {
+            return {head: match[1], attr: spec.attr, autofill: spec.autofill};
+        }
+    }
+}
+
+// We're finding the new uirevision before supplyDefaults, so do the
+// inheritance manually. Note that only `undefined` inherits - other
+// falsy values are returned.
```
alexcjohnson: Otherwise we'd need to call `supplyDefaults` just to get the `uirevision` attributes, then if anything changed due to the GUI edits, call it all over again. The inheritance is simple enough that I just replicated it here.

src/plot_api/plot_api.js
```diff
@@ -2362,6 +2362,193 @@ exports._guiRestyle = guiEdit(restyle);
...
+                // TODO: This test for undefined is to account for the case where
+                // the value was filled in automatically in gd.layout,
+                // like axis.range/autorange. In principle though, if the initial
+                // plot had a value and the new plot removed that value, we would
+                // want the removal to override the GUI edit and generate a new
+                // auto value. But that would require figuring out what value was
+                // in gd.layout *before* the auto values were filled in, and
+                // storing *that* in preGUI... oh well, for now at least I limit
+                // this to attributes that get autofilled, which AFAICT among
+                // the GUI-editable attributes is just axis.range/autorange.
```
alexcjohnson: A fairly involved and kind of strange edge case, but I still may try to address it before we merge this:
- the initial plot has an explicit range
- the user zooms/pans to a different range
- `react` is called again, with the SAME `uirevision` but the explicit range is removed (this step seems weird to me, but it could happen...)

The correct behavior here would be to autorange, but currently we instead keep the GUI-defined range, since we can't tell that the original pre-GUI range wasn't actually something we filled in automatically during plotting. If we ever get around to getting rid of that late `gd.layout` mutation this problem will go away.

src/plots/cartesian/select.js
```diff
@@ -666,6 +666,15 @@ function isOnlyOnePointSelected(searchTraces) {
...
     var i, searchInfo, cd, trace;
 
+    // before anything else, update preGUI if necessary
+    for(i = 0; i < searchTraces.length; i++) {
+        var fullInputTrace = searchTraces[i].cd[0].trace._fullInput;
+        var tracePreGUI = gd._fullLayout._tracePreGUI[fullInputTrace.uid];
+        if(tracePreGUI.selectedpoints === undefined) {
+            tracePreGUI.selectedpoints = fullInputTrace._input.selectedpoints || null;
+        }
+    }
```
alexcjohnson: Missed this one earlier as it doesn't go through `restyle` or emit a `plotly_restyle` event - but anyway it wants to be a special case so that we don't push individual array elements into `_tracePreGUI`, we link the array as a unit.

BTW I'm working off `trace._fullInput` here but I notice farther down we go straight from `trace` to `trace._input` - have we ever looked at selections on `groupby` traces?

test/jasmine/tests/plot_api_react_test.js
```diff
@@ -886,3 +887,490 @@ describe('resizing with Plotly.relayout and Plotly.react', function() {
         .then(done);
     });
 });
+
+
+describe('Plotly.react and uirevision attributes', function() {
```
alexcjohnson: I didn't write tests for every possible GUI-editable attribute yet, it's probably worthwhile doing that though. The missing ones:

- [ ] ternary zoom/pan
- [ ] polar zoom/pan & rotation
- [ ] mapbox zoom/pan
- [ ] parcoords `constraintrange`
- [ ] `editable: true`: shape positions, annotation positions & text, plot, axis, & colorbar titles, colorbar & legend positions

alexcjohnson: Heh yes, definitely worthwhile - quite a few of them exposed broken cases, either in my new code or in their existing code. AFAIK they're all covered now.

src/plots/mapbox/mapbox.js
```diff
@@ -138,22 +138,22 @@ proto.createMap = function(calcData, fullLayout, resolve, reject) {
     map.on('moveend', function(eventData) {
         if(!self.map) return;
 
-        var view = self.getView();
-
-        opts._input.center = opts.center = view.center;
-        opts._input.zoom = opts.zoom = view.zoom;
-        opts._input.bearing = opts.bearing = view.bearing;
-        opts._input.pitch = opts.pitch = view.pitch;
```
alexcjohnson: @etpinard because this was getting triggered by `map.setCenter` etc (see existing comment below) we were unnecessarily writing these settings back to `gd.layout` on initial plotting. It doesn't cause any test failures to move these into the protected block below, do you see any issues with this change?

src/plots/layout_attributes.js
```diff
@@ -234,6 +234,15 @@ module.exports = {
             'Defaults to `layout.uirevision`.'
         ].join(' ')
     },
+    selectionrevision: {
```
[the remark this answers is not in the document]
etpinard: Ha I see, `data[i].uirevision` also works for `selectedpoints` :ok_hand:

src/plots/plots.js
etpinard: Could you add a few lines about this `uid` vs trace index logic in the `plots/attributes.js` uirevision description?

src/plots/mapbox/mapbox.js
etpinard: I can't think of any potential issues here :ok_hand: 

Just make sure all the `@noCI` mapbox tests still pass.

src/components/rangeslider/draw.js
etpinard: Thinking about this again, I think I would prefer reverting to the old `'xaxis.range': [/* */]` update to preserve backward compatibility. We should add an item about in [`v2`](#420) checklist though.

src/plot_api/plot_api.js
etpinard: This here is probably not ideal performance-wise, but I kinda like having these patterns declared in block, so :ok_hand:

src/traces/parcoords/plot.js
etpinard: This probably deserves a new issue :rocket:

src/plots/layout_attributes.js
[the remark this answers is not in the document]
alexcjohnson: > Ha I see, `data[i].uirevision` also works for `selectedpoints` 👌

Does it? I don't think so, at least I don't see it when I try that.

> What's the rationale here for putting `selectionrevision` in layout instead of in the individual traces?

I can certainly think of a use case for resetting `selectedpoints` but not resetting any of the other trace attributes - any time the basic meaning of the trace is unchanged but the point indices are altered, like if you had a trace showing annual data and you change it to monthly or you change the date range. And I guess you could imagine extending this argument to say `selectedpoints` could be invalidated for one trace but not for others, though it would create a weird UX situation in which the selected points in the trace where `selectedpoints` did NOT get reset get lost amid the sea of no-longer-deselected points in the trace where `selectedpoints` DID get reset. So between that and the fact that when the user creates a selection it's not a per-trace action, that's my argument for making `selectedrevision` a `layout` attribute. The app developer would give this attribute some value that depends on every input that can affect the identity of point indices in any trace.

Now there's still a question about what `uirevision` attributes the other trace attributes should link to. For the most part it seems like `uid` takes care of resetting these attributes (and note that if `uid` changes, even attributes linked to `uirevision` attributes `layout` will reset)... I linked `colorbar.(x|y)` to `editrevision` because this is more about the overall layout of the plot than about anything in the trace itself. And I made `visible: 'legendonly'` controlled by `legend.uirevision`, because (a) that's what controls `layout.hiddenlabels` for `pie` traces, and (b) the only use case I could think of for it, aside from those covered by `trace.uid`, is a "reset legend visibility" control (so `legend.uirevision` would just be set to the `nclicks` for this control).

That leaves `name`, `colorbar.title`, and `dimensions[i].constraintrange` (and presumably `dimensionorder` and `columnorder` when we get to those) controlled by `trace.uirevision`. I can't really think of a use case for this (again, aside from those covered by `trace.uid`), but it does seem like if there ever was one the attribute would belong in the trace, and I don't want to leave any attributes *only* under the control of the main `layout.uirevision` with no inheritance, since then to reset one of those attributes without resetting the whole rest of the plot would require setting *all* the other lower-level attributes to some constant value.

src/plot_api/plot_api.js
alexcjohnson: We have a tentative plan to :hocho: undo/redo in v2 https://github.com/plotly/plotly.js/issues/420#issuecomment-322873840, with the rationale:

> we used these in our old workspace, but this is really not the right level to be managing this issue, as plots may be coupled to other application state (ie changes in data arrays) and using our queue would muck up that correspondence.

so unless we reverse that conclusion let's not build anything else using it.

src/components/rangeslider/draw.js
etpinard: Thanks! Resolved.

src/plots/layout_attributes.js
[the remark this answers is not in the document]
etpinard: > Does it? I don't think so, at least I don't see it when I try that.

Ah. I think I got that from:

https://github.com/plotly/plotly.js/blob/874a72aa86778ef223a65d8ce77b1af8b761fdd0/src/plots/attributes.js#L173-L174

> I can certainly think of a use case for resetting `selectedpoints` but not resetting any of the other trace attributes

> So between that and the fact that when the user creates a selection it's not a per-trace action, that's my argument for making `selectedrevision` a `layout` attribute.

You got me convinced. Thanks for writing your thought process down.

> but it does seem like if there ever was one the attribute would belong in the trace, and I don't want to leave any attributes only under the control of the main layout.uirevision with no inheritance

Good point. I have no problem leaving `trace.uirevision` in there even though it has (very) limited use cases.

src/plot_api/plot_api.js
etpinard: That looks ok to me. Great tests :+1:

src/plots/cartesian/select.js
alexcjohnson: Actually it looks like selections work fine with groupby 😌 there's some potential confusion around `pointNumber` vs `pointIndex` in the event data, but all the information is there (`curveNumber` and `pointIndex` give you the index in the input arrays, `pointNumber` is the index within the group), and visible behavior is all correct.

The only thing that looks odd to me is `fullTrace.selectedpoints`:
- Immediately after drawing it has the selected input array indices from items in that group, but `fullTrace._fullInput.selectedpoints` is missing
- After you make an edit (`restyle` or zoom, for example) `fullTrace.selectedpoints` and `fullTrace._fullInput.selectedpoints` *both* have the entire array of input array indices.

I don't see any practical effects of this difference, though it's possible `Plotly.react`'s diffing algo will care.

Anyway I think I've found a `uirevision` + `selectedpoints` bug, investigating.

alexcjohnson: > Anyway I think I've found a `uirevision` + `selectedpoints` bug, investigating.

haha looks like it's just the missing `fullInput.selectedpoints` I mentioned above - fix coming, and perhaps I'll make `fullTrace.selectedpoints` self-consistent across edits as well while I'm at it.

alexcjohnson: alright, 🔒 in https://github.com/plotly/plotly.js/pull/3236/commits/090231bce09a0ee064b0741ec261b9b1d10de9b3 - which should also speed things up a bit in the normal non-transform case where `data` and `fullData` can point to the same array (as they will after an edit anyway) so we only need to build it once.

src/plots/layout_attributes.js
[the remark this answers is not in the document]
alexcjohnson: > Ah. I think I got that from:

Good catch, that was wrong, thanks! -> https://github.com/plotly/plotly.js/pull/3236/commits/5a27c0043e46342d73b8d986f9ce995fc48ead60

nicolaskruchten:
I think we should hold off on merging this guy to master until we can do a prerelease and let Dash users play with it a bit... @chriddyp what do you think?

chriddyp:
> But maybe @chriddyp would like to make a dash pre-release with it?

Yeah, I'd like to have a prerelease out there for a week or so, in similar spirit to https://github.com/plotly/dash-core-components/pull/374 and https://community.plot.ly/t/exploring-a-transitions-api-for-dcc-graph/15468

etpinard:
:dancer: for me.

alexcjohnson:
@chriddyp do you want to make a parallel release candidate for this feature in dcc, or would you rather have a branch combining both features so that `0.39.0rc4` could be said to strictly build on `0.39.0rc3`? LMK, I can make that branch if you like.

chriddyp:
> a parallel release candidate for this feature in dcc

A parallel release works for me 👍 Users can try each feature out independently

chriddyp:
>  LMK, I can make that branch if you like.

That would be great!

alexcjohnson:
A branch containing both transitions and uirevision, to be used for dcc `0.39.0rc4`: https://github.com/plotly/plotly.js/compare/transitions-plus-uirevision?expand=1
https://18863-45646037-gh.circle-artifacts.com/0/plotly.min.js

chriddyp:
>  a parallel release candidate for this feature in dcc

For those reading along, see https://github.com/plotly/dash-core-components/pull/387 for the dash-core-components prerelease and https://community.plot.ly/t/preserving-ui-state-like-zoom-in-dcc-graph/15793/3 for the installation instructions and examples 🎉

nicolaskruchten:
Last chance attempt to change the name... would we consider `uidescriptor` or `uidigest` or something like that instead of `revision`? `revision` still strongly suggests monotonically-increasing numbers to me, like version numbers.

etpinard:
> `revision` still strongly suggests monotonically-increasing numbers to me, like version numbers.

`datarevision` also needs not to be monotonically increasing to work. Still voting `uirevision`.

nicolaskruchten:
> `datarevision` also needs not to be monotonically increasing to work

Yep, I have the same problems with that key name :) But perhaps that ship has indeed sailed. ⛵️

alexcjohnson:
I agree it's not a great name, but at this point I think it's the least bad. I wouldn't insist on concordance with `datarevision`, since that one at least has a very common monotonic use case (ie real time or as-available data)... but that's at least a vote in its favor, and all the other options we've floated IMO have some sort of equally or more problematic association.

src/plots/polar/layout_attributes.js
```diff
@@ -136,6 +136,9 @@ var radialAxisAttrs = {
     }
 };
 
+// radial title is not gui-editable, so it needs dflt: '', similar to carpet axes.
+radialAxisAttrs.title.text.dflt = '';
```
alexcjohnson: FYI @rmoestl - one we missed during #3276

alexcjohnson:
@etpinard a few updates to make this work compatible with camera up #3256 and title alignment #3276, and some looser tests for titles and tickson #3275 to work on my machine. Does this all look OK to merge now that we have the 👍 from @chriddyp ?

etpinard:
Yep, ok to merge :dancer: !

PR #3242, opened 2018-11-13
regl-scatter2d bug fix - to keep trace colors 

### LINKED ISSUE #3232: Scattergl traces with multiple colors set sometimes incorrectly format points in other scattergl traces
see https://codepen.io/alexcjohnson/pen/JeXOGX?editors=1010

The large point above the line should be orange (as it is in the legend) but it ends up blue, taking its color from the previous trace.

Fiddling around with the settings, there seem to be various ways to trigger or avoid this behavior. See also https://codepen.io/donolingerPHX/pen/qQZdxE?editors=1010

Seems to have come in with v1.36.0 - perhaps #2499?

archmoj:
Fixes #3232 in the `regl-scatter2d` module.
@etpinard 
@alexcjohnson

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "regl": "^1.3.7",
     "regl-error2d": "^2.0.5",
     "regl-line2d": "^3.0.12",
-    "regl-scatter2d": "git://github.com/a-vis/regl-scatter2d.git#45e7ab163fe8991bca3dfc31fba7e9cb4a62a570",
+    "regl-scatter2d": "git://github.com/a-vis/regl-scatter2d.git#12ea41d8a80ea743d8a4e5f474a69b89ac14ae2b",
     "regl-splom": "^1.0.4",
     "right-now": "^1.0.0",
     "robust-orientation": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "regl": "^1.3.7",
     "regl-error2d": "^2.0.5",
     "regl-line2d": "^3.0.12",
-    "regl-scatter2d": "^3.0.6",
+    "regl-scatter2d": "git://github.com/a-vis/regl-scatter2d.git#0515a47c5468363d19654fcb9383d7135c93c8ad",
     "regl-splom": "^1.0.4",
     "right-now": "^1.0.0",
     "robust-orientation": "^1.1.3",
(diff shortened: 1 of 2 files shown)


etpinard:
Yeah, that will fix the bug. I suspect users will notice some performance deterioration after on graphs with 256 colors, but I'm ok with showing the correct colors slowly as opposed to the wrong colors fast - until we figure out a way to use WebGL textures for graphs with more than 255 colors (tracked in https://github.com/gl-vis/regl-scatter2d/issues/14)

@archmoj would you mind adding a mock to :lock: this bug down? Grabbing the data/layout from https://codepen.io/alexcjohnson/pen/JeXOGX?editors=1010 should suffice.

package.json
```diff
@@ -103,7 +103,7 @@
     "regl": "^1.3.7",
     "regl-error2d": "^2.0.5",
     "regl-line2d": "^3.0.12",
-    "regl-scatter2d": "^3.0.6",
+    "regl-scatter2d": "git://github.com/a-vis/regl-scatter2d.git#0515a47c5468363d19654fcb9383d7135c93c8ad",
```
archmoj: https://github.com/gl-vis/regl-scatter2d/compare/plotly-issue-3232

archmoj:
@etpinard thanks for the comment. In respect to what you mentioned I revised other parts of `regl-scatter2d` which would allow using 65535 color palette.

package.json
etpinard: So this brings the max number of colors in a WebGL texture to 256x256=65536, and falls back to plain buffer beyond that? That's amazing! I gotta be honest, I don't fully understand your `split8Uint` dance, maybe you could explain it to me in person before or after the meeting tomorrow?

Now, looks like your latest commit made several image test cases fail, do you know why that is?

archmoj: Yes I would fix this today. Don't worry about the failed image test for the moment. There are a bunch of commits which are not yet pushed to gl-vis.

etpinard:
@archmoj looks like you got the tests to :white_check_mark: 

Now, could you submit a PR to `regl-scatter2d` and ping @dy to see what he thinks?

archmoj:
@etpinard I would say we are pretty close to fix this. But there is still one thing remained that will show up in the coming tests.

archmoj:
Closed in regard with #3328. But I prefer not to delete the branch too for further work...

PR #3254, opened 2018-11-16
Refactor Axes.doTicks and Axes.doTicksSingle

etpinard:
_please review this PR commit-by-commit_

`Axes.doTicksSingle` is currently used to compute and draw ticks, but also grid lines, tick labels, axis titles ... and a few more things. It is used by cartesian axes (via `Axes.doTicks`), colorbars, ternary axes and polar axes. Making that one single function work for all these cases turned `Axes.doTicksSingle` into hackcity :fearful: 

In brief, this PR:

- first breaks up `Axes.doTicksSingle` into `drawTicks`, `drawGrid`, `drawZeroLine` and `drawLabels`
- wraps the cartesian ticks, grid, label, title sequence into a method called `Axes.drawOne`
- renames `Axes.doTicks` -> `Axes.draw`, this one draws multiple (or all) cartesian axes
- breaks up a few more reusable parts from `Axes.drawOne`
- breaks up `drawTitle`, `calcBoundingBox` and `doAutoMargin`- which only get called by cartesian axes -  from `drawLabels` 
- adapts colorbar, ternary and polar axis drawing code 
- removes a few hacks in ternary and polar subplots. I'm particularly :smile: about https://github.com/plotly/plotly.js/pull/3254/commits/944482c0a481617ab3714102995fbd3a94fc9d14

----

**Note:** commit https://github.com/plotly/plotly.js/commit/5858067ed1b65742f2456fd1a5b60f3ea414f5b5 is potentially **dangenerous**. No tests are failing because of it, but ignoring this comment:

https://github.com/plotly/plotly.js/blob/3a32ac00cf5e8619932f874960f11b4daf7ceb40/src/plots/cartesian/axes.js#L1940-L1943

makes me nervous. 

----

A few things in `axes.js` could still be ironed out to make it even more "reusable":

- a lot of things depend on `ax._id.chaAt(0)`, this could made more general by adding a new `opts` flag e.g. `orientation: 'h' | 'v'`
- tick sign logic is still a bit of a mess
- clip-or-not-to-clip axis ticks logic is not great either
- `Axes.drawLabels` could be made more d3-esque, but I worrying about potential performance losses.

but hopefully you'll find this PR a step in the right direction.

cc @plotly/plotly_js 

---

TODO:

- [x] finish :pencil2:  jsDocs after initial :+1: -- **done in:** https://github.com/plotly/plotly.js/pull/3254/commits/ea6e76f33abb8eb767ddbd22f187df8cc060935c


alexcjohnson:
@etpinard this is awesome. Huge step in the right direction!

Is all of this stuff sufficiently generalized that we could call most of it `components/axes/` and leave in `cartesian/axes.js` only the stuff that's really specific to cartesian? Wouldn't have to do it now, but... is there going to be a better time?

> commit 5858067 is potentially generous. No tests are failing because of it, but ignoring this comment...

Other than reading that comment I have no recollection of what's going on there... so I guess we just test what it seems to be worried about, which is dynamic title scoot. Do we have such tests? In the `mathjax` mock I see that both tick labels and axis title support mathjax, I guess in principle we should have 4 test cases: (regular/mathjax tick labels) * (regular/mathjax title), shifting back and forth between collision and no collision and verifying that the title moves to the right place.

> a lot of things depend on `ax._id.chaAt(0)`, this could made more general by adding a new `opts` flag e.g. `orientation: 'h' | 'v'`

Or even `ax._orientation` - you're moving away from that pattern in order to make the axis objects passed to the drawing routines require less mocking, but I feel like this one is meaningful enough as part of the axis' identity - does it ever want different values for the same axis in different contexts? Anyway it's used in some other places besides the drawing code - `setConvert`, scattered about the autotick & tick value/text code, and other places.

Anyway aside from ensuring we have tests for 5858067 I wouldn't consider any of my comments ^^ blocking, just opportunities to push this even farther, if you've got the appetite for it.

src/plots/cartesian/axes.js
```diff
@@ -1805,7 +1805,7 @@ axes.drawTicks = function(gd, ax, opts) {
     ticks.enter().append('path')
         .classed(cls, 1)
         .classed('ticks', 1)
-        .classed('crisp', 1)
+        .classed('crisp', !opts.noCrisp)
```
alexcjohnson: Yeah, that's cool! But I'm not a huge fan of negated option names - could we use:
`opts.crisp !== false`?

etpinard: Done in https://github.com/plotly/plotly.js/pull/3254/commits/016b8e485ba5e586e4b4dbfbf77c988d2298f089

etpinard:
> I guess in principle we should have 4 test cases: (regular/mathjax tick labels) * (regular/mathjax title), shifting back and forth between collision and no collision and verifying that the title moves to the right place.

Looks like everything is ok: https://github.com/plotly/plotly.js/pull/3254/commits/6ccc4573639b5e16fd3157d7036fa4017fd30bb5

... by the way :arrow_heading_up: are our first MathJax jasmine tests. About time.

alexcjohnson:
Love it. 💃

etpinard:
> Is all of this stuff sufficiently generalized that we could call most of it `components/axes/` and leave in `cartesian/axes.js` only the stuff that's really specific to cartesian?

I gave this a go this morning. I couldn't easily decide where to put a few things (e.g. `cartesian/axis_ids.js` and few methods in `cartesian/axes.js`), so I'll defer that for later. 

Thinking about this, it might be nice to make a distinction between "registerable" (i.e optional) component modules (e.g. `annotations/`, `shapes/`) and component modules that are mostly intended to DRY things up (e.g. `color/`, `colorscale/` ... and the soon-to-be `axes/`). Maybe the latter could be placed in different directory e.g. `src/common/` and leave `src/components/` for "registerable" modules?

PR #3256, opened 2018-11-17
Bug fix: applying camera up.z vector at scene init

### LINKED ISSUE #2425: Initial camera up vector ignored
default camera position of plot behaves like default value of up vector have not been overridden
clicking "reset camera .*" button fixes that
moreover split second after page load plot will move from overridden up state to default up state 

more info in comment to below example

```
import plotly
import plotly.graph_objs as go

import pandas as pd

# Read data from a csv
z_data = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/api_docs/mt_bruno_elevation.csv')

data = [
    go.Surface(
        z=z_data.as_matrix()
    )
]
layout = go.Layout(
    scene=dict(
    camera=dict(center=dict(y=0, x=0, z=0),
                eye=dict(y=0.362, x=-1.213, z=1.792),
                up=dict(y=0.984, x=0.045, z=-0.168),#default position of plot behaves like default value have not been overridden
                # comment above line and initial state of plot would be the same, clicking "reset camera .*" button fixes that
                #moreover when above line is not commented plot will move from overridden up state to default up state split second after page load
                )
    )
)
fig = go.Figure(data=data, layout=layout)
plotly.offline.plot(fig, filename="filename")

```

archmoj:
Fixes #2425 by seting `dragmode: 'orbit'` at init time to allow tilted camera z vectors and then possibly switch to `turntable` mode as used to be desired default user mode.

Background: in plotly.js `dragmode: 'turntable'` was applied as the default camera behaviour which is desirable for most cases and it modifies & locks camera z-axis rotation. However in certain cases when the user had specified camera z-axis this value was ignored at the start.
The first approach applied here was to try to turn on `dragmode: 'orbit'` in those edge cases. The prototypes was working interactively but that introduced new side-effects and required complicated changes.
On the other hand the second approach applied is to simply set `orbit` as the default and then switch to `turntable` mode in all cases where the camera.z.up vector is different from being upwards i.e. [0,0,1].

@etpinard 
@alexcjohnson


src/plots/gl3d/index.js
```diff
@@ -51,6 +51,14 @@ exports.plot = function plotGl3d(gd) {
             scene = sceneLayout._scene;
 
         if(!scene) {
+            if(fullLayout[sceneId].camera.up.z !== 1 &&
+                fullLayout[sceneId].dragmode === 'turntable') {
+                fullLayout[sceneId].dragmode = 'orbit';
```
alexcjohnson: Would be better to do this in [`supplyDefaults`](https://github.com/plotly/plotly.js/blob/3a32ac00cf5e8619932f874960f11b4daf7ceb40/src/plots/gl3d/layout/defaults.js#L113)

Not sure where we ensure `||up||===1` but it doesn't look like that's done in `supplyDefaults`, so I guess at that point we might need a more complicated test (`up.x || up.y || up.z < 0` perhaps?)

test/image/mocks/gl3d_snowden_altered.json
```diff
@@ -24,8 +24,8 @@
                 "z": 2
             },
             "up": {
-                "x": 1,
-                "y": 0,
+                "x": 0,
+                "y": 1,
```
alexcjohnson: So... `gl3d_ibm_plot` has the old (verrrry old) camera format, and converts to the new format with a tilted `up`, and you allowed it to shift into `orbit` mode; that's great. And `gl3d_surface_lighting` you converted to the new camera format and turntable-compatible `up`, so the image changed because it's somewhat different from the output of the previous old->new conversion; also great.

But what's going on with `gl3d_snowden_altered`? The image didn't change, but the change you made to `up` is *not* to `{x: 0, y: 0, z: 1}` so I'd have expected the image to change, since now it will render in orbit mode.

src/plots/gl3d/index.js
archmoj: I like the idea of having more complicated test as you mentioned. On the other hand I was wondering that it is now rather difficult for the users to compute and pass z vector components. Could we add a function to normalize user inputs for this vector before reaching there too?

alexcjohnson: Only the direction of `up` matters, right? So feels to me like we shouldn't require it to be normalized at all. If the user passes in (0,0,0) we should revert to (0,0,1), otherwise seems to me we should normalize it for our own use but leave the un-normalized value in `layout` and `fullLayout`.

src/plots/gl3d/layout/layout_attributes.js
```diff
@@ -140,7 +140,7 @@ module.exports = {
         valType: 'enumerated',
         role: 'info',
         values: ['orbit', 'turntable', 'zoom', 'pan', false],
-        dflt: 'turntable',
+        dflt: 'orbit',
```
etpinard: Hmm. I would prefer removing the `dflt` key entirely - as the default value is now dynamic. 

Could you also add some info in the `description` about the new orbit-vs-turntable logic?

test/jasmine/tests/gl3d_plot_interact_test.js
```diff
@@ -394,7 +394,7 @@ describe('Test gl3d plots', function() {
 
     it('@gl should be able to reversibly change trace type', function(done) {
         var _mock = Lib.extendDeep({}, mock2);
-        var sceneLayout = { aspectratio: { x: 1, y: 1, z: 1 } };
+        var sceneLayout = { aspectratio: { x: 1, y: 1, z: 1 }, dragmode: 'turntable' };
```
etpinard: Why did you need to change this? I'm fearing a breaking change here.

etpinard: So without setting `dragmode: 'turntable'` here, it would default to `'orbit'`? 

What are the `camera` settings in this case?

src/plots/gl3d/layout/layout_attributes.js
archmoj: Once I tried to remove `dragmode.dflt`, the default behaviour for drag using right-click (orbit/turntable) is not activated.
So we may keep it there.

alexcjohnson: You need to pass the dynamic default to `dragmode` here:
https://github.com/plotly/plotly.js/blob/c9eed54f2fb16f9d2c87d5a7a0fad60c046bccc0/src/plots/gl3d/layout/defaults.js#L141

`getDfltFromLayout` is part of it but then there's the dynamic default based on `camera`. I don't think you want to be *altering* `sceneIn.dragmode` with `enableTurnTable`, you just want to be setting the dynamic default. The logic is something like:

- If `getDfltFromLayout` returns a value, use that.
- If not but the user provides a `camera` that's not `{x: 0, y: 0, z: >0}` `dfltDragmode='orbit'`
- If neither of those, `dfltDragmode = 'turntable'`
- Then you call `coerce('dragmode', dfltDragmode`) so the user can still explicitly provide whatever.
- The only case that's still weird at that point is if the user explicitly says `dragmode: 'turntable', camera: {up: {not z-up}}`. My inclination at that point is to set `sceneOut.camera.up` (and NOT `sceneIn.camera.up`) to `{x: 0, y: 0, z: 1}`. That won't change the behavior - the user might still wonder why `up` is not being honored - but at least it would raise an alarm if you call `Plotly.validate`.

test/jasmine/tests/gl3d_plot_interact_test.js
alexcjohnson: @etpinard notice this isn't a setting, it's an expectation. But I agree, we shouldn't be mutating the input `scene`, per my [comment above](https://github.com/plotly/plotly.js/pull/3256/files#r235165793)

src/plots/gl3d/layout/layout_attributes.js
archmoj: Thanks @alexcjohnson, I will modify it to be this way.

src/plots/gl3d/layout/defaults.js
```diff
@@ -110,6 +110,31 @@ function handleGl3dDefaults(sceneLayoutIn, sceneLayoutOut, coerce, opts) {
...
+            if(sceneLayoutIn.camera &&
+                sceneLayoutIn.camera.up) {
+
+                var x = sceneLayoutIn.camera.up.x;
+                var y = sceneLayoutIn.camera.up.y;
+                var z = sceneLayoutIn.camera.up.z;
+
+                if(!x || !y || !z) {
+                    dragmode = 'turntable';
+                } else if(z / Math.sqrt(x * x + y * y + z * z) > 0.999) {
```
etpinard: This PR is getting better by the minute. Thanks for your efforts @archmoj 

Now, would there be a way to tweak this condition such that `gl3d_ibm-plot` and `gl3d_surface-lighting` still default to `dragmode: 'turntable'`?

archmoj: Good point @etpinard. Let me check.

src/plot_api/helpers.js
```diff
@@ -114,7 +114,7 @@ exports.cleanLayout = function(layout) {
                 scene.camera = {
                     eye: {x: eye[0], y: eye[1], z: eye[2]},
                     center: {x: center[0], y: center[1], z: center[2]},
-                    up: {x: mat[1], y: mat[5], z: mat[9]}
+                    up: {x: 0, y: 0, z: 1} // we just ignore calculating camera z up in this case
```
archmoj: Previous calculations of `camera.up` for `gl3d_surface-lighting` & `gl3d_ibm-plot` didn't set up vector close to [0,0,1].
I thought we may ignore camera z up if `cameraposition` depreciated setting is used.

test/image/mocks/gl3d_text-weirdness.json
```diff
@@ -42,22 +42,6 @@
...
-          -0.32134727420767745
-        ],
-        [
-          0,
-          0,
-          0
-        ],
-        1.8771354322421991
-      ]
-    },
```
archmoj: I thought this might not be necessary. If one wants to enable `cameraposition` it is required to use `"scene"` instead of `"undefined"`! Also now I am curious about this mock & why it was named `weirdness`? Also if we could possibly remove it or reduce the width/height.

etpinard: Wow. Thanks for fixing this. That `undefined` shouldn't be there.

src/plot_api/helpers.js
etpinard: Nice solution.

Issue #3260, opened 2018-11-19
[Bug?] Strange clip-path problem 


Braintelligence:
Hi everyone.

So I have this strange problem, where the clip-path in the SVG is wrong due to addition of the location to the clip-path url.

I have the exact same HTML offline, where when you generate the SVG from a `file://` location you get something like:
```SVG
<defs id="defs-fac894">
        <g class="clips">
            <clipPath id="clipfac894xyplot" class="plotclip"><rect width="518" height="350"/></clipPath>
            ...
```
together with:
```SVG
<g class="cartesianlayer">
        <g class="subplot xy">
            ...
            <g class="plot" transform="translate(42, 50)" clip-path="url(#clipfac894xyplot)">
            ...
```

But when you do the same online you get:
```SVG
<defs id="defs-6f068e">
        <g class="clips">
            <clipPath id="clip6f068exyplot" class="plotclip"><rect width="518" height="350"/></clipPath>
            ...
```
together with:
```SVG
<g class="cartesianlayer">
        <g class="subplot xy">
             ...
            <g class="plot" transform="translate(42, 50)" clip-path="url(https://www.mydomain.com#clip6f068exyplot)">
             ...
```

**If you use `.downloadImage` on the original state of each plot, both images are identical.** If you do this **with a zoomed state**, where you **pass the `x.range` and `y.range`** to the `layout` for `.downloadImage` **the resulting image isn't clipped correctly _in the online version_**.

This is an image that results of the first code (used offline with file://-URL) after zooming: [svg_offline.zip](https://github.com/plotly/plotly.js/files/2594588/svg_offline.zip)

This is an image that results of the latter code (used online with https://-URL) after zooming: [svg_online.zip](https://github.com/plotly/plotly.js/files/2594590/svg_online.zip)

When you change 
```SVG
<g class="plot" transform="translate(42, 50)" clip-path="url(https://www.mydomain.com#clip6f068exyplot)">
``` 
to 
```SVG
<g class="plot" transform="translate(42, 50)" clip-path="url(#clip6f068exyplot)">
``` 
(in other words: remove the location part from the url) you get the desired image.

This was tested with the current Firefox 63 as well as Firefox 52 ESR and current Microsoft Edge.

Braintelligence:
Some more notes, that I'll edit on the fly:
1. This doesn't happen when providing the offline files statically via a python http server on localhost.

etpinard:
> where when you generate the SVG from a `file://` location

[...]

PR #3272, opened 2018-11-20
Do not add <base> href to SVG clip paths during toImage

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3260

@plotly/plotly_js to do this "properly" i.e. w/o doing something hacky in `Snapshot.toSVG` like

```js
d3.selectAll('[clipPath'].each(function() {
  var s = d3.select(this);
  var cp = s.attr('clipPath');
  s.attr('clipPath'. cp.replace(URL_REGEX, '');
})
```

I made `Drawing.setClipUrl` expect `gd` as third argument. This means we can stash the `<base>` href on `gd._context` (instead of on `Drawing`) and use it in `setClipUrl`.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -110,9 +110,6 @@ exports.plot = function(gd, data, layout, config) {
     // so we can share cached text across tabs
     Drawing.makeTester();
 
-    // clear stashed base url
-    delete Drawing.baseUrl;
-
     // collect promises for any async actions during plotting
     // any part of the plotting code can push to gd._promises, then
     // before we move to the next step, we check that they're all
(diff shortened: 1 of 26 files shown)


src/plot_api/to_image.js
```diff
@@ -138,6 +138,7 @@ function toImage(gd, opts) {
     // extend config for static plot
     var configImage = Lib.extendFlat({}, config, {
         staticPlot: true,
+        exportedPlot: true,
```
etpinard: This here means that #3260 is only fixed for `Plotly.toImage` and `Plotly.downloadImage`, the other (unofficial) methods on `Plotly.Snapshot` will still be buggy.

antoinerg:
Looks good to me :dancer:

src/plot_api/plot_api.js
```diff
@@ -419,7 +416,16 @@ function opaqueSetBackground(gd, bgColor) {
...
 function setPlotContext(gd, config) {
-    if(!gd._context) gd._context = Lib.extendDeep({}, defaultConfig);
+    if(!gd._context) {
+        gd._context = Lib.extendDeep({}, defaultConfig);
+
+        // stash <base> href, used to make robust clipPath URLs
+        var base = d3.select('base');
+        gd._context.baseUrl = base.size() && base.attr('href') ?
+            window.location.href.split('#')[0] :
+            '';
```
alexcjohnson: This isn't something we will ever expose as a setting to users, right? I'd make it `_baseUrl` then.

alexcjohnson:
I'm happy now, and your update to @Braintelligence's fiddle https://github.com/plotly/plotly.js/issues/3260#issuecomment-440847345 works for me - so 💃 from my side but perhaps we can wait for @Braintelligence to confirm.

Braintelligence:
Sorry guys, I am on vacation. I can confirm tomorrow, but if my fiddle doesn't behave like described anymore, it should be alright!

Issue #3273, opened 2018-11-20
changing `autocolorscale` via Plotly.restyle only works on first try


antoinerg:
As shown in the following Codepen: https://codepen.io/anon/pen/VVybMO?editors=1000

etpinard:
This one is a side effect from mutating `gd.data[0].colorscale`

In the codepen above, the graph starts out with `gd.data[0].colorscale: 'Viridis'`, after the first `restyle(gd, 'autocolorscale', false)`, line

https://github.com/plotly/plotly.js/blob/11bec79d010a44694bf70b357bd11e11ae874ccb/src/components/colorscale/calc.js#L92

sets `gd.data[0].colorscale: [[0,"rgb(220,220,220)"],[0.2,"rgb(245,195,157)"],[0.4,"rgb(245,160,105)"],[1,"rgb(178,10,28)"]]` and the original `Viridis` is gone.

I wonder if there's a way to make colorscale attributes behave like autobin attributes post https://github.com/plotly/plotly.js/pull/3044 ?

Issue #2925, opened 2018-08-18
Sequential/Diverging defaults


nicolaskruchten:
`colorway` is going to be a very useful parameter for the new Themes feature of Chart Studio, because by setting just that attribute and saving it in a template, one will be able to provide custom defaults for colors of new traces, plus pie charts.

I think it will be necessary to add the equivalent custom defaults for sequential and diverging colorscales (overriding the current built-in default white-to-red and blue-to-white-to-red that appear when you first try to color by data in the editor), such that a template can specify just these three to prepopulate the default colors in all three modes: categorical, sequential and diverging.

nicolaskruchten:
To clarify: this would be two new keys in `layout`, something like `layout.sequentialscale` and `layout.divergingscale`which would be arrays of color-strings, and whose defaults would be the current defaults plotly.js uses for those kinds of scales when `colorscale` is not specified.

nicolaskruchten:
Even better :)

nicolaskruchten:
This now seems fairly necessary, as Themes support in Chart Studio Cloud doesn't seem to grab the Theme-provided sequential colorscales ... or at least plotly.js isn't offering them up as defaults in `fullData` ...? http://plot.ly/create and try to make a scatter plot colored by a column with a theme.

jonmmease:
This would be great. Right now themes have to individually set defaults for every colorscale across all trace types.  It would be nice to be able to specify this in one place (or one place for sequential and one place for diverging).

nicolaskruchten:
Could we assume that diverging is the concatenation of increasing and
decreasing sequential or do we need all three?

jonmmease:
I'm not used to thinking of increasing and decreasing as separate colorscale types.  Could you elaborate?  I was picturing that the defaults would be for sequential and diverging colorscales.

nicolaskruchten:
I’m with you Jon but see comment above from Alex :)

jonmmease:
Ah, missed that.  @alexj how does plotly.js decide which scale to use?

alexcjohnson:
Right, the negative sequential is unusual, we only use it when you have uniformly negative data and you ask for `autocolorscale` (If you don't explicitly ask for `autocolorscale` we default to the diverging blue->gray->red). The idea being that the goal with a sequential colorscale is generally to vary from a "weak" color representing small values to a "strong" one representing large values, and "larger" more often means "bigger absolute value" than ">"

[...]

PR #3274, opened 2018-11-20
initial implementation of `layout.colorscale`

antoinerg:
Attempt at closing #2925. As far as I can tell it works well but test coverage is pretty minimal at the moment.

In the process of making this PR, I uncovered issue #3273 .

### DIFF
--- a/src/components/colorscale/calc.js
+++ b/src/components/colorscale/calc.js
@@ -84,9 +84,9 @@ module.exports = function calc(trace, vals, containerStr, cLetter) {
     doUpdate(autoAttr, (auto !== false || (min === undefined && max === undefined)));
 
     if(container.autocolorscale) {
-        if(min * max < 0) scl = scales.RdBu;
-        else if(min >= 0) scl = scales.Reds;
-        else scl = scales.Blues;
+        if(min * max < 0) scl = container.diverging || scales.RdBu;
+        else if(min >= 0) scl = container.sequential || scales.Reds;
+        else scl = container.sequentialminus || scales.Blues;
 
         // reversescale is handled at the containerOut level
         doUpdate('colorscale', scl, container.reversescale ? flipScale(scl) : scl);
--- a/src/components/colorscale/defaults.js
+++ b/src/components/colorscale/defaults.js
@@ -18,7 +18,6 @@ var colorbarDefaults = require('../colorbar/defaults');
 var isValidScale = require('./is_valid_scale');
 var flipScale = require('./flip_scale');
 
-
 module.exports = function colorScaleDefaults(traceIn, traceOut, layout, coerce, opts) {
     var prefix = opts.prefix,
         cLetter = opts.cLetter,
@@ -43,7 +42,13 @@ module.exports = function colorScaleDefaults(traceIn, traceOut, layout, coerce, 
     var autoColorscaleDflt;
     if(sclIn !== undefined) autoColorscaleDflt = !isValidScale(sclIn);
     coerce(prefix + 'autocolorscale', autoColorscaleDflt);
-    var sclOut = coerce(prefix + 'colorscale');
+
+    var layoutColorscale = layout.colorscale || {};
+    containerOut.diverging = layoutColorscale.diverging;
+    containerOut.sequential = layoutColorscale.sequential;
+    containerOut.sequentialminus = layoutColorscale.sequentialminus;
+    var dfltScl = containerOut.diverging;
+    var sclOut = coerce(prefix + 'colorscale', dfltScl);
 
     // reversescale is handled at the containerOut level
     var reverseScale = coerce(prefix + 'reversescale');
--- a/src/plots/layout_attributes.js
+++ b/src/plots/layout_attributes.js
@@ -188,6 +188,29 @@ module.exports = {
         editType: 'calc',
         description: 'Sets the default trace colors.'
     },
+    colorscale: {
+        sequential: {
+            valType: 'colorscale',
+            dflt: false,
+            role: 'style',
+            editType: 'calc',
+            description: 'Sets the default sequential colorscale for positive values.'
+        },
+        sequentialminus: {
+            valType: 'colorscale',
+            dflt: false,
+            role: 'style',
+            editType: 'calc',
+            description: 'Sets the default sequential colorscale for negative values.'
+        },
+        diverging: {
+            valType: 'colorscale',
+            dflt: false,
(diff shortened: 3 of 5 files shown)


src/plots/layout_attributes.js
```diff
@@ -188,6 +188,29 @@ module.exports = {
         editType: 'calc',
         description: 'Sets the default trace colors.'
     },
+    colorscale: {
```
etpinard: Let's put these new attributes and default logic in the [`Colorscale`](https://github.com/plotly/plotly.js/tree/master/src/components/colorscale) component module. 

Create two new files:

- src/components/colorscale/layout_attributes.js
- src/components/colorscale/layout_defaults.js

and link their `exports` in src/components/colorscale/index.js as `layoutAttributes` and `supplyLayoutDefaults`. The latter will get automatically picked in in `Plotly.supplyLayoutModuleDefaults` here:

https://github.com/plotly/plotly.js/blob/695f3114c86fee146c7f8e4b8700fa0b00d5c304/src/plots/plots.js#L1512-L1518

src/plots/layout_attributes.js
```diff
@@ -188,6 +188,29 @@ module.exports = {
...
+        sequential: {
+            valType: 'colorscale',
+            dflt: false,
```
etpinard: What would happen if we set `dflt: scales.RdBu` ?

src/components/colorscale/defaults.js
```diff
@@ -43,7 +42,15 @@ module.exports = function colorScaleDefaults(traceIn, traceOut, layout, coerce,
...
     coerce(prefix + 'autocolorscale', autoColorscaleDflt);
-    var sclOut = coerce(prefix + 'colorscale');
+
+    var layoutColorscale = layout.colorscale || {};
+    containerOut.diverging = layoutColorscale.diverging;
+    containerOut.sequential = layoutColorscale.sequential;
+    containerOut.sequentialminus = layoutColorscale.sequentialminus;
+    var dfltScl = containerOut.diverging;
+    var sclOut;
+    if(dfltScl) sclOut = coerce(prefix + 'colorscale', dfltScl);
```
etpinard: Ha I see. We wouldn't need this extra `if-else` block if you made the `layout.colorscale.*.dflt` `undefined` (or removed them entirely) instead of `false`. The relevant line in `coerce.js` is:

https://github.com/plotly/plotly.js/blob/695f3114c86fee146c7f8e4b8700fa0b00d5c304/src/lib/coerce.js#L384

etpinard: ... But looking at this again, do we really need logic in colorscale/defaults.js **and** in `colorscale/calc.js`?

Ideally, 

- I would leave colorscale/defaults.js untouched
- make the `layout.colorscale.*.dflt` match `scales.RdBu`, `scales.Reds` and `scales.Blues`, making those `container.diverging || scales.RdBu;` in colorscale/calc.js unnecessary.
- handle the autocolorscale logic in colorscale/calc.js

Now, I'm just thinking out-loud. @antoinerg let me know if my proposal has any flaws.

src/plots/plots.js
```diff
@@ -1362,9 +1362,11 @@ plots.supplyLayoutGlobalDefaults = function(layoutIn, layoutOut, formatObj) {
     coerce('hidesources');
 
     coerce('colorway');
-    coerce('colorscale.sequential');
-    coerce('colorscale.sequentialminus');
-    coerce('colorscale.diverging');
+    if(layoutIn.colorscale) {
```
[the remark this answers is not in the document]
antoinerg: Coercing `valType: colorScale` will always give it a default colorscale. I initially wanted to be falsey if not set by the user. There are other ways to do that.

src/plots/layout_attributes.js
antoinerg: That's another way of doing it and I think it's much better than the current approach. We'd have the defaults clearly defined in the attributes where it should be :tada:

src/components/colorscale/defaults.js
antoinerg: I think this makes a lot of sense. Let me rework this PR!

antoinerg:
Commit 93ce241 refactors `colorscale_test.js`. 

I still need to improve test coverage to make sure `layout.colorscale.(sequential|sequentialminus|diverging)` are used appropriately when `autorange: true` and no colorscale is provided by the user. It should already work but it should be :lock: down!

antoinerg:
Commit ee222d5 improves the test coverage!

src/traces/bar/calc.js
```diff
@@ -42,10 +42,18 @@ module.exports = function calc(gd, trace) {
 
     // auto-z and autocolorscale if applicable
     if(hasColorscale(trace, 'marker')) {
-        colorscaleCalc(trace, trace.marker.color, 'marker', 'c');
+        colorscaleCalc(gd, trace, {
```
etpinard: Looking nice :gem:

src/components/colorscale/calc.js
```diff
@@ -84,9 +86,9 @@ module.exports = function calc(trace, vals, containerStr, cLetter) {
     doUpdate(autoAttr, (auto !== false || (min === undefined && max === undefined)));
 
     if(container.autocolorscale) {
-        if(min * max < 0) scl = scales.RdBu;
-        else if(min >= 0) scl = scales.Reds;
-        else scl = scales.Blues;
+        if(min * max < 0) scl = gd._fullLayout.colorscale.diverging;
```
etpinard: minor :cow2: , usually when a method uses `fullLayout` from `gd` we put a

```js
```

just below the `function (`

etpinard:
Nicely done @antoinerg 

Since this PR will mostly help out template users, we should add a JSON mock that uses one of the new attribute in a template.

test/image/mocks/colorscale_template.json
```diff
@@ -0,0 +1,157 @@
...
+      4.299876
+    ],
+    "type": "heatmap",
+    "autocolorscale": true
+  }],
+  "layout": {
+    "template": {
+      "layout": {
+        "title": "heatmap template",
+        "colorscale": {
```
etpinard: What if we put `autocolorscale: true` in the `template` container, does it still work?

antoinerg: Good call! I modified the mock in dca4350

etpinard:
Nicely done :dancer:

Issue #2601, opened 2018-05-02
Categorical ticks and gridlines halfway between category labels


yangmqglobe:
When use categorical axis, the grid still drew in the middle of a category which you can find in [here](https://codepen.io/anon/pen/erWLjw). By the way, the [striped-rows](http://getbootstrap.com/docs/4.1/content/tables/#striped-rows) style would be great for this situation(Yes, it is a table, but it can be used in plots right?).

alexcjohnson:
Good point, ticks and gridlines halfway between the category labels is often better than aligned with the labels, as they are for numeric or date axes. Not always, for example label-aligned may be better for scatter/bubble plots. Also I'm not sure what the right behavior is when categories get too dense and we can't show them all...

Striped rows is interesting, I see a few examples of it elsewhere, like [here](https://canvasjs.com/docs/charts/basics-of-creating-html5-chart/tick-marks-grid-lines-and-interlaced-colors/) - it's not an effect I'd use myself, I think it makes interpreting the data harder. But we'd entertain a PR to add it 😄

yangmqglobe:
em...My idea is not striped rows but striped columns, so the two adjacent category can distinguish easily, this will highly affect the group boxplot, group violin plot, etc.

alexcjohnson:
> striped rows but striped columns, so the two adjacent category can distinguish easily

Nice, I'd call that the same feature (just applied to the x axis instead of y) but that does seem like a more useful application of it. Note that if you zoom out in your example you can see the issue I was worried about, when categories get too dense and the stripes (and/or ticks/lines) get put in what's suddenly a weird place:
![screen shot 2018-05-03 at 10 02 50 am](https://user-images.githubusercontent.com/2678795/39581280-31ecc368-4eb9-11e8-8724-770013401d8c.png)

> Another advance would be the category region hover highlight

interesting, kind of a generalization of the idea of spikelines https://github.com/plotly/plotly.js/pull/1461 - perhaps this could fit into that framework.

yangmqglobe:
In my opinion, for the dense data, user/developer should change their design like zoom their axis or change the plot type like this one
![image](https://user-images.githubusercontent.com/17685088/39611379-d05f2fc8-4f1c-11e8-8738-58a3630f0e31.png)
When there are too much bar, then you should use polar coordinates, so I think the stripe style should be an option weather for xaxis or yaxis, but the user/develop should consider is it fix for their work.

nicolaskruchten:
[...]

PR #3275, opened 2018-11-20
Add `tickson: 'boundaries'` for category cartesian axes

etpinard:
Closes https://github.com/plotly/plotly.js/issues/2601 - adding `tickson` to date axes and special spike support should get issues of their own.

cc @plotly/plotly_js


src/components/colorbar/draw.js
```diff
@@ -198,6 +198,7 @@ module.exports = function draw(gd, id) {
                 letter: 'y',
                 font: fullLayout.font,
                 noHover: true,
+                noTickson: true,
```
alexcjohnson: until we make discrete colorscales perhaps 😅

src/plots/cartesian/axes.js
```diff
@@ -1610,14 +1610,41 @@ axes.drawOne = function(gd, ax, opts) {
...
+        // draw ticks and grid lines 1/2 a 'category' to the left,
+        // add one item at axis tail
+        var valsBoundaries = vals.map(function(d) {
+            var d2 = Lib.extendFlat({}, d);
+            d2.x -= 0.5;
+            return d2;
+        });
+        // not used for labels; no need to worry about the other tickTextObj keys
+        var d2 = Lib.extendFlat({}, vals[vals.length - 1]);
+        d2.x += 0.5;
```
alexcjohnson: -0.5 for the other ticks is great, but +0.5 may need altering in case `dtick !== 1` (perhaps `+= dtick - 0.5`?). Have you experimented with this case? see eg https://github.com/plotly/plotly.js/issues/2601#issuecomment-386312022

src/plots/cartesian/layout_attributes.js
```diff
@@ -304,6 +304,19 @@ module.exports = {
             'the axis lines.'
         ].join(' ')
     },
+    tickson: {
+        valType: 'enumerated',
+        values: ['labels', 'boundaries'],
+        role: 'info',
+        editType: 'ticks',
```
alexcjohnson: dflt?

alexcjohnson:
Do we want a longer default `ticklen` when you turn on `tickson: 'boundaries'`? And we definitely need to not push the labels away from the axis (ie to be beyond the ticks).

There are probably also some unpleasant effects on collision avoidance. Imagine the case:
`'A very long title', 'short', 'Another very long title'`
They could all avoid touching each other yet overlap the boundaries which would still be unacceptable, at least if we're showing ticks, though it might still be better to rotate in this case even with just grid lines. And perhaps we want to only auto-rotate to 90 degrees, omitting the step at 30 degrees?

src/plots/cartesian/layout_attributes.js
etpinard: Good call. Done in -> https://github.com/plotly/plotly.js/pull/3275/commits/78ef22ce125429ae0339b355069ad29f6ce60dc6

src/plots/cartesian/axes.js
etpinard: Ah good eye -> https://github.com/plotly/plotly.js/pull/3275/commits/6d0a3fc926a423061b954611def71abae858e73b

etpinard:
Thanks very much @alexcjohnson for the thorough review. I overlooked many subtle details. :face_with_head_bandage: 

> And we definitely need to not push the labels away from the axis (ie to be beyond the ticks).

Oh right. Done in -> https://github.com/plotly/plotly.js/pull/3275/commits/78ff4d519086101126edb8c211c84965e0dc0f7e

> There are probably also some unpleasant effects on collision avoidance. Imagine the case:
> `'A very long title', 'short', 'Another very long title'`
> They could all avoid touching each other yet overlap the boundaries which would still be unacceptable, at least if we're showing ticks, though it might still be better to rotate in this case even with just grid lines. And perhaps we want to only auto-rotate to 90 degrees, omitting the step at 30 degrees?

I went with auto-rotating straight to 90 degrees with or w/o ticks in -> https://github.com/plotly/plotly.js/pull/3275/commits/09ac230b9698b09bccc2b729aa81f6601ba84633

> Do we want a longer default `ticklen` when you turn on `tickson: 'boundaries'`? 

I'm not opposed. How much longer are you thinking?

src/plots/cartesian/axes.js
```diff
@@ -2298,14 +2339,26 @@ axes.drawLabels = function(gd, ax, opts) {
...
                     // any overlap at all - set 30 degrees
                     autoangle = 30;
                     break;
                 }
             }
 
+            if(ax.tickson === 'boundaries') {
+                for(i = 0; i < lbbArray.length; i++) {
+                    if(
+                        (vals[i].xl !== null && (lbbArray[i].left - ax.l2p(vals[i].xbnd[0])) < 2) ||
```
etpinard: Using a 2px padding, same as for `bBoxIntersect` above.

alexcjohnson:
> > Do we want a longer default `ticklen` when you turn on `tickson: 'boundaries'`?
> 
> I'm not opposed. How much longer are you thinking?

Hmm, looking at it in more detail, I think perhaps the current default is fine. Definitely once we get to multi-level axes, ticks for the outer levels will need to by default extend beyond the inner level's labels, but we can leave it as is for now.

src/plots/cartesian/axes.js
```diff
@@ -1621,18 +1635,18 @@ axes.drawOne = function(gd, ax, opts) {
...
+        // no need to worry about the other tickTextObj keys
+        var valsBoundaries = [];
+        var _push = function(d, bndIndex) {
+            var xb = d.xbnd[bndIndex];
+            if(xb !== null) {
+                valsBoundaries.push(Lib.extendFlat({}, d, {x: xb}));
+            }
+        };
+        for(i = 0; i < vals.length; i++) _push(vals[i], 0);
+        _push(vals[i - 1], 1);
```
alexcjohnson: nice solution, keeping `xbnd` with the rest of the tick calculations.

test/image/mocks/tickson_boundaries.json
```diff
@@ -38,12 +39,31 @@
...
+    },
+
+    {
+      "mode": "markers",
+      "marker": {"symbol": "square"},
+      "name": "with overlapping tick labels",
+      "x": ["A very long title", "short", "Another very long title"],
+      "y": [1, 4, 2],
+      "xaxis": "x4",
+      "yaxis": "y4"
```
alexcjohnson: Nice, just squeaks in unrotated if we draw this without `tickson: 'boundaries'` 🎉 
screen shot 2018-11-26 at 1 20 37 pm

alexcjohnson:
OK, looks great! 💃

Issue #882, opened 2016-08-25
Chart Title Alignment


CalvinFernandez:
Would it be possible to provide alignment API for chart title, "left", "center" "right" ?

etpinard:
Yeah. We should really add this.

The _reason_ why nobody has done this yet is that the plot title at the moment is set via two root `layout` level attributes (`layout.title` and `layout.titlefont`) and adding more `layout.title****` attribute would feel like polluting the root layout namespace.

Ideally, the main title should have its own layout object container, for example:

```js
 layout.title = { 
  text: 'My title', 
  font: {
    size: 20,
    color: 'red'
  },
  x: 0,
  y: 1.2,
  xanchor: 'left',
  yanchor: 'bottom',
};
```

similar to `layout.legend`. So, to add `layout.title: {}` in a backward-compatible way would require some conversion step `layout.title` -> `layout.title.text` and `layout.titlefont` -> `layout.title.font`.

Moreover, axes titles (e.g. `layout.xaxis.title`) currently follow the same `**title` / `**titlefont` pattern. To preserve consistency between title attributes,  we should also allow axes title to be input as `layout.xaxis.title.text` / `layout.xaxis.title.font`.

mviertel:
Any progress on this?

AlexVvx:
Hey, I created an pr on that issue, could you please take a look?
https://github.com/AlexVvx/plotly.js/pull/3

alexcjohnson:
Thanks for taking this on @AlexVvx! I'm torn about the right way to specify alignment. I'm going to continue the discussion here for a bit until we get the right attributes. The closest precedent we have is to use plot fraction, like legend & annotations. But it's hard to match the current behavior:
- Vertically, we just split the top margin (after accounting for anything that may have auto-expanded that margin, such as a legend)
- Horizontally, we center on the whole container, so if for example the right margin is much bigger due to a colorbar, the title will *not* be centered on the plot area.

The most common alternative title placement is in the top left corner (also whole-container referenced), but occasionally you want the left edges of the title and plot area aligned.

Also keep in mind dynamic sizing, so pixel positioning isn't so great, unless perhaps we have a way to anchor to various plot features (left edge, left plot area edge, center, plot area center, right edge, right plot area edge...)

So the most general is perhaps:
- `xref` and `'yref'`: `'container'` (including margins) `'paper'` (plot area only, not so great a name in this context but matches our usage elsewhere)
[...]

PR #3276, opened 2018-11-21
882 - Chart Title Alignment

rmoestl:
This PR adds the capability to **align the main plot title** as requested in issue #882.

**New title attributes structure**
Besides the introduction of the discussed alignment options, this PR also transitions to a new `title` attribute structure. `layout.title` has become `layout.title.text` and `layout.titlefont` has become `layout.title.font`. The same goes for any other title definitions within `layout`, e.g.  `xaxis`, `yaxis`, `polar.radialaxis` and so on.

The structure of title attributes within `data` remains unchanged as it won't have added any significant value to users compared to the additional effort this would have taken (e.g. transitioning `pie`).

**To be discussed**
- Two new `TODO`s were added that probably need a closer look by a core maintainer.
- The new title structure has changed event data of `plotly_relayout` which especially showed in two broken and thus updated tests in `legend_test.js`. Users are still able to use the old attributes structure as an input to `Plotly.relayout` and `Plotly.update`. But any handlers listening on `plotly_relayout` events, that assume the old title structure, need to be updated.
- `Lib.warn` calls in `cleanLayout` have not been added so far since the original issue is about two years old as of writing and every other clean-up code in `cleanLayout` does not yet emit warnings.

### DIFF
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -91,6 +91,19 @@ exports.cleanLayout = function(layout) {
                 }
                 delete ax.autotick;
             }
+
+            // title -> title.text
+            if(!Lib.isPlainObject(ax.title)) {
+                ax.title = {
+                    text: ax.title
+                };
+            }
+
+            // titlefont -> title.font
+            if(Lib.isPlainObject(ax.titlefont) && !Lib.isPlainObject(ax.title.font)) {
+                ax.title.font = ax.titlefont;
+                delete ax.titlefont;
+            }
         }
 
         // modifications for 3D scenes
@@ -176,20 +189,10 @@ exports.cleanLayout = function(layout) {
         }
     }
 
-    // Check for old-style title definitions
+    // Check for old-style title definition
     if(!Lib.isPlainObject(layout.title)) {
         layout.title = {
             text: layout.title
-        };
-    }
-    if(layout.xaxis && !Lib.isPlainObject(layout.xaxis.title)) {
-        layout.xaxis.title = {
-            text: layout.xaxis.title
-        };
-    }
-    if(layout.yaxis && !Lib.isPlainObject(layout.yaxis.title)) {
-        layout.yaxis.title = {
-            text: layout.yaxis.title
         };
     }
 
@@ -197,16 +200,6 @@ exports.cleanLayout = function(layout) {
     if(Lib.isPlainObject(layout.titlefont) && !Lib.isPlainObject(layout.title.font)) {
         layout.title.font = layout.titlefont;
         delete layout.titlefont;
-    }
-    if(layout.xaxis &&
-      Lib.isPlainObject(layout.xaxis.titlefont) && !Lib.isPlainObject(layout.xaxis.title.font)) {
-        layout.xaxis.title.font = layout.xaxis.titlefont;
-        delete layout.xaxis.titlefont;
-    }
-    if(layout.yaxis &&
-      Lib.isPlainObject(layout.yaxis.titlefont) && !Lib.isPlainObject(layout.yaxis.title.font)) {
-        layout.yaxis.title.font = layout.yaxis.titlefont;
-        delete layout.yaxis.titlefont;
     }
 
     /*
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -137,6 +137,11 @@ exports.cleanLayout = function(layout) {
 
                 delete scene.cameraposition;
             }
+
+            // clean axis titles
+            cleanTitle(scene.xaxis);
+            cleanTitle(scene.yaxis);
+            cleanTitle(scene.zaxis);
         }
     }
 
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -176,6 +176,23 @@ exports.cleanLayout = function(layout) {
         }
     }
 
+    // Check for old-style title definitions
+    if(!Lib.isPlainObject(layout.title)) {
+        layout.title = {
+            text: layout.title
+        };
+    }
+    if(layout.xaxis && !Lib.isPlainObject(layout.xaxis.title)) {
+        layout.xaxis.title = {
+            text: layout.xaxis.title
+        };
+    }
+    if(layout.yaxis && !Lib.isPlainObject(layout.yaxis.title)) {
+        layout.yaxis.title = {
+            text: layout.yaxis.title
+        };
+    }
+
     /*
      * Moved from rotate -> orbit for dragmode
      */
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -193,6 +193,22 @@ exports.cleanLayout = function(layout) {
         };
     }
 
+    // Check for old-style title font definitions
+    if(Lib.isPlainObject(layout.titlefont) && !Lib.isPlainObject(layout.title.font)) {
+        layout.title.font = layout.titlefont;
+        delete layout.titlefont;
+    }
+    if(layout.xaxis &&
+      Lib.isPlainObject(layout.xaxis.titlefont) && !Lib.isPlainObject(layout.xaxis.title.font)) {
+        layout.xaxis.title.font = layout.xaxis.titlefont;
+        delete layout.xaxis.titlefont;
+    }
+    if(layout.yaxis &&
+      Lib.isPlainObject(layout.yaxis.titlefont) && !Lib.isPlainObject(layout.yaxis.title.font)) {
+        layout.yaxis.title.font = layout.yaxis.titlefont;
+        delete layout.yaxis.titlefont;
+    }
+
     /*
      * Moved from rotate -> orbit for dragmode
      */
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -100,12 +100,14 @@ exports.cleanLayout = function(layout) {
         // modifications for polar
         else if(polarAttrRegex && polarAttrRegex.test(key)) {
             var polar = layout[key];
+
             cleanTitle(polar.radialaxis);
         }
 
         // modifications for ternary
         else if(ternaryAttrRegex && ternaryAttrRegex.test(key)) {
             var ternary = layout[key];
+
             cleanTitle(ternary.aaxis);
             cleanTitle(ternary.baxis);
             cleanTitle(ternary.caxis);
@@ -199,7 +201,7 @@ exports.cleanLayout = function(layout) {
         }
     }
 
-    // Check for old-style title definition
+    // clean plot title
     cleanTitle(layout);
 
     /*
@@ -211,7 +213,7 @@ exports.cleanLayout = function(layout) {
     // supported, but new tinycolor does not because they're not valid css
     Color.clean(layout);
 
-    // also clean the layout container in layout.template
+    // clean the layout container in layout.template
     if(layout.template && layout.template.layout) {
         exports.cleanLayout(layout.template.layout);
     }
@@ -229,6 +231,8 @@ function cleanAxRef(container, attr) {
 
 function cleanTitle(titleContainer) {
     if(titleContainer) {
+
+        // title -> title.text
         if(typeof titleContainer.title === 'string') {
             titleContainer.title = {
                 text: titleContainer.title
@@ -236,9 +240,13 @@ function cleanTitle(titleContainer) {
         }
 
         // titlefont -> title.font
-        if(titleContainer.title &&
-          Lib.isPlainObject(titleContainer.titlefont) &&
-          !Lib.isPlainObject(titleContainer.title.font)) {
+        var oldFontAttrSet = Lib.isPlainObject(titleContainer.titlefont);
+        var newFontAttrSet = titleContainer.title && Lib.isPlainObject(titleContainer.title.font);
+        if(oldFontAttrSet && !newFontAttrSet) {
+            if(!titleContainer.title) {
+                titleContainer.title = {};
+            }
+
             titleContainer.title.font = titleContainer.titlefont;
             delete titleContainer.titlefont;
         }
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -53,6 +53,7 @@ exports.cleanLayout = function(layout) {
     }
 
     var axisAttrRegex = (Plots.subplotsRegistry.cartesian || {}).attrRegex;
+    var polarAttrRegex = (Plots.subplotsRegistry.polar || {}).attrRegex;
     var sceneAttrRegex = (Plots.subplotsRegistry.gl3d || {}).attrRegex;
 
     var keys = Object.keys(layout);
@@ -92,6 +93,9 @@ exports.cleanLayout = function(layout) {
                 delete ax.autotick;
             }
 
+            // TODO Should this be done with typeof as well (see radialaxes)?
+            // At least code would be more obvious!
+
             // title -> title.text
             if(!Lib.isPlainObject(ax.title)) {
                 ax.title = {
@@ -103,6 +107,28 @@ exports.cleanLayout = function(layout) {
             if(Lib.isPlainObject(ax.titlefont) && !Lib.isPlainObject(ax.title.font)) {
                 ax.title.font = ax.titlefont;
                 delete ax.titlefont;
+            }
+        }
+
+        // modifications for radial axes
+        else if(polarAttrRegex && polarAttrRegex.test(key)) {
+            var polar = layout[key];
+
+            if(polar.radialaxis) {
+
+                // title -> title.text
+                if(typeof polar.radialaxis.title === 'string') {
+                    polar.radialaxis.title = {
+                        text: polar.radialaxis.title
+                    };
+                }
+
+                // titlefont -> title.font
+                if(Lib.isPlainObject(polar.radialaxis.titlefont) &&
+                  !Lib.isPlainObject(polar.radialaxis.title.font)) {
+                    polar.radialaxis.title.font = polar.radialaxis.titlefont;
+                    delete polar.radialaxis.titlefont;
+                }
             }
         }
 
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -211,6 +211,11 @@ exports.cleanLayout = function(layout) {
     // supported, but new tinycolor does not because they're not valid css
     Color.clean(layout);
 
+    // also clean the layout container in layout.template
+    if(layout.template && layout.template.layout) {
+        exports.cleanLayout(layout.template.layout);
+    }
+
     return layout;
 };
 
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -34,7 +34,7 @@ exports.clearPromiseQueue = function(gd) {
 // before it gets used for anything
 // backward compatibility and cleanup of nonstandard options
 exports.cleanLayout = function(layout) {
-    var i, j;
+    var i, j, k;
 
     if(!layout) layout = {};
 
@@ -54,6 +54,7 @@ exports.cleanLayout = function(layout) {
 
     var axisAttrRegex = (Plots.subplotsRegistry.cartesian || {}).attrRegex;
     var polarAttrRegex = (Plots.subplotsRegistry.polar || {}).attrRegex;
+    var ternaryAttrRegex = (Plots.subplotsRegistry.ternary || {}).attrRegex;
     var sceneAttrRegex = (Plots.subplotsRegistry.gl3d || {}).attrRegex;
 
     var keys = Object.keys(layout);
@@ -128,6 +129,38 @@ exports.cleanLayout = function(layout) {
                   !Lib.isPlainObject(polar.radialaxis.title.font)) {
                     polar.radialaxis.title.font = polar.radialaxis.titlefont;
                     delete polar.radialaxis.titlefont;
+                }
+            }
+        }
+
+        // modifications for ternary
+        else if(ternaryAttrRegex && ternaryAttrRegex.test(key)) {
+            var ternary = layout[key];
+
+            var ternaryKeys = Object.keys(ternary);
+            for(k = 0; k < ternaryKeys.length; k++) {
+                var ternaryKey = ternaryKeys[k];
+                if(
+                  ternaryKey === 'aaxis' ||
+                  ternaryKey === 'baxis' ||
+                  ternaryKey === 'caxis'
+                ) {
+                    var ternaryAxis = ternary[ternaryKey];
+
+                    // TODO 882 DRY that up with polar, axes and main title
+                    // title -> title.text
+                    if(typeof ternaryAxis.title === 'string') {
+                        ternaryAxis.title = {
+                            text: ternaryAxis.title
+                        };
+                    }
+
+                    // titlefont -> title.font
+                    if(Lib.isPlainObject(ternaryAxis.titlefont) &&
+                      !Lib.isPlainObject(ternaryAxis.title.font)) {
+                        ternaryAxis.title.font = ternaryAxis.titlefont;
+                        delete ternaryAxis.titlefont;
+                    }
                 }
             }
         }
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -34,7 +34,7 @@ exports.clearPromiseQueue = function(gd) {
 // before it gets used for anything
 // backward compatibility and cleanup of nonstandard options
 exports.cleanLayout = function(layout) {
-    var i, j, k;
+    var i, j;
 
     if(!layout) layout = {};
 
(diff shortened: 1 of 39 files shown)


src/plot_api/helpers.js
```diff
@@ -185,6 +213,11 @@ exports.cleanLayout = function(layout) {
     // supported, but new tinycolor does not because they're not valid css
     Color.clean(layout);
 
+    // clean the layout container in layout.template
+    if(layout.template && layout.template.layout) {
+        exports.cleanLayout(layout.template.layout);
```
etpinard: Good call here. I guess this means templates with `layout.xaxis1` didn't work previously, correct?

src/plot_api/helpers.js
```diff
@@ -196,6 +229,30 @@ function cleanAxRef(container, attr) {
     }
 }
 
+function cleanTitle(titleContainer) {
+    if(titleContainer) {
+
+        // title -> title.text
+        if(typeof titleContainer.title === 'string') {
```
etpinard: We'll need to be less stringent here. Even though `layout.title` used to be `valType: 'string'`, we also access numbers:

https://github.com/plotly/plotly.js/blob/11bec79d010a44694bf70b357bd11e11ae874ccb/src/lib/coerce.js#L120-L128

For example,

```js
Plotly.newPlot(gd, [], {title: 100, xaxis: {title: 0}, yaxis: {title: 1}})
```

gives:

image

-------


So

```
if('title' in titleContainer) { // ...
```

should be ok.

src/plot_api/plot_api.js
```diff
@@ -494,7 +494,7 @@ function setPlotContext(gd, config) {
     context._hasZeroWidth = context._hasZeroWidth || gd.clientWidth === 0;
 }
 
-function plotPolar(gd, data, layout) {
+function plotLegacyPolar(gd, data, layout) {
```
etpinard: Good call renaming this function :trophy: 

Looks like legacy polar's `layout.radialaxis` did not support a `title`, so :ok_hand:

src/plot_api/plot_api.js
```diff
@@ -1686,6 +1686,74 @@ function _restyle(gd, aobj, traces) {
...
+    keys = Object.keys(layoutObj);
+    for(i = 0; i < keys.length; i++) {
+        key = keys[i];
+        value = layoutObj[key];
+
+        if(key.indexOf('titlefont') > -1) {
+            replace(key, key.replace('titlefont', 'title.font'));
+        }
+
+        if(typeof value === 'string' && oldAxisTitleRegExp.test(key)) {
```
etpinard: Here too, we have to accept more than just `'string'` types.

src/plot_api/plot_api.js
```diff
@@ -1686,6 +1686,74 @@ function _restyle(gd, aobj, traces) {
...
+        // know the potentially existing `title.text` property.
+        layoutObj['title.font'] = layoutObj.titlefont;
+        delete layoutObj.titlefont;
+    }
+
+    // Note, that updating x-axis and y-axis title fonts
+    // was never supported unless (i) using string
+    // attributes or (ii) passing `xaxis.title` / `yaxis.title`
+    // again. And these cases are covered by
+    // helpers.cleanLayout anyways.
```
etpinard: Hmm. Could you share an example of `(ii)`?

src/plot_api/plot_api.js
```diff
@@ -1686,6 +1686,74 @@ function _restyle(gd, aobj, traces) {
...
+/**
+ * Converts deprecated layout "string attributes" to
+ * "string attributes" of the current API to ensure backwards compatibility.
+ *
+ * E.g. Maps {'xaxis.title': 'A chart'} to {'xaxis.title.text': 'A chart'}
+ *
+ * @param layoutObj
+ */
+function cleanDeprecatedLayoutAttributeStrings(layoutObj) {
+    var oldAxisTitleRegExp = /axis\d{0,2}.title$/;
```
etpinard: We should reuse 

https://github.com/plotly/plotly.js/blob/11bec79d010a44694bf70b357bd11e11ae874ccb/src/plots/cartesian/constants.js#L20

:recycle:

src/plot_api/plot_api.js
```diff
@@ -1726,6 +1794,9 @@ exports.relayout = function relayout(gd, astr, val) {
 
     if(Object.keys(aobj).length) gd.changed = true;
 
+    cleanDeprecatedLayoutAttributes(aobj);
+    cleanDeprecatedLayoutAttributeStrings(aobj);
```
etpinard: Could we call these two things from `_relayout` to avoid duplication?

src/plots/gl3d/layout/axis_defaults.js
```diff
@@ -56,7 +56,15 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, options) {
             options.fullLayout);
 
         coerce('gridcolor', colorMix(containerOut.color, options.bgColor, gridLightness).toRgbString());
-        coerce('title', axName[0]);  // shouldn't this be on-par with 2D?
+        coerce('title.text', axName[0]);  // shouldn't this be on-par with 2D?
+
+        // TODO 882 coercing old 'titlefont' was missing. Why? Activating the below would break a lot image tests...
```
etpinard: No need to worry here. `title.font` (and so did `titlefont`) is coerced in `handleAxisDefaults` a couple lines above this line.

In fact, `title.text` (and so did `title`) is also coerced in `handleAxisDefaults`. This line here coerces it again with the "correct" gl3d default value.

src/plots/pad_attributes.js
```diff
@@ -8,37 +8,46 @@
...
+/**
+ * Creates a set of padding attributes.
+ *
+ * @param {object} opts
+ *   @param {string} editType:
+ *     the editType for all pieces of this padding definition
+ *
+ * @return {object} attributes object containing {t, r, b, l} as specified
+ */
+module.exports = function(opts) {
```
etpinard: Nicely done here :ok_hand:

src/plots/polar/polar.js
```diff
@@ -483,9 +483,12 @@ proto.updateRadialAxisTitle = function(fullLayout, polarLayout, _angle) {
     var sina = Math.sin(angleRad);
 
     var pad = 0;
+
+    // Hint: no need to check if there is in fact a title.text set
+    // because if plot is editable, pad needs to be calculated anyways.
```
etpinard: Hmm. Could we elaborate here? Why does `pad` need to be calculated?

src/plot_api/subroutines.js
```diff
@@ -450,18 +454,125 @@ function findCounterAxisLineWidth(ax, side, counterAx, axList) {
...
+            y: getMainTitleY(fullLayout, dy),
+            'text-anchor': textAnchor,
+            dy: dy
         }
     });
 };
 
+function getMainTitleX(fullLayout, textAnchor) {
+    var title = fullLayout.title;
+    var _size = fullLayout._size;
```
etpinard: minor :cow2: we usually use `var gs = fullLayout._size` (for **g**raph **s**ize).

src/plot_api/subroutines.js
```diff
@@ -450,18 +454,125 @@ function findCounterAxisLineWidth(ax, side, counterAx, axList) {
...
+            case 'paper':
+                return _size.t + _size.h - _size.h * title.y + vPadShift;
+            case 'container':
+            default:
+                return fullLayout.height - fullLayout.height * title.y + vPadShift;
+        }
+    }
+}
+
+function getMainTitleTextAnchor(fullLayout) {
```
etpinard: [Non-blocking as this is a trivial amount of code] It might be nice to move  

https://github.com/plotly/plotly.js/blob/master/src/components/legend/anchor_utils.js

to `src/lib/` and reuse it here.

src/plot_api/plot_api.js
etpinard: Ha I see, this is use for all (cartesian, ternary, polar, scene) axes, so we can't just reuse the cartesian `attrRegex`. We should create a new regex: `counterRegex('axis')` instead.

test/jasmine/tests/titles_test.js
```diff
@@ -1,13 +1,1060 @@
...
+        expectCapLineAlignsWithTopEdgeOf(paperElemSelector);
+    });
+
+    function extendLayout(titleAttrs) {
+        return {
+            title: Lib.extendFlat({text: 'A Chart Title'}, titleAttrs),
+
+            // This needs to be set to have a DOM element that represents the
+            // exact size of the plot area.
+            plot_bgcolor: '#f9f9f9',
```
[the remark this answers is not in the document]
etpinard: .... no need to change this of course. With the comment above, your solution is crystal clear :crystal_ball:

etpinard:
Another very impressive and very well-tested PR from @rmoestl (we'll miss you :cry: )

It's a bit of a shame that only ~100 new lines in `subroutines.js` we're needed to get this to work, but almost 10x more lines had to be patched to preserve backward compatibility and `title` attribute consistency. Thanks for your hard work :muscle: 

----

Now,

> The structure of title attributes within data remains unchanged as it won't have added any significant value to users compared to the additional effort this would have taken (e.g. transitioning pie).

This is a tricky topic. On one hand, now would be the perfect time to transition all other `title` attributes to the new signature (that's `colorbar.title` and pie `title`, anything else?). But on the other, maybe there's no need to have `title` as an attribute container in traces. Colorbar has `title`, `titleside` and `titlefont`. Pie traces have `title`, `titleposition` and `titlefont`. Will we ever need more attributes? Maybe that's something we can discuss in tomorrow's meeting.

> Two new TODOs were added that probably need a closer look by a core maintainer.

Answers (unless I spotted the wrong `TODO`s) -> https://github.com/plotly/plotly.js/pull/3276#discussion_r235521986 and https://github.com/plotly/plotly.js/pull/3276#discussion_r235524798

> The new title structure has changed event data of plotly_relayout which especially showed in two broken and thus updated tests in legend_test.js. Users are still able to use the old attributes structure as an input to Plotly.relayout and Plotly.update. But any handlers listening on plotly_relayout events, that assume the old title structure, need to be updated.

Thanks for bringing this up. This could potentially break some apps. The most problematic use-case I can think of is: an app with an `{editable: true}` graph that listens to `plotly_relayout` to update something else in the app when a user edits the graph title. The event handler would be something like:

```js
gd.on('plotly_relayout', d => {
  // on master
  if(d.title) { /* update with new d.title value */ } 

  // on this branch
  if(d['title.text']) { /* ... */ }
})
```

So what if we made the relayout event data have two keys `'title'` and `'title.text'` that way old and new signatures would be supported?

> Lib.warn calls in cleanLayout have not been added so far since the original issue is about two years old as of writing and every other clean-up code in cleanLayout does not yet emit warnings.

Yeah, making `cleanLayout` (and `cleanData`) call `Lib.warn`  can wait another PR. **But** in the meantime, we should add info in the `title.text` and `title.font` descriptions (in `plots/layout_attributes.js`)  about how we changed the API **and** add a `_deprecated` container (see [example](https://github.com/plotly/plotly.js/blob/11bec79d010a44694bf70b357bd11e11ae874ccb/src/traces/bar/attributes.js#L199-L207)) so that with the help of https://github.com/plotly/documentation/issues/1104, the old attribute still show up in our docs.

src/plot_api/helpers.js
rmoestl: Ah, I wasn't aware of that. But it's certainly good behavior!

`if('title' in titleContainer) { // ...` would be `true` when `title` is an `object` as well.
So I think 
`if(typeof titleContainer.title === 'string' || typeof titleContainer.title === 'number') { // ...`
will do the job.

src/plot_api/plot_api.js
rmoestl: I think the gist should be that if `xaxis` or `yaxis` is passed to relayout, a `replot` will get triggered. And `replot` calls and thus cleans deprecated attributes anyways. 
However, I've decided to throw away the code of `cleanDeprecatedLayoutAttributes` because `cleanDeprecatedLayoutAttributeStrings` effectively does the same thing to `titlefont`. Though I renamed `cleanDeprecatedLayoutAttributeStrings` now. Hope the code gets a little bit clearer this way.

src/plot_api/plot_api.js
rmoestl: Yep!

src/plot_api/plot_api.js
rmoestl: Absolutely! Done.

src/plot_api/plot_api.js
rmoestl: I have considered using it. Problem right now is that all regular expressions start with `^` at the moment. 

Should I add another optional parameter to `counter` that allows to opt out of that?

src/plot_api/subroutines.js
rmoestl: Oh, thanks.

src/plots/gl3d/layout/axis_defaults.js
rmoestl: Okay, thanks. Removed it.

src/plot_api/helpers.js
etpinard: Well done. Thanks!

test/jasmine/bundle_tests/plotschema_test.js
```diff
@@ -228,7 +228,7 @@ describe('plot schema', function() {
 
         assertPlotSchema(
             function(attr, attrName, attrs, level, attrString) {
-                if(attr && isPlainObject(attr[DEPRECATED])) {
+                if(attr && isPlainObject(attr[DEPRECATED]) && isValObject(attr[DEPRECATED])) {
```
etpinard: Thanks :+1:

src/plot_api/plot_api.js
etpinard: Nicely done :ok_hand:

src/plot_api/plot_api.js
etpinard: Ha good point. 

> Should I add another optional parameter to counter that allows to opt out of that?

That, or just replacing `\d{0,2}` with `([2-9]|[1-9][0-9]+)?` to match the same axis name patterns. Your call.

etpinard:
> Long story short, I suggest to retain this behavior – passing back what was originally provided.

+1 from me. I like this solution.


> Have you discussed the question in the meeting?

Unfortunately, no. @alexcjohnson was also not present. So, let's ask that question again:

Should we also change `title` / `titlefont` and other `title*` trace attribute to `title.text` / `title.font` / ... in this PR?

src/plots/plots.js
```diff
@@ -1315,6 +1315,19 @@ plots.supplyLayoutGlobalDefaults = function(layoutIn, layoutOut, formatObj) {
         return Lib.coerce(layoutIn, layoutOut, plots.layoutAttributes, attr, dflt);
     }
 
+    // Special treatment for title.y that can be 'auto'
+    // or a number between 0 and 1.
+    function coerceTitleY() {
```
etpinard: cc @alexcjohnson here's another attribute that would benefit from adding a special string value to `valType: 'number'` as discussed in https://github.com/plotly/plotly.js/pull/3109#discussion_r227543929 and https://github.com/plotly/plotly.js/issues/2834#issuecomment-433529733

etpinard: [no need to tackle this in this PR of course @rmoestl]

src/plots/polar/polar.js
rmoestl: Thanks. Done.

src/plots/polar/layout_attributes.js
```diff
@@ -117,7 +117,12 @@ var radialAxisAttrs = {
 
     hoverformat: axesAttrs.hoverformat,
 
-    editType: 'calc'
+    editType: 'calc',
+
+    _deprecated: {
+        title: overrideAll(axesAttrs._deprecated.title, 'plot', 'from-root'),
```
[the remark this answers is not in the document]
rmoestl: Okay, thanks.

src/plot_api/plot_api.js
rmoestl: Okay. I've added another parameter to `Lib.counterRegex` and reused it here.

src/plots/plots.js
alexcjohnson: That would be great, but would this not work with `type: 'number', dflt: 'auto'`? Anyway as is I don't see where this version brings in the bounds `[0, 1]`?

src/plots/plots.js
```diff
@@ -1315,6 +1315,19 @@ plots.supplyLayoutGlobalDefaults = function(layoutIn, layoutOut, formatObj) {
...
+        if(layoutIn.title && isNumeric(layoutIn.title.y)) {
+            var propOut = Lib.nestedProperty(layoutOut, 'title.y');
+            var opts = Lib.nestedProperty(plots.layoutAttributes, 'title.y').get();
```
alexcjohnson: If we keep this, it doesn't need to be `nestedProperty` - `plots.layoutAttributes.title.y` would work, right?

src/plots/plots.js
rmoestl: Oh, my bad. I removed min and max from the attribute definition after I think plotschema_test was having failures. It comlained that min and max were defined for `valType: 'any'`. I should have stayed with `valType: 'number'` in the first place. I should never have removed min and max without at least checking the bounds in the custom `coerceTitleY`.

src/plots/plots.js
rmoestl: You're right. `Lib.valObjectMeta.number.coerceFunction` used to check the defined min/max bounds but the removal of min/max in the attribute definition rendered that obsolete (see above) :man_facepalming:

src/plot_api/subroutines.js
etpinard: Thanks very much for https://github.com/plotly/plotly.js/pull/3276/commits/86a88e24db36aedcbc0e5db8a7e6f086a1185c6c :muscle:

src/plots/plots.js
rmoestl: > `type: 'number', dflt: 'auto'`

I quickly tested that and it looks good I believe. I'll add another commit.

alexcjohnson:
> Should we also change `title` / `titlefont` and other `title*` trace attribute to `title.text` / `title.font` / ... in this PR?

`carpet.(a|b)axis.title` seems to me like it maps naturally to cartesian axis titles, so I feel like it would be confusing NOT to move this to the new structure. And at that point seems to me we should convert them all - ie any remaining `title` -> `title.text` and `title*` -> `title.*`

src/plots/plots.js
rmoestl: See https://github.com/plotly/plotly.js/pull/3276/commits/5fb7b8f4aa180ee894dce63646510407fb535777

src/plots/plots.js
rmoestl: See commit https://github.com/plotly/plotly.js/pull/3276/commits/5fb7b8f4aa180ee894dce63646510407fb535777

rmoestl:
> And at that point seems to me we should convert them all - ie any remaining `title` -> `title.text` and `title*` -> `title.*`

Converted `pie`, `carpet` and `colorbar` now. `colorbar` component is used in a lot of trace types of course.

src/traces/carpet/plot.js
```diff
@@ -213,16 +213,16 @@ function drawAxisTitle(gd, layer, trace, t, xy, dxy, axis, xa, ya, labelOrientat
...
+        offset += titleSize + axis.title.offset;
 
         var labelNorm = labelOrientation.angle + (labelOrientation.flip < 0 ? 180 : 0);
         var angleDiff = (labelNorm - orientation.angle + 450) % 360;
         var reverseTitle = angleDiff > 90 && angleDiff < 270;
 
         var el = d3.select(this);
 
-        el.text(axis.title || '')
+        el.text(axis.title.text)
```
rmoestl: FYI: removed `|| ''` because an empty title wouldn't be part of the joined `data` anyways.

src/traces/carpet/axis_defaults.js
```diff
@@ -203,11 +203,6 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, options)
     // but no, we *actually* want to coerce this.
     coerce('tickmode');
 
-    if(!containerOut.title || (containerOut.title && containerOut.title.length === 0)) {
```
rmoestl: I've a question about this one: when migrating this piece of code to the new nested structure, `trace.aaxis.title` becomes `undefined` when being used in `carpet/plot`. I tracked down `relinkPrivateKeys` (called by the `supplyDefaults` routine) to be the cause. 

I see two options:
- Code in `carpet` that wants to access `[ab]axis.title.text` has to always check if `[ab]axis.title` is defined. But doesn't that undermine the *coercion* mechanism?
- Do not erase `title.font` and `title.offset` when title is empty. By the way, what is the reason here to delete already coerced attributes?

alexcjohnson: > what is the reason here to delete already coerced attributes?

Usually we do it the other way around - we only coerce the attributes we need, so in this case it would look like
```js
var title = coerce('title.text');
if(title) { // what's up with the (title && title.length === 0) clause? useless, right?
    coerceFont('title.font');
    coerce('title.offset');
}
```

In fact, I guess in principle *all* of the title attributes want this kind of logic - if there's no title there should be no attributes *describing* the title. There are a few reasons for this: efficiency (don't do unnecessary work), validation (tell the user if they provided attributes that aren't used), RCE (the editor should only allow you to edit attributes that will do something)

But you're right that it gets awkward if that leads to the whole title container not existing. What if we give all the `title.text` attributes `dflt: ''`? That would ensure the `title` container always exists, and in fact would be good for RCE, since the `title.text` attribute will always be there as a signal that it can be edited. Does that break anything?

rmoestl: > What if we give all the `title.text` attributes `dflt: ''`

This won't have any effect on any titles but `carpet` axes titles. Why? All `supplyDefaults` (except `carpet` and `pie`) supply a default title along the lines of `coerce('title', layoutOut._dfltTitle.plot);`. These default titles are defined in `plots.supplyDefaults` by the way: 
```js
newFullLayout._dfltTitle = {
    plot: _(gd, 'Click to enter Plot title'),
    x: _(gd, 'Click to enter X axis title'),
  ...
};
```
`pie.title` has `dflt: ''`. So in the light of that, I suggest to simply do the same with `carpet` axes titles.

Okay with that?

> `if(title) { // what's up with the (title && title.length === 0) clause? useless, right?`

:smile: Yeah, I was wondering about that clause as well. In fact, I was wondering if there was any JavaScript weirdness I don't yet know that suggests to do things this way. But yeah, I also came to the conclusion that this clause is useless.

Meta: Part of the problem I think is that in this particular line the programmer hasn't expressed his intention well enough (i.e. is title empty?). That's why I tend to extract such conditions into local variables whose names express my intend – you may noticed that when reviewing my contributions. I haven't researched it, but I guess that JS build chains or at least JS engines are inlining these extractions back into `if` anyways (at least in Java JVM land that's the case). So I'd hope that this won't hurt memory and runtime efficiency.

alexcjohnson: > So in the light of that, I suggest to simply do the same with carpet axes titles.

Agreed, thanks for clarifying.

> So I’d hope that this won’t hurt memory and runtime efficiency. 

Clarity >> micro-optimization - you’re probably correct about this getting optimized away but even if not, small price to pay for more maintainable code.

rmoestl: :+1: Done.

src/traces/carpet/axis_defaults.js
```diff
@@ -203,6 +203,11 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, options)
     // but no, we *actually* want to coerce this.
     coerce('tickmode');
 
+    if(!containerOut.title.text) {
+        delete containerOut.title.font;
+        delete containerOut.title.offset;
+    }
```
alexcjohnson: Is there some reason it doesn't work to invert this pattern? ie only coerce these attributes if we have a title, rather than always coercing them then deleting if there's no title?

There is a lot of this pattern above - coerce everything and then delete - and it's needed when a value provided to one of these other attributes affects the default of the controlling attribute - like if there's an explicit start line color or width we default to show the start line, so we need to coerce `startlinecolor` and `startlinewidth` before we can correctly coerce `startline` but then if start line was explicitly hidden we need to delete color and width. But I don't think that applies here.

src/components/colorbar/defaults.js
```diff
@@ -59,7 +59,7 @@ module.exports = function colorbarDefaults(containerIn, containerOut, layout) {
     handleTickLabelDefaults(colorbarIn, colorbarOut, coerce, 'linear', opts);
     handleTickMarkDefaults(colorbarIn, colorbarOut, coerce, 'linear', opts);
 
-    coerce('title', layout._dfltTitle.colorbar);
-    Lib.coerceFont(coerce, 'titlefont', layout.font);
-    coerce('titleside');
+    coerce('title.text', layout._dfltTitle.colorbar);
+    Lib.coerceFont(coerce, 'title.font', layout.font);
+    coerce('title.side');
```
alexcjohnson: Oh hmm, I was going to say we should not coerce `title.font` and `title.side` here but because of `_dfltTitle` you'd have to explicitly set `title.text = ''` to make these irrelevant, and even then in principle in `editable: true` mode we would need these attributes. So I guess here - and everywhere else the titles are editable within the plot - it's correct as is. And the only places that doesn't apply are carpet (discussed elsewhere) and pie (already handled correctly) 🎉

rmoestl: I agree. That was my concern as well (plus my guess that we might need `font` to calculate things like automargin even if we don't have a title set). But I resisted to raise that because I haven't done the research in the code to underpin my claim.

src/traces/carpet/axis_defaults.js
rmoestl: > Is there some reason it doesn't work to invert this pattern?

I don't think so, but I can't say for sure TBH. What could be a reason for checking for an empty title and deleting title attributes at the end? In theory, every call to `coerce` could clear `containerOut.title`. But since there's no comment on the "why", let's reverse the logic as you suggested and keep :crossed_fingers: we haven't broke anything. Does that sound like a plan?

alexcjohnson:
Beautiful. As I was giving the code a final look I saw a few extra "oh wait, what about..." only to find you'd already addressed and thoroughly tested them all. Love it.

💃

etpinard:
I'm going to add a second :dancer: here. Amazing work @rmoestl !

PR #3280, opened 2018-11-22
implement watermark config option

### LINKED ISSUE #3268: watermark config key
Per https://github.com/plotly/streambed/issues/10708 I'd like to add a default-false boolean key to `config` called `watermark` (or similar) which, when true, would cause the modebar logo to always be visible, even if the modebar is not.

antoinerg:
Closes #3268 exactly as described/requested by @nicolaskruchten's in his first comment https://github.com/plotly/plotly.js/issues/3268#issue-382485186


src/css/_modebar.scss
```diff
@@ -5,12 +5,12 @@
     z-index: 1001;
 }
 
-.modebar--hover {
+.modebar--hover > :not(.watermark) {
```
antoinerg: Apparently `:not` is supported IE9+: https://caniuse.com/#search=%3Anot

antoinerg: CSS child selector should also work in IE9+. The reason we use it is because we cannot override the `opacity` of a child if the parent has a CSS `opacity`.

etpinard: Nice find. But we should test this out in IE9 and IE11 (from browserstack) before merging this thing.

src/plot_api/plot_config.js
```diff
@@ -133,6 +133,9 @@ module.exports = {
     // add the plotly logo on the end of the mode bar
     displaylogo: true,
 
+    // watermark the images with the company's logo
+    watermark: false,
```
etpinard: Yesss! `false` by default :ok_hand:

etpinard:
Looking good!

It should be possible to add the a few test case in 

https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/modebar_test.js

to :lock: down the new `config.watermark` logic.

src/css/_modebar.scss
antoinerg: Works in IE11:
ie11_watermark

IE9/10 doesn't work on master :neutral_face: 
ie9_fail_master2

I will open an issue for the latter.

etpinard: Oh wait, right the main plotly.js bundle doesn't work in ie9.

You'll need to use `plotly-basic.js`:

- https://cdn.plot.ly/plotly-basic-latest.min.js, or
- from https://github.com/plotly/plotly.js/blob/master/lib/index-basic.js

antoinerg: Works in IE9 :tada: 

ie9_watermark

About using `plotly-basic.js` in IE9 versus `plotly.js`: is this documented somewhere? If not I should open an issue over at the documentation project.

etpinard: More precisely, all [SVG-based traces](https://github.com/plotly/plotly.js/blob/922bad611cd8a48f77ab2086d6a935b9eebab6f5/test/jasmine/bundle_tests/ie9_test.js#L3-L15) should work in IE9. 

We could maybe add some info here ->

https://github.com/plotly/plotly.js/tree/master/dist#to-support-ie9

etpinard:
:dancer: nicely done.

Issue #3262, opened 2018-11-19
Broken rendering with flat surface and scatter3d traces


slishak:
https://codepen.io/anon/pen/KrXrgJ

In the above codepen, the 3D axis appears to have been corrupted. This only appears to occur when there are both a surface and scatter3d, all the z coordinates are identical, and the z coordinate can not be expressed as `2^n` where n is an integer.

With `z = Math.pow(2, -1)`:
![image](https://user-images.githubusercontent.com/12515343/48703236-5f1b3900-ebf3-11e8-87c3-390b28b15abc.png)

With `z = Math.pow(2, -1) + 0.1`:
![image](https://user-images.githubusercontent.com/12515343/48703217-5460a400-ebf3-11e8-8dc9-aa578ea48024.png)

etpinard:
Thanks for the report.

archmoj:
This [demo](https://codepen.io/MojtabaSamimi/pen/bQZEJZ?editors=0010) illustrates a similar issue when having a `mesh3d` with a surface. So it is not related to `scatter3d` and the bug is likely on the `surface3d` side.

Issue #2208, opened 2017-12-15
Hover picking over this surface is off


etpinard:
First reported on https://community.plot.ly/t/3d-contained-surface-not-in-accordance-with-the-3d-containing-space/7327

See https://codepen.io/etpinard/pen/xpwGgV?editors=0010

![peek 2017-12-15 10-48](https://user-images.githubusercontent.com/6675409/34049457-7694e726-e185-11e7-958e-04b0592d09ae.gif)

etpinard:
cc @archmoj this may be fixed by your work in https://github.com/plotly/plotly.js/pull/3281

archmoj:
This is a closed issue? It may not be addressed there.
@etpinard Anyway I updated your codepen to point to look at the branch [here](https://codepen.io/MojtabaSamimi/pen/RqBLLZ).

archmoj:
Revised [Demo](https://codepen.io/MojtabaSamimi/pen/pQZdLb) with better fix on another branch.

etpinard:
@archmoj did https://github.com/plotly/plotly.js/pull/3299 fix this one?

archmoj:
@etpinard Yes it would. Here is the updated [demo](https://codepen.io/MojtabaSamimi/pen/xmEbLm) using latest code, in case you were interested to test.

archmoj:
Fixed in #3281.

Issue #2812, opened 2018-07-12
Implement "connectgaps" for surface traces


benauriemma:
I'm trying to create a surface plot, but it is not always the case that every point within the range I'm plotting over is defined.  That is, my z matrix might resemble:
```json
[
    [0, 1, 2, 3],
    [1, 2, 3, 4],
    [2, 3, undefined, 5],
    [3, 4, 5, 6]
]
```

If I try to create a surface plot using the above as the z matrix, the surface gets cut off and not all of the data is plotted.  If I replace ```undefined``` with ```null``` or ```NaN```, the same happens.  If I replace ```undefined``` with empty, the code fails.

The only way I can get the rest of the data to plot is by replacing ```undefined``` with ```0```.  However, this "pulls" the surrounding surface down to 0, which is misleading and especially problematic when all other data is much greater than 0.

I would be shocked if plotting an incomplete dataset is not a very common use case, and currently there's no good way to do it.  I believe it would be usefull to provide a more intelligent handling for missing data.  This could be leaving part of the surface out, or interpolating the missing value from surrounding values.

etpinard:
Adding a way to interpolate (assuming that's what you mean by "intelligent handling") for missing `z` values in surface would be great addition. In plotly.js lingo, we call this `connectgaps: true` as is already available for heatmap and contour traces. 

@benauriemma ok if change the issue title accordingly?

benauriemma:
@etpinard , yes, I think that would be a great way of handling it intelligently.  Github is not allowing me to change the title of the issue right now, but I will try again later.

"connectgaps for surface plots with missing data"

PR #3281, opened 2018-11-22
Improving the quality of surface objects to better follow the data by using LCM or highly composite numbers

### LINKED ISSUE #2713: Surface plots do not quite follow the data
See https://codepen.io/alexcjohnson/pen/gKLBgR?editors=0010
screen shot 2018-06-08 at 5 00 44 pm

I was trying to get the contour lines to exactly match the data, but notice that the line at `y=1.3` (last black line in the red) is not quite straight (even though the data at `y=1.3` is uniformly `1.0`) and the line at `y=1.4` (first black line in the green) does not quite reach the correct value of `0.898` like its neighbors at larger `y` do. Digging in a bit deeper, it seems like the contours are following the surface very closely - perhaps not precisely, there are still a few spots where they seem to disappear behind the surface, but it's pretty close. But the surface itself seems to disagree a little bit with the data, which you can see by zooming in near one of these corners:
screen shot 2018-06-08 at 2 47 27 pm
It looks like either a) the surface is smoothing a little bit around these sharp corners, or b) the surface is actually creating a finer mesh than the data points, and this mesh is not aligned with the data so it misses the corners. Either way it's not doing the right thing.

cc @etpinard as mentioned in #2712

archmoj:
Fixes #2713 by increasing the resolution of surface objects.
With previous set up resolution (128x128) surfaces with any remarkable changes in z-axis were not displayed well.
Also to mention: the issue #2208 and #3262 would be fixed in this PR as well.
@etpinard 
@alexcjohnson


test/jasmine/tests/gl3d_plot_interact_test.js
```diff
@@ -66,7 +66,7 @@ describe('Test gl3d plots', function() {
         gd = createGraphDiv();
         ptData = {};
 
-        jasmine.DEFAULT_TIMEOUT_INTERVAL = 4000;
+        jasmine.DEFAULT_TIMEOUT_INTERVAL = 8000;
```
etpinard: Hmm. I was about to ask if you noticed any performance deterioration on your branch. I guess this here answers my question. :thinking:

etpinard: ... would there be a way to increase the resolution just for traces that _need it_? In other words, is a way to determine if a given `z` 2D array requires extra resolution?

archmoj: One idea could be to allow resetting this resolution via the api.
I can imagine of auto-detect solutions; but considering animations I suggest we keep this constant.

archmoj: @etpinard Rather than using `1024`, now by simply using `120` instead of `128' the performance is untouched; while the quality of the graphs are remarkably improved. If we want to go further than that the next magic values that could be applied are 180, 240, 360, 720 and 840.

etpinard: Fantastic! Could you explain briefly why `120` is better than `128`?

archmoj: 128 = 2<sup>7</sup> so it is divisible by (1 + 7) numbers {1, 2, 4, 8, 16, 32, 64, 128}
120 = 2<sup>3</sup>x3<sup>1</sup>x5<sup>1</sup> therefore divisible by (1+3)x(1+1)x(1+1) = 16 numbers {1, 2, 3, 4, 5, 6, 8, 10, 12, 15, 16, 20, 24, 30, 60, 120}.
So it is more likely that the refined grid match actual data points in this case.
https://en.wikipedia.org/wiki/Highly_composite_number
Base on this investigation I am now thinking to rewriting the function that refines scale.

etpinard: Wow @archmoj is the man! Thanks!

archmoj:
@alexcjohnson & @etpinard
Kindly find the details below concerning the performance.
.
gl3d_contour-lines2 
call: first, second, third
old: 1186ms, 368ms, 230ms
new: 1239ms, 392ms, 222ms
.
gl3d_surface-lighting 
call: first, second, third
old: 735ms, 395ms, 300ms
new: 735ms, 395ms, 250ms
.
gl3d_xy-defined-ticks 
call: first, second, third
old: 787ms,388ms, 324ms
new: 795ms,384ms, 322ms

test/image/mocks/gl3d_ibm-plot.json
```diff
@@ -196,21 +196,6 @@
                     20.0538,
                     17.7003,
                     15.4519
-                ],
```
archmoj: Note: this extra unused data impacted the `gl3d_ibm-plot.json` (i.e. the undesirable highlighted edge on the right).

alexcjohnson: Interesting... Here's what `contour` does in a similar situation:
screen shot 2018-11-26 at 7 33 12 pm
It seems to extrapolate the last differences in x & y. That's a better analog for `surface` than `heatmap`, where both n and n+1 points in x or y compared to the same dimension of z have a clear interpretation.
Doesn't need to be addressed now, but at some point we should probably ensure all our 2D mappings handle this case identically.

archmoj: There is also this related to interpolation. https://github.com/plotly/plotly.js/issues/2812

package.json
```diff
@@ -84,7 +84,7 @@
     "gl-select-box": "^1.0.2",
     "gl-spikes2d": "^1.0.1",
     "gl-streamtube3d": "^1.1.1",
-    "gl-surface3d": "^1.3.7",
+    "gl-surface3d": "git://github.com/gl-vis/gl-surface3d.git#2a1728fae4d9ee30e0c34e945859234cc489e32a",
```
archmoj: https://github.com/gl-vis/gl-surface3d/pull/14

src/traces/surface/convert.js
```diff
@@ -342,46 +373,87 @@ proto.update = function(data) {
...
+            rawCoords[1][j][k] = this.getYat(j, k, data.ycalendar, sceneLayout.yaxis);
+            rawCoords[2][j][k] = this.getZat(j, k, data.zcalendar, sceneLayout.zaxis);
+        }
+    }
 
-    fill(coords[2], function(row, col) {
-        return zaxis.d2l(z[col][row], 0, zcalendar) * scaleFactor[2];
-    });
+    // Note: log axes are not defined in surfaces yet.
+    // but they could be defined here...
```
etpinard: They seem to work ok in `latest`: https://codepen.io/etpinard/pen/MzZLap?editors=0010

src/traces/surface/convert.js
```diff
@@ -104,59 +129,212 @@ function isColormapCircular(colormap) {
...
+    for(var i = 0; i < shortPrimes.length; i++) {
+        n *= Math.pow(
+            shortPrimes[i], Math.max(A[i], B[i])
+        );
     }
+    return n;
+}
 
-    return 1.0;
+function ayyarLCM(A) {
```
etpinard: What does `ayyar` stand for here?

alexcjohnson: haha I'm assuming y<->r

etpinard: Typo of the week!

src/traces/surface/convert.js
archmoj: A [demo](https://codepen.io/MojtabaSamimi/pen/GwPzOR?editors=0010) displaying the issue with logs in z direction. Noting that similar issue may also happen in x and y directions when using log.

archmoj:
Good news! It seems that as another result of this PR, issue #3262 would also be fixed. Here is a [demo](https://codepen.io/MojtabaSamimi/pen/gQErrq).

archmoj:
@alexcjohnson we were wondering if you could please review the latest changes of this PR?
Thank you.

src/plots/gl3d/scene.js
```diff
@@ -380,6 +380,11 @@ function computeTraceBounds(scene, trace, bounds) {
                 }
             }
         }
+
+        if(trace.type === 'surface') {
+            bounds[0][d] += trace._objectOffset[d];
+            bounds[1][d] += trace._objectOffset[d];
+        }
```
alexcjohnson: @archmoj as discussed just now let's get a clear test case (mock I guess) where `_objectOffset` matters - offhand this looks weird to me as we're recalculating min and max from the input data `trace.(x|y|z)` - which is redundant anyway since you already did this in `surface/convert` - and then adding the `_objectOffset` to it, which seems like double-counting the middle.

alexcjohnson:
@archmoj as discussed the improved grid refinement system is a big improvement for a good number of cases, very nice! I still think we'll do better if we don't refine the grid at all, and simply use the coordinates we're given - performance as well as probably some cases it still doesn't draw the surface quite right. But that can wait for later, perhaps the regl refactor whenever we might get to that.

src/traces/surface/convert.js
```diff
@@ -104,59 +129,212 @@ function isColormapCircular(colormap) {
...
+};
+
+var highlyComposites = [1, 2, 4, 6, 12, 24, 36, 48, 60, 120, 180, 240, 360, 720, 840, 1260];
+
+var MIN_RESOLUTION = highlyComposites[9];
+var MAX_RESOLUTION = highlyComposites[13];
+
+proto.estimateScale = function(resSrc, axis) {
+    // console.log("axis=", axis);
+    // console.log("resSrc=", resSrc);
```
alexcjohnson: 🔪

alexcjohnson:
> Regarding our discussion concerning computing the ranges twice ... I could still try using `Lib.aggnums` there.

There are two separate issues here: DRY 🌴 and performance 🐎.

🌴 we could investigate replacing *both* of these implementations with `aggNums` but if it's at all complicated don't worry about it.

🐎 I would just imagine keeping both copies, but short-circuiting the one in `scene` if it has already been done (and stashed on the trace) in `convert`.

test/image/mocks/gl3d_surface_contour_precision.json
```diff
@@ -0,0 +1,33 @@
+{
+    "data": [
```
alexcjohnson: I love this mock! Unfortunately I'm seeing some more weird behavior when I play with it... all these axis features flicker above & below the data.
screen shot 2018-12-06 at 6 29 11 pm

archmoj: Those two mocks are intended to possibly show improvements and at the same time certain remaining problems related to data precision. Are you experiencing this on the latest release as well? I've just fetch master as there was another [PR](#3302) that could help.

archmoj: Concerning @alexcjohnson observation, I tested another new mock with CI on current master branch (i.e. without these changes applied); and the same z-fighting existed [there](https://20938-45646037-gh.circle-artifacts.com/0/test_images/gl3d_parametric_surface_data_precision.png). Now I may confirm that this problem is not related to the changes of this branch related to objectOffset.

package.json
```diff
@@ -84,7 +84,7 @@
     "gl-select-box": "^1.0.2",
     "gl-spikes2d": "^1.0.1",
     "gl-streamtube3d": "^1.1.1",
-    "gl-surface3d": "^1.3.7",
+    "gl-surface3d": "git://github.com/gl-vis/gl-surface3d.git#0d9201dfdc3bee7c4157e4602357d28f75ba2b00",
```
archmoj: https://github.com/gl-vis/gl-surface3d/pull/14

src/traces/surface/convert.js
archmoj: This log issue is now reported here https://github.com/plotly/plotly.js/issues/3312

alexcjohnson:
Looks good to me! 💃 after merging `gl-surface3d` and updating its ref, and making an issue for https://github.com/plotly/plotly.js/pull/3281#discussion_r242356771

etpinard:
:tada: this will be part of 1.43.0!

test/image/mocks/gl3d_surface_contour_precision.json
archmoj: OK. That issue is now reported in #3346.

Issue #3277, opened 2018-11-21
Fill tozeroy with a null filled y array errors out


GlennMatthys:
In a trace where 'y' is all nulls, Plotly errors out with "Cannot read property 'fillcolor' of undefined" when fill is set to 'tozeroy'.

![image](https://user-images.githubusercontent.com/31989842/48847463-e8706e00-eda1-11e8-979c-2cecbeb41e9b.png)

[JSFiddle](https://jsfiddle.net/GlennMatthysECrane/9tky5gLz/1/)

Expected result: Plotly draws the graph without erroring out.

PR #3282, opened 2018-11-22
Fix fills on segment-less marker-less traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3277 - introduced by https://github.com/plotly/plotly.js/commit/1f3c10c735927d57b78151be8c4d43bdf35e53d7

cc @plotly/plotly_js


alexcjohnson:
Sure, I was going to suggest deleting the element so we don't need to trust as much that we always set the data, but since it's a singleton element that would be awkward. So, looks good. 💃

etpinard:
> but since it's a singleton element that would be awkward

My thoughts exactly.

Issue #168, opened 2016-01-07
eventdata content consistency


aderidder:
Hi,

In order to capture the points in my scatterplot, I'm currently using plotly_relayout. After playing around with this for a bit, I found that the eventdata is somewhat unpredictable, making it hard/confusing to work with:
- autoscale button clicked: Object {xaxis.autorange: true, yaxis.autorange: true}
- zoom button clicked: Object {xaxis.range: Array[2], yaxis.range: Array[2]}
- drag zoom: Object {xaxis.range[0]: <value>, xaxis.range[1]: <value>, yaxis.range[0]: <value>, yaxis.range[1]: <value>}
- drag zoom, but leave your plot at the left side: Object {xaxis.range[1]: <value>, yaxis.range[0]: <value>, yaxis.range[1]: <value>}
- pan: similar to drag zoom, but if you drag along an axis (e.g. don't change the y, only the x), the eventdata only contains information about the x

It would be very useful if the eventdata's content would be consistent, e.g. always provide Object {xaxis.range: Array[2], yaxis.range: Array[2]}

Thanks,
Sander

etpinard:
@aderidder great suggestion. Thanks!

The `plotly_relayout` is in dire need of a redesign. Maybe the solution lies in https://github.com/plotly/plotly.js/issues/145 where *axis* each interaction would emit an event with a similar data object.

chrisgervang:
It appears the rangeslider pan and zoom is consistently `{xaxis.range: Array[2], yaxis.range: Array[2]}`. So to handle relayout event I now do this:

```
const newSliderRange: [number, number] = e['xaxis.range']; //Plotly format for zooming with the slider
const newZoomRange: [number, number] = [e['xaxis.range[0]'], e['xaxis.range[1]']]; //Plotly format for zooming within the graph

const newRange = !!newZoomRange[0] || !!newZoomRange[1] ? newZoomRange : newSliderRange;
console.log("EVENT", e, newRange);
```

jackparmer:
Hiya! Plotly now has improved and consolidated Plotly.js event documentation here:

https://plot.ly/javascript/plotlyjs-events/

Please feel free to suggest documentation improvements in a pull request:

https://github.com/plotly/documentation/tree/source-design-merge/_posts/plotly_js/events

etpinard:
Pie and Sankey event data should be refactor during the v2 push to look more like the other traces.

See https://github.com/plotly/plotly.js/issues/1875

[...]

PR #3284, opened 2018-11-23
hovertemplate support for sankey

### LINKED ISSUE #3269: implement `hovertemplate` for Sankey traces
Custom HTML code for both links and nodes.

antoinerg:
Closes #3269

### DIFF
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -13,6 +13,7 @@ var plotAttrs = require('../../plots/attributes');
 var colorAttrs = require('../../components/color/attributes');
 var fxAttrs = require('../../components/fx/attributes');
 var domainAttrs = require('../../plots/domain').attributes;
+var hovertemplateAttrs = require('../../components/fx/hovertemplate_attributes');
 
 var extendFlat = require('../../lib/extend').extendFlat;
 var overrideAll = require('../../plot_api/edit_types').overrideAll;
@@ -146,6 +147,9 @@ var attrs = module.exports = overrideAll({
             ].join(' ')
         },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
+        hovertemplate: hovertemplateAttrs({}, {
+            keys: []
+        }),
         description: 'The nodes of the Sankey plot.'
     },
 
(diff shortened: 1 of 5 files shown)


antoinerg:
I am not sure the description of `hovertemplate` attributes is satisfactory. I'd like to discuss the changes made in commit d952713.

Finally, because the inclusion `hovertemplate` in `Fx.loneHover` calls have been tested in other traces, is it necessary to test it again in `sankey_test.js`? Is it OK to only check for the existence of the expected variables in eventData?

src/traces/sankey/attributes.js
```diff
@@ -146,6 +147,10 @@ var attrs = module.exports = overrideAll({
             ].join(' ')
         },
         hoverlabel: fxAttrs.hoverlabel, // needs editType override,
+        hovertemplate: hovertemplateAttrs({}, {
```
etpinard: Hmm. Sankey event data is a bit of mess:

image

image

There's probably way more stuff it in than we need. Moreover, some of it isn't consistent with event data for other traces e.g. the `trace` key should be `fullData`. I'm not how much we can change it w/o making breaking changes.

Oh well, these descriptions look ok to me.

etpinard:
> Finally, because the inclusion `hovertemplate` in `Fx.loneHover` calls have been tested in other traces, is it necessary to test it again in `sankey_test.js`?

I would add one "true" `hovertemplate` test case where we check that a user-defined `hovertemplate` gives the correct hover label content. You're right though, no need to re-test all the `loneHover` logic again :ok_hand:

src/traces/sankey/attributes.js
antoinerg: I added `fullData` to the eventData in commit 3204274

antoinerg:
> I would add one "true" `hovertemplate` test case where we check that a user-defined `hovertemplate` gives the correct hover label content.

Done in commit 3204274!

src/traces/sankey/attributes.js
etpinard: v2 changes tracked in https://github.com/plotly/plotly.js/issues/168#issuecomment-315874345

etpinard:
@antoinerg oops I forgot about this PR.

Looks good :dancer:

PR #3285, opened 2018-11-23
Add (some) IE11 testing on AppVeyor

etpinard:
resolves https://github.com/plotly/plotly.js/issues/431

Looks like AppVeyor has gotten a lot better in the last three years, adding IE11 tests was relatively easy.

Note that this iteration only tests the `lib_test.js` suite. We should add more IE11 tests eventually, perhaps by adding a specialty suite in `test/jasmine/bundle_tests/ie11_test.js` or by adding a `@ie11` jasmine tag.

But as discussed in https://github.com/plotly/plotly.js/issues/431#issuecomment-435079403, adapting all our jasmine tests (especially our pixel-based tests) might be too much of a pain for little benefit.

cc @plotly/plotly_js

### DIFF
--- a/test/jasmine/tests/lib_test.js
+++ b/test/jasmine/tests/lib_test.js
@@ -1558,12 +1558,16 @@ describe('Test lib.js:', function() {
 
     describe('cleanNumber', function() {
         it('should return finite numbers untouched', function() {
-            [
-                0, 1, 2, 1234.567,
-                -1, -100, -999.999,
-                Number.MAX_VALUE, Number.MIN_VALUE, Number.EPSILON,
-                -Number.MAX_VALUE, -Number.MIN_VALUE, -Number.EPSILON
-            ].forEach(function(v) {
+            var vals = [0, 1, 2, 1234.567, -1, -100, -999.999];
+
+            if(!Lib.isIE()) {
+                vals.push(
+                    Number.MAX_VALUE, Number.MIN_VALUE, Number.EPSILON,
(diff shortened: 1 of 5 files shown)


test/jasmine/tests/lib_test.js
```diff
@@ -1558,12 +1558,16 @@ describe('Test lib.js:', function() {
...
-                Number.MAX_VALUE, Number.MIN_VALUE, Number.EPSILON,
-                -Number.MAX_VALUE, -Number.MIN_VALUE, -Number.EPSILON
-            ].forEach(function(v) {
+            var vals = [0, 1, 2, 1234.567, -1, -100, -999.999];
+
+            if(!Lib.isIE()) {
+                vals.push(
+                    Number.MAX_VALUE, Number.MIN_VALUE, Number.EPSILON,
+                    -Number.MAX_VALUE, -Number.MIN_VALUE, -Number.EPSILON
+                );
```
alexcjohnson: Oh dear, IE doesn't support these? We use `MAX_VALUE`:

https://github.com/plotly/plotly.js/blob/077a1d770cb7dc9fa8c2588bfb64a80607cf9095/src/constants/numerical.js#L26

etpinard: fixed in -> https://github.com/plotly/plotly.js/pull/3285/commits/50218deb708cab59260df5bf5096d5a61988eb2d

alexcjohnson:
Excellent! 💃 

I like the idea of an `@ie11` tag. That would make it nice and easy (and local) to add new tests to ie.

Issue #3279, opened 2018-11-22
Missing labels when y-axis reverse in contour


poldz123:
When using the layout for contour:
```
var layout = {
      yaxis: {
        autorange: "reversed",
        fixedrange: true
      },
      xaxis: {
        side: "top",
        fixedrange: true
      },
      margin: {
          l: 40,
          r: 50,
          t: 30,
          b: 10
        }
    };
```

It is missing the labels of the contour but if the `autorange` is removed it is back again.

etpinard:
Seems to work as expected in: https://codepen.io/etpinard/pen/vQjNEM?editors=0010

Can you share a "fully" reproduce example to help us out? Thank you.

PR #3293, opened 2018-11-26
Fix contour label px bounds for reversed axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3279 - cc @plotly/plotly_js


alexcjohnson:
Nice, thanks! 💃

PR #3301, opened 2018-12-02
Prevent scatter3d to only display the first point on hover: issue 3258 

### LINKED ISSUE #3258: scatter3d always selects point 0 on mouse hover
From version 1.34.0 and onward, mouse hover on `scatter3d` always picks point number 0. Version 1.33.1 works fine. The problem exists in both Chrome and Firefox on Ubuntu 16.04 and Ubuntu 18.04. Here is a small example:

```js
'use strict';
{
  async function main(){
    const data = [
      {
        type: 'scatter3d',
        mode: 'markers',
        x: Array.from( { length: 25 }, () => Math.random() ),
        y: Array.from( { length: 25 }, () => Math.random() ),
        z: Array.from( { length: 25 }, () => Math.random() )
      }
    ];
    const layout = {};

    const plot = document.createElement('div')
    plot.style = 'width: 90vw; height: 90vh'
    document.body.appendChild(plot)

    await Plotly.plot(plot, data, layout)
    plot.on('plotly_hover', event => console.log(event) )
  }
  main()
}
```

And here is a screenshot of the bug (notice where the mouse is pointing):
plotly_scatter3d_hover_bug

archmoj:
Fixes #3258 for Ubuntu.
Bug was related to linking issues within gl-scatter3d module...
@etpinard 
@alexcjohnson


etpinard:
Thanks for investigating this bug @archmoj !

> It seems Linux tries to optimise the performance by removing the alpha channel (sometimes after warm up)

sounds like it was very painful to debug :muscle: 

---

Your fix does make things work on my Ubuntu machine. Now, it does strike me a little bit _hacky_. I mean an opacity of `0.99` isn't really an opacity of `1`, and probably makes things render more slowly as some the aforementioned optimization is now skipped. 

I wonder if we could instead pass a flag into `gl-scatter3d`'s `pickHover` routine telling it "all points in this trace have `opacity: 1`"?

package.json
```diff
@@ -80,7 +80,7 @@
     "gl-plot2d": "^1.4.0",
     "gl-plot3d": "^1.6.0",
     "gl-pointcloud2d": "^1.0.1",
-    "gl-scatter3d": "^1.1.0",
+    "gl-scatter3d": "git://github.com/gl-vis/gl-scatter3d.git#c252e1cd723aed657bdc8b022efe6befcfe9f729",
```
archmoj: https://github.com/gl-vis/gl-scatter3d/pull/13

archmoj:
Thank @etpinard for encouraging! It was quite a bad bug. Different approach is used in this revision. No need to tweak the opacity and the baselines didn't changed.

log.txt
```diff
@@ -0,0 +1,4458 @@
+
+> plotly.js@1.42.5 test-jasmine /home/mojtaba/plotly/webgl/plotly.js
```
etpinard: :zap:

Issue #3075, opened 2018-10-04
rangeslider and autorange: "reversed" incompatibility


LloydLS:
Hi,

`rangeslider` does not seem to work well with a reversed x axis. Double clicking to autorange the main plot always sets the x axis back to an ascending direction. Also, setting a `range` under `layout.xaxis` disables the rangeslider's normal operation, and makes the two plots act the same (rather than being able to make a small "rectangular" area for the main plot to show).

```
"data": [
    {
        "name": "Points",
        "type": "line",
        "mode": "lines",
        "hoverinfo": "x",
        "showlegend": false,
        "line": {
            "color": "blue",
            "shape": "spline"
        },
        "x": "[-1,0,1,2,3,4,5,6,7,8,9,10,11]",
        "y": "[0,10,20,30,40,50,60,70,80,90,100,110,120]"
    },
],
"layout": {
    "title": "",
    "xaxis": {
        "title": "",
        "zeroline": false,
        "autorange": "reversed",
        "rangeslider": {
            "yaxis": {
               "rangemode": "match"
           },
           "autorange": false,
           "range": [
               11,
               0
           ],
           "visible": true,
       }
   },
   "yaxis": {
    "visible": false
    }
}
```

Thanks! (Plotly is great btw!)

etpinard:
CodePen: https://codepen.io/etpinard/pen/LgNwRN?editors=1010

Thanks for the report.

PR #3304, opened 2018-12-03
Rangeslider on reversed axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3075

cc @plotly/plotly_js


alexcjohnson:
Very nice fix! But what's up with the fuzzy line in the test image (between 10&9, 8&7, 3&2)?
screen shot 2018-12-04 at 4 42 40 pm

etpinard:
> But what's up with the fuzzy line in the test image (between 10&9, 8&7, 3&2)?

Looks like it's a weird nw.js artifact. Here's how it looks like using orca:

image

PR #3307, opened 2018-12-05
remove "Edit in Chart Studio" button by default

### LINKED ISSUE #316: Should the 'send to cloud' option be disabled by default?
The CDN hosted version of the library includes the toolbar link to immediately copy the data elsewhere. It would be nice to use the CDN version without this seemingly insecure option that can be clicked accidentally. There does not seem to be a way to disable this option. The Codepen examples all use a custom build and not the CDN version, so this issue is not apparent on the examples.

Alternatively, perhaps hosting can be done at CDNJS or Google if this button is a condition of the cost of the CDN hosting.

alexcjohnson:
Closes #316 - removes the "Edit in Chart Studio" button by default, and adds a new config arg `showSendToCloud` to re-enable it.

cc @etpinard @nicolaskruchten @chriddyp any comments in particular on the name of this new arg?

### DIFF
--- a/src/components/modebar/manage.js
+++ b/src/components/modebar/manage.js
@@ -64,7 +64,8 @@ module.exports = function manageModeBar(gd) {
         buttonGroups = getButtonGroups(
             gd,
             context.modeBarButtonsToRemove,
-            context.modeBarButtonsToAdd
+            context.modeBarButtonsToAdd,
+            context.showSendToCloud
         );
     }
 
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
Works for me.

etpinard:
:dancer:

nicolaskruchten:
We’ll definitely need to adjust stuff in streambed when bumping the dep there

PR #3317, opened 2018-12-10
Axis labels applied in the hovering popup

### LINKED ISSUE #2618: Axis title could also change the labels in the hovering popup

archmoj:
Fix #2618 
@plotly/plotly_js

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -944,7 +944,7 @@ function createHoverText(hoverData, opts, gd) {
             yLetter = gd._fullLayout.scene.yaxis.title.text || 'y';
             zLetter = gd._fullLayout.scene.zaxis.title.text || 'z';
         }
-        else if(!gd._fullLayout.ternary) {
+        else if(!gd._fullLayout.ternary && !gd._fullLayout.title) {
             xLetter = gd._fullLayout.xaxis.title.text || 'x';
             yLetter = gd._fullLayout.yaxis.title.text || 'y';
         }
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -944,7 +944,7 @@ function createHoverText(hoverData, opts, gd) {
             yLetter = gd._fullLayout.scene.yaxis.title.text || 'y';
             zLetter = gd._fullLayout.scene.zaxis.title.text || 'z';
         }
-        else {
+        else if(!gd._fullLayout.ternary) {
             xLetter = gd._fullLayout.xaxis.title.text || 'x';
             yLetter = gd._fullLayout.yaxis.title.text || 'y';
         }
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -935,10 +935,24 @@ function createHoverText(hoverData, opts, gd) {
             }
         }
 
+        var xLetter;
+        var yLetter;
+        var zLetter;
+
+        if(gd._fullLayout.scene) {
+            xLetter = gd._fullLayout.scene.xaxis.title.text || 'x';
+            yLetter = gd._fullLayout.scene.yaxis.title.text || 'y';
+            zLetter = gd._fullLayout.scene.zaxis.title.text || 'z';
+        }
+        else {
+            xLetter = gd._fullLayout.xaxis.title.text || 'x';
+            yLetter = gd._fullLayout.yaxis.title.text || 'y';
+        }
+
         if(d.zLabel !== undefined) {
-            if(d.xLabel !== undefined) text += 'x: ' + d.xLabel + '<br>';
-            if(d.yLabel !== undefined) text += 'y: ' + d.yLabel + '<br>';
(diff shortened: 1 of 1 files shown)


etpinard:
Hmm. I think I would prefer an "opt-in" version - at least in v1.x as some might consider this a breaking change. 

To do so, I propose adding a new attribute to add axes the support hover called: `'hovertitle'` and use that in `fx/hover.js` like you did here.

We should make sure that all axes (cartesian, 3d, ternary, polar and maybe geo and mapbox) support this new attribute.

antoinerg:
There's also the possibility of implementing `hovertemplate` for 3D.

src/plots/cartesian/layout_attributes.js
```diff
@@ -593,6 +593,15 @@ module.exports = {
             '*%H~%M~%S.%2f* would display *09~15~23.46*'
         ].join(' ')
     },
+    hovertitle: {
+        valType: 'boolean',
```
etpinard: That's not what I had in mind.

I was thinking of making `hovertitle` a string that defaults to the current behavior (in v1.x, and potentially default to the axis `title` value in v2.x).

archmoj: The `hovertitle` is now a `string` instead of a `boolean`. 
Other axis types have `_hovertitle` attribute. Wondering if we should change them to `hovertitle`?

src/components/fx/hover.js
```diff
@@ -25,6 +25,8 @@ var Registry = require('../../registry');
 var helpers = require('./helpers');
 var constants = require('./constants');
 
+var MAJOR_VERSION = 1; // i.e. Plotly version
```
etpinard: Nop. Please don't include version-specific logic. Write code for `v1.x`, we'll patch this up in a `v2` branch when we get there.

test/image/mocks/gl3d_surface_after_heatmap.json
```diff
@@ -2,28 +2,59 @@
     "data": [
       {
         "type": "heatmap",
-        "x": [0, 1, 2],
-        "y": [0, 1, 2],
+        "x": [0, 1, 2, 3, 4],
```
etpinard: Why did you patch this mock?

src/plots/cartesian/layout_attributes.js
etpinard: Sounds like those `_hovertitle` are now obsolete. Would you mind :hocho: ing them? Thanks!

src/plots/cartesian/layout_attributes.js
```diff
@@ -593,6 +593,15 @@ module.exports = {
             '*%H~%M~%S.%2f* would display *09~15~23.46*'
         ].join(' ')
     },
+    hovertitle: {
+        valType: 'string',
+        dflt: '',
```
etpinard: Remove this line. The `hovertitle` default isn't blank. As it depends on which axis where coercing, we can't declare a default value here, so the best thing we can do is to not declare any default value.

src/plots/cartesian/axis_defaults.js
```diff
@@ -56,7 +56,10 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
 
     handleCategoryOrderDefaults(containerIn, containerOut, coerce, options);
 
-    if(axType !== 'category' && !options.noHover) coerce('hoverformat');
+    if(axType !== 'category' && !options.noHover) {
+        coerce('hoverformat');
+        coerce('hovertitle');
```
etpinard: Could we give this a default value? By implementing something like:

```js
coerce('hovertitle', options.dfltHoverTitle);
```

Each `handleAxisDefaults` caller would have to pass another `options` key, but this would make the chart editor devs (aka RCE) much happier (as the default value will be automatically filled in the RCE panels) and the logic in `Fx.hover` simpler.

etpinard:
Dropping from 1.44.0.

I'd like @archmoj to focus on https://github.com/plotly/plotly.js/pull/3438

test/image/mocks/gl3d_surface_after_heatmap.json
archmoj: This was a new mock added quite recently. I thought we could expand a bit to allow interactivelytesting `hovertitle` on gl3d & gl2d.

Issue #3255, opened 2018-11-16
Plotly.react problems when switching traces from 'linear' scatter -> 'category' scatterpolar


etpinard:
See https://codepen.io/etpinard/pen/vQJKLd from https://community.plot.ly/t/swapping-from-bubble-to-scatterpolar/15602/10

alexcjohnson:
Wow, that's crazy. I should ask what the heck this line (which is resulting in a whole attempted "re"plot without any associated `calc`) is doing in `supplyDefaults`:

https://github.com/plotly/plotly.js/blob/3a32ac00cf5e8619932f874960f11b4daf7ceb40/src/plots/plots.js#L482-L483

But I'm afraid I'd have to ask myself from 4 years ago 🙄 
Hopefully we can just move this (and the `ax.setScale` loop below it) out of `supplyDefaults` and later in the process, and deal with any consequences in a more rational way!

Kind of amazed this is the first we've seen of this bug... oh well, no time like the present!

PR #3300, opened 2018-11-30
Multicategory axis type

etpinard:
(attempts to) resolves https://github.com/plotly/plotly.js/issues/2799 by adding a :sparkles: new :sparkles: axis type: `'multicategory'` - which displays two (maybe more later) axis labels levels along with "dividers" in-between each category group. 

**N.B.** This branch is up-to-date with the latest master that includes `uirevision`.

--- 

Things to TODO:

- [x] add support for `type: 'multicategory'` on y-axes
- [x] add support to heatmap and contours
- [x] add support in `histogram*` traces
- <strike>[ ] add some way to configure ticks/dividers per category level</strike> - **leave for later**
- [x] improve `axis.automargin` test coverage, **partly done in:** https://github.com/plotly/plotly.js/pull/3300/commits/80f47014b8bbfb0743fbd81ebd56295d629e3be0, more to come in https://github.com/plotly/plotly.js/issues/3305
- [x] test as many edge cases I can think of

Things skipped:

- `ticktext` / `tickvals` are not coerced on `type: 'multicategory'` axes
- support for `transforms`

cc @plotly/plotly_js @nicolaskruchten

PR #3323, opened 2018-12-11
Clean up automargin pipeline a bit

alexcjohnson:
@etpinard I based this branch off `multicategory` (including commits from this morning) because we were both working in axis drawing code.

The main goal was to get [`plots.doAutoMargin` out of `supplyDefaults`](https://github.com/plotly/plotly.js/commit/858f7c2aea73f03f06215c280dd11c3429d11584#diff-ad4f76ccd6044ed16514297078e13b84L501) - because that can lead to a full redraw of the plot if margins have changed, which is completely inappropriate for `supplyDefaults` - somehow it never caused explicit problems before, but with big enough changes in `Plotly.react`, ie #3255, it does.

It turned out the biggest challenge with this was rangesliders. The main thing I did to fix this was to combine rangeslider and axis automargins to happen [together](https://github.com/plotly/plotly.js/commit/bdabf48c298001f490368618db0bd4281ae22a89#diff-f6d5ed25a5d6f57c70063e1e1bbc407dR1886) - note though that in principle both can contribute independently to the margins, if you have a rangeslider (on the bottom) and a top axis with long labels (oh right, I meant to add a test for that case - I'll add that)

One baseline image, `range_slider_rangemode`, changed in this PR https://github.com/plotly/plotly.js/commit/b2a4b769cb73494d060fedd498357ef5f916f5ab - just a little bit, but I believe it was in fact wrong before, as you can see by interacting with this mock on master: start dragging the range and the box expands vertically a little bit, to match what's in the new baseline. On this branch there's no such jump. Not quite sure *what* in this PR fixed it 😅 but it's fixed.
range_slider_rangemode


test/jasmine/tests/splom_test.js
```diff
@@ -892,9 +892,9 @@ describe('Test splom interactions:', function() {
             return Plotly.relayout(gd, {width: 4810, height: 3656});
         })
         .then(function() {
-            assetsFnCall('after', {
-                cleanPlot: 4,       // 3 three from supplyDefaults, once in drawFramework
-                supplyDefaults: 3,  // 1 from relayout, 1 from automargin, 1 in drawFramework
+            assertFnCall('after', {
+                cleanPlot: 3,       // 2 from supplyDefaults, once in drawFramework
+                supplyDefaults: 2,  // 1 from relayout, 1 in drawFramework
```
alexcjohnson: A bit of a less iterative plotting pipeline now...

etpinard: :tada:

src/plots/cartesian/axes.js
```diff
@@ -1628,12 +1628,20 @@ axes.drawOne = function(gd, ax, opts) {
...
     var subplotsWithAx = axes.getSubplots(gd, ax);
 
     var vals = ax._vals = axes.calcTicks(ax);
 
+    // Add a couple of axis properties that should cause us to recreate
+    // elements. Used in d3 data function.
+    var axInfo = [ax.mirror, mainLinePosition, mainMirrorPosition].join('_');
+    for(i = 0; i < vals.length; i++) {
+        vals[i].axInfo = axInfo;
+    }
```
alexcjohnson: I think it was always incorrect in fact to have `makeDataFn` refer to axis attributes directly... they need to be in the data that's bound to the ticks by D3, otherwise changing them will not cause those ticks to be recreated. The right thing to do is what I do here: push the relevant axis attributes into the tick data, so the old ticks have the old axis attributes and the new ones have new attributes. Then `makeDataFn` goes away, it's just a static function `tickDataFn`.

I didn't add any new tests for this change, it was made in order to fix existing tests, which I guess were failing because we aren't doing as much full redrawing with the simplified pipeline.

src/plot_api/subroutines.js
```diff
@@ -113,11 +110,6 @@ function lsInner(gd) {
         ax._mainMirrorPosition = (ax.mirror && counterAx) ?
             getLinePosition(ax, counterAx,
                 alignmentConstants.OPPOSITE_SIDE[ax.side]) : null;
-
-        // Figure out which subplot to draw ticks, labels, & axis lines on
-        // do this as a separate loop so we already have all the
-        // _mainAxis and _anchorAxis links set
-        ax._mainSubplot = findMainSubplot(ax, fullLayout);
```
alexcjohnson: moved this into `plots.linkSubplots`, which is both a more obvious place for it and happens earlier (in `supplyDefaults`) so `_mainSubplot` is available sooner.

src/plots/plots.js
```diff
@@ -864,8 +868,57 @@ plots.linkSubplots = function(newFullData, newFullLayout, oldFullData, oldFullLa
...
+
+    // finally, we can find the main subplot for each axis
+    // (on which the ticks & labels are drawn)
+    for(i = 0; i < axList.length; i++) {
+        ax = axList[i];
+        ax._mainSubplot = findMainSubplot(ax, newFullLayout);
+    }
 };
 
+function findMainSubplot(ax, fullLayout) {
```
etpinard: Is moving `findMainSubplot` from `lsInner` to `supplyDefaults` necessary? I suspect this will make `supplyDefaults` for large splom traces significantly slower.

src/plots/cartesian/axes.js
etpinard: Nicely done. It might be nice to try to :hocho: this block

https://github.com/plotly/plotly.js/blob/3ea5c82de6790a3471fea10774fc50b9c3023ae6/src/plots/cartesian/axes.js#L1570-L1584

entirely in favor of better/more-general d3 data functions. I suspect this may lead to performance degradation on pan, so I would need to be careful.

src/plots/plots.js
alexcjohnson: Something or other failed with it in `lsInner`, though I don't recall what. I'll take a look at 🐎 for large sploms, I'm sure there's a shortcut we can take there.

test/jasmine/tests/heatmap_test.js
```diff
@@ -298,6 +298,11 @@ describe('heatmap calc', function() {
 
         fullTrace._extremes = {};
 
+        // clearCalc used to be (oddly enough) part of supplyDefaults.
```
etpinard: Really? How? `On master` I get:

image

src/plots/plots.js
etpinard: > I'm sure there's a shortcut we can take there.

Yeah, I would be nice to leave `findMainSubplot` in `supplyDefaults` and add a `fullLayout._hasOnlyLargeSploms` condition to bypass the loop-over-subplot-for-all-axes loops.

src/plots/cartesian/axes.js
```diff
@@ -2208,26 +2221,39 @@ axes.drawGrid = function(gd, ax, opts) {
     opts = opts || {};
 
     var cls = ax._id + 'grid';
+    var vals = opts.vals;
+    var counterAx = opts.counterAxis;
+    if(ax.showgrid === false) {
+        vals = [];
+    }
+    else if(counterAx && axes.shouldShowZeroLine(gd, ax, counterAx)) {
```
etpinard: Does this fix any bugs, or just a nicer way to filter grid lines on top of zero lines?

test/jasmine/tests/heatmap_test.js
alexcjohnson: oh good catch - it wasn't `clearCalc` itself, but [the `setScale` call I removed](https://github.com/plotly/plotly.js/pull/3323/commits/858f7c2aea73f03f06215c280dd11c3429d11584#diff-ad4f76ccd6044ed16514297078e13b84L507) which contains a [fallback for `_categories` and `_categoriesMap`](https://github.com/plotly/plotly.js/blob/3ea5c82de6790a3471fea10774fc50b9c3023ae6/src/plots/cartesian/set_convert.js#L351-L354) - I'll update the comment to describe that. I still don't think we can remove that fallback... though we're closer now.

etpinard: Got it. Thanks for the explanation!

alexcjohnson: Actually looks like we *can* remove that fallback now 🎉  https://github.com/plotly/plotly.js/pull/3323/commits/da30f160b510c88c4c884d901fb37704df52c7e4

src/plots/cartesian/axes.js
alexcjohnson: hmm, I thought it fixed some obscure bug but I'm not seeing what it is now. But mostly I just wanted to get away from the pattern of "create extras and then delete them," and 🌴 the logic of `axes.shouldShowZeroLine`

etpinard: Ok. No need to dig deeper, I was just curious. I like this pattern much better too!

src/plots/plots.js
alexcjohnson: Large sploms usually avoid the subplot loop there - the [first try](https://github.com/plotly/plotly.js/pull/3323/files/858f7c2aea73f03f06215c280dd11c3429d11584..bdabf48c298001f490368618db0bd4281ae22a89#diff-ad4f76ccd6044ed16514297078e13b84R894) based on `_anchorAxis` works unless the labels are unanchored - upper half with lower labels, or both halves with no diagonal.

Anyway I added an extra collection of counteraxes and subplots *per axis* in https://github.com/plotly/plotly.js/pull/3323/commits/f55e7695209a235ced67c27d39ef4055782b93c4, which means we don't have to search the full subplot list anymore here (and in fact `Axes.getSubplots` is completely unused within plotly.js, but we still have it in streambed ATM). So even when splom *does* go through this loop, it now only takes 0.3ms at 20 dimensions - ~2ms at 50 dimensions if it's O(n^2).

etpinard: > works unless the labels are unanchored - upper half with lower labels, or both halves with no diagonal.

Ha right, I remember making that happen 3 or 4 splom-perf PRs ago. 

> So even when splom _does_ go through this loop, it now only takes 0.3ms at 20 dimensions - ~2ms at 50 dimensions if it's O(n^2).

Beautiful :racehorse:

alexcjohnson:
> note though that in principle both can contribute independently to the margins, if you have a rangeslider (on the bottom) and a top axis with long labels (oh right, I meant to add a test for that case - I'll add that)

Working on this has uncovered some more bugs - will address those in a separate bugfix PR.

etpinard:
:dancer: let's get this in https://github.com/plotly/plotly.js/pull/3300

PR #3325, opened 2018-12-12
Make corrections to Finnish locale translations

lmalmber:
Months and days of the week are written lowercase in Finnish. The list of short days also included a duplicate Sunday in place of Saturday.

Another library I use in the same project with Plotly.js does this correctly, which is why I noticed this:
[https://github.com/charliekassel/vuejs-datepicker/blob/master/src/locale/translations/fi.js](https://github.com/charliekassel/vuejs-datepicker/blob/master/src/locale/translations/fi.js)

Sources:
- Institute for the Languages of Finland: [Months](http://www.kielitoimistonohjepankki.fi/ohje/169), [Days](http://www.kielitoimistonohjepankki.fi/haku/viikonp%C3%A4iv%C3%A4/luokka/929/ohje/190)
- I am a native speaker of Finnish


etpinard:
Thanks! Would you be interested in filling up the `dictionary` translating expression from https://github.com/plotly/plotly.js/blob/master/dist/translation-keys.txt ?

lmalmber:
Hey, I added the dictionary translations. I don't currently know any other Finnish person I could refer to double check them, but I'm fairly satisfied with the translations.

etpinard:
Thanks very much! This will be part of 1.43.0.

Issue #2434, opened 2018-03-02
Mirrored axes have a huge bounding box


nicolaskruchten:
setting `mirror: 'allticks'` on an axis results in a giant bounding box for the axis, which causes problems with the new `automargin` flag on axes, and the positioning of the rangeslider. You can see this problem in action by setting `automargin` on any of the axes here (once we release 1.35.0) https://codepen.io/nicolaskruchten/pen/WMWvVz

This is due to the fact that the mirrored axis is drawn with the same path as the 'main' axis where the ticklabels are. It would make some sense to draw that secondary axis in a different element, so that the bounding box of the main axis is independent of the `mirror` attribute.

alexcjohnson:
Really what we need to do is get rid of the `getBoundingClientRect` calls there altogether. This causes all sorts of problems, see #1988

PR #3326, opened 2018-12-12
Multicategory inside ticks fixes

etpinard:
Follow-up on https://github.com/plotly/plotly.js/pull/3300. Two multicategory fixes discovered when trying to clean up our `ax._boundingBox` mess.

**Edit:** This PR fixes one more bug: https://codepen.io/nicolaskruchten/pen/roOzQO?editors=0010

cc @plotly/plotly_js


etpinard:
A new era of reviews is starting now. Asking @archmoj and @antoinerg to review this thing!

test/image/mocks/multicategory-inside-ticks.json
```diff
@@ -0,0 +1,76 @@
...
+    },
+    "xaxis2": {
+      "title": "MULTI-CATEGORY",
+      "range": [-1, 6.5],
+      "showline": true,
+      "tickfont": {"size": 16},
+      "ticks": "inside",
+      "ticklen": 10,
+      "tickcolor": "red",
+      "tickwidth": 2,
```
archmoj: I noticed `tickwidth` may be considered in the calculation of labels. However using a large number e.g. `tickwidth: 50` could force rotate the labels which might not be desirable.
newplot 5

test/image/mocks/multicategory-inside-ticks.json
```diff
@@ -0,0 +1,76 @@
...
+      "x": [
+        ["2017", "2017", "2017", "2017", "2018", "2018", "2018"],
+        ["q1", "q2", "q3", "q4", "q1", "q2", "q3" ]
+      ],
+      "y": [1, 2, 3, 1, 3, 2, 3, 1],
+      "xaxis": "x2",
+      "yaxis": "y2"
+    }
+  ],
+  "layout": {
```
archmoj: It may be good idea to add `"width": 750` and `"height": 450` layout info.

src/plots/cartesian/axes.js
```diff
@@ -1781,11 +1779,12 @@ axes.drawOne = function(gd, ax, opts) {
     if(ax.type === 'multicategory') {
         var labelLength = 0;
         var pad = {x: 2, y: 10}[axLetter];
```
archmoj: I didn't understand why different padding were needed for x and y directions? It is not related to this PR though.

test/image/mocks/multicategory-inside-ticks.json
etpinard: It _may_ no be desirable, but it is _intended_ for now.

https://github.com/plotly/plotly.js/blob/4f8628f4c7b8f0d604f1f6e7585e962428316cff/src/plots/cartesian/axes.js#L2517

test/image/mocks/multicategory-inside-ticks.json
```diff
@@ -0,0 +1,76 @@
...
+    "grid": {"rows": 2, "columns": 1, "pattern": "independent"},
+    "xaxis": {
+      "side": "top",
+      "title": "MULTI-CATEGORY",
+      "range": [-0.5, 7],
+      "showline": true,
+      "tickfont": {"size": 16},
+      "ticks": "inside",
+      "ticklen": 10,
```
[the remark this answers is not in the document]
etpinard: Ah good eye, I didn't know we allowed negative `ticklen`. We probably shouldn't.

src/plots/cartesian/axes.js
archmoj: True. Most of the texts are horizontal so that's OK to have a stronger divider. But what if the width of image was less than the height?

test/image/mocks/multicategory-inside-ticks.json
```diff
@@ -0,0 +1,76 @@
...
+  ],
+  "layout": {
+    "grid": {"rows": 2, "columns": 1, "pattern": "independent"},
+    "xaxis": {
+      "side": "top",
+      "title": "MULTI-CATEGORY",
+      "range": [-0.5, 7],
+      "showline": true,
+      "tickfont": {"size": 16},
+      "ticks": "inside",
```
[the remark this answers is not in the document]
etpinard: > But when `ticks: null` or `ticks: false`

null, and false aren't valid values:

https://github.com/plotly/plotly.js/blob/4f8628f4c7b8f0d604f1f6e7585e962428316cff/src/plots/cartesian/layout_attributes.js#L307

So, when you _set_ `ticks` to null or false, it defaults to `"outside"`:

https://github.com/plotly/plotly.js/blob/4f8628f4c7b8f0d604f1f6e7585e962428316cff/src/plots/cartesian/tick_mark_defaults.js#L24

test/image/mocks/multicategory-inside-ticks.json
etpinard: Can I asked why? So that it looks better in the test dashboard?

src/plots/cartesian/axes.js
alexcjohnson: I agree, because otherwise the labels run into each other in the reading direction. Which says you might want the larger padding on x too, if both levels of label are rotated.

test/image/mocks/multicategory-inside-ticks.json
```diff
@@ -0,0 +1,76 @@
+{
+  "data": [
+    {
+      "type": "bar",
+      "x": [
+        ["2017", "2017", "2017", "2017", "2018", "2018", "2018"],
+        ["q1", "q2", "q3", "q4", "q1", "q2", "q3" ]
+      ],
+      "y": [1, 2, 3, 1, 3, 2, 3, 1]
```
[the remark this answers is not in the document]
etpinard: Right, that `1` is not _displayed_, but it is still something worth testing: we take the minimum length of "x" and "y" and _length_ for "x" in the multicategory case isn't trivial. 

So effectively, this is just another way to test:

https://github.com/plotly/plotly.js/blob/4f8628f4c7b8f0d604f1f6e7585e962428316cff/src/lib/array.js#L140-L142

test/image/mocks/multicategory-inside-ticks.json
archmoj: When trying to download the image to show the impact of large ticks on label orientation, I noticed that the result was quite different from what was displayed on the screen.

src/plots/cartesian/axes.js
etpinard: this line:

https://github.com/plotly/plotly.js/blob/e1ac60b1078b728277f2ba887cf5ad98b050f77b/src/plots/cartesian/axes.js#L1787

appears to do an ok job:

```js
var l = 'loooooooooong'
var l2 = 'looooooooooo oooooong'

Plotly.newPlot(gd, [{
  type: 'bar',
  x: [
    [l, l, l2, l2],
    [l, l2, l, l2]
  ],
  y: [1, 2, 3, 1]
}], {
  xaxis: {automargin: true},
  width: 300,
  height: 700
})
```

newplot 15

test/image/mocks/multicategory-inside-ticks.json
etpinard: That's expected, label-rotation is computed in pixel-space.

test/image/mocks/multicategory-inside-ticks.json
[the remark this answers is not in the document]
archmoj: Could someone applies a negative `ticklen` to reverse `ticks: 'outside'/'inside'` ?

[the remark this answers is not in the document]
etpinard: Ah

https://github.com/plotly/plotly.js/blob/e1ac60b1078b728277f2ba887cf5ad98b050f77b/src/plots/cartesian/layout_attributes.js#L348-L355

`ticklen` has `min: 0`, so that shouldn't work.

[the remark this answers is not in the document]
archmoj: Thanks for the reference.

archmoj:
Nicely done :dancer:

etpinard:
Nice job reviewing! :tada:

src/plots/cartesian/axes.js
etpinard: Anyway, I'll revisit this problem when fixing https://github.com/plotly/plotly.js/issues/2434 - which I hope I'll get to soon :crossed_fingers:

Issue #2802, opened 2018-07-11
IE11: scattergl: problems with marker symbols


mojoaxel:
We have problems with custom marker-symbols in a `scattergl` chart. 

Chrome and Firefox are OK. 
This seems to be only a Problem in IE (11.0.9600.19002)

```
WEBGL11042: INVALID_OPERATION: useProgram: Program not linked
```

![image](https://user-images.githubusercontent.com/600565/42578521-da718318-8526-11e8-9546-7f84fb757be6.png)

Live Example: https://jsbin.com/zokekob/edit?js,output

etpinard:
cc @dy is this the expected behavior? I remember you saying we only support of symbols in IE11?

etpinard:
related https://github.com/plotly/plotly.js/issues/2750

dy:
Yes, custom symbols are likely broken now in scattergl. That requires a bit of time to fix, I will give a try.

mojoaxel:
@dy Any update on this? For now we switched from `scattergl` to `scatter` on IE. Is there a chance that this will get fixed?

mojoaxel:
I can confirm, that this issue is solved :+1:  
but, as @etpinard mentioned there are still some warnings in the example above:

![image](https://user-images.githubusercontent.com/6675409/43597729-ee9b13a0-9650-11e8-91e6-d9cf1fbf7319.png)

dy:
@etpinard just figured out. That error is actually just a log in console, there is no way to turn it off. That piece of code feature-detects WebGL. I can hardcode browser detection instead, to mute the log for IE11 specifically.

someusersomeuser:
version 1.42.0 and later
marker symbols not work
error: "Path2D" is not defined

PR #3333, opened 2018-12-13
Bump svg-path-sdf

### LINKED ISSUE #3332: scatterpolargl display issue on IE11
Here's an example that works in Chrome, but not in IE11.
https://bl.ocks.org/dmt0/6b47f10114efd50c3c6bba0c1a806045/95a201ad81b846da4dfc3ac2f767287b35ea26cf

From @alexcjohnson on slack:

> [T]here’s a function `isPath2DSupported` but it throws if `Path2D` is undefined (as in IE11) https://github.com/dy/svg-path-sdf/blob/master/index.js#L83-L95 https://github.com/dy/svg-path-sdf/pull/1 (also seems like we should be able to call that function just once and cache the result, right?)

dy:
Fixes #3332
@etpinard @dmt0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3927,7 +3927,8 @@
         "code-point-at": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "concat-map": {
           "version": "0.0.1",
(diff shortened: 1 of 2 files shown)


etpinard:
Nicely done. We still get these:

image

but that's a duplicate of https://github.com/plotly/plotly.js/issues/2802#issuecomment-414420063 - the graph renders fine.

Merging!

PR #3343, opened 2018-12-17
fixes #3336 - don't try to makeCalcdata on visible:false traces

### LINKED ISSUE #3336: Regression with autobins
Found this while upgrading RCE for 1.43.0 but I think it was introduced earlier: https://codepen.io/nicolaskruchten/pen/LMNGNP?editors=0010

etpinard:
... in `Histogram.calc` loop looking for `isFirstVisible` traces.

cc @plotly/plotly_js - a follow-up of https://github.com/plotly/plotly.js/pull/3044

cc @nicolaskruchten 

peek 2018-12-17 14-01


archmoj:
Good fix :dancer: from my corner_|.

PR #3345, opened 2018-12-17
fix #3330 - fix hoverChanged for traces with pointNumbers

### LINKED ISSUE #3330: plotly_hover fires only once on histograms
https://codepen.io/cpsievert/pen/QzjmQq

Note how the 'x' event data is logged only once when hovering over the different bars. This regression appears to have happened somewhere between v1.31 and v1.32 (i.e. it works with v1.31.0 but not v1.32.0)

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3330 cc @plotly/plotly_js

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -1565,8 +1565,11 @@ function hoverChanged(gd, evt, oldhoverdata) {
     for(var i = oldhoverdata.length - 1; i >= 0; i--) {
         var oldPt = oldhoverdata[i];
         var newPt = gd._hoverdata[i];
+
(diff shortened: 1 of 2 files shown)


antoinerg:
Nice fix @etpinard ! 

Is it documented anywhere that histograms return `pointNumbers` instead of `pointNumber`?

etpinard:
> Is it documented anywhere that histograms return `pointNumbers` instead of `pointNumber`?

Not at the moment, tracked in -> https://github.com/plotly/documentation/issues/1210

antoinerg:
Looks good to me :dancer:

Issue #3347, opened 2018-12-18
fixedrange=True is not applied correctly when updating figure through Plotly.react


Marc-Andre-Rivet:
This issue relates to https://github.com/plotly/dash/issues/508.

Setting an empty figure, later overwritten to a figure with data and layout does not exhibit expected behavior for `fixedrange` in axis.

Example that exhibits the incorrect behavior: https://codepen.io/anon/pen/JwRxOo
Minimal example that exhibits the correct behavior: https://codepen.io/anon/pen/xmEBVR

Neither of these examples exhibit this problem when used directly (w/o the react wrapper)

![fixedrange](https://user-images.githubusercontent.com/8092993/50165439-00f57a80-02b3-11e9-88b4-e5254f5a51d1.gif)

PR #3351, opened 2018-12-19
Fix axis autorange on doubleClick

etpinard:
... on graphs w/ `fixedrange:true` axes and w/o set `_rangeInitial`

fixes https://github.com/plotly/plotly.js/issues/3347 cc @plotly/plotly_js

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -703,13 +703,14 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             for(i = 0; i < axList.length; i++) {
                 ax = axList[i];
 
-                if(!ax._rangeInitial) {
-                    attrs[ax._name + '.autorange'] = true;
-                }
-                else {
-                    rangeInitial = ax._rangeInitial;
-                    attrs[ax._name + '.range[0]'] = rangeInitial[0];
-                    attrs[ax._name + '.range[1]'] = rangeInitial[1];
+                if(!ax.fixedrange) {
+                    if(!ax._rangeInitial) {
+                        attrs[ax._name + '.autorange'] = true;
+                    } else {
+                        rangeInitial = ax._rangeInitial;
+                        attrs[ax._name + '.range[0]'] = rangeInitial[0];
+                        attrs[ax._name + '.range[1]'] = rangeInitial[1];
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/plots/cartesian/dragbox.js
```diff
@@ -703,13 +703,14 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
                 ax = axList[i];
 
-                if(!ax._rangeInitial) {
-                    attrs[ax._name + '.autorange'] = true;
-                }
-                else {
-                    rangeInitial = ax._rangeInitial;
-                    attrs[ax._name + '.range[0]'] = rangeInitial[0];
-                    attrs[ax._name + '.range[1]'] = rangeInitial[1];
+                if(!ax.fixedrange) {
```
archmoj: I thought we may need to check for having zero in the first element of range array here.
Here is a [demo](https://codepen.io/MojtabaSamimi/pen/vvyqgJ?editors=0010) which fails using with `yaxis2.range=[0,1]`.

test/jasmine/tests/click_test.js
```diff
@@ -900,6 +901,96 @@ describe('Test click interactions:', function() {
...
+                expect(gd._fullLayout.xaxis._rangeInitial).toBe(undefined);
+                expect(gd._fullLayout.yaxis._rangeInitial).toBe(undefined);
+                expect(gd._fullLayout.yaxis2).toBe(undefined);
+            })
+            .then(function() { return Plotly.react(gd, fig); })
+            .then(function() {
+                _assert('after react into fig', {
+                    xRng: [0, 4],
+                    yRng: [-1, 5],
+                    y2Rng: [-1, 1]
```
archmoj: We may also add the case of [0,1] here.

etpinard: What's special about the `[0,1]` case?

src/plots/cartesian/dragbox.js
archmoj: The demo is actually working with [0,1]. So need to worry there.

test/jasmine/tests/click_test.js
archmoj: The test is perfect the way it is.

archmoj:
A :dancer: from my side.
Thank you @etpinard for the fix.

PR #3355, opened 2018-12-19
Updating to d3-sankey 0.7.1

antoinerg:
This will be the starting point to migrate to https://github.com/tomshanley/d3-sankey-circular which closely ressembles d3-sankey v0.7.1 API.

The biggest difference with the previous version of d3-sankey is a change of coordinate (x, dx) -> (x0, x1).

This is still a WIP although all tests are passing and only one baseline had to be slightly updated.


package.json
```diff
@@ -57,15 +57,18 @@
...
     "3d-view": "^2.0.0",
-    "@plotly/d3-sankey": "^0.5.1",
     "alpha-shape": "^1.0.0",
     "array-range": "^1.0.1",
     "canvas-fit": "^1.5.0",
     "color-normalize": "^1.3.0",
     "convex-hull": "^1.0.3",
     "country-regex": "^1.1.0",
     "d3": "^3.5.12",
+    "d3-sankey": "git://github.com/antoinerg/d3-sankey.git#4f37ed8d3578b545a8569ecd74583f373768e900",
```
antoinerg: The only appreciable change we have with d3-sankey is:
https://github.com/d3/d3-sankey/compare/master...antoinerg:fix-large-padding#diff-968964796da46b424b314b4359ccde31

src/traces/sankey/render.js
```diff
@@ -218,22 +107,50 @@ function linkModel(uniqueKeys, d, l) {
     };
 }
 
-function nodeModel(uniqueKeys, d, n) {
+function linkPath() {
```
antoinerg: This path generator for links used to lived in our fork of d3-sankey and was added in this commit:
https://github.com/plotly/d3-sankey/commit/80c33c31f414c3e263a8bd49e445c281d65ac6b0 . It doesn't need to live in there and could be moved back in here.

This and the fix for large node padding are the only improvements that we had over d3-sankey v0.4.2.

antoinerg: To substantiate this last comment, you can review the [changes of our forked d3-sankey has compared to v0.4.2](https://github.com/d3/d3-sankey/compare/master...plotly:master#diff-968964796da46b424b314b4359ccde31).

src/traces/sankey/convert-to-d3-sankey.js
```diff
@@ -0,0 +1,77 @@
...
+            value: +val
+        });
+    }
+
+    var hasNodeColorArray = isArrayOrTypedArray(nodeSpec.color);
+    var nodes = [];
+    var removedNodes = false;
+    var nodeIndices = {};
+
+    for(i = 0; i < nodeCount; i++) {
```
archmoj: Non-blocking comment: when deleting from a list one may consider reversing the loop count and simply drop (pop) items from the array.

archmoj:
@antoinerg Great work! I was wondering if the status of this PR should be changed to review-able?

antoinerg:
Thanks @archmoj! This is by no means urgent and I feel like it might be a bit too early for a review. There is still the possibility, tat we get rid of d3-sankey altogether if it doesn't fit our requirements. I'll give you an update about this soon.

package.json
antoinerg: Yes, thank you @archmoj! The only changes I made are are in `src/sankey.js`.

etpinard:
@antoinerg Looks like all the commits here are now part of https://github.com/plotly/plotly.js/pull/3406

Ok if we close this one?

PR #3360, opened 2018-12-20
Regression bug fix to draw 3d traces after heatmap  

### LINKED ISSUE #3359: Bug with latest version, opts.dflt is undefined
Unfortunately upgrading to the latest version (v1.43) has not been possible for me. I particularly want it due to the fix of #3306.

I am unsure how to replicate the bug (it may be due to have a large amount of null data) but can show a screenshot of the issue. It happened once I upgraded to 1.43 and going back to 1.42.5 fixed the issue.

screenshot 2018-12-20 10 46 35

archmoj:
Fixes #3359 by not calling `autoType` function on z axis for 3d traces after axis type was to `multicategory` by `heatmap` trace.
@etpinard


src/plots/cartesian/type_defaults.js
```diff
@@ -91,6 +91,7 @@ function setAutoType(ax, data) {
         }
     }
     else {
+        if(axLetter === 'z') opts.noMultiCategory = true;
```
etpinard: That works. But `opts.noMultiCategory` should already be `true` from 

https://github.com/plotly/plotly.js/blob/9beb88d02ac0a0566238e1038159c75408476069/src/plots/cartesian/type_defaults.js#L60

I suspect `d0` in this scope is `fullData[0]` when it should be the first trace on the gl3d subplot.

I hope that makes sense :smile:

archmoj: You are right. It seems to be the first trace on fullData which is not invisible.
The issue is not only affected 3d traces. The example below also fails on master:
<pre>
Plotly.newPlot(gd, {
    "data": [
      {
        "type": "heatmap",
        "x": [0, 1, 2],
        "y": [0, 1, 2],
        "z": [
          [0, 1, 0],
          [1, 0, 1],
          [0, 1, 0]
        ]
      },
      {
        "type": "scattergl",
        "x": [0, 1, 2],
        "y": [0, 1, 2]
      }
    ],
    "layout": {
        "title": "scattergl plot on top of 2d heatmap",
        "width": 600,
        "height": 400
    }
});
</pre>

archmoj: @etpinard The 2d case was mentioned above failed on v.1.43.0; but works on master. So I would fetch master changes and fix this for the 3D case as you mentioned.

src/plots/gl3d/layout/defaults.js
```diff
@@ -97,10 +97,19 @@ function handleGl3dDefaults(sceneLayoutIn, sceneLayoutOut, coerce, opts) {
...
+                opts.fullData[i]
+            );
+        }
+    }
+
     supplyGl3dAxisLayoutDefaults(sceneLayoutIn, sceneLayoutOut, {
         font: opts.font,
         scene: opts.id,
-        data: opts.fullData,
+        data: fullGl3dData,
```
etpinard: Nicely done, but let's :recycle: 

https://github.com/plotly/plotly.js/blob/0078626c6ef3809898cfceab76e8ba227897e046/src/plots/get_data.js#L89-L126

archmoj: @etpinard The try to reuse the getSubPlotData function was not successful.
Could I revert this change?

archmoj: Many thanks @etpinard for the help figuring this out!
Now the tests are all passed.

etpinard:
Nice fix :dancer: !

Issue #3356, opened 2018-12-20
multicategory sorting bugs


alexcjohnson:
Based on a report by @nicolaskruchten 

If you create `multicategory` data that aren't in order, there are different bugs whether the out-of-order values are within one trace or across multiple traces. See https://codepen.io/alexcjohnson/pen/vvgGOB?editors=0010

**Across multiple traces (step 1)**: The combinations in the second trace are tacked on at the end, rather than inserted in the series after the existing values having the same outer category. Then when we're drawing labels, the outer categories are not contiguous, so the labels end up in a weird place. In this example, [4,2] is the 4th item and [4,1] is the 7th (with [5,1] and [6,2] between, from the first trace) so the 4 is drawn halfway between the 5th and 6th items... but that's on a divider between outer categories 5 and 6!
![screen shot 2018-12-19 at 10 07 47 pm](https://user-images.githubusercontent.com/2678795/50261439-8c1d6000-03da-11e9-829c-0d6b4ce43785.png)

**Within one trace (step 2)**: The combinations are correctly sorted, BUT the x data are placed in their unsorted positions.
![screen shot 2018-12-19 at 10 13 37 pm](https://user-images.githubusercontent.com/2678795/50261616-5af15f80-03db-11e9-89df-e1c61c2ac34b.png)

PR #3362, opened 2018-12-21
Multicategory sorting fixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3356 - and made me realised that `multicategory` are slightly more complicated than first expected. 

In brief, we need to (1) scan the data array (2) sort the resulting multi-category list (3) loop over the data arrays again to retrieve their `calcdata` values.

cc @plotly/plotly_js


src/plots/plots.js
```diff
@@ -2617,7 +2617,7 @@ plots.doCalcdata = function(gd, traces) {
     }
 
     // clear stuff that should recomputed in 'regular' loop
-    if(hasCalcTransform) clearAxesCalc(axList);
+    if(hasCalcTransform) setupAxisCategories(axList);
```
alexcjohnson: `, fullData`?

src/plots/plots.js
```diff
@@ -2675,9 +2675,13 @@ plots.doCalcdata = function(gd, traces) {
...
 
-function clearAxesCalc(axList) {
+function setupAxisCategories(axList, fullData) {
     for(var i = 0; i < axList.length; i++) {
-        axList[i].clearCalc();
+        var ax = axList[i];
+        ax.clearCalc();
+        if(ax.type === 'multicategory') {
+            ax.setupMultiCategory(fullData);
+        }
```
alexcjohnson: `clearCalc` is also used in `rangeslider.draw` and `gl2d/scene2d` - do these need `setupMultiCategory`? Should that just be part of `clearCalc` for multicategory axes?

etpinard: > `clearCalc` is also used in `rangeslider.draw` and `gl2d/scene2d` - do these need `setupMultiCategory`?

`setupMultiCategory` only needs to be called once (here in `doCalcdata`, to fill in `ax._muticatList`), so rangeslider don't need it, that's why I didn't put the `setupMultiCategory` logic in `clearCalc`. As for `gl2d/scene2d`, gl2d subplots don't support `multicategory` axes, so I think we're :ok_hand:

src/plots/plots.js
alexcjohnson: maybe a can of worms, but is there a test for multicat + calctransforms (or non-calc-transforms for that matter)?

etpinard: Multicategory axes don't support transforms. I'm not sure how gracefully things fail though.

alexcjohnson: OK, that's fine. Nothing more from me - I'll let one of the other folks give this a once-over and the 💃

etpinard: Transforms + multicategory tracked in -> https://github.com/plotly/plotly.js/issues/3363

test/image/mocks/multicategory-sorting.json
```diff
@@ -0,0 +1,37 @@
...
+        "yaxis": "y2"
+    },
+
+    {
+        "x": [[1, 2, 3, 4, 5, 6, 4, 5, 6, 7, 8, 9],
+              [1, 2, 1, 2, 1, 2, 1, 2, 1, 2, 1, 2]],
+        "y":  [1, 2, 3, 4, 5, 6, 1, 2, 3, 4, 5, 6],
+        "xaxis": "x3",
+        "yaxis": "y3"
+    }],
```
archmoj: @etpinard Would you please adding these two traces to the end of this mock? I was curious about the `null` cases. But your algorithm was able to ignore them. 
<pre>
{
        "x": [[1, 2, 3, 4, 5, 6],
              [1, 2, 1, 2, null, 2]],
        "y":  [1, 2, 3, 4, 5, 6],
        "xaxis": "x4",
        "yaxis": "y4"
    },
    {
        "x": [[4, 5, 6, 7, 8, 9],
              [1, null, 1, 2, 1, 2]],
        "y":  [1, 2, 3, 4, 5, 6],
        "xaxis": "x4",
        "yaxis": "y4"
    }
</pre>

etpinard: Good call! Done in -> 96edd7215

archmoj:
:dancer:

PR #3365, opened 2018-12-21
Surface additional checks for handling empty z arrays and minimum number of rows and columns

### LINKED ISSUE #3364: Surface trace error with empty z data
As also reported in here  https://github.com/plotly/plotly.js/issues/3306#issuecomment-449330809
The user may get errors creating a `surface` trace in scenarios e.g. when the z array is empty.
Here is a [demo](https://codepen.io/MojtabaSamimi/pen/OrpwWj).

archmoj:
Fixes #3364.
The surface trace would be invisible if the z array is defined but is empty or when the columns/row array includes less than two items. 
@etpinard


test/image/mocks/gl3d_surface-with-no-z-data.json
```diff
@@ -0,0 +1,33 @@
+{
+    "data": [
```
etpinard: Could we turn these test cases into jasmine tests instead? This `surface_test.js` block 

https://github.com/plotly/plotly.js/blob/997e720c8f40fda68b98e4c365baf1ac62e84cfd/test/jasmine/tests/surface_test.js#L9

should be a nice place to test the default logic.

src/traces/surface/defaults.js
```diff
@@ -23,14 +23,18 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
     }
 
+    var x = coerce('x');
+    var y = coerce('y');
+
     var z = coerce('z');
-    if(!z) {
+    if(!z || !z.length ||
+       (x ? (x.length < 2) : false) ||
```
etpinard: I would prefer keeping things consistent with `contour` traces here. 1-item contour traces are (like 1-item surfaces) ill-defined, but still appear as `visible: true` in the `fullData` (see -> https://codepen.io/etpinard/pen/oJZyEO)

To avoid drawing errors, we should handle the 1-item case in `surface/convert.js`. Thanks!

etpinard:
Nicely done :dancer:

PR #3369, opened 2018-12-22
Mesh3d cell data checks

### LINKED ISSUE #2630: mesh3d: handle messy data
Currently `mesh3d` does an [in-place modification](https://github.com/plotly/plotly.js/blob/a132b85b899dd490216d12823be56ad3861f091f/src/traces/mesh3d/defaults.js#L47) of the input data during `supplyDefaults`, noticed in the course of https://github.com/plotly/plotly.js/pull/2629. We should not be modifying the user data, rather making a new cleaned array, and we shouldn't be doing any of this in `supplyDefaults` either. Also in other places (like sankey) we reject non-integers rather than truncating, we should do that here too. This probably indicates there is other strange behavior there with messy data as well.

archmoj:
Fix #2630 by removing the indices clean up from the `defaults` code. 
@plotly/plotly_js


src/traces/mesh3d/defaults.js
```diff
@@ -34,20 +34,13 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
 
     if(!coords) {
         traceOut.visible = false;
         return;
     }
 
-    if(indices) {
-        // otherwise, convert all face indices to ints
-        indices.forEach(function(index) {
-            for(var i = 0; i < index.length; ++i) index[i] |= 0;
```
etpinard: Hmm. So you didn't move this to `mesh3d/convert.js`? Just :hocho: ing it didn't break any tests?

Can you try plotting a mesh3d trace with non-integer indices and see what happens?

archmoj: @etpinard Thanks for the review. 
That's true! One could get bad errors in case of having non-integer ids. 
So, in the revised code I added map functions to round the float values to integers. 
However; there is now extra logic to check and handle for such cases. A series of jasmine tests are also added to define the expected behaviour in various scenarios.

src/traces/mesh3d/convert.js
```diff
@@ -71,19 +71,21 @@ proto.update = function(data) {
...
+        toDataCoords(layout.zaxis, data.z, scene.dataScale[2], data.zcalendar)
+    );
 
     var cells;
     if(data.i && data.j && data.k) {
-        cells = zip3(data.i, data.j, data.k);
-    }
-    else if(data.alphahull === 0) {
+        cells = zip3(
+            data.i.map(function(e) { return Math.round(e); }),
```
etpinard: Let's use a for-loop here instead of three `.map`s for speed.

src/traces/mesh3d/defaults.js
```diff
@@ -34,18 +34,58 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
+        return;
+    }
+
+    // test for size of indices
+    if(
+        traceIn.i && Lib.isArrayOrTypedArray(traceIn.i) &&
+        traceIn.j && Lib.isArrayOrTypedArray(traceIn.j) &&
+        traceIn.k && Lib.isArrayOrTypedArray(traceIn.k)
+    ) {
+        if(traceIn.k.length !== 0 && (
```
etpinard: Can we combine this `if(...)` with the one above?

src/traces/mesh3d/defaults.js
```diff
@@ -34,18 +34,58 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
+            traceIn.i.length !== traceIn.j.length ||
+            traceIn.j.length !== traceIn.k.length)) {
+            traceOut.visible = false;
+            return;
+        }
+    }
+
+    var allIndices = readComponents(['i', 'j', 'k']);
```
etpinard: This logic here belongs in `convert.js`. We should never loop over data arrays in `supplyDefaults`.

src/traces/mesh3d/convert.js
archmoj: Good hint. Replaced those as well as other instances of using `map` in `mesh3d/convert.js` https://github.com/plotly/plotly.js/pull/3369/commits/83060bc00a5cc17b1144f486a01e02903b1b9524 .

src/traces/mesh3d/defaults.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3369/commits/a041e51366a6c5b281ec8f7e50026ce2acafbc94

src/traces/mesh3d/convert.js
```diff
@@ -105,6 +134,18 @@ proto.update = function(data) {
...
+
+        if(
+            data.i.length !== data.j.length ||
+            data.j.length !== data.k.length ||
+            !hasValidIndices(data.i, numVertices) ||
+            !hasValidIndices(data.j, numVertices) ||
+            !hasValidIndices(data.k, numVertices) ||
+            !isTriangle([data.i, data.j, data.k])
+        ) {
+            data.visible = false;
```
etpinard: We can't set `visible: false` here. `fullData` keys can't be mutated after the defaults step (unless they are `_` properties) for our `Plotly.react` diffing to work correctly.

Just removing this line here will suffice (i.e. and keeping the early return below) should suffice.

src/traces/mesh3d/convert.js
```diff
@@ -52,7 +53,69 @@ proto.handlePick = function(selection) {
...
+    for(var i = 0; i < len; i++) {
+        if(!isIndex(list[i], numVertices)) {
+            return false;
+        }
+    }
+    return true;
+}
+
+// avoid pointing to an identical vertex twice
+function isTriangle(lists) {
```
etpinard: This criterion looks new. Was this leading to any bugs before?

archmoj: No bugs before. So I simply removed that check.

test/jasmine/tests/mesh3d_test.js
```diff
@@ -3,55 +3,225 @@ var createGraphDiv = require('../assets/create_graph_div');
...
+                x: [0, 1, 0.5, 0.5],
+                y: [0, 0.5, 1, 0.5],
+                z: [0, 0.5, 0.5, 1],
+                i: [0, 0, 0, 1.00001],
+                j: [1, 1, 2, 1.99999],
+                k: [2, 3, 3, 3.00001],
+                type: 'mesh3d'
+            }])
+            .then(function() {
+                assertVisibility(true, 'to be visible');
```
etpinard: Right, this is correct: `fullData` will have `visible:false`. But this doesn't test any of the new logic in convert.js.

Could you add a check for e.g.

```
gd._fullLayout.scene._scene.glplot.objects[0].positions
```

which should be an ok proxy.

archmoj: OK. Will do so.

etpinard: Thanks, sorry for not catching this sooner.

archmoj: No problem. Previously we get different visibility values there...

src/traces/mesh3d/convert.js
```diff
@@ -52,7 +52,53 @@ proto.handlePick = function(selection) {
...
-    return colors.map(str2RgbaArray);
+    var b = [];
+    var len = colors.length;
+    for(var i = 0; i < len; i++) {
+        b[i] = str2RgbaArray(colors[i]);
+    }
+    return b;
+}
+
+// Unpack position data
```
archmoj: @etpinard wondering where could these functions go if we want to reuse these functions in other places e.g. in iso-surfaces?

test/jasmine/tests/mesh3d_test.js
```diff
@@ -3,55 +3,298 @@ var createGraphDiv = require('../assets/create_graph_div');
...
-        })
-        .then(function() {
-            _assert([true, 0, 10], [true, 0, 3]);
-
-            return Plotly.purge(gd);
-        })
-        .catch(failTest)
-        .then(done);
+        function assertPositions(exp, msg) {
+            expect(gd._fullLayout.scene._scene.glplot.objects[0].positions.length !== undefined).toBe(exp, msg);
```
etpinard: Can we test the actual `length` values, not just whether `positions` and `cells` are arrays like you did in your latest commit?

archmoj: Good call. I'll fix that.

src/traces/mesh3d/convert.js
etpinard: I'd vote for putting helper functions like this one in a new file `src/traces/mesh3d/helpers.js`.

src/traces/mesh3d/defaults.js
```diff
@@ -34,19 +34,21 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
     }
 
-    if(indices) {
-        // otherwise, convert all face indices to ints
-        indices.forEach(function(index) {
-            for(var i = 0; i < index.length; ++i) index[i] |= 0;
-        });
+    // three indices should be all provided or not
+    if(
+        (traceIn.i && (!traceIn.j || !traceIn.k)) ||
```
etpinard: Ha, you're using `traceIn` here. So something like

```
{
  i: 1,
  j: true,
  k: {}
}
```

would have `visible:true`.

Could we be more stringent and use `traceOut` instead, to ensure that i,j,k are all non-empty arrays.

archmoj: Good Call. Now `traceOut` is applied.
Noting that `visible` would "only" be `false` in case of bad 'vertices'.

etpinard: Hmm. I don't understand `traceOut.i`, `traceOut.j` and `traceOut.k` should be defined at this stage. They get filled in during the `coerce()` calls in `readComponents` below.

archmoj: You are right. That is fixed in https://github.com/plotly/plotly.js/pull/3369/commits/29f5a06c2aa5be4101c33ed649f8ef74a19a0d71
Now the traces become invisible in case of bad index arrays.

etpinard:
All right, good enough for now :dancer: 

Thanks!

Issue #3235, opened 2018-11-09
Support div id in Plotly.makeTemplate


alexcjohnson:
ie if you can do `Plotly.newPlot('myDiv', figure)` you should also be able to do `Plotly.makeTemplate('myDiv')`

(cc @nicolaskruchten)

PR #3375, opened 2018-12-26
Add support for div id as 1st arg to Plotly.makeTemplate

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3235 cc @plotly/plotly_js


archmoj:
Looks good to me.
:dancer:

PR #3372, opened 2018-12-24
Add postshrinkwrap step to replace http:// -> https://

etpinard:
More info: 

- https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/44
- https://www.npmjs.com/package/replace (using this as a cross-platform `sed`)

so that 

```
git checkout master
npm i
```

doesn't lead to

image

ever again!

cc @plotly/plotly_js

PR #3382, opened 2019-01-01
Compact http to https with zero dependency

archmoj:
This PR is the follow up to #3372. 
In order to reduce the chance of undesirable changes to the code by using a general purpose `replace` program, regex and its dependencies, a new module is developed https://www.npmjs.com/package/chttps that could be maintained by the Plotly team and may possibly be handy in fixing similar `http > https` fixes in other places namely in `gl-vis` module upgrades.
@plotly/plotly_js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1591,9 +1591,9 @@
       }
     },
     "chttps": {
-      "version": "1.0.4",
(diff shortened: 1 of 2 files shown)


etpinard:
Nice find. Thanks :dancer:

Issue #3378, opened 2018-12-28
uirevision + 3D view changes


alexcjohnson:
Something's broken with `uirevision` and 3D scene view changes:
```js
Plotly.newPlot(gd,[{type:'surface',z:[[1,2,3],[3,1,2],[2,3,1]]}],{uirevision:true})
// then rotate & zoom, before redrawing:
Plotly.react(gd,[{type:'surface',z:[[1,2,3],[3,1,2],[2,3,1]]}],{uirevision:true})
> WARN: unrecognized GUI edit: up.x
> WARN: unrecognized GUI edit: up.y
> WARN: unrecognized GUI edit: up.z
> WARN: unrecognized GUI edit: center.x
> WARN: unrecognized GUI edit: center.y
> WARN: unrecognized GUI edit: center.z
> WARN: unrecognized GUI edit: eye.x
> WARN: unrecognized GUI edit: eye.y
> WARN: unrecognized GUI edit: eye.z
// and the view jumps back to the original
```
Looks like these get recorded in `_fullLayout._preGUI` without prepending the scene name.

alexcjohnson:
There seem to be problems with `geo` and `mapbox` as well... with different details but the same result: GUI viewport changes are not persisted by `uirevision`.

PR #3394, opened 2019-01-03
React + uirevision fixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3378 - that is `uirevision` is now functional with subplots that do not call `_guilRelayout` at interactions' end (i.e. gl3d, geo and mapbox).

These subplots have to use `_storeDirectGUIEdit` to fill `gd._fullLayout._preGUI` and get `uirevision` to work correctly. On master, `_preGUI` was filled with the "new" values instead of the "old" (i.e. "pre-interaction") values that the restyle/relayout/update/react logic expects. In brief, fixing this bug was mostly just a matter of reordering things.

cc @plotly/plotly_js @alexcjohnson

### DIFF
--- a/src/plots/mapbox/mapbox.js
+++ b/src/plots/mapbox/mapbox.js
@@ -32,9 +32,6 @@ function Mapbox(opts) {
     // unique id for this Mapbox instance
     this.uid = fullLayout._uid + '-' + this.id;
 
-    // full mapbox options (N.B. needs to be updated on every updates)
-    this.opts = fullLayout[this.id];
-
     // create framework on instantiation for a smoother first plot call
     this.div = null;
     this.xaxis = null;
@@ -57,9 +54,7 @@ module.exports = function createMapbox(opts) {
 
 proto.plot = function(calcData, fullLayout, promises) {
     var self = this;
-
-    // feed in new mapbox options
-    var opts = self.opts = fullLayout[this.id];
+    var opts = fullLayout[self.id];
 
     // remove map and create a new map if access token has change
     if(self.map && (opts.accesstoken !== self.accessToken)) {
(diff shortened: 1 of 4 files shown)


test/jasmine/tests/plot_api_react_test.js
```diff
@@ -1806,3 +1807,67 @@ describe('Plotly.react and uirevision attributes', function() {
...
+                z: [[1, 2, 3], [3, 1, 2], [2, 3, 1]]
+            }], {
+                width: 500,
+                height: 500,
+                uirevision: true
+            });
+        }
+
+        // mocking panning/scrolling is brittle,
+        // this here is enough to to trigger the relayoutCallback
```
alexcjohnson: but this presumably doesn't actually change `camera.eye`, right? Would it be possible to do something that changes the `eye`, even if it's not robust what it changes it *to*? Just "not the default", and verify that after another `_react` it's again "not the default"? Or is it so brittle we can't even ensure it will register anything at all?

src/plots/mapbox/mapbox.js
```diff
@@ -32,9 +32,6 @@ function Mapbox(opts) {
     // unique id for this Mapbox instance
     this.uid = fullLayout._uid + '-' + this.id;
 
-    // full mapbox options (N.B. needs to be updated on every updates)
-    this.opts = fullLayout[this.id];
```
alexcjohnson: Very nice simplification!

test/jasmine/tests/plot_api_react_test.js
etpinard: > but this presumably doesn't actually change `camera.eye`, right?

Yes, you were correct. 

> Would it be possible to do something that changes the `eye`, even if it's not robust what it changes it _to_? Just "not the default", and verify that after another `_react` it's again "not the default"?

Good call. Here's my attempt -> https://github.com/plotly/plotly.js/pull/3394/commits/b880717ba8b86602a62d8010a8028695556c0103

alexcjohnson: Ah clever! I'll take it :)

alexcjohnson:
Beautiful, thanks for cleaning all this up! 💃

PR #3396, opened 2019-01-04
better default order for multiple hovers

### LINKED ISSUE #3050: Misalignment of horizontal box/violin hover labels
Split out of @alexcjohnson 's  https://github.com/plotly/plotly.js/issues/2970#issuecomment-424499773

----------

The misalignment in the original multi-plot codepen https://codepen.io/kotwgarnku/pen/gdxaPN when you set the first plot to `closest` is still there (and yes, the codepen is now using 1.41.3). To be clear, I'm talking about how eg the "min" label is not pointing to the "min" value, but to where that value would be according to the bottom plot. There also seems to be a problem with the overlap avoidance algorithm when one of the labels is too long and gets flipped to the top - it's putting them out of order and then stretching them across each other.
screen shot 2018-09-25 at 4 52 23 pm

### LINKED ISSUE #3389: Horizontal box hover labels out of order
This is pretty rough... median to the left of min and max to the left of upper:

image

Can we enforce an order here?

nicolaskruchten:
Closes #3389 AND closes #3050 :)

### DIFF
--- a/src/traces/box/hover.js
+++ b/src/traces/box/hover.js
@@ -145,14 +145,25 @@ function hoverOnBoxes(pointData, xval, yval, hovermode) {
 
     // box plots: each "point" gets many labels
     var usedVals = {};
-    var attrs = ['med', 'min', 'q1', 'q3', 'max'];
+    var attrs = ['min'];
+
+    if(trace.boxpoints || trace.points) {
+        attrs.push('lf');
+    }
+
+    attrs.push('q1', 'med');
 
     if(trace.boxmean || (trace.meanline || {}).visible) {
         attrs.push('mean');
     }
+
+    attrs.push('q3');
+
     if(trace.boxpoints || trace.points) {
-        attrs.push('lf', 'uf');
+        attrs.push('uf');
     }
+
+    attrs.push('max');
 
     for(var i = 0; i < attrs.length; i++) {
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Note that this makes #2970 a bit worse, because it'll show only 'min' and not 'median' any longer, because that's now the first hover.

nicolaskruchten:
LMK if you like the fix and I'll fix the tests

etpinard:
Thanks for the PR!

Going from:

peek 2019-01-04 15-30

to:

peek 2019-01-04 15-31

is amazing :tada: 

I'm a little confused about _why_ your patch works though. 

`Fx.hover` doesn't seem to care about the ordering of the labels of vertical box/violin traces and  of horizontal box/violin traces under `hovermode:'y'`. I'll try to dig a little deeper. Maybe `Fx.hover` is just missing a `.sort()` for horizontal labels.

nicolaskruchten:
Hehehe like I said, I just looked at it and thought "seems like they're laying out in order and jamming each other up" ... I didn't dig into the root cause at all, but if there's a deeper/more elegant fix I'm all for it.

etpinard:
My attempt in -> https://github.com/plotly/plotly.js/pull/3401

Issue #3388, opened 2019-01-03
the toolbar's background is always visible, not just on hover


chriddyp:
It looks like the toolbar's background is always visible, not just when you hover. In the image below, see how the toolbar's background is visible.
![image](https://user-images.githubusercontent.com/1280389/50647233-0ce56b80-0f46-11e9-999c-64a3af20c335.png)

I'm on Firefox 64.0 and plotly.js 1.43.1

etpinard:
In a codepen: https://codepen.io/etpinard/pen/gZepKm?editors=1010

etpinard:
The culprit is https://github.com/plotly/plotly.js/pull/3280

PR #3397, opened 2019-01-04
Add :hover styleRule for modebar.bgcolor

etpinard:
... and a `.ease-bg` class to match modebar button opacity transition.

fixes https://github.com/plotly/plotly.js/issues/3388  - for example:

```js
Plotly.newPlot('graph', [{
  y: [1, 2, 1]
}], {
  width: 500,
  paper_bgcolor: '#d3d3d3',
  margin: {t: 0},
  modebar: {bgcolor: 'red'}
}, {
  watermark: true
})
```
peek 2019-01-04 12-58

cc @plotly/plotly_js - I would appreciate if modebar-expert @antoinerg would test this one out. Thanks!

### DIFF
--- a/src/components/modebar/modebar.js
+++ b/src/components/modebar/modebar.js
@@ -52,21 +52,22 @@ proto.update = function(graphInfo, buttons) {
     this.element.setAttribute('id', modeBarId);
     this._uid = modeBarId;
 
-    if(context.displayModeBar === 'hover') {
-        this.element.className = 'modebar modebar--hover';
-    }
-    else this.element.className = 'modebar';
+    this.element.className = 'modebar';
+    if(context.displayModeBar === 'hover') this.element.className += ' modebar--hover ease-bg';
 
     if(fullLayout.modebar.orientation === 'v') {
         this.element.className += ' vertical';
         buttons = buttons.reverse();
     }
 
+    var style = fullLayout.modebar;
+    var bgSelector = context.displayModeBar === 'hover' ? '.js-plotly-plot .plotly:hover ' : '';
+
     Lib.deleteRelatedStyleRule(modeBarId);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId, 'background-color: ' + fullLayout.modebar.bgcolor);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn .icon path', 'fill: ' + fullLayout.modebar.color);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn:hover .icon path', 'fill: ' + fullLayout.modebar.activecolor);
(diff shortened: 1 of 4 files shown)


etpinard:
Ping @antoinerg - could you take a look at this today. I'll :heart:  to include this in 1.43.2. Thanks!

src/components/modebar/modebar.js
```diff
@@ -52,21 +52,22 @@ proto.update = function(graphInfo, buttons) {
...
 
+    var style = fullLayout.modebar;
+    var bgSelector = context.displayModeBar === 'hover' ? '.js-plotly-plot .plotly:hover ' : '';
+
     Lib.deleteRelatedStyleRule(modeBarId);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId, 'background-color: ' + fullLayout.modebar.bgcolor);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn .icon path', 'fill: ' + fullLayout.modebar.color);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn:hover .icon path', 'fill: ' + fullLayout.modebar.activecolor);
-    Lib.addRelatedStyleRule(modeBarId, '#' + modeBarId + ' .modebar-btn.active .icon path', 'fill: ' + fullLayout.modebar.activecolor);
+    Lib.addRelatedStyleRule(modeBarId, bgSelector + '#' + modeBarId, 'background-color: ' + style.bgcolor);
```
antoinerg: Would this work without `bgSelector`? `modeBarId` should already be unique so this might not be necessary.

etpinard: Maybe there's a way (not a CSS expert here), but I think that `:hover` pseudo-class must be tagged onto the graph div class.

test/jasmine/tests/modebar_test.js
```diff
@@ -1336,16 +1336,55 @@ describe('ModeBar', function() {
...
+                style = window.getComputedStyle(gd._fullLayout._modeBar.element);
+                expect(style.backgroundColor).toBe('rgba(0, 0, 0, 0)');
+                expect(getStyleRule().rules[3].style.backgroundColor).toBe(colors[1]);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('changes background color (displayModeBar: true)', function(done) {
+            function getStyleRule() {
```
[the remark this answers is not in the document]
etpinard: Good call. Done in -> a4fb8cd37

src/components/modebar/modebar.js
antoinerg: Ok thanks for the explanation. This makes sense!

antoinerg:
Thanks for the fix and the nice tests @etpinard :dancer:

PR #3398, opened 2019-01-04
adding hovertemplates to polar bar and scatter and ternary scatter

nicolaskruchten:
I mostly just followed the pattern in https://github.com/plotly/plotly.js/pull/3126 :)


etpinard:
Nice. Would you be interested in adding tests to help us :lock: down this feature?

src/traces/barpolar/attributes.js
```diff
@@ -69,6 +71,9 @@ module.exports = {
     marker: barAttrs.marker,
 
     hoverinfo: scatterPolarAttrs.hoverinfo,
+    hovertemplate: hovertemplateAttrs({}, {
```
etpinard:
```
hovertemplate: hovertemplateAttrs(),
```

is probably good enough for all the trace types patched in this PR. That extra `keys` list is meant for traces with event data keys that do not match coordinates (e.g. histogram traces). No need to add new `constants.js` files as well.

src/traces/scatterpolargl/defaults.js
```diff
@@ -35,6 +35,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     coerce('mode', len < PTS_LINESONLY ? 'lines+markers' : 'lines');
     coerce('text');
     coerce('hovertext');
+    coerce('hovertemplate');
```
etpinard: **N.B.** `hovertemplate` should not get coerced for `hoveron: 'fills'`, just like for `scatter` traces:

https://github.com/plotly/plotly.js/blob/a78600a5689696b63e051454a6976c242698bc18/src/traces/scatter/defaults.js#L78

src/traces/scatterternary/defaults.js
```diff
@@ -61,6 +61,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
 
     coerce('text');
     coerce('hovertext');
+    coerce('hovertemplate');
```
etpinard: .... similarly here, as `scatterternary` also support an `hoveron: 'fills'` mode.

nicolaskruchten:
Any recommendation about where to add these tests? Is there a pattern somewhere for testing hovers?

nicolaskruchten:
OK, I added tests for `scatterpolar` and `scatterpolargl` because the test harness was set up for it. I tried to replicate the test harness for `barpolar` and `scatterternary` but I wasn't able to trick `assertHoverLabelContent` into finding anything :(

etpinard:
> I tried to replicate the test harness for `barpolar` and `scatterternary` but I wasn't able to trick

Right, because `barpolar` and `scatterternary` hover test don't actually test `Fx.hover` (where the template substitutions happen). They instead test their trace module's `.hoverPoints` method - which _only_ feeds in `pointData.hovertemplate` to `Fx.hover`.

So, an _ok_ test for both `barpolar` and `scatterternary` would be to assert that `pointData.hovertemplate` is defined after `hoverPoints()`.

nicolaskruchten:
OK, done :)

etpinard:
Nice PR! :dancer: - thanks for your help @nicolaskruchten !

nicolaskruchten:
Happy to help! I might submit a followup for `scattergeo` and `scattermapbox` if they're also using the easy case.

etpinard:
> I might submit a followup for `scattergeo` and `scattermapbox`

Yeah, those should be straight-forward and similar to this PR.

Trace types `scatter3d`, `surface`, `mesh3d` and `parcats` might be more of a pain (but doable, their implementation would be similar to `pie` in https://github.com/plotly/plotly.js/pull/3126

VeraZab:
so nice to see it automagically work in RCE :)

PR #3399, opened 2019-01-04
Expose mapbox minzoom, maxzoom, and symbol-placement properties

rockg:
This exposes the mapbox minzoom, maxzoom, and symbol-placement properties. Currently this only exposes them to the `mapbox` argument and not to `Scattermapbox`. 

Some items:
1. It wasn't clear to me where tests in `mapbox_test.js` this should as I don't see other for strict mapbox arguments (most of the tests are for Scattermapbox). 
2. I followed the instructions for development on the Contributing page and step 4 tripped me up. The page opens, but there's nothing there despite the "mocks search" and the description suggests there should be tests in the dashboard.

@etpinard


etpinard:
Thanks very much for the PR!

I won't have the time to look at this in detail before we release 1.44.0 (in ~ 1 week from now). This does seem like a nice addition to the library!

etpinard:
@plotly/plotly_js how do you feel about the proposed attributes:

- `layout.mapbox?.layers[i].(min|max)zoom`
- `layout.mapbox?.layers[i].symbol.placement`

Should we attempt to stay 1-to-1 with the mapbox-gl API or should we try to stay consistent with similar existing plotly.js attributes? I'm thinking

- a more plotly.js-esque version of `(min|max)zoom` would be: `zoomrange` or just `range`
- similarly `placement` -> `position`

Any opinions on this topic?

alexcjohnson:
I'd vote to leave them as they are. `(min|max)zoom` is doing something very different from our other range-type attributes, and I imagine it's very common to only want to set one of them. `placement` is somewhat analogous to `textposition`, but seems to me it's unique enough that it will help people figure it out if its name matches where they should look in the mapbox docs.

rockg:
Another thing that I can add here is some documentation to map the plot attributes to mapbox attributes. I didn't see this anywhere and it is not obvious without looking at the code the correspondence between the two. I can add this in the reference section for each attribute.

etpinard:
> Another thing that I can add here is some documentation to map the plot attributes to mapbox attributes

Sure, go for it!

----

Moreover, could send us (better yet, check it into to [`test/image/mocks/`](https://github.com/plotly/plotly.js/tree/master/test/image/mocks)) a "data" / "layout" JSON file that showcases these new attributes? Thanks!

etpinard:
We'll be making a minor release (`v1.45.0`) in the next few days - and it would be nice to include this PR in it.

@rockg Are you still planning on pushing another commit to this branch? If not, no worries, we'll gladly bring this PR over the finish line.

rockg:
I should have time to do it by Wednesday and will let you know if I won't. Sorry for the delay.

src/plots/mapbox/layout_attributes.js
```diff
@@ -217,7 +223,15 @@ var attrs = module.exports = overrideAll({
                 dflt: 2,
                 role: 'style',
                 description: [
-                    'Sets the line width.',
+                    'Sets the line width (mapbox.layer.paint.line-width).',
+                    'Has an effect only when `type` is set to *line*.'
+                ].join(' ')
+            },
+            dash: {
+                valType: 'data_array',
```
etpinard: Does `mapbox.layer.paint.line-dasharray` also support scalars?

rockg: "Optional array of numbers greater than or equal to 0" [here](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layers-line)

etpinard: Thanks!

etpinard:
Thanks @rockg 

I added the baseline to your new mock

image


Looking good!

rockg:
@etpinard Thanks. Unfortunately the line label and min/max zoom aren't easily tested by the baseline as those are somewhat interactive.

etpinard:
Yeah, that's ok I guess. plotly.js is just passing things to mapbox-gl here. I'm sure mapbox has tests for those things.

Merging! Thank you very much for your contribution!

PR #3400, opened 2019-01-04
Errorbars inherit color from line or marker color

### LINKED ISSUE #3392: Error bars not inheriting color

image

archmoj:
fix #3392 for 5 traces.

Note 1: For `bar` and `histogram` cases `marker.line.color` is applied as the default color. If undefined then `marker.color' is used if defined. 

Note 2: For `scatter`, `scatter3d` and `scattergl` cases `line.color` is applied as the default color. If undefined then `marker.color' is used if defined. Also note that `marker.line.color` is not used in this regard; and the trace line color and errorbar lines would use `colorway` in that scenario.

@plotly/plotly_js


archmoj:
now at #3408

PR #3401, opened 2019-01-04
Fix horizontal box/violin hover label misalignment under hovermode:'closest'

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3050 (although it might worth moving https://github.com/plotly/plotly.js/issues/3050#issuecomment-424508921 into a new issue).

The idea behind this patch came from @nicolaskruchten 's https://github.com/plotly/plotly.js/pull/3396 - which revealed that something was wrong with how box/violin horizontal hover labels were sorted before the avoid-overlap algo. This PR is essentially a continuation of https://github.com/plotly/plotly.js/pull/3043

I have to add some tolerance in the kde hover label text assertion, but other than that this PR is ready to review cc @plotly/plotly_js

### DIFF
--- a/test/jasmine/tests/box_test.js
+++ b/test/jasmine/tests/box_test.js
@@ -312,6 +312,29 @@ describe('Test box hover:', function() {
         },
         nums: 'look:0.7',
         name: ''
+    }, {
+        desc: 'orientation:h | hovermode;y',
+        mock: require('@mocks/box_grouped_horz.json'),
+        pos: [430, 130],
+        nums: [
+            'max: 1', 'mean ± σ: 0.6833333 ± 0.2409472', 'min: 0.3',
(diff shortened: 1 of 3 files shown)


etpinard:
Note that this PR doesn't make https://github.com/plotly/plotly.js/issues/2970 "worse" - unlike https://github.com/plotly/plotly.js/pull/3396#issuecomment-451473288

test/jasmine/tests/box_test.js
```diff
@@ -312,6 +312,29 @@ describe('Test box hover:', function() {
         },
         nums: 'look:0.7',
         name: ''
+    }, {
+        desc: 'orientation:h | hovermode;y',
```
nicolaskruchten: Colon or semicolon?

nicolaskruchten:
:dancer: if the colon/semicolon is a display-only thing, which it appears to be.

PR #3404, opened 2019-01-06
Cartesian small domain

### LINKED ISSUE #3403: small domain does not display correctly.
In this [codepen][1], the first attempt on `Plotly.newPlot` is not working.

I have two Y-axes. The bottom has a domain of `[0,0.5]`, while the second has the `plot_layout.yaxis2.domain` = `[0.945, 0.95]`. However, the second Y-axis will overlap the whole plot. I expected it only covers the top 94.5% to 95%. I don’t know why this does not work. If I simply change the domain to `[0.94, 0.95]` covering top 94% to 95%, it suddenly worked. Why does this happen?

  [1]: https://codepen.io/anon/pen/wRmQLN

archmoj:
Fix #3403


src/plots/cartesian/position_defaults.js
```diff
@@ -72,7 +72,7 @@ module.exports = function handlePositionDefaults(containerIn, containerOut, coer
         // in the axes popover to hide domain for the overlaying axis.
         // perhaps I should make a private version _domain that all axes get???
         var domain = coerce('domain', dfltDomain);
-        if(domain[0] > domain[1] - 0.001) containerOut.domain = dfltDomain;
+        if(domain[0] > domain[1] - (1 / 32768.0)) containerOut.domain = dfltDomain;
```
etpinard: Where does that `32768` number come from?

etpinard: (No need for the trailing `.0` in JS by the way)

archmoj: 32768 = 2 ^ 15 is ofcourse the maximum signed short seems to be the maximum width/height on Chrome.
After more research I noticed [this npm module](https://www.npmjs.com/package/canvas-size) updated today with good info! According to that we may reduce that to something like 16,384 for FF. Older versions of IE have lower limits.

etpinard: Interesting. Thanks for the info!

I don't adding a dependency for this is worth it. I suggest finding the max of min values listed. 

Inspecting

https://github.com/jhildenbiddle/canvas-size/blob/52382c95de0b66c447d032f3d8455c6aaaec1fc6/src/index.js#L55-L72

looks like `1 / 4096` will be good enough.

archmoj: Done in https://github.com/plotly/plotly.js/pull/3404/commits/a7c4485f78963c99fc8354b6f29493ee98aa9767

etpinard: Thanks! Maybe you could put a comment above this line explaining where that `4096` comes from.

etpinard:
Ok. Nicely done :dancer:

PR #3406, opened 2019-01-07
d3-sankey-circular

### LINKED ISSUE #2636: Feature request: Sankey - support circular references
Circular references for Sankey graphs are not currently supported:
`ERROR: Circularity is present in the Sankey data. Removing all nodes and links.`

This would be a fantastically useful optional ability for Plotly.js.

I have found an example that uses d3.js here:
http://bl.ocks.org/tomshanley/11392275
...but it is not as developer-friendly as the Plotly.js library.

antoinerg:
Initial integration of `d3-sankey-circular` to close #2636 

This branch is based off PR https://github.com/plotly/plotly.js/pull/3355 which brings `d3-sankey` to version 0.7.1. 

When detecting circular links, we swap from `d3-sankey`(https://github.com/d3/d3-sankey) to `d3-sankey-circular` (https://github.com/tomshanley/d3-sankey-circular). This is possible because `d3-sankey-circular` has a similar API to `d3-sankey` 0.7.x and this is why PR https://github.com/plotly/plotly.js/pull/3355 was an important stepping stone.

Here are example baselines:

To do:
- [x] Add dragging and a force model
- [x] Links are not rectangles anymore but path (so we don't have a fill property, but only stroke)
- [x] Style link on hover


package.json
```diff
@@ -58,15 +58,19 @@
...
     "3d-view": "^2.0.0",
-    "@plotly/d3-sankey": "^0.5.1",
     "alpha-shape": "^1.0.0",
     "array-range": "^1.0.1",
     "canvas-fit": "^1.5.0",
     "color-normalize": "^1.3.0",
     "convex-hull": "^1.0.3",
     "country-regex": "^1.1.0",
     "d3": "^3.5.12",
+    "d3-sankey": "git://github.com/antoinerg/d3-sankey.git#4f37ed8d3578b545a8569ecd74583f373768e900",
```
antoinerg: Minimal changes over latest d3-sankey https://github.com/d3/d3-sankey/compare/master...antoinerg:fix-large-padding

antoinerg:
Here's a Codepen to generate random mocks with circular links: https://codepen.io/anon/pen/LMBZbK

You can play around with `N`, `L` and `targetOffset` in the source code.

AryehGielchinsky:
Hi antoinerg, thank you for your work bringing circular Sankey to plotly. I'm beginner/intermediate at python, beginner at plotly, and don't know any JavaScript. Is there any way I could help?

antoinerg:
Hello @AryehGielchinsky and thank you for your interest in plotly.js. Sankey is undergoing majors changes as we're trying several new features and getting feedback from the community is useful.

What's your use case for circular Sankey? Do you have interesting datasets with circularity that we could use as examples?

antoinerg:
With 3ed9ba5, circular links type is updated after moving nodes (`top` <-> `bottom`). In the following GIF, we see how link 3 to 2 goes from `top` to `bottom` upon moving towards the bottom of the plot, and link 5 to 0 goes from `bottom` to `top`.

sankey_link_type_change

antoinerg:
As of right now, circular Sankey does not take into account `link.line.width` and `link.line.color`. The reason for this is that a link is a simple svg path that is NOT closed. Therefore their appearance can only be controlled with `stroke-width` and the color of the `stroke` itself.

We don't have any mocks showcasing attributes `link.line.(width|color)` but IMHO I don't see much value for them since any non-zero border `link.line.width !== 0` just looks wrong and out of place to me:
sankey_link_lines

Translating the current open path to closed ones to support `link.line.(width|color)` would involved modifying function `createCircularPathString` defined in:
https://github.com/antoinerg/d3-sankey-circular/blob/40acb8385fd5dbda5272634583be5b40e8cf69b1/src/sankeyCircular.js#L983-L1158

@etpinard Should we support those 2 attributes in circular Sankey?

etpinard:
> Should we support those 2 attributes in circular Sankey?

Eventually it would be nice, but no need to do this now. Just make sure to mention this limitation in the `link.line.{color,width}` attribute `description`.

AryehGielchinsky:
@antoinerg I work for a financial/social media website. I'd like to use Sankey to visualize user traffic through different areas of our site. Often users will view several pages of the same type, which would require a circular Sankey. If that sounds like a good example to you, I could check if we'd be willing to release that data publicly.

antoinerg:
@AryehGielchinsky Sounds like an interesting use case for Sankey! Depending on your data, the layout (that is the position of the nodes/links) may or may not be optimal. I'd like to get a better sense of how well the current algorithm performs for various datasets. Keep me posted :)

antoinerg:
> Eventually it would be nice, but no need to do this now. Just make sure to mention this limitation in the `link.line.{color,width}` attribute `description`.

Done in 6fb3fb4e68649398fa3c91c32cc1f337f35d457b !

package.json
```diff
@@ -58,15 +58,19 @@
...
+    "d3-sankey-circular": "git://github.com/antoinerg/d3-sankey-circular.git#e5760737ef64afade7906bd018149bf1da080e2d",
```
antoinerg: Minimal changes over latest d3-sankey-circular:

 https://github.com/tomshanley/d3-sankey-circular/compare/master...antoinerg:support-update

antoinerg:
@etpinard I should probably add tests to make sure `sankey.update()` works properly for every (as of right now there's only 2) Sankey generators.

package.json
etpinard: Hmm. So we no longer need https://github.com/plotly/d3-sankey ? I thought [i read](https://github.com/plotly/plotly.js/issues/2636#issuecomment-389330558) that we made changes in our `@plotly/d3-sankey` that got rejected by `d3/d3-sankey`.

package.json
etpinard: Cool. We'll have to make a new (or move it?) to the `@plotly` organization.

package.json
```diff
@@ -58,15 +58,19 @@
...
+    "d3-array": "1",
```
etpinard: Looks like we don't need to list `d3-array`, `d3-collection` and `d3-interpolate` in our deps. This here looks like an artefact from https://github.com/plotly/plotly.js/pull/3355/commits/03a3cb578bc33a06e0c27ed1d91bb79c32e673f4

package.json
etpinard: ... unless you think your patches are general enough to hope that a PR would get merged over at https://github.com/tomshanley/d3-sankey-circular

src/traces/sankey/convert-to-d3-sankey.js
```diff
@@ -0,0 +1,80 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Lib = require('../../lib');
+var isArrayOrTypedArray = Lib.isArrayOrTypedArray;
+var isIndex = Lib.isIndex;
+
+module.exports = function(trace) {
```
etpinard: I'd vote for bringing this over to `sankey/calc.js` and remove this file. I don't see the need to have two <100 lines of code file side-by-side.

etpinard: ... there's a good reason for it that I can't think of

etpinard: Regardless of where this routine ends up, thanks very much for making sankey a little more _plotly.js-esque_ :tada:

src/traces/sankey/render.js
```diff
@@ -220,23 +124,55 @@ function linkModel(uniqueKeys, d, l) {
...
-    uniqueKeys[basicKey] = (foundKey || 0) + 1;
-    var key = basicKey + '__' + uniqueKeys[basicKey];
+    var key = basicKey + '__' + i;
 
     // for event data
     n.trace = d.trace;
     n.curveNumber = d.trace.index;
 
+    // additionnal coordinates
+    n.dx = n.x1 - n.x0;
```
etpinard: Can we move this up in `nodeModel()` to have:

```js
var visibleThickness = n.dx;
var visibleLength = Math.max(0.5, n.dy);
```

as before?

src/traces/sankey/render.js
```diff
@@ -321,7 +251,7 @@ function textFlip(d) {return d.horizontal ? 'scale(1 1)' : 'scale(-1 1)';}
 function nodeTextColor(d) {return d.darkBackground && !d.horizontal ? 'rgb(255,255,255)' : 'rgb(0,0,0)';}
 function nodeTextOffset(d) {return d.horizontal && d.left ? '100%' : '0%';}
 
-// event handling
+// // event handling
```
etpinard: :zap: that extra `//`

src/traces/sankey/render.js
```diff
@@ -386,17 +320,24 @@ function attachDragHandler(sankeyNode, sankeyLink, callbacks) {
...
+                d.node.y1 = y + d.visibleHeight / 2;
             } else {
                 if(d.arrangement === 'freeform') {
-                    d.node.x = x;
+                    d.node.x0 = x - d.visibleWidth / 2;
+                    d.node.x1 = x + d.visibleWidth / 2;
+                    // d.x0 = x;
                 }
-                d.node.y = Math.max(d.node.dy / 2, Math.min(d.size - d.node.dy / 2, y));
+                // d.node.y = Math.max(d.node.dy / 2, Math.min(d.size - d.node.dy / 2, y));
```
etpinard: :zap:

src/traces/sankey/plot.js
```diff
@@ -73,12 +73,14 @@ function linkHoveredStyle(d, sankey, visitNodes, sankeyLink) {
     var label = sankeyLink.datum().link.label;
 
     sankeyLink.style('fill-opacity', 0.4);
+    sankeyLink.style('stroke-opacity', 0.4);
```
etpinard: Why do we need this now?

src/traces/sankey/render.js
```diff
@@ -489,33 +487,42 @@ module.exports = function(svg, styledData, layout, callbacks) {
         .style('fill', 'none');
 
     var sankeyLink = sankeyLinks.selectAll('.' + c.cn.sankeyLink)
-        .data(function(d) {
-            var uniqueKeys = {};
```
etpinard: Why is the `uniqueKeys` stash not used anymore?

src/traces/sankey/render.js
```diff
@@ -220,23 +124,55 @@ function linkModel(uniqueKeys, d, l) {
...
 }
 
-function nodeModel(uniqueKeys, d, n) {
+function linkPath() {
+    var curvature = 0.5;
+    function path(d) {
+        if(d.circular) {
+            return d.link.path; // TODO: turn this into a closed path to support link.line.(width|color)
+        } else {
+            var x0 = d.link.source.x1;
```
etpinard: Why do we need this new `linkPath` logic? I can't find this anymore in the `master` version of this file.

test/jasmine/tests/sankey_d3_sankey_circular_test.js
```diff
@@ -0,0 +1,76 @@
+// var Plotly = require('@lib/index');
+// var attributes = require('@src/traces/sankey/attributes');
+
+var d3SankeyCircular = require('d3-sankey-circular');
+var mockCircular = require('@mocks/sankey_circular.json');
+var convertToD3Sankey = require('@src/traces/sankey/convert-to-d3-sankey');
+
+describe('d3-sankey-ciruclar', function() {
```
etpinard: No need for two sankey test files. Can you move this `describe` to `sankey_test.js` and :hocho: this file?

test/jasmine/tests/sankey_d3_sankey_test.js
```diff
@@ -0,0 +1,130 @@
...
+        'value': 2
+    }, {
+        'source': 3,
+        'target': 4,
+        'value': 4
+    }]
+};
+
+
+describe('d3-sankey', function() {
```
etpinard: Similarly here, can you move this describe to sankey_test.js and :hocho: this file?

package.json
antoinerg: Those changes were minimal. One of the biggest was that we generated the SVG path for the links in it. As you found out in this PR, we can do operation in plotly.js instead.

package.json
antoinerg: Mmm... I guess we have a shot at getting a PR merged in. Basically, `d3-sankey-circular` is missing the API function `update()` present in `d3-sankey`. This is useful for implementing interactivity/animations.

src/traces/sankey/render.js
antoinerg: Basically, what's in the `else` clause used to live in our fork of `d3-sankey`.

etpinard:
Nice work @antoinerg !

I'm not too familiar with how the current version of `sankey` looks and feels, so I would be nice to check that this PR doesn't break anything (rendering-wise AND during interactions) in a few more graphs e.g from: https://plot.ly/feed/?q=plottype:sankey

A few more comments,

> * Links are not rectangles anymore but path (so we don't have a fill property, but only stroke)

About this here, this is true only for circular layout, correct? 

> I should probably add tests to make sure `sankey.update()` works properly for every (as of right now there's only 2) Sankey generators.

Yes please! Looks like the current `sankey_test.js` suite only tests restyle for `visible`:

https://github.com/plotly/plotly.js/blob/51085c41c6f3a0eede8da65e59a1be2e43ca256a/test/jasmine/tests/sankey_test.js#L314-L335

there should be more testing (not just for circular layout :smirk: ) testing restyle calls involving `link` / `node`.

package.json
etpinard: Ok that make sense. Do you think we have a shot at getting

https://github.com/d3/d3-sankey/compare/master...antoinerg:fix-large-padding

merged into `d3/d3-sankey`?

package.json
etpinard: > I guess we have a shot at getting a PR merged in

PR away please :rocket: !

package.json
antoinerg: There hasn't been anything merged into `master` in `d3/d3-sankey` since Jul 13, 2017 so I'm not sure... I opened a PR almost a month ago https://github.com/d3/d3-sankey/pull/63 and there has been no activity since then.

package.json
antoinerg: It got merged in pretty quickly! https://github.com/tomshanley/d3-sankey-circular/pull/28

It is not yet published on npmjs.com and I don't know if it will be. Is it OK to point to a git repo like I was doing? If it is, we should keep a copy of it in case the repo disappears. Let me know what you prefer!

src/traces/sankey/plot.js
antoinerg: Actually we won't need to touch `stroke-opacity` on hover anymore if we render links as closed paths instead of open paths with `stroke-width`

This file will end up more similar to the original one!  :tada:

antoinerg:
Here's a potentially unwanted behaviour:
sankey_misplaced_hover

cc @etpinard

alexcjohnson:
> hovers appear in the center of circular loops.

Good 👁! For forward links the bounding box center is always on the link, but for circular links it would be great to move the hover box anchor to the center of the back stretch.

src/traces/sankey/plot.js
antoinerg: Removed in d1d852ab3a4270d3b7a847ba84b6b0ae5e43e967

src/traces/sankey/render.js
antoinerg: I added even more code to generate SVG path in d1d852ab3a4270d3b7a847ba84b6b0ae5e43e967 just above that function.

The reason it's in here is because `d3-sankey` and `d3-sankey-circular` use open paths with a large `stroke-width` to render links whereas we use closed path that are filled. There are pros and cons to both approaches. You can see this PR that was refused for more details https://github.com/d3/d3-sankey/pull/24 . Therefore, because we have our own take on how to render those links, that logic should probably live in our repo.

package.json
```diff
@@ -58,15 +58,19 @@
...
-    "@plotly/d3-sankey": "^0.5.1",
     "alpha-shape": "^1.0.0",
     "array-range": "^1.0.1",
     "canvas-fit": "^1.5.0",
     "color-normalize": "^1.3.0",
     "convex-hull": "^1.0.3",
     "country-regex": "^1.1.0",
     "d3": "^3.5.12",
+    "d3-sankey": "git://github.com/antoinerg/d3-sankey.git#4f37ed8d3578b545a8569ecd74583f373768e900",
+    "d3-sankey-circular": "git://github.com/antoinerg/d3-sankey-circular.git#a298acf674f0a9c7158243d45814cd8060dad728",
```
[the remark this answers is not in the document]
antoinerg: Those changes got merged in :rocket: !

src/traces/sankey/convert-to-d3-sankey.js
antoinerg: Done in 04a186c55523373f7b87eb505bc7c02865ee8be4

src/traces/sankey/render.js
```diff
@@ -13,179 +13,83 @@ var d3 = require('d3');
...
-    }
-}
-
-function switchToSankeyFormat(nodes) {
-    // sankey uses x, y as top left
-    for(var i = 0; i < nodes.length; i++) {
-        nodes[i].y = nodes[i].y - nodes[i].dy / 2;
-    }
-}
+var interpolateNumber = require('d3-interpolate').interpolateNumber;
```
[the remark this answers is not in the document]
antoinerg: Our updated version of `d3-sankey` (https://github.com/antoinerg/d3-sankey/tree/fix-large-padding) doesn't have `d3-interpolate` as a dependency anymore so I think this is needed. Correct me if I'm wrong!

package.json
antoinerg: Done in dee73b87c23f4667992718d97251be27385f823b

test/jasmine/tests/sankey_d3_sankey_circular_test.js
antoinerg: Done in 4539d285285ebb22d6d2ce9ddd300e9e1de03b60

test/jasmine/tests/sankey_d3_sankey_test.js
antoinerg: Done in 4539d285285ebb22d6d2ce9ddd300e9e1de03b60

antoinerg:
> > hovers appear in the center of circular loops.
> 
> Good ! For forward links the bounding box center is always on the link, but for circular links it would be great to move the hover box anchor to the center of the back stretch.

@alexcjohnson Is the following what you had in mind? I am not sure what part is the "back stretch".

sankey_fixed_hover

alexcjohnson:
> @alexcjohnson Is the following what you had in mind? I am not sure what part is the "back stretch".

Yep, you got it exactly as I had in mind! https://en.wikipedia.org/wiki/Backstretch

src/traces/sankey/render.js
```diff
@@ -216,27 +120,177 @@ function linkModel(uniqueKeys, d, l) {
...
+    n.dx = n.x1 - n.x0;
+    n.dy = n.y1 - n.y0;
     var visibleThickness = n.dx;
     var visibleLength = Math.max(0.5, n.dy);
 
     var basicKey = n.label;
-    var foundKey = uniqueKeys[basicKey];
-    uniqueKeys[basicKey] = (foundKey || 0) + 1;
-    var key = basicKey + '__' + uniqueKeys[basicKey];
+    var key = basicKey + '__' + i;
```
antoinerg: This key should always be unique regardless of the node's label because of `i`: it is the index of the node in the graph array and it will always be unique.

Please correct me if I'm wrong!

src/traces/sankey/render.js
antoinerg: It doesn't seem like we need a stash to make sure we generate unique keys for our links and nodes. See comment https://github.com/plotly/plotly.js/pull/3406/files#r249568603

Therefore I :hocho: it to simplify the code.

src/traces/sankey/render.js
etpinard: That's sound ok to me :ok_hand: 

Resolved.

antoinerg:
In commit 9b36090, I :lock: down `sankey.update()`. It allowed me to discover a little bug in `d3-sankey-circular`: it wouldn't force circular links to go from top to bottom on the first invocation of `update()`. Because we make the call to `update()` in the animation loop over and over, this error on first call was not perceivable to the user.

I will send a PR of the fix upstream. BTW, its maintaner @tomshanley said he could publish an npm package as soon as he gets back from holiday (https://github.com/tomshanley/d3-sankey-circular/pull/28#issuecomment-456227142).

antoinerg:
@etpinard Tomorrow I intend to create/update our own forks for both `d3-sankey` and `d3-sankey-circular` and publish npm packages for them. Was updating our deps the only remaining thing blocking this PR?

etpinard:
> Was updating our deps the only remaining thing blocking this PR?

Yes. Your new tests look good!

test/jasmine/tests/sankey_test.js
```diff
@@ -908,208 +908,252 @@ function assertNoLabel() {
...
+        node.x0 += delta[0];
+        node.x1 += delta[0];
+        node.y0 += delta[1];
+        node.y1 += delta[1];
 
-    it('controls the padding between nodes', function() {
-        expect(sankey.nodePadding()).toEqual(10, 'incorrect nodePadding');
-    });
+        // Update links
+        var updatedGraph = sankey.update(graph);
```
etpinard: Actually, can we make these test use `restyle` instead of (the internal) `sankey.update`?

antoinerg: > Actually, can we make these tests use `restyle` instead of (the internal) `sankey.update`?

I don't think we can do that right now because we don't have any attributes to control the position of the nodes (so no `restyle` call can change the position of the nodes). Right now, `sankey.update` is only called when nodes are dragged around with the mouse.

Maybe tests using `restyle` could be included in a future PR implementing attributes `node.x` and `node.y`? Or maybe I should add those attributes here right now?

etpinard: Ha I see. Good point.

Can you add (just) one `restyle` test that update circular to non-circular node and links data? Or better yet, one `react` test that goes from one circular mock to a non-circular one?

etpinard: One last TODO :arrow_up:

etpinard:
Amazing work!

:dancer: :dancer: :dancer:

PR #3408, opened 2019-01-07
finalist: Errorbars inherit color from line or marker color

### LINKED ISSUE #3392: Error bars not inheriting color

archmoj:
fix #3392 for 5 traces.

Note 1: For `bar` and `histogram` cases `marker.line.color` is applied as the default color. If undefined then `marker.color' is used if defined. 

Note 2: For `scatter`, `scatter3d` and `scattergl` cases `line.color` is applied as the default color. If undefined then `marker.color' is used if defined. Also note that `marker.line.color` is not used in this regard; and the trace line color and errorbar lines would use `colorway` in that scenario.

@plotly/plotly_js


etpinard:
Hmm. I don't think we should be considering this a bug fix. This changes a few of our baselines.

Personally, I think this PR belongs in the `v2` push, but I'd be _ok_ with calling this a feature set for the next minor v1.44.0.

src/traces/bar/defaults.js
```diff
@@ -70,10 +70,13 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
 
+    var lineColor = (traceOut.marker.line || {}).color;
+    var markerColor = (traceOut.marker || {}).color;
+
     // override defaultColor for error bars with defaultLine
     var errorBarsSupplyDefaults = Registry.getComponentMethod('errorbars', 'supplyDefaults');
-    errorBarsSupplyDefaults(traceIn, traceOut, Color.defaultLine, {axis: 'y'});
-    errorBarsSupplyDefaults(traceIn, traceOut, Color.defaultLine, {axis: 'x', inherit: 'y'});
+    errorBarsSupplyDefaults(traceIn, traceOut, lineColor || markerColor || Color.defaultLine, {axis: 'y'});
+    errorBarsSupplyDefaults(traceIn, traceOut, lineColor || markerColor || Color.defaultLine, {axis: 'x', inherit: 'y'});
```
alexcjohnson: I don’t think `markerColor` belongs here. That would be invisible against the bars themselves, right? I like `lineColor` though.

archmoj: Yes totally agree.

alexcjohnson:
True, it contains a bit of a change, but mostly I’d call this fixing a bug. To me it’s not such a change that it should be considered breaking but I can understand not calling it just a patch.

src/traces/bar/defaults.js
archmoj: @alexcjohnson there is also [this PR](https://github.com/plotly/plotly.js/pull/3414) for possibly only fixing this on scatter traces...

etpinard:
> To me it’s not such a change that it should be considered breaking but I can understand not calling it just a patch.

Sounds good, let's panned this one for `1.44.0`.

etpinard:
:dancer: nicely done @archmoj !

PR #3414, opened 2019-01-07
fix 3392 scatter errorbars inherit color

### LINKED ISSUE #3392: Error bars not inheriting color

archmoj:
fix #3392 for 3 `scatter` traces only.

Unlike #3408 there is no baseline change if applied for scatter traces.
@etpinard in respect to [your comment](https://github.com/plotly/plotly.js/pull/3408#issuecomment-451972760)

For `scatter`, `scatter3d` and `scattergl` cases `line.color` is applied as the default color. If undefined then `marker.color' (if defined) would be used.


etpinard:
https://github.com/plotly/plotly.js/pull/3408 is now planned for 1.44.0. Closing.

PR #3421, opened 2019-01-08
Updating dependencies of gl-vis modules

archmoj:
Various gl-vis modules are updated e.g. to use glslify 6.4.1, etc. 

@etpinard 
CC: @plotly/plotly_js


etpinard:
Is `glslify` the only module you updated in our gl-vis deps? Sorry for being lazy and not checking myself :smile:

package-lock.json
```diff
@@ -4698,10 +4697,71 @@
...
         "gl-vao": "^1.1.1",
         "gl-vec4": "^1.0.1",
-        "glslify": "^6.1.0",
+        "glslify": "^6.4.1",
         "robust-orientation": "^1.1.3",
         "split-polygon": "^1.0.0",
         "vectorize-text": "^3.2.0"
+      },
+      "dependencies": {
+        "glslify": {
```
etpinard: Hmm. That seems off. Looks like `glslify` didn't get dedupe.

Can you run `npm ls glslify` and paste its result here?

etpinard:
Ok,

image


nicely done. :muscle: 

Go ahead and bump the gl-vis modules' versions.

package.json
```diff
@@ -101,10 +101,10 @@
     "point-cluster": "^3.1.4",
     "polybooljs": "^1.2.0",
     "regl": "^1.3.11",
-    "regl-error2d": "^2.0.5",
-    "regl-line2d": "^3.0.12",
-    "regl-scatter2d": "^3.1.2",
-    "regl-splom": "^1.0.5",
+    "regl-error2d": "git://github.com/gl-vis/regl-error2d#0d27313d932fdb8ffaed17931b7258958bce0b88",
```
archmoj: @etpinard missing `npm` permission to publish this patch. Hopefully @dy would help?

package-lock.json
```diff
@@ -4806,19 +4806,26 @@
...
         "gl-vao": "^1.1.3",
-        "glsl-out-of-range": "^1.0.3",
+        "glsl-out-of-range": "^1.0.4",
         "glsl-read-float": "^1.0.0",
-        "glslify": "^6.1.0",
+        "glslify": "^7.0.0",
         "ndarray": "^1.0.16"
+      },
+      "dependencies": {
+        "binary-search-bounds": {
```
etpinard: Could you check `npm ls binary-search-bounds`? I think I tried using `binary-search-bounds@2.x` in gl3d, but I don't think that worked out. We should double-check.

package.json
etpinard: I don't have npm publish rights for `regl-error2d` either :cry:

etpinard: -> https://github.com/gl-vis/regl-error2d/issues/3 :crossed_fingers:

etpinard: @archmoj you should now have `npm publish` rights.

archmoj: I do have. Let me publish the new package. Thanks @etpinard.

package-lock.json
etpinard: @archmoj this here is now the only potentially-blocking item.

What are the differences between `binary-search-bounds` v1.x and v2.x?

archmoj: I think we may leave that for another PR. It seems that I don't have permission to patch and publish those modules...

archmoj: https://github.com/mikolalysenko/binary-search-bounds/commit/d047eeeff4cb388d423b017b8f3100dda5adefa6#diff-b9cfc7f2cdf78a7f4b91a753d10865a2
This may be relevant to the difference between version 1 and 2.
`ndarray`s are not supported in Version 2.

etpinard:
Ok, some before and after:

on `master`:

image

image

now on this branch:

image

which is :tada: , and

image

which, as you pointed out, only has https://github.com/mikolalysenko/binary-search-bounds/commit/d047eeeff4cb388d423b017b8f3100dda5adefa6 diff between v1.x and v2.x. So I'd call this an improvement!

etpinard:
@archmoj let's merge this in :dancer: 

Our gl-vis modules have never been this well maintained!

archmoj:
Thanks @etpinard for another encouraging and great review.
Yes I think we may merge this earlier so that to give the team more time to possibly test any side effect before the v.1.44.0 release.

Issue #3337, opened 2018-12-14
responsive: true breaks scroolZoom: true


IvanJosipovic:
With the version 1.42.5 of plotly.js enabling **responsive: true** breaks **scroolZoom: true**

Working:
https://codepen.io/anon/pen/YdqpoO

Broken:
https://codepen.io/anon/pen/REaozq

etpinard:
Probably from the same weird block identified in https://github.com/plotly/plotly.js/pull/3085

https://github.com/plotly/plotly.js/blob/9bbe2956f87bf2f42101db8b7e08375b63dc613a/src/plots/cartesian/dragbox.js#L437-L442

etpinard:
Commenting out:

https://github.com/plotly/plotly.js/blob/9bbe2956f87bf2f42101db8b7e08375b63dc613a/src/plots/cartesian/dragbox.js#L437-L442

indeed "fixes" this bug. I'll try to get patch similar to https://github.com/plotly/plotly.js/pull/3085 merged in time for `v1.44.0`.

PR #3424, opened 2019-01-09
Do not disable scrollZoom scroll{Width,Height} > client{Width,Height}

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2371 and https://github.com/plotly/plotly.js/issues/3337 by including a commit from @marfoldi's https://github.com/plotly/plotly.js/pull/3085.

Removing the `scroll{Width,Height} > client{Width,Height}` early return fixes two bugs:

- it fixes https://github.com/plotly/plotly.js/issues/3337 where `{responsive: true, scrollZoom: true}` + graph div height `vh` units doesn't work, as in that scenario `gd.client{Width,Height}` is `0` leading to an early return.
- it fixes https://github.com/plotly/plotly.js/issues/2371 where we get an early return when page is scrollable. See https://github.com/plotly/plotly.js/pull/3085#pullrequestreview-170331290 for more details.

There are probably other ways to fix these bugs, but that `scroll{Width,Height} > client{Width,Height}` early return feels out-of-place. It's probably a relic from the workspace days, where we set `scrollZoom:true`, but wanted some extra do-not-scroll logic. As scroll zoom is disable by default for cartesian subplots, I don't think overriding a `scrollZoom: true` user config is a good idea.

I hope @plotly/plotly_js agrees. If so, I'll add a few tests.


etpinard:
@antoinerg would you mind taking a look at this one at some point this week?

etpinard:
Oh wait. I still have to add a few tests for this one.

Asking for @alexcjohnson 's opinion on this potentially-controversial patch.

alexcjohnson:
Agreed. Explicit is better than implicit.

etpinard:
> Oh wait. I still have to add a few tests for this one.

Done in https://github.com/plotly/plotly.js/pull/3424/commits/3ee07987d59a9bea8d39e70f1b5e77e8aa8f813e and https://github.com/plotly/plotly.js/pull/3424/commits/09e0a5d8b58a7f2e67a4d7237f9c5358349cb92a

----

@antoinerg I know you're busy, but would you mind taking a :eye: at this PR (should be quick :wink: )

src/plots/cartesian/dragbox.js
```diff
@@ -430,17 +430,8 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
...
-        var pc = gd.querySelector('.plotly');
-
         recomputeAxisLists();
 
-        // if the plot has scrollbars (more than a tiny excess)
-        // disable scrollzoom too.
-        if(pc.scrollHeight - pc.clientHeight > 10 ||
-                pc.scrollWidth - pc.clientWidth > 10) {
-            return;
-        }
```
antoinerg: I am not sure what this early return was meant to achieve but I guess it's OK to :hocho: considering it doesn't have any associated tests...

antoinerg:
Thank you for the fix @etpinard! It seems to work well: https://codepen.io/anon/pen/roRdmB

If the removed code really is a relic, then :dancer:

Issue #3432, opened 2019-01-12
"legend.valign" doesn't work for pie charts


michaelbabyn:
If I set `valign='top'` in a pie chart, multi-line legend items will still be centered around the legend marker. 

[codepen for pie chart with issue](https://codepen.io/michaelbabyn/pen/jXQJME)
[codepen for scatter chart demonstrating the desired behavior](https://codepen.io/michaelbabyn/pen/VqVRKx)

etpinard:
Thanks for the report!

PR #3435, opened 2019-01-14
legend valign fix for pie traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3432 - a follow-up for https://github.com/plotly/plotly.js/pull/3263

cc @plotly/plotly_js


antoinerg:
Very nice fix and a good mock! That was fast @etpinard 

:dancer:

Issue #3427, opened 2019-01-11
layout.metatext


nicolaskruchten:
For Chart Studio we would need:

* A new data-array key in layout called `metatext`, intended to contain strings
* The ability to use `%{metatext[<x>]}` in string keys like `layout.title`, `layout.*axis.title`, `layout.annotations[*].text` (and possibly one or two other places TBD)

etpinard:
Just to confirm, we want something like:

```js
Plotly.newPlot(gd, [{
  y: [1,2,1]
}], {
   metatext: ['a', 'b'],
   title: 'my title %{metatext[0]} - %{metatext[1]}'
})
```

yielding:

![image](https://user-images.githubusercontent.com/6675409/51040979-d7b0cd00-1586-11e9-900a-384ddd31cb36.png)

?

nicolaskruchten:
yes indeed!

PR #3439, opened 2019-01-15
Metatext

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3427

cc @plotly/plotly_js and @nicolaskruchten


alexcjohnson:
Does it need to be text? Should we just call it `layout.meta`? I'm imagining in the short term templating numbers (potentially with a format spec), and longer term possibly using this as a way to share other settings across multiple traces, axes, etc.

etpinard:
> Should we just call it `layout.meta`

I don't have a strong opinion.

nicolaskruchten:
Looks great to me! `meta` works for me, although there is no requirement today to support anything beyond text, we may as well leave the door open.

etpinard:
> Looks great to me! `meta` works for me

done in https://github.com/plotly/plotly.js/pull/3439/commits/b03e77769a817ae85b40b4571ef8e2f473bf0dc0

nicolaskruchten:
💃

nicolaskruchten:
Just so the plotly.py machinery etc doesn't complain about Chart Studio sliding that in...

etpinard:
> do we need to add anything for `metasrc`

That's done automatically here:

https://github.com/plotly/plotly.js/blob/7d67f8011a0200cc09535cebde8de34199f85a09/src/plot_api/plot_schema.js#L617-L622

nicolaskruchten:
ok great :)

Issue #3440, opened 2019-01-17
Box plot: hover position of points is off


aaronpries:
In a box plot with boxpoints='all', the hover position of the boxes shows the correct position of the data points. The boxpoints are shifted to the left and the hoverinfo shows their x-position in the figure, which is not the same as the x-position of the data points, see the attached figure (created with plotly.py: [FigureWidget.txt](https://github.com/plotly/plotly.js/files/2768130/FigureWidget.txt)).
![boxplot_hover](https://user-images.githubusercontent.com/9529130/51311227-b8afb080-1a48-11e9-824a-dca040ad0412.png)

Is this desired? I would have expected to see the x-value of the data points, not the position in the figure. If a jitter is added, this may be even more confusing.

alexcjohnson:
Thanks @aaronpries, that’s most definitely not correct, it should show the data value as you say.

etpinard:
Thanks for the report @aaronpries !

etpinard:
In a codepen: https://codepen.io/etpinard/pen/NeVaYR?editors=1010

PR #3441, opened 2019-01-17
Show box/violin position when hovering over pts

etpinard:
... instead of showing the point's position - fixes https://github.com/plotly/plotly.js/issues/3440

https://codepen.io/etpinard/pen/NeVaYR?editors=1010

now looks like:

peek 2019-01-17 10-27

Note that this was only a problem for box/violin traces with numeric positions (e.g. with set `x` and `y` arrays) under `hovermode:'closest'`, as deeper down `Fx.hover` 

https://github.com/plotly/plotly.js/blob/cc4597257b411688b2bbc6dbe90f8a6e78cc96c8/src/components/fx/hover.js#L1341-L1348

calls [`Axes.hoverLabelText`](https://github.com/plotly/plotly.js/blob/cc4597257b411688b2bbc6dbe90f8a6e78cc96c8/src/plots/cartesian/axes.js#L1008-L1025) which then formats the positions back to categories with: 

https://github.com/plotly/plotly.js/blob/cc4597257b411688b2bbc6dbe90f8a6e78cc96c8/src/plots/cartesian/axes.js#L1164-L1168

### DIFF
--- a/src/traces/box/hover.js
+++ b/src/traces/box/hover.js
@@ -240,15 +240,25 @@ function hoverOnPoints(pointData, xval, yval) {
         name: trace.name,
         x0: xc - rad,
         x1: xc + rad,
-        xLabelVal: pt.x,
         y0: yc - rad,
         y1: yc + rad,
-        yLabelVal: pt.y,
         spikeDistance: pointData.distance
     });
-    var pLetter = trace.orientation === 'h' ? 'y' : 'x';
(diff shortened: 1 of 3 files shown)


etpinard:
cc @plotly/plotly_js

antoinerg:
Thanks for the fix @etpinard! Test is clear and simple :ok_hand: 

Codepen with the fix: https://codepen.io/anon/pen/BMaNzV

:dancer:

Issue #3390, opened 2019-01-03
Violins and Boxes not honoring 'group'


nicolaskruchten:
I'm seeing some odd inconsistencies in overlay vs group modes with boxes/violins... In all of the following cases, `violinmode = "group"`:

![image](https://user-images.githubusercontent.com/203523/50653157-9bfa7f80-0f56-11e9-9b5f-3075998f2b72.png)

![image](https://user-images.githubusercontent.com/203523/50653171-a61c7e00-0f56-11e9-8bc3-7cd98574ad9a.png)

![image](https://user-images.githubusercontent.com/203523/50653190-af0d4f80-0f56-11e9-8c5a-ea6858065d03.png)

![image](https://user-images.githubusercontent.com/203523/50653203-b92f4e00-0f56-11e9-823d-58ff0eae5cc9.png)

The last one is where the problem is: why are they overlaid?

etpinard:
> The last one is where the problem is: why are they overlaid?

Would it be too much to ask to share the "data" / "layout" of that graph. I'm not too familiar with `px` (yet). Thanks!

nicolaskruchten:
https://plot.ly/~nicolaskruchten/252

nicolaskruchten:
OK, so the reason it's not grouping is because of this block: https://github.com/plotly/plotly.js/blob/master/src/traces/box/cross_trace_calc.js#L80-L84

I'm not super clear on what it's doing or why but temporarily deleting that block does force things back into grouping mode, like it says in the comment. BUT I then get this odd behaviour, where it's grouping "too much"... it's like it doesn't think that the various "female" groups are the same and so it's leaving a bunch of extra room in each subplot :)

![image](https://user-images.githubusercontent.com/203523/50712127-51ecc900-103e-11e9-9efd-eb12a46e3f13.png)

nicolaskruchten:
Maybe related? https://github.com/plotly/plotly.py/issues/1389

nicolaskruchten:
Hmm actually the “too much grouping” is present in the first two screenshots as well, just less obviously-so. That might just be a separate issue.

etpinard:
Consider the following:

```
Plotly.newPlot('graph', [{
  type: 'box',
  x0: 1,
  y: [1,2,1,2,1,2,3,4,4],
}, {
  type: 'box',
  x0: 2,
  y: [2,1,2,3,3,1,0,0,1]
}], {
  boxmode: 'group'
})
```

which currently looks like:

[...]

PR #3445, opened 2019-01-17
Remove box ungrouping magic

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3390 as per @nicolaskruchten 's https://github.com/plotly/plotly.js/issues/3390#issuecomment-455016558 

To be merged into https://github.com/plotly/plotly.js/pull/3234 to avoid conflicts in `box/cross_trace_calc.js`

cc @plotly/plotly_js


antoinerg:
@etpinard I am not too familiar with `box` trace but I am a bit confused by the result of the new baseline you introduced in this PR. According to the description of `boxmode`:
`If *group*, the boxes are plotted next to one another centered around the shared location.`. If I only have one box at a given location, I would expect the *shared location* to be the box's center position. Yet, the boxes in baseline `box_single-group.json` are not aligned anymore with the tick labels :thinking: Is this normal?

etpinard:
> I would expect the _shared location_ to be the box's center position. Yet, the boxes in baseline `box_single-group.json` are not aligned anymore with the tick labels  Is this normal?

See discussion in https://github.com/plotly/plotly.js/issues/3390#issuecomment-454969238

antoinerg:
> See discussion in #3390 (comment)

I think I understand after a careful second read. I agree, let's :hocho: that magic!

:dancer:

PR #3453, opened 2019-01-22
Description of borderwidth

DenrizSusam:
description of borderwidth should be "Sets the border width" instead of "Sets the border color..."


etpinard:
Thanks! I'll fix the failing test in a different PR.

PR #3460, opened 2019-01-22
Allow fixedrange axes to scaleanchor with `constrain:domain`

etpinard:
Compare

- https://codepen.io/etpinard/pen/daPyXd 
- https://codepen.io/etpinard/pen/MLYWjy

where the first one has `fixedrange:true` on both x and y axes, and the second does not. This was first reported in https://community.plot.ly/t/axis-constraints-without-zoom-use-scaleanchor-with-fixedrange/18354

Now, the scaleanchor logic is skipped for `fixedrange:true`. This makes sense when constraining by range (from https://github.com/plotly/plotly.js/pull/1522), but does it also make sense when constraining by domain (from https://github.com/plotly/plotly.js/pull/1767)? 

Commit https://github.com/plotly/plotly.js/commit/dee136d2eab529f27cc7472a93bcb6c75f8cfed4 was added in the first scaleanchor PR #1522, to:

https://github.com/plotly/plotly.js/blob/e4e1149e20312892e261f6d349ef34c25a6cbf34/src/plots/cartesian/constraint_defaults.js#L58-L61

I can't imagine a scenario where `constrain:'domain'` would lead to an infinite loop.

@alexcjohnson @plotly/plotly_js can you think any?


alexcjohnson:
The prohibition on `fixedrange` wasn't due to infinite loops, it was just that if you've fixed a range and constrained another range based on that, the second range would be fixed too, at least in span.

I suppose if *both* axes are `fixedrange` *and* `constrain: 'domain'` (as in your mock) there are no concerns: one of them will get a reduced domain but both will keep the specified ranges. If *one* is `fixedrange` (and `constrain: 'domain'`), the other one could in principle be either `constrain`, and it would need to be restricted to allow pan but not zoom.

etpinard:
Will be part of https://github.com/plotly/plotly.js/pull/3506

PR #3380, opened 2018-12-31
Make vectorize-text functions top-level

archmoj:
Fixes the error reported in #3379.
@plotly/plotly_js

PR #3463, opened 2019-01-23
Disable dark highlights from Mesh3d

### LINKED ISSUE #3462: Black highlights on mesh3d on some devices
Areas of the `mesh3d` & `isosurface` that are opposite to the light source may be rendered in black on some devices e.g. Android.
This is a regression from #3415. 
newplot 18

archmoj:
Fixes #3462 using newly published `gl-mesh3d` module by not allowing negative highlights.
Example below illustrate a desirable result after using `gl-mesh3d` v.2.0.6
@etpinard cc: @plotly/plotly_js 
newplot 1 1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4854,9 +4854,8 @@
       }
     },
     "gl-mesh3d": {
-      "version": "2.0.5",
-      "resolved": "https://registry.npmjs.org/gl-mesh3d/-/gl-mesh3d-2.0.5.tgz",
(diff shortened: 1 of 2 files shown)


etpinard:
Nice fix. Is there any way to test this fix?

archmoj:
I used a mobile with [this codepen](https://codepen.io/MojtabaSamimi/pen/zexNbP).
Unfortunately I don't see a way to write a test for it.

etpinard:
Ok, sounds good :dancer:

etpinard:
Thanks for the quick fix @archmoj !

This thing will be out today, but I'll wait a couple more hours before releasing 1.44.1 in case other regression are detected.

archmoj:
It would be great if we could follow up with #3380.

PR #3469, opened 2019-01-23
Make surface contour levels configurable

### LINKED ISSUE #485: Make surface contour levels configurable
... with `autocontours`, `ncontours`, `contour.start`, `contour.end` and `contour.size`.

Looks like [`gl-surface3d`](https://github.com/gl-vis/gl-surface3d/blob/master/surface.js#L124) already has that feature; it would just a matter of exposing the contour calc [step](https://github.com/plotly/plotly.js/blob/1f406e5bf7c4396f3371699c4f61cb406f61b2e6/src/traces/contour/calc.js#L24-L48) so that the surface convert step can reuse it to fill the `contourLevels` gl-surface3d property.

archmoj:
Attempt to address original ticket #485 to make surface contours adjustable and aligned with `heatmap` & `contour` trace.
In this PR `contour.start`, `contour.end` and `contour.size` are added and could be applied by the user to slice surface graphs on desired positions other than tick positions which is the default behaviour.

[Codepen example](https://codepen.io/MojtabaSamimi/pen/gyrrXq?editors=1000)

@plotly/plotly_js


etpinard:
Hmm. In https://github.com/plotly/plotly.js/issues/485, we propose making surface contour settings on-par with the attributes for `contour` traces. @archmoj are your new attributes equivalent to `contour`'s

> `autocontours`, `ncontours`, `contour.start`, `contour.end` and `contour.size`

? If not, could you explain why your new attribute make more sense for `surface` traces.

archmoj:
> Hmm. In #485, we propose making surface contour settings on-par with the attributes for `contour` traces. @archmoj are your new attributes equivalent to `contour`'s
> 
> > `autocontours`, `ncontours`, `contour.start`, `contour.end` and `contour.size`
> 
> ? If not, could you explain why your new attributes make more sense for `surface` traces.

I thought this setup could provide more flexibility so that the users can put the contours where they may need.
For example if they want them to be non-uniformly spaced, or in case of skipping one or few contours from the list, it is as simple as providing an array of desired world locations (i.e. similar to iso-surface slices). 
Also in regards to @alexcjohnson's comment https://github.com/plotly/plotly.js/issues/485#issuecomment-397604971 similar options are provided for relative locations on/between grid points.

etpinard:
I would prefer reaching parity with the attributes for `contour`, but I could be convinced otherwise.

archmoj:
> I would prefer reaching parity with the attributes for `contour`, but I could be convinced otherwise.

Would you like me to implement `ncontours`, `contour.start`, `contour.end` and `contour.size` as well?
That's relatively easy.

alexcjohnson:
> > I would prefer reaching parity with the attributes for `contour`, but I could be convinced otherwise.
> 
> Would you like me to implement `ncontours`, `contour.start`, `contour.end` and `contour.size` as well?
> That's relatively easy.

Definitely - the primary goal should be that you can take anything that works in a `type: 'contour'` trace, change just the `type` attribute to `'surface'`, and it still displays all the same contours.

Then at some point we can extend BOTH of these to support an array of contour values - perhaps `contour.values` - for the nonuniform case.

src/traces/surface/attributes.js
```diff
@@ -44,6 +44,62 @@ function makeContourAttr(axLetter) {
...
+            min: 0,
+            role: 'style',
+            editType: 'plot',
+         // impliedEdits: {'^autocontour': false},
+            description: [
+                'Sets the step between each contour level.',
+                'Must be positive.'
+            ].join(' ')
+        },
+        locations: {
```
etpinard: Do we really need `locations` and `relative` to :white_check_mark: this feature?

I would prefer making `surface` on-par with `contour` to start.

archmoj: Good call. Let's put this on 1.47.0.

archmoj: Dropped those features in https://github.com/plotly/plotly.js/pull/3469/commits/98aebc52f9d56eee09a95828d851f75395903e7c as `surface` code is already pretty complicated for us to maintain.

src/traces/surface/convert.js
```diff
@@ -349,19 +352,54 @@ proto.refineCoords = function(coords) {
...
+            if(
+                this.contourSize[i] > 0 &&
+                this.contourStart[i] !== null &&
+                this.contourEnd[i] !== null &&
+                this.contourEnd[i] > this.contourStart[i]
+            ) {
+                useNewLevels[i] = true;
+
+                for(j = this.contourStart[i]; j < this.contourEnd[i]; j += this.contourSize[i]) {
+                    value = j * this.scene.dataScale[i];
```
etpinard: Have you tried re-using [`setContours`](https://github.com/plotly/plotly.js/blob/master/src/traces/contour/set_contours.js#L16)? It probably gives similar results, but it would be nice to double-check.

archmoj: I tried that on a different branch. Also wanted to add `autocontours` and `ncontours` in this PR. 
But unfortunately that could add too much complexity while surface contours are in 3 directions not only z.

etpinard:
It would've been nice to bring this a bit more on-par with `contour`, but I guess that can wait for the regl-rewrite.

This will be a nice feature for 1.47.0 :dancer:

PR #3465, opened 2019-01-23
Do not set camera mode to turntable when up.z is zero

archmoj:
Fixes #3425 by not defaulting to `turntable` when `camera.up.z` is set to zero.
@etpinard

PR #3475, opened 2019-01-24
change camera up vector in plot api react test

archmoj:
Remaining change from PR #3465.
Fix https://github.com/plotly/plotly.js/pull/3465#issuecomment-457238107 but using z value other than zero.
@etpinard


etpinard:
:dancer: !!

PR #3483, opened 2019-01-25
No more allowing specular greater than one on mesh3d

### LINKED ISSUE #3482: White highlights of mesh3d on some devices
Areas of the `mesh3d` & `isosurface` may be rendered entirely in white on some devices e.g. Android.
Similar to issue #3462 and PR #3463 concerning the `minimum` specular there should be a limit for `maximum` specular level too.
before

archmoj:
Fixes #3482 by setting the limit of one and bumping `gl-mesh3d` module to 2.0.7
Image below illustrates the correct behaviour when compared to the one presented on #3482.
@etpinard 
after

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4854,9 +4854,9 @@
       }
     },
     "gl-mesh3d": {
-      "version": "2.0.6",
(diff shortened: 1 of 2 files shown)


etpinard:
:dancer: !

PR #3484, opened 2019-01-25
responsive tests: check size of DOM elements with getBoundingClientRect()

antoinerg:
This makes for more robust tests since a `style` attribute can override the `width` and `height` attribute of a DOM element. We now check for both to be safe.

@plotly/core-contributors

### DIFF
--- a/test/jasmine/tests/config_test.js
+++ b/test/jasmine/tests/config_test.js
@@ -553,7 +553,7 @@ describe('config argument', function() {
 
     describe('responsive figure', function() {
         var gd;
-        var data = [{x: [1, 2, 3, 4], y: [5, 10, 2, 8]}];
+        var data = [{type: 'scatter', x: [1, 2, 3, 4], y: [5, 10, 2, 8]}];
         var width = 960;
         var height = 800;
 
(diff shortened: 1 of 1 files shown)


test/jasmine/tests/config_test.js
```diff
@@ -553,7 +553,7 @@ describe('config argument', function() {
 
     describe('responsive figure', function() {
         var gd;
-        var data = [{x: [1, 2, 3, 4], y: [5, 10, 2, 8]}];
+        var data = [{type: 'scatter', x: [1, 2, 3, 4], y: [5, 10, 2, 8]}];
```
antoinerg: It is `scatter` by default, but it is clearer if we make it explicit.

test/jasmine/tests/config_test.js
```diff
@@ -594,6 +601,13 @@ describe('config argument', function() {
             .then(delay(RESIZE_DELAY))
             .then(function() {
                 checkLayoutSize(elWidth / 2, elHeight / 2);
+
+                var mainSvgs = document.getElementsByClassName('main-svg');
+                checkElementsSize(mainSvgs, elWidth / 2, elHeight / 2);
+
+                var canvases = document.getElementsByTagName('canvas');
```
antoinerg: Checking the size of the `canvas` elements sandwiched between the SVG elements is also very important.

VeraZab:
@antoinerg , maybe use `@plotly/plotly_js` ? `@plotly/core-contributors` notifies a lot more people, who don't necessarily work with plotly.js :)

etpinard:
:dancer: thanks!

Next time though:

- ping @plotly/plotly_js 
- I don't see why this commit needs a PR of its own. It could have been part of https://github.com/plotly/plotly.js/pull/3486

PR #3486, opened 2019-01-25
fix responsive figures with WebGL

### LINKED ISSUE #3485: responsive figures with WebGL are not properly sized
As shown in this Codepen, responsive figures using WebGL do no get properly sized in modern CSS layouts such as `flexbox`:  https://codepen.io/anon/pen/RvWvBa

antoinerg:
Closes #3485 

Commit 16e1f4f simply adds `scattergl` in the responsive test suite and fails. Commit f8c959d makes the tests pass.

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -244,8 +244,6 @@ exports.plot = function(gd, data, layout, config) {
                     'position': 'absolute',
                     'top': 0,
                     'left': 0,
-                    'width': '100%',
-                    'height': '100%',
                     'overflow': 'visible',
                     'pointer-events': 'none'
                 });
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -105,7 +105,8 @@ exports.plot = function(gd, data, layout, config) {
 
     // hook class for plots main container (in case of plotly.js
     // this won't be #embedded-graph or .js-tab-contents)
-    d3.select(gd).classed('js-plotly-plot', true);
+    d3.select(gd).classed('js-plotly-plot', true)
+      .style('position', 'relative');
 
     // off-screen getBoundingClientRect testing space,
     // in #js-plotly-tester (and stored as Drawing.tester)
@@ -536,8 +537,7 @@ function plotLegacyPolar(gd, data, layout) {
     var paperDiv = plotContainer.selectAll('.svg-container')
         .data([0]);
     paperDiv.enter().append('div')
-        .classed('svg-container', true)
-        .style('position', 'relative');
+        .classed('svg-container', true);
 
     // empty it everytime for now
     paperDiv.html('');
@@ -3742,8 +3742,7 @@ function makePlotFramework(gd) {
     // Make the svg container
     fullLayout._paperdiv = fullLayout._container.selectAll('.svg-container').data([0]);
     fullLayout._paperdiv.enter().append('div')
-        .classed('svg-container', true)
-        .style('position', 'relative');
+        .classed('svg-container', true);
 
     // Make the graph containers
     // start fresh each time we get here, so we know the order comes out
--- a/test/jasmine/tests/config_test.js
+++ b/test/jasmine/tests/config_test.js
@@ -551,230 +551,232 @@ describe('config argument', function() {
         });
     });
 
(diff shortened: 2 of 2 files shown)


src/plot_api/plot_api.js
```diff
@@ -244,8 +244,6 @@ exports.plot = function(gd, data, layout, config) {
                     'position': 'absolute',
                     'top': 0,
                     'left': 0,
-                    'width': '100%',
-                    'height': '100%',
```
antoinerg: I don't think those two lines were ever needed since a few lines later we set the size of the canvas via its attributes anyway:
https://github.com/plotly/plotly.js/blob/f8c959d3c2f9c7be683754a9422b1b3cda7e032e/src/plot_api/plot_api.js#L253-L255

etpinard: Nice find!

test/jasmine/tests/config_test.js
```diff
@@ -551,230 +551,232 @@ describe('config argument', function() {
...
+            }
 
-        it('@flaky should still be responsive if the plot is purged and replotted', function(done) {
-            fillParent(1, 1);
-            Plotly.plot(gd, data, {}, {responsive: true})
-            .then(function() {return Plotly.newPlot(gd, data, {}, {responsive: true});})
-            .then(testResponsive)
-            .then(done);
-        });
+            it('@flaky should resize when the viewport width/height changes', function(done) {
```
etpinard: We should add the `traceType` value inside these `it`s so that if the scatter or the scattergl cases fail, we'll know which one did. For example,

```js
it('@flaky should resize when the viewport width/height changes (case ' + traceType + ')', function(done) {
// ...
```

test/jasmine/tests/config_test.js
```diff
@@ -551,230 +551,232 @@ describe('config argument', function() {
...
-            parent.style.top = '0';
-            parent.style.left = '0';
-        });
-
-        afterEach(function() {
-            Plotly.purge(gd); // Needed to remove all event listeners
-            document.body.removeChild(parent);
-            viewport.reset();
-        });
+    ['scatter', 'scattergl'].forEach(function(traceType) {
```
antoinerg: > We should add the traceType value inside these its so that if the scatter or the scattergl cases fail, we'll know which one did. For example,

I think this line allows to clearly identify which case fails. Am I mistaken?

cc @etpinard

antoinerg: Running `npm run test-jasmine -- config` will a failing test gives me:
screenshot_2019-01-25_12-35-18

What am I missing?

Thank you @etpinard

etpinard: `scatter` or `scattergl`.

antoinerg: Maybe it's not prominent enough but it's already there:
screenshot_2019-01-25_12-35-18

etpinard: Oops. It's in the `describe` block:

```
describe('responsive ' + traceType + ' trace', function() {
```

My bad.

antoinerg:
@etpinard Yes, I think I should clean up that CSS and fix this here.

src/plot_api/plot_api.js
```diff
@@ -105,7 +105,8 @@ exports.plot = function(gd, data, layout, config) {
 
     // hook class for plots main container (in case of plotly.js
     // this won't be #embedded-graph or .js-tab-contents)
-    d3.select(gd).classed('js-plotly-plot', true);
+    d3.select(gd).classed('js-plotly-plot', true)
+      .style('position', 'relative');
```
etpinard: Hmm. I'll call that a breaking change.

We could turn this on just when `responsive` is set to `true`. But then again, are there any other solutions?

antoinerg: Yes, that's certainly a breaking change and it breaks a lot of tests anyway. 

I have another branch in which I fix the responsive issues for both WebGL #3485 and the modebar https://github.com/plotly/plotly.js/issues/3499 that doesn't rely on modifying the parent. I will submit a new separate PR with a cleaner commit history and I will close the current one.

antoinerg:
Closing in favor of https://github.com/plotly/plotly.js/pull/3486

Issue #3037, opened 2018-09-21
Waterfall trace type


nicolaskruchten:
We should create a new type of trace for waterfall charts. These are very useful for understand how sums decompose/recompose into each other, see: http://support.zebrabi.com/contribution-analysis-waterfall-charts/

This would be distinct from bar charts in that the position of each bar depends on the position of the previous one, without the trace having to contain this math.

Highcharts has a neat way of dealing with the "computed remnants" by having a flag on certain values that say "take this bar down to zero":  https://www.highcharts.com/demo/waterfall

see also #1015

nicolaskruchten:
Multiple waterfall trace types on the same subplot would work like grouped bars, with each waterfall being independent, similar to:

![image](https://user-images.githubusercontent.com/203523/45898653-99f62180-bda8-11e8-8054-f981c634afbf.png)

nicolaskruchten:
I could see the API being just `x` and `y` with the magic `y` value of `NaN` causing the 'computed' case. We could then provide increasing/decreasing styles, like we do in financial traces, as well as computed styles, line styles etc. And the hoverlabels would indicate the delta, which we cannot currently do with offset bar traces.

nicolaskruchten:
@etpinard is my data proposal above reasonably clear or... ?

nicolaskruchten:
Essentially this models waterfalls a lot like lines, with special handling for `nan`, and different rendering/hovering

nicolaskruchten:
Unfortunately this kind of thing wouldn't quite be possible: we wouldn't be able to stack the green and orangy ones... they would be split:

![image](https://user-images.githubusercontent.com/203523/51040178-c5ce2a80-1584-11e9-9426-20e2a77eb80f.png)

etpinard:
> the magic `y` value of `NaN` causing the 'computed' case.

Hmm. That sounds a little bit inconsistent with our current API. Currently, `NaN`s are handled like other non-numeric "bad" values (e.g. `null` and `undefined`): they get skipped and don't show up on the graph. We did get a similar proposal in https://github.com/plotly/plotly.js/issues/2941, but I'm not sure I like it. Note that `NaN` is **not** JSON.

What if instead we had an array attribute with the same length as the coordinates e.g. 

```js
sums: ['', '', 'intermediate', '', '', 'total']
```

where items with `''` would show up as waterfalls.

nicolaskruchten:
[...]

PR #3488, opened 2019-01-28
Volume - new gl3d trace

archmoj:
Supersedes #3479 initial prototype.
New trace type for `volume` data visualisation.
Here is an [interactive demo](https://codepen.io/MojtabaSamimi/pen/OdNGMZ?editors=0010)
gl3d_volume_airflow
Note: now `opacityscale` attribute options could be applied to emphasize on maximum or minimum values as well as both `extremes`.


jackparmer:
Interesting view on isosurface vs volume plots by a lab at Lawrence Livermore:
image
https://wci.llnl.gov/simulation/computer-codes/visit/screenshots

lib/index.js
```diff
@@ -26,6 +26,7 @@ Plotly.register([
     require('./scatter3d'),
     require('./surface'),
     require('./isosurface'),
+    require('./volume4d'),
```
etpinard: `volume4d` interesting. Why do you prefer this over `volume`?

src/traces/isosurface/attributes.js
```diff
@@ -236,34 +235,20 @@ colorscaleAttrs('', {
 
     colorbar: colorbarAttrs,
 
-    // Flat shaded mode
-    flatshading: {
-        valType: 'boolean',
-        role: 'style',
+    opacityscale: {
+        valType: 'enumerated',
+        values: ['min', 'max', 'extremes', 'center'],
```
etpinard: Hmm. Could we also support 2D arrays similar to `colorscale` e.g.

```js
opacityscale: [[0, 0], [0.5, 0.2], [1, 0.9]]
```

?

src/traces/volume4d/convert.js
```diff
@@ -0,0 +1,132 @@
...
+        var rawId = selection.data.index;
+
+        var x = this.data._x[rawId];
+        var y = this.data._y[rawId];
+        var z = this.data._z[rawId];
+
+        var height = this.data._Ys.length;
+        var depth = this.data._Zs.length;
+
+        var i = findNearestOnAxis(x, this.data._Xs).id;
```
etpinard: ... and "nearest" here means nearest on the "top" slice, correct?

etpinard: ... which I think is _ok_ for the first iteration, but could be potentially improved in the future.

src/traces/isosurface/attributes.js
archmoj: Yes I think we could/should. However; from the user perspective it might be easier to select a `function` that could be linear or non-linear compared to declaring `linear` opacity values that way.

etpinard: Right, the two signatures would be supported. Users could either use 

```js
opacityscale: 'Center';

// or
opacityscale: [[0, 0], [0.5, 0.2], [1, 0.9]];
```

similar to how `colorscale` has special string value e.g. `Reds`, `Viridis` ...

lib/index.js
archmoj: I think `volume3d` & to some extends `volume` both refer to three-dimensions which this trace is not really about (i.e. exploring the data in 4 dimensions). If the value was declared as a `vector` instead of `scalar` (case of plotting wind directions & speed) the data could be 6d/7d.

src/traces/isosurface/attributes.js
archmoj: Good pointer. 
Working on this now...

src/traces/volume4d/convert.js
archmoj: Would you mind hovering on `gl3d_volume4d_airflow` in your browser?

jackparmer:
@archmoj do you have any more screenshots of the state of the art of this trace type so far?

archmoj:
> @archmoj do you have any more screenshots of the state of the art of this trace type so far?

@jackparmer here is another example using `opacityscale`.
plotly_volume_trace

etpinard:
TODO:

- [ ] make to add support for `hovertemplate`, once https://github.com/plotly/plotly.js/pull/3530 is merged.

src/components/opacityscale/attributes.js
```diff
@@ -0,0 +1,59 @@
...
+ *   - opacityscaleDflt {string}:
+ *     overrides the opacityscale dflt
+ *
+ *   - editTypeOverride {boolean} (dflt: ''):
+ *     most of these attributes already require a recalc, but the ones that do not
+ *     have editType *style* or *plot* unless you override (presumably with *calc*)
+ *
+ * @return {object}
+ */
+module.exports = function opacityScaleAttrs(context, opts) {
```
etpinard: We don't need a new `components/` module just for this thing. I can't imagine we'll need `opacityscale` for traces other than `volume` at any point.

Could you merge these new files into one and place it in  `traces/volume/opacityscale.js`.

src/lib/coerce.js
```diff
@@ -173,7 +174,22 @@ exports.valObjectMeta = {
         requiredOpts: [],
         otherOpts: ['dflt'],
         coerceFunction: function(v, propOut, dflt) {
-            propOut.set(scales.get(v, dflt));
+            propOut.set(colorscales.get(v, dflt));
+        }
+    },
+    opacityscale: {
```
etpinard: We don't a new `valType` just for `opacityscale`. In the `volume` attribute, you can set `valType` to `any` and coerce "manually" in the `volume/defaults.js`.

src/traces/isosurface/fn_find-nearest-on-axis.js
```diff
@@ -0,0 +1,29 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+var findNearestOnAxis = function(w, arr) {
```
etpinard: I would prefer if `findNearestOnAxis` and `generateIsoMeshes` were brought back in `isosurface/convert.js`.

To reuse them in `volume/convert.js`, you'll need to turn

```js
module.exports = createIsosurfaceTrace;
```

into 

```js
module.exports = {
  createIsosurfaceTrace: createIsosurfaceTrace,
  findNearestOnAxis: findNearestOnAxis,
  generateIsoMeshes: generateIsoMeshes
}
```

in `isosurface/convert.js` and then `require('../isosurface/convert')` in `volurme/convert.js`

src/traces/volume/attributes.js
```diff
@@ -0,0 +1,75 @@
...
+                'maximum iso-values. Often useful when either caps or',
+                'surfaces are disabled or filled with values less than 1.'
+            ].join(' ')
+        },
+        fill: {
+            valType: 'number',
+            role: 'style',
+            min: 0,
+            max: 1,
+            dflt: 1,
```
etpinard: Ha. I see the `spaceframe.fill` is different for `volume`. Nice touch!

src/traces/isosurface/attributes.js
```diff
@@ -225,9 +225,13 @@ var attrs = module.exports = overrideAll(extendFlat({
             'If trace `hoverinfo` contains a *text* flag and *hovertext* is not set,',
             'these elements will be seen in the hover labels.'
         ].join(' ')
-    },
+    }
 },
 
+opacityscaleAttrs('', {
```
etpinard: Hmm. Does this actual do anything? Do we really want to add `opacityscale` to `isosurface` traces?

test/image/compare_pixels_test.js
```diff
@@ -97,7 +97,9 @@ function untestableFilter(mockName) {
         mockName === 'font-wishlist' ||
         mockName.indexOf('gl2d_') !== -1 ||
         mockName.indexOf('mapbox_') !== -1 ||
-        mockName.indexOf('gl3d_cone-') !== -1
+        mockName.indexOf('gl3d_cone-') !== -1 ||
+        mockName.indexOf('gl3d_volume_between-ranges') !== -1 ||
```
etpinard: Ok. We should add those `gl3d_volume` mocks to

https://github.com/plotly/plotly.js/blob/02e51d8f5ba5601f53ee3dc839ba05fb12c7f8a2/tasks/noci_test.sh#L31-L39

also.

etpinard:
@archmoj what's the status on the PR? Are you still working on it? Should I pull down your branch to try it out?

archmoj:
> @archmoj what's the status of this PR? Are you still working on it? Should I pull down your branch to try it out?

@etpinard thanks for another great review. 
I would likely continue working on this branch after #3521 and more progress on the `waterfall` (#3037). 
If you wanted to interactively try it out that would be great too.

etpinard:
One more TODO:

- add support for `hovertext` - similar to https://github.com/plotly/plotly.js/pull/3553

archmoj:
> One more TODO:
> 
> * [ ]  add support for `hovertext` - similar to #3553

Good Call. Done in https://github.com/plotly/plotly.js/pull/3488/commits/320485f09a58e36dde6de446c32200477ade081f.

archmoj:
@etpinard this PR should be quite `review-able` now.

etpinard:
> this PR should be quite `review-able` now.

Nice. I'll do that at some point next week.

src/traces/volume/defaults.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return arr;
+}
+
+module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
+    function coerce(attr, dflt) {
+        return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
+    }
+
+    isosurfaceDefaults(traceIn, traceOut, defaultColor, layout);
```
etpinard: Hmm. I don't think we can do that. `isosurface` don't have similar attributes, but they have different `dflt` for `volume`, correct?

test/image/compare_pixels_test.js
```diff
@@ -106,8 +106,10 @@ else {
  *
  */
 function untestableFilter(mockName) {
-    var cond = !(
+    var cond =
+    !(
         mockName === 'font-wishlist' ||
+        mockName.indexOf('gl3d_volume_') !== -1 ||
```
etpinard: Hmm. That's less than ideal. Do you have an idea why we can't pixel-compare `volume` mocks?

Moreover, have you tried generating the `volume` baseline using `orca`?

src/traces/volume/defaults.js
archmoj: There is only one attribute `spaceframe` that has different `dflt`.

src/traces/volume/defaults.js
```diff
@@ -0,0 +1,70 @@
...
+    } else if(opacityscale === 'min') {
+        traceOut.opacityscale = [[0, 1], [1, MIN]];
+    } else if(opacityscale === 'extremes') {
+        traceOut.opacityscale = createWave(1, MIN);
+    } else if(!isValidScaleArray(opacityscale)) {
+        traceOut.opacityscale = undefined;
+    }
+};
+
+function isValidScaleArray(scl) {
```
etpinard: Very nicely done :1st_place_medal:

test/image/compare_pixels_test.js
archmoj: I think there is one mock that caused a problem. Now with @antoinerg parallel PR in place I could fix this. 
Second question: No I have not tried `volume` with `orca`.

src/traces/volume/defaults.js
etpinard: Ok, where is `spaceframe` getting coerced?

archmoj: Good catch!

test/image/compare_pixels_test.js
etpinard: > Second question: No I have not tried `volume` with `orca`.

Cool, could you give it a try?

etpinard: See

https://github.com/plotly/plotly.js/blob/c6573487f33ffe02f38f443690e5fdb824fdb161/tasks/noci_test.sh#L31-L38

for a sample orca command.

etpinard:
Looks like adding those volume tests to `npm run test-image` makes our CI run slower by about 1 minute:

image

but oh well, I'm ok with this :ok_hand: . We should try to keep to number of `volume` down to a minimum though.

etpinard:
@archmoj could you tell us a little bit more about how your `volume` renderer compare to MATLAB's https://www.mathworks.com/help/matlab/volume-visualization.html ? Thanks!

archmoj:
> Looks like adding those volume tests to `npm run test-image` makes our CI run slower by about 1 minute:
> 
> image
> 
> but oh well, I'm ok with this  . We should try to keep to number of `volume` down to a minimum though.

Good observation.

archmoj:
> Looks like adding those volume tests to `npm run test-image` makes our CI run slower by about 1 minute:

That was basically the result of increasing the waiting time between the batches. 
Now reset to normal in https://github.com/plotly/plotly.js/pull/3488/commits/0aa25f04fc6d7ce8b977dfeda2745b474e283955.

src/traces/volume/defaults.js
archmoj: Addressed in https://github.com/plotly/plotly.js/pull/3488/commits/77f7317b40f922cadca5b2aa6609732c99b8bbab.

src/traces/isosurface/iso_defaults.js
```diff
@@ -0,0 +1,100 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Registry = require('../../registry');
+var colorscaleDefaults = require('../../components/colorscale/defaults');
+
+module.exports = function supplyIsoDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
```
etpinard: Nicely done! But I would prefer having this `supplyIsoDefaults` in `isosurface/defaults.js`. 

We're trying to move away form this pattern where each defaults-related functions would get their own files. We now prefer having just one `defaults.js` file per trace module and have its reusable pieces exported along with the main trace `supplyDefaults.` function. See [`box/defaults.js`](https://github.com/plotly/plotly.js/blob/master/src/traces/box/defaults.js) for an example.

etpinard:
@archmoj can you add `volume` and `isosurface` to the gl3d partial bundle list: https://github.com/plotly/plotly.js/blob/master/lib/index-gl3d.js

archmoj:
> @archmoj can you add `volume` and `isosurface` to the gl3d partial bundle list: https://github.com/plotly/plotly.js/blob/master/lib/index-gl3d.js

Yes. Thanks for the link.

src/traces/isosurface/iso_defaults.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3488/commits/af17adc5a50ecae9d3d063241b2b94f8ef2e9bac.

etpinard:
This PR is looking good!

@archmoj two things:

- can you confirm that `orca` can export your `volume` traces ok?
- can you write down a few words about how your `volume` renderer compares to MATLAB's https://www.mathworks.com/help/matlab/volume-visualization.html ?

Thanks very much!

archmoj:
> * can you write down a few words about how your `volume` renderer compares to MATLAB's https://www.mathworks.com/help/matlab/volume-visualization.html ?

Thanks for the link. From that link I also found this interesting link:
https://www.mathworks.com/help/matlab/visualize/overview-of-volume-visualization.html

It looks that the way matlab does this require multiple steps to create the patch. Since our `isosurface` was capable of generating two or multiple surfaces between iso-values, and at the same time drawing iso-caps and slices those features are accessible to the user in `volume` trace too.
Thinking of visualisation of wind data in 3D (x, y, z, u, v, w), one could generate an example to display opaque `slices`, `caps` from `isosurface` and use `volume` on wind speed {`value = (u*u + v*v + z*z)^0.5`}, then add `cone` or `streamtube` plots to draw 3D wind direction vectors (u, v, w).
Some other 'nice-to-have' features (for future work) could be adding attributes to `isosurface` & `volume` for clipping/reducing volume data similar to https://www.mathworks.com/help/matlab/ref/subvolume.html and drawing `contour-slices` on the plot https://www.mathworks.com/help/matlab/ref/contourslice.html.

etpinard:
@archmoj the `gl3d_volume` can't render in time (i.e. before the timeout) on my machine, making `npm run test-image`. 

Do we have to test such large mocks in order to :lock: down the behaviour?

archmoj:
> @archmoj the `gl3d_volume` mocks can't render in time (i.e. before the timeout) on my machine, making `npm run test-image` fail 
> 
> Do we have to test such large mocks in order to  down the behaviour?

Should we rename the mocks?
@etpinard if you are still on the branch could you please check the `orca` export?

etpinard:
> Should we rename the mocks?

No, I think we should reduce the size of their data arrays.

etpinard:
> if you are still on the branch could you please check the `orca` export?

Are you having issue running orca?

archmoj:
> > if you are still on the branch could you please check the `orca` export?
> 
> Are you having issues running orca?

Yes - I needed to reinstall it. But then it didn't work : (

archmoj:
> > Should we rename the mocks?
> 
> No, I think we should reduce the size of their data arrays.

OK. I will take care of that.

archmoj:
@etpinard I was able to test `orca` export.
When possible please retry running the test-image with the reduced volume mocks on your machine.
Thank you very much in advance.

etpinard:
@archmoj commit https://github.com/plotly/plotly.js/pull/3488/commits/e23b2c39173f215438198bdcd5ad4d9395626237 is brilliant. Thanks very much for doing this.

I can confirm that the image tests are passing smoothly on my machine and `orca` exports are looking crisp!

Let's get this in :dancer: :dancer: :dancer:

PR #3491, opened 2019-01-28
replace prepend() calls since they aren't supported in IE11

### LINKED ISSUE #3489: Vertical modebars break plots on IE11
On IE11, setting the `layout` parameter `modebar: {'orientation': 'v'}` will break plots on IE11, with the following error:
```javascript
Unhandled promise rejection TypeError: Object doesn't support property or method 'prepend'
```

This is likely related to [this](https://github.com/plotly/plotly.js/pull/3068/files#diff-979486c8ea6cdf2d004e24d69d537f3cR85) line in #3068  using `element.prepend()`. The bug can likely be easily fixed by using [this](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend#Polyfill) polyfill suggested by MDN.

"Plunker" pen demonstrating the issue (graph does not display on IE11):
https://plnkr.co/edit/mnQV4RWKhAQOLUX225IL?p=preview

image 5

I had to use Plunker as CodePen et al is not compatible with IE11.

antoinerg:
Fixes #3489

### DIFF
--- a/src/components/modebar/modebar.js
+++ b/src/components/modebar/modebar.js
@@ -88,7 +88,9 @@ proto.update = function(graphInfo, buttons) {
             }
 
             if(fullLayout.modebar.orientation === 'v') {
-                this.element.prepend(logoGroup);
+                d3.select(this.element).insert(function() {
+                    return logoGroup;
+                }, ':first-child');
             } else {
                 this.element.appendChild(logoGroup);
             }
--- a/src/components/modebar/modebar.js
+++ b/src/components/modebar/modebar.js
@@ -88,9 +88,7 @@ proto.update = function(graphInfo, buttons) {
             }
 
             if(fullLayout.modebar.orientation === 'v') {
-                d3.select(this.element).insert(function() {
-                    return logoGroup;
-                }, ':first-child');
+                this.element.insertBefore(logoGroup, this.element.childNodes[0]);
             } else {
                 this.element.appendChild(logoGroup);
             }
--- a/test/jasmine/assets/unpolyfill.js
+++ b/test/jasmine/assets/unpolyfill.js
@@ -18,5 +18,17 @@
                 ].join(' '));
             }
         });
+
+        Object.defineProperty(item, 'prepend', {
+            configurable: true,
+            enumerable: true,
+            writable: true,
+            value: function remove() {
+                throw Error([
+                    'test/jasmine/assets/unpolyfill.js error: calling ChildNode.prepend()',
+                    'which is not available in IE.'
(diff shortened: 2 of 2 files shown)


antoinerg:
@etpinard The failing example provided by @wbrgss was tested to work under IE11 on Windows 10 by using the build from the current branch: https://plnkr.co/edit/SDQf3lmpKgltoTEoNYW8?p=preview

src/components/modebar/modebar.js
```diff
@@ -88,7 +88,9 @@ proto.update = function(graphInfo, buttons) {
             }
 
             if(fullLayout.modebar.orientation === 'v') {
-                this.element.prepend(logoGroup);
+                d3.select(this.element).insert(function() {
```
etpinard: Hmm. Does

```js
d3.select(this.element).insert(logoGroup, ':first-child');
```

work also?

antoinerg: I tried that initially but it does not work:
screenshot_2019-01-28_17-07-49

etpinard: Ha right, does

```js
d3.select(this.element).insert(d3.select(logoGroup), ':first-child');
```

work?

antoinerg: > d3.select(this.element).insert(d3.select(logoGroup), ':first-child');

Also fails with: 
screenshot_2019-01-28_17-18-23

Would you prefer that I to turn the snippet from this PR into a one-liner?

etpinard: Cool. Thanks for checking!

etpinard:
Nicely done :dancer: !

src/components/modebar/modebar.js
alexcjohnson: Non- 🚫, but at that point the vanilla solution seems simpler:

https://github.com/plotly/plotly.js/blob/ea7941b97fbaf11967f3c118a8e130036c8169f4/src/plot_api/subroutines.js#L148

https://github.com/plotly/plotly.js/blob/ea7941b97fbaf11967f3c118a8e130036c8169f4/src/lib/svg_text_utils.js#L105-L106

antoinerg: Thanks @alexcjohnson! I agree it is simpler. Done in commit 4b164a26506e60f29e854600c76042cc9b94c53d

Updated demo https://plnkr.co/edit/3Rfo4svtr6PUDCiCRM3l?p=preview

etpinard:
:dancer: (bis). Nice job!

PR #3498, opened 2019-01-30
Displaying axis labels other than x, y & z in the hovering popup using hovertitle attribute

### LINKED ISSUE #2618: Axis title could also change the labels in the hovering popup

archmoj:
Supersedes #3317.
Fix #2618 and #265.
A new attribute `hovertitle` could be applied with each axis to enable this feature.
For example in 3D scenes if the x, y, z dimensions are presenting latitude, longitude and elevation (or other coordinates e.g. polar) this info could be displayed in the hovering popup instead of x, y and z.
@plotly/plotly_js


nicolaskruchten:
I wonder if we shouldn't just prioritize `hovertemplate` instead and then they can do this directly?

antoinerg:
@nicolaskruchten I wonder the same thing since it should be possible to do anything with `hovertemplate` already :thinking:

src/components/fx/hover.js
```diff
@@ -931,10 +931,24 @@ function createHoverText(hoverData, opts, gd) {
...
+            if(gd._fullLayout.yaxis) yLetter = gd._fullLayout.yaxis.hovertitle || yLetter;
+        }
+
         if(d.zLabel !== undefined) {
-            if(d.xLabel !== undefined) text += 'x: ' + d.xLabel + '<br>';
-            if(d.yLabel !== undefined) text += 'y: ' + d.yLabel + '<br>';
-            text += (text ? 'z: ' : '') + d.zLabel;
+            if(d.xLabel !== undefined) text += xLetter + ': ' + d.xLabel + '<br>';
+            if(d.yLabel !== undefined) text += yLetter + ': ' + d.yLabel + '<br>';
+            text += (text ? zLetter + ': ' : '') + d.zLabel;
```
etpinard: Hmm. So `hovertitle` for cartesian axes only has an effect for trace with a "z" field. That's ok, but should make that clear in the attribute description.

src/plots/polar/layout_attributes.js
```diff
@@ -118,6 +118,7 @@ var radialAxisAttrs = {
     // might need a 'titleside' and even 'titledirection' down the road
 
     hoverformat: axesAttrs.hoverformat,
+    hovertitle: axesAttrs.hovertitle,
```
etpinard: We could set a "hard" `dflt value: `'r'` for `radialaxis.hovertitle` and `'θ'` for `angularaxis.hovertitle`

test/image/mocks/gl3d_ibm-plot.json
```diff
@@ -304,6 +304,7 @@
         "scene": {
             "xaxis": {
                 "title": "Term",
+                "hovertitle": "Term",
```
etpinard: Adding `hovertitle` to a few mock is ok, but we should really test this issue jasmine tests.

- https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/hover_label_test.js for cartesian 
- https://github.com/plotly/plotly.js/blob/884d2fa4cf76529bd6cb819fa2d4fec1f43fadfa/test/jasmine/tests/ternary_test.js#L120-L174 for ternary 
- https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/gl3d_plot_interact_test.js#L107 for gl3d
- https://github.com/plotly/plotly.js/blob/884d2fa4cf76529bd6cb819fa2d4fec1f43fadfa/test/jasmine/tests/scatterpolar_test.js#L83-L200 for polar
- https://github.com/plotly/plotly.js/blob/884d2fa4cf76529bd6cb819fa2d4fec1f43fadfa/test/jasmine/tests/carpet_test.js#L633-L681 for carpet

----

Oh and we should also `hovertitle` to `geo.lonaxis` and `geo.lataxis` as displayed on https://codepen.io/etpinard/pen/gqKjeB

archmoj:
Closing now that `hovertemplate` is supported for 3D traces as well.

Issue #3318, opened 2018-12-11
Sankey: colorscales per component, linked to concentration


antoinerg:
A single flow can be made up of multiple components (this is already allowed - multiple connections between the same pair of nodes). Often those components exist across the whole diagram... so for example you're tracking oil and water as they move through some process. We need a way to 1) label links by component, 2) color all links of the same component, either with a constant per-component color or a per-component colorscale that maps to its concentration within the flow. This is particularly important for trace components: perhaps component benzene goes from grey at zero to red at 10ppm

antoinerg:
Each link has a concentration at the input AND the output of a node as shown in branch `sankey-link-concentration`:
https://github.com/plotly/plotly.js/blob/2ffa5634852396835c67cec62c553888e5c32f6a/src/traces/sankey/render.js#L71-L87

Therefore, I think we should support gradients as fill color for links to visualize the concentration at both ends. The start and end colors of this gradient should be either
- [ ] The colors of the source and target nodes (to close https://github.com/plotly/plotly.js/issues/3261)
- [ ] A single color for each link specified by the already existing attribute `link.color`
- [ ] The color of the link based on its label and a colorway
- [ ] A colorscale for each link that map to its concentration

alexcjohnson:
Oh I was thinking of concentration not as a fraction of the total of flows into or out of a node, but as a fraction of total parallel flows - ie all links with the same source AND target nodes, or from a physical standpoint the concentration within a specific pipe. In that case it's constant along the link.

Also note that the value of a single link as a fraction of total inflow/outflow is not necessarily the concentration of that component coming in or out, as there may be multiple paths containing that component.

antoinerg:
> Oh I was thinking of concentration not as a fraction of the total of flows into or out of a node, but as a fraction of total parallel flows - ie all links with the same source AND target nodes, or from a physical standpoint the concentration within a specific pipe. In that case it's constant along the link.

Ok so we should also compute the fraction of the link versus the total flow connecting the same source and target nodes.

> Also note that the value of a single link as a fraction of total inflow/outflow is not necessarily the concentration of that component coming in or out, as there may be multiple paths containing that component.

Nice observation. Basically, the component concentration will be given by summing the aforementionned ratios for links with the same label.

@alexcjohnson Should we support gradients nonetheless? How do you feel about feature request https://github.com/plotly/plotly.js/issues/3261

etpinard:
> Should we support gradients nonetheless?

[...]

PR #3501, opened 2019-01-31
Sankey: colorscales per component, linked to concentration

antoinerg:
An attempt at closing #3318


antoinerg:
Here's a Codepen showcasing the new feature introduced in this PR: https://codepen.io/antoinerg/pen/qgJbOr

By default in Sankey, when we hover on a link, its opacity is reduced. I think this looks odd in the Codepen above.

@alexcjohnson what do you think?

src/traces/sankey/render.js
```diff
@@ -10,6 +10,7 @@
 
 var c = require('./constants');
 var d3 = require('d3');
+var sum = require('d3-array').sum;
```
etpinard: Ha this here only works between `d3-sankey` already depends on `d3-array`.

image

We should add `d3-array` to the (plotly.js) dependencies OR use plain loops and save one line in our `package.json`.

antoinerg: I think I will do the latter and save one line in our `package.json`.

I should probably replace all the `forEach` while I'm at it right?

alexcjohnson:
> Hovering on links will display the new computed statistics.

Nice - perhaps we can include the concentration in the label by default, for links that have a concentration colorscale?

> By default in Sankey, when we hover on a link, its opacity is reduced. I think this looks odd in the Codepen above.

That's weird, normally opacity is *increased* on hover - is that because normally the opacity is low but here its high, and on hover it gets some set value that's more than the default but less than we have here?

Anyway it seems like we probably don't want an opacity change at all when the color is displaying quantitative information via a colorscale - especially since scales like you've used, with white at one end, will probably be common, and these scales are equivalent to opacity. But nor do we want to *decrease* opacity on hover even for non-colorscaled links. Perhaps we just need to find a different way to highlight the hovered links if their opacity is above some threshold and/or they have a colorscale. Perhaps adding, thickening, or increasing the contrast of the link outline?

By the same token, we probably want the hover label to be fully opaque when it's describing a colorscaled link, so it will have the same color as the link.

src/traces/sankey/render.js
etpinard: > I should probably replace all the `forEach` while I'm at it right?

That would be amazing :muscle:

antoinerg: Done in c228b37

antoinerg:
> Nice - perhaps we can include the concentration in the label by default, for links that have a concentration colorscale?

Done in c228b37

> That's weird, normally opacity is _increased_ on hover - is that because normally the opacity is low but here its high, and on hover it gets some set value that's more than the default but less than we have here?

Yes you are right: it used to simply set the opacity of 0.4 which is higher than the default but less than we had here.

> Perhaps we just need to find a different way to highlight the hovered links if their opacity is above some threshold and/or they have a colorscale. Perhaps adding, thickening, or increasing the contrast of the link outline?

In c228b37, I simply removed hover effects for links. @etpinard Do you have any suggestion of what to do to highlight the hovered links? Is there any precedent in `plotly.js` for such a situation? Displaying a link outline seems like a good idea but I wanted  to double-check with you first.

> By the same token, we probably want the hover label to be fully opaque when it's describing a colorscaled link, so it will have the same color as the link.

Done in c228b37

etpinard:
> @etpinard Do you have any suggestion of what to do to highlight the hovered links? Is there any precedent in `plotly.js` for such a situation?

Do you really "need" to highlight them?

peek 2019-02-15 14-58

their positioning and coloring makes the hover labels pretty unambiguous to me, at least in this case.

antoinerg:
> Do you really "need" to highlight them?

@etpinard I was wondering the exact same thing. I'd be curious to have @alexcjohnson's opinion on this one!

etpinard:
> I'd be curious to have @alexcjohnson's opinion on this one!

Good call. I'm voting _no_.

alexcjohnson:
That's a pretty simple diagram, but in more complicated ones like the energy flows example a highlight can really help to see clearly what nodes the link connects.

I'd be OK with omitting highlighting for now when the link is opaque, but we should keep it in the default case and reserve the right to add something like outlining later on.

antoinerg:
@etpinard It seems like @alexcjohnson is OK with not highlighting colorscaled links (at least for now). Nothing precludes us from changing their appearance later on so I'm comfortable with the status quo.

I agree that the existing appearance of hovered non-colorscaled links is useful and I confirm that it is still  intact.

As of right now, I'm not testing in jasmine that the flow statistics are properly computed although the mock effectively :lock: this down. Let me know if that is sufficient.

package-lock.json
```diff
@@ -4142,8 +4142,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
```
etpinard: @antoinerg which version of `npm` are you using?

Your PR doesn't change any of our dependencies, so it shouldn't change the `package-lock`. Can you revert these changes?

src/traces/sankey/render.js
```diff
@@ -616,6 +686,7 @@ module.exports = function(svg, calcData, layout, callbacks) {
           .attr('d', linkPath())
           .call(attachPointerEvents, sankey, callbacks.linkEvents);
 
+
```
etpinard: :zap: no need to commit this newline.

etpinard:
Nicely done @antoinerg !

I made two minor comments. This thing is set to be part of 1.45.0!

src/traces/sankey/render.js
antoinerg: Done in f628b549433df3f86244125a2e66929781929f69

package-lock.json
antoinerg: I now updated to npm `6.8.0`, ran `npm i ` and the `package-lock.json` has now zero changes! Sorry about that... I need a safety check on my dev machine.

Done f628b549433df3f86244125a2e66929781929f69

etpinard:
Nicely done. :dancer:

Issue #1549, opened 2017-04-03
Matching axes


alexcjohnson:
We should allow two (or more) axes to not just share a scaling, but actually force precisely equal `range`. Use case is plots like [this one](https://plot.ly/~jackp/16209/machine-learning-classifier-comparison.embed):
![machine-learning-classifier-comparison](https://cloud.githubusercontent.com/assets/2678795/23995287/43807cd0-0a1f-11e7-965d-e6b4f30cbb63.png)

This came up in https://github.com/plotly/plotly.js/issues/272#issuecomment-287043579 but was not addressed in #1522 - opening this as a new issue so we don't forget about it.

tishihar94:
Hi, any progress with this? 

I have two side-by-side scatter plots that are identical, except for the colors of each point (One plot is colored by cluster and the other has a colorscale). It would be great if I could synchronize the zoom/pan so that both the X and Y axis are preserved But I can only get it to synchronize the Y-axis even if I do `shared_xaxes=True` and `shared_yaxes=True`.

alexcjohnson:
We're probably going to need to do this as part of #2372, which is up next, so that the x and y for the same variable can have matched ranges (keeping the diagonal panes diagonal as well). But when we do this it will include matched x and matched y axes as well as x matching y.

mschilli87:
@alexcjohnson: https://github.com/plotly/plotly.js/issues/2372 was closed via https://github.com/plotly/plotly.js/pull/2505 but this is still open. So is this possible by now and if so how? I have to scatter plots below each other via `subplot` and I'd like to have the same axes range on both and zoom synchronously. Am I on the right track?

alexcjohnson:
@mschilli87 I'm afraid matching axes was not included in the initial SPLOM implementation and has not yet been done on its own, so it's correct for this issue to remain open. In the meantime you may be able to listen to `plotly_relayout` events and issue new `Plotly.relayout` calls to keep the axes in sync - but be careful about avoiding infinite loops!

nicolaskruchten:
A few items that come to mind for ideal 'matched' axes:

* category order should match
* histogram auto-binning should align

alexcjohnson:
Absolutely, autorange should take into account all data on any of the matched axes. This is going to be a bit more complicated than just throwing all of that data into the autorange calculation though, because two axes should be able to match `range` *without* matching `_length` - particularly useful for x matching y without forcing the plot to be square.

[...]

PR #3506, opened 2019-02-01
Implementing matching axes

etpinard:
resolves https://github.com/plotly/plotly.js/issues/1549 - with a :sparkles: **new** :sparkles: cartesian axis attribute `ax.matches` :exclamation: 

-------------

### Demos:

Matching `x`, `x2` and `x3` axes:

peek 2019-02-01 17-26

Matching splom axes (notice the constrained zoombox on the diagonal):

peek 2019-02-01 17-28

Or funkier things, like matching an overlaying axis:

peek 2019-02-01 17-27

-----

### Notable features:

- matching axes get matching autorange values https://github.com/plotly/plotly.js/commit/6f2bb788a866012e8f99bd9bc26079a50ae8a846, https://github.com/plotly/plotly.js/commit/c5f9e74e1df0b4f2c3a3248e64e71d22916711ea
- matching axes get matching categories https://github.com/plotly/plotly.js/commit/d9c2d4e2cca5b3abeba8c4702ccd6d8b6b5ac806
- matching axes get matching autobins https://github.com/plotly/plotly.js/commit/91185051bfcce07edd5853c1bc4a894ecfa21031
- a new `splom` dimension attribute (`axis.matches`) to turn on matching axes behavior for splom-generated axes https://github.com/plotly/plotly.js/commit/c290adfa26f47da57c731854315ff2e45ff0792a
- guaranteed fun when dragging around subplots with matching axes

----

More technically, I attempted to reuse most of the `scaleanchor` logic, for better and worse as some things didn't match (pun :smirk: ) too well. I also cleaned up a few things in the `cartesian_interact_test.js` suite (in https://github.com/plotly/plotly.js/commit/13007e029507d89410a222f635f3f9726e1fbb27, https://github.com/plotly/plotly.js/commit/e962e96ac31cf8329026e701399efef52fc3793f https://github.com/plotly/plotly.js/commit/ca1de5db42324c813caa0d03b56d25d4a451310f)

I'll tag this thing as `reviewable`, but I'll hunt around for edge cases again on Monday. I'll not 100% convinced the `scaleanchor` + `matches` mock (added in https://github.com/plotly/plotly.js/commit/8c09944acfe016dfd3ea0880e734f37e8dc18650) does the right thing (**update:** won't be done in this PR). Moreover, I'll try to incorporate https://github.com/plotly/plotly.js/pull/3460 into this branch (**done in:** https://github.com/plotly/plotly.js/pull/3506/commits/99a9edb4f968be0eeba19dbfd42741c1c3576bd5)

cc @plotly/plotly_js


src/traces/histogram/cross_trace_defaults.js
```diff
@@ -50,15 +50,16 @@ module.exports = function crossTraceDefaults(fullData, fullLayout) {
         binDirection = traceOut.orientation === 'v' ? 'x' : 'y';
         // in overlay mode make a separate group for each trace
         // otherwise collect all traces of the same subplot & orientation
-        group = isOverlay ? traceOut.uid : (traceOut.xaxis + traceOut.yaxis + binDirection);
-        traceOut._groupName = group;
-
+        group = traceOut._groupName = isOverlay ? traceOut.uid : (
+            getAxisGroup(fullLayout, traceOut.xaxis) +
+            getAxisGroup(fullLayout, traceOut.yaxis) +
```
etpinard: **N.B.** currently, both x and y need to match in order to be considered in the same auto-bin group.

etpinard:
Remaking https://plot.ly/~jackp/16209/machine-learning-classifier-comparison.embed with matching axes gives :arrow_right:  https://codepen.io/etpinard/pen/NojNwY

alexcjohnson:
This is awesome @etpinard, so much fun to play with!

> I'll not 100% convinced the `scaleanchor` + `matches` mock (added in [8c09944](https://github.com/plotly/plotly.js/commit/8c09944acfe016dfd3ea0880e734f37e8dc18650)) does the right thing.

yeah, that one has some problems... looks like after a zoom/pan it manages to satisfy all the constraints (the very shrunken subplots I think are correct), but initially or after an autorange it does not.
axes_scaleanchor-with-matches

Here's a more extreme example. Kind of a pathological case, where each subplot has x&y matching, and the x of each subplot is supposed to scale with the y of the previous... this is the case I think compounds height/width differences, so it should be expected that either the first or last subplot gets its data very compressed if you stretch it to a very non-square size. Right now it appears to apply all the `scaleanchor` constraints, but all the `matches` constraints except the last are lost.
screen shot 2019-02-02 at 8 22 00 am
```js
Plotly.newPlot(gd,[
  {y:[1,2]},
  {y:[1,2],xaxis:'x2',yaxis:'y2'},
  {y:[1,2],xaxis:'x3',yaxis:'y3'},
  {y:[1,2],xaxis:'x4',yaxis:'y4'}
], {
  yaxis: {matches: 'x'},
  xaxis2: {scaleanchor: 'y'},
  yaxis2: {matches: 'x2'},
  xaxis3: {scaleanchor: 'y2'},
  yaxis3: {matches: 'x3'},
  xaxis4: {scaleanchor: 'y3'},
  yaxis4: {matches: 'x4'},
  grid: {rows: 2, columns: 2, pattern: 'independent'},
  width: 500,
  height: 400
})
```

I haven't looked at your code in detail yet, but in order to satisfy this kind of chained constraints seems to me it may be necessary to generate `scaleanchor` constraints based on each `matches` constraint (using `gs._w` and `gs._h`) and use that to run through the `scaleanchor` algorithm, and only after that apply the `matches` constraints explicitly.

etpinard:
TODO:

- [ ] match `histogram2d` binning of matching axes

etpinard:
@alexcjohnson here's an attempt at satisfying both `matches` and `scaleanchor`: https://github.com/plotly/plotly.js/commit/74cf76c394f157a2c8d62c42ce9c43b6d949e101

In that commit, matching `autorange:true` axes first have their range value computed **then** we enforce the `scaleanchor` constraint as if their range was set. This way, only their `domain` is tweaked during `enforceAxisConstraints`. I'm not 100% sure this is what users expect, but it seems to make things work ok.

Now, about your 

```js
Plotly.newPlot(gd,[
  {y:[1,2]},
  {y:[1,2],xaxis:'x2',yaxis:'y2'},
  {y:[1,2],xaxis:'x3',yaxis:'y3'},
  {y:[1,2],xaxis:'x4',yaxis:'y4'}
], {
  yaxis: {matches: 'x'},
  xaxis2: {scaleanchor: 'y'},
  yaxis2: {matches: 'x2'},
  xaxis3: {scaleanchor: 'y2'},
  yaxis3: {matches: 'x3'},
  xaxis4: {scaleanchor: 'y3'},
  yaxis4: {matches: 'x4'},
  grid: {rows: 2, columns: 2, pattern: 'independent'},
  width: 500,
  height: 400
})
```

in https://github.com/plotly/plotly.js/pull/3506/commits/9f2fad133f7f11e795076317701372f9ab3e22af, we disallowed `matches` and `scaleanchor` constraints with different target axes. Perhaps I misinterpreted 

> I guess it may be clearer that way, as it means “range matches AND scale matches”... just seems weird as the only allowed values of `scaleanchor` would be `’x’` or nothing.

from your https://github.com/plotly/plotly.js/issues/1549#issuecomment-458323206

etpinard:
@nicolaskruchten we don't currently match auto-binning for `histogram2d` traces on the same subplot. At present, I think it would feel weird to add a way to make them match across subplots, but have no way to make them match on the same subplot. I'm thinking we need another `histogram2d` attribute (e.g. `bingroup`). Could we split this item into another issue?

nicolaskruchten:
> Could we split this item into separate issue?

Yep!

etpinard:
@nicolaskruchten @alexcjohnson would it a big deal if we disallowed simultaneous use of `matches` and `scaleanchor` on the set of axes for this first iteration?

nicolaskruchten:
In principle that's fine... I would assume that `matched` totally supersedes `scaleanchor` anyway, no?

alexcjohnson:
In principle you can have both, if you have `constrain: 'domain'` too. Anyway, I'm fine with omitting it from v1 as long as it's clear we can add it later. Let's have `matches` win and drop `scaleanchor` when necessary until they're supported together.

etpinard:
Great. Thanks! Sorry for quitting on this problem, but I'll like to get going on `sunburst` ASAP.

etpinard:
> . Let's have `matches` win and drop `scaleanchor` when necessary until they're supported together.

Done in https://github.com/plotly/plotly.js/pull/3506/commits/91431ec425052d0a5ee90182b08adafc6c53cd15

@plotly/plotly_js this PR is now ready to review :eyes: As this PR touches some fairly important cartesian subplot concepts, I'll like both @antoinerg and @archmoj to review this thing. Thanks!

nicolaskruchten:
Just to make sure I get the scaleanchor limitations... it means it’s not possible to have two plots with matched axes that are both constrained to be physically square?

etpinard:
> it means it’s not possible to have two plots with matched axes that are both constrained to be physically square?

that's correct.

nicolaskruchten:
> that's correct. (sorry 😢 )

I think that's totally reasonable to start with!

etpinard:
Ping @archmoj @antoinerg can one of you please review this thing today? Thank you.

src/plot_api/subroutines.js
```diff
@@ -737,7 +737,7 @@ exports.doAutoRangeAndConstraints = function(gd) {
             ax = Axes.getFromId(gd, id);
             ax.range = rng.slice();
             ax._input.range = rng.slice();
-            ax.setScale(0);
+            ax.setScale();
```
archmoj: Good catch!

src/plots/cartesian/layout_attributes.js
```diff
@@ -211,6 +213,23 @@ module.exports = {
...
+        ],
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'If set to another axis id (e.g. `x2`, `y`), the range of this axis',
+            'will match the range of the corresponding axis in data-coordinates space.',
+            'Moreover, matching axes share auto-range values, category lists and',
+            'histogram auto-bins.',
+            'Note that setting axes simultaneously in both a `scaleanchor` and a `matches` constraint',
+            'is currently forbidden.'
```
antoinerg: Is it obvious that for axes to match, they need to have the same type? If it isn't, maybe we should mention it here. On a related note, should we fire `lib.warn` when that happens?

etpinard: > Is it obvious that for axes to match, they need to have the same type?

I'll add it to the description. Thanks!




> . On a related note, should we fire `lib.warn` when that happens?

It should fall into this 

https://github.com/plotly/plotly.js/blob/91431ec425052d0a5ee90182b08adafc6c53cd15/src/plots/cartesian/constraints.js#L70-L72

`Lib.warn` call.

src/plots/cartesian/dragbox.js
```diff
@@ -468,6 +482,7 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
             for(i = 0; i < xaxes.length; i++) {
                 zoomWheelOneAxis(xaxes[i], xfrac, zoom);
             }
+            updateMatchedAxes(matches.isSubplotConstrained ? yaxes : matches.xaxes, xaxes[0].range);
```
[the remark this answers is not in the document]
etpinard: No. They come from

https://github.com/plotly/plotly.js/blob/91431ec425052d0a5ee90182b08adafc6c53cd15/src/plots/cartesian/dragbox.js#L93-L94

which are always defined.

src/plots/cartesian/set_convert.js
```diff
@@ -307,10 +307,25 @@ module.exports = function setConvert(ax, fullLayout) {
             var traceIndices = ax._traceIndices;
             var i, j;
 
+            var matchGroups = fullLayout._axisMatchGroups;
+            if(matchGroups && matchGroups.length && ax._categories.length === 0) {
```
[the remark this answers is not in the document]
etpinard: If `ax._categories` isn't defined here, then I would prefer this to fail. `ax.setupMultiCategory` should always be called after `ax.clearCalc` where `ax._categories` is initialized.

src/plots/cartesian/dragbox.js
```diff
@@ -922,14 +969,22 @@ function zoomAxRanges(axList, r0Fraction, r1Fraction, updates, linkedAxes) {
     }
 
     // zoom linked axes about their centers
-    if(linkedAxes && linkedAxes.length) {
+    if(linkedAxes.length) {
```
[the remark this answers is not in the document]
etpinard: Ok. Done -> https://github.com/plotly/plotly.js/pull/3506/commits/70b10fff46eef9b7cd601317764ed9e3e4515324

src/plots/cartesian/layout_attributes.js
etpinard: > I'll add it to the description.

-> https://github.com/plotly/plotly.js/pull/3506/commits/772efe519828e6efdb436e4dfbb8a6275b230bdf

archmoj:
@antoinerg it is a :dancer: from my side, if you have no more comments?

antoinerg:
It is also a :dancer: from me! 

@etpinard eventually I'd be curious to know what hurdles you ran into and how long or hard it woud be to support constrained axes.

Anyway, nice PR :muscle:

etpinard:
> I'd be curious to know what hurdles you ran into and how long or hard it woud be to support constrained axes.

That's in -> https://github.com/plotly/plotly.js/issues/3539

Issue #2985, opened 2018-09-07
Automargin clipping ticklabels


sglyon:
I'm trying to use `*axis.automargin` to get the _minimum_ necessary margin to display all tick labels/axis titles.

To do that I set `layout.margin.[tblr]` to 0 and then set `layout.[xy]axis.automargin` to true.

When I do this and do not have axis titles, the tick labels are slightly clipped. See this pen for an example: https://codepen.io/anon/pen/XPzEeN

When in the pen if you uncomment the line that sets `layout.yaxis.title = ""`, you'll notice that the ticks are no longer clipped.

sglyon:
Update, if you do have a yaxis title that is long enough to run into ticks, it will cover them up.

I've updated the pen with another section that has

```js
  // Uncomment the line below and the yaxis.title will cover up some ticks
  // layout.yaxis.title = "This is y and it runs into ticks"
```

and if you uncomment the long title you can see what I mean.

It is very possible that I'm simply trying to use this in the wrong way. If there is a better/easier way to reach my goal (minimum margin where nothing is clipped), I'd love to hear it

etpinard:
Related: https://github.com/plotly/plotly.js/issues/2434

zbjornson:
Here's a particularly egregious example of I think the same issue: https://codepen.io/anon/pen/XOWWNa

cldougl:
One of the goals of ddk is to provide reduced margins by default so we're running into this issue in most of our examples as the margin size is templated as in the example below: 
https://codepen.io/plotly/pen/bzBOjL?editors=1010

etpinard:
Would setting a margin of 1px (together with `ax.automargin:true`) like in https://codepen.io/etpinard/pen/omBbBv

wbrgss:
@etpinard Thanks for taking a look. I don't think this will work for us though; it seems any `margin` smaller than the tick size will mean clipped ticks — see the x-axis ticks on
https://codepen.io/anon/pen/GzrjxN

I've tried the `layout.yaxis.title = ""` trick, which appears to work well for most graphs, but that's not going to work for us with components that make use of plots flush with their container, such as the following (sorry for the lack of codepen)
![datacard](https://user-images.githubusercontent.com/9068290/52030685-fa2f6980-24e6-11e9-9085-fd50a8d52611.png)

If `layout.yaxis.title = ""` hack really is the best solution I think I'll do my best to work around the above issue with the flush-with-border plots.

[...]

PR #3510, opened 2019-02-04
Fix axis automargin push offset

etpinard:
fixes https://github.com/plotly/plotly.js/issues/2985 

I attempted to fix this bug by addressing https://github.com/plotly/plotly.js/issues/1988 and by association https://github.com/plotly/plotly.js/issues/2434, but that turned out to be too much of an endeavour. The DDK folks seems to really not like #2985, so here's a quick(er) fix.

In brief, we were previously assuming that the ticks + tick labels started right on the edge of the axis line when computing the auto-margin push value. This is incorrect, there's a slight pad between tick labels and ticks that must be taken into account to get the correct push.

See updated bug report codepens:

- https://codepen.io/anon/pen/XPzEeN -> https://codepen.io/etpinard/pen/GzvjoX
- https://codepen.io/anon/pen/GzrjxN -> https://codepen.io/etpinard/pen/gqxwPB

cc @plotly/plotly_js


antoinerg:
Very nice PR @etpinard ! Thanks for fixing this issue :dancer:

etpinard:
Thanks for the review @antoinerg !

It might be nice to have some of the DDK folks try out 

https://25536-45646037-gh.circle-artifacts.com/0/plotly.min.js

before merging this cc @cldougl @wbrgss

wbrgss:
> This is incorrect, there's a slight pad between tick labels and ticks that must be taken into account to get the correct push.

Yeah, that's what it looked like to me as well. I've tested the above `plotly.min.js` with DDK, and can confirm this PR appears to fix #2985. This will be a great help; thank you @etpinard!

etpinard:
Thanks @wbrgss for checking! Merging :rocket: 

I'll release 1.44.3 tomorrow with this patch included.

Issue #3459, opened 2019-01-22
Can't show two scattercarpet traces on two different carpets


etpinard:
See https://codepen.io/etpinard/pen/bzGyvm 

Original post: https://github.com/plotly/plotly.py/issues/1398

PR #3512, opened 2019-02-04
Clip scattercarpet traces with clipPath of ref'ed carpet

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3459 

cc @plotly/plotly_js


alexcjohnson:
Ha right, all our other examples have each carpet on its own subplot...
💃

Issue #3455, opened 2019-01-22
Bad placement of aaxis LaTeX title in a scatterternary plot #1399


jonmmease:
Originally reported in https://github.com/plotly/plotly.py/issues/1399 by @empet
> 

CodePen: https://codepen.io/jonmmease/pen/pGoLNg?editors=1010

![screen shot 2019-01-22 at 6 27 11 am](https://user-images.githubusercontent.com/15064365/51532691-ca110800-1e0e-11e9-9519-072a5a8f94d8.png)

This problem is present in Chrome, so it's not a Firefox only issue like some of the other LaTeX problems.

PR #3513, opened 2019-02-04
Fix mathjax placement in ternary aaxis titles

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3455 -- cc @plotly/plotly_js 

I was really hoping to get rid of this hacky block 

https://github.com/plotly/plotly.js/blob/84e24aaf529000ec14fbe64aeeb4bcdc6ccac474/src/lib/svg_text_utils.js#L127-L142

entirely, but since our MathJax test coverage is lackluster, I chose the safe road.


etpinard:
@antoinerg or @archmoj could one of review this thing at some point today?

Thank you!

src/lib/svg_text_utils.js
```diff
@@ -134,7 +134,7 @@ exports.convertToTspans = function(_context, gd, _callback) {
                 else if(svgClass[0] === 'l') {
                     newSvg.attr({x: _context.attr('x'), y: dy - (newSvgH / 2)});
                 }
-                else if(svgClass[0] === 'a') {
+                else if(svgClass[0] === 'a' && svgClass.indexOf('atitle') === -1) {
```
archmoj: Could we use 
<pre>
if(svgClass === 'atitle') {
</pre>
here?

etpinard: Why would that be better?

archmoj: I was wondering of an edge case where 'atitle' appears in other position in the text. 
Though I was not able find any word containing those letters in order!
Anyway could we use this condition instead to test if `atitle` only occurs at the start of the string?
<pre>
else if(svgClass[0] === 'a' && svgClass.indexOf('atitle') !== 0) {
</pre>

etpinard: Ok. Done in -> https://github.com/plotly/plotly.js/pull/3513/commits/c7878b381dc5822faf20570abfa74a814297e4b1

archmoj:
Thanks @etpinard for fixing this bug.
:dancer: when the tests were passed.

Issue #3525, opened 2019-02-07
react + legendonly bug


alexcjohnson:
Calling `Plotly.react` from a state with all traces `visible: 'legendonly'` to a new normal state breaks the plot until a fresh `newPlot` call.
https://codepen.io/alexcjohnson/pen/RvxLNy?editors=1010

Originally reported in https://github.com/plotly/dash/issues/595

Separate note: why can't I click the legend lines/symbols to show/hide traces? I have to click the name or the space immediately above or below the legend line, but the lines/symbols themselves don't do it.

alexcjohnson:
This is something to do with `uirevision` - simply setting all traces `visible: 'legendonly'` in the initial figure does not trigger this bug.

etpinard:
Hmm, very strange. Using `restyle(gd, 'visible', 'legendonly')` instead of clicking on the legend items gives the correct result -> https://codepen.io/etpinard/pen/pGpdJP?editors=1010

So yeah, this is consistent with something being up with `uirevision`.

PR #3527, opened 2019-02-08
Fix react+uirevision when removing traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3525

cc @plotly/plotly_js

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2539,7 +2539,7 @@ function getTraceIndexFromUid(uid, data, tracei) {
         if(data[i].uid === uid) return i;
     }
     // fall back on trace order, but only if user didn't provide a uid for that trace
-    return data[tracei].uid ? -1 : tracei;
+    return (!data[tracei] || data[tracei].uid) ? -1 : tracei;
 }
 
 function valsMatch(v1, v2) {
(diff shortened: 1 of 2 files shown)


etpinard:
@antoinerg @archmoj  looking for reviewers on this one

I'll like to make another patch release this week before 1.45.0.

src/plot_api/plot_api.js
```diff
@@ -2539,7 +2539,7 @@ function getTraceIndexFromUid(uid, data, tracei) {
         if(data[i].uid === uid) return i;
     }
     // fall back on trace order, but only if user didn't provide a uid for that trace
-    return data[tracei].uid ? -1 : tracei;
+    return (!data[tracei] || data[tracei].uid) ? -1 : tracei;
```
archmoj: Wondering if we may add few more checks to test `tracei` to be a valid index?
<pre>
return (tracei === undefined || tracei < 0 || !data[tracei] || data[tracei].uid) ? -1 : tracei;
</pre>

etpinard: `getTraceIndexFromUid` gets called here:

https://github.com/plotly/plotly.js/blob/7d117741d0e93f79831b65faa78a113bf3d8109b/src/plot_api/plot_api.js#L2628

and `index` is filled here:

https://github.com/plotly/plotly.js/blob/7d117741d0e93f79831b65faa78a113bf3d8109b/src/plots/plots.js#L1019

so I can't be anything other than positive number.

archmoj:
:dancer:
@etpinard Thanks for the fix.

Issue #3402, opened 2019-01-05
Violins and Boxes in subplots oddly offset


nicolaskruchten:
Split off of #3390 ...

the image below (see https://plot.ly/~nicolaskruchten/260) shows that each subplot "leaves room" for traces that are present only in other subplots, leading to undesirable horizontal offsets:

![image](https://user-images.githubusercontent.com/203523/50726988-ac3f6580-10e2-11e9-90f5-17cd27594d09.png)

nicolaskruchten:
Upon further reflection, I realize why this is: there is nothing saying that the traces in the vertically-offset subplots are actually the same as the other ones and hence should be aligned. It only looks that way because they're coloured the same. I haven't tried it but I assume the same thing will happend with grouped bard.

I think this calls for a new cross-trace argument allowing for grouping of groups or something. I've already used `legendgroup` above to link the traces' visibilities together, and I think we need a new mechanism for linking offset/grouped categorical traces across subplots (this doesn't affect numerical axes, or overlaid/stacked situations).

nicolaskruchten:
Interestingly, this does *not* happen with grouped bars:

![image](https://user-images.githubusercontent.com/203523/50729594-a8bfd480-110a-11e9-9fa2-0604a3c39f69.png)

nicolaskruchten:
(although the order of the bars in the lowest subplot is odd)

nicolaskruchten:
After further thought, given that we already do what seems like the right thing for `bar` (and `scatter`! see below), I think we should align `box` and `violin` with this same behaviour.

![image](https://user-images.githubusercontent.com/203523/50775250-a7ee8600-1263-11e9-82c0-efabf3180969.png)

nicolaskruchten:
To be a bit clearer, here are two plots: same except that one is bar and one is violin.

![online_graph_maker_ _plotly_chart_studio](https://user-images.githubusercontent.com/203523/51194331-d64e1000-18b8-11e9-9d7e-8d5c2f74db23.png)

![online_graph_maker_ _plotly_chart_studio-2](https://user-images.githubusercontent.com/203523/51194335-d8b06a00-18b8-11e9-922d-ddf15b04ca26.png)

Right now there's no way to have the bars offset horizontally among the subplots if e.g. traces 0 and 2 (blue and green) are *un*-related. Similarly, there's no way to align the violins such that trace 0 is vertically aligned with trace 2 if the *are* related.

[...]

PR #3529, opened 2019-02-08
Implement alignmentgroup and offsetgroup

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3402 - at the part listed in https://github.com/plotly/plotly.js/issues/3402#issuecomment-461583613

In brief, this PR add two attributes `alignmentgroup` and `offsetgroup` to bar, histogram, box and violin traces. These attributes allow to have more control over how the cross-trace offset and width values are computed. Examining the `bar-alignment-offset` mock is probably the best way to grasp what these new attributes do.

cc @plotly/plotly_js and @nicolaskruchten - it would be nice if you could take a :eyes: at the new mocks, thanks!

some TODOs:

- [x] Incorporate `alignmentgroup` and `offsetgroup` on matching axes, once https://github.com/plotly/plotly.js/pull/3506 is merged
- [x] Decide what `alignmentgroup` and `offsetgroup` should do across trace types


src/traces/bar/defaults.js
```diff
@@ -77,5 +76,54 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
...
+
     Lib.coerceSelectionMarkerOpacity(traceOut, coerce);
+}
+
+function handleGroupingDefaults(traceIn, traceOut, layout, coerce) {
+    var orientation = traceOut.orientation;
+    // TODO make this work across matching axes too?!?
+    // TODO should this work per trace-type?
+    //      one set for bar/histogram another for box/violin?
+    //      or just one set for all trace trace types?
```
etpinard: Does anyone here have a strong opinion about this? 

Should `alignmentgroup` and `offsetgroup` work across trace types? Or should we match bar and histogram and box and violins separately?

alexcjohnson: Seems like making it work across trace types would be the most flexible - you can always just give different values to different trace types, which I guess is what the default would be. It's also probably what I'd expect as a user. It's an awfully weird use case, so not worth a huge effort, but if we omit it now and wanted to add it later it would be a breaking change.

nicolaskruchten: Bars above for counts (say) and boxes below for distributions? Seems not that weird to me :) kinda useful actually!

alexcjohnson: Ah yeah good point. That would magically work if you have the same traces in each case, and in the same order... but break those assumptions (which could happen for good reasons) and you're back to manual alignment if we don't implement this cross-type.

etpinard: Done in https://github.com/plotly/plotly.js/pull/3529/commits/e746b088d6ff929577416706f59c2d739f850640

etpinard:
@plotly/plotly_js this PR is now `reviewable`. We'll also _need_ this one for 1.45.0.

@archmoj could you review this one?

archmoj:
Excellent PR :dancer:
@etpinard would you mind adding [a codepen example](https://codepen.io/MojtabaSamimi/pen/bzzPJB?editors=0010) to the PR description which may be helpful for the Q&A? @VeraZab

etpinard:
All the new mocks in a codepen:

https://codepen.io/etpinard/pen/mvoJoz

arlowhite:
@etpinard thank you for this feature! One minor issue is that traces do not re-layout when toggled (reproduced in your "mocks"). So the user cannot turn-off traces using the legend to view the remaining traces at a larger size.

etpinard:
@arlowhite thanks, would you mind filing a bug report at https://github.com/plotly/plotly.js/issues/new ?

Issue #85, opened 2015-12-08
Showing totals for stacked bar charts


tlrdstd:
I've been trying to show totals on my stacked bar charts. I'd prefer to show them as an integral part of the graph, but I'd settle for hoverinfo if I had to.

This jsfiddle using annotations is the closest I've come: http://jsfiddle.net/u6f0q14n/3/

![](https://cloud.githubusercontent.com/assets/194300/11659874/70e8f35e-9d98-11e5-98d3-ed37c998a9cf.PNG)

That works, until I click the legend to hide a given trace:

![](https://cloud.githubusercontent.com/assets/194300/11659914/9acc5ec2-9d98-11e5-85f2-f9b610169281.PNG)

The annotations (totals) are left hanging in the air, and the graph does not nicely resize to the updated data. As best I can tell, the provided config options do not offer a pre-packaged way to solve this.

PPPW:
I looked at #34, I wonder will it show total in 'markers+text' for stacked bar? It seems for me, if I specify 'text' for each series in stacked bar, then it will display text near each series, not sure is it possible to display a total number for all series, using 'markers+text'? Thanks!

nicolaskruchten:
I'd also like the ability to show totals in hovertemplates.

Perhaps we want a `texttemplate` ?

nicolaskruchten:
this seems like a stack-level rather than trace-level thing. related to #3614 ?

mangel2095:
I had the same need, and maybe it's far to being the best solution, but it works, I used `restyleData ` property like input in the callback, and additionaly put the figure of the plot like state, in this way, with de `restyleData`, detect when user interact with the figure. I save in a list the legends that are `visible` in the data figure, then, when define annotations only do it with the visibles legends.

 Basically is recalculate all,  but at least that happens in a short time lapse.

sean-wallace:
This would be a great feature for stacked area charts too!

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #3531, opened 2019-02-10
Waterfall charts

### LINKED ISSUE #3037: Waterfall trace type

archmoj:
Fixes #3037.
A prototype for `waterfall` new trace type mainly based on `bar`.

TODO:
- [x] add a selection test
- [x] Why aren't arrayOk `offset` and `width` working?
- [x] add `total` color container an keep `marker` similar to bars
- [x] implement waterfallmode, waterfallgap and waterfallgroupgap
- [x] fixup `cliponaxis:false` support
- [x] fixup lasso/select

[codepen](https://codepen.io/MojtabaSamimi/pen/BbmVVd?editors=0010)
waterfall_profit-loss_2018_positive-negative


src/traces/bar/calc.js
```diff
@@ -31,9 +33,25 @@ module.exports = function calc(gd, trace) {
...
+    // set position and size (as well as for waterfall total size)
+    var previousSum = 0;
+    for(i = 0; i < serieslen; i++) {
+        cd[i] = {
+            p: pos[i],
+            s: size[i]
+        };
+
+        if(isWaterfall) {
+            if(cd[i].s === undefined) {
```
etpinard: So, in order to draw partial sums, one has to set `size` as an array with `undefined` entries corresponding to the location of the partial sums?

As I said before in https://github.com/plotly/plotly.js/issues/3037#issuecomment-454864262, I'm very much against that. But oh well, for a quick prototype, this is :ok_hand:

src/traces/bar/cross_trace_calc.js
```diff
@@ -36,7 +36,8 @@ function crossTraceCalc(gd, plotinfo) {
         var fullTrace = fullTraces[i];
         if(
             fullTrace.visible === true &&
-            Registry.traceIs(fullTrace, 'bar') &&
+            (Registry.traceIs(fullTrace, 'bar') ||
+            Registry.traceIs(fullTrace, 'waterfall')) &&
```
etpinard: We should a new trace module category that encompasses `bar`, `histogram` and `waterfall` traces

https://github.com/plotly/plotly.js/blob/02e51d8f5ba5601f53ee3dc839ba05fb12c7f8a2/src/traces/bar/index.js#L31

or better yet (if it works out), maybe the `waterfall` trace module could be declared with the `'bar'` category?

src/traces/bar/calc.js
```diff
@@ -17,7 +17,9 @@ var calcSelection = require('../scatter/calc_selection');
 module.exports = function calc(gd, trace) {
```
etpinard: Thanks for reusing almost :sparkle: **all** :sparkle: of `Bar`.

Now, I would prefer if we use an inheritance pattern more similar to how `barpolar` calls `Bar` methods:

https://github.com/plotly/plotly.js/blob/02e51d8f5ba5601f53ee3dc839ba05fb12c7f8a2/src/traces/barpolar/calc.js#L76-L104

instead of having `type === 'waterfall'` clauses inside `traces/bar/`. In other words, `traces/bar/` `shouldn't have to know anything about waterfall traces.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,68 @@
...
+    base: barAttrs.base,
+
+    offset: barAttrs.offset,
+
+    width: barAttrs.width,
+
+    marker: extendFlat({}, barAttrs.marker, {
+        shape: {
+            valType: 'enumerated',
+            values: ['rectangle', 'triangle'],
```
etpinard: Do we "really" want this? @archmoj have you seen anything like in the wild?

src/traces/waterfall/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+
+'use strict';
+
+var Waterfall = {};
+
+Waterfall.attributes = require('./attributes');
+Waterfall.layoutAttributes = require('../bar/layout_attributes');
+Waterfall.supplyDefaults = require('./defaults');
+Waterfall.supplyLayoutDefaults = require('../bar/layout_defaults');
```
etpinard: Interesting. Should waterfall trace adhere to `barmode`, `bargroupgap` and `bargap`?

Having a group and an overlay `barmode` probably makes sense. I doubt we that `barmode: 'stack'` should be implemented.

etpinard: `barnorm` is probably useful too.

Now, maybe we should add a separate set of attribute just for waterfall traces e.g. `waterfallmode`, `waterfallgap` ...

I doubt that graphs with bar and waterfall on the same subplots are very common.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,68 @@
...
+
+    orientation: barAttrs.orientation,
+
+    base: barAttrs.base,
+
+    offset: barAttrs.offset,
+
+    width: barAttrs.width,
+
+    marker: extendFlat({}, barAttrs.marker, {
```
etpinard: We should have two `marker` containers for waterfall traces: one for increasing bars and one for decreasing bars, similar to how `ohlc` and `candlestick` do it.

https://github.com/plotly/plotly.js/blob/02e51d8f5ba5601f53ee3dc839ba05fb12c7f8a2/src/traces/ohlc/attributes.js#L89-L91

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,68 @@
...
+    x0: barAttrs.x0,
+    dx: barAttrs.dx,
+    y: barAttrs.y,
+    y0: barAttrs.y0,
+    dy: barAttrs.dy,
+
+    r: barAttrs.r,
+    t: barAttrs.t,
+
+    text: barAttrs.text,
```
etpinard: From https://github.com/plotly/plotly.js/issues/3037#issuecomment-459806749, 

> also note we will want to support `text` display which should be the delta value, not the absolute value (and the start, delta and end values should be available in the hovertemplate if possible).

Is this the case?

Moreover, we should probably have `increasing.*textfont` and `decreasing.*textfont` container. I can imagine users wanting different text font colours for increasing vs decreasing bars.

etpinard:
@archmoj can you point me the "real" waterfall mocks you added? I see that you remade all the `bar` mocks with waterfall traces. That's cool, but not 100% helpful as some bar attributes won't apply to waterfall traces (e.g. as mentioned in https://github.com/plotly/plotly.js/pull/3531#discussion_r255351883)

Similarly, copying the entire `bar_test.js` suite and switching traces to `type: 'waterfall'` is an overkill. We don't have to test "all" the `Bar` logic that we're reusing to render waterfall traces.

src/traces/waterfall/attributes.js
etpinard: Maybe we could implement https://github.com/plotly/plotly.js/issues/85 if the to-be-added logic is sufficient similar in the stacked bar case.

archmoj:
> @archmoj can you point me the "real" waterfall mocks you added?

Most of them start with `waterfall_profit-loss_`. Thanks.

src/traces/waterfall/index.js
alexcjohnson: Even group mode is extremely hard to read, but it would be much clearer if we made connecting lines - something like `connectors: {width, color, dash, mode: ('between'|'spanning')}`?

Dunno about the name and values for `mode` but I'm thinking `between` would be like this:
between
and `spanning` like:
spanning
My own preference would be for spanning solid black lines to be the default, because it makes clearer which end of the bar is the start and which is the end... not too important for it to be the default but I do think the connector feature should be there from the start.

There are also examples online of lines connecting the midpoints of the two bars, I'm not a fan but some people may want it:
halfway connectors

FWIW `stack` does pop up here and there https://peltiertech.com/excel-waterfall-charts-bridge-charts/:
stack
But I don't see it as particularly important at least to start.

src/traces/bar/cross_trace_calc.js
archmoj: Yes the `waterfall` could be declared as `'bar'` category. 
Done in 56786530ff31d865aefd30ad4ebab4cff7487f94.

src/traces/waterfall/calc.js
```diff
@@ -0,0 +1,123 @@
...
+
+'use strict';
+
+var Axes = require('../../plots/cartesian/axes');
+var hasColorscale = require('../../components/colorscale/helpers').hasColorscale;
+var colorscaleCalc = require('../../components/colorscale/calc');
+var arraysToCalcdata = require('./arrays_to_calcdata');
+var calcSelection = require('../scatter/calc_selection');
+
+function extractOperators(list) {
```
etpinard: Interesting, but I can't allow. Again, this would make `waterfall` data array "too different" from our other trace types.

What's wrong with having a separate data array e.g. `sums` as proposed in https://github.com/plotly/plotly.js/issues/3037#issuecomment-454864262 ?

archmoj: Using a pattern like this https://github.com/plotly/plotly.js/blob/b0fdceda91abdbf2ccfccde1b4106db208931ea5/test/image/mocks/waterfall_profit-loss_2018_colorscale.json#L7-L23 we no longer use `magic` numbers and it may be easier for the users to add and figure out the calculation related for each row.
But yes I see your point @etpinard and would likely switch to use an array for the operations there.

archmoj: Revised in https://github.com/plotly/plotly.js/pull/3531/commits/4383a71e7f52c8ac4aeb3313cc44d3f3146a929c.

nicolaskruchten:
OK, this is starting to look really hot!

Some feedback:

* I worry a lot about the `operators` vector: I don't think we should support `-` and `%` out of the box, and I have a big concern about the `=` operator because it basically just causes the corresponding value to be ignored, which opens up room for lots of errors. I would strongly favour treating everything as `+` and `null`/`NaN` as `=` so it's all encoded in just one place. We would then add special handling for the very first one (i.e. is a "+ from 0" or "initial ="?)
* I'm OK with not supporting stacking for now, but if side-by-side'ing works that's great. Stacking seems problematic anyways because if values with the same `x` value have different signs it'll get super hard to read.
* I think we should remove error bars, triangles and continuous color
* We should have increasing/decreasing/intermediate color specifications, which default to red/green/blue, and each `waterfall` trace should have a single legend marker which shows all three colors (similar to `ohlc` and `candlestick` (and collapses down to a square if all three are the same!)
* We should support explicit CSS per-marker coloring to override the above, with some heuristic to generate the colors in the legend marker

The above sums up to us not being able to encode charts like the one below (especially the legend), but I think that's something I can live with:

image

archmoj:
@nicolaskruchten Awesome feedback. 
Much appreciated.

etpinard:
> * I would strongly favour treating everything as `+` and `null`/`NaN` as `=`

I am strongly against that. I don't want `null` meaning something different for `waterfall` when it is means "skip this point" for all our other trace types. 

Again, what's wrong with `sums` https://github.com/plotly/plotly.js/pull/3531#discussion_r256146667

nicolaskruchten:
> Again, what's wrong with sums

Because it would discard/ignore the corresponding value in the data array, which IMO leaves too much room for error, especially in a UI.

nicolaskruchten:
> I don't want null meaning something different for waterfall when it is means "skip this point" for all our other trace types 

I would argue that this *is* treating `null` as 'skip this point' in the sense that it doesn't add to the running total. And this is visually represented by drawing a bar to the baseline.

nicolaskruchten:
In line charts, we don't *just* skip the point, we also visually represent this by not connecting the lines.

etpinard:
@nicolaskruchten What made you change your mind since https://github.com/plotly/plotly.js/issues/3037#issuecomment-454874172 ?

nicolaskruchten:
I realized that this situation is actually really scary:

```
values: [1,2,3,4,5],
operators: ['+','+','=','+','+]
```

The number 3 is in the data and yet is ignored.

It would be the equivalent of having, in line traces:

```
values: [1,2,3,4,5],
drawlines: [true, true, false, true, true]
```

etpinard:
Yes, I agree that `operator` thing had to go. But what about my https://github.com/plotly/plotly.js/issues/3037#issuecomment-454864262 ?

nicolaskruchten:
Same problem:

```
values: [1,2,3,4,5],
sums: ['', '', 'intermediate', '', 'total']
```

The number 3 is ignored.

etpinard:
Why is that bad?

nicolaskruchten:
I think it's just really ambiguous/dangerous to have data arrays where we throw away/ignore data. It seems easy to misalign things and have a chart that shows the wrong output. I would prefer those be forced to be `null` or something as well.

nicolaskruchten:
Maybe if we actually rendered the delta of whatever number was there it would solve my problem. Like the 3 would cause the 'intermediate' bar to start 3 points higher up.

archmoj:
> Maybe if we actually rendered the delta of whatever number was there it would solve my problem. Like the 3 would cause the 'intermediate' bar to start 3 points higher up.
That may hide the issue: for example if there is -3 before 3 it won't be obvious from the graph that there could be a problem with input data.

nicolaskruchten:
No I meant the bar that drops to the axis would start where the other bar normally would. I’ll draw you a sketch tomorrow :)

nicolaskruchten:
Some food for thought... this is how Mighty Excel does it: https://www.microsoft.com/en-us/microsoft-365/blog/2015/08/04/introducing-the-waterfall-chart-a-deep-dive-to-a-more-streamlined-chart/

image

In the image above, I marked `d` with "set as total".

nicolaskruchten:
So actually there are a few independent things happening that we might want to model in this spec for any given datum:

* Does this value affect the running total?
* Is this value a relative one or an absolute one?
* Do we draw the bar starting at the axis or starting at the end point of the previous one?

In general, the answers are "yes, relative, previous". For Excel's totals, the answers are "yes, absolute, axis". The way we've considered "intermediate" values are as "no, relative (if you assume the value is 0), axis".

nicolaskruchten:
OK here is a proposal (sorry, no obviously-silly name after all, but the day is young!)

In the image below, the top two specs yield the same figure but for different reasons, and the right hand one has the correct subtotal in the data. The bottom two specs show the same incorrect total in two ways, whose visual representation is highlighted in yellow.

img_1673

The three modes are:

* `rel` aka `relative` aka `''` aka `null`: affects the running total, relative, drawn from the previous relative point
* `abs_from_prev`: doesn't affect the running total, absolute, drawn from previous relative point
* `abs_from_axis`: resets the running total, absolute, drawn from the axis

On the extreme right hand side are some variations on proposals for a unified legend glyph which includes all three colors, and could gracefully handle two- or one-color cases should they arise (i.e. only increasing, or no-totals)

@alexcjohnson @chriddyp feedback welcome! (already chatted about this with @archmoj and @etpinard  in person)

alexcjohnson:
Oh I see, as you've envisioned it a nonzero `abs_from_prev` would put the *base* of the bar somewhere other than zero but not affect the *end* of the bar, nor the running total. That seems weird, unless we can identify a real use case for it (and even if we can, I kind of feel like a different attribute (`base` probably, to match `bar` traces) would be cleaner.

I was imagining it would behave identically to `abs_from_axis`, but with the actual calculated running total added on. So with a value of `-75` you would get exactly the bottom right graph, whereas `+75` would yield a ramp up to a total of 225, but the bottom of this bar would always be at 0.

Does that change what they should be called? Perhaps `relative`, `total` (for `abs_from_axis`), and `total difference`?

BTW elsewhere for multi-word attribute *values* we've used lowercase with spaces.

alexcjohnson:
re: legend items - the ones we have for finance data make it intuitive, I think, which color/style is up and which is down:
screen shot 2019-03-07 at 5 44 17 pm
We have a tougher problem here, as we need to distinguish up, down, and total. The first one you gave may do it, but the upper left triangle should be up, the lower right should be down. But I wonder if just drawing a little waterfall would do:
screen shot 2019-03-07 at 5 57 13 pm

nicolaskruchten:
I'm fine with your legend glyph idea @alexcjohnson but I'm afraid I don't follow your comments re the `abs` options... can you make some similar diagrams to mine so I can compare?

I agree that the bar that drops almost down to zero but not quite is a bit odd :)

alexcjohnson:
My `total` is the same as your `abs_from_axis`, and my `total difference` has the same effect on the running total as `relative` (rather than resetting the running total as `total`/`abs_from_axis` does), but the bar is drawn down to the axis like `total`.
img_8779

I still don't like the names though, and I'm hoping we can find a set of three single words to use. Perhaps this new one is actually `total`, as it's totaling up all the values up to and including that point, and `abs_from_axis` just gets called `absolute` since it's giving a completely new value independent of what came before. That's also nice because for efficiency (mostly for users' sake if they're typing by hand) we could just look at the first letter, `r`/`t`/`a`.

nicolaskruchten:
Ah I see, thanks.

I think my two options were meant to make sense as “sanity checks for precomputed totals” as you mentioned earlier, whereas your total-diff option doesn’t seem to do this any more as far as I can tell..?

nicolaskruchten:
And yes, single words would be great :)

alexcjohnson:
They’re identical sanity checks: the “correct” entry (zero for total diff or the running total for total/abs_from_axis) is connected to the previous with a flat line, an “incorrect” entry has a sloped line. In fact they have an identical space of possible outcomes: total diff with value x is drawn the same as total/abs_from_axis with value running_total + x

nicolaskruchten:
OK. It's a stretch for me to imagine an intended use for non-zero total diff (in either case, tbh) but it has the important property of showing that number rather than just ignoring it. So can we call by `abs_from_previous` the `total` and my `abs_from_axis` the `absolute`? My reasoning there is that `absolute` has more of a 'reset to this point and move on' connotation, whereas `total` has more of a 'show me the total and move on' connotation (to me). Maybe `subtotal` is even clearer in that usually subtotals happen mid-way?

src/traces/bar/cross_trace_calc.js
```diff
@@ -108,7 +108,7 @@ function setGroupPositions(gd, pa, sa, calcTraces) {
     collectExtents(calcTraces, pa);
 }
 
-function initBase(gd, pa, sa, calcTraces) {
```
etpinard: No need to do this.

Please keep the more-general `(gd, pa, sa, calcTraces)` call signatures for all methods in this file. that way they can potentially be easily reused by other trace module.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+    increasing: directionAttrs(),
+    decreasing: directionAttrs(),
+
+    connector: {
+        color: lineAttrs.color,
+        width: lineAttrs.width,
+        dash: lineAttrs.dash,
+        mode: {
+            valType: 'enumerated',
+            values: ['begin+end', 'steps', false],
```
etpinard: `'begin+end'` makes this attribute look like a `flaglist`, which it is not.

These here look similar to the `line.shape` attribute:

https://github.com/plotly/plotly.js/blob/7e68a2e56f5cfca1286163bda63b5f8263c43451/src/traces/scatter/attributes.js#L235

Maybe `connector.shape: 'hv' | 'hvh'` would be best.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+            editType: 'style',
+            color: barAttrs.marker.line.color,
+            width: barAttrs.marker.line.width
+        }
+    };
+}
+
+module.exports = {
+
+    valuetype: {
```
etpinard: I'm not a big fan of `valuetype` which to me sounds like a scalar. I'm voting for `sums` but I have no strong opinion.

src/traces/waterfall/cross_trace_calc.js
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+'use strict';
+
+var Registry = require('../../registry');
+var barCrossTraceCalc = require('../bar/cross_trace_calc').crossTraceCalc;
+
+module.exports = function crossTraceCalc(gd, plotinfo) {
+
+    barCrossTraceCalc(gd, plotinfo);
```
etpinard: Hmm. What happens when we have `bar` and `waterfall` traces on the same subplot?

src/traces/waterfall/defaults.js
```diff
@@ -0,0 +1,133 @@
...
+    coerce('offset');
+    coerce('width');
+
+    coerce('text');
+    coerce('hovertext');
+    coerce('hovertemplate');
+
+    var textPosition = coerce('textposition');
+
+    var hasBoth = Array.isArray(textPosition) || textPosition === 'auto';
```
etpinard: This is tricky logic. Isn't this the same as in `bar/defaults.js`? If so, can we split this from `bar/defaults.js` and call it from here?

src/traces/waterfall/defaults.js
```diff
@@ -0,0 +1,133 @@
...
+    handleDirection(coerce, 'increasing', INCREASING_COLOR);
+    handleDirection(coerce, 'decreasing', DECREASING_COLOR);
+    handleDirection(coerce, 'marker', defaultColor);
+
+    coerce('selected.marker.color');
+    coerce('unselected.marker.color');
+
+    Lib.coerceSelectionMarkerOpacity(traceOut, coerce);
+
+    traceOut._autoMarkerColor = !!(
```
etpinard: Nop, use `Colorscale.hasColorscale`:

https://github.com/plotly/plotly.js/blob/7e68a2e56f5cfca1286163bda63b5f8263c43451/src/components/colorscale/helpers.js#L20-L45

src/traces/waterfall/layout_attributes.js
```diff
@@ -0,0 +1,50 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
+    barmode: {
```
etpinard: I'm not sure we should be reusing the `bar` layout attributes, but then again I'm not sure what _should_ happen on subplots with bar and waterfall traces coexisting. To me, it seems safer to have a different set of layout attributes for waterfall traces: `waterfallmode`, `waterfallgap` and `waterfallgroupgap`, I could be wrong though.

test/image/compare_pixels_test.js
```diff
@@ -99,6 +99,8 @@ function untestableFilter(mockName) {
         mockName.indexOf('mapbox_') !== -1
     );
 
+    cond = (mockName.indexOf('waterfall') !== -1); // only run waterfall on CI.
```
etpinard: :zap: don't commit stuff like that on branches associated with PRs. Thanks!

test/image/mocks/waterfall_multiline_labels.json
```diff
@@ -0,0 +1,161 @@
+{
+    "data":[
+        {
+            "type": "waterfall",
```
etpinard: We can probably remove this mock. It doesn't test anything new, am I right?

test/jasmine/assets/mock_lists.js
```diff
@@ -14,6 +14,9 @@ var svgMockList = [
     ['axes_enumerated_ticks', require('@mocks/axes_enumerated_ticks.json')],
     ['axes_visible-false', require('@mocks/axes_visible-false.json')],
     ['bar_and_histogram', require('@mocks/bar_and_histogram.json')],
+    ['waterfall', require('@mocks/waterfall_profit-loss_2018_colorscale.json')],
```
etpinard: Adding just one `waterfall` mock here would suffice here.

test/jasmine/tests/waterfall_test.js
```diff
@@ -0,0 +1,1506 @@
...
+            y: [1, 2, 10, 30]
+        }]);
+
+        var cd = gd.calcdata;
+        assertPointField(cd, 'x', [[1, NaN, NaN, 15]]);
+        assertPointField(cd, 'y', [[1, 3, 13, 43]]);
+    });
+});
+
+describe('Waterfall.crossTraceCalc (formerly known as setPositions)', function() {
```
etpinard: :zap: the ` (formerly known as setPositions)` please

package-lock.json
```diff
@@ -4129,8 +4129,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
```
etpinard: No need to commit this. 

Which npm version are you on?

etpinard:
Nice work @archmoj ! Looks like you're almost there.

The biggest unsolved questions (to me) are:

- what to do with the `waterfall` layout attributes? Should we reuse the `bar*` set or add waterfallmode, waterfallgap and waterfallgroupgap
- what to rename your `valuetype` attribute? I'm voting for `sums`

etpinard:
Moreover, did you get box/lasso selections to work? If so, could you add a test in `select_test.js`? Thanks!

src/traces/waterfall/attributes.js
```diff
@@ -53,20 +62,20 @@ module.exports = {
 
     constraintext: barAttrs.constraintext,
 
-    cliponaxis: barAttrs.cliponaxis,
```
etpinard: Actually, having `cliponaxis` from the start would be nice. @archmoj are there any complications with `cliponaxis` that I can't think of?

nicolaskruchten:
I agree that `valuetype` isn't great, but `sums` doesn't seem right either... "sum = relative", "sum = total" ?

src/traces/waterfall/attributes.js
archmoj: I agree with you that 'begin+end' may be confused with flag-lists.
On the other hand 'hv', 'vh', 'hvh' and 'vhv' may be confusing whereas both horizontal and vertical orientations are supported.

archmoj:
> I agree that `valuetype` isn't great, but `sums` doesn't seem right either... "sum = relative", "sum = total" ?

@alexcjohnson It would be great to have your thoughts on this as well. Thanks!

src/traces/waterfall/attributes.js
alexcjohnson: (copying @nicolaskruchten's comment so I can see the discussion alongside the code)
> I agree that `valuetype` isn't great, but `sums` doesn't seem right either... "sum = relative", "sum = total" ?

How about `measure`? Generic enough that people would need to read the docs to understand its meaning, but correspondingly free of baggage.

Is there even a use case for this being a scalar? I'd think this should be a `data_array` - also that `dflt` should be omitted and the expected values listed in `description`. I guess the one case I could imagine wanting to make easier than having to provide a whole array is putting a single sum at the end, so we could allow a scalar to mean "let all but the last one be relative, and do this (absolute or total) with the last one."

src/traces/waterfall/defaults.js
archmoj: Good call. Revised.

src/traces/waterfall/attributes.js
archmoj: Added back. 
Not sure why on the mock below it continues clipping the first trace even if it was set to false. 
https://github.com/plotly/plotly.js/blob/fed3c97af3997a99e05ddb7fc18a5220d571c235/test/image/baselines/waterfall_cliponaxis-false.png

src/traces/waterfall/cross_trace_calc.js
archmoj: A mock is added to display waterfall with bars. We also have one for waterfall with histogram.
https://github.com/plotly/plotly.js/blob/fed3c97af3997a99e05ddb7fc18a5220d571c235/test/image/baselines/waterfall_and_bar.png

src/traces/waterfall/defaults.js
archmoj: I removed this block as colorscale is not going to be supported there.

src/traces/waterfall/attributes.js
etpinard: I'm a fan of `measure` :+1:

src/traces/waterfall/defaults.js
etpinard: Thanks!

src/traces/waterfall/defaults.js
etpinard: Nicely done!

src/traces/waterfall/attributes.js
etpinard: Yeah, looks like `cliponaxis` isn't working.

src/traces/waterfall/attributes.js
etpinard: You're right. My mistake!

I'm ok with `values: ['steps', 'spanning', 'between', false],` as in your latest iteration.

src/traces/waterfall/layout_attributes.js
etpinard: After a quick talk with @alexcjohnson ,  I think making waterfall-specific trace layout attributes is the way to go for now. That is, these here should be `waterfallmode`, `waterfallgap` and `waterfallgroupgap`.

src/traces/waterfall/attributes.js
```diff
@@ -78,9 +80,9 @@ module.exports = {
 
     orientation: barAttrs.orientation,
 
-    offset: barAttrs.offset,
+    offset: extendFlat({}, barAttrs.offset, { arrayOk: false }),
```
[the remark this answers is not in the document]
archmoj: They used to work on separated bars. But then I noticed it would become difficult to read the `waterfall` graph and more complicated to compute the connector endpoints.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+'use strict';
+
+var barAttrs = require('../bar/attributes');
+var lineAttrs = require('../scatter/attributes').line;
+var extendFlat = require('../../lib/extend').extendFlat;
+
+function directionAttrs() {
+    return {
+        editType: 'style',
+        color: barAttrs.marker.color,
```
etpinard: We should override the description here and for `marker.line.color` as waterfall trace don't support numeric colorscales.

etpinard: Moreover, what currently happens when we have

```js
{
  type: 'waterfall',
  x: [/* */],
  y: [/* */],
  marker: {color: [1,2,3]}
}
```

?

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+
+    base: {
+        valType: 'number',
+        dflt: null,
+        arrayOk: false,
+        role: 'info',
+        editType: 'calc',
+        description: [
+            'Sets where the bar base is drawn (in position axis units).'
```
etpinard: What does `base` do? I suspect it simply offsets the whole trace along the value axis?

src/traces/waterfall/cross_trace_calc.js
```diff
@@ -0,0 +1,50 @@
...
+    var fullTraces = gd._fullData;
+    var calcTraces = gd.calcdata;
+    var waterfalls = [];
+    var i;
+
+    for(i = 0; i < fullTraces.length; i++) {
+        var fullTrace = fullTraces[i];
+        if(
+            fullTrace.visible === true &&
+            Registry.traceIs(fullTrace, 'bar') &&
```
etpinard: We can :zap: this line. The `fullTrace.type === 'waterfall'` check below should suffice.

src/traces/waterfall/attributes.js
archmoj: Yes exactly. Here is a [codepen(https://codepen.io/MojtabaSamimi/pen/oVaYbX?editors=0010) with `base`.

src/traces/waterfall/attributes.js
archmoj: Now they turn to black. So you are right about overriding the description.

src/traces/waterfall/defaults.js
```diff
@@ -0,0 +1,89 @@
...
+
+    function coerce(attr) {
+        return Lib.coerce(traceOut._input, traceOut, attributes, attr);
+    }
+
+    for(var i = 0; i < fullData.length; i++) {
+        traceOut = fullData[i];
+
+        traceIn = traceOut._input;
+        if(fullLayout.barmode === 'group') {
```
etpinard: We should filter out non-waterfall traces here. Actually this whole block (copied from `bar/defaults.js`) is not-optimal. We shouldn't loop over all the traces when `(bar|waterfall)mode !== 'group'`.

Something like:

```js
if(fullLayout.waterfallmode === 'group') {
  for(var i = 0; i < fullData.length; i++) {
     traceOut = fullData[i];
     if(traceOut.type === 'waterfall') {
        traceIn = traceOut._input;
        handleGroupingDefaults(traceIn, traceOut, fullLayout, coerce);
     }
  }
}

**Comment by etpinard** [comment]
### Updated list of unresolved things:

- [ ] implement waterfallmode, waterfallgap and waterfallgroupgap
- [ ] fixup `cliponaxis:false` support
- [ ] Why aren't arrayOk `offset` and `width` working?
- [ ] add a selection test
- [ ] override `?marker.color` description waterfall

File: src/traces/waterfall/attributes.js (hunk 24 above)
Comment (etpinard) [reply src/traces/waterfall/attributes.js]: Great. It makes sense to have `base` as `arrayOk:false` for waterfalls :ok_hand:

File: src/traces/waterfall/attributes.js (hunk 23 above)
Comment (etpinard) [reply src/traces/waterfall/attributes.js]: > Now they turn to black. 

... which I guess is ok? Are we planning on ever adding colorscale support for waterfall traces?

File: src/traces/waterfall/attributes.js (hunk 22 above)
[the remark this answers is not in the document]
Comment (etpinard) [reply src/traces/waterfall/attributes.js]: So, they _still_ work and you made them `arrayOk:false` to avoid confusion?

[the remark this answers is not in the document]
Comment (etpinard) [reply src/traces/waterfall/attributes.js]: I think that arrayOk `offset` and `width` are still valuable, especially if users want something more general than the set [waterfallmode, waterfallgap and waterfallgroupgap].

If things get too messy bringing back arrayOk `offset` and `width`, I would prefer :hocho: `offset` and `width` entirely from this first iteration than having them in but as `arrayOk:false`.

File: src/traces/waterfall/attributes.js (hunk 23 above)
Comment (archmoj) [reply src/traces/waterfall/attributes.js]: I think they could be pretty useful for cases to illustrate `extreme` jumps. 
Here is one example found on the web: https://i.ytimg.com/vi/tapO6Dcf9i4/maxresdefault.jpg

File: src/traces/waterfall/attributes.js (hunk 22 above)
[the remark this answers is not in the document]
Comment (archmoj) [reply src/traces/waterfall/attributes.js]: Thanks for the comments @etpinard. OK I would try to keep them there.

File: src/traces/waterfall/attributes.js [hunk 27]
```diff
@@ -0,0 +1,113 @@
...
+    orientation: barAttrs.orientation,
+
+    offset: extendFlat({}, barAttrs.offset, { arrayOk: false }),
+
+    width: extendFlat({}, barAttrs.width, { arrayOk: false }),
+
+    marker: directionAttrs(),
+    increasing: directionAttrs(),
+    decreasing: directionAttrs(),
+
```
etpinard: Let's add `totals` container too

src/traces/waterfall/attributes.js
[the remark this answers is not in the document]
archmoj: Done in https://github.com/plotly/plotly.js/pull/3531/commits/a05352fa492c89dae5cde3b2afbae445811750fe.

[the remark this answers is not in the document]
etpinard: Looking good. Thanks!

nicolaskruchten:
I'd like to do a final review pass on the API today after the meeting if you can hold off on dancering till later today :)

archmoj:
> I'd like to do a final review pass on the API today after the meeting if you can hold off on dancering till later today :)

The latest API is presented in [this codepen](https://codepen.io/MojtabaSamimi/pen/LaXezJ?editors=0010). Noting that in addition to `marker`, `increasing` & `decreasing`, another container for `totals` would be added for sums.

nicolaskruchten:
Can we get the hovers to be the same colors as the bars?

image

archmoj:
> Can we get the hovers to be the same colors as the bars?
> 
> image

Yes. Just discussed that with @etpinard.

nicolaskruchten:
when `connector.mode` is `between` could we add some line segments next to the bars so there's not a jagged effect like this?

image

nicolaskruchten:
Basically in the case above we'd have a matching-green line segment above the first bar and another one below the second bar.

archmoj:
> Basically in the case above we'd have a matching-green line segment above the first bar and another one below the second bar.

I see the point there. But the data representation may be slightly off with higher line width values.

nicolaskruchten:
Yes, it would be off by a couple of pixels, but I think that's ok. We can make it yet another mode if you like. This same effect exists in spanning mode.

archmoj:
> Yes, it would be off by a couple of pixels, but I think that's ok. We can make it yet another mode if you like. This same effect exists in spanning mode.

If that is OK, I may give a try. Now wondering what should happen when using connector with dash lines?

alexcjohnson:
> Yes, it would be off by a couple of pixels, but I think that's ok. We can make it yet another mode if you like. This same effect exists in spanning mode.

I guess that's OK - we can sacrifice a little fidelity for aesthetics. But in spanning mode the line is a different color, so it's intuitive (and precisely correct) to read the center of that line as the data value.

> we'd have a matching-green line segment above the first bar and another one below the second bar.

Do it just by extending the existing rect the right number of pixels, rather than adding a new line. A new line would make a mess with opacity, and may end up looking a subpixel longer due to antialiasing.

> Now wondering what should happen when using connector with dash lines?

I don't think this changes anything. In spanning mode the bar will end halfway across the line, and in between mode the bars should still extend by the width of the line.

etpinard:
> * [ ] implement waterfallmode, waterfallgap and waterfallgroupgap
> * [ ]  fixup `cliponaxis:false` support

@archmoj looks like the two items :arrow_up: are still missing. Can I do anything to help out?

archmoj:
> > * [ ]  implement waterfallmode, waterfallgap and waterfallgroupgap
> > * [ ]   fixup `cliponaxis:false` support
> 
> @archmoj looks like the two items  are still missing. Can I do anything to help out?

Thanks for asking @etpinard. Yes indeed my first attempt for `barmode` > `waterfallmode` was not successful. It looks to me that we need to wait for https://github.com/plotly/plotly.js/pull/3680 to be merged first to avoid conflicts?

etpinard:
> Yes indeed my first attempt for `barmode` > `waterfallmode` was not successful. It looks to me that we need to wait for #3680 to be merged first to avoid conflicts?

Don't wait. This PR is waaaaaaay more important that other fix. I'll handle the merge conflict.

src/traces/waterfall/attributes.js
etpinard: ... We still need to override the description.

src/components/drawing/index.js
```diff
@@ -360,8 +360,23 @@ drawing.pointStyle = function(s, trace, gd) {
     });
 };
 
+drawing.waterfallStyle = function(d, trace) {
```
etpinard: This thing doesn't do any drawing now. Can we move that to `waterfall/calc.js`?

src/components/legend/style.js
```diff
@@ -241,8 +242,49 @@ module.exports = function style(s, gd) {
...
+
+            if(w) {
+                p.call(Color.stroke, d0.mlc || container.color);
+            }
+        });
+    }
+
     function styleBars(d) {
         var trace = d[0].trace;
+        if(trace.type === 'waterfall') return d;
```
etpinard: Why do we need this line? In other words, why is `styleBars` ever calls via waterfall traces?

src/components/drawing/index.js
```diff
@@ -360,8 +360,23 @@ drawing.pointStyle = function(s, trace, gd) {
...
+    var marker = trace.marker;
+
+    if(d.isSum) {
+        return trace.totals || marker;
+    } else {
+        if(d.rawS > 0) return trace.increasing || marker;
+        if(d.rawS < 0) return trace.decreasing || marker;
```
etpinard: Hmm. This logic hasn't been updated it looks like.

@archmoj did we agree that `marker.color` if set would take precedence over `(increasing|decreasing|totals).marker.color`

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+    cliponaxis: barAttrs.cliponaxis,
+
+    orientation: barAttrs.orientation,
+
+    offset: barAttrs.offset,
+
+    width: barAttrs.width,
+
+    marker: directionAttrs(),
```
etpinard: @archmoj are you still planning on implementing `marker`?

src/traces/waterfall/defaults.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function crossTraceDefaults(fullData, fullLayout) {
+    var traceIn, traceOut;
+
+    function coerce(attr) {
+        return Lib.coerce(traceOut._input, traceOut, attributes, attr);
+    }
+
+    for(var i = 0; i < fullData.length; i++) {
```
etpinard: Just like I mentioned in https://github.com/plotly/plotly.js/pull/3531#discussion_r267427724 - no need to loop over all the traces when `waterfallmode` isn't `group`.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+                'Sets the shape of connector lines.'
+            ].join(' ')
+        },
+        editType: 'plot'
+    },
+
+    offsetgroup: barAttrs.offsetgroup,
+    alignmentgroup: barAttrs.offsetgroup,
+
+    selected: barAttrs.selected,
```
etpinard: These attribute containers don't appear to be implemented. No big deal, but please remove them from the attribute declaration.

src/components/legend/style.js
```diff
@@ -241,8 +242,49 @@ module.exports = function style(s, gd) {
         txt.selectAll('text').call(Drawing.textPointStyle, tMod, gd);
     }
 
+    function styleWaterfalls(d) {
```
etpinard: image

Could we not show the "totals" part in the legend symbol when the trace only shows increasing/decreasing bars.

src/traces/waterfall/attributes.js
etpinard: Actually, we should keep `[un]selected.marker.opacity`

nicolaskruchten:
Could we avoid the little gap in the corner here?
image

src/traces/bar/cross_trace_calc.js
```diff
@@ -56,7 +56,7 @@ function crossTraceCalc(gd, plotinfo) {
 function setGroupPositions(gd, pa, sa, calcTraces) {
     if(!calcTraces.length) return;
 
-    var barmode = gd._fullLayout.barmode || gd._fullLayout.waterfallmode;
+    var barmode = (gd._fullLayout.barmode !== undefined) ? gd._fullLayout.barmode : gd._fullLayout.waterfallmode;
```
etpinard: Hmm. This seems wrong. What happens on subplots with both bar and waterfall traces? In the case, both `barmode` and `waterfallmode` are present in `fullLayout` and here the waterfall traces would be computed using `barmode` instead of `waterfallmode`.

Please have a look at how `barpolar` handles this problem:

https://github.com/plotly/plotly.js/blob/7b2946a4c2c6cd6ab5691e5a1fc37b8809831287/src/traces/barpolar/calc.js#L76-L104

where we don't call `Bar.crossTraceCalc` direction, but instead we call `setGroupPositions` with a mock `gd` object containing the correct `*mode`, `*groupgap` and `*gap` attribute.

src/traces/waterfall/attributes.js
```diff
@@ -0,0 +1,113 @@
...
+    decreasing: directionAttrs(),
+
+    connector: {
+        color: lineAttrs.color,
+        width: lineAttrs.width,
+        dash: lineAttrs.dash,
+        mode: {
+            valType: 'enumerated',
+            values: ['steps', 'spanning', 'between', false],
+            dflt: 'spanning',
```
etpinard: Judging by a quick search for waterfalls in the wild, these connectors aren't too common. Would it be ok if we made the default here `false`?

etpinard: ... and actually, having `mode: false` isn't very plotly.js-esque. It would be best to have `connector.visible: false`

etpinard: TODO:

- [ ] add `connector.visible: true | false` , drop `connector.mode: false`
- [ ] drop `connector.mode: 'steps'` (which looks buggy)
- [ ] tweak `Bar.plot` so that it draws bars `connector.line.width / 2` bigger in `connector.mode: 'between'`

etpinard: - [ ] make sure `restyle(gd, 'connector.visible', true | false)` works + add jasmine test

etpinard: @nicolaskruchten @alexcjohnson any strong opinions on whether or not connectors should be turned on by default?

etpinard: - [ ] turn `connector.(color|dash|width)` into -> `connector.line.(color|dash|width)`

src/traces/waterfall/defaults.js
archmoj: Addressed in https://github.com/plotly/plotly.js/pull/3531/commits/6b3eeab43c5d7ea382bd47306fd355367df14a9b.

etpinard:
**Update:** branch https://github.com/plotly/plotly.js/compare/master...waterfall_tests-tmp has the latest (most bug-free) version of waterfall.

We're currently still testing a few things on that branch. We'll rebase those commits those tomorrow morning.

archmoj:
> **Update:** branch [master...waterfall_tests-tmp](https://github.com/plotly/plotly.js/compare/master...waterfall_tests-tmp) has the latest (most bug-free) version of waterfall.
> 
> We're currently still testing a few things on that branch. We'll rebase those commits those tomorrow morning.

@etpinard few more tests are added here: https://github.com/plotly/plotly.js/compare/waterfall_tests-tmp...waterfall_more-tests

etpinard:
:dancer: :dancer: :dancer: 

Awesome work!

PR #3532, opened 2019-02-10
Bump dev deps 2019-02-10

etpinard:
fixing this low-severity security alert

image

https://github.com/plotly/plotly.js/network/alert/package-lock.json/lodash/open

going from

image

to 

image

and bumping a few other dev deps along the way.

cc @plotly/plotly_js


package-lock.json
```diff
@@ -4047,7 +4110,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
archmoj: I've got a difference when I `npm i`. 
It seems we could remove this line (`"optional": true`) as well as few others below.
@etpinard please let me know if you wanted me to push a commit to remove those?

archmoj:
Just as another note, this is the only difference I noticed between the current master and the new publish, which should be OK.
screenshot from 2019-02-10 19-42-42

package-lock.json
etpinard: Which version of `npm` are you using? I'm on 6.7.0

etpinard:
> Just as another note, this is the only difference I noticed between the current master and the new publish, which should be OK.

Those are just randomly-generated string ids. This PR only upgrade dev dependencies, so yeah we're good.

package-lock.json
archmoj: My npm version is at 6.4.1

etpinard: Can you bump your npm version to 6.7.0 and try `npm i` again?

archmoj: Cool! With 6.7.0 I've got no difference.

archmoj:
@etpinard Thanks for the PR.
Well done. 
:dancer:

Issue #3526, opened 2019-02-08
SVG errors for heatmap with all NaNs 


etpinard:
From https://community.plot.ly/t/training-garbage-error-in-heatmap-with-nans/19265

See https://codepen.io/Squerut/pen/XOZWqB

PR #3534, opened 2019-02-11
Fix annotations SVG errors on trace-less subplots

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3526 - which was slightly misreported, this bug has nothing to do with heatmaps. All cartesian autoranged subplots w/o a single `visible:true` trace were affected.

cc @plotly/plotly_js looking a for a quick review. I'll like to include this in tomorrow's `v1.44.4`. Thanks!


archmoj:
Good fixes.
:dancer:
Thanks @etpinard.

PR #3536, opened 2019-02-12
Fix bug in scattergl text update

MPursche:
Hello,

This pull request fixes a bug in `scattergl` plots when text labels are used and the number of datasets is changed. Changing the number of `scattergl` plots resulted in an error because the number of `Text` objects was not adapted to the number of plots.

Greetings
Marcel


etpinard:
Brilliant PR. Amazing fix. Thanks very much!

PR #3366, opened 2018-12-21
Finance trace hover fixups

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3298 by making data coordinates match `gd.calcdata` items 1-to-1 and fixes a finance trace hover edge case - cc @plotly/plotly_js

PR #3537, opened 2019-02-12
Fix finance chart null item hover

jakubkottnauer:
Fixes the bug visible here: https://codepen.io/jakubkottnauer/pen/gqKgeQ

The issue happens when triggering a hover on finance charts via `Fx.hover` and when the element at the position is `null` - visually it works correctly but there are errors in console.


etpinard:
Very solid fix. Thanks very much!

etpinard:
Related: https://github.com/plotly/plotly.js/pull/3366

Issue #3541, opened 2019-02-13
[Feature] add trace name to be referenced from 'metasrc'


priyatharsan:
https://codepen.io/priyatharsan/pen/aXjYaB

VeraZab:
I also think that for rangeselector.buttons, it has not been implemented.
Is this an easy add on?

PR #3548, opened 2019-02-15
More layout meta support

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3541 by adding `layout.meta` support for:

- `trace.name`, both in legend items and hover labels
- `rangeselector` buttons `label`
- `updatemenus` buttons `label`
- `sliders` steps `label`

@VeraZab I also added `TEMPLATE_STRING_REGEX` to `Lib` as requested.

cc @plotly/plotly_js


antoinerg:
Original Codepen now fixed with `plotly.js` from this branch: https://codepen.io/antoinerg/pen/gqQbGK

antoinerg:
Thanks @etpinard for the fix!

However, I'm wondering if `layout.meta` is available in `hovertemplate`?

etpinard:
> However, I'm wondering if the metadata in `layout.meta` is accessible from `hovertemplate`?

Not yet.

antoinerg:
This is non-blocking, but I think it would be as simple as adding a 4th argument `{meta: meta}` when calling `Lib.hovertemplateString` on this line:
https://github.com/plotly/plotly.js/blob/1808fefef52ae6836fb2b1ecb4b4789b3342a144/src/components/fx/hover.js#L978 . If you do add it, an additional test would be needed.

If you prefer not to add `hovertemplate` support in this PR, this is ready to :dancer:

etpinard:
> as adding a 4th argument `{meta: meta}` when calling `Lib.hovertemplateString` on this line

If it's that easy, then _might as well_ -> https://github.com/plotly/plotly.js/pull/3548/commits/a0ff6a17ff2dbada075e80d246e1d217f91eea8b

antoinerg:
> If it's that easy, then _might as well_ -> [a0ff6a1](https://github.com/plotly/plotly.js/commit/a0ff6a17ff2dbada075e80d246e1d217f91eea8b)

Glad that it really was that easy and thanks for improving `hovertemplate` again :medal_sports: 

:dancer: once the tests pass on CircleCI!

PR #3549, opened 2019-02-15
Add midpoint attr to color scales

### LINKED ISSUE #3509: Color scale midpoint
Right now to use a diverging color scale, the user has to do some math outside of plotly.js to ensure that the midpoint is at an appropriate value in data-space. This math is actually almost impossible to do in the case of `histogram2d` as the data values are computed within plotly.js... I propose a new attribute `cmid` (or `zmid` as appropriate) which would make this easier. 

If `mid` is provided and neither `min` nor `max` then after the existing automatic computation, whichever of those two is closest to the midpoint will be pushed away by an appropriate amount. If either `min` or `max` is provided but not both, then the unspecified one will be set at a distance from `mid` equal to the distance of the specified one. If both `min` and `max` are provided, this setting does nothing.

Feedback before I start implementing?

nicolaskruchten:
Closes #3509 

LMK what kinds of extra tests you'd like or if you want me to just modify an existing mock or something, or if I've mangled some of the coercion machinery :)


nicolaskruchten:
I've love to get this into 1.45 if it's uncontroversial :)

etpinard:
I'm still 100% convinced `cmid` is the right name for this attribute. I'd like make this attribute consistent with a future axis attributes coerced when we have `rangemode: 'symmetric'`. Maybe `rangemid` would be better?

etpinard:
At present `rangemode: 'symmetric'` means an axis range centered around 0, but we could make that more general in the future.

nicolaskruchten:
I think we should stick with the current naming convention for right now, and when we do something `range`-related in the future we can carry `mid` along with it.

etpinard:
> the current naming convention

which is?

nicolaskruchten:
`marker.rangemid` is really... unintuitive, in the same way as `marker.showscale` is... what range? what scale?

nicolaskruchten:
> > the current naming convention
> which is?

cmin, cmax, cmid ...?

etpinard:
ok, but we're planning on changing `cmin` / `cmax` for `crange`

----------

Does anyone in @plotly/plotly_js have an opinion about this?

nicolaskruchten:
In future when we have `layout.coloraxis` I think we should use `axis`-like terminology like `range` etc for sure but I feel `cmid` fits in nicely with what's here atm.

That said, if `rangemid` is what it takes to get this merged in, I can change it. Might be odd in the `histogram2d` case where it's `zmin`/`zmax` at the trace level tho. `trace.rangemid` ?

nicolaskruchten:
> ok, but we're planning on changing `cmin` / `cmax` for `crange`

And the day we do that, we should change `cmid` also, is what I'm saying.

etpinard:
Making _changes_ to attribute is never fun. This is exactly what I'm trying to avoid.

nicolaskruchten:
What's the rationale for `cmin`/`cmax` -> `crange` ?

etpinard:
> What's the rationale for `cmin`/`cmax` -> `crange` ?

Bring color "axis" attribute closer to data axis attributes.

nicolaskruchten:
> Bring color "axis" attribute closer to data axis attributes.

I would leave the existing `cmin`/`cmax` attributes alone (and add `cmin`!), and then use the `range` terminology only in the future `layout.coloraxis` object. Then in V2 we can drop the trace-specific colorscale/colorbar stuff.

etpinard:
> I would leave the existing `cmin`/`cmax` attributes alone (and add `cmid`)!

If you really want to merge this ASAP, sure what can that!

Now, are you planning on fixing the tests (we should also add jasmine tests making sure `restyle` and `react` work ok in and out of `cmid`) or you want me to take this thing to the finish line?

nicolaskruchten:
I can finish it up, just wanted to get feedback on the approach, make sure I didn't miss anything obvious.

nicolaskruchten:
> If you really want to merge this ASAP, sure what can that!

Mostly as you say it's a pain to change existing attrs, so I think we should plan to leave the existing `cmin`/`cmax` alone for that reason.

nicolaskruchten:
So I'm trying to fix the tests and confused about why the `scattergeo` one is failing but there is `scattermapbox` code in the stack trace. I'm not sure why execution is going down that path... any hints?

etpinard:
> So I'm trying to fix the tests and confused about why the `scattergeo` one is failing

https://github.com/plotly/plotly.js/blob/0714286d8c485a55bd0666f4a45f853c4f5112e5/src/plots/geo/layout/layout_attributes.js#L96-L102

Thanks for uncovering a pretty embarrassing typo. Switching to `scattergeo` makes the `scattergeo_test.js` suite pass.

Now, for `scattermapbox`, you'll need to add `cmid` to 

https://github.com/plotly/plotly.js/blob/0714286d8c485a55bd0666f4a45f853c4f5112e5/src/traces/scattermapbox/attributes.js#L73-L99

as `scattermapbox` doesn't use the colorscale attribute [generator](https://github.com/plotly/plotly.js/blob/master/src/components/colorscale/attributes.js) (yet).

nicolaskruchten:
Thanks for looking into it @etpinard ! I'm confused about what the "typo" is in the code you linked above... can you clarify please?

etpinard:
> I'm confused about what the "typo" is in the code you linked above... can you clarify please?

Oops. Wrong link:

https://github.com/plotly/plotly.js/blob/25fa0c20691092b18435a59d53c24c9fb470b5e9/test/jasmine/tests/scattergeo_test.js#L93

nicolaskruchten:
Ahhh that makes much more sense, thank you :)

nicolaskruchten:
OK despite best intentions I havent found the time to write tests here unfortunately... could someone in @plotly/plotly_js take over the restyle/relayout tests plz? :)

etpinard:
Ok. I'll do that this morning.

src/components/colorscale/attributes.js
```diff
@@ -160,6 +162,21 @@ module.exports = function colorScaleAttrs(context, opts) {
         ].join('')
     };
 
+    attrs[mid] = {
+        valType: 'number',
+        role: 'info',
+        dflt: null,
+        editType: editTypeOverride || 'plot',
```
nicolaskruchten: @etpinard I wasn't sure what to put for `editType` and `impliedEdits` btw so I copied over the settings from `max` above... maybe it should match `auto` instead? Or be something else? Likely your tests will expose any problems but I just wanted to shine a light here.

etpinard: Restyle got fixed in https://github.com/plotly/plotly.js/pull/3549/commits/23c1cf0832a585a47b4231dbd9e9adfd785d382a

Is that the behaviour you had in mind?

test/jasmine/tests/plot_api_test.js
```diff
@@ -1195,6 +1195,77 @@ describe('Test plot api', function() {
             .then(done);
         });
 
+        it('turns on cauto when cmid is edited', function(done) {
```
nicolaskruchten: this is an odd description... `cmid` only has effect if `cauto` is true, but doesn't turn on `cauto` per se, right?

src/components/colorscale/defaults.js
```diff
@@ -33,9 +33,14 @@ module.exports = function colorScaleDefaults(traceIn, traceOut, layout, coerce,
...
     var validMinMax = isNumeric(minIn) && isNumeric(maxIn) && (minIn < maxIn);
-    coerce(prefix + cLetter + 'auto', !validMinMax);
-    coerce(prefix + cLetter + 'min');
-    coerce(prefix + cLetter + 'max');
+    var auto = coerce(prefix + cLetter + 'auto', !validMinMax);
+
+    if(auto) {
+        coerce(prefix + cLetter + 'mid');
+    } else {
+        coerce(prefix + cLetter + 'min');
```
nicolaskruchten: this line and the next used to always run, but now only conditionally... might this break anything? I'm not sure about the implication of not calling `coerce()`.

src/components/colorscale/attributes.js
nicolaskruchten: I must admit I don't fully grok the various nuances of the types of edits, but the test reads fine at first glance!

src/components/colorscale/defaults.js
etpinard: No tests are failing because of it.

test/jasmine/tests/plot_api_test.js
etpinard: > but doesn't turn on `cauto` per se, right?

yes, it has to. If not, `restyle(gd, 'marker.cmid', 0)` wouldn't work.

etpinard: by "work" here I mean it wouldn't apply that new cmid value when `cmin` and `cmax` are set.

nicolaskruchten: OK I see what you mean.

src/components/colorscale/defaults.js
nicolaskruchten: ok

nicolaskruchten: There is a test failure that mentions `coerce`:

> Test surface supplyDefaults should coerce 'c' attributes with 'c' values regardless of `'z' if 'c' is present FAILED

etpinard: Ah good eye.

Well `cmin` and `cmax` have no effect

https://github.com/plotly/plotly.js/blob/bb1239a31a56d72ec664091ca7e1c0b42adb8b2f/src/components/colorscale/calc.js#L31-L37

when `cauto` is true.

nicolaskruchten: I'm easy. Like I said, I don't know the impact of calling `coerce` and ignoring the attr vs not calling `coerce` :)

etpinard: Calling the behaviour correct and adapting the test in https://github.com/plotly/plotly.js/pull/3549/commits/ff6781226bf5717efdcf0c9041b9e5f0b6e7abc6

etpinard:
@plotly/plotly_js can I get a review for this thing?

archmoj:
> @plotly/plotly_js can I get a review for this thing?

@etpinard - sure, I'll review this.

etpinard:
> However; it looks`cmin` has no effect in the mentioned example when `cmid` is defined?

Right, that's the expected behaviour it seems like:

https://github.com/plotly/plotly.js/blob/ff6781226bf5717efdcf0c9041b9e5f0b6e7abc6/src/components/colorscale/attributes.js#L171-L177

archmoj:
> Right, that's the expected behaviour it seems like:

Then it may be reasonable to modify the description of `cmin` and `cmax`? to include for example:
<pre>'Has no effect when `cmin` is defined.'</pre>.

etpinard:
> Then it may be reasonable to modify the description of `cmin` and `cmax`? to include for example:
> 
> 'Has no effect when `cmid` is defined.'

That's not true. Say you have:

```
{
  cmin: 0,
  cmax: 2,
  cmid: 2
}
```


then here `cmid` is _defined_, but ignored.

archmoj:
Nicely done.
:dancer:

nicolaskruchten:
Yes the idea is that if you want to specify two points on the curve they must be min/max... not min/mid or mid/max. Easier to implement and document. The mid doc is a bit indirect in that it says “no effect if auto false” and then the auto doc says “false if min and max are set” but I figure this is still correct/readable.

PR #3426, opened 2019-01-10
d3-sankey-circular with grouping

antoinerg:
Very rough attempt at closing #3321 

This prototype implements grouping via a new `group` attribute. For now, it can be called via `Plotly.restyle`.

This PR is based off PR https://github.com/plotly/plotly.js/pull/3406

Live demo: https://codepen.io/anon/pen/gZBMKp

Issue #3321, opened 2018-12-11
Sankey: Grouped nodes + "zoom-like" interactions


antoinerg:
Add the ability to define a set of nodes to be a "group", then there is an interaction where you can collapse this group into a single composite node, and re-expand it later.

antoinerg:
Partially closed by #3556 

Opening an issue for the "zoom-like" and mouse interactions: https://github.com/plotly/plotly.js/issues/3582

PR #3556, opened 2019-02-19
Sankey: group nodes

antoinerg:
Closes part of #3321 by implementing grouping of nodes via a new `node.groups` attribute. For now, it can be called via `Plotly.restyle`.

It is based off a recent `master` and it supersedes PR https://github.com/plotly/plotly.js/pull/3426

Checklist:
- [x] Remove transitions on first render
- [x] Disallow nodes to be in more than one group


antoinerg:
Here's a Codepen showcasing the new mock introduced in this PR: https://codepen.io/antoinerg/pen/MLxjmB

src/snapshot/helpers.js
```diff
@@ -15,6 +15,7 @@ exports.getDelay = function(fullLayout) {
     return (
         fullLayout._has('gl3d') ||
         fullLayout._has('gl2d') ||
+        fullLayout._has('sankey') ||
```
etpinard: Why do we need this?

src/traces/sankey/constants.js
```diff
@@ -15,8 +15,8 @@ module.exports = {
     sankeyIterations: 50,
     forceIterations: 5,
     forceTicksPerFrame: 10,
-    duration: 500,
-    ease: 'cubic-in-out',
+    duration: 350,
+    ease: 'quart-in-out',
```
etpinard: Why?

src/traces/sankey/attributes.js
```diff
@@ -89,6 +89,16 @@ var attrs = module.exports = overrideAll({
             role: 'info',
             description: 'The shown name of the node.'
         },
+        groups: {
+            valType: 'data_array',
```
etpinard: It might be better to declare this as an `info_array` with `dimensions: 2` similar to 

https://github.com/plotly/plotly.js/blob/bb1239a31a56d72ec664091ca7e1c0b42adb8b2f/src/components/grid/index.js#L54-L69

that way wouldn't need to worry about its structure during `sankey/calc.js`.

src/traces/sankey/calc.js
```diff
@@ -17,8 +17,10 @@ var isIndex = Lib.isIndex;
 var Colorscale = require('../../components/colorscale');
 
 function convertToD3Sankey(trace) {
-    var nodeSpec = trace.node;
-    var linkSpec = trace.link;
+    // var nodeSpec = trace.node;
+    // var linkSpec = trace.link;
+    var nodeSpec = Lib.extendDeep({}, trace.node);
+    var linkSpec = Lib.extendDeep({}, trace.link);
```
etpinard: Do we really need to extend them? This can be slow.

src/snapshot/helpers.js
antoinerg: The opacity of links is now zero on `enter()` which is why we need to wait a bit here. I understand this is not a super clean solution and I could either:
- remove animations altogether
- remove animations when saving to image

src/traces/sankey/calc.js
```diff
@@ -34,7 +36,32 @@ function convertToD3Sankey(trace) {
...
+        var group = groups[i];
+        // Build a lookup table to quickly find in which group a node is
+        if(Array.isArray(group)) {
+            for(j = 0; j < group.length; j++) {
+                var nodeIndex = group[j];
+                var groupIndex = nodeCount + i;
+                groupLookup[nodeIndex] = groupIndex;
+            }
+        } else {
+            Lib.warn('node.groups must be an array, default to empty array []');
```
etpinard: If we declare `groups` an `info_array`, we wouldn't need this. `Plotly.validate` would be enough to catch bad inputs.

src/traces/sankey/render.js
```diff
@@ -67,6 +67,30 @@ function sankeyModel(layout, d, traceIndex) {
         Lib.warn('node.pad was reduced to ', sankey.nodePadding(), ' to fit within the figure.');
     }
 
+    // Create transient nodes for animations
+    Object.keys(calcData._groupLookup).forEach(function(nodePointNumber) {
```
etpinard: `for(var k in calcData._groupLookup)` is generally faster.

src/traces/sankey/render.js
```diff
@@ -352,8 +376,11 @@ function nodeModel(d, n, i) {
     var visibleThickness = n.dx;
     var visibleLength = Math.max(0.5, n.dy);
 
-    var basicKey = n.label;
-    var key = basicKey + '__' + i;
+    var key = 'node_' + n.pointNumber;
+    // If it's a group, it's mutable and should be unique
+    if(n.group) {
+        key = 'group_' + Math.floor(1e12 * (1 + Math.random()));
```
etpinard: Use 

https://github.com/plotly/plotly.js/blob/7da95e1af39b4b1f98abc2a43deb6b117435931d/src/lib/index.js#L261-L307

instead.

src/traces/sankey/constants.js
antoinerg: `duration: 350` for it to complete before saving to image
`ease: 'quart-in-out'` just makes it much easier to follow the trajectories of nodes on relayout/grouping. I guess, we should have a discussion about animations.

src/traces/sankey/render.js
```diff
@@ -701,13 +732,17 @@ module.exports = function(gd, svg, calcData, layout, callbacks) {
         })
         .style('stroke-width', function(d) {
             return salientEnough(d) ? d.linkLineWidth : 1;
-        });
+        })
+        .attr('d', linkPath());
```
etpinard: Did moving this `attr()` call down to the merge selection fix any bugs?

src/snapshot/helpers.js
etpinard: Oh we have an animation on `enter()` now. Was this a requirement?

If not, I'd vote :hocho: that out for now. Animation on first-draw is something that we'll need to add at some point, but we should do that for traces, not just for `sankey`.

src/traces/sankey/render.js
```diff
@@ -79,7 +79,7 @@ function sankeyModel(layout, d, traceIndex) {
             }
         }
 
-        graph.nodes.push({
+        graph.nodes.unshift({
```
etpinard: Cool. Can we add a test to :lock: this down?

src/traces/sankey/constants.js
etpinard: Ha I see. Voting :-1: on animations.

src/traces/sankey/attributes.js
antoinerg: I didn't realize we had `info_array`. Did you mean `dimensions: 1`? 

For example: the following seem to work fine:
```javascript
        groups: {
            valType: 'info_array',
            dimensions: 1,
            freeLength: true,
            items: {valType: 'data_array', dflt: [], editType: 'calc'},
            role: 'calc',
            description: [
                'Groups of nodes.',
                'Each group is defined by an array with the indices of the nodes it contains.',
                'Multiple groups can be specified.'
            ].join(' ')
        },
```

etpinard: Why do the inner items need to be `data_array`?

src/traces/sankey/render.js
antoinerg: Yes it did. After a grouping operation, some links stay in the figure needs (they don't `enter` the scene) but their `d` attribute still need to be updated.

src/traces/sankey/attributes.js
antoinerg: Oh I see! The following works fine too:
```javascript
        groups: {
            valType: 'info_array',
            dimensions: 2,
            freeLength: true,
            items: {valType: 'number', editType: 'calc'},
            role: 'calc',
            description: [
                'Groups of nodes.',
                'Each group is defined by an array with the indices of the nodes it contains.',
                'Multiple groups can be specified.'
            ].join(' ')
        },
```

etpinard: > The following works fine too:

Nice. I'd vote for this configuration. I don't think `groups` is really a "data" array.

antoinerg: Done in 1cd206f7fdba75b010650331fc1f7a726ec488df

src/traces/sankey/calc.js
antoinerg: No, we actually don't need to do that! 

:hocho: in 1cd206f7fdba75b010650331fc1f7a726ec488df

src/traces/sankey/calc.js
antoinerg: Nice!  :hocho: in 1cd206f7fdba75b010650331fc1f7a726ec488df

src/traces/sankey/render.js
antoinerg: Done in 1cd206f7fdba75b010650331fc1f7a726ec488df  :racehorse:

src/traces/sankey/render.js
antoinerg: I didn't know we had this. I could also reuse it elsewhere in Sankey as can be seen in 1cd206f7fdba75b010650331fc1f7a726ec488df !

src/snapshot/helpers.js
antoinerg: Out of curiosity, is there an easy way to tell that we're in the process of creating an image and force the opacity of nodes and links to be `1` if that's the case?

etpinard: You can use `gd._context.staticPlot`

src/traces/sankey/constants.js
antoinerg: About animations: in this [Codepen](https://codepen.io/antoinerg/pen/MLxjmB), toggling between `Ungroup [[]]` and `Group [[[0,2]]]`  shows where the animations can shine. It really helps to see what nodes are being merged/unmerged.

I'm also on the fence about enabling them because of the timeout I had to add to `src/snapshot/helpers.js`. On the other hand, I think the animations do add value and can look great.

Mabye @alexcjohnson has an opinion on this one?

BTW, we can simply comment out the couple `.style('opacity', 0)` calls to get rid of the new animations added in this PR so no additional work will be required either way.

etpinard: Animation for sankey trace will be useful, no doubt. But I don't think they should be triggered from plain `restyle` calls. Our other traces don't do that. I think we should try to make sankey animations follow how `scatter` animations behave.

@antoinerg you can try playing around with `transitionOpts` and `makeOnCompleteCallback` as  in [`Scatter.plot`](https://github.com/plotly/plotly.js/blob/7da95e1af39b4b1f98abc2a43deb6b117435931d/src/traces/scatter/plot.js#L25). You'll need those to get animations to end properly (i.e. return a Promise) and have access to the transition options.

src/snapshot/helpers.js
antoinerg: I removed the delay in `src/snapshot/helpers.js` while still passing the image tests

See commit b90fda1 and b4d7ff3

src/traces/sankey/constants.js
antoinerg: @etpinard If you interact with the menu in mock `sankey_energy` you will see animations on `restyle`. If I :hocho: animations, can those ones also be removed? Would you consider that a breaking change?

etpinard: Arrg. We'll have to keep the existing animations. Otherwise yes, that would be a breaking change.

etpinard:
1.45.0 is getting delayed (due to my buggy PRs), so let's try get this in 1.45.0!

etpinard:
@antoinerg let me know if you need help to finish up this PR today.

src/traces/sankey/render.js
```diff
@@ -701,13 +744,17 @@ module.exports = function(gd, svg, calcData, layout, callbacks) {
         })
         .style('stroke-width', function(d) {
             return salientEnough(d) ? d.linkLineWidth : 1;
-        });
+        })
+        .attr('d', linkPath());
 
-    sankeyLink.transition()
-       .ease(c.ease).duration(c.duration)
-       .attr('d', linkPath());
```
antoinerg: Animating the SVG path for links would look odd. Transitioning from a normal link to a circular link one for example would look broken. It makes more sense for the links to disappear and reappear, hence the change here...

etpinard:
@antoinerg can you fix:

image

?

src/traces/sankey/render.js
etpinard: Nicely done!

src/traces/sankey/constants.js
etpinard: Resolved!

src/traces/sankey/render.js
etpinard: Have you tried adding a test for this?

src/traces/sankey/render.js
etpinard: Same here. We should try :lock: this down.

etpinard:
Awesome work @antoinerg 

Looks like this PR only needs two more interaction tests:

- https://github.com/plotly/plotly.js/pull/3556#discussion_r260032672
- https://github.com/plotly/plotly.js/pull/3556#discussion_r260032570

and it will be ready to :rocket:

src/traces/sankey/render.js
antoinerg: > Have you tried adding a test for this?

No, I haven't. What specifically would you test?

etpinard: that an updated node gets the correct `'d'` string.

src/traces/sankey/render.js
antoinerg: Done in ffbafa7

src/traces/sankey/render.js
antoinerg: Done in ffbafa7

antoinerg:
@etpinard I added tests in ffbafa7 as requested. In the process of writing them, I encountered a show-stopping bug for certain combinations of nodes: for example in mock `sankey_energy` if you try to group nodes 1 and 3:
```javascript
Plotly.restyle(gd,{'node.groups':[[[1,3]]]})
```
CPU usage goes to 100% and the operation never ends. The issue seems limited to `d3-sankey`: it works fine if the underlying Sankey generator is `d3-sankey-circular`. One option, is to force the use of `d3-sankey-circular` when a `node.groups` is defined and not empty. That way, we would not have any breaking changes.

I would still like to investigate why exactly `d3-sankey` chokes the way it does for some combinations of nodes :thinking:

antoinerg:
The problem mentionned in https://github.com/plotly/plotly.js/pull/3556#issuecomment-467257612 is now fixed in 7094887584b7bb9a3e2ca17fef03deba138858f6.

The problem was that the check for circularity was done prior to grouping nodes. Grouping certain combination of nodes (see 1.) can create a circularity and `d3-sankey` cannot process the resulting graph (hence the infinite loop). By checking for circularity after grouping, we can appropriately switch to `d3-sankey-circular`.

1. This is obivous: if we have 1 -> 2 -> 3 and we group 1 and 3, we get 2 <-> 3 which now has a circular link.

etpinard:
Nicely done :dancer:

Issue #3560, opened 2019-02-21
Using matches when opposite axis is categorical causes error


nicolaskruchten:
This pen shows the problem: https://codepen.io/nicolaskruchten/pen/gqyyVL?editors=0010

nicolaskruchten:
Note that this only happens with the `x` axis AFAICT... doing the same trick with `x` doesn't fail.

nicolaskruchten:
Hmm doesn't even seem related to grid per se.

nicolaskruchten:
Here's the same problem with scatter, no grid, on Y: https://codepen.io/nicolaskruchten/pen/LqvoNV?editors=0010

nicolaskruchten:
OK so actually it's not just `y` ... same thing happens on `x` if I swap everything! Basically if the non-shared axis is categorical it doesn't work, but if both are linear it does, in both permutations! See https://codepen.io/nicolaskruchten/pen/LqvoNV?editors=0010

etpinard:
That's the expected behaviour. Matching axes only work on axes of the same type.

nicolaskruchten:
No... in this case I'm matching continuous to continuous, but the OTHER axis is categorical.

nicolaskruchten:
i.e. both `y`s are continuous, and the `x` is categorical, and I'm trying to match `y2` to `y`.

etpinard:
Ah I see. On it!

PR #3565, opened 2019-02-21
Fix category init for non-matching axes on graphs with matching axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3560 - a fixup of https://github.com/plotly/plotly.js/pull/3506

cc @plotly/plotly_js 

@nicolaskruchten updated codepens:

- https://codepen.io/etpinard/pen/MLdLQZ?editors=0010
- https://codepen.io/etpinard/pen/VgOgXj?editors=0010


src/plots/cartesian/set_convert.js
```diff
@@ -575,13 +575,21 @@ module.exports = function setConvert(ax, fullLayout) {
 
     // should skip if not category nor multicategory
     ax.clearCalc = function() {
+        var init = function() {
```
archmoj: For this function name, it may be better to use something like `doClear` or `emptyCategories` since it is not initialised by default on top of the function.

src/plots/cartesian/set_convert.js
```diff
@@ -598,14 +606,14 @@ module.exports = function setConvert(ax, fullLayout) {
                         ax._categories = categories;
                         ax._categoriesMap = categoriesMap;
                     } else {
-                        ax._categories = [];
-                        ax._categoriesMap = {};
+                        init();
                     }
                 }
+
+                if(!found) init();
```
archmoj: I am wondering about the position of this line in relation to the `var found = false;` and wanted to double check something:
With this setup if the item is not found in the first iteration (when i === 0), the `init` function would be called. If that is correct, then `var found = false;` could be moved inside the `i` loop?

src/plots/cartesian/set_convert.js
etpinard: Good call. Done in https://github.com/plotly/plotly.js/pull/3565/commits/598a67bce659e737ecb4dfe11a645135b60f31a5

src/plots/cartesian/set_convert.js
etpinard: You're right! Thanks for spotting that. 

Done in https://github.com/plotly/plotly.js/pull/3565/commits/598a67bce659e737ecb4dfe11a645135b60f31a5 + I added a `break` as one `ax` can only be in one match group at a time.

archmoj:
Looks good.
:dancer:

Issue #3561, opened 2019-02-21
Automargin regression since 1.44.2 ?


nicolaskruchten:
See this pen: https://codepen.io/nicolaskruchten/pen/yZrdKO?editors=0010 

It takes a long time to run (almost hangs in Chrome/OSX!) and then the solution is to add a bunch of padding at the bottom, which didn't happen on 1.44.2, as in this pen: https://codepen.io/nicolaskruchten/pen/QYPeWE

nicolaskruchten:
Maybe related to https://github.com/plotly/plotly.js/pull/3510 ?

etpinard:
Why does `xaxis2` have a title in your codepen?

Is this the desired behaviour: https://codepen.io/etpinard/pen/qgGxaW?editors=0010

nicolaskruchten:
> Why does `xaxis2` have a title in your codepen?

it doesn't need one.

> Is this the desired behaviour: https://codepen.io/etpinard/pen/qgGxaW?editors=0010

Yes. But if you're using a template (which plotly.py and Chart Studio do by default) then you get `automargin: true` on all axes by default. Turning `automargins` on on both here even without the title makes the plot short again.

etpinard:
Cool. On it!

destradafilm:
> See this pen: https://codepen.io/nicolaskruchten/pen/yZrdKO?editors=0010
> 
> It takes a long time to run (almost hangs in Chrome/OSX!) and then the solution is to add a bunch of padding at the bottom, which didn't happen on 1.44.2, as in this pen: https://codepen.io/nicolaskruchten/pen/QYPeWE

I'm seeing this same bug with automargin and I'm on v1.46.0, I even tested in v1.47 and still get it too. I think this hasn't been fixed.

etpinard:
> I'm seeing this same bug

By _this same bug_, you mean the page hangs?

destradafilm:
yeah, I've been trying to recreate it but seems to need the perfect conditions to break. Even the original codepen submitted by @nicolaskruchten breaks sometimes for me. 

can you test this one? it has a forced width / height (in pixels) to what the width / height in the original codepen (in vh) was rendered to on my browser. 

https://codepen.io/destrada/pen/PgBroN?editors=1010

Eventually I get 
`plotly.min.js:1 Uncaught RangeError: Maximum call stack size exceeded`

[...]

PR #3566, opened 2019-02-21
Fix automargin regression for superimposed subplots

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3561 - a fixup of https://github.com/plotly/plotly.js/pull/3510

cc @plotly/plotly_js 

@nicolaskruchten see updated codepen:

- https://codepen.io/etpinard/pen/Gzaavr?editors=0010


antoinerg:
Looks good to me :dancer: ! 

Maybe @archmoj should have a look to be extra sure :eye:

src/plots/cartesian/axes.js
```diff
@@ -2647,11 +2648,37 @@ function drawDividers(gd, ax, opts) {
...
+            _offset: gs.l + (ax.position || 0) * gs.w,
+            _length: 0
+        };
+    }
+
+    if(side === 'top' || side === 'left') {
+        return anchorAxis._offset;
+    } else if(side === 'bottom' || side === 'right') {
+        return anchorAxis._offset + anchorAxis._length;
+    }
```
archmoj: Should this function `return` a default value if both conditions were false?

etpinard: That should never happen.

PR #3571, opened 2019-02-23
Detect and increase pixelRatio in 3D plots when anti-alias is not enabled by hardware

### LINKED ISSUE #3387: 3D graph axis rendering in FF64 + mac
I've noticed graphs that look perfectly fine in Chrome look "dodgy" in Firefox. I'm not sure if it is to do with antialiasing or what but the labels and axis look many times better in Chrome.

Take the documentation's 3D surface graph for example: https://codepen.io/plotly/pen/MaxrwW

Firefox:
screenshot 2019-01-03 10 31 06

Chrome:
screenshot 2019-01-03 10 31 31

Is there anything that can be done to improve this? Firefox definitely gets a less polished result. While it looks "ok" in the screenshots, if you actually interact with the graphs Firefox is even worse.

archmoj:
Fixes #3387 and would also improve the quality of images produced using `saveImage` & by CI.
More description - coming soon.


archmoj:
Closing in the favour of #3573 which was able to address more bugs while keeping more baselines unchanged.

Issue #3559, opened 2019-02-20
Error using scattergl when plot has text and non-text data


Jeffrey72:
https://codepen.io/anon/pen/JxVZeR?editors=1111

Reproduction:
In the codepen select some points using the lasso tool, and then click a zoom button. 

Notable features of the codepen are that both data traces have the type 'scattergl' and only one of the two traces has a mode that contains 'text'.
(I used Chrome to reproduce)

Bug: 
The traces disappear, and there is an error in the dev-tools console:
Uncaught TypeError: Cannot read property 'color' of undefined
    at styleTextSelection (plotly-latest.js:173530)
    at plotly-latest.js:173210
    at Array.forEach (<anonymous>)
    at Object.plot (plotly-latest.js:173208)
    at plotOne (plotly-latest.js:117826)
    at Object.exports.plot (plotly-latest.js:117750)
    at exports.drawData (plotly-latest.js:110083)
    at Object.lib.syncOrAsync (plotly-latest.js:99315)
    at relayout (plotly-latest.js:106232)
    at wrappedEdit (plotly-latest.js:106760)

Note: 
A naive solution may just be to check if the textOptions at the current index is undefined before continuing.

PR #3575, opened 2019-02-25
Call styleTextSelection only on hasText traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3559

cc @plotly/plotly_js you can compare old vs updated in codepens:

- old: https://codepen.io/anon/pen/JxVZeR?editors=1111
- updated: https://codepen.io/etpinard/pen/YBbova?editors=1111


archmoj:
Nicely done.
:dancer:

Issue #1851, opened 2017-07-04
Persistent Selection Handles and Draggable Box Select


chriddyp:
In addition to customized selection styles (https://github.com/plotly/plotly.js/issues/1847) and persistent selection data (https://github.com/plotly/plotly.js/issues/1848), it'd be nice if selection box widget could be made persistent and modifiable. 

Crossfilter.js has a great example of this:

![image](https://user-images.githubusercontent.com/1280389/27840065-0b36f286-60c4-11e7-95d0-b59b11bf4c57.png)

etpinard:
Like https://github.com/plotly/plotly.js/issues/181 ?

monfera:
@etpinard having read the original post https://github.com/plotly/plotly.js/issues/181#issue-126633751 it looks like that was more about specific issues back then. The description doesn't mention the above select box resizing. Maybe that issue was renamed to reflect intention for working on it; let me rename it back so it sticks to its original feedback and let's continue with this this issue for the select box resizing manipulation.

siakouchine:
Hello Plotly team, do you have any news about this issue? The example of behavior defined in Crossfilter in terms of persistent selection handles and draggable box is exactly what we are looking for for our plots. Do you think it could be implemented shortly?

mlisthenewcool:
Hi, any news about this issue from Plotly team ? Something like [brushable-scatterplot-matrix](https://observablehq.com/@d3/brushable-scatterplot-matrix) would be awesome !
Did someone managed to find a way around this issue ?

nicolaskruchten:
This isn't on our roadmap at the moment but we're still very interested in this feature if someone wants to work on some pull requests or sponsor it!

gvwilson:
Hi - this issue has been sitting for a while, so as part of our effort to tidy up our public repositories I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our stack. Cheers - @gvwilson

PR #3577, opened 2019-02-25
Clear selection outline during 'axrange' relayout calls

etpinard:
Fixing an unreported regression from some past axis-range relayout performance work.

Currently, selecting a region then clicking on the modebar zoom in/out buttons does not clear the selection outlines:

peek 2019-02-25 12-35

this PR fixes that.

cc @plotly/plotly_js and https://github.com/plotly/plotly.js/issues/1851 were this behaviour will be augmented.


archmoj:
Thanks for the fix. @etpinard Would you be interested to add [this codepen](https://codepen.io/MojtabaSamimi/pen/EMxZWM?editors=1010) to the PR description? 
Nicely done. :dancer:

Issue #2334, opened 2018-02-05
Missing data in scattergl plot when I go juste above 10000 points (regression)


mlaily:
Hi,

I just updated plotly.js from 1.30 to the latest 1.33, and my plot is broken.

I’m rendering about 75000 points with a scattergl plot, and after the update, it looks like most of the points are not rendered at all. What is odd though, is that when I hover over the plot, I can still see popups showing when my mouse is over places where points are supposed to be.

After trying different things, I noticed everything is rendering fine up to 10000 data points. If I go just above, say 10001, my plot is broken and most of the points are not rendered!

Everything worked fine with plotly.js 1.30.

I made two versions of my code, one with 10000 points, and the other with 10001. See for yourself: https://0.x2a.yt/other/private/plotly-test/10000.html
https://0.x2a.yt/other/private/plotly-test/10001.html

(This bug report is a copy/paste from the following forum topic: https://community.plot.ly/t/regression-missing-data-in-scattergl-plot-when-i-go-juste-above-10000-points/8152)

etpinard:
cc @dfcreative

dfcreative:
That is a very good edge case for [snap-points-2d](https://github.com/gl-vis/snap-points-2d), thank you @mlaily. In 1.30.0 we had _gl-vis fancy-scattergl_, which has no points snapping enabled, therefore points were rendered to the plot directly, although slower than now in terms of interactions.

Now at 1e5 points the quadtree algorithm for points clustering gets triggered, which is good for cases of randomish distributions, but not good for linearly aligned points.

We can consider increasing [`TOO_MANY_POINTS`](https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/index.js#L41) constant to defer optimized rendering mode, but essentially that won't fix the problem.

Alternately, that can be addressed in upcoming _point-cluster_ component, although we would have to think how to detect proper clustering mode.

Alternately, we can provide `snap` flag or number for `scattergl` trace, disabling/thresholding point snapping.

mlaily:
Haha, you are welcome! :D

If I understand correctly, fixing the general case seems hard, but I think I would be happy with an option to disable clustering altogether.

That said, maybe I'm not using the optimal representation for my data. Would you have any advice by any chance?
The data set is my last.fm listening history, and I'm trying to reveal interesting trends over time.
I first thought using a heat map would be the best representation, but I found it not detailed enough, or lacking the webgl performance if I use too many points...
[...]

PR #3578, opened 2019-02-25
No scattergl clustering below 1e5 pts

etpinard:
An attempts at fixing https://github.com/plotly/plotly.js/issues/2334 by skipping point clustering (done by [`point-cluster`](https://github.com/dy/point-cluster)) for `scattergl` traces with less than 1e5 pts.

Previously, we were subject to the `regl-scatter2d` default value of [`1e4`](https://github.com/gl-vis/regl-scatter2d/blob/091a1a70296dbb81cea4a320812e62ba6e928353/scatter.js#L213) meaning `scattergl` traces with more than 1e4 pts but less than 1e5 pts (our [`TOO_MANY_POINTS`](https://github.com/plotly/plotly.js/blob/13b925cc2c946fddf78dd0201ace5f9e986de545/src/traces/scattergl/index.js#L114-L116) constant) were clustered [from `regl-scatter2d`](https://github.com/gl-vis/regl-scatter2d/blob/091a1a70296dbb81cea4a320812e62ba6e928353/scatter.js#L547-L553) whereas traces with more than 1e5 pts were clustered from `Scattergl.calc`. 

With the PR, I noticed that panning is slightly slower for graphs 1e4 < pts < 1e5, but otherwise performance appears the same. I think this small performance lost is worth it to get the correct rendering and to make us call `point-cluster` from one place i.e. from `Scattergl.calc`.

Here are some updated codepens from https://github.com/plotly/plotly.js/issues/2334:

- https://codepen.io/etpinard/pen/RdwrJy
- https://codepen.io/etpinard/pen/Ygzwve
- https://codepen.io/etpinard/pen/VRwedy?editors=1010

We may eventually still want to lower the `maxDepth` clustering option as @dy has suggested in https://github.com/plotly/plotly.js/issues/2334#issuecomment-465784675 to potentially correct rendering of traces with > 1e5 pts, but I'd like to wait to find a currently failing example before doing so. We could also try to bring back clustering for 1e4 < pts < 1e5 traces together with a lower `maxDepth` down the road.

cc @plotly/plotly_js I'd like to slide to in the 1.45.0 - as this PR may have consequences I don't want to release in a patch version.


archmoj:
Awesome fix! 
[Before](https://codepen.io/MojtabaSamimi/pen/MxWoyV?editors=0010)
[After](https://codepen.io/MojtabaSamimi/pen/ZPEKKK?editors=0010)
Thanks @etpinard.
Concerning too many line numbers in `gl2d_no-clustering.json` and also in order to improve the mock and baseline test with points with opacity, would you please replace it with the data printed to console [here](https://codepen.io/MojtabaSamimi/pen/ZPEKKK?editors=0010)?

etpinard:
@archmoj I like the way `gl2d_no-clustering.json` easily demonstrated the fix (i.e. one pt _was_ missing, not it is _not_), so instead of replacing it, I simply added a new mock in https://github.com/plotly/plotly.js/pull/3578/commits/f6e8918ca7baa8afb94b639cf3d33136dde77aff

I hope you don't mind.

archmoj:
Excellent! Thanks.
:dancer:

Issue #2520, opened 2018-04-03
Feature Request: Sankey Node Default Order


baraaorabi:
I think it would be cool to be able to force an ordering on the nodes of a sankey diagram. I imagine the order in which nodes appear in the label list of a sankey data object would be a reasonable choice.

davhin:
I would also highly value the ability to order the nodes in a sankey diagram.

davhin:
@antoinerg I see you have assigned yourself to this issue, thank you for that! Since I have no  JS-experience and cannot help out on the coding side is there anything I can do?

skylogic004:
I agree, this would be very useful. 

In my case, I have about 25 target nodes and plotly puts most of them in the correct order that I would like, but without fail it always has to put 2 of them in some random place, breaking the cleanliness of the plot. Manual ordering would help me fix this.

Also, in my case, the targets are unique to a source, meaning that there should never be a need for connections to overlap (which I suppose is a different issue, as whatever mechanism is used to auto-sort nodes really should be able to prevent overlap when it's not necessary).

jwilson8767:
@antoinerg 
Possibly of use: https://stackoverflow.com/questions/33461868/ordering-nodes-in-d3-sankey-diagram
And even more recently: https://github.com/d3/d3-sankey/pull/53

antoinerg:
Also, although we can drag nodes within a column, this isn't reflected in the state so on save/reload, export etc this info is lost. This position needs to be part of the state! Ordering should also work.

ndeku:
Agree this would add value to the Sankey diagrams in Plotly. Without consistent node ordering the diagrams are not reliably reproducible.

Mahdis-z:
The new feature `node.(x|y)` sets the position of nodes in [Sankey](https://plot.ly/python/sankey-diagram/#define-node-position). Please follow this [ issue](https://github.com/plotly/plotly.js/issues/4373) for `sorting` feature.

Issue #3563, opened 2019-02-21
Sankey nodes jump on dragging


alexcjohnson:
When you drag a sankey node, on first mouse move it jumps vertically (or horizontally if the diagram is horizontal). Bug exists on master but not in the latest release 1.44.4
![sankey-drag-jump](https://user-images.githubusercontent.com/2678795/53183247-f4e6ab80-35c8-11e9-9924-e17e391efd3c.gif)
cc @antoinerg

antoinerg:
Thank you @alexcjohnson for reporting this issue. It has to do with a change of coordinates that is done prior to running the force simulation.

However, I think I resolved this issue at some point. Did you run `npm i` before starting your devtools?

PR #3583, opened 2019-02-26
Sankey: add attributes `node.(x|y)` 

antoinerg:
This closes parts of https://github.com/plotly/plotly.js/issues/2520.

It adds attributes `node.(x|y)` to specify the position of nodes. 

Upon reordering the nodes by dragging them with the mouse, we update those attributes to persist the state. This allow saving the current state of the figure to an image whereas it used to always export the initial position of the nodes.

This PR also greatly improves our Sankey tests by testing drag operations for every arrangement! This allowed me to catch a regression as discussed in https://github.com/plotly/plotly.js/pull/3583#issuecomment-468780206 .


test/jasmine/tests/sankey_test.js
```diff
@@ -1009,56 +1011,116 @@ describe('sankey tests', function() {
         });
     });
 
-    describe('Test drag interactions:', function() {
-        var gd;
+    describe('Test drag interactions', function() {
+        ['freeform', 'perpendicular'].forEach(function(arrangement) {
```
antoinerg: Unfortunately, I am struggling quite a bit to get tests to pass for `snap` (which is the default `arrangement`). It seems like by the time we call `getBoundingClientRect();` in `getNodeCoords`, the DOM hasn't been updated yet and the tests fail. I think we should try to add it here so we :lock: down the dragging behavior.

antoinerg:
One thing I dislike about the current branch is that `ky`, the ratio between the height in pixel of a node and its value, is computed by the underlying Sankey generator. So although the user can position the nodes by specifying `node.(x|y)`, there's no way of knowing how tall the nodes will be. This makes it impossible to decide ahead of times what would be good values of `node.y` to prevent nodes from overlapping vertically...

I am thinking we might want to add a new attribute `node.scale` :thinking: What do you think?

cc @alexcjohnson @plotly/plotly_js

alexcjohnson:
When you drag nodes around by GUI, the others slide around to avoid overlaps. What if we did the same here? You provide positions but then we slide them around, keeping the ordering fixed, to avoid overlaps. It would still have some weird aspects, like if you're on a column that doesn't span the whole height and you *want* the nodes clustered in certain places you'd have to artificially squish them and trust us to sort it out.

`node.scale` could still be an interesting idea - though it's not just nodes, it applies to links too. But that seems like a separate feature, and in any case we don't want to allow overlapping nodes so we'll have to reserve the right to modify any positioning the user provides.

antoinerg:
Thank you @alexcjohnson for the quick reply! I will give this a bit more thought. In the meantime, here are a few comments:

> in any case we don't want to allow overlapping nodes

If the arrangement is `freeform` or `perpendicular`, we actually allow nodes to overlap. In fact, dragging in `freeform` allow nodes to be anywhere! This is why I set out to support arbitrary positioning via `node.(x|y)` to be on par with what can be done with the mouse.

If we decide to get rid of overlapping nodes, I could definitely try to implement your suggestion:
> you provide positions but then we slide them around, keeping the ordering fixed, to avoid overlaps.

Finally,
> `node.scale` could still be an interesting idea - though it's not just nodes, it applies to links too.

You're right: it should be a top-level attribute (ie. `scale` instead of `node.scale`).

alexcjohnson:
Ah OK, I guess I hadn't seen the interaction modes that allow overlaps to persist. In that case those should still be allowed, and should play nice with explicit positioning. But seems to me the default mode should be to respect explicit positioning but then push apart overlaps.

test/jasmine/tests/sankey_test.js
antoinerg: Resolved in 1a89a1e! We now have tests for drag interactions for every arrangement! :tada:

antoinerg:
Issue https://github.com/plotly/plotly.js/issues/3563 is also present for arrangement `freeform` and `perpendicular` (and hence in `master`) but it is fixed in the current PR and now has proper tests.

antoinerg:
> But seems to me the default mode should be to respect explicit positioning but then push apart overlaps.

@alexcjohnson After the nodes are pushed apart in `snap` mode (which is the default), should I update the `node.(x|y)` to reflect the new positions of the nodes?

antoinerg:
> @alexcjohnson After the nodes are pushed apart in snap mode (which is the default), should I update the node.(x|y) to reflect the new positions of the nodes?

I elected to not update `node.(x|y)` if the user specified them explicitly. This way, the user can switch from `freeform` to `snap` to `freeform` again.

peek 2019-03-05 12-06

chriddyp:
Will this work with `uirevision`?

antoinerg:
> Will this work with `uirevision`?

@chriddyp it now does with commit 8ddfe7e ! Thanks for the review :smile_cat:

etpinard:
@antoinerg what's the status on this one? Are planning on adding anything else to this PR?

src/traces/sankey/render.js
```diff
@@ -163,6 +168,93 @@ function sankeyModel(layout, d, traceIndex) {
...
+            // Change node's x position to align it with its column
+            dx = colX - node.x0;
+            node.x0 += dx, node.x1 += dx;
+
+        }
+        return columns;
+    }
+
+    // Force node position
+    if(trace.node.x.length !== 0 && trace.node.y.length !== 0) {
```
etpinard:
```js
if(trace.node.x.length > 0 && trace.node.y.length > 0) {
```

would save us 4 characters :smirk:

etpinard: and

```js
if(trace.node.x.length && trace.node.y.length) {
```

would save us 8 characters in total!

src/traces/sankey/render.js
```diff
@@ -628,13 +735,31 @@ function snappingForce(sankeyNode, forceKey, nodes, d) {
...
+function persistFinalNodePositions(d, gd) {
+    var x = [];
+    var y = [];
+    for(var i = 0; i < d.graph.nodes.length; i++) {
+        var nodeX = (d.graph.nodes[i].x0 + d.graph.nodes[i].x1) / 2;
+        var nodeY = (d.graph.nodes[i].y0 + d.graph.nodes[i].y1) / 2;
+        x.push(nodeX / d.figure.width);
+        y.push(nodeY / d.figure.height);
+    }
+    Plotly._guiRestyle(gd, {
```
etpinard: Strange that this doesn't result in circular dependency loop.

In other modules, we usually `var Registry = require('../../registry')` and use `Registry.call('_guiRestyle', /**/)`

src/traces/sankey/render.js
```diff
@@ -688,6 +813,9 @@ module.exports = function(gd, svg, calcData, layout, callbacks) {
         firstRender = true;
     });
 
+    // To prevent animation on dragging
+    var dragcover = gd.querySelector('.dragcover');
```
etpinard: Can we use `gd._fullLayout._dragCover` here? That `querySelector()` could be slow.

test/jasmine/assets/check_overlap.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+function compare(baseRects, compareRects) {
+    return baseRects.left < compareRects.right &&
+            baseRects.right > compareRects.left &&
+            baseRects.top < compareRects.bottom &&
+            baseRects.bottom > compareRects.top;
+}
+
+module.exports = function checkOverlap(base, elements) {
```
etpinard: Brilliant :sparkles:

etpinard:
> but it is fixed in the current PR and now has proper tests.
> Will this work with `uirevision`?

Amazing work :tada:

test/image/mocks/sankey_x_y.json
```diff
@@ -0,0 +1,28 @@
+{
+    "data": [
+        {
+            "type": "sankey",
+            "arrangement": "freeform",
```
etpinard: Do all the potential values of `arrangement` have an effect of the node layout when `node.x` and `node.y` are set?

src/traces/sankey/render.js
antoinerg: Mmm :thinking: should I change it?

etpinard: I would prefer, yes. Thanks!

test/image/mocks/sankey_x_y.json
antoinerg: > Do all the possible values of arrangement have an effect of the node layout when node.x and node.y are set?

Short answer: only `snap` can override `node.(x|y)`

Long answer:
When `node.(x|y)` is set, the nodes are forced to sit at the specified locations except in the `snap` arrangement. When drag and dropping, `snap` is the only arrangement in which we forbid nodes from overlapping so I made sure we keep honoring this promise.

The goal of this PR was to make it possible to replicate any state accessible via drag and drop by setting `node.(x|y)` and hence allowing to save to export images.

src/traces/sankey/render.js
antoinerg: Yes we can! Done in 40fb223fe48cacb155f294193fedbffc15a93069

src/traces/sankey/render.js
antoinerg: Sure thing! Done in 40fb223fe48cacb155f294193fedbffc15a93069

src/traces/sankey/render.js
antoinerg: Done in 40fb223fe48cacb155f294193fedbffc15a93069! :smile_cat:

antoinerg:
> @antoinerg what's the status on this one? Are planning on adding anything else to this PR?

I wasn't planning on adding anything else to this PR. It is bigger than I initially anticipated and I don't want to complicate it further at this stage.

test/image/mocks/sankey_x_y.json
etpinard: Ok, thanks for the info!

Could we then make `arrangement` default to `freeform` whenever `node.(x|y)` are set? That way, a user that wants to programmatically tweak the node position via `node.(x|y)` can do so w/o having to worry about changing the `arrangement` value.

antoinerg: Done in 65c7201

etpinard:
Nicely done :dancer: :dancer:

PR #3586, opened 2019-02-27
support locales in hovertemplate

### LINKED ISSUE #3580: hovertemplate d3-format numbers ignoring locale
When I set the locale, the dates are formatted correctly. But the numbers using the d3-format ignores that locale and instead follows the `en-us` default format and not the defined format from Plotly.

As per the built-in locales. => https://github.com/plotly/plotly.js/blob/master/lib/locales/fr.js#L86-L87
For FR, the separator should be a space " " and the decimal should be a comma ",".

Here is a sandbox of my implementation.
[Edit react-plotly.js DOM element is null or undefined -- fixed](https://codesandbox.io/s/381790o4rm?codemirror=1&fontsize=14)

Hover over the bars on the chart.

The $000,000.00 should look like this $000 000,00 in FR

Maybe I'm missing something obvious though.

antoinerg:
This PR closes #3580

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -10,6 +10,7 @@
 'use strict';
 
 var d3 = require('d3');
+var d3format = require('d3-format');
 var isNumeric = require('fast-isnumeric');
 
 var numConstants = require('../constants/numerical');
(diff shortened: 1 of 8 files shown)


src/lib/index.js
```diff
@@ -1076,7 +1078,9 @@ lib.hovertemplateString = function(string, labels) {
         }
 
         if(format) {
-            value = d3.format(format.replace(TEMPLATE_STRING_FORMAT_SEPARATOR, ''))(value);
+            var fmt = d3format;
+            if(locale) fmt = fmt.formatLocale(locale);
+            value = fmt.format(format.replace(TEMPLATE_STRING_FORMAT_SEPARATOR, ''))(value);
```
alexcjohnson: this seems a little funny - can you just use `fullLayout._d3locale` and avoid pulling in a new complete copy of `d3-format`? Anyway that's going to have different capabilities from the one in d3v3 that we already have and use, which will be confusing to users.

antoinerg: > this seems a little funny - can you just use `fullLayout._d3locale` and avoid pulling in a new complete copy of `d3-format`

Nice catch :eye: ! I agree this was a bad idea. In 1faaf8b6c3c161e243ed32e2bbf139c678b135ea, I reuse our existing formatting system.

PR #3589, opened 2019-02-28
Display hoverlabels above modebar

### LINKED ISSUE #3471: Always overlap modebar with hoverlabels
Hi there, it would be awesome if we could avoid this:
grafik
No matter if I put the modebar to orientation = v or h hoverlabels on the right side will land behind the modebar and thus be hard to read. Especially bad for narrow displays.

I tried to make the modebar only appear on hover as demonstrated in some issues/PRs but I couldn't find such an option.

Other than that I can't think of a situation where you'd **not** want hoverlabels to always overlap the modebar.

antoinerg:
Fixes #3471 

We now insert hoverlabels in a third `svg` element right after the modebar. This ensures that hoverlabels are always visible:
screenshot_2019-02-28_17-44-55

We also :hocho: `z-index` for the modebar: it is now sandwiched based on its position in the DOM. This should fix the following issue seen in a Dash app where the `z-index:1001` really was problematic:
53552498-b65a6f00-3b3b-11e9-805c-e37bba0753b2

### DIFF
--- a/build/plotcss.js
+++ b/build/plotcss.js
@@ -31,7 +31,7 @@ var rules = {
     "X .cursor-n-resize": "cursor:n-resize;",
     "X .cursor-ne-resize": "cursor:ne-resize;",
     "X .cursor-grab": "cursor:-webkit-grab;cursor:grab;",
-    "X .modebar": "position:absolute;top:2px;right:2px;z-index:1001;",
+    "X .modebar": "position:absolute;top:2px;right:2px;",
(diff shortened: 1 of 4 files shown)


etpinard:
Nice fix! 

Adding a new `.main-svg` svg node to ALL our graphs might be a bit too much for a patch release. So, I'll slide this one in the `1.46.0` milestone and let's wait a few days before merging it

etpinard:
good to merge :dancer:

Issue #3590, opened 2019-02-28
[feature-request] pie/sunburst label text direction


destradafilm:
feature request for pie chart label options -  to control how inside plot label directions are set. 
As you see here, some of the labels from the values that are bigger slices face different directions compared to each other (one with the text bottom facing the outside of the donut, the other with the bottom facing the inside of the donut). The smaller slices have text with the bottom of text right aligned to center of donut.

![plotly pie chart labels](https://user-images.githubusercontent.com/6043499/53606019-87152380-3b76-11e9-9824-b96d8b8ab222.PNG)

alexcjohnson:
Ha good timing, we were just talking about this today regarding the sunburst charts @etpinard is working on right now, but yes, would be great to implement for pie too.

destradafilm:
nice! I didn't even know you had a sunburst charts. Where can I find examples/docs on sunburst charts?

alexcjohnson:
Sunburst is not yet available, but coming soon!

etpinard:
I'm thinking of adding `textrotation: 'radial' | 'tangent' | 'none' | 'auto'` - where `'auto'` is the current default.

Any objections?

alexcjohnson:
Perhaps change `'none'` to `'fixed'` or `'constant'` so it can at some point be modified with another attribute containing an angle - or an array of angles? Other than that, 👍

destradafilm:
@etpinard thanks for taking care of this. 

- I'm wondering if 'tangent' will ensure that the text will always face one direction (in comparison to the preview I posted above: the tangential text with the largest slices in pie chart 1 and 2 face different directions)
- What will be the behavior if a user chooses 'tangential' but the text is super long? Will the text ellipse? Will it get super small like it currently does when you choose textposition: 'inside' ?

alexcjohnson:
That's a good point @destradafilm - I would assume the default would be to always have the text "right-side-up" as referenced to the screen, but that's the behavior you don't want since a biggest slice <50% like your chart 2 would have its text rotated right, but >50% like the other two would have it rotated left. A behavior where you always have the top (for tangential) or right (for radial) edge of the text toward either the inside or outside of the circle would need to be a separate attribute.

parthhhh:
Hello, wondering if this feature is still being implemented?

parthhhh:
@etpinard Thanks for merging this issue, looking forward to this feature.

parthhhh:
[...]

PR #3594, opened 2019-03-01
Sunburst traces

etpinard:
Sunburst traces are coming to plotly.js :sun_with_face: 

peek 2019-03-01 15-10

### 1st iteration features

- custom click iterations: clicking on branches zooms in, click on relative roots (in the middle) zooms out
- per-sector coloring through `marker.colors` with per-branch color inheritance
- optional per-sector values and two ways to sum them via `branchvalues: 'total' | 'extra'`
- fully-featured hover, with `hovertext`, `hoverinfo` and `hovertemplate` support
- `pie`-like sector coloring via `sunburstcolorway` and `extendsunburstcolors`
- "leaf"-only opacity setting (to easily recreate common sunburst examples in the wild where leaves are less opaque then branches e.g. https://observablehq.com/@d3/d3-zoomable-sunburst)
- the same auto-text-fitting algo used in `pie` traces

### TODOs

- [x] agree on attribute names:
  + `labels` / `parents` / `values` or `label` / `value` / `parent` ?
  + `level` / `maxdepth` can anyone think of less-confusing names?
- [x] agree on `branchvalues` default `'total'` or `'extra'`?
- [x] experiment with `storke-linejoin` and `stroke-miterlimit` - see 1b21aae15a38b963f952b74f4da563f6ce609c41
- [x] make `marker.line.width: 1` the default - **done** in c1ccad65b
- [x] agree on click transitions easing and duration
- [x] fixup `react` + transitions edge cases
- [x] hook up `uirevision` **done in:** 0da13dbe7
- [x] mucho jasmine tests
- [x] a few more mocks

### Things that may be added in this PR

-  smooth transitions on click interactions + compatibility with `Plotly.react` transition API. I'll give this a try next week. **THIS IS GOING IN -->**  aa3adf33e 
-  an enumerated `marker.colorinheritance` attribute to offer more ways to inherit sector colors
-  allow `labels` / `parents` duplication when unambiguous (`d3-hierarchy` is very strict)
-  replace `d3-hierarchy` ([`hierarchy`](https://github.com/d3/d3-hierarchy#hierarchy), [`stratify`](https://github.com/d3/d3-hierarchy#stratify), [`partition`](https://github.com/d3/d3-hierarchy#partition)) calls with own implementation

---

cc @plotly/plotly_js @chriddyp @nicolaskruchten


nicolaskruchten:
This looks really nice!

One concern, re the text, which may also apply to pies... Is it possible to control the size so it's not 'auto'? Right now I worry about the larger text calling more attention to certain slices just because the text label is smaller therefore gets scaled up. Also aesthetically is it possible to just say "do a radial layout always even if text could be bigger/more horizontal"? Or possibly "do a symmetrical layout of text"? In certain cases, symmetry will be more highly-valued that absolute maximum readability...

etpinard:
> Is it possible to control the size so it's not 'auto'?

yes, via `insidetextfont.size` 

> Also aesthetically is it possible to just say "do a radial layout always even if text could be bigger/more horizontal"?

Funny, a user wrote in about that in https://github.com/plotly/plotly.js/issues/3590 last week. I might try to :white_check_mark: in this PR.

nicolaskruchten:
> yes, via insidetextfont.size

OK cool, does that disable the rotations or does it still try to get the most-horizontal-that-still-fits or... ?

etpinard:
> does that disable the rotations

no, that's what https://github.com/plotly/plotly.js/issues/3590 requests.

etpinard:
First-cut transition work is on:

https://github.com/plotly/plotly.js/compare/sunburst...sunburst-transitions

Here's an sample codepen: https://codepen.io/etpinard/pen/Bbpqvo,

I would appreciate some feedback. Thanks!

antoinerg:
@etpinard Sunburst will be a strong addition to plotly.js :muscle: and animations do add a little bit of wow factor :star2: 

Would it be possible/easy to have the labels appear _after_ the resizing of the sectors is done. I'm asking because the labels are sometimes out of place during the transition. For example, when going from `Aromas` back to the root, you see the lines delimiting the sectors flying above the labels of the leaf nodes. This intermediate state, where labels are in between sectors and traversed by lines, is not correct or meaningful so I think we should hide it from the user.

nicolaskruchten:
WOW the sector tweens are gorgeous! nice job!!

That said, I agree with @antoinerg re the text. If we can translate the text with the sector tween, it should only start to appear after the sectors have stopped moving IMO.

etpinard:
Update with text-position transitions on https://github.com/plotly/plotly.js/compare/sunburst...sunburst-transitions 

examples --> https://codepen.io/etpinard/pen/LayPQQ

antoinerg:
> examples --> https://codepen.io/etpinard/pen/LayPQQ

Text-position transition looks even better than I imagined :tada: :muscle: Just superb!

nicolaskruchten:
OMG so awesome!

chriddyp:
yeah that's gold 🏅

jonmmease:
So nice @etpinard!

what is double click intended to do on these?  Seems to sometimes hide some of the wedges but I haven't been able to grok exactly what's going on.

On second try, it's not double clicking per-say, it's that clicking on the sunburst while it's animating sometimes causes some odd behavior with wedges disappearing.  Maybe disable click interactions during transitions?

etpinard:
### Pre-vacation update

Updated example codepen: https://codepen.io/etpinard/pen/ZPybzB 

I squashed all my transitions work into one large commit  https://github.com/plotly/plotly.js/pull/3594/commits/aa3adf33e45085034709372819b3363393f665dc where:

- sector paths **and** text are tweened :tada: 
- `sunburst` will be our 2nd trace with support for transitions via `Plotly.animate` and `Plotly.react`
- clicking on a sector _simply_ calls `Plotly.animate` with the appropriate arguments.
- the sector click handler is disabled during transitions, so the odd "wedges disappearing" behaviour is a thing of the past

The `TODO` list in the PR description has been updated. Please let me know if there's any other blocking items for this first iteration.

Now, I'm thinking of removing `leaf.textposition` and making sunburst traces only support inside text for all sectors except the hierarchy root. Getting outside text to tween properly will take at least 1 full day: as most sunburst in the wild only have inside text, removing `leaf.textposition` sounds ok to me. Please let me know if you disagree.

Moreover, I'm thinking of not implementing attributes `sort`, `direction` and `rotation` in this first iteration. Please let me know if that's ok.

etpinard:
Tagging this as `reviewable` cc @plotly/plotly_js

src/traces/sunburst/attributes.js
```diff
@@ -0,0 +1,145 @@
...
+'use strict';
+
+var plotAttrs = require('../../plots/attributes');
+var hovertemplateAttrs = require('../../components/fx/hovertemplate_attributes');
+var domainAttrs = require('../../plots/domain').attributes;
+var pieAtts = require('../pie/attributes');
+
+var extendFlat = require('../../lib/extend').extendFlat;
+
+// TODO should we use singular `label`, `parent` and `value`?
```
antoinerg: For `sankey` traces, we went with singular attribute names. Maybe we should stick to that? 

What do @nicolaskruchten, @alexcjohnson @plotly/plotly_js think?

nicolaskruchten: I favour singular everywhere

src/traces/sunburst/attributes.js
```diff
@@ -0,0 +1,145 @@
...
+    parents: {
+        valType: 'data_array',
+        editType: 'calc',
+        description: [
+            'Sets the parent sectors for each of the sunburst sectors.',
+            'Empty string items \'\' are understood to reference',
+            'the root node in the hierarchy.',
+            'If `ids` is filled, `parents` items are understood to be "ids" themselves.',
+            'When `ids` is not set, plotly attempts to find matching items in `labels`,',
+            'but beware there must be unique.'
```
antoinerg: replace there by they :point_right:  but beware _they_ must be unique

src/traces/sunburst/attributes.js
etpinard: So, should we change the `pie` attributes `labels` and `values` accordingly?

etpinard: ... and `marker.colors`

nicolaskruchten: ...of course I favour consistency above all!

antoinerg: > So, should we change the pie attributes labels and values accordingly?
> ... and marker.colors

Ok so it seems like we have more plurals than singulars.

> ...of course I favour consistency above all!

Me too so I guess we should stick with what we already have right now (ie. plural form). Maybe in the future, we could support both.

src/traces/sunburst/calc.js
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    // TODO might be better to replace stratify() with our own algorithm
+    var root;
+    try {
+        root = d3Hierarchy.stratify()
+            .id(function(d) { return d.id; })
+            .parentId(function(d) { return d.pid; })(cd);
+    } catch(e) {
+        return Lib.warn('Failed to build sunburst hierarchy.');
```
antoinerg: It might be useful to return the error message from `stratify()`:
```
return Lib.warn('Failed to build sunburst hierarchy with error: `' + e.message + '`');
```
to help understand why a given dataset is failing to render.

src/traces/sunburst/calc.js
```diff
@@ -0,0 +1,237 @@
...
+
+                    if(d.children) {
+                        var partialSum = d.children.reduce(function(a, c) {
+                            return a + c.data.data.v;
+                        }, 0);
+                        if(v < partialSum) {
+                            d.value = partialSum;
+                            return Lib.warn([
+                                'Total value for node', d.data.data.id,
+                                'is smaller than the sum of its children.'
```
antoinerg: There are two ways of denoting the size of sectors depending on `branchvalues`:
1- `total`: which is in absolute values
2- `extra`: which is relative to the sum of its children

Because we filter negative values, we can only reach this `Lib.warn` in `total`. In `extra`, we would have failed earlier when calling `stratify()`. At the moment, we handle the situation of parents that are smaller than children in two different manners.

I think the current behvaior of `extra` is more sane: it simply refuses to render something it can't correctly do. Until we add support for rendering parents that are smaller than the sum of its children (by padding the remaining space for example), I think it would make sense to bail out here to be correct _and_ consistent with `extra`.

Sorry for the long comment but I wanted to add context for other potential reviewers.
@plotly/plotly_js @nicolaskruchten

src/traces/sunburst/attributes.js
alexcjohnson: At one point I think the idea was `arrayOk` attributes - which could be scalars OR arrays - should be singular, and `data_array` attributes should be plural. But I don't think we were ever consistent about that, and anyway it doesn't work to pluralize things like `x`. So yeah, if I were to redo it all today I'd probably vote 100% singular (and perhaps we should make that change in a major bump?), but at this point wherever we're using a name that's already in the schema somewhere we should try to match it.

antoinerg:
> agree on branchvalues default 'total' or 'extra'?

I'm still debating with myself about this one... I don't have a compelling argument for changing it from the current default `extra` to `total`. Maybe someone else does @plotly/plotly_js?

src/traces/sunburst/attributes.js
etpinard: > but at this point wherever we're using a name that's already in the schema somewhere we should try to match it.

those were my thoughts exactly :ok_hand:

antoinerg: :+1: I guess you can :zap: that TODO line and add it to our wishlist for v2

src/traces/sunburst/plot.js
```diff
@@ -0,0 +1,806 @@
...
+var svgTextUtils = require('../../lib/svg_text_utils');
+var setCursor = require('../../lib/setcursor');
+var appendArrayPointValue = require('../../components/fx/helpers').appendArrayPointValue;
+
+var transformInsideText = require('../pie/plot').transformInsideText;
+var formatPieValue = require('../pie/helpers').formatPieValue;
+var styleOne = require('./style').styleOne;
+
+var CLICK_TRANSITION_TIME = 1000;
+var CLICK_TRANSITION_EASING = 'linear'; // TODO or something else?
```
antoinerg: > agree on click transitions easing and duration

I vote for `linear` easing as we currently have. 

As for the transition time, 1000ms is nice to showcase the animation but it is too long considering it prevents from clicking again. This means that after zooming 4 levels deep with one click, it then takes 4 clicks times 1000ms to get back to the root again. This is too long. I think we shouldn't prevent clicking for more than 500 ms.

Therefore, I personally like anything between 250 and 500ms. We have time to see the animation but it doesn't block clicking.

@etpinard, maybe in the future we could this configurable via an attribute?

src/traces/sunburst/calc.js
etpinard: > In `extra`, we would have failed earlier when calling `stratify()`
> I think the current behvaior of `extra` is more sane: it simply refuses to render something it can't correctly do

Hmm. I'm a little confused @antoinerg . Can you share an example of `stratify()` failing under `branchvalues:'extra'`?

src/traces/sunburst/plot.js
etpinard: Hmm. 500ms is fast, too fast for my liking:

Peek 2019-03-28 09-47

Can we settle on 750ms?



> we shouldn't prevent clicking for more than 500 ms.

You're right here. It would be nice to not have to prevent clicks during transitions. We'll keep that in mind for the next iteration.

src/traces/sunburst/attributes.js
```diff
@@ -0,0 +1,145 @@
...
+        valType: 'data_array',
+        editType: 'calc',
+        description: [
+            'Sets the values associated with each of the sunburst sectors.',
+            'Use with `branchvalues` to determine how the values are summed.'
+        ].join(' ')
+    },
+    branchvalues: {
+        valType: 'enumerated',
+        values: ['total', 'extra'],
```
etpinard: rename to `'remainder'`

src/traces/sunburst/calc.js
antoinerg: If you use a negative value in `extra`, we will filter that node out. If that node was a parent, `stratify()` will fail because it will be missing a node.

src/traces/sunburst/plot.js
antoinerg: > You're right here. It would be nice to not have to prevent clicks during transitions. We'll keep that in mind for the next iteration.

If you don't have to prevent clicks during transitions, I could even agree with going with 1000ms! I love the animations :smile: In the meantime, if you prefer 750ms, let's go with that!

In a future iteration, I'd be curious to see it makes sense to make the animation time proportional to the number of levels you're zooming in. The deeper you zoom, the more things move around and maybe longer animations would be beneficial for those cases while zooming only one level would be faster :thinking:

src/traces/sunburst/attributes.js
etpinard: done in https://github.com/plotly/plotly.js/pull/3594/commits/1258f2da606954952b7e2bfc1c82eb7a274b2ef1

src/traces/sunburst/attributes.js
etpinard: done in done in https://github.com/plotly/plotly.js/pull/3594/commits/1258f2da606954952b7e2bfc1c82eb7a274b2ef1

src/traces/sunburst/attributes.js
etpinard: TODO :zap: in https://github.com/plotly/plotly.js/pull/3594/commits/1258f2da606954952b7e2bfc1c82eb7a274b2ef1

added to v2 list:

image

src/traces/sunburst/plot.js
etpinard: Lowered to 750ms in https://github.com/plotly/plotly.js/pull/3594/commits/b59be1d732b8412dfda8af60e114cb1c4050806b

src/traces/sunburst/calc.js
etpinard: Thanks very much for the tip @antoinerg !

Done in https://github.com/plotly/plotly.js/pull/3594/commits/4a2f7a5a59e7868b1a21222647d2632cb7d3edb0

src/traces/sunburst/calc.js
etpinard: Good call! Done in https://github.com/plotly/plotly.js/pull/3594/commits/4a2f7a5a59e7868b1a21222647d2632cb7d3edb0

antoinerg:
This is one hell of a PR :sun_with_face: ! Amazing work @etpinard :muscle:

:dancer: :dancer: :dancer:

mwouts:
Thanks for the amazing contribution! I love the new sunburst plots! For those who can't wait for the next plotly.py release, have a look at the example below to see how to generate those plots in Python right now. 

[World population](https://github.com/mwouts/world_bank_data#ready-for-an-interative-tutorial)

PR #3597, opened 2019-03-02
fix reset camera buttons to work after switching projection between perspective and orthographic

### LINKED ISSUE #3596: Reset camera to default or the last saved position may break the plot after switching to a new projection type
After switching gl3d projection type, clicking reset camera may remove the scene!
But the plot would appear after clicking other modebar buttons e.g. pan/orbit.
[codepen](https://codepen.io/MojtabaSamimi/pen/eXZKzQ?editors=0010).

archmoj:
Fixes the issue described in #3596.
[codepen Before](https://codepen.io/MojtabaSamimi/pen/eXZKzQ?editors=0010).
[codepen After](https://codepen.io/MojtabaSamimi/pen/PLNadO?editors=0010).

fixes #3596

### DIFF
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -338,8 +338,6 @@ modeBarButtons.resetCameraLastSave3d = {
 function handleCamera3d(gd, ev) {
     var button = ev.currentTarget;
     var attr = button.getAttribute('data-attr');
-    if(attr !== 'resetLastSave' && attr !== 'resetDefault') return;
-
     var fullLayout = gd._fullLayout;
     var sceneIds = fullLayout._subplots.gl3d;
     var aobj = {};
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -358,13 +358,6 @@ function handleCamera3d(gd, ev) {
             aobj[key + '.eye'] = null;
             aobj[key + '.center'] = null;
         }
-
-        var newOrtho = (scene.viewInitial.projection.type === 'orthographic');
-        var oldOrtho = scene.camera._ortho;
-
-        if(newOrtho !== oldOrtho) {
-            aobj[key + '.projection'] = scene.viewInitial.projection;
-        }
     }
 
     Registry.call('_guiRelayout', gd, aobj);
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -347,22 +347,15 @@ function handleCamera3d(gd, ev) {
         var key = sceneId + '.camera';
         var scene = fullLayout[sceneId]._scene;
 
-        if(attr === 'resetDefault' || attr === 'resetLastSave') {
-            aobj[key] = {
-                'up': scene.viewInitial.up,
-                'eye': scene.viewInitial.eye,
-                'center': scene.viewInitial.center,
-                'projection': {
-                    type: (scene.camera._ortho) ?
-                    'orthographic' : 'perspective'
-                }
-            };
-        }
+        if(attr === 'resetLastSave') {
+            aobj[key + '.up'] = scene.viewInitial.up;
+            aobj[key + '.eye'] = scene.viewInitial.eye;
+            aobj[key + '.center'] = scene.viewInitial.center;
 
-        if(attr === 'resetDefault') {
-            aobj[key].up = null;
-            aobj[key].eye = null;
-            aobj[key].center = null;
+        } else if(attr === 'resetDefault') {
+            aobj[key + '.up'] = null;
+            aobj[key + '.eye'] = null;
+            aobj[key + '.center'] = null;
         }
     }
 
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -348,7 +348,7 @@ function handleCamera3d(gd, ev) {
         var scene = fullLayout[sceneId]._scene;
 
         if(attr === 'resetDefault' || attr === 'resetLastSave') {
-            aobj[key] = Lib.extendDeep({}, scene.cameraInitial);
+            aobj[key] = Lib.extendDeep({}, scene.viewInitial);
 
             aobj[key].projection = {
                 type: (scene.camera._ortho) ? 'orthographic' : 'perspective'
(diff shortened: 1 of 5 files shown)


src/components/modebar/buttons.js
```diff
@@ -347,15 +347,19 @@ function handleCamera3d(gd, ev) {
         var key = sceneId + '.camera';
         var scene = fullLayout[sceneId]._scene;
 
-        if(attr === 'resetDefault') {
+        if(attr === 'resetDefault' || attr === 'resetLastSave') {
             aobj[key] = Lib.extendDeep({}, scene.cameraInitial);
```
etpinard: Wouldn't it be easier to _just_ save keys `up`, `eye` and `center` in `scene.cameraInitial` instead of saving the all `scene.camera` keys and then doing this fancy footwork to get the correct `projection.type`?

In other words, do we really need a `cameraInitial` stash? Could we instead just use a "view" stash:

```js
scene.viewInitial = {
  up: {x: /**/, y: /**/, z: /**/},
  eye: {x: /**/, y: /**/, z: /**/},
  center: {x: /**/, y: /**/, z: /**/}
}
```

archmoj: Good call. 
Done in https://github.com/plotly/plotly.js/pull/3597/commits/9e1aa7af86a8dab6e807e8ddea8dca826031cc41.

src/components/modebar/buttons.js
```diff
@@ -347,15 +347,19 @@ function handleCamera3d(gd, ev) {
         var key = sceneId + '.camera';
         var scene = fullLayout[sceneId]._scene;
 
+        if(attr === 'resetDefault' || attr === 'resetLastSave') {
+            aobj[key] = Lib.extendDeep({}, scene.viewInitial);
+
+            aobj[key].projection = {
```
etpinard: We shouldn't need this after 9e1aa7af86a8dab6e807e8ddea8dca826031cc41

etpinard: Actually, you'll need to tweak `aobj` / `key` logic. In the end, `aobj` should look like:

```js
aobj = {
  'scene.camera.up': {/**/},
  'scene.camera.eye': {/**/},
  'scene.camera.center': {/**/}
}
```

as opposed to

```js
aobj = {
  'scene.camera': {
    up: {/**/},
    eye: {/**},
    center: {/**}
  }
}
```

src/components/modebar/buttons.js
```diff
@@ -348,10 +348,14 @@ function handleCamera3d(gd, ev) {
...
         if(attr === 'resetDefault' || attr === 'resetLastSave') {
-            aobj[key] = Lib.extendDeep({}, scene.viewInitial);
-
-            aobj[key].projection = {
-                type: (scene.camera._ortho) ? 'orthographic' : 'perspective'
+            aobj[key] = {
+                'up': scene.viewInitial.up,
+                'eye': scene.viewInitial.eye,
+                'center': scene.viewInitial.center,
+                'projection': {
```
etpinard: We shouldn't need to update `projection` at all, am I right?

archmoj: You are right. But in order to that we need to explicitly pass the current projection here.

etpinard: Ok, that's a bug then. Calling

```
Plotly.relayout('graph', 'scene.camera.up', {/**/})
```

should not change the projection type.

alexcjohnson: don't we just need to flatten it to `aobj[key + '.up'] = ...` etc?

etpinard:
Nicely done :dancer:

src/components/modebar/buttons.js
archmoj: Thanks @alexcjohnson & @etpinard.
Flattening was really helpful.

archmoj:
> Nicely done
@etpinard If you don't mind, there is one more commit ready to push which help switching the projection to graph default.

src/components/modebar/buttons.js
```diff
@@ -351,12 +353,18 @@ function handleCamera3d(gd, ev) {
...
             aobj[key + '.up'] = null;
             aobj[key + '.eye'] = null;
             aobj[key + '.center'] = null;
         }
+
+        var newOrtho = (scene.viewInitial.projection.type === 'orthographic');
+        var oldOrtho = scene.camera._ortho;
+
+        if(newOrtho !== oldOrtho) {
+            aobj[key + '.projection'] = scene.viewInitial.projection;
```
etpinard: Wait. Why do we need this?

test/jasmine/tests/gl3d_plot_interact_test.js
```diff
@@ -1490,7 +1490,7 @@ describe('Test gl3d relayout calls', function() {
         .then(done);
     });
 
-    it('@gl should maintain projection type when resetCamera buttons clicked after switching projection type from perspective to orthographic', function(done) {
+    it('@gl resetCamera buttons should be able to reset projection type after switching projection type from perspective to orthographic', function(done) {
```
etpinard: Why do you change your mind here?

archmoj: Now that we could switch projections to initial projection I thought we may change this.
If the user is not happy with the changes made to the camera and wants to get back to initial/home view this allows that. [Codepen](https://codepen.io/MojtabaSamimi/pen/JzbRjb?editors=0010).

etpinard:
> Now that we could switch projections to initial projection I thought we may change this.
If the user is not happy with the changes made to the camera and wants to get back to initial/home view this allows that.

I disagree. The reset camera buttons should reset only the attribute that can be changed via user interactions i.e. drag/scroll/clicking on modebar buttons. As we can't change the `projection.type` via user interactions (we need a `relayout` call here), it shouldn't be part of `viewInitial` and shouldn't be handled on modebar button clicks.

I'm voting for the version pre https://github.com/plotly/plotly.js/pull/3597/commits/8444b99579fa42834b2cf8ab45dffa865e31b584

src/components/modebar/buttons.js
```diff
@@ -338,6 +338,8 @@ modeBarButtons.resetCameraLastSave3d = {
 function handleCamera3d(gd, ev) {
     var button = ev.currentTarget;
     var attr = button.getAttribute('data-attr');
+    if(attr !== 'resetLastSave' && attr !== 'resetDefault') return;
```
etpinard: When does this happen?

etpinard:
Well earned :dancer: 

Thanks for following through from my comments.

Issue #913, opened 2016-09-07
Add support for non-JSON image mocks


etpinard:
PR https://github.com/plotly/plotly.js/pull/850 must us realised that not all plotly.js features are JSON serializable e.g. `Float32Array`.

**Possible solution:** we could try to make the image tests routine and the test dashboard accept `js` files as mock. These `js` files would generate `{ data: [], layout: {} }` object.

For example,

```js
// in test/image/mocks/float32array.js

module.exports = function() {
  return {
    data: [{
      xy: new Float32Array([1,2,3,4])
    }]
  }
}
```

etpinard:
Ping @plotly/plotly_js 

Might be a good time to tackle this.

etpinard:
So, I propose:

- keep the `test/image/mocks/` folder intact. A lot of things (e.g. https://rreusser.github.io/plotly-mock-viewer/) are looking into this folder from `http://raw.githubusercontent.com/` to do stuff with the JSON mocks.
- add a `test/image/mock-generators/` folder with js files such that: `test/image/mock-generators/some-mock.js` would correspond to `test/image/mocks/some-mock.json`

then

- before `npm run test-image`, `npm run baseline` and `npm run test-export` we would run these `test/image/mock-generators/` scripts in Node.js to then update `text/image/mocks/` before generating baselines or running the image tests.

Now, we would still need to keep the output of `test/image/mock-generators/` pure JSON (e.g. no TypedArrays). Down the road (when / if we switch to `orca` to run our image tests), we could try to generate the mocks in the front-end by adding a new `orca` renderer.

We would also need to find a way to keep `test/image/mock-generators/` and `test/image/mocks/` in-sync. I can think of two solutions:

+ PRs that add files to `test/image/mock-generators/` **do not add** the corresponding files in `test/image/mocks/`. `test/image/mocks/` would be added during version commits.
  - :+1: no extra step (that could be sometimes forgotten) for developers
  - :-1: new `test/image/mocks/` JSON files are not available until next plotly.js version

+ PRs that add files to `test/image/mock-generators/`, **must add** the corresponding files in `test/image/mocks/`
  - :+1:  `test/image/mocks/` JSON files are always available
  - :-1:  we would have to add extra machinery (probably a syntax test) to make sure `test/image/mock-generators/` and `test/image/mocks/` are in-sync

Thougths?

archmoj:
[...]

PR #3598, opened 2019-03-03
Fix parcoords dimensions values & line coloring error with integer TypedArray

### LINKED ISSUE #3595: parcoords line coloring error with integer TypedArray
There seems to be an issue with specifying line color of a `parcoords` trace with an integer TypedArray. All of the values seem to be rounded to the min or max color value.

CodePen: https://codepen.io/jonmmease/pen/QoNgMe?editors=1010
```javascript
var trace = {
  type: 'parcoords',
  line: {
	  color: new Int32Array([0, 1, 2])
	  // color: new Float32Array([0, 1, 2])
          // color: [0, 1, 2]	  
  },
	
  dimensions: [{
    range: [1, 5],
    label: 'A',
    values: [1,4,3]
  }, {    
    range: [1,5],
    label: 'B',
    values: [3,1.5,2],
  }, {
    range: [1, 5],
    label: 'C',
    values: [2,4,1],
  }]
};

var data = [trace]

Plotly.plot('graphDiv', data, {}, {showSendToCloud: true});
```

When `line.color` is specified as a standard array or floating point TypedArray:

newplot 4

When `line.color` is specified as an integer TypedArray:
newplot 3

archmoj:
Fixes two issues reported & discovered in #3595 resulted from `integer` rounding errors. 
- [x] case of `color.line` typed arrays e.g. int32
- [x] case of `dimensions.values` typed arrays e.g. int32

@plotly_js
[Codepen Before](https://codepen.io/jonmmease/pen/QoNgMe?editors=1010).
[Codepen After](https://codepen.io/MojtabaSamimi/pen/GeZzjj?editors=1010).

### DIFF
--- a/src/traces/parcoords/calc.js
+++ b/src/traces/parcoords/calc.js
@@ -46,10 +46,6 @@ function constHalf(len) {
     return out;
 }
 
-function isTypedArray(a) {
-    return !Array.isArray(a) && Lib.isArrayOrTypedArray(a);
-}
-
 function convertTypedArray(a) {
-    return (isTypedArray(a)) ? Array.prototype.slice.call(a) : a;
+    return (Lib.isTypedArray(a)) ? Array.prototype.slice.call(a) : a;
 }
--- a/src/traces/parcoords/calc.js
+++ b/src/traces/parcoords/calc.js
@@ -14,7 +14,13 @@ var Lib = require('../../lib');
 var wrap = require('../../lib/gup').wrap;
 
 module.exports = function calc(gd, trace) {
-    var cs = !!trace.line.colorscale && Lib.isArrayOrTypedArray(trace.line.color);
+
+    for(var i = 0; i < trace.dimensions.length; i++) {
+        trace.dimensions[i].values = convertTypedArray(trace.dimensions[i].values);
+    }
+    trace.line.color = convertTypedArray(trace.line.color);
+
+    var cs = !!trace.line.colorscale && Array.isArray(trace.line.color);
     var color = cs ? trace.line.color : constHalf(trace._length);
     var cscale = cs ? trace.line.colorscale : [[0, trace.line.color], [1, trace.line.color]];
 
@@ -38,4 +44,12 @@ function constHalf(len) {
         out[i] = 0.5;
     }
     return out;
+}
+
+function isTypedArray(a) {
+    return !Array.isArray(a) && Lib.isArrayOrTypedArray(a);
+}
+
+function convertTypedArray(a) {
+    return (isTypedArray(a)) ? Array.prototype.slice.call(a) : a;
 }
--- a/src/traces/parcoords/defaults.js
+++ b/src/traces/parcoords/defaults.js
@@ -20,12 +20,7 @@ var maxDimensionCount = require('./constants').maxDimensionCount;
 var mergeLength = require('./merge_length');
 
 function handleLineDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
-
     var lineColor = coerce('line.color', defaultColor);
-    if(!Array.isArray(lineColor) && Lib.isArrayOrTypedArray(lineColor)) {
-        // should convert typed arrays e.g. integers to real numbers
-        lineColor = traceOut.line.color = Array.prototype.slice.call(lineColor);
-    }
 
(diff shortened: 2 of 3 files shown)


antoinerg:
Thanks for the fix @archmoj! It looks good to me :+1: 

I am wondering if we should add a parcoords mock with `line.color` as a typedArray to :lock: this down.

archmoj:
@antoinerg thanks for the review & comment. I am not sure if we could make a JSON mock with typed arrays. So I may add a jasmine test.

archmoj:
@antoinerg Please note that https://github.com/plotly/plotly.js/pull/3598/commits/e8cc760f642b932486638ef31a9cbf682926e06c is also added concerning my comment [here](https://github.com/plotly/plotly.js/issues/3595#issuecomment-469066307).

antoinerg:
> I am not sure if we could make a JSON mock with typed arrays.

You're probably right :smile: Thanks for the jasmine tests!

I confirm this PR fix both issues in #3595 but I wonder why `coerce` isn't already doing the right thing and whether copying the `TypedArray` into an `Array` is avoidable. Therefore, I will leave it up to @etpinard to give you a dancer!

Nice work @archmoj

src/traces/parcoords/defaults.js
```diff
@@ -49,6 +49,11 @@ function dimensionDefaults(dimensionIn, dimensionOut) {
     }
 
     var values = coerce('values');
+    if(!Array.isArray(values) && Lib.isArrayOrTypedArray(values)) {
+        // should convert typed arrays e.g. integers to real numbers
+        values = dimensionOut.values = Array.prototype.slice.call(values);
```
etpinard: Nop. Let's not do this during defaults to remain consistent with other typed-array-to-array blocks.

Move this to `parcoords/calc.js`.

src/traces/parcoords/defaults.js
```diff
@@ -20,7 +20,12 @@ var maxDimensionCount = require('./constants').maxDimensionCount;
 var mergeLength = require('./merge_length');
 
 function handleLineDefaults(traceIn, traceOut, defaultColor, layout, coerce) {
+
     var lineColor = coerce('line.color', defaultColor);
+    if(!Array.isArray(lineColor) && Lib.isArrayOrTypedArray(lineColor)) {
+        // should convert typed arrays e.g. integers to real numbers
+        lineColor = traceOut.line.color = Array.prototype.slice.call(lineColor);
```
etpinard: same here.

etpinard:
> but I wonder why `coerce` isn't already doing the right thing and whether copying the `TypedArray` into an `Array` is avoidable

because we don't want to do that in general. If we can keep typed arrays as typed array and handle them correct downstream, then we shouldn't convert them. See for example `ax.makeCalcdata` which is called by most trace modules:

https://github.com/plotly/plotly.js/blob/ecfa45a0f00b1b49c2fda9853e6ac69fa06c7a8f/src/plots/cartesian/set_convert.js#L516-L537

etpinard:
> I am not sure if we could make a JSON mock with typed arrays.

See https://github.com/plotly/plotly.js/issues/913

src/traces/parcoords/defaults.js
antoinerg: > If we can keep typed arrays as typed array and handle them correct downstream, then we shouldn't convert them.

That makes perfect sense!

src/traces/parcoords/calc.js
```diff
@@ -39,3 +45,11 @@ function constHalf(len) {
     }
     return out;
 }
+
+function isTypedArray(a) {
```
etpinard: No need for this. `Lib.isTypedArray` [is](https://github.com/plotly/plotly.js/blob/ecfa45a0f00b1b49c2fda9853e6ac69fa06c7a8f/src/lib/index.js#L29) a [thing](https://github.com/plotly/plotly.js/blob/ecfa45a0f00b1b49c2fda9853e6ac69fa06c7a8f/src/lib/array.js#L23-L26).

archmoj: @etpinard Thanks for the note. 
Done in e577e49.

etpinard:
Nicely done :dancer:

PR #3609, opened 2019-03-06
Fix handling of horizontal legendgroup

### LINKED ISSUE #1913: Horizontal legend wrapping is broken when legendgroups are used
jsfiddle repro here: https://jsfiddle.net/6oq39dpy/3/

Normally, when a horizontal legend contains many items, they get wrapped onto multiple lines.

However, if legend groups are used at all -- even if only a single item is assigned to its own legend group -- nothing is wrapped, and legend items extending beyond the width of the plot are invisible and inaccessible.

To make wrapping appear again in the jsfiddle repro, just comment out line 10, which sets a legend group for the first trace.

While searching for an already open issue for this, I found #769, which led to the original PR that added support for wrapping horizontal legends in the first place, and mentions legend groups, but only as a workaround for the original problem. I've only skimmed the related code so far so I don't know exactly what causes the bug. :-)

antoinerg:
Fixes #1913

See https://github.com/plotly/plotly.js/issues/1913#issuecomment-469848892 for details.


etpinard:
Can you tell us how `legend.tracegroupgap` affects legend groups on `orientation: 'h'` legends now?

antoinerg:
> Can you tell us how legend.tracegroupgap affects legend groups on orientation: 'h' legends now?

@etpinard `legend.tracegroupgap` changes the spacing between legend items when horizontal. If I set it to 50 in the new mock, we get:
newplot 6

I don't know if this behavior was intentional but the current PR preserves it.

antoinerg:
I think we need to introduce a new legend attribute to control the spacing between trace items independently of the spacing between grouped items. The new attribute could be named `legend.tracegap` and would complement the current `legend.tracegroupgap`.

cc @nicolaskruchten @plotly/plotly_js

nicolaskruchten:
I don't mind in principle, but can't we get away without it? It seems to me that horizontal legends are just like vertical ones, no?

antoinerg:
> I don't mind in principle, but can't we get away without it?

Yes we can get away without it if we're happy with solely controlling the spacing between groups of items and NOT have control over the spacing between items.

>  It seems to me that horizontal legends are just like vertical ones, no?

In vertical legends, grouped items are stacked vertically. In horizontal legends, they are also stacked vertically. One may argue that they should be stacked horizontally (as done in the current PR). If we want to keep them stacked vertically even in horizontal legends, the following community PR https://github.com/plotly/plotly.js/pull/3628 is more appropriate.

cc @nicolaskruchten

nicolaskruchten:
> In horizontal legends, they are also stacked vertically

This is the current behaviour?

antoinerg:
> In horizontal legends, they are also stacked vertically

> > This is the current behaviour?

Yes it is. This is how horizontal legends with groups are rendered on `master`:
newplot (10)
where items are overflowing on the right (which is essentially what we're trying to fix #1913). Also, `legend.tracegroupgap` which is supposed to
> Sets the amount of vertical space (in px) between legend groups.

actually controls the horizontal spacing between all items if there is at least one group in the legend.

nicolaskruchten:
Well, if we currently stack them vertically then we can't change it, right? I think PR #3628 makes sense to me: keep vertically-aligned groups and wrap whole groups, honouring `tracegroupgap`. Any downsides to that @antoinerg  ?

antoinerg:
> I think PR #3628 makes sense to me: keep vertically-aligned groups and wrap whole groups, honouring tracegroupgap. Any downsides to that @antoinerg ?

I agree PR #3628 is less of a breaking change. The only downside I can think of is that vertically-aligned groups will take up more space than they would if stacked horizontally.

cc @nicolaskruchten 

Close in favor of PR #3628

PR #3612, opened 2019-03-06
CircleCI: improve caching for faster startup time

antoinerg:
This PR slightly improves the execution time of our CircleCi :

1. I removed the steps to store and restore the cache because they were ineffective (the cache key would never match). We can reintroduce them later if we really want to save off a few seconds out of running `npm i`.

2. Instead of doing a long git checkout process for every jobs, we pass the content of the `plotly.js` directory to subsequent jobs after `build`. This add rougly 5 seconds to the `persist_to_workspace` step. However, we reduce the startup time of all the following jobs by ~ 18 seconds:
newplot 22

This will make running tests in parallel more efficient and save us N * 18 seconds of computation time each run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - run:
           name: Install dependencies
           command: |
-            npm install
+            npm ci
       - run:
           name: List dependency versions
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,10 +14,6 @@ jobs:
     working_directory: ~/plotly.js
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - v{{ .Environment.CIRCLE_CACHE_VERSION }}-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
-            - v{{ .Environment.CIRCLE_CACHE_VERSION }}-deps-master-{{ checksum "package-lock.json" }}
       - run:
           name: Install dependencies
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -14,10 +14,6 @@ jobs:
     working_directory: ~/plotly.js
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - v{{ .Environment.CIRCLE_CACHE_VERSION }}-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
antoinerg: Notice here we use the checksum of "package-lock.json" but when we save the cache, we the checksum "package.json". Does two cache names will never match so this is why I removed this "dead" code.

etpinard:
Nice find. Great work!

I'm curious though. Can you try replacing our `npm install`  step with [`npm ci`](https://docs.npmjs.com/cli/ci.html), and see if we can :hocho: a few more seconds?

antoinerg:
> I'm curious though: can you try replacing our npm install step with npm ci, and see if we can hocho a few more seconds?

`npm ci` seems faster: roughly 6 seconds versus 15-18 seconds for `npm i`. I will add it to the current PR.

etpinard:
> `npm ci` seems faster: roughly 6 seconds versus 15-18 seconds for `npm i`

:tada: !

antoinerg:
Now using `npm ci` in 98c962d !

etpinard:
:dancer: :dancer:

alexcjohnson:
cc @Marc-Andre-Rivet re: `npm ci` - nice find @etpinard @antoinerg 🏆

Issue #3622, opened 2019-03-11
`violin` trace: why an area is calculated if all values are 0?


micosacak:
Please see my question in [stackoverflow](https://stackoverflow.com/questions/55094041/plotly-js-why-an-area-is-calculated-if-all-values-are-0?noredirect=1#comment96932560_55094041)

I have an array of values from 0-10 with and 20 groups in each. For some groups or categories all values are 0. When I use violin plot, there is an area calculated which is below 0, but I want to get thedifferent that I got from Seurat package VlnPlot. Is there a way to overcome that problem? Here is a part of java script code:

```
var my_violin = [];
var vln_layout = {};
var my_violin =[
{
    type:"violin",
    x:xvalues,
    y:my_categories,
    points:"all",
    pointpos:0,
    color:"black",
    width:0.75,
    marker:{color:"black",size:4},
    jitter:1,
    span:0,
    transforms:[
    {
        type:"groupby",
        groups:my_categories,
        styles:my_styles
    }
    ]
}
]

var vln_layout = {title: "my title",    xaxis:{showline: false,showgrid: false,zeroline: false,
    categoryorder: 'category ascending'}, yaxis: {showline: false,showgrid: false,zeroline:false},
    showlegend:true,legend:{traceorder:'normal',categoryorder:"ascending"}};

Plotly.newPlot('myDiv2', my_violin, vln_layout, {showSendToCloud: false});
```

antoinerg:
Thank you @micosacak for your interest in plotly.js and for reporting this. 

When reporting an issue, it is more convenient for the maintainers if you provide a Codepen showing the problem. I prepared one here: https://codepen.io/antoinerg/pen/QoqWBb. Can you fork it and add the missing variables `xvalues`, `my_categories` and `my_styles`?

Thank you!

micosacak:
Thanks. I do not how to use Codepen. Here are the values.
```
[...]

PR #3625, opened 2019-03-12
`violin`: handle zero standard deviation

antoinerg:
Closes issue #3622 

This is by far the most concise way I could fix the issue. I only added 4 lines to `src/traces/violin/calc.js` (starting from line 38).


src/traces/violin/calc.js
```diff
@@ -33,28 +33,36 @@ module.exports = function calc(gd, trace) {
...
-        for(var k = 0, t = span[0]; t < (span[1] + step / 2); k++, t += step) {
-            var v = kde(t);
-            cdi.density[k] = {v: v, t: t};
-            maxKDE = Math.max(maxKDE, v);
+        if (cdi.min === cdi.max) {
+            cdi.spanZero = true;
+            span = cdi.span = [cdi.min, cdi.max];
+            cdi.density = [{v:1, t: span[0]}];
+            maxKDE = Math.max(maxKDE, 1);
+        } else {
```
antoinerg: The following is exactly the same code we had before.

alexcjohnson:
Much better 🍻
Question though: what if someone *wants* a nonzero bandwidth despite the single-valued data? Ideally they should still be able to provide a `trace.bandwidth` and get that behavior. I might even suggest doing that in the `groups-over-matching-axes` and `violin-offsetgroups` mocks to restore the previous look and make it easier to see what's going on.

src/traces/violin/calc.js
```diff
@@ -33,28 +33,37 @@ module.exports = function calc(gd, trace) {
...
-        cdi.density = new Array(n);
-
-        for(var k = 0, t = span[0]; t < (span[1] + step / 2); k++, t += step) {
-            var v = kde(t);
-            cdi.density[k] = {v: v, t: t};
-            maxKDE = Math.max(maxKDE, v);
+        var span;
+
+        if(cdi.min === cdi.max) {
+            cdi.spanZero = true;
```
alexcjohnson: Is this used?

antoinerg:
> Question though: what if someone wants a nonzero bandwidth despite the single-valued data? Ideally they should still be able to provide a trace.bandwidth and get that behavior. I might even suggest doing that in the groups-over-matching-axes and violin-offsetgroups mocks to restore the previous look and make it easier to see what's going on.

Right... I was approaching this the wrong way. Closing in favor of https://github.com/plotly/plotly.js/pull/3626

PR #3626, opened 2019-03-12
`violin`: honor `bandwidth` attribute even if values in a trace are identical

### LINKED ISSUE #2779: Improve bandwidth calculation for single-value violins
From @alexcjohnson in https://github.com/plotly/plotly.js/pull/2775

> But if we think this is a real use case we should use other violins on the same subplot to pick a bandwidth that's reasonable for the whole collection, like we do with [histograms](https://github.com/plotly/plotly.js/blob/master/src/traces/histogram/calc.js#L350)

### LINKED ISSUE #3622: `violin` trace: why an area is calculated if all values are 0?

antoinerg:
Following botched PR https://github.com/plotly/plotly.js/pull/3625, this PR closes #3622 by honoring `trace.bandwidth` even if all the values are identical.

When no bandwidth is specified and all values are identical, it draws the violin as a line. Therefore, this PR also closes #2779!


antoinerg:
At the moment, the default bandwidth for a violin with zero standard deviation (ie. all values are identical) is 1. Therefore, you will get a significant violin area as shown in this [comment](https://github.com/plotly/plotly.js/issues/3622#issuecomment-471691858) which is not what other libraries do (like Seaborn for example). For people who don't know the details of violin plot (like me yesterday), this looks weird/wrong.

@alexcjohnson do you think we should greatly reduce the bandwidth when all values are identical so the violin collapses?

alexcjohnson:
Right, I think the *default* behavior with all identical values should be zero width, as you had in #3625. But if someone provides an explicit `bandwidth`, we should honor that as you do here.

antoinerg:
> Right, I think the default behavior with all identical values should be zero width, as you had in #3625. But if someone provides an explicit bandwidth, we should honor that as you do here.

With commit 5acb518 we have a mock showcasing both situations (identical values with and without explicit bandwidth). I had to explicitly set the bandwidth to 1 on an existing mock to preserve baselines intact.

test/image/mocks/groups-over-matching-axes.json
```diff
@@ -86,6 +86,9 @@
     }
   ],
   "layout": {
+    "template": {
+      "data": {"violin": [{"bandwidth": 1}]}
```
alexcjohnson: Nice use of `template` 🏆

alexcjohnson:
💃 
This seems like more than a bugfix, so merge if you're ready for 1.46 features.

etpinard:
@antoinerg looks like this PR resolves https://github.com/plotly/plotly.js/issues/2779 as well, correct?

test/image/mocks/groups-over-matching-axes.json
etpinard: Hmm. I think I would prefer updating the baselines instead of tweaking the JSON mocks.

antoinerg:
> @antoinerg looks like this PR resolves #2779 as well, correct?

Yes it does! I updated my first comment to reflect that. Nice catch :eyes:

test/image/mocks/groups-over-matching-axes.json
antoinerg: Done in 4698df0

alexcjohnson: Not a big deal, but to me the original baselines are a lot easier to look at and understand what's going on. But of course either way it'll serve the function of catching unintended changes.

etpinard:
good to merge :dancer:

dellshape:
Hi, updating here, passed by here because I got the same problem just now. But turns out now you can solve this by setting spanmode to "hard", therefore the "span goes from the sample's minimum to its maximum value."

PR #3628, opened 2019-03-12
Fix #1913 by pushing overflowed legend items to a new line

### LINKED ISSUE #1913: Horizontal legend wrapping is broken when legendgroups are used

nathanjenx:
Fixes the handling of horizontal grouped legend items see #1913. Maintains the vertical grouping and includes gaps between traces (see below screenshot)
image
@antoinerg @etpinard 

(edit: updated to show current behaviour, original available [here](https://user-images.githubusercontent.com/17784905/54193763-a4b98580-44b2-11e9-928d-df12b110357c.png))


antoinerg:
Thank you @nathanjenx for submitting this PR :tada: 

It seems like two mocks are not matching the baselines. Could you update the baselines by running:
```
$ npm run baseline legend_horizontal_groups
$ npm run baseline legendgroup_horizontal_wrapping
```

I will review the rest of your PR shortly :smile_cat:

antoinerg:
About `tracegroupgap`, its description says that it
>Sets the amount of vertical space (in px) between legend groups. 

As of right now, it also affects the horizontal spacing. Could you get rid of this behaviour by not adding `traceGroupGap` to the widths that are computed?

cc @nathanjenx

nathanjenx:
Removed `traceGroupGap` from all the width's & fixed up where the newline resets too (incase it's not 0) from:
` groupXOffsets[groupXOffsets.length - 1] = 0;`
to:
`groupXOffsets[groupXOffsets.length - 1] = groupXOffsets[0];`
@antoinerg

src/components/legend/draw.js
```diff
@@ -586,23 +586,58 @@ function computeLegendDimensions(gd, groups, traces) {
...
+                return a + b[0].height;
+            }, 0);
 
-            opts._width += opts.tracegroupgap + groupWidth;
+            maxWidth = Math.max(maxWidth, groupWidth);
+            maxHeight = Math.max(maxHeight, groupHeight);
+            maxItems = Math.max(maxItems, group.length);
+        });
+
+        maxHeight += opts.tracegroupgap || 5;
```
antoinerg: When `tracegroupgap` is `0`, which is falsey, it uses value `5` which is not what we want. I would :hocho: `|| 5`.

nathanjenx: Makes sense, I ripped it from https://github.com/plotly/plotly.js/blob/f279235ed99f3b3fda9805cdfd9b44378003eceb/src/components/legend/draw.js#L672

I guess it was put there to default to 5 which is redundant now it defaults to 10, should I 🔪that also?

antoinerg:
Thank you @nathanjenx for commit b9542c44673ec5eeba3524c3dd3ff68e279733ae which removes `tracegroupgap` from horizontal margin.

I now realize however that we should still have some horizontal spacing between items to be in line with what we have when there are no grouped items. To do so, we should add a `traceGap` of 5 px to be equivalent to [L672 of draw.js](https://github.com/nathanjenx/plotly.js/blob/f279235ed99f3b3fda9805cdfd9b44378003eceb/src/components/legend/draw.js#L672).

While you're at it, I think we should replace L672 from:
`var traceGap = opts.tracegroupgap || 5;` to `var traceGap = 5;`
Otherwise, `tracegroupgap` has an effect on legends without grouped items which shouldn't be the case.

src/components/legend/draw.js
antoinerg: See https://github.com/plotly/plotly.js/pull/3628#issuecomment-472135605 . Actually, if we reach that block, `opts.tracegroupgap` is still undefined since we only coerce the attribute if the legend has grouped items:
https://github.com/plotly/plotly.js/blob/7e68a2e56f5cfca1286163bda63b5f8263c43451/src/components/legend/defaults.js#L104

Thanks for cleaning that up :smile: !

nathanjenx:
> Thank you @nathanjenx for commit [b9542c4](https://github.com/plotly/plotly.js/commit/b9542c44673ec5eeba3524c3dd3ff68e279733ae) which removes `tracegroupgap` from horizontal margin.
> 
> I now realize however that we should still have some horizontal spacing between items to be in line with what we have when there are no grouped items. To do so, we should add a `traceGap` of 5 px to be equivalent to [L672 of draw.js](https://github.com/nathanjenx/plotly.js/blob/f279235ed99f3b3fda9805cdfd9b44378003eceb/src/components/legend/draw.js#L672).
> 
> While you're at it, I think we should replace L672 from:
> `var traceGap = opts.tracegroupgap || 5;` to `var traceGap = 5;`
> Otherwise, `tracegroupgap` has an effect on legends without grouped items which shouldn't be the case.

Is commit 51178ec2187db86eb17234eb171142f4a7e30af5 what you had in mind?

src/components/legend/draw.js
```diff
@@ -609,13 +609,15 @@ function computeLegendDimensions(gd, groups, traces) {
             maxItems = Math.max(maxItems, group.length);
         });
 
-        maxHeight += opts.tracegroupgap || 5;
+        maxHeight += opts.tracegroupgap;
+        var traceGroupGap = 5;
```
antoinerg: To prevent confusion, maybe we should call this variable `traceGap` instead of `traceGroupGap`.

Maybe you can change this in your next commit along with updated baselines (I expect at least one baseline will change)!

Thanks again @nathanjenx

src/components/legend/draw.js
```diff
@@ -586,23 +586,60 @@ function computeLegendDimensions(gd, groups, traces) {
...
+
+        maxHeight += opts.tracegroupgap;
+        var traceGroupGap = 5;
+        maxWidth += traceGroupGap;
+
+        var groupXOffsets = [opts._width];
+        var rowHeights = [];
+        var rowNum = 1;
+        for(var i = 0, n = groupData.length; i < n; i++) {
+            opts._width += maxWidth ;
```
antoinerg: Linter will complain about this extra white-space before the semicolon!

src/components/legend/draw.js
nathanjenx: No problemo, will get that sorted now!

archmoj:
Concerning the changes in [this baseline](https://github.com/plotly/plotly.js/pull/3628/files?short_path=3367b6a#diff-3367b6ae7c2fe5d6fbdfd3a6630f48a1), I am wondering one should find a way to keep the axis line where it was before?

antoinerg:
@nathanjenx The horizontal spacing issue is still present.

If I remove all `legendgroup` from your mock, I get the following:
newplot (11)

Now if I add a `legendgroup` to the first trace, I get:
newplot (12)

I expected the horizontal spacing to remain the same :thinking: 

> Concerning the changes in this baseline, I am wondering one should find a way to keep the axis line where it was before?

Indeed, it would be better if this baseline didn't change at all. @nathanjenx let me know if this is easy/possible!

Thank you again!

nathanjenx:
@antoinerg the horizontal spacing was off because I was updating the width then checking it then adding it again, rather than checking if we would be too long + then adding the width if we aren't.

@antoinerg, @archmoj the axis line was off as I originally added traceGroupGap to EVERY row, not just new ones. The changes to the horizontal spacing of the legend items you see now are due to the traceGap being set to 5 (not the 10 it was inheriting from opts.traceGroupGap).

Should be all sorted now!  🎉

antoinerg:
Thank you @nathanjenx for the thorough revision! It looks good to me :dancer:

test/image/mocks/legendgroup_horizontal_wrapping.json
```diff
@@ -0,0 +1,57 @@
+{
```
etpinard: For comparison, mock looks like: 

image


on master.

src/components/legend/draw.js
```diff
@@ -620,11 +658,12 @@ function computeLegendDimensions(gd, groups, traces) {
 
                 groupHeight += textHeight;
             });
-
-            opts._height = Math.max(opts._height, groupHeight);
         });
 
-        opts._height += 10 + borderwidth * 2;
+        opts._height = 10 + (borderwidth * 2) + (rowNum * maxHeight) + ((rowNum - 1) * traceGap);
```
etpinard: Thanks very much for the PR @nathanjenx ! After playing around on your branch, this here doesn't appear quite right. Is `legend.tracegroupgap` taken into account?

From your new `legendgroup_horizontal_wrapping` mock, I expect that calling `Plotly.relayout(gd, 'legend.tracegroupgap', 40)` would increase the bottom margin. But now, the legend is clipped:

image

antoinerg: Thank you so much @etpinard for catching this one! It should definitely be `opts.tracegroupgap` instead of `traceGap` at the end of line 663.

I'll let @nathanjenx fix!

nathanjenx: @antoinerg quite right, that was the case.
@etpinard, @antoinerg I've refactored the code to make it more immediately obvious where the values come from!

etpinard: Thanks @nathanjenx for the quick fix! :tada: 

Would you be interested in adding a jasmine test making sure `relayout(gd, 'legend.tracegroupgap', ??)` makes the bottom margin resize accordingly?

If so, add an `it` block to this [`describe`](https://github.com/plotly/plotly.js/blob/fc236e2802cdfe60fee1320de55cf80af2b11eda/test/jasmine/tests/legend_test.js#L593-L708) block, asserting `gd._fullLayout._size` should be enough. You can also try to assert the legend bounding box to :lock: this down. Thanks very much for your hard work :muscle:

etpinard: No news from @nathanjenx 

@antoinerg could you take over this one to bring it to the finish line? :checkered_flag:

You should be able to push to this PR after:

```
git remote add nathanjenx git@github.com:nathanjenx/plotly.js.git
git fetch nathanjenx fix1913-horizontal-legened-groups:fix1913-horizontal-legened-groups
git checkout fix1913-horizontal-legened-groups

# then
git push nathanjenx fix1913-horizontal-legened-groups
```

antoinerg: Done in 131145d

Thanks again @nathanjenx for the great work :tada:

src/components/legend/draw.js
```diff
@@ -586,23 +588,54 @@ function computeLegendDimensions(gd, groups, traces) {
...
-        var groupXOffsets = [opts._width];
+        var maxHeight = 0;
+        var maxWidth = 0;
         var groupData = groups.data();
 
-        for(var i = 0, n = groupData.length; i < n; i++) {
-            var textWidths = groupData[i].map(function(legendItemArray) {
+        var maxItems = 0;
+
+        groupData.forEach(function(group) {
```
etpinard: **NON-BLOCKING**

We could probably clean up these few lines a little:

- use for-loop here instead of `forEach`
- :zap: that double empty new line
- use [`Lib.aggNums`](https://github.com/plotly/plotly.js/blob/12ce7253e5a3995824ec514a42be01aadbfdb6ea/src/lib/stats.js#L28-L44) instead of `Math.max.apply` and `group.reduce`
- make the `for(var i = 0, n = groupData.length; i < n; i++)` loops `for(var i = 0; i < groupData.length; i++)`, i.e. no need to define `n` (I think).

etpinard:
@antoinerg Thanks for adding that test! I made one very non-blocking comment. Feel free to ignore it.

I'll let you the honours of merging this thing :dancer: :dancer: 

Thank again @nathanjenx :1st_place_medal:

src/components/legend/draw.js
antoinerg: Done in a5fa247 except for `group.reduce`.

Can I really replace `group.reduce` with  `Lib.aggNums`?

etpinard: > Can I really replace `group.reduce` with `Lib.aggNums`?

Maybe with something like:

```js
Lib.aggNums(function(a, b) { return a + b[0].height; }, 0, groups)
```

similar to [`Lib.mean`](https://github.com/plotly/plotly.js/blob/12ce7253e5a3995824ec514a42be01aadbfdb6ea/src/lib/stats.js#L54-L57), but you're right, that's probably an overkill here as all the `b[0].height` should be defined. :dancer:

nathanjenx:
@etpinard @antoinerg sorry for going dark at the end there! Cheers for pushing it over the line! 💃💃

PR #3634, opened 2019-03-13
CircleCI: run jobs in parallel

antoinerg:
This PR is a follow-up to https://github.com/plotly/plotly.js/pull/3612 to try to improve the runtime of our test suite and its robustness.

Here we do so by leveraging the `parallelism` option offered in CircleCI. I moved out `test-bundle` into its own job and proceeded to run `test-jasmine` and `test-image` in parallel.

The `test-image` routine had to modified for it to accept a list of mocks as command-line arguments (it used to only support one globbing pattern which wasn't suitable).

I selected a parallelism of 2 for a few jobs in order for them to run in ~ 4 minutes. Because we need to `build` first (~ 2 minutes), we end up with a total running time of roughly 6 minutes:
Screenshot_2019-03-13_17-52-09

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -44,8 +44,8 @@ case $1 in
     jasmine)
         set_tz
 
-        npm run test-jasmine -- --skip-tags=gl,noCI,flaky || EXIT_STATE=$?
-        npm run test-bundle || EXIT_STATE=$?
+        SUITE=$(circleci tests glob "$ROOT/test/jasmine/tests/*" | circleci tests split)
+        npm run test-jasmine -- $SUITE --skip-tags=gl,noCI,flaky || EXIT_STATE=$?
 
         exit $EXIT_STATE
         ;;
@@ -53,7 +53,7 @@ case $1 in
     jasmine2)
         set_tz
 
-        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl))
+        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl | circleci tests split))
 
         for s in ${SHARDS[@]}; do
             retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI
@@ -65,7 +65,7 @@ case $1 in
     jasmine3)
         set_tz
 
-        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=flaky))
+        SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=flaky | circleci tests split))
 
         for s in ${SHARDS[@]}; do
             retry npm run test-jasmine -- "$s" --tags=flaky --skip-tags=noCI
@@ -82,6 +82,12 @@ case $1 in
     image2)
         npm run test-export     || EXIT_STATE=$?
         npm run test-image-gl2d || EXIT_STATE=$?
+        exit $EXIT_STATE
+        ;;
+
+    bundle)
+        set_tz
+        npm run test-bundle || EXIT_STATE=$?
         exit $EXIT_STATE
         ;;
 
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -45,7 +45,7 @@ case $1 in
         set_tz
 
         SUITE=$(circleci tests glob "$ROOT/test/jasmine/tests/*" | circleci tests split)
-        npm run test-jasmine -- $SUITE --skip-tags=gl,noCI,flaky || EXIT_STATE=$?
+        npm run test-jasmine -- $SUITE --skip-tags=gl,noCI,flaky --showSkipped || EXIT_STATE=$?
 
         exit $EXIT_STATE
         ;;
@@ -56,7 +56,7 @@ case $1 in
         SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl | circleci tests split))
 
         for s in ${SHARDS[@]}; do
-            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI
+            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI --showSkipped
         done
 
         exit $EXIT_STATE
@@ -68,7 +68,7 @@ case $1 in
         SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=flaky | circleci tests split))
 
         for s in ${SHARDS[@]}; do
-            retry npm run test-jasmine -- "$s" --tags=flaky --skip-tags=noCI
+            retry npm run test-jasmine -- "$s" --tags=flaky --skip-tags=noCI --showSkipped
         done
 
         exit $EXIT_STATE
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -82,7 +82,6 @@ case $1 in
 
     image2)
         npm run test-export     || EXIT_STATE=$?
-        npm run test-image-gl2d || EXIT_STATE=$?
         exit $EXIT_STATE
         ;;
 
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -75,7 +75,8 @@ case $1 in
         ;;
 
     image)
-        npm run test-image      || EXIT_STATE=$?
+        SUITE=$(find $ROOT/test/image/mocks/ -type f -printf "%f\n" | circleci tests split)
+        npm run test-image -- $SUITE --filter || EXIT_STATE=$?
         exit $EXIT_STATE
         ;;
 
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "test-export": "node tasks/test_export.js",
     "test-syntax": "node tasks/test_syntax.js && npm run find-strings -- --no-output",
     "test-bundle": "node tasks/test_bundle.js",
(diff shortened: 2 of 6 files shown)


etpinard:
This is brilliant. Thanks for looking into this @antoinerg !

Before going any further, I'd like to get @bpostlethwaite 's approval. 

Off this branch, we have:

image

that is, a max of 8 containers are running simultaneously and 4 of those containers set `parallelism: 2`. @bpostlethwaite is this going to be take too juice much out of CircleCI plan?

antoinerg:
> that is, a max of 8 containers are running simultaneously and 4 of those containers set parallelism: 2. @bpostlethwaite is this going to be take too juice much out of our CircleCI plan?

@etpinard this is a very valid concern that I was going to bring up to your attention. Little correction: I think we have _3_ job with a `parallelism` of 2 so in total we use `5*1 + 3*2 = 11` containers. 

One thing that concerned me was whether CircleCI would wait for 11 containers to be available to even start or would it execute each job sequentially if we're short on containers. I assume it does the later but we should check to make sure we our tests do not get queued up for long period of times.

bpostlethwaite:
Yes this is fine. If we need more containers we'll buy them. Please let me know if there is significant queuing!

etpinard:
@antoinerg looks like we have the green light :rocket: 

Before merging this thing, I noticed that the tests that get skipped are no longer logged. For example, in `test-jasmine` container off master, we have:

image

If I remember correctly, the `spec` reporter

https://github.com/plotly/plotly.js/blob/61fb58ee9a342701abc3830dcfe845f760d6fa29/test/jasmine/karma.conf.js#L114

is responsible for this, along with its config object:

https://github.com/plotly/plotly.js/blob/61fb58ee9a342701abc3830dcfe845f760d6fa29/test/jasmine/karma.conf.js#L247-L257

----

I'm not sure how useful logging the skipped tests is. What do you think @antoinerg? Maybe we could add a CLI argument to allow devs to switch from 

```js
var reporters = ['dots', 'spec'];

// to
var reportes = ['progress'];
```

Thoughts?

etpinard:
Oh one more thing too, at present we shouldn't have to split the `gl2d_*` mocks from the others during `npm run test-image` (hmm but maybe the `heatmapgl` and `pointcloud` mocks still need their own test call).

So it would be nice, to bring all the pixel comparison tests into one container. Once that's done perhaps bump the parallelism to `3` in that container.

antoinerg:
> Oh one more thing too, at present we shouldn't have to split the `gl2d_*` mocks from the others during `npm run test-image` (hmm but maybe the `heatmapgl` and `pointcloud` mocks still need their own test call).

One big difference between them and "regular" mocks is that we render them in queue instead of in batch:
https://github.com/plotly/plotly.js/blob/12ce7253e5a3995824ec514a42be01aadbfdb6ea/package.json#L37-L38
Out of curiosity, what makes you think the following warning may not apply anymore?
https://github.com/plotly/plotly.js/blob/12ce7253e5a3995824ec514a42be01aadbfdb6ea/test/image/compare_pixels_test.js#L69

Anyway, I will test it out in a separate branch and let you know the result :smile_cat:

etpinard:
> > 
> 
> One big difference between them and "regular" mocks is that we render them in queue instead of in batch

Good eye. Yeah, we shouldn't have to run the `gl2d` image tests in a queue any more, but maybe the `heatmapgl` and `pointcloud` still behave weird.

etpinard:
> Out of curiosity, what makes you think the following warning may not apply anymore?

That one sounds obsolete.

antoinerg:
> Oh one more thing too, at present we shouldn't have to split the gl2d_* mocks from the others during npm run test-image (hmm but maybe the heatmapgl and pointcloud mocks still need their own test call).

> So it would be nice, to bring all the pixel comparison tests into one container. Once that's done perhaps bump the parallelism to 3 in that container.

Done in 609cdf8 and seems to work fine.

antoinerg:
> I'm not sure how useful logging the skipped tests is. What do you think @antoinerg? Maybe we could add a CLI argument to allow devs to switch from

I think it is useful to show the skipped tests. I added a flag for it in b99672b9af6d156aa237412d7ef92cefedb66b0c.

package.json
```diff
@@ -35,7 +35,6 @@
     "pretest": "node tasks/pretest.js",
     "test-jasmine": "karma start test/jasmine/karma.conf.js",
     "test-image": "node tasks/test_image.js",
-    "test-image-gl2d": "node tasks/test_image.js gl2d_* --queue",
```
etpinard: Nice find! 

Can you :zap: that `npm run test-image-gl2d` piece in:

```
    "test": "npm run test-jasmine && npm run test-bundle && npm run test-image && npm run test-image-gl2d && npm run test-syntax && npm run lint",
```

etpinard: and actually `test` should be:

```js
"test": "npm run test-jasmine -- --nowatch && npm run test-bundle && npm run test-image && npm run test-export && npm run test-syntax && npm run lint",
```

I think anybody actually uses this command, but let's keep it up-to-date for documentation purposes.

test/image/compare_pixels_test.js
```diff
@@ -129,34 +119,6 @@ function untestableFilter(mockName) {
...
- * having issues with dealing with the two different
- * program binding algorithm.
- *
- * The problem will be solved by switching all our
- * WebGL-based trace types to regl.
- *
- * More info here:
- * https://github.com/plotly/plotly.js/pull/1037
- */
-function sortGl2dMockList(mockList) {
```
etpinard: Looks like we'll have to keep this:

Running

```
npm run test-image
```

off this branch fails at `gl2d_pointcloud-basic`:

image

test/image/compare_pixels_test.js
```diff
@@ -120,7 +110,7 @@ else {
 function untestableFilter(mockName) {
     var cond = !(
         mockName === 'font-wishlist' ||
-        mockName.indexOf('gl2d_') !== -1 ||
+        // mockName.indexOf('gl2d_') !== -1 ||
```
etpinard: :zap: this line completely.

.circleci/test.sh
```diff
@@ -65,23 +65,29 @@ case $1 in
...
         ;;
 
     image2)
         npm run test-export     || EXIT_STATE=$?
-        npm run test-image-gl2d || EXIT_STATE=$?
+        exit $EXIT_STATE
+        ;;
+
+    bundle)
+        set_tz
```
etpinard: :zap: - no need to set the timezone in containers that run image tests (they have their own timezone in their docker container).

test/image/compare_pixels_test.js
```diff
@@ -48,37 +49,51 @@ var QUEUE_WAIT = 10;
...
+    allMockList = allMockList.concat(mockList);
+});
+
+// To get rid of duplicates
+Array.prototype.unique = function() {
+    return this.filter(function(value, index, self) {
+        return self.indexOf(value) === index;
+    });
+};
+allMockList = allMockList.unique();
```
etpinard: Please rewrite this as:

```js
// To get rid of duplicates
function unique(value, index, self) {
    reurn self.indexOf(value) === index;
};
allMockList = allMockList.filter(unique);
```

adding things to the prototype of builtin objects is a (very) bad pattern.

.circleci/test.sh
archmoj: @etpinard would you please provide a pointer to this?

etpinard: Not sure what else can I say.

test/image/compare_pixels_test.js
etpinard: ... and while at it can you :zap: the `gl2d` and `gl3d_cone` lines in comment above this function:

```js
/* Test cases:
 *
 * - font-wishlist
 * - all gl2d
 * - all mapbox
 * - gl3d_cone-*
 *
 * don't behave consistently from run-to-run and/or
 * machine-to-machine; skip over them for now.
 *
 */
```

Thanks!

etpinard:
image

a thing of beauty. Thanks @antoinerg - looks like you're almost there!

package.json
antoinerg: Done in be336489a93f6600c6ac668f4f7b46b323d99e53

test/image/compare_pixels_test.js
antoinerg: I could reproduce this issue locally! Nice catch :eyes: 

With be336489a93f6600c6ac668f4f7b46b323d99e53, I reintroduced `sortGl2dMockList` and running `npm run test-image` completed without error :tada:

test/image/compare_pixels_test.js
antoinerg: Done in be336489a93f6600c6ac668f4f7b46b323d99e53

test/image/compare_pixels_test.js
antoinerg: I agree! Sorry about that. Fixed in be336489a93f6600c6ac668f4f7b46b323d99e53

test/image/compare_pixels_test.js
etpinard: image

works for me too :tada:

etpinard:
Amazing work! :dancer: :dancer: 

Feel free to ignore https://github.com/plotly/plotly.js/pull/3634#pullrequestreview-218903730

PR #3637, opened 2019-03-14
Preserve marker size in scattergl traces with different glPixelRatio values

### LINKED ISSUE #3246: config option plotGlPixelRatio changes size of markers
As shown in this Codepen: https://codepen.io/anon/pen/aQwzEM

archmoj:
Fix #3246 for `scattergl` trace.
[Before codepen](https://codepen.io/anon/pen/aQwzEM).
[After codepen](https://codepen.io/MojtabaSamimi/pen/rRJgRK).

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "regl": "^1.3.11",
     "regl-error2d": "^2.0.6",
     "regl-line2d": "3.0.13",
-    "regl-scatter2d": "^3.1.3",
+    "regl-scatter2d": "git://github.com/gl-vis/regl-scatter2d.git#0fcb2a5b014ee1d4a743ed72bbcfb8ba05ace39b",
     "regl-splom": "^1.0.6",
     "right-now": "^1.0.0",
     "robust-orientation": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "regl": "^1.3.11",
     "regl-error2d": "^2.0.6",
     "regl-line2d": "3.0.13",
-    "regl-scatter2d": "git://github.com/gl-vis/regl-scatter2d.git#0fcb2a5b014ee1d4a743ed72bbcfb8ba05ace39b",
+    "regl-scatter2d": "git://github.com/gl-vis/regl-scatter2d.git#6a8f3576467f41bc2f8d6a4e78c646e933532439",
     "regl-splom": "^1.0.6",
     "right-now": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -104,7 +104,7 @@
     "regl": "^1.3.11",
     "regl-error2d": "^2.0.6",
     "regl-line2d": "3.0.13",
-    "regl-scatter2d": "^3.1.3",
+    "regl-scatter2d": "git://github.com/gl-vis/regl-scatter2d.git#6a8f3576467f41bc2f8d6a4e78c646e933532439",
```
archmoj: https://github.com/gl-vis/regl-scatter2d/compare/fix3246-glPixelRatio-change-marker-size

antoinerg:
Thanks @archmoj for looking into this issue! This might give a new life to my old Orca PR https://github.com/plotly/plotly.js/pull/3237 :smile:

Can you explain why we're completely dropping `glPixelRatio` from the `regl-scatter2d` module? I understand it wasn't working properly before but isn't it supposed to do something like control the resolution/sharpness of markers? @alexcjohnson was hinting at this in https://github.com/plotly/plotly.js/issues/3246#issuecomment-439175019 . I'm not really familiar with the 3D modules so it'd be nice if you could shed light upon this for me and tell me if `glPixelRatio` should be added in the future or not!

archmoj:
We don't completely drop `glPixelRatio` from `regl-scatter2d` and one could find the variable all over the place. But it looks that it was applied twice in the sizes. And since the `plotGlPixelRatio` is generally set to `2.0` on `plotly.js` it is replaced by `2.0` in the vertex shader programs so that the graphs may remain unchanged.

antoinerg:
> We don't completely drop glPixelRatio from regl-scatter2d and one could find the variable all over the place. But it looks that it was applied twice in the sizes. And since the plotGlPixelRatio is generally set to 2.0 on plotly.js it is replaced by 2.0 in the vertex shader programs so that the graphs may remain unchanged.

I was misled by the bad `grep` command I was using. Sorry about that! I now see we're still using `glPixelRatio` :+1: ! I will let @etpinard review this one. Thank you again :smile:

etpinard:
Cool. Thanks @archmoj 

Can you turn https://github.com/gl-vis/regl-scatter2d/compare/fix3246-glPixelRatio-change-marker-size into a PR and ping @dy 

Tagging this under 1.46.0

etpinard:
Dropping from 1.46.0

PR #3644, opened 2019-03-18
Bug fix errorbars on bar trace may disappear after switch modebar to select and double click

### LINKED ISSUE #3630: Error lines disappear on bar plots after box select and double click on the white space
Visit https://rreusser.github.io/plotly-mock-viewer/#bar_errorbars_inherit_color.
1. Select `Box Select` from the modebar.
2. Double click on graph white space.
[Codepen](https://codepen.io/MojtabaSamimi/pen/pYKqJL?editors=1000).

All errorbars (except one) would disappear.
Screenshot from 2019-03-12 11-02-29

archmoj:
Fix #3630.
Codepens:
[Before](https://codepen.io/MojtabaSamimi/pen/pYKqJL?editors=1000).
[After](https://codepen.io/MojtabaSamimi/pen/PLaXOK?editors=0010).
@plotly/plotly_js

### DIFF
--- a/src/traces/bar/style.js
+++ b/src/traces/bar/style.js
@@ -71,6 +71,8 @@ function styleOnSelect(gd, cd) {
         stylePointsInSelectionMode(s, trace, gd);
     } else {
         stylePoints(s, trace, gd);
+
+        Registry.getComponentMethod('errorbars', 'style')(s);
     }
 }
 
--- a/test/jasmine/tests/select_test.js
+++ b/test/jasmine/tests/select_test.js
@@ -688,6 +688,45 @@ describe('Click-to-select', function() {
         }
     });
 
+    it('should maintain style of errorbars after double click cleared selection (bar case)', function(done) {
+        Plotly.newPlot(gd, { // Note: this call should be newPlot not plot
+            data: [{
+                x: [0, 1, 2],
+                y: [100, 200, 400],
+                type: 'bar',
+                marker: {
+                    color: 'yellow'
+                },
+                error_y: {
+                    type: 'sqrt'
+                }
+            }],
+            layout: {
+                dragmode: 'select'
+            }
+        })
+        .then(function() {
+            var x = 100;
+            var y = 100;
+            drag([[x, y], [x, y]]); // first empty drag
+            return doubleClick(x, y); // then double click
+        })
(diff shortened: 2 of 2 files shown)


src/traces/bar/style.js
```diff
@@ -71,6 +71,8 @@ function styleOnSelect(gd, cd) {
         stylePointsInSelectionMode(s, trace, gd);
     } else {
         stylePoints(s, trace, gd);
+
+        Registry.getComponentMethod('errorbars', 'style')(s);
```
etpinard: Hmm. I'm not sure this is the correct fix. Look at https://codepen.io/etpinard/pen/gEKqgj?editors=1010 (the same as your _before_ example, but with scatter traces). Selection doesn't dim the error bars.

etpinard: Back in 1.36.0, selecting bars didn't dim the associated error bars: https://codepen.io/etpinard/pen/jJKdYN?editors=1000

We should probably eventually add a way to dim the error bars on selections (by adding a `[un]selected.error_(x|y).color` or by adding a trace-wide `[un]selected.opacity`), but that's for another day.

To _fix_ this bug, I think we should make sure error bars on bar traces don't get dimmed on selection by default.

archmoj: The bar version is different from `scatter` possibly because in `bar` the select deals with both bar points and errorbar points. So I still think it may be a good patch to solve the regression.

etpinard: .... and why should scatter and bar error bars behave differently?

etpinard: ... but yeah you're right that the regression is way worse than the inconsistency this fix brings up.

archmoj: > .... and why should scatter and bar error bars behave differently?

In `scatter` within `stylePoint` function `selectAll('path.point')` is applied; whereas in `bar` within `stylePoint` function `selectAll('path')` is applied which resulted in errorbars being selected after click.

etpinard:
Like you said https://github.com/plotly/plotly.js/issues/3630 is a (pretty bad) regression, so I'll give you a :dancer: to get this in the next release.

That said, please open a new issue describing the selected/unselected inconsistency between bar and scatter. Thanks!

Issue #3603, opened 2019-03-04
Hover Labels Overlap in 2d scatter plot


tfg250:
In my 2D scatter plot, the hover labels for closely spaced traces overlap when I think they should be stacked. It appears to be related to the level of zoom. Is the intended behavior for the labels to overlap when zoomed out?

A quick example is here:
[https://plot.ly/~tfg250/141](https://plot.ly/~tfg250/141)

In this plot, the initial view shows this:
![image](https://user-images.githubusercontent.com/15915420/53747764-fc803d00-3e71-11e9-83d0-2815745e6c2d.png)

Zooming in a bit and it changes to the expected view:
![image](https://user-images.githubusercontent.com/15915420/53747818-1a4da200-3e72-11e9-9a03-b402c5b2da53.png)

python code used to generate this plot:
```python
traces=[]
for i in range(10):
    x=[i,i+1]
    y=[i,i+1]
    traces.append(go.Scatter(x=x,y=y))

for i in range(10):
    x=[i,i+1]
    y=[i+1,i+2]
    traces.append(go.Scatter(x=x,y=y))

layout = go.Layout(xaxis=dict(range=[0, 100]),yaxis=dict(range=[0, 100]))

fig = go.Figure(data=traces,layout=layout)
py.plot(fig, filename='test.html')
```

antoinerg:
Thank you @tfg250 for reporting this issue!

antoinerg:
Codepen for this issue: https://codepen.io/antoinerg/pen/MxEjeo

PR #3645, opened 2019-03-18
Fix avoid-overlap hover label algo when "too close" label get filtered out

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3603

before: https://codepen.io/antoinerg/pen/MxEjeo

Peek 2019-03-18 15-39

after: https://codepen.io/etpinard/pen/LaBpYm

Peek 2019-03-18 16-04

In this example, we enter this block:

https://github.com/plotly/plotly.js/blob/fc236e2802cdfe60fee1320de55cf80af2b11eda/src/components/fx/hover.js#L872-L877

where `hoverData` gets filtered down showing two visible hover labels, but isn't updated for the downstream `hoverAvoidOverlaps` call where the "old" (longer) `hoverData` array is used here:

https://github.com/plotly/plotly.js/blob/fc236e2802cdfe60fee1320de55cf80af2b11eda/src/components/fx/hover.js#L1099-L1114

cc @plotly/plotly_js

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -704,7 +704,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
     var hoverLabels = createHoverText(hoverData, labelOpts, gd);
 
-    hoverAvoidOverlaps(hoverData, rotateLabels ? 'xa' : 'ya', fullLayout);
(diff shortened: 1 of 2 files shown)


antoinerg:
Thank you for the fix @etpinard ! It looks good to me :dancer:

Issue #3619, opened 2019-03-08
Compare Mode Bar button is for 'x' axis not 'y'axis when all Scatter Chart traces are Horizontal


PaulRCooper:
The compare mode bar button behaviour is for 'x' axis rather than even when you set all the traces to horizontal (orientation of 'h'). 

If you set the traces as horizontal the Mode Bar and set the hovermode to 'y' the button is unselected (so is the closest button) even though the expected 'y' hovermode behaviour occurs when hovering.

See example at [https://jsbin.com/wibivof](https://jsbin.com/wibivof). Here I have 3 charts: 

1.  The first is a horizontal bar chart showing the correct and expected behaviour in both the chart and the button. 

1. The second chart shows a horizontal scatter chart (all traces are horizontal) with hovermode set to 'y' and you get the correct behaviour on hovering but the button is unselected. If you click on the compare button you get 'x' axis compare behaviour instead. To get 'y' mode behaviour requires a refresh of the page.

1. The third chart is a copy of the second chart but includes a workaround to force the correct behaviour using the plotly_afterplot event to update the hovermode and the button state.

Possibly relates to issue [#778](https://github.com/plotly/plotly.js/issues/778)

etpinard:
The second chart shows a horizontal scatter chart (all traces are horizontal) with hovermode set to 'y' and you get the correct behaviour on hovering but the button is unselected. If you click on the compare button you get 'x' axis compare behaviour instead. To get 'y' mode behaviour requires a refresh of the page.

Behavior w/o setting `hovermode` to `'y'` is in: https://codepen.io/etpinard/pen/eXWXyy

Thanks for the report!

PR #3646, opened 2019-03-18
Fix hovermode default logic for stacked scatter traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3619 - so that by default graphs with horizontally stacked scatter traces get the correct `hovermode: 'y'` default (that is, until https://github.com/plotly/plotly.js/issues/778 in v2 :smirk: )

cc @plotly/plotly_js note that `orientation` is only coerced into `fullData` once per stack. Unnecessary keys are removed [here](
https://github.com/plotly/plotly.js/blob/05884832cb59c9ec159141b791a4645d09b2d7c3/src/traces/scatter/stack_defaults.js#L86).

### DIFF
--- a/src/components/fx/layout_defaults.js
+++ b/src/components/fx/layout_defaults.js
@@ -28,7 +28,7 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
         } else {
             // flag for 'horizontal' plots:
             // determines the state of the mode bar 'compare' hovermode button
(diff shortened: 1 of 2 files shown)


antoinerg:
Codepen before: https://codepen.io/etpinard/pen/eXWXyy
Codepen after: https://codepen.io/antoinerg/pen/LaBNmz

antoinerg:
Thank you @etpinard for improving the default for horizontally stacked scatter traces and for the clear test!

:dancer:

PR #3652, opened 2019-03-19
histogram: remove gap when `barmode` is relative

### LINKED ISSUE #3649: bargap in barmode=relative
If I make a multi-trace histogram with `barmode` set to `"relative"`, I would expect the bargap to stay 0, like it does in `"stack"`.

Per @alexcjohnson:

> We should really coerce barmode first, then invoke this logic only for `barmode !== 'group'`. And of course update it when bar `stackgroup` becomes a thing…

antoinerg:
Fixes #3649 but I am not sure this is what @alexcjohnson had in mind!

cc @nicolaskruchten


src/traces/bar/layout_defaults.js
```diff
@@ -46,7 +48,6 @@ module.exports = function(layoutIn, layoutOut, fullData) {
 
     if(!hasBars) return;
```
etpinard: we should 

```js
delete layoutOut.barmode;
```

when `hasBars` is false, so that it doesn't pollute `fullLayout`.

src/traces/bar/layout_defaults.js
```diff
@@ -46,7 +46,10 @@ module.exports = function(layoutIn, layoutOut, fullData) {
         }
     }
 
-    if(!hasBars) return;
+    if(!hasBars) {
+        delete layoutOut.barmode;
```
etpinard: `++` a jasmine test in `bar_test.js` plz :smirk:

antoinerg: a jasmine test that checks `_fullLayout` doesn't have a `barmode`?

etpinard: yeah, that and the converse. Similar to https://github.com/plotly/plotly.js/blob/ef932cfad9eb35fb7986a3740fdb42464d6e8b7f/test/jasmine/tests/bar_test.js#L214-L228

antoinerg: Done in bc3890b

alexcjohnson:
That's exactly what I had in mind, thx!

etpinard:
Ok, let's start merging things for 1.46.0 

:dancer: :dancer: :dancer:

nicolaskruchten:
Thanks guys!

PR #3591, opened 2019-03-01
Fixing cone plots lighting & rendering on some hardware including CircleCI image test 

archmoj:
The `gl-cone3d` module in this PR is aligned with the latest `gl-mehs3d` code in regards with lighting position, shading algorithms. Also some unused files, codes and parameters removed from the `gl-cone3d` code. 
Finally the indexing and floating points are debugged & fixed so that the baselines which include the `cone` trace type could be tested using `CircleCI` image test.
@plotly/plotly_js

PR #3656, opened 2019-03-20
No need to test `cone` traces during pre-release checklist

etpinard:
a follow-up of https://github.com/plotly/plotly.js/pull/3591

cc @archmoj


archmoj:
:dancer:

Issue #3618, opened 2019-03-08
Plotly_unhover not triggered following restyle in plotly_hover in Pie Charts


PaulRCooper:
Hi,

I found a issue with restyling the chart in plotly_hover events stops the triggering of plotly_unhover event and also stops the hover text from disappearing. Please note it works in other chart types.

I have created an example showing the issue with the first chart showing the issue, a second chart without restyling and a third showing it working with restyling in a Bar Chart. In all cases I have added code to write "hover" or "unhover" to the console in the appropriate event.

Example web-page can be found here: [https://jsbin.com/pavopaq](https://jsbin.com/pavopaq)

PaulRCooper:
No love for this one? Is more information required?

antoinerg:
Thank you @PaulRCooper for clearly reporting this issue!

PR #3662, opened 2019-03-22
Fixup pie/sunburst unhover post plot/calc restyle calls

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3618 and another (pretty bad) pie bug, to be merged into the sunburst PR in https://github.com/plotly/plotly.js/pull/3594

Updated jsbin: https://output.jsbin.com/lesidaxoru/1

cc @plotly/plotly_js

### DIFF
--- a/src/traces/pie/base_plot.js
+++ b/src/traces/pie/base_plot.js
@@ -16,8 +16,7 @@ exports.name = 'pie';
 exports.plot = function(gd) {
     var Pie = Registry.getModule('pie');
     var cdPie = getModuleCalcData(gd.calcdata, Pie)[0];
-
-    if(cdPie.length) Pie.plot(gd, cdPie);
+    Pie.plot(gd, cdPie);
 };
 
 exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout) {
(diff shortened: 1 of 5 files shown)


src/traces/pie/plot.js
```diff
@@ -313,12 +313,12 @@ function attachFxHandlers(sliceTop, gd, cd) {
 
     // hover state vars
     // have we drawn a hover label, so it should be cleared later
-    var hasHoverLabel = false;
+    if(!('_hasHoverLabel' in trace)) trace._hasHoverLabel = false;
```
archmoj: What about using the line below?
<pre>trace._hasHoverLabel = !!trace._hasHoverLabel;</pre>

etpinard: I don't want to cast `undefined` to `false`.

src/traces/pie/base_plot.js
```diff
@@ -16,8 +16,7 @@ exports.name = 'pie';
 exports.plot = function(gd) {
     var Pie = Registry.getModule('pie');
     var cdPie = getModuleCalcData(gd.calcdata, Pie)[0];
-
-    if(cdPie.length) Pie.plot(gd, cdPie);
+    Pie.plot(gd, cdPie);
```
archmoj: @etpinard Could you please elaborate why the condition check is no longer required here?

etpinard: Because we need `Pie.plot` to `.exit().remove()` the old pie traces.

archmoj:
@etpinard Thanks.
Looks very good.
:dancer:

PR #3661, opened 2019-03-22
Clean CI run of gl hovers

archmoj:
A follow up of PR #3585 now to help run the `gl+hover` tests on the CI much more smoothly! Recently we had to re-run `jasmine2` container few times after it failed several times which was not optimal. 
@plotly/plotly_js

PR #3599, opened 2019-03-04
Adding third container for jasmine tests to avoid flaky fails

archmoj:
Fixes #3587.
@plotly/plotly_js

PR #3667, opened 2019-03-23
Rearrange jasmine test retries

archmoj:
A follow up of #3661 and #3599 in order to have more robust image tests with less retries
- [x] applied different retry numbers for gl and flaky containers
- [x] reduced number of retries for jasmine-2 test
- [x] reduced number of retries for jasmine-3 test
- [x] set shards limit to 1 i.e. helped fix various side effects between different tests
- [x] removed `noCI` tags from a number of tests
- [x] adjusted timeouts of some tests to run on CI
@plotly/plotly_js

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -6,13 +6,13 @@ set +o pipefail
 
 ROOT=$(dirname $0)/..
 EXIT_STATE=0
-MAX_AUTO_RETRY=5
 
 log () {
     echo -e "\n$1"
 }
 
 # inspired by https://unix.stackexchange.com/a/82602
+MAX_AUTO_RETRY=1
 retry () {
     local n=1
 
@@ -54,9 +54,9 @@ case $1 in
         set_tz
 
         SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=gl))
-
+        MAX_AUTO_RETRY=2
         for s in ${SHARDS[@]}; do
-            retry npm run test-jasmine -- "$s" --tags=gl --skip-tags=noCI
+            retry npm run test-jasmine -- "${s}" --tags=gl --skip-tags=noCI
         done
 
         exit $EXIT_STATE
@@ -66,9 +66,9 @@ case $1 in
         set_tz
 
         SHARDS=($(node $ROOT/tasks/shard_jasmine_tests.js --tag=flaky))
-
+        MAX_AUTO_RETRY=5
         for s in ${SHARDS[@]}; do
-            retry npm run test-jasmine -- "$s" --tags=flaky --skip-tags=noCI
+            retry npm run test-jasmine -- "${s}" --tags=flaky --skip-tags=noCI
         done
 
         exit $EXIT_STATE
--- a/tasks/shard_jasmine_tests.js
+++ b/tasks/shard_jasmine_tests.js
@@ -15,7 +15,7 @@ var argv = minimist(process.argv.slice(2), {
         limit: ['l'],
     },
     default: {
-        limit: 20
+        limit: 1
     }
 });
 
--- a/test/jasmine/tests/gl2d_plot_interact_test.js
+++ b/test/jasmine/tests/gl2d_plot_interact_test.js
@@ -215,7 +215,7 @@ describe('Test gl plot side effects', function() {
         .then(done);
     });
 
-    it('@noCI @gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
+    it('@gl should fire *plotly_webglcontextlost* when on webgl context lost', function(done) {
         var _mock = Lib.extendDeep({}, require('@mocks/gl2d_12.json'));
 
         function _trigger(name) {
@@ -551,7 +551,7 @@ describe('Test gl2d plots', function() {
         .then(done);
     });
 
-    it('@noCI @gl should display selection of big number of regular points', function(done) {
+    it('@gl should display selection of big number of regular points', function(done) {
         // generate large number of points
         var x = [];
         var y = [];
@@ -582,7 +582,7 @@ describe('Test gl2d plots', function() {
         .then(done);
     });
 
-    it('@noCI @gl should display selection of big number of miscellaneous points', function(done) {
+    it('@gl should display selection of big number of miscellaneous points', function(done) {
         var colorList = [
             '#006385', '#F06E75', '#90ed7d', '#f7a35c', '#8085e9',
             '#f15c80', '#e4d354', '#2b908f', '#f45b5b', '#91e8e1',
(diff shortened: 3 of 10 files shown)


tasks/shard_jasmine_tests.js
```diff
@@ -15,7 +15,7 @@ var argv = minimist(process.argv.slice(2), {
         limit: ['l'],
     },
     default: {
-        limit: 20
+        limit: 1
```
etpinard: So each test suite is runned separately?

archmoj: Yes. That's exactly what I wanted there for `gl` and `flaky`. I noticed e.g. setting timeout in one suite may have an effect on another suite. After having this in place the tests started to pass properly.

etpinard:
@archmoj I'm a little puzzled by this PR. If it makes our tests on CI less flaky than great, but I'd like to why and _how much_ less flaky they get.

> * applied different retry numbers for gl and flaky containers
> reduced number of retries for jasmine-2 test
>  reduced number of retries for jasmine-3 test

Why does this help?

> * set shards limit to 1 i.e. helped fix various side effects between different tests

That sounds like an overkill to me. For example, if our tests are less flaky, but take 5x longer to run, this won't help us much. So yeah I'm asking you: after say CI 20 runs, are our tests 2x, 5x, 10x less flaky? How much slower are they?

> * removed `noCI` tags from a number of tests such as gl2d lasso/select & parcoods.

:tada: nice!

> * adjusted timeouts of some tests to run on CI
>   @plotly/plotly_js

:ok_hand: 

--------

All in all, https://github.com/plotly/plotly.js/pull/3634 sounds more promising. It might be a good idea to wait until @antoinerg is done with that one before spending more time trying to monkey-patch our tests.

archmoj:
@etpinard it seems more `flaky` tests are passing in the first run. The timing are not too bad. And now with @antoinerg great PR #3634 being merged we have: 
Screenshot from 2019-03-28 22-59-33

etpinard:
> it seems more `flaky` tests are passing in the first run

Cool. Can you re-run the tests 10 times and see what's the success rate?

archmoj:
After 10 runs the success rate seems to be around 80 percent i.e. including no `flaky` suite fails.
https://circleci.com/gh/plotly/workflows/plotly.js/tree/reduce-num-retries-in-jasmine
Screenshot from 2019-04-01 12-54-30

etpinard:
> After 10 runs the success rate seems to be around 80 percent

Is that better than on `master`?

test/jasmine/tests/gl2d_plot_interact_test.js
```diff
@@ -20,6 +20,8 @@ function countCanvases() {
     return d3.selectAll('canvas').size();
 }
 
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 5000;
```
etpinard: Does this apply only to this test suite or to all test suite that get bundled?

tasks/shard_jasmine_tests.js
etpinard: Can you leave to default to `20` and pass a `--limit 1` in the relevant test commands in `.circleci/test.sh`.

Do all `npm run test-jasmine` commands need to have `--limit 1`?

.circleci/test.sh
```diff
@@ -6,13 +6,13 @@ set +o pipefail
...
 ROOT=$(dirname $0)/..
 EXIT_STATE=0
-MAX_AUTO_RETRY=5
 
 log () {
     echo -e "\n$1"
 }
 
 # inspired by https://unix.stackexchange.com/a/82602
+MAX_AUTO_RETRY=1
```
etpinard: Why would we want to lower `MAX_AUTO_RETRY`?

archmoj: It looks flaky tests (jasmine3) sometime required more retries. The gl tests on the other hand are slow and may not need that number of retries. Specially if they fail for a good reason, we don't want to wait too long to be notified that the test is actually failed.

test/jasmine/tests/gl2d_plot_interact_test.js
archmoj: Good question. I noticed when using shard with high limit number (e.g. 20), changing a timout in one test has an impact on the other tests in another suites. That may be one reason to try to keep them separated. Anyway a better solution may be to have a timeout setup for every suite/describe block to avoid side effects?

etpinard: > Anyway a better solution may be to have a timeout setup for every suite/describe block to avoid side effects?

Yes, that's a better solution.

.circleci/test.sh
etpinard: > The gl tests on the other hand are slow and may not need that number of retries.

Well, I suspect the gl tests are slow (now) because you reduced the number of test per shards to 1.

To me, having `MAX_AUTO_RETRY=5` for all retry loops is the best of both world. Sufficient retry attempt, but failing runs that take to long to exit.

archmoj:
> > After 10 runs the success rate seems to be around 80 percent
> 
> Is that better than on `master`?

Always hard to tell which one is better. On this branch less noCI flags are applied with less retries. I hope that [less is more](https://en.wikipedia.org/wiki/Less_is_more) and that we can have even more robust tests.

test/jasmine/tests/gl2d_plot_interact_test.js
archmoj: @etpinard @antoinerg Any default value there to start with? Should they be set for every describe block? Or having them in every suites is only required?

etpinard:
> On this branch less noCI flags are applied with less retries

I agree, there's a lot of _good_ on this branch, no doubt.

Here are my recommendations (to get this thing merged):

- Bring back `MAX_AUTO_RETRY=5` for all `retry` loops
- Bring back the `limit: 20` default in the shard utility
- Lower the shard `limit` for the non-`@gl` test-jasmine commands
- Bump the timeout in `gl2d_plot_interact_test.js` per `it` block as opposed to that `jasmine.DEFAULT_TIMEOUT_INTERVAL = 5000;` line in the file scope

archmoj:
> I agree, there's a lot of _good_ on this branch, no doubt.
> 
> Here are my recommendations (to get this thing merged):
> 
> * Bring back `MAX_AUTO_RETRY=5` for all `retry` loops
> * Bring back the `limit: 20` default in the shard utility
> * Lower the shard `limit` for the non-`@gl` test-jasmine commands
> * Bump the timeout in `gl2d_plot_interact_test.js` per `it` block as opposed to that `jasmine.DEFAULT_TIMEOUT_INTERVAL = 5000;` line in the file scope

@etpinard Thanks for the recommendations. 
Applied in https://github.com/plotly/plotly.js/pull/3667/commits/24aafc81c698d60f26c4e73c592040874451b2f3.

etpinard:
Ok, let's get this in :dancer:

PR #3678, opened 2019-03-25
ensure we create only one `modebar-container`

### LINKED ISSUE #3677: Modebar buttons not generating correctly
Some issues I noticed on `master`, most likely post https://github.com/plotly/plotly.js/pull/3589

### Multiple `<div.modebar-container>` 

image

One should be enough right?

### More `<div.modebar-container>` keep getting added on `newPlot`

(and possible other update signatures)

image

----

We'll need to fix this one before releasing 1.46.0

antoinerg:
Closes #3677

- [x] Test that the final DOM sandwich is the one we want ( hoverlabels > modebar > the rest )

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -3802,6 +3802,7 @@ function makePlotFramework(gd) {
         .classed('gl-container', true);
 
     fullLayout._paperdiv.selectAll('.main-svg').remove();
+    fullLayout._paperdiv.selectAll('.modebar-container').remove();
 
     fullLayout._paper = fullLayout._paperdiv.insert('svg', ':first-child')
         .classed('main-svg', true);
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -3802,7 +3802,7 @@ function makePlotFramework(gd) {
         .classed('gl-container', true);
 
     fullLayout._paperdiv.selectAll('.main-svg').remove();
-    fullLayout._paperdiv.selectAll('.modebar-container').remove();
+    fullLayout._paperdiv.select('.modebar-container').remove();
 
     fullLayout._paper = fullLayout._paperdiv.insert('svg', ':first-child')
         .classed('main-svg', true);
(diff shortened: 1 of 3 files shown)


src/plot_api/plot_api.js
```diff
@@ -3802,6 +3802,7 @@ function makePlotFramework(gd) {
         .classed('gl-container', true);
 
     fullLayout._paperdiv.selectAll('.main-svg').remove();
+    fullLayout._paperdiv.selectAll('.modebar-container').remove();
```
antoinerg: We remove the existing modebar because:
https://github.com/plotly/plotly.js/blob/3e9d162cc4dd9f5b54f3d8e375b46e55fbfb82dc/src/plot_api/plot_api.js#L3792-L3796

etpinard: `.select` should suffice here, as there's only one `<div.modebar-container>` per graph.

antoinerg: Done in 567eddedd5ef3e1fd9e45ee0c03a6cea5d279d20

antoinerg: Done in 567eddedd5ef3e1fd9e45ee0c03a6cea5d279d20

test/jasmine/tests/hover_label_test.js
```diff
@@ -3221,6 +3221,36 @@ describe('hovermode defaults to', function() {
...
+            var infolayer = document.getElementsByClassName('infolayer');
+            var modebar = document.getElementsByClassName('modebar-container');
+            var hoverlayer = document.getElementsByClassName('hoverlayer');
+
+            expect(infolayer.length).toBe(1);
+            expect(modebar.length).toBe(1);
+            expect(hoverlayer.length).toBe(1);
+
+            var compareMask = infolayer[0].compareDocumentPosition(modebar[0]);
+            expect(compareMask).toBe(Node.DOCUMENT_POSITION_FOLLOWING, '.modebar-container appears after the .infolayer');
```
etpinard: Wow. I didn't know that thing existed. NICE!

antoinerg: Me neither! Just found out about them today: https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition :smile:

etpinard:
Thanks for the quick fix :dancer:

Issue #3651, opened 2019-03-19
same-position/within-trace bar placement when barmode=group


nicolaskruchten:
Consider the following pen: https://codepen.io/nicolaskruchten/pen/RdYWdz?editors=0010

The red and cyan bars aren't being shown because within each trace at the same position, the bars are behaving as if the barmode was "overlay". I believe that the default behaviour should be "relative" instead because it's the only one that always shows all the data!). 

If we can't agree that the current behaviour is a bug, then we should create a new attr that defaults to "overlay" but allows us to set it to "relative" or "stack" as required.

etpinard:
Interestingly under `barmode: 'relative'`, we appear to do the right thing already: https://codepen.io/etpinard/pen/GeeXmQ?editors=0010

This might just be an issue with `barmode: 'group'`.

nicolaskruchten:
Yes, this is exclusively a problem with barmode=group.

etpinard:
@nicolaskruchten I'm starting to think we could merge a "fix" for this bug w/o adding a new attribute.

Here's my first attempt: https://github.com/plotly/plotly.js/compare/unhide-grouped-bars-within-trace - which doesn't break any existing test.

Your codepen now looks like: https://codepen.io/etpinard/pen/YggmQx?editors=0010

![image](https://user-images.githubusercontent.com/6675409/54950357-7ce90980-4f17-11e9-8e26-f13ac10cd7b3.png)


and note that the new logic only has an effect when `barmode: 'group'`, so under `barmode: 'overlay'` we (still) get:

![image](https://user-images.githubusercontent.com/6675409/54950395-8e321600-4f17-11e9-9d79-0f510869285b.png)

nicolaskruchten:
This is pretty much my ideal fix I think, without looking at the implementation

PR #3680, opened 2019-03-25
Unhide overlapping grouped bars within trace

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3651 - where I decided to not add another attribute as the problem described in https://github.com/plotly/plotly.js/issues/3651 only affects bar graphs with `barmode: 'group'`. See https://github.com/plotly/plotly.js/issues/3651#issuecomment-476356081 for more on the topic.


src/traces/bar/cross_trace_calc.js
```diff
@@ -57,52 +57,57 @@ function setGroupPositions(gd, pa, sa, calcTraces) {
...
+            // exclude from the group those traces for which the user set an offset
+            excluded = [];
+            included = [];
+            for(i = 0; i < calcTraces.length; i++) {
+                calcTrace = calcTraces[i];
+                fullTrace = calcTrace[0].trace;
+
+                if(fullTrace.offset === undefined) included.push(calcTrace);
+                else excluded.push(calcTrace);
+            }
```
archmoj: Could we put this logic in a function and reuse it for the case of `relative` below?

src/traces/bar/cross_trace_calc.js
```diff
@@ -201,12 +214,12 @@ function setGroupPositionsInGroupMode(gd, pa, sa, calcTraces) {
...
-    var stack = barmode === 'stack';
-    var relative = barmode === 'relative';
     var barnorm = fullLayout.barnorm;
-    var separateNegativeValues = relative;
-    var dontMergeOverlappingData = !(barnorm || stack || relative);
-    var sieve = new Sieve(calcTraces, separateNegativeValues, dontMergeOverlappingData);
+
+    var sieve = new Sieve(calcTraces, {
+        separateNegativeValues: barmode === 'relative',
+        dontMergeOverlappingData: !(barnorm || barmode === 'stack' || barmode === 'relative')
```
archmoj: The function now has a more clear interface block which is great. 
Now that the keys would be presented in the `.min` files could we zip `separateNegativeValues` into e.g. `separateNegVals` and `dontMergeOverlappingData` into a short form?

etpinard:
Thanks for the review @archmoj - I'll until the `waterfall` PR is merged before addressing the comments.

src/traces/bar/cross_trace_calc.js
etpinard: Done in c019c5ca8

src/traces/bar/cross_trace_calc.js
etpinard: I'll pass if you don't mind. That would a pretty ugly function (we would have to pass functions `setGroupPositionsInGroupMode`, etc) and only save a few lines.

Issue #3654, opened 2019-03-19
pie chart hover label hidden


destradafilm:
long hoverlabels get hidden outside of plot boundries instead of making sure they fit within the plot container's range

the label for this slice in the pie chart is '1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16' but you can only see from `7` onward. 

![Screen Shot 2019-03-19 at 4 26 25 PM](https://user-images.githubusercontent.com/6043499/54646144-cf33b180-4a63-11e9-82d4-57c3eed53f6e.jpg)

Maybe the label placement can be different and if the label is wider than the chart, it can wrap text?

etpinard:
Thanks for posting!

What are your graph's width, height and margins here?

Or better yet, can you share a reproducible codepen showing the problem? Thank you!

destradafilm:
Hi @etpinard 

Here is a codepen with the example. Thanks for taking a look

https://codepen.io/destrada/pen/BbbGgJ

etpinard:
I'll submit a patch that constrains the hover labels to fit inside the graph's bounding box, but just so you know your long green label still won't fit:

destradafilm:
it's been almost 5 years since I opened this case - I came across the same bug again. Here is a codepen
https://codepen.io/destrada/pen/NWmxrVe

screenshot of the label being clipped out of visible chart area
<img width="745" alt="Screen Shot 2024-03-07 at 3 28 25 PM" src="https://github.com/plotly/plotly.js/assets/6043499/a2833bd3-66a9-48e0-9655-0f51b5d69f72">

PR #3681, opened 2019-03-25
Fit center-aligned hover labels within graph's horizontal span

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3654

updated codepen: https://codepen.io/etpinard/pen/bZJNBj

cc @plotly/plotly_js

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -1062,13 +1062,22 @@ function createHoverText(hoverData, opts, gd) {
             d.pos = hty;
             anchorStartOK = htx + dx / 2 + txTotalWidth <= outerWidth;
             anchorEndOK = htx - dx / 2 - txTotalWidth >= 0;
+
             if((d.idealAlign === 'left' || !anchorStartOK) && anchorEndOK) {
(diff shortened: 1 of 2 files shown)


src/components/fx/hover.js
```diff
@@ -1062,13 +1062,22 @@ function createHoverText(hoverData, opts, gd) {
...
                 d.anchor = 'start';
-            } else d.anchor = 'middle';
+            } else {
+                d.anchor = 'middle';
+
+                var txHalfWidth = txTotalWidth / 2;
+                var overflowR = htx + txHalfWidth - outerWidth;
+                var overflowL = htx - txHalfWidth;
+                if(overflowR > 0) htx -= overflowR;
+                if(overflowL < 0) htx += -overflowL;
```
antoinerg: Beautiful! It makes sure we never have an overflow on the left which is perfect for languages that read left to right. 

One day, we should allow reversing that order for languages that read right to left :smile_cat:

antoinerg:
Thank you for the fix @etpinard! It looks good to me :dancer:

PR #3685, opened 2019-03-26
Hovertemplate for box/violin points

etpinard:
Slipping this in 1.46.0.

This PR first fixes an unreported bug for scatter traces with `hoveron: 'points+fills'` and then adds `hovertemplate` support for box/violin points. Please note that `hovertemplate` has no effect on the box/violin mean/q1/q3/kde etc labels. This part will remain unresolved until we can spec the API.

cc @plotly/plotly_js and @nicolaskruchten

### DIFF
--- a/test/jasmine/tests/hover_label_test.js
+++ b/test/jasmine/tests/hover_label_test.js
@@ -1938,6 +1938,38 @@ describe('hover info', function() {
         .catch(failTest)
         .then(done);
     });
+
+    it('should fallback to regular hover content when hoveron does not support hovertemplate', function(done) {
+        var gd = createGraphDiv();
+        var fig = Lib.extendDeep({}, require('@mocks/scatter_fill_self_next.json'));
+
+        fig.data.forEach(function(trace) {
+            trace.hoveron = 'points+fills';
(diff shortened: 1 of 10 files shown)


test/jasmine/tests/hover_label_test.js
```diff
@@ -1938,6 +1938,38 @@ describe('hover info', function() {
...
+
+        fig.layout.hovermode = 'closest';
+        fig.layout.showlegend = false;
+        fig.layout.margin = {t: 0, b: 0, l: 0, r: 0};
+
+        Plotly.plot(gd, fig)
+        .then(function() { _hoverNatural(gd, 180, 200); })
+        .then(function() {
+            assertHoverLabelContent({
+                nums: 'trace 1',
```
etpinard: Before the fix, this used to show `'%{x} | %{y}'`, that is the raw hover template string.

antoinerg:
This PR introduces a first-level attribute `hovertemplate` that applies only to points. This seems like a reasonable default and it shouldn't be a problem in the future as I think our next API will have second-level (or higher) attributes `hovertemplate` for the other elements.

Thanks for the addition @etpinard ! :dancer:

Issue #3653, opened 2019-03-19
Violin kde all 0 in marginal subplot


nicolaskruchten:
Compare the following plots, which differ only in that one has a second violin on the right: https://plot.ly/~nicolaskruchten/290 and https://plot.ly/~nicolaskruchten/292

![image](https://user-images.githubusercontent.com/203523/54633829-562e5d00-4a57-11e9-8bd8-a7ffcdc03ba1.png)

Why is it that when there is a second marginal violin on the right, the top one has KDE=0 everywhere?

etpinard:
Might have been fixed during @antoinerg 's https://github.com/plotly/plotly.js/pull/3626

@antoinerg wanna check?

antoinerg:
Thank you @nicolaskruchten for reporting this issue.

The following figure https://plot.ly/~nicolaskruchten/290 is indeed not rendering properly. 
Here it is in a https://codepen.io/antoinerg/pen/vPzxzL

antoinerg:
I forgot to mention but the above Codepen uses a recent plotly.js build that includes #3626. The issue is still present and will require a new fix!

antoinerg:
Although the violins are rendered differently in the two figures produced by @nicolaskruchten, the `calcData` is the same (ie. same bandwidth and same density/kde). The problem is probably related to how the axis range is computed. To be followed.

antoinerg:
The problem reported occurs because both violins have the same trace name (which is a whitespace [see JSON](https://plot.ly/~nicolaskruchten/290/lifeexp-vs-gdppercap.json)). Because of this, by default, they end up in the same `scalegroup` with a default `scalemode` of `width`. Rescaling their width together makes no sense since they represent two very different quantities (ie. the density of `lifeExp` and `gdpPercap` respectively).

Question: is the default behavior of grouping traces in the same `scalegroup` if they have the same trace name a good default? I lean towards yes.

Unfortunately for @nicolaskruchten, he's not out of the woods just yet: there is also the issue of hover labels being empty for the top violin :disappointed:

nicolaskruchten:
I don't think we should do any kind of automatic grouping based on the trace name, as a gut reaction. Seems far too brittle.

etpinard:
> I don't think we should do any kind of automatic grouping based on the trace name,

Hmm, I'd call removing that automatic grouping

[...]

PR #3687, opened 2019-03-27
update description for `violin.scalegroup`

antoinerg:
This PR is an attempt at resolving #3653 by documenting how the default value for `violin.scalegroup` is set.

Even if the current behavior it's not what we may want in the long-term, I think we should at least document how it is working at the moment.


etpinard:
I like it :dancer:

PR #3688, opened 2019-03-28
Update vertextcolor description to note the range of the values

### LINKED ISSUE #3348: update `vertextcolor` description to note the range of the values
I was just reading through a community member's notebook (https://plot.ly/~empet/15040/plotly-mesh3d-from-a-wavefront-obj-f/#/) and came across this comment:
image

Indeed, looks like the `vertexcolor` description is a little terse, could mention that this color needs to be normalized between `0` and `1`, unlike other color mappings:
https://github.com/plotly/plotly.js/blob/83fd6512178a0881ed9a960ab88a4c2393e88b04/src/traces/mesh3d/attributes.js#L150-L158

archmoj:
Fix #3348.
@plotly/plotly_js

### DIFF
--- a/src/traces/mesh3d/attributes.js
+++ b/src/traces/mesh3d/attributes.js
@@ -163,7 +163,10 @@ module.exports = extendFlat({
         editType: 'calc',
         description: [
             'Sets the color of each vertex',
-            'Overrides *color*.'
+            'Overrides *color*. While Red, green and blue colors',
+            'are in the range of 0 and 255; in the case of having',
+            'vertex color data in RGBA format, the alpha color',
+            'should be normalized to be between 0 and 1.'
         ].join(' ')
     },
     facecolor: {


src/traces/mesh3d/attributes.js
```diff
@@ -163,7 +163,10 @@ module.exports = extendFlat({
         editType: 'calc',
         description: [
             'Sets the color of each vertex',
-            'Overrides *color*.'
+            'Overrides *color*. While Red, green and blue colors',
+            'are in the range of 0 and 255; in the case of having',
+            'vertex color data in RGBA format, the alpha color',
+            'should be normalized to be between 0 and 1.'
```
etpinard: Hmm. That sounds like an odd `mesh3d` behavior that doesn't a correspondence in other traces. Can you confirm?

archmoj: Actually there is also this bug https://github.com/plotly/plotly.js/issues/3741 in `gl-mesh3d`.

archmoj: > Hmm. That sounds like an odd `mesh3d` behavior that doesn't have a correspondence in other traces. Can you confirm?

@etpinard It is actually the case for other 3d traces e.g. `scatter3d`: [codepen](https://codepen.io/MojtabaSamimi/pen/wZojoM?editors=0010).

etpinard: I'm starting to get a little confused by this issue.

First, see https://codepen.io/etpinard/pen/yrMmpJ?editors=0010 - looks like `scatter3d` is behaving inconsistently. Colors set as `rgb(200,100,150,0.2)` don't show an alpha channel in `scatter`. To show that alpha channel, one has to use the `rgba` prefix e.g. `rgba(200,100,150,0.2)`. That's probably something we could add to the `marker.color` attribute description:

https://github.com/plotly/plotly.js/blob/1d0acf035ac054c983c3e6bb1b4a691a0c8cc8be/src/components/colorscale/attributes.js#L108-L114

Next, if I'm understanding correctly https://github.com/plotly/plotly.js/issues/3348 is about `vertexcolor` accepting colors as `[1,0,0,1]` AND `[255,0,0,255`] see https://codepen.io/etpinard/pen/rbyXvm and https://codepen.io/etpinard/pen/GPNwYX for examples which yield the same graph using different `vertexcolor` denominator.

----

Things are a little messy here. So feel free to take this issue wherever you please.

archmoj: @etpinard thanks for the review.
You are right. There is a problem with `Alpha > 1` not being mapped to 0 - 1 range.
That one should actually be labelled as a bug!

archmoj:
This should be addressed as a bug.
Closing.

etpinard:
@archmoj feel free to reopen this PR. :dancer: - the issue I noticed in https://github.com/plotly/plotly.js/pull/3688#discussion_r274193906 about rgb vs rgba colors can be taken care of in a separate PR.

Issue #3686, opened 2019-03-26
stackgroup + groupby transforms bug


nicolaskruchten:
The following Pen shows the issue: https://codepen.io/nicolaskruchten/pen/GeaZBz (the 'b' trace is missing!)

PR #3692, opened 2019-03-28
Fix stacked scatter for groupby traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3686

Updated codepen: https://codepen.io/etpinard/pen/MxNoXp

cc @plotly/plotly_js


antoinerg:
Thank you @etpinard for the nice little fix :+1: 

:dancer:

Issue #3147, opened 2018-10-24
Unexpect fill connection rendering with extreme values


jonmmease:
Plotlyjs translation of https://github.com/plotly/plotly.py/issues/1008

When creating a rectangle using 4 `scatter` points (3 line segments) and `fill: "toself"` the connecting edge of the rectangle behaves oddly when the rectangle is much much larger than the viewport.

CodePen:  https://codepen.io/jonmmease/pen/WaapQa

```JavaScript
var yval = 1e4
fig = {
  "data": [
    {
      "fill": "toself",
      "type": "scatter",
      "x": [
        1,
        10,
        10,
        1
      ],
      "y": [
        yval,
        yval,
        -yval,
        -yval
      ]
    }
  ],
  "layout": {
    "yaxis": {
      "range": [
        -0.5,
        0.5
      ]
    }
  }
}

Plotly.newPlot('div1', fig);
```
![newplot-1](https://user-images.githubusercontent.com/15064365/47432705-d2549b00-d76c-11e8-9e36-5fca5a599303.png)

But it looks as expected with `yval` is 1e2:
![newplot-2](https://user-images.githubusercontent.com/15064365/47432747-eb5d4c00-d76c-11e8-8cc5-e1fc8a01986f.png)

It also work as expected with `scattergl` traces.

PR #3696, opened 2019-03-29
Always include the very first and last points for fills

etpinard:
... during the line-decimation algo - fixes https://github.com/plotly/plotly.js/issues/3147

cc @plotly/plotly_js


test/image/mocks/ultra_zoom_fill.json
```diff
@@ -0,0 +1,21 @@
+{
+  "data": [
```
etpinard: which looks like:

image


on `master`.

antoinerg:
Thanks for the fix @etpinard !

:dancer:

src/traces/scatter/line_points.js
```diff
@@ -387,7 +391,9 @@ module.exports = function linePoints(d, opts) {
 
             clusterRefDist = ptDist(clusterHighPt, clusterStartPt);
 
-            if(clusterRefDist < getTolerance(clusterHighPt, nextPt) * minTolerance) continue;
+            // #3147 - always include the very first and last points for fills
+            if(!(fill && (pti === 0 || len - pti === 1)) &&
```
alexcjohnson: I've never seen `pti === len - 1` written quite that way 🤔

etpinard: yeah, that _was_ weird, fixed in https://github.com/plotly/plotly.js/pull/3696/commits/9e8c3e9783860cc741bb3fb9bd0e9a1aa754ab44

archmoj:
:dancer:

PR #3707, opened 2019-04-01
Fix `npm run test-image -- *` for mock lists w/o gl2d_* mocks

etpinard:
A follow-up of https://github.com/plotly/plotly.js/pull/3634/

Currently on `master`:

```
npm run test-image -- 10
```

gives us:

image

as we're trying to sort `['gl2d_pointcloud-basic', 'gl2d_heatmapgl']` to the front even though they're not in the list of mocks we're trying to test.

cc @plotly/plotly_js


archmoj:
Thanks for the fix.
:dancer:

PR #3708, opened 2019-04-01
Add waterfall trace to finance bundle

etpinard:
Quick follow-up of https://github.com/plotly/plotly.js/pull/3531

`waterfall` is a finance trace and we should make it part of the finance bundle and npm package.

cc @plotly/plotly_js

### DIFF
--- a/lib/index-finance.js
+++ b/lib/index-finance.js
@@ -15,7 +15,8 @@ Plotly.register([
     require('./histogram'),
     require('./pie'),
     require('./ohlc'),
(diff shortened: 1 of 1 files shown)


archmoj:
@etpinard Thanks for the fix.
Is it on 1.47 or 1.46?

etpinard:
> Is it on 1.47 or 1.46?

In 1.46.0 of course. I clicked on the wrong milestone. Thanks for spotting that!

PR #3711, opened 2019-04-02
Format numbers in waterfall hover

### LINKED ISSUE #3710: Waterfall initial field rounding
https://rreusser.github.io/plotly-mock-viewer/#waterfall_line

I'd expect the tooltip to display 1 in this case, not 1.000000000000002

Screen Shot 2019-04-01 at 3 42 12 PM

archmoj:
Fix #3710 by rounding 10 digits then parsing as float. 
[Codepen](https://codepen.io/MojtabaSamimi/pen/zXxbjG).
@plotly/plotly_js

### DIFF
--- a/src/traces/waterfall/hover.js
+++ b/src/traces/waterfall/hover.js
@@ -8,7 +8,8 @@
 
 'use strict';
 
-var Color = require('../../components/color');
+var hoverLabelText = require('../../plots/cartesian/axes').hoverLabelText;
+var opacity = require('../../components/color').opacity;
 var hoverOnBars = require('../bar/hover').hoverOnBars;
 
 var DIRSYMBOL = {
@@ -16,22 +17,25 @@ var DIRSYMBOL = {
     decreasing: '▼'
 };
 
-function formatNumber(a) {
-    return parseFloat(a.toPrecision(10));
-}
-
 module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     var point = hoverOnBars(pointData, xval, yval, hovermode);
     if(!point) return;
 
     var cd = point.cd;
     var trace = cd[0].trace;
+    var isHorizontal = (trace.orientation === 'h');
+
+    var vAxis = isHorizontal ? pointData.xa : pointData.ya;
+
+    function formatNumber(a) {
+        return hoverLabelText(vAxis, a);
+    }
 
     // the closest data point
     var index = point.index;
     var di = cd[index];
 
-    var sizeLetter = (trace.orientation === 'h') ? 'x' : 'y';
+    var sizeLetter = isHorizontal ? 'x' : 'y';
 
     var size = (di.isSum) ? di.b + di.s : di.rawS;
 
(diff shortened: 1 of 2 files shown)


src/traces/waterfall/hover.js
```diff
@@ -34,16 +38,16 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     if(!di.isSum) {
         // format delta numbers:
         if(size > 0) {
-            point.extraText = size + ' ' + DIRSYMBOL.increasing;
+            point.extraText = formatNumber(size) + ' ' + DIRSYMBOL.increasing;
```
alexcjohnson: Use `Axes.hoverLabelText`, it's made specifically for this purpose:
https://github.com/plotly/plotly.js/blob/c6573487f33ffe02f38f443690e5fdb824fdb161/src/traces/box/hover.js#L170

archmoj: Thanks for the review @alexcjohnson.
Revised in https://github.com/plotly/plotly.js/pull/3711/commits/07b6c9c296b2037ebda7d12619e31354fc263296.

src/traces/waterfall/hover.js
```diff
@@ -22,28 +23,35 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
...
+            point.extraText = '(' + (formatNumber(-size)) + ') ' + DIRSYMBOL.decreasing;
         } else {
             return;
         }
         // display initial value
-        point.extraText += '<br>Initial: ' + (di.b + di.s - size);
+        point.extraText += '<br>Initial: ' + formatNumber(di.b + di.s - size);
     } else {
-        point[sizeLetter + 'LabelVal'] = size;
+        point[sizeLetter + 'LabelVal'] = formatNumber(size);
```
alexcjohnson: not quite sure where this would break down, but in principle `*LabelVal` should be a number, and `*Label` should be the string you want to use to display that number.

alexcjohnson: Perhaps non-default separators, or tickprefix/suffix, would break this as is

archmoj: In `bar` traces we use a similar pattern: https://github.com/plotly/plotly.js/blob/07b6c9c296b2037ebda7d12619e31354fc263296/src/traces/bar/hover.js#L138.
Do you think there would be a problem there?

alexcjohnson: Bar uses `*LabelVal` correctly, it's a number. I'm guessing that if there's no `*Label` then later on `*LabelVal` gets converted to a string using `hoverLabelText` but I don't know. If so, perhaps you can just leave this one as a number too.

Whew, the hover system is certainly ripe for refactoring... or at least documenting.

etpinard: > Whew, the hover system is certainly ripe for refactoring... or at least documenting.

:pray:

archmoj: > Bar uses `*LabelVal` correctly, it's a number. I'm guessing that if there's no `*Label` then later on `*LabelVal` gets converted to a string using `hoverLabelText` but I don't know. If so, perhaps you can just leave this one as a number too.

OK. Let's keep this as it is in this PR.

etpinard:
Nice fix. Good enough to be released in 1.46.1 :dancer:

Issue #3582, opened 2019-02-26
Sankey: mouse interactions with groups


antoinerg:
In Sankey traces, it would be nice to be able to group nodes with a lasso and conversely to be able to expand grouped nodes.

PR #3712, opened 2019-04-02
sankey: implement node grouping via mouse selection

antoinerg:
Closes https://github.com/plotly/plotly.js/issues/3582

Peek 2019-03-14 17-08

- [x] Add a home button to ungroup all groups (not seen in the above GIF)
- [x] Add jasmine tests


src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
     }
 };
+
+exports.updateFx = function(gd) {
+    for(var i = 0; i < gd._fullData.length; i++) {
+        subplotUpdateFx(gd, i);
+    }
+};
+
+var dragOptions = [];
```
etpinard: Why do we need to track the `dragOptions` entries in the file scope?

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+            yaxis: yaxis,
+            fillRangeItems: Lib.noop
+        },
+        subplot: i,
+        // create mock x/y axes for hover routine
+        xaxes: [xaxis],
+        yaxes: [yaxis],
+        doneFnCompleted: function(selection) {
+            var newGroups;
+            var oldGroups = gd._fullData[i].node.groups.slice();
```
etpinard: Nicely done playing it safe here as the wanted `gd._fullData[i]` ref could be different in this scope than in the `subplotUpdateFx` scope.

Usually in other drag callbacks of the likes, we would write this as:

```js
// ...
var traceNow = gd._fullData[i];
var oldGroups = traceNow.node.groups.slice();
// ...
```

to make it clear that we using the "now" `fullData` item.

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+            var newGroup = [];
+
+            function findNode(pt) {
+                return gd._fullData[i]._sankey.graph.nodes.find(function(n) {
```
etpinard: What's `find` here? Is `nodes` a plain JS array? Is so, we can't use this:

image

src/traces/sankey/render.js
```diff
@@ -836,6 +836,23 @@ module.exports = function(gd, svg, calcData, layout, callbacks) {
         .style('pointer-events', 'auto')
         .attr('transform', sankeyTransform);
 
+    sankey.each(function(d, i) {
+        gd._fullData[i]._sankey = d;
+
+        // Draw dragbox
+        Lib.ensureSingle(gd._fullLayout._draggers, 'rect', 'bg-' + i, function(el) {
```
etpinard: Having something like `bgsankey-' + trace.uid` would be safer.

Note that you'll need to remove this dragger layer in 

https://github.com/plotly/plotly.js/blob/c6573487f33ffe02f38f443690e5fdb824fdb161/src/traces/sankey/base_plot.js#L29-L36

src/traces/sankey/select.js
```diff
@@ -0,0 +1,36 @@
...
+    var fullData = cd[0].trace;
+
+    var nodes = fullData._sankey.graph.nodes;
+
+    for(var i = 0; i < nodes.length; i++) {
+        var node = nodes[i];
+        if(node.partOfGroup) continue; // Those are invisible
+
+        // TODO: decide on selection criteria, using centroid for now
+        var pos = [(node.x0 + node.x1) / 2, (node.y0 + node.y1) / 2];
```
etpinard: Voting :+1: on centroid.

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+exports.updateFx = function(gd) {
+    for(var i = 0; i < gd._fullData.length; i++) {
+        subplotUpdateFx(gd, i);
+    }
+};
+
+var dragOptions = [];
+function subplotUpdateFx(gd, index) {
+    var i = index;
+    var fullData = gd._fullData[i];
```
etpinard: You should call this `trace` (if no `gd.data[i]` are present in this scope) or `fullData` (otherwise).

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+
+exports.updateFx = function(gd) {
+    for(var i = 0; i < gd._fullData.length; i++) {
+        subplotUpdateFx(gd, i);
+    }
+};
+
+var dragOptions = [];
+function subplotUpdateFx(gd, index) {
+    var i = index;
```
etpinard: Keep this as `index`. We try to keep one-letter variables (like `i`) for for-loop counters.

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+
+    var dragMode = fullLayout.dragmode;
+    var cursor = fullLayout.dragmode === 'pan' ? 'move' : 'crosshair';
+    var bgRect = fullData._bgRect;
+
+    setCursor(bgRect, cursor);
+
+    var xaxis = {
+        _id: 'x',
+        c2p: function(v) { return v; },
```
etpinard: You can use `Lib.identity`

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+                    }
+                    // Flag group for removal from existing list of groups
+                    oldGroups[node.pointNumber - fullData.node._count] = false;
+                } else {
+                    newGroup.push(node.pointNumber);
+                }
+            }
+
+            newGroups = oldGroups
+                .filter(function(g) { return g; })
```
etpinard: `.filter(Boolean)` would be cleaner :paintbrush:

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+                .concat([newGroup]);
+
+            Registry.call('_guiRestyle', gd, {
+                'node.groups': [ newGroups ]
+            }, i).catch(Lib.noop); // TODO will this ever fail?
```
etpinard: Don't catch. If it fails, we'll like to know about it.

src/traces/sankey/base_plot.js
```diff
@@ -34,3 +40,94 @@ exports.clean = function(newFullData, newFullLayout, oldFullData, oldFullLayout)
...
+        }
+    };
+
+    dragOptions[i].prepFn = function(e, startX, startY) {
+        prepSelect(e, startX, startY, dragOptions[i], dragMode);
```
etpinard: ... and I guess `prepSelect` does the right thing when `dragmode` isn't `select` nor `lasso`?

src/plots/cartesian/select.js
```diff
@@ -284,6 +286,8 @@ function prepSelect(e, startX, startY, dragOptions, mode) {
                 dragOptions.mergedPolygons.length = 0;
                 [].push.apply(dragOptions.mergedPolygons, mergedPolygons);
             }
+
+            doneFnCompleted(selection);
```
etpinard: I like this solution a lot. Very clean :1st_place_medal:

src/traces/sankey/base_plot.js
etpinard: Ahh. This is tricky actually for `sankey`, dragmode values `zoom` and `pan` don't really make sense. I'm not sure if we can do anything about it in v1. Would it be ok to have `zoom` and `pan` correspond to the current "drag-on-nodes" behaviour?

etpinard:
Looking good @antoinerg - some minor comments, but you're for sure on the right track :racing_car:

src/traces/sankey/base_plot.js
antoinerg: In the current implementation we don't. 

I :hocho:'ed it in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: I agree that's a good convention: done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: Thank you @etpinard for pointing that out. This `find` has no place in plotly.js. It is replaced by a for loop in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/render.js
antoinerg: Done in 20d8c92d8454cc804e71b84a90469df807d14898 and :lock: down with a test.

src/traces/sankey/select.js
antoinerg: Ok :+1:  I :hocho: the TODO in 20d8c92d8454cc804e71b84a90469df807d14898

Maybe we should add a visual cue to let the user know a given node is now part of a selection :thinking: This could also be added later on.

src/traces/sankey/base_plot.js
antoinerg: Done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: I agree that this is a good convention! Done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: Sweet :smile_cat: 

Done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: I didn't know this! Done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: I don't know what I was thinking. Thank you @etpinard 

Done in 20d8c92d8454cc804e71b84a90469df807d14898

src/traces/sankey/base_plot.js
antoinerg: > Would it be ok to have zoom and pan correspond to the current "drag-on-nodes" behavior?

Yes, I think it would be OK. I will return early in `subplotUpdateFx` when `dragMode` is either `zoom` or `pan`.

antoinerg: Done in aa17f1b6b32091fc171e97dfad4e3e18ee331421

src/components/modebar/buttons.js
```diff
@@ -513,6 +513,17 @@ function toggleHover(gd) {
     Registry.call('_guiRelayout', gd, 'hovermode', newHover);
 }
 
+modeBarButtons.resetSankeyGroup = {
+    name: 'resetSankeyGroup',
+    title: function(gd) { return _(gd, 'Ungroup all nodes'); },
+    icon: Icons.home,
+    click: function(gd) {
+        Registry.call('restyle', gd, {
+            'node.groups': [[]],
```
etpinard: As we discussed in person, it might be better to `restyle` back to the "initial" view rather than removing all groups.

To do so:

- stash the 'initial' `node.(x|y)` and `groups` in `fullData[i]` (see mapbox [example](https://github.com/plotly/plotly.js/blob/07822743b03e9bc1a3decd429c02b0452af49732/src/plots/mapbox/index.js#L83-L90))
- call `restyle` with the values in that stash each in the modebar button handler (see [example](https://github.com/plotly/plotly.js/blob/07822743b03e9bc1a3decd429c02b0452af49732/src/components/modebar/buttons.js#L596-L614))

etpinard: @antoinerg let me know if you need help implementing this!

antoinerg: > call restyle with the values in that stash each in the modebar button handler (see example)

I don't think I can reuse `resetView` because it does a `relayout` whereas we need a `restyle`.

etpinard: Right, no need to use `resetView()` found in `modebar/buttons.js`, but you should try to replicate the same pattern.

antoinerg: Done in commit 506da4c! Let me know if it is satisfactory. 

I will now write Jasmine tests for the reset button. Is there anything else you would like me to add?

etpinard: > I will now write Jasmine tests for the reset button. Is there anything else you would like me to add?

Nop, once that jasmine test is in, this one will be good to go!

antoinerg: Done in f8a4073

etpinard:
Excellent!

You might not feel that way @antoinerg as this PR didn't add that many lines to `src/`, but I think this is your best PR to date. Reusing stuff, following our existing patterns and writing good interaction tests are all difficult things that not many past plotly.js devs have even attempted. 

Well-earned :dancer:

Issue #3714, opened 2019-04-02
Presence of textfont without text mangles bar traces


cpsievert:
If you comment out `textfont` in this codepen it gives you the correct result

https://codepen.io/cpsievert/pen/EJjrdd

PR #3701, opened 2019-03-31
Draw blank bars as empty paths + allow blank bars to draw outsidetext

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3115 

cc @plotly/plotly_js

PR #3715, opened 2019-04-02
Don't try to access textfont when bar has no text

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3714 - a fixup of my https://github.com/plotly/plotly.js/pull/3701


etpinard:
cc @plotly/plotly_js

archmoj:
Thanks for the PR.
:dancer: when you get all the green tests.

Issue #3306, opened 2018-12-04
Surface graph has issue with small time ranges


ShaneHudson:
Hello,

I've had an issue for quite a while and thought it would be good to raise here. It's an issue with graphs that are over a small amount of time. I've attempted to make a demonstration but the screenshot below will show my issue more clearly.

Take a look at the following:
A) https://jsfiddle.net/Lv4qpoyu/3/
B) https://jsfiddle.net/5au0xepd/1/

In A) you can see some random data over 50 minutes. In B) you can see random data over 50 seconds. A) works perfectly well, nice little graph. B) looks like an utter mess and is near impossible to use.

So that brings us to C), the screenshots (the first taken a few months ago, the second today) below that shows my real dataset. In a 2D heatmap it works fine but in a 3D surface graph it shows the data correctly but with the axis completely messed up. Sometimes it looks worse than others, but all the time there is a continuation of the date (y) axis. Sometimes it requires you to zoom out to see, but it is always there in the distance. This makes it unusable to me if I want to put it in front of users.

![44929865-5e8da780-ad54-11e8-8795-0c2353f09984](https://user-images.githubusercontent.com/487005/49437514-983ad800-f7b3-11e8-9e94-ef22a9d7cf8e.png)

![screenshot 2018-12-04 11 03 52](https://user-images.githubusercontent.com/487005/49437762-51011700-f7b4-11e8-9633-b8ccc7f92aa1.png)

When I've looked online for help I found [someone else](https://stackoverflow.com/questions/52636143/plot-dates-on-3d-axis-using-plotly-framework) having the exact same issue but they were using the python version of the library with a scatter3d.

This seems to be a bug opposed to an issue in my code. I have autorange set to false, the range of data is correct and is using moment dates. Does anyone have any idea what is causing this issue?

Cheers
Shane

ShaneHudson:
It looks like this may be related to https://github.com/plotly/plotly.js/issues/3155 but would appreciate any help on the matter.

archmoj:
@ShaneHudson Thank you very much for reporting this issue.

ShaneHudson:
Also another thing I've noticed, potentially related to this, is if there is no data on the graph but ranges are set on the axis then for the heatmap it shows correct ranges but the surface plot shows from 2000 to 2001.

archmoj:
@ShaneHudson Could you please create a demo related to the second issue?

ShaneHudson:
[...]

PR #3721, opened 2019-04-04
Adjust local time offset help improve display short time periods (hours) in gl3d plots

archmoj:
As partly described in https://github.com/plotly/plotly.js/issues/3306 when dealing with short times on gl3d axes, there are two problems involved:
One is that we need to deal with float numbers right close to precision limits (noting that floats are generally good when the data is uniformly distributed around zero).  This could be improved when https://github.com/plotly/plotly.js/issues/2646 is addressed.
The second problem is that the date ticks presented in the browser shifted depending on user time zone.
This PR is an attempt to solve the second issue. 
- [ ] need to figure out a better way to adjust the time in CI tests and in generating the gl3d baselines.  

@alexcjohnson 
@plotly/plotly_js


archmoj:
Reworked in #4903.

Issue #3322, opened 2018-12-11
Sankey: Hover info extensions


antoinerg:
Add support for multiple tooltips ("compare" hover mode) and concentration display.
Optionally display one hover box for each component in a flow, and optionally display the concentration of a component within the flow in its hover box.

PR #3730, opened 2019-04-04
sankey: compare links in a flow on hover

antoinerg:
Closes https://github.com/plotly/plotly.js/issues/3322

Screenshot_2019-04-04_17-30-43

- [x] Add jasmine tests

### DIFF
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -156,42 +156,49 @@ module.exports = function plot(gd, calcData) {
         if(gd._fullLayout.hovermode === false) return;
         var obj = d.link.trace.link;
         if(obj.hoverinfo === 'none' || obj.hoverinfo === 'skip') return;
-        var rootBBox = gd._fullLayout._paperdiv.node().getBoundingClientRect();
-        var hoverCenterX;
-        var hoverCenterY;
-        if(d.link.circular) {
-            hoverCenterX = (d.link.circularPathData.leftInnerExtent + d.link.circularPathData.rightInnerExtent) / 2 + d.parent.translateX;
-            hoverCenterY = d.link.circularPathData.verticalFullExtent + d.parent.translateY;
-        } else {
-            var boundingBox = element.getBoundingClientRect();
-            hoverCenterX = boundingBox.left + boundingBox.width / 2 - rootBBox.left;
-            hoverCenterY = boundingBox.top + boundingBox.height / 2 - rootBBox.top;
+
+        var hoverItems = [];
+
+        // For each related links, create a hoverItem
+        for(var i = 0; i < d.flow.links.length; i++) {
+            var link = d.flow.links[i];
+            obj = link;
+            var hoverCenterX;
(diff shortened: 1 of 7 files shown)


antoinerg:
Right now, only the top link within a flow has its hover label exactly above it. All the ones below will be shifted. Should I change this behavior such that the hovered link always has its hover label above it? I think this would be ideal but I'm not sure how easy/hard it would be to implement. :thinking:

src/traces/sankey/plot.js
```diff
@@ -156,51 +156,63 @@ module.exports = function plot(gd, calcData) {
...
-        }
 
-        var hovertemplateLabels = {valueLabel: d3.format(d.valueFormat)(d.link.value) + d.valueSuffix};
-        d.link.fullData = d.link.trace;
+        var hoverItems = [];
+
+        // For each related links, create a hoverItem
+        for(var i = 0; i < d.flow.links.length; i++) {
+            var link = d.flow.links[i];
+            if(gd._fullLayout.hovermode === 'closest' && d.link.pointNumber !== link.pointNumber) continue;
```
etpinard: So, `hovermode: 'x'` or `'y'` gives enables the "compare" mode? That's a little unfortunate because `sankey` don't have x nor y axes, but I guess it's not worth adding a `hovermode: 'compare'` just for sankey.

etpinard: ... so I'm ok with this :ok_hand:

etpinard:
> l but I'm not sure how easy/hard it would be to implement

give it a shot! If you think it's the ideal behavior, it's definitely worth spending at least one more day on this feature.

antoinerg:
> give it a shot! If you think it's the ideal behavior, it's definitely worth spending at least one more day on this feature.

I agree that the current behavior is not good enough. It should be possible to display any hover label via the mouse. For this to work, we need to center it or at least making sure it's not overflowing. Right now it just overflows:
Screenshot_2019-04-04_18-32-47

antoinerg:
With commit f00b542 we can reach all hoverlabels and fix https://github.com/plotly/plotly.js/pull/3730#issuecomment-480088462:
Screenshot_2019-04-05_15-11-28

etpinard:
> With commit [f00b542](https://github.com/plotly/plotly.js/commit/f00b542c76fc4f7c7fa21fd324eb2bf9c30e2b84) we can reach all hoverlabels and fix [#3730 (comment)](https://github.com/plotly/plotly.js/pull/3730#issuecomment-480088462):

:tada: :champagne: :1st_place_medal:

etpinard:
Brilliant work. :dancer:

etpinard:
@antoinerg one sankey `@noCI` test is failing on `master` is my machine:

image

Is it passing on your setup?

archmoj:
@etpinard That one also fails on my machine.

antoinerg:
It fails sometimes but it does pass more often than not. I guess this one should be a `@flaky`.

Issue #3727, opened 2019-04-04
legend.marker.sizemode


nicolaskruchten:
We should add `legend.marker.sizemode = data (default) | constant`

related: #3602

nicolaskruchten:
related: #2745 and #2080

etpinard:
Full list of related issues:

- https://github.com/plotly/plotly.js/issues/2745
- https://github.com/plotly/plotly.js/issues/2967
- https://github.com/plotly/plotly.js/issues/2744
- https://github.com/plotly/plotly.js/issues/2080
- https://github.com/plotly/plotly.js/issues/1701
- https://github.com/plotly/plotly.js/issues/1560
- https://github.com/plotly/plotly.js/issues/98

PR #3732, opened 2019-04-04
Add legend `itemsizing`

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3727 - by adding `legend.itemsizing` with values `'constant'` and `'trace'` (the default and current behavior).

I chose `itemsizing` over `sizemode` as we already have a `sizemode` attribute in the `scatter*` traces which does something sufficiently different in my mind to not replicate the name over to `layout.legend`. Let me know if you can think of a better name.


src/components/legend/style.js
```diff
@@ -17,16 +17,38 @@ var Color = require('../color');
 
 var subTypes = require('../../traces/scatter/subtypes');
 var stylePie = require('../../traces/pie/style_one');
+var pieCastOption = require('../../traces/pie/helpers').castOption;
+
+var CST_MARKER_SIZE = 12;
+var CST_LINE_WIDTH = 5;
+var CST_MARKER_LINE_WIDTH = 2;
+var MAX_LINE_WIDTH = 10;
+var MAX_MARKER_LINE_WIDTH = 5;
```
etpinard: Please see `legend-constant-itemsizing.png` below and let me know if we can set better constant values.

src/components/legend/style.js
```diff
@@ -17,16 +17,38 @@ var Color = require('../color');
...
 
 module.exports = function style(s, gd) {
+    var fullLayout = gd._fullLayout;
+    var legend = fullLayout.legend;
+    var constantItemSizing = legend.itemsizing === 'constant';
```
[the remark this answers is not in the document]
etpinard: Why?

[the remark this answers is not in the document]
archmoj: Simply thought it may be more readable that we deal with a boolean.

test/image/mocks/legend-constant-itemsizing.json
```diff
@@ -0,0 +1,169 @@
...
+          "width": [ 10, 15, 14 ],
+          "color": "#444"
+        }
+      }
+    },
+    {
+      "type": "waterfall",
+      "name": "waterfall",
+      "base": 5,
+      "y": [ 1, -1, 1 ],
```
archmoj: Could you please add these lines into it?
<pre>
"y": [ 1, -0.5, null],
"measure": [ "", "", "total"],
</pre>

test/image/mocks/legend-constant-itemsizing.json
```diff
@@ -0,0 +1,169 @@
...
+        "marker": {
+          "line": {
+            "width": 20
+          }
+        }
+      },
+      "decreasing": {
+        "marker": {
+          "line": {
+            "width": 40
```
archmoj: Please set `width` to 10 here.
Thanks.

src/components/legend/style.js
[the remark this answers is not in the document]
etpinard: I'm not a fan of those unnecessary `(` `)`.

test/image/mocks/legend-constant-itemsizing.json
archmoj: A little clarification on why we may need to tweak this one:
I thought this mock is a good example showcasing different legends. And for waterfall the default is to have 3 colors. If  we accept my previous comment to use `measure` list to display the total, then this `width` of 40 is too much and hides the decreasing bar.

archmoj:
> @etpinard thanks for the PR.
> Concerning those gl3d baseline changes, I am wondering if we may apply half of the sizes?

@etpinard In fact this PR solved the issue with those gl3d baselines I mentioned earlier. So there is nothing to worry there.

src/components/legend/style.js
etpinard: cc @nicolaskruchten @alexcjohnson

nicolaskruchten: do these differ from the current defaults?

etpinard: Which _default_ are you referring to here?

Legend items at the moment don't have static defaults, their style depend on the style of their corresponding traces.

nicolaskruchten: so I was assuming that constant sizing meant literally "whatever the size would have been if `marker.size` wasn't variable" so I'm confused about what these constants here do

etpinard: You mean what's the default `marker.size` value for a scatter trace?

https://github.com/plotly/plotly.js/blob/5d69e77fde779dc57861e108de720b08a14a3fc4/src/traces/scatter/attributes.js#L365

and it looks like:

image

I'd argue that's too small when `marker.line.width>0` so I made it twice as big.

etpinard: Ping. Is everyone ok with values I picked for `legend.itemsizing: 'constant'`?

antoinerg: > Is everyone ok with values I picked for `legend.itemsizing: 'constant'`?

@etpinard sorry for the late reply. I like how the mock currently looks. However, I have 2 questions:
1. Did you consider sizing it proportionally to the `legend.font.size`? Right now, let's say if `legend.font.size = 30`, the items may look a bit small:
Screenshot_2019-04-08_18-44-32

2. Do we plan to support an attribute to control the size of the items/symbols? If so, would it make more sense to have an `item.(property1|property2|...)` container?

etpinard: > 1. Did you consider sizing it proportionally to the `legend.font.size`?

no



> 2\. Do we plan to support an attribute to control the size of the items/symbols? If so, would it make more sense to have an `item.(property1|property2|...)` container?

https://github.com/plotly/plotly.js/issues/3735

antoinerg: Well do you think it is a good idea? The ratio of item size vs font size is good in the current mock. Maybe we should :lock: it down. That being said, I'm open to leaving it as it is since we're planning to implement #3735.

archmoj:
OK. My comments above were non-blocking.
:dancer: 
Another great PR @etpinard that would make many people happy.

test/image/mocks/legend-constant-itemsizing.json
etpinard: done in https://github.com/plotly/plotly.js/pull/3732/commits/c6301a1b921a6b324acaea57b5853e5a550b5c7a

test/image/mocks/legend-constant-itemsizing.json
etpinard: done in https://github.com/plotly/plotly.js/pull/3732/commits/c6301a1b921a6b324acaea57b5853e5a550b5c7a

archmoj:
Awesome.
:dancer:

bulam:
`itemsizing = 'constant'` doesn't seem to work in a 3D scatter plot. The legend item circles are still smaller

Issue #3728, opened 2019-04-04
hoverlabel.namelength = 0 should hide the extra hoverlabel thing


nicolaskruchten:
Please and thank you! (if can be slid into 1.47.... :)

PR #3734, opened 2019-04-05
Fix namelength=0 case + implement namelength to loneHover traces

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3728

cc @plotly/plotly_js


test/jasmine/tests/hover_label_test.js
```diff
@@ -805,6 +805,40 @@ describe('hover info', function() {
                     name: 'trace one'
                 });
             })
+            .then(function() {
+                var nl = [[0, 0, 0], [0, 0, 0]];
+                return Plotly.restyle(gd, 'hoverlabel.namelength', [nl, nl]);
+            })
+            .then(function() {
+                // hoverlabel.namelength has no effect when <extra></extra> is set
```
etpinard: N.B. here `namelength` is set to 0, but `hovertemplate` is set with content inside `<extra>`. Here `namelength` is ignore and the extra (name) label still shows up.

Please let me know if you think `namelength` should (essentially) override `<extra>`.

archmoj:
Looks good me. 
@antoinerg @nicolaskruchten any opinion on @etpinard question: 
<pre>Do you think namelength should (essentially) override < extra > ?</pre>

nicolaskruchten:
I think namelength should do nothing if `extra` is set in hovertemplate

alexcjohnson:
Hmm, I was going to argue the opposite - if you template some field into `<extra>`, it could still be useful to limit it with `namelength`. What if we just default to `namelength: -1` when `hovertemplate` is used, but still honor it if provided?

antoinerg:
> Hmm, I was going to argue the opposite - if you template some field into `<extra>`, it could still be useful to limit it with `namelength`. What if we just default to `namelength: -1` when `hovertemplate` is used, but still honor it if provided?

I think I prefer your suggestion @alexcjohnson. I think it's both useful and more consistent with our current behavior: I believe all hover labels should be stylable in the same way whether they use `hovertemplate` or not.

@etpinard I realize that the two functions `Fx.(lone|multi)Hover` do not call `cleanPoint` which is where your fix takes place. Therefore, the current PR will not work with `sunburst`, `sankey`, `parcats`, webGL and possibly others. This begs the question: is there any reason we're not calling `cleanPoint` in functions `Fx.(lone|multi)Hover`?

etpinard:
> This begs the question: is there any reason we're not calling `cleanPoint` in functions `Fx.(lone|multi)Hover`?

Ha, good eye `namelength` never got implemented for those traces.

nicolaskruchten:
Well, the `namelength` vs `extra` nomenclature is odd. Also labelling that widget in the RCE will be interesting if they interact.

I don't have a strong preference.

antoinerg:
> Well, the namelength vs extra nomenclature is odd. 

I regret using `<extra>` as a tag. `<name>` would have been more appropriate since it literally sets what by default is the trace name in hoverlabels. Should we start supporting both tag names?

As for `namelength`, it should do what the doc says:
> Sets the default length (in number of characters) of the trace name in the hover labels for all traces.

etpinard:
This PR is ready again for review.

------------

> What if we just default to `namelength: -1` when `hovertemplate` is used, but still honor it if provided?

Done in https://github.com/plotly/plotly.js/pull/3734/commits/e9f84035155aa9e3a3c0bedf94b0240e8ce3da13

> Therefore, the current PR will not work with `sunburst`, `sankey`, `parcats`, webGL and possibly others.

`namelength` should now work for all our trace type that support `hoverlabel.*` thanks to https://github.com/plotly/plotly.js/pull/3734/commits/918d22e731c701c888d5e88be1a495a3390f21c2 and https://github.com/plotly/plotly.js/pull/3734/commits/7f1645d187cf961b7409ed5e283de76ef2980c5f

src/plots/gl3d/scene.js
```diff
@@ -70,6 +71,7 @@ function render(scene) {
     if(lastPicked !== null) {
         var pdata = project(scene.glplot.cameraParams, selection.dataCoordinate);
         trace = lastPicked.data;
+        var traceNow = gd._fullData[trace.index];
```
archmoj: @etpinard Could you please describe why `traceNow` is needed here?
Thanks in advance.

etpinard: Because `hoverlabel.*` edits don't (and shouldn't have to) replot the scene. As `Plots.supplyDefaults` creates new `gd._fullData` and `gd._fullLayout` refs on every call, we need to use the right "now" trace object when hovering.

antoinerg:
Thanks for all the tests! It looks good to me and I think it is ready to go. I will let @archmoj give you a dancer if he is satisfied with your answer.

Absolutely non-blocking but could you set `hoverlabel.namelength` to `-1` in the mock `sunburst_values` to make sure the hover labels stay the same.

archmoj:
@etpinard thanks for the PR.
:dancer:
Also - To help Q&A you may modify the PR description & briefly describe what it does.

PR #3736, opened 2019-04-05
apply new eslint rule in surface convert file

archmoj:
Lint fix after merging PR #3469.
cc: @etpinard


etpinard:
:dancer: thanks!

Issue #3743, opened 2019-04-08
toImageButtonOptions arg to download current size


nicolaskruchten:
We need to add a JSON-serializable flag to `config.toImageButtonOptions`  to resolve #1576

Issue #1576, opened 2017-04-10
feature request: when downloading plot as png using the modebar button, use the current size of the plot as the size of the png


finger563:
Currently it does not use the size of the plot as it has been rendered into the browser. The data ranges and such are all the same, but since the size is different the legend may overlap parts that it did not when viewed in the browser and the aspect ratio may be different.

etpinard:
Unfortunately, that's more like a backward-incompatible change than a feature request. If we do choose to do this, we'll have to wait for [`v2`](https://github.com/plotly/plotly.js/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20v2).

In the meantime, here's a workaround: http://codepen.io/etpinard/pen/NpvYOo

finger563:
Thanks for the feedback and adding my own options to that code snippet worked well :)

jayanthcl:
Hi,
Thanks for suggesting the workaround

> Unfortunately, that's more like a backward-incompatible change than a feature request. If we do choose to do this, we'll have to wait for [`v2`](https://github.com/plotly/plotly.js/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20v2).
> 
> In the meantime, here's a workaround: http://codepen.io/etpinard/pen/NpvYOo

I am using Angular wrapper for Plotly (https://github.com/plotly/angular-plotly.js/)
Unfortunately, I cannot figure out the proper syntax for this workaround.
Is there a way to use Plotly module from my Angular component?

Will appreciate a lot your help!

ijustmadethisforplotly:
Is there any progress on this feature?- It kinda seems important to be able to save your plots as an png :)

ijustmadethisforplotly:
Bump

nicolaskruchten:
@etpinard would this really be a "breaking" change? In what scenarios would this not be a net improvement over the current arbitrary dimensions of the downloaded PNG when no width/height are set?

etpinard:
https://github.com/plotly/plotly.js/pull/3746 add a JSON-serializable way to do effectively what is presented in https://github.com/plotly/plotly.js/issues/1576#issuecomment-293043413

-----

Next, as @antoinerg points out in https://github.com/plotly/plotly.js/pull/3746#issuecomment-481012418, maybe we could:

> Could config options `toImageButtonOptions: {width: null, height: null}` be the default when responsive: true?

[...]

PR #3746, opened 2019-04-08
Add way for `toImage` to export images with current graph width/height

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3743 ... in a slightly more general way than proposed in that issue. 

This PR adds special logic to `Plotly.toImage` (which is called by `Plotly.downloadImage`, which is itself called when clicking on the `toImage` modebar button) for options `width:null` and `height:null`.

In brief, `Plotly.toImage(gd, {width:null, height:null})` will use the width/height values found in `gd._fullLayout` to generate its output image. Equivalently, 

```js
Plotly.react(gd, data, layout, {
  toImageButtonOptions: {width: null, height: null}
})
```

and clicking on the `toImage` modebar button will generate an image using the current `gd._fullLayout.width` and `gd._fullLayout.height` values - which (should at least) always store the current graph width/height values.

@plotly/plotly_js @nicolaskruchten - let me know what you think!

### DIFF
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -6,7 +6,6 @@
 * LICENSE file in the root directory of this source tree.
 */
 
-
 'use strict';
 
 var Registry = require('../../registry');
@@ -44,7 +43,6 @@ var modeBarButtons = module.exports = {};
  * @param {boolean} [toggle]
  *      is the button a toggle button?
  */
-
 modeBarButtons.toImage = {
     name: 'toImage',
     title: function(gd) {
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -65,7 +65,7 @@ modeBarButtons.toImage = {
         }
 
         ['filename', 'width', 'height', 'scale'].forEach(function(key) {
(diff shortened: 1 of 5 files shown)


nicolaskruchten:
Hah that’s actually what I had initially tried and thought might work so I’m totally down with that approach. /cc @chriddyp

chriddyp:
lgtm!

antoinerg:
Beautiful feature @etpinard !

Could config options `toImageButtonOptions: {width: null, height: null}` be the default when `responsive: true`?

etpinard:
> Could config options `toImageButtonOptions: {width: null, height: null}` be the default when `responsive: true`?

I'd call that a breaking change. For sure in v2 though.

antoinerg:
> I'd call that a breaking change. For sure in v2 though.

Ok, then the only request I have is to add it to the v2 wishlist!

:dancer:

etpinard:
> Ok, then the only request I have is to add it to the v2 wishlist!

:pencil2: in https://github.com/plotly/plotly.js/issues/1576#issuecomment-481015960 - for better context info.

Issue #3726, opened 2019-04-04
Hoverlabel contents alignment


nicolaskruchten:
We should add an attr to control hoverlabel contents alignment, mostly to be able to force right-alignment even when the hoverlabel is on the left.

etpinard:
Related: https://github.com/plotly/plotly.js/issues/3010

nicolaskruchten:
Well, what I'm proposing is much narrower than #3010 but yes that would be nice too :)

Can we get this attr slotted in for 1.47 ? with default = `auto` encompassing the current behaviour ?

etpinard:
By _alignment_ here, you mean:

- the alignment of the text content inside the hover label with respect the hover label container? or
- the alignment of the hover label with respect to its corresponding data point(s)?

nicolaskruchten:
Text within box

nicolaskruchten:
related #260

PR #3753, opened 2019-04-09
Align hover label text content with *hoverlabel.align*

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3726 by adding a new attribute under the `hoverlabel` trace and layout containers.

See demo: https://codepen.io/etpinard/pen/MRJvLr?editors=1010

I picked _align_ as its behavior reminded me of `align` in the annotations attributes:

https://github.com/plotly/plotly.js/blob/1dcee933ec8ac0648d6e5fc3476d923d19cb8eb7/src/components/annotations/attributes.js#L87-L99

Let me know if you can think of a better name.

### DIFF
--- a/src/components/fx/attributes.js
+++ b/src/components/fx/attributes.js
@@ -9,49 +9,25 @@
 'use strict';
 
 var fontAttrs = require('../../plots/font_attributes');
+var hoverLabelAttrs = require('./layout_attributes').hoverlabel;
+var extendFlat = require('../../extend').extendFlat;
 
 module.exports = {
     hoverlabel: {
-        bgcolor: {
-            valType: 'color',
-            role: 'style',
+        bgcolor: extendFlat({}, hoverLabelAttrs.bgcolor, {
             arrayOk: true,
-            editType: 'none',
-            description: [
-                'Sets the background color of the hover labels for this trace'
-            ].join(' ')
-        },
-        bordercolor: {
-            valType: 'color',
-            role: 'style',
+            description: 'Sets the background color of the hover labels for this trace'
+        }),
+        bordercolor: extendFlat({}, hoverLabelAttrs.bordercolor, {
             arrayOk: true,
-            editType: 'none',
-            description: [
-                'Sets the border color of the hover labels for this trace.'
-            ].join(' ')
-        },
+            description: 'Sets the border color of the hover labels for this trace.'
+        }),
         font: fontAttrs({
             arrayOk: true,
             editType: 'none',
             description: 'Sets the font used in hover labels.'
         }),
-        namelength: {
-            valType: 'integer',
-            min: -1,
-            dflt: 15,
-            arrayOk: true,
-            role: 'style',
-            editType: 'none',
-            description: [
-                'Sets the length (in number of characters) of the trace name in',
-                'the hover labels for this trace. -1 shows the whole name',
-                'regardless of length. 0-3 shows the first 0-3 characters, and',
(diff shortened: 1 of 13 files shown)


nicolaskruchten:
The default is center? I would have thought the default would be “auto” to describe the current behaviour

nicolaskruchten:
Also re the description: this doesn’t just apply to “texts right? It applies to the entire contents of the hoverlabel, even if it’s built with hovertemplate etc?

antoinerg:
> Also re the description: this doesn’t just apply to “texts right? It applies to the entire contents of the hoverlabel, even if it’s built with hovertemplate etc?

`hovertemplate` was implemented so that it reuses all the existing logic. So yes the `hoverlabel.align` does apply to the entire content of the hoverlabel even if it's built with `hovertemplate`! 

I think we could replace \`text\` by text in the description.

> The default is center? I would have thought the default would be “auto” to describe the current behaviour

The default is actually auto. However, `center` is not yet implemented in this PR. @nicolaskruchten is this a requirement?

etpinard:
> The default is center? I would have thought the default would be “auto” to describe the current behaviou

The default is `'auto'`.

etpinard:
Wow. Sorry I made things confusing. The block I pasted in the PR description is in `annotations/attributes.js` not the new `hoverlabel.align` attribute declaration which is:

```js
        align: {
            valType: 'enumerated',
            values: ['left', 'right', 'auto'],
            dflt: 'auto',
            role: 'style',
            editType: 'none',
            description: [
                'Sets the horizontal alignment of the text content within hover label box.',
                'Has an effect only if the hover label text spans more two or more lines'
            ].join(' ')
        },
```

as found in `fx/layout_attributes.js`.

antoinerg:
@etpinard about your https://github.com/plotly/plotly.js/issues/3726#issuecomment-481017727 . Since there are two different things we may want to align, wouldn't it be safer to call this new attribute `textalign` instead of `align`. I could see `align` being used to align with respect to the data points as you pointed out whereas in the current PR we deal with text alignment.

etpinard:
> Since there are two different things we may want to align, wouldn't it be safer to call this new attribute `textalign` instead of `align`

Fair. But you could say the same thing about annotations.

nicolaskruchten:
`center` is not important to me BTW :)

etpinard:
> `center` is not important to me BTW :)

:tada: that's what I thought.

nicolaskruchten:
✨ from me :)

antoinerg:
:dancer: once the tests pass!

etpinard:
Going with `hoverlabel.align` for consistency with annotations.

ScottGledhill:
Screenshot 2019-05-28 at 16 40 05
When no room issues with data it will display in the middle

Screenshot 2019-05-28 at 16 44 14
If the data is on the left itll show an arrow leftwards and text will align this way also

Screenshot 2019-05-28 at 16 40 37
If the data is on the right hand side itll show an arrow right and text will also align this way.

Plotly makes it really really hard to change this default behaviour, and i'm not sure if we should want to, forcing it right on the first screen shot would look weird, and having the arrow pointing left but the text aligned right would also look strange in my opinion.

After discussing with Ameer and Pete, I recommend keeping the default behaviour, if it does need to be done, someone with much more css skills than me will have to pick it up :)

nicolaromano:
Would it be possible to allow center-aligning text as well?

Issue #3754, opened 2019-04-10
Uncaught (in promise) TypeError: doneFnCompleted is not a function


lavor:
Hi, the release version 1.47.0 adds call of method `doneFnCompleted`:
https://github.com/plotly/plotly.js/commit/627fd74d07bdcfb4407c9241e5f00d92e09ab293#diff-8c5d8f688f2bb41a18d35397e5f58864R290

This method is initialized only in sankey's base_plot.js:
https://github.com/plotly/plotly.js/commit/627fd74d07bdcfb4407c9241e5f00d92e09ab293#diff-2d7d5b2875d00ef73d8c35008560ea17R90

Otherwise  `doneFnCompleted` it is not initialized and a browser reports an error whenever the Box select or the Lasso select from toolbar is used:
Uncaught (in promise) TypeError: doneFnCompleted is not a function

etpinard:
Thanks very much for the report :+1:

Issue #1705, opened 2017-05-18
promiseError disappeared?


alexcjohnson:
We still call `lib.promiseError` in [`lib.syncOrAsync`](https://github.com/plotly/plotly.js/blob/master/src/lib/index.js#L297) but it has disappeared? What do we want to have happen here?

rreusser:
Is this similar to `fail_test.js`?

https://github.com/plotly/plotly.js/blob/master/test/jasmine/assets/fail_test.js

etpinard:
From https://github.com/plotly/plotly.js/commit/7fa9243bd5005df988321d1bc730d5d3989bf63b#diff-e64ef427412e9bab4929b8b93ca12d96

jklimke:
Hi, this came to my notice as i am receiving constantly warnings when using plotly in an promise based environment. As the promiseError function does not exist anymore, it should not be passed to promises for error handling as this causes a warning

```
Warning: .then() only accepts functions but was passed: [object Undefined], [object Undefined]    at Object.push../node_modules/plotly.js/src/lib/index.js.lib.syncOrAsync (http://localhost:3000/packs/js/vendors~._node_modules_plotly.js_src_c-9f52df4a5f1ee8c919dc.chunk.js:3256:38)    at relayout (http://localhost:3000/packs/js/vendors~._node_modules_plotly.js_src_c-9f52df4a5f1ee8c919dc.chunk.js:10431:22)    at Object.push../node_modules/plotly.js/src/registry.js.exports.call (http://localhost:3000/packs/js/vendors~._node_modules_pl-6439151412dfacf8a22d.chunk.js:246:42)    at http://localhost:3000/packs/js/vendors~._node_modules_plotly.js_src_c-9f52df4a5f1ee8c919dc.chunk.js:28740:16
```

Deleting the line should not have side effects:

https://github.com/plotly/plotly.js/blob/e27ca3c691924b760a697f7176854cd91ae4dcd5/src/lib/index.js#L459

So please change it to the following code an leave error handling to the caller that provided a promise.
```
lib.syncOrAsync = function(sequence, arg, finalStep) {
    var ret, fni;

    function continueAsync() {
        return lib.syncOrAsync(sequence, arg, finalStep);
    }

    while(sequence.length) {
        fni = sequence.splice(0, 1)[0];
        ret = fni(arg);

        if(ret && ret.then) {
            return ret.then(continueAsync);
        }
    }

    return finalStep && finalStep(arg);
};

PR #3755, opened 2019-04-10
Don't try to call doneFnCompleted when it isn't set

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3754

@plotly/plotly_js here's the hotfix, we should investigate why our tests didn't catch this though.

### DIFF
--- a/src/plots/cartesian/select.js
+++ b/src/plots/cartesian/select.js
@@ -15,6 +15,7 @@ var Registry = require('../../registry');
 var Color = require('../../components/color');
 var Fx = require('../../components/fx');
 
+var Lib = require('../../lib');
 var polygon = require('../../lib/polygon');
 var throttle = require('../../lib/throttle');
 var makeEventData = require('../../components/fx/helpers').makeEventData;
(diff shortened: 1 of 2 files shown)


archmoj:
:dancer:

antoinerg:
Thanks for the fix @etpinard !!! :dancer:

etpinard:
Thanks for the review! I'm still trying to figure why our `select_test.js` are eating up console errors. 

I'll release this fix later this afternoon.

etpinard:
https://github.com/plotly/plotly.js/pull/3755/commits/98563547a97bf94bb60976b04f29ec46b27d992c :lock: down the fix. I was hoping to find a better to look for errors inside promise chains (I even tried bringing back `Lib.promiseError` cc https://github.com/plotly/plotly.js/issues/1705), but to no avail.

Oh well, I hope you think this will be good enough for 1.47.1.

etpinard:
Merging this thing in. 1.47.1 will be out in ~20 minutes

Issue #3162, opened 2018-10-25
Update eslint rules


antoinerg:
Following @etpinard's https://github.com/plotly/plotly.js/pull/3158#discussion_r228292109, we should probably update our linter syntax to reflect the current coding style.

In particular we should probably add `"one-var": { "initialized": "never", "uninitialized": "consecutive" }` and fix any possible errors.

**Update:** the `"initialized"` part got done in https://github.com/plotly/plotly.js/pull/3374

alexcjohnson:
> We should also add the camelcase rule

though we do have a few legacy snake_case attributes (`paper_bgcolor`, `error_x`...)

etpinard:
> egacy snake_case attributes (`paper_bgcolor`, `error_x`...)

Right, but with `{"properties": "never"}` we would only be looking at variable and function names.

etpinard:
One more from https://github.com/plotly/plotly.js/pull/3125/files#r227541345:

> Maybe we could add https://eslint.org/docs/rules/no-extra-parens to our lintinting rules.

etpinard:
from https://github.com/plotly/plotly.js/issues/3020

----

consider enabling https://eslint.org/docs/rules/no-shadow - can save us from some very confusing bugs.

As @etpinard mentions in #3017:
> That would require some work though, as no-shadow fails in more than 500 times in our codebase.

etpinard:
Potential new rule: `/*eslint quote-props: ["error", "as-needed"]*/`

https://eslint.org/docs/rules/quote-props

which fits the style in most (but not all!) of our src files.

etpinard:
Maybe update our [`no-multiple-empty-lines`](https://eslint.org/docs/rules/no-multiple-empty-lines) rule down to `{"max": 1}`

archmoj:
We previously started to dislike `else` statements sitting on new lines and tried to fix them in other PRs.
![lint-else](https://user-images.githubusercontent.com/33888540/55886817-6c24ce80-5b7a-11e9-8750-84db849c07e9.png)

To put all `else` and `catch` statements inline with the closing curly braces, I suggest we use `"brace-style": [2, "1tbs", {"allowSingleLine": true}],` instead of `"brace-style": [0, "stroustrup", {"allowSingleLine": true}],` then run `npm run lint -- --fix` on a separate PR (possibly now that we have only a few PRs open).
We could then fix just a few edge cases where there is a comment on top of an `else` statement.

[...]

PR #3757, opened 2019-04-11
Eslint 1tbs help update else statement blocks

archmoj:
Follow up of https://github.com/plotly/plotly.js/issues/3162#issuecomment-481713080 & https://github.com/plotly/plotly.js/issues/3162#issuecomment-481827954, 
[a new lint rule]( 38a9db3fce7ff6e94ec8a4399bf70a91becc31e8) is applied in this PR.

@plotly/plotly_js

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,7 @@
     "eslint:recommended"
   ],
   "parserOptions": {
-    "ecmaVersion": 5,
+    "ecmaVersion": 5
   },
   "env": {
     "commonjs": true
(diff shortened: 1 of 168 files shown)


etpinard:
Like I said in https://github.com/plotly/plotly.js/issues/3162#issuecomment-481827954, I'm a fan of this new rule, but perhaps we should let @plotly/plotly_js come to consensus first.

antoinerg:
I don't have a strong opinion on which linting rules we should adopt. If both @etpinard and @archmoj like it better, we should adopt it!

etpinard:
Alright, let's get this in :dancer:

Issue #3373, opened 2018-12-25
Inside texts in stacked bar charts are cut off for log scale


Ejdamm:
If the inside text for the bottom bar in a stacked bar chart does not fit it will be cut off, instead of shrinking. This only happens for log scale.

Look at this codepen example and compare how "giraffes in LA zoo" text is shrunk while "orangutangs in SF zoo" text is cut off. 
https://codepen.io/Ejdamm/pen/ebRJMN

etpinard:
You'll need to set `cliponaxis: false`. We currently clip text nodes around the subplot limit by default at the moment. Here's an example: https://codepen.io/etpinard/pen/BvdxvV (I made the text red, to see the unclipped text better).

Ejdamm:
Okay that helps but what I really want and think would be default behavior is that the text shrinks like in the rest of the bars.

You see in the orange giraffe bar, the text doesn't fit so it have shrunk instead of overflowing.

Ejdamm:
Yes the behavior for linear scales is consistent and what I am looking for in the log scale. My actual data are sometimes 100 for one bar and 1M on another so linear scale is not an option. Also the chart is made automatically every week so i can't set the fontsize manually. Thanks for the tips though!

PR #3762, opened 2019-04-11
Handle special bar insidetext case on log size axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3373

before: https://codepen.io/etpinard/pen/QPgLpK?editors=0010
after: https://codepen.io/etpinard/pen/dLRyoy?editors=1010

cc @plotly/plotly_js


src/traces/bar/plot.js
```diff
@@ -226,6 +226,21 @@ function appendBarText(gd, bar, calcTrace, i, x0, x1, y0, y1) {
     var insideTextFont = style.getInsideTextFont(trace, i, layoutFont, barColor);
     var outsideTextFont = style.getOutsideTextFont(trace, i, layoutFont);
 
+    // Special case: don't use the c2p(v, true) value on log size axes,
```
etpinard: See 

https://github.com/plotly/plotly.js/blob/1d0acf035ac054c983c3e6bb1b4a691a0c8cc8be/src/plots/cartesian/set_convert.js#L68-L80

for more info that `true` option.

antoinerg:
Thanks for the fix! Looks good to me :dancer:

PR #3763, opened 2019-04-11
Over 40 chart types

chriddyp:
Well, more like 40 exactly. But if you count `mode: line` and `mode: markers` as 2 then we'd be at 41.


etpinard:
Free :beers: to anyone that makes a script that auto-updates that part of the README on `npm version`.

-------

In the meantime, :dancer:

chriddyp:
my seasonal contribution to plotly.js

PR #3764, opened 2019-04-11
DRY loneHover routine

antoinerg:
`Fx.loneHover` logic is already contained in `Fx.multiHover` so I remove the duplication here.

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -77,7 +77,7 @@ exports.hover = function hover(gd, evt, subplot, noHoverEvent) {
 };
 
 /*
- * Draw a single hover item in a pre-existing svg container somewhere
+ * Draw a single hover item or an array of hover item in a pre-existing svg container somewhere
  * hoverItem should have keys:
  *    - x and y (or x0, x1, y0, and y1):
  *      the pixel position to mark, relative to opts.container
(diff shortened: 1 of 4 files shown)


etpinard:
Brilliant!

I'm a little worried about this block:

https://github.com/plotly/plotly.js/blob/26bc2bf4701efb7b7df0357fe065896c19c41661/src/components/fx/hover.js#L168-L189

affecting traces the use `Fx.loneHover` in bad ways. Could you manually check that `pie`, `gl3d`, `sunburst` and `sankey` traces look the same?

----

Moreover, maybe we could go one step further by merging `multiHovers` with `loneHover` completely. Looks like all you need is something like:

```js
return Array.isArray(hoverItems) ? hoverLabel : hoverLabel.node(); 
```

at the end.

antoinerg:
> Could you manually check that `pie`, `gl3d`, `sunburst` and `sankey` traces look the same?

I added a `console.log(d.offset)` after L188 and made sure `d.offset` is zero for all the traces you mentioned. In the future, if we support different alignment schemes for hover labels, we should definitely add test this in jasmine.

> Moreover, maybe we could go one step further by merging multiHovers with loneHover completely. 

Do you mean that I should remove all occurrences of `Fx.loneHover` in our codebase? I guess it would make sense to do so.

etpinard:
> Do you mean that I should remove all occurrences of `Fx.loneHover` in our codebase?

I would do opposite. Keep `loneHover` and merge `multiHovers` into `loneHover`. `loneHover` has more history in the repo (it's been there ever since `gl3d` got its hover labels).

etpinard:
Awesome :dancer:

PR #3766, opened 2019-04-12
Set enabled:false when filter target array is empty

etpinard:
closes https://github.com/plotly/plotly.js/issues/2908 - by implementing https://github.com/plotly/plotly.js/issues/2908#issuecomment-428754392

cc @plotly/plotly_js 

before/after example: https://codepen.io/etpinard/pen/wYdppB / https://codepen.io/etpinard/pen/WWOVBV


src/transforms/filter.js
```diff
@@ -138,10 +138,16 @@ exports.supplyDefaults = function(transformIn) {
     var enabled = coerce('enabled');
 
     if(enabled) {
+        var target = coerce('target');
+
+        if(Lib.isArrayOrTypedArray(target) && target.length === 0) {
```
archmoj: @etpinard Thanks for the PR.
Cool! 
What about handling a case like this: `[ [ ] ]`?  
See [codepen](https://codepen.io/MojtabaSamimi/pen/PgKYpr).

etpinard: Good call. Thanks for the review!

etpinard: @archmoj this turned out to be a bit more of a project than anticipated. 

See https://github.com/plotly/plotly.js/pull/3766/commits/3a5a4c2b2761dd232838b6c59109650a308a3077 - which makes things work for all trace types except `scattercarpet` and in some world-calendar scenario. I'll try to get those two cases fixed before `1.48.0`, but in the meantime here's "a good chuck" of the solution.

etpinard: > which makes things work for all trace types except `scattercarpet` and in some world-calendar scenario.

Now fixed by https://github.com/plotly/plotly.js/pull/3766/commits/b6432380de4b3474b0bec9b2f02d2bd49e416f15 and https://github.com/plotly/plotly.js/pull/3766/commits/40c8abc10cb029d1300c6b0c696acd71299bb3c9

Making this PR ready for a second :eye:

archmoj:
@etpinard you mentioned a possible conflict on `funnel` PR here: 
https://github.com/plotly/plotly.js/pull/3817#discussion_r279558929

etpinard:
> @etpinard you mentioned a possible conflict on `funnel` PR here:
> [#3817 (comment)](https://github.com/plotly/plotly.js/pull/3817#discussion_r279558929)

and the fix is in https://github.com/plotly/plotly.js/pull/3766/commits/84c3606648629ea1ba9fa14b8f39dcc0f9804312

archmoj:
Looks very good. 
Should we handle `[ [ ] ]` case also?

etpinard:
> Should we handle `[ [ ] ]` case also?

It is handled as it results in `trace._length=0`

archmoj:
> > Should we handle `[ [ ] ]` case also?
> 
> It is handled as it results in `trace._length=0`

Could you please provide a codepen example for that?
Thanks.

etpinard:
> Could you please provide a codepen example for that?

https://codepen.io/etpinard/pen/dEYQXW

archmoj:
> > Could you please provide a codepen example for that?
> 
> https://codepen.io/etpinard/pen/dEYQXW

Thanks. Now could you please check [this one](https://codepen.io/MojtabaSamimi/pen/QRjJKY). 
Is that the expected behaviour?

etpinard:
> Is that the expected behaviour?

I think so. It's not great, but I think trace with `_length===0` should still get a legend item as mentioned in https://github.com/plotly/plotly.js/issues/2908#issuecomment-428754392

archmoj:
Excellent.
[ [ :dancer: ] ]

Issue #3777, opened 2019-04-16
Waterfall textinfo


nicolaskruchten:
I think we'll need some `textinfo` for waterfalls, so as to be able to show the total value as text.

archmoj:
Here is [a prototype](https://codepen.io/MojtabaSamimi/pen/pBpZry?editors=0010) by reusing some of the logic in `pie`.

nicolaskruchten:
Pretty cool! What are the available flags? Can I show value+initial or something? If so, what does that end up looking like?

archmoj:
@nicolaskruchten the early prototype used to apply the options identical to `pie`.
However as @etpinard pointed out https://github.com/plotly/plotly.js/pull/3782#issuecomment-484089320 we may have different set of flags for waterfalls.
What about this list? 
[`label`, `text`, `sum value`, `diff value`, `diff percent`, `sum percent`]

nicolaskruchten:
what would those look like? one per line?

nicolaskruchten:
I would do:

* `value` (the delta value!), `initial`, `final`, and `text` (which is whatever was provided in the `text` attr)
* I would display these one-per-line

archmoj:
I like `delta` over `value`. Value is a bit confusing considering base, etc. 
Here is an updated [codepen](https://codepen.io/MojtabaSamimi/pen/WWMmEq?editors=0010).

Issue #3504, opened 2019-02-01
Funnel charts


nicolaskruchten:
Funnel charts are kind of interesting because they're kind of like stacked bar charts, with variable width bars, often with sloping sides, which makes for the potential for double-encoding. A standard stacked bar chart has each segment with the same width, and the height of the segment encodes the datum. On the other extreme you can have bars where each are the same height, and the width encodes the datum, like this: https://www.amcharts.com/demos/funnel-chart/

In both of those cases, the area is proportional to the datum. If you start sloping the sides, then the area no longer matches the datum and it's just the height: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/funnel/

So we could maybe add some options to `bar` to make the version without sloping sides, but it may make more sense to create a new trace type here. We'll want some special legend and color handling, as well as some computations for hovertemplate like 'percentage of total' and 'percentage difference with previous stage' etc.

alexcjohnson:
I really dislike the version where the whole side is sloped - I get that this makes the analogy to an actual funnel clearer, but it makes the data value highly ambiguous. But connecting bars with a sloping region in the gap, like the amcharts example does, strikes me as really cool, and would be a nice addition to regular bar charts too. This came up recently, I can't seem to find it now, but it's a great effect for stacked bars with some sort of stack sorting, because it shows you when two series change order in the stack.

At that point it would be fairly easy to imagine this just as another attribute to a bar chart to shift it to symmetric about the size-axis origin, but perhaps enough of the default attributes should change that it's worth making a trace type for it (and just use the bar code to render it)

alexcjohnson:
Oh I was thinking of this not as stacked at all, but a single trace of horizontal bars of equal height. That's how the amcharts example does it, and it shows you all the relevant information clearly: the total starting size, the size and amount lost at each stage, and it's intuitive that each stage is a subset of the previous stage.

The way that highcharts example shows it, where the vertical size of each segment is different, strikes me as totally inappropriate. Yes, it shows you the size at each stage (at least if you can figure out how to interpret the sloping sides), but the total stack size is completely meaningless and there's no way to visually see what's lost at each stage.

nicolaskruchten:
OK, so the a pie-like thing makes sense, if we want to be able to let people have different colours for each stage. I actually think that separate legend items per stage could be problematic if we allow click-to-hide :)

alexcjohnson:
[...]

PR #3782, opened 2019-04-17
exposing pie insert text function

archmoj:
This PR refactors and exposes `pie` insert-text function which appear to be useful in adding `textinfo` feature for `waterfall` & `funnel` in the next release.
See #3777,  #3504 and https://github.com/plotly/plotly.js/issues/3504#issuecomment-483681727.

@plotly/plotly_js

### DIFF
--- a/src/traces/pie/calc.js
+++ b/src/traces/pie/calc.js
@@ -85,27 +85,36 @@ function calc(gd, trace) {
     // include the sum of all values in the first point
     if(cd[0]) cd[0].vTotal = vTotal;
 
-    // now insert text
     if(trace.textinfo && trace.textinfo !== 'none') {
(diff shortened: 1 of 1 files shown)


etpinard:
Hmm. I'm not this is worth pursuing. Here's a couple reasons:

- the flags for `bar` and `waterfall` will be different, so as is we can't _just_ :recycle: `insideText`
- I'm not a fan of computing the full on-graph text content in `calc`, I would prefer having this in `plot` like for sunburst [traces](https://github.com/plotly/plotly.js/blob/172c3e6e6691b0fff362cbe55d32124c6130ef9f/src/traces/sunburst/plot.js#L718-L743).

archmoj:
Great! Closing in the favour of having this in `plot`.

PR #3783, opened 2019-04-17
fix svgTextUtils for FF: use child element instead of SVG to getSize()

### LINKED ISSUE #2616: mathjax doesn't render for firefox
image

originally reported in https://github.com/plotly/streambed/issues/2674, reported again in https://github.com/plotly/dash-core-components/pull/194

antoinerg:
Closes #2616 

The fix was tested in `Mozilla Firefox 62.0.3` in Linux. I try to fix the following misbehaving routine in FF:
https://github.com/plotly/plotly.js/blob/fe842c4cc7fa38fa8fb77b36a4010d7722cbe7b0/src/lib/svg_text_utils.js#L22-L24
However, after inspecting the DOM in Firefox, it turns out that it sets the size of the embedded `svg.gtitle-math` elements to a very large value. However, it sizes properly its first and only child element. We use this one instead when calling `getSize()`.

Codepen before: https://codepen.io/antoinerg/pen/oOEaPz
Codepen after: https://codepen.io/antoinerg/pen/ZZrqmd

Before:
newplot(1)
After:
newplot(2)

### DIFF
--- a/src/lib/svg_text_utils.js
+++ b/src/lib/svg_text_utils.js
@@ -114,10 +114,11 @@ exports.convertToTspans = function(_context, gd, _callback) {
                 .style({overflow: 'visible', 'pointer-events': 'none'});
 
                 var fill = _context.node().style.fill || 'black';
-                newSvg.select('g').attr({fill: fill, stroke: fill});
+                var g = newSvg.select('g');
+                g.attr({fill: fill, stroke: fill});
 
-                var newSvgW = getSize(newSvg, 'width');
-                var newSvgH = getSize(newSvg, 'height');
+                var newSvgW = getSize(g, 'width');
+                var newSvgH = getSize(g, 'height');
                 var newX = +_context.attr('x') - newSvgW *
                     {start: 0, middle: 0.5, end: 1}[_context.attr('text-anchor') || 'start'];
                 // font baseline is about 1/4 fontSize below centerline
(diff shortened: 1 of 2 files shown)


etpinard:
Brilliant fix! 

Wanna try to :lock: this down in a test?

Good news: `npm run test-jasmine -- --bundleTest=mathjax --FF` appears to fail on `master`. (1) Wanna make sure it passed off your branch? Once (1) is done, simply adding one more task (corresponding to `npm run test-jasmine -- --bundleTest=mathjax --FF` in [`test_bundle.js`](https://github.com/plotly/plotly.js/blob/master/tasks/test_bundle.js) should be good enough.

antoinerg:
> Good news: `npm run test-jasmine -- --bundleTest=mathjax --FF` appears to fail on `master`. (1) Wanna make sure it passed off your branch?

It works on branch `fix-2616`!

> Once (1) is done, simply adding one more task (corresponding to `npm run test-jasmine -- --bundleTest=mathjax --FF` in [`test_bundle.js`](https://github.com/plotly/plotly.js/blob/master/tasks/test_bundle.js) should be good enough.

I added a test in `test_bundle.js` for MathJax in Firefox in b1177a6. It gets executed on CircleCI as the last step:
Screenshot_2019-04-17_15-13-06

etpinard:
One of my favourite bug fixes of the year so far.

:dancer: :dancer: :dancer:

jonmmease:
Awesome!!! :tada:
This one has annoyed me ever since I got things working again in chrome last year.  Thanks for digging in!

etpinard:
To be released in tomorrow morning's `v1.47.3`

PR #3788, opened 2019-04-17
Waterfall connector.line.color coerce and default fixup

### LINKED ISSUE #3784: Waterfall connector.line.color not in _fullData
The reason there's no connector-line-color widget appearing in the RCE is that that attr isn't getting coerced (or something!) and hence not appearing in `_fullData`:

image

archmoj:
Fixes #3784 thanks to https://github.com/plotly/plotly.js/issues/3784#issuecomment-484237834.

@plotly/plotly_js 
cc: @nicolaskruchten


etpinard:
Nice. Do you have an idea why we need to update the baselines?

archmoj:
To avoid updating the baseline we may default to `#000` instead of `#444`.

etpinard:
Ha. Good to know! I prefer `#444`, but maybe someone with better eyesight would consider this a breaking change :thinking:

alexcjohnson:
`#000` should be considered a bug, it should match everything else.

etpinard:
Nicely done @archmoj :dancer:

PR #3789, opened 2019-04-18
Waterfall connector.line.width react bug in mode between

### LINKED ISSUE #3787: Waterfalls reacting weirdly for connector line thickness
The following CodePen emulates the `react()` calls that the RCE emits: https://codepen.io/nicolaskruchten/pen/vMdPNp?editors=0010 ... basically it mutates the figure and returns it with the identity of the `data` array changed, in order to trigger the `react()` logic. This works everywhere except `connector.line.width` in waterfalls for some reason ...

archmoj:
Fixes #3787.
[Codepen After](https://codepen.io/MojtabaSamimi/pen/ROMbwr?editors=0010).

@plotly/plotly_js 
cc: @nicolaskruchten

### DIFF
--- a/src/traces/waterfall/attributes.js
+++ b/src/traces/waterfall/attributes.js
@@ -96,7 +96,9 @@ module.exports = {
     connector: {
         line: {
             color: extendFlat({}, lineAttrs.color, {dflt: Color.defaultLine}),
-            width: lineAttrs.width,
+            width: extendFlat({}, lineAttrs.width, {
+                editType: 'plot', // i.e. to adjust bars is mode: 'between'. See https://github.com/plotly/plotly.js/issues/3787
+            }),
             dash: lineAttrs.dash,
             editType: 'plot'
         },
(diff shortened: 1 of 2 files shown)


src/traces/waterfall/attributes.js
```diff
@@ -96,7 +96,9 @@ module.exports = {
     connector: {
         line: {
             color: extendFlat({}, lineAttrs.color, {dflt: Color.defaultLine}),
-            width: lineAttrs.width,
+            width: extendFlat({}, lineAttrs.width, {
```
nicolaskruchten: does a similar fix need to be applied anywhere else in this trace? I haven't rigorously tested each attr, I just happened to test this one a lot due to our conversations around the line extensions etc :)

etpinard:
Ha, Good catch from this block:

https://github.com/plotly/plotly.js/blob/4f7592e2ba7a64bd07e4ba5ce1ee4f1fdcf05940/src/traces/bar/plot.js#L41-L45

which can't be transferred to `waterfal/style.js`.

------------

Nicely done :dancer:

PR #3790, opened 2019-04-18
Implementation of textinfo for waterfall traces

archmoj:
Addressing #3777 for `waterfall`.
A similar code could be useful for bar-like `funnel` traces.
[codepen](https://codepen.io/MojtabaSamimi/pen/xeWxbO?editors=0010).

@plotly/plotly_js 
cc: @nicolaskruchten


src/traces/bar/plot.js
```diff
@@ -475,12 +476,51 @@ function getTransform(textX, textY, targetX, targetY, scale, rotate) {
...
+    }
+    if(hasFlag('initial')) {
+        thisText.push(pieHelpers.formatPieValue(initial, separators));
+    }
+    if(hasFlag('text')) {
+        var tx = Lib.castOption(trace, cdi.i, 'text');
+        if(tx) thisText.push(tx);
+    }
+
+    return thisText.join('<br>'); // TODO: ' | ' may be a better dividor for bar
```
etpinard: Yep, it might be better to have `.join(' | ')` for horizontal waterfalls. But since I think most users will set only one `textinfo` flag, I doubt it's worth adding magic here.

src/traces/waterfall/attributes.js
```diff
@@ -76,6 +77,15 @@ module.exports = {
     hovertext: barAttrs.hovertext,
     hovertemplate: barAttrs.hovertemplate,
 
+    // labels (legend is handled by plots.attributes.showlegend and layout.hiddenlabels)
```
etpinard: I think you can :zap: this comment.

src/traces/waterfall/attributes.js
```diff
@@ -76,6 +77,15 @@ module.exports = {
...
+    textinfo: extendFlat({}, pieAtts.textinfo, {
+        editType: 'plot',
+        flags: ['text', 'delta', 'initial', 'final'],
```
etpinard: I think these flags are :ok_hand: 

We could maybe add `label`, but it might awkward as for `waterfall` it doesn't match a `labels` data array like for pie and sunburst traces. Maybe adding plain `'x'` and `'y'` flags would be best.

src/traces/bar/plot.js
```diff
@@ -469,12 +470,57 @@ function getTransform(textX, textY, targetX, targetY, scale, rotate) {
...
+        var tx = Lib.castOption(trace, cdi.i, 'text');
+        if(tx) text.push(tx);
+    }
+
+    if(trace.type === 'waterfall') {
+        var delta = +cdi.rawS || cdi.s;
+        var final = cdi.v;
+        var initial = final - delta;
+
+        if(hasFlag('initial')) text.push(pieHelpers.formatPieValue(initial, separators));
```
etpinard: Try using `Axes.tickText`

cc https://github.com/plotly/plotly.js/compare/fix3777-textinfo-waterfall-fin...funnel-new-proto#commitcomment-33323818

src/traces/bar/plot.js
```diff
@@ -469,12 +471,64 @@ function getTransform(textX, textY, targetX, targetY, scale, rotate) {
...
 }
 
-function getText(trace, index) {
-    var value = helpers.getValue(trace.text, index);
+function getText(calcTrace, index, xa, ya) {
+    var trace = calcTrace[0].trace;
+
+    var value;
+    if(!trace.textinfo || trace.textinfo === 'none') {
+        value = helpers.getValue(trace.text, index);
```
etpinard: Hmm. `textinfo: 'none'` should mean no on-graph text even if `text` is set. In fact, `textinfo: 'none'` is essentially equivalent to `textposition: 'none'`, perhaps we don't need it.

src/traces/waterfall/defaults.js
```diff
@@ -45,6 +45,8 @@ function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
     coerce('width');
 
     coerce('text');
+    coerce('textinfo');
```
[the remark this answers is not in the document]
archmoj: Done in https://github.com/plotly/plotly.js/pull/3790/commits/bed68427ea138dfb004500581a39350a6eec041b.

src/traces/bar/plot.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/3790/commits/bed68427ea138dfb004500581a39350a6eec041b.

etpinard: Please add jasmine test.

etpinard: Thanks!

test/image/mocks/waterfall_profit-loss_2018vs2019_textinfo_base.json
```diff
@@ -0,0 +1,157 @@
...
+    "layout": {
+        "title": {
+            "text": "Profit and loss statement 2018 vs 2019<br>waterfall chart"
+        },
+        "yaxis": {
+            "type": "category",
+            "autorange": "reversed"
+        },
+        "xaxis": {
+            "type": "linear"
```
etpinard: Can you try adding `ticksuffix` and `tickprefix` to these axes to see if things work as planned?

archmoj: Good call.
Done in the `funnel` PR https://github.com/plotly/plotly.js/pull/3817/commits/6a3a13c6f20348a5637eaa87a81d07eb28125145.

archmoj:
> What does `pie` do in this case?

Good question. No new lines when there is empty text.

archmoj:
> What does `pie` do in this case?

Good question. `Pie` does not leave a blank line when text is empty.

etpinard:
:dancer: nice job!

PR #3791, opened 2019-04-18
fixup waterfall connectors on date axes

### LINKED ISSUE #3785: Waterfalls handling date axes weirdly
This figure looks weird: connectors aren't rendering in the right place?

image

```json
{
  "data": [
    {
      "type": "waterfall",
      "mode": "markers",
      "decreasing": {
        "line": {
          "color": "#7f7f7f"
        }
      },
      "increasing": {
        "line": {
          "color": "#17becf"
        }
      },
      "x": [
        "2010-01-01",
        "2010-07-01",
        "2011-01-01",
        "2011-07-01",
        "2012-01-01",
        "2012-06-01"
      ],
      "xsrc": "dates",
      "y": [
        2,
        1,
        3,
        5,
        4,
        6
      ],
      "ysrc": "jagged ints"
    }
  ],
  "layout": {
    "xaxis": {
      "range": [
        "1998-06-24 10:40",
        "2228-12-06 13:20"
      ],
      "autorange": true
    },
    "yaxis": {
      "range": [
        -0.5,
        5.5
      ],
      "autorange": true
    },
    "autosize": true
  },
  "frames": []
}
```

archmoj:
Fixes #3785. 
This is similar to the way connectors would be handled in the `funnel` trace.
[codepen After](https://codepen.io/MojtabaSamimi/pen/jRzEmV?editors=0010).

@plotly/plotly_js 
cc: @nicolaskruchten


src/traces/waterfall/cross_trace_calc.js
```diff
@@ -65,6 +65,11 @@ module.exports = function crossTraceCalc(gd, plotinfo) {
             if(di.isSum === false) {
                 di.s0 += (j === 0) ? 0 : cd[j - 1].s;
             }
+
+            if(j + 1 < cd.length) {
```
etpinard: Non blocking, but I think `j < cd.length - 1` is more readable.

src/traces/waterfall/plot.js
```diff
@@ -60,11 +60,6 @@ function plotConnectors(gd, plotinfo, cdModule, traceLayer) {
             var x2, y2;
             var x3, y3;
 
-            var delta = 0;
-            if(i + 1 < len && Array.isArray(trace.offset)) {
-                delta -= trace.offset[i + 1] - trace.offset[i];
```
etpinard: Hmm. Where is `offset` taken in consideration? Do we have a mock with arrayOk `trace.offset` settings?

archmoj: Yes we have: https://github.com/plotly/plotly.js/blob/master/test/image/mocks/waterfall_custom.json

Also here is [a codepen](https://codepen.io/MojtabaSamimi/pen/ZZxoMa?editors=0010) using this branch.

src/traces/waterfall/cross_trace_calc.js
archmoj: I wanted to simply note that `j + 1` result is going to be used inside the `cd` array below.
So if you don't mind we may keep this the way it is.

src/traces/waterfall/plot.js
etpinard: Awesome!

etpinard:
Nicely done :dancer:

PR #3792, opened 2019-04-18
New lint rule to disallow named function expressions

archmoj:
A follow up of https://github.com/plotly/plotly.js/issues/3162,  https://github.com/plotly/plotly.js/issues/3162#issuecomment-483916923 and the discussion with @etpinard concerning redundant and sometime confusing function names this PR adds a new lint rule to disallow extra function names when they are not required. Image below illustrates one example where the function is named `put` and exported as `get`. 
https://github.com/plotly/plotly.js/blob/d2f8976f96f2a89ab9b6ac527338537e315d3b6e/src/traces/bar/sieve.js#L87
Screenshot from 2019-04-18 16-38-50

 @plotly/plotly_js


etpinard:
Ah ooops. I think I might have unestimated what adding this new rule meant this morning (in-person convo)

I very much dislike removing the function names from things we export off `module.exports` e.g.:

image

etpinard:
@archmoj maybe https://eslint.org/docs/rules/func-name-matching would give better results.

archmoj:
> @archmoj maybe https://eslint.org/docs/rules/func-name-matching would give better results.

Awesome! That one was able to detect the mismatches: https://github.com/plotly/plotly.js/pull/3792/commits/6d1609ab38cf50845d72b3ed7716072020bfc75f.

etpinard:
Ok. Nicely done :dancer:

PR #3811, opened 2019-04-26
Better did-margin-change check

### LINKED ISSUE #3561: Automargin regression since 1.44.2 ?

etpinard:
which hopefully should be enough to fix #3561 

cc @plotly/plotly_js

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -198,7 +198,7 @@ function plot(gd, data, layout, config) {
      * start async-friendly code - now we're actually drawing things
      */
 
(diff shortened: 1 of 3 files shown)


archmoj:
@etpinard thanks for the fix.
:dancer:

PR #3817, opened 2019-04-29
Funnel traces

### LINKED ISSUE #3117: Inside Bar Text Automatically Rotates
With a stacked bar chart, when a part of the stack becomes too small, inside positioned text automatically rotates 90 degrees. It would be useful to be able to control whether or not the text rotates.

archmoj:
This PR adds 'bar-like' `funnel` traces on Cartesian coordinates towards addressing #3504.
It would also add two `insidetext` features to `waterfall`, `bar` & `histogram` which help control the (`none` | `auto`) rotation and (`start` | `middle` | `end`) anchor position of those trace types (i.e. to fix #3117) as well as `funnel` traces . 

[Example codepen](https://codepen.io/MojtabaSamimi/pen/BEEYvd?editors=0010).

@plotly/plotly_js


src/components/drawing/index.js
```diff
@@ -110,8 +110,10 @@ drawing.hideOutsideRangePoints = function(traceGroups, subplot) {
         var trace = d[0].trace;
         var xcalendar = trace.xcalendar;
         var ycalendar = trace.ycalendar;
-        var selector = trace.type === 'bar' ? '.bartext' :
+        var selector =
```
etpinard: Non-blocking (so feel free to ignore), but maybe we could clean this up by adding a field on `trace._module` (i.e. on trace `index.js` exports) e.g. 

```js
// in bar/index.js

module.exports = {
  // ...
  hideOutsideRangeSelector: '.bartext'
};
```

src/traces/bar/attributes.js
```diff
@@ -86,6 +86,28 @@ module.exports = {
...
+    insidetextrotate: {
+        valType: 'enumerated',
+        values: ['auto', 'none'],
+        dflt: 'auto',
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Determines if texts could automatically be rotated to fit inside bars in `textposition` *inside* mode.'
+        ].join(' ')
+    },
```
etpinard: We'll need approval from @plotly/plotly_js and @nicolaskruchten for these new bar/waterfall/funnel attributes (perhaps during this week's meeting).

I'm a fan of `insidetextanchor` :ok_hand:  

For `insidetextrotate`, perhaps `textangle` (as proposed in https://github.com/plotly/plotly.js/issues/3117) would be better. If implementing a general `textangle` behavior would take too long, we could start with implementing an _enumerated_ `textangle` with values `auto`, `0` and `90`.

src/traces/funnel/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+    offset: barAttrs.offset,
+    width: barAttrs.width,
+
+    marker: barAttrs.marker,
+
+    connector: {
+        fillcolor: {
+            valType: 'color',
+            role: 'style',
+            editType: 'plot', // TODO: could we set this to 'style' ?
```
etpinard: Yes, but make sure to add a jasmine test to :lock: the behaviour down.

src/traces/funnel/calc.js
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+        size = ya.makeCalcdata(trace, 'y');
+        pos = xa.makeCalcdata(trace, 'x');
+    }
+
+    // create the "calculated data" to plot
+    var serieslen = Math.min(pos.length, size.length);
+    var cd = new Array(serieslen);
+
+    trace._base = [];
```
etpinard: Maybe add comment above this line to describe how `_base` is different than `base` for funnel traces?

src/traces/funnel/cross_trace_calc.js
```diff
@@ -0,0 +1,74 @@
...
+            _axisMatchGroups: fullLayout._axisMatchGroups,
+            _alignmentOpts: fullLayout._alignmentOpts,
+            barmode: fullLayout.funnelmode,
+            barnorm: fullLayout.funnelnorm,
+            bargap: fullLayout.funnelgap,
+            bargroupgap: fullLayout.funnelgroupgap
+        }
+    };
+
+    setGroupPositions(mockGd, xa, ya, funnelsVert);
```
etpinard: Next non-blocking clean-up opportunity: instead of creating a "mock" `gd` object here and for `waterfall` and `barpolar`, we should make `setGroupPosition` accept an `opts` option container with keys `mode`, `norm`, `gap` and `groupgap`. 

Again, this is non-blocking.

src/traces/funnel/defaults.js
```diff
@@ -0,0 +1,79 @@
...
+
+    handleText(traceIn, traceOut, layout, coerce, false);
+
+    coerce('marker.color', defaultColor);
+    coerce('marker.line.color', Color.defaultLine);
+    coerce('marker.line.width');
+
+    var connectorVisible = coerce('connector.visible');
+    if(connectorVisible) {
+        coerce('connector.fillcolor', Color.addOpacity(defaultColor, 0.5));
```
etpinard: I'm a fan of this, but we should probably double-check with @plotly/plotly_js and @nicolaskruchten

src/traces/funnel/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+    textfont: barAttrs.textfont,
+    insidetextfont: barAttrs.insidetextfont,
+    outsidetextfont: barAttrs.outsidetextfont,
+    constraintext: barAttrs.constraintext,
+
+    cliponaxis: barAttrs.cliponaxis,
+    orientation: barAttrs.orientation,
+
+    offset: barAttrs.offset,
+    width: barAttrs.width,
```
etpinard: `offset` and `width` seems to _work_ correctly, but I'd vote for removing them. I can't think of a scenario where a user would want to set these attributes.

test/image/mocks/blank-bar-outsidetext_horizontal.json
```diff
@@ -0,0 +1,33 @@
+{
+  "data": [
+    {
+      "type": "bar",
```
etpinard: Thanks for improving test coverage here! Could you combine these horizontal traces in mock into `blank-bar-outsidetext.json` in a separate subplot? That way we won't introduce too many new mocks

src/traces/funnel/defaults.js
```diff
@@ -0,0 +1,79 @@
...
+
+            handleGroupingDefaults(traceIn, traceOut, fullLayout, coerce);
+        }
+    }
+}
+
+module.exports = {
+    supplyDefaults: supplyDefaults,
+    crossTraceDefaults: crossTraceDefaults,
+    handleGroupingDefaults: handleGroupingDefaults
```
etpinard: We don't need to export this `handleGroupingDefaults`, right?

src/traces/funnel/plot.js
```diff
@@ -0,0 +1,154 @@
...
+        var group = Lib.ensureSingle(plotGroup, 'g', 'regions');
+
+        if(!trace.connector || !trace.connector.visible) {
+            group.remove();
+            return;
+        }
+
+        var isHorizontal = (trace.orientation === 'h');
+
+        if(!plotinfo.isRangePlot) cd0.node3 = plotGroup;
```
etpinard: We shouldn't need this line here. `node3` should be already linked up in `barPlot`.

src/traces/funnel/style.js
```diff
@@ -0,0 +1,60 @@
...
+
+        gTrace.selectAll('.point > path').each(function(di) {
+            if(!di.isBlank) {
+                var cont = trace.marker;
+
+                d3.select(this)
+                    .call(Color.fill, di.mc || cont.color)
+                    .call(Color.stroke, di.mlc || cont.line.color)
+                    .call(Drawing.dashLine, cont.line.dash, di.mlw || cont.line.width)
+                    .style('opacity', trace.selectedpoints && !di.selected ? 0.3 : 1);
```
etpinard: For bonus points, as `funnel` have (only) one `marker` style container, it should be relatively easy to implement `[un]selected.marker` styles like bar traces have:

https://github.com/plotly/plotly.js/blob/adf9bee14a8badcb5b2f49065fd594623421c82e/src/traces/bar/attributes.js#L199-L216

Maybe you can come back to when `funnelarea` is completed.

test/image/mocks/funnel_11.json
```diff
@@ -0,0 +1,249 @@
+{
+    "data": [
```
etpinard: It doesn't look this mock locks down anything special. Could you remove it?

test/image/mocks/funnel_and_bar.json
```diff
@@ -0,0 +1,55 @@
+{
+    "data": [
```
etpinard: Looks like there's already one pretty complete `funnel_axis_with_other_traces` mock. Can we remove `funnel_and_bar`, `funnel_and_histogram` and `funnel_line`? Or do these mocks lock down something in particular?

archmoj: Yes we can remove funnel and `bar`, `waterfall` and `line` but not the `histogram` for the moment.
I don't know why the histogram didn't properly show up on `funnel_axis_with_other_traces` setup.

archmoj: [Codepen](https://codepen.io/MojtabaSamimi/pen/eoojYo?editors=0010) please notice how the orange `histogram` has bars of equal size!

src/plots/cartesian/layout_defaults.js
```diff
@@ -93,6 +99,80 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
+        return (trace.xaxis) ? id2name(trace.xaxis) : undefined; // TODO: why these should be left undefined?
+    }
+
+    function getYname() {
+        return (trace.yaxis) ? id2name(trace.yaxis) : undefined; // TODO: why these should be left undefined?
+    }
+
+    function includesOnly(desiredType, xaNameIn, yaNameIn) {
+        var result = false;
+        for(var k = 0; k < fullData.length; k++) {
```
etpinard: Hmm. Am I seeing this right, we could `O(fullData.length^2)` loops? Is that really necessary to detect if all traces on a given axis are `funnel` traces. Am I missing something?

src/traces/bar/plot.js
```diff
@@ -530,5 +519,32 @@ function calcTextinfo(calcTrace, index, xa, ya) {
         if(hasFlag('final')) text.push(formatNumber(final));
     }
 
+    if(trace.type === 'funnel') {
+        if(hasFlag('value')) text.push(formatNumber(cdi.s));
```
etpinard: Do we have a mock testing `arrayOk` `textinfo` settings?

src/traces/bar/plot.js
```diff
@@ -530,5 +519,32 @@ function calcTextinfo(calcTrace, index, xa, ya) {
...
+            if(hasMultiplePercents) tx += ' of initial';
+            text.push(tx);
+        }
+        if(hasFlag('percent previous')) {
+            tx = formatNumber(Math.round(100 * cdi.difR), true) + '%';
+            if(hasMultiplePercents) tx += ' of previous';
+            text.push(tx);
+        }
+        if(hasFlag('percent total')) {
+            tx = formatNumber(Math.round(100 * cdi.sumR), true) + '%';
```
etpinard: Why are you using the `hover:true` option here?

archmoj: Otherwise the correct text didn't show up possibly related to the hidden axis?

etpinard: > Otherwise the correct text didn't show up possibly related to the hidden axis?

Probably because we don't coerce `tickformat`, `ticksuffix` and `tickprefix` when `ax.visible: false`. We should fix that.

test/image/mocks/funnel_axis.json
```diff
@@ -0,0 +1,143 @@
...
+            "xaxis": "x4",
+            "yaxis": "y4"
+        }
+    ],
+    "layout": {
+        "width": 800,
+        "height": 800,
+        "dragmode": "pan",
+        "xaxis": {
+            "type": "category",
```
etpinard: You should need to set `type: 'category'` in these mocks. Please :hocho:

test/image/mocks/funnel_axis.json
```diff
@@ -0,0 +1,143 @@
...
+        },
+        "yaxis3": {
+            "type": "category",
+            "domain": [
+                0.52,
+                1
+            ],
+            "autorange": "reversed"
+        },
+        "yaxis4": {
```
etpinard: Oh, when dragging this subplot

Peek 2019-04-29 17-41

I get:

image

test/image/mocks/funnel_axis_with_other_traces.json
```diff
@@ -0,0 +1,196 @@
...
+        "xaxis": {
+            "type": "category",
+            "domain": [
+                0,
+                0.48
+            ],
+            "autorange": true
+        },
+        "xaxis2": {
+            "type": "category",
```
etpinard: I _think_ you're missing `anchor` settings in this mock to make the y-axis labels not show up on top of one another.

We should have:

```
layout.xaxis2.anchor: 'y2'
layout.yaxis2.anchor: 'x2'
layout.xaxis3.anchor: 'y3'
layout.yaxis3.anchor: 'x3'
layout.xaxis4.anchor: 'y4'
layout.yaxis4.anchor: 'x3'
```

test/image/mocks/funnel_axis_with_other_traces.json
```diff
@@ -0,0 +1,196 @@
...
+                1
+            ],
+            "autorange": true
+        },
+        "xaxis3": {
+            "domain": [
+                0,
+                0.48
+            ],
+            "autorange": true
```
etpinard: ... and no need to set `autorange: true` for all these axes. It is the default.

test/image/mocks/funnel_cliponaxis-false.json
```diff
@@ -0,0 +1,109 @@
...
+  "layout": {
+	"funnelmode": "overlay",
+	"legend": {
+      "x": 0.5,
+      "xanchor": "center",
+      "y": -0.05,
+      "yanchor": "top"
+	},
+    "xaxis": {
+      "showline": true,
```
etpinard: Nice touch here - this makes that mock more much meaningful :trophy:

src/traces/funnel/hover.js
```diff
@@ -0,0 +1,57 @@
...
+    point.extraText = [
+        formatNumber(0.1 * (Math.round(1000 * di.begR))) + '% of initial',
+        formatNumber(0.1 * (Math.round(1000 * di.difR))) + '% of previous',
+        formatNumber(0.1 * (Math.round(1000 * di.sumR))) + '% of total'
+    ].join('<br>');
+    // TODO: Should we use pieHelpers.formatPieValue instead ?
+
+    point.color = getTraceColor(trace, di);
+
+    return [point];
```
etpinard: Peek 2019-04-29 17-50


I think the default hover label looks a little cluttered. Maybe we shouldn't show that position label and/or show 2 of those 3 percentages?

src/traces/funnel/layout_attributes.js
```diff
@@ -0,0 +1,64 @@
...
+            'Determines how bars at the same location coordinate',
+            'are displayed on the graph.',
+            'With *stack*, the bars are stacked on top of one another',
+            'With *group*, the bars are plotted next to one another',
+            'centered around the shared location.',
+            'With *overlay*, the bars are plotted over one another,',
+            'you might need to an *opacity* to see multiple bars.'
+        ].join(' ')
+    },
+    funnelnorm: {
```
etpinard: I don't think there's a single mock locking down `funnelnorm` at the moment, could you add one (unless I missed it)?

test/image/mocks/funnel_horizontal_stack_basic.json
```diff
@@ -0,0 +1,62 @@
+{
+    "data": [
+        {
+            "name": "plotly.js",
```
etpinard: Something looks off in the blue hover label path:

Peek 2019-04-29 17-55


This is probably not related to your work, can you confirm that `bar` traces show the same problem?

test/jasmine/tests/select_test.js
```diff
@@ -2192,6 +2192,75 @@ describe('Test select box and lasso per trace:', function() {
         .then(done);
     });
 
+    it('@noCI @flaky should work for funnel traces', function(done) {
```
etpinard: Funnel traces should probably have special `selectPoints` logic. Instead of selecting bars by their mid-position pt at full height, we should probably use their mid-position pt at **mid** height.

test/jasmine/tests/funnel_test.js
```diff
@@ -0,0 +1,1719 @@
...
+            return Plotly.restyle(gd, 'connector.visible', true);
+        })
+        .then(function() {
+            _assertNumberOfFunnelConnectorNodes(4);
+        })
+        .catch(failTest)
+        .then(done);
+    });
+
+    it('should be able to deal with blank bars on transform', function(done) {
```
etpinard: :sos: This one will conflict with https://github.com/plotly/plotly.js/pull/3766

test/image/mocks/funnel_11.json
etpinard: Ha I see you're using this mock in the hover tests. It's ok to keep it.

test/image/mocks/funnel_and_bar.json
etpinard: > I don't know why the histogram didn't properly show up on

_not_ probably as in the histogram bars have gaps in-between them?

etpinard: Oh no it's worse, the histogram bar height are wrong. It must be a conflict between the `bar` and `bar-like` categories.

src/traces/funnel/attributes.js
archmoj: TBD cc: @nicolaskruchten @alexcjohnson

src/traces/funnel/calc.js
archmoj: Good call. Done in 4fb5a2b166d7cfeda81a1783c71b1d550c0760ca.

src/traces/funnel/cross_trace_calc.js
archmoj: Good call. May addressed in 75a5cb43c5fca0d1e23a7d7a6c5f5ac053114e4c.

test/image/mocks/blank-bar-outsidetext_horizontal.json
archmoj: Done in 1504463.

src/traces/funnel/defaults.js
archmoj: Good catch! 
Fixed for `funnel` and `waterfall` in https://github.com/plotly/plotly.js/pull/3817/commits/68bd3715e940a0e15e8de38b0b0d217f891d76c6.

test/image/mocks/funnel_and_bar.json
archmoj: > Looks like there's already one pretty complete `funnel_axis_with_other_traces` mock. Can we remove `funnel_and_bar`, `funnel_and_histogram` and `funnel_line`? Or do these mocks lock down something in particular?

Right. Reduced mocks in https://github.com/plotly/plotly.js/pull/3817/commits/1434148694d331480025d49501557a167a9be76c.

src/traces/funnel/plot.js
archmoj: Good catch again! Fixed for waterfalls and funnels in https://github.com/plotly/plotly.js/pull/3817/commits/3d5b8ff69b7eb72fe6e5779d15cdaa3ed010412d.

src/traces/bar/plot.js
archmoj: No we don't have any. 
Should `textinfo` be `arrayOk`?
I think in `pie` it is not. Anyway for the moment `arrayOk`: `false` is added here: https://github.com/plotly/plotly.js/pull/3817/commits/8d02bcff0f997dd901a89d9b08af1199ef705cbb for better clarification.

test/image/mocks/funnel_axis.json
archmoj: Done in https://github.com/plotly/plotly.js/pull/3817/commits/c6155dbd1e2bd146848a0ada80cd7de44ada07ad.

test/image/mocks/funnel_axis_with_other_traces.json
archmoj: Thanks for the note. Done in https://github.com/plotly/plotly.js/pull/3817/commits/020a34b9744acb754eb41ee101ff8e4788cf96d9.

test/image/mocks/funnel_axis_with_other_traces.json
archmoj: Done in https://github.com/plotly/plotly.js/pull/3817/commits/c6155dbd1e2bd146848a0ada80cd7de44ada07ad.

src/traces/funnel/attributes.js
archmoj: Revised in https://github.com/plotly/plotly.js/pull/3817/commits/4c741c7f477d53033ec320cbbb52c9c8c1d33f1d. Also regarding the test coverage we have these two blocks: https://github.com/plotly/plotly.js/blob/4c741c7f477d53033ec320cbbb52c9c8c1d33f1d/test/jasmine/tests/funnel_test.js#L1132-L1204.

nicolaskruchten:
Aye, one of the side-effects of using normal cartesian coordinates here is that we're forcing every user of `funnel` to set `trace.orientation="h"` and `layout.yaxis.autorange = "reversed"` which is pretty obscure.

nicolaskruchten:
So maybe I'm looking at an out of date codepen but https://codepen.io/MojtabaSamimi/pen/BEEYvd?editors=0010 here `textinfo = "percent of initial"` isn't working

nicolaskruchten:
In `stack` mode, "% of total" is ambiguous... we take it to mean "percent of total *of trace*" but there is another more total total: total of *all traces*.

archmoj:
> So maybe I'm looking at an out of date codepen but https://codepen.io/MojtabaSamimi/pen/BEEYvd?editors=0010 here `textinfo = "percent of initial"` isn't working

Just updated the codepen now the flag is `percent initial` without `of`. Thanks for the note @nicolaskruchten.

src/traces/funnel/hover.js
nicolaskruchten: I think this is fine, and the user can pare down to what they like. The vast majority of funnels will be single-trace, so it won't seem that bad.

src/traces/funnel/attributes.js
nicolaskruchten: I can't think of a reasonable use-case for setting these, so let's remove them for now and we can always add them back in later if it turns out there is a killer app for them.

archmoj:
> In `stack` mode, "% of total" is ambiguous... we take it to mean "percent of total _of trace_" but there is another more total total: total of _all traces_.

@nicolaskruchten Thanks for bringing this up. I was also wondering `% of sum` may be less ambiguous?

src/traces/funnel/defaults.js
nicolaskruchten: This is what makes the fill color the same color as the bar but lighter? This is as requested by me, so I approve :)

src/traces/bar/attributes.js
nicolaskruchten: For funnels, I'd rather we not do the funky rotation stuff by default :)

archmoj: Oh yes. As discussed for funnels the auto rotation and text anchors of inside texts are set differently from bars and waterfalls: 
https://github.com/plotly/plotly.js/blob/4c741c7f477d53033ec320cbbb52c9c8c1d33f1d/src/traces/funnel/attributes.js#L39-L40

nicolaskruchten: Well, IMO all new traces shouldn't do it... is it possible to disable it by default for waterfall also or is it too late? sunburst? :)

etpinard: > shouldn't do it

Shouldn't do what? Auto-Rotate the labels?

src/traces/funnel/attributes.js
etpinard: Thanks. Resolved.

src/traces/funnel/calc.js
etpinard: Thanks!

src/traces/funnel/cross_trace_calc.js
etpinard: I'm not a fan of https://github.com/plotly/plotly.js/commit/75a5cb43c5fca0d1e23a7d7a6c5f5ac053114e4c - we could do better.

This diff in `barpolar/calc.js` is particularly not great:

```js
-   setGroupPositions(mockGd, aAxis, rAxis, barPolarCd);
+   setGroupPositions('bar', mockGd, aAxis, rAxis, barPolarCd);
```

I would much prefer having:

```js
// barpolar/calc.js

setGroupPositions(gd, aAxis, rAxis, {
  mode: polarLayout.barmode,
  norm: polarLayout.barnorm,
  gap: polarLayout.bargap,
  groupgap: polarLayout.bargroupgap
})
```

src/traces/bar/attributes.js
etpinard: TBD Thursday :busts_in_silhouette:

src/traces/funnel/defaults.js
etpinard: Thanks!

src/traces/funnel/plot.js
etpinard: One more here:

https://github.com/plotly/plotly.js/blob/3d5b8ff69b7eb72fe6e5779d15cdaa3ed010412d/src/traces/funnel/plot.js#L92

src/traces/bar/plot.js
etpinard: I can see value in having `arrayOk:true` for `textinfo`, but no need to do this in this PR.

src/traces/funnel/attributes.js
alexcjohnson: Removing sounds good, only thing I can think of is shifting the bars wrt the axis labels, ie you want all the labels at one end of the bars... but I don’t know why you would want that.

test/image/mocks/funnel_and_bar.json
etpinard: Ok, thanks. I'll keep this the thread unresolved as the `histogram` + `funnel` problem appears to still be a thing.

test/image/mocks/funnel_axis_with_other_traces.json
etpinard: Thanks!

src/traces/funnel/defaults.js
alexcjohnson: Shouldn’t we be adding opacity to `marker.color`? So if you set your own color the fill matches it by default?

etpinard:
> `trace.orientation="h"` and `layout.yaxis.autorange = "reversed"`

It would be easy to default `orientation` to `'h'` in funnel traces. 

We _could_ also make default `autorange` to `'reversed'` on y-axes that only have funnels (the same subset of graphs that will get `axis.visible:false`). That one might be too _magical_ and might lead to some undesired consequences though.

src/traces/funnel/hover.js
etpinard: > I think this is fine, and the user can pare down to what they like. The vast majority of funnels will be single-trace, so it won't seem that bad.

I would be in favour of a different `hoverinfo` default for traces that appear on subplots with multiple `funnel` traces.

nicolaskruchten: ok

nicolaskruchten:
> We could also make default autorange to 'reversed'

Let's investigate what might break as a result of this magic, as I think it will be very helpful. Just imagining the UI in Chart Studio for this and the confusion that users will have is... unpleasant.

src/traces/bar/attributes.js
nicolaskruchten: > Shouldn't do what? Auto-Rotate the labels?

yes. discuss thursday :)

test/jasmine/tests/select_test.js
archmoj: Good call. Addressed in eea4a72a695e238832506803a2358360cce55c69 and updated the PR codepen.

test/image/mocks/funnel_and_bar.json
archmoj: Histogram case is added in https://github.com/plotly/plotly.js/pull/3817/commits/95076408fb63493313b8c3757ebe2966c521ea8a.

etpinard: @archmoj in https://github.com/plotly/plotly.js/commit/95076408fb63493313b8c3757ebe2966c521ea8a, it seems that the histogram trace coexist properly with funnel traces. Is that correct? In other words, are the concerns from

>  It must be a conflict between the bar and bar-like categories.

not founded?

test/jasmine/tests/select_test.js
etpinard: I like it. Thanks!

test/image/mocks/funnel_and_bar.json
archmoj: That is correct. & I think so.

etpinard: Ok, great. Resolved.

archmoj:
> Hmm the baseline still looks odd:
> 
> image
> 
> It looks like you have an stray x-axis covering up the `x` labels.

@etpinard That one is an old version. It is fixed thanks to the instruction you sent.
https://github.com/plotly/plotly.js/pull/3817/files?short_path=c401c67#diff-c401c672de381c6580de57289d146845

src/traces/funnel/cross_trace_calc.js
archmoj: @etpinard Now it should be addressed in https://github.com/plotly/plotly.js/pull/3817/commits/86c0c092c7962d7b66e2131300bb2665b3e39378.
Thanks.

test/image/mocks/blank-bar-outsidetext_horizontal.json
etpinard: Thansk!

src/traces/funnel/cross_trace_calc.js
etpinard: https://github.com/plotly/plotly.js/commit/86c0c092c7962d7b66e2131300bb2665b3e39378 is lovely.

Resolved.

test/image/mocks/blank-bar-outsidetext_horizontal.json
archmoj: <b>Correction</b>: done in https://github.com/plotly/plotly.js/pull/3817/commits/95076408fb63493313b8c3757ebe2966c521ea8a.

src/traces/funnel/plot.js
archmoj: Good eye!
Fixed in https://github.com/plotly/plotly.js/pull/3817/commits/bdbbeb1b0438c25bcdf2566b7c19a2b64e2db745.

etpinard: Resolved.

src/traces/bar/plot.js
archmoj: Actually this was not related to hidden axes. We need to pass `true` to `ticktext` function to enable high precision and similarly for `waterfalls` as done in  https://github.com/plotly/plotly.js/pull/3817/commits/cad2728b964a373cd7593f93217501c4588b1799.

src/traces/funnel/attributes.js
archmoj: I think we may keep both. Here is [an example of use](https://codepen.io/MojtabaSamimi/pen/BEeMrz?editors=0010).

archmoj: <b>Hint</b>: click on each section of the `plotly.js` building.

test/image/mocks/funnel_axis.json
archmoj: Addressed in https://github.com/plotly/plotly.js/pull/3817/commits/04049f2ca3c440d4e4da429b1481f58cdd660daa .

src/traces/funnel/defaults.js
archmoj: Good call. Fixed in https://github.com/plotly/plotly.js/pull/3817/commits/6c8a661fe4b458915ce8428a765c9a6dd0efd3b1.

src/traces/funnel/layout_attributes.js
archmoj: `Funnelnorm` is no longer supported: https://github.com/plotly/plotly.js/pull/3817/commits/2a3c3543ed8fc64016fa16104dfa82715633822a.

test/image/mocks/funnel_axis.json
etpinard: Thanks!

src/components/drawing/index.js
etpinard: So you didn't need to add `Drawing.setClipUrl` calls as in:

https://github.com/plotly/plotly.js/blob/aec946e8570c042d632352fc58eaf99276797bd5/src/traces/bar/plot.js#L170


in https://github.com/plotly/plotly.js/pull/3817/commits/04049f2ca3c440d4e4da429b1481f58cdd660daa to get this to work?

src/traces/funnel/defaults.js
etpinard: Oh, good thinking about handling the arrayOk case, I not entirely sure your

```js
        var defaultFillcolor = Color.addOpacity(
            Lib.isArrayOrTypedArray(markerColor) ? markerColor[0] : markerColor, 0.5
        );
        coerce('connector.fillcolor', defaultFillcolor);
```

is ideal. How hard would it be to add `arrayOk:true` support for `connector.fillcolor`?

src/traces/funnel/layout_attributes.js
etpinard: Nicely done :tada:

src/traces/funnel/defaults.js
etpinard: Alternatively, we could not set the `connector.fillcolor` default in the case where `marker.color` is an array, and then turn 

https://github.com/plotly/plotly.js/blob/6a3a13c6f20348a5637eaa87a81d07eb28125145/src/traces/funnel/style.js#L41-L43

into

```js
        gTrace.selectAll('.regions').each(function(di) {
            d3.select(this).selectAll('path')
               .style('stroke-width', 0)
               .call(Color.fill, trace.connector.fillcolor || di.mc);
        });
```

alexcjohnson: that's weird though - the connector links bars `i` and `i+1`, so which `marker.color` do we take, `i` or `i+1`? I mean, I'd be OK with accepting an array of length `n-1` (though... `_length` & transforms 🤦‍♂), but pulling `n-1` items out of a length-`n` array seems like a bad idea.

I'd fall back on something like `rgba(0,0,0,0.5)` as the default when `marker.color` is an array. Not even `defaultColor`, as the user has already said they don't want that.

archmoj: Totally OK to change this to the way it should be. 
However; one may argue that the current behaviour is consistent with the way labels are presented. Also since funnel charts often have greater values in their very first elements (as the whole) it may be OK to present them this way.    
https://github.com/plotly/plotly.js/pull/3817/commits/6c8a661fe4b458915ce8428a765c9a6dd0efd3b1?short_path=4dff010#diff-4dff0100e1e69e85f5abd051fbd5c9f2 
Screenshot from 2019-05-01 20-06-03

etpinard: > I'd fall back on something like `rgba(0,0,0,0.5)` as the default when `marker.color` is an array.

I'm a fan of this :ok_hand:

src/plots/cartesian/axis_defaults.js
```diff
@@ -69,15 +67,22 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
     // try to get default title from splom trace, fallback to graph-wide value
     var dfltTitle = splomStash.label || layoutOut._dfltTitle[letter];
 
+
+    handleTickLabelDefaults(containerIn, containerOut, coerce, axType, options);
```
etpinard: Unfortunately, there's a bunch of attributes that get coerced in `handleTickLabelDefaults` that will have no effect when `ax.visible:false`.

I suggest moving back  `handleTickLabelDefaults` back below the `if(!visible) ...`  statement and moving the `coerce()` calls for `tickformat`, `tickprefix` and `ticksuffix` here above the `if(!visible) ...`

archmoj: Good call. Revised in https://github.com/plotly/plotly.js/pull/3817/commits/e968da68abbc70c67a0b08f646d266a2e1754dd2.

src/traces/funnel/attributes.js
archmoj: `arrayOk` `width` & `offset` are dropped from `funnel` here:
https://github.com/plotly/plotly.js/blob/c29b58d25b1d61749ddeded441420979f8ad8d13/src/traces/funnel/attributes.js#L49-L50 and the tests are adjusted.
As mentioned, `waterfall` may have a similar issue. So if you think that's a bug, please feel free to open an issue. Thanks.

src/traces/funnel/defaults.js
archmoj: This is fixed here: 
https://github.com/plotly/plotly.js/blob/c29b58d25b1d61749ddeded441420979f8ad8d13/src/traces/funnel/defaults.js#L64-L68
Note it was also required to take into account the `trace.opacity`.
So here is the new baseline https://github.com/plotly/plotly.js/blob/c29b58d25b1d61749ddeded441420979f8ad8d13/test/image/baselines/funnel_custom.png
which renders https://github.com/plotly/plotly.js/blob/e968da68abbc70c67a0b08f646d266a2e1754dd2/test/image/mocks/funnel_custom.json

src/components/drawing/index.js
archmoj: We had that here: https://github.com/plotly/plotly.js/blob/e968da68abbc70c67a0b08f646d266a2e1754dd2/src/traces/funnel/plot.js#L74-L76

etpinard: Cool. Thanks!

src/traces/funnel/defaults.js
etpinard: Nicely done :trophy:

src/traces/funnel/hover.js
etpinard: Closing this one. No need to this!

src/plots/cartesian/axis_defaults.js
etpinard: Nicely done :ok_hand:

src/traces/funnel/attributes.js
etpinard: Thanks!

src/traces/funnel/defaults.js
```diff
@@ -40,7 +40,9 @@ function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
 
     // TODO: move this block to bar handleText if/when textinfo implimented for bars/histograms
     if(traceOut.textposition !== 'none') {
-        coerce('textinfo');
+        var defaultTextinfo =
+            Lib.isArrayOrTypedArray(text) ? 'text+value' : 'value';
+        coerce('textinfo', defaultTextinfo);
```
etpinard: Or-par with `pie`:

image

thanks!

src/traces/funnel/attributes.js
```diff
@@ -0,0 +1,86 @@
...
+            editType: 'plot',
+            description: [
+                'Determines if connector regions and lines are drawn.'
+            ].join(' ')
+        },
+        editType: 'plot'
+    },
+
+    offsetgroup: barAttrs.offsetgroup,
+    alignmentgroup: barAttrs.offsetgroup
```
etpinard: should be `alignmentgroup: barAttrs.alignmentgroup`

src/traces/funnel/defaults.js
```diff
@@ -0,0 +1,90 @@
...
+        return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
+    }
+
+    var len = handleXYDefaults(traceIn, traceOut, layout, coerce);
+    if(!len) {
+        traceOut.visible = false;
+        return;
+    }
+
+    coerce('orientation', (traceOut.x && !traceOut.y) ? 'h' : 'v');
```
etpinard: Remember, we should make the default `orientation: 'h'` - unless `(!traceOut.x && traceOut.y)`

src/traces/funnel/attributes.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/3817/commits/61b6068cda0761ed877846a00cb110da4b0f98f3. Note: there was a similar problem in `waterfall`.

src/traces/bar/attributes.js
archmoj: General `textangle` for `bar`, `waterfall` and `funnel` addressed in https://github.com/plotly/plotly.js/pull/3817/commits/380db2ecaa750ae317c06e2aaf1d31aac3139bdf. `Funnel` has a different default i.e. `0` not `auto`.

src/traces/funnel/attributes.js
etpinard: Thanks!

src/traces/funnel/defaults.js
etpinard: Resolved!

test/image/mocks/funnel_axis.json
```diff
@@ -30,6 +31,7 @@
             ],
             "textposition": "inside",
             "insidetextanchor": "start",
+            "textangle": 0,
```
etpinard: Do we _need_ this line? Isn't `textangle: 0` by default for funnel traces?

src/traces/bar/attributes.js
etpinard: Looking good! Thanks for going the extra mile and making `textrotate` work for general angles! Your two new mocks are absolutely brilliant.

Now, I see that you updated the `bar_attrs_relative.png` baseline. Why is that?

etpinard: One more thing, `textangle` for `textposition: 'outside'` looks off. For example:

```js
Plotly.newPlot(gd, [{
  type: 'bar',
  y: [1, 2, 1],
  text: ['Loong', 'Loonggggerr', 'Suuuupeeeer Loooooong'],
  textposition: 'outside',
  textangle: -20,
  cliponaxis: false
}])
```

gives:

image

I think we should move the text-anchor so the text nodes are fully outside the bars.

----------

Now, I don't want you spending too much time on fixing `textangle` for `textposition: 'outside'` as it isn't part of our requirements for 1.48.0. If you think fixing that will take you more than a few hours, please make `textangle` work only for `textposition: 'inside'` for now and move on. Thanks!

src/plots/cartesian/layout_defaults.js
archmoj: Good call. Reworked in https://github.com/plotly/plotly.js/pull/3817/commits/a650170bfcee35743e0b3eb5aa42e2e12b8518d6.

test/image/mocks/funnel_axis.json
archmoj: No we don't need them anymore. Fixed in https://github.com/plotly/plotly.js/pull/3817/commits/a7446899f8c48924509c843a42c9554412a4aa9c.

src/traces/bar/attributes.js
archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/3817/commits/30b172a613626b491b8b814b4a392d45f8b4c681.

archmoj: Still extra padding required for this [codepen](https://codepen.io/MojtabaSamimi/pen/OYPOLd?editors=0010).

archmoj: Now fixed in https://github.com/plotly/plotly.js/pull/3817/commits/446baccaed5712736a21fea85f33078347db0ef5.
[Codepen](https://codepen.io/MojtabaSamimi/pen/joEYxZ?editors=0010)

src/plots/cartesian/layout_defaults.js
etpinard: Wow. So clean. Thanks!

test/image/mocks/funnel_horizontal_stack_basic.json
archmoj: Copied from @alexcjohnson message.
<i>Here’s the original comment about rotated & shifted hover labels - the issue got closed without spinning out a new issue for this part that’s still open </i> https://github.com/plotly/plotly.js/issues/3050#issuecomment-424508921

etpinard: @archmoj #3050 is closed. Can you open a new issue?

src/traces/bar/attributes.js
etpinard: > Now fixed in [446bacc](https://github.com/plotly/plotly.js/commit/446baccaed5712736a21fea85f33078347db0ef5).

Very nice!

Now, what about

> Now, I see that you updated the bar_attrs_relative.png baseline. Why is that?

test/image/mocks/funnel_horizontal_stack_basic.json
archmoj: Open this one: https://github.com/plotly/plotly.js/issues/3838.

etpinard: Thanks!

src/traces/funnel/defaults.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3817/commits/b93e3a5a83b6561ac6258a59f274b5fc87630c3e.
What about `autorange: "reversed"`?

etpinard: > Done in b93e3a5.

Nice! Can you mention something about the distinct behaviour in the funnel `orientation` attribute `description`?

etpinard: > What about `autorange: "reversed"`?

That item is still unresolved.

etpinard:
@archmoj you're almost there. Here's my list of blocking items:

- [ ] Figure out why the `bar_attrs_relative.png` baseline changed
- [ ] Make `autorange` default to 'reversed' on y-axes that only have horizontal funnels
- [ ] Improve the `orientation` attribute description

archmoj:
> @archmoj you're almost there. Here's my list of blocking items:

@etpinard Thanks for the review and checklist.

> * [ ]  Figure out why the `bar_attrs_relative.png` baseline changed

Good call. Done in https://github.com/plotly/plotly.js/pull/3817/commits/6d50386f775a993990eb772495b034e03a36ebc0 so that the none of the current mocks change.

> * [ ]  Make `autorange` default to 'reversed' on y-axes that only have horizontal funnels

Done in https://github.com/plotly/plotly.js/pull/3817/commits/d0e8fe569eaea6b7be4926e7e2c0741242362727.

> * [ ]  Improve the funnel `orientation` attribute description

Done in https://github.com/plotly/plotly.js/pull/3817/commits/fd1f6ca218c70dee563c3ab424067548ccedaf29.

archmoj:
@nicolaskruchten @alexcjohnson @etpinard 
Please double check `funnel` and `waterfall` `textinfo` print order & naming here: 
https://github.com/plotly/plotly.js/blob/6d50386f775a993990eb772495b034e03a36ebc0/src/traces/bar/plot.js#L536-L580

src/plots/cartesian/layout_defaults.js
```diff
@@ -180,6 +187,13 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
 
         var overlayableAxes = getOverlayableAxes(axLetter, axName);
 
+        var visibleDflt =
+            (axLetter === 'x' && !xaMustDisplay[axName] && xaMayHide[axName]) ||
+            (axLetter === 'y' && !yaMustDisplay[axName] && yaMayHide[axName]);
+
+        var reverseDflt =
+            (axLetter === 'y' && !yaMustForward[axName] && yaMayBackward[axName]);
```
etpinard: So clean. Wow :100: / :100:

etpinard:
This PR is a triumph :trophy: 

Not only do we get a new trace type in (i.e. `funnel` :tornado: ), but we also get:

- new `bar` and `waterfall` attribute `insidetextanchor` :anchor: 
- new `bar` and `waterfall` attribute `textangle`, supporting general [-180,180] angles :rotating_light: 
- through https://github.com/plotly/plotly.js/pull/3790, new `waterfall` attribute `textinfo` :memo: 
- multiple internal cleanups in _bar-like_ land :palm_tree: 

@archmoj I know you still have a question about the `textinfo` behaviour in https://github.com/plotly/plotly.js/pull/3790, so you might want to before wait merging this PR, but regardless

:dancer: :dancer: :dancer: :dancer: :dancer: :dancer: :dancer:

src/traces/funnel/defaults.js
etpinard: Now resolved.

src/traces/bar/attributes.js
etpinard: Now resolved!

etpinard:
Just to note a few things this PR _didn't_ do:

- `[un]selected.marker` styles
- `arrayOk` support for `textinfo`
- fix https://github.com/plotly/plotly.js/issues/3838
- `funnelarea` :smirk:

etpinard:
> @nicolaskruchten @alexcjohnson @etpinard
> Please double check `funnel` and `waterfall` `textinfo` print order & naming here:

:+1: from me.

archmoj:
@etpinard Thanks very much for the review and encouragement.
I will move forward and merge this thing into https://github.com/plotly/plotly.js/pull/3790.
There I wait if we need any changes regarding this comment: https://github.com/plotly/plotly.js/pull/3817#issuecomment-490210429.

PR #3819, opened 2019-04-29
reduce plot api imports

archmoj:
Refactoring plot_api files (i.e. excluding `plot_api.js`) to reduce imported functions and help increase code readability.

@etpinard

### DIFF
--- a/src/plot_api/helpers.js
+++ b/src/plot_api/helpers.js
@@ -12,15 +12,15 @@
 var isNumeric = require('fast-isnumeric');
 var m4FromQuat = require('gl-mat4/fromQuat');
 
-var Registry = require('../registry');
 var Lib = require('../lib');
-var Plots = require('../plots/plots');
-var AxisIds = require('../plots/cartesian/axis_ids');
+var subplotsRegistry = require('../plots/plots').subplotsRegistry;
 var Color = require('../components/color');
 
-var cleanId = AxisIds.cleanId;
-var getFromTrace = AxisIds.getFromTrace;
-var traceIs = Registry.traceIs;
+var cleanId = require('../plots/cartesian/axis_ids').cleanId;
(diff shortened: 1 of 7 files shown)


etpinard:
To be honest @archmoj, I don't see how this makes our code more _readable_.

I agree our import/export statements could be standardised better, but I think we can wait when we'll make a push for ES6 to spend time on this.

etpinard:
@archmoj ok to close this thing?

archmoj:
> To be honest @archmoj, I don't see how this makes our code more _readable_.
> 
> I agree our import/export statements could be standardised better, but I think we can wait when we'll make a push for ES6 to spend time on this.

Your call. 
Anyway here is one example if one searches for `isPlainObject` in https://github.com/plotly/plotly.js/blob/b6503ac2682bab35e7e6949696a7ce029cba00da/src/plot_api/plot_schema.js there are two ways which is called i.e. with/without `Lib.`.

etpinard:
Ok, I agree the patch you made to `plot_schema.js` is :ok_hand:

src/plot_api/helpers.js
```diff
@@ -12,15 +12,15 @@
 var isNumeric = require('fast-isnumeric');
 var m4FromQuat = require('gl-mat4/fromQuat');
 
-var Registry = require('../registry');
 var Lib = require('../lib');
-var Plots = require('../plots/plots');
-var AxisIds = require('../plots/cartesian/axis_ids');
+var subplotsRegistry = require('../plots/plots').subplotsRegistry;
```
archmoj: The only function used from `Plots` in this file is `subplotsRegistry`.

etpinard: Is there a way to enforce this _rule_ using eslint? If not, than we don't _need_ to do this.

Issue #3824, opened 2019-05-01
Using "update" method in button created using updatemenus removes the modebar


hashimmoosavi:
The issue can be reproduced on the Plotly website itself. Navigate to https://plot.ly/javascript/custom-buttons/#update-button

When you hover over the chart, you can view the modebar. 

<img width="773" alt="Screen Shot 2019-05-01 at 11 57 30 AM" src="https://user-images.githubusercontent.com/3678012/57011067-83e6f400-6c08-11e9-9930-eee87e681463.png">

When you click on any button, the modebar disappears.

<img width="817" alt="Screen Shot 2019-05-01 at 11 57 40 AM" src="https://user-images.githubusercontent.com/3678012/57011080-8d705c00-6c08-11e9-8112-a60355b5b8aa.png">

etpinard:
Thanks for reporting!

etpinard:
The report feels similar to https://github.com/plotly/plotly.js/issues/3809

This is probably a side-effect of #3589

sermeor:
I know the issue is closed, but it happens the same when using restyle method, which can be seen in the demos online (https://plotly.com/javascript/custom-buttons/). I was wondering if that can also be fixed?

PR #3825, opened 2019-05-01
Plotly.update replot fixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3824 

-----------

`Plotly.update` calls like

```js
Plotly.update(gd, {visible: false}, {annotations: [{text: 'a'}]});
```

that lead to restyle flag `fullReplot:true` and relayout flag `layoutReplot:true` erroneously resulted in a [`makePlotFramework`](https://github.com/plotly/plotly.js/blob/adf9bee14a8badcb5b2f49065fd594623421c82e/src/plot_api/plot_api.js#L3679-L3813) call, which (among other things) cleared the modebar container (added in https://github.com/plotly/plotly.js/pull/3589).

In brief, `makePlotFramework` on "empty" graph divs, I suspect other `Plotly.update` bugs went undetected because of this.

--------------

cc @plotly/plotly_js I'll ask @antoinerg to review this thing. Thanks!

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -2362,20 +2362,12 @@ function update(gd, traceUpdate, layoutUpdate, _traces) {
     // fill in redraw sequence
     var seq = [];
 
-    if(restyleFlags.fullReplot && relayoutFlags.layoutReplot) {
-        var data = gd.data;
-        var layout = gd.layout;
-
-        // clear existing data/layout on gd
-        // so that Plotly.plot doesn't try to extend them
-        gd.data = undefined;
-        gd.layout = undefined;
-
-        seq.push(function() { return exports.plot(gd, data, layout); });
+    if(relayoutFlags.layoutReplot) {
+        // N.B. works fine when both
+        // relayoutFlags.layoutReplot and restyleFlags.fullReplot are true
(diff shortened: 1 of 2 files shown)


src/plot_api/plot_api.js
```diff
@@ -2362,20 +2362,12 @@ function update(gd, traceUpdate, layoutUpdate, _traces) {
...
     var seq = [];
 
-    if(restyleFlags.fullReplot && relayoutFlags.layoutReplot) {
-        var data = gd.data;
-        var layout = gd.layout;
-
-        // clear existing data/layout on gd
-        // so that Plotly.plot doesn't try to extend them
-        gd.data = undefined;
-        gd.layout = undefined;
```
etpinard: **N.B.** This block was added in https://github.com/plotly/plotly.js/commit/9bad6082b5e1a91d43c17c6c8b12c0b683d71e2a - with proper test coverage.

antoinerg:
Codepen after the fix: https://codepen.io/antoinerg/pen/YbyevG

Looks good to me :dancer:

sermeor:
I know the issue is closed, but it happens the same when using **restyle method**, which can be seen in the demos online (https://plotly.com/javascript/custom-buttons/). I was wondering if that can also be fixed?

PR #3803, opened 2019-04-24
Shared color axes

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3431 - to be merged in https://github.com/plotly/plotly.js/pull/3786

@plotly/plotly_js please review this PR commit-by-commit. The "main" commits are:

- https://github.com/plotly/plotly.js/commit/ae3f407be227ca7e3fc4b20deb6775490b4583d5 where `layout.coloraxis?` attribute declarations and default logic are added.
- https://github.com/plotly/plotly.js/commit/5bb341ebbb0d80eae121d986627ae5823e3ff4b1 implements the cross-trace-defaults, calc and colorbar drawing.
- https://github.com/plotly/plotly.js/commit/bd81ac21ecedf1f4ca3d47d63671acc4920959f6 handles the `contour` trace edge cases.

PR #3827, opened 2019-05-01
cartesian 2dMap: implement sorting of categorical axes

### LINKED ISSUE #1097: heatmaps don't respect axis categoryarray
Hey Plotly,

It looks like you can use `axis.categoryarray` to reorder the bars in a bar chart. I'm trying to do the same thing with heatmaps (that is, reorder the heatmap columns) but it only seems to change the tick labels, not actually move the data. See http://codepen.io/hellochar/pen/WGBEow . This is concerning since the data being displayed is not longer accurate to the data that was fed in.

 Is there another way to specify the heatmap ordering, other than by modifying the x/y arrays manually?

### LINKED ISSUE #1117: Heatmap should look up category position when building brick data
The 2nd plot should start at `Team B`, instead of `Team A`.

```{python}
fig = make_subplots(rows=2, cols=1, shared_yaxes=True, shared_xaxes=True)

z = [[.1, .3, .5],  
     [1.0, .8, .6],
     [.6, .4, .2]]

x = ['Team A', 'Team B', 'Team C']
y = ['Game Three', 'Game Two', 'Game One']

fig.append_trace(go.Heatmap(z=z, x=x, y=y), 1, 1)

z = [[.3, .5],  
     [.8, .6],
     [.4, .2]]

x = ['Team B', 'Team C']
y = ['Game Three', 'Game Two', 'Game One']

fig.append_trace(go.Heatmap(z=z, x=x, y=y), 2, 1)
iplot(fig)
```

```
fig.data

[{'type': 'heatmap',
  'x': ['Team A', 'Team B', 'Team C'],
  'xaxis': 'x1',
  'y': ['Game Three', 'Game Two', 'Game One'],
  'z': [[.1, .3, .5],  
     [1.0, .8, .6],
     [.6, .4, .2]],
  'yaxis': 'y1'},
 {'type': 'heatmap',
  'x': ['Team B', 'Team C'],
  'xaxis': 'x1',
  'y': ['Game Three', 'Game Two', 'Game One'],
  'z': [[.3, .5],  
     [.8, .6],
     [.4, .2]],
  'yaxis': 'y2'}]

fig.layout

{'xaxis1': {'anchor': 'y2', 'domain': [0.0, 1.0]},
 'yaxis1': {'anchor': 'free', 'domain': [0.575, 1.0], 'position': 0.0},
 'yaxis2': {'anchor': 'x1', 'domain': [0.0, 0.425]}}
```
newplot 6

https://codepen.io/anon/pen/qqdqWq

antoinerg:
Closes #1097 (see below https://github.com/plotly/plotly.js/pull/3827#issuecomment-488405250)

Closes #1117 (see below https://github.com/plotly/plotly.js/pull/3827#issuecomment-489853085)


antoinerg:
Codepen before: https://codepen.io/antoinerg/pen/bJyOjX
Codepen after: https://codepen.io/antoinerg/pen/EJzGdv

src/traces/heatmap/clean_2d_array.js
```diff
@@ -30,12 +30,23 @@ module.exports = function clean2dArray(zOld, transpose) {
         old2new = function(zOld, i, j) { return zOld[i][j]; };
     }
 
+    var xMap = function(i) {return i;};
+    var yMap = function(i) {return i;};
+    if(trace && trace.type === 'heatmap') {
```
etpinard: Hmm. Why _just_ `heatmap`?

antoinerg: This was the low-hanging fruit. I guess I should enlarge the scope of this PR and implement sorting of categorical axes for `contour`, `histogram2d` and all other traces relying on `heatmap`'s calc routine.

etpinard: Yes please!

alexcjohnson: definitely useful for `histogram2d`! I have qualms from a viz-theory standpoint about sorting an axis of a contour plot: if you're drawing contours, which interpolate between data points, it strongly implies a continuous axis, so you really have no business with categories unless they're some sort of a hack of what's really a continuous axis... and sorting a continuous axis is meaningless.

That said, we allow it in the schema, therefore it's a bug if it doesn't work!

antoinerg: Histogram and contour are now working in 6b06516

By allowing all traces (except `carpet` and `contourcarpet`), the only test that would fail was in `bar_test.js`. In it, we would create a `contour` trace without `x` and `y` data. In that case, `trace._x` and `trace._y` would be empty and it failed. To fix this, we check whether those arrays are empty and ignore categorical sorting if they are. After all, if the data has no categories attached to it, it's not clear how it should be sorted.

src/traces/heatmap/clean_2d_array.js
```diff
@@ -10,7 +10,7 @@
 
 var isNumeric = require('fast-isnumeric');
 
-module.exports = function clean2dArray(zOld, transpose) {
+module.exports = function clean2dArray(zOld, transpose, trace, xa, ya) {
```
etpinard: Minor style tip: no need here to pass `transpose` and `trace` as `transpose = trace.transpose`. 

One example of a better call signature:

```js
function clean2dArray(zOld, trace, xa, ya) { /* */ }

File: src/traces/heatmap/calc.js [hunk 3]
```diff
@@ -64,14 +64,17 @@ module.exports = function calc(gd, trace) {
         } else {
             x = trace.x ? xa.makeCalcdata(trace, 'x') : [];
             y = trace.y ? ya.makeCalcdata(trace, 'y') : [];
+
+            trace._x = x;
+            trace._y = y;
```
etpinard: Minor style tip: we could merge this with the above statements and have:

```js
x = trace._x = trace.x ? xa.makeCalcdata(trace, 'x') : [];
// similarly for y
```

src/traces/heatmap/clean_2d_array.js
```diff
@@ -30,12 +30,23 @@ module.exports = function clean2dArray(zOld, transpose) {
         old2new = function(zOld, i, j) { return zOld[i][j]; };
     }
 
+    var xMap = function(i) {return i;};
```
etpinard: We should use `Lib.identity` here.

src/traces/heatmap/clean_2d_array.js
```diff
@@ -30,12 +30,23 @@ module.exports = function clean2dArray(zOld, transpose) {
...
+    var yMap = function(i) {return i;};
+    if(trace && trace.type !== 'carpet' && trace.type !== 'contourcarpet') {
+        if(ya && ya.type === 'category') {
+            if(trace._y.length) yMap = function(i) {return trace._y[i];};
+        }
```
etpinard: Minor style tip: the above three line could become:

```js
      if(ya && ya.type === 'category' && trace.y._length) {
          yMap = function(i) { return trace._y[i]; };
      }
```

etpinard:
Thanks very much @antoinerg . This PR is looking good!

etpinard:
Oh wait, codepen https://codepen.io/anon/pen/qqdqWq from https://github.com/plotly/plotly.js/issues/1117 still looks off using your latest commit -> https://codepen.io/etpinard/pen/ROXLbB

etpinard:
... it would be ok to leave #1117 un-addressed in this PR by the way. I hope to fix two bugs in one go, but it doesn't have to be the case :smile:

antoinerg:
@etpinard Although I can get `heatmapgl` to support `categoryorder` and `categoryarray` on first render, it doesn't work on `Plotly.relayout` (somehow the `calc` routine is not called again). Maybe this is related to https://github.com/plotly/plotly.js/issues/3833 which I just opened. :thinking:

etpinard:
> Although I can get `heatmapgl` to support `categoryorder` and `categoryarray` on first render, it doesn't work on `Plotly.relayout`

That's ok. If you're using `heatmapgl`, I hope that you have a large data set that isn't categorical, so I have 100% no problem calling this an edge case.

Thanks for opening https://github.com/plotly/plotly.js/issues/3833 !

src/traces/heatmap/clean_2d_array.js
antoinerg: Done in 59edb5704f076b438e6b8460b3e03097d61748ab

src/traces/heatmap/calc.js
antoinerg: Done in 59edb5704f076b438e6b8460b3e03097d61748ab

src/traces/heatmap/clean_2d_array.js
antoinerg: Done in 59edb5704f076b438e6b8460b3e03097d61748ab

src/traces/heatmap/clean_2d_array.js
antoinerg: Done in 59edb5704f076b438e6b8460b3e03097d61748ab

antoinerg:
@etpinard Ok so in the end, there are minimal changes to the code. Would you like to see more tests?

test/jasmine/tests/histogram2d_test.js
```diff
@@ -157,6 +170,60 @@ describe('Test histogram2d', function() {
                 [0, 0, 0, 1]
             ]);
         });
+
+        ['histogram2d', 'histogram2dcontour'].forEach(function(traceType) {
+            it('@flaky should sort z data based on axis categoryorder for ' + traceType, function() {
```
etpinard: @antoinerg do you have idea why this needs `@flaky` ? I mean `_module.calc` routines _can't_ be flaky right?

etpinard:
> Ok so in the end, there are minimal changes to the code. Would you like to see more tests?

Test coverage look good for this fix. Thanks very much! Although that I'm a little worried about that `@flaky` tag you added.

Now, could you investigate (i.e. no _need_ to fix this now) why you patch here does not resolve https://github.com/plotly/plotly.js/issues/1117 (which I also suspect will need a patch in `clean2darray`)? Thanks!

test/jasmine/tests/histogram2d_test.js
antoinerg: @etpinard I am puzzled by this. Locally the tests run fine but on the CI it would keep failing. Commit 22cfa82 (prior to adding flaky) just wouldn't pass `test-jasmine` on CircleCI and it failed with the same error 4-5 times in a row...

antoinerg: On CircleCI, it seems to consistently fail with:
Screenshot_2019-05-06_12-27-23
:man_shrugging:

antoinerg:
> Now, could you investigate (i.e. no _need_ to fix this now) why your patch here does not resolve #1117 (which I also suspect will need a patch in `clean2darray`)? Thanks!

Interestingly the current PR changes the result for the Codepen of issue #1117:
plotly.js from `master`: https://codepen.io/anon/pen/qqdqWq
plotly.js from this PR: https://codepen.io/antoinerg/pen/gJbWZd

I agree that #1117 is a nasty bug. I will investigate further.

antoinerg:
Commit e966030 closes #1117:
Codepen before: https://codepen.io/anon/pen/qqdqWq
Codepen after: https://codepen.io/antoinerg/pen/WBbyyP

antoinerg:
One detail I would like to discuss is whether we really want to support categories that are numerical. Right now, `gd._fullLayout.xaxis._categories` can be an array of integers. However, whenever `findCategories` gets called, it is turned into an array of strings by returning a call to `Object.keys`:
https://github.com/plotly/plotly.js/blob/03b7e09fa4fc6c842a28d8ba02547658cd43b6fe/src/plots/cartesian/category_order_defaults.js#L11-L42
This needs to be adapted if we want to support numerical categories.

test/image/mocks/heatmap_shared_categories.json
```diff
@@ -0,0 +1,49 @@
...
+    "xaxis": "x",
+    "y": ["Game Three", "Game Two", "Game One"],
+    "z": [
+      [0.3, 0.5],
+      [0.8, 0.6],
+      [0.4, 0.2]
+    ],
+    "yaxis": "y2"
+  }],
+  "layout": {
```
etpinard: Nice mock! Can you try using `coloraxis` from https://github.com/plotly/plotly.js/pull/3803 so that we don't have two overlapping (ugly) colorbars? Thanks!

etpinard:
> One detail I would like to discuss is whether we really want to support categories that are numerical. 

Yes.

Right now, `gd._fullLayout.xaxis._categories` can be an array of integers. However, whenever `findCategories` gets called, it is turned into an array of strings by returning a call to `Object.keys`

Why is that a big deal? I don't see a problem with tuning `typeof number` items in `_categories` into strings if that makes things easier.

src/traces/heatmap/clean_2d_array.js
```diff
@@ -30,12 +31,38 @@ module.exports = function clean2dArray(zOld, transpose) {
...
+    };
+
+    function axisMapping(ax) {
+        if(trace && trace.type !== 'carpet' && trace.type !== 'contourcarpet' &&
+            ax && ax.type === 'category' && trace['_' + ax._id.charAt(0)].length) {
+            var axMapping = [];
+            for(i = 0; i < ax._categories.length; i++) {
+                axMapping.push((trace['_' + ax._id.charAt(0) + 'Map'] || trace[ax._id.charAt(0)]).indexOf(ax._categories[i]));
+            }
+            return function(i) {return axMapping[i] === -1 ? undefined : axMapping[i];};
```
etpinard: You should be using `var BADNUM = require('../../constants/numerical').BADNUM` instead of `undefined` here.

src/traces/heatmap/clean_2d_array.js
```diff
@@ -30,12 +31,38 @@ module.exports = function clean2dArray(zOld, transpose) {
...
+        if(i === undefined || j === undefined) return undefined;
+        return old2new(zOld, i, j);
+    };
+
+    function axisMapping(ax) {
+        if(trace && trace.type !== 'carpet' && trace.type !== 'contourcarpet' &&
+            ax && ax.type === 'category' && trace['_' + ax._id.charAt(0)].length) {
+            var axMapping = [];
+            for(i = 0; i < ax._categories.length; i++) {
+                axMapping.push((trace['_' + ax._id.charAt(0) + 'Map'] || trace[ax._id.charAt(0)]).indexOf(ax._categories[i]));
```
etpinard: adding `var axLetter = ax._id.charAt(0)` above probably wouldn't hurt.

etpinard: Oh and can we use `ax._categoriesMap` instead of a (potentially) costly `indexOf` into `ax._categories`?

src/traces/heatmap/clean_2d_array.js
```diff
@@ -9,16 +9,17 @@
 'use strict';
 
 var isNumeric = require('fast-isnumeric');
+var Lib = require('../../lib');
 
-module.exports = function clean2dArray(zOld, transpose) {
+module.exports = function clean2dArray(zOld, trace, xa, ya) {
```
etpinard: I'm curious also, you have a bunch of `if(ax && ax[/* some key */])` statements in this routine. Is there a situation where `clearn2dArray` is called w/o `xa` and/or `ya`? 

If not, I'd just assume that they exist.

src/traces/heatmap/clean_2d_array.js
```diff
@@ -9,16 +9,17 @@
...
     var rowlen, collen, getCollen, old2new, i, j;
 
     function cleanZvalue(v) {
         if(!isNumeric(v)) return undefined;
         return +v;
     }
 
-    if(transpose) {
+    if(trace && trace.transpose) {
```
etpinard: Similarly for `trace`. Is there a situation where `clean2dArray` w/o `trace`?

src/traces/heatmap/clean_2d_array.js
antoinerg: > Oh and can we use `ax._categoriesMap` instead of a (potentially) costly `indexOf` into `ax._categories`?

Unfortunately, we cannot. For example, in the `heatmap_shared_categories` mock, `_categoriesMap` links category `Team C` with index 2. That index is out of range for the second trace since it only has 2 columns. We really need to check whether each trace has data for a given category (and fill with `BADNUM`) to fix #1117.

src/traces/heatmap/clean_2d_array.js
antoinerg: > Is there a situation where `clearn2dArray` is called w/o `xa` and/or `ya`?

I think it did occur yes. Let me confirm that for you!

src/traces/heatmap/clean_2d_array.js
antoinerg: > Why is that a big deal? I don't see a problem with turning `typeof number` items in `_categories` into strings if that makes things easier.

The problem is that we need a mechanism to check whether a trace has data associated with a given category. That is why we do an `indexOf` call. Of course, if the category and the input data have a different type, this lookup will not work properly.

> > One detail I would like to discuss is whether we really want to support categories that are numerical.
> 
> Yes.

Let me clarify my question. Should we consider `3` (a number) and `'3'` (a string) to represent different categories? If not, would it be okay to stringify everything?

etpinard: > Should we consider `3` (a number) and `'3'` (a string) to represent different categories?

yes -> https://codepen.io/etpinard/pen/NVqgdZ?editors=1010

antoinerg: Using `BADNUM` in ebf545ff33bd2f9fee38b516d73c891a2a2d1bee

src/traces/heatmap/clean_2d_array.js
antoinerg: > adding `var axLetter = ax._id.charAt(0)` above probably wouldn't hurt.

Done in ebf545ff33bd2f9fee38b516d73c891a2a2d1bee

test/image/mocks/heatmap_shared_categories.json
antoinerg: Done in 793c2a346e07d4dfa5504a20dd95d6fa47290ff0

src/traces/heatmap/clean_2d_array.js
antoinerg: > Similarly for `trace`. Is there a situation where `clean2dArray` w/o `trace`?

Ok, probably not :sweat_smile: . Reverted in ebf545ff33bd2f9fee38b516d73c891a2a2d1bee

src/traces/heatmap/clean_2d_array.js
antoinerg: > > Should we consider `3` (a number) and `'3'` (a string) to represent different categories?
> 
> yes -> https://codepen.io/etpinard/pen/NVqgdZ?editors=1010

I'm confused. It seems to me that this Codepen show that  `1` and `'1'` are considered to be the **same** category.

etpinard: > It seems to me that this Codepen shows that `1` and `'1'` are considered to be the **same** category.

Yes, that's what I meant. Sorry. Edited above.

src/traces/heatmap/clean_2d_array.js
antoinerg: Turns out this was necessary:
Screenshot_2019-05-07_11-50-53

src/traces/heatmap/clean_2d_array.js
antoinerg: > Ok, that's very unfortunate, because
> 
> ```js
> for(i = 0; i < ax._categories.length; i++) {
>   trace[ax._id.charAt(0)]).indexOf(ax._categories[i]))
> }
> ```
> 
> can be very costly.

What we could do instead is build a `_traceCategoriesMap` for each trace. It would map categories to their position within the trace-level `z` data. This might be a better idea now that I think about it.

test/jasmine/tests/histogram2d_test.js
antoinerg: I could remove the "flaky" tag with commit 4d4b0b748d71fb6ef272369de06f4d00c6e45e43 ! :tada:

src/traces/heatmap/clean_2d_array.js
antoinerg: @etpinard I think can remove `indexOf` with the following:
```javascript
for(i = 0; i < trace[axLetter].length; i++) {
    axMapping[trace[axLetter][i]] = i;
}
return function(i) {return axMapping.hasOwnProperty(ax._categories[i]) ? axMapping[ax._categories[i]] : BADNUM;};
```
Would this be good enough?

antoinerg: I went ahead and committed the suggested change above in f180c79 !

antoinerg:
@etpinard I updated the various Codepens to use the latest version f180c79!

src/traces/heatmap/clean_2d_array.js
etpinard: I didn't know you could use `hasOwnProperty` on an array. Two concerns:

- Is that support across browsers? 
- is that really faster than `indexOf`?

etpinard: What if instead you made `_(x|y)CategoryMap` lookup objects? 

That is, instead of:

```js
   if(ax1 && ax1.type === 'category') {
        trace['_' + var1Name + 'CategoryMap'] = col1vals.map(function(v) { return ax1._categories[v];});
    }
```

we could have

```js
   if(ax1 && ax1.type === 'category') {
        var lookup = trace['_' + var1Name + 'CategoryMap'] = {};
        for(var i = 0; i < col1vals.length; i++) {
           var v = col1vals[i];
           lookup[v] = i;
       }
    }
```

antoinerg: About your questions regarding `hasOwnProperty`:
- we have it in several places already and it seems [well supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#Browser_compatibility)
- `hasOwnProperty` is roughly 6 times faster than `indexOf`

However, we now run it for every element so in the end, I'm not sure which is faster.

etpinard: Sorry @antoinerg I got confused (again). Your logic looks right to me. 

The way you have it setup right now, I _think_ we can get rid of the `axMapping.hasOwnProperty(ax._categories[i])` and just have:

```js
function(i) { 
  var ind = axMapping[ax._categories[i]];
  return ind + 1 ? ind : BADNUM;
};
```

where the `+ 1` takes care of the `ind === 0` case. 

This comment is non-blocking as I don't expect much of a performance gain from it :smirk:

etpinard:
Nice work @antoinerg !

Thanks for commit https://github.com/plotly/plotly.js/pull/3827/commits/0c41776c7dcdf30eb596d5fcd5cc64d64d99087b - which should be a step in the right direction towards cleaning up our categories stashes (we still have a long way to go though).

The best part about this PR are the tests, so let's get in :dancer:

--------

I'm not sure if GitHub is smart enough to close multiple issues at once, so make sure both https://github.com/plotly/plotly.js/issues/1097 and https://github.com/plotly/plotly.js/issues/1117 are closed off correctly.

src/traces/heatmap/clean_2d_array.js
antoinerg: > where the `+ 1` takes care of the `ind === 0` case.

Awesome trick :) Done in c7cd1f3

etpinard: :dancer: :dancer:

Issue #3836, opened 2019-05-06
Box legend marker with invisible box but visible points


nicolaskruchten:
If I set the box fill and line to 0 opacity but have visible points, the legend entry is invisible. In this case it would be nice to have the legend entry look like the points.

nicolaskruchten:
For context, I'm trying to add strip plots to `px`:

![image](https://user-images.githubusercontent.com/203523/57203810-1a0c7880-6f81-11e9-90c9-e549c7fd0403.png)

etpinard:
Should we add "type: 'strip'" traces to plotly.js?

nicolaskruchten:
could do, but it's really really easy to emulate with `box` (or even `violin`) for now so it's really not a priority, so long as we can slide in a fix for the legend :)

alexcjohnson:
Sure, get a quick legend fix in, but `type: 'strip'` would be nice at some point - feels awfully hacky to do it via box or violin (instructions: make a box with no box, or a violin with no violin. Take your pick!) and there may be some extra functionality that doesn't make sense in the existing types.

etpinard:
> , but `type: 'strip'` would be nice at some point -

That or adding `jitter` to scatter traces (cc https://github.com/plotly/plotly.js/issues/299), Personally, I prefer adding a `type: 'strip'` trace type.

alexcjohnson:
Right, `scatter.jitter` is related, but not a replacement for `type: 'strip'` - especially grouping & offsets on the position axis, but also it's not clear to me if we can (or even if we want to) adjust scatter jitter for the local density of points like we do for box/violin jitter.

nicolaskruchten:
Glad you guys are into it! I'm not opposed to a `strip` type, but I'd still like a fix for the legend in the meantime :)

etpinard:
> but I'd still like a fix for the legend in the meantime :)

Yep, that fix we'll be in 1.48.0

nicolaskruchten:
> ... adjust scatter jitter for the local density of points like we do for box/violin jitter.

I didn't realize we did that but looking at the screenshot above we clearly do, which is cool. It's sort of halfway to "swarm plots"

Seaborn has both but they're quite distinct: https://seaborn.pydata.org/generated/seaborn.swarmplot.html and https://seaborn.pydata.org/generated/seaborn.stripplot.html

nicolaskruchten:
Could we imagine an attr that would emulate swarm plots more directly?

[...]

PR #3846, opened 2019-05-08
Draw box/violin pts as legend item when other parts have opacity 0

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3836

before: https://codepen.io/etpinard/pen/pmjGEL?editors=0010
after: https://codepen.io/MojtabaSamimi/pen/JqYQjG?editors=0010 


src/components/legend/style.js
```diff
@@ -356,12 +356,27 @@ module.exports = function style(s, gd) {
 
         pts.each(function() {
             var p = d3.select(this);
-            var w = boundLineWidth(undefined, trace.line, MAX_MARKER_LINE_WIDTH, CST_MARKER_LINE_WIDTH);
 
-            p.style('stroke-width', w + 'px')
-                .call(Color.fill, trace.fillcolor);
+            if((trace.boxpoints === 'all' || trace.points === 'all') &&
```
etpinard: @nicolaskruchten do you use other `(box)points` values for your `px` strip traces?

If so, we might need to add logic for `marker.outliercolor`, `marker.line.outliercolor` and `marker.line.outlierwidth` in the block below.

alexcjohnson: might be interesting to add a mode where we show all points but style [suspected]outliers differently - but currently these other modes both hide the non-outlier points entirely - which seems like an awfully bad idea to combine with hiding the box/violin!

nicolaskruchten: for `px` I was planning on using `'all'`, because of @alexcjohnson's comment :) so this is looking great to me :)

archmoj:
[Codepen After](https://codepen.io/MojtabaSamimi/pen/JqYQjG?editors=0010).

antoinerg:
Looks good to me :dancer:

PR #3848, opened 2019-05-09
Display bar outlines with height zero

### LINKED ISSUE #3797: Bar chart: draw the outline, even if the height is zero.
For a bar of zero height, if the line width is specified, we should be able to see the bar border, even though we don't see the bar.

In this example:
https://codepen.io/RichardNeill/pen/EJRqmN

B is a normal bar, with 10px border, as it should be.
C is a bar with height = null.  It is, correctly, completely invisible.

But A is a bar with height = 0. It should actually appear as a 20px high rectangle (the border should still be displayed), even though the bar area itself isn't.  Try setting the value to 0.0001
and you'll see what I expect. In addition, the hover tooltip is (wrongly) suppressed - it doesn't appear in "show closest data on hover" mode.

archmoj:
Fix #3797 i.e. for bars as well as histograms.
Note: a similar fix was not desirable for waterfalls/funnels.
And the text is not printed when the bar height is zero.

@plotly/plotly_js


src/traces/bar/plot.js
```diff
@@ -231,7 +235,7 @@ function appendBarText(gd, plotinfo, bar, calcTrace, i, x0, x1, y0, y1, opts) {
 
     if(!text ||
         textPosition === 'none' ||
-        (calcBar.isBlank && (
+        ((calcBar.isBlank || x0 === x1 || y0 === y1) && (
```
etpinard: So here we don't show text associated with zero-height bars?

etpinard: Oh sorry, we don't show **inside** text associated with zero-height bars. Nicely done :ok_hand:

src/traces/bar/plot.js
```diff
@@ -67,9 +67,12 @@ module.exports = function plot(gd, plotinfo, cdModule, traceLayer, opts) {
     var bartraces = Lib.makeTraceGroups(traceLayer, cdModule, 'trace bars').each(function(cd) {
         var plotGroup = d3.select(this);
         var trace = cd[0].trace;
+        var isWaterfall = (trace.type === 'waterfall');
+        var isFunnel = (trace.type === 'funnel');
+        var isBar = !(isWaterfall || isFunnel);
```
etpinard: Ok, so `isBar` includes histogram traces. This looks ok, as I don't think you can get a zero-height bar in histogram to go through here. But I think it would be safer to that have this as `trace.type === 'bar'`.

archmoj: Good call. 
Done in 97190c3.

archmoj:
TODO: we need to display `funnel` bar outlines with zero height.

etpinard:
@archmoj reading up https://github.com/plotly/plotly.js/issues/3797#issuecomment-485415054 again, in particular:

> Also re: hover - makes sense to include the lines, though this gets tricky if there’s overlap like stacking.

we should add a hover test, locking down the behaviour.

etpinard:
> we should add a hover test, locking down the behaviour.

done in https://github.com/plotly/plotly.js/pull/3848/commits/7224fe4d137670207b865493639f28795a8d29e3

:dancer: from me.

Feel free to merge this thing!

archmoj:
@etpinard thanks very much for the help on this.
Merging.

PR #3810, opened 2019-04-25
scattergl selection fixes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3740, https://github.com/plotly/plotly.js/issues/2958

before (context in #3740): https://codepen.io/Convo/pen/GLZYdw
after: https://codepen.io/etpinard/pen/xemRqV

@plotly/plotly_js see commit msg in https://github.com/plotly/plotly.js/commit/81367f13e3ee5bb6abfeda5aa12d36c182fac466 for more info.

PR #3853, opened 2019-05-10
Fixup jasmine2 to have more consistent CI runs

archmoj:
Recently the `Jasmine2` container started to constantly failling on the `CircleCI`. It is also take a long time to fail i.e. about 15-20min.
This PR rearranged some of the tests in the `gl2d_double_click_test` simply by moving shared functions to the top, avoid nested `describe` blocks. 
A new flag titled `dbgl` is added for a new container to isolate flaky `gl` tests. 
Parallelism of `test-jasmine2` increased from 2 to 3 so that the tests complete around the same time i.e. close to 5 min. 
Finally the shard limit reduced to 1 in order to help reduce the chance of possible side effects of running one test on the other.

@plotly/plotly_js

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
     docker:
       # need '-browsers' version to test in real (xvfb-wrapped) browsers
       - image: circleci/node:10.9.0-browsers
-    parallelism: 2
+    parallelism: 5
     working_directory: ~/plotly.js
     steps:
       - attach_workspace:
@@ -77,7 +77,7 @@ jobs:
   test-image:
     docker:
       - image: plotly/testbed:latest
-    parallelism: 4
+    parallelism: 3
     working_directory: /var/www/streambed/image_server/plotly.js/
     steps:
       - attach_workspace:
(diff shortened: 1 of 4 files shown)


etpinard:
Thanks for taking a look into this @archmoj 

You're right that `test-jasmine2` has been very bad lately (after since we merged https://github.com/plotly/plotly.js/pull/3810)

Now, I can't say I'm a fan of this solution. I don't think we should be adding another container (that has a cost :moneybag: you know) and I don't think we should be adding another `@tag` (which has a maintenance cost).

Here are some ideas:

- Have you tried splitting up the `@gl` test into more `_test.js` files?
- There must be something in our `@gl` that's causing the flakiness? We should try:
  + Adding longer delays between drag commands
  + Adding delays between each `it` - to allow the container to (hopefully) slow down its CPU usage

archmoj:
> * Have you tried splitting up the `@gl` test into more `_test.js` files?

It is a very good idea. Namely the `gl2d_double_click_test.js`.

archmoj:
Closing in the favour of #3859.

PR #3859, opened 2019-05-13
Reorganise gl tests for less flakiness

etpinard:
my follow-up on https://github.com/plotly/plotly.js/pull/3853#issuecomment-491294321- which appears to make our flaky `@gl` tests less flaky (3/3 successful runs so far).

In brief,

- The hover/click tests in `gl3d_plot_interact_test.js` are moved to new suite `gl3d_hover_click_test.js`
- `scattergl` interactions tests are divided up evenly (and logically) in `gl2d_plot_interact_test.js`, `scattergl_test.js` and `scattergl_select_test.js` 

cc @plotly/plotly_js It's the not the end of the story, but this PR should be a solid step forward toward non-flaky `@gl` tests.

PR #3845, opened 2019-05-08
Introducing histogram* bingroup attributes

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3749 (the 1d overlay case) and resolves https://github.com/plotly/plotly.js/issues/3515 (the 2d case).

Users can now group traces such that their autobin settings are _compatible_ (see https://github.com/plotly/plotly.js/issues/3515#issuecomment-460706202 for more info on what _compatible_ means).

Examples:
- https://codepen.io/etpinard/pen/YbyOKO?editors=0010 (1d overlay)
- https://codepen.io/etpinard/pen/arvazV (2d)

In brief, this PR:

- adds attribute `bingroup` to `histogram`, `histogram2d` and `histogram2dcontour` traces
- `histogram2d` and `histogram2dcontour` traces also get new attributes `xbingroup` and `ybingroup` to allow users to group x-bins and y-bins separately.

To do so, I decided to:

- make all histogram* traces (i.e. 1d and 2d) go through the same `crossTraceDefaults` method
- make all histogram* traces use that same `calcAllAutoBins` routine

@plotly/plotly_js best to review this PR commit-by-commit. Starting by reviewing the jasmine tests added in https://github.com/plotly/plotly.js/commit/359fbbc9267fa2f241062dc63cbadb4d1130b3d2 might make things easier to grasp.

PR #3864, opened 2019-05-14
sort categorical Cartesian axes by value

### LINKED ISSUE #3606: value ascending / value descending
I'm making categorical histograms (with and without `z` and `histfunc: sum`). I'd like to be able to sort these histograms in descending by value order without using transforms.

Maybe this would fit under `categoryorder: 'value ascending'` and `categoryorder: 'value descending'`? https://github.com/plotly/plotly.js/pull/419/files#diff-0e41c2e162564438ff091d0ed6b5b455R472

antoinerg:
Closes #3606 by introducing the possibility to sort by values using `categoryorder`.

Most of the work involved writing tests to make sure everything works properly across all traces. Note that `scattergl` and `splom` required special handling since they differ from other traces.

Supported traces (:lock: down with tests of course):
- [x] scatter
- [x] bar
- [x] box
- [x] heatmap
- [x] histogram
- [x] histogram2d
- [x] contour
- [x] violin
- [x] funnel
- [x] waterfall
- [x] scattergl
- [x] splom
- [x] ohlc
- [x] candlestick
- [x] histogram2dcontour

Supported aggregations:
- [x] min
- [x] max
- [x] sum
- [x] mean (unweighted across all values)
- [x] median (unweighted across all values)

I chose aggregations (`min`, `max` and `sum`) because they are unambiguously defined regardless of the order of operation (the minimum of the traces' minimum is equal to the minimum of all values across all traces). Other operations such as "take the mean" are ambiguous: is it the mean of all values across all traces or the mean of each traces' means (at the moment we assume the former for mean/median). To be clear, we need the ability to specify how to aggregate at each step:
1. how to aggregate multiple values belonging to a trace
2. how to aggregate the aggregated value obtained in step 1

This will probably require making a new attribute. Do you have any suggestions for this? Related https://github.com/plotly/plotly.js/issues/3606#issuecomment-492694202

Coverage:
- [x] support sorting matching axes
- [x] support sorting of  both x and y axes

TODO:
- [x] splom: get rid `indexOf` calls

FUTURE WORK:
- move sorting logic in category_order_default.js to `Plots.doCalcData` (https://github.com/plotly/plotly.js/pull/3864#issuecomment-492707571)

cc @nicolaskruchten @plotly/plotly_js


antoinerg:
For reasons unknown, running the tests on CircleCI today resulted a ridiculously low success rate :disappointed:

etpinard:
> , running the tests on CircleCI today resulted in a ridiculously low success rate

Have you rebased off master since https://github.com/plotly/plotly.js/pull/3859 got merged?

antoinerg:
> > , running the tests on CircleCI today resulted in a ridiculously low success rate
> 
> Have you rebased off master since #3859 got merged?

Yes I did!

etpinard:
> Yes I did!

Cool. Looks like `test-jasmine` is responsible for most of the failures today. Maybe we should tag a few more of those tests as `@flaky`.

src/plots/plots.js
```diff
@@ -2843,10 +2843,176 @@ plots.doCalcdata = function(gd, traces) {
...
+
+function sortAxisCategoriesByValue(axList, gd) {
+    var affectedTraces = [];
+    var i, j, k, l, o;
+    for(i = 0; i < axList.length; i++) {
+        var ax = axList[i];
+        if(ax.type !== 'category') continue;
+
+        // Order by value
+        var match = ax.categoryorder.match(sortAxisCategoriesByValueRegex);
```
etpinard: In situations like this one, we usually prefer using `sortAxisCategoriesByValueRegex.test(ax.categoryorder)` which always returns a boolean.

src/plots/plots.js
```diff
@@ -2843,10 +2843,176 @@ plots.doCalcdata = function(gd, traces) {
...
+            // Store values associated with each category
+            var categoriesValue = [];
+            for(j = 0; j < ax._categories.length; j++) {
+                categoriesValue.push([ax._categories[j], []]);
+            }
+
+            // Collect values across traces
+            for(j = 0; j < ax._traceIndices.length; j++) {
+                var traceIndex = ax._traceIndices[j];
+                var fullData = gd._fullData[traceIndex];
```
etpinard: We usually call items in `gd._fullData` `fullTrace` or simply `trace` when unambiguous with items in `gd.data`

src/plots/plots.js
```diff
@@ -2843,10 +2843,176 @@ plots.doCalcdata = function(gd, traces) {
...
+
+                // Skip over invisible traces
+                if(fullData.visible !== true) continue;
+
+                var type = fullData.type;
+                if(type === 'histogram') delete fullData._autoBinFinished;
```
etpinard: Post https://github.com/plotly/plotly.js/pull/3845, we'll need this to be `Registry.traceIs(trace, 'histogram')`. Would you mind changing this line to 

```js
if(Registry.traceIs(trace, 'histogram')) delete fullData._autoBinFinished;
```

thanks!

src/plots/plots.js
```diff
@@ -2843,10 +2843,176 @@ plots.doCalcdata = function(gd, traces) {
...
+
+                var cd = gd.calcdata[traceIndex];
+                for(k = 0; k < cd.length; k++) {
+                    var cdi = cd[k];
+                    var cat, catIndex, value;
+
+                    // If `splom`, collect values across dimensions
+                    if(type === 'splom') {
+                        // Find which dimension the current axis is representing
+                        var currentDimensionIndex = cdi.trace[ax._id.charAt(0) + 'axes'].indexOf(ax._id);
```
etpinard: Try using the lookups `_xaxes`  and `_yaxes` defined here in ->

https://github.com/plotly/plotly.js/blob/06ca2c125e2411e0fbafc09d203d6821ab551c46/src/traces/splom/defaults.js#L96-L97

to avoid using an `.indexOf`.

src/plots/plots.js
```diff
@@ -2843,10 +2843,176 @@ plots.doCalcdata = function(gd, traces) {
...
+
+                        // Apply logic to associated x axis
+                        if(ax._id.charAt(0) === 'y') {
```
etpinard: You're using `ax._id.charAt(0)` a lot, we usually call this `axLetter`.

src/traces/bar/calc.js
```diff
@@ -33,7 +33,7 @@ module.exports = function calc(gd, trace) {
 
     // set position and size
     for(var i = 0; i < serieslen; i++) {
-        cd[i] = { p: pos[i], s: size[i] };
+        cd[i] = { p: pos[i], s: size[i], v: size[i] };
```
etpinard: From the 

```js
 value = cdi.s || cdi.v || cdi.y;
```

line in `sortAxisCategoriesByValue`, looks like you no longer need to set `v`.

src/traces/ohlc/calc.js
```diff
@@ -86,6 +86,10 @@ function calcCommon(gd, trace, x, ya, ptFunc) {
             pt.i = i;
             pt.dir = increasing ? 'increasing' : 'decreasing';
 
+            // For categoryorder, store low and high
+            pt.x = pt.pos;
+            pt.y = [li, hi];
```
etpinard: Clever. No need to store open, high, low and close values :+1:

test/image/mocks/sort_by_value_matching_axes.json
```diff
@@ -0,0 +1,43 @@
+{
+  "data": [{
+    "type": "scatter",
+    "orientation": "v",
+    "x": ["a", "b", "c"],
+    "y": [7, 2, 3],
+
+    "z": [
```
etpinard: We can :hocho: `z` and `dimensions` out of this mock

test/image/mocks/sort_by_value_matching_axes.json
```diff
@@ -0,0 +1,43 @@
...
+      "label": "DimensionA",
+      "values": ["a", "b", "c"]
+    }, {
+      "label": "DimensionB",
+      "values": [7, 2, 3]
+    }]
+  }, {
+    "type": "bar",
+    "x": ["a", "b", "c"],
+    "y": [10, 20, 30],
```
etpinard: Hmm. I don't understand why the categories show up as `[b, c, a]`. Shouldn't the `y` values for the bar trace by considered also?

test/jasmine/tests/calcdata_test.js
```diff
@@ -869,6 +871,216 @@ describe('calculated data and points', function() {
...
+                        {
+                            label: 'DimensionB',
+                            values: b
+                        }
+                    ]
+                });
+            }
+
+            supportedCartesianTraces.forEach(function(trace) {
+                ['xaxis', 'yaxis'].forEach(function(axId) {
```
etpinard: `axId` is usually reserved for `x2`, `y4`, etc. For `xaxis` and `yaxis` we usually use `axName`

test/jasmine/tests/calcdata_test.js
```diff
@@ -869,6 +871,216 @@ describe('calculated data and points', function() {
...
+                            var data = [7, 2, 3];
+                            var baseMock = { data: [makeData(trace.type, cat, data, axId)], layout: {}};
+                            var mock = Lib.extendDeep({}, baseMock);
+                            mock.layout[axId] = { type: 'category', categoryorder: categoryorder};
+
+                            // Set ordering
+                            var finalOrder = ['b', 'c', 'a'];
+                            if(categoryorder === 'value descending') finalOrder.reverse();
+
+                            if(trace.type.match(/histogram/)) {
```
etpinard: Could we reuse `checkAggregatedValue` here?

etpinard:
Nicely done @antoinerg ! Test coverage looks impressive. 

I would be _ok_ leaving `histogram2dcontour` support out of this PR if it proves too tedious to add. 

What I would like to see though: the sorting logic in https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/category_order_defaults.js getting moved to `Plots.doCalcdata` and possibly merged into your `sortAxisCategoriesByValue` routine. That way all things `categoryorder` will be handled in the same place, only during `calc` edits.

src/plots/plots.js
antoinerg: Good to know! In this case, however, I need to know the matches (is it ascending or descending).

etpinard: > I need to know the matches

Ha right, I missed that. Resolved.

test/image/mocks/sort_by_value_matching_axes.json
antoinerg: > Hmm. I don't understand why the categories show up as `[b, c, a]`. Shouldn't the `y` values for the bar trace by considered also?

At the moment, not if they are on a different axis: only the traces associated with the axis to be sorted are accounted for. In this mock, the `bar` trace is on different axes namely `(x|y)axis2`. It ends inheriting the order of categories of `xaxis` because it `matches` it but that's all. I thought this behavior would be useful.

We could account for the values of all traces across all matching axes IF we also support a way to specify traces to ignore as suggested @nicolaskruchten (see https://github.com/plotly/plotly.js/issues/3606#issuecomment-492694202) so that we can reproduce the current baseline.

etpinard: > not if they are on a different axis: only the traces associated with the axis to be sorted are accounted for. In this mock, the `bar` trace is on different axes namely `(x|y)axis2`. It ends inheriting the order of categories of `xaxis` because it `matches` it but that's all. I thought this behavior would be useful.

Thanks for the explanation. That makes perfect sense :+1:

src/plots/plots.js
antoinerg: Replaced with `fullData` by `fullTrace` in 2d379f7647bec909b806f56276b6f6225578bc6b

src/plots/plots.js
antoinerg: Done in 2d379f7647bec909b806f56276b6f6225578bc6b

src/plots/plots.js
antoinerg: Replaced `ax._id.charAt(0)` by `axLetter` in 2d379f7647bec909b806f56276b6f6225578bc6b

src/traces/bar/calc.js
antoinerg: You're right! Nice catch :eyes: 

Removed in 2d379f7647bec909b806f56276b6f6225578bc6b

test/image/mocks/sort_by_value_matching_axes.json
antoinerg: Done in 2d379f7647bec909b806f56276b6f6225578bc6b

test/jasmine/tests/calcdata_test.js
antoinerg: Much clearer thanks! Done in 2d379f7647bec909b806f56276b6f6225578bc6b

test/jasmine/tests/calcdata_test.js
antoinerg: Absolutely! I also realized that very first test was very similar to a subsequent one so I :hocho: it.

See for 2d379f7647bec909b806f56276b6f6225578bc6b for the changes!

antoinerg:
> I would be _ok_ leaving `histogram2dcontour` support out of this PR if it proves too tedious to add.

I ended up adding support for `histogram2dcontour` in 76683ba !

src/plots/plots.js
```diff
@@ -2861,6 +2861,31 @@ var sortAxisCategoriesByValueRegex = /(value|sum|min|max) (ascending|descending)
 function sortAxisCategoriesByValue(axList, gd) {
     var affectedTraces = [];
     var i, j, k, l, o;
+
+    function zMapCategory(type, ax, value) {
+        var axLetter = ax._id.charAt(0);
+        if(type === 'histogram2dcontour') {
```
etpinard: Interesting. Could you explain why we need special logic for `histogram2dcontour` but not `histogram2d`?

src/plots/plots.js
antoinerg: I got rid of `indexOf` calls for `splom` in commit fac239b7b8db0ac7d9f68850c75a9bca1c03a4c6

src/plots/plots.js
antoinerg: For `histogram2d`: `gd.calcdata[0][0].z` will have dimensions of `x.length` and `y.length`.

For `histogram2dcontour`: `gd.calcdata[0][0].z` will have dimensions of `x.length + 1` and `y.length + 1`. If the `x` or `y` are on categorical axes, the dimensions will be `(x|y).length + 2`. Not sure why that is :man_shrugging:

etpinard: Thanks for the info!

src/plots/plots.js
etpinard: Excellent!

src/plots/plots.js
```diff
@@ -2886,13 +2885,25 @@ function sortAxisCategoriesByValue(axList, gd) {
         }
     }
 
+    var aggFn = {
+        'min': function(values) {return Lib.aggNums(Math.min, null, values);},
+        'max': function(values) {return Lib.aggNums(Math.max, null, values);},
+        'sum': function(values) {return Lib.aggNums(function(a, b) { return a + b;}, null, values);},
+        'value': function(values) {return Lib.aggNums(function(a, b) { return a + b;}, null, values);},
```
etpinard: Hmm. Written in this way, I'm thinking maybe _total_ (as in `categoryorder: 'total ascending'`) would make more sense. Thoughts?

src/plots/plots.js
```diff
@@ -2886,13 +2885,25 @@ function sortAxisCategoriesByValue(axList, gd) {
...
+        'mean': function(values) {return Lib.mean(values);},
+        'median': function(values) {values.sort(); var mid = Math.round((values.length - 1) / 2); return values[mid];}
```
etpinard: I'm surprised our linter didn't complain here. Would you mind rewriting this as

```js
'median': function(values) {
    values.sort(); 
    var mid = Math.round((values.length - 1) / 2); 
    return values[mid];
}
```

etpinard: There's probably an eslint [rule](https://eslint.org/docs/rules/) for this :microscope:

src/lib/stats.js
```diff
@@ -74,6 +74,22 @@ exports.stdev = function(data, len, mean) {
     return Math.sqrt(exports.variance(data, len, mean));
 };
 
+/**
+ * median of a finite set of numbers
+ * reference page: https://en.wikipedia.org/wiki/Median#Finite_set_of_numbers
+**/
+exports.median = function(data, len) {
```
etpinard: Isn't this the same as `Lib.interp(data, 0.5)` ?

src/plots/plots.js
antoinerg: I agree that `value` is not a good name. I think the vision was to have `value` switch the ordering mode similar to what `array` does (with `array`, we look for the order in `categoryarray`). I think we should reserve `value` for this purpose. I will rename the current `value` to `total`.

src/lib/stats.js
antoinerg: `Lib.interp` is similar except that it doesn't sort the data. Would you prefer that I rewrite `median` to:
```javascript
exports.median = function(data) {
    var b = data.slice().sort();
    retunr exports.interp(b, 0.5);
};

Comment (etpinard) [reply src/lib/stats.js]: Ha. Good eye! You can ignore my comment. Feel free to rewrite `Lib.median` as

```js
exports.median = function(data) {
    var b = data.slice().sort();
    retunr exports.interp(b, 0.5);
};
```

or not.

antoinerg: I rewrote it as suggested and tests `Lib.median` against non-sorted values to :lock: down its behavior in 54b3ff93d95eef21660e66c49cde1f50b59b31a2

src/plots/plots.js
antoinerg: Done in 9aaa10b59a62cd4595f665ec2860839384d4e047

antoinerg:
As discussed in the original issue https://github.com/plotly/plotly.js/issues/3606#issuecomment-492694202, in the future, we will need to add new attributes to support different ways of sorting categories. @etpinard suggested that we create a new container to group all those attributes. I think it would be a good idea.

Right now, everything is stuffed into `categoryorder` such as `category ascending`, `sum descending`, etc... Instead of enumerating all the possible combinations, those two parameters (example `category` and `ascending`) could/should be specified via their own separate attribute. Whenever we split those two, we should also add additional attributes to refine sorting by value as discussed in https://github.com/plotly/plotly.js/issues/3606#issuecomment-492694202 and following comments. We could still support our current `categoryorder` by mapping them to an equivalent set of attributes in the new container. 

Please let me know if you see any problems with this approach and whether the current state of this PR can be merged as is.

etpinard:
My take on https://github.com/plotly/plotly.js/pull/3864#issuecomment-493574928 :

Our `category*` attributes are already kind of a mess, moving towards a `axis.category.*` attribute container sounds inevitable. Per https://github.com/plotly/plotly.js/issues/3606#issuecomment-492694202, it looks like we haven't spec'ed out all the potential category attributes yet, so I think _just_ filling in `categoryorder` is way to go for now. At some point, we can introduce `axis.category.*` and convert "old" `categoryorder` and `catergoyarray` during `cleanData`.

-------

All in all, this PR LGTM. Well-earned :dancer:

PR #3865, opened 2019-05-14
Trace meta text

etpinard:
a follow-up https://github.com/plotly/plotly.js/pull/3439 and https://github.com/plotly/plotly.js/pull/3548 - where we: 

- allow `meta` to be set as an `{}` (not just as an `[]`) and 
- allow users to set `meta` info in each trace container.

--------

To make it easier for users to grab trace `meta` info within the trace where `meta` is set, we proposed the following precedence rules:

Given

```js
Plotly.newPlot(gd, [{
  y: [1, 2, 1],
  meta: { yname: "dollars" }
},{
  y: [2, 3, 4],
  marker: {color: [1, 2, 3]},
  meta: { yname: "temperature" }
}], {
  meta: [
    'Company B'
  ]
})
```

Setting

```js
gd.data[0].name = '%{meta.yname}' // => 'dollars'
gd.data[1].marker.colorbar.title.text = '%{meta.yname}' // => 'temperature'
```

but `%{meta.yname}` gives a blank string in layout fields (e.g. `layout.xaxis.title.text`).

One could still use `layout.meta` within the traces as

```js
gd.data[0].name = '%{meta[0]}' // =>  'Company B' (as key '0' is not overridden by trace.meta)

// or 
gd.data[0].name = '%{layout.meta[0]}' // => 'Company B'
```

Moreover, one can target trace `meta` info in layout fields as

```js
gd.layout.yaxis.title.text = '%{data[0].meta.yname} vs. %{data[1].meta.yname}'
```

--- 

I hope @plotly/plotly_js will like it.


nicolaskruchten:
This looks good to me. I had forgotten that in 1.47, an un-prefixed `trace.name = '%{meta[0]}'` will resolve to `layout.meta[0]`. I think it's perhaps not too late to break that and require the `layout.` prefix always? 

Either way, this PR looks like we need, thanks!

etpinard:
> This looks good to me. I had forgotten that in 1.47, an un-prefixed `trace.name = '%{meta[0]}'` will resolve to `layout.meta[0]`.

That will still work. The `meta` ref for trace is `extendFlat({}, layout.meta, trace.meta)`.

So,

```js
trace = {
  meta: {yname: 'Y'},
  name: '%{meta.yname} - %{meta[0]}'
}
layout = {
  meta: ['Company B']
}
```

will print `Y - Company B`

src/components/titles/index.js
```diff
@@ -97,8 +99,10 @@ function draw(gd, titleClass, options) {
         if(!editable) txt = '';
     }
 
-    if(fullLayout.meta) {
-        txt = Lib.templateString(txt, {meta: fullLayout.meta});
+    if(options._meta) {
```
archmoj: Wondering if we need to bypass `[ ]` or `{ }` here?

alexcjohnson:
> The `meta` ref for trace is `extendFlat({}, layout.meta, trace.meta)`.

I guess if there isn't a `trace.meta` it makes sense to fall back on `layout.meta` - especially for backward compatibility for anyone currently using `layout.meta` in trace contexts. But merging the two seems a bit confusing, esp. since we're allowing - expecting even - that one will be an array and the other an object. I'd vote to just use `trace.meta || layout.meta`.

nicolaskruchten:
The downside of falling back is that if today I have 

```
trace = {
  name: '%{meta[0]}'
}
layout = {
  meta: ['Company B']
}
```

And then I set `trace.meta = {}` it will break in a way which is a bit confusing perhaps.

It's not a huge concern either way. I think there are probably zero users of `layout.meta` PLUS `trace.anything = '%{meta[ ... ]}'` in the wild today (including Chart Studio) so I'd personally be OK with the case above not working as of 1.48.

src/components/titles/index.js
etpinard: No need. `Lib.termplateString` handles those cases correctly.

nicolaskruchten:
"interesting" concern... 

```
data[0] = {
   meta: {x: 'hi'},
   name: '%{meta.x}',
   hovertemplate: '%{name}'
}
```

What will show up in the hover? What if I don't set `hovertemplate`? I assume the trace name shows up in the `extra` section as `hi`?

etpinard:
```js
Plotly.newPlot(gd, [{
   y: [1], hoverinfo: 'all',
   meta: {x: 'hi'},
   name: '%{meta.x}',
   hovertemplate: '%{data.name}'
}])
```

currently shows 
Peek 2019-05-17 11-37

and w/o `hovertemplate`:

```js
Plotly.newPlot(gd, [{
   y: [1], hoverinfo: 'all',
   meta: {x: 'hi'},
   name: '%{meta.x}',
   hovertemplate: '%{data.name}'
}])
```

we get:

Peek 2019-05-17 11-38

nicolaskruchten:
Yeah, that's what I figured. OK for now, and probably forever, given the crazy recursion possibilities. Wouldn't want to make our hovertemplate system Turing-complete.

etpinard:
Is everyone happy with the behaviour here? 

@antoinerg would you mind reviewing this at some point today? Thanks!

nicolaskruchten:
I’m happy!

antoinerg:
> Is everyone happy with the behaviour here?

I was really happy until I realized the following test fails:
```javascript
            Plotly.update(gd, {
                meta: {pizza: 'is yummy'},
                hovertemplate: 'TRACE -- %{meta.yname}<extra>%{meta.xname}</extra>'
            }, {
                meta: {yname: 'Yy', xname: 'Xx'}
            })
            .then(function() {
                Fx.hover('graph', evt, 'xy');

                assertHoverLabelContent({
                    nums: 'TRACE -- Yy',
                    name: 'Xx',
                    axis: '0.388'
                });
            })
```
If `meta` is defined at the trace-level and the requested key is missing, we don't make the effort of looking into the `layout.meta` object to see if it exists there. As a user, I kinda expected it to do so but maybe it's a bad idea? :thinking: 

Anyway, this behavior could be achieved by making `Lib.templateString` a variadic function (similar to `Lib.hovertemplateString`) and changing `Lib.templateString(d.name, d.trace._meta)` to `Lib.templateString(d.name, d.trace._meta, d.trace._meta.layout)`.

Let me know what you think @nicolaskruchten @plotly/plotly_js

nicolaskruchten:
I really think we should drop access to `layout.meta` via `meta` in `trace.any_template`. This feature was requested by me and even we never used it in prod in Chart Studio because what we really needed was this PR. I very strongly doubt anyone is using it in the wild: it wasn't documented anywhere.

antoinerg:
> I really think we should drop access to `layout.meta` via `meta` in `trace.any_template`.

Oh, I see @nicolaskruchten! I didn't realize this was a conscious decision. @alexcjohnson can you confirm that you are OK with this? If so, then the PR would be good to go since its new behavior is well :lock: down by the mock.

alexcjohnson:
> I really think we should drop access to `layout.meta` via `meta` in `trace.any_template`.

Confirmed, I'm happy with this.

antoinerg:
:dancer: :dancer: 
:dancer: :dancer:

nicolaskruchten:
> Confirmed, I'm happy with this.

Well, that's not what we went with here, right? here it uses `layout.meta` as `meta` if `trace.meta` isn't present. I was thinking we wouldn't even do that. But I'm not hung up on it :)

antoinerg:
> here it uses `layout.meta` as `meta` if `trace.meta` isn't present. 

Yes it does @nicolaskruchten.

> I was thinking we wouldn't even do that. But I'm not hung up on it :)
 
Ok good. I'll let @etpinard decide on this one.

nicolaskruchten:
well it's merged, so @etpinard already decided. Works for me, just wanted to make sure we're all on the same page. No further work or discussion required on this today :)

Issue #3814, opened 2019-04-29
Tickmark precision with percentage formatting


jonmmease:
Originally reported in https://github.com/plotly/plotly.py/issues/1537.

When the axis tickformat is set to `'p'` for percentage, the ticks are often displayed with an unexpectedly large amount of precision.

CodePen: https://codepen.io/jonmmease/pen/vMMNXJ?editors=1010
```javascript
var fig = {
  "data": [
    {
      "type": "scatter",
      "y": [
        0,
        1.0
      ]
    }
  ],
  "layout": {
    "yaxis": {
      "tickformat": "p",
      "automargin": true,
    }
  }
}


Plotly.newPlot('myDiv', fig, {showSendToCloud: true});
```
![newplot](https://user-images.githubusercontent.com/15064365/56892795-b2b96a80-6a4e-11e9-94c3-15c7e285166a.png)

PR #3867, opened 2019-05-15
Bump dev deps 2019-05-15

etpinard:
fixes https://github.com/plotly/plotly.js/network/alert/package-lock.json/tar/open (via https://github.com/sass/node-sass/issues/2625) and bumps a few more dev deps cc @plotly/plotly_js

PR #3868, opened 2019-05-15
Tickformat "p" fix

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3814 and fixes a bad commit from https://github.com/plotly/plotly.js/pull/3867

before: https://codepen.io/jonmmease/pen/vMMNXJ?editors=1010
after: <strike>https://codepen.io/etpinard/pen/NVpNEV?editors=1010</strike> https://codepen.io/etpinard/pen/yWbVqd?editors=1010

cc @plotly/plotly_js


archmoj:
@etpinard Thanks for the fix. 
Unfortunately there is another issue with small ranges. See [codepen](https://codepen.io/MojtabaSamimi/pen/OYpjPN?editors=0010) where the Y-axis range is printed as 0 to 100% whereas it should be 0 to 0.01%. 
Could we possibly address that in this PR?

etpinard:
> Y-axis range is printed as 0 to 100% whereas it should be 0 to 0.01%.

Oh boy. Thanks for noticing that!

archmoj:
> > Y-axis range is printed as 0 to 100% whereas it should be 0 to 0.01%.
> 
> Oh boy. Thanks for noticing that!

For the `funnelarea` I would add the following function to the `Lib`. Thought it may be of interest.
<pre>
lib.formatPercent = function(ratio, n) {
    n = n || 0;
    var str = (Math.round(100 * ratio * Math.pow(10, n)) * Math.pow(0.1, n)).toFixed(n) + '%';
    for(var i = 0; i < n; i++) {
        if(str.indexOf('.') !== -1) {
            str = str.replace('0%', '%');
            str = str.replace('.%', '%');
        }
    }
    return str;
};
</pre>

etpinard:
@archmoj how do you compute the `n` passed to your `formatPercent` function?

archmoj:
> @archmoj how do you compute the `n` passed to your `formatPercent` function?
That's the desired precision which in the case of financial traces is 0 or 1. 
But one may pass that after calculating what precision needed or simply by using a higher default e.g. 6.

etpinard:
I'm having second thoughts about my attempt at fixing this bug.

Closing for now.

Issue #3855, opened 2019-05-10
[bug] wrong range values for boxplot of layout type category


destradafilm:
When I inspect the range returned after the 'plotly_selected' callback in a boxplot with a categorical axis. I get the string values which is fine

But, if I want to get the pixel values or better yet the range that the selection translates to, I get strange results. 

I tried `d2p` and `d2c`  (`c2p` and `c2r` didnt produce any results)
and the results dont seem right. I believe there is a bug here unless I'm wrong

See the screen shot below
![Screen Shot 2019-05-10 at 11 48 41 AM](https://user-images.githubusercontent.com/6043499/57554892-e12d2480-732f-11e9-9e68-a4b63616601b.jpg)

Here is the codepen:
https://codepen.io/destrada/pen/EzKeMb

alexcjohnson:
`[undefined, undefined]` certainly looks like a bug. Thanks! Presumably we're trying to convert back to `d` "data value" format, but that doesn't make sense for categorical axes - for these we should be using `c` "calcdata", ie based on the category serial numbers.

(Note that if we ever get around to the log-range breaking change this will just be the `r` "range" format, as that will be defined as "`d` for all continuous axis types, `c` for discrete axis types." Right now it's not the same as `r` because log ranges are specified by their log values.)

destradafilm:
@alexcjohnson right, that's what I assumed that I should be using the `c` formating but none of those worked for me, even when I had actual values returned from the callback

Also, any chance there's documentation for these conversions anywhere? I basically had to try all of them 'trial and error' to find the best conversion function

alexcjohnson:
> any chance there's documentation for these conversions anywhere?

Good question, @etpinard ? The only place I'm aware of is buried deep in the code:

https://github.com/plotly/plotly.js/blob/e26f8f4c9c75a0381e75dd4a56035f7fbf07fe73/src/plots/cartesian/set_convert.js#L40-L55

That describes what the letters `d/c/l/p/r` mean, so you can extrapolate to the `x2y` functions. Perhaps we were trying to keep our options open, thinking if we don't document these functions we can change them without considering it a breaking change? 🙈 But obviously people like yourself have found them and depend on them so at this point we really should document them publicly.

etpinard:
> But obviously people like yourself have found them and depend on them so at this point we really should document them publicly.

Or include `px` value as part of the event data.

destradafilm:
> Or include px values as part of the event data.
and the range values (of the selection box) too?

PR #3869, opened 2019-05-15
Fix selection range event data on category axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3855 (at least the part identify in https://github.com/plotly/plotly.js/issues/3855#issuecomment-491428329)

before: https://codepen.io/destrada/pen/EzKeMb
after: https://codepen.io/etpinard/pen/pmeEXQ

cc @plotly/plotly_js


antoinerg:
Thank you for the fix and the very nice tests!

:dancer:

PR #3856, opened 2019-05-10
Miscellaneous geo fixes

etpinard:
Three geo fixes.  

The fix in https://github.com/plotly/plotly.js/commit/68bd47b754faac9e7a773f8d725444b1903bc08d is the most interesting where I spent most of the last two days trying to bring back my [`sane-topojson`](https://github.com/etpinard/sane-topojson) package back to life (a few things have changed in the [Natural Earth](https://www.naturalearthdata.com/) dataset and the [`topojson`](https://github.com/topojson/topojson) utility since 2016 :stuck_out_tongue_closed_eyes: ). I settled for a hacky but simple patch for now ... and until `sane-topojson` is back on solid grounds (cc https://github.com/etpinard/sane-topojson/pull/12)

@plotly/plotly_js reviewing this PR commit-by-commit should be straight-forward enough, but please let me know if you'd like a few codepen examples.

PR #1077, opened 2016-10-25
New topojson files!

etpinard:
fixes https://github.com/plotly/plotly.js/issues/899

Requires [`sane-topojson@2.0.0`](https://github.com/etpinard/sane-topojson)

PR #3874, opened 2019-05-16
Use sane-topojson v3

### LINKED ISSUE #3779: USA-states incorrectly drawn as foreign regions in high resolution choropleth
I’ve been working on a choropleth (through plotly.py originally) and things had been going fine at 110 resolution, but I decided to switch to 50 and things got weird. The problem I’m seeing is that I’m providing a list of ‘USA-states’ and it’s deciding to map some of them to similarly named world states, e.g. WA (Washington) is mapped on Western Australia and PA (Pennsylvania) is mapped on Pará, Brazil. This problem does not occur at all at the lower 110 resolution where WA is correctly mapped to the USA state as requested.

I’d actually like to take advantage of this hidden feature for other projects but I can only get it to happen through this glitch, not through choice. I’ve tried various other formats like "WA, AUS" or "AUS-WA" and both "USA-states" and "countries" formats. The only thing that causes Western Australia to be drawn is putting “WA” and “USA-states” and resolution=50.

https://jsfiddle.net/Thlayli/3jza9sov/
or
https://plot.ly/~thlayli/22/number-of-artists-by-country-or-region/#/ 1

etpinard:
The `sane-topojson` package is about to get a major upgrade https://github.com/etpinard/sane-topojson/pull/12: 
- it now ships topojsons generated with the latest [Natural Earth](http://www.naturalearthdata.com/downloads/) version (v.4.1.0, we were using v2)
- it now tracks 'governing unit' in the `subunits` layer properties (with key `gu`) allowing us to fix #3779 in an un-hacky way unlike https://github.com/plotly/plotly.js/pull/3856/commits/68bd47b754faac9e7a773f8d725444b1903bc08d 

That said, the new topojsons are _different_ then the current ones. Not by much (see https://github.com/plotly/plotly.js/commit/4eb79b966fe4884b5295083911e75661e291a185, https://github.com/plotly/plotly.js/commit/4336c304a2348d27c0a232a40a29eefcf201d121 and https://github.com/etpinard/sane-topojson/pull/12/commits/8624ec41ab73e6d1b2be8e40478154b2dc3670b3 to dig deeper), but they are different. As we don't version the topojsons on our CDN (e.g. at https://cdn.plot.ly/world_110m.json) used by default by [`Plotly.newPlot`](https://github.com/plotly/plotly.js/blob/acccdab66d78f942f0d7635615845bf26768a378/src/plot_api/plot_config.js#L333-L345), I'm a little reluctant to just merge this in. Perhaps we should retire the un-version URLs and start using e.g. https://cdn.plot.ly/world_v3.0.0_110m.json where the `vX.Y.Z` would correspond to the `sane-topojson` version?

Oh well, maybe I'm overthinking this. We did "override" the topojsons URLs back in https://github.com/plotly/plotly.js/pull/1077 and we didn't hear anyone complain about it. 

I'm interested in hearing what @plotly/plotly_js will have to say.


etpinard:
N.B. this PR is based off https://github.com/plotly/plotly.js/pull/3856

alexcjohnson:
The only changes I notice are Crimea transferring to Russia (accurate, whatever one thinks of its propriety) and in geo_canadian-cites (quelqu'un a oublié l'accent sur "cité"?) it seems Canada has lost a bunch of lakes, including the Quebec Eye. Much as I like lakes I don't think this is a big deal, and I don't think we should support multiple versions.

etpinard:
> it seems Canada has lost a bunch of lakes, including the Quebec Eye.

correct, `north_america_50m` has 

image

has 68 fewer lakes for some reason.

etpinard:
> geo_canadian-cites (quelqu'un a oublié l'accent sur "cité"?)

Haha good eye https://github.com/plotly/plotly.js/pull/3874/commits/51c6d1ca016c7bf874b68a79db905a9247e83dab

> Much as I like lakes I don't think this is a big deal, and I don't think we should support multiple versions.

Great. I'll merge this PR into https://github.com/plotly/plotly.js/pull/3856  :ok_hand:

PR #3876, opened 2019-05-17
Funnelarea traces

archmoj:
Plotly `funnelarea` traces based on `pie` & bar/funnel `insidetext` refactors.

resolves https://github.com/plotly/plotly.js/issues/3504

@plotly/plotly_js 

TODOs: 
- [x] add jasmine tests to cover different scenarios of  having both `pie` and `funnelarea` 
- [x] fixup `scalegroup` to handle scenarios with different dimension ratios  
- [x] improve `aspectratio` logic to take into account pixel values i.e. layout & domain ratios 

Codepen examples:
1. [Demo with pie and funnel](https://codepen.io/MojtabaSamimi/pen/arwVaw?editors=0010).
2. [Recat demo](https://codepen.io/MojtabaSamimi/pen/RmgjeY?editors=0010).
3. [Demo for testing various scaleGroup X aspectRatio X baseRatio](https://codepen.io/MojtabaSamimi/pen/dERZgm?editors=0010).


src/traces/pie/index.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
     moduleType: 'trace',
     name: 'pie',
     basePlotModule: require('./base_plot'),
-    categories: ['pie', 'showLegend'],
+    categories: ['pie-like', 'pie', 'showLegend', 'hiddenLabels'],
```
etpinard: Hmm. I don't think you're using `hiddenlabels` as a category anywhere.

src/traces/pie/layout_defaults.js
```diff
@@ -16,7 +16,9 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut) {
     function coerce(attr, dflt) {
         return Lib.coerce(layoutIn, layoutOut, layoutAttributes, attr, dflt);
     }
-    coerce('hiddenlabels');
+
     coerce('piecolorway', layoutOut.colorway);
     coerce('extendpiecolors');
+
+    if(!layoutOut.hasOwnProperty('hiddenlabels')) coerce('hiddenlabels');
```
etpinard: Why do we need this extra `if` clause?

src/plots/plots.js
```diff
@@ -467,6 +466,7 @@ plots.supplyDefaults = function(gd, opts) {
     newFullLayout._hasGL2D = newFullLayout._has('gl2d');
     newFullLayout._hasTernary = newFullLayout._has('ternary');
     newFullLayout._hasPie = newFullLayout._has('pie');
+    newFullLayout._hasFunnelarea = newFullLayout._has('funnelarea');
```
etpinard: No need for this. Please :hocho: 

These `fullLayout._has??` fields are just there for backward compatibility (mostly for our old toolpanel code). We shouldn't add new ones. If you need to add a `funnelarea`-only block somewhere downstream, you should use `fullLayout._has('funnelarea')` instead.

src/lib/index.js
```diff
@@ -1159,3 +1159,31 @@ lib.fillText = function(calcPt, trace, contOut) {
 lib.isValidTextValue = function(v) {
     return v || v === 0;
 };
+
+lib.formatPercent = function(ratio, n) {
```
etpinard: Interesting. I'm having second-thoughts about my attempt at fixing `tickformat:'p'` in https://github.com/plotly/plotly.js/pull/3868

src/traces/funnelarea/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+
+var extendFlat = require('../../lib/extend').extendFlat;
+
+module.exports = {
+    labels: pieAttrs.labels,
+    // equivalent of x0 and dx, if label is missing
+    label0: pieAttrs.label0,
+    dlabel: pieAttrs.dlabel,
+    values: pieAttrs.values,
+    marker: sunburstMarker,
```
etpinard: So, `funnelarea` traces like sunburst have a default `marker.line.width` of 1 and use `paper_color` to set the default `maker.line.color` ? I think that's ok, but I would prefer making this more explicit here:


```
    marker: {
        colors: pieAttrs.colors,
  
        line: {
            color: extendFlat({}, pieAtts.marker.line.color, {
                dflt: null,
                description: [
                    'Sets the color of the line enclosing each sector.',
                    'Defaults to the `paper_bgcolor` value.'
                ].join(' ')
            }),
            width: extendFlat({}, pieAtts.marker.line.width, {dflt: 1}),
            editType: 'calc'
        },
        editType: 'calc'
    }
```

That way both `sunburst` and `funnelarea` inherit from `pie`, and we don't need to make `funnelarea` inherit from `sunburst` too.

src/traces/funnelarea/layout_attributes.js
```diff
@@ -0,0 +1,43 @@
...
+* Copyright 2012-2019, Plotly, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var hiddenlabels = require('../pie/layout_attributes').hiddenlabels;
```
etpinard: It might be nice to add a line in the `hiddenlabels` description mentioning that it applies to `pie` and `funnelarea` labels simultaneously.

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+    Lib.makeTraceGroups(fullLayout._funnelarealayer, cdModule, 'trace').each(function(cd) {
+        var plotGroup = d3.select(this);
+        var cd0 = cd[0];
+        var trace = cd0.trace;
+
+        setCoords(cd);
+
+        // TODO: miter might look better but can sometimes cause problems
+        // maybe miter with a small-ish stroke-miterlimit?
+        plotGroup.attr('stroke-linejoin', 'round');
```
etpinard: We can't :hocho: this I think, it shouldn't make a difference on `funnelarea` trapezoids

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+
+        plotGroup.each(function() {
+            var slices = d3.select(this).selectAll('g.slice').data(cd);
+
+            slices.enter().append('g')
+                .classed('slice', true);
+            slices.exit().remove();
+
+            var quadrants = [
```
etpinard: `quadrant` isn't used anywhere for `funnelarea`. You can  :hocho: it

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+                .classed('titletext', true);
+            titleTextGroup.exit().remove();
+
+            titleTextGroup.each(function() {
+                var titleText = Lib.ensureSingle(d3.select(this), 'text', '', function(s) {
+                    // prohibit tex interpretation as above
+                    s.attr('data-notex', 1);
+                });
+
+                var txt = fullLayout.meta ?
```
etpinard: We'll need to generalise this to be inline with https://github.com/plotly/plotly.js/pull/3865 - either in this PR or in #3865 depending on which one gets merged first.

src/traces/funnelarea/layout_attributes.js
etpinard: Ah you did already. Thanks very much! Resolved.

src/traces/pie/index.js
archmoj: Yes. We have not use them yet.

src/traces/pie/layout_defaults.js
archmoj: To coerce once.

test/jasmine/tests/funnelarea_test.js
```diff
@@ -0,0 +1,1408 @@
...
+        }])
+          .then(function() {
+              _assertTitle('base', 'yo', 'rgb(0, 0, 255)');
+              _verifyTitle(true, false, true, false, false);
+          })
+          .catch(failTest)
+          .then(done);
+    });
+
+    it('still support the deprecated `title` structure (backwards-compatibility)', function(done) {
```
etpinard: Hmm. Maybe we shouldn't support `title: 'text'` signature for new trace types.

src/traces/pie/layout_defaults.js
etpinard: Are there any side-effects from coercing `hiddelabels` twice? If not, it's not worth adding complexity to skip one `coerce` call.

test/jasmine/tests/funnelarea_test.js
```diff
@@ -0,0 +1,1408 @@
...
+        expect(point.fullData).toBe(gd._fullData[0]);
+        expect(point.label).toBe('4');
+        expect(point.value).toBe(5);
+        expect(point.color).toBe('#1f77b4');
+        expect(point.id).toEqual(['maggie']);
+        expect(point.customdata).toEqual([{9: 10}]);
+
+        // for backward compat - i/v to be removed at some point?
+        expect(point.i).toBe(point.pointNumber);
+        expect(point.v).toBe(point.value);
```
etpinard: These should **not** appear in the `funnelarea` event data.

src/lib/index.js
```diff
@@ -1159,3 +1159,31 @@ lib.fillText = function(calcPt, trace, contOut) {
...
+    for(var i = 0; i < n; i++) {
+        if(str.indexOf('.') !== -1) {
+            str = str.replace('0%', '%');
+            str = str.replace('.%', '%');
+        }
+    }
+    return str;
+};
+
+lib.convexPolygonArea = function(points) {
```
etpinard: Looks like you're only using this function in `funnelarea_test.js`. Please move it to `funnelarea_test.js`, we don't want to bundle things we don't use in `src/`.

test/jasmine/tests/funnelarea_test.js
```diff
@@ -0,0 +1,1408 @@
...
+                unhoverCnt: 0,
+                hoverLabel: 1
+            });
+        })
+        .catch(failTest)
+        .then(done);
+    });
+});
+
+describe('Test funnelarea calculated areas', function() {
```
etpinard: Nice test :tada:

src/traces/pie/index.js
etpinard: let's :hocho: it then

src/traces/funnelarea/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+        flags: ['label', 'text', 'value', 'percent', 'name']
+    }),
+
+    hovertemplate: hovertemplateAttrs({}, {
+        keys: ['label', 'color', 'value', 'percent', 'text']
+    }),
+
+    textfont: pieAttrs.textfont,
+    insidetextfont: pieAttrs.insidetextfont,
+    outsidetextfont: pieAttrs.outsidetextfont,
```
etpinard: There's no way to set `textposition: 'outside'` at the moment for `funnelarea`, so we should :hocho: this.

src/traces/funnelarea/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+
+    title: extendFlat({}, pieAttrs.title, {
```
etpinard: I would prefer declaring `title` as:

```js
title: {
  text: pieAttrs.title,
  font: pieAttrs.font,
  position: extendFlat({}, pieAttrs.position, {
    values: ['top left', 'top center', 'top right']
  }),

  editType: 'plot'
}
```

for clarity.

src/traces/funnelarea/attributes.js
```diff
@@ -0,0 +1,85 @@
...
+        min: 0,
+        max: 1,
+        dflt: 0.333,
+        editType: 'plot',
+        description: [
+            'Sets the ratio between bottom length and maximum top length.'
+        ].join(' ')
+    },
+
+    sort: extendFlat({}, pieAttrs.sort, { dflt: false })
```
etpinard: I would :hocho: this, to leave room for improvement for future iterations. @archmoj what made you choose to add this in afterall?

archmoj: 1. Aggregated funnelarea graphs are not looking great without sorting.
2. In case of having a sorted pie with not sorted funnelarea labels may strangely hide other colors.

etpinard: > 1. Aggregated funnelarea graphs are not looking great without sorting.

Ok, but if `sort` defaults to false, then "Aggregated funnelarea graphs" will still "not look great" by default.

> In case of having a sorted pie with not sorted funnelarea labels may strangely hide other colors.

Can you share an example of this case?

src/traces/funnelarea/defaults.js
```diff
@@ -0,0 +1,76 @@
...
+
+    var lineWidth = coerce('marker.line.width');
+    if(lineWidth) coerce('marker.line.color', layout.paper_bgcolor);
+
+    coerce('marker.colors');
+
+    coerce('scalegroup');
+
+    var textData = coerce('text');
+    var textInfo = coerce('textinfo', Array.isArray(textData) ? 'text+value' : 'value');
```
etpinard: Here you're using `value` as the default, similar to `funnel` traces but unlike `pie` which uses `percent`.

I'd vote for making this align with `pie`: and using `percent` / `text+percent` as default values.

src/traces/funnelarea/calc.js
```diff
@@ -0,0 +1,24 @@
...
+'use strict';
+
+var pieCalc = require('../pie/calc');
+
+function calc(gd, trace) {
+    return pieCalc.calc(gd, trace);
+}
+
+function crossTraceCalc(gd) {
+    pieCalc.crossTraceCalc(gd, { type: 'funnelarea' });
```
etpinard: I was skeptical at first, but looks like reusing `pie/calc.js` almost verbatim works fine for various `piecolorway` and `funnleareacolorway` combos. For example,

```js
var labels = ['Lead', 'Pipeline', 'Proposal', 'Negotiation', 'Closed (Won)'];
var values = [23, 15, 11, 6, 2.1];

var colorwayLight = ['#7fc97f', '#beaed4','#fdc086']
var colorwayDark = ['#1b9e77','#d95f02','#7570b3']

Plotly.newPlot(gd, [{
  type: 'pie',
  labels: labels,
  values: values,
  domain: {row: 0, column: 0}
}, {
  type: 'funnelarea',
  labels: labels,
  values: values,
  domain: {row: 0, column: 1}
}], {
  grid: {rows: 1, columns: 2},
  piecolorway: colorwayLight,
  funnelareacolorway: colorwayDark,
})
```

yields:

image

generating the correct "extended" pie and funnelarea colors separately.

Nice job!

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+                    line(pt.BR, pt.BL) +
+                    line(pt.BL, pt.TL) +
+                    'Z';
+
+                slicePath.attr('d', shape);
+
+                // add text
+                var textPosition = pieHelpers.castOption(trace.textposition, pt.pts);
+                var sliceTextGroup = sliceTop.selectAll('g.slicetext')
+                    .data(pt.text && (textPosition !== 'none') ? [0] : []);
```
etpinard: This works. But currently for `funnelarea`, `textposition` isn't an available attributes. So these lines here are equivalent to: 

```js
var sliceTextGroup = sliceTop.selectAll('g.slicetext'.data(pt.text ? [0] : []);
```
---

That said, if we're planning to add support for `textposition: 'outside'` at some points, it might best to add `textposition` in this PR as an enumerated with values `['inside', 'none']`.

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+                    var transform;
+
+                    var x0, x1;
+                    var y0 = Math.min(pt.BL[1], pt.BR[1]);
+                    var y1 = Math.max(pt.TL[1], pt.TR[1]);
+
+                    x0 = Math.max(pt.TL[0], pt.BL[0]);
+                    x1 = Math.min(pt.TR[0], pt.BR[0]);
+
+                    transform = getTransformToMoveInsideBar(x0, x1, y0, y1, textBB, {
```
etpinard: This call is :trophy:

src/traces/funnelarea/defaults.js
```diff
@@ -0,0 +1,76 @@
...
+            moduleHasConstrain: false,
+            moduleHasCliponaxis: false,
+            moduleHasTextangle: false,
+            moduleHasInsideanchor: false
+        });
+    }
+
+    handleDomainDefaults(traceOut, layout, coerce);
+
+    var title = coerce('title.text');
```
etpinard: I could imagine adding logic for a `trace.name` becoming the default `title.text` when there are multiple `funnelarea` traces on the same graph.

Thoughts?

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+                titleText.attr('transform',
+                    'translate(' + transform.x + ',' + transform.y + ')' +
+                    (transform.scale < 1 ? ('scale(' + transform.scale + ')') : '') +
+                    'translate(' + transform.tx + ',' + transform.ty + ')');
+            });
+        });
+    });
+};
+
+function move(a) {
```
etpinard: This thing is only called once. I would :hocho:

src/traces/funnelarea/plot.js
```diff
@@ -0,0 +1,306 @@
...
+
+function setCoords(cd) {
+    if(!cd.length) return;
+
+    var cd0 = cd[0];
+
+    var aspectratio = cd0.trace.aspectratio;
+    var h = cd0.trace.baseratio;
+    if(h > 0.999) h = 0.999;
+    var h2 = Math.pow(h, 2);
```
etpinard: That's a cheeky line. What happens when `h=1`?

etpinard: ... ha I see, `v0` blows up. Is there a way to handle the `h=1` case separately?

archmoj: The origin moves to `infinity` which should be avoided. By the way `h=1` or `h=0.999` could be used to generate rectangles.

archmoj: > ... ha I see, `v0` blows up. Is there a way to handle the `h=1` case separately?

Yes. That might be a better option later on the road. But I think for the moment we could/should skip that.

etpinard: Sure, skipping is fine for now! 

Make sure to add a `TODO` above this line though.

test/jasmine/tests/funnelarea_test.js
etpinard: Fixed in https://github.com/plotly/plotly.js/commit/511bef61db73aa81c946ddb721c4604cc5e5091e - off a branch of mine.

Ok if I pushed this commit onto this branch?

archmoj: @etpinard. Thanks. OK please move forward.

test/jasmine/tests/funnelarea_test.js
etpinard: Thinking about this again. I think we should go all-in with the new API. Let's **not** support `title: 'text'`.

archmoj: Good. & What about `funnel`?

etpinard:
Nice work @archmoj :1st_place_medal: 

Most of my comments should be easy to resolve. In brief, we need to:

- [ ] figure out what to do with `sort`
- [ ] figure out what to do with `textposition`
- [ ] agree on the `tetinfo` default: `text+value` (like funnel) vs `text+percent` (like pie)

----

About

> * improve `aspectratio` logic to take into account pixel values i.e. layout & domain ratios

Let's meetup tomorrow to try to finish up the details there.

test/jasmine/tests/funnelarea_test.js
etpinard: `funnel` doesn't have a `title` attribute.

test/jasmine/tests/funnelarea_test.js
etpinard: https://github.com/plotly/plotly.js/pull/3876/commits/7c3369bbd3f64959e67ea4f38d255f97af1a33f8 

done, thanks!

test/jasmine/tests/funnelarea_test.js
archmoj: Good point. As another point the `funnelarea` mocks are using the new API. But we may need to drop/adapt few jasmine tests.

src/traces/funnelarea/defaults.js
archmoj: cc: @nicolaskruchten

src/traces/pie/index.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/abc3b39d159e5d4f53dc170b1d11f4fc3bc5f535.

src/traces/pie/layout_defaults.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/abc3b39d159e5d4f53dc170b1d11f4fc3bc5f535.

src/plots/plots.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/abc3b39d159e5d4f53dc170b1d11f4fc3bc5f535.

src/traces/funnelarea/attributes.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/316aff912a4cab9a6b899e86764977e0ee82766f.

src/traces/funnelarea/plot.js
archmoj: Another good call. Done in https://github.com/plotly/plotly.js/pull/3876/commits/7f2dc2abd32705842ce091474fb2b951a29221c3.

src/traces/funnelarea/plot.js
archmoj: Cut in https://github.com/plotly/plotly.js/pull/3876/commits/7e6b9aee160110a8d17f3be746f6ccf44531501d.

src/traces/funnelarea/defaults.js
nicolaskruchten: Agreed with @etpinard

archmoj: Thanks @nicolaskruchten for the quick reply.

src/traces/funnelarea/plot.js
archmoj: Cut in https://github.com/plotly/plotly.js/pull/3876/commits/7e6b9aee160110a8d17f3be746f6ccf44531501d.

src/traces/funnelarea/plot.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/7e6b9aee160110a8d17f3be746f6ccf44531501d.

src/traces/funnelarea/defaults.js
archmoj: Done in 8bd1875b390d91c07a8a8795b7eb9a84600c4ec9.

src/lib/index.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3876/commits/fa5b27332b474e86c1236e9af4c1ccdcbb7312ce. Also more tests were added for calculated areas in different `scalegroups`.

src/traces/funnelarea/attributes.js
archmoj: Revised in https://github.com/plotly/plotly.js/pull/3876/commits/38b474d3a4c47bf6a761ac9c0f2d5f0367632d35.
Not clear why `pie` uses `editType: 'calc'`?

src/traces/funnelarea/calc.js
archmoj: @etpinard Thanks for testing this.
Included this mock in https://github.com/plotly/plotly.js/pull/3876/commits/de8b5176a605452b58f4144b02414cd722625ece.

src/lib/index.js
etpinard: I just closed off https://github.com/plotly/plotly.js/pull/3868 (for now)

src/traces/funnelarea/attributes.js
etpinard: > Not clear why pie uses editType: 'calc'?

Good :eye: `editType: 'plot'` should suffice for pie and funnelarea `title.*` updates.

src/traces/funnelarea/attributes.js
archmoj: issue 1: [Codepen](https://codepen.io/MojtabaSamimi/pen/gJXmqE?editors=0010);
issue 2: [Codepen](https://codepen.io/MojtabaSamimi/pen/zQPZzK?editors=0010).

src/traces/funnelarea/attributes.js
```diff
@@ -0,0 +1,109 @@
...
+        colors: pieAttrs.marker.colors,
+        line: {
+            color: extendFlat({}, pieAttrs.marker.line.color, {
+                dflt: null,
+                description: [
+                    'Sets the color of the line enclosing each sector.',
+                    'Defaults to the `paper_bgcolor` value.'
+                ].join(' ')
+            }),
+            width: extendFlat({}, pieAttrs.marker.line.width, {dflt: 1}),
```
etpinard: I'm having second thoughts about this today. :thinking: 

I think I would prefer making the dflt `marker.line.width` 0 (like pie)

image


vs:

image

Any strong opinions from @plotly/plotly_js and/or @nicolaskruchten ?

src/traces/funnelarea/attributes.js
etpinard: You're right @archmoj - `funnelarea` traces that have aggregated values look a lot better sorted AND are a lot more useful.

So I think we should still :hocho: to `sort` attribute, but behind the scenes sort aggregated funnelarea calcdata during the calc step. I can't think of a reason to _not_ sort aggragated `funnelarea` traces.

src/components/legend/style.js
```diff
@@ -435,13 +436,24 @@ module.exports = function style(s, gd) {
...
-            .selectAll('path.legendpie')
-            .data(Registry.traceIs(trace, 'pie') && trace.visible ? [d] : []);
-        pts.enter().append('path').classed('legendpie', true)
+        var isVisible = (!desiredType) ? Registry.traceIs(trace, desiredType) :
+            (trace.type === desiredType && trace.visible);
+
+        var pts = d3.select(lThis).select('g.legendpoints')
+            .selectAll('path.legend' + desiredType)
+            .data(isVisible ? [d] : []);
+        pts.enter().append('path').classed('legend' + desiredType, true)
```
etpinard: Starting a new thread regarding  @archmoj 's:

> In case of having a sorted pie with not sorted funnelarea labels may strangely hide other colors.

from https://github.com/plotly/plotly.js/pull/3876#discussion_r285733483 which doesn't _just_ happen via `sort`.

---------

Consider the `funnelarea_pie_colorways` mock:

image

where the pie and funnelarea traces are set with the same `labels`/`values`. Here the legend draws one item per shared label using the colors of the pie trace (as that trace comes first in `gd.data`).

This is confusing as clicking on the legend items toggles both the pie and funnelarea items associated with the label. 

Potential solutions:

- drop `funnelareacolorway`, make funnelarea and pie traces always share the same colorway (set under `layout.colorway` or `piecolorway`)
- split the legend item square into two halves (in general into N parts), one half with the pie color, the other with funnelarea color

Right now, I think prefer splitting the legend item square. Perhaps this is sufficiently edgy case that we could leave for later though.

src/traces/funnelarea/attributes.js
etpinard: done in https://github.com/plotly/plotly.js/pull/3876/commits/32c96d2926d41c7b557635fc1da8dcfb765f5eaf

src/traces/funnelarea/plot.js
etpinard: Taking care of this one!

src/traces/funnelarea/plot.js
etpinard: Nicely done in https://github.com/plotly/plotly.js/pull/3876/commits/8af4d12318381574bf280a88ec348d7c0a8b79b2

src/components/legend/style.js
nicolaskruchten: why does this happen? I thought I saw a mock where the funnel and pie had the same colors. Is this because `piecolorway != funnelcolorway` ?

etpinard: > Is this because `piecolorway != funnelcolorway`

Yes in the case of `funnelarea_pie_colorways`. The same outcome can happen with via `marker.colors` too.

nicolaskruchten: hehe ah ok. it's quite an edge-case indeed. I would be fine with punting to another PR, although the split legend thing is a neat idea. Would it solve similar multi-pie/marker-color issues? i.e. two pies, same label, different marker.color?

etpinard: > Would it solve similar multi-pie/marker-color issues? i.e. two pies, same label, different marker.color?

Yes it could.

> I would be fine with punting to another PR, although the split legend thing is a neat idea

I'll open a new issue about it.

etpinard: Now tracked in https://github.com/plotly/plotly.js/issues/3885

Resolved.

src/traces/funnelarea/attributes.js
etpinard: Issue 1 and Issue 2 have been resolved in the scope of this PR.

src/traces/funnelarea/plot.js
etpinard: Done during latest master merge in https://github.com/plotly/plotly.js/pull/3876/commits/1426be7b6ea701f31e6e73d65a3190670818ce5c

src/traces/pie/layout_attributes.js
```diff
@@ -9,14 +9,16 @@
...
 module.exports = {
-    /**
-     * hiddenlabels is the pie chart analog of visible:'legendonly'
-     * but it can contain many labels, and can hide slices
-     * from several pies simultaneously
-     */
     hiddenlabels: {
         valType: 'data_array',
-        editType: 'calc'
+        role: 'info', // TODO: or style ?
```
etpinard: `'info'` !

src/traces/funnelarea/attributes.js
etpinard: restyle tests for `title.position` are added in https://github.com/plotly/plotly.js/commit/08eecaa28ba5bb759a3ab6dbc14434704ba679f8 (my branch)

the other attributes that @archmoj switched from editType `'calc'` to `'plot'` are properly tested in `pie_test.js`.

test/jasmine/tests/calcdata_test.js
```diff
@@ -985,6 +986,7 @@ describe('calculated data and points', function() {
                                 expect(gd._fullLayout[trace.type === 'splom' ? 'xaxis' : axName]._categories).toEqual(finalOrder, 'wrong order');
                             }
                         })
+                        .then(delay(100))
```
etpinard: cc @antoinerg i haven't seen `test-jasmine` fail since I added this in 10+ runs :tada:

src/traces/funnelarea/attributes.js
etpinard: No comments in ~24h (meaning no strong opinions, right :smirk: ), so I'll let @archmoj decide what he thinks looks best for `funnelarea`.

src/traces/funnelarea/attributes.js
etpinard: Done in https://github.com/plotly/plotly.js/pull/3876/commits/dd673a21af3e9510aa39d17cc2ccb0e25dab91f3

src/traces/funnelarea/defaults.js
etpinard: After some thoughts, it might be best to keep the pie/funnelarea symmetry for `scalegroup`.

Closing.

src/traces/funnelarea/attributes.js
archmoj: I also like the integrity in the `pie` version i.e. with no white lines.
But since the edges between two colour may not be rendered perfectly without gap, I suggest we keep them (by the way i.e. similar to `sunburst`). In addition it may help the figures be clear when printing a report document in black & white.

etpinard:
Awesome work @archmoj `funnelarea` is looking rock solid :muscle: 

v1.48.0 will be the first time a release will feature two new trace types added by the same dev. Be proud :1st_place_medal: 

:dancer: :dancer: (once CI gets its life back in order).

archmoj:
> Awesome work @archmoj `funnelarea` is looking rock solid 
> 
> v1.48.0 will be the first time a release will feature two new trace types added by the same dev. Be proud 
> 
>   (once CI gets its life back in order).

@etpinard obviously your review & supervision helped that possible.
Cheers.

PR #3888, opened 2019-05-22
`plotly_relayouting` event: live updates during panning/zooming

### LINKED ISSUE #2082: Add an event for panning and zooming that show live details as the range changes

antoinerg:
This closes #2082 by adding a `plotly_relayouting` event to emit live updates of coordinates change when panning and zooming.

- [x] cartesian
- [x] geo
  - [x] clipped
  - [x] non-clipped
  - [x] scoped
- [x] mapbox
- [x] polar
  - [x] angular drag
  - [x] radial drag
  - [x] rotation drag
- [x] gl3d
- [x] ternary

I used the abandoned PR https://github.com/plotly/plotly.js/pull/2606 for guidance but I had to modify it and write tests.

Tests could probably be DRYed up significantly but it gets the job done. I now need to investigate why they fail on CircleCI .


src/plots/geo/zoom.js
```diff
@@ -164,6 +165,7 @@ function zoomNonClipped(geo, projection) {
 
         didZoom = true;
         geo.render();
+        geo.graphDiv.emit('plotly_relayouting', {'projection.scale': projection.scale() / geo.fitScale});
```
etpinard: I think you're missing one `.emit` call in `zoomClipped`

https://github.com/plotly/plotly.js/blob/7a66dafc5e2b38a907b7757849240f432d6e2372/src/plots/geo/zoom.js#L195-L287

test/jasmine/tests/ternary_test.js
```diff
@@ -514,6 +514,33 @@ describe('ternary plots', function() {
...
+                    relayoutCallback = jasmine.createSpy('relayoutCallback');
+                    gd.on('plotly_relayout', relayoutCallback);
+                    gd.on('plotly_relayouting', function(e) {
+                        events.push(e);
+                    });
+                    return drag(path);
+                })
+                .then(function() {
+                    expect(events.length).toEqual(path.length - 1);
+                    // expect(relayoutCallback).toHaveBeenCalledTimes(1);
```
etpinard: :zap:

etpinard: ... you're right, we don't really need that `expect` statement. There's no other way to fill `events`.

test/jasmine/tests/geo_test.js
```diff
@@ -2066,4 +2066,31 @@ describe('Test geo zoom/pan/drag interactions:', function() {
         .catch(failTest)
         .then(done);
     });
+
+    describe('plotly_relayouting', function() {
```
etpinard: Geo subplots have three distinct zoom handlers (depending on the projection used). We should add one test per zoom handler. 

Note also, that we don't need to test `dragmode: 'zoom'` and `dragmode: 'pan'`, both dragmode do the same thing as geo subplot do not support zoombox interactions (yet).

src/plots/geo/zoom.js
antoinerg: Thanks for pointing that out @etpinard. I also just realized that I should output more information on `plotly_relayouting` to be on par with `plotly_relayout`. Right now, we're missing latitude and longitude.

test/jasmine/tests/geo_test.js
etpinard: > We should add one test per zoom handler.

Similar to:

https://github.com/plotly/plotly.js/blob/7a66dafc5e2b38a907b7757849240f432d6e2372/test/jasmine/tests/geo_test.js#L1650-L1651

https://github.com/plotly/plotly.js/blob/7a66dafc5e2b38a907b7757849240f432d6e2372/test/jasmine/tests/geo_test.js#L1750-L1751

https://github.com/plotly/plotly.js/blob/7a66dafc5e2b38a907b7757849240f432d6e2372/test/jasmine/tests/geo_test.js#L1839-L1840

test/jasmine/tests/cartesian_interact_test.js
```diff
@@ -188,6 +188,48 @@ describe('main plot pan', function() {
...
+            mouseEvent('mouseup', x1, y1);
+        }
+
+        var nsteps = 10; var events = []; var relayoutCallback;
+        Plotly.plot(gd, mock.data, mock.layout).then(function() {
+            // Switch to pan mode
+            modeBar = gd._fullLayout._modeBar;
+            var buttonPan = selectButton(modeBar, 'pan2d');
+            buttonPan.click();
+            expect(buttonPan.isActive()).toBe(true); // switched on dragmode
```
etpinard: Non-blocking, but

```js
var mock = Lib.extendDeep({}, require('@mocks/10.json');
mock.layout.dragmode = 'pan'
```

is a much simpler way to start a test in `dragmode: 'pan'`

test/jasmine/tests/polar_test.js
```diff
@@ -1247,6 +1247,70 @@ describe('Test polar interactions:', function() {
...
+                .then(function() {
+                    expect(events.length).toEqual(path.length - 1);
+                    expect(events[0]['polar.radialaxis.range']).toBeCloseToArray([6, 11], 0.1);
+                    expect(relayoutCallback).toHaveBeenCalledTimes(1);
+                })
+                .catch(failTest)
+                .then(done);
+            });
+        });
+        it('should emit plotly_relayouting events on angular drag', function(done) {
```
etpinard: What about radial drag:

Peek 2019-05-22 14-47

test/jasmine/tests/ternary_test.js
etpinard: Oh wait, I see what's going on. `plotly_relayout` isn't emitted on `pan` as `ternary_simple` in zoomed out on first render.

Try 

```js
var fig = Lib.extendDeep({}, require('@mocks/ternary_simple'));
fig.layout.ternary.aaxis.min = 0.15
```

and then both `pan` and `zoom` should trigger a `plotly_relayout` on drag-end.

etpinard:
Nicely done @antoinerg ! Thanks very much for taking this (no so fun) task on!

You're almost there. Let me know if I can help out with anything.

test/jasmine/tests/ternary_test.js
antoinerg: I thought this was a bug last night but it's not. Thank you for figuring it out for me :male_detective:

test/jasmine/tests/geo_test.js
antoinerg: Done in 2545746698588ac3f3edaa91bec2a1ac98ac33b3

src/plots/geo/zoom.js
antoinerg: In 2545746698588ac3f3edaa91bec2a1ac98ac33b3, I check that each key in the event data of `plotly_relayout` is also present in the event data of `plotly_relayouting`.

test/jasmine/tests/geo_test.js
etpinard: Awesome :tada:

test/jasmine/tests/ternary_test.js
antoinerg: I wrote my last comment too fast. `ternary_simple` is not zoomed in on the first render and I am not sure why my `plotly_relayout` event listener is called :thinking:

src/plots/geo/zoom.js
etpinard: Fantastic!

test/jasmine/tests/ternary_test.js
etpinard: > I am not sure why my `plotly_relayout` event listener is called

You're calling `drag` here, so shouldn't `plotly_relayout` get called once at the end of the drag motion?

test/jasmine/tests/polar_test.js
antoinerg: Done in 12658db141231bf1e8bc2606f23342c07dcdda40

test/jasmine/tests/ternary_test.js
antoinerg: It should but it doesn't at the moment... It's a bit disappointing because I would like to check that all the keys in the event data of `plotly_relayout` are also present in the event data of `plotly_relayouting`...

test/jasmine/tests/cartesian_interact_test.js
antoinerg: Thanks @etpinard 

Done in 12658db141231bf1e8bc2606f23342c07dcdda40

test/jasmine/tests/ternary_test.js
etpinard: Ok well no need to waste time on this detail. Resolved.

etpinard:
@antoinerg looks good! Thanks very (very) for the PR!

One `cartesian_interact` test is failing on CI and on my machine:

image

is it passing on your setup?

antoinerg:
> @antoinerg looks good! Thanks very (very) for the PR!
> 
> One `cartesian_interact` test is failing on CI and on my machine:
> 
> image
> 
> is it passing on your setup?

I don't remember it failing before but it does fail at the moment on my desktop.

antoinerg:
@etpinard It seems like the test that currently fails is affected by the timing of events. It checks the axis range _right after_ starting to drag. I suspect that now that we emit a `plotly_relayouting` event, _right after_ is at a different time:
https://github.com/plotly/plotly.js/blob/6e2da35615c39642a7b4f7e860c1163a7f94909f/test/jasmine/tests/cartesian_interact_test.js#L1464-L1483

A workaround is to update those "assert values". Doing so in commit https://github.com/plotly/plotly.js/commit/a76a52acdda5181fa82b1ac2f2e4033f340e5c2d makes the tests pass both locally and on the CI. Does that seem like a reasonable modification?

etpinard:
> It checks the axis range _right after_ starting to drag. I suspect that now that we emit a `plotly_relayouting` event, _right after_ is at a different time:

That makes perfect sense. So adding `plotly_relayouting` _does_ affect perf (by a little bit) it looks like.



> A workaround is to update those "assert values". Doing so in commit [a76a52a](https://github.com/plotly/plotly.js/commit/a76a52acdda5181fa82b1ac2f2e4033f340e5c2d) makes the tests pass both locally and on the CI. Does that seem like a reasonable modification?

Yep, nicely done. Could you just manually check that zoom-drag https://codepen.io/etpinard/pen/eaVQze?editors=1111 work correctly on your branch?

antoinerg:
> Yep, nicely done. Could you just manually check that zoom-drag https://codepen.io/etpinard/pen/eaVQze?editors=1111 works correctly on your branch?

It looks fine as far as I can tell: https://codepen.io/antoinerg/pen/Jqpebm?editors=0001

etpinard:
Awesome work :dancer: !

Issue #3872, opened 2019-05-16
histogram2d: wrong hover labels


antoinerg:
In `histogram2d`, the hover labels display wrong `y` coordinates:
![Screenshot_2019-05-16_15-57-50](https://user-images.githubusercontent.com/301546/57883249-a590cf80-77f3-11e9-9804-0a914a7a6a7e.png)

Codepen: https://codepen.io/antoinerg/pen/EzmmdP?editors=0010

etpinard:
In https://codepen.io/antoinerg/pen/EzmmdP?editors=0010 (the codepen url pasted above) hover seem to behave correctly:

![Peek 2019-05-22 10-37](https://user-images.githubusercontent.com/6675409/58183615-a0ad9f00-7c7d-11e9-97da-bf61e25ddd2f.gif)

antoinerg:
Sorry for the inaccurate Codepen @etpinard. I updated it to be an `histogram2d`. You should see the bug now.

etpinard:
Oh boy, that's bad. Thanks for spotting that @antoinerg !!

etpinard:
Looks like this was working properly in 1.31.2 https://codepen.io/etpinard/pen/BeJJvv and started misbehaving in 1.32.0 https://codepen.io/etpinard/pen/BeJJvv potential as a side-effect of https://github.com/plotly/plotly.js/pull/2113 and/or https://github.com/plotly/plotly.js/pull/2213 (-> full [changelog](https://github.com/plotly/plotly.js/releases/tag/v1.32.0))

PR #3890, opened 2019-05-22
Use correct index to lookup unique histogram2d y vals

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3872 - by patching only 3 characters in `histogram2d/calc.js` :smirk: @plotly/plotly_js 

before: https://codepen.io/antoinerg/pen/EzmmdP?editors=0010 (which has been broken since https://github.com/plotly/plotly.js/pull/2113/ released in `1.32.0` :expressionless: )
after: https://codepen.io/etpinard/pen/xNpMOX


antoinerg:
:smiley: beautiful fix @etpinard

Thank you for fixing this nasty issue :dancer:

PR #3891, opened 2019-05-22
fix pie-like legend when line color array has length greater than 3

### LINKED ISSUE #3871: Pie legend color is not correct after the third one
[codepen](https://codepen.io/MojtabaSamimi/pen/qGmmWB?editors=0010)

archmoj:
Fix #3871.

@plotly/plotly_js


src/lib/index.js
```diff
@@ -644,11 +643,7 @@ lib.minExtend = function(obj1, obj2) {
         if(k.charAt(0) === '_' || typeof v === 'function') continue;
         else if(k === 'module') objOut[k] = v;
         else if(Array.isArray(v)) {
-            if(k === 'colorscale') {
```
etpinard: Hmm. This probably slows down performance in some cases. 

Can we make try adding another special (similar to `k === 'colorscale'`) instead?

alexcjohnson: Or perhaps a third arg to the function, to override `arrayLen = 3`, that gets invoked only when needed?

etpinard: > Or perhaps a third arg to the function, to override `arrayLen = 3`, that gets invoked only when needed?

Good call - that would work!

archmoj: Revised in c960d0d6c3e1005f143c30e05aa795cb477010a6.

etpinard: As `Lib.minExtend` is used in the toolpanel code, I would prefer if we would try to fix this bug without patching `Lib.minExtend`.

Sorry for the confusion @archmoj

alexcjohnson: a third arg would be fine with toolpanel. I wouldn't want to add an exception like `color` because in some contexts that can be enormous. I would much prefer to pick out `pie` traces explicitly and extend the full length in that case, since we wouldn't be bothered by poor perf if someone gives us a pie with 100K slices 😏

etpinard:
nicely done :dancer:

PR #3904, opened 2019-05-24
Scatter3d and scattergl handling rgb colors with extra alpha values

### LINKED ISSUE #3902: Scatter3d & Scattergl should not display alpha when having rgb colors not rgba colors
[Scatter3d codepen](https://codepen.io/MojtabaSamimi/pen/EzQMaQ?editors=0010).
[Scattergl codepen](https://codepen.io/MojtabaSamimi/pen/EzEjRX?editors=0010).

archmoj:
Fixes #3902 .
Colors passed as `rgb` with an extra `alpha` value could cause some traces such as `scatter3d` & `scattegl` to enable opacity (as if `rgba` was applied).
This happens while [color-normalize](https://www.npmjs.com/package/color-normalize) parses alpha for `rgb` similar to `rgba`. 

Also dependencies of various gl modules including regl-splom, regl-scatter2d, regl-line2d, regl-error2d, gl-text are patched and updated in this PR.

@plotly/plotly_js 

TODO:
- [x] Adding more tests


archmoj:
This is now `reviewable` excluding two reverted commits. 
And here is all those patches end up:
https://github.com/colorjs/color-parse/compare/master...archmoj:rgb-has-no-alpha

etpinard:
I think patching `color-parse` (in your https://github.com/colorjs/color-parse/compare/master...archmoj:rgb-has-no-alpha) is the way to go, but that would probably be called a breaking change for `color-parse`.

You can try making a PR right away, but I suspect you'll have to make that an option in order to get it merged.

package-lock.json
```diff
@@ -9300,37 +9327,59 @@
...
+        "color-rgba": "git://github.com/archmoj/color-rgba.git#92f2babae20c1eec60fd9185719baf3f85a7ba56",
         "flatten-vertex-data": "^1.0.2",
         "glslify": "^7.0.0",
         "image-palette": "^2.1.0",
         "is-iexplorer": "^1.0.0",
         "object-assign": "^4.1.1",
         "parse-rect": "^1.2.0",
         "pick-by-alias": "^1.2.0",
-        "point-cluster": "^3.1.4",
+        "point-cluster": "^3.1.5",
```
etpinard: Hmm. Do we really want to bump `point-cluster`?

archmoj: It has already been applied within the master branch of `regl-scatter2d`: https://github.com/gl-vis/regl-scatter2d/commit/cda45ae30e5e1f841d65ced2404c9e36a618415f. Yet we don't have a new reg-scatter2d release.

etpinard: Right, so can we wait until we need to bump `regl-scatter2d` to bump `point-cluster`?

archmoj: Yes. Exactly.

etpinard: ... unless you feel comfortable with https://github.com/dy/point-cluster/commit/8419f968f7818ce9f347e7ac2bf4940f2fcfa337

etpinard: Ah I see, `regl-scatter2d` uses `color-parse`. Yep this package-lock diff makes perfect sense. Sorry for the confusion :ok_hand:

etpinard:
@archmoj will it be possible to get this PR in 1.48.0? Do you need anything else from Dima?

archmoj:
The only blocker is this PR: https://github.com/colorjs/color-normalize/pull/4.
I suspect the reason behind the failed test could be node 12 & not color-rgba 2.1.1?

archmoj:
@etpinard finally this thing is now reviewable.

etpinard:
:dancer: thanks!

etpinard:
@archmoj Looks like this PR broke `./tasks/noci_test.sh image`. On `master`, I get:

```
$ ./tasks/noci_test.sh image
renderer error - Uncaught ReferenceError: Plotly is not defined
          Chrome version 59.0.3071.115
          Electron version 1.8.4
```

adding `--debug` to 


https://github.com/plotly/plotly.js/blob/b6edba5db568cda2ff246a12aa1552281c988fc8/tasks/noci_test.sh#L31-L38

we get:

image

which appears to come from `to-px`, added in commit https://github.com/mikolalysenko/to-px/commit/bd2ce00d081f37a6d3c3ee4cee47f774b09429c9 released in `to-px@1.1.0` 

---------

Other info:

- `npm ls to-px` gives:

image

- this sounds similar to https://github.com/plotly/plotly.js/pull/2783


Maybe bumping down `to-px` to `1.0.1` in `gl-text` would fix the problem @archmoj ?

etpinard:
... of course this will be :no_entry: blocking :no_entry: the release of `v1.49.0`

archmoj:
I am on it.

Issue #3906, opened 2019-05-24
Strange label position in sunburst plot


jcalve:
https://codepen.io/anon/pen/rgdmxV

The last parent in the plot, in this case "Parent Name Seven" appears in the middle of the plot unlike the others.

It doesn't seem to happen with higher a number of children per parent.

etpinard:
Thanks for the report

PR #3907, opened 2019-05-24
Fix sunburst text in sectors centered around theta=180

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3906

before: https://codepen.io/anon/pen/rgdmxV
after: https://codepen.io/etpinard/pen/oRqMxP


etpinard:
cc @plotly/plotly_js

archmoj:
Good catch!
:dancer:

etpinard:
thanks for the quick review!!

Issue #3794, opened 2019-04-19
Inconsistent parcoords behaviour when data is outside range


slishak:
Example codepen: https://codepen.io/anon/pen/xejYmv?editors=1010

In this example, dimension `C` has an out-of-range point at `C=4`. On page load, this line is shown in grey (I'm not sure if this should be the case, as it should be selectable in dimensions `A` and `B`):
![image](https://user-images.githubusercontent.com/12515343/56428087-6dfb3a00-62be-11e9-9136-7d594af68d4d.png)

Attempting to select the line on dimension `B` confirms that it is not selectable:
![image](https://user-images.githubusercontent.com/12515343/56428125-92efad00-62be-11e9-9514-c5b503b63e0f.png)

However, when clearing the constraint range by clicking on the `B` axis, the line disappears completely:
![image](https://user-images.githubusercontent.com/12515343/56428148-a69b1380-62be-11e9-9c5d-40df45d467a6.png)

And from that point onwards, it reappears whenever a constraint range is applied, and disappears again when they are cleared.

In my opinion, the default behaviour should be for lines going out of the range of one of the coordinates to not have any special treatment. This is because you might have a large number of coordinates, and points can have invalid data for one dimension while still being valid for all of the rest of the dimensions (and the status of this point as an outlier compared to the expected range could even make this point more interesting to the user, so shouldn't be hidden).

etpinard:
Thanks for reporting!

Issue #2950, opened 2018-08-28
Parcoords: cannot select the first enumerated value with a zero-width range


alexcjohnson:
The Parallel Coordinates plot does not select the first enumerated value when the individual values are selected as their own constraint ranges.

For example, if the values are [1 ,2, 3] and the constraint range is [[1, 1], [2, 2], [3, 3]], bars will show up next to 2 & 3, but not 1.

This following codepen demonstrates this issue: https://codepen.io/donolingerPHX/pen/oPxQyM

Look at the IndivConstsEnum dimension. Bars show up next to all values except Delta1.

Changing the range for Delta1 (value 1) to [1, 1.01] is enough to get the bar to show up.

The problem seems to exist on the first enumerated value of any dimension, even if the selection is out of order (eg `[[2, 2], [1, 1], [3, 3]]` it's still `1` that doesn't show up).

reported by @donolingerPHX

PR #3915, opened 2019-05-28
Fix2950 parcoords cannot select first enum

archmoj:
Fix for #2950 and https://github.com/plotly/phoenix-integration/issues/201.

The bug was introduced at this line; whereas `i === 1` should have been applied. 
https://github.com/plotly/plotly.js/blob/4a8dd1c326f9e4a98ae090e9748123718f9d55db/src/traces/parcoords/axisbrush.js#L55

The logic within low and high functions are also refactored and merged into one function.

@plotly/plotly_js


etpinard:
Nice fix and cleanup!

:dancer: 

---

Does this also fix https://github.com/plotly/plotly.js/issues/3794 ?

PR #3917, opened 2019-05-28
treating rgba colors similar to rgb on parallel coordinates

### LINKED ISSUE #3820: Using rgba colors in colorscales for parcoords plots results in all black lines.
## Observed 

The `parcoords` plot behaves differently for `rgba` color strings than others (`rgb`, `hex`, `string`) when used in a colorscale. When using `rgba` all lines are rendered black. When using a different color string type in the colorscale, lines are drawn in their respective colors.

## Expected

The type of the color string should not impact the functionality of the colors in the trace.

## Reproduction

This is the definition of the colorscale used in the example.

### Broken

Codepen: https://codepen.io/SaschaJust/pen/JVqoaL

```js
"colorscale": [
            [
                0,
                "rgba(255,0,0,1)"
            ],
            [
                1,
                "rgba(0,0,255,1)"
            ]
        ]
```

parcoords_rgba

### Working

If you replace the `rgba` strings by their `rgb` counterparts, the plot renders as expected.

https://codepen.io/SaschaJust/pen/xeNGOy

```js
"colorscale": [
            [
                0,
                "rgb(255,0,0)"
            ],
            [
                1,
                "rgb(0,0,255)"
            ]
        ]
```

parcoords_rgb

Notice that the colorbar works fine in both cases.

archmoj:
Fixes #3820 for `parcoords`.

@plotly/plotly_js


etpinard:
Nice fix :dancer: !

PR #3924, opened 2019-05-31
fix for ScatterGL hover and click events give the 'top' point instead of the 'bottom' point

archmoj:
Fix for https://github.com/plotly/phoenix-integration/issues/241 i.e. in `mode`: `closest`.

@plotly/plotly_js

### DIFF
--- a/src/traces/scattergl/index.js
+++ b/src/traces/scattergl/index.js
@@ -732,7 +732,7 @@ function hoverPoints(pointData, xval, yval, hovermode) {
             }
         }
     } else {
-        for(i = 0; i < ids.length; i++) {
+        for(i = ids.length - 1; i > -1; i--) {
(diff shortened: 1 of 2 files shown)


etpinard:
Brilliant fix and test :dancer:

secretwpn:
Hi!
I was about to open an issue about this, but tried to search first and came across this issue.

Currently if I have a bunch of `scattergl` traces with `hovermode: closest` and there is a cluster of overlapping points,
the `plotly_click` event handler will get only the "bottom of the stack" point, which is completely counter-intuitive - if the point is rendered on the top (closest to user) - I expect this point to come with an event, not the one that is buried under the whole stack.

But apparently this is an intended behavior coming from this fix?

What is the reason for this? 
Is there any way to get the topmost point from the event?

Thanks!

PR #3925, opened 2019-05-31
Correct parcoords tickvals and ticktext documentation - no tickmode property

### LINKED ISSUE #3313: parcoords tickvals documentation mentions non-existant tickmode property
The schema documentation for `parcoords.dimension.tickvals` reads:

```
Sets the values at which ticks on this axis appear. Only has an
effect if `tickmode` is set to "array". Used with `ticktext`.
```

And `parcoords.dimension.ticktext`:
```
Sets the text displayed at the ticks position via `tickvals`.
Only has an effect if `tickmode` is set to "array". Used with
`tickvals`.
```

But there is no `parcoords.dimension.tickmode` property.  These properties work fine on their own, so it looks like this is just a documentation issue.

archmoj:
Simply fixes #3313.

@plotly/plotly_js

### DIFF
--- a/src/traces/parcoords/attributes.js
+++ b/src/traces/parcoords/attributes.js
@@ -41,8 +41,18 @@ module.exports = {
         },
         // TODO: better way to determine ordinal vs continuous axes,
         // so users can use tickvals/ticktext with a continuous axis.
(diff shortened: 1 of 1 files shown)


etpinard:
:dancer: Thanks!

PR #3932, opened 2019-06-03
improve robustness of circular Sankey 

### LINKED ISSUE #3813: Sankey error state
In https://github.com/plotly/plotly.py/issues/1536, @adurivault discovered a combination of Sankey links that causes Plotly.js to freeze up.

CodePen: https://codepen.io/jonmmease/pen/OGGPjL?editors=1010
```JavaScript
var fig = {
  "data": [
    {
      "link": {
        "source": [
          2,
          2,
          4,
          2,
          3,
          0,
          1,
          0
        ],
        "target": [
          0,
          4,
          2,
          3,
          2,
          4,
          2,
          1
        ],
        "value": [
          1,
          1,
          1,
          1,
          1,
          1,
          1,
          1
        ]
      },
      "node": {
        "label": [
          "A",
          "B",
          "C",
          "D",
          "E"
        ]
      },
      "type": "sankey"
    }
  ],
  "layout": {}
}

Plotly.newPlot('myDiv', fig, {showSendToCloud: true});
```
Interestingly, if any single link is commented out, then the diagram renders successfully.

antoinerg:
Closes #3813

As reported in #3813:
1st dataset:
Before: https://codepen.io/antoinerg/pen/BebjwJ?editors=0010 (hangs indefinitely)
After: https://codepen.io/antoinerg/pen/arMvXX

2nd dataset:
Before: https://codepen.io/joetristano/pen/wbKjbo (hangs indefinitely)
After: https://codepen.io/antoinerg/pen/eaXJEo

Minified bundle: https://40395-45646037-gh.circle-artifacts.com/0/dist/plotly.min.js


package.json
```diff
@@ -66,7 +66,7 @@
     "d3-force": "^1.0.6",
     "d3-hierarchy": "^1.1.8",
     "d3-interpolate": "1",
-    "d3-sankey-circular": "0.33.0",
+    "d3-sankey-circular": "git@github.com:antoinerg/d3-sankey-circular.git#8943bbd049c4651dbb230aa0e55ee8aad5d85afd",
```
antoinerg: https://github.com/tomshanley/d3-sankey-circular/compare/master...antoinerg:8943bbd049c4651dbb230aa0e55ee8aad5d85afd

antoinerg:
One baseline was altered by the hotfix and unfortunately, it is for the worse... The layout was better before. On average, however, the difference should not be appreciable.

etpinard:
> One baseline was altered by the hotfix and unfortunately, it is for the worse... The layout was better before. On average, however, the difference should not be appreciable.

Ok, interesting. Do you have an idea of how common data similar to `sankey_circular_process` is? 

For what it's worth, I'm ok with this patch. Deteriorating some cases while fixing cases that were 100% broken sounds like a win to me, but will I don't think we should include this in a patch release; this might have to wait for 1.49.0

package.json
etpinard: You might want to make a PR to `d3-sankey-circular`. Maybe the author will an opinion and/or tips about your patch

antoinerg: Ok, I sent a PR upstream: https://github.com/tomshanley/d3-sankey-circular/pull/36 !

antoinerg:
> Ok, interesting. Do you have an idea of how common data similar to sankey_circular_process is?

I really can't say. I haven't seen that many Sankey diagrams circular links in the wild.

> For what it's worth, I'm ok with this patch. Deteriorating some cases while fixing cases that were 100% broken sounds like a win to me,

It is definitely a win. That being said, down the road, it might be worth investigating ways to improve the default position of nodes.

package.json
etpinard: Hmm. That's frustrating. No news from the `d3-sankey-circular` maintainers. We might have to make a new package out of your fork e.g. `@plotly/d3-sankey-circular`

antoinerg: No problem! I will make a release today

antoinerg: Done in 5fed1a5

etpinard: :tada: https://www.npmjs.com/package/@plotly/d3-sankey-circular

etpinard:
Nicely done :dancer: !

Issue #3929, opened 2019-06-03
contour labels missing if x/y data is reversed


alexcjohnson:
Related to https://github.com/plotly/plotly.js/issues/3279 but this time the x/y data in the trace rather than the axis range: if it's reversed, either for x, y, or both, the labels will not appear.

https://codepen.io/alexcjohnson/pen/pmGBWq?editors=0010

PR #3934, opened 2019-06-04
Fix contour label rendering for non-monotonically increasing x/y

etpinard:
fixes https://github.com/plotly/plotly.js/issues/3929 

Following up on https://github.com/plotly/plotly.js/pull/3293 and https://github.com/plotly/plotly.js/pull/3352 

before: https://codepen.io/alexcjohnson/pen/pmGBWq?editors=0010
after: https://codepen.io/etpinard/pen/xNeZwB?editors=1010

cc @plotly/plotly_js


src/traces/contour/plot.js
```diff
@@ -244,16 +244,19 @@ function makeLinesAndLabels(plotgroup, pathinfo, gd, cd0, contours, perimeter) {
             .attr('data-notex', 1)
             .call(Drawing.font, contours.labelfont);
 
+        var _x = cd0.x.slice().sort(Lib.sorterAsc);
+        var _y = cd0.y.slice().sort(Lib.sorterAsc);
```
alexcjohnson: non-⛔️ but if we only need the bounds we can do better than a full sort of the arrays.

archmoj: Addressed in https://github.com/plotly/plotly.js/pull/3934/commits/9f6d1b63a0f9e2bc7676847d099937c8a868033e.

archmoj:
Excellent!
:dancer:

PR #3951, opened 2019-06-11
Bump dev deps 11-06-2019 edition

etpinard:
@plotly/plotly_js this should be enough to clear our security [alerts](https://github.com/plotly/plotly.js/network/alerts).


archmoj:
:dancer:

Issue #2466, opened 2018-03-09
[Webpack] Module not found: Error: Can't resolve 'vertx' in '/home/user/app/node_modules/plotly.js/dist'


brussee:
Error with plotly.js 1.35.1 and webpack 3.11.0 when importing: `import Plotly from 'plotly.js/dist/plotly-with-meta'`
```
./node_modules/plotly.js/dist/plotly-with-meta.js
Module not found: Error: Can't resolve 'vertx' in '/home/user/app/node_modules/plotly.js/dist'
 @ ./node_modules/plotly.js/dist/plotly-with-meta.js 61621:16-26
 @ ./src/entryPoint.js
 @ multi ./src/entryPoint.js webpack-hot-middleware/client?path=http://localhost:3000/__webpack_hmr
```

brussee:
Exact same issue with `./node_modules/plotly.js/dist/plotly.min.js` and `./node_modules/plotly.js/dist/plotly.js` as well.
`Module not found: Error: Can't resolve 'vertx' in '/home/user/app/node_modules/plotly.js/dist'`

nicolaskruchten:
Just to confirm: this is showing up as a warning, and compilation proceeds, right?

chiaramdelucia:
@nicolaskruchten I am having the same issue and yes, it is logging as a `warning` and compilation proceeds.

chiaramdelucia:
But, to note, I am now getting too many WebGL contexts in the console, where as before I wasn't. Not sure if this is related, but it did pop up with the 1.35.1 and this warning.

etpinard:
> Not sure if this is related, but it did pop up with the 1.35.1 and this warning.

That shouldn't be related. @chiaramdelucia would you mind sharing a reproducible example? Thank you!

chiaramdelucia:
@etpinard I didn't really think it would be, but thought I would note it JIC. 

As for reproducible code, here is a [codepen](https://codepen.io/chiaramdelucia/pen/ppMdOb?editors=1010). This isn't *exactly* what happens in my app, but it is similar. In the pen's console you'll see the following on the first button click, but not after that.
`Error: WebGL warning: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one.` 

In my app, which uploads a dataset, it prints the following on the third time a dataset is uploaded. 
`WARNING: Too many active WebGL contexts. Oldest context will be lost.`

etpinard:
[...]

PR #3952, opened 2019-06-12
Bump more dependencies - edition 2019 June 11

archmoj:
Bump es6-promise and topojson-client;
Uninstall font-atlas-sdf i.e. an unused dependency.

@plotly/plotly_js 

<b>Correction:</b>
No `es6-promise` and `topojson-client` in this PR.
Only removing  `font-atlas-sdf` dependency here.


etpinard:
Thanks, but we can't _just_ bump those.


Bumping `es6-promise` is a breaking change (they don't have the same browser support if I remember correctly, cc https://github.com/plotly/plotly.js/issues/2466)

`topojson-client` also does something different, it uses a bunch more memory and makes `geo` subplot slower we tried it when working on https://github.com/plotly/plotly.js/pull/3856

etpinard:
> Only removing `font-atlas-sdf` dependency here.

Good eye. We swap out `font-atlas-sdf` for [`svg-path-sdf`](https://github.com/dy/svg-path-sdf) during the regl `scattergl` push https://github.com/plotly/plotly.js/pull/2258

:dancer: thanks!

archmoj:
@etpinard Awesome. 
Thanks for finding out the reason why it was useful.

Issue #3960, opened 2019-06-13
all textinfo/hoverinfo flags should be available to hovertemplate


nicolaskruchten:
Following on from #3954 ... I'm not sure if this is or isn't the case for `pie` already, but it should be 🔒'ed down please :)

archmoj:
Thanks for the report.
The `percent` flag seem to be not available in `hoverinfo` of `pie` & `funnelarea`.

archmoj:
<b>Correction</b>: there is no issue found with `pie` and `funnelarea` with `percent` flag.

etpinard:
@archmoj how's our test coverage for `funnelarea` + `hoverinfo` and `hovertemplate` ?

archmoj:
> @archmoj how's our test coverage for `funnelarea` + `hoverinfo` and `hovertemplate` ?

I <i>think</i> that is pretty good.

For `hovertemplate` we have:
https://github.com/plotly/plotly.js/blob/d7d2bdd66e1afaff222eb4e5dc76e7b21db1069a/test/jasmine/tests/funnelarea_test.js#L965-L993

And for `hoverinfo` we have:
https://github.com/plotly/plotly.js/blob/d7d2bdd66e1afaff222eb4e5dc76e7b21db1069a/test/jasmine/tests/funnelarea_test.js#L877-L892

etpinard:
Yep, good enough.

I think you can close this issue @archmoj

PR #3958, opened 2019-06-13
Implement percentage flags to control funnel hoverinfo and hovertemplete

archmoj:
fix #3954 
Control percentages via `hoverinfo` and `hovertemplate`.
Note: the flags are in `camelCase` format. 

[codepen](https://codepen.io/MojtabaSamimi/pen/orXwKX?editors=0010)

@plotly/plotly_js 
cc: @nicolaskruchten

Issue #3954, opened 2019-06-12
Funnel hoverinfo incomplete


nicolaskruchten:
In building the RCE controls for `funnel` I notice that this requirement wasn't met: https://github.com/plotly/plotly.js/issues/3504#issuecomment-486423160 ... I can't turn the three percentages on or off via the `hoverinfo` flaglist as far as I can tell.

archmoj:
Is there is a similar issue with `bar` and `waterfall`?

nicolaskruchten:
We didn't discuss anything specific for waterfall, but we could discuss it later. Let's fix the funnel case for now please, as it's a well-spec'ed requirement we already agreed to :)

archmoj:
Should we add `hoverinfo` flags (initial | previous | total) for each of those percentages or turning off the three using `none` or `skip` is enough?

nicolaskruchten:
All three please, same as `textinfo`, just like in the spec :)

nicolaskruchten:
Question: is it possible to access these via `hovertemplate` in some way?

etpinard:
> Question: is it possible to access these via `hovertemplate` in some way?

We can make that possible, yes!

nicolaskruchten:
Thanks! The `hoverinfo` thing is pretty important, the `hovertemplate` is more of a nice-to-have. I'd say more generally that as a rule, anything that is accessible to `hoverinfo` should be accessible to `hovertemplate`. Should I create a separate issue for that?

Issue #3816, opened 2019-04-29
texttemplate


nicolaskruchten:
This idea keeps coming up as we discuss various questions... Here's an issue for it!

We should have a `texttemplate` attribute wherever we accept `text` or `textinfo`, with similar semantics as `hovertemplate`.

antoinerg:
Ok so are the affected traces/components:
```bash
$ find . -name "attributes.js" -exec ack -l "\s(textinfo|text):" {} \;
src/traces/box/attributes.js
src/traces/funnel/attributes.js
src/traces/violin/attributes.js
src/traces/pointcloud/attributes.js
src/traces/sunburst/attributes.js
src/traces/funnelarea/attributes.js
src/traces/scatterpolar/attributes.js
src/traces/contourcarpet/attributes.js
src/traces/splom/attributes.js
src/traces/indicator/attributes.js
src/traces/waterfall/attributes.js
src/traces/scatterpolargl/attributes.js
src/traces/mesh3d/attributes.js
src/traces/histogram/attributes.js
src/traces/ohlc/attributes.js
src/traces/scattergeo/attributes.js
src/traces/candlestick/attributes.js
src/traces/cone/attributes.js
src/traces/scattergl/attributes.js
src/traces/surface/attributes.js
src/traces/bar/attributes.js
src/traces/choroplethmapbox/attributes.js
src/traces/pie/attributes.js
src/traces/streamtube/attributes.js
src/traces/volume/attributes.js
src/traces/barpolar/attributes.js
src/traces/scatterternary/attributes.js
src/traces/isosurface/attributes.js
src/traces/choropleth/attributes.js
src/traces/scattercarpet/attributes.js
src/traces/contour/attributes.js
src/traces/densitymapbox/attributes.js
src/traces/scattermapbox/attributes.js
src/traces/scatter/attributes.js
src/traces/scatter3d/attributes.js
src/traces/heatmap/attributes.js
src/components/annotations3d/attributes.js
src/components/colorbar/attributes.js
src/components/annotations/attributes.js
```

etpinard:
More like:

```js
// pie-like traces
src/traces/pie/attributes.js
src/traces/sunburst/attributes.js
src/traces/funnelarea/attributes.js

// bar-like
src/traces/bar/attributes.js
src/traces/funnel/attributes.js
src/traces/waterfall/attributes.js

// scatter-like
src/traces/scatter/attributes.js
src/traces/scatterpolar/attributes.js
[...]

PR #3963, opened 2019-06-14
fixup hoverinfo & hovertemplate initial, delta and final flags for waterfalls

archmoj:
A follow up of #3954, #3958 and to address #3960.
The `initial`, `delta` and `final` displayed in the `waterfall` hover now could be controlled and turned off using `hovertemplate` and `hoverinfo`.

Also to note for consistency the initial values and delta from `base` is also displayed over the total bars.  

[Codepen before](https://codepen.io/MojtabaSamimi/pen/OeMLEr?editors=0010)
[Codepen after](https://codepen.io/MojtabaSamimi/pen/BgjLxw?editors=1000)

@plotly/plotly_js 
cc: @nicolaskruchten

### DIFF
--- a/src/traces/waterfall/hover.js
+++ b/src/traces/waterfall/hover.js
@@ -28,7 +28,7 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     var vAxis = isHorizontal ? pointData.xa : pointData.ya;
 
     function formatNumber(a) {
-        return (a === undefined) ? '' : hoverLabelText(vAxis, a);
+        return hoverLabelText(vAxis, a);
     }
 
     // the closest data point
@@ -37,20 +37,16 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
 
     var size = (di.isSum) ? di.b + di.s : di.rawS;
 
-    if(di.isSum) {
-        point.final = undefined;
-        point.initial = undefined;
-        point.delta = size - di.b;
-    } else {
+    if(!di.isSum) {
         point.initial = di.b + di.s - size;
         point.delta = size;
         point.final = point.initial + point.delta;
-    }
 
-    var v = formatNumber(Math.abs(point.delta));
-    point.deltaLabel = size < 0 ? '(' + v + ')' : v;
-    point.finalLabel = formatNumber(point.final);
-    point.initialLabel = formatNumber(point.initial);
+        var v = formatNumber(Math.abs(point.delta));
+        point.deltaLabel = size < 0 ? '(' + v + ')' : v;
+        point.finalLabel = formatNumber(point.final);
+        point.initialLabel = formatNumber(point.initial);
+    }
 
     var hoverinfo = di.hi || trace.hoverinfo;
     var text = [];
@@ -60,18 +56,20 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
 
         var hasFlag = function(flag) { return isAll || parts.indexOf(flag) !== -1; };
 
-        if(hasFlag('final') && point.finalLabel !== '') {
-            text.push(point.finalLabel);
-        }
-        if(hasFlag('delta') && point.deltaLabel !== '') {
-            if(size < 0) {
-                text.push(point.deltaLabel + ' ' + DIRSYMBOL.decreasing);
-            } else {
-                text.push(point.deltaLabel + ' ' + DIRSYMBOL.increasing);
+        if(!di.isSum) {
+            if(hasFlag('final')) {
+                text.push(point.finalLabel);
             }
-        }
-        if(hasFlag('initial') && point.initialLabel !== '') {
-            text.push('Initial: ' + point.initialLabel);
+            if(hasFlag('delta')) {
+                if(size < 0) {
+                    text.push(point.deltaLabel + ' ' + DIRSYMBOL.decreasing);
+                } else {
+                    text.push(point.deltaLabel + ' ' + DIRSYMBOL.increasing);
+                }
+            }
+            if(hasFlag('initial')) {
+                text.push('Initial: ' + point.initialLabel);
+            }
         }
     }
 
--- a/src/traces/waterfall/hover.js
+++ b/src/traces/waterfall/hover.js
@@ -38,17 +38,43 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
     var size = (di.isSum) ? di.b + di.s : di.rawS;
 
     if(!di.isSum) {
-        // format delta numbers:
-        if(size > 0) {
-            point.extraText = formatNumber(size) + ' ' + DIRSYMBOL.increasing;
-        } else if(size < 0) {
-            point.extraText = '(' + (formatNumber(-size)) + ') ' + DIRSYMBOL.decreasing;
-        } else {
-            return;
-        }
-        // display initial value
-        point.extraText += '<br>Initial: ' + formatNumber(di.b + di.s - size);
+        point.initial = di.b + di.s - size;
+        point.delta = size;
+        point.final = point.initial + point.delta;
+    } else {
+        point.final = size;
+        point.initial = di.b;
+        point.delta = point.final - point.initial;
     }
+
+    point.finalLabel = formatNumber(point.final);
+    point.deltaLabel = formatNumber(Math.abs(point.delta));
+    point.initialLabel = formatNumber(point.initial, 1);
+
+    var hoverinfo = di.hi || trace.hoverinfo;
+    var text = [];
+    if(hoverinfo && hoverinfo !== 'none' && hoverinfo !== 'skip') {
+        var isAll = (hoverinfo === 'all');
+        var parts = hoverinfo.split('+');
+
+        var hasFlag = function(flag) { return isAll || parts.indexOf(flag) !== -1; };
+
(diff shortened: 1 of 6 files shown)


nicolaskruchten:
This is cool, but there are some differences between the final text in the on-plot text and the in-hover text:

* the triangle is missing in the on-plot text
* the negative amounts are in (accounting-format) in the in-hover text

@etpinard what's the right way forward here?

image

archmoj:
> This is cool, but there are some differences between the final text in the on-plot text and the in-hover text:
> 
> * the triangle is missing in the on-plot text
> * the negative amounts are in (accounting-format) in the in-hover text
> 
> @etpinard what's the right way forward here?
> 
> image

Good call. 
Also to note that currently we only display delta values in accounting style. 
Should we change anything there feel free to comment here or open a new issue.

etpinard:
> This is cool, but there are some differences between the final text in the on-plot text and the in-hover text:

Oh right, we could have done a better job explaining the `textinfo` behavior to you in https://github.com/plotly/plotly.js/pull/3790 (part of `1.48.0`).  Our apologies.

That said, (1) nothing guarantees than `textinfo` flags and `hoverinfo` flags ended up rendering the same e.g. we should probably always show more precision on-hover than on-graph. (2) Having those `Initial: ` and `Final: ` prefixes on-graph we have made most text item scale down (to fit the bar dimensions) making the number smaller than expected for most users. 

Come to think of it, the currently behavior isn't that bad as I doubt that most users will use a  `textinfo` with more than 1 flags. Oh well, `texttemplate` (ref: https://github.com/plotly/plotly.js/issues/3816) should make everyone happy.

src/traces/waterfall/hover.js
```diff
@@ -38,18 +38,44 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
...
+        var parts = hoverinfo.split('+');
+
+        var hasFlag = function(flag) { return isAll || parts.indexOf(flag) !== -1; };
+
+        if(hasFlag('final') && point.finalLabel !== '') {
+            text.push('Final: ' + point.finalLabel);
+        }
+        if(hasFlag('delta') && point.deltaLabel !== '') {
+            if(size < 0) {
+                text.push('(' + point.deltaLabel + ') ' + DIRSYMBOL.decreasing);
```
etpinard: this here should be the `deltaLabel` value so that 

```
hoverinfo: 'delta'
```

and

```
hovertemplate: '%{delta}'
```

render the same.

etpinard: As for `Final: ` and `Initial: ` prefixes, they should **not** be part of `finalLabel` / `initalLabel`.

See example:

https://codepen.io/etpinard/pen/ewJNVJ

--------

In fact, those prefixes should only be there when there are more than 1 of 'initial', 'delta' and 'final' flags.

etpinard: In brief, _`xLabel`_ should have the default-formatted label text for _`x`_.

I'd argue that `if(size < 0)` logic for delta should be considered _formatting_, but the `DIRSMBOL` suffixes should **not**.

So in brief,

v = formatNumber(Math.abs(point.delta));
point.deltaLabel = size < 0 ? '(' + v + ')' : v;

// ...

    if(size < 0) {
                text.push(point.deltaLabel + DIRSYMBOL.decreasing))
      } else {
                text.push(point.deltaLabel +  DIRSYMBOL.increasing);
            }

test/jasmine/tests/waterfall_test.js
```diff
@@ -1379,13 +1379,72 @@ describe('waterfall hover', function() {
...
+            Plotly.plot(gd, mock)
+            .then(_hover)
+            .then(function() {
+                expect(d3.selectAll('g.hovertext').size()).toBe(0);
+            })
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('should turn on percentages with hoveinfo all', function(done) {
```
etpinard: Looks like a `<ctrl-c><ctrl-v>` gone bad, there are no _percentages_ here :smirk:

test/jasmine/tests/waterfall_test.js
```diff
@@ -1424,7 +1486,7 @@ describe('waterfall hover', function() {
             })
             .then(function() {
                 assertHoverLabelContent({
-                    nums: '2.2\n4.4 ▲\nInitial: −2.2',
+                    nums: '2.2\nFinal: 2.2\n4.4 ▲\nInitial: −2.2',
```
etpinard: Hmm. So wait `Final: ` is new?

test/jasmine/tests/waterfall_test.js
```diff
@@ -1460,31 +1522,31 @@ describe('waterfall hover', function() {
             .then(function() {
                 var out = _hover(gd, 0, 1000.5, 'closest');
                 expect(out.yLabelVal).toEqual(1002.201);
-                expect(out.extraText).toEqual(undefined);
+                expect(out.extraText).toEqual('Final: $1,002.201m<br>$2.2m ▲<br>Initial: $1,000.001m');
```
etpinard: Why did these assertions change?

test/jasmine/tests/waterfall_test.js
archmoj: Good eye! It was 2am.

test/jasmine/tests/waterfall_test.js
archmoj: We did not fill any extra text for blue bars (totals). But it is pretty useful to display this info for them for example when there is a `base` other than zero.

etpinard: Ok, but what does _initial_ and _final_ refer to for total bars?

archmoj: The base and top of the bar.

test/jasmine/tests/waterfall_test.js
archmoj: Yes.

test/jasmine/tests/waterfall_test.js
etpinard: I don't think base/top here represent _inital_ and _final_ values.

test/jasmine/tests/waterfall_test.js
etpinard: Was this a requirement anywhere?

archmoj: No. And we could possibly drop that. 
But that make it clear what the value is pointing to.

test/jasmine/tests/waterfall_test.js
archmoj: Could be displayed in [codepen after](https://codepen.io/MojtabaSamimi/pen/KjdxgR?editors=0010).

nicolaskruchten: I agree that initial and final don't make sense in 'total' cases

archmoj:
This is ready for the second review.

src/traces/waterfall/hover.js
```diff
@@ -37,19 +37,46 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
...
-        if(size > 0) {
-            point.extraText = formatNumber(size) + ' ' + DIRSYMBOL.increasing;
-        } else if(size < 0) {
-            point.extraText = '(' + (formatNumber(-size)) + ') ' + DIRSYMBOL.decreasing;
-        } else {
-            return;
+    if(di.isSum) {
+        point.final = undefined;
+        point.initial = undefined;
+        point.delta = size - di.b;
```
etpinard: Hmm. I'm not sure we can call this a "delta" . `point.y` is probably best for the total bars.

test/jasmine/tests/waterfall_test.js
```diff
@@ -1424,7 +1486,7 @@ describe('waterfall hover', function() {
             })
             .then(function() {
                 assertHoverLabelContent({
-                    nums: '2.2\n4.4 ▲\nInitial: −2.2',
+                    nums: '2.2\n2.2\n4.4 ▲\nInitial: −2.2',
```
etpinard: Why do we have an extra `2.2` now?

archmoj: In respect to https://github.com/plotly/plotly.js/pull/3963#discussion_r293856811 displaying delta seems to be useful especially when there is a base; also when hovering using `hoverinfo`: `delta`.

archmoj: The extra 2.2 here is the `final` value which is the same as `y`.

src/traces/waterfall/hover.js
archmoj: In respect to https://github.com/plotly/plotly.js/pull/3963#discussion_r293856811 displaying delta seems to be useful especially when there is a base; also when hovering using `hoverinfo`: `delta`.

archmoj: To make it similar to current behaviour, this option is dropped in https://github.com/plotly/plotly.js/pull/3963/commits/8432d8f770d912d90532f151d3abbf07e138ba23.

test/jasmine/tests/waterfall_test.js
archmoj: Addressed in https://github.com/plotly/plotly.js/pull/3963/commits/0a588366201a20d2d90483d19e42f89e06555126.

etpinard:
Very nice implementation :dancer:

Issue #3962, opened 2019-06-13
"Uncaught TypeError" during hover over while zoomed into boxplot


wunschsa-lilly:
Hello Plotly Community!

Recently I have tried developing a boxplot merged with a scatter plot locally. I noticed performance problems and increased memory usage when I would zoom in to specific areas of the chart and tried hovering over the areas. I also found that the text displayed next to data points of these zoomed in regions appeared in a different format. Upon inspecting the console in chrome, I have come to find out that an error was being thrown each time this obscure hover event occurred. Below is a screen shot of the error from the console:

![image](https://user-images.githubusercontent.com/49691260/59470183-d14fa700-8deb-11e9-86f1-310ddcdc9aee.png)

The number in the top left increases quickly as you continue to hover over the data, which in turn causes the performance issues. I figured this was a problem with my code or React, but when I tried the Javascript example online, I was able to replicate the same error.  The Code Pen example below is the simple example that you can also find on the Plotly JS website for boxplots (https://plot.ly/javascript/box-plots/).

**How to Replicate the Error:**  If you zoom into this plot and it does not show the Q1 box value, the error will occur. If the Q1 box value is present, then the error does not exist and the plot functions normally. 

When Q1 is zoomed out of the plot (causes error):
![image](https://user-images.githubusercontent.com/49691260/59471781-132f1c00-8df1-11e9-9b79-db0e76e2a425.png)

When Q1 is NOT zoomed out of the plot (no error):
![image](https://user-images.githubusercontent.com/49691260/59471839-45d91480-8df1-11e9-80cb-d9b7681bf02a.png)

I was able to fix this problem in the plotly.js script (starts at while loop on line 91003). It occurs because an "undefined" value is passed as an array instead of the minimum Q1 value. 

Let me know if there is any further information you need on my end to assist with this small bug.

Code Pen: https://codepen.io/wunschsa/pen/gNaRQP?editors=1111

Plotly is an amazing tool, thank you for the capability.

etpinard:
Reproducible example is zoomed view: https://codepen.io/etpinard/pen/XLXWOY?editors=1111

Thanks very much for reporting @wunschsa-lilly !!

etpinard:
Related: https://github.com/plotly/plotly.js/pull/3889

etpinard:
Looks like this was working fine back in `1.45.2`: https://codepen.io/etpinard/pen/YowGmZ?editors=1011

[...]

PR #3889, opened 2019-05-22
Fix hoverAvoidOverlaps for zoomed-in violin traces

etpinard:
fixes #https://github.com/plotly/plotly.js/issues/3875

before: https://codepen.io/etpinard/pen/wberrP?editors=1010
after: https://codepen.io/etpinard/pen/NVXEOw?editors=1010

cc @plotly/plotly_js

PR #3965, opened 2019-06-14
More zoomed-in box/violin hover labels fixes 

etpinard:
@plotly/plotly_js This presents a few commits that alters (mostly fixes :smirk: ) the box/violin hover behaviour when their corresponding value axis range is within the box/violin q1/q3 values. This PR is essentially generalises https://github.com/plotly/plotly.js/pull/3889

Consider,

```js
Plotly.newPlot(gd, [{
  type: 'box',
  y: [1,2,2,3]
}], {
  yaxis: {range: [1.6, 2.4]},
  hovermode: 'x'
})
```

Currently on master (ever since #3645, released in `v1.45.3`), hovering over the box gets us an exception on this line

https://github.com/plotly/plotly.js/blob/06901623e988b7ad036207a905a4a032da186357/src/components/fx/hover.js#L1187

as reported in https://github.com/plotly/plotly.js/issues/3962

---

Commit https://github.com/plotly/plotly.js/commit/a4b771c6cbb4f17319e35dcaca78d6a71bd9aed1 is enough to _fix_ the problem and matches the pre-`1.45.3` behaviour:

Peek 2019-06-14 17-02

but notice that `min` and `q3` are shown and `q1` and `max` are not, which looked odd to me.

---

Commit https://github.com/plotly/plotly.js/commit/1634ec80148d8c5d26933dd48a966ff21b298473 made `q1` and `q3` show instead:

Peek 2019-06-14 17-05

which I think is better as we show both "innermost" values, but maybe we could do even better.

---

Commit https://github.com/plotly/plotly.js/commit/ce5e5f96d58e8f8b9a1f8947e7b38344d5507977 made all box labels show i.e. `min`, `q1`, `median`, `q3` and `max` while avoiding overlaps on the y-axis bounds:

Peek 2019-06-14 17-08

which I feel matches similar scenarios (e.g zoomed-in stack bars [codepen](https://codepen.io/etpinard/pen/bPEqQL)) better, but please let me know if you disagree.


antoinerg:
Thank you for fixing #3962 @etpinard !

As far as I'm concerned, this PR deserves a dancer as it improves things. However, I'll let @nicolaskruchten and/or @alexcjohnson comment on the decision of showing all boxes (ie. https://github.com/plotly/plotly.js/commit/ce5e5f96d58e8f8b9a1f8947e7b38344d5507977).

alexcjohnson:
Yep, much better! 💃

Does look a little weird to have the arrows pointing right at the edge of the plot area, but of course this is what we do in other contexts where you're hovering on something off-plot. So maybe at some point we can alter that somehow to make it clearer they're not actually right there. Maybe as simple as getting rid of the arrow altogether, so these ones only have the box?

etpinard:
> Maybe as simple as getting rid of the arrow altogether, so these ones only have the box?

Follow-up in https://github.com/plotly/plotly.js/issues/3985

etpinard:
Ok, let's start merging things for `1.49.0`

PR #3967, opened 2019-06-17
Split and reuse functions reside inside index files of scattergl scatterpolargl and splom traces

archmoj:
Currently the `index` files of all traces except `scattergl`, `scatterpolargl` & `splom` are compact and only declare different functionalities of the trace.

Having similar pattern in `scattegl` and other two can help improve the readability & requiring of different parts (e.g. when requiring parts of `scattergl` in `scatterpolargl`).

In this PR the functions inside the `index` files of the three are separated and moved to `calc`, `plot`, `select`, `hover`, `helpers` and `update_scene` files.

@plotly/plotly_js


src/traces/scattergl/style.js
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+'use strict';
+
+var Lib = require('../../lib');
+var Color = require('../../components/color');
+
+var DESELECTDIM = require('../../constants/interactions').DESELECTDIM;
+
+function styleTextSelection(cd) {
```
etpinard: Hmm, yeah this one isn't ideal. It is under a `style.js`, but it's not really a `_module.style` method. Ok if we move this thing to `./helpers.js` instead?

src/traces/splom/style.js
```diff
@@ -0,0 +1,34 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Lib = require('../../lib');
+var calcColorscale = require('../scatter/colorscale_calc');
+var convertMarkerStyle = require('../scattergl/convert').markerStyle;
+
+function editStyle(gd, cd0) {
```
etpinard: Same here, I'm not this belong in a `style.js` file. I'd be ok we either:

- moving this logic to `splom/index.js`
- renaming this file to `edit_style.js`

etpinard: Same here, I'm not this belong in a `style.js` file. I'd be ok we either:

- moving this logic to `splom/index.js`
- renaming this file to `edit_style.js`

etpinard: Same here, I'm not this belong in a `style.js` file. I'd be ok we either:

- moving this logic to `splom/index.js`
- renaming this file to `edit_style.js`

etpinard: Same here, I'm not this belong in a `style.js` file. I'd be ok we either:

- moving this logic to `splom/index.js`
- renaming this file to `edit_style.js`

src/traces/scattergl/index.js
```diff
@@ -970,13 +20,13 @@ module.exports = {
...
-    selectPoints: selectPoints,
+    calc: require('./calc'),
+    plot: require('./plot'),
+    hoverPoints: hover.hoverPoints,
+    selectPoints: require('./select'),
 
-    sceneUpdate: sceneUpdate,
-    calcHover: calcHover,
+    sceneUpdate: require('./scene_update'),
+    calcHover: hover.calcHover,
```
etpinard: No need to export these off `ScatterGl` anymore (I think).

etpinard: No need to export these anymore (I think).

etpinard:
Thanks for doing this @archmoj !!

I like the way you split the cross-trace plot logic into an `update_scene.js` file. That's probably something we'll standardise across all regl traces old (parcoords) and upcoming (gl3d).

src/traces/scattergl/index.js
archmoj: Good point!
Done in bb12f68.

src/traces/splom/style.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/3967/commits/e27ea9c8b482c87f2918702f0f7f7770fd88a087.

etpinard:
Nicely done @archmoj !!

:dancer:

Issue #3106, opened 2018-10-15
Plotly relayout autorange fails with automargins.


justnick21:
When `fixLabelOverlaps` is called (axes.js:1807) via:     

> Plotly.relayout('plotly', {
>   "xaxis.autorange": true,
>   "yaxis.autorange": true
> });

`tester = drawing.tester.node();` inside `src/components/drawing/index.js.drawing.bBox` fails because there is no tester node. The `js-plotly-tester` is on the page, just not referenced in the drawing object.

Can be fixed by adding `if (!drawing.tester) drawing.makeTester()` on the line above.

I saw some discussion about the performance of the measuring so I thought I would raise as a bug before making a PR.

alexcjohnson:
Thanks @justnick21 - But I haven't managed to reproduce this error. Here's a try based on one of our existing automargin mocks: https://codepen.io/alexcjohnson/pen/wYPdBN?editors=1010
Can you make a codepen, or modify that one, to show the issue?

justnick21:
Sorry I am unable to reproduce this issue in a codepen. Not sure why the key is no longer on that object. Stack trace attached but without spending a lot of time reproducing this in a pen I am afraid I can't give any more details

`TypeError: Cannot read property 'node' of undefined
    at Object.push../node_modules/plotly.js/src/components/drawing/index.js.drawing.bBox (index.js:926)
    at SVGGElement.<anonymous> (axes.js:1803)
    at plotly.js:21021
    at d3_selection_each (plotly.js:21028)
    at Array.d3_selectionPrototype.each (plotly.js:21020)
    at fixLabelOverlaps (axes.js:1798)
    at Object.push../node_modules/plotly.js/src/lib/index.js.lib.syncOrAsync (index.js:401)
    at drawLabels (axes.js:1961)
    at Object.push../node_modules/plotly.js/src/plots/cartesian/axes.js.axes.doTicksSingle (axes.js:2132)
    at axes.js:1480
    at Object.push../node_modules/plotly.js/src/lib/index.js.lib.syncOrAsync (index.js:401)
    at Object.push../node_modules/plotly.js/src/plots/cartesian/axes.js.axes.doTicks (axes.js:1477)
    at Object.push../node_modules/plotly.js/src/plot_api/subroutines.js.exports.doTicksRelayout (subroutines.js:480)
    at doTicks (plot_api.js:1638)
[...]

PR #3971, opened 2019-06-18
ensure that tester property of drawing object exists

wryczko:
When I tried to customize modeBarButtons by using imported objects (import PlotlyButtons from 'plotly.js/src/components/modebar/buttons') I got following error:

`TypeError: Cannot read property 'node' of undefined at Object.push../node_modules/plotly.js/src/components/drawing/index.js.drawing.bBox (index.js:926)`

I've noticed this issue was once explained here https://github.com/plotly/plotly.js/issues/3106, but no actions have been taken. I can confirm, that this issue exists, but I have reproduced it differently.

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -942,6 +942,9 @@ drawing.bBox = function(node, inTester, hash) {
     if(inTester) {
         testNode = node;
     } else {
+        if(!drawing.tester) {
+            drawing.makeTester();
+        }
         tester = drawing.tester.node();
 
         // copy the node to test into the tester


etpinard:
Thanks very much for the PR!

Before we start thinking about merging it, could you please share a reproducible example of the bug you're trying to fix? I understand that your issue appears similar to https://github.com/plotly/plotly.js/issues/3106, but we haven't been able to reproduce that one (yet). Once you figured out how to replicate the problem, we'll be able to lock your fix down in a test.

Cheers!

etpinard:

If someone can replicate the problem in a codepen, please open an issue!

Better yet, if someone can write a failing test case and then write a patch fixing it, please open a PR!

federicomassa:
This happened to me and independently found the same way to solve it. The fix works for me, but when I tried to replicate it on CodePen I didn't manage to, the issue just didn't show up.

PR #3972, opened 2019-06-18
 responsive handler: do not resize if gd is hidden

### LINKED ISSUE #3866: Bug: Resize on hidden graph divs causes Error
Discussion: https://community.plot.ly/t/dynamically-change-responsive-layout/23002
Repo sample: https://codesandbox.io/s/9qy1p1k8jy

Error:
```
Error: Resize must be passed a displayed plot div element.
    at eval (plotly-basic.js:59636)
    at new Promise (<anonymous>)
    at Object.plots.resize (plotly-basic.js:59629)
    at gd._responsiveChartHandler (plotly-basic.js:41558)
```

antoinerg:
Fixes #3866 using approach number 2 from comment https://github.com/plotly/plotly.js/issues/3866#issuecomment-503226330

### DIFF
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -178,15 +178,11 @@ function plot(gd, data, layout, config) {
         gd.calcdata[i][0].trace = gd._fullData[i];
     }
 
-    function isHidden(gd) {
-        var display = window.getComputedStyle(gd).display;
-        return !display || display === 'none';
-    }
     // make the figure responsive
     if(gd._context.responsive) {
         if(!gd._responsiveChartHandler) {
             // Keep a reference to the resize handler to purge it down the road
-            gd._responsiveChartHandler = function() { if(!isHidden(gd)) Plots.resize(gd).catch(Lib.noop); };
+            gd._responsiveChartHandler = function() { if(!Lib.isHidden(gd)) Plots.resize(gd); };
 
             // Listen to window resize
             window.addEventListener('resize', gd._responsiveChartHandler);
--- a/src/plot_api/plot_api.js
+++ b/src/plot_api/plot_api.js
@@ -178,11 +178,15 @@ function plot(gd, data, layout, config) {
         gd.calcdata[i][0].trace = gd._fullData[i];
     }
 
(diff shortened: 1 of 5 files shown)


src/plot_api/plot_api.js
```diff
@@ -178,11 +178,15 @@ function plot(gd, data, layout, config) {
         gd.calcdata[i][0].trace = gd._fullData[i];
     }
 
+    function isHidden(gd) {
+        var display = window.getComputedStyle(gd).display;
+        return !display || display === 'none';
+    }
```
etpinard: Isn't this the same helpers as in `Plots.resize`? 

Can we make this a `Lib` function?

src/plot_api/plot_api.js
```diff
@@ -178,11 +178,15 @@ function plot(gd, data, layout, config) {
...
     // make the figure responsive
     if(gd._context.responsive) {
         if(!gd._responsiveChartHandler) {
             // Keep a reference to the resize handler to purge it down the road
-            gd._responsiveChartHandler = function() { Plots.resize(gd); };
+            gd._responsiveChartHandler = function() { if(!isHidden(gd)) Plots.resize(gd).catch(Lib.noop); };
```
etpinard: Do we _really_ need to `.catch` here?

etpinard:
Thanks for taking this on @antoinerg !

Pinning this for `v1.49.0`

src/plot_api/plot_api.js
antoinerg: No, we don't! The whole point is to not ignore possible rejected promises. Thanks, @etpinard for noticing this.

Fixed in 913bb6877df1e7668748ad4b3e535e38bbf25e6b

src/plot_api/plot_api.js
antoinerg: Yes, good idea!

Done in 913bb6877df1e7668748ad4b3e535e38bbf25e6b

etpinard:
thanks again for doing this @antoinerg :dancer:

etpinard:
(let's start merging things for 1.49.0)

PR #3977, opened 2019-06-21
Fix emitting event "plotly_relayouting" with wrong data

maun:
Currently "plotly_relayouting" events are emitted when creating a zoombox. These events use the updates to the layout which would be applied when zooming. Since no relayouting takes place until the zoombox is applied I removed the event.

### DIFF
--- a/src/plots/cartesian/dragbox.js
+++ b/src/plots/cartesian/dragbox.js
@@ -410,7 +410,6 @@ function makeDragBox(gd, plotinfo, x, y, w, h, ns, ew) {
 
         updateZoombox(zb, corners, box, path0, dimmed, lum);
         computeZoomUpdates();
-        gd.emit('plotly_relayouting', updates);
         dimmed = true;
     }
 


maun:
To fix the test I could remove them or set the expected relayouting count to zero. An alternative would be to change the event to something else, eg "plotly_zoombox-resized". This would be enable tests like ('handles y-only to xy back to y-only in single zoombox drag motion').

alexcjohnson:
@maun `plotly_relayouting` events are not intended to represent actual changes to the layout - that's what `plotly_relayout` events are for. These events are intended for live updates before the change has actually been applied to the plot. See https://github.com/plotly/plotly.js/pull/3888 and https://github.com/plotly/plotly.js/issues/2082

maun:
@alexcjohnson I implemented my own range slider and wanted to synchronise the movement with the panning on the graph. This works by using `plotly_relayouting `, but fails when using the zoombox since this also emits this event.

jonmmease:
@alexcjohnson I would agree with @maun that this behavior is a bit surprising.  The main usecase that I've had in mind for `plotly_relayouting` is 2D Histogram resampling (http://datashader.org/ from Python, but could also be pure JavaScript).  During pan actions I would want to be able to process the entire viewport at each moment during the pan (not only at the end when `plotly_relayout` is emitted), but it would not make sense for this usecase to process the zoombox region only during box zoom.

Is there anything in the event data to help distinguish between pan and box zoom?

maun:
@jonmmease Unfortunately not, the event is the same. If you agree to change this somehow by renaming the zoombox event or adding another property I would update my PR.

alexcjohnson:
That makes sense, thanks for clarifying. I don't think we want to remove zoombox from this event entirely, I can still imagine uses for it. But labeling it so you could choose to ignore zoombox or whichever ones you don't want seems like a good solution. It's a little weird as we only have the relayout object, but can we just add an `interactionType` key to that object? @etpinard @antoinerg does that seem reasonable?

jonmmease:
> can we just add an interactionType key to that object?

How about `_interactionType` so that it's clear that this isn't part of the schema?

etpinard:
Wouldn't

```js
gd.on('plotly_relayouting', function(d) {
  gd._fullLayout.dragmode // => 'pan' or 'zoom'
})
```

also give you the right answer?

maun:
@etpinard Just tried this, for me always "pan" is returned. Probably because I set the default dragmode to panning.

melecptp:
Hello everyone,
This discussion is quite old and I do not know if this issue is still relevant. 
But I need to distinguish zoom events from pan events during relayouting and having a `interactionType` key would be very helpful.
I agree with @maun, the dragmode does not really solve the problem since it only gives the selected mode in the modebar. But one could also pan the figure by grabbing the x/y axis even if the activated mode is zoom.

PR #3978, opened 2019-06-21
trace type `indicator`

### LINKED ISSUE #3659: Indicator trace type (gauge+bignum+bullet)
Let's tackle three of the ones from #2221 in a single `indicator` trace type.

antoinerg:
This PR introduces a new trace type `indicator` and closes #3659

Codepen to animate mocks: https://codepen.io/antoinerg/full/dBZdpY (using bundle built off 1a991d5)

- [x] Improve test coverage of mocks (`gauge.axis.visible`)
- [x] Provide codepen to animate aforementionned mocks


lib/index-finance.js
```diff
@@ -18,7 +18,8 @@ Plotly.register([
     require('./ohlc'),
     require('./candlestick'),
     require('./funnel'),
-    require('./waterfall')
+    require('./waterfall'),
+    require('./indicator')
```
archmoj: What is the best drawing order for this trace?
Should it be on top of everything else?

etpinard: I'd say `indicator` traces should be above all other traces.

Right now,

https://github.com/plotly/plotly.js/pull/3978/files#diff-2941ab69a12080c0633ff4ac8ea3aa83R3783

@antoinerg has it above all other SVG traces but below the WebGL traces. Perhaps we should place `_indicatorlayer` in `fullLayout._toppaper` next to colorbars, legends, updatemenus and sliders? This would be the first ever trace in `_toppaper` :smirk:

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+    }
+}));
+
+module.exports = {
+    mode: {
+        valType: 'flaglist',
+        editType: 'calc',
+        role: 'info',
+        flags: ['number', 'delta', 'gauge'],
+        dflt: 'number'
```
etpinard: `description` please.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+            description: [
+                'Set the font used to display the title'
+            ].join(' ')
+        }),
+        editType: 'plot'
+    },
+    number: {
+        valueformat: {
+            valType: 'string',
+            dflt: '.3s',
```
etpinard: It might be nice to be consistent with other `*format` attributes in use currently [e.g. `tickformat` for axis](https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/plots/cartesian/layout_attributes.js#L549-L564) where the `dflt` is left blank and we let [`Axes.tickText`](https://github.com/plotly/plotly.js/blob/master/src/plots/cartesian/axes.js#L934-L1004) handle things.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+            valType: 'enumerated',
+            values: ['top', 'bottom', 'left', 'right'],
+            role: 'info',
+            dflt: 'bottom',
+            editType: 'plot',
+            description: [
+                'Sets the position of delta with respect to the number.'
+            ].join(' ')
+        },
+        relative: {
```
etpinard: Just to document something me and @antoinerg discussed in-person last week.

----------------

One potential addition to `indicator` would be to add a way to show both "absolute" and "relative" delta widgets on the trace same `indicator` trace. 

One way to achieve this would be to add `'delta2'` flag in mode and have something like:

```js
{
  mode: 'number+delta+delta2'
  delta: {},
  delta2: {relative: true}
}
```

OR, replace `delta` with `deltarel` and `deltaabs`

```js
{
  mode: 'number+deltarel+deltaabs'
  deltarel: {},
  deltaabs: {}
}
```

making this `relative` attribute obsolete.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+            valType: 'enumerated',
+            editType: 'plot',
+            role: 'info',
+            dflt: 'angular',
+            values: ['angular', 'bullet'],
+            description: [
+                'Set the shape of the gauge'
+            ].join(' ')
+        },
+        value: extendDeep({}, gaugeBarAttrs, {
```
etpinard: Hmm. I'm not sure I like having `value` both as an attribute name and an attribute container. Maybe `bar` as in `gauge.bar.color` or `pane` (like [highcharts](https://www.highcharts.com/demo/gauge-solid)) would be better?

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+        },
+        editType: 'plot'
+    },
+    delta: {
+        reference: {
+            valType: 'number',
+            role: 'info',
+            editType: 'calc',
+            description: [
+                'Sets the reference value to compute the delta.'
```
etpinard: Maybe add something in the `description` about what happens when `reference` isn't set? 

In the case, it uses the "last" value as reference, correct?

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+        ].join(' ')
+    },
+    // position
+    domain: domainAttrs({name: 'indicator', trace: true, editType: 'calc'}),
+
+    vmin: {
+        valType: 'number',
+        editType: 'calc',
+        role: 'info',
+        dflt: 0,
```
etpinard: Interesting `dflt` here. I guess this make sense for most cases (e.g. for money), but maybe setting a min value could be called _too magical_.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+                'which is similar to those of Python. See',
+                'https://github.com/d3/d3-format/blob/master/README.md#locale_format'
+            ].join(' ')
+        },
+        font: extendFlat({}, textFontAttrs, {
+            description: [
+                'Set the font used to display main number'
+            ].join(' ')
+        }),
+        suffix: {
```
etpinard: I suspect will need a `prefix` attribute too, e.g for users wanting to show `$` before their formatted `value`.

src/traces/indicator/defaults.js
```diff
@@ -0,0 +1,162 @@
...
+    traceOut._hasNumber = traceOut.mode.indexOf('number') !== -1;
+    traceOut._hasDelta = traceOut.mode.indexOf('delta') !== -1;
+    traceOut._hasGauge = traceOut.mode.indexOf('gauge') !== -1;
+
+    coerce('value');
+    coerce('vmin');
+    coerce('vmax', 1.5 * traceOut.value);
+
+    // Number attributes
+    var auto = [];
```
etpinard: Safer to initialise `auto` as `new Array(2)` here before filling in `auto[0]` and `auto[1]`.

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+                    color: trace.gauge.bordercolor,
+                    width: trace.gauge.borderwidth
+                },
+                thickness: 1
+            };
+        }
+
+        // Prepare angular gauge layers
+        var data = cd.filter(function() {return isAngular;});
+        var angularGauge = plotGroup.selectAll('g.angular').data(data);
```
etpinard: `cd` here has always `length===1`, correct?

In this case, 

```js
var angularGauge = plotGroup.selectAll('g.angular').data(isAngular ? cd : []);
```

would probably be more readable and make the (slightly confusing) `data` variable obsolete.

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+    ax._id = 'xbulletaxis';
+    ax.domain = [bulletLeft, bulletRight];
+    ax.setScale();
+
+    vals = Axes.calcTicks(ax);
+    transFn = Axes.makeTransFn(ax);
+    tickSign = Axes.getTickSigns(ax)[2];
+
+    shift = size.t + size.h;
+    if(ax.visible) {
```
etpinard: I don't think `visible` is part of the `gauge.axis` attribute currently. Might be a nice addition.

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+        s
+            .attr('width', function(d) { return Math.max(0, ax.c2p(d.range[1] - d.range[0]));})
+            .attr('x', function(d) { return ax.c2p(d.range[0]);})
+            .attr('y', function(d) { return 0.5 * (1 - d.thickness) * bulletHeight;})
+            .attr('height', function(d) { return d.thickness * bulletHeight; });
+    }
+
+    // Draw bullet background, steps
+    var boxes = [gaugeBg].concat(trace.gauge.steps);
+    var targetBullet = bullet.selectAll('g.targetBullet').data(boxes);
```
etpinard: No :camel:s in class names, please.

https://stackoverflow.com/questions/1547986/can-i-use-camel-case-in-css-class-names

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+                .each('interrupt', function() { onComplete && onComplete(); })
+                .attrTween('text', function() {
+                    var that = d3.select(this);
+                    var interpolator = d3.interpolateNumber(cd[0].lastY, cd[0].y);
+                    return function(t) {
+                        that.text(fmt(interpolator(t)) + bignumberSuffix);
+                    };
+                });
+        } else {
+            number.text(fmt(cd[0].y) + bignumberSuffix);
```
etpinard: I was expecting `Lib.constrain(cd[0].y, vmin, vmax)` here.

Does `vmin` and `vmax` only have an effect in for `mode` values that include `'gauge'`?

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+
+    var data = cd.filter(function() {return trace.gauge.threshold.value;});
+    var threshold = bullet.selectAll('g.threshold').data(data);
+    threshold.enter().append('g').classed('threshold', true).append('line');
+    threshold.select('line')
+        .attr('x1', ax.c2p(trace.gauge.threshold.value))
+        .attr('x2', ax.c2p(trace.gauge.threshold.value))
+        .attr('y1', (1 - trace.gauge.threshold.thickness) / 2 * bulletHeight)
+        .attr('y2', (1 - (1 - trace.gauge.threshold.thickness) / 2) * bulletHeight)
+        .style('stroke', trace.gauge.threshold.line.color)
```
etpinard: Please use `Color.stroke` 

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/components/color/index.js#L84-L87

whenever you need to set `stroke` for consistency and to guarantee that SVG exports of `indicator` traces will work in Illustrator.

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+            if(value === 0) return '-';
+            return (value > 0 ? trace.delta.increasing.symbol : trace.delta.decreasing.symbol) + deltaFmt(value);
+        };
+        var deltaFill = function(d) {
+            return d.delta >= 0 ? trace.delta.increasing.color : trace.delta.decreasing.color;
+        };
+        var delta = numbers.select('tspan.delta');
+        delta
+            .call(Drawing.font, trace.delta.font)
+            .style('fill', deltaFill)
```
etpinard: Please use `Color.fill`:

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/components/color/index.js#L89-L95

... for similar reasons as `Color.stroke`.

src/snapshot/tosvg.js
```diff
@@ -88,7 +88,7 @@ module.exports = function toSVG(gd, format, scale) {
     // fill whatever container it's displayed in regardless of plot size.
     svg.node().style.background = '';
 
-    svg.selectAll('text')
+    svg.selectAll('text,tspan')
```
etpinard: Please add test in `snapshot_test.js` similar to:

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/test/jasmine/tests/snapshot_test.js#L258-L307

but for `indicator` `<tspan>`s.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+// var plotAttrs = require('../../plots/attributes');
+// var domainAttrs = require('../../plots/domain').attributes;
```
etpinard: :zap: those commented out lines.

src/traces/indicator/plot.js
antoinerg: > Does vmin and vmax only have an effect in for mode values that include 'gauge'?

Initially no, but right now, it's only used for setting the range of the `gauge`. I guess we could also constrain the `number` :thinking:

src/traces/indicator/defaults.js
```diff
@@ -0,0 +1,162 @@
...
+        bignumberFontSize = traceOut.number.font.size;
+    }
+
+    // delta attributes
+    var deltaFontSize;
+    if(traceOut._hasDelta) {
+        coerce('delta.font.color', layout.font.color);
+        coerce('delta.font.family', layout.font.family);
+        coerce('delta.font.size', 'auto');
+        if(traceOut.delta.font.size === 'auto') {
```
etpinard: We probably don't need to handle a special `'auto'` value. Omitting the `dflt` in `*.font.size` attribute declarations, we can handle the case where e.g. here `traceOut.delta.font.size` is `undefined`.

src/traces/indicator/attributes.js
```diff
@@ -0,0 +1,408 @@
...
+            valType: 'string',
+            role: 'info',
+            editType: 'plot',
+            description: [
+                'Sets the title of this indicator.'
+            ].join(' ')
+        },
+        align: {
+            valType: 'enumerated',
+            values: ['left', 'center', 'right'],
```
etpinard: I think `title.align` corresponds to what most users will find useful for `indicator` traces, but note that `align` somewhat conflicts with `textposition` in pie traces

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/traces/pie/attributes.js#L208-L222

and the `(x|y)ref` / `(x|y)anchor` attributes for `layout.title`:

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/plots/layout_attributes.js#L51-L130

and there's `side` in colorbars:

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/plots/layout_attributes.js#L51-L130

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+
+function fitTextInsideCircle(el, radius) {
+    // compute scaling ratio to have text fit within specified radius
+    var textBB = Drawing.bBox(el.node());
+    var elRadius = Math.sqrt((textBB.width / 2) * (textBB.width / 2) + textBB.height * textBB.height);
+    var ratio = radius / elRadius;
+    return [ratio, textBB, radius];
+}
+
+// Draw gauge's min and max in text
```
etpinard: :zap:

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+    handleAxisPositionDefaults(axisIn, axisOut, coerce, axisOptions);
+
+    return axisOut;
+}
+
+function strTranslate(x, y) {
+    return 'translate(' + x + ',' + y + ')';
+}
+
+function strRotate(angle) {
```
etpinard: This thing is only used once. :hocho:

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+        .attr('class', function(d) { return d;})
+        .attr('dx', function(d, i) {
+            // Add padding to the second tspan when it's a one-liner
+            if(i === 1) {
+                var pos = trace.delta.position;
+                if(pos === 'left' || pos === 'right') return 10;
+            }
+            return null;
+        });
+    sel.exit().remove();
```
etpinard: You'll need a `sel.order()` call to make sure that

```js
Plotly.newPlot(gd, [{mode: 'number+delta'}])
```

has the same `<tspan>` order as after

```js
Plotly.restyle(gd, 'mode', 'delta')
Plotly.restyle(gd, 'mode', 'number+delta')
```

test/image/mocks/indicator_bignumber.json
```diff
@@ -0,0 +1,197 @@
...
+          "symbol": "+"
+        },
+        "valueformat": ".1%",
+        "reference": 400,
+        "relative": true
+      },
+      "gauge": {
+        "bgcolor": "rgba(255, 255, 255, 0.25)"
+      },
+      "max": 500,
```
etpinard: `vmax`?

src/traces/indicator/defaults.js
```diff
@@ -0,0 +1,162 @@
...
+    function coerceGaugeAxis(attr, dflt) {
+        return Lib.coerce(axisIn, axisOut, attributes.gauge.axis, attr, dflt);
+    }
+    if(traceOut._hasGauge) {
+        gaugeIn = traceIn.gauge;
+        gaugeOut = Template.newContainer(traceOut, 'gauge');
+        coerceGauge('shape');
+        var isBullet = traceOut._isBullet = traceOut.gauge.shape === 'bullet';
+        if(!isBullet) {
+            coerce('title.align', 'center');
```
etpinard: Doesn't the current behaviour for bullets have essentially `title.align: 'right'`?

src/traces/indicator/plot.js
```diff
@@ -0,0 +1,788 @@
...
+
+        // Position numbers based on mode and set the scaling logic
+        var numbersX, numbersY, numbersScaler;
+        var numbersAlign = trace.align || 'center';
+
+        numbersY = centerY;
+        if(!hasGauge) {
+            numbersX = size.l + position[numbersAlign] * size.w;
+            numbersScaler = function(el) {
+                return fitTextInsideBox(el, 0.9 * size.w, 0.9 * size.h);
```
etpinard: I'm curious why you picked `0.9` as factor here? How bad do things look if we replace it with `1`?

etpinard: Having `0.9` for angular gauge makes sense as we want some sort of padding inside the arc, but for gauge-less `indicator` traces, it might be best to try to fit the `domain` box.

src/traces/indicator/attributes.js
antoinerg: I don't have a strong opinion for this one but having a fixed number of digits with SI suffixes as a default does make sense if we want the number to always fit within the angular gauge regardless of how big or small the value is :thinking:

src/traces/indicator/attributes.js
antoinerg: That's definitely a possibility! 

I think I prefer having `delta2` as it would imply it is to be displayed to the right of `delta`. If we have `deltarel` and `deltaabs`, we'd have to add an attribute to set their position relative to one another. Basically, we'd be trading attribute `relative` for `relativeposition`...

src/traces/indicator/attributes.js
antoinerg: I think `bar` makes sense. Maybe @plotly/plotly_js @nicolaskruchten have another idea?

src/traces/indicator/plot.js
etpinard: ... or alternatively make move `vmin` and `vmax` to the `gauge.*` attribute container

test/jasmine/tests/indicator_test.js
```diff
@@ -0,0 +1,438 @@
...
+        function assertGauge(shape, cnt) {
+            assertElementCnt('g.' + shape, cnt);
+            assertElementCnt('g.' + shape + 'axis', cnt);
+        }
+        function assert(flags) {
+            // flags is an array denoting whether the figure [hasNumber, hasDelta, hasAngular, hasBullet]
+            assertElementCnt('tspan.number', flags[0]);
+            assertElementCnt('tspan.delta', flags[1]);
+            assertGauge('angular', flags[2]);
+            assertGauge('bullet', flags[3]);
```
etpinard: Can we also assert that number, delta, angular and bullet pieces preserve their order after these `restyle` calls below.

You can try using 

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/test/jasmine/assets/custom_assertions.js#L309-L370

to help you out.

src/traces/indicator/attributes.js
antoinerg: When it isn't set, it defaults to the current value.

test/jasmine/tests/indicator_test.js
```diff
@@ -0,0 +1,438 @@
...
+
+describe('Indicator animations', function() {
+    var gd;
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+    afterEach(destroyGraphDiv);
+
+    it('should be able to transition via `Plotly.react`', function(done) {
+        var mock = {data: [{type: 'indicator', value: 100}], layout: {}};
```
etpinard: Can we also check that `indicator` traces with bullet/angular gauges can also be smoothly transitioned using `Plotly.react`?

src/traces/indicator/attributes.js
antoinerg: By the way, I also set `vmax` to 1.5 times the current value. This is going to break for negative values. What would be good defaults for `v(min|max)`?

etpinard:
Awesome work @antoinerg !

I'm a fan of your _`numbersScales`_ abstractions and your clever use of `Drawing.bBox` to place the indicator title. 

Your baselines looks great and they appear to :lock: down a lot of the behaviour. Maybe you could add one more baseline to lock down `indicator` together with `template` and `layout.grid` ?

etpinard:
... oh and @antoinerg maybe you could share a codepen example of `indicator` used `Plotly.react` transition or `Plotly.animate`?

src/traces/indicator/attributes.js
etpinard: > for this one but having a fixed number of digits with SI suffixes as a default does make sense if we want the number to always fit within the angular gauge regardless of how big or small the value is

Good point here. I don't have a strong opinion either.

src/traces/indicator/attributes.js
etpinard: Great, is there a test locking down that behaviour?

src/traces/indicator/defaults.js
antoinerg: > Doesn't the current behaviour for bullets have essentially `title.align: 'right'`?

They do yes. However, it cannot be changed. That's why I thought it shouldn't be coerced. Should I coerce it?

src/traces/indicator/attributes.js
etpinard: Sorry for the confusion, I was under the impression that `vmin` and `vmax` were also used to constraint the `mode: 'number'` text rendering. 

Your `vmin: 0` and `vmax: 1.5 * value` default sounds ok for `mode: 'gauge'`, but then again maybe `vmin` and `vmax` should be under `gauge.vmin` and `gauge.vmax` :thinking:

etpinard:
Oh one more thing,

Peek 2019-06-25 16-29

the "Toggle show closest data on hover" modebar button shouldn't not appear when only `indicator` traces are present on the graph.

src/traces/indicator/defaults.js
etpinard: Could it support `align: 'center'` and `align: 'left'` eventually (i.e. not in this PR)?

src/traces/indicator/attributes.js
antoinerg: > Sorry for the confusion, I was under the impression that `vmin` and `vmax` were also used to constraint the `mode: 'number'` text rendering.

It used to be the case but no more: we now resize the text by measuring its actual size in the viewport.

> Your `vmin: 0` and `vmax: 1.5 * value` default sounds ok for `mode: 'gauge'`, but then again maybe `vmin` and `vmax` should be under `gauge.vmin` and `gauge.vmax`

Yes, let's move `v(min|max)` under `gauge` where they belong.

src/traces/indicator/attributes.js
antoinerg: I'm willing to either rename the attribute to `position` as in `pie` or change the API altogether. 

FYI, the `(title|number).align` attributes were added at the last minute in order to mimic DDK's "data cards" in which the text is left-aligned and position to the left of the domain.

test/image/mocks/indicator_bignumber.json
antoinerg: Cruft that we left over! Thanks for noticing :eyes:

src/traces/indicator/defaults.js
antoinerg: > Could it support `align: 'center'` and `align: 'left'` eventually (i.e. not in this PR)?

My thought was that because the title is outside the domain in `bullet`, the notion of alignment is not well-defined: it's unclear how much whitespace we have on the left (ie. maybe there's another plot there).

src/traces/indicator/plot.js
antoinerg: > Oh one more thing,
> 
> Peek 2019-06-25 16-29
> 
> the "Toggle show closest data on hover" modebar button shouldn't not appear when only `indicator` traces are present on the graph.

What would be the cleanest to achieve that?

src/traces/indicator/attributes.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/attributes.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830 with test to :lock: down the behavior

src/traces/indicator/attributes.js
antoinerg: Added with tests in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/defaults.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Added in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Thanks so much for noticing this! Camelcases removed in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/snapshot/tosvg.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/attributes.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/defaults.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

test/image/mocks/indicator_bignumber.json
antoinerg: Removed in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/plot.js
antoinerg: Now fit the domain box in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

test/jasmine/tests/indicator_test.js
antoinerg: Done in e6d8525ab07c56ca10bdbac6762b06d1d57bf830

src/traces/indicator/attributes.js
etpinard: > Yes, let's move `v(min|max)` under `gauge` where they belong.

:+1:

src/traces/indicator/attributes.js
```diff
@@ -327,6 +340,9 @@ module.exports = {
             description: 'Sets the width (in px) of the border enclosing the gauge.'
         },
         axis: overrideAll({
+            visible: extendDeep({}, axesAttrs.visible, {
```
etpinard: `extendFlat` should suffice here.

src/traces/indicator/defaults.js
```diff
@@ -37,17 +37,18 @@ function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
...
+    var auto = new Array(2);
     var bignumberFontSize;
     if(traceOut._hasNumber) {
         coerce('number.valueformat');
         coerce('number.font.color', layout.font.color);
         coerce('number.font.family', layout.font.family);
-        coerce('number.font.size', 'auto');
-        if(traceOut.number.font.size === 'auto') {
+        coerce('number.font.size');
+        if(!traceOut.number.font.size) {
```
etpinard: That works if `font.size===0` is invalid. Do we accept `font.size===0` elsewhere?

Otherwise `traceOut.number.font.size === undefined` should work too.

src/traces/indicator/plot.js
etpinard: We should add a mock here showing off `gauge.axis.visible: false` for both gauge shapes.

src/traces/indicator/plot.js
etpinard: Merged with https://github.com/plotly/plotly.js/pull/3978#pullrequestreview-254121221

src/traces/indicator/defaults.js
etpinard: Marking as resolved. Follow-up in https://github.com/plotly/plotly.js/pull/3978#pullrequestreview-254682391

src/traces/indicator/attributes.js
etpinard: Or maybe `gauge.range` would be better and more consistent with `gauge.steps[i].range` :thinking:

lib/index-finance.js
antoinerg: Done in 03cb7c3b9b4fbd8cb29889457e0399363bb0c1cf with and :lock: with image test

src/traces/indicator/attributes.js
antoinerg: I went ahead and renamed `gauge.value` to `gauge.bar` in 95de45724b62a7a1c3d22434af838bd8c690cbfe

src/traces/indicator/attributes.js
antoinerg: I chose to replace attributes `v(min|max)` with a single attribute `gauge.axis.range`.

See commit 95de45724b62a7a1c3d22434af838bd8c690cbfe

src/traces/indicator/plot.js
antoinerg: I test this in 95de45724b62a7a1c3d22434af838bd8c690cbfe and it seems to be OK without a `sel.order()`. To be investigated

test/jasmine/tests/indicator_test.js
antoinerg: It seems like the order is preserved as can be seen from the test introduced in commit 95de45724b62a7a1c3d22434af838bd8c690cbfe

src/traces/indicator/attributes.js
antoinerg: Done in 95de45724b62a7a1c3d22434af838bd8c690cbfe

src/traces/indicator/defaults.js
antoinerg: I went with `traceOut.number.font.size === undefined` in 95de45724b62a7a1c3d22434af838bd8c690cbfe

antoinerg:
> the "Toggle show closest data on hover" modebar button shouldn't not appear when only `indicator` traces are present on the graph.

Traces part of category `noHover` do not have hover buttons following commits c796e57e65477f9a283ea8f811085721e439cf55 + aa6ba8fab5769bf56bce125c04f3c3cd3cbc5ff8
Screenshot_2019-06-26_17-30-37

lib/index-finance.js
etpinard: Nice. I'm glad that went smoothly!

src/traces/indicator/attributes.js
etpinard: :ok_hand:

src/traces/indicator/plot.js
etpinard: The test look fine. Thanks for adding those! Something else must up with a upstream `.data()` call.

Oh well. No need to waste time taking a deeper dive at this problem. :ok_hand:

test/jasmine/tests/indicator_test.js
etpinard: Same conclusion as in https://github.com/plotly/plotly.js/pull/3978#discussion_r298265531

Resolved.

antoinerg:
The default for `gauge.bgcolor` is black. We should probably change it otherwise we get a not so pretty:
```javascript
Plotly.newPlot(gd,[{type:'indicator', mode:'gauge', value:10}])
```
newplot - 2019-06-27T132131 367

etpinard:
> Any suggestion?

Oh right. Good eye! We should be `coerce('bgcolor', layoutOut.paper_bgcolor);`

antoinerg:
> ... oh and @antoinerg maybe you could share a codepen example of `indicator` used `Plotly.react` transition or `Plotly.animate`?

Here's the requested Codepen: https://codepen.io/antoinerg/full/dBZdpY

src/traces/indicator/index.js
```diff
@@ -0,0 +1,30 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+module.exports = {
+    moduleType: 'trace',
+    name: 'indicator',
+    basePlotModule: require('./base_plot'),
+    categories: ['svg', 'noOpacity', 'noHover'],
```
archmoj: It could be useful to if this trace had an adjustable opacity.

etpinard: Mojtaba _"Mr. Opacity"_ Samimi

etpinard:
All right. Time to merge this thing!

https://codepen.io/antoinerg/full/dBZdpY is one of the coolest plotly.js codepen in recent memory :1st_place_medal: 

Not bad for @antoinerg 's first `new trace type`.

:dancer: :dancer: :dancer: 

----

@antoinerg feel free to open a new issue about the array-container / `coerce2` + template bugs you noticed. They should be tackled in (a) separate PR(s).

PR #3983, opened 2019-06-25
Fix boxplot with categoryorder

### LINKED ISSUE #3970: Some categoryorder creates invisible extra series for boxplot
Hi,

For boxplot, I've found that `categoryorder` of (total|min|max|sum|mean|median) in boxplot generates invisible data to the plot.

[Original baseline with no categoryorder](https://codepen.io/etpinard/pen/OOZMqW)
[with categoryorder of 'median descending'](https://codepen.io/anon/pen/LKZXRy)

Screenshot. Second plot is the trace with categoryorder.
image

antoinerg:
Closes #3970 

This seems a bit hacky to me to it does fix the issue at hand.


antoinerg:
Codepens:
 - Before: https://codepen.io/anon/pen/LKZXRy
 - After: https://codepen.io/antoinerg/pen/rEzGoL

etpinard:
Thanks for the fix. Nicely done!

While at it, it would be nice to investigate if the other "per-graph" variables initialised in `Plots.doCalcdata`:

https://github.com/plotly/plotly.js/blob/795220f815afcf79c73885715379b606afc1b4a1/src/plots/plots.js#L2755-L2764

lead to similar bugs when sorting categories per value.

----

Regardless this is a great fix, so :dancer:

Issue #3592, opened 2019-03-01
[Feature Request] use own style.json instead of mapbox access token 


anderl80:
For plotly scatter mal plots:

I'd wish the package also offered the possibility to use a self defined style.json like when you use Mapbox natively in JS. Using this specification (and using OSM tiles) makes the mapbox_access_token obsolete.

https://github.com/plotly/plotly.py/issues/1431

etpinard:
Thanks for posting. We'll make that simpler this spring.

In the meantime you can get this to work by providing a dummy mapbox access token: https://codepen.io/etpinard/pen/KbxgXb

etpinard:
Related https://github.com/plotly/plotly.js/issues/3173, https://github.com/plotly/plotly.js/issues/2988

PR #3987, opened 2019-06-26
mapbox-gl@1.0.0 + improvements

etpinard:
_This PR will serve as a base for all v1.49.0 mapbox-related improvements._

---

In brief, this PR:

- bumps mapbox-gl to `v1.0.0` in https://github.com/plotly/plotly.js/commit/2d4fe9b055af4cce1aa37be99c770ec8da7ddd11 (which went smooth apart from https://github.com/plotly/plotly.js/commit/5b43b8cc901603682bc456ecd0db9d02dc99b950) :tada: 
- commit https://github.com/plotly/plotly.js/pull/3987/commits/41d167f5c3cf6bf35559d66fe44f0ec52f43c65a allows users to use third-party `layout.mapbox.style` (e.g. from Open-Street Map, see mock in https://github.com/plotly/plotly.js/commit/23ffe4fad21f48dab70a49236cfe7c6e0aa9da90). This resolves https://github.com/plotly/plotly.js/issues/3592 

cc @plotly/plotly_js


src/plots/mapbox/index.js
```diff
@@ -69,14 +66,7 @@ exports.plot = function plot(gd) {
...
         if(!mapbox) {
-            mapbox = createMapbox({
-                gd: gd,
-                container: fullLayout._glcontainer.node(),
-                id: id,
-                fullLayout: fullLayout,
-                staticPlot: gd._context.staticPlot
-            });
-
+            mapbox = new Mapbox(gd, id);
```
antoinerg: Nice clean up here :+1:

antoinerg:
I am a bit surprised that the markers in the legend are now smaller in `test/image/baselines/mapbox_0.png`. Can you tell me what changes introduced in this PR are causing this?

etpinard:
> Can you tell me what changes introduced in this PR are causing this?

I forgot to update the mapbox baselines in https://github.com/plotly/plotly.js/pull/3732

antoinerg:
Thanks @etpinard for the very nice PR! I also noticed that `mapbox_fill` now renders properly the `fillcolor`  :+1: 

:dancer:

etpinard:
TODO:

- [ ] add `layout.mapbox.style: 'osm'` as an alias to 

```js
{
  "style": {
        "id": "osm",
        "version": 8,
        "sources": {
          "simple-tiles": {
            "type": "raster",
            "tiles": [
              "https://a.tile.openstreetmap.org/{z}/{x}/{y}.png",
              "https://b.tile.openstreetmap.org/{z}/{x}/{y}.png"
            ],
            "tileSize": 256
          }
        },
        "layers": [
          {
            "id": "simple-tiles",
            "type": "raster",
            "source": "simple-tiles",
            "minzoom": 0,
            "maxzoom": 22
          }
        ]
      }
}
```

etpinard:
So @nicolaskruchten , the only parameters we need to agree on are:

- `tileSize` - which is _the minimum visual size to display tiles for this layer_. The default is 512, we could lower it to 256 (like mapbox does in their docs). But all in all this should not be a big deal.
- `minzoom` / `maxzoom` - which I thought of setting for a max range of `0` to `22`

etpinard:
> * add `layout.mapbox.style: 'osm'` as an alias

Done in https://github.com/plotly/plotly.js/pull/3987/commits/36e3f06680ac0d80117a12a5f7cfd4141e3c750d

Example: https://codepen.io/etpinard/pen/jjYKJp

etpinard:
@plotly/plotly_js I'm looking for a final review.

This PR now includes:

- `mapbox-gl@v1.1.0`
- support for custom non-mapbox base styles
- new `open-street-map` base style
- new trace type `choroplethmabpox`
- new trace type `densitymapbox`
- support for raster and image layout mapbox layers
- `choropleth` `marker.line.color` dflt fix

archmoj:
> So @nicolaskruchten , the only parameters we need to agree on are:
> 
> * `tileSize` - which is _the minimum visual size to display tiles for this layer_. The default is 512, we could lower it to 256 (like mapbox does in their docs). But all in all this should not be a big deal.
> * `minzoom` / `maxzoom` - which I thought of setting for a max range of `0` to `22`

According to https://blog.mapbox.com/512-map-tiles-cb5bfd6e72ba, it looks they suggested using 512 resolution.
Also that may help reduce the number of requests to the server in addition to less i/o process for caching.

etpinard:
> Also that may help reduce the number of requests to the server in addition to less i/o process for caching.

How? I'm curious.

archmoj:
> > Also that may help reduce the number of requests to the server in addition to less i/o process for caching.
> 
> How? I'm curious.

I have rather limited experience on this. 
Anyway regarding i/o caching more tiles means more files to seek for.
I think one advantage of tiles of 256 x 256 was that the integer index could simply fit in an unsigned short integers (0-65535). That may be one reason it is used in bucket rendering engines. On the other hand requesting for a tile of 512 x 512 may reduce the number of requests by four; plus they may have done other optimisations e.g. related to clipping text printed close the tile boundaries.

etpinard:
@archmoj so, I'm assuming you'd vote for having `tileSize:512`? I don't have a strong opinion. I can't notice a performance different between the two (on my laptop using the plotly wifi)

archmoj:
> @archmoj so, I'm assuming you'd vote for having `tileSize:512`? I don't have a strong opinion. I can't notice a performance different between the two (on my laptop using the plotly wifi)

Yes in the case of `mapbox`.

etpinard:
Upon further inspection, `tileSize:256` (top subplot) looks a lot better then `tileSize:512` (bottom subplot):

image

What do you think @archmoj

archmoj:
> Upon further inspection, `tileSize:256` (top subplot) looks a lot better then `tileSize:512` (bottom subplot):
> 
> image
> 
> What do you think @archmoj ?

@etpinard 
Just to double check: 
Did you clean the cache?

etpinard:
> Did you clean the cache?

How do I do that?

archmoj:
Confirmed. Let's use 256.

archmoj:
This is likely the first time ever that multiple `new trace types` were added in one PR!
Brilliant work @etpinard 
:1st_place_medal: 
:dancer:

PR #3988, opened 2019-06-26
Introducing choroplethmapbox traces

etpinard:
_Based off https://github.com/plotly/plotly.js/pull/3987_

-----------

This PR adds a new mapbox-based trace type: `choroplethmapbox` :tada: :world_map: :artificial_satellite: 

Example: https://codepen.io/etpinard/pen/orGQNe

In brief, `choroplethmapbox` traces have a very similar API as their geo-subplot cousin `choropleth`. Both use `locations` and `z` to set the polygon-to-color mapping. But unlike `choropleth`, `choroplethmapbox` traces (currently) must be set with a `geojson` attribute that contains polygon coordinates.

Note that, this PR also fixes a `choropleth` bug in https://github.com/plotly/plotly.js/commit/91e03193ad374a1537caf10d0a16ad97c6f778fc - where we didn't use the correct `marker.line.color` default.

### Things that work

- `geojson` can be set either as URL string or `{ }` GeoJSON object. We accept GeoJSONs of type `Feature` and `FeatureCollection` and geometries of type `Polygon` and `MultiPolygon`.
- colorscale / coloraxis work just like `choropleth`
- hover and selections reuse the `choropleth` logic, where the polygons' centroid is used.
- GeoJSON `properties` fields are accessible in `hovertemplate` ([example](https://codepen.io/etpinard/pen/JQreKJ))

### Things worthy of discussion

- The `below` default. `below` (like in [`mapbox.layers`](https://github.com/plotly/plotly.js/blob/a6393bce2f27183d1c6a304e6a00057dc401f9b7/src/plots/mapbox/layout_attributes.js#L147-L157)) sets the `id` of the layer just below where the `choroplethmapbox` trace will be plotted. I tentatively chose the layer just above the last "water" layer. That is, `choroplethmapbox` traces are plotted below roads and labels by default.
- Maybe instead of some "smart" default for `below`, we could plot `choroplethmapbox` traces above all background layers (but below `scattermapbox` traces) and have a default `marker.opacity` of e.g. 0.7?

### Things that we could add

- topojson support, that way users could (easily) use https://cdn.plot.ly/world_110m.json and friends 
- `locationmode`. The current behavior would correspond to `locationmode: 'geojson-id'`. Some users might expect to identify GeoJSON features using their `properties.name` (this would correspond to e.g. `locationmode: 'geojson-prop-name'`. Moreover, we could also support the `locationmode` values found in geo traces: [`'ISO-3', 'USA-states', 'country names'`](https://github.com/plotly/plotly.js/blob/a6393bce2f27183d1c6a304e6a00057dc401f9b7/src/traces/scattergeo/attributes.js#L42-L51)
- some `zkey` attribute as an alternative to the `z` array. For example, `zkey: 'value'` would read off the `properties.value` values in the GeoJSON set by `geojson`

cc @plotly/plotly_js @nicolaskruchten @chriddyp


src/plots/mapbox/mapbox.js
```diff
@@ -111,125 +114,56 @@ proto.createMap = function(calcData, fullLayout, resolve, reject) {
...
+proto.fetchMapData = function(calcData) {
+    var promises = [];
+
+    function fetch(url) {
+        return new Promise(function(resolve, reject) {
+            d3.json(url, function(err, d) {
+                if(err) {
+                    delete PlotlyGeoAssets[url];
+                    var msg = err.status === 404 ?
+                        ('GeoJSON at URL ' + url + ' does not exist.') :
```
archmoj: Adding double quotations before & after url may help the message to be slightly more readable.

src/traces/choroplethmapbox/index.js
```diff
@@ -0,0 +1,40 @@
...
+        if(cd) {
+            var trace = cd[0].trace;
+            trace._glTrace.updateOnSelect(cd);
+        }
+    },
+
+    moduleType: 'trace',
+    name: 'choroplethmapbox',
+    basePlotModule: require('../../plots/mapbox'),
+    categories: ['mapbox', 'gl', 'noOpacity'],
```
archmoj: Later on the road, it may be of interest to have traces with opacity. Right?

src/traces/choroplethmapbox/convert.js
```diff
@@ -0,0 +1,244 @@
...
+// (just like we currently do for geo choropleth polygons),
+// maybe instead it would make more sense to compute the centroid
+// of each polygon and consider those on hover/select
+function findCentroid(feature) {
+    var geometry = feature.geometry;
+    var poly;
+
+    if(geometry.type === 'MultiPolygon') {
+        var coords = geometry.coordinates;
+        var maxArea = 0;
```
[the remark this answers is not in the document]
etpinard: Why would an _area_ be less than 0?

src/traces/choroplethmapbox/plot.js
```diff
@@ -0,0 +1,152 @@
...
+
+    var mapLayers = this.subplot.map.getStyle().layers;
+    var out = '';
+
+    // find layer just above top-most "water" layer
+    for(var i = 0; i < mapLayers.length; i++) {
+        var layerId = mapLayers[i].id;
+
+        if(typeof layerId === 'string') {
+            var isWaterLayer = layerId.indexOf('water') === 0;
```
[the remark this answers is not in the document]
etpinard: We pick the layer over the _last_ layer starting with "water"

src/traces/choroplethmapbox/index.js
etpinard: `opacity` here for `choroplethmapbox` trace would be equivalent to `marker.opacity`, so we don't need it.

lib/index-mapbox.js
```diff
@@ -11,7 +11,8 @@
 var Plotly = require('./core');
 
 Plotly.register([
-    require('./scattermapbox')
+    require('./scattermapbox'),
+    require('./choroplethmapbox')
```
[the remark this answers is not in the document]
etpinard: See https://github.com/plotly/plotly.js/blob/cb202a8e47631e20555de382d2bbc7393625519b/src/plots/mapbox/mapbox.js#L201-L204

The order in which we `Plotly.register` traces has no impact on their ordering on the graph.

src/traces/choroplethmapbox/index.js
archmoj: What about dimming marker.lines? Trace opacity may be handy for adjusting both.

etpinard: Maybe, but I'm not going to do this now for consistency with `choropleth`:

https://github.com/plotly/plotly.js/blob/cb202a8e47631e20555de382d2bbc7393625519b/src/traces/choropleth/index.js#L26

src/plots/mapbox/mapbox.js
etpinard: done in https://github.com/plotly/plotly.js/pull/3988/commits/e509047a6c99db6f2e452d20c304587ce51ce1cb

src/traces/choropleth/calc.js
```diff
@@ -20,13 +19,29 @@ module.exports = function calc(gd, trace) {
     var len = trace._length;
     var calcTrace = new Array(len);
 
+    var isNonBlankString = function(v) { return v && typeof v === 'string'; };
```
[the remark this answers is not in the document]
etpinard: ok done in https://github.com/plotly/plotly.js/pull/3988/commits/e509047a6c99db6f2e452d20c304587ce51ce1cb

src/traces/choroplethmapbox/defaults.js
```diff
@@ -0,0 +1,47 @@
...
+    traceOut._length = Math.min(locations.length, z.length);
+
+    coerce('below');
+
+    coerce('text');
+    coerce('hovertext');
+    coerce('hovertemplate');
+
+    coerce('marker.line.color');
+    coerce('marker.line.width');
```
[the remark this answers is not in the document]
etpinard: Good call!! Done in -> https://github.com/plotly/plotly.js/pull/3988/commits/654c08e6c927253ab34019167876ba8942468912

src/traces/choropleth/attributes.js
```diff
@@ -40,7 +41,7 @@ module.exports = extendFlat({
     }),
     marker: {
         line: {
-            color: scatterGeoMarkerLineAttrs.color,
+            color: extendFlat({}, scatterGeoMarkerLineAttrs.color, {dflt: defaultLine}),
```
antoinerg: Changing the default here changes a lot of baselines! How did you choose the new color? Also, should we change `scattergeo`'s default as well?

archmoj: There was a bug in the previous setup not picking the right default color. The new baselines are OK.

etpinard: _discussed in-person with @antoinerg_

Resolved.

etpinard:
Is everyone subscribed to the PR ok with the behavior?

etpinard:
No news, good news I guess. Merging this PR into https://github.com/plotly/plotly.js/pull/3987 to have all 1.49.0 mapbox items in one place

Issue #3110, opened 2018-10-16
ENH: Add support for heatmap MapBox layers


jonmmease:
In a [forum post](https://community.plot.ly/t/anything-in-python-plotly-can-create-guardian-heat-map-on-maps/14544) a user asked if there is a way to create heatmap-style plots on a map. e.g.

![55b748a71faa544cd6478eb9aa0c0af59e7adf4d](https://user-images.githubusercontent.com/15064365/47007631-a864fe00-d106-11e8-9410-e529b6316221.jpeg)

It looks like MapBox supports a `heatmap` layer that would support this (See https://www.mapbox.com/mapbox-gl-js/example/heatmap-layer/), so it would be nice to add support for this layer type in the future.

nk9:
Seconded!

SterlingButters:
Thirded

inon-peled:
Definitely useful.

etpinard:
done in https://github.com/plotly/plotly.js/pull/3993 - merged in master via https://github.com/plotly/plotly.js/pull/3987

PR #3993, opened 2019-06-28
Introducing densitymapbox traces

etpinard:
Based https://github.com/plotly/plotly.js/pull/3988 (which is based off https://github.com/plotly/plotly.js/pull/3987)

Resolves https://github.com/plotly/plotly.js/issues/3110

---------

This PR adds a new mapbox-based trace type: `densitymapbox` 

Example: https://codepen.io/etpinard/pen/KjZXOr

`densitymapbox` traces are essentially just a wrapper around [`mapbox-gl`'s "heatmap" layers](https://docs.mapbox.com/mapbox-gl-js/example/heatmap-layer/) - which we chose not the call _`heatmapmapbox`_ as their algorithm is "too different" from our `heatmap` traces.

Please refer to the comments and descriptions in https://github.com/plotly/plotly.js/commit/7e15d817deff6df4f1750728d45ef3533b6e8118 for more info.

---

cc @plotly/plotly_js @nicolaskruchten @chriddyp


src/traces/densitymapbox/plot.js
```diff
@@ -0,0 +1,131 @@
...
+
+        subplot.setOptions(id, 'setLayoutProperty', opts.layout);
+
+        if(opts.layout.visibility === 'visible') {
+            subplot.setOptions(id, 'setPaintProperty', opts.paint);
+        }
+    }
+};
+
+proto._addLayers = function(optsAll) {
```
antoinerg: Can we reuse the one from `src/traces/choroplethmapbox/plot.js`? I think it is identical

src/traces/densitymapbox/plot.js
```diff
@@ -0,0 +1,131 @@
...
+    ];
+
+    // previous 'below' value,
+    // need this to update it properly
+    this.below = null;
+}
+
+var proto = DensityMapbox.prototype;
+
+proto.update = function(calcTrace) {
```
antoinerg: Also very similar to the function in `src/traces/choroplethmapbox/plot.js`. I'll leave it up to you to decide whether to reuse it or not.

src/traces/densitymapbox/plot.js
etpinard: We could, but I don't think it's worth it.

src/traces/densitymapbox/plot.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    this.below = below;
+};
+
+proto._removeLayers = function() {
+    var map = this.subplot.map;
+    var layerList = this.layerList;
+
+    for(var i = 0; i < layerList.length; i++) {
```
archmoj: The loop order may be reversed here.
Removing from last item of the list towards the first item may have the benefit of having the valid index. 
On another note - I was wondering if the list should be set to an empty array at the end of this process?

etpinard: Done in https://github.com/plotly/plotly.js/pull/3993/commits/1add90efdfb7d930ff3c0c880199cdeef776b5e2

antoinerg:
It looks great :tada: ! Nice job @etpinard :muscle: 

 :dancer:

meghnakayam:
This looks great. Is there a way to incorporate the same in python for Plotly Dash?

nicolaskruchten:
@meghnakayam yep! This is coming to Plotly.py/Dash in a couple of weeks :)

Issue #3992, opened 2019-06-27
World Choropleth map issue with showcoastlines: false


michaelbabyn:
Originally reported in /plotly/documentation#1378, [this choropleth example](https://plot.ly/python/choropleth-maps/#world-choropleth-map) doesn't load anymore due to a regression introduced in 1.48. See this [codepen with `layout.geo.showcoastlines: true` using plotly.js 1.47.0](https://codepen.io/michaelbabyn/pen/gNXzep) and the [same plot using version 1.48.0](https://codepen.io/michaelbabyn/pen/agVGQQ) of plotly.js.

I see the following error in the console:

```
Uncaught (in promise) TypeError: Cannot read property 'objects' of null
    at n.getTopojsonFeatures (plotlyjs-bundle.7315e6932bed.js:1)
    at u (plotlyjs-bundle.7315e6932bed.js:1)
    at Object.e.exports [as plot] (plotlyjs-bundle.7315e6932bed.js:1)
    at Object.g.generalUpdatePerTraceModule (plotlyjs-bundle.7315e6932bed.js:1)
    at y.x.update (plotlyjs-bundle.7315e6932bed.js:1)
    at y.x.plot (plotlyjs-bundle.7315e6932bed.js:1)
    at Object.plot (plotlyjs-bundle.7315e6932bed.js:1)
    at r.drawData (plotlyjs-bundle.7315e6932bed.js:1)
    at Object.l.syncOrAsync (plotlyjs-bundle.7315e6932bed.js:1)
    at i (plotlyjs-bundle.7315e6932bed.js:1)
```

PR #3994, opened 2019-06-28
Fetch topojson when 1+ trace(s) has `locationmode`

etpinard:
... so that `_module.plot` can find the relevant polygon and/or  centroids, even though no base layer is drawn - fixes https://github.com/plotly/plotly.js/issues/3992 introduced in https://github.com/plotly/plotly.js/pull/3856/commits/b04102b745df2a4c79e86c27b8bd1e7d108fab67

before: https://codepen.io/michaelbabyn/pen/agVGQQ
after: https://codepen.io/etpinard/pen/YoYYyO

@plotly/plotly_js

### DIFF
--- a/src/plots/geo/geo.js
+++ b/src/plots/geo/geo.js
@@ -80,6 +80,12 @@ proto.plot = function(geoCalcData, fullLayout, promises) {
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


archmoj:
@etpinard thanks for the fix.
:dancer:

PR #4007, opened 2019-07-02
indicator: position delta using drawing.bBox

antoinerg:
Because guessing the size of the text to vertically align it wasn't robust enough.

I introduced a new mock `indicator_datacard3` to showcase the 12 possible combinations of `delta.position` and `align`. Changing the font size of either `number` or `delta` now always yield a correct alignment.

Codepen: https://codepen.io/antoinerg/pen/qzYpeV


src/traces/indicator/plot.js
```diff
@@ -650,11 +622,46 @@ function drawNumbers(gd, plotGroup, cd, opts) {
                 return deltaFormatText(deltaValue(cd[0]));
             });
         }
+
+        delta.attr('data-unformatted', deltaFormatText(deltaValue(cd[0])));
```
etpinard: I'm curious. Why did you use `data-unformatted` here?

src/snapshot/tosvg.js
```diff
@@ -88,7 +88,7 @@ module.exports = function toSVG(gd, format, scale) {
     // fill whatever container it's displayed in regardless of plot size.
     svg.node().style.background = '';
 
-    svg.selectAll('text,tspan')
+    svg.selectAll('text')
```
etpinard: Nice touch!

src/traces/indicator/plot.js
antoinerg: Because `Drawing.bBox` fails without it when it tries to hash the node:
https://github.com/plotly/plotly.js/blob/44633a13a3e022a4565d7b230b9db3e1e1ac83aa/src/components/drawing/index.js#L990-L991

etpinard: Cool. Thanks for the info! Nice find!

test/image/mocks/indicator_datacard3.json
```diff
@@ -0,0 +1,140 @@
+{
+  "data": [{
```
etpinard: Nice mock!

For reference, on master this mock looks like


image

etpinard:
Looks good @antoinerg - thanks for the quick turnaround

:dancer:

test/image/mocks/indicator_datacard3.json
antoinerg: It looks fine on `master` but if you change the font-size of delta or number, the alignment might break.

Issue #3771, opened 2019-04-15
"Failed - network error" when downloading large png using toImage button


slishak:
I haven't yet found the actual limiting condition, but after some threshold of plot width/height/number of data points, I am unable to download a png of a Plotly plot. See following codepen:

https://codepen.io/anon/pen/WWZBOB

With `n = 20000`, I can't download the png, Chrome just shows the following message:
![image](https://user-images.githubusercontent.com/12515343/56132175-9ebf3480-5f89-11e9-808e-b51b0cf73b32.png)

With `n = 10000`, it works. It also works with a 2000x2000 figure layout and 20000 points.

etpinard:
Thanks for reporting!

Interestingly, downloading your graph in FF66 works fine. Chrome must put a URL-length limit in `<a>` with `download` attribute like we use here:

https://github.com/plotly/plotly.js/blob/4529fe92c6a13eb0ebeb4ede38592e67183d8df6/src/snapshot/filesaver.js#L45-L52

I found [this stackoverflow post](https://stackoverflow.com/a/41755526) claiming that Chrome uses a 2MB limit. 

We should probably try using [`URL.createObjectURL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL).

etpinard:
PoC in https://github.com/plotly/plotly.js/compare/createObjectURL 

We should be able to get a fix in the next minor 1.49.0 (I wouldn't want to try this our in a patch release).

etpinard:
https://codepen.io/etpinard/pen/zVdjQg which uses https://42931-45646037-gh.circle-artifacts.com/0/dist/plotly.min.js 

appears to work correctly on Chrome 70 and FF67 on my laptop.

---

Before making a PR, we still need to test the legacy fallbacks on IE10, IE11 and Safari.

antoinerg:
I manually tested saving mock `14.json` to image by clicking the snapshot button in the modebar.

- [x] IE11: works
- [ ] Safari 12 on macOS Mojave: it goes to the following page instead of saving to file:
![Screenshot_2019-06-27_20-13-51](https://user-images.githubusercontent.com/301546/60308728-3f31bd80-9918-11e9-8695-457478964a5d.png)
- [ ] IE10: cannot even render the figure even when uncommenting https://github.com/plotly/plotly.js/blob/3d383b796d5738b9f93356db2cc5b4b51570e4a5/devtools/test_dashboard/index.html#L22-L24

[...]

PR #4008, opened 2019-07-02
Use URL.createObjectURL during Plotly.downloadImage

etpinard:
... to not hit the 2MB URL-length limit in Chrome - which resolves https://github.com/plotly/plotly.js/issues/3771 - and improve `downloadImage` performance.

before: https://codepen.io/anon/pen/WWZBOB
after: https://codepen.io/etpinard/pen/MMGGVW (click to "Download plot as a png" modebar button)

All-format playground: https://codepen.io/etpinard/pen/QXraeW 

@plotly/plotly_js

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -733,6 +733,11 @@ lib.isIE = function() {
     return typeof window.navigator.msSaveBlob !== 'undefined';
 };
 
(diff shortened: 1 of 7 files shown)


antoinerg:
Looks good to me! Thanks for the improvement :dancer:

PR #4009, opened 2019-07-02
gl-text patch to use to-px at 1.0.1

archmoj:
Fixes noCI test: https://github.com/plotly/plotly.js/pull/3904#issuecomment-507852140 as a result of upgrading `to-px` from 1.0.1 to 1.1.0

@etpinard

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5181,9 +5181,9 @@
       }
     },
     "gl-text": {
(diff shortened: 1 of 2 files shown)


etpinard:
:dancer: thanks for the (very) quick turnaround!

PR #3966, opened 2019-06-15
Refactor parcoords - bug fixes - reuse axes ticktext - support pseudo html labels - features for label angle & position

archmoj:
Refactoring `parcoords` code helped fixing:
https://github.com/plotly/plotly.js/issues/3794
https://github.com/plotly/plotly.js/issues/3946

Reusing Cartesian axes `tickText` function for linear axes (`parcoords` has linear and ordinal axis types) 
https://github.com/plotly/plotly.js/issues/3930

Support for pseudo-html and MathJax formatting in `parcoords` dimension labels
https://github.com/plotly/plotly.js/issues/3206

And label line breaks useful for:
https://github.com/plotly/phoenix-integration/issues/242
Also two new attributes are added at trace data level to help control the angle and position of labels 

The 3 vertex shader programs as well as 2 helper functions are all combined in one vertex shader & optimised.
Also to mention low precision flag in fragment shader is replaced by high precision flag for better output on different devices. 

To investigate performance benchmarks please visit the examples below and find the console.logs after refresh!
1050ms [codepen 1 before](https://codepen.io/MojtabaSamimi/pen/WqrOrb?editors=0010)
0700ms [codepen 1 after](https://codepen.io/MojtabaSamimi/pen/PrZjwL?editors=0010)

[codepen 2 before](https://codepen.io/MojtabaSamimi/pen/WqGQMB?editors=0010)
[codepen 2 after](https://codepen.io/MojtabaSamimi/pen/xoEwQr?editors=0010)

@plotly/plotly_js

PR #4011, opened 2019-07-03
parcoords - bug fix handling undefined ranges and dticks for tickText

archmoj:
In addition to some refactors (e.g. to avoid using one variable name in two scopes), 
this PR fixes the `tickText` defaults for undefined ranges and ticks to be used by `tickText`.

Please note the changes in two baselines would make them look similar to what they look before #3966. 
@plotly/plotly_js


src/traces/parcoords/defaults.js
```diff
@@ -55,6 +55,19 @@ function dimensionDefaults(dimensionIn, dimensionOut, parentOut, opts) {
         coerce('ticktext');
         coerce('tickformat');
         var range = coerce('range');
+        if(!range) {
```
etpinard: Hmm I would move this down to the plot step. That way we don't have to loop over the data during the defaults.

etpinard:
Nice fix @archmoj - I made one small comment.

src/traces/parcoords/defaults.js
archmoj: Good call.
Done in https://github.com/plotly/plotly.js/pull/4011/commits/bf02101c5d8c51e2db07065f1a7ab20ce0e8ff55.

etpinard:
Excellent. :dancer: 

Sorry about those intermittent test failures. I'm working on it at the moment.

PR #4013, opened 2019-07-03
parcoords - reusing Lib.aggNums in finding extreme ranges

archmoj:
Follow up of #4011.
Reusing Lib.aggnum in finding dimension ranges e.g. for better support of numbers passed as strings.

@plotly/plotly_js


etpinard:
nice catch - :dancer:

Issue #3737, opened 2019-04-05
Bar attributes shouldn't have anim:true in the schema


cpsievert:
Perhaps this is expected, but I found it surprising since `bar`'s y atttribute has a `anim: true` flag in this config (this means it should work with the react transitioning API, right?)

https://codepen.io/cpsievert/pen/axZZeP

etpinard:
Duplicate of https://github.com/plotly/plotly.js/issues/1019

etpinard:
Ha oops. Sorry @cpsievert I misread the issue. I thought we were asking whether bar traces can be smoothly transitioned. The answer is no.

But you're right, `anim: true` for `x`, `y` etc should be there in the `bar` trace schema. Have you noticed `anim: true` showing up anywhere else?

cpsievert:
No, but I haven't really been playing around with `Plotly.react()` much at all yet

PR #4024, opened 2019-07-05
Schema animatable + anim fixes

etpinard:
resolves https://github.com/plotly/plotly.js/issues/3737

In brief, `anim: true` specifications in attribute declarations are now dropped in modules that don't set `animatable: true` during `Plotly.PlotSchema.get()` 

would be nice to slip this one in 1.49.0 - @plotly/plotly_js


archmoj:
Excellent.
:dancer:

PR #4029, opened 2019-07-08
`indicator`: several improvements/fixes

antoinerg:
This PR attempts to fix a few visual issues that can occur on animation:
- [x] fill color of `delta` is not consistent with the currently displayed value (ce3f2d109df17d5443ce77593a56cc76004c224e)
- [x] precompute the final size and position of numbers to rescale properly when the animation start (hence preventing overlap). We effectively save one `Drawing.bBox` call (1e7f21d)
- [x] numbers (either `number` or `delta`) can jump around if the text size changes a lot from the start value to end value. We solve this by aligning the numbers using the minimum X offset seen up to this point (2c378e0).
- [x] fix transition on first animation (6e9def2fac19e3f6ad45dc138745b688ce32acbf)

Additional bug fix:
- [x] it reverts to default `valueformat` for both `number` and `delta` if an empty string is specified (4d2ad95)
- [x] fix bullet with range spanning negative values (3b9028a424fa891852db5300dd91bd1d9c8c83a8)
- [x] do not display SI suffix when passing through zero if no SI suffixes are present in start and end value (8cea8e7)
- [x] add support for blank valueformat 
- [x] do not draw value on gauge with `null` value
- [x] render exponent properly

https://codepen.io/antoinerg/pen/bPQEPr?editors=0010


src/traces/indicator/defaults.js
```diff
@@ -63,7 +64,8 @@ function supplyDefaults(traceIn, traceOut, defaultColor, layout) {
         }
         coerce('delta.reference', traceOut.value);
         coerce('delta.relative');
-        coerce('delta.valueformat', traceOut.delta.relative ? '2%' : '.3s');
+        coerce('delta.valueformat');
+        if(!traceOut.delta.valueformat) traceOut.delta.valueformat = traceOut.delta.relative ? '+2%' : '+.3s';
```
etpinard: As you know, I'm not a fan of these `+` prefixes. They do make animations look better, but again not all `indicator` traces will be used with animations. 

I'd vote for bringing back `.3s` and adding some info in the `delta.tickformat` description about `+.3s` for users wanting more fixed-bbox delta tickers.

But, I'll let @plotly/plotly_js vote on this.

src/traces/indicator/plot.js
```diff
@@ -769,17 +823,40 @@ function strTranslate(x, y) {
     return 'translate(' + x + ',' + y + ')';
 }
 
-function fitTextInsideBox(el, width, height) {
+function fitTextInsideBox(textBB, width, height) {
     // compute scaling ratio to have text fit within specified width and height
-    var textBB = Drawing.bBox(el.node());
+    // var textBB = Drawing.bBox(el.node());
```
etpinard: :zap:

src/traces/indicator/plot.js
```diff
@@ -769,17 +823,40 @@ function strTranslate(x, y) {
...
     var ratio = Math.min(width / textBB.width, height / textBB.height);
     return [ratio, textBB, width + 'x' + height];
 }
 
-function fitTextInsideCircle(el, radius) {
+function fitTextInsideCircle(textBB, radius) {
     // compute scaling ratio to have text fit within specified radius
-    var textBB = Drawing.bBox(el.node());
+    // var textBB = Drawing.bBox(el.node());
```
etpinard: :zap:

antoinerg: Done in 68c2aefa8ab6af09c598b3739149e2d5e89155d9

src/traces/indicator/plot.js
antoinerg: Done in 68c2aefa8ab6af09c598b3739149e2d5e89155d9

antoinerg:
> Do we have a mock testing the "range spanning negative values" case?

Added on in 68c2aefa8ab6af09c598b3739149e2d5e89155d9

src/traces/indicator/defaults.js
antoinerg: Let me remove the `+` and check if animations are OK. I added more padding between `number` and `delta` so we might be fine!

archmoj: imo the dafualt should be set to blank; then under the hood '.3s' could be used (for the moment) until we figure out how to make blank work perfectly.

antoinerg: I removed the `+` sign by default in commit b942832

antoinerg:
Codepen has been update to use commit b942832: https://codepen.io/antoinerg/pen/bPQEPr?editors=0010

etpinard:
@antoinerg are you planning on addressing

> I would prefer if *tickformat: '' would result in the default Axes.tickText formatting

as mentioned in https://github.com/plotly/plotly.js/pull/4029#commitcomment-34221603 ?

Is `Axes.tickText`  still giving you trouble? Maybe @archmoj can help?

That said, this is a small detail, I'd be ok punting this for later. The fixes here in this PR are fantastic!

antoinerg:
> Is `Axes.tickText` still giving you trouble? Maybe @archmoj can help?

It was giving me `NaN` and I didn't investigate much more. I still think using exponent or SI notation is a better default but consistency with the axes is also good... If @archmoj can get it to work I could get behind it.

archmoj:
> Is `Axes.tickText` still giving you trouble? Maybe @archmoj can help?

Yes the blank is not working.
I am on it.

antoinerg:
> Yes the blank is not working.
> I am on it.

Thanks @archmoj ! Don't push directly on this branch as I will need to test it thoroughly before accepting the change.

archmoj:
By defaulting `tickformat` to `.3s` we would get `1.00` and `10.0` as presented on the image on the left.
What is we could mock linear axis in a way that the `tickText` work with blank `tickformat`? 
In that case we can get `1` and `10` as presented on the image on the right.
Screenshot from 2019-07-16 14-43-20
So I am wondering if the blank may be a better default?

First [demo using modified version of '.3s' format](https://codepen.io/MojtabaSamimi/pen/QeWZYJ?editors=1000) - I noticed the `m` suffix may still show up sometimes. 
Second [demo using `plolt.js` Cartesian blank format](https://codepen.io/MojtabaSamimi/pen/vqozNE?editors=0010) - using different setup https://github.com/plotly/plotly.js/commit/1a0565883abf39bac783a64e4bb0af1d273ea2d5

cc: @nicolaskruchten @alexcjohnson @jonmmease 
Please also consider removing `Math.floor` from the JS code to compare transition between non-integers in two cases.

nicolaskruchten:
Generally speaking, I prefer "blank as the default" rather than 3-decimals.

nicolaskruchten:
But I don't think the default for negative values should be red + triangle-down + parens + minus-sign. that feels like overkill squared :)

antoinerg:
> But I don't think the default for negative values should be red + triangle-down + parens + minus-sign. that feels like overkill squared :)

@nicolaskruchten Which one should we ship as a default? A minus sign or parentheses?

nicolaskruchten:
Minus sign, no parens plz

archmoj:
What about default formatting for `relative` deltas? 
It could default to either '2%' or blank. But which one is best?

etpinard:
> What about default formatting for `relative` deltas?
> It could default to either '2%' or blank. But which one is best?

`'2%'` is probably fine here.

archmoj:
> > What about default formatting for `relative` deltas?
> > It could default to either '2%' or blank. But which one is best?
> 
> `'2%'` is probably fine here.

Great! I also think so.

etpinard:
@antoinerg feel free to merge this PR :dancer: for me!

antoinerg:
Thank you @etpinard and @archmoj for helping me bring this to the finish line :checkered_flag:

PR #4030, opened 2019-07-08
edit error type description

cldougl:
I think in the description this "array" should be "data" to match the options listed in values above


etpinard:
Good :eye: !

:dancer:

PR #4035, opened 2019-07-11
Improve `mapbox.style` description + use mapbox-gl v1.1.1

etpinard:
@plotly/plotly_js @nicolaskruchten


antoinerg:
This description is much clearer :+1:

antoinerg:
:dancer: from me

src/plots/mapbox/layout_attributes.js
```diff
@@ -48,10 +48,16 @@ var attrs = module.exports = overrideAll({
         dflt: constants.styleValueDflt,
         role: 'style',
         description: [
-            'Sets the Mapbox map style.',
-            'Either input one of the default Mapbox style names or the URL to a custom style',
-            'or a valid Mapbox style JSON.',
-            'From OpenStreetMap raster tiles, use *open-street-map*.'
+            'Sets the Mapbox base map style.',
+            'Base map styles are rendered below all traces and layout layers.',
+            'Either input one of the default Mapbox style names:', constants.styleValuesMapbox, '.',
```
chriddyp:
```suggestion
            'Either input one of the default Mapbox style names: ', constants.styleValuesMapbox, '.',
```

src/plots/mapbox/layout_attributes.js
```diff
@@ -48,10 +48,16 @@ var attrs = module.exports = overrideAll({
...
+            'Note that to use these, a Mapbox access token must be set either in the `accesstoken` attribute',
+            'or in the `mapboxAccessToken` config option.',
+            'From OpenStreetMap raster tiles, use:', constants.styleValueOSM, '.',
```
chriddyp:
```suggestion
            'From OpenStreetMap raster tiles, use: ', constants.styleValueOSM, '.',
```

src/plots/mapbox/layout_attributes.js
```diff
@@ -48,10 +48,16 @@ var attrs = module.exports = overrideAll({
...
+            'No access token is required to render the', constants.styleValueOSM, 'style.',
+            'One can also set `style` as a URL to a Mapbox custom style, e.g. created in Mapbox Studio.',
+            'Finally, one can set `style` as a Mapbox style JSON, see',
```
chriddyp:
```suggestion
            'Finally, one can set `style` as a Mapbox style JSON, see ',
```

chriddyp: nvm - i'm assuming a space is added after every line

src/plots/mapbox/layout_attributes.js
antoinerg: In the same vein, `constants.styleValuesMapbox.join(', ')` to ensure we have a whitespace after each comma.

antoinerg: `['Possible values are:', ['a','b','c']].join(' ')` -> "Possible values are:  a,b,c"

`['Possible values are:',['a','b','c'].join(', ')].join(' ')` -> "Possible values are: a, b, c"

src/plots/mapbox/layout_attributes.js
```diff
@@ -48,10 +48,16 @@ var attrs = module.exports = overrideAll({
         dflt: constants.styleValueDflt,
         role: 'style',
         description: [
-            'Sets the Mapbox map style.',
-            'Either input one of the default Mapbox style names or the URL to a custom style',
-            'or a valid Mapbox style JSON.',
-            'From OpenStreetMap raster tiles, use *open-street-map*.'
+            'Sets the Mapbox base map style.',
+            'Base map styles are rendered below all traces and layout layers.',
```
nicolaskruchten: I would include a bit more detail here:
* Base map styles can include multiple layers
* Base map layers are always rendered below traces and layers defined in the adjacent `layout.mapbox.layers` attribute

src/plots/mapbox/layout_attributes.js
```diff
@@ -48,10 +48,16 @@ var attrs = module.exports = overrideAll({
         dflt: constants.styleValueDflt,
         role: 'style',
         description: [
-            'Sets the Mapbox map style.',
-            'Either input one of the default Mapbox style names or the URL to a custom style',
-            'or a valid Mapbox style JSON.',
-            'From OpenStreetMap raster tiles, use *open-street-map*.'
+            'Sets the Mapbox base map style.',
```
nicolaskruchten: I would remove the word "Mapbox" here as it's ambiguous if we're talking about Mapbox-provided styles or not.

src/plots/mapbox/layout_attributes.js
nicolaskruchten: I would reverse the order in which we explain this: not [name, url, json] but rather [json, osm, url, name].

My reasoning is that really the intent of this attribute is to specify or identify the JSON object that will be used to render the map. So we should say that first. Then we should say that we provide a shortcut to a particular OSM-based one which you can pass in as a string. Then we should say that you can create and host these styles in the Mapbox studio app, in which case you can provide the URL here and you'll need a token. Then finally we should say that Mapbox provides some default hosted styles and for convenience we have shortcuts for these enumerated things, which are expanded to URLs by prepending `"mapbox://styles/mapbox/"` or whatever. 

I would draw a distinction between the Mapbox-service-provided shortcut names and our new plotly.js-provided shortcut name just to be clear. For example until reading the actual source of https://github.com/plotly/plotly.js/blob/c1ef6911da054f3b16a7abe8fb2d56019988ba14/src/plots/mapbox/constants.js it wasn't clear to me that the "basic", "light", "dark" etc were not specified in our source tree.

src/plots/mapbox/layout_attributes.js
nicolaskruchten: For my own edification: is it possible to use the new `below` attribute to render `choroplethmapbox` traces below layers specified in this `layout.mapbox.style.layers` object or do we somehow constrain the behaviour of `below` to only look at those in `layout.mapbox.layers` ?

nicolaskruchten: Tangent on a tangent: with the new `*mapbox` trace types, how does inter-trace rendering order work? Should `below` be added to all three `mapbox` traces?

etpinard:
More from @nicolaskruchten on slack:

what does this line mean? https://github.com/plotly/plotly.js/blob/master/src/plots/mapbox/layout_attributes.js#L141

Is this line up to date wrt to `raster` and `image` source types? https://github.com/plotly/plotly.js/blame/master/src/plots/mapbox/layout_attributes.js#L120

would be good to clarify this line if it only applies when `sourcetype` is `geojson`: https://github.com/plotly/plotly.js/blob/master/src/plots/mapbox/layout_attributes.js#L143 

----

these will be addressed in this PR.

nicolaskruchten:
Would be good to point out that this is only needed for certain values of `layout.mapbox.style`: https://github.com/plotly/plotly.js/blob/master/src/plots/mapbox/layout_attributes.js#L40

nicolaskruchten:
The relationship between `layers.type` and `layers.sourcetype` is pretty nebulous to me at the moment... Would be good to clarify.

nicolaskruchten:
The attr for `choroplethmapbox.below` here https://github.com/plotly/plotly.js/blob/master/src/traces/choroplethmapbox/attributes.js#L63 should be expanded a bit to say something like "the default value is 'water', which is the ID of a layer that is often present in Mapbox styles. If no layer with this ID is implicitly or explicitly defined via `layout.mapbox.style` or explicitly present in `layout.mapbox.layers` then this trace will be positioned as usual."

... or something. that's basically a clearer description of the current behaviour and assumptions IMO.

src/plots/mapbox/layout_attributes.js
etpinard: > is it possible to use the new `below` attribute to render `choroplethmapbox` traces below layers specified in this `layout.mapbox.style.layers`

Yes.

> how does inter-trace rendering order work

like this:

https://github.com/plotly/plotly.js/blob/14133af4fc205a7d00b158bb91e4afec845fd1d3/src/plots/mapbox/mapbox.js#L200-L204

`densitymapbox` has a `below` attribute:

https://github.com/plotly/plotly.js/blob/14133af4fc205a7d00b158bb91e4afec845fd1d3/src/traces/densitymapbox/attributes.js#L67-L79

only `scattermapbox` does not. I didn't think that would be necessary. I can add `below` to `scattermapbox` if you want.

src/plots/mapbox/layout_attributes.js
etpinard: > My reasoning is that really the intent of this attribute is to specify or identify the JSON object that will be used to render the map. So we should say that first

I disagree here. As long as our default `mapbox.style` is a Mapbox-service style, I think it makes more sense to advertise the other available Mapbox-service style values first. This is of course subject to change when/if we make `'open-street-map'` the default in a major version bump. I really can't imagine that many users playing around with the `mapbox.style` JSON structure. That said, you have veto rights @nicolaskruchten - just writing down my opinion here.


-------------

> I would draw a distinction between the Mapbox-service-provided shortcut names and our new plotly.js-provided shortcut name just to be clear

I don't fully understand what we mean here. Would turning:

```
            'For OpenStreetMap raster tiles, use: ', constants.styleValueOSM, '.',
            'No access token is required to render the', constants.styleValueOSM, 'style.',
```

into 

```
            'For OpenStreetMap raster tiles, use the Plotly shortcut value: ', constants.styleValueOSM, '.',
            'No access token is required to render the', constants.styleValueOSM, 'style.',
```

be ok?

etpinard:
> The attr for `choroplethmapbox.below` here https://github.com/plotly/plotly.js/blob/master/src/traces/choroplethmapbox/attributes.js#L63 should be expanded a bit to say something like "the default value is 'water', which is the ID of a layer that is often present in Mapbox styles.

That's not exactly true, here's the logic:

https://github.com/plotly/plotly.js/blob/14133af4fc205a7d00b158bb91e4afec845fd1d3/src/traces/choroplethmapbox/plot.js#L106-L132

The default `below` value is the layer just above the last `water*` layer. It's unfortunate that mapbox-gl doesn't have an `above` attribute, it would have been easier to describe.

src/plots/mapbox/layout_attributes.js
etpinard: done in https://github.com/plotly/plotly.js/pull/4035/commits/ca762e96d9f8ea77966cfe4918c292e48661f3b0 

Thanks!

src/plots/mapbox/layout_attributes.js
etpinard: https://github.com/plotly/plotly.js/pull/4035/commits/ca762e96d9f8ea77966cfe4918c292e48661f3b0

src/plots/mapbox/layout_attributes.js
etpinard: Good call. Done in https://github.com/plotly/plotly.js/pull/4035/commits/ca762e96d9f8ea77966cfe4918c292e48661f3b0

src/plots/mapbox/layout_attributes.js
nicolaskruchten: > I disagree here. As long as our default mapbox.style is a Mapbox-service style,

Our intent with 1.49 is to permit non-Mapbox, non-OSM layers to be specified as the lowest "base" layers... The only way to do this is to specify this JSON inline. This needs to be more clearly documented front and center IMO.

I also still think that `open-street-map` should be the default if no mapbox access token is present.

etpinard: > I also still think that open-street-map should be the default if no mapbox access token is present.

I'd call that _too magical_.

nicolaskruchten: Why? Right now if you don't give a token, you get nothing. Something is better than nothing.

alexcjohnson: I'm not sure which side I fall on here in the end, but let's say a user wants mapbox, they make a plot but forget to provide their token, and it shows up as OSM... how do they debug this? How do they figure out that it's the lack of token that gave them OSM? Feels like "you get nothing" is a clearer signal that you just omitted something, whereas "you get OSM" sends you down the path of trying to find where exactly you asked for OSM - answer: nowhere, but if you're lucky you'll find that the docs for which provider to use tell you about the token via their description of the default. Can we improve on this experience?

nicolaskruchten: OK, that's an actual objection I can respond to :)

In this case I think that I would say that if neither the token nor the style are provided, the console error message that currently says "token missing" should say "please specify a style". Instead of the current behaviour which automatically encourages mapbox without leaving the door open to non-mapbox styles. The current message is appropriate when the style is a mapbox style and no token is provided because then the user is explicitly requesting a mapbox style and should provide a token.

src/plots/mapbox/layout_attributes.js
nicolaskruchten: I think for completeness and symmetry would should add `below` to `scattermapbox`, yes.

src/plots/mapbox/layout_attributes.js
nicolaskruchten: (follow-up in streambed)

etpinard:
Based of this branch's latest commit

https://github.com/plotly/plotly.js/blob/ca762e96d9f8ea77966cfe4918c292e48661f3b0/src/plots/mapbox/layout_attributes.js

here are some **TODOs** that will done on this branch:

- [ ] Improve "mapbox token error message" https://github.com/plotly/plotly.js/issues/4059

- [ ] rewrite `mapbox.style` description as:

```
Defines the map layers that are rendered by default below the trace layers defined in `data`, 
which are themselves by default rendered below the layers defined in `layout.mapbox.layers`.

These layers can be defined either explicitly as a Mapbox Style object which can contain multiple 
layer definitions that load data from any public or private Web Map Service (WMS), or implicitly 
by using one of the builtin plotly.js style objects which use WMSes which do not require any 
access tokens, or by using a  default Mapbox style or custom Mapbox style URL, both of 
which require a Mapbox access token.

Mapbox Style objects are of the form described in the Mapbox GL JS documentation available at
https://docs.mapbox.com/mapbox-gl-js/style-spec

The built-in plotly.js styles objects are... (provide name and the tile source)

The built-in Mapbox styles are...

Mapbox style URLs are of the form...
```

- [ ] Clarify relationship between `layers.type` and `layers.sourcetype`

- [ ] anything else?

src/plots/mapbox/layout_attributes.js
etpinard: Now in https://github.com/plotly/plotly.js/issues/4049

nicolaskruchten:
This is already a big set of improvements, thank you very much :)

One small adjustment to my initial proposal: the phrase `public or private Web Map Service (WMS)` should be expanded to say `public or private Tile Map Service (TMS or XYZ) or Web Map Service (WMS)`

etpinard:
@plotly/plotly_js @nicolaskruchten this PR is now `reviewable` again.

Note that I merged master (with @antoinerg 's attributions and non-mapbox styles work) and I bumped `mapbox-gl` to [`v1.1.1`](https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#111) - so this is essentially on-track to be the "last" mapbox PR for 1.49.0 :smirk:

src/plots/mapbox/layout_attributes.js
```diff
@@ -39,19 +39,38 @@ var attrs = module.exports = overrideAll({
         description: [
             'Sets the mapbox access token to be used for this mapbox map.',
             'Alternatively, the mapbox access token can be set in the',
-            'configuration options under `mapboxAccessToken`.'
+            'configuration options under `mapboxAccessToken`.',
+            'Note that accessToken are only required when `style`',
```
nicolaskruchten: I think there's a line missing here... "when `style` ... something?"

src/plots/mapbox/layout_attributes.js
```diff
@@ -39,19 +39,38 @@ var attrs = module.exports = overrideAll({
...
-        values: constants.styleValuesMapbox.concat(constants.styleValueOSM),
+        values: constants.styleValuesMapbox.concat(Object.keys(constants.styleValuesNonMapbox)),
         dflt: constants.styleValueDflt,
         role: 'style',
         description: [
-            'Sets the Mapbox map style.',
-            'Either input one of the default Mapbox style names or the URL to a custom style',
-            'or a valid Mapbox style JSON.',
-            'From OpenStreetMap raster tiles, use *open-street-map*.'
+            'Defines the map layers that are rendered by default below the trace layers defined in `data`,',
```
nicolaskruchten: 👍 to this new description!

src/plots/mapbox/layout_attributes.js
```diff
@@ -117,7 +137,7 @@ var attrs = module.exports = overrideAll({
                 'Sets the source data for this layer (mapbox.layer.source).',
                 'Source can be either a URL,',
                 'a geojson object (with `sourcetype` set to *geojson*)',
-                'or an array of tile URLS (with `sourcetype` set to *vector*).'
+                'or an array of URLs (with `sourcetype` set to *vector* or *raster*).'
```
nicolaskruchten: what does `sourcetype` have to be if `source` is a URL?

src/plots/mapbox/layout_attributes.js
```diff
@@ -145,10 +165,13 @@ var attrs = module.exports = overrideAll({
             dflt: 'circle',
             role: 'info',
             description: [
-                'Sets the layer type (mapbox.layer.type).',
-                'Support for *raster*, *background* types is coming soon.',
-                'Note that *line* and *fill* are not compatible with Point',
-                'GeoJSON geometries.'
+                'Sets the layer type,',
+                'that is the how the layer data set in `source` will be rendered',
+                'With `sourcetype` set to *geojson*, *circle*, *line*, *fill* and *symbol* are available',
```
nicolaskruchten:
```suggestion
                'With `sourcetype` set to *geojson*, the following values are allowed: *circle*, *line*, *fill* and *symbol*',
```

src/plots/mapbox/layout_attributes.js
```diff
@@ -145,10 +165,13 @@ var attrs = module.exports = overrideAll({
...
+                'but note that *line* and *fill* are not compatible with Point',
+                'GeoJSON geometries.',
+                'With `sourcetype` set to *vector*, *circle*, *line*, *fill* and *symbol* are available.',
```
nicolaskruchten:
```suggestion
                'With `sourcetype` set to *vector*, the following values are allowed: *circle*, *line*, *fill* and *symbol*.',
```

src/plots/mapbox/layout_attributes.js
```diff
@@ -145,10 +165,13 @@ var attrs = module.exports = overrideAll({
...
+                'With `sourcetype` set to *raster* or `*image*`, only the *raster* value is available.'
```
nicolaskruchten:
```suggestion
                'With `sourcetype` set to *raster* or `*image*`, only the *raster* value is allowed.'
```

test/jasmine/tests/mapbox_test.js
```diff
@@ -292,6 +292,22 @@ describe('mapbox credentials', function() {
                 lon: [10, 20, 30],
                 lat: [10, 20, 30]
             }]);
+        }).toThrow(new Error(constants.missingStyleErrorMsg));
+
+        expect(Lib.error).toHaveBeenCalledTimes(0);
```
nicolaskruchten: is this right? `Lib.error` is not called?

etpinard: Correct, we "only" throw an error.

etpinard: ... I could change that of course.

nicolaskruchten: No changes needed, just ensuring I was reading things right!

src/plots/mapbox/layout_attributes.js
nicolaskruchten: And actually I would leave the word "tile" in there, as it's not super clear what the URL should point to otherwise. We should try to say what the non-array URL should point to also: an image? a GeoJSON file?

src/plots/mapbox/layout_attributes.js
etpinard: The full text is:

https://github.com/plotly/plotly.js/blob/d81f28cf985f5ec912b83d60b0bb0b066665a6bb/src/plots/mapbox/layout_attributes.js#L40-L45

src/plots/mapbox/layout_attributes.js
etpinard: Now this is:

https://github.com/plotly/plotly.js/blob/29c2bf9ce97403ae844ac0c27c3be3d67da1948d/src/plots/mapbox/layout_attributes.js#L137-L142

src/plots/mapbox/layout_attributes.js
etpinard: Now this is:

https://github.com/plotly/plotly.js/blob/29c2bf9ce97403ae844ac0c27c3be3d67da1948d/src/plots/mapbox/layout_attributes.js#L170-L178

src/plots/mapbox/layout_attributes.js
nicolaskruchten: ah ok, i was thrown off by the parenthetical in the middle of the sentence. Works.

nicolaskruchten:
💃 thanks very much, this is a huge step up in terms of users' ability to bootstrap their understanding of how to do things :)

PR #4037, opened 2019-07-11
simplify opts passed in indicator plot

archmoj:
This PR simplifies the setup used for options passed between functions within `indicator` plot functions and reduces the number of variables in the code. 

@plotly/plotly_js


etpinard:
I have no opinion about this PR. I'll let @antoinerg and @archmoj decide what they prefer.

antoinerg:
:dancer:

Issue #3052, opened 2018-09-26
Tickformat attribute descriptions should link to d3 v3 timeformat docs


IMM0rtalis:
Tickformat with "%V" for weekdays is not working: 
https://codepen.io/anon/pen/NLQpEz

It should work as described in https://github.com/d3/d3-time-format/blob/master/README.md#locale_format

Where is the source of the formatting? Can anyone provide a link pls?

All the best

etpinard:
We're (still unfortunately) using d3 v3. So that doc page isn't adequate. We should update the `tickformat` descriptions:

https://github.com/plotly/plotly.js/blob/a070a8d2dd89f8ecbee89eecee77979298bfd886/src/plots/cartesian/layout_attributes.js#L509-L518

and link it to 

https://github.com/d3/d3-3.x-api-reference/blob/master/Time-Formatting.md#format

https://github.com/plotly/plotly.js/issues/424 is the issue regarding bumping d3.

Thanks for reporting.

alexcjohnson:
For completeness, the items d3v3 is missing that v5 has are:

`%Q` - milliseconds since UNIX epoch.

`%s` - seconds since UNIX epoch.

`%u` - Monday-based (ISO 8601) weekday as a decimal number [1,7].

`%V` - ISO 8601 week of the year as a decimal number [01, 53].

`%f` - microseconds as a decimal number [000000, 999999]. Frustratingly, d3 did something different than we did with `%f`, which we added (before it was part of d3) to support more digits and increased flexibility over `%L`. They added it as a fixed 6 digits of precision, whereas we support any precision up to 6 digits (using `%{n}f`) and we truncate trailing zeros. The fixed precision of only 3 (`%L`) or 6 digits is, IMO, a significant limitation (though I can see wanting to keep trailing zeros, we could easily add that) so I guess we'll be keeping this particular difference from D3 long-term...

Perhaps not worth it until we upgrade our d3, but it would be possible for us to add `Q,s,u,V` similarly to how we added `f`.

etpinard:
> Frustratingly, d3 did something different than we did with `%f`,
> but it would be possible for us to add `Q,s,u,V` similarly to how we added `f`.

Ha good point, I forgot about that `%f` discrepancy. Thanks for the reminder!

cooperok:
[...]

PR #4044, opened 2019-07-12
Fix tickformat attribute description links

etpinard:
See https://github.com/plotly/plotly.js/issues/3052#issuecomment-424731196

cc @plotly/plotly_js


archmoj:
Nicely done.
:dancer:

PR #4056, opened 2019-07-17
Bug fix - do not display empty bars when line.width is zero

### LINKED ISSUE #4047: Extra bar traces show up in stacked bar chart in relative barmode in >=1.48.0
For stacked bar charts with barmode='relative', super-thin bar traces show up on the plot when the bar height *should* be zero:
https://codepen.io/anon/pen/xoZOxP

This bug was introduced in 1.48.0 it seems, as the plot renders as expected in 1.47.4:
https://codepen.io/anon/pen/dBBJxW

This issue occurs in Chrome and Edge. I have not tried FF, but given the nature of the bug (extra path elements), I imagine this will occur on any browser.

archmoj:
Fixes #4047.
Empty `bar` and `funnel` rectangles should not be displayed when marker.line.width is set to zero.

[Codepen after](https://codepen.io/MojtabaSamimi/pen/bXbmba)

Also addressing #4057 for bars in this PR.
[Codepen after](https://codepen.io/MojtabaSamimi/pen/mNbYNL)

@plotly/plotly_js


src/traces/bar/plot.js
```diff
@@ -102,12 +102,23 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts) {
...
+                y0 === y1 ||
+                !isNumeric(x0) ||
+                !isNumeric(x1) ||
+                !isNumeric(y0) ||
+                !isNumeric(y1)
             );
+            // display zeros if line.width > 0
+            if(isBlank && shouldDisplayZeros) {
+                if(isHorizontal ? x1 - x0 === 0 : y1 - y0 === 0) {
+                    if(di.mlw || trace.marker.line.width > 0) {
```
etpinard: It's unclear whether this means `di.mlw || (trace.marker.line.width > 0)` or `(di.mlw || trace.marker.line.width) > 0`. Both are equivalent in this scenario, but to me 

```js
if(di.mlw > 0 || trace.marker.line.width > 0)
```

or simply

```js
if(di.mlw || trace.marker.line.width)
```

would be better.

---

Moreover, I would prefer if we combined all three `if()` into one. I don't see the need to have :three: if statements here. Oh well, this part in non-blocking. I'll let @archmoj decide what's best.

test/image/mocks/bar_display_height_zero_only_line_width.json
```diff
@@ -0,0 +1,340 @@
...
+            "insidetextanchor": "middle",
+            "cliponaxis": false,
+            "xaxis": "x4",
+            "yaxis": "y4"
+        },
+        {
+            "type": "bar",
+            "marker": {
+                "line": {
+                    "width": [10, 0, 10, 0, 10]
```
etpinard: Looks like this mock only has array `marker.line.width` value. Could you add one trace that test scaler `marker.line.width`? 

Maybe you can leave this mock alone and simply add another mock using https://codepen.io/MojtabaSamimi/pen/bXbmba , so that we also test the relative-stacked case.

etpinard:
Thanks for taking this one @archmoj - I made two minor comments.

src/traces/bar/plot.js
archmoj: @etpinard quick question:
I am wondering if `di.mlw` could be a negative number?

etpinard: Probably, as we don't coerce `arrayOk` attributes. That logic must be in a `Drawing.style*` method (please double-check though), so at this stage in `plot.js` we should **not** assume that `di.mlw >= 0`. Good eye!

archmoj:
@etpinard Thanks for the review!
This is now ready for the second round.
As I updated the PR description please note that the #4057 is also addressed for bars.

src/traces/bar/plot.js
```diff
@@ -130,8 +137,7 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts) {
                     mc = cont.color;
                 }
             } else {
-                lw = (di.mlw + 1 || trace.marker.line.width + 1 ||
-                    (di.trace ? di.trace.marker.line.width : 0) + 1) - 1;
```
archmoj: @etpinard is there a need for this second line condition? i.e. <pre>di.trace ? di.trace.marker.line.width : 0</pre>.

src/traces/bar/helpers.js
```diff
@@ -64,3 +64,12 @@ exports.getValue = function(arrayOrScalar, index) {
     else if(index < arrayOrScalar.length) value = arrayOrScalar[index];
     return value;
 };
+
+exports.getLineWidth = function(trace, di) {
+    var w =
+        (0 < di.mlw) ? di.mlw :
+        (0 < trace.marker.line.width) ? trace.marker.line.width : // Note: we don't want to get true & return an array when trace.marker.line.width is an array containing zeros.
```
etpinard: We can mostly drop this `trace.marker.line.width` will never be negative after the defaults.

src/traces/bar/helpers.js
```diff
@@ -64,3 +64,12 @@ exports.getValue = function(arrayOrScalar, index) {
     else if(index < arrayOrScalar.length) value = arrayOrScalar[index];
     return value;
 };
+
+exports.getLineWidth = function(trace, di) {
```
etpinard: Bars don't have `line.width`, but still calling this `getMarkerLineWidth` would be better.

src/traces/bar/helpers.js
archmoj: But it could be an array. For example if `line.width` is set to [0,1,2] in the first item di.mlw is zero.

src/lib/index.js
```diff
@@ -459,10 +459,10 @@ lib.noneOrAll = function(containerIn, containerOut, attrList) {
  * @param {object} cd : calcdata trace
  * @param {string} cdAttr : calcdata key
  */
-lib.mergeArray = function(traceAttr, cd, cdAttr) {
+lib.mergeArray = function(traceAttr, cd, cdAttr, isNumber) {
     if(lib.isArrayOrTypedArray(traceAttr)) {
         var imax = Math.min(traceAttr.length, cd.length);
-        for(var i = 0; i < imax; i++) cd[i][cdAttr] = traceAttr[i];
+        for(var i = 0; i < imax; i++) cd[i][cdAttr] = isNumber ? +traceAttr[i] : traceAttr[i];
```
etpinard: What we should really do here is: drop this new `isNumber` argument and call `isNumeric`.

```js
for(var i = 0; i < imax; i++) {
  var v = traceAttr[i];
  cd[i][cdAttr] = isNumerix(v) ? +v : v;
}
```

This might result in a performance drop, but this sounds like an acceptable change until we drop `gd.calcdata` all together.

src/traces/bar/plot.js
etpinard: Wow. Nice catch! That `di.trace` looks pretty odd. Maybe the old [toolpanel](https://github.com/plotly/streambed/tree/master/shelly/toolpanel/static/toolpanel/src/js) uses it. Would you wind trying to `grep` in the streambed directory?

src/lib/index.js
archmoj: What about keeping `isNumber` argument and only apply `isNumerix` on numeric fields e.g. `width` not `color`? That would help keep a better performance.

src/traces/bar/helpers.js
etpinard: True, so `Array.isArray(trace.marker.line.width` would be a better check on this line.

src/traces/bar/plot.js
archmoj: grep

src/lib/index.js
etpinard: Yeah we could do that. But remember `marker.color` can have numeric items. I presume that they are handling elsewhere (probably somewhere in `component/colorscale/`).

---

Moreover, passing a plain boolean to a function like 

```js
mergeArray(markerLine.width, cd, 'mlw', true);
```

is considered a bad pattern (to many) as you can't guess what the function does by looking at the line-of-code of the caller (here, you would need to open `src/lib/index.js` to see what the `true` does)

Here are two "better" patterns:

```js
mergeArray(markerLine.width, cd, 'mlw', {coerceNumber: true});

// or
mergeArray(markerLine.width, cd, 'mlw', function(v) { return isNumeric(v) ? +v : v; });
```

Personally I prefer the latter (as it is more general) and it consistent with how [`Lib.fillArray`](https://github.com/plotly/plotly.js/blob/master/src/lib/index.js#L479-L487) and [`Lib.castOption`](https://github.com/plotly/plotly.js/blob/218a9195832aab94b1d8bd0ca4653c159d2f25ce/src/lib/index.js#L498-L512) work.

src/traces/bar/plot.js
etpinard: and what if you grep for `Bar.plot`?

archmoj: grep_barPlot

etpinard: Great! I think we're good to :hocho: that `di.trace` condition.

src/traces/bar/helpers.js
archmoj: We also need to check for `null`s and other things. Let's keep the logic as is.

etpinard: `trace.marker.line.width` should never be `null` at this stage.

Do you have an example of that?

archmoj: You are right. Because this is coerced it shouldn't happen.

src/lib/index.js
archmoj: @etpinard thanks for the insight.
In https://github.com/plotly/plotly.js/pull/4056/commits/3ee7dbffe00cf5a899ee9858abe221e06e9e2e3f the second pattern you mentioned is applied to expand `mergeArray` functionalities.
Also in order to avoid requiring `fast-is-numeric` in `arrays_to_calcdata` steps in different traces, new functions were added to the `Lib`.

etpinard: Adding `Lib.mergeArrayCastPositive` is a very good idea. But less not add `Lib.mergeArrayCastNumber` and `Lib.mergeArrayCastIfNumber` if you're not going to use them anywhere.

src/lib/index.js
```diff
@@ -459,13 +459,39 @@ lib.noneOrAll = function(containerIn, containerOut, attrList) {
...
  */
-lib.mergeArray = function(traceAttr, cd, cdAttr, isNumber) {
+lib.mergeArray = function(traceAttr, cd, cdAttr, fn) {
+    var hasFn = typeof fn === 'function';
     if(lib.isArrayOrTypedArray(traceAttr)) {
         var imax = Math.min(traceAttr.length, cd.length);
-        for(var i = 0; i < imax; i++) cd[i][cdAttr] = isNumber ? +traceAttr[i] : traceAttr[i];
+        for(var i = 0; i < imax; i++) {
+            var v = traceAttr[i];
+            cd[i][cdAttr] = hasFn ? fn(v) : v;
```
etpinard: I wonder what's faster 

```js
var hasFn = typeof fn === 'function';
for(var i = 0; i < N; i++) {
  arrayOut[i] = hasFn ? fn(arrayIn[i]) : arrayIn[i];
}
```

or 

```js
fn = typeof fn === 'function' ? fn : Lib.noop;
for(var i = 0; i < N; i++) {
  arrayOut[i] = fn(arrayIn[i]);
}
```

No need to change anything, I'm just thinking out-loud.

src/lib/index.js
etpinard: .... unless you're planning on fixing more `Lib.mergeArray` issues in this PR :smile:

src/lib/index.js
archmoj: Interesting question. And I like the minimalist version with `Lib.noop`!
That may depend on the browser.
Having two loops may also be fast.
```js 
var i = 0;
if (typeof fn === 'function') {
  for(; i < N; ++i) arrayOut[i] = fn(arrayIn[i]);
} else {
  for(; i < N; ++i) arrayOut[i] = arrayIn[i];
}
```

src/lib/index.js
archmoj: Let's do that in a separate PR.
For now I cut them from the Lib in https://github.com/plotly/plotly.js/pull/4056/commits/579a1b55c248350de6c2d461c87ac0a05bd0eef9.

etpinard:
Nicely done :dancer: 

Thanks for the fixes!

Issue #4057, opened 2019-07-17
Problem with marker.line.width arrays containing scientific number strings


archmoj:
Scientific numbers in marker.line.width arrays may cause `bar` plot to fill the rest of plot in black!
[Codepen](https://codepen.io/MojtabaSamimi/pen/zgOXdJ?editors=0010).

etpinard:
Good eye. Replacing `"1e1"` with `1e1` gives the correct result: https://codepen.io/etpinard/pen/KOPYJo?editors=0010

I wonder what we do elsewhere with exponent-notation inside strings (e.g. `"1e1"`). Do they even pass `isNumeric`?

archmoj:
It is pretty strange. Only happens with `bar` and not with `histogram`, `funnel` or `waterfall`.
This might be a better [codepen](https://codepen.io/MojtabaSamimi/pen/oKvOOg?editors=0010) to illustrate the issue.

archmoj:
A regression may be involved here since 1.29.0:
[Codepen](https://codepen.io/MojtabaSamimi/pen/dxbEpX?editors=1000).

etpinard:
@archmoj wanna close this one too?

archmoj:
No. In respect to https://github.com/plotly/plotly.js/pull/4056#discussion_r305034311 let's keep that open until we apply similar fixes (already applied for bars) for other traces.

PR #4063, opened 2019-07-19
Filter numbers during calc step

archmoj:
Follow up of https://github.com/plotly/plotly.js/pull/4056 &
Fixes https://github.com/plotly/plotly.js/issues/4057.

Fix and apply `Lib.mergeArrayCastPositive` during `calc` step of various traces to ensure having positive inputs for size attributes.
@plotly/plotly_js

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -474,7 +474,7 @@ lib.mergeArray = function(traceAttr, cd, cdAttr, fn) {
 lib.mergeArrayCastPositive = function(traceAttr, cd, cdAttr) {
     return lib.mergeArray(traceAttr, cd, cdAttr, function(v) {
         var w = +v;
-        return isNaN(w) ? NaN : w > 0 ? w : 0;
+        return !isFinite(w) ? 0 : w > 0 ? w : 0;
     });
 };
 
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -474,7 +474,7 @@ lib.mergeArray = function(traceAttr, cd, cdAttr, fn) {
 lib.mergeArrayCastPositive = function(traceAttr, cd, cdAttr) {
     return lib.mergeArray(traceAttr, cd, cdAttr, function(v) {
         var w = +v;
-        return w > 0 ? w : 0;
+        return isNaN(w) ? NaN : w > 0 ? w : 0;
     });
 };
 
--- a/test/jasmine/tests/bar_test.js
+++ b/test/jasmine/tests/bar_test.js
@@ -412,14 +412,14 @@ describe('Bar.calc', function() {
         var gd = mockBarPlot([{
             marker: {
                 line: {
-                    width: [2, 1, 0, -1, false, true, null, [], -Infinity, Infinity, NaN, {}]
+                    width: [2, 1, 0, -1, false, true, null, [], -Infinity, Infinity, NaN, {}, '12+1', '1e1']
                 }
             },
-            y: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+            y: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
         }], {});
 
         var cd = gd.calcdata;
-        assertPointField(cd, 'mlw', [[2, 1, 0, 0, 0, 1, 0, 0, 0, Infinity, NaN, NaN]]);
+        assertPointField(cd, 'mlw', [[2, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 10]]);
     });
 });
 
--- a/test/jasmine/tests/bar_test.js
+++ b/test/jasmine/tests/bar_test.js
@@ -407,6 +407,20 @@ describe('Bar.calc', function() {
         assertPointField(cd, 'x', [[1, NaN, NaN, 15]]);
         assertPointField(cd, 'y', [[1, 2, 10, 30]]);
     });
+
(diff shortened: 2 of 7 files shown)


etpinard:
@archmoj have you noticed any broken behaviour that can be traced back to not having positive sizes in the calc step?

etpinard:
I'm thinking if might be safer to move some of these checks to `Drawing`?

archmoj:
> @archmoj have you noticed any broken behaviour that can be traced back to not having positive sizes in the calc step?

For example missing a circle in this `scatter` plot:
[before](https://codepen.io/MojtabaSamimi/pen/ymYrQO?editors=0010)
[after](https://codepen.io/MojtabaSamimi/pen/rXOboj?editors=0010)

archmoj:
> I'm thinking if might be safer to move some of these checks to `Drawing`?

I would say it is better if one checks the inputs as early as possible which is in this case the `clac` step.
But we may also add additional checks in `Drawing` to bypass values that make no sense.

etpinard:
> For example missing a circle in this `scatter` plot:

Ok nice!


Now, what about the other `calcdata` field you're sanitizing: 

- funnelarea `marker.line.width`
- scatter `textfont.size`
- scatter `marker.size`
- scatter `marker.opacity`

are they showing bugs currently on master?

I'm just trying to make sure we don't sanitize things twice.

src/lib/index.js
```diff
@@ -474,7 +474,7 @@ lib.mergeArray = function(traceAttr, cd, cdAttr, fn) {
 lib.mergeArrayCastPositive = function(traceAttr, cd, cdAttr) {
     return lib.mergeArray(traceAttr, cd, cdAttr, function(v) {
         var w = +v;
-        return w > 0 ? w : 0;
+        return isNaN(w) ? NaN : w > 0 ? w : 0;
```
etpinard: I think you want `isNumeric` here. Have tried something like

```js
'marker.line.width': [NaN, null, undefined, [], {}, 0]`
```

?

archmoj: Works OK with those cases. Using `isNumeric` might be slow.

etpinard: Have you tried running benchmarks here? I doubt that `isNumeric` is that much slower than `isNaN`

etpinard: Moreover, could we lock those cases down in a `_module.calc` test?

archmoj:
> > For example missing a circle in this `scatter` plot:
> 
> Ok nice!
> 
> Now, what about the other `calcdata` field you're sanitizing:
> 
> * funnelarea `marker.line.width`
> * scatter `textfont.size`
> * scatter `marker.size`
> * scatter `marker.opacity`
> 
> are they showing bugs currently on master?
> 
> I'm just trying to make sure we don't sanitize things twice.

Also regarding funnelarea `marker.line.width` we have {
  [codepen before](https://codepen.io/MojtabaSamimi/pen/QeyZVG?editors=0010),
  [codepen after](https://codepen.io/MojtabaSamimi/pen/oKbPZZ?editors=0010)
}

src/lib/index.js
archmoj: Tests are added in https://github.com/plotly/plotly.js/pull/4063/commits/39b9840933957640fe26e9a9f9c30e1a5625c782.

test/jasmine/tests/bar_test.js
```diff
@@ -407,6 +407,20 @@ describe('Bar.calc', function() {
...
+            marker: {
+                line: {
+                    width: [2, 1, 0, -1, false, true, null, [], -Infinity, Infinity, NaN, {}]
+                }
+            },
+            y: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+        }], {});
+
+        var cd = gd.calcdata;
+        assertPointField(cd, 'mlw', [[2, 1, 0, 0, 0, 1, 0, 0, 0, Infinity, NaN, NaN]]);
```
etpinard: Does having `NaN` is `calcdata[i][j].mlw` lead to console errors?

etpinard: console errors e.g. when drawing a svg path?

archmoj: I didn't notice any:
[Scatter codepen](https://codepen.io/MojtabaSamimi/pen/bXpGbw?editors=0010)
[Bar codepen](https://codepen.io/MojtabaSamimi/pen/MNyWwQ?editors=0010)

archmoj: But that's a very good question.
Now I am wondering if we should patch this test instead of the `lib.mergeArrayCastPositive` to accept 0 instead of `NaN`?
shouldCheckForNaN

etpinard: > I didn't notice any:

Ok, but looks like marker pts with `mlw = NaN` render as if they had `mlw = 0`, so I would prefer making `NaN` cast to `0` during calc.

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4063/commits/f9db523c7fd838da60607bd9eabf8e84190b97d9 and fixed `scattegeo` test.

etpinard:
Nicely done @archmoj - thanks very much!!

archmoj:
@etpinard may I merge this PR?

etpinard:
Oops - I forgot the :dancer:

etpinard:
Yes, please merge this PR - thanks again!

PR #4070, opened 2019-07-22
Improve axis tick increment

### LINKED ISSUE #3814: Tickmark precision with percentage formatting

archmoj:
Fixes #3814.
This PR removes some precision errors happening during tick increments on axes.
Two patches are added here:
First and most importantly we use a dynamic pattern similar to (10 * 0.1 + 10 * 0.2) / 10 to get exactly `0.3` as the result not `0.30000000000000004`.
Then we check if the length of new x (i.e. x0 + dtick) as an string is greater than the total lengths of previous value and `dtick` we round the result to 12 digits. This helps cover few edge cases e.g. where 3 * 0.3 sums up to `0.8999999999999999` not `0.9`.
To view before vs after changes please view https://github.com/plotly/plotly.js/commit/485503378319bee02acfae86ba5f716bcec9d902?short_path=4bd0c49#diff-4bd0c497a76299dd674f7b3092fb6aad.

Also to mention it appears that there is currently a bug in d3 formatting where 0.55 in `p` does not return a rounded 55%.
d3format55p.

@plotly/plotly_js


antoinerg:
> Also to mention it appears that there is currently a bug in d3 formatting where 0.55 in `p` does not return a rounded 55%.

Just to be clear, this bug exists in our outdated version of d3-format. On newer versions, this bug is gone: https://runkit.com/embed/ooxgp8bijh5w

antoinerg:
Thanks for this PR @archmoj ! It already improves things quite a bit.

Do you think it would be a lot of work to switch to the newer `d3-format` (which is its own little library) instead of using the built-in one from `d3v3`? `grep -R d3.format src/` does not return that many occurrences.

archmoj:
> Thanks for this PR @archmoj ! It already improves things quite a bit.
> 
> Do you think it would be a lot of work to switch to the newer `d3-format` (which is its own little library) instead of using the built-in one from `d3v3`? `grep -R d3.format src/` does not return that many occurrences.

> > Also to mention it appears that there is currently a bug in d3 formatting where 0.55 in `p` does not return a rounded 55%.
> 
> Just to be clear, this bug exists in our outdated version of d3-format. On newer versions, this bug is gone: https://runkit.com/embed/ooxgp8bijh5w

Thanks for the note. But that one still returns extra zeros!

archmoj:
> Thanks for this PR @archmoj ! It already improves things quite a bit.
> 
> Do you think it would be a lot of work to switch to the newer `d3-format` (which is its own little library) instead of using the built-in one from `d3v3`? `grep -R d3.format src/` does not return that many occurrences.

Good question. However; I think we should try that on a separate branch/PR. 
@antoinerg are you going to open a new issue for it?

antoinerg:
> Thanks for the note. But that one still returns extra zeros!

You're right! Sorry for the misinformation.

> Good question. However; I think we should try that on a separate branch/PR.
> @antoinerg are you going to open a new issue for it?

Sounds good. Well, if upgrading `d3-format` doesn't fix a bug, I think I can wait until we upgrade `d3`.

antoinerg:
Ok, thank you @archmoj! I think I'll let @etpinard or @alexcjohnson review this one.

src/plots/cartesian/axes.js
```diff
@@ -846,7 +847,27 @@ axes.tickIncrement = function(x, dtick, axrev, calendar) {
     var axSign = axrev ? -1 : 1;
 
     // includes linear, all dates smaller than month, and pure 10^n in log
-    if(isNumeric(dtick)) return x + axSign * dtick;
+    if(isNumeric(dtick)) {
+        // Note 1:
+        // 0.3 != 0.1 + 0.2 but 0.3 == ((10 * 0.1) + (10 * 0.2)) / 10
+        // Attempt to use integer steps to increment
+        var magic = 1 / dtick;
```
antoinerg: Do you have a reference explaining that "magic"?

archmoj: Unfortunately no. But the trick appear to be working.

etpinard: Hmm, do we really need this block for `tickformat` values other than `'p'`?

archmoj: Yes it is also required e.g. with `s`.
Please find the new mock added in https://github.com/plotly/plotly.js/pull/4070/commits/740574b301991c39b2f868355565b639ab1e0b38 which currently renders as:
before4064b

etpinard: > Yes it is also required e.g. with `s`.

Ok so could then have 

```js
if(isNumeric(dtick) && (tickformat.indexOf('s') !== -1 || tickformat.indexOf('p') !== -1)) {
  // your new logic
}
```

that would make this patch a lot less stressful to merge.

src/plots/cartesian/axes.js
```diff
@@ -846,7 +847,27 @@ axes.tickIncrement = function(x, dtick, axrev, calendar) {
...
+            magic * x +
+            magic * axSign * dtick
+        ) / magic;
+
+        // Note 2: now we may also consider rounding to cover few more edge cases
+        var lenDt = ('' + dtick).length;
+        var lenX0 = ('' + x).length;
+        var lenX1 = ('' + newX).length;
+
+        if(lenX1 > lenX0 + lenDt) { // this is likey a rounding error!
```
antoinerg: @archmoj Just a thought here: if you know the number of decimals in `dtick`, you also know that the number of decimals in the final number should never exceed it. So if `newX.split('.')[1].length > dtickDecimalLength` then you can round with a precision of `dtickDecimalLength`. I have the feeling it could fix the `55%` but I may be wrong.

archmoj: @antoinerg thanks for sharing this.
That one not only depends on the `dtick` but also on the previous tick value. For example if we start from 1000.0001 and `dtick`=1, the next tick should be at `1001.001`.

Also I should clarify that the case of `55%` is a `d3` bug which is separated from what is going to be fixed in this PR i.e. providing better raw tick values.

src/plots/cartesian/axes.js
archmoj: @antoinerg @etpinard the logic is now moved to `Lib` in https://github.com/plotly/plotly.js/pull/4070/commits/916bf80e8d26d67c474bda6547c1ef4e1e53fe50 and I improved test coverage in https://github.com/plotly/plotly.js/pull/4070/commits/86b135dccaf5848338aaa1d909e9b48c2b3fa156.

src/lib/increment_numeric.js
```diff
@@ -0,0 +1,37 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+module.exports = function incrementNumeric(x, delta) {
+    if(!delta) return x;
+
```
archmoj: One may simply `return x + delta` here to find out various failing tests.

src/lib/increment_numeric.js
```diff
@@ -0,0 +1,37 @@
...
+    // e.g. 0.3 * 3 = 0.8999999999999999
+    var lenDt = ('' + delta).length;
+    var lenX0 = ('' + x).length;
+    var lenX1 = ('' + newX).length;
+
+    if(lenX1 > lenX0 + lenDt) { // likely a rounding error!
+        newX = +parseFloat(newX).toPrecision(12);
+    }
+
+    return newX;
```
etpinard: Thanks for adding those tests @archmoj - I understand (a little bit) more what you're attempting. 

----

That said, I feel like we should be extending `numFormat`:

https://github.com/plotly/plotly.js/blob/96fe2624a0df16b50ceb3061ff18d95b6a2120da/src/plots/cartesian/axes.js#L1311-L1408

where instead of exiting early

https://github.com/plotly/plotly.js/blob/96fe2624a0df16b50ceb3061ff18d95b6a2120da/src/plots/cartesian/axes.js#L1338

for when `tickformat` is set, we could construct the formatted number ourselves for `tickformat` values that include `s` and `p`. To me, that sounds a lot more robust than these floating-point tricks that gets us a rounded-off tick step. What do you think?

archmoj: I <i>think</i> it would be interesting to fix the issue at the root cause & thanks to `JavaScript`. 
These rounding issues happen in other languages too. 
Following images illustrate examples in `Fortran` & `C++`.
fortran

cpp

etpinard: > it would be interesting to fix the issue at the root cause

What is the _root_ cause in your mind here?

src/plots/cartesian/axes.js
antoinerg: > That one not only depends on the `dtick` but also on the previous tick value. For example if we start from 1000.0001 and `dtick`=1, the next tick should be at `1001.001`.

Wouldn't taking the maximal number of decimals in either `dtick` or the `start` number work?

alexcjohnson:
@archmoj I did a little testing, and I think you may be on to something useful. AFAICT, as long as we add/subtract integers, and then divide by a power of 10 (again, explicitly rounded to be an integer), we're protected from these floating point errors over a quite wide range of values. (I tested `i/10^d`, for `d` in [1,6] and `i` in [1, 3e8], with both numerator and denominator rounded to integers, and never saw any extra digits)

For performance purposes I suspect we'd want to calculate up front the power of 10 that makes both `tick0` and `dtick` into integers (which in itself is nontrivial if you want to cover weird cases like `1.001e-6` - but I'm sure you can figure it out 😅) and round it:
`p = Math.round(Math.pow(10, digits))`
then incrementing is just:
`x1 = Math.round((x + delta) * p) / p`

Now, the question is what cases that actually fixes? We could use it to fix cases we don't handle correctly right now in our own formatting - for example `{tick0: 0.001, dtick: 1.001}` which currently just rounds to 0, 1, 2... though there may be an easier way to fix that, just by improving the logic in the [numeric case of `autoTickRound`](https://github.com/plotly/plotly.js/blob/6d09846fee9ec3a5f004839fc3122f040afd1460/src/plots/cartesian/axes.js#L836-L848) to work more like the [date case](https://github.com/plotly/plotly.js/blob/6d09846fee9ec3a5f004839fc3122f040afd1460/src/plots/cartesian/axes.js#L824-L833) (this is the tick0/dtick logic I was thinking of earlier - funny that we do this for dates but not numbers currently!)

But it won't generally fix `tickformat: 'p'` or `tickformat: 's'` (or `'e'`, `'f'`, or `'g'` for that matter). In both of these d3 internally may multiply a fraction by a power of 10 - and no matter how well we ensure the fraction has the correct string representation, we can't ensure the end product will be free of floating point rounding errors. Honestly I don't think there *is* a good solution for this - it's a problem inherent to floating point numbers, and anything we do to fix it will improperly round some cases where the user really wanted the extra precision.

Really the only solution I see is to encourage users never to use bare `p` or `s`, always include a length specifier. It's not a great solution, as it means the format can't change with zoom for example... unless we want to monkey with the d3 format language to let us insert an auto precision 😱

archmoj:
> @archmoj I did a little testing, and I think you may be on to something useful. AFAICT, as long as we add/subtract integers, and then divide by a power of 10 (again, explicitly rounded to be an integer), we're protected from these floating point errors over a quite wide range of values. (I tested `i/10^d`, for `d` in [1,6] and `i` in [1, 3e8], with both numerator and denominator rounded to integers, and never saw any extra digits)

@alexcjohnson Thanks very much for testing this.

archmoj:
> Now, the question is what cases that actually fixes? We could use it to fix cases we don't handle correctly right now in our own formatting - for example `{tick0: 0.001, dtick: 1.001}` which currently just rounds to 0, 1, 2... though there may be an easier way to fix that, just by improving the logic in the [numeric case of `autoTickRound`]

Good point. Most of these test would fail with the master version: https://github.com/plotly/plotly.js/blob/76b036859248c43518bd4f15f5dfa35444f665ed/test/jasmine/tests/lib_increment_numeric_test.js

@alexcjohnson Thank for mentioning 'e' & 'g' as well.
Having `{tick0: 0.001, dtick: 1.001, tickformat: 'e | g | p | s' } these are [before](https://codepen.io/MojtabaSamimi/pen/xvdxVZ?editors=0010) & [after](https://codepen.io/MojtabaSamimi/pen/OKmJWZ?editors=0010) codepens

etpinard:
In order to potentially get this PR merged:

- [ ] try finding papers about this. We can't be the first ones hitting this problem
- [ ] try your tests on differents browsers (maybe using https://github.com/karma-runner/karma-sauce-launcher)
- [ ] try the `55.00000000000000000` input using the latest `d3-format`. If still buggy, fill report over to d3

archmoj:
> * [ ]  try the `55.00000000000000000` input using the latest `d3-format`. If still buggy, fill report over to d3

This case is fixed in d3 > v3.

etpinard:
@archmoj I stumbled upon https://github.com/MikeMcl/decimal.js/ - which might be of interest for this PR.

archmoj:
Reworked in #5114.

Issue #1956, opened 2017-08-16
Add option to format dates as UTC


brian428:
Currently, Plotly offers the `tickformat` property to specify date formats. But the results are always shown in local time. I see no way to format the dates as UTC other than manually transforming all of the dates to offset based on the current time zone so they **look** like the equivalent UTC time. And obviously, this is a **major** pain.

D3 can certainly handle [UTC dates](https://github.com/d3/d3-time-format/blob/master/README.md#locale_utcFormat), so Plotly really needs to expose something to allow dates to be shown as UTC.

alexcjohnson:
To be precise (and yes, we're aware that this is a mess), Plotly ONLY shows dates in UTC (you'll notice every day on a plotly plot has 24 hours, for example), but if you feed in native js Date objects or milliseconds to mean dates (rather than date strings) we assume what you really wanted was the (UTC) date with the same representation as the date you provided has in the local timezone.

See #1532 for more info - one of these days we will need to build in real time zone support, but that's a substantial project (largely due to DST issues)

schwabts:
* About half a year ago I wrote the code **"Timezone Table"** below to show a table of all 24 hours of a day correctly aligned for several selected timezones in a Jupyter Notebook.
* For getting this done the **package `pendulum`** (as well as **`pytz`**) was of greatest help. It provides the abstraction of timezones from any local time to Python but making use of that in Plotly probaly involves bringing this functionality to Javascript, React, or something else I am not familiar with.
* However, I am even clueless right now about how to create a Dash app that would simply show the current (real-) time of _two **different** timezones_ simultaneously. Clearly, if we want to avoid having all display updates needed to be triggered over the net this involves clientside_callbacks, AFAIK.
* But I don't even manage to run the "[example with two Outputs](https://community.plotly.com/t/display-correct-time-in-browsers-timezone/49789/5)" in the answers to [Display correct time in browsers timezone](https://community.plotly.com/t/display-correct-time-in-browsers-timezone/49789). After adding `dcc.Interval(id="clock", interval=1000)` to the layout and brackets around the `Output`-s the remaining Error is
```
dc[namespace] is undefined
(This error originated from the built-in JavaScript code that runs Dash apps. Click to see the full stack trace or open your browser's console.)
```
But even if "the example with two Outputs" worked I believe it would display the **same** time twice. Could anyone run it sucessfully?

**Timezone Table:**

```
import pandas as pd
import pytz
import pendulum

parameters = {
[...]

Issue #1532, opened 2017-03-30
Shapes and annotations do not handle Date objects or strings with time zones


harbulot:
At the moment (version 1.25.1), data traces can handle Date objects, which will then be displayed with what seems to be the browser's timezone settings. For example, `date = new Date("2017-03-30T00:00:00.000Z")` used on the X-axis in a trace, with a browser in the UTC+1 timezone will be displayed as "Mar 30, 2017, 01:00", which is indeed correct.

However, that same date object cannot be used for `x0`/`x1` in a shape on the same object.

When using `date.toISOString()` instead (both for data and shapes), the timezone indicator (e.g. `Z`) at the end of the string is ignored and the string is considered to be within the local time zone.

Here is a jsfiddle to demonstrate the problem: https://jsfiddle.net/tkhmx8Le/

![image](https://cloud.githubusercontent.com/assets/80994/24502702/cad4dbc2-1546-11e7-8ef5-d84bcb540c0c.png)

Effectively, the problem is two-fold (and might lead to two separate issues):
* It would be good if `shapes` and `annotations` also supported date objects.
* ISO-8601 date parsing should take into account the `Z` (or generally `+xxxx` or `+xx:xx`) timezone indicator before building its internal UTC representation. This is certainly related to #1003 (and subsequent work).

alexcjohnson:
Thanks for the report @harbulot 

Yes, that's certainly inconsistent. Using `Date` objects within data traces is supported for backward compatibility, but - again for backward compatibility - it works in a strange way. I think you understand it correctly, but just for clarity for other folks reading this: we interpret the date in the local time zone but then discard the time zone and treat it as UTC, so for example if you supply a `Date` that amounts to Mar 30, 2017, 01:00 in your local time zone, we treat it as Mar 30, 2017 01:00 UTC and draw the axis as UTC.

Because of this strange behavior, it's not recommended to use `Date` objects to build plots. The present behavior is included for backward compatibility to a time before shapes existed and when annotations needed to be positioned using epoch milliseconds (in the local time zone).

With that caveat, I would be willing to consider it a bug that shapes and annotations do not support positioning by `Date` objects. It shouldn't be difficult to fix that, although we might not get to it right away (I'd be happy to review a PR to do it though!).

[...]

PR #4071, opened 2019-07-22
texttemplate: initial implementation

antoinerg:
Add `textemplate` support to traces with `textinfo`:

- [x] pie
- [x] sunburst
- [x] funnelarea
- [x] bar
- [x] waterfall
- [x] funnel
- [x] scatter


src/components/fx/texttemplate_attributes.js
```diff
@@ -0,0 +1,53 @@
...
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var FORMAT_LINK = require('../../constants/docs').FORMAT_LINK;
+
+module.exports = function(opts, extra) {
```
etpinard: `texttemplate` isn't really an _fx_, could you move this file to `plots/texttemplate_attributes.js`?

src/traces/pie/plot.js
```diff
@@ -941,8 +943,62 @@ function setCoords(cd) {
     }
 }
 
+function formatSliceLabel(gd, pt, cd0) {
```
etpinard: Very nicely done here!

src/traces/scatter/defaults.js
```diff
@@ -37,6 +37,7 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
     var defaultMode = !stackGroupOpts && (len < constants.PTS_LINESONLY) ?
         'lines+markers' : 'lines';
     coerce('text');
+    coerce('texttemplate');
```
etpinard: So here, maybe we shouldn't coerce `text` when `texttemplate` is set? What do you think @antoinerg ?

src/traces/sunburst/attributes.js
```diff
@@ -129,6 +130,9 @@ module.exports = {
         editType: 'plot',
         flags: ['label', 'text', 'value']
     }),
+    texttemplate: texttemplateAttrs({editType: 'plot'}, {
+        keys: ['label', 'text', 'value', 'color']
```
etpinard: Similar here, does anyone else think it's odd to have `%{text}` in a `texttemplate`?

test/jasmine/assets/check_texttemplate.js
```diff
@@ -0,0 +1,28 @@
+var Plotly = require('@lib/index');
+var Lib = require('@src/lib');
+var failTest = require('../assets/fail_test');
+var createGraphDiv = require('../assets/create_graph_div');
+var destroyGraphDiv = require('../assets/destroy_graph_div');
+
+'use strict';
+
+module.exports = function checkTextTemplate(mock, selector, tests) {
```
etpinard: Very nice test!

---

Could you also defaults tests for each of the traces asserting that `textinfo` isn't coerced when `texttemplate` is set?

src/lib/index.js
```diff
@@ -1005,14 +1021,15 @@ var maximumNumberOfHoverTemplateWarnings = 10;
...
- * @param {obj}     d3 locale
  * @param {string}  input string containing %{...:...} template strings
  * @param {obj}     data object containing fallback text when no formatting is specified, ex.: {yLabel: 'formattedYValue'}
+ * @param {obj}     d3 locale
  * @param {obj}     data objects containing substitution values
  *
  * @return {string} templated string
  */
-lib.hovertemplateString = function(string, labels, d3locale) {
+lib.templateFormatString = function(string, labels, d3locale) {
```
etpinard: Very nice abstraction!

Is `Lib.templateFormatString` called directly anywhere outside `src/lib/index.js`? If not, we don't have to export it on `Lib`?

src/traces/bar/plot.js
```diff
@@ -534,6 +537,64 @@ function getTextPosition(trace, index) {
     return helpers.coerceEnumerated(attributeTextPosition, value);
 }
 
+function calcTexttemplate(fullLayout, calcTrace, index, xa, ya) {
```
etpinard: It would be nice to merge this routine with `calcTextinfo` where we could first set the `obj.??Label` values and then use those to translate the `textinfo` flags.

etpinard:
Very nice PR!

Your custom jasmine assertions is very clean, but still we should add one or two mocks showing `texttemplate`.

Nice job squeezing `scatter` in here. It should be very trivial to add the other SVG `scatter*` traces. We may have to work a bit more for `scatter3d`, `scattergl` and `scatterpolargl`, but that shouldn't be too difficult as well.

----

Now, we'll have to agree on how `text` and `texttemplate` work together. To me, `texttemplate` is simply a more general `text` (as well as a more general `textinfo` :smirk: )

Some users might like to do:

```js
{
  text: ['a', 'b', 'c'],
  texttemplate: 'My text: %{text}'
}
```

which is a little easier to input than:

```js
{
  text: ['My text: a', 'My text: b', 'My text c']
}
```

**but** one could also do

```js
{
  customdata: ['a', 'b', 'c'],
  texttemplate: 'My text: %{customdata}'
}
```

So I'm thinking we shouldn't coerce `text` when `texttemplate` is set.

I'd like to know if @plotly/plotly_js agrees.

etpinard:
Oooops, I didn't notice the `texttemplate` which is very nice and very much sufficient. Ignore that part of my last comment :stuck_out_tongue:

src/traces/scatter/defaults.js
antoinerg: Very good question! Because `text` is an available flag in `textinfo`, I decided to make it available in `texttemplate` but that does not have to be the case.

test/jasmine/assets/check_texttemplate.js
antoinerg: Can I add this test in the `checkTextTemplate` routine to keep the code DRY or should I add a new entry in the appropriate `describe('Test traceName defaults')` block for each traces?

etpinard: > Can I add this test in the `checkTextTemplate` routine to keep the code DRY

Yeah, that would be fine.

antoinerg:
Thanks for review @etpinard !

I'd be OK with not making `%{text}` available in `texttemplate` but can't decide which is best. Maybe @nicolaskruchten and/or @alexcjohnson could chime (look at the second part of https://github.com/plotly/plotly.js/pull/4071#issuecomment-514244768 for details).

src/lib/index.js
antoinerg: Done in ee0a309ed478c121b6693c5833d45cf1c9e1f1ec

src/components/fx/texttemplate_attributes.js
antoinerg: Done in ce3c9cccca5b9ae3a2408681463a3b2f49d903ca

src/traces/sunburst/attributes.js
antoinerg: > **but** one could also do
> 
> ```js
> {
>   customdata: ['a', 'b', 'c'],
>   texttemplate: 'My text: %{customdata}'
> }
> ```

Actually, this isn't working in the current implementation! It needs to be fixed!

nicolaskruchten:
What's the downside of making `text` available in `texttemplate`? `customdata` is definitely workable, but if there's little downside, why frustrate the occasional user by making it not work?

etpinard:
> What's the downside of making `text` available in `texttemplate`?

- `text` sets on-graph text when `texttemplate` is omitted,
- `text` doesn't do anything when `texttemplate` is set w/o `%{text}`
- `text` could show up when `textemplate` includes a `%{text}`, but then `customdata` can do the exact same.

nicolaskruchten:
> `text` doesn't do anything when `texttemplate` is set w/o `%{text}`

That's pretty compelling.

src/traces/scatter/defaults.js
antoinerg: > So here, maybe we shouldn't coerce `text` when `texttemplate` is set?

If we remove `%{text}` from `texttemplate`, we still need to coerce it if `hovertemplate` is set.

etpinard: Hmm, very good point. 

I guess it would kinda of a mess to have logic for: 

_Do not coerce `text` is `texttemplate` is set OR `hovertemplate` is set_

---

So, we should coerce `text` always, but make it clear in the `texttemplate` attribute description that if `text` and `texttemplate` are set, `texttemplate` is used for on-graph text and `text` entries are available via `%{text}`.

nicolaskruchten: > So, we should coerce text always, but make it clear in the texttemplate attribute description that if text and texttemplate are set, texttemplate is used for on-graph text and text entries are available via %{text}.

I like this.

etpinard: Cool. Let's go with that!

antoinerg:
@etpinard In `hovertemplate`, should `%{text}` be the output string of `texttemplate`?

> Why did you do this? `hovertemplate` has
> 
> https://github.com/plotly/plotly.js/blob/a9e74805a18fc4e8ed3ab57bf501661a2c656762/src/components/fx/hovertemplate_attributes.js#L35
> 
> Both `dflt: ''` or not setting `dflt` at all seem correct to me (as `''` is falsy), but we should be consistent.

Sorry for the inconsistency. I guess I should stick with `''` since we already have it?

etpinard:
> Sorry for the inconsistency. I guess I should stick with `''` since we already have it?

Sure :ok_hand:

src/traces/sunburst/attributes.js
antoinerg: Fixed in e0bde45fe84d78d07f4d02882cd76a62f0345fa0

test/jasmine/assets/check_texttemplate.js
antoinerg: Done in https://github.com/plotly/plotly.js/commit/9330290d01efb3e24aa2f39b37d2feb34237ab73

antoinerg:
I tried to add support for all scatter traces in https://github.com/plotly/plotly.js/pull/4071/commits/902d2d37c276a9858beacb818f7ca250ac25f073 but gave up the ones using WebGL or mapbox because there work a bit differently.

src/traces/scattergeo/attributes.js
```diff
@@ -64,6 +65,9 @@ module.exports = overrideAll({
             'these elements will be seen in the hover labels.'
         ].join(' ')
     }),
+    texttemplate: texttemplateAttrs({editType: 'plot'}, {
+        keys: ['latlon', 'text']
```
etpinard: Hmm. This should be 

```js
keys: ['lat', 'lon', 'location', 'text']
```

similar to the `hovertemplate` keys

src/components/drawing/index.js
```diff
@@ -689,15 +689,23 @@ drawing.textPointStyle = function(s, trace, gd) {
...
         if(!text && text !== 0) {
             p.remove();
             return;
         }
 
+        if(template) {
+            var extra = {};
+            if(d.tx) extra.text = d.tx;
+            if(d.data) extra.customdata = d.data;
+            text = Lib.texttemplateString(text, {}, gd._fullLayout._d3locale, d, trace._meta || {}, extra);
```
etpinard: Hmm, sorry for just noticing this now. You're using the fields directly off `calcdata[i][j]` which have gone through `ax.d2c` 

For example, off this branch:

```js
Plotly.newPlot(gd, [{
  type: 'bar',
  textposition: 'outside',
  x: ['2019-01-01', '2019-02-01'],
  y: [1, 2],
  hovertemplate: '%{x}',
  texttemplate: '%{x}'
}])
```

gives:

Peek 2019-07-24 09-25

Instead, we should be using the `d` (d for data) filled in `fullData` - for values w/o a corresponding `textinfo` flag.

Hover uses [`appendArrayPointValue`](https://github.com/plotly/plotly.js/blob/96fe2624a0df16b50ceb3061ff18d95b6a2120da/src/components/fx/helpers.js#L159-L184) - maybe it could be reused here when `texttemplate` is set.

etpinard:
Dropping from 1.49.0

src/components/drawing/index.js
antoinerg: > Hover uses [`appendArrayPointValue`](https://github.com/plotly/plotly.js/blob/96fe2624a0df16b50ceb3061ff18d95b6a2120da/src/components/fx/helpers.js#L159-L184) - maybe it could be reused here when `texttemplate` is set.

Using `appendArrayPointValue` improves things a lot: we get `customdata`, `lat`, `lon`, `location` and other variables available in the `texttemplate` without additional code. For the specific mock you provided, I get:
newplot - 2019-07-24T125556 290
which is already a nice improvement. Should I aim to get the exact same formatting as in `hovertemplate`?

etpinard: Very good question. I can upsides to showing `Jan 1, 2019` and to showing `2019-01-01`.

Could be a good thing to discuss in tomorrow's meeting.

src/traces/scattergeo/attributes.js
antoinerg: Fixed by 0b516f7086c2fff2e9e4009234cc804169eb2aa1 and tested in https://github.com/plotly/plotly.js/commit/7d216806f075a938bc0fbdbd9149ee09c4bac027

antoinerg:
Some progress on the `gl` front can be seen here: https://github.com/plotly/plotly.js/compare/texttemplate...texttemplate-gl . I'm not sure it's the best place to handle texttemplate :thinking: so let's discuss about it.

Thanks again for the review and comments @etpinard

etpinard:
> I'm not sure it's the best place to handle texttemplate  so let's discuss about it.

Nothing to discuss here. You found the perfect place :ok_hand: , just make sure to call `Plotly.purge(gd)` before you `destroyGraphDiv();` in your `checktemplate` wrapper.

src/components/drawing/index.js
antoinerg: As discussed, we should add support formatting dates using `d3-time-format` when a pipe operator is present ex.: `%{x|yyyy-mm-dd}`

src/lib/index.js
```diff
@@ -1067,7 +1067,7 @@ function templateFormatString(string, labels, d3locale) {
 
         if(format) {
             var fmt;
-            if(d3locale) {
+            if(d3locale && d3locale.numberFormat) {
```
etpinard: Why do we need this extra condition now?

test/jasmine/assets/check_texttemplate.js
```diff
@@ -27,17 +31,30 @@ module.exports = function checkTextTemplate(mock, selector, tests) {
     tests.push(['%{customdata}', customdata]);
 
     tests.forEach(function(test) {
-        it('should support texttemplate', function(done) {
+        it(isGL ? '@gl' : '' + 'should support texttemplate', function(done) {
```
etpinard: Could you split the non-gl from the gl test in two separate `it`s (see [how `select_test.js` does it](https://github.com/plotly/plotly.js/blob/ffa6e3468b592e45db2a00bb65805a92fd59c4fb/test/jasmine/tests/select_test.js#L619-L664) . Combining them here will throw off test [sharding](https://github.com/plotly/plotly.js/blob/master/tasks/shard_jasmine_tests.js)

src/traces/scatter/attributes.js
```diff
@@ -166,6 +167,10 @@ module.exports = {
             'these elements will be seen in the hover labels.'
         ].join(' ')
     },
+
+    texttemplate: texttemplateAttrs({arrayOk: true}, {
```
etpinard: I think you can :hocho: that `{arrayOk: true}` part. Doesn't `texttempalteAttrs` return an object with `arrayOk: true` by default?

src/lib/index.js
antoinerg: I'll remove this in a future commit. I added it to my TODO above:
> for GL traces, pass through the current d3locale to texttemplateString

test/jasmine/assets/check_texttemplate.js
antoinerg: @etpinard can you have a look at commit https://github.com/plotly/plotly.js/commit/05f472229f99e0f6bd82200f6192a7ec2fb17638 ? Are the non-gl tests corectly split from the gl?

src/components/drawing/index.js
antoinerg: @etpinard Should I handle this in a separate PR?

test/jasmine/assets/check_texttemplate.js
etpinard: Hmm. Come to think of it, we might have to refactor `checkTextTemplate` wrapper so that each `it` is in the `_test.js` files, not in the `assets/` folder to make test sharding work as expected.

Would it be too much to ask to have (e.g. for the `sunburst_test.js` suite):

```js
describe('Test sunburst texttemplate:', function() {
  var trace = {
        type: 'sunburst',
        labels: ['Eve', 'Cain', 'Seth', 'Enos', 'Esther'],
        values: [11, 12, 13, 14, 15],
        text: ['1', '2', '3', '4', '5'],
        parents: ['', 'Eve', 'Eve', 'Seth', 'Seth' ]
   }

   var selector = 'g.slicetext'
   
   [
      ['txt: %{label}', ['txt: Eve', 'txt: Cain', 'txt: Seth', 'txt: Enos', 'txt: Esther']],
      [['txt: %{label}', '%{text}', 'value: %{value}'], ['txt: Eve', '2', 'value: 13', '', '']],
      ['%{color}', ['rgba(0,0,0,0)', '#1f77b4', '#ff7f0e', '#1f77b4', '#1f77b4']]
    ].forEach(function(test, i) {
      it('should support texttemplate - case ' + i, function(done) {
          checkTextTemplate([trace], selector, test).catch(failTest).then(done);
      });
    });
});
```

That way all the `it`s would remain in `test/jasmine/tests/` and sharding would work as expected.


You could combine the "should not coerce textinfo when texttemplate" `it` inside `checkTextTemplate` or have `assets/check_texttemplate.js` export two methods e.g. `checkTextTemplateDefaults` and `checkTextTemplateRendering`

src/lib/index.js
antoinerg: Fixed in cbbcfcbe3635eb34a51ef2e8a8b35f0c885c7323

src/traces/scatter/attributes.js
antoinerg: Done in cbbcfcbe3635eb34a51ef2e8a8b35f0c885c7323

test/jasmine/assets/check_texttemplate.js
antoinerg: Ok, I now understand how `shard_jasmine_tests.js` proceeds. Unfortunately, it misses out on `it` blocks that are within required files or in `for` loops. The latter is a common pattern used throughout our tests: https://github.com/plotly/plotly.js/blob/ffa6e3468b592e45db2a00bb65805a92fd59c4fb/test/jasmine/tests/scatterpolargl_test.js#L276-L283

I think we should find a more robust way of counting the number of `it` blocks in test files... :thinking:

src/components/drawing/index.js
etpinard: > Should I handle this in a separate PR?

I would prefer having just one texttemplate PR, but that's up to you.

test/jasmine/assets/check_texttemplate.js
etpinard: > Unfortunately, it misses out on `it` blocks that are within required files or in `for` loops.

Does it really _miss_ out? Or does it just count them as one `it` test?

etpinard: > I think we should find a more robust way of counting the number of `it` blocks in test files...

I agree. The best way to do this would be to write a [karma plugin](http://karma-runner.github.io/4.0/dev/plugins.html) for this where we would have access to all the "generated" `it`s at runtime.

src/traces/scattergl/convert.js
```diff
@@ -111,7 +113,27 @@ function convertTextStyle(trace) {
...
     var i;
 
-    optsOut.text = trace.text;
+    var texttemplate = trace.texttemplate;
+    if(texttemplate) {
+        var d3locale = gd._fullLayout._d3locale;
+        optsOut.text = [];
+        var pt;
+        if(Array.isArray(texttemplate)) {
+            for(i = 0; i < texttemplate.length; i++) {
```
etpinard: So here you're filling up to `texttemplate.length`. What does:

```js
Plotly.newPlot(gd, [{
  type: 'scattergl',
  y: [1, 2, 1],
  texttemplate: ['%{y} a', '%{y} b']
}])
```

show when hovering over the third pt?

antoinerg: It errors out at the moment. Thanks for pointing this out...

antoinerg: Although it seems like on `master`, it also gives out an error if the `text` is shorter than the data array:
Screenshot_2019-08-13_14-10-16

test/jasmine/assets/check_texttemplate.js
antoinerg: @etpinard upon a second look, I'm not even sure it's worth counting the `it` at all considering most of the time there's only one file per shard (see below). Therefore, one could argue that all we need is to filter out files without any tests matching a given tag in order to skip the wasteful build/run step.

### Breakdown of shards
The folllowing line:
https://github.com/plotly/plotly.js/blob/c91f88b0f6f41bdfda46ecf28be06bf534d993c0/.circleci/test.sh#L56
yields:
```
contourgl_test.js
gl2d_click_test.js
gl2d_plot_interact_test.js
gl3d_hover_click_test.js
gl3d_plot_interact_test.js
mapbox_test.js
mesh3d_test.js
parcoords_test.js
plot_api_react_test.js
scatter3d_test.js
scattergl_select_test.js
scattergl_test.js
scattermapbox_test.js
splom_test.js
streamtube_test.js
surface_test.js
cone_test.js
scatterpolargl_test.js
isosurface_test.js,cartesian_test.js
volume_test.js,colorscale_test.js
select_test.js,gl3dlayout_test.js
pointcloud_test.js,modebar_test.js
choroplethmapbox_test.js,densitymapbox_test.js
```
Except for the last 5 lines, each shard only has one file.

The following line: https://github.com/plotly/plotly.js/blob/c91f88b0f6f41bdfda46ecf28be06bf534d993c0/.circleci/test.sh#L67
yields:
```
config_test.js
select_test.js
shapes_test.js
animate_test.js
polar_test.js
```
Then again, each shard only has one file.

etpinard: > considering most of the time there's only one file per shard (see below).

_currently_ there's only one file per shard, but there's nothing guaranteeing that we won't change `--limit` value in the future.

Oh well, I think having all the `it`s in `test/jasmine/tests/` is a good idea regardless, but I guess we can leave that for another PR if you prefer.

antoinerg: I understand that we may want to increase that limit.

> I guess we can leave that for another PR if you prefer.

Cool, I think I would prefer to tackle this in another PR considering there are other files under `test/jasmine/assets/` that are not related to this PR and contain `it` specs.

src/components/drawing/index.js
antoinerg: If I understand correctly, we have limited support for timezones at the moment (https://github.com/plotly/plotly.js/issues/1532, https://github.com/plotly/plotly.js/issues/1956) so is the following all we need https://github.com/plotly/plotly.js/compare/texttemplate...template-format-date ?

etpinard: Comment in -> https://github.com/plotly/plotly.js/commit/3a59d94281a19d90d220f5ae072a7dd6a5d3d7a7#r34758960

etpinard: Comment in -> https://github.com/plotly/plotly.js/commit/3a59d94281a19d90d220f5ae072a7dd6a5d3d7a7#r34758960

src/traces/funnelarea/attributes.js
```diff
@@ -55,15 +55,15 @@ module.exports = {
     }),
 
     texttemplate: texttemplateAttrs({editType: 'plot'}, {
-        keys: ['label', 'color', 'value', 'percent', 'text']
+        keys: ['label', 'color', 'value', 'text', 'percentInitial', 'percentTotal', 'percentPrevious']
```
archmoj: Those should be fixed in `funnel` not `funnelarea`.

src/traces/funnelarea/attributes.js
```diff
@@ -55,15 +55,15 @@ module.exports = {
...
     }),
 
     hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
         flags: ['label', 'text', 'value', 'percent', 'name']
     }),
 
     hovertemplate: hovertemplateAttrs({}, {
-        keys: ['label', 'color', 'value', 'percent', 'text']
+        keys: ['label', 'color', 'value', 'text', 'percentInitial', 'percentTotal', 'percentPrevious']
```
archmoj: Also here:
Those should be fixed in `funnel` not `funnelarea`.

src/traces/funnelarea/attributes.js
antoinerg: My bad! Fixed in 8c339df8c17b60af28ad38831924b03ec1141acc

src/traces/funnelarea/attributes.js
antoinerg: Fixed in 8c339df8c17b60af28ad38831924b03ec1141acc

src/traces/bar/plot.js
```diff
@@ -555,6 +560,66 @@ function getTextPosition(trace, index) {
...
+    function formatNumber(v) {
+        var sAxis = isHorizontal ? xa : ya;
+        return tickText(sAxis, +v, true).text;
+    }
+
+    var cdi = calcTrace[index];
+    var obj = {};
+
+    obj.label = cdi.p;
+    obj.labelLabel = formatLabel(cdi.p);
```
antoinerg: This line sets the default for `label` when no formatting is provided in the template. If it's a date, it will automatically get the default date formatting (ie. `%b %-d, %Y`).

etpinard:
Awesome PR @antoinerg !!

:dancer: :dancer: :dancer: - let's get the 1.50.0 :train: up and running!

PR #4081, opened 2019-07-26
Revise the output of bar text transform functions

archmoj:
This PR makes `bar` transform text functions useful for transitions and on par with `pie` versions. 
Having an `object` instead of `string` in the result, one could read properties such as `scale` and work with them as needed in transitions of `treemap` traces.

@plotly/plotly_js


etpinard:
Ok, looks good to me :dancer:

PR #4082, opened 2019-07-26
Bug fix parcoords when all the data is zero

### LINKED ISSUE #4080: Parcoords plot causes out of memory error or browser hang if data is all zeroes
Parallel coordinates plots cause browser out of memory errors (on Chrome) or browser hangs (on Firefox) if a dimension's values are all zeroes and the dimension's range is not specified or also set to `[0, 0]`.

Minimal example:  https://jsfiddle.net/JonathonW/p5rj2uv4/8/

This issue is not reproducible if the range is explicitly set (to something other than `[0, 0]`), or if the dimension's values are non-zero (even if they are the same--  `value: [1, 1]` does not cause the same problem).

archmoj:
Fixes #4080 for `parcoords`, also
adds a condition check in `Axes` to avoid encountering `log(0)` in computing range-exponent.

@plotly/plotly_js

### DIFF
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -839,7 +839,7 @@ function autoTickRound(ax) {
         // 2 digits past largest digit of dtick
         ax._tickround = 2 - Math.floor(Math.log(dtick) / Math.LN10 + 0.01);
 
-        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1]));
+        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1])) || 1;
 
         var rangeexp = Math.floor(Math.log(maxend) / Math.LN10 + 0.01);
         if(Math.abs(rangeexp) > 3) {
--- a/src/plots/cartesian/axes.js
+++ b/src/plots/cartesian/axes.js
@@ -839,8 +839,7 @@ function autoTickRound(ax) {
         // 2 digits past largest digit of dtick
         ax._tickround = 2 - Math.floor(Math.log(dtick) / Math.LN10 + 0.01);
 
-        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1])) || 1;
-
+        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1]));
         var rangeexp = Math.floor(Math.log(maxend) / Math.LN10 + 0.01);
         if(Math.abs(rangeexp) > 3) {
             if(isSIFormat(ax.exponentformat) && !beyondSI(rangeexp)) {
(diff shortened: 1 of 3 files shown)


src/plots/cartesian/axes.js
```diff
@@ -839,7 +839,7 @@ function autoTickRound(ax) {
         // 2 digits past largest digit of dtick
         ax._tickround = 2 - Math.floor(Math.log(dtick) / Math.LN10 + 0.01);
 
-        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1]));
+        var maxend = Math.max(Math.abs(rng[0]), Math.abs(rng[1])) || 1;
```
etpinard: Hmm. I'm not a fan of this. We shouldn't have to patch this deep in `axes.js`. If cartesian axes don't need this `|| 1` fallback, `parcoords` shouldn't need it.

Could you help understand why we need to `log` a range value of 0?

archmoj: I could simply remove this from here. But I thought we should check for that to avoid log(0).

etpinard: Ha I see, https://github.com/plotly/plotly.js/pull/4082/commits/fde0d320b747abbb3f92da79bd0fc7bbd1624509 is enough to fix the problem. Sorry for the confusion.

archmoj: Yes. That should be enough. Do you want me to undo https://github.com/plotly/plotly.js/pull/4082/commits/765ca5a62b88eb94bf64dfe32f1eb364d5aff0cb?

etpinard: If `maxend` is 0 here, it leads to a pretty serious malfunction for this routine. Just mocking it to `1` isn't really the behaviour we want. In fact, the fallback is just there to not make `Math.log` blowup. Cartesian (and polar and maybe other axes) ranges are sanatised during [`ax.cleanRange`](https://github.com/plotly/plotly.js/blob/38cd9e26e4e05c3d175c5b251b2c9a213f03ee56/src/plots/cartesian/set_convert.js#L387-L449), so `maxend=0` never happens and the fallback is never used, so I'd be ok with dropping the fallback.

That said, since we're starting to reuse the more of the `Axes` logic, I think it would be best to log and error using `Lib.error` - similar to

https://github.com/plotly/plotly.js/blob/38cd9e26e4e05c3d175c5b251b2c9a213f03ee56/src/plots/cartesian/axes.js#L935

So @archmoj you could _just_ drop it, or drop it and add a `Lib.error` when `maxend === 0`

archmoj: @etpinard thanks for your review.
Dropped condition check of `Axes` in https://github.com/plotly/plotly.js/pull/4082/commits/096ed736d27392563f47741fc482a52937bfa118.

etpinard:
Thanks very much !! :dancer: :dancer:

PR #4083, opened 2019-07-27
Parcoords bug fix - do not draw full select line when constraint range falls below range

archmoj:
Fixes https://github.com/plotly/phoenix-integration/issues/240.
i.e. the Parallel Coordinates plot draws the pink constraints line so that it spans the entire length of the dimension when the constraint range falls below the range.

[codepen Before](https://codepen.io/MojtabaSamimi/pen/VomxMZ)
[codepen After](https://codepen.io/MojtabaSamimi/pen/JgbvrV)

@plotly/plotly_js


etpinard:
Brilliant fix and test mock :dancer:

PR #4090, opened 2019-07-30
Bump scatter3d to 1.2.2 

archmoj:
i.e to remove an unused variable from the shader
https://github.com/gl-vis/gl-scatter3d/commit/a07e1fe5624c9376227b795cfed9b51ce165f107

@etpinard @antoinerg

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4697,9 +4697,9 @@
       }
     },
     "gl-scatter3d": {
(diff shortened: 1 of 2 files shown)


antoinerg:
Looks like we don't need that variable! Nice catch @archmoj 

:dancer:

PR #4094, opened 2019-07-31
Pin down mapbox-gl version

etpinard:
so that `npm i plotly.js` always results in installing the correct `mapbox-gl` version

cc @plotly/plotly_js


archmoj:
:dancer:

PR #4096, opened 2019-07-31
Gl3d hover bug fix - pick most recent point if two share identical position

archmoj:
Addressing https://github.com/plotly/phoenix-integration/issues/250
where in the scenes with transparent `traces` (or `axes`), the hover algorithm was not able to pick the latest data points if they were on positions similar to the previous points. 

The Z-Fighting bug in `gl-plot3d`  module could be fixed by https://github.com/gl-vis/gl-plot3d/pull/20.

[Codepen before](https://codepen.io/MojtabaSamimi/pen/PMjbOZ)
[Codepen after](https://codepen.io/MojtabaSamimi/pen/QegGdd)

@plotly/plotly_js


etpinard:
Nicely done :dancer: !!

PR #4099, opened 2019-08-01
Use Axes.prepTicks not Axes.calcTicks for indicator number and delta

etpinard:
cc @plotly/plotly_js - I totally forgot about [`Axes.prepTicks`](https://github.com/plotly/plotly.js/blob/dab7d8365e87f43f829124803261c4ae1c566c28/src/plots/cartesian/axes.js#L500-L546) which does all the things required to then call `Axes.tickText` w/o having to compute tick-value arrays like in `Axes.calcTicks`. 

`Axes.prepTicks` is currently used in the contour and carpet code (that I've been staring at for ~1 week) and should lead to an enjoyable performance boost.


archmoj:
@etpinard Thanks for the info and fix.
:dancer:

PR #4100, opened 2019-08-02
using Attrs as the short form of Attributes everywhere

archmoj:
There were few cases in the code where `Atts` is used next to `Attrs`. 
While both refer to `Attributes` the `Atts` forms are simply replaced in this PR.

@plotly/plotly_js


etpinard:
thanks :dancer:

PR #4104, opened 2019-08-03
refactoring drawing index

archmoj:
This PR refactores `drawing` index file so that it would have a clear module interface.

@plotly/plotly_js 

Also to note: 
- Commit e14338e0ac97f24f9c312cf9006f178e9041e23b avoids keeping extra arrays in memory.
- Commit ef6206c2225a727929a101625ea631b2ea5b933f fixes a bug where falsely hashes in some scenarios could have increased the counter and result in removing all the saved bounding boxes.
https://github.com/plotly/plotly.js/blob/ffa6e3468b592e45db2a00bb65805a92fd59c4fb/src/components/drawing/index.js#L982-L983

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -27,14 +27,18 @@ var DESELECTDIM = require('../../constants/interactions').DESELECTDIM;
 var subTypes = require('../../traces/scatter/subtypes');
 var makeBubbleSizeFn = require('../../traces/scatter/make_bubble_size_func');
 
+var symbolList = [];
+var symbolNames = [];
+var symbolFuncs = [];
+
 module.exports = {
     // data:
     tester: undefined,
     testref: undefined,
 
-    symbolList: [],
-    symbolNames: [],
-    symbolFuncs: [],
+    symbolList: symbolList,
+    symbolNames: symbolNames,
+    symbolFuncs: symbolFuncs,
     symbolNoDot: {},
     symbolNoFill: {},
     symbolNeedLines: {},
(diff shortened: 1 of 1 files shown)


src/components/drawing/index.js
```diff
@@ -27,13 +27,67 @@ var DESELECTDIM = require('../../constants/interactions').DESELECTDIM;
 var subTypes = require('../../traces/scatter/subtypes');
 var makeBubbleSizeFn = require('../../traces/scatter/make_bubble_size_func');
 
-var drawing = module.exports = {};
+module.exports = {
+    // data:
+    tester: this.tester,
+    testref: this.testref,
```
etpinard: What's `this` in the case? The module scope function? If so, that's a bad pattern, because it would probably brake things when bundling the code using bundler that does not scope every module into its own function.

src/components/drawing/index.js
```diff
@@ -203,42 +257,35 @@ drawing.fillGroupStyle = function(s) {
...
-drawing.symbolFuncs = [];
-drawing.symbolNeedLines = {};
-drawing.symbolNoDot = {};
-drawing.symbolNoFill = {};
-drawing.symbolList = [];
-
 Object.keys(SYMBOLDEFS).forEach(function(k) {
     var symDef = SYMBOLDEFS[k];
-    drawing.symbolList = drawing.symbolList.concat(
+    module.exports.symbolList = module.exports.symbolList.concat(
```
etpinard: Can't you just define `symbolList` above the `module.exports = {/* */}` and use `symbolList` instead of `module.exports.symbolList`?

etpinard:
> This PR refactores hover index file

By _hover_, you mean _drawing_, or is there another commit coming?

src/components/drawing/index.js
archmoj: Good catch.
Fixed in https://github.com/plotly/plotly.js/pull/4104/commits/e9fe58ccb74f4cfce33d9973c7f4925dd1f77127.

archmoj:
> > This PR refactores hover index file
> 
> By _hover_, you mean _drawing_, or is there another commit coming?

Oops. The `Drawing` is correct; and the `Hover` is wrong. 
Fixed the PR title and description accordingly.

src/components/drawing/index.js
archmoj: The short answer ifs`No`; since the `symbolList` and friends exists as local variables too.

archmoj: Hmm. I think I could fix that one too.

src/components/drawing/index.js
```diff
@@ -875,11 +926,28 @@ drawing.makeTester = function() {
  * @return {object}: a plain object containing the width, height, left, right,
  *   top, and bottom of `node`
  */
-drawing.savedBBoxes = {};
-var savedBBoxesCount = 0;
-var maxSavedBBoxes = 10000;
+var MAX_SAVED_BBOXES = 10000;
```
archmoj: I suggest using a power of two for this constant e.g. 8192 or 4096 for faster get/set results.

PR #4105, opened 2019-08-03
Process data uri images synchronously

jonmmease:
This PR is the result of some time spent debugging an odd behavior I was seeing when using plotly.js layout images to display interactive results generated by [Datashader](http://datashader.org/).

The basic idea is that every time the viewport of a plot changes (pan or zoom), a new image is generated by datashader to be displayed for that viewport.  By encoding these images a data uris, I've been able to display them pretty successfully using `layout.image` objects.

But one problem has been that sometimes the images appear to jump right before they are updated.  Here's what that looks like.

datashader_jump

After playing with this for a while, I realized that what's happening is that the image position (with the old image) is getting updated a moment before the new image is displayed.  Here's a pure JavaScript example that shows the problem a little more clearly. (gist: https://gist.github.com/jonmmease/d2793251fdb1f370e483d54ca6561690).  In this snippet, each time the viewport changes, `Plotly.react` is used to update the image with a new `source` and a new position, where the position is always the bottom right corner of the viewport.

datauri_before

Notice how the old image is moved to the bottom right corner before the new image is displayed.

After digging into plotly.js a bit, I realized that the reason this is happening is that the displayed image is updated in an `img.onload` callback that is executed after the image has loaded.   This callback then builds and saves the data uri form of the image using a canvas element.

This PR makes a small adjustment to bypass the `onload` callback for images that are specified as data uri's already.  In this case we can use the `layout.image.source` string directly.  With this change, images specified as data uris will update at the same time that their position is updated. This change is in commit 536c90e. Commit 0a397c4 contains the indentation that I omitted from 536c90e in order to make a cleaner diff of the logic changes.

Here is what the example above looks like in this PR:
datauri_after

Let me know if there are any specific tests you'd like to see for this change.  I did notice that there is already a data uri image being tested by the `layout_image` mock.

### DIFF
--- a/src/components/images/draw.js
+++ b/src/components/images/draw.js
@@ -84,7 +84,7 @@ module.exports = function draw(gd) {
             var img = new Image();
             this.img = img;
 
-            if (d.source && d.source.slice(0, 5) === 'data:') {
+            if(d.source && d.source.slice(0, 5) === '') {
                 img.src = d.source;
                 thisImage.attr('xlink:href', d.source);
                 resolve();
--- a/src/components/images/draw.js
+++ b/src/components/images/draw.js
@@ -89,31 +89,31 @@ module.exports = function draw(gd) {
                 thisImage.attr('xlink:href', d.source);
                 resolve();
             } else {
-            // If not set, a `tainted canvas` error is thrown
-            img.setAttribute('crossOrigin', 'anonymous');
-            img.onerror = errorHandler;
-            img.onload = function() {
-                var canvas = document.createElement('canvas');
-                canvas.width = this.width;
-                canvas.height = this.height;
+                // If not set, a `tainted canvas` error is thrown
+                img.setAttribute('crossOrigin', 'anonymous');
+                img.onerror = errorHandler;
+                img.onload = function() {
+                    var canvas = document.createElement('canvas');
+                    canvas.width = this.width;
+                    canvas.height = this.height;
 
-                var ctx = canvas.getContext('2d');
-                ctx.drawImage(this, 0, 0);
+                    var ctx = canvas.getContext('2d');
+                    ctx.drawImage(this, 0, 0);
 
-                var dataURL = canvas.toDataURL('image/png');
+                    var dataURL = canvas.toDataURL('image/png');
 
-                thisImage.attr('xlink:href', dataURL);
+                    thisImage.attr('xlink:href', dataURL);
 
-                // resolve promise in onload handler instead of on 'load' to support IE11
-                // see https://github.com/plotly/plotly.js/issues/1685
-                // for more details
-                resolve();
-            };
+                    // resolve promise in onload handler instead of on 'load' to support IE11
+                    // see https://github.com/plotly/plotly.js/issues/1685
+                    // for more details
+                    resolve();
+                };
 
 
-            thisImage.on('error', errorHandler);
+                thisImage.on('error', errorHandler);
 
-            img.src = d.source;
+                img.src = d.source;
             }
 
             function errorHandler() {
--- a/src/components/images/draw.js
+++ b/src/components/images/draw.js
@@ -84,7 +84,7 @@ module.exports = function draw(gd) {
             var img = new Image();
             this.img = img;
 
-            if(d.source && d.source.slice(0, 5) === '') {
+            if(d.source && d.source.slice(0, 5) === 'data:') {
                 img.src = d.source;
                 thisImage.attr('xlink:href', d.source);
                 resolve();
--- a/src/components/images/draw.js
+++ b/src/components/images/draw.js
@@ -84,6 +84,11 @@ module.exports = function draw(gd) {
             var img = new Image();
             this.img = img;
 
+            if (d.source && d.source.slice(0, 5) === 'data:') {
+                img.src = d.source;
+                thisImage.attr('xlink:href', d.source);
+                resolve();
+            } else {
             // If not set, a `tainted canvas` error is thrown
             img.setAttribute('crossOrigin', 'anonymous');
             img.onerror = errorHandler;
@@ -109,6 +114,7 @@ module.exports = function draw(gd) {
             thisImage.on('error', errorHandler);
 
             img.src = d.source;
+            }
 
             function errorHandler() {
                 thisImage.remove();
--- a/src/components/images/draw.js
+++ b/src/components/images/draw.js
@@ -74,21 +74,20 @@ module.exports = function draw(gd) {
     function setImage(d) {
         var thisImage = d3.select(this);
 
-        if(this.img && this.img.src === d.source) {
+        if(this._imgSrc === d.source) {
             return;
         }
 
         thisImage.attr('xmlns', xmlnsNamespaces.svg);
 
-        var imagePromise = new Promise(function(resolve) {
-            var img = new Image();
-            this.img = img;
+        if(d.source && d.source.slice(0, 5) === 'data:') {
+            thisImage.attr('xlink:href', d.source);
+            this._imgSrc = d.source;
+        } else {
+            var imagePromise = new Promise(function(resolve) {
+                var img = new Image();
+                this.img = img;
 
-            if(d.source && d.source.slice(0, 5) === 'data:') {
-                img.src = d.source;
-                thisImage.attr('xlink:href', d.source);
-                resolve();
-            } else {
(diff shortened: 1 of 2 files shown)


jonmmease:
I'm not sure what's going on with the tests.
```
    ✗ @gl should update the scene camera - orthographic case (3494ms)
	Error: Timeout - Async callback was not invoked within 3000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```

And another timeout.  I already tried restarting them once with no change. Any ideas?

etpinard:
> And another timeout. I already tried restarting them once with no change. Any ideas?

Tests are passing now.

src/components/images/draw.js
```diff
@@ -84,31 +84,37 @@ module.exports = function draw(gd) {
...
             this.img = img;
 
-            // If not set, a `tainted canvas` error is thrown
-            img.setAttribute('crossOrigin', 'anonymous');
-            img.onerror = errorHandler;
-            img.onload = function() {
-                var canvas = document.createElement('canvas');
-                canvas.width = this.width;
-                canvas.height = this.height;
+            if(d.source && d.source.slice(0, 5) === 'data:') {
```
etpinard: `d.source` should always be a string at this stage, so you could use:

```js
if(d.source.indexOf('data:') === 0) {
  // ...
}
```

instead.

src/components/images/draw.js
```diff
@@ -84,31 +84,37 @@ module.exports = function draw(gd) {
...
+                img.src = d.source;
```
etpinard: I see that you're setting `img.src` here even though the `Image` instance isn't used in this case. That's not a bad idea as this makes the early return a few lines above

```js
        if(this.img && this.img.src === d.source) {
            return;
        }
```

work.

But setting `img.src` spins off the `img.onload`, `img.onerror`, etc event chain. I'm not sure how if that can lead to performance losses, but it would be easy to stash the `d.source` data in something like `img._src` instead to skip potential `img.onload` / `img.onerror` work.

etpinard: We could even go one step further by

- if `source` is a data uri, do not wrap logic in a `Promise` (and do not append `gd._promises`)
- if `source` is a data uri, do not instantiate `Image`
- use `thisImage.attr('xlink:href')` as the "stash" value instead of `img.src`

jonmmease: Yeah, that's a good point. `img._src` sounds good.

src/components/images/draw.js
jonmmease: 👍

etpinard:
Thanks @jonmmease - this works!

Before merging this, it would be cool to add one test checking that layout images with data uri as `source` do not call `canvas.getContext` and/or `canvas.toDataURL`. Here's one way we currently use to spy on canvas methods:

https://github.com/plotly/plotly.js/blob/ffa6e3468b592e45db2a00bb65805a92fd59c4fb/test/jasmine/tests/heatmap_test.js#L735-L741

etpinard:
Oh and @jonmmease , I put this under the `1.50.0` milestone. This _could_ be released as part of a patch if urgent, but I would prefer releasing this in a minor. I hope you don't mind.

jonmmease:
Thanks @etpinard,  yeah that test approach make sense. I'll give it a go!

Not urgent, 1.50.0 is fine. Thanks for the quick review!

src/components/images/draw.js
alexcjohnson: `d.source` can be large though, isn't `slice(0, 5)` going to be a lot faster than `indexOf`?

jonmmease: That's a good point @alexcjohnson. @etpinard, anything you don't like about `slice` here?

jonmmease:
Ok, in c310c92 I pulled the data uri image handling outside of the promise. So in this case we don't create the promise, construct an `Image`, or run `onload`.  And I added a `this._imgSrc` property to save off the input image string for later comparison.

In dd585b2, I added a test to make sure that we don't construct any `Image` elements for the data uri case, but we do construct an image and call `toDataURL` for the image url case.

**Note:** I tried to avoid the `setTimeout` call in the test, but I couldn't figure out how to get Jasmine to check that a variable remains undefined, without timing out the test. `expect(newCanvasElement).not.eventually.toBeDefined()`, `expect(newCanvasElement).eventually.not.toBeDefined()`, and `expect(newCanvasElement).eventually.toBeUndefined()` all timed out.

src/components/images/draw.js
etpinard: It's not obvious to me why `slice` would be faster than `indexOf() === 0`. Has anyone tested that out on jsperf before?

jonmmease: My thinking was that If the string starts with `'data:'`, then `indexOf()` would terminate right away. But if not, `indexOf()` would search the full string and not bail out until it reaches the end, whereas slice never needs to look past the first 5 characters.  If the string doesn't start with `data:`, then I guess it should be a URL which hopefully isn't terribly long. So it might not matter either way.

test/jasmine/tests/layout_images_test.js
```diff
@@ -290,6 +294,31 @@ describe('Layout images', function() {
...
+            spyOn(document, 'createElement').and.callFake(function(elementType) {
+                var element = originalCreateElement.call(document, elementType);
+                if(elementType === 'canvas') {
+                    newCanvasElement = element;
+                    spyOn(newCanvasElement, 'toDataURL');
+                }
+                return element;
+            });
+
+            Plotly.relayout(gd, 'images[0].source', dataUriImage).then(function() {
```
etpinard: Non-blocking but something like:

```js
        it('should only create canvas if url image', function(done) {
            var originalCreateElement = document.createElement;
            var newCanvasElement;
            spyOn(document, 'createElement').and.callFake(function(elementType) {
                var element = originalCreateElement.call(document, elementType);
                if(elementType === 'canvas') {
                    newCanvasElement = element;
                    spyOn(newCanvasElement, 'toDataURL');
                }
                return element;
            });

            Plotly.relayout(gd, 'images[0].source', dataUriImage)
            .then(function() {
                expect(newCanvasElement).toBeUndefined();
            })
            .then(function() { return Plotly.relayout(gd, 'images[0].source', pythonLogo); })
            .then(function() {
                expect(newCanvasElement).toBeDefined();
                expect(newCanvasElement.toDataURL).toHaveBeenCalledTimes(1);
            })
            .catch(failTest)
            .then(done)
        });
```

should work.

src/components/images/draw.js
alexcjohnson: I guess in as far as the two options we care about are a short string starting `http`
and a long string starting `data:`, `indexOf` should be fine. The case to worry about would be a long string that doesn't start `data:`... but we don't expect that and wouldn't care if it's slow.

jsperf shows a win for `indexOf` https://jsperf.com/slice-vs-indexof-acj but I don't trust it, there's no way it could actually do `indexOf` in 2 clock cycles, compiler must have turned it into a constant. This isn't happening in a loop so either option is fast enough to be not worth worrying about.

test/jasmine/tests/layout_images_test.js
jonmmease: The trouble there is that the first `expect(newCanvasElement).toBeUndefined();` will also pass for a url image, because the `onload` callback won't necessarily have completed yet. At least that's what I concluded was happening when I tried this.

etpinard: > The trouble there is that the first `expect(newCanvasElement).toBeUndefined();` will also pass for a url image, because the `onload` callback won't necessarily have completed yet.

Hmm, that shouldn't happen, the `Plotly.relayout` promise shouldn't resolve before the `onload` callback is executed.

jonmmease: Ok, I'll double check...

jonmmease: Yeah, you're right. Not sure what I was trying before.  I updated the test to use this structure.

etpinard:
Very strong PR @jonmmease ! Thanks very much for your help!

We haven't started merging things for `1.50.0`, so I won't give you a `: dancer :` just yet. We'll most likely release 1 or 2 more patch releases in the `1.49.x` series before.

jonmmease:
Thanks for the reviews @etpinard, nice to dip my toe into the code base again!

PR #4108, opened 2019-08-06
Refactoring color index

archmoj:
Most of our `component` indices have clear `module.exports` interface block.
This PR adds a similar one for `color` component i.e to increase the readability of the code base.

@plotly/plotly_js

PR #4109, opened 2019-08-06
refactor drag element and hover indices - add module.exports blocks

archmoj:
Similar to https://github.com/plotly/plotly.js/pull/4108 `dragelement` is the last `component` object where a module.exports block is added i.e. to avoid mutating the exports object.

Thanks @plotly/plotly_js for your time reviewing this one too.

### DIFF
--- a/src/components/dragelement/unhover.js
+++ b/src/components/dragelement/unhover.js
@@ -14,9 +14,12 @@ var getGraphDiv = require('../../lib/dom').getGraphDiv;
 
 var hoverConstants = require('../fx/constants');
 
-var unhover = module.exports = {};
+module.exports = {
+    wrapped: wrapped,
+    raw: raw
+};
 
-unhover.wrapped = function(gd, evt, subplot) {
+function wrapped(gd, evt, subplot) {
     gd = getGraphDiv(gd);
 
     // Important, clear any queued hovers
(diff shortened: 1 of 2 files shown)


src/components/dragelement/unhover.js
```diff
@@ -14,22 +14,25 @@ var getGraphDiv = require('../../lib/dom').getGraphDiv;
 
 var hoverConstants = require('../fx/constants');
 
-var unhover = module.exports = {};
+module.exports = {
```
etpinard: Wanna go one step further and always put the `module.exports` at the end of the file?

archmoj: I think that's OK to have that on top where one could easily view inputs and outputs of the module.

etpinard: Well, that's why exactly people don't like wasting time on syntax PRs like this one. If we can't add a test (e.g. an eslint rule), then how can we tell whether the "before" or the "after" is better?

archmoj: I don't think we should consider this as a syntax PR. I would say it is more related to the difference between using function declarations and function expressions at the run time.
Maybe @antoinerg could help investigate the benchmark diffs?
Plus in large files having a clear interface can help the developers have a quick view of the methods available.

etpinard: > I would say it is more related to the difference between using function declarations and function expressions at the run time.

> Maybe @antoinerg could help investigate the benchmark diffs?

Ok, well if you can detect a performance improvement, then YES I'm all for this PR. Before that happens, I have no interest in this PR. 

> Plus in large files having a clear interface can help the developers have a quick view of the methods available.

One could also argue: inputs (at the top) / new things (in the middle) / output (at the bottom) is easier to read. In addition, most files that have a `module.exports = { /* */ }` block currently, have it at the bottom of the file. But like I said before, these are simply opinion, so please let's not waste time arguing over this.

PR #4122, opened 2019-08-13
Add correct number formater

Domino987:
This adds the missing decimal and thousands separator for the italian language as also mentioned [here](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html).
@ghtmtt, @alexcjohnson, @archmoj, @etpinard

Thanks for your interest in plotly.js!


etpinard:
Thanks very much!

PR #4123, opened 2019-08-13
fix for invisible Sankey traces

### LINKED ISSUE #4119: Visibility property on multiple Sankey traces breaks plots
Same console error shown in [plotly/plotly.js#3798](https://github.com/plotly/plotly.js/issues/3798)

Was attempting to use the visibility flag dropdown shown [here](https://plot.ly/javascript/dropdowns/#add-two-dropdown-menus-to-a-chart-with-plotly.js). When I stumbled upon the above linked issue with the same error message that I was getting.

If both of the visible flags are true, it works fine. If multiple traces exist where at least one visible flag is false, the below error is output.

[JSFiddle](https://jsfiddle.net/9fmyo20c/) with reproduction of bug.

Error output in console matching linked issue:
```
Uncaught TypeError: Cannot read property 'groups' of undefined 
    at plot (plotly-latest.js:180349)
    at Object.exports.plot (plotly-latest.js:179746)
    at exports.drawData (plotly-latest.js:118147)
    at Object.lib.syncOrAsync (plotly-latest.js:107400)
    at Object.plot (plotly-latest.js:112750)
    at Object.newPlot (plotly-latest.js:113020)
    at window.onload ((index):66)
```

antoinerg:
Closes #4119 

Before: https://jsfiddle.net/9fmyo20c/
After: https://jsfiddle.net/qakmy2x6/

### DIFF
--- a/src/traces/sankey/plot.js
+++ b/src/traces/sankey/plot.js
@@ -129,6 +129,7 @@ module.exports = function plot(gd, calcData) {
 
     // stash initial view
     for(var i = 0; i < gd._fullData.length; i++) {
+        if(!gd._fullData[i].visible) continue;
         if(gd._fullData[i].type !== cn.sankey) continue;
         if(!gd._fullData[i]._viewInitial) {
             var node = gd._fullData[i].node;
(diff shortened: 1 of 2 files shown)


etpinard:
Thanks! :dancer:

PR #4125, opened 2019-08-13
scattergl: fill text arrays up to trace._length if necessary

### LINKED ISSUE #4124: scattergl: text arrays shorter than data arrays throws error
Plotting the following gives an error:
```javascript
Plotly.newPlot(gd,[{
        "type": "scattergl",
        "mode": "markers+text",
        "x": [3, 2, 1, 0],
        "y": [0, 1, 4, 9],
        "text": ["1", "2", "3"],
        "textposition": "top center"
    }])
```
Screenshot_2019-08-13_14-10-16

antoinerg:
Closes #4124

### DIFF
--- a/src/traces/scattergl/convert.js
+++ b/src/traces/scattergl/convert.js
@@ -112,6 +112,11 @@ function convertTextStyle(trace) {
     var i;
 
     optsOut.text = trace.text;
+    if(Array.isArray(optsOut.text)) {
+        for(i = optsOut.text.length; i < count; i++) {
+            optsOut.text[i] = '';
+        }
+    }
     optsOut.opacity = trace.opacity;
     optsOut.font = {};
     optsOut.align = [];
--- a/test/jasmine/tests/scattergl_test.js
+++ b/test/jasmine/tests/scattergl_test.js
@@ -105,6 +105,20 @@ describe('end-to-end scattergl tests', function() {
         }).catch(failTest).then(done);
     });
 
+    it('@gl should handle a plot with less text labels than data points', function(done) {
+        expect(function() {
+            Plotly.plot(gd, [{
+                'type': 'scattergl',
+                'mode': 'markers+text',
+                'x': [3, 2, 1, 0],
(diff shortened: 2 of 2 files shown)


etpinard:
Excellent :dancer:

src/traces/scattergl/convert.js
```diff
@@ -112,6 +112,11 @@ function convertTextStyle(trace) {
     var i;
 
     optsOut.text = trace.text;
+    if(Array.isArray(optsOut.text)) {
+        for(i = optsOut.text.length; i < count; i++) {
+            optsOut.text[i] = '';
+        }
+    }
```
alexcjohnson: ⚠️ this is mutating the original data array - I don't think we want to do that.

Screen Shot 2019-08-13 at 5 29 39 PM

antoinerg: Nice catch @alexcjohnson !

Because this file conflicts with my `texttemplate`, branch I will have to modify it anyway. How should I solve this problem? Can I simply `slice` it or is this a no go for performance?

alexcjohnson: I'd say slice is fine - as long as you only do it when `text.length < count`. I think it's totally reasonable to say "for best performance make sure your arrays are all the same length" and prioritize stability over edge-case performance. The presumably higher-perf alternative would be to track down where the array is being indexed and fall back to a blank string at that point... but looks like that's deep in a regl submodule so that sounds painful 😅

etpinard: Probably not a big deal. But since we're planning on releasing 1.49.3 and this PR got merged, @antoinerg can you make another PR off master to fix and test the `optsOut.text` mutation?

etpinard: .... instead of fixing it in https://github.com/plotly/plotly.js/pull/4071

antoinerg: @etpinard Done in https://github.com/plotly/plotly.js/pull/4126

PR #4126, opened 2019-08-14
scattergl: do not mutate text array

antoinerg:
Fix for PR #4071

See https://github.com/plotly/plotly.js/pull/4125#discussion_r313621904 for details


etpinard:
Thanks :dancer:

Issue #4136, opened 2019-08-19
mirror: "all" - all axis lines and ticks are mirrored


DaanyWaay:
Ticks should be mirrored only if "allticks" is selected.
https://codepen.io/pen/ZEzpYWY

etpinard:
essential the bug part of https://github.com/plotly/plotly.js/issues/2539 - thanks very much for posting!

thchr:
I noted this over in #4140 but realized it might make more sense to note it in this issue instead:

> This seems to still be be broken on v2.2.1: https://codepen.io/thchr/pen/eYWGjQM
> I.e., the tick marks are on both y-axes of the first subplot.

PR #4140, opened 2019-08-20
Don't draw ticks on tick subplots when mirror:'all'

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4136

before: https://codepen.io/DaanyWaay/pen/ZEzpYWY
after: https://codepen.io/etpinard/pen/zYOoYJV

and see the `mirror-all-vs-allticks` image diff in https://github.com/plotly/plotly.js/commit/cc191783c1a111031ed7f14ebc0f2c2d7355e284

cc @plotly/plotly_js


archmoj:
@etpinard thanks very much for the fix & test.
:dancer:

thchr:
This seems to still be be broken on v2.2.1: https://codepen.io/thchr/pen/eYWGjQM ?
I.e., the tick marks are on both y-axes of the first subplot.

Issue #4142, opened 2019-08-21
Missing translation for "Download plot" in localization files


Batalex:
Hi,
I noticed that the localization files do not have the translation key "Download plot" (they have the legacy "Download plot as png"). I checked three of them (**fr**, **de**, & **es**), but maybe some of them do have this key.

I do not know how they were generated, but  if needed I would be glad to contribute on the french ones.

etpinard:
This one got fixed in https://github.com/plotly/plotly.js/pull/4148

PR #4148, opened 2019-08-26
add missing translation keys

Batalex:
See https://github.com/plotly/plotly.js/issues/4142

This PR adds the missing "Download plot" localization key for the following languages:
- cs
- de
- es
- fr
- it
- ja
- pt-br
- sw

The others languages either have already the missing key or do not translate modebar buttons.
I am by no mean a translator (Deepl/Translate for the win). Native speakers feedbacks are welcome even if it seemed "simple".

### DIFF
--- a/lib/locales/cs.js
+++ b/lib/locales/cs.js
@@ -26,6 +26,7 @@ module.exports = {
         'Double-click on legend to isolate one trace': 'Dvojklikem na legendu izolujete jedinou datovou sadu',  // components/legend/handle_click.js:90
         'Double-click to zoom back out': 'Dvojklikem vrátíte zvětšení',                                  // plots/cartesian/dragbox.js:299
         'Download plot as a png': 'Uložit jsko PNG',                                                       // components/modebar/buttons.js:52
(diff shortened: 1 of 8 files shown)


Domino987:
@etpinard it has my ok for de

ghtmtt:
@etpinard +1 for Italian too

fhurta:
@etpinard cs translation is fine

Batalex:
Thanks everyone for your time! 

@Domino987 My german is awful, but how do you feel about the "Download plot as a png" just above? Would "Graphen als PNG herunterladen" be better as well?

Domino987:
@Batalex yes that would be better as well.

jins-tkomoda:
@etpinard ja translation is also fine!

n-riesco:
@etpinard `lib/locales/es.js` LGTM

etpinard:
Thanks everyone!

We're only missing a :+1: from a `sw` speaker

etpinard:
Ok, we're still missing a :+1: from a `sw` speaker, but since the translation appears straightforward

image

I'm ok trusting Deepl here. Merging. Set for release in 1.50.0

Issue #4132, opened 2019-08-16
Ability to constrain legend background to the legend text


chriddyp:
In this example, the legend's background color spans the entire width of the graph. It'd be nice if it only spanned the legend text, so that the lines on the right of the graph aren't beneath the semi-transparent white legend background.

![image](https://user-images.githubusercontent.com/1280389/63194794-f0162800-c03e-11e9-966e-e60795aaad46.png)

etpinard:
Fixed in https://github.com/plotly/plotly.js/compare/drop-firstRender-in-legend-draw

will be part of 1.50.0

chriddyp:
nice!

etpinard:
Actually, I can't even reproduce this bug on latest, let alone on my dev branch.

My attempt: https://codepen.io/etpinard/pen/OJLXRJa?editors=1010

Which font are you using?

etpinard:
https://codepen.io/etpinard/pen/OJLXRJa?editors=1010 now has a minimal reproducible example.

The problem happens with horizontal legends when `legendgroup` is set.

jensb:
How do you get the semi-transparent legend background in your screenshot above?
It's possible to set bgcolor=rgba(x,y,z,128) for example but it's ignored. Only 0 (transparent) or >0 (non-transparent) seem to have any effect. 
Thanks!

Issue #771, opened 2016-07-24
Legend overlaps since 1.11.0


lp-jonathan:
This appears to be a regression introduced in 1.11.0 - probably by: https://github.com/plotly/plotly.js/issues/53 and is still evident in "latest;

In 1.10.2 and earlier, when a legend was too wide and would overlap a pie chart, the legend would not show.
Since 1.11.0, the legend is shown overlapping the pie chart.

This can be replicated very easily with the following data declaration (default everything else):

```
var data = [{
    labels:['hello','this is a very long string and bad things are likely to happen', 'short string','moderately long string'],
    values:[4,4,4,4],
    type:'pie'
}];
```

Of these two behaviours, not showing the legend is preferred. The ideal solution (feature request I guess) would be to scale the legend to the a suitable non-overlapping size, then wrap the text within that.

etpinard:
Reproducible example here: http://codepen.io/etpinard/pen/dXZQGP

Is this really a bug though? To me the behaviour in `latest` makes more sense - where by default (when no `layout.legend` options have been set) the legend items are squeezed into the graph.

That said, something **is** up with legend positioning. Take a look at http://codepen.io/etpinard/pen/mELQPA - even by setting `legend.x` to something `>  1`, the legend refuses to be placed on the right of the graph.

I doubt that affect only pie charts. The legend drawing code is pretty trace type agnostic. To be investigated.

lp-jonathan:
Thanks for creating the reproduction.

Curious - in my application when I had a too-long legend, it simply 
didn't show the legend, I assumed that was a design feature. But your 
first 1.10 example shows the legend but mostly obscured. In that 
scenario neither is ideal. The second example is the best of the three.

My suggestion would be to enable word-wrap and allow the developer to 
specify the maximum width of the legend.

n-riesco:
The logic to handle oversized legends is [here](https://github.com/psalmody/plotly.js/blob/5b5a102874974be5d726bc7ef6edf54c16746de7/src/components/legend/draw.js#L165-L193).

etpinard:
Another report in https://github.com/plotly/plotly.js/issues/840

------------

From @odow 

> Narrow screens cause horizontal legends with lots of entries to bug.

> https://codepen.io/odow/pen/mEarpy

> In this example, the three column layout is fine, but when the width forces a two column layout, the legend is no longer below the plot area but on top of, and scrolling is enabled.

---------

From @n-riesco 

[...]

PR #4160, opened 2019-08-30
Multiple legend positioning and sizing fixes

etpinard:
This PR essentially supersedes https://github.com/plotly/plotly.js/pull/3024, fixes https://github.com/plotly/plotly.js/issues/4132 and will resolve https://github.com/plotly/plotly.js/issues/771

In brief,

- commit https://github.com/plotly/plotly.js/commit/de369f38bcd4eaf019345cd844d73e4cdba03030 improves legend `x` / `y` attribute declarations
- commits https://github.com/plotly/plotly.js/commit/1cf347a379f200f495012831539fa117e21c16a5 and https://github.com/plotly/plotly.js/commit/f12c62c27c65156d836cbdbadf28cb543974ca14 remove obsolete logic from the `legend/draw.js`
- mocks of the problems listed in https://github.com/plotly/plotly.js/issues/771 (with baselines as of 1.49.4) are added in https://github.com/plotly/plotly.js/commit/bf586d4a74c5f5ce9b5cf77e8fb5481ae1e00899 (some of the mocks were taken from https://github.com/plotly/plotly.js/pull/3024 - thanks @antoinerg !)
- https://github.com/plotly/plotly.js/commit/488cca19cc30022c025cfdfc7f499abf66ea7565 is the most important commit, it fixes most of the issues presented in https://github.com/plotly/plotly.js/issues/771
- https://github.com/plotly/plotly.js/commit/c33c92fb4c90e6f79a2dca7ac706405b52f9640d adds extra logic around `margin.autoexpand: false` in order to present a solution to a #771 scenario
- commits https://github.com/plotly/plotly.js/commit/1e6862a6ce2d28b5e87bb608013d3a66c45f18d4 and https://github.com/plotly/plotly.js/commit/1f00a40ef07aac47e8ae7d068182ec165b767baf add some `Lib.log` calls that should make legend auto-margin / positioning edge cases easier to spot
- commits https://github.com/plotly/plotly.js/commit/2b2bd9b64678e157bba661a4f721891d63e59600 and https://github.com/plotly/plotly.js/commit/6b0befeca81dd6791d4baa094578a58dd8bef504 present and fix https://github.com/plotly/plotly.js/issues/4132
- the rest are essentially linting commits

Reviewers may be interested in reviewing 

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/components/legend/draw.js#L98-L118

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/components/legend/draw.js#L462-L627

in `legend/draw.js` with their corresponding blocks on the current master:

https://github.com/plotly/plotly.js/blob/37e80cbc10ef8bf93b4dee1c1e7de043745da676/src/components/legend/draw.js#L133-L195

https://github.com/plotly/plotly.js/blob/37e80cbc10ef8bf93b4dee1c1e7de043745da676/src/components/legend/draw.js#L541-L724

cc @plotly/plotly_js


nicolaskruchten:
Is this on the way to resolving https://github.com/plotly/plotly.js/issues/1199 or unrelated?

etpinard:
@antoinerg would you mind reviewing this one?

antoinerg:
Thanks for the nice PR and the clear breakdown of your different commits :+1: 

Here's my first comment/question following Alex's suggestion on #3024 to:
> Can you check what happens with config: {editable: true} when you drag the legend around and bump up against these various limits? Does the legend end up where you expected after you drop it?

@etpinard is this the expected behavior:

Peek 2019-09-03 12-26

etpinard:
Ok, here's what happens when you drag the legend to the right off the `legend_negative_x` mock under `editable:true`:

- `relayout` with a new `legend.x` (and `legend.y`) is called on mouse up
- the new automargin push values are set to `0` here:

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/plots/plots.js#L1824

as the legend width is larger than half the layout width.

- when the new `legend.x` is less than 1, this results in no right-margin push.
- as we can't fit the legend within the graph's width with the new  `legend.x`, we constrain its position here:

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/components/legend/draw.js#L108

snapping the legend back to the left.

-----------

Now here's what happens when one drags the legend to the right past the plot area:

Peek 2019-09-03 14-46

here, the max horizontal legend width is computed with:

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/components/legend/draw.js#L528-L531

is `fullLayout.width / 2`, and the right-margin gets pushed.

---------
----------

It's not obvious to me how we should _fix_ this problem. I'm thinking we could refine 

https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/plots/plots.js#L1824

by considering the push-margin `x` (and `y`) position before "dropping" the push-margin values. In https://github.com/plotly/plotly.js/pull/3024, looks like @antoinerg tried to scale down the push-margin values instead of dropping them. If someone has a better idea, please let me know.

antoinerg:
> * as we can't fit the legend within the graph's width with the new  `legend.x`, we constrain its position here:
> 
> https://github.com/plotly/plotly.js/blob/6b0befeca81dd6791d4baa094578a58dd8bef504/src/components/legend/draw.js#L108
> 
> snapping the legend back to the left.

@etpinard It seems however that it does not snap back to the left enough: the item for trace 0 is clipped (the blue trace).

etpinard:
> It seems however that it does not snap back to the left enough: the item for trace 0 is clipped (the blue trace).

Hmm. I can't replicate that. Can you check the `fullLayout.legend.x` value for which this happens?

antoinerg:
> Hmm. I can't replicate that. Can you check the `fullLayout.legend.x` value for which this happens? That would help me out a lot!

The value for `gd._fullLayout.legend.x` will depend on how far I drop the legend to the right. On Chrome and FF it reproducibly position the legend such that `trace 0` is outside the figure and clipped. :thinking:

etpinard:
> Is this on the way to resolving #1199 or unrelated?

This PR fixes a bunch of positioning bugs for horizontal legends, so it should make it easier for users to workaround the problems of #1199. But this PR does **not** make an attempt at fixing axis label <--> legend overlaps.

etpinard:
@antoinerg thanks very much for spotting that bug!! 

There was a race condition in `Legend.draw` - commit https://github.com/plotly/plotly.js/pull/4160/commits/42d5e4e0e5f9012a141cf91bcdf683436842d890 fixes it.

Dragging around the `legend_negative_x` legend in `editable:true` mode now does this:

Peek 2019-09-10 15-17

which appears "ok" to me. There are still some less-than-ideal scenarios. These should be much easier to solve once https://github.com/plotly/plotly.js/issues/2704 is completed. Let me know what you think!

antoinerg:
This is a massive improvement for legend positioning. It seems like the several issues grouped in #771 are fixed as well as #4132 :tada: 

> There are still some less-than-ideal scenarios.

I think the most confusing scenario for users is when they set `legend.x` to a value between 0 and 1. In this case, we don't expand the right margin and force the legend's position (as shown in your GIF). If the user really wants the legend to be at the provided `legend.x`, he/she will need to manually increase the right margin. It's probably OK for now but maybe we should improve on this once #2704 is completed.

One last non-blocking comment: I noticed that `layout.margin.autoexpand` doesn't have a description. It might be a good idea to add one either here or in #2704 .

Great job :muscle: @etpinard 

:dancer: :dancer: :dancer:

etpinard:
Thanks very much for the review @antoinerg 

I pushed https://github.com/plotly/plotly.js/pull/4160/commits/d0431514fbc55e56a35417f65fbbd01ca2055fab adding a description for `autoexpand`

Merging.

destradafilm:
> Has the suggested "fitinside" option ever been added? I've got an issue with legends crawling up to overlap the chart after resizing the window, and nothing in any of these commits or issues fixes it - but your codepen demoing "fitinside" on issue #771 had the correct behavior.

I agree,  I still see the problem of either choosing the legend overlapping the chart or legend with cut off text when legend labels are long. Would it be possible to have an option to wrap text in the legend or overflow auto with a max-width and a scroll bar?

PR #4167, opened 2019-09-05
Fix multi-axis transition "leaks"

### LINKED ISSUE #4153: Axis animation with subplots leaks to other axis
Every button here should only animate a single axis:
https://jsfiddle.net/9onwtvgu/

But clicking the first button also animates the second and third axis and clicking the second button also animates the third axis.

The problem seems to be that `editX` and `editY` are not initialised to `null` here:
https://github.com/plotly/plotly.js/blob/fbbcd20f65d1b61373d6625973a8885382816e82/src/plots/plots.js#L2424-L2436

The same problem might be found here as well (but not related to animations, I assume):
https://github.com/plotly/plotly.js/blob/fbbcd20f65d1b61373d6625973a8885382816e82/src/plots/plots.js#L2521-L2533

etpinard:
fixes #4153 - by resetting the loop variables in old/new transition prepare loops

before: https://jsfiddle.net/9onwtvgu/
after: https://jsfiddle.net/seh2Lx1n/

cc @plotly/plotly_js

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2407,26 +2407,25 @@ plots.transition = function(gd, data, layout, traces, frameOpts, transitionOpts)
                 var xr0 = xa.range.slice();
                 var yr0 = ya.range.slice();
 
-                var xr1;
+                var xr1 = null;
+                var yr1 = null;
+                var editX = null;
+                var editY = null;
+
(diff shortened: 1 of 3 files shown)


archmoj:
@etpinard Great fix.
:dancer:

Issue #3896, opened 2019-05-23
Bar length transitions


nicolaskruchten:
Subset of  #3456 for roadmap purposes...

Bar length should support transitions either via frame-animation or react() with transitions. This includes stacked bars which should transition without gaps or overlaps.

etpinard:
Duplicate of https://github.com/plotly/plotly.js/issues/1019

etpinard:
From Ricky's https://github.com/plotly/plotly.js/pull/1143#issuecomment-278722499

> I'm hoping will get a chance to loop back to it in the next week or so. Getting it started was pretty easy, but I don't yet know all the ins and outs of bar graphs and need maybe a day or two to lock down the remaining details. barmode: 'relative' is currently the one that was causing problems, though I suspect there are a decent number of details I still need to address.

and his TODO list in https://github.com/plotly/plotly.js/pull/1647#issue-118903659 had

- fix barmode: relative
- test
- object constancy for bar text

-----------

So, I sounds getting bars under `barmode: 'relative'` will be trickiest part to :white_check_mark:

nicolaskruchten:
Heh yeah. And we changed the behaviour of group barmode since then also to look more like relative. And we added connectors ;)

Issue #3931, opened 2019-06-03
Leftover traces in animations


nicolaskruchten:
In the following pen https://codepen.io/nicolaskruchten/pen/KLEprV?editors=0010 the trace in `data` has no `uid` but in `frames` there are `uids` ... hitting 'play' results in the original trace still being visible but not hoverable. I would expect that `animate()` would clean up the original rendering there?

etpinard:
We don't animate trace enter/exit at the moment cc https://github.com/plotly/plotly.js/issues/932

So transitions with `redraw: false` won't work as expected. Commenting the `redraw: false` lines in your codepen make things work ok https://codepen.io/etpinard/pen/RmOXPb?editors=0010

More in https://github.com/plotly/plotly.js/issues/1688

nicolaskruchten:
so when you say "animate trace exit" you mean "we leave them there on purpose even if they're no longer present in the active frame"? That's what I'm seeing right now, and the markers that are left aren't hoverable...

nicolaskruchten:
Setting redraw to true does fix this, but is very janky, as the animation appears to 'stall' at every frame 
:(

etpinard:
removing the `isFullReplot` condition around:

https://github.com/plotly/plotly.js/blob/d75e956db2fffe23bcf818c3ca9eb8952f68eab2/src/traces/scatter/plot.js#L80-L82

seems to solve the problem and doesn't make any test fail.

I can't remember what Ricky envisioned with that `isFullReplot` condition, but we should probably manually test out a few things before making a PR.

etpinard:
cc @antoinerg who may have copied that 

```js
 if(isFullReplot) { 
     join.exit().remove(); 
 } 
```

block in `bar/plot.js` while working on bar transitions.

etpinard:
Here's the `bar` version of @nicolaskruchten 's codepen (https://codepen.io/nicolaskruchten/pen/KLEprV i.e. with `redraw:false`):

https://codepen.io/etpinard/pen/MWgMEmx?editors=0010

![Peek 2019-09-30 18-01](https://user-images.githubusercontent.com/6675409/65920212-72b73480-e3ac-11e9-848d-b10f9b60d13c.gif)

after @antoinerg's bar transition work in #4180 and #4186.

-------

[...]

PR #4180, opened 2019-09-10
bar transitions

antoinerg:
Introduces transitions for `bar` and closes https://github.com/plotly/plotly.js/issues/3896 . The following can now be tweened:
- [x] length
- [x] width
- [x] text translates, rotates and scales properly
- [x] fill color
- [x] error bars
- [x] border width and color

[Codepen showcasing different mocks](https://codepen.io/antoinerg/pen/rNBygGO)
[Codepen showcasing that all `barmode` work properly](https://codepen.io/antoinerg/pen/KKPNGyz?editors=1000)

### checkTransition

Another notable addition in this PR is `checkTransition` which facilitate the testing of transitions. By hijacking `Date.now()`, it is possible to move and freeze time in order to inspect the resulting SVG at any moment. This way, we can guarantee that tweening works on the intended elements. See https://github.com/plotly/plotly.js/commit/8166eb70757616a204b83d292b37c45152f170c9 for details.


src/traces/bar/plot.js
```diff
@@ -51,7 +51,28 @@ function getXY(di, xa, ya, isHorizontal) {
     return isHorizontal ? [s, p] : [p, s];
 }
 
-function plot(gd, plotinfo, cdModule, traceLayer, opts) {
+function transition(selection, opts, makeOnCompleteCallback) {
```
etpinard: Clean wrapper!

It's a little different than what `scatter/plot.js` does, in particular this piece:

https://github.com/plotly/plotly.js/blob/46866fdbb1fdf2e12437a48252f190d07807aa00/src/traces/scatter/plot.js#L67-L73

@antoinerg have you noticed "out of order" transitions?

Your wrapper plays better with [`Lib.makeTraceGroups`](https://github.com/plotly/plotly.js/blob/master/src/lib/make_trace_groups.js), so if there are no side-effects, maybe we could try this pattern in `scatter/plot.js` and maybe fix https://github.com/plotly/plotly.js/issues/3931 while at it.

etpinard: One drawback from your pattern:

- we have to pass `opts` and `makeOnCompleteCallback` around in all the `transition` calls

etpinard: The `scatter/plot.js` uses this trick:

https://github.com/plotly/plotly.js/blob/46866fdbb1fdf2e12437a48252f190d07807aa00/src/traces/scatter/plot.js#L126-L128

src/traces/bar/plot.js
```diff
@@ -178,12 +201,19 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts) {
                 y1 = fixpx(y1, y0);
             }
 
-            Lib.ensureSingle(bar, 'path')
+            var sel = transition(Lib.ensureSingle(bar, 'path'), opts, makeOnCompleteCallback);
+            sel
                 .style('vector-effect', 'non-scaling-stroke')
-                .attr('d', isBlank ? 'M0,0Z' : 'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
+                .attr('display', function() { return isBlank ? 'none' : null;})
```
etpinard: Interesting move.

Do you have an example of a bar transitioning in and out of `isBlank`?

antoinerg: I should mention here that this was necessary otherwise bars transitioning from `isBlank` to a given value would start growing from the top left corner of the figure (starting at 0, 0)! We need to have blank bars at the right place for transitions in and out of `isBlank` to work. However, even though we have them in the SVG, they shouldn't appear (or we break several mocks) which is why we set them to `display: none`.

> Do you have an example of a bar transitioning in and out of isBlank?

Let me produce a Codepen for you

antoinerg: Here's the requested Codepen: https://codepen.io/antoinerg/pen/PoYRBzw

As you can see, there is a problem going from a finite value to blank. The bar disappears right away! Will need to fix this. Very nice catch :male_detective: @etpinard

test/jasmine/assets/check_transitions.js
```diff
@@ -0,0 +1,115 @@
...
+                clockTick(startTime, test[0]);
+                currentTime = test[0];
+            }
+            return assert(test);
+        };
+    });
+
+    // Run all tasks
+    return promiseSerial(p.concat(checkTests))
+        .then(function() {
```
etpinard: What if something goes wrong during the promise chain? Does `Date.now` return to its original self after?

test/jasmine/assets/check_transitions.js
```diff
@@ -0,0 +1,115 @@
...
+
+// A test array is made of
+// [ms since start of transition, selector, (attr|style), name of attribute, array of values to be found]
+// Ex.: [0, '.point path', 'style', 'fill', ['rgb(31, 119, 180)', 'rgb(31, 119, 180)', 'rgb(31, 119, 180)']]
+function assert(test) {
+    var msg = 'at ' + test[0] + 'ms, selection ' + test[1] + ' has ' + test[3];
+    var cur = [];
+    Plotly.d3.selectAll(test[1]).each(function(d, i) {
+        if(test[2] === 'style') cur[i] = this.style[test[3]];
+        if(test[2] === 'attr') cur[i] = Plotly.d3.select(this).attr(test[3]);
```
etpinard: :cow: you could just use `d3.select` instead of `Plotly.d3.select`

test/jasmine/assets/check_transitions.js
```diff
@@ -0,0 +1,115 @@
...
+    return Promise.resolve(true);
+}
+
+function assertAttr(A, B, cb, msg) {
+    var a = cb(A);
+    var b = cb(B);
+    expect(a).withContext(msg + ' equal to ' + JSON.stringify(b)).toEqual(b);
+}
+
+function round(str) {
```
etpinard: I guess this is fine.

I would have prefered extracting the numbers out of the transform strings and then use [`.toBeCloseArray`](https://github.com/plotly/plotly.js/blob/46866fdbb1fdf2e12437a48252f190d07807aa00/test/jasmine/assets/custom_matchers.js#L58-L73), but this wrapper is a huge :trophy:

test/jasmine/tests/bar_test.js
```diff
@@ -2548,3 +2549,96 @@ function assertTraceField(calcData, prop, expectation) {
...
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+
+    afterEach(function() {
+        Plotly.purge(gd);
+        destroyGraphDiv();
+    });
+
+    it('for bar fill color', function(done) {
```
etpinard: These tests are :100: 

From your

- [x] length
- [x] width
- [x] text translates, rotates and scales properly
- [x] fill color
- [x] error bars
- [x] border width and color

checklist, looks like we're only missing :lock:s for `marker.line.width` and `marker.line.color`

etpinard: and could you add one test for horizontal bars to test this clause:

https://github.com/plotly/plotly.js/blob/bda810fd90be1ccb6993ec856d8ff59f007aa772/src/traces/bar/plot.js#L141

test/jasmine/tests/transition_test.js
```diff
@@ -444,8 +444,8 @@ describe('Plotly.react transitions:', function() {
         var layout = {transition: {duration: 10}};
 
         // sanity check that this test actually tests what was intended
-        var Bar = Registry.modules.bar._module;
-        if(Bar.animatable || Bar.attributes.marker.line.width.anim !== true) {
+        var Violin = Registry.modules.violin._module;
+        if(Violin.animatable || Violin.attributes.marker.line.width.anim !== true) {
```
etpinard: Very nice touch here!

test/image/mocks/animation_bar.json
```diff
@@ -0,0 +1,22 @@
...
+    "x": ["A", "B", "C"],
+    "y": [24, 5, 8],
+    "error_y": {"array": [3, 2, 1]}
+    }],
+  "layout": {
+    "width": 400,
+    "height": 400,
+    "yaxis": {"range": [0, 30]}
+  },
+  "frames": [
```
etpinard: For reference, calling `Plotly.animate(gd)` off this mock is very fun:


Peek 2019-09-10 11-50

etpinard:
Awesome PR @antoinerg !!

src/traces/bar/plot.js
etpinard: **This comment is non-blocking**

src/traces/bar/plot.js
```diff
@@ -325,9 +354,12 @@ function appendBarText(gd, plotinfo, bar, calcTrace, i, x0, x1, y0, y1, opts) {
                 (textPosition === 'outside') ?
                 outsideTextFont : insideTextFont);
 
+        var currentTransform = textSelection.attr('transform');
+        textSelection.attr('transform', '');
         textBB = Drawing.bBox(textSelection.node()),
```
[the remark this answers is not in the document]
etpinard: **This comment is non-blocking**

test/jasmine/assets/check_transitions.js
etpinard: **This comment is non-blocking**

antoinerg: > I would have prefered extracting the numbers out of the transform strings and then use [`.toBeCloseArray`](https://github.com/plotly/plotly.js/blob/46866fdbb1fdf2e12437a48252f190d07807aa00/test/jasmine/assets/custom_matchers.js#L58-L73), but this wrapper is a huge

This might indeed be necessary because of the discrepancy in the size of text elements in different browsers/OS which ends affecting the exact transform strings.

src/traces/bar/plot.js
antoinerg: Fixed in https://github.com/plotly/plotly.js/pull/4180/commits/495d036d44636923d9e6d55d7261629c8c7a81ac

test/jasmine/assets/check_transitions.js
antoinerg: Done in c871cdb5292bf7c3e75431d3936c2229966cb37

test/jasmine/assets/check_transitions.js
antoinerg: In c871cdb5292bf7c3e75431d3936c2229966cb37, I made sure it is restored by adding a `catch` block

test/jasmine/tests/bar_test.js
antoinerg: > checklist, looks like we're only missing s for `marker.line.width` and `marker.line.color`

Added in https://github.com/plotly/plotly.js/pull/4180/commits/cc871cdb5292bf7c3e75431d3936c2229966cb37

test/jasmine/assets/check_transitions.js
etpinard: Hmm, looks like there are indeed cross-browser differences.

Running `npm run test-jasmine -- bar` on Chrome 70 in Ubuntu 18.04, the two cases with on-graph bar text fail for me, see details below:

<details>
Chrome 70.0.3538 (Linux 0.0.0) bar tweening for vertical bar height and text position FAILED
	Error: at 0ms, selection text.bartext has transform equal to ["translate(90 58)","translate(269 239)","translate(450 210)"]: Expected $[0] = 'translate(90 58)' to equal 'translate(90 56)'.
	Expected $[1] = 'translate(269 239)' to equal 'translate(270 237)'.
	Expected $[2] = 'translate(450 210)' to equal 'translate(450 208)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:48:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210857:43
	Error: at 300ms, selection text.bartext has transform equal to ["translate(90,166)","translate(269,130)","translate(450,187)"]: Expected $[0] = 'translate(90,166)' to equal 'translate(90,164)'.
	Expected $[1] = 'translate(269,130)' to equal 'translate(270,128)'.
	Expected $[2] = 'translate(450,187)' to equal 'translate(450,185)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210878:20
	Error: at 600ms, selection text.bartext has transform equal to ["translate(90 239)","translate(269 58)","translate(450 172)"]: Expected $[0] = 'translate(90 239)' to equal 'translate(90 237)'.
	Expected $[1] = 'translate(269 58)' to equal 'translate(270 56)'.
	Expected $[2] = 'translate(450 172)' to equal 'translate(450 170)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210878:20
Chrome 70.0.3538 (Linux 0.0.0) bar tweening for horizontal bar length and text position FAILED
	Error: at 0ms, selection text.bartext has transform equal to ["translate(99 230)","translate(505 140)","translate(248 50)"]: Expected $[0] = 'translate(99 230)' to equal 'translate(100 229)'.
	Expected $[1] = 'translate(505 140)' to equal 'translate(506 139)'.
	Expected $[2] = 'translate(248 50)' to equal 'translate(249 49)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:48:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210857:43
	Error: at 300ms, selection text.bartext has transform equal to ["translate(227,230)","translate(389,140)","translate(287,50)"]: Expected $[0] = 'translate(227,230)' to equal 'translate(228,229)'.
	Expected $[1] = 'translate(389,140)' to equal 'translate(391,139)'.
	Expected $[2] = 'translate(287,50)' to equal 'translate(287,49)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210878:20
	Error: at 600ms, selection text.bartext has transform equal to ["translate(312 230)","translate(313 140)","translate(312 50)"]: Expected $[0] = 'translate(312 230)' to equal 'translate(314 229)'.
	Expected $[1] = 'translate(313 140)' to equal 'translate(314 139)'.
	Expected $[2] = 'translate(312 50)' to equal 'translate(313 49)'.
	    at <Jasmine>
	    at assertAttr (/tmp/test/jasmine/assets/check_transitions.js:109:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210918:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:98:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210907:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/1ea018d52a3b10a9a939f0f793fee4b4.browserify.js:210878:20
Chrome 70.0.3538 (Linux 0.0.0): Executed 112 of 112 (2 FAILED) (4.036 secs / 3.5 secs)
</details>

src/traces/bar/plot.js
etpinard: Oh I see what your doing. Setting a blank path `M0,0Z` breaks the transition, but setting a `y0 = y1` (or `x0 = x1`) rectangle would end up showing the `marker.line` around the bar (which we don't want to do for blank bars)

That `display` trick is clever, but unfortunately it doesn't work under the `frame: {redraw: false}` option: https://codepen.io/etpinard/pen/bGbvZxq - see the `marker.line` is there after the transition is completed.

etpinard: For reference, using the current 

```js
sel.attr('d', isBlank ? 'M0,0Z' : 'M' + x0 + ',' + y0 + 'V' + y1 + 'H' + x1 + 'V' + y0 + 'Z')
```

we get:

Peek 2019-09-10 15-51

test/jasmine/assets/check_transitions.js
antoinerg: I introduced tolerance for `transform` checks in https://github.com/plotly/plotly.js/pull/4180/commits/86ba54ad454c804c29298017b7a497df0829c828

src/traces/bar/plot.js
antoinerg: > That `display` trick is clever, but unfortunately it doesn't work under the `frame: {redraw: false}` option: https://codepen.io/etpinard/pen/bGbvZxq - see the `marker.line` is there after the transition is completed.

I was unaware of that `redraw` thingy. It didn't actually make sense to me why https://github.com/plotly/plotly.js/commit/495d036d44636923d9e6d55d7261629c8c7a81ac would work :laughing: 

I think the solution will be to set `display: none` on transition end and interupt.

etpinard: The solution might be to make

https://github.com/plotly/plotly.js/blob/9b5f0898e13c24f2ad23c9745cac07e48c0e35d9/src/components/drawing/index.js#L445

```js
sel.style('stroke-width', (d.isBlank ? 0 : lineWidth) + 'px');
```

while keeping your new:

https://github.com/plotly/plotly.js/blob/cc871cdb5292bf7c3e75431d3936c2229966cb37/src/traces/bar/plot.js#L141-L142

and dropping the `attr('display', /**/)` lines

test/jasmine/assets/check_transitions.js
etpinard: One of those tests is still failing for me:

<details>
Chrome 70.0.3538 (Linux 0.0.0) bar tweening for vertical bar height and text position FAILED
	Error: at 0ms, selection text.bartext has transform equal to ["translate(89.796875 58)","translate(269.375 238.5)","translate(449.78125 210)"]: Expected 89.796875,58,269.375,238.5,449.78125,210 to be close to 90,56,270,236.5,450,208 
	    at <Jasmine>
	    at assertCloseTo (/tmp/test/jasmine/assets/check_transitions.js:117:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210926:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:99:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210908:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:48:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210857:43
	Error: at 300ms, selection text.bartext has transform equal to ["translate(89.796875,166.3)","translate(269.375,130.2)","translate(449.78125,187.2)"]: Expected 89.796875,166.3,269.375,130.2,449.78125,187.2 to be close to 90,164.3,270,128.20000000000002,450,185.2 
	    at <Jasmine>
	    at assertCloseTo (/tmp/test/jasmine/assets/check_transitions.js:117:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210926:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:99:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210908:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210878:20
	Error: at 600ms, selection text.bartext has transform equal to ["translate(89.796875 238.5)","translate(269.375 58)","translate(449.78125 172)"]: Expected 89.796875,238.5,269.375,58,449.78125,172 to be close to 90,236.5,270,56,450,170 
	    at <Jasmine>
	    at assertCloseTo (/tmp/test/jasmine/assets/check_transitions.js:117:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210926:67)
	    at assert (/tmp/test/jasmine/assets/check_transitions.js:99:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210908:13)
	    at /tmp/test/jasmine/assets/check_transitions.js:69:1 <- /tmp/dbdeaeadcc8a079807751cc57fad1a9c.browserify.js:210878:20
</details>

----

Bumping the tolerance to `2.5` (from `2`) make the tests pass.

Some info for future PRs: in my experience precision-based tolerances (e.g. `- Math.log10(2 * precision)` doesn't work so well for cross-browser differences in text sizing/positioning (like here). Absolute-value-based tolerances (e.g. using `toBeWithin` or `toBeWithinArray`) appears to work better.

src/traces/bar/plot.js
antoinerg: > The solution might be to make
> 
> https://github.com/plotly/plotly.js/blob/9b5f0898e13c24f2ad23c9745cac07e48c0e35d9/src/components/drawing/index.js#L445
> 
> ```js
> sel.style('stroke-width', (d.isBlank ? 0 : lineWidth) + 'px');
> ```
> 
> while keeping your new:
> 
> https://github.com/plotly/plotly.js/blob/cc871cdb5292bf7c3e75431d3936c2229966cb37/src/traces/bar/plot.js#L141-L142
> 
> and dropping the `attr('display', /**/)` lines

That seemed like a really good idea so I went with it. Unfortunately, the `stroke-width` also transitions which probably isn't what we want. Can you think of another way than setting `display:none` on `transition.end` or `transition.interrupt`?

Thank you @etpinard !

etpinard: > Unfortunately, the `stroke-width` also transitions which probably isn't what we want.

I don't see a problem with that. I mean the `stroke-width` of blank pt is effectively 0, right?

Maybe @archmoj has something to say on this topic.

> Can you think of another way than setting `display:none` on `transition.end` or `transition.interrupt`?

This sounds very hacky and bug-prone to me. I would prefer fixing this problem at the tween level.

test/jasmine/assets/check_transitions.js
```diff
@@ -0,0 +1,134 @@
...
+}
+
+function assertEqual(A, B, cb, msg) {
+    var a = cb(A);
+    var b = cb(B);
+    expect(a).withContext(msg + ' equal to ' + JSON.stringify(a)).toEqual(b);
+}
+
+function assertCloseTo(A, B, precision, cb, msg) {
+    precision = - Math.log10(2 * precision);
```
antoinerg: I wasn't aware we had `toBeWithinArray` so we don't need this ugly conversion from absolute precision to exponential one! :tada:

src/traces/bar/plot.js
antoinerg: > I don't see a problem with that. I mean the `stroke-width` of blank pt is effectively 0, right

Not necessarily, you can have a large `stroke-width` with a `null` value and that qualifies as a blank.

Here's a [Codepen showing the fix you proposed](https://codepen.io/antoinerg/pen/zYOWQJz) built off https://github.com/plotly/plotly.js/commit/bda91f7b373e500434c2d884989da33071a8491d. As you can see, the bar has a decreasing line width as it goes to `null` even though we did not explicitly ask to transition that property. One may argue this is good enough but it's not perfect. Let me know what you think!

etpinard: > Not necessarily, you can have a large `stroke-width` with a `null` value and that qualifies as a blank.

Sure, but the `stroke-width` does not render for blank pts e.g 

```js
  Plotly.newPlot(gd, [{
    type: 'bar',
    x: ['A', 'B', 'C'],
    y:  [null, 5, 3];
    marker: {
      line: {
        width: 10
      }
    }
  }])
```

shows up as:

image

i.e. the left-most bar has (effectively) no `marker.line.width`

etpinard: If someone complains about this, we could potentially add a custom `.attrTween` function for `stroke-width`, but I honestly don't see the need to spend time on this at this time.

test/jasmine/assets/check_transitions.js
antoinerg: I set the tolerance to 3 in 16c88f8!

src/traces/bar/plot.js
antoinerg: Sounds good! I cherry-picked that fix since you approve of it :)

src/components/drawing/index.js
```diff
@@ -442,7 +442,7 @@ drawing.singlePointStyle = function(d, sel, trace, fns, gd) {
                 fill: 'none'
             });
     } else {
-        sel.style('stroke-width', lineWidth + 'px');
+        sel.style('stroke-width', (d.isBlank ? 0 : lineWidth) + 'px');
```
etpinard: Great - tests are passing. Looks like this didn't break @archmoj `isBlank` work from https://github.com/plotly/plotly.js/pull/4056 :tada:

src/traces/bar/plot.js
etpinard: Nicely done! Resolved.

etpinard:
image

@antoinerg please don't do that again on branches associated with open PRs. This makes the `git pull` process a little more annoying than it needs to be. Thanks

---------

test/jasmine/assets/check_transitions.js
etpinard: Very very nice! Thanks!!

etpinard:
Awesome work @antoinerg - this PR is a huge step forward in transitions land.

As we add more `animatable: true` traces, it will be interesting to see if your `transition` wrapper will work in other scenarios - see https://github.com/plotly/plotly.js/pull/4180#pullrequestreview-286183512 for more info.

:dancer: :dancer: :dancer:

PR #4182, opened 2019-09-10
Added docs about building Plotly correctly with Angular

marcj:
Added correct howto to build plotly + angular. Removed old variants since `ng eject` is not possible anymore and not building with webGl is basically useless..

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -42,32 +42,57 @@ browserify index.js > bundle.js
 
 ## Angular CLI
 
-Currently Angular CLI uses Webpack under the hood to bundle and build your Angular application.
-Sadly it doesn't allow you to override its Webpack config in order to add the plugin mentioned in the [Webpack](#webpack) section.
-Without this plugin your build will fail when it tries to build glslify for WebGL plots.
+Since Angular uses webpack under the hood and doesn't allow easily to change it's webpack configuration. There is some work needed using a custom-webpack builder to get things going.
 
-Currently 2 solutions exists to circumvent this issue:
+1. Install [`@angular-builders/custom-webpack`](https://www.npmjs.com/package/@angular-builders/custom-webpack) and [ify-loader@v1.1.0+](https://github.com/hughsk/ify-loader)
+2. Create a new `extra-webpack.config.js` beside `angular.json`.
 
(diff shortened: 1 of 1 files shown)


etpinard:
Thanks very much for the PR!

I've never used plotly.js with angular, so I'll ping @andrefarzat and @TomDemulierChevret who could maybe helps us review your patch.

TomDemulierChevret:
I don't have time to test this right now, but it seems pretty alright.

The only issue for me is that it should include an example with a partial bundle (and if possible a custom bundle) like the existing documentation.

marcj:
What is a "partial bundle" and where is the custom bundle?

TomDemulierChevret:
https://github.com/plotly/plotly.js/blob/master/dist/README.md#partial-bundles
https://github.com/plotly/plotly.js#modules

You don't want to bundle the whole plotly.js lib in your app if you only use a specific subset of charts.

thatsmeta:
I found this PR very helpful when trying to build plotly with Angular.
As far as I understand, you can create a custom bundle in about the same way as described here: https://github.com/plotly/plotly.js#modules.

```
import Plotly from 'plotly.js/lib/core';
import scatter from 'plotly.js/lib/scatter';
import scattergl from 'plotly.js/lib/scattergl';

Plotly.register([
    scatter,
    scattergl
]);

export { Plotly };
```

BUILDING.md
```diff
@@ -42,32 +42,60 @@ browserify index.js > bundle.js
 
 ## Angular CLI
 
-Currently Angular CLI uses Webpack under the hood to bundle and build your Angular application.
-Sadly it doesn't allow you to override its Webpack config in order to add the plugin mentioned in the [Webpack](#webpack) section.
-Without this plugin your build will fail when it tries to build glslify for WebGL plots.
+Since Angular uses webpack under the hood and doesn't allow easily to change it's webpack configuration. There is some work needed using a custom-webpack builder to get things going.
```
alexcjohnson, line 45:
```suggestion
Since Angular uses webpack under the hood and doesn't allow easily to change it's webpack configuration, there is some work needed using a `custom-webpack` builder to get things going.
```

PR #4184, opened 2019-09-10
Fix toggleHover and resetViews modebar buttons for some partial bundle + graph setups 

### LINKED ISSUE #4164: unhandled exception when using plotly-cartesian.js  with scatterternary
Hello,

When the "Toggle show closest data on hover" button is clicked there is an exception triggered in function getNextHover3d, approximately line 395.  sceneIds is undefined.

If I switch to the full dist, the exception does not occur. 
I noticed this in our app using Plotly version 1.48.3 of npm installed plotly.js-cartesian-dist, however it seems to be there with 1.49.4 as well.

Attached in the zip file is a simple html file with ternary plot that demonstrates the problem.  Open dev tools and you'll see the exception.

[Three.zip](https://github.com/plotly/plotly.js/files/3571976/Three.zip)

if you change the dist to point to https://cdnjs.cloudflare.com/ajax/libs/plotly.js/1.49.4/plotly.js the error does not occur.

etpinard:
fix #4164 - by adding fallbacks to `fullLayout._subplots[k]` entries in modebar buttons handlers called on graphs with different subplot arrangements that may or may not be part of one's partial or custom bundle.

The reason why some `fullLayout._subplots[k]` are empty in partial bundles is due to this routine here:

https://github.com/plotly/plotly.js/blob/9b5f0898e13c24f2ad23c9745cac07e48c0e35d9/src/plots/plots.js#L584-L624

which loops over the **registered** base-plot modules - which is a subset of all the available base-plot modules.

cc @plotly/plotly_js - we don't publish partial bundles on CI, so I can't show you a before/after from https://github.com/plotly/plotly.js/issues/4164#issuecomment-530110586 - I hope you'll find the fix trivial enough to not mind. Please let me know if that's not case.

### DIFF
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -343,7 +343,7 @@ function handleCamera3d(gd, ev) {
     var button = ev.currentTarget;
     var attr = button.getAttribute('data-attr');
     var fullLayout = gd._fullLayout;
-    var sceneIds = fullLayout._subplots.gl3d;
+    var sceneIds = fullLayout._subplots.gl3d || [];
     var aobj = {};
 
     for(var i = 0; i < sceneIds.length; i++) {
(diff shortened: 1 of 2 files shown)


archmoj:
@etpinard  Thanks for the fix and adding tests.
:dancer:

Issue #3456, opened 2019-01-22
More transitions


nicolaskruchten:
On the heels of #1849 being closed, here's a list of new transition types/tweens that we'll want to add, probably piece by piece, trace by trace:

- [ ] mark position tweening for all scatter types, and for bars (i.e. ranked bars that change rank)
- [ ] size tweening for everything that has a size: marker size for all scatter traces, height of bar/barpolar, box and violin parameters
- [ ] color tweening for everything that has a color: bars, choropleths, boxes, whatever
- [ ] range tweening for everything that has a range

zoohair:
May I also suggest some zoom+pan transition effect for maps (geo and mapbox) ?

etpinard:
Related: https://github.com/plotly/plotly.js/issues/1687

etpinard:
Related: https://github.com/plotly/plotly.js/issues/932

elaksana:
Any near-future plans to fix element persistence for contour plots?  #4491

iseessel:
Anyone working on this? Would love to have some smooth transitions for scatter3d!

LeoDiep:
I would like to add the candlestick and bar on the list, please

PR #4186, opened 2019-09-11
bar transitions: support object constancy

antoinerg:
Add support for object constancy to bar transitions!

[Codepen](https://codepen.io/antoinerg/pen/vYBjREv)

### DIFF
--- a/src/traces/bar/plot.js
+++ b/src/traces/bar/plot.js
@@ -31,6 +31,13 @@ var appendArrayPointValue = require('../../components/fx/helpers').appendArrayPo
 // padding in pixels around text
 var TEXTPAD = 3;
 
+function keyFunc(d) {return d.id;}
+function getKeyFunc(trace) {
+    if(trace.ids) {
+        return keyFunc;
+    }
+}
+
 function dirSign(a, b) {
     return (a < b) ? 1 : -1;
 }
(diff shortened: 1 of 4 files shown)


etpinard:
Awesome!

Next stop: let's try to find a nice way to test this.

etpinard:
Moreover, can anyone else detect that the text font color transition before the bars' positions are swapped?

antoinerg:
> Moreover, can anyone else detect that the text font color transition before the bars' positions are swapped?

I confirm that is the case. The reason is that we don't have transitions for text size/color but only for its `transform`...

etpinard:
> The reason is that we don't have transitions for text size/color but only for its `transform`... I could definitely add those in before 1.50.

Cool. If you don't have the time to add it for 1.50.0, no big deal. Just make sure to add to the TODO list in https://github.com/plotly/plotly.js/issues/3456

etpinard:
@antoinerg ok to merge?

antoinerg:
@etpinard It looks good to me. I will add text color and font-size tweening in a separate PR

:dancer:

etpinard:
No problem @antoinerg I updated https://github.com/plotly/plotly.js/issues/3456 accordingly. Thanks!!

Issue #4156, opened 2019-08-29
Zoom doesn't work if not enough pixels are selected on X axis


TheJavaGuy:
For some reason, zoom doesn't work if we don't highlight enough pixels over X axis.

Example of unsuccessful zoom:
![plotly_zoom_bug_1](https://user-images.githubusercontent.com/11942401/63956783-6b3bfd00-ca87-11e9-966d-b36946d5ba85.gif)

Example of successful zoom:
![plotly_zoom_bug_2](https://user-images.githubusercontent.com/11942401/63956805-742cce80-ca87-11e9-9e18-b419f6c2bc74.gif)

Code is at https://codepen.io/etpinard/pen/pLOMXR

etpinard:
Thanks for reporting!

etpinard:
The problem is with the block:

https://github.com/plotly/plotly.js/blob/b45304597ff639c7b9e38adf6a6b48efc47455af/src/plots/cartesian/dragbox.js#L432-L436

which removes the zoombox and skips the `relayout` call on drag-end when the zoombox width or height is less than 16px (as [`MINDRAG:=8`](https://github.com/plotly/plotly.js/blob/b45304597ff639c7b9e38adf6a6b48efc47455af/src/plots/cartesian/constants.js#L36)). 

During drag, there's some [logic](https://github.com/plotly/plotly.js/blob/b45304597ff639c7b9e38adf6a6b48efc47455af/src/plots/cartesian/dragbox.js#L353-L404) around `MINZOOM` (set to [20px](https://github.com/plotly/plotly.js/blob/b45304597ff639c7b9e38adf6a6b48efc47455af/src/plots/cartesian/constants.js#L42)), but still, there are ways to have a zoombox passing `MINZOOM` requirements, but that have `Math.min(box.h, box.w) < MINDRAG * 2`, as reported in this issue.

The easiest way to fix this bug, seems to simply move the problematic `removeZoombox` block from the `zoomDone` to the `zoomMove` handler.

Alternatively, we could simply remove the problematic `removeZoombox` block, allowing for smaller zoom boxes relayouts. At first try, this doesn't appear to lead to any undesired effects.

PR #4188, opened 2019-09-13
Fix tiny zoombox behavior

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4156

where zoomboxes smaller than `2*MINDRAG=16px` now lead to an axis range relayout

Before:

Peek 2019-09-13 12-04

After:

Peek 2019-09-13 12-10

cc @plotly/plotly_js


archmoj:
Brilliant fix.
:dancer:

Issue #4193, opened 2019-09-16
slight mod to /devtools/test_dashboard/server.js for Windows dev


greg9504:
Ran into [this problem](https://github.com/http-party/http-server/issues/525) with node http server and ecstatic on Windows.  
Basically the npm start command produces a web page with:

> This page isn’t working
> localhost redirected you too many times.
> Try clearing your cookies.
> ERR_TOO_MANY_REDIRECTS

and the address bar looks like:

> http://localhost:3000/devtools/test_dashboard/////////////////////////

From reading the above linked issue it doesn't look like ecstatic will be updated to 'fix' the issue.

However there is a simple work around, just add index.html to the served url.  So line 26 in server.js becomes:

old:
`open('http://localhost:' + PORT + '/devtools/test_dashboard');`

new:
`open('http://localhost:' + PORT + '/devtools/test_dashboard/index.html');`

and everything works.
It's minor, but figured I'd post this to save someone else some time if they run into the problem.  At least it will show up in the search here with the work around.

etpinard:
Thanks for the headsup! Would you be interested in making a PR?

greg9504:
Sure, I'll give it a try.

greg9504:
OK, after an embarrassing amount of time trying to figure out how to clean up my git fork (I stupidly committed to my master some unrelated work a long time ago), I think I have the PR done correctly.

PR #4198, opened 2019-09-18
change devtools/test_dashboard/server.js to use full url ...

greg9504:
... include index.html to prevent ERR_TOO_MANY_REDIRECTS error under Windows when running the 'npm start' script.

fixes https://github.com/plotly/plotly.js/issues/4193


etpinard:
Thanks very much @greg9504 !

thornjad:
This should no longer be necessary with http-server version 0.12.0

PR #4204, opened 2019-09-19
Update RU and UK locales and fix FR locale keys

KravtsovDV:
In commit https://github.com/plotly/plotly.js/commit/c76b02f37edf2f72649a466136537565b6ef7567 word 'titre' in locale keys (e.g. 'Click to enter X axis **titre**') was replaced with 'title' (e.g. 'Click to enter X axis **title**'), because otherwise those keys are not functional (example: https://codepen.io/KravtsovDV/pen/yLBRoeO).

In commit https://github.com/plotly/plotly.js/commit/18dcbe56cc9de161103ca24688f69cdc4a3eba01 locales for russian and ukrainian were updated. @kserbez can check them.

I have some doubts concerning some translations. Among them:

- I know no equivalent of KDE abbreviation, so it was translated using full (and long) phrases. This could be a problem for small plots.

- I don't know where 'concentration:' in Sankey diagram is used. Better translation than the one I used might exist.

- There are many variants of months abbreviations, so there could be better ones.

Also, there were three-letter abbreviations of days initially, but two-letter abbreviations are more common in russian and ukrainian.


etpinard:
Oh wow, thanks very much for fixing my dumb typos in https://github.com/plotly/plotly.js/pull/4204/commits/c76b02f37edf2f72649a466136537565b6ef7567 !

For 

> * I don't know where 'concentration:' in Sankey diagram is used. Better translation than the one I used might exist.

Here's an example https://codepen.io/etpinard/pen/LYPXVbv?editors=1010 concentration hover info only appears when one sets a `concentrationscale` in a `sankey` trace.

For

> * I know no equivalent of KDE abbreviation, so it was translated using full (and long) phrases. This could be a problem for small plots.

> * There are many variants of months abbreviations, so there could be better ones.

Hmm, it's not obvious to me what's best. Reaching out to another native speaker might be the way to go.

Thanks very much for your help!!

KravtsovDV:
Judging by CodePen example I think current translation for 'concentration:' should be OK.

kserbez:
I've checked the locales as native speaker. Everithing seems fine.

shortMonths and shortDays look way better now!

As for KDE abbreviation, I don't know any abbreviation analog in Russian and Ukrainian and I haven't found them in the Internet. So, current version is semantically ok but might be a little bit long. If it's a problem than one of the solutions could be changing it simply to "KDE".

etpinard:
Thanks very much @KravtsovDV and @kserbez for your help. Merging. This will be part of `v1.50.0`.

PR #4208, opened 2019-09-20
include color-rgba in package and bump gl-cone3d as well as gl-streamtube3d

### LINKED ISSUE #4207: Missing dependency in package.json
`color-rgba` should be included in dependencies:

https://github.com/plotly/plotly.js/blob/56f6983360febe9d5b3bf7427ada9863c372561a/src/traces/parcoords/parcoords.js#L12

I suspect this is a [phantom dependency](https://rushjs.io/pages/advanced/phantom_deps/), as it almost certainly exists in a dependency of plotly.js.

archmoj:
Fix #4207 
The `color-rgba` module is required by `parcoods` trace while it was not listed in the package file.

Also bumped `gl-cone3d` and `gl-streamtube3d` with similar issue & updated their dependencies:
- https://github.com/gl-vis/gl-cone3d/issues/15#issuecomment-533377090
- https://github.com/gl-vis/gl-streamtube3d/issues/9#issuecomment-533374318

cc: @etpinard


etpinard:
Thank you! :dancer:

PR #4209, opened 2019-09-20
remove duplicate keys and data from gl3d_isosurface_math mock

archmoj:
Removed duplicate data volume from a gl3d mock.
cc: @etpinard


etpinard:
Ok - :dancer:

PR #4211, opened 2019-09-23
Add constraint info to parcats click/hover events

jonmmease:
This PR adds a new `constraints` key to the event data emitted by parcats click/hover events.  The keys are dimension indices or `color`, and the values are category values.

When clicking on a path, there will be a key for each dimension. When clicking on a category rectangle, there will be a key for the dimension the category belongs to. The `color` key is only added if:
 1. A color array was provided
 2. The click was on a path or on a category with `hoveron: 'color'`


etpinard:
Nicely done @jonmmease ! :dancer:

jonmmease:
Thanks!

Issue #4212, opened 2019-09-23
scattergl hover error with nulls


bhogan-mitre:
It appears that null values are not handled well by scattergl.  The data that is present is displayed as expected, for example 3 of 4 points are shown in the screenshot below.  However mousing around the plot area throws a lot of errors from the `hoverPoints` function.  For example, `Uncaught TypeError: Cannot read property 't' of undefined`.

![Screen Shot 2019-09-23 at 1 00 49 PM](https://user-images.githubusercontent.com/32336739/65446660-d1a40900-de02-11e9-8c7b-6cb4637c436b.png)

Reproducible example:
https://jsfiddle.net/bhogan_mitre/u41pn9eo/

I believe this is the underlying function at fault: https://github.com/plotly/plotly.js/blob/master/src/traces/scattergl/hover.js#L15

Would it be appropriate to add some checks in that function on `cd` before trying to access `cd[0]`?  I have also seen errors further down where it makes it past this line:
```
var stash = cd[0].t;
```

but fails here:
```
var x = stash.x;
```

I will note that this is only an issue with `scattergl` and not the normal SVG `scatter`.

etpinard:
Thanks very much for reporting!

etpinard:
Looks like this is regression from 1.45.0.

Things were working ok in 1.44.4: https://codepen.io/etpinard/pen/VwZgjom?editors=1010

antoinerg:
It looks like it was working in 1.46.1: https://codepen.io/antoinerg/pen/mdbvrBx?editors=1010
and that the regression was introduced in 1.47.0: https://codepen.io/antoinerg/pen/BaBMLmq?editors=1010

etpinard:
Interesting. It's failing in 1.45.0 though: https://codepen.io/etpinard/pen/NWKoRMJ?editors=1010

antoinerg:
Oh, I'm sorry, disregard https://github.com/plotly/plotly.js/issues/4212#issuecomment-534219458

PR #4213, opened 2019-09-23
Fix scattergl hover over nulls

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4212 - a bug introduced in 1.45.0 as a side-effect of https://github.com/plotly/plotly.js/pull/3578

before: https://codepen.io/etpinard/pen/LYPqWYZ (hover over `x=3` and check console)
after: https://codepen.io/etpinard/pen/YzKBNbp?editors=1010

In `scatter/hover.js`, we used to mutate `pointData.cd` in order to get `scattergl` traces to work with the parts of `Fx.hover` that expect `gd.calcdata[i]` to have `length=trace._length` (which isn't the case for `scattergl`). Before https://github.com/plotly/plotly.js/pull/3578, we used to return early here

https://github.com/plotly/plotly.js/blob/184c0b089ac32ab6a902747b68e2525f5d5a7c20/src/traces/scattergl/hover.js#L48

and the wrong `pointData.cd` mutation did not occur.

cc @archmoj

### DIFF
--- a/test/jasmine/tests/gl2d_click_test.js
+++ b/test/jasmine/tests/gl2d_click_test.js
@@ -346,6 +346,36 @@ describe('Test hover and click interactions', function() {
         .then(done);
     });
 
+    it('@gl should not error when scattergl trace has missing points', function(done) {
+        var _mock = {
+            data: [{
+                type: 'scattergl',
+                mode: 'markers',
+                x: [1, 2, 3, 4],
+                y: [10, 15, null, 17],
+            }],
+            layout: {
+                width: 500,
+                height: 500
+            }
+        };
+
+        Plotly.plot(gd, _mock)
+        .then(function() {
+            gd.on('plotly_hover', function() {
+                fail('should not trigger plotly_hover event');
+            });
+        })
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/gl2d_click_test.js
```diff
@@ -346,6 +346,36 @@ describe('Test hover and click interactions', function() {
         .then(done);
     });
 
+    it('@gl should not error when scattergl trace has missing points', function(done) {
```
archmoj: Interestingly when passing a blank string instead of null there is no error. 
Something like ``` x: [1, 2, 3, 4], y: [10, 15, 16, '' ]```
That may considered to be useful if one wants to set the last items of array to undefined values (without losing the length). 
But curious to know where that handled?

etpinard: Hmm. I'm getting `Cannot read property 't' of undefined` errors on https://codepen.io/etpinard/pen/LYPqWbb which has `''` swapped in for `null`

Issue #3801, opened 2019-04-23
Plotly mobile: Legend cannot be scrolled


avsdev-cw:
Initially I thought this issue was the same as #2844, however I think it is different in one key point.

When displaying a plotly graph on a mobile device (particularly where the page has it's own scrollbar) and a large number of traces, the legend does not capture the press+drag event in the same way as the main plot.  This means that the legend cannot be scrolled up or down on these devices.

Plotly.js version: v1.47.3

Example:
https://codepen.io/anon/pen/MRqOgE

etpinard:
Thanks for reporting.

etpinard:
Related: https://github.com/plotly/plotly.js/issues/2844

andrewbaldock:
Here's a PR you might consider for this issue: https://github.com/plotly/plotly.js/pull/3873

PR #4214, opened 2019-09-24
Add test for legend scroll via touch

etpinard:
a followup on https://github.com/plotly/plotly.js/pull/3873 which will fix https://github.com/plotly/plotly.js/issues/3801

cc @archmoj


archmoj:
@etpinard 
Beautifully done.
FYI - the new test is also passed on my machine.
:dancer:

PR #4219, opened 2019-09-26
Revise hierarchy referencing and fixup parent percentage at the root node of treemap and sunburst 

archmoj:
This PR 
- simplifies the plot and hover code for displaying percentages on `sunburst` and `treemap` plots
- fixes a bug at the root node | now the hover/text template/info are available there
- adds jasmine tests for treemap `hovertemplate` at different levels 
- uploads the CI version of a baseline which started failing after we merged #4185 
@etpinard


src/traces/sunburst/helpers.js
```diff
@@ -52,10 +52,11 @@ exports.getPtId = function(pt) {
 };
 
 exports.getPtLabel = function(pt) {
-    return pt.data.data.label;
+    var label = pt.data.data.label;
+    return label === undefined ? pt.data.data.pid : label;
```
etpinard: Hmm. I don't like this. This helper is supposed to return the `label` not the "label or the pid".

I'd would leave `getPtLabel` along, and move this extra logic in `getParent` or make another helper.

test/jasmine/tests/sunburst_test.js
```diff
@@ -1345,7 +1345,7 @@ describe('Test sunburst texttemplate without `values` should work at root level:
         ['path: %{currentPath}', ['path: /', 'path: Eve/', 'path: Eve/', 'path: Eve/', 'path: Eve/', 'path: Eve', 'path: Eve/Seth', 'path: Eve/Seth/', 'path: Eve/Awan/']],
         ['%{percentRoot} of %{root}', ['100% of Eve', '33% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve']],
         ['%{percentEntry} of %{entry}', ['100% of Eve', '33% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve']],
-        ['%{percentParent} of %{parent}', ['%{percentParent} of %{parent}', '100% of Seth', '33% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '50% of Seth', '100% of Awan']],
+        ['%{percentParent} of %{parent}', ['100% of Eve', '100% of Seth', '33% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '17% of Eve', '50% of Seth', '100% of Awan']],
```
etpinard: Wait. Is this the `root` node sector? You're saying the parent of the root node is itself?

test/jasmine/tests/treemap_test.js
```diff
@@ -629,6 +629,189 @@ describe('Test treemap hover:', function() {
     });
 });
 
+describe('Test treemap hover with levels', function() {
```
etpinard: Can you merge this `describe` block with the other `Test treemap hover` block? The current `run` wrapper should allow you to pass your own `traces` definitions.

src/traces/sunburst/helpers.js
archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/4219/commits/df5e1219dd2892b29d9c6a2fe5ec9f3dd3e4f2ae.

test/jasmine/tests/sunburst_test.js
archmoj: Good call. Fixed in https://github.com/plotly/plotly.js/pull/4219/commits/54360be4be39d7d77a8f88fb147909efa6b037d4.

test/jasmine/tests/treemap_test.js
archmoj: Combined and added one more test for leaves in https://github.com/plotly/plotly.js/pull/4219/commits/54360be4be39d7d77a8f88fb147909efa6b037d4.

test/jasmine/tests/sunburst_test.js
etpinard: I don't get it. After https://github.com/plotly/plotly.js/commit/54360be4be39d7d77a8f88fb147909efa6b037d4, looks like `%{parent}` is now `''`, but `%{percentParent}` is still `100%` for the root node. 

Can we have `'%{percentParent} of %{parent}'` resolve as `'%{percentParent} of %{parent}'` for the root node (i.e. the `percentParent` and `parent` keys are not defined) like before?

I know this is not ideal, but I don't like filling in `'%{percentParent}` and `%{parent}` with _wrong_ value for the root node just to make the template string look better.

src/traces/sunburst/helpers.js
etpinard: Thanks!

test/jasmine/tests/treemap_test.js
etpinard: Ok.

src/traces/sunburst/calc.js
```diff
@@ -128,7 +128,8 @@ exports.calc = function(gd, trace) {
         cd.unshift({
             hasMultipleRoots: true,
             id: dummyId,
-            pid: ''
+            pid: '',
+            label: ''
```
archmoj: @etpinard wondering are you OK with this change?

etpinard: What difference(s) does it make?

archmoj: The changing tests in https://github.com/plotly/plotly.js/commit/df5e1219dd2892b29d9c6a2fe5ec9f3dd3e4f2ae#diff-07da929bf23f8ed55f7b47eb51c0a739
I suggest we keep it blank so that the `getPath` helper function could see it as a blank string not an undefined:
https://github.com/plotly/plotly.js/blob/54360be4be39d7d77a8f88fb147909efa6b037d4/src/traces/sunburst/helpers.js#L165-L167

etpinard: If those `calcdata` diffs are the only difference, then that's not a problem :ok_hand:

test/jasmine/tests/sunburst_test.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4219/commits/7155917f2dd0f4d440abb0fc89f86ff77c88439a.

etpinard:
Nicely done :dancer: thanks!!

PR #4227, opened 2019-09-26
Treemap text positions for out of range cases when transitioning with maxdepth

archmoj:
Fixes the jump happening at the end of entering new text elements of `treemap` graph when `maxdepth` is set.

In the example below the inner text grows too big during the transition and then returns to its normal on transition end.

Peek 2019-09-25 11-08

[Codepen before](https://codepen.io/MojtabaSamimi/pen/gOYyLdq?editors=0010)
[Codepen after](https://codepen.io/MojtabaSamimi/pen/GRKLNLe?editors=0010)

@etpinard

PR #4230, opened 2019-09-27
Bump mapbox-gl to v1.3.2

etpinard:
mostly to fix https://github.com/plotly/plotly.js/network/alert/package-lock.json/esm/open

See what's next in mapbox-gl since the last version bump we made:

- https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#120
- https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#121
- https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#130
- https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#131

cc @archmoj


etpinard:
Note that `mapbox-gl` now allow uses to set their access token _per map_ as opposed to during module instantiation (in https://github.com/mapbox/mapbox-gl-js/pull/8364). 

This means that we could implement a way to plot multiple mapbox subplots with different mapbox access token on the same plotly.js graph.

src/plots/mapbox/mapbox.js
```diff
@@ -367,7 +367,10 @@ proto.resolveOnRender = function(resolve) {
         if(map.loaded()) {
             map.off('render', onRender);
             // resolve at end of render loop
-            setTimeout(resolve, 0);
+            //
+            // Need a 10ms delay (0ms should suffice to skip a thread in the
+            // render loop) to workaround mapbox-gl bug introduced in v1.3.0
+            setTimeout(resolve, 10);
```
etpinard: @archmoj w/o this extra 10ms delay, call sequences like

```js
// off the `mapbox_0` mock
.then(() => Plotly.restyle(gd, 'visible', true))
.then(() => Plotly.restyle(gd, 'visible', 'legendonly', [1]))
```

fails and logs

image

--------------

I suspect this is caused by https://github.com/mapbox/mapbox-gl-js/pull/8633 - but I haven't investigated further.

etpinard:
Looks like we'll also need https://github.com/plotly/plotly.js/commit/422d5d3b0c23fe654ccc1263f41913fac4c953bd to make `test-image` pass on this branch - cc https://github.com/plotly/plotly.js/pull/4227#issuecomment-535763313

---

Let's wait a little bit before. Ideally, I would merge master into this branch after https://github.com/plotly/plotly.js/pull/4219 gets merged.

src/plots/mapbox/mapbox.js
archmoj: OK. I'll leave this conversation open so that it is noticed by in the QA.

archmoj:
@etpinard thanks very much for the PR.
:dancer:

Issue #3371, opened 2018-12-24
Passive event listener error on mobile on drag


athenatan:
Hi,
I got this error from trying to scroll on bar chart surface (on mobile platforms).
`Unable to preventDefault inside passive event listener due to target being treated as passive.`
Also, this makes the hover doesn't work as well.
How to fix this ?
Thank you.

etpinard:
Could you share a the "data" / "layout" you used to generate the graph in question? Thank you.

autukill:
Test with Chrome's Device Mode:
- https://plot.ly/javascript/line-and-scatter/
- [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080

![passive event](https://user-images.githubusercontent.com/5186639/50556618-4c2d7580-0d16-11e9-85a6-19e7035fad9d.png)

etpinard:
Thanks @autukill

PR #4231, opened 2019-09-27
[dragelement] Make touchmove event listener non passive

dan8f:
Fixes https://github.com/plotly/plotly.js/issues/3371.

Touch event listeners are passive by default in some browsers and calling preventDefault inside passive event listeners is not allowed.

This is the only case I've come across but there might be more cases in the code base.


etpinard:
This looks good. I didn't notice any regressions while testing on mobile Chrome and FF

Thanks for the PR!!

---------

I'll go ahead on merge this thing even though `test-image` fails. This problem has been fixed in `master` and won't be an issue going forward.

PR #4242, opened 2019-10-01
use marker.colors with colorscale in treemap and sunburst plots

archmoj:
This PR adds ways to use `marker.colors` numbers with `colorscale`.
Also to mention the `marker.colorscale` would be defaulted if `marker.colors` is not empty.

Other changes within this PR also include:
- implement `marker.depthfade` and drop `marker.opacitybase` and `marker.opacitystep`.
- default pathbar.opacity to 1.

@etpinard 
cc: @nicolaskruchten


src/traces/sunburst/calc.js
```diff
@@ -189,9 +189,10 @@ exports.calc = function(gd, trace) {
     var pullColor;
     var scaleColor;
     var colors = trace.marker.colors || [];
+    var hasColors = !!colors.length;
     trace._hasColorscale = hasColorscale(trace, 'marker');
```
etpinard: Why don't we need 

```js
trace._hasColorscale = hasColorscale(trace, 'marker', 'colors')
```

here?

test/jasmine/tests/sunburst_test.js
```diff
@@ -317,6 +342,88 @@ describe('Test sunburst calc:', function() {
...
+        expect(cd[4].color).toEqual('rgba(0, 255, 0, 1)');
+        expect(cd[5].color).toEqual('rgba(0, 255, 255, 1)');
+        expect(cd[6].color).toEqual('rgba(0, 0, 255, 1)');
+        expect(cd[7].color).toEqual('rgba(255, 0, 255, 1)');
+        expect(cd[8].color).toEqual('rgba(255, 255, 255, 1)');
+    });
+
+    it('should use *marker.colors* numbers with colorscale', function() {
+        _calc({
+            marker: { colors: [1, 2, 3, 4, 5, 6, 7, 8, 9], colorscale: 'Portland' },
```
etpinard: Hmm. What about the cases where `colorscale` isn't set?

By the looks of it, `hasColorscale`

https://github.com/plotly/plotly.js/blob/003f52e4829c082547962bb2e1e0096259ca3d4d/src/components/colorscale/helpers.js#L20-L45

returns `true` when `colorscale` is set regardless of if `marker.color(s)` has numeric items.

src/traces/sunburst/calc.js
archmoj: Is it OK to call `hasColorscale` once in the `default.js` and set `trace._hasColorscale` there?

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4242/commits/2e179940a459bad7208c3ed84393f4065ddcb849.

test/jasmine/tests/sunburst_test.js
archmoj: I think that case is covered:
https://github.com/plotly/plotly.js/blob/1c136f4b1257b4a0558f5bc5699f517324b64300/test/jasmine/tests/treemap_test.js#L276-L291

etpinard: > I think that case is covered:

Not really, this default test essentially just asserts that 

https://github.com/plotly/plotly.js/blob/003f52e4829c082547962bb2e1e0096259ca3d4d/src/lib/coerce.js#L172-L174

returns the `Reds` colorscale array - which is overwritten during the calc step under `autocolorscale: true`:

https://github.com/plotly/plotly.js/blob/003f52e4829c082547962bb2e1e0096259ca3d4d/src/components/colorscale/calc.js#L70-L76

---

With `marker.colors: [0]` it _turns out_ that auto colorscale (after the calc step) is "Reds", but try setting `marker.colors: [-1]`: this makes `marker.colorscale` "Reds" after the defaults, but then "Blues" after the calc step.

So, please a calc test w/o a set colorscale. Thanks!

src/traces/sunburst/calc.js
etpinard: Nicely done!

test/jasmine/tests/sunburst_test.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4242/commits/ba492cf4ce9396b374abfe6f3fdfb020ad1be12c.

src/traces/treemap/style.js
```diff
@@ -36,37 +36,55 @@ function styleOne(s, pt, trace, opts) {
...
     if(hovered) {
         lineColor = trace._hovered.marker.line.color;
         lineWidth = trace._hovered.marker.line.width;
         opacity = trace._hovered.marker.opacity;
     } else {
-        if(helpers.isHierarchyRoot(pt)) {
+        var isRoot = helpers.isHierarchyRoot(pt);
+
+        if(!pt.onPathbar && !trace._hasColorscale && trace.marker.depthfade) {
+            var fadedColor = Color.combine(Color.addOpacity(trace._backgroundColor, 1), fillColor);
```
archmoj: Please note that we could set the `1` on this line to 0.5 or another number if the fading effect is too much.

src/traces/treemap/attributes.js
```diff
@@ -139,29 +139,16 @@ module.exports = {
...
-            max: 1,
-            dflt: 0.5,
             description: [
-                'Sets the increment for opacity of the headers based on the depth from the entry.',
-                'This options is not available when having a `colorscale`.'
+                'Fades headers towards the background.',
+                'When `marker.colors` are not set within the trace it is defaulted to *false*;',
+                'otherwise it is defaulted to *true*.',
+                'This options is not available when having a `colorscale` or',
+                'when .'
```
etpinard: Should read: 

```
'This options is not available when having a `colorscale`.
```

src/traces/treemap/draw_descendants.js
```diff
@@ -65,16 +65,25 @@ module.exports = function drawDescendants(gd, cd, entry, slices, opts) {
...
-        if(pt.depth >= trace._maxDepth) {
+        var depth = pt.depth;
+        if(depth >= trace._maxDepth) {
+            // hide slices that won't show up on graph
             pt.x0 = pt.x1 = (pt.x0 + pt.x1) / 2;
             pt.y0 = pt.y1 = (pt.y0 + pt.y1) / 2;
+        } else {
+            minVisibleDepth = Math.min(minVisibleDepth, depth);
+            maxVisibleDepth = Math.max(maxVisibleDepth, depth);
         }
```
etpinard: Please bring this out of the key function. 

You can have something like:

```js
 var minVisibleDepth = Infinity;
 var maxVisibleDepth = -Infinity;
sliceData.forEach(function(pt) {
        var depth = pt.depth;
        if(depth >= trace._maxDepth) {
            // hide slices that won't show up on graph
            pt.x0 = pt.x1 = (pt.x0 + pt.x1) / 2;
            pt.y0 = pt.y1 = (pt.y0 + pt.y1) / 2;
        } else {
            minVisibleDepth = Math.min(minVisibleDepth, depth);
            maxVisibleDepth = Math.max(maxVisibleDepth, depth);
        }
})

slices = slices.data(sliceData, function(pt) { return helpers.getPtId(pt); });
// or even ;)
slices = slices.data(sliceData, helpers.getPtId)
```

src/traces/treemap/draw_descendants.js
```diff
@@ -185,14 +194,6 @@ module.exports = function drawDescendants(gd, cd, entry, slices, opts) {
...
         });
 
-        if(helpers.isOutsideText(trace, pt)) {
-            // consider in/out diff font sizes
-            pt.transform.targetY -= (
-                helpers.getOutsideTextFontKey('size', trace, pt, fullLayout.font) -
-                helpers.getInsideTextFontKey('size', trace, pt, fullLayout.font)
-            );
-        }
-
```
etpinard: Why did you :hocho: this?

src/traces/treemap/plot.js
```diff
@@ -84,11 +84,16 @@ function plotOne(gd, cd, element, transitionOpts) {
...
     var entry = helpers.findEntryWithLevel(hierarchy, trace.level);
+    var isRoot = helpers.isHierarchyRoot(entry);
+
     var maxDepth = helpers.getMaxDepth(trace);
     var hasVisibleDepth = function(pt) {
         return pt.data.depth - entry.data.depth < maxDepth;
     };
 
+    // copy
+    trace._backgroundColor = fullLayout.paper_bgcolor;
```
etpinard: This is a little dirty, but I guess passing `gd` to `styleOne` isn't much better. Passing `paper_bgcolor` as a `styleOne` `opts` would be a pain too ... so let's leave like you have it (for now).

etpinard: Resolving this comment. Please don't address this in this PR :smirk:

src/traces/treemap/draw_descendants.js
archmoj: With that in place the text was printed outside the domain on this mock: https://github.com/plotly/plotly.js/blob/429324d66493681f67751fff07503dba243bd332/test/image/baselines/treemap_values.png

etpinard: Great. Nice catch. Resolved.

src/traces/treemap/style.js
etpinard: These would be the baselines if we set this to `0.5`:

https://github.com/plotly/plotly.js/compare/treemap-sunburst-colors-in-colorscale...treemap-sunburst-colors-in-colorscale-etpinard

I think I prefer 0.5 over 1, but maybe something in-between would be best.

etpinard: With `0.75`:

https://github.com/plotly/plotly.js/compare/treemap-sunburst-colors-in-colorscale...treemap-sunburst-colors-in-colorscale-etpinard2

What do you think @archmoj ?

etpinard:
Awesome job @archmoj !!

As for the failing tests:

https://github.com/plotly/plotly.js/blob/429324d66493681f67751fff07503dba243bd332/test/jasmine/tests/treemap_test.js#L119-L129

- [ ] This one should not test the `marker.depthfade` isn't coerced when a marker `colorscale` is present.
- [ ] We should also add one default test checking that `marker.depthfade` defaults to false when `marker.colors` is set

--- 

https://github.com/plotly/plotly.js/blob/429324d66493681f67751fff07503dba243bd332/test/jasmine/tests/treemap_test.js#L1193-L1235

- [ ] That one can be simply :hocho: (you _could_ add a restyle test for `marker.depthfade`, but that's 100% not required)


---- 

https://github.com/plotly/plotly.js/blob/429324d66493681f67751fff07503dba243bd332/test/jasmine/tests/treemap_test.js#L1237-L1280

- [ ] I'm not 100% sure what's going here, but I think we simply need to adjust the assertion values

src/traces/treemap/style.js
alexcjohnson: I like `0.75` - still gives good visibility to very deep levels but a clear fade while you're near the top. 🏆

etpinard: Done in https://github.com/plotly/plotly.js/pull/4242/commits/f068cb22f75d124f5428391e230b6a4d36506c39

Thank you very much @alexcjohnson for taking a :eye: at it!

src/traces/treemap/draw_descendants.js
archmoj: Thank you!
Fixed in https://github.com/plotly/plotly.js/pull/4242/commits/280a0a1baf7b7c9e8c4965599ed96402a38fa746. Also revised draw_ancestors and sunburst functions.

etpinard:
One more thing:

- [ ] make sure `pathbar.opacity` is `1` by default

src/traces/treemap/attributes.js
archmoj: Good eye. Also fixed few more typos like that.

src/traces/treemap/draw_descendants.js
etpinard: Awesome!!

archmoj:
> One more thing:
> 
> * [ ]  make sure `pathbar.opacity` is `1` by default

It is `1` when a `colorscale` is present and `0.5` otherwise.
Is that OK?

etpinard:
> It is `1` when a `colorscale` is present and `0.5` otherwise.
> Is that OK?

I don't have a strong opinion about it, but looks like @nicolaskruchten would prefer `1` always.

> I reran https://codepen.io/MojtabaSamimi/pen/poozNQx using this build and the brown went away, which is nice, but I'm not sure why the pathbar is so faded?
actually it was faded before, and I only now noticed it... if I set a marker.color to red I would expect the pathbar element to be fully red and not faded, no?

archmoj:
> > It is `1` when a `colorscale` is present and `0.5` otherwise.
> > Is that OK?
> 
> I don't have a strong opinion about it, but looks like @nicolaskruchten would prefer `1` always.
> 
> > I reran https://codepen.io/MojtabaSamimi/pen/poozNQx using this build and the brown went away, which is nice, but I'm not sure why the pathbar is so faded?
> > actually it was faded before, and I only now noticed it... if I set a marker.color to red I would expect the pathbar element to be fully red and not faded, no?

OK. Let's go with that. Actually it would work better for text transitions.

src/traces/treemap/attributes.js
```diff
@@ -139,29 +139,20 @@ module.exports = {
...
+            values: [true, false, 'reversed'],
             editType: 'style',
             role: 'style',
-            min: 0,
-            max: 1,
-            dflt: 0.5,
             description: [
-                'Sets the increment for opacity of the headers based on the depth from the entry.',
-                'This options is not available when having a `colorscale`.'
+                'Fades headers towards the background. Background color with opacity of 0.75',
```
etpinard: Fades _sectors_ towards the background?

etpinard: even better: Fades _sector colors_ towards the background

src/traces/treemap/attributes.js
```diff
@@ -139,29 +139,20 @@ module.exports = {
...
+                'is used as the destination fading color.',
+                'When `marker.colors` are not set within the trace it is defaulted to *false*;',
+                'otherwise it is defaulted to *true*.',
```
etpinard: Hmm. This should say: This option is unavailable when a `colorscale` is present, defaults to *false* when `marker.colors` is set, but otherwise defaults to *true*.

src/traces/treemap/attributes.js
```diff
@@ -139,29 +139,20 @@ module.exports = {
...
+                'The *reversed* option could be used to invert the direction of fade',
+                'in which the top elements within hierarchy drawn with fully saturated colors',
+                'while the leaves would be faded towards the background color',
+                'This option is not available when a `colorscale` is present.'
             ].join(' ')
```
etpinard: Here's what I would write as `description`:

```
Determines if the sector colors are faded towards the background from the leaves up to the headers.
This option is unavailable when a colorscale is present, defaults to false when marker.colors is set, 
but otherwise defaults to true.
When set to *reversed*, the fading direction is inverted, 
that is the top elements within hierarchy are drawn with fully saturated colors 
while the leaves are faded towards the background color.
```

src/traces/treemap/style.js
```diff
@@ -35,39 +35,65 @@ function styleOne(s, pt, trace, opts) {
...
     } else {
-        if(helpers.isHierarchyRoot(pt)) {
+        var isRoot = helpers.isHierarchyRoot(pt);
+
+        if(!trace._hasColorscale) {
+            if(pt.onPathbar) {
+                // Combining colors with the background color.
+                // Better not to have transparent segments.
+                // Also this helps pathbar texts appear nicely below previous segments.
+                fillColor = Color.combine(Color.addOpacity(trace._backgroundColor, 1 - trace.pathbar.opacity), fillColor);
```
etpinard: Hmm. This logic is ok, but we can't use `pathbar.opacity` for this. Opacity attributes should set some opacity DOM attribute (in SVG traces at least :smirk: )

So, we should either:

- come up with another attribute name to replace `pathbar.opacity`
- or, drop `pathbar.opacity` altogether

At this stage, I would personally vote for dropping `pathbar.opacity`. If some users still want to have some opacity in their pathbar, they could always set an `rgba` color.

src/traces/treemap/style.js
```diff
@@ -35,39 +35,65 @@ function styleOne(s, pt, trace, opts) {
...
-            pt.onPathbar ? trace.pathbar.opacity :
-                depthFade(pt.data.depth - trace._entryDepth);
     }
 
     s.style('stroke-width', lineWidth)
-        .call(Color.fill, cdi.color)
+        .call(Color.fill, fillColor)
         .call(Color.stroke, lineColor)
-        .style('opacity', opacity);
+        .style('opacity', null);
```
etpinard: Do we need to unset `opacity` here? That is, do we set `s.style('opacity', /**/)` elsewhere?

src/traces/treemap/style.js
archmoj: OK. Let's drop that.

src/traces/treemap/style.js
etpinard: Cool. Let's :hocho: it.

etpinard:
@archmoj I'm not sure I understand this mock:

image

Why do the `marker.depthfade: 'reversed'` traces show color gradient while the `marker.depthfade: true` don't?

etpinard:
Moreover, in 

image

why does the root node show up in white now?

archmoj:
> @archmoj I'm not sure I understand this mock:
> 
> image
> 
> Why do the `marker.depthfade: 'reversed'` traces show color gradient while the `marker.depthfade: true` don't?

When `depthfade` is true the leaves are fully saturated. Since there is no space for headers (pad = 0), there is no gradient displayed on the figures on the right.

src/traces/treemap/attributes.js
archmoj: Thanks.
Done in https://github.com/plotly/plotly.js/pull/4242/commits/3b104b143050a6bbeb0853486dd2ac5a3a82802c.

src/traces/treemap/style.js
archmoj: Dropped in https://github.com/plotly/plotly.js/pull/4242/commits/3b104b143050a6bbeb0853486dd2ac5a3a82802c.

src/traces/treemap/style.js
archmoj: Dropped in https://github.com/plotly/plotly.js/pull/4242/commits/3b104b143050a6bbeb0853486dd2ac5a3a82802c but then the opacity 0 is actually used for handling root without color special case in https://github.com/plotly/plotly.js/pull/4242/commits/1dff468da972eb9417c8b3c690e9accfafdf30eb.

archmoj:
> Moreover, in
> 
> image
> 
> why does the root node show up in white now?

To display coloured root cases the root should be drawn.
This is now fixed by https://github.com/plotly/plotly.js/pull/4242/commits/1dff468da972eb9417c8b3c690e9accfafdf30eb.

src/traces/treemap/style.js
```diff
@@ -35,37 +35,57 @@ function styleOne(s, pt, trace, opts) {
...
+    var isRoot = helpers.isHierarchyRoot(pt);
+    var opacity = 1;
 
     if(hovered) {
         lineColor = trace._hovered.marker.line.color;
         lineWidth = trace._hovered.marker.line.width;
-        opacity = trace._hovered.marker.opacity;
     } else {
-        if(helpers.isHierarchyRoot(pt)) {
+        if(isRoot && fillColor === 'rgba(0,0,0,0)') {
```
etpinard: This solution correctly handle the default root case that sets `cdi.color` to `rgba(0,0,0,0)` here:

https://github.com/plotly/plotly.js/blob/1dff468da972eb9417c8b3c690e9accfafdf30eb/src/traces/sunburst/calc.js#L269-L270

but, we'll probably need to extend it all other `rgba` colors with a `0` alpha channel - as the current depthfade algorithm gives a white `fillColor`.

etpinard:
All right. Time to merge this thing :dancer: 

---

@archmoj please keep in mind https://github.com/plotly/plotly.js/pull/4242#pullrequestreview-298239454 for future treemap work.

archmoj:
@etpinard Thanks very much for allocating enough time for this PR.

Issue #28, opened 2015-11-23
Provide TypeScript definitions (aka d.ts file)


david-bouyssie:
The TypeScript transpiler is a useful tool for the development of web applications with a large codebase.

Popular graphics libraries such as Highcharts provides typescript definition files allowing to use their API in a safe way when called from TypeScript code.

It would be awesome if such a definition file was available for plotly.js.
The work consists in creating a TypeScript interface for each object of the public Javascript API.

xuyang2:
+1

etpinard:
I've never used typescript before. 

Could you point me to good resources or examples of other projects that support typescript (I couldn't find the relevant file in the highcharts repo).

xuyang2:
@etpinard take a look at https://github.com/DefinitelyTyped/DefinitelyTyped/

david-bouyssie:
@etpinard the Handbook from the Microsoft website is a good starting point.
http://www.typescriptlang.org/Handbook
You may also be interested by their tutorials and samples:
http://www.typescriptlang.org/Tutorial
http://www.typescriptlang.org/Samples

This blog is also an interesting resource: http://www.johnpapa.net/typescriptpost1/

As xuyang2 said the DefinitelyTyped repo is the place the search for existing TS declaration files.
Here is the folder corresponding to Highcharts: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/highcharts

About IDE support:
* Visual Studio has a built-in support of TypeScript
* Intellij and Webstorm also provide a support through a free plugin. I think it now installed by default.
* Under Eclipse (I develop my TS project under Eclipse) I suggest to use the free plugin from the palantir company: https://github.com/palantir/eclipse-typescript

Now about the methodology you have two possibilities:
1. Keep the development of plotly.js in pure JS and provide TS declaration files. They will contain interfaces describing your functions, classes and objects. This will be done in a "d.ts" file (d standing for declaration).
2. Develop plotly.js directly in TS. You will benefit from the type-safety of the language which simplifies the refactoring of big projects. The compiler will generate for you a JS file that could be used as a dro-pin replacement of plotly.js and a plotly.d.ts file corresponding to the API of your plotly.ts file. This plotly.d.ts will be useful for other TS developers.

I will suggest to start with approach 1. and to switch to 2. if you think it is useful for your team.
Note that the declarations defined in approach 1 could leveraged when switching to the TS dev, so you won't loose your work.

[...]

PR #4255, opened 2019-10-04
add instruction for TypeScript usage

simon04:


etpinard:
First off, thanks for the PR.

Hmm, I'm not 100% sure we should include this in our Readme as we have no control over the [typescript](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/plotly.js) definitions (cf https://github.com/plotly/plotly.js/issues/28). 

Furthermore, no current plotly.js team member has any experience using typescript, so we can't really put our stamp-of-approval on them.

At the very least, I don't think we include typescript info in the "Quick start options" section. Personally, I think this belongs in a new section below "Building plotly.js".

nicolaskruchten:
We could add a section that says something like "Related third-party projects" or something and list the TS thing there? I agree we don't want to stamp our approval to the TS typings at this time, not because they're bad, but because we can't commit to maintaining them.

simon04:
Thank you, @etpinard and @nicolaskruchten, for your review and comments. I agree that having a separate section for 3rd party contribution is more adequate, and addressed this in the commit https://github.com/plotly/plotly.js/pull/4255/commits/f5e332983d7155b54e1fcc7c85bee1f5308c01b6.

archmoj:
@simon04 could you please merge master into this branch and resolve the conflict?

Issue #4233, opened 2019-09-29
3D plots broken in Chrome 77 on GNU/Linux and Android


dpq:
### Summary
In Chrome@Linux v.77, 3D plots are broken.

### Steps to reproduce
For example, visit the [docs](https://plot.ly/python/3d-scatter-plots/). Other 3D plots used in production are similarly broken in Chrome 77. 

### What is observed
In error console, ```WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost```. In the browser:
![image](https://user-images.githubusercontent.com/64569/65834424-569b9080-e2e3-11e9-973b-b3b17affd814.png)

### What is expected
Normally functioning 3D plots.

### Additional information
WebGL is enabled; get.webgl.org shows a spinning cube; Chrome WebGL experiments such as Aquarium work properly.

etpinard:
Thanks very much reporting.

We've had some reports already in https://github.com/plotly/plotly.js/issues/3640, could you tell which Linux flavor/version you're using and some info about the hardware you're on. Thank you!

archmoj:
With Chrome 77 on Ubuntu the very first graph displays the error and then a combination of working and non-working veiwports like the one below:
![Screenshot from 2019-09-30 10-35-01](https://user-images.githubusercontent.com/33888540/65888820-2947f480-e36e-11e9-9944-bbfbef82138a.png)

longwei66:
Same here, not working with Version 77.0.3865.90 (Build officiel) (64 bits) - Ubuntu 18.04

WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost

See example here : https://datapleth.io/blog/2019/08/28/2019-08-28_the_french_yield_curve_problem/
![image](https://user-images.githubusercontent.com/1811577/65969313-da30bb00-e464-11e9-9f8c-9e4af0bdb0fb.png)

No problem on firefox 69.0.1 (64 bits) and chromium v. 76 on another computer (wayland).

The example was generated by plotly on R :+1: 
https://cloud.r-project.org/src/contrib/plotly_4.9.0.tar.gz

Source code here : 
https://github.com/longwei66/datapleth/blob/master/content/blog/2019-08-28_the_french_yield_curve_problem.Rmd

archmoj:
@dpq @longwei66 
Could you please test [this codepen](https://codepen.io/MojtabaSamimi/pen/pozMRaJ?editors=0010) in Chrome 77 browser?
Thanks.

[...]

PR #4256, opened 2019-10-06
Enable second try to create gl3d scenes for Chrome 77 webgl bug

archmoj:
Addresses https://github.com/plotly/plotly.js/issues/4233.
The very first try to create a 3-D scene in Chrome 77 is not successful.
This patch enables the second attempt to create `gl3d` scene in case the first try was not successful.
@etpinard


etpinard:
Thanks @archmoj :dancer: 

Let's merge this in, but let's keep https://github.com/plotly/plotly.js/issues/4233 open until we found a better solution.

KamalGalrani:
@archmoj 
This fixes the wegl not supported error. But I need to resize the browser for it to render. Before that I get a blank screen

PR #4268, opened 2019-10-10
Check for typed array in minExtend

jtvd78:
Modified minExtend to check for both arrays and TypedArrays.

Currently, if a trace contains a typedArray, minExtend will treat it it as if it's not an array and will re-create the entire array as opposed to simply taking a slice of the array. To fix this, this change checks if the current value is an array in addition to checking if it is a TypedArray.

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -660,7 +660,7 @@ lib.minExtend = function(obj1, obj2) {
         v = obj1[k];
         if(k.charAt(0) === '_' || typeof v === 'function') continue;
         else if(k === 'module') objOut[k] = v;
-        else if(Array.isArray(v)) {
+        else if(lib.isArrayOrTypedArray(v)) {
             if(k === 'colorscale') {
                 objOut[k] = v.slice();
             } else {
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -660,12 +660,14 @@ lib.minExtend = function(obj1, obj2) {
         v = obj1[k];
         if(k.charAt(0) === '_' || typeof v === 'function') continue;
         else if(k === 'module') objOut[k] = v;
-        else if(lib.isArrayOrTypedArray(v)) {
+        else if(Array.isArray(v)) {
             if(k === 'colorscale') {
                 objOut[k] = v.slice();
             } else {
                 objOut[k] = v.slice(0, arrayLen);
             }
+        } else if(lib.isTypedArray(v)) {
+            objOut[k] = v.subarray(0, arrayLen);
         } else if(v && (typeof v === 'object')) objOut[k] = lib.minExtend(obj1[k], obj2[k]);
         else objOut[k] = v;
     }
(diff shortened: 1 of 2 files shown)


jtvd78:
Sorry for the failing PR. Looks like I gotta put a little more work into this

etpinard:
Thanks for the PR!

Can I ask why you're patching this line? Does not considering typed arrays lead to any bugs? If so, can you provide a reproducible example? If it's only a performance gain, could you provide some before/after benchmarks?

jtvd78:
I've created a benchmark to show the problem's I'm having. The project I'm working on involves plotting many thousands of points, and due to performance problems, we decided to try to use TypedArrays instead of standard javascript arrays. 

The issue we're running into seems to be related to rendering the legend. minExtend ends up re-creating the entire trace due to it not recognizing that the trace we pass in is a TypedArray. 

Please see this codepen for more details: 

https://codepen.io/jvandort/pen/OJJyQpq

Open up chrome dev tools, click on the performance tab, and begin recording. Then, click 'Reload' once, and then again. 

The first time, Plotly.React is called with a standard javascript array. The second time, it is called with a TypedArray. You can see that minExtend takes up the majority of the processing time for the second round (with TypedArray) since it ends up re-creating the entire array. 

I've also attached a screenshot of the chrome  dev tools performance tab while I ran the procedure that I posted above. 

proof

etpinard:
Wow. Thanks so much for spotting that @jtvd78 !

You're absolutely right: something is up in our legend code and your patch does seem to resolve most of the slow perf.

Let me see if I can identify why some of our tests are failing.

etpinard:
Ok, I got the tests to pass on this branch:

https://github.com/plotly/plotly.js/compare/pr-4268-etpinard

----

Turns out that your patch did fix a bug:

```js
        Plotly.newPlot(gd, [{
            x: new Float32Array([1, 2, 3]),
            y: new Float32Array([1, 2, 1]),
            marker: {
                size: new Float32Array([40, 30, 20]),
                color: new Float32Array([20, 30, 10]),
                cmin: 10,
                cmax: 30,
                colorscale: [
                    [0, 'rgb(255, 0, 0)'],
                    [0.5, 'rgb(0, 255, 0)'],
                    [1, 'rgb(0, 0, 255)']
                ]
            }
        }], {
          showlegend: true
        })
```

currently renders as

image

and now off your branch

image

with the correct legend item marker color :tada: so that `scatter_test.js` should now be updated :white_check_mark: 

Next, `.slice` isn't supported for typed arrays on some (old) browsers (like the nw.js version we used for our image tests). Using `.subarray` for typed arrays in `Lib.minExtend` makes the image tests pass again.

----

@jtvd78 feel free to pull down my branch and cherry-pick my commits or make the changes yourself.

If you haven't replied by next Tuesday, I'll merge my [`pr-4268-etpinard`](https://github.com/plotly/plotly.js/compare/pr-4268-etpinard) branch so that it is released as part of `v1.50.1`.

jtvd78:
Not sure what's going on here. I've updated my branch with your changes but it seems there are still failing tests

etpinard:
@jtvd78 the failures on CI were just intermittent. After rebuilding the tests are now :white_check_mark: 

Thanks very much again for this PR! Merging!!

----

> but I can't seem to run `npm run test-jasmine` without getting failing tests, even on master. Am I doing something wrong?

Yeah, that can happen unfortunately. The tests are tuned for the hardware and Chrome version found on our CI. For 99% of our tests that doesn't matter, but for some tests that depend on assertions in pixel space and font rendering we have noticed some machine-to-machine discrepancies.

Issue #4157, opened 2019-08-29
Cannot read property _subplot of undefined


acrhg:
I have a dashboard with several tabs and charts/tables. The tab that renders on initial page load has an array of five gauge charts (pie charts cut in half with a hole and a line drawn for an indicator) at the top. It looks like this: 
<a href="http://google.com.au/" rel="some text">![Foo](https://aws1.discourse-cdn.com/standard17/uploads/plot/original/2X/4/42f77eb3ae4c1a67a08091dc78c5542bad71f033.png)</a>

Occasionally, one of the charts will not render, like this:
<a href="http://google.com.au/" rel="some text">![Bar](https://aws1.discourse-cdn.com/standard17/uploads/plot/original/2X/5/518ebdb8a0c7edac31c8cc6857daccd7018702ce.png)</a>

Here is the error I get in the console:

<a href="http://google.com.au/" rel="some text">![Bar](https://aws1.discourse-cdn.com/standard17/uploads/plot/original/2X/5/5087766b7e62bad3fcecb10fd83f9bb3f448a72b.png)</a>

Original post is [here](https://community.plot.ly/t/chart-randomly-does-not-render/27978). I know this is the part of dash framework but I think something it has to do with the plotly.js. Any help would be really appreciated. Thanks

etpinard:
Thanks for posting!

It is indeed an error coming from plotly.js. 

The first step in debugging would to convert your example into pure plotly.js. Maybe someone could help us out.

etpinard:
I wasn't able to replicate the problem when adapting the python code provided in https://community.plot.ly/t/chart-randomly-does-not-render/27978 in JS in a codepen.

Closing due to lack of reproducible example.

slishak:
I know it doesn't help much, but for the record I have also been experiencing this issue for a while. I've so far also been unable to create a reproducible example, but it appears to be caused by repeatedly redrawing a Plotly graph in Dash. I've usually only see it after working with an app for a while and redrawing a plot 20-30 times with different data and layouts, eventually it will throw this error and stop updating until the page is refreshed. Anecdotally it seems to happen most often when I change the layout (e.g. number of subplots). I think it's related to Dash re-using some of the resources from the existing canvas when drawing an updated figure?

alexcjohnson:
reported again https://community.plot.ly/t/dash-app-with-js-clientside-callback-sometimes-crashes/29599

etpinard:
Ok - reopening.

[...]

PR #4269, opened 2019-10-10
Do not attempt to re-hover on exiting subplots

etpinard:
hopefully fixes https://github.com/plotly/plotly.js/issues/4157 and related `Cannot read _subplot of undefined` issues.

@plotly/plotly_js this bug has been plaguing dash users for some times now (see https://community.plot.ly/t/dash-app-with-js-clientside-callback-sometimes-crashes/29599/5 for a reproducible case).

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -224,17 +224,11 @@ function _hover(gd, evt, subplot, noHoverEvent) {
     for(var i = 0; i < len; i++) {
         var spId = subplots[i];
 
-        // 'cartesian' case
-        var plotObj = plots[spId];
-        if(plotObj) {
+        if(plots[spId]) {
+            // 'cartesian' case
             supportsCompare = true;
-
-            // TODO make sure that fullLayout_plots axis refs
-            // get updated properly so that we don't have
-            // to use Axes.getFromId in general.
-
-            xaArray[i] = Axes.getFromId(gd, plotObj.xaxis._id);
-            yaArray[i] = Axes.getFromId(gd, plotObj.yaxis._id);
+            xaArray[i] = plots[spId].xaxis;
+            yaArray[i] = plots[spId].yaxis;
         } else if(fullLayout[spId] && fullLayout[spId]._subplot) {
             // other subplot types
             var _subplot = fullLayout[spId]._subplot;
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -235,13 +235,15 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
             xaArray[i] = Axes.getFromId(gd, plotObj.xaxis._id);
             yaArray[i] = Axes.getFromId(gd, plotObj.yaxis._id);
-            continue;
+        } else if(fullLayout[spId] && fullLayout[spId]._subplot) {
+            // other subplot types
+            var _subplot = fullLayout[spId]._subplot;
+            xaArray[i] = _subplot.xaxis;
+            yaArray[i] = _subplot.yaxis;
+        } else {
+            Lib.warn('Unrecognized subplot: ' + spId);
+            return;
         }
-
-        // other subplot types
-        var _subplot = fullLayout[spId]._subplot;
-        xaArray[i] = _subplot.xaxis;
-        yaArray[i] = _subplot.yaxis;
     }
 
     var hovermode = evt.hovermode || fullLayout.hovermode;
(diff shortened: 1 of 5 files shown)


etpinard:
To include in `v1.50.1`

src/plots/cartesian/graph_interact.js
```diff
@@ -59,7 +59,7 @@ exports.initInteractions = function initInteractions(gd) {
                 // This is on `gd._fullLayout`, *not* fullLayout because the reference
                 // changes by the time this is called again.
                 gd._fullLayout._rehover = function() {
-                    if(gd._fullLayout._hoversubplot === subplot) {
+                    if((gd._fullLayout._hoversubplot === subplot) && gd._fullLayout._plots[subplot]) {
```
archmoj: What about putting this condition test inside `Fx.hover`?

etpinard: Why would that be better?

archmoj: Just in case that the function is called from different places.
Otherwise; this performs better.

alexcjohnson:
Nice 🎉 presumably this works for disappearing non-cartesian subplots as well as cartesian?

etpinard:
> presumably this works for disappearing non-cartesian subplots as well as cartesian?

Presumably, we shouldn't be calling `Plots.rehover` on non-cartesian subplots ... but hold on looks like mapbox subplots define a `_rehover` function too:

image

Thanks for the tip!

I'll double-check if something similar happens to mapbox subplots before merging this PR.

src/plots/cartesian/graph_interact.js
etpinard: You're making a good point here, consider:

```js
        /*var data = [
            {y: [1], hoverinfo: 'y'},
            {y: [2], hoverinfo: 'y', xaxis: 'x2', yaxis: 'y2'}
        ];*/

        var data2 = [
            {y: [1], hoverinfo: 'y'}
        ];

        var layout = {
            xaxis: {domain: [0, 0.5]},
            xaxis2: {anchor: 'y2', domain: [0.5, 1]},
            yaxis2: {anchor: 'x2'},
            width: 400,
            height: 200,
            margin: {l: 0, t: 0, r: 0, b: 0},
            showlegend: false
        };
Plotly.newPlot(gd, data2, layout)

// which is part of our semi-official API
Plots.Fx.hover(gd, {}, 'x2y2')
```

fails with the same `Cannot read _subplot of undefined ` error.

---

So, I'll keep the check and `graph_interact.js` (to not have to call `Fx.hover` for nothing) AND  handle the empty subplot case in `Fx.hover` (+ a test).

Thanks for the headsup @archmoj !

alexcjohnson:
> Presumably, we shouldn't be calling `Plots.rehover` on non-cartesian subplots

that's not the implication of these comments... anyway regardless of the eventual *fix*, the *test* is pretty clear, hover on a subplot of each type while it disappears. Seems a bit overkill for a very simple fix though, also I've never seen users complaining about this with anything but cartesian. But it would be nice to have at least manually tested that this works.

https://github.com/plotly/plotly.js/blob/25fe1c929672461e0b915a737134f6a975edf8f8/src/components/fx/hover.js#L227-L241

src/plots/cartesian/graph_interact.js
etpinard: done in https://github.com/plotly/plotly.js/pull/4269/commits/89e52f71eb663bbaa0e045757c78c38454991569

etpinard:
@archmoj this PR is ready for a 2nd round of review.

src/components/fx/hover.js
```diff
@@ -224,24 +224,20 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
-
-            // TODO make sure that fullLayout_plots axis refs
-            // get updated properly so that we don't have
-            // to use Axes.getFromId in general.
-
-            xaArray[i] = Axes.getFromId(gd, plotObj.xaxis._id);
-            yaArray[i] = Axes.getFromId(gd, plotObj.yaxis._id);
-            continue;
+            xaArray[i] = plots[spId].xaxis;
+            yaArray[i] = plots[spId].yaxis;
```
etpinard: `gd._fullLayout._plots[/* subplot id */].(x|y).axis` is now always filled in properly thanks for [`Plots.linkSubplots`](https://github.com/plotly/plotly.js/blob/25fe1c929672461e0b915a737134f6a975edf8f8/src/plots/plots.js#L825-L946) called during `Plots.supplyDefaults`.

The `TODO` comment above was over three years old.

archmoj:
Excellent.
:dancer: :dancer:

PR #4283, opened 2019-10-17
Fix width of box plots with log-scale axis

s417-lama:
Fixes an issue of box plots that box width is also log-scaled with log axis (fixes plotly#3508).
I changed the setting of box widths from raw-value-based one to linearized-value-based one.

This PR is pre-reviewed in https://github.com/s417-lama/plotly.js/pull/1 by @etpinard .


src/plots/cartesian/autorange.js
```diff
@@ -371,16 +374,22 @@ function findExtremes(ax, data, opts) {
...
+            dmin = ax.c2l(di) - vpadminus(i);
+            dmax = ax.c2l(di) + vpadplus(i);
+        } else {
+            vmin = di - vpadminus(i);
+            vmax = di + vpadplus(i);
+            // special case for log axes: if vpad makes this object span
+            // more than an order of mag, clip it to one order. This is so
+            // we don't have non-positive errors or absurdly large lower
+            // range due to rounding errors
+            // if(isLog && vmin < vmax / 10) vmin = vmax / 10;
```
etpinard: What happens when you uncomment this line?

src/traces/box/cross_trace_calc.js
```diff
@@ -125,7 +125,6 @@ function setPositionOffset(traceType, gd, boxList, posAxis) {
         t.bPos = bPos;
         t.bdPos = bdPos;
         t.wHover = wHover;
-
```
etpinard: Minor :cow2: comment: can you bring back this newline. Thanks!

src/plots/cartesian/autorange.js
s417-lama: It seems to be an error. Fixed in commit 1876e8c5caaad78ca7d50fa0282c8e1c780b9d00
(Maybe I forgot to uncomment this line while checking its behaviour)

src/traces/box/cross_trace_calc.js
```diff
@@ -213,6 +212,7 @@ function setPositionOffset(traceType, gd, boxList, posAxis) {
             padded: padded,
             vpadminus: vpadminus,
             vpadplus: vpadplus,
+            vpadLinearized: true,
```
etpinard: So, could you explain why we _need_ to pad in linear space (`l`) for log axes as opposed to in calc space (`c`)? 

Padding in calc space on log axes seems to work ok in other scenarios like scatter-marker traces.

test/image/mocks/box_log_scale.json
```diff
@@ -0,0 +1,94 @@
...
+                10,
+                10,
+                10,
+                100,
+                100,
+                100,
+                100,
+                100
+            ],
+            "type": "box"
```
etpinard: Have you tried plotting these samples as `type: 'violin'` traces?

Violin and box traces share a lot of the same code, so it might be nice to fix the log-position problems for violin traces in the same go. 

@s417-lama please let us know if you're not interested in working on a fix for `violin` traces. We should be able to find some time to work on it before releasing `v1.51.0`.

src/traces/box/cross_trace_calc.js
s417-lama: I think paddings in scatter-marker traces are not based on values (`vpad`), but pixels (`ppad`). So the padding is not affected by axes types.
Another example that uses values as margins is box traces, but box traces should be based on `c` values rather than `l` values.

test/image/mocks/box_log_scale.json
s417-lama: Oh, I didn't consider violin traces.
I would be glad for you to work on violin plots. I appreciate it.

etpinard: Great. Thanks!! The fix for violins should be very similar than the one for boxes.

s417-lama: Fixed! 349ef640517fc2429e8fb9364f7c30925507fddc

test/image/mocks/box_log_scale.json
```diff
@@ -0,0 +1,94 @@
+{
```
etpinard: Thanks very much for taking a shot a `violin` traces on log axes!

There seem to be a small problem with the position of the violin "box":

image

the tall line at the right of the box should be drawn where the shading from the one-side violin ends.

etpinard: Actually this has nothing to do with your PR, all vertical one-sided violins have their box line shifted in the wrong direction e.g.: 

https://codepen.io/etpinard/pen/RwwZZzj?editors=1010

----

No need to address this here. I'll make a PR in the next few minutes fixing that.

etpinard:
Awesome work @s417-lama - this is one of the best community PRs we've received in recent memory. 

I was hoping to find a slightly less convoluted way to dance between `c`, `l` and `p` coordinates, but looks like your work is as clean it gets. 

Thank you very much. Merging. This will be released in `v1.51.0`.

jeonsimon:
@etpinard Could you check my issue?  https://github.com/plotly/plotly.py/issues/3970#issue-1468932954

PR #4284, opened 2019-10-17
Update connectgaps description in respect to heatmap and contour default logics

archmoj:
The `connectgaps` attributes of `heatmap` and `contour` traces has some default logics which could result in enabling by default.
So they are not `dflt: false`.

This PR updates the description and add tests for different scenarios.

@etpinard


etpinard:
Nicely done :dancer: . Thanks!

Issue #3898, opened 2019-05-23
raster trace type: matplotlib imshow equivalent


nicolaskruchten:
Could/should we have some trace type that would be equivalent to matplotlib's `imshow()` for showing full-color images? https://matplotlib.org/api/_as_gen/matplotlib.pyplot.imshow.html

basically heatmap but with r,g,b channels or something?

nicolaskruchten:
cc @jonmmease @emmanuelle

nicolaskruchten:
I feel like this would help Python adoption, as this comes up every once in a while.

nicolaskruchten:
Thinking out loud here... `imshow` accepts either a single-channel vector, or RGB or RGBA, but we in principle could also accept HSV or other color-space specifications. This would allow us to do some cool combined-colormap type stuff where hue is mapped to one vector and lightness to another.

etpinard:
related: https://github.com/plotly/plotly.js/issues/332

nicolaskruchten:
this kind of thing https://en.wikipedia.org/wiki/Domain_coloring

![image](https://user-images.githubusercontent.com/203523/58272019-77fbd700-7d5b-11e9-9b17-f7867875200b.png)

alexcjohnson:
I'm for it - I really like the idea of being able to specify 3 (or 2 or 4) 2D arrays and specifying whether they're RGB, HSV, HSL, or whatever. I had some instances in grad school of multiple independent 2D channels of data I wanted to overlay... never did find a mapping that was intelligible for 3 channels but for 2 channels either R+G, R+B, or L+H worked fairly well in my case.

emmanuelle:
It would be great, yes!

emmanuelle:
RGB channels for a start would be enough for 99% cases I think

nicolaskruchten:
The third example on this page shows how a `rastermapbox` related trace type might work: http://datashader.org/topics/gerrymandering.html

YoniChechik:
@nicolaskruchten it would be so useful to computer vision in python! hope to see it soon. Thanks!

antoinerg:
> it would be so useful to computer vision in python! hope to see it soon. Thanks!

@YoniChechik for your use cases are there additional features you would find valuable in this new trace type?

YoniChechik:
[...]

PR #4289, opened 2019-10-21
image trace type

antoinerg:
This PR introduces a new `image` trace type to display images on cartesian axes and closes https://github.com/plotly/plotly.js/issues/3898 

Codepen showing the different mocks: https://codepen.io/antoinerg/pen/OJJPxJK

Quick overview:
- [x] Support different colormodels (`rgb`, `rgba`, `hsl` and `hsla`).
- [x] Can rescale data via `zmin` and `zmax`
- [x] Display data and color information on hover

Todo:
- [x] handle `z` array with missing values (ie. columns with shorter length) (done in 15c06fa1814fa41b2c7a4a60d6d06cd6b90c4a5d)
- [x] support categorial axes
- [x] support log axis

To discuss:
- [x] the trace will be name `image` ~~decide the trace's name. ~~ Scroll down below to read different points of view on the matter. Here's a compiled a list of candidates:
  1. `image`
  2. `raster`
  3. `rasterimage`
  4. `pixmap`
  5. `imshow`
- [x] decide on the default hover label. Display the color or not?


etpinard:
Is this PR `in progress` or `reviewable`?

antoinerg:
@etpinard It's ready for a first round of review !

lib/index.js
```diff
@@ -56,6 +56,7 @@ Plotly.register([
 
     require('./sankey'),
     require('./indicator'),
+    require('./image'),
```
etpinard: Can you move this `require` statement up in this block:

https://github.com/plotly/plotly.js/blob/e86c95b4b2abe646d7ab4e311fcd40cc61f1eaea/lib/index.js#L15-L25

with all the other cartesian traces.

src/traces/image/attributes.js
```diff
@@ -0,0 +1,84 @@
...
+    colormodel: {
+        valType: 'enumerated',
+        values: ['rgb', 'rgba', 'hsl', 'hsla'],
+        dflt: 'rgb',
+        role: 'info',
+        editType: 'plot',
+        description: 'Color model used to map the numerical color components described in `z` into colors.'
+    },
+    zmin: {
+        valType: 'data_array',
```
etpinard: We should make `zmin` and `zmax` have `valType: 'info_array'`:

https://github.com/plotly/plotly.js/blob/6e9bafb8285f7247407a503e7dc33e44ff637f31/src/lib/coerce.js#L254-L321

src/traces/image/attributes.js
```diff
@@ -0,0 +1,84 @@
...
+    zmax: {
+        valType: 'data_array',
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Array defining the higher bound for each color component.',
+            'For example, for the `rgba` colormodel, the default value is [255, 255, 255, 1].'
+        ].join(' ')
+    },
+    x0: {
```
etpinard: So no `x` and `y` for this iteration? Is that intended?

src/traces/image/attributes.js
```diff
@@ -0,0 +1,84 @@
...
+        valType: 'number',
+        dflt: 0,
+        role: 'info',
+        editType: 'calc',
```
etpinard: You might need `editType: 'calc+clearAxisTypes'` for `x0` and `y0` in order to restyle to work properly in situations like:

```js
Plotly.newPlot(gd, [{
  z: /**/,
  x0: 2
}])

Plotly.restyle(gd, 'x0', 'a')
```

src/traces/image/attributes.js
```diff
@@ -0,0 +1,84 @@
...
+        dflt: 1,
+        role: 'info',
+        editType: 'calc',
+        description: 'Set the pixel\'s vertical size'
+    },
+    hoverinfo: extendFlat({}, plotAttrs.hoverinfo, {
+        flags: ['x', 'y', 'z', 'color', 'name']
+    }),
+    hovertemplate: hovertemplateAttrs({}, {
+        keys: ['z', 'c', 'colormodel']
```
etpinard: What's `c` here? Is it the same value as for the `'color'` hoverinfo flag?

src/traces/image/calc.js
```diff
@@ -0,0 +1,35 @@
...
+var Axes = require('../../plots/cartesian/axes');
+
+module.exports = function calc(gd, trace) {
+    var xa = Axes.getFromId(gd, trace.xaxis || 'x');
+    var ya = Axes.getFromId(gd, trace.yaxis || 'y');
+
+    var x0 = trace.x0 - trace.dx / 2;
+    var y0 = trace.y0 - trace.dy / 2;
+    var h = trace.z.length;
+    var w = trace.z[0].length;
```
etpinard: So we assume `z` to be a 2D arrays with inner arrays all of the same the same length?

src/traces/image/calc.js
```diff
@@ -0,0 +1,35 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+// var Registry = require('../../registry');
+// var Lib = require('../../lib');
```
etpinard: :zap: those commented lines please

src/traces/image/defaults.js
```diff
@@ -0,0 +1,36 @@
...
+var Lib = require('../../lib');
+var attributes = require('./attributes');
+var constants = require('./constants');
+
+module.exports = function supplyDefaults(traceIn, traceOut) {
+    function coerce(attr, dflt) {
+        return Lib.coerce(traceIn, traceOut, attributes, attr, dflt);
+    }
+    var z = coerce('z');
+    if(z === undefined || !z.length) {
```
etpinard: We might want to check that `z[0]` and `z[0].length` are also truthy here. I can't imagine having 1-dimensional raster data - am I right?

src/traces/image/attributes.js
etpinard: ... moreover, do we really need `colormodel` in here? It's a per-trace attribute, so it should be available using:

```js
hovertemplate: 'using color model: %{fullData.colormodel}'
```

src/traces/image/hover.js
```diff
@@ -0,0 +1,79 @@
...
+* LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+var Fx = require('../../components/fx');
+var Lib = require('../../lib');
+var constants = require('./constants');
+
+// var Axes = require('../../plots/cartesian/axes');
```
etpinard: :zap:

src/traces/image/attributes.js
```diff
@@ -0,0 +1,84 @@
...
+            'For example, for the `rgba` colormodel, the default value is [0, 0, 0, 0].'
+        ].join(' ')
+    },
+    zmax: {
+        valType: 'data_array',
+        role: 'info',
+        editType: 'plot',
+        description: [
+            'Array defining the higher bound for each color component.',
+            'For example, for the `rgba` colormodel, the default value is [255, 255, 255, 1].'
```
etpinard: It might be worth spelling out the default values for all the available color models in the `description`.

src/traces/image/hover.js
```diff
@@ -0,0 +1,79 @@
...
+    var c = trace._scaler(cd0.z[ny][nx]);
+    var s = constants.colormodel[colormodel].suffix;
+
+    var colorstring = [];
+    if(trace.hovertemplate || fmtColor) {
+        colorstring.push('[' + [c[0] + s[0], c[1] + s[1], c[2] + s[2]].join(', '));
+        if(dims === 4) colorstring.push(', ' + c[3] + s[3]);
+        colorstring.push(']');
+        colorstring = colorstring.join('');
+        pointData.extraText = '<span style="text-transform:uppercase">' + colormodel + '</span>: ' + colorstring;
```
etpinard: Let's get rid of this `<span>` and upper-case the `colormodel` string in JS using `.toUpperCase()`.

src/traces/image/hover.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    var py = ya.c2p(cd0.y0 + (ny + 0.5) * trace.dy);
+    var xVal = cd0.x0 + (nx + 0.5) * trace.dx;
+    var yVal = cd0.y0 + (ny + 0.5) * trace.dy;
```
etpinard: Does `image` work on categorical axes? If so, I suspect this part here will break.

src/traces/image/plot.js
```diff
@@ -0,0 +1,146 @@
...
+'use strict';
+var d3 = require('d3');
+var Lib = require('../../lib');
+var xmlnsNamespaces = require('../../constants/xmlns_namespaces');
+var constants = require('./constants');
+
+module.exports = {};
+
+// Generate a function to scale color components according to zmin/zmax and the colormodel
+module.exports.scaler = function(trace) {
```
etpinard: Hmm, do we need to export this function?

test/jasmine/tests/calcdata_test.js
```diff
@@ -886,7 +886,7 @@ describe('calculated data and points', function() {
             });
 
             // excludedTraces are traces that do not support sorting by value
-            var excludedTraces = [ 'carpet', 'contourcarpet'];
+            var excludedTraces = [ 'carpet', 'contourcarpet', 'image'];
```
etpinard: Can you turn this logic into a new trace-module category, just like

https://github.com/plotly/plotly.js/blob/e86c95b4b2abe646d7ab4e311fcd40cc61f1eaea/src/traces/image/index.js#L23

Naming it `'noSortingByValue'` would be good enough.

test/jasmine/tests/image_test.js
```diff
@@ -0,0 +1,381 @@
...
+            z: [[]]
+        };
+        supplyDefaults(traceIn, traceOut);
+        expect(traceOut.visible).toBe(false);
+
+        traceIn = {
+            z: [[], [], []]
+        };
+        supplyDefaults(traceIn, traceOut);
+        expect(traceOut.visible).toBe(false);
```
etpinard: Hmm. How does `visible` get set to false here during the defaults?

test/jasmine/tests/image_test.js
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        it('should display RGB channel values', function(done) {
+            var mockCopy = Lib.extendDeep({}, mock);
+            mockCopy.data[0].colormodel = 'rgb';
+            Plotly.newPlot(gd, mockCopy)
+            .then(function() {_hover(205, 125);})
+            .then(function() {
+                assertHoverLabelContent({
+                    nums: 'x: 25.5\ny: 14.5\nz: [54, 136, 153]\n<tspan style="text-transform:uppercase">rgb</tspan>: [54, 136, 153]',
```
etpinard: So you're showing the color by default. I thought we decided on just `x+y+z+text` in last week's meeting. Please correct me if I'm wrong.

src/traces/image/defaults.js
archmoj: That could be useful option with typed arrays.

src/traces/image/plot.js
```diff
@@ -0,0 +1,146 @@
...
+            c[k] = s[k](c[k]);
+        }
+        return c;
+    };
+};
+module.exports.plot = function(gd, plotinfo, cdimage, imageLayer) {
+    var xa = plotinfo.xaxis;
+    var ya = plotinfo.yaxis;
+
+    Lib.makeTraceGroups(imageLayer, cdimage, 'im').each(function(cd) {
```
etpinard: We'll need to fill in 

https://github.com/plotly/plotly.js/blob/6e9bafb8285f7247407a503e7dc33e44ff637f31/src/plots/cartesian/constants.js#L65-L74

with `'imagelayer'`. I'd say image traces should go below `heatmap` traces OR one position above. What do you think @antoinerg 

Moreover, we should add a mock of an image and heatmap trace on the same subplot.

etpinard:
Very nicely done @antoinerg :tada: - I think you found the perfect way to get this new trace type out fast with solid performance and not much brittle logic. Thank you very much for that!

I made a few comments - none of them should be difficult to resolve. I'll add one more comment here:

- I'd still vote for naming this trace `raster`. I'm not a fan of having a the term "image" refer to things in the layout AND the data.

etpinard:
Oh and one more thing @antoinerg: I didn't notice any funky CSS tricks like you had in one of your proof-of-concepts. I guess you decide to not go down that route, can you confirm?

src/traces/image/attributes.js
etpinard: ... moreover `x0` and `y0` should be `valType: 'any'` to support categorical axes.

src/traces/image/attributes.js
antoinerg: > What's `c` here? Is it the same value as for the `'color'` hoverinfo flag?

`c` is the color array. When using the `'color'` hoverinfo flag, we display the `c` array WITH the colormodel in front of it ie. `%{colormodel}: [%{c[0]}, %{c[1]}, %{c[2]}]`. Should I rename `c` to `color` for `hovertemplate`?

src/traces/image/calc.js
antoinerg: Yes, that's what this code assumes!

test/jasmine/tests/image_test.js
antoinerg: Yes, the code currently shows the color by default. You're right that this isn't what we discussed in the meeting. I'm OK with removing `color` but I also think it could be a default. It's your call :)

cc @nicolaskruchten @etpinard

src/traces/image/attributes.js
antoinerg: > Oh and one more thing @antoinerg: I didn't notice any funky CSS tricks like you had in one of your proof-of-concepts. I guess you decide to not go down that route, can you confirm?

I confirm that I :zap: the CSS trick I had in my proof of concept. If you want me to, I could add a few comments/references about CSS `image-rendering: pixelated` which could potentially be used in the future to speed up rendering on browsers that support it (almost all of them except Safari and potentially SVG editors like Illustrator/Inkscape).

antoinerg:
> Oh and one more thing @antoinerg: I didn't notice any funky CSS tricks like you had in one of your proof-of-concepts. I guess you decide to not go down that route, can you confirm?

I confirm that I :zap: the CSS trick I had in my proof of concept. If you want me to, I could add a few comments/references about CSS image-rendering: pixelated which could potentially be used in the future to speed up rendering on browsers that support it (almost all of them except Safari and potentially SVG editors like Illustrator/Inkscape).

I decided not to go that route because it's a bit unsafe and would require manually testing all the different browsers we want to support...

src/traces/image/attributes.js
antoinerg: As far as I know, this wasn't a requirement for that iteration so I did not include it. For your information, most other libraries do not support this either. It is usually assumed that an image is made of pixels of identical size arranged in a regular fashion as in a square crystal lattice. I'd be surprised if users ask for `x` and `y` but we'll see :smile:

test/jasmine/tests/calcdata_test.js
antoinerg: Very nice idea! Done in e7eeb38b53bb358222880329af43962fe92142e9

src/traces/image/attributes.js
nicolaskruchten: `x0` and `dx` seem enough for a v1... any obvious cases where that wouldn't be enough? @jonmmease @emmanuelle @alexcjohnson ?

src/traces/image/plot.js
antoinerg: > with `'imagelayer'`. I'd say image traces should go below `heatmap` traces OR one position above. What do you think @antoinerg

I don't have a strong preference but like you, I think I would lean to having `image` underneath. This is what I've done in 10c9914c5886de4eb7a9f32a80b747d389c6ce4e

lib/index.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/attributes.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3. Please let me know if it's :ok_hand:

src/traces/image/calc.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/defaults.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/hover.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/hover.js
antoinerg: Done in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/plot.js
antoinerg: Not in the current state of the code :sweat_smile:

I do not export it anymore in 73c0740f53fc82a70196d1993ba2824db2be85a3

src/traces/image/calc.js
etpinard: Ok, I guess it's fine to only render 2D arrays with inner arrays all of the same the same length, but how do jagged arrays like:

```
var trace = {
  type: 'image',
  z: [
    [[0,0,1], [0,0,1], [0,0,1]], 
    [[0,1,0], [0,1,0]], 
    [[1,0,0]]
}
```

get handled?

test/jasmine/tests/calcdata_test.js
etpinard: Thanks!!

src/traces/image/plot.js
etpinard: Very nicely done @antoinerg !

Thanks!!

src/traces/image/defaults.js
etpinard: Thanks!

src/traces/image/attributes.js
etpinard: Moreover, @antoinerg can you a few test cases checking what happens when `zmin` and `zmax` are partially set, for example:

```js
var trace = {
  zmin: [0, 1],
  zmax: [0]
}
```

antoinerg: In commit b998a17b249cfa93068e5ea719774d9357b45b1b, I removed the `dimensions` key and added tests to confirm that partially set `zmin` and `zmax` end up with a complete array!

antoinerg:
> Hmm. I'm not sure this `.concat` gets picked up properly by
> 
> https://github.com/plotly/plotly.js/blob/6e9bafb8285f7247407a503e7dc33e44ff637f31/tasks/compress_attributes.js#L14-L17
> 
> Can you double-check that the `description` field gets correctly removed in the `Plotly.PlotSchema.get()` output?

Nice catch @etpinard :male_detective: . You are right that it wasn't correctly removed in `Plotly.PlotSchema.get().traces.image.attributes.zmin.description`.

It is now fixed in b998a17b249cfa93068e5ea719774d9357b45b1b

src/traces/image/attributes.js
antoinerg: Very good idea! Done in https://github.com/plotly/plotly.js/pull/4289/commits/73c0740f53fc82a70196d1993ba2824db2be85a3 and subsequently fixed in b998a17b249cfa93068e5ea719774d9357b45b1b

src/traces/image/attributes.js
etpinard: `dimensions` is still showing for `zmin`:

https://github.com/plotly/plotly.js/blob/b998a17b249cfa93068e5ea719774d9357b45b1b/src/traces/image/attributes.js#L43

antoinerg: Oh my god, sorry about that :man_facepalming: 

Fixed in dec946902cdc69fbab792148730f77bbf9c0ade6

etpinard: No worries!! `zmin` and `zmax` are now looking :100:

src/traces/image/attributes.js
etpinard: Thanks!!

src/traces/image/calc.js
antoinerg: Right now, it will crash in a fire. Let's handle that properly. 

Whenever a pixel is undefined, we could
1. Draw nothing (which will show whatever is behind)
4. Draw pixel with a color set by the user via a new attribute
2. Draw pixel with all color component set to their minimum

Whenever we hover on an undefined pixel, we could:
1. Not show any hover label
2. Indicated this pixel is not defined
 
I'm tempted to go with 1 and 1. What do you think @etpinard ?

archmoj: I may vote for 1 and 2. cc: https://github.com/plotly/plotly.js/pull/4291
Then for `not show any hover` implementing `hoverongaps: false`.

antoinerg: I went with 1 and 1 in 15c06fa1814fa41b2c7a4a60d6d06cd6b90c4a5d but I'm willing to change it to whatever's best :smile_cat:

etpinard: https://github.com/plotly/plotly.js/commit/15c06fa1814fa41b2c7a4a60d6d06cd6b90c4a5d is :ok_hand: - thanks!

We could eventually add `hoverongaps:true` to image trace, but I think a default of `false` makes more sense for image traces.

emmanuelle:
Regarding categorical axes, I don't think we need them for images.

jackparmer:
> I'd still vote for naming this trace raster. 

FWIW, I also like `raster` for its precision. Besides the `layout.image` confusion, `image` makes me think of HTML images, where I can load up any ol' SVG or JPEG into my chart. I think we're begging for community.plot.ly confusion with `image` because it's not as unique or precise a name as `raster`. `raster` is also easier from a marketing perspective. Imagine us announcing "New chart type! Images!" It's going to be a head scratcher for many - definitely doesn't sound as novel, cool, or useful.

src/traces/image/hover.js
antoinerg: It doesn't at the moment. 

However, I added support to it in this [throwaway branch](https://github.com/plotly/plotly.js/compare/pr-image...antrg-test) along with improvements for log axis type. You can see it in action in this [Codepen](https://codepen.io/antoinerg/pen/BaaWjxe).

It shows that:
- `x0` and `y0` are working
- now if `dx` or `dy` is not equal to 1, we may get strange results similar to what `heatmap` currently does.

@etpinard is this good enough?

emmanuelle:
> > I'd still vote for naming this trace raster.
> 
> FWIW, I also like `raster` for its precision. Besides the `layout.image` confusion, `image` makes me think of HTML images, where I can load up any ol' SVG or JPEG into my chart. I think we're begging for community.plot.ly confusion with `image` because it's not as unique or precise a name as `raster`. `raster` is also easier from a marketing perspective. Imagine us announcing "New chart type! Images!" It's going to be a head scratcher for many - definitely doesn't sound as novel, cool, or useful.

I disagree here. What people expect is to have the equivalent of `imshow` in plotly. Practitioners dealing with images (camera pictures, microscopy, etc) call them images and never raster. I completely understand that there is a risk of confusion with `layout.image`, but I think `image` is better than `raster` because people are more familiar with it.

emmanuelle:
I think we will add an `imshow` function somewhere in the Python API anyway. If we decide that we don't want `Image` I'd rather have `Rasterimage` than just `Raster`.

emmanuelle:
Regarding the confusion we can make an new image tutorial where we make sure to explain that `layout.image` is something different

jackparmer:
👍Makes sense, I'm back on the `image` train now!

archmoj:
What about naming it `pixmap`?
https://en.wikipedia.org/wiki/Bitmap

YoniChechik:
> What about naming it `pixmap`?
> https://en.wikipedia.org/wiki/Bitmap

why don't it just be `imshow` like anywhere else?

src/traces/image/calc.js
etpinard: Oh one more thing @antoinerg, instead of

```js
var w = trace.z[0].length;
```

you should use

```js
var maxRowLength = require('../../lib').maxRowLength;
// ...
var w = maxRowLength(trace.z);
```

etpinard: ... to handle cases where the max row length isn't in `z[0]`.

src/traces/image/hover.js
etpinard: The behaviour looks good @antoinerg -- thanks!

One tip: use `d2c` in `calc.js`, for example:

https://github.com/plotly/plotly.js/blob/82e33e2491192eacab1db026af7b2b98cc6e14ab/src/traces/image/calc.js#L17-L20

should just be:

```js
x0 = xa.d2c(trace.x0) - trace.dx / 2;
y0 = ya.d2c(trace.y0) - trace.dy / 2;
```

where `d2c` for linear and log axes is simply an is-numeric check.

src/traces/image/calc.js
antoinerg: Done in 2d393ddab83a5492afaf8a2c2690ddd7da3cb15f

etpinard: Great. Resolved!

src/traces/image/hover.js
antoinerg: :ok_hand: done in fa04037cdad8a130068f08f0af417915b0147015

src/traces/image/attributes.js
antoinerg: ... well I set `valType: 'any'`. I didn't change the `editType` however. Let me check if it's really necessary.

antoinerg: It was needed indeed. Thanks @etpinard :1st_place_medal: 

Fixed in https://github.com/plotly/plotly.js/pull/4289/commits/0eb6604ab3470c28ee866d4cb2f90087ce81d7db

emmanuelle:
@YoniChechik `imshow` is a function name (the name of an action, show the image). For the name of the trace, we're looking for a good name for the object to be plotted. But it's likely that we'll also have an `imshow` function in the python API.

emmanuelle:
@archmoj `pixmap` is not used by anybody else (I think). We should stick with names which are familiar to users.

src/traces/image/attributes.js
etpinard: Ok - can you add a test to lock it down? thanks.

etpinard:
I'll go ahead and hide all the image vs raster comments. They make it hard for devs and reviewers to focus on the remaining items to implement.

We'll choose the definitive trace name during tomorrow morning's meeting. Thank you.

src/traces/image/attributes.js
etpinard: FWIW I'm also very much ok with not implementing `x` and `y` in this iteration.

src/traces/image/attributes.js
antoinerg: Of course :smile_cat: I added one in e2a9c6e77a425c96d2878168c204427d7460d797

etpinard: Thanks!!

test/jasmine/tests/image_test.js
etpinard: Having `x+y+z+color+all` as default is pretty nice when one sets `zmin` and `zmax`, but otherwise the `color` flag is essentially that same as `z` flag.

So, I'm thinking we should either:

1) set the `hoverinfo` default to `x+y+z+color+all` when `zmin` and/or `zmax` is set, fallback to `x+y+z+all` otherwise
2) consider that `zmin` and `zmax` won't be set very often by users and then `x+y+z+all` is a fine "hard" default for `hoverinfo`.

---

I'd personally vote for 2)

src/traces/image/attributes.js
etpinard: So currently,

```js
Plotly.newPlot(gd, [{
  type: 'image',
  z: [[ [255,0,0] ]],
  hoverinfo: 'color'
}])

// => shows `RBG: [255, 0, 0]` on hover
```

But

```js
Plotly.newPlot(gd, [{
  type: 'image',
  z: [[ [255,0,0] ]],
  hovertemplate: '%{color}'

// => shows `%{color}` 
}])
```

... which I'm not a fan of. I think we should try to make `hoverinfo` flags and `hovertemplate` token match. So yeah, I think renaming `c` to `color` is the way to go.

---

Moreover, I really like that `%{c[0]}`, `%{c[1]}` and friends (soon to be renamed `color[0]`, `color[1]`, ...) work, but it would nice to make

```js
Plotly.newPlot(gd, [{
  type: 'image',
  z: [[ [255,0,0] ]],
  hovertemplate: '%{c[3]}'

// => currently shows `%{c[3]}` 
}])
```

show `1` instead, even when the alpha channel isn't set.

antoinerg: I renamed `c` to `color` in a98ac44ae81685920979202770200a1e2a05fcba.

Let me think about what to do for `%{color[3]}`. I'm tempted to leave it as is because, in the context of `rgb` or `hsl`, the 4th channel is undefined and shouldn't exist.

etpinard: > Let me think about what to do for `%{color[3]}`. I'm tempted to leave it as is because, in the context of `rgb` or `hsl`, the 4th channel is undefined and shouldn't exist.

That's fair. I think we should show _something_ for `%{color[3]}`, rendering it as just `%{color[3]}` makes me think that the token is invalid, which isn't really true. In my mind, it's the channel that's undefined. So I'd vote for showing any of: `null`,`false`, `''`, `1` or `undefined`. 

But this is a very tiny detail, so feel free to simply add a `TODO` in the code about it and move on to more important things.

test/jasmine/tests/image_test.js
antoinerg: Reading your comment, I realized that the trace lacked `hovertext` and `text` attributes. I added both in fd51e206f7bc3cdc6b4c6b13559c7a6535c771a3

antoinerg:
> I don't we need that, this `appendArrayPointValue` call
> 
> https://github.com/plotly/plotly.js/blob/07ef0008795c9dccf9d5c5e917697babb243df86/src/components/fx/helpers.js#L154
> 
> should automatically include text in the image event data.

Nice catch :male_detective: ! It is :hocho:  in 5d7f9a42ea1a44d9d979176f436dc5beebc392d8

src/traces/image/attributes.js
antoinerg: This one is actually tricky to implement. I would need to set `color[3]` to `1` but then it would show up in hover template `%{color}` which we don't want. The template string routine does not allow overriding the result of accessing an individual array element without ALSO overriding the result when accessing the whole array...

test/jasmine/tests/image_test.js
antoinerg: Ok, so I went with your suggestion number 2.

The change is introduced in e4cac27013813e94a4439809d2de1c53e90c0f95 along with a test!

src/traces/image/attributes.js
antoinerg: I added a `TODO` in the code in commit https://github.com/plotly/plotly.js/commit/e4cac27013813e94a4439809d2de1c53e90c0f95

etpinard: Thanks! Resolving!

test/jasmine/tests/image_test.js
etpinard: Awesome :tada:

etpinard:
Great, all my comments have been resolved. Awesome work @antoinerg !!

PSA: we're picking `image` (not `raster`, not `pixmap`, not anything else). Turns out I was the only one that didn't like `image`.

:dancer: :dancer: :dancer:

antoinerg:
> PSA: we're picking `image` (not `raster`, not `pixmap`, not anything else). Turns out I was the only one that didn't like `image`.

I am with you @etpinard: I'm not a big fan of `image` either and would have preferred `pixmap`. It seems like `image` is more popular so let's give the people what they want!

archmoj:
> PSA: we're picking `image` (not `raster`, not `pixmap`, not anything else). Turns out I was the only one that didn't like `image`.

Me too!

pfeatherstone:
Fantastic stuff. Any plans of updating the docs? At the moment i'm using the demo for guidance

PR #4292, opened 2019-10-21
Keep orthographic zoom scales in fullLayout

### LINKED ISSUE #4274: When mouse-zooming with 3d orthographic projection, no exposed property is updated
See the codepen https://codepen.io/emmanuelle-plotly/pen/QWWyjxg?editors=0011 and try to zoom using the mouse wheel. One would expect one property of the figure to be updated, such as `scene.camera.eye` or `scene.aspectratio`, but none of them changes. 

In `perspective` projection, the `scene.camera.eye` is updated when wheel-zooming. 

At the moment it is not possible to dump to JSON a figure which has been zoomed and then to reproduce it because the information about the zoom is not included.

archmoj:
Fix #4274.
Orthographic scroll zoom changes `aspectratio` values. These changes should be reflected in `fullLayout` and also `relayout` should be emitted.

[Before](https://codepen.io/emmanuelle-plotly/pen/QWWyjxg?editors=0011)
[After](https://codepen.io/MojtabaSamimi/pen/dyyNNWO?editors=0011)

In addition, there was a mistake noticed in calling arguments i.e. when try recovering the context which was fixed in https://github.com/plotly/plotly.js/pull/4292/commits/d486dcd56560feca1c8e486bffee31e822f8fda4.

Also to mention commit https://github.com/plotly/plotly.js/pull/4292/commits/496059ca3fe1c9baf4a6d690be7987c5cf5cb39a simplifies calling arguments by applying scene `pixelRatio`.

@plotly/plotly_js

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.1.1",
     "gl-plot2d": "^1.4.2",
-    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#4d41ca21b3c0a25db8776ce92261995323bb62e8",
+    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#83e9c9406976e5ee92bf1f1851e4366e516c06f7",
     "gl-pointcloud2d": "^1.0.2",
     "gl-scatter3d": "^1.2.2",
     "gl-select-box": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.1.1",
     "gl-plot2d": "^1.4.2",
-    "gl-plot3d": "^2.2.2",
+    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#4d41ca21b3c0a25db8776ce92261995323bb62e8",
     "gl-pointcloud2d": "^1.0.2",
     "gl-scatter3d": "^1.2.2",
     "gl-select-box": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.1.1",
     "gl-plot2d": "^1.4.2",
-    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#83e9c9406976e5ee92bf1f1851e4366e516c06f7",
+    "gl-plot3d": "^2.3.0",
     "gl-pointcloud2d": "^1.0.2",
     "gl-scatter3d": "^1.2.2",
     "gl-select-box": "^1.0.3",
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -9,9 +9,10 @@
 
 'use strict';
 
-var createCamera = require('gl-plot3d').createCamera;
-var createPlot = require('gl-plot3d').createScene;
-var mouseWheel = require('mouse-wheel');
+var glPlot3d = require('gl-plot3d');
+var createCamera = glPlot3d.createCamera;
+var createPlot = glPlot3d.createScene;
+
 var getContext = require('webgl-context');
 var passiveSupported = require('has-passive-events');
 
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.1.1",
     "gl-plot2d": "^1.4.2",
-    "gl-plot3d": "^2.2.2",
+    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#4d41ca21b3c0a25db8776ce92261995323bb62e8",
```
etpinard: Have you made a PR for this patch?

src/plots/gl3d/scene.js
```diff
@@ -243,7 +244,26 @@ function tryCreatePlot(scene, cameraObject, pixelRatio, canvas, gl) {
         }
     }
 
-    return failed < 2;
+    if(failed < 2) {
+        scene.wheelListener = mouseWheel(scene.glplot.canvas, function(dx, dy) {
```
etpinard: A couple questions:

- why can't we add logic to this block:

https://github.com/plotly/plotly.js/blob/4a649f01f21fbf3ec814242a0b1f93ca9d2f8067/src/plots/gl3d/scene.js#L296-L300

which looks like also adds a mouse-wheel handler to the scene

- Shouldn't we also emit a `plotly_relayout` event with the `aspectratio` keys just like we do here:

https://github.com/plotly/plotly.js/blob/4a649f01f21fbf3ec814242a0b1f93ca9d2f8067/src/plots/gl3d/scene.js#L283-L290

for perspective projections?

archmoj:
@etpinard thanks very much for the review.
Setting the status to `in progress`...

src/plots/gl3d/scene.js
archmoj: Addressed in https://github.com/plotly/plotly.js/pull/4292/commits/fe19e47430efc9ebab484ce5113bd7841b62a5cd

package.json
archmoj: https://github.com/gl-vis/gl-plot3d/pull/24
@etpinard Thanks!

src/plots/gl3d/scene.js
```diff
@@ -260,21 +262,45 @@ function initializeGLPlot(scene, pixelRatio, canvas, gl) {
...
     var gd = scene.graphDiv;
 
+    var makeUpdate = function() {
+        var update = {};
+
+        // camera updates
+        update[scene.id + '.camera'] = getLayoutCamera(scene.camera);
+
+        // scene updates
+        update[scene.id + '.aspectratio'] = scene.glplot.getAspectratio();
```
etpinard: This should **only** part of the event data on scene with orthographic projections

etpinard: ... in other words, the event data should have the minimal set of keys that changed during the (here: scroll) interaction.

src/plots/gl3d/scene.js
```diff
@@ -839,15 +872,33 @@ proto.saveCamera = function saveCamera(layout) {
...
+            ) {
+                hasChanged = true;
+            }
+        }
+    }
+
     if(hasChanged) {
         var preGUI = {};
         preGUI[this.id + '.camera'] = cameraDataLastSave;
+        preGUI[this.id + '.aspectratio'] = aspectDataLastSave;
```
etpinard: Here too, we should only store the aspectratio change for orthographic projections.

archmoj: Done in https://github.com/plotly/plotly.js/pull/4292/commits/b92230a0b99f883dd009e89890afc7555450a7a2.

src/plots/gl3d/scene.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4292/commits/b92230a0b99f883dd009e89890afc7555450a7a2.

archmoj: And in https://github.com/plotly/plotly.js/pull/4292/commits/d825c0e18de420cb15cd30747f366e394729bff7.

archmoj:
@etpinard This PR should be ready for another review.
Thanks!

src/plots/gl3d/scene.js
```diff
@@ -835,19 +868,75 @@ proto.saveCamera = function saveCamera(layout) {
...
+            cameraFullNP.set(cameraData);
+        }
+
+        if(aspectChanged) {
+            aspectNestedProp.set(aspectData);
+
+            var aspectFullNP = Lib.nestedProperty(fullLayout, this.id + '.aspectratio');
+            aspectFullNP.set(aspectData);
+
+            this.glplot.redraw();
```
etpinard: Hmm, why do we need to redraw here?

test/jasmine/tests/gl3d_plot_interact_test.js
```diff
@@ -903,16 +1164,117 @@ describe('Test gl3d drag and wheel interactions', function() {
...
+                buttons: 1,
+                noCover: true
+            });
+        })
+        .then(function() {
+            expect(events.length).toEqual(nsteps);
+            expect(relayoutCnt).toEqual(1);
+            Object.keys(relayoutEvent).sort().forEach(function(key) {
+                expect(Object.keys(events[0])).toContain(key);
+            });
```
etpinard: Can you also check that this set of relayout event keys does not include `scene.aspectratio`?

src/plots/gl3d/scene.js
archmoj: We used to redraw in scene wheel listener after updating `gl-plot3d.aspect` which is now moved to here so that we record the new before update.

etpinard: I guess that's ok, it's just feels a little odd to me that a method called `saveLayout` ends up redrawing things in some cases.

Non-blocking.

archmoj: We could possibly make `gl3d/scene.js` file more readable at one point.  
I keep this in mind when we have time to refactor it in another PR.

etpinard: > I keep this in mind when we have time to refactor it in another PR.

Awesome. Thank you very much cleaning up this code!!

test/jasmine/tests/gl3d_plot_interact_test.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4292/commits/3b28faef2ddd72d1ce6f6d2812f5b1f01917075f. I also added few more tests.

etpinard:
Nicely done @archmoj - :dancer:

PR #4304, opened 2019-10-25
do not accept domain ranges where end is not greater than start

### LINKED ISSUE #4300: pie - problem with reversed domain ranges
In this [codepen](https://codepen.io/MojtabaSamimi/pen/LYYyzwa?editors=0010) the first `pie` graph (i.e. on the left) does not look right! 
The graphs reversed and the text is presented `outside` as a result of bad input data where `domain.x`[1] is not greater than `domain.x`[0].

We shouldn't allow that here: https://github.com/plotly/plotly.js/blob/99fc02f50cb7e10a616b1c4928e5f7e16809c57c/src/plots/domain.js#L111-L132

archmoj:
Fix #4300

[Before](https://codepen.io/MojtabaSamimi/pen/LYYyzwa?editors=0010)
[After](https://codepen.io/MojtabaSamimi/pen/zYYzGOo?editors=0010)

@plotly/plotly_js

### DIFF
--- a/src/plots/domain.js
+++ b/src/plots/domain.js
@@ -131,6 +131,6 @@ exports.defaults = function(containerOut, layout, coerce, dfltDomains) {
     var y = coerce('domain.y', dfltY);
 
     // don't accept bad input data
-    if(!(x[0] < x[1])) containerOut.domain.x = dfltX;
-    if(!(y[0] < y[1])) containerOut.domain.y = dfltY;
+    if(!(x[0] < x[1])) containerOut.domain.x = dfltX.slice();
+    if(!(y[0] < y[1])) containerOut.domain.y = dfltY.slice();
 };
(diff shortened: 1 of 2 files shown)


src/plots/domain.js
```diff
@@ -127,6 +127,10 @@ exports.defaults = function(containerOut, layout, coerce, dfltDomains) {
         }
     }
 
-    coerce('domain.x', dfltX);
-    coerce('domain.y', dfltY);
+    var x = coerce('domain.x', dfltX);
+    var y = coerce('domain.y', dfltY);
+
+    // don't accept bad input data
+    if(!(x[0] < x[1])) containerOut.domain.x = dfltX;
```
etpinard: I would `dfltX.slice()` here. I'm sure it ever happens for traces calling `plots/domain.js`, but we do mutate the domain sometimes (e.g. when applying `scaleanchor` constraints), so I would keep things on the safe side.

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4304/commits/74773fb1c28104c896873bf19022d6a871f0b24e

etpinard:
Thanks very much :dancer:

PR #4306, opened 2019-10-25
Handle data with identical positions in cone plots

### LINKED ISSUE #4072: Cone trace doesn't show any if double-up
I found that if I have two cones at the same location, then no cones show up, and no error appears.

This also happens if they are just in close proximity, like 0.01 off

Reproducible example: https://codepen.io/anon/pen/KOVwYd?editors=0010

archmoj:
Fix #4072 using `gl-cone3d` 1.5.1 patch release.

[Before](https://codepen.io/MojtabaSamimi/pen/ZEEyQmY?editors=0010)
[After](https://codepen.io/MojtabaSamimi/pen/vYYZLQp?editors=0010)

@etpinard

### DIFF
--- a/test/jasmine/tests/cone_test.js
+++ b/test/jasmine/tests/cone_test.js
@@ -177,6 +177,39 @@ describe('Test cone autorange:', function() {
     });
 });
 
+describe('Test cone autorange:', function() {
+    var gd;
+
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+
+    afterEach(function() {
+        Plotly.purge(gd);
+        destroyGraphDiv();
+    });
+
+    it('@gl should skip identical positions in calculating cone vectorScale', function(done) {
+        Plotly.plot(gd, {
+            data: [
+                {
+                    type: 'cone',
+                    x: [-1, -1, -3, -4],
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/cone_test.js
```diff
@@ -177,6 +177,39 @@ describe('Test cone autorange:', function() {
...
+                    x: [-1, -1, -3, -4],
+                    y: [1, 1, 3, 4],
+                    z: [1, 1, 3, 4],
+                    u: [-1, 2, -3, -4],
+                    v: [1, -2, 3, 4],
+                    w: [1, -2, 3, 4]
+                }
+            ]
+        }).then(function() {
+            expect(gd._fullLayout.scene._scene.glplot.objects[0].vectorScale).toBeCloseTo(0.2857, 4);
```
etpinard: ... and I guess `vectorScale` used to be `0` here?

archmoj: Exactly!

etpinard: Thanks! Brilliant test and brilliant fix!

etpinard:
:dancer:

PR #4307, opened 2019-10-25
image: improve layout defaults

antoinerg:
This PR adds a smart layout default of `scaleanchor` to make sure pixels are squares. 

It also fixes the following issue: before this PR, having an `image` trace would always reverse the y-axis. Now it only does so if `image` is the only trace type present in the plot.

Both smart defaults are now tested.


src/plots/cartesian/constraints.js
```diff
@@ -53,14 +55,14 @@ exports.handleConstraintDefaults = function(containerIn, containerOut, coerce, a
     // 'matches' wins over 'scaleanchor' (for now)
     var scaleanchor, scaleOpts;
 
-    if(!matches && containerIn.scaleanchor && !(containerOut.fixedrange && constrain !== 'domain')) {
+    if(!matches && (containerIn.scaleanchor || opts.scaleanchorDflt) && !(containerOut.fixedrange && constrain !== 'domain')) {
```
archmoj: This line is getting too long.
It would be nice to put conditions on multiple lines.

src/plots/cartesian/constraints.js
```diff
@@ -53,14 +55,14 @@ exports.handleConstraintDefaults = function(containerIn, containerOut, coerce, a
...
         scaleOpts = getConstraintOpts(constraintGroups, thisID, allAxisIds, layoutOut, constrain);
         scaleanchor = Lib.coerce(containerIn, containerOut, {
             scaleanchor: {
                 valType: 'enumerated',
                 values: scaleOpts.linkableAxes || []
             }
-        }, 'scaleanchor');
+        }, 'scaleanchor', opts.scaleanchorDflt);
```
archmoj: To have a more clear interface with input arguments, would you mind setting
```js
var scaleanchorDflt = opts.scaleanchorDflt
```
on top of the function and then use all those optional attributes within the function?
It can also help reduce the chance of typos that may silently happen in the object properties. 
Thanks!

src/plots/cartesian/layout_defaults.js
```diff
@@ -74,24 +76,24 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
                 if(xaName) xaMayHide[xaName] = true;
-                if(yaName) yaMayBackward[yaName] = true;
+                if(yaName) yaMayReverse[yaName] = true;
             } else {
                 if(yaName) yaMayHide[yaName] = true;
             }
+            yaMustNotScaleanchor[yaName] = true;
+        } else if(trace.type === 'image') {
+            if(yaName) yaMayReverse[yaName] = true;
+            if(yaName) yaMayScaleanchor[yaName] = true;
```
archmoj: We can combine these two `if(yaName)` conditions.

src/plots/cartesian/layout_defaults.js
```diff
@@ -299,7 +301,15 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
         axLayoutIn = layoutIn[axName];
         axLayoutOut = layoutOut[axName];
 
-        handleConstraintDefaults(axLayoutIn, axLayoutOut, coerce, allAxisIds, layoutOut);
+        var scaleanchorDflt = null;
+        if(axLetter === 'y' && !axLayoutIn.hasOwnProperty('scaleanchor') && !yaMustNotScaleanchor[axName] && yaMayScaleanchor[axName]) {
```
archmoj: This line is also rather long. So it would be nice to have it on multiple lines.

test/image/mocks/image_adventurer.json
```diff
@@ -1 +1 @@
...
..., 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]], [[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]], "x0": 0.5, "y0": 0.5, "colormodel": "rgba", "type": "image"}], "layout": {"yaxis": {"dtick": 1, "range": [40, 7], "zeroline": false}, "xaxis": {"dtick": 1, "zeroline": false}, "width": 400, "height": 400, "margin": {"t": 0, "b": 0, "l": 0, "r": 0}}}
```
archmoj: Curious to know how these numbers for the mock are achieved?
Did you read the raw data from a raw bitmap file (i.e. `.bmp`) using cpp? 
Or decoded a `.png` with javascript or python? Or...

test/image/mocks/image_zmin_zmax.json
```diff
@@ -60,6 +60,14 @@
...
     "width": 400,
     "height": 800,
+    "yaxis": {"scaleanchor": null},
+    "yaxis2": {"scaleanchor": null},
+    "yaxis3": {"scaleanchor": null},
+    "yaxis4": {"scaleanchor": null},
+    "yaxis5": {"scaleanchor": null},
+    "yaxis6": {"scaleanchor": null},
+    "yaxis7": {"scaleanchor": null},
+    "yaxis8": {"scaleanchor": null},
```
archmoj: Can we use `false` instead of `null`?

archmoj: Or even better `true` instead of `null`?

test/image/mocks/image_adventurer.json
etpinard: @antoinerg what did you change in this JSON file?

test/image/mocks/image_axis_type.json
```diff
@@ -28,6 +28,7 @@
       "width": 400, "height": 600, "title": {"text": "Image on categorical and log axes"},
       "grid": {"rows": 2, "columns": 1, "pattern": "independent"},
       "xaxis2": {"type": "log"},
-      "yaxis2": {"type": "log"}
+      "yaxis": {"scaleanchor": null},
+      "yaxis2": {"type": "log", "scaleanchor": null}
```
etpinard: `null` isn't a valid `scaleanchor` value. Please remove all the `scalanchor: 'null'` lines from the mocks. If you want to test that `scaleanchor: null` does the right thing, use a jasmine test.

test/image/mocks/image_zmin_zmax.json
etpinard: We should :hocho: all these lines instead.

test/jasmine/tests/image_test.js
```diff
@@ -444,7 +500,7 @@ describe('image hover:', function() {
                 zmax: [1, 1, 1],
                 text: [['A', 'B', 'C'], ['D', 'E', 'F']],
                 hovertemplate: '%{text}<extra></extra>'
-            }], layout: {width: 400, height: 400}};
+            }], layout: {width: 400, height: 400, yaxis: {scaleanchor: null}}};
```
etpinard: Setting `scaleanchor: 'null'` is fine here (adapting the pixel coordinates to hover on is annoying :smirk: ), but can you add a comment above this explain why `scaleanchor` is set to `null`? Thanks!

test/jasmine/tests/image_test.js
```diff
@@ -100,6 +100,62 @@ describe('image supplyDefaults', function() {
...
+describe('image smart layout defaults', function() {
+    var gd;
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+
+    afterEach(destroyGraphDiv);
+
+    it('should reverse yaxis if only images are present', function(done) {
+        Plotly.newPlot(gd, [{type: 'image', z: [[[255, 0, 0]]]}])
```
etpinard: Non-blocking. Using `newPlot` is fine, but I would prefer making these pure `supplyDefaults` tests for speed.

test/image/mocks/image_axis_type.json
etpinard: ... this mean you'll have to update a few baselines, I hope you don't mind.

src/plots/cartesian/layout_defaults.js
antoinerg: That's 4 lines instead of 2 but I'll make the change :)

test/image/mocks/image_adventurer.json
antoinerg: @etpinard I removed the `scaleanchor` that was initially present since it is now the default. This way, this mock also tests the smart default!

antoinerg: > Curious to know how these numbers for the mock are achieved?
> Did you read the raw data from a raw bitmap file (i.e. `.bmp`) using cpp?
> Or decoded a `.png` with javascript or python? Or...

@archmoj I read the a `png` using the `imread`  function in Python:
```python
#! /usr/bin/env nix-shell
#! nix-shell -i python3 -p python3 python3Packages.scikitimage -p python3 python3Packages.numpy
import json
from skimage.io import imread
import numpy as np
# rgb_img = imread('small.jpg')

rgb_img = imread('/home/antoine/Downloads/Individual
Sprites/adventurer-idle-2-00.png')
# rgb_img[:, :, 3] = rgb_img[:, :, 3] / 255
figure = {
    "data": [{
        "z": rgb_img.tolist(),
        "x": 0.5,
        "y": 0.5,
        "colormodel": "rgba",
        "type": 'image'
    }],
    "layout": {
        "yaxis": {
            "scaleanchor": "x",
            "dtick": 1,
            "range": [-40, -5],
            "zeroline": False
        },
        "xaxis": {
            "dtick": 1,
            "zeroline": False
        },
        "width": 400,
        "height": 400,
        "margin": {"t": 0, "b": 0, "l": 0, "r": 0}
    }
}
json = json.dumps(figure)
print(json)
```

antoinerg:
Sorry about that:
Screenshot_2019-10-28_11-31-59
but I pushed a bad commit inadvertently. I didn't change any commits you've seen already.

src/plots/cartesian/constraints.js
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

src/plots/cartesian/constraints.js
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

src/plots/cartesian/layout_defaults.js
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

src/plots/cartesian/layout_defaults.js
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

test/image/mocks/image_zmin_zmax.json
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6 for all the mocks

test/image/mocks/image_axis_type.json
antoinerg: Done in d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

test/jasmine/tests/image_test.js
antoinerg: I actually went ahead and removed the `scaleanchor`. Guessing the new hover position wasn't so hard :smile_cat: 

See d71c3b345d2eecb7b3087c383f00f5f102b9c2f6

etpinard: Even better. Thanks!!

test/jasmine/tests/image_test.js
antoinerg: Ok so I tried to transform a test from using `newPlot` to `supplyAllDefaults`. However, it fails since the range doesn't seem to get set. Any idea how to solve this?
```javascript
    var supplyAllDefaults = require('../assets/supply_defaults');
    fit('should reverse yaxis if only images are present', function() {
        gd = {};
        gd.data = [{type: 'image', z: [[[255, 0, 0]]]}];
        gd.layout = {};
        supplyAllDefaults(gd);
        expect(gd._fullLayout.yaxis.range[0]).toBeGreaterThan(gd._fullLayout.yaxis.range[1]);
    });
```

antoinerg: Never mind my last comment. I will check instead for `autorange === 'reversed'`. Commit underway

test/image/mocks/image_axis_type.json
etpinard: Ok, thanks to the new baselines I realised that the `constrain` default is `'range'`.

If I recall correctly, both @emmanuelle and @nicolaskruchten prefered the look of `constrain: 'domain'` for image traces.

Personally, I think `constrain: 'range'` looks just fine and it's probably not worth adding more smart (aka magic) defaults in this case. 

Would it be ok to keep `constrain: 'range'` as default here?

src/plots/cartesian/layout_defaults.js
etpinard: Non-blocking (to me :smile: ) but I think you forgot to commit this line.

test/jasmine/tests/image_test.js
etpinard: Exactly, `supplyDefaults` won't give you the `range`, but assert `autorange` is 100% sufficient here.

antoinerg: Done in 97a8d246f0025ae12204781d68582fc7235ff106

src/plots/cartesian/layout_defaults.js
antoinerg: Nice catch! Done in 2053c583d6ca1cc2dffe4632738dd120abf67988

test/image/mocks/image_axis_type.json
antoinerg: Mmm :thinking: I'm not sure. I think `constrain: 'domain'` looks a bit better when a single `image` trace is present. I don't know how I feel about adding some more magic defaults. I'll let @emmanuelle and @nicolaskruchten chime in.

About the smart defaults in `image`, where should we document them? I guess within the trace description in its `index.js` file?

etpinard: > About the smart defaults in `image`, where should we document them? I guess within the trace description in its `index.js` file?

Yep, in the `image/index.js` description would be :ok_hand:

etpinard:
:dancer: for me - thanks very much @antoinerg

antoinerg:
>  for me - thanks very much @antoinerg

Awesome! I'll wait a bit before merging this in case we want a smart layout default of `constrain: 'domain'`.

BTW, it is implemented in this [branch](https://github.com/plotly/plotly.js/compare/image-scaleanchor...image-constrain). It's not super DRY but it is tested.

emmanuelle:
> > for me - thanks very much @antoinerg
> 
> Awesome! I'll wait a bit before merging this in case we want a smart layout default of `constrain: 'domain'`.
> 
> BTW, it is implemented in this [branch](https://github.com/plotly/plotly.js/compare/image-scaleanchor...image-constrain). It's not super DRY but it is tested.

Yes, `constrain: 'domain'` would look nicer (no background visible around the image)

antoinerg:
> Yes, `constrain: 'domain'` would look nicer (no background visible around the image)

Ok, I opened a PR against the current one: https://github.com/plotly/plotly.js/pull/4313

@etpinard can you review it?

etpinard:
> Yes, `constrain: 'domain'` would look nicer (no background visible around the image)

Ok maybe, but I'd argue it's much more magical. 

@archmoj what do you think?

archmoj:
> > Yes, `constrain: 'domain'` would look nicer (no background visible around the image)
> 
> Ok maybe, but I'd argue it's much more magical.
> 
> @archmoj what do you think?

I think the `image` trace is getting too magical when compared to other `Cartesian` traces.

antoinerg:
> I think the `image` trace is getting too magical when compared to other `Cartesian` traces.

I understand and share your sentiment. Now the question is where should that logic be? If it lives in plotly.py, then we'll also have to include it in our other wrappers (R and others) (ie. the logic will have to be duplicated downstream which is maybe not ideal).

Considering we already decided to automagically set `scaleanchor` and `autorange`, adding `constrain` to the list would be consistent with our current philosophy/efforts of providing better layout defaults for `image`. Now the real question is whether `constrain: 'domain'` really is better. If it is, I would be tempted to consider including https://github.com/plotly/plotly.js/compare/image-scaleanchor...image-constrain

cc @nicolaskruchten @alexcjohnson @emmanuelle @archmoj

nicolaskruchten:
I’m on the domain train.

alexcjohnson:
To my mind there's not a whole lot of difference in magicality between turning on `scaleanchor` and that plus `domain`. If that's the right default user experience, let's do it. Honestly if we had implemented `domain` in the first iteration of constraints it would probably be the default anyway.

(I haven't looked at that other branch @antoinerg but I hope `domain` is set on both axes...)

etpinard:
Ooof, I think I need a vacation.

archmoj:
> * but we'll change the on/off logic: **any** subplot that has **one or more** `image` trace should get a `scaleanchor` and a domain constraint (regardless of what other trace types there are on that subplot).

And what about `image` reversing the Y-axis? It would make sense to keep the axis reversed when other traces are added on top.

etpinard:
> And what about `image` reversing the Y-axis? I

Good call! Yes, that too. Every `image`-induced axis default should be turned on on every subplot containing one or more `image` trace.

nicolaskruchten:
That all sounds great to me.

antoinerg:
I'm glad we reached a consensus! I'd like to thank everybody for their valuable input.

@etpinard I went ahead and made the changes described in https://github.com/plotly/plotly.js/pull/4307#issuecomment-547442754 in PR https://github.com/plotly/plotly.js/pull/4313 which we would merge in the current PR. Let me know if that is :ok_hand:

test/jasmine/tests/image_test.js
```diff
@@ -100,6 +101,82 @@ describe('image supplyDefaults', function() {
...
 
+describe('image smart layout defaults', function() {
+    var gd;
+    beforeEach(function() {
+        gd = createGraphDiv();
+    });
+
+    afterEach(destroyGraphDiv);
+
+    it('should reverse yaxis if only images are present', function() {
```
etpinard: should say `should reverse yaxis if images are present` i.e. no `only`

test/jasmine/tests/image_test.js
```diff
@@ -100,6 +101,82 @@ describe('image supplyDefaults', function() {
...
+
+    it('should NOT reverse yaxis if it\'s already defined', function() {
+        gd = {};
+        gd.data = [{type: 'image', z: [[[255, 0, 0]]]}];
+        gd.layout = {yaxis: {autorange: false}};
+        supplyAllDefaults(gd);
+        expect(gd._fullLayout.yaxis.autorange).toBe(false);
+    });
+
+    it('should set scaleanchor to make square pixels if only images are present', function() {
```
etpinard: similarly here, :hocho: that `only`

test/jasmine/tests/image_test.js
```diff
@@ -100,6 +101,82 @@ describe('image supplyDefaults', function() {
...
+    });
+
+    it('should NOT set scaleanchor if it\'s already defined', function() {
+        gd.data = [{type: 'image', z: [[[255, 0, 0]]]}];
+        gd.layout = {yaxis: {scaleanchor: 'x3'}};
+        supplyAllDefaults(gd);
+        expect(gd._fullLayout.yaxis.scaleanchor).toBe(undefined);
+    });
+
+    it('should constrain axes to domain if only images are present', function() {
```
etpinard: :hocho:  `only`

test/jasmine/tests/image_test.js
antoinerg: Should I :hocho: the whole test instead?

test/jasmine/tests/image_test.js
antoinerg: Again, should I :hocho: the whole test?

src/plots/cartesian/layout_defaults.js
```diff
@@ -39,8 +39,9 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
     var yaMayHide = {};
     var xaMustDisplay = {};
     var yaMustDisplay = {};
-    var yaMustForward = {};
-    var yaMayBackward = {};
+    var yaMustNotReverse = {};
```
archmoj: What was the problem with `Forward` and `Backfard` here?
`yaMustForward` may be less confusing that `yaMustNotReverse`.

etpinard: The attribute value of interest here is `reversed`. There's no `autorange: 'forward'`

test/jasmine/tests/image_test.js
etpinard: I'd just :hocho: the `only`. There's a nice progression in complexity in this `declare` block

src/plots/cartesian/layout_defaults.js
archmoj: Thanks. Resolved.

antoinerg: I made that change when introducing `yaMayScaleanchor` and `yaMayConstrain` (which have now disappeared).  Finding good name/word to evoke the opposite of `Scaleanchor` and `Constrain` was difficult. Instead, I decided to always reuse the same name/word but prepend a `Not` in front. I would argue that this pattern is simpler: you know right away it's the opposite thing. I realize now however that `MustDisplay` could/should be `MustNotHide`.

@archmoj I can revert it if you and @etpinard  really dislike it.

etpinard:
My comments about the `only` string are non-blocking

:dancer: - thanks very much @antoinerg for the quick turnaround!!

test/jasmine/tests/image_test.js
antoinerg: Sweet! Done in https://github.com/plotly/plotly.js/pull/4307/commits/408b46ba5d1bddcd496b9397f1864b0749ecec89

test/jasmine/tests/image_test.js
antoinerg: Done in https://github.com/plotly/plotly.js/pull/4307/commits/408b46ba5d1bddcd496b9397f1864b0749ecec89

test/jasmine/tests/image_test.js
antoinerg: Done in https://github.com/plotly/plotly.js/pull/4307/commits/408b46ba5d1bddcd496b9397f1864b0749ecec89

PR #4312, opened 2019-10-28
Handle invalid entry during treemap plot

### LINKED ISSUE #4308: Better error handling in treemap
Hi,

When given data for treemap is somehow flawed, Plotly terminates with error. I think there needs to be some error handling. 

In my case, plotting stops here with `Cannot read property 'data' of undefined`.
https://github.com/plotly/plotly.js/blob/990a96050bbab12e1e405d0974985b2fea54d2e6/src/traces/treemap/plot.js#L87

This is because `calc` has failed to build `cd.hierarchy`, but `plotOne` expects `hierarchy` to be there. 

Since treemap doesn't have axis, I don't know what a blank treemap plot would look like, but I suppose that's waht we need here.

archmoj:
Fix #4308

@etpinard

### DIFF
--- a/src/traces/treemap/plot.js
+++ b/src/traces/treemap/plot.js
@@ -82,9 +82,20 @@ function plotOne(gd, cd, element, transitionOpts) {
     var cd0 = cd[0];
     var trace = cd0.trace;
     var hierarchy = cd0.hierarchy;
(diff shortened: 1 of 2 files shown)


etpinard:
:dancer:

PR #4314, opened 2019-10-28
Fix box/meanline offset for one-sided vertical violins

etpinard:
fixes https://github.com/plotly/plotly.js/pull/4283#discussion_r339739636 

cc @archmoj and @s417-lama


archmoj:
@etpinard thanks very much for the PR.
:dancer:

PR #4316, opened 2019-10-29
Preserve line gaps in scattergl plot

### LINKED ISSUE #3551: scattergl & gap between points : strange behavior
Hello,

i find an interesting bug when using scattergl et gap between point.

I made an exemple with 2 traces :
- first trace have a gap with data in Y as [1,null,1,2]. The data is not the same for all point. It works as expected.
- second trace have a gap with data in Y as [3,null,3,3]. The data is the same for all points except the null.
It doesn't works because there is a line between the first & the third point.

An exemple is visible here : https://codepen.io/anon/pen/KJbEoX

No problem with scatter.

archmoj:
Fix #3551 
`array-normalize` module use to normalize `NaN` values which is now patched and used by `reg-line2d` 3.0.15.

@etpinard


test/image/mocks/gl2d_scattergl_gaps.json
```diff
@@ -0,0 +1,49 @@
+{
```
etpinard: Which currently renders as:

image


https://codepen.io/etpinard/pen/poorMMY

etpinard:
Awesome fix! 

:dancer: :dancer: :dancer:

PR #4317, opened 2019-10-29
append coverSlip to fullscreen element instead of body

### LINKED ISSUE #4259: Rangeslider not working in fullscreen
I have a graph with multiple traces, that uses rangeslider and rangeselector to navigate.

The rangeslider works fine on normal window size, even when it is resized, but when it goes in fullscreen mode it stop working.

Looks like there is some issue with the mouse events.

When I exit the fullscreen mode it works again, just looks in some cases bound to the mouse move, but one or 2 clicks and it start working again perfectly.

In fullscreen the graph zoom works, and the range selector works too.

Here there is an example of the issue:

https://codepen.io/alexmario74/pen/yLLyGgB?editors=0010

FlorianMaak:
Resolves #4259 

This PR checks if there is a fullscreen element and appends dragelements coverSlip to this element instead of documents body.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4295,6 +4295,11 @@
       "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8="
     },
+    "fscreen": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/fscreen/-/fscreen-1.0.2.tgz",
+      "integrity": "sha1-xMUdltgZ11oZ1yjg30Rfm+m7mE8="
+    },
     "fsevents": {
       "version": "2.0.7",
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.0.7.tgz",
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Thanks @FlorianMaak! But it looks to me as though the unprefixed `fullscreenElement` is not yet available everywhere - [perhaps only 45%](https://caniuse.com/#feat=fullscreen)? In some of our other packages we use [`fscreen`](https://www.npmjs.com/package/fscreen) to normalize this.

etpinard:
Thanks very much for the PR @FlorianMaak - beyond browser support, it would be nice to write a test in the [`dragelement_test.js`](https://github.com/plotly/plotly.js/blob/master/test/jasmine/tests/dragelement_test.js) suite to lock down your fix. Would be interested in attempting to write a test?

FlorianMaak:
Sure. Going to migrate to fscreen and add tests as soon, as I'm at office tomorrow.

FlorianMaak:
**Relating Tests:** I think, we need to edit the karma config to work with fullscreen. Per default karma does not support fullscreen mode :/

https://github.com/karma-runner/karma/issues/2292

archmoj:
@FlorianMaak could you please merge master into this branch and resolve the conflict?

gvwilson:

Issue #4301, opened 2019-10-24
`image` future directions


antoinerg:
Following the introduction of `image` trace type in https://github.com/plotly/plotly.js/pull/4289, here are a few future directions this trace type could go:
- add support for `x` and `y`
- add new color models
- add optimized rendering pathway in browsers that support it by drawing the image once and rely on CSS `image-rendering: pixelated` to provide a pixelated view of it

etpinard:
- renaming it `pixmap` :smile:

archmoj:
I was wondering it might be of interest to have an option of printing numbers (or other characters) instead of drawing rectangles. Something similar to `pixmap` actually!
https://franz.com/support/documentation/current/doc/cg/cg-pixmaps.htm
Also that reminds me of this article http://www.cs.utah.edu/~gk/papers/vissym04/vissym04kindlmann.pdf which @nicolaskruchten showed me a while ago may be possible using vectors.

archmoj:
Also to metion: 
- supporting typed arrays

emmanuelle:
I have the impression that at the moment nan values are converted to 0. We could instead keep the same behaviour as `Heatmap` where nan regions are transparent and show the background.

antoinerg:
> I have the impression that at the moment nan values are converted to 0. We could instead keep the same behaviour as `Heatmap` where nan regions are transparent and show the background.

I agree we should probably clean things up when we're not receiving numbers and show a transparent color. This is implemented in the following https://github.com/plotly/plotly.js/compare/master...image-nan. Now the question is what to display in the hover for color? A transparent pixel should not exist in `rgb` or `hsl` so there's no way of defining it there. I think not displaying anything would be better. As you know, I'm leaving in vacation tomorrow but feel free to take over the `image-nan` branch!

emmanuelle:
The optimized rendering pathway would be great to display large images (1000x1000 and larger) in reasonable time. I'm a big fan of adding this possibility.

antoinerg:
Commit https://github.com/plotly/plotly.js/commit/d9a17cd8cb9e2798028321f71a101a52d089030a shows an early example using `image-rendering: pixelated`.

gvwilson:
[...]

PR #4325, opened 2019-10-31
Skip non-numeric values in image trace plot 

archmoj:
Fix in respect to @emmanuelle 's comment
https://github.com/plotly/plotly.js/issues/4301#issuecomment-548103964
i.e. by drawing those pixels with full transparency.

[before vs after demo](https://codepen.io/MojtabaSamimi/pen/oNNoMBE?editors=0010)

@alexcjohnson 

cc: @nicolaskruchten 
cc: @plotly/plotly_js


test/image/mocks/image_non_numeric.json
```diff
@@ -0,0 +1,16 @@
+{
+  "data": [{
+    "type": "image",
+    "hoverinfo": "all",
+    "z": [
+      [
+        [true, true, true],
+        [false, false, false]
```
alexcjohnson: True, there's something nice about ensuring we see the right axis ranges even when *all* of the pixels are invalid. But perhaps then a second mock (or a second subplot on the same one?) that includes a couple of valid pixels as well as various kinds of invalid pixels - any single channel missing for example.

archmoj: Good call. That mock should be improved.

archmoj: Done in https://github.com/plotly/plotly.js/pull/4325/commits/d6abde903a8988edc218b63498381d9d2fa472a5.

etpinard:
Nice PR @archmoj - can I why you made some of the colorscale attributes `editType: 'calc'` now? I suspect this was to make the plot step faster, but I'd like to confirm with you.

archmoj:
@etpinard Yes. That was for speed and ( if I recall ) not for calling `isNumeric` during the `plot`.

etpinard:
> Yes. That was for speed and ( if I recall ) not for calling `isNumeric` during the `plot`.

Yeah, that very nice. Most colorscale computations for other traces are done during the calc step, so this is fantastic :+1:

Issue #2477, opened 2018-03-15
lightposition behaves weirdly


alexcjohnson:
`lightposition` behaves similarly for both `mesh3d` and `surface`, but it's nearly unusable. Generally x is left (-) to right (+) and y is bottom (-) to top(+), but z has effects that defy easy categorization other than z=1 being "special". Sometimes x and y flip randomly, and anyway they don't seem to have the same scales, ie x=y does not put the light at 45 degrees.

etpinard:
https://github.com/plotly/plotly.js/pull/3415 will bring `mesh3d` and `surface` lightposition to the same level. 

It is still broken, but we might want to wait until v2 or the regl rewrite (cc #https://github.com/plotly/plotly.js/issues/949) before spending more time on this.

archmoj:
As illustrated in this [demo](https://codepen.io/MojtabaSamimi/pen/jOOLrKL?editors=0010),
if we choose to apply `lightposition.y` for `surface` similar to other `gl3d` traces, and enable `flat shading` for `mesh3d` traces the result looks OK. 
I can imagine finding such a setup could be pretty frustrating for users.
What about revising the defaults of those parameters?
I don't think we should enable smoothing for `mesh3d` by default.
And why `surface` plot should have a different light position by default?

alexcjohnson:
`flatshading` definitely helps - in fact without it I can't really figure out any physical meaning to the lighting of `mesh3d` traces, but with it there does seem to be, for any given `lightposition`, a light that looks like it's coming from *somewhere* fixed wrt the scene as the scene is rotated. And moreover that fixed somewhere is the same for all traces. So based on that alone I'd probably enable `flatshading` by default. And I'm highly in favor of standardizing all the `lightposition` defaults to something with intelligible consequences, with all traces matching each other. In fact, can we add a `scene.lightposition` that all the traces in a scene inherit from if not provided at the trace level?

Now, as to where this light appears to be, my description above still seems to hold. I'd add that when both x and y are fairly small (±hundreds or less) or when z is significant (<-10 or >1) it no longer seems possible to assign the light a fixed position as the scene rotates - I can align the cubes or saddle surfaces in what should be symmetric viewpoints (90-degree rotations of the cube, 180-degree rotations of the saddle) and I get different light regions.

So I definitely would not close this issue. To me this would be closed when the meaning of `lightposition.x/y/z` can be precisely defined, something like:
[...]

PR #4344, opened 2019-11-08
remove noOpacity flag from surface index

archmoj:
This PR simply removes the confusing `noOpacity` flag from `surface` index file.
Please note that the opacity used to be (& would remain) available in `surface` plot as the attribute was added by module `opacity` later in the process by `plot_schema`.

@alexcjohnson 
cc: @plotly/plotly_js


test/image/mocks/gl3d_traces-with-opacity.json
```diff
@@ -0,0 +1,173 @@
...
+      "b": 0,
+      "l": 0,
+      "r": 0
+    },
+    "width": 1000,
+    "height": 500,
+    "scene1": {
+      "domain": {
+        "x": [0, 0.25],
+        "y": [0, 0.25]
```
alexcjohnson:
```suggestion
        "y": [0, 0.5]
```

archmoj: Good eye! Done in https://github.com/plotly/plotly.js/pull/4344/commits/0dd87488852d7950fa8b9757d188dc8b3506ba82.
I also think we may want to close https://github.com/plotly/plotly.js/issues/2477.

etpinard:
Thanks @archmoj - I'd call this a (documentation) bug fix as `opacity` is currently omitted from the `surface` trace plot schema.

PR #4346, opened 2019-11-12
Do not coerce contour & colorscale attributes for mesh3d when not needed

archmoj:
It appears there is no need for coercing `colorscale` and `reversescale` attributes of `mesh3d`.
Those attributes are handled by `colorscaleDefaults` function later in the `default` process in case the trace has `intensity`.

I addition, now the `contour` attributes won't be coerced when `contour.show` is disabled. See https://github.com/plotly/plotly.js/pull/4346/commits/67e7fbf43598d1d501f8419a7f6372c02f600ba4.

@alexcjohnson 
cc: @plotly/plotly_js

### DIFF
--- a/src/traces/mesh3d/defaults.js
+++ b/src/traces/mesh3d/defaults.js
@@ -68,8 +68,6 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         'contour.show',
         'contour.color',
         'contour.width',
-        'colorscale',
-        'reversescale',
         'flatshading',
         'alphahull',
         'delaunayaxis',
--- a/src/traces/mesh3d/defaults.js
+++ b/src/traces/mesh3d/defaults.js
@@ -65,14 +65,17 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         'lightposition.x',
         'lightposition.y',
         'lightposition.z',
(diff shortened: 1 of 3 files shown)


antoinerg:
> It appears there is no need for coercing `colorscale` and `reversescale` attributes of `mesh3d`.
> Those attributes are handled by `colorscaleDefaults` function later in the `default` process in case the trace has `intensity`.

Would this be easy to test via a jasmine test?

archmoj:
> > It appears there is no need for coercing `colorscale` and `reversescale` attributes of `mesh3d`.
> > Those attributes are handled by `colorscaleDefaults` function later in the `default` process in case the trace has `intensity`.
> 
> Would this be easy to test via a jasmine test?

Good call! 
Addressed in https://github.com/plotly/plotly.js/pull/4346/commits/67e7fbf43598d1d501f8419a7f6372c02f600ba4.

antoinerg:
Thank you @archmoj for adding tests in https://github.com/plotly/plotly.js/commit/67e7fbf43598d1d501f8419a7f6372c02f600ba4 !

Non-blocking comment: you could significantly DRY the code by copying your `traceIn` object and removing/adding properties needed for each test case.

Other than that :dancer:

etpinard:
Nicely done @archmoj !

PR #4347, opened 2019-11-12
Improve gl3d coloraxis mock

archmoj:
This PR add a `surface` trace to `coloraxis` mock of 3-D traces.
- adjust camera position in scene
- use isosurface with a defined bound
- reverse one of the colorscales

@antoinerg


antoinerg:
Ideally, @etpinard should review this mock since he's the one who created it. However, since you're only adding things to it, it's probably safe to merge.

:dancer:

etpinard:
thanks @archmoj !

Issue #4349, opened 2019-11-13
mesh3d error when missing a vertexcolor


archmoj:
[demo](https://codepen.io/MojtabaSamimi/pen/YzzJWYM?editors=0010)

PR #4353, opened 2019-11-14
Handle missing vertexcolor and facecolor in mesh3d

### LINKED ISSUE #4348: mesh3d error when missing a facecolor
[demo](https://codepen.io/MojtabaSamimi/pen/PooyNgL?editors=0010)

archmoj:
Fix #4348 & #4349
`gl-mesh3d` used to stop when encountering a missing `vertexcolor` or `facecolor` which is fixed by `gl-mesh3d` v2.1.2 and v2.1.3 patches.
This PR bumps `gl-mesh3d` and adds an image test.

@plotly/plotly_js


alexcjohnson:
rename `mising` -> `missing`

missing vertex color reverts to white, but missing face color is omitted. Is that intentional / desirable? Offhand I'd think we'd want the faces to be white as well. Or perhaps a light gray in both cases...

archmoj:
We could use gray color for both cases which is pretty easy to do.
But I think when encountring missing data we should display the face or vertex as invisible.
In that regard, we could try turning vertices with missing `vertexcolor` to full transparent.
@alexcjohnson what do you suggest?

alexcjohnson:
In general I feel like data should not disappear if some secondary characteristic of it (color in this case, but more generally anything beyond its position) is missing or invalid. For example scatter traces with a short `marker.color` array treat the missing items as black. (that said I see that a short `marker.size` array yields size=0, this argument would say it should yield the default size)

This is mostly useful for debugging, so you can see what parts you got right and what parts have issues. If your goal is to make a transparent face, you should specify this explicitly rather than relying on an implicit "bad data becomes transparent"

archmoj:
> In general I feel like data should not disappear if some secondary characteristic of it (color in this case, but more generally anything beyond its position) is missing or invalid. For example scatter traces with a short `marker.color` array treat the missing items as black. (that said I see that a short `marker.size` array yields size=0, this argument would say it should yield the default size)
> 
> This is mostly useful for debugging, so you can see what parts you got right and what parts have issues. If your goal is to make a transparent face, you should specify this explicitly rather than relying on an implicit "bad data becomes transparent"

@alexcjohnson Good call.
Done in https://github.com/plotly/plotly.js/pull/4353/commits/3066f32660586ef4736597a8b6eb787ab4570088.

PR #4359, opened 2019-11-18
Skip dimensions including non-numeric values on parcoords trace

### LINKED ISSUE #4341: Should skip missing data in parcoords
Reported in https://community.plot.ly/t/plotly-graph-objects-parcoords-nan-replaced-by-0/29725
`If some of the columns contain nan during the display of plotly.graph_objects.Parcoords the data are displayed as 0, it will be nice to don’t to treat the value as a nan and not as 0`.

[demo](https://codepen.io/MojtabaSamimi/pen/oNNyWPm?editors=0010)

archmoj:
Fix #4341 
Sets `dimension.visible` to `false` when encountering non-numeric values.

@plotly/plotly_js


test/image/mocks/gl2d_parcoords_skip_null.json
```diff
@@ -0,0 +1,37 @@
...
+                        2,
+                        3
+                    ]
+                },
+                {
+                    "label": "B",
+                    "values": [
+                        -1,
+                        null,
+                        1
```
etpinard: Wait. `-1` and `1` are numeric, why doesn't this dimension show up?

PR #4360, opened 2019-11-18
Refactor gl3d scene

### LINKED ISSUE #4236: Bad webgl alpha config for 3-D scenes
Noticed blending issues (see the yellow areas on the image below) with Ubuntu + FireFox. 
This could be related to `premultipliedAlpha: true` instead of `false`.
See `gl-plot3d` at https://github.com/gl-vis/gl-plot3d/blob/ccfc4318ec9d382900f7b80b17b1630092fe3953/scene.js#L84.

blendingFireFoxUbuntu

archmoj:
This PR adds setters to `gl-plot3d` and refactors functions/methods of `gl3d/scene.js` file to use `gl-plot3d` functions to change camera attributes and avoid two camera instances, etc.

@plotly/plotly_js 
Please review this PR commit by commit.

After all, this PR does NOT f_i_x  #4236 by passing a different config (premultipliedAlpha: false) to `gl-plot3d` module. As used to be suggested in this [demo](https://codepen.io/MojtabaSamimi/pen/LYYaERd) {to compare before and after effects on Windows(7/10) using Chrome/Firefox/Opera as well as on Ubuntu using Firefox}.
That may be addressed in separated PR in future.


package.json
```diff
@@ -81,7 +81,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.1.1",
     "gl-plot2d": "^1.4.2",
-    "gl-plot3d": "^2.3.0",
+    "gl-plot3d": "git://github.com/gl-vis/gl-plot3d.git#e1858a63bb6617474145ee4ae37106eddf3b3fa3",
```
archmoj: https://github.com/gl-vis/gl-plot3d/pull/25

src/plots/gl3d/scene.js
```diff
@@ -97,7 +97,7 @@ proto.tryCreatePlot = function() {
         gl: scene.gl,
         glOptions: {
             preserveDrawingBuffer: isMobile,
-            premultipliedAlpha: true,
+            premultipliedAlpha: false,
```
etpinard: Hmm. Does this lead to any changes in performance ?

etpinard:
> Fix #4236 by passing different config to `gl-plot3d` module.

Is the only change in the WebGL config the `premultipliedAlpha` value?

src/plots/gl3d/scene.js
archmoj: Good question. It shouldn't. Anyway I would double check.

archmoj:
> > Fix #4236 by passing different config to `gl-plot3d` module.
> 
> Is the only change in the WebGL config the `premultipliedAlpha` value?

Yes that is the only change in the config.

archmoj:
before vs after windows10+chrome78
beforeAfter_windows10_chrome78b

archmoj:
before vs after windows10+chrome78

beforeAfter_windows10_chrome78

archmoj:
before vs after windows10+firefox70

beforeAfter_windows10_firefox70

archmoj:
Also added a codepen to the PR description to help compare the two options.
After figuring out which `premultipliedAlpha` option to use as the default, it may be best to provide both options in the API.
Anyway applying `premultipliedAlpha: false` would make the graphs render similar to save as `png` version. 
cc: @alexcjohnson @antoinerg

archmoj:
Then in order to keep the best view and consistent png results we could default `layout.scene.bgcolor` to white instead of `rgba(0,0,0,0)`.

src/plots/gl3d/scene.js
archmoj: We decided not to change this config in this PR for now: https://github.com/plotly/plotly.js/pull/4360/commits/ade22a674393762b8a7262c368acb5e0fae0661f.

archmoj:
@etpinard all the tests passed and this PR should be ready to merge.

src/plots/gl3d/scene.js
```diff
@@ -33,7 +34,233 @@ var computeTickMarks = require('./layout/tick_marks');
...
+proto.tryCreatePlot = function() {
+    var scene = this;
+    var opts = {
+        canvas: scene.canvas,
+        gl: scene.gl,
+        glOptions: {
+            preserveDrawingBuffer: isMobile,
+            premultipliedAlpha: true,
+            antialias: true
+        },
```
etpinard: Can I ask why you're now passing these options from plotly.js? The corresponding option object current on master is:

```js
    var glplotOptions = {
        canvas: canvas,
        gl: gl,
        container: scene.container,
        axes: scene.axesOptions,
        spikes: scene.spikeOptions,
        pickRadius: 10,
        snapToData: true,
        autoScale: true,
        autoBounds: false,
        cameraObject: cameraObject,
        pixelRatio: pixelRatio
    };
```

that is, the `glOptions` container is new.

archmoj: Where is that set on master?

etpinard: https://github.com/plotly/plotly.js/blob/8122c2127c12b50cb20207d676aa24e3a786f305/src/plots/gl3d/scene.js#L201-L213

archmoj: https://github.com/gl-vis/gl-plot3d/blob/d206f1f110d427b2baae0462964bc59cf7f59dae/scene.js#L80-L87
The `glOptions` used to be left undefined and hidden to `plolty.js` developers. 
Now that are explicitly defined, it should be easier to figure out what the differences between static & interactive plots are.

etpinard: Ok, that's fair. But then `is-mobile` becomes a top-level dependency for plotly.js and we should add it to our `package.json`.

archmoj: Good catch. OK if I add it to `package.json`?

etpinard: > OK if I add it to `package.json`

yes please!

archmoj: Simply listed in https://github.com/plotly/plotly.js/pull/4360/commits/ff2ecc796dd0607b7a07baac8a20e6104adbfe31.
But also noticed that it was added in package-lock before which is not clear why?
https://github.com/plotly/plotly.js/commit/75de9e5dea

etpinard: > But also noticed that it was added in package-lock before which is not clear why?

Yeah it was in the package-lock before, because the package-lock list all the dependencies not just the top-level ones. It's a best practice to include all the top-level dependencies in the package.json, and make npm handles the nested properties part in the package-lock.

etpinard:
:dancer: - nicely done!

Before merging, could you change the PR title to something like "Refactor gl3d scene" as this PR no longer disable `premultipliedAlpha`.

PR #4364, opened 2019-11-19
Option to set class on given shape using className attribute

mitja-p:
I am trying to allow the functionality described here and open possible other uses:
https://community.plot.ly/t/lock-the-editing-of-few-shapes/25033/3

In this case you could set CSS for example: 
.shapelayer path.locked {
    pointer-events: none !important;
}

### DIFF
--- a/src/components/shapes/draw.js
+++ b/src/components/shapes/draw.js
@@ -93,7 +93,8 @@ function drawOne(gd, index) {
         var attrs = {
             'data-index': index,
             'fill-rule': 'evenodd',
-            d: getPathString(gd, options)
+            d: getPathString(gd, options),
+            class: options._input.className
         };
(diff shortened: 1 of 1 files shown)


etpinard:
Hi @mitja-p 

We don't allow users to set custom class names anywhere in the library at the moment. So what your PR is implementing would be a first, and hence is unlikely to get merged.

Sounds like the end-result you're looking for is to disable `{ediatable: true}` interactions for _some_ shapes. Without a doubt, this would a nice addition to the library. We already have similar logic for annotations behind the [`captureevents`](https://github.com/plotly/plotly.js/blob/dbf02f1b47f2b0e17a1b78fabee815e3cde18715/src/components/annotations/attributes.js#L520-L532) boolean attribute.

If you're interested in making a PR that adds `captureevents` to shapes, please go ahead and we'll help you get it ready to be merged in master.

etpinard:

PR #4365, opened 2019-11-19
An option to ignore certain shapes when doing autosize

mitja-p:
In some cases it is useful to ignore some shapes when auto sizing as they might stretch beyond current range on purpose or we do not even care where they are. By adding an option autorange (default: true), we can individually remove some shapes from auto sizing algorithm.

### DIFF
--- a/src/components/shapes/calc_autorange.js
+++ b/src/components/shapes/calc_autorange.js
@@ -23,6 +23,7 @@ module.exports = function calcAutorange(gd) {
 
     for(var i = 0; i < shapeList.length; i++) {
         var shape = shapeList[i];
+        if (shape._input.autorange==false) continue;
         shape._extremes = {};
 
         var ax, bounds;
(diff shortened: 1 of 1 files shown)


gvwilson:

gvwilson:
@emilykl when you have a sec could you please let us know if this was closed by your earlier autosizing work? thanks - @gvwilson

emilykl:
Hi @mitja-p, thank you very much for the contribution.

If I understand correctly, the intention is for the chart developer to be able to specify certain shapes to exclude from axis autorange calculations, is that right?

If so I don't believe this change will have the desired effect. If you want to add a new shape attribute (something like `autorangeignore`) it should be added in [`src/components/shapes/attributes.js`](https://github.com/plotly/plotly.js/blob/master/src/components/shapes/attributes.js).

Feel free to also share a code example demonstrating the desired outcome.

mitja-p:
Yes, that is right. I forgot the exact purpuse, but I think it was for helper lines that should be ignored when doing auto range as they are visual aid and not a thing that needs to be fully visible.

emilykl:
@mitja-p I'm going to close this PR for the moment since more work is needed — please feel free to reopen at any point if/when you have the chance to continue working on it. Thank you for your interest in contributing to Plotly.js!

If you prefer, you can also [open a feature request issue](https://github.com/plotly/plotly.js/issues/new?template=feature_request.md) describing the behavior you'd like to see, and we or a community member will get to it when we can.

Issue #4337, opened 2019-11-06
coloraxis range not react()ing


nicolaskruchten:
As raised in https://github.com/plotly/react-plotly.js/issues/167 and at least one Dash-related report somewhere that I can't find just now, it seems like `layout.coloraxis` isn't `react()`ing completely... codepen: https://codepen.io/nicolaskruchten/pen/YzzLXmy?editors=0010

nicolaskruchten:
Basically the range grows but doesn't shrink :)

alexcjohnson:
Another example, from https://github.com/plotly/dash-core-components/issues/702: https://codepen.io/alexcjohnson/pen/JjjwBKE?editors=0010

PR #4366, opened 2019-11-19
Clear relinked cmin/cmax values in shared axes

etpinard:
... early in `Plots.doCalcata`, such that the cross-trace cmin/cmax aggregation starts from scratch.

This fixes https://github.com/plotly/plotly.js/issues/4337 - a bug that originates from the first shared coloraxis implementation (cf https://github.com/plotly/plotly.js/pull/3803 https://github.com/plotly/plotly.js/commit/5bb341ebbb0d80eae121d986627ae5823e3ff4b1)

before: https://codepen.io/nicolaskruchten/pen/YzzLXmy?editors=0010
after: https://codepen.io/etpinard/pen/rNNPvxy?editors=0010

before: https://codepen.io/alexcjohnson/pen/JjjwBKE
after: https://codepen.io/etpinard/pen/gOOqzPE

cc @archmoj

### DIFF
--- a/src/plots/plots.js
+++ b/src/plots/plots.js
@@ -2828,6 +2828,15 @@ plots.doCalcdata = function(gd, traces) {
         );
     }
 
+    // clear relinked cmin/cmax values in shared axes to start aggregation from scratch
+    for(var k in fullLayout._colorAxes) {
+        var cOpts = fullLayout[k];
+        if(cOpts.cauto !== false) {
+            delete cOpts.cmin;
+            delete cOpts.cmax;
+        }
+    }
+
     var hasCalcTransform = false;
 
     function transformCalci(i) {
--- a/test/jasmine/tests/colorscale_test.js
+++ b/test/jasmine/tests/colorscale_test.js
@@ -1101,6 +1101,42 @@ describe('Test colorscale restyle calls:', function() {
         .then(done);
     });
 
+    it('should update coloraxis cmin/cmax on color value changes', function(done) {
+        function fig(mc) {
+            return {
+                data: [{
+                    mode: 'markers',
+                    y: [1, 2, 3],
+                    marker: {
+                        coloraxis: 'coloraxis',
+                        color: mc
+                    }
+                }]
+            };
+        }
+
+        function _assert(msg, cmin, cmax) {
+            return function() {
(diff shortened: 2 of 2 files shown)


src/plots/plots.js
```diff
@@ -2828,6 +2828,15 @@ plots.doCalcdata = function(gd, traces) {
         );
     }
 
+    // clear relinked cmin/cmax values in shared axes to start aggregation from scratch
+    for(var k in fullLayout._colorAxes) {
```
archmoj: Do we need to turn this to
```js
for(var k in (fullLayout._colorAxes || [])) {
```
?

src/plots/plots.js
```diff
@@ -2828,6 +2828,15 @@ plots.doCalcdata = function(gd, traces) {
...
+        var cOpts = fullLayout[k];
+        if(cOpts.cauto !== false) {
+            delete cOpts.cmin;
+            delete cOpts.cmax;
```
archmoj: Should we do something similar for `cmid`?

src/plots/plots.js
etpinard: `_colorAxes` is initialised during `Plots.supplyDefaults`:

https://github.com/plotly/plotly.js/blob/513491e6a8f20e247881609eefa74ffdb6c7c319/src/plots/plots.js#L390

which should always be called before `Plots.doCalcdata`, so adding a fallback here would be pretty much useless.

src/plots/plots.js
etpinard: No, we don't need that.

`cmid` is not relinked during the [cross-trace-defaults](https://github.com/plotly/plotly.js/blob/513491e6a8f20e247881609eefa74ffdb6c7c319/src/components/colorscale/cross_trace_defaults.js#L23-L47) and we don't aggregate its value during `calc`, like we do for `cmin` and `cmax` :arrow_heading_down: 

https://github.com/plotly/plotly.js/blob/513491e6a8f20e247881609eefa74ffdb6c7c319/src/components/colorscale/calc.js#L37-L38

https://github.com/plotly/plotly.js/blob/513491e6a8f20e247881609eefa74ffdb6c7c319/src/components/colorscale/calc.js#L47-L48

archmoj:
@etpinard thanks very much for the info.
:dancer:

PR #4385, opened 2019-11-25
Fix bug wheel event for IE.

ZgerskiyVadim:
Wheel event not work in IE.
With additional code that i wrote works fine. Check it please. Thank you


etpinard:
Hi @ZgerskiyVadim - thanks very much for the PR!!

Which version of IE are you referring to?

ZgerskiyVadim:
IE 11

src/plots/cartesian/dragbox.js
```diff
@@ -1202,6 +1202,7 @@ function attachWheelEventHandler(element, handler) {
     if(!supportsPassive) {
         if(element.onwheel !== undefined) element.onwheel = handler;
         else if(element.onmousewheel !== undefined) element.onmousewheel = handler;
+        else { element.addEventListener('wheel', handler, {passive: false}); }
```
alexcjohnson: *sigh* IE is so ~broken~ weird. `element.onmousewheel` seems `null` (what we're looking for as `!==undefined` for "will handle events if set") for html elements but not for svg elements. And likewise setting `onwheel` doesn't appear to do anything but `addEventListener('wheel', ...)` does.

But is there a case in which this function would get called multiple times on the same element? If so that would behave differently for the two methods, and we'd better remove any previous wheel handlers first.

ZgerskiyVadim: Yeah, this function would get called multiple times. For call function only once for element maybe create some flag for element and check it before add this event.

For example:

if (!element.isAddedWheelEvent) {
element.addEventListener('wheel', handler, {passive: false});
element.isAddedWheelEvent = true;
} 

Unfortunately
Note: In IE, the wheel event is only supported with the addEventListener() method. There is no onwheel attribute on DOM objects.
https://www.w3schools.com/jsref/event_onwheel.asp

src/plots/cartesian/dragbox.js
```diff
@@ -1202,6 +1202,10 @@ function attachWheelEventHandler(element, handler) {
     if(!supportsPassive) {
         if(element.onwheel !== undefined) element.onwheel = handler;
         else if(element.onmousewheel !== undefined) element.onmousewheel = handler;
+        else if(element.isAddedWheelEvent) {
```
etpinard: Hmm. Shouldn't this be 

```js
else if(!element.isAddedWheelEvent) { 
// ...
```

?

Xerotico: Hey! You are right!
I just fixed it

archmoj:
@ZgerskiyVadim
Nicely done.
:dancer:

PR #4387, opened 2019-11-25
test_dashboard fix for ie11

alexcjohnson:
not sure if anyone else still uses `test_dashboard` (`npm start`) but I needed these two fixes to get it to run in IE11. There's still an error later on about calling `forEach` on a `NodeList` from `querySelectorAll` but that one didn't stop me from getting the graph I wanted so I didn't hunt that down 😅


etpinard:
:dancer: - thanks!!

Issue #4088, opened 2019-07-29
What to do with notched box traces with "inverted ends" ?


destradafilm:
I'm implementing the boxplot notched option, but don't really know what it is exactly and how to explain it to users that choose that option. The plotly documentation doesn't explain how it's calculated

the only thing I found useful online when researching was this:
https://sites.google.com/site/davidsstatistics/home/notched-box-plots

> The Notch - displays the a confidence interval around the median which is normally based on the median +/- 1.57 x IQR/sqrt of n.  According to Graphical Methods for Data Analysis (Chambers, 1983) although not a formal test the, if two boxes' notches do not overlap there is ‘strong evidence’ (95% confidence) their medians differ.>

Also, there are inverted ends when I add notches. I'm not sure if this is a bug or expected behavior
<img width="520" alt="Screen+Shot+2019-07-29+at+4 17 03+PM" src="https://user-images.githubusercontent.com/6043499/62087353-ab0c8c00-b21e-11e9-911c-ecdde6cf377d.png">

etpinard:
> Also, there are inverted ends when I add notches.

Thanks for reporting. Could you share the data/layout you used to help us debug?

destradafilm:
Here is a codepen with an example of the bug:
https://codepen.io/destrada/pen/WVONOK

destradafilm:
hi @etpinard . I'm wondering if this has been labeled as a bug and if the codepen I provided helps show what the bug is? Thanks!

etpinard:
The problem here as to do with samples that have median confidence intervals that go beyond Q1 (and/or Q3) and beyond the lower (and/or upper) fence value. 

After a quick search, I couldn't find any notched boxplot examples that show "inverted ends". I suspect that some libraries bound the confidence interval to the IQR. But from https://github.com/plotly/plotly.js/pull/2305 (the PR that added notched boxes), looks like the author (@krassowski ) was ok with those "inverted ends" as that PR added two mocks showing inverted ends:

- https://rreusser.github.io/plotly-mock-viewer/#box_notched
- https://rreusser.github.io/plotly-mock-viewer/#box_horz_notched

If "inverted ends" are a thing, then the problem is in our autorange-finding routine which should consider the median confidence interval in its computations.

If "inverted ends" were added in by mistake, then we should simply bound the median confidence interval accordingly in the calc step.

alexcjohnson:
[...]

PR #4388, opened 2019-11-26
Fix notched box autorange

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4088 - after confirmation that notched boxes with inverted ends are valid in https://github.com/plotly/plotly.js/issues/4088#issuecomment-558504399

before: https://codepen.io/destrada/pen/WVONOK
after: https://codepen.io/etpinard/pen/YzzoeKe

cc @plotly/plotly_js


archmoj:
@etpinard thanks for the PR.
Looks great to me.
:dancer:

PR #4391, opened 2019-11-29
Move bar transform functions to bar helpers

archmoj:
Before addressing https://github.com/plotly/plotly.js/issues/4247
this PR remove bar transform functions namely `toMoveInsideBar` & `toMoveOutsideBar` from `bar/plot` to `bar/helpers` as those function are required by traces e.g. `treemap`.
This can also help make bar plot code (which used by `histogram`, `funnel` and `waterfall`) more compact.
 
@plotly/plotly_js

### DIFF
--- a/src/traces/bar/helpers.js
+++ b/src/traces/bar/helpers.js
@@ -11,6 +11,7 @@
 var isNumeric = require('fast-isnumeric');
 var tinycolor = require('tinycolor2');
 var isArrayOrTypedArray = require('../../lib').isArrayOrTypedArray;
+var TEXTPAD = require('./constants').TEXTPAD;
 
 exports.coerceString = function(attributeDefinition, value, defaultValue) {
     if(typeof value === 'string') {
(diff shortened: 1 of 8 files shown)


etpinard:
I'm not sure I approve. `helpers.js` files are usually kept from utilities that are used across different methods in a given trace module. 

For example, `castOption` for pies

https://github.com/plotly/plotly.js/blob/dbf02f1b47f2b0e17a1b78fabee815e3cde18715/src/traces/pie/helpers.js#L37-L40

is called in `plot.js` and `style_one.js`

https://github.com/plotly/plotly.js/blob/dbf02f1b47f2b0e17a1b78fabee815e3cde18715/src/traces/pie/plot.js#L79

https://github.com/plotly/plotly.js/blob/dbf02f1b47f2b0e17a1b78fabee815e3cde18715/src/traces/pie/style_one.js#L16-L17

---

So, I don't see anything wrong with keeping all bar "plotting" subroutines in `bar/plot.js`. I don't see anything wrong with having `treemap` require `bar/plot.js` either. Moving stuff around simply to make things _more compact_ (which is subjective, by the way) is usually a waste of time.

archmoj:
I see the point that `helpers` is not the best place for putting these functions.
So maybe we should put them in a separate file (or leave them where they are).
By the way here is a comment from `treemap` PR which is quite related:
https://github.com/plotly/plotly.js/pull/4185#discussion_r323926387

etpinard:
> So maybe we should put them in a separate file (or leave them where they are).

I vote for leaving them in `bar/plot.js`.

Issue #2856, opened 2018-07-30
Add zoom buttons to Mapbox plots


maxwell8888:
Is it possible to add zoom controls to the hoverbar for Mapbox plots? I have had complaints from users that they couldn't work out how to zoom in and out of my Mapbox charts.
Regular Plotly maps have + - zoom buttons in the hoverbar - see screenshot from https://plot.ly/javascript/scatter-plots-on-maps/:
![image](https://user-images.githubusercontent.com/28952593/43389984-0ad256dc-93e5-11e8-89f3-433bf231f0f3.png)
However, these buttons don't seem to be available in Mapbox plots - see screenshot from https://plot.ly/javascript/scattermapbox/:
![image](https://user-images.githubusercontent.com/28952593/43390144-66154536-93e5-11e8-80c0-5012a50a823a.png)

zoohair:
any ETA on this :) ?

johnWilshire:
This would be really useful feature.

etpinard:
done in https://github.com/plotly/plotly.js/pull/4398

PR #4398, opened 2019-12-02
Add modebar buttons zoomInMapbox and zoomOutMapbox

etpinard:
Implements https://github.com/plotly/plotly.js/issues/2856

The initial commit uses a scalar value of `1.3`, which gives a smaller zoom increment than for geo and cartesian subplots (they use `2`). I personally prefer this behaviour for the vector-tiled mapbox subplots. Please try out the example codepen below and tell me if you agree.

demo: https://codepen.io/etpinard/pen/qBEBPmN

cc @plotly/plotly_js

### DIFF
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -641,7 +641,7 @@ function handleMapboxZoom(gd, ev) {
     var val = button.getAttribute('data-val');
     var fullLayout = gd._fullLayout;
     var subplotIds = fullLayout._subplots.mapbox || [];
-    var scalar = 1.3;
+    var scalar = 1.05;
     var aObj = {};
 
     for(var i = 0; i < subplotIds.length; i++) {
--- a/src/components/modebar/buttons.js
+++ b/src/components/modebar/buttons.js
@@ -618,6 +618,42 @@ modeBarButtons.resetViewMapbox = {
     }
 };
 
+modeBarButtons.zoomInMapbox = {
+    name: 'zoomInMapbox',
+    title: function(gd) { return _(gd, 'Zoom in'); },
+    attr: 'zoom',
+    val: 'in',
+    icon: Icons.zoom_plus,
+    click: handleMapboxZoom
+};
+
+modeBarButtons.zoomOutMapbox = {
+    name: 'zoomOutMapbox',
+    title: function(gd) { return _(gd, 'Zoom out'); },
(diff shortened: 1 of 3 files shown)


src/components/modebar/buttons.js
```diff
@@ -618,6 +618,42 @@ modeBarButtons.resetViewMapbox = {
...
+    icon: Icons.zoom_minus,
+    click: handleMapboxZoom
+};
+
+function handleMapboxZoom(gd, ev) {
+    var button = ev.currentTarget;
+    var val = button.getAttribute('data-val');
+    var fullLayout = gd._fullLayout;
+    var subplotIds = fullLayout._subplots.mapbox || [];
+    var scalar = 1.3;
```
archmoj: This is changing fast (specially when having angles for display urban areas).
I suggest using `1.02` instead of `1.3` here.

etpinard: Right, but I thought value lower than 1.3 were "too slow" for world-scope maps.

We could introduce some sort of non-linear scale, but then that might be too inconsistent with the other zoom in/out buttons we have.

archmoj: I don't think a constant close to `1.02` is "too slow" for world maps.
Would you like me to make a codepen to help compare the two?

etpinard: > I suggest using `1.02`

> Would you like me to make a codepen to help compare the two?

I'm curious, how did you come to suggesting `1.02` with testing it beforehand?

archmoj: I tested it by changing it on the local branch.

etpinard: Ok, so sounds like you should be the one making a codepen

archmoj: My pleasure :slightly_smiling_face:

archmoj: [Demo](https://codepen.io/MojtabaSamimi/pen/VwYwxPo?editors=1000) to compare zoom speed 1.3 on the left with 1.02 on the right.

etpinard: @archmoj I picked `1.05` in my latest commit - which is somewhere between our two values.

Try out: https://codepen.io/etpinard/pen/wvBWpRQ

Is that good enough for you?

archmoj:
Thanks @etpinard 
:dancer: :dancer:

Issue #4399, opened 2019-12-02
ENH: provide mapbox viewport coordinates in relayout data


jonmmease:
I would like the mapbox subplot relayout data to include the `lon`/`lat` of the viewport corners. This would make it possible to use a mapbox subplot to display an image layer that is dynamically generated for the current viewport.

The relayout data currently includes `mapbox.center`, `mapbox.zoom`, `mapbox.bearing`, and `mapbox.pitch`.  I would like to propose that we include a `mapbox.coordinates` property that contains an array of the lon/lat coordinates of the four corners of the viewport. 

This would match the `coordinates` property that is expected in `image` layers. See https://docs.mapbox.com/mapbox-gl-js/example/image-on-a-map/.  

These points can be generated from the mapbox API by following the example in 
https://github.com/mapbox/mapbox-gl-js/issues/2375#issuecomment-204297638.

Question:
Could `coordinates` be something that is added only to `relayout` data, or would it need to be a new figure property as well?  The reason it's not clear to me that it should be a property is that it would be read-only.  I don't think we could really support setting `coordinates` as a way to set where to place the viewport (at least in the case of non-zero `bearing` and `pitch`).

etpinard:
> Could `coordinates` be something that is added only to `relayout` data, or would it need to be a new figure property as well?

It _could_, but it would be a first. At the moment, we only include key-values of the attributes that changed during an interaction.

I think I would prefer emitting `coordinates` in a different `plotly_` event.

nicolaskruchten:
`coordinates` is derivable from `center`/`zoom`/`bearing`/`pitch` but it's much, much easier for that derivation to happen in plotly.js via the mapbox code itself. It makes sense to me to add this in the same event, as a favour to non-JS users of this data :)

alexcjohnson:
For now though `plotly_relayout` event data maps exactly to `Plotly.relayout` arguments, so in principle folks may be using this to accumulate changes that get fed back to the same or another plot later. If we add in other things that *look* like attributes, this would lead to invalid figures - mostly silently so, but confusing and potentially painful in case there's any chance we'd add attributes with these names later.

But perhaps we could add something that is clearly NOT an attribute? Like:
`_derived: {coordinates: ...}`

nicolaskruchten:
> But perhaps we could add something that is clearly NOT an attribute? Like:
_derived: {coordinates: ...}

Yes, this is the kind of thing I was thinking

etpinard:
[...]

PR #4413, opened 2019-12-05
Add coordinates of mapbox view as a derived property in plotly_relayo…

jonmmease:
Closes https://github.com/plotly/plotly.js/issues/4399, following @alexcjohnson 's suggestion from https://github.com/plotly/plotly.js/issues/4399#issuecomment-561215028.

### DIFF
--- a/src/plots/mapbox/mapbox.js
+++ b/src/plots/mapbox/mapbox.js
@@ -135,7 +135,8 @@ proto.createMap = function(calcData, fullLayout, resolve, reject) {
     Promise.all(promises).then(function() {
         self.fillBelowLookup(calcData, fullLayout);
         self.updateData(calcData);
-        self.updateLayout(fullLayout, true);
+        self.updateLayout(fullLayout);
+        self.updateDerived();
         self.resolveOnRender(resolve);
     }).catch(reject);
 };
@@ -346,7 +347,7 @@ proto.updateData = function(calcData) {
     }
 };
 
-proto.updateLayout = function(fullLayout, initialView) {
+proto.updateLayout = function(fullLayout) {
     var map = this.map;
     var opts = fullLayout[this.id];
 
@@ -365,11 +366,10 @@ proto.updateLayout = function(fullLayout, initialView) {
     } else {
         map.scrollZoom.disable();
     }
-    if(initialView === true) {
-        // Add derived properties to viewInitial so they are included in
-        // the plotly_relayout event that is emitted when the plot is reset
-        this.viewInitial._derived = this.getView()._derived;
-    }
+};
+
+proto.updateDerived = function() {
+    this.viewInitial._derived = this.getView()._derived;
 };
 
 proto.resolveOnRender = function(resolve) {
--- a/src/plots/mapbox/mapbox.js
+++ b/src/plots/mapbox/mapbox.js
@@ -461,7 +461,7 @@ proto.initFx = function(calcData, fullLayout) {
             optsNow._input.bearing = optsNow.bearing = viewNow.bearing;
             optsNow._input.pitch = optsNow.pitch = viewNow.pitch;
 
-            gd.emit('plotly_relayout', self.getViewEdits(viewNow, true));
+            gd.emit('plotly_relayout', self.getViewEditsWithDerived(viewNow));
         }
         wheeling = false;
 
@@ -509,7 +509,7 @@ proto.initFx = function(calcData, fullLayout) {
 
     function emitUpdate() {
         var viewNow = self.getView();
-        gd.emit('plotly_relayouting', self.getViewEdits(viewNow, true));
+        gd.emit('plotly_relayouting', self.getViewEditsWithDerived(viewNow));
     }
 
     map.on('drag', emitUpdate);
@@ -532,7 +532,7 @@ proto.initFx = function(calcData, fullLayout) {
         optsNow._input.pitch = optsNow.pitch = viewNow.pitch;
 
         gd.emit('plotly_doubleclick', null);
-        gd.emit('plotly_relayout', self.getViewEdits(viewNow, true));
+        gd.emit('plotly_relayout', self.getViewEditsWithDerived(viewNow));
     });
 
     // define event handlers on map creation, to keep one ref per map,
@@ -771,12 +771,9 @@ proto.getView = function() {
     };
 };
 
-proto.getViewEdits = function(cont, derived) {
+proto.getViewEdits = function(cont) {
     var id = this.id;
     var keys = ['center', 'zoom', 'bearing', 'pitch'];
-    if(derived === true) {
-        keys.push('_derived');
-    }
     var obj = {};
 
     for(var i = 0; i < keys.length; i++) {
@@ -784,6 +781,13 @@ proto.getViewEdits = function(cont, derived) {
         obj[id + '.' + k] = cont[k];
     }
 
+    return obj;
+};
+
+proto.getViewEditsWithDerived = function(cont) {
+    var id = this.id;
+    var obj = this.getViewEdits(cont);
+    obj[id + '._derived'] = cont._derived;
     return obj;
 };
 
--- a/src/plots/mapbox/mapbox.js
+++ b/src/plots/mapbox/mapbox.js
@@ -136,7 +136,6 @@ proto.createMap = function(calcData, fullLayout, resolve, reject) {
         self.fillBelowLookup(calcData, fullLayout);
         self.updateData(calcData);
         self.updateLayout(fullLayout);
-        self.updateDerived();
         self.resolveOnRender(resolve);
     }).catch(reject);
 };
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/mapbox_test.js
```diff
@@ -1226,11 +1226,18 @@ describe('@noCI, mapbox plots', function() {
             expect(layout.zoom).toBeCloseTo(zoom);
         }
 
-        function _assert(center, zoom) {
+        function _assert(center, zoom, lonMin, latMin, lonMax, latMax) {
```
archmoj: `lonMin` could sometimes be greater than `lonMax`.
What about using `lon1`, `lon2`, `lat1` and `lat2` here?

test/jasmine/tests/mapbox_test.js
```diff
@@ -1226,11 +1226,18 @@ describe('@noCI, mapbox plots', function() {
...
             _assertLayout(center, zoom);
 
             expect([evtData['mapbox.center'].lon, evtData['mapbox.center'].lat]).toBeCloseToArray(center);
             expect(evtData['mapbox.zoom']).toBeCloseTo(zoom);
+            expect(evtData['mapbox._derived']).toEqual({
+                coordinates: [
+                    [lonMin, latMax],
+                    [lonMax, latMax],
+                    [lonMax, latMin],
+                    [lonMin, latMin]
```
archmoj: Although mapbox for the moment does not support tiles in various coordinates system, I could imagine we may want to make this more general: https://github.com/archmoj/grib2class/blob/a4730169516fc532298385b0115d7a254e50d3f8/grib2class.js#L506-L536
So for latitude/longitude system I was wondering if we could return something like: 
```js
grid: {
  type: 'Latitude/longitude',
  lat1: 
  lat2:
  lon1:
  lon2:
}
```

test/jasmine/tests/mapbox_test.js
jonmmease: :+1:

test/jasmine/tests/mapbox_test.js
jonmmease: I went with the 2D array to match the expected form of the `coordinates` property in the mapbox image layer: https://docs.mapbox.com/mapbox-gl-js/example/image-on-a-map/

Note that the coordinates of all 4 corners of the viewport are required in general because in the presence of `bearing` and `pitch` the viewport can be a trapezoid.

That said, I'm not opposed to having some sort of `type` or `units` property in there if that makes sense to you.

archmoj: @jonmmease Thanks for the info. 
Now it make complete sense to me to keep it the way it is.
Would you be interested to adding tests to cover cases with pitch and bearing?
Also if you could possibly add one test for a scanrio where lon1 > lon2, that would be fantastic.

jonmmease: From what I can tell, this logic will never produce a case where lon1 > lon2.  This is because mapbox will report longitude values less than zero and/or greater than 360 so that there is no discontinuity within the current viewport.  Because of that, I think the lonMin/lonMax names above are accurate.

For testing pitch and bearing, I suppose we could set up a test case that simulates a drag using the right mouse button. But I guess I don't personally see a lot of value in that since none of the plotly.js logic here depends on pitch or bearing. Do you see it differently?

src/plots/mapbox/mapbox.js
```diff
@@ -365,6 +365,11 @@ proto.updateLayout = function(fullLayout) {
     } else {
         map.scrollZoom.disable();
     }
+    if(initialView === true) {
```
etpinard: Hmm. Can you move this new block to the `proto.createMap` method instead of making `updateLayout` take two different flavors?

src/plots/mapbox/mapbox.js
```diff
@@ -747,17 +752,32 @@ proto.getView = function() {
...
     };
 };
 
-proto.getViewEdits = function(cont) {
+proto.getViewEdits = function(cont, derived) {
     var id = this.id;
     var keys = ['center', 'zoom', 'bearing', 'pitch'];
+    if(derived === true) {
+        keys.push('_derived');
+    }
```
etpinard: I'm not a fan of adding boolean arguments to functions. 

By making

```js
getViewEdits({}, true)
```

you don't know what the `true` mean until you scroll down to the `getViewEdits` definition.

I would prefer turning `getViewEdits({}, true)` into `getViewEdits({}, {derived: true})` or split `getViewEdits` into a version w/o the `_derived` container and one (e.g. named `getViewEditsWithDerived`) with it.

src/plots/mapbox/mapbox.js
```diff
@@ -747,17 +752,32 @@ proto.getView = function() {
     var mapCenter = map.getCenter();
     var center = { lon: mapCenter.lng, lat: mapCenter.lat };
 
+    var canvas = map.getCanvas();
+    var w = canvas.width;
+    var h = canvas.height;
+    var cUL = map.unproject([0, 0]).toArray();
+    var cUR = map.unproject([w, 0]).toArray();
+    var cLR = map.unproject([w, h]).toArray();
+    var cLL = map.unproject([0, h]).toArray();
```
archmoj: What about renaming `cLL` to `cBL` here?
Or you could simply use `c00`, `c01`, `c10`, c11` for the variable names.

archmoj: Or even better (IMO) is to simply to return without declaring extra variables and instead add inline comments
```js
    return {
        center: center,
        zoom: map.getZoom(),
        bearing: map.getBearing(),
        pitch: map.getPitch(),
        _derived: {
            coordinates: [
                map.unproject([0, 0]).toArray(), // top left
                map.unproject([w, 0]).toArray(), // top right
                map.unproject([w, h]).toArray(), // bottom right
                map.unproject([0, h]).toArray()  // bottom left
            ]
        }
    };
```

test/jasmine/tests/mapbox_test.js
archmoj: > From what I can tell, this logic will never produce a case where lon1 > lon2. This is because mapbox will report longitude values less than zero and/or greater than 360 so that there is no discontinuity within the current viewport. Because of that, I think the lonMin/lonMax names above are accurate.

I still think that by changing the `bearing` (e.g. to 180) one could get `lat1 > lat2` and `lon1 > lon2` at any location.
Am I missing something?

jonmmease: Oh, good point. Changed names in 4b2623fda84e818e600c9539ea45b3ce80b558ba.

test/jasmine/tests/mapbox_test.js
jonmmease: Done in 4b2623fda84e818e600c9539ea45b3ce80b558ba

src/plots/mapbox/mapbox.js
jonmmease: I pulled this out into a separate `updateDerived` method in 0ed7584c5f07ca1e9e2c8c8c8eedcb684b686138.

src/plots/mapbox/mapbox.js
jonmmease: I went with `getViewEditsWithDerived` in 45ddd806fd4679486c71dd52ebc92c6af5079688.

src/plots/mapbox/mapbox.js
jonmmease: Yeah, I agree. Inlining looks better than introducing awkward variable names in this case. Done in cacf1e9a8c6df92368b1b3c2d9f9f8a6f91445aa.

jonmmease:
Thanks very much for the review @archmoj and @etpinard. I believe I made the requested updates.  Anything else?

jonmmease:
Oh, I just noticed that these changes were resulting in a "unrecognized GUI edit" warning on the `_derived` property when resetting the plot with the modebar button.  I added a check to prevent this in https://github.com/plotly/plotly.js/pull/4413/commits/cf2c4fc46b6fd87f3900af79011c179beb7da2cf, but let me know if you see a more elegent way to do this.

etpinard:
> Oh, I just noticed that these changes were resulting in a "unrecognized GUI edit" warning on the `_derived` property when resetting the plot with the modebar button

Ideally, `'mapbox._derived.coordinates[i][j]` keys should never slip into the `fullLayout._preGUI` object. So yeah we could probably do better than your attempt in https://github.com/plotly/plotly.js/pull/4413/commits/cf2c4fc46b6fd87f3900af79011c179beb7da2cf. 

Looks like you were successful at that on user pan and scroll-zoom: `fullLayout._preGUI` only contains mapbox `center`, `zoom`, `bearing` and `pitch` keys after those.

 The logic for the modebar "reset" button is here:

https://github.com/plotly/plotly.js/blob/0be6afd71947717cb3efe5e34da2ee8caa5c3187/src/components/modebar/buttons.js#L626-L636

it relies on `viewInitial` - which does get filled with `_derived` in your new `updateDerived` mapbox subplot instance method. So I see two ways of fixing the problem,

- not stash `_dervied` values key-values into `viewInitial` (right? do we really need to stash them?)
- have the "reset" modebar button handler skip over `_` keys in `viewInitial`

jonmmease:
Thank for taking a look @etpinard, I went ahead and removed `_derived` from `viewInitial` since this wasn't needed after all.

etpinard:
:dancer: thanks @jonmmease !!

jonmmease:
Thanks! merging

Issue #4414, opened 2019-12-06
pie should handle case of zero total values


archmoj:
See console error: https://codepen.io/MojtabaSamimi/pen/BayoLgj
Also happens for `funnelarea`.

etpinard:
Nice find!

PR #4416, opened 2019-12-07
Handle invalid values and zero totals for pie and funnelarea

archmoj:
Fixes https://github.com/plotly/plotly.js/issues/4414 

[Before](https://codepen.io/MojtabaSamimi/pen/BayoLgj)
[After](https://codepen.io/MojtabaSamimi/pen/LYEpjoX)

@plotly/plotly_js


src/traces/pie/calc.js
```diff
@@ -25,26 +24,26 @@ function calc(gd, trace) {
...
-    var hasVals = isArrayOrTypedArray(vals) && vals.length;
+    var len = trace._length;
+    var hasValues = trace._hasValues && len;
 
     var i, pt;
 
     if(trace.dlabel) {
-        labels = new Array(vals.length);
-        for(i = 0; i < vals.length; i++) {
+        labels = [];
```
etpinard: What's wrong with `labels = new Array(len);`. I'm not a fan an implicit array-pushing e.g.

```js
var arr = [];
for(var i = 0; i < N; i++) {
  arr[i] = i;
}

like you have here.

File: src/traces/pie/defaults.js [hunk 2]
```diff
@@ -13,21 +13,52 @@ var attributes = require('./attributes');
...
+
+    if(!isFinite(len)) len = 0;
+
+    if(len && hasValues) {
+        var sum = 0;
+        for(var i = 0; i < len; i++) {
+            var v = +values[i];
+            if(v < 0) {
+                sum = 0;
+                break;
```
etpinard: Hmm. Don't you mean `continue;` here instead of `break;` ?

If I understand correctly, this makes any trace with one (or more) negative values has `_length=0` ?

src/traces/pie/defaults.js
```diff
@@ -13,21 +13,52 @@ var attributes = require('./attributes');
...
+            }
+            sum += v;
+        }
+        if(!sum) len = 0;
+    }
```
etpinard: Moreover, you're potentially looping here over all the `values` items. This is something we usually defer to the calc step to keep `supplyDefaults` performance scaling as the number of attributes NOT the number of data array items. 

As `pie-like` trace usually don't have that many data array items, this is not a big deal, but still could you move some of this logic to the calc step?

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4416/commits/1967fd459ef07a8e09edc52a997d74539bd9914a.

src/traces/pie/defaults.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4416/commits/1967fd459ef07a8e09edc52a997d74539bd9914a.

src/traces/pie/calc.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4416/commits/1967fd459ef07a8e09edc52a997d74539bd9914a.

etpinard:
Nicely done - :dancer: !

PR #4418, opened 2019-12-08
Mapbox react updates

jonmmease:
### Overview
This PR makes two changes to make updates to mapbox-based plots using `react` smoother. The use-case I'm working on is a Dash app that uses Datashader to update an image layer in response to each `plotly_relayout` event.

Note: These this branch was initially based on https://github.com/plotly/plotly.js/pull/4413

#### Change 1: Don't update map position while panning/zooming.

36a2650 prevents plotly.js from updating the mapbox view while a pan/zoom is in process.  This fixes the stuttering that occurs when quickly repeating zoom or pan actions.

Before:
pan_before

After:
pan_after

#### Change 2: Use `updateImage` to update image, rather than delete and recreate layer.

When possible, 7e355d7 uses the mapbox [`updateImage`](https://docs.mapbox.com/mapbox-gl-js/api/#map#updateimage) function to update an existing image in place rather than removing and recreating the source and layer.  It also adds the `raster-fade-duration: 0` paint option (which is suggested by the `updateImage` docstring). Together, these changes remove the image flashing that previously occurred between image updates.

Before:
zoom_before

After:
zoom_after


src/plots/mapbox/mapbox.js
```diff
@@ -460,18 +462,21 @@ proto.initFx = function(calcData, fullLayout) {
             optsNow._input.zoom = optsNow.zoom = viewNow.zoom;
             optsNow._input.bearing = optsNow.bearing = viewNow.bearing;
             optsNow._input.pitch = optsNow.pitch = viewNow.pitch;
-
             gd.emit('plotly_relayout', self.getViewEditsWithDerived(viewNow));
         }
-        wheeling = false;
+        if(evt.originalEvent && evt.originalEvent.type === 'mouseup') {
```
etpinard: Does this cover the case where one starts dragging on a mapbox subplot and continues into the graph's margins?

src/plots/mapbox/mapbox.js
```diff
@@ -351,10 +353,12 @@ proto.updateLayout = function(fullLayout) {
...
-    map.setCenter(convertCenter(opts.center));
-    map.setZoom(opts.zoom);
-    map.setBearing(opts.bearing);
-    map.setPitch(opts.pitch);
+    if(!this.dragging && !this.wheeling) {
+        map.setCenter(convertCenter(opts.center));
+        map.setZoom(opts.zoom);
+        map.setBearing(opts.bearing);
+        map.setPitch(opts.pitch);
+    }
```
etpinard: Very interesting solution. I think this should work. Would you be interested in trying to write a few tests to lock this down?

src/plots/mapbox/layers.js
```diff
@@ -70,6 +84,13 @@ proto.needsNewLayer = function(opts) {
     );
 };
 
+proto.updateImage = function(opts) {
+    var map = this.subplot.map;
+    map.getSource(this.idSource).updateImage({
```
etpinard: This is absolutely brilliant :sparkles: 

Thanks very much! Let's get a test case in to lock down this beautiful optimization.

etpinard:
> Note: These this branch is based on #4413, so I'll need to rebase on master after that one is merged.

That's ok. But next time, please base your PR off your base branch. Here at plotly.js, we're not fans of rebasing and `git push -f` on branches associated with open PRs.

src/plots/mapbox/mapbox.js
jonmmease: Yes, this event is triggered even when the drag ends outside the window

jonmmease:
Thanks for taking a look @etpinard, I'll start looking at adding tests next.

src/plots/mapbox/layers.js
jonmmease: Test added in https://github.com/plotly/plotly.js/pull/4418/commits/4158916421deb440690dd3379a2783419a95112c

src/plots/mapbox/mapbox.js
jonmmease: Tests added in https://github.com/plotly/plotly.js/pull/4418/commits/4f0c5e7a526b3c9c4b800b4d301f2377b3203e9b

jonmmease:
I added a few tests that cover the new functionality

Also, it looks like there are some unrelated failures in the image test suite for some webgl chart types.

etpinard:
Awesome work @jonmmease !!

:dancer: :dancer: :dancer:

jonmmease:
Thanks for taking the time @etpinard! merging

PR #4420, opened 2019-12-12
Consistent text mode for bar-like & pie-like traces and feature to control text orientation inside pie/sunburst slices

archmoj:
Resolves https://github.com/plotly/plotly.js/issues/4247 and
resolves https://github.com/plotly/plotly.js/issues/3590

In addition to simplifying (e.g. removing extra translations) and unifying transformations between bar-like and pie-like traces,  this PR adds a new `layout` attribute called `uniformtext` with `mode` (false | 'hide' | 'show'). By enabling this option the text within traces of the same type (namely `bar`, `waterfall`, `funnel`, `funnelarea`, `pie`, `sunburst` and `treemap`) would be filtered and displayed with an identical text size greater or equal than the value defined by the `minsize`.

Also this PR adds a new attribute called `insidetextorientation` to `pie` and `sunburst` traces which allows more control over the orientation of text within the slices.

Keeping text consistent during transitions is not within the scope of this PR; but it may be addressed by a separated PR.

[simple demo](https://codepen.io/MojtabaSamimi/pen/zYxBLJe)
[sunburst text orientation demo](https://codepen.io/MojtabaSamimi/pen/MWYpooK)

@plotly/plotly_js 
cc: @nicolaskruchten


src/plots/layout_attributes.js
```diff
@@ -140,6 +140,36 @@ module.exports = {
...
+            valType: 'enumerated',
+            values: [false, 'hide', 'show'],
+            dflt: false,
+            role: 'info',
+            editType: 'plot',
+            description: [
+                'Determines how the font size for various text',
+                'elements are uniformed between each trace type.',
+                'If the computed text sizes were smaller than',
+                'the minimum size defined by `minsize`',
```
etpinard: [..] by `uniformtext.minsize`

src/plots/layout_attributes.js
```diff
@@ -140,6 +140,36 @@ module.exports = {
...
+                'using *hide* option hides the text; and',
+                'using *show* option shows the text without further downscaling.',
+                'Please note that if the size defined by `minsize` is greater than',
+                'the font size defined by trace, the `minsize` would be used.'
```
etpinard: [..] then the `minsize` _is_ used.

src/plots/layout_attributes.js
```diff
@@ -140,6 +140,36 @@ module.exports = {
...
+            ].join(' ')
+        },
+        minsize: {
+            valType: 'number',
+            min: 0,
```
etpinard: Hmm. I thought we agreed on allowing `minsize: 0` and making `minsize`  have `min: -1`

src/lib/index.js
```diff
@@ -1182,31 +1182,30 @@ lib.isHidden = function(gd) {
     return !display || display === 'none';
 };
 
-lib.getTextTransform = function(opts) {
+lib.getTextTransform = function(opts, isCenter) {
```
etpinard: Why not place the `isCenter` option inside the `opts` object? I'm not a fan of generalising existing functions by adding boolean positional arguments, they make it hard to understand what the functions does when just looking at the caller.

---

Otherwise, thanks very much for cleaning this up. I see that you got a few more traces to reuse this routine - nice job!

src/traces/bar/style.js
```diff
@@ -20,8 +20,46 @@ var attributeInsideTextFont = attributes.insidetextfont;
...
+        );
+
+        t.each(function(d) {
+            var transform = d.transform;
+
+            transform.scale = minSize / transform.fontSize;
+            d3.select(this).attr('transform', Lib.getTextTransform(transform, isCenter));
+
+            if(shouldHide && transform.hide) {
+                d3.select(this).text(' ');
```
etpinard: Hmm. It's probably best to use `d3.select(this).attr('display', 'none')` to hide text elements. That way we could have something like:

```js
d3.select(this).attr('display', shouldHide && transform.hide ? 'none' : null);
```

similar to 

https://github.com/plotly/plotly.js/blob/b2102a504a480493d9e474e15489c4874d5f6bac/src/components/drawing/index.js#L98-L101

making the hide/show logic independent of `_module.plot`.

src/traces/bar/style.js
```diff
@@ -57,7 +95,9 @@ function stylePoints(sel, trace, gd) {
 function styleTextPoints(sel, trace, gd) {
     sel.selectAll('text').each(function(d) {
         var tx = d3.select(this);
-        var font = determineFont(tx, d, trace, gd);
+        var font = Lib.extendFlat({}, determineFont(tx, d, trace, gd), {});
```
etpinard: This looks like a typo. 

```js
var font = determineFont(tx, d, trace, gd);
```

should be good enough, right? If not then, we should move the `extendFlat` call to `determineFont`.

src/traces/funnelarea/plot.js
```diff
@@ -78,10 +80,10 @@ module.exports = function plot(gd, cdModule) {
                 slicePath.attr('d', shape);
 
                 // add text
-                piePlot.formatSliceLabel(gd, pt, cd0);
+                formatSliceLabel(gd, pt, cd0);
                 var textPosition = pieHelpers.castOption(trace.textposition, pt.pts);
                 var sliceTextGroup = sliceTop.selectAll('g.slicetext')
-                    .data(pt.text && (textPosition !== 'none') ? [0] : []);
+                    .data(pt.text && (textPosition !== 'none') ? [pt] : []);
```
etpinard: Why did `[0]` become `[pt]` ?

src/traces/funnelarea/plot.js
```diff
@@ -94,36 +96,41 @@ module.exports = function plot(gd, cdModule) {
...
                     var x0, x1;
-                    var y0 = Math.min(pt.BL[1], pt.BR[1]);
-                    var y1 = Math.max(pt.TL[1], pt.TR[1]);
+                    var y0 = Math.min(pt.BL[1], pt.BR[1]) + cy;
+                    var y1 = Math.max(pt.TL[1], pt.TR[1]) + cy;
 
-                    x0 = Math.max(pt.TL[0], pt.BL[0]);
-                    x1 = Math.min(pt.TR[0], pt.BR[0]);
+                    x0 = Math.max(pt.TL[0], pt.BL[0]) + cx;
+                    x1 = Math.min(pt.TR[0], pt.BR[0]) + cx;
```
etpinard: These changes are pretty subtle. Can you explain why you needed to that?

src/traces/treemap/draw_ancestors.js
```diff
@@ -141,23 +141,31 @@ module.exports = function drawAncestors(gd, cd, entry, slices, opts) {
             s.attr('data-notex', 1);
         });
 
+        var font = Lib.extendFlat({}, helpers.determineTextFont(trace, pt, fullLayout.font, trace.pathdir), {});
```
etpinard: Here too. Shouldn't

```js
var font = helpers.determineTextFont(trace, pt, fullLayout.font, trace.pathdir);
```

be enough?

src/traces/treemap/draw_descendants.js
```diff
@@ -182,16 +182,20 @@ module.exports = function drawDescendants(gd, cd, entry, slices, opts) {
             s.attr('data-notex', 1);
         });
 
+        var font = Lib.extendFlat({}, helpers.determineTextFont(trace, pt, fullLayout.font), {});
```
etpinard: (same as :arrow_up: )

test/jasmine/tests/treemap_test.js
```diff
@@ -1395,8 +1395,8 @@ describe('Test treemap tweening:', function() {
             _assert('update b to new position', 'd', 'b',
                 'M221.75,136L611,136L611,361L221.75,361Z'
             );
-            _assert('move B text to new position', 'transform', 'B', [221.75126, 0]);
-            _assert('move b text to new position', 'transform', 'b', [224.75150, 0]);
+            _assert('move B text to new position', 'transform', 'B', [222.75, 126]);
+            _assert('move b text to new position', 'transform', 'b', [225.75, 150]);
```
etpinard: Hmm. I'm getting these new jasmine test failures locally:

image

Is that due to some local vs CI discrepancy?

src/traces/pie/attributes.js
```diff
@@ -181,6 +181,21 @@ module.exports = {
     textfont: extendFlat({}, textFontAttrs, {
         description: 'Sets the font used for `textinfo`.'
     }),
+    insidetextorientation: {
+        valType: 'enumerated',
+        role: 'info',
+        values: ['h', 'r', 't', 'auto'],
```
etpinard: Hmm. I'd vote for `'horizontal', 'rotational'`, `'tangential'`, `'auto'`. I see your point for `'h'` making it on-par with the typical `orientation` values, but values `'r'` and `'t'` look pretty obscure to me.

src/traces/pie/defaults.js
```diff
@@ -70,6 +70,10 @@ module.exports = function supplyDefaults(traceIn, traceOut, defaultColor, layout
         if(hasOutside) {
             coerce('automargin');
         }
+
+        if(textposition !== 'none' && textposition !== 'outside') {
```
etpinard: Wouldn't `textposition === 'inside' || textposition === 'auto'` be more readable?

src/traces/pie/plot.js
```diff
@@ -144,15 +145,19 @@ function plot(gd, cdModule) {
                         s.attr('data-notex', 1);
                     });
 
+                    var font = Lib.extendFlat({}, textPosition === 'outside' ?
+                        determineOutsideTextFont(trace, pt, fullLayout.font) :
+                        determineInsideTextFont(trace, pt, fullLayout.font), {}
+                    );
```
etpinard: One more strange `extendFlat` call comment,

src/traces/pie/plot.js
```diff
@@ -164,36 +169,39 @@ function plot(gd, cdModule) {
...
                         transform = transformInsideText(textBB, pt, cd0);
                         if(textPosition === 'auto' && transform.scale < 1) {
-                            sliceText.call(Drawing.font, trace.outsidetextfont);
-                            if(trace.outsidetextfont.family !== trace.insidetextfont.family ||
-                                    trace.outsidetextfont.size !== trace.insidetextfont.size) {
+                            var newFont = Lib.extendFlat({}, trace.outsidetextfont, {});
+                            newFont.size = Math.max(newFont.size, fullLayout.uniformtext.minsize || 0);
+
+                            sliceText.call(Drawing.font, newFont);
+                            if(newFont.family !== font.family || newFont.size !== font.size) {
```
etpinard: We shouldn't need this check. The caching mechanism in `Drawing.bBox` should be able to catche cases where `sliceText` has the same font family and size as in the previous call.

src/traces/sunburst/plot.js
```diff
@@ -148,8 +150,9 @@ function plotOne(gd, cd, element, transitionOpts) {
     // slice path generation fn
     var pathSlice = function(d) { return Lib.pathAnnulus(d.rpx0, d.rpx1, d.x0, d.x1, cx, cy); };
     // slice text translate x/y
-    var transTextX = function(d) { return cx + d.pxmid[0] * d.transform.rCenter + (d.transform.x || 0); };
-    var transTextY = function(d) { return cy + d.pxmid[1] * d.transform.rCenter + (d.transform.y || 0); };
+
+    var getTargetX = function(d) { return cx + (d.pxtxt || d.pxmid)[0] * d.transform.rCenter + (d.transform.x || 0); };
+    var getTargetY = function(d) { return cy + (d.pxtxt || d.pxmid)[1] * d.transform.rCenter + (d.transform.y || 0); };
```
etpinard: What's this new `d.pxtxt` for?

src/traces/sunburst/plot.js
```diff
@@ -248,26 +253,29 @@ function plotOne(gd, cd, element, transitionOpts) {
             s.attr('data-notex', 1);
         });
 
+        var font = Lib.extendFlat({}, helpers.determineTextFont(trace, pt, fullLayout.font), {});
```
etpinard: (duplicate comment about strange `Lib.extendFlat` call)

src/traces/funnelarea/plot.js
archmoj: The text elements here used to have two `translate` calls. We need to reduce that to only one for keeping track of actual translation (in the transform object).

src/traces/bar/style.js
archmoj: A copy is needed since the font.size may be replaced by `minsize`.

etpinard: Ok that explains the

```js
 var font = Lib.extendFlat({}, determineFont(tx, d, trace, gd));
```

part, but isn't third argument in 

```js
var font = Lib.extendFlat({}, determineFont(tx, d, trace, gd), {});
```

useless?

src/traces/funnelarea/plot.js
archmoj: To have access to text `tranform`.

test/jasmine/tests/treemap_test.js
archmoj: Strange. They pass on my machine as well as CI.

etpinard:
@archmoj I still need to try to understand your new `insidetextorientation` logic and explore all the details you're locking down in the test mocks, but nonetheless here's a first review.

Writing what we discussed a week ago: in a not-so-distant future (i.e. whenever we'll attempt to implement https://github.com/plotly/plotly.js/issues/2000) we should try to compute the bounding boxes of all the trace text items **before** plotting them. In brief, this means adding logic to the `calc` or `crossTraceCalc` step - which would unfortunately make them async in general to support MathJax (unless we decide to ditch MathJax for e.g. KaTeX before doing this). Not having to resize the text elements during `style` would essentially eliminate the transition problems you're referring to in the PR header (example here: https://codepen.io/MojtabaSamimi/pen/NWPrzOG?editors=1000). Personally, these transition problems are minor enough at the moment that I'd vote keeping them as is until we tweak the pipeline.

----

Before thinking about merging this PR, I'd like to see some jasmine tests to see how things behave when calling <strike>`relayout`</strike> (sorry) `react` with changes in the `uniformtext.*` attributes.

src/traces/pie/attributes.js
archmoj: `rotational` or `radial`?
Also to note something like `insidetextorientation: 'tangential'` is hard to read & write.

src/traces/pie/defaults.js
archmoj: Good point.

src/traces/pie/plot.js
archmoj: Good call. I'll give it a try.

etpinard:
Moreover, @archmoj could you share a JSON mock or a codepen of a _simple_ case  that uses `uniformtext`? Your test mocks are great for locking multiple things down in one go, but they are a little hard grasp for reviewers.

src/traces/sunburst/plot.js
archmoj: When fitting horizontal text in `pie` and `sunburst` the best position may not be the middle point of the arc. So `pxtxt` keeps the position of text if it is moved form the middle.

uniformtext_pie_8_horizontal

sunburst

nicolaskruchten: these are really nice... loving the consistent alignment of labels in the sunburst :)

nicolaskruchten: (well, semi-consistent!) We should maybe special-case the noon and 6 o'clock positions as 'preferred' if they result in the same size as 12:05 or 6:15, say :)

archmoj:
> Moreover, @archmoj could you share a JSON mock or a codepen of a _simple_ case that uses `uniformtext`? Your test mocks are great for locking multiple things down in one go, but they are a little hard grasp for reviewers.

Here is a [simple demo](https://codepen.io/MojtabaSamimi/pen/zYxBLJe) which also added to the PR description.

src/plots/layout_attributes.js
archmoj: Disabling `uniformtext` using `mode: false` (or possibly `none` ?) made the process much simpler.
In addition we may want to add other options for unifying e.g. rotations in future; and in that sense depending on the value of `minsize` is not great.

src/traces/bar/style.js
archmoj: Correct. We could/should drop that.
I can make a centralize function for those font getters.
Should `uniform text.minsize` override the font.size(s) defined by traces? If not we could use the minimum among them.

src/traces/funnelarea/plot.js
etpinard: Brilliant - thanks!

src/lib/index.js
etpinard: ... oh can you add a jsDoc comment on top of this function? I have a feeling we'll be reusing this function in the future, so it would be nice to document its arguments and behaviour.

src/traces/bar/style.js
etpinard: > Should uniform text.minsize override the font.size(s) defined by traces?

Like override them in `gd.data[i]` and `gd._fullData[i]`? If so, then the answer is NO. We shouldn't mutate that. 



> If not we could use the minimum among them.

That sounds about right. Adding something like

```js
Lib.applyUniformtext = function(gd, baseFont) {
  var out = Lib.extendFlat({}, baseFont);
  out.size = Math.max(baseFont.size, gd._fullLayout.uniformtext.minsize || 0);
  return out;
}
```

could be a big win!

src/traces/pie/attributes.js
etpinard: yep, I meant to type `'radial'`. Edited above.

src/traces/bar/style.js
archmoj: Thanks for the hint.
Would you please clarify which size should be used in the scenario below?
When one of the traces defined its font size to be e.g. 8; while the uniform.minsize is set to 16.
In the current implementation 16 wins over 8.

src/plots/layout_attributes.js
nicolaskruchten: what would `-1` mean here?

archmoj: Right now. Nothing. Any value less than zero is not allowed as treated as default zero.

src/traces/pie/plot.js
```diff
@@ -164,36 +169,39 @@ function plot(gd, cdModule) {
...
-                        'translate(' + translateX + ',' + translateY + ')' +
-                        (transform.scale < 1 ? ('scale(' + transform.scale + ')') : '') +
-                        (transform.rotate ? ('rotate(' + transform.rotate + ')') : '') +
-                        'translate(' +
-                            (-(textBB.left + textBB.right) / 2) + ',' +
-                            (-(textBB.top + textBB.bottom) / 2) +
-                        ')');
+                    transform.fontSize = font.size;
+                    recordMinTextSize(trace.type, transform, fullLayout);
+                    cd[i].transform = transform;
```
etpinard: Why do we need to stash the transform info in the calcdata?

src/traces/funnelarea/plot.js
etpinard: Where? You have

```
sliceTextGroup.each(function() {
  // ...
})
```

 below, not extracting the `pt` object.

src/traces/pie/plot.js
archmoj: In `resizeText` we need to have access to translate, scale and rotation raw values without parsing the style of the element.

src/traces/bar/style.js
```diff
@@ -20,8 +20,46 @@ var attributeInsideTextFont = attributes.insidetextfont;
...
+                t = gTrace.selectAll('g.points').selectAll('g.point').selectAll('text');
+        }
+
+        var isCenter = (
+            traceType === 'pie' ||
+            traceType === 'sunburst'
+        );
+
+        t.each(function(d) {
+            var transform = d.transform;
```
archmoj: @etpinard here we read transform data from the calc data.

src/traces/funnelarea/plot.js
archmoj: See https://github.com/plotly/plotly.js/pull/4420/files#r357786862

src/traces/pie/plot.js
archmoj: See https://github.com/plotly/plotly.js/pull/4420/files#r357786862

src/traces/pie/plot.js
```diff
@@ -549,59 +559,126 @@ function prerenderTitles(cdModule, gd) {
...
+            rotate: 0
+        };
 
-    // max size text can be inserted inside without rotating it
-    // this inscribes the text rectangle in a circle, which is then inscribed
-    // in the slice, so it will be an underestimate, which some day we may want
-    // to improve so this case can get more use
-    var transform = {
-        scale: rInscribed * r * 2 / textDiameter,
+        if(transform.scale >= 1) return transform;
```
etpinard: Hmm. That's an odd early return. When does this happen?

archmoj: That's from the current logic. 
Which means if the horizontal text fits at the middle, then no need to try radial and tengential.

src/traces/pie/plot.js
```diff
@@ -549,59 +559,126 @@ function prerenderTitles(cdModule, gd) {
...
+
+                var closestEdge = dStart < dStop ? dStart : dStop;
+
+                var newT;
+                if(key === 'tan') {
+                    newT = calcTanTransform(textBB, r, ring, closestEdge, 0);
+                } else { // case of 'rad'
+                    newT = calcRadTransform(textBB, r, ring, closestEdge, Math.PI / 2);
+                }
+                newT._repos = getCoords(r, angle);
```
etpinard: What does `_repos` stand for?

archmoj: `Re-position` text if it is not at the center of the arc.

src/traces/pie/plot.js
```diff
@@ -549,59 +559,126 @@ function prerenderTitles(cdModule, gd) {
...
+
+                allTransforms.push(newT);
+            }
+        };
+
+        for(var i = 3; i >= -3; i--) { // to cover all cases with trace.rotation added
+            considerCrossing(Math.PI * (i + 0.0), 'tan');
```
etpinard:
```js
considerCrossing(Math.PI * i, 'tan');
```

suffices. No need to convert `i` to a float in JS.

archmoj: Good eye. That one remained from the alignment of a series of calls to the function when there was not this loop. So I'd remove that.

src/traces/bar/style.js
etpinard: Right ok I see - but it would be cleaner to read in the calc data from their parent `<g>`. Propagating d3-bound data using `selectAll` calls can lead to weird problems, for example:

https://github.com/plotly/plotly.js/blob/b2102a504a480493d9e474e15489c4874d5f6bac/src/traces/scattercarpet/plot.js#L32-L34

----

So here, it would be best to have your `t` variable select the `g.points` or `g.slicetext` and here have:

```js
t.each(function(d) {
  var el = d3.select(this).select('text')

  // and then apply changes to `el`
});

File: src/plots/layout_attributes.js (hunk 3 above)
Comment (etpinard) [reply src/plots/layout_attributes.js]: Ok nevermind - I think I got confused by your previous iteration where you have `min: 1`.

Closing.

File: src/traces/pie/plot.js (hunk 18 above)
Comment (etpinard) [reply src/traces/pie/plot.js]: Ok - I get that's fine though I might be hard to keep that thing in-sync in the future. Closing.

File: src/traces/funnelarea/plot.js (hunk 7 above)
Comment (etpinard) [reply src/traces/funnelarea/plot.js]: Cross-referencing https://github.com/plotly/plotly.js/pull/4420#discussion_r357814274

File: src/traces/pie/plot.js (hunk 21 above)
Comment (etpinard) [reply src/traces/pie/plot.js]: Ok non-blocking, but since you copy over its values to `pxtxt` below, maybe it would be best to name `newT.pxtxt`.

File: src/traces/pie/defaults.js (hunk 13 above)
Comment (archmoj) [reply src/traces/pie/defaults.js]: Actually that does not work; whereas `textposition` could be also an array.

File: src/traces/bar/style.js (hunk 6 above)
Comment (etpinard) [reply src/traces/bar/style.js]: > When one of the traces defined its font size to be e.g. 8; while the uniform.minsize is set to 16.

That's a hard one. I can see arguments for both ways. On one hard, it would make things more flexible if we allow users to override the uniform-text-size results. But then again, if they really want that, they can set all the `textfont.size` in the traces manually w/o even touching `uniformtext`.

File: src/traces/pie/plot.js (hunk 20 above)
Comment (etpinard) [reply src/traces/pie/plot.js]: Ha I see. Thanks for pointing that out. The diff was hard to read.

File: src/traces/pie/defaults.js (hunk 13 above)
Comment (etpinard) [reply src/traces/pie/defaults.js]: Ok well, I'd vote for:

```js
if(textposition === 'inside' || textposition === 'auto || Array.isArray(textposition)) {
  // ...
}
```

src/plots/layout_attributes.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/plots/layout_attributes.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/lib/index.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/bar/style.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/treemap/draw_ancestors.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/treemap/draw_descendants.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/attributes.js
archmoj: Changed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/defaults.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/plot.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/plot.js
archmoj: Removed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/sunburst/plot.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/plot.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/traces/pie/plot.js
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4420/commits/a27dc29ca87303fd53bf994bfe1203b3c5e96887.

src/lib/index.js
etpinard: Ok. Now could you add info about all the different keys that the function expect in the `transform` object? For example, it would be nice to shed light on difference between `textX` and `targetX`

src/traces/bar/style.js
etpinard: Much better. Thanks!!

src/traces/pie/attributes.js
etpinard: Thank you very much!

src/traces/pie/plot.js
etpinard: Great - I'm glad that worked out!

src/lib/index.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4420/commits/1479943d15bb8fa09e7a3fa913ed2f0d576ea345

src/traces/sunburst/plot.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4420/commits/1479943d15bb8fa09e7a3fa913ed2f0d576ea345.

src/traces/bar/style.js
archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/4420/commits/eab205ae31606fa42f5e057f19da27d3389ca3f1.

src/lib/index.js
etpinard: Thanks!

src/traces/bar/style.js
etpinard: Very nice. Thanks! Can you now replace back the new `.data([pt])` with old `.data([0])` as pointed out in https://github.com/plotly/plotly.js/pull/4420#pullrequestreview-331517397

archmoj:
> Before thinking about merging this PR, I'd like to see some jasmine tests to see how things behave when calling `relayout` (sorry) `react` with changes in the `uniformtext.*` attributes.

Good call. React tests for `uniformtext` `mode` and `minsize` for `bar`, `watrefall`, `funnel`, `funnelarea` and `pie` as well as `insidetextorientation` for `pie` are added in https://github.com/plotly/plotly.js/pull/4420/commits/15ce24d17eb7415a5bc04ee7dafa6797db6c7c81.

src/traces/bar/style.js
archmoj: In order to access transform (for `uniformtext`) similar to other traces, this is still needed for `pie` and `funnelarea`.

etpinard: Can't you grab the `transform` object from the parent `<g>`?

archmoj: Fixed by https://github.com/plotly/plotly.js/pull/4420/commits/603e0cce4ada75443932ab415e7b0f193e60c9ae thanks to @etpinard.

etpinard: Resolved!

test/jasmine/tests/treemap_test.js
etpinard: Now fixed!!

etpinard:
Awesome work @archmoj - this PR looks good to merge :dancer: 

---

To sum up, this PR:

- adds new layout attributes `uniformtext.mode` and `uniformtext.size` which determines how the font size for various text elements are uniformed between each trace type
- add new `pie` and `sunburst` attribute `insidetextorientation` which determines the orientation of text inside slices

PR #4271, opened 2019-10-11
Correct streamtube positions and colouring

archmoj:
Fixes #4206 and fixes #4275 

The `gl-streamtube3d` module used to make assumptions about the order of input data which could be false in cases where the data cube was not filled in orders than `xyz` (i.e first `x` | next `y` | last  `z`).
This PR addresses this issue by considering the order of data in `calc` step and using this info inside `gl-streamtube3d` module.

[Before fixing 4206](https://codepen.io/MojtabaSamimi/pen/XWWmwMN)
[After fixing 4206](https://codepen.io/MojtabaSamimi/pen/mddeXqa)

[Before fixing 4275](https://codepen.io/MojtabaSamimi/pen/XWWXZRK)
[After fixing 4275](https://codepen.io/MojtabaSamimi/pen/YzzweWE)

In addition `gl-cone3d` module is patched in this PR.

Please refer to the following `gl-vis` PRs for more details on the changes within dependecies:
https://github.com/gl-vis/gl-cone3d/pull/16
https://github.com/gl-vis/gl-cone3d/pull/17
https://github.com/gl-vis/gl-cone3d/pull/18
https://github.com/gl-vis/gl-streamtube3d/pull/10
https://github.com/gl-vis/gl-streamtube3d/pull/11

@plotly/plotly_js

PR #4431, opened 2019-12-18
Handle streamtube coordinates in string format and handle different data orders in isosurface and volume

### LINKED ISSUE #4427: Streamtube problem input string numbers
This [demo](https://codepen.io/MojtabaSamimi/pen/QWwGOmG?editors=0010) works; but this [demo](https://codepen.io/MojtabaSamimi/pen/eYmBeKy?editors=0010) does not.
Basically the calc step is unable to figure out the direction of data since the values are in string format.

### LINKED ISSUE #4440: Isosurface and volume expecting certain input data order
Since `isosurface` as well as `volume` API were designed to be similar to `streamtube` they show
a problem similar to #4206.
[Demo](https://codepen.io/MojtabaSamimi/pen/mdywPxE?editors=0010) illustrates that currently `isosurface` & possibly `volume` cannot handle different orders of input grid.

archmoj:
Fixes #4427:
Whereas raw x, y and z values were used in `streamtube` calc step, in order to find the order of data, it was necessary to ensure casting to numbers before comparing two values.
[Before](https://codepen.io/MojtabaSamimi/pen/eYmBeKy?editors=0010)
[After](https://codepen.io/MojtabaSamimi/pen/LYExdqg?editors=1000)

And fixes #4440:
Similar to #4271, this PR fixed the data cube order issue so that the `value` as well as `x`, `y` and `z` arrays could be filled in different orders.
[Before](https://codepen.io/MojtabaSamimi/pen/mdywPxE?editors=0010)
[After](https://codepen.io/MojtabaSamimi/pen/oNgwzya?editors=0010)

@plotly/plotly_js


src/traces/streamtube/calc.js
```diff
@@ -25,11 +25,14 @@ module.exports = function calc(gd, trace) {
     var slen = 0;
     var startx, starty, startz;
     if(trace.starts) {
-        startx = trace.starts.x || [];
-        starty = trace.starts.y || [];
-        startz = trace.starts.z || [];
+        startx = filter(trace.starts.x || []);
```
etpinard: Wouldn't it be better to loop over all the `starts.(x|y|z)` values once and convert all the numeric strings to numbers?

etpinard: Wait, that's what `filter` does essentially.

So we shouldn't need all the `+v` calls downstream of here, right?

archmoj: When the length is known, it converts and copies the raw values.
So during the calculations for length it was needed to cast in few more locations.

etpinard: Ok, so wouldn't it be easier to always convert the raw values? This would be more similar to how we do things in other trace modules e.g. when calling `ax.makeCalcdata` .

src/traces/streamtube/calc.js
```diff
@@ -206,6 +206,9 @@ function distinctVals(col) {
 function filter(arr, len) {
     if(len === undefined) len = arr.length;
 
+    // no need for casting typed arrays to numbers
+    if(Lib.isTypedArray(arr)) return arr.slice(0, len);
```
etpinard: It's better to use `arr.subarray()` for typed arrays, as `.slice` isn't supported in older browsers.

archmoj: Good call.
Wondering if we need to slice the array at all?

archmoj: `slice` (or subarry) is needed for `distictVals` to work correctly.
But it may be better not to slice if the length is correct.

src/traces/isosurface/convert.js
```diff
@@ -154,17 +149,32 @@ function generateIsoMeshes(data) {
...
+    var Zs = data._Zs;
 
     var width = Xs.length;
     var height = Ys.length;
     var depth = Zs.length;
 
-    function getIndex(i, j, k) {
-        return k + depth * j + depth * height * i;
-    }
+    var filled = GRID_TYPES.indexOf(data._gridFill.replace(/-/g, '').replace(/\+/g, ''));
```
etpinard: In https://github.com/plotly/plotly.js/issues/4440, you mentioned that `volume` traces are affected by this bug too. So don't you need to bring this new logic over to `volume/convert.js` ?

archmoj: Good eye!

src/traces/streamtube/calc.js
```diff
@@ -165,32 +208,43 @@ module.exports = function calc(gd, trace) {
...
-        yMin = Math.min(yMin, sy);
+function distinctVals(col) {
+    return Lib.distinctVals(filter(col)).vals;
+}
 
-        var sz = startz[i];
-        zMax = Math.max(zMax, sz);
-        zMin = Math.min(zMin, sz);
-    }
+function filter(arr, len) {
```
etpinard: I think `filter` has is never called with the `len` argument.

src/traces/isosurface/convert.js
archmoj: Since `volume/convert` reused parts of `isosurface/convert`, no change was needed.

src/traces/streamtube/calc.js
```diff
@@ -165,32 +208,43 @@ module.exports = function calc(gd, trace) {
...
+        fill: gridFill
+    };
+}
 
-        var sy = starty[i];
-        yMax = Math.max(yMax, sy);
-        yMin = Math.min(yMin, sy);
+function distinctVals(col) {
+    return Lib.distinctVals(filter(col)).vals;
+}
```
etpinard: I think this should be

```js
function distinctVals(col, len) {
    return Lib.distinctVals(filter(col, len)).vals;
}
```

src/traces/isosurface/convert.js
etpinard: Cool! I don't think any of your new mocks have `volume` traces in them. Can you add a new mock to lock down the behaviour for `volume` traces?

src/traces/streamtube/calc.js
```diff
@@ -61,13 +99,18 @@ module.exports = function calc(gd, trace) {
...
-    var firstY;
-    var firstZ;
+    var firstX, lastX;
+    var firstY, lastY;
+    var firstZ, lastZ;
     if(len) {
-        firstX = x[0];
-        firstY = y[0];
-        firstZ = z[0];
+        firstX = +x[0];
```
etpinard: Again, why not convert all the x, y, z items to numbers first thing in `calc`, so that we don't have to worry about casting strings to numbers in `processGrid`, `filter` and `lessThan`

src/traces/isosurface/convert.js
archmoj: Good call. Added in https://github.com/plotly/plotly.js/pull/4431/commits/6d062292c01fd69e1c853d653d73a5ed11e9a0f7.
We also have jasmine tests for `volume`.

src/traces/streamtube/calc.js
archmoj: Good eye.
Done in https://github.com/plotly/plotly.js/pull/4431/commits/b71677a1255b52cab4dae423de4d501a6f0b4a81.

src/traces/streamtube/calc.js
etpinard: > But it may be better not to slice if the length is correct.

that could be a nice optimisation for typed arrays, but for regular arrays we'll need to slice in general to cast numerical-strings items w/o mutating.

src/traces/streamtube/calc.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4431/commits/df9b4c49e62f9fc44624e3fd5f7c37f6bafb1acd.

src/traces/isosurface/calc.js
```diff
@@ -10,9 +10,19 @@
...
+    trace._len = Math.min(
+        trace.x.length,
+        trace.y.length,
+        trace.z.length,
+        trace.value.length
+    );
+
+    ['value', 'x', 'y', 'z'].forEach(function(e) {
+        trace['_' + e] = filter(trace[e], trace._len);
+    });
```
etpinard: Ah, this is ugly. What if I want to know where `_x`, `_y`, `_z`, `_value` are set? With this here, grepping for e.g. `_x` won't help me. 

Please rewrite this as in four almost-identical lines:

```js
trace._value = filter(trace.value, trace._len);
// ...
```

src/traces/streamtube/calc.js
```diff
@@ -21,6 +21,10 @@ function calc(gd, trace) {
         trace.w.length
     );
 
+    ['u', 'v', 'w', 'x', 'y', 'z'].forEach(function(e) {
+        trace['_' + e] = filter(trace[e], trace._len);
+    });
```
etpinard: Same here. Please rewrite this as six almost-identical lines.

etpinard:
Nice job @archmoj - this PR is looking very nice now, one commit away from merging.

src/traces/isosurface/calc.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4431/commits/c82da07617b8cac581eea4a4cc291a710683d320.

src/traces/streamtube/calc.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4431/commits/c82da07617b8cac581eea4a4cc291a710683d320.

src/traces/isosurface/calc.js
etpinard: awesome!

etpinard:
Beautiful fix!! Thanks very much for bringing some sanity into streamtube-land

:dancer: :dancer: :dancer:

Issue #4430, opened 2019-12-18
reversescale doesn't reverse colorbar on contour plot with coloring='heatmap'


harupy:
### Bug

`reversescale: True` doesn't reverse the colorbar on a contour plot with `{contours: {coloring: 'heatmap'}}`.

In the plot below, the colorbar indicates the red region has larger values, but it actually has smaller values.

![plotly-issue](https://user-images.githubusercontent.com/17039389/71063572-498a9100-21b0-11ea-9f96-b4fffa88227a.gif)

### Code

```javascript
var data = [{
  z: [[10, 10.625, 12.5, 15.625, 20],
      [5.625, 6.25, 8.125, 11.25, 15.625],
      [2.5, 3.125, 5., 8.125, 12.5],
      [0.625, 1.25, 3.125, 6.25, 10.625],
      [0, 0.625, 2.5, 5.625, 10]],
  x: [-9, -6, -5 , -3, -1],
  y: [0, 1, 4, 5, 7],
  type: 'contour',
  contours: {
    coloring: 'heatmap'
  },
  reversescale: true,
}];

var layout = {
  title: 'Setting the X and Y Coordinates in a Contour Plot'
};

Plotly.newPlot('myDiv', data, layout);
```
 
### Codepen
https://codepen.io/harupy/pen/bGNgBOz

harupy:
Without `{contours: {coloring: 'heatmap'}}`, it works.

![plotly-issue-2](https://user-images.githubusercontent.com/17039389/71064532-a8510a00-21b2-11ea-944a-444014dbc0f6.gif)


### Code

```javascript
var data = [{
  z: [[10, 10.625, 12.5, 15.625, 20],
       [5.625, 6.25, 8.125, 11.25, 15.625],
       [2.5, 3.125, 5., 8.125, 12.5],
       [0.625, 1.25, 3.125, 6.25, 10.625],
       [0, 0.625, 2.5, 5.625, 10]],
  x: [-9, -6, -5 , -3, -1],
  y: [0, 1, 4, 5, 7],
  type: 'contour',
  reversescale: true,
}];

var layout = {
  title: 'Setting the X and Y Coordinates in a Contour Plot'
};

Plotly.newPlot('myDiv', data, layout);
```

### Codepen
https://codepen.io/harupy/pen/NWPddqQ

harupy:
Do we need to flip the colorscale here?
[...]

PR #4437, opened 2019-12-20
Fix colorbar of reversescale colorbars of heatmap-coloring contours

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4430

before: https://codepen.io/etpinard/pen/ZEYePxW 
after: https://codepen.io/etpinard/pen/JjoWzYb

cc @archmoj


archmoj:
Great! Could you please add a test for the case of using coloraxis as well?

etpinard:
Ok I'll add a subplot on the new mock.

But in my opinion, we don't need that as the trace-colorscale vs coloraxis logic is inside `Colorscale.extractOpts` which is already tested elsewhere.

archmoj:
If so, just testing in a codepen is enough.
Thank you!

archmoj:
Concerning `before` codepen it is better to use v.1.51.3 not the latest.

etpinard:
> Concerning `before` codepen it is better to use v.1.51.3 not the latest.

Sure, edited above.

> Great! Could you please add a test for the case of using coloraxis as well?

done in https://github.com/plotly/plotly.js/pull/4437/commits/af025a2a0bc3aaa5c99283e5d8df6f78a1bd52e5

archmoj:
:dancer: :dancer: after all green light on the CI.

Issue #4429, opened 2019-12-17
Wrong bar layout on the horizontal axis when using 'histogram' in plotly-cartesian


cebamps:
When using the `plotly-cartesian` in place of the `plotly` distribution in the latest version (1.51.3, also reproducible in 1.51.1), for some inputs the histogram bars are incoherently laid out on the horizontal axis.

Some other inputs do not seem cause that issue.

The problem does not appear when substituting `https://cdn.plot.ly/plotly-1.51.3.js`.

### Minimal reproduction of the issue
https://jsfiddle.net/c8gfb6rp/2/
```js
var x1 = [1,3];
var x2 = [1,20];

var trace1 = {
  x: x1,
  type: "histogram",
};
var trace2 = {
  x: x2,
  type: "histogram",
};
var data = [trace1, trace2];
var layout = {barmode: "stack", height: 200, width: 400, margin: {t:30, b:30}};

Plotly.plot("plot", data, layout);
```

### Screenshots
Here is what it looks like in the jsfiddle above.

Using `plotly-cartesian` (bug):
![hist-plotly-cartesian](https://user-images.githubusercontent.com/12797742/70982809-efa6a000-20b7-11ea-938c-e1a3392058c0.png)

Using `plotly` (expected behavior):
![hist-plotly](https://user-images.githubusercontent.com/12797742/70982808-efa6a000-20b7-11ea-8834-eeb0a915232e.png)

etpinard:
Thanks for reporting!

PR #4439, opened 2019-12-23
Fix histogram bingroup algo when calendars module is not registered

etpinard:
By adding a fallback in the trace calendar getter in `Histogram.crossTraceCalc` so that when the Calendars module isn't registered (thus making `_fullData[i].(x|y)calendar === undefined`), we're still able to group traces in the correct bin groups.

fixes https://github.com/plotly/plotly.js/issues/4429

- before: https://jsfiddle.net/c8gfb6rp/2/
- after: https://jsfiddle.net/Lg5ucrew/
  + w/ https://gist.github.com/etpinard/ff8d940dfb831b54d41baa4e58f6af9d#file-plotly-cartesian-js 
  + together with https://gitcdn.xyz/

Note also, that at present we only register `calendars` in the "main" plotly.js bundles:

https://github.com/plotly/plotly.js/blob/97ba9c3903844ef9fef10010b62089b481fdc764/lib/index.js#L92-L95

cc @archmoj

### DIFF
--- a/src/traces/histogram/cross_trace_defaults.js
+++ b/src/traces/histogram/cross_trace_defaults.js
@@ -55,7 +55,7 @@ module.exports = function crossTraceDefaults(fullData, fullLayout) {
         if(!groupName) groupName = fallbackGroupName;
 
         var axType = getAxisType(traceOut, binDir);
-        var calendar = traceOut[binDir + 'calendar'];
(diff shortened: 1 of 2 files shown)


archmoj:
Nicely done.
:dancer:

Issue #4438, opened 2019-12-23
Recent toImage problems


etpinard:
Currently on `master`, `./tasks/noci_test.sh image` outputs

```
exported mapbox_0, in 2223.43075 ms
exported mapbox_angles, in 1665.376115 ms
exported mapbox_bubbles, in 1178.323781 ms
exported mapbox_bubbles-text, in 1258.123461 ms
exported mapbox_carto-style, in 835.072415 ms
exported mapbox_choropleth0, in 995.291759 ms
export error 525 for mapbox_choropleth0-legend - plotly.js error
  {"type":"error"}
exported mapbox_choropleth-multiple, in 1718.333484 ms
exported mapbox_choropleth-raw-geojson, in 1662.899144 ms
exported mapbox_connectgaps, in 1533.387381 ms
exported mapbox_custom-style, in 1875.924 ms
exported mapbox_density0, in 1156.254864 ms
export error 525 for mapbox_density0-legend - plotly.js error
  {"type":"error"}
exported mapbox_density-mulitple, in 1433.619021 ms
export error 525 for mapbox_density-mulitple_legend - plotly.js error
  {"type":"error"}
exported mapbox_earthquake-density, in 1440.466529 ms
exported mapbox_fill, in 1072.05885 ms
exported mapbox_geojson-attributes, in 2012.565512 ms
exported mapbox_layers, in 2141.55931 ms
exported mapbox_osm-style, in 834.893128 ms
exported mapbox_stamen-style, in 894.038581 ms
exported mapbox_symbol-text, in 1134.548921 ms
exported mapbox_texttemplate, in 1025.165576 ms
exported mapbox_white-bg-style, in 696.264266 ms

```

where the mapbox mocks added in https://github.com/plotly/plotly.js/pull/4386 fail to export using orca. 

These problematic mocks also fail to export via `Plotly.toImage` and `Plotly.downloadImage` in the test-dashboard console.

----

I suspect this problem is similar to https://github.com/plotly/plotly.js/pull/2914 and https://github.com/plotly/plotly.js/pull/1694

etpinard:
While attempting to fix this bug, we should test orca image-exports for tall the mocks added https://github.com/plotly/plotly.js/pull/4386

PR #4386, opened 2019-11-25
Add title feature to legend and support legend option for various traces

archmoj:
resolves https://github.com/plotly/plotly.js/issues/276 and
resolves https://github.com/plotly/plotly.js/issues/2642 and
resolves #3468

List of new `layout` attributes:
- `legend.title`
- `legend.title.text` 
- `legend.title.font`
- `legend.title.side`

List of new traces with `legend`:
- `choropleth`
- `choroplethmapbox`
- `cone`
- `densitymapbox`
- `heatmap`
- `histogram2d`
- `isosurface`
- `mesh3d`
- `streamtube`
- `surface`
- `volume` 

@plotly/plotly_js 
cc: @nicolaskruchten 

[before vs after demo](https://codepen.io/MojtabaSamimi/pen/gOOJNKg?editors=0010)

PR #4441, opened 2019-12-24
Fix image exports for new #4386 mocks with SVG gradients

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4438 - a follow-up from https://github.com/plotly/plotly.js/pull/4386.

Please note that exporting those new mocks worked fine in our (old) `imagetest` container, the problem was only noticed during a `./tasks/noci_test.sh` run.

----

Commit https://github.com/plotly/plotly.js/commit/b4ac79e613152d308f647304870c5ffd8a5035ce is the bare-bone fix and test :lock: 

Commit https://github.com/plotly/plotly.js/commit/55d76b2688f0016f608e5924e589be98e2c3fa67 introduces a more robust way of handling gradient URL strings during image exports. Let me know what you think!

cc @archmoj

### DIFF
--- a/src/components/drawing/index.js
+++ b/src/components/drawing/index.js
@@ -300,9 +300,10 @@ drawing.gradient = function(sel, gd, gradientID, type, colorscale, prop) {
         }
     }
 
-    var fullID = 'g' + gd._fullLayout._uid + '-' + gradientID;
+    var fullLayout = gd._fullLayout;
+    var fullID = 'g' + fullLayout._uid + '-' + gradientID;
 
-    var gradient = gd._fullLayout._defs.select('.gradients')
+    var gradient = fullLayout._defs.select('.gradients')
         .selectAll('#' + fullID)
         .data([type + colorStops.join(';')], Lib.identity);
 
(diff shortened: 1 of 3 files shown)


test/jasmine/tests/snapshot_test.js
```diff
@@ -344,6 +344,24 @@ describe('Plotly.Snapshot', function() {
...
+            it('- legend3dandfriends case', function(done) {
+                var fig = Lib.extendDeep({}, require('@mocks/geo_choropleth-legend.json'));
+
+                Plotly.plot(gd, fig)
+                .then(function() { return Plotly.Snapshot.toSVG(gd); })
+                .then(function(svg) {
+                    var svgDOM = parser.parseFromString(svg, 'image/svg+xml');
+
+                    var fillItems = svgDOM.getElementsByClassName('legend3dandfriends');
+                    expect(fillItems.length).toBe(4, '# of legend items');
```
etpinard: Before this PR, `fillItems.length` was `0`

src/components/drawing/index.js
```diff
@@ -333,6 +334,13 @@ drawing.gradient = function(sel, gd, gradientID, type, colorscale, prop) {
 
     sel.style(prop, getFullUrl(fullID, gd))
         .style(prop + '-opacity', null);
+
+    var className2query = function(s) {
+        return '.' + s.attr('class').replace(/\s/g, '.');
+    };
+    var k = className2query(d3.select(sel.node().parentNode)) +
+        '>' + className2query(sel);
+    fullLayout._gradientUrlQueryParts[k] = 1;
```
etpinard: We could go further by trying to construct a full xpath for each nodes with a gradient url, but I think this here is good enough for now.

archmoj:
@etpinard 
Thanks very much for the fix.
:dancer: :dancer:

PR #4447, opened 2019-12-27
Replace mapbox access token with token from plotly-js-tests account

etpinard:
cc @plotly/plotly_js


archmoj:
Thank you!
:dancer:

Issue #4417, opened 2019-12-08
Rangeslider range expands beyond data range and can’t be reset


asdFletcher:
I am copying this issue from a post on the plotly forum (https://community.plot.ly/t/rangeslider-range-expands-beyond-data-range-and-cant-be-reset/32193). Apparently this is a bug, and I am hopeful that posting it here helps others. 

Basically: The rangeslider range can be expanded beyond the range of the actual data. Once this happens, it’s not possible to reset the range slider range.

Any help on this issue would be appreciated! Is it a known issue? Is my config wrong?

Here is as [gif](https://aws1.discourse-cdn.com/business7/uploads/plot/original/2X/2/28399259ce54005c63f46e4c2920a887593aa3c2.gif) (which is a copy of the one on the forum post):

Here is a reproducible code sandbox: https://codesandbox.io/s/cool-hill-zdlh7
Uses: plotly-js, and react

I’ve tried a variety of things to fix this, but have not found a solution.

Is there a way to set the range programatically?

for instance: if range > limit, range = limit
Is there a config/layout option to limit the range?

I have found working examples that DON’T exhibit this behavior here:
https://plot.ly/python/range-slider/

---

Possibly I can look into fixing this myself but I am not familiar with the codebase. In addition, I see lots of room for improvement in the rangeslider behavior, adopting ideas from the C3 sub-graph, and the like.

Some possible changes to the  behavior:
We could make it so the range slider can't move beyond the graph's actual data range. Although, would this cause issues for a graph that allows zooming out far away from the data?

We could simply allow the range slider to collapse back a preferred size which matches the entire data set. This must happen on graph initialization anyway. It could be tied to double-click.

---

Also, a little off topic, but it may not even make sense to have the range slider directly attached to the original graph... what if it was simply a separate graph component, fed the same (or additional) data as the main graph, except with a selectable area. This might also address some of the customizability limitations of the rangeslider. In fact, the zoom-rectangle and left-right zoom handle functionality is already there, the graph just hast to NOT zoom in after the selection is made and report it back to a callback, which can be fed to another graph. Similar to the selectable area of the [react-vis 2d-draggable-area](https://uber.github.io/react-vis/examples/showcases/misc)

jay-bis:
Came across your post today, and just finished working on a solution I believe works. I tested it out on a few examples and it seems to get rid of the problem.

[...]

PR #4448, opened 2019-12-28
ensure newRng stays in bounds

jay-bis:
Created a few checks to make sure the slider's range doesn't go out of its normal bounds. This is an attempt to fix the bug outlined by this [gif](https://aws1.discourse-cdn.com/business7/uploads/plot/original/2X/2/28399259ce54005c63f46e4c2920a887593aa3c2.gif).

I'm not sure what the preferred behavior would be when you do drag the slider to the max/min bounds, but what this solution does is snap the slider back to the graph's corresponding max/min x value when the slider is dragged past the max/min.

### DIFF
--- a/src/components/rangeslider/draw.js
+++ b/src/components/rangeslider/draw.js
@@ -99,6 +99,15 @@ module.exports = function(gd) {
                 ];
             }
 
+            // make sure the slider's axis range (axRng) doesn't go above the
+            // axis max or below the axis min (rng)
+            if(axRng[1] > rng[1]) {
+                newRng[1] = rng[1];
+            }
+            if(axRng[0] < rng[0]) {
+                newRng[0] = rng[0];
+            }
+
             opts.range = opts._input.range = Lib.simpleMap(newRng, axisOpts.l2r);
         }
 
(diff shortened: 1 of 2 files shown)


etpinard:
Hi @jay-bis - are you attempting to fix the same bug as presented in https://github.com/plotly/plotly.js/issues/4417#issuecomment-569348207 here?

jay-bis:
Yes, I am! My first commit caused some of the tests to fail and was still a bit buggy; my newest commit passed all the tests and seems to work a bit more smoothly, and gets rid of the bug as far as I've tested it.

etpinard:
Ok great - thanks!

Your second commit looks good. Would you be interested in trying to write a few test cases to lock down your fix?

If so, the new tests should be similar to 

https://github.com/plotly/plotly.js/blob/e778644e1182e13126db6b801af6dd0a386372ba/test/jasmine/tests/range_slider_test.js#L152-L202

jay-bis:
For sure! I'll get started on that.

klausbreyer:
@etpinard I just stumbled upon #4417  and wanted kindly ask if there is any chance, that this will be merged in soon?

YMNNs:
@archmoj It really works, please consider merging #4448.

archmoj:
We somehow missed the updates on this PR. 
Thanks very much for the PR.
:dancer:
I am going to release a patch right after.

markchagers:
As of today, this is still an issue even in v2.27
Any chance of this getting fixed for real?

PR #4453, opened 2019-12-30
enables loading locales before Plotly

### LINKED ISSUE #4146: Order-independent script tags (MathJax & locales)
Spurred by https://github.com/plotly/dash/issues/880 as it's annoying to have to control script tag ordering in dash.

Currently the docs say MathJax must be loaded *before* plotly.js, and locales must be loaded *after*. This is because plotly.js looks for `typeof MathJax !== 'undefined'` during script execution, and if it finds MathJax it configures it. Whereas the opposite happens with locales: the script immediately calls `Plotly.register`, so it assumes that the `Plotly` object exists.

For MathJax we could simply defer this configuration to the first `Plotly.plot` call - and set some flag so we don't do it again.

For locales it's a little more complicated, we would have to do something like call `Plotly.register` if it exists, otherwise stash the locale in some global like `window.PlotlyLocales` (in a way that's compatible with multiple locale scripts), and then during the plotly.js script execution grab these and register them. At first I was thinking of deferring this to the first `Plotly.plot` call along with the MathJax config, and that may still work, I just worry people may want to inspect the available locales and would be confused that they don't see the right ones.

antoinerg:
This PR enables loading locales before plotly.js is on the page and partially fixes #4146.

In the following Codepens, we expect months to be in French (it should be `Mai` instead of `May`):
[Before](https://codepen.io/antoinerg/pen/vYEeNrg)
[After](https://codepen.io/antoinerg/pen/zYxEqeJ)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -65,6 +65,11 @@ register([
     require('./locale-en-us')
 ]);
 
+// locales that are present in the window should be loaded
+if(window.PlotlyLocales && window.PlotlyLocales.length) {
+    register(window.PlotlyLocales);
+}
+
 // plot icons
 exports.Icons = require('./fonts/ploticon');
 
--- a/src/core.js
+++ b/src/core.js
@@ -66,9 +66,9 @@ register([
 ]);
 
 // locales that are present in the window should be loaded
-if(window.PlotlyConfig && window.PlotlyConfig.locales) {
-    register(window.PlotlyConfig.locales);
-    delete window.PlotlyConfig.locales;
+if(window.PlotlyLocales && Array.isArray(window.PlotlyLocales)) {
+    register(window.PlotlyLocales);
+    delete window.PlotlyLocales;
 }
 
 // plot icons
--- a/src/core.js
+++ b/src/core.js
@@ -66,8 +66,9 @@ register([
 ]);
 
 // locales that are present in the window should be loaded
-if(window.PlotlyLocales && window.PlotlyLocales.length) {
+if(window.PlotlyLocales && Array.isArray(window.PlotlyLocales)) {
     register(window.PlotlyLocales);
+    delete window.PlotlyLocales;
 }
 
 // plot icons
--- a/src/core.js
+++ b/src/core.js
@@ -66,9 +66,9 @@ register([
 ]);
 
 // locales that are present in the window should be loaded
-if(window.PlotlyLocales && Array.isArray(window.PlotlyLocales)) {
-    register(window.PlotlyLocales);
-    delete window.PlotlyLocales;
+if(window.PlotlyConfig && window.PlotlyConfig.locales) {
+    register(window.PlotlyConfig.locales);
+    delete window.PlotlyConfig.locales;
 }
 
 // plot icons
--- a/tasks/util/wrap_locale.js
+++ b/tasks/util/wrap_locale.js
@@ -6,8 +6,8 @@ var intoStream = require('into-stream');
 
 var constants = require('./constants');
 
-var prefix = 'Plotly.register(';
-var suffix = ');';
+var prefix = 'var locale=';
+var suffix = ';if(typeof Plotly === \'undefined\') {window.PlotlyLocales = window.PlotlyLocales || []; window.PlotlyLocales.push(locale);} else {Plotly.register(locale);}';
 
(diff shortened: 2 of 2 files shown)


src/core.js
```diff
@@ -65,6 +65,11 @@ register([
     require('./locale-en-us')
 ]);
 
+// locales that are present in the window should be loaded
+if(window.PlotlyLocales && window.PlotlyLocales.length) {
```
etpinard:
```js
if(window.PlotlyLocales && Array.isArray(window.PlotlyLocales)) {
// ...
}
```

would be safer. Note that `Plotly.register([])` is handled correctly (essentially a noop) at the moment.

etpinard:
This is an interesting solution. I'm not a fan of us polluting the window namespace once more with this new `PlotlyLocales` global, but I can't really think of a better solution, so :ok_hand:

src/core.js
```diff
@@ -65,6 +65,11 @@ register([
...
+    register(window.PlotlyLocales);
```
etpinard: Maybe we could after this line have:

```js
delete window.PlotlyLocales;
```

to de-pollute the window namespaces after we're done registering the locales.

antoinerg: Done in f169380281377e3af40bb0772112d8ad9084f092

src/core.js
antoinerg: Done in f169380281377e3af40bb0772112d8ad9084f092

antoinerg:
> This is an interesting solution. I'm not a fan of us polluting the window namespace once more with this new `PlotlyLocales` global, but I can't really think of a better solution, so

@etpinard Would you prefer me using a `locales` key in the `window.PlotlyConfig` object. I know it is used to configure MathJax [here](https://github.com/plotly/plotly.js/blob/0c4cebbfaf1e0fba44855adc99dd890715d22982/tasks/stats.js#L68-L78).

src/core.js
archmoj: Since it is a temporary variable, it may be better to add an underscore i.e. `window._PlotlyLocales`.

antoinerg: Thanks for the suggestion! I'll let @etpinard choose between the different options (see also https://github.com/plotly/plotly.js/pull/4453#issuecomment-569838778)

archmoj: Wondering using [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) may be a possibility?

etpinard:
> Would you prefer me using a `locales` key in the `window.PlotlyConfig` object. I know it is used to configure MathJax [here](https://github.com/plotly/plotly.js/blob/0c4cebbfaf1e0fba44855adc99dd890715d22982/tasks/stats.js#L68-L78).

I forgot about `PlotlyConfig`. Thanks for pointing this out! Yes, I'd vote for putting your new locales stash in `PlotlyConfig`.

alexcjohnson:
Only issue with using `PlotlyConfig` is if a user includes `window.PlotlyConfig = {MathJaxConfig: 'local'}` - which is the documented way to support parallel MathJax usage - we'll still have some order-dependent behavior because that will wipe out the locales if they're loaded before the `PlotlyConfig` script and that script is loaded before plotly.js. I think it'd be better to keep `PlotlyConfig` reserved for explicitly user-provided behavior. I think `window.PlotlyLocales` is fine. That said, if we want to be really clever (and hard to debug...) we might even be able to use `window.Plotly`, as this won't exist yet during the execution of the plotly.js script, right?

antoinerg:
That's a very good point @alexcjohnson. I reverted https://github.com/plotly/plotly.js/pull/4453/commits/b4b20a7e562a8c0d03aad10a12fe535143e28dfa to use `window.PlotlyLocales`. Should we underscore it or is it OK as it is?

alexcjohnson:
I'm OK with `window.PlotlyLocales`, unless anyone feels strongly about the underscore.

etpinard:
> I'm OK with `window.PlotlyLocales`, unless anyone feels strongly about the underscore.

Yeah, that's fine :ok_hand: 

---

@antoinerg can you merge the latest master into your branch to get the syntax (header) tests to pass?

etpinard:
Nicely done @antoinerg - :dancer: !

Issue #4482, opened 2020-01-10
geo.visible vs template...showland


nicolaskruchten:
Right now, `layout.geo.visible=false` doesn't override `layout.template.layout.geo.showland=true` but it should.

archmoj:
Bug [demo](https://codepen.io/MojtabaSamimi/pen/ZEYobOQ?editors=0010)

alexcjohnson:
Hmm this is a bit of a tricky case but I think you're right to consider it a bug.  We have two inheritance pathways: the more-specific value from the template and the more-general value set explicitly, and normally the template wins but here we want the explicit-but-general value to win. I'm wondering if there are other cases like this - for example, should `layout.xaxis.color` beat out `layout.template.layout.xaxis.linecolor` to determine final `fullLayout.xaxis.linecolor`?

An additional wrinkle though: what if you set BOTH in the template and NEITHER in the main layout? ie `layout.template.layout.geo.showland=true` and `layout.template.layout.geo.visible=false` and you don't set `layout.geo.visible` or `layout.geo.showland`? We need to ensure that in this case `showland` will end up `true`, so we can't even trust the coerced value of `fullLayout.geo.visible`

nicolaskruchten:
Yes, agreed that it's tricky.

Agreed that `layout.template.layout.geo.visible=false` and `layout.template.layout.geo.showland=true` and `layout.geo.visible=null` and `layout.geo.showland=null` should result in land showing only IMO, so you can't use the "normal" machinery.

archmoj:
Should we override all these? Or only 'showland'?
```
'lonaxis.showgrid',
'lataxis.showgrid',
'showcoastlines',
'showocean',
'showland',
'showlakes',
'showrivers',
'showcountries',
'showsubunits',
'showframe'
```

nicolaskruchten:
all of the above except for `*axis.showgrid` I would think.

archmoj:
Why should we skip `*axis.showgrid`?

nicolaskruchten:
Ah I see that `geo.visible` currently impacts `showgrid` per https://github.com/plotly/plotly.js/commit/0909f339c0e481b2ba882491b3e544d5cc5eb47d  so... OK for `showgrid` too.

etpinard:
> I'm wondering if there are other cases like this - for example, should `layout.xaxis.color` beat out `layout.template.layout.xaxis.linecolor` to determine final `fullLayout.xaxis.linecolor`?

[...]

PR #4483, opened 2020-01-10
geo.visible false should override template.layout.geo.show*

archmoj:
Fixes: #4482 | [before](https://codepen.io/MojtabaSamimi/pen/ZEYobOQ) vs [after](https://codepen.io/MojtabaSamimi/pen/KKwRgvv)

@plotly/plotly_js 


src/plots/plots.js
```diff
@@ -1446,6 +1446,29 @@ plots.supplyLayoutGlobalDefaults = function(layoutIn, layoutOut, formatObj) {
 
     var template = layoutIn.template;
     if(Lib.isPlainObject(template)) {
+        if( // geo.visible false should override template.layout.geo.show* - see issue 4482
+            layoutIn.geo &&
+            layoutIn.geo.visible === false &&
+            template.layout &&
+            template.layout.geo
```
alexcjohnson: This is a clever solution! But if we do it this way it will need some extending. `template.layout.geo` will apply to `layout.geo` but also `layout.geo2`, `.geo3`, etc... unless there's also a `template.layout.geo2`, in which case THAT will only apply to `layout.geo2`, nothing before or after.

Oh actually... that means this strategy won't work. If you have `template.layout.geo.showland=true`, no `template.layout.geo2`, but `layout.geo.visible=true` and `layout.geo2.visible=false`, what do you do with the template? It would need to appear `show*=false` to `geo2` but the original setting to `geo`.

So I think the solution is going to have to be at the `supplyDefaults` level, not at the template level... where it seems like we'll need a special coerce variant involving `geoLayoutIn.visible`, `geoLayoutOut._template.visible`, `geoLayoutIn[x]`, and `geoLayoutOut._template[x]`

Unless you want to, in this case, copy `newTemplate.layout.geo` into `newTemplate.layout.geo2` and modify only one of them... but it seems to me given all this extra complexity a `supplyDefaults` solution will be easier.

test/jasmine/tests/geo_test.js
```diff
@@ -688,6 +688,51 @@ describe('Test Geo layout defaults', function() {
...
+        keys.forEach(function(k) {
+            layout.template.layout.geo[k] = true;
+        });
+
+        var data = [{ lon: [0], lat: [0], type: 'scattergeo' }];
+
+        Plotly.plot(gd, data, layout)
+        .then(function() {
+            keys.forEach(function(k) {
+                expect(gd._fullLayout.template.layout.geo[k]).toBe(false, k);
```
alexcjohnson: At the end of the day we don't care what's in `_fullLayout.template` - only the values of `_fullLayout.geo*[k]`. Would you mind modifying the test to look at these attributes, under both of the cases we discussed: `visible: false` in the main layout as well as `visible: false` in the template?

src/plots/plots.js
archmoj: Good call. Fixed by https://github.com/plotly/plotly.js/pull/4483/commits/7e62ce45f14c1a1abdb05c49a631ca4f63439019.

test/jasmine/tests/geo_test.js
archmoj: Good call. Fixed by https://github.com/plotly/plotly.js/pull/4483/commits/7e62ce45f14c1a1abdb05c49a631ca4f63439019.

nicolaskruchten:
You’re doing the axis grid show attrs too right?

archmoj:
> You’re doing the axis grid show attrs too right?

Oh no! I misread your reply here: https://github.com/plotly/plotly.js/issues/4482#issuecomment-573139716

Thanks for pointing out!
So I need to add one more commit to this PR.

archmoj:
> You’re doing the axis grid show attrs too right?

Now done in https://github.com/plotly/plotly.js/pull/4483/commits/1b905a1588752b6fd46be1ea1544ffd881fb88f3.

src/plots/geo/layout_defaults.js
```diff
@@ -46,6 +46,28 @@ function handleGeoDefaults(geoLayoutIn, geoLayoutOut, coerce, opts) {
     var isClipped = geoLayoutOut._isClipped = !!constants.lonaxisSpan[projType];
 
     var visible = coerce('visible');
+    if(visible === false) {
```
alexcjohnson: I think this is still incorrect when the `visible=false` came from the template (which we still need a test for - showing that all the `show*` end up `true` in that case). I guess the condition here should be `if(geoLayoutIn.visible === false)`

archmoj: Addressed in https://github.com/plotly/plotly.js/pull/4483/commits/639913e2f7e77e93a4cd0db40a1a19f67d2b82e4.

src/plots/geo/layout_defaults.js
```diff
@@ -46,6 +46,29 @@ function handleGeoDefaults(geoLayoutIn, geoLayoutOut, coerce, opts) {
     var isClipped = geoLayoutOut._isClipped = !!constants.lonaxisSpan[projType];
 
     var visible = coerce('visible');
+    if(geoLayoutIn.visible === false) {
+        visible = geoLayoutOut.visible = true;
```
alexcjohnson: Why are we setting `geoLayoutOut.visible = true` here? Looks to me as though it doesn't matter for the end result, is that right? RCE would want the `false` to remain intact.

test/jasmine/tests/geo_test.js
```diff
@@ -650,39 +650,93 @@ describe('Test Geo layout defaults', function() {
...
-        it('- base case', function() {
-            layoutIn = {
-                geo: { visible: false }
-            };
+        [true, false, undefined].forEach(function(q) {
+            it('- base case | ' + q, function() {
+                layoutIn = {
+                    template: {
+                        layout: {
+                            geo: {
```
alexcjohnson: sorry to be a pain, but I still don't see a test case that sets `template.layout.geo.visible=false`, does NOT set `layout.geo.visible`, and verifies that `template.layout.geo.show* settings are still honored in `fullLayout.geo`

src/plots/geo/layout_defaults.js
archmoj: My bad.

nicolaskruchten:
Reminder: we need this merged and released by end of day today if at all possible :)

test/jasmine/tests/geo_test.js
archmoj: Added in https://github.com/plotly/plotly.js/pull/4483/commits/9448c8f6b6d968d76e6a49b4bb92a78ecb783b99.

PR #4506, opened 2020-01-23
Match Cartesian axes with identical invisible reference

### LINKED ISSUE #4501: Axis matching limitation
This is a pretty problematic case for Plotly.py: https://codepen.io/nicolaskruchten/pen/ExaOMRr?editable=true

In this case, all axes have `matches` set to the first axis (`x` or `y`) but those axes have no traces, so no information is being shared among the axes that *do* have traces.

I'd be tempted to call this a bug... thoughts @alexcjohnson @archmoj ?

This is a problem for Plotly.py because `make_subplots` creates subplots and sets `matches` *before* knowing which subplots have traces.

archmoj:
Fixes #4501.

Demos:
- [Before](https://codepen.io/MojtabaSamimi/pen/mdyvRJo?editors=0010)
-  [After](https://codepen.io/MojtabaSamimi/pen/gObqgaJ?editors=1000)

@alexcjohnson 
cc: @nicolaskruchten @plotly/plotly_js

### DIFF
--- a/src/plots/cartesian/constraints.js
+++ b/src/plots/cartesian/constraints.js
@@ -19,6 +19,7 @@ var FROM_BL = require('../../constants/alignment').FROM_BL;
 
 exports.handleConstraintDefaults = function(containerIn, containerOut, coerce, opts) {
     var allAxisIds = opts.allAxisIds;
+    var layoutIn = opts.layoutIn;
     var layoutOut = opts.layoutOut;
     var scaleanchorDflt = opts.scaleanchorDflt;
     var constrainDflt = opts.constrainDflt;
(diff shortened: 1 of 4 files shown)


nicolaskruchten:
Does this PR check to make sure that the new `matches` target is a valid one with traces? Or if not, do all references get cleaned up? For example if x2, x3, x4, x5 are all matching x and x is invalid and we make them all point to x2 but then x2 is invalid also do we then match them all to x3?

archmoj:
> Does this PR check to make sure that the new `matches` target is a valid one with traces? Or if not, do all references get cleaned up? For example if x2, x3, x4, x5 are all matching x and x is invalid and we make them all point to x2 but then x2 is invalid also do we then match them all to x3?

It does only search within valid axes. So this should not be a problem.
BTW it would be great to test the demo posted on the PR description with more combinations.

alexcjohnson:
One thing this approach misses is if the lost axis sets an explicit `range` or `rangemode` https://codepen.io/alexcjohnson/pen/wvBNXgY?editors=0010
That seems like something we want to support, right?

nicolaskruchten:
Ah drat. Also `categoryarray` and any other axis-related option I would guess. That is a drawback to this approach: you need to copy everything over basically

archmoj:
Then, should x.range override x2.range if both specified?

alexcjohnson:
> should x.range override x2.range if both specified?

The answer to that is "exactly whatever would have happened if there was data on the xy subplot" - I don't know how we handle that though I'd guess that `xaxis.range` wins.

archmoj:
Closing in respect to #4529.

PR #4519, opened 2020-01-24
Improve pie and sunburst inside text pad 

### LINKED ISSUE #4518: no pad applied in pie & sunburst inside text
Unlike bar-like traces in `pie` and `sunburst` no textpad is applied for fitting text inside slices.
This causes problems even with small line width values. 
[pie demo](https://codepen.io/MojtabaSamimi/pen/WNbmoOL?editors=0010)
Noting that in `sunburst` which has default white `marker.line.color` and white text on dark slices (e.g. blue) the text may be mixed with the outside area.  
[sunburst demo](https://codepen.io/MojtabaSamimi/pen/Baybpoe?editors=0010)

archmoj:
Fixes #4518 by considering `TEXTPAD` similar to bar-like traces 
when computing scales for `radial` and `tangential` orientation options.

Thanks very much in advance for reviewing revised baselines!

@plotly/plotly_js 
cc: @nicolaskruchten


etpinard:
The new baselines look good! Thanks for doing this @archmoj 

I'm not 100% comfortable releasing this into a patch, as the new pad value can be the difference between an inside text and an outside text. This would be a great addition to `v1.53.0`. 

One small comment: it would be nice to combine all the new mocks in https://github.com/plotly/plotly.js/pull/4519/commits/6b8de97d50801885142242b773d2e6a57c9d5374 into a single figure.

etpinard:
:dancer:

PR #4521, opened 2020-01-27
Accept u, v and w in isosurface and volume APIs 

### LINKED ISSUE #4517: Accept u, v and w in isosurface and volume APIs similar to cone and streamtube APIs
When working with CFD (computational fluid dynamics) simulation the data could be in the form of 3-D vector (u, v, w). To facilitate interfacing with raw data (and adding iso-surface layers on top of stream tube or cone plots), the `value` array including wind speed can be computed internally.

archmoj:
Resolves #4517 
By allowing `u`, `v` & `w` to be passed instead of `value` in `isosurface` and `volume` traces, one could simply (combine &) switch between traces such as `cone` and `streamtube` with some CFD data in hand.

@plotly/plotly_js 
cc: @jackparmer


archmoj:
cc: @peterdsharpe, @empet, @emmanuelle

empet:
I'm not familiar with  volumes that can have the attributes u, v, w.
For  a Cone trace (u,v, w) represents the direction of a vector field at the corresponding point (x, y, z).
For volumes I can think at u, v, w like parameters, similar to parameters in parameterized surfaces,
i.e. u, v, w are derived from a 3d meshgrid, and then
x, y, z are functions of (u,v,w):
 x(u,v,w), y(u,v,w), z(u,v,w).

The values at x, y, z are theoretically defined by a function f:

`vals = f(x(u,v,w), y(u,v,w), z(u,v,w))`

When x(u,v,w)=u, y(u,v,w)=v, z(u,v,w) we can recover the actual definition of a volume.
If this is a case it's a nice extension. Otherwise, please give more details to be able 
to express an opinion.

archmoj:
> When x(u,v,w)=u, y(u,v,w)=v, z(u,v,w) we can recover the actual definition of a volume.
> If this is a case it's a nice extension. Otherwise, please give more details to be able
> to express an opinion.

@empet thanks for your feedback.
Yes. That's the case. The `value` is computed from `sqrt(u^2+v^2+z^2)` here:
https://github.com/plotly/plotly.js/blob/afc50bfac346f6c77802f4510c61ebc77bac9503/src/traces/isosurface/calc.js#L70

empet:
@archmoj   I cannot understand what relevance has u*u+v*v+w*w for a brain read from a 3d image,.  as well as  when I want to plot an isosurface, let us say f(x, y, z) = f(u,v,w)= u*u*u- 2*u*v*w= 2. in a given volume  u in [-1,1], v in [0, 3], w in [-1,2].

archmoj:
> @archmoj I cannot understand what relevance has u_u+v_v+w_w for a brain read from a 3d image,. as well as when I want to plot an isosurface, let us say f(x, y, z) = f(u,v,w)= u_u_u- 2_u_v_w= 2. in a given volume with u in [-1,1], v in [0, 3], w in [-1,2].

This is more relevant in the context of wind/wave simulation.
However I can imagine a case where one may be interested to display the difference between two MRI images of the same person. In that case every point could be moved by dx, dy, dz in x, y & z directions where we can get u, v, w components.

archmoj:
@jackparmer wondering this PR could be useful in CFD apps like this: https://dash-gallery.plotly.host/dash-aerosandbox/

PR #4522, opened 2020-01-28
Fix - hide empty bars in interactive mode

### LINKED ISSUE #4499: Null values show up as thin bars in Bar chart if you set an axis range including 0.
If you have null values in y, those show up as thin bars in Bar chart if you set an axis range that includes 0.

image

Here is a pen. 

https://codepen.io/kei51/pen/KKwraMX

It doesn't happen on an older version of Plotly like 1.47.4.

https://codepen.io/kei51/pen/XWJypEO

archmoj:
Fixes #4499 

[before](https://codepen.io/MojtabaSamimi/pen/rNabvjJ?editors=0010) vs [after](https://codepen.io/MojtabaSamimi/pen/MWYRdMN?editors=1000)

@plotly/plotly_js


src/traces/bar/plot.js
```diff
@@ -190,7 +191,9 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts, makeOnCompleteCallback)
                     d3.round(Math.round(v) - offset, 2) : v;
             }
 
-            function expandToVisible(v, vc) {
+            function expandToVisible(v, vc, hideZeroHeight) {
```
etpinard: Can you change the 3rd argument to `hideZeroSpan`?

etpinard:
Thanks @archmoj for the quick fix. After some testing, looks like the behaviour is now correct. I am a little worried though about how complex the logic in `bar/plot.js` has become.

At the very least, we should add two need variables to make the logic easier to digest:

- one for `isHorizontal && (x1 - x0 !== 0)` e.g. setting a variable named `spanHorizontal`
- one for `!isHorizontal && (y1 - y0 !== 0)` e.g. " " " " `spansVertical`

which could be in 

```js
            // display zeros if line.width > 0
            if(isBlank && shouldDisplayZeros && helpers.getLineWidth(trace, di) && (isHorizontal ? x1 - x0 === 0 : y1 - y0 === 0)) {
                isBlank = false;
            }
```

and below in

```js
if(!gd._context.staticPlot) {
  // ...

  if(spansHorizontal) {
                x0 = fixpx(x0, x1);
                x1 = fixpx(x1, x0);
  }
  if(spansVertical) {
                y0 = fixpx(y0, y1);
                y1 = fixpx(y1, y0);
  }
}
```

instead of monkey-patching the `expandVisible` helper for the degenerate case.

----

Does that sound ok @archmoj ?

archmoj:
@etpinard that makes sense.
Alternatively we may consider adding "only" this line:
```js
if(v === vc) return v;
``` 
to `expandToVisible` function without adding extra logic to bar.

etpinard:
> Alternatively we may consider adding "only" this line:
> 
> ```js
> if(v === vc) return v;
> ```

Hmm. Wouldn't that defeat the purpose of `expandVisible`. Right now, if `v === vc`, `v` gets `floor`ed or `ceil`ed.

etpinard:
... and unfortunately we can't rely on image test here as this piece of logic in inside a `!gd._context.staticPlot` block.

archmoj:
OK. Addressing your https://github.com/plotly/plotly.js/pull/4522#issuecomment-579843027 sounds the safest fix.

etpinard:
Yep, that's a step in the right direction - thanks :dancer:

PR #4523, opened 2020-01-28
Update insidetextorientation description for pie and sunburst traces

jdamiba:
The purpose of this PR is to update the text description for the pie and sunburst trace types. 

@plotly/plotly_js

### DIFF
--- a/src/traces/pie/attributes.js
+++ b/src/traces/pie/attributes.js
@@ -187,13 +187,16 @@ module.exports = {
         values: ['horizontal', 'radial', 'tangential', 'auto'],
         dflt: 'auto',
         editType: 'plot',
+            
         description: [
-            'Determines the orientation of text inside slices.',
-            'With *auto* the texts may automatically be',
-            'rotated to fit with the maximum size inside the slice.',
-            'Using *horizontal* option forces text to be horizontal.',
-            'Using *radial* option forces text to be radial.',
-            'Using *tangential* option forces text to be tangential.'
+            'The `insidetextorientation` attribute controls the',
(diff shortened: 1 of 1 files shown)


archmoj:
@jdamiba Nicely done. The very first PR to plotly.js!
 
:dancer:

src/traces/pie/attributes.js
```diff
@@ -188,12 +188,14 @@ module.exports = {
...
         editType: 'plot',
         description: [
-            'Determines the orientation of text inside slices.',
-            'With *auto* the texts may automatically be',
-            'rotated to fit with the maximum size inside the slice.',
-            'Using *horizontal* option forces text to be horizontal.',
-            'Using *radial* option forces text to be radial.',
-            'Using *tangential* option forces text to be tangential.'
+            'The `insidetextorientation` attribute controls the',
+            'orientation of the text inside chart sectors.',
```
alexcjohnson: The updated text looks great! Minor comment though - we generally don't start with "The `xxx` attribute" - I would 🔪 it here (start "Controls the orientation") unless we want to have a discussion about moving to that grammar everywhere. If we want the public-facing docs to have that grammar we could  automatically prepend it there (along with lowercasing the first character) if we first go through all the descriptions and check that they're all written in a way that would admit such a lead-in. A quick look shows that *most* descriptions currently start with a verb (usually "Sets" or "Determines") but a few do not.

jdamiba: I edited the text to be  being with a verb.

archmoj:
:dancer: :dancer:

PR #4529, opened 2020-01-29
Handle 'missing' matching axes

etpinard:
fixes https://github.com/plotly/plotly.js/issues/4501 - attempts to supersede @archmoj 's https://github.com/plotly/plotly.js/pull/4506 by implementing "solution 1" as presented in https://github.com/plotly/plotly.js/issues/4501#issuecomment-579371483

---

Examples:

- minimal case where we have traces on subplots x2y2, x3y3, x4y4, all x-axes have `matches: 'x'`, all y-axes have `matches: 'y'`
  + off `v1.52.1`: https://codepen.io/MojtabaSamimi/pen/wvBRGyV - `matches: 'x'` and `matches: 'y'` are invalid, and the axes are not linked
  + off this branch: https://codepen.io/etpinard/pen/gObJBXm?editors=1010 `matches: 'x'` and `matches: 'y'` are now valid
- same as previous, but with a set `xaxis.range`
  + off @archmoj 's #4506 branch: https://codepen.io/alexcjohnson/pen/wvBNXgY?editors=0010 - that `xaxis.range` setting gets ignored
  + off this branch: https://codepen.io/etpinard/pen/KKwLGZe?editors=1010 - the `xaxis.range` gets coerced and is the value used for the matching x-axes
- similar to previous, but now with traces on subplots x3y3 and x4y4 only and x4 matching x3, x3 matching x2 and x2 matching x
  + this case is also working off this branch: https://codepen.io/etpinard/pen/GRgaYxW

Pinging @alexcjohnson @nicolaskruchten who may want to try these examples.

---

For the reviewers, commit https://github.com/plotly/plotly.js/commit/33078b0c12e17b5cee7e7240f430da67bac1a44c adds a way to track (in `fullLayout._axisMatchGroups`) and coerce the 'missing' axes which are referenced via the `matches` attributes but that do not have traces on them (i.e. are not coerced into `fullLayout` otherwise). Commit df1c59f7ab502b078ed62106d669eaab52aafbdf makes the downstream logic work: as far as I could tell we only needed to skip missing axes during `Axes.drawOne` and during the `_axisMatchGroups` autorange sync operation.


nicolaskruchten:
Just to confirm: this does or could work with other attributes than `range` like `categoryarray` and `alignmentgroup` and so on?

alexcjohnson:
Looks great! My only code comment is it may be possible to 🌴 the missing & real axis coercion code a bit, but OTOH that could make it less readable... so definitely non-⛔️

> Just to confirm: this does or could work with other attributes than `range` like `categoryarray` and `alignmentgroup` and so on?

It certainly looks like that should work, it does a fairly complete coercion of attributes on these missing axes, making no explicit mention of `range` in the process. But yeah, may be nice to include tests of these other attributes.

etpinard:
> this does or could work with other attributes than `range` like `categoryarray` and `alignmentgroup` and so on?

Does anyone have a px example ready that uses `categoryarray` and `alignmentgroup`?

etpinard:
> My only code comment is it may be possible to  the missing & real axis coercion code a bit, but OTOH that could make it less readable... so definitely non-

Yeah, I paid special attention to making this new code as readable as it can be. The cartesian `supplyLayoutDefaults` logic is already pretty complex.

src/plots/cartesian/layout_defaults.js
```diff
@@ -239,12 +263,60 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
         coerce('title.standoff');
 
+        addMissingMatchedAxis(axLayoutIn.matches);
+
+        axLayoutOut._input = axLayoutIn;
+    }
+
+    // coerce the 'missing' axes
+    i = 0;
+    while(i < missingMatchedAxisIds.length) {
```
archmoj: What about using a `for` loop here instead of `while`?

etpinard: I prefer using `while` when the terminal value (here `missingMatchedAxisIds.length`) can changed inside the loop.

archmoj: Thanks for the note.

etpinard: No problem. I should have highlighted this. Calling `addMissingMatchedAxis` again inside this `while` is how I got the nested cases to work.

nicolaskruchten:
Here's a PX-output case with alignmentgroup set: https://codepen.io/nicolaskruchten/pen/ExaOMRr?editable=true

nicolaskruchten:
If you set `categoryarray` on the `x` axis there, we would like it to propagate.

etpinard:
Thanks for the example @nicolaskruchten 

Things appear to work ok with `categoryarray` set in the missing `x` axis: https://codepen.io/etpinard/pen/QWwXKPd

**but** one must set `xaxis.type: 'category'` first, because as the `x` axis doesn't have any traces associated with it, we get the default `xaxis.type: 'linear'`. Recall that matching axes of different `type` values is forbidden. 

I suspect that setting `type: 'category'` isn't ideal in the PX context, so let me try to find a way to _propagate_ the axis type through to the 'missing' axes.

etpinard:
> Things appear to work ok with `categoryarray` set in the missing `x` axis:

Actually, the x categories are reversed, we should show Male then Female. On it.

nicolaskruchten:
Hmm well today PX sets `categoryorder`/`categoryarray` a little oddly and I can't remember why: it sets it on every x axis in the bottom row and every y axis in the left column. That seems like a leftover from the pre-`matches` days. PX could just set it on all axes though if that makes things easier. Arguably setting it on the bottom-most row is enough, as there will always be at least one valid x axis in the bottom-most row for PX plots.

nicolaskruchten:
(and ditto one valid y-axis in the left-most column)

but that's a very PX-specific property.

etpinard:
Axis type propagation and `categoryarray` behaviour on missing axes is now fixed:

https://codepen.io/etpinard/pen/MWYMpGm

@archmoj this PR is ready for another round of review.

test/jasmine/tests/cartesian_interact_test.js
```diff
@@ -1535,6 +1535,59 @@ describe('axis zoom/pan and main plot zoom', function() {
...
+                [['yaxis', 'yaxis2', 'yaxis3', 'yaxis4'], [1.156, 2.335], {dragged: true}]
+            ]))
+            .then(doDblClick('x3y3', 'nsew'))
+            .then(_assert('after double-click on x3y3 subplot', [
+                [['xaxis', 'xaxis2', 'xaxis3', 'xaxis4'], [0.8206, 3.179], {autorange: true}],
+                [['yaxis', 'yaxis2', 'yaxis3', 'yaxis4'], [0.9103, 2.0896], {autorange: true}]
+            ]))
+            .catch(failTest)
+            .then(done);
+        });
```
archmoj: This test fails on my machine.
Screenshot from 2020-01-30 20-32-37

test/image/compare_pixels_test.js
```diff
@@ -104,6 +104,7 @@ var FLAKY_LIST = [
     'treemap_coffee',
     'treemap_textposition',
     'treemap_with-without_values_template',
+    'treemap_with-without_values',
```
archmoj: Despite black listing this mock, the image test still fails!

etpinard: Do you have any idea why?

archmoj: I am trying to fix it.

test/jasmine/tests/cartesian_interact_test.js
etpinard: Hmm. Interesting. It passes fine on my laptop and on CI. This test isn't new. Does it pass on master on your laptop?

test/image/compare_pixels_test.js
etpinard: Would it be ok if I simply update the baseline?

test/jasmine/tests/cartesian_interact_test.js
archmoj: On my machine, that fails with master too.
Resolving...

etpinard: Ok, well looks like adding a bit of tolerance should be enough to make it pass on your machine. Feel free to push a commit that bumps the tolerance value right on this branch.

test/image/compare_pixels_test.js
archmoj: You could pick https://github.com/plotly/plotly.js/commit/2589451c1c0734de9c0268b7af6093bd6f94d461.

etpinard: Wanna push that commit on this branch? It's your work afterall. Thanks very much for looking into this!

archmoj: Not a big deal. 
OK. I'll push.

archmoj:
@etpinard thanks very much for this great PR.

:dancer: :dancer: :dancer:  
:dancer: :dancer: :dancer:  
:dancer: :dancer:

PR #4536, opened 2020-01-31
add the word background to bgcolor description so it will get indexed by algolia

jdamiba:
The purpose of this PR is to update the description for the `paper_bgcolor` and `plot_bgcolor` layout attributes to add the word `background`. 

This is so that users who search for `background` in the search bar on https://plot.ly/javascript will see these attributes as results.


etpinard:
Excellent. Thanks! :dancer:

Issue #4415, opened 2019-12-06
Contour negative valued layer is not visible


poldz123:
Negative numbers within the layer are not that visible, it looks like the symbol it is part of the line.

Adding extra padding should fix this visibility issue only for negative symbols.

![contour-sample](https://user-images.githubusercontent.com/4010507/70355749-e65c3e80-1840-11ea-9d34-8c712d1cfe57.png)

etpinard:
Thanks for pointing this out.

Maybe instead of some vertical adjustment, we could implement https://github.com/plotly/plotly.js/issues/1597 to make the minus sign more visible.

etpinard:
> Adding extra padding should fix this visibility issue only for negative symbols.

Oh you mean some padding between the negative sign and the start of the contour line i.e. tweaking the label bounding box dimensions to "pad" the text of the contour labels. That's actually a great idea with or without https://github.com/plotly/plotly.js/issues/1597 and is probably much easier to implement. 

Thanks for posting @poldz123 !

alexcjohnson:
BTW whenever this gets addressed, looks like contour labels sometimes use the regular hyphen "-", not minus sign "−" like we do for all other negatives

https://github.com/plotly/plotly.js/blob/162afe491bccda9284e7c3b903a45bd9ac5cf836/src/constants/numerical.js#L62

Looking at https://rreusser.github.io/plotly-mock-viewer/#contour_edge_cases - perhaps when `contours.labelformat` is specified we don't fix this character, but we do when auto-formatting? Compare the "-1.0e-6" to "−1µ"
<img width="266" alt="Screen Shot 2019-12-09 at 12 03 11 PM" src="https://user-images.githubusercontent.com/2678795/70456373-fad84b00-1a7b-11ea-91f0-811693ecd656.png">

@poldz123 does your example use `contours.labelformat`? Or is there another case we're missing? We *do* use the correct minus sign for axis ticks with `axis.tickformat`, so seems this issue is just with contour labels.

poldz123:
@alexcjohnson I do use the `contours.labelformat` with the value: `.0f`

etpinard:
@poldz123 I'm attempting to fix this problem in https://github.com/plotly/plotly.js/pull/4540, maybe you could try out this dev bundle:

https://65959-45646037-gh.circle-artifacts.com/0/dist/plotly.min.js

in your use case to see if the results are to your likings. Thanks!

poldz123:
@etpinard 
[...]

PR #4540, opened 2020-01-31
Contour label extra pad and correct minus sign

etpinard:
my attempt to fix https://github.com/plotly/plotly.js/issues/4415

Best to review this PR by comparing the baseline changes in each commit. The changes induces in the first commit https://github.com/plotly/plotly.js/commit/a32f76f490e014768a93f76c74db08e558c82275 should be uncontroversial. I'm not 100% sure about the results in commit https://github.com/plotly/plotly.js/commit/30bee29643071ed77356829df721ea1b25a60be0 though: most baselines look at lot better to me, but some appear to have a little too much padding e.g. `cheater_contour`, `contour_label-formatting-via-colorbar`, `contour_lines_coloring`, `heatmap_multicategory`.

@plotly/plotly_js please let me know what you think.


src/traces/contour/constants.js
```diff
@@ -72,5 +72,8 @@ module.exports = {
         INITIALSEARCHPOINTS: 10,
         // number of binary search iterations after the initial wide search
         ITERATIONS: 5
-    }
+    },
+
+    // number of px around labels text clip out of contour lines below
+    LABEL_LINE_CLIP_PAD: 2
```
archmoj: Could you please try 3px horizontal padding here?

src/traces/contour/plot.js
```diff
@@ -543,8 +544,8 @@ function locationCost(loc, textOpts, labelData, bounds) {
 }
 
 exports.addLabelData = function(loc, textOpts, labelData, labelClipPathData) {
-    var halfWidth = textOpts.width / 2;
-    var halfHeight = textOpts.height / 2;
+    var halfWidth = textOpts.width / 2 + LABEL_LINE_CLIP_PAD;
+    var halfHeight = textOpts.height / 2 + LABEL_LINE_CLIP_PAD;
```
archmoj: After viewing the baseline changes, I think we may not need to add vertical padding to the height.

alexcjohnson: I agree, let's not add vertical padding. In fact if there was a way to make it even a tighter fit vertically than it currently is on master, I'd be all for it. Probably not though, assuming that comes from a bounding box... but the goal is to avoid clipping neighboring contour lines as much as possible, since that can cause confusion about which line is actually being labeled.

src/traces/contour/constants.js
alexcjohnson: 2 px looks like plenty to me, but I might argue for making it a fraction of the font size (or text height) so that it scales with the text.

etpinard:
Thanks for adding commits @archmoj and taking over this PR !!

Looks like your solution uses a constant padding and does not attempt to implement @alexcjohnson's

> but I might argue for making it a fraction of the font size (or text height) so that it scales with the text

idea, so I'll let @alexcjohnson confirm that the updated baselines look :ok_hand:

src/traces/contour/plot.js
```diff
@@ -544,24 +543,28 @@ function locationCost(loc, textOpts, labelData, bounds) {
 }
 
 exports.addLabelData = function(loc, textOpts, labelData, labelClipPathData) {
-    var halfWidth = textOpts.width / 2;
-    var halfHeight = textOpts.height / 2;
+    var w = textOpts.width + 4;
+    var h = Math.max(0, textOpts.height - 4);
```
alexcjohnson: This looks great for all the baselines we test. But I'm still concerned about it in case someone uses a very different font size, particularly shrinking height by a constant if someone were to use a very small font size, perhaps for publication when it's expected the user can zoom in on the PDF or something.

So yeah, I'd still like to make both the +4 and -4 adjustments relative to font size, or if you think there's some ill effect we'd expect from that, at least ensure the -4 isn't too large a fraction of the height.

archmoj:
@alexcjohnson this PR should be ready for another review.
[before](https://codepen.io/MojtabaSamimi/pen/XWbEBww?editors=0010) vs [after](https://codepen.io/MojtabaSamimi/pen/vYORaqy?editors=1000)
Also to note: it would be nice to have it in a minor release.

PR #4539, opened 2020-01-31
Fix hover of mesh3d traces with face color and face intensity

archmoj:
Fixes #4527 by displaying `facecolor` or face-intensity at the center of each face.

[demo1 before](https://codepen.io/MojtabaSamimi/pen/yLydGzJ?editors=0010) vs [demo1 after](https://codepen.io/MojtabaSamimi/pen/qBEzLPo?editors=0010)

[demo2 before](https://codepen.io/MojtabaSamimi/pen/mdyZaqv?editors=0010) vs [demo2 after](https://codepen.io/MojtabaSamimi/pen/PowrXOv?editors=0010)

@plotly/plotly_js 
cc: @empet

Issue #4445, opened 2019-12-27
Complete mesh3d API to accept cell intensity


archmoj:
The `gl-vis/mesh3d` module supports cell intensity. 
But at the moment one could only pass vertex intensity using `plotly.js` API.
cc: @jackparmer

PR #4541, opened 2020-02-02
Improve mesh3d tests using cell intensitymode 

archmoj:
A follow up of #4539 and #4445 this PR adds tests for `mesh3d` with `cell` `intensitymode` and hover on different grid types e.g. "delaunay`, "alpha-hull" and "convex-hull".


etpinard:
Thank you very much :dancer: !

Issue #4502, opened 2020-01-21
iPad Pro 12.9 (13) and iPad 7th (13) Not Detected Properly using Safari


jneilliii:
There appears to be an issue with WebGL detection on larger iPad Pro 12.9" tablets.  I found ticket #280 that addressed this at one point for other iOS devices, but not sure if this is a regression or not.  Additional details can be found [here](https://github.com/jneilliii/OctoPrint-BedLevelVisualizer/issues/180) of the problem.  It seems that if the browser is made smaller the graphing works fine, but once maximized to the full size of the tablet, the error "WebGL is not supported by your browser" is displayed.  The same result happens for this user as well on the example pages [here](https://plot.ly/javascript/3d-surface-plots/).

archmoj:
Which browser (& version) you are using?

RaoulDuke-Esq:
Browser is Safari on iPadOS 13.3

archmoj:
Replicated on:
![Screenshot from 2020-01-22 08-48-54](https://user-images.githubusercontent.com/33888540/72899564-3eb5a580-3cf4-11ea-8506-41c685f9c69c.png)

Using Safari only (Chrome works well),

Showing:
![Screenshot from 2020-01-22 08-45-28](https://user-images.githubusercontent.com/33888540/72899572-42e1c300-3cf4-11ea-83fb-8974e58fc6c2.png)

tsangiotis:
Having the same problem in iPad Air 10.5 (2019)

archmoj:
Error: `gl-fbo: Framebuffer incomplete attachment` during `createPlot`: https://github.com/plotly/plotly.js/blob/e7f38c0dc7b789e53055e5bc693f349d99a594f0/src/plots/gl3d/scene.js#L137.

archmoj:
Got it to work by internally forcing `isMobile=true` in `plotly.js` and `gl-plot3d`. 
Here is a [Demo](https://codepen.io/MojtabaSamimi/pen/oNgKWrg?editors=0010).
The `is-mobile` module needs a patch to detect iPad Pro 12.9 (13) and iPad 7th (13) when using Safari.
Related: https://github.com/juliangruber/is-mobile/issues/17

archmoj:
Also see: https://stackoverflow.com/questions/58019463/how-to-detect-device-name-in-safari-on-ios-13-while-it-doesnt-show-the-correct

PR #4542, opened 2020-02-02
Match createFBO arguments with gl-fbo arguments

archmoj:
To make debuging easier (e.g. for #4502) , this PR matches calls from `gl-plot2d` and `gl-plot3d` to create frame buffer object to basic `createFBO` arguments as listed here: 
https://github.com/stackgl/gl-fbo/blob/d8fc2fecb701097df1954f7268a206165d634609/fbo.js#L377.

@etpinard


etpinard:
Thanks! :dancer:

PR #4546, opened 2020-02-03
Fix 3-D config for webgl buffer on iPad Pro & iPad 7th + iOs v13 + Safari 

### LINKED ISSUE #4502: iPad Pro 12.9 (13) and iPad 7th (13) Not Detected Properly using Safari

archmoj:
Fixes #4502 | [Demo](https://codepen.io/MojtabaSamimi/pen/gObVddE?editors=1000)

@plotly/plotly_js 
cc: @jackparmer

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.3.0",
     "gl-plot2d": "^1.4.3",
-    "gl-plot3d": "^2.4.1",
+    "gl-plot3d": "^2.4.2",
     "gl-pointcloud2d": "^1.0.2",
     "gl-scatter3d": "^1.2.2",
     "gl-select-box": "^1.0.3",
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -15,7 +15,6 @@ var createPlot = glPlot3d.createScene;
 
 var getContext = require('webgl-context');
 var passiveSupported = require('has-passive-events');
-var isMobile = require('is-mobile')({ tablet: true });
 
 var Registry = require('../../registry');
 var Lib = require('../../lib');
@@ -30,6 +29,32 @@ var project = require('./project');
 var createAxesOptions = require('./layout/convert');
 var createSpikeOptions = require('./layout/spikes');
 var computeTickMarks = require('./layout/tick_marks');
+
+var isMobile = require('is-mobile');
+var tablet = isTablet();
+
+function isTablet() {
+    var navigator = window.navigator;
+    var userAgent = navigator.userAgent;
+
+    var result = isMobile({
+        ua: userAgent,
+        tablet: true
+    });
+
+    // handle iPad pro or iPad with iOs 13 using Safari
+    // see https://github.com/plotly/plotly.js/issues/4502
+    if(
+        result === false &&
+        userAgent.indexOf('Macintosh') !== -1 &&
+        userAgent.indexOf('Safari') !== -1 &&
+        navigator.maxTouchPoints > 1
+    ) {
+        result = true;
+    }
+
+    return result;
+}
 
 
 var STATIC_CANVAS, STATIC_CONTEXT;
@@ -96,7 +121,7 @@ proto.tryCreatePlot = function() {
         canvas: scene.canvas,
         gl: scene.gl,
         glOptions: {
-            preserveDrawingBuffer: isMobile,
+            preserveDrawingBuffer: tablet,
             premultipliedAlpha: true,
             antialias: true
         },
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.3.0",
     "gl-plot2d": "^1.4.3",
-    "gl-plot3d": "^2.4.1",
+    "gl-plot3d": "^2.4.2",
```
archmoj: https://github.com/gl-vis/gl-plot3d/pull/27

src/plots/gl3d/scene.js
```diff
@@ -31,6 +30,32 @@ var createAxesOptions = require('./layout/convert');
...
+        tablet: true
+    });
+
+    // handle iPad pro or iPad with iOs 13 using Safari
+    // see https://github.com/plotly/plotly.js/issues/4502
+    if(
+        result === false &&
+        userAgent.indexOf('Macintosh') !== -1 &&
+        userAgent.indexOf('Safari') !== -1 &&
+        navigator.maxTouchPoints > 1
```
etpinard: How did you find this solution?

archmoj: Inspired by info on https://stackoverflow.com/questions/58019463/how-to-detect-device-name-in-safari-on-ios-13-while-it-doesnt-show-the-correct

src/plots/gl3d/scene.js
```diff
@@ -31,6 +30,32 @@ var createAxesOptions = require('./layout/convert');
 var createSpikeOptions = require('./layout/spikes');
 var computeTickMarks = require('./layout/tick_marks');
 
+var isMobile = require('is-mobile');
+var tablet = isTablet();
+
+function isTablet() {
+    var navigator = window.navigator;
```
etpinard: Does `is-mobile` setup a guard for cases when `window.navigator` is undefined?

src/plots/gl3d/scene.js
```diff
@@ -31,6 +30,32 @@ var createAxesOptions = require('./layout/convert');
...
+    var userAgent = navigator.userAgent;
```
etpinard: and similarly here, if `is-mobile` guards against undefined `navigator.userAgent`, we should too here.

src/plots/gl3d/scene.js
archmoj: Good call.

src/plots/gl3d/scene.js
archmoj: Good call.

src/plots/gl3d/scene.js
archmoj: I don't see any check there concerning `navigator`.

src/plots/gl3d/scene.js
archmoj: Revised in https://github.com/plotly/plotly.js/pull/4546/commits/436fca46dad2226913b07f3ab0d01b04cb0ad03c.

src/plots/gl3d/scene.js
etpinard: OK, can you add one just to be sure?

etpinard: (I wouldn't someone running plotly.js in some sort of weird environment have their app break in a patch release)

archmoj: Actually it is added in https://github.com/plotly/plotly.js/pull/4546/commits/436fca46dad2226913b07f3ab0d01b04cb0ad03c here:
```js
if(!ua && typeof navigator !== 'undefined') ua = navigator.userAgent;
```

etpinard: Sure, we can still can breakage on this line below:

```js
navigator.maxTouchPoints > 1
```

if `window.navigator` isn't defined.

archmoj: Addressed in https://github.com/plotly/plotly.js/pull/4546/commits/aa28ebafd53a0fc5cb34a54b5f0718cb0aad4f43.

etpinard:
Sweet. Thanks very much

:dancer: - once tests pass!

archmoj:
@etpinard Thanks for the review.
After this I would open a PR on `is-mobile` to possibly simply this for future.

archmoj:
> @etpinard Thanks for the review.
> After this I would open a PR on `is-mobile` to possibly simply this for future.

PR to `is-mobile`: https://github.com/juliangruber/is-mobile/pull/22

Issue #3640, opened 2019-03-15
»Webgl is not supported by your browser« in Brave, although it is supported


Xjs:
Opening https://plot.ly/javascript/3d-scatter-plots/ in the Brave browser (https://brave.com, it's based on Chromium) yields »Webgl is not supported by your browser - visit http://get.webgl.org for more info«. Going to http://get.webgl.org in Brave, however, says »Your browser supports WebGL« and successfully displays an animation.

It would be nice if plotly WebGL plots would work in Brave.

archmoj:
It is working by disabling `Brave Shields`.
![Brave](https://user-images.githubusercontent.com/33888540/54437626-6965c380-470b-11e9-8573-5b930928f40f.png)

Xjs:
I see. Can confirm that allowing »All Device Recognition« it can be viewed. However, I think this shouldn't be the default state. Do I understand it correctly that Plotly tries to identify the browser and if it can't, it won't attempt to display WebGL? Or do you think this is an issue with Brave wrongly classifying something Plotly does?

archmoj:
Thanks @Xjs for testing that out. And glad that it worked.

<i>Regarding your first question: </i>
>  Do I understand it correctly that Plotly tries to identify the browser and if it can't, it won't attempt to display WebGL? 

No that is not the case.

<i>And in respect to the second question: </i>
> Or do you think this is an issue with Brave wrongly classifying something Plotly does?

It may be the case. I'll leave it to my colleague @etpinard who is super knowledgeable about browsers as well as `Plotly.js`.

hoerup:
I have the same problem with Firefox 65

etpinard:
Is this still a problem @Xjs ?

Xjs:
With »device recognition blocked« this is still a problem. I'm running Brave Dev `Version 0.66.71 Chromium: 75.0.3770.38 (Official Build) dev (64-bit)`.

hoerup:
Still a problem with FF 67 on fedora 29

mosiman:
Same deal here, but FF 69 on Arch. I'm using Wayland, so I'm not sure if that could be relevant to anything. Also getting `Framebuffer not complete` in the console. (`https://get.webgl.org` works fine)

tvst:
I see this on Chrome on Ubuntu. Plotly WebGL used to work, by the way. So this is a regression. I just don't know _whose_ regression it is...

[...]

PR #4548, opened 2020-02-04
Bump is-mobile to handle iOS 13 + iPad case

archmoj:
Thanks to @juliangruber, now that [a new `is-mobile` minor is out](https://github.com/juliangruber/is-mobile/pull/22#issuecomment-581949135), we could simply let this module detect iOS13+iPad case and remove the logic added in #4546.

@plotly/plotly_js

PR #4549, opened 2020-02-04
Detect correct device in the second attempt to create 3-D scene

archmoj:
Follow up of  and 4546 and #4548, and in regards with a number of comments from users experiencing webgl setup e.g.
https://github.com/plotly/plotly.js/issues/3640#issuecomment-505355009
https://github.com/plotly/plotly.js/issues/3640#issuecomment-474091716
https://github.com/plotly/plotly.js/issues/3640#issuecomment-529726285
https://github.com/plotly/plotly.js/issues/3640#issuecomment-566523569
this PR uses a different webgl config for `preserveDrawingBuffer` in the second attempt to create `gl3d` scene. With this in place the graph should be displayed instead of a webgl error namely 
`Error: WebGL warning: checkFramebufferStatus: Framebuffer not complete` 
even if `is-mobile` module was unable to detect the correct device.

[Demo](https://codepen.io/MojtabaSamimi/pen/qBdWVjB?editors=1000)

@plotly/plotly_js 
cc: @hoerup

### DIFF
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -91,8 +91,9 @@ function Scene(options, fullLayout) {
 
 var proto = Scene.prototype;
 
-proto.tryCreatePlot = function() {
+proto.prepareOptions = function() {
     var scene = this;
+
     var opts = {
         canvas: scene.canvas,
         gl: scene.gl,
@@ -132,20 +133,37 @@ proto.tryCreatePlot = function() {
         opts.canvas = STATIC_CANVAS;
     }
 
-    var failed = 0;
+    return opts;
+};
+
+proto.tryCreatePlot = function() {
+    var scene = this;
+
+    var opts = scene.prepareOptions();
+
+    var success = true;
 
     try {
         scene.glplot = createPlot(opts);
     } catch(e) {
-        failed++;
-        try { // try second time to fix issue with Chrome 77 https://github.com/plotly/plotly.js/issues/4233
-            scene.glplot = createPlot(opts);
-        } catch(e) {
-            failed++;
+        if(scene.staticMode) {
+            success = false;
+        } else { // try second time
+            // make fresh object for options
+            opts = scene.prepareOptions();
+
+            try {
+                // invert preserveDrawingBuffer setup which could be resulted from is-mobile not detecting the right device
+                opts.glOptions.preserveDrawingBuffer = !opts.glOptions.preserveDrawingBuffer;
+
+                scene.glplot = createPlot(opts);
+            } catch(e) {
+                success = false;
+            }
         }
     }
 
-    return failed < 2;
+    return success;
 };
 
 proto.initializeGLCamera = function() {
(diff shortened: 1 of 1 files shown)


src/plots/gl3d/scene.js
```diff
@@ -132,20 +133,34 @@ proto.tryCreatePlot = function() {
...
-        try { // try second time to fix issue with Chrome 77 https://github.com/plotly/plotly.js/issues/4233
-            scene.glplot = createPlot(opts);
-        } catch(e) {
-            failed++;
+        if(scene.staticMode) {
+            success = false;
+        } else { // try second time
+            try {
+                // invert preserveDrawingBuffer setup which could be resulted from is-mobile not detecting the right device
+                isMobile = opts.glOptions.preserveDrawingBuffer = !opts.glOptions.preserveDrawingBuffer;
```
etpinard: Hmm, I'm not sure I like this. 

If the "main" problem here is `is-mobile` not doing its job in some cases, then we should try to fix `is-mobile` instead of working around it behind the scenes.

At the very (very) least we should `Lib.warn` something here to let the user (and the devs) know what's going on, and gather some data on the situations where `is-mobile` is buggy.

----

To be more robust, what if instead of flipping `preserveDrawingBuffer` here, we allow users (via a new config option) to set `preserveDrawingBuffer` to `true` to essentially override `is-mobile` when they want?

archmoj: > To be more robust, what if instead of flipping `preserveDrawingBuffer` here, we allow users (via a new config option) to set `preserveDrawingBuffer` to `true` to essentially override `is-mobile` when they want?

The users may not know in advance the correct setup for the devices.

archmoj: > Hmm, I'm not sure I like this.
> 
> If the "main" problem here is `is-mobile` not doing its job in some cases, then we should try to fix `is-mobile` instead of working around it behind the scenes.
 
> At the very (very) least we should `Lib.warn` something here to let the user (and the devs) know what's going on, and gather some data on the situations where `is-mobile` is buggy.

Good call. Done in https://github.com/plotly/plotly.js/pull/4549/commits/a26e5e2262437a96cad71da9cc4b2177064b1334.

etpinard: > The users may not know in advance the correct setup for the devices.

Yes, I know that. But do you agree that the best way to fix this problem would be to _fix_ `is-mobile`?

etpinard: .... but now with the `Lib.warn` message I guess we can expect some users to tell us about it - which may lead us to fix `is-mobile` for these cases. 

I'm happy with this solution. That `Lib.warn` message is very clear.

etpinard:
:dancer:

src/plots/gl3d/scene.js
archmoj: I totally agree we should try fixing `is-mobile` to help detect the right device.
Still this PR helps detect/by-pass such edge cases.

archmoj:
Also to mention this PR fixed #3640 | [Demo](https://codepen.io/MojtabaSamimi/pen/JjdYapq?editors=1000)

richardoptibrium:
I'm running 1.54.2 but it's still an issue.

PR #4554, opened 2020-02-05
Apply utf-8 charset in test_dashboard

archmoj:
Fixes non-ASCII characters in test_dashboard.

@plotly/plotly_js


etpinard:
:dancer: - thanks!

PR #4566, opened 2020-02-10
Fix download images to match live gl3d plot

### LINKED ISSUE #4378: difference between live 3-D graph with downloaded PNG
Is this a bug?

Bunny

https://rreusser.github.io/plotly-mock-viewer/#gl3d_bunny

archmoj:
Fixes #4378 | [Demo](https://codepen.io/MojtabaSamimi/pen/dyoozBN?editors=1000)

@plotly/plotly_js 
cc: @alexcjohnson


src/plots/gl3d/scene.js
```diff
@@ -1003,6 +1003,23 @@ function flipPixels(pixels, w, h) {
...
+    for(var i = 0; i < h; ++i) {
+        for(var j = 0; j < w; ++j) {
+            var k = 4 * (w * i + j);
+
+            var a = pixels[k + 3]; // alpha
+            if(a > 0) {
+                var q = 255 / a;
+
+                for(var l = 0; l < 3; ++l) { // RGB
+                    pixels[k + l] = Math.min(q * pixels[k + l], 255);
```
etpinard: Can explain why we need to apply this transform when exporting images?

archmoj: Unlike openGL, webGL `premultipliedAlpha` in enabled by default.
To get correct RGB values from the webgl context it was needed to devide RGB values by alpha.

etpinard: Ok, what would happen if we instead disable `premultipliedAlpha` on image exports?

archmoj: Good question; but that didn't work: [demo](https://codepen.io/MojtabaSamimi/pen/bGddoLJ?editors=1000).

etpinard:
:dancer:

PR #4568, opened 2020-02-10
Fix for displaying narrow bars in interactive mode

### LINKED ISSUE #4567: Regression of bar plot axis scaling between 1.51.1 and 1.51.2
@archmoj 

https://github.com/plotly/plotly.js/commit/96ef28a76de609ca11adf475adba59a574ce3e59#diff-41fd60f1e6a06a5340748d3f4b3754e6R218-R225

This commit broke (my) bar plots which are **updated in chunks**. Horizontal axis of the bar plot stopped rescaling correctly between 1.51.1 and 1.51.2.
This has been working for years so this should be considered as a regression.

By removing this if statement it starts working again: https://github.com/plotly/plotly.js/blob/acc9c21c0239b4500b9787ab92203327b3abd033/src/traces/bar/plot.js#L218

**New behaviour:**
image

**Version <= 1.51.1 (or above if-statement removed in 1.51.2):**
image

If required, I'll try to construct a simple example to reproduce.

### LINKED ISSUE #4573: inconsistent interactive bar display
Missing a number of bars in this [demo](https://codepen.io/MojtabaSamimi/pen/QWbjyNM?editors=0010).
This is a bug different from #4567 while in the codepen v1.52.0 is used (not v1.52.2).

archmoj:
Fixes #4567 | [Demo: before vs after](https://codepen.io/MojtabaSamimi/pen/YzXXYWy?editors=1000)
and fixes #4573 | [before](https://codepen.io/MojtabaSamimi/pen/QWbjyNM?editors=0010) vs [after](https://codepen.io/MojtabaSamimi/pen/VwLveKw?editors=1000)

@plotly/plotly_js
cc: @brubbel


src/traces/bar/plot.js
```diff
@@ -215,14 +214,10 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts, makeOnCompleteCallback)
...
-                    x1 = fixpx(x1, x0);
-                }
-                if(spansVertical) {
-                    y0 = fixpx(y0, y1);
-                    y1 = fixpx(y1, y0);
-                }
+                x0 = fixpx(x0, x1, isHorizontal);
+                x1 = fixpx(x1, x0, isHorizontal);
+                y0 = fixpx(y0, y1, !isHorizontal);
+                y1 = fixpx(y1, y0, !isHorizontal);
```
etpinard: Can you explain how this commit changes the logic.

archmoj: All four positions i.e. x0, x1, y0, y1 (not only two of hem) should possibly be rounded by `fixpx` function.
In case the bar is `horizontal` the `fixpx` function may decide not to round (floor and ciel) zero spans i.e. when `x0 === x1`.
Similarly, when a bar is `vertical` the `fixpx` will return early for zero spans i.e. when `y0 === y1`.
However, in case of having narrow lines e.g. a `vertical` bar the `fixpx` should round values to upper and lower values even if when `x0 === x1`. Also when a `horizontal` bar encounters `y0 === y1` narrow lines should be displayed simalr to v1.51.1 and earlier versions.

archmoj:
@etpinard I am working on few more commits for this PR.

src/traces/bar/plot.js
```diff
@@ -149,14 +149,13 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts, makeOnCompleteCallback)
...
-            if(isBlank && shouldDisplayZeros && helpers.getLineWidth(trace, di) && (isHorizontal ? x1 - x0 === 0 : y1 - y0 === 0)) {
+            if(isBlank && shouldDisplayZeros &&
+                helpers.getLineWidth(trace, di) &&
+                (isHorizontal ? x0 === x1 : y0 === y1)) {
                 isBlank = false;
             }
             di.isBlank = isBlank;
 
-            if(isBlank && isHorizontal) x1 = x0;
-            if(isBlank && !isHorizontal) y1 = y0;
```
archmoj: Those two lines added in https://github.com/plotly/plotly.js/commit/f49828a05f?short_path=73806af#diff-73806afb28fb251e00e170f2218e263a appear to cause some rounding problem. @antoinerg do you recall why they were needed?
Now that they removed the new test added in https://github.com/plotly/plotly.js/pull/4568/commits/7f3fb34fc3300f7670a9d92e0054802efb105ac2 with defined range pass.

src/traces/bar/plot.js
etpinard: Ok, can you try to write a comment explaining that above this line:

```js
if(hideZeroSpan && v === vc) return v;
```

??

archmoj: Done in https://github.com/plotly/plotly.js/pull/4568/commits/c1ee8c2279058d61e7075a9967899a038b75e45e.

src/traces/bar/plot.js
etpinard: Hmm right, this may have to do with transitions.

@antoinerg can you send us links to your old bar transition examples?

archmoj: Putting the logic back in https://github.com/plotly/plotly.js/pull/4568/commits/4c323bcd03ea4cd768c4ce46f1e786aa3113bc51; but only when there is transition | [Demo](https://codepen.io/MojtabaSamimi/pen/Exjjqqx?editors=1000).

etpinard: Can I ask why? Did you notice some less-than-ideal behaviour w/o this new `withTransition` logic?

archmoj: [Here](https://codepen.io/MojtabaSamimi/pen/VwLvwyq?editors=1000) when applying the very first commit of this PR, the bars were disappeared after third zoom out (on the graphs on the right).

etpinard: Ok great, can you try to lock that down in a test?

etpinard: ... and can you add a comment above:

```js
            if(isBlank && withTransition) {
                if(isHorizontal) {
                    x1 = x0;
                } else {
                    y1 = y0;
                }
            }

```

archmoj: > Ok great, can you try to lock that down in a test?

That one was added in  https://github.com/plotly/plotly.js/pull/4568/commits/7f3fb34fc3300f7670a9d92e0054802efb105ac2.

etpinard: Hmm. I don't see a test locking down your https://github.com/plotly/plotly.js/pull/4568/commits/4c323bcd03ea4cd768c4ce46f1e786aa3113bc51 commit.

archmoj: > ... and can you add a comment above:
> 
> ```js
>             if(isBlank && withTransition) {
>                 if(isHorizontal) {
>                     x1 = x0;
>                 } else {
>                     y1 = y0;
>                 }
>             }
> ```

Done in https://github.com/plotly/plotly.js/pull/4568/commits/8ebc0301f1eaf3f4e54eeebab7923bc2bf6d96be.

src/traces/bar/plot.js
```diff
@@ -157,6 +157,7 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts, makeOnCompleteCallback)
             }
             di.isBlank = isBlank;
 
+            // for empty bars, ensure start and end positions are equal when having transition
```
etpinard: What about:

```
for empty bars, make start and end positions equal to ensure a smooth transition
```

src/traces/bar/plot.js
archmoj: > Hmm. I don't see a test locking down your [4c323bc](https://github.com/plotly/plotly.js/commit/4c323bcd03ea4cd768c4ce46f1e786aa3113bc51) commit.

Added in https://github.com/plotly/plotly.js/pull/4568/commits/494775bfccc1e517917ecabb7cb494c3412d9e53.

archmoj:
Please note that bug reported in #4573 is also fixed in this PR.

src/traces/bar/plot.js
```diff
@@ -140,25 +140,29 @@ function plot(gd, plotinfo, cdModule, traceLayer, opts, makeOnCompleteCallback)
...
+                    !isNumeric(y0) ||
+                    !isNumeric(y1)
+                );
+            }
+
+            // record isBlank
             di.isBlank = isBlank;
 
-            // for empty bars, ensure start and end positions are equal when having transition
-            if(isBlank && withTransition) {
```
archmoj: With revised logic, now there is no need for handling transitions differently.

etpinard:
Awesome work. Thanks very much for adding those transition tests!

I hope this is the last of our blank bar fixup. :dancer:

PR #4578, opened 2020-02-12
Preserve gl3d scene aspectratio after orthographic scroll zoom

### LINKED ISSUE #4514: Orthographic projection + zoom breaks hover and Plotly.restyle
See https://codepen.io/luthaf/pen/oNgmavz

After zooming, the hover label no longer correspond to the mouse cursor position:
Capture du 2020-01-23 17-34-35

Additionally, if I try to call Plotly.restyle after zooming, the camera resets to the default position. I don't know if this is intended or accidental.

---

Maybe related to #4239

archmoj:
Fixes #4514 restyle part
After refactoring parts of the gl3d/scene code related to computing `aspectratio`,
this PR fixes the bug in commit https://github.com/plotly/plotly.js/commit/e29aceb930208840dc27e5d2bdac782332fab505.
[Demo](https://codepen.io/MojtabaSamimi/pen/vYONaww?editors=1000) | Please change scales using scroll, then select a point. It shouldn't display initial scales.

@plotly/plotly_js

### DIFF
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -471,12 +471,12 @@ proto.recoverContext = function() {
 var axisProperties = [ 'xaxis', 'yaxis', 'zaxis' ];
 
 function computeTraceBounds(scene, trace, bounds) {
-    var sceneLayout = scene.fullSceneLayout;
+    var fullSceneLayout = scene.fullSceneLayout;
 
     for(var d = 0; d < 3; d++) {
         var axisName = axisProperties[d];
         var axLetter = axisName.charAt(0);
-        var ax = sceneLayout[axisName];
+        var ax = fullSceneLayout[axisName];
         var coords = trace[axLetter];
         var calendar = trace[axLetter + 'calendar'];
         var len = trace['_' + axLetter + 'length'];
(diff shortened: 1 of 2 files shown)


src/plots/gl3d/scene.js
```diff
@@ -725,45 +726,44 @@ proto.plot = function(sceneData, fullLayout, layout) {
...
-
-            aspectRatio = axesScaleRatio;
-        } else {
-            /*
-             * USE EQUAL MODE WHEN AXIS RANGE DIMENSIONS ARE HIGHLY UNEQUAL
-             */
-            aspectRatio = [1, 1, 1];
-        }
-    } else if(fullSceneLayout.aspectmode === 'cube') {
+    var aspectmode = scene._aspectmode || fullSceneLayout.aspectmode;
```
etpinard: What's the difference between `scene._aspectmode` and `fullSceneLayout.aspectmode` ?

archmoj: `scene._aspectmode` is an internal variable which is set after `aspectratio` values are computed.

etpinard: Are there situations where `scene._aspectmode` and `fullSceneLayout.aspectmode` aren't the same?

archmoj: They are. Namely after using scroll zoom.

etpinard: Ok, but shouldn't the changes to `aspectmode` during scroll zoom propagate to `fullLayout` ?

archmoj: I initially tried to achieve that on another branch using `relayout` updates. 
Since tests started failing, I thought that may not be ideal to override `fullLayout` in this case.

etpinard: I'm curious. Which tests fail?

archmoj: > Are there situations where `scene._aspectmode` and `fullSceneLayout.aspectmode` aren't the same?

OK. This time I was able to make that work without a temporary variable: https://github.com/plotly/plotly.js/pull/4578/commits/1828797e27f6394dc4f27a7be1fe0f6a762d5037

etpinard:
Brilliant. :dancer: !

PR #4579, opened 2020-02-13
Include gl3d scene.aspectmode changes in relayout updates

archmoj:
Follow up of #4578 this PR adds gl3d `scene.aspectmode` changes in relayout updates.

[Demo](https://codepen.io/MojtabaSamimi/pen/ExjPNRV?editors=0010)

@plotly/plotly_js

### DIFF
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -783,15 +783,15 @@ proto.plot = function(sceneData, fullLayout, layout) {
      */
     scene.glplot.setAspectratio(fullSceneLayout.aspectratio);
 
-    // save 'initial' camera view settings for modebar button
+    // save 'initial' aspectratio & aspectmode view settings for modebar buttons
     if(!scene.viewInitial.aspectratio) {
         scene.viewInitial.aspectratio = {
             x: fullSceneLayout.aspectratio.x,
             y: fullSceneLayout.aspectratio.y,
             z: fullSceneLayout.aspectratio.z
         };
-
-        // also keep track of aspectmode here
+    }
+    if(!scene.viewInitial.aspectmode) {
         scene.viewInitial.aspectmode = fullSceneLayout.aspectmode;
     }
 
--- a/src/plots/gl3d/scene.js
+++ b/src/plots/gl3d/scene.js
@@ -155,7 +155,7 @@ proto.tryCreatePlot = function() {
                     'webgl setup failed possibly due to',
                     isMobile ? 'disabling' : 'enabling',
                     'preserveDrawingBuffer config.',
-                    'The device may not be supported by isMobile module!',
+                    'The device may not be supported by is-mobile module!',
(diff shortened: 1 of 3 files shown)


src/plots/gl3d/scene.js
```diff
@@ -219,6 +219,12 @@ proto.initializeGLPlot = function() {
         if(scene.isAspectChanged(layout)) {
             // scene updates
             update[scene.id + '.aspectratio'] = scene.glplot.getAspectratio();
+
+            if(layout[scene.id].aspectmode !== 'manual') {
```
etpinard: How do we get back to the original `aspectmode` ?

archmoj: Should we?

etpinard: When resetting the view (e.g via the modebar buttons), yes.

archmoj: Good call. I think we have some tests related to that.

archmoj: https://github.com/plotly/plotly.js/blob/4d24c36c93d1d1cca4ad4a195cbaf43f53e963fe/test/jasmine/tests/plot_api_test.js#L394-L411

etpinard: So, to be clear, once a user scrolls into a orthographic scene, the `aspectmode` is turned to `'manual'` and the only way to get back to original `aspectmode` is by resetting the view?

archmoj: Or we should keep that in `viewInitial` props?

etpinard: How is it handled currently?

archmoj: `apsectmode` added to `viewInitial` in https://github.com/plotly/plotly.js/pull/4579/commits/1fdc49c7c22dc3b73f7a0f3572b675ea2457b5b7 and tests are added in https://github.com/plotly/plotly.js/pull/4579/commits/6ae280967e5e2e91762d28cc77290516735efc83.

src/plots/gl3d/scene.js
```diff
@@ -790,6 +790,9 @@ proto.plot = function(sceneData, fullLayout, layout) {
             y: fullSceneLayout.aspectratio.y,
             z: fullSceneLayout.aspectratio.z
         };
+
+        // also keep track of aspectmode here
+        scene.viewInitial.aspectmode = fullSceneLayout.aspectmode;
```
etpinard: Shouldn't this be

```js
if(!scene.viewInitial.aspectmode) {
  scene.viewInitial.aspectmode = fullSceneLayout.aspectmode
}
```

that way we don't override `viewInitial` on updates?

archmoj: Good call.
We could. Done in https://github.com/plotly/plotly.js/pull/4579/commits/c730f1276ad167adf44b13cf8482fb317d88843c.

etpinard:
Thanks - :dancer: !

PR #4580, opened 2020-02-13
Ensure text fits inside sunburst sectors with zero values

### LINKED ISSUE #4569: sunburst text does not fit inside for 0-valued sectors
See https://codepen.io/emmanuelle-plotly/pen/JjdYPBN?editable=true, for the sector with 0 value the text label is still displayed even when setting `uniformtext` parameters. Changing the value to a very small number (e.g. 1.e-7) fixes the issue. Raised by https://community.plot.ly/t/help-with-sunburst-using-plotly-express-and-path/34823/

archmoj:
Fixes #4569 | [Demo before vs after](https://codepen.io/MojtabaSamimi/pen/YzXwPmq?editors=1000)

@plotly/plotly_js 
cc: @emmanuelle


src/traces/pie/plot.js
```diff
@@ -559,18 +559,31 @@ function prerenderTitles(cdModule, gd) {
...
 function transformInsideText(textBB, pt, cd0) {
-    var textDiameter = Math.sqrt(textBB.width * textBB.width + textBB.height * textBB.height);
+    var r = cd0.r || pt.rpx1;
+    var rInscribed = pt.rInscribed;
+
+    var isEmpty = pt.startangle === pt.stopangle;
+    if(isEmpty) {
+        return {
+            rCenter: 1 - rInscribed,
+            scale: 0,
```
etpinard: So, by _fitting_ inside text of zero-value sunburst sectors, you mean setting their `scale` transform to `0`, correct?

archmoj: Correct.

etpinard:
Nicely done :dancer:

src/traces/pie/plot.js
archmoj: The other option (not implemented here) may be not to draw sectors with zero values.

etpinard: Or set them with `display: 'none'`, to keep the sector <--> node map 1:1

etpinard: ... but really, this is fine. It allows us to keep all the logic in `transformInsideText`.

PR #4601, opened 2020-02-25
Fix to make identical bundles on different nodes

archmoj:
Possibly since v.1.36.1 a random number was attached to one of the GLSL variables of shader programs used in `gl-line3d` module e.g. `encode_float_1540259130(pixelArcLength)` resulting in non-deterministic bundles as reported in https://github.com/plotly/streambed/issues/13816.

This PR fixes this issue by removing one linking step within `gl-line3d` module.

@plotly/plotly_js 
cc: @nicolaskruchten @dmt0 @tarzzz

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "gl-contour2d": "^1.1.6",
     "gl-error3d": "^1.0.15",
     "gl-heatmap2d": "^1.0.5",
-    "gl-line3d": "git://github.com/gl-vis/gl-line3d.git#1514fbaf898e5929c77b03dfb5ec6d08edbcd80a",
+    "gl-line3d": "1.2.0",
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.3.0",
     "gl-plot2d": "^1.4.3",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "gl-contour2d": "^1.1.6",
     "gl-error3d": "^1.0.15",
     "gl-heatmap2d": "^1.0.5",
-    "gl-line3d": "^1.1.11",
+    "gl-line3d": "git://github.com/gl-vis/gl-line3d.git#66628995ce0cd532dba0f526e2f0357c7acea5c8",
     "gl-mat4": "^1.2.0",
     "gl-mesh3d": "^2.3.0",
     "gl-plot2d": "^1.4.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
     "gl-contour2d": "^1.1.6",
     "gl-error3d": "^1.0.15",
     "gl-heatmap2d": "^1.0.5",
-    "gl-line3d": "^1.1.11",
+    "gl-line3d": "git://github.com/gl-vis/gl-line3d.git#1514fbaf898e5929c77b03dfb5ec6d08edbcd80a",
```
archmoj: https://github.com/gl-vis/gl-line3d/pull/12

etpinard:
So, if I understand correctly, browserify + glslify currently bundles the [`encode_float`](https://github.com/mikolalysenko/glsl-read-float/blob/e827ba9ad653d478752c2af4bec8edd2d16a1e41/index.glsl#L4-L40) function with some additional hash (e.g. `encode_float_1540259130` or `encode_float_1604150559`) leading to bundles with potentially different checksum output. @archmoj is this correct?

If so, what's special about the shader program in `glsl-read-float`? Why don't other glsl packages exhibit this behaviour? Shouldn't we instead try to solve this problem at the glslify-level? Maybe we can set a flag that makes the additional hash reproducible?

archmoj:
> So, if I understand correctly, browserify + glslify currently bundles the [`encode_float`](https://github.com/mikolalysenko/glsl-read-float/blob/e827ba9ad653d478752c2af4bec8edd2d16a1e41/index.glsl#L4-L40) function with some additional hash (e.g. `encode_float_1540259130` or `encode_float_1604150559`) leading to bundles with potentially different checksum output. @archmoj is this correct?

Correct.

archmoj:
> If so, what's special about the shader program in `glsl-read-float`? Why don't other glsl packages exhibit this behaviour? 

`gl-line3d` is the only module that uses `glsl-read-float` module. And there is a linking problem between `glsl-read-float` and `gl-line3d` due to casting doubles to shorts between two shaders on different hardware.

archmoj:
> Shouldn't we instead try to solve this problem at the glslify-level? 

We could. But that require spending more time debugging. 
I suggest we open an issue on `gl-line3d` or `glsl-read-float` when/if this PR merged.

etpinard:
> I suggest we open an issue on `gl-line3d` or `glsl-read-float` when/if this PR merged.

yep, good idea.

etpinard:
:dancer:

PR #4610, opened 2020-03-02
check-node-version: accept any nodejs version in 12.x

antoinerg:
I was wondering if we could loosen the requirement for nodejs from `12.13` to `12` to accommodate developers that may not be running this specific version and may not want to install nvm. Do you see any risk in doing that? Also, the latest 12 version is `12.16.1`.

cc @alexcjohnson @etpinard

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "start-image_viewer": "node devtools/image_viewer/server.js",
     "start": "npm run start-test_dashboard",
     "baseline": "node tasks/baseline.js",
-    "preversion": "check-node-version --node 12.13 --npm 6.13 && npm-link-check && npm ls --prod",
+    "preversion": "check-node-version --node 12 --npm 6.13 && npm-link-check && npm ls --prod",
     "version": "npm run build && git add -A dist src build",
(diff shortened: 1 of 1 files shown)


etpinard:
We have two options here. Either loosen the requirements or update the requirements now (and possibly every time there's a new Node version).

I highly doubt that having a requirement on the major version of Node will lead to any problems. In other words, this PR seems fine.

I can't say the same for `npm`. I think sticking to a requirement on the minor version of `npm` is the way to go, as minor updates have lead to changing `package-lock` behaviour in the past (less so recently though).

antoinerg:
> I highly doubt that having a requirement on the major version of Node will lead to any problems. In other words, this PR seems fine.

Good! I think so too

> I can't say the same for npm. I think sticking to a requirement on the minor version of npm is the way to go, as minor updates have lead to changing package-lock behaviour in the past (less so recently though).

`npm` is easy to install so I'm fine with a strict requirement there.

archmoj:
Thanks @etpinard for the review.
:dancer: 
@antoinerg please include this PR in the changelog for the upcoming release.

etpinard:
> please include this PR in the changelog for the upcoming release.

If you feel it's necessary, but we usually keep the Changelog for user-facing things only.

Issue #4210, opened 2019-09-23
[Feature Request] Visually skipped/warped range


Braintelligence:
Hello everyone, long time no new issue from my side 😛 

I have a feature request: I'd like to make something like this happen:

![image](https://user-images.githubusercontent.com/6761196/65419405-d961ae80-ddfe-11e9-9b83-b763e9ca4923.png)

Use-case are bar charts that don't start at zero. You want people to realize that you didn't actually properly start at zero to better show the difference on a smaller scale.

Formally for this example you could define that any zoom on the y-axis that doesn't start at 0 will show this special bit (as shown in my image file); if you start at 0 or less, this part will disappear.

Issue #1382, opened 2017-02-15
Should remove breaks for OHLC, Candlestick


mckelvin:
[The code to reproduce](https://codepen.io/mckelvin/pen/NdJpjx) is based on [simple-candlestick-chart-example](https://plot.ly/javascript/candlestick-charts/#simple-candlestick-chart-example) . The difference is:

**11th, 12th are not existed** in the dataset (simply because they're weekdays/non-trading days), but candlesticks in the chart should be continued, i.e.: the breaks of 11th, 12th should be removed like [what highcharts do](http://www.highcharts.com/stock/demo/candlestick).

This behaviour should be the default one for OHLC and Candlestick charts, otherwise it will become [terrible for real world dataset](https://plot.ly/~mckelvin/5/).

* * *

https://codepen.io/mckelvin/pen/ygwMQz is my workaround by using `tickvals` and `ticktext`. But there's another problem: **nticks doesn't work when tickmode=array** as mentioned in docs, there will be [too many tick labels](https://plot.ly/~mckelvin/3/#plot) if there're too many ticks.

etpinard:
Agreed. But, I don't like your tone. Saying that  plotly.js _should be_ like Highcharts isn't a very good argument. Thanks.

mckelvin:
@etpinard Thanks for your reply. Sorry but you might misunderstand me. I don't mean to argue for the feature as it's very common in OHLC/Candlestick charts and the link is just provided as a demo which contains the feature I expected. Both plotly and highcharts are great softwares but neither of them is a copycat.

etpinard:
@mckelvin no worries. Thanks for posting!

jackparmer:
Current proposal is to either add a new axis type:
`type: 'date-no-we'`

Or a flag for the axis object:
`show_weekend: false`

cc @jackwluo @cpsievert @santosjorge

Anyone out there interested in sponsoring this one? :clinking_glasses: 
https://plot.ly/products/consulting-and-oem/

jackwluo:
Implement is as a 'financial' flag for axis, since different markets have different holidays

jackparmer:
@jackwluo - That's an interesting point. Can you think of any online reference for which markets respect which holidays?

It looks like mpl and ggplot user have also come across this and have their own workarounds:

[ggplot](http://stackoverflow.com/questions/32893176/ggplot2-remove-blank-space-for-weekends-and-holidays-from-x-axis-dates)
[...]

PR #4614, opened 2020-03-03
Introduce cartesian axis breaks

etpinard:
which presents a solution for removing weekends, holidays and non-work hours as requested in https://github.com/plotly/plotly.js/issues/1382 and lays the groundwork from "visually skipped" ranges as in the screenshot in https://github.com/plotly/plotly.js/issues/4210.

----

In brief, this PR adds a `breaks` array attribute to cartesian axes. Each item inside breaks defines an axis break either using a 2-item `bounds` array or a free-length array of `values`. When using `bounds` on date axes, one can set a pattern to generate breaks e.g. over all weekends or non-work hours. When using `values`, one can set the "spread" of each value using `dvalue`. The inclusive/exclusive behaviour at the bounds is determined by the `operation` attributes (as used in filter transforms).

Some examples:

- breaks on linear x and y axes: https://codepen.io/etpinard/pen/eYNErKK?editors=1010
- breaks using `pattern` on date axis: https://codepen.io/etpinard/pen/MWwvGBV?editors=1010
- breaks using `values` on date axis: https://codepen.io/etpinard/pen/PoqKedG

To implement axis `breaks`,

- we first set coordinates that fall within the breaks to `BADNUM` (more info in https://github.com/plotly/plotly.js/commit/baf753aad7cb77d9dff47cf946a77d67ba2aca9e). Please note that this implies that lines remain continuous across a break if there's no (x,y) pt inside the break in question.
- sanitize the breaks present inside the axis range during `ax.setScale` (i.e. when the range is known) and adapt the `l2p` and `p2l` mappings (more info in https://github.com/plotly/plotly.js/commit/259eafa1e805def29df18289a0f0292ffda0cc43)
- Use the sum of the piecewise data-space ranges between the breaks (instead of simply `range[1] - range[0]`) for autorange padding, ticks placement computations and during drag interactions.

cc @archmoj @alexcjohnson @nicolaskruchten @antoinerg


src/plots/cartesian/axis_defaults.js
```diff
@@ -117,5 +121,68 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
...
+                if(rng[0] < rng[1]) {
+                    if(bnds[0] < rng[0] && bnds[1] > rng[1]) {
+                        itemOut.enabled = false;
+                        return;
+                    }
+                } else if(bnds[0] > rng[0] && bnds[1] < rng[1]) {
+                    itemOut.enabled = false;
+                    return;
+                }
+            }
```
archmoj: What about handling `autorange: 'reversed'` case?
In the following demos I tried to flip the axes which it didn't work.
https://codepen.io/MojtabaSamimi/pen/yLNzLra?editors=0010
https://codepen.io/MojtabaSamimi/pen/bGdoGyG?editors=0010
https://codepen.io/MojtabaSamimi/pen/QWbqWRj?editors=0010

etpinard: Ah right, I forgot to test those. They won't be handled here as this logic has to do with set (i.e. `autorange: false` ranges, but yeah I'll get this fixed. Thanks!

src/plots/cartesian/set_convert.js
```diff
@@ -486,6 +549,51 @@ module.exports = function setConvert(ax, fullLayout) {
...
         }
 
+        // set of "N" disjoint breaks inside the range
+        ax._breaks = [];
+        // length of these breaks in value space
+        ax._lBreaks = 0;
+        // l2p slope (same for all intervals)
+        ax._m2 = 0;
+        // set of l2p offsets (one for each of the (N+1) piecewise intervals)
+        ax._B = [];
```
archmoj: I am afraid using `_B` here may be too abstract. 
How about `_offBr` or other alternatives?

src/plots/cartesian/set_convert.js
```diff
@@ -486,6 +549,51 @@ module.exports = function setConvert(ax, fullLayout) {
             ax._b = -ax._m * rl0;
         }
 
+        // set of "N" disjoint breaks inside the range
+        ax._breaks = [];
+        // length of these breaks in value space
+        ax._lBreaks = 0;
```
archmoj: `ax._lenBr` may read better.

src/plots/cartesian/set_convert.js
etpinard: I can change it if you want, but `ax._B` refers to `ax._b` which is used on axes w/o breaks.

archmoj: Then adding a comment may be enough.

src/plots/cartesian/set_convert.js
```diff
@@ -565,6 +673,141 @@ module.exports = function setConvert(ax, fullLayout) {
...
+                                step = ONEDAY;
+
+                                t = r0 + r0PatternDelta * ONEHOUR -
+                                    r0Date.getUTCMinutes() * ONEMIN -
+                                    r0Date.getUTCSeconds() * ONESEC -
+                                    r0Date.getUTCMilliseconds();
+                                break;
+                        }
+
+                        while(t <= r1) {
```
archmoj: Could we add a finite test for `t` before entring this loop?
Is there any scenario (e.g. on reversed axes) in which step could be negative?

nicolaskruchten:
Looks great! Do we have any control over the visual aspect of the break? Can we put in a gap or a squiggle or something?

etpinard:
> Do we have any control over the visual aspect of the break? Can we put in a gap or a squiggle or something?

Not at the moment.

https://github.com/plotly/plotly.js/blob/187c93afe4f210ceef20a1e67aa66d36109cfec7/src/plots/cartesian/layout_attributes.js#L346-L375

nicolaskruchten:
OK. I feel like introducing breaks without visible gaps is a bit scary in other contexts than "drop weekends for business metrics" ... what's the LOE on visible gaps?

etpinard:
> what's the LOE on visible gaps?

Probably 1-2 days for gaps w/o squiggle on the axis. I suspect that getting everyone happy with the squiggle look/behaviour will take a few days on its own.

src/plots/cartesian/layout_attributes.js
```diff
@@ -248,6 +248,135 @@ module.exports = {
...
+            role: 'info',
+            items: [
+                {valType: 'any', editType: 'calc'},
+                {valType: 'any', editType: 'calc'}
+            ],
+            editType: 'calc',
+            description: [
+                'Sets the lower and upper bounds of this axis break.',
+                'Can be used with `operation` to determine the behavior at the bounds.',
+                'On *date* axes, it can be used `pattern`.'
```
nicolaskruchten: missing "with" before "pattern" ?

nicolaskruchten:
Zooming way in to the pattern date mock yields some odd tick labels: 

image

src/plots/cartesian/layout_attributes.js
```diff
@@ -248,6 +248,135 @@ module.exports = {
...
+            ].join(' ')
+        },
+        dvalue: {
+            // TODO could become 'any' to add support for 'months', 'years'
+            valType: 'number',
+            role: 'info',
+            editType: 'calc',
+            min: 0,
+            description: [
+                'Sets the spread of each `values` item.',
```
nicolaskruchten: I don't really understand what this means... "spread" is unclear in this context, can you clarify please?

etpinard: An axis break is an continuous interval, so when setting an axis break via a set of values, one needs to also specify the distance on the axis that one value corresponds to.

src/plots/cartesian/layout_attributes.js
etpinard: Good eye! Fixed in https://github.com/plotly/plotly.js/pull/4614/commits/76a265e75e7f6f67169710ec292b625ef9b8e172

src/plots/cartesian/layout_attributes.js
nicolaskruchten: As discussed, I think "size" is less confusing than "spread" and an inline example would help a lot :)

nicolaskruchten:
To capture what was discussed today: for 1.53 let's activate this feature for date axes only please. We'll need to think about axis gap and bar charts etc before turning this on on linear/log axes I think.

etpinard:
TODO: 

- [ ] coerce breaks only on `type: 'date'` axes, adapt tests and mocks!
- [ ] make sure axis breaks work on axes with `range[0] > range[1]`

see comments over on: https://github.com/plotly/plotly.js/compare/axis-breaks...axis-breaks-fixups

@archmoj will attempt to fix these tomorrow and on Monday the 9th. 

- [ ] better "first tick" algorithm e.g. what to do when the canonical first auto tick falls into a break? See https://github.com/plotly/plotly.js/pull/4614/commits/b5aaf92ce91e7f76101cfebd41ccfa101d747521
- [ ] better "overlapping tick in either side of an axis break fix" algorithm, see comment in https://github.com/plotly/plotly.js/pull/4614/commits/6bec94d3869b8af7a9260142e5127bc95a1662a8

I should be able to look at these when I get back Tuesday the 10th.

archmoj:
There is a console error in this [demo](https://codepen.io/MojtabaSamimi/pen/eYNyBZV?editors=0010).

archmoj:
@etpinard my branch in at https://github.com/plotly/plotly.js/tree/axis-breaks-fixups_ms

src/plots/cartesian/layout_attributes.js
etpinard: Done in https://github.com/plotly/plotly.js/pull/4614/commits/7080f900074a59c6af3fd664999d8493e9d2fe84

etpinard:
> There is a console error in this [demo](https://codepen.io/MojtabaSamimi/pen/eYNyBZV?editors=0010).

Looks like this one is just an example of a more general bug. Commenting out the `axis.breaks` and replacing the weekends dates with `NaN`s gives the same console error: https://codepen.io/etpinard/pen/dyoJxKe?editors=0010

@archmoj you've worked on this routine in `bar/plot.js` recently, do you have an idea why this happens?

archmoj:
> > There is a console error in this [demo](https://codepen.io/MojtabaSamimi/pen/eYNyBZV?editors=0010).
> 
> Looks like this one is just an example of a more general bug. Commenting out the `axis.breaks` and replacing the weekends dates with `NaN`s gives the same console error: https://codepen.io/etpinard/pen/dyoJxKe?editors=0010
> 
> @archmoj you've worked on this routine in `bar/plot.js` recently, do you have an idea why this happens?

Something went wrong in [v1.50.0](https://github.com/plotly/plotly.js/releases/tag/v1.50.0) | [demo-console-error](https://codepen.io/MojtabaSamimi/pen/abOEerE?editors=1000).
Possibly a side effect of `Add transition support to bar trace length, width, on-graph text positioning,
marker style and error bars` [#4180, #4186]

archmoj:
It could be related to 

Screenshot from 2020-03-10 15-18-07

in https://github.com/plotly/plotly.js/pull/4180/commits/f49828a05fdf593097083842b31cbc2f8f777c11

archmoj:
> TODO:
> 
> * [ ]  make sure axis breaks work on axes with `range[0] > range[1]`

On a reversed x-axis the hover position sometimes is in the wrong position: [demo](https://codepen.io/MojtabaSamimi/pen/JjdpPNm?editors=1000).

archmoj:
On axes with `autorange: 'reversed'` of this [demo](https://codepen.io/MojtabaSamimi/pen/mdJXbQe?editors=0010) autorange does not fit.

archmoj:
It appears (at the moment) `cartesian` `breaks` are not compatible with `gl2d` traces namely `scattergl`: [demo](https://codepen.io/MojtabaSamimi/pen/OJVQJXg?editors=0010). 
Should we warn the user in that case?

etpinard:
> Should we warn the user in that case?

Fixing the behaviour for `scattergl` should not be hard.

etpinard:
https://github.com/plotly/plotly.js/pull/4614/commits/dcceb7693ed2a68264a295b9dca22c2e2ec65224 fixes the overlapping ticks of https://github.com/plotly/plotly.js/pull/4614#issuecomment-595242555

Updated codepen: https://codepen.io/etpinard/pen/rNVJaex?editors=1010

archmoj:
> > TODO:
> > 
> > * [ ]   make sure axis breaks work on axes with `range[0] > range[1]`
> 
> On a reversed x-axis the hover position sometimes is in the wrong position: [demo](https://codepen.io/MojtabaSamimi/pen/JjdpPNm?editors=1000).

@etpinard please pick https://github.com/plotly/plotly.js/commit/23c566f98b4ab92b463b726eb3132dea279b26f4 to fix hover on reversed ranges.

etpinard:
> @etpinard please pick [23c566f](https://github.com/plotly/plotly.js/commit/23c566f98b4ab92b463b726eb3132dea279b26f4) to fix hover on reversed ranges.

Thanks!! I know I should be doing this, but I'm running out of time: can you try adding a test to lock this down similar to 

https://github.com/plotly/plotly.js/blob/dcceb7693ed2a68264a295b9dca22c2e2ec65224/test/jasmine/tests/hover_label_test.js#L2707-L2773

etpinard:
> > Should we warn the user in that case?
> 
> Fixing the behaviour for `scattergl` should not be hard.

I was wrong. This will actually be very tricky to get right for `scattergl` - most likely `regl-scatter2d`, `regl-line2d` and `regl-error2d` will have to be patched. I doubt I'll have time to do this before Friday.

archmoj:
> > > Should we warn the user in that case?
> > 
> > 
> > Fixing the behaviour for `scattergl` should not be hard.
> 
> I was wrong. This will actually be very tricky to get right for `scattergl` - most likely `regl-scatter2d`, `regl-line2d` and `regl-error2d` will have to be patched. I doubt I'll have time to do this before Friday.

Thanks for the investigation. Could we `visible: false` scattergl on date axes with breaks?

alexcjohnson:
> > Should we warn the user in that case?
> Could we `visible: false` scattergl on date axes with breaks?

Yes, let's leave gl out of this PR. Hide them - and log a warning - in that case. But @archmoj please make an issue to come back to this soon, I can see this being important for some users.

etpinard:
> * better "first tick" algorithm e.g. what to do when the canonical first auto tick falls into a break?

Commit https://github.com/plotly/plotly.js/pull/4614/commits/28c328dfd3f9f31c83b727b610f4329f2b0c6149 made things a bit better. Still room for improvement though.

archmoj:
> > > Should we warn the user in that case?
> > > Could we `visible: false` scattergl on date axes with breaks?
> 
> Yes, let's leave gl out of this PR. Hide them - and log a warning - in that case. But @archmoj please make an issue to come back to this soon, I can see this being important for some users.

Good call. Tracked in #4630.

archmoj:
> It could be related to
> 
> Screenshot from 2020-03-10 15-18-07
> 
> in [f49828a](https://github.com/plotly/plotly.js/commit/f49828a05fdf593097083842b31cbc2f8f777c11)

@etpinard 
Now that `NaN` positions are handled in #4634, 
would you mind merging master to this branch?

archmoj:
Console error when zooming out (+10 times) on this [demo](https://codepen.io/MojtabaSamimi/pen/WNvzQdd?editors=1010).

archmoj:
Another console error when zooming in (+2 times) on this [demo](https://codepen.io/MojtabaSamimi/pen/WNvzQdd?editors=1010).

etpinard:
Nice catch. It's the same problem in both demos. The modebar zoom button calls a relayout with a range falling inside a break. I'm not exactly sure what to do here.

archmoj:
Also to note by panning left and right (n times) the x-axis scale would change (which I assume it shouldn't) until the line would be drawn vertically where get an error.

etpinard:
> Also to note by panning left and right (n times) the x-axis scale would change (which I assume it shouldn't) until the line would be drawn vertically where get an error.

Hmm. That shouldn't happen. The logic here must be off:

https://github.com/plotly/plotly.js/blob/28c328dfd3f9f31c83b727b610f4329f2b0c6149/src/plots/cartesian/dragbox.js#L1017-L1020

archmoj:
Thanks @etpinard the following can help:
```js
var p0 = 0;
var p1 = axi._length;
var d0 = axi.p2l(p0 + pix) - axi.p2l(p0);
var d1 = axi.p2l(p1 + pix) - axi.p2l(p1);
var delta = (d0 + d1) / 2;

axi.range = [
    axi.l2r(axi._rl[0] - delta),
    axi.l2r(axi._rl[1] - delta)
];
```

etpinard:
> > * better "first tick" algorithm e.g. what to do when the canonical first auto tick falls into a break?
> 
> Commit [28c328d](https://github.com/plotly/plotly.js/commit/28c328dfd3f9f31c83b727b610f4329f2b0c6149) made things a bit better. Still room for improvement though.

Improved once more in https://github.com/plotly/plotly.js/pull/4614/commits/1b55b42f5ce3c6986774c3866d70630a72d6721a

-------

Pan fixes from @archmoj are in https://github.com/plotly/plotly.js/pull/4614/commits/40d57fa6c4e29a5fb7d10cf528ef9975f08792f5

--------



> > > Should we warn the user in that case?
> > > Could we `visible: false` scattergl on date axes with breaks?
> 
> Yes, let's leave gl out of this PR. Hide them - and log a warning - in that case. But @archmoj please make an issue to come back to this soon, I can see this being important for some users.

Done in https://github.com/plotly/plotly.js/pull/4614/commits/49b40534107c1d5cf91b9765e38e2a79064e776c

archmoj:
Great PR! 
:medal_military: :medal_military: :1st_place_medal: 

Thanks very much @etpinard for tackling all this.
:dancer: :man_dancing:

etpinard:
Ok, merging to get a first round of QA rolling (I suspect a few bug reports will come out of it).

Update examples:

breaks using pattern on date axis: https://codepen.io/etpinard/pen/PoqRNoL
breaks using values on date axis: https://codepen.io/etpinard/pen/OJVvNJG

nicolaskruchten:
🏆 thanks @etpinard 🙇

kenahoo:
Is this functionality part of any release?  Or is it still experimental?  I'm having a little trouble finding how to do it in the docs.

archmoj:
> Is this functionality part of any release? Or is it still experimental? I'm having a little trouble finding how to do it in the docs.

@kenahoo yes it is available since `v1.53.0`
You could find javascript documentation at https://plotly.com/javascript/reference/layout/xaxis/#layout-xaxis-rangebreaks.
The documentation for Python at the bottom of this page https://plotly.com/python/time-series/ could also be of interest.

PR #4618, opened 2020-03-06
Corrected typo in histogram hover

emmanuelle:
I think there was a typo here, but I may be missing something.

### DIFF
--- a/src/traces/histogram/hover.js
+++ b/src/traces/histogram/hover.js
@@ -27,7 +27,7 @@ module.exports = function hoverPoints(pointData, xval, yval, hovermode) {
         pointData[posLetter + 'Label'] = hoverLabelText(pointData[posLetter + 'a'], di.ph0, di.ph1);
     }
 
-    if(trace.hovermplate) pointData.hovertemplate = trace.hovertemplate;
+    if(trace.hovertemplate) pointData.hovertemplate = trace.hovertemplate;
 
     return pts;
 };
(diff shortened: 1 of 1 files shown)


archmoj:
Good catch.
@emmanuelle so because of this, `hovertemplate` was not working on `histogram`?

emmanuelle:
I'm very confused because `hovertemplate` does work as I just checked (screenshot from Python sorry but you'll get the idea)
image

emmanuelle:
of maybe the `if` was always True?

archmoj:
Most likely this should be labled as `bug`.
Anyway it would be good to know which issue we are closing.
@antoinerg any idea?

antoinerg:
Thank you @emmanuelle for spotting this ridiculous line of code of mine. Actually, it is completely unnecessary and it should be removed. All tests pass without it: https://github.com/plotly/plotly.js/commit/d88d2438c1fe2fae95a804d48beaf755bc96b073

Please remove it and I'll approve the change :smile_cat:

archmoj:
Please wait `Plotly.PlotSchema.get().traces.histogram.attributes.hovertemplate` gives me 
```js
{valType: "string", dflt: "", editType: "none", arrayOk: true}
```
So in the case of blank string (i.e. the plotly.js default) `hovertemplate` is disabled.

antoinerg:
@archmoj I'm not sure I understand what you're getting at? Are you saying that https://github.com/plotly/plotly.js/commit/d88d2438c1fe2fae95a804d48beaf755bc96b073 is not OK?

archmoj:
That commit is good; and we likely need to keep the if statement. 
But wondering what should be the result in the case of a blank item in the array vs the blank default? Here is a [demo](https://codepen.io/MojtabaSamimi/pen/VwLyrjj?editors=0010).

archmoj:
TODO: 
- [ ] remove if statement as proposed in https://github.com/plotly/plotly.js/pull/4618#issuecomment-596792452

archmoj:
ping @emmanuelle

archmoj:
:dancer:

Issue #3841, opened 2019-05-07
support MathJax in hover labels


antoinerg:
As of right now, `hover(template|text)` does not properly parse/render LaTeX symbols

gvwilson:
Hi - we are currently trying to tidy up Plotly's public repositories to help us focus our efforts on things that will help users most. Since this issue has been sitting for several years, I'm going to close it. If it's still a concern, we'd be grateful if you could open a new issue (with a short reproducible example if appropriate) so that we can add it to our backlog. Thanks for your help - @gvwilson

PR #4620, opened 2020-03-09
unified hover label

### LINKED ISSUE #4477: unified hoverlabels
We need a new `hovermode` like `xunified`/`yunified` that would produce a single hoverlabel that would contain info about all the points at a given `x`/`y` ... similar to `hovermode=x` but with a single hover. Like this:

image

Other libraries which implement this include eCharts (ignore the "follow-me" aspect here, it's not in-scope for this issue and it's discussed here #3899 ) https://www.echartsjs.com/examples/en/editor.html?c=line-stack and https://www.fusioncharts.com/ (scroll down a bit)

antoinerg:
Closes #4477 by introducing hover modes `x unified` and `y unified`!

Codepen:
- https://codepen.io/antoinerg/pen/MWwpPpa
- https://codepen.io/antoinerg/pen/bGdYejZ 

TODO:
- [x] test traces with MathJax in their names (not working but see https://github.com/plotly/plotly.js/pull/4620#issuecomment-597233916)
- [x] style symbol based on on `(marker | line).color` (ec8c373ecdd3b98f4478e6afa52151418dc0fd69)
- [x] investigate problem with mock `legend_horizontal_autowrap` (fixed in https://github.com/plotly/plotly.js/pull/4620/commits/ae5b9ccfd0fc3a0f45467b4aa721f1b5e06def4b)
- [x] fix waterfall symbols


archmoj:
Great PR!
@antoinerg could we possibly use color of each point instead of the legend?
Here is a [demo](https://codepen.io/MojtabaSamimi/pen/JjdMeBX?editors=1000), where icons for traces with `(marker | line).color` arrays and colorscale currently paint in black.

src/components/fx/hover.js
```diff
@@ -543,7 +548,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         var thisSpikeDistance;
         for(var i = 0; i < pointsData.length; i++) {
             thisSpikeDistance = pointsData[i].spikeDistance;
-            if(thisSpikeDistance < minDistance && thisSpikeDistance <= spikedistance) {
+            if(thisSpikeDistance <= minDistance && thisSpikeDistance <= spikedistance) {
```
archmoj: @antoinerg Could you please explain why this change was required?

src/components/legend/defaults.js
```diff
@@ -17,7 +17,7 @@ var basePlotLayoutAttributes = require('../../plots/layout_attributes');
 var helpers = require('./helpers');
 
 
-module.exports = function legendDefaults(layoutIn, layoutOut, fullData) {
+module.exports = function legendDefaults(layoutIn, layoutOut, fullData, fullLayout) {
```
archmoj: Let's make the 4th argument an option
```js
module.exports = function legendDefaults(layoutIn, layoutOut, fullData, opts) {
  if(!opts) opts = {};
```
and then pass desired options like this:
```js
legend.supplyLayoutDefaults(mockLayoutIn, mockLayoutOut, gd._fullData, {
  paper_bgcolor: fullLayout.paper_bgcolor,
  font: fullLayout.font
});
```

src/components/legend/draw.js
```diff
@@ -30,26 +30,44 @@ var getLegendData = require('./get_legend_data');
...
 var helpers = require('./helpers');
 
-module.exports = function draw(gd) {
+module.exports = function draw(gd, opts) {
     var fullLayout = gd._fullLayout;
     var clipId = 'legend' + fullLayout._uid;
+    var layer;
 
-    if(!fullLayout._infolayer || !gd.calcdata) return;
+    // Check whether this is the main legend (ie. called without any opts)
```
archmoj: This is little hacky.
What if we pass proper options to draw legend.
In that case one may look into `opts._main` to figure out the mocked legend (for unified hover) from the actual legend.

src/components/fx/hover.js
```diff
@@ -906,11 +912,96 @@ function createHoverText(hoverData, opts, gd) {
...
+            legend: {
+                title: {text: t0},
+                bgcolor: fullLayout.paper_bgcolor,
+                borderwidth: 1,
+                tracegroupgap: 7,
+                orientation: 'v'
+            }
+        };
+        var mockLayoutOut = {};
+        legend.supplyLayoutDefaults(mockLayoutIn, mockLayoutOut, gd._fullData, fullLayout);
```
etpinard: Could you mock `paper_bgcolor` into `mockLayoutOut` instead of passing `fullLayout` to `legend.supplyLayoutDefauls` ?

src/components/fx/hover.js
```diff
@@ -25,6 +25,8 @@ var Registry = require('../../registry');
 var helpers = require('./helpers');
 var constants = require('./constants');
 
+var legend = require('../legend');
```
etpinard: We usually use `Legend` (not `legend`) when requiring the legend module. 

Personally, I prefer require just the methods we need. I think here only:

```js
var legendSupplyDefaults = require('../legend/defaults');
var legendDraw = require('./legend/draw');
```

would suffice.

etpinard:
Very cool! 

@antoinerg have you tested `x unified` and `y unified` on graphs that have traces with MathJax symbol in their `name` attributes? I suspect this won't work smoothly as `Fx.hover` does not handle MathJax promises.

We may not need to handle unified hover labels + MathJax for the first iteration, but it would be nice to confirm

test/image/mocks/hovermode_xunified.json
```diff
@@ -0,0 +1,45 @@
+{
```
archmoj: You may drop this mock and baseline, if only needed for jasmine test.

src/plots/cartesian/layout_defaults.js
```diff
@@ -249,12 +249,14 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
-        var spikemode = coerce2('spikemode');
+        var unifiedHover = layoutIn.hovermode && ['x unified', 'y unified'].indexOf(layoutIn.hovermode) !== -1;
+        var unifiedSpike = unifiedHover && axLetter === layoutIn.hovermode.charAt(0);
+        var spikecolor = coerce2('spikecolor', unifiedHover ? axLayoutOut.color : undefined);
+        var spikethickness = coerce2('spikethickness', unifiedHover ? 1.5 : undefined);
+        var spikedash = coerce2('spikedash', unifiedHover ? 'dot' : undefined);
+        var spikemode = coerce2('spikemode', unifiedHover ? 'across' : undefined);
         var spikesnap = coerce2('spikesnap');
-        var showSpikes = coerce('showspikes', !!spikecolor || !!spikethickness || !!spikedash || !!spikemode || !!spikesnap);
+        var showSpikes = coerce('showspikes', unifiedSpike || !!spikecolor || !!spikethickness || !!spikedash || !!spikemode || !!spikesnap);
```
archmoj: Please use `!!unifiedSpike` here to ensure casting to boolean.

src/components/fx/hover.js
```diff
@@ -950,8 +950,13 @@ function createHoverText(hoverData, opts, gd) {
             var texts = getHoverLabelText(hoverData[j], true, hovermode, fullLayout, t0);
             var text = texts[0];
             var name = texts[1];
-            hoverData[j].text = name + ' : ' + text;
             hoverData[j].name = name;
+            if(name) {
```
archmoj: What happens if `name === 0`?

archmoj:
Hover is not showing up on this [mock](https://codepen.io/MojtabaSamimi/pen/qBdpgov?editors=1000) | possibly related to not having enough space?

src/components/legend/defaults.js
antoinerg: Following @etpinard's https://github.com/plotly/plotly.js/pull/4620#discussion_r390317487 and yours, I realized that I can simply set the `font`, `papger_bgcolor` and `title.font` in the `mockLayoutIn` and not have to pass the `fullLayout` object to `legend.supplyLayoutDefaults`. `fullLayout` is only needed to figure out the default values in case they are missing.

src/components/fx/hover.js
antoinerg: Done in 9ea8bfdab29495a2eaf96884a36dff71a38aff67

src/components/fx/hover.js
antoinerg: As explained in https://github.com/plotly/plotly.js/pull/4620#discussion_r390488261, I can simply set those in `mockLayoutIn`. Changes made in 9ea8bfdab29495a2eaf96884a36dff71a38aff67

src/components/legend/defaults.js
antoinerg: Done in https://github.com/plotly/plotly.js/commit/9ea8bfdab29495a2eaf96884a36dff71a38aff67

archmoj: Resolved in https://github.com/plotly/plotly.js/pull/4620/commits/9ea8bfdab29495a2eaf96884a36dff71a38aff67.

src/components/fx/hover.js
archmoj: Resolving. Done in https://github.com/plotly/plotly.js/pull/4620/commits/9ea8bfdab29495a2eaf96884a36dff71a38aff67.

src/components/fx/hover.js
archmoj: Resolving. Done in https://github.com/plotly/plotly.js/pull/4620/commits/9ea8bfdab29495a2eaf96884a36dff71a38aff6.

src/plots/cartesian/layout_defaults.js
antoinerg: Done in 36bc6b5dc2f27ed3c2b90066c0b1afc4199c692e

src/components/fx/hover.js
antoinerg: Fixed in 36bc6b5dc2f27ed3c2b90066c0b1afc4199c692e by checking for empty string instead of truthy! Thanks @archmoj

src/plots/cartesian/layout_defaults.js
archmoj: !!Thanks!! @antoinerg 
Resolving...

src/components/fx/hover.js
archmoj: Resolved.

test/image/mocks/hovermode_xunified.json
antoinerg: Done in e82e010da369a347d8d13db18290b34ca710db79

src/components/fx/hover.js
antoinerg: I can't remember why so I reverted the change in 818223d6c245f4ebf091ff9ac6890ed31e4039bc ! Thanks @archmoj :male_detective:

archmoj: @antoinerg thanks for the revision.
Resolved.

antoinerg:
> @antoinerg have you tested `x unified` and `y unified` on graphs that have traces with MathJax symbols in their `name` attributes? I suspect this won't work smoothly as `Fx.hover` does not handle MathJax promises.
> 
> We may not need to handle unified hover labels + MathJax for the first iteration, but it would be nice to confirm

@etpinard I just tested it and you were right that `Fx.hover` does not handle MathJax properly. At the moment, if the trace name has MathJax in it, it will appear as a raw string in the hover label. However, this is on par with the other hover modes so I consider it out of scope for this PR. However, I'd be curious to know how much work you think it would be to add this feature.

etpinard:
> At the moment, if the trace name has MathJax in it, it will appear as a raw string in the hover label.

That's good. 

> However, this is on par with the other hover modes so I consider it out of scope for this PR

I agree 100% cc https://github.com/plotly/plotly.js/issues/3841

antoinerg:
> @antoinerg could we possibly use color of each point instead of the legend?
> Here is a [demo](https://codepen.io/MojtabaSamimi/pen/JjdMeBX?editors=1000), where icons for traces with `(marker | line).color` arrays and colorscale currently paint in black.

This is addressed in https://github.com/plotly/plotly.js/pull/4620/commits/ec8c373ecdd3b98f4478e6afa52151418dc0fd69

archmoj:
> > @antoinerg could we possibly use color of each point instead of the legend?
> > Here is a [demo](https://codepen.io/MojtabaSamimi/pen/JjdMeBX?editors=1000), where icons for traces with `(marker | line).color` arrays and colorscale currently paint in black.
> 
> This is addressed in [ec8c373](https://github.com/plotly/plotly.js/commit/ec8c373ecdd3b98f4478e6afa52151418dc0fd69)

Wow! Hover on those mocks look very nice. Here is the [updated demo](https://codepen.io/MojtabaSamimi/pen/xxGYbGR?editors=1000).

archmoj:
For hovering on `waterfall` we may want to do something similar to hovering on other `bar`-like traces i.e. not to use waterfall legend. 
Here is a [demo](https://codepen.io/MojtabaSamimi/pen/ExjQaPy?editors=1000) using current branch.

antoinerg:
> For hovering on `waterfall` we may want to do something similar to hovering on other `bar`-like traces i.e. not to use waterfall legend.

Hopefully this is the last trace type to fix. Basically, I need to make sure the following function gets everything it needs to properly style things:
https://github.com/plotly/plotly.js/blob/8f049fddbac0ca0382816984b8526857e9714fe6/src/components/legend/style.js#L288-L319

I added it to the TODO at the top.

archmoj:
One more question:
Could we reverse the order when they are stacked on the y-axis? Here is a [demo](https://codepen.io/MojtabaSamimi/pen/RwPQNZJ?editors=1000) using current behaviour.

alexcjohnson:
> Could we reverse the order when they are stacked on the x-axis?

Good catch @archmoj - I think it's probably a good idea to keep the order consistent from point to point, but perhaps we should honor `legend.traceorder` in unified hover labels?

antoinerg:
> For hovering on `waterfall` we may want to do something similar to hovering on other `bar`-like traces i.e. not to use waterfall legend.

Commit ba3de2f88574d719d5fef1553be1c257d243d3bc improves things slightly for `waterfall` but I'm not sure what should be displayed. I'm confused about the style of legend items to begin with. In `waterfall_months`, all legend items have 3 colors:
Screenshot_2020-03-11_00-11-35

But if only display one trace, 3 legend items now have 2 colors
Screenshot_2020-03-11_00-11-48

@archmoj is that expected?

antoinerg:
> Good catch @archmoj - I think it's probably a good idea to keep the order consistent from point to point, but perhaps we should honor legend.traceorder in unified hover labels?

With 0822b93bb08b3009f0495bf8de5f9d56f031305f, we now use `legend.traceorder` by default, if it's defined. @archmoj I think that's the behavior you were hoping for :smile_cat:

archmoj:
> > Good catch @archmoj - I think it's probably a good idea to keep the order consistent from point to point, but perhaps we should honor legend.traceorder in unified hover labels?
> 
> With [0822b93](https://github.com/plotly/plotly.js/commit/0822b93bb08b3009f0495bf8de5f9d56f031305f), we now use `legend.traceorder` by default, if it's defined. @archmoj I think that's the behavior you were hoping to get

@antoinerg  Thanks very much. By considering `traceorder` the order of items in hover is now improved.
In addition to that, I still think that it may be a good idea to flip the order on a reversed y-axis:
Here are two examples:
[funnel demo](https://codepen.io/MojtabaSamimi/pen/yLNvbzX?editors=1000)
[bar demo](https://codepen.io/MojtabaSamimi/pen/qBdxmay?editors=1000)

archmoj:
> > For hovering on `waterfall` we may want to do something similar to hovering on other `bar`-like traces i.e. not to use waterfall legend.
> 
> Commit [ba3de2f](https://github.com/plotly/plotly.js/commit/ba3de2f88574d719d5fef1553be1c257d243d3bc) improves things slightly for `waterfall` but I'm not sure what should be displayed. I'm confused about the style of legend items to begin with. In `waterfall_months`, all legend items have 3 colors:
> Screenshot_2020-03-11_00-11-35
> 
> But if only display one trace, 3 legend items now have 2 colors
> Screenshot_2020-03-11_00-11-48
> 
> @archmoj is that expected?

When a `waterfall` trace is `visible: false` we don't look into the data to see if it has `total`s or not.
So the legend of invisible `waterfall`s is OK.
But concerning `watrefall` icons on unified hovers, IMO we shouldn't display triangles at all. Instead it can be a simple square identical to other bar-like traces.

antoinerg:
> In addition to that, I still think that it may be a good idea to flip the order on a reversed y-axis:
> Here are two examples:
> [funnel demo](https://codepen.io/MojtabaSamimi/pen/yLNvbzX?editors=1000)
> [bar demo](https://codepen.io/MojtabaSamimi/pen/qBdxmay?editors=1000)

As it stands, `waterfall_custom` has the same `traceorder` in the legend and the hover label. Do you also think it might be a good idea to flip the order of legend items when on a reversed y-axis? Why change one and not the other? The same could be said about `bar_stackto1`.

I'm not sure how to proceed cc @nicolaskruchten @alexcjohnson cc @archmoj

alexcjohnson:
Let's leave any sort of extra axis flip logic out of this PR. If we want to add logic to set `legend.traceorder` explicitly based on reversing the y axis, we could consider that in a separate issue (@archmoj feel free to open an issue to discuss) - and then the hover label order would follow from that. Both of the demos pointed to by @archmoj would benefit from flipping the legend as well as flipping the hover labels.

antoinerg:
> But concerning `watrefall` icons on unified hovers, IMO we shouldn't display triangles at all. Instead it can be a simple square identical to other bar-like traces.

@archmoj I added this in commit ae5b9ccfd0fc3a0f45467b4aa721f1b5e06def4b . We now get:
Screenshot_2020-03-11_12-21-33
which is a really good improvement IMO.

src/components/fx/hover.js
antoinerg: I think this equal sign was needed after all. 

When setting `spikedistance` to `-1` as we do for unified hover label, `thisSpikeDistance` is `Infinity` and `minDistance` is also `Infinity`. Without an equal sign, the condition is never met and we don't get spikelines. This can be seen in the first above Codepen: the spikeline does not appear anymore.

antoinerg: This looks like a spike line bug.

archmoj:
@antoinerg any idea why we get error over `candlestick` mocks? Here is a [demo](https://codepen.io/MojtabaSamimi/pen/QWbQawY?editors=1000).

antoinerg:
> @antoinerg any idea why we get error over `candlestick` mocks? Here is a [demo](https://codepen.io/MojtabaSamimi/pen/QWbQawY?editors=1000).

Should be fixed in https://github.com/plotly/plotly.js/pull/4620/commits/bca4a708026a9bf3378561e87f8c094929256547

antoinerg:
> @antoinerg any idea why we get error over `candlestick` mocks? Here is a [demo](https://codepen.io/MojtabaSamimi/pen/QWbQawY?editors=1000).

Fixed in b3f5abbe3baada6aca325447936b6c1b92d75e69 .

antoinerg:
@archmoj I made a fixup to my latest commit and force pushed. See https://github.com/plotly/plotly.js/pull/4620/commits/8308111c5fcfba2d595a54771614f2b5beec6504

src/components/legend/style.js
```diff
@@ -288,6 +291,14 @@ module.exports = function style(s, gd) {
     function styleWaterfalls(d) {
         var trace = d[0].trace;
 
+        if(d[0]._distinct && trace.type === 'waterfall') {
```
archmoj: Now that we test for `trace.type === 'waterfall'` multiple times, let's make a variable and use it in those two places.
```js
var isWaterfall = trace.type === 'waterfall';
```

antoinerg: Done in 67b1fa69210d9b38006d4b41205fb85770e3c835

src/components/legend/draw.js
```diff
@@ -30,26 +30,44 @@ var getLegendData = require('./get_legend_data');
...
+    if(!opts) {
```
archmoj: IMO, if this is really an option, on top of the function we should be able to add
```js
if(!opts) opts = {}:
```
But we can't do that right now!
So this line should be something like
```js
if(!opts._unifiedHover) {
```
and we must pass `_unifiedHover` in unified-hover `opts` object.

antoinerg: I'm not sure I see why this
```
if(!opts) opts = {};
```
is much better than:
```
if(!opts) {
  opts = fullLayout.legend || {};
  ... some more defaults values ...
}
```
Can you show me a commit with the changes you'd like to see @archmoj ?

archmoj:
@antoinerg any idea why `spikelines` are missing on this [demo](https://codepen.io/MojtabaSamimi/pen/ZEGrrNo?editors=1000)?

antoinerg:
> @antoinerg any idea why `spikelines` are missing on this [demo](https://codepen.io/MojtabaSamimi/pen/ZEGrrNo?editors=1000)?

Probably because of https://github.com/plotly/plotly.js/issues/4626 which should be fixed by  https://github.com/plotly/plotly.js/pull/4627

src/components/fx/hover.js
```diff
@@ -906,11 +913,112 @@ function createHoverText(hoverData, opts, gd) {
...
+                if(cd.dir) pt.dir = cd.dir;
+            }
+            pt._distinct = true;
+
+            legendOpts.entries.push([pt]);
+        }
+        legendOpts.layer = container;
+
+        // Return early if nothing is hovered on
+        if(legendOpts.entries.length === 0) return;
```
archmoj: Could we exit earlier before mocking legend?
```js
if(!hoverData.length) return;
```

antoinerg: Done in a9c899b02597408eb45fa39d54f3e51c84a55722

src/components/legend/draw.js
archmoj: See https://github.com/plotly/plotly.js/commit/72abd2fb0f72344b13065319c66d0211094fb73f

src/components/fx/layout_attributes.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
     hovermode: {
         valType: 'enumerated',
         role: 'info',
-        values: ['x', 'y', 'closest', false],
+        values: ['x', 'y', 'closest', false, 'x unified', 'y unified'],
```
archmoj: @antoinerg would you please update the description of `hovermode` in respect to these two new attributes?

archmoj:
Wondering if we need to swap `x unified` and `y unified` here?
https://github.com/plotly/plotly.js/blob/8f049fddbac0ca0382816984b8526857e9714fe6/src/plot_api/plot_api.js#L1733-L1741

src/components/fx/layout_attributes.js
antoinerg: @nicolaskruchten could you tackle this one (ie. updating the description of `hovermode` to include `x unified` and `y unified`) ?

antoinerg:
> Wondering if we need to swap `x unified` and `y unified` here?
> https://github.com/plotly/plotly.js/blob/8f049fddbac0ca0382816984b8526857e9714fe6/src/plot_api/plot_api.js#L1733-L1741

@archmoj I am not sure what that piece of logic is supposed to be doing. I'll have to defer to @alexcjohnson and @etpinard !

alexcjohnson:
Oy yes restyle... in principle yes, we should include the new values in the `orientationaxes` block. I don't know if even chart studio uses that anymore, and I highly doubt anyone else does, but better safe than sorry I guess.

src/components/fx/layout_attributes.js
archmoj: ping @nicolaskruchten

nicolaskruchten: Please insert, before the existing paragraph, something like:

If *closest*, a single hoverlabel will appear for the closest point within the `hoverdistance`.
If *x* (or *y*), multiple hoverlabels will appear for multiple points at the closest x- (or y-) coordinate within the `hoverdistance`, with the caveat that no more than one hoverlabel will appear per trace.
If *x unified* (or *y unified*), a single hoverlabel will appear multiple points at the closest x- (or y-) coordinate within the `hoverdistance`, with the caveat that no more than one hoverlabel will appear per trace. In this mode, `spikelines` are enabled by default perpendicular to the specified axis.
If *false*, hover interactions are disabled.

Edit to taste and/or to correct inaccuracies :)

alexcjohnson:
Just watching @nicolaskruchten demo this and saying "I have to reload because there’s no button for unified hover" - makes me think perhaps unified vs separate hover should be a new layout attribute (`unifiedhover: true`?) rather than part of `hovermode`. That way the modebar buttons would still function to toggle compare vs closest without losing your preference of what compare mode looks like.

src/components/fx/layout_attributes.js
```diff
@@ -61,6 +61,16 @@ module.exports = {
         editType: 'modebar',
         description: [
             'Determines the mode of hover interactions.',
+            'If `closest`, a single hoverlabel will appear',
```
archmoj: Please put flags between two * and attributes between ` chars.
E.g. this line should be 
```js
'If *closest*, a single `hoverlabel` will appear',
```

antoinerg:
> Just watching @nicolaskruchten demo this and saying "I have to reload because there’s no button for unified hover" - makes me think perhaps unified vs separate hover should be a new layout attribute (`unifiedhover: true`?) rather than part of `hovermode`. That way the modebar buttons would still function to toggle compare vs closest without losing your preference of what compare mode looks like.

That's an interesting idea. Either this or we add a third button? cc @nicolaskruchten

nicolaskruchten:
We don't want to add a new attribute because there's no real meaning to unified + closest. I also don't really want to add a new button today because the mode bar is already pretty big IMO.

Maybe in unified hover mode the two existing buttons disappear by default?

alexcjohnson:
> We don't want to add a new attribute because there's no real meaning to unified + closest.

By itself that's true, but it serves the purpose of saving the fact that you when you enable compare mode you want it to have unified hover.

antoinerg:
> Maybe in unified hover mode the two existing buttons disappear by default?

This could be achieved in 2 lines of code.

antoinerg:
I kinda like having `(x|y) unified` be a `hovermode` as it currently is. What do you think of the idea of adding a third button down the road (for the next release for example)?

cc @archmoj @nicolaskruchten @alexcjohnson

alexcjohnson:
I'd be OK with, as a short-term measure, removing the hover mode buttons in one of the unified modes. Perhaps down the line this is just one more reason to put some menus into the mode bar - one for hovermode, another for drag mode...

antoinerg:
> I'd be OK with, as a short-term measure, removing the hover mode buttons in one of the unified modes.

I remove the modebar buttons when hovermode is `(x|y) unified`.

> Perhaps down the line this is just one more reason to put some menus into the mode bar - one for hovermode, another for drag mode...

I can open an issue for this one if you'd like!

src/plots/cartesian/layout_defaults.js
```diff
@@ -249,12 +249,14 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
         handleTypeDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions);
         handleAxisDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions, layoutOut);
 
-        var spikecolor = coerce2('spikecolor');
-        var spikethickness = coerce2('spikethickness');
-        var spikedash = coerce2('spikedash');
-        var spikemode = coerce2('spikemode');
+        var unifiedHover = layoutIn.hovermode && ['x unified', 'y unified'].indexOf(layoutIn.hovermode) !== -1;
```
antoinerg: Let's replace `layoutIn` by `layoutOut`

src/plots/cartesian/layout_defaults.js
```diff
@@ -249,12 +249,14 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
...
+        var unifiedSpike = unifiedHover && axLetter === layoutIn.hovermode.charAt(0);
```
antoinerg: Let's replace `layoutIn` by `layoutOut`

eddy-geek:
great work. two remarks from using it (through the python api)

* The first line containing "X" value seems to be oblivious to `hovertemplate` image
Is it worth an issue or just documentation update or am I missing something?

* With plotly express `color=` you get a ton of repetition by default. Example based on the first *Line* example from the plotly.py doc:
image
```python
import plotly.express as px

df = px.data.gapminder().query("continent=='Oceania'")
fig = px.line(df, x="year", y="lifeExp", color='country').update_layout(hovermode='x unified')
fig.show()
```
I don't see an easy fix, but better documentation on the correct `hover*` arguments to use would help.

nicolaskruchten:
Indeed, the `x` value here is not subject to `hovertemplate` ... we may add a `layout.xaxis.hovertemplate` for this purpose (TBD).

Regarding Plotly Express, yes, in the docs here https://plotly.com/python/hover-text-and-formatting/ we recommend forcing `hovertemplate=None` when using compare or unified hover with PX.

nicolaskruchten:
Here is the result:

image

jcunhafonte:
@antoinerg This looks awesome. Thank you.

Screenshot 2020-04-21 at 10 27 49
When there are many traces the information gets hidden.
Any hint or suggestion to have the legend with scroll or above the graph?

alexcjohnson:
Thanks @jcunhafonte - moved to a new issue https://github.com/plotly/plotly.js/issues/4786

Issue #4243, opened 2019-10-02
[Request] Allow `customdata` attribute to be accessed by `hovertemplate` using `sankey`,


matthewchan15:
I am/was giving a workshop to an enterprise customer where they want the ability/flexibility to be able to pass in multiple sets of external metadata to `customdata`, that can be than used by `hovertemplate` for the `sankey` graph.

We tried using `customdata` from the top level `data` property, but were not able to access it. Currently `hovertemplate` can only access whatever is within it's scope, e.g `node` or `link`. And, it can only take in one data set (`list` data structure). This could be helpful for other graph types as well.

CC: @alb-erandelcastillo

etpinard:
Right, the problem here is: should we assume `customdata` to have length of `node.label` or of `link.value` / `link.source`? In brief, `customdata` for sankey traces does not match 1:1 with the other data arrays. 

So perhaps, we could expose two `customdata` attributes: one inside the `node` container and one inside the `link` container - similar to how we currently expose `node.hovertemplate` and `link.hovertemplate`.

---

This ticket is essentially equivalent to https://github.com/plotly/plotly.js/issues/2504

antoinerg:
> So perhaps, we could expose two `customdata` attributes: one inside the `node` container and one inside the `link` container - similar to how we currently expose `node.hovertemplate` and `link.hovertemplate`.

This proposal makes a whole lot of sense. :+1:

emmanuelle:
Can you confirm that it is not possible to iterate through the `customdata` from a `sankey.node`? I'm trying to do this with @matthewchan15, for example in https://codepen.io/emmanuelle-plotly/pen/PoqOWPE?editors=1010 if I put `fullData.customdata` in the hovertemplate I get the whole array, and if I put `fullData.customdata[0]` I only get the first element but for all nodes. Is there a workaround like a different shape of the customdata or a different way to write the hovertemplate?

antoinerg:
Unfortunately, at the moment, it is not possible to iterate through the `customdata` from a `sankey.node`. We would need to do what is described in https://github.com/plotly/plotly.js/issues/4243#issuecomment-537574876 . It shouldn't be too complicated to do. Is this needed urgently?

emmanuelle:
Thank you @antoinerg. I don't know how urgent this is for @matthewchan15, but there have been several requests on the forum in order to use `customdata` for hover in a `sankey` trace (which is not possible at the moment).

matthewchan15:
[...]

PR #4621, opened 2020-03-10
sankey: introducing attributes (node|link).customdata

antoinerg:
Closes https://github.com/plotly/plotly.js/issues/4243 by introducing attributes `node.customdata` and `link.customdata` as suggested in https://github.com/plotly/plotly.js/issues/4243#issuecomment-537563628.

### DIFF
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -207,6 +207,13 @@ var attrs = module.exports = overrideAll({
                 'If `link.color` is omitted, then by default, a translucent grey link will be used.'
             ].join(' ')
         },
+        customdata: {
+            valType: 'data_array',
+            editType: 'calc',
+            description: [
+                'Assigns extra data each link.'
+            ].join(' ')
+        },
         line: {
             color: {
                 valType: 'color',
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -84,6 +84,9 @@ var attrs = module.exports = overrideAll({
         description: 'Sets the font for node labels'
     }),
 
+    // Remove top-level customdata
+    customdata: undefined,
+
     node: {
         label: {
             valType: 'data_array',
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -135,7 +135,7 @@ var attrs = module.exports = overrideAll({
             valType: 'data_array',
             editType: 'calc',
             description: [
-                'Assigns extra data each node.'
+                'Assigns extra data to each node.'
             ].join(' ')
         },
         line: {
@@ -214,7 +214,7 @@ var attrs = module.exports = overrideAll({
             valType: 'data_array',
             editType: 'calc',
             description: [
-                'Assigns extra data each link.'
+                'Assigns extra data to each link.'
             ].join(' ')
         },
         line: {
--- a/src/traces/sankey/attributes.js
+++ b/src/traces/sankey/attributes.js
@@ -128,6 +128,13 @@ var attrs = module.exports = overrideAll({
                 'what is beneath the node.'
             ].join(' ')
         },
+        customdata: {
+            valType: 'data_array',
+            editType: 'calc',
+            description: [
+                'Assigns extra data each node.'
+            ].join(' ')
+        },
         line: {
             color: {
                 valType: 'color',
--- a/src/traces/sankey/calc.js
+++ b/src/traces/sankey/calc.js
@@ -116,6 +116,7 @@ function convertToD3Sankey(trace) {
     // Process nodes
     var totalCount = nodeCount + groups.length;
     var hasNodeColorArray = isArrayOrTypedArray(nodeSpec.color);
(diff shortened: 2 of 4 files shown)


etpinard:
Nice!

I see that `customdata` is currently part of the `sankey` root-level attributes:

image

Maybe we should remove it, like we currently do for `parcats`:

https://github.com/plotly/plotly.js/blob/80262b5194d625d750260b1dc64cb952cd1fb1cd/src/traces/parcats/attributes.js#L217-L218

emmanuelle:
Thank you @antoinerg for the great reactivity :-). Would it make sense to add a test where the node or link customdata has more than one dimensions, to check that it's possible to display several values from the customdata in the hovertemplate?

antoinerg:
> Maybe we should remove it, like we currently do for `parcats`

Done in https://github.com/plotly/plotly.js/pull/4621/commits/63ef3e9fb7f42d82493b0b965c23b1aebdcb5d2a

antoinerg:
> Thank you @antoinerg for the great reactivity :-). Would it make sense to add a test where the node or link customdata has more than one dimensions, to check that it's possible to display several values from the customdata in the hovertemplate?

@emmanuelle what do you mean by more than one dimension? Right now, I made it so that if `customdata` is an array, accessing `customdata` from an hovertemplate will return the value at this point's index (ie. `customdata[pointNumber]`). Is that what you're after?

emmanuelle:
@antoinerg yes. I just wanted to be sure that if `customdata = [['a', 'b'], ['c', 'd']]`, it is possible to access either `a` with `customdata[0]` in the template, or `b` with  `customdata[1]`, etc.

antoinerg:
> @antoinerg yes. I just wanted to be sure that if `customdata = [['a', 'b'], ['c', 'd']]`, it is possible to access either `a` with `customdata[0]` in the template, or `b` with `customdata[1]`, etc.

Thanks for the clarification! I'm not sure this is handled properly. I'll add a test :)

antoinerg:
@emmanuelle tested to work properly in https://github.com/plotly/plotly.js/pull/4621/commits/c43873a6a6eacef71764861f937a0810bb06f077

src/traces/sankey/attributes.js
```diff
@@ -128,6 +131,13 @@ var attrs = module.exports = overrideAll({
                 'what is beneath the node.'
             ].join(' ')
         },
+        customdata: {
+            valType: 'data_array',
+            editType: 'calc',
+            description: [
+                'Assigns extra data each node.'
```
alexcjohnson:
```suggestion
                'Assigns extra data to each node.'
```

src/traces/sankey/attributes.js
```diff
@@ -200,6 +210,13 @@ var attrs = module.exports = overrideAll({
                 'If `link.color` is omitted, then by default, a translucent grey link will be used.'
             ].join(' ')
         },
+        customdata: {
+            valType: 'data_array',
+            editType: 'calc',
+            description: [
+                'Assigns extra data each link.'
```
alexcjohnson:
```suggestion
                'Assigns extra data to each link.'
```

jzamalloa1:
Thank you very much all for working on this much needed feature! Do you know if this (node|customdata, link|customdata) will be pushed to the plotly sankey python library? That would be extremely useful, thanks!

archmoj:
I think so.

alexcjohnson:
@jzamalloa1 you mean the sankey trace type in the `plotly` package? Yes, the next release of the package - due out within the next week I would expect - will include this feature.

xanderwallace85:
> @jzamalloa1 you mean the sankey trace type in the `plotly` package? Yes, the next release of the package - due out within the next week I would expect - will include this feature.

Great! Same for the python library?

antoinerg:
> Great! Same for the python library?

Yes, the Python library will also get the improvements very shortly.

xanderwallace85:
> > Great! Same for the python library?
> 
> Yes, the Python library will also get the improvements very shortly.

Any news about the updated Python library? :(

PGrothaus:
It's in the new python release 4.6.0. Thanks a lot everyone involved, it's immensely useful!

xanderwallace85:
> It's in the new python release 4.6.0. Thanks a lot everyone involved, it's immensely useful!

Hi @PGrothaus,
Can it be that this same problem with customdata is still present in the sunburst?

xanderwallace85:
> It's in the new python release 4.6.0. Thanks a lot everyone involved, it's immensely useful!

Hi @PGrothaus,
I believe the problem is not completely solved in plotly. For instance, regarding the sunburst, the case to follow doesn't work properly:
```
customdata=np.dstack((a, b)),
hovertemplate='<br>%{customdata[0]:.3f} <br> %{customdata[1]:.3f} ',
```

However, if I have simply 
```
customdata=a,
hovertemplate='<br>%{customdata:.3f} <br>',
```
then everything works smoothly.

antoinerg:
@xanderwallace85 I think it would be better if you opened a separate issue for the sunburst case you're describing!

Thank you for using plotly.js :)

Issue #3805, opened 2019-04-24
layout.spikedistance = -1 causes missing spike lines over non-scatter traces.


pope1ni:
When using [`layout.spikedistance`](https://plot.ly/python/reference/#layout-spikedistance) with a value of `-1`, which is supposed to mean "no cutoff", and [`layout.xaxis.spikesnap`](https://plot.ly/python/reference/#layout-xaxis-spikesnap) with a value of `'cursor'`, I expected the spike line to be always visible in the plot area when moving the cursor across it. Indeed, when looking at my [codepen](https://codepen.io/anon/pen/eoPXJE), this is the case for the first scatter chart.

When I have other types of traces, the `-1` value doesn't seem to work in the same way and the spike line isn't rendered or can occasionally be seen between the traces. Changing to a value of `9999` then works, but only over the traces and not between.

Interestingly, combining scatter and bar partially works, but results in some flickering when hovering the mouse back and forth.

pope1ni:
Annoyingly it looks as though my codepen didn't save everything, but I have partially fixed it to show enough of the problem.

etpinard:
Thanks for the report!

pope1ni:
Fixed by #4637.

PR #4627, opened 2020-03-11
fix for spikedistance set to infinity

### LINKED ISSUE #4626: spike lines do not appear when `spikedistance`
At least in some traces: https://codepen.io/antoinerg/pen/BaNYJve

To be clear, this test will fail: https://github.com/plotly/plotly.js/commit/9d1325c75c1526caf174f0c4c93504c8c71c8e81

Affects traces:
- bar
- ohlc

antoinerg:
Fixes #4626 and https://github.com/plotly/plotly.js/issues/3805

Commit https://github.com/plotly/plotly.js/commit/9d1325c75c1526caf174f0c4c93504c8c71c8e81 adds a test which fail.

For #4626:
Before: https://codepen.io/antoinerg/pen/BaNYJve
After: https://codepen.io/antoinerg/pen/oNXEENy

For #3805:
Before: https://codepen.io/anon/pen/eoPXJE
After: https://codepen.io/antoinerg/pen/ExjQQYw

cc @archmoj @alexcjohnson @etpinard

### DIFF
--- a/src/traces/bar/hover.js
+++ b/src/traces/bar/hover.js
@@ -39,7 +39,6 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     var isClosest = (hovermode === 'closest');
     var isWaterfall = (trace.type === 'waterfall');
     var maxHoverDistance = pointData.maxHoverDistance;
-    var maxSpikeDistance = pointData.maxSpikeDistance;
 
     var posVal, sizeVal, posLetter, sizeLetter, dx, dy, pRangeCalc;
 
@@ -161,7 +160,7 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     pointData.valueLabel = hoverLabelText(sa, pointData[sizeLetter + 'LabelVal']);
 
     // spikelines always want "closest" distance regardless of hovermode
-    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2 + maxSpikeDistance - maxHoverDistance;
+    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2;
     // they also want to point to the data value, regardless of where the label goes
     // in case of bars shifted within groups
     pointData[posLetter + 'Spike'] = pa.c2p(di.p, true);
--- a/test/jasmine/tests/hover_spikeline_test.js
+++ b/test/jasmine/tests/hover_spikeline_test.js
@@ -404,6 +404,33 @@ describe('spikeline hover', function() {
         .then(done);
     });
 
+    it('correctly responds to setting the spikedistance to -1 by increasing ' +
+        'the range of search for points to draw the spikelines to Infinity in a bar trace', function(done) {
+        var _mock = {
+            data: [{type: 'bar', y: [5, 6, 8, 10]}],
+            layout: {xaxis: {showspikes: true}}
+        };
+
+        Plotly.plot(gd, _mock).then(function() {
+            _hover({xval: 0, yval: 2});
+            _assert(
+                [[147.5, 370, 147.5, 241.75]],
+                []
+            );
+
+            _setSpikedistance(-1);
+        })
+        .then(function() {
+            _hover({xval: 0, yval: 2});
+            _assert(
+                [[147.5, 370, 147.5, 241.75]],
+                []
+            );
+        })
+        .catch(failTest)
+        .then(done);
+    });
+
     it('correctly responds to setting the spikedistance to 0 by disabling ' +
         'the search for points to draw the spikelines', function(done) {
         var _mock = makeMock('toaxis', 'closest');


src/components/fx/hover.js
```diff
@@ -549,7 +549,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
         var thisSpikeDistance;
         for(var i = 0; i < pointsData.length; i++) {
             thisSpikeDistance = pointsData[i].spikeDistance;
-            if(thisSpikeDistance < minDistance && thisSpikeDistance <= spikedistance) {
+            if(thisSpikeDistance <= minDistance && thisSpikeDistance <= spikedistance) {
```
antoinerg: The problem stems from `thisSpikeDistance` being `Infinity` and `minDistance` also being `Infinity`. Without an equal sign, this will never be true and no spikelines will appear and hence the bug.

archmoj:
@antoinerg Thanks very much for the PR.
Looks good to my eyes.
I'll let a second reviewer to give you dancers.

src/components/fx/hover.js
alexcjohnson: Are there any cases this changes *which* point we get spikes to? It might still be OK in that case, I just want to know what those cases are. The loop indicates this way if we have multiple items in `pointData` all with the same `spikeDistance`, previously we'd get the first one, but with the new logic we'll get the last one.

antoinerg: > Are there any cases this changes which point we get spikes to? 

@alexcjohnson I'm not sure how to proceed to find out.

antoinerg: Should we check whether `minDistance` is `Infinity` and handle that case separately?

archmoj: That's a much better solution, IMO.

archmoj:
```js
if((minDistance === Infinity || thisSpikeDistance < minDistance) &&
    thisSpikeDistance <= spikedistance
) {
```
looks like the right logic to apply.

antoinerg: Ok so I went with this safer approach. If `spikedistance` is `Infinity` we visit that `if` block. It should preserve any existing behavior but fix the edge case for `Infinity`.

See https://github.com/plotly/plotly.js/pull/4627/commits/583ad647bea336db2c7b844991a94624d762239b

alexcjohnson: The key is to first identify the relevant test cases. I suspect the relevant ones are (with `hovermode: 'x'`):
- Two scatter traces with exactly identical points, so there aren't closer and farther ones. Maybe we don't care which one gets the spike line... though offhand I'd suggest the last one, since it'll be drawn on top so you'll see a spike line color matching the visible point.
- Stacked bar traces with the cursor at a larger (or smaller) y value than any of the bars. Ideally I'd think it would be nice to draw the spike line to the one that's closest; but that would require getting away from using `Infinity` for all points outside the bar, at least for [`sizeFn`](https://github.com/plotly/plotly.js/blob/8f049fddbac0ca0382816984b8526857e9714fe6/src/traces/bar/hover.js#L104). Perhaps we give `Fx.inbox` another (optional) argument `failVal` that, in this case, will be set very large, and increasing the farther away you get from the bar either above or below, but not infinite. Or, at that point it seems silly to use `Fx.inbox` at all, just inline this logic.

archmoj:
Nicely done.
:dancer:

archmoj:
Please hold on. There is a new comment: https://github.com/plotly/plotly.js/pull/4627#discussion_r391292113.

src/components/fx/hover.js
antoinerg: Thanks for the clarification @alexcjohnson . I now better understand the purpose of this code. I will take another approach.

antoinerg:
I made a mess earlier and cleaned up the git history with a `git rebase`. I left commit https://github.com/plotly/plotly.js/pull/4627/commits/ae7db42b1d73ebcbc1a0513abf5ed78008a67271 with the failing test to fix and made a new solution in  https://github.com/plotly/plotly.js/pull/4627/commits/ebf75e6a89b477015fad590c1eede011682be031 .

antoinerg:
I think my first solution has to go back in otherwise we have too many non-scatter traces for which no spike lines appear for `spikedistance: -1`.

@alexcjohnson can you have a look at [branch `fix-spikedistance`](https://github.com/plotly/plotly.js/compare/fix-spikedistance). It contains new tests which make sure that:
- all cartesian traces produce spike lines when `spikedistance: -1` (first commit)
- test that spikeline is drawn up to the last point as you described
- test that spikeline is drawn up to the closest bar

If it's satisfactory, I can open a new PR or simply force push here. Let me know what you prefer.

antoinerg:
@archmoj you might also want to review my new branch described in https://github.com/plotly/plotly.js/pull/4627#issuecomment-598453328

alexcjohnson:
> @alexcjohnson can you have a look at branch fix-spikedistance. It contains new tests...

Looks great, my only question about it is the test "correctly select the closest bar even when setting spikedistance to -1" - you do `_hover({xpx: 600, ypx: 400});` and `_hover({xpx: 580, ypx: 200});`, which look like they're within the bounds of the blue and green bars at E? That's important (though it would be clearer with exactly the same `xpx`), but I would have expected to also see `ypx` corresponding to positions not *on* any bars, but above and below the entire stack - which would probably require increasing the range to create space above and below. I'm not sure if that works right now but I think the goal of #3805 would be that if you get a hover point from a bar, and you set `spikedistance=-1`, you should get a spike line to the closest one.

archmoj:
> If it's satisfactory, I can open a new PR or simply force push here. Let me know what you prefer.

Force-pushing on PRs is discouraged on the plotly.js repo.
Please simply add the commit to this PR or open a new PR.
Thank you!

antoinerg:
> my only question about it is the test "correctly select the closest bar even when setting spikedistance to -1" - you do `_hover({xpx: 600, ypx: 400});` and `_hover({xpx: 580, ypx: 200});`, which look like they're within the bounds of the blue and green bars at E?

I'm not completely sure: maybe it's above or below. I tested it manually on mock `bar_stack-with-gaps` to make sure we get the closest one.

> That's important (though it would be clearer with exactly the same `xpx`)

Yes they should have the same `xpx`, sorry for making things confusing. I will fix that.

> but I would have expected to also see ypx corresponding to positions not on any bars, but above and below the entire stack - which would probably require increasing the range to create space above and below. I'm not sure if that works right now 

Right now, for `bar` traces, spike lines will appear only if you hover inside the bar itself for any value  `spikedistance > 0` (this is also the case on `master`). My PR `fix-spikedistance` only fixes the case `spikedistance = -1` which is key for unified hover label. I'm not sure if I have time to tackle the other case (ie `spikedistance > 0 `) at the moment.

> if you get a hover point from a bar, and you set spikedistance=-1, you should get a spike line to the closest one.

I think this case is covered: https://codepen.io/antoinerg/pen/vYORyLV

cc @alexcjohnson

alexcjohnson:
You're so close! Right now you get a spike to the last bar when you're off the stack, not the closest one (in the gif below see behavior when the mouse is below, spike goes to the green bar). fixing that would require tweaking `sizefn` to include some large-but-not-infinite value when off the bar, that grows larger as you get farther away.
Kapture 2020-03-13 at 8 46 42

This is not a super important case, so if you want to move on that's fine, but if you do please make a new issue for this.

antoinerg:
> You're so close! Right now you get a spike to the last bar when you're off the stack, not the closest one (in the gif below see behavior when the mouse is below, spike goes to the green bar). fixing that would require tweaking `sizefn` to include some large-but-not-infinite value when off the bar, that grows larger as you get farther away.
> Kapture 2020-03-13 at 8 46 42
> 
> This is not a super important case, so if you want to move on that's fine, but if you do please make a new issue for this.

I missed that behavior! Let me see if I can get it right :) Thanks for the review @alexcjohnson  :male_detective:

antoinerg:
> Please simply add the commit to this PR or open a new PR.

I'll open a new PR. Closing in favor of #4637

PR #4635, opened 2020-03-12
expose parcoords context line color

### LINKED ISSUE #3939: Add option to set parcoords context line color and opacity
The deselected lines in all `parcoords` plots now use a constant color (i.e. `#777` and auto opacity logic (i.e. depending on the number of overlaid lines).
In some scenarios it may be useful to let the user override the default.

archmoj:
resolves #3939 | [Demo](https://codepen.io/MojtabaSamimi/pen/eYNMNvR?editors=0010)
by adding new attribute `unselected.line.color`.

TODO: 
- [x] finalize attribute name
- [x] add test

@plotly/plotly_js 
cc: @chriddyp

### DIFF
--- a/src/traces/parcoords/attributes.js
+++ b/src/traces/parcoords/attributes.js
@@ -141,18 +141,20 @@ module.exports = {
         description: 'The dimensions (variables) of the parallel coordinates chart. 2..60 dimensions are supported.'
     }),
 
-    bgline: {
-        color: {
-            valType: 'color',
-            dflt: '#777',
-            role: 'style',
-            editType: 'plot',
-            description: [
-                'Sets the color of lines in the background',
-                'i.e. unselected lines.'
-            ].join(' ')
+    unselected: {
+        line: {
+            color: {
+                valType: 'color',
+                dflt: '#777',
+                role: 'style',
+                editType: 'plot',
+                description: [
+                    'Sets the color of lines in the background',
+                    'i.e. unselected lines.'
+                ].join(' ')
+            },
+            editType: 'calc'
         },
-
         editType: 'calc'
     },
 
--- a/src/traces/parcoords/attributes.js
+++ b/src/traces/parcoords/attributes.js
@@ -141,6 +141,21 @@ module.exports = {
         description: 'The dimensions (variables) of the parallel coordinates chart. 2..60 dimensions are supported.'
     }),
 
+    bgline: {
+        color: {
+            valType: 'color',
+            dflt: '#777',
+            role: 'style',
+            editType: 'plot',
+            description: [
+                'Sets the color of lines in the background',
+                'i.e. unselected lines.'
+            ].join(' ')
+        },
+
+        editType: 'calc'
+    },
+
     line: extendFlat({editType: 'calc'},
         colorScaleAttrs('line', {
             // the default autocolorscale isn't quite usable for parcoords due to context ambiguity around 0 (grey, off-white)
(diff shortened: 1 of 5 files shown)


src/traces/parcoords/attributes.js
```diff
@@ -141,6 +141,21 @@ module.exports = {
         description: 'The dimensions (variables) of the parallel coordinates chart. 2..60 dimensions are supported.'
     }),
 
+    bgline: {
+        color: {
```
alexcjohnson: If we follow the example of `scatter` we get `unselected.line.color`. It seems a bit overly-nested, though `splom` does this too and all it has in `unselected` is `marker`. I think we should do it, gives us maximum extensibility as well as matching other trace types.

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4635/commits/9b4711b74cd5c30dfccdce3ef27b875c2856f65a.

src/traces/parcoords/attributes.js
```diff
@@ -141,6 +141,23 @@ module.exports = {
...
+                role: 'style',
+                editType: 'plot',
+                description: [
+                    'Sets the color of lines in the background',
+                    'i.e. unselected lines.'
+                ].join(' ')
+            },
+            editType: 'calc'
+        },
+        editType: 'calc'
```
alexcjohnson: if `color` has `editType: 'plot'` and it's the only thing in these containers, they can have `'plot'` as well.

archmoj: I recall, Etienne mentioned that for containers we should use `calc`.

test/jasmine/tests/parcoords_test.js
```diff
@@ -910,6 +922,62 @@ describe('parcoords Lifecycle methods', function() {
...
+
+            return Plotly.restyle(gd, 'unselected.line.color', '#00F');
+        })
+        .then(function() {
+            var rgb = getAvgPixelByChannel(testLayer);
+            expect(rgb[0]).toBe(0, 'no red');
+            expect(rgb[1]).toBe(0, 'no green');
+            expect(rgb[2]).not.toBe(0, 'blue');
+
+            return Plotly.restyle(gd, 'unselected.line.color', 'rgba(0,0,0,0)');
```
alexcjohnson: Let's turn one of these `restyle` calls into a `react` call.

archmoj: If you don't mind, I'd be happy to keep this one, and add a new one to use `react`.

archmoj: Done in https://github.com/plotly/plotly.js/pull/4635/commits/1af1c90bf445b37d2ddab8e1942358fb5ab2d3c2.

src/traces/parcoords/attributes.js
alexcjohnson: I believe containers should use the most intense `editType` of anything inside them, and should use `calc` if there's any array or `arrayOk` attribute inside.

test/jasmine/tests/parcoords_test.js
alexcjohnson: That works for me!

src/traces/parcoords/attributes.js
archmoj: Thanks for the note regarding `arrayOk` items.
Changed all to `plot` in https://github.com/plotly/plotly.js/pull/4635/commits/b4dabb224c37724951f638462dcb1421ab85011e.

PR #4637, opened 2020-03-13
fix for spikedistance=-1

### LINKED ISSUE #4626: spike lines do not appear when `spikedistance`

antoinerg:
Supersedes https://github.com/plotly/plotly.js/pull/4627 and fixes #4626 and #3805

For #4626:
Before: https://codepen.io/antoinerg/pen/BaNYJve
After: https://codepen.io/antoinerg/pen/oNXEENy

For #3805:
Before: https://codepen.io/antoinerg/pen/eYNMrBq
After: https://codepen.io/antoinerg/pen/ExjQQYw

cc @archmoj @alexcjohnson @etpinard

I opened issue https://github.com/plotly/plotly.js/issues/4638 to deal with an edge case.

### DIFF
--- a/src/traces/bar/hover.js
+++ b/src/traces/bar/hover.js
@@ -160,7 +160,7 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     pointData.valueLabel = hoverLabelText(sa, pointData[sizeLetter + 'LabelVal']);
 
     // spikelines always want "closest" distance regardless of hovermode
-    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2;
+    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2 - maxHoverDistance;
     // they also want to point to the data value, regardless of where the label goes
     // in case of bars shifted within groups
     pointData[posLetter + 'Spike'] = pa.c2p(di.p, true);
--- a/src/traces/bar/hover.js
+++ b/src/traces/bar/hover.js
@@ -39,7 +39,6 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     var isClosest = (hovermode === 'closest');
     var isWaterfall = (trace.type === 'waterfall');
     var maxHoverDistance = pointData.maxHoverDistance;
-    var maxSpikeDistance = pointData.maxSpikeDistance;
 
     var posVal, sizeVal, posLetter, sizeLetter, dx, dy, pRangeCalc;
 
@@ -161,7 +160,7 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     pointData.valueLabel = hoverLabelText(sa, pointData[sizeLetter + 'LabelVal']);
 
     // spikelines always want "closest" distance regardless of hovermode
-    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2 + maxSpikeDistance - maxHoverDistance;
+    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2;
     // they also want to point to the data value, regardless of where the label goes
     // in case of bars shifted within groups
     pointData[posLetter + 'Spike'] = pa.c2p(di.p, true);
--- a/test/jasmine/tests/hover_spikeline_test.js
+++ b/test/jasmine/tests/hover_spikeline_test.js
@@ -404,6 +404,29 @@ describe('spikeline hover', function() {
         .then(done);
     });
 
+    it('correctly select the closest bar even when setting spikedistance to -1', function(done) {
+        var mock = require('@mocks/bar_stack-with-gaps');
+        var mockCopy = Lib.extendDeep({}, mock);
+        mockCopy.layout.xaxis.showspikes = true;
+        mockCopy.layout.yaxis.showspikes = true;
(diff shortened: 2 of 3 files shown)


archmoj:
@antoinerg thanks for the follow ups!
Would you mind updating `before` codepen to use `v1.52.3` instead of latest?
That way one could easily compare the two after a new release.

src/traces/bar/hover.js
```diff
@@ -161,7 +160,7 @@ function hoverOnBars(pointData, xval, yval, hovermode) {
     pointData.valueLabel = hoverLabelText(sa, pointData[sizeLetter + 'LabelVal']);
 
     // spikelines always want "closest" distance regardless of hovermode
-    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2 + maxSpikeDistance - maxHoverDistance;
+    pointData.spikeDistance = (sizeFn(di) + thisBarPositionFn(di)) / 2 - maxHoverDistance;
```
antoinerg: `maxHoverDistance` is necessary here because it is added in the function `sizeFn`. Now, why would a function compute the size of a bar would contain `maxHoverDistance` is beyond my comprehension.

test/jasmine/tests/hover_spikeline_test.js
```diff
@@ -404,6 +404,29 @@ describe('spikeline hover', function() {
...
     });
 
+    it('correctly select the closest bar even when setting spikedistance to -1', function(done) {
+        var mock = require('@mocks/bar_stack-with-gaps');
+        var mockCopy = Lib.extendDeep({}, mock);
+        mockCopy.layout.xaxis.showspikes = true;
+        mockCopy.layout.yaxis.showspikes = true;
+        mockCopy.layout.spikedistance = -1;
+
+        Plotly.newPlot(gd, mockCopy)
```
archmoj: Non-blocking:
We could us `newPlot` > `plot` here.

test/jasmine/tests/hover_spikeline_test.js
```diff
@@ -593,4 +616,139 @@ describe('spikeline hover', function() {
         .catch(failTest)
         .then(done);
     });
+
+    it('correctly draws lines up to the last point', function(done) {
+        Plotly.newPlot(gd, [
```
archmoj: Non-blocking:
We could us `newPlot` > `plot` here.

test/jasmine/tests/hover_spikeline_test.js
```diff
@@ -593,4 +616,139 @@ describe('spikeline hover', function() {
...
+                        spikedistance: -1,
+                        xaxis: {showspikes: true},
+                        yaxis: {showspikes: true},
+                        zaxis: {showspikes: true},
+                        title: {text: trace.type},
+                        width: 400, height: 400
+                    }
+                };
+
+                Plotly.newPlot(gd, mock)
```
archmoj: Non-blocking:
We could us `newPlot` > `plot` here.

test/jasmine/tests/hover_spikeline_test.js
alexcjohnson: Is that desirable? `plot` is not part of the API we want anyone using, so I'd kind of rather not use it in tests either, even if it's a little faster.

src/traces/bar/hover.js
alexcjohnson: Yeah, super confusing. It's not computing the size of a bar, it's computing a distance measure along the size axis. Is it necessary to remove `maxSpikeDistance` here? Or it just doesn't have an understood purpose?

test/jasmine/tests/hover_spikeline_test.js
archmoj: Thanks for the note. Resolving...
And please feel free to open an issue if you prefer to chage similar cases in all the tests.

src/traces/bar/hover.js
antoinerg: > Is it necessary to remove maxSpikeDistance here?

Keeping `maxSpikeDistance` in leads to:
Screenshot_2020-03-13_15-47-32

> Or it just doesn't have an understood purpose?

I'll be honest: I also don't understand its purpose. What is `spikedistance` supposed to represent? Why would we add `maxSpikeDistance` and substract `maxHoverDistance` to it at the trace-level? It seems to me that this logic should live in `fx/hover.js` no?

test/jasmine/tests/hover_spikeline_test.js
alexcjohnson: I wouldn't bother converting the rest unless there are problems. Sometimes there's a good reason to go for the faster one, in particular I seem to recall a lot of overhead to `newPlot` in WebGL tests? But that overhead comes with more assurance that there's no cross-test interaction, which has lead to hard-to-debug errors in the past. @etpinard any words of wisdom to add here?

src/traces/bar/hover.js
alexcjohnson: yeah this logic should certainly be cleaned up at some point... super hacky. Perhaps as part of #4638 🙏

PR #4639, opened 2020-03-13
Fix reverse autorange on cartesian axes with breaks

archmoj:
A follow up of remaining work #4614 related to `autorange: 'reversed'` cases on axes with `breaks`
as well as some simplification for `l2p` & `p2l` functions.

@plotly/plotly_js


src/plots/cartesian/set_convert.js
```diff
@@ -575,12 +566,13 @@ module.exports = function setConvert(ax, fullLayout) {
...
-            var signAx = rl0 > rl1 ? -1 : 1;
+            var axReverse = rl0 > rl1;
+            var signAx = axReverse ? -1 : 1;
 
             if(ax._breaks.length) {
                 for(i = 0; i < ax._breaks.length; i++) {
                     brk = ax._breaks[i];
-                    ax._lBreaks += (brk.max - brk.min);
+                    ax._lBreaks += brk.max - brk.min;
                 }
```
etpinard: One small recommendation: I think it would be best to keep `ax._lBreaks` positive-definite. So many below this loop we could have

```js
ax._lBreaks = Math.abs(ax._lBreaks)
```

?

archmoj: I tried that at one point. However, allowing negative helped handling reversed cases much easier. For now I added a comment in front of `_lBreaks` description.

etpinard: > For now I added a comment in front of `_lBreaks` description.

That's great!

etpinard:
:dancer: - thanks very much @archmoj long live plotly.js!

PR #4641, opened 2020-03-14
rename axis.breaks to axis.rangebreaks

archmoj:
Follow up of feature added in #4614 for `v1.53.0`
Fixes `codegen` noting that `break` is a reserved keyword in python.

@plotly/plotly_js


src/plots/cartesian/layout_attributes.js
```diff
@@ -284,17 +284,17 @@ module.exports = {
             role: 'info',
             editType: 'calc',
             description: [
-                'Determines a pattern on the time line that generates breaks.',
+                'Determines a pattern on the time line that generates rangebreaks.',
```
alexcjohnson:
```suggestion
                'Determines a pattern on the time line that generates breaks.',
```
Not a big deal, but a few of these were the *word* breaks, not the attribute name...

src/plots/cartesian/layout_attributes.js
```diff
@@ -284,17 +284,17 @@ module.exports = {
...
                 'If *%w* - Sunday-based weekday as a decimal number [0, 6].',
                 'If *%H* - hour (24-hour clock) as a decimal number [0, 23].',
                 'These are the same directive as in `tickformat`, see',
                 'https://github.com/d3/d3-time-format#locale_format',
                 'for more info.',
                 'Examples:',
                 '- { pattern: \'%w\', bounds: [6, 0], operation: \'[]\' }',
-                '  breaks from Saturday to Monday (i.e. skips the weekends).',
+                '  rangebreaks from Saturday to Monday (i.e. skips the weekends).',
```
alexcjohnson:
```suggestion
                '  breaks from Saturday to Monday (i.e. skips the weekends).',
```

src/plots/cartesian/layout_attributes.js
```diff
@@ -284,17 +284,17 @@ module.exports = {
...
                 '- { pattern: \'%H\', bounds: [17, 8] }',
-                '  breaks from 5pm to 8am (i.e. skips non-work hours).'
+                '  rangebreaks from 5pm to 8am (i.e. skips non-work hours).'
```
alexcjohnson:
```suggestion
                '  breaks from 5pm to 8am (i.e. skips non-work hours).'
```

src/plots/cartesian/layout_attributes.js
```diff
@@ -270,7 +270,7 @@ module.exports = {
             ],
             editType: 'calc',
             description: [
-                'Sets the lower and upper bounds of this axis break.',
+                'Sets the lower and upper bounds of this axis rangebreak.',
```
alexcjohnson: hmm somehow as a noun I'm ambivalent about it - probably I'd change the noun usages in descriptions back to break(s) but it's not as clear to me as when break is a verb...

src/plots/gl2d/camera.js
```diff
@@ -158,7 +158,7 @@ function createCamera(scene) {
                                 Math.abs(dx) * dydx * (dy >= 0 ? 1 : -1);
 
                             // gl-select-box clips to the plot area bounds,
-                            // which breaks the axis constraint, so don't allow
+                            // which rangebreaks the axis constraint, so don't allow
```
alexcjohnson:
```suggestion
                            // which breaks the axis constraint, so don't allow
```

PR #4480, opened 2020-01-10
Implement surface opacityscale & correct display of transparent surfaces

archmoj:
Resolves #4331 | [demo](https://codepen.io/MojtabaSamimi/pen/jOEZagL)
and
fixes #4478 | [before](https://codepen.io/MojtabaSamimi/pen/abzYYxK) vs [after](https://codepen.io/MojtabaSamimi/pen/BayrxGZ)

This PR adds a new attribute titled `opacityscale` to `surface` trace.
Using that it is possible to make certain parts of the surface plot more transparent.
Also this PR removes extra transparency displayed by surface plots which would match `surface.opacity` and `mesh3d.opacity`.

@plotly/plotly_js

PR #4642, opened 2020-03-15
Fix to display transparent surface with contours when opacity is less than one 

archmoj:
Fixes a bug introduced in #4480 to avoid regression in `v1.53.0`.
As illustrated in this [codepen](https://codepen.io/MojtabaSamimi/pen/yLNjBWb?editors=0010) the `surface` with `opacity` < 1 and `contours` was not displayed transparent!

@plotly/plotly_js


package.json
```diff
@@ -88,7 +88,7 @@
     "gl-select-box": "^1.0.3",
     "gl-spikes2d": "^1.0.2",
     "gl-streamtube3d": "^1.4.0",
-    "gl-surface3d": "^1.5.0",
+    "gl-surface3d": "git://github.com/gl-vis/gl-surface3d.git#68c1ee93f6c1d9674bea3aa120ddf1da30304db6",
```
archmoj: https://github.com/gl-vis/gl-surface3d/commit/68c1ee93f6c1d9674bea3aa120ddf1da30304db6

PR #4644, opened 2020-03-15
Handle hover over rangebreaks for bar-like traces

archmoj:
Following #4614, this PR fixes a bug for `bar`-like traces
and avoids console errors when hover crosses an axis which has a `rangebreak`.

@plotly/plotly_js


antoinerg:
LGTM thanks @archmoj 
:dancer:

Issue #4655, opened 2020-03-17
Range break semantics


nicolaskruchten:
Per Slack, we want to have three modes of operation, after #4653 is merged:

* pattern="day of week"/bounds: in this mode, we accept only integer values in bounds, and operation is always `[]`
* pattern="hour"/bounds/operation: in this mode we accept floats in bounds and operation is honoured for points at the boundaries
* values/operation: in this mode we accept ISO datetime values and operation is honoured for points at the boundaries

nicolaskruchten:
@alexcjohnson you've thought about this more than I have... can we unify the behaviour of `operation` even in day-of-week mode somehow with respect to points on the boundaries?

archmoj:
> * pattern="hour"/bounds/operation: in this mode we accept floats in bounds and operation is honoured for points at the boundaries

Could you please provide an input+output example for this case?

archmoj:
> * values/operation: in this mode we accept ISO datetime values and operation is honoured for points at the boundaries

To be "honoured", should the points be included in the plot?  i.e. excluded from the `rangebreak`

nicolaskruchten:
For "hour" and "values" modes I'm just describing the current behaviour I believe.

archmoj:
Regarding the first case,
Another option may be 
```js
pattern: 'day', bounds: ['Sat', 'Sun']
```
instead of 
```js
pattern: 'day of week', bounds: [6, 0]
```

alexcjohnson:
Yes, for `'hour'` and `'values'` modes @nicolaskruchten has, I believe, described the status quo accurately, and we can keep that.

Right now in `'day of week'`, after we use `operation` to determine which days to remove (which I think we can all agree at this point we're no longer going to do, but keep reading because I have some concerns about what we should do instead), the boundary behavior is what in the other modes would have been `operation='[)'` - that is, the left edge of the break range, ie Saturday at 00:00:00.000, is "within the break" so a data point there will not be drawn, but the right edge of the break range, Monday at 00:00:00.000 *will* be drawn. So if you give this data:
Friday 23:59:59.998, Friday 23:59:59.999, Monday 00:00:00.000, Monday 00:00:00.001
The result will be four equally spaced points, no values between that last Friday and the first Monday point would be allowed, and no two points with different data values would ever share the same screen position.

This seems to me like it will work well in the short term for the two main cases I can think of:
[...]

PR #4661, opened 2020-03-18
Handle weekday strings in rangebreaks

archmoj:
Follow up of #4614 and #4655 
- [x] convert string `bounds` e.g. `Monday` (or 'mon`) to day of week indices between [0, 6]. 
- [x] auto default to `day of week` when using day strings e.g. `Monday` or 'mon`.
- [x] update description
- [x] add jasmine tests

[Demo](https://codepen.io/MojtabaSamimi/pen/BaNVPdN?editors=0010).

@plotly/plotly_js

### DIFF
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -8,6 +8,8 @@
 
 'use strict';
 
+var isNumeric = require('fast-isnumeric');
+
 var Registry = require('../../registry');
 var Lib = require('../../lib');
 
@@ -22,6 +24,7 @@ var handleLineGridDefaults = require('./line_grid_defaults');
 var setConvert = require('./set_convert');
 
 var DAY_OF_WEEK = require('./constants').WEEKDAY_PATTERN;
+var HOUR = require('./constants').HOUR_PATTERN;
 
 /**
  * options: object containing:
@@ -175,13 +178,44 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
                 }
             }
             var pattern = coerce('pattern', dfltPattern);
-
-            if(bnds.length === 2 && pattern === DAY_OF_WEEK) {
+            if(pattern && bnds.length !== 2) {
+                itemOut.enabled = false;
+                return;
+            }
+            if(pattern === DAY_OF_WEEK) {
                 for(i = 0; i < 2; i++) {
                     q = indexOfDay(bnds[i]);
                     if(q) {
                         // convert to integers i.e 'Sunday' --> 0
                         itemOut.bounds[i] = bnds[i] = q - 1;
+                    }
+                }
+            }
+            if(pattern) {
+                // ensure types and ranges
+                for(i = 0; i < 2; i++) {
+                    q = bnds[i];
+                    switch(pattern) {
+                        case DAY_OF_WEEK :
+                            if(isNumeric(q)) q = Math.floor(q);
+
+                            q = Math.floor(+q);
+                            if(!(q >= 0 && q < 7)) {
+                                itemOut.enabled = false;
+                                return;
+                            }
+                            // use int [0, 7)
+                            itemOut.bounds[i] = bnds[i] = q;
+                            break;
+
+                        case HOUR :
+                            if(!(q >= 0 && q <= 24)) { // accept 24
+                                itemOut.enabled = false;
+                                return;
+                            }
+                            // use float [0, 24]
+                            itemOut.bounds[i] = bnds[i] = q;
+                            break;
                     }
                 }
             }
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -225,6 +225,6 @@ var dayStrToNum = {
 function indexOfDay(v) {
     if(typeof v !== 'string') return;
     return dayStrToNum[
-        String(v).substr(0, 3).toLowerCase()
+        v.substr(0, 3).toLowerCase()
     ];
 }
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -160,12 +160,7 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
 
     if(enabled) {
         var bnds = coerce('bounds');
-
         if(bnds && bnds.length >= 2) {
-            if(bnds.length > 2) {
-                itemOut.bounds = itemOut.bounds.slice(0, 2);
-            }
-
             var dfltPattern = '';
             var i, q;
             if(bnds.length === 2) {
@@ -178,10 +173,6 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
                 }
             }
             var pattern = coerce('pattern', dfltPattern);
-            if(pattern && bnds.length !== 2) {
-                itemOut.enabled = false;
-                return;
-            }
             if(pattern === DAY_OF_WEEK) {
                 for(i = 0; i < 2; i++) {
                     q = indexOfDay(bnds[i]);
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -165,17 +165,19 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
 
             var dfltPattern = '';
             var i, q;
-            for(i = 0; i < bnds.length; i++) {
-                q = indexOfDay(bnds[i]);
-                if(q) {
-                    dfltPattern = DAY_OF_WEEK;
-                    break;
+            if(bnds.length === 2) {
+                for(i = 0; i < 2; i++) {
+                    q = indexOfDay(bnds[i]);
+                    if(q) {
+                        dfltPattern = DAY_OF_WEEK;
+                        break;
+                    }
                 }
             }
             var pattern = coerce('pattern', dfltPattern);
 
-            if(pattern === DAY_OF_WEEK) {
-                for(i = 0; i < bnds.length; i++) {
+            if(bnds.length === 2 && pattern === DAY_OF_WEEK) {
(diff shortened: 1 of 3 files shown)


src/plots/cartesian/layout_attributes.js
```diff
@@ -280,16 +280,17 @@ module.exports = {
         pattern: {
             valType: 'enumerated',
             values: [DAY_OF_WEEK, HOUR, ''],
-            dflt: '',
             role: 'info',
             editType: 'calc',
             description: [
                 'Determines a pattern on the time line that generates breaks.',
-                'If *' + DAY_OF_WEEK + '* - Sunday-based weekday as a decimal number [0, 6].',
+                'If *' + DAY_OF_WEEK + '* - days of the week in English e.g. \'Sunday\' or `\sun\`',
```
alexcjohnson: Let's add a note like "matching is case-insensitive and considers only the first three characters."

src/plots/cartesian/axis_defaults.js
```diff
@@ -161,6 +163,27 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
                 itemOut.bounds = itemOut.bounds.slice(0, 2);
```
alexcjohnson: I don't think you have to do this `slice(0, 2)` - `coerce` for `info_array` already does this.
But if somehow I'm mistaken about that, you'd need to also shorten `bnds` so you don't lengthen `itemOut.bounds` again in the loops below.

src/plots/cartesian/layout_attributes.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4661/commits/05ab23ca6f8d51100f04c9bdc91e053d926f0c24.

src/plots/cartesian/axis_defaults.js
```diff
@@ -189,3 +210,18 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
...
+    'tue',
+    'wed',
+    'thu',
+    'fri',
+    'sat'
+];
+
+function indexOfDay(v) {
+    var str = String(v).substr(0, 3).toLowerCase();
+    return weekSTR.indexOf(str);
```
alexcjohnson: This could be sped up a bit with an object lookup rather than `indexOf`
```js
var dayStrToNum = {
    sun: 0,
    mon: 1,
    ...
}

return dayStrToNum[str];
```

src/plots/cartesian/axis_defaults.js
archmoj: I don't see any place where `itemOut.bounds.length` is changed.

src/plots/cartesian/axis_defaults.js
alexcjohnson: and possibly bail out early if `v` isn't already a string, rather than converting to a string only to discard it later.

src/plots/cartesian/axis_defaults.js
```diff
@@ -161,6 +163,27 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
...
+            }
+            var pattern = coerce('pattern', dfltPattern);
+
+            if(pattern === DAY_OF_WEEK) {
+                for(i = 0; i < bnds.length; i++) {
+                    q = indexOfDay(bnds[i]);
+                    if(q !== -1) {
+                        // convert to integers i.e 'Monday' --> 1
+                        itemOut.bounds[i] = bnds[i] = q;
+                    }
```
alexcjohnson: Do we also want to validate other values here? ie turn `'0'` through `'6'` into numbers and reject all other strings, numbers outside that range, and non-integers?

src/plots/cartesian/axis_defaults.js
alexcjohnson: `itemOut.bounds[i] = bnds[i] = q;`

src/plots/cartesian/axis_defaults.js
archmoj: cc: @nicolaskruchten

src/plots/cartesian/axis_defaults.js
archmoj: Good calls! Done in https://github.com/plotly/plotly.js/pull/4661/commits/db2c3a3af9018477def1a6abb22b88b70c5081ab.
Curious to know if there is a good article on this?
At the moment, there are quite a number of places that `indexOf !== -1` is applied, so please feel free to open an issue for that.
Thanks.

src/plots/cartesian/axis_defaults.js
```diff
@@ -189,3 +210,21 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
         }
     }
 }
+
+// these numbers are one more than what bounds would be mapped to
```
alexcjohnson: clever way to avoid having to distinguish 0 from undefined :)

src/plots/cartesian/axis_defaults.js
```diff
@@ -189,3 +210,21 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
...
+    wed: 4,
+    thu: 5,
+    fri: 6,
+    sat: 7
+};
+
+function indexOfDay(v) {
+    if(typeof v !== 'string') return;
+    return dayStrToNum[
+        String(v).substr(0, 3).toLowerCase()
```
alexcjohnson:
```suggestion
        v.substr(0, 3).toLowerCase()
```

src/plots/cartesian/axis_defaults.js
alexcjohnson: We can ignore this for now, but at some point it may be nice to play with what happens when users enter bad data, and whether validating it here would help them figure out how to fix it.

src/plots/cartesian/axis_defaults.js
archmoj: OMG. Fixed in https://github.com/plotly/plotly.js/pull/4661/commits/c48f551fb2c93fe856f340a55c470f6a8c54ea1b.

src/plots/cartesian/axis_defaults.js
archmoj: Good call. I revised the new logic in https://github.com/plotly/plotly.js/pull/4661/commits/550a3756da650fe3f3481fa13ce42092c39e5000.

src/plots/cartesian/axis_defaults.js
archmoj: OK. I think I could fix that in this PR.

nicolaskruchten:
I <3 this!! cc @chriddyp 

image

src/plots/cartesian/axis_defaults.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4661/commits/e481a13deba3322d6279dd4f0acddc04f297ebbd.

src/plots/cartesian/axis_defaults.js
alexcjohnson: Again, you don't need to worry about the length of `bnds` at all - `coerce` handles that so `bnds` (and `itemOut.bounds`) is either `undefined` (since it has no default value) or it's an array of length 2 - or less, I just tested this, if you pass in a shorter array `coerce` isn't going to pad it since the inner values don't have defaults either, but that's OK, it'll just look like undefined for missing values and get filtered out in the type checks.

Can you please take all these length checks out? We put a lot of effort into making `coerce` do the right thing so the individual `supplyDefaults` functions could be simple and readable.

archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/4661/commits/84c2a6fa29a7b015a010ab5b52033a1318b9fece.

src/plots/cartesian/axis_defaults.js
```diff
@@ -155,10 +160,55 @@ function rangebreaksDefaults(itemIn, itemOut, containerOut) {
...
+            }
+            if(pattern) {
+                // ensure types and ranges
+                for(i = 0; i < 2; i++) {
+                    q = bnds[i];
+                    switch(pattern) {
+                        case DAY_OF_WEEK :
+                            if(isNumeric(q)) q = Math.floor(q);
+
+                            q = Math.floor(+q);
```
alexcjohnson: redundant - and dangerous - `q=''` would be turned into `0` for example. We should only cast to number when the value has already passed `isNumeric` (but might be a numeric string).

Also I don't think we should `floor`, we should just discard values that aren't integers (ie `enabled=false`). For one thing, that way if we ever add support for fractional days it won't break any existing behavior.

Can't this numeric validation be pushed into the `if(pattern === DAY_OF_WEEK)` loop above, and only executed if we didn't already convert from a day name to integer?

archmoj: Good call. Fixed in https://github.com/plotly/plotly.js/pull/4661/commits/9a0416aefb17fd0d5ac3bd26f91abb721b934a6d.

src/plots/cartesian/axis_defaults.js
alexcjohnson: Not sure about an article... you find a lot of people saying "hash lookup is O(1) but indexOf is O(n)" but that's as far as it usually goes.

Here's a jsperf showing the general trends: https://jsperf.com/array-indexof-vs-object-key-lookup2/12 - basically, object lookups hardly scale with the number of keys but `indexOf` scales linearly. So `indexOf` can be a nice way to simplify `if (a===b || a===c || a===d)` -> `if ([b,c,d].indexOf(a) !== -1)` - but more than a few items and it's better to turn the problem into an object lookup.

Also of course object lookups only support strings... looking forward to when we stop supporting IE and can use [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)!

archmoj: @alexcjohnson thank you!!

PR #4664, opened 2020-03-19
compare/unified hover include all points at same coordinate

### LINKED ISSUE #4656: compare/unified hover should include all points at same coordinate
In x-unified and x-compare modes, it should never be possible for there to be two data with the same x-coordinate and only one of them is displayed in a hoverlabel. I know compare mode doesn't currently behave this way but I consider it a bug.

The solution proposed by @alexcjohnson seems good to me:

* Step 1: find hover points the way we do now
* Step 2: loop back over traces not in the set of hover points we already found, and see if any of them have a point with position-axis value exactly matching what we found for the other point(s)

antoinerg:
Fixes #4656 

Codepen: https://codepen.io/antoinerg/pen/gOpKwPg

First commit simply wrap existing logic into a function. Second commit changes the existing behavior.

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -45,6 +45,8 @@ var YSHIFTY = Math.sin(YA_RADIANS);
 var HOVERARROWSIZE = constants.HOVERARROWSIZE;
 var HOVERTEXTPAD = constants.HOVERTEXTPAD;
 
+var XY = {x: 1, y: 1};
+
 // fx.hover: highlight data on hover
 // evt can be a mousemove event, or an object with data about what points
 //   to hover on
@@ -630,7 +632,7 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
     // If in compare mode, select every point at position
     if(hoverData[0].length !== 0 &&
-      ['x', 'y'].indexOf(mode) !== -1 &&
+      XY[mode] &&
       hoverData[0].trace.type !== 'splom' // TODO: add support for splom
     ) {
         var hd = hoverData[0];
@@ -657,7 +659,6 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
         // Remove duplicated hoverData points
         // note that d3 also filters identical points in the rendering steps
-        // TODO: use ES6 map
         var repeated = {};
         hoverData = hoverData.filter(function(hd) {
             var key = hoverDataKey(hd);
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -365,176 +365,178 @@ function _hover(gd, evt, subplot, noHoverEvent) {
     // find the closest point in each trace
     // this is minimum dx and/or dy, depending on mode
     // and the pixel position for the label (labelXpx, labelYpx)
-    for(curvenum = 0; curvenum < searchData.length; curvenum++) {
-        cd = searchData[curvenum];
+    function findHoverPoints(vals) {
+        for(curvenum = 0; curvenum < searchData.length; curvenum++) {
+            cd = searchData[curvenum];
 
-        // filter out invisible or broken data
-        if(!cd || !cd[0] || !cd[0].trace) continue;
+            // filter out invisible or broken data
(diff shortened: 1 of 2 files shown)


src/components/fx/hover.js
```diff
@@ -625,6 +628,46 @@ function _hover(gd, evt, subplot, noHoverEvent) {
 
     hoverData.sort(function(d1, d2) { return d1.distance - d2.distance; });
 
+    // If in compare mode, select every point at position
+    if(hoverData[0].length !== 0 &&
+      ['x', 'y'].indexOf(mode) !== -1 &&
```
archmoj: concerning
```js
['x', 'y'].indexOf(mode) !== -1 
```
creating a dynamic array of strings just for index checking might be slow.
Could we create a constant e.g.
```js
var XY = ['x', 'y'];
```
at the upper scope and reuse that here?

src/components/fx/hover.js
```diff
@@ -625,6 +628,46 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
+      hoverData[0].trace.type !== 'splom' // TODO: add support for splom
+    ) {
+        var hd = hoverData[0];
+        var cd0 = hd.cd[hd.index];
+        var isGrouped = (fullLayout.boxmode === 'group' || fullLayout.violinmode === 'group');
```
archmoj: Non-blocking:
Even for `group` which is an array of chars, one may use a const variable.

src/components/fx/hover.js
```diff
@@ -625,6 +628,46 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
+        if(ax.type === 'category') xVal = ax._categoriesMap[xVal];
+        if(ax.type === 'date') xVal = ax.d2c(xVal);
+        if(cd0 && cd0.t && cd0.t.posLetter === ax._id && isGrouped) {
+            xVal += cd0.t.dPos;
+        }
+
+        var yVal = hd.yVal;
+        ax = hd.ya;
+        if(ax.type === 'category') yVal = ax._categoriesMap[yVal];
+        if(ax.type === 'date') yVal = ax.d2c(yVal);
```
archmoj: Here is the second time checking those types.
Let's make 
```js
var isCategory = ax.type === 'category';
var isDate = ax.type === 'date';
```
or use of a `switch` & `case` or put them in one `if` statement.

src/components/fx/hover.js
alexcjohnson: or even `var XY = {x: 1, y: 1}` -> `if(XY[mode])` 😅

src/components/fx/hover.js
```diff
@@ -625,6 +628,46 @@ function _hover(gd, evt, subplot, noHoverEvent) {
...
+        if(cd0 && cd0.t && cd0.t.posLetter === ax._id && isGrouped) {
+            yVal += cd0.t.dPos;
+        }
+
+        findHoverPoints(xVal, yVal);
+
+        // Remove duplicated hoverData points
+        // note that d3 also filters identical points in the rendering steps
+        // TODO: use ES6 map
```
alexcjohnson: I don't think we can make that a TODO just yet - perhaps if & when we convert the whole syntax to ES6 we can make sure we include polyfills for this and other feature, THEN we can use it whenever we wish...

src/components/fx/hover.js
antoinerg: It's not the same axis object the second time around!

antoinerg:
> This looks great. Just a few nonblocking comments.

Great! I'll merge as soon as the tests pass

src/components/fx/hover.js
antoinerg: Done in https://github.com/plotly/plotly.js/pull/4664/commits/b8b9395f573a8b70e8263aac2ef154de49d6de5f

src/components/fx/hover.js
archmoj: Oops. Resolving.

archmoj:
Nicely done! 
:dancer: :dancer:

Issue #4667, opened 2020-03-20
New rangebreaks and unified hover should work from templates


nicolaskruchten:
I should be able to set `rangebreaks` and `hovermode="x unified"` inside `layout.template.layout` instead of `layout` and have things work identically.

As an aside: this is an invariant which should be the case for all attributes! I just QA'ed these two by chance but we should probably look into previous recently-added attributes to ensure that they work.

nicolaskruchten:
Possibly see https://github.com/plotly/plotly.js/pull/4620/files#diff-1602f9e66d27cf60c7469f31f03b8053

archmoj:
Closed via #4669 and #4670.

PR #4669, opened 2020-03-22
Set unified hovermode via template

archmoj:
Addressing #4667 i.e. to set unified `hovermode` via `template`.
Commit https://github.com/plotly/plotly.js/commit/2c76d79c18b564bd0383a304c416cc527c3c0cac fixes the bug.
Other commits make helper functions to check `hovermode`.

@plotly/plotly_js

### DIFF
--- a/src/components/fx/helpers.js
+++ b/src/components/fx/helpers.js
@@ -12,13 +12,13 @@ var Lib = require('../../lib');
 
 // look for either subplot or xaxis and yaxis attributes
 // does not handle splom case
-exports.getSubplot = function getSubplot(trace) {
+exports.getSubplot = function(trace) {
     return trace.subplot || (trace.xaxis + trace.yaxis) || trace.geo;
 };
 
 // is trace in given list of subplots?
 // does handle splom case
-exports.isTraceInSubplots = function isTraceInSubplots(trace, subplots) {
+exports.isTraceInSubplots = function(trace, subplots) {
     if(trace.type === 'splom') {
         var xaxes = trace.xaxes || [];
         var yaxes = trace.yaxes || [];
@@ -36,7 +36,7 @@ exports.isTraceInSubplots = function isTraceInSubplots(trace, subplots) {
 };
 
 // convenience functions for mapping all relevant axes
-exports.flat = function flat(subplots, v) {
+exports.flat = function(subplots, v) {
     var out = new Array(subplots.length);
     for(var i = 0; i < subplots.length; i++) {
         out[i] = v;
@@ -44,7 +44,7 @@ exports.flat = function flat(subplots, v) {
     return out;
 };
 
-exports.p2c = function p2c(axArray, v) {
+exports.p2c = function(axArray, v) {
     var out = new Array(axArray.length);
     for(var i = 0; i < axArray.length; i++) {
         out[i] = axArray[i].p2c(v);
@@ -52,12 +52,12 @@ exports.p2c = function p2c(axArray, v) {
     return out;
 };
 
-exports.getDistanceFunction = function getDistanceFunction(mode, dx, dy, dxy) {
+exports.getDistanceFunction = function(mode, dx, dy, dxy) {
     if(mode === 'closest') return dxy || exports.quadrature(dx, dy);
     return mode.charAt(0) === 'x' ? dx : dy;
 };
 
-exports.getClosest = function getClosest(cd, distfn, pointData) {
+exports.getClosest = function(cd, distfn, pointData) {
     // do we already have a point number? (array mode only)
     if(pointData.index !== false) {
         if(pointData.index >= 0 && pointData.index < cd.length) {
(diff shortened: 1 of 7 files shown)


src/plots/cartesian/layout_defaults.js
```diff
@@ -249,8 +249,9 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
         handleTypeDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions);
         handleAxisDefaults(axLayoutIn, axLayoutOut, coerce, defaultOptions, layoutOut);
 
-        var unifiedHover = layoutIn.hovermode && ['x unified', 'y unified'].indexOf(layoutIn.hovermode) !== -1;
-        var unifiedSpike = unifiedHover && axLetter === layoutIn.hovermode.charAt(0);
+        var hovermode = layoutOut.hovermode || layoutIn.hovermode;
```
alexcjohnson: Does it not work with just `layoutOut.hovermode`? Do we need the `layoutIn` fallback for some specific case?

archmoj: Some tests failed without the `layoutIn` fall back.

alexcjohnson: If this is just unit tests (ie we're not running the entire `supplyDefaults` pipeline so earlier steps in that should have set `layoutOut.hovermode` were not called, then I'd rather change the test than include this fallback. If there's a case where the full pipeline does not behave correctly then this worries me, as that case presumably would not work with templates.

archmoj: `hovermode` coerce with auto defaults depending on `cartesian` is
https://github.com/plotly/plotly.js/blob/6de4c751a3d702ea13f20214349460840cfab6af/src/components/fx/layout_defaults.js#L36-L44
called at the end of pipeline here:
https://github.com/plotly/plotly.js/blob/6de4c751a3d702ea13f20214349460840cfab6af/src/plots/plots.js#L1672-L1678
I am not sure how to proceed if we want to remove `layoutIn.hovermode` fallback.

archmoj: On another note - here is test which is failing, if we remove `layoutIn` fallback:
https://github.com/plotly/plotly.js/blob/6de4c751a3d702ea13f20214349460840cfab6af/test/jasmine/tests/hover_label_test.js#L4037-L4052

alexcjohnson: If that test is failing without the `layoutIn` fallback, it means that spikes will not be properly enabled by `hovermode` set in the `template`, right?

So I guess the code here is called by the `basePlotModules` loop:

https://github.com/plotly/plotly.js/blob/6de4c751a3d702ea13f20214349460840cfab6af/src/plots/plots.js#L1639-L1647

I think the easiest thing to do would be to move `hovermode` coercion here, right at the top of this function. That will involve making a special `coerce` function using `fx/layout_attributes`, but it will ensure the ordering is correct.

archmoj: @alexcjohnson Good call. Revised in the commits below.

src/components/fx/helpers.js
```diff
@@ -12,13 +12,13 @@ var Lib = require('../../lib');
 
 // look for either subplot or xaxis and yaxis attributes
 // does not handle splom case
-exports.getSubplot = function getSubplot(trace) {
```
alexcjohnson: nonblocking, but at one point these names were useful in stack traces. That may or may not not be the case anymore, based on changes to our build and sourcemap processes... but historically that was the reason for names like this that are otherwise unused.

src/components/fx/layout_defaults.js
```diff
@@ -9,40 +9,24 @@
...
-        } else {
-            // flag for 'horizontal' plots:
-            // determines the state of the mode bar 'compare' hovermode button
-            layoutOut._isHoriz = isHoriz(fullData, layoutOut);
-            hovermodeDflt = layoutOut._isHoriz ? 'y' : 'x';
-        }
-    } else hovermodeDflt = 'closest';
-
-    var hoverMode = coerce('hovermode', hovermodeDflt);
+    var hoverMode = handleHoverModeDefaults(layoutIn, layoutOut, fullData);
```
alexcjohnson: We talked about short-circuiting this one, right? something like:
```js
var hoverMode = layoutOut.hovermode;
if (hoverMode === undefined) {
    hoverMode = handleHoverModeDefaults(layoutIn, layoutOut, fullData);
}
```
Is that a possibility? Or would that not work for some reason?

archmoj: `hovermode` and `clickmode` use special `coerce` function which returns previously coerced value.
https://github.com/plotly/plotly.js/blob/abf3a557080ae16b25ca0da57d88ba956a535b75/src/components/fx/hovermode_defaults.js#L14-L20

PR #4670, opened 2020-03-22
Set Cartesian axis attributes via template

archmoj:
Commit https://github.com/plotly/plotly.js/pull/4670/commits/5f8b280fc78b0dd8ad88440a317cb0d35811bb24 addresses #4667 i.e. to set axis `type`, `rangebreaks` and `tickformatstops` from 
`template`.

@plotly/plotly_js

### DIFF
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -126,14 +126,15 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce, 
     }
 
     if(axType === 'date') {
-        var rangebreaks = containerIn.rangebreaks || axTemplate.rangebreaks;
+        handleArrayContainerDefaults(containerIn, containerOut, {
+            name: 'rangebreaks',
+            inclusionAttr: 'enabled',
+            handleItemDefaults: rangebreaksDefaults
+        });
 
-        if(Array.isArray(rangebreaks) && rangebreaks.length) {
-            handleArrayContainerDefaults(containerIn, containerOut, {
-                name: 'rangebreaks',
-                inclusionAttr: 'enabled',
-                handleItemDefaults: rangebreaksDefaults
-            });
+        if(!containerOut.rangebreaks.length) {
+            delete containerOut.rangebreaks;
+        } else {
             setConvert(containerOut, layoutOut);
 
             if(layoutOut._has('scattergl') || layoutOut._has('splom')) {
--- a/src/plots/cartesian/axis_defaults.js
+++ b/src/plots/cartesian/axis_defaults.js
@@ -105,10 +105,10 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce, 
 
     if(options.automargin) coerce('automargin');
 
-    var isMultiCategory = containerOut.type === 'multicategory';
+    var isMultiCategory = axType === 'multicategory';
 
     if(!options.noTickson &&
-        (containerOut.type === 'category' || isMultiCategory) &&
+        (axType === 'category' || isMultiCategory) &&
         (containerOut.ticks || containerOut.showgrid)
     ) {
         var ticksonDflt;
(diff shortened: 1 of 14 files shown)


nicolaskruchten:
I think we should leave #4671 alone for now please... this code already works as desired for e.g. annotations. Check out this Python example: https://plot.ly/python/templates/#including-tuple-elements-in-a-theme

archmoj:
In JS, it appears those were not working [before](https://codepen.io/MojtabaSamimi/pen/xxGaMbv?editors=0010) | [after](https://codepen.io/MojtabaSamimi/pen/qBdMgZE?editors=1000).

@nicolaskruchten @alexcjohnson am I missing something?

nicolaskruchten:
This works for me with the production bundle: https://codepen.io/nicolaskruchten/pen/wvaENrm?editors=0010

nicolaskruchten:
@alexcjohnson I'll let you judge the safety of these changes, keeping in mind that I would rather keep existing bugs in the interests of getting 1.53 and 1.54 done ASAP and not introducing subtle new bugs. For 1.53 we just need `rangebreaks` and `hovermode = "* unified"` to work with templates IMO.

alexcjohnson:
@archmoj the way templates interact with container arrays is a bit subtle as it contains features for both defaults for all array items and overrides for specific items (which may not be at the same index in the array), but in general it works. Take a look at the original templates PR https://github.com/plotly/plotly.js/pull/2761, esp. [this test](https://github.com/plotly/plotly.js/pull/2761/files#diff-10fdc9e067ffc2e976a05e6d98cebbf1R57) and the [`arrayTemplater` function](https://github.com/plotly/plotly.js/pull/2761/files#diff-44b7021288b6b3409a7d11ef712b75eeR144)

Please use those for context before deciding what's a bug.

nicolaskruchten:
Ah, with respect to your before/after: you need to set `name` for that to work... https://plot.ly/javascript/reference/#layout-annotations-items-annotation-name

src/plots/cartesian/layout_defaults.js
```diff
@@ -230,7 +230,13 @@ module.exports = function supplyLayoutDefaults(layoutIn, layoutOut, fullData) {
                 axHasImage[axName]
               ));
 
+        var axTemplate;
+        if(Lib.isPlainObject(layoutOut._template)) {
+            axTemplate = layoutOut._template[axLayoutOut._name];
```
alexcjohnson: This isn't quite right - `Template.newContainer` also lets `xaxis` from the template be used for `xaxis2`, xaxis3` etc in the actual plot. But do you need `axTemplate` at all? Can't you just reference `containerOut._template` where you currently reference `options.axTemplate`?

archmoj: @alexcjohnson Good call. Fixed in https://github.com/plotly/plotly.js/pull/4670/commits/2494c4e1ad8718bcd4e23d559e3caa0509162a44.

src/plots/cartesian/axis_defaults.js
```diff
@@ -124,8 +125,9 @@ module.exports = function handleAxisDefaults(containerIn, containerOut, coerce,
         }
     }
 
-    if(containerOut.type === 'date') {
-        var rangebreaks = containerIn.rangebreaks;
+    if(axType === 'date') {
+        var rangebreaks = containerIn.rangebreaks || axTemplate.rangebreaks;
```
alexcjohnson: This isn't quite right... `[]` is truthy, also `axTemplate.rangeBreaks` items will extend the array if it contains named items.

Can we instead do:

```js
handleArrayContainerDefaults(...);

if (containerOut.rangebreaks.length) {
   // setConvert and gl / splom check
}
else {
   delete containerOut.rangebreaks;
}
```

Same for `tickformatstops`

archmoj: Good idea. Done in https://github.com/plotly/plotly.js/pull/4670/commits/f824902d0c044d3fcd7342b45efd20461248f6d1.

PR #4677, opened 2020-03-23
Improve auto ticks over axes with rangebreaks

### LINKED ISSUE #4666: gridlines, ticks and ticklabels within rangebreaks
If one or more gridlines/ticks/ticklabels would have been drawn within a rangebreak, we should draw the first one at the break.

Possibly involves https://github.com/plotly/plotly.js/pull/4614/files#diff-f6d5ed25a5d6f57c70063e1e1bbc407dR616

archmoj:
Follow up of #4614, 
This PR improves ticks over axes with `rangebreaks` and fixes #4666.
Demos: [Before](https://codepen.io/MojtabaSamimi/pen/vYOVEOP?editors=0010) vs [After](https://codepen.io/MojtabaSamimi/pen/bGdmNVV?editors=1000).

@plotly/plotly_js


src/plots/cartesian/axes.js
```diff
@@ -611,31 +611,43 @@ axes.calcTicks = function calcTicks(ax) {
...
-        // so that some hopefully fall between rangebreaks
-        if(ax.tickmode === 'auto' && tickVals.length < nTicksBefore / 6) {
-            axes.autoTicks(ax, ax._roughDTick / 3);
-            autoTickRound(ax);
-            ax._tmin = axes.tickFirst(ax);
-            generateTicks();
-            tickVals = tickVals.filter(function(d) {
-                return ax.maskBreaks(d.value) !== BADNUM;
-            });
+        // add tick at the start of every rangebreak
```
alexcjohnson: That's too much - we should only add ticks to breaks that would get a tick anyway.

Compare this screenshot from "before"
Screen Shot 2020-03-23 at 6 09 03 PM

With a similar range "after"
Screen Shot 2020-03-23 at 6 09 57 PM

That actually tried to make even more ticks than you see, but some later step removed lots of them because it couldn't draw them without overlapping.

I'm thinking the above `tickVals = tickVals.filter` needs to somehow turn into:
- Find all the ticks within the SAME break - note that this isn't necessarily "all the consecutive ticks with `maskBreaks === BADNUM`, because you could have two consecutive ticks in different breaks. For example, month ticks for February and March, in years when both of those start on a weekend (because February is exactly 4 weeks long). I don't know how to do that, we may need a new helper function that's effectively inverse of `maskBreaks`, so if you're in a break it returns the beginning (or end?) of that break.
- Collapse those all to a single tick, positioned at the break, but labeled with its original value.
- Previously I think we discussed using the first value, but now I'm thinking the value of the last tick would make more sense, particularly for date axes where currently the ticks essentially label the time period that follows them. For example, if you remove nights AND weekends but zoom to where ticks are daily, so that Saturday, Sunday, and Monday ticks are within the same break, you'd rather see the re-inserted tick say Monday rather than Saturday.

archmoj: @alexcjohnson thanks for the feedback.
I started some adjustments to use end of the break and reduce the effective length of axis by the total length of the breaks in  https://github.com/plotly/plotly.js/pull/4677/commits/7b627e53fbf10c25180a49df03dfb0ae3ef38029.
I am working on another commit to possibly improve this.

src/plots/cartesian/axes.js
```diff
@@ -611,31 +613,50 @@ axes.calcTicks = function calcTicks(ax) {
...
+                    }
+                }
+            }
+        }
 
-        // remove ticks falling inside rangebreaks
-        tickVals = tickVals.filter(function(d) {
-            return ax.maskBreaks(d.value) !== BADNUM;
+        // sort
+        tickVals = tickVals.sort(function(a, b) {
```
alexcjohnson: Why do we need to sort? We didn't before, did we?

src/plots/cartesian/axes.js
```diff
@@ -611,31 +613,50 @@ axes.calcTicks = function calcTicks(ax) {
...
+                var value = tickVals[t].value;
+                if(ax.maskBreaks(value) === BADNUM) {
+                    // find which break we are in
+                    for(var k = 0; k < ax._rangebreaks.length; k++) {
+                        var brk = ax._rangebreaks[k];
+                        if(value >= brk.min && value < brk.max) {
+                            // replace with break end
+                            tickVals[t] = {
+                                minor: false,
+                                value: brk.max
```
alexcjohnson: I suspect just replacing the value with `brk.max` is going to get us in trouble at some point, ie when the end of the break would get a label that's different from the position within the break where we actually wanted the tick to be. Often the tick label formatting is sufficiently rounded that it will be correct, but this need not always be the case.

Can we
- keep track of the actual value that was requested in addition - ie do something like `tickVals[t] = {minor: false, value: brk.max, requestedValue: value}`
- when we reduce to one tick within the break, keep only the one with the largest requested value
- make sure that even if it's `brk.max` that's used to *position* the tick, we use `requestedValue` to generate its label.

src/plots/cartesian/axes.js
archmoj: Good eye. It was needed with previous commits.
Now it is removed by https://github.com/plotly/plotly.js/pull/4677/commits/ea4d8a2bc5160577132212819b1fb554bf006a67.

src/plots/cartesian/axes.js
```diff
@@ -655,14 +676,25 @@ axes.calcTicks = function calcTicks(ax) {
...
+        var label = axes.tickText(
             ax,
             tickVals[i].value,
             false, // hover
             tickVals[i].minor // noSuffixPrefix
         );
+
+        if(ax.rangebreaks) { // this might be useful in general - but applying it only to rangebreaks for now
+            if(label.text !== prevLabel) {
+                ticksOut.push(label);
```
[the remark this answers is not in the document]
archmoj: Also no need for those changes anymore.
Reverting the changes in this part by ea4d8a2.

src/plots/cartesian/axes.js
archmoj: @alexcjohnson 
Good call. Done in https://github.com/plotly/plotly.js/pull/4677/commits/7a1c649bfc71b55f4ca22be37ec44067bd19c083, https://github.com/plotly/plotly.js/pull/4677/commits/d1ba367c00743c740db4116ab29c0ba62601ffff and https://github.com/plotly/plotly.js/pull/4677/commits/e718e3e05d382b0736014638ca31bdc8aa9eaf23.

alexcjohnson:
@archmoj huge improvement! I'm just seeing one problem in a few of the mocks: the second line (date part) of the tick label is missing on all but the first label; it should be present whenever the date changes. Somewhere we have code that compares adjacent tick labels and removes the second line when it matches the previous tick label, that's perhaps running too early, before the ticks and their text are fully defined?

Screen Shot 2020-03-24 at 7 56 56 PM

axes_breaks-night_autorange-reversed is in this screenshot, axes_breaks-weekends-weeknights also shows this.

Interestingly, axes_breaks-weekends_autorange-reversed shows a change but I think I prefer the new behavior: the "first" tick of each new day is not the left-most or bottom-most, it's the earliest date. I could see arguments either way, but I think earliest date is better, as it labels the day (or month, or year, whatever) on the closest tick to when that period starts.

PR #4687, opened 2020-03-25
unified hover: honor layout.hoverlabel

### LINKED ISSUE #4683: hoverlabel formatting in unified mode
The intention for unified hoverlabel formatting is that if `layout.hoverlabel` isn't specified, it will use the settings from `layout.legend` for e.g. `bgcolor` and `font` etc, otherwise it will use the `layout.hoverlabel` settings.

Right now it seems like neither work to override the hoverlabel formatting in `x unified` mode

### LINKED ISSUE #4693: rangebreaks hovermode=closest offset
Try hovering over points here: https://codepen.io/nicolaskruchten/pen/oNXQmLR?editors=1010

antoinerg:
Fixes #4683

TODO
- [x] honor `bordercolor` (https://github.com/plotly/plotly.js/pull/4687/commits/531fe79de9f74bdc4394a38d2fc4b07d807657bf)
- [x] make sure `hoverlabel.bgcolor` is opaque by default (https://github.com/plotly/plotly.js/commit/cc8ca1c3071565a5ad2b3fda1907ebd170b1419b)

### DIFF
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -992,6 +992,7 @@ function createHoverText(hoverData, opts, gd) {
                 title: {text: t0, font: fullLayout.hoverlabel.font},
                 font: fullLayout.hoverlabel.font,
                 bgcolor: fullLayout.hoverlabel.bgcolor || fullLayout.paper_bgcolor,
+                bordercolor: fullLayout.hoverlabel.bordercolor,
                 borderwidth: 1,
                 tracegroupgap: 7,
                 traceorder: fullLayout.legend ? fullLayout.legend.traceorder : undefined,
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -989,8 +989,8 @@ function createHoverText(hoverData, opts, gd) {
         var mockLayoutIn = {
             showlegend: true,
             legend: {
-                title: {text: t0, font: fullLayout.font},
-                font: fullLayout.font,
+                title: {text: t0, font: fullLayout.hoverlabel.font},
+                font: fullLayout.hoverlabel.font,
                 bgcolor: fullLayout.hoverlabel.bgcolor,
                 borderwidth: 1,
                 tracegroupgap: 7,
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -991,7 +991,7 @@ function createHoverText(hoverData, opts, gd) {
             legend: {
                 title: {text: t0, font: fullLayout.hoverlabel.font},
                 font: fullLayout.hoverlabel.font,
-                bgcolor: fullLayout.hoverlabel.bgcolor,
+                bgcolor: fullLayout.hoverlabel.bgcolor || fullLayout.paper_bgcolor,
                 borderwidth: 1,
                 tracegroupgap: 7,
                 traceorder: fullLayout.legend ? fullLayout.legend.traceorder : undefined,
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -991,7 +991,7 @@ function createHoverText(hoverData, opts, gd) {
             legend: {
                 title: {text: t0, font: fullLayout.font},
                 font: fullLayout.font,
-                bgcolor: fullLayout.paper_bgcolor,
+                bgcolor: fullLayout.hoverlabel.bgcolor,
                 borderwidth: 1,
                 tracegroupgap: 7,
                 traceorder: fullLayout.legend ? fullLayout.legend.traceorder : undefined,
--- a/src/components/fx/hover.js
+++ b/src/components/fx/hover.js
@@ -991,7 +991,7 @@ function createHoverText(hoverData, opts, gd) {
             legend: {
                 title: {text: t0, font: fullLayout.hoverlabel.font},
                 font: fullLayout.hoverlabel.font,
-                bgcolor: fullLayout.hoverlabel.bgcolor || fullLayout.paper_bgcolor,
+                bgcolor: fullLayout.hoverlabel.bgcolor,
                 bordercolor: fullLayout.hoverlabel.bordercolor,
                 borderwidth: 1,
                 tracegroupgap: 7,
--- a/src/components/fx/hoverlabel_defaults.js
+++ b/src/components/fx/hoverlabel_defaults.js
@@ -14,17 +14,25 @@ var isUnifiedHover = require('./helpers').isUnifiedHover;
 module.exports = function handleHoverLabelDefaults(contIn, contOut, coerce, opts) {
     opts = opts || {};
 
+    function inheritFontAttr(attr) {
+        if(!opts.font[attr]) {
+            if(contIn.legend && contIn.legend.font && contIn.legend.font[attr]) {
+                opts.font[attr] = contIn.legend.font[attr];
+            } else if(contIn.font && contIn.font[attr]) {
+                opts.font[attr] = contIn.font[attr];
+            }
+        }
+    }
+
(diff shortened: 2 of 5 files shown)


antoinerg:
With the changes made in this PR, a bunch of mocks are now broken in unified hover modes because they contain this in their layout:
```json
        "legend": {
            "font": {
                "family": "", 
                "size": 0, 
                "color": ""
            }, 
```
This is all over our mocks and I wonder why that is the case?

@alexcjohnson

alexcjohnson:
Ah yeah, that's in a lot of really old mocks... my recollection is these came from the predecessor to chart studio, which at one point had a bug that it was saving effectively `fullData` and `fullLayout` instead of `data` and `layout`. Nobody would create things like that today, and aside from unified hover they have no effect, right? So I'd say we can safely remove them from these mocks.

antoinerg:
I don't particularly like the changes I made in `hoverlabel_defaults.js` and I wonder whether we could solve the problem differently (we could at least make the code nicer). At the very least, the new tests show that the behavior requested by @nicolaskruchten now works.

I must say however that inheriting the `bgcolor` from the legend will look odd in many mocks for which the legend background is transparent.

nicolaskruchten:
We explicitly force the legend bgcolor to transparent in some mocks?

antoinerg:
> We explicitly force the legend bgcolor to transparent in some mocks?

We do. For example: https://github.com/plotly/plotly.js/blob/60425d2d3c8f00f20e786bae906e22c9f1b236ca/test/image/mocks/11.json#L240

src/components/fx/hover.js
```diff
@@ -989,9 +989,10 @@ function createHoverText(hoverData, opts, gd) {
         var mockLayoutIn = {
             showlegend: true,
             legend: {
-                title: {text: t0, font: fullLayout.font},
-                font: fullLayout.font,
-                bgcolor: fullLayout.paper_bgcolor,
+                title: {text: t0, font: fullLayout.hoverlabel.font},
+                font: fullLayout.hoverlabel.font,
+                bgcolor: fullLayout.hoverlabel.bgcolor || fullLayout.paper_bgcolor,
```
alexcjohnson: why do we need the `paper_bgcolor` fallback here?

src/components/fx/hoverlabel_defaults.js
```diff
@@ -9,10 +9,32 @@
...
 
 var Lib = require('../../lib');
+var isUnifiedHover = require('./helpers').isUnifiedHover;
 
 module.exports = function handleHoverLabelDefaults(contIn, contOut, coerce, opts) {
     opts = opts || {};
 
+    function inheritFontAttr(attr) {
+        if(!opts.font[attr]) {
+            if(contIn.legend && contIn.legend.font && contIn.legend.font[attr]) {
```
alexcjohnson: is `contOut.legend` not available yet? Could we rearrange so it *is* available? As it is this won't work with templates.

alexcjohnson:
Re: transparent legend bgcolor - there are two interesting cases: legend over the margin, and legend over the plot. Both of these will happen from time to time, but I'd say we should fall back on `plot_bgcolor` because we know that trace symbols will be visible against that. Specifically, what if we do `color.combine(legend.bgcolor, layout.plot_bgcolor)`, so we always get 100% opaque?

src/components/fx/hover.js
antoinerg: If `hoverlabel.bgcolor` is undefined, we need a fallback. We can either specify it here or specify `paper_bgcolor` in `mockLayoutOut` in order for `legendSupplyDefaults` to use it as a fallback. Let me know what you prefer!

src/components/fx/hoverlabel_defaults.js
antoinerg: > is `contOut.legend` not available yet? Could we rearrange so it _is_ available?

It's not available yet. I'm not sure how difficult it would be to rearrange so it is. I'm not super familiar with the `supplyDefaults` routine but it does look like the legend (which depend on data) is coerced in the final steps:
https://github.com/plotly/plotly.js/blob/7c67896603d6c61d515a5fb88d653580bd4e748a/src/plots/plots.js#L440-L441

src/components/fx/hover.js
archmoj: Do we have a test for this?
Should we update the descriptions?

antoinerg: Yes, the background color is well tested to make sure that the color is set by layout attribute `hoverlabel.bgcolor` > `legend.bgcolor` > `paper_bgcolor`.

We should definitely update the description.

src/components/fx/hoverlabel_defaults.js
antoinerg: Done in https://github.com/plotly/plotly.js/pull/4687/commits/dfd73b86ab1ebcf8448245583035396a43c71b07 which results in simpler code (in particular I don't have to check for the existence of nested keys in `legend` object: if it's been coerced, everything is there)!

antoinerg:
> Re: transparent legend bgcolor - there are two interesting cases: legend over the margin, and legend over the plot. Both of these will happen from time to time, but I'd say we should fall back on `plot_bgcolor` because we know that trace symbols will be visible against that. Specifically, what if we do `color.combine(legend.bgcolor, layout.plot_bgcolor)`, so we always get 100% opaque?

I think we should get something always opaque by default and I like the idea of using `color.combine`. However, I wonder if we should use `layout.paper_bgcolor` instead of `layout.plot_bgcolor`.

Using `plot_bgcolor` leads to this result in mock `11`:
Screenshot_2020-03-26_12-52-41

cc @alexcjohnson @nicolaskruchten

alexcjohnson:
> I wonder if we should use `layout.paper_bgcolor` instead of `layout.plot_bgcolor`.

Yes, looking at it again I think you're right. Also because the default for `legend.bgcolor` is `layout.paper_bgcolor`. Either way there will be unpleasant cases but `paper_bgcolor` probably has fewer of them.

antoinerg:
> > I wonder if we should use `layout.paper_bgcolor` instead of `layout.plot_bgcolor`.
> 
> Yes, looking at it again I think you're right. Also because the default for `legend.bgcolor` is `layout.paper_bgcolor`. Either way there will be unpleasant cases but `paper_bgcolor` probably has fewer of them.

Ok, I went with `paper_bgcolor` in https://github.com/plotly/plotly.js/pull/4687/commits/cc8ca1c3071565a5ad2b3fda1907ebd170b1419b. Thanks for letting me know about `color.combine`! It's a nice addition.

src/components/fx/hover.js
alexcjohnson: Just closing the loop based on our off-GH conversation: `layout.hoverlabel.bgcolor` is empty by default, so that each hoverlabel knows to take its color from its trace. It only has a value if you explicitly set one, that will become the common hover label color for all traces.

src/core.js
```diff
@@ -42,8 +42,8 @@ register(require('./traces/scatter'));
 
 // register all registrable components modules
 register([
-    require('./components/fx'),
     require('./components/legend'),
+    require('./components/fx'),
```
alexcjohnson: Heh that was easy 😅 A little fragile though. Can we add a comment at least, saying that fx needs to come after legend for `supplyDefaults` to work correctly? I wonder if there are any other couplings like this?

antoinerg: Done in https://github.com/plotly/plotly.js/pull/4687/commits/0d406a02fd927cbba70cd3b9a5be58b265cae3cc

It's a bit fragile although the `hoverlable_test` would fail if the order was changed!

PR #4690, opened 2020-03-26
Disable plotly server URL by default

### LINKED ISSUE #4689: Change default value of config.plotlyServerURL to ""
And update the schema

archmoj:
Resolves #4689 
This PR changes the default value for `config.plotlyServerURL` option to empty string in https://github.com/plotly/plotly.js/commit/514aef4b444e12cb8cf80c2b0407e2b4802f4558.
Please also note that empty string would be ignored in `Plotly.sendDataToCloud` in https://github.com/plotly/plotly.js/commit/f6611843ecda399cca16f8a81da5d01f72070070,
Therefore in order to `Edit in Chart Studio` (using `modebar` buttons `editInChartStudio` or `sendDataToCloud`) the `config.plotlyServerURL` is required.

@plotly/plotly_js

### DIFF
--- a/src/plot_api/plot_config.js
+++ b/src/plot_api/plot_config.js
@@ -34,10 +34,10 @@ var configAttributes = {
         valType: 'string',
         dflt: '',
         description: [
-            'When set it determines base URL e.g. \'https://plotly.com\' for',
-            'the \'Edit in Chart Studio\' (aka sendDataToCloud) mode bar button',
+            'When set it determines base URL for',
+            'the \'Edit in Chart Studio\' `showEditInChartStudio`/`showSendToCloud` mode bar button',
             'and the showLink/sendData on-graph link.',
-            'To enable sending your data to Plotly\'s public cloud, you need to',
+            'To enable sending your data to Chart Studio Cloud, you need to',
             'set both `plotlyServerURL` to \'https://chart-studio.plotly.com\' and',
             'also set `showSendToCloud` to true.'
         ].join(' ')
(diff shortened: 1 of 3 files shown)


src/plot_api/plot_config.js
```diff
@@ -32,10 +32,11 @@ var configAttributes = {
...
     plotlyServerURL: {
         valType: 'string',
-        dflt: 'https://plot.ly',
+        dflt: '',
         description: [
-            'Sets base URL for the \'Edit in Chart Studio\' (aka sendDataToCloud) mode bar button',
-            'and the showLink/sendData on-graph link'
+            'When set it determines base URL for',
+            'the \'Edit in Chart Studio\' (aka sendDataToCloud) mode bar button',
+            'and the showLink/sendData on-graph link.'
```
alexcjohnson: Can we at least include the url for CSC in the description? I guess that should be `https://chart-studio.plotly.com`? Something like "To enable sending your data to Plotly's public cloud, you need to set both plotlyServerURL to 'https://chart-studio.plotly.com' and showSendToCloud to true"

archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4690/commits/7d5492d07667d163b9cd31a850b97ce61119565b using `https://plotly.com`

archmoj: Extra info added in https://github.com/plotly/plotly.js/pull/4690/commits/9959cb89065cee2c7bcfb07a48372babc6a591a5.

alexcjohnson: Thanks - I'd take `https://plotly.com` out, since I believe that's not going to work, and just include `https://chart-studio.plotly.com`

nicolaskruchten: Let's not use `https://plotly.com` as this will confuse people, let's only use `https://chart-studio.plotly.com` as the example.

Let's also not call it "Plotly's public cloud" but rather "Chart Studio Cloud" please.

Finally, let's replace `sendDataToCloud` (which is not one of the options!) with `showEditInChartStudio/showSendToCloud` which are slight variants on each other...

nicolaskruchten:
> Therefore in order to Edit in Chart Studio the config.plotlyServerURL is required.

Yes please.

src/plot_api/plot_config.js
archmoj: Revised in https://github.com/plotly/plotly.js/pull/4690/commits/d81e59829a66796eceefd63153ba0c428255bc74.

PR #4695, opened 2020-03-27
Avoid tight daily labels on axes with rangebreaks when using "day of week" pattern

archmoj:
[Demo](https://codepen.io/MojtabaSamimi/pen/KKpbVGg?editors=0010)

Follow up of #4614 and #4688:
Commit https://github.com/plotly/plotly.js/commit/c340efc1cbd23ab3a52d8962ea5acc50ed89357e exercised 2 -> 7 rounding which leads to following commits.

As a result we only round `dtick` to 1 if it is really close to 1. 

@plotly/plotly_js


nicolaskruchten:
Hmm yeah that mock does seem sparse. But I think erring on the side of sparseness is better than sometimes having unreadably-dense plots like this one:

image

archmoj:
> Hmm yeah that mock does seem sparse. But I think erring on the side of sparseness is better than sometimes having unreadably-dense plots like this one:
> 
> image

@nicolaskruchten thanks very much for posting the image related to the ( old ) undesirable behaviour.

src/plots/cartesian/axes.js
```diff
@@ -806,11 +806,11 @@ axes.autoTicks = function(ax, roughDTick) {
             roughDTick /= ONEAVGMONTH;
             ax.dtick = 'M' + roundDTick(roughDTick, 1, roundBase24);
         } else if(roughX2 > ONEDAY) {
-            ax.dtick = roundDTick(roughDTick, ONEDAY, roundDays);
-
             if(ax._hasDayOfWeekBreaks) {
-                if(ax.dtick === 2 * ONEDAY) ax.dtick = ONEDAY;
-                else if(ax.dtick === 3 * ONEDAY) ax.dtick = 7 * ONEDAY;
+                ax.dtick = roundDTick(roughDTick, ONEDAY, [1, 1.001, 7, 14]);
+                if(ax.dtick !== ONEDAY) ax.dtick = 7 * ONEDAY;
```
alexcjohnson: Is this different from just saying `roundDTick(roughDTick, ONEDAY, [1, 7, 14])`? The underlying function is `Lib.roundUp` so I'd expect anything too big for 1 would already bump up to 7. Also I think this would convert 2-week ticks into 1-week?

archmoj: > Also I think this would convert 2-week ticks into 1-week?

Doog eye. Fixed in https://github.com/plotly/plotly.js/pull/4695/commits/79b6d95b73d727289b456616baea39db458b495c.

But we still need to fix something possibly in `Lib.dateTick0` to compute `dtick` on those `reversed` cases.

alexcjohnson: again though, is this different from just `roundDTick(roughDTick, ONEDAY, [1, 7, 14])`?

archmoj: I think that is different. The issue was almost related to numbers close to 1 (e.g. 2) not those close to 7. Anyway it seems there is no more need for that hacy roundingSet as the logic is changed in the commits below.

archmoj:
@nicolaskruchten would you mind testing the updated [demo](https://codepen.io/MojtabaSamimi/pen/KKpbVGg?editors=0010) in the PR description.
Thanks!

PR #4696, opened 2020-03-27
Do not show tick labels falling inside rangebreaks

### LINKED ISSUE #4692: rangebreaks sometimes draws the gridline on the wrong point
See Jan 3 here: https://codepen.io/nicolaskruchten/pen/oNXQmLR?editors=1010

archmoj:
Follow up of #4614 and #4684,
this PR fixes #4692 | demo: [before](https://codepen.io/MojtabaSamimi/pen/GRJPrmM?editors=1010) vs [after](https://codepen.io/MojtabaSamimi/pen/QWbzdGV?editors=1010).

@plotly/plotly_js


archmoj:
TODO slack message from @alexcjohnson 

<i>"The thing to look out for in testing is whether, due to the rounding of the displayed value, we can get to a situation that the label shown actually corresponds to a situation past the first data point beyond the break. ie the label says Tuesday’s date but there are data points for Monday afternoon to the right of it.
With any luck that just won’t happen, even if we combine day of week and hour breaks, but that’s the key thing to test.</i>

alexcjohnson:
Out of the mocks here, the only one that bothers me is axes_breaks-night_autorange-reversed.png

Screen Shot 2020-03-27 at 2 58 17 PM

The "05:59" labels are weird, why don't they say "06:00"? But everything else is great, and even this one is more correct than before, just uglier. It's an obscure enough edge case that I'd be happy merging and releasing like this, and flagging this for follow-up later.

archmoj:
:dancer: 
Merging...

PR #4699, opened 2020-03-28
Fixup rangebreaks l2p and p2l functions 

### LINKED ISSUE #4693: rangebreaks hovermode=closest offset

### LINKED ISSUE #4700: rangebreaks mapping on reversed ranges
[demo](https://codepen.io/MojtabaSamimi/pen/jOPXazX?editors=0010).

archmoj:
Following of #4614,
this PR fixes `l2p` & `p2l` functions 
- [x] correct reversed ranges fixes #4700 
see https://github.com/plotly/plotly.js/pull/4699/commits/11c8ea2b8d180115d1250c018de1c47a1c9866cf?short_path=563a470#diff-563a470dab197f9857d9545f11805d89

- [x] find correct hover indices and fixes #4693
Hover demo: [before](https://codepen.io/MojtabaSamimi/pen/ExjGbgY?editors=1000) vs [after](https://codepen.io/MojtabaSamimi/pen/jOPXaMv?editors=1000)
[Hover demo2: bar + reversed range](https://codepen.io/MojtabaSamimi/pen/rNVoRNo?editors=0010)

@plotly/plotly_js


alexcjohnson:
What's going on with autorange in axes_breaks-night_autorange-reversed.png ? Both the orange and red subplots prior to this PR have tight clipping on one side, and after this PR the tight clipping is on the other side. Neither side should have tight clipping, they should both get 5% padding. Also what's with the label `23:22` that's in the same place both before and after this PR, despite the axis range changing?

Screen Shot 2020-03-29 at 2 13 28 PM

src/plots/cartesian/set_convert.js
```diff
@@ -198,51 +198,55 @@ module.exports = function setConvert(ax, fullLayout) {
...
+                var max = brk.max;
 
-                if(pos < min) break;
-                if(pos > max) q = nextI;
+                if(signAx * pos < signAx * min) break;
+                if(signAx * pos > signAx * max) q = nextI;
                 else {
                     // when falls into break, pick 'closest' offset
-                    q = pos > (min + max) / 2 ? nextI : i;
+                    q = signAx * pos > signAx * (min + max) / 2 ? nextI : i;
```
alexcjohnson: `l2p` is a hot path, so it's worth optimizing a good deal - even if it means some duplicated code, though I'm not sure if that would help here, and even if the code needs more comments and is less self-documenting. For example:
- `signAx * pos` could be done once up front, and `var min = signAx * brk.min` etc would simplify these conditionals a bit
- `first` and `last` look unnecessary
- Flip `_m2` for y axes after using it to calculate `_B`, so we don't need the `(isY ? -1 : 1) *` down below.
- Move `var isY = axLetter === 'y'` to the outer scope, where `axLetter` was defined.

`p2l` I think is generally not as hot, though maybe for some traces it's used in hover? Not sure. (Note that `c2p` uses `l2p` and `p2c` uses `p2l`)

archmoj: Good calls. Addressed in the commits below.

archmoj:
> What's going on with autorange in axes_breaks-night_autorange-reversed.png ? Both the orange and red subplots prior to this PR have tight clipping on one side, and after this PR the tight clipping is on the other side. Neither side should have tight clipping, they should both get 5% padding. Also what's with the label `23:22` that's in the same place both before and after this PR, despite the axis range changing?
> 
> Screen Shot 2020-03-29 at 2 13 28 PM

Fixed by https://github.com/plotly/plotly.js/pull/4699/commits/749430d7be8d9b5f4f4250773025cd517457e5b1.

PR #4714, opened 2020-03-30
Skip zeroline when negative sizes are present on positive base and vice versa

### LINKED ISSUE #4701: autorange should not display zero line for certain bars with positive base and negative value
See for example https://codepen.io/emmanuelle-plotly/pen/bGdOLmR?editable=true, the axis starts at 0 and therefore a large part of the domain is not used to represent data. For other sets of data the automatic range works well. 

Triggered by https://github.com/plotly/plotly.py/issues/2327

archmoj:
Fixes #4701 for `bar` and `waterfall`.

@plotly/plotly_js


src/traces/bar/cross_trace_calc.js
```diff
@@ -512,7 +512,7 @@ function setBaseAndTop(sa, sieve) {
             pts.push(top);
             if(bar.hasB) pts.push(base);
 
-            if(!bar.hasB || !(bar.b > 0 && bar.s > 0)) {
+            if(!bar.hasB || !(bar.b > 0 && bar.b + bar.s > 0)) {
```
alexcjohnson: Per https://github.com/plotly/plotly.js/issues/4701#issuecomment-606110937 I'm really curious if there's anything wrong with just:
`if (!bar.hasB || !bar.b)`
or maybe even
`if (!bar.b)`
depending on what these values are.

If you negate all base and size values - so all top and bottom values are <0 - we should also not draw the axis to zero. And if there are already values on both sides of zero, the `tozero: true` flag is irrelevant.

Also: there's similar logic down in `normalizeBars` ~line 700 that probably needs updating as well.

archmoj: For now added logic & tests to handle negative case similar to the positive.

archmoj: > Per [#4701 (comment)](https://github.com/plotly/plotly.js/issues/4701#issuecomment-606110937) I'm really curious if there's anything wrong with just:
> `if (!bar.hasB || !bar.b)`
> or maybe even
> `if (!bar.b)`
> depending on what these values are.
> 
That would change `bar_attrs_overlay` & `bar_attrs_group` baselines.

PR #4725, opened 2020-04-02
Fix "extremes" opacityscale function

### LINKED ISSUE #4723: The opacityscale, 'extremes', has constant opacity equal to 1
The function `createWave()` [https://github.com/plotly/plotly.js/blob/d37895bd64227433229f2b3f91c9e14fd6ec7b89/src/traces/surface/defaults.js#L19](https://github.com/plotly/plotly.js/blob/d37895bd64227433229f2b3f91c9e14fd6ec7b89/src/traces/surface/defaults.js#L19) generates an opacityscale with all opacities equal to 1, because:
```
 Math.max(1, Math.min(0, v)) // line 27
```
is 1 for any v.

Replacing it by:
```
 Math.min(1, Math.max(0, v)) 
```
removes the actual drawback, and leads to a non-constant opacityscale.

archmoj:
Fixes #4723 for `volume` & `surface` 
thanks to investigation by @empet.

@plotly/plotly_js


antoinerg:
LGTM :dancer:

PR #4734, opened 2020-04-06
Relocate first tick when falls inside a rangebreak

### LINKED ISSUE #4722: Tick labels when using hourly rangebreaks
This pen https://codepen.io/nicolaskruchten/pen/qBdwQKw?editors=0010 shows oddly-spaced tick labels both by default, and when setting `dtick` to `24*60*60*1000` (daily) or `60*60*1000` (hourly).

archmoj:
Fixes #4722,
this PR improves tick positioning by moving the first tick when it falls inside a `rangebreak`.

Demos: [before](https://codepen.io/MojtabaSamimi/pen/KKpOzVG?editors=0010) vs [after](https://codepen.io/MojtabaSamimi/pen/gOpVrgX?editors=1000).

@plotly/plotly_js


alexcjohnson:
What's going on in test/image/baselines/axes_breaks-gridlines.png ? The ticks seem to have moved to the last day of the month, which means "Mar 2015" is actually at March 31, ie essentially where an April 1 tick should have been, so it looks like we're labeling April as March.

This seems consistent on the autorange view, but if you zoom in a bit to 1-month ticks things get worse: some ticks seem to be on the last day of the month, others on the first, so sometimes you get two identical ticks:
Screen Shot 2020-05-06 at 4 19 40 PM

I guess I'm nervous about doing this at the level of `tick0` in general - particularly if you have some funny non-periodic breaks, shifting `tick0` could result in other ticks that don't need to move at all getting moved. Seems to me the solution needs to be about moving each individual tick out of a break if it falls within one.

This PR also still has the odd spacing issue for automatically-determined `tick0` and `dtick` values. See test/image/baselines/axes_breaks-weekends-weeknights.png - to solve that we'll likely need to adjust this algorithm when there are hourly ticks > 1 h along with hour rangebreaks.
Screen Shot 2020-05-06 at 4 38 45 PM

PR #4736, opened 2020-04-06
Translation to Portuguese (Portugal)

rmmineiro:
Portuguese (Portugal) translation based on Portuguese (Brazil). Made a few changes and in other cases kept the Brazil version due to lack of context.

Thanks for your interest in plotly.js!

### Translations:

- Make a PR directly to the main repository
- Please @ mention a few other speakers of this language who can help review your translations.
- If you've omitted any keys from [dist/translation_keys.txt](https://github.com/plotly/plotly.js/blob/master/dist/translation-keys.txt) - which means they will fall back on the US English text - just make a short comment about why in the PR description: the English text works fine in your language, or you would like someone else to help translating those, or whatever the reason.
- You should only update files in `lib/locales/`, not those in `dist/`


archmoj:
Thanks @rmmineiro 
Could you please @ mention a few other speakers of this language who can help review your translations?

rmmineiro:
@archmoj from the top of my head I can think of a few, say Cristiano Ronaldo, I´m just not sure he is using github :) Now, seriously, I can try to have it checked by a 'certified' translator but anyone looking at this will need to be familiar with plotly context which does not make things easier (I am not using the complete distribution myself). So, if I don´t get back to you in a few days is because I do not have any changes to propose. Looking back at the translation I suggest one change in the key 'Zoom' to ' Ampliar'. I have doubts in a few cases such as ´Pan´ but I guess 'Mover' is not too far off. In fact, all Portuguese (Brazil) translations are mostly understandable in Portuguese (Portugal), it´s just that some expressions are not commonly used (some hurting the ear a bit, say ‘Restaurar visão‘ or ´Fechamento´). Finally, I am using all these keys without any English fall back because I have users that do not understand English and would have difficulty using the graphs otherwise.

archmoj:
@joaoalf do you know anyone that could help with reviewing this PR?

rmmineiro:
@archmoj can you point me to a turntable rotation example?

archmoj:
> @archmoj can you point me to a turntable rotation example?

Here is a [demo](https://codepen.io/MojtabaSamimi/pen/yrMxEV?editors=0010).

andrefarzat:
@alexcjohnson It seems ok for me 💃

Issue #4733, opened 2020-04-06
Write test to validate mocks


archmoj:
cc: https://github.com/plotly/plotly.js/issues/4707#issuecomment-609786084

alexcjohnson:
From #4762: before we consider this issue closed, new mocks must automatically be part of the validation test. IMO that's actually more important than making sure we validate all the existing mocks, though both parts would be useful.

> > My primary question is about the hard-coded list in `mocks_test` - seems like we should be able to
> > 
> > * read in the list of _all_ mocks from the filesystem - perhaps if this step is fast enough we can just have `karma.conf.js` do it before every test run and write the results to a file in `tests/jasmine/assets`?
> > * only list a set of mocks or patterns to ignore - and then over time we can reduce that to an empty list.
> 
> @alexcjohnson
> That is a very good idea we could try in another PR.
> At first I did try to adapt our image test for this purpose.
> But it was rather complicated to acess the system files from the jasmine test.

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for several years, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Thanks for your help - @gvwilson

Issue #4707, opened 2020-03-29
How best to interpret/parse "items"  with no "role" or "valType"


hmf:
I realize this may not be the best place to ask this. If so please redirect me to a more appropriate forum.  

My testing of a Scala wrapper failed on (among others) these mocks:

*  [gl2d_parcoords_select_first_last_enum.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/gl2d_parcoords_select_first_last_enum.json)
* [cheater_constraint_greater_than_with_hill.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/cheater_constraint_greater_than_with_hill.json)
* [mocks/box-violin-x0-category-position.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/box-violin-x0-category-position.json)
* [indicator_format_extremes.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/indicator_format_extremes.json)
* [splom_lower-nodiag-matching.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/splom_lower-nodiag-matching.json)
* [bar-alignment-offset.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/bar-alignment-offset.json)
* [cheater_constraint_outer_range_hi_top.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/cheater_constraint_outer_range_hi_top.json)
* [fonts.json](https://github.com/plotly/plotly.js/blob/master/test/image/mocks/fonts.json)

I have realized that all have a schema with the following _shape_ (example based on `gl2d_parcoords_select_first_last_enum.json`):

```json
                "dimensions": {
                    "items": {
                        "dimension": {
                            "label": {
                                "valType": "string",
                                "role": "info",
                                "editType": "plot",
                                "description": "The shown name of the dimension."
                            },
                            "tickvals": {
                                "valType": "data_array",
                                "editType": "plot",
                                "description": "Sets the values at which ticks on this axis appear.",
                                "role": "data"
                            },
                           "role": "object"
                    },
                    "role": "object"
                },
```

Looking at the JSON documents, it seems like `"items"` is a JSON array that contains objects of a specific _type_ (in the example above it would be  `"dimension"`). But the schema does not seem to make this explicit. Trying to work out how to parse this consistently. How should I go about it - can I assume that this is always so?

[...]

PR #4738, opened 2020-04-07
Deleted "steps" attribute from test/image/mocks/indicator_attrs.json …

hmf:
As discussed in issue #4707 @nicolaskruchten indicated the the mock may not be correct. I changed as per the description. @archmoj  may also be interested in checking this change (related #4733).


archmoj:
@hmf thanks very much for the PR.
The mock is corrected by #4740.

PR #4740, opened 2020-04-07
Fix steps in indicator attributes mock

archmoj:
Supersedes #4738.
The `steps` should be a list.

@plotly/plotly_js 
cc: @hmf


nicolaskruchten:
Yep! I just didn't want to make @hmf feel like he had to go through the whole "regenerate the baselines" thing :)

PR #4752, opened 2020-04-10
Bump mapbox-gl to 1.9.2

archmoj:
Maybe best to merge this in the next minor.

@plotly/plotly_js


alexcjohnson:
Yes, this definitely deserves a minor, not a patch.

Does this impact any of the mapbox baselines? Those are in the `@noCI` tests, right?

Curious whether this would have addressed https://github.com/plotly/dash-core-components/issues/756 and friends - hopefully we won't find out, and the underlying bug will never bubble up in this repo!

archmoj:
> Does this impact any of the mapbox baselines? Those are in the `@noCI` tests, right?
> 
No change came to my eyes. 
Maybe we could/should test this better.

PR #4762, opened 2020-04-14
Validate mocks - first step

archmoj:
In addition to heavy cleanup and updating (by applying some non-deprecated attributes e.g. title object), this PR adds jasmine tests for validating all current mocks using `Plotly.validate`.
The test called `mock_test` is drafted in commit 8c1a9d93f989de6d7030922487ea0a93a967a356 which currently bypasses `undefined layout`, `invisible traces`, `coloraxis`, `dynamic` e.g. `textposition` arrays.
With this relaxation,
before modifying any mocks running the test resulted in  
TOTAL: 1014 | FAILED: 487 | SUCCESS: 527
After updating the mocks the situation improved as listed below:
TOTAL: 1014 | FAILED: 174 | SUCCESS: 840.

@alexcjohnson 
cc: @plotly/plotly_js 

cc: #4733


nicolaskruchten:
@alexcjohnson let's review this once shape-drawing is done :)

alexcjohnson:
@archmoj I just had a quick look at this - very nice! Let's not touch it until 1.54.0 is released, but as this doesn't impact features at all we can include it in a patch after that.

My primary question is about the hard-coded list in `mocks_test` - seems like we should be able to
- read in the list of *all* mocks from the filesystem - perhaps if this step is fast enough we can just have `karma.conf.js` do it before every test run and write the results to a file in `tests/jasmine/assets`?
- only list a set of mocks or patterns to ignore - and then over time we can reduce that to an empty list.

archmoj:
> My primary question is about the hard-coded list in `mocks_test` - seems like we should be able to
> 
> * read in the list of _all_ mocks from the filesystem - perhaps if this step is fast enough we can just have `karma.conf.js` do it before every test run and write the results to a file in `tests/jasmine/assets`?
> * only list a set of mocks or patterns to ignore - and then over time we can reduce that to an empty list.

@alexcjohnson 
That is a very good idea we could try in another PR. 
At first I did try to adapt our image test for this purpose. 
But it was rather complicated to acess the system files from the jasmine test.
I think the way the required mocks are listed is OK atm. 
It is actually handy for fixing those many mocks.

test/jasmine/tests/mock_test.js
```diff
@@ -0,0 +1,2081 @@
...
+*/
+
+describe('@noCI mock validation', function() {
+    list.forEach(function(name) {
+        var figure = figs[name];
+        if(figure) {
+            it('validating mock: "' + name + '"', function() {
+                var out = Plotly.validate(
+                    figure.data,
+                    figure.layout || {} // shouldn't Plotly.validate handle undefined layout?
```
alexcjohnson, line 2055: Yes, I think you're right, it should. Or even undefined data, like if all you want are shapes and annotations. Feel free to make a new issue for this, or just fix it in this PR if you like.

alexcjohnson:
> That is a very good idea we could try in another PR.

🍻 fair enough - added to #4733 so we address it before considering that issue closed.

This looks great. You currently have the require statements for the mocks all commented out - can we enable the ones you've fixed already, just comment out the still-failing ones, and call this PR done?

test/jasmine/tests/mock_test.js
archmoj: Good point.
New issue: https://github.com/plotly/plotly.js/issues/4797.
Resolving...

archmoj:
> This looks great. You currently have the require statements for the mocks all commented out - can we enable the ones you've fixed already, just comment out the still-failing ones, and call this PR done?

Done in https://github.com/plotly/plotly.js/pull/4762/commits/efd85187a24461a3e94ca4b52e962118c9f03333.

PR #4775, opened 2020-04-20
Draw & adjust new shapes over image trace and cartesian subplots

### LINKED ISSUE #4374: Shape drawing
It would be interesting to port to plotly.js some shape-drawing features which exist in dash-canvas at the moment (https://dash.plot.ly/canvas). In an existing figure, the user would have the possibility to
- create new shapes such as rectangles, lines, freehand lines (paths), by selecting a specific type of drag mode in the modebar
- modify these shapes (moving them, and if possible scaling them and rotating them). Deleting the last shape would be a plus.
- retrieving the geometry of shapes would probably be done thanks to `relayoutData` events in Dash.

Also discussed: 
- having a `new_shape_style` attribute to set the style of shapes individually.
- all the drag modes would not be in the modebar by default but could be added individually

archmoj:
Resolves #4374.

As illustrated in [demo](https://codepen.io/MojtabaSamimi/pen/KKdgKZM?editors=0010), this PR enables drawing and editing shapes over `image` trace (as well as `cartesian` subplots).

New `layout` attributes are 
```
{
    `shape`: {
        `fillrule`: 'evenodd' | 'nonzero'
        `editable`: false | true
    },

    `newshape`: {
        `line`: {
            `width`:
            `dash`:
            `color`:
        },
        `fillcolor`:
        `fillrule`:  'evenodd' | 'nonzero'
        `opacity`:
        `layer`: 'above' | 'below'
        `drawdirection`: 'diagonal' | 'horizontal' | 'vertical' | 'ortho' 
    }
}
```

New modebar buttons could simply be added by passing `strings` (instead of `objects`) in `config`.
For example:
```js
modeBarButtonsToAdd: [
  'drawline',
  'drawopenpath',
  'drawclosedpath',
  'drawcircle',
  'drawrect',
  'eraseshape'
]
```

@plotly/plotly_js 
@emmanuelle


alexcjohnson:
> New `dragmode` options are `linedraw`, `rectdraw`, `ellipsedraw`, `closedfreedraw` and `openfreedraw`.

Thoughts on naming? I might vote for:
`drawline`
`drawrect`
`drawcircle` (because even though it doesn't need to be 1:1, the shape you make is `type: 'circle'`),
`drawfreeclosed`, `drawfreeopen` (or maybe `drawclosedpath` and `drawopenpath`?)

src/components/modebar/manage.js
```diff
@@ -170,6 +170,21 @@ function getButtonGroups(gd) {
         dragModeGroup.push('select2d', 'lasso2d');
     }
 
+    if(
+        // fullLayout._has('ternary') ||
+        // fullLayout._has('mapbox') ||
```
alexcjohnson, line 175: That would be nice, but clearly out of scope here 🔪

src/components/shapes/defaults.js
```diff
@@ -30,18 +30,22 @@ function handleShapeDefaults(shapeIn, shapeOut, fullLayout) {
     }
 
     var visible = coerce('visible');
-
     if(!visible) return;
 
+    var dfltType = shapeIn.path ? 'path' : 'rect';
+    var shapeType = coerce('type', dfltType);
```
alexcjohnson, line 36: hmm not new here, just rearranged, but template support... may be better to do
```js
path = coerce('path');
var dfltType = path ? 'path' : 'rect';
var shapeType = coerce('type', dfltType);
if (type !== 'path') delete shapeOut.path;
```

src/components/shapes/attributes.js
```diff
@@ -235,8 +235,30 @@ module.exports = templatedArray('shape', {
         role: 'info',
         editType: 'arraydraw',
         description: [
-            'Sets the color filling the shape\'s interior.'
+            'Sets the color filling the closed shape\'s interior.'
```
alexcjohnson, line 238:
```suggestion
            'Sets the color filling the shape\'s interior. Only applies to closed shapes.'
```

src/components/shapes/attributes.js
```diff
@@ -235,8 +235,30 @@ module.exports = templatedArray('shape', {
...
         ].join(' ')
     },
+    fillrule: {
+        valType: 'enumerated',
+        values: ['evenodd', 'nonzero'],
+        dflt: 'evenodd',
+        role: 'info',
+        editType: 'arraydraw',
+        description: [
+            'Determines the path\'s interior.',
```
alexcjohnson, line 248:
```suggestion
            'Determines which regions of complex paths constitute the interior.',
```

archmoj:
> > New `dragmode` options are `linedraw`, `rectdraw`, `ellipsedraw`, `closedfreedraw` and `openfreedraw`.
> 
> Thoughts on naming? I might vote for:
> `drawline`
> `drawrect`
> `drawcircle` (because even though it doesn't need to be 1:1, the shape you make is `type: 'circle'`),
> `drawfreeclosed`, `drawfreeopen` (or maybe `drawclosedpath` and `drawopenpath`?)

Good call. Revised in https://github.com/plotly/plotly.js/pull/4775/commits/a58539412618d497cd4d24e7f6912431f2e9328d.

src/components/shapes/defaults.js
archmoj: Good eye! Thanks for the hint.
Addressed in https://github.com/plotly/plotly.js/pull/4775/commits/092a80e5976566ab15211e14926affb9ca99033b.

src/plots/cartesian/handle_outline.js
```diff
@@ -0,0 +1,95 @@
...
+
+function activateShape(gd, path, drawShapes) {
+    var element = path.node();
+    var id = +element.getAttribute('data-index');
+
+    for(var q = 0; q < gd._fullLayout.shapes.length; q++) {
+        var shapeIn = gd._fullLayout.shapes[q]._input;
+        if(q === id && shapeIn.editable) {
+            gd._fullLayout._activeShapeIndex = q;
+            break;
```
alexcjohnson, line 23: Does this need to be a loop? Can't you just inspect `gd._fullLayout.shapes[id]`?

alexcjohnson: Also why do we need to look at `_input.editable`? Does that not make it into `_fullLayout`?

src/plots/cartesian/handle_outline.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+function deactivateShape(gd) {
+    clearOutlineControllers(gd);
+
+    var shapes = [];
+    for(var q = 0; q < gd._fullLayout.shapes.length; q++) {
+        var shapeIn = gd._fullLayout.shapes[q]._input;
+        shapes.push(shapeIn);
```
alexcjohnson, line 38: Isn't this just recreating `gd.layout.shapes`? In which case the `_guiRelayout` call would be a noop... what am I missing?

There's also a situation where `gd._fullLayout.shapes` is not 1:1 with `gd.layout.shapes` - if you have named shapes in the template, without matching items in `shapes`. I don't know if it will do everything you need, but there's a template-aware helper for editing array items that should at least help understand the scope

https://github.com/plotly/plotly.js/blob/22dd9c1d5711575ba34c3d9f42ccf8be12398350/src/plot_api/plot_template.js#L251-L252

src/components/shapes/attributes.js
archmoj: Thanks for your suggestion @alexcjohnson. 
Resolved by https://github.com/plotly/plotly.js/pull/4775/commits/838bda91346430351e0f705d1a8f452693a6f6b5.

src/components/shapes/attributes.js
archmoj: Thanks for your suggestion @alexcjohnson. 
Resolved by https://github.com/plotly/plotly.js/pull/4775/commits/838bda91346430351e0f705d1a8f452693a6f6b5.

src/plots/cartesian/handle_outline.js
archmoj: Good eye! Revised in https://github.com/plotly/plotly.js/pull/4775/commits/90a7ec7dd1cfc776b880904c7c095a3b342cd0c7.

src/plots/layout_attributes.js
```diff
@@ -444,6 +445,112 @@ module.exports = {
...
+        opacity: {
+            valType: 'number',
+            min: 0,
+            max: 1,
+            dflt: 0.5,
+            role: 'info',
+            editType: 'none',
+            description: [
+                'Sets the opacity of the active shape.',
+                'If using a value greater than half,',
```
nicolaskruchten, line 546: this doesn't seem accurate... It looks like it's more on the other side: if an **inactive** shape has an opacity greater than 0.5, then we can't create a new shape by starting a click-drag inside that shape.

archmoj: Good eye. Fixed in https://github.com/plotly/plotly.js/pull/4775/commits/cedad524b52d2b373f3d801f9ce354804f0e3c0a.

src/plots/cartesian/select.js
```diff
@@ -494,8 +603,32 @@ function coerceSelectionsCache(evt, gd, dragOptions) {
...
+
+        var fullLayout = gd._fullLayout;
+        var zoomLayer = fullLayout._zoomlayer;
+
+        var outlines = zoomLayer.selectAll('.select-outline-' + plotinfo.id);
+        if(outlines && gd._fullLayout._drawing) {
+            // add shape
+            var shapes = addNewShapes(outlines, dragOptions);
+            if(shapes) {
+                Registry.call('relayout', gd, {
```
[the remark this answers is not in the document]
archmoj: Fixed in https://github.com/plotly/plotly.js/pull/4775/commits/83b6cd241878a43415e75015551be4ef4771c30e.

src/plots/cartesian/handle_outline.js
archmoj: > Isn't this just recreating `gd.layout.shapes`? In which case the `_guiRelayout` call would be a noop... what am I missing?
> 
> There's also a situation where `gd._fullLayout.shapes` is not 1:1 with `gd.layout.shapes` - if you have named shapes in the template, without matching items in `shapes`. I don't know if it will do everything you need, but there's a template-aware helper for editing array items that should at least help understand the scope
> 
> https://github.com/plotly/plotly.js/blob/22dd9c1d5711575ba34c3d9f42ccf8be12398350/src/plot_api/plot_template.js#L251-L252

Good call. Revised not to have any `_guiRelayout`:
https://github.com/plotly/plotly.js/blob/f7643ccb923f8efcda3a16735b59a5f3a9950dc5/src/components/shapes/draw.js#L708-L735

archmoj: > There's also a situation where `gd._fullLayout.shapes` is not 1:1 with `gd.layout.shapes` - if you have named shapes in the template, without matching items in `shapes`. I don't know if it will do everything you need, but there's a template-aware helper for editing array items that should at least help understand the scope

Very good point. Fixes are made so that the shapes from `template` don't cause major problem for new shapes as well as shapes on the `layout`. 
We'll get back to improve edits using `arrayEditor` in a separate PR.

archmoj:
@alexcjohnson all the tests are now passed on CI and locally; so this should be ready for review.
Thanks!

jackparmer:
image

Amazing demo!! 200 lines of JS!

archmoj:
Tests for shape-drawing are added in  
https://github.com/plotly/plotly.js/pull/4775/commits/bea923041a658b4772947526ddb058bf5926bb88, https://github.com/plotly/plotly.js/pull/4775/commits/dffe30a9043c2455b740ff8c7f4bce25c672c973 and https://github.com/plotly/plotly.js/pull/4775/commits/3c80f7d8b941d17b4de677279db2144b66c6c639.

src/components/modebar/buttons.js
```diff
@@ -134,6 +134,58 @@ modeBarButtons.lasso2d = {
     click: handleCartesian
 };
 
+modeBarButtons.drawclosedpath = {
+    name: 'drawclosedpath',
+    title: function(gd) { return _(gd, 'Draw Closed Freeform'); },
```
alexcjohnson, line 139:
```suggestion
    title: function(gd) { return _(gd, 'Draw closed freeform'); },
```

Not a big deal, but I think a single initial cap makes more sense for all of the new buttons. That's how we capitalize all the other button titles, except `Edit in Chart Studio` (makes sense as Chart Studio is a proper name) and `Box Select` and `Lasso Select` - and I guess we should change those two at some point but unfortunately that would require editing all the translation files that already use these... so let's leave them alone for now and just change the new ones in this PR.

src/components/modebar/manage.js
```diff
@@ -170,6 +179,26 @@ function getButtonGroups(gd) {
...
+    // accept pre-defined buttons as string
+    if(Array.isArray(buttonsToAdd)) {
+        var newList = [];
+        for(var i = 0; i < buttonsToAdd.length; i++) {
+            var b = buttonsToAdd[i];
+            if(typeof b === 'string') {
+                if(DRAW_MODES.indexOf(b) !== -1) {
+                    if(
+                        // fullLayout._has('ternary') ||
+                        fullLayout._has('mapbox') ||
```
alexcjohnson, line 191: 🔪 ternary, and I didn't think we want this in mapbox either, do we? We don't support shapes on either of those subplot types, right?

src/components/shapes/attributes.js
```diff
@@ -235,8 +235,30 @@ module.exports = templatedArray('shape', {
...
+        ].join(' ')
+    },
+    editable: {
+        valType: 'boolean',
+        role: 'info',
+        dflt: false,
+        editType: 'calc+arraydraw',
+        description: [
+            'Determines whether the shape could be activated for edit or not.',
+            'Please note that setting to *false* has no effect in case `config.editable` is set to true'
```
alexcjohnson, line 259: Is it just `false` that has no effect if `config.editable` is `true` (or more specifically `config.edits.shapePosition`)? I thought if the old editable mode was enabled then for the moment the new one with activation would be available at all. If that's the case, this should say something like:
```
Has no effect when the older editable shapes mode is enabled via
`config.editable` or `config.edits.shapePosition`
```

src/components/shapes/draw.js
```diff
@@ -90,23 +108,74 @@ function drawOne(gd, index) {
...
+            path.style('pointer-events',
+                !couldHaveActiveShape(gd) || (
+                    lineWidth < 2 || ( // not has a remarkable border
+                        !isOpen && Color.opacity(fillColor) * opacity > 0.5 // not too transparent
+                    )
+                ) ? 'all' : 'stroke'
+            );
+        }
+
+        path.node().addEventListener('click', function() { return clickFn(path); });
```
alexcjohnson, line 174: Why the extra layer here, vs. just

```
path.node().addEventListener('click', function() { return activateShape(gd, path); });
```
?

src/components/shapes/draw.js
```diff
@@ -383,20 +464,30 @@ function setupDragElement(gd, shapePath, shapeOptions, index, shapeLayer) {
...
         } else {
-            var newN = (~dragMode.indexOf('n')) ? n0 + dy : n0;
-            var newS = (~dragMode.indexOf('s')) ? s0 + dy : s0;
-            var newW = (~dragMode.indexOf('w')) ? w0 + dx : w0;
-            var newE = (~dragMode.indexOf('e')) ? e0 + dx : e0;
+            var has = function(str) { return dragMode.indexOf(str) !== -1; };
+            var hasN = has('n');
+            var hasS = has('s');
+            var hasW = has('w');
+            var hasE = has('e');
```
alexcjohnson, line 471: Oh good, `~` is a bit esoteric, this is much cleaner and 🌴
 🎉

src/plots/cartesian/helpers.js
```diff
@@ -0,0 +1,57 @@
...
+* This source code is licensed under the MIT license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+
+'use strict';
+
+function getAxId(ax) {
+    return ax._id;
+}
```
alexcjohnson, line 14: I know you just moved this from one place to another, but to my mind it's harder to remember that this function exists somewhere than to just remember that axes have an `_id` attribute. Non-blocking, but I'd 🔪

src/plots/cartesian/new_shape.js
```diff
@@ -0,0 +1,917 @@
...
+        } else { // other polylines
+            cell[indexJ][1] = x0 + dx;
+            cell[indexJ][2] = y0 + dy;
+        }
+
+        redraw();
+    }
+
+    function endDragVertexController(evt) {
+        Lib.noop(evt);
```
alexcjohnson, line 157: ?

src/plots/cartesian/new_shape.js
```diff
@@ -0,0 +1,917 @@
...
+}
+
+function displayOutlines(polygons, outlines, dragOptions, nCalls) {
+    if(!nCalls) nCalls = 0;
+
+    var gd = dragOptions.gd;
+
+    function redraw() {
+        // recursive call
+        displayOutlines(polygons, outlines, dragOptions, nCalls++);
```
alexcjohnson, line 63: what's `nCalls`? Doesn't seem used. Debugging?

src/plots/layout_attributes.js
```diff
@@ -444,6 +445,112 @@ module.exports = {
         editType: 'modebar'
     },
 
+    newshape: {
```
alexcjohnson, line 448: Can we define these attributes somewhere in `components/shapes` and import them here?

src/plots/mapbox/mapbox.js
```diff
@@ -549,7 +558,7 @@ proto.updateFx = function(fullLayout) {
     var dragMode = fullLayout.dragmode;
     var fillRangeItems;
 
-    if(dragMode === 'select') {
+    if(rectMode(dragMode)) {
```
alexcjohnson, line 561: Why did this change? Wouldn't drawing modes just do nothing on mapbox? Same question on line 601 and a couple places in ternary.

src/plots/plots.js
```diff
@@ -1524,6 +1527,21 @@ plots.supplyLayoutGlobalDefaults = function(layoutIn, layoutOut, formatObj) {
     coerce('modebar.activecolor', Color.addOpacity(modebarDefaultColor, 0.7));
     coerce('modebar.uirevision', uirevision);
 
+    coerce('newshape.drawdirection');
```
alexcjohnson, line 1530: These too would be nice to move to `components/shapes` - and referenced with `Registry.getComponentMethod`.

test/jasmine/tests/draw_shape_test.js
```diff
@@ -0,0 +1,1364 @@
...
+    }
+}
+
+function fixDates(str) {
+    // hack to conver date axes to some numbers to parse with parse-svg-path
+    str = str.replace(/ /g, '');
+    str = str.replace(/_/g, '');
+    str = str.replace(/-/g, '');
+    str = str.replace(/:/g, '');
+    return str;
```
alexcjohnson, line 93:
```suggestion
    return str.replace(/[ _\-:]/g, '');
```

src/components/modebar/buttons.js
archmoj: Good call. Updated in https://github.com/plotly/plotly.js/pull/4775/commits/44dec73c5f682ef97454ea86a31c0ff4eb30d11a.

src/components/shapes/attributes.js
archmoj: Description is updated by https://github.com/plotly/plotly.js/pull/4775/commits/d99b44b414b8f90feca4f285a579d0cc6306f42f.

And here is the function that should control the two, I think.
https://github.com/plotly/plotly.js/blob/63ef7aaf311858101c7ed3075a40472eb6341cd1/src/components/shapes/draw.js#L74-L77.

src/components/modebar/manage.js
archmoj: I removed `ternary` and added a comment regarding `mapbox` in https://github.com/plotly/plotly.js/pull/4775/commits/1f196994c9cfff253a76f5af4af9dd7b69956732.

src/plots/cartesian/helpers.js
archmoj: Done in https://github.com/plotly/plotly.js/pull/4775/commits/a6662067ae8bb6b62f124f5cd0ed1272e1fcc00b.

src/plots/layout_attributes.js
archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/4775/commits/63ef7aaf311858101c7ed3075a40472eb6341cd1.

src/plots/plots.js
archmoj: Good call. Addressed in https://github.com/plotly/plotly.js/pull/4775/commits/63ef7aaf311858101c7ed3075a40472eb6341cd1.

src/components/shapes/draw.js
archmoj: Good eye. No more needed now. Fixed in https://github.com/plotly/plotly.js/pull/4775/commits/d7e24b83426f276f977bfa710dae3dd457c4d94d.

src/components/modebar/manage.js
alexcjohnson: OK fair enough

src/plots/cartesian/new_shape.js
archmoj: Good Call. Revised in https://github.com/plotly/plotly.js/pull/4775/commits/c5ff5dfeec859a1c46cfc1494155bab8a1270bd0.

src/plots/cartesian/new_shape.js
archmoj: Good Call. Now used in https://github.com/plotly/plotly.js/pull/4775/commits/c5ff5dfeec859a1c46cfc1494155bab8a1270bd0.

src/plots/mapbox/mapbox.js
archmoj: These plot types support `select` & `lasso`.
Drawing on `ternary` still needs a little offset fix somewhere which if out of scope; but drawing on the `mapbox` works (rather useful for highlighting some areas) i.e. in paper coordinate.

archmoj:
@alexcjohnson here is a [demo](https://codepen.io/MojtabaSamimi/pen/YzyQOaE?editors=0010) with shapes on template.

jackparmer:
Incredible work as always Mojtaba!!

emmanuelle:
Hurray! :tada: :heart_eyes:

pfebrer96:
Just wanted to say that this is amazing :)

I hope you don't mind this quick question: Are these changes automatically available in react-plotly.js? I updated plotly.py and it works, but when I try it on react I just get a white canvas. 

Thanks in advance and sorry if the question does not belong here.

nicolaskruchten:
Glad you like it!

In principle these should be available in react-plotly.js if you upgrade the version of plotly.js via `npm install` or `yarn install`, yes.

pfebrer96:
Makes sense :smile: 

I somehow thought that plotly.js would get updated automatically when I reinstalled react-plotly.js. All good now, thanks!

PR #4778, opened 2020-04-20
notifier: Use full font stack as elsewhere

edg2s:


archmoj:
@edg2s thanks for the PR.
:dancer:

PR #4783, opened 2020-04-22
Addressing transform sort order with null values

oshikryu:
Updating sort function to handle null value case
Changing sort transform tests to include null value

This should fix the issue in https://github.com/plotly/plotly.js/issues/4784

Example working codepen:
https://codepen.io/oshikryu/pen/rNOjeZb

### DIFF
--- a/src/transforms/sort.js
+++ b/src/transforms/sort.js
@@ -12,6 +12,8 @@ var Lib = require('../lib');
 var Axes = require('../plots/cartesian/axes');
 var pointsAccessorFunction = require('./helpers').pointsAccessorFunction;
 
+var BADNUM = require('../constants/numerical').BADNUM;
+
 exports.moduleType = 'transform';
 
 exports.name = 'sort';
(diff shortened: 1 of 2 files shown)


src/transforms/sort.js
```diff
@@ -136,8 +136,22 @@ function getIndices(opts, targetArray, d2c, len) {
...
     switch(opts.order) {
         case 'ascending':
-            return function(a, b) { return d2c(a.v) - d2c(b.v); };
+            return function(a, b) {
+                if(a.v === null) {
+                    return -1;
+                } else if(b.v === null) {
+                    return 1;
+                }
+                return d2c(a.v) - d2c(b.v);
```
alexcjohnson, line 145: Thanks for the PR @oshikryu - looking good, and your test additions are excellent!

I think we want to treat any sort of invalid data the same as `null` - which, after running through `d2c`, becomes `BADNUM` from [`src/constants/numerical`](https://github.com/plotly/plotly.js/blob/eca4b3bf7ce3731912dcc2243a42da8b01bc8395/src/constants/numerical.js#L19). `BADNUM` is just `undefined` but for clarity and consistency we prefer to use the named constant.

So this would look something like:
```js
var ac = d2c(a.v);
var bc = d2c(b.v);
if(ac === BADNUM) {
    return -1;
}
if(bc === BADNUM) {
    return 1;
}
return ac - bc;
```

Also I suppose this can be up for discussion, but in several other contexts (eg Dash DataTable) we sort null values always at the end - the idea being you want them out of the way where you mostly ignore them. Here you're sorting them as "extra negative", which puts them last when sorting descending, but FIRST when sorting ascending.

oshikryu: Good point about "extra negative" for `BADNUM` values. I added the reference to `BADNUM` in the sort and updated the tests

oshikryu:
I'm not really sure what downstream errors were produced with this change. I could use some guidance @alexcjohnson

PR #4793, opened 2020-04-30
Sanitize sourceattribution in mapbox layers

archmoj:
@plotly/plotly_js 
@scjody 

cc: https://github.com/plotly/streambed/issues/14452


test/jasmine/tests/svg_text_utils_test.js
```diff
@@ -512,3 +512,115 @@ describe('svg+text utils', function() {
         });
     });
 });
+
+describe('sanitizeHTML', function() {
+    'use strict';
+
+    describe('convertToTspans', function() {
```
alexcjohnson, line 519: we're not testing `convertToTspans` here, can we remove this level of nesting?

PR #4794, opened 2020-04-30
Fixup default angle of markers on scattermapbox

archmoj:
The `dflt` for `angle` should be set to `null` not to zero.
Fixes
https://github.com/plotly/plotly.js/pull/4575#issuecomment-622037809

@plotly/plotly_js


alexcjohnson:
My concern with `null` is that it won’t set the attribute at all in `fullData`, which is a problem for RCE, it won't know that this attribute should be available. Perhaps we can use `'auto'` as the default, and then strip it out before sending to the underlying mapbox library?

PR #4802, opened 2020-05-01
Rebuild package lock - May 2020

archmoj:
In respect to #https://github.com/plotly/plotly.js/issues/4796
This PR helps reduce the number of warnings namely for `kind-of` package.

Before: 
before

After:
after

See commit https://github.com/plotly/plotly.js/commit/a883997cc0a46284877372794ac2f1138075a0b5 for details.

Also to mention, the audit files were created using  the commands below:
```
npm audit --json > audit.json
npm audit | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" > audit.txt
```

@plotly/plotly_js

PR #4800, opened 2020-05-01
Dev install acorn, espree and mkdirp modules

archmoj:
In respect to #https://github.com/plotly/plotly.js/issues/4796
This PR helps reduce the number of warnings for `acorn`using `regl`.

Before: 
before

After:
after

See commit https://github.com/plotly/plotly.js/commit/cabe146d31410f2a6544df84cff9b80f746b7f61 for details.

Also to mention, the audit files were created using  the commands below:
```
npm audit --json > audit.json
npm audit | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" > audit.txt
```

@plotly/plotly_js

PR #4811, opened 2020-05-04
Downgrade acorn to v6 

archmoj:
Folowing #4800 and #4802, now using `acorn` v7 the `npm run preversion` step failed with the following message:
Screenshot from 2020-05-04 18-21-00

This PR fixes that by using latest v6 version i.e. without increasing audit warnings.

cc: @plotly/plotly_js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -289,9 +289,10 @@
       }
     },
     "acorn": {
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Dancer

PR #4814, opened 2020-05-05
Handle rangebreak gaps in `candlestick` & `ohlc`

### LINKED ISSUE #4742: Candlestick body width bug with range breaks
Hello everyone,

I am using plotly with python and found a bug with the new (very useful) rangebreaks feature.

When using candlestick graphs it generates some kind of *yaxis size reduction* on the body of every candles, resulting in what look likes **vertical lines** for **each candles**.

I created a codepen with and without rangebreaks as the only difference.

Thanks a lot for this great library,

[codepen link](https://codepen.io/matdecac/pen/JjYPwRv)

Mathias

### LINKED ISSUE #4795: Using hour rangebreaks causes candlestick chart to fail
Using an hour rangebreak, in my case `{'pattern': 'hour', 'bounds': [16, 9.5]}` (2nd last line) causes the candlestick chart to not plot open/close boxes and it breaks the hovering feature to preview data. There is also more space between points but I assumed that was because the lack of candle boxes.

Data:
[data.xlsx](https://github.com/plotly/plotly.py/files/4536643/data.xlsx)

Code:
`import pandas as pd
from plotly.offline import plot
import plotly.graph_objects as go
from plotly.subplots import make_subplots

df = pd.read_excel('data.xlsx')

#Plot Chart
fig = make_subplots(          
    rows=2, cols=1, 
    subplot_titles=('Price', 'Volume'),
    shared_xaxes=True,
    vertical_spacing=0.3)
fig.add_trace(go.Candlestick( #Candlesticks
    x=df['date'], open=df['open'], high=df['high'], low=df['low'], close=df['close']),
    row=1, col=1)
fig.add_trace(go.Bar( #Volume
    x=df['date'], y=df['volume'],
    marker_color='blue'),
    row=2, col=1,)
fig.update_layout(
    title_text='AAPL Technical Analysis',
    xaxis_rangeslider_visible=True,
    showlegend=False)
fig.update_xaxes(
    rangebreaks=[
        {'pattern': 'day of week', 'bounds': [6, 1]}, #Break weekends
        {'pattern': 'hour', 'bounds': [16, 9.5]}]) #Break after hours and market close

plot(fig)`

Remove `{'pattern': 'hour', 'bounds': [16, 9.5]}` to see correct candlesticks but I don't want to plot times where the market is after hours or closed.

Python 3.7.7, all modules are updated

Thanks

archmoj:
Fixes #4742 and fixes #4795 by hangling gaps in `distinctVals` function.
Demo1: [Before](https://codepen.io/MojtabaSamimi/pen/bGVaZeJ?editors=0010) | [After](https://codepen.io/MojtabaSamimi/pen/zYvpbKo?editors=1000)
Demo2: [Before](https://codepen.io/MojtabaSamimi/pen/GRpQjLq?editors=0010) | [After](https://codepen.io/MojtabaSamimi/pen/JjYpRQQ?editors=1000)

@plotly/plotly_js


test/image/mocks/axes_breaks-candlestick.json
```diff
@@ -0,0 +1,75 @@
+{
+    "data": [
+        {
+            "type": "candlestick",
+            "x": [
```
nicolaskruchten, line 5: in this mock, it would be great to see some data in the next visible range, like after 9am the next day, just to capture some of the tick labelling etc

src/lib/search.js
```diff
@@ -74,20 +75,39 @@ exports.distinctVals = function(valsIn) {
...
     vals.sort(exports.sorterAsc);
 
-    var l = vals.length - 1;
-    var minDiff = (vals[l] - vals[0]) || 1;
-    var errDiff = minDiff / (l || 1) / 10000;
-    var v2 = [vals[0]];
+    var first;
+    for(first = 0; first < vals.length; first++) {
+        if(vals[first] !== BADNUM) break;
+    }
```
alexcjohnson, line 81: Oh interesting, I wouldn't have guessed that the problem traced back to `distinctVals`, nice find!

TIL:
> all undefined elements are sorted to the end of the array, with no call to `compareFunction`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

So perhaps all we need is the loop to find `last` and the rest of this can stay as it was, as we'll never have `undefined` anywhere else after a `sort`?

alexcjohnson: maybe worth testing this in IE before relying on it. I've confirmed in Chrome, FF, Safari on my mac so seems robust on modern browsers.

If it were to fail on IE, I wouldn't trust that sorting works at all then with interspersed `undefined` values, and better would be to filter these out before even starting (instead of the `slice()` call)

archmoj: @alexcjohnson good catch! Not working on IE11.

archmoj: @alexcjohnson Correction. It appear to be working on IE11.
IE11

archmoj: @alexcjohnson correction number 2: It is sometimes working and sometime not.
So let's fix that for IE11.

archmoj:
Also we should possibly addess https://github.com/plotly/plotly.js/issues/4818 in this PR which seems related to applying `distinctVals`.

archmoj:
> Also we should possibly addess #4818 in this PR which seems related to applying `distinctVals`.

The `heatmap` bug is not related to `distinctVals` at all. So we are not going to fix that in this PR.

src/lib/search.js
archmoj: OK. The logic is modified by https://github.com/plotly/plotly.js/pull/4814/commits/2f3b098f78180ecb36ed36f27b96c35c13352d96 & tested in IE11 and it works.

test/image/mocks/axes_breaks-candlestick.json
archmoj: Good point. I added another mock in https://github.com/plotly/plotly.js/pull/4814/commits/8a4bae3333621ff4d75d1b82defe2613c959fbc8.
And we are going to revise tick behaviour (including these mocks) in #4734 after this PR is merged.

Issue #4818, opened 2020-05-07
plot & hover problems of heatmap (case of 2d-z array) with rangebreaks


archmoj:
[demo](https://codepen.io/MojtabaSamimi/pen/BaoraoP?editors=0010)
which is similar to: #4742  & #4795.

archmoj:
The problem is only with the case of 2d `z` array.
Case of 1d-z-array works fine: [demo](https://codepen.io/MojtabaSamimi/pen/dyYmGpg?editors=0010).

PR #4821, opened 2020-05-07
Fix rangebreaks on heatmap with 2d z array

archmoj:
Fixes: #4818 | demo: [before](https://codepen.io/MojtabaSamimi/pen/BaoraoP?editors=0010) vs [after](https://codepen.io/MojtabaSamimi/pen/qBOoZVB?editors=1000).

@plotly/plotly_js


test/image/mocks/axes_breaks-heatmap1d.json
```diff
@@ -0,0 +1,111 @@
...
+                "2020-01-02 18:00",
+                "2020-01-02 18:10",
+                "2020-01-02 17:00",
+                "2020-01-02 17:10",
+                "2020-01-02 17:20",
+                "2020-01-02 17:30",
+                "2020-01-02 17:40",
+                "2020-01-02 17:50",
+                "2020-01-02 18:00",
+                "2020-01-02 18:10"
```
alexcjohnson, line 69: These tests include some removed data, which is good - but can we also include some data on the other side of the break, ie the next morning?

alexcjohnson:
Does this fix also support `histogram2d` with rangebreaks?

test/image/mocks/axes_breaks-heatmap1d.json
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4821/commits/e2cf395aeec7ea1c2950a7579d157671f1c85fdb.

archmoj:
> Does this fix also support `histogram2d` with rangebreaks?

Not sure about that one. Is there any problem you noticed with `histogram2d` and `rangebreaks`?

alexcjohnson:
I haven't tried `histogram2d` with `rangebreaks` but I know it shares a lot of code with `heatmap` so seems natural to address them together.

archmoj:
I added tests for `contour` in https://github.com/plotly/plotly.js/pull/4821/commits/a7676bd2798c335add2abf052e230572eba69a35.
The `histogram2d` case is tracked in #4825 amd it would be addressed in a separate PR.

PR #4832, opened 2020-05-14
Fix react category order of matching axes

### LINKED ISSUE #4718: Data order is not applied by React when using matching axes
Originally reported in https://github.com/plotly/dash/issues/1163.
[MINIMAL DEMO](https://codepen.io/MojtabaSamimi/pen/GRJeVwd?editors=0010)

### LINKED ISSUE #4826: faceting react() bug
Sorry this isn't a more minimal codepen but there's something seriously wrong with the category orders here: https://codepen.io/nicolaskruchten/pen/rNOKjpE

Note: the JSON here is basically what you'd get from PX in a Dash app when changing some filters.

archmoj:
Fixes #4718 and fixes #4826 | [before](https://codepen.io/MojtabaSamimi/pen/PoPBgQa) vs [after](https://codepen.io/MojtabaSamimi/pen/eYpjoMN).

@plotly/plotly_js

### DIFF
--- a/src/lib/relink_private.js
+++ b/src/lib/relink_private.js
@@ -25,7 +25,7 @@ module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
         var toVal = toContainer[k];
 
         if(toVal === fromVal) continue;
-        if(k !== '_categories' && toContainer.matches) continue;
+        if(toContainer.matches && k === '_categoriesMap') continue;
 
         if(k.charAt(0) === '_' || typeof fromVal === 'function') {
             // if it already exists at this point, it's something
--- a/src/lib/relink_private.js
+++ b/src/lib/relink_private.js
@@ -24,9 +24,9 @@ module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
         var fromVal = fromContainer[k];
         var toVal = toContainer[k];
 
-        if(toVal === fromVal) {
-            continue;
-        }
+        if(toVal === fromVal) continue;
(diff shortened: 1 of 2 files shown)


src/lib/relink_private.js
```diff
@@ -24,9 +24,9 @@ module.exports = function relinkPrivateKeys(toContainer, fromContainer) {
         var fromVal = fromContainer[k];
         var toVal = toContainer[k];
 
-        if(toVal === fromVal) {
-            continue;
-        }
+        if(toVal === fromVal) continue;
+        if(k !== '_categories' && toContainer.matches) continue;
```
alexcjohnson, line 28: Wait, what's going on here? Am I missing something or won't this skip everything *except* `_categories`, when `matches` is defined?

Maybe what this is saying is we actually *can* skip every private key for `matches` axes?

archmoj: Oops! Good eye!
Fixed in https://github.com/plotly/plotly.js/pull/4832/commits/2295e505b7b8f52b31e23d0e4c7bdcd3f27db4e2.

archmoj:
What is hapening (I think) is that this block https://github.com/plotly/plotly.js/blob/4d383f2fceecc320d60a81593d66e0619b367c0d/src/plots/cartesian/set_convert.js#L877-L909 links the matching items and then those would be replaced by the relink function.
I added few more tests in https://github.com/plotly/plotly.js/pull/4832/commits/c2dc252d4c3ea64a30ee13fdb03b12499ebc2e3c.
& merging...

PR #4851, opened 2020-05-20
Fix setting array tickmode on date & log axes 

### LINKED ISSUE #4848: Explicitly setting tickmode to array disables tickvals in date axis
Initially reported in https://github.com/plotly/plotly.py/issues/2478

This codepen is a bit puzzling: https://codepen.io/nicolaskruchten/pen/MWaZBwa ... I'm explicitly setting `tickmode` to `array` and seeing the default tick labels. But if I comment out that `tickmode` line it does the right thing, even though the schema docs say that this configuration should imply `tickmode=array`!

archmoj:
This PR fixes #4848 | [Before](https://codepen.io/MojtabaSamimi/pen/MWaZLQd) vs [After](https://codepen.io/MojtabaSamimi/pen/ZEbVwxg).

It also allows using [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays) for `tickvals`.

@plotly/plotly_js


alexcjohnson:
Ha OK, so we're just removing the restriction on `axType`. Makes me wonder why that was there in the first place but that's probably lost to the mists of time... looks great and it does appear to work! One thing I was worried about is if on log axes `tickvals` would need to be in linearized form (0,1,2), but they're in data form (1,10,100) as they should be 🎉 

Let's add a mock showing both axis types - here's your codepen modified to include log, that would be sufficient I think: https://codepen.io/alexcjohnson/pen/ExVrYrP - then it would be ready to go!
Screen Shot 2020-05-20 at 10 53 48 PM

PR #4861, opened 2020-05-22
Bump node sass, espree v7 and karma v5

archmoj:
Reduces dev-dependency warnings.

Before: 
before

After:
after

@plotly/plotly_js


package-lock.json
```diff
@@ -9522,62 +9639,142 @@
...
+        "get-caller-file": {
+          "version": "2.0.5",
+          "resolved": "https://registry.npmjs.org/get-caller-file/-/get-caller-file-2.0.5.tgz",
+          "integrity": "sha512-DyFP3BM/3YHTQOCUL/w0OZHR0lpKeGrxotcHWcqNEdnltqFwXVfhEBQ94eIo34AfQpo0rGki4cyIiftY06h2Fg==",
+          "dev": true
+        },
+        "is-fullwidth-code-point": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-2.0.0.tgz",
+          "integrity": "sha1-o7MKXE8ZkYMWeqq5O+764937ZU8=",
```
antoinerg, line 9700: I wonder why this one is hashed using `sha1`?

cc @archmoj

antoinerg:
:woman_dancing:

package-lock.json
archmoj: I think this is related to the way each module was published. Some older modules appear to use `sha1` instead of `sha512`; but I may be wrong.

Issue #4863, opened 2020-05-25
The `preprocess` is broken.


cy6eria:
**Problem**: The `npm run preprocess` fails with message: "TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Object".
**OS**: Mac OS 10.15.4
**Node**:  14.2.0
**Version**: With HEAD on 5010b80dcfdb66a2d6ce8a6650ee7982a5f01a73.

The cause of the problem is in the `tasks/util/update_version.js` The `falafel` package returns an Object, but `fs.writeFile` expect `string`, `Buffer`, `TypedArray` or `DataView`.

archmoj:
That may be related to the node version. Have you tried using node 12 instead?

cy6eria:
@archmoj hi, i almost sure that it is. I tested with 12.16.3 and it works. I've created a simple fix and it doesn't affect the script on 12.16.3, but works with 14.2.0.

gvwilson:
Hi - we are trying to tidy up the stale issues and PRs in Plotly's public repositories so that we can focus on things that are still important to our community. Since this one has been sitting for several years, I'm going to close it; if it is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. Thanks for your help - @gvwilson

PR #4864, opened 2020-05-25
The `preprocess` script fix

cy6eria:
This PR belongs to my #4863 issue.

### DIFF
--- a/tasks/util/update_version.js
+++ b/tasks/util/update_version.js
@@ -11,7 +11,7 @@ module.exports = function updateVersion(pathToFile) {
             if(isVersionNode(node)) node.update('\'' + pkg.version + '\'');
         });
 
-        fs.writeFile(pathToFile, out, function(err) {
+        fs.writeFile(pathToFile, out.toString(), function(err) {
             if(err) throw err;
         });
     });


archmoj:
In regard with https://github.com/plotly/plotly.js/issues/4863#issuecomment-633587743, right now we use node 12, not 14. So I am not sure if we could move forward with this PR at this moment.
cc: @alexcjohnson

alexcjohnson:
Thanks for the investigation and PR @cy6eria!

@archmoj The tests pass in this PR, do they exercise `preprocess` / `update_version`? If this patch works on both node 12 and node 14, it seems like a good idea to accept it, as it will make things easier once we're ready to bump to 14. If it breaks our workflow on 12 then we can leave this PR open and use it as the base for our eventual upgrade to 14.

tasks/util/update_version.js
```diff
@@ -11,7 +11,7 @@ module.exports = function updateVersion(pathToFile) {
             if(isVersionNode(node)) node.update('\'' + pkg.version + '\'');
         });
 
-        fs.writeFile(pathToFile, out, function(err) {
+        fs.writeFile(pathToFile, out.toString(), function(err) {
```
archmoj, line 14: Let's test for the type of `out` to be `object` before casting to string.

cy6eria: Hey! Sorry for a long answer. I don't think that it's necessary. It doesn't matter what `out` is. It inherit `toString()` method from `Object`. Even if it's string it has a `toString` method.

PR #4882, opened 2020-05-28
Bump fuse.js v6

archmoj:
@plotly/plotly_js

### DIFF
--- a/devtools/test_dashboard/devtools.js
+++ b/devtools/test_dashboard/devtools.js
@@ -2,7 +2,7 @@
 
 /* global Plotly:false */
 
-var Fuse = require('fuse.js');
+var Fuse = require('fuse.js/dist/fuse.common.js');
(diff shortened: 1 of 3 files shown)


devtools/test_dashboard/devtools.js
```diff
@@ -160,7 +160,17 @@ setInterval(function() {
...
+    // isCaseSensitive: false,
+    // includeScore: false,
+    // shouldSort: true,
+    // includeMatches: false,
+    // findAllMatches: false,
+    // minMatchCharLength: 1,
+    // location: 0,
+    // threshold: 0.6,
+    // distance: 100,
+    // useExtendedSearch: false,
```
alexcjohnson, line 173: Are these comments here for a reason?

archmoj: Simply taken from `fuse.js` docs, illustrating options and their defaults.

PR #4900, opened 2020-06-04
Fix bar lengths in milliseconds from base

archmoj:
Fixes #/4558.

[demo 1: Gantt chart example](https://codepen.io/MojtabaSamimi/pen/pogvWvx?editors=0010)
[demo 2: before vs after](https://codepen.io/MojtabaSamimi/pen/zYrxEYV?editors=1000)

@plotly/plotly_js


src/traces/bar/calc.js
```diff
@@ -19,12 +19,14 @@ module.exports = function calc(gd, trace) {
...
     if(trace.orientation === 'h') {
-        size = xa.makeCalcdata(trace, 'x');
-        pos = ya.makeCalcdata(trace, 'y');
+        size = xa.makeCalcdata(trace, 'x', msUTC);
+        pos = ya.makeCalcdata(trace, 'y', msUTC);
     } else {
-        size = ya.makeCalcdata(trace, 'y');
-        pos = xa.makeCalcdata(trace, 'x');
+        size = ya.makeCalcdata(trace, 'y', msUTC);
+        pos = xa.makeCalcdata(trace, 'x', msUTC);
```
alexcjohnson, line 29: `pos` should not get `msUTC`, only `size`. This would matter only in a very weird situation: The position axis has type date and data as milliseconds; this should be treated as local milliseconds whether or not the data has a `base` applied along the size axis (which may have any type, date or numeric).

src/plots/cartesian/set_convert.js
```diff
@@ -103,7 +103,15 @@ module.exports = function setConvert(ax, fullLayout) {
...
-                ms = dateTime2ms(new Date(msRounded)) + msecTenths / 10;
+
+                var d = new Date(msRounded);
+                ms = dateTime2ms(d) + msecTenths / 10;
+                if(msUTC) {
+                    // For now it is only used
+                    // to fix bar length in milliseconds.
+                    // It could be applied in other places in v2
+                    ms += d.getTimezoneOffset() * ONEMIN;
+                }
```
alexcjohnson, line 114: Can't we just (right after `v = +v`) say `if(msUTC) return v;` and skip the whole `dateTime2ms` step entirely in this case? Otherwise aside from the overhead there will be a subtle bug for very large negative values #2747

src/traces/bar/calc.js
archmoj: Oops! Fixed in https://github.com/plotly/plotly.js/pull/4900/commits/2ae0fdfb53dfd3ce600347a2924f25793c4a7b5d.

src/plots/cartesian/set_convert.js
archmoj: Good call. Done in https://github.com/plotly/plotly.js/pull/4900/commits/2ae0fdfb53dfd3ce600347a2924f25793c4a7b5d.

archmoj: Actually the modified file was not saved. Let me push another commit...

archmoj: Fixed by https://github.com/plotly/plotly.js/pull/4900/commits/bb5bec34bd03214b4e026a6fc0f054379fd2f009.

test/jasmine/tests/bar_test.js
```diff
@@ -2600,6 +2600,26 @@ describe('bar hover', function() {
...
+            .catch(failTest)
+            .then(done);
+        });
+
+        it('should display the correct bar length passed in milliseconds from base', function() {
+            var out = _hover(gd, 0.5, 0.75, 'y');
+
+            var xEnd = out.style[2];
+            expect(xEnd).not.toBe(1580670000000);
+            expect(xEnd).toBe(1580688000000);
```
alexcjohnson, line 2620: If you're testing `toBe`, what do you gain from testing `not.toBe`? 🤔

archmoj: Just to ensure no one change the test.

PR #4903, opened 2020-06-05
Fix gl3d ticks when converting dates to milliseconds

### LINKED ISSUE #4902: gl3d ticks are off by timezone
[Bug demo](https://codepen.io/MojtabaSamimi/pen/QWywJdW?editors=0010).
cc: #3306.

archmoj:
Supersedes #3721 and fixes #4902.
[Demo: Before vs After](https://codepen.io/MojtabaSamimi/pen/BajybOB?editors=0010).

@plotly/plotly_js


src/plots/cartesian/set_convert.js
```diff
@@ -90,7 +90,7 @@ module.exports = function setConvert(ax, fullLayout) {
      * - inserts a dummy arg so calendar is the 3rd arg (see notes below).
      * - defaults to ax.calendar
      */
-    function dt2ms(v, _, calendar, msUTC) {
+    function dt2ms(v, _, calendar, opts) {
```
alexcjohnson, line 93: good call. At some point I bet we could collapse all these extra args into a single `opts` arg (ie `dt2ms(v, opts)`) but that's a big and potentially risky project - which some users may consider a breaking change though the `ax.a2b` functions are sufficiently "internal" that I personally wouldn't be bothered by it. Anyway not a change to make now.

PR #4910, opened 2020-06-09
bump and relax ndarray version

archmoj:
@plotly/plotly_js


Frick-David:
is this ndarrary from numpy? why is this in plotly.js I am super curious as I have been passing ndarrays from my python backend to plotly.js

PR #4925, opened 2020-06-16
only take container computed size in px

alexcjohnson:
When you plot into a hidden div, `window.getComputedStyle` cannot resolve width and height into pixels if they are given in some other form (for example `'100%'`) so it leaves the original value. We shouldn't try to use that value, instead we should fall back on the schema default height & width in this case.

@archmoj this needs to go in the upcoming Dash release, so may require its own patch release if 1.54.3 isn't available in the next day or so. (sorry for the rush!)


archmoj:
Nicely done.
The CI as well as `no_ci` tests are also passed.
[ :dancer: ] | :dancer: | :dancer:

PR #4930, opened 2020-06-18
drop cwise and inline ndarray-homography

### LINKED ISSUE #4796: Security warnings in static-module and minimist
We need to resolve https://github.com/plotly/plotly.py/issues/2385 and https://github.com/plotly/jupyterlab-chart-editor/issues/47 by avoiding https://www.npmjs.com/advisories/758 and https://www.npmjs.com/advisories/1179

archmoj:
The follow up of #4929, 

zero

no longer depend on [cwise](https://www.npmjs.com/package/cwise) resolves #4796 
also to simplify the build process 
and optimized for 2d operations as well as inverting the input matrix once.

@plotly/plotly_js


src/traces/surface/convert.js
```diff
@@ -333,18 +378,17 @@ proto.refineCoords = function(coords) {
     var padWidth = 1 + width + 1;
     var padHeight = 1 + height + 1;
     var padImg = ndarray(new Float32Array(padWidth * padHeight), [padWidth, padHeight]);
+    var X = invert([], [
+        scaleW, 0, 0,
+        0, scaleH, 0,
+        0, 0, 1
+    ]);
```
alexcjohnson, line 385: http://mathonline.wikidot.com/inverse-of-diagonal-matrices seems like just `1/scaleW` and `1/scaleH` then we can 🔪 `invert` too?

alexcjohnson:
Fantastic work! Can we add a test to 🔒 the "`found 0 vulnerabilities`" audit result?

src/traces/surface/convert.js
archmoj: Good call! I also removed some loops in https://github.com/plotly/plotly.js/pull/4930/commits/db6f9e7efe643bdab2df3fc168296e5c71158b1b.

archmoj:
> Fantastic work! Can we add a test to  the "`found 0 vulnerabilities`" audit result?

Hmm.. not in this PR. But let's think about it.

nicolaskruchten:
> Can we add a test to the "found 0 vulnerabilities" audit result?

this will fail due to exogenous reasons all the time and block unrelated PRs unfortunately I think.

alexcjohnson:
> this will fail due to exogenous reasons all the time and block unrelated PRs unfortunately I think.

I guess. Can we put it in [`noci_test.sh`](https://github.com/plotly/plotly.js/blob/master/tasks/noci_test.sh) then, so we at least catch any new issues before making a release? It would be a shame to go through all this work and not be able to say future releases are all clean as of their release date. That would still give us the option to release with a known vulnerability if fixing it is more than a small task, but it would be a conscious decision at least.

archmoj:
OK. We could add 'npm audit' step after 'npm ci' to the release checklist.

PR #4988, opened 2020-07-08
Bump and control the version of lodash in dev-dependencies

archmoj:
Addressing 12 audit warnings https://www.npmjs.com/advisories/1523.

@alexcjohnson

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,7 @@
     "karma-spec-reporter": "0.0.32",
     "karma-verbose-reporter": "0.0.6",
     "karma-viewport": "^1.0.6",
+    "lodash": "^4.17.18",
     "madge": "^3.9.2",
     "minify-stream": "^2.0.1",
     "minimist": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "karma-spec-reporter": "0.0.32",
     "karma-verbose-reporter": "0.0.6",
     "karma-viewport": "^1.0.6",
-    "lodash": "^4.17.18",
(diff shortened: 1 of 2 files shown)


archmoj:
Wondering if we should wait until https://github.com/lodash/lodash/issues/4844 is resolved?

package.json
```diff
@@ -152,6 +152,7 @@
     "karma-spec-reporter": "0.0.32",
     "karma-verbose-reporter": "0.0.6",
     "karma-viewport": "^1.0.6",
+    "lodash": "^4.17.19",
```
alexcjohnson, line 155: do we need to add it here? Isn't it sufficient to have lodash bumped in `package-lock.json`?

alexcjohnson:
> Wondering if we should wait until lodash/lodash#4844 is resolved?

Seems like if our builds and tests run, that issue isn't a problem for us, right? And anyway it's not clear yet that they're in any rush to fix it.

package.json
archmoj: I think it is better to keep it this way e.g. in case one wants to regenerate the lock as well the for the yarn upgrade.

alexcjohnson: I just don't want us to be in the habit of including sub-dependencies in `package.json`, is this one special in some way? Are we waiting for something that will let us remove this?

archmoj: Until all those sub-dependencies switch to force using newer version, I think we need to control the version at root.
We have few other dev-dependencies e.g. `acorn`, `static-eval` that controlled like this.

alexcjohnson: OK, fair enough.

PR #4993, opened 2020-07-09
Implement period ticklabelmode on cartesian date axes

### LINKED ISSUE #4911: (x|y)axis.ticklabelmode = instant | period
This issue is scoped to 2D `type = date` axes when there is no `ticktext`. 

We need to add a new attribute (proposed name `ticklabelmode`) whose default value (`instant`) describes the current behaviour of our axes.

The new behaviour (`period`) would cause the tick label (not the tick or gridline!) to be positioned at the midpoint of the period that its current formatting represents (note: this means that it's sensitive to `tickformat` and `tickformatstops`). So for example if the ticklabel is `Jan` then instead of positioning it at `YYYY-01-01 00:00` (the 'instant') we would position it at `YYYY-01-15 12:00` (the midpoint of the 'period').

archmoj:
Resolves #4911.
[Demo](https://codepen.io/MojtabaSamimi/pen/vYLjqeV)

@plotly/plotly_js


src/plots/cartesian/layout_attributes.js
```diff
@@ -478,6 +478,20 @@ module.exports = {
             'to the left/bottom of labels.'
         ].join(' ')
     },
+    ticklabelmode: {
+        valType: 'enumerated',
+        values: ['', 'period'],
+        dflt: '',
```
alexcjohnson, line 484: Per https://github.com/plotly/plotly.js/issues/4911 can we make the default be `'instant'` instead of `''`?

src/plots/cartesian/layout_attributes.js
```diff
@@ -478,6 +478,20 @@ module.exports = {
...
+        role: 'info',
+        editType: 'ticks',
+        description: [
+            'Determines where tick labels are drawn with respect to their',
+            'corresponding ticks and grid lines.',
+            'Only has an effect for axes of `type` *date*',
+            'When set to *period*, tick labels are drawn in the middle of the period',
+            'between ticks.'
```
nicolaskruchten, line 492: this is actually not quite the desired behaviour: if the ticks are spaced every 2 months and the labels are `Jan` and `Mar`, the `Jan` label should appear on Jan 15, *not* on Feb 1.

nicolaskruchten: Here's an example... https://codepen.io/nicolaskruchten/pen/ZEQRBWw?editors=0010 

In "instant" mode we see 4 "month" labels spaced two months apart. In "period" mode we probably should see only 3 labels, but they should remain Jan/Mar/May, and they should be positioned on Jan 15, Mar 15, May 15

nicolaskruchten: I'll add a clearer codepen in the original issue

alexcjohnson:
I said I was going to comment on #4911 with some more details, and I never did that, my apologies. But now that there's a PR, let me comment here in the context of your new mock. Here's the original - that has the same data and ranges as your new mock - thanks for using the same data, this makes it very clear what's going on!
Screen Shot 2020-07-10 at 9 08 51 AM
And here it is modified with `ticklabelmode: 'period'`:
Screen Shot 2020-07-10 at 9 10 23 AM

First, it looks like the labels moved the wrong direction - ie right now the "1950" label appears between the 1900 and 1950 tick marks, it should be between 1950 and 2000.

Second, as @nicolaskruchten [noted](https://github.com/plotly/plotly.js/pull/4993#discussion_r452836768) we don't want the labels in the middle of the range, but in the middle of the period DENOTED BY THAT LABEL. So the label "2000" doesn't go halfway between the ticks for 2000 and 2050, it goes halfway between the beginning of 2000 and the beginning of 2001. The important thing here is what's the smallest increment of time included in the label? If it's years, the label shifts half a year into the future. If it's months, the label shifts half a month. If it's days, the label shifts half a day. It doesn't matter if `dtick` is 2 or 3 or 6 months, or even 10 years - if the label specifies months and nothing smaller, it shifts by half a month.

This needs to work whether the tick format is automatic, specified by `tickformat`, or pulled out of `tickformatstops`, which likely means we're going to have to parse the relevant format string and find the smallest time unit specifier it includes. There's an edge case here that @nicolaskruchten brought up in conversation with me: what if `dtick` is smaller than this smallest date part? Like someone specifies `tickformat='%b'` (months) and then zooms in to days? It'll look weird, as we'll have repeated "Jan" tick labels, but I think the thing to do here is fall back on half of `dtick` - ie the label shift is half the SMALLER of: the smallest unit displayed or `dtick`.

Third, and this is the part I've discussed now separately with both @archmoj and @nicolaskruchten but never written down, I think `'period'` should cease to apply once you get to minutes precision, and we treat such times as always instants. I think this has practical benefits, in that we display minutes very often when one minute is a quite small fraction of the range, so shifting by half a minute would end up just looking like we misaligned the label; and conceptually once you start displaying times like "12:00" my feeling is people naturally read these as instants and not periods.

What about hour precision? It's not part of our default formatting, that goes straight from days to minutes, so you won't get it by zooming in on a graph with automatic labels. You'll only see this if you explicitly specify a `tickformat` or `tickformatstop` with hours as the smallest unit. My feeling is people do think of "the 12 o'clock hour" as a period so if they've gone to the trouble of specifying this AND `ticklabelmode='period'` we should honor it.

src/plots/cartesian/layout_attributes.js
archmoj: We could consider it for sure.
However, I wanted to avoid testing strings here https://github.com/plotly/plotly.js/blob/4292f08339470f00e6266698f842bd30fb009145/src/plots/cartesian/axes.js#L2535 which may slow down a bit.

alexcjohnson: Yes, let's use `'instant'`. That particular test isn't in a very hot path, and I don't think an exact string match is all that much overhead anyway.

archmoj: Done in https://github.com/plotly/plotly.js/pull/4993/commits/9f3e1db2b3475e54db5492a57fc087efb3797fb4.

archmoj:
Changes made in https://github.com/plotly/plotly.js/pull/4993/commits/0e124c2f9c8e1564d505e286badc12aa21b69309 are reflected in the codepen at PR description.

alexcjohnson:
Looking much better! The day and month label offsets look perfect. The only issue I see there is that sometimes there should be an extra label at the beginning (when the tick is off-plot but the label position is on-plot), and sometimes there should NOT be a label at the end (when the tick is on-plot and the label position is off-plot). See the top row in the below image.

Year tick labels have this same issue, but sometimes it looks like some of the labels get offset by half a year while others get offset by half a month? See the bottom row in the below image. When you drag this axis around sometimes a single tick will shift back and forth between half-month and half-year offsets, when it's near the end of the range.

And when minutes or smaller are shown, the labels still offset. This shift should not happen at all in that case. That's the middle three rows.

Screen Shot 2020-07-20 at 4 27 20 PM

Finally, if you set an explicit `tickformat`, the smallest field included in the format should determine the offset, NOT `dtick`. See the image below where I set `tickformat: "%Y %b %d"` so the offset should be half a day, despite `dtick` being one month.

Screen Shot 2020-07-20 at 4 42 02 PM

src/plots/cartesian/axes.js
```diff
@@ -620,6 +619,16 @@ axes.calcTicks = function calcTicks(ax, opts) {
...
     generateTicks();
 
+    var isPeriod = ax.ticklabelmode === 'period';
+    var addOneTickToStart = isPeriod && tickVals.length > 2;
+
+    if(addOneTickToStart) {
+        tickVals.push({
+            minor: false,
+            value: 2 * tickVals[0].value - tickVals[1].value
+        });
```
alexcjohnson, line 629: Looking good! Maybe this is what you were referring to this morning, but I see a little funny behavior with the first label: when the first tick is present, the first label is correct, but when the first tick is missing, the first label is sometimes incorrect. I suspect this can be fixed by using `tickIncrement` and inverting the 3rd arg, something like:
`value: axes.tickIncrement(tickVals[0].value, ax.dtick, !axrev, ax.caldendar)`

Should be 1886, second one is correct:
Screen Shot 2020-07-23 at 4 06 10 PM
Screen Shot 2020-07-23 at 4 06 39 PM

Should be 1887, second one is correct:
Screen Shot 2020-07-23 at 4 08 55 PM
Screen Shot 2020-07-23 at 4 09 07 PM

And with months, May is correct:
Screen Shot 2020-07-23 at 4 14 15 PM
Screen Shot 2020-07-23 at 4 14 28 PM

Day tick labels look to have the correct values, but I notice a first one with no tick doesn't get the year, it goes on the original first tick, which is now the second one. Maybe this is as simple as putting this tick first rather than last in the array? ie `unshift` rather than `push`?
Screen Shot 2020-07-23 at 4 16 47 PM

archmoj: Thanks for the review @alexcjohnson 
Let me give this part a second try / tries.

archmoj: @alexcjohnson I applied your suggestions in https://github.com/plotly/plotly.js/pull/4993/commits/79385c24e5fe80aff8008f8818a11582146031a4 and https://github.com/plotly/plotly.js/pull/4993/commits/66a98be64f76ecfb43ed04da72500251f4bbe694.

src/plots/cartesian/axes.js
```diff
@@ -620,6 +620,15 @@ axes.calcTicks = function calcTicks(ax, opts) {
 
     generateTicks();
 
+    var isPeriod = ax.ticklabelmode === 'period';
+    if(isPeriod) {
+        // add one label to show pre tick0 period
+        tickVals = [{
+            minor: false,
+            value: axes.tickIncrement(tickVals[0].value, ax.dtick, !axrev, ax.caldendar)
+        }].concat(tickVals);
```
alexcjohnson, line 629: not a big deal, but `tickVals.unshift({...})` is slightly more efficient here.

archmoj: Good call. 
I also fixed `reversed` range positioning and added few tests in https://github.com/plotly/plotly.js/pull/4993/commits/83f5cddbcbdac4f5c36f70f043a9156aaa1c0829.

PR #4994, opened 2020-07-09
Fix logging and notifyOnLogging to work with Plotly.newPlot calls

### LINKED ISSUE #4555: Make notify-on-logging work with newPlot
Right now we can't use this function from Python because we only set `config` during `newPlot`.

archmoj:
Fixes #4555 thanks to https://github.com/plotly/plotly.js/issues/4555#issuecomment-597776722.

TODO: 
- [ ] test this from python
- [x] add a jasmine test

@plotly/plotly_js


nicolaskruchten:
If we're concerned about this refactoring shaking something loose at runtime that causes silent/subtle failures, could we try adding some extra runtime checks at the top of all the function definitions that we're changing which error out if the first argument is not as expected, and then run the test suite? That might flush out some extra silent failures? We wouldn't commit this code into master but as an extra one-time safety check?

Otherwise, would this be a safer change if we added an argument instead of changing the first one?

archmoj:
> If we're concerned about this refactoring shaking something loose at runtime that causes silent/subtle failures, could we try adding some extra runtime checks at the top of all the function definitions that we're changing which error out if the first argument is not as expected, and then run the test suite? That might flush out some extra silent failures? We wouldn't commit this code into master but as an extra one-time safety check?
> 
> Otherwise, would this be a safer change if we added an argument instead of changing the first one?

@nicolaskruchten thanks very much for the feedback.

Adding this new `gd` parameter to the end of function arguments is not very helpful; while those functions may be called without full set of arguments.

On the other hand checking the first argument in the test looks like a good idea; although in the end we may not catch anything new since those tests are passing already. Anyway I could give it a try.

archmoj:
Notes by @alexcjohnson:

> I’m a bit uncomfortable with the logging PR - seems like it’s passing gd essentially everywhere, when there are really only a few places we call log/warn/error.
> 
> One thing we could do to make it easier is if you have access to any DOM node, allow the log functions to search its parents for a _context
> 
> And some of them, like randstr, we could probably just get rid of the log message, turn it into throwing an error.

archmoj:
> Notes by @alexcjohnson:
>
> > One thing we could do to make it easier is if you have access to any DOM node, allow the log functions to search its parents 

Searching DOM elements for `gd` is little hacky (in my opinion) and could add to the complexity of this PR.

PR #5010, opened 2020-07-22
Implement horizontal and vertical markers for arrow charts

### LINKED ISSUE #4935: Triangle markers that end on points rather than center on them
In order to make an "arrow chart" out of a scatter trace, we need to be able to have arrowhead/triangle markers whose extreme end is at the point, rather than markers centered on a point: right now if I want an arrow from 10 to 20, the tip of the triangle overshoots.

I propose we add four new markers `arrowhead-*` for this purpose.

image

archmoj:
Resolves #4935 | [demo](https://codepen.io/MojtabaSamimi/pen/dyGamry?editors=0010)

List of new symbols:
```
'arrow-up'
'arrow-down'
'arrow-left'
'arrow-right'
'arrow-bar-up'
'arrow-bar-down'
'arrow-bar-left'
'arrow-bar-right'
```

@plotly/plotly_js


alexcjohnson:
Very nice! Just two comments:
- These should have `noDot: true`, and the bar variants should have `needLine: true` (now, whether we're actually doing what we should with those flags is another matter...)
- Instead of `arrow-to-bar-up` can we shorten to just `arrow-bar-up`?

PR #5011, opened 2020-07-22
Keep hidden grid lines of oriented traces when color is defined by template

### LINKED ISSUE #4980: Template quirk
A strange effect of #4904:
- If you want to use the template to override default colors for axis lines that are hidden by default, now you need to explicitly hide those lines.
- We decided in that PR that this behavior is more self-consistent, because that's how the regular figure attributes work: if you provide a color, it makes the line visible by default. So the template should do that too, otherwise it has a strange difference vs a regular figure.
- EXCEPT that templates are often used just to provide new defaults but the intent is not to override other parts of the logic. In the case of axis lines this has two undesirable effects:
  - Sometimes the data itself causes the lines to show. For example bar charts by default turn on the zero line. But now if you have a template that recolors the zero line and then turns it back off, bar charts won't be able to automatically turn it back on.
  - If the user provides a new color for the line in the main figure, it won't cause the line to turn on, as it would without the template present.

Both of these are demonstrated in https://codepen.io/alexcjohnson/pen/VweyRJe?editors=1010 - without the template, both the xaxis line and the x=0 line show:
Screen Shot 2020-07-03 at 3 03 07 PM
But with the template they both disappear:
Screen Shot 2020-07-03 at 3 03 23 PM

Since there *is* a clear use case for both variants, I kind of think we need a new concept in the template, of a "new default" value, that doesn't count as having provided a value, so can't affect any other logic.

@archmoj @nicolaskruchten @wbrgss thoughts?

archmoj:
Fixes #4980 | [demo](https://codepen.io/wbrgss/pen/GRozJza)

@plotly/plotly_js 
cc: @wbrgss 

TODO: 
- [x] add test


wbrgss:
At first glance this seems to solve the problem I described in https://github.com/plotly/plotly.js/issues/4980#issuecomment-662108834 out-of-the-box (i.e. without changes to the relevant template)

I'm using https://82470-45646037-gh.circle-artifacts.com/0/dist/plotly.min.js from the `publish` step of this PR

Much appreciated @archmoj!

archmoj:
:1st_place_medal: @wbrgss thanks very much for tying that out.
To catch similar issues in future, I think, it may be a good idea if we could add a number of mocks (for image test on the plotly.js repo) that use the exact templates applied by DDK.

alexcjohnson:
Per our discussion just now: In the short term this leaves too many open questions for us to feel confident about the solution. So the steps @archmoj is going to take are:
- Revert #4904 and close this PR
- Release a new patch version
- Open a new PR that combines #4904 and this PR, to be used as a basis for discussion about how to make a coherent and complete fix for the original `Lib.coerce2` issues that we identified and support all relevant use cases.

archmoj:
Closing not that #5016 is open.

